PR #7874, opened 2016-03-02
[rgw] RGWLib::env is not used so remove it

badone:
Coverity complains about RGWLib::env.port being uninitialised but
RGWLib::env is not used at all so just remove it.


badone:
@dalgaaf FYI

mattbenjamin:
weird--ok

PR #7876, opened 2016-03-02
hammer: redhat-lsb-core dependency was dropped, but is still needed

smithfarm:
http://tracker.ceph.com/issues/14946


dachary:
It passed a run of the upgrade suite : http://tracker.ceph.com/issues/14692#note-31

PR #7877, opened 2016-03-02
infernalis: redhat-lsb-core dependency was dropped, but is still needed

smithfarm:
http://tracker.ceph.com/issues/14945


dachary:

PR #7882, opened 2016-03-02
mon: unconfuse object count skew message

branch-predictor:
"Pool <pool> has too few pgs" is okay assuming it does not take other
pools into account. And since it does, it is confusing in the following
scenario:

1. Create two pools, one with small pg count and one with large
   pg count
2. Put a whole lot of objects in smaller pool, resulting in "too few
   pgs" warning on that pool, which is expected behavior.
3. Put a whole lot of objects in larger pool, warning goes away.
   Suddenly smaller pool has plenty of PGs?

Current message suggests adding more nodes (or PGs) to pool, when
actually it's warning about significantly more objects in that
particular pool than in the other pools.


liewegas:
how about "has many objects per pg (too few pgs?)" ?

branch-predictor:
"has many more objects per pg than average (too few pgs?)"?

liewegas:
sounds good

branch-predictor:
Done.

PR #7884, opened 2016-03-02
librbd: allocate new journal tag after acquiring exclusive lock

dillaman:
The journal tags are used to delineate epochs within an image.


jdurgin:
lgtm

dillaman:
Fixed compile errors against rebased master

PR #7885, opened 2016-03-02
os/filestore/FileJournal: set block size via config option

liewegas:
We were setting the block_size as the MIN of the min size (4096) and the
block size reported for the journal file/device.  In reality, file systmes
report all kinds of crazy block sizes.  Usually it's the page size, but
sometimes it is larger (e.g., 128KB for ZFS), and that's not actually what
we want.  Using a size smaller than the file systems block size is not
optimal for performance, but it doesn't affect how the IO happens--as long
as it is larger than the hardware sector size, which is either 512 or
4096 bytes.  And our min was hard-coded at 4096.

So, instead, just set a config option to specify teh block size, and
default that to 4096.

The other uses of this constant we about *alignment* of memory buffers
for the purposes of direct IO.  Rename the constant but do not change
the logic.  That means we continue to use 4k alignment for direct io even
if the device has 512 byte sectors, but that's fine--no reason to use
a smaller alignment.

(cherry picked from commit c8048ce4bedf0d1ba5d448795c46668f2724f29d)

### DIFF
--- a/src/common/config_opts.h
+++ b/src/common/config_opts.h
@@ -916,6 +916,7 @@ OPTION(filestore_debug_verify_split, OPT_BOOL, false)
 OPTION(journal_dio, OPT_BOOL, true)
 OPTION(journal_aio, OPT_BOOL, true)
 OPTION(journal_force_aio, OPT_BOOL, false)
+OPTION(journal_block_size, OPT_INT, 4096)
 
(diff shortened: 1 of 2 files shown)


liewegas:
This is needed for infernalis because it switched from following the page size (possiblye 64k) to forcing 4k, breaking actual clusters.

PR #7889, opened 2016-03-02
XIO: Isolate the frequently accessed atomic variable 'connected' on a separate cache line

vsjulu:
@tchaikov  I am re-submitting the changes to isolate the 'connected' variable on a cache line. Since the last discussion, you had one remaining comment - you wanted to remove the 'pad' and instead align the next data item ( entitiy_inst_t peer) in the structure. We want to stick to the pad because we dont want to tie this to whatever be the next data item. The comment conveys the idea to whoever is going through the code. If we align the next data item and a new item is introduced again in between, it may introduce more data items into the same cache line..

### DIFF
--- a/src/msg/xio/XioConnection.h
+++ b/src/msg/xio/XioConnection.h
@@ -65,7 +65,8 @@ public:
 private:
   XioConnection::type xio_conn_type;
   XioPortal *portal;
-  atomic_t connected;
+  atomic_t connected __attribute__((align(64)));
+  char pad_after[56]; 	//Padding to isolate the frequently accessed atomic variable 'connected' on a separate cache line
   entity_inst_t peer;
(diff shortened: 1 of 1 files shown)


mattbenjamin:
Ok, is there a blocker on this conceptually?

vsjulu:
@mattbenjamin -  No, this is not blocking anything. It is a part of performance enhancement though.

mattbenjamin:
Kefu is correct, GCC and clang compilers will emit warnings regarding this with our current compile settings.  Can you fix?
-tx

athanatos:
It doesn't look like there's much movement on this one, so I'm going to close it for now

@vsjulu Feel free to reopen once you get a chance to address the comments!

PR #7891, opened 2016-03-02
rgw: multiple Swift API compliance improvements for TempURL

rzarzynski:


rzarzynski:
@yehudasa also s3-tests haven't found any regression.

PR #7893, opened 2016-03-03
mon: osdmonitor: trigger an immediate propose if any newly down osd i…

xiexingguo:
…s detected during tick()

### DIFF
--- a/src/mon/OSDMonitor.cc
+++ b/src/mon/OSDMonitor.cc
@@ -1625,18 +1625,21 @@ bool OSDMonitor::can_mark_in(int i)
   return true;
 }
 
-void OSDMonitor::check_failures(utime_t now)
+bool OSDMonitor::check_failures(utime_t now)
 {
+  bool propose_now = false;
   for (map<int,failure_info_t>::iterator p = failure_info.begin();
        p != failure_info.end();
        ++p) {
     if (can_mark_down(p->first)) {
-      check_failure(now, p->first, p->second);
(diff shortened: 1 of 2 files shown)


src/mon/OSDMonitor.cc
```diff
@@ -1646,6 +1649,8 @@ bool OSDMonitor::check_failure(utime_t now, int target_osd, failure_info_t& fi)
 
   utime_t grace = orig_grace;
   double my_grace = 0, peer_grace = 0;
+  if (propose_now)
+    *propose_now = false;
```
liewegas: hmm, I woudl drop this, and the weird condition when passing the arg or NULL above, and then only set the value to true

src/mon/OSDMonitor.cc
```diff
@@ -2591,7 +2600,8 @@ void OSDMonitor::tick()
   utime_t now = ceph_clock_now(g_ceph_context);
 
   // mark osds down?
-  check_failures(now);
+  if(check_failures(now))
```
liewegas: s/if(/if (/

src/mon/OSDMonitor.cc
liewegas: and hopefully this is the only caller and we can make the argument required instead of optional?

xiexingguo: @liewegas  Actually check_failure() will also be called by prepare_failure(), that is why I try to make this argument optional.

xiexingguo:
Others are fixed.

PR #7573, opened 2016-02-08
Wip wjw freebsd tests

wjwithagen:
If everything for Ceph compiles under FreeBSD
And src/test/run-cli-tests runs to succes
Start compiling the tests

PR #7897, opened 2016-03-03
common/obj_bencher.cc: make verify error fatal

branch-predictor:
When run without "--no-verify", all verification errors are noted,
but they are not forwarded/reported anywhere else but to cerr, which
will cause automated testing to ignore them. Make seq_read_bench and
rand_read_bench return -EIO on any verification error which will,
in turn, return it back to caller (which include test_rados_tool.sh).

Fixes: #14971

### DIFF
--- a/src/common/obj_bencher.cc
+++ b/src/common/obj_bencher.cc
@@ -827,7 +827,7 @@ int ObjBencher::seq_read_bench(int seconds_to_run, int num_objects, int concurre
 
   completions_done();
 
-  return 0;
+  return (errors > 0 ? -EIO : 0);
 
  ERR:
   lock.Lock();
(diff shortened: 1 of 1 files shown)


dachary:
<pre>
[  FAILED  ] Buffer.constructors (2 ms)
</pre>
is the only failure.

dachary:
@branch-predictor could you reference the tracker issue at df001c0 ?

wjwithagen:
@branch-predictor @dachary 
I had the same error bother me in a testrun yesterday evening in #7573 
See the last red mark of all commit. Exact the same eror for code I have not touched in a long time

branch-predictor:
@dachary done!

dachary:
@branch-predictor now that it's in master, it can be **git cherry-pick -x** for the infernalis/hammer backports. thanks !

PR #7901, opened 2016-03-03
libcephfs: fix python tests and fix getcwd on missing dir

jcsp:


gregsfortytwo:
@ukernel, please review

jcsp:
Updated to run tests as root.  Nothing rooty about the tests, other than that in teuthology the filesystem's root dir has uid 0, so so does everything that wants to write there.

jcsp:
Passing here (along with flock change): http://qa-proxy.ceph.com/teuthology/jspray-2016-03-15_08:58:50-fs:basic-wip-pycephfs-flock---basic-mira/62101/teuthology.log

PR #7906, opened 2016-03-03
journal: re-use common threads between journalers

dillaman:
The rbd-mirror daemon will instantiate potentially thousands of journalers.  To avoid spawning two threads per journaler, now rbd-mirror can reuse the same threads between journalers.

### DIFF
--- a/src/journal/FutureImpl.cc
+++ b/src/journal/FutureImpl.cc
@@ -3,14 +3,15 @@
 
 #include "journal/FutureImpl.h"
 #include "common/Finisher.h"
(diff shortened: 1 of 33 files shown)


jdurgin:
`TestJournalPlayer.PrefetchSkippedObject` failed in the make check bot

dillaman:
@jdurgin I was able to repeat that failure on master -- I'll fix it under PR #7884 if it's still an issue after the code changes made to JournalPlayer in that commit.

dillaman:
@jdurgin Unable to repeat the issue under PR #7884 branch

PR #7908, opened 2016-03-03
cmake: Cython modules building with cmake

alimaredia:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,6 +48,11 @@ include(CheckIncludeFiles)
 include(CheckIncludeFileCXX)
 include(CheckFunctionExists)
 
+#put all the libs and binaries in one place
+set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
+set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
+
 CHECK_FUNCTION_EXISTS(fallocate CEPH_HAVE_FALLOCATE)
 CHECK_FUNCTION_EXISTS(posix_fadvise HAVE_POSIX_FADVISE)
 CHECK_FUNCTION_EXISTS(posix_fallocate HAVE_POSIX_FALLOCATE)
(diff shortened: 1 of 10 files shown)


alimaredia:
@jdurgin @cbodley @jcsp care to take a quick look?

src/CMakeLists.txt
```diff
@@ -1287,6 +1289,7 @@ add_custom_target(vstart DEPENDS
 add_custom_target(cephfs_testing DEPENDS
     vstart
     rados
+    cython_modules
```
cbodley: I'm not sure that `cephfs_testing` is the right place to put the only dependency on cython bindings. do you guys have a better idea of which targets will actually depend on `cython_rbd` and `cython_cephfs`?

CMakeLists.txt
```diff
@@ -48,6 +48,11 @@ include(CheckIncludeFiles)
 include(CheckIncludeFileCXX)
 include(CheckFunctionExists)
 
+#put all the libs and binaries in one place
+set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
```
jcsp: Was this actually necessary?  I don't necessarily have a problem with introducing lib, bin folders, but it seems unrelated to this change.

Presumably vstart will need updating, as will https://github.com/ceph/ceph-qa-suite/blob/master/tasks/cephfs/vstart_runner.py

Might it make sense to do the lib/ bin/ thing in another patch later?

src/CMakeLists.txt
jcsp: Nothing we actually _build_ depends on them, right?  So it would only be testing-related targets that will run using the modules that would have the dependency, and then presumably packaging one day.

CMakeLists.txt
alimaredia: @jcsp I'm introducing the lib, bin directories thing in another PR i'll have out today, which might take a little longer to get merged. I wanted to keep things consistent when I rebase There are some updates to vstart.sh in that one as well.

jcsp: I guess the vstart updates need pulling in to this PR too, if otherwise vstart will still not work for cmake users

alimaredia: For the time being let's move this conversation to here: https://github.com/ceph/ceph/pull/7912

tchaikov:
@alimaredia is this change still relevant? as #8351 has been merged.

alimaredia:
@tchaikov nope, this PR can be deleted

PR #7909, opened 2016-03-03
ceph-detect-init: return correct value on recent SUSE distros

smithfarm:
http://tracker.ceph.com/issues/14770 Fixes: #14770


smithfarm:
@dachary Tested manually on Tumbleweed. Since the `release` value of Leap will be 42.1, it is obvious that it will work on Leap as well, though I have not tested it.

PR #7910, opened 2016-03-03
buffer: increment history alloc as well in raw_combined

athanatos:
Introduced in 73dcd26fc8798e09d93520c583d63edb28ffc631

Fixes: #14955


athanatos:
Probably don't need to run a suite against this, going to wait for the builds to finish though.

liewegas:
:+1:

PR #7912, opened 2016-03-03
Wip cmake make check

alimaredia:
As long as these commit's don't break make check that jenkins runs, and the numerous other changes are agreed upon, there's no reason this CMake work shouldn't go be in ASAP


alimaredia:
@cbodley scrutanize away!

src/vstart.sh
```diff
@@ -21,50 +21,34 @@ if [ -n "$VSTART_DEST" ]; then
...
+        [ -z "$INIT_CEPH" ] && INIT_CEPH=$CEPH_BIN/init-ceph
+        [ -z "$CEPH_LIB" ] && CEPH_LIB=$CEPH_BUILD_DIR/lib
+        [ -z "$OBJCLASS_PATH" ] && OBJCLASS_PATH=$CEPH_LIB
+        [ -z "$EC_PATH" ] && EC_PATH=$CEPH_LIB
+else
+        [ -z "$CEPH_BIN" ] && CEPH_BIN=.
+        [ -z "$CEPH_LIB" ] && CEPH_LIB=.libs
+        [ -z $EC_PATH ] && EC_PATH=$CEPH_LIB
+        [ -z $CS_PATH ] && CS_PATH=$CEPH_LIB
+        [ -z $OBJCLASS_PATH ] && OBJCLASS_PATH=$CEPH_LIB
```
alimaredia: @jcsp here are the changes made to vstart.sh. A good place to look at where some of these variables are defined is cmake/modules/AddCephTest.cmake

src/test/common/test_blkdev.cc
```diff
@@ -19,13 +20,15 @@ TEST(blkdev, get_block_device_base) {
   ASSERT_EQ(-EINVAL, get_block_device_base("/etc/notindev", buf, 100));
 
   for (int i=0; i<2; ++i) {
-    const char *root = "";
-    if (i == 0)
-      root = "test/common/test_blkdev_sys_block";
-    set_block_device_sandbox_dir(root);
+    string root;
+    if (i == 0) {
+      string CEPH_ROOT = getenv("CEPH_ROOT");
```
dillaman: Protect against ```getenv``` returning nullptr.

cmake/modules/AddCephTest.cmake
```diff
@@ -0,0 +1,24 @@
...
+  add_dependencies(check ${test_name})
+  set_property(TEST 
+    ${test_name}
+    PROPERTY ENVIRONMENT 
+    CEPH_ROOT=${CMAKE_SOURCE_DIR}
+    CEPH_BIN=${CMAKE_RUNTIME_OUTPUT_DIRECTORY}
+    CEPH_LIB=${CMAKE_LIBRARY_OUTPUT_DIRECTORY}
+    LD_LIBRARY_PATH=${CMAKE_BINARY_DIR}/lib
+    PATH=${CMAKE_RUNTIME_OUTPUT_DIRECTORY}:${CMAKE_SOURCE_DIR}/src:$ENV{PATH}
+    PYTHONPATH=${CMAKE_LIBRARY_OUTPUT_DIRECTORY}/cython_modules/lib.linux-x86_64-2.7:${CMAKE_SOURCE_DIR}/src/pybind:${CMAKE_SOURCE_DIR}/src/pybind/rados)
```
cbodley: > lib.linux-x86_64-2.7

this part will probably need to be different for x86, arm, freebsd, etc

src/vstart.sh
jcsp: No objection from me to any of the vstart stuff here (I'm assuming you've tried it and it all works) -- can you separate it into its own commit and bring it over to #7908?

Might also be worth checking that none of the other regular cmake users (@oritwas @tchaikov perhaps?) have any comments on the change to build everything in bin/ and lib/ before we merge it in #7908

src/common/buffer.cc
```diff
@@ -1085,7 +1085,7 @@ static simple_spinlock_t buffer_debug_lock = SIMPLE_SPINLOCK_INITIALIZER;
   template<bool is_const>
   bool buffer::list::iterator_impl<is_const>::operator!=(const buffer::list::iterator_impl<is_const>& rhs) const
   {
-    return bl == rhs.bl && off == rhs.off;
+    return bl != rhs.bl || off != rhs.off;
```
cbodley: this is completely unrelated, but appears to be a valid bug fix

src/compressor/snappy/CMakeLists.txt
```diff
@@ -1,7 +1,8 @@
 # snappy
 
+# absolute path because $snappy_sources used outside of this directory
 set(snappy_sources
-  CompressionPluginSnappy.cc
+  ${CMAKE_SOURCE_DIR}/src/compressor/snappy/CompressionPluginSnappy.cc
```
cbodley: can we get around the absolute path thing by making this an object library?

src/erasure-code/jerasure/CMakeLists.txt
```diff
@@ -31,7 +31,7 @@ add_library(ec_jerasure_generic SHARED
   $<TARGET_OBJECTS:erasure_code_objs>
   )
 add_dependencies(ec_jerasure_generic ${CMAKE_SOURCE_DIR}/src/ceph_ver.h)
-target_link_libraries(ec_jerasure_generic ${EXTRALIBS})
+target_link_libraries(ec_jerasure_generic crush ${EXTRALIBS})
```
cbodley: i don't understand why the erasure code libs would depend on crush. if you're getting linker errors without this, we should figure out what's pulling crush stuff in

src/pybind/cephfs/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+add_custom_target(cython_cephfs ALL
+  COMMAND
+  LDFLAGS=-L${CMAKE_LIBRARY_OUTPUT_DIRECTORY}
+  CYTHON_BUILD_DIR=${CMAKE_BINARY_DIR}/src/pybind/cephfs
+  CFLAGS=\"-I${CMAKE_SOURCE_DIR}/src -I${CMAKE_BINARY_DIR}/include -I${CMAKE_SOURCE_DIR}/src/include -std=c++11\"
+  python ${CMAKE_SOURCE_DIR}/src/pybind/cephfs/setup.py build --build-base ${CYTHON_MODULE_DIR} --verbose 
+  WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/src/pybind/cephfs
```
cbodley: you can use `${CMAKE_CURRENT_SOURCE_DIR}` here

src/test/ObjectMap/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+set_target_properties(ceph_test_object_map PROPERTIES COMPILE_FLAGS
+  ${UNITTEST_CXX_FLAGS})
+target_link_libraries(ceph_test_object_map
+  os
+  common
+  ${UNITTEST_LIBS}
+  global
+  ${EXTRALIBS}
+  ${CMAKE_DL_LIBS}
+  )
```
cbodley: can you use the `add_ceph_test()` for these targets? or are they not supposed to be part of `make check`?

alimaredia:
@dachary @tchaikov As long as make check passes in jenkins with automake, do you anticipate any of these changes breaking any parts of the Ceph CI processes (ex: teuthology)?

cmake/modules/AddCephTest.cmake
jdurgin: and dependent on python version

src/test/cls_rbd/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
+# cls_test_cls_rbd
+add_executable(ceph_test_cls_rbd
+  test_cls_rbd.cc
+  ${CMAKE_SOURCE_DIR}/src/common/TextTable.cc
+  ${CMAKE_SOURCE_DIR}/src/common/secret.c
```
cbodley: are these two not part of libcommon? we shouldn't have to recompile them here

src/test/cls_hello/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
...
+target_link_libraries(ceph_test_cls_hello
+  librados
+  global
+  ${EXTRALIBS}
+  ${BLKID_LIBRARIES}
+  ${CMAKE_DL_LIBS}
+  radostest
+  ${UNITTEST_LIBS}
+  )
+
```
cbodley: shouldn't these cls tests run with `make check`?

src/test/common/CMakeLists.txt
```diff
@@ -0,0 +1,190 @@
+# get_command_descriptions
+add_executable(get_command_descriptions
+  get_command_descriptions.cc
+  ${CMAKE_SOURCE_DIR}/src/common/TextTable.cc
```
cbodley: use libcommon or an object library instead of rebuilding TextTable.cc

src/test/ObjectMap/CMakeLists.txt
jdurgin: these aren't in `make check` in automake, but ceph_test_object_map could be added (the keyvaluedb tests use an actual object store and aren't as suitable)

src/test/compressor/CMakeLists.txt
```diff
@@ -0,0 +1,76 @@
+add_library(ceph_example SHARED
+  compressor_plugin_example.cc
+  ${CMAKE_SOURCE_DIR}/src/compressor/Compressor.cc
```
cbodley: object library for Compressor.cc

src/test/compressor/test_compression_plugin_snappy.cc
```diff
@@ -38,12 +38,17 @@ int main(int argc, char **argv) {
   global_init(NULL, args, CEPH_ENTITY_TYPE_CLIENT, CODE_ENVIRONMENT_UTILITY, 0);
   common_init_finish(g_ceph_context);
 
-  int r = system("mkdir -p .libs/compressor");
+  const char* env = getenv("CEPH_LIB");
+  std::string directory(env ? env : "lib");
+  string mkdir_compressor = "mkdir -p " + directory + "/compressor";
```
cbodley: i would prefer to fail with an error if we don't have a CEPH_LIB. an empty `directory` here can't possibly be correct, and will try to modify /

src/test/encoding/CMakeLists.txt
```diff
@@ -0,0 +1,4 @@
+# scripts
+add_ceph_test(check-generated.sh ${CMAKE_SOURCE_DIR}/src/test/encoding/check-generated.sh)
+add_ceph_test(readable.sh ${CMAKE_SOURCE_DIR}/src/test/encoding/readable.sh)
```
cbodley: can replace `${CMAKE_SOURCE_DIR}/src/test/encoding` with `${CMAKE_CURRENT_SOURCE_DIR}`

src/test/erasure-code/CMakeLists.txt
```diff
@@ -1,4 +1,6 @@
-# make check tests for erasure-code directory
+
+add_ceph_test(test-erasure-code.sh ${CMAKE_SOURCE_DIR}/src/test/erasure-code/test-erasure-code.sh)
+add_ceph_test(test-erasure-eio.sh ${CMAKE_SOURCE_DIR}/src/test/erasure-code/test-erasure-eio.sh)
```
cbodley: can replace `${CMAKE_SOURCE_DIR}/src/test/erasure-code` with `${CMAKE_CURRENT_SOURCE_DIR}`

src/test/mon/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+#scripts
+add_ceph_test(misc.sh ${CMAKE_SOURCE_DIR}/src/test/mon/misc.sh)
+add_ceph_test(mkfs.sh ${CMAKE_SOURCE_DIR}/src/test/mon/mkfs.sh)
+add_ceph_test(mon-created-time.sh ${CMAKE_SOURCE_DIR}/src/test/mon/mon-created-time.sh)
+add_ceph_test(mon-handle-forward.sh ${CMAKE_SOURCE_DIR}/src/test/mon/mon-handle-forward.sh)
+add_ceph_test(mon-ping.sh ${CMAKE_SOURCE_DIR}/src/test/mon/mon-ping.sh)
+add_ceph_test(mon-scrub.sh ${CMAKE_SOURCE_DIR}/src/test/mon/mon-scrub.sh)
+add_ceph_test(osd-crush.sh ${CMAKE_SOURCE_DIR}/src/test/mon/osd-crush.sh)
+add_ceph_test(osd-erasure-code-profile.sh ${CMAKE_SOURCE_DIR}/src/test/mon/osd-erasure-code-profile.sh)
+add_ceph_test(osd-pool-create.sh ${CMAKE_SOURCE_DIR}/src/test/mon/osd-pool-create.sh)
```
cbodley: can replace `${CMAKE_SOURCE_DIR}/src/test/mon` with `${CMAKE_CURRENT_SOURCE_DIR}`

src/test/osd/CMakeLists.txt
```diff
@@ -0,0 +1,91 @@
...
+add_ceph_test(osd-config.sh ${CMAKE_SOURCE_DIR}/src/test/osd/osd-config.sh)
+add_ceph_test(osd-markdown.sh ${CMAKE_SOURCE_DIR}/src/test/osd/osd-markdown.sh) 
+add_ceph_test(osd-reactivate.sh ${CMAKE_SOURCE_DIR}/src/test/osd/osd-reactivate.sh) 
+add_ceph_test(osd-reuse-id.sh ${CMAKE_SOURCE_DIR}/src/test/osd/osd-reuse-id.sh)
+#osd-scrub-repair hangs
+#add_ceph_test(osd-scrub-repair.sh ${CMAKE_SOURCE_DIR}/src/test/osd/osd-scrub-repair.sh) 
+add_ceph_test(osd-scrub-snaps.sh ${CMAKE_SOURCE_DIR}/src/test/osd/osd-scrub-snaps.sh)
+
+#osd-copy-from.sh needs to be run out of ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}
+add_test(NAME osd-copy-from.sh COMMAND bash ${CMAKE_SOURCE_DIR}/src/test/osd/osd-copy-from.sh WORKING_DIRECTORY ${CMAKE_RUNTIME_OUTPUT_DIRECTORY})
```
cbodley: can replace `${CMAKE_SOURCE_DIR}/src/test/osd` with `${CMAKE_CURRENT_SOURCE_DIR}`

src/test/pybind/CMakeLists.txt
```diff
@@ -0,0 +1,2 @@
+add_ceph_test(test_ceph_daemon.py ${CMAKE_SOURCE_DIR}/src/test/pybind/test_ceph_daemon.py)
+add_ceph_test(test_ceph_argparse.py ${CMAKE_SOURCE_DIR}/src/test/pybind/test_ceph_argparse.py)
```
cbodley: `${CMAKE_CURRENT_SOURCE_DIR}`

src/test/cls_hello/CMakeLists.txt
jdurgin: I'd leave anything out of make check if it's not there in automake - originally we tried to keep `make check` only running unit tests, later added more functionality when it started getting run automatically on PRs, and then had to start taking out some less useful tests since it became too long.

So there are plenty of ceph_test_* programs that shouldn't be in `make check`, while all of unittest_* should be included

dachary:
> As long as make check passes in jenkins with automake, do you anticipate any of these changes breaking any parts of the Ceph CI processes (ex: teuthology)?

The only thing that could break teuthology is if you change files in the qa directory or in what is found in the ceph packages.

src/vstart.sh
oritwas: I don't have any problem with building everything in bin/ and lib/

cbodley:
@alimaredia there are some linker errors with the test_librgw_file* targets with undefined references to heap profiler stuff

they only try to build under `make`, not `make check`, so they need to be updated to use `add_ceph_unittest()`

cbodley:
> The only thing that could break teuthology is if you change files in the qa directory or in what is found in the ceph packages.

@dachary qa/workunits/ceph-helpers.sh was changed to support out-of-tree builds (relative paths to binaries were replaced with environment variables). It was pointing to stuff in .libs (which wouldn't come from installed ceph packages), so I would assume that it isn't used by teuthology?

cbodley:
the diff for src/test/CMakeLists.txt is too big for github, so comments here..

464: `add_ceph_test(test-ceph-helpers.sh ${CMAKE_SOURCE_DIR}/src/test/test-ceph-helpers.sh)`
use `${CMAKE_CURRENT_SOURCE_DIR}`

478: `add_dependencies(check run-tox)` should be run-tox-ceph-disk?

481: `add_dependencies(check run-tox)` should be run-tox-ceph-detect-init?

483: `set_property(TEST run-tox-ceph-disk ...`
use `add_ceph_test()` if it works, or update `PYTHONPATH=` for cython path

dachary:
> qa/workunits/ceph-helpers.sh was changed to support out-of-tree builds (relative paths to binaries were replaced with environment variables). It was pointing to stuff in .libs (which wouldn't come from installed ceph packages), so I would assume that it isn't used by teuthology?

It is used by teuthology, reason why it is in qa/workunits:

workunits/cephtool/test.sh:3:source $(dirname $0)/../ceph-helpers.sh
workunits/rbd/journal.sh:3:. $(dirname $0)/../ceph-helpers.sh
workunits/rbd/rbd-nbd.sh:3:. $(dirname $0)/../ceph-helpers.sh
workunits/rbd/test_admin_socket.sh:7:. $(dirname $0)/../ceph-helpers.sh

cmake/modules/AddCephTest.cmake
alimaredia: Great points, ill come up with a fix

src/test/ObjectMap/CMakeLists.txt
alimaredia: @jdurgin should I go ahead and change the prefix from ceph_test_* to unittest_* and add it to make check?

jcsp:
@alimaredia is this one still a ways off?  is it worth me grabbing the vstart parts out of this and adding to #7908?

cbodley:
@alimaredia have we seen this branch run through teuthology yet? if so, i'd like to get it merged as is. we can fix the cython 'lib.linux-x86_64-2.7' thing as a separate pr when you get to it

dmick:
retest this please

src/test/librados/CMakeLists.txt
```diff
@@ -0,0 +1,172 @@
...
+target_link_libraries(ceph_test_rados_api_aio
+  librados ${UNITTEST_LIBS} radostest)
+
+# ceph_test_rados_api_list
+add_executable(ceph_test_rados_api_list
+  list.cc
+  )
+set_target_properties(ceph_test_rados_api_list PROPERTIES COMPILE_FLAGS
+  ${UNITTEST_CXX_FLAGS})
+target_link_libraries(ceph_test_rados_api_list
```
tchaikov: need to link against "global", see 769c0affc4408061fa4787c7e910e4f48b9f3bf8

tchaikov:
@alimaredia i am trying to get your branch to pass the build at http://gitbuilder.sepia.ceph.com/gitbuilder-ceph-tarball-trusty-amd64-cmake/#origin/wip-cmake-make-check, so i added my fix on top of yours, please feel free to squash it into your corresponding commit(s) (83ff75091f6f795faf23582968813d2c67b5540b).

src/test/librados_test_stub/TestClassHandler.cc
```diff
@@ -43,7 +45,9 @@ void TestClassHandler::open_class(const std::string& name,
 void TestClassHandler::open_all_classes() {
   assert(m_class_handles.empty());
 
-  DIR *dir = ::opendir(".libs");
+  const char* env = getenv("CEPH_LIB");
+  std::string CEPH_LIB(env ? env : "lib");
+  DIR *dir = ::opendir(CEPH_LIB.c_str());
   if (dir == NULL) {
```
trociny: @alimaredia @cbodley  Now if CEPH_LIB env is not set, "lib" is used instead of previously used ".libs". Was it intentionally?
```
% ./unittest_rbd_mirror               
2016-04-18 09:10:23.392541 7fb9f8fdeb00 -1 WARNING: the following dangerous and experimental features are enabled: *
[==========] Running 58 tests from 11 test cases.
[----------] 1 test from TestMockImageReplayer
[ RUN      ] TestMockImageReplayer.Blah
test/librados_test_stub/TestClassHandler.cc: In function 'void librados::TestClassHandler::open_all_classes()' thread 7fb9f8fdeb00 time 2016-04-18 09:10:23.500613
test/librados_test_stub/TestClassHandler.cc: 52: FAILED assert(false)
```

alimaredia: @trociny This was done intentionally. ".libs" is an autotools convention and since we are moving toward using CMake. The "lib" was chosen because all the libraries in a CMake build go to ${CMAKE_BUILD_DIR}/lib.

I would love it if you transitioned to CMake and CTest (the environment variables are predefined there all you have to do is run `ctest -V -R unittest_rbd_mirror` but I understand your concern. 

One thing I can do is add a check for if opendir(".libs") is true i can set *dir to that to make sure this test and others with the similar revisions work without running make check in the automake build. How does that sound?

trociny: @alimaredia  If it is intentional, I am fine with it. Not sure it is necessary to do any additional checks for compatibility, it might only increase the transition period. A more useful thing I think is to announce the new way to developers. It may already be done and I just missed it? At least a message in ceph-dev would be nice.

dzafman: @trociny @alimaredia I added the automatic setting of CEPH_LIB using .libs in my pull request #8649 just for the ceph_objectstore_tool.py test.

alimaredia: @trociny I've made the appropriate adjustments in https://github.com/ceph/ceph/pull/8770. I did end up changing "lib" to ".libs" since that is a good default that CEPH_LIB should revert back to for the time being

PR #7915, opened 2016-03-03
librados: stat2 with higher time precision

yehudasa:
extend librados, objclass api to support higher resolution stat (and mtime for librados) operations.

### DIFF
--- a/src/include/rados/librados.hpp
+++ b/src/include/rados/librados.hpp
@@ -51,8 +51,6 @@ namespace librados
   typedef void *completion_t;
   typedef void (*callback_t)(completion_t cb, void *arg);
 
-  typedef void *ceph_real_time_t;
-
   class CEPH_RADOS_API ListObject
   {
   public:
@@ -364,7 +362,6 @@ namespace librados
 
     void mtime(time_t *pt);
     void mtime2(struct timespec *pts);
-    void mtime2(ceph_real_time_t *pmtime);
 
     void create(bool exclusive);
     void create(bool exclusive,
@@ -477,7 +474,6 @@ namespace librados
 
     void stat(uint64_t *psize, time_t *pmtime, int *prval);
     void stat2(uint64_t *psize, struct timespec *pts, int *prval);
-    void stat2(uint64_t *psize, ceph_real_time_t *pmtime, int *prval);
     void getxattr(const char *name, bufferlist *pbl, int *prval);
     void getxattrs(std::map<std::string, bufferlist> *pattrs, int *prval);
     void read(size_t off, uint64_t len, bufferlist *pbl, int *prval);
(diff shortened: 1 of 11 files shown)


src/include/rados/librados.hpp
```diff
@@ -10,6 +10,7 @@
 #include <utility>
 #include "memory.h"
 #include "buffer.h"
+#include "common/ceph_time.h"
```
jdurgin: is this suitable for a public interface? seems like we might want to avoid exposing it

yehudasa: not less suitable than buffer.h imo

src/include/rados/librados.hpp
```diff
@@ -355,13 +356,21 @@ namespace librados
   class CEPH_RADOS_API ObjectWriteOperation : public ObjectOperation
   {
   protected:
-    time_t *pmtime;
+    ceph::real_time rt;
```
jdurgin: changing members breaks the ABI - should add it to ObjectOperationImpl instead, move the implementation of mtime and mtime2 to librados.cc, and make sure the old version of mtime() is still ok to call

src/include/rados/librados.hpp
jdurgin: since we'll want a c version anyway, maybe it would be better to use the same type in the c++ api to avoid exposing ceph_time

jdurgin: buffer.h was a mistake - it should have hidden a lot more so we would be able to change it more easily without breaking librados users

liewegas: we can use struct timespec from time.h... there are to/from methods in the ceph_time classes.

jdurgin: struct timespec or two uint64_t sound good to me

yehudasa:
@jdurgin @liewegas pushed a second version

liewegas:
This looks okay to me, though I'm a little iffy on the ceph_real_time_t typedef.  @jdurgin ?

jdurgin:
I'd rather not add the ceph_real_time_t version. I don't think it's worth extra messiness in the librados api to avoid an extra field in a wrapper callback the few places we use mtime internally.

Other than that looks good. Thanks for adding tests and objclass support too.

yehudasa:
@liewegas @jdurgin as was discussed on irc before implementing the ceph_real_time_t, not having it would mean that it'd complicate implementation of different operations quite a bit. Specifically when dealing with async operations, it's not always trivial to just have an extra conversion at the endpoint, since it'd require implementing unnecessary callbacks. This is silly, because we use real_time all around, so we'd just need to do extra work all the time. Either we drop the use of ceph::real_time altogether and switch to timespec all around, or we expose it somehow.

jdurgin:
Yes, it's a bit of extra work to use internally. Creating a callback that converts it, and using it everywhere we use mtime internally (not that many places) doesn't seem like a lot of work. There's already a single main place stat is used in rgw, and it's not hard to make a wrapper that always does the conversion.

I don't think it's worth exposing through the librados api, where it's of no use to anyone outside of ceph.git, to avoid that.

yehudasa:
@jdurgin but it is useful to everyone who is inside ceph.git. And since inside ceph.git we're using ceph::real_time, we should have this interface. It might not affect rbd much, but mtime for rgw is all over the place.

jdurgin:
rgw is only calling stat as an async op in a single place afaict.

yehudasa:
@jdurgin @liewegas now without ceph_real_time_t

jdurgin:
lgtm

liewegas:
I'll get it on my next wip-sage-testing pass

mattbenjamin:
do y'all need this rebased?

liewegas:
nope, should be ok as is.

PR #7919, opened 2016-03-04
FileJournal：delete FULL_WAIT state

jack-changtao:
For file joural , I think  FULL_WAIT of file journal state is useless

### DIFF
--- a/src/os/filestore/FileJournal.cc
+++ b/src/os/filestore/FileJournal.cc
@@ -1206,20 +1206,8 @@ void FileJournal::do_write(bufferlist& bl)
     Mutex::Locker locker(finisher_lock);
     journaled_seq = writing_seq;
 
-    // kick finisher?
-    //  only if we haven't filled up recently!
-    if (full_state != FULL_NOTFULL) {
-      dout(10) << "do_write NOT queueing finisher seq " << journaled_seq
-	       << ", full_commit_seq|full_restart_seq" << dendl;
-    } else {
-      if (plug_journal_completions) {
-	dout(20) << "do_write NOT queueing finishers through seq " << journaled_seq
-		 << " due to completion plug" << dendl;
-      } else {
-	dout(20) << "do_write queueing finishers through seq " << journaled_seq << dendl;
-	queue_completions_thru(journaled_seq);
-      }
-    }
+    dout(20) << "do_write queueing finishers through seq " << journaled_seq << dendl;
+    queue_completions_thru(journaled_seq);
   }
 }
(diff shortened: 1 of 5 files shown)


branch-predictor:
Can you explain why it is useless?

jack-changtao:
@branch-predictor 
I am not sure about this, because I am not sure of  FULL_WAIT state is usefull ?

branch-predictor:
This code was added to handle the case when journal was filled up and no space is left for new journal entries - in that case, entries will not be written to disk but queued and written once some space will be reclaimed. Once you remove this code, journaling will be no longer reliable in that case.
Also, you removed a call to journal->commit_start() in JournalingObejctstore which will prevent committing changes to underlying journal if there's any.

jack-changtao:
@branch-predictor 

Thanks for you reply! 

I still has some confused!   Every time before journal write, it calls check_for_full to check journal space. If it has no journal space, the journal entry is still on writeq and will be written once some space will be reclaimed.

why " journaling will be no longer reliable" ?

branch-predictor:
@jack-changtao FULL_WAIT causes journaling code to keep flushing entries while still queuing new entries (i. e. remains in state similar to FULL_FULL, except it actually isn't full anymore). See the description of FULL_WAIT in original comment:
```
  /*
   * full states cycle at the beginnging of each commit epoch, when commit_start()
   * is called.
   *   FULL - we just filled up during this epoch.
   *   WAIT - we filled up last epoch; now we have to wait until everything during
   *          that epoch commits to the fs before we can start writing over it.
   *   NOTFULL - all good, journal away.
   */
```

jack-changtao:
@branch-predictor 

sorry， I still can‘t got it. If it is remote, what is   "no longer reliable" it will happen? can you gave a use case ? Thanks very much !

branch-predictor:
@jack-changtao I gave an use case above (ensuring that once journal gets filled up, it will be emptied entirely before accepting new entry).

jack-changtao:
@branch-predictor 

Once journal gets filled up , the journal commit will wait untill there is space.
The journal should be trimmed all if it all has been applied!

jack-changtao:
@liewegas  @yuyuyu101  @XinzeChi   

Do you have time to review this ?  or point out the state FULL_WAIT use ?

Thanks very much !

athanatos:
It's necessary for some annoying parallel journaling edge case iirc.  If the journal fills up and the filesystem actually write ahead of the current journal state, we need to sync twice for some reason.

PR #7920, opened 2016-03-04
infernalis: PK11_DestroyContext() is called twice if PK11_DigestFinal() fails

badone:
http://tracker.ceph.com/issues/14960
This patch set resolves multiple crashes in infernalis including tracker 13826 and 13527


dachary:

PR #7921, opened 2016-03-04
Modifications to rdb-openstack.rst

RustShen:
the default value of 'enabled_backends' is 'lvm' in cinder.conf.

the documentation about cinder configuration doesn't cover changing it to ceph-ssd.

I think modifications about it should be made.

otherwise OpenStack will still use lvm as storage backend.


dillaman:
Can this be closed as a duplicate of #7923?

PR #7922, opened 2016-03-04
hammer: PK11_DestroyContext() is called twice if PK11_DigestFinal() fails

badone:
http://tracker.ceph.com/issues/14960
This patch set resolves multiple crashes in hammer including tracker 13826 and 13527


dachary:
@liewegas does this backport look good to merge ? It passed the hammer upgrade suite ( http://tracker.ceph.com/issues/14692#note-52 ) and the rados suite http://tracker.ceph.com/issues/14692#note-51).

PR #7923, opened 2016-03-04
Modifications to rbd-openstack.rst

RustShen:
Original configuration is not working 

'enabled_backends' should be changed according to openstack configuration guide.

''A list of backend names to use. These backend names should
''be backed by a unique [CONFIG] group with its options (list
''value)
''enabled_backends=<None>
 

Signed-off-by: RustShen <rustinpeace@163.com>

### DIFF
--- a/doc/rbd/rbd-openstack.rst
+++ b/doc/rbd/rbd-openstack.rst
@@ -255,6 +255,11 @@ OpenStack requires a driver to interact with Ceph block devices. You must also
 specify the pool name for the block device. On your OpenStack node, edit
 ``/etc/cinder/cinder.conf`` by adding::
 
+    [DEFAULT]
+    ...
+    enabled_backends = ceph
+    ...
+    [ceph]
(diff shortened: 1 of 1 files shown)


RustShen:
Hi, I'm new to this.
Can this commit be merged, or should I do some modifications about it before merging.There seems to be a checking failure in default.

The duplication is closed.I changed the wrong part of docs without knowing, sorry for that.

jdurgin:
could you add the openstack version where this started being required as well (maybe as a comment in before the config settings) - also squash that last commit into the first one (git commit rebase -i HEAD^, say squash for the 2nd commit and exit the editor). thanks!

RustShen:
Rebasing is done , thx for help.
About the openstack version, I checked that this configuration is in Juno,Kilo and Liberty, but I'm not sure when it starts.Since the previous versions have been end of life, this configuration should be a default setting for openstack to use rbd.

jdurgin:
Thanks for checking!

tobegit3hub:
Nice job.

PR #7925, opened 2016-03-04
infernalis: Objects promoted unnecessarily when deleting from cache

rajatg1989:
http://tracker.ceph.com/issues/14964


dachary:

PR #7927, opened 2016-03-04
rgw: RGWZoneParams::create should not handle -EEXIST error

oritwas:
It should be handled by the caller


liewegas:
passed rados run

PR #7928, opened 2016-03-04
tools/rados: reduce "rados put" memory usage by op_size

branch-predictor:
indata.append(char*, len) causes buffer "buf" to be physically copied
to new memory block. Make static buffer pointing to buf so we don't need
any intermediate copies. This reduces memory usage by op_size, and also
cpu usage by a bit (no memmove necessary).


tchaikov:
lgtm.

PR #7930, opened 2016-03-04
debian: package librgw_file* tests

ktdreyer:
The `make install` command installs these files, but the Debian packaging did not include them.


ktdreyer:
The equivalent RPM packaging fix was cb9cfb7291c33607c213f0b56c9786316b3961b8 . cc'ing @mattbenjamin in case he wants to review

cbodley:
I just commented on https://github.com/ceph/ceph/pull/7912 about these targets - I think they should be run by `make check`. Is there a reason we should be installing them?

ktdreyer:
The files are already in the RPM, so we should either remove the utilities from the list of things that get installed during `make install`, or we should put them into the RPM packaging.

PR #7931, opened 2016-03-04
debian: tighten sub-package version dependencies

ktdreyer:
Prior to this change, users could possibly install some ceph sub-packages with differing versions. Make each sub-package explicitly depend on a particular version of its dependencies.

I've purposly excluded ceph-test's dependencies for now (see http://tracker.ceph.com/issues/10989 for background.)


liewegas:
passed rados suite

liewegas:
merged manually

PR #7933, opened 2016-03-04
doc: add ceph-detect-init(8) source to dist tarball

ktdreyer:
Prior to this commit, the tarball from `make dist` did not include the ceph-detect-init(8) man page rST source.


dachary:
<pre>
</pre>

PR #7934, opened 2016-03-04
etc/sysconfig/ceph: set 128MB tcmalloc cache size

ktdreyer:
Bump up from tcmalloc's default cache size.


ktdreyer:
CC @markhpc

markhpc:
This looks fine to me so long as we are willing to accept the higher memory usage by default.  FWIW, if async messenger ever becomes default we may be able to drop back down to 32MB.

PR #7942, opened 2016-03-05
mon: OSDMonitor: improve pool creation logic

xiexingguo:
This pr consists of two changes:

    The first one is that we move sanity-check for fast_read option of erasure coding
    pool ahead, so we don't dirty pending_inc map and thus can make a clean exit 
    on failure.

    The second one is that we move sanity-check for duplicated request of pool-create
    ahead,  this is safe because we only update new_pool_names[] field of penging_inc
    map at the end of creation, which undoubtedly reflects that the creation is a success.

    The first one shall be considered as a bug-fix while the second one is a cleanup, 
    which is proposed for the sake of efficiency.

### DIFF
--- a/qa/workunits/cephtool/test.sh
+++ b/qa/workunits/cephtool/test.sh
@@ -1374,6 +1374,7 @@ function test_mon_pg()
 
   ceph osd primary-affinity osd.0 .9
   expect_false ceph osd primary-affinity osd.0 -2
+  expect_false ceph osd primary-affinity osd.9999 .5
   ceph osd primary-affinity osd.0 1
 
   ceph osd pg-temp 0.0 0 1 2
--- a/qa/workunits/cephtool/test.sh
+++ b/qa/workunits/cephtool/test.sh
@@ -1235,6 +1235,7 @@ function test_mon_osd_pool()
   ceph osd pool mksnap data datasnap
   rados -p data lssnap | grep datasnap
   ceph osd pool rmsnap data datasnap
+  expect_false ceph osd pool rmsnap pool_fake snapshot
   ceph osd pool delete data data --yes-i-really-really-mean-it
 
   ceph osd pool create data2 10
--- a/qa/workunits/cephtool/test.sh
+++ b/qa/workunits/cephtool/test.sh
@@ -1041,6 +1041,9 @@ function test_mon_osd()
   expect_false "ceph osd blacklist $bl/-1"
   expect_false "ceph osd blacklist $bl/foo"
 
+  # test with wrong address
+  expect_false "ceph osd blacklist 1234.56.78.90/100"
+
   # Test `clear`
   ceph osd blacklist add $bl
   ceph osd blacklist ls | grep $bl
--- a/src/mon/OSDMonitor.cc
+++ b/src/mon/OSDMonitor.cc
@@ -1694,15 +1694,17 @@ bool OSDMonitor::can_mark_in(int i)
   return true;
 }
 
-void OSDMonitor::check_failures(utime_t now)
+bool OSDMonitor::check_failures(utime_t now)
 {
+  bool found_failure = false;
   for (map<int,failure_info_t>::iterator p = failure_info.begin();
        p != failure_info.end();
        ++p) {
     if (can_mark_down(p->first)) {
-      check_failure(now, p->first, p->second);
+      found_failure |= check_failure(now, p->first, p->second);
     }
   }
+  return found_failure;
 }
 
 bool OSDMonitor::check_failure(utime_t now, int target_osd, failure_info_t& fi)
@@ -2666,7 +2668,8 @@ void OSDMonitor::tick()
   utime_t now = ceph_clock_now(g_ceph_context);
 
   // mark osds down?
-  check_failures(now);
+  if (check_failures(now))
+    do_propose = true;
 
   // mark down osds out?
 
--- a/src/mon/OSDMonitor.cc
+++ b/src/mon/OSDMonitor.cc
@@ -2414,6 +2414,12 @@ bool OSDMonitor::prepare_remove_snaps(MonOpRequestRef op)
   for (map<int, vector<snapid_t> >::iterator p = m->snaps.begin();
        p != m->snaps.end();
        ++p) {
+
+    if (!osdmap.have_pg_pool(p->first)) {
+      dout(10) << " ignoring removed_snaps " << p->second << " on non-existent pool " << p->first << dendl;
+      continue;
+    }
+
     pg_pool_t& pi = osdmap.pools[p->first];
     for (vector<snapid_t>::iterator q = p->second.begin();
 	 q != p->second.end();
--- a/src/mon/OSDMonitor.cc
+++ b/src/mon/OSDMonitor.cc
@@ -5702,6 +5702,7 @@ bool OSDMonitor::prepare_command_impl(MonOpRequestRef op,
     if (!newcrush.name_exists(name)) {
       err = -ENOENT;
       ss << "item " << name << " does not exist";
+      goto reply;
     } else {
       int id = newcrush.get_item_id(name);
       if (!newcrush.check_item_loc(g_ceph_context, id, loc, (int *)NULL)) {
@@ -5714,6 +5715,7 @@ bool OSDMonitor::prepare_command_impl(MonOpRequestRef op,
 	} else {
 	  ss << "cannot link item id " << id << " name '" << name
              << "' to location " << loc;
+          goto reply;
 	}
       } else {
 	ss << "no need to move item id " << id << " name '" << name
--- a/src/mon/OSDMonitor.cc
+++ b/src/mon/OSDMonitor.cc
@@ -4757,6 +4757,24 @@ int OSDMonitor::prepare_new_pool(string& name, uint64_t auid,
     dout(10) << " prepare_pool_stripe_width returns " << r << dendl;
     return r;
   }
+  
+  bool fread = false;
+  if (pool_type == pg_pool_t::TYPE_ERASURE) {
+    switch (fast_read) {
+      case FAST_READ_OFF:
+        fread = false;
+        break;
+      case FAST_READ_ON:
+        fread = true;
+        break;
+      case FAST_READ_DEFAULT:
+        fread = g_conf->mon_osd_pool_ec_fast_read;
+        break;
+      default:
+        *ss << "invalid fast_read setting: " << fast_read;
+        return -EINVAL;
+    }
+  }
 
   for (map<int64_t,string>::iterator p = pending_inc.new_pool_names.begin();
        p != pending_inc.new_pool_names.end();
@@ -4771,6 +4789,7 @@ int OSDMonitor::prepare_new_pool(string& name, uint64_t auid,
   pg_pool_t empty;
   pg_pool_t *pi = pending_inc.get_new_pool(pool, &empty);
   pi->type = pool_type;
+  pi->fast_read = fread; 
   pi->flags = g_conf->osd_pool_default_flags;
   if (g_conf->osd_pool_default_flag_hashpspool)
     pi->set_flag(pg_pool_t::FLAG_HASHPSPOOL);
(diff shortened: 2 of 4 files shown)


xiexingguo:
change log:
1、add sanity check for fast-read setting during pool set
2、make debug_fake_ec_pool option cancelable
3、some cleanups

xiexingguo:
@tchaikov Done. Thank you for your attention. Feel free to change the assignee to yourself it you'd like to review these changes:-)

xiaoxichen:
could you stash the commits? something like "fix typo" and "remove redundant semicolon“ doesnt make sense to stay in the commit log.

xiexingguo:
I don't think so, I'd like them to stay this way for better tracing or backporting purpose or anything alike(e.g.,  for me,  one benefit of separating different changes into small, isolated commits would be the convenience to drop the incorrect changes and keep those which remains valid). BTW, I don't commit all these changes in a batch, but found them in a couple of days. That is why I keep appending new commits into this pr.

xiexingguo:
@tchaikov Could you do me a favor of reviewing this one for me too? This pr is somewhat big but most of the changes are cleanups. And I'll be very appreciated if I can receive some precious advice from you. Thanks again:-)

xiexingguo:
@tchaikov Ping~

src/mon/OSDMonitor.cc
```diff
@@ -1625,18 +1625,21 @@ bool OSDMonitor::can_mark_in(int i)
...
+  bool propose_now = false;
   for (map<int,failure_info_t>::iterator p = failure_info.begin();
        p != failure_info.end();
        ++p) {
     if (can_mark_down(p->first)) {
-      check_failure(now, p->first, p->second);
+      check_failure(now, p->first, p->second, propose_now ? NULL : &propose_now);
     }
   }
+  return propose_now;
```
xiaoxichen: bool is_failure_found = false;
 if (can_mark_down(p->first)) {
      is_failure_found |= check_failure(now, p->first, p->second);		
      }		      
}
return is_failure_found;

Seems did same thing as your code but simpler?

src/mon/OSDMonitor.cc
```diff
@@ -4599,6 +4599,14 @@ int OSDMonitor::prepare_new_pool(string& name, uint64_t auid,
...
   if (name.length() == 0)
     return -EINVAL;
+
+  for (map<int64_t,string>::iterator p = pending_inc.new_pool_names.begin();
+       p != pending_inc.new_pool_names.end();
+       ++p) {
+    if (p->second == name)
+      return 0;
+  }
+
```
tchaikov: i am not sure i see the point of moving this block around?

xiexingguo: So if we receive duplicated request of pool-creation request, we can fast catch it、skip over unnecessary sanity check and make quick response.

src/mon/OSDMonitor.cc
```diff
@@ -5596,6 +5596,7 @@ bool OSDMonitor::prepare_command_impl(MonOpRequestRef op,
     if (!newcrush.name_exists(name)) {
       err = -ENOENT;
       ss << "item " << name << " does not exist";
+      goto reply;
```
tchaikov: the commit message is misleading. `wait_for_finished_proposal()` does not trigger a new proposal.

src/mon/OSDMonitor.cc
```diff
@@ -6440,6 +6440,10 @@ bool OSDMonitor::prepare_command_impl(MonOpRequestRef op,
       wait_for_finished_proposal(op, new Monitor::C_Command(mon, op, 0, rs,
                                                 get_last_committed() + 1));
       return true;
+    } else {
+      ss << "osd." << id << " does not exist";
+      err = -ENOENT;
+      goto reply;
```
tchaikov: could you add a simple test for this change?

src/mon/OSDMonitor.cc
xiexingguo: check_failure() may also return true in this case: https://github.com/ceph/ceph/pull/7942/files#diff-0a5db46a44ae9900e226289a810f10e8R1705

In the above case we may don't want to trigger a duplicated propose, I guess.

src/mon/OSDMonitor.cc
xiexingguo: sure

src/mon/OSDMonitor.cc
xiexingguo: OK. Will fix it.

src/mon/OSDMonitor.cc
```diff
@@ -7657,7 +7657,7 @@ bool OSDMonitor::prepare_pool_op(MonOpRequestRef op)
     case POOL_OP_CREATE_SNAP:
     case POOL_OP_DELETE_SNAP:
       if (!pool->is_unmanaged_snaps_mode()) {
-        bool snap_exists = pool->snap_exists(m->name.c_str());
+        bool snap_exists = pool->snap_exists(m->name.c_str()) > 0;
```
tchaikov: i don't see the problem of the conversion to bool. it's well define in C++ standard 4.12 [conv.bool].

xiexingguo: OK, will remove this change.

src/mon/OSDMonitor.cc
```diff
@@ -2345,6 +2345,12 @@ bool OSDMonitor::prepare_remove_snaps(MonOpRequestRef op)
   for (map<int, vector<snapid_t> >::iterator p = m->snaps.begin();
        p != m->snaps.end();
        ++p) {
+
+    if (!osdmap.have_pg_pool(p->first)) {
+      dout(10) << " ignoring removed_snaps " << p->second << " on non-existent pool " << p->first << dendl;
```
tchaikov: could you add a simple test for this change?

xiexingguo: I'll try...

src/mon/OSDMonitor.cc
tchaikov: and could you elaborate a little bit on how this can improve the current availability of cluster?

xiexingguo:
@tchaikov  Would you kindly tell me where is the proper place to add these tests? Thanks a lot:)

tchaikov:
and i am with @xiaoxichen on squashing these cosmetic changes:

- 03ba3ad016b19c57756bb87688685ecb966f2ec5
- ee369d546cffea3a7e701a0320a364fc933b0f4f
- 001b39b3f18aa14f427feaf8e89ad70e1160f4d3

xiexingguo:
@tchaikov It's OK if you persist.

> Would you kindly tell me where is the proper place to add these tests? Thanks a lot:)

Any suggestion about my question? Thanks.

tchaikov:
> Would you kindly tell me where is the proper place to add these tests?
```
$ git grep 'osd primary-affinity'
doc/man/8/ceph.rst:Subcommand ``primary-affinity`` adjust osd primary-affinity from 0.0 <=<weight>
doc/man/8/ceph.rst:     ceph osd primary-affinity <osdname (id|osd.id)> <float[0.0-1.0]>
doc/rados/operations/crush-map.rst:     ceph osd primary-affinity <osd-id> <weight>
qa/workunits/cephtool/test.sh:  ceph osd primary-affinity osd.0 .9
qa/workunits/cephtool/test.sh:  expect_false ceph osd primary-affinity osd.0 -2
qa/workunits/cephtool/test.sh:  ceph osd primary-affinity osd.0 1
src/mon/MonCommands.h:COMMAND("osd primary-affinity " \
src/mon/MonCommands.h:  "adjust osd primary-affinity from 0.0 <= <weight> <= 1.0", \
src/mon/OSDMonitor.cc:  } else if (prefix == "osd primary-affinity") {
```

so probably you could put the one for "ceph osd primary-affinity" into "qa/workunits/cephtool/test.sh" ?

xiexingguo:
@tchaikov OK, thank you so very much. I'll give another try.

src/mon/OSDMonitor.cc
tchaikov: the same applies to the "skipped" checks. imho, we might want to focus on the performance of hotspots instead of the error handling paths.

xiexingguo: OK, I will take this change down.

xiexingguo:
@tchaikov All fixed as you suggested except the "osd pool rmsnap" one. I am worried the test case added can not fully verified such a complicated case. So perhaps I shall drop that change for now if you insist. And ping me if there is anything I need to further clarify.

xiexingguo:
@tchaikov Since you are there, could you please take a look at this again? Thanks.

xiexingguo:
@tchaikov Ping

tchaikov:
@xiexingguo sorry for the latency, will review this changeset tomorrow.

xiexingguo:
You are welcome:-)

qa/workunits/cephtool/test.sh
```diff
@@ -1355,6 +1355,7 @@ function test_mon_pg()
 
   ceph osd primary-affinity osd.0 .9
   expect_false ceph osd primary-affinity osd.0 -2
+  expect_false ceph osd primary-affinity osd.9999 .5
```
tchaikov: the sha1 of the referenced commit will change after the merge, maybe you can squash this change into the one it is testing.

qa/workunits/cephtool/test.sh
```diff
@@ -1213,6 +1213,7 @@ function test_mon_osd_pool()
   ceph osd pool mksnap data datasnap
   rados -p data lssnap | grep datasnap
   ceph osd pool rmsnap data datasnap
+  expect_false ceph osd pool rmsnap pool_fake snapshot
```
tchaikov: could you squash this change into the one it is testing?

xiexingguo:
Morning, done.

src/mon/OSDMonitor.cc
xiaoxichen: i think usually we can finished the proposal within one tick(5s by default IIRC), if not, it would be OK(or better) to trigger the proposal again. Make sense?

xiexingguo: @xiaoxichen Thanks for your attention. Your idea is OK with me. Let's wait and see @tchaikov 's idea before going any futher:)

tchaikov: i agree with @xiaoxichen , the propose_timer will be cancelled if the `tick()` decides to propose the pending changes at seeing a failed OSD, even if a pending change for the failed OSD was scheduled by the `PaxosService::prepare_update()` because of a failure report.

tchaikov: > we have to wait long enough to sentence the guilty OSD to death.

just want to clarify.. how long is "long enough"? i assume it will be triggered by some other changes so it can be piggy back with other pending changes? or maybe another MOSDFailure message targeting the same OSD?

src/mon/OSDMonitor.cc
```diff
@@ -7493,6 +7493,8 @@ bool OSDMonitor::prepare_command_impl(MonOpRequestRef op,
     ss << "will thrash map for " << thrash_map << " epochs";
     ret = thrash();
     err = 0;
+    if (ret)
```
tchaikov: in addition to "what", could you put the motive of this change into the commit, say, "why"?

also, `wait_for_finished_proposal()` does not trigger an update.

xiexingguo:
@tchaikov  Fix https://github.com/ceph/ceph/pull/7942/commits/539f34aef3334b48c2c2620a81e664ebf78b4a95 as @xiaoxichen 's gorgeous advice.
The time before monitor can make a decision an osd is dead is not constant, and you may check check_failure() for detail. Usually this change can be caught by other pending changes, but maybe it's not. I guess that's why the tick() thread shall do this check again.

No offend, it's OK to drop this change with me if you insist:)

tchaikov:
ah, i am good with this change. just wanted to check with you to see if i understand it correctly.

src/mon/OSDMonitor.cc
```diff
@@ -6625,8 +6625,11 @@ bool OSDMonitor::prepare_command_impl(MonOpRequestRef op,
     string addrstr;
     cmd_getval(g_ceph_context, cmdmap, "addr", addrstr);
     entity_addr_t addr;
-    if (!addr.parse(addrstr.c_str(), 0))
+    if (!addr.parse(addrstr.c_str(), 0)) {
```
tchaikov: could have simple test for this also.

tchaikov:
reviewed, would be better if we can have 
- a test for ef7df2c .
- and more detailed commit message for 8673f3d

tchaikov:
lgtm, @jecluis mind taking a look?

xiexingguo:
@tchaikov  Thank you for your careful review. I have learned so much, as always.

tchaikov:
@xiexingguo i am removing the jewel milestone from this pull request. as i think, despite that a80d6c5 helps with the performance of the cluster when some osd should be marked, but this PR is not ready for merge at the moment. so let's move it out of jewel.

xiexingguo:
@tchaikov OK:)

PR #7943, opened 2016-03-05
init-ceph.in: allow case-insensitive true in `osd crush update on start'

llua:
I noticed this when openstack/puppet-ceph sliently converted my 'true'
to 'True' in ceph.conf.

Signed-off-by: Eric Cook llua@gmx.com


liewegas:
lgtm

PR #7774, opened 2016-02-24
[DNM] rbd-mirror: ImageReplayer async start/stop

trociny:

PR #7944, opened 2016-03-05
rbd-mirror: ImageReplayer async start/stop

trociny:
This is reopened #7774, which was accidentally closed.

### DIFF
--- a/qa/workunits/rbd/rbd_mirror.sh
+++ b/qa/workunits/rbd/rbd_mirror.sh
@@ -1,0 +1,300 @@
+#!/bin/sh
+
+LOC_CLUSTER=local
+RMT_CLUSTER=remote
+POOL=mirror
+RBD_MIRROR_PID_FILE=
+RBD_MIRROR_ASOK=
+SRC_DIR=$(readlink -f $(dirname $0)/../../../src)
+TEMPDIR=
+
+#
+# Functions
+#
+
+setup()
+{
+    local c
+    trap cleanup INT TERM EXIT
+
+    TEMPDIR=`mktemp -d`
+
+    cd ${SRC_DIR}
+    ./mstart.sh ${LOC_CLUSTER} -n
+    ./mstart.sh ${RMT_CLUSTER} -n
+
+    ln -s $(readlink -f run/${LOC_CLUSTER}/ceph.conf) \
+       ${TEMPDIR}/${LOC_CLUSTER}.conf
+    ln -s $(readlink -f run/${RMT_CLUSTER}/ceph.conf) \
+       ${TEMPDIR}/${RMT_CLUSTER}.conf
+
+    cd ${TEMPDIR}
+
+    start_mirror
+
+    ceph --cluster ${LOC_CLUSTER} osd pool create ${POOL} 64 64
+    ceph --cluster ${RMT_CLUSTER} osd pool create ${POOL} 64 64
+
+    rbd --cluster ${LOC_CLUSTER} mirror pool enable ${POOL} pool
+    rbd --cluster ${RMT_CLUSTER} mirror pool enable ${POOL} pool
+
+    rbd --cluster ${LOC_CLUSTER} mirror pool peer add ${POOL} ${RMT_CLUSTER}
+    rbd --cluster ${RMT_CLUSTER} mirror pool peer add ${POOL} ${LOC_CLUSTER}
+}
+
+cleanup()
+{
+    test  -n "${RBD_MIRROR_NOCLEANUP}" && return
+
+    set +e
+
+    stop_mirror
+
+    cd ${SRC_DIR}
+
+    ./mstop.sh ${LOC_CLUSTER}
+    ./mstop.sh ${RMT_CLUSTER}
+
+    rm -Rf ${TEMPDIR}
+}
+
+start_mirror()
+{
+    RBD_MIRROR_PID_FILE=${TEMPDIR}/rbd-mirror.pid
+    RBD_MIRROR_ASOK=${TEMPDIR}/rbd-mirror.asok
+
+    rbd-mirror \
+	--cluster ${LOC_CLUSTER} \
+	--pid-file=${RBD_MIRROR_PID_FILE} \
+	--log-file=${TEMPDIR}/rbd-mirror.log \
+	--admin-socket=${RBD_MIRROR_ASOK} \
+	--debug-rbd=30 --debug-journaler=30 \
+	--debug-rbd_mirror=30 \
+	--daemonize=true
+}
+
+stop_mirror()
+{
+    if [ -z "${RBD_MIRROR_PID_FILE}" ]
+    then
+	return 0
+    fi
+
+    local pid
+    pid=$(cat ${RBD_MIRROR_PID_FILE} 2>/dev/null) || :
+    if [ -n "${pid}" ]
+    then
+	kill ${pid}
+	for s in 1 2 4 8 16 32; do
+	    sleep $s
+	    ps auxww | awk -v pid=${pid} '$2 == pid {print; exit 1}' && break
+	done
+	ps auxww | awk -v pid=${pid} '$2 == pid {print; exit 1}'
+    fi
+    rm -f ${RBD_MIRROR_ASOK}
+    rm -f ${RBD_MIRROR_PID_FILE}
+    RBD_MIRROR_PID_FILE=
+    RBD_MIRROR_ASOK=
+}
+
+flush()
+{
+    local image=$1
+    local image_id cmd
+
+    test -n "${RBD_MIRROR_ASOK}"
+
+    image_id=$(remote_image_id ${image})
+    test -n "${image_id}"
+
+    cmd=$(ceph --admin-daemon ${RBD_MIRROR_ASOK} help |
+		 sed -nEe 's/^.*"(rbd mirror flush.*'${image_id}'])":.*$/\1/p')
+    test -n "${cmd}"
+    ceph --admin-daemon ${TEMPDIR}/rbd-mirror.asok ${cmd}
+}
+
+wait_for_image_replay_started()
+{
+    local image=$1
+    local image_id s
+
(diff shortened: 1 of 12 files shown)


jdurgin:
looking good in general, the threadpool will conflict with @dillaman's #7906

trociny:
@dillaman @jdurgin I have a branch where this is rebased against  #7906:
https://github.com/trociny/ceph/commits/wip-rbd-mirror-image-replayer-async-on-wip-14869

If (when) #7906 is merged to master I will update this PR

trociny:
Rebased against the recent master.

After rebase I have been observing an issue with replaying snap remove (see e1072df). Also after rebase my rbd_mirror.sh test  fails when comparing remote and local image. This is I am going to investigate.

trociny:
@dillaman it crashes when snap_remove is replayed to local image.

dillaman:
@trociny Ah -- ok, I see the issue.

dillaman:
Hopefully addressed by PR #8028

trociny:
@dillaman #8028 looks like help with snap_remove crashes but now I have been observing another crush, reported under valgrind as:
```
==5317== Invalid read of size 8
==5317==    at 0x448B09: librbd::journal::Replay<librbd::ImageCtx>::handle_op_complete(unsigned long, int) (Replay.cc:651)
==5317==    by 0x394338: Context::complete(int) (Context.h:64)
==5317==    by 0x41E0D9: finish (Operations.cc:50)
==5317==    by 0x41E0D9: complete (Context.h:64)
==5317==    by 0x41E0D9: librbd::(anonymous namespace)::C_NotifyUpdate<librbd::ImageCtx>::complete(int) (Operations.cc:43)
==5317==    by 0x57F22E: ThreadPool::worker(ThreadPool::WorkThread*) (WorkQueue.cc:128)
==5317==    by 0x58014F: ThreadPool::WorkThread::entry() (WorkQueue.h:440)
==5317==    by 0xAAD60A3: start_thread (pthread_create.c:309)
==5317==    by 0xC7C104C: clone (clone.S:111)
==5317==  Address 0x2cac1610 is 0 bytes inside a block of size 256 free'd
==5317==    at 0xA8C3360: operator delete(void*) (vg_replace_malloc.c:507)
==5317==    by 0x40B569: librbd::Journal<librbd::ImageCtx>::stop_external_replay() (Journal.cc:732)
==5317==    by 0x3A3053: rbd::mirror::ImageReplayer::on_stop_journal_replay_shut_down_finish(int) (ImageReplayer.cc:689)
==5317==    by 0x395C1B: operator() (function_template.hpp:767)
==5317==    by 0x395C1B: FunctionContext::finish(int) (Context.h:460)
==5317==    by 0x394338: Context::complete(int) (Context.h:64)
==5317==    by 0x57F22E: ThreadPool::worker(ThreadPool::WorkThread*) (WorkQueue.cc:128)
==5317==    by 0x58014F: ThreadPool::WorkThread::entry() (WorkQueue.h:440)
==5317==    by 0xAAD60A3: start_thread (pthread_create.c:309)
==5317==    by 0xC7C104C: clone (clone.S:111)
```

So, I guess I need some flush before (in) `m_local_image_ctx->journal->stop_external_replay()`?

dillaman:
Yes -- you need to shut down ```librbd::journal::Replay``` not just delete it.  

https://github.com/trociny/ceph/blob/wip-rbd-mirror-image-replayer-async/src/librbd/Journal.cc#L935

trociny:
@dillaman Actually I do shut down:

https://github.com/trociny/ceph/blob/wip-rbd-mirror-image-replayer-async/src/tools/rbd_mirror/ImageReplayer.cc#L678

But this recalled me about another issue with the replay shutdown. When you added `bool cancel_ops` param to Replay::shut_down(), you updated ImageReplayer::stop() to call it with `cancel_ops=false`. When I did the same in async stop, I got:

```
`librbd/journal/Replay.cc: 181: FAILED assert(op_event.on_start_ready != nullptr)
 ceph version 10.0.4-2080-g7ccf678 (7ccf678303fad0d498bb23e48d1988c547f495e2)
 1: (ceph::__ceph_assert_fail(char const*, char const*, int, char const*)+0x82) [0x7f2bcc6e2452]
 2: (librbd::journal::Replay<librbd::ImageCtx>::shut_down(bool, Context*)+0x40f) [0x7f2bcc5b353f]
 3: (rbd::mirror::ImageReplayer::on_stop_journal_replay_shut_down_start()+0x123) [0x7f2bcc513cd3]
```
As it worked for me with `cancel_ops=true` and I thought it would be rather safe to cancel ops here (the journal commit position would not be updated and the ops would replay on the next run) I decided to change this to `true`. I wanted to discussed this with you later but forgot.

dillaman:
@trociny Missed that -- thanks.  I am having a hard time reconciling the backtrace since on master there is no code where the invalid read occurs:  https://github.com/ceph/ceph/blob/master/src/librbd/journal/Replay.cc#L651

dillaman:
OK, I think I'll need to build your branch because it seems like there is an op in-flight that is not being tracked somehow (or it raced with the destructor and the delete operation)

dillaman:
With regard to the shut_down cancel_ops, not sure how that failed the assert.  Again, I would need to repeat it and look at the logs.  OpEvent::on_op_finish_event and OpEvent::on_start_ready are supposed to be essentially mutually exclusive (so if you didn't have an ::on_op_finish_event, you should have an ::on_start_ready).

More generally, you should cancel the ops when you are shutting down due to an error and don't cancel the ops when normally shutting down.

trociny:
@dillaman In case you still have not figured out what Replay.cc:651 is, it is here:
```
#0  librbd::journal::Replay<librbd::ImageCtx>::handle_op_complete (this=0x7fff68010ed0, op_tid=1, r=0)
    at librbd/journal/Replay.cc:651
651       CephContext *cct = m_image_ctx.cct;
(gdb) l
646       return on_op_complete;
647     }
648
649     template <typename I>
650     void Replay<I>::handle_op_complete(uint64_t op_tid, int r) {
651       CephContext *cct = m_image_ctx.cct;
652       ldout(cct, 20) << this << " " << __func__ << ": op_tid=" << op_tid << ", "
653                      << "r=" << r << dendl;
654
655       OpEvent op_event;
(gdb) p m_image_ctx
$1 = (librbd::ImageCtx &) @0x7850000000001206: <error reading variable>
(gdb) p m_image_ctx.cct
Cannot access memory at address 0x7850000000001206
```

I differs from the master because of your patch from #8028 applied. I just have pushed my local branch to have it in sync with github.

dillaman:
@trociny I've been running a branch with #8028 and this PR merged with the following patch:

```
diff --git a/src/tools/rbd_mirror/ImageReplayer.cc b/src/tools/rbd_mirror/ImageReplayer.cc
index abe9d1a..332a8b9 100644
--- a/src/tools/rbd_mirror/ImageReplayer.cc
+++ b/src/tools/rbd_mirror/ImageReplayer.cc
@@ -576,7 +576,7 @@ void ImageReplayer::on_start_fail_finish(int r)
   }
 
   if (m_local_replay) {
-    shut_down_journal_replay(true);
+    shut_down_journal_replay(false);
     m_local_image_ctx->journal->stop_external_replay();
     m_local_replay = nullptr;
   }
@@ -1126,7 +1126,7 @@ cleanup:
   } else {
     dout(20) << "removing temporary snapshot " << m_snap_name << dendl;
     // XXXMG: disabled temporary, as it makes replay crash.
-    //r1 = remote_image_ctx->operations->snap_remove(m_snap_name.c_str());
+    r1 = remote_image_ctx->operations->snap_remove(m_snap_name.c_str());
     if (r1 < 0) {
       derr << "error removing snapshot " << m_snap_name << " of remote image "
           << m_remote_image_id << ": " << cpp_strerror(r1) << dendl;
```

Running ```unittest_rbd_mirror --gtest_repeat=1000 --gtest_break_on_failure --gtest_filter=TestImageReplayer.*``` doesn't crash for me.  Running under valgrind I see a couple memory leaks at the conclusion of the test, but no read/write after free.  I'll keep running it in the background overnight so see if I get a crash.

trociny:
@dillaman `unittest_rbd_mirror` works for me too, but when I try `ceph_test_rbd_mirror` I immediately get "FAILED assert(op_event.on_start_ready != nullptr)".

trociny:
@dillaman Looking in logs how the run for unittest_rbd_mirror differs from ceph_test_rbd_mirror, I see that in unittest_rbd_mirror, Replay::handle_op_complete() is always executed before shut_down(), so m_op_events is always empty. While in ceph_test_rbd_mirror case, handle_event() (where on_op_finish_event is cleared) is executed, but shut_down() happens before handle_op_complete().

So it looks like there is a window when on_op_finish_event is already cleared but the event is not removed yet from m_op_events. If shut_down() is run in this window, `if (op_event.on_op_finish_event != nullptr)` branch is skipped and "FAILED assert(op_event.on_start_ready != nullptr)" is triggered.

trociny:
@dillaman The issue is easily reproduced adding `sleep(1)` at the begin of  Replay<I>::handle_op_complete().

dillaman:
@trociny Awesome -- thanks for the reproducer.  I pushed a fix for that crash to #8028.  I am going to re-test with this PR on top to see if I can repeat the valgrind issue.

trociny:
@dillaman the valgring issue I observed only with ceph_test_rbd_mirror (not unittest_rbd_mirror) and `m_local_replay->shut_down(true, ctx)` (cancel_ops) in ImageReplayer. 

`sleep(1)` in handle_op_complete() does not help to reproduce this with unittest_rbd_mirror, but reveals "FAILED assert(op_it != m_op_events.end())" in handle_op_complete().

dillaman:
@trociny Yes, the issue was that an inflight snap remove was "canceled" by shut down, resulting in a call to "handle_op_complete".  However, since the snap remove was actually in-flight when it was canceled, it eventually calls back into "handle_op_complete", resulting in the issue.  The fix should be in my PR now.

PR #7945, opened 2016-03-05
ceph.in: Minor python3 specific changes

saru95:
Modified print and input statements to work with both Python2 and Python3.


tchaikov:
@saru95 looks good modulo that it would be better not to add a space after `print`. if you run pep8 against the code, it will complain that
```
E211 whitespace before '('
```

see https://www.python.org/dev/peps/pep-0008/ and https://pypi.python.org/pypi/pep8

PR #7948, opened 2016-03-06
msg: remove duplicated code - local_delivery will now call 'enqueue'

avnerbh:
This simplify the code and clarify that for each local message,
 'local_delivery_thread' simply tries to 'fast_dispatch' it,
 or it will 'enqueue' it for the dispatch_thread (like any standard message)

Signed-off-by: Avner BenHanoch <avnerb@mellanox.com>


tchaikov:
lgtm with qa run.

gregsfortytwo:
Huh, I think those paths used to be subtly different, but they look the same now. LGTM with that QA run!

PR #7949, opened 2016-03-06
Event: fix clock skew problem

wjin:
Default constructor of utime_t generates the same default time value,
so after the loop, we only keep the last list from map 'time_events'
into map 'change', thus it will lead to time event lost if swap them.


yuyuyu101:
good catch

yuyuyu101:
http://pulpito.ceph.com/haomai-2016-03-12_12:34:10-rados-wip-haomai-testing-distro-basic-multi/

PR #7952, opened 2016-03-07
mds: allow client to request caps when opening file

ukernel:
Allow client to request xattrs while opening file. MDS adds xattrs
and corresponding caps to the reply. This is important for selinux
on kcephfs.

Fixes: #14360

### DIFF
--- a/src/client/Client.cc
+++ b/src/client/Client.cc
@@ -128,6 +128,8 @@ using namespace std;
 #define O_DIRECT 0x0
 #endif
 
+#define DEBUG_GETATTR_CAPS (CEPH_CAP_XATTR_SHARED)
+
 void client_flush_set_callback(void *p, ObjectCacher::ObjectSet *oset)
 {
   Client *client = static_cast<Client*>(p);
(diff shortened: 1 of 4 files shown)


gregsfortytwo:
This needs a qa-suite test.

Also, it doesn't seem to include the bits for actually sending back xattrs on an open?

ukernel:
the new line "mdr->getattr_caps = mask" in handle_client_open() does the trick

gregsfortytwo:
Oh, because that's used in encode_inodestat() to include the xattrs on reply.

Still needs a test though. :)

ukernel:
https://github.com/ceph/ceph-qa-suite/pull/876

gregsfortytwo:
Test failed: http://pulpito.ceph.com/gregf-2016-03-16_13:01:47-fs-greg-fs-testing-316---basic-mira/66370/

gregsfortytwo:
@ukernel, also http://pulpito.ceph.com/gregf-2016-03-16_13:01:47-fs-greg-fs-testing-316---basic-mira/66414/

gregsfortytwo:
Oh, didn't read closely enough. Looks like it just needs to run against a different conf file.

>2016-03-16T13:15:26.261 INFO:tasks.cephfs_test_runner:Only have 1 clients, require 2

gregsfortytwo:
Pushed a quick change moving the test into multiclient and will test again.

gregsfortytwo:

PR #7953, opened 2016-03-07
[BlueStore] os/bluestore/KernelDevice: Make invalidate_cache work.

majianpeng:


liewegas:
I think before we start fiddling with any of the overlay write heuristics we need to make it work.. which means rewriting _do_write_overlays(). 

But I'm not sure we should bother yet?

majianpeng:
For _do_write_overlays, i don't know why you add "assert(0 == "this is all broken");". Could you explain in detain?  Thanks!

PR #7955, opened 2016-03-07
test/bufferlist: Avoid false-positive tests

ErwanAliasr1:
The current code is using the length of ptr to determine if the string is
correct.

As 'ptr' is a copy of the original 'str', it sounds safer to compare against the
original string size of the consider the actual size of the string : if 'ptr' is
shorter than 'str', the actual code will PASS while 'ptr' is broken.

This patch offer to use the same testing logic as per "ptr(const ptr& p,
unsigned o, unsigned l)" test by using a memcmp against the original string
length.


liewegas:
oh i see!  sounds good.

ErwanAliasr1:
Thx @liewegas

PR #7967, opened 2016-03-08
common/strtol.cc: fix the coverity warnings

tchaikov:
* promote the compared types properly to address the signed/unsigned
  comparison warnings. this also fixes the potential problems of
  slicing a compared type down to a "smaller" type before the
  comparison.
* check for the width of resulting type and shift bits caused by SI
  prefix to avoid the -Wshift-count-overflow warnings. this again
  is a potential issue, as shifting n bits of an integer of m bits
  width, where n >= m, leads to undefined behaviour.
* add a test for the 2nd fixed issue.

### DIFF
--- a/src/common/strtol.cc
+++ b/src/common/strtol.cc
@@ -166,14 +166,21 @@ T strict_si_cast(const char *str, std::string *err)
     *err = "strict_sistrtoll: value should not be negative";
     return 0;
   }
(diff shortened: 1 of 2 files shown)


badone:
LGTM

tchaikov:
changelog

* update the test to use a fresh err string every time, as we don't reset the err string if no error.

PR #7968, opened 2016-03-08
fix update pool peers

jazeltq:
Right now, the update pool peers is not work. So fix it.

Signed-off-by: Tianqing <tianqing@unitedstack.com>


jdurgin:
nice catch, we clearly need tests for this case

jdurgin:
oops, this is already fixed in #7958, closing this one as we don't need the extra copy of peers

PR #7970, opened 2016-03-08
msg/async: _try_send trim already sent for outcoming_bl more efficient

Yan-waller:

### DIFF
--- a/src/msg/async/AsyncConnection.cc
+++ b/src/msg/async/AsyncConnection.cc
@@ -395,9 +395,7 @@ ssize_t AsyncConnection::_try_send(bool send, bool more)
   // trim already sent for outcoming_bl
   if (sent_bytes) {
     if (sent_bytes < outcoming_bl.length()) {
-      bufferlist bl;
-      outcoming_bl.splice(sent_bytes, outcoming_bl.length()-sent_bytes, &bl);
-      bl.swap(outcoming_bl);
+      outcoming_bl.splice(0, sent_bytes);
     } else {
       outcoming_bl.clear();
     }


src/msg/async/AsyncConnection.cc
```diff
@@ -394,13 +394,8 @@ ssize_t AsyncConnection::_try_send(bool send, bool more)
 
   // trim already sent for outcoming_bl
   if (sent_bytes) {
-    if (sent_bytes < outcoming_bl.length()) {
-      bufferlist bl;
-      outcoming_bl.splice(sent_bytes, outcoming_bl.length()-sent_bytes, &bl);
-      bl.swap(outcoming_bl);
```
yuyuyu101: we only need to splice to a nullptr when sent_bytes < outcoming_bl.length()

src/msg/async/AsyncConnection.cc
```diff
@@ -394,13 +394,8 @@ ssize_t AsyncConnection::_try_send(bool send, bool more)
...
-    } else {
-      outcoming_bl.clear();
-    }
+    assert(sent_bytes <= outcoming_bl.length());
+    outcoming_bl.splice(0, sent_bytes);
```
Yan-waller: Yes, we want to trim already sent part, from my point of view, It seems to be easier to understand, and achieve the same effect as above (both sent_bytes < outcoming_bl.length() and sent_bytes = outcoming_bl.length()). It seems to be unnecessary to define bl and do swap operation. tks.

yuyuyu101: clear is still needed since it will be faster

Yan-waller:
@yuyuyu101 Done.

xiexingguo:
@Yan-waller Need rebase

PR #7971, opened 2016-03-08
selinux: Update policy to grant additional access

branto1:
This adds dac_override capability and ability to manage var_lock_t labelled locks.

EDIT: No need to bump the version and relabel the files as we do not change any context of the files, we just allow additional access.


liewegas:
passed rados teuthology run

PR #7688, opened 2016-02-18
buffer: hide iterator_impl symbols

tchaikov:
buffer::list::iterator_impl symbols are referenced by const_iterator
and iterator, and are exposed as weak symbols. if a source file is
compiled using the buffer.h, the produced object file will reference
these symbols as well, so we'd better hiding them and avoid using
them in the header file.
as a side-effect, buffer::list::const_iterator is also hidden, but
currently we don't have any librados client using this class, so
we can just leave it as an internal class at this moment.

Fixes: #14788
Signed-off-by: Kefu Chai <kchai@redhat.com>

PR #7974, opened 2016-03-08
buffer: add symmetry operator==() and operator!=()

tchaikov:
* add const_iterator(const iterator&) so an iterator can be converted to
  const_iterator
* add tests for operator==() and operator!=()

### DIFF
--- a/src/include/buffer.h
+++ b/src/include/buffer.h
@@ -263,6 +263,10 @@ namespace buffer CEPH_BUFFER_API {
     unsigned _memcopy_count; //the total of memcopy using rebuild().
     ptr append_buffer;  // where i put small appends.
 
+  public:
+    class iterator;
+
+  private:
     template <bool is_const>
     class CEPH_BUFFER_DETAILS iterator_impl
       : public std::iterator<std::forward_iterator_tag, char> {
@@ -281,6 +285,7 @@ namespace buffer CEPH_BUFFER_API {
       unsigned off; // in bl
       list_iter_t p;
       unsigned p_off;   // in *p
+      friend class iterator_impl<true>;
 
     public:
       // constructor.  position.
@@ -289,6 +294,7 @@ namespace buffer CEPH_BUFFER_API {
       iterator_impl(bl_t *l, unsigned o=0);
       iterator_impl(bl_t *l, unsigned o, list_iter_t ip, unsigned po)
 	: bl(l), ls(&bl->_buffers), off(o), p(ip), p_off(po) {}
+      iterator_impl(const list::iterator& i);
 
       /// get current iterator offset in buffer::list
       unsigned get_off() const { return off; }
@@ -304,12 +310,11 @@ namespace buffer CEPH_BUFFER_API {
 
       void advance(int o);
       void seek(unsigned o);
-      bool operator!=(const iterator_impl& rhs) const;
       char operator*() const;
       iterator_impl& operator++();
       ptr get_current_ptr() const;
 
-      bl_t& get_bl() { return *bl; }
+      bl_t& get_bl() const { return *bl; }
 
       // copy data out.
       // note that these all _append_ to dest!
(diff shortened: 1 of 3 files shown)


src/include/buffer.h
```diff
@@ -304,7 +310,6 @@ namespace buffer CEPH_BUFFER_API {
 
       void advance(int o);
       void seek(unsigned o);
-      bool operator!=(const iterator_impl& rhs) const;
```
jdurgin: I think we need this for abi compat since it was present in infernalis - not entirely sure for these operators though

https://techbase.kde.org/Policies/Binary_Compatibility_Issues_With_C%2B%2B

src/include/buffer.h
```diff
@@ -318,6 +323,15 @@ namespace buffer CEPH_BUFFER_API {
       void copy(unsigned len, list &dest);
       void copy(unsigned len, std::string &dest);
       void copy_all(list &dest);
+
+      friend bool operator==(const iterator_impl& lhs,
+			     const iterator_impl& rhs) {
+	return lhs.bl == rhs.bl && lhs.off == rhs.off;
```
jdurgin: should implement all these in buffer.cc (particularly because this is a public api)

src/include/buffer.h
tchaikov: jdurgin. the iterator_impl symbols are hidden from librados clients. see https://github.com/ceph/ceph/pull/7974/files#diff-e2d9d51a484aa6d89093b0bcfc70a1e0R271, and `CEPH_BUFFER_DETAILS` is defined as `__attribute__ ((visibility ("hidden")))`.

src/include/buffer.h
tchaikov: @jdurgin i refactored it to use the iterator_impl's public methods instead of accessing its member variable directly. does it look better to you?

tchaikov: and again, they are not public apis in the sense of symbol visibility.

src/include/buffer.h
tchaikov: but @jdurgin i did break the ABI compatibility in #7688 which violates https://techbase.kde.org/Policies/Binary_Compatibility_Examples#Unexport_or_remove_a_class: the iterator_impl was referenced in the header file by the public APIs so, as a sequence it is exported as weak symbols. and in #7688, i fixed it by 

1. moving the implementation of public interface into the .cc file as you suggested and
1. hiding the template symbols by using `__attribute__ ((visibility ("hidden")))`

this helps to bring the backward compatibility of librados clients but breaks the forward compatibility of librados clients compiled with v9.1.0 and v9.2.0. that's why @smithfarm suggest we bump the so version.

src/include/buffer.h
jdurgin: I see, sorry I didn't look closely enough to see the visibility("hidden") yesterday.

This looks fine then. Just need to be sure that the potentially inlined iterator::operators will still work if we change them or iterator_impl in the future. I doubt that will be necessary though.

src/include/buffer.h
jdurgin: I don't think we should bump the so version in this case, since it will cause issues with librados C users (e.g. packages like qemu or libvirt that depend on librados2), and it would only affect a c++ librados program compiled against infernalis iiuc. The move to c++11 in infernalis required a recompile for c++ librados users already, so another one for jewel doesn't seem worth the trouble an so version bump brings (especially this close to jewel).

tchaikov:
tested at http://pulpito.ceph.com/kchai-2016-03-11_19:18:43-rados-wip-kefu-testing---basic-smithi/ , failed tests are known issues and not related to the tested changes:

- http://tracker.ceph.com/issues/15077
- http://tracker.ceph.com/issues/15047
- http://tracker.ceph.com/issues/15106
- http://tracker.ceph.com/issues/15113

PR #7975, opened 2016-03-08
config: fix osd_crush_initial_weight

JiYou:
1. Fix missing config osd_crush_initial_weight in config_opts.h.
2. Add related docs for osd_crush_initial_weight.

### DIFF
--- a/doc/rados/configuration/pool-pg-config-ref.rst
+++ b/doc/rados/configuration/pool-pg-config-ref.rst
@@ -66,6 +66,16 @@ Ceph configuration file.
 :Default: ``1``. Typically a host containing one or more Ceph OSD Daemons.
 
 
+``osd crush initial weight``
+
+:Description: The initial crush weight for newly added osds into crushmap.
+
+:Type: Double
+:Default: ``the size of newly added osd in TB``. By default, the initial crush
+          weight for the newly added osd is set to its volume size in TB.
+          See `Weighting Bucket Items`_ for details.
+
+
 ``osd pool default crush replicated ruleset`` 
 
 :Description: The default CRUSH ruleset to use when creating a replicated pool.
@@ -164,3 +174,4 @@ Ceph configuration file.
 
 .. _pool: ../../operations/pools
 .. _Monitoring OSDs and PGs: ../../operations/monitoring-osd-pg#peering
+.. _Weighting Bucket Items: ../../operations/crush-map#weightingbucketitems
--- a/src/common/config_opts.h
+++ b/src/common/config_opts.h
@@ -550,6 +550,8 @@ OPTION(osd_client_message_cap, OPT_U64, 100)              // num client messages
 OPTION(osd_pg_bits, OPT_INT, 6)  // bits per osd
 OPTION(osd_pgp_bits, OPT_INT, 6)  // bits per osd
 OPTION(osd_crush_chooseleaf_type, OPT_INT, 1) // 1 = host
+// This parameter is not consumed by ceph C code but the upstart scripts.
+// OPTION(osd_crush_initial_weight, OPT_DOUBLE, 0) // the initial weight is for newly added osds.
 OPTION(osd_pool_use_gmt_hitset, OPT_BOOL, true) // try to use gmt for hitset archive names if all osds in cluster support it.
 OPTION(osd_pool_default_crush_rule, OPT_INT, -1) // deprecated for osd_pool_default_crush_replicated_ruleset
 OPTION(osd_pool_default_crush_replicated_ruleset, OPT_INT, CEPH_DEFAULT_CRUSH_REPLICATED_RULESET)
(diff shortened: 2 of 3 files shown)


src/common/config_opts.h
```diff
@@ -550,6 +550,7 @@ OPTION(osd_client_message_cap, OPT_U64, 100)              // num client messages
 OPTION(osd_pg_bits, OPT_INT, 6)  // bits per osd
 OPTION(osd_pgp_bits, OPT_INT, 6)  // bits per osd
 OPTION(osd_crush_chooseleaf_type, OPT_INT, 1) // 1 = host
+OPTION(osd_crush_initial_weight, OPT_DOUBLE, 0) // 0 means default value for osd added into crushmap.
```
tchaikov: this option is not read by ceph code. so i am sure we need to add it here.

doc/rados/configuration/pool-pg-config-ref.rst
```diff
@@ -66,6 +66,13 @@ Ceph configuration file.
 :Default: ``1``. Typically a host containing one or more Ceph OSD Daemons.
 
 
+``osd crush initial weight``
+:Description: The initial crush weight for the new osd just added into crushmap.
+:Type: Double
+:Default: ``0``. The initial crush weight for the new osd added into crushmap, by
```
tchaikov: this is not accurate, the default value is calculated using the size of the osd data volume . it's something like "sizeof($volume)/1TB".

JiYou:
@tchaikov thanks for your helpful suggestion.
I've checked ceph-osd-prestart.sh which acts like your comment.
thanks very much.

tchaikov:
@JiYou and we don't need to change the config_opt.h.

> this option is not read by ceph code. so i am not sure we need to add it here.

xiaoxichen:
@tchaikov , would the config_opt.h act as the collection of config knobs?  I mean, as we don't have a centrolized doc listing the config parameters, config_opt.h somehow act like such doc for some of us. Otherwise it would be too hard for even a experienced user to know this parameters.

xiaoxichen:
but yes, it's not used by ceph C code

tchaikov:
> config_opt.h somehow act like such doc for some of us. 

@xiaoxichen good point, if you insist, i guess we can keep it, but as a comment. what do you think?

JiYou:
@tchaikov @xiaoxichen Thanks for your suggestions. I will check and change it.

xiaoxichen:
@tchaikov something like :

// This parameters is not consumed by ceph but the upstart scripts
// OPTION(osd_crush_initial_weight, OPT_DOUBLE, 0) // the initial weight for newly added OSD.

Works for you ?

tchaikov:
@xiaoxichen yeah!

JiYou:
@tchaikov @xiaoxichen many thanks for you guys' suggestions.

src/common/config_opts.h
```diff
@@ -550,6 +550,8 @@ OPTION(osd_client_message_cap, OPT_U64, 100)              // num client messages
 OPTION(osd_pg_bits, OPT_INT, 6)  // bits per osd
 OPTION(osd_pgp_bits, OPT_INT, 6)  // bits per osd
 OPTION(osd_crush_chooseleaf_type, OPT_INT, 1) // 1 = host
+// This parameter is not consumed by ceph C code but the upstart scripts.
+OPTION(osd_crush_initial_weight, OPT_DOUBLE, 0) // the initial weight is for newly added osds.
```
xiaoxichen: Also comment this line out.

JiYou: OK

doc/rados/configuration/pool-pg-config-ref.rst
```diff
@@ -164,3 +172,4 @@ Ceph configuration file.
 
 .. _pool: ../../operations/pools
 .. _Monitoring OSDs and PGs: ../../operations/monitoring-osd-pg#peering
+.. _Weighting Bucket Items: ../../operations/crush-map.rst#Weighting Bucket Items
```
tchaikov: i am not sure if this works
> crush-map.**rst**#Weighting Bucket Items

JiYou: I will check.

doc/rados/configuration/pool-pg-config-ref.rst
```diff
@@ -164,3 +172,4 @@ Ceph configuration file.
 
 .. _pool: ../../operations/pools
 .. _Monitoring OSDs and PGs: ../../operations/monitoring-osd-pg#peering
+.. _Weighting Bucket Items: ../../operations/crush-map#Weighting Bucket Items
```
tchaikov: again this does not work. "Weighting Bucket Items" does not point to any anchor in crush-map.rst. you might want to create a hyper link target in that document first, then reference it from here.

i'd suggest you try `./admin/build-doc` and test before submitting the document related change.

JiYou:
@tchaikov Updated the doc and tried with admin/build-doc and & brower.
Now it works. Thanks for your time and patient.

PR #7976, opened 2016-03-08
Jewel 15008

ukernel:
http://tracker.ceph.com/issues/15008

### DIFF
--- a/src/client/Client.cc
+++ b/src/client/Client.cc
@@ -362,7 +362,10 @@ void Client::tear_down_cache()
 
 inodeno_t Client::get_root_ino()
 {
-  return root->ino;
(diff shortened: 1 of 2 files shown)


gregsfortytwo:
This looks good but we need a test validating its behavior.

ukernel:
https://github.com/ceph/ceph-qa-suite/pull/874

gregsfortytwo:
Stuff's noisy but this passed in http://pulpito.ceph.com/gregf-2016-03-18_16:26:03-fs-greg-fs-testing-317---basic-mira/ and http://pulpito.ceph.com/gregf-2016-03-23_11:13:03-fs-greg-fs-testing-317---basic-mira/

PR #7977, opened 2016-03-08
hammer: revert "OSD crashed when reached pool's max_bytes quota"

Vicente-Cheng:
http://tracker.ceph.com/issues/15019

### DIFF
--- a/src/osd/ReplicatedPG.cc
+++ b/src/osd/ReplicatedPG.cc
@@ -2293,8 +2293,10 @@ void ReplicatedPG::execute_ctx(OpContext *ctx)
     return;
   }
 
-  if (result == -ENOSPC) {
-    reply_ctx(ctx, result);
+  // check for full
+  if (ctx->delta_stats.num_bytes > 0 &&
+      pool.info.has_flag(pg_pool_t::FLAG_FULL)) {
+    reply_ctx(ctx, -ENOSPC);
     return;
   }
 
@@ -5684,15 +5686,6 @@ int ReplicatedPG::prepare_transaction(OpContext *ctx)
   if (ctx->op_t->empty() && !ctx->modify) {
     unstable_stats.add(ctx->delta_stats);
     return result;
-  }
-
-  // check for full
-  if ((ctx->delta_stats.num_bytes > 0 ||
-       ctx->delta_stats.num_objects > 0) &&
-      (pool.info.has_flag(pg_pool_t::FLAG_FULL) ||
-       get_osdmap()->test_flag(CEPH_OSDMAP_FULL))) {
-    dout(20) << __func__ << ": full, returning -ENOSPC" << dendl;
-    return -ENOSPC;
   }
 
   // clone, if necessary


dachary:
building in gitbuilders under the branch pr-7977

dachary:
* *pass* http://pulpito.ceph.com/loic-2016-03-08_18:43:43-fs-pr-7977---basic-multi/

asheplyakov:
This patch reintroduces the data corruption (in memory obc has been updated, but  the corresponding changes don't get written to the disk if the pool is full, which causes an error in copy-get operation).

Also dropping writes when the OSD is full is a correct behavior, isn't it?

asheplyakov:
The failed test tries to write 144 MB to the pool backed by ~ 95 MB (10^8 bytes) memstore. This operation should fail. The OSD is correctly rejecting writes. I think it's the test which should be fixed, not the OSD code. Not rejecting writes when the storage is full easily triggers further data corruption.

gregsfortytwo:
I'm not sure exactly why we're reverting, but this particular patch is an incomplete backport of a big full-space-handling rework. The way it's responding broke the MDS in at least one test (but did not break things in master), although we opted to continue with the release since it was during a shutdown anyway. We had a discussion on email and I think in the tracker as well which doesn't seem to be referenced from this PR or any of the linked tickets.
(That test is supposed to fill things up btw; it's testing full-space behavior!)

asheplyakov:
@gregsfortytwo 

> That test is supposed to fill things up btw; it's testing full-space behavior!

The test assumes it's possible to write more data (144 MB) than the storage capacity (95 MB) which is wrong.

> this particular patch is an incomplete backport of a big full-space-handling rework

The patch does two things:
- move the check for full to the correct place (before updating the cached obc)
- check for full OSD in addition to full pool (don't attempt to write to a full storage)

These changes does not require any messages or on disk format changes, and does not change the behavior, yet they are enough to fix/prevent the data corruption (and make cache pools usable).

dachary:
@gregsfortytwo I translated what I suppose you meant by "We had a discussion on email and I think in the tracker as well which doesn't seem to be referenced from this PR or any of the linked tickets." at http://tracker.ceph.com/issues/15019#note-7

I would be grateful if you could review that and confirm I'm not pointing at incorrect mails or notes.

dachary:
@asheplyakov does http://tracker.ceph.com/issues/15019#note-7 make sense to you ?

gregsfortytwo:
@dachary that's definitely part of it, but I'm a bit confused because it's missing the part with Josh where he actually elucidated the part about the incomplete backport of full handling, and his argument that we should do the backport. I thought it was in that email chain you found, so I'm not sure where the conversation took place...(maybe in irc? :/) You might want to ask him.

asheplyakov:
@dachary 
> does http://tracker.ceph.com/issues/15019#note-7 make sense to you?
yes

dachary:
@jdurgin do you remember the conversation @gregsfortytwo is refering to ? It looks like everyone is in agreement that we should not revert. But a log of this conversation would be useful for posterity :-)

PR #7979, opened 2016-03-08
rbd-mirror: remote to local cluster image sync

dillaman:
This PR currently includes commits from several dependent in-review PRs.  The following functionality is still required:

- [x] create snapshot and register sync point in journal
- [x] invoke snapshot copy state machine
- [x] loop over all object numbers (calculate max from snapshots and HEAD version) and invoke object copy state machine
- [x] sync snapshot object maps
- [ ] periodically update sync point in journal


dillaman:
@jdurgin I think this is good-to-go for review.  I opened task 15108 to add a periodic update of the last sync'ed object number and I opened 15109 to integrate this to within the ImageReplayer.

PR #7980, opened 2016-03-08
mon: initialize last_* timestamps on new pgs to creation time

liewegas:
Currently, when you create a pool, until the PGs peer you generate a health
error like

            8 pgs are stuck inactive for more than 300 seconds

which is inaccurate and misleading.  Instead, set the timestamps to the
creation time so that warnings don't appear until it's clear they're stuck.

Fixes: #14952


jcsp:
:+1: (just checked that this clears the HEALTH_ERRs right after cluster creation)

PR #7981, opened 2016-03-08
osdc/Objecter: fix narrow race with tid assignment

liewegas:
The operation completion could finish and be freed before we
do the info->register_tid assignment.  Avoid this by doing the
assignment in _op_submit itself.

Fixes: #14364

### DIFF
--- a/src/librados/IoCtxImpl.cc
+++ b/src/librados/IoCtxImpl.cc
@@ -745,7 +745,7 @@ int librados::IoCtxImpl::aio_operate_read(const object_t &oid,
   Objecter::Op *objecter_op = objecter->prepare_read_op(oid, oloc,
 		 *o, snap_seq, pbl, flags,
 		 onack, &c->objver);
-  c->tid = objecter->op_submit(objecter_op);
+  objecter->op_submit(objecter_op, &c->tid);
   return 0;
 }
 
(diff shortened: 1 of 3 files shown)


jdurgin:
lgtm

jdurgin:
once the int/ceph_tid_t compile error is fixed, lgtm

tchaikov:
```
[ RUN      ] TestJournalPlayer.PrefetchSkippedObject
test/journal/test_JournalPlayer.cc:377: Failure
Value of: wait_for_entries(player, 5, &entries)
  Actual: false
Expected: true
[  FAILED  ] TestJournalPlayer.PrefetchSkippedObject (10006 ms)
```

is this relevant?

jdurgin:
The failure is from journaling changes for rbd - checking it out.

I did look again at the bug this is addressing, and I think the race was in assigning to the completion in librados, which still needs fixing (maybe pass a pointer to it to the objecter methods called in IoCtxImpl so it can be assigned under the lock)

liewegas:
updated to fix the librados caller, too.

liewegas:
passed rados suite in its current form, but doesn't address the other librados callers.

jdurgin:
lgtm, covered all the librados callers

PR #7983, opened 2016-03-08
log: do not repeat errors to stderr

liewegas:
If we get an error writing to the log, log it only once to stderr.
This avoids generating, say, 72 GB of ENOSPC errors in
teuthology.log when /var/log fills up.

Fixes: #14616

### DIFF
--- a/src/log/Log.cc
+++ b/src/log/Log.cc
@@ -53,6 +53,7 @@ Log::Log(SubsystemMap *s)
     m_fd(-1),
     m_uid(0),
     m_gid(0),
+    m_fd_last_error(0),
     m_syslog_log(-2), m_syslog_crash(-2),
     m_stderr_log(1), m_stderr_crash(-1),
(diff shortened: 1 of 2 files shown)


src/log/Log.cc
```diff
@@ -337,9 +338,15 @@ void Log::_flush(EntryQueue *t, EntryQueue *requeue, bool crash)
       if (do_fd) {
         buf[buflen] = '\n';
         int r = safe_write(m_fd, buf, buflen+1);
-        if (r < 0)
-          cerr << "problem writing to " << m_log_file << ": " << cpp_strerror(r)
-	       << std::endl;
+	if (r > 0)
```
tchaikov: seems `safe_write()` does not return an `r > 0`

tchaikov:
lgtm modulo the nit.

tchaikov:
lgtm

zmc:
:beers:

PR #6781, opened 2015-12-03
common/WrrQueue: Add new Weighted Round Robin Queue

rldleblanc:
# Summary
This Weighted Round Robin Queue strives to provide fairness between all priorities based on the weight of their priority but help prevent starving lower priorities. Based on testing, the standard deviation of IOPs and bandwidth between multiple clients are reduced meaning that each client is getting about the same amount of work done. The queue is showing improved client performance for saturated spindle OSDs, but SSDs are showing a small drop in performance. The time to enqueue and dequeue is not much different between them, so it seems that the new behavior is exposing something else.

I've tested the code on my 6 node cluster and have dropped single OSDs and multiple OSDs out at the same time and have not been able to find any more bugs. This PR hard codes the use or WrrQueue instead of PrioritizedQueue and we may want an option to select it at run time. Sam said he could help with that if it was decided to go that way.

# Test Data
## FIO
Test|Values|Total IOPs|Total KB/s|Read IOPs|Read KB/s|Write IOPs|Write KB/s
---|------|----------|----------|----------|---------|----------|----------
4K Spindle Prio|Mean|45.83|189.75|23.08|95.43|22.75|94.32
  |Std. Dev.|7.01|27.54|3.55|13.67|3.49|13.90
4K Spindle Wrr|Mean|49.25|202.74|24.67|101.48|24.58|101.26
  |Std. Dev.|10.32|41.40|5.07|20.22|5.27|21.22
4M Spindle Prio|Mean|5.92|28,382.72|3.25|15,169.68|2.67|13,213.04
  |Std. Dev.|1.71|6,796.01|0.83|3,687.70|0.94|3,109.35
4M Spindle Wrr|Mean|6.73|31,762.36|3.45|16,951.55|3.27|14,810.82
  |Std. Dev.|0.86|3,419.74|0.50|1,779.21|0.45|1,642.91
4K SSD Prio|Mean|771|3,090.56|386.25|1,548.25|384.75|1,542.31
  |Std. Dev.|40.81|163.18|20.68|82.30|20.13|80.88
4K SSD Wrr|Mean|640.17|2,566.48|320.17|1,283.63|320.00|1,282.85
  |Std. Dev.|22.76|90.88|11.57|46.20|11.20|44.68
4M SSD Prio|Mean|10.42|46,948.08|5.33|24,111.08|5.08|22,837.00
  |Std. Dev.|1.85|7,181.80|0.85|3,342.93|1.04|3,842.92
[...]

PR #7984, opened 2016-03-08
common: WeightedPriorityQueue Re-Add Round Robin for Classes

rldleblanc:
In some corner cases the WPQ without round robin dequeue of classes caused starvation of other classes. This adds the round robin dequeue of classes back in as intrusive containers.

Signed-off-by: Robert LeBlanc <robert.leblanc@endurance.com>

### DIFF
--- a/src/common/WeightedPriorityQueue.h
+++ b/src/common/WeightedPriorityQueue.h
@@ -69,6 +69,12 @@ class WeightedPriorityQueue :  public OpQueue <T, K>
         Klass(K& k) :
           key(k)
           {}
+      friend bool operator< (const Klass &a, const Klass &b)
+        { return a.key < b.key; }
+      friend bool operator> (const Klass &a, const Klass &b)
+        { return a.key > b.key; }
+      friend bool operator== (const Klass &a, const Klass &b)
+        { return a.key == b.key; }
       void insert(unsigned cost, T& item, bool front) {
         if (front) {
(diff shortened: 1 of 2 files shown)


theanalyst:
This makes the master builds green again with boost 1.60.0

theanalyst:
Can we get this in before we cut a release to address the failures with boost 1.60

kmroz:
+1 @theanalyst

tchaikov:
@rldleblanc lgtm. did you perform a performance test against this chanceset as you did in #6781 ?

rldleblanc:
@tchaikov I did some tests on a previous version of this code (I pulled it out thinking it might be ok with the round robin between the clients). I had to clean it up a little for this pull request. In that test, I only did the SSDs because the spindles were already maxed out in the STL version. I did have about ~15% decrease in enqueue/dequeue times and 6.41% improvement on 4K SSD and 10.57% increase on SSD 4M.

Today is my last day at Endurance so I won't be able to do any more testing as I won't have access to the gear.

theanalyst:
@tchaikov @liewegas could we get this in one of Jewel RCs, it helps avoid build failures in distros that ship boost 1.60+

kmroz:
@liewegas - Awesome, thanks for adding this. Will make our lives a bit simpler :)
@theanalyst - Thanks for following up.

athanatos:
lgtm

PR #7985, opened 2016-03-08
rgw ldap

mattbenjamin:
this has the basics + S3 and librgw integration (swift to follow, should be reviewable for the general content)


yehudasa:
@mattbenjamin this PR is still marked as DNM, maybe remove it from the title?

mattbenjamin:
Also making the simple_bind opeation use a local LDAP, since it must be thread safe.  (I'm only sort of sure the underlying operation is thread safe, but will check.)

yehudasa:
Merged this PR accidentally instead of the correct branch with some build fixes. Reverted this PR, merged #8076 instead.

PR #7986, opened 2016-03-08
xxhash: use clone of xxhash.git; add .gitignore

liewegas:


liewegas:
https://github.com/ceph/xxHash/pull/1

PR #7994, opened 2016-03-09
Jewel 14996

ukernel:


src/client/Client.cc
```diff
@@ -10974,13 +10988,7 @@ int Client::ll_opendir(Inode *in, int flags, dir_result_t** dirpp,
...
   }
 
-  int r = 0;
-  if (vino.snapid == CEPH_SNAPDIR) {
-    *dirpp = new dir_result_t(in);
-  } else {
-    r = _opendir(in, dirpp, uid, gid);
-  }
-
+  int r = _opendir(in, dirpp, uid, gid);
```
gregsfortytwo: This looks fine to me, but can you explain why it's okay to reroute the snapshotted directories through _opendir, with the extra baggage that implies?

gregsfortytwo:
This is a simple enough patchset to clean stuff up, but you referenced unifying the open file and dir handling in the ticket. Did you decide that was too complicated?

ukernel:
Because I found that unifying open file/dir requires changing the API

src/client/Client.cc
ukernel: the extra baggage does not hurt and its overhead is low

gregsfortytwo:
http://pulpito.ceph.com/gregf-2016-03-09_00:11:22-fs-greg-fs-testing-38-1---basic-mira/
two failures are python libcephfs tests included in testing, and the other one is http://tracker.ceph.com/issues/14962. So this looks good.

PR #7995, opened 2016-03-09
osd: bluestore: allow _dump_onode dynamic accept log level

majianpeng:
Make _dump_onode dynamic accept different log level in different case.

### DIFF
--- a/src/os/bluestore/BlueStore.cc
+++ b/src/os/bluestore/BlueStore.cc
@@ -4829,11 +4829,11 @@ void BlueStore::_do_read_all_overlays(bluestore_wal_op_t& wo)
   return;
 }
 
-void BlueStore::_dump_onode(OnodeRef o)
+void BlueStore::_dump_onode(OnodeRef o, int log_level)
 {
-  if (!g_conf->subsys.should_gather(ceph_subsys_bluestore, 30))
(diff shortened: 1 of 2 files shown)


src/os/bluestore/BlueStore.h
```diff
@@ -578,7 +578,7 @@ class BlueStore : public ObjectStore {
 
   void _assign_nid(TransContext *txc, OnodeRef o);
 
-  void _dump_onode(OnodeRef o);
+  void _dump_onode(OnodeRef o, int log_leverl=30);
```
vshankar: minor nit: "int log_level=30"

src/os/bluestore/BlueStore.cc
```diff
@@ -5598,14 +5598,14 @@ int BlueStore::_do_write(
     if (p->second.has_flag(bluestore_extent_t::FLAG_UNWRITTEN)) {
       derr << __func__ << " left behind an unwritten extent, out of sync with "
 	   << "_do_allocate" << dendl;
-      _dump_onode(o);
+      _dump_onode(o, 0);
```
vshankar: Any reason why to invoke this as _dump_onode(o, 0) rather than _dump_onode(o, 30) which would preserve the earlier behaviour?

PR #6953, opened 2015-12-16
Multi-filesystem support

jcsp:

PR #7998, opened 2016-03-09
Multi-filesystem followups

jcsp:
This is a tidyup for https://github.com/ceph/ceph/pull/6953

Failing to pass features through to FSMap::encode has no effect a the moment, because it doesn't have any feature flags, but to avoid this being a gotcha in future we should pass features through.

### DIFF
--- a/src/messages/MFSMap.h
+++ b/src/messages/MFSMap.h
@@ -26,16 +26,18 @@ class MFSMap : public Message {
   bufferlist encoded;
 
   version_t get_epoch() const { return epoch; }
-  bufferlist& get_encoded() { return encoded; }
+  const FSMap & get_fsmap() {return fsmap;}
 
   MFSMap() : 
     Message(CEPH_MSG_FS_MAP), epoch(0) {}
-  MFSMap(const uuid_d &f, FSMap *fsmap) :
-    Message(CEPH_MSG_FS_MAP), epoch(fsmap->get_epoch())
+  MFSMap(const uuid_d &f, const FSMap &fsmap_) :
+    Message(CEPH_MSG_FS_MAP), epoch(fsmap_.get_epoch())
   {
-    fsmap->encode(encoded, -1);
+    fsmap = fsmap_;
   }
 private:
+  FSMap fsmap;
+
   ~MFSMap() {}
 
 public:
@@ -48,11 +50,11 @@ public:
   void decode_payload() {
     bufferlist::iterator p = payload.begin();
     ::decode(epoch, p);
-    ::decode(encoded, p);
+    ::decode(fsmap, p);
   }
   void encode_payload(uint64_t features) {
     ::encode(epoch, payload);
-    ::encode(encoded, payload);
+    ::encode(fsmap, payload, features);
   }
 };
 
--- a/src/mon/MDSMonitor.cc
+++ b/src/mon/MDSMonitor.cc
@@ -2419,7 +2419,7 @@ void MDSMonitor::check_sub(Subscription *sub)
 
   if (sub->type == "fsmap") {
     if (sub->next <= fsmap.get_epoch()) {
-      sub->session->con->send_message(new MFSMap(mon->monmap->fsid, &fsmap));
+      sub->session->con->send_message(new MFSMap(mon->monmap->fsid, fsmap));
       if (sub->onetime) {
         mon->session_map.remove_sub(sub);
       } else {
--- a/src/mon/MDSMonitor.cc
+++ b/src/mon/MDSMonitor.cc
@@ -133,13 +133,13 @@ void MDSMonitor::update_from_paxos(bool *need_bootstrap)
   assert(version >= fsmap.epoch);
 
   // read and decode
-  mdsmap_bl.clear();
-  int err = get_version(version, mdsmap_bl);
+  fsmap_bl.clear();
+  int err = get_version(version, fsmap_bl);
   assert(err == 0);
 
-  assert(mdsmap_bl.length() > 0);
+  assert(fsmap_bl.length() > 0);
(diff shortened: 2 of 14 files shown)


gregsfortytwo:
@jcsp, can you rebase this? I think it's just the last commit that's for real.

gregsfortytwo:
@jcsp this broke cmake build. http://gitbuilder.sepia.ceph.com/gitbuilder-ceph-tarball-trusty-amd64-cmake/log.cgi?log=c21e3a49fb5fdcfe2cb63d655c4eda39a9c1c9bd

>In file included from /srv/autobuild-ceph/gitbuilder.git/build/src/include/types.h:18:0,
from /srv/autobuild-ceph/gitbuilder.git/build/src/include/utime.h:23,
from /srv/autobuild-ceph/gitbuilder.git/build/src/include/cephfs/libcephfs.h:19,
from cephfs.cpp:361:
/srv/autobuild-ceph/gitbuilder.git/build/src/include/int_types.h:4:22: fatal error: acconfig.h: No such file or directory
#include "acconfig.h"

jcsp:
The cmake breakage will be because the branch had my hack for cython in it (I remove it when I'm done hacking)

gregsfortytwo:
Testing passed: http://pulpito.ceph.com/gregf-2016-03-15_04:07:43-fs-greg-fs-testing-314---basic-mira/
Still pending review.

jcsp:
This PR targeted for master, but we'll want to get at least some of this stuff into Jewel (the "mds: fix mds_info_t::dump" commit especially is a clear bug).

jcsp:
@gregsfortytwo ready for review now

src/messages/MFSMap.h
```diff
@@ -26,16 +26,18 @@ class MFSMap : public Message {
...
 
   MFSMap() : 
     Message(CEPH_MSG_FS_MAP), epoch(0) {}
-  MFSMap(const uuid_d &f, FSMap *fsmap) :
-    Message(CEPH_MSG_FS_MAP), epoch(fsmap->get_epoch())
+  MFSMap(const uuid_d &f, const FSMap &fsmap_) :
+    Message(CEPH_MSG_FS_MAP), epoch(fsmap_.get_epoch())
   {
-    fsmap->encode(encoded, -1);
+    fsmap = fsmap_;
```
gregsfortytwo: Hmm. Do we have any reason to think the fsmap reference will still be valid by the time we get around to encoding it?

gregsfortytwo: ...wow did I read that wrong. Never mind.

src/messages/MFSMap.h
```diff
@@ -48,11 +50,11 @@ class MFSMap : public Message {
   void decode_payload() {
     bufferlist::iterator p = payload.begin();
     ::decode(epoch, p);
-    ::decode(encoded, p);
+    ::decode(fsmap, p);
   }
   void encode_payload(uint64_t features) {
     ::encode(epoch, payload);
-    ::encode(encoded, payload);
+    ::encode(fsmap, payload, features);
```
gregsfortytwo: So if we're changing the wire encoding like this it certainly needs to go in Jewel. Don't think that will be a problem as we're still merging master in to Jewel; sound good to you?

src/mon/MDSMonitor.cc
```diff
@@ -144,7 +144,9 @@ void MDSMonitor::update_from_paxos(bool *need_bootstrap)
   // new map
   dout(4) << "new map" << dendl;
   print_map(fsmap, 0);
-  fsmap.sanity();
+  if (!g_conf->mon_mds_skip_sanity) {
+    fsmap.sanity();
+  }
```
gregsfortytwo: We want to do the config check in FSMap itself; Don't Repeat Yourself and that way any future callers don't need to remember to check it.

gregsfortytwo:
So the other follow-up I was interested in and should be done quickly is that we reproduce the entire MDSMap::decode() function in FSMap::decode(). It seems to be working now but is awfully fragile in that state. Can you whip up a patch that makes it reference MDSMap::decode instead?
(So actually because of the ENCODE_START macros you'll want to put the real work into an MDSMap::_decode() function which takes struct_v as well as the iterator&, and then call into that from both decodes, but you get the idea.)

src/mon/MDSMonitor.cc
jcsp: I decided that repeating it in MDSMonitor was the lesser of two evils compared with passing a reference to the config object through into FSMap

src/messages/MFSMap.h
jcsp: I didn't think passing explicit features arguments (vs. how it was before) to ::encode actually changed the encoding on the wire, does it?

Obviously changing the actual feature bits changes things, but this is just changing it to pass through the globally configured features as opposed to having a hardcoded set further down.

jcsp:
I haven't forgotten about refactoring the decode method, but let's save that for another PR

src/messages/MFSMap.h
gregsfortytwo: I wasn't talking about using the feature bits. There's a difference between encoding an FSMap into the payload, and encoding a bufferlist (of an encoded FSMap) into the payload. (There is the usual wrapping around a bufferlist, right? I don't think it's just an append.)

jcsp: Oh yes, quite right.  We're dropping the leading four bytes of the bl's length.

src/mon/MDSMonitor.cc
gregsfortytwo: Mmm, okay. Sometimes I miss having globals you can reference gratuitously. ;)

gregsfortytwo:

PR #8000, opened 2016-03-09
msg/async: skip unnecessary steps when parsing simple messages

branch-predictor:
Many simple messages (like ACKs) do not have data, so there's no
need to go through STATE_OPEN_MESSAGE_READ_DATA_PREPARE and
STATE_OPEN_MESSAGE_READ_DATA states. In case when also there's no
middle chunk, we can jump straight from STATE_OPEN_MESSAGE_READ_FRONT
to STATE_OPEN_MESSAGE_READ_FOOTER_AND_DISPATCH.

### DIFF
--- a/src/msg/async/AsyncConnection.cc
+++ b/src/msg/async/AsyncConnection.cc
@@ -730,8 +730,18 @@ void AsyncConnection::process()
 
             ldout(async_msgr->cct, 20) << __func__ << " got front " << front.length() << dendl;
           }
-          state = STATE_OPEN_MESSAGE_READ_MIDDLE;
-          break;
+          if (current_header.middle_len) {
+            state = STATE_OPEN_MESSAGE_READ_MIDDLE;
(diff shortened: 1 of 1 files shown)


src/msg/async/AsyncConnection.cc
```diff
@@ -730,7 +730,16 @@ void AsyncConnection::process()
 
             ldout(async_msgr->cct, 20) << __func__ << " got front " << front.length() << dendl;
           }
-          state = STATE_OPEN_MESSAGE_READ_MIDDLE;
+          if (current_header.middle_len) {
```
yuyuyu101: actually, I expect we remove break statement and flow down the state, I think it will be cheap too.

branch-predictor: Yeah. Tbh, yesterday I thought it's not looping, but actually it is as long as state changes, so my note related to mutexes is not true anyway.

branch-predictor:
Superseded by #8072.

PR #8005, opened 2016-03-09
pybind: flag an RBD image as closed regardless of result code

dillaman:


dillaman:
http://qa-proxy.ceph.com/teuthology/joshd-2016-03-08_20:12:27-rbd-master---basic-mira/48172/teuthology.log

PR #8008, opened 2016-03-09
mon: default ec min_size to k+1

cernceph:
If m OSDs are down and we allow writes to the remaining k, the
PG could suffer split brain problems if one of these k goes down
during subsequent recovery. Default min_size to k+1 to be a bit
safer.

### DIFF
--- a/src/mon/OSDMonitor.cc
+++ b/src/mon/OSDMonitor.cc
@@ -4519,7 +4519,7 @@ int OSDMonitor::prepare_pool_size(const unsigned pool_type,
       err = get_erasure_code(erasure_code_profile, &erasure_code, ss);
       if (err == 0) {
 	*size = erasure_code->get_chunk_count();
-	*min_size = erasure_code->get_data_chunk_count();
+	*min_size = MIN(erasure_code->get_data_chunk_count() + 1, *size);
       }
     }
     break;


athanatos:
Not split brain, right?  Just inability to peer (that is, we'll go unavailable, not return wrong answers).

athanatos:
Not that this isn't a good idea...

athanatos:
I do think this messes up existing tests, you'll have to check through and see whether it would change the min_size for any of the ceph-qa-suite ec tests and update them to put it back.

dvanders:
I've updated the commit message.
I'll check the tests now.

dvanders:
@athanatos I'm finally getting my head around the qa-suite. I didn't see a way to override the min_size for ec pools, so I made that optional in rados.py. Does this look ok: https://github.com/dvanders/ceph-qa-suite/commit/0551d838d71f1fb5e7b227f4d9b811edf971e3e6

I spent a bit of time first on the make check tests, since I was confused why they weren't failing now. Turns out that because we use 10 osd clusters so this won't be a problem there.

athanatos:
@dvanders That ceph-qa-suite thing looks right, can you push it as a branch and submit a ceph-qa-suite PR linked in a comment here (I've got a corresponding ceph-qa-suite wip-sam-testing)?  I'll get this one on the next testing round.

yuriw:

gregsfortytwo:
This appears to have broken unit tests, if you look at http://gitbuilder.sepia.ceph.com/gitbuilder-ceph-tarball-trusty-amd64-basic/#origin/master

I pushed a revert to check.

yuriw:
@liewegas ??

zphj1987:
default K=2 M=1
and can not down 1 osd? pg to be incomplete

imjustmatthew:
Wow, this really needs to be better documented somewhere. This effectively causes k=2, m=1 erasure coded pools to block during recovery from a single failed OSD when they would be capable of recovering normally otherwise. Ideally though, recovery needs to somehow proceed autonomously even if we're going to keep blocking client writes.

dvanders:
Maybe we should do:
```
  if m < 2:
    min_size = k
  else:
    min_size = k+1
```

liewegas:
@dvanders i was thinking the same thing.  @neha-ojha @tchaikov ?

tchaikov:
👍 , or put another way
```
  min_size = k + min(1, m - 1)
```

neha-ojha:
looks good to me

liewegas:
https://github.com/ceph/ceph/pull/26894

PR #8009, opened 2016-03-09
test: reproducer for writeback CoW deadlock

dillaman:
Refs: #14988


dillaman:
Reproducer hits the condition only on hammer and infernalis branches -- pushing new test case to master for backport.

dillaman:
Now there's something you don't see everyday:

> [ RUN      ] TestJournalReplay.SnapRename
> test/librbd/journal/test_Replay.cc:451: Failure
> Expected: (((__u64)(-2))) != (snap_id), actual: 18446744073709551614 vs 18446744073709551614
> [  FAILED  ] TestJournalReplay.SnapRename (65 ms)

PR #8015, opened 2016-03-10
DNM rgw: Add an implicit tenant for Keystone

zaitcev:
This may be a bit confusing because Keystone already has perfectly good tenants, which have absolutely nothing with our own tenants.


yehudasa:
PR #8139 replaces this one

PR #8016, opened 2016-03-10
rpm,deb: remove conditional BuildRequires for btrfs-progs

ErwanAliasr1:
By pull request 7742, the btrfs-progs package was considered as a BuildRequires
only when --with tests was engaged like :

    if %{with tests}
    BuildRequires: btrfsprogs
    %endif

That's perfectly valid for a spec file.

The issue we have is the following :
- yum-builddep called by install-deps.sh is used to prepare the build env by
installing the needed BuildRequires.
- %{with test} is defined by using a %bcond_with
- yum-builddep doesn't consider %{with test} as valid
- yum-builddep doesn't install the btrfs package

As per discussions with the yum team, there is no way to engage conditional flags with
yum-builddep.

So this patch, as per discussions with Nathan Cutler & Loic Dachary, is removing
the condition arond the BuildRequires.

Note that all BuildRequires defined with a %bcond_with would be affected by this
issue. The current specfile only have %bcond_without conditional BuildRequires
which is fine.

Fixes: #15042

### DIFF
--- a/ceph.spec.in
+++ b/ceph.spec.in
@@ -152,9 +152,7 @@ BuildRequires:	libbz2-devel
 %if 0%{with tcmalloc}
 BuildRequires:	gperftools-devel
 %endif
-%if 0%{with tests}
 BuildRequires:  btrfsprogs
-%endif
 BuildRequires:	mozilla-nss-devel
 BuildRequires:	keyutils-devel
 BuildRequires:	libatomic-ops-devel
@@ -165,9 +163,7 @@ BuildRequires:	python-Cython
 %if 0%{?_with_systemd}
 Requires:	systemd
 %endif
-%if 0%{with tests}
 BuildRequires:	btrfs-progs
-%endif
 BuildRequires:	nss-devel
 BuildRequires:	keyutils-libs-devel
 BuildRequires:	libatomic_ops-devel


smithfarm:
@ktdreyer What do you think about this approach? Another approach just occurred to me: the failing tests themselves could be re-done to not run if the relevant btrfs programs are not installed.

ErwanAliasr1:
@smithfarm but you cannot really get what is the underlying filesystem before running the test. At least, the current approach insures that it will work everywhere. btrfs-progs is not a big dep.

smithfarm:
@ErwanAliasr1 But I wrote "to not run if the relevant btrfs programs are not installed" not "to run only if the underlying filesystem is btrfs".

ErwanAliasr1:
@smithfarm I meant that if your underlying fs is btrfs and if the btrfs progs are not installed then many tests would be bypasses which is not very nice.

smithfarm:
OK, thanks for the clarification. Let's wait and see what the others think.

ErwanAliasr1:
Others ? What's your thoughts on that ?

ErwanAliasr1:
@ktdreyer any ideas on that issue  we are discussing ?

wjwithagen:
On 16-3-2016 10:53, Erwan Velu wrote:
> @ktdreyer <https://github.com/ktdreyer> any ideas on that issue we are
> discussing ?

It would also require removing all the btrfs related code from the
testscripts. And digging into other sources to see where btrfs-isms are
being used.

If removing the dependancie on removing just a package is the purpose,
I'd say: why bother, if you're leaving the rest in.

my 2 cts,
--WjW

ErwanAliasr1:
@wjwithagen I've been fixing the btrfs issues in the test script and use them on a daily basis. So it works. Its just a matter of accepting keeping that build dep without any conditional statement.

liewegas:
dropping the conditional seems fine to me!

ErwanAliasr1:
So do I @liewegas. I'm updating this way.

ErwanAliasr1:
Can we merge that PR ?

smithfarm:

PR #8017, opened 2016-03-10
osd: try remove more stale osdmaps

tchaikov:
if the scale of cluster is large, there are chances that the OSD fails
to trim the cached osdmap in a timely manner. and sometimes, it is just
unable to keep up with the incoming osdmap, so the osdmap cache can
keep building up to over 250GB in size. in this change, we loose the
limit of the transaction size of removing stale osdmap so it can remove
at least 1/2 stale osdmaps. the limit was
 max(osd_target_transaction_size, m->get_last() - m->get_first())

Fixes: #13990

### DIFF
--- a/src/osd/OSD.cc
+++ b/src/osd/OSD.cc
@@ -223,6 +223,7 @@ OSDService::OSDService(OSD *osd) :
   pg_epoch_lock("OSDService::pg_epoch_lock"),
   publish_lock("OSDService::publish_lock"),
   pre_publish_lock("OSDService::pre_publish_lock"),
+  max_oldest_map(0),
   peer_map_epoch_lock("OSDService::peer_map_epoch_lock"),
   sched_scrub_lock("OSDService::sched_scrub_lock"), scrubs_pending(0),
   scrubs_active(0),
@@ -1183,7 +1184,7 @@ MOSDMap *OSDService::build_incremental_map_msg(epoch_t since, epoch_t to,
                                                OSDSuperblock& sblock)
 {
   MOSDMap *m = new MOSDMap(monc->get_fsid());
-  m->oldest_map = sblock.oldest_map;
+  m->oldest_map = max_oldest_map;
   m->newest_map = sblock.newest_map;
 
   for (epoch_t e = to; e > since; e--) {
@@ -1223,7 +1224,7 @@ void OSDService::send_incremental_map(epoch_t since, Connection *con,
     if (since < sblock.oldest_map) {
       // just send latest full map
       MOSDMap *m = new MOSDMap(monc->get_fsid());
-      m->oldest_map = sblock.oldest_map;
+      m->oldest_map = max_oldest_map;
       m->newest_map = sblock.newest_map;
       get_map_bl(to, m->maps[to]);
       send_map(m, con);
@@ -2161,6 +2162,7 @@ int OSD::init()
   service.init();
   service.publish_map(osdmap);
   service.publish_superblock(superblock);
+  service.max_oldest_map = superblock.oldest_map;
 
   osd_lock.Unlock();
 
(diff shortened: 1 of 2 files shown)


gregsfortytwo:
Hmm, my recollection is that we limited the size of map trimming for a reason — in particular, if the cluster was dirty for a long time you could have a *lot* of maps suddenly get marked as stale. Enough of them to time out the OSD on the LevelDB transaction deleting them.

So we need some fixed limit which isn't relative to the number of stale maps, but perhaps we can try and scale how much of our disk throughput we devote to map trimming as the rate of map generation goes up? :/

tchaikov:
thanks Greg. do you mean BlueStore? FileStore is not supported by a KeyValueDB. and osdmaps are stored to ObjectStore on OSD.

so can i break the transactions into smaller pieces so map trimming's remove operations won't starve the other osd ops on ObjectStore or timeout the underlying KeyValueDB?

gregsfortytwo:
Oh, I was thinking OSDMaps go into the omap key-value area, but that's not actually true, is it? :)
The important bit was the size of the remove transaction needs to be limited to a size the OSD can actually perform in a reasonable amount of time. I'm a bit fuzzy on it but just thought I remembered that going by; @athanatos undoubtedly remembers more. Breaking up the transaction into smaller ones should be fine but is essentially what the limit is already supposed to be doing — is it failing just because if generating maps fast enough we might see more new OSDMaps in a single MOSDMap message than the osd_target_transaction_size limit?

tchaikov:
> Oh, I was thinking OSDMaps go into the omap key-value area, but that's not actually true, is it? :)

no. each (incremental) osdmap is stored as an individual object in meta collection.

> @athanatos undoubtedly remembers more.

thank you, i will ping him for more info.

>  is it failing just because if generating maps fast enough we might see more new OSDMaps in a single MOSDMap message than the osd_target_transaction_size limit?

basically because 

1. a MOSDMap message came in, but somehow the cached osdmaps in PG prevent the OSD from trimming some stale osdmaps. 
1. this OSD populates the latest osdmaps to its peers, but this time the oldest_map fields it greater than the one it received
1. this effect accumulates and is populated across the cluster

that worsen the problem when the cluster is larger. see http://tracker.ceph.com/issues/13990 for more details.

tchaikov:
note to myself

1. we should break the removing ops into smaller transactions. and we can even have a dedicated transaction for trimming maps in the normal case.
1. before each transaction, we `publish_superblock()` with the increased `superlock.oldest_map`, so other components in OSD won't access the osdmaps being removed. this actually fixes a bug in existing code.
1. `assert()` that the `service.map_cache.cached_key_lower_bound()` does not decrease blow `min` after the trimming. that would be a disaster if we are trimming a cached osdmap.

tchaikov:
changelog:

  removed.
  without enforcing an upper bound of
  max(conf->osd_target_transaction_size, m->get_last() - m->get_first())
  we had before.

tchaikov:
the hammer backport is available at https://github.com/tchaikov/ceph/commit/70a1c6df77e2977d2b2fded738a7b80a5cbbaa1b

tchaikov:
being tested at http://pulpito.ceph.com/kchai-2016-03-11_19:18:43-rados-wip-kefu-testing---basic-smithi/

tchaikov:
changelog

- rebased against master
- reset the `num` of pending removal commits after queue the transaction

tchaikov:
will rebase against #8096 once it gets merged.

tchaikov:
changelog

- rebase against #8096 
- populate the maximum oldest_map using oldest_map, see http://tracker.ceph.com/issues/13990#note-79

liewegas:
I don't think b545fb4 is a good idea:

1. If an OSD is way behind on trimming, or if the lower bound on last_epoch_clean jumps forward a long way at once, we will do a huge series of delete transactions all at once that delete everything, whereas previously we would slowly catch up as we get more maps.
2. Assuming we chage that, we lose the logic that we trim *at least* as many maps as we just received.

I didn't read the bug super closely, but it seems like just the last patch would be enough to address the original problem?  Or maybe adjusting the 'trim at least' logic to do 2x what we received or something?

tchaikov:
>  previously we would slowly catch up as we get more maps.

no, we cannot. as the frequency of receiving osdmap messages is relatively much lower than that of new osdmap is generated.

> but it seems like just the last patch would be enough to address the original problem?

agreed. will remove https://github.com/ceph/ceph/commit/b545fb49c7ecf00594fdbdacd698a23166f56be8 from this PR.

tchaikov:
changelog

- rebase against master
- remove the "remove all stale maps" change

liewegas:
merged manually

PR #8018, opened 2016-03-10
DNM: drop permissions of rgw/civetweb after startup - revised

kmroz:
An alternative to pull#7313.

Differs in that it stores --setuser/--setgroup strings in CephContext to be used by civetweb directly when dropping permissions.

Fixes: http://tracker.ceph.com/issues/13600


kmroz:
Marked DNM as it needs a quick sanity test (soon).

PR #7313, opened 2016-01-21
drop permissions of rgw/civetweb after startup

kmroz:
http://tracker.ceph.com/issues/13600

Start rgw/civetweb as root, then rely on civetweb to drop permissions to specified user (or "ceph" by default).

In order to tighten this solution even more, the following commits should be backported to the civetweb submodule:
https://github.com/civetweb/civetweb/commit/5e753cc
https://github.com/civetweb/civetweb/commit/a3c460c
My understanding is that a new branch would need to be created under ceph/civetweb with these commits added, followed by civetweb submodule commit. @yehudasa - is that something you might be able to help with?

PR #8019, opened 2016-03-10
rgw: drop permissions of rgw/civetweb after startup

kmroz:
An alternative to https://github.com/ceph/ceph/pull/7313

Differs in that it stores --setuser/--setgroup strings in CephContext to be used by civetweb directly when dropping permissions.

Fixes: http://tracker.ceph.com/issues/13600

### DIFF
--- a/src/common/ceph_context.cc
+++ b/src/common/ceph_context.cc
@@ -441,6 +441,8 @@ CephContext::CephContext(uint32_t module_type_, int init_flags_)
     _init_flags(init_flags_),
     _set_uid(0),
     _set_gid(0),
+    _set_uid_string(),
+    _set_gid_string(),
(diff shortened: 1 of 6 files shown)


kmroz:
Marked as DNM as needs some sanity testing (soon).

kmroz:
Quick sanity test on my side passes:

fastcgi frontend - 2016-03-10 14:01:04.628487 7fddee481840  0 set uid:gid to 487:486 (ceph:ceph)
civetweb frontend - 2016-03-10 14:02:50.581841 7f3deba69840  0 deferred set uid:gid to 487:486 (ceph:ceph)

User under which radosgw runs is dropped to "ceph" in both cases via the appropriate code paths.

PR #8020, opened 2016-03-10
osd: use foreach in a small assortment of Session-handling loops

songbaisen:
osd: use foreach in a small assortment of Session-handling loops

### DIFF
--- a/src/osd/OSD.h
+++ b/src/osd/OSD.h
@@ -1343,19 +1343,12 @@ public:
 	   session_waiting_for_pg.begin();
 	 i != session_waiting_for_pg.end();
 	 ++i) {
-      for (set<Session*>::iterator j = i->second.begin();
-	   j != i->second.end();
-	   ++j) {
-	(*j)->put();
-      }
+        std::for_each(i->second.begin(), i->second.end(), mem_fun(&Session::put));
     }
+
     session_waiting_for_pg.clear();
 
-    for (set<Session*>::iterator i = session_waiting_for_map.begin();
-	 i != session_waiting_for_map.end();
-	 ++i) {
-      (*i)->put();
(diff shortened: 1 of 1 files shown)


gregsfortytwo:
This is cute but the functions involved are invoked infrequently. Did you have some reason for targeting them?

songbaisen:
@gregsfortytwo Thanks for review! 
The reason is make the code more sort and simple.The most reason is make the code run more faster。

[root@client101 testsong]# ./test9
t1: 1772284 ns
t2: 1705622 ns

gcc version 4.8.5 20150623 (Red Hat 4.8.5-4) (GCC)


class Test{
public:
void put()
{
    delete this;
}

};

int main()
{
   unsigned long long t1, t2;
   t1 = gettickcount();
   vector<Test *> v1;
   vector<Test *> v2;
   for(int i=0; i<100000; i++)
   {
       Test *T = new Test();
       v1.push_back(T);
       v2.push_back(T);
   }
   t1 = gettickcount();
   vector<Test *> :: iterator iter = v1.begin();
   while(iter != v1.end())
   {
       (*iter)->put();
       ++iter;
   }
   t1 = gettickcount() - t1;

   t2 = gettickcount();
   for_each(v1.begin(), v1.end(),mem_fun(&Test::put));
   t2 = gettickcount() - t2;
   printf("t1: %llu ns \n", t1);
   printf("t2: %llu ns \n", t2);
}

athanatos:
This might be worthwhile if there were other changes happening in the same area, but I don't really want to merge this on its own.  Also, it would be annoying to add debugging if we found that we wanted to add debugging in those loops.

songbaisen:
@athanatos :ok: :smile:

PR #8022, opened 2016-03-10
xio: fix compilation against latest accelio

roidayan:
Do not use C struct initilization style to avoid eclipse errors.
We also don't use c++ struct initalize as members are not specified
and it just happens that in latest accelio version one of the fields
has changed in the middle of the struct.

### DIFF
--- a/src/msg/xio/XioMessenger.cc
+++ b/src/msg/xio/XioMessenger.cc
@@ -965,15 +965,11 @@ ConnectionRef XioMessenger::get_connection(const entity_inst_t& dest)
       << xio_uri << dendl;
 
     /* XXX client session creation parameters */
-    struct xio_session_params params = {
-      .type		= XIO_SESSION_CLIENT,
-      .initial_sn	= 0,
-      .ses_ops		= &xio_msgr_ops,
-      .user_context	= this,
-      .private_data	= NULL,
-      .private_data_len = 0,
-      .uri		= (char *)xio_uri.c_str()
-    };
+    struct xio_session_params params = {};
+    params.type         = XIO_SESSION_CLIENT;
+    params.ses_ops      = &xio_msgr_ops;
+    params.user_context = this;
+    params.uri          = xio_uri.c_str();
 
     XioConnection *xcon = new XioConnection(this, XioConnection::ACTIVE,
 					    dest);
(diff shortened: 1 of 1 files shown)


src/msg/xio/XioMessenger.cc
```diff
@@ -965,15 +965,14 @@ ConnectionRef XioMessenger::get_connection(const entity_inst_t& dest)
...
-    struct xio_session_params params = {
-      .type		= XIO_SESSION_CLIENT,
-      .initial_sn	= 0,
-      .ses_ops		= &xio_msgr_ops,
-      .user_context	= this,
-      .private_data	= NULL,
-      .private_data_len = 0,
-      .uri		= (char *)xio_uri.c_str()
-    };
+    struct xio_session_params params = {};
```
asheplyakov: Perhaps you mean
`memset(&params, 0, sizeof(params));`

src/msg/xio/XioMessenger.cc
```diff
@@ -986,16 +985,14 @@ ConnectionRef XioMessenger::get_connection(const entity_inst_t& dest)
...
-      .session           = xcon->session,
-      .ctx               = this->portals.get_portal0()->ctx,
-      .conn_idx          = 0, /* XXX auto_count */
-      .enable_tos        = 0,
-      .tos               = 0,
-      .pad               = 0,
-      .out_addr          = NULL,
-      .conn_user_context = xcon
-    };
+    struct xio_connection_params xcp = {};
```
asheplyakov: Same here

src/msg/xio/XioMessenger.cc
roidayan: assignment of {} should do the same. am i mistaken?

roidayan:
Hi,
pinging for more comments.
Thanks

roidayan:
hi @mattbenjamin, can you help with review? tia.

roidayan:
ping

cbodley:
now that the structs are zero-initialized, we can take out the extra assignments to 0/NULL. looks good otherwise

roidayan:
@cbodley did the change. thanks.

roidayan:
not sure how tags matter but maybe this should also be tagged as bug fix. we need this commit to be able to compile against latest accelio because of the pad member used her which does not exists now in accelio.

PR #8028, opened 2016-03-10
librbd: replaying a journal op post-refresh requires locking

dillaman:


dillaman:
@trociny Can you try this patch to see if it resolves your snap remove issue?

jdurgin:
lgtm

PR #8032, opened 2016-03-10
DNM: Wip split lfn -- bug 14766

athanatos:


athanatos:
Not tested yet, DNM

athanatos:
Actually fails ceph_test_objectstore, still working.

PR #8038, opened 2016-03-11
crushtool: Don't crash when called on a file that isn't a crushmap

badone:
Currently we crash due to an uncaught exception when we fail to decode a crushmap.

Catch the exception and exit gracefully.

Fixes: #8286


tchaikov:
@badone can you add a test in src/test/cli/crushtool ?

badone:
@tchaikov done

tchaikov:
lgtm

PR #8039, opened 2016-03-11
osd/ReplicatedPG: be more careful about calling publish_stats_to_osd()

gregsfortytwo:
We had moved the call out of eval_repop into a lambda, but that left out
a few other code paths and is fragile. So just call it unconditionally in
eval_repop() instead.

Fixes: #14962

### DIFF
--- a/src/osd/ReplicatedPG.cc
+++ b/src/osd/ReplicatedPG.cc
@@ -3010,8 +3010,6 @@ void ReplicatedPG::execute_ctx(OpContext *ctx)
 	log_op_stats(
 	  ctx);
 
-      publish_stats_to_osd();
-
       if (m && m->wants_ondisk() && !ctx->sent_disk) {
 	// send commit.
 	MOSDOpReply *reply = ctx->reply;
@@ -8355,6 +8353,7 @@ void ReplicatedPG::eval_repop(RepGather *repop)
   if (repop->all_applied && repop->all_committed) {
     repop->rep_done = true;
 
+    publish_stats_to_osd();
     calc_min_last_complete_ondisk();
 
     for (auto p = repop->on_success.begin();


gregsfortytwo:
@athanatos I'll test this through the FS suite but can you check if my reasoning makes any sense?
http://tracker.ceph.com/issues/14962

src/osd/ReplicatedPG.cc
```diff
@@ -8303,6 +8301,7 @@ void ReplicatedPG::eval_repop(RepGather *repop)
 	 repop->on_committed.erase(p++)) {
       (*p)();
     }
+    publish_stats_to_osd();
```
liewegas: I think this should go a bit further down, here https://github.com/ceph/ceph/pull/8039/files#diff-72747d40a424e7b5404366b557ff12a3R8357 next to calc_min_last_complete_ondisk().  We could also do it as an on_complete lambda, but I think it makes sense to do this unconditionally.

gregsfortytwo: Uh, maybe? This is where it was located previously, and immediately follows where it would be called as part of an on_complete lambda, so I put it back as that's the minimal change and I'm unsure if there are any side effects to be worried about. (Though I definitely don't see anything that makes me think publish_stats_to_osd() is sensitive to much of anything.)

athanatos: Yeah, next to calc_min_last_complete_ondisk would be better.

gregsfortytwo:
And @yehudasa reports his cluster does not crash with this branch included. Hurray!

gregsfortytwo:
This passed my cephfs testing: http://pulpito.ceph.com/gregf-2016-03-15_04:07:43-fs-greg-fs-testing-314---basic-mira/

gregsfortytwo:
@athanatos @liewegas can we merge this now? :)

PR #8043, opened 2016-03-11
DNM: rgw: get usage rest api now return capacity used by each bucket

chenji-kael:
It`s very useful to let users know how much capacity
they have used in each bucket, Venders also need this
function to charge fee by per bucket
Signed-off-by: Ji Chen insomnia@139.com

### DIFF
--- a/src/rgw/rgw_auth_s3.cc
+++ b/src/rgw/rgw_auth_s3.cc
@@ -29,8 +29,6 @@ static const char *signed_subresources[] = {
   "torrent",
   "uploadId",
   "uploads",
-  "start-date",
-  "end-date",
   "versionId",
   "versioning",
   "versions",
--- a/src/rgw/rgw_rados.cc
+++ b/src/rgw/rgw_rados.cc
@@ -12228,6 +12228,31 @@ int RGWRados::cls_user_sync_bucket_stats(rgw_obj& user_obj, rgw_bucket& bucket)
   return 0;
 }
 
+int RGWRados::cls_user_get_bucket_stats(rgw_bucket& bucket, cls_user_bucket_entry& entry)
+{
+  map<string, struct rgw_bucket_dir_header> headers;
+  int r = cls_bucket_head(bucket, RGW_NO_SHARD, headers);
+  if (r < 0) {
+    ldout(cct, 20) << "cls_bucket_header() returned " << r << dendl;
+    return r;
+  }
+  
+  bucket.convert(&entry.bucket);
+
+  map<string, struct rgw_bucket_dir_header>::iterator hiter = headers.begin();
+  for (; hiter != headers.end(); ++hiter) {
+    map<uint8_t, struct rgw_bucket_category_stats>::iterator iter = hiter->second.stats.begin();
+    for (; iter != hiter->second.stats.end(); ++iter) {
+      struct rgw_bucket_category_stats& header_stats = iter->second;
+      entry.size += header_stats.total_size;
+      entry.size_rounded += header_stats.total_size_rounded;
+      entry.count += header_stats.num_entries;
+    }
+  }
+
+  return 0;
+}
+
 int RGWRados::update_user_bucket_stats(const string& user_id, rgw_bucket& bucket, RGWStorageStats& stats)
 {
   cls_user_bucket_entry entry;
--- a/src/rgw/rgw_rest_s3.cc
+++ b/src/rgw/rgw_rest_s3.cc
@@ -368,6 +368,15 @@ static void dump_usage_categories_info(Formatter *formatter, const rgw_usage_log
   formatter->close_section(); // Category
 }
 
+static void dump_usage_bucket_info(Formatter *formatter, const std::string& name, const cls_user_bucket_entry& entry)
+{
+  formatter->open_object_section("Entry");
+  formatter->dump_string("Bucket", name);
+  formatter->dump_int("Bytes", entry.size);
+  formatter->dump_int("Bytes_Rounded", entry.size_rounded);
+  formatter->close_section(); // entry
+}
+
 void RGWGetUsage_ObjStore_S3::send_response()
 {
   if (op_ret < 0)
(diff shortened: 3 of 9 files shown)


mattbenjamin:
@yehudasa prospectively do you see any issue with this?

oritwas:
@chenji-kael , can you open a tracker issue for this?

chenji-kael:
@oritwas 
OK, FEATURE #16191 created

oritwas:
@chenji-kael , can you add it to commit message:
Fixes: url of the tracker issue

chenji-kael:
@oritwas done

src/rgw/rgw_auth_s3.cc
```diff
@@ -29,8 +29,6 @@ static const char *signed_subresources[] = {
   "torrent",
   "uploadId",
   "uploads",
-  "start-date",
```
oritwas: I think you remove those by mistake?

src/rgw/rgw_rados.cc
```diff
@@ -11288,6 +11288,31 @@ int RGWRados::cls_user_sync_bucket_stats(rgw_obj& user_obj, rgw_bucket& bucket)
...
+{
+  map<string, struct rgw_bucket_dir_header> headers;
+  int r = cls_bucket_head(bucket, RGW_NO_SHARD, headers);
+  if (r < 0) {
+    ldout(cct, 20) << "cls_bucket_header() returned " << r << dendl;
+    return r;
+  }
+  
+  bucket.convert(&entry.bucket);
+
```
oritwas: it will be safer if you zeroed the entry size, size_rounded and count fields

src/rgw/rgw_rest_s3.cc
```diff
@@ -408,14 +417,28 @@ void RGWGetUsage_ObjStore_S3::send_response()
        formatter->dump_int("SuccessfulOps", total_usage.successful_ops);
        formatter->close_section(); // total 
        formatter->close_section(); // user
-     } 
-     formatter->dump_int("TotalBytes", header.stats.total_bytes);
```
oritwas: I would keep those in the summary section

chenji-kael:
@oritwas ok

src/rgw/rgw_auth_s3.cc
chenji-kael: @oritwas 
no , start-date and end-date is only parameters， should not be treated as sub-resource ， I introduced this mistake at last commit, so I think it need wipe out.
In fact I think usage should really be put here, what`s your opinion?

chenji-kael:
@oritwas I have update this pr, please check it

src/rgw/rgw_rados.cc
```diff
@@ -11288,6 +11288,31 @@ int RGWRados::cls_user_sync_bucket_stats(rgw_obj& user_obj, rgw_bucket& bucket)
...
+  map<string, struct rgw_bucket_dir_header>::iterator hiter = headers.begin();
+  for (; hiter != headers.end(); ++hiter) {
```
theanalyst: can we use C++11 syle auto for the for loops here

src/rgw/rgw_user.cc
```diff
@@ -87,6 +87,44 @@ int rgw_user_sync_all_stats(RGWRados *store, const rgw_user& user_id)
...
+  do {
+    RGWUserBuckets user_buckets;
+    ret = rgw_read_user_buckets(store, user_id, user_buckets, marker,
+				string(), max_entries, false, &is_truncated);
+    if (ret < 0) {
+      ldout(cct, 0) << "failed to read user buckets: ret=" << ret << dendl;
+      return ret;
+    }
+    map<string, RGWBucketEnt>& buckets = user_buckets.get_buckets();
+    for (map<string, RGWBucketEnt>::iterator i = buckets.begin();
```
theanalyst: see earlier comment

src/rgw/rgw_rest_s3.cc
chenji-kael: @oritwas 
IMHO, summary is a sum of all data search from usage log. and it represent process datas
but "TotalBytes" "TotalBytesRounded" "TotalEntries" are current capacity used by user, It`s a finally result
So I think put them in "CapacityUsed" is more suitable

theanalyst: I'm not sure whether changing the json response here is a good idea, there might be other consumers who are built on top the rest api. For eg. the openstack ceilometer rgw plugin iirc used the "Summary" bits for quite sometime.. not sure what the state is right now

chenji-kael: @theanalyst 
If really someone else used this api ，I agree to keep it as before。
I will change it back

theanalyst: I'm not sure completley, but iirc the first ceilometer rgw plugin did use the usage api, so might be worth worth asking if there are other consumers of this api as well

yehudasa:
@theanalyst @oritwas is this one ready?

oritwas:
it looks like the comments were not fixed.

oritwas:
also it will be good to join https://github.com/ceph/ceph/pull/9495 with this pull request

chenji-kael:
@oritwas @yehudasa 
in fact, I do not know how to treat this PR now due to  #9495. and most comments here have been fixed except using some C++ 11 feature

yehudasa:
@oritwas @theanalyst @chenji-kael what is the status of this PR?

oritwas:
@chenji-kael, please rebase

src/rgw/rgw_rados.cc
oritwas: you need to zero entry values (see my previous comment)

chenji-kael:
@oritwas rebase done

yehudasa:
@oritwas ping

cbodley:
compression support was added to radosgw recently, and `struct rgw_bucket_category_stats` now includes an `actual_size` field that represents the compressed size in rados. we might consider exposing that information also

src/rgw/rgw_user.cc
```diff
@@ -83,6 +83,44 @@ int rgw_user_sync_all_stats(RGWRados *store, const rgw_user& user_id)
...
+      marker = i->first;
+
+      RGWBucketEnt& bucket_ent = i->second;
+      cls_user_bucket_entry entry;
+      ret = store->cls_user_get_bucket_stats(bucket_ent.bucket, entry);
+      if (ret < 0) {
+        ldout(cct, 0) << "ERROR: could not get bucket stats: ret=" << ret << dendl;
+        return ret;
+      }
+      buckets_usage_map.insert(pair<string, cls_user_bucket_entry>(bucket_ent.bucket.name, entry));
```
cbodley: `buckets_usage_map.emplace(bucket_ent.bucket.name, entry);` should work here

oritwas:
jenkins test this please

dachary:
jenkins test this please (asok)

liewegas:
@oritwas what should we do with this pr?

oritwas:
The changes needed are small.
I will create a new pr based on this one and add the needed changes.

oritwas:
opened #14287 with the fixes

PR #8044, opened 2016-03-11
osd: fix dump_ops_in_flight races

dzafman:
Fixes: #8885

### DIFF
--- a/src/common/TrackedOp.h
+++ b/src/common/TrackedOp.h
@@ -193,6 +193,7 @@ public:
   }
 
   double get_duration() const {
+    Mutex::Locker l(lock);
     if (!events.empty() && events.rbegin()->second.compare("done") == 0)
       return events.rbegin()->first - get_initiated();
     else
@@ -201,6 +202,7 @@ public:
 
   void mark_event(const string &event);
   virtual const char *state_string() const {
+    Mutex::Locker l(lock);
     return events.rbegin()->second.c_str();
   }
   void dump(utime_t now, Formatter *f) const;
--- a/src/common/TrackedOp.h
+++ b/src/common/TrackedOp.h
@@ -104,8 +104,8 @@ public:
     RWLock::WLocker l(lock);
     tracking_enabled = enable;
   }
-  void dump_ops_in_flight(Formatter *f, bool print_only_blocked=false);
-  void dump_historic_ops(Formatter *f);
+  bool dump_ops_in_flight(Formatter *f, bool print_only_blocked=false);
+  bool dump_historic_ops(Formatter *f);
   void register_inflight_op(xlist<TrackedOp*>::item *i);
   void unregister_inflight_op(TrackedOp *i);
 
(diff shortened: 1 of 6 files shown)


dzafman:
Haven't written an automated test case, yet.  Rebased to jewel so waiting for my own make check to verify.

src/messages/MOSDOp.h
```diff
@@ -48,8 +48,10 @@ class MOSDOp : public Message {
   pg_t pgid;
   bufferlist::iterator p;
   // Decoding flags. Decoding is only needed for messages catched by pipe reader.
-  bool partial_decode_needed;
-  bool final_decode_needed;
+  // Transition from true -> false without locks being held
+  // Can never see final_decode_needed == false and partial_decode_needed == true
+  volatile bool partial_decode_needed;
+  volatile bool final_decode_needed;
```
gregsfortytwo: I'm just looking up the volatile keyword here, but I'm not sure this works, strictly speaking. http://en.cppreference.com/w/cpp/language/cv and https://en.wikipedia.org/wiki/Volatile_(computer_programming)#In_C_and_C.2B.2B are both pretty clear that ```volatile``` is not "for inter-thread communication"

tchaikov: agreed, can we use `atomic<bool>` instead here?

gregsfortytwo: I think that'll get expensive; don't we dereference those bools every time we look at an MOSDOp? I *think* we're just using those as flags to determine if we need to grab a lock and do the decode before looking at memory, or if we can look at the memory directly, which means presumably with careful coding it's okay if we read a stale value (we'd just grab the lock needlessly). But apparently it's being done wrongly, or maybe it really is just that the memory reads are being reordered incorrectly (which is what ```volatile``` prevents)...

athanatos: Yeah, volatile really isn't appropriate here.  atomic<bool> would be cleanest.

src/common/TrackedOp.h
```diff
@@ -166,13 +170,15 @@ class TrackedOp {
...
+    is_registered(false)
   {
     RWLock::RLocker l(tracker->lock);
     if (tracker->tracking_enabled) {
       tracker->register_inflight_op(&xitem);
       events.push_back(make_pair(initiated_at, "initiated"));
-      is_tracked = true;
+      // Tells tracking_start() whether to set is_tracked or not
+      is_registered = true;
     }
```
athanatos: Hmm, this does solve the problem, but I think the better solution would be to move this code into a TrackedOp::register method and call it in OpTracker::create_request.  That way, we never have a half-constructed object registered in the first place and can avoid having both is_tracked and is_registered.

athanatos: That is, we can have just is_tracked rather than both.

dzafman: I have the fix for this by moving that constructor code into tracking_start() so that is_registered isn't needed.

athanatos:
lgtm

liewegas:
passed my rados run

PR #8046, opened 2016-03-11
msg: async: fix perf counter description and simplify _send_keepalive_or_ack

xiexingguo:


yuyuyu101:
http://pulpito.ceph.com/haomai-2016-03-12_12:34:10-rados-wip-haomai-testing-distro-basic-multi/

PR #8048, opened 2016-03-11
rgw: cleanup of debug messages

oritwas:
rgw: use ldout instead of derr and lderr and increase logging level


yehudasa:
@oritwas PR #8058 rebases your work here

PR #8049, opened 2016-03-11
hammer: keyring permisions for mon deamon

xiaoxichen:
http://tracker.ceph.com/issues/15022


dachary:
It passed the hammer upgrade suite ( http://tracker.ceph.com/issues/14692#note-52 ) and the rados suite http://tracker.ceph.com/issues/14692#note-51).

PR #8050, opened 2016-03-11
os/filestore: fix punch hole usage in _zero

liewegas:
If we punch a hole that extends past EOF, ObjectStore semantics are
that the file size is also extended.  Do that.

Note that this bug was hidden before because we weren't passing
KEEP_SIZE to fallocate until 7bd95b595fddb8a4e618a2c7df1ba04eccf0829d
and the fallocate was *always* failing with EOPNOTSUPP (making us fall
back to writing actual zeros).

### DIFF
--- a/src/os/filestore/FileStore.cc
+++ b/src/os/filestore/FileStore.cc
@@ -3249,10 +3249,29 @@ int FileStore::_zero(const coll_t& cid, const ghobject_t& oid, uint64_t offset, 
     goto out;
   }
 
+  struct stat st;
+  ret = ::fstat(**fd, &st);
+  if (ret < 0) {
+    ret = -errno;
+    lfn_close(fd);
+    goto out;
+  }
+
   // first try fallocate
   ret = fallocate(**fd, FALLOC_FL_KEEP_SIZE | FALLOC_FL_PUNCH_HOLE, offset, len);
-  if (ret < 0)
+  if (ret < 0) {
     ret = -errno;
+  } else {
+    // ensure we extent file size, if needed
+    if (offset + len > st.st_size) {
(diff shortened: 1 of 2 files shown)


src/os/filestore/FileStore.cc
```diff
@@ -3239,38 +3239,60 @@ int FileStore::_zero(const coll_t& cid, const ghobject_t& oid, uint64_t offset,
...
-  }
+    // first try fallocate
+    ret = fallocate(**fd, FALLOC_FL_KEEP_SIZE | FALLOC_FL_PUNCH_HOLE,
+		    offset, len);
+    if (ret < 0) {
+      ret = -errno;
+    } else {
+      // ensure we extent file size, if needed
+      if (offset + len > st.st_size) {
+	ret = ::ftruncate(**fd, offset + len);
```
yuyuyu101: I'm a little nervous about this, what if we want to zero 10->5 and size is 12. from my glancing view, 12-15 will be zeroed, but what about 10-12 section(not zeroed)?

Another thought: it looks like PUNCH_HOLE isn't fullly match _zero semantic, FALLOC_FL_ZERO_RANGE may better

liewegas: The fallocate call should zero 10-12.  Truncate just ensure we end up with 
size=15 at the end.

I think PUNCH_HOLE is a closer fit.  Both zero teh region logically, but 
ZERO_RANGE preallocates, and I don't think we want that.

athanatos:
lgtm

src/os/filestore/FileStore.cc
yuyuyu101: oh, I have a misunderstand on this bug.

PR #8053, opened 2016-03-11
rgw: improve memory management in AWS4 auth

rzarzynski:


zaitcev:
The try-catch is one thing, but why is it necessary to apply unique_ptr? Where is the leak? If you know a scenario, place put it into the commit message. I, for one, cannot readily see where the pointer is being improperly copied.

rzarzynski:
> why is it necessary to apply unique_ptr? Where is the leak?

I **never** said we have any leak there. Switch to ```std::unique_ptr``` is purely for code maintainability. It's a very clear measure to mark who is responsible for controlling the part's life and makes life easier. Lowering risk of having dangling pointers and double free could be seen as an example of that. 

After move to C++11 smart pointers became quite popular across the code.

zaitcev:
> I never said we have any leak there. Switch to std::unique_ptr is purely for code maintainability.
Sounds fine to me, then. I just wish the commit message contained this reasoning. For example, "This improves maintainability."

rzarzynski:
I can fully agree that being explicit is a good thing. I've just pushed the commit with message amended according to your comment.

zaitcev:
Looks fine to me.

PR #8054, opened 2016-03-11
common: SubProcess: Avoid buffer corruption when calling err()

ErwanAliasr1:
Some code like crush/CrushTester.cc, uses err() to get the output status of a spawned command.
    
As per reported in bug #15011, some tests were failing because of some utf8
decoding issues. The implied buffers were generated by the output of err()
function of SubProcess.
    
err() is returning the str().c_str() value of 'errstr' object which is not living much longer that the function itself.
    
This patch returns a std::string which avoid the temporary allocation.
    
Since this commit, the make check is now passing everytime.
    
 Fixes: #15011

### DIFF
--- a/src/test/test_subprocess.cc
+++ b/src/test/test_subprocess.cc
@@ -38,7 +38,7 @@ TEST(SubProcess, True)
   SubProcess p("true");
   ASSERT_EQ(p.spawn(), 0);
   ASSERT_EQ(p.join(), 0);
-  ASSERT_TRUE(p.err()[0] == '\0');
+  ASSERT_TRUE(p.err().c_str()[0] == '\0');
 }
 
 TEST(SubProcess, False)
@@ -46,7 +46,7 @@ TEST(SubProcess, False)
   SubProcess p("false");
   ASSERT_EQ(p.spawn(), 0);
   ASSERT_EQ(p.join(), 1);
-  ASSERT_FALSE(p.err()[0] == '\0');
+  ASSERT_FALSE(p.err().c_str()[0] == '\0');
 }
 
 TEST(SubProcess, NotFound)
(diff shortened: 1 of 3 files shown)


liewegas:
I don't think we want to print errors to stderr in the python bindings... @jdurgin ?

ErwanAliasr1:
The real meaning was being able to determine what command failed. As its ran inside a Python Thread,  I found not really obvious the way to get it.

I don't have any particular affinity with this commit as long as we can get that information easily. That's a side commit that I can move to another PR if it become too much controversal.

jdurgin:
Yeah, libraries like rados.pyx shouldn't write to stderr like that. I think in this case we should rely on ceph's logging system in librados, and add any extra debugging we may need there.

The 2nd commit looks good, just need to fix up any other err() callers (test/librbd/fsx.cc at least)

ErwanAliasr1:
Ok, I'll work on that next week and update the PR accordingly. Thanks.

ErwanAliasr1:
Removed the 1st commit, will rework it differently.
Applied the changes as requested.
Enjoy,

jdurgin:
lgtm, thanks!

ErwanAliasr1:
Thx for the merge

src/test/test_subprocess.cc
```diff
@@ -46,7 +46,7 @@ TEST(SubProcess, False)
   SubProcess p("false");
   ASSERT_EQ(p.spawn(), 0);
   ASSERT_EQ(p.join(), 1);
-  ASSERT_FALSE(p.err()[0] == '\0');
+  ASSERT_FALSE(p.err().c_str()[0] == '\0');
```
trociny: Nitpicking, I would use `ASSERT_TRUE(p.err().empty())` instead.

PR #8059, opened 2016-03-11
rados: add ceph:: namespace to bufferlist type

noahdesu:
After a fresh install, applications that had been including rados.hpp cannot resolve bufferlist, which is in ceph:: scope.

Signed-off-by: Noah Watkins <noahwatkins@gmail.com>


liewegas:
lgtm, but want to build test this first on all the distros

jdurgin:
we should really have a test for this (added http://tracker.ceph.com/issues/15100)

PR #8060, opened 2016-03-11
Wip rgw highres 3

mattbenjamin:
Depends on the librados stat2() change.

There are unfixed warnings marking places where new header parse/emit may be needed.  Will fix after merge.


yehudasa:
closing it in favour of #8108

PR #8061, opened 2016-03-11
SSL, Civetweb, and STS

adamemerson:
This pull request includes super awesome security stuff. Like TLS. And STS.


CMakeLists.txt
```diff
@@ -156,6 +157,8 @@ if(WITH_SPDK)
   set(HAVE_SPDK TRUE)
 endif(WITH_SPDK)
 
+CHECK_LIBRARY_EXISTS (curl curl_multi_wait "" HAVE_CURL_MULTI_WAIT)
```
cbodley: cmake/modules/Findlibcurl.cmake already has a `CHECK_SYMBOL_EXISTS(curl_multi_wait "curl.h" HAVE_CURL_MULTI_WAIT)`. was it not working for you?

src/rgw/rgw_op.cc
```diff
@@ -658,6 +659,15 @@ bool RGWOp::generate_cors_headers(string& origin, string& method, string& header
   return true;
 }
 
+/**
+ * Return a callable that can invoke dump_access_control().
+ */
+
+boost::function<void()> RGWOp::dump_access_control_f()
+{
+  return boost::bind(dump_access_control, s, this);
```
cbodley: why boost over `std::function` and `std::bind`?

src/sts/rgw_rest_sts.h
```diff
@@ -0,0 +1,365 @@
...
+class assume_role_with_saml_request {
+public:
+  string arn;
+  string policy;
+  int duration_seconds;
+  string principal_arn;
+  string saml_assertion;
+};
+
+#if 0
```
cbodley: are these `#if 0` blocks likely to be needed?

PR #8062, opened 2016-03-11
ceph-detect-init: return systemd on Debian Jessie

smithfarm:
http://tracker.ceph.com/issues/15086 Fixes: #15086

### DIFF
--- a/src/ceph-detect-init/ceph_detect_init/debian/__init__.py
+++ b/src/ceph-detect-init/ceph_detect_init/debian/__init__.py
@@ -14,4 +14,9 @@ def choose_init():
             return 'systemd'
         else:
             return 'upstart'
+    if distro.lower() in ('debian'):
+        if codename in ('squeeze', 'wheezy'):
+            return 'sysvinit'
+        else:
+            return 'systemd'
     return 'sysvinit'


smithfarm:
Tested in a Jessie container. Without this patch, ceph-detect-init returns "sysvinit". With the patch, it returns "systemd".

src/ceph-detect-init/ceph_detect_init/debian/__init__.py
```diff
@@ -14,4 +14,7 @@ def choose_init():
             return 'systemd'
         else:
             return 'upstart'
+    if distro.lower() in ('debian'):
+        if codename >= 'jessie':
```
liewegas: the debian version names aren't sorted... i think we need to enumerate the "old" distros and map them to sysvinit, and then default to systemd for others.

i think that is just squeeze and wheezy -> sysvinit.. anything older won't work anyway.

smithfarm:
@liewegas OK. By the way, the old distros are already enumerated - read on.

In the Jessie container, `codename` evaluates to an empty string:

```
>>> import platform
>>> print platform.linux_distribution()
('debian', '8.3', '')
```

The code takes this into account and enumerates here: https://github.com/ceph/ceph/blob/master/src/ceph-detect-init/ceph_detect_init/__init__.py#L87-L92

smithfarm:
Tested on Jessie only so far.

liewegas:
Thanks, looks good to me!

smithfarm:
Should I retarget this to master?

liewegas:
merged manually

PR #8065, opened 2016-03-11
MDSMap -> FSMap decoding fixes

gregsfortytwo:
Fix a failure when upgrading monitor clusters that have *no* filesystem, not just those which do have an FS. And make "fs new" idempotent again.


gregsfortytwo:
Okay, Yehuda had a vstart cluster which was configured *without* a filesystem and which was still crashing on startup with my previous round of fixes...but now I can't even reproduce that.

Nonetheless, I believe these patches are what we want — certainly the change to make "fs new" idempotent is needed!

gregsfortytwo:
And @yehudasa reports his cluster does not crash with this branch included. Hurray!

gregsfortytwo:
Closed in favor of the above.

PR #8066, opened 2016-03-12
Wip rgw swift versioning disabled

yehudasa:
(partial) support for swift object versioning. The feature is disabled by default. The idea is to have to data structure changes in now, so that it make feature backport feasible later.


jdurgin:
seems ok to me

yehudasa:
@jdurgin danke

PR #8069, opened 2016-03-12
Adding documentation on how to use new dynamic throttle scheme

somnathr:

### DIFF
--- a/src/doc/dynamic-throttle.txt
+++ b/src/doc/dynamic-throttle.txt
@@ -1,0 +1,127 @@
+------
+TOPIC:
+------
+Dynamic backoff throttle is been introduced in the Jewel timeframe to produce
+a stable and improved performance out from filestore. This should also improve 
+the average and 99th latency significantly.
+
+-----------
+WHAT IS IT?
+-----------
+The old throttle scheme in the filestore is to allow all the ios till the 
+outstanding io/bytes reached some limit. Once crossed the threshold, it will not
+allow any more io to go through till the outstanding ios/bytes goes below the 
+threshold. That's why once it is crossed the threshold, the write behavior 
+becomes very spiky. 
+The idea of the new throttle is, based on some user configurable parameters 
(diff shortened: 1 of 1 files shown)


src/doc/dynamic-throttle.txt
```diff
@@ -0,0 +1,129 @@
...
+
+  filestore_queue_max_bytes and filestore_queue_max_ops:
+  -----------------------------------------------------
+
+  This directly depends on how filestore op_wq will be growing in the memory. So,
+  user needs to calculate how much memory he can give to each OSD. Bigger value
+  meaning current/max ratio will be smaller and throttle will be relaxed a bit.
+  This could help in case of 100% write but in case of mixed read/write this can
+  induce more latencies for reads on the objects those are not yet been applied 
+  i.e yet to reach in it's final filesystem location.
```
athanatos: You should mention that once you reach max throughput, increasing these further only increases memory usage and latency (bad).

src/doc/dynamic-throttle.txt
```diff
@@ -0,0 +1,129 @@
...
+  This is the delay tunables and one should do all effort here so that the current
+  doesn't reach to the max. For HDD and smaller block size, the old throttle may
+  make sense, in that case we should set these delay multiple to 0.
+
+It is recommended that user should use ceph_smalliobench tool to see the effect 
+of these parameters (and decide) on your HW before deploying OSDs. 
+Following scripts is a wrapper around small_io_bench and shows a nice latency 
+distribution.
+
+  https://github.com/athanatos/ceph-perf-scripts
```
athanatos: I really don't want this in the docs, it would be very confusing to users.  I'd rather you explained how to use the relevant smalliobenchfs options directly.  If you want, you can factor out the code for summarizing the smalliobenchfs output and put it into ceph.git as another file in another PR.

somnathr:
@athanatos sorry for the delay, I have addressed your comments..

PR #8070, opened 2016-03-12
time: should use signed type for time_point::duration::rep

tchaikov:
otherwise the timespan between two `Clock::time_point`s could be
incredibly large due to unsigned integer overflow. and da0f660
covered this bug up in a wrong way.

### DIFF
--- a/src/common/ceph_time.h
+++ b/src/common/ceph_time.h
@@ -63,7 +63,7 @@ namespace ceph {
       typedef duration::period period;
       // The second template parameter defaults to the clock's duration
       // type.
-      typedef std::chrono::time_point<real_clock> time_point;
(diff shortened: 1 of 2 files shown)


adamemerson:
We have signespan for that purpose.

tchaikov:
@adamemerson i see. and in this patch, i am suggesting to use `signedspan` as the `duration` type for the `clock_point` types.

tchaikov:
changelog:

- use signed type for time_point::duration::rep (was use signed type for time_point::rep, which is wrong, as pointed out by Adam)

tchaikov:
@adamemerson i fixed my PR. could you take a look again?

tchaikov:
changelog
- rebased against master

PR #8071, opened 2016-03-12
osd: initialize last_recalibrate field at construction

xiexingguo:
So we don't get an overflowed duration on the first entry of
promote_throttle_recalibrate() call.


xiexingguo:
@liewegas Not very sure of this.

liewegas:
lgtm

PR #8078, opened 2016-03-14
msg: async: time_events may be lost after correcting the system clock.

Yan-waller:
list changed[utime_t()] will be overwritten sinces utime_t() is allways zero.


yuyuyu101:
already fixed in master

PR #8079, opened 2016-03-14
[DNM] rbd-mirror: integrate with image sync state machine

dillaman:


dillaman:
@jdurgin This is only DNM since it includes the very large PR #7979

jdurgin:
nice decomposition of sync into small state machines - makes it much easier to read. merged manually

PR #8080, opened 2016-03-14
osdc: Fix bug "New added osd always down when full"

sglwlb:
When full flag is set in osdmap, Objecter will call
_maybe_request_map to subscribe newer map. However, it will
cause new added osd has no change to start boot and stay at
down state. This patch fix this problem.

Signed-off-by: Wu Libin hzwulibin@gmail.com

### DIFF
--- a/src/osdc/Objecter.cc
+++ b/src/osdc/Objecter.cc
@@ -1109,6 +1109,7 @@ void Objecter::handle_osd_map(MOSDMap *m)
   bool cluster_full = _osdmap_full_flag();
   bool was_pausewr = osdmap->test_flag(CEPH_OSDMAP_PAUSEWR) || cluster_full ||
     _osdmap_has_pool_full();
+  bool first_map = false;
   map<int64_t, bool> pool_full_map;
   for (map<int64_t, pg_pool_t>::const_iterator it
 	 = osdmap->get_pools().begin();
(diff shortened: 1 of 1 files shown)


sglwlb:
Here is the problem i described:
http://tracker.ceph.com/issues/15025

smithfarm:
@sglwlb Does this patch fix http://tracker.ceph.com/issues/15025 ? If so, please add a line like this to the commit message:

```
Fixes: #15025
```

This helps with later debugging/forensics.

sglwlb:
@smithfarm updated.

liewegas:
I'm unable to reproduce this on current master:

maetl:src (wip-mon-rocksdb) 10:10 PM $ ./ceph -s | grep full
            3 full osd(s)
            flags full,sortbitwise

....
./init-ceph restart osd.0
...

maetl:src (wip-mon-rocksdb) 10:11 PM $ ./ceph -s | grep osds
     osdmap e20: 3 osds: 3 up, 3 in

I think we need a test for this, something like one of the scripts at test/osd/*.sh.  If this only affects older releases (hammer), then we can backpor the test and verify the fix there.

liewegas:
@sglwlb ping?

dachary:

tchaikov:
@sglwlb  ping?

liewegas:
Please reopen if this is still a bug!

PR #8084, opened 2016-03-14
rgw: accept data only at the first time in response to a request

sunfch:
Because the member "max_response" in class "RGWRESTSimpleRequest" is
initialized 0, data packages cannot be accepted when the function
"receive_data()" is called again.

This patch initialize the value of "max_response" to "content-length".

Signed-off-by: sunspot <sunspot0105@gmail.com>

### DIFF
--- a/src/rgw/rgw_rest_client.cc
+++ b/src/rgw/rgw_rest_client.cc
@@ -22,6 +22,22 @@ int RGWRESTSimpleRequest::get_status()
   return status;
 }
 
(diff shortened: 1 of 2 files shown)


cbodley:
can you help me understand why receive_data() is being called multiple times? I haven't seen this one in testing

sunfch:
@cbodley 
when the content length of http body data are larger than 1500 bytes, 
the receive_data() function could be called multiple times because of the limit of Ethernet MTU. 
The content length of your http header could be less than 1500 bytes.

PR #8085, opened 2016-03-14
os/bluestore/KernelDevice: silence compilation #warnings

atheism:


liewegas:
the rgw type needs to be uint64_t since it's encoded.
the warning was fixed in another pr.

thanks!

PR #8089, opened 2016-03-14
mon/PGMonitor: reliably mark PGs state

liewegas:
In beb99d708a5051405a32f46fc10cef473bfddce3 we partially addressed
the case where a acting_primary update raced with an attemp to
mark the pg state.  However, if the acting_primary changed, we
didn't check the new acting_primary to ensure that it is also
down and we should still mark things stale.

Also, printing out the acting_primary aids debugging.

### DIFF
--- a/src/mon/PGMonitor.cc
+++ b/src/mon/PGMonitor.cc
@@ -1281,9 +1281,11 @@ epoch_t PGMonitor::send_pg_creates(int osd, Connection *con, epoch_t next)
   return last + 1;
 }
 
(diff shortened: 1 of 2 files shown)


liewegas:
@tchaikov mind taking a look?

liewegas:
I'm trying to track down a bug where the mon marked every pg stale but I can't figure out why.. the extra debugging here will help a bit.

tchaikov:
lgtm

PR #8090, opened 2016-03-14
test: fixup and improvements for rbd-mirror test

trociny:

### DIFF
--- a/qa/workunits/rbd/rbd_mirror.sh
+++ b/qa/workunits/rbd/rbd_mirror.sh
@@ -1,4 +1,53 @@
 #!/bin/sh
+#
+# rbd_mirror.sh - test rbd-mirror daemon
+#
+# The scripts starts two ("local" and "remote") clusters using mstart.sh script,
+# creates a temporary directory, used for cluster configs, daemon logs, admin
+# socket, temporary files, and launches rbd-mirror daemon.
+#
+# There are several env variables useful when troubleshooting a test failure:
+#
+#  RBD_MIRROR_NOCLEANUP - if not empty, don't run the cleanup (stop processes,
+#                         destroy the clusters and remove the temp directory)
+#                         on exit, so it is possible to check the test state
+#                         after failure.
+#  RBD_MIRROR_TEMDIR    - use this path when creating the temporary directory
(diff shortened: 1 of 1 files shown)


trociny:
To troubleshoot a test failure I use the following commands:

From src dir start the test in NOCLEANUP mode (so it possible to check its state after failure) and with TEMPDIR pointing to known location):
```
RBD_MIRROR_NOCLEANUP=1 RBD_MIRROR_TEMDIR=/tmp/tmp.rbd_mirror ../qa/workunits/rbd/rbd_mirror.sh
```
After failure cd to TEMPDIR and check status:
```
cd /tmp/tmp.rbd_mirror
less rbd-mirror.log
rbd --cluster remote -p mirror journal status --image test
ceph --admin-daemon rbd-mirror.asok help
...
```
Run cleanup:
```
cd - #SRCDIR
RBD_MIRROR_TEMDIR=/tmp/tmp.rbd_mirror ../qa/workunits/rbd/rbd_mirror.sh cleanup
```

jdurgin:
Could you add that comment on how to use it to the script (perhaps a comment on using the script at the top of the file)?

qa/workunits/rbd/rbd_mirror.sh
```diff
@@ -113,26 +118,50 @@ flush()
...
 }
 
+wait_for_image_replay_started()
+{
+    local image=$1
+
+    wait_for_image_replay_state ${image} started
+}
+
+wait_for_image_replay_stopped()
```
jdurgin: looks like this isn't used yet

trociny: Yes, I was going to use this to test live image replayer removal after an image is removed (journaling is disabled). But right now I can't remove the image when it is replayed due to "Device or resource busy" error. Still I think it is desirable feature and we will be able to do this in some future, and this function will be useful then.

jdurgin: makes sense to leave it in then. I think Jason fixed that EBUSY error with his aio_notify patch that's now in master.

trociny:
Rebased after adding "how to use" into the scrip.

PR #8091, opened 2016-03-14
rgw: Remote the new rgw_swift_account_in_url option

zaitcev:
A recent commit introduced rgw_swift_account_in_url, which appears
to be not necessary. This patch uses token format to pass the
information that current code extracts from the option. This way
a smooth upgrades without undue configuration are possible, and
the administrative load is not raised. No need to document what
software can do automatically either.

When I presented this originally in one of community BJ meetings,
it may be that Radoslaw pointed out that adjustments to TempURL
were needed. But I cannot recall if it were true and if yes, what was
the problem. I may be missing something though. This commit permits
to find it out and document the issue, if any.

### DIFF
--- a/src/common/config_opts.h
+++ b/src/common/config_opts.h
@@ -1221,7 +1221,6 @@ OPTION(rgw_swift_url_prefix, OPT_STR, "swift") // entry point for which a url is
 OPTION(rgw_swift_auth_url, OPT_STR, "")        // default URL to go and verify tokens for v1 auth (if not using internal swift auth)
 OPTION(rgw_swift_auth_entry, OPT_STR, "auth")  // entry point for which a url is considered a swift auth url
 OPTION(rgw_swift_tenant_name, OPT_STR, "")  // tenant name to use for swift access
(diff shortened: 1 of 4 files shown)


rzarzynski:
@zaitcev

Hi Pete.

I'm afraid that the solution is only partial. Yes, in some cases you are perfectly able to smuggle the information whether URL contains the account part or not. That's the responsibility of introduced ```AUTH_rgwts``` if my understanding is correct.

Unfortunately, there are use cases where you don't have such possibility. One of them is TempURL - client doesn't provide token but a signature (HMAC basing on a key set on account) that is generated outside of RGW. This is only about the TempURL but other authentication mechanisms might be affected as well. I mean here especially situations when you want to operate through Swift API on someone else's account you have permissions to (ACLs).

Regards,
Radek

oritwas:
@zaitcev , what is the status of this pr?

zaitcev:
I think Radoslaw is likely wrong about this. TempURL can be fixed with a known prefix, as far as I know. But it's valid that all such issues have to be enumerated and proven, and I never can get around to doing it.

PR #8092, opened 2016-03-14
rgw: guard memory alloc and fix error handling in RGWBulkDelete

rzarzynski:


src/rgw/rgw_op.cc
```diff
@@ -4473,18 +4471,25 @@ void RGWBulkDelete::pre_exec()
...
 void RGWBulkDelete::execute()
 {
-  deleter = std::unique_ptr<Deleter>(new Deleter(store, s));
+  try {
+    deleter = std::unique_ptr<Deleter>(\
+          new RGWBulkDelete::Deleter(store, s));
+  } catch (std::bad_alloc) {
+    op_ret = -ENOMEM;
+    return;
+  }
```
cbodley: i'm not convinced that catching allocation failures here is a good idea. if we return -ENOMEM at this point, are we actually able to send a reply back to the client without any other allocations? this is an error path that we don't have a way to test. once we're out of memory, there's little we can do to recover gracefully so it's generally best to crash early

yehudasa: right, we don't handle in anywhere else

rzarzynski: Casey's argumentation convinced me. Although there is is `strdup` where we try to recover through `-ENOMEM`, I think it's much better to exit early. I will rework the PR soon.

mattbenjamin:
@rzarzynski do you have new code for this?  if not, can I close for now?

rzarzynski:
@mattbenjamin: I will upload the rest (like the `op_ret` fix) as a separate PR. Closing this.

PR #8093, opened 2016-03-14
rgw: Allow to serve Swift off the URL root

zaitcev:
In most public clouds OpenStack Swift URL is based on the tree
root, like:
  https://swift.domain.com/v1/AUTH_tenant

Our convention is to have a path, like:
  https://domain.com/swift/v1/AUTH_tenant

The path is configurable, but even if someone manages to set
an empty prefix, we print headers with %s/%s/v1 where the extra
slash is baked in. In addition, the empty prefix means "use the
/swift default".

To provide compatibility, this patch uses a trick. The empty
prefix remains to mean "/swift", but then we use "/" (one slash)
to create the needed empty prefix.

Signed-off-by: Pete Zaitcev <zaitcev@redhat.com>

### DIFF
--- a/doc/radosgw/config-ref.rst
+++ b/doc/radosgw/config-ref.rst
@@ -749,10 +749,14 @@ Swift Settings
 
 ``rgw swift url prefix``
 
-:Description: The URL prefix for the Swift API. 
+:Description: The URL prefix for the Swift StorageURL that goes in front of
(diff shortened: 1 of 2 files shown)


yehudasa:
@rzarzynski can you take a look at this one?

rzarzynski:
@yehudasa, @zaitcev: it looks good to me.

PR #8094, opened 2016-03-14
librbd: handle unregistering the image watcher when disconnected

dillaman:


jdurgin:
looks good - should it have an issue for backporting too?

dillaman:
@jdurgin It only affects the jewel branch

PR #8095, opened 2016-03-14
Wip rgw sync fixes

yehudasa:


src/rgw/rgw_data_sync.cc
```diff
@@ -1230,10 +1244,13 @@ class RGWDataSyncCR : public RGWCoroutine {
...
               RGWDataSyncShardControlCR *cr = new RGWDataSyncShardControlCR(sync_env, sync_env->store->get_zone_params().log_pool,
                                                         iter->first, iter->second);
+              cr->get();
               shard_crs_lock.Lock();
               shard_crs[iter->first] = cr;
               shard_crs_lock.Unlock();
-              spawn(cr, true);
+              RGWCoroutinesStack *stack = spawn(cr, true);
+              stack->get();
+              stacks[stack] = iter->first;
```
cbodley: I'm not sure what `map<RGWCoroutinesStack *, int> stacks` accomplishes here (I don't see it being accessed elsewhere), and this reference from `stack->get()` isn't ever released.

src/rgw/rgw_coroutine.h
```diff
@@ -271,6 +271,9 @@ class RGWCoroutine : public RefCountedObject, public boost::asio::coroutine {
   int wait(const utime_t& interval);
   bool drain_children(int num_cr_left); /* returns true if needed to be called again */
   void wakeup();
+  void restart() {
+    set_state(RGWCoroutine_Run);
+  }
```
cbodley: I don't see this called anywhere

src/rgw/rgw_data_sync.cc
yehudasa: Yeah, my original change was a bit more complicated, was tracking spawned crs, and then re-executing. But ended up with the current solution. Need to remove this line.

src/rgw/rgw_coroutine.h
yehudasa: yuo

yehudasa:
@cbodley pushed a fix, merging it

yehudasa:
@cbodley *fixed according to your comments (removed unneeded stuff)

PR #8097, opened 2016-03-14
crushtool: add ability of precise testing of placement group calculation.

aclamk:
Now pool id can be passed and proper JENKINS hash is used.
Signed-off-by: Adam Kupczyk akupczyk@mirantis.com

### DIFF
--- a/src/crush/CrushTester.h
+++ b/src/crush/CrushTester.h
@@ -18,6 +18,7 @@ class CrushTester {
   int ruleset;
   int min_x, max_x;
   int min_rep, max_rep;
+  int64_t pool_id;
 
   int num_batches;
   bool use_crush;
@@ -172,6 +173,7 @@ public:
       ruleset(-1),
       min_x(-1), max_x(-1),
       min_rep(-1), max_rep(-1),
+      pool_id(-1),
       num_batches(1),
       use_crush(true),
       mark_down_device_ratio(0.0),
@@ -302,6 +304,11 @@ public:
   void set_min_x(int x) {
     min_x = x;
   }
+
+  void set_pool_id(int64_t x){
+    pool_id = x;
+  }
+
   int get_min_x() const {
     return min_x;
   }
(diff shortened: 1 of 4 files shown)


src/crush/CrushTester.h
```diff
@@ -18,6 +18,7 @@ class CrushTester {
   int ruleset;
   int min_x, max_x;
   int min_rep, max_rep;
+  int pool_id;
```
liewegas: may as well use int64_t since that was pool ids are

src/crush/CrushTester.h
```diff
@@ -302,6 +304,10 @@ class CrushTester {
   void set_min_x(int x) {
     min_x = x;
   }
+
+  void set_pool_id(int x){
```
liewegas: ditto

tchaikov: @aclamk might want to update the type of the parameter as well.

tchaikov:
@aclamk you are including irrelevant commits in your pull request, could you remove them?

aclamk:
@tchaikov THANKS. I really dont know how I ended with Mykola commits here :-(

liewegas:
I think this is covered by osdmaptool --test-map-pg and --test-map-all-pgs (or whatever it is).. right?

aclamk:
Summary:
I cannot use osdmaptool in its current state

Question:
Can I modify osdmaptool to:
- allow custom ruleset
- allow custom pool id
- allow custom replication size

Reasoning:
I don't see equivalence between "osdmaptool --test-map-pg" and "crushtool --pool-id".
True, 
1) osdmaptool uses rjenkins as hash to crush, 
2) osdmaptool prints number of PGs per OSD
But:
1) osdmaptool automagically creates "pool 0 'rbd' replicated size 3 min_size 2 crush_ruleset 0 object_hash rjenkins pg_num 22528 pgp_num 22528 last_change 0 flags hashpspool stripe_width 0". I don't see any possibility to create additonal pools nor specify different ruleset or replication size.
2) you can only select only one PG with osdmaptool invocation, and it must be PG from pool 0; with crushtool pool id, ruleset and replication size can be selected

liewegas:
Fair enough!

PR #8100, opened 2016-03-14
cmake: avoid false-positive LDAP header detect

mattbenjamin:
Pre-setting a value in OPENLDAP_INCLUDE_DIR broke the subsequent
test.

@oritwas look ok?


oritwas:
looks great

PR #8101, opened 2016-03-14
os/bluestore: small fixes in bluestore StupidAllocator

jjhuo:
Main change is to avoid duplicated searches during allocating free extents.

### DIFF
--- a/src/os/bluestore/StupidAllocator.cc
+++ b/src/os/bluestore/StupidAllocator.cc
@@ -85,15 +85,15 @@ static uint64_t aligned_len(btree_interval_set<uint64_t>::iterator p,
 }
 
 int StupidAllocator::allocate(
-  uint64_t need_size, uint64_t alloc_unit, int64_t hint,
+  uint64_t want_size, uint64_t alloc_unit, int64_t hint,
   uint64_t *offset, uint32_t *length)
 {
   std::lock_guard<std::mutex> l(lock);
-  dout(10) << __func__ << " need_size " << need_size
+  dout(10) << __func__ << " want_size " << want_size
 	   << " alloc_unit " << alloc_unit
 	   << " hint " << hint
 	   << dendl;
-  uint64_t want = MAX(alloc_unit, need_size);
+  uint64_t want = MAX(alloc_unit, want_size);
   int bin = _choose_bin(want);
   int orig_bin = bin;
 
@@ -107,7 +107,7 @@ int StupidAllocator::allocate(
     for (bin = orig_bin; bin < (int)free.size(); ++bin) {
       p = free[bin].lower_bound(hint);
       while (p != free[bin].end()) {
-	if (aligned_len(p, alloc_unit) >= need_size) {
+	if (aligned_len(p, alloc_unit) >= want_size) {
 	  goto found;
 	}
 	++p;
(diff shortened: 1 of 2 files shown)


src/os/bluestore/StupidAllocator.cc
```diff
@@ -98,6 +98,7 @@ int StupidAllocator::allocate(
   int orig_bin = bin;
 
   auto p = free[0].begin();
+  auto end = free[0].end();
```
tchaikov: better off moving the definition of `end` to where it is used. BTW, its value is always overwritten there.

src/os/bluestore/StupidAllocator.cc
```diff
@@ -139,10 +141,11 @@ int StupidAllocator::allocate(
     }
   }
 
-  // search down (origin)
+  // search down (from origin, and skip searched extents by hint)
```
tchaikov: `free[0..orig_bin]` is not searched at all. so we can not skip any of them.

src/os/bluestore/StupidAllocator.cc
jjhuo: Sure, will do.

src/os/bluestore/StupidAllocator.cc
jjhuo: Ideally, we should only search up for free extents, since previous bin of _choose_bin() only contains free extents with length < need_size. I am not sure why Sage had those two "search down" loops in the first place, maybe he used this as last resort in case code has bugs? @liewegas do you think, we should remove those two "search down" loops at all?

tchaikov: @jjhuo i concur with you. in theory we should return `-ENOSPC` here, before searching in the smaller bins in vain.

tchaikov: could have another commit in this PR to address this, i'd say.

tchaikov:
@jjhuo need to rebase.

src/os/bluestore/StupidAllocator.cc
liewegas: the allocator is allowed to return a short extent if it can't find a big one.. that's why it searches the smaller bins (in descending order).  'need_size' should probably be named 'want_size'.

jjhuo:
@tchaikov rebase is done, thanks for help!

src/os/bluestore/StupidAllocator.cc
tchaikov: @jjhuo so we need to keep as it is.

src/os/bluestore/StupidAllocator.cc
```diff
@@ -119,7 +118,7 @@ int StupidAllocator::allocate(
   // search up (from origin, and skip searched extents by hint)
   for (bin = orig_bin; bin < (int)free.size(); ++bin) {
     p = free[bin].begin();
-    end = hint ? free[bin].lower_bound(hint) : free[bin].end();
+    auto end = hint ? free[bin].lower_bound(hint) : free[bin].end();
```
tchaikov: can you squash this change into the previous one? as we are not expecting the fix-the-fix commit in a single PR.

tchaikov:
maybe you need to remove the "// search down (from origin, and skip searched extents by hint)" change.

jjhuo:
Finished the code squash.
And as Sage said, allocate() will search down and return smaller extents, that means this fix will skip those unnecessary searches as well. I have ran ceph_test_objectstore and saw same behaviors.

liewegas:
lgtm, thanks!

PR #8102, opened 2016-03-14
librados: cancel aio notification linger op upon completion

dillaman:


jdurgin:
lgtm

PR #8103, opened 2016-03-14
common: signed coarse time

adamemerson:


src/common/ceph_time.h
```diff
@@ -259,6 +259,14 @@ namespace ceph {
 	return time_point(seconds(ts.tv_sec) + nanoseconds(ts.tv_nsec));
       }
     };
+
+    template<typename Clock, typename Duration>
+    signedspan difference(
+      typename std::chrono::time_point<Clock, Duration> minuend,
+      typename std::chrono::time_point<Clock, Duration> subtrahend) {
+      return signedspan(minuend.time_since_epoch().count() -
```
tchaikov: nit, this statement actually casts an unsigned variable to a signed one. and the width of `Duration::rep` "happen" to be identical to `signedspan`, that's why the tests pass. can we put this way?
```c++
      using rep_type = typename std::make_signed<typename Duration::rep>::type;
      return signedspan(rep_type(minuend.time_since_epoch().count()) -
                        rep_type(subtrahend.time_since_epoch().count()));
```
i think it would be, well, safer.

or we can have a
```c++
     static_assert(sizeof(typename Duration::rep) ==
                    sizeof(typename signedspan::rep), "bad duration type");
```
to make sure that the cast always works. 

what do you think?

adamemerson: Actually, I think the biggest problem is resolution. I'll throw something in to fix that that should address the width-worry, too.

src/common/ceph_time.h
```diff
@@ -26,6 +26,7 @@ struct ceph_timespec;
 namespace ceph {
   namespace time_detail {
     using std::chrono::duration_cast;
+    using std::chrono::duration_cast;
```
tchaikov: might want to remove this line?

src/common/ceph_time.h
```diff
@@ -259,6 +260,44 @@ namespace ceph {
...
+	    std::chrono::duration<typename std::make_signed<Rep2>::type,
+				  Period2> >::type;
+	return srep(srep(minuend).count() - srep(subtrahend).count());
+      }
+
+      template<typename Clock, typename Duration1, typename Duration2>
+      inline auto difference(
+	typename std::chrono::time_point<Clock, Duration1> minuend,
+	typename std::chrono::time_point<Clock, Duration2> subtrahend)
+	-> typename std::common_type<
```
tchaikov: could also use 

```c++
  -> decltype(difference(minuend.time_since_epoch(), 
                         subtrahend.time_since_epoch()))
```

for less typing, and more copy & paste =)

PR #8106, opened 2016-03-14
cmake: add FindOpenSSL.cmake

mattbenjamin:
Use find_package to enforce the dependency up front (we should be
using this pattern throughout top-level CMakeLists.txt).

(Counting this a bugfix for new SSL support.)


cmake/modules/FindOpenSSL.cmake
```diff
@@ -0,0 +1,37 @@
...
+
+if (OPENSSL_FOUND)
+  message(STATUS "Found ldap: ${OPENSSL_INCLUDE_DIR}")
+else (OPENSSL_FOUND)
+  if (NOT OPENSSL_INCLUDE_DIR)
+    message(FATAL_ERROR "Missing required ssl/ssl.h (openssl-devel)")
+  else (NOT OPENSSL_INCLUDE_DIR)
+    message (FATAL_ERROR "Missing required OpenSSL libraries")
+  endif (NOT OPENSSL_INCLUDE_DIR)
+endif (OPENSSL_FOUND)
```
tchaikov: @mattbenjamin we can simplify the lines from 17 to 31 (inclusive) using 
```
include(FindPackageHandleStandardArgs)
find_package_handle_standard_args(OpenSSL DEFAULT_MSG LIBSSL OPENSSL_INCLUDE_DIR)
```

cmake/modules/FindOpenSSL.cmake
```diff
@@ -0,0 +1,37 @@
...
+set(OPENSSL_LIB_DIR "${OPENSSL_PREFIX}/lib")
+
+find_path(OPENSSL_INCLUDE_DIR ssl/ssl.h NO_DEFAULT_PATH PATHS
+  /usr/include
+  /opt/local/include
+  /usr/local/include
+  "${OPENSSL_PREFIX}/include"
+  )
+
+find_library(LIBSSL NAMES ssl)
```
tchaikov: can we set `OPENSSL_LIBS` here?

mattbenjamin: setting openssl_libs would improve things;  on the simplification, I don't think it's helpful, really

liewegas:
@mattbenjamin ping

mdw-at-linuxbox:
I just talked with Matt - I'll look at making this progress.

cbodley:
cmake comes with its own FindOpenSSL.cmake module (/usr/share/cmake/Modules/FindOpenSSL.cmake on fedora, /usr/share/cmake-x.y/ on ubuntu). can we use that instead?

mdw-at-linuxbox:
I like casey's idea of using the cmake version of FindOpenSSL.  So I pushed a commit that uses this, cleans up a bit of stuff (like building right w/o openssl), and updated master for it.

PR #8108, opened 2016-03-15
Wip rgw highres 3

yehudasa:
adapt rgw to use higher resultion clock


mattbenjamin:
Offhand, this looks great.

PR #8111, opened 2016-03-15
cmake: cleanup cmake/modules/Findblkid.cmake

tchaikov:
* fix the typo of BLKID_INCLUDE_DIR
* remove unnecessary code dups with find_package_handle_standard_args()


cbodley:
we get `BLKID_LIBRARY` as the output of find_library(), but pass `BLKID_LIBRARIES` to find_package_handle_standard_args(). the `set( BLKID_LIBRARIES ${BLKID_LIBRARY} )` line was removed, which made this work previously

PR #8112, opened 2016-03-15
mon/OSDMonitor: Always satisfty get_osdmap request

xiaoxichen:
We would like to cap the max map per message to osd_map_message_max,
but we also need to fully satisfy (return all the map requested) the
GET_OSDMAP request by seperate to multiple messages.

Fixes 15130

### DIFF
--- a/src/mon/OSDMonitor.cc
+++ b/src/mon/OSDMonitor.cc
@@ -1442,21 +1442,34 @@ bool OSDMonitor::preprocess_get_osdmap(MonOpRequestRef op)
   epoch_t first = get_first_committed();
   epoch_t last = osdmap.get_epoch();
   int max = g_conf->osd_map_message_max;
-  for (epoch_t e = MAX(first, m->get_full_first());
-       e <= MIN(last, m->get_full_last()) && max > 0;
-       ++e, --max) {
-    int r = get_version_full(e, reply->maps[e]);
-    assert(r >= 0);
+
+  int full_map_got = 0;
+  int inc_map_got = 0;
+  epoch_t full_first = MAX(first, m->get_full_first());
+  epoch_t inc_first = MAX(first, m->get_inc_first());
+
+  while (full_map_got <= m->get_full_last()
+          && inc_map_got <= m->get_inc_last()) {
+    epoch_t e;
+    for (e = MAX(full_first, full_map_got);
+         e <= MIN(last, m->get_full_last()) && max > 0;
+         ++e, --max) {
+      int r = get_version_full(e, reply->maps[e]);
+      assert(r >= 0);
+    }
+    full_map_got = e;
+
+    for (e = MAX(inc_first, inc_map_got);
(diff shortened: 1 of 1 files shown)


src/mon/OSDMonitor.cc
```diff
@@ -1442,21 +1442,30 @@ bool OSDMonitor::preprocess_get_osdmap(MonOpRequestRef op)
...
-    int r = get_version(e, reply->incremental_maps[e]);
-    assert(r >= 0);
-  }
-  reply->oldest_map = get_first_committed();
-  reply->newest_map = osdmap.get_epoch();
-  mon->send_reply(op, reply);
+
+  int full_map_got = 0;
+  int inc_map_got = 0;
+  while (full_map_got <= m->get_full_last() && inc_map_got <= m->get_inc_last()) {
```
tchaikov: could be better to wrap the line, as it's over 80 chars.

src/mon/OSDMonitor.cc
```diff
@@ -1442,21 +1442,34 @@ bool OSDMonitor::preprocess_get_osdmap(MonOpRequestRef op)
...
+
+  int full_map_got = 0;
+  int inc_map_got = 0;
+  epoch_t full_first = MAX(first, m->get_full_first());
+  epoch_t inc_first = MAX(first, m->get_inc_first());
+
+  while (full_map_got <= m->get_full_last()
+          && inc_map_got <= m->get_inc_last()) {
+    epoch_t e;
+    for (e = MAX(full_first, full_map_got);
```
tchaikov: we can always start from the previous `full_map_got`. how about set it to `full_first` out side of the loop?

tchaikov:
and could you change the "Fixes 15130" line to "Fixes: #15130"?

xiaoxichen:
@jecluis thanks， will fix it.

Is that the right way to fix the bug?

gregsfortytwo:
I really don't think this is the right way around to fix that bug. Except in the case of permanent subscriptions, the monitor's message pattern is very clearly one request->one reply and breaking that will do weird things. I think in most paths the OSD in fact does request additional maps if it gets a response that doesn't get it all the way there, and it's probably a simple off-by-one error in the map handling code.

jecluis:
I hadn't really looked into the ticket before commenting on this pull request. I'm inclined to agree with @gregsfortytwo . You should make sure this is not something on the osd map handling code; after that's done, and if no problem surfaces, then we can start thinking how to fix that from the monitor side.

Also, I don't like the idea of hanging for too long sending messages to the OSD. We're holding the mon lock while we're in this path and we should be as quick as possible about it. Leaving the monitor stuck in this path for an unbounded amount of time is not the greatest of ideas, as it will only bring us unbounded pain.

xiaoxichen:
@gregsfortytwo @jecluis , Thanks, yes seems can fix this on OSD side, which is definitely better.

PR #8114, opened 2016-03-15
qa/workunits/rados/test.sh: test tmap_migrate

liewegas:


liewegas:
We need to fix the test, too.. see http://tracker.ceph.com/issues/15135

PR #8115, opened 2016-03-15
Jewel cephfs data scan misc

ukernel:


src/tools/cephfs/DataScan.h
```diff
@@ -147,7 +147,8 @@ class MetadataTool
    * Construct a synthetic InodeStore for a normal file
    */
   void build_file_dentry(
-    inodeno_t ino, uint64_t file_size, time_t file_mtime,
+    inodeno_t ino, uint64_t nfiles,
```
jcsp: Maybe this is a typo?  It doesn't make sense to call file_size nfiles here, because this is the function for creating normal file dentries, not directories.

jcsp:
:+1: apart from one minor comment (+ I checked that this passes TestDataScan locally)

ukernel:
Sorry, it's typo. I removed it.

gregsfortytwo:

PR #8117, opened 2016-03-15
osd: add sanity check for send_incremental_map()

xiexingguo:


liewegas:
What did you observe that led you to this change?  Skipping maps will make the destination go ask the mon for the maps in the gap, which probably isn't what we want in general.

xiexingguo:
Negative. May be it is just some kind of paranoia thought. So I prefer to close it out for now.

PR #8118, opened 2016-03-15
fix 2 warnings

liewegas:


yehudasa:
@liewegas PR conflicts (os/filestore.cc has changed since).

yehudasa:
ahrm.. that was os/filestore/FileStore.cc

yehudasa:
@liewegas it seems that the second commit is not needed anymore, you can drop it.

yehudasa:
This PR is not needed now (code has changed in master)

PR #8121, opened 2016-03-15
mon: remove 'mds setmap'

liewegas:


liewegas:
@jecluis @jcsp mind reviewing?

liewegas:
passed rados run

PR #7556, opened 2016-02-07
mon: have mon-specific features

jecluis:
UPDATE: fixed patch set according to @liewegas review. See below for highlights.

This patch set adds the following functionality to the monitors:

1. have a 'features' field in the MonMap, holding the bare minimum features required
by ALL the monitors in the cluster.

2. we now have 'sticky' features, which will be automatically activated - and set on
monmap - once all the quorum members support them. These can not be disabled
or reverted. To be used sparingly.

3. have a 'flags' field in the MonMap, holding (possibly) transient features that are
required by all monitors in the quorum.

4. monitors require a given set of features to join a quorum; these will be assessed
relying on `monmap.features` and `monmap.flags`.

5. any given flag can be set in `monmap.flags` once all the monitors in a quorum
are able to understand said flag. These can also be unset. Setting/unsetting flags
will likely require user intervention (or some action that triggers a flag to be set/unset).

`Signed-off-by: Joao Eduardo Luis <joao@suse.de>`

PR #6855, opened 2015-12-08
DNM: mon: MonMap: hold mon info (addr,etc) in dedicated class

jecluis:
I'm pushing this branch, as it is, solely for discussion; should it get the okay to merge, that will happen as part of a broader pull request.

Even if this change seems unwarranted, it's meant as a basis for a having multiple networks on the monitors (i.e., instead of just having a public network). Keeping any given monitor's info on a dedicated struct/class allows us to more easily access it's information and extend the number of, say, addresses. The alternative would be to add clutter to the monmap in ways that make me feel all sorts of wrong.

Reviews will be most welcome.

`Signed-off-by: Joao Eduardo Luis <joao@suse.de>`

PR #8126, opened 2016-03-15
mon: have mon-specific features & rework internal monmap structures

jecluis:
This pull request is the resulting branch from merging two features:

1. mon-specific features (previously, PR #7556)
2. holding mon addr and friends on a structure dedicated to hold mon info (previously, PR #6855)

Further info on each will be found at their original PR's.

### DIFF
--- a/src/mon/MonMap.h
+++ b/src/mon/MonMap.h
@@ -195,11 +195,11 @@ class MonMap {
 
   uuid_d& get_fsid() { return fsid; }
 
-  unsigned size() {
+  unsigned size() const {
     return mon_info.size();
   }
 
-  epoch_t get_epoch() { return epoch; }
+  epoch_t get_epoch() const { return epoch; }
   void set_epoch(epoch_t e) { epoch = e; }
 
   /**
@@ -257,7 +257,7 @@ class MonMap {
     calc_ranks();
   }
 
-  bool contains(const string& name) {
+  bool contains(const string& name) const {
     return mon_info.count(name);
   }
 
@@ -270,8 +270,8 @@ class MonMap {
    * @returns true if monmap contains a monitor with address @p;
    *          false otherwise.
    */
-  bool contains(const entity_addr_t &a) {
-    for (map<string,mon_info_t>::iterator p = mon_info.begin();
+  bool contains(const entity_addr_t &a) const {
+    for (map<string,mon_info_t>::const_iterator p = mon_info.begin();
          p != mon_info.end();
          ++p) {
       if (p->second.public_addr == a)
--- a/src/mon/MonMap.h
+++ b/src/mon/MonMap.h
@@ -105,88 +105,9 @@ class MonMap {
     return (persistent_features | optional_features);
   }
 
-  void sanitize_mons(map<string,entity_addr_t>& o) {
-
-    // if mon_info is populated, it means we decoded a map encoded
-    // by someone who understands the new format (i.e., is able to
-    // encode 'mon_info'). This means they must also have provided
-    // a properly populated 'mon_addr' (which we have dropped with
-    // this patch), 'o' being the contents of said map. In this
-    // case, 'o' must have the same number of entries as 'mon_info'.
-    //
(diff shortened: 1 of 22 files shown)


jecluis:
fwiw, the whole thing has gone through a rados thrash suite. Failed tests have been rerun successfully.

src/mon/MonMap.h
```diff
@@ -19,44 +19,176 @@
...
+    // Please note that we use a 'set' here instead of resorting to
+    // std::sort() because we need more info than that's available in
+    // the vector. The vector will thus be ordered by, e.g., public_addr
+    // while only containing the names of each individual monitor.
+    // The only way of achieving this with std::sort() would be to first
+    // insert every mon_info_t entry into a vector 'foo', std::sort() 'foo'
+    // with custom comparison functions, and then copy each invidual entry
+    // to a new vector. Unless there's a simpler way, we don't think the
+    // added complexity makes up for the additional memory usage of a 'set'.
+    set<mon_info_t, rank_cmp> tmp;
```
tchaikov: - can we move the implementation of this method into MonMap.cc to speed up the compilation?
- could also define the comparator using `WRITE_CMP_OPERATORS_2`.

src/mon/MonMap.h
```diff
@@ -64,116 +196,154 @@ class MonMap {
   uuid_d& get_fsid() { return fsid; }
 
   unsigned size() {
-    return mon_addr.size();
+    return mon_info.size();
```
tchaikov: could take this chance to mark `size()` as a const method.

src/mon/MonMap.h
```diff
@@ -64,116 +196,154 @@ class MonMap {
...
+  /**
+   * Check if monmap contains a monitor with address @p a
+   *
+   * @note checks for all addresses a monitor may have, public or otherwise.
+   *
+   * @param a monitor address
+   * @returns true if monmap contains a monitor with address @p;
+   *          false otherwise.
+   */
   bool contains(const entity_addr_t &a) {
```
tchaikov: could add `const` here also.

src/mon/MonmapMonitor.h
```diff
@@ -83,6 +84,26 @@ class MonmapMonitor : public PaxosService {
 
  private:
   bufferlist monmap_bl;
+
+  class C_ApplyFeatures : public Context {
```
tchaikov: as this class is not used elsewhere, can we define it in MonmapMonitor.cc at the place where it is used, so we are able to find it's implementation right where it is used?

src/mon/mon_types.h
```diff
@@ -234,4 +235,217 @@ struct C_MonOp : public Context
...
+  bool empty() const {
+    return features == 0;
+  }
+
+  /**
+   * Set difference of our features in respect to @p other
+   *
+   * Returns all the elements in our features that are not in @p other
+   *
+   * @returns all the features in @p other but not in our features.
```
tchaikov: this line conflicts with the line above. but i love the above one better =)

liewegas:
@jecluis ping

liewegas:
I think this has to wait.  And I can't think of a reason why having this field in jewel helps us...

jecluis:
Yeah, it can wait as there's nothing actually leveraging it. It would just have been nice in order to time it with the jewel feature bit, but assuming we'll have a kraken feature bit it shouldn't matter much.

src/mon/MonMap.h
jecluis: is it just me thinking that `WRITE_CMP_OPERATORS_2`, along with all the remaining macros for more than one parameter, sort of just obfuscate the code?

tchaikov: right. agree with you. cross that out!

src/mon/mon_types.h
jecluis: yes, this was an artifact from before the semantic change. it'll be fixed on the next push.

jecluis:
see https://github.com/ceph/ceph/pull/10907 instead

PR #8128, opened 2016-03-15
tools/cephfs: fix tmap_upgrade

jcsp:
Move it up to handle the command before we try
and initialize DataScan::driver, because drivers
require a target filesystem, and that doesn't
apply to this command.

Fixes: #15135


jcsp:
Tested this locally with test_rados_api_tmap_migrate and TestDataScan from qa suite.

PR #8130, opened 2016-03-15
mon: do not panic at seeing requests without session

tchaikov:
http://tracker.ceph.com/issues/15113

### DIFF
--- a/src/mon/Monitor.cc
+++ b/src/mon/Monitor.cc
@@ -18,6 +18,7 @@
 #include <signal.h>
 #include <limits.h>
 #include <cstring>
+#include <boost/scope_exit.hpp>
 
 #include "Monitor.h"
 #include "common/version.h"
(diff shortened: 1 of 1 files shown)


tchaikov:
@jecluis mind review it?

src/mon/Monitor.cc
```diff
@@ -3329,40 +3329,63 @@ void Monitor::handle_route(MonOpRequestRef op)
   }
 }
 
+namespace {
+  struct ReentrantLock {
+    const bool initially_locked;
+    Mutex &mutex;
+    ReentrantLock(Mutex& m)
+      : initially_locked(m.is_locked_by_me()),
```
liewegas: I don't think we should ever use is_locked* as a conditional--only for debug assertions.  Notably, this is not available from std::mutex.

liewegas: Better to make the caller(s) who don't already hold lock take it before calling resend_routed_requests?

tchaikov: so we have following call stack:

Monitor::init() => with lock: bootstrap() => win_standalone_election() => win_election() => Monitor::finish_election() => Monitor::resend_routed_requests()

Elector::expire() => Elector::victory() => **with lock:** Monitor::win_election() => Monitor::finish_election() => Monitor::resend_routed_requests()

Elector::expire() => **with lock:** bootstrap() => win_standalone_election() => win_election() => Monitor::finish_election() => Monitor::resend_routed_requests()

the "with lock:" in bold font are added in this PR.

tchaikov:
changelog

- rework to avoid using "is_locked_by_me"
- rebase against master.

not tested, so "[DNM]".

tchaikov:
changelog

- remove the commit to hold the mon lock when resend_routed_requests(): the safe timer does hold the lock for us, so it's not necessary and actually wrong!
- rebased against master.

liewegas:
passed rados suite

tchaikov:
@jecluis ping?

jecluis:
LGTM

PR #8131, opened 2016-03-15
tests: Improving 'make check' execution time

ErwanAliasr1:
When running make -j x check, we face a weird situation where the makefile
targets are spawn in parallel up to "x" but some targets is very very
long and sequential.

This pull request aims at removing this bottlenecks to make a faster build process.

The initial running time was 30mn on a recent laptop and dropped to 14mn.

### DIFF
--- a/qa/workunits/ceph-helpers.sh
+++ b/qa/workunits/ceph-helpers.sh
@@ -147,6 +147,50 @@ function test_teardown() {
 #######################################################################
 
 ##
+# Sends a signal to a single daemon.
+# This is a helper function for kill_daemons
+#
+# After the daemon is sent **signal**, its actual termination
+# will be verified by sending it signal 0. If the daemon is
+# still alive, kill_daemon will pause for a few seconds and
+# try again. This will repeat for a fixed number of times
+# before kill_daemon returns on failure. The list of
+# sleep intervals can be specified as **delays** and defaults
+# to:
+#
+#  0 1 1 1 2 3 5 5 5 10 10 20 60 60 60 120
+#
+# This sequence is designed to not require a sleep time (0) if the
+# machine is fast enough and the daemon terminates in a fraction of a
+# second. The increasing sleep numbers should give plenty of time for
+# the daemon to die even on the slowest running machine. If a daemon
+# takes more than a few minutes to stop (the sum of all sleep times),
+# there probably is no point in waiting more and a number of things
+# are likely to go wrong anyway: better give up and return on error.
+#
+# @param pid the process id to send a signal
+# @param send_signal the signal to send
+# @param delays sequence of sleep times before failure
+#
+function kill_daemon() {
+    local pid=$(cat $1)
+    local send_signal=$2
+    local delays=${3:-0 0 1 1 1 2 3 5 5 5 10 10 20 60 60 60 120}
+    local exit_code=1
+    for try in $delays ; do
+         sleep $try
+         if kill -$send_signal $pid 2> /dev/null ; then
+            exit_code=1
+         else
+            exit_code=0
+            break
+         fi
+         send_signal=0
+    done;
+    return $exit_code
+}
+
+##
 # Kill all daemons for which a .pid file exists in **dir**.  Each
 # daemon is sent a **signal** and kill_daemons waits for it to exit
 # during a few minutes. By default all daemons are killed. If a
@@ -168,24 +212,6 @@ function test_teardown() {
 # if at least one daemon remains, this is treated as an 
 # error and the function return 1.
 #
-# After the daemon is sent **signal**, its actual termination
-# will be verified by sending it signal 0. If the daemon is
-# still alive, kill_daemons will pause for a few seconds and
-# try again. This will repeat for a fixed number of times
-# before kill_daemons returns on failure. The list of
-# sleep intervals can be specified as **delays** and defaults
-# to:
-#
-#  0 1 1 1 2 3 5 5 5 10 10 20 60 60 60 120
-#
-# This sequence is designed to not require a sleep time (0) if the
-# machine is fast enough and the daemon terminates in a fraction of a
-# second. The increasing sleep numbers should give plenty of time for
-# the daemon to die even on the slowest running machine. If a daemon
-# takes more than a few minutes to stop (the sum of all sleep times),
-# there probably is no point in waiting more and a number of things
-# are likely to go wrong anyway: better give up and return on error.
-#
 # @param dir path name of the environment
 # @param signal name of the first signal (defaults to TERM)
 # @param name_prefix only kill match daemons (defaults to all)
@@ -198,27 +224,17 @@ function kill_daemons() {
     local dir=$1
     local signal=${2:-TERM}
     local name_prefix=$3 # optional, osd, mon, osd.1
-    local delays=${4:-0 0 1 1 1 2 3 5 5 5 10 10 20 60 60 60 120}
-
+    local delays=$4 #optional timing
     local status=0
+    local pids=""
+
     for pidfile in $(find $dir 2>/dev/null | grep $name_prefix'[^/]*\.pid') ; do
-        pid=$(cat $pidfile)
-        local send_signal=$signal
-        local kill_complete=false
-        for try in $delays ; do
-            sleep $try
-            if kill -$send_signal $pid 2> /dev/null ; then
-                kill_complete=false
-            else
-                kill_complete=true
-                break
-            fi
-            send_signal=0
-        done
-        if ! $kill_complete ; then
-            status=1
-        fi
+	run_in_background pids kill_daemon $pidfile $signal $delays
     done
+
+    wait_background pids
+    status=$?
+
     $trace && shopt -s -o xtrace
     return $status
 }
--- a/qa/workunits/ceph-helpers.sh
+++ b/qa/workunits/ceph-helpers.sh
@@ -158,10 +158,10 @@ function test_teardown() {
 # sleep intervals can be specified as **delays** and defaults
 # to:
 #
-#  0 1 1 1 2 3 5 5 5 10 10 20 60 60 60 120
+#  0.1 0.2 1 1 1 2 3 5 5 5 10 10 20 60 60 60 120
 #
-# This sequence is designed to not require a sleep time (0) if the
-# machine is fast enough and the daemon terminates in a fraction of a
+# This sequence is designed to run first a very short sleep time (0.1)
+# if the machine is fast enough and the daemon terminates in a fraction of a
 # second. The increasing sleep numbers should give plenty of time for
 # the daemon to die even on the slowest running machine. If a daemon
 # takes more than a few minutes to stop (the sum of all sleep times),
@@ -175,10 +175,9 @@ function test_teardown() {
 function kill_daemon() {
     local pid=$(cat $1)
     local send_signal=$2
-    local delays=${3:-0 0 1 1 1 2 3 5 5 5 10 10 20 60 60 60 120}
+    local delays=${3:-0.1 0.2 1 1 1 2 3 5 5 5 10 10 20 60 60 60 120}
     local exit_code=1
     for try in $delays ; do
-         sleep $try
          if kill -$send_signal $pid 2> /dev/null ; then
             exit_code=1
          else
@@ -186,8 +185,43 @@ function kill_daemon() {
             break
          fi
          send_signal=0
+         sleep $try
     done;
     return $exit_code
+}
+
+function test_kill_daemon() {
+    local dir=$1
+    setup $dir || return 1
+    run_mon $dir a --osd_pool_default_size=1 || return 1
+    run_osd $dir 0 || return 1
+
+    name_prefix=osd
+    for pidfile in $(find $dir 2>/dev/null | grep $name_prefix'[^/]*\.pid') ; do
+        #
+        # sending signal 0 won't kill the daemon
+        # waiting just for one second instead of the default schedule
+        # allows us to quickly verify what happens when kill fails
+        # to stop the daemon (i.e. it must return false)
+        #
+        ! kill_daemon $pidfile 0 1 || return 1
+        #
+        # killing just the osd and verify the mon still is responsive
+        #
+        kill_daemon $pidfile TERM || return 1
+    done
+
+    ceph osd dump | grep "osd.0 down" || return 1
+
+    for pidfile in $(find $dir -name "*.pid" 2>/dev/null) ; do
+        #
+        # kill the mon and verify it cannot be reached
+        #
+        kill_daemon $pidfile TERM || return 1
+        ! ceph --connect-timeout 60 status || return 1
+    done
+
+    teardown $dir || return 1
 }
 
 ##
--- a/qa/workunits/ceph-helpers.sh
+++ b/qa/workunits/ceph-helpers.sh
@@ -1283,6 +1283,83 @@ function test_display_logs() {
 }
 
 #######################################################################
+##
+# Spawn a command in background and save the pid in the variable name
+# passed in argument. To make the output reading easier, the output is
+# prepend with the process id.
+#
+# Example:
+#   pids1=""
+#   run_in_background pids1 bash -c 'sleep 1; exit 1'
+#
+# @param pid_variable the variable name (not value) where the pids will be stored
+# @param ... the command to execute
+# @return only the pid_variable output should be considered and used with **wait_background**
+#
+function run_in_background() {
+    local pid_variable=$1
+    shift;
+    # Execute the command and prepend the output with its pid
+    # We enforce to return the exit status of the command and not the awk one.
+    ("$@" |& awk '{ a[i++] = $0 }END{for (i = 0; i in a; ++i) { print PROCINFO["pid"] ": " a[i]} }'; return ${PIPESTATUS[0]}) &
+    eval "$pid_variable+=\" $!\""
+}
+
+function test_run_in_background() {
+    local pids
+    run_in_background pids sleep 1
+    run_in_background pids sleep 1
+    test $(echo $pids | wc -w) = 2 || return 1
+    wait $pids || return 1
+}
+
+#######################################################################
+##
+# Wait for pids running in background to complete.
+# This function is usually used after a **run_in_background** call
+# Example:
+#   pids1=""
+#   run_in_background pids1 bash -c 'sleep 1; exit 1'
+#   wait_background pids1
+#
+# @param pids The variable name that contains the active PIDS. Set as empty at then end of the function.
+# @return returns 1 if at least one process exits in error unless returns 0
+#
+function wait_background() {
+    # We extract the PIDS from the variable name
+    pids=${!1}
+
+    return_code=0
+    for pid in $pids; do
+        if ! wait $pid; then
+            # If one process failed then return 1
+            return_code=1
+        fi
+    done
+
+    # We empty the variable reporting that all process ended
+    eval "$1=''"
+
+    return $return_code
+}
+
+
+function test_wait_background() {
+    local pids=""
+    run_in_background pids bash -c "sleep 1; exit 1"
+    run_in_background pids bash -c "sleep 2; exit 0"
+    wait_background pids
+    if [ $? -ne 1 ]; then return 1; fi
+
+    run_in_background pids bash -c "sleep 1; exit 0"
+    run_in_background pids bash -c "sleep 2; exit 0"
+    wait_background pids
+    if [ $? -ne 0 ]; then return 1; fi
+
+    if [ ! -z "$pids" ]; then return 1; fi
+}
+
+#######################################################################
 
 ##
 # Call the **run** function (which must be defined by the caller) with
--- a/qa/workunits/ceph-helpers.sh
+++ b/qa/workunits/ceph-helpers.sh
@@ -940,7 +940,7 @@ function get_num_active_clean() {
     # grep -v '^$' 
     ceph --format xml pg dump pgs 2>/dev/null | \
         $XMLSTARLET sel -t -m "//pg_stat/state[$expression]" -v . -n | \
-        grep -v '^$' | wc -l
+        grep -cv '^$'
 }
 
 function test_get_num_active_clean() {
--- a/qa/workunits/ceph-helpers.sh
+++ b/qa/workunits/ceph-helpers.sh
@@ -1040,32 +1040,36 @@ function test_is_clean() {
 
 ##
 # Wait until the cluster becomes clean or if it does not make progress
-# for $TIMEOUT seconds. The function **is_clean** is used to determine
-# if the cluster is clean. Progress is measured either vian the
-# **get_is_making_recovery_progress** predicate or if the number of
-# clean PGs changes.
+# for $TIMEOUT seconds.
+# Progress is measured either via the **get_is_making_recovery_progress**
+# predicate or if the number of clean PGs changes (as returned by get_num_active_clean)
 #
 # @return 0 if the cluster is clean, 1 otherwise
 #
 function wait_for_clean() {
     local status=1
-    local num_active_clean=$(get_num_active_clean)
+    local num_active_clean=-1
     local cur_active_clean
     local -i timer=0
-    while ! is_clean ; do
-        if get_is_making_recovery_progress ; then
-            timer=0
-        elif (( timer >= $TIMEOUT )) ; then
-            ceph report
-            return 1
-        fi
+    local num_pgs=$(get_num_pgs)
+    test $num_pgs != 0 || return 1
 
+    while true ; do
+        # Comparing get_num_active_clean & get_num_pgs is used to determine
+        # if the cluster is clean. That's almost an inline of is_clean() to
+        # get more performance by avoiding multiple calls of get_num_active_clean.
         cur_active_clean=$(get_num_active_clean)
+        test $cur_active_clean = $num_pgs && break
         if test $cur_active_clean != $num_active_clean ; then
             timer=0
             num_active_clean=$cur_active_clean
+        elif get_is_making_recovery_progress ; then
+            timer=0
+        elif (( timer >= $(($TIMEOUT * 10)))) ; then
+            ceph report
+            return 1
         fi
-        sleep 1
+        sleep .1
         timer=$(expr $timer + 1)
     done
     return 0
--- a/src/test/ceph_objectstore_tool.py
+++ b/src/test/ceph_objectstore_tool.py
@@ -44,7 +44,8 @@ def wait_for_health():
     print "Wait for health_ok...",
     tries = 0
     while call("./ceph health 2> /dev/null | grep -v 'HEALTH_OK\|HEALTH_WARN' > /dev/null", shell=True) == 0:
-        if ++tries == 150:
+        tries += 1
+        if tries == 150:
             raise Exception("Time exceeded to go to health")
         time.sleep(1)
     print "DONE"
--- a/src/test/ceph_objectstore_tool.py
+++ b/src/test/ceph_objectstore_tool.py
@@ -44,9 +44,9 @@ def wait_for_health():
     print "Wait for health_ok...",
     tries = 0
     while call("./ceph health 2> /dev/null | grep -v 'HEALTH_OK\|HEALTH_WARN' > /dev/null", shell=True) == 0:
-        if ++tries == 30:
+        if ++tries == 150:
             raise Exception("Time exceeded to go to health")
-        time.sleep(5)
+        time.sleep(1)
     print "DONE"
 
 
@@ -1799,15 +1799,11 @@ def main(argv):
         vstart(new=False)
         wait_for_health()
 
-        time.sleep(20)
-
         cmd = "./ceph osd pool set {pool} pg_num 2".format(pool=SPLIT_POOL)
         logging.debug(cmd)
         ret = call(cmd, shell=True, stdout=nullfd, stderr=nullfd)
         time.sleep(5)
         wait_for_health()
-
-        time.sleep(15)
 
         kill_daemons()
 
--- a/src/test/encoding/check-generated.sh
+++ b/src/test/encoding/check-generated.sh
@@ -19,13 +19,6 @@ for type in `./ceph-dencoder list_types`; do
     num=`./ceph-dencoder type $type count_tests`
     echo "$num $type"
     for n in `seq 1 1 $num 2>/dev/null`; do
-	if ! ./ceph-dencoder type $type select_test $n encode decode; then
-	    echo "**** $type test $n encode+decode check failed ****"
-	    echo "   ceph-dencoder type $type select_test $n encode decode"
-	    failed=$(($failed + 3))
-	    continue
-	fi
-
 	safe_type=$type
 	# BitVector<2> needs some escaping to avoid bash issues with <>
 	if [ "$type" = "BitVector<2>" ]; then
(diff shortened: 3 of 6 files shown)


src/test/encoding/readable.sh
```diff
@@ -114,15 +118,86 @@ for arversion in `ls -v $dir/archive`; do
...
+     for pid in "$@"; do
+       shift
+       if kill -0 "$pid" 2>/dev/null; then
+         debug "$pid is still alive."
+         set -- "$@" "$pid"
+       elif wait "$pid"; then
+         debug "$pid exited with zero exit status."
+       else
+         debug "$pid exited with non-zero exit status."
+         ((++errors))
```
tchaikov: this is bash extension not POSIX compatible. in other words, this might not work in some distro, for instance debian derivatives, which point `/bin/sh` to `dash`.

ErwanAliasr1:
@tchaikov Repushed with the fixes you asked for. I rebased & updating the commit comment to explain that we can override the number of // jobs by using MAX_PARALLEL_JOBS

dachary:
:+1:

branch-predictor:
:+1:, here it reduced make check time from 28 down to 17 minutes.

wjwithagen:
On 15-3-2016 17:15, Erwan Velu wrote:
> When running make -j x check, we face a weird situation where the makefile
> targets are spawn in parallel up to "x" but one of those target is very very
> long and sequential.
>
> The "readable.sh" test is trying to run ~7.9K tests where 5.3K are actually
> executed.
>
> The current code is taking 23mn on a recent laptop (Intel(R) Core(TM)
> i7-4810MQ CPU @ 2.80GHz, 32GB of RAM & SSD).
>
> This patch implements parallelism to speed up this process which is not
> really CPU and
> neither IO bound.
>
> By default, readable.sh is now using the number of logical processors to
> determine
> the level of parallelism (by using nproc).
>
> On the same system, where nproc=8, the resulting execution time is 5m55
> seconds :
> 4x faster than the original code.
>
> The global 'make check' is therefore getting faster too and dropped from
> 30 to
> 16 minutes : 2x faster than the original code.

Something I've contemplating is to separate the checks into at least three
parts:
	simple (unit) tests with no benchmark component
	convoluted tests which require ods and mons to run
	benchmarks that really take a long time.

And then have make run these tests in sequence.
Problem for me now is that benchmarks and complexer tests are run, whilest
there could be trouble in the more simpel tests.

Just haven't gotten around it. Next to that unittests are not always without
benchmark part and/or requiring osds/mons

--WjW

ErwanAliasr1:
Just added a new commit to save one more minute without impacting the loadavg.

ErwanAliasr1:
Please don't merge that PR now, I found many other options to optimize the make check. I'll update the PR accordingly and tell when it's ready.

dachary:
@ErwanAliasr1 we usually prefix the title of a pull request that must not be merged with DNM (Do Not Merge). When it's ready, just remove the prefix :-)

src/test/encoding/readable.sh
```diff
@@ -4,23 +4,25 @@ dir=../ceph-object-corpus
...
 
-tmp1=`mktemp /tmp/typ-XXXXXXXXX`
-tmp2=`mktemp /tmp/typ-XXXXXXXXX`
-
 failed=0
 numtests=0
+pids=""
 
 myversion=`./ceph-dencoder version`
+DEBUG=0
```
tchaikov: @ErwanAliasr1 could you revise the title of your commit message with something like: "test/encoding/readable.sh: parallelize the tests" ?

so it is more specific?

tchaikov:
tested the readable.sh changes: 52m53.915s => 24m20.108s. will continue with other commits tomorrow.

src/test/encoding/readable.sh
ErwanAliasr1: Good point, fixed & repushed.

src/test/ceph_objectstore_tool.py
```diff
@@ -44,9 +44,9 @@ def wait_for_health():
     print "Wait for health_ok...",
     tries = 0
     while call("./ceph health 2> /dev/null | grep -v 'HEALTH_OK\|HEALTH_WARN' > /dev/null", shell=True) == 0:
-        if ++tries == 30:
+        if ++tries == 150:
```
tchaikov: this is actually a bug. in python "+" is an identity operator. so "++tries" evaluates to "try".

ErwanAliasr1: Fixed by commit https://github.com/ceph/ceph/pull/8131/commits/820ed6b1b000a15a0fc2a8efc954dfb8185ada50

src/test/ceph_objectstore_tool.py
```diff
@@ -44,7 +44,8 @@ def wait_for_health():
     print "Wait for health_ok...",
     tries = 0
     while call("./ceph health 2> /dev/null | grep -v 'HEALTH_OK\|HEALTH_WARN' > /dev/null", shell=True) == 0:
-        if ++tries == 150:
+        tries = tries + 1
```
tchaikov: could be
```py
tries += 1
```

and my first name is "Kefu"

also ideally, you could have an empty line before the "Signed-off-by" line. "git commit -s" can do it nicely.

ErwanAliasr1:
Most of the work is done here. I think its time for review + merge.
If I found something else I'll make a 2nd PR for it.

ErwanAliasr1:
Rebased & pushed.

qa/workunits/ceph-helpers.sh
```diff
@@ -197,26 +197,30 @@ function kill_daemons() {
...
     local signal=${2:-TERM}
     local name_prefix=$3 # optional, osd, mon, osd.1
-    local delays=${4:-0 0 1 1 1 2 3 5 5 5 10 10 20 60 60 60 120}
+    local delays=${4:-0.1 0.2 1 1 1 2 3 5 5 5 10 10 20 60 60 60 120}
 
     local status=0
+    pids=""
     for pidfile in $(find $dir 2>/dev/null | grep $name_prefix'[^/]*\.pid') ; do
-        pid=$(cat $pidfile)
+        (pid=$(cat $pidfile)
```
dachary: move that part to a separate function, it's becoming too big

qa/workunits/ceph-helpers.sh
```diff
@@ -197,26 +197,30 @@ function kill_daemons() {
...
-        done
-        if ! $kill_complete ; then
-            status=1
-        fi
+            sleep $try
+        done; exit $exit_code) &
+	pids="$pids $!"
+    done
+    for pid in $pids; do
+	if ! wait $pid; then
```
dachary: if $pid terminated before wait gets there, wait will return 127 which will be confused as an error when it really means the daemon is already dead

dachary: I like this optimization :-) It needs to be tested in test_kill_daemons (the race mentionned above can be faked by making it so the first daemon in the list blocs and the second one terminates quickly).

dachary: and the documentation needs updating

qa/workunits/ceph-helpers.sh
```diff
@@ -1032,23 +1036,25 @@ function test_is_clean() {
...
 
+    while true ; do
         cur_active_clean=$(get_num_active_clean)
+        test $cur_active_clean = $num_pgs && break
         if test $cur_active_clean != $num_active_clean ; then
             timer=0
             num_active_clean=$cur_active_clean
+        elif get_is_making_recovery_progress ; then
+            timer=0
+        elif (( timer >= $TIMEOUT )) ; then
```
dachary: timer * 10

qa/workunits/ceph-helpers.sh
```diff
@@ -1032,23 +1036,25 @@ function test_is_clean() {
 #
 function wait_for_clean() {
```
dachary: the documentation needs updating

src/test/ceph_objectstore_tool.py
```diff
@@ -1796,16 +1797,12 @@ def main(argv):
         vstart(new=False)
         wait_for_health()
 
-        time.sleep(20)
```
dachary: I unfortunately think there is a reason for that sleep but it's unclear what it is.

dachary:
to validate this series of change, a single make check is not enough. Since it modifies timing and timeouts in a number of places, make check should be run in a loop for a day or so with no failure.

src/test/encoding/check-generated.sh
```diff
@@ -24,16 +24,21 @@ for type in `./ceph-dencoder list_types`; do
...
 
-	./ceph-dencoder type $type select_test $n dump_json > $tmp1
-	./ceph-dencoder type $type select_test $n encode decode dump_json > $tmp2
-	./ceph-dencoder type $type select_test $n copy dump_json > $tmp3
-	./ceph-dencoder type $type select_test $n copy_ctor dump_json > $tmp4
+	./ceph-dencoder type $type select_test $n dump_json > $tmp1 & pids="$!"
+	./ceph-dencoder type $type select_test $n encode decode dump_json > $tmp2 & pids="$pids $!"
+	./ceph-dencoder type $type select_test $n copy dump_json > $tmp3 & pids="$pids $!"
+	./ceph-dencoder type $type select_test $n copy_ctor dump_json > $tmp4 & pids="$pids $!"
+	for pid in $pids; do wait $pid; done
```
dachary: watch out for the -127 return that indicates the process finished early

src/test/encoding/check-generated.sh
```diff
@@ -65,10 +70,11 @@ for type in `./ceph-dencoder list_types`; do
...
 
-	if ./ceph-dencoder type $type is_deterministic
-	then
-	    ./ceph-dencoder type $type select_test $n encode export $tmp1
-	    ./ceph-dencoder type $type select_test $n encode decode encode export $tmp2
+	if [ $deterministic -ne 0 ]; then
+	    ./ceph-dencoder type $type select_test $n encode export $tmp1 & pids="$!"
+	    ./ceph-dencoder type $type select_test $n encode decode encode export $tmp2 & pids="$pids $!"
+	    for pid in $pids; do wait $pid; done
+
```
dachary: dito

dachary:
the pattern proc & pids+="$!" ... wait for $pids to complete is used in many places and looks useful. It deserves a function in ceph-helpers.sh with various tests for border cases such as when a pid terminates early. There also is a potentially confusing mix of the output of all the process that are run in parallel. We could either print them as they go and prefix each line with the command. Or redirect each stdout/stderr in a separate file and print them one after the other so the developer does not need to sort which output belongs to which command when trying to figure out the source of the problem.

wjwithagen:
@ErwanAliasr1 @dachary 
I did some hacking in qa/workunits/ceph-helpers.sh to adopt to FreeBSD. if you want look at it they are in #7573, but that deserves a big cleanup.

One of the difficult parts there is that is seems that under FreeBSD OSD/MONs only die after 
2* kill TERM.

But I've also become rather annoyed at the fact that with failures the dumps of the OSD/MONs are in the output. Especially during porting/testing/debugging it is much more useful if the tests in testdir are not removed and can be used for post-mortum debugging. I've added tests on the ENV CEPH_DEBUG, but any other way of achieving this would be welcom. 

I also noticed @ldachary mentions "update manual" but in all attempts to find manuals about the testing, I was not able to find anything. (Perhaps because I don't make the manuals during building. But on the Ceph site I did not find much)

src/test/ceph_objectstore_tool.py
ErwanAliasr1: Fixed. Sorry for the mispelling.

qa/workunits/ceph-helpers.sh
ErwanAliasr1: Done.

qa/workunits/ceph-helpers.sh
ErwanAliasr1: Done.

src/test/ceph_objectstore_tool.py
ErwanAliasr1: I had the exact same remark. But as long the wait_for_health is complete it should be ok. If not, then we shall make a specific test instead of waiting an amount of time which doesn't guarantee anything.
As far as we cannot determine the meaning of such wait, my proposal here is to remove it and add a real test if someone notices a regression. I didn't had any during my tests.

src/test/encoding/check-generated.sh
ErwanAliasr1: ceph-dencoder is lasting long so this is very unlikely but if so, we are in the same behavior right ? I mean the return code of the ceph-dencoder wasn't tested before. Shall we ?

src/test/ceph_objectstore_tool.py
dachary: > if someone notices a regression

this is going to be a painful process since a) the problem is most probably going to be transient, b) the person observing it is not going to figure out this removal could be the cause. 

the removal of these two sleep() is desirable but they deserve a separate pull request, discussion with david and very careful thinking

src/test/encoding/check-generated.sh
dachary: the whole script runs with **set -e** and the expected behavior is that it aborts as soon as something fails

dachary: > ceph-dencoder is lasting long so this is very unlikely 

It will happen if **ceph-dencoder type $type select_test $n dump_json** finishes after **ceph-dencoder type $type select_test $n encode decode dump_json**

dachary:
Hum, actually wait $pid will only return -127 if the pid has never been a child of the current shell. So there is nothing special to do to handle -127.

src/test/ceph_objectstore_tool.py
ErwanAliasr1: @dzafman Can you enlight us on the need of having those sleeps here ?

src/test/encoding/check-generated.sh
ErwanAliasr1: Fair enough. That kills my PR then. I have to think on how to handle that.

ErwanAliasr1: We are safe against the return code 127 as per https://gist.github.com/ErwanAliasr1/1969b33d2c538a65d8ed

We only get 127 if the process never existed which is not the case here.

ErwanAliasr1: So for me the code is safe as wait will assert if one pid return failed and will not get 127 because of races between process.

ErwanAliasr1:
@dachary Updated the branch to implement & use helpers. Remains the logging issue to fix.

qa/workunits/ceph-helpers.sh
```diff
@@ -146,6 +146,27 @@ function test_teardown() {
 #######################################################################
 
 ##
+# Kills a single daemon. This is an helper function fo kill_daemons
```
dachary: Could you please document it as a first class citizen ? All functions in this file are helpers and they should all be individually tested and documented. The general idea is that someone interested in a given function does not need to look beyond the part of the file where it's defined, documented and tested. That locality is very valuable for the casual user.

qa/workunits/ceph-helpers.sh
```diff
@@ -1023,32 +1033,36 @@ function test_is_clean() {
 
 ##
 # Wait until the cluster becomes clean or if it does not make progress
-# for $TIMEOUT seconds. The function **is_clean** is used to determine
-# if the cluster is clean. Progress is measured either vian the
-# **get_is_making_recovery_progress** predicate or if the number of
-# clean PGs changes.
+# for $TIMEOUT seconds.
+# Comparing get_num_active_clean & get_num_pgs is used to determine
+# if the cluster is clean. That's almost an inline of is_clean() to
```
dachary: the documentation is designed to be read by the user and should not contain implementation details. it should focus on documenting precisely each aspect of function behavior and error handling, pitfall etc. 

the reason why is_clean is inlined could usefuly be added as a comment within the function itself, as a reminder for the implementer

qa/workunits/ceph-helpers.sh
```diff
@@ -1282,6 +1296,53 @@ function test_display_logs() {
 }
 
 #######################################################################
+##
+# run_in_background() function spawn the command line passed in
```
dachary: s/run_in_background() function // (like the other descriptions)

qa/workunits/ceph-helpers.sh
```diff
@@ -1282,6 +1296,53 @@ function test_display_logs() {
...
+# argument in background and save the pid in the variable name passed
+# in 1st argument like :
+#
+# pids1=""
+# run_in_background pids1 bash -c 'sleep 1; exit 1'
```
dachary: Add a section with @param & @return as for the other fucntions

qa/workunits/ceph-helpers.sh
```diff
@@ -1282,6 +1296,53 @@ function test_display_logs() {
...
+#
+function run_in_background() {
+    # The first argument is the name of the PID variable
+    # We execute everything passed in argument
+    # And save the running pid in the variable name
+    local pid_variable=$1
```
dachary: the above comment is redundant with the documentation and can be removed

qa/workunits/ceph-helpers.sh
```diff
@@ -1282,6 +1296,53 @@ function test_display_logs() {
...
+    shift;
+    "$@" & eval "$pid_variable+=\" $!\""
+}
```
dachary: add a test_run_in_background() function that verifies it does the right thing

qa/workunits/ceph-helpers.sh
```diff
@@ -1282,6 +1296,53 @@ function test_display_logs() {
...
+    done
+
+    # We empty the variable reporting that all process ended
+    eval "$1=''"
+
+    return $return_code
+}
+
+
+#######################################################################
```
dachary: remove that extra separator

qa/workunits/ceph-helpers.sh
```diff
@@ -146,6 +146,27 @@ function test_teardown() {
...
+	    exit_code=0
+            break
+         fi
+         send_signal=0
+         sleep $try
+    done;
+    return $exit_code
+}
+
+##
```
dachary: add a test_kill_deamon function (moving code from test_kill_daemons most probably)

qa/workunits/ceph-helpers.sh
ErwanAliasr1: Done in the new version of that commit.

qa/workunits/ceph-helpers.sh
ErwanAliasr1: Fixed in the new version of that commit.

qa/workunits/ceph-helpers.sh
ErwanAliasr1: Done in last version of the commit.

qa/workunits/ceph-helpers.sh
ErwanAliasr1: Done in last version of the commit.

qa/workunits/ceph-helpers.sh
ErwanAliasr1: Done in last version of the commit.

qa/workunits/ceph-helpers.sh
ErwanAliasr1: Done in last version.

qa/workunits/ceph-helpers.sh
ErwanAliasr1: kill_daemons doesn't do much except calling kill_daemon several time.
The actual test_kill_daemons test the feature pretty well. Having that test passing implies that kill_daemon is ok. Seems overkill to me to add such test.

qa/workunits/ceph-helpers.sh
ErwanAliasr1: Added in the last version of the commit.

qa/workunits/ceph-helpers.sh
```diff
@@ -146,6 +146,50 @@ function test_teardown() {
 #######################################################################
 
 ##
+# Sends a signal to a single daemon.
+# This is an helper function fo kill_daemons
```
dachary: s/an/a/ s/fo/for/

qa/workunits/ceph-helpers.sh
```diff
@@ -146,6 +146,50 @@ function test_teardown() {
...
+#
+# After the daemon is sent **signal**, its actual termination
+# will be verified by sending it signal 0. If the daemon is
+# still alive, kill_daemons will pause for a few seconds and
```
dachary: s/daemons/daemon/

qa/workunits/ceph-helpers.sh
```diff
@@ -146,6 +146,50 @@ function test_teardown() {
...
+# try again. This will repeat for a fixed number of times
+# before kill_daemons returns on failure. The list of
+# sleep intervals can be specified as **delays** and defaults
+# to:
+#
+#  0.1 0.2 1 1 1 2 3 5 5 5 10 10 20 60 60 60 120
+#
+# This sequence is designed to not require a sleep time (0) if the
```
dachary: s/not require a sleep time (0)/only sleep 0.1 second/

qa/workunits/ceph-helpers.sh
dachary: You can create the test function by moving some of the test cases from test_kill_daemons. The important point here is that someone maintaining this function should not need to look anywhere else to fix a bug or add a feature. It should be self contained. Sure, the kill_deamons tests also cover kill_daemon. But when a newcomer works on kill_daemon, it is more work for her/him to figure that out. 

This helper file has been carefully designed and maintained to preserve its maintainability and usability.

qa/workunits/ceph-helpers.sh
```diff
@@ -167,24 +211,6 @@ function test_teardown() {
...
-#  0 1 1 1 2 3 5 5 5 10 10 20 60 60 60 120
-#
-# This sequence is designed to not require a sleep time (0) if the
-# machine is fast enough and the daemon terminates in a fraction of a
-# second. The increasing sleep numbers should give plenty of time for
-# the daemon to die even on the slowest running machine. If a daemon
-# takes more than a few minutes to stop (the sum of all sleep times),
-# there probably is no point in waiting more and a number of things
-# are likely to go wrong anyway: better give up and return on error.
-#
```
dachary: the @param delay below is no longer valid

qa/workunits/ceph-helpers.sh
```diff
@@ -197,27 +223,17 @@ function kill_daemons() {
     local dir=$1
     local signal=${2:-TERM}
     local name_prefix=$3 # optional, osd, mon, osd.1
-    local delays=${4:-0 0 1 1 1 2 3 5 5 5 10 10 20 60 60 60 120}
-
+    local delays=$4 #optional timing
```
dachary: I think this argument is only used by test_kill_daemons and can be removed now that it moved to kill_daemon.

qa/workunits/ceph-helpers.sh
```diff
@@ -1023,32 +1039,36 @@ function test_is_clean() {
 
 ##
 # Wait until the cluster becomes clean or if it does not make progress
-# for $TIMEOUT seconds. The function **is_clean** is used to determine
-# if the cluster is clean. Progress is measured either vian the
-# **get_is_making_recovery_progress** predicate or if the number of
-# clean PGs changes.
+# for $TIMEOUT seconds.
+# Progress is measured either via the **get_is_making_recovery_progress**
+# predicate or if the number of clean PGs changes.
```
dachary: s/number of clean PGs/number of clean PGs as returned by **get_num_active_clean**/

qa/workunits/ceph-helpers.sh
```diff
@@ -1282,6 +1302,70 @@ function test_display_logs() {
...
+##
+# Spawn a command in background and save the pid in the variable name
+# passed in argument
+#
+# Example:
+#   pids1=""
+#   run_in_background pids1 bash -c 'sleep 1; exit 1'
+#
+# @param pid_variable the variable name (not value) where the pids will be stored
+# @param $@ the command to execute
```
dachary: **$@** also includes the first argument, use **...** instead (that's how it's done in other functions as well, C prototype style).

qa/workunits/ceph-helpers.sh
```diff
@@ -1282,6 +1302,70 @@ function test_display_logs() {
...
+    done
+
+    # We empty the variable reporting that all process ended
+    eval "$1=''"
+
+    return $return_code
+}
+
+
+function test_run_in_background() {
```
dachary: s/test_run_in_background/test_wait_background/

qa/workunits/ceph-helpers.sh
```diff
@@ -1282,6 +1302,70 @@ function test_display_logs() {
...
+#
+# @param pid_variable the variable name (not value) where the pids will be stored
+# @param $@ the command to execute
+# @return only the pid_variable output should be considered and used with **wait_background**
+#
+function run_in_background() {
+    local pid_variable=$1
+    shift;
+    "$@" & eval "$pid_variable+=\" $!\""
+}
```
dachary: Here it is:

<pre>
function test_run_in_background() {
   local pids
   run_in_background pids sleep 1
   test "$pids" || return 1
   wait $pids || return 1
}
</pre>

src/test/encoding/check-generated.sh
```diff
@@ -17,23 +19,35 @@ for type in `./ceph-dencoder list_types`; do
...
     for n in `seq 1 1 $num 2>/dev/null`; do
-	if ! ./ceph-dencoder type $type select_test $n encode decode; then
+	safe_type=$type
+	# BitVector<2> needs some escaping to avoid bash issues with <>
+	if [ "$type" = "BitVector<2>" ]; then
+	    safe_type="BitVector\<2\>"
+	fi
+
+	pids=""
+	run_in_background pids bash -c "./ceph-dencoder type $safe_type select_test $n dump_json > $tmp1"
```
dachary: You can 

    run_in_background pids ./ceph-dencoder type $safe_type select_test $n dump_json > $tmp1

there is no need for the bash -c. But it does not really matter ;-)

src/test/osd/osd-scrub-repair.sh
```diff
@@ -75,8 +75,13 @@ function corrupt_and_repair_two() {
     #
     # 1) remove the corresponding file from the OSDs
     #
-    objectstore_tool $dir $first SOMETHING remove || return 1
-    objectstore_tool $dir $second SOMETHING remove || return 1
+    pids=""
+    run_in_background pids objectstore_tool $dir $first SOMETHING remove
```
dachary: should something fail here, it will be very difficult to sort out what comes from the first process and what comes from the second process. I suggest run_in_background redirects stdout/stderr and print them in order, when the process terminates. Since the pids are collected in order, the output will also be displayed in order, which will help a lot to debug.

src/test/encoding/readable.sh
```diff
@@ -114,15 +122,86 @@ for arversion in `ls -v $dir/archive`; do
...
+            running_jobs=$(($running_jobs - 1))
+        done
+        running_jobs=0
+    fi
   done
 done
 
-rm -f $tmp1 $tmp2
-
 if [ $failed -gt 0 ]; then
```
dachary: this refactor is extensive. since readable.sh is essential to preserve data integrity, it's worth having more reviewers as well as tests that demonstrate it will keep catching the errors it is supposed to catch. it would be bad (that happened to me more than once) to realize in two months time that a slight mistake turned the test into something that succeeds when it should fail ;-)

dachary: I propose this is turned into a separate pull request as it is relatively independant from the rest.

qa/workunits/ceph-helpers.sh
ErwanAliasr1: That's an helper as it starts with a vowel sound.
Fixed for the for.

qa/workunits/ceph-helpers.sh
ErwanAliasr1: thx, fixed.

qa/workunits/ceph-helpers.sh
ErwanAliasr1: It was possible to pass this argument to kill_daemons so I was propagating the value. Is that an issue ? That still could be useful.

qa/workunits/ceph-helpers.sh
ErwanAliasr1: fixed

qa/workunits/ceph-helpers.sh
ErwanAliasr1: fixed thx

qa/workunits/ceph-helpers.sh
ErwanAliasr1: Why does mine is incorrect ?

qa/workunits/ceph-helpers.sh
ErwanAliasr1: done

qa/workunits/ceph-helpers.sh
ErwanAliasr1: Oh you want a test for both. Ok.

ErwanAliasr1: added. thx.

qa/workunits/ceph-helpers.sh
ErwanAliasr1: I will look at that tomorrow

qa/workunits/ceph-helpers.sh
ErwanAliasr1: fixed.

src/test/encoding/readable.sh
ErwanAliasr1: The main goal was a performance gain. But would be nice to get more review on it yes.

src/test/encoding/check-generated.sh
ErwanAliasr1: I'm not sure because of the redirection. I don't really check how it would be handled.

src/test/osd/osd-scrub-repair.sh
ErwanAliasr1: Yes I still have to work on the interlaced output stream.

qa/workunits/ceph-helpers.sh
dachary: It's not used for anything but testing the function itself and such testing should now be done in kill_daemon.

qa/workunits/ceph-helpers.sh
dachary: nit: could you group run_in_background and test_run_in_background together and separate them from wait_background and test_wait_in_background with a ####################################################################### line ? So it is like all other functions in the file. Exactly one function, one test and the documentation :-)

src/test/encoding/check-generated.sh
dachary: you can leave it as it is, that's ok

qa/workunits/ceph-helpers.sh
ErwanAliasr1: This have been removed.

qa/workunits/ceph-helpers.sh
ErwanAliasr1: Done.

src/test/osd/osd-scrub-repair.sh
ErwanAliasr1: Fixed in my latest commit by prepending the output with the process id.

ErwanAliasr1:
I have an issue with one test, I'll fix it asap

src/test/ceph_objectstore_tool.py
dzafman: @ErwanAliasr1  It is fine to remove this sleep and the sleep(15) below too.

ErwanAliasr1: Thx @dzafman.

ErwanAliasr1:
On my side the PR is over & solved all the open issues. @ldachary if you want to make a final check on it, that's ok on my side.

ErwanAliasr1:
The branch have been rebased on the current master too.

ErwanAliasr1:
Rebased against current master tree.

dachary:
:+1: @ErwanAliasr1 could you please rebase one last time to get a clean make check ? I assume the failure does not come from your series.

ErwanAliasr1:
@dachary I'm still up to date with master, I can't rebase anymore

ErwanAliasr1:
@dachary Rebased against master to get fix from PR #8429 . That should be ok this way.

dachary:
Let's hope it's not tainted by http://tracker.ceph.com/issues/15375

PR #8133, opened 2016-03-15
packaging: move cephfs repair tools to ceph-common

ktdreyer:
The CephFS repair tools are generally useful to run on any node, not just the MDS nodes themselves.

Move the utilities from the ceph-mds package to the ceph-common package.

Fixes http://tracker.ceph.com/issues/15145

### DIFF
--- a/ceph.spec.in
+++ b/ceph.spec.in
@@ -974,6 +974,9 @@ rm -rf $RPM_BUILD_ROOT
 %{_bindir}/ceph-rbdnamer
 %{_bindir}/ceph-syn
 %{_bindir}/ceph-crush-location
+%{_bindir}/cephfs-data-scan
+%{_bindir}/cephfs-journal-tool
+%{_bindir}/cephfs-table-tool
 %{_bindir}/rados
 %{_bindir}/rbd
 %{_bindir}/rbd-replay
@@ -1050,9 +1053,6 @@ fi
 #################################################################################
 %files mds
 %{_bindir}/ceph-mds
-%{_bindir}/cephfs-journal-tool
-%{_bindir}/cephfs-table-tool
-%{_bindir}/cephfs-data-scan
 %{_mandir}/man8/ceph-mds.8*
 %if 0%{?_with_systemd}
 %{_unitdir}/ceph-mds@.service
--- a/ceph.spec.in
+++ b/ceph.spec.in
@@ -266,6 +266,7 @@ Summary:	Ceph Common
 Group:		System Environment/Base
 Requires:	librbd1 = %{epoch}:%{version}-%{release}
 Requires:	librados2 = %{epoch}:%{version}-%{release}
+Requires:	libcephfs1 = %{epoch}:%{version}-%{release}
 Requires:	python-rados = %{epoch}:%{version}-%{release}
 Requires:	python-rbd = %{epoch}:%{version}-%{release}
 Requires:	python-cephfs = %{epoch}:%{version}-%{release}
--- a/debian/ceph-common.install
+++ b/debian/ceph-common.install
@@ -7,6 +7,9 @@ usr/bin/ceph-dencoder
 usr/bin/ceph-rbdnamer
 usr/bin/ceph-syn
 usr/bin/ceph-crush-location
+usr/bin/cephfs-data-scan
+usr/bin/cephfs-journal-tool
+usr/bin/cephfs-table-tool
 usr/bin/rados
 usr/bin/rbd
 usr/bin/rbdmap
--- a/debian/ceph-mds.install
+++ b/debian/ceph-mds.install
@@ -1,5 +1,2 @@
 usr/bin/ceph-mds
-usr/bin/cephfs-journal-tool
-usr/bin/cephfs-table-tool
-usr/bin/cephfs-data-scan
 usr/share/man/man8/ceph-mds.8


ktdreyer:
cc @gregsfortytwo in case he wants to review

gregsfortytwo:
Looks fine to me, I guess?
That's about as much positive as I can say about packaging changes. :)

branto1:
LGTM, I'll wait for gitbuilders to finish the builds and merge. The jenkins build seems to have failed due to some changes in crushtool cli which should not be related to this commit in any way.

smithfarm:
@ktdreyer You say useful on "any node" - does that include clients? If only useful within the cluster, wouldn't it be better to put these in ceph-base?

ktdreyer:
Good question. My understanding from Greg is that this includes clients.

branto1:
You usually run fsck/repair tools on the computer (client/node) that mounts the file-system, I'm not 100% sure about it in this case, though.

gregsfortytwo:
These tools will generally require highly-capable keys (they need read/write access to CephFS' metadata pool[s]), but otherwise can be run from anywhere with cluster access, and you'll want to run multiple processes. I think ceph-common is an okay location for them but it is more of a convenience than a requirement so if we're trying to keep client-side installs slim maybe it's better-suited for the tools repo or something. (I've not heard of ceph-base before; it would probably work as well?)

smithfarm:
My vote would be for ceph-base, which is guaranteed to be installed on all server-side nodes. I agree to keep ceph-common slim. One selfish reason for that: ceph-common is part of SUSE Linux Enterprise (SLE) proper, as opposed to SUSE Enterprise Storage (SES). Anything that is to be added to SLE has to be justified why it is being added, so this creates extra work for me personally.

smithfarm:
(Not that I'm unwilling to take on the extra work - just want to make sure it is justified. Thanks!)

branto1:
I personally would vote for ceph-common. Going from the regular file-system, I'd definitely expect to be able to run fsck/repair tool from the computer that mounts the file-system.

smithfarm:
@branto1 That's a good point, but ceph-common is installed *everywhere* whereas fsck is a separate package.

gregsfortytwo:
I really don't have a dog in this show and am happy to leave it up to the packagers to work out — you guys have a better perspective on that than I do. :) Unless @jcsp cares I think you'll have to decide.

branto1:
@smithfarm: ceph-common already requires some of the cephfs bits but not all of them (which is imo quite weird and I have already noticed it previously) -- it requires python-cephfs but does not require libcephfs1. That would probably change with this commit, though as I suppose the tools are linked against libcephfs1 and rpmbuild will resolve the deps properly.

ktdreyer:
>  I suppose the tools are linked against libcephfs1 and rpmbuild will resolve the deps properly.

Yeah, although I think we would still want to require the precise N-V-R of libcephfs1 to be on the safe side.

smithfarm:
> Yeah, although I think we would still want to require the precise N-V-R of libcephfs1 to be on the safe side.

I am all for being on the safe side, but this is against openSUSE packaging policy, so I either have to ask for (and justify) a special exception or I have to drop Requires: lines.

This is currently a problem (for me) in libradosstriper1 and librbd1, both of which have a hard runtime dependency on librados2 this-and-this version.

The argument (as I understand it) is that shared libraries are supposed to be versioned using the shared libraries versioning mechanism, because a user might want to have two different versions installed at the same time.

ktdreyer:
> The argument (as I understand it) is that shared libraries are supposed to be versioned using the shared libraries versioning mechanism

Right, this is what I thought too, but it doesn't seem to always be the case in Ceph. It would be ideal if upstream always bumped the soname when there were breaking changes, but I've actually found errors in the field when the RPMs did not exactly match 100%. At a minimum it makes trying to trace with `-debuginfo` RPMs a nightmare when the versions differ slightly.

I think the problem is that Ceph's own ABI compat tests are very ad-hoc (it's basically "run ceph-test with a differing version in teuthology and see what breaks"). It would be ideal if we had a true ABI checker like https://www.sourceware.org/libabigail/ as a part of Teuthology.

smithfarm:
OK, I'm fine with whatever you guys decide. The "extra work" of putting this stuff into ceph-common isn't a big deal.

gregsfortytwo:
If our bad ABI compatibility testing is causing trouble for
packagers/downstream, maybe one of them would like to improve our ABI
compatibility testing to improve everybody's lives? ;)

(Seriously, most of the tech leads don't really know anything about
packaging so we're relying on you guys to tell us when it's broken and to
make the improvements!)

branto1:
@ktdreyer: It looks like this passed the git builders, we should probably also add the libcephfs1 with the precise N-V-R before merging though. Can you update this branch?

branto1:
Passed the gitbuilders again, merging.

smithfarm:
@ktdreyer, @gregsfortytwo: Tracking the ABI compatibility testing at http://tracker.ceph.com/issues/15160

smithfarm:
Given that we have split the server package, the logical next step would be to split the `ceph-common` package as well. Then we could have a package called, e.g., `cephfs-tools` containing cephfs tools only, and clients that are only interested in cephfs could install that instead of the entire `ceph-common` package as they must do now.

branto1:
@smithfarm: That might be a bit tricky as it would probably get messy (new dependencies?) and I'm not sure if it would help you with the packaging in any way -- we should probably make the new ceph-common require cephfs-tools, etc. Saying that, I'm not oppossed to that idea. I'm just not sure whether it is worth all the work.

PR #8134, opened 2016-03-15
msg/direct: DirectMessenger

mattbenjamin:
DirectMessenger is a dependency of libosd, which permits embedding of an OSD in other processes, such as an NFS-Ganesha or RGW server instance.

(I'm hoping for a Kraken milestone for this.)


yuyuyu101:
Makefile.am?

src/msg/direct/DirectMessenger.h
```diff
@@ -0,0 +1,82 @@
...
+
+  // unimplemented Messenger interface
+  void set_addr_unknowns(entity_addr_t &addr) {}
+  int get_dispatch_queue_len() { return 0; }
+  double get_dispatch_queue_max_age(utime_t now) { return 0; }
+  ceph::timespan get_dispatch_queue_max_age() {
+    return ceph::timespan(0); }
+  void set_cluster_protocol(int p) {}
+  int lazy_send_message(Message *m, const entity_inst_t& dst) { return EINVAL; }
+  int lazy_send_message(Message *m, Connection *con) { return EINVAL; }
```
yuyuyu101: any background on this api?

gregsfortytwo:
This is clearly not a "real" messenger (eg, it only allows one "direct peer" per messenger), so can you write some documentation about what is implemented and what guarantees are provided?

Amongst other things I'm concerned about the locking; it looks like message delivery happens basically in-line. Doesn't that mean any attempt at simultaneous delivery is likely to deadlock?

liewegas:
Can you explain how this will be used/useful?  It looks like there is only one peer.  How is that useful?  For example, we can't even drive an OSD with test traffic because it can't talk to a mon.

cbodley:
In the libosd use case, a DirectMessenger is used in addition to the existing messengers, and only sends messages between the library interface and the OSD. For example, an nfs data server could embed a ceph osd and use the DirectMessenger to deliver its requests in-process.

gregsfortytwo:
Based on what I've heard so far, it's not clear to me if we really want to pretend this is a Messenger. Perhaps a simple "Channel" concept or something would be clearer.

@mattbenjamin, you appear to have responded to my earlier comment by...editing my answer? (Or else I went a little bit schizophrenic and forgot about it.) Probably things will be clearer if you submit comments inline instead. ;)

cbodley:
@gregsfortytwo Given that the OSD pipeline is based on Messages, and the OSD needs a Messenger to send the replies back, we chose to use a Messenger for this hack. But I'm happy with whatever abstraction makes it easy to inject work and get responses.

Another use case to consider is what @adamemerson called LogicalOSD, where multiple OSDs are running out of a single process. We'll presumably want something that routes around the other Messengers to deliver messages between those OSDs, and a DirectMessenger that only supports two peers is probably not ideal for this.

gregsfortytwo:
@cbodley, did it actually take advantage of the messenger replies in the DirectMessenger? That would be a case for *sigh* eventually splitting up our Messenger interface into some kind of "Replyable" and proper "Messenger" interfaces.

I just am not very comfortable having a Messenger which doesn't satisfy *most* of the interface and locking rules of that interface, and hoping that we don't invoke any of the "bad" functions on it.

liewegas:
On Tue, 3 May 2016, Casey Bodley wrote:
> Another use case to consider is what @adamemerson called LogicalOSD, where multiple OSDs are running out of a single process. We'll presumably
> want something that routes around the other Messengers to deliver messages between those OSDs, and a DirectMessenger that only supports two
> peers is probably not ideal for this.

This is my concern, as it's the main use case I care about, and it seems 
unlikely that DirectMessenger is helpful there.

cbodley:
I believe this work started before the refactoring that moved send_message() to Connection. Message has a ConnectionRef, so replies are going through that.

Could we use DirectConnection without needing any of the Messenger pieces?

gregsfortytwo:
That's what I'm asking you. :p

I forgot we had Connection-based replies now, which certainly narrow the Messenger-requiring bits even more.

cbodley:
@gregsfortytwo Okay, I agree that DirectMessenger doesn't need to implement Messenger. Call it something like DirectChannel instead. It just needs to wrap a pair of DirectConnections, allow you to wire up the Dispatchers on each side, and provide a get_connection() to return the sending side

liewegas:
Closing this for now.  We can refactor it into a DirectConnection/Channel/Whatever when we have a user?

mattbenjamin:
I think you guys are reasoning via category mistake.  While Messenger the class isn't "doing much", "Messenger" is really a family of related abstractions: Messenger, Connection, Message, entity, Policy.  Rather than saving reasoning or implementation work, inventing a new Messenger-like abstraction that isn't a Messenger but shares Connection reduces logical clarity--using all the related interfaces in the usual way enforces consistency--removing it makes inconsistent behavior more likely.  As we discussed at ceph-next/vault, Channel in particular is an abstraction which we might someday want in -all- the Messengers, not just this one.  The situation is the same for OSD/Messenger federation--as soon as we have that, DirectMessenger will need the ability to direct Messages/OPs at all of the resident OSDs, not just one.  Casey pointed out that we might wish to enable logical connections to  specific OSDs--that's a good point, but it's a flex point trading added connection overhead for maybe some other benefit;  if we did want THAT, I'd suggest we call that a Channel and ensure that it works consistently with the rest of the Messenger system--probably by extending entity?

gregsfortytwo:
Yes, and that family of related functions has well (if not *clearly*) defined behaviors associated with each name. This class passes messages, but it is *not a Messenger*. For instance, it is very easy to construct a system which deadlocks this, which would not deadlock under any of the other Messengers.

That doesn't make it bad, or un-useful. But we need a better-defined interface than "some of the Messenger, but not most of the important bits". (I suggested "channel" as the first name in my head, but it does collide with the sub-messenger channels so it's probably a bad one.)

PR #8138, opened 2016-03-15
rbd: add support for mirror image promotion/demotion/resync

dillaman:


dillaman:
@jdurgin I quickly threw these CLI commands together since I need to reference them in the admin documentation.

src/librbd/internal.cc
```diff
@@ -2483,9 +2500,117 @@ int validate_pool(IoCtx &io_ctx, CephContext *cct) {
...
+      return r;
+    } else if (mirror_uuid == Journal<>::LOCAL_MIRROR_UUID) {
+      lderr(cct) << "image is already primary" << dendl;
+      return -EINVAL;
+    } else if (mirror_uuid != Journal<>::ORPHAN_MIRROR_UUID && !force) {
+      lderr(cct) << "image is still primary within a remote cluster" << dendl;
+      return -EBUSY;
+    }
+
+    r = Journal<>::allocate_tag(ictx, Journal<>::LOCAL_MIRROR_UUID);
```
jdurgin: I guess we might want an interlock with rbd-mirror here so that it doesn't have any i/o in flight to the local image after it's promoted, so it can be used after promotion immediately

dillaman: Yeah -- might be good for me to throw a TODO here

PR #8141, opened 2016-03-16
crush: fix the problem can not find the define item below the bucket

songbaisen:
crush: fix the problem can not find the define item below the bucket

Fixes: #15154

### DIFF
--- a/src/crush/CrushCompiler.cc
+++ b/src/crush/CrushCompiler.cc
@@ -815,6 +815,54 @@ void CrushCompiler::dump(iter_t const& i, int ind)
     dump(i->children.begin() + j, ind+1); 
 }
 
+/**
+*  This function fix the problem like below
+*   rack using_foo { item foo }  
+*   host foo { ... }
+*
+*  if an item being used by a bucket is defined after that bucket. 
+*  CRUSH compiler will create a map by which we can 
+*  not identify that item when selecting in that bucket.
+**/
+int CrushCompiler::adjust_bucket_item_place(iter_t const &i)
+{
+  map<string,set<string> > bucket_items;
+  map<string,iter_t> bucket_itrer;
+  vector<string> buckets;
+  for (iter_t p = i->children.begin(); p != i->children.end(); ++p) {
+    if ((int)p->value.id().to_long() == crush_grammar::_bucket) {
+      string name = string_node(p->children[1]);
+      buckets.push_back(name);
+      bucket_itrer[name] = p;
+      //skip non-bucket-item children in the bucket's parse tree
+      for (unsigned q=3; q < p->children.size()-1; ++q) {
+        iter_t sub = p->children.begin() + q;
+        if ((int)sub->value.id().to_long() 
+          == crush_grammar::_bucket_item) {
+          string iname = string_node(sub->children[1]);
+          bucket_items[name].insert(iname);
+        }         
+      }       
+    }     
+  }
+  
+  //adjust the bucket
+  for (unsigned i=0; i < buckets.size(); ++i) { 
+    for (unsigned j=i+1; j < buckets.size(); ++j) {
+      if (bucket_items[buckets[i]].count(buckets[j])) {
+        if (bucket_items[buckets[j]].count(buckets[i])) {
+          err << "bucket  '" <<  buckets[i] << "' and bucket '"
+          << buckets[j] << "' are included each other" << std::endl;
+          return -1; 
+        } else {  
+	   std::iter_swap(bucket_itrer[buckets[i]], bucket_itrer[buckets[j]]);
+        } 
+      } 
+    }
+  }
+	
+  return 0;
+}
 
 int CrushCompiler::compile(istream& in, const char *infn)
 {
@@ -880,7 +928,11 @@ int CrushCompiler::compile(istream& in, const char *infn)
 	<< " error: parse error at '" << line_val[line].substr(pos) << "'" << std::endl;
     return -1;
   }
-
+  
+  int r = adjust_bucket_item_place(info.trees.begin());
+  if (r < 0) {
+    return r;
+  }
   //out << "parsing succeeded\n";
   //dump(info.trees.begin());
   return parse_crush(info.trees.begin());
--- a/src/crush/CrushCompiler.h
+++ b/src/crush/CrushCompiler.h
@@ -52,8 +52,8 @@ class CrushCompiler {
   void find_used_bucket_ids(iter_t const& i);
   int parse_crush(iter_t const& i);  
   void dump(iter_t const& i, int ind=1);
-
   string consolidate_whitespace(string in);
+  int adjust_bucket_item_place(iter_t const &i);
 
 public:
   CrushCompiler(CrushWrapper& c, ostream& eo, int verbosity=0)
--- a/src/test/Makefile.am
+++ b/src/test/Makefile.am
@@ -71,6 +71,7 @@ check_SCRIPTS += \
 	test/cephtool-test-mds.sh \
 	test/cephtool-test-rados.sh \
        test/test_pool_create.sh \
+       test/test_crush_bucket.sh \
 	unittest_bufferlist.sh \
 	test/encoding/check-generated.sh \
 	test/mon/osd-pool-create.sh \
--- a/src/test/test_crush_bucket.sh
+++ b/src/test/test_crush_bucket.sh
@@ -1,0 +1,51 @@
+#!/bin/bash
+
+
+#Generic test_crush_bucket  test
+#
+
+# Includes
+source ../qa/workunits/ceph-helpers.sh
+function run() {
+    local dir=$1
+    shift
+
+    export CEPH_MON="127.0.0.1:17119"
+    export CEPH_ARGS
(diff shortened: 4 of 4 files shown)


songbaisen:
When we define the item below the bucket in the map.txt.Then we compile the map,txt

to the map.bin.It will error for cannnot find the item. It is very amazing for usr that

who have define it,but the ceph told him than can not find it.SO slove it!

liewegas:
Can you add a test case that demonstrates the bug?  We aren't follwoing what the problem is.  Thanks!

songbaisen:
@liewegas  Ok,SIr. I will add it later.

songbaisen:
@liewegas :Hello.Sir. I have add the test.Thank you!

songbaisen:
@liewegas  The unit test error as below looks like  the Jenkins bug.
FATAL: org.acegisecurity.providers.UsernamePasswordAuthenticationToken cannot be cast to

And somebody have have report this error to Jenkins  as below.
https://issues.jenkins-ci.org/browse/JENKINS-34775

src/crush/CrushCompiler.cc
```diff
@@ -815,6 +815,42 @@ void CrushCompiler::dump(iter_t const& i, int ind)
...
+  //adjust the bucket
+  for(unsigned i=0; i<buckets.size(); i++)
+    for(unsigned j=i+1; j<buckets.size(); j++) {
+      if(bucket_items[buckets[i]].count(buckets[j])) {
+        if(bucket_items[buckets[j]].count(buckets[i])) {
+          err << "bucket  '" <<  buckets[i] << "' and bucket '"<< buckets[j] << "' are included each other" << std::endl;
+          return -1; }
+        else {  
+	   std::iter_swap(bucket_itrer[buckets[i]],bucket_itrer[buckets[j]]);
+        }
```
liewegas: Please clean up upt eh whitespace:

* for ( not for(
* if ( not if(
* } else { not {\nelse {
* avoid long lines > 80 chars
* please include { } on the outer for loop
Thanks!

src/crush/CrushCompiler.cc
```diff
@@ -880,7 +916,12 @@ int CrushCompiler::compile(istream& in, const char *infn)
 	<< " error: parse error at '" << line_val[line].substr(pos) << "'" << std::endl;
     return -1;
   }
-
+  
+  int r = adjust_bucket_item_place(info.trees.begin());
+  if(r < 0)
+  {
```
liewegas: if (r < 0) {

src/crush/CrushCompiler.cc
songbaisen: @liewegas  Thank you sir.Done it. I will remember this!:smile:

src/crush/CrushCompiler.cc
```diff
@@ -815,6 +815,44 @@ void CrushCompiler::dump(iter_t const& i, int ind)
...
+  vector<string> buckets;
+  for (iter_t p = i->children.begin(); p != i->children.end(); p++) {
+    if ((int)p->value.id().to_long() == crush_grammar::_bucket) {
+      string name = string_node(p->children[1]);
+      buckets.push_back(name);
+      bucket_itrer[name] = p;
+      for (unsigned q=3; q < p->children.size()-1; q++) {
+        iter_t sub = p->children.begin() + q;
+        string tag = string_node(sub->children[0]);
+        if (tag == "item") {
```
tchaikov: could you compare the `id().to_long()` with `_bucket_item` also, instead of hardwiring it to "item"?

src/crush/CrushCompiler.cc
```diff
@@ -815,6 +815,44 @@ void CrushCompiler::dump(iter_t const& i, int ind)
     dump(i->children.begin() + j, ind+1); 
 }
 
+int CrushCompiler::adjust_bucket_item_place(iter_t const &i)
+{
+  map<string,set<string> > bucket_items;
+  map<string,iter_t> bucket_itrer;
+  vector<string> buckets;
+  for (iter_t p = i->children.begin(); p != i->children.end(); p++) {
```
tchaikov: use `++p` to non-primitive iterator.

src/crush/CrushCompiler.cc
```diff
@@ -815,6 +815,44 @@ void CrushCompiler::dump(iter_t const& i, int ind)
...
+    if ((int)p->value.id().to_long() == crush_grammar::_bucket) {
+      string name = string_node(p->children[1]);
+      buckets.push_back(name);
+      bucket_itrer[name] = p;
```
tchaikov: could add some comment noting why the children start at 3. for example, to skip non-bucket-item children in the bucket's parse tree. or, just enumerate all its children and note down "item" children.

src/crush/CrushCompiler.cc
```diff
@@ -815,6 +815,44 @@ void CrushCompiler::dump(iter_t const& i, int ind)
     dump(i->children.begin() + j, ind+1); 
 }
 
+int CrushCompiler::adjust_bucket_item_place(iter_t const &i)
```
tchaikov: could you elaborate a little bit what this function is doing in the comments? personally, i think "adjust" is quite vague.

tchaikov:
and you need to revise your commit message, 

```
Fixes: http://tracker.ceph.com/issues/15154
```

src/crush/CrushCompiler.cc
songbaisen: @tchaikov  Thank you sir for review! I will add some comments about this function.And the function name use "change" or "move" take place the "adjust"?

src/crush/CrushCompiler.cc
```diff
@@ -816,6 +816,48 @@ void CrushCompiler::dump(iter_t const& i, int ind)
...
+  for (iter_t p = i->children.begin(); p != i->children.end(); ++p) {
+    if ((int)p->value.id().to_long() == crush_grammar::_bucket) {
+      string name = string_node(p->children[1]);
+      buckets.push_back(name);
+      bucket_itrer[name] = p;
+      //skip non-bucket-item children in the bucket's parse tree
+      for (unsigned q=3; q < p->children.size()-1; ++q) {
+        iter_t sub = p->children.begin() + q;
+        if ((int)sub->value.id().to_long() == crush_grammar::_bucket_item) {
+          string iname = string_node(sub->children[1]);
```
songbaisen: Here is very strange.Use the command git show --color it does not look the same as here.

+        if ((int)sub->value.id().to_long() == crush_grammar::_bucket_item) {
+          string iname = string_node(sub->children[1]);
+          bucket_items[name].insert(iname);
+        }

songbaisen:
@tchaikov  Thank you sir! Done it! :+1: :+1: :+1: :tea: :beers:

src/crush/CrushCompiler.cc
```diff
@@ -816,6 +816,49 @@ void CrushCompiler::dump(iter_t const& i, int ind)
 }
 
 
+//Sometimes when defined the bucket above the item which the bucket
```
tchaikov: sorry, i don't follow you. could you put an example here?

src/crush/CrushCompiler.cc
```diff
@@ -816,6 +816,49 @@ void CrushCompiler::dump(iter_t const& i, int ind)
...
+      }       
+    }     
+  }
+  
+  //adjust the bucket
+  for (unsigned i=0; i < buckets.size(); ++i) { 
+    for (unsigned j=i+1; j < buckets.size(); ++j) {
+      if (bucket_items[buckets[i]].count(buckets[j])) {
+        if (bucket_items[buckets[j]].count(buckets[i])) {
+          err << "bucket  '" <<  buckets[i] << "' and bucket '"\
```
tchaikov: could you remove the '\', and indent the next line?

src/crush/CrushCompiler.cc
songbaisen: @tchaikov : OK sir.How about  the example like below?
before adjust                      after adjust
root default {                     host localhost {
  id -1                                 id -2          
  item localhost weight 0.000
}                                   }

host localhost {                   root default {
  id -2                                 id -1          
                                       item localhost weight 0.000
}                                    }

tchaikov: oic.

so the case is: 
```
rack using_foo { item foo }
host foo { ... }
```

if an item being used by a bucket is defined *after* that bucket. CRUSH compiler will create a map by which we can not identify that item when selecting in that bucket.

songbaisen: @tchaikov  Yes sir.Thank you!

liewegas:
-    export CEPH_MON="127.0.0.1:17109"
+    export CEPH_MON="127.0.0.1:17119"

(must be a unique port!)

songbaisen:
@liewegas  Thank you sir! Learn it. The mon must have unique port here is the reason that when begin unit test we will run a lot of unit script.And may be each script may be set up a mon. Is that right? If so where we can known which port we have used.Should we define a file let people know the max port have used.

liewegas:
They are all in a similar range.. you can 'git grep CEPH_MON= test' to see them all

songbaisen:
@liewegas  Oh Thank you sir! Learn it!

PR #8145, opened 2016-03-16
[RFC][BlueStore]Bluestore async read

majianpeng:
Implement async-read for KernelDevice. Using async-read mainly for wal-op.

### DIFF
--- a/src/os/bluestore/KernelDevice.cc
+++ b/src/os/bluestore/KernelDevice.cc
@@ -116,6 +116,11 @@ int KernelDevice::open(string p)
       goto out_fail;
     }
     size = s;
+
+    //For rotational disk, use async_read is better.
+    if (!g_conf->bluestore_wal_async_read && block_device_is_rotational(path.c_str()))
+      g_conf->set_val("bluestore_wal_async_read", "true");
+
   } else {
     size = st.st_size;
   }
--- a/src/os/bluestore/KernelDevice.cc
+++ b/src/os/bluestore/KernelDevice.cc
@@ -256,6 +256,15 @@ void KernelDevice::_aio_thread()
 		 << " ioc " << ioc
 		 << " with " << left << " aios left" << dendl;
 	assert(r >= 0);
+
+	std::map<uint64_t, bufferlist*>::iterator it =  ioc->aio_read_queue.find(aio[i]->offset);
+	if (it != ioc->aio_read_queue.end()) {
+	  bufferlist *bl = it->second;
+	  bl->clear();
+	  bl->claim_append(aio[i]->bl);
+	  --ioc->num_reading;
+	}
+
 	if (left == 0) {
 	  // check waiting count before doing callback (which may
 	  // destroy this ioc).
@@ -548,3 +557,22 @@ int KernelDevice::invalidate_cache(uint64_t off, uint64_t len)
   return r;
 }
 
+int KernelDevice::aio_read(uint64_t off, uint64_t len, bufferlist *pbl, IOContext *ioc, bool buffered)
+{
+  dout(20) << __func__ << " " << off << "~" << len << dendl;
+  assert(off % block_size == 0);
+  assert(len % block_size == 0);
+  assert(len > 0);
+  assert(off < size);
+  assert(off + len <= size);
+
+  _aio_log_start(ioc, off, len);
+  ioc->pending_aios.push_back(FS::aio_t(ioc, buffered ? fd_buffered : fd_direct));
+  ++ioc->num_pending;
+  ++ioc->num_reading;
+  assert(ioc->aio_read_queue.find(off) == ioc->aio_read_queue.end());
+  ioc->aio_read_queue[off] = pbl;
+  FS::aio_t& aio = ioc->pending_aios.back();
+  aio.pread(off, len);
+  return 0;
+}
--- a/src/os/bluestore/NVMEDevice.cc
+++ b/src/os/bluestore/NVMEDevice.cc
@@ -948,3 +948,10 @@ int NVMEDevice::invalidate_cache(uint64_t off, uint64_t len)
   dout(5) << __func__ << " " << off << "~" << len << dendl;
   return 0;
 }
+
+int NVMEDevice::aio_read(uint64_t off, uint64_t len, bufferlist *pbl,
+                     IOContext *ioc,
+                     bool buffered)
+{
+  return -EOPNOTSUPP;
+}
--- a/src/common/blkdev.cc
+++ b/src/common/blkdev.cc
@@ -171,6 +171,10 @@ bool block_device_support_discard(const char *devname)
   return get_block_device_int_property(devname, "discard_granularity") > 0;
 }
 
+bool block_device_is_rotational(const char *devname)
+{
+  return get_block_device_int_property(devname, "rotational") > 0;
+}
 int block_device_discard(int fd, int64_t offset, int64_t len)
 {
   uint64_t range[2] = {(uint64_t)offset, (uint64_t)len};
@@ -248,6 +252,11 @@ int get_device_by_uuid(uuid_d dev_uuid, const char* label, char* partition,
 {
   return -EOPNOTSUPP;
 }
+
+bool block_device_is_rotational(const char *devname)
+{
+  return -EOPNOTSUPP;
+}
 #elif defined(__FreeBSD__)
 #include <sys/disk.h>
 
@@ -274,6 +283,12 @@ int get_device_by_uuid(uuid_d dev_uuid, const char* label, char* partition,
 {
   return -EOPNOTSUPP;
 }
+
+bool block_device_is_rotational(const char *devname)
+{
+  return -EOPNOTSUPP;
+}
+
 #else
 int get_block_device_size(int fd, int64_t *psize)
 {
@@ -292,6 +307,10 @@ int block_device_discard(int fd, int64_t offset, int64_t len)
 
 int get_device_by_uuid(uuid_d dev_uuid, const char* label, char* partition,
 	char* device)
+{
+  return -EOPNOTSUPP;
+}
+bool block_device_is_rotational(const char *devname)
 {
   return -EOPNOTSUPP;
 }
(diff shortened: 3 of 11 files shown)


src/os/bluestore/KernelDevice.cc
```diff
@@ -548,3 +557,22 @@ int KernelDevice::invalidate_cache(uint64_t off, uint64_t len)
...
+  assert(len > 0);
+  assert(off < size);
+  assert(off + len <= size);
+
+  _aio_log_start(ioc, off, len);
+  ioc->pending_aios.push_back(FS::aio_t(ioc, buffered ? fd_buffered : fd_direct));
+  ++ioc->num_pending;
+  ++ioc->num_reading;
+  assert(ioc->aio_read_queue.find(off) == ioc->aio_read_queue.end());
+  ioc->aio_read_queue[off] = pbl;
```
yuyuyu101: is it possible exists overlap for aio_read_queue? aio_read same offset twice?

src/os/bluestore/KernelDevice.cc
```diff
@@ -256,6 +256,15 @@ void KernelDevice::_aio_thread()
...
 	assert(r >= 0);
+
+	std::map<uint64_t, bufferlist*>::iterator it =  ioc->aio_read_queue.find(aio[i]->offset);
+	if (it != ioc->aio_read_queue.end()) {
+	  bufferlist *bl = it->second;
+	  bl->clear();
+	  bl->claim_append(aio[i]->bl);
+	  --ioc->num_reading;
+	}
+
```
yuyuyu101: I think we need to add io type to FS::aio_t too. It looks like a little stupid to check aio_read_queue regardless the aio type.

src/os/bluestore/NVMEDevice.cc
```diff
@@ -948,3 +948,10 @@ int NVMEDevice::invalidate_cache(uint64_t off, uint64_t len)
   dout(5) << __func__ << " " << off << "~" << len << dendl;
   return 0;
 }
+
+int NVMEDevice::aio_read(uint64_t off, uint64_t len, bufferlist *pbl,
+                     IOContext *ioc,
+                     bool buffered)
+{
+  return -EOPNOTSUPP;
```
yuyuyu101: actually, NVMEDevice support aio_read is very cheap, just copy read impl and remove the last few lines.

src/os/bluestore/KernelDevice.cc
majianpeng: For that, i don't make sure.

src/os/bluestore/KernelDevice.cc
majianpeng: At firstly, i used this method which check read/write type. But 
In aio_write:
>>if (g_conf->bdev_inject_crash &&
>>	rand() % g_conf->bdev_inject_crash == 0) {
>>      derr << __func__ << " bdev_inject_crash: dropping io " << off << "~" << len
>>	   << dendl;
>>      // generate a real io so that aio_wait behaves properly, but make it
>>      // a read instead of write, and toss the result.
>>      aio.pread(off, len);
>>      ++injecting_crash;
In fact,  we can get io type from 'struct iocb".(see: http://compbio.cs.toronto.edu/repos/snowflock/xen-3.0.3/tools/libaio/src/libaio.h)
We can firstly get the type and then still need check aio_read_queue.

src/os/bluestore/KernelDevice.cc
```diff
@@ -116,6 +116,11 @@ int KernelDevice::open(string p)
       goto out_fail;
     }
     size = s;
+
+    //For rotational disk, use async_read is better.
+    if (!g_conf->bluestore_wal_async_read && block_device_is_rotational(path.c_str()))
```
yuyuyu101: why ssd isn't suitable for async read?

majianpeng: Are you mean the default is use async-read?

src/os/bluestore/KernelDevice.cc
```diff
@@ -548,3 +557,22 @@ int KernelDevice::invalidate_cache(uint64_t off, uint64_t len)
   return r;
 }
 
+int KernelDevice::aio_read(uint64_t off, uint64_t len, bufferlist *pbl, IOContext *ioc, bool buffered)
```
yuyuyu101: the other thing is buffered case, I'm not sure aio+buffered is ok to data consistence. aio_write will pick up different way regard to `buffered` flag.

majianpeng: For aio-write, i think the reason is the sync-method. But i'll check again.

majianpeng: @liewegas . For aio_write w/ buffer mode, why does not it use aio mode?

liewegas: aio only works with O_DIRECT. if you do it on a non-O_DIRECT it 
is actually synchronous (or errors out, i forget).

majianpeng: Thanks very much. I don't know this. I'll review the kernel code.
Can it change from buffer to direct for wal-read?

majianpeng: @liewegas . From this(http://lse.sourceforge.net/io/aio.html)
What Works?
>>AIO read and write on raw (and O_DIRECT on blockdev)
>>AIO read and write on files opened with O_DIRECT on ext2, ext3, jfs, xfs
What Does Not Work?
>>AIO read and write on files opened without O_DIRECT (i.e. normal buffered filesystem >>AIO). On ext2, ext3, jfs, xfs and nfs, these do not return an explicit error, but quietly default >>to synchronous or rather non-AIO behaviour (i.e io_submit waits for I/O to complete in these cases). For most other filesystems, -EINVAL is reported.

From the "What works". it need open w/ O_DIRECT for raw block.
But for "What Does Not Work",, it don't say for raw block w/o O_DIRECT.
Have you time to explain in detail? Thanks!

liewegas: On Wed, 16 Mar 2016, majianpeng wrote:
> @liewegas . From this(http://lse.sourceforge.net/io/aio.html)
> What Works?
> 
>             AIO read and write on raw (and O_DIRECT on blockdev)
>             AIO read and write on files opened with O_DIRECT on ext2, ext3, jfs, xfs
>             What Does Not Work?
>             AIO read and write on files opened without O_DIRECT (i.e. normal buffered filesystem >>AIO). On ext2, ext3, jfs, xfs
>             and nfs, these do not return an explicit error, but quietly default >>to synchronous or rather non-AIO behaviour (i.e
>             io_submit waits for I/O to complete in these cases). For most other filesystems, -EINVAL is reported.
> 
> From the "What works". it need open w/ O_DIRECT for raw block.
> But for "What Does Not Work",, it don't say for raw block w/o O_DIRECT.
> Have you time to explain in detail? Thanks!

It's not clear from that doc whether we would get EINVAL or synchronous 
behavior, but either way it isn't async.  Would need to dig through the 
code to know which, but it doesn't matter since it won't do what we 
want.

Doing a sync write into cache and then sync_file_range will do the right 
thing in all cases except where we have hit the dirty page limit, and I 
expect that will be hard to do if we are throttling our outstanding writes 
(current cap defaults to 200MB of memory from outstanding requests, IIRC).  
I don't think it will be a problem.

majianpeng: Yeah. Other question: 
a)for read, is is ok for async mode w/o O_DIRECT?
b)If a)not, can it change read-mode from buffer to direct for wal-read?

liewegas: > Yeah. Other question:
> a)for read, is is ok for async mode w/o O_DIRECT?

It will be synchronous.

> b)If a)not, can it change read-mode from buffer to direct for wal-read?

The WAL path currently has to be buffered because we are lazy about 
tracking multiples reads/writes to the same range in memory.

majianpeng: Thanks very much! sage

majianpeng:
@liewegas . I review _do_wal_op again and again. I still don't know " NOTE: we are doing all reads and writes buffered so that we can avoid worrying about multiple RMW cycles over the same blocks."
Can you give me a example about 'multiple RMW cycles over the same blocks"? For a transaction, i think no this case. Thanks!

liewegas:
On Mon, 21 Mar 2016, majianpeng wrote:
> @liewegas . I review _do_wal_op again and again. I still don't know " NOTE: we are doing all reads and writes buffered so that we can avoid
> worrying about multiple RMW cycles over the same blocks."
> Can you give me a example about 'multiple RMW cycles over the same blocks"? For a transaction, i think no this case. Thanks!

We may have a single transaction with two or more write calls, such 
that _do_write does a small WAL read/modify/write write over the same 
block.  If we don't use buffered mode, the second one will re-read the 
original block off disk and not see the effect of the first write.

To fix that we'd need to maintain a buffer cache of some sort (local to 
_do_wal_op would work).  Or we can just use buffered mode, like we do 
now...

majianpeng:
@liewegas . If i understand correctly, this case is you said: write(off=0, len=10k), write(off=10K, len=10k). The first write-op need read off=8K, the later write also need read off=8K. 
In fact, my patch also can do this(need modify) and remove the second read-op.
But this don't use for this case which: on traaction:write(off=0k, len=10k), The second transaction:write(off=10K, len=10k).

BTW: for OP_COPY & OP_ZERO, the read can use direct-mode. Is it ok?

yuyuyu101:
@majianpeng sage means we also need read in _write process.

From my view, it would be better we solve this like buffer result within transaction if we really want to get async read. And we also need more states to maintain async read process. Hmm, I hope no other problems if switching to async read.

In one word, it would be big refactor....

xiaoxichen:
Previously in filestore, synchronise read is never be a problem even for SSDs, not to say HDD.
So what's initiative of making it async?

majianpeng:
For wal-unaligned-write, it need read 4K data. My purpose put this read into state of STATE_AIO_WAIT rather than in STATE_KV_DONE.  Because for sync_wal_apply, this do by kv_sync_thread

liewegas:
Closing this because teh code has changed, and I think we want to have a discussion about how to do aio read in general before implementing it.  I'm kind of waiting for the other aio read change to land for filestore..

PR #8146, opened 2016-03-16
osd: try to flush/evict object from cache tier based on time period

david-z:
Currently cache tier's flushing/evicting triggering mechanism is based on dirty ratio and full dirty ratio. This is useful to control cache tier pool's space utilization. 

However one user case that we encountered is that customer doesn't care too much on such ratio towards their requirement. Customers want to move cold data to cheap base pool from upper expensive but small cache tier pool. They want to see only hot data staying at cache tier pool.

So we add such feature on cache tier with a configurable option, together with cache_min_flush_age, osd agent can flush objects which meet cache_min_flush_age to base pool periodically and then evict them.

Fixes: #15157

### DIFF
--- a/src/common/config_opts.h
+++ b/src/common/config_opts.h
@@ -573,6 +573,7 @@ OPTION(osd_agent_max_low_ops, OPT_INT, 2)
 OPTION(osd_agent_min_evict_effort, OPT_FLOAT, .1)
 OPTION(osd_agent_quantize_effort, OPT_FLOAT, .1)
 OPTION(osd_agent_delay_time, OPT_FLOAT, 5.0)
+OPTION(osd_agent_flush_evict_interval, OPT_U64, 0)
 
 // osd ignore history.last_epoch_started in find_best_info
 OPTION(osd_find_best_info_ignore_history_les, OPT_BOOL, false)
--- a/src/osd/ReplicatedPG.cc
+++ b/src/osd/ReplicatedPG.cc
@@ -11869,6 +11869,9 @@ bool ReplicatedPG::agent_work(int start_max, int agent_flush_quota)
     }
   }
 
+  // record latest flush/evict timestamp of this pg
+  info.history.last_flush_evict_stamp = ceph_clock_now(NULL);
+  
   if (++agent_state->hist_age > g_conf->osd_agent_hist_halflife) {
     dout(20) << __func__ << " resetting atime and temp histograms" << dendl;
     agent_state->hist_age = 0;
@@ -12188,6 +12191,7 @@ bool ReplicatedPG::agent_choose_mode(bool restart, OpRequestRef op)
   }
 
   {
+  utime_t now = ceph_clock_now(NULL);
   uint64_t divisor = pool.info.get_pg_num_divisor(info.pgid.pgid);
   assert(divisor > 0);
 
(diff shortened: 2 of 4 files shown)


l-mb:
Would it make sense to configure a cache_max_flush_age to correspond to cache_min_flush_age, to specify an age after which objects must be flushed?

david-z:
@l-mb 

In ceph, cache_min_flush_age means the minimum age of an object before the cache tiering agent flushes a recently modified (or dirty) object to the backing storage pool. So after this age, objects will be flushed. Is this you asking?

What I add "osd_agent_flush_evict_interval" is to trigger checking objects whether they need to be flushed or evicted.

l-mb:
@david-z That's not what cache_min_flush_age means, quite. An object older than this *may* be flushed, but it also may not be. (It essentially locks younger objects into the cache pool, unless it hits target_max_bytes/objects.)

A cache_max_flush_age could be useful for making sure that older objects are guaranteed to be written back in your usecase.

david-z:
@l-mb 

Pls check ceph's document for cache_min_flush_age: http://docs.ceph.com/docs/master/rados/operations/cache-tiering/#cache-age

I kinda understand what you are trying to say. Let us roughly go over cache tier's flushing logic first (evicting logic is similar):
1. when one pg's object number/bytes hits the dirty ratio (which is calculated by a simple formula, not just cache pool's target_max_bytes/objects), the agent's flush mode will be set to be active (low or high).
2. when the agent work is started, it will list some objects from objectstore. 
3. then agent will iterate each object. If agent's flush mode is not idle, here is active (low or hight) now, agent will run some checks on this object, one of which is to check whether this object meets cache_min_flush_age. If all the checks are passed, this object can be flushed.

What this PR does is to change flush mode to be active periodically no matter what dirty ratio is, then agent can get the chance to check object's age. I don't know why cache_max_flush_age is useful here, because if we don't change agent's flush mode, no object can be flushed. If agent's flush mode is changed to be active, cache_min_flush_age can do the work.

Pls correct me and point me the code piece if I am wrong. Thanks.

athanatos:
@david-z

I don't think this makes much sense.  First, I don't think that timestamp needs to go into the pg history unless the interval is meant to be a pretty long time.  Second, I think the entire point of this series is to make sure that we try to flush/evict at least once per osd_agent_flush_evict_interval.  The reason for this is to ensure that we get a chance to reduce the size of the cache pool even if it's not full based on object age?  I'm not sure I understand that.  Once the user has configured a cache tier, surely it's better to keep it close to the target size, right?  If the user wants a smaller target size, they can simply reduce that config?

I'm closing this for now.  If you disagree or think I've missed something, feel free to reopen, but I think it would help if you included a document in doc/dev/osd_internals explaining what the new config is meant to do, how to use it, and how it fits in with the other config options (doc/dev/osd_internals/cache_tiering.rst I'd think).

david-z:
@athanatos 

The reason for this is like you said, which is to ensure that we get a chance to reduce the size of the cache pool even if it's not full based on object age.

I understand your concern and question. For our case, the reason we use this feature, not simply reduce the config, is because we configured all SSDs as cache tier and we almost can't predict end-user's traffic, but we know end-user usually won't access their files again which are older than a certain time, e.g, 3 months. 

If end-user's traffic is very high some time and dirty objects reaches the target, then cache tier's default flushing/evicting will be triggered. If end-user's traffic is very low some time and the dirty ratio is not reached, we hope to try flushing/evicting those older files, then once high traffic is coming again, we have enough space at cache tier without doing flushing/evicting. 

I know this feature might be very tight to our use case, not common for every body else. So it is ok for now to close it.

PR #8147, opened 2016-03-16
osd/OSD.cc: finish full_map_request every MOSDMap message.

xiaoxichen:
We remember the range of requested full map in requested_full_first/last
and prevent sending duplicate requests.

But monitor will cap the reply to osd_map_message_max number of maps, for example,
OSD request [100, 200] while monitor only return [100,149], previous code think
[150, 200] is dup and prevent the OSD to send out the request, which is wrong.

Fix this by clear the requested_full_first/last field at the end of handle_osd_map.

Fixes: #15130

### DIFF
--- a/src/osd/OSD.cc
+++ b/src/osd/OSD.cc
@@ -4933,6 +4933,17 @@ void OSD::request_full_map(epoch_t first, epoch_t last)
   monc->send_mon_message(req);
 }
 
+void OSD::finish_full_map_request()
+{
+  if (requested_full_first == 0 && requested_full_last == 0)
+    return;
+  //Had requested some map but didn't receive in this message,
+  //This might because monitor capping the message to osd_map_message_max
+  dout(10) << __func__ << "still missing " << requested_full_first
+	   << ".." << requested_full_last << ", but now give up." << dendl;
+  requested_full_first = requested_full_last = 0;
+}
+
 void OSD::got_full_map(epoch_t e)
 {
   assert(requested_full_first <= requested_full_last);
@@ -6705,6 +6716,9 @@ void OSD::handle_osd_map(MOSDMap *m)
 
   // even if this map isn't from a mon, we may have satisfied our subscription
   monc->sub_got("osdmap", last);
+
+  if (!m->maps.empty())
+    finish_full_map_request();
 
   if (last <= superblock.newest_map) {
     dout(10) << " no new maps here, dropping" << dendl;
--- a/src/osd/OSD.h
+++ b/src/osd/OSD.h
@@ -2052,6 +2052,7 @@ protected:
   epoch_t requested_full_first, requested_full_last;
 
   void request_full_map(epoch_t first, epoch_t last);
+  void finish_full_map_request();
   void got_full_map(epoch_t e);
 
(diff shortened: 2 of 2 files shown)


xiaoxichen:
@liewegas ,would you mind take a review? this bug block us upgrading to jewel (development) release

src/osd/OSD.cc
```diff
@@ -6706,6 +6719,9 @@ void OSD::handle_osd_map(MOSDMap *m)
   // even if this map isn't from a mon, we may have satisfied our subscription
   monc->sub_got("osdmap", last);
 
+  if (!m->maps.empty())
```
tchaikov: @xiaoxichen i think this piece of code is repeating the `got_full_map()` if the monitor is not capping the returned MOSDMap.
```c++
  if (requested_full_first == requested_full_last) {
    dout(10) << __func__ << " " << e << ", requested " << requested_full_first
	     << ".." << requested_full_last
	     << ", now done" << dendl;
    requested_full_first = requested_full_last = 0;
```
can we reset the (requested_full_first, requested_full_last) only if the `last` is smaller than `requested_full_last`.

xiaoxichen: I put the logic inside the finish_full_map_request function, so if monitor is not capping the MOSDMap, the requested_full_first = requested_full_last =0, the log will not be printed out.

Or,

Whould you prefer to change the code like 

if ((requested_full_first <= requested_full_last && ! m->maps.empty() )
     finish_full_map_request();

Is this more clear?

tchaikov: ahh, i understand now. no, i think the current approach is better. it hides the `requested_full_first` and `requested_full_last` in *_full_map*() methods.

src/osd/OSD.cc
```diff
@@ -4933,6 +4933,19 @@ void OSD::request_full_map(epoch_t first, epoch_t last)
   monc->send_mon_message(req);
 }
 
+void OSD::finish_full_map_request()
+{
+  if (requested_full_first == 0 && requested_full_last ==0)
```
tchaikov: could you add spaces around '=='?

src/osd/OSD.cc
```diff
@@ -4933,6 +4933,19 @@ void OSD::request_full_map(epoch_t first, epoch_t last)
...
+    return;
+  //Had requested some map but didn't receive in this message,
+  //This might because monitor capping the message to  osd_map_message_max
+  dout(10) << __func__ <<" finished handle MOSDMap message, still missing "
```
tchaikov: add a space after "<<". and imho, this message could be more concise by removing "finished handle MOSDMap message" and " and finish this request.".

tchaikov:
aside from the formatting nit, the change looks good to me. @liewegas what do you think?

xiaoxichen:
@tchaikov comments addressed:) thanks.

xiaoxichen:
@liewegas Ping

athanatos:
The OSD side looks reasonable assuming that the mon behavior is to forget about the full map request after sending fewer maps than requested.

liewegas:
it is

PR #8148, opened 2016-03-16
xio: add prefix to xio msgr logs

roidayan:
This will help identify the msgr instance.
This is the same prefix style as the other msgrs.

Change-Id: I47aeaf5e62c5c728691affb67858bd791ea20ba9


mattbenjamin:
this looks fine to me--I'll ping for an ack, and hopefully merge

roidayan:
ping

PR #8149, opened 2016-03-16
test: rbd-mirror: add "switch to the next tag" test

trociny:


trociny:
@dillaman This test (when run as ceph_test_rbd_mirror) may reproduce the issue with the replay getting stuck I observe with rbd-mirror.

PR #8151, opened 2016-03-16
rbd-mirror: replayers async create/destroy

rjfd:
Implemented the TODOs left in Mirror.cc regarding the creation and destruction of Replayer objects in an asynchronous way.

### DIFF
--- a/src/tools/rbd_mirror/Mirror.cc
+++ b/src/tools/rbd_mirror/Mirror.cc
@@ -30,7 +30,8 @@ namespace mirror {
 Mirror::Mirror(CephContext *cct) :
   m_cct(cct),
   m_lock("rbd::mirror::Mirror"),
-  m_local(new librados::Rados())
+  m_local(new librados::Rados()),
+  m_replayers_lock("rbd::mirror::Mirror::replayers_map")
 {
   cct->lookup_or_create_singleton_object<Threads>(m_threads,
                                                   "rbd_mirror::threads");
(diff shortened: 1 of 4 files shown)


src/tools/rbd_mirror/Replayer.cc
```diff
@@ -44,50 +93,10 @@ Replayer::~Replayer()
...
-
-  // TODO: make interval configurable
-  m_pool_watcher.reset(new PoolWatcher(m_remote, 30, m_lock, m_cond));
-  m_pool_watcher->refresh_images();
-
-  m_replayer_thread.create("replayer");
-
-  return 0;
+  dout(20) << "init replaying for " << m_peer << dendl;
+  m_threads->work_queue->queue(new C_InitReplayer(this, on_finish), 0);
```
dillaman: This doesn't really make this async -- it just makes it synchronous on the work_queue thread.  I don't think it's going to be possible to make the connection to a remote cluster asynchronous.  Given that we should only be talking to 1 or 2 peers, it probably isn't that big of a deal.

src/tools/rbd_mirror/Mirror.cc
```diff
@@ -80,28 +90,42 @@ void Mirror::update_replayers(const map<peer_t, set<int64_t> > &peer_configs)
...
-  for (auto it = m_replayers.begin(); it != m_replayers.end();) {
+  m_replayers_lock.Lock();
+  for (auto it = m_replayers.begin(); it != m_replayers.end(); ++it) {
     peer_t peer = it->first;
     if (peer_configs.find(peer) == peer_configs.end()) {
-      dout(20) << "removing replayer for " << peer << dendl;
-      m_replayers.erase(it++);
-    } else {
-      ++it;
+      m_threads->work_queue->queue(new C_ShutdownReplayer(&m_replayers,
```
dillaman: To make the shutdown async, you would need to add a new "shut_down" method to Replayer which takes a context.  It would store the context, set the "stopping" flag, and when the thread exits, it would invoke the callback.  Again, given that we will only have one or two peers, I am not sure it is worth the immediate effort.

rjfd:
I agree with both your comments. I wrote this code to learn more about the asynchronous mechanisms used in Ceph, but I guess I failed on that. So, If I need to run some code asynchronously in Ceph, what is the boilerplate code that I should use? 

Also, I will later close this PR.

dillaman:
The big thing is that you need to avoid any synchronous methods.  In general, the convention is that if a method is async, it generally takes a "Context *" callback to invoke when complete.

PR #8152, opened 2016-03-16
vstart: use ip instead of ifconfig utility

theanalyst:
Making vstart not use ifconfig which may not be supported in all
distributions

### DIFF
--- a/src/vstart.sh
+++ b/src/vstart.sh
@@ -390,7 +390,7 @@ if [ -n "$ip" ]; then
 else
     echo hostname $HOSTNAME
     # filter out IPv6 and localhost addresses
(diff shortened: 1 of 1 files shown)


wjwithagen:
@theanalyst 
That is definitely not going to fly on FreeBSD since it does not have ip(1).
I'd prefer to keep ifconfig since that is closest to what I would expect to be available on just about any decent Linux/Unix platform....

theanalyst:
@wjwithagen yeah I kind of got that after I submitted the patch.. I guess nearly everybody has ifconfig (even though it may be in sbin) I'll close this

wjwithagen:
On 16-3-2016 14:15, Abhishek L wrote:
> Closed #8152 <https://github.com/ceph/ceph/pull/8152>.

FreeBSD does have it in /sbin, indeed.

--WjW

PR #8156, opened 2016-03-16
osd/PG: indicate in pg query output whether ignore_history_les would help

liewegas:
Put a hint in the PG query output if setting the
osd_find_best_info_ignore_history_les = true would allow peering to
proceed.  This is difficult to diagnose even for a developer (you need
to read the peer info *very* carefully) but is easy to hint.

### DIFF
--- a/src/osd/PG.cc
+++ b/src/osd/PG.cc
@@ -970,8 +970,9 @@ PG::Scrubber::~Scrubber() {}
  *  3) Prefer current primary
  */
 map<pg_shard_t, pg_info_t>::const_iterator PG::find_best_info(
-  const map<pg_shard_t, pg_info_t> &infos) const
+  const map<pg_shard_t, pg_info_t> &infos, bool *history_les_bound) const
 {
+  assert(history_les_bound);
   /* See doc/dev/osd_internals/last_epoch_started.rst before attempting
    * to make changes to this process.  Also, make sure to update it
    * when you find bugs! */
@@ -982,6 +983,7 @@ map<pg_shard_t, pg_info_t>::const_iterator PG::find_best_info(
        ++i) {
     if (!cct->_conf->osd_find_best_info_ignore_history_les &&
 	max_last_epoch_started_found < i->second.history.last_epoch_started) {
+      *history_les_bound = true;
       max_last_epoch_started_found = i->second.history.last_epoch_started;
     }
     if (!i->second.is_incomplete() &&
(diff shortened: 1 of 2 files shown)


liewegas:
@athanatos what do you think?  this would have helped me earlier this week with a cluster i was rescuing

src/osd/PG.cc
```diff
@@ -5975,6 +5978,18 @@ boost::statechart::result PG::RecoveryState::Peering::react(const QueryState& q)
   }
   q.f->close_section();
 
+  if (history_les_bound) {
+    q.f->open_object_section("peering_blocked_by_options");
```
dzafman: Even though you warn about it, I'm nervous calling this "peering_blocked_by_options."  Someone might assume that they should leave osd_find_best_info_ignore_history_les permanently set to true.

I wish the option were called "osd_find_best_info_ignore_history_les_and_lose_data_i_really_really_mean_it" or invert it as "osd_find_best_info_check_data_integrity_history_les" defaulting to true.

liewegas: On Wed, 16 Mar 2016, David Zafman wrote:
> 
> In src/osd/PG.cc:
> 
> > @@ -5975,6 +5978,18 @@ boost::statechart::result PG::RecoveryState::Peering::react(const QueryState& q)
> >    }
> >    q.f->close_section();
> >  
> > +  if (history_les_bound) {
> > +    q.f->open_object_section("peering_blocked_by_options");
> 
> Even though you warn about it, I'm nervous calling this "peering_blocked_by_options." Someone might assume that they should leave
> osd_find_best_info_ignore_history_les permanently set to true.
> 
> I wish the option were called "osd_find_best_info_ignore_history_les_and_lose_data_i_really_really_mean_it" or invert it as
> "osd_find_best_info_check_data_integrity_history_les" defaulting to true.

Yeah... we could not mention the option, maybe?  Although the field that 
mentions the option also mentioned that you will likely lose data.  But

athanatos:
<sjusthm> sage: assert that the pointer is non-null at the top or check for null before assigning
<sjusthm> but it looks entirely reasonable to me
<sjusthm> david's point is valid, it's a dangerous option...
* mx (~myeho@66.193.98.66) has joined #ceph-devel
<sjusthm> you could make it something like peering_blocked_by_details
<sjusthm> with peering_blocked_by_history_les_bound

src/osd/PG.cc
```diff
@@ -970,7 +970,7 @@ PG::Scrubber::~Scrubber() {}
  *  3) Prefer current primary
  */
 map<pg_shard_t, pg_info_t>::const_iterator PG::find_best_info(
-  const map<pg_shard_t, pg_info_t> &infos) const
+  const map<pg_shard_t, pg_info_t> &infos, bool *history_les_bound) const
```
athanatos: Assert that the pointer is non-null at the top or check for null before assigning -- someone's going to try to call this with a nullptr on the assumption that it's optional.

src/osd/PG.cc
athanatos: Probably best not to mention the option (it's kludgy anyway).  Maybe make it say:

    peering_blocked_by_details : [
        {
            detail: peering_blocked_by_history_les_bound
        }
    ]
}

PR #8157, opened 2016-03-16
[DNM] rgw: replace rest conn parameter list with vector

kmroz:


kmroz:
Goes a bit deeper... closing for now.

PR #8159, opened 2016-03-16
rbd-mirror: use pool/image names in asok commands

trociny:
Now the commands look similar to rbd cache asok commands.


jdurgin:
lgtm

dillaman:
lgtm

PR #8160, opened 2016-03-16
os/bluestore/NVMEDevice: make IO thread using dpdk launch

yuyuyu101:
Previously we will call rte_eal_init in caller thread as well as dpdk
resources will init in that thread. It will cause all threads spawned
later will bind to the specified core.

And io thread will be spawned via dpdk api, otherwise, dpdk memory
has huge performance degraded problem.

### DIFF
--- a/src/os/bluestore/NVMEDevice.cc
+++ b/src/os/bluestore/NVMEDevice.cc
@@ -21,6 +21,7 @@
 #include <fcntl.h>
 #include <unistd.h>
 #include <map>
+#include <thread>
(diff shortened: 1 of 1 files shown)


yuyuyu101:
I think we should make this target to Jewel since this PR fix the serve performance problem caused by incorrect thread spawn :-(

yuyuyu101:
Although it's just a experiment feature, it should be better for users to try this.

liewegas:
yep

PR #8162, opened 2016-03-16
rbd-mirror: use the mirroring directory to detect candidate images

rjfd:
Fixes #15142

Ticket url: http://tracker.ceph.com/issues/15142


dillaman:
lgtm -- should merge with #8204

PR #8163, opened 2016-03-16
osd/ReplicatedPG: clear watches on change after applying repops

liewegas:
apply_and_flush_repops does any in-flight request side-effects,
including setting up new watches.  Clear out watches after that
so we don't leave any behind.

Fixes: #15151


liewegas:
I was able to quickly reproduce the original crash w/ a vstart cluster, but can't after a couple hours with the fix.

athanatos:
lgtm

PR #8164, opened 2016-03-16
rgw: handle initial slashes properly in BulkDelete of Swift API.

rzarzynski:


rzarzynski:
@oritwas: I've pushed the amended commit a minute ago. There are absolutely no changes in the code. Only the `Fixes` tag has been added in commit message to make backporting easier.

PR #8166, opened 2016-03-16
rgw: rest and http client code to use param vectors

kmroz:
Trying to knock down a //TODO item, in replacing the use of a list for a vector in some rgw REST parameter lists.

Since we're only inserting to the back of the parameter list, a vector should be a more efficient choice for a container structure.

### DIFF
--- a/src/rgw/rgw_admin.cc
+++ b/src/rgw/rgw_admin.cc
@@ -1345,7 +1345,7 @@ static int send_to_url(const string& url, const string& access,
   key.id = access;
   key.key = secret;
 
-  list<pair<string, string> > params;
+  param_vec_t params;
   RGWRESTSimpleRequest req(g_ceph_context, url, NULL, &params);
 
   bufferlist response;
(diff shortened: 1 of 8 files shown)


cbodley:
cool, thanks for taking this on!

the patch looks good in general. the only part that sticks out is that rgw_rest_client.h doesn't have access to these typedefs, and has to know that it's a vector.

would you mind moving param_pair_t and param_vec_t into rgw_rest_client.h while you're at it, so we can use param_vec_t in rgw_rest_client.* and rgw_admin.cc?

kmroz:
@cbodley - Thanks for the review. Good suggestion on location of param_pair_t and param_vec_t. Testing something now.

kmroz:
@cbodley - I ended up moving param_pair_t and param_vec_t one level higher to rgw_http_client.h. That way I could also cleanly use vectors for the various header lists inherited from RGWHTTPClient.
This could certainly use a run through some test infra :)

cbodley:
looks good, thanks. this may have to wait a bit though - with the jewel release coming up, we'll only be merging bug fixes into master

kmroz:
Cool, thanks. Waiting until after jewel is released makes total sense.

kmroz:
@cbodley - Needed a rebase.

kmroz:
@cbodley - In your opinion, anything else needed here? Thanks.

cbodley:
ran this through s3tests without regression. thanks again for taking the initiative on this one, and sorry for the delays in merging!

kmroz:
@cbodley - No worries, thanks for getting back to this.

PR #8168, opened 2016-03-16
Rgw ldap fixes

mattbenjamin:


src/rgw/Makefile.am
```diff
@@ -92,6 +91,10 @@ librgw_la_SOURCES = \
 	rgw/rgw_xml_enc.cc \
 	rgw/rgw_website.cc
 
+if WITH_OPENLDAP
+librgw_la_SOURCES += rgw/rgw_ldap
```
yehudasa: @mattbenjamin rgw/rgw_ldap.cc?

src/rgw/rgw_ldap.h
```diff
@@ -80,7 +82,35 @@ namespace rgw {
...
+
+  class LDAPHelper
+  {
+  public:
+    LDAPHelper(std::string _uri, std::string _binddn, std::string _searchdn,
+	      std::string _dnattr)
+      {}
+
+    int init() {
+      return -EINVAL;
```
yehudasa: -ENOTSUPP maybe?

src/rgw/Makefile.am
mattbenjamin: @yehudasa er, yeah

(it works as is because rgw_ldap.cc has no text--but I left it in place for future)

repushed

src/rgw/rgw_ldap.h
mattbenjamin: ok, I forget that's a real error sometimes

PR #8170, opened 2016-03-16
Wip rgw sync fixes 3

yehudasa:


src/rgw/rgw_http_client.cc
```diff
@@ -632,9 +632,12 @@ void *RGWHTTPManager::reqs_thread_entry()
...
   RWLock::WLocker rl(reqs_lock);
-  map<uint64_t, rgw_http_req_data *>::iterator iter = reqs.begin();
-  for (; iter != reqs.end(); ++iter) {
-    _finish_request(iter->second, -ECANCELED);
+  set<rgw_http_req_data *> all_reqs;
+  for (auto iter : reqs) {
+    all_reqs.insert(iter.second); /* can't call _finish_request() on iter.second, it will modify reqs */
+  }
+  for (auto iter : all_reqs) {
+    _finish_request(iter, -ECANCELED);
```
cbodley: good catch! i'll pull this into my testing branch

though it might be simpler to do something like this:
```
while (!reqs.empty()) {
  _finish_request(reqs.begin()->second, -ECANCELED);
}
```
alternatively, you could move reqs into a temporary and loop over that:
```
auto all_reqs = std::move(reqs); // prevent _finish_request() from removing reqs during the loop
for (auto iter : all_reqs) {
```

cbodley:
With these fixes, I've been unable to reproduce the segfaults and multisite tests are passing consistently.

src/rgw/rgw_http_client.cc
yehudasa: @cbodley yeah, seems more elegant. I wouldn't go for the `while (!reqs.empty())` solutions, because it relies on _finish_request() to do the right thing (which doesn't remove it directly, but calls _complete_request()).

yehudasa:
@cbodley reworked commit following your comment, added another fix to a separate issue

cbodley:
lgtm

PR #8171, opened 2016-03-16
qa/workunits/rados/test-upgarde-from-9.2.sh

liewegas:
Skip certain tests when upgrading from 9.2 to x (jewel).


liewegas:
see also https://github.com/ceph/ceph-qa-suite/pull/886

PR #8172, opened 2016-03-16
test: rbd-mirror: compare positions using all fields

trociny:
Comparing by only entry_tid may give false positives.

### DIFF
--- a/qa/workunits/rbd/rbd_mirror.sh
+++ b/qa/workunits/rbd/rbd_mirror.sh
@@ -216,21 +216,14 @@ get_position()
     local image=$1
     local id_regexp=$2
 
-    # Parse line like below, looking for the first entry_tid
+    # Parse line like below, looking for the first position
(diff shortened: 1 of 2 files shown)


dillaman:
@trociny One of the tasks I want to get to next week is properly recording ownership within the journal tags to support failover/failback.  For simple tests, I think the tag_tids will always align between two clients -- but for more complicated tests, I would expect them to diverge.  For example, a peer that attaches to a remote journal after many exclusive locks (i.e. tag allocations) would result in a high tag tid on the remote side aligning with a small tag_tid on the local side.

trociny:
@dillaman I guess what you describe is possible when comparing positions in different journals. But here I am comparing positions for two clients in the same journal. Shouldn't they always match exactly (object_number, tag_tid, entry_tid) if the positions are equal? They both pointing to the same event in the journal.

dillaman:
Within the same journal, the tags should match. In that case, lgtm.

PR #5096, opened 2015-06-27
common/buffer: add bufferlist::const_iterator

tchaikov:
* unlike its cousin, `bufferlist::const_iterator` is not exported using
  CEPH_BUFFER_API. will do it once we think it will have external users.

Signed-off-by: Kefu Chai <kchai@redhat.com>

PR #8173, opened 2016-03-16
buffer: fix iterator_impl visibility through typedef

noahdesu:
The following program doesn't compile because of symbol visibility issues.
While bufferlist::iterator is a class implementation with visibility specified,
it is unclear after google-fu how to do the same through typedef.
```c++
int main()
{
 ceph::bufferlist bl;
 ceph::bufferlist::const_iterator it = bl.begin();
 (void)it;
 return 0;
}
```
[nwatkins@bender ~]$ g++ -Wall -std=c++11 -Iinstall/include -Linstall/lib -o test test.cc -lrados
/tmp/cciR9MUj.o: In function `main':
test.cc:(.text+0x43): undefined reference to `ceph::buffer::list::iterator_impl<true>::iterator_impl(ceph::buffer::list::iterator const&)'
/usr/bin/ld: test: hidden symbol `_ZN4ceph6buffer4list13iterator_implILb1EEC1ERKNS1_8iteratorE' isn't defined
/usr/bin/ld: final link failed: Bad value

Signed-off-by: Noah Watkins <noahwatkins@gmail.com>

### DIFF
--- a/src/include/buffer.h
+++ b/src/include/buffer.h
@@ -57,10 +57,8 @@
 
 #if __GNUC__ >= 4
   #define CEPH_BUFFER_API  __attribute__ ((visibility ("default")))
-  #define CEPH_BUFFER_DETAILS __attribute__ ((visibility ("hidden")))
 #else
   #define CEPH_BUFFER_API
-  #define CEPH_BUFFER_DETAILS
 #endif
 
 #if defined(HAVE_XIO)
@@ -270,7 +268,7 @@ namespace buffer CEPH_BUFFER_API {
 
   private:
     template <bool is_const>
-    class CEPH_BUFFER_DETAILS iterator_impl
+    class CEPH_BUFFER_API iterator_impl
       : public std::iterator<std::forward_iterator_tag, char> {
     protected:
       typedef typename std::conditional<is_const,


noahdesu:
This might need to become a tracker ticket... While this patch fixes my use case, it looks like it makes the jenkins build blow up with bufferlist iterator errors.

jdurgin:
cc @tchaikov

tchaikov:
@noahdesu this is expected. see #5096. i intended to be more conservative when it came to symbol visibility. if we need to offer the librados client the `buffer::list::const_iterator`. we need to

- replace the `CEPH_BUFFER_DETAILS` specifier of `iterator_impl` with `CEPH_BUFFER_API`. as `buffer::list::const_iterator` inherit quite a few of its symbols from this tempate
- and that's all =)

noahdesu:
@tchaikov sounds good to me! the way I ran into this is that in Google gtest framework if one builds an assertion `ASSERT_EQ(bl1, bl2);` then gtest will compile that as `const ceph::buffer::list::const_iterator it = bl1.begin()`.

tchaikov:
@noahdesu i see, because `ASSERT_EQ(bl1, bl2)` tries to print the bufferlist. as gtest's default printer detects that buffer::list is a STL container with iterator and const_iterator defined. (see  `IsContainerTest` and `DefaultPrintTo()` ), and it will iterate through the bufferlist and dereference the const_iterator and print the value of the elements in bufferlist. if the `ASSERT_EQ` fails.

the unittest_bufferlist compiles and runs fine with following patch
```diff
diff --git a/src/test/bufferlist.cc b/src/test/bufferlist.cc
index 008b8cd..a55ea8b 100644
--- a/src/test/bufferlist.cc
+++ b/src/test/bufferlist.cc
@@ -2604,6 +2604,7 @@ TEST(BufferList, compare) {
   ASSERT_FALSE(a == ab);
   ASSERT_FALSE(ac == ab);
   ASSERT_TRUE(ab == ab);
+  ASSERT_EQ(ab, ab);
 }
```

could you show me a minimal reproducer? but i have no strong objection to unlock the `buffer::list::const_iterator` though.

noahdesu:
The code you have in the comparison test above `ASSERT_EQ(ab, ab)` is what I was using to reproduce the problem. My test is running outside of the Ceph tree against an installed version of librados. Is it possible that the visibility issue isn't showing up for in-tree tests?

tchaikov:
>  Is it possible that the visibility issue isn't showing up for in-tree tests?

yes. seems librados hides all invisible symbols when it is being linked against libcommon.a which contains bufferlist.o

and yes, i am running `make unittest_bufferlist` against the ceph source tree. guess that's the way we build this unittest.

if you want to export `buffer::list::const_iterator`, i am okay with this decision. but we need another set of eyes on it.

jdurgin:
I've no problem with exposing a const iterator. Just need to be careful about the api exposed, like any public method that we can't break.

noahdesu:
@tchaikov how's that look? i'm not really up to speed on all of the api breakage issues.

tchaikov:
@noahdesu could you remove the definition of "CEPH_BUFFER_DETAILS" along with this change, as it is not used elsewhere. aside from this, lgtm.

noahdesu:
Cool, all fixed.

tchaikov:
lgtm. nit: we can improve the commit message title a little bit: "buffer: expose the buffer::list::const_iterator".

smithfarm:
Created tracker issue and prepended its URL to the PR description.

tchaikov:
thanks Nathan!

Nathan Cutler <notifications@github.com>于2016年8月3日 周三13:44写道：

PR #8174, opened 2016-03-16
mon/MonClient: make authenticate timeout more reliable

liewegas:
The LibRadosMiscConnectFailure.ConnectFailure test occasionally
fails with a .000001 timeout.  I'm pretty sure it's because this
cond always manages a signal queued by the time it goes to wait,
or gets a signal before the deadline.

Fix by adding an outter check as well so that we'll error out
regardless of the queued signal vs timeout check order in the
cond.

This is my best theory as to what is going on, at least... I don't
see any other way we'd get a success with the timeout set.

Fixes: #13392

### DIFF
--- a/src/mon/MonClient.cc
+++ b/src/mon/MonClient.cc
@@ -461,7 +461,11 @@ int MonClient::authenticate(double timeout)
     ldout(cct, 10) << "authenticate will time out at " << until << dendl;
   while (state != MC_STATE_HAVE_SESSION && !authenticate_err) {
     if (timeout > 0.0) {
-      int r = auth_cond.WaitUntil(monc_lock, until);
+      utime_t now = ceph_clock_now(cct);
+      int r = ETIMEDOUT;
+      if (now < until) {
+	r = auth_cond.WaitUntil(monc_lock, until);
+      }
       if (r == ETIMEDOUT) {
 	ldout(cct, 0) << "authenticate timed out after " << timeout << dendl;
 	authenticate_err = -r;


liewegas:
passed rados run, although the original bug was pretty hard to hit, so that doesn't mean it's fixed.

src/mon/MonClient.cc
```diff
@@ -461,7 +461,11 @@ int MonClient::authenticate(double timeout)
     ldout(cct, 10) << "authenticate will time out at " << until << dendl;
   while (state != MC_STATE_HAVE_SESSION && !authenticate_err) {
     if (timeout > 0.0) {
-      int r = auth_cond.WaitUntil(monc_lock, until);
+      utime_t now = ceph_clock_now(cct);
+      int r = ETIMEDOUT;
+      if (now < until) {
```
tchaikov: in http://tracker.ceph.com/issues/13992, 

```c++
  ASSERT_NE(0, rados_connect(cluster));
  ASSERT_NE(0, rados_connect(cluster));
```
the monclient gets authenticated sooner than we expect. but the second test failed. i think there is a race condition that the first 

> it's because this cond always manages a signal queued by the time it goes to wait, or gets a signal before the deadline.

but i am not sure this is what happened in http://tracker.ceph.com/issues/13992

1. _reopen_session() in MonClient::authenticate() // with monc_lock
1. while (state != MC_STATE_HAVE_SESSION && !authenticate_err) :
1. time reaches "until"
1. WaitUntil(until)
  1. it release the monc_lock and start waiting on the condition
1. ms_dispatch()
  1. acquires the monc_lock
  1. handle_auth() change the state and wakeup the waiters

- because the ms_dispatch() can not signal the condition variable without acquiring the lock, i think the signal can not be fired before `authenticate()` starts to wait. even if the condition is signalled before we start to wait, the `pthread_cond_timedwait()` would return `ETIMEDOUT`, because it's edge triggered.


since MonClient is using conf->client_mount_timeout for the timeout in seconds, maybe we can just use "0.0000000001" (1e-9) instead of 0.000001 (1e-6) as the timeout?

tchaikov:
and need to s/13392/13992/ in the commit message.

liewegas:
https://github.com/ceph/ceph/pull/8209

PR #8175, opened 2016-03-17
osd: add cmpext and writesame ops

mikechristie:
The following patches made over the ceph master branch
implement OSD side support for VMware VAAI's Atomic Test
and Set (ATS) and Write Same (Zero) requests.

ATS is used for operations like locking and heartbeats. It
is implemented by as the SCSI COMPARE_AND_WRITE command which
requires the device to read N blocks, compare them to data
sent with the command, and if equal, write N blocks.

Zero is used to initialize blocks to zero. It is implemented
as the SCSI WRITE_SAME command which passes the device a
block's worth of data and has it write it multiple times.

More info on VAAI can be found here:
http://www.vmware.com/resources/techresources/10337

The krbd patches which use these requests are in vaai branch of
this tree:
https://github.com/mikechristie/linux-kernel

I did not submit them in this thread, because they depend on other
patches that are still being reviewed upstream and I did not want
to waste people's time reviewing them if they change. These OSD side
patches should be ok to review and merge, because the op format 
and implementation should not change.

v3
- Add CMPEXT miscompare offset back.
v2
- Integrated David's tracing fixes.
- Dropped CMPEXT miscompare offset.

### DIFF
--- a/src/osd/ReplicatedPG.cc
+++ b/src/osd/ReplicatedPG.cc
@@ -3655,6 +3655,39 @@ int ReplicatedPG::do_xattr_cmp_str(int op, string& v1s, bufferlist& xattr)
   }
 }
 
+int ReplicatedPG::do_extent_cmp(OpContext *ctx, OSDOp& osd_op)
+{
+  ceph_osd_op& op = osd_op.op;
+  vector<OSDOp> read_ops(1);
+  OSDOp& read_op = read_ops[0];
+  int result = 0;
+
+  read_op.op.op = CEPH_OSD_OP_SYNC_READ;
+  read_op.op.extent.offset = op.extent.offset;
+  read_op.op.extent.length = op.extent.length;
+  read_op.op.extent.truncate_seq = op.extent.truncate_seq;
+  read_op.op.extent.truncate_size = op.extent.truncate_size;
+
+  result = do_osd_ops(ctx, read_ops);
+  if (result < 0) {
+    derr << "do_extent_cmp do_osd_ops failed " << result << dendl;
+    return result;
+  }
+
+  if (read_op.outdata.length() != osd_op.indata.length())
+    return -EINVAL;
+
+  for (uint64_t p = 0; p < osd_op.indata.length(); p++) {
+    if (read_op.outdata[p] != osd_op.indata[p]) {
+      ::encode(p, osd_op.outdata);
+      osd_op.outdata.claim_append(read_op.outdata);
+      return -EILSEQ;
+    }
+  }
+
+  return 0;
+}
+
 int ReplicatedPG::do_writesame(OpContext *ctx, OSDOp& osd_op)
 {
   ceph_osd_op& op = osd_op.op;
@@ -4158,6 +4191,11 @@ int ReplicatedPG::do_osd_ops(OpContext *ctx, vector<OSDOp>& ops)
     switch (op.op) {
       
       // --- READS ---
+
+    case CEPH_OSD_OP_CMPEXT:
+      tracepoint(osd, do_osd_op_pre_extent_cmp, soid.oid.name.c_str(), soid.snap.val, oi.size, oi.truncate_seq, op.extent.offset, op.extent.length, op.extent.truncate_size, op.extent.truncate_seq);
+      result = do_extent_cmp(ctx, osd_op);
+      break;
 
     case CEPH_OSD_OP_SYNC_READ:
       if (pool.info.require_rollback()) {
--- a/src/osd/ReplicatedPG.cc
+++ b/src/osd/ReplicatedPG.cc
@@ -3655,6 +3655,37 @@ int ReplicatedPG::do_xattr_cmp_str(int op, string& v1s, bufferlist& xattr)
   }
 }
 
+int ReplicatedPG::do_writesame(OpContext *ctx, OSDOp& osd_op)
+{
+  ceph_osd_op& op = osd_op.op;
+  vector<OSDOp> write_ops(1);
+  OSDOp& write_op = write_ops[0];
+  uint64_t write_length = op.writesame.length;
+  int result = 0;
+
+  if (write_length % op.writesame.data_length)
+    return -EINVAL;
+
+  if (op.writesame.data_length != osd_op.indata.length()) {
+    derr << "invalid length ws data length " << op.writesame.data_length << " actual len " << osd_op.indata.length() << dendl;
+    return -EINVAL;
+  }
+
+  while (write_length) {
+    write_op.indata.append(osd_op.indata.c_str(), op.writesame.data_length);
+    write_length -= op.writesame.data_length;
+  }
+
+  write_op.op.op = CEPH_OSD_OP_WRITE;
+  write_op.op.extent.offset = op.writesame.offset;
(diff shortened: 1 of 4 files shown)


yuyuyu101:
Hi @mikechristie 

in my mind, VAAI is a block layer api spec which should integrate into librbd, why not implement this in RADOS instead of librbd?

yuyuyu101:
@gregsfortytwo yep, I also means we could impl detail logics in librbd layer, and leave generic CAS things to rados.

Oh, I guess you will impl it in kernel rbd firstly.

mikechristie:
@yuyuyu101, Yes, it is implemented in the kernel only right now. Here are the patches that are being proposed upstream:

https://github.com/mikechristie/linux-kernel/commits/vaai

@gregsfortytwo, You mean implement them as rbd cls ops in src/cls/rbd/cls_rbd.cc right? I swear I saw a comment in the ceph code saying something about being able to compare extents could be useful, so I went wild and implemented both as a OSD ops. I cannot find this comment, so maybe I was seeing things :) I think you are probably right and the only user would be rbd for this ESX support.

gregsfortytwo:
Yeah, that's what I meant. (Or perhaps as a separate object class, if that's desirable in terms of auth caps or something.)

I don't care that much and @athanatos is your real gate, but in general it's nice to not expand public interfaces too much. ;)

mikechristie:
Ok, actually, it looks like it is less invasive on the kernel side to implement as cls ops. The existing libceph and rbd kernel code seem to have the infrastructure I need to implement them.

So if no one wants them as osd ops, it makes the kernel side nicer to do them as cls ops.

yuyuyu101:
@mikechristie why not consider impl librbd firstly? related to iscsi target problem?

mikechristie:
@yuyuyu101,

I am new'ish to ceph. What do you mean by implement in librbd? Is that in the userspace lib ceph/src/librbd? If my target was in userpsace like tgt or tcmu, I could do something like this

(tgt making rbd_*() lib calls)
https://github.com/fujita/tgt/blob/master/usr/bs_rbd.c

I think this approach, but tgt is being replaced with the upstream LIO target in most distros. LIO *was* all kernel based, and so that was one reason I was focusing on krbd support. I also felt performance would be better and it would be a easier implementation.

jdurgin:
I also recall someone thinking a generic op would be useful, but can't find where. A class op can still be reused easily though - it also makes testing easier (no need to plumb through new client calls in userspace either).

So new class methods (don't really care if they're cls_rbd or a new cls) sound good to me.

ddiss:
These changes both look good to me:
Reviewed-by: David Disseldorp <ddiss@suse.de>

Though I'm not against implementing these as class operation, I think a generic cmpext op will be useful outside of just RBD - atomic compare-and-swap operations are already possible using xattrs and OMAP values. Allowing the same functionality with object data is complementary to this IMO.

If extra tests are needed for these operations, I'll gladly put my hand up.

jdurgin:
class ops are just as reusable as new rados ops. I don't see much reason to add more rados ops (like these) that could be implemented as class ops directly. class ops are easier to write, and can be used without bloating inferfaces - no need to plumb new ops through the Objecter and librados on the client side, for example.

For testing this, we'll want to update src/test/librados/(maybe misc.cc?) so we can detect regressions without going through extra layers like an application on top of iscsi.

dillaman:
I too remember this discussion last year -- I thought it was @liewegas who suggested they be ops as opposed to Mike's original implementation using class methods.

src/osd/ReplicatedPG.cc
```diff
@@ -5043,6 +5074,13 @@ int ReplicatedPG::do_osd_ops(OpContext *ctx, vector<OSDOp>& ops)
       }
       break;
 
+    case CEPH_OSD_OP_WRITESAME:
+      ++ctx->num_write;
+      tracepoint(osd, do_osd_op_pre_writesame, soid.oid.name.c_str(), soid.snap.val, oi.size, op.writesame.offset, op.writesame.length, op.writesame.data_length);
+
+     result = do_writesame(ctx, osd_op);
+     break;
```
liewegas: nit: indentation looks wonky?

liewegas:
Rados ops seem fine to me, but yeah, a class would also work.  @athanatos ?

liewegas:
We need functional tests in test/librados/misc.cc that verify these work as expected.

athanatos:
Rados ops seem reasonable.  I'm ok either way on that part.

+1 on the functional tests in test/librados/misc.cc

I'd also like ceph_test_rados (src/test/osd/RadosModel.h and src/test/osd/TestRados.cc) extended with these operations.  CMPEXT can probably be tested as part of the normal write op by randomly adding a CMPEXT which should succeed to one of the writes and sometimes (rand() % 3 or something) adding a CMPEXT that should fail with a write that would be noticed later in read if it happened (0's or something).  WRITESAME should probably be its own op (Object.h will need to be changed a bit).

wonzhq:
@mikechristie why the CMPEXT op just does the comparison? I think it should do a write if the comparison result is equal.

ddiss:
@wonzhq: similar to the xattr ops, atomic test-and-set updates are handled on the client side by issuing compound cmpext+write requests.

yuyuyu101:
@mikechristie are you still working on this? If target to jewel and no much free time, I would like to help

liewegas:
@yuyuyu101 we'd definitely appreciate help, especially with the ceph_test_rados changes.

mikechristie:
@yuyuyu101, I might have messed up on my tests. I will push my current branch later today, and then we can see what still needs to be done and you can help with any of it.

yuyuyu101:
@mikechristie great, look forward to this

ddiss:
Hi,

I hope I'm not duplicating work that Mike has already done, but I've pushed a preliminary writesame librados API and some basic ceph_test_rados_api_misc tests to https://github.com/ddiss/ceph/tree/mc-wip-osd-vaai
It's all atop Mike's existing OSD changes.

I'll continue adding writesame tests, and work on a cmpext API today/tomorrow. Preliminary feedback would be much appreciated.

liewegas:
@ddiss looks good!

mikechristie:
Hey ddiss,

A little overlap. Thanks for working on it!.

I am having trouble pushing my branch. Just so I can get it out to you and yuyuyu101, I just made a new branch wip-osd-vaai2.

Here is a link:
https://github.com/mikechristie/ceph/commits/wip-osd-vaai2

It has the beginnings of cmpext librados support and tests. I did not do the C API cmpext stuff yet.

ddiss:
Taking a look now - thanks Mike!

mikechristie:
Hey ddiss,

You do not have to actually waste your time and review the ws/cmp rados patches. Just use them for a guide if you have not started the cmpext stuff. I do not think some of the patches build against the tree I uploaded against. I just pushed them incase you wanted to look at them while you were working on this.

I think your ws patches are more complete so we should go forward with yours.

ddiss:
I've pushed a new round of writesame changes to my branch, with improved test coverage (aio, etc.). I'm now expanding on Mike's cmpext API changes and tests.

jdurgin:
closing this pr since it's been superseded by #8568 and a future cmpext patchset

PR #8177, opened 2016-03-17
global/signal_handler: print thread name in signal handler

majianpeng:
Now most thread has own thread name. So print thread name in signal
handle.

### DIFF
--- a/src/global/signal_handler.cc
+++ b/src/global/signal_handler.cc
@@ -91,14 +91,18 @@ static void handle_fatal_signal(int signum)
   // case, SA_RESETHAND specifies that the default signal handler--
   // presumably dump core-- will handle it.
   char buf[1024];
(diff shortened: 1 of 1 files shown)


src/global/signal_handler.cc
```diff
@@ -91,14 +91,18 @@ static void handle_fatal_signal(int signum)
   // case, SA_RESETHAND specifies that the default signal handler--
   // presumably dump core-- will handle it.
   char buf[1024];
+  char pthread_name[17] = {0}; //limited by 16B include terminating null byte.
```
tchaikov: so it should be `pthread_name[16]`?

majianpeng:
@tchaikov  updated. Thanks!

tchaikov:
lgtm

PR #8178, opened 2016-03-17
os/bluestore/BlueFS: Before reap ioct, it should wait io complete.

majianpeng:
When change write mode of FileWrite(in _flush_range) from buffer to
direct. osd met segment fault.
This because call _close_writer don't check related ios whether
complete.


majianpeng:
@liewegas . When test, i modify BlueFS::_flush_range code which change buffer mode to direct mode.
Using fio, every test met segment fault.  although i don't change whether correct. But it think it should not depend on this.

PR #8180, opened 2016-03-17
Jewel 15167

ukernel:


jcsp:
Looks sensible.  I guess we could alternatively choose to block splitting when scrub was in progress, but this is simpler.

gregsfortytwo:
Hmm, I think that breaks the scrub guarantee we want to have though: "anything in a tree when you start scrubbing, that doesn't get moved, will have been validated once scrubbing finishes". Users don't really control splitting so we need to handle it for them. :/ This can be an intermediate solution I guess since we aren't enabling directory fragmentation either...

Although actually I thought we *did* block split while scrubbing was queued (implicitly with some lock grabbing). Maybe I imagined that.

gregsfortytwo:

PR #8181, opened 2016-03-17
os/BlueStore: reduce unnecessary memory cost in 'aio_write'

tanghaodong25:
preallocate memory for zero buffer, reducing memory cost repeatly constructing bufferlist object.

Signed-off-by: haodong tang <haodong.tang@intel.com>

### DIFF
--- a/src/os/bluestore/KernelDevice.cc
+++ b/src/os/bluestore/KernelDevice.cc
@@ -44,8 +44,6 @@ KernelDevice::KernelDevice(aio_callback_t cb, void *cbpriv)
     aio_thread(this),
     injecting_crash(0)
 {
-  zeros = buffer::create_page_aligned(1048576);
-  zeros.zero();
 }
 
 int KernelDevice::_lock()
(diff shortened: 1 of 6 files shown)


yuyuyu101:
zero buffer uses "zeros" bufferptr which needn't alloc memory.

src/os/bluestore/KernelDevice.cc
```diff
@@ -456,12 +456,16 @@ int KernelDevice::aio_zero(
   assert(off + len <= size);
 
   bufferlist bl;
+  bufferlist t(len);
```
liewegas: * no len arg to the bufferlist ctor
* no need for t at all.. let's just append zeroes directly to bl inside the loop

liewegas:
There is actually a bufferlist::append_zero(len) method.. let's use that instead, and optimize it?

liewegas:
failed ceph_test_objectstore, 
/a/sage-2016-05-23_05:47:51-rados-wip-sage-testing---basic-smithi/208868

liewegas:
I think we can drop aio_zero completely, actually... there aren't any users, and the new lextent+blob format makes it unnecessary to zero regions of disk.

PR #8182, opened 2016-03-17
rbd-mirror: make remote context respect env and argv config params

trociny:


jdurgin:
nice catch, lgtm

dillaman:
lgtm

PR #8183, opened 2016-03-17
Wip rgw fix mdlog

oritwas:


src/rgw/rgw_rest_log.cc
```diff
@@ -78,10 +78,14 @@ void RGWOp_MDLog_List::execute() {
   } 
 
   if (period.empty()) {
-    ldout(s->cct, 5) << "Missing period id" << dendl;
-    http_ret = -EINVAL;
-    return;
+    ldout(s->cct, 5) << "Missing period id trying to use current" << dendl;
+    RGWPeriod current_period;
+    http_ret = current_period.init(s->cct, store);
```
yehudasa: @oritwas can't we get current period from store->realm.current_period?

oritwas: yes I will fix it

oritwas:
@yehudasa  pushed a new versiom

src/rgw/rgw_rest_log.cc
```diff
@@ -78,10 +78,15 @@ void RGWOp_MDLog_List::execute() {
...
-    ldout(s->cct, 5) << "Missing period id" << dendl;
-    http_ret = -EINVAL;
-    return;
+    ldout(s->cct, 5) << "Missing period id trying to use current" << dendl;
+    period = store->get_current_period_id();
+    if (period.empty()) {
+      ldout(s->cct, 5) << "Missing period id" << dendl;
+      http_ret = -EINVAL;
+      return;
+    }
```
cbodley: This is a fix for the sync agent, which will always send an empty period id, right? And the sync agent is not compatible with a new multisite configuration?

If the gateway isn't running a new multisite config (i.e. realm.current_period is empty), it should just accept and use the empty period id.

But if the gateway does have a valid current_period, I think it should still reject requests that don't give a period id. The remote side needs to be explicit about which period it's syncing, so falling back to the current_period is probably not the right thing to do.

So I would suggest this change instead:
```c++
// reject empty period ids if we have a current_period
if (period.empty() && !store->get_current_period_id().empty()) {
  ldout(s->cct, 5) << "Missing period id" << dendl;
  http_ret = -EINVAL;
  return;
}
```

yehudasa: @cbodley this is for the sync agent, which doesn't send a period id at all. The idea is to allow a path for migration from old config to new one without breaking the sync agent. That means that if users created a new configuration with a realm and a period, then we should still allow it to continue working as it was working before.

cbodley: once we reconfigure with a realm and period, our new multisite sync stuff kicks in. can that and the sync agent run at the same time without causing problems?

yehudasa: @cbodley new sync can be disabled, and I don't really see any issue with both happening concurrently (they use separate objects to track status).

PR #8184, opened 2016-03-17
rgw: add zone delete to rgw-admin help

theanalyst:
This operation is already implemented (OPT_ZONE_DELETE), however not
mentioned in the help, just adding to the help options


src/rgw/rgw_admin.cc
```diff
@@ -108,6 +108,7 @@ void _usage()
   cout << "  zonegroup-map get          show zonegroup-map\n";
   cout << "  zonegroup-map set          set zonegroup-map (requires infile)\n";
   cout << "  zone create                create a new zone\n";
+  cout << "  zone delete                delete a zone\n";
```
yehudasa: @theanalyst also need to update `test/cli/radosgw-admin/help.t`

theanalyst: @yehudasa updated

PR #8185, opened 2016-03-17
rbd-mirror: fix long termination due to 30sec wait in Mirror::run loop

trociny:

### DIFF
--- a/src/tools/rbd_mirror/Mirror.cc
+++ b/src/tools/rbd_mirror/Mirror.cc
@@ -40,6 +40,10 @@ Mirror::Mirror(CephContext *cct, const std::vector<const char*> &args) :
 void Mirror::handle_signal(int signum)
 {
   m_stopping.set(1);
+  {
+    Mutex::Locker l(m_lock);
+    m_cond.Signal();
+  }
 }
 
 int Mirror::init()


src/tools/rbd_mirror/main.cc
```diff
@@ -58,14 +58,18 @@ int main(int argc, const char **argv)
...
   int r = mirror->init();
   if (r < 0) {
     std::cerr << "failed to initialize: " << cpp_strerror(r) << std::endl;
     goto cleanup;
   }
 
-  mirror->run();
+  while (!stopping.read()) {
+    struct timeval sleep_time{1, 0};
+    ::select(0, NULL, NULL, NULL, &sleep_time);
```
jdurgin: rather use c++11's sleep_for or a plain sleep() (no need for hi-res here)

src/tools/rbd_mirror/Mirror.cc
```diff
@@ -30,15 +30,25 @@ namespace mirror {
...
 
-void Mirror::handle_signal(int signum)
+Mirror::~Mirror()
 {
+  dout(20) << "terminating" << dendl;
+
   m_stopping.set(1);
+  {
+    Mutex::Locker l(m_lock);
+    m_cond.Signal();
```
dillaman: Wouldn't just signaling the condition be enough of a change to interrupt the 30 second sleep?

src/tools/rbd_mirror/main.cc
trociny: I used select() here because I knew it would interrupt if signal were received (so no delay before shutdown). I will check if sleep()/sleep_for.

src/tools/rbd_mirror/Mirror.cc
trociny: I think it is not safe to do in a signal handler. That is why I had to add an additional thread.

trociny:
Ah, actually, I should have looked how SignalHandler framework work first! I thought register_async_signal_handler() is just a wrapper for sigaction(), but actually we have a thread for handling signals.

So it looks it is safe to just signal the condition in Mirror::handle_signal().  I have rebased the PR.

dillaman:
lgtm

PR #8186, opened 2016-03-17
rgw: optimize BulkDelete authorization and response printing

rzarzynski:
According to how object removal is authorized by verify_permission()
method of RGWDeleteObj class, we don't need to examine ACL of each
single object specified to delete by the BulkDelete of Swift API.

This commit removes the unnecessary check.


rzarzynski:
Both *Tempest* and *s3-tests* haven't found a regression here. Removing the *DNM* prefix.

PR #8188, opened 2016-03-17
systemd: Use the same restart limits as upstart

branto1:
Currently, the systemd daemons are not restarted on failure. This patch
adds this functionality and sets the defaults to those defined in
upstart. This resolves to 3 fails per 30 minutes for osd, mon and mds
and 5 fails per 30 seconds for radosgw.

### DIFF
--- a/systemd/ceph-mds@.service
+++ b/systemd/ceph-mds@.service
@@ -16,6 +16,9 @@ ProtectHome=true
 ProtectSystem=full
 PrivateTmp=true
 TasksMax=infinity
+Restart=on-failure
+StartLimitInterval=30min
+StartLimitBurst=3
 
 [Install]
 WantedBy=ceph-mds.target
--- a/systemd/ceph-mon@.service
+++ b/systemd/ceph-mon@.service
@@ -22,6 +22,9 @@ ProtectHome=true
 ProtectSystem=full
 PrivateTmp=true
 TasksMax=infinity
+Restart=on-failure
+StartLimitInterval=30min
+StartLimitBurst=3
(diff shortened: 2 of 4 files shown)


liewegas:
@tmuthamizhan i seem to remember restarts being an issue a few months back, and not worrying about systemd because it didn't do any restarts.  do you remember what the issue was?

gregsfortytwo:
@liewegas are you thinking of something different than http://tracker.ceph.com/issues/11798, which is why we set the limits explicitly to begin with? :)

liewegas:
That's probably it, but I vaguely remember telling someone "don't worry 
about that--systemd doesn't do restarts at all".  Maybe/probably it was 
just about not needed to fix that issue.

branto1:
@liewegas: I vaguely remember something about teuthology starting the ceph daemons directly and systemd not being able to restart the services that were not started thorugh it. It should not be related to this PR, though.

gregsfortytwo:
https://bugzilla.redhat.com/show_bug.cgi?id=1262977 is titled "[RFE] verify that systemd config prevents repeatedly restarting daemons", but seems to have turned into "make it match upstart". :) Maybe it's fine as-is? But I think we probably want to have *some* restart capability.

branto1:
Rebased on top of latest master because of the conflicts.

liewegas:
@ktdreyer @smithfarm look ok to you?

ktdreyer:
Thanks!
`Reviewed-by: Ken Dreyer <kdreyer@redhat.com>`

PR #8189, opened 2016-03-17
librados examples: link and include from current source tree by default.

chardan:
Fixes #15100. The Makefile now by default links and includes from the
current source tree's librados and offers another option for building
examples against the system's.

Signed-off-by: Jesse Williamson <nerd.cpp@gmail.com>

### DIFF
--- a/examples/librados/Makefile
+++ b/examples/librados/Makefile
@@ -1,15 +1,35 @@
-CXXFLAGS=-std=c++0x -I../../src/include
-CFLAGS=-I../../src/include
-all: librados_hello_world librados_hello_world_c
 
-librados_hello_world: hello_world.cc
-	g++ -g -c hello_world.cc -o hello_world.o $(CXXFLAGS)
-	g++ -g hello_world.o -lrados -o librados_hello_world $(LDFLAGS)
+CXX?=g++
+CXX_FLAGS?=-std=c++11 -Wall -Wextra -Werror -g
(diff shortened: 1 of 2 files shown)


chardan:
To verify:
* build local librados
* type "make"
* see local include path in compiler commands:
-I../../src/include
* see programs linked to objects in build tree:
> ldd hello_world_cpp hello_world_c | grep librados
       librados.so.2 => ../../src/.libs/librados.so.2 (0x00007f0cc0bb7000)
       librados.so.2 => ../../src/.libs/librados.so.2 (0x00007fcb81ec6000)

* make clean
* make all-system
* see no special include path in compiler commands (currently just means no -I)
* see programs linked to system libraries:
> ldd hello_world_cpp hello_world_c | grep librados
       librados.so.2 => /usr/lib64/librados.so.2 (0x00007fcef2d10000)
       librados.so.2 => /usr/lib64/librados.so.2 (0x00007f4ce08d6000)

chardan:
@jdurgin Please have a look. Thanks!

jdurgin:
Looking good in general. We'll want to add a script (e.g. qa/workunits/rados/test_build.sh) to build this against system packages. It can assume librados and its headers are installed, and build in cwd, grabbing these example files from git.ceph.com.

Then we can add a simple job to run that script to ceph-qa-suite/suites/rados/singleton-nomsgr/all

PR #8190, opened 2016-03-18
Wip rgw sync fixes 4

yehudasa:


src/rgw/rgw_rados.cc
```diff
@@ -6411,14 +6412,20 @@ int RGWRados::fetch_remote_obj(RGWObjectCtx& obj_ctx,
 
   string obj_name = dest_obj.bucket.name + "/" + dest_obj.get_object();
 
-  RGWOpStateSingleOp opstate(this, client_id, op_id, obj_name);
+  RGWOpStateSingleOp *opstate = NULL;
```
cbodley: it would be nice if we could avoid the heap allocation here.. it looks like you could leave opstate on the stack and replace the `if (opstate) {` blocks with `if (record_op_state) {`

otherwise, consider using std::unique_ptr to make sure we don't miss a delete in an error path

src/rgw/rgw_data_sync.cc
```diff
@@ -551,15 +551,16 @@ static string full_data_sync_index_shard_oid(const string& source_zone, int shar
...
 
-  bucket_instance_meta_info() : mtime(0) {}
+  bucket_instance_meta_info() {}
 
   void decode_json(JSONObj *obj) {
     JSONDecoder::decode_json("key", key, obj);
     JSONDecoder::decode_json("ver", ver, obj);
-    JSONDecoder::decode_json("mtime", mtime, obj);
+    string s;
+    JSONDecoder::decode_json("mtime", s, obj);
```
cbodley: why decode into a temporary string just to discard it?

is `RGWMetadataManager::get()` where this gets encoded? it's hard to tell, but it encodes the same keys

yehudasa: @cbodley yeah, this is missing the creation of mtime from s. Right, it's encoded in RGWMetadataManager::get().

PR #8191, opened 2016-03-18
qa/workunits/rest/test.py: don't use newfs

liewegas:
It doesn't work anymore.


liewegas:
@jcsp is this right?

jcsp:
Yep, looks correct.

PR #8193, opened 2016-03-18
os/bluestore: a few fixes

liewegas:


liewegas:
https://github.com/ceph/ceph-qa-suite/pull/894

src/os/bluestore/BlueStore.cc
```diff
@@ -5466,10 +5514,13 @@ int BlueStore::_do_write(
     if (offset >= bp->first &&
 	offset > tail_start &&
 	offset + length >= o->onode.size &&
+	o->tail_offset == tail_start &&
 	o->tail_bl.length() &&
 	(offset / block_size == (o->onode.size - 1) / block_size)) {
       dout(20) << __func__ << " using cached tail" << dendl;
       assert((offset & block_mask) == (o->onode.size & block_mask));
+      // wait for any related wal writes to commit
+      txc->osr->wait_for_wal_on_seq(o->tail_txc_seq);
```
yuyuyu101: There are lots of changes in bluestore, so I'm not sure whether catch the real problem.

I guess the problem is a previous wal op have overlap with later tail append op. If crash happen,we can't reply correct result via WAL. Because the later op may write data successfully but don't have wal data part. If we reply all entries, it will let previous op rewrite old data?

liewegas: This problem only comes up because we are making use of the cached state, 
which doesn't happen during wal replay. The problem is we may have a wal 
op that wrote the data we're caching, but it hasn't committed yet.. and 
then we use the cached data to queue a new write that doesn't need wal and 
orders before it.

PR #8194, opened 2016-03-18
os/bluestore/KernelDevice.cc: small fix in buffer flush

tanghaodong25:
should set 'io_since_flush' to 0 after sync operation finish, otherwise data in buffer will not be synced to block device if sync operation get error.

Signed-off-by: haodong tang <haodong.tang@intel.com>

### DIFF
--- a/src/os/bluestore/KernelDevice.cc
+++ b/src/os/bluestore/KernelDevice.cc
@@ -202,6 +202,7 @@ int KernelDevice::flush()
   if (r < 0) {
     r = -errno;
     derr << __func__ << " fdatasync got: " << cpp_strerror(r) << dendl;
+    assert(0);
   }
   dout(5) << __func__ << " in " << dur << dendl;;
   return r;
--- a/src/os/bluestore/KernelDevice.cc
+++ b/src/os/bluestore/KernelDevice.cc
@@ -176,15 +176,12 @@ void KernelDevice::close()
 
 int KernelDevice::flush()
 {
-  // serialize flushers, so that we can avoid weird io_since_flush
-  // races (w/ multipler flushers).
-  Mutex::Locker l(flush_lock);
-  if (io_since_flush.read() == 0) {
+  bool ret = io_since_flush.compare_and_swap(1, 0);
+  if (!ret) {
     dout(10) << __func__ << " no-op (no ios since last flush)" << dendl;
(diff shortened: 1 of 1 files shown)


yuyuyu101:
Actually I don't think it's wrong. If we set flag later and new io comming the gap between fdatasync and "set", we will miss the io needed to be flush

tanghaodong25:
Thanks for your reply! You are right, indeed this situation exits and is it necessary to set 'io_since_flush' to 1 if 'fdatasync' operation get error?

majianpeng:
@yuyuyu101 . Ask other question: The reason why use mutext "// serialize flushers, so that we can avoid weird io_since_flush / races (w/ multipler flushers).". But i think we can remove this only replace 
io_since_flush.read() w/ io_since_flush.exchange(0).   What do you think of it?

yuyuyu101:
for most of flush ops, atomic_t is enough and do thing as expected. For a little racing cases, it may trigger two flush calls which still can be accept. Since flush is a fast path, atomic_t looks better still now.

yuyuyu101:
@tanghaodong25 bluestore.cc now ignore the return value of flush failed, I think you could add error handle in bluestore.cc or just insert assert(0) in KernelDevice::flush

liewegas:
Let's assert(0) for now

tanghaodong25:
a) add 'insert(0)'.
b) use atomic compare and swap to replace mutex lock.

src/os/bluestore/KernelDevice.cc
```diff
@@ -176,15 +176,14 @@ void KernelDevice::close()
 
 int KernelDevice::flush()
 {
-  // serialize flushers, so that we can avoid weird io_since_flush
-  // races (w/ multipler flushers).
-  Mutex::Locker l(flush_lock);
   if (io_since_flush.read() == 0) {
```
yuyuyu101: actually, you need to do atomic compare here.

tanghaodong25:
@yuyuyu101  add more fixes.

```c++
if io_since_flush == 0:
    // ret will be 0, and do nothing
    return;
else
    // io_since_flush will swap to 0
    // ret will be 1
    fdatasync();
```

Thanks.

yuyuyu101:
LGTM

PR #8195, opened 2016-03-18
cmake changes

tchaikov:
no regression found, and there's a substantial reduction in the size of
generated librados library (both stripped and not).


liewegas:
rebase?

tchaikov:
@liewegas rebased.

alimaredia:
@tchaikov looks fine to me, one question I have: is CMAKE_INSTALL_LIBDIR replacing "lib" as the DESTINATION everywhere in the code or just in the places in your commit?

tchaikov:
@alimaredia i tried to fix this inconsistency whenever i see it. ideally we should use `${CMAKE_INSTALL_LIBDIR}` everywhere.

PR #8197, opened 2016-03-18
rgw:response information is error when geting token of swift account

qiankunzheng:
The header is X-Trans-Id instead of x-amz-request-id in the response header

Fixes:#15195
Signed-off-by: Qiankun Zheng <zheng.qiankun@h3c.com>


qiankunzheng:
@oritwas 
Hi,could you review it? thanks

PR #8198, opened 2016-03-18
osd: drop unnecessary transaction cleanup

xiexingguo:
The lock() method will ensure 'dirty_info' will be false,
and within this function's scope I see no possibility to
set 'dirty_info' true. Thus I guess the transaction cleanup
logic before exit is never reachable and therefore shall be
considered as redundant and can be safely removed.

### DIFF
--- a/src/osd/PG.cc
+++ b/src/osd/PG.cc
@@ -2008,12 +2008,7 @@ void PG::_activate_committed(epoch_t epoch, epoch_t activation_epoch)
     }
   }
 
-  if (dirty_info) {
-    ObjectStore::Transaction t;
-    write_if_dirty(t);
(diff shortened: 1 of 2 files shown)


xiexingguo:
@liewegas Have a look?

xiexingguo:
@tchaikov You might want to review this one too, since you are there :-)

src/osd/PG.cc
```diff
@@ -2008,13 +2008,6 @@ void PG::_activate_committed(epoch_t epoch, epoch_t activation_epoch)
     }
   }
 
-  if (dirty_info) {
```
asheplyakov: What about
`assert(!dirty_info);`

xiexingguo: good advice, thank you, sir!

PR #8199, opened 2016-03-18
rgw: Use def bucket quota when bucket quota disable

Aran85:
if not bucket quota cant lower than def quota


Aran85:
@dang could you have a look at the quick change? thanks a lot!

yehudasa:
The code has changed here, and I think @dang has addressed the issue being fixed here. Closing it, let me know if you think we still need to fix this somehow.

PR #8201, opened 2016-03-18
mon: when set the pool size should consider the crush rule size

songbaisen:
mon: when set the pool size should consider the crush rule size

### DIFF
--- a/src/mon/OSDMonitor.cc
+++ b/src/mon/OSDMonitor.cc
@@ -4863,6 +4863,19 @@ int OSDMonitor::prepare_command_pool_set(map<string,cmd_vartype> &cmdmap,
       ss << "pool size must be between 1 and 10";
       return -EINVAL;
     }
+    const int64_t crush_ruleset = p.get_crush_ruleset();
+    const int64_t minsize = osdmap.crush->get_rule_mask_min_size(crush_ruleset);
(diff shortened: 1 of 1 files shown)


liewegas:
This should wait until after jewel.  I thinkwe need to

* finally get rid of ruleset vs rule id
* ensure that there is a path to change to a different rule *and* change the pool size where the first and second rule have non-overlapping min_size..max_size

songbaisen:
@liewegas  Thank you for you reply. I can not understand why the two rules can not have the overlapping  min_size..max_size.

In my option if pool A sise is 4  use  rule 0, rul0 min_size is 1  max_size.5 
change to rule 1,rule 1 min_size is 3 max_size.5 .The two rules have the overlapping  should
be ok.

@tchaikov  HI!Is that sage mean what i said.

tchaikov:
i am not sure if i understand your query. but ruleset and ruleid are different things here. you are using one as another.

songbaisen:
@tchaikov  HI!Thank you for your reply.I have a  doubt.Here if do not use ruleset,how does the pool can know which rule it is used.

liewegas:
I think we're better served by eliminating ruleset entirely.

songbaisen:
@liewegas  Yes.Sir.

PR #8202, opened 2016-03-18
osdc/Objecter: dout log after assign tid

XinzeChi:
This would help to debug

### DIFF
--- a/src/osdc/Objecter.cc
+++ b/src/osdc/Objecter.cc
@@ -2276,11 +2276,6 @@ ceph_tid_t Objecter::_op_submit(Op *op, shunique_lock& sul)
   _send_op_account(op);
 
   // send?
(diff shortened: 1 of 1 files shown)


src/osdc/Objecter.cc
```diff
@@ -2321,6 +2314,14 @@ ceph_tid_t Objecter::_op_submit(Op *op, shunique_lock& sul)
   OSDSession::unique_lock sl(s->lock);
   if (op->tid == 0)
     op->tid = last_tid.inc();
+
+  // send?
```
tchaikov: might need to leave this comment where it was.

tchaikov:
lgtm but need to keep the comment as it is.

tchaikov:
lgtm

PR #8203, opened 2016-03-18
rbd-mirror: don't link dynamically with librados

trociny:
It fixes the issue with multiple instances of lockdep static
variables, which led to lockdep assert violations due to locks
registered in one variables and unregistered in another.

### DIFF
--- a/src/test/Makefile-client.am
+++ b/src/test/Makefile-client.am
@@ -507,8 +507,8 @@ ceph_test_rbd_mirror_image_replay_LDADD = \
 	librbd_api.la \
 	$(LIBRBD_TYPES) \
 	libjournal.la \
-	$(LIBRADOS) $(LIBOSDC) \
-	librados_internal.la \
+	librados_api.la \
+	$(LIBRADOS_DEPS) \
 	libcls_rbd_client.la \
 	libcls_lock_client.la \
 	libcls_journal_client.la \
--- a/src/tools/Makefile-client.am
+++ b/src/tools/Makefile-client.am
@@ -131,8 +131,8 @@ rbd_mirror_LDADD = \
 	librbd_api.la \
 	$(LIBRBD_TYPES) \
 	libjournal.la \
-	$(LIBRADOS) $(LIBOSDC) \
-	librados_internal.la \
+	librados_api.la \
+	$(LIBRADOS_DEPS) \
 	libcls_rbd_client.la \
 	libcls_lock_client.la \
 	libcls_journal_client.la \


dillaman:
It would be great if we could fix the lock naming to ensure they are unique instead of statically linking to librados.

trociny:
@dillaman This is not the issue with lock naming. lockdep_register/unregister supports locks with the same name (using refs).

I think the problem is that rbd-mirror has two instances of lockdep.o, one is from librados.so and one is statically linked in rbd-mirror. As a result, we have two instances of lockdep static variables that used for registering locks (lockdep_mutex, lock_ids, lock_names...). As a result it is possible that a lock with name 'XXX' is registered in one lockdep instance and gets id=Y, and unregistered in another instance, actually unregistering other lock that happened to have id=Y here.

In rbd-mirror I observe, that HeartbeatMap::m_rwlock is registered with id=9 in one lockdep, while in another lockdep PerfCountersCollection is registered with id=9. Then when the remote context is destroyed the second lockdep is used for unregistering both locks, and it crashed on the second unregister.

I guess it is unsafe to link both statically librados_internal and dynamically librados?

dillaman:
@trociny I think the way to solve this is to move lockdep's state from being global static variables and instead register them as a CephContext ```lookup_or_create_singleton_object``` singleton.  That would prevent the odd mix-n-match between the static and dynamic linked versions.

trociny:
@dillaman Using lookup_or_create_singleton_object for lockdep looks like a good idea. Though not sure right now about details and how much changes it will require.

But anyway, isn't it  a bad idea to have both static and dynamic linked versions of the code? Won't it create other potential issues, either similar to lockdep or different?

E.g. theoretically, if a library is linked dynamically, it is not required to use the same version the binary was built with. But now consider if internal functions diverged and we have two instances of these functions?

dillaman:
@trociny It's not ideal, but since we version lock our rpm/debs, we are generally safe.  Basically, everything in "common" is already linked in twice to librados / librbd.  librbd even directly manipulates the CephContext allocated by librados -- so when that data structure changes (especially md_config) and the two libraries are out-of-sync, you can expect a crash.

trociny:
@dillaman I am working on using lookup_or_create_singleton_object() for lockdep singleton. I will create a new PR when it is ready.

trociny:
@dillaman Actually, I got stuck with the solution.

The initial idea was that I could use global (not static) variables. I had a misunderstanding about how linking for global variables works and thought that global variables do not suffer from this problem (only one instance of global variable still exists after mixing dynamic and static linking). So the plan was to change type for global g_lockdep from bool (int) to a structure which would keep all lockdep state variables. See the initial patch (it was planned to be improved if it worked):

https://github.com/trociny/ceph/commit/a282389048eff641ba626fa7e3d03c3ad1762a14

But it turned out that I was wrong and there is the same issue with global vars as it is with static.

So right now I can't think out a good solution. To fix this we would need to get rid of using global variables at all. It could be possible if ceph context was passed to RWLock and Mutex, so it could be used by lockdep (then I could use lookup_or_create_singleton_object() for creating and accessing lockdep singleton). Right now Mutex constructor has ceph context as an optional paramer and RWLock does not have ceph context at all, and I am not sure it is easy to fix it (or rather all its users).

Also it might not be a very good idea to limit lockdep to one ceph context -- theoretically, dependencies between locks from different contexts are still possible, and we will not be able to track them. (it might not be an issue on practice though)

jdurgin:
For now let's just use helgrind instead - this will still work if we switch to std::mutex in the future too.

dillaman:
Closing to #8476

PR #8208, opened 2016-03-18
mon: mark_down_pgs in lockstep with pg_map's osdmap epoch

liewegas:
This fixes the stale PGs we've been seeing.  And probably
several corner cases that are hard to see.

### DIFF
--- a/src/mon/PGMonitor.cc
+++ b/src/mon/PGMonitor.cc
@@ -1159,27 +1159,35 @@ bool PGMonitor::map_pg_creates()
         up_primary != s->up_primary ||
         acting !=  s->acting ||
         acting_primary != s->acting_primary) {
-      dout(20) << __func__ << "  " << pgid << " "
-               << " acting_primary: " << s->acting_primary
-               << " -> " << acting_primary
-               << " acting: " << s->acting << " -> " << acting
-               << " up_primary: " << s->up_primary << " -> " << up_primary
-               << " up: " << s->up << " -> " << up
-               << dendl;
-
       pg_stat_t *ns = &pending_inc.pg_stat_updates[pgid];
-      *ns = *s;
+      if (osdmap->get_epoch() > ns->reported_epoch) {
+	dout(20) << __func__ << "  " << pgid << " "
+		 << " acting_primary: " << s->acting_primary
+		 << " -> " << acting_primary
+		 << " acting: " << s->acting << " -> " << acting
+		 << " up_primary: " << s->up_primary << " -> " << up_primary
+		 << " up: " << s->up << " -> " << up
+		 << dendl;
 
-      // note epoch if the target of the create message changed
-      if (acting_primary != ns->acting_primary)
-	ns->mapping_epoch = osdmap->get_epoch();
+	// only initialize if it wasn't already a pending update
+	if (ns->reported_epoch == 0)
+	  *ns = *s;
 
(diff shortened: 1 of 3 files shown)


liewegas:
@tchaikov @jecluis review please?

liewegas:
passed rados suite

liewegas:
added a fix for a race between map_pg_create and a pg_stat update, observed at http://pulpito.ceph.com/sage-2016-03-20_11:34:21-rados-wip-sage-testing-distro-basic-smithi/76029

src/mon/PGMonitor.cc
```diff
@@ -1310,9 +1310,19 @@ void PGMonitor::_try_mark_pg_stale(
...
+	   << pg_map.last_osdmap_epoch << dendl;
+  if (pg_map.last_osdmap_epoch == 0)
+    return false;
+
+  // use the OSDMap that matches the one pg_map has consumed.
+  std::unique_ptr<OSDMap> osdmap;
+  bufferlist bl;
+  int err = mon->osdmon()->get_version_full(pg_map.last_osdmap_epoch, bl);
+  assert(err == 0);
+  osdmap.reset(new OSDMap);
```
tchaikov: could use `std::make_unique()`.

src/mon/PGMonitor.cc
```diff
@@ -1310,9 +1310,19 @@ void PGMonitor::_try_mark_pg_stale(
...
-  dout(10) << "check_down_pgs" << dendl;
+  dout(10) << "check_down_pgs last_osdmap_epoch "
+	   << pg_map.last_osdmap_epoch << dendl;
+  if (pg_map.last_osdmap_epoch == 0)
+    return false;
+
+  // use the OSDMap that matches the one pg_map has consumed.
+  std::unique_ptr<OSDMap> osdmap;
+  bufferlist bl;
+  int err = mon->osdmon()->get_version_full(pg_map.last_osdmap_epoch, bl);
```
tchaikov: the OSDMonitor kicks pgmon in `OSDMonitor::update_from_paxos()`, so the pgmon can consume (`check_osd_map()`) the latest osdmaps once the monitor is **active**[1]. and pg_map.last_osdmap_epoch is updated in `PGMonitor::update_from_paxo()` when the monitor is still **updating**.

so, is there a chance that `check_down_pgs()` gets called **after** `pg_map.last_osdmap_epoch` is updated, and **before** `check_osd_map()` is called? in other words:

1. pg_map.set_last_osdmap_epoch(last_osdmap_epoch);
1. PGMonitor::check_down_pgs()
1. check_osd_map()

---
[1] see `PGMonitor::check_osd_map()`
```c++
  if (!is_writeable()) {
    dout(10) << __func__ << " -- pgmap not writeable, waiting" << dendl;
    wait_for_writeable_ctx(new RetryCheckOSDMap(this, epoch));
    return;
  }
```

the monitor is not writable until it is done with refresh (the `update_from_paxos()` calls).

tchaikov: please ignore the comment above, just read b67ac8c318f8bf635b221b626040b802d2f119c2 , which renders the case above possible

tchaikov:
lgtm modulo the suggestion to use `std::make_unique()`

liewegas:
will do one more rados run with the new commit

PR #8209, opened 2016-03-18
ceph_test_rados_misc: shorten mount timeout

liewegas:
This might fix #13992.


tchaikov:
lgtm this is my best guess. i also tried to understand why the 2nd test failed, nothing came to my mind. maybe we just need to have a smaller timeout.

PR #8210, opened 2016-03-18
osd: add 'proxy' cache mode

liewegas:
The 'forward' mode is flawed for the same reason that redirects
are flawed: they allow the reordering of requests.  Add a proxy
mode that can be used instead of forward.  (Later we can consider
dropping forward entirely?)

Fixes: #12814


liewegas:
https://github.com/ceph/ceph-qa-suite/pull/906

athanatos:
lgtm

PR #8211, opened 2016-03-18
doc: draft notes for jewel

liewegas:


dillaman:
@liewegas Github isn't allowing inline comments:

Major changes since Infernalis:

* The exclusive-lock, object-map, fast-diff, and journaling features can be enabled or disabled dynamically.  The deep-flatten can be disabled dynamically but not re-enabled.

* The RBD CLI has been rewritten to provide command-specific help and full bash completion support.

* RBD snapshots can now be renamed.

Minor changes since Infernalis (worth mentioning):

* The default RBD image features for new images have been updated to enable the following: exclusive lock, object map, fast-diff, and deep-flatten. These features are not currently supported by the RBD kernel driver nor older RBD clients. These features can be disabled on a per-image basis via the RBD CLI or the default features can be updated to the pre-Jewel setting by adding the following to the client section of the Ceph configuration file:
  rbd default features = 1

* The rbd legacy image format (version 1) is deprecated with the Jewel release.

doc/release-notes.rst
```diff
@@ -823,6 +823,14 @@ We recommend that all infernalis v9.2.0 users upgrade.
...
 For more detailed information, see :download:`the complete changelog <changelog/v9.2.1.txt>`.
 
+Upgrading
+---------
+
+* Some symbols wrongly exposed by librados in v9.1.0 and v9.2.0 were removed.
+  If you compiled your own application against librados shipped with these
+  releases, it is very likely referencing these removed symbols. So you will
+  need to recompile it.
+
```
jdurgin: only c++ applications, right?

dillaman: Correct -- it was in buffer.h

jdurgin:
"multiple MDS servers" - is active/standy not considered stable?

jdurgin:
> There is new support for asynchronous replication

perhaps add "(mirroring)" to introduce the term there, and mention the new rbd-mirror daemon

theanalyst:
github doesst seem to allow inline comments on this changeset.. on RGW:
`The new multitenancy infrastructure improves compatibility with
Swift (which has a provide namespace of containers for each
tenant).` -> (which provides ...)

PR #8216, opened 2016-03-18
qa/workunits/rbd: new online maintenance op tests

dillaman:

### DIFF
--- a/qa/workunits/rbd/qemu_dynamic_features.sh
+++ b/qa/workunits/rbd/qemu_dynamic_features.sh
@@ -1,0 +1,44 @@
+#!/bin/sh -ex
+
+if [[ -z "${IMAGE_NAME}" ]]; then
+  echo image name must be provided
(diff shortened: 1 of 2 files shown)


qa/workunits/rbd/qemu_dynamic_features.sh
```diff
@@ -0,0 +1,39 @@
...
+  while ! is_qemu_running ; do
+    echo "*** Waiting for QEMU"
+    sleep 30
+  done
+}
+
+wait_for_qemu
+while is_qemu_running ; do
+  echo "*** Disabling all features"
+  rbd feature disable ${IMAGE_NAME} journaling || true
```
jdurgin: seems like we should check the return status when these are meant to succeed - maybe have an inital setup that gets the image to a known state with || true

dillaman:
@jdurgin pushed

PR #8218, opened 2016-03-18
os/kstore: we have flush_txns is to make sure read after write for th…

jupiturliu:
…e same object. That's the reason why we add _txc_finalize(osr,txc) after we add the

transaction. However, since we did not call STATE_FINISHING at all, The _txf_finish will never be called and flush_txns of object will never be removed.
It will make the read on the same object hanging there forever. We will combine KV_DONE and STATE_FINISHING.

Signed-off-by: James Liu <james.liu@ssi.samsung.com>


athanatos:
@liewegas Is this still relevant?

PR #8219, opened 2016-03-18
test_pool_create.sh: put test files in the test dir so they are cleaned up

jdurgin:
If they aren't cleaned up, tarball gitbuilders are unhappy.


gregsfortytwo:
LGTM, let's merge if the jenkins builder is happy with it.

PR #8223, opened 2016-03-19
hammer: Bucket owner isn't changed after unlink/link

Vicente-Cheng:
http://tracker.ceph.com/issues/15089


Vicente-Cheng:
need more info about this backport, close now.

oritwas:
user_id is not used so just remove the line.
instead of system_obj_set_attr you can use the method set_attr

Vicente-Cheng:
@oritwas 
thanks for your help! reopen on #8583

PR #8224, opened 2016-03-19
rbdmap: add manpage

smithfarm:
http://tracker.ceph.com/issues/15212 Fixes: #15212

### DIFF
--- a/doc/man/8/rbdmap.rst
+++ b/doc/man/8/rbdmap.rst
@@ -1,0 +1,43 @@
+:orphan:
+
+=========================================
+ rbdmap -- map RBD devices at boot time
+=========================================
+
+.. program:: rbdmap
+
+Synopsis
+========
+
+| **rbdmap**
+
+
+Description
+===========
+
+**rbdmap** is a shell script that is intended to be run at boot time
+by the init system (sysvinit, upstart, systemd). It looks for an
+environment variable ``RBDMAPFILE``, which defaults to
+``/etc/ceph/rbdmap``. This file is expected to contain a list of RBD
+devices to be mapped (via the ``rbd map`` command) and mounted at boot
+time.
+
+
+Options
+=======
+
+None
+
+
+Availability
+============
+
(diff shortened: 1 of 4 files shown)


doc/man/8/rbdmap.rst
```diff
@@ -0,0 +1,43 @@
...
+ rbdmap -- map RBD devices at boot time
+=========================================
+
+.. program:: rbdmap
+
+Synopsis
+========
+
+| **rbdmap**
+
```
branto1: We could use the basic usage info, here. I.e. rbdmap map | unmap.

doc/man/8/rbdmap.rst
```diff
@@ -0,0 +1,43 @@
...
+
+Description
+===========
+
+**rbdmap** is a shell script that is intended to be run at boot time
+by the init system (sysvinit, upstart, systemd). It looks for an
```
branto1: While it is true that it is usually run by init system, a user can also easily run it to achieve the same thing without having to have it run every time a computer starts -- that is why I originally kept it as a script in /usr/bin.

Alternatively, we could skip creating the man page altogether and move the script to /usr/lib/... (libexecdir). I'm not sure how useful the standalone script in bindir is and it depends mostly on that.

smithfarm:
> While it is true that it is usually run by init system, a user can also easily run it to achieve the same thing without having to have it run every time a computer starts -- that is why I originally kept it as a script in /usr/bin.
> 
> Alternatively, we could skip creating the man page altogether and move the script to /usr/lib/... (libexecdir). I'm not sure how useful the standalone script in bindir is and it depends mostly on that.

If there is a reasonable use case for running it manually (and I think there is), we should keep it where it is. I revamped the manpage a little - can you take another look?

branto1:
I also think there is a valid use case for this being in bindir (well, that is why I originally put it there). With the recent changes to the wording and after mentioning map/unmap in the man page, lgtm,

PR #8225, opened 2016-03-19
packaging: align radosgw package description

smithfarm:
First, make the Debian package description mention that RGW aims to
implement the Swift API.

Second, replace the RPM package description with the Debian one, both for
consistency and because the Debian one is better.


kmroz:
Looks good. Thanks Nathan.

PR #8227, opened 2016-03-20
xio: xio_init needs to be called before any other xio function

roidayan:
Latest accelio code removes the attribute to call xio_init on load
and XioMessenger will first try to initialize portals and create xio context.
This commit will fix this behavior and call xio_init first.

### DIFF
--- a/src/msg/xio/XioMessenger.cc
+++ b/src/msg/xio/XioMessenger.cc
@@ -248,11 +248,116 @@ static string xio_uri_from_entity(const string &type,
   return xio_uri;
 } /* xio_uri_from_entity */
 
+void XioInit::package_init(CephContext *cct) {
+   if (! initialized.read()) {
+
+     mtx.Lock();
+     if (! initialized.read()) {
+
+       xio_init();
+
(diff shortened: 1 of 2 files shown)


yuyuyu101:
I think common/Initialize.h could meet your demand.

cbodley:
@roidayan looks good to me, if I understand right: the issue is that the XioMessenger ctor was initializing its `portals` member before the body called xio_init() - so moving xio_init() into a base class ctor fixes the order of initialization

@yuyuyu101 I don't think static initialization will work here, since the xio init depends on a CephContext

roidayan:
@cbodley yes that's correct.

I want also later want to add a new commit to call xio_dtor from XioInit which we don't do today so I will need ~XioInit which I don't have in Initialize class.

cbodley:
@roidayan sounds good. there's already a `nInstances` counter that we should be able to use to tell when the last messenger destructs. it would also be nice to encapsulate that, `intialized`, and `mtx` as static members of XioInit while you're at it

PR #8229, opened 2016-03-21
recovery optimize for log-based recovery

sysnote:
This PR is based on #7325, overwrite Ops can be recover as partial content. Otherwise, run recovery as the normal process;check peer_features when set can_recover_partial in do_osd_ops, this can solve upgrade compitable problem.


athanatos:
This seems to be a less complete version of https://github.com/ceph/ceph/pull/7325.  I'm closing it for now.  If you disagree, a ceph-devel conversation would probably be a good idea.

Peerless-Jing:
have you ever run fio+librbd to test the performance of random write? I came across an error when running this, this error is like "rbd_open failed. fio_rbd_connect failed."
Anyway, could you please offer me a way to test the performance of random write? thx a lot! Looking forward to your replying!!

PR #8230, opened 2016-03-21
osd: fix bugs for omap ops

majianpeng:


src/osd/ReplicatedPG.cc
```diff
@@ -5774,7 +5774,7 @@ int ReplicatedPG::do_osd_ops(OpContext *ctx, vector<OSDOp>& ops)
       ctx->mod_desc.mark_unrollbackable();
       ++ctx->num_write;
       {
-	if (!obs.exists || oi.is_whiteout()) {
+	if (!obs.exists || oi.is_whiteout() || !oi.is_omap()) {
```
tchaikov: if the object does not have any omap entry, i think we can consider this op done, and "break" right away instead of returning a '-ENOENT'?

src/osd/ReplicatedPG.cc
```diff
@@ -5533,7 +5533,7 @@ int ReplicatedPG::do_osd_ops(OpContext *ctx, vector<OSDOp>& ops)
 	tracepoint(osd, do_osd_op_pre_omapgetkeys, soid.oid.name.c_str(), soid.snap.val, start_after.c_str(), max_return);
 	set<string> out_set;
 
-	if (pool.info.supports_omap()) {
+	if (obs.oi.is_omap()) {
```
tchaikov: could just reference `oi.is_omap()` instead.

tchaikov: and there are two "Signed-off-by: Jianpeng Ma <jianpeng.ma@intel.com>" in your commit message, could you remove one of them?

tchaikov:
i am not sure about f56e0d68cffb0f23df64c2381a46fbdae0c5c8a2.

src/osd/ReplicatedPG.cc
majianpeng: For this, you're right. Before change, if object don't have omap, it return 0. After this, it return -ENOENT. It change the semantics. So apply you said.

tchaikov:
lgtm

tchaikov:
lgtm

PR #8231, opened 2016-03-21
common: fire all timer events if the system time is adjusted back

wonzhq:
When the system time is adjusted to some time ago becasue of sync
or something else, some timers don't work properly, such as the
tick timer. Fire all the events if we see this situation.

Signed-off-by: Zhiqiang Wang <zhiqiang@xsky.com>

### DIFF
--- a/src/common/Timer.cc
+++ b/src/common/Timer.cc
@@ -85,8 +85,31 @@ void SafeTimer::timer_thread()
 {
   lock.Lock();
   ldout(cct,10) << "timer_thread starting" << dendl;
+  utime_t last;
   while (!stopping) {
+    // fire all the events already in the queue if the system time
+    // is changed to some time ago
     utime_t now = ceph_clock_now(cct);
+    if (now < last) {
+      last = now;
+      vector<Context *> callbacks;
+      for (scheduled_map_t::iterator p = schedule.begin();
+           p != schedule.end(); ++p)
+        callbacks.push_back(p->second);
+      events.clear();
+      schedule.clear();
+
+      if (!safe_callbacks)
+        lock.Unlock();
+      for (vector<Context *>::iterator it = callbacks.begin();
+           it != callbacks.end(); ++it)
+        (*it)->complete(0);
+      if (!safe_callbacks)
+        lock.Lock();
+      now = ceph_clock_now(cct); // recalc now
+    } else {
+      last = now;
+    }
     
     while (!schedule.empty()) {
       scheduled_map_t::iterator p = schedule.begin();


athanatos:
I think the right answer here is to actually use mono_time for the cases where we aren't scheduling for absolute time.

athanatos:
I'm closing it based on my above comment, reopen if you disagree!

wonzhq:
@athanatos yep, using monotonic time could be a better choice. But MacOS and FreeBSD don't support monotonic time...

wjwithagen:
On 4-5-2016 08:52, wonzhq wrote:
> @athanatos <https://github.com/athanatos> yep, using monotonic time
> could be a better choice. But MacOS and FreeBSD don't support monotonic
> time...

Not sure where you got that?

But at least for FreeBSD:

/usr/include/time.h:#define CLOCK_MONOTONIC     4
/usr/include/time.h:#define CLOCK_MONOTONIC_PRECISE     11      /* 
FreeBSD-specific. */
/usr/include/time.h:#define CLOCK_MONOTONIC_FAST        12      /* 
FreeBSD-specific. */

Read FAST == COARSE

--WjW

wonzhq:
@wjwithagen thanks for the info. I don't have a FreeBSD system, and get that from the internet (http://stackoverflow.com/questions/11338899/are-there-any-well-behaved-posix-interval-timers). Maybe it's a little old.

For MacOS, I just did a quick grep on my macbook, and it doesn't have CLOCK_MONOTONIC defined.

wjwithagen:
On 4-5-2016 14:57, wonzhq wrote:
> @wjwithagen <https://github.com/wjwithagen> thanks for the info. I don't
> have a FreeBSD system, and get that from the internet
> (http://stackoverflow.com/questions/11338899/are-there-any-well-behaved-posix-interval-timers).
> Maybe it's a little old.

Using Google: "FreeBSD manual MONOTONIC" is your friend. :)
https://www.freebsd.org/cgi/man.cgi?query=clock_gettime
And that is from the horse's mouth.

--WjW

adamemerson:
Use the timer in common/ceph_timer.h. It's templated, you can specify that it run on ceph::mono_clock and it will do exactly what you want. You'll have to make ticks use functors or lambdas or something rather than Context*, but we want to do that anyway. (And for ticks, if you use reschedule_me, it saves an allocation on every tick.)

See the timer in Objecter for an example.

wonzhq:
@adamemerson great, so we already have this!

Two quick questions:
1. Seems like this piece of code doesn't take MacOS into account?
2. Does the std condition variable use mono time by default in wait_until?

adamemerson:
1. ceph_timer.h should not take Mac OS into account. The correct way to take Mac OS into account would be to add another clause to mono_clock::now() to support it. There's an example of how to get monotonic time in libcxx we could base ours off of if we wanted to.

2. wait_until is a tempalte function. It takes a std::chrono::time_point which is itself a template. As far as the C++ standard library is concerned, there is no default. So, if you give it a real (wallclock) time, it waits until after that wallclock time. If you give it a monotonic time, it waits until that monotoic time comes around.

PR #8234, opened 2016-03-21
mon: calculate reporters_by_subtree when adjust_heartbeat_grace is false

wonzhq:
In the case mon_osd_adjust_heartbeat_grace is set to false, should
also calculate reporters_by_subtree.

### DIFF
--- a/src/mon/OSDMonitor.cc
+++ b/src/mon/OSDMonitor.cc
@@ -1766,16 +1766,6 @@ bool OSDMonitor::check_failure(utime_t now, int target_osd, failure_info_t& fi)
     for (map<int,failure_reporter_t>::iterator p = fi.reporters.begin();
 	 p != fi.reporters.end();
 	 ++p) {
-      // get the parent bucket whose type matches with "reporter_subtree_level".
-      // fall back to OSD if the level doesn't exist.
-      map<string, string> reporter_loc = osdmap.crush->get_full_location(p->first);
-      map<string, string>::iterator iter = reporter_loc.find(reporter_subtree_level);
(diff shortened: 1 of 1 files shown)


liewegas:
It seems like this shouldb e possible without cut and pasting the code.. can you refactor so there aren't two copies of the loop that builds the map?

wonzhq:
I was thinking to save some time to avoid looping the map `fi.reporters` twice in the case when mon_osd_adjust_heartbeat_grace is true. Does this matter? @liewegas

wonzhq:
Updated, pls take another look! @liewegas

athanatos:
@wonzhq Can you rebase?

wonzhq:
Rebased.

src/mon/OSDMonitor.cc
```diff
@@ -1777,6 +1777,7 @@ bool OSDMonitor::check_failure(utime_t now, int target_osd, failure_info_t& fi)
 	   << " + " << peer_grace << "), max_failed_since " << max_failed_since
 	   << dendl;
 
+
```
athanatos: @wonzhq This doesn't look right.  Did something go wrong with the rebase?

wonzhq:
Oops, the change in the second commit has been in the master. I've removed it from this PR.

tchaikov:
@wonzhq #8749 fixes the same issue in a different way, @Aran85 keeps just a single loop over `fi.reporters`. personally, i am good with both approaches. but seems @liewegas prefers the "single" loop one. so i am closing yours.

PR #8235, opened 2016-03-21
rbd-mirror: asok commands to get status and flush on Mirror and Replayer level

trociny:


src/tools/rbd_mirror/ImageReplayer.h
```diff
@@ -117,21 +117,21 @@ class ImageReplayer {
...
-   *    |
-   *    v
-   * <stopping>
-   *    |
-   *    v
-   * JOURNAL_REPLAY_SHUT_DOWN
-   *    |
-   *    v
+   *    v-----------------------------------------------\
+   * <replaying> --------------> <flushin_replay>       |
```
dillaman: s/flushin_replay/flushing_replay/g

trociny: Thanks! fixed.

PR #8236, opened 2016-03-21
ReplicatedPG: add thread timeout suspend/reset on time-cost fragments

cxwshawn:

### DIFF
--- a/src/osd/PG.h
+++ b/src/osd/PG.h
@@ -2306,7 +2306,9 @@ public:
     ThreadPool::TPHandle &handle
   ) = 0;
 
-  virtual void do_op(OpRequestRef& op) = 0;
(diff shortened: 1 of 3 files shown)


cxwshawn:
@tchaikov  @liewegas  please help review

athanatos:
@cxwshawn  We definitely can't put the hb handle into the OpContext, it gets passed to other threads.  Second, if the do_op thread blocks on read because the filesystem is hung, we actually want the osd to crash.  With this patch, a do_op thread hung in read might fail to kill the osd, which would be bad.  I'm closing this for now.  It does make sense to try to reduce false positives here, do you have data to indicate what kinds of slow reads are resulting in false positives?  That would be a good discussion for #ceph-devel

PR #8237, opened 2016-03-21
cmake: fix mrun to handle cmake build structure

oritwas:

### DIFF
--- a/src/mrun
+++ b/src/mrun
@@ -5,7 +5,14 @@
 root=`dirname $0`
 run_name=$1
 command=$2
+CEPH_BIN=$root
+CEPH_CONF_PATH=$root/run/$run_name
+
+if [ -e CMakeCache.txt ]; then
+    CEPH_BIN=$PWD/src
+    CEPH_CONF_PATH=$PWD/run/$run_name
+fi
 
 shift 2
 
-$root/$command -c $root/run/$run_name/ceph.conf "$@"
+$CEPH_BIN/$command -c $CEPH_CONF_PATH/ceph.conf "$@"


oritwas:
@cbodley I pushed a newer version that handles the conf file too

src/mrun
```diff
@@ -1,11 +1,18 @@
...
 [ $# -lt 2 ] && echo "usage: $0 <name> <command> [params...]" && exit 1
 
 root=`dirname $0`
 run_name=$1
 command=$2
+CEPH_BIN=$root
+CEPH_CONF_PATH=$root/run/$run_name
+
+if [ -e CMakeCache.txt ]; then
+    CEPH_BIN=$PWD/src
```
cbodley: note that this bit is going to change to `$PWD/bin` once @alimaredia's cmake check work (https://github.com/ceph/ceph/pull/7912) gets merged

src/mrun
```diff
@@ -1,11 +1,18 @@
-#!/bin/sh
+#!/bin/sh -x
```
cbodley: helpful for debugging, but we probably don't want to merge this change

cbodley:
thanks @oritwas! for my multisite testing, I had been hacking around this with symlinks. after removing those and testing this pr, there are still a couple issues running ceph/src/test/rgw/test_multi.py. but I'll add fixes for those in a separate pull request - this one looks good :+1:

cbodley:
see https://github.com/ceph/ceph/pull/8246 for the related multisite test fixes

oritwas:
@cbodley removed the -x

PR #8238, opened 2016-03-21
mon: fix mixed-version MDSMonitor

jcsp:
Encode old-style until mon quorum features indicate
jewel, forbid creating multiple filesystems, and
when creating a single filesystem use the legacy fscid.

Fixes: #15223


jcsp:
Oops, meant to target jewel.

PR #8241, opened 2016-03-21
vstart: support independent configuration and startup of rgw instances

kmroz:
RGW was a bit neglected in some of our scripting. Trying to remedy this.

This PR:
1. Allows restarting rgw independently of the cluster.
2. Tries to provide better support for multiple rgw instances (same
   cluster, or multi-cluster via mstart.sh) by providing independent config
   sections, etc.
3. Allows setting base port number on which to listen (instances will
   increment base port by instance number).
4. Changing base port number when re-starting overrides rgw_frontends
   allowing new instance to run on changed port.
5. Generates proper pid/log/asok/etc files in out/client.rgw.\* (needed to properly stop rgw instances via mstop.sh)

Fixes: http://tracker.ceph.com/issues/15226

### DIFF
--- a/src/vstart.sh
+++ b/src/vstart.sh
@@ -377,7 +377,11 @@ $SUDO rm -f core*
 
 test -d $CEPH_OUT_DIR || mkdir $CEPH_OUT_DIR
 test -d $CEPH_DEV_DIR || mkdir $CEPH_DEV_DIR
-$SUDO rm -rf $CEPH_OUT_DIR/*
+
+# Only clear out pidfiles/heartbeats/logs/etc on a $new run.
+if [ "$new" -eq 1 ]; then
+    $SUDO rm -rf $CEPH_OUT_DIR/*
+fi
 test -d gmon && $SUDO rm -rf gmon/*
 
 [ "$cephx" -eq 1 ] && [ "$new" -eq 1 ] && test -e $keyring_fn && rm $keyring_fn
--- a/src/vstart.sh
+++ b/src/vstart.sh
@@ -141,6 +141,7 @@ usage=$usage"\t-o config\t\t add extra config parameters to all sections\n"
 usage=$usage"\t--mon_num specify ceph monitor count\n"
 usage=$usage"\t--osd_num specify ceph osd count\n"
 usage=$usage"\t--mds_num specify ceph mds count\n"
+usage=$usage"\t--rgw_num specify ceph rgw count\n"
 usage=$usage"\t--rgw_port specify ceph rgw http listen port\n"
 usage=$usage"\t--bluestore use bluestore as the osd objectstore backend\n"
 usage=$usage"\t--memstore use memstore as the osd objectstore backend\n"
@@ -213,6 +214,10 @@ case $1 in
             ;;
     --mds_num )
             CEPH_NUM_MDS=$2
+            shift
+            ;;
+    --rgw_num )
+            CEPH_NUM_RGW=$2
             shift
             ;;
     --rgw_port )
--- a/src/vstart.sh
+++ b/src/vstart.sh
@@ -102,6 +102,7 @@ start_mon=0
 start_mds=0
 start_osd=0
 start_rgw=0
+rgw_port_override=0
 ip=""
 nodaemon=0
 smallmds=0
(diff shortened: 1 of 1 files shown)


src/vstart.sh
```diff
@@ -617,6 +635,29 @@ EOF
...
+if [ "$start_rgw" -eq 1 ]; then
+    for rgw in `seq 0 $((CEPH_NUM_RGW-1))`
+    do
+	if [ "$new" -eq 1 ]; then
+            mkdir -p $CEPH_DEV_DIR/client.rgw.$rgw
+	    key_fn=$CEPH_DEV_DIR/client.rgw.$rgw/keyring
+	    if [ $overwrite_conf -eq 1 ]; then
+		    cat <<EOF >> $conf_fn
+[client.rgw.$rgw]
+        rgw frontends = fastcgi, civetweb port=$((CEPH_RGW_PORT+$rgw))
```
theanalyst: should we drop fastcgi here?

kmroz: @theanalyst - I wondered about this also, but ultimately left it as it was in the [global] section.

theanalyst:
LGTM

kmroz:
@theanalyst - Cool, thanks for reviewing.

kmroz:
@liewegas / @yehudasa - Sage, Yehuda - Both of you guys put a lot into vstart.sh. If you have a moment, a review of this from you would be awesome. Thanks!

kmroz:
Hey RGWers - Anything else needed to improve this PR?

liewegas:
@yehudasa ping

kmroz:
A gentle bump :)

kmroz:
@yehudasa - ping++

cbodley:
i gave this a try and was able to verify that the `--rgw_port` override stuff works as expected. one thing that didn't work, though, was increasing the `--rgw_num` after the initial run

this has to do with adding separate `[client.rgw.#]` sections to ceph.conf for each gateway. it seems like the `rgw_port_override` stuff would be a lot simpler if we didn't do this, and instead passed the `--rgw_frontends` argument explicitly on the command line. you can still use `--name client.rgw.#` to get separate log files. do you buy that?

kmroz:
@cbodley - Yes, you're right. In my mind, `--rgw_num` is something that's relevant when provisioning the test cluster (ie. when `-n` is specified). My thinking was that manipulating the number of daemon instances would warrant tear down of the cluster, followed by re-provisioning. Anyway, I'm flexible. We can aim to have `--rgw_num` manipulate the number of instances even on restart if you think it's a worthwhile bit.

As to `--rgw-frontends`, I was hoping to abstract this from the user.

I will re-visit this PR more thoroughly, hopefully this week.

Thanks for the review!

cbodley:
@kmroz agreed that other ceph daemons need to be provisioned for in the initial setup, but radosgw is more of a client than a daemon in this sense. you can add or remove gateways at runtime, and they'll cooperate through rados to share the storage

i doubt many people will need to add/remove gateways in a vstart cluster, but i think it would be nice to support as long as we don't sacrifice other features. which brings us to..

> As to `--rgw-frontends`, I was hoping to abstract this from the user.

by abstracting this, you mean having ceph.conf remember the port number so they don't have to specify a `--rgw_port` or `CEPH_RGW_PORT` next time?

i can see this being useful in a multisite setup; the gateway endpoints have to be included in the zone configuration, so you don't want the ports changing.. we've been using mstart.sh for multisite testing, which runs vstart.sh for multiple clusters and overrides `CEPH_RGW_PORT` for each (there's also src/test/rgw/test_multi.py that runs mstart.sh)

yehudasa:
@kmroz @cbodley status of this PR?

kmroz:
@yehudasa @cbodley - Sorry, was MIA for some time. Let me go over the notes of this PR again. As I recall, the major problem with my approach here (pointed out by Casey) was that I was treating RGW as a daemon (akin to MONs and OSDs) where really it is a Ceph client.

dachary:

liewegas:
@kmroz ping

kmroz:
@sage - I think this should be closed.  It's horribly outdated.

PR #8242, opened 2016-03-21
read cluster_name.conf for multi-cluster

jazeltq:
Right now, the unit service  will find the ceph.conf by default.
So add -c to read cluster_name.conf.


trociny:
I failed to see the purpose of this patch. When `--cluster XXX` option is specified, `XXX.conf` is used.

jdurgin:
If --cluster foo isn't reading foo.conf for any of these commands, we should fix that (it's a bug) rather than passing -c

jazeltq:
yes.

PR #8244, opened 2016-03-21
rgw_ldap fixes

mattbenjamin:
1. only attempt auth on tokens which pass valid() check
2. ignore false positive search result (and fix result in that case)


src/rgw/rgw_token.cc
```diff
@@ -128,7 +128,7 @@ int main(int argc, char **argv)
 	std::cout << "decoded expanded token: " << token2 << std::endl;
       }
     }
-    std::cout << to_base64(token_str) << std::endl;
```
yehudasa: @mattbenjamin shouldn't this be under the `if (verbose)` section?

PR #8245, opened 2016-03-21
rgw: fix error message for zone delete

theanalyst:
Currently if a zone is not a part of a zonegroup, an error message
is printed that zone doesn't exist, and the zone gets deleted anyway.
Since this is exhibited only if the zone isn't a part of a zonegroup
and we allow creation of such zones, clarify that the zone wasn't a
part of zonegroup instead

Fixes: #14951

EDIT: added fixes http://tracker.eph.com/issues/14951


theanalyst:
@yehudasa / @oritwas I'm assuming we allow creation of zones without attaching them to a zonegroup, if not this may not be the direction for the fix

PR #8246, opened 2016-03-21
rgw: multisite tests don't change working directory

cbodley:
this allows the multisite tests (src/test/rgw/test_multi.py) to be invoked from the cmake build directory, and use pwd to find the `run` directory used by mstart.sh and friends

### DIFF
--- a/src/mrgw.sh
+++ b/src/mrgw.sh
@@ -11,7 +11,7 @@ port=$2
 
 shift 2
 
-run_root=$script_root/run/$name
+run_root=run/$name
 pidfile=$run_root/out/radosgw.${port}.pid
 asokfile=$run_root/out/radosgw.${port}.asok
 logfile=$run_root/out/radosgw.${port}.log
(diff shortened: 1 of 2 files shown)


src/mrgw.sh
```diff
@@ -11,7 +11,7 @@ port=$2
 
 shift 2
 
-run_root=$script_root/run/$name
+run_root=run/$name
```
yehudasa: @cbodley does it mean that running mrgw.sh from different locations will end up with different instances?

cbodley: that's correct. though in the cmake case, running mrgw.sh from anywhere but the cmake build directory will fail because it doesn't know how the find the binaries. in the automake case, I assume that it will fail if it doesn't see a ./run directory from mstart

yehudasa: @cbodley I'm not sure that the way to go. We don't want to force users who use the different m* tools (mrun, mstart.sh, mrgw.sh) to run everything from the same directory. If that's a problem with cmake, we may need to add new configurables, either via a config file, or via yet another env variable.

cbodley:
closing for now, this needs more thought

PR #8247, opened 2016-03-21
mon: do not send useless pg_create messages for split pgs

liewegas:


src/mon/PGMonitor.cc
```diff
@@ -1256,11 +1256,14 @@ epoch_t PGMonitor::send_pg_creates(int osd, Connection *con, epoch_t next)
              << dendl;
     last = q->first;
     for (set<pg_t>::iterator r = q->second.begin(); r != q->second.end(); ++r) {
+      pg_stat_t &st = pg_map.pg_stat[*r];
+      if (st.parent_split_bits)
+	continue;
```
tchaikov: can we skip the split PGs in `PGMap::stat_pg_add()` at the very beginning? the `pg_map.pg_stat` is always updated at that moment.

liewegas: Which call to stat_pg_add?  I'm not sure what you  mean by beginning.

tchaikov: i am wondering if we can have something like
```diff
--- a/src/mon/PGMap.cc
+++ b/src/mon/PGMap.cc
@@ -470,7 +470,8 @@ void PGMap::stat_pg_add(const pg_t &pgid, const pg_stat_t &s,
 
   if (s.state & PG_STATE_CREATING) {
     creating_pgs.insert(pgid);
-    if (s.acting_primary >= 0) {
+    if (s.acting_primary >= 0 &&
+	!pg_stat[pgid].parent_split_bits)) {
       creating_pgs_by_osd_epoch[s.acting_primary][s.mapping_epoch].insert(pgid);
     }
   }
```

liewegas: Oh, I see.  Yeah, that would be much better!

tchaikov:
lgtm

liewegas:
I think the original was wrong,a ctually.

* moved condition to outer if, so we don't populate creating_pg set
* added matching condition to stat_pg_sub too

PR #8250, opened 2016-03-21
python: avoid long paths

dachary:
As a temporary measure because it overflows when jenkins tries to build
in a deep path. Run make with:

make CEPH_BUILD_VIRTUALENV=/tmp

and it will use these directories instead of the default.

### DIFF
--- a/src/ceph-detect-init/Makefile.am
+++ b/src/ceph-detect-init/Makefile.am
@@ -53,13 +53,15 @@ EXTRA_DIST += \
 	ceph-detect-init/tests/test_all.py \
 	ceph-detect-init/tox.ini
 
-ceph-detect-init-all: ceph-detect-init/virtualenv
+export CEPH_DETECT_INIT_VIRTUALENV = ${CEPH_BUILD_VIRTUALENV}/ceph-detect-init-virtualenv
 
-ceph-detect-init/virtualenv:
-	cd $(srcdir)/ceph-detect-init ; ../tools/setup-virtualenv.sh ; test -d wheelhouse && export NO_INDEX=--no-index ; virtualenv/bin/pip install $$NO_INDEX --use-wheel --find-links=file://$$(pwd)/wheelhouse -e .
+ceph-detect-init-all: ${CEPH_DETECT_INIT_VIRTUALENV}
+
+${CEPH_DETECT_INIT_VIRTUALENV}:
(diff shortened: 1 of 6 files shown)


alfredodeza:
this means that the build needs to specify these dirs now? If this is temporary how can we make sure to revert these env vars when they are no longer needed?

alfredodeza:
If the exporting of environment variables are absolutely required then they will need to be added to the build scripts.

For RPMs: https://github.com/ceph/ceph-build/blob/master/ceph-build/build/build_rpm
For DEBs: https://github.com/ceph/ceph-build/blob/master/ceph-build/build/build_deb

src/ceph-detect-init/Makefile.am
```diff
@@ -55,11 +55,13 @@ EXTRA_DIST += \
 
 ceph-detect-init-all: ceph-detect-init/virtualenv
 
+CEPH_DETECT_INIT_VIRTUALENV := .
```
alfredodeza: can we have just one environment variable? Something like `CEPH_BUILD_VIRTUALENVS` and then specify the actual virtualenv for ceph-detect-init like:

    ${CEPH_BUILD_VIRTUALENVS}/ceph-detect-init/

dachary:
> If the exporting of environment variables are absolutely required then they will need to be added to the build scripts.

This is a jenkins specific workaround, you should indeed add them to the jenkins build files. Please let me know if that fixes the issue.

src/ceph-detect-init/Makefile.am
alfredodeza: @ldachary I didn't see this one addressed. Care to comment (or fix?) :)

dachary: I think best to have this hack on a per-module basis rather than a general thing shared among modules. Less chances for mistakes.

alfredodeza: the proposed solution would allow a per-module basis:

    ${CEPH_BUILD_VIRTUALENVS}/ceph-detect-init/
    ${CEPH_BUILD_VIRTUALENVS}/ceph-disk/

The only thing a build system would need to define is the parent virtualenv dir

dachary:
repushed with the required changes

src/ceph-detect-init/Makefile.am
```diff
@@ -55,11 +55,17 @@ EXTRA_DIST += \
 
 ceph-detect-init-all: ceph-detect-init/virtualenv
 
+ifeq ($(origin CEPH_BUILD_VIRTUALENV), undefined)
+  CEPH_DETECT_INIT_VIRTUALENV = .
+else
```
alfredodeza: this is not valid here because it needs an `if` not an `ifeq`, or the other way around, it would need a `else ifeq`

It breaks the build too: https://jenkins.ceph.com/job/ceph-pull-requests/3461/console

    src/ceph-detect-init/Makefile.am:60: error: else without if

dachary:
the make check failure is environmental, ./run-make-check.sh on a new machine succeeds.

PR #8251, opened 2016-03-22
vstart: make -k with optional mon_num.

majianpeng:
Default mon_num=3, if we change this and using -k to create a new
cluster. It don't work. So we must record the old mon_num.
The same for osd_num/mds_num/rgw_num.


src/vstart.sh
```diff
@@ -275,6 +275,13 @@ esac
 shift
 done
 
+if [ "$overwrite_conf" -eq 0 ]; then
+  CEPH_NUM_MON=`cat $conf_fn|grep CEPH_NUM_MON|awk -F= '{print $2}'`
```
tchaikov: nit, could just put
```awk
CEPH_NUM_MON=`awk -F= '/CEPH_NUM_MON/{print $2}' $conf_fn`
```

tchaikov:
lgtm apart from a nit.

dillaman:
This PR now causes `MON=1 OSD=1 MDS=0 ./vstart.sh -k` to fail:

`./vstart.sh: line 428: [: 1: unary operator expected`

src/vstart.sh
```diff
@@ -275,6 +275,13 @@ esac
 shift
 done
 
+if [ "$overwrite_conf" -eq 0 ]; then
+  CEPH_NUM_MON=`awk -F= '/CEPH_NUM_MON/{print $2}' $conf_fn`
```
dillaman: Can we default these if the `awk` fails?

tchaikov: ahh, i overlooked this. will fix it tomorrow.

PR #8253, opened 2016-03-22
rgw: Do not send a Content-Type on a '304 Not Modified' response

wido:
When we say the Content has not changed we should not respond
with a content type which defaults to binary/octet stream.

Fixes: #15119

This should fix issue http://tracker.ceph.com/issues/15119


smithfarm:
@wido: I tried to backport this to hammer but fb4e5cc does not cherry-pick cleanly and I'm not sure if `status_num` is defined in `RGWMongoose::complete_request()` in hammer. Could you do the backport? It's http://tracker.ceph.com/issues/15283

PR #8259, opened 2016-03-22
rgw: return -EACCESS for system requests also

theanalyst:
In a multisite scenario, if a user created in a secondary zone tries to
create a bucket, fail with AccessDenied instead of a NoSuchKey, which
doesn't make sense for a create Bucket request for eg.

Fixes: #15234

### DIFF
--- a/src/rgw/rgw_rest_s3.cc
+++ b/src/rgw/rgw_rest_s3.cc
@@ -3600,7 +3600,7 @@ int RGW_Auth_S3::authorize_v4(RGWRados *store, struct req_state *s)
       int ret = rgw_get_user_info_by_uid(store, effective_uid, effective_user);
       if (ret < 0) {
         ldout(s->cct, 0) << "User lookup failed!" << dendl;
-        return -ENOENT;
+        return -EACCES;
       }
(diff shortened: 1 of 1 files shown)


theanalyst:
There is also a place in swift code where we return a -ENOENT, however on local testing,  fixing the S3, seemed to work for Swift requests as well as when a zone forwards the request we automatically use the S3 API.

theanalyst:
test this please

theanalyst:
@yehudasa / @cbodley WDYT?

cbodley:
I was under the impression that we were handling user creation/sync similar to the way we handle buckets, so I'm surprised to see that secondary zones can have their own private users. Is this intended @yehudasa?

theanalyst:
> I was under the impression that we were handling user creation/sync similar to the way we handle buckets, so I'm surprised to see that secondary zones can have their own private users. 

at least what I observed is that you could create users in secondary zone and this doesn't sync back to the master zone, (I kind of mentioned that in the documentation that we have to always create users in master zone), not sure which log (mdlog?) syncs the user info and if it is bidirectional?

PR #8261, opened 2016-03-22
rbd-mirror: enabling/disabling pool mirroring should update the mirroring directory

rjfd:
Fixes #15217

Ticket url: http://tracker.ceph.com/issues/15217

### DIFF
--- a/src/librbd/internal.cc
+++ b/src/librbd/internal.cc
@@ -2779,6 +2779,34 @@ int validate_mirroring_enabled(ImageCtx *ictx) {
     return 0;
   }
 
+  int list_mirror_images(IoCtx& io_ctx,
+                         std::set<std::string>& mirror_image_ids) {
+    CephContext *cct = reinterpret_cast<CephContext *>(io_ctx.cct());
+
+    std::string last_read = "";
+    int max_read = 1024;
+    int r;
+    do {
+      std::map<std::string, std::string> mirror_images;
+      r =  cls_client::mirror_image_list(&io_ctx, last_read, max_read,
+                                             &mirror_images);
+      if (r < 0) {
+        lderr(cct) << "error listing mirrored image directory: "
+             << cpp_strerror(r) << dendl;
+        return r;
+      }
+      for (auto it = mirror_images.begin(); it != mirror_images.end(); ++it) {
+        mirror_image_ids.insert(it->first);
+      }
+      if (!mirror_images.empty()) {
+        last_read = mirror_images.rbegin()->first;
+      }
+      r = mirror_images.size();
+    } while (r == max_read);
+
+    return 0;
+  }
+
   int mirror_mode_set(IoCtx& io_ctx, rbd_mirror_mode_t mirror_mode) {
     CephContext *cct = reinterpret_cast<CephContext *>(io_ctx.cct());
     ldout(cct, 20) << __func__ << dendl;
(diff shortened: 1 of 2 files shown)


src/librbd/internal.cc
```diff
@@ -2797,6 +2822,58 @@ int validate_mirroring_enabled(ImageCtx *ictx) {
...
+        r = img_ctx->state->open();
+        if (r < 0) {
+          lderr(cct) << "error opening image "<< img_pair.first << ": "
+            << cpp_strerror(r) << dendl;
+          delete img_ctx;
+          continue;
+        }
+
+        uint64_t features;
+        r = get_features(img_ctx, &features);
```
dillaman: Slight simplification: you have inside librbd and have direct access to the ImageCtx.  Just use `ImageCtx::test_features`

src/librbd/internal.cc
```diff
@@ -2797,6 +2822,58 @@ int validate_mirroring_enabled(ImageCtx *ictx) {
...
+        r = cls_client::mirror_mode_set(&io_ctx, current_mirror_mode);
+        if (r < 0) {
+          lderr(cct) << "Failed to rollback mirror mode: " << cpp_strerror(r)
+            << dendl;
+        }
+        return r;
+      }
+
+      for (const auto& img_pair : images) {
+        ImageCtx *img_ctx = new ImageCtx("", img_pair.second, nullptr, io_ctx,
```
dillaman: Might be nice to just directly query the image's features instead of fully opening the image (which requires a lot of back-and-forth). Use `librbd::cls_client::get_features` and pass it the header oid (computed from the image ID extracted from the directory via `librbd::util::header_name`) and CEPH_NOSNAP for the snap id.

src/librbd/internal.cc
```diff
@@ -2872,6 +2872,55 @@ int validate_mirroring_enabled(ImageCtx *ictx) {
...
+      }
+
+      for (const auto& img_id : image_ids) {
+        ImageCtx *img_ctx = new ImageCtx("", img_id, nullptr, io_ctx, false);
+        r = img_ctx->state->open();
+        if (r < 0) {
+          lderr(cct) << "error opening image id "<< img_id << ": "
+            << cpp_strerror(r) << dendl;
+          delete img_ctx;
+          continue;
```
dillaman: I think if you have any error disabling mirroring on an image you just need to abort -- don't want to leave the pool / images in an inconsistent state.  Same comment below.

src/librbd/internal.cc
```diff
@@ -2872,6 +2872,55 @@ int validate_mirroring_enabled(ImageCtx *ictx) {
...
+        if (r < 0) {
+          lderr(cct) << "Failed to set mirror mode to image: "
+            << cpp_strerror(r) << dendl;
+          return r;
+        }
+        needs_rollback = true;
+      }
+
+      std::vector<std::string> image_ids;
+      r = cls_client::mirror_image_list(&io_ctx, &image_ids);
```
dillaman: I'm almost tempted to say that if the mirror mode is per-image and they still have images w/ mirroring enabled, you should abort out.

src/librbd/internal.cc
rjfd: I agree

src/librbd/internal.cc
rjfd: I'm not convinced that we should that. 
Also, why do we need to have two mirroring modes (pool / image) at the pool level? In my understanding, the mirroring state of an image is what decides if the image is mirrored or not, so why not just have enabled/disable states for the pool as well?

dillaman: There is the pool setting and the per-image setting (and really the per-image setting isn't a setting, it's the advertisement in the rbd_mirroring directory).

If you have explicit control over which images are mirrored, you set the pool to image-mode.  If you want all images to automatically be mirrored (assuming the images have journaling enabled), set it to image-mode.  When in pool-mode, it really doesn't make sense to allow the user to run "rbd mirror image enable / disable" so we should prohibit it.  Similarly, if you are disabling image mode but you haven't already manually disabled mirroring on all images, we should remind them to do it.

rjfd: Just to be clear, when you said:
> If you want all images to automatically be mirrored (assuming the images have journaling enabled), set it to image-mode
You wanted to say, "set it to pool-mode" right?

If that is the case, I agree with you, that we should prohibit "rbd mirror image enable/disable" when in pool-mode.

dillaman: D'oh -- yes, when in pool-mode, we probably shouldn't allow the user to play with "rbd mirror image enable / disable" since that is effectively putting it into a quasi-image-mode.

rjfd: Great! Should I create a new ticket to change the behavior of "rbd mirror image enable / disable" according to what we discussed above?

rjfd:
Pushed new changes to the previous implementation.

src/librbd/internal.cc
dillaman: That would be great. Thx!

rjfd:
@dillaman  I forgot to implement the behavior of when you change from MODE_IMAGE to MODE_DISABLED. I will abort in the case that there is at least one image with mirroring enabled.

I will notify you when I push the additional changes. Thanks

rjfd:
@dillaman I'm done with the changes. You can review when you have the time.
Thanks!

rjfd:
Two of the unit tests pushed in this PR failed. But I just ran them on my local environment and passed! Dunno what's happening.

rjfd:
Solved the problem, and already pushed the fix.

src/librbd/internal.cc
```diff
@@ -2797,11 +2845,155 @@ int validate_mirroring_enabled(ImageCtx *ictx) {
...
+        if (r < 0) {
+          lderr(cct) << "error getting features for image " << img_pair.first
+            << ": " << cpp_strerror(r) << dendl;
+          return r;
+        }
+
+        if ((features & RBD_FEATURE_JOURNALING) != 0) {
+          ImageCtx *img_ctx = new ImageCtx("", img_pair.second, nullptr,
+                                           io_ctx, false);
+          r = img_ctx->state->open();
```
dillaman: Is it possible to enable it w/o fully opening the image?  I had to write a similar helper for one of my open PRs:
https://github.com/ceph/ceph/pull/8287/commits/643ae7e1c37aa18b431174187a6824b324951948#diff-254555dde8dcfb7fb908791ab8214b92R203

src/librbd/internal.cc
```diff
@@ -2797,11 +2845,155 @@ int validate_mirroring_enabled(ImageCtx *ictx) {
...
+          }
+
+          rb_state.img_ctxs.push_back(std::make_pair(img_ctx, true));
+        }
+      }
+    }
+
+    rb_state.do_rollback = false;
+
+    if (next_mirror_mode != cls::rbd::MIRROR_MODE_IMAGE) {
```
dillaman: Perhaps just return on line 2854 to avoid this extra check.

src/librbd/internal.cc
rjfd: Yes, I could use a helper similar to yours, but I still have to check if this image is the primary using the Journal<>::is_tag_owner, which also needs to be adapted to not require an ImageCtx. Am I right?

dillaman: Good point --- probably not worth creating tons of helper methods to optimize for this case.

rjfd:
Fixed the commits and rebased on top of master.

PR #8262, opened 2016-03-22
os/bluestore: add a new config option, bluestore_no_device_flush

jjhuo:
To avoid flush/sync on whole device for some setup which doesn't need this at all.
And removed some obsolete bluestore config options which may confuse users.


liewegas:
what devices do you ahve in mind?  won't a flush/fsync on those devices be a no-op?

maybe the option should be bluestore_allow_buffered_writes or something.  also, this may cause problems with the wal path, which relies on the cache/buffer for correctness (it may do mulitple r/m/w's on the same block)

jjhuo:
@liewegas, Sage, please kindly see my commit msg below, thanks!

Currently, in BlueStore _kv_sync_thread, it will flush/sync the whole block device before committing every batched kv transaction. For 4KB rbd block workloads on high end SSD, this flush/sync could happen hundreds or thousands of times every second, since RocksDB batched kv commit is also very fast. Each flush/sync operation not only will walk through kernel page cache radix tree and flush dirty page cache, and also it will flush the SSD internal disk cache(expensive too).

For rbd block workloads on SSD, there are situations where those flush/sync are totally not necessary, for example, if bluestore_min_alloc_size is set 4KB, bluestore WAL is avoided, and all writes are direct AIO by default. There is no reason to flush page cache. In addition to that, many enterprise SSDs are
equiped with power loss protection capacitors to guarantee that write data cached inside SSDs will be flushed into Flash memory fully in case of power failure. So there is no need to flush disk cache either.

This patch will add a new config option, bluestore_no_device_flush, for users who need to avoid those flush/sync for their setup and device.

yuyuyu101:
actually, we want to use pagecache since it will be any cache in upper layer

jjhuo:
@yuyuyu101 This new option doesn't conflict with bluestore_default_buffered_read, with or without this new option set, bluestore can still do buffered read and store all the read into page cache. This patch only eliminate flush/sync after AIO writes for some setup who doesn't need them.
BTW, mixing buffered read and direct IO write is not that recommended by kernel guys.

liewegas:
> BTW, mixing buffered read and direct IO write is not that recommended by kernel guys.

The problem is with mixing concurrent buffered and direct IO to the same 
pages--which we don't do.  Or buffered reads triggered by readahead, which 
we disable.

jjhuo:
@liewegas After discussed this topic in Hackathon, I have ran some benchmarking tests to evaluate the performance impact of fdatasync on whole device(200 times per second) while writing SSD heavily with 4KB random IO, I only saw about 2.5% performance drop. Since disk cache flushing mechanism varies in different devices, I am closing this PR.

PR #8264, opened 2016-03-22
radosgw-admin: 'period commit' supplies user-readable error messages

cbodley:


cbodley:
added a commit to explain EACCES on 'realm pull', re: http://tracker.ceph.com/issues/15200

PR #8265, opened 2016-03-22
config_opts: disable filestore throttle soft backoff by default

athanatos:
Performance testing suggests that HDD osds don't benefit and
in some cases are slowed down.

Also, adjust the other defaults to make more sense.

### DIFF
--- a/src/common/config_opts.h
+++ b/src/common/config_opts.h
@@ -1058,18 +1058,18 @@ OPTION(filestore_queue_max_bytes, OPT_U64, 100 << 20)
 OPTION(filestore_caller_concurrency, OPT_INT, 10)
 
 /// Expected filestore throughput in B/s
-OPTION(filestore_expected_throughput_bytes, OPT_DOUBLE, 100 << 20)
+OPTION(filestore_expected_throughput_bytes, OPT_DOUBLE, 200 << 20)
 /// Expected filestore throughput in ops/s
-OPTION(filestore_expected_throughput_ops, OPT_DOUBLE, 100)
+OPTION(filestore_expected_throughput_ops, OPT_DOUBLE, 200)
 
-/// Filestore max delay multiple (probably don't need to change)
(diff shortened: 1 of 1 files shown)


athanatos:
Tested, /a/samuelj-2016-03-21_13:58:56-rados-wip-sam-testing-distro-basic-smithi

liewegas:
The current settings make more sense for SSD's I take it?

I wonder if we should bifurcate these options into 'hdd' and 'ssd' versions, and make filestore select the appropriate set of settings based on the /sys/block/.../queue/rotational flag?  It's a bit tedious to map the data path back to a block device, but I want the same thing for bluestore for similar reasons...

athanatos:
@liewegas 

Actually, the current settings make sense for nothing.  This set of configs just does a better job of matching the old ones (so that people don't see regressions).  I don't yet have a good general set of ssd configs, we've already seen that nvme and ssds seem to require different configs from each other.  For now, I'd like to at least not have a regression on spinning.

markhpc:
The only really obvious improvements we saw with the Intel NVMe drives on the incerta nodes was in small 50% mixed random workloads, where somnath's tuning (ie throttles wide open) appeared to be about 12% faster than the old default and about 6% faster than Sam's tuning.  Even those improvements might be suspect though as fio seems to be having a clock/timing issue that may cause some processes to run for longer than others.  In some cases we've had to go back and generate our own run averages from the throughput logs with the tails chopped off to get accurate results.  I suspect the improvement is real though since we see it gradually improve as the IO size increases in multiple write tests (This issue often manifests as a lone spikey sample).

On HDDs we can gain what appears to be a slight advantage in large write throughput by opening the throttles up, but it's probably not enough of one to be worth the memory and/or other potential issues that come with it.

Over all, at least on the incerta nodes, the performance variation between all of these tunings on both HDDs and SSDs was slight compared (especially in the write path) to all of the other improvements that have been made since hammer.  I believe Somnath is seeing bigger differences on his setup, so it may be that some hardware may be more reliant on optimized settings than others.

PR #8267, opened 2016-03-22
osd: bail out of _committed_osd_maps if we are shutting down

athanatos:
Introduced in b839a06c1aba24ecf0780387a605192ff929f158
Fixes: 14309


athanatos:
Tested: /a/samuelj-2016-03-21_13:58:56-rados-wip-sam-testing-distro-basic-smithi/78333/remote

PR #8268, opened 2016-03-23
rgw: convert plain object to versioned (with null version) when removing

yehudasa:
Fixes #15243

When removing a plain null versioned object (was created prior to bucket versioning
enabled), need to convert the bucket index representation to a versioned one. This
is needed so that all the versioning mechanics play together.

### DIFF
--- a/src/cls/rgw/cls_rgw.cc
+++ b/src/cls/rgw/cls_rgw.cc
@@ -1073,6 +1073,9 @@ public:
     initialized = true;
   }
 
+  void set_epoch(uint64_t epoch) {
+    instance_entry.versioned_epoch = epoch;
+  }
 
   int unlink_list_entry() {
     string list_idx;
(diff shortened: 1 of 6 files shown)


src/cls/rgw/cls_rgw.cc
```diff
@@ -1018,6 +1018,7 @@ static int write_obj_entries(cls_method_context_t hctx, struct rgw_bucket_dir_en
   }
   string instance_list_idx;
   get_list_index_key(instance_entry, &instance_list_idx);
+        instance_idx.c_str(), instance_list_idx.c_str(), (int)instance_entry.versioned_epoch);
```
oritwas: missing line maybe?

yehudasa:
@oritwas pushed a fix. That was an extra leftover line from a debug message that I missed when cleaned up things.

PR #8269, opened 2016-03-23
tools/rados: add parameter "--obj-offset" for put cmd.

majianpeng:
In development, we need write object at options offset. But 'rados put'
don't support this. So add write command w/ options offset.

### DIFF
--- a/src/tools/rados/rados.cc
+++ b/src/tools/rados/rados.cc
@@ -82,7 +82,7 @@ void usage(ostream& out)
 "\n"
 "OBJECT COMMANDS\n"
 "   get <obj-name> [outfile]         fetch object\n"
-"   put <obj-name> [infile]          write object\n"
+"   put <obj-name> [infile]  [--offset offset]  write object start offset(default 0)\n"
 "   truncate <obj-name> length       truncate object\n"
 "   create <obj-name>                create object\n"
 "   rm <obj-name> ...[--force-full]  [force no matter full or not]remove object(s)\n"
@@ -389,7 +389,7 @@ static int do_copy_pool(Rados& rados, const char *src_pool, const char *target_p
 }
 
 static int do_put(IoCtx& io_ctx, RadosStriper& striper,
-		  const char *objname, const char *infile, int op_size,
+		  const char *objname, const char *infile, int op_size, uint64_t obj_offset,
 		  bool use_striper)
 {
   string oid(objname);
@@ -408,7 +408,7 @@ static int do_put(IoCtx& io_ctx, RadosStriper& striper,
   }
   char *buf = new char[op_size];
   int count = op_size;
-  uint64_t offset = 0;
+  uint64_t offset = obj_offset;
   while (count != 0) {
     count = read(fd, buf, op_size);
     if (count < 0) {
@@ -417,7 +417,7 @@ static int do_put(IoCtx& io_ctx, RadosStriper& striper,
       goto out;
     }
     if (count == 0) {
-      if (!offset) { // in case we have to create an empty object
+      if (offset == obj_offset) { // in case we have to create an empty object & if obj_offset > 0 do a hole
 	if (use_striper) {
 	  ret = striper.write_full(oid, indata); // indata is empty
 	} else {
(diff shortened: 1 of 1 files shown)


yangdongsheng:
Hi, @majianpeng , one question here, why we can't add a --obj-offset in rados put command to achieve what we want? In addition, even we have to introduce a rados write here, maybe it's better to do a refactor on do_put to make it support offset parameter. Example:
```
static int _do_write(IoCtx& io_ctx, RadosStriper& striper,
          const char *objname, const char *infile, uint64_t obj_offset, int op_size,
          bool use_striper)
``` 

And then, 
```
do_put() {
    return _do_write(..., obj_offset=0, ...)
}
```

```
do_write(){
    return _do_write(..., obj_offset=offset, ...)
}
```

What do you think? But first of all, why we have to introduce a new subcommand of rados write? 

Thanx :)

majianpeng:
At first, i did it like you said. But i think put don't mean write for object op. So..

majianpeng:
Why add 'rados write'?  In development, i offset need do rado_write_api to test. But no existing tools. I only modify src/test/librados/c_write_operations to do test. Add this to do test for me is very helpful. So i think this maybe help others.

yangdongsheng:
Hi @majianpeng , Let's focus on the first point, whether we have to introduce a new subcommand of 'rados write' rather than reuse 'rados put'. IIUC, you don't think 'put' is a good choice to write data into a object, right? 

Hmm,,,please consider this scenario, we have already put an object of obj_test into pool test. then we do another rados put to overwrite it.
```
(1). rados -p test put obj_test /tmp/data_1
(2). rados -p test put obj_test /tmp/data_2
```
Then I think the step (2) is actually a writing, right? So I think at least currently the rados put is acting like a writer.

But the problem you meet here is there is no functionality to write an object with a specified offset. I think just introducing an option of --obj-offset in `rados put` is a better idea than introducing a `rados write`, which introduces a redundancy and makes a lot of confusion I am afraid.

Thanx :)

tchaikov:
yeah, i am with @yangdongsheng that it makes more sense to simply add "--offset" to "put".

majianpeng:
updated! Thanks very much.

src/tools/rados/rados.cc
```diff
@@ -82,7 +82,7 @@ void usage(ostream& out)
 "\n"
 "OBJECT COMMANDS\n"
 "   get <obj-name> [outfile]         fetch object\n"
-"   put <obj-name> [infile]          write object\n"
+"   put <obj-name> [infile]  [--obj-offset offset]  write object at offset\n"
```
tchaikov: could you use "--offset" instead? as "--obj-offset" is not a global option. i guess one won't be take it as the offset of something else. what do you think?

src/tools/rados/rados.cc
```diff
@@ -431,15 +431,15 @@ static int do_put(IoCtx& io_ctx, RadosStriper& striper,
     }
     indata.append(buffer::ptr(buffer::create_static(count, buf)));
     if (use_striper) {
-      if (offset == 0)
+      if (obj_offset == 0)
```
tchaikov: if the obj_offset is not 0, we should be able to punch a hole using
```
$ rados put obj-name path/to/a/zero-sized-file 128
```

but we just stop if the `count` is 0 with your change, and a zero-sized object is created instead.

src/tools/rados/rados.cc
```diff
@@ -431,15 +431,15 @@ static int do_put(IoCtx& io_ctx, RadosStriper& striper,
...
 	ret = striper.write_full(oid, indata);
       else
-	ret = striper.write(oid, indata, count, offset);
+	ret = striper.write(oid, indata, count, obj_offset);
```
tchaikov: can we use `offset` instead, and initialize if using `obj_offset` at https://github.com/ceph/ceph/pull/8269/files#diff-c24acfa82f15c38ae9370cab8630a579R411 ?

tchaikov:
@majianpeng could you add a simple test in qa/workunits/rados/test_rados_tool.sh? then it's good to qa run and merge. sorry for the latency. this PR just fell off the crack on my table.

Liuchang0812:
It's useful. Should we reopen this PR currently

tchaikov:
@Liuchang0812 this PR is not closed. we just need a test for it.

tchaikov:
superseded by #12674

PR #8270, opened 2016-03-23
os/bluestore/bluestore: For truncate, no need zero(old_size, ROUND_UP_TO(old_size, block_size))

majianpeng:
…_TO(old_size, block_size))

In _do_write, we can make sure (old_size, ROUND_UP_TO(old_size,block_size)) must padded by zero.
So when truncate, we can skip this area. This can reduce one RMW op.


liewegas:
I changed this around so tha thte trailing partial block is no longer zeroed.

PR #8271, opened 2016-03-23
doc: rgw_region_root_pool option should be in [global]

vumrao:
doc: rgw_region_root_pool option should be in [global]
Fixes: #15244

### DIFF
--- a/doc/radosgw/config-ref.rst
+++ b/doc/radosgw/config-ref.rst
@@ -709,9 +709,17 @@ configuration.
 ``rgw region root pool``
 
 :Description: The pool for storing all region-specific information.
+              Not used in Ceph version ``Jewel``.
 :Type: String
 :Default: ``.rgw.root``
 
+.. versionadded:: Jewel
+
+``rgw zonegroup root pool``
+
+:Description: The pool for storing all zonegroup-specific information.
+:Type: String
+:Default: ``.rgw.root``
 
 
 .. versionadded:: v.67
--- a/doc/radosgw/federated-config.rst
+++ b/doc/radosgw/federated-config.rst
@@ -341,11 +341,12 @@ Add Instances to Ceph Config File
 On an admin node, add an entry for each instance in the Ceph configuration file
 for your Ceph Storage Cluster(s). For example:: 
 
-	...
+	[global]
+	rgw region root pool = .us.rgw.root     # Deprecated in Jewel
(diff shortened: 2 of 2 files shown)


theanalyst:
I think this should only make it in infernalis and older.. iirc in jewel, `rgw_region_root_pool` should become `rgw_zonegroup_root_pool`, also can you check whether this behaviour is the same in the current master as well?

theanalyst:
@vumrao also see commit https://github.com/ceph/ceph/commit/796e4703095a79519adc65fcd4169c0d2eb3ba64

vumrao:
@theanalyst Thanks for your inputs. 

- Yes  in Jewel  regions would be *zonegroup*.

- I think then it would be better to send this change to hammer and infernalis and modify this one for rgw_zonegroup_root_pool.

- radosgw-admin zonegroups list
read_default_id : -2
{
    "default_info": "",
    "zonegroups": [
        "default"
    ]
}

- I checked in latest Jewel and it is same it should be in global.

$ ceph osd pool rename .rgw.root .IN.rgw.root
pool '.rgw.root' renamed to '.IN.rgw.root'

$ radosgw-admin zonegroups list
failed to list zonegroups: (2) No such file or directory

- Added in global section of ceph.conf 
$ cat /etc/ceph/ceph.conf 
rgw_zonegroup_root_pool = .IN.rgw.root

$ radosgw-admin zonegroups list
read_default_id : -2
{
    "default_info": "",
    "zonegroups": [
        "default"
    ]
}

- Works perfectly fine.
- But if I am moving it to [client.rgw.<hostname>] section it is failing.
- Only name has changed but the location is same [global] section.
- I will rebase with * new* name : rgw_zonegroup_root_pool for master and will send new patches for hammer and infernalis.

vumrao:
I think it would be good to make it like this - *New in version - Jewel* as given in below link for *rgw region root pool* option in version 0.67:
http://docs.ceph.com/docs/master/radosgw/config-ref/#pools

theanalyst:
> I think it would be good to make it like this - New in version - Jewel as given in below link for rgw region root pool option in version 0.67:
http://docs.ceph.com/docs/master/radosgw/config-ref/#pools

Yes, makes sense

vumrao:
@theanalyst thanks

theanalyst:
LGTM

vumrao:
@yehudasa thanks

PR #8275, opened 2016-03-23
osd: fix dirtying info without correctly setting drity_info field

xiexingguo:
Also add an option to reduce resetting tp handle a little, which shall save CPU a little bit.


liewegas:
LGTM

PR #8276, opened 2016-03-23
python: avoid long paths (part 2)

dachary:


dachary:
this is tested to work with

    CEPH_BUILD_VIRTUALENV=/tmp/ make TESTS=ceph-detect-init/run-tox.sh check

note the trailing / in CEPH_BUILD_VIRTUALENV

src/ceph-detect-init/Makefile.am
```diff
@@ -53,7 +53,7 @@ EXTRA_DIST += \
 	ceph-detect-init/tests/test_all.py \
 	ceph-detect-init/tox.ini
 
-export CEPH_DETECT_INIT_VIRTUALENV = ${CEPH_BUILD_VIRTUALENV}/ceph-detect-init-virtualenv
+export CEPH_DETECT_INIT_VIRTUALENV = ${CEPH_BUILD_VIRTUALENV}ceph-detect-init-virtualenv
```
alfredodeza: does this mean that the `CEPH_BUILD_VIRTUALENV` env var needs to end in a trailing slash to work properly?

dachary: yes

PR #8279, opened 2016-03-23
build/ops: in jessie virtualenv is in package virtualenv

dachary:
The python scripts are not yet compatible with python3, yet it is the
default on jessie. Force the creation of the virtualenv to use python2.7
instead. The wheelhouse is already explicitly populated for both python3
and python2.7 by install-deps.sh, regardless of the default interpreter.


dachary:
passes ./run-make-check.sh with
<pre>
test/docker-test.sh --os-type debian --os-version jessie --shell
</pre>

PR #8281, opened 2016-03-23
doc: rgw adding a very basic multisite doc

theanalyst:
Very basic documentation for the new multisite features, based on emails in the ceph-devel list
TODO
- Hyperlinks within and updating index
- Configuration and installation guide

### DIFF
--- a/doc/radosgw/multisite.rst
+++ b/doc/radosgw/multisite.rst
@@ -1,0 +1,715 @@
+=============
+RGW Multisite
+=============
+
+.. versionadded:: Jewel
+
+From Ceph release Jewel and beyond, you may configure each :term`Ceph Object
+Gateway` to work in an active active zone configuration, allowing for writes to
+non-master zones. Following are the basic terminologies that would be used:
+
+- **Zone**: A zone is *logical* grouping of one or more Ceph Object Gateway
+  instances. There will be one Zone that should be designated as the master zone
+  in a zonegroup, which will handle all bucket and user creation.
+
+- **Zonegroup**: A zonegroup consists of multiple zones, this approximately
+  corresponds to what used to be called as a region in pre Jewel releases for
+  federated deployments. There should be a master zonegroup that will handle
+  changes to the system configuration.
+
+- **Zonegroup map**: A zonegroup map is a configuration structure that holds the
+  map of the entire system, ie. which zonegroup is the master, relationships
+  between different zonegroups and certain configurables like storage policies.
+
+- **Realm**: A realm is a container for zonegroups, this allows for separation
+  of zonegroups themselves between clusters. It is possible to create multiple
+  realms, making it easier to run completely different configurations in the same
+  cluster.
+
+- **Period**: A period holds the configuration structure for the current state
+  of the realm. Every period contains a unique id and an epoch. A period's epoch
+  is incremented on every commit operation. Every realm has an associated
+  current period, holding the current state of configuration of the zonegroups
+  and storage policies. Any configuration change for a non master zone will
+  increment the period's epoch. Changing the master zone to a different zone
+  will trigger the following changes:
+  - A new period is generated with a new period id and epoch of 1
+  - Realm's current period is updated to point to the newly generated period id
+  - Realm's epoch is incremented
+
+Pool Configuration
+==================
+
+About this guide
+================
+
+In this guide we create a single zone group, with three separate zones, which
+actively sync data between them. For the purpose of the guide, we create 2 zones
+in the same cluster, and the third zone in a different cluster. There is no sync
+agent involved for mirroring data changes between the RadosGWs and this allows
+for a much simpler configuration scheme and active-active configurations. Please
+note that metadata operations such as creating a user would still need to go
+through the master zone, however data operations such as creation of buckets
+objects etc. can be handled by any of the zones.
+
+System Keys
+===========
+
+While configuring zones, RadosGW, expects creation of zone a system user, with
+an S3-like access and secret keys. This allows another radosgw instance to pull
+the configuration remotely, with the access and secret keys. For making
+scripting and use of configuration management tools easier, it would be easier
+to generate the access-key and secret-key before hand and use these. For the
+purpose of the guide we assume an access key and secret key are set in the
+environment. For eg., something like::
+
+  $ SYSTEM_ACCESS_KEY=1555b35654ad1656d805
+  $ SYSTEM_SECRET_KEY=h7GhxuBLTrlhVUyxSPUKUV8r/2EI4ngqJxD7iBdBYLhwluN30JaT3Q12
+
+Generally access keys are 20 alphanumeric characters, secret keys are
+40 alphanumeric characters (they can contain +/= chars as well). These
+keys can be generated in shell for eg::
+
+  $ SYSTEM_ACCESS_KEY=$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 20 | head -n 1)
+  $ SYSTEM_SECRET_KEY=$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 40 | head -n 1)
+
+Naming conventions
+==================
+
+This section describes the process of setting up a master zone. For the purpose
+of the guide, we'll assume a zonegroup called ``us`` spanning the United States,
+which will be our master zonegroup. This will contain two zones written in a
+``{zonegroup}-{zone}`` format, this is just a convention only and you are free
+to choose a format you prefer to set these names. So in summary:
+
+- Master Zonegroup: United States ``us``
+- Master zone: United States, East Region 1: ``us-east-1``
+- Secondary zone: United States, East Region 2: ``us-east-2``
+- Secondary zone: United States, West Region: ``us-west``
+
+This will be a part of a larger realm, say ``gold``. The zones ``us-east-1`` and
+``us-east-2`` are part of the same ceph cluster, of which we'll make
+``us-east-1`` as the primary. ``us-west`` will be in a different ceph cluster.
+
+Creating a realm
+----------------
+
+Configure a realm called ``gold``, and also make this the default ::
+
+  # radosgw-admin realm create --rgw-realm=gold --default
+  {
+    "id": "4a367026-bd8f-40ee-b486-8212482ddcd7",
+    "name": "gold",
+    "current_period": "09559832-67a4-4101-8b3f-10dfcd6b2707",
+    "epoch": 1
+  }
+
+
+Note that every realm has an id, which allows for flexibility like renaming the
+realm later, should the need arise. The ``current_period`` changes whenever we
+change anything in the master zone. The ``epoch`` is incremented when there's a
+change in configuration of the master zone, which results in a change of current
+period.
+
+Deleting the default zonegroup
+------------------------------
+
+A simple installation of radosgw would assume the default zonegroup called
+"default", since we no longer need this we begin by removing the default
+zonegroup::
+
+  # radosgw-admin zonegroup delete --rgw-zonegroup=default
+
+
+Creating a master zonegroup
+---------------------------
+
+We'll be creating a zonegroup called ``us`` as a master zonegroup. A master
+zonegroup will be in control of the zonegroup map and propagate changes to the
+rest of the system. We will also set this zonegroup as the default, which allows later commands to use this zonegroup without explicitly mentioning it with the rgw-zonegroup switch.
+
+::
+
+   # radosgw-admin zonegroup create --rgw-zonegroup=us --endpoints=http://rgw1:80 --master --default
+   {
+    "id": "d4018b8d-8c0d-4072-8919-608726fa369e",
+    "name": "us",
+    "api_name": "us",
+    "is_master": "true",
+    "endpoints": [
+        "http:\/\/rgw1:80"
+    ],
+    "hostnames": [],
+    "hostnames_s3website": [],
+    "master_zone": "",
+    "zones": [],
+    "placement_targets": [],
+    "default_placement": "",
+    "realm_id": "4a367026-bd8f-40ee-b486-8212482ddcd7"
+    }
+
+Alternatively, we make this zonegroup as the default zonegroup via the following
+command ::
+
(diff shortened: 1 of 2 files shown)


doc/radosgw/multisite.rst
```diff
@@ -0,0 +1,487 @@
...
+    "name": "gold",
+    "current_period": "09559832-67a4-4101-8b3f-10dfcd6b2707",
+    "epoch": 1
+  }
+
+
+Note that every realm has an id, which allows for flexibility like renaming the
+realm later, should the need arise. The ``current_period`` changes whenever we
+change anything in the master zone. The ``epoch`` is incremented when there's a
+change in configuration which doesn't modify the master zone
```
cbodley: > The ``current_period`` changes whenever we change anything in the master zone. The ``epoch`` is incremented when there's a change in configuration which doesn't modify the master zone

this is a bit off, from my understanding. any change to the configuration that doesn't change which zone is the master_zone will just increment the _period's_ epoch. on the other hand, changing the master_zone to point to a different zone will a) generate a new period id with period epoch=1, b) assign realm's current_period to the new period's id, and c) increment the realm's epoch. be careful not to confuse the realm's epoch with the period's epoch

doc/radosgw/multisite.rst
```diff
@@ -0,0 +1,487 @@
...
+  federated deployments. There should be a master zonegroup that will handle
+  changes to the system configuration.
+
+- **Zonegroup map**: A zonegroup map is a configuration structure that holds the
+  map of the entire system, ie. which zonegroup is the master, relationships
+  between different zonegroups and certain configurables like storage policies.
+
+- **Realm**: A realm is a container for zonegroups, this allows for separation
+  of zonegroups themselves between clusters. It is possible to create multiple
+  realm, making it easier to run completely different configurations in the same
```
yehudasa: multiple realms

doc/radosgw/multisite.rst
theanalyst: Thanks for clarifying, I'll update

doc/radosgw/multisite.rst
```diff
@@ -0,0 +1,487 @@
...
+- Secondary zone: United States, West Region: ``us-west``
+
+This will be a part of a larger realm, say ``gold``
+
+Creating a realm
+----------------
+
+Configure a realm called ``gold``, which for ::
+
+  # radosgw-admin realm create --rgw-realm=gold
```
oritwas: I would add the --default flag to set the realm as default so you don't need to provide in all the admin commands

doc/radosgw/multisite.rst
```diff
@@ -0,0 +1,487 @@
...
+---------------------------
+
+We'll be creating a zonegroup called ``us`` as a master zonegroup. A master
+zonegroup will be in control of the zonegroup map and propagate changes to the
+rest of the system. We will also set this zonegroup as the default, which allows
+explicitly mentioning the ``rgw-zonegroup`` switch for later commands.
+
+::
+
+   # radosgw-admin zonegroup create --rgw-zonegroup=us --endpoints=http://rgw1:80 --master
```
oritwas: also add --default

doc/radosgw/multisite.rst
```diff
@@ -0,0 +1,487 @@
...
+    "hostnames": [],
+    "hostnames_s3website": [],
+    "master_zone": "",
+    "zones": [],
+    "placement_targets": [],
+    "default_placement": "",
+    "realm_id": "4a367026-bd8f-40ee-b486-8212482ddcd7"
+    }
+
+Now we make this as the default zonegroup via the following command ::
```
oritwas: you can add the flag --default when creating the zonegroup

doc/radosgw/multisite.rst
```diff
@@ -0,0 +1,487 @@
...
+    "realm_id": "4a367026-bd8f-40ee-b486-8212482ddcd7"
+    }
+
+   # radosgw-admin zone default --rgw-zone=us-east
+
+Next we add the zone to the zonegroup::
+
+  # radosgw-admin zonegroup add --rgw-zonegroup=us --rgw-zone=us-east
+
+
```
oritwas: We allow adding the zone to the zonegroup in the create command by adding --rgw-zonegroup=<zg> and setting it as default by adding --default.

doc/radosgw/multisite.rst
```diff
@@ -0,0 +1,487 @@
...
+  # radosgw-admin realm default --rgw-realm=gold
+
+Secondary Zone Configuration
+----------------------------
+
+We first set the default zonegroup to the created ``us`` zonegroup::
+
+  # radosgw-admin zonegroup default --rgw-zonegroup=us
+
+Next we create the new zone, ``us-west``, with the same system keys::
```
oritwas: Just to make it simpler, I would have created the secondary zone on the master zone and than would do just one period update --commit.

doc/radosgw/multisite.rst
```diff
@@ -0,0 +1,487 @@
...
+  --access-key=$SYSTEM_ACCESS_KEY --secret-key=$SYSTEM_SECRET_KEY
+  {
+    "id": "4a367026-bd8f-40ee-b486-8212482ddcd7",
+    "name": "gold",
+    "current_period": "b5e4d3ec-2a62-4746-b479-4b2bc14b27d1",
+    "epoch": 2
+  }
+
+  # radosgw-admin realm default --rgw-realm=gold
+
```
oritwas: you are missing period pull

doc/radosgw/multisite.rst
theanalyst: Sure makes sense, I'll change that

doc/radosgw/multisite.rst
theanalyst: ack

doc/radosgw/multisite.rst
theanalyst: Hm ok, I didn't know of this approach, sounds simpler, I'll update teh docs

theanalyst: Not able to get this working completely yet, the secondary ceph cluster does see the realm, period and zonegroups, however a zone list still turns up empty

theanalyst:
thanks for the reviews, pushed some commits addressing the comments, (can be squashed later if necessary). 

@oritwas In the last commit cfee62a0e79218729fc6d0f5135a9f8b5366975a I tried updating the doc to create the secondary zone from the first cluster and pull the period, realm etc. see if this makes sense?

doc/radosgw/multisite.rst
cbodley: @oritwas I looked into making this work (creating a zone on the primary cluster and pulling it to the secondary), but I don't think it's actually possible with our design. 'zone create' will create a local `RGWZoneParams` object that contains all of the pool names, access key, etc. but the period only stores them as `RGWZone` - which is an external representation that tells other zones how to connect to/use it. this means that a 'period pull' can't get the `RGWZoneParams` from one cluster to another

so zones need to be created on their own cluster, for now. we might consider adding a 'zone pull' command, but I'm not sure it's a good idea..

oritwas: @cbodley , the problem is that the zone id will be different. Maybe store the RGWZoneParams in the period or create the zone when doing period pull?

oritwas: @theanalyst, after discussing it with @cbodley:
If we have two different clusters:
on the master: realm creation, zonegroup creation, master zone creation, period update --commit
on the second zone: realm pull, period pull, create second zone, period update --comitt
For one cluster we can have a simpler way of configuration as everything can be done on the master gateway including the second zone creation (no need for realm pull and period pull).

theanalyst: @oritwas @cbodley 
> If we have two different clusters:
on the master: realm creation, zonegroup creation, master zone creation, period update --commit
on the second zone: realm pull, period pull, create second zone, period update --comitt
For one cluster we can have a simpler way of configuration as everything can be done on the master gateway including the second zone creation (no need for realm pull and period pull). 

Alright, so if its a single cluster the second approach works,. sounds useful.. and for 2 different clusters the std. approach ... thanks! I'll try this locally once and update the docs

theanalyst:
@oritwas @cbodley I've updated the doc with your comments, changed the doc a bit to show both the possibilities, ie. zone1 and zone2 in same cluster, zone2 in different clusters. Do take a look

doc/radosgw/multisite.rst
```diff
@@ -0,0 +1,710 @@
...
+While configuring zones, RadosGW, expects creation of zone a system user, with
+an S3 like access and secret keys. This allows another radosgw instance to pull
+the configuration remotely, with the access and secret keys. For making
+scripting and use of configuration management tools easier, it would be easier
+to generate the access-key and secret-key before hand and use these. For the
+purpose of the guide we assume an access key and secret key are set in the
+environment. For eg., something like::
+
+  $ SYSTEM_ACCESS_KEY=1555b35654ad1656d805
+  $ SYSTEM_SECRET_KEY=h7GhxuBLTrlhVUyxSPUKUV8r/2EI4ngqJxD7iBdBYLhwluN30JaT3Q==
```
jrosenboom: Can you add an example how to create these keys? Would be bad if people just use the sample from the docs because it is too complicated to find out how to build your own.

doc/radosgw/multisite.rst
```diff
@@ -0,0 +1,710 @@
...
+Gateway` to work in an active active zone configuration, allowing for writes to
+non-master zones. Following are the basic terminologies that would be used:
+
+- **Zone**: A zone is *logical* grouping of one or more Ceph Object Gateway
+  instances. There will be one Zone that should be designated as the master zone
+  in a zonegroup, which will handle all bucket and user creation.
+
+- **Zonegroup**: A zonegroup consists of multiple zones, this approximately
+  corresponds to what used to be called as a region in pre Jewel releases for
+  federated deployments. There should be a master zonegroup that will handle
```
jrosenboom: Would this section just replace the section about federated deployments at some point in time? Otherwise that should at least get a warning that it doesn't work anymore with Jewel, it fails when there is no realm defined.

doc/radosgw/multisite.rst
```diff
@@ -0,0 +1,710 @@
...
+Creating a master zone
+----------------------
+
+Next we create a zone, and make it as the default zone. Note that for metadata
+operations like user creation you would want to use this zone. We also add it to
+the zonegroup
+
+::
+
+   # radosgw-admin zone create --rgw-zonegroup=us --rgw-zone=us-east-1 --endpoints=http://rgw1:80 --access-key=$ZONE_ACCESS_KEY --secret=$ZONE_SECRET_KEY --default --master
```
jrosenboom: You didn't define `$ZONE_ACCESS_KEY` earlier, should this be `$SYSTEM_ACCESS_KEY` instead?

doc/radosgw/multisite.rst
theanalyst: As far as ceph is concerned, I believe we may not actually have an alphanumeric restriction,  but for the sake of uniformity, we can probably generate keys by getting 20 random alphanumeric chars for access, and 40 for secret

`cat /dev/urandom | tr -dc 'a-zA-Z0-9 | fold -w 20  | head -n1' and 40 for secret is what comes to my mind.. my bash-fu isn't that great and any alternatives are welcome :-)

doc/radosgw/multisite.rst
theanalyst: Ah thanks, yeah I'll fix this

doc/radosgw/multisite.rst
```diff
@@ -0,0 +1,710 @@
...
+for a much simpler configuration scheme and active-active configurations. Please
+note that metadata operations such as creating a user would still need to go
+through the master zone, however data operations such as creation of buckets
+objects etc. can be handled by any of the zones.
+
+System Keys
+===========
+
+While configuring zones, RadosGW, expects creation of zone a system user, with
+an S3 like access and secret keys. This allows another radosgw instance to pull
```
cbodley: > While configuring zones, RadosGW, expects creation of zone a system user, with an S3 like access and secret keys.

While configuring zones, RadosGW expects creation of a system user, with S3-like access and secret keys.

theanalyst: Thanks, I'll get this

doc/radosgw/multisite.rst
```diff
@@ -0,0 +1,710 @@
...
+
+
+Creating a master zonegroup
+---------------------------
+
+We'll be creating a zonegroup called ``us`` as a master zonegroup. A master
+zonegroup will be in control of the zonegroup map and propagate changes to the
+rest of the system. We will also set this zonegroup as the default, which allows
+explicitly mentioning the ``rgw-zonegroup`` switch for later commands. We also
+make this the default zonegroup
```
cbodley: > We will also set this zonegroup as the default, which allows explicitly mentioning the ``rgw-zonegroup`` switch for later commands. We also make this the default zonegroup

We will also set this zonegroup as the default, which allows later commands to use this zonegroup without explicitly mentioning it with the ``rgw-zonegroup`` switch.

doc/radosgw/multisite.rst
```diff
@@ -0,0 +1,710 @@
...
+  # radosgw-admin realm pull --url=http://rgw1:80
+  --access-key=$SYSTEM_ACCESS_KEY --secret=$SYSTEM_SECRET_KEY
+  {
+    "id": "4a367026-bd8f-40ee-b486-8212482ddcd7",
+    "name": "gold",
+    "current_period": "b5e4d3ec-2a62-4746-b479-4b2bc14b27d1",
+    "epoch": 2
+  }
+
+  # radosgw-admin period pull --url=http://rgw1:80 --acess-key=$SYSTEM_ACCESS_KEY --secret=$SYSTEM_SECRET_KEY
```
cbodley: > --acess-key=

--access-key=

cbodley:
nits aside, the content looks great. i like how you added a third zone to show the difference between configuration on the same cluster vs. a separate cluster

theanalyst:
@cbodley thanks, addressed the comments, squashed the commits, learnt a lot by creating the doc as well :)

PR #8284, opened 2016-03-23
rgw: fix lockdep false positive

yehudasa:

### DIFF
--- a/src/rgw/rgw_data_sync.cc
+++ b/src/rgw/rgw_data_sync.cc
@@ -1239,17 +1239,16 @@ public:
 
       yield {
         if  ((rgw_data_sync_info::SyncState)sync_status.sync_info.state == rgw_data_sync_info::StateSync) {
-          case rgw_data_sync_info::StateSync:
-            for (map<uint32_t, rgw_data_sync_marker>::iterator iter = sync_status.sync_markers.begin();
-                 iter != sync_status.sync_markers.end(); ++iter) {
-              RGWDataSyncShardControlCR *cr = new RGWDataSyncShardControlCR(sync_env, sync_env->store->get_zone_params().log_pool,
-                                                        iter->first, iter->second);
-              cr->get();
-              shard_crs_lock.Lock();
-              shard_crs[iter->first] = cr;
-              shard_crs_lock.Unlock();
-              spawn(cr, true);
-            }
+          for (map<uint32_t, rgw_data_sync_marker>::iterator iter = sync_status.sync_markers.begin();
+               iter != sync_status.sync_markers.end(); ++iter) {
+            RGWDataSyncShardControlCR *cr = new RGWDataSyncShardControlCR(sync_env, sync_env->store->get_zone_params().log_pool,
+                                                                          iter->first, iter->second);
+            cr->get();
+            shard_crs_lock.Lock();
+            shard_crs[iter->first] = cr;
+            shard_crs_lock.Unlock();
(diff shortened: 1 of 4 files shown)


src/rgw/rgw_data_sync.cc
```diff
@@ -1239,17 +1242,16 @@ class RGWDataSyncCR : public RGWCoroutine {
...
-                 iter != sync_status.sync_markers.end(); ++iter) {
-              RGWDataSyncShardControlCR *cr = new RGWDataSyncShardControlCR(sync_env, sync_env->store->get_zone_params().log_pool,
-                                                        iter->first, iter->second);
-              cr->get();
-              shard_crs_lock.Lock();
-              shard_crs[iter->first] = cr;
-              shard_crs_lock.Unlock();
-              spawn(cr, true);
-            }
+          for (map<uint32_t, rgw_data_sync_marker>::iterator iter = sync_status.sync_markers.begin();
```
theanalyst: nit: cant we use auto here?

src/rgw/rgw_data_sync.cc
```diff
@@ -1239,17 +1242,16 @@ class RGWDataSyncCR : public RGWCoroutine {
 
       yield {
         if  ((rgw_data_sync_info::SyncState)sync_status.sync_info.state == rgw_data_sync_info::StateSync) {
-          case rgw_data_sync_info::StateSync:
```
oritwas: why was this removed?

src/rgw/rgw_data_sync.cc
```diff
@@ -1182,9 +1182,12 @@ class RGWDataSyncCR : public RGWCoroutine {
   }
 
   ~RGWDataSyncCR() {
+    shard_crs_lock.Lock();
```
cbodley: shouldn't need to lock in a dtor - if we get here while operate() is still running, that would indicate a separate ref counting issue. did you see this happening?

src/rgw/rgw_data_sync.cc
yehudasa: @theanalyst yes we can, but this is just indentation fix (and removed extra line), so I wouldn't go into changing it just now

src/rgw/rgw_data_sync.cc
yehudasa: @oritwas this line shouldn't have been there.. there's no switch statement, it only compiles because the reenter/yield expand into a switch statement. The `if` there replaced the original switch.

src/rgw/rgw_data_sync.cc
yehudasa: @cbodley you have a point, I've seen shard_crs inconsistency, but can't be sure what the culprit was.

src/rgw/rgw_data_sync.cc
cbodley: it looks like this used to be a switch statement that changed to an if on the line above. i assume the compiler didn't complain about the leftover `case` here because the boost `reenter` macro is using a switch under the covers

src/rgw/rgw_data_sync.cc
theanalyst: Ah ok, no issues

src/rgw/rgw_data_sync.cc
yehudasa: @cbodley removed this

yehudasa:
@cbodley added a couple more fixes, and made the change according to your comment

PR #8260, opened 2016-03-22
test/test-erasure-code.sh: disable pg_temp priming

liewegas:
fixes make check

PR #8285, opened 2016-03-23
debian/control: try installing virtualenv first, if it exists

dmick:
Debian Jessie (at least) now has the virtualenv command in
package 'virtualenv', which depends on python3-virtualenv, and
that's how you get the virtualenv command there.  Earlier
Debian-based releases still use python-virtualenv.  Jessie's
virtualenv command still creates a python2 venv by default.

This will likely require a followup fix for install-deps.sh
to handle the | correctly; that's the next investigation


dachary:
the order in which the packages are mentioned here do not matter. If not installing recommends and python-virtualenv is somehow found to be in the dependency chain before virtualenv, we may end up having python-virtualenv only installed on jessie and no virtualenv. And the virtualenv | python-virtualenv is satisfied.

dachary:
<pre>
</pre>

pbuilder build --binary-arch --distribution jessie --basetgz /opt/jessie.tgz --buildresult /opt --debbuildopts -j12 ../ceph_10.0.5.dsc

with this pull request installs virtualenv. The comment should be updated to mention the fact that the problem is specific to pbuilder (because the default policy of ubuntu / jessie is to isntall the recommends).

dmick:
Updated the commit message to contain more of what I know now

dmick:
Check failure's not my fault, man

dmick:
Probably want https://github.com/ceph/ceph/pull/8286 at the same time

dachary:
the make check failure is unrelated and has been fixed by https://github.com/ceph/ceph/pull/8260

PR #8288, opened 2016-03-24
Makefile-env.am: set a default for CEPH_BUILD_VIRTUALENV

dmick:
We need to shorten the virtualenv path for non-Jenkins builds
too; gitbuilders were also overflowing the interpreter path limit


dmick:
testing local build to be sure this is right

dmick:
Manual test looks fine; virtualenvs created in /tmp as expected.

dmick:
Since @jdurgin asked, I verified that "make clean" removes the dir, and in case make clean isn't run, setup-virtualenv.sh removes and rebuilds it

PR #8290, opened 2016-03-24
Fix unittest for chain_xattr.

yangdongsheng:
When we run unittest_chain_xattr, if there is a default xattr enabled, such as security.selinux, this test would fail.
```
  attr security.selinux
  attr user.foo
  attr user.foo@1
[       OK ] chain_xattr.chunk_aligned (2 ms)
[ RUN      ] chain_xattr.listxattr
test/objectstore/chain_xattr.cc:251: Failure
Value of: ::memcmp(expected, actual, buffer_size)
  Actual: 117
Expected: 0
[  FAILED  ] chain_xattr.listxattr (1 ms)
[----------] 3 tests from chain_xattr (333 ms total)

[==========] 3 tests from 1 test case ran. (333 ms total)
[  PASSED  ] 2 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] chain_xattr.listxattr

 1 FAILED TEST
FAIL unittest_chain_xattr (exit status: 1)
```

This patch record the original xattr in the expected buffer, then unittest_chain_xattr works well now.

### DIFF
--- a/src/test/objectstore/chain_xattr.cc
+++ b/src/test/objectstore/chain_xattr.cc
@@ -220,6 +220,35 @@ TEST(chain_xattr, chunk_aligned) {
   ::unlink(file);
 }
 
+void get_vector_from_xattr(vector<string> &xattrs, char* xattr, int size) {
+  char *end = xattr + size;
+  while (xattr < end) {
+    if (*xattr == '\0' )
+      break;
+    xattrs.push_back(xattr);
+    xattr += strlen(xattr) + 1;
+  }
+}
+
+bool listxattr_cmp(char* xattr1, char* xattr2, int size) {
+  vector<string> xattrs1;
+  vector<string> xattrs2;
+  get_vector_from_xattr(xattrs1, xattr1, size);
+  get_vector_from_xattr(xattrs2, xattr2, size);
+
+  if (xattrs1.size() != xattrs2.size())
+    return false;
+
+  std::sort(xattrs1.begin(), xattrs1.end());
+  std::sort(xattrs2.begin(), xattrs2.end());
+  std::vector<string> diff;
+  std::set_difference(xattrs1.begin(), xattrs1.end(),
+			  xattrs2.begin(), xattrs2.end(),
+			  diff.begin());
+
+  return diff.empty();
+}
+
 TEST(chain_xattr, listxattr) {
   const char* file = FILENAME;
   ::unlink(file);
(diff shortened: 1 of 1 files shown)


src/test/objectstore/chain_xattr.cc
```diff
@@ -230,13 +230,32 @@ TEST(chain_xattr, listxattr) {
   const string x(LARGE_BLOCK_LEN, 'X');
   const int y = 1234;
 
+  int orig_size = chain_listxattr(file, NULL, 0);
+  char *orig_buffer = NULL;
+  string orig_str;
```
tchaikov: no need to have an `orig_str`. `memcpy(3)` would suffice.

src/test/objectstore/chain_xattr.cc
```diff
@@ -230,13 +230,32 @@ TEST(chain_xattr, listxattr) {
...
+  if (orig_size) {
+    orig_buffer = (char*)calloc(1, orig_size);
```
tchaikov: `malloc(3)` is good enough in this case.

src/test/objectstore/chain_xattr.cc
tchaikov: nvm, i see you are using `strcpy()` here and the target is malloc'ed.

src/test/objectstore/chain_xattr.cc
yangdongsheng: good point thanx @tchaikov

yangdongsheng:
@tchaikov Updated, please help to review again, thanx :)

tchaikov:
lgtm

wjwithagen:
@yangdongsheng @tchaikov 
There is one major problem with the code in this test, when run on FreeBSD.
in " ASSERT_EQ(0, ::memcmp(expected, actual, buffer_size));" is assumed that the attributes are returned in the same order as they are inserted. At which point the memcmp works.
FreeBSD does not do this.
And in the manual pages both in Linux and in FreeBSD there is no guarantee that such a feature actually is guaranteed. eg. it could be that a different filesystem also has a different return strategy.
But that would require a masivve rewrite of this one line compare. Which I have not (yet) done.

yangdongsheng:
@wjwithagen thanx, good point. My solution about it would be split the expected and actual into name list and then do a sorting. then compare them one by one. But I am really a newbie to C++ :(. I am not sure how to implement it currently, but I am glad to do an investigation. Thanx .

src/test/objectstore/chain_xattr.cc
```diff
@@ -230,13 +230,32 @@ TEST(chain_xattr, listxattr) {
...
 
+  int orig_size = chain_listxattr(file, NULL, 0);
+  char *orig_buffer = NULL;
+  string orig_str;
+  if (orig_size) {
+    orig_buffer = (char*)malloc(orig_size);
+    chain_flistxattr(fd, orig_buffer, orig_size);
+    orig_str = string(orig_buffer);
+    orig_size = orig_str.size();
+  }
```
tchaikov: @yangdongsheng you can use the boost for splitting the concatenated xattrs, like:
```c++
  using strings = std::vector<std::string>;
  std::string s;
  int size = chain_listxattr(file, NULL, 0);
  s.resize(size);
  chain_flistxattr(fd, &s[0], s.size());
  strings attrs;
  boost::split(attrs, s, boost::is_from_range('\0', '\0'));
  std::sort(attrs.begin(), attrs.end());
```

wjwithagen: @tchaikov 
That's using the power of boost. Something I'm not fully into also.
It's like in Java: some many classes to choose from that it is hard to fully understand all of them.

tchaikov: agreed. it's difficult to *fully* understand *all* of them. but if somethings goes wrong, guess we can also try to dive into it to figure out the root cause.

yangdongsheng: @tchaikov Thanx for your tip.

@wjwithagen 'boost' is widely used in ceph, I think that's not a problem to take a use of boost here. 

Anyway, I will update this branch next week, thanx guys. :)

yangdongsheng:
Hi @wjwithagen and @tchaikov it's updated and rebased against master. Please help to take a look.

src/test/objectstore/chain_xattr.cc
```diff
@@ -18,6 +18,7 @@
  * GNU Library Public License for more details.
  *
  */
+#include <boost/algorithm/string.hpp>
```
tchaikov: if you choose to reinvent the `split()`, maybe we can dispense with this line?

yangdongsheng: @tchaikov yes, you are right, This is not necessary any more. :)

yangdongsheng:
@tchaikov Could you help to take a look again?  thanx :)

tchaikov:
lgtm

wjwithagen:
lgtm 
On FreeBSD it now passed without excluding any tests.
Nice work.

yangdongsheng:
@tchaikov It sounds ready to go, mind merging it now? :)

tchaikov:
as it has not passed the rados qa run, and it is not a bug fix targeting jewel. maybe we need to wait until the release of jewel (10.2.1).

src/test/objectstore/chain_xattr.cc
```diff
@@ -230,20 +259,41 @@ TEST(chain_xattr, listxattr) {
...
+    index = orig_size + 1;
+  }
+  ::strcpy(expected + index, name1.c_str());
+  ::strcpy(expected + index + name1.size() + 1, name2.c_str());
+  char* actual = (char*)malloc(buffer_size);
+  ::memset(actual, '\0', buffer_size);
   ASSERT_LT(buffer_size, chain_listxattr(file, NULL, 0)); // size evaluation is conservative
   chain_listxattr(file, actual, buffer_size);
-  ASSERT_EQ(0, ::memcmp(expected, actual, buffer_size));
+  ASSERT_EQ(true, listxattr_cmp(expected, actual, buffer_size));
```
tchaikov: could use ASSERT_TRUE()

src/test/objectstore/chain_xattr.cc
```diff
@@ -230,20 +259,41 @@ TEST(chain_xattr, listxattr) {
...
   ::memset(actual, '\0', buffer_size);
   chain_flistxattr(fd, actual, buffer_size);
-  ASSERT_EQ(0, ::memcmp(expected, actual, buffer_size));
+  ASSERT_EQ(true, listxattr_cmp(expected, actual, buffer_size));
```
tchaikov: ditto.

yangdongsheng:
@tchaikov  thanx for the review. Branch is updated and rebased. please help to review again. :)

wjwithagen:
LGTM
Tested it last night on FreeBSD, and passed

tchaikov:
lgtm

src/test/objectstore/chain_xattr.cc
wjwithagen: LGTM
Still working for me

PR #8291, opened 2016-03-24
osd: more fixes for incorrectly dirtying info; resend reply for duplicated scrub-reserve req

xiexingguo:


xiexingguo:
@liewegas More fixes

liewegas:
How would the scrub message get lost "due to a network problem"?  The osd <-> osd message passing is lossless

liewegas:
The rest looks right to me.  Can we just drop that one commit?

xiexingguo:
Sure, dropped.

PR #8294, opened 2016-03-24
rgw:Use count fn in RGWUserBuckets for quota check

theanalyst:
We already have a count function from RGWUserBuckets class which gives
the total size of the map, use that instead of getting the map and doing
a count.


mattbenjamin:
@oritwas  @yehudasa this looks promising

PR #8052, opened 2016-03-11
hammer: test/librados/tier.cc doesn't completely clean up EC pools

xiaoxichen:
http://tracker.ceph.com/issues/15051

PR #7883, opened 2016-03-02
hammer: osd: use GMT time for the object name of hitsets

tchaikov:
http://tracker.ceph.com/issues/13288

PR #8295, opened 2016-03-24
osd/osd_types: encode pg_pool_t the old way

tchaikov:
if we upgrade from an old cluster with monitor 0.94, which is still
suffered from tracker#12410. then the crc error of re-encoded osdmap
can prevent the osd from requesting more osdmaps. as the monitor
fails to send a single last map due to the off-by-one bug.

a reproducer is https://github.com/ceph/ceph/pull/8052#issuecomment-198750979

Fixes: #13288

### DIFF
--- a/src/osd/osd_types.cc
+++ b/src/osd/osd_types.cc
@@ -1491,6 +1491,59 @@ void pg_pool_t::encode(bufferlist& bl, uint64_t features) const
     return;
   }
 
+  if ((features & CEPH_FEATURE_OSD_HITSET_GMT) == 0) {
+    // CEPH_FEATURE_OSD_HITSET_GMT requires pg_pool_t v21 which has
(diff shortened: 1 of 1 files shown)


tchaikov:
it fixes http://qa-proxy.ceph.com/teuthology/kchai-2016-03-23_22:01:54-upgrade-jewel---basic-multi/82401/teuthology.log

tchaikov:
@liewegas not sure if jewel should have this fix. this only happens if we upgrade from an **old** hammer cluster.

tchaikov:
this is a "forward" port of a commit in #7883

tchaikov:
the jenkins test failed for no reason, strange enough..

liewegas:
The user is required to upgrade to 0.94.4 before moving on to infernalis or jewel.  Does that mean they avoid this problem?

tchaikov:
@liewegas yes!

PR #8296, opened 2016-03-24
build: set hardening flags conditionally

kmroz:
Last night I was stepping (actually hopping :D) through radosgw in gdb due to the optimization level set during compilation. I went digging where to set -O0... do_autogen.sh provides a "-O" option, however this broke compilation because FORTIFY_SOURCE=2 (along with a few others) are being set via HARDENING_CFLAGS in src/Makefile-env.am under a "if LINUX" conditional.

This patch _tries_ to fix this by adding a "--without-hardening" option to do_autogen.sh/configure.ac to be used alongside "-d" and "-O".

First pass at fixing: http://tracker.ceph.com/issues/15263

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -106,6 +106,13 @@ AC_PROG_LIBTOOL
 AC_SUBST(AM_CXXFLAGS)
 AM_CXXFLAGS="${AM_CXXFLAGS}"
 
+# Hardened build
+AC_ARG_WITH([hardening],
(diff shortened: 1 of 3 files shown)


tchaikov:
#8301 is another way to fix this.

kmroz:
@tchaikov Yes it is. Let's get one of them committed so we can build with -O0. Though I suspect this is not much of a tragedy for the upcoming release :)

PR #8297, opened 2016-03-24
cls_rbd: mirror_image_list should return global image id

dillaman:
The global image id is needed to crosslink images when an image is
replicated between multiple clusters.

### DIFF
--- a/src/tools/rbd_mirror/PoolWatcher.cc
+++ b/src/tools/rbd_mirror/PoolWatcher.cc
@@ -105,19 +105,28 @@ void PoolWatcher::refresh_images(bool reschedule)
       continue;
     }
 
-    // only format 2 images can be mirrored, so only check the format
-    // 2 rbd_directory structure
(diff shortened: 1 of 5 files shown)


src/tools/rbd_mirror/PoolWatcher.cc
```diff
@@ -105,19 +105,28 @@ void PoolWatcher::refresh_images(bool reschedule)
...
+             << cpp_strerror(r) << dendl;
+        continue;
+      }
+      for (auto it = mirror_images.begin(); it != mirror_images.end(); ++it) {
+        image_ids.insert(it->first);
+      }
+      if (!mirror_images.empty()) {
+        last_read = mirror_images.rbegin()->first;
+      }
+      r = mirror_images.size();
```
jdurgin: won't this not fetch all the image ids?

jdurgin: nvm, misread

PR #8298, opened 2016-03-24
build: many fixes for freebsd

wjwithagen:
Enough changes are added that the autotools can do their work and generate
working Makefiles that will start compilation. 
But compilation on FreeBSD stops due to incompatibilities

To fix these a new PULL will follow


gregsfortytwo:
Looks like this has passed tests, which is as much review as I'm capable of giving it...but I suspect some of our build guys (or maybe @ukernel, who's done some bsd porting work before?) would like to look at it first.

wjwithagen:
This is the least invasive part, I guess, of all I have.
And it was extracted from the previous, too large, pull I had: #7573.
From what I saw @ukernel was mainly looking at the Darwin end of life.
@tchaikov already had some remarks which I tried to integrate.
and some of the stuff was in commits by @trociny which I tried to use here as well.
But I'm awaiting suggestions and hope the pull to goes in.
So I can do the next set of commits to get compilation fro FreeBSD in order.

README.FreeBSD
```diff
@@ -0,0 +1,56 @@
+
+The FreeBSD build will build most of the tools in Ceph.
+Note that the RBD dependant items will not work since FreeBSD does not have RBS
+
+Build Prerequisites
+===================
+
+Compiling and building Ceph is tested on 11-CURRENT 
+It uses the CLANG toolset taht is available, which needs to be atleast 3.7
```
trociny: s/taht/that/

README.FreeBSD
```diff
@@ -0,0 +1,56 @@
...
+
+Building Ceph
+=============
+ - Go and start building
+	./do_freebsd.sh
+
+Parts not (yet) included:
+=========================
+
+ - RBD
```
trociny: s/RBD/RBD kernel client/

Userspace RBD is available under FreeBSD (both librbd and rbd tool (without linux specific commands) are buildable).

README.FreeBSD
```diff
@@ -0,0 +1,56 @@
...
+   Rados Block Devices is implemented in the Linux kernel
+   It seems that there used to be a userspace implementation first.
+   And perhaps ggated could be used as a template since it does some of the
+   same, other than just between 2 disks. And it has a userspace counterpart.
```
trociny: FYI, I started implementation based on geom_gate but it got stack, partially due to other activity, but also due to continuous frustration when build breaks after next master rebase.

do_freebsd.sh
```diff
@@ -0,0 +1,14 @@
...
+	    sudo ./install-deps.sh
+	fi
+	. ./autogen_freebsd.sh
+	./autogen.sh 
+	./configure ${CONFIGURE_FLAGS} 
+	( cd src/gmock/gtest; patch < /usr/ports/devel/googletest/files/patch-bsd-defines ) 
+	gmake -j$NPROC ENABLE_GIT_VERSION=OFF 
+	gmake -j$NPROC check ENABLE_GIT_VERSION=OFF CEPH_BUFFER_NO_BENCH=yes 
+
+
```
trociny: Why this indentation here?

src/kv/Makefile.am
```diff
@@ -16,12 +16,17 @@ if WITH_SLIBROCKSDB
...
-        NPROC = sysctl -n hw.ncpu
+NPROC = sysctl -n hw.ncpu
+ROCKSDBCXX_FLAGS += "-Wmismatched-tags"
+else
+NPROC = nproc
 endif
+
 rocksdb/librocksdb.a:
-	cd rocksdb && CC="${CC}" CXX="${CXX}" EXTRA_CXXFLAGS=-fPIC PORTABLE=1 $(MAKE) -j$(shell ${NPROC}) static_lib
+	cd rocksdb && CC="${CC}" CXX="${CXX}" EXTRA_CXXFLAGS="${ROCKSDBCXX_FLAGS}" PORTABLE=1 ${MAKE} -j$(shell ${NPROC}) static_lib
```
trociny: The commit log says "Fix NPROC and add compilerflags" but actually you adds only the flags. It is a little confusing.

src/test/run-cli-tests
```diff
@@ -41,7 +41,16 @@ mkdir -p "$FAKE_HOME"
 # cram doesn't like seeing the same foo.t basename twice on the same
 # run, so run it once per directory
 FAILED=0
+FAILEDTOOLS=""
 for tool in "$SRCDIR"/cli/*; do
+    if [ x`uname`x = xFreeBSDx ]; then
+    	if [ x"$tool"x = x"$SRCDIR/cli/rbd"x ]; then
```
trociny: The rbd tool already exists under FreeBSD. It has different set of commands though, so we need something more complicated here to test FreeBSD version. But in the first just skipping it might be ok.

README.FreeBSD
wjwithagen: 'mmmm,
I got so much breakage whilest building, that I move that way back on my planning.
But nice to know that I was wrong, and I'll rephrase.

do_freebsd.sh
wjwithagen: @trociny Leftovers from earlier versions. Will fix.

src/kv/Makefile.am
wjwithagen: @trociny 
I don't really get the point you are making?
For Clang/FreeBSD we add -Wmismatched-tags (which has an error, as I now see)
And I'm using your suggestion to fetch NPROC from sysctl

src/test/run-cli-tests
wjwithagen: Skipping this for the time is rather essential, since otherwise the Makefiles do not continue to the next set of tests. And then it it not possible to get a decent coverage when we Jenkins-ify the building. And I haven't looked into it, but I guess that adopting the tests for FreeBSD is rather big by itself.

src/kv/Makefile.am
trociny: @wjwithagen "fetch NPROC from sysctl" has already been added (before your patch), so the message is confusing.

wjwithagen: @trociny 
You are right...
Reason I changed it a bit is because in the current implementation nproc is always executed, even if not available. And I think that gmake even stops then.
So even the linux part needs to be conditionalised, not to break on FreeBSD.
But I'll fix the commit message to call it a rearange?

wjwithagen:
@gregsfortytwo @tchaikov 
Gentleman, is there anything more I need to do to get this merged?
I think all issues are addressed?

wjwithagen:
@liewegas 
Not to be inpatient ...
But this is already waiting for 2 weeks without further comments to be merged.
Is there anything further I need to do, or do I need to wait for a certain merge window?

tchaikov:
sorry for the latency, @wjwthagen. i think we still need to wait until jewel's stable release: the "10.2.1" tag. before that, we can merge only the bug fixes into master, see https://github.com/ceph/ceph/blob/master/doc/dev/index.rst#stable-release-candidates-i-e-x-1-z-phase-1. guess that's why we prioritize the bugs fixes and release related changes over pull requests for introducing new features, like the ones fro your.

wjwithagen:
Hi @tchaikov That is not a problem. FreeBSD has its own freeze periods a well. Only HEAD has a more relaxed approach to new features. Assuming that FreeBSD is a feature for Ceph 
So if that is the point I'll sit and wait. :)

But hen we should perhaps also already work on the next set of patches. So that we do not end up waiting for 11.2.1 to be released....
That was what I tried doing with #7573, but the end conclusion there is that is has grown too big.
So what is convenient? Of the reviewers to understand what is going on, for me to actually maintain the complete set of diff.

I now also understand that small diff which are actually "fixes" are better submitted as small bug-fix pulls. I will take that into the considerations of where to put the submission.

src/rgw/Makefile.am
```diff
@@ -93,6 +93,11 @@ librgw_la_SOURCES = \
 
 if WITH_OPENLDAP
 librgw_la_SOURCES += rgw/rgw_ldap.cc
+
+if FREEBSD
```
tchaikov: nit, could you move this block down to https://github.com/ceph/ceph/pull/8298/commits/886d907bd4930d7d1f1626a9f3e1611c3d8901d3#diff-d8612c96429ff7fbf2de590a08bca6dbR123, so all LIBRGW_DEPS related code can be grouped together?

configure.ac
```diff
@@ -17,7 +17,7 @@ RPM_RELEASE=0
 AC_SUBST(RPM_RELEASE)
 if test -d ".git" ; then
   AC_CHECK_PROG(GIT_CHECK, git, yes)
-  if test x"$GIT_CHECK" = x"yes"; then
+  if test x"$GIT_CHECK" = x"yes" -a x"$freebsd" != x"yes"; then
```
tchaikov: @wjwithagen i think we can leave it unchanged. yes, FreeBSD does not use rpm package manager. but it does not hurt to have the `RPM_RELEASE`.

wjwithagen: Emmm, it will generate an error/warning. And I don't like those if I can prevent them.
Now the whole autobuild stuff is littered with warnings. (atleast when I run it under FreeBSD) so it is just another one of may.
That being said, FreeBSD does have a rpm-package:
  /usr/ports/archivers/rpm4
I do not have that installed, but I'm not sure what would happen if it was.
So the alternative is to add it to the dependancies. Then I'm not sure what happens if it start building a FreeBSD-Ceph-RPM 8-)
I would just leave it as is..

src/rgw/Makefile.am
wjwithagen: Sure, done.

wjwithagen:
@liewegas 
Hi Sage,
I see these labels being attached to PULLs (like now: needs-qa), is there a document that describes these labels and what they do, what't the purpose, ....

configure.ac
tchaikov: okay. makes sense to me =)

src/test/Makefile-client.am
```diff
@@ -234,7 +235,7 @@ ceph_test_rados_api_aio_CXXFLAGS = $(UNITTEST_CXXFLAGS)
 bin_DEBUGPROGRAMS += ceph_test_rados_api_aio
 
 ceph_test_rados_api_list_SOURCES = test/librados/list.cc
-ceph_test_rados_api_list_LDADD = $(LIBRADOS) $(UNITTEST_LDADD) $(RADOS_TEST_LDADD)
+ceph_test_rados_api_list_LDADD = $(LIBRADOS) $(UNITTEST_LDADD) $(RADOS_TEST_LDADD) $(CEPH_GLOBAL)
```
tchaikov: @wjwithagen i don't quite understand this change, why we need $(CEPH_GLOBAL) here?

as ceph_test_rados_api_list is a typical librados client, in theory, it should only rely on librados.

tchaikov:
@wjwithagen for this one, see doc/dev/index.rst

> The Ceph community has access to the `Sepia lab
<http://ceph.github.io/sepia/>`_ where integration tests can be run on
real hardware. Other developers may add tags like "needs-qa" to your PR.
This allows PRs that need testing to be merged into a single branch and
tested all at the same time. Since teuthology suites can take hours
(even days in some cases) to run, this can save a lot of time.

src/test/Makefile-client.am
wjwithagen: @tchaikov 
Reason I needed to add this was that during linking I got unresolved references to ldap routines.
So I added CEPH_GLOBAL, which fixed it. And actually never looked back.

wjwithagen:
@tchaikov 
Right that one makes sense. How about the definition of all the other tags....
I've also seen, just being curious..:
build/ops 
core 
pending-discussion 
feature
bug fix

src/test/Makefile-client.am
tchaikov: @wjwithagen this does sound suspicious. as librados client should not reference ldap routines. could you post the error messages when linking this test executable?

wjwithagen: On 6-5-2016 05:18, Kefu Chai wrote:
> In src/test/Makefile-client.am
> <https://github.com/ceph/ceph/pull/8298#discussion_r62284821>:
>
>> @@ -234,7 +235,7 @@ ceph_test_rados_api_aio_CXXFLAGS = $(UNITTEST_CXXFLAGS)
>>  bin_DEBUGPROGRAMS += ceph_test_rados_api_aio
>>
>>  ceph_test_rados_api_list_SOURCES = test/librados/list.cc
>> -ceph_test_rados_api_list_LDADD = $(LIBRADOS) $(UNITTEST_LDADD) $(RADOS_TEST_LDADD)
>> +ceph_test_rados_api_list_LDADD = $(LIBRADOS) $(UNITTEST_LDADD) $(RADOS_TEST_LDADD) $(CEPH_GLOBAL)
>
> @wjwithagen <https://github.com/wjwithagen> this does sound suspicious.
> as librados client should not reference ldap routines. could you post
> the error messages when linking this test executable?

Eeergh, I'll have to rebuild.
Killing the current one, and will remove this addition, we'll see what 
happens

wjwithagen: @tchaikov 
Below the error that I get if I remove CEPH_GLOBAL from the LDADD list
test/librados/ceph_test_rados_api_list-list.o: In function `main':
/usr/srcs/Ceph/work/ceph/src/test/librados/list.cc:973: undefined reference to `global_init(std::__1::vector<char const*, std::__1::allocator<char const*> >*, std::__1::vector<char const*, std::__1::allocator<char const*> >&, unsigned int, code_environment_t, int, char const*, bool)'
/usr/srcs/Ceph/work/ceph/src/test/librados/list.cc:974: undefined reference to `g_ceph_context'

tchaikov:
lgtm

tchaikov:
@wjwithagen 

> build/ops 

the scripts for delivering the ceph: like building and packaging systems.

> core 

mostly related to ceph-osd and ceph-mon 

> pending-discussion 

this is a new label, i guess it's for pull requests still need more discussion about its design and intention before accepting reviews in details. please note that some reviews even try to apply the PR and play with it.

> feature

just new stuff, in contrary to bug-fix. 

> bug fix

i think bug fix is more likely get more attention. as it fixes existing code, and sometimes, we need to backport fixes applying to older releases also.

liewegas:
This is preventing teh ISA erasure-code plugin from being built (on linux).  See this failure

http://pulpito.ceph.com/sage-2016-05-07_10:20:37-rados-wip-sage-testing---basic-smithi/176935

which is caused because the isa.so isn't in the package here: 

http://gitbuilder.ceph.com/ceph-deb-trusty-x86_64-basic/sha1/13bc3ed90956334753ed8360c6d4dbe23d5d2939/pool/main/c/ceph/

wjwithagen:
@liewegas Ehhhmmm right..
Never had this thrown on my plate thusfar. 

Just for my understanding:
The testrun consists out of the Ceph/Ceph tree and all pulls you have added to 'wip-sage-testing', and that is thrown into teuthology?

And I'm assuming that the libex_isa.so should have been in the package that was build for this testrun.
Whose URL is the second link you mention, and the package created by 'wip-sage-testing'?

And not I have the task of finding out why the package builder does not contain libec_isa.so?

liewegas:
The packages include usr/lib/ceph/erasure-code/*, so the problem is just that the build isn't creating the isa plugin after these changes.  My guess is it is the arch changes around the yasm check, as the isa build is conditioned on HAVE_BETTER_YASM_ELF64 in erasure-code/Makefile.am.

liewegas:
<pre>checking for yasm... yes
we have a modern and working yasm
x86_64: =: No such file or directory
x86_64: =: No such file or directory
we are not x86_64 && !x32
</pre>
is the build output, at http://gitbuilder.sepia.ceph.com/gitbuilder-ceph-deb-trusty-amd64-basic/log.cgi?log=13bc3ed90956334753ed8360c6d4dbe23d5d2939

wjwithagen:
right, that would do it.
It had to be something like that, since this is one of the first tests to run, so it seems

I just started building again in my CentOS7 box. But this is faster.
I saw that this is run on Ubuntu, I guess that I'll need a Ubuntu VM as well.

And I noticed that 13bc3ed90956334753ed8360c6d4dbe23d5d2939 in the log-URLis the build_sha1 in pulpito. 
Is that something I could have known? Because thus far I only looked at the Jenkins logs.

autogen_freebsd.sh
```diff
@@ -0,0 +1,63 @@
+# Set the FreeBSD specific configure flags
+FREEBSD_CONFIGURE_FLAGS=
+if [ x`uname`x = x"FreeBSD"x ]; then
+    MAKE=gmake
+    # We need at least something > clang 3.4
+    # tested with package clang37 on FreeBSD 10.2 ( Which has 3.4 as default )
+    if clang -v 2>&1 | grep -q "3.8" ; then
+        CC=clang
+        CXX=clang++
+    elif -f /usr/local/bin/clang37 ]; then
```
tchaikov: elif [ -f /usr/local/bin/clang37 ]; then

wjwithagen:
@liewegas It is a wierd error:
the code in configure.ac is like
     echo 'we have a modern and working yasm'
     arch_x86_64=0
     if [ `uname -p` = "amd64" ]; then
        arch_x86_64=1
     elif which arch 2>&1 > /dev/null; then
        [ `arch` = "x86_64" ] && arch_x86_64=1
     fi

And I just checked Ubuntu has both executables in /usr/bin, (at least the ubuntu we run our cluster on). And they both return x86_64, but it seems that the return result is empty 
Now I can guard the compares, but that only obfuscates the fact that there is no uname or arch executed. 
Given the autoconfig language it could very well be that a [ is too special and that the Ubuntu version brings this to surface so I'll replace with test(1).

configure.ac
```diff
@@ -243,7 +243,14 @@ AC_CHECK_PROG(YASM_CHECK, yasm, yes)
 if test x"$YASM_CHECK" = x"yes"; then
   if yasm -f elf64 src/common/crc32c_intel_fast_asm.S -o /dev/null; then
      echo 'we have a modern and working yasm'
-     if test `arch` = "x86_64" ; then
+     # FreeBSD has uname -p not arch(1), so do not exec arch
+     arch_x86_64=0
+     if [ `uname -p` = "amd64" ]; then
+	arch_x86_64=1
+     elif which arch 2>&1 > /dev/null; then 
+	[ `arch` = "x86_64" ] && arch_x86_64=1
```
tchaikov: @wjwithagen the bracket is removed by autoconf. we'd better stick to 
```sh
if test
```

do_freebsd.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh -xve
+NPROC=`sysctl -n hw.ncpu`
+
+if [ x"$1"x = x"--deps" ]; then
```
tchaikov: `x"$1"x = x"--deps"x` ?
and i think 
```sh
if [ "$1" = "--deps" ]; then
```
would suffice

tchaikov:
@wjwithagen 13bc3ed90956334753ed8360c6d4dbe23d5d2939 is the sha1 of sage's testing branch tested by http://gitbuilder.sepia.ceph.com/gitbuilder-ceph-deb-trusty-amd64-basic/log.cgi?log=13bc3ed90956334753ed8360c6d4dbe23d5d2939. and his branch is now overwritten, so 
```
git branch --contains 13bc3ed90956334753ed8360c6d4dbe23d5d2939
```
does not show anything now.

do_freebsd.sh
wjwithagen: @tchaikov yes and no...
I do not like comparing strings to "empty"' values. Sometimes strange things will happen.
And perhaps it is old-school, but I've been taught to wrap compares like this with x""x

tchaikov: the reason why we need `[ x$1 = x"foo" ]` is that we don't have `$1` quoted. iiuc, the old school shell looks like:
```sh
if [ x$1= x--deps ]; then
```

configure.ac
wjwithagen: @tchaikov that was my conclusion too, in the end

wjwithagen:
@tchaikov  @liewegas 
Error should be fixed now.
This pull does not reenter building, any particular reason why not.

wjwithagen:
@tchaikov @liewegas 
Ignore, since it might have just crossed testing to be started at the same point.

liewegas:
@tchaikov this passed a rados run.. ready to merge?

liewegas:
http://pulpito.ceph.com/sage-2016-05-16_01:43:51-rados-wip-sage-testing---basic-smithi/

tchaikov:
@liewegas yes!

wjwithagen:
@liewegas Cool
@tchaikov @trociny @gregsfortytwo Thanx for all the support and help.

Up and onto the next set of commits from #7573 Another 81++ to go.
And still more in my local work. 
But little steps will get us there. I'll start working on a new set to actually get us finished compilation with Clang >= 3.7 ...
Please note that I saw that on CentOS 7.2 Clang is version 3.4.
And with that version I was not able to build the tree, let alone get it to run correctly.
So that would require looking for packages that give at least Clang 3.7  on Linux, of compiling with Clang would ever become a target.

wjwithagen:
@liewegas @tchaikov @trociny @gregsfortytwo 
There is a new pull #9178, with more FreeBSD compile time stuff
Hopefully this pull is a manageable chunk to get feedback on.
Once we have this one in as well, the real interesting commits start to get all the tests working.

PR #8299, opened 2016-03-24
messages/MOSDOp: clear reqid inc for v6 encoding

liewegas:
For the new v7 encoding, we put the client incarnation in the reqid so that
we can have it at the beginning of the encoding.  However, we do *not*
want to do this for the v6 encoding, as that version of get_reqid() will
return reqid if it is not == osd_reqid_t().  Previously we were setting
the field in the constructor, but since it's specific to the v7 encoding,
put it there (in encode_payload()) instead.

Fixes: #15230

### DIFF
--- a/src/messages/MOSDOp.h
+++ b/src/messages/MOSDOp.h
@@ -315,7 +315,13 @@ struct ceph_osd_request_head {
 
       ::encode(retry_attempt, payload);
       ::encode(features, payload);
-      ::encode(reqid, payload);
+      if (reqid.name != entity_name_t() || reqid.tid != 0) {
+	::encode(reqid, payload);
+      } else {
+	// don't include client_inc in the reqid for the legacy v6
+	// encoding or else we'll confuse older peers.
+	::encode(osd_reqid_t(), payload);
+      }
     } else {
       // new, reordered, v7 message encoding
       header.version = HEAD_VERSION;


src/messages/MOSDOp.h
```diff
@@ -323,7 +319,15 @@ struct ceph_osd_request_head {
...
-      ::encode(reqid, payload);
+      if (reqid == osd_reqid_t()) {
+	// also put the client_inc in reqid.inc, so that get_reqid() can
+	// be used before the full message is decoded.
+	osd_reqid_t fake_reqid;
+	fake_reqid.inc = inc;
+	::encode(fake_reqid, payload);
+      } else {
+	::encode(reqid, payload);
+      }
```
gregsfortytwo: Doesn't this need to use the actual reqid, not just fill in the inc?

liewegas: This is the reqid == osd_reqid_t path...

src/messages/MOSDOp.h
```diff
@@ -323,7 +319,15 @@ struct ceph_osd_request_head {
       ::encode(osdmap_epoch, payload);
       ::encode(flags, payload);
       ::encode(reassert_version, payload);
-      ::encode(reqid, payload);
+      if (reqid == osd_reqid_t()) {
+	// also put the client_inc in reqid.inc, so that get_reqid() can
+	// be used before the full message is decoded.
+	osd_reqid_t fake_reqid;
+	fake_reqid.inc = inc;
```
jdurgin: compile error from the bot:

```
./messages/MOSDOp.h: In member function 'virtual void MOSDOp::encode_payload(uint64_t)':
./messages/MOSDOp.h:326:19: error: 'inc' was not declared in this scope
  fake_reqid.inc = inc;
```

src/messages/MOSDOp.h
gregsfortytwo: d'oh. Something about this this whole setup was confusing me previously, but I agree it looks good now.

src/messages/MOSDOp.h
gregsfortytwo: Yeah, that should be client_inc here. Bad copy from the constructor.

PR #8257, opened 2016-03-22
#15065: Add standby_for_fscid in MMDSBeacon and handle it

jcsp:
This PR depends on https://github.com/ceph/ceph/pull/7998, the new stuff here starts at the "s/standby_for_ns/standby_for_fscid/" commit

PR #8300, opened 2016-03-24
mds: Fix standby replay MDS in upgrades and simplify standby replay logic (#15591)

jcsp:
Depends on https://github.com/ceph/ceph/pull/8257

### DIFF
--- a/src/mds/MDSMap.cc
+++ b/src/mds/MDSMap.cc
@@ -19,11 +19,6 @@
 using std::stringstream;
 
 
-const mds_rank_t MDSMap::MDS_NO_STANDBY_PREF(-1);
-const mds_rank_t MDSMap::MDS_STANDBY_ANY(-2);
-const mds_rank_t MDSMap::MDS_STANDBY_NAME(-3);
-const mds_rank_t MDSMap::MDS_MATCHED_ACTIVE(-4);
-
 // features
 CompatSet get_mdsmap_compat_set_all() {
   CompatSet::FeatureSet feature_compat;
@@ -83,6 +78,7 @@ void MDSMap::mds_info_t::dump(Formatter *f) const
   f->dump_int("standby_for_rank", standby_for_rank);
   f->dump_int("standby_for_fscid", standby_for_fscid);
   f->dump_string("standby_for_name", standby_for_name);
+  f->dump_bool("standby_replay", standby_replay);
   f->open_array_section("export_targets");
   for (set<mds_rank_t>::iterator p = export_targets.begin();
        p != export_targets.end(); ++p) {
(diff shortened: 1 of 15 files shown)


gregsfortytwo:
@jcsp be great to rebase this now that stuff is merged.

gregsfortytwo:
You know, I was going to wait until after Jewel to merge this but it's dragged on so long and this isn't exactly high-risk so I'll probably put it in. Just going to wait until a bunch of other stuff gets merged in case dropping the oneshot replay or changing the output causes test text match failures in the integration testing.

You did run this through vstart unit testing, right @jcsp?

jcsp:
fs suite run here: http://pulpito.ceph.com/jspray-2016-04-29_19:36:29-fs-wip-standby-replay-reorg---basic-mira/

(the 1 failure is an osd one)

jcsp:

src/mds/MDSMap.h
```diff
@@ -140,12 +130,15 @@ class MDSMap {
     mds_rank_t standby_for_rank;
     std::string standby_for_name;
     fs_cluster_id_t standby_for_fscid;
+    bool standby_replay;
```
gregsfortytwo: I'm concerned we may be inappropriately losing the standby-replay intentions of older daemons; will come back to that...

src/mds/MDSMap.cc
```diff
@@ -462,6 +459,9 @@ void MDSMap::mds_info_t::decode(bufferlist::iterator& bl)
   if (struct_v >= 6) {
     ::decode(standby_for_fscid, bl);
   }
+  if (struct_v >= 7) {
+    ::decode(standby_replay, bl);
+  }
```
gregsfortytwo: I think we should be trying to infer standby_replay state of the MDS on decode, if it's an older encoding. Otherwise it can lose the state, right, and I don't think it'll get it re-granted automatically unless maybe the MDS requests it to go back (not sure).

gregsfortytwo:
Running an upgrade test as well: http://pulpito.ceph.com/gregf-2016-05-02_11:18:38-upgrade:infernalis-x-wip-standby-replay-reorg---basic-mira/

src/mds/MDSMap.cc
jcsp: In the case of a daemon that is not in standby replay state but wants to be: old MDS daemons will continue to send beacons that request state standby replay, which MMDSBeacon::decode will interpret as standby_replay=true, which gets picked up in MDSMonitor::prepare_beacon.

gregsfortytwo: Okay, so the map will swap it to be a regular standby but its next beacon will request standby-replay and life will then go on as normal? Sounds fine.

jcsp:
The upgrade test failed about half the jobs after it couldn't lock nodes

gregsfortytwo:
turns out upgrade tests hard-code certain OSes and the mira machines lack the numbers for some of them. :(

PR #8301, opened 2016-03-24
cmake: only build with _FORTIFY_SOURCE with -O

tchaikov:
http://tracker.ceph.com/issues/15263 Fixes: #15263


tchaikov:
@kmroz ping

tchaikov:
@cbodley could you take a look at this change?

cbodley:
@tchaikov the cmake bits look fine

PR #8302, opened 2016-03-24
common: update config parsing error for clarity

gregsfortytwo:
http://tracker.ceph.com/issues/15235 describes a user getting this
message despite not trying to open the config file. And indeed, it's
what gets spat out if internal_safe_to_start_threads is false. Update
the output to reflect that being the most likely error.message

### DIFF
--- a/qa/workunits/cephtool/test.sh
+++ b/qa/workunits/cephtool/test.sh
@@ -814,6 +814,8 @@ function test_mon_mds()
   ceph mds remove_data_pool $data3_pool
   ceph osd pool delete data2 data2 --yes-i-really-really-mean-it
   ceph osd pool delete data3 data3 --yes-i-really-really-mean-it
+  expect_false ceph mds set_max_mds 4
+  ceph mds set allow_multimds true --yes-i-really-mean-it
   ceph mds set_max_mds 4
   ceph mds set_max_mds 3
   ceph mds set_max_mds 256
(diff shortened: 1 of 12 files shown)


tchaikov:
parse_config_files() bails out if internal_safe_to_start_threads is true.

and i think http://tracker.ceph.com/issues/15235 failed due to something else other than http://tracker.ceph.com/issues/14144.

because, 

- main()
    - global_init()
        - global_pre_init()
            - conf->parse_config_files()
                - if internal_safe_to_start_threads: return ENOSYS
    - common_init_finish()
        - start_service_thread()
            - internal_safe_to_start_threads = true

`parse_config_files()` is called before we start any threads. and the "internal_safe_to_start_threads" flag is set when we `start_service_thread()` in `common_init()` after `global_init()` which calls `parse_config_files()` in `global_pre_init()`. and MDS has barely done anything before `common_init_finish()` so it can hardly interfere with the "internal_safe_to_start_threads" conf var.

maybe we can also `dout_emergency(cpp_strerror(ret))` before `_exit(1)`?

gregsfortytwo:
Bah, I misread that conditional pretty badly. I think the bug is somehow still a result of the FD issue though (mostly just because that's one glaring problem with the system).

gregsfortytwo:
Well, I accidentally overwrote my branch and it was apparently wrong anyway, so killing this.

PR #8303, opened 2016-03-25
os/bluestore/BlueStore: fix bug for using cache_tail.

majianpeng:
If using cache tail and the extent isn't unwritten, this mean overwrite
the existed tail-data(all or part). It must using wal rather than
directly write.
For example:
write(0-2K)-->cache tail 2k, offset=0
write(2k, 60K)-->using cache tail
   write(0, 64K)-->overwrite the exsited data (0, 2k)
Because overwrite existed data, for data integrity, it must use
wal-mode.

### DIFF
--- a/src/os/bluestore/BlueStore.cc
+++ b/src/os/bluestore/BlueStore.cc
@@ -5520,6 +5520,7 @@ int BlueStore::_do_write(
 
     // use cached tail block?
     uint64_t tail_start = o->onode.size - o->onode.size % block_size;
+    bool partly_overwrite_tail = false;
     if (offset >= bp->first &&
 	offset > tail_start &&
 	offset + length >= o->onode.size &&
@@ -5545,6 +5546,9 @@ int BlueStore::_do_write(
 	bl.swap(t);
       } else {
 	bufferlist t;
+	//partly overwrite cache tail. Block size don't equal sector size
+	//so block update isn't a atomic op.
+	partly_overwrite_tail = true;
 	t.substr_of(o->tail_bl, 0, tail_off);
 	offset -= t.length();
 	length += t.length();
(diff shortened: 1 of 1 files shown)


liewegas:
Can you please add a test to store_test.cc that reproduces this case?  Preferably using the min_alloc_size config value to calculate the write sizes/offsets?

majianpeng:
For the final result, those don't have any different. So i don't know how to check whether use wal-mode or other?

majianpeng:
updated, please review again. Thanks!

liewegas:
I separated just hte test case into a new commit at http://github.com/liewegas/ceph/commit/wip-cache-tail-bug but I can't reproduce the issue.

majianpeng:
Consider those op:
  write(off=0, len=2048)
  write(off=1024, len=63K)
The following is the message:
>>bluestore(/mnt/ceph/src/dev/osd2) _do_write #0:9ca9c3fd:::intel:head# 0~2048 - have 0 bytes in 0 extents
>>bluestore(/mnt/ceph/src/dev/osd2) _do_allocate #0:9ca9c3fd:::intel:head# 0~2048 - have 0 bytes in 0 extents
>>bluestore(/mnt/ceph/src/dev/osd2)   io falls within 0~65536
>>bluestore(/mnt/ceph/src/dev/osd2)   not yet allocated
>>bluestore(/mnt/ceph/src/dev/osd2)   must alloc 0~65536
>>stupidalloc reserve need 65536 num_free 9659678720 num_reserved 0
>>stupidalloc allocate want_size 65536 alloc_unit 65536 hint 0
>>bluestore(/mnt/ceph/src/dev/osd2) _do_allocate  alloc 0: 1077739520~65536:unwritten
>>bluestore(/mnt/ceph/src/dev/osd2) _do_write  chunk 0~2048 extent 0: 1077739520~65536:unwritten
>>bluestore(/mnt/ceph/src/dev/osd2) _pad_zeros_tail cached 2048 of tail block at 0
>>bluestore(/mnt/ceph/src/dev/osd2) _pad_zeros_tail pad 2048 on back, now 0~4096
>>bluestore(/mnt/ceph/src/dev/osd2) _do_write write 0~4096 x_off 0
>>bluestore(/mnt/ceph/src/dev/osd2) _do_overlay_trim #0:9ca9c3fd:::intel:head# 0~4096
>>bluestore(/mnt/ceph/src/dev/osd2) _do_write extending size to 2048
>>bluestore(/mnt/ceph/src/dev/osd2) _write 0.1_head #0:9ca9c3fd:::intel:head# 0~2048 = 0
>>bluestore(/mnt/ceph/src/dev/osd2) _write 0.1_head #0:9ca9c3fd:::intel:head# 1024~64512
>>bluestore(/mnt/ceph/src/dev/osd2) _do_write #0:9ca9c3fd:::intel:head# 1024~64512 - have 2048 bytes in 1 extents
>>bluestore(/mnt/ceph/src/dev/osd2) _do_allocate #0:9ca9c3fd:::intel:head# 1024~64512 - have 2048 bytes in 1 extents
>>bluestore(/mnt/ceph/src/dev/osd2)   io falls within 0~65536
>>bluestore(/mnt/ceph/src/dev/osd2)   will presumably WAL
>>bluestore(/mnt/ceph/src/dev/osd2) _do_write  chunk 1024~64512 extent 0: 1077739520~65536
>>bluestore(/mnt/ceph/src/dev/osd2) _do_write using cached tail
>>bluestore(/mnt/ceph/src/dev/osd2) _pad_zeros pad 0 + 0 on front/back, now 0~65536
>>bluestore(/mnt/ceph/src/dev/osd2) _do_write write 0~65536 x_off 0
>>bluestore(/mnt/ceph/src/dev/osd2) _do_write extending size to 65536

For this case, it should use wal. But from the mesages, it don't use. So if met error, the data maybe change partly. If no error occur, the final data is expected. We only from the messages find the bug.
Should we remove the test case?

liewegas:
Oh I see.  I think in this case it is actually safe, though, since the 
block update should be atomic (old value or new value), and we ignore 
garbage after EOF.  This was actually why I switched from the old 
scheme (tail of final block had to be zeros) to this scheme (tail of block 
is undefined)... so that we could do small appends without WAL.

majianpeng:
block_size don't equal hard-sector size.  For example, hdd: ST1000NM0033-9ZM173 , the physical sector size is 512B. The block_size from stat() is 4096B.  So block update(old value partly update) isn't a atomic op. We should consider this case.

liewegas:
Hmm, yeah that's true.  And now that I think about it again the atomic part isn't actually important, as long as we don't scramble the data that is already there.  Since the area we're modifying is past the end of the object/file, we don't care what garbage is there until we commit the transaction to update the size.  Which means this should be safe to overwrite.

PR #8305, opened 2016-03-25
vstart: make -k with optional mon_num (part 2)

tchaikov:
vstart: use default CEPH_NUM_* if awk fails

see #8251


tchaikov:
@dillaman could you help take a look? thanks!

src/vstart.sh
```diff
@@ -278,10 +282,14 @@ shift
 done
 
 if [ "$overwrite_conf" -eq 0 ]; then
-  CEPH_NUM_MON=`awk -F= '/CEPH_NUM_MON/{print $2}' $conf_fn`
-  CEPH_NUM_OSD=`awk -F= '/CEPH_NUM_OSD/{print $2}' $conf_fn`
-  CEPH_NUM_MDS=`awk -F= '/CEPH_NUM_MDS/{print $2}' $conf_fn`
-  CEPH_NUM_RGW=`awk -F= '/CEPH_NUM_RGW/{print $2}' $conf_fn`
+    MON=`awk -F= '/CEPH_NUM_MON/{print $2}' $conf_fn`
```
dillaman: Can you just do something like: CEPH_NUM_MON=`awk -F= '/CEPH_NUM_MON/{print $2}' $conf_fn` before line 79?  If the awk fails, it will fall back to the defaults.

tchaikov: if we go this way, we will always continue with whatever `CEPH_NUM_{MON,OSD,MDS,RGW}` the last vstart run used, even if "-k" is not passed in. in other words,
```bash
$ MON=1 OSD=1 MDS=0 ./vstart.sh
$ ./stop
$ MON=1 OSD=2 MDS=0 ./vstart.sh
# what? we have only 1 OSD?
```

if we move these lines after line 91, we overwrite the options for a non "-k" vstart cluster.

dillaman: Fair enough.

trociny:
This comment is rather for the initial PR (part 1) and can be freely ignored... Anyway, I don't not like very much when sed/awk/... is used to extract values from files like this (that can be parsed by a specialized parser). E.g. the current awk parsing will break when ceph.conf happens to include CEPH_NUM_MON word more than one time.

Instead, we could add e.g. "client.vstart.sh" section to the generated ceph.conf:
```
[client.vstart.sh]
  ceph num mon = 5
  ...
```
And parse it using `ceph-conf`:
```
MON=`ceph-conf -c $conf_fn --name client.vstart.sh ceph_num_mon`
```

tchaikov:
@trociny thanks for the suggestion! i will update this PR later on.

tchaikov:
changelog

- store the number of {osd,mon,mds,rgw} as settings in the section of "[client.vstart.sh]"
- read the remembered using ceph-conf instead of parsing the ceph.conf using awk.

hope this is okay to introduce the dependency on ceph-conf. but guess we don't use "-k" in our testing, so it's relatively safe.

dillaman:
My only comment is that perhaps given the distance between usages of "client.vstart.sh", should it use an environment variable "constant"?

tchaikov:
changelog

- use a variable for "client.vstart.sh" instead of hardwiring it to "client.vstart.sh" at two separated places

tchaikov:
@dillaman ping?

src/vstart.sh
```diff
@@ -229,6 +231,10 @@ case $1 in
 	    cephx=0
 	    ;;
     -k )
+	    if [ ! -r $conf_fn ]; then
+	        echo "can not use old configuration: $conf_fn not readable." >&2
```
dillaman: Super minor: s/can not/cannot/g

tchaikov:
changelog

- s/can not/cannot/

@dillaman thanks for the review, fixed and repushed.

tchaikov:
@dillaman ping?

PR #8307, opened 2016-03-25
mon: paxos small optimization on store_state

songbaisen:
mon: paxos small optimization on store_state

### DIFF
--- a/src/mon/Paxos.cc
+++ b/src/mon/Paxos.cc
@@ -357,7 +357,7 @@ bool Paxos::store_state(MMonPaxos *m)
     // ignore everything if values start in the future.
     dout(10) << "store_state ignoring all values, they start at " << start->first
 	     << " > last_committed+1" << dendl;
-    start = m->values.end();
+    return false;
   }
 
   // push forward the start position on the message's values iterator, up until


src/mon/Paxos.cc
```diff
@@ -358,6 +358,7 @@ bool Paxos::store_state(MMonPaxos *m)
     dout(10) << "store_state ignoring all values, they start at " << start->first
 	     << " > last_committed+1" << dendl;
     start = m->values.end();
+    return changed;
```
tchaikov: if we want to early return here, maybe we can remove the line of `start = m->values.end();` also.

songbaisen:
@tchaikov  Yes,Thank you for view.I have done it.

jecluis:
Also, personally, I think returning just 'false' would be a bit more readable than returning 'changed' (which is set to false at this point).

songbaisen:
@jecluis  Yes,Done it.It make clear here.

athanatos:
@jecluis It's ok to skip remove_legacy_versions() in that case?

athanatos:
This has passed tests (sjust@teuthology:/a/samuelj-2016-05-16_20:02:21-rados-wip-sam-testing-distro-basic-smithi).  Once someone reviews it, it can merge.

@tchaikov @jecluis Is this ok?

tchaikov:
lgtm.

PR #8317, opened 2016-03-27
Gentoo support for ceph-disk / ceph-detect-init; pip speedup

robbat2:
Done because this was very useful while testing static-sites.

### DIFF
--- a/src/ceph-detect-init/ceph_detect_init/gentoo/__init__.py
+++ b/src/ceph-detect-init/ceph_detect_init/gentoo/__init__.py
@@ -1,0 +1,37 @@
+import os
+distro = None
+release = None
+codename = None
+
+
+# From ceph-disk, but there is no way to access it since it's not in a module
+def is_systemd():
+    """
+    Detect whether systemd is running;
+    WARNING: not mutually exclusive with openrc
+    """
+    with open('/proc/1/comm', 'rb') as i:
+        for line in i:
+            if 'systemd' in line:
+                return True
+    return False
+
+
+def is_openrc():
+    """
+    Detect whether openrc is running.
+    """
+    OPENRC_CGROUP = '/sys/fs/cgroup/openrc'
+    return os.path.isdir(OPENRC_CGROUP)
+
+
+def choose_init():
+    """Select a init system
+
+    Returns the name of a init system (upstart, sysvinit ...).
+    """
+    if is_openrc():
+        return 'openrc'
+    if is_systemd():
+        return 'systemd'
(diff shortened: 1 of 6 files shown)


dx9:
@robbat2 :+1: 07b76fd fixes ceph-disk on Alpine too.

Had to fix the error below to get ceph-disk/run-tox.sh to pass

ceph_disk/main.py:2850:80: E501 line too long (82 > 79 characters)

robbat2:
@dx9 does Alpine need a block added to _normalized_distro_name?

dx9:
@robbat2 Don't think so

Alpine doesn't appear to implement lsb_release format. platform.linux_distribution() detects /etc/alpine-release which only has a version.

cat /etc/alpine-release
3.3.3

robbat2:
jenkins: retest this please

robbat2:
jenkins: retest this please (again)

src/ceph-detect-init/ceph_detect_init/gentoo/__init__.py
```diff
@@ -0,0 +1,37 @@
...
+            if 'systemd' in line:
+                return True
+    return False
+
+
+def is_openrc():
+    """
+    Detect whether openrc is running.
+    """
+    OPENRC_CGROUP = '/sys/fs/cgroup/openrc'
```
prometheanfire: Does openrc ALWAYS create this cgroup now?  I thought it was optional.

src/ceph-detect-init/requirements.txt
```diff
@@ -1 +0,0 @@
-argparse
```
prometheanfire: was this removed because it is an assumed builtin?

src/ceph-disk/requirements.txt
```diff
@@ -1 +0,0 @@
-argparse
```
prometheanfire: same

src/ceph-detect-init/requirements.txt
robbat2: Removing it since builtin means this tool can be built without it trying to hit network at all (it ignored the system copy otherwise)

onyb: Currently `setup.py` adds it to requirements (and installs over network) only if it is not a builtin for the particular Python version.

Also take a look at `setup.py` file in my pull request https://github.com/ceph/ceph/pull/9157

tchaikov:

tchaikov:
lgtm

PR #8323, opened 2016-03-28
msg/async: add numevents statistics for external_events

Yan-waller:
Maybe we need to add the statistics for external_events.

### DIFF
--- a/src/msg/async/Event.cc
+++ b/src/msg/async/Event.cc
@@ -430,6 +430,7 @@ int EventCenter::process_events(int timeout_microseconds)
         if (e)
           e->do_request(0);
         cur_process.pop_front();
+        numevents++;
       }
     }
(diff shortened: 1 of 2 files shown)


src/msg/async/Event.cc
```diff
@@ -339,7 +339,7 @@ int EventCenter::process_events(int timeout_microseconds)
   // Must set owner before looping
   assert(owner);
   struct timeval tv;
-  int numevents;
```
yuyuyu101: we don't need to init this variable since it must be assigned later.

Yan-waller: @yuyuyu101  thanks for your review, and I have fixed it.

Yan-waller:
@yuyuyu101 Ping

PR #8324, opened 2016-03-28
mon: remove remove_legacy_versions()

tchaikov:
as it's required to upgrade to hammer before moving to jewel, and
hammer is already using the single-paxos monitor. and we convert
the store.db when starting up from an old monitor (bobtail). we
stopped the conversion since 1d814b7.


jecluis:
LGTM

PR #8329, opened 2016-03-28
osd: persist info into disk if we fixed same_interval_since during load_pgs()

xiexingguo:
Also drop a useless perf counter and fix a perf counter name


xiexingguo:
@liewegas Not very sure of https://github.com/ceph/ceph/pull/8329/commits/5510f43fbf556f55ce8aecf19c8d9bfd2ea8332a. Please help to check.

PR #8330, opened 2016-03-28
common: fix race during optracker switches between enabled/disabled mode

xiexingguo:
The RWLock lock is made private to protect tracking_enabled changing
but we are reading tracking_enabled beyond its protection during
unregister_inflight_op(). So this is a slight race conditon.

### DIFF
--- a/src/common/TrackedOp.cc
+++ b/src/common/TrackedOp.cc
@@ -164,6 +164,7 @@ void OpTracker::unregister_inflight_op(TrackedOp *i)
   }
   i->_unregistered();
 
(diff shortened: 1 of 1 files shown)


dzafman:
My testing doesn't find the lack of a lock here matters.  The peek at the value of tracking_enabled is a clean race.

dzafman:
LGTM  It makes sense to put this in.  If we ever wanted to purge the tracking history when disabling tracking we need this lock.

@liewegas Let's merge this once it passes wip-sage-testing.

PR #8333, opened 2016-03-28
script: subscription-manager support (part 3)

dachary:
Renaming subman.py in subman is cute but does not prevent the generation
of the py[co]. Exclude them explicitly from packaging instead.


dachary:
pushed in the gitbuilders under the branch wip-subman-facts

PR #8334, opened 2016-03-28
script: subscription-manager support (part 3)

dachary:
Renaming subman.py in subman is cute but does not prevent the generation
of the py[co]. Exclude them explicitly from packaging instead.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -155,7 +155,7 @@ ceph_libexec_SCRIPTS = ceph_common.sh ceph-osd-prestart.sh
 
 if ENABLE_SUBMAN
 submandir = /etc/cron.hourly
-subman_SCRIPTS = script/subman
+subman_DATA = script/subman
 endif
 
 # tests to actually run on "make check"; if you need extra, non-test,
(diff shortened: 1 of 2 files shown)


dachary:
pushed in the gitbuilders under the name wip-subman-facts

branto1:
@dachary: I was looking for some ways to avoid creating these files and two possible solutions came up (it might be best if we used them both to be sure):
- use `#!/usr/bin/env python -B` -- this should help avoid creating the files when run as a script
- add subman script to subman_DATA instead of subman_SCRIPTS -- this should tell autotools to not create the pyc/pyo files

dachary:
@branto1 the gitbuilders seem to be happy about the current branch. If you think this is not the way to go, I'll revise according to your suggestions and re-submit. Just let me know what you prefer :-)

branto1:
@dachary Sure, this solution should work but it is mostly just a work-around which we could hopefully avoid if we patched the build system instead of the spec file. Could you try to incorporate the suggestions to see if that helps? We can fall back to this approach if it does not help.

dachary:
@branto1 repushed with the suggested changes and also pushed to gitbuilders

branto1:
I tried building the wip-subman-facts branch on rhel and it went as expected, everything (the cron.hourly subman script) was packaged properly, no pyc/pyo files. The gibuilders seemed to have built this ok as well. I think we can merge this.

PR #8336, opened 2016-03-28
Pipe: let peer retry when existing is closing

yuyuyu101:
If accepting a new connection and the existing just mark down,
we need to let peer retry connect and exit accept process.

Just like AsyncConnection.cc:
https://github.com/ceph/ceph/blob/master/src/msg/async/AsyncConnection.cc#L1686

Fixes: http://tracker.ceph.com/issues/14266

### DIFF
--- a/src/msg/simple/Pipe.cc
+++ b/src/msg/simple/Pipe.cc
@@ -466,6 +466,16 @@ int Pipe::accept()
     existing = msgr->_lookup_pipe(peer_addr);
     if (existing) {
       existing->pipe_lock.Lock(true);  // skip lockdep check (we are locking a second Pipe here)
+      if (existing->state == STATE_CLOSED) {
+        ldout(msgr->cct, 1) << __func__ << " existing racing replace or mark_down happened while replacing."
+                            << " existing_state=" << existing->get_state_name() << dendl;
+	reply.tag = CEPH_MSGR_TAG_RETRY_GLOBAL;
+	reply.global_seq = existing->peer_global_seq;  // so we can send it below..
+	existing->pipe_lock.Unlock();
+	msgr->lock.Unlock();
+	goto reply;
+      }
+
       if (existing->reader_dispatching) {
 	/** we need to wait, or we can deadlock if downstream
 	 *  fast_dispatchers are (naughtily!) waiting on resources


gregsfortytwo:
Can you explain how we get into this situation?

I think if we get into that block we aren't *allowed* to accept the connection as it's been explicitly marked down and the other side hasn't acknowledged that yet.

yuyuyu101:
@gregsfortytwo if osd side firstly accept one connection and register pipe, but failed at https://github.com/ceph/ceph/blob/master/src/msg/simple/Pipe.cc#L732 . while this connection isn't a lossy connection, it won't unregister pipe. Then we accept a new connection and lookup this pipe, we will go through this condition

src/msg/simple/Pipe.cc
```diff
@@ -466,6 +466,16 @@ int Pipe::accept()
     existing = msgr->_lookup_pipe(peer_addr);
     if (existing) {
       existing->pipe_lock.Lock(true);  // skip lockdep check (we are locking a second Pipe here)
+      if (existing->state == STATE_CLOSED) {
+        ldout(msgr->cct, 1) << __func__ << " existing racing replace or mark_down happened while replacing."
+                            << " existing_state=" << existing->get_state_name() << dendl;
+	reply.tag = CEPH_MSGR_TAG_RETRY_GLOBAL;
+	reply.global_seq = existing->peer_global_seq;  // so we can send it below..
```
tchaikov: we might need to unlock `existing->pipe_lock` also here.

src/msg/simple/Pipe.cc
```diff
@@ -466,6 +466,16 @@ int Pipe::accept()
     existing = msgr->_lookup_pipe(peer_addr);
     if (existing) {
       existing->pipe_lock.Lock(true);  // skip lockdep check (we are locking a second Pipe here)
+      if (existing->state == STATE_CLOSED) {
+        ldout(msgr->cct, 1) << __func__ << " existing racing replace or mark_down happened while replacing."
+                            << " existing_state=" << existing->get_state_name() << dendl;
+	reply.tag = CEPH_MSGR_TAG_RETRY_GLOBAL;
```
tchaikov: why not send a RETRY_SESSION message and ask the peer to bump up its cseq? so its cseq won't collide with the existing one. if this makes sense to you, maybe we can reuse the "retry_session:", in other words
```diff
--- a/src/msg/simple/Pipe.cc
+++ b/src/msg/simple/Pipe.cc
@@ -524,11 +524,12 @@ int Pipe::accept()
 
       if (connect.connect_seq == existing->connect_seq) {
 	// if the existing connection successfully opened, and/or
-	// subsequently went to standby, then the peer should bump
+	// subsequently went to standby/closed, then the peer should bump
 	// their connect_seq and retry: this is not a connection race
 	// we need to resolve here.
 	if (existing->state == STATE_OPEN ||
-	    existing->state == STATE_STANDBY) {
+	    existing->state == STATE_STANDBY ||
+	    existing->state == STATE_CLOSED) {
 	  ldout(msgr->cct,10) << "accept connection race, existing " << existing
 			      << ".cseq " << existing->connect_seq
 			      << " == " << connect.connect_seq
```

tchaikov:
i think we should send the RETRY_SESSION message instead of the RETRY_GLOBAL message. so "retry_session:" logic can be reused.

gregsfortytwo:
This is still in my queue for deeper review and I'm planning to go through a bunch of PRs today so I hope to reach it.

But my instinct without digging too far into it, is that this is too broad and violates our mark_down rules. If we've got potential failures during connection setup that result in this state, we probably need to handle them there. Don't take that as gospel, but off-hand I'm pretty confident. :)

src/msg/simple/Pipe.cc
yuyuyu101: you mean line 474?

tchaikov: very sorry, i missed it!

src/msg/simple/Pipe.cc
yuyuyu101: https://github.com/ceph/ceph/commit/a640ca93307e92f8032881e35fdd74b60f0091d0

retry_session will cause side effect we may not want to have

yuyuyu101:
@gregsfortytwo yes, I agree these changes like a dirty fix. The main problem is I don't have brave to refactor too much in Pipe, it's really a adventure... If we want to clean these unexpected state changes, the lock usage need to be refactor but it is so disordered now!

gregsfortytwo:
Okay, I went through this more carefully and updated http://tracker.ceph.com/issues/14266. In short, I don't see how we can get into this situation at all without the other side misbehaving, and the proposed patch is definitely too broad a fix to be safe. :( So I'm closing this PR.

@yuyuyu101, I'm a bit concerned that the similar hunk in AsyncMessenger is almost certainly unsafe in the same ways. What I wonder is if you added it because you saw this more frequently in AsyncMessenger — if so, there might be some extra clues for us?

PR #8339, opened 2016-03-28
make openssl easier to use.

mdw-at-linuxbox:
Augment configuration so it's easy to support http & https in one civetweb frontend.

Get rid of an incorrect warning when using openssl.

And document the thing so people can use it.


yehudasa:
@mdw-at-linuxbox this PR doesn't deal with openssl linking issue, is there another fix for that?

mdw-at-linuxbox:
The ssl linking stuff is already pushed onto jewel, so that's all done.

djzort:
+1 to include this asap

mdw-at-linuxbox:
I created a new pull request (10335) to replace this one - I have a newer branch that is updated against a more current master, and knows about sonames for libssl.so | libcrypto.so .

PR #8340, opened 2016-03-28
build/ops: libgoogle-perftools-dev not on arm64 14.04.1

dachary:
Signed-off-by: Loic Dachary loic@dachary.org


liewegas:
@dachary I think we can close this now right?  Since we don't build arm on 14.04 at all?

dachary:
ok

PR #8346, opened 2016-03-29
msg/async/Event: replace ceph_clock_now with coarse_mono_clock

yuyuyu101:
reduce cpu utilization on real clock latency

### DIFF
--- a/src/msg/async/Event.cc
+++ b/src/msg/async/Event.cc
@@ -229,20 +229,7 @@ uint64_t EventCenter::create_time_event(uint64_t microseconds, EventCallbackRef 
 
   ldout(cct, 10) << __func__ << " id=" << id << " trigger after " << microseconds << "us"<< dendl;
   EventCenter::TimeEvent event;
-  utime_t expire;
-  struct timeval tv;
-
-  if (microseconds < 5) {
-    tv.tv_sec = 0;
-    tv.tv_usec = microseconds;
-  } else {
-    expire = ceph_clock_now(cct);
-    expire.copy_to_timeval(&tv);
-    tv.tv_sec += microseconds / 1000000;
-    tv.tv_usec += microseconds % 1000000;
-  }
-  expire.set_from_timeval(&tv);
-
+  clock_type::time_point expire = clock_type::now() + std::chrono::microseconds(microseconds);
   event.id = id;
   event.time_cb = ctxt;
   time_events[expire].push_back(event);
@@ -260,8 +247,7 @@ void EventCenter::delete_time_event(uint64_t id)
   if (id >= time_event_next_id)
     return ;
 
-  for (map<utime_t, list<TimeEvent> >::iterator it = time_events.begin();
-       it != time_events.end(); ++it) {
+  for (auto it = time_events.begin(); it != time_events.end(); ++it) {
     for (list<TimeEvent>::iterator j = it->second.begin();
          j != it->second.end(); ++j) {
       if (j->id == id) {
@@ -290,9 +276,8 @@ void EventCenter::wakeup()
 int EventCenter::process_time_events()
 {
   int processed = 0;
-  time_t now = time(NULL);
-  utime_t cur = ceph_clock_now(cct);
-  ldout(cct, 10) << __func__ << " cur time is " << cur << dendl;
+  clock_type::time_point now = clock_type::now();
+  ldout(cct, 10) << __func__ << " cur time is " << now << dendl;
 
   time_lock.Lock();
   /* If the system clock is moved to the future, and then set back to the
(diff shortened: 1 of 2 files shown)


src/msg/async/Event.cc
```diff
@@ -260,8 +247,7 @@ void EventCenter::delete_time_event(uint64_t id)
   if (id >= time_event_next_id)
     return ;
 
-  for (map<utime_t, list<TimeEvent> >::iterator it = time_events.begin();
-       it != time_events.end(); ++it) {
+  for (auto it = time_events.begin(); it != time_events.end(); ++it) {
```
tchaikov: better off using `auto&` to avoid the copying. might want to consider using the range-for statement instead, see http://www.stroustrup.com/C++11FAQ.html#for.

src/msg/async/Event.cc
```diff
@@ -341,39 +322,33 @@ int EventCenter::process_events(int timeout_microseconds)
...
-      tv.tv_usec += timeout_microseconds % 1000000;
-    }
-    shortest.set_from_timeval(&tv);
+    clock_type::time_point shortest;
+    shortest = now + std::chrono::microseconds(timeout_microseconds); 
 
     Mutex::Locker l(time_lock);
-    map<utime_t, list<TimeEvent> >::iterator it = time_events.begin();
+    auto it = time_events.begin();
     if (it != time_events.end() && shortest >= it->first) {
```
tchaikov: [OT] why would we update `shortest` if `it ->first == shortest`?

src/msg/async/Event.h
```diff
@@ -85,6 +86,7 @@ class EventDriver {
  * EventCenter maintain a set of file descriptor and handle registered events.
  */
 class EventCenter {
+  using clock_type = ceph::coarse_real_clock;
```
tchaikov: i think what we really want here is `ceph::coarse_mono_clock`.

src/msg/async/Event.cc
yuyuyu101: because I want to use map::erase(iterator)

tchaikov: then we can not update `it` here.

tchaikov: sorry, scratch that. didn't notice that you returned right away.

tchaikov:
lgtm

markhpc:
Do we have any idea what the effects of this are yet?

yuyuyu101:
@markhpc yep, because process_events is in fast path, in benchmark case, clock_gettime will occur 7-8% cpu util. I don't dive into the impl of coarse clock, I guess it will return last timestamp if the gap is so short. From the result, the cpu util on clock_gettime is missing in "perf top" with this pr

tchaikov:
@markhpc the `clock_gettime(CLOCK_REALTIME_COARSE)` does not read the time from hardware, it just read the last tick. so it's way faster. see http://lwn.net/Articles/342018/

gregsfortytwo:
I think the real question is "are we sure coarse clocks are close enough?" I'm not sure how these time events are used in the AsyncMessenger — if they're linux ticks that's accurate enough for many purposes, but probably not for eg our op tracking.

wjwithagen:
On 30-3-2016 20:07, Gregory Farnum wrote:
> I think the real question is "are we sure coarse clocks are close
> enough?" I'm not sure how these time events are used in the
> AsyncMessenger — if they're linux ticks that's accurate enough for many
> purposes, but probably not for eg our op tracking.

If you look at the real implementation of gettime, than it runs as
	COARSE on Linux
	FAST on FreeBSD
but on other platforms it falls back to regular gettime.

So there it will kill performance.

As does getting a better resolution, atleast on FreeBSD the diff in
calltime is major.

mattbenjamin:
What Greg said, basically.  We should be using the coarse/fast clocks whenever their resolution is sufficient, when in a fast path (as here).  The supported resolution of any given clock is given by clock_getres(), so if we want to be systematic and define a clear set of guidelines for when different clocks should be used, we can do so.  What do you think @adamemerson ?

adamemerson:
Looking at it, time events seem to be scheduled with a microsecond count (relative to the present). So using coarse_mono_clock would be my preference and taking ceph::timespan instead of a number of microseconds.
It doesn't look to have any particular resolution requirements. clock_getres isn't bad but I'd rather use it compile-time tests rather than at runtime.
Also, I'm not sure if you want to use a `std::map` whose value is a `std::list` in the fast path. It might be better to use `boost::intrusive::set` with the hook in TimeEvent.
Also, it might be better to use `std::function<void(int)>` may be preferable to the EventCallback abstract base class.

src/msg/async/Event.cc
```diff
@@ -303,18 +288,14 @@ int EventCenter::process_time_events()
    * events to be processed ASAP when this happens: the idea is that
    * processing events earlier is less dangerous than delaying them
    * indefinitely, and practice suggests it is. */
-  bool clock_skewed = false;
-  if (now < last_time) {
-    clock_skewed = true;
-  }
+  bool clock_skewed = now < last_time;
```
tchaikov: if we are using coarse_mono_clock, then we don't need to worry about clock_skewed?

src/msg/async/Event.h
```diff
@@ -85,6 +86,7 @@ class EventDriver {
  * EventCenter maintain a set of file descriptor and handle registered events.
  */
 class EventCenter {
+  using clock_type = ceph::coarse_mono_clock;
```
tchaikov: thought you wanted to use `coarse_real_clock`... so we can update the commit message accordingly?

src/msg/async/Event.cc
yuyuyu101: Hmm, I'm considering remove this

liewegas:
So.. the coarse clock is ~1ms resolution (1 tick) on Linux. Is that sufficient?

This passed a rados run, but not sure what the verdict is.

yuyuyu101:
ms level is fine to this usage. If high resolution needed, we can add a separated interface later.

PR #8347, opened 2016-03-29
msg/async: AF_UNIX socket support

branch-predictor:
This changeset adds AF_UNIX sockets to Async Messenger for improved
performance and reduction of IP network stack load (which include
any firewall rules, if necessary) of communication between daemons
and clients on same host. Since it's quite common to have multiple
OSDs on single host, there's no reason to have them talk through IP
protocol even if Linux kernel is smart enough to not pipe data
through complete network stack in that case.
Improves max throughput of Async messenger.
New config knobs added:

  - ms async use local socket (bool, default = false)
    Enable or disable AF_UNIX sockets usage
  - ms async require local socket (bool, default = false)
    If true, any failure to set up listening local socket is considered
    fatal.
  - ms async sndbuf (integer, default = 0/system default)
    Allows setting the size of send buffer of AF_UNIX socket, on Linux
    3.10.42 it defaults to around 208KB.

### DIFF
--- a/src/common/config_opts.h
+++ b/src/common/config_opts.h
@@ -205,6 +205,10 @@ OPTION(ms_async_set_affinity, OPT_BOOL, true)
 OPTION(ms_async_affinity_cores, OPT_STR, "")
 OPTION(ms_async_send_inline, OPT_BOOL, true)
 
+OPTION(ms_async_use_local_socket, OPT_BOOL, false) // should async msgr use AF_UNIX sockets for communication on localhost?
+OPTION(ms_async_require_local_socket, OPT_BOOL, false) // should local socket bind error be fatal?
+OPTION(ms_async_sndbuf, OPT_INT, 0) // local socket send buffer size
+
 OPTION(inject_early_sigterm, OPT_BOOL, false)
 
 OPTION(mon_data, OPT_STR, "/var/lib/ceph/mon/$cluster-$id")
--- a/src/include/types.h
+++ b/src/include/types.h
@@ -21,6 +21,7 @@
 #include "uuid.h"
 
 #include <netinet/in.h>
+#include <sys/un.h>
 #include <fcntl.h>
 #include <string.h>
 
--- a/src/msg/async/AsyncConnection.cc
+++ b/src/msg/async/AsyncConnection.cc
@@ -32,7 +32,7 @@
 #undef dout_prefix
 #define dout_prefix _conn_prefix(_dout)
 ostream& AsyncConnection::_conn_prefix(std::ostream *_dout) {
-  return *_dout << "-- " << async_msgr->get_myinst().addr << " >> " << peer_addr << " conn(" << this
+  return *_dout << "-- " << async_msgr->get_myinst().addr << " >> " << established_peer_addr << " conn(" << this
                 << " sd=" << sd << " :" << port
                 << " s=" << get_state_name(state)
                 << " pgs=" << peer_global_seq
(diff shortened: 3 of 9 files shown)


branch-predictor:
Some perf numbers (cached reads):

4 parallel IOs:


16 parallel IOs:



128 parallel IOs:


local_512 is an local AF_UNIX socket with buffer size set to 512kb, capped by OS to around 448kb.

src/common/config_opts.h
```diff
@@ -205,6 +205,10 @@ OPTION(ms_async_set_affinity, OPT_BOOL, true)
 OPTION(ms_async_affinity_cores, OPT_STR, "")
 OPTION(ms_async_send_inline, OPT_BOOL, true)
 
+OPTION(ms_async_use_local_socket, OPT_BOOL, false) // should async msgr use AF_UNIX sockets for communication on localhost?
+OPTION(ms_async_require_local_socket, OPT_BOOL, false) // should local socket bind error be fatal?
+OPTION(ms_async_sndbuf, OPT_INT, 0) // local socket send buffer size
```
avnerbh: I would add at the end of the comment:
`// value of 0 indicates system default`

mattbenjamin:
These are not quite the graphs I might have naively expected   Do you have any sense of any possible benefit in small-message latency or maybe system memory utilization?

branch-predictor:
@mattbenjamin It's actually better than I expected. For small reads and writes, my small PoC shown that Unix socket were slower than TCP sockets, but it's not the case here (which may be due to lower latency). Larger reads and writes are improved by less than in my PoC, but then, Async messenger does far more than my PoC and having performance improved by 1G/s instead of ~2,5G/s doesn't sound that bad. I haven't measured memory utilization yet.

liewegas:
I think it's more likely that we'll be combining OSDs into a single process (to share caches, messenger connections, etc).  I don't think this is worth the complexity.

yuyuyu101:
hmm, if federated feature could be impl, it more make sense.... @branch-predictor what do you think?

branch-predictor:
On Thu, Apr 28, 2016 at 06:58:44PM -0700, Sage Weil wrote:
> I think it's more likely that we'll be combining OSDs into a single process (to share caches, messenger connections, etc).  I don't think this is worth the complexity.

It's still less complex than combined OSDs, and can be disabled at any time.
Also, considering Ceph's fail-fast/fail-hard nature, I assume that many users won't
go that (combined OSDs) way.

-- 
Piotr Dałek
branch@predictor.org.pl
http://blog.predictor.org.pl

branch-predictor:
On Thu, Apr 28, 2016 at 07:57:57PM -0700, Haomai Wang wrote:
> hmm, if federated feature could be impl, it more make sense.... @branch-predictor what do you think?

What is "federated feature"? If you mean combined OSDs, then no, since kn
that case we don't need kernel for IPC. 

-- 
Piotr Dałek
branch@predictor.org.pl
http://blog.predictor.org.pl

yuyuyu101:
"If you mean combined OSDs, then no, since kn that case we don't need kernel for IPC."

what's the meaning for later part?

branch-predictor:
On Sat, Apr 30, 2016 at 01:23:01AM -0700, Haomai Wang wrote:
> "If you mean combined OSDs, then no, since kn that case we don't need kernel for IPC."
> 
> what's the meaning for later part?

For OSDs in separate processes, we need kernel for inter-process
communication (IPC), either for loopback or unix sockets. In case of
single process hosting multiple OSDs, shared memory (passing pointers)
is enough, no kernel arbitration is required.

-- 
Piotr Dałek
branch@predictor.org.pl
http://blog.predictor.org.pl

liewegas:
We think this is not worth the complexity to maintain at this time, given how rarely it comes up.

PR #8348, opened 2016-03-29
osd: fix failure report handling during ms_handle_connect()

xiexingguo:
On connecting to a new monitor, we will resend everything
including the osd failure reports previously sent.
To realize this, we call requeue_failures() to transfer inflight
failure reports from failure_pending to failure_queue first, and
then call send_failures() to do the real delivery job.

The problem here is that the send_failures() never sends a
failure report again if it successfully detects that the doomed osd
is already in the failure_pending set, which is necessary as we don't
want to report monitor of the same osd failure twice in normal case.

This pr solves the above problem by erasing the record from failure_pending
set simultaneously during the requeue_failures() process. So the
succeeding call to send_failures() can resend the failure reports correctly.


xiexingguo:
@liewegas More fixes and cleanups.

PR #8350, opened 2016-03-29
mon: do not alloc monmap on heap in Elector::dispatch()

songbaisen:
mon: elector dispatch do not alloc memory on monmap

### DIFF
--- a/src/mon/Elector.cc
+++ b/src/mon/Elector.cc
@@ -430,10 +430,10 @@ void Elector::dispatch(MonOpRequestRef op)
 	return;
       }
 
-      MonMap *peermap = new MonMap;
-      peermap->decode(em->monmap_bl);
-      if (peermap->epoch > mon->monmap->epoch) {
-	dout(0) << em->get_source_inst() << " has newer monmap epoch " << peermap->epoch
+      MonMap peermap;
+      peermap.decode(em->monmap_bl);
+      if (peermap.epoch > mon->monmap->epoch) {
+	dout(0) << em->get_source_inst() << " has newer monmap epoch " << peermap.epoch
 		<< " > my epoch " << mon->monmap->epoch 
 		<< ", taking it"
 		<< dendl;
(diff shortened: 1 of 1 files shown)


src/mon/Elector.cc
```diff
@@ -453,7 +452,6 @@ void Elector::dispatch(MonOpRequestRef op)
 		<< " < my epoch " << mon->monmap->epoch 
 		<< dendl;
       } 
-      delete peermap;
```
branch-predictor: Do we really need to construct MonMap instance by hand?

songbaisen:
@branch-predictor  Good eyes man..For performance here is really does not need to new memory every time.

src/mon/Elector.h
```diff
@@ -44,7 +45,11 @@ class Elector {
    * The Monitor instance associated with this class.
    */
   Monitor *mon;
-
+  /**
+   * The instance of the peer monitor map
+   *
+   */ 
+  MonMap peermap;
```
tchaikov: why would you want to keep the peermap around as a member variable?

songbaisen: @tchaikov  consider the  branch-predictor  idea.do not need to  new and free every time.

src/mon/Elector.cc
```diff
@@ -430,10 +430,9 @@ void Elector::dispatch(MonOpRequestRef op)
 	return;
       }
 
-      MonMap *peermap = new MonMap;
-      peermap->decode(em->monmap_bl);
-      if (peermap->epoch > mon->monmap->epoch) {
-	dout(0) << em->get_source_inst() << " has newer monmap epoch " << peermap->epoch
+      peermap.decode(em->monmap_bl);
```
tchaikov: could just use a local variable? guess that's what @branch-predictor means.

songbaisen:
@tchaikov  Thank you sir! Got it.

yangdongsheng:
@songbaisen Sorry, but I would ask, what's the benefit from this change? what I can see here is just moving the peermap location from heap to stack, what's the difference? Could you give a more explanation? thanx

songbaisen:
@yangdongsheng  Hello.Nice to meet you.Thank you for your reply.Here is not very a lot difference.Just for the safe in future.If any body change the code before delete,we may have a risk of memory leak.For prevent that happen change it.

branch-predictor:
@tchaikov @songbaisen yep, I meant having an MonMap instance as local variable. This way we're leak-free *and* we're one heap (de)allocation less.

athanatos:
Passed tests: sjust@teuthology:/a/samuelj-2016-05-16_20:02:21-rados-wip-sam-testing-distro-basic-smithi

@tchaikov Does this look ok?

PR #8351, opened 2016-03-29
cmake: Build cython modules and change paths to bin/, lib/

jcsp:
This is derived from https://github.com/ceph/ceph/pull/7908 and https://github.com/ceph/ceph/pull/7912

Ali's original change included changing build paths for other binaries, so this branch includes the changes elsewhere to accomodate that.  There is also a change in ceph-qa-suite for the paths https://github.com/ceph/ceph-qa-suite/pull/916

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,6 +48,11 @@ include(CheckIncludeFiles)
 include(CheckIncludeFileCXX)
 include(CheckFunctionExists)
 
+#put all the libs and binaries in one place
+set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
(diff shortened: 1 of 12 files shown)


oritwas:
don't we need to fix stop.sh?

jcsp:
@oritwas you're right: I forget about stop.sh because I just use the killall path of it

jcsp:
Updated for stop.sh

cbodley:
thanks for pulling these patches out! i was able to build cleanly, and vstart/stop worked from the cmake build directory as expected :+1:

jdurgin:
The ceph.in path changes broke the automake python path in a dev setup. fixed in https://github.com/ceph/ceph/pull/8360

PR #8352, opened 2016-03-29
rbd-mirror: fix missing increment for iterators

runsisi:
rbd-mirror: fix missing increment of iterators in `Replayer::set_sources`.

Signed-off-by: runsisi <runsisi@hust.edu.cn>

### DIFF
--- a/src/tools/rbd_mirror/Replayer.cc
+++ b/src/tools/rbd_mirror/Replayer.cc
@@ -274,11 +274,15 @@ void Replayer::set_sources(const PoolImageIds &pool_image_ids)
       for (auto images_it = pool_images.begin();
 	   images_it != pool_images.end();) {
 	if (stop_image_replayer(images_it->second)) {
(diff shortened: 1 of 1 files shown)


xiexingguo:
This change looks good. @dillaman  What do you think?

src/tools/rbd_mirror/Replayer.cc
```diff
@@ -273,10 +273,14 @@ void Replayer::set_sources(const map<int64_t, set<string> > &images)
 	   images_it != pool_images.end();) {
 	if (stop_image_replayer(images_it->second)) {
 	  pool_images.erase(images_it++);
```
dillaman: While fixing this, can we also switch this (and below) to something like `images_it = pool_images.erase(images_it)`?

runsisi:
@dillaman ok, fixed it, thx~

xiexingguo:
LGTM

PR #8353, opened 2016-03-29
client: pass 'newly issued caps' to Client::check_cap_issue()

ukernel:
Client::check_cap_issue() expects caller to pass 'newly issued caps'
to it. But in Client::check_cap_issue(), varible 'issued' is caps
client alread has.

Fixes: http://tracker.ceph.com/issues/15303

### DIFF
--- a/src/client/Client.cc
+++ b/src/client/Client.cc
@@ -4807,7 +4807,7 @@ void Client::handle_cap_grant(MetaSession *session, Inode *in, Cap *cap, MClient
   if (m->get_op() == CEPH_CAP_OP_IMPORT && m->get_wanted() != wanted)
     check = true;
 
-  check_cap_issue(in, cap, issued);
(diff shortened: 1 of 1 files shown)


gregsfortytwo:
@jcsp, please test with Ramana.

jcsp:
This does indeed make the the reproducer test pass

gregsfortytwo:
@ukernel, did you check the other caller of check_cap_issue()? The call chain around add_update_caps() looks mostly okay but isn't entirely obvious.

I also wonder if we should change the parameter name on check_cap_issue() from "issued" to "newly_issued" or something, in case of future users. (Although it looks like this error originates from when Sage created the function.)

ukernel:
I checked the other caller, the usage should be OK

gregsfortytwo:

PR #8357, opened 2016-03-29
osd: update crush_location from ceph-osd on startup

liewegas:
This is (1) just cleaner and better, and (2) it means that the crush weight matches ObjectStore::statfs's value, which means that for bluestore we get an accurate weight (whereas df does not give you the right value)

### DIFF
--- a/src/common/config_opts.h
+++ b/src/common/config_opts.h
@@ -614,7 +614,7 @@ OPTION(osd_pgp_bits, OPT_INT, 6)  // bits per osd
 OPTION(osd_crush_chooseleaf_type, OPT_INT, 1) // 1 = host
 OPTION(osd_pool_use_gmt_hitset, OPT_BOOL, true) // try to use gmt for hitset archive names if all osds in cluster support it.
 OPTION(osd_crush_update_on_start, OPT_BOOL, true)
-OPTION(osd_crush_initial_weight, OPT_DOUBLE, 0) // the initial weight is for newly added osds.
+OPTION(osd_crush_initial_weight, OPT_DOUBLE, -1) // if >=0, the initial weight is for newly added osds.
 OPTION(osd_pool_default_crush_rule, OPT_INT, -1) // deprecated for osd_pool_default_crush_replicated_ruleset
 OPTION(osd_pool_default_crush_replicated_ruleset, OPT_INT, CEPH_DEFAULT_CRUSH_REPLICATED_RULESET)
 OPTION(osd_pool_erasure_code_stripe_width, OPT_U32, OSD_POOL_ERASURE_CODE_STRIPE_WIDTH) // in bytes
--- a/src/common/config_opts.h
+++ b/src/common/config_opts.h
@@ -417,6 +417,8 @@ OPTION(client_use_faked_inos, OPT_BOOL, false)
 OPTION(client_mds_namespace, OPT_INT, -1)
 
 OPTION(crush_location, OPT_STR, "")       // whitespace-separated list of key=value pairs describing crush location
+OPTION(crush_location_hook, OPT_STR, "")
+OPTION(crush_location_hook_timeout, OPT_INT, 10)
 
 OPTION(objecter_tick_interval, OPT_DOUBLE, 5.0)
 OPTION(objecter_timeout, OPT_DOUBLE, 10.0)    // before we ask for a map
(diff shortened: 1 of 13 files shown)


liewegas:
This also fixes an annoying startup timeout issue where the crush update might timeout, and we have to decide whether to not start the osd at all (that's no fun) or start it with a potentially incorrect location.  Doing it from the osd itself lets us block indefinitely.  This was pretty easy to hit if, say, the whole cluster is rebooting and the osds come up well before the mons form a quorum.

src/crush/CrushLocation.cc
```diff
@@ -0,0 +1,105 @@
...
+    return update_from_hook();
+  }
+
+  // start with a sane default
+  char hostname[HOST_NAME_MAX + 1];
+  int r = gethostname(hostname, sizeof(hostname)-1);
+  if (r < 0)
+    strcpy(hostname, "unknown_host");
+  std::lock_guard<std::mutex> l(lock);
+  loc.clear();
```
tchaikov: this overwrites the crush location just collected from config file??

src/common/config_opts.h
```diff
@@ -610,6 +610,7 @@ OPTION(osd_crush_chooseleaf_type, OPT_INT, 1) // 1 = host
 // This parameter is not consumed by ceph C code but the upstart scripts.
 // OPTION(osd_crush_initial_weight, OPT_DOUBLE, 0) // the initial weight is for newly added osds.
```
tchaikov: can also remove this comment. it was added to document the the behaviour in  src/ceph-osd-prestart.sh

src/osd/OSD.cc
```diff
@@ -2719,6 +2723,57 @@ int OSD::shutdown()
...
+    string("\"args\": [");
+  for (multimap<string,string>::iterator p = loc.begin(); p != loc.end(); ++p) {
+    if (p != loc.begin())
+      cmd += ", ";
+    cmd += "\"" + p->first + "=" + p->second + "\"";
+  }
+  cmd += "]}";
+
+  dout(10) << __func__ << " cmd: " << cmd << dendl;
+  vector<string> vcmd(1);
```
tchaikov: nit, could just put
```c++
vector<string> cmd{cmd};
```

src/crush/CrushLocation.cc
liewegas: see line 87.. we use the config value if it is specified.  otherwise, we try teh hook.  otherwise, we have a sane default.

liewegas:
@tchaikov updated

liewegas:
I don't really like the osd_crush_initial_weight == 0 means not to use it, though.  I think we should make negative mean 'use the real size', and 0 mean 'weight it to 0'.

src/crush/CrushLocation.cc
tchaikov: oh, got it. we **return** right away.

tchaikov:
yeah, should document this change.

lgtm!

smithfarm:
@liewegas There is a request [1] to backport this to jewel or, failing that, to add a timeout to ceph-osd-prestart.sh in jewel. What do you think?

[1] http://tracker.ceph.com/issues/15213#note-3

smithfarm:
Jewel backport PR: #9256

liewegas:
I don't think we need to backport this one.  It redoes the way that 
startup is handled.. I'd rather wait for kraken.

liewegas:
I cannot for the life of me figure out why we made this a multimap instead of a map.  @tchaikov do you have any idea?  Why would we have a location with, say, two racks or hosts specified?

https://github.com/ceph/ceph/blob/master/src/crush/CrushLocation.h#L15

tchaikov:
i think it's allowed that an item (bucket) occurs twice in the crushmap. guess that's why we use a `multimap` instead of a `map` here. see https://github.com/ceph/ceph/pull/988#r8540960

but probably this feature is not yet fully supported

1. the `multimap` will be degraded to a plain `map` when osd daemon boots, OSD::init() -> OSD::update_crush_location() -> OSDMonitor::prepare_command_impl("crush create-or-move") -> CrushWrapper::parse_loc_map(). so even an OSD is assigned to two buckets in a location map, only the last one will be used when it is created.
1. `get_immediate_parent()`

PR #8358, opened 2016-03-29
releases: what is merged where and when ?

dachary:

PR #8365, opened 2016-03-30
doc:Wrong count for 10.0.4 desc

scienceluo:
V10.0.4 was not "the fourth" but "the fifth" development release before Jewel.

Signed-off-by : luo kexue luo.kexue@zte.com.cn

### DIFF
--- a/doc/release-notes.rst
+++ b/doc/release-notes.rst
@@ -5,7 +5,7 @@
 v10.0.4
 =======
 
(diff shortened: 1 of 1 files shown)


xiexingguo:
Will be recovered by https://github.com/ceph/ceph/pull/8358/files#diff-d31d1b025d45818c51583cfa5f76c9c9R1953, close yours.

PR #8370, opened 2016-03-30
osd/PG: cleanup: delete find_best_info again

jack-changtao:
After called find_best_info find a auth_pg_shard, it must be not incomplete,
so it will not enter into logic of find_best_info again in complete_infos.

Signed-off-by: Tao Chang <changtao@hihuron.com>

### DIFF
--- a/src/osd/PG.cc
+++ b/src/osd/PG.cc
@@ -1325,26 +1325,7 @@ bool PG::choose_acting(pg_shard_t &auth_log_shard_id, bool *history_les_bound)
     return false;
   }
 
-  if ((up.size() &&
-      !all_info.find(up_primary)->second.is_incomplete() &&
-      all_info.find(up_primary)->second.last_update >=
-       auth_log_shard->second.log_tail) &&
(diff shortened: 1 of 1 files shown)


src/osd/PG.cc
```diff
@@ -1325,26 +1325,6 @@ bool PG::choose_acting(pg_shard_t &auth_log_shard_id, bool *history_les_bound)
     return false;
   }
```
athanatos: @jack-changtao Can you add an assert(!auth_log_shard->second.is_incomplete()); ?  Otherwise, I think this is right.

jack-changtao:
@athanatos 
Have changed it, please review, thanks very much!

athanatos:
I think this is probably right, marked needs-qa.

athanatos:
Heh, this turns out to be a fix for http://tracker.ceph.com/issues/13967 adding to wip-sam-testing.

athanatos:
sjust@teuthology:/a/samuelj-2016-06-21_10:29:56-rados-wip-sam-testing-distro-basic-smithi

PR #8371, opened 2016-03-30
cls_rbd: fix -EEXIST checking in cls::rbd::image_set

runsisi:
`read_key` may return other negative codes(e.g. `-EIO`) other than  `-ENOENT`, so we'd better not to return  `-EEXIST` blindly to our caller.

Signed-off-by: runsisi <runsisi@hust.edu.cn>

### DIFF
--- a/src/cls/rbd/cls_rbd.cc
+++ b/src/cls/rbd/cls_rbd.cc
@@ -3091,8 +3091,12 @@ int image_set(cls_method_context_t hctx, const string &image_id,
     std::string global_id_key = global_key(mirror_image.global_image_id);
     std::string image_id;
     r = read_key(hctx, global_id_key, &image_id);
-    if (r != -ENOENT) {
+    if (r >= 0) {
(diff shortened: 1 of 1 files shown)


src/cls/rbd/cls_rbd.cc
```diff
@@ -3087,7 +3087,13 @@ int image_set(cls_method_context_t hctx, const string &image_id,
     std::string global_id_key = global_key(mirror_image.global_image_id);
     std::string image_id;
     r = read_key(hctx, global_id_key, &image_id);
-    if (r != -ENOENT) {
+    if (r < 0) {
```
dillaman: Instead of the nested 'if' statements, how about something like:

```
if (r >= 0) {
  return -EEXIST;
} else if (r != -ENOENT) {
  CLR_ERR(...)
  return r;
}
```

runsisi:
@dillaman ok, fixed as suggested, thank u very much.

PR #8372, opened 2016-03-30
qa: update rest test cephfs calls

jcsp:
Since the ceph task was already creating filesystems
during setup, I presume that these calls only ever
worked when they were using the same names as
the existing filesystem.

Fixes: http://tracker.ceph.com/issues/15309

### DIFF
--- a/qa/workunits/rest/test.py
+++ b/qa/workunits/rest/test.py
@@ -166,9 +166,9 @@ if __name__ == '__main__':
     r = expect('mds/dump.xml', 'GET', 200, 'xml')
     assert(r.tree.find('output/mdsmap/created') is not None)
 
-    expect('osd/pool/create?pg_num=1&pool=fsmetadata', 'PUT', 200, '')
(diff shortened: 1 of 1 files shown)


liewegas:
2016-03-30T15:06:45.122 INFO:tasks.workunit.client.0.smithi042.stderr:Traceback (most recent call last):
2016-03-30T15:06:45.122 INFO:tasks.workunit.client.0.smithi042.stderr:  File "/home/ubuntu/cephtest/workunit.client.0/rest/test.py", line 199, in <module>
2016-03-30T15:06:45.122 INFO:tasks.workunit.client.0.smithi042.stderr:    assert('info' in r.myjson['output']['mdsmap'])
2016-03-30T15:06:45.122 INFO:tasks.workunit.client.0.smithi042.stderr:KeyError: 'mdsmap'


from http://pulpito.ceph.com/sage-2016-03-30_13:23:09-rados-wip-sage-testing-distro-basic-smithi/98173

PR #8373, opened 2016-03-30
ceph-detect-init/run-tox.sh: FreeBSD: No init detect

wjwithagen:
- Freebsd has /etc/rc.d with scripts to start daemons at boottime. 
   Test for that are currently not integrated with Ceph tests

### DIFF
--- a/src/ceph-detect-init/run-tox.sh
+++ b/src/ceph-detect-init/run-tox.sh
@@ -17,6 +17,11 @@
 # GNU Library Public License for more details.
 #
 
+if [ x"`uname`"x = xFreeBSDx ]; then
+    echo FreeBSD init system has not been integrated.
+    exit 0
+fi
+
 # run from the ceph-detect-init directory or from its parent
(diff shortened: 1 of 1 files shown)


dachary:
Could you instead
<pre>
if test "`uname`"  = FreeBSD ; then
 echo warning
 exit 0
fi
</pre>
it's a detail but makes it clearer that it's a FreeBSD only thing.

wjwithagen:
On 30-3-2016 13:37, Loic Dachary wrote:
> Could you instead
> 
> if test "`uname`"  = FreeBSD ; then
>  echo warning
>  exit 0
> fi
> 
> it's a detail but makes it clearer that it's a FreeBSD only thing.

Sure, I liked the case-construct because it would be easier to extend to
other OSes, since I also ran into AIX, sun, Darwin, ... defines.
And when they want to run the tests it is easier to rinse-repeat and
extend. But for just FreeBSD it is fine.

dachary:
<pre>
</pre>

PR #8375, opened 2016-03-30
librbd: disable image mirroring when image is removed

rjfd:
Fixes: #15265

Ticket url: http://tracker.ceph.com/issues/15265

### DIFF
--- a/src/librbd/internal.cc
+++ b/src/librbd/internal.cc
@@ -280,8 +280,20 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force) {
   std::set<cls::journal::Client> clients;
   std::string header_oid;
 
+  int r = cls_client::mirror_image_get(&ictx->md_ctx, ictx->id,
+      &mirror_image_internal);
+  if (r < 0 && r != -ENOENT) {
+    lderr(cct) << "cannot disable mirroring: " << cpp_strerror(r) << dendl;
+    return r;
+  } else if (r == -ENOENT) {
+    // mirroring is not enabled for this image
+    ldout(cct, 20) << "ignoring disable command: mirroring is not enabled "
+      "for this image" << dendl;
+    return 0;
+  }
+
   bool is_primary;
-  int r = Journal<>::is_tag_owner(ictx, &is_primary);
+  r = Journal<>::is_tag_owner(ictx, &is_primary);
   if (r < 0) {
     lderr(cct) << "cannot disable mirroring: failed to check tag ownership: "
       << cpp_strerror(r) << dendl;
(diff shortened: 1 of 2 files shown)


src/librbd/internal.cc
```diff
@@ -1943,6 +1943,19 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force) {
 	return r;
       }
 
+      if (!old_format) {
+        r = mirror_image_disable_internal(ictx, true);
```
dillaman: We probably want `false` for the force flag.  If the image isn't primary, it is probably being replicated.   If a delete is truly desired, promote it first or delete the primary image and let the deletion propagate over to the peer cluster.

src/librbd/internal.cc
```diff
@@ -1943,6 +1943,19 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force) {
...
+        if (r < 0) {
+          lderr(cct) << "error disabling image mirroring: " << cpp_strerror(r)
+                     << dendl;
+          ictx->owner_lock.put_read();
+          ictx->state->close();
+          return r;
+        }
+
+        // TODO: http://tracker.ceph.com/issues/14421
```
dillaman: I think rbd-mirror will be responsible for propagating the deletion (it should also delete non-primary images if mirroring is disabled).  When the peer rbd-mirror daemon notices that it has a non-primary image owned by a cluster that no longer lists the image in its mirroring directory, it can "force" delete the local, non-primary image.

src/librbd/internal.cc
rjfd: I agree

src/librbd/internal.cc
rjfd: If I understood correctly, if we remove the image from the cluster that owns the image (primary image), we just execute the code above. Then, it's the rbd-mirror daemon responsibility to check if the image disappears from the mirroring directory and delete the non-primary image.

I'll remove the TODO comment.

dillaman: Correct -- we cannot assume that the local librbd has access to the remote peer clusters. We will let the remote rbd-mirror daemon "pull" the deletion.

rjfd:
Fixed commit

src/librbd/internal.cc
```diff
@@ -280,8 +280,21 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force) {
   std::set<cls::journal::Client> clients;
   std::string header_oid;
 
+  int r = cls_client::mirror_image_get(&ictx->md_ctx, ictx->id,
+      &mirror_image_internal);
+  if (r < 0 && r != -ENOENT) {
+    lderr(cct) << "cannot disable mirroring: " << cpp_strerror(r) << dendl;
+    return r;
+  }
+  else if (r == -ENOENT) {
```
jdurgin: nit: else should be on the same line as } and {

rjfd:
@jdurgin fixed :)

PR #8376, opened 2016-03-30
os/bluestore: Minor code simplification for read handler.

ifed01:
Replaces lower_bound with upper_bound call to locate blocks more precisely and to avoid the need for subsequent iterator repositioning.

### DIFF
--- a/src/os/bluestore/BlueStore.cc
+++ b/src/os/bluestore/BlueStore.cc
@@ -2577,28 +2577,18 @@ int BlueStore::_do_read(
 
   // loop over overlays and data fragments.  overlays take precedence.
   bend = o->onode.block_map.end();
-  bp = o->onode.block_map.lower_bound(offset);
+  bp = o->onode.block_map.upper_bound(offset);
   if (bp != o->onode.block_map.begin()) {
     --bp;
   }
   oend = o->onode.overlay_map.end();
-  op = o->onode.overlay_map.lower_bound(offset);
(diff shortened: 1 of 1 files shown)


src/os/bluestore/BlueStore.cc
```diff
@@ -2577,28 +2577,18 @@ int BlueStore::_do_read(
...
-      continue;
-    }
-    if (bp != bend && bp->first + bp->second.length <= offset) {
-      dout(30) << __func__ << " skip frag " << bp->first << " " << bp->second
-	       << dendl;
-      ++bp;
-      continue;
-    }
+    assert(op == oend || (op->first + op->second.length > offset));
+    assert(bp == bend || (bp->first + bp->second.length > offset));
```
xiexingguo: I doubt the above assert is right. 
```
bp = o->onode.block_map.upper_bound(offset);
   if (bp != o->onode.block_map.begin()) {
     --bp;
   }
```
Consider ```bp = o->onode.block_map.upper_bound(offset);``` the statement only, we may assert this way.
But after the  ```--bp;``` statement and if there is a gap in the block map, it is possible that we may still hit the``` bp != bend && bp->first + bp->second.length <= offset``` condition and this assert is going to fire under this case.

ifed01:
@xiexingguo,
yep, you are right. Thanks!

xiexingguo:
@ifed01 You are welcome:)

PR #8383, opened 2016-03-31
DNM: CephFS Lockouts

gregsfortytwo:
This PR disables the CephFS features we don't consider stable in time for Jewel.

Specifically, it strengthens the multi-FS warning and adds a lockout on multi-MDS clusters and
directory fragmentation.

Currently untested.

### DIFF
--- a/src/mds/MDSMap.cc
+++ b/src/mds/MDSMap.cc
@@ -134,6 +134,8 @@ void MDSMap::dump(Formatter *f) const
 {
   f->dump_int("epoch", epoch);
   f->dump_unsigned("flags", flags);
+  f->dump_unsigned("ever_allowed_features", ever_allowed_features);
+  f->dump_unsigned("explicitly_allowed_features", explicitly_allowed_features);
   f->dump_stream("created") << created;
   f->dump_stream("modified") << modified;
   f->dump_int("tableserver", tableserver);
@@ -547,7 +549,7 @@ void MDSMap::encode(bufferlist& bl, uint64_t features) const
   ::encode(cas_pool, bl);
 
   // kclient ignores everything from here
-  __u16 ev = 9;
+  __u16 ev = 10;
   ::encode(ev, bl);
   ::encode(compat, bl);
   ::encode(metadata_pool, bl);
@@ -560,8 +562,8 @@ void MDSMap::encode(bufferlist& bl, uint64_t features) const
   ::encode(failed, bl);
   ::encode(stopped, bl);
   ::encode(last_failure_osd_epoch, bl);
-  ::encode(ever_allowed_snaps, bl);
-  ::encode(explicitly_allowed_snaps, bl);
+  ::encode(ever_allowed_features, bl);
+  ::encode(explicitly_allowed_features, bl);
   ::encode(inline_data_enabled, bl);
   ::encode(enabled, bl);
   ::encode(fs_name, bl);
@@ -624,11 +626,20 @@ void MDSMap::decode(bufferlist::iterator& p)
   if (ev >= 4)
     ::decode(last_failure_osd_epoch, p);
   if (ev >= 6) {
-    ::decode(ever_allowed_snaps, p);
-    ::decode(explicitly_allowed_snaps, p);
+    if (ev < 10) {
+      // previously this was a bool about snaps, not a flag map
+      bool flag;
+      ::decode(flag, p);
+      legacy_mds_map.ever_allowed_features = flag ? CEPH_MDSMAP_ALLOW_SNAPS : 0;
+      ::decode(flag, p);
+      legacy_mds_map.explicitly_allowed_features = flag ? CEPH_MDSMAP_ALLOW_SNAPS : 0;
+    } else {
+      ::decode(ever_allowed_features, p);
+      ::decode(explicitly_allowed_features, p);
+    }
   } else {
-    ever_allowed_snaps = true;
-    explicitly_allowed_snaps = false;
+    ever_allowed_features = CEPH_MDSMAP_ALLOW_SNAPS;
+    explicitly_allowed_features = 0;
   }
   if (ev >= 7)
     ::decode(inline_data_enabled, p);
--- a/src/mds/MDSMap.cc
+++ b/src/mds/MDSMap.cc
@@ -630,16 +630,23 @@ void MDSMap::decode(bufferlist::iterator& p)
       // previously this was a bool about snaps, not a flag map
       bool flag;
       ::decode(flag, p);
-      legacy_mds_map.ever_allowed_features = flag ? CEPH_MDSMAP_ALLOW_SNAPS : 0;
+      ever_allowed_features = flag ? CEPH_MDSMAP_ALLOW_SNAPS : 0;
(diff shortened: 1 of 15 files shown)


gregsfortytwo:
I didn't target this at jewel but it definitely needs to go there!

src/mon/MDSMonitor.cc
```diff
@@ -1508,6 +1508,9 @@ class FlagSetHandler : public FileSystemCommandHandler
     string flag_val;
     cmd_getval(g_ceph_context, cmdmap, "val", flag_val);
 
+    string confirm;
+    cmd_getval(g_ceph_context, cmdmap, "confirm", confirm);
```
jcsp: doc/cephfs/administration.rst has a reference to enable_multiple that could be updated for this

src/mon/MDSMonitor.cc
```diff
@@ -1521,7 +1524,12 @@ class FlagSetHandler : public FileSystemCommandHandler
         ss << "Multiple-filesystems are forbidden until all mons are updated";
         return -EINVAL;
       }
-
+      if (confirm != "--yes-i-really-mean-it") {
+	ss << "Multiple FSes is an EXPERIMENTAL and BARELY-TESTED configuration"
```
jcsp: Since it's user facing i'd say "filesystems" instead of FSes

src/mds/MDSMap.h
```diff
@@ -259,8 +259,8 @@ class MDSMap {
 
   void set_snaps_allowed() {
     set_flag(CEPH_MDSMAP_ALLOW_SNAPS);
-    ever_allowed_snaps = true;
-    explicitly_allowed_snaps = true;
+    ever_allowed_features &= CEPH_MDSMAP_ALLOW_SNAPS;
```
jcsp: I think you mean "|=", not "&="

src/mds/MDSMap.cc
```diff
@@ -134,6 +134,8 @@ void MDSMap::dump(Formatter *f) const
 {
   f->dump_int("epoch", epoch);
   f->dump_unsigned("flags", flags);
+  f->dump_unsigned("ever_allowed_features", ever_allowed_features);
```
jcsp: It's awkward to output an integer in JSON output, because to understand it someone will have to come and read the source code to learn what the flags mean.  I'd prefer to see a list of strings here

src/test/encoding/types.h
```diff
@@ -192,6 +192,10 @@ TYPE_FEATUREFUL(InodeStore)
 TYPE_FEATUREFUL(MDSMap)
 TYPE_FEATUREFUL(MDSMap::mds_info_t)
 
+#include "mds/FSMap.h"
+//TYPE_FEATUREFUL(Filesystem)
```
jcsp: Commented out line here

src/mon/MDSMonitor.cc
```diff
@@ -1903,6 +1916,64 @@ class SetHandler : public FileSystemCommandHandler
...
+      int r = parse_bool(val, &enable_multimds, ss);
+      if (r != 0) {
+	return r;
+      }
+
+      if (!enable_multimds) {
+	fsmap.modify_filesystem(fs->fscid,
+	     [](std::shared_ptr<Filesystem> fs)
+		{
+		  fs->mds_map.set_multimds_allowed();
```
jcsp: I think this is enabling it where you meant to disable it

src/mon/MDSMonitor.cc
```diff
@@ -1903,6 +1916,64 @@ class SetHandler : public FileSystemCommandHandler
...
+      int r = parse_bool(val, &enable_dirfrags, ss);
+      if (r != 0) {
+	return r;
+      }
+
+      if (!enable_dirfrags) {
+	fsmap.modify_filesystem(fs->fscid,
+	     [](std::shared_ptr<Filesystem> fs)
+		{
+		  fs->mds_map.set_dirfrags_allowed();
```
jcsp: Same, setting where it should be clearing

jcsp:
I wonder if we ought to be a bit more consistent with our user-facing messages, since these are likely to be one of the first things that people see when poking around (possibly before reading docs or speaking to us).

Currently we have: 

- Multi-MDS clusters are unstable and will probably break your FS!
- directory fragmentation is unstable and may break your FS! 
- Multiple FSes is an EXPERIMENTAL and BARELY-TESTED configuration that may break your entire cluster, probably prevent serious support, and irrevocably marks your cluster.
- Snapshots are unstable and will probably break your FS!

In general, I think that the word "experimental" would sum up what we mean here neatly.  Otherwise we're liable to get questions based on the wording that try to divine what we mean by the difference between "will probably break" and "may break", etc etc.

What about standardising these to something like "Warning: this feature is experimental.  Enabling experimental features may cause problems up to and including loss of data.  Consult the documentation at ceph.com, and if unsure, do no proceed.  Add --yes-i-really-mean-it if you are certain you wish to enable it".

We could then add a page to the docs that lists the experimental feature flags, and that would be where we could wax more lyrical about the situation as needed.

src/mon/MDSMonitor.cc
gregsfortytwo: Fixed.

src/mds/MDSMap.h
gregsfortytwo: Fixed.

src/mds/MDSMap.cc
gregsfortytwo: These aren't really intended for anybody except us to look at and I don't want to devote strings and screen space to reminding users about features we don't want them turning on.

src/test/encoding/types.h
gregsfortytwo: Ah, right. I was trying to decide if we'd ever want to decode a Filesystem separate from an FSMap, since it didn't come with all the functions required for this at the time I was doing debugging. Probably want to just drop it, but any thoughts? (And can you think of any other new types we're missing?)

src/mon/MDSMonitor.cc
gregsfortytwo: Fixed.

src/mon/MDSMonitor.cc
gregsfortytwo: Yeesh, what was wrong with me that day.

gregsfortytwo:
Hm. I started out with the snapshot one and then tried to modify the others to reflect relative levels of risk. You're probably right though.

src/mon/MDSMonitor.cc
gregsfortytwo: Done.

gregsfortytwo:
I'll squash these down if you're happy with them, @jcsp.

src/test/encoding/types.h
jcsp: Yeah, I don't think we need the Filesystem one, it's never encoded on its own anywhere

jcsp:
Changes look good -- what did you think about using the same message for all the "--yes-i-really-mean-it" paths?  Not a blocker.

gregsfortytwo:
On Thu, Apr 7, 2016 at 6:34 AM, John Spray <notifications@github.com> wrote:

> Changes look good -- what did you think about using the same message for
> all the "--yes-i-really-mean-it" paths? Not a blocker.
>
>
Oh whoops, I did the hard part of documenting them and forgot to
standardize the messages. Will do that too.

gregsfortytwo:

PR #8384, opened 2016-03-31
 crush: when take place the crush map should consider the rule is in …

songbaisen:
crush: when take place the crush map should consider the rule is in use

When we remove crush rule in cli we have consider this situation.But when 
take place the crush map,we missing consider it.This will result in all the pg 
in the remove rule pool become to unclean.


songbaisen:
This error may be not relate to my code .For pass it may be need to modefy the osd-crush.sh TEST_crush_repair_faulty_crushmap function.

PR #8386, opened 2016-03-31
client: mount non-default filesystem by name

ukernel:
To mount non-default filesytem, user needs to provide mds namespace ID.
This is inconvenience.

This patch makes user be able to mount filesystem by name. To do this,
client first subscribes to FSMap. Subscribe to mdsmap.<ID> after knowning
ID of the filesystem.

### DIFF
--- a/src/client/Client.cc
+++ b/src/client/Client.cc
@@ -59,6 +59,7 @@
 #include "messages/MClientCapRelease.h"
 #include "messages/MMDSMap.h"
 #include "messages/MFSMap.h"
+#include "messages/MFSMapUser.h"
 
 #include "mon/MonClient.h"
 
@@ -244,7 +245,7 @@ Client::Client(Messenger *m, MonClient *mc)
     objecter_finisher(m->cct),
     tick_event(NULL),
     monclient(mc), messenger(m), whoami(m->get_myname().num()),
-    cap_epoch_barrier(0), fsmap(nullptr),
+    cap_epoch_barrier(0), fsmap(nullptr), fsmap_user(nullptr),
     last_tid(0), oldest_tid(0), last_flush_tid(1),
     initialized(false), authenticated(false),
     mounted(false), unmounting(false),
@@ -488,18 +489,6 @@ int Client::init()
   objecter->start();
 
   monclient->set_want_keys(CEPH_ENTITY_TYPE_MDS | CEPH_ENTITY_TYPE_OSD);
-
-  std::string want = "mdsmap";
-  const auto &want_ns = cct->_conf->client_mds_namespace;
-  if (want_ns != FS_CLUSTER_ID_NONE) {
-    std::ostringstream oss;
-    oss << want << "." << want_ns;
-    want = oss.str();
-  }
-  ldout(cct, 10) << "Subscribing to map '" << want << "'" << dendl;
-
-  monclient->sub_want(want, 0, 0);
-  monclient->renew_subs();
 
   // logger
   PerfCountersBuilder plb(cct, "client", l_c_first, l_c_last);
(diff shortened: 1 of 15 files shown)


ukernel:
https://github.com/ceph/ceph-qa-suite/pull/924

jcsp:
Currently FSMap is only used in Client for sending commands (i.e. normal clients don't use it, but "ceph tell" does).  When we introduce auth caps that limit clients to particular filesystems, it is probably okay for fetching FSMap to simply require a global access, as it is only used for administrative things.

However, if clients are using the FSMap for ordinary mounts, then when we introduce auth caps to limit access[1], we would have to start generating cut-down FSMaps in the mon in order to restrict clients to only see certain filesystems.  That isn't desirable (FSMap is a much larger structure than clients really need to resolve the IDs to names, anyway).

I think it would make more sense to introduce a new thing to watch on the mons which was just an ID->name mapping.  It would be easier for the kclient to decode, and it would be quicker/simpler to generate on-the-fly when we are introducing the client auth caps to limit to certain filesystems.

1. http://tracker.ceph.com/issues/15070

jcsp:
On second thoughts, I suppose in the upgrade path we could just say that old clients would just always require access to all filesystems, and just refuse to send an fsmap to an old client if it didn't.  But since adding a new thing on the mon to return just the id->filesystem mapping is comparatively simple, that's still probably less work

ukernel:
It would be much simpler if monitor supports  mdsmap.<fs_name> subscription. The question is should monitor continue to support  mdsmap.<fs id> subscription.

src/messages/MFSMapUser.h
```diff
@@ -0,0 +1,60 @@
...
+#ifndef CEPH_MFSMAPCOMPACT_H
+#define CEPH_MFSMAPCOMPACT_H
+
+#include "msg/Message.h"
+#include "mds/FSMapUser.h"
+#include "include/ceph_features.h"
+
+class MFSMapUser : public Message {
+ public:
+  epoch_t epoch;
```
jcsp: the epoch and encoded members are unused

src/mds/FSMapUser.h
```diff
@@ -0,0 +1,81 @@
...
+    fs_cluster_id_t cid;
+    std::string name;
+    fs_info_t() : cid(FS_CLUSTER_ID_NONE) {}
+    void encode(bufferlist& bl, uint64_t features) const;
+    void decode(bufferlist::iterator &bl);
+  };
+
+  epoch_t epoch;
+  fs_cluster_id_t legacy_client_fscid;
+  bool enable_multiple;
```
jcsp: What is the use case for a client to know about enable_multiple?  We may not need to include this field

src/client/Client.cc
```diff
@@ -5281,6 +5284,49 @@ int Client::authenticate()
...
+  } while (r == -EAGAIN);
+
+  if (r < 0) {
+    lderr(cct) << "Failed to learn FSMap version: " << cpp_strerror(r) << dendl;
+    return r;
+  }
+
+  ldout(cct, 10) << __func__ << " learned FSMap version " << fsmap_latest << dendl;
+
+  if (user) {
```
jcsp: We should separate this out so that clients would usually fetch `fsmap.user` and only fetch `fsmap` when they need it for doing "tell" commands

ukernel: I don't understand what do you mean here? client does fetch fsmap.user in normal case

src/messages/MFSMapUser.h
ukernel: I will remove encoded member. maybe we should keep epoch member. (to match MFSMap class)

jcsp:
@ukernel please can you rebase this on master, it collided with the check_subs fix

src/client/Client.cc
jcsp: Sorry, I read it wrong the first time.

jcsp:
Reviewed-by, good to go once it's rebased.

Tested here http://pulpito.ceph.com/jspray-2016-05-26_17:59:25-fs-wip-jcsp-testing-20160527---basic-mira/ (for the test branch I just did a naive merge replacing the old check_subs logic with the block that was added here)

jcsp:
@ukernel ping, can you rebase please?

PR #8389, opened 2016-03-31
hammer: rgw: convert plain object to versioned (with null version) when removing

kmroz:
Fixes http://tracker.ceph.com/issues/15254


kmroz:
Needed to rebase after a large commit set went into ceph:hammer-backports

kmroz:
Targeting wrong branch. Closing and will re-open.

PR #8390, opened 2016-03-31
osd: drop map_lock during tick to reduce look contention

xiexingguo:
Because in tick() procedure, we already have osd_lock in hand
    to prevent osdmap from changing. Therefore we don't need to
    acquire map_lock again in order to call maybe_update_heartbeat_peers(),
    which is a pure waste and may cause lock concentration as the
    tick_without_osd_lock periodically need the same lock too.
    
    Only when called from tick_without_osd_lock() or queue_want_up_thru()
    from the peeringWQ, we shall require the map_lock in hand to safely access
    osdmap as we don't with osd_lock held in those cases.
    

xiexingguo:
@liewegas And one more fix about promote_throttle()( although not very sure: )

xiexingguo:
@liewegas More fixes and ping.

src/osd/OSD.cc
```diff
@@ -4182,15 +4182,11 @@ void OSD::tick()
   logger->set(l_osd_cached_crc_adjusted, buffer::get_cached_crc_adjusted());
 
   if (is_active() || is_waiting_for_healthy()) {
-    map_lock.get_read();
```
tchaikov: typo in the title of the commit message:
> look concentration
should be
> lock contention

src/osd/OSD.h
```diff
@@ -791,7 +791,7 @@ class OSDService {
   bool promote_throttle() {
     // NOTE: lockless!  we rely on the probability being a single word.
     promote_counter.attempt();
-    if ((unsigned)rand() % 1000 > promote_probability_millis.read())
+    if ((unsigned)rand() % 1001 >= promote_probability_millis.read())
```
tchaikov: i think it's fine to always consider promoting an object when we have a "1000" `promote_probability_millis`. and `promote_probability_millis` is not always 1000, so the throttling still works as expected.

xiexingguo:
dropped that change and @tchaikov  it is not a typo. Actually I was looking for the proper word for a long time and thanks very much for the reminder, it really helps.

liewegas:
passed rados run

xiexingguo:
@tchaikov Ping

src/osd/OSD.cc
```diff
@@ -3251,6 +3251,8 @@ struct pistate {
 void OSD::build_past_intervals_parallel()
 {
   map<PG*,pistate> pis;
+  set<PG*> fixed;
```
tchaikov: @xiexingguo could you elaborate a little bit the rationale of this change in the commit message?

xiexingguo: @tchaikov update and please check commit message for detail. 
Thanks a lot.

xiexingguo:
Oops, sorry for the misclick, opened again:)

src/osd/OSD.cc
tchaikov: and i don't think we need "set" here.

src/osd/OSD.cc
```diff
@@ -3264,8 +3266,10 @@ void OSD::build_past_intervals_parallel()
 
       epoch_t start, end;
       if (!pg->_calc_past_interval_range(&start, &end, superblock.oldest_map)) {
-        if (pg->info.history.same_interval_since == 0)
+        if (pg->info.history.same_interval_since == 0) {
           pg->info.history.same_interval_since = end;
+          fixed.insert(pg);
```
tchaikov: maybe we can have
```c++
pis.start = pis.end = 0;
pis.same_interval_since = end;
```
and skip the interval calculation later on. and do not set `pg->dirty_big_info` when preparing the transactions if the `pis.begin == pis.end`.

@dzafman what do you think? shall we write the pg info if the pg has no past intervals?

src/osd/OSD.cc
```diff
@@ -3251,6 +3251,8 @@ struct pistate {
 void OSD::build_past_intervals_parallel()
 {
   map<PG*,pistate> pis;
+  set<PG*> fixed;
+  OSDMapRef cur_map, last_map;
```
tchaikov: could also move these definitions to https://github.com/ceph/ceph/pull/8390/commits/657c7f01b99e08f10bc8d613c20a0870df619cbc#diff-fa6c2eba8356ae1442d1bf749beacfdfR3287, closer to where the defined variables are used. but i'd suggest reuse the logic here after confirmed by @dzafman .

xiexingguo: @tchaikov  Thanks for the review and good suggestion, let's wait and see.
:+1:

xiexingguo:
@tchaikov I kill the commit we have a discussion at yesterday because I think it is irrelevant. Also append more cleanups here, would you kindly help looking at this again? Thanks a lot, sir.

PR #8391, opened 2016-03-31
MDSMonitor: introduce command 'fs set_default <fs_name>'

ukernel:
set default filesystem for legacy client


jcsp:
Looks good.  Perhaps add a test to the command line tests?

gregsfortytwo:

PR #8392, opened 2016-03-31
Makefile-env.am: set a default for CEPH_BUILD_VIRTUALENV (part 3)

dachary:
The default values in the run-tox.sh scripts must be updated as well, so
that it is possible to run them from the command line for debug
purposes. Otherwise it is mandatory to run them from

  make TESTS=ceph-detect-init/run-tox.sh check

which introduces a significant delay. The alternative is to run with

  CEPH_DETECT_INIT_VIRTUALENV:=/tmp/ceph-detect-init-virtualenv \
    ceph-detect-init/run-tox.sh

but it's difficult to figure out.

Followup of fa450896ace454dbd140b2d4c71535ab015fc407

### DIFF
--- a/src/ceph-detect-init/run-tox.sh
+++ b/src/ceph-detect-init/run-tox.sh
@@ -23,7 +23,7 @@ if [ x"`uname`"x = xFreeBSDx ]; then
 fi
 
 # run from the ceph-detect-init directory or from its parent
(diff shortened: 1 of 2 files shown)


dmick:
tox confuses and frightens me.  passing this to @alfredodeza

alfredodeza:
this looks OK to me, unsure why the build failed

alfredodeza:
test this please

PR #8402, opened 2016-03-31
configure: Add -D_LARGEFILE64_SOURCE to Linux build.

ilc:
Without -D_LARGEFILE64_SOURCE we can't build against libxfs,
because off64_t must be defined.

Builds on Fedora 23 are broken without this fix.

"mock" builds now pass with this patch in place.


robbat2:
+1, this also explains a Gentoo build failure I saw.

PR #8404, opened 2016-03-31
radosgw-admin: fix for 'realm pull'

cbodley:
this was causing teuthology failures in 'radosgw-admin realm pull':

  ERROR creating control for new realm realm0: (17) File exists

'realm pull' calls RGWRealm::create() with exclusive=false, but RGWRealm::create_control() was using exclusive=true

### DIFF
--- a/src/rgw/rgw_rados.cc
+++ b/src/rgw/rgw_rados.cc
@@ -350,7 +350,7 @@ int RGWZoneGroup::read_default_id(string& default_id, bool old_format)
   return RGWSystemMetaObj::read_default_id(default_id, old_format);
 }
 
-int RGWZoneGroup::set_as_default()
+int RGWZoneGroup::set_as_default(bool exclusive)
 {
   if (realm_id.empty()) {
     /* try using default realm */
@@ -363,7 +363,7 @@ int RGWZoneGroup::set_as_default()
     realm_id = realm.get_id();
   }
 
-  return RGWSystemMetaObj::set_as_default();
+  return RGWSystemMetaObj::set_as_default(exclusive);
 }
 
 int RGWSystemMetaObj::init(CephContext *_cct, RGWRados *_store, bool setup_obj, bool old_format)
@@ -441,7 +441,7 @@ int RGWSystemMetaObj::use_default(bool old_format)
   return read_default_id(id, old_format);
 }
 
-int RGWSystemMetaObj::set_as_default()
+int RGWSystemMetaObj::set_as_default(bool exclusive)
 {
   string pool_name = get_pool_name(cct);
   string oid  = get_default_oid();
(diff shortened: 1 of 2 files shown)


cbodley:
@oritwas added a second commit that changes how `RGWSystemMetaObj::set_as_default()` works for 'realm create' and 'zone create', to address http://tracker.ceph.com/issues/15319

src/rgw/rgw_rados.h
```diff
@@ -783,7 +783,7 @@ class RGWSystemMetaObj {
   }
   int init(CephContext *_cct, RGWRados *_store, bool setup_obj = true, bool old_format = false);
   virtual int read_default_id(string& default_id, bool old_format = false);
-  virtual int set_as_default();
+  virtual int set_as_default(bool exclusive);
```
oritwas: Let set the default vaule to false , it will result in less code change

src/rgw/rgw_rados.cc
```diff
@@ -733,12 +725,11 @@ int RGWRealm::create(bool exclusive)
...
   }
-
-  if (first_realm) { /* this is racy, but it's fine */
-    ret = set_as_default();
-    if (ret < 0) {
-      ldout(cct, 0) << "WARNING: failed to set realm as default realm, ret=" << ret << dendl;
-    }
+  // try to set as default, but pass exclusive=true so we don't override an
+  // existing default
+  ret = set_as_default(true);
```
oritwas: The problem here is that we may race with another create operation. Can you add a comment?

src/rgw/rgw_rados.cc
```diff
@@ -1516,11 +1501,11 @@ int RGWZoneParams::create(bool exclusive)
...
   }
 
-  if (zones.empty()) { /* first zone? maybe, it's a racy check */
-    r = set_as_default();
-    if (r < 0) {
-      ldout(cct, 10) << "WARNING: failed to set zone as default, r=" << r << dendl;
-    }
+  // try to set as default, but pass exclusive=true so we don't override an
+  // existing default
+  r = set_as_default(true);
```
oritwas: add a comment here too

src/rgw/rgw_rados.cc
cbodley: Passing `exclusive=true` makes `set_as_default()` an atomic operation, as I understand it. Two calls to `RGWRealm::create()` can go in parallel, but only one can successfully `set_as_default()`.

oritwas: True , but it could be the newer create realm command. It won't be deterministic. It will be the first one to execute set_as_default (or more correctly to execute the put_obj), better from the previous version but still racy

src/rgw/rgw_rados.cc
```diff
@@ -733,12 +725,11 @@ int RGWRealm::create(bool exclusive)
...
     return ret;
   }
-
-  if (first_realm) { /* this is racy, but it's fine */
-    ret = set_as_default();
-    if (ret < 0) {
-      ldout(cct, 0) << "WARNING: failed to set realm as default realm, ret=" << ret << dendl;
-    }
+  // try to set as default. may race with another create, so pass exclusive=true
+  // so we don't override an existing default
```
cbodley: @oritwas how does this look?

oritwas: great

PR #8405, opened 2016-03-31
common: Fix ObjBencher clean_up when no prefix specified

dzafman:
Fix rados cleanup when there is no benchmark_last_metadata available
Support --all option with rados cleanup
Prevent rados cleanup from being used to remove non rados bench objects

Fixes: http://tracker.ceph.com/issues/15338

### DIFF
--- a/src/common/obj_bencher.cc
+++ b/src/common/obj_bencher.cc
@@ -45,11 +45,10 @@ static std::string generate_object_prefix_nopid() {
 }
 
 static std::string generate_object_prefix(int pid = 0) {
-  if (!cached_pid) {
-    if (!pid)
-      pid = getpid();
+  if (pid)
     cached_pid = pid;
-  }
+  else if (!cached_pid)
+    cached_pid = getpid();
 
   std::ostringstream oss;
   oss << generate_object_prefix_nopid() << "_" << cached_pid;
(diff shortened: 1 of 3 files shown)


dzafman:
The testing is part of this and runs during make check.

src/common/obj_bencher.cc
```diff
@@ -1084,31 +1079,67 @@ int ObjBencher::rand_read_bench(int seconds_to_run, int num_objects, int concurr
...
+    r = fetch_bench_metadata(run_name_meta, &op_size, &object_size, &num_objects, &prevPid);
+    if (r < 0) {
       return r;
     }
-  }
 
-  r = clean_up(num_objects, prevPid, concurrentios);
-  if (r != 0) return r;
+    r = clean_up(num_objects, prevPid, concurrentios);
+    if (r != 0) return r;
```
athanatos: Whitespace thing?

src/common/obj_bencher.cc
```diff
@@ -1084,31 +1079,67 @@ int ObjBencher::rand_read_bench(int seconds_to_run, int num_objects, int concurr
...
 
-  r = sync_remove(run_name_meta);
-  if (r != 0) return r;
+    r = sync_remove(run_name_meta);
+    if (r != 0) return r;
+  }
```
athanatos: Same here?

athanatos:
lgtm

src/common/obj_bencher.cc
dzafman: @athanatos Yes, this and the previous case are now inside a while loop, so are indented 1 additional level.

PR #8406, opened 2016-04-01
debian/rules: put init-ceph in /etc/init.d/ceph, not ceph-base

dmick:
When the package name changed from ceph to ceph-base, dh_installinit
started installing the init script into /etc/init.d/ceph-base.  Fix
this by using --name ceph with dh_installinit, which requires
1) naming the .init file ceph-base.ceph.init, and
2) calling dh_installinit separately for each package

http://tracker.ceph.com/issues/15329

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -51,7 +51,7 @@ build-stamp: configure-stamp
 
 	$(MAKE)
 
-	cp src/init-ceph debian/ceph-base.init
+	cp src/init-ceph debian/ceph-base.ceph.init
 	cp src/init-radosgw debian/radosgw.init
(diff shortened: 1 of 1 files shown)


tchaikov:
lgtm.

Dan, could you put `Fixes: http://tracker.ceph.com/issues/15329` right before your `Signed-off-by` line, see https://github.com/ceph/ceph/blob/master/SubmittingPatches.rst#preparing-and-sending-patches. we start using this notation couple days ago.

dmick:
I can, although the URL is already there, and I know tracker is going to pay no attention to it anyway...but I will fix it up

PR #8407, opened 2016-04-01
common/util: add support for distro info from /etc/os-release file

dx9:


src/common/util.cc
```diff
@@ -115,47 +115,100 @@ int get_fs_stats(ceph_data_stats_t &stats, const char *path)
...
+      lderr(cct) << "os_release_parse - failed to open /etc/os-release: " << cpp_strerror(ret) << dendl;
+    return;
+  }
+
+  file_values_parse(vm, fp, m, cct);
+
+  fclose(fp);
+}
+
+static void distro_detect(map<string, string> *m, CephContext *cct)
```
liewegas: may as well return an error code here?

PR #8408, opened 2016-04-01
mds: fix potential null pointer access; message reference leak

xiexingguo:


xiexingguo:
@gregsfortytwo Glad you are back, would you have a look at this? Thanks

gregsfortytwo:
It's in my queue. I don't think those message puts are correct though and haven't had a chance to go check things. (If you document why you think those kinds of things are wrong, it goes a lot faster in review.)

src/mds/MDSDaemon.cc
```diff
@@ -1245,6 +1245,7 @@ bool MDSDaemon::handle_core_message(Message *m)
     if (mds_rank) {
       mds_rank->handle_osd_map();
     }
+    m->put();
```
xiexingguo: @gregsfortytwo Take this one for example. In all the above cases, we are going to return true to indicate we have successfully handled these messages  and therefore are responsible for drop reference of these messages(to get them correctly released). If you take a look at  https://github.com/ceph/ceph/pull/8408/commits/358c81d67786bae27ab45d5f3507d0efb8238026#diff-89e22a97695f436e6645ce25f7bfb985R1223, I think the one is exactly the same.

src/mds/MDSDaemon.cc
```diff
@@ -989,6 +989,7 @@ void MDSDaemon::handle_mds_map(MMDSMap *m)
             // has taken our ID, we don't want to keep restarting and
             // fighting them for the ID.
             suicide();
+            m->put();
```
xiexingguo: @gregsfortytwo  And for this one, because on any other exits(check https://github.com/ceph/ceph/pull/8408/commits/473371996d5b7c2402d1ef7d0901a5e350ba9673#diff-89e22a97695f436e6645ce25f7bfb985L914 and https://github.com/ceph/ceph/pull/8408/commits/473371996d5b7c2402d1ef7d0901a5e350ba9673#diff-89e22a97695f436e6645ce25f7bfb985R1035) are decreasing the message's reference by one.  
Although this one is not necessary because we are going to shutdown anyway, but it is a good habit to do the cleanup.

xiexingguo:
update commit message also to make things clear

xiexingguo:
@gregsfortytwo @jcsp Ping

xiexingguo:
@jcsp Mind having a look at this one too? Thanks. As sir Gregory seems to be not available at this moment:)

gregsfortytwo:
Ah, I was worried about the MOSDMap put – I believe that once upon a time, we didn't handle that message in order to make sure it got passed to the Objecter. But apparently we are now the last Dispatcher in the chain and should be dumping it.

gregsfortytwo:
...or actually, maybe it's just from changing our internal interfaces all over...since apparently we were reporting the message as handled anyway.

gregsfortytwo:

PR #8410, opened 2016-04-01
rgw: add a few more help options in admin interface

theanalyst:
Added a few missing commandline switches for zone related and orphans related help

### DIFF
--- a/src/rgw/rgw_admin.cc
+++ b/src/rgw/rgw_admin.cc
@@ -197,8 +197,11 @@ void _usage()
   cout << "   --realm-new-name=<realm new name> realm new name\n";
   cout << "   --rgw-zonegroup=<zonegroup>   zonegroup name\n";
   cout << "   --rgw-zone=<zone>         zone in which radosgw is running\n";
+  cout << "   --zone-id=<zone id>       zone id\n";
   cout << "   --zone-new-name=<zone>    zone new name\n";
+  cout << "   --source-zone             specify the source zone (for data sync)\n";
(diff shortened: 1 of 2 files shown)


src/rgw/rgw_admin.cc
```diff
@@ -197,8 +197,11 @@ void _usage()
   cout << "   --realm-new-name=<realm new name> realm new name\n";
   cout << "   --rgw-zonegroup=<zonegroup>   zonegroup name\n";
   cout << "   --rgw-zone=<zone>         zone in which radosgw is running\n";
+  cout << "   --zone-id=<zone id>       zone id\n";
   cout << "   --zone-new-name=<zone>    zone new name\n";
+  cout << "   --source-zone             specify the source zone (for data sync)";
```
xiexingguo: @theanalyst The false is true, I think you need to append a **newline** here at least.

theanalyst: Ah ok thanks!

theanalyst:
Hm I think whitespacing is an issue in the help file.. causing cli tests to fail.. Let me fix this

theanalyst:
It was basically tabs in the file (that need to be present on a few newlines)

xiexingguo:
:+1:  It's really hard to distinguish these. I think it's better to run UT locally, which shall help to generate the right help.t.

theanalyst:
@xiexingguo yeah actually the problem was my editor stripped lines with empty whitespaces by default, which caused the issue.. now all is well :)

PR #8412, opened 2016-04-01
libradosstriper: async rm, truncate

sponce:
optimized truncation and removal of striped objects in the radosstriper library by using the async API of the rados layer.
The previous implementation was a first, inefficient one.


liewegas:
@sponce what's the status of this?  looks like it needs a rebase at least

liewegas:
@sponce ping

jdurgin:
this is included in #10049

PR #8416, opened 2016-04-01
librados: modify Pipe::connect() to return the error code - part 1

vumrao:
librados: modify Pipe::connect() to return the error code - part 1
Fixes: http://tracker.ceph.com/issues/15308

### DIFF
--- a/src/msg/simple/Pipe.cc
+++ b/src/msg/simple/Pipe.cc
@@ -883,7 +883,7 @@ int Pipe::connect()
   pipe_lock.Unlock();
   
   char tag = -1;
-  int rc;
+  int rc = -1;
   struct msghdr msg;
   struct iovec msgvec[2];
   int msglen;
@@ -901,7 +901,8 @@ int Pipe::connect()
   // create socket?
   sd = ::socket(peer_addr.get_family(), SOCK_STREAM, 0);
   if (sd < 0) {
-    lderr(msgr->cct) << "connect couldn't created socket " << cpp_strerror(errno) << dendl;
+    rc = -errno;
+    lderr(msgr->cct) << "connect couldn't created socket " << cpp_strerror(rc) << dendl;
     goto fail;
   }
 
@@ -913,15 +914,17 @@ int Pipe::connect()
   ldout(msgr->cct,10) << "connecting to " << peer_addr << dendl;
   rc = ::connect(sd, (sockaddr*)&peer_addr.addr, peer_addr.addr_size());
   if (rc < 0) {
+    rc = -errno;
     ldout(msgr->cct,2) << "connect error " << peer_addr
-	     << ", " << cpp_strerror(errno) << dendl;
+	     << ", " << cpp_strerror(rc) << dendl;
     goto fail;
   }
 
   // verify banner
   // FIXME: this should be non-blocking, or in some other way verify the banner as we get it.
-  if (tcp_read((char*)&banner, strlen(CEPH_BANNER)) < 0) {
-    ldout(msgr->cct,2) << "connect couldn't read banner, " << cpp_strerror(errno) << dendl;
+  rc = tcp_read((char*)&banner, strlen(CEPH_BANNER));
+  if (rc < 0) {
+    ldout(msgr->cct,2) << "connect couldn't read banner, " << cpp_strerror(rc) << dendl;
     goto fail;
   }
   if (memcmp(banner, CEPH_BANNER, strlen(CEPH_BANNER))) {
@@ -935,8 +938,9 @@ int Pipe::connect()
   msg.msg_iov = msgvec;
   msg.msg_iovlen = 1;
   msglen = msgvec[0].iov_len;
-  if (do_sendmsg(&msg, msglen)) {
-    ldout(msgr->cct,2) << "connect couldn't write my banner, " << cpp_strerror(errno) << dendl;
+  rc = do_sendmsg(&msg, msglen);
+  if (rc < 0) {
+    ldout(msgr->cct,2) << "connect couldn't write my banner, " << cpp_strerror(rc) << dendl;
     goto fail;
   }
 
@@ -950,8 +954,9 @@ int Pipe::connect()
 #endif
     addrbl.push_back(std::move(p));
   }
-  if (tcp_read(addrbl.c_str(), addrbl.length()) < 0) {
-    ldout(msgr->cct,2) << "connect couldn't read peer addrs, " << cpp_strerror(errno) << dendl;
+  rc = tcp_read(addrbl.c_str(), addrbl.length());
+  if (rc < 0) {
+    ldout(msgr->cct,2) << "connect couldn't read peer addrs, " << cpp_strerror(rc) << dendl;
     goto fail;
   }
   {
@@ -987,8 +992,9 @@ int Pipe::connect()
   msg.msg_iov = msgvec;
   msg.msg_iovlen = 1;
   msglen = msgvec[0].iov_len;
-  if (do_sendmsg(&msg, msglen)) {
-    ldout(msgr->cct,2) << "connect couldn't write my addr, " << cpp_strerror(errno) << dendl;
+  rc = do_sendmsg(&msg, msglen);
+  if (rc < 0) {
+    ldout(msgr->cct,2) << "connect couldn't write my addr, " << cpp_strerror(rc) << dendl;
     goto fail;
   }
   ldout(msgr->cct,10) << "connect sent my addr " << msgr->my_inst.addr << dendl;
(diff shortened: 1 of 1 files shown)


vumrao:
@jdurgin Hello Josh as discussed I have sent this first part of fix for pipe::connect() for tracker : http://tracker.ceph.com/issues/15308

gregsfortytwo:
You can't just set errno on all those callchains. tcp_read() calls into tcp_read_wait() which returns -1 if the socket has been closed, for instance — nothing to do with errno.

I am also not 100% sure that you can rely on errno's value being set correctly once you've exited your initial function context; the rules for it around that or threading or something are a bit funky.

vumrao:
Thank you Greg for your inputs.

>You can't just set errno on all those callchains. tcp_read() calls into tcp_read_wait() which returns -1 if the socket has been closed, for instance — nothing to do with errno.

Sure I will make the change for tcp_read and then I checked Pipe::do_sendmsg() and same goes for do_sendmsg(). This function do have a standard call to sendmsg() and I think we can handle errno properly here. I am sending the next patch set with making changes to do_sendmsg().
Please let me know your inputs.

> I am also not 100% sure that you can rely on errno's value being set correctly once you've exited your initial function context; the rules for it around that or threading or something are a bit funky.

For this I had discussion with Josh and because of this I made it as part 1 . We discussed any developer who has more background of code , he can send patch for part 2. Part 2 patch will take pipe::connect() returned errno to reach properly to rados_connect().

yuyuyu101:
I think part 2 is a little difficult for current impl. we don't have any way to notify the caller who want to connect to a endpoint except ms_handle*.

gregsfortytwo:
>>I am also not 100% sure that you can rely on errno's value being set correctly once you've exited your initial function context; the rules for it around that or threading or something are a bit funky.

>For this I had discussion with Josh and because of this I made it as part 1 . We discussed any developer who has more background of code , he can send patch for part 2. Part 2 patch will take pipe::connect() returned errno to reach properly to rados_connect().

I mean, you may need to return errno from do_sendmsg and tcp_read, rather than reading it out in Pipe::connect directly. Not certain, though.

vumrao:
> I think part 2 is a little difficult for current impl. we don't have any way to notify the caller who want to connect to a endpoint except ms_handle*. 

Thanks @yuyuyu101 for your inputs. I am sure it is little difficult :). @jdurgin and I discussed the same.

vumrao:
> I mean, you may need to return errno from do_sendmsg and tcp_read, rather than reading it out in Pipe::connect directly. Not certain, though.

Thanks Greg, sorry I misunderstood it. In `do_sendmsg()` I had added `errno`. Now I have added `errno` in `tcp_read` , `tcp_read_wait()` and `tcp_read_nonblocking()`.

src/msg/simple/Pipe.cc
```diff
@@ -2440,8 +2453,11 @@ int Pipe::write_message(const ceph_msg_header& header, const ceph_msg_footer& fo
 
 int Pipe::tcp_read(char *buf, unsigned len)
 {
-  if (sd < 0)
-    return -1;
+  int rc = 0;
+  if (sd < 0) {
+    rc = -errno;
```
jdurgin: errno isn't getting set by anything - it'll just have an unrelated value here.

For these sanity checks, tcp_{read,write} should return a specific error, perhaps EINVAL

src/msg/simple/Pipe.cc
```diff
@@ -2452,25 +2468,32 @@ int Pipe::tcp_read(char *buf, unsigned len)
       }
     }
 
-    if (tcp_read_wait() < 0)
-      return -1;
+    if (tcp_read_wait() < 0) {
+      rc = -errno;
```
jdurgin: reading errno should be moved to right after the syscall, not in any of the callers of these wrappers. that'll avoid as much as possible getting invalid errno values, and avoiding accidental resetting of it if the code changes

src/msg/simple/Pipe.cc
```diff
@@ -2565,18 +2594,20 @@ ssize_t Pipe::buffered_recv(char *buf, size_t len, int flags)
 
 ssize_t Pipe::tcp_read_nonblocking(char *buf, unsigned len)
 {
+  int rc = -1;
   ssize_t got = buffered_recv(buf, len, MSG_DONTWAIT );
```
jdurgin: same thing here, the errno value is only useful all the way in do_recv here, buffered_recv and tcp_read_nonblocking should just pass the return code up, and not read errno themselves, since they are not calling any syscalls that set it

vumrao:
@jdurgin Thank you Josh for your inputs and explaining things in detail. Sending new patch set.

jdurgin:
lgtm now, thanks

vumrao:
Thank you Josh!

src/msg/simple/Pipe.cc
```diff
@@ -901,7 +901,8 @@ int Pipe::connect()
   // create socket?
   sd = ::socket(peer_addr.get_family(), SOCK_STREAM, 0);
   if (sd < 0) {
-    lderr(msgr->cct) << "connect couldn't created socket " << cpp_strerror(errno) << dendl;
+    rc = -errno;
+    lderr(msgr->cct) << "connect couldn't created socket " << cpp_strerror(rc) << dendl;
```
tchaikov: why change this line, `cpp_strerror(errno)` is better here.

vumrao: Thanks Kefu, I have made the change.

src/msg/simple/Pipe.cc
```diff
@@ -2244,7 +2257,7 @@ int Pipe::do_sendmsg(struct msghdr *msg, unsigned len, bool more)
     }
   }
   restore_sigpipe();
-  return 0;
+  return rc;
```
tchaikov: why you are changing this line? imo, it would be easier for the reader if we are sure that this branch always return `0`?

src/msg/simple/Pipe.cc
```diff
@@ -2495,7 +2511,7 @@ int Pipe::tcp_read_wait()
   if (!(pfd.revents & POLLIN))
     return -1;
 
-  return 0;
+  return rc;
```
tchaikov: ditto.

src/msg/simple/Pipe.cc
```diff
@@ -2210,16 +2221,18 @@ int Pipe::do_sendmsg(struct msghdr *msg, unsigned len, bool more)
...
     if (r == 0) 
       ldout(msgr->cct,10) << "do_sendmsg hmm do_sendmsg got r==0!" << dendl;
-    if (r < 0) { 
-      ldout(msgr->cct,1) << "do_sendmsg error " << cpp_strerror(errno) << dendl;
+    if (r < 0) {
+      rc = -errno; 
+      ldout(msgr->cct,1) << "do_sendmsg error " << cpp_strerror(rc) << dendl;
       restore_sigpipe();
-      return -1;
+      return rc;
```
tchaikov: could return -errno right away without introducing a local "rc" here.

vumrao: @tchaikov ack will make the changes.

src/msg/simple/Pipe.cc
```diff
@@ -901,6 +901,7 @@ int Pipe::connect()
   // create socket?
   sd = ::socket(peer_addr.get_family(), SOCK_STREAM, 0);
   if (sd < 0) {
+    rc = -errno;
     lderr(msgr->cct) << "connect couldn't created socket " << cpp_strerror(errno) << dendl;
```
jdurgin: the reason to avoid cpp_strerror(errno) here is perhaps a bit paranoid, but it's to avoid errno being changed before we get to printing it here.

e.g. lderr() or operator << or other things called indirectly could call some function that sets errno, causing us to lose the original error code. hence the preference for cpp_strerror(rc).

vumrao: Thanks Josh for your inputs, sorry I missed to explain this to Kefu, why we did like this.

tchaikov: josh, makes sense. thanks for the explanation. yeah, `clock_gettime(2)` could set errno.

tchaikov:
lgtm

vumrao:
Thank you Josh and Kefu for your help in this PR.

src/msg/simple/Pipe.cc
dzafman: I prefer to keep this change, since it is better to get errno saved before something might make another system call before cpp_strerror() is called now or in the future.

tchaikov: thanks @dzafman, josh also reminded me this at https://github.com/ceph/ceph/pull/8416#discussion_r58825773 .

vumrao:
doing rebase to trigger the build again.

vumrao:
@liewegas Thank you.

PR #8417, opened 2016-04-01
librbd: avoid throwing error if mirroring is unsupported

dillaman:
Attempting to remove an image will remove the image from the mirroring
directory.  However, if the OSD is older and doesn't support this
new feature, avoid throwing an error.


jdurgin:
lgtm

PR #8419, opened 2016-04-01
common: fix time_t cast in decode

adamemerson:


tchaikov:
Adam, could you add a `Fixes: http://tracker.ceph.com/issues/15330` line right before your "Signed-off-by" label?

adamemerson:
@tchaikov And _so_ it has been done.

liewegas:
lgtm

src/common/ceph_time.h
```diff
@@ -382,10 +382,13 @@ void encode(const std::chrono::time_point<Clock, Duration>& t,
 template<typename Clock, typename Duration>
 void decode(std::chrono::time_point<Clock, Duration>& t,
 	    bufferlist::iterator& p) {
-  uint32_t s, ns;
+  uint32_t s;
+  uint32_t ns;
```
tchaikov: @adamemerson i don't quite understand you change. the commit message reads 

> Since we're decoding 32-bit integers, just use uint32_t

we are using `uint32_t`, aren't we?

> and then cast them to what utime_t expects.

and i think
```c++
struct timespec ts = {s, ns};
```
and 
```c++
  struct timespec ts = {
    static_cast<time_t>(s),
    static_cast<long int>(ns)};
```
are equivalent. but the later casts the temp variables explicitly. could you help me understand this commit? thanks.

adamemerson: The difference is in the [definition of struct timespec](http://pubs.opengroup.org/onlinepubs/007908775/xsh/time.h.html).
````
struct timespec {
  time_t tv_sec;
  long tv_nsec;
};
````
time_t is remarkably unspecified. On x86_64 Linux it is a signed 64-bit integer. So we can assign an unsigned 32-bit integer to it with no trouble at all. On i386 Linux, it is a signed 32-bit integer. Assigning a uint32_t to it will generate a warning.

Similarly, 'long' on x86_64 Linux is a 64-bit signed integer. Assigning a uint32_t to that is fine. But on i386 Linux, it generates a warning.

By explicitly casting we avoid having 'narrowing conversion' warnings fill the console of anyone who dares compile under i386.

tchaikov: thank you, got it!

tchaikov:
lgtm.

PR #7303, opened 2016-01-20
add custom move operations for ObjectStore::Transaction

cbodley:
I wrote up a quick unit test to make sure that ObjectStore::Transaction's default move operations were doing what we expected them to, and it turns out that they weren't.

What I learned is that the compiler treats plain-old-data types specially when generating their default move operations. Because the builtin types don't have default initial values, it reverts to a normal copy and makes no effort to restore the source object to default values. See 'Trivial move constructor' at http://en.cppreference.com/w/cpp/language/move_constructor for more information.

I added a custom move constructor and move assignment operator to TransactionData that does restore the default values, and the unit tests pass as expected.

PR #8421, opened 2016-04-02
os/ObjectStore: add noexcept to ensure move ctor is used

tchaikov:
otherwise vector::push_back() will use the copy ctor if it resizes,
to enforce its strong exception guarantee.


tchaikov:
@cbodley would you please take a look? this is a follow-up of #7303.

liewegas:
passed rados run

cbodley:
@tchaikov good catch :+1:

PR #8428, opened 2016-04-04
ceph-rest-api: fix fs/flag/set

liewegas:


liewegas:
@dmick do you remember why the URL prefix was doing that?

liewegas:
rebased this on @jcsp's fix, so the test is already passing, but this is still needed to fix the fs/flag/set api via ceph-rest-api.

PR #8432, opened 2016-04-04
mon: warn if 'sortbitwise' flag is not set and no legacy OSDs are present

liewegas:
If the user has upgraded, they should set sortbitwise.


liewegas:
@athanatos ?

PR #8437, opened 2016-04-05
os/kv, os/bluestore: Enable SingleDelete with RocksDB and Bluestore

jjhuo:
SingleDelete is useful for log-insert-merge tree based key value store, such as RocksDB, to avoid more LSM compactions for already deleted key value pairs.

By use of single delete, RocksDB should be able to remove deleted wal/overlay entries with only one compaction in theory, when they land on level0. This should reduce RocksDB WAF.

### DIFF
--- a/src/kv/KeyValueDB.h
+++ b/src/kv/KeyValueDB.h
@@ -88,6 +88,16 @@ public:
       const std::string &k	      ///< [in] Key to remove
       ) = 0;
 
+    /// Remove Single Key which exists and was not overwritten.
+    /// This API is only related to performance optimization, and should only be 
+    /// re-implemented by log-insert-merge tree based keyvalue stores(such as RocksDB). 
+    /// If a key is overwritten (by calling set multiple times), then the result
+    /// of calling rm_single_key on this key is undefined.
+    virtual void rm_single_key(
+      const std::string &prefix,   ///< [in] Prefix to search for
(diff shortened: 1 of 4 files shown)


jjhuo:
[       OK ] ObjectStore/StoreTest.TryMoveRename/3 (201 ms)
[----------] 176 tests from ObjectStore/StoreTest (1107318 ms total)
[==========] 176 tests from 1 test case ran. (1107318 ms total)
[  PASSED  ] 176 tests.

markhpc:
Would it be possible to get test results on the reduction in compactions using this code vs the old code?

liewegas:
oops, i didn't see this, and coded up https://github.com/ceph/ceph/pull/8708

liewegas:
pulled in your last commit as i hadn't covered the overlay case

jjhuo:
Sure, no problem. Either way is fine with me.
Sorry it took me a while to get some preliminary results, had to wait for a new machine and SSD.
For 1 hour fio_rbd 100% 4KB random write test, I saw WAF reported by RocksDB dropped from 9.5 to 8.8.

jjhuo:
For 6 hour fio_rbd 100% 4KB random write test, I saw WAF reported by RocksDB dropped from 9.5 to 8.6, and IOPS improved by 10%.

PR #8438, opened 2016-04-05
cmake: fix the build of test_rados_api_list

tchaikov:
the libglobal linkage was added in 769c0af, so add it to cmake
accordingly.


yuyuyu101:
cool, I just met this.... cmake always broken after a round of merge...

PR #8441, opened 2016-04-05
rgw: Set Access-Control-Allow-Origin to a Asterisk if allowed in a rule

wido:
Before this patch the RGW would respond with the Origin send by the client in the request
if a wildcard/asterisk was specified as a valid Origin.

This patch makes sure we respond with a header like this:

  Access-Control-Allow-Origin: *

This way a resource can be used on different Origins by the same browser and that browser
will use the content as the asterisk.

We also keep in mind that when Authorization is send by the client different rules apply.
In the case of Authorization we may not respond with an Asterisk, but we do have to
add the Vary header with 'Origin' as a value to let the browser know that for different
Origins it has to perform a new request.

More information: https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS

Fixes: #15348

### DIFF
--- a/src/rgw/rgw_cors.cc
+++ b/src/rgw/rgw_cors.cc
@@ -116,6 +116,13 @@ static bool is_string_in_set(set<string>& s, string h) {
   return false;
 }
 
+bool RGWCORSRule::has_wildcard_origin() {
+  if (allowed_origins.find("*") != allowed_origins.end())
+    return true;
+
+  return false;
+}
+
 bool RGWCORSRule::is_origin_present(const char *o) {
   string origin = o;
   return is_string_in_set(allowed_origins, origin);
--- a/src/rgw/rgw_cors.h
+++ b/src/rgw/rgw_cors.h
@@ -80,6 +80,7 @@ public:
     ::decode(exposable_hdrs, bl);
     DECODE_FINISH(bl);
   }
+  bool has_wildcard_origin();
   bool is_origin_present(const char *o);
   void format_exp_headers(std::string& s);
   void erase_origin_if_present(std::string& origin, bool *rule_empty);
--- a/src/rgw/rgw_op.cc
+++ b/src/rgw/rgw_op.cc
@@ -637,6 +637,18 @@ bool RGWOp::generate_cors_headers(string& origin, string& method, string& header
   if (!rule)
     return false;
 
+  /*
+   * Set the Allowed-Origin header to a asterisk if this is allowed in the rule
+   * and no Authorization was send by the client
+   *
+   * The origin parameter specifies a URI that may access the resource.  The browser must enforce this.
+   * For requests without credentials, the server may specify "*" as a wildcard,
+   * thereby allowing any origin to access the resource.
+   */
+  const char *authorization = s->info.env->get("HTTP_AUTHORIZATION");
+  if (!authorization && rule->has_wildcard_origin())
+    origin = "*";
+
   /* CORS 6.2.3. */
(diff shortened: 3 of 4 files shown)


oritwas:
@wido,
I tried running s3tests on this code and it causes several errors and even a crash:
s3tests.functional.test_s3.test_cors_origin_response ... ERROR
s3tests.functional.test_s3.test_atomic_read_1mb ... ERROR
s3tests.functional.test_s3.test_atomic_read_4mb ... ERROR
s3tests.functional.test_s3.ttest_s3.test_atomic_read_8mb - crashe

oritwas:
@wido , ignore this.
It looks like my build environment got corrupted . I will clean everything and test again

oritwas:
Confirmed:
s3tests.functional.test_s3.test_cors_origin_response fail and 
s3tests.functional.test_s3.test_atomic_read_1mb crashes the gateway

wido:
@oritwas Thanks for the tests! I don't know why the gateway crashes. That's odd...

Kind of busy at the moment, but I'll look into this asap.

wido:
@oritwas I have the tests running now and even on master I see multiple tests failing.

Could you share your command to run the s3 tests so that we are running it the same way?

wido:
@oritwas I think I figured it out. I pushed a new version of my code and should be there now.

Tests also pass:

<pre>root@ceph-builder:~/s3-tests# S3TEST_CONF=/root/s3.conf ./virtualenv/bin/nosetests s3tests.functional.test_s3:test_cors_origin_response
.
Ran 1 test in 7.223s

OK
root@ceph-builder:~/s3-tests# S3TEST_CONF=/root/s3.conf ./virtualenv/bin/nosetests s3tests.functional.test_s3:test_atomic_read_1mb
.
Ran 1 test in 0.283s

OK
root@ceph-builder:~/s3-tests#</pre>

wido:
I verified with Wireshark that the proper headers are being returned in these cases.

I also opened a PR in the s3-tests repo to add tests for this use-case.

indykish:
@wido Is this fix shipped in `Jewel` ? http://ceph.com/releases/

I couldn't locate `15348` in the release-notes though.

wido:
No, it is not. I hope it will be in 10.2.3 since I'm waiting for this fix to come out as well :)

rajesh-rajagopal:
@wido i tried to fix the CORS problem in jewel using this boto configuration 

import boto
from boto.s3.connection import S3Connection
host = 'madai.megam.io'
access_key = '--------'
secret_key = '-------'
print "keys"
conn = S3Connection(
    host=host,
    aws_access_key_id=access_key,
    aws_secret_access_key=secret_key,
     is_secure=True,
    calling_format = boto.s3.connection.OrdinaryCallingFormat(),)
print "connected"
my_cors_conf = """
<CORSConfiguration>
 <CORSRule>
   <AllowedOrigin>*</AllowedOrigin>
   <AllowedMethod>GET</AllowedMethod>
   <AllowedMethod>POST</AllowedMethod>
   <AllowedHeader>*</AllowedHeader>
</CORSRule>
</CORSConfiguration>
"""
print "bucket creation"
bucket = conn.get_bucket('----')
print "bucket"
bucket.set_cors_xml(my_cors_conf)

but its doesn't working ..

indykish:
@wido  @liewegas  Will the fix help to `set_cors` via `boto` ?

wido:
No, this only fixes the return headers send by the RADOS Gateway. Nothing with Python boto

indykish:
@wido  Thanks

PR #8442, opened 2016-04-05
rbd-mirror: replay status now available via the rbd api and cli

trociny:


trociny:
@dillaman @jdurgin This is still WIP.

Right now status object that is stored in rbd_mirroring object contains two fileds: state and description. The description has details about the current state, e.g:
```
  state:       replaying
  description: replaying, master_position=[object_number=3, tag_tid=2, entry_tid=3], mirror_position=[object_number=3, tag_tid=2, entry_tid=3]
```
The global_image_id is used as a key.

The status also contains metadata that helps users to understand if info is outdated (e.g. due to rbd-mirror daemon not running). Metadata includes 1) origin of "set status" request, and  2) last update time. Metadata is generated on the server side, when processing "set status" method call, so it is transparent to the user.

On start the rbd-mirror daemon registers a watcher for rbd_mirroring object. When "get status" method is called by a user, it compares status' origin with currently registered watchers. If the origin is not found among the watchers, the status is returned with "stale" flag set. This is done on the server side, and for this I had to add an objclass method to list watchers. Also last_update_time metadata is return.

Right now "stale" and "update_time" metadata are returned in two additional params for mirror_image_status_get method. Not sure it is the best solution. They could be added to status object. Or a additional object "metadata" could be created. What do you think?

Some yet unresolved issues:

- In rbd I use cls::rbd methods directly. I think I will add RBD API methods and will use them instead.
- The output for `rbd mirror pool status` command contains image IDs (and no image names), which would not be very convenient for users. To get image names I would need to call cls_client::dir_get_name, or open every image. Or I could cache image name in status. What do you think?
- If the image replayer in "SYNCING" state, it would be nice , to store syncing progress in "description". I have not looked yet how difficult it is to get it.
- Currently unittest fails due to:
```
  2016-04-05 15:37:57.955039 7fd88f728b00 -1 Failed to load class: .libs/libcls_rbd.so: undefined symbol: _Z21cls_cxx_list_watchersPvP25obj_list_watch_response_t
```
Not sure why it fails if obj_list_watch_response_t is fully defined in the included osd_types.h. I will investigate this. BTW, unittest_rbd_mirror fails to load also these objects:
```
2016-04-05 15:37:57.951510 7fd88f728b00 -1 Failed to load class: .libs/libcls_hello.so: undefined symbol: _ZTI10PGLSFilter
2016-04-05 15:37:57.952249 7fd88f728b00 -1 Failed to load class: .libs/libcls_user.so: undefined symbol: _Z23cls_cxx_map_read_headerPvPN4ceph6buffer4listE
2016-04-05 15:37:57.953890 7fd88f728b00 -1 Failed to load class: .libs/libcls_version.so: undefined symbol: _Z19cls_gen_rand_base64Pci
2016-04-05 15:37:57.953956 7fd88f728b00 -1 Failed to load class: .libs/libcls_log.so: undefined symbol: _Z23cls_cxx_map_read_headerPvPN4ceph6buffer4listE
2016-04-05 15:37:57.958821 7fd88f728b00 -1 Failed to load class: .libs/libcls_rgw.so: undefined symbol: _Z23cls_cxx_map_read_headerPvPN4ceph6buffer4listE
2016-04-05 15:37:57.960766 7fd88f728b00 -1 Failed to load class: .libs/libcls_cephfs.so: undefined symbol: _ZTI10PGLSFilter
2016-04-05 15:37:57.966028 7fd88f728b00 -1 Failed to load class: .libs/libcls_refcount.so: undefined symbol: _Z14cls_cxx_removePv
```

dillaman:
@trociny The issue with ```_Z21cls_cxx_list_watchersPvP25obj_list_watch_response_t``` is because librados_test_stub doesn't implement 100% of the librados (and cls) API.  Instead, it only implements what has been necessary to run librbd / rbd-mirror thus far.

You would need to add an implementation for ```cls_cxx_list_watchers``` to https://github.com/ceph/ceph/blob/master/src/test/librados_test_stub/LibradosTestStub.cc#L1009.

dillaman:
@trociny Agree that the global image ids are not too useful.  It should be fine to just run a "cls_client::dir_list" to retrieve the image name -> image id mapping, reverse it, and translate the image id to name.

For the syncing state, I agree it would be good to provide deeper context.  Since `ImageSync` is a state machine, it should be relatively easy to update the progress.  When `ImageSync` gets to the `ObjectCopyRequest` state, that class should provide the percent complete since it will most likely be the longest running state.

src/cls/rbd/cls_rbd.cc
```diff
@@ -3676,6 +3867,127 @@ int mirror_image_remove(cls_method_context_t hctx, bufferlist *in,
...
 
+/**
+ * Input:
+ * none
+ *
+ * Output:
+ * @param std::vector<std::string>: collection of global_image_ids
+ * @returns 0 on success, negative error code on failure
+ */
+int mirror_image_status_list(cls_method_context_t hctx, bufferlist *in,
```
dillaman: Is this cls method used for anything?

dillaman:
@trociny 

> Right now "stale" and "update_time" metadata are returned in two additional params for mirror_image_status_get method. Not sure it is the best solution. They could be added to status object. Or a additional object "metadata" could be created. What do you think?

I assume this was because you didn't want to encode `stale` and `update_time` within the on-disk structure? Definitely seems like `update_time` belongs in the structure since it appears you are encoding it along with the state.  As for `stale`, you could create a derived class which is only used for the `cls_client` API which includes the state.

On a related note, for `mirror pool status`, it might be a good idea to batch up multiple `get` calls into a single op -- otherwise a pool of hundreds of images will take forever to query.  Not critical, but potentially important.

src/cls/rbd/cls_rbd.cc
trociny: It was used initially for `mirror pool status`, but is not used now. I will use your suggestion to batch up multiple get calls into a single op, and this method is most likely to be removed from the final version.

src/tools/rbd/action/MirrorPool.cc
```diff
@@ -399,6 +503,10 @@ Shell::Action action_info(
   {"mirror", "pool", "info"}, {},
   "Show information about the pool mirroring configuration.", {},
   &get_info_arguments, &execute_info);
+Shell::Action action_status(
+  {"mirror", "pool", "status"}, {},
```
jdurgin: It's probably worth having a general summary command (or option for this command), that doesn't list the details for each image, but just a high-level status like OK - images are all within some threshold, WARN if some images are out of sync by more than some threshold, and ERR if info is stale or any images are stuck (split-brain, or maybe just behind by a larger threshold)?

jdurgin: Maybe useful to have an option to specify a single image to check on as well

jdurgin: Perhaps the high-level summary could use separate class methods and avoid image name lookups to be a lot faster when there are tons of images. With batching maybe this isn't worth optimizing right now though.

trociny: @jdurgin 

> It's probably worth having a general summary command

Yes, thanks for the remainder! I had plans for the summary command. I think by default `mirror pool status` will output only summary status, and when `--detail` option is specified it will output status for every image in the pool

> Maybe useful to have an option to specify a single image to check on as well

Isn't it what `mirror image status` does?

jdurgin: On 04/05/2016 10:23 PM, Mykola Golub wrote:
> In src/tools/rbd/action/MirrorPool.cc
> <https://github.com/ceph/ceph/pull/8442#discussion_r58652430>:
>
>> @@ -399,6 +503,10 @@ Shell::Action action_info(
>>    {"mirror", "pool", "info"}, {},
>>    "Show information about the pool mirroring configuration.", {},
>>    &get_info_arguments, &execute_info);
>> +Shell::Action action_status(
>> +  {"mirror", "pool", "status"}, {},
>
> @jdurgin <https://github.com/jdurgin>
>
>     It's probably worth having a general summary command
>
> Yes, thanks for the remainder! I had plans for the summary command. I
> think by default |mirror pool status| will output only summary status,
> and when |--detail| option is specified it will output status for every
> image in the pool

Sounds good

>     Maybe useful to have an option to specify a single image to check on
>     as well
>
> Isn't it what |mirror image status| does?

Ah, yes, somehow missed that

trociny:
@dillaman @jdurgin 

The PR is updated. I still have plans to extend RBD API to include "get" and "list" status functions, but would like to know if you like cls interface, and the approach in general.

Here are some examples:
```
% rbd --cluster cluster1 -p mirror mirror pool status --verbose
health: WARNING
images: 4 total
    1 syncing
    3 replaying

test/1011515f007c:
  global_id:   e0bd1416-2b6a-4b3e-96a2-9d1630f31e29
  state:       replaying
  description: replaying, master_position=[object_number=3, tag_tid=11, entry_tid=99], mirror_position=[object_number=3, tag_tid=11, entry_tid=99]
  last_update: 2016-04-11 16:16:57.254165

test2/101812200854:
  global_id:   ad0f6d02-3baf-409c-b5c3-c175f2fe15a6
  state:       syncing
  description: bootstrapping, IMAGE_SYNC/COPY_SNAPSHOTS
  last_update: 2016-04-11 16:17:09.272254

test1/1018515f007c:
  global_id:   11340f3c-aec5-46fc-b5c9-d0be863e5c41
  state:       replaying
  description: replaying, master_position=[object_number=3, tag_tid=2, entry_tid=3], mirror_position=[object_number=3, tag_tid=2, entry_tid=3]
  last_update: 2016-04-11 16:16:57.484031

test3/10185bd062c2:
  global_id:   565feb28-2cc5-42ef-9c42-482882554479
  state:       replaying
  description: replaying, master_position=[object_number=3, tag_tid=1, entry_tid=3], mirror_position=[]
  last_update: 2016-04-11 16:17:09.397410

% rbd --cluster cluster2 -p mirror mirror pool status --verbose
health: OK
images: 4 total
    4 stopped

test/10102ae8944a:
  global_id:   e0bd1416-2b6a-4b3e-96a2-9d1630f31e29
  state:       stopped
  description: remote image is non-primary or local image is primary
  last_update: 2016-04-11 16:18:04.286298
...

% rbd --cluster cluster1 -p mirror mirror image status test     
test/1011515f007c:
  global_id:   e0bd1416-2b6a-4b3e-96a2-9d1630f31e29
  state:       replaying
  description: replaying, master_position=[object_number=3, tag_tid=11, entry_tid=99], mirror_position=[object_number=3, tag_tid=11, entry_tid=99]
  last_update: 2016-04-11 16:18:57.254935
```

So, `mirror pool status` outputs summary, and (if --verbose option is specified) detailed status. 
`mirror image status` outputs status for one image.

Questions I have:

- How do you like cls interface? Particularly cls_client::mirror_image_status_list().
- Is "stale" name is good? May be "offline"/"online", or some other?
- Format for status output:
```
<image_name>/<image_id>:
  global_id: <global_image_id>
  ...
```
may be change to
```
<image_name>/<image_id>/<global_image_id>:
  ...
```
The last one is easier to grep, but for parsing I would use json or xml format, while the plain format should be the most human readable, and the first variant looks better for me from this POV.

dillaman:
@trociny Looking good.

Now that you bring it up, I guess "stale" could get confusing when you think of it in terms of the contents of the image (i.e. the async playback will usually result in a "stale" image until the journal entries catch up).  Given that, perhaps online/offline would be better (or re-use up/down terminology from OSD status).

Not sure how I feel about the `pair<cls::rbd::MirrorImageStatus, bool>` interface to the cls method -- it might make it harder to extend in the future if needed.  It might just be easier to add the stale state to the struct, store it on disk, but overwrite it within the class methods before sending it back to the client.

For the `status_get_summary` cls interface, if the records are stale, perhaps it just makes sense to mark those images as the `UNKNOWN` state.  That would eliminate the need to pass two maps and probably better aligns to the true state of the images (i.e. we don't know since it is stale). 

With regard to the format of the status output, I would actually vote for moving the image id down to its own line prefixed by "id:".  To the casual user, the image id doesn't mean much (we only semi-expose it via the 'rbd info' block_name_prefix anyway).

trociny:
@dillaman I was unwilling to move "stale" to the status struct due to its nature, but looking how much pain it causes to pass it around I tend to agree that it is not so bad. Will do this.

As for moving image id down in the status output, I used it as a key because was not sure if cases were possible when image name for an image_id does not exists (not found). I guess it is possible only due to a bug, and can be ignored, so I will move the id down as you suggest. 

Also agree about `status_get_summary` change.

trociny:
@dillaman rebased after adressing your comments. Now the output looks like below:
```
test:
  id:          1014515f007c
  global_id:   c5770f73-a939-481f-a70a-c2fa99f0382d
  state:       up+replaying
  description: replaying, master_position=[object_number=3, tag_tid=11, entry_tid=99], mirror_position=[object_number=3, tag_tid=11, entry_tid=99]
  last_update: 2016-04-12 15:05:59.170998
```
```
test:
  id:          1014515f007c
  global_id:   c5770f73-a939-481f-a70a-c2fa99f0382d
  state:       down+stopped
  description: stopped
  last_update: 2016-04-12 15:46:24.874658
```

It still lacks RBD API, because I get stuck thinking about a good interface.

At first I thought to hide status implementation and use only methods to retrieve status details:
```
typedef void *rbd_mirror_image_status_t;
int rbd_mirror_image_get_status(rbd_image_t image, rbd_mirror_image_status_t *mirror_image_status);
void rbd_mirror_image_status_free(rbd_mirror_image_status_t mirror_image_status);
int rbd_mirror_image_status_global_image_id(rbd_mirror_image_status_t mirror_image_status, char* global_image_id, size_t *maxlen);
void rbd_mirror_image_status_state(rbd_mirror_image_status_t mirror_image_status, rbd_mirror_image_status_state_t *state);
int rbd_mirror_image_status_state_string(rbd_mirror_image_status_t mirror_image_status, char* state, size_t *maxlen);
int rbd_mirror_image_status_description(rbd_mirror_image_status_t mirror_image_status, char* description, size_t *maxlen);
void rbd_mirror_image_status_last_update(rbd_mirror_image_status_t mirror_image_status, time_t *last_update);
bool rbd_mirror_image_status_is_up(rbd_mirror_image_status_t mirror_image_status);
bool rbd_mirror_image_status_is_down(rbd_mirror_image_status_t mirror_image_status);
```
But looking at rbd_mirror_image_* functions we already have, it would differ from those, so making it similar to what we already have it could be:
```
typedef struct {
  rbd_mirror_image_status_state_t state;
  char *description;
  time_t last_update;
  uint64_t flags;
} rbd_mirror_image_status_t;

int rbd_mirror_image_get_status(rbd_image_t image, rbd_mirror_image_status_t *mirror_image_status, size_t status_size);
int rbd_mirror_image_status_list(const char *start, size_t max, char *image_names, rbd_mirror_image_info_t *images, rbd_mirror_image_status_t *image_statuses, size_t *len);
int rbd_mirror_image_status_summary(rbd_mirror_image_status_state_t *states, int *counts, size_t *maxlen);
```
But this would mean more manual work on user side...

src/cls/rbd/cls_rbd_types.h
```diff
@@ -73,13 +74,81 @@ struct MirrorImage {
...
+  MIRROR_IMAGE_STATUS_STATE_SYNCING         = 2,
+  MIRROR_IMAGE_STATUS_STATE_STARTING_REPLAY = 3,
+  MIRROR_IMAGE_STATUS_STATE_REPLAYING       = 4,
+  MIRROR_IMAGE_STATUS_STATE_STOPPING_REPLAY = 5,
+  MIRROR_IMAGE_STATUS_STATE_STOPPED         = 6,
+};
+
+enum MirrorImageStatusFlag {
+  MIRROR_IMAGE_STATUS_FLAG_UP               = 1<<0,
+  MIRROR_IMAGE_STATUS_FLAG_DOWN             = 1<<1,
```
dillaman: Do we need the second flag?  If it's not up, it should be considered down.

src/cls/rbd/cls_rbd.cc
```diff
@@ -3162,6 +3174,299 @@ int image_remove(cls_method_context_t hctx, const string &image_id) {
...
+    decode(ondisk_status, it);
+  } catch (const buffer::error &err) {
+    CLS_ERR("could not decode status for mirrored image, global id '%s'",
+	    global_image_id.c_str());
+    return -EIO;
+  }
+
+  obj_list_watch_response_t watchers;
+  r = cls_cxx_list_watchers(hctx, &watchers);
+  if (r < 0) {
```
dillaman: Should this just be ```if (r < 0 && r != -ENOENT)```?

src/cls/rbd/cls_rbd.cc
```diff
@@ -3162,6 +3174,299 @@ int image_remove(cls_method_context_t hctx, const string &image_id) {
...
+      }
+
+      if (watchers.find(status.origin) == watchers.end()) {
+	CLS_LOG(20, "removing stale status object for key %s",
+		key.c_str());
+	int r1 = cls_cxx_map_remove_key(hctx, key);
+	if (r1 < 0) {
+	  CLS_ERR("error removing stale status for key '%s': %s",
+		  key.c_str(), cpp_strerror(r1).c_str());
+	  return -r1;
```
dillaman: ```return r1```

src/cls/rbd/cls_rbd.cc
```diff
@@ -3676,6 +3981,157 @@ int mirror_image_remove(cls_method_context_t hctx, bufferlist *in,
...
+    return r;
+  }
+
+  ::encode(states, *out);
+  return 0;
+}
+
+/**
+ * Input:
+ *
```
dillaman: Minor: missing "none"

src/cls/rbd/cls_rbd_types.cc
```diff
@@ -127,5 +132,92 @@ std::ostream& operator<<(std::ostream& os, const MirrorImage& mirror_image) {
...
+    os << "replaying";
+    break;
+  case MIRROR_IMAGE_STATUS_STATE_STOPPING_REPLAY:
+    os << "stopping_replay";
+    break;
+  case MIRROR_IMAGE_STATUS_STATE_STOPPED:
+    os << "stopped";
+    break;
+  default:
+    os << "invalid (" << static_cast<uint32_t>(state) << ")";
```
dillaman: Minor: can this be "unknown" instead of "invalid"?  If we ever expand the states, an older client would interpret the new state as "invalid" which might be alarming.

src/test/cli/rbd/help.t
```diff
@@ -48,13 +48,15 @@
       mirror image enable         Enable RBD mirroring for an image.
       mirror image promote        Promote an image to primary for RBD mirroring.
       mirror image resync         Force resync to primary image for RBD mirroring.
+      mirror image status         Show RDB mirroring status for image.
```
dillaman: Minor: s/for image/for an image/g

src/tools/rbd/action/MirrorImage.cc
```diff
@@ -187,6 +189,85 @@ int execute_resync(const po::variables_map &vm) {
...
+						  &status);
+  if (r < 0) {
+    std::cerr << "rbd: failed to get status for image " << image_name << ": "
+	      << cpp_strerror(r) << std::endl;
+    return r;
+  }
+
+  if (formatter != nullptr) {
+    formatter->open_object_section("image");
+    formatter->dump_string("image_name", image_name);
```
dillaman: Minor: probably safe to drop the "image_" prefix here and below since you are within an "image" object section -- implied.

src/tools/rbd/action/MirrorPool.cc
```diff
@@ -374,6 +376,159 @@ int execute_info(const po::variables_map &vm) {
...
+      }
+      for (auto it = mirror_images.begin(); it != mirror_images.end(); ++it) {
+	const std::string &image_id = it->first;
+	std::string &global_image_id = it->second.global_image_id;
+	cls::rbd::MirrorImageStatus &status =
+	  statuses.find(image_id) != statuses.end() ?
+	  statuses[image_id] : unknown_status;
+
+	std::string image_name;
+	r = librbd::cls_client::dir_get_name(&io_ctx, RBD_DIRECTORY, image_id,
```
dillaman: I would load the image directory once before the start of the `do` loop so that you don't need to have individual round-trips to the OSD to convert the image id to image name.

src/tools/rbd/action/MirrorPool.cc
```diff
@@ -374,6 +376,159 @@ int execute_info(const po::variables_map &vm) {
...
+					     &image_name);
+	if (r < 0) {
+	  std::cerr << "rbd: failed to get name for image " << image_id << ": "
+		    << cpp_strerror(r) << std::endl;
+	}
+
+	if (formatter != nullptr) {
+	  formatter->open_object_section("image");
+	  formatter->dump_string("image_name", image_name);
```
dillaman: Minor: same comment as above -- the "image_" prefix seems semi-redundant within the "image" object section.

src/tools/rbd_mirror/ImageReplayer.cc
```diff
@@ -874,6 +907,142 @@ void ImageReplayer<I>::shut_down_journal_replay(bool cancel_ops)
...
+
+  // TODO: make status update interval configurable.
+  m_threads->timer->add_event_after(30, m_update_status_task);
+}
+
+template <typename I>
+void ImageReplayer<I>::start_update_status_task()
+{
+  FunctionContext *ctx = new FunctionContext(
+    [this](int r) {
```
dillaman: If  you take ```m_threads->timer_lock``` here, you could set ```m_update_status_task = nullptr;``` here and then eliminate the ```bool from_timer``` argument to ```reschedule_update_status_task``` and ```update_mirror_image_status```.

src/tools/rbd_mirror/ImageSync.cc
```diff
@@ -157,6 +166,8 @@ void ImageSync<I>::send_copy_image() {
     return;
   }
 
+  update_progress("COPY_IMAGE");
```
dillaman: It would be nice to pass a wrapped ```ProgressContext``` to the ```ImageCopyRequest``` so it can update on the actual image copy (which might take a while).  It should be throttled -- so perhaps the ```ProgressContext``` callback should take a boolean argument for "should_throttle" (or similar).

src/tools/rbd_mirror/Replayer.cc
```diff
@@ -113,6 +113,33 @@ class ReplayerAdminSocketHook : public AdminSocketHook {
   Commands commands;
 };
 
+class MirrorStatusWatchCtx : public librados::WatchCtx2 {
```
dillaman: What is this class used for?  I don't believe you need to be watching an object in order to send it notifications.

dillaman: Never mind -- forgot about the "up/down" logic

dillaman: Probably should use the ```librbd::ObjectWatcher``` since it handles error recovery.

src/tools/rbd_mirror/Replayer.cc
```diff
@@ -383,6 +419,50 @@ void Replayer::set_sources(const PoolImageIds &pool_image_ids)
...
+  uint64_t instance_id = librados::Rados(ioctx).get_instance_id();
+
+  dout(20) << "pool_id=" << pool_id << ", instance_id=" << instance_id << dendl;
+
+  librados::ObjectWriteOperation op;
+  librbd::cls_client::mirror_image_status_remove_down(&op);
+
+  librados::AioCompletion *comp = librados::Rados::aio_create_completion();
+
+  int r = ioctx.aio_operate(RBD_MIRRORING, comp, &op);
```
dillaman: ```ioctx.operate``` vs ```ioctx.aio_operate```

src/tools/rbd_mirror/image_replayer/BootstrapRequest.cc
```diff
@@ -95,19 +96,23 @@ BootstrapRequest<I>::BootstrapRequest(librados::IoCtx &local_io_ctx,
...
-    m_client_meta(client_meta), m_on_finish(on_finish) {
+    m_client_meta(client_meta), m_on_finish(on_finish),
+    m_progress_ctx(progress_ctx) {
 }
 
 template <typename I>
 BootstrapRequest<I>::~BootstrapRequest() {
   assert(m_remote_image_ctx == nullptr);
+
+  delete(m_progress_ctx);
```
dillaman: If ```BootstrapRequest``` is going to own the progress context pointer, maybe move the ```BootstrapProgressContext``` class here and construct it here.

src/tools/rbd_mirror/ImageReplayer.cc
```diff
@@ -874,6 +907,142 @@ void ImageReplayer<I>::shut_down_journal_replay(bool cancel_ops)
...
+    default:
+      assert(!"invalid state");
+    }
+  }
+
+  dout(20) << "status=" << status << dendl;
+
+  librados::ObjectWriteOperation op;
+  librbd::cls_client::mirror_image_status_set(&op, m_global_image_id, status);
+  librados::AioCompletion *comp = librados::Rados::aio_create_completion();
```
dillaman: Perhaps it might be a good idea to attach a callback to this to act as a rate limiter (don't send another update until this update is ACKed) and to ensure we don't have any pending status updates before destructing the object.

src/cls/rbd/cls_rbd_types.h
trociny: Actually, it can be up, down or unknown. That is why I decided to have two flags. Not sure how useful it is, it was expected to be used only internally, but right now does not look very useful. In 'RBD API', I use bool:
```
  typedef struct {
    mirror_image_status_state_t state;
    std::string description;
    time_t last_update;
    bool up;
  } mirror_image_status_t;
```
So may be use bool instead of flags for internal structure too?

dillaman: How do you get to the unknown status state?

src/cls/rbd/cls_rbd.cc
trociny: yes. fixed

src/cls/rbd/cls_rbd_types.cc
trociny: We already have "unknown" as a valid state, and I did not want to use the same word for different meanings. Now, thinking about it, if I change "invalid" to "unknown" it may still be ok, because it is still recognizable: for MIRROR_IMAGE_STATUS_STATE_UNKNOWN we will have just "unknown", and for the default: "unknown (XXX)".

Or may be rename MIRROR_IMAGE_STATUS_STATE_UNKNOWN state (to what)?

dillaman: Yeah -- I was guessing that was the goal.  I think "unknown" and "unknown (xyz)" is enough to distinguish the states.

src/cls/rbd/cls_rbd_types.h
trociny: Well, when the status is stored on disk, it is stored with no any flag set. And until you compare the status origin with current watchers you don't know if it is up or down. When a flag (UP or DOWN) is set you can be sure, that it has been checked against running watchers. If you remove DOWN flag, and the status comes with no UP, it would mean either it is DOWN or it has not been checked at all. Actually right now it is mostly useful for debugging and testing only, so I don't mind removing this. What do you prefer, a bool or flags with one flag (currently)?

dillaman: Probably better to have this and the API match.  The only future-proofing that I can think of right now is for when we add horizontal scaling to rbd-mirroring.  If we have an orderly transition of one image from rbd-mirror instance A to instance B, it might be nice to avoid showing a warning.  However, I don't think adding a flag will help us in that situation.

src/tools/rbd_mirror/image_replayer/BootstrapRequest.cc
trociny: But then I will need to pass ImageReplayer as an argument to BootstrapRequest, to be able to construct BootstrapRequest?

dillaman: I was thinking ```ImageReplayer``` would have a ```ImageReplayerProgressContext``` proxy which it passes to ```BootstrapRequest``` as a ```ProgressContext*``` (owned by ImageReplayer -- could even be a member variable of ImageReplayer since the lifetime is intertwined).

src/tools/rbd_mirror/ImageReplayer.cc
trociny: Right now I use aio_flush before destructing to ensue we don't have pending updates. I didn't think about a rate limiter because amount of updates until the replayer went to replay state was always a finite number, and after this it was updated with 30 sec period.

Now, when we want to update progress of ImageCopyRequest, the number of updates becomes unknown, so I will try your suggestion.

trociny:
@dillaman I think I have addressed all your comments.

trociny:
@dillaman I have also added displaying "number of entries behind master" in status.

src/include/rbd/librbd.h
```diff
@@ -208,6 +225,12 @@ CEPH_RBD_API int rbd_mirror_peer_set_client(rados_ioctx_t io_ctx,
 CEPH_RBD_API int rbd_mirror_peer_set_cluster(rados_ioctx_t io_ctx,
                                              const char *uuid,
                                              const char *cluster_name);
+CEPH_RBD_API int rbd_mirror_image_status_list(rados_ioctx_t io_ctx,
+    const char *start, size_t max, char **image_names,
```
dillaman: Is it ```image_names``` or ```image_ids```?

dillaman: Never mind -- see that it converts ids to names within the API.

src/tools/rbd_mirror/ImageReplayer.cc
```diff
@@ -176,6 +176,225 @@ void ImageReplayer<I>::BootstrapProgressContext::update_progress(
...
+    on_finish->complete(-EAGAIN);
+    return;
+  }
+
+  m_on_finish = on_finish;
+  get_registered_clients();
+}
+
+template <typename I>
+void ImageReplayer<I>::ReplayStatusMaker::get_registered_clients() {
```
dillaman: Do you need to query the registered clients?  They should be auto-updating whenever a journal header update occurs.

src/tools/rbd_mirror/ImageReplayer.h
```diff
@@ -195,6 +196,35 @@ class ImageReplayer {
     ImageReplayer<ImageCtxT> *replayer;
   };
 
+  struct ReplayStatusMaker {
```
dillaman: Super nitpicks: (1) it might be nice to break this out into the "image_replayer" namespace since it is small, doesn't need to depend on accessing ImageReplayer private state (at least that I could see), and could allow unit testing (in the future) -- (2) the "Maker" suffix doesn't sit right with me for some reason, perhaps "Generator" or "Formatter"?

dillaman:
@trociny Looks good to me minus the couple questions/nits.

src/tools/rbd_mirror/ImageReplayer.cc
trociny: This was I tried first (the cached clients).  But I got strange results, I don't recall exactly, it looked like data were outdated or master position was behind mirror, so entries_behind_master could be negative. I did not investigate that much though (anyway I needed to query osd for other data). If you think it is unexpected and worth further investigation I will do.

src/tools/rbd_mirror/ImageReplayer.h
trociny: Actually it accesses private m_remote_journaler and m_local_mirror_uuid. Sure it can be resolved, so I will move it out into the "image_replayer".

I don't like "Maker" very much too, will change.

src/tools/rbd_mirror/ImageReplayer.cc
trociny: @dillaman Looking at it again, actually querying the registered clients instead of using the cached ones gives the same results and does not resolve possible negative entries_behind_master:
```
2016-04-26 09:15:12.180129 7f5e2effd700 20 rbd-mirror: ImageReplayer[3/103b74b0dc51]::ReplayStatusMaker::make_description: m_master_position=[object_number=3, tag_tid=2, entry_tid=3], m_mirror_position=[object_number=3, tag_tid=2, entry_tid=3], m_entries_behind_master=0
2016-04-26 09:15:42.181220 7f5e2effd700 20 rbd-mirror: ImageReplayer[3/103b74b0dc51]::ReplayStatusMaker::make_description: m_master_position=[object_number=3, tag_tid=3, entry_tid=859], m_mirror_position=[object_number=3, tag_tid=3, entry_tid=1471], m_entries_behind_master=-612
2016-04-26 09:16:12.182310 7f5e2effd700 20 rbd-mirror: ImageReplayer[3/103b74b0dc51]::ReplayStatusMaker::make_description: m_master_position=[object_number=2, tag_tid=3, entry_tid=2702], m_mirror_position=[object_number=3, tag_tid=3, entry_tid=3327], m_entries_behind_master=-625
2016-04-26 09:16:30.526062 7f5e2effd700 20 rbd-mirror: ImageReplayer[3/103b74b0dc51]::ReplayStatusMaker::make_description: m_master_position=[object_number=3, tag_tid=3, entry_tid=4319], m_mirror_position=[object_number=3, tag_tid=3, entry_tid=4319], m_entries_behind_master=0
2016-04-26 09:17:00.527246 7f5e2effd700 20 rbd-mirror: ImageReplayer[3/103b74b0dc51]::ReplayStatusMaker::make_description: m_master_position=[object_number=2, tag_tid=4, entry_tid=790], m_mirror_position=[object_number=3, tag_tid=4, entry_tid=1503], m_entries_behind_master=-713
2016-04-26 09:17:30.528257 7f5e2effd700 20 rbd-mirror: ImageReplayer[3/103b74b0dc51]::ReplayStatusMaker::make_description: m_master_position=[object_number=2, tag_tid=4, entry_tid=2734], m_mirror_position=[object_number=3, tag_tid=4, entry_tid=3391], m_entries_behind_master=-657
2016-04-26 09:17:48.072964 7f5e2effd700 20 rbd-mirror: ImageReplayer[3/103b74b0dc51]::ReplayStatusMaker::make_description: m_master_position=[object_number=3, tag_tid=4, entry_tid=4319], m_mirror_position=[object_number=3, tag_tid=4, entry_tid=4319], m_entries_behind_master=0
2016-04-26 09:18:18.074030 7f5e2effd700 20 rbd-mirror: ImageReplayer[3/103b74b0dc51]::ReplayStatusMaker::make_description: m_master_position=[object_number=3, tag_tid=5, entry_tid=855], m_mirror_position=[object_number=3, tag_tid=5, entry_tid=1663], m_entries_behind_master=-808
2016-04-26 09:18:48.086746 7f5e2effd700 20 rbd-mirror: ImageReplayer[3/103b74b0dc51]::ReplayStatusMaker::make_description: m_master_position=[object_number=3, tag_tid=5, entry_tid=3003], m_mirror_position=[object_number=3, tag_tid=5, entry_tid=3359], m_entries_behind_master=-356
2016-04-26 09:19:18.088519 7f5e2effd700 20 rbd-mirror: ImageReplayer[3/103b74b0dc51]::ReplayStatusMaker::make_description: m_master_position=[object_number=3, tag_tid=5, entry_tid=4319], m_mirror_position=[object_number=3, tag_tid=5, entry_tid=4319], m_entries_behind_master=0
2016-04-26 09:19:22.354803 7f5e2effd700 20 rbd-mirror: ImageReplayer[3/103b74b0dc51]::ReplayStatusMaker::make_description: m_master_position=[object_number=3, tag_tid=5, entry_tid=4319], m_mirror_position=[object_number=3, tag_tid=5, entry_tid=4319], m_entries_behind_master=0
2016-04-26 09:19:52.355642 7f5e2effd700 20 rbd-mirror: ImageReplayer[3/103b74b0dc51]::ReplayStatusMaker::make_description: m_master_position=[object_number=0, tag_tid=6, entry_tid=1204], m_mirror_position=[object_number=3, tag_tid=6, entry_tid=1727], m_entries_behind_master=-523
2016-04-26 09:20:25.273524 7f5e2effd700 20 rbd-mirror: ImageReplayer[3/103b74b0dc51]::ReplayStatusMaker::make_description: m_master_position=[object_number=7, tag_tid=6, entry_tid=4319], m_mirror_position=[object_number=0, tag_tid=6, entry_tid=2140], m_entries_behind_master=2179
2016-04-26 09:20:55.274942 7f5e2effd700 20 rbd-mirror: ImageReplayer[3/103b74b0dc51]::ReplayStatusMaker::make_description: m_master_position=[object_number=7, tag_tid=6, entry_tid=4319], m_mirror_position=[object_number=4, tag_tid=6, entry_tid=4316], m_entries_behind_master=3
```
 I think this is because the mirror commits its position after every entry apply, while the master client may delay.

I will use the cached clients, though there is no idea if negative entries_behind_master can be fixed/improved somehow.

dillaman: On Tue, Apr 26, 2016 at 1:38 AM, Mykola Golub <notifications@github.com>
wrote:

> I will use the cached clients, though there is no idea if negative
> entries_behind_master can be fixed/improved somehow.
>

Really?  rbd-mirror flushes the commit position after each entry is applied
or is that just a side-effect of the test script constantly flushing it?
We'll always have to deal with the possibility that the master hasn't
committed the update.  Just should guard for that possibility within the
status output (e.g. m_entries_behind_master=0 instead of
m_entries_behind_master=-612).


-- 
Jason

trociny: Ah, yes, I guess this is the result of the test script constantly flushing. The ImageReplayer just calls Journaler::committed() on every entry applied.

BTW, I have updated PR. I am not sure about guarding entries_behind_master against negative values. Outputing 0 in this case may also look confusing. Maybe don't print entries_behind_master=X in this case at all?
Or if master_position less than mirror_position set master_position to mirror?

dillaman:
@trociny Playing with it I see the following:

> $ rbd --cluster cluster2 --pool mirror mirror image status foo
> foo:
>   global_id:   dc2f0b2b-9c06-4041-9725-fded9e4e6a44
>   state:       up+replaying
>   description: replaying, master_position=[object_number=6, tag_tid=2, entry_tid=22598], mirror_position=[object_number=6, tag_tid=2, entry_tid=29598], entries_behind_master=-7000
>   last_update: 2016-04-28 18:42:39

So it would would be good to keep "entries_behind_master" >= 0.  However, even after the master commit position caught up, the journal status never updated again:

> $ rbd --cluster cluster1 --pool mirror journal status 103574b0dc51
> minimum_set: 1
> active_set: 1
> registered clients: 
> 	[id=, commit_position=[positions=[[object_number=6, tag_tid=2, entry_tid=22598], [object_number=5, tag_tid=2, entry_tid=22597], [object_number=4, tag_tid=2, entry_tid=22596], [object_number=7, tag_tid=2, entry_tid=22595]]], state=connected]
> 	[id=96464429-94eb-4f0a-bc40-90394f63f1f7, commit_position=[positions=[[object_number=6, tag_tid=2, entry_tid=29598], [object_number=5, tag_tid=2, entry_tid=29597], [object_number=4, tag_tid=2, entry_tid=29596], [object_number=7, tag_tid=2, entry_tid=29595]]], state=connected]

trociny:
@dillaman I left a note earlier about keeping "entries_behind_master" >= 0, asking for an advise. I am not very happy with the solution just to set entries_behind_master = 0 when it is negative. It will still look confusing for the user. May be just don't print entries_behind_master in this case at all?
Another option: if entries_behind_master is negative, set it to 0 and set master_position = mirror_position?

I would prefer just not printing entries_behind_master in such cases at all, still I am interested in your opinion.

As for status not being updated, did you wait for 30 seconds when rechecking the status (in replaying state it is updated in 30 sec interval)? Was 'last_update' time changed? I will try to reproduce it.

trociny:
@dillaman retested and failed to see the position getting stuck in replay status. It is updated to me (with expected up to 30 sec delay).

I have updated PR to keep "entries_behind_master" >= 0 (after thinking more about it, I think it is better than not printing it in such cases, it would unnecessary complicate monitoring scripts).

dillaman:
@trociny Yes, it was updating (didn't verify if it was every 30 seconds).  I'll try to reproduce it today and investigate:

> $ rbd --cluster cluster2 --pool mirror mirror image status foo
> foo:
>   global_id:   dc2f0b2b-9c06-4041-9725-fded9e4e6a44
>   state:       up+replaying
>   description: replaying, master_position=[object_number=6, tag_tid=2, entry_tid=22598], mirror_position=[object_number=6, tag_tid=2, entry_tid=29598], entries_behind_master=-7000
>   last_update: 2016-04-28 18:47:39

dillaman:
Merged manually to master

PR #8446, opened 2016-04-05
osd: fix/clean up full map request handling

liewegas:
- send the right number of maps from the mon
- request more maps if the osd didn't get them all
- clean up the code


tchaikov:
lgtm

PR #8452, opened 2016-04-05
pybind: fix unicode handling in CephFSVolumeClient::purge

jcsp:
os.path.join is sensitive to string encoding, but
just doing a straight substitution should not be.

### DIFF
--- a/src/pybind/ceph_volume_client.py
+++ b/src/pybind/ceph_volume_client.py
@@ -526,7 +526,10 @@ class CephFSVolumeClient(object):
             d = self.fs.readdir(dir_handle)
             while d:
                 if d.d_name not in [".", ".."]:
-                    d_full = os.path.join(root_path, d.d_name)
+                    # Do not use os.path.join because it is sensitive
+                    # to string encoding, we just pass through dnames
+                    # as byte arrays
+                    d_full = "{0}/{1}".format(root_path, d.d_name)
                     if d.is_dir():
                         rmtree(d_full)
                     else:


jcsp:
Tested by https://github.com/ceph/ceph-qa-suite/pull/934

jcsp:

jdurgin:
Looks like that run didn't include test_purge (not using the wip-15266 suite branch).

Maybe I'm missing something, but I'm not seeing how using format() handles unicode better than os.path.join - what are the types of args that fail with the latter and succeed with the former?

onyb:
Python 3 documentation suggests the following for `os.path` module:
> Applications are encouraged to represent file names as (Unicode) character strings.

In this particular case, one possible fix in PY2 could be: `d_full = os.path.join(root_path, unicode(d.d_name))`. Note that `d_full` in the above expression would also be a Unicode literal.

I hope things will be simpler after we have a compatibility layer (for PY2 and PY3 dual support), during Google Summer of Code.

jcsp:
@jdurgin yeah, this was a weird one. I think Manila is giving us share IDs (which are very much ascii) as unicode python objects.

```
>>> os.path.join(u"/foo", chr(0xc3))
  File "<stdin>", line 1, in <module>
  File "/usr/lib64/python2.7/posixpath.py", line 73, in join
    path += '/' + b
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 1: ordinal not in range(128)
>>> "{0}/{1}".format(u"/", chr(0xc3))
'//\xc3'
```

My reproducer PR wasn't actually using unicode properly so I've updated that too (can see it failing if you back out the join vs. format change)

jdurgin:
Ah, ran into similar passing of unicode from glance, cinder, and nova - since we didn't have a high-level library like this we ended up forcing everything to utf-8-encoded string types before passing to librbd. Of course, nothing actually used non-ascii chars for e.g. uuids, so it was just a nuisance, but allowed it to work with existing versions of the python bindings.

jcsp:
Passing test run here: http://pulpito.ceph.com/jspray-2016-05-12_04:34:36-fs-wip-jcsp-testing-20160511---basic-smithi/187784

Any objections to merging?

jdurgin:
lgtm fwiw

PR #8454, opened 2016-04-05
rbd_mirror.sh: add env vars to allow running under teuthology

jdurgin:
This works in both a dev environment and in teuthology, with
the teuthology and ceph-qa-suite branches wip-multi-cluster.

### DIFF
--- a/qa/workunits/rbd/rbd_mirror.sh
+++ b/qa/workunits/rbd/rbd_mirror.sh
@@ -66,6 +66,15 @@ POOL=mirror
 SRC_DIR=$(readlink -f $(dirname $0)/../../../src)
 TEMPDIR=
 
+# These vars facilitate running this script in an environment with
+# ceph installed from packages, like teuthology. These are not defined
+# by default.
+#
+# RBD_MIRROR_USE_EXISTING_CLUSTER - if set, do not start and stop ceph clusters
+# RBD_MIRROR_USE_EXISTING_DAEMON - if set, use an existing instance of rbd-mirror
+#                                  running as ceph client $CEPH_ID. If empty,
+#                                  this script will start and stop rbd-mirror
+
 #
 # Functions
 #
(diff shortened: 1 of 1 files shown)


jdurgin:
See also: https://github.com/ceph/teuthology/pull/841 https://github.com/ceph/ceph-qa-suite/pull/937

jdurgin:
@trociny any comments on this?

qa/workunits/rbd/rbd_mirror.sh
```diff
@@ -167,22 +168,14 @@ stop_mirror()
...
     local daemon=$1
-    local pid_file=$(daemon_pid_file $daemon)
-
-    pid=$(cat $(daemon_pid_file "${daemon}"))
-
-    test -n "${pid}"
-
-    ps auxww | awk -v pid=${pid} '$2 == pid' |
-	sed -nEe 's/^.*--cluster +([^ ]+).*$/\1/p'
+    echo "${daemon}" | sed s/_daemon$//
```
trociny: You could use:
```
echo ${daemon%_daemon}
```

But actually, although I like the idea to make automatic detection, I don't like very much that you relay on daemon name being "${CLUSTER}_daemon", while CLUSTER{X}_DAEMON vars are defined in the header section, so a user could change them for her purposes, e.g. set `CLUSTER1_DAEMON=primary` , and it will break the script.

Thus I would remove CLUSTERXX_DAEMON vars then, so they are built by the script internally and users don't have an impression that daemon name can be redefined. And in functions, instead of passing `daemon` arg, you could pass `cluster` arg and `daemon` var is built locally as ${cluster}_daemon.

qa/workunits/rbd/rbd_mirror.sh
```diff
@@ -75,7 +75,7 @@ TEMPDIR=
 daemon_asok_file()
 {
     local daemon=$1
-    local cluster=$2
+    local cluster=$(daemon_local_cluster "${daemon}")
```
trociny: I thing this is not right. The daemon have two asok files -- for local and remote cluster. It looks like with you change `daemon_asok_file` will generate path to local asok only?

qa/workunits/rbd/rbd_mirror.sh
```diff
@@ -68,6 +68,16 @@ POOL=mirror
...
+# These vars facilitate running this script in an environment with
+# ceph installed from packages, like teuthology. These are not defined
+# by default.
+#
+# USE_EXISTING_CLUSTER - if set, do not start and stop ceph clusters
+# UNMANAGED_RBD_MIRROR - if set, do not start and stop rbd-mirror
+# CEPH_ID              - id portion of client.$id ceph user that rbd-mirror uses
+#                        this is assumed to be the same for all clusters, and is
+#                        only used if UNMANAGED_RBD_MIRROR is set
+
```
trociny: I would prefer such vars follow the convention to have RBD_MIRROR_ prefix. Or do they are used by teuthology not only for rbd_mirror? Also, if CEPH_ID is used only together with  UNMANAGED_RBD_MIRROR, may be we can combain them? I.e. what about these?
```
RBD_MIRROR_USE_EXISTING_CLUSTER = empty|non-empty
RBD_MIRROR_USE_RBD_MIRROR = empty|${CEPH_ID}
```

jdurgin: CEPH_ID is set for all workunits by teuthology; I'll add the prefix to the others

jdurgin:
@trociny how does this look now?

trociny:
@jdurgin lgtm

jdurgin:
merged manually

PR #8457, opened 2016-04-05
build: Respect TMPDIR for virtualenv.

robbat2:
Gentoo's normal build process uses a sandbox to catch writes outside the
build environment; this includes providing a value other than /tmp for
TMPDIR. Use TMPDIR by default for CEPH_BUILD_VIRTUALENV.


dachary:
@robbat2 I assume TMPDIR is defined by gentoo, right ?

robbat2:
Yes, eg (as I have avahi rebuilding on my desktop right now)
`TMPDIR=/dev/shm/portage/net-dns/avahi-0.6.32/temp`

PR #8458, opened 2016-04-05
common/config_opts: remove invalid config options of bluestore

jjhuo:
Those invalid options are probably leftover from newstore,
and are not used anymore.

Signed-off-by: Jianjian Huo <jianjian.huo@ssi.samsung.com>


tchaikov:
lgtm.

PR #8460, opened 2016-04-05
rgw: nfs fixes

mattbenjamin:
This change fixes 2 outstanding RGW-NFS bugs, including an important change to object permission handling.

@cbodley @oritwas please check?

Thanks,

Matt


mattbenjamin:
With this change, I got a successful rsync from nfs-ganesha.  I got some messages to stderr, as well, so still chasing it down--but it's worth merging this.

oritwas:
ACK

PR #8461, opened 2016-04-05
build: change ccache invocation to path style

robbat2:
If ccache is invoked as two arguments eg CXX='ccache g++', then pybind
linking can fail because of long-standing distutils bug 8027. Some
distributions patch distutils for this, but it's not consistent.

The alternative way to use ccache is by altering PATH, but it needs
slight care because the actual directory for the ccache symlinks is not
consistent between distributions.

X-Distutils-Bug-URL: https://bugs.python.org/issue8027

### DIFF
--- a/make-debs.sh
+++ b/make-debs.sh
@@ -84,7 +84,12 @@ fi
 if test $NPROC -gt 1 ; then
     j=-j${NPROC}
 fi
-PATH=/usr/lib/ccache:$PATH dpkg-buildpackage $j -uc -us
+CCACHE_PATH=$PATH
+for d in /usr/{lib64,lib,lib32,libexec}/ccache{,/bin} ; do
+    test -d $d && test -x $d/g++ && CCACHE_PATH=$d:$PATH && break
+done
+
+PATH=${CCACHE_PATH} dpkg-buildpackage $j -uc -us
 cd ../..
 mkdir -p $codename/conf
 cat > $codename/conf/distributions <<EOF
--- a/run-make-check.sh
+++ b/run-make-check.sh
@@ -65,9 +65,13 @@ function run() {
     if test -f ./install-deps.sh ; then
 	$DRY_RUN ./install-deps.sh || return 1
     fi
-    $DRY_RUN ./autogen.sh || return 1
+    CCACHE_PATH=$PATH
+    for d in /usr/{lib64,lib,lib32,libexec}/ccache{,/bin} ; do
+        test -d $d && test -x $d/g++ && CCACHE_PATH=$d:$PATH && break
+    done
+    $DRY_RUN ./autogen.sh PATH=$CCACHE_PATH || return 1
     $DRY_RUN ./configure "$@"  --with-librocksdb-static --disable-static --with-radosgw --with-debug --without-lttng \
-        CC="ccache gcc" CXX="ccache g++" CFLAGS="-Wall -g" CXXFLAGS="-Wall -g" || return 1
+        PATH=$CCACHE_PATH CFLAGS="-Wall -g" CXXFLAGS="-Wall -g" || return 1
     $DRY_RUN make $BUILD_MAKEOPTS || return 1
     $DRY_RUN make $CHECK_MAKEOPTS check || return 1
     $DRY_RUN make dist || return 1


robbat2:
@joshd ping, patch to implement the ccache change we discussed on 2016/04/04.

jdurgin:
lgtm, just tried the same thing with the gitbuilders and it worked: https://github.com/ceph/autobuild-ceph/pull/72

tchaikov:
@dmick i recalled that you were mumbling in #sepia about this ccache issue couple days ago. =)

dmick:
oh *I* get it.
the distutils bug is that it picks one word out of the linker definition to use as the linker definition with          linker[i] = self.compiler_cxx[i].  The utter hack to pass over env k=v for c++ is irrelevant to us because we're not using env....but that last statement is outside the "if env", where it has no business, and it changes
linker from the default compiler setting "ccache g++" to just "ccache".  (i.e. it's not allowing for the possibility of validly having more than one word in the compiler).

It's stunning to me that this has existing for 6 years in distutils without a patch.

I guess the workaround is as good as any other solution, and less likely to hit silly errors like this, assuming the globbing is sufficient for all distros we support and that all the scripts use bash?....

tchaikov:
hi gents, i fixed it in another way, see #8523.

robbat2:
@tchaikov  the problem only occurs for C++ linking, which your patch doesn't change.

Here's the broken distutils link() function:
https://github.com/python/cpython/blob/2.7/Lib/distutils/unixccompiler.py#L152
Look at the _need_link block.

If target_lang is "c", then linker is correctly copied from linker_so/linker_exe (which you are setting via LDSHARED);
But when target_lang is "c++", the linker variable is then still overwritten with the first item from compiler_cxx :-(.

tchaikov:
@robbat2 i tested the change in my linux box, it works as expected:

before `linker[i] = self.compiler_cxx[i]`, 

- `linker` is `['ccache', '/usr/bin/gcc-6', '-shared', '-L/home/kefu/dev/ceph/build/lib', '-I/home/kefu/dev/ceph/src/include', '-std=c++11']`
- `self.compiler_cxx` is `['ccache', '/usr/bin/g++-6']`

and we just let `self.compiler_cxx[0]` to overwrite `self.linker[0]`, and it will be a no-op because `ccache` is not changed this way. so the resulting `linker` after overwriting is
```py
 ['ccache', '/usr/bin/gcc-6', '-shared', '-L/home/kefu/dev/ceph/build/lib', '-I/home/kefu/dev/ceph/src/include', '-std=c++11']
```

so i think the work-around works for us. am i missing anything?

robbat2:
I'm also juggling 3 distro's variations of that function, but ok, I suppose your method would work for CMake for now, since CMake handles the use-ccache decision on it's own, whereas for autoconf we're deciding outside.

+1 if we include both versions.

tchaikov:
@jdurgin @robbat2 i ported my fix to automake also at #8523. is under testing at gitbuilder as wip-8523.

tchaikov:
```sh
set -- CC='ccache gcc' CXX='ccache g++'
```
oh, we disabled it.

jdurgin:
somewhat related, C++ isn't even necessary to compile the python bindings: https://github.com/ceph/ceph/pull/8381

jdurgin:
@tchaikov enabled ccache again on the trusty tarball gb (possibly the only one where it was working before) and pushed your fix as wip-ccache-pybind (building here: http://gitbuilder.sepia.ceph.com/gitbuilder-ceph-tarball-trusty-amd64-basic/log.cgi?log=2be6017c95f9f5f6509164f3a83b0e0c2ca93fbf)

tchaikov:
@jdurgin thanks it builds fine. i am rebasing my changes against the master to include the bash/dash fix (PR #8521).

tchaikov:
build and test fine, but it failed due to
```
error: Added files:
+ cat .git/added-files
src/0
```

"forcerebuild" to see if it is reproducible.

jdurgin:
@tchaikov that error is present without the patch too, saw it earlier today on master

tchaikov:
oh, right. i see it in http://gitbuilder.sepia.ceph.com/gitbuilder-ceph-tarball-trusty-amd64-basic/log.cgi?log=18a5fddb90d180315e1817c0f7341101636408b2 now

that's embarrassing... i will try "make check" in my ubuntu 14.04 docker.

jdurgin:
the 'added files' check is only run by the tarball gitbuilders afaik, not part of 'make check'

tchaikov:
oh, right! got it.

tchaikov:
@jdurgin it does run "make check", see https://github.com/ceph/autobuild-ceph/blob/master/build-ceph.sh#L101

jdurgin:
@tchaikov yeah it runs make check first, but the 'added files' part is unique to that script https://github.com/ceph/autobuild-ceph/blob/master/build-ceph.sh#L127

tchaikov:
oh, okay.

tchaikov:
turns out a bug in testing, fixed at #8574. and it's being tested at http://gitbuilder.sepia.ceph.com/gitbuilder-ceph-tarball-trusty-amd64-basic/log.cgi?log=c462eed00ea3f887d2af7de74f48d5fb992c968e .

tchaikov:
@jdurgin and @robbat2 build passed at http://gitbuilder.sepia.ceph.com/gitbuilder-ceph-tarball-trusty-amd64-basic/log.cgi?log=fc81bec888eba3be488c34508d4ae3521372b175

tchaikov:
@robbat2 can we close this PR? as we can always set CC and CXX to enable ccache?

robbat2:
Closing for now, we can revisit if the issue comes back.

PR #8466, opened 2016-04-06
Jewel 15387

ukernel:


src/common/fs_types.cc
```diff
@@ -60,11 +64,15 @@ void file_layout_t::to_legacy(ceph_file_layout *fl) const
   fl->fl_cas_hash = 0;
   fl->fl_object_stripe_unit = 0;
   fl->fl_unused = 0;
-  // in the legacy encoding, pool 0 was undefined.
-  if (pool_id >= 0)
-    fl->fl_pg_pool = pool_id;
-  else
+  if (pool_id >= 0 && pool_id < (uint32_t)-1) {
```
jcsp: In what cases do we need the to_legacy behaviour?  Is there something client side that relies on getting 0s here sometimes?

jcsp:
See https://github.com/ceph/ceph/pull/8455/files

PR #8469, opened 2016-04-06
tests: be more generous with test timeout

dachary:
When the thrasher is in action together with a validater (lockdep or
valgrind), a single test may hang for more than 360 seconds. Increase to
1200: it does not matter if the value is large, only that it prevents
the test from hanging forever.

Fixes: http://tracker.ceph.com/issues/15403

### DIFF
--- a/src/test/librados/test.h
+++ b/src/test/librados/test.h
@@ -48,7 +48,7 @@ class TestAlarm
 {
 public:
   TestAlarm() {
-    alarm(360);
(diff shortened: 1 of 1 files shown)


dachary:
@smithfarm would you have time to review this ?

smithfarm:
My understanding is that this is just taking into account that OpenStack is slower than Sepia. In that case, :+1: 

PR #8470, opened 2016-04-06
hammer: tests: be more generous with test timeout

dachary:
http://tracker.ceph.com/issues/15699

When the thrasher is in action together with a validater (lockdep or
valgrind), a single test may hang for more than 360 seconds. Increase to
1200: it does not matter if the value is large, only that it prevents
the test from hanging forever.

Fixes: http://tracker.ceph.com/issues/15403

(cherry picked from commit af89474b3fb2c4aa63680aa6b30d71fad2fdd373)

### DIFF
--- a/src/test/librados/test.h
+++ b/src/test/librados/test.h
@@ -44,7 +44,7 @@ class TestAlarm
 {
 public:
   TestAlarm() {
-    alarm(360);
+    alarm(1200);
   }
   ~TestAlarm() {
     alarm(0);


smithfarm:
Changelog:
* added backport tracker URL to commit message

smithfarm:
@dachary This PR is in the latest round of hammer-backports integration tests, which passed a rados run (the only failures are a valgrind false positive that has since been fixed by https://github.com/ceph/teuthology/pull/915 and http://tracker.ceph.com/issues/15139 which is an infrastructure issue with two of the tests) - for details, see: http://tracker.ceph.com/issues/15895#note-18

PR #8471, opened 2016-04-06
msg/async: add missing DelayedDelivery and delay injection

branch-predictor:
Delay injection was missing from a few spots, also, DelayedDelivery
was added.

Fixes: http://tracker.ceph.com/issues/15372

### DIFF
--- a/src/msg/async/AsyncConnection.cc
+++ b/src/msg/async/AsyncConnection.cc
@@ -176,9 +176,9 @@ static void alloc_aligned_buffer(bufferlist& data, unsigned len, unsigned off)
 }
 
 AsyncConnection::AsyncConnection(CephContext *cct, AsyncMessenger *m, EventCenter *c, PerfCounters *p)
-  : Connection(cct, m), async_msgr(m), logger(p), global_seq(0), connect_seq(0), peer_global_seq(0),
-    out_seq(0), ack_left(0), in_seq(0), state(STATE_NONE), state_after_send(0), sd(-1), port(-1),
-    write_lock("AsyncConnection::write_lock"), can_write(WriteStatus::NOWRITE),
+  : Connection(cct, m), delay_thread(NULL), async_msgr(m), logger(p), global_seq(0), connect_seq(0), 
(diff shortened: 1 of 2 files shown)


branch-predictor:
@yuyuyu101 @athanatos mind taking a look?

yuyuyu101:
my main concern is we need to handle deliver thread sync. Could we just use create_time_event to delay message deliver and dispatch? it will more make sense to AsyncMessenger? it will result another thread dispatch message, I think it will be problem later while I drop lock in eventcenter

branch-predictor:
At this point let's make sure it's functionally equivalent to SimpleMessenger, once that works we'll have something to revert to in case changes to make it fully async prove unreliable or otherwise buggy. Other than that, I fully agree.

src/msg/async/AsyncConnection.cc
```diff
@@ -207,6 +207,18 @@ AsyncConnection::~AsyncConnection()
...
+    delete delay_thread;
+}
+
+void AsyncConnection::maybe_start_delay_thread()
+{
+  if (!delay_thread &&
+      async_msgr->cct->_conf->ms_inject_delay_type.find(ceph_entity_type_name(peer_type)) != string::npos) {
+    lsubdout(msgr->cct, ms, 1) << "setting up a delay queue on Pipe " << this << dendl;
+    delay_thread = new DelayedDelivery(this, async_msgr);
+    delay_thread->create("ms_pipe_delay");
```
yuyuyu101: need to grep the changes to replace pipe with async

branch-predictor: Repushed with name changes only.

liewegas:
Mostly works, but my run hit two issues:

http://tracker.ceph.com/issues/15440
http://tracker.ceph.com/issues/15443

yuyuyu101:
@branch-predictor need rebase

branch-predictor:
@yuyuyu101 rebased.

liewegas:
subsumed by #8607

PR #8473, opened 2016-04-06
rbd-mirror: Delete local image mirror when remote image mirroring is disabled

rjfd:
Ticket url: http://tracker.ceph.com/issues/14421

### DIFF
--- a/src/librbd/internal.cc
+++ b/src/librbd/internal.cc
@@ -289,7 +289,8 @@ int mirror_image_enable_internal(ImageCtx *ictx) {
   return 0;
 }
 
-int mirror_image_disable_internal(ImageCtx *ictx, bool force) {
+int mirror_image_disable_internal(ImageCtx *ictx, bool force,
+                                  bool remove=true) {
   CephContext *cct = ictx->cct;
 
   cls::rbd::MirrorImage mirror_image_internal;
@@ -397,17 +398,19 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force) {
   }
 
 remove_mirroring_image:
-  r = cls_client::mirror_image_remove(&ictx->md_ctx, ictx->id);
-  if (r < 0) {
-    lderr(cct) << "failed to remove image from mirroring directory: "
-      << cpp_strerror(r) << dendl;
-    return r;
-  }
+  if (remove) {
+    r = cls_client::mirror_image_remove(&ictx->md_ctx, ictx->id);
+    if (r < 0) {
+      lderr(cct) << "failed to remove image from mirroring directory: "
+                 << cpp_strerror(r) << dendl;
+      return r;
+    }
 
-  ldout(cct, 20) << "removed image state from rbd_mirroring object" << dendl;
+    ldout(cct, 20) << "removed image state from rbd_mirroring object" << dendl;
 
-  if (is_primary) {
-    // TODO: send notification to mirroring object about update
+    if (is_primary) {
+      // TODO: send notification to mirroring object about update
+    }
   }
 
   return 0;
@@ -1980,7 +1983,8 @@ remove_mirroring_image:
     return 0;
   }
 
-  int remove(IoCtx& io_ctx, const char *imgname, ProgressContext& prog_ctx)
+  int remove(IoCtx& io_ctx, const char *imgname, ProgressContext& prog_ctx,
+             bool force)
   {
     CephContext *cct((CephContext *)io_ctx.cct());
     ldout(cct, 20) << "remove " << &io_ctx << " " << imgname << dendl;
@@ -2001,12 +2005,31 @@ remove_mirroring_image:
 
       ictx->owner_lock.get_read();
       if (ictx->exclusive_lock != nullptr) {
-        r = ictx->operations->prepare_image_update();
-        if (r < 0 || !ictx->exclusive_lock->is_lock_owner()) {
-	  lderr(cct) << "cannot obtain exclusive lock - not removing" << dendl;
-	  ictx->owner_lock.put_read();
-	  ictx->state->close();
-          return -EBUSY;
+        if (force) {
+          // releasing read lock to avoid a deadlock when upgrading to
+          // write lock in the shut_down process
+          ictx->owner_lock.put_read();
+          if (ictx->exclusive_lock != nullptr) {
+            C_SaferCond ctx;
+            ictx->exclusive_lock->shut_down(&ctx);
+            r = ctx.wait();
+            if (r < 0) {
+              lderr(cct) << "error shutting down exclusive lock"
+                         << cpp_strerror(r) << dendl;
+              ictx->state->close();
+              return r;
+            }
+            assert (ictx->exclusive_lock == nullptr);
+            ictx->owner_lock.get_read();
+          }
+        } else {
+          r = ictx->operations->prepare_image_update();
+          if (r < 0 || !ictx->exclusive_lock->is_lock_owner()) {
+	    lderr(cct) << "cannot obtain exclusive lock - not removing" << dendl;
+	    ictx->owner_lock.put_read();
+	    ictx->state->close();
+            return -EBUSY;
+          }
         }
       }
 
@@ -2049,7 +2072,7 @@ remove_mirroring_image:
       }
 
       if (!old_format) {
-        r = mirror_image_disable_internal(ictx, false);
+        r = mirror_image_disable_internal(ictx, force, !force);
         if (r < 0 && r != -EOPNOTSUPP) {
           lderr(cct) << "error disabling image mirroring: " << cpp_strerror(r)
                      << dendl;
@@ -2116,6 +2139,14 @@ remove_mirroring_image:
 		     << cpp_strerror(-r) << dendl;
         }
 	return r;
+      }
+
+      ldout(cct, 2) << "removing image from rbd_mirroring object..." << dendl;
+      r = cls_client::mirror_image_remove(&io_ctx, id);
+      if (r < 0 && r != -ENOENT) {
+        lderr(cct) << "failed to remove image from mirroring directory: "
+                   << cpp_strerror(r) << dendl;
+        return r;
       }
     }
 
--- a/src/librbd/internal.cc
+++ b/src/librbd/internal.cc
@@ -335,10 +335,6 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force,
     return r;
   }
 
-  if (!is_primary) {
-    goto remove_mirroring_image;
-  }
-
   r = MirroringWatcher<>::notify_image_updated(
     ictx->md_ctx, cls::rbd::MIRROR_IMAGE_STATE_DISABLING,
     ictx->id, mirror_image_internal.global_image_id);
@@ -397,7 +393,6 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force,
     }
   }
 
-remove_mirroring_image:
   if (remove) {
     r = cls_client::mirror_image_remove(&ictx->md_ctx, ictx->id);
     if (r < 0) {
--- a/src/librbd/internal.h
+++ b/src/librbd/internal.h
@@ -13,6 +13,7 @@
 #include "include/buffer_fwd.h"
 #include "include/rbd/librbd.hpp"
 #include "include/rbd_types.h"
+#include "librbd/parent_types.h"
 
 enum {
   l_librbd_first = 26000,
@@ -88,6 +89,7 @@ namespace librbd {
   bool image_options_is_empty(rbd_image_options_t opts);
 
   int snap_set(ImageCtx *ictx, const char *snap_name);
+
   int list(librados::IoCtx& io_ctx, std::vector<std::string>& names);
   int list_children(ImageCtx *ictx,
 		    std::set<std::pair<std::string, std::string> > & names);
@@ -125,7 +127,7 @@ namespace librbd {
   int is_exclusive_lock_owner(ImageCtx *ictx, bool *is_owner);
 
   int remove(librados::IoCtx& io_ctx, const char *imgname,
-	     ProgressContext& prog_ctx);
+	     ProgressContext& prog_ctx, bool force=false);
   int snap_list(ImageCtx *ictx, std::vector<snap_info_t>& snaps);
   int snap_exists(ImageCtx *ictx, const char *snap_name, bool *exists);
   int snap_is_protected(ImageCtx *ictx, const char *snap_name,
--- a/src/librbd/internal.h
+++ b/src/librbd/internal.h
@@ -90,6 +90,8 @@ namespace librbd {
 
   int snap_set(ImageCtx *ictx, const char *snap_name);
 
+  int list_images_v2(librados::IoCtx& io_ctx,
+      std::map<std::string, std::string>& images);
   int list(librados::IoCtx& io_ctx, std::vector<std::string>& names);
   int list_children(ImageCtx *ictx,
 		    std::set<std::pair<std::string, std::string> > & names);
--- a/src/test/rbd_mirror/test_ImageDeleter.cc
+++ b/src/test/rbd_mirror/test_ImageDeleter.cc
@@ -1,0 +1,442 @@
+// -*- mode:C; tab-width:8; c-basic-offset:2; indent-tabs-mode:t -*-
+// vim: ts=8 sw=2 smarttab
+/*
+ * Ceph - scalable distributed file system
+ *
+ * Copyright (C) 2016 SUSE LINUX GmbH
+ *
+ * This is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License version 2.1, as published by the Free Software
+ * Foundation.  See file COPYING.
+ *
+ */
+#include "include/rados/librados.hpp"
+#include "include/rbd/librbd.hpp"
+#include "include/stringify.h"
+#include "cls/rbd/cls_rbd_client.h"
+#include "tools/rbd_mirror/ImageDeleter.h"
+#include "tools/rbd_mirror/ImageReplayer.h"
+#include "tools/rbd_mirror/Threads.h"
+#include "librbd/ImageCtx.h"
+#include "librbd/ImageState.h"
+#include "librbd/Operations.h"
+#include "librbd/Journal.h"
+#include "librbd/internal.h"
+#include "librbd/Utils.h"
+#include "test/rbd_mirror/test_fixture.h"
+
+#include "test/librados/test.h"
+#include "gtest/gtest.h"
+
+#define GLOBAL_IMAGE_ID "global_image_id"
+#define GLOBAL_CLONE_IMAGE_ID "global_image_id_clone"
+
+#define dout_subsys ceph_subsys_rbd_mirror
+
+using rbd::mirror::RadosRef;
+using rbd::mirror::TestFixture;
+using namespace librbd;
+using cls::rbd::MirrorImageState;
+
+
+void register_test_rbd_mirror_image_deleter() {
+}
+
+class TestImageDeleter : public TestFixture {
+public:
+
+  static int64_t m_local_pool_id;
+
+  const std::string m_local_mirror_uuid = "local mirror uuid";
+  const std::string m_remote_mirror_uuid = "remote mirror uuid";
+
+  static void SetUpTestCase() {
+    TestFixture::SetUpTestCase();
+
+    m_local_pool_id = _rados->pool_lookup(_local_pool_name.c_str());
+  }
+
+  void SetUp() {
+    TestFixture::SetUp();
+
+    librbd::mirror_mode_set(m_local_io_ctx, RBD_MIRROR_MODE_IMAGE);
+
+    m_deleter = new rbd::mirror::ImageDeleter(_rados,
+        m_threads->timer, &m_threads->timer_lock);
+
+    EXPECT_EQ(0, create_image(rbd, m_local_io_ctx, m_image_name, 1 << 20));
+    ImageCtx *ictx = new ImageCtx(m_image_name, "", "", m_local_io_ctx,
+                                  false);
+    EXPECT_EQ(0, ictx->state->open());
+    m_local_image_id = ictx->id;
+
+    cls::rbd::MirrorImage mirror_image(GLOBAL_IMAGE_ID,
+                                MirrorImageState::MIRROR_IMAGE_STATE_ENABLED);
+    EXPECT_EQ(0, cls_client::mirror_image_set(&m_local_io_ctx, ictx->id,
+                                              mirror_image));
+
+    demote_image(ictx);
+    EXPECT_EQ(0, ictx->state->close());
+  }
+
+  void TearDown() {
+    remove_image();
+    TestFixture::TearDown();
+    delete m_deleter;
+  }
+
+  void remove_image(bool force=false) {
+    if (!force) {
+      cls::rbd::MirrorImage mirror_image;
+      int r = cls_client::mirror_image_get(&m_local_io_ctx, m_local_image_id,
+                                           &mirror_image);
+      EXPECT_EQ(1, r == 0 || r == -ENOENT);
+      if (r != -ENOENT) {
+        mirror_image.state = MirrorImageState::MIRROR_IMAGE_STATE_ENABLED;
+        EXPECT_EQ(0, cls_client::mirror_image_set(&m_local_io_ctx,
+                                                  m_local_image_id,
+                                                  mirror_image));
+      }
+      promote_image();
+    }
+    NoOpProgressContext ctx;
+    int r = remove(m_local_io_ctx, m_image_name.c_str(), ctx, force);
+    EXPECT_EQ(1, r == 0 || r == -ENOENT);
+  }
+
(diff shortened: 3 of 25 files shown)


src/librbd/internal.h
```diff
@@ -86,9 +87,13 @@ namespace librbd {
   bool image_options_is_empty(rbd_image_options_t opts);
 
   int snap_set(ImageCtx *ictx, const char *snap_name);
+  int list_images_v2(librados::IoCtx& io_ctx,
```
dillaman: Used?

src/include/Context.h
```diff
@@ -463,6 +464,19 @@ class FunctionContext : public Context {
   boost::function<void(int)> m_callback;
 };
 
+class LambdaContext : public Context {
```
dillaman: Minor: can we hold off on this for now and just use FunctionContext which also takes lambda?  I'd rather see a cleanup ticket to switch from FunctionContext / boost::function to the C++11 native.

src/librbd/internal.cc
```diff
@@ -1913,6 +1921,20 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force) {
       unknown_format = false;
       id = ictx->id;
 
+      if (force) {
+        C_SaferCond ctx;
+        ictx->exclusive_lock->shut_down(&ctx);
```
dillaman: `exclusive_lock` might not be initialized yet -- perhaps move this down to be within the block near line 1939.  Probably should be holding a write lock against the `owner_lock` as well.

src/tools/rbd_mirror/ImageDeleter.cc
```diff
@@ -0,0 +1,367 @@
...
+
+using librados::IoCtx;
+using librbd::ImageCtx;
+using namespace librbd;
+
+namespace rbd {
+namespace mirror {
+
+ImageDeleter::ImageDeleter(string remote_cluster_name, RadosRef local_cluster)
+  : m_remote_name(remote_cluster_name),
```
dillaman: Is there a reason to tie an image deleter to the remote cluster?  I was thinking that we would only need one for rbd-mirror.

src/tools/rbd_mirror/ImageDeleter.h
```diff
@@ -0,0 +1,122 @@
...
+    std::string to_string();
+  };
+
+  enum FailedStep {
+    FAILED_ON_DISABLE_MIRRORING,
+    FAILED_ON_REMOVE_SNAPS,
+    FAILED_ON_REMOVE_IMAGE
+  };
+
+  class FailedDelete {
```
dillaman: Might be nice to combine `FailedDelete` and `DeleteInfo` into a single structure.  That way you could avoid having both the Thread and SafeTimer both running / deleting at the same time -- and you could re-use the existing rbd-mirror SafeTimer to just move failed items from a  "retry" queue back into the main queue (to be processed when the main thread becomes available).

src/tools/rbd_mirror/ImageDeleter.cc
```diff
@@ -0,0 +1,367 @@
...
+                                                   string& image_id,
+                                                   bool *has_snapshots,
+                                                   bool *has_children) {
+
+  *has_snapshots = false;
+  *has_children = false;
+
+  string header_oid = librbd::util::header_name(image_id);
+  ::SnapContext snapc;
+  int r = cls_client::get_snapcontext(ioctx, header_oid, &snapc);
```
dillaman: Thinking about the case where the local has been offline and missed journal events for snap remove, perhaps the image deleter should effectively purge all existing snapshots on the image (assuming it has no active children first).

src/tools/rbd_mirror/ImageDeleter.cc
```diff
@@ -0,0 +1,367 @@
...
+                               &img_name);
+  if (r < 0) {
+    derr << "error retrieving local image name for image id "
+         << del_info->local_image_id << ": " << cpp_strerror(r) << dendl;
+    enqueue_failed_delete(std::move(del_info), FAILED_ON_DISABLE_MIRRORING, r,
+                          retries);
+    return;
+  }
+
+  switch (step) {
```
dillaman: Why worry about tracking the state?  You should be able to just fall right through w/o worry (i.e. if mirroring has already been disabled, it's a no-op).

src/tools/rbd_mirror/ImageDeleter.cc
dillaman: Never mind -- I see that you are doing that after invoking this helper function.

src/tools/rbd_mirror/ImageDeleter.cc
```diff
@@ -0,0 +1,367 @@
...
+    derr << "error retrieving snapshot context for image id " << image_id
+         << ": " << cpp_strerror(r) << dendl;
+    return r;
+  }
+
+  for (const auto& snap_id : snapc.snaps) {
+    *has_snapshots = true;
+    parent_spec pspec(ioctx->get_id(), image_id, snap_id);
+    set<pair<string, string> > names;
+    r = librbd::list_children(*ioctx, pspec, names);
```
dillaman: The snap_unprotect operation will already check for children, so this might not be needed (and you might not need the new `list_children` method).

src/tools/rbd_mirror/ImageDeleter.cc
```diff
@@ -0,0 +1,367 @@
...
+      dout(20) << "snapshot " << imgctx->name << "@" << snap.name
+               << " was deleted" << dendl;
+
+    }
+
+    imgctx->state->close();
+  }
+
+failed_on_remove_image:
+  class SilentProgressContext : public ProgressContext {
```
dillaman: `librbd::NoOpProgressContext` in `librbd/internal.h`

src/tools/rbd_mirror/ImageDeleter.cc
```diff
@@ -0,0 +1,367 @@
...
+                                           mirror_image);
+  if (r < 0) {
+    derr << "cannot disable mirroring for image id "
+         << del_info->local_image_id << ": " << cpp_strerror(r) << dendl;
+    enqueue_failed_delete(std::move(del_info), FAILED_ON_DISABLE_MIRRORING, r,
+                          retries);
+    return ;
+  }
+
+  r = cls_client::mirror_image_remove(&ioctx, del_info->local_image_id);
```
dillaman: Hmm -- we might have a race condition here.  If we crash after removing the local image from the local mirroring directory, we won't know that we need to delete the image when we recover.  The current `librbd::remove` method will disable mirroring -- but I think it needs to be broken up into two parts: (1) the current `mirror_image_disable_internal` call does everything except remove it from the directory, and (2) remove from the mirroring directory after removing from the image directory.

src/tools/rbd_mirror/Replayer.cc
```diff
@@ -220,6 +224,115 @@ int Replayer::init()
   return 0;
 }
 
+void Replayer::read_local_cluster_mirroring_info() {
```
dillaman: Alternative: on start-up, you could populate a `map<pool_id, set<global image id> >` for all local images.  On the first call to `set_sources`, remove from the collection all global image ids from your local collection that exist in the remote.  Any images that remain in the collection are local images that no longer exist on the mirror.

src/tools/rbd_mirror/Replayer.cc
```diff
@@ -220,6 +224,115 @@ int Replayer::init()
...
+      }
+      for (auto it = mirror_images.begin(); it != mirror_images.end(); ++it) {
+        std::string remote_image_id;
+        r = librbd::cls_client::mirror_image_get_image_id(&remote_ioctx,
+                                                          it->second,
+                                                          &remote_image_id);
+        if (r == -ENOENT) {
+          dout(20) << "remote image with global id " << it->second << "does "
+                   << "not exist. Scheduling delete of local mirror" << dendl;
+          m_image_deleter->schedule_image_delete(ioctx.get_id(), it->first,
```
dillaman: Very important to not schedule any images for deletion if they are locally flagged as primary.  Need to guard here and also on shut-down (unless both deletion paths were combined in `set_sources` as noted above).

src/tools/rbd_mirror/ImageDeleter.cc
```diff
@@ -44,14 +46,85 @@ using namespace librbd;
...
+public:
+  ImageDeleterAdminSocketHook(CephContext *cct, ImageDeleter *image_del,
+                              string& cluster_name) :
+    admin_socket(cct->get_admin_socket()),
+    cluster_name(cluster_name) {
+
+    std::string command;
+    int r;
+
+    command = "rbd image-deleter status "+cluster_name;
```
dillaman: Minor: suggested alternative command naming: "rbd mirror deletion status"

src/tools/rbd_mirror/Replayer.cc
```diff
@@ -399,7 +403,17 @@ bool Replayer::stop_image_replayer(unique_ptr<ImageReplayer<> > &image_replayer)
...
+        [&image_replayer, this] (int r) {
+          if (m_image_deleter) {
+            m_image_deleter->schedule_image_delete(
+                          image_replayer->get_local_pool_id(),
+                          image_replayer->get_local_image_id(),
+                          image_replayer->get_global_image_id());
+          }
+        }
+    );
+    image_replayer->stop(ctx);
```
dillaman: We probably also want to guard to make sure we cannot create a new ImageReplayer if there is a pending delete scheduled for an image.  I am worried about the case where a user quickly deletes an image and recreates it.

src/tools/rbd_mirror/ImageDeleter.cc
```diff
@@ -0,0 +1,367 @@
...
+  switch (step) {
+    case FAILED_ON_REMOVE_SNAPS:
+      goto failed_on_remove_snaps;
+    case FAILED_ON_REMOVE_IMAGE:
+      goto failed_on_remove_image;
+    case FAILED_ON_DISABLE_MIRRORING:
+      break;
+  }
+
+  mirror_image.global_image_id = del_info->global_image_id;
```
dillaman: Probably good to ensure / guard that the local image is actually has the same global image id.  Getting a -EEXIST from `mirror_image_set` lets you know somehow another image id is now registered with that global id (bug?).  Getting a -EINVAL indicates the image now has a new global id (so somehow this image had mirroring disabled/re-enabled).

src/include/Context.h
rjfd: Sure, no problem. 
Is the plan to substitute all occurrences of boost::function by C++11 native? I can do that afterwards :-)

src/tools/rbd_mirror/ImageDeleter.cc
rjfd: You're right. In my first implementation I would only delete an non-primary image if the primary images was really deleted (and not only disabled) and to know that I would have to contact the remote cluster. That is why ImageDeleter was tied with a remote cluster.
Now we just need one ImageDeleter.

src/librbd/internal.h
rjfd: Another leftover of the first implementation. I was using it in ImageDeleter.cc. Not anymore.

src/librbd/internal.cc
rjfd: The reason I'm shutting down `exclusive_lock` here is because it needs to be done before the call to `ictx->operations->prepare_image_update()`

Why shouldn't the `exclusive_lock` be initialized at this point? Since ImageState::open() is synchronous, I though it would not be a problem to do this here.

dillaman: If the exclusive lock feature is enabled, the `exclusive_lock` should be non-null -- but if an notify event comes in on the header while you have it open, there is a potential for a race if you aren't holding the lock.  If the exclusive lock feature isn't enabled (e.g. this wasn't a mirrored image, but just a regular ol' ```rbd rm foo```-call, it might be null)

src/include/Context.h
dillaman: There isn't a formulated plan for this in particular, but there has been talk about switching to C++11 primitives where available.  Such cleanup work should probably be left until the kraken release at this point.

src/librbd/internal.cc
rjfd: > there is a potential for a race if you aren't holding the lock.
Which lock? the `owner_lock`?

dillaman: Yeah -- ```owner_lock``` protects ```ImageCtx::exclusive_lock```.

src/tools/rbd_mirror/ImageDeleter.cc
rjfd: I was just trying to save some processing steps, but I agree that it may not worth the effort

src/tools/rbd_mirror/ImageDeleter.cc
dillaman: Yeah -- I think it can only be a bug for -EEXIST.  In such a case, log it and abort seems reasonable for now.  I think -EINVAL is possible if the user force-promotes the local image, disables mirroring, and re-enables it (e.g. crazy user).  Again, I think log and abort seems reasonable.

src/tools/rbd_mirror/ImageDeleter.cc
rjfd: I think, I'll move the call of `mirror_image_remove` to after the call of  `librbd::remove`, and make necessary checks (check for -ENOENT) during all the execution of `process_image_delete`.

src/tools/rbd_mirror/Replayer.cc
rjfd: I'm going to check this condition inside `ImageDeleter::process_image_delete`

src/tools/rbd_mirror/Replayer.cc
rjfd: I have a problem here. The information we keep for deleting an image is it's local_id and global_image_id.
If the user disables mirroring for the primary image and then re-enables it, the global_image_id will be different, so we will not be able check if there is a deletion scheduled. 
We could also keep the remote_id of the primary image, but it will also not work in the case where the user deletes and re-creates the primary image.

What will be the consequence of create a new ImageReplayer when there is a pending delete? Is because there still exists a local image with the same name? I that is the case, maybe what we should keep track in the DeleteInfo structure is the image name, which is common in the remote and local cluster.

dillaman: Yeah, that was the case I was worried about -- I think we should match on remote image id or remote image name.

rjfd:
@dillaman finished the changes addressing your comments. I made big changes, so it will be as almost as the first time you are reviewing the code :stuck_out_tongue_closed_eyes:

rjfd:
Finished implementation and rebased on top of master

src/test/rbd_mirror/test_ImageDeleter.cc
```diff
@@ -0,0 +1,210 @@
...
+#define GLOBAL_IMAGE_ID "global_image_id"
+
+#define dout_subsys ceph_subsys_rbd_mirror
+
+using rbd::mirror::RadosRef;
+
+void register_test_rbd_mirror_image_deleter() {
+}
+
+class TestImageDeleter : public ::testing::Test {
```
dillaman: If you inherit from the "TestFixture" class, it creates and connects to the two pools automatically.

src/test/rbd_mirror/test_ImageDeleter.cc
```diff
@@ -0,0 +1,210 @@
...
+
+    rbd::mirror::ImageReplayer<>::BootstrapParams bootstap_params(m_image_name);
+    C_SaferCond cond;
+    replayer->start(&cond, &bootstap_params);
+    ASSERT_EQ(0, cond.wait());
+
+    m_local_image_id = replayer->get_local_image_id();
+    m_local_image_name = replayer->get_local_image_name();
+
+    if (seconds > 0) {
```
dillaman: I'm probably missing something, but what's the purpose of the sleeps?  At a quick glance, it seems like it could be problematic/racy in the build jobs.

src/tools/rbd_mirror/PoolWatcher.cc
```diff
@@ -115,7 +116,13 @@ void PoolWatcher::refresh_images(bool reschedule)
         continue;
       }
       for (auto it = mirror_images.begin(); it != mirror_images.end(); ++it) {
-        image_ids.insert(ImageIds(it->first, it->second));
+        std::string image_name;
+        r = dir_get_name(&ioctx, RBD_DIRECTORY, it->first, &image_name);
```
dillaman: I would recommend a single pass of retrieving all image names in the pool, followed by retrieving the mirrored images, and then cross-reference.  I'm not sure we can skip adding the mirrored image to the collection since that would be interpreted as a deleted image propagation.  Maybe a boost::optional<std::string> for the image name?

src/tools/rbd_mirror/ImageDeleter.cc
```diff
@@ -0,0 +1,491 @@
...
+  dout(20) << "enter" << dendl;
+
+  Mutex::Locker l(m_delete_lock);
+  m_delete_queue.push_front(unique_ptr<DeleteInfo>(
+        new DeleteInfo(local_pool_id, local_image_id, local_image_name,
+                       global_image_id)));
+  m_delete_queue_cond.Signal();
+}
+
+void ImageDeleter::is_image_scheduled_for_delete(const std::string& image_name,
```
dillaman: Perhaps a name like ```wait_for_scheduled_deletion``` might be better?

src/test/rbd_mirror/test_ImageDeleter.cc
rjfd: yeah, this was a nasty hack that I used to ensure (or not) that I wouldn't stop image replayer while it was still holding the local temporary snapshot. As a side effect, I found that if I stop during image replayer sync, the interruption handler does not remove the temporary snapshot, and the ImageDeleter will not be able to delete the local image because it fails while trying to delete the temporary snapshot with an -EROFS error. 

What do you suggest to ensure that the the sync is finished?

src/test/rbd_mirror/test_ImageDeleter.cc
rjfd: I tried to use it, but it was easier to do the tests cleanup by deleting pools which goes against the logic of TestFixture.

src/tools/rbd_mirror/ImageDeleter.cc
rjfd: agreed

src/test/rbd_mirror/test_ImageDeleter.cc
dillaman: Are you referring to the "FailedDeletePoolENOENT" test case where it deletes the local pool?  Could just create a pool specifically for that test case.

src/test/rbd_mirror/test_ImageDeleter.cc
dillaman: It would probably be better to test ImageDeleter in isolation from ImageReplayer -- and you could add a test case to ImageReplayer (or one of its subordinate classes) to verify it properly interacts with ImageDeleter.

src/librbd/internal.cc
```diff
@@ -1931,12 +1935,31 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force) {
...
       if (ictx->exclusive_lock != nullptr) {
-        r = ictx->operations->prepare_image_update();
-        if (r < 0 || !ictx->exclusive_lock->is_lock_owner()) {
-	  lderr(cct) << "cannot obtain exclusive lock - not removing" << dendl;
-	  ictx->owner_lock.put_read();
-	  ictx->state->close();
-          return -EBUSY;
+        if (force) {
+          // releasing read lock to avoid a deadlock when upgrading to
+          // write lock in the stut_down process
```
dillaman: Typo: "shut_down"

src/test/rbd_mirror/test_ImageDeleter.cc
rjfd: That's great idea! Thanks

rjfd:
@dillaman Pushed the changes and rebased on top of master.

src/librbd/internal.cc
```diff
@@ -389,14 +390,16 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force) {
...
+      lderr(cct) << "failed to remove image from mirroring directory: "
+                 << cpp_strerror(r) << dendl;
+      return r;
+    }
 
-  ldout(cct, 20) << "removed image state from rbd_mirroring object" << dendl;
+    ldout(cct, 20) << "removed image state from rbd_mirroring object" << dendl;
+  }
 
   if (is_primary) {
```
dillaman: Will need to delay this notification when ```!remove```.

rjfd: Yes, I forgot about that. Thx!

dillaman:
@rjfd can you rebase and repush?

rjfd:
@dillaman done!

dillaman:
@rjfd Hit a race condition while manually testing.  I had an image replicated between two clusters and I disabled mirroring on the primary image.  The ```ImageDeleter``` properly picked up that it should delete the image, but when it tried to acquire the exclusive lock in order to delete it, it ended up blacklisting itself:

> 2016-05-03 09:13:25.883965 7f0bca7fc700 10 librbd::exclusive_lock::AcquireRequest: retrieved exclusive locker: client.4196@127.0.0.1:0/1749366697
> 2016-05-03 09:13:25.883969 7f0bca7fc700 10 librbd::exclusive_lock::AcquireRequest: send_get_watchers
> 2016-05-03 09:13:25.884277 7f0bca7fc700 10 librbd::exclusive_lock::AcquireRequest: handle_get_watchers: r=0
> 2016-05-03 09:13:25.884284 7f0bca7fc700 10 librbd::exclusive_lock::AcquireRequest: send_blacklist
> 2016-05-03 09:13:26.016784 7f0b6cff9700 10 librbd::exclusive_lock::AcquireRequest: handle_blacklist: r=0

The reason is because the ImageReplayer state machine was running on the same image -- and currently had the image open w/ the exclusive lock acquired.

src/tools/rbd_mirror/ImageDeleter.cc
```diff
@@ -0,0 +1,536 @@
...
+    del_info->notify(r);
+    return;
+  } else if (r < 0) {
+    derr << "cannot disable mirroring for image id "
+         << curr_deletion->local_image_id << ": " << cpp_strerror(r) << dendl;
+    enqueue_failed_delete(r);
+    return ;
+  }
+
+  dout(20) << "setted local image mirroring to disable" << dendl;
```
dillaman: Minor: just saw this in the logs -- should be "set local image mirroring state to disabling"

dillaman:
@rjfd Correction: I am seeing the ImageDeleter attempt to list the snapshots on the local image (to delete), detect that it needs to delete a snapshot, attempt to acquire the exclusive lock, and fail opening the journal since the local image is not primary.  This causes the exclusive lock to be kept in a locked state but the image is closed.  When ImageDeleter retries, it sees that the lock is till held (by itself) and causes the blacklist.

I'll fix the issue where the lock is not properly released upon journal error.  I think we will also need to expand the ```exclusive_lock::Policy``` to prevent opening the journal when you attempt to unprotect or delete snapshots.  I can create a PR for that and you could just swap out ```ImageCtx::exclusive_lock_policy``` with a new one that disables opening the journal.

rjfd:
@dillaman That is a good idea! I'll add more unit test to cover that case.

dillaman:
@rjfd Actually -- in the short term, you should be able to directly unset the RBD_FEATURE_JOURNALING bit after opening the image but before proceeding with snapshot manipulation.  Looking at adjusting the policy to support this looks like it might be more effort than it really solves.

rjfd:
@dillaman pushed the changes and rebased with master
Added more unit tests to test the cases of images with snapshots and clones.

src/tools/rbd_mirror/ImageReplayer.cc
```diff
@@ -404,6 +404,9 @@ void ImageReplayer<I>::handle_bootstrap(int r) {
 
   {
     Mutex::Locker locker(m_lock);
+    m_name = m_local_ioctx.get_pool_name() + "/" + m_local_image_ctx->name;
```
rjfd: TODO: remove this line

dillaman:
@rjfd Hopefully last thing for this PR: the 'qa/workunits/rbd/rbd_mirror.sh' test needs to be updated because it currently fails.  It would be good to also add a simple deletion propagation test to that file.

rjfd:
@dillaman added a test case for the image deletion propagation. Also fixed a problem with the disabling of mirroring for non-primary images in commit c429f61 .

rjfd:
@dillaman Just rebased on top of master to resolve the conflicts. Also added one more bugfix to 48f005f

dillaman:
@rjfd thx -- I am going to build a test branch this afternoon with all the 'wip-jason-testing' bugs.

dillaman:
http://qa-proxy.ceph.com/teuthology/jdillaman-2016-05-22_12:32:14-rbd-wip-jd-testing---basic-mira/207665/teuthology.log

Invalid memory access detected in TestImageDeleter.Delete_NonPrimary_Image test

dillaman:
Resolved invalid memory access and merged manually

PR #8474, opened 2016-04-06
pycephfs: using namespace for data isolation

xiaoxichen:


src/pybind/ceph_volume_client.py
```diff
@@ -439,19 +441,33 @@ def create_volume(self, volume_path, size=None, data_isolated=False):
...
-                    'pool': pool_name
-                })
-            self.fs.setxattr(path, 'ceph.dir.layout.pool', pool_name, 0)
+            isolate_by_namespace = bool(self.rados.conf_get("cephFSClient_data_isolate_by_namespace"))
+            if isolate_by_namespace:
+                namespace = "{0}{1}".format(self.POOL_NS_PREFIX, volume_path.volume_id)
+                log.info("create_volume: {0}, using rados namespace {1} to isolate data.".format(volume_path, namespace))
+                try:
+                    self.fs.setxattr(path, 'ceph.dir.layout.pool_namespace', namespace, 0)
+                except:
```
jcsp: All exception handlers must specify what exception they are catching, catch-all handlers are bad because they might catch things like connection errors (we should let these get raised further), rather than the kind of thing we are expecting (errors from the particular filesystem call)

src/common/config_opts.h
```diff
@@ -396,7 +396,8 @@ OPTION(client_metadata, OPT_STR, "")
 OPTION(client_acl_type, OPT_STR, "")
 OPTION(client_permissions, OPT_BOOL, true)
 OPTION(client_dirsize_rbytes, OPT_BOOL, true)
-
+// consume by ceph_volume_client, use rados namespace to isolate cephFS volume instead of using pool
+OPTION(cephFSClient_data_isolate_by_namespace, OPT_BOOL, true)
```
jcsp: I don't think we need to make it optional: CephFSVolumeClient was only released with Jewel, so it does not have to work with any ceph versions that don't have the namespace support.

src/pybind/ceph_volume_client.py
```diff
@@ -439,19 +441,33 @@ def create_volume(self, volume_path, size=None, data_isolated=False):
...
-            pool_name = "{0}{1}".format(self.POOL_PREFIX, volume_path.volume_id)
-            pool_id = self._create_volume_pool(pool_name)
-            mds_map = self._rados_command("mds dump", {})
-            if pool_id not in mds_map['data_pools']:
-                self._rados_command("mds add_data_pool", {
-                    'pool': pool_name
-                })
-            self.fs.setxattr(path, 'ceph.dir.layout.pool', pool_name, 0)
+            isolate_by_namespace = bool(self.rados.conf_get("cephFSClient_data_isolate_by_namespace"))
+            if isolate_by_namespace:
```
jcsp: I suggest we move the namespace isolation up out of the "if data_isolated" block, so that it is always used for all volumes

jcsp:
This is largely correct, just need to change the logic to always use namespaces (not just in data_isolated) and make sure that any exception handlers are for a specific class of exception.

jcsp:
You could also update tests here: https://github.com/ceph/ceph-qa-suite/blob/master/tasks/cephfs/test_volume_client.py to assert that data is going into a namespace correctly.  The vstart_runner.py script in the same directory should enable you to run the tests quickly against a vstart cluster.

xiaoxichen:
@jcsp thanks, one question, cant cephfsvolumeclinet work
 with pre jewel release?  Seems It can if we dont use namespace as it only use standard libcephfs api.

src/pybind/ceph_volume_client.py
xiaoxichen: I think it would be better to leave it inside "data_isolated" block, as user may want to create a public share(volume) with no namespace specified(or a ns called "public") , and in the authorization by default grant user access to the public namespace. Is that reasonable?

src/common/config_opts.h
xiaoxichen: Reasonable, there is no means to use pool to isolate while namespace supported, if doesn't support, it will fall back to pool automatically.

src/pybind/ceph_volume_client.py
xiaoxichen: Changed my mind, the data_isolated flag is passed from outside, so if someone mistakenly passed data_isolated= True for a volume which is not isolated by pool , in current code, we will DESTROY THE WHOLE DATA POOL.

xiaoxichen:
@jcsp , basically the decision is whether we need to support older release.
a)  only support ceph after jewel  which means namespace is always there, in no way we will use pool.
b)  need to work with hammer/infernalis , which means we need to fall back to use pool when namespace doesnt present.


Isolate by pool is dangerous because remove pool is dangerous, we have no way to check whether the pool is safe for remove. But if we want to support older release, we had to go with it.

jcsp:
This is definitely only for Jewel and newer, there are other features used here like the filtered "session evict" command that are only supported in jewel.

I would actually like to keep the data_isolated option (for using separate pools) as well as using rados namespaces (used always), because we might also want to use the pool isolation to control placement (e.g. put one share on SSDs) in addition to having the security isolation from the namespaces.

xiaoxichen:
@jcsp , cool, only for Jewel make the logic much clear.
Yes, we can always use rados namespace(as it's surely supported) for isolation. The pool is somehow like a placement rule desision, the "data_isolated" option is somehow misleading and limit.Say, if I would like one share on SSDs(default to HDD), I have to mark data_isolated to True, and got a Pool called ${POOL_PREFIX}_mySSDShare, then I need to adjust the ruleset/replication level for the pool.

Instead of doing this, would it be better change "data_isolated" option to "data_pool", which default to POOL_INHERITE. Thus user can specified an exisiting pool (NS ensure the isolation already), or leave it as default means inherite from parent dir.

Make sense?

xiaoxichen:
And we no longer need to delete a pool...now we are super safe.

src/common/config_opts.h
```diff
@@ -396,7 +396,6 @@ OPTION(client_metadata, OPT_STR, "")
 OPTION(client_acl_type, OPT_STR, "")
 OPTION(client_permissions, OPT_BOOL, true)
 OPTION(client_dirsize_rbytes, OPT_BOOL, true)
-
```
zhiteng: Is this change (removing blank line) necessary?

xiaoxichen: Nop, will remove it....

src/pybind/ceph_volume_client.py
```diff
@@ -519,9 +528,19 @@ def rmtree(root_path):
...
+        else:
+            log.info("Volume '{0}' is isolated by rados namespace {1}, no need to delete pool.".format(
+                volume_path.volume_id, namespace))
+
         rmtree(trashed_volume)
 
-        if data_isolated:
+        if data_isolated and namespace == "":
+        # only delete pool when layout.pool_namespace doesn't support or unset
+        # which indicate we use pool to isolate the data.
```
zhiteng: comments are under-indent

src/pybind/ceph_volume_client.py
```diff
@@ -439,20 +441,27 @@ def create_volume(self, volume_path, size=None, data_isolated=False):
...
-            if pool_id not in mds_map['data_pools']:
-                self._rados_command("mds add_data_pool", {
-                    'pool': pool_name
-                })
-            self.fs.setxattr(path, 'ceph.dir.layout.pool', pool_name, 0)
-
+        namespace = "{0}{1}".format(self.POOL_NS_PREFIX, volume_path.volume_id)
+        log.info("create_volume: {0}, using rados namespace {1} to isolate data.".format(volume_path, namespace))
+        try:
+            self.fs.setxattr(path, 'ceph.dir.layout.pool_namespace', namespace, 0)
```
zhiteng: It's confusing that namespace is used regardless 'data_isolated' flag.  If the flag was meant to represent whether volume should be created in a way that access to it can be controlled separately, then this flag should be honored.

jcsp: Think of "data_isolated" as referring to physical isolation (i.e. potentially using different OSDs).  We might in the future add "metadata_isolated" to use a separate filesystem with separate MDSs.

data_isolated is kind of a weird thing, so for the moment I'd like to just keep it completely separate from the namespace stuff -- everything should use a namespace by default, and data_isolated should be the special case

xiaoxichen:
@jcsp , simplify a lot as safely assume rados namespace is there. See if it works for you.

xiaoxichen:
@jcsp  ping:)

src/pybind/ceph_volume_client.py
```diff
@@ -443,7 +445,9 @@ def create_volume(self, volume_path, size=None, data_isolated=False):
         if size is not None:
             self.fs.setxattr(path, 'ceph.quota.max_bytes', size.__str__(), 0)
 
+        # data_isolated means create a seperate pool for this volume
         if data_isolated:
+            log.info("create_volume: {0}, create pool {1} as data_isolated =True.".format(volume_path, pool_name))
```
ajarr: pool_name not assigned yet.

src/pybind/ceph_volume_client.py
```diff
@@ -453,6 +457,11 @@ def create_volume(self, volume_path, size=None, data_isolated=False):
                 })
             self.fs.setxattr(path, 'ceph.dir.layout.pool', pool_name, 0)
 
+        # enforce security isolation, create a seperate pool for this volume
```
ajarr: pool or namespace?

src/pybind/ceph_volume_client.py
xiaoxichen: Done

src/pybind/ceph_volume_client.py
xiaoxichen: Oops,fixed

xiaoxichen:
@jcsp ping

jcsp:
(Tested this locally)

PR #8476, opened 2016-04-06
rbd-mirror: workaround for intermingled lockdep singletons

dillaman:
If librados constructs the CephContext, lockdep will be initialized
in both the librados shared address space and also within the app
address space.  This breaks lockdep's functionality and results in
a crash upon shutdown.

### DIFF
--- a/src/tools/rbd_mirror/Replayer.cc
+++ b/src/tools/rbd_mirror/Replayer.cc
@@ -5,6 +5,9 @@
 
 #include "common/Formatter.h"
 #include "common/admin_socket.h"
+#include "common/ceph_argparse.h"
+#include "common/code_environment.h"
+#include "common/common_init.h"
 #include "common/debug.h"
 #include "common/errno.h"
 #include "include/stringify.h"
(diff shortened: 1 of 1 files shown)


dillaman:
@trociny @rjfd FYI -- after some digging, I created this workaround for the lockdep issues.  Short term I think this will resolve the crashes on shutdown when lockdep is enabled.

rjfd:
@dillaman lgtm --- already tested and the assertion failure is no more.
I was wondering if the only option to make lockdep truly singleton would be to make it a library on its own.

dillaman:
@rjfd Yeah -- we run into lots of issues due to dual linkage of libcommon/libosdc/etc in librados and librbd/other apps.

jdurgin:
This reminds me we might want to use CINIT_FLAG_UNPRIVILEGED_DAEMON_DEFAULTS for common_preinit() to get logs and asok by default for the remote cluster too.

trociny:
@dillaman Good idea, but there is an issue with the patch: by some reason command line options are not applied. You can see this running rbd_mirror.sh with NOCLEANUP, and observe that asok and log files are not created in tmp directory but in run/cluster2/out directory (as specified in cluster2.conf).

So far I have failed to understand why.

src/tools/rbd_mirror/Replayer.cc
```diff
@@ -143,37 +146,59 @@ int Replayer::init()
...
+    derr << "could not parse environment for " << m_peer << ":"
+         << cpp_strerror(r) << dendl;
+    cct->put();
     return r;
   }
 
   if (!m_args.empty()) {
-    r = m_remote->conf_parse_argv(m_args.size(), &m_args[0]);
+    // librados::Rados::conf_parse_argv
+    r = cct->_conf->parse_argv(args);
```
trociny: @dillaman s/args/m_args/ :-)

dillaman:
Updated and fixed -- thx.

trociny:
lgtm

PR #8477, opened 2016-04-06
journal: fix race condition between Future and journal shutdown

dillaman:


jdurgin:
unittest failure looks real

jdurgin:
otherwise lgtm

dillaman:
@jdurgin Indeeed -- fun with async callbacks.

PR #8479, opened 2016-04-07
hammer: installation fix for debian ubuntu using vstart

rjain-iitr:


dachary:
@rjain-iitr I don't see why that would be necessary, could you explain the reason for this patch ?

asheplyakov:
--force-yes is inherently dangerous and should not be used except for very special situations

dachary:
@rjain-iitr feel free to re-open whenever you're ready to address the comments :-)

PR #8482, opened 2016-04-07
os/bluestore/BlueStore: preallocate space when use file instead of blockdevice

majianpeng:
Avoid failure because lack of space.

### DIFF
--- a/src/os/bluestore/BlueStore.cc
+++ b/src/os/bluestore/BlueStore.cc
@@ -1639,31 +1639,34 @@ int BlueStore::_setup_block_symlink_or_file(
 	  VOID_TEMP_FAILURE_RETRY(::close(fd));
 	  return r;
 	}
+
+	if (g_conf->bluestore_block_preallocate_file) {
 #ifdef HAVE_POSIX_FALLOCATE
-	r = ::posix_fallocate(fd, 0, size);
-	if (r < 0) {
-	  r = -errno;
-	  derr << __func__ << " failed to prefallocate " << name << " file to "
-	       << size << ": " << cpp_strerror(r) << dendl;
-	  VOID_TEMP_FAILURE_RETRY(::close(fd));
-	  return r;
-	}
-#else
-	char data[1024*128];
-	for (uint64_t off = 0; off < size; off += sizeof(data)) {
-	  if (off + sizeof(data) > size)
-	    r = ::write(fd, data, size - off);
-	  else
-	    r = ::write(fd, data, sizeof(data));
+	  r = ::posix_fallocate(fd, 0, size);
 	  if (r < 0) {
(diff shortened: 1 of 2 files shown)


src/os/bluestore/BlueStore.cc
```diff
@@ -1640,6 +1640,31 @@ int BlueStore::_setup_block_symlink_or_file(
...
+	r = ::posix_fallocate(fd, 0, size);
+	if (r < 0) {
+	  r = -errno;
+	  derr << __func__ << " failed to prefallocate " << name << " file to "
+	       << size << ": " << cpp_strerror(r) << dendl;
+	  VOID_TEMP_FAILURE_RETRY(::close(fd));
+	  return r;
+	}
+#else
+	char data[1024*128];
```
tchaikov: can we use `ftruncate(2)` instead?

majianpeng: truncate don't alloc space. Or i miss something?

tchaikov: no, it does not.

majianpeng: Sorry to reply so later. The code already use ftruncate.  But this don't allocate space. So like FileJournal, avoid later no space use and it should allocate space.

tchaikov: @majianpeng i agree. but file backend for bluestore is mostly for testing usage. imho, i am not sure if we want to complicate the code to avoid the ENOSPAC in testing env.

majianpeng: Yes, found this bug was in my testing env.  Now i am writing code about PMEMdevice(like NVMEDevice) which can use file instead(for wal of rocksdb). So this isn't only for testing usage.

liewegas:
@majianpeng can you guard this with a bdev_preallocate_file bool option?  and maybe default to false, since it having it unallocated can be convenient for development.

yuriw:
Pls resolve merge conflicts:

[yuriw@smithi028 src]$ git pull https://github.com/majianpeng/ceph.git bluestore-prefallocate
remote: Counting objects: 15, done.
remote: Total 15 (delta 12), reused 12 (delta 12), pack-reused 3
Unpacking objects: 100% (15/15), done.
From https://github.com/majianpeng/ceph
 * branch            bluestore-prefallocate -> FETCH_HEAD
Auto-merging src/os/bluestore/BlueStore.cc
Auto-merging src/common/config_opts.h
CONFLICT (content): Merge conflict in src/common/config_opts.h
Automatic merge failed; fix conflicts and then commit the result.

liewegas:
rebase please?

majianpeng:
@yuriw . Update. Thanks!

PR #8484, opened 2016-04-07
mds: use epoch as incarnation counter (#15399)

jcsp:
To avoid the need to manage the per-map arrange of rank to incarnation across multiple filesystems.


ukernel:
looks goods

gregsfortytwo:
Okay, I'm forgetting part of the standup conversation now, but I thought we decided that we still needed an additional way of distinguishing between MDSes of the same rank in different filesystems. Do you want to fix that here?

jcsp:
Yes, let's not merge this until I've done the using-gids-as-names part.

gregsfortytwo:
Okay, we want to do this one for Jewel after all. Putting off the better fixes until Kraken since they're more involved and require MDSMap changes to add addresses.

liewegas:
Reviewed-by:

jcsp:

PR #8486, opened 2016-04-07
librados: Revert "rados: Add new field flags for ceph_osd_op.copy_get."

liewegas:
This reverts commit 2b4acfb1b808b98132a771cf1089063c0f7a75b5.

We fixed this on the receiving end in
fc51ce2a837e8e878d46d8ca54531aa7bd5d01b2 a few weeks after this commit
was added.  It's ugly and no longer needed.

This *does* mean that we'll start to copy some data and then abort
partway through, but that's better than an kludgey protocol/interface.

### DIFF
--- a/src/include/rados.h
+++ b/src/include/rados.h
@@ -446,10 +446,6 @@ enum {
 };
 
 enum {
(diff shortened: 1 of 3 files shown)


athanatos:
lgtm

PR #8488, opened 2016-04-07
rgw: fix problem deleting objects begining with double underscores

oritwas:
Fixes: http://tracker.ceph.com/issues/15318

### DIFF
--- a/src/rgw/rgw_op.cc
+++ b/src/rgw/rgw_op.cc
@@ -2994,7 +2994,8 @@ void RGWDeleteObj::execute()
     return;
   }
 
-  rgw_obj obj(s->bucket, s->object);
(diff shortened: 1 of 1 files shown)


mattbenjamin:
I see how this avoids the __ logic in the RGWDeleteObj path, and that appears unconditional.  Can you explain why it is never relevant in that path?  (Apologies if I'm misunderstanding.)

oritwas:
When we send an object to the osd we add an underscore to objects start with an underscore , so when we get back an object  that starts with two underscore we need to remove one. This is done is the ctor of rgw_obj that gets an rgw_obj_key parameter .
Here we get the object name from the url and there is no need to remove the underscore.  rgw_obj as another ctor that doesn't remove the underscore so I used it instead.

PR #8489, opened 2016-04-07
systemd: make Ceph daemons dependent upon time-sync.target

smithfarm:
Fixes: http://tracker.ceph.com/issues/15419

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -101,7 +101,8 @@ Recommends: btrfs-tools,
             ceph-mds (= ${binary:Version}),
             librados2 (= ${binary:Version}),
             libradosstriper1 (= ${binary:Version}),
-            librbd1 (= ${binary:Version})
+            librbd1 (= ${binary:Version}),
+            ntp | time-daemon
 Replaces: ceph-common (<< 0.78-500), python-ceph (<< 0.92-1223),
 	  ceph-test (<< 0.94-1322), ceph (<< 10)
 Breaks: python-ceph (<< 0.92-1223), ceph-test (<< 0.94-1322), 
@@ -600,6 +601,7 @@ Depends: ceph-common (= ${binary:Version}),
          librgw2 (= ${binary:Version}),
          ${misc:Depends},
          ${shlibs:Depends}
+Recommends: ntp | time-daemon
 Description: REST gateway for RADOS distributed object store
  RADOS is a distributed object store used by the Ceph distributed
  storage system.  This package provides a REST gateway to the
--- a/ceph.spec.in
+++ b/ceph.spec.in
@@ -249,6 +249,7 @@ Requires:      findutils
 Requires:      which
 %if 0%{?suse_version}
 Requires:      lsb-release
+Recommends:    ntp-daemon
 %endif
 %if 0%{?fedora} || 0%{?rhel}
 Requires:      redhat-lsb-core
--- a/systemd/ceph-mds@.service
+++ b/systemd/ceph-mds@.service
@@ -1,7 +1,7 @@
 [Unit]
 Description=Ceph metadata server daemon
-After=network-online.target local-fs.target
-Wants=network-online.target local-fs.target
+After=network-online.target local-fs.target time-sync.target
+Wants=network-online.target local-fs.target time-sync.target
 PartOf=ceph-mds.target
 
 [Service]
--- a/systemd/ceph-mon@.service
+++ b/systemd/ceph-mon@.service
@@ -5,8 +5,8 @@ Description=Ceph cluster monitor daemon
 #   http://www.freedesktop.org/wiki/Software/systemd/NetworkTarget
 # these can be removed once ceph-mon will dynamically change network
 # configuration.
-After=network-online.target local-fs.target ceph-create-keys@%i.service
-Wants=network-online.target local-fs.target ceph-create-keys@%i.service
+After=network-online.target local-fs.target time-sync.target ceph-create-keys@%i.service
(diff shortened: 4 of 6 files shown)


liewegas:
What happens if ntp isn't installed?  It's not a package requirement

smithfarm:
If a provider of `time-sync.target` is not installed, Ceph will not work correctly, right? This patch makes this dependency explicit.

I agree that I need to find a multi-distro way of specifying the package dependency. I've suggested adding "Requires: ntp" some time ago, but that was shot down because someone might be using chrony or systemd-timesyncd instead of ntp.

liewegas:
We need to adjust the package dependency before changing this, tho, or 
else the user can install it and have it not work at all.

On debian, there is a pseudo-package 'time-daemon' we can depend on, so 
that's easy enough.  I don't see one of the rpm universe, though.  Maybe 
just depends ntp | chronyd is sufficient?

smithfarm:
Thanks! Marked DNM until the packaging aspect is addressed.

smithfarm:
https://packages.debian.org/jessie/time-daemon

smithfarm:
```
systemd-timedated is a system service that may be used as a mechanism to change the 
system clock and timezone, as well as to enable/disable NTP time synchronization.  
systemd-timedated is automatically activated on request and terminates itself when it is unused.

The tool timedatectl(1) is a command-line client to this service.
```

smithfarm:
I'm not sure, but it looks like `systemd-timedated` might provide sufficient clock synchronization to run a Ceph cluster without needing either `ntp` or `chrony`.

smithfarm:
From https://wiki.archlinux.org/index.php/systemd-timesyncd

> systemd-timesyncd is a daemon that has been added for synchronizing the system clock across the network. It implements an SNTP client. In contrast to NTP implementations such as chrony or the NTP reference server this only implements a client side, and does not bother with the full NTP complexity, focusing only on querying time from one remote server and synchronizing the local clock to it. Unless you intend to serve NTP to networked clients or want to connect to local hardware clocks this simple NTP client should be more than appropriate for most installations.

tserong:
We might not strictly need a package dependency at all; `Wants=` is meant to be a weak requirement (systemd.unit(5) says: "if the listed units fail to start or cannot be added to the transaction, this has no impact on the validity of the transaction as a whole"). I haven't actually tested this myself though.

liewegas:
Ah, that would be perfect... then we can address the requirement 
separately (or make it a Recommends:, or whatever).

liewegas:
@javacruft @ktdreyer care to weigh in?

smithfarm:
Rebased and added commit fd3b3ab to recommend `ntp-daemon` metapackage on SUSE.

smithfarm:
Changed the debian commit to only recommend `time-daemon`, not require it.

smithfarm:
I looked at the CentOS/Fedora ntp and chrony spec files and the ones I found by googling did not have any "provides".

smithfarm:
Changelog:
* rebased
* added commit 6c43404 to recommend `ntp-daemon` metapackage on SUSE
* Changed the debian commit to only recommend time-daemon, not require it.

liewegas:
I'm just slightly nervous that the After line isn't soft the same way Wants is?  Did you test that systemd still behaves when no time daemon is installed?

Otherwise, LGTM!

smithfarm:
Have not tested yet, but doing so now.

ktdreyer:
my concern is the same as Sage's (is it ok to have "After" and "Wants" here?) Otherwise this LGTM

```
Reviewed-by: Ken Dreyer <kdreyer@redhat.com>
```

smithfarm:
For my test, I instantiated four CentOS 7.2 VMs. On one I installed ceph-deploy and on the other three I used ceph-deploy to install ceph and get a mon cluster running.

What I found out was it's quite difficult to convince systemd to **not** synchronize the clock, but I did manage to do it by disabling ntpd.service, chronyd.service, and time-sync.target **and** by issuing the command `timedatectl set-ntp no` - and then reboot the node.

After this I ran the following sequence of commands to verify that I can stop and start the monitor unit without any problems: http://paste2.org/65evCptc

Then I found that disabling "time-sync.target" just sets it to "static", so it can still get run. And indeed starting the ceph-mon unit causes it to get run and that causes chronyd.service to be started. 

So I **removed the chrony package completely** and rebooted again, and ran the test again. Here is the result: http://paste2.org/2bY4ZFLZ

As you can see, even without chrony installed, and `time-sync.target` runs just fine.

My conclusion is that this patch is OK to merge.

smithfarm:
Now I noticed that the `ntp` package was installed the entire time, although the `ntp.service` was disabled. So I removed the ntp package and tried the test again. Again, the `time-sync.target` unit starts even though there is no way for the system to sync the time. See the log: http://paste2.org/PCaLMVMH

debian/control
```diff
@@ -101,7 +101,8 @@ Recommends: btrfs-tools,
             ceph-mds (= ${binary:Version}),
             librados2 (= ${binary:Version}),
             libradosstriper1 (= ${binary:Version}),
-            librbd1 (= ${binary:Version})
+            librbd1 (= ${binary:Version}),
+            time-daemon
```
javacruft: time-daemon is a virtual package so you need to provide a primary 'real' package as well:

   ntp | time-daemon

That way, its either ntp or another package that Provides: time-daemon if also provided on the apt-get call.

smithfarm:
Changelog
* `Recommends: ntp | time-daemon` in the Debian commit ( per @javacruft )

liewegas:
Thanks, Nathan, for testing!

PR #8490, opened 2016-04-07
rbd: journal reset should disable/re-enable journaling feature

dillaman:
Fixes: http://tracker.ceph.com/issues/15097


trociny:
@dillaman Then may be also remove librbd::Journal::reset()? It is currently not used (I planned to used it for `rbd journal reset` but faced a linking issue). It also does remove/create, which as I understand doesn't work nice.

dillaman:
@trociny If it isn't used -- probably makes sense to delete it.

PR #8495, opened 2016-04-07
osd/ReplicatedPG::_rollback_to: update the OMAP flag

athanatos:
We don't simply copy over the whole flag field because not all values
should necessarily be carried over from the clone.  Particularly,
FLAG_CACHE_PIN should not be overridden.

Fixes: http://tracker.ceph.com/issues/14777
Backport: hammer, infernalis


athanatos:
Tested samuelj-2016-04-06_16:46:34-rados-wip-sam-testing-distro-basic-smithi

Failures appear to be unrelated to this patch.

PR #8496, opened 2016-04-07
os/filestore: force lfn attrs to be written atomically, restructure name length limits

athanatos:
Tested: samuelj-2016-04-06_16:46:34-rados-wip-sam-testing-distro-basic-smithi -- failures appear to be unrelated

ceph-qa-suite branch to follow

### DIFF
--- a/src/osd/ReplicatedPG.cc
+++ b/src/osd/ReplicatedPG.cc
@@ -1627,13 +1627,38 @@ void ReplicatedPG::do_op(OpRequestRef& op)
     return;
   }
 
+  hobject_t head(m->get_oid(), m->get_object_locator().key,
+		 CEPH_NOSNAP, m->get_pg().ps(),
+		 info.pgid.pool(), m->get_object_locator().nspace);
+
   // object name too long?
-  unsigned max_name_len = MIN(g_conf->osd_max_object_name_len,
-                              osd->osd->store->get_max_object_name_length());
-  if (m->get_oid().name.size() > max_name_len) {
-    dout(4) << "do_op '" << m->get_oid().name << "' is longer than "
-            << max_name_len << " bytes" << dendl;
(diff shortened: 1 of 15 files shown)


athanatos:
DNM quite yet

src/osd/ReplicatedPG.cc
```diff
@@ -1627,16 +1627,29 @@ void ReplicatedPG::do_op(OpRequestRef& op)
...
-  if (m->get_oid().name.size() > max_name_len) {
-    dout(4) << "do_op '" << m->get_oid().name << "' is longer than "
-            << max_name_len << " bytes" << dendl;
+  uint64_t name_length = m->get_oid().name.size() +
+    m->get_object_locator().key.size() +
+    m->get_object_locator().nspace.size();
+  if (name_length > g_conf->osd_max_object_name_len) {
+    dout(4) << "do_op name+namespace+locator is longer than "
+            << g_conf->osd_max_object_name_len
+	    << " bytes" << dendl;
```
liewegas: I don't think this hunk is necessary if the ObjectStore has its own check.  Also, I don't think it makes sense.  If the max length is 16, then "0123456789abcdef" is a valid name.  I should be able to specify an equally long key so that I can make "foo" stored next to it...

liewegas: Instead, I would separately enforce that key is <= max_object_name_len.  And either use the same for namespace, or specify a new tunable that restricts it.

athanatos:
DNM: possibly failed test

athanatos:
http://pulpito.ceph.com/samuelj-2016-04-10_21:23:13-rados-wip-sam-testing-distro-basic-smithi

Only bug from that run, seems to be unrelated:
http://tracker.ceph.com/issues/15458

Ready for review again.

athanatos:
@liewegas ping

PR #8502, opened 2016-04-08
crushtool: do not allow to compile crushmap with repeat ruleset.

songbaisen:
crushtool: do not allow to compile crushmap with repeat ruleset.

### DIFF
--- a/src/crush/CrushCompiler.cc
+++ b/src/crush/CrushCompiler.cc
@@ -608,6 +608,10 @@ int CrushCompiler::parse_rule(iter_t const& i)
   }
 
   int ruleset = int_node(i->children[start]);
+  if (ruleset_name.count(ruleset)) {
+    err << "ruleset '" << ruleset << "' already defined in rule '" << ruleset_name[ruleset] <<"'\n"<< std::endl;
+    return -1;
+  }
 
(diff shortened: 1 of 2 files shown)


src/crush/CrushCompiler.cc
```diff
@@ -608,6 +608,10 @@ int CrushCompiler::parse_rule(iter_t const& i)
   }
 
   int ruleset = int_node(i->children[start]);
+  if (ruleset_name.count(ruleset)) {
+    err << "ruleset '" << ruleset << "' already defined in rule '" << ruleset_name[ruleset] <<"'\n"<< std::endl;
+    return -1;
```
tchaikov: in `CrushCompiler::parse_crush()`, only the last returned `r` is returned. so in the test case of src/test/cli/crushtool/check-overlapped-rules.t, a partial compiled crush map is generated. if user is using a script to generate the map, he/she might not notice this problem.

songbaisen: @tchaikov Hi! I think we want to do the same thing.But I  forbid the repeat  ruleset earlier.If the user have define the same ruleset I will forbid  him to generate crushmap from crushmap.txt.

PR #8507, opened 2016-04-08
rgw-admin: remove unused iterator and fix error message

kmroz:


mattbenjamin:
@oritwas don't see any issue with these...

PR #8509, opened 2016-04-08
DNM rgw-admin: remove misleading error message

kmroz:
In RGWZoneGroup::remove_zone(), we already log a message when trying to
remove a non-existent zone from a zonegroup. No need to treat this as an
error condition.


kmroz:
Nonsense - we don't treat ENOENT as a true error. Closing.

PR #8510, opened 2016-04-08
DNM os/bluestore: New data structures and read/write/zero/truncate handler implementations embedded into Bluestore

ifed01:
This is a prototype code implementing lextent/blob/pextent structure triple at bluestore to support compression/checksum verification. This patch introduces data structure modifications and read/write/zero/fiemap/clone op implementations along with corresponding UT coverage.
Code integrates into Bluestore and replaces old data structures for all object operations.
ceph_test_objectstore test case PASSED.

Open tasks:
- Remove legacy WAL structures and related code, most of other legacy stuff has been already removed.
- Add deferred processing to implement merge procedures that require R/M/W sequence
- Add calls and settings for compressor/checksum verification

### DIFF
--- a/src/os/bluestore/ExtentManager.h
+++ b/src/os/bluestore/ExtentManager.h
@@ -15,7 +15,7 @@
  */
 
 #ifndef CEPH_OSD_EXTENT_MANAGER_H
-#define CEPH_OSD_EXTENT_ANAGER_H
+#define CEPH_OSD_EXTENT_MANAGER_H
 
 #include <list>
 #include <map>
@@ -26,38 +26,66 @@
 class ExtentManager{
 
 public:
+  struct CheckSumInfo{
+    uint8_t csum_type;               //  enum bluestore_blob_t::CSumType
+    uint8_t csum_block_order;
+    CheckSumInfo() : csum_type(bluestore_blob_t::CSUM_NONE), csum_block_order(12) {}
+    CheckSumInfo(uint8_t _type, uint8_t _csum_block_order) : csum_type(_type), csum_block_order(_csum_block_order) {}
+  };
 
-  struct DeviceInterface
+  struct CompressInfo{
+    uint8_t compress_type;
+    CompressInfo() : compress_type(0xff) {}
+  };
+
+  struct BlockOpInterface
   {
-    virtual ~DeviceInterface() {}
+    virtual ~BlockOpInterface() {}
     virtual uint64_t get_block_size() = 0;
 
     virtual int read_block(uint64_t offset, uint32_t length, void* opaque, bufferlist* result) = 0;
+    virtual int write_block(uint64_t offset, const bufferlist& data, void* opaque) = 0;
+    virtual int zero_block(uint64_t offset, uint32_t length, void* opaque) = 0;
+
+    //method to allocate pextents, depending on the store state can return single or multiple pextents if there is no contiguous extent available
+    virtual int allocate_blocks(uint32_t length, void* opaque, bluestore_extent_vector_t* result) = 0;
+
+    virtual int release_block(uint64_t offset, uint32_t length, void* opaque) = 0;
 
   };
+
   struct CompressorInterface
   {
     virtual ~CompressorInterface() {}
+    virtual int compress(const CompressInfo& cinfo, uint32_t source_offs, uint32_t length, const bufferlist& source, void* opaque, bufferlist* result) = 0;
     virtual int decompress(const bufferlist& source, void* opaque, bufferlist* result) = 0;
   };
   struct CheckSumVerifyInterface
   {
     virtual ~CheckSumVerifyInterface() {}
-    virtual int verify(bluestore_blob_t::CSumType, uint32_t csum_block_size, const vector<char>& csum_data, const bufferlist& source, void* opaque) = 0;
+
+    virtual int calculate(bluestore_blob_t::CSumType, uint32_t csum_value_size, uint32_t csum_block_size, uint32_t source_offs, uint32_t source_len, const bufferlist& source, void* opaque, vector<char>* csum_data) = 0;
+    virtual int verify(bluestore_blob_t::CSumType, uint32_t csum_value_size, uint32_t csum_block_size, const bufferlist& source, void* opaque, const vector<char>& csum_data ) = 0;
   };
 
-  ExtentManager(DeviceInterface& device, CompressorInterface& compressor, CheckSumVerifyInterface& csum_verifier)
-    : m_device(device), m_compressor(compressor), m_csum_verifier(csum_verifier) {
+
+  ExtentManager(BlockOpInterface& blockop_inf, CompressorInterface& compressor, CheckSumVerifyInterface& csum_verifier)
+    : m_blockop_inf(blockop_inf), m_compressor(compressor), m_csum_verifier(csum_verifier) {
   }
 
-  int write(uint64_t offset, uint32_t length, void* opaque, const bufferlist& bl);
+  int write(uint64_t offset, const bufferlist& bl, void* opaque, const CheckSumInfo& check_info, const CompressInfo* compress_info);
+  int zero(uint64_t offset, uint64_t length, void* opaque);
+  int truncate(uint64_t offset, void* opaque);
   int read(uint64_t offset, uint32_t length, void* opaque, bufferlist* result);
+
+  uint64_t get_max_blob_size() const;
+  uint64_t get_min_alloc_size() const;
 
 protected:
 
   bluestore_blob_map_t m_blobs;
   bluestore_lextent_map_t m_lextents;
-  DeviceInterface& m_device;
+  BlockOpInterface& m_blockop_inf;
   CompressorInterface& m_compressor;
   CheckSumVerifyInterface& m_csum_verifier;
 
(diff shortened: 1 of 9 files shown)


allensamuels:
The data structures look great in general -- one issue comes to mind after reading the code (BTW, it's great to have BOTH of them to review at the same time):

In the non-compressed world, if you're reading a portion of a blob you will need to linearly scan the pextent vector to find the portions that you want to read. This could be VERY time consuming for a large object that was encoded as a small number of blobs. I think the right solution here is to leave the data structures alone and to assume (enforce) that the write path simply doesn't create this situation. In other words, when appending to an existing part of an object, the write path will have to balance extending the existing blob with creating a new blob in order to avoid long linear searches in the read path. This is probably a trivial thing, but I thought I'd mention here to make a record of it.

----

When looking into the checksum case, it wasn't clear if the following case was handled correctly:

(1) no compression
(2) checksum covers more than one physical block. (i.e., say, a checksum for every TWO blocks on the device)
(3) requested read starts in the middle of a checksum-area

In this case, you'll need to read more of the data of a physical extent and discard that extra data AFTER doing the checksum computation. 

I believe you correctly handle the case where the length of the read doesn't match the checksum-area size, but I didn't see how that was handled with it was the offset that didn't match.

----

I think the read path doesn't care what the blocksize of the device is. It just issues the reads that need to happen to get the job done. In particular when checksums are enabled, you may need to read data in units of the checksum-area. This is controlled by the checksums that are encoded in the onode, not the device block size. I believe this is the same situation as the linear scan described above, i.e., you rely on the write path to do the correct thing. The read path doesn't really have any options, it's a slave to the write path's previous choices.

----

I'm a bit concerned about some of the raw blob pointers that are passed around. Since blobs can be shared (in an enode), I'm concerned about the underlying blob storage data structure. In other words, I'm worried about the rules for loading/unloading of enodes affecting this code. (Maybe I just need to study that part of BlueStore better). I believe that Sage has stated in in the past that the range of the clone operation is restricted in such a way as to ensure that all clients of an enode have the same hash code, meaning they are in the same PG, which *might* be enough to prevent premature unloading of enodes -- but you're not immune to you own enode load/unloads. But we've also talked about moving the PG lock to be a RW lock rather than an exclusive (I think the SK telecom guys identified this as a serious performance bottleneck for multiple reads) which would likely cause this to be a significant issue.

----
One minor note. I found this code pattern at least twice (maybe thrice?)

xyz::iterator it = container.find(key);
...
if (it != container.end()) {
   it-><do something>;
} else {
   container[key].<do_samething>;
}

I personally found it minor-confusing and probably slower than the code that I feel is more obvious:

xyz& value = container[key];
...
value.<do something>

----

One -- possibly minor -- performance optimization. The decompressor code doesn't know how much data out of the blob will actually be used. If it had this knowledge, it might be able to prematurly terminate the decompression. Not all decompressors have that ability -- but some do and it would be ashame to not take advantage of it.

----

Here's a nastier one. This code is written to be blocking, i.e., synchronous w.r.t. each user-level I/O operation. With a movement toward thread-per-core/run-to-completion code we're going to want to make it asynchronous.

ifed01:
@allensamuels please find some replies below (others to follow):

> When looking into the checksum case, it wasn't clear if the following case was handled correctly:
This has to work properly. There is a UT to cover the case. Please see TEST(bluestore_extent_manager, read_checksum)
 ...
  mgr.read(0x7000, 0x4000, NULL, &res);
that results in read 0x6000~2000 (checksum block size is 0x2000)
  ASSERT_TRUE(mgr.checkRead(ReadTuple(0x6000, 0x2000)));

and compare to the same call at
TEST(bluestore_extent_manager, read)
  ASSERT_TRUE(mgr.checkRead(ReadTuple(0x7000, 0x1000)));
--------------------------------------------------------------------------------
> I think the read path doesn't care what the blocksize of the device is.

Currently there is a generic method that determines the read block size:
ExtentManager::get_read_block_size

Depending on the checksum mode is on or off it returns actual block size that read handler has to use: either device block size or checksum one stored for the blob. In that regard Extent manager is almost block size agnostic.
We definitely can make a small step forward and force write handler to store device block size for blobs that lack checksum but IMHO there is no much value in that. 
---------------------------------------------------------------------------------------
>I'm a bit concerned about some of the raw blob pointers that are passed around. 
Please note that pointers in my code are used for temporary data structures. They exist during specific read handling ONLY. I.e. I don't store blob or pextent pointer longer than single client read is being processed. They aren't shared between multiple read calls. IMO in sync mode read operation should be performed under (read?)lock on the object thus using the pointers is safe. My suggestion is to have locks out of Extent Manager, e.g. before read method call. Thus we can add them when integrating EM to the rest of the code.  Async handling probably brings more challenge into this though. But I suggest to stabilize sync operation mode first.

---------------------------------------------------------------------------------------
>One minor note. I found this code pattern at least twice (maybe thrice?)
Fixed. Thanks for pointing out.

---------------------------------------------------------------------------------------
>One -- possibly minor -- performance optimization. The decompressor code doesn't know how >much data out of the blob will actually be used. 
Well, my view on the issue is as follow:
There are two lengths decompressor potentially cares of: blob length and compressed block length that is less or equal to blob length. In fact decompressor uses the second one while ExtentManager delivers data based on the first one. IMO most of the time (or always?) we'll have blob length = compressed block length as we allocate blobs for compression results. And EM implicitly provides blob length to the decompressor via decompress call - it's source.length(). 
If for some (unclear for me at the moment) reason these lengths diverge this should be treated out of EM - by the code that wraps actual compressor invocation. Such code can prefix compressed block with the length. It's supposed to do the same for comp. algorithm already.
Anyway I think EM should be agnostic to such specific details.

ifed01:
>Here's a nastier one. This code is written to be blocking, i.e., synchronous w.r.t. each user-level I/O operation. With a movement toward thread-per-core/run-to-completion code we're going to want to make it asynchronous.
>
In general I agree.
But I suggest to stabilize the sync mode of operation first - both read and write - and then proceed with the async one. Moreover it seems that we lacks async mode in Bluestore at the moment that gives another point to postpone such activity for a while.

ifed01:
@allensamuels 
>In the non-compressed world, if you're reading a portion of a blob you will need to linearly scan the pextent vector to find the portions that you want to read. 
>
As you mix both pextents and blobs let me reconfirm first - you are talking exactly about sequential pextent scan, not blob one, right? I.e. the case when single blob has multiple underlying pextents and determining proper pextent subset requires such a scan.
As far as I understand we can have such a case in quite a rare scenario you highlighted during dev-list discussion -  when there is no contiguous space at the device to allocate. And that was primarily for compressed scenario.
Hopefully resulting pextent vectors aren't that large. Anyway I like the idea to enforce write code to provide some protection for the case. But what's about a simple idea to do that by limiting blob size - as it's multiple of allocation unit we'll have such an enforcement automatically- max pextent vector size is max_blob_size / min_allocation_size. And we aren't planning to have huge blobs anyway for compressed case...

allensamuels:
>>I think the read path doesn't care what the blocksize of the device is.

>Currently there is a generic method that determines the read block size:
ExtentManager::get_read_block_size
Depending on the checksum mode is on or off it returns actual block size that read handler has to use: either device block size or checksum one stored for the blob. In that regard Extent manager is almost block size agnostic.
We definitely can make a small step forward and force write handler to store device block size for blobs that lack checksum but IMHO there is no much value in that.

When checksums are on, we agree that you have to use a blocksize that's a function of the checksum information. 

When checksums are off, why not just round the read up to the nearest direct I/O boundary (512 or 4K) and let the underlying device drivers deal with it?

ifed01:
@allensamuels 
>When checksums are on, we agree that you have to use a blocksize that's a function of the checksum information.
When checksums are off, why not just round the read up to the nearest direct I/O boundary (512 or 4K) and let the underlying device drivers deal with it?
>

And that's exactly what I tried to implement.
When checksum is on - blocksize is taken from the corresponding field of the blob structure. And it's writer responsibility to set it properly.
When checksum is off - block size to be retrieved from the device. Current implementation uses abstract DeviceInterface to enable UT but actual implementation to use the mentioned I/O boundary.

allensamuels:
>> One -- possibly minor -- performance optimization. The decompressor code doesn't know how much data out of the blob will actually be used.

> Well, my view on the issue is as follow:
There are two lengths decompressor potentially cares of: blob length and compressed block length that is less or equal to blob length. In fact decompressor uses the second one while ExtentManager delivers data based on the first one. IMO most of the time (or always?) we'll have blob length = compressed block length as we allocate blobs for compression results. And EM implicitly provides blob length to the decompressor via decompress call - it's source.length(). 
If for some (unclear for me at the moment) reason these lengths diverge this should be treated out of EM - by the code that wraps actual compressor invocation. Such code can prefix compressed block with the length. It's supposed to do the same for comp. algorithm already.
Anyway I think EM should be agnostic to such specific details.

IMO, there are three lengths: 
1 blob physical length -- length of blob on the physical device (amount of compressed data)
2 blob logical length -- length of blob when uncompressed
3 requested data length -- amount of data that must be uncompressed to satisfy the read operation (essentially offset + length of original read operation)

For decompressors that *must* process the entire input data stream (i.e., have no ability to stop prematurely) then (3) is irrelevant and the upper layers will simply have to discard the excess data.

For decompressors that do have the ability to stop prematurely you'll want to use (3) as the decompress length, not (2).

allensamuels:
>> In the non-compressed world, if you're reading a portion of a blob you will need to linearly scan the pextent vector to find the portions that you want to read.

> As you mix both pextents and blobs let me reconfirm first - you are talking exactly about sequential pextent scan, not blob one, right? I.e. the case when single blob has multiple underlying pextents and determining proper pextent subset requires such a scan.
As far as I understand we can have such a case in quite a rare scenario you highlighted during dev-list discussion - when there is no contiguous space at the device to allocate. And that was primarily for compressed scenario.
Hopefully resulting pextent vectors aren't that large. Anyway I like the idea to enforce write code to provide some protection for the case. But what's about a simple idea to do that by limiting blob size - as it's multiple of allocation unit we'll have such an enforcement automatically- max pextent vector size is max_blob_size / min_allocation_size. And we aren't planning to have huge blobs anyway for compressed case...

It's only rare if the write-path is aware of the potential danger of making a blob too large. In the append situation, the write-path has the choice of extending the last blob or creating a new one. As long as it is prevented from creating a blob that's too large, this is a non-problem.

allensamuels:
> When checksum is off - block size to be retrieved from the device. Current implementation uses abstract DeviceInterface to enable UT but actual implementation to use the mentioned I/O boundary.

My point is that this isn't a function of the device -- I think it's a constant 512 (essentially an operating system constant), though we might choose to increase it to a constant 4K or page size because of memory allocation issues.

src/os/bluestore/ExtentManager.h
```diff
@@ -0,0 +1,94 @@
...
+ *
+ * This is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License version 2.1, as published by the Free Software
+ * Foundation.  See file COPYING.
+ *
+ */
+
+#ifndef CEPH_OSD_EXTENT_MANAGER_H
+#define CEPH_OSD_EXTENT_ANAGER_H
```
liewegas: typo

liewegas:
1. I like the ExtentManager structure since it means we'll be able to unit test the code that updates the extent mappings.
2. Lots of style nits--I have a really hard time reading the code.  Easy to clean up.
3. Why rename Enode to Bnode?
4. Right now we only load up the enode if we need to update ref counts etc... Now the ExtentManager ctor needs it unconditionally.  I think we should make an upcall to get a ref for the mapping so that it's only loaded when necessary.
5. I think we'll want to instantiate the ExtentManager inside the Onode so that we don't have to (re)construct it on every operation.
6. So far this appears to do all io directly.. no wal operations.  I haven't read through the read/write paths carefully yet, but is that right?

I'm trying to figure out if we want to jump into a totally new code base for all of this or if we want to adapt any of the current code to the new structure (e.g., as an altnerate ExtentManager implementation).

liewegas:
I think we have to take an incremental approach here.

1. introduce new lextent, blob, etc types
2. convert existing code to use new types
3. modularize io path logic of existing implementation
4. introduce incremental changes or alternative implementation with new logic

ifed01:
>So far this appears to do all io directly.. no wal operations. I haven't read through the read/write paths carefully yet, but is that right?

@liewegas Yes. That's what I'm planning to do next

ifed01:
>I think we'll want to instantiate the ExtentManager inside the Onode so that we don't have to (re)construct it on every operation.

@liewegas EM construction is pretty trivial - just an assignment of several references. Thus I don't see any drawbacks to create it on the stack on every operation. Instead having it in Onode unnecessarily increases onode's memory footprint. Anyway final decision is up to you...

liewegas:
I worked on the first part yesterday:

https://github.com/ceph/ceph/pull/8928

I'm getting to the point of converting the write methods to the new 
structures, though, and I can't decide if it'll be better to move 
this code into ExtentManager first.  I have a bunch of cosmetic/cleanup 
suggestions for the interface.. let me reread the patches and see what 
makes sense.

What's your irc nick?

ifed01:
>Why rename Enode to Bnode
 @liewegas IMHO that's a completely different structure thus needed to name it somehow. And it was easier to integrate new code having both Enode and B node together for some time. Will easily rename though if needed.

liewegas:
:+1:

ifed01:
>Right now we only load up the enode if we need to update ref counts etc... Now the ExtentManager ctor needs it unconditionally. I think we should make an upcall to get a ref for the mapping so that it's only loaded when necessary.
@liewegas Not sure I understand your suggestion here. Please note however that contrary to Enode we create Bnode for any Onode,  not for ones that have clones. Bnode is not just a ref count tracker but an 'independent' owner for blob mapping. Thus we load Bnode with the first Onode that needs it and cache/lru it as usual.

liewegas:
I think we can keep the behavior where bnode is loaded only on demand, and we keep blobs in onode until they are shared.  Right?

liewegas:
Reading through more of this and now I'm thinking there isn't much of anything to save from the current read/write handling code.  But there is a lot of cleanup to do on the ExtentManager interface and code.

Let's start with the *Interface interfaces...

* Is there any point of having 3 separate interfaces, when in all cases it's the same upward-facing interface?
* We're going to need to add a cache in here at some point. I'm a bit worried the indirection here is going to make that harder.
* If we're breaking down the read process into the first step being to make a complete read plan, then doing IO, then decompress, then assemble result, then we can put the planning into a simple helper that is easy to unit test independently... without putting a harness around the part that does IO.  Similarly, we already have something like this on the write side in the form of the IOContext.  I think it might be simpler and cleaner than all the wrappers around block IO functions and opaque pointers getting past around.

ifed01:
>What's your irc nick

Just created - ifed01
Please note that I'm at vacation this week thus probably  wouldn't  respond promptly.

ifed01:
>I think we can keep the behavior where bnode is loaded only on demand, and we keep blobs in onode until they are shared. Right?

@liewegas yes we can. But making a share will be irreversible op. Once blobs are separated from the Onode there is no easy way to put them back since blob ref count is used for both shares and multiple lextents within a single object.
And such change will definitely  bring more complexity thus we need a strong argument for it.  Are you trying to avoid double read from kv this way?

liewegas:
Yeah, understood it's one-way.  Main concern is to avoid the 2x 
kv read in the common case of no sharing.

ifed01:
> Is there any point of having 3 separate interfaces, when in all cases it's the same upward-facing interface?

My points :
1. There is no strong  need  to implement checksumming and compression at bluestore class. Currently these are just stubs. We can have standalone final implementations for them for the sake of better modularity.
2. (Pretty academic) Single role - single entity. Thus different interfaces for different tasks.

ifed01:
>We're going to need to add a cache in here at some point. I'm a bit worried the indirection here is going to make that harder.

I need more information on your plans in this area to be able to provide any valuable feedback. However  my intuition says that one can put such a cache before or after the EM. Or even within it. Thus no much worries on that from my side.

ifed01:
>If we're breaking down the read process into the first step....

Will think about that....

ifed01:
>Yeah, understood it's one-way.  Main concern is to avoid the 2x
kv read in the common case of no sharing.

OK.  Will implement this way...

allensamuels:
If we’re going to re-write it. Let’s do it in a style that makes it easy to convert to a non-threaded, run to completion model (futures/promises??) That’s where the OSD is headed and this will be some of the first code that will want to get converted.


Allen Samuels
Software Architect, Fellow, Systems and Software Solutions
[cid:image001.png@01CF4C2D.A7116C90]

PR #8512, opened 2016-04-08
librbd: Fixed bug in disabling non-primary image mirroring

rjfd:
It is only possible to remove image from rbd_mirroring object if the
mirroring image state is set to DISABLING.

### DIFF
--- a/src/librbd/internal.cc
+++ b/src/librbd/internal.cc
@@ -312,13 +312,10 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force) {
     return r;
   }
 
-  if (!is_primary) {
-    if (!force) {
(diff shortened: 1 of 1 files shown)


src/librbd/internal.cc
```diff
@@ -318,7 +318,6 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force) {
         " disable mirroring" << dendl;
       return -EINVAL;
     }
-    goto remove_mirroring_image;
```
dillaman: Minor: can we collapse the nested-ifs down to a single if statement now that we don't need this 'goto'?

rjfd:
@dillaman done

PR #8513, opened 2016-04-08
fixup unit test for keyvaluedb. add merge test case

allensamuels:
Fix tracker# 15435


liewegas:
https://github.com/ceph/ceph/pull/8703

PR #8515, opened 2016-04-09
install-deps.sh: use 'pip install --upgrade' to prepare virtualenv

runsisi:
or the command ```pip install  'setuptools >= 0.8' 'pip >= 7.0' 'wheel >= 0.24'```
will always ignore the local configured pypi mirror and try to get updated pip
from https://bitbucket.org/ which results:

Downloading/unpacking pip>=7.0
  Could not fetch URL https://bitbucket.org/pypa/setuptools/downloads/ctypes-1.0.2.win32-py2.4.exe#md5=9092a0ad5a3d79fa2d980f1ddc5e9dbc: There was a problem confirming the ssl certificate: <urlopen error EOF occurred in violation of protocol (_ssl.c:765)>
  Will skip URL https://bitbucket.org/pypa/setuptools/downloads/ctypes-1.0.2.win32-py2.4.exe#md5=9092a0ad5a3d79fa2d980f1ddc5e9dbc when looking for download links for pip>=7.0

### DIFF
--- a/install-deps.sh
+++ b/install-deps.sh
@@ -110,7 +110,7 @@ function activate_virtualenv() {
     if ! test -d $env_dir ; then
         virtualenv --python $interpreter $env_dir
         . $env_dir/bin/activate
-        if ! populate_wheelhouse install ; then
+        if ! populate_wheelhouse "install --upgrade" ; then
             rm -rf $env_dir
             return 1
         fi


liewegas:
@alfredodeza @dmick

dmick:
wait...pip install ignores the mirror but pip install --upgrade does not?  That doesn't sound right at all; is that documented behavior?

dmick:
@runsisl did you see the question?

runsisi:
@dmick sorry i did not notice this :(
maybe this problem is specific to our build environment, the build machine can connect to the outside world only through the corporate proxy.
the install command, i.e., `pip install 'setuptools >= 0.8' 'pip >= 7.0' 'wheel >= 0.24'`, can complete successfully now but very slow, i think it tries the local configured pypi mirror as the last resort, if i add the option `--upgrade` then it works like a charm.
i only find a comment that states (see pypa/pip#1805)
```The best way would be to use your distro pip to install another pip version : pip install -U pip==1.4```
feel free to close this if you think this is not a issue :)

dmick:
but your comment is to upgrade pip itself, not to use upgrade while installing requirements....

I don't doubt there's something going wrong, but it doesn't seem like there's yet a diagnosis, much less a clear idea that this change fixes it.

runsisi:
yes, maybe i need to do some more digging, so i will close it for now
thank your very much, @dmick :)

PR #8516, opened 2016-04-09
journal: fix final result for JournalTrimmer::C_RemoveSet

runsisi:
if the first journal object got removed successfully while later objects
failed then the final result should be set to fail


xiexingguo:
looks good

PR #8521, opened 2016-04-10
check-generated.sh: can't source bash from sh

stiopaa1:
'source' is not sh command and executing bash commands in sh won't
work. We could try to replace it with:
'. ../qa/workunits/ceph-helpers.sh'
but the above file uses bash.

### DIFF
--- a/src/test/encoding/check-generated.sh
+++ b/src/test/encoding/check-generated.sh
@@ -1,4 +1,4 @@
-#!/bin/sh -e
+#!/bin/bash -e
 
 source ../qa/workunits/ceph-helpers.sh
 


stiopaa1:
Not sure if we want to to keep check-generated.sh to use sh, If so I guess we need to change ceph-helpers.sh to to use sh as well?(which shouldn't be a lot of work and I'm happy to do it).

tchaikov:
> Not sure if we want to to keep check-generated.sh to use sh

thanks for fixing this @stiopaa1 ! i don't think it's worth your time to translate the bashism in `ceph-helpers.sh` to POSIX compatible counterparts. @dachary what do you think?

tchaikov:
lgtm

dachary:
@tchaikov @stiopaa1 I think it's safe to assume all contexts in which ceph runs have bash

wjwithagen:
@dachary @tchaikov @stiopaa1 
Last time I discussed this with @liewegas when I wanted /bin/sh for some FreeBSD stuff, The conclusion is that bash is the most steady denominator even in all the Linux distros, because there are distros where sh equals to dash. Which has its on set of problems.

tchaikov:
@wjwithagen thanks for your insights.

pushed to gitbuilder as wip-15426 for testing.

PR #8523, opened 2016-04-10
cmake: pass RULE_LAUNCHER_* to cython

tchaikov:
* pass the environment variables using `env` to pass environment
  varibles with space(s) in them to the COMMAND in add_custom_target.
  otherwise, cmake will try to quote the space with "\". this breaks the
  generated command line.
* add a comment for ccache to note that we do not expect ccache to speed
  up linking. we use it as the linker's launcher to workaround
  https://bugs.python.org/issue8027. to be specific,
  distutils.UnixCCompiler.link overwrites the first linker CLI's arg
  using the the first C++ compiler's CLI arg, if "env" is not used to
  launch the linker. this breaks the cythonization of our pybind APIs.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,6 +25,9 @@ if(WITH_CCACHE)
   if(CCACHE_FOUND)
     message(STATUS "Building with ccache: ${CCACHE_FOUND}, CCACHE_DIR=$ENV{CCACHE_DIR}")
     set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE ccache)
+    # ccache does not accelerate link (ld), but let it handle it. by passing it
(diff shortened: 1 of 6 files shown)


robbat2:
@tchaikov  See https://github.com/ceph/ceph/pull/8461#issuecomment-208034584 your patch doesn't effect ccache C++ linking.

tchaikov:
build passed at http://gitbuilder.sepia.ceph.com/gitbuilder-ceph-tarball-trusty-amd64-basic/log.cgi?log=fc81bec888eba3be488c34508d4ae3521372b175

jdurgin:
lgtm

tchaikov:
@liewegas @athanatos can we include this change in jewel?

PR #8525, opened 2016-04-10
ceph-helpers.sh: only use mon*pid files when killing MONs

wjwithagen:
FreeBSD once in a while forgets to remove *pid files (this is probably a bug).
But taking care of it this way is probably much in line of what is actually needs to be done

### DIFF
--- a/qa/workunits/ceph-helpers.sh
+++ b/qa/workunits/ceph-helpers.sh
@@ -293,7 +293,7 @@ function test_kill_daemons() {
     #
     # kill the mon and verify it cannot be reached
     #
-    kill_daemons $dir TERM || return 1
+    kill_daemons $dir TERM mon || return 1
     ! ceph --connect-timeout 60 status || return 1
     teardown $dir || return 1
 }


qa/workunits/ceph-helpers.sh
```diff
@@ -293,7 +293,7 @@ function test_kill_daemons() {
     #
     # kill the mon and verify it cannot be reached
     #
-    kill_daemons $dir TERM || return 1
+    kill_daemons $dir TERM mon || return 1
```
tchaikov: `kill_daemons` does not accept a forth parameter, see
```sh
# @param pid the process id to send a signal
# @param send_signal the signal to send
# @param delays sequence of sleep times before failure
```

or you need to filter the output of `find $dir -name "*.pid" 2>/dev/null` with a prefix of "mon" so it only kills `mon.*pid` under the testdir?

wjwithagen:
@tchaikov 
That is weird. This is code I first tried to use, but then actually opted for waht you are suggesting:
Fix the find in the second run and make it select *mon*pid files only.
I'll close this one and submit the correct one again.

wjwithagen:
@tchaikov #8530 is a new attempt.

PR #8530, opened 2016-04-11
ceph-helpers.sh: only use mon*pid files when killing MONs

wjwithagen:
FreeBSD once in a while forgets to remove *pid files (this is probably a bug).
But taking care of it this way is probably much in line of what is actually needs to be done


tchaikov:
lgtm

PR #8531, opened 2016-04-11
librados/TestCase: skip tests if SetUpTestCase() fails

tchaikov:
see https://github.com/google/googletest/issues/247, gtest does not fail
its tests even if SetUpTestCase() fails. after shutdown(), the behaviour
of RadosClient is undefined. so we'd better avoid using it anymore in
that case.

### DIFF
--- a/src/test/librados/TestCase.cc
+++ b/src/test/librados/TestCase.cc
@@ -13,16 +13,26 @@ rados_t RadosTestNS::s_cluster = NULL;
 void RadosTestNS::SetUpTestCase()
 {
   pool_name = get_temp_pool_name();
-  ASSERT_EQ("", create_one_pool(pool_name, &s_cluster));
+  auto res = create_one_pool(pool_name, &s_cluster);
+  if (!res.empty()) {
+    s_cluster = NULL;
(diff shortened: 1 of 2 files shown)


liewegas:
this breaks ceph_test_rados_api_misc


2016-04-24T23:46:24.650 INFO:tasks.workunit.client.0.smithi045.stdout:                 api_misc: [ RUN      ] LibRadosTwoPoolsECPP.CopyFrom
2016-04-24T23:46:24.650 INFO:tasks.workunit.client.0.smithi045.stdout:                 api_misc: test/librados/TestCase.cc:618: Failure
2016-04-24T23:46:24.651 INFO:tasks.workunit.client.0.smithi045.stdout:                 api_misc: Failed
2016-04-24T23:46:24.651 INFO:tasks.workunit.client.0.smithi045.stdout:                 api_misc: cluster not connected, skipping this tests
2016-04-24T23:46:24.651 INFO:tasks.workunit.client.0.smithi045.stdout:                 api_misc: [  FAILED  ] LibRadosTwoPoolsECPP.CopyFrom (2271 ms)

PR #8532, opened 2016-04-11
osd: miscellaneous fixes

xiexingguo:


xiexingguo:
@liewegas Batch of core-related fixes and cleanups. Please help to review, thanks:)

xiexingguo:
@liewegas Set milestone to jewel for attention.  Reset if these changes turn out to be irrelevant.

liewegas:
Several of these I'm nervous about (the locking change, map epoch change, etc).  Let's backport as needed once jewel is out.

src/osd/PG.cc
```diff
@@ -5944,7 +5944,7 @@ boost::statechart::result PG::RecoveryState::Peering::react(const AdvMap& advmap
     post_event(advmap);
     return transit< Reset >();
   }
-  
+  pg->check_full_transition(advmap.lastmap, advmap.osdmap);
```
liewegas: The Peering and Active states are both sub-states of Started, which means the check should already get handled there, right?

src/osd/PG.cc
```diff
@@ -3569,22 +3569,6 @@ void PG::sub_op_scrub_unreserve(OpRequestRef op)
   clear_scrub_reserved();
 }
 
-void PG::sub_op_scrub_stop(OpRequestRef op)
```
liewegas: @tchaikov no callers for this now.. but is this something we might want later?  e.g., if 'noscrub' is set on the cluster or something?

src/osd/PG.cc
xiexingguo: Oops, that's true and I am going to drop this change now. And thanks for the review:)

src/osd/PG.cc
tchaikov: @liewegas, sorry for the late response. this change was introduced by @jdurgin in 378f84c1a65875d32ac6b8802a245f1d8de06373, but i don't quite understand how we can stop the ongoing scrub by clearing `scrubber.reserved`.

PR #8533, opened 2016-04-11
rgw_rest_admin: return -EINVAL for improper user info requests

theanalyst:
for /admin/user GET requests, if a uid is not given, we ultimately
return an AccessDenied error, as we initialize an anon user by default
where operations like key initialization will throw an -EACCESS. Since
the actual failure here is that uid was not specified, return an
-EINVAL, HTTP 400 error instead which is more clearer.

Fixes: http://tracker.ceph.com/issues/15455

### DIFF
--- a/doc/radosgw/adminops.rst
+++ b/doc/radosgw/adminops.rst
@@ -238,7 +238,7 @@ Request Parameters
 :Description: The user for which the information is requested.
 :Type: String
 :Example: ``foo_user``
-:Required: No
+:Required: Yes
 
 
 Response Entities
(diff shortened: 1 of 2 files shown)


theanalyst:
The jenkins build failure seems to be on 
```
FAIL: test/erasure-code/test-erasure-eio.sh
FAIL: test/erasure-code/test-erasure-code.sh
```
which shouldn't be related to this changeset

theanalyst:
test this please

theanalyst:
Even if we want to return `-EACCESS`, we should probably return when we find `uid.empty()` itself, no need to go through and fail later I guess.

PR #8535, opened 2016-04-11
hammer: rbd snap rollback: restore the link to parent

asheplyakov:
So snapshot, flatten, rollback of a cloned image does not loose any data

Fixes: #14512

### DIFF
--- a/src/librbd/internal.cc
+++ b/src/librbd/internal.cc
@@ -413,6 +413,48 @@ int validate_pool(IoCtx &io_ctx, CephContext *cct) {
     rados_completion->release();
   }
 
+  int rollback_parent(ImageCtx *ictx, uint64_t snap_id)
+  {
+    assert(ictx);
+    assert(ictx->parent_lock.is_locked());
+    assert(ictx->snap_lock.is_locked());
+
+    CephContext *cct = ictx->cct;
+    int r = 0;
+    std::map<librados::snap_t, SnapInfo>::const_iterator it = ictx->snap_info.find(snap_id);
+    if (it == ictx->snap_info.end()) {
+      ldout(cct, 10) << __func__ << ": no such snapshot: " << snap_id << dendl;
+      return -ENOENT;
+    }
+    const SnapInfo& snap_info(it->second);
+    if (ictx->parent_md == snap_info.parent) {
+      ldout(cct, 20) << __func__ << ": nop: head and snapshot have the same parent" << dendl;
+      return 0;
+    }
+    if (ictx->parent_md.spec.pool_id != -1) {
+       // remove the old parent link first, otherwise cls_client::set_parent
+       // will fail with -EEXISTS
+       ldout(cct, 20) << __func__ << ": removing the old parent link" << dendl;
+       r = cls_client::remove_parent(&ictx->md_ctx, ictx->header_oid);
+       if (r < 0) {
+         ldout(cct, 10) << __func__ << ": failed to remove parent link: "
+                        << cpp_strerror(r) << dendl;
+         return r;
+       }
+    }
+    if (snap_info.parent.spec.pool_id != -1) {
+      ldout(cct, 20) << __func__ << ": updating the parent link" << dendl;
+      r = cls_client::set_parent(&ictx->md_ctx, ictx->header_oid,
+                                 snap_info.parent.spec, snap_info.parent.overlap);
+      if (r < 0) {
+        ldout(cct, 10) << __func__ << ": failed to set parent link: "
+                       << cpp_strerror(r) << dendl;
+        return r;
+      }
+    }
+    return 0;
+  }
+
   int rollback_image(ImageCtx *ictx, uint64_t snap_id,
 		     ProgressContext& prog_ctx)
   {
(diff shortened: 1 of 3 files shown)


src/librbd/internal.cc
```diff
@@ -444,6 +444,30 @@ int validate_pool(IoCtx &io_ctx, CephContext *cct) {
...
+
+    {
+      RWLock::WLocker snap_locker(ictx->snap_lock);
+      RWLock::WLocker parent_locker(ictx->parent_lock);
+      std::map<librados::snap_t, SnapInfo>::const_iterator it = ictx->snap_info.find(snap_id);
+      if (it != ictx->snap_info.end()) {
+          const SnapInfo& si(it->second);
+          // current parent does not match the one of the snapshot being reverting to
+          bool should_update_parent = ictx->parent_md.spec != si.parent.spec ||
+                  ictx->parent_md.overlap != si.parent.overlap;
```
dillaman: If the overlap changes (but parent record still exists), the cls `set_parent` method will return `-EEXISTS` when trying to update the parent record.  Therefore, you will need to `remove_parent` first.  Also need to support the case where HEAD is associated with a parent but the rollback snapshot doesn't have a parent.

src/test/librbd/test_snapshot_flatten_rollback_cloned.py
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env python
```
dillaman: Can you move this test to `test_rbd.py`?  Probably enough to just verify parent spec after rollback.  Would also be good to add tests for various combinations of updates (no parent->parent, parent->no parent, overlap size change, etc).

asheplyakov:
@dillaman

> If the overlap changes (but parent record still exists), the cls set_parent method will return -EEXISTS when trying to update the parent record. Therefore, you will need to remove_parent first.

Done.

> Also need to support the case where HEAD is associated with a parent but the rollback snapshot doesn't have a parent.

Done.

> Can you move this test to test_rbd.py?

Done.

> Probably enough to just verify parent spec after rollback. 

I've added parent info checks, however think it's important to verify if the actual data matches the expected one.

> Would also be good to add tests for various combinations of updates (no parent->parent, parent->no parent, overlap size change, etc).

I've implemented no parent -> parent, parent -> no parent tests, however I don't quite understand which operations can change the overlap size. Any hints, please?

dillaman:
@asheplyakov The only way to change the overlap size is to shrink the image.  So if you resize the image down past the original parent size, the overlap is now smaller.

asheplyakov:
@dillaman 
> The only way to change the overlap size is to shrink the image. 

I've added a test which verifies that rollback handles such a modification. Does the patch look OK now?

src/librbd/internal.cc
```diff
@@ -413,6 +413,58 @@ int validate_pool(IoCtx &io_ctx, CephContext *cct) {
...
+  int rollback_parent(ImageCtx *ictx, uint64_t snap_id)
+  {
+    assert(ictx);
+    assert(ictx->parent_lock.is_locked());
+    assert(ictx->snap_lock.is_locked());
+
+    CephContext *cct = ictx->cct;
+    int r = 0;
+    std::map<librados::snap_t, SnapInfo>::const_iterator it = ictx->snap_info.find(snap_id);
+    assert(it != ictx->snap_info.end());
```
dillaman: Probably sufficient to just return ```-ENOENT``` in this case.

src/librbd/internal.cc
```diff
@@ -413,6 +413,58 @@ int validate_pool(IoCtx &io_ctx, CephContext *cct) {
...
+    const SnapInfo& snap_info(it->second);
+    if (snap_info.parent.spec.snap_id == CEPH_NOSNAP) {
```
dillaman: Minor: while technically correct to check ```snap_id == CEPH_NOSNAP```, since the rest of librbd uses ```pool_id == -1``` for the same check, it might make it clearer (here and below)

src/librbd/internal.cc
```diff
@@ -413,6 +413,58 @@ int validate_pool(IoCtx &io_ctx, CephContext *cct) {
...
+      if (ictx->parent_md.spec.snap_id == CEPH_NOSNAP) {
```
dillaman: Minor: instead of the nested ```if```'s, can we just test:

```
if (ictx->parent_md == snap_info.parent) { 
  return 0;
}

if (ictx->parent_md.pool_id != -1) {
  ... remove parent ...
}
if (snap_info.parent.pool_id != -1) {
  ... set parent ...
} 
```

src/test/pybind/test_rbd.py
```diff
@@ -1048,3 +1048,96 @@ def test_follower_write(self):
             for offset in [0, IMG_SIZE / 2]:
                 read = image2.read(offset, 256)
                 eq(data, read)
+
+
+class TestRollbackTricky(object):
```
dillaman: s/TestRollbackTricky/TestCloneRollback/

src/librbd/internal.cc
asheplyakov: Done

src/librbd/internal.cc
asheplyakov: Done

src/test/pybind/test_rbd.py
asheplyakov: Done

src/librbd/internal.cc
asheplyakov: Done

dillaman:
lgtm

dillaman:
http://pulpito.ceph.com/jdillaman-2016-05-11_07:32:22-rbd-wip-jd-testing---basic-mira/

dillaman:
👍 -- good to merge once 0.94.7 is released.

PR #8537, opened 2016-04-11
librados: fix to make external functions visible in librados.so

rjfd:
This PR is motivated by the problems reported in PRs #8476 #8203.

The specific problem with lockdep static global variables occurs because the exported lockdep functions are not externally visible in librados.so. Therefore, when a program is statically linked with common/lockdep.cc and dynmically linked with librados.so, it will create two versions of the lockdep's static global variables.

The reason why lockdep's functions are not externally visible in librados.so is due to the linker option:
`-Xcompiler -Xlinker -Xcompiler '--exclude-libs=ALL'`

I don't see a significant increase in size of the the shared library file when this option is removed. So is this option critical to some other issue?

### DIFF
--- a/src/common/Makefile.am
+++ b/src/common/Makefile.am
@@ -174,8 +174,10 @@ LIBCOMMON_DEPS += -lrt -lblkid
 endif # LINUX
 
 libcommon_la_SOURCES = common/buffer.cc
-libcommon_la_LIBADD = $(LIBCOMMON_DEPS)
-noinst_LTLIBRARIES += libcommon.la
+libcommon_la_LIBADD = $(LIBCOMMON_DEPS) $(CRYPTO_LIBS)
+libcommon_la_CXXFLAGS = ${AM_CXXFLAGS}
+libcommon_la_LDFLAGS = ${AM_LDFLAGS} -version-info 1:0:0
+lib_LTLIBRARIES += libcommon.la
 
 noinst_HEADERS += \
 	common/BackTrace.h \
--- a/src/include/buffer.h
+++ b/src/include/buffer.h
@@ -270,7 +270,7 @@ namespace buffer CEPH_BUFFER_API {
 
   private:
     template <bool is_const>
-    class CEPH_BUFFER_DETAILS iterator_impl
+    class CEPH_BUFFER_API iterator_impl
       : public std::iterator<std::forward_iterator_tag, char> {
     protected:
       typedef typename std::conditional<is_const,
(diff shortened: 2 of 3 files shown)


dillaman:
@rjfd The goal is to only export symbols that are part of the public API -- and avoid exporting things that are private.

rjfd:
@dillaman I think that is achieved by the -fvisibility=hidden compiler option, right?

dillaman:
@rjfd ... but we don't want symbols from linked-in static libraries to be exported either (since they are not public).  

Currently:
> $ nm -D .libs/librados.so | grep " T " | wc -l
> 670

Proposed change:
> $ nm -D .libs/librados.so | grep " T " | wc -l
> 3184

rjfd:
@dillaman ok -- let me check how can I only export the lockdep functions, without exporting everything else.

tchaikov:
@rjfd i think the real question is, are lockdep functions supposed to be exposed as part of librados interfaces?

rjfd:
@tchaikov I also agree that lockdep functions should not be part of librados interfaces. 
Why don't we distribute libcommon as a shared library, and avoid to statically linking libcommon with librados?

tchaikov:
you mean *distribute* libcommon to our user?

rjfd:
yes, although I don't quite know the implications of doing this (it might be a stupid thing :-p)

rjfd:
I pushed a new change that creates libcommon.so. What do you guys think?

liewegas:
Is lockdep the only thing with problematic global variables?  Since it's a developer-only tool I'm not sure it warrants all of the effort to switch around the packaging (which is a real headache).

rjfd:
What if we just take out lockdep from libcommon, and turn it into a dynamic lib? It will only be used in developer-mode.
I'm continuing the discussion around this problem because I think it is really annoying when someone hits this bug, which is not so easy to debug.

dillaman:
@rjfd Did you recently hit it again after the fix to rbd-mirror?

rjfd:
@dillaman no I haven't.

dillaman:
@rjfd I'm not sure it's worth the effort to break it out since it's future is in flux since more pieces of code are directly using C++11 locks and shipping an extra .so file that would only be used for development purposes is not ideal.

Since this isn't directly affecting rbd-mirror anymore but to avoid such pain in the future, perhaps the register/unregister commands could also take lock pointer so it can fail-fast if you attempt to unregister an id owned by a different lock.

athanatos:
@dillaman @rjfd @liewegas This doesn't seem worthwhile, I'll close it on my next pass unless there is consensus that something needs to be done.

liewegas:
This needs a respin with cmake anyway since we're about to kill autotools.

PR #8539, opened 2016-04-11
Wip Xio heavy messengers and resources

vuhuong:
Patch set to identify and allocate properly resources (number of portals/contexts, max number of connections per portal/context), then use all portals/contexts for both passive and active connections

### DIFF
--- a/src/msg/xio/XioMessenger.cc
+++ b/src/msg/xio/XioMessenger.cc
@@ -541,9 +541,9 @@ int XioMessenger::session_event(struct xio_session *session,
   case XIO_SESSION_CONNECTION_CLOSED_EVENT: /* orderly discon */
   case XIO_SESSION_CONNECTION_DISCONNECTED_EVENT: /* unexpected discon */
   case XIO_SESSION_CONNECTION_REFUSED_EVENT:
-    ldout(cct,2) << xio_session_event_types[event_data->event]
-      << " user_context " << event_data->conn_user_context << dendl;
     xcon = static_cast<XioConnection*>(event_data->conn_user_context);
+    ldout(cct,2) << xio_session_event_types[event_data->event]
+      << " xcon " << xcon << " session " << session  << dendl;
     if (likely(!!xcon)) {
       Spinlock::Locker lckr(conns_sp);
       XioConnection::EntitySet::iterator conn_iter =
@@ -565,9 +565,9 @@ int XioMessenger::session_event(struct xio_session *session,
     }
     break;
   case XIO_SESSION_CONNECTION_TEARDOWN_EVENT:
-    ldout(cct,2) << xio_session_event_types[event_data->event]
-      << " user_context " << event_data->conn_user_context << dendl;
     xcon = static_cast<XioConnection*>(event_data->conn_user_context);
+    ldout(cct,2) << xio_session_event_types[event_data->event]
+      << " xcon " << xcon << " session " << session << dendl;
     xcon->on_teardown_event();
     break;
   case XIO_SESSION_TEARDOWN_EVENT:
@@ -1004,7 +1004,7 @@ ConnectionRef XioMessenger::get_connection(const entity_inst_t& dest)
      * we can always set it explicitly */
     struct xio_connection_params xcp = {};
     xcp.session           = xcon->session;
-    xcp.ctx               = this->portals.get_portal0()->ctx;
+    xcp.ctx               = xcon->portal->ctx;
     xcp.conn_user_context = xcon;
 
     xcon->conn = xio_connect(&xcp);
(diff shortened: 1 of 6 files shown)


vuhuong:
Matt & Casey

Could you help to review this pull request?

src/msg/xio/XioPortal.h
```diff
@@ -131,28 +131,24 @@ class XioPortal : public Thread
...
+    pthread_mutex_init(&mtx, NULL);
 
-      /* associate this XioPortal object with the xio_context handle */
-      struct xio_context_attr xca;
-      xca.user_context = this;
-      xio_modify_context(ctx, &xca, XIO_CONTEXT_ATTR_USER_CTX);
+    struct xio_context_params ctx_params;
+    memset(&ctx_params, 0, sizeof(ctx_params));
+    ctx_params.user_context = this;
+    ctx_params.max_conns_per_ctx = max_conns;
```
cbodley: it's not obvious to me what accelio is doing with max_conns_per_ctx, and what happens when this is exceeded. adding a comment here would be really helpful

src/msg/xio/XioMessenger.cc
```diff
@@ -403,6 +409,32 @@ int XioMessenger::pool_hint(uint32_t dsize) {
...
+int XioMessenger::get_nportals(const string &mname)
+{
+  int nportals = 1;
+
+  if (mname == "client" || mname == "cluster")
+    nportals = max(cct->_conf->xio_portal_threads, 1);
+
+  return nportals;
+}
+
```
cbodley: making decisions based on the messenger's name seems prone to errors down the road. consider applications other than ceph-osd, where names like "client" and "cluster" would be common choices, but would have unexpected side-effects on the messenger's behavior. not to mention changes to these names within ceph-osd, where other developers would have no idea that XioMessenger would need to be updated as well

ideally, i think we'd want to use the `Messenger::Policy` to pass in generic hints about the messenger and its expected load, similar to the way that it handles message- and byte-throttling. the downside is that changes to the Messenger interface would need input and buy-in from more of the community

mattbenjamin: @cbodley, @vuhuong, @liewegas, @gregsfortytwo Agree with Casey this feels like policy, and more importantly, it doesn't seem like we want the messenger "names" in the interface.  Is there a way we can make some XioMessenger-specific policy that does the work explicitly, so we avoid names and also avoid changing shared policy structures (which we pretty much can't change 'till Kraken)?

yuyuyu101: looks reasonable and worth to give more hits!

src/msg/xio/XioPortal.h
vuhuong: it gives accelio a hint on total number of connections sharing the context's resources so that Accelio can allocate precisely its internal primary task pool... Without this setting, Accelio is choosing the default value which is not enough for *heavy* messengers on scaling and is too much for light messengers.

I'll add a comment.

vuhuong:
@mattbenjamin  To avoid specific names and/or changing shared policy structures, could we pass in a xio's specific list of properties or general parameters to Messenger::create(&prop_opts) something like:
ceph-osd, ceph-mon, ceph-mds... or any client that matter to xio/accelio resources will do:

Messenger::Propertylist ms_props;
ms_props.insert(Messenger::Property("xio_max_conns_per_portal", g_conf->xio_max_conns_per_portal));
msgr = Messenger::create(g_ceph_context, g_conf->ms_type, entity_name_t, "lname", getpid(), &ms_props);

and in XioMessenger:

XioMessenger::XioMessenger(Messenger::Propertylist &opts)
{
...
   if (opts) {
       pr_iter = opts->find("xio_max_conns_per_portal");
       max_conns = boost::get<int>(pr_iter->second);
   }
....
}

@cbodley I will revised the patch set to remove name specific for now since we need cct->_conf->xio_max_conns_per_portal passed in accelio's context creation and the second patch to spread both active and passive connections on portals

roidayan:
Hi all,
the 2 commits are really important for xio messenger stability. Vu updated the commit to remove the messenger name comparison. can you review again and consider to merge this pull request before kraken milestone?
thanks in advance.

cbodley:
@vuhuong @roidayan thanks. can you help clarify how `ctx_params.max_conns_per_ctx` works? is it just a hint for preallocation of resources, or will accelio actually fail new connections over that limit? (and if it fails, is that part of an existing/tested error path?)

roidayan:
it's a hint for accelio for preallocation of tasks in its primary pool. accelio will not check this to actually count number of connections.

vuhuong:
@cbodley it gives accelio a hint on total number of connections sharing the context's resources so that Accelio can allocate precisely its internal primary task pool... Without this setting, Accelio is choosing the default value which is not enough for heavy messengers on scaling and is too much for light messengers.

I already added the comment in the updated patchset

PR #8543, opened 2016-04-12
librbd: do not return a failure if a peer cannot be notified of update

dillaman:
Maintenance ops that fail to notify a peer of the update should not
report a failure to the original user.

Fixes: http://tracker.ceph.com/issues/15351


jdurgin:
lgtm

PR #8545, opened 2016-04-12
os/filestore: fix return type mismatch for lfn parse

xiexingguo:
According to https://github.com/ceph/ceph/pull/8496/commits/755c685f2d09c36b53bc39589a77fa73faba5149,
we shall return a int value instead of bool now.
So if we fail to parse an object, we shall reset "false" into "-EINVAL" instead,
as "false" is always translated into 0, which is supposed to indicate a success.


tchaikov:
lgtm

xiexingguo:
@tchaikov  Thanks for the quick review, and merge:)

athanatos:
lgtm as well

PR #8546, opened 2016-04-12
mds: fix auth caps with hyphen in path (#15465)

jcsp:
Quirk of the boost spirit char() fn, hyphens
are treated as range delimiters unless they're
the last character in the set.

Fixes: #15465


ukernel:
LGTM

jcsp:

PR #8547, opened 2016-04-12
common: drop a redundant warning about experimental features

zaitcev:
At first I thought perhaps to filter out blanket enablement with "*", or change the level of the warning. But what I found is a surprising redundancy, which makes this warning worthless to begin with! Thus this commit simply removes the silly thing. The commit has a longer explanation.

If anyone sees something I'm missing here, please comment.


liewegas:
:+1:

liewegas:
Note that this means we won't see a warning for features we don't use.  That's probably sane, though.

liewegas:
see https://github.com/ceph/ceph/pull/14170

PR #8550, opened 2016-04-12
objclass: honour do_osd_ops() results before going any further

xiexingguo:
This is because:
 1. all the other methods are doing the same.
 2. the origin results of do_osd_ops() may get overwritten by -ENOMEM,
    which is misleading.
 3. if the class methods return negative results, normally it's not
    the responsibility for the caller being in charge of freeing out
    buffers. Therefore this change saves our some efforts and avoids
    potential memory leaks(Though I doubt this shall happen).


liewegas:
lgtm

PR #8554, opened 2016-04-12
client: report root's quota in statfs

jcsp:
When user is mounted a quota-restricted inode
as the root, report that inode's quota status
as the filesystem statistics in statfs.

This allows us to have a fairly convincing illusion
that someone has a filesystem to themselves, when
they're really mounting a restricted part of
the larger global filesystem.

### DIFF
--- a/src/client/Client.cc
+++ b/src/client/Client.cc
@@ -8752,23 +8752,7 @@ int Client::statfs(const char *path, struct statvfs *stbuf)
   tout(cct) << "statfs" << std::endl;
 
   ceph_statfs stats;
-
-  Mutex lock("Client::statfs::lock");
-  Cond cond;
-  bool done;
-  int rval;
-
-  objecter->get_fs_stats(stats, new C_SafeCond(&lock, &cond, &done, &rval));
-
-  client_lock.Unlock();
-  lock.Lock();
-  while (!done)
-    cond.Wait(lock);
-  lock.Unlock();
-  client_lock.Lock();
-
   memset(stbuf, 0, sizeof(*stbuf));
-
   /*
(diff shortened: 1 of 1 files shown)


src/client/Client.cc
```diff
@@ -8779,15 +8763,51 @@ int Client::statfs(const char *path, struct statvfs *stbuf)
...
+
+  C_SaferCond cond;
+  objecter->get_fs_stats(stats, &cond);
+
+  client_lock.Unlock();
+  int rval = cond.wait();
+  client_lock.Lock();
+
+  if (cct->_conf->client_quota && root_ancestor
+      && root_ancestor->quota.max_bytes) {
```
ukernel: maybe we should use get_quota_root(root) instead of root_ancestor

liewegas:
Should this be a config option maybe to control whether it is the quota vs full fs limit?  We can default to the quota one...

gregsfortytwo:
Yeah, what Sage said.

Maybe even set it up as a server-side config that gets passed to the client on connect? Although it's not secure since the client still sees the whole cluster, it's more of a cluster admin decision than a user one.

src/client/Client.cc
jcsp: What is the advantage to doing that?  It's not obvious to me

ukernel: I think root_ancestor->quota is filesystem root's quota. It's possible that the mounted sub-directory or its ancestor directories have quota

liewegas: I think that's what root_ancestor is?  Either the mounted inode, or the 
first ancestor that has a quota set.

jcsp:
Superceded (by accident!) by https://github.com/ceph/ceph/pull/8747

PR #8555, opened 2016-04-12
msg/xio: simple fixes

avnerbh:


src/common/buffer.cc
```diff
@@ -659,7 +659,7 @@ static simple_spinlock_t buffer_debug_lock = SIMPLE_SPINLOCK_INITIALIZER;
   public:
     struct xio_reg_mem *mp;
     xio_mempool(struct xio_reg_mem *_mp, unsigned l) :
-      raw((char*)mp->addr, l), mp(_mp)
+      raw((char*)_mp->addr, l), mp(_mp)
```
mattbenjamin: ouch!

src/msg/xio/XioMessenger.cc
```diff
@@ -575,7 +575,7 @@ enum bl_type
 #define MAX_XIO_BUF_SIZE 1044480
 
 static inline int
-xio_count_buffers(buffer::list& bl, int& req_size, int& msg_off, int& req_off)
+xio_count_buffers(const buffer::list& bl, int& req_size, int& msg_off, int& req_off)
```
mattbenjamin: ok

src/msg/xio/XioMessenger.cc
```diff
@@ -620,7 +620,7 @@ xio_count_buffers(buffer::list& bl, int& req_size, int& msg_off, int& req_off)
 }
 
 static inline void
-xio_place_buffers(buffer::list& bl, XioMsg *xmsg, struct xio_msg*& req,
+xio_place_buffers(const buffer::list& bl, XioMsg *xmsg, struct xio_msg*& req,
 		  struct xio_iovec_ex*& msg_iov, int& req_size,
```
mattbenjamin: ok

src/msg/xio/XioConnection.cc
```diff
@@ -198,40 +198,40 @@ static inline XioDispatchHook* pool_alloc_xio_dispatch_hook(
   return xhook;
 }
 
-int XioConnection::on_msg_req(struct xio_session *session,
```
mattbenjamin: ok, thanks for the effort

src/msg/xio/XioConnection.cc
```diff
@@ -477,7 +477,7 @@ void XioConnection::msg_release_fail(struct xio_msg *msg, int code)
     " (" << xio_strerror(code) << ")" << dendl;
 } /* msg_release_fail */
 
-int XioConnection::flush_input_queue(uint32_t flags) {
```
mattbenjamin: ok--this is a comment on the whole commit--why is this a misleading name?
my naming had "input" queue for queuing before any attempt to call xio_send_msg, and "out" queue for msgs we need to retry on xio_send_msg (iirc)

src/msg/xio/XioConnection.cc
```diff
@@ -542,7 +542,7 @@ int XioConnection::discard_input_queue(uint32_t flags)
 	break;
       case XioSubmit::INCOMING_MSG_RELEASE:
 	deferred_q.erase(q_iter);
-	portal->release_xio_rsp(static_cast<XioRsp*>(xs));
```
mattbenjamin: ok, consistent with prior commit

mattbenjamin:
Hi @avnerbh ,

Mostly ACK, I only have the one question inline, about the naming of queues;  Interested in any feedback on how Xio sees things

avnerbh:
Hi @mattbenjamin 

Thanks for your care and for all your comments.

I changed the names because they confused me when I tried to understand the code in order to add features.

Regarding queues' names: in general, something that we **send** to another machine is usually called "**out**" traffic/queue and something that we **recv** from another machine is usually called  "**in**" traffic/queue. 

This terminology also matches ceph concept.  See for example:
**out_q in AsyncConnection**:
```
map<int, list<pair<bufferlist, Message*> > > out_q;  // priority queue for outbound msgs
```
**out_q in Pipe**/PipeConnection:
```
    map<int, list<Message*> > out_q;  // priority queue for outbound msgs
    DispatchQueue *in_q;
```

This also makes **XioConnection** to match accelio's terminology, for example, we place the **outgoing ceph message** and its headers in the "**out**" field of the **struct xio_msg**.  
See for example the following code snippet from **accelio** header files:
```
struct xio_msg {
	/* more fields... */
	struct xio_vmsg		in;		/**< incoming side of message */
	struct xio_vmsg		out;		/**< outgoing side of message */
	/* more fields... */
};
```

PR #8558, opened 2016-04-12
msg: mark daemons down on RST + ECONNREFUSED

branch-predictor:
When a daemon goes down (because it is killed or suicided because of assert)
on otherwise healthy machine, TCP stack will take care of its connections
and send RST (connection reset) packets to all daemons that were connected
to downed daemon. OSDs already handle that and attempt to reconnect, with
grace timer starting to count. When the grace timer runs out and daemons
still cannot reconnect, OSD in question is marked down.
This changeset adds additional handler (handle_refused()) to the dispatchers
and code that detects when connection attempt fails with ECONNREFUSED error
(connection refused) which is a clear indication that host is alive, but
daemon isn't, so daemons can instantly mark the other side as undoubtly
downed without the need for grace timer.
This changeset also adds more info to connections so figuring out which OSD
actually failed is a bit easier.
In current state, only OSDs take advantage of handle_refused() facility,
but I don't see why other daemons shouldn't.

Signed-off-by: Piotr Dałek git@predictor.org.pl

### DIFF
--- a/src/common/config_opts.h
+++ b/src/common/config_opts.h
@@ -815,6 +815,7 @@ OPTION(osd_op_history_duration, OPT_U32, 600) // Oldest completed op to track
 OPTION(osd_target_transaction_size, OPT_INT, 30)     // to adjust various transactions that batch smaller items
 OPTION(osd_failsafe_full_ratio, OPT_FLOAT, .97) // what % full makes an OSD "full" (failsafe)
 OPTION(osd_failsafe_nearfull_ratio, OPT_FLOAT, .90) // what % full makes an OSD near full (failsafe)
+OPTION(osd_fast_fail_on_connection_refused, OPT_BOOL, true) // immediately mark OSDs as down once they refuse to accept connections
 
 OPTION(osd_pg_object_context_cache_count, OPT_INT, 64)
 OPTION(osd_tracing, OPT_BOOL, false) // true if LTTng-UST tracepoints should be enabled
--- a/src/messages/MOSDFailure.h
+++ b/src/messages/MOSDFailure.h
@@ -24,22 +24,41 @@ class MOSDFailure : public PaxosServiceMessage {
   static const int HEAD_VERSION = 3;
 
  public:
+  enum {
+    FLAG_ALIVE = 0,      // use this on its own to mark as "I'm still alive"
+    FLAG_FAILED = 1,     // if set, failure; if not, recovery
+    FLAG_IMMEDIATE = 2,  // known failure, not a timeout
+  };
+  
   uuid_d fsid;
   entity_inst_t target_osd;
-  __u8 is_failed;
+  __u8 flags;
   epoch_t       epoch;
   int32_t failed_for;  // known to be failed since at least this long
 
   MOSDFailure() : PaxosServiceMessage(MSG_OSD_FAILURE, 0, HEAD_VERSION) { }
   MOSDFailure(const uuid_d &fs, const entity_inst_t& f, int duration, epoch_t e)
     : PaxosServiceMessage(MSG_OSD_FAILURE, e, HEAD_VERSION),
-      fsid(fs), target_osd(f), is_failed(true), epoch(e), failed_for(duration) { }
+      fsid(fs), target_osd(f),
+      flags(FLAG_FAILED),
+      epoch(e), failed_for(duration) { }
+  MOSDFailure(const uuid_d &fs, const entity_inst_t& f, int duration, 
+              epoch_t e, __u8 extra_flags)
+    : PaxosServiceMessage(MSG_OSD_FAILURE, e, HEAD_VERSION),
+      fsid(fs), target_osd(f),
+      flags(extra_flags),
+      epoch(e), failed_for(duration) { }
 private:
   ~MOSDFailure() {}
 
 public: 
   entity_inst_t get_target() { return target_osd; }
-  bool if_osd_failed() { return is_failed; }
+  bool if_osd_failed() const { 
+    return flags & FLAG_FAILED; 
+  }
+  bool is_immediate() const { 
+    return flags & FLAG_IMMEDIATE; 
+  }
   epoch_t get_epoch() { return epoch; }
 
   void decode_payload() {
@@ -49,9 +68,9 @@ public:
     ::decode(target_osd, p);
     ::decode(epoch, p);
     if (header.version >= 2)
-      ::decode(is_failed, p);
+      ::decode(flags, p);
     else
-      is_failed = true;
+      flags = FLAG_FAILED;
     if (header.version >= 3)
       ::decode(failed_for, p);
     else
@@ -63,14 +82,15 @@ public:
     ::encode(fsid, payload);
     ::encode(target_osd, payload, features);
     ::encode(epoch, payload);
-    ::encode(is_failed, payload);
+    ::encode(flags, payload);
     ::encode(failed_for, payload);
   }
 
   const char *get_type_name() const { return "osd_failure"; }
   void print(ostream& out) const {
     out << "osd_failure("
-	<< (is_failed ? "failed " : "recovered ")
+	<< (if_osd_failed() ? "failed " : "recovered ")
+	<< (is_immediate() ? "immediate " : "timeout ")
 	<< target_osd << " for " << failed_for << "sec e" << epoch
 	<< " v" << version << ")";
   }
--- a/src/msg/async/AsyncConnection.cc
+++ b/src/msg/async/AsyncConnection.cc
@@ -878,6 +878,10 @@ ssize_t AsyncConnection::_process_connection()
         r = cs.is_connected();
         if (r < 0) {
           ldout(async_msgr->cct, 1) << __func__ << " reconnect failed " << dendl;
+          if (r == -ECONNREFUSED) {
+            ldout(async_msgr->cct, 2) << __func__ << " connection refused!" << dendl;
+            dispatch_queue->queue_refused(this);
+          }
           goto fail;
         } else if (r == 0) {
           ldout(async_msgr->cct, 10) << __func__ << " nonblock connect inprogress" << dendl;
--- a/src/msg/simple/Pipe.cc
+++ b/src/msg/simple/Pipe.cc
@@ -924,9 +924,13 @@ int Pipe::connect()
   ldout(msgr->cct,10) << "connecting to " << peer_addr << dendl;
   rc = ::connect(sd, peer_addr.get_sockaddr(), peer_addr.get_sockaddr_len());
   if (rc < 0) {
-    rc = -errno;
+    int stored_errno = errno;
     ldout(msgr->cct,2) << "connect error " << peer_addr
-	     << ", " << cpp_strerror(rc) << dendl;
+	     << ", " << cpp_strerror(stored_errno) << dendl;
+    if (stored_errno == ECONNREFUSED) {
+      ldout(msgr->cct, 2) << "connection refused!" << dendl;
+      msgr->dispatch_queue.queue_refused(connection_state.get());
+    }
     goto fail;
   }
 
--- a/src/osd/OSD.cc
+++ b/src/osd/OSD.cc
@@ -4776,6 +4776,11 @@ bool OSD::ms_handle_reset(Connection *con)
   return true;
 }
 
+bool OSD::ms_handle_refused(Connection *con)
+{
+  return false;
+}
+
 struct C_OSD_GetVersion : public Context {
   OSD *osd;
   uint64_t oldest, newest;
--- a/src/osd/OSD.cc
+++ b/src/osd/OSD.cc
@@ -4778,7 +4778,33 @@ bool OSD::ms_handle_reset(Connection *con)
 
 bool OSD::ms_handle_refused(Connection *con)
 {
-  return false;
+  if (!cct->_conf->osd_fast_fail_on_connection_refused)
+    return false;
+
+  OSD::Session *session = (OSD::Session *)con->get_priv();
+  dout(1) << "ms_handle_refused con " << con << " session " << session << dendl;
+  if (!session)
+    return false;
+  int type = con->get_peer_type();
+  // handle only OSD failures here
+  if (monc && (type == CEPH_ENTITY_TYPE_OSD)) {
+    OSDMapRef osdmap = get_osdmap();
+    if (osdmap) {
+      int id = osdmap->identify_osd(con->get_peer_addr());
+      if (osdmap->is_up(id)) {
+	// I'm cheating mon heartbeat grace logic, because we know it's not going
+	// to respawn alone. +1 so we won't hit any boundary case.
+	monc->send_mon_message(new MOSDFailure(monc->get_fsid(),
+						  osdmap->get_inst(id),
+						  cct->_conf->osd_heartbeat_grace + 1,
+						  osdmap->get_epoch(),
+						  MOSDFailure::FLAG_IMMEDIATE | MOSDFailure::FLAG_FAILED
+						  ));
+      }
+    }
+  }
+  session->put();
+  return true;
 }
 
 struct C_OSD_GetVersion : public Context {
--- a/src/osd/OSD.cc
+++ b/src/osd/OSD.cc
@@ -4790,8 +4790,8 @@ bool OSD::ms_handle_refused(Connection *con)
   if (monc && (type == CEPH_ENTITY_TYPE_OSD)) {
     OSDMapRef osdmap = get_osdmap();
     if (osdmap) {
-      int id = osdmap->identify_osd(con->get_peer_addr());
-      if (osdmap->is_up(id)) {
+      int id = osdmap->identify_osd_on_all_channels(con->get_peer_addr());
+      if (id >= 0 && osdmap->is_up(id)) {
 	// I'm cheating mon heartbeat grace logic, because we know it's not going
 	// to respawn alone. +1 so we won't hit any boundary case.
 	monc->send_mon_message(new MOSDFailure(monc->get_fsid(),
@@ -5116,8 +5116,7 @@ void OSD::send_failures()
 
 void OSD::send_still_alive(epoch_t epoch, const entity_inst_t &i)
 {
-  MOSDFailure *m = new MOSDFailure(monc->get_fsid(), i, 0, epoch);
-  m->is_failed = false;
+  MOSDFailure *m = new MOSDFailure(monc->get_fsid(), i, 0, epoch, MOSDFailure::FLAG_ALIVE);
   monc->send_mon_message(m);
 }
 
--- a/src/client/Client.cc
+++ b/src/client/Client.cc
@@ -12380,6 +12380,12 @@ void Client::ms_handle_remote_reset(Connection *con)
   }
 }
 
+bool Client::ms_handle_refused(Connection *con)
+{
+  ldout(cct, 1) << "ms_handle_refused on " << con->get_peer_addr() << dendl;
+  return false;
+}
+
 bool Client::ms_get_authorizer(int dest_type, AuthAuthorizer **authorizer, bool force_new)
 {
   if (dest_type == CEPH_ENTITY_TYPE_MON)
--- a/src/client/Client.h
+++ b/src/client/Client.h
@@ -569,6 +569,7 @@ protected:
   void ms_handle_connect(Connection *con);
   bool ms_handle_reset(Connection *con);
   void ms_handle_remote_reset(Connection *con);
+  bool ms_handle_refused(Connection *con);
   bool ms_get_authorizer(int dest_type, AuthAuthorizer **authorizer, bool force_new);
 
   int authenticate();
--- a/src/librados/RadosClient.cc
+++ b/src/librados/RadosClient.cc
@@ -478,6 +478,10 @@ void librados::RadosClient::ms_handle_remote_reset(Connection *con)
 {
 }
 
+bool librados::RadosClient::ms_handle_refused(Connection *con)
+{
+  return false;
+}
 
 bool librados::RadosClient::_dispatch(Message *m)
 {
--- a/src/librados/RadosClient.h
+++ b/src/librados/RadosClient.h
@@ -58,6 +58,7 @@ private:
   void ms_handle_connect(Connection *con);
   bool ms_handle_reset(Connection *con);
   void ms_handle_remote_reset(Connection *con);
+  bool ms_handle_refused(Connection *con);
 
   Objecter *objecter;
 
--- a/src/mds/Beacon.h
+++ b/src/mds/Beacon.h
@@ -94,6 +94,7 @@ public:
   void ms_handle_connect(Connection *c) {}
   bool ms_handle_reset(Connection *c) {return false;}
   void ms_handle_remote_reset(Connection *c) {}
+  bool ms_handle_refused(Connection *c) {return false;}
 
   void notify_mdsmap(MDSMap const *mdsmap);
   void notify_health(MDSRank const *mds);
--- a/src/mds/MDSDaemon.cc
+++ b/src/mds/MDSDaemon.cc
@@ -1271,6 +1271,12 @@ void MDSDaemon::ms_handle_remote_reset(Connection *con)
   }
 }
 
+bool MDSDaemon::ms_handle_refused(Connection *con)
+{
+  // do nothing for now
+  return false;
+}
+
 bool MDSDaemon::ms_verify_authorizer(Connection *con, int peer_type,
 			       int protocol, bufferlist& authorizer_data, bufferlist& authorizer_reply,
 			       bool& is_valid, CryptoKey& session_key)
--- a/src/mds/MDSDaemon.h
+++ b/src/mds/MDSDaemon.h
@@ -138,6 +138,7 @@ class MDSDaemon : public Dispatcher, public md_config_obs_t {
   void ms_handle_connect(Connection *con);
   bool ms_handle_reset(Connection *con);
   void ms_handle_remote_reset(Connection *con);
+  bool ms_handle_refused(Connection *con);
 
  protected:
   // admin socket handling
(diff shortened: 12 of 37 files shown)


branch-predictor:
This is more of PoC that doesn't take Async messenger into consideration (yet). Comments are welcome.

src/msg/simple/Pipe.cc
```diff
@@ -1336,7 +1336,11 @@ void Pipe::fault(bool onread)
     ldout(msgr->cct,10) << "fault already connecting, reader shutting down" << dendl;
     return;
   }
-  
+  if (errno == ECONNREFUSED) {
+    ldout(msgr->cct, 10) << "connection refused!" << dendl;
+    if (connection_state->clear_pipe(this))
+      msgr->dispatch_queue.queue_refused(connection_state.get());
+  }
```
gregsfortytwo: I haven't done a careful review, but the idea of this patch looks good.

This part isn't really safe though. We can get errors which don't directly call fault() or meander a bit before ending up here. We'll want some kind of shutdown error flag we can set at the failure site and check here or elsewhere.

yuyuyu101:
looks a interesting idea, refuse is a good semantic to decide behavior

src/msg/simple/Pipe.cc
branch-predictor: On Tue, Apr 12, 2016 at 07:35:43AM -0700, Gregory Farnum wrote:
> I haven't done a careful review, but the idea of this patch looks good.
> 
> This part isn't really safe though. We can get errors which don't directly call fault() or meander a bit before ending up here. We'll want some kind of shutdown error flag we can set at the failure site and check here or elsewhere.

I'll look into this, but I'm quite sure that ECONNREFUSED here would only
occur after attempt to connect() to port that isn't listening.

branch-predictor:
Rebased and moved connection error checking to right after connect, that fixes some test failures. Thanks @gregsfortytwo!

src/osd/OSD.cc
```diff
@@ -4679,6 +4679,30 @@ struct C_OSD_GetVersion : public Context {
...
+  if (monc && (type == CEPH_ENTITY_TYPE_OSD)) {
+    OSDMapRef osdmap = get_osdmap();
+    if (osdmap && osdmap->is_up(id)) {
+      // I'm cheating mon heartbeat grace logic, because we know it's not going
+      // to respawn alone. +1 so we won't hit any boundary case.
+      monc->send_mon_message(new MOSDFailure(monc->get_fsid(),
+                                                osdmap->get_inst(id),
+                                                cct->_conf->osd_heartbeat_grace + 1,
+                                                osdmap->get_epoch()
+                                                ));
```
liewegas: Let's update teh message properly.. something like

diff --git a/src/messages/MOSDFailure.h b/src/messages/MOSDFailure.h
index a1032e6..0ff1342 100644
--- a/src/messages/MOSDFailure.h
+++ b/src/messages/MOSDFailure.h
@@ -24,22 +24,34 @@ class MOSDFailure : public PaxosServiceMessage {
   static const int HEAD_VERSION = 3;
 
  public:
+  enum {
+    FLAG_FAILED = 1,     // if set, failure; if not, recovery
+    FLAG_IMMEDIATE = 2,  // known failure, not a timeout
+  };
+  
   uuid_d fsid;
   entity_inst_t target_osd;
-  __u8 is_failed;
+  __u8 flags;
   epoch_t       epoch;
   int32_t failed_for;  // known to be failed since at least this long
 
   MOSDFailure() : PaxosServiceMessage(MSG_OSD_FAILURE, 0, HEAD_VERSION) { }
   MOSDFailure(const uuid_d &fs, const entity_inst_t& f, int duration, epoch_t e)
     : PaxosServiceMessage(MSG_OSD_FAILURE, e, HEAD_VERSION),
-      fsid(fs), target_osd(f), is_failed(true), epoch(e), failed_for(duration) { }
+      fsid(fs), target_osd(f),
+      flags(FLAG_FAILED),
+      epoch(e), failed_for(duration) { }
 private:
   ~MOSDFailure() {}
 
 public: 
   entity_inst_t get_target() { return target_osd; }
-  bool if_osd_failed() { return is_failed; }
+  bool if_osd_failed() {
+    return flags & FLAG_FAILED;
+  }
+  bool is_immediate() {
+    return flags & FLAG_IMMEDIATE;
+  }
   epoch_t get_epoch() { return epoch; }
 
   void decode_payload() {
@@ -49,9 +61,9 @@ public:
     ::decode(target_osd, p);
     ::decode(epoch, p);
     if (header.version >= 2)
-      ::decode(is_failed, p);
+      ::decode(flags, p);
     else
-      is_failed = true;
+      flags = FLAG_FAILED;
     if (header.version >= 3)
       ::decode(failed_for, p);
     else
@@ -62,14 +74,15 @@ public:
     ::encode(fsid, payload);
     ::encode(target_osd, payload);
     ::encode(epoch, payload);
-    ::encode(is_failed, payload);
+    ::encode(flags, payload);
     ::encode(failed_for, payload);
   }
 
   const char *get_type_name() const { return "osd_failure"; }
   void print(ostream& out) const {
     out << "osd_failure("
-       << (is_failed ? "failed " : "recovered ")
+       << (if_osd_failed() ? "failed " : "recovered ")
+       << (is_immediate() ? "immediate " : "timeout ")
        << target_osd << " for " << failed_for << "sec e" << epoch
        << " v" << version << ")";
   }

then we can use this for other stuff too, like a watchdog.

liewegas:
Looking at this again, I think this approach makes a lot of sense, and should be safe, too.  Worst case, we mark an osd down quickly instead of slowly, which is not much of a problem.

Can you break this apart into separate patches?  e.g.,

 - add the new msgr callback
 - implement it in msg/simple
 - update failure message
 - update mon to use immediate flag
 - update osd to implement hook

src/osd/OSD.cc
branch-predictor: On Fri, Apr 22, 2016 at 09:28:13AM -0700, Sage Weil wrote:
> Let's update teh message properly.. something like
> 
> [..]
>   public:
> +  enum {
> +    FLAG_FAILED = 1,     // if set, failure; if not, recovery
> +    FLAG_IMMEDIATE = 2,  // known failure, not a timeout
> +  };
> +  
> [..] 
> then we can use this for other stuff too, like a watchdog.

And introduce other daemon failure modes, too.

-- 
Piotr Dałek

branch-predictor:
@liewegas I have broken it into a bunch of separate commits, added new failure flag to MOSDFailure class and new option to disable new behaivior.

yuyuyu101:
back from ML, we can continue review this?

src/msg/simple/Pipe.cc
```diff
@@ -914,9 +914,13 @@ int Pipe::connect()
   ldout(msgr->cct,10) << "connecting to " << peer_addr << dendl;
   rc = ::connect(sd, (sockaddr*)&peer_addr.addr, peer_addr.addr_size());
   if (rc < 0) {
-    rc = -errno;
+    int stored_errno = errno;
     ldout(msgr->cct,2) << "connect error " << peer_addr
-	     << ", " << cpp_strerror(rc) << dendl;
+	     << ", " << cpp_strerror(stored_errno) << dendl;
+    if (stored_errno == ECONNREFUSED) {
+      ldout(msgr->cct, 10) << "connection refused!" << dendl;
```
yuyuyu101: I'd like to set level to 0

liewegas: I would go with 1 or 2

src/common/config_opts.h
```diff
@@ -188,6 +188,7 @@ OPTION(ms_rwthread_stack_bytes, OPT_U64, 1024 << 10)
 OPTION(ms_tcp_read_timeout, OPT_U64, 900)
 OPTION(ms_pq_max_tokens_per_priority, OPT_U64, 16777216)
 OPTION(ms_pq_min_cost, OPT_U64, 65536)
+OPTION(ms_immediate_fail_on_connection_refused, OPT_BOOL, true) // mark osd down immediately after receiving ECONNREFUSED during (re)connection
```
liewegas: I think this should be an osd option, osd_fast_fail_on_connection_refused, and the conditional should go in the OSD's ms_refused callback

liewegas:
I like this better!

1. Please make the second patch only add the callbacks, but leave them empty.. don't do anytihng fancy in OSD yet.
2. then update the failure message
3. then in the final patch add the osd config option that will send a fast fail message from OSD::ms_refused.

yuyuyu101:
if you could add async msgr support, it would be good since it should be only two line code

branch-predictor:
@yuyuyu101 Done, please take a look.
@liewegas rebased and split again.

yuyuyu101:
@branch-predictor I think we set peer osd id in Session instead of in Connection, it would be make more sense.

yuyuyu101:
Another question, do we really need immediate flag? If connection refused, we only send a normal MOSDFailure message. If osd down, just like heartbeat, multi osds will send MOSDFailure in handle_refuse function. So it won't occur too much latency?

Because I concern the immediate flag will cause osd flipping in unknown env..... Like message delay?

branch-predictor:
On Wed, May 25, 2016 at 03:09:05AM -0700, Haomai Wang wrote:
> Another question, do we really need immediate flag? If connection refused, we only send a normal MOSDFailure message. If osd down, just like heartbeat, multi osds will send MOSDFailure in handle_refuse function. So it won't occur too much latency?
> 
> Because I concern the immediate flag will cause osd flipping in unknown env..... Like message delay? 

We set FLAG_IMMEDIATE only in ms_handle_refused as a signal that we're
certain that OSD won't get up soon. In other cases (regular hertbeat
timeout) we don't set it and it's up to mons to decide whether they should
be marked as down or not. As for flapping, we discussed it on ceph-devel ml
and came to conclusion that it requires either broken firewall or network
configuration to cause this, and these are more serious issues that should
be resolved first before worrying about OSDs flapping (either way, flapping
OSDs could be good for getting someone's attention).

-- 
Piotr Dałek
branch@predictor.org.pl
http://blog.predictor.org.pl

yuyuyu101:
ok.

On Wed, May 25, 2016 at 6:20 PM, Piotr Dałek <notifications@github.com>
wrote:

> On Wed, May 25, 2016 at 03:09:05AM -0700, Haomai Wang wrote:
> > Another question, do we really need immediate flag? If connection
> refused, we only send a normal MOSDFailure message. If osd down, just like
> heartbeat, multi osds will send MOSDFailure in handle_refuse function. So
> it won't occur too much latency?
> >
> > Because I concern the immediate flag will cause osd flipping in unknown
> env..... Like message delay?
>
> We set FLAG_IMMEDIATE only in ms_handle_refused as a signal that we're
> certain that OSD won't get up soon. In other cases (regular hertbeat
> timeout) we don't set it and it's up to mons to decide whether they should
> be marked as down or not. As for flapping, we discussed it on ceph-devel ml
> and came to conclusion that it requires either broken firewall or network
> configuration to cause this, and these are more serious issues that should
> be resolved first before worrying about OSDs flapping (either way, flapping
> OSDs could be good for getting someone's attention).
>
> --
> Piotr Dałek
> branch@predictor.org.pl
> http://blog.predictor.org.pl
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ceph/ceph/pull/8558#issuecomment-221532745>
>



-- 

Best Regards,

Wheat

liewegas:
I think it will be okay (no flapping) in the ECONNREFUSED case.  At 
least, the only case I can think of where it would flap would be if there 
was some weird network firewall rule that sent RST instead of dropping 
packets, and did so intermittently.

Aside from that, though, I can see other cases where we *know* the process 
is dead (e.g., a watchdog process) and the immediate flag will be 
useful/needed.

yuyuyu101:
I guess we need to have qa test to cover this? For example, enable refuse option and stop one osd, then make sure mon receives OSDFauilre within 0.5s.

branch-predictor:
On Sat, Jun 04, 2016 at 07:11:09AM -0700, Haomai Wang wrote:
> I guess we need to have qa test to cover this? For example, enable refuse option and stop one osd, then make sure mon receives OSDFauilre within 0.5s.

Good idea, I'll add one.

-- 
Piotr Dałek
branch@predictor.org.pl
http://blog.predictor.org.pl

yuyuyu101:
ping @branch-predictor

branch-predictor:
Sorry it took a while, been busy lately.
Rebased and added test to both CMake and autotools.

src/messages/MOSDFailure.h
```diff
@@ -24,22 +24,41 @@ class MOSDFailure : public PaxosServiceMessage {
   static const int HEAD_VERSION = 3;
 
  public:
+  enum {
+    FLAG_ALIVE = 0,      // use this on its own to mark as "I'm still alive"
```
liewegas: I would leave this out..

src/messages/MOSDFailure.h
```diff
@@ -24,22 +24,41 @@ class MOSDFailure : public PaxosServiceMessage {
...
     : PaxosServiceMessage(MSG_OSD_FAILURE, e, HEAD_VERSION),
-      fsid(fs), target_osd(f), is_failed(true), epoch(e), failed_for(duration) { }
+      fsid(fs), target_osd(f),
+      flags(FLAG_FAILED),
+      epoch(e), failed_for(duration) { }
+  MOSDFailure(const uuid_d &fs, const entity_inst_t& f, int duration, 
+              epoch_t e, __u8 extra_flags)
+    : PaxosServiceMessage(MSG_OSD_FAILURE, e, HEAD_VERSION),
+      fsid(fs), target_osd(f),
+      flags(FLAG_FAILED | extra_flags),
```
liewegas: and just pass full flags (not |'ing in FAILED)

src/osd/OSD.cc
```diff
@@ -5093,7 +5120,7 @@ void OSD::send_failures()
 void OSD::send_still_alive(epoch_t epoch, const entity_inst_t &i)
 {
   MOSDFailure *m = new MOSDFailure(monc->get_fsid(), i, 0, epoch);
-  m->is_failed = false;
+  m->flags = MOSDFailure::FLAG_ALIVE;
```
liewegas: and then just pass in 0 (or maybe define FLAG_NONE?) to the ctor

src/osd/OSD.cc
```diff
@@ -4783,6 +4783,33 @@ bool OSD::ms_handle_reset(Connection *con)
...
+  // handle only OSD failures here
+  if (monc && (type == CEPH_ENTITY_TYPE_OSD)) {
+    OSDMapRef osdmap = get_osdmap();
+    if (osdmap && osdmap->is_up(id)) {
+      // I'm cheating mon heartbeat grace logic, because we know it's not going
+      // to respawn alone. +1 so we won't hit any boundary case.
+      monc->send_mon_message(new MOSDFailure(monc->get_fsid(),
+                                                osdmap->get_inst(id),
+                                                cct->_conf->osd_heartbeat_grace + 1,
+                                                osdmap->get_epoch(), MOSDFailure::FLAG_IMMEDIATE
```
liewegas: MOSDFAilure::FLAG_FAILURE | MOSDFailure::FLAG_IMMEDIATE

branch-predictor:
@liewegas done.

src/msg/async/AsyncConnection.cc
```diff
@@ -1019,6 +1032,10 @@ ssize_t AsyncConnection::_process_connection()
         r = net.reconnect(get_peer_addr(), sd);
         if (r < 0) {
           ldout(async_msgr->cct, 1) << __func__ << " reconnect failed " << dendl;
+          if (r == -ECONNREFUSED) {
+            ldout(async_msgr->cct, 2) << __func__ << " connection refused!" << dendl;
+            center->dispatch_event_external(connection_refused_handler);
```
yuyuyu101: async msgr now also use dispatch_queue-> queue_refused !

branch-predictor: On Thu, Jul 14, 2016 at 10:19:57AM -0700, Haomai Wang wrote:
> > @@ -1019,6 +1032,10 @@ ssize_t AsyncConnection::_process_connection()
> >          r = net.reconnect(get_peer_addr(), sd);
> >          if (r < 0) {
> >            ldout(async_msgr->cct, 1) << __func__ << " reconnect failed " << dendl;
> > +          if (r == -ECONNREFUSED) {
> > +            ldout(async_msgr->cct, 2) << __func__ << " connection refused!" << dendl;
> > +            center->dispatch_event_external(connection_refused_handler);
> 
> async msgr now also use dispatch_queue-> queue_refused !

Updated and rebased.

-- 
Piotr Dałek
branch@predictor.org.pl
http://blog.predictor.org.pl

src/msg/async/AsyncConnection.cc
```diff
@@ -135,6 +146,8 @@ AsyncConnection::AsyncConnection(CephContext *cct, AsyncMessenger *m, DispatchQu
 {
   read_handler = new C_handle_read(this);
   write_handler = new C_handle_write(this);
+
+  connection_refused_handler = new C_handle_connection_refused(async_msgr, this);
```
yuyuyu101: need to delete..

src/msg/async/AsyncConnection.cc
```diff
@@ -87,6 +87,17 @@ class C_clean_handler : public EventCallback {
...
+class C_handle_connection_refused : public EventCallback {
+  AsyncMessenger *msgr;
+  AsyncConnectionRef conn;
+
+ public:
+  C_handle_connection_refused(AsyncMessenger *m, AsyncConnectionRef c): msgr(m), conn(c) {}
+  void do_request(int id) {
+    msgr->ms_deliver_handle_refused(conn.get());
+  }
+};
```
yuyuyu101: need to delete..

src/msg/async/AsyncConnection.h
```diff
@@ -331,6 +331,8 @@ class AsyncConnection : public Connection {
   utime_t backoff;         // backoff time
   EventCallbackRef read_handler;
   EventCallbackRef write_handler;
+
+  EventCallbackRef connection_refused_handler;
```
yuyuyu101: need to

src/msg/async/AsyncConnection.h
```diff
@@ -406,6 +408,8 @@ class AsyncConnection : public Connection {
     shutdown_socket();
     delete read_handler;
     delete write_handler;
+
+    delete connection_refused_handler;
```
yuyuyu101: and this

branch-predictor:
@yuyuyu101 deleted.

src/msg/async/AsyncConnection.h
```diff
@@ -406,6 +407,7 @@ class AsyncConnection : public Connection {
     shutdown_socket();
     delete read_handler;
     delete write_handler;
+
```
yuyuyu101: blank line

src/msg/async/AsyncConnection.h
```diff
@@ -331,6 +331,7 @@ class AsyncConnection : public Connection {
   utime_t backoff;         // backoff time
   EventCallbackRef read_handler;
   EventCallbackRef write_handler;
+
```
yuyuyu101: black

src/msg/async/AsyncConnection.cc
```diff
@@ -135,6 +135,7 @@ AsyncConnection::AsyncConnection(CephContext *cct, AsyncMessenger *m, DispatchQu
 {
   read_handler = new C_handle_read(this);
   write_handler = new C_handle_write(this);
+
```
yuyuyu101: blank

src/msg/simple/SimpleMessenger.cc
```diff
@@ -357,6 +357,7 @@ Pipe *SimpleMessenger::connect_rank(const entity_addr_t& addr,
 			static_cast<PipeConnection*>(con));
   pipe->pipe_lock.Lock();
   pipe->set_peer_type(type);
+  pipe->set_peer_id(num);
```
yuyuyu101: I found we only set peer id when connecting, what if the connection is accepted to build?

src/osd/OSD.cc
```diff
@@ -4779,6 +4779,34 @@ bool OSD::ms_handle_reset(Connection *con)
...
+      // I'm cheating mon heartbeat grace logic, because we know it's not going
+      // to respawn alone. +1 so we won't hit any boundary case.
+      monc->send_mon_message(new MOSDFailure(monc->get_fsid(),
+                                                osdmap->get_inst(id),
+                                                cct->_conf->osd_heartbeat_grace + 1,
+                                                osdmap->get_epoch(),
+                                                MOSDFailure::FLAG_IMMEDIATE | MOSDFailure::FLAG_FAILED
+                                                ));
+    }
+  }
```
yuyuyu101: session leak, session->put()

branch-predictor: Fixed

yuyuyu101:
I felt a little nervous to pass osd id via caller code. what if use OSDMap::identify_osd(const entity_addr_t& addr) to get osd id in ms_handle_refused()? @liewegas

yuyuyu101:
hmm, I think we don't have a way let server side connection get peer id in messenger level. so I think identitfy_osd via entity_addr_t is a reasonable way

liewegas:
@yuyuyu101 Yeah, I was a bit nervous about that too. That means we would drop 6841eff.

branch-predictor:
identify_osd() alone is not enough as it doesn't take hb addrs into account. I just added identify_osd_on_all_channels() to work around it, not influencing any existing users.

yuriw:
@branch-predictor I am trying to rebuild with this PR and got conflicts:

[yuriw@smithi018 build]$ git pull https://github.com/branch-predictor/ceph.git bp-mark-down-on-perm-rst
From https://github.com/branch-predictor/ceph
 * branch            bp-mark-down-on-perm-rst -> FETCH_HEAD
Auto-merging src/test/msgr/test_msgr.cc
CONFLICT (modify/delete): src/test/Makefile.am deleted in HEAD and modified in c8a9ce3d2fd7845b46daa1b9a3b9c0d1503c6184. Version c8a9ce3d2fd7845b46daa1b9a3b9c0d1503c6184 of src/test/Makefile.am left in tree.
Auto-merging src/osdc/Objecter.cc
Auto-merging src/osd/OSD.cc
Auto-merging src/msg/DispatchQueue.h
Auto-merging src/common/config_opts.h
Auto-merging src/client/Client.h
Auto-merging src/client/Client.cc
Automatic merge failed; fix conflicts and then commit the result.

pls fix and assign "needs-qa" tag

branch-predictor:
@yuriw This is because of https://github.com/ceph/ceph/pull/11007 being merged. Rebased - should be OK now (only the Makefile.am part of commit was removed).
Unfortunately I don't have tagging rights so I can't put needs-qa tag.

yuriw:
@branch-predictor thx
ack @liewegas

yuriw:
http://pulpito.front.sepia.ceph.com:80/yuriw-2016-09-14_16:05:47-rados-wip-yuri-testing_2016_09_14-distro-basic-smithi/

http://pulpito.ceph.com/yuriw-2016-09-15_15:38:24-rados-wip-yuri-testing_2016_09_14-distro-basic-smithi/

nh2:
(Edit: I filed this as a new Ceph bug https://tracker.ceph.com/issues/78385 -- just commenting here so that subscribers on this issue can see it conveniently and perhaps chime in.)

> At
> least, the only case I can think of where it would flap would be if there
> was some weird network firewall rule that sent RST instead of dropping
> packets, and did so intermittently.

Hello, I am from the future 10 years later, and of course this happened to me.

`osd_fast_fail_on_connection_refused` can cause that a single mis-configured machine (or network equipment on the way to it) takes down the entire cluster.

In my case, we block Ceph ports `6800-*` by default with `iptables`, permitting only known IPs in order to reduce the attack surface of there being a 0-day vulnerability in the Ceph protocol. Port 3300 does not need `iptables` blocking because we run that over our VPN for protection; we cannot run `6800-*` (main OSD traffic) on the VPN because it's too slow / unstable because UDP-based VPNs tend to drop too many packets indiscriminately.

We use `REJECT` instead of `DROP` in `iptables` because that's makes for easier debugging, compared to seeing hanging connections.

Now, we tried to add a new OSD machine to our cluster, and did so _before_ allow-listing it in `iptables`. The effect:

* The new machine made itself to the `mon`s via the unblocked port 3300.
* The new machine's OSDs got `Connection refused` when talking to any other OSDs due to the `iptables` `REJECT`.
* The new machine thus told the MON to shut down all other OSDs. Because of `osd_fast_fail_on_connection_refused`'s  `FLAG_IMMEDIATE`, the MON acted on that without any quorum/voting/`mon_osd_min_down_reporters`/`mon_osd_reporter_subtree_level`.
* OSDs all around the cluster started restarting/flapping.

This way, a single problematic machine took down the entire Ceph cluster.

It can be reproduced with, running on 1 machine:

```sh
iptables --wait --insert OUTPUT --protocol tcp --match multiport --destination-ports 6800:7568 --jump REJECT --reject-with tcp-reset

# to undo:
iptables --wait --delete OUTPUT --protocol tcp --match multiport --destination-ports 6800:7568 --jump REJECT --reject-with tcp-reset
```

On one hand, Ceph assumes bug-free software (e.g. does not implement Byzantine fault tolerance) and only protects against network downtime, not really misconfigurations.
On the other hand, it's a bit impractical that a single misconfigured node (or network equipment on the way to it) can cause the entire cluster to go down.

Just setting `osd_fast_fail_on_connection_refused = false` is not a good solution, as that really makes detection of working OSDs extremely slow, or not happen at all: When it's `false`, after using my `iptables` invocation above, Ceph takes 30 seconds until `SLOW_OPS` shows up in `ceph -w`, and all OSDs (including the ones of the one that just cut all its connections) stay marked as `up` for apparently forever (I waited 70 minutes so far), and the cluster is hanging. So in this case, the OSDs that _should_ be ejected from the cluster are not.

This leaves us in this situation:

* `osd_fast_fail_on_connection_refused = true` ejects unproblematic OSDs from the cluster
* `osd_fast_fail_on_connection_refused = false` doesn't eject problematic OSDs from the cluster

For a fix, I am wondering if there is a way to implement `osd_fast_fail_on_connection_refused` still as "pretty immediate", but without listening to the vote of just a single machine.
* For example, shouldn't the other machines talking to truly down OSDs (e.g. crashed processes) ALL get `Connection refused` pretty much instantly?

PR #8559, opened 2016-04-12
rgw: fix unsafe c_str() usage

liewegas:


src/common/buffer.cc
```diff
@@ -1780,6 +1780,18 @@ static simple_spinlock_t buffer_debug_lock = SIMPLE_SPINLOCK_INITIALIZER;
     return _buffers.front().c_str();  // good, we're already contiguous.
   }
 
+  string buffer::list::to_str() const {
+    string s;
+    for (std::list<ptr>::const_iterator p = _buffers.begin();
+	 p != _buffers.end();
+	 ++p) {
+      if (p->length()) {
+	s += string(p->c_str(), p->length());
```
cbodley: consider using `s.append(p->c_str(), p->length())` to avoid allocating and copying from a temporary string. if you know the total length, you can also use `s.reserve(len)` at the beginning to avoid extra allocations on append

liewegas:
@cbodley thanks, updated!

liewegas:
http://pulpito.ceph.com/sage-2016-04-12_08:47:13-rgw-wip-15643---basic-mira/

PR #8560, opened 2016-04-12
osd: refuse to start if object name limits are not possible

liewegas:
Unfortunately we don't mount() until after the fork.  It goes in the  log, though, and systemd should still capture it since derr echos to stderr too.

This does change the workflow for devs on ext4 because they have to pass --short.  This seems better than having librados and rgw tests unexpectedly fail.

### DIFF
--- a/src/osd/OSD.cc
+++ b/src/osd/OSD.cc
@@ -2037,7 +2037,13 @@ int OSD::init()
       derr << "   osd max object namespace len = "
 	   << g_conf->osd_max_object_namespace_len << dendl;
       derr << cpp_strerror(r) << dendl;
-      goto out;
+      if (g_conf->osd_check_max_object_name_len_on_startup) {
+	goto out;
+      }
+      derr << "osd_check_max_object_name_len_on_startup = false, starting anyway"
+	   << dendl;
+    } else {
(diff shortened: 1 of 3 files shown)


src/vstart.sh
```diff
@@ -162,6 +164,9 @@ case $1 in
     --new | -n )
 	    new=1
 	    ;;
+    --long )
```
dachary: Did you mean 

<pre>
--short)
   long=0 
</pre>

src/osd/OSD.cc
```diff
@@ -2022,6 +2022,25 @@ int OSD::init()
   int rotating_auth_attempts = 0;
   const int max_rotating_auth_attempts = 10;
 
+  // sanity check long object name handling
+  {
+    hobject_t l;
+    l.oid.name = string(g_conf->osd_max_object_name_len, 'n');
+    l.set_key(string(g_conf->osd_max_object_name_len, 'k'));
+    l.nspace = string(g_conf->osd_max_object_namespace_len, 's');
+    r = store->validate_hobject_key(l);
```
athanatos: This uses the config values for max attr size to validate the resulting key.  Is that the intent?

dachary:
<pre>
</pre>

PR #8561, opened 2016-04-12
rgw-admin: sync zones as part of realm pull

kmroz:
Up until now, a realm pull from a master to a peer, would sync realm and period information. Within the period, one would find the period map containing zone groups. Zone groups would also be synced to the peer. Syncing zones attached to each zone group, however, would make life a bit easier for the administrator. This PR tries to address this.

@oritwas - Orit, as we discussed on IRC. Please have a look when you have a moment. Thanks!

### DIFF
--- a/src/rgw/rgw_admin.cc
+++ b/src/rgw/rgw_admin.cc
@@ -2791,6 +2791,34 @@ int main(int argc, char **argv)
 	  }
 	}
 
+	// Perform zone sync as well.
+	// TODO: Should we only do this if above realm.udpate() was successful?
+	const RGWPeriodMap& p_map = period.get_map();
+	for (map<string, RGWZoneGroup>::const_iterator zg_it = p_map.zonegroups.begin(); zg_it != p_map.zonegroups.end(); ++zg_it) {
+	  const RGWZoneGroup zg = zg_it->second;
+	  ldout(store->ctx(), 20) << "Creating missing zones for zonegroup: " << zg.get_name() << ":" << zg.get_id() << dendl;
+	  for (map<string, RGWZone>::const_iterator z_it = zg.zones.begin(); z_it != zg.zones.end(); ++z_it) {
+	    const RGWZone z = z_it->second;
+	    ldout(store->ctx(), 20) << "Creating zone: " << z.name << ":" << z.id << dendl;
+	    // Let's try to create the zone.
+	    RGWZoneParams zp(z.id, z.name, zg.realm_id);
+	    ret = zp.init(g_ceph_context, store, false);
+	    if (ret < 0 && ret != -EEXIST) {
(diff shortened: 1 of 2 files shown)


src/rgw/rgw_rados.cc
```diff
@@ -1496,7 +1496,7 @@ int RGWZoneParams::create(bool exclusive)
   }
 
   r = RGWSystemMetaObj::create(exclusive);
-  if (r < 0) {
+  if (r < 0 && r != -EEXIST) {
```
oritwas: this will cause error when create_default races with other create_default, it is up to the caller to handle EEXIST errors

kmroz: Ahh...ok. Will fix it up tomorrow. Thanks.

cbodley:
I think there's a conceptual disconnect here that we've had trouble articulating.

The `RGWZoneParams` object contains local information, like the pool names, that aren't relevant to the other clusters/zones. It only needs to exist on the cluster where it runs.

`RGWZone` is the external representation that contains information that the other clusters need to communicate with that zone - things like endpoints and how to do sync. `RGWZone` is not a rados object, and only lives within its `RGWZoneGroup`.

So the issue is that there isn't a way for the period (and the `RGWZoneGroup`s that it contains) to carry all of the information that another cluster would need to reconstruct the local `RGWZoneParams` object.

As I recall from @yehudasa on the matter, this separation was meant to allow a zone to move from one zonegroup to another, or to potentially be part of multiple zonegroups at once - for example, a master of one zonegroup, and a read-only replica in another zonegroup. This isn't currently supported, but was part of the original design.

I agree that it would be more convenient for an admin to do all of the configuration in one place. But because the local zone information is not part of the period (which is our mechanism for maintaining a consistent view of the multisite configuration across clusters), we can't really provide that for zones.

src/rgw/rgw_admin.cc
```diff
@@ -2791,6 +2791,34 @@ int main(int argc, char **argv)
...
+	    RGWZoneParams zp(z.id, z.name, zg.realm_id);
+	    ret = zp.init(g_ceph_context, store, false);
+	    if (ret < 0 && ret != -EEXIST) {
+	      cerr << "unable to initialize zone: " << cpp_strerror(-ret) << std::endl;
+	      return -ret;
+	    }
+	    // Will attach access keys to zone.
+	    zp.system_key.id = access_key;
+	    zp.system_key.key = secret_key;
+	    zp.realm_id = zg.realm_id;
```
cbodley: unfortunately, this is missing most of the information (pool names, especially) that the gateway would need to use this zone locally - see my explanation for more

kmroz: I see what you mean. I neglected to factor in differing pool names, etc :/ Thanks for reviewing.

I will close the PR. However, @cbodley, this design you speak of, is it documented somewhere?

PR #8562, opened 2016-04-12
rbd: add rbd object-map check

fullerdj:
Add a CLI option to verify the object map for an unmapped image
for debugging purposes. Syntax:

rbd object-map check <image-name>

This operation is not supported on currently mapped images.

Any inconsistencies that could affect the correctness of future
operations are noted and the object map (and fast diff state, if
enabled) is invalidated. Inconsistencies that do not affect
correctness will be emitted at debug level 1 or higher.

Consolidate code path with object-map rebuild and remove newly
unused code.

Fixes: http://tracker.ceph.com/issues/14867

### DIFF
--- a/src/include/rbd/librbd.hpp
+++ b/src/include/rbd/librbd.hpp
@@ -204,6 +204,8 @@ public:
   /* object map feature */
   int rebuild_object_map(ProgressContext &prog_ctx);
 
+  int check_object_map(ProgressContext &prog_ctx);
+
   int copy(IoCtx& dest_io_ctx, const char *destname);
   int copy2(Image& dest);
   int copy3(IoCtx& dest_io_ctx, const char *destname, ImageOptions& opts);
--- a/src/librbd/Operations.cc
+++ b/src/librbd/Operations.cc
@@ -13,6 +13,7 @@
 #include "librbd/Utils.h"
 #include "librbd/operation/FlattenRequest.h"
 #include "librbd/operation/RebuildObjectMapRequest.h"
+#include "librbd/operation/ObjectMapIterate.h"
 #include "librbd/operation/RenameRequest.h"
 #include "librbd/operation/ResizeRequest.h"
 #include "librbd/operation/SnapshotCreateRequest.h"
@@ -274,6 +275,18 @@ struct C_InvokeAsyncRequest : public Context {
   }
 };
 
+template <typename I>
+bool needs_invalidate(I& image_ctx, uint64_t object_no,
+		     uint8_t current_state, uint8_t new_state) {
+  if ( (current_state == OBJECT_EXISTS ||
+	current_state == OBJECT_EXISTS_CLEAN) &&
+       (new_state == OBJECT_NONEXISTENT ||
+	new_state == OBJECT_PENDING)) {
+    return false;
+  }
+  return true;
+}
+
 } // anonymous namespace
 
 template <typename I>
@@ -421,6 +434,48 @@ void Operations<I>::execute_rebuild_object_map(ProgressContext &prog_ctx,
     new operation::RebuildObjectMapRequest<I>(
       m_image_ctx, new C_NotifyUpdate<I>(m_image_ctx, on_finish), prog_ctx);
   req->send();
+}
+
+template <typename I>
+int Operations<I>::check_object_map(ProgressContext &prog_ctx) {
+  CephContext *cct = m_image_ctx.cct;
+  ldout(cct, 5) << this << " " << __func__ << dendl;
+  int r = m_image_ctx.state->refresh_if_required();
+  if (r < 0) {
+    return r;
+  }
+
+  r = invoke_async_request("check object map", true,
+                           boost::bind(&Operations<I>::check_object_map, this,
+                                       boost::ref(prog_ctx), _1),
+			   [] (Context *c) { c->complete(-EOPNOTSUPP); });
+
+  return r;
+}
+
+template <typename I>
+void Operations<I>::object_map_iterate(ProgressContext &prog_ctx,
+				       operation::ObjectIterateWork<I> handle_mismatch,
+				       Context *on_finish) {
+  assert(m_image_ctx.owner_lock.is_locked());
+  assert(m_image_ctx.exclusive_lock == nullptr ||
+         m_image_ctx.exclusive_lock->is_lock_owner());
+
+  if (!m_image_ctx.test_features(RBD_FEATURE_OBJECT_MAP)) {
+    on_finish->complete(-EINVAL);
+    return;
+  }
+
+  operation::ObjectMapIterateRequest<I> *req =
+    new operation::ObjectMapIterateRequest<I>(m_image_ctx, on_finish,
+					      prog_ctx, handle_mismatch);
+  req->send();
+}
+
+template <typename I>
+void Operations<I>::check_object_map(ProgressContext &prog_ctx,
+				     Context *on_finish) {
+  object_map_iterate(prog_ctx, needs_invalidate, on_finish);
 }
 
 template <typename I>
--- a/src/librbd/operation/ObjectMapIterate.cc
+++ b/src/librbd/operation/ObjectMapIterate.cc
@@ -1,0 +1,293 @@
+// -*- mode:C++; tab-width:8; c-basic-offset:2; indent-tabs-mode:t -*-
+// vim: ts=8 sw=2 smarttab
+
+#include "librbd/operation/ObjectMapIterate.h"
+#include "common/dout.h"
+#include "common/errno.h"
+#include "librbd/AsyncObjectThrottle.h"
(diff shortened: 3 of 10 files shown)


src/tools/rbd/action/ObjectMapCheck.cc
```diff
@@ -0,0 +1,74 @@
+// -*- mode:C++; tab-width:8; c-basic-offset:2; indent-tabs-mode:t -*-
```
dillaman: Can this just be combined with the current "ObjectMap.cc" file to keep all object map operations together?

src/include/rbd/librbd.hpp
```diff
@@ -188,6 +188,8 @@ class CEPH_RBD_API Image
   /* object map feature */
   int rebuild_object_map(ProgressContext &prog_ctx);
 
+  int check_object_map(ProgressContext &prog_ctx);
```
dillaman: Can you add it to the C and Python APIs as well?

src/librbd/Operations.cc
```diff
@@ -358,6 +359,60 @@ void Operations<I>::execute_rebuild_object_map(ProgressContext &prog_ctx,
...
+  ldout(cct, 5) << this << " " << __func__ << dendl;
+  int r = m_image_ctx.state->refresh_if_required();
+  if (r < 0) {
+    return r;
+  }
+
+  r = invoke_async_request("check object map", true,
+                           boost::bind(&Operations<I>::check_object_map, this,
+                                       boost::ref(prog_ctx), _1),
+			   [] (Context *c) { c->complete(-EOPNOTSUPP); });
```
dillaman: It would be good to add the necessary RPC plumbing to support this when the exclusive lock cannot be acquired -- otherwise you'd never be able to verify the object map on opened images.

fullerdj: We discussed this during standup at one point; since this is more of a debugging tool than a useful operation in itself, it was thought that users would want to use rebuild in that case. Also, there is not necessarily the expectation of consistency during active I/O, so it's likely there would be false positives.

Also, I wanted to be the first to commit a C++11 lambda.

src/librbd/Operations.cc
```diff
@@ -358,6 +359,60 @@ void Operations<I>::execute_rebuild_object_map(ProgressContext &prog_ctx,
...
+  }
+
+  operation::ObjectMapIterateRequest<I> *req =
+    new operation::ObjectMapIterateRequest<I>(m_image_ctx, on_finish,
+					      prog_ctx, handle_mismatch);
+  req->send();
+}
+
+template <typename I>
+int needs_invalidate(I& image_ctx, uint64_t object_no,
```
dillaman: Can we swap this to return a bool instead of an int?  Also, can we move this up to the anonymous namespace section at the top of the file.

src/librbd/operation/ObjectMapIterate.cc
```diff
@@ -0,0 +1,292 @@
...
+
+    ldout(cct, 10) << "C_VerifyObjectCallback::object_map_action"
+		   << " object " << image_ctx.get_object_name(m_object_no)
+		   << " state " << (int)state
+		   << " new_state " << (int)new_state << dendl;
+
+    if (state != new_state) {
+      int r = 0;
+
+      if (m_handle_mismatch)
```
dillaman: Do we need this guard?  Is it valid to be null?

src/librbd/operation/ObjectMapIterate.cc
```diff
@@ -0,0 +1,292 @@
...
+	r = m_handle_mismatch(image_ctx, m_object_no, state, new_state);
+      if (r) {
+	lderr(cct) << "object map error: object "
+		   << image_ctx.get_object_name(m_object_no)
+		   << " marked as " << (int)state << ", but should be "
+		   << (int)new_state << dendl;
+	m_invalidate->test_and_set();
```
dillaman: If we detect that the object map is invalid, do we need to continue checking the rest of the object map?  You could eliminate the atomic and just pass back an error code to the object throttle -- which will finish all in-flight ops and bubble the error back up to the main state machine.

src/librbd/operation/ObjectMapIterate.cc
fullerdj: No, that would be a programming error.

Do we generally handle cases like this with assert() or just ignore them?

src/tools/rbd/action/ObjectMapCheck.cc
```diff
@@ -0,0 +1,74 @@
...
+namespace action {
+namespace object_map_check {
+
+namespace at = argument_types;
+namespace po = boost::program_options;
+
+static int do_object_map_check(librbd::Image &image, bool no_progress)
+{
+  utils::ProgressContext pc("Object Map Check", no_progress);
+  int r = image.check_object_map(pc);
```
dillaman: This should return 0 even if the object map was invalidated?  Maybe output a message saying that the object map was invalidated.

src/tools/rbd/action/ObjectMapCheck.cc
fullerdj: Well, right now there is a 1:1 correspondence between tool actions and the source files that implement them (cf. ObjectMapUpdateRequest.cc). ObjectMap.cc contains the library tooling and this seemed like a user-level operation like the others in operations/. I'll change it if you want, though.

src/librbd/operation/ObjectMapIterate.cc
fullerdj: From a purely functional perspective, that wouldn't bother me. This is mainly a debugging tool, so I would like to see all problems with the object map, not just the first one. If you're in production, you should just use rebuild.

src/tools/rbd/action/ObjectMapCheck.cc
fullerdj: The object map invalidate request prints a message.

src/librbd/Operations.cc
dillaman: Fair enough -- we can always add it later.

src/librbd/operation/ObjectMapIterate.cc
dillaman: assert() is fine -- it shows the expectation that it shouldn't be null.

src/include/rbd/librbd.hpp
dillaman: Can ignore since it's just for debugging.

src/librbd/operation/ObjectMapIterate.cc
dillaman: Fair enough

src/tools/rbd/action/ObjectMapCheck.cc
dillaman: This is for the cli -- most of the related commands are grouped into the same file.

dillaman: I'd still like to see all the object map commands handles within "src/tools/rbd/action/ObjectMap.cc"

fullerdj: I misconstrued this note to refer to librbd/operation instead of tools/rbd/action. Sorry about that; I'll merge the files.

dillaman:
@fullerdj unittest_librbd is failing on the TestLibRBD.CheckObjectMap test case.

fullerdj:
Where/how do you see that? It passes on my test setup, maybe I’m missing something.

dillaman:
@fullerdj I just re-merged and re-built it and still see the failure:

> $ RBD_FEATURES=61 ./unittest_librbd --gtest_filter=TestLibRBD.CheckObjectMap
> Note: Google Test filter = TestLibRBD.CheckObjectMap
> [==========] Running 1 test from 1 test case.
> [----------] Global test environment set-up.
> [----------] 1 test from TestLibRBD
> seed 16854
> [ RUN      ] TestLibRBD.CheckObjectMap
> using new format!
> 0.00% done
> 25.00% done
> 50.00% done
> 75.00% done
> test/librbd/test_librbd.cc:3727: Failure
> Value of: (flags & RBD_FLAG_OBJECT_MAP_INVALID) != 0
>   Actual: false
> Expected: true
> [  FAILED  ] TestLibRBD.CheckObjectMap (35 ms)
> [----------] 1 test from TestLibRBD (35 ms total)

fullerdj:
This should be fixed now.

dillaman:
Now this fails on "RBD_FEATURES=109 ./unittest_librbd"

> [ RUN      ] TestLibRBD.CheckObjectMap
> using new format!
> test/librbd/test_librbd.cc:3723: Failure
> Value of: bl1.contents_equal(bl2)
>   Actual: true
> Expected: false
>

fullerdj:
This should actually be fixed now.

dillaman:
lgtm

yuriw:

PR #8563, opened 2016-04-12
rgw: file unix

mattbenjamin:
Above Unix attrs support, this change does use std::map::emplace to assign attributes in several RGWOps.

Not sure if that can go on Jewel, but it has to pass s3-tests anyway--this branch passes my simple S3 object tests as well as the RGW NFS tests.

### DIFF
--- a/src/rgw/rgw_op.h
+++ b/src/rgw/rgw_op.h
@@ -717,6 +717,10 @@ public:
 		 supplied_md5_b64(NULL), supplied_etag(NULL),
 		 data_pending(false) {}
 
+  void emplace_attr(std::string&& key, buffer::list&& bl) {
+    attrs.emplace(key, bl); /* key and bl are r-value refs */
+  }
+
   virtual void init(RGWRados *store, struct req_state *s, RGWHandler *h) {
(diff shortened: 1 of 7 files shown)


src/rgw/rgw_op.h
```diff
@@ -545,6 +546,10 @@ class RGWCreateBucket : public RGWOp {
 public:
   RGWCreateBucket() : has_cors(false) {}
 
+  void emplace_attr(std::string&& key, buffer::list&& bl) {
+    attrs.emplace(key, bl); /* key and bl are r-value refs */
```
cbodley: confusing, but `key` and `bl` are lvalues here.. so you need to re-`move()` them:
`attrs.emplace(std::move(key), std::move(bl));`

i found this to be a decent explanation as to why: [Is an Rvalue Reference an Rvalue?](http://thbecker.net/articles/rvalue_references/section_05.html)

mattbenjamin: I don't find it convincing--but the whole point is to get the compiler to dtrt, so not relevant.

PR #8566, opened 2016-04-12
rgw: RGWPeriod::reflect() sets master zonegroup as default

cbodley:
This patch improves the configuration of a secondary cluster by establishing a default zonegroup after the initial 'realm pull'.

Previously, the initial output of 'zonegroup list' was:
```
read_default_id : -2
{
    "default_info": "",
    "zonegroups": [
        "us",
        "default"
    ]
}
```
With these patches, the output is as expected:
```
read_default_id : 0
{
    "default_info": "792474ac-d2f1-4a29-9c48-ce2dd1f116cf",
    "zonegroups": [
        "us"
    ]
}
```

### DIFF
--- a/src/rgw/rgw_rados.cc
+++ b/src/rgw/rgw_rados.cc
@@ -1204,6 +1204,14 @@ int RGWPeriod::reflect()
       ldout(cct, 0) << "ERROR: failed to store zonegroup info for zonegroup=" << iter.first << ": " << cpp_strerror(-r) << dendl;
       return r;
     }
+    if (zg.is_master_zonegroup()) {
+      // set master as default if no default exists
+      r = zg.set_as_default(true);
+      if (r == 0) {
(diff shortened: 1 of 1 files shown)


src/rgw/rgw_rados.cc
```diff
@@ -1204,6 +1204,8 @@ int RGWPeriod::reflect()
       ldout(cct, 0) << "ERROR: failed to store zonegroup info for zonegroup=" << iter.first << ": " << cpp_strerror(-r) << dendl;
       return r;
     }
+    // set as default if no default exists
+    zg.set_as_default(true);
```
oritwas: I think it is better to set the master zonegroup as default and not the first.

cbodley: sure. i think i'll also add a ldout message if it successfully changes the default

cbodley:
@oritwas updated to only call set_as_default() on master zg

PR #8567, opened 2016-04-12
radosgw-admin: allow 'period pull --url' without full multisite config

cbodley:
to support 'period pull --remote', radosgw-admin needs a complete multisite configuration to look up remote endpoints by name

however, 'period pull' will also accept a --url argument, which doesn't require a working configuration. this patch splits the handling of these two cases, so that 'period pull --url' can succeed when run on a cluster that hasn't yet been fully configured

### DIFF
--- a/src/rgw/rgw_admin.cc
+++ b/src/rgw/rgw_admin.cc
@@ -1333,9 +1333,18 @@ static int send_to_remote_gateway(const string& remote, req_info& info,
   return ret;
 }
 
-static int send_to_url(const string& url, RGWAccessKey& key, req_info& info,
+static int send_to_url(const string& url, const string& access,
+                       const string& secret, req_info& info,
                        bufferlist& in_data, JSONParser& parser)
 {
+  if (access.empty() || secret.empty()) {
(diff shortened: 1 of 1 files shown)


src/rgw/rgw_admin.cc
```diff
@@ -2734,6 +2760,10 @@ int main(int argc, char **argv)
       break;
     case OPT_REALM_PULL:
       {
+        if (url.empty()) {
+          cerr << "A --url must be provided." << std::endl;
+          return EINVAL;
+        }
```
theanalyst: +1 Thanks for getting this, I was  about to ask are there any cases a realm pull can even work without a `--url`

kmroz:
Cool, LGTM also.

src/rgw/rgw_admin.cc
```diff
@@ -2536,6 +2542,26 @@ int main(int argc, char **argv)
 	}
       }
       break;
+    case OPT_PERIOD_PULL: // period pull --url
+      {
+        if (url.empty()) {
```
oritwas: don't you need to check that remote is empty too?

PR #8568, opened 2016-04-13
OSD and librados writesame support

ddiss:
This is a follow up to https://github.com/ceph/ceph/pull/8175 , which adds a librados API and tests atop the OSD writesame implementation from @mikechristie. I'll submit the cmpext support separately.
A few things potentially worth changing:
- allow for librados zero-length writesame requests
- use .*write_same() instead of *writesame(), to be consistent with write_full()
- add librados striper API and tests (?)

### DIFF
--- a/src/include/rados.h
+++ b/src/include/rados.h
@@ -256,6 +256,9 @@ extern const char *ceph_osd_state_name(int s);
 	f(CACHE_PIN,	__CEPH_OSD_OP(WR, DATA, 36),	"cache-pin")        \
 	f(CACHE_UNPIN,	__CEPH_OSD_OP(WR, DATA, 37),	"cache-unpin")      \
 									    \
+	/* ESX/SCSI */							    \
+	f(WRITESAME,	__CEPH_OSD_OP(WR, DATA, 38),	"write-same")	    \
+									    \
 	/** multi **/							    \
 	f(CLONERANGE,	__CEPH_OSD_OP(WR, MULTI, 1),	"clonerange")	    \
 	f(ASSERT_SRC_VERSION, __CEPH_OSD_OP(RD, MULTI, 2), "assert-src-version") \
@@ -533,6 +536,11 @@ struct ceph_osd_op {
 			__le64 expected_object_size;
 			__le64 expected_write_size;
 		} __attribute__ ((packed)) alloc_hint;
+		struct {
+			__le64 offset;
+			__le64 length;
+			__le64 data_length;
+		} __attribute__ ((packed)) writesame;
 	};
 	__le32 payload_len;
 } __attribute__ ((packed));
--- a/src/include/rados/librados.h
+++ b/src/include/rados/librados.h
@@ -283,8 +283,8 @@ struct rados_cluster_stat_t {
  *   rados_write_op_assert_version()
  * - Creating objects: rados_write_op_create()
  * - IO on objects: rados_write_op_append(), rados_write_op_write(), rados_write_op_zero
- *   rados_write_op_write_full(), rados_write_op_remove, rados_write_op_truncate(),
- *   rados_write_op_zero()
+ *   rados_write_op_write_full(), rados_write_op_writesame(), rados_write_op_remove,
+ *   rados_write_op_truncate(), rados_write_op_zero()
  * - Hints: rados_write_op_set_alloc_hint()
  * - Performing the operation: rados_write_op_operate(), rados_aio_write_op_operate()
  */
@@ -1334,6 +1334,24 @@ CEPH_RADOS_API int rados_write_full(rados_ioctx_t io, const char *oid,
                                     const char *buf, size_t len);
 
 /**
+ * Write the same *data_len* bytes from *buf* multiple times into the
+ * *oid* object. *write_len* bytes are written in total, which must be
+ * a multiple of *data_len*. The value of *write_len* and *data_len*
+ * must be <= UINT_MAX/2.
+ *
+ * @param io the io context in which the write will occur
+ * @param oid name of the object
+ * @param buf data to write
+ * @param data_len length of the data, in bytes
+ * @param write_len the total number of bytes to write
+ * @param off byte offset in the object to begin writing at
+ * @returns 0 on success, negative error code on failure
+ */
+CEPH_RADOS_API int rados_writesame(rados_ioctx_t io, const char *oid,
+                                   const char *buf, size_t data_len,
+                                   size_t write_len, uint64_t off);
+
+/**
  * Efficiently copy a portion of one object to another
  *
  * If the underlying filesystem on the OSD supports it, this will be a
@@ -1888,6 +1906,29 @@ CEPH_RADOS_API int rados_aio_append(rados_ioctx_t io, const char *oid,
 CEPH_RADOS_API int rados_aio_write_full(rados_ioctx_t io, const char *oid,
 			                rados_completion_t completion,
 			                const char *buf, size_t len);
+
+/**
+ * Asychronously write the same buffer multiple times
+ *
+ * Queues the writesame and returns.
+ *
+ * The return value of the completion will be 0 on success, negative
+ * error code on failure.
+ *
+ * @param io the io context in which the write will occur
+ * @param oid name of the object
+ * @param completion what to do when the writesame is safe and complete
+ * @param buf data to write
+ * @param data_len length of the data, in bytes
+ * @param write_len the total number of bytes to write
+ * @param off byte offset in the object to begin writing at
+ * @returns 0 on success, -EROFS if the io context specifies a snap_seq
+ * other than LIBRADOS_SNAP_HEAD
+ */
+CEPH_RADOS_API int rados_aio_writesame(rados_ioctx_t io, const char *oid,
+			               rados_completion_t completion,
+			               const char *buf, size_t data_len,
+				       size_t write_len, uint64_t off);
 
 /**
  * Asychronously remove an object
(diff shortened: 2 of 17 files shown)


src/include/rados.h
```diff
@@ -256,6 +256,9 @@ extern const char *ceph_osd_state_name(int s);
 	f(CACHE_PIN,	__CEPH_OSD_OP(WR, DATA, 36),	"cache-pin")        \
```
jdurgin: should get rid of the 'cherry-picked from' in the commit message, since that ref won't exist after this one is merged

jdurgin:
looking good, nice a thorough on the various forms of librados i/o operations. writesame as the method name is fine with me. A libradosstriper api isn't necessary. 

The one other test @athanatos was interested in seeing was for ceph_test_rados - extending src/test/osd/RadosModel.h to add writesame ops.

src/osd/ReplicatedPG.cc
```diff
@@ -5085,6 +5119,12 @@ int ReplicatedPG::do_osd_ops(OpContext *ctx, vector<OSDOp>& ops)
       }
       break;
 
+    case CEPH_OSD_OP_WRITESAME:
+      ++ctx->num_write;
+      tracepoint(osd, do_osd_op_pre_writesame, soid.oid.name.c_str(), soid.snap.val, oi.size, op.writesame.offset, op.writesame.length, op.writesame.data_length);
+      result = do_writesame(ctx, osd_op);
```
yuyuyu101: write_update_size_and_usage?

src/osd/ReplicatedPG.cc
```diff
@@ -3701,6 +3701,40 @@ int ReplicatedPG::do_xattr_cmp_str(int op, string& v1s, bufferlist& xattr)
...
+  if (!op.writesame.data_length || write_length % op.writesame.data_length)
+    return -EINVAL;
+
+  if (op.writesame.data_length != osd_op.indata.length()) {
+    derr << "invalid length ws data length " << op.writesame.data_length << " actual len " << osd_op.indata.length() << dendl;
+    return -EINVAL;
+  }
+
+  while (write_length) {
+    write_op.indata.append(osd_op.indata.c_str(), op.writesame.data_length);
```
yuyuyu101: I think this should be append ptr instead of char*, otherwise we will result in oom if write_length is large(we permit UINT_64_MAX/2)

yuyuyu101:
:+1:

src/include/rados/librados.h
```diff
@@ -1334,6 +1334,24 @@ CEPH_RADOS_API int rados_write_full(rados_ioctx_t io, const char *oid,
...
+ * @param oid name of the object
+ * @param buf data to write
+ * @param data_len length of the data, in bytes
+ * @param write_len the total number of bytes to write
+ * @param off byte offset in the object to begin writing at
+ * @returns 0 on success, negative error code on failure
+ */
+CEPH_RADOS_API int rados_writesame(rados_ioctx_t io, const char *oid,
+                                   const char *buf, size_t data_len,
+                                   size_t write_len, uint64_t off);
```
yuyuyu101: I suggest we change write_len to write_count, so we don't need to care about the data_len align with write_len. And it will be more intuitionistic to caller to reduce potential mistake.

ddiss:
@jdurgin and @yuyuyu101: thanks a lot for the feedback, will integrate your suggested changes and re-push.

src/include/rados.h
ddiss: Oops, will drop that in the next round.

src/include/rados/librados.h
ddiss: Am a little undecided here, as we may choose to drop the alignment requirement in future.
I'll leave the wire protocol as is, and rework the client API to take a write_count multiplier.

src/osd/ReplicatedPG.cc
ddiss: do_writesame() recurses with do_osd_ops(op = CEPH_OSD_OP_WRITE), so I don't think an extra write_update_size_and_usage() call is needed here.

src/include/rados/librados.h
jdurgin: If you think we might drop the alignment requirement I'd leave the api as-is, so we don't need a new version of the function to handle that

ddiss: Thanks. The new round I just pushed uses the same data_len + write_len API.

ddiss:
Ping - did anyone get a chance to look at the latest round of changes?

jdurgin:
Yes, it's looking good. Ran it through the rados suite last week, and it passed, but I've been traveling this week.

Will look at merging next week when I'm at a real computer.

src/include/rados/librados.hpp
```diff
@@ -683,6 +685,9 @@ namespace librados
      * NOTE: this call steals the contents of @param bl.
      */
     int write_full(const std::string& oid, bufferlist& bl);
+    int writesame(const std::string& oid, bufferlist& bl,
+		  size_t data_len, size_t write_len,
```
yuyuyu101: why would we have data_len? could we assume bl.length == data_len?

src/librados/librados.cc
```diff
@@ -3279,6 +3303,27 @@ extern "C" int rados_write_full(rados_ioctx_t io, const char *o, const char *buf
...
+				const char *buf,
+				size_t data_len,
+				size_t write_len,
+				uint64_t off)
+{
+  tracepoint(librados, rados_writesame_enter, io, o, buf, data_len, write_len, off);
+  if ((data_len > UINT_MAX/2) || (write_len > UINT_MAX/2))
+    return -E2BIG;
+  if ((data_len == 0) || (write_len % data_len))
+    return -EINVAL;
```
yuyuyu101: it looks dup with IoCtxImpl argument check

src/librados/librados.cc
```diff
@@ -4338,6 +4383,27 @@ extern "C" int rados_aio_write_full(rados_ioctx_t io, const char *o,
...
 }
 
+extern "C" int rados_aio_writesame(rados_ioctx_t io, const char *o,
+				   rados_completion_t completion,
+				   const char *buf, size_t data_len,
+				   size_t write_len, uint64_t off)
+{
+  tracepoint(librados, rados_aio_writesame_enter, io, o, completion, buf,
+						data_len, write_len, off);
+  if ((data_len > UINT_MAX/2) || (write_len > UINT_MAX/2))
```
yuyuyu101: ditto

src/include/rados/librados.hpp
ddiss: Yes, that would be a valid assumption.
I added the data_len parameter to be consistent with write(const std::string& oid, bufferlist& bl, size_t len, uint64_t off) here. I can remove the parameter if desired.

src/librados/librados.cc
ddiss: You're right, I'll fix this (and rados_aio_writesame) in the next round - thanks!

src/include/rados/librados.hpp
yuyuyu101: yes, I think it could remove redundant argument check

ddiss: @jdurgin: any preference?

jdurgin: sounds good to me. Not sure why we have the len parameter for write().

ddiss:
Changes since last version:
- removed duplicate parameter checks
- removed redundant data_len parameter

Feedback appreciated

yuyuyu101:
good to me, thanks!

jdurgin:
lgtm too

src/osd/ReplicatedPG.cc
```diff
@@ -3701,6 +3701,40 @@ int ReplicatedPG::do_xattr_cmp_str(int op, string& v1s, bufferlist& xattr)
...
+    write_length -= op.writesame.data_length;
+  }
+
+  write_op.op.op = CEPH_OSD_OP_WRITE;
```
yuyuyu101: Hmm, I think we need to use CEPH_OSD_OP_ZERO instead. Otherwise, writesame op doesn't actually improve vmware zero operation performance.

yuyuyu101: I mean for writezero the special but common case, we need to handle this properly..

ddiss: I don't think that kind of optimisation belongs here - IMO it would be much better handled on the client (i.e. iSCSI gateway). That way it could avoid sending the (zeroed) buffer in the first place.

yuyuyu101: no, I mean ObjectStore support zero operation like fallocate api. It will much better than write zero bufferlist.

yuyuyu101: I can't follow your meaning, what's the meaning of "That way it could avoid sending the (zeroed) buffer in the first place."

jdurgin: I think David is suggesting that the client should send a zero osd op instead of using writesame to zero out an object.

yuyuyu101: ok

ddiss: @jdurgin: Exactly.

@yuyuyu101: The (current) use-case for CEPH_OSD_OP_WRITESAME is to offload handling of SCSI WRITE SAME I/O to the OSDs.
SCSI WRITE SAME can be used by a SCSI initiator in a couple of ways:
1) write an arbitrary buffer multiple times to disk
2) trim/discard a range on a thin-provisioned disk (UNMAP=1)

For type (2) WRITE SAME handling, the kernel client uses CEPH_OSD_OP_DELETE, CEPH_OSD_OP_TRUNCATE or CEPH_OSD_OP_ZERO to implement the trim/discard behaviour.

For type (1) WRITE SAME handling, CEPH_OSD_OP_WRITESAME will initially be used. CEPH_OSD_OP_ZERO could be used later to optimise the case where the buffer is all zeros.

PR #8570, opened 2016-04-13
osd: clear requeue_scrub_on_unblock flag after requeue scrub

xiexingguo:
We'd better clear requeue_scrub_on_unblock flag after successfully
requeuing scrub, this is for:
  1. if a newly initiated scrub is blocked by objectcontext again,
     reset requeue_scrub_on_unblock to true will be satisfied enough
     to meet such a appeal.
  2. so we don't push pg to the scrub queue when we don't want to
     by accident.

### DIFF
--- a/src/osd/ReplicatedPG.cc
+++ b/src/osd/ReplicatedPG.cc
@@ -9288,8 +9288,10 @@ void ReplicatedPG::kick_object_context_blocked(ObjectContextRef obc)
     objects_blocked_on_snap_promotion.erase(i);
   }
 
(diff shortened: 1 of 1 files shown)


xiexingguo:
@tchaikov Please help to confirm, thanks.

tchaikov:
lgtm.

tchaikov:
imo, the chance is rare that an object can be blocked twice. but it's always good to avoid unnecessary scrubs.

xiexingguo:
@tchaikov  Agree, thanks for the quick review:)

PR #8572, opened 2016-04-13
osd: allow eviction on non-legacy watched object

wonzhq:
For v2 watch, the watchers are disconnected while doing eviction.
Clients can reconnect to the object in the base. This is to support
removing a cache tier on active RBD pools where the RBD image header
is watched by active clients.

Fixes: http://tracker.ceph.com/issues/14865
Signed-off-by: Zhiqiang Wang zhiqiang@xsky.com

### DIFF
--- a/src/osd/ReplicatedPG.cc
+++ b/src/osd/ReplicatedPG.cc
@@ -4593,10 +4593,23 @@ int ReplicatedPG::do_osd_ops(OpContext *ctx, vector<OSDOp>& ops)
 	  result = -EBUSY;
 	  break;
 	}
-	if (!oi.watchers.empty()) {
+
+	// return EBUSY if we have legacy watcher. For v2 watch, the watchers
+	// are disconnected in _delete_oid, clients will reconnect to the object
+	// in the base after the eviction.
+	bool have_legacy_watcher = false;
+        for (map<pair<uint64_t, entity_name_t>, WatchRef>::iterator it =
+	     ctx->obc->watchers.begin(); it != ctx->obc->watchers.end(); ++it) {
+	  if (it->second->is_legacy_watch()) {
+	    have_legacy_watcher = true;
+	    break;
+	  }
+        }
+	if (have_legacy_watcher) {
 	  result = -EBUSY;
 	  break;
 	}
+
 	if (soid.snap == CEPH_NOSNAP) {
 	  result = _verify_no_head_clones(soid, ssc->snapset);
 	  if (result < 0)
--- a/src/osd/Watch.h
+++ b/src/osd/Watch.h
@@ -227,6 +227,7 @@ public:
   entity_name_t get_entity() const { return entity; }
   entity_addr_t get_peer_addr() const { return addr; }
   uint32_t get_timeout() const { return timeout; }
+  bool is_legacy_watch() const { return !will_ping; }
 
   /// Generates context for use if watch timeout is delayed by scrub or recovery
   Context *get_delayed_cb();
--- a/src/test/librados/tier.cc
+++ b/src/test/librados/tier.cc
@@ -2717,6 +2717,121 @@ TEST_F(LibRadosTwoPoolsPP, CachePin) {
   cluster.wait_for_latest_osdmap();
 }
 
+
+int notify_sleep = 0;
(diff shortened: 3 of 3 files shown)


src/osd/ReplicatedPG.cc
```diff
@@ -4593,10 +4593,23 @@ int ReplicatedPG::do_osd_ops(OpContext *ctx, vector<OSDOp>& ops)
...
 	  break;
 	}
-	if (!oi.watchers.empty()) {
+
+	// return EBUSY if we have legacy watcher. For v2 watch, the watchers
+	// are disconnected in _delete_oid, clients will reconnect to the object
+	// in the base after the eviction.
+	bool have_legacy_watcher = false;
+        for (map<pair<uint64_t, entity_name_t>, WatchRef>::iterator it =
+	     ctx->obc->watchers.begin(); it != ctx->obc->watchers.end(); ++it) {
```
liewegas: for (auto w : ctx->obc->watchers)

liewegas:
I wonder if we can do better.  For example, the disconnect we queue in _delete_oid passes true for send_disconnect.  If we don't send a disconnect, it could try to reconnect to the same object in the base tier.... except that we force promotion for watch ops.

Hrm.. in order for somethign like that to work, I think the base tier would need to disconnect watchers when the object is promoted, so that we can be sure all watches are on the same tier?

liewegas:
In the meantime, this is probably better than nothing... but...

wonzhq:
@liewegas I agree with that the base tier needs to disconnect watchers when the object is promoted. But when evicting an object, isn't it better to send disconnects to all the watchers, so that they can reconnect to the base object ASAP (we are removing the cache tier, and the cache mode has been changed to forward in this case)?

liewegas:
Hmm, you're right that we need to send something, since there isn't a PG 
change to make the client reconnect.  But a disconnect like this is an 
error that surfaces up to the librados user, and I think what we actually 
want is for the objecter to do it's internal reconnect attempt (like it 
does if the PG moves), and surface an error only if that fails.

I guess this is fine in the interim, since well-written clients will 
try to reconnect and revalidate all of their state.  But it's more violent 
than it should be.

We probably need a new event type CEPH_WATCH_EVENT_MUST_RECONNECT or 
something...

wonzhq:
Got you. Yes it's better to come up with a new event type.

liewegas:
@athanatos what do you think we should do here?

1) merge this as is.  it'll let you cache-flush an rbd header, and librbd should reconnect.
2) extend the watch/notify protocol so that there is a graceful way to ask the client to reconnect.

I think 2 will get messy.

liewegas:
@wonzhq I think we need to add a new test/librados/tier.cc test that creates an object in teh cache tier, sets a watch, and does a cache-flush and evict, and verifies evict returns 0 and the watch  gets disconnected.

wonzhq:
@liewegas, sorry I wasn't able to work on this in the last couple of days for some reason. So we are going ahead with option 1? Personally I prefer option 2. Anyway, I'll add the test case ASAP.

wonzhq:
@liewegas test added, pls take another look, thx!

athanatos:
We need a test that verifies rbd's behavior as well.

athanatos:
Option 1 is I think adequate for now.  If we want to do option 2, first ste

athanatos:
Oops, didn't mean to close.

athanatos:
@wonzhq We still need an rbd test.

dillaman:
@wonzhq ping

wonzhq:
Sorry, I'm not sure how to add the rbd test. Pls advise..

athanatos:
@dillaman What would @wonzhq need to do to get rbd to exercise the code?

dillaman:
@wonzhq If you could write a bash script in "qa/workunits/rbd" that creates an image, runs "rbd watch" against the image in the background, ensure the image header is promoted to the cache pool, demote it, and run "rbd resize" against the image and verify that your background "rbd watch" sees the notification.

liewegas:
@wonzhq ping

liewegas:
please reopen when you have time to address this!

PR #8573, opened 2016-04-13
hammer: ECBackend: send write and read sub ops on behalf of client ops at normal priority

Vicente-Cheng:
http://tracker.ceph.com/issues/15293


smithfarm:

smithfarm:
@athanatos This PR passed a /200 rados run on Ubuntu. None of the failures were reproducible. For details see http://tracker.ceph.com/issues/15895#note-18

athanatos:
lgtm

PR #8574, opened 2016-04-13
test/encoding/readable: use [ for "test" not ((

tchaikov:
this also fixes following error in the ubuntu gitbuilder

error: Added files:
+ cat .git/added-files
src/0

where sh points to dash.

### DIFF
--- a/src/test/encoding/readable.sh
+++ b/src/test/encoding/readable.sh
@@ -149,7 +149,7 @@ waitall() { # PID...
          errors=$(($errors + 1))
        fi
      done
-     (("$#" > 0)) || break
+     [ $# -eq 0 ] && break
      sleep ${WAITALL_DELAY:-1}
     done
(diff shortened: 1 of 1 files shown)


tchaikov:
@ErwanAliasr1 could you take a look?

ErwanAliasr1:
I'm ok with the syntax but aren't we expecting to test $? instead of $# ?

tchaikov:
iiuc, we are testing the number of pids to wait for, but not the return code of last command.

ErwanAliasr1:
Oh ok, didn't read it like that.
+1 for me then.

tchaikov:
build and test passed at http://gitbuilder.sepia.ceph.com/gitbuilder-ceph-tarball-trusty-amd64-basic/log.cgi?log=fc81bec888eba3be488c34508d4ae3521372b175

PR #8577, opened 2016-04-13
osdc/Objecter: small cleanup for osd op completion lock

runsisi:
OSDSession::get_lock also checks the emptiness of oid name, so no need
to check it twice


scienceluo:
For double insurance?

PR #8581, opened 2016-04-13
doc: adding pg_num to pools documentation

emenguy:
Current documentation only makes reference to pgp_num to get or set value.
http://docs.ceph.com/docs/hammer/rados/operations/pools/

pg_num documentation already exists here http://docs.ceph.com/docs/master/rados/operations/placement-groups/#set-the-number-of-placement-groups

Had to change pgp_num description to distinguish them.

Signed-off-by: Etienne Menguy <etienne.menguy@corp.ovh.com>

### DIFF
--- a/doc/rados/operations/pools.rst
+++ b/doc/rados/operations/pools.rst
@@ -266,12 +266,21 @@ You may set values for the following keys:
               
 :Type: Integer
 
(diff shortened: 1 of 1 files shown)


theanalyst:
LGTM

PR #8582, opened 2016-04-13
MDS: don't pin purging denties

ukernel:


gregsfortytwo:
@jcsp I think you looked at this some already, so please provide any comments you already had.

gregsfortytwo:
@ukernel, what testing was done?

ukernel:
I set mds_max_purge_files to 0, then create and delete lots of files.

jcsp:
I've exercised this a bit here, haven't found any bugs.  I see the new push, was that fixes for snapshots?

It's kind of awkward to test because when we set a small cache size, we also need to aggressively flush the journal to avoid the cache size being exceeded due to references to log segments pinning the inodes in cache.

We do still have the trimmed_strays set here, which will grow indefinitely, although of course it uses much less space per dentry than actually having it pinned in the cache.

I would like to have an automated test that triggers some trimming: set cache size to N, delete more than N files (with max purge ops = 0), see that the inode count goes down to N, and then set max purge ops to something finite and see that all the strays eventually get purged.  That would exercise both the path where we trim, and also the path where we recover things we trimmed.

It seems like scan_stray_dir is still going to load all the stray dirs at once on startup.  That seems like a pretty big issue, unless there is some mitigation that I'm not seeing?

ukernel:
On Wed, May 11, 2016 at 11:01 PM, John Spray <notifications@github.com>
wrote:

> I've exercised this a bit here, haven't found any bugs. I see the new
> push, was that fixes for snapshots?
>

I add code that open inode->snaprealm's  past parents if the inode had been
trimmed.


> It's kind of awkward to test because when we set a small cache size, we
> also need to aggressively flush the journal to avoid the cache size being
> exceeded due to references to log segments pinning the inodes in cache.
>
> We do still have the trimmed_strays set here, which will grow
> indefinitely, although of course it uses much less space per dentry than
> actually having it pinned in the cache.
>
> I would like to have an automated test that triggers some trimming: set
> cache size to N, delete more than N files (with max purge ops = 0), see
> that the inode count goes down to N, and then set max purge ops to
> something finite and see that all the strays eventually get purged. That
> would exercise both the path where we trim, and also the path where we
> recover things we trimmed.
>
> It seems like scan_stray_dir is still going to load all the stray dirs at
> once on startup. That seems like a pretty big issue, unless there is some
> mitigation that I'm not seeing?
>
scan_stray_dir() loads stray directory one by one.  After calling
maybe_eval_stray(), stray dentries can be trimmed from cache. It's not
perfect but much better than old code.

jcsp:
> scan_stray_dir() loads stray directory one by one.  After calling
maybe_eval_stray(), stray dentries can be trimmed from cache. It's not
perfect but much better than old code.

Yeah, it is a significant improvement.  I think we should merge this once it's passed an fs suite.

ukernel:

PR #8585, opened 2016-04-13
rgw: call rgw_log_usage_finalize() on reconfiguration

cbodley:
RGWRealmReloader was calling rgw_log_usage_init() with the new RGWRados instance, but never cleaned up the previous one with rgw_log_usage_finalize()

Fixes a crash caused by the previous UsageLogger using the RGWRados instance after it was freed by RGWRealmReloader.


mattbenjamin:
@oritwas did you see a problem with this?

oritwas:
lgtm

PR #8586, opened 2016-04-13
osd: fix watch reconnect race

liewegas:


athanatos:
lgtm

src/msg/simple/DispatchQueue.cc
```diff
@@ -150,6 +150,14 @@ void DispatchQueue::entry()
       lock.Unlock();
 
       if (qitem.is_code()) {
+	if (g_conf->ms_inject_internal_delays > 0 &&
```
xiexingguo: nit, ``` g_conf->ms_inject_internal_delays &&``` will suffice

src/msg/simple/DispatchQueue.cc
```diff
@@ -150,6 +150,14 @@ void DispatchQueue::entry()
...
+	    g_conf->ms_inject_delay_probability &&
+	    (rand() % 1000) / 1000.0 > g_conf->ms_inject_delay_probability) {
```
xiexingguo: The inject probability judgement logic is a little bit confusing. The chance of injecting delay shall go high as the ms_inject_delay_probability increases, e.g., 
```
1. if ms_inject_delay_probability == 0, we don't inject delay;
2. if ms_inject_delay_probability == 1, we shall 100% inject delay;
```
It's quite the opposite logic here, especially when ms_inject_delay_probability is set to 1, we will absolutely **not** inject delay, as ```(rand() % 1000) / 1000.0```  is always smaller than 1.

xiexingguo:
Otherwise LGTM

tchaikov:
looks like it's a case where CAS is needed..

i am thinking about a racing here. we have two threads:
- over a new connection, the client's RECONNECT op leads to `Watch::connect(con)`
- when the old connection is reset, `OSD::ms_handle_reset(con)` gets called, `WatchConState::reset(con)` is in turn called.

let's look at following sequence:

1 . in `WatchConState::reset(con)`// because con is the one being reset, so, we are about to reset `this->conn`. but we are switched to another thread before we reset it.
 ```c++
// with pg lock
if ((*i)->is_connected(con)) {
  // we switch to another thread here
  (*i)->disconnect();
}
  ```
2 . in `Watch::connect(con)` // good, this->conn is not the one we are connected, change it to the new conn
```c++
if (this->conn == con) return;
this->conn = con;
```
3 . continue with `WatchConState::reset(con)` // and the `this->conn` gets reset. `ETIMEDOUT` waiting ahead.
```c++
if ((*i)->is_connected(con)) {
  // we continue from here
  (*i)->disconnect();
}
```

it sounds like a fabricated case. but it could happen in real-world. i don't see we have the pg lock acquired in `ReplicatedPG::do_osd_op_effects()` when calling `watch->connect()`, probably we should? otherwise we need an `atomic<>` member variable in `Watch` class for the CAS operation so `Watch.conn` is safe when being updated by multiple threads.

xiexingguo:
@tchaikov Well done.





> it sounds like a fabricated case. but it could happen in real-world. i don't see we have the pg lock acquired in ReplicatedPG::do_osd_op_effects() when calling watch->connect(), probably we should?

I doubt we can do this, because we may call  ReplicatedPG::do_osd_op_effects()  in the following sequences:
```
ReplicatedPG::do_op(OpRequestRef& op)
                                |
                               V
ReplicatedPG::execute_ctx(OpContext *ctx)
                                |
                               V
ReplicatedPG::do_osd_op_effects()
```
which means we shall already have pg-lock in hand under this case.

tchaikov:
thanks @xiexingguo, i also suspect that the do_osd_op_effects() is guarded by the pg lock, but failed to find the place where we acquire the lock. could you point me to that line?

xiexingguo:
@tchaikov In the above case will take pg lock at line https://github.com/ceph/ceph/blob/master/src/osd/OSD.cc#L8610.
And at line https://github.com/ceph/ceph/blob/master/src/osd/ReplicatedPG.cc#L2991 we will call do_osd_op_effects() if op is pure read or we encounter a error.

In the above case, we are in the ShardedOpWQ with pg lock held.
But we may also call do_osd_op_effects() in FileStore's OpWQ, in which case we are unprotected by pg lock.

xiexingguo:
@tchaikov   Ping me if I fail to make myself crystal clear:)

tchaikov:
thank you @xiexingguo . got it now.

but the change lgtm. the chance of the racing is rare. maybe we can fix it after jewel.

liewegas:
The on_success callbacks also are called with pg lock held, from eval_repop.  Not called from filestore.

xiexingguo:
I finally figured it out, thanks:)

PR #8588, opened 2016-04-13
rgw: try to use current period id in a few more cases

yehudasa:
If missing period id, try to use the current period id. This is needed
for sync agent compatibility.


oritwas:
lgtm

PR #8592, opened 2016-04-14
rgw: Add a test for multi-tenancy

zaitcev:
This is mostly a PR so we can talk about it. I know other people extended test_multi.py for their purposes and I'm looking to factor out some common code. But if this goes in, it's fine too - I see no harm or technical debt, only profit.

The biggest area of concern is how I fixed the double-eval problem. For the best compatibility it's probably beneficial to fix as high in the stack as possible. Ergo, test-rgw-call.sh.


mattbenjamin:
lgtm -- if no one objects in today's call, we should just merge it

oritwas:
lgtm

zaitcev:
Matt, Orit - I have just nullified your comments, sorry. I fixed a couple of XXX-es, and for that I had to patch existing code. I need to know if I broke other tests, including those not yet merged.

yehudasa:
lgtm

zaitcev:
Build errored out somewhere else with:

  CXX      osd/libosd_a-OSD.o
  CXX      osd/libosd_a-OSDCap.o
osd/OSD.cc: In member function ‘int OSD::init()’:
osd/OSD.cc:2040:19: error: ‘struct md_config_t’ has no member named ‘osd_check_max_object_name_len_on_startup’
       if (g_conf->osd_check_max_object_name_len_on_startup) {
                   ^

I'm going to wait for the trunk to churn a bit, then re-pull and re-push.

PR #8596, opened 2016-04-14
os/filestore: add sanity check for stat() syscall

xiexingguo:
The syscall stat() may fail here, the hardlink output shall
be assigned a rubbish value as a result and mislead caller.


liewegas:
Do you have a log for when you saw ENOENT from rmcoll on bluestore?  It doesn't do log replay, so if there is a bad rmcoll it's probably because the caller is buggy.

xiexingguo:
I don't. I happened to  remember that FileStore handled ENOENT on OP_RMCOLL that way, so I think we'd better keep pace with FileStore. Would you like me to drop that change or both of them?

liewegas:
FileStore only needed to do this because the log replay items needed to be 
idempotent.  BlueStore won't need it.

The other patch looks ok!

PR #8597, opened 2016-04-14
rgw: handle no current_period in is_syncing_bucket_meta

oritwas:
Fix bucket instance leak


yehudasa:
lgtm

PR #8598, opened 2016-04-14
mds: fix usage() display 

renhwztetecs:
The usage using derr, will display the date and time;
remove unnecessary break.

Signed-off-by: renhwztetecs <ren.huanwen@zte.com.cn>

### DIFF
--- a/src/ceph_mds.cc
+++ b/src/ceph_mds.cc
@@ -107,8 +107,8 @@ int main(int argc, const char **argv)
       break;
     }
     else if (ceph_argparse_flag(args, i, "--help", "-h", (char*)NULL)) {
+      // exit(1) will be called in the usage()
       usage();
-      break;
     }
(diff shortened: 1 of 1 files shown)


src/ceph_mds.cc
```diff
@@ -108,7 +108,6 @@ int main(int argc, const char **argv)
     }
     else if (ceph_argparse_flag(args, i, "--help", "-h", (char*)NULL)) {
       usage();
-      break;
```
jcsp: It's not obvious when reading that usage() will ultimately call exit(), so a comment would be nice here

jcsp:
+1 for the fix to the usage part, Reviewed-By

Would be nice to add a comment where the break was removed

renhwztetecs:
@jcsp 
thanks for your review
I  add a comment where the break was removed

jcsp:

PR #8599, opened 2016-04-14
os: fix submit_transaction return codes

yangdongsheng:
Fixes: http://tracker.ceph.com/issues/15346

This series of patches are going to introduce a mechanism to allow transaction to be aborted if we meet something wrong in submitting transaction.

### DIFF
--- a/src/os/bluestore/BlueStore.cc
+++ b/src/os/bluestore/BlueStore.cc
@@ -3720,7 +3720,7 @@ void BlueStore::_txc_finish_io(TransContext *txc)
 	   p->state == TransContext::STATE_IO_DONE);
 }
 
-int BlueStore::_txc_finalize(OpSequencer *osr, TransContext *txc)
+void BlueStore::_txc_finalize(OpSequencer *osr, TransContext *txc)
 {
   dout(20) << __func__ << " osr " << osr << " txc " << txc
 	   << " onodes " << txc->onodes << dendl;
@@ -3764,8 +3764,6 @@ int BlueStore::_txc_finalize(OpSequencer *osr, TransContext *txc)
     get_wal_key(txc->wal_txn->seq, &key);
     txc->t->set(PREFIX_WAL, key, bl);
   }
-
-  return 0;
 }
 
 void BlueStore::_txc_finish_kv(TransContext *txc)
@@ -4262,7 +4260,6 @@ int BlueStore::queue_transactions(
   Context *onreadable_sync;
   ObjectStore::Transaction::collect_contexts(
     tls, &onreadable, &ondisk, &onreadable_sync);
-  int r;
 
   // set up the sequencer
   OpSequencer *osr;
@@ -4290,8 +4287,7 @@ int BlueStore::queue_transactions(
     _txc_add_transaction(txc, &(*p));
   }
 
-  r = _txc_finalize(osr, txc);
-  assert(r == 0);
+  _txc_finalize(osr, txc);
 
   throttle_ops.get(txc->ops);
   throttle_bytes.get(txc->bytes);
--- a/src/os/bluestore/BlueStore.cc
+++ b/src/os/bluestore/BlueStore.cc
@@ -1785,7 +1785,7 @@ int BlueStore::mkfs()
     } else {
       fm->release(reserved, end, t);
     }
-    db->submit_transaction_sync(t);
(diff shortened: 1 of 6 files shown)


yuyuyu101:
interesting idea, but I'm not sure what's the benefit get from this? Currently if ObjectStore can't handle a error, it will go to crash. If we promote this unexpected to upper, I think it may be too heavy to let caller process. We need a lot of things to make it consistence instead of just return error to client.

yangdongsheng:
@yuyuyu101 Yes, currently, the filestore will crash if we meet an error in submitting transaction. But Kstore and bluestore do not handle it. TBH, I was considering to crash it immediately by assert(0). But I think getting a chance such as on_abort to transaction is a better idea, (even we finally do assert(0) in the on_abort Context.)

Anyway, good point about the consistence here about it, at least, I think from my point of view, we need to allow a transaction to be aborted, that sounds more natural, I think.

But about how to handle it in C_OSD_RepopAbort, maybe crash it by assert(0), or maybe keep it alive but returning EIO, I am not sure now.

I want to continue this work and figure out what to do in C_OSD_RepopAbort, is that a good plan?

liewegas:
I also don't know what benefit we get from a transaction abort.  We have to be really careful not to commit anything that's incorrect, and if we encounter an error that suggests a hardware failure or bug, both of which should trigger an osd abort.  At least under the current model.

In the future, we might want to be able to "fail" on a PG-basis.  Or stop the entire OSD instance (including the ObjectStore) for when we put multiple OSDs in a single process.  In the latter case it's a matter of a more graceful stop (than assert(0)), not an 'abort' in the traditional sense.

yangdongsheng:
Okey, @yuyuyu101 and @liewegas , so both of you are suggesting to assert(0) immediately if we encounter an error in submitting transaction in current model, right? Hmmm, yes, the consistency point makes sense to me.

Okey, anyway, we did not handle some cases in Kstore and Bluestore where we should assert(0) currently, I will update this branch soon.

Thank you two guys. :)

yangdongsheng:
Hi, @yuyuyu101 and @liewegas , as a new comer to ceph, I have one question here, could you help to answer it?

When we are writing data, the ReplicateBackend will issue_op to different replicas. then there are two different cases might happen to crash:

(1), the primary osd failed to write data into device, (maybe hardware error)
Then the primary osd will crash immediately, but the client (maybe rados put command) is still listening on the reply message, how can it know this writing is failed and stop waiting?

(2). the replica failed to write data into device, 
The similar question with the case (1), if one replica encounter an error and crashed, how can the primary know it? If we assert(0) immediately in replica without send_message_osd_cluster() to primary. 

There must be something I am missing, could you help me to understand it? thanx in advance. :)

yangdongsheng:
Is that peering? So in these cases:
(1), primary osd down,
Then we will find it by tick(), then choose another primary and calculate a new acting list. then a peering would happen, so all of this are transparent to the client (rados put command), right? then rados put command will not fail but will wait a longer time for the completion, right?

(2), replica osd down,
Then we will find it by tick() similarly, then kick it out and calculate a new acting list. then a peering would happen.

Am I understanding it correctly?

liewegas:
1- the client will resend when it learns about the failure from a new osdmap.
2- the primary will restart the request when it learns about the replica failure from a new osdmap and repeers.

1- right
2- right.  although in both these cases, if we think there is a failure we tell the mon, but it is the mon publishing a new osdmap that makes it official. the heartbeat checks in tick just help the mon figure it out.

liewegas:
os/kstore/KStore.cc: In member function ‘virtual int KStore::queue_transactions(ObjectStore::Sequencer*, std::vector<ObjectStore::Transaction>&, TrackedOpRef, ThreadPool::TPHandle*)’:
os/kstore/KStore.cc:2393:7: warning: unused variable ‘r’ [-Wunused-variable]
   int r;
       ^
  CXX      osd/libosd_a-ReplicatedBackend.o
os/bluestore/BlueStore.cc: In member function ‘virtual int BlueStore::queue_transactions(ObjectStore::Sequencer*, std::vector<ObjectStore::Transaction>&, TrackedOpRef, ThreadPool::TPHandle*)’:
os/bluestore/BlueStore.cc:4263:7: warning: unused variable ‘r’ [-Wunused-variable]
   int r;
       ^

yangdongsheng:
@liewegas  thanx, it's updated now.

PR #8600, opened 2016-04-14
ceph_test_rados: make long name ~300 chars, (not ~800)

liewegas:

### DIFF
--- a/src/test/osd/TestRados.cc
+++ b/src/test/osd/TestRados.cc
@@ -56,10 +56,7 @@ public:
       oid << m_op;
       if (m_op % 2) {
 	// make it a long name
-	oid << " ";
-	for (unsigned i = 0; i < 300; ++i) {
-	  oid << i;
-	}
+	oid << " " << string(300, 'o');
       }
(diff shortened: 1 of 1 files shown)


src/test/osd/TestRados.cc
```diff
@@ -56,10 +56,7 @@ class WeightedTestGenerator : public TestOpGenerator
       oid << m_op;
       if (m_op % 2) {
 	// make it a long name
-	oid << " ";
-	for (unsigned i = 0; i < 300; ++i) {
-	  oid << i;
-	}
+	oid << " " << string(300, 'o');
```
tchaikov: @liewegas i don't get it, we append 301 spaces to oid. and we append a string of 301 spaces. so nothing changed in this patch. it's but a refactor unlike claimed by the commit message?

liewegas: we were appending "1" "2" "3" "4" ... "299" "300".

tchaikov: oh, right!

tchaikov:
lgtm.

PR #8601, opened 2016-04-14
rgw: aws4: handle UNSIGNED-PAYLOAD under header auth

jmunhoz:
Fixes: http://tracker.ceph.com/issues/15499

Signed-off-by: Javier M. Mellid <jmunhoz@igalia.com>

### DIFF
--- a/src/rgw/rgw_rest_s3.cc
+++ b/src/rgw/rgw_rest_s3.cc
@@ -3523,24 +3523,33 @@ int RGW_Auth_S3::authorize_v4(RGWRados *store, struct req_state *s)
 
   /* handle request payload */
 
-  /* from rfc2616 - 4.3 Message Body
-   *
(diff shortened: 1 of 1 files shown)


jmunhoz:
In the jargon of AWS S3, this patch adds the missing bits to handle the 'Unsigned payload option' in 'Transfer Payload in a Single Chunk' (Authorization Header way)

s3-tests has not coverage for the combo (PUT + UNSIGNED-PAYLOAD + Authorization Header) cause our test cases use boto2 with the 'Signed payload option' mainly. Testing with libcloud requires changing the host/region to speak to RGW and using the verify_hash=False parameter in  'driver.upload_object' method.

One RGW S3 provider in libcloud is almost finished but I am running some testing before asking to go upstream in libcloud for the next days.

yehudasa:
lgtm

PR #8602, opened 2016-04-14
ceph_volume_client: evict client also based on mount path

ajarr:
Evict clients based on not just their auth ID, but also based on the
volume path mounted. This is needed for the Manila use-case, where
the clients using an auth ID are denied further access to a share.

Fixes: http://tracker.ceph.com/issues/15045

Signed-off-by: Ramana Raja <rraja@redhat.com>

### DIFF
--- a/src/pybind/ceph_volume_client.py
+++ b/src/pybind/ceph_volume_client.py
@@ -215,16 +215,22 @@ class CephFSVolumeClient(object):
         self.cluster_name = cluster_name
         self.auth_id = auth_id
 
-    def evict(self, auth_id, timeout=30):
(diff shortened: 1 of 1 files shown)


gregsfortytwo:
Can you describe how you tested this? :)

ajarr:
Tested by https://github.com/ceph/ceph-qa-suite/pull/968

gregsfortytwo:
@jcsp, your review

gregsfortytwo:
Do you guys want a manila or similar label in github? I'm trying to decide if that's too fine-grained, but it's a project with sort of its own timelines so I'm thinking maybe it's good.

PR #8603, opened 2016-04-14
Fixes for python omap method return values.

rmechler:
Signed-off-by: Roland Mechler <rmechler@cisco.com>

### DIFF
--- a/src/pybind/rados/rados.pyx
+++ b/src/pybind/rados/rados.pyx
@@ -2799,7 +2799,9 @@ returned %d, but should return zero on success." % (self.name, ret))
             int _flags = flags
 
         with nogil:
-            rados_write_op_operate(_write_op.write_op, self.io, _oid, &_mtime, _flags)
+            ret = rados_write_op_operate(_write_op.write_op, self.io, _oid, &_mtime, _flags)
+        if (ret != 0):
+            raise make_ex(ret, "Failed to operate write op for oid %s" % oid)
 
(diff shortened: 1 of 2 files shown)


jdurgin:
lgtm, nice catch

src/pybind/rados/rados.pyx
```diff
@@ -2799,7 +2799,9 @@ returned %d, but should return zero on success." % (self.name, ret))
             int _flags = flags
 
         with nogil:
-            rados_write_op_operate(_write_op.write_op, self.io, _oid, &_mtime, _flags)
+            ret = rados_write_op_operate(_write_op.write_op, self.io, _oid, &_mtime, _flags)
+        if (ret != 0):
```
tchaikov: could also remove the parenthesis here.

rmechler: I used parentheses because I copied the style from similar code elsewhere in the file. There seems to be a mix of styles (parentheses vs not) in the file. So, should I (a) leave this as is, (b) remove the parentheses only on the code I've added, or (c) make the style consistent (no parentheses) everywhere in the file?

jdurgin: I'd suggest (b) for this pr. A separate pr removing parens from ifs everywhere in pybind would be great. It's just a historical accident that kept getting copied.

jdurgin:
passed python2 and 3 tests locally

PR #8604, opened 2016-04-14
mds: FSmap decode fix

gregsfortytwo:
I left out a struct_v guard when decoding the ever_enabled_multiple boolean in FSMap. Add it!


liewegas:
:+1: fixed the lab cluster

PR #8605, opened 2016-04-14
rgw: return 503 for requests when master zone is down

theanalyst:
This patch basically tries to return 503 Service Unavailable when `forward_request_to_master` fails during bucket create, we currently return 400 Invalid Argument which is wrong, as this is more of a server side error than a client side improper request. Following are the changes

1.Adding a HTTP Error Code for 503
2.Printing the curl_easy_perform error code too along with status 
3. Bailing out when forward request to master fails and returning a service unavailable

Posting this for review, I'm still confused on 3 on whether  we return 503s for general failures of forward_request_to_master itself or should we make the rest_client handle when process request fails with an -EINVAL and return 503 there

Fixes: http://tracker.ceph.com/issues/15501

### DIFF
--- a/src/rgw/rgw_http_client.cc
+++ b/src/rgw/rgw_http_client.cc
@@ -121,7 +121,7 @@ int RGWHTTPClient::process(const char *method, const char *url)
 
   CURLcode status = curl_easy_perform(curl_handle);
   if (status) {
-    dout(0) << "curl_easy_perform returned error: " << error_buf << dendl;
+    dout(0) << "curl_easy_perform returned status " << status << " error: " << error_buf << dendl;
     ret = -EINVAL;
   }
   curl_easy_getinfo(curl_handle, CURLINFO_RESPONSE_CODE, &http_status);
--- a/src/rgw/rgw_op.cc
+++ b/src/rgw/rgw_op.cc
@@ -1928,8 +1928,9 @@ void RGWCreateBucket::execute()
   if (!store->is_meta_master()) {
     JSONParser jp;
     op_ret = forward_request_to_master(s, NULL, store, in_data, &jp);
-    if (op_ret < 0)
+    if (op_ret < 0) {
       return;
+    }
 
     JSONDecoder::decode_json("entry_point_object_ver", ep_objv, &jp);
     JSONDecoder::decode_json("object_ver", objv, &jp);
--- a/src/rgw/rgw_common.h
+++ b/src/rgw/rgw_common.h
@@ -180,6 +180,7 @@ using ceph::crypto::MD5;
 #define ERR_USER_SUSPENDED       2100
 #define ERR_INTERNAL_ERROR       2200
 #define ERR_NOT_IMPLEMENTED      2201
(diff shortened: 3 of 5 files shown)


theanalyst:
I've pushed another version at https://github.com/theanalyst/ceph/tree/trial2 basically moving the logic back to rest_client instead,

src/rgw/rgw_op.cc
```diff
@@ -1928,8 +1928,14 @@ void RGWCreateBucket::execute()
   if (!store->is_meta_master()) {
     JSONParser jp;
     op_ret = forward_request_to_master(s, NULL, store, in_data, &jp);
-    if (op_ret < 0)
+    if (op_ret < 0) {
+      if (op_ret == -EINVAL){
+	// Taking the easy way out, basically some issue in processing request from our side
+	// return a 5XX instead of an -EINVAL to the end user
+	op_ret = -ERR_SERVICE_UNAVAILABLE;
+      }
```
cbodley: returning 503 seems reasonable to me. but we should probably move this up into `forward_request_to_master()` so that the other callers (`RGWSetBucketVersioning` and `RGWDeleteBucket`) do the same

theanalyst: @cbodley I've pushed another version where I've moved to up forward_request itself and see the value of process(), partly because if the master genuinely returned a -EINVAL, we shouldn't be reraising a 500

theanalyst:
@cbodley moved to `forward_request` and returning only if curl fails with an err, just to make sure that we don;t reraise any genuine `einval` from master zone .. for eg. when bucket location is specified wrongly etc.

src/rgw/rgw_http_client.cc
```diff
@@ -121,7 +121,7 @@ int RGWHTTPClient::process(const char *method, const char *url)
 
   CURLcode status = curl_easy_perform(curl_handle);
   if (status) {
-    dout(0) << "curl_easy_perform returned error: " << error_buf << dendl;
+    dout(0) << "curl_easy_perform returned status " << status << " error: " << error_buf << dendl;
```
cbodley: which CURLcode are you seeing for these failures? can we map that code to `ERR_SERVICE_UNAVAILABLE` here, instead of going through `EINVAL`?

theanalyst: Yeah I'm seeing couldn't connect 7 https://curl.haxx.se/libcurl/c/libcurl-errors.html , we could return here itself, but then since this is commonly used by all other http requests, for eg. keystone connections etc. I don't know the implications fully yet

cbodley: how about this?
```
if (status == CURLE_COULDNT_CONNECT) {
  ret = -ERR_SERVICE_UNAVAILABLE;
} else if (status) {
  ret = -EINVAL;
}
```

theanalyst:
@cbodley @yehudasa what do you think of this?

cbodley:
tested manually by killing the master zone gateway, and verified that bucket creation on secondary zone fails with 503. no regression in s3tests

PR #8606, opened 2016-04-14
rbd: loosen image spec validation requirements

dillaman:


PendingReleaseNotes
```diff
@@ -0,0 +1,9 @@
+v10.2.0
```
jdurgin: at this point you can modify doc/release-notes.rst directly for what's currently the jewel rc section

src/tools/rbd/action/ImportDiff.cc
```diff
@@ -203,7 +203,7 @@ int execute(const po::variables_map &vm) {
   std::string snap_name;
   r = utils::get_pool_image_snapshot_names(
     vm, at::ARGUMENT_MODIFIER_NONE, &arg_index, &pool_name, &image_name,
-    &snap_name, utils::SNAPSHOT_PRESENCE_NONE);
```
jdurgin: this still assumes the image already exists, so should be SPEC_VALIDATION_NONE

dillaman: good catch -- I was in robot mode.

dillaman:
@jdurgin updates pushed

jdurgin:
lgtm, running through cli tests

PR #8607, opened 2016-04-15
msg/async: add missing DelayedDelivery and delay injection

yuyuyu101:


yuyuyu101:
make delay happen within thread to avoid deadlock problem #8471

src/msg/async/AsyncConnection.h
```diff
@@ -124,9 +127,53 @@ class AsyncConnection : public Connection {
...
+    Mutex delay_lock;
+    AsyncMessenger *msgr;
+    EventCenter *center;
+
+   public:
+    explicit DelayedDelivery(AsyncMessenger *omsgr, EventCenter *c)
+      : delay_lock("AsyncConnection::DelayedDelivery::delay_lock"),
+        msgr(omsgr), center(c) { }
+    ~DelayedDelivery() {
+      for (auto &&i : register_time_events)
```
tchaikov: `i` is still a universal reference not a rvalue reference. `auto` is better here. since `i` is just a `uint64_t`.

src/msg/async/AsyncConnection.cc
```diff
@@ -2408,6 +2428,53 @@ void AsyncConnection::handle_ack(uint64_t seq)
...
+  } else {
+    msgr->ms_deliver_dispatch(m);
+  }
+}
+
+class C_flush_messages : public EventCallback {
+  std::deque<std::pair<utime_t, Message*> > delay_queue;
+  AsyncMessenger *msgr;
+ public:
+  C_flush_messages(std::deque<std::pair<utime_t, Message*> > q, AsyncMessenger *m): delay_queue(std::move(q)), msgr(m) {}
```
tchaikov: would be better if the parameter type of `q` is `std::deque<std::pair<utime_t, Message*> >&&`, otherwise it will be copied when being passed in.

src/msg/async/AsyncConnection.h
```diff
@@ -124,9 +127,53 @@ class AsyncConnection : public Connection {
...
+        center->delete_time_event(i);
+      assert(delay_queue.empty());
+    }
+    void do_request(int id);
```
tchaikov: nit, better off marking it with the `override` specifier.

yuyuyu101:
passed a small part of tests(http://pulpito.ceph.com/haomai-2016-04-23_16:08:23-rados:thrash-wip-15372-distro-basic-smithi)

@liewegas for cautious I hope you could take a larger one with this pr...

yuyuyu101:
oh sorry, no.. still has one crash case

yuyuyu101:
@liewegas done, need more large test suite.

liewegas:
<pre>msg/async/Event.cc: In member function ‘int EventCenter::process_time_events()’:
msg/async/Event.cc:296:9: error: ‘cur’ was not declared in this scope
     if (cur >= it->first || clock_skewed) {
         ^
</pre>
needs rebase

yuyuyu101:
done

PR #8611, opened 2016-04-15
Jewel open symlink

ukernel:


liewegas:
retarget at master, we can backport

PR #8612, opened 2016-04-15
osd/ReplicatedPG: fix implementation of register_on_success()

xiexingguo:
We shall push the callbacks into on_success instead, I think.

### DIFF
--- a/src/osd/ReplicatedPG.h
+++ b/src/osd/ReplicatedPG.h
@@ -567,7 +567,7 @@ public:
     }
     template <typename F>
     void register_on_success(F &&f) {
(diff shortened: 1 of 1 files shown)


liewegas:
http://pulpito.ceph.com/sage-2016-04-15_13:04:02-rados-wip-sage-testing2-distro-basic-smithi/

looks good here.. @athanatos what do you think?

athanatos:
sjust@teuthology:/a/samuelj-2016-04-16_14:14:41-rados-wip-sam-testing-distro-basic-smithi/135269$ 
 
Good here too.  The fix looked definitely right, I was mostly worried that we were covering another bug.  Mergin

PR #8613, opened 2016-04-15
osd: some debug output for old osdmap trimming

liewegas:
I'm trying to figure out why osd.114 (among others) on the lab cluster is not trimming old maps.  My guess is a leaked ref that's making the cache lower bound stick...


tchaikov:
hammer is also suffered from this, see http://tracker.ceph.com/issues/13990

i pulled together a branch (https://github.com/ceph/ceph/commits/wip-hammer-13990) to add very verbose log and a script (http://tracker.ceph.com/issues/13990) to process the log. if we are able to reproduce this issue that would be great.

PR #8615, opened 2016-04-15
mailmap: Abhishek Lekshmanan affiliation

theanalyst:
Adding an entry to normalize the personal email id and work email id (As reviews/and commits assoc'ed with gmail still got normalized to previous email)

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -7,8 +7,8 @@
 #
 #
 Aaron Bassett <abassett@gmail.com>
-Abhishek Lekshmanan <abhishek.lekshmanan@ril.com> <abhishek.lekshmanan@gmail.com>
 Abhishek Lekshmanan <abhishek@suse.com> <alekshmanan@suse.com>
+Abhishek Lekshmanan <abhishek@suse.com> <abhishek.lekshmanan@gmail.com>
 Adam C. Emerson <aemerson@redhat.com>
 Adam Twardowski <adam.twardowski@gmail.com>
 Ahoussi Armand <ahoussi.say@telecom-bretagne.eu> <delco225>


.mailmap
```diff
@@ -9,6 +9,7 @@
 Aaron Bassett <abassett@gmail.com>
 Abhishek Lekshmanan <abhishek.lekshmanan@ril.com> <abhishek.lekshmanan@gmail.com>
```
dachary: this one should then be removed, right ?

theanalyst: I'm not sure, without removing it also, if do the credits.sh script from prev release I'm getting the correct output?

dachary: In any case only one of them will be taken into account. There is no way to resolve this so that past contributions stay with the previous mapping. There would need to be a time range for which a mailmap entry is valid, but that's not supported. In other words, unless you change your personal email, contributions from the past will be remapped incorrectly when using a recent .mailmap. There always is the option of using the former .mailmap to get the correct affiliation, which is probably the right thing to do.

theanalyst: Ah ok, what is to be done here then, I'm ok with closing this. not that much of a deal, anyway most of the commits dont use this email id, only the `reviewed-by` since that is added by the dev. doing the merge. (for eg. in 10.1.2 contributor credits it appeared in reviewed-by mostly)

dachary: I think it would make sense to remove the
<pre>
Abhishek Lekshmanan <abhishek.lekshmanan@ril.com> <abhishek.lekshmanan@gmail.com>
</pre>
line.

theanalyst: K will do that

theanalyst:
@dachary updated as you suggested

PR #8616, opened 2016-04-15
[DNM] hammer: build/ops: include rocksdb sources in make dist

asheplyakov:
make dist does not recurse into SUBDIRS if DIST_SUBDIRS are defined.
As a result rocksdb source was missing in the tarball if the source
has been configured --with-librocksdb-static. Unconditionally add
rocksdb to DIST_SUBDIRS to avoid the problem. As a side effect files
rocksdb's configure script, Makefile.in, and various autoconf/libtool
bits are also included in the dist tarball.

Signed-off-by: Alexey Sheplyakov asheplyakov@mirantis.com

### DIFF
--- a/src/Makefile-rocksdb.am
+++ b/src/Makefile-rocksdb.am
@@ -1,3 +1,6 @@
+
+## include rocksdb sources no matter what current build options are
+DIST_SUBDIRS += rocksdb
 if WITH_SLIBROCKSDB
   SUBDIRS += rocksdb
 else


smithfarm:
@asheplyakov Please rebase.

liewegas:
Hrm.. do we really care about rocksdb source for hammer?  It wasn't really used at that point.

liewegas:
As long as the package builds okay for teuthology it ought to be okay for release too

asheplyakov:
@liewegas 
> do we really care about rocksdb source for hammer?

The point is to make release tarballs independent on ./configure options (to not confuse downstream maintainers with unexpectedly large diffs)

smithfarm:
Build fails with: `make[3]: *** No rule to make target `distdir'.  Stop.`

PR #8617, opened 2016-04-15
debian: start ceph-mon-all and ceph-osd-all on package install …

liewegas:


ktdreyer:
Those "ceph-*-all" upstart targets are buggy and I can't wait to drop them (they easily get out of sync with the state of the daemons :(

That said, I guess this is ok with me

liewegas:
:+1:

dmick:
Do we *really* want to start services on installation?  I know traditionally we have done, but it's absolutely forbidden in Fedora, and the reason is justifiable: installing a package should not open ports, I should have to do that consciously.  I know a lot of Debian packages violate this, but I think it's a bad idea overall.

Stopping on package removal makes total sense.

ktdreyer:
I think Debian policy is to start on install, right?

dmick:
I misunderstood what was actually being started; *-all don't actually start daemons, they just need to themselves be in the right state so they can be stopped later

PR #8618, opened 2016-04-15
rgw: file setattr

mattbenjamin:
Posted for review only.  The first 20 commits are already in a PR.  The top commit "works" but may not replicate, among other potential problems.


mattbenjamin:
@liewegas I'm updating this now, need to use a different path to update attrs

src/include/rados/librados.hpp
```diff
@@ -375,6 +375,7 @@ namespace librados
     void zero(uint64_t off, uint64_t len);
     void rmxattr(const char *name);
     void setxattr(const char *name, const bufferlist& bl);
+	void setxattr(const char *name, const buffer::list&& bl);
```
oritwas: nitpick, formatting ...

src/rgw/rgw_file.cc
```diff
@@ -85,18 +90,25 @@ namespace rgw {
...
+     * object locator w/o trailing slash */
+
+    /* mutating path */
+    std::string obj_path{parent->relative_object_name()};
+    if ((obj_path.length() > 0) &&
+	(obj_path.back() != '/'))
+      obj_path += "/";
+    obj_path += path;
+
+    for (auto ix : { 0, 1, 2 }) {
```
oritwas: can we use constants here?

src/rgw/rgw_file.h
```diff
@@ -777,21 +851,23 @@ namespace rgw {
 	  goto retry; /* !LATCHED */
 	}
 	/* LATCHED, LOCKED */
-	if (! (fh->flags & RGWFileHandle::FLAG_LOCK))
+	if (! (flags & RGWFileHandle::FLAG_LOCK))
```
oritwas: why do you need this change? it looks suspicious

src/rgw/rgw_file.cc
```diff
@@ -261,19 +324,45 @@ namespace rgw {
...
+	unref(rgw_fh);
+	get<0>(mkr) = nullptr;
+	return mkr;
       }
 
       string uri = "/" + bname; /* XXX get rid of URI some day soon */
       RGWCreateBucketRequest req(get_context(), get_user(), uri);
+
+      /* save attrs */
+      req.emplace_attr(RGW_ATTR_UNIX_KEY1, std::move(ux_key));
```
oritwas: I think the move is not needed here as emplace_attr uses it.

src/rgw/rgw_file.cc
mattbenjamin: yes, probably should in general

src/rgw/rgw_file.cc
mattbenjamin: I'll look at the params;  it should be harmless to annotate it, though

mattbenjamin:
@oritwas I think this version should be ready to merge, can you revisit it?

I have not replaced the quoted '/' and "/" with a constant in this change to avoid churn, I'll be happy to send a separate change to improve that.

src/include/rados/librados.hpp
oritwas: still not fixed ...

PR #8619, opened 2016-04-15
rgw: fix binfo_cache, RGWShardCollectCR

yehudasa:


mattbenjamin:
does look like a good idea, now that you mention it

PR #8623, opened 2016-04-15
pybind: removed unneccesary parentheses from if statements

rmechler:
Signed-off-by: Roland Mechler <rmechler@cisco.com>


jdurgin:
lgtm, thanks!

PR #8590, opened 2016-04-14
doc/release-notes: jewel updates

liewegas:
Signed-off-by: Sage Weil <sage@redhat.com>

PR #8627, opened 2016-04-16
doc: add missing asterisks in 10.1.2 release notes

smithfarm:
These two missing asterisks caused a formatting melt-down in the list of bullet
points.


smithfarm:
Ah, now I see this is fixed in #8590

PR #8629, opened 2016-04-16
librados: always set the outstrlen for rados_ping_monitor()

tchaikov:
should set outstrlen even if the returned payload is empty. and we do
handle the empty string in do_out_buffer().


tchaikov:
it's not a fix for http://tracker.ceph.com/issues/15454 . as per the mon log, the monitor did send a non-zero payload to the client.

PR #8632, opened 2016-04-17
osd: fix be_deep_scrub() can't catch omap error

xiexingguo:
be_deep_scrub() will scan the omap of the specified object also.
However, this can not work well currently because iter->status()
does not return -EIO on error, it returns -1 instead.

### DIFF
--- a/src/osd/ReplicatedBackend.cc
+++ b/src/osd/ReplicatedBackend.cc
@@ -805,7 +805,8 @@ void ReplicatedBackend::be_deep_scrub(
       poid, ghobject_t::NO_GEN, get_parent()->whoami_shard().shard));
   assert(iter);
   uint64_t keys_scanned = 0;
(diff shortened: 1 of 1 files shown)


src/osd/ReplicatedBackend.cc
```diff
@@ -806,6 +806,13 @@ void ReplicatedBackend::be_deep_scrub(
   assert(iter);
   uint64_t keys_scanned = 0;
   for (iter->seek_to_first(); iter->valid() ; iter->next(false)) {
+    if (iter->status() < 0) {
```
tchaikov: can we change this way?
```c++
for (iter->seek_to_first(); !iter->status() && iter->valid(); iter->next(false)) {
```

and move the check out of the for loop?

xiexingguo:
Done, thanks for the review:)

liewegas:
@tchaikov ping

tchaikov:
lgtm

PR #8635, opened 2016-04-18
cls_rbd: fix loop condition when listing image metadata

runsisi:
if `max_return` is 0 (i.e., the rbd client wants to list all metadata of the
image), then we should not use `max_return` as a part of the loop
condition
and there is also a small typo fix


liewegas:
@dillaman

yuyuyu101:
LGTM

PR #8638, opened 2016-04-18
rgw: fix input param 'last_run' when call "inspect_all_shards" function

weiqiaomiao:
This patch fix for ObjectExpirer Process:
A ObjectExpirer Process Round is start from 'last_run' to 'round_start' ,
if the last process round break in function ```"RGWObjectExpirer::process_single_shard"```
because of time expired when 'truncated' is true(as follow code):
```
void RGWObjectExpirer::process_single_shard(const string& shard,
                                         const utime_t& last_run,
                                         const utime_t& round_start)
{
   ...
   ret = store->objexp_hint_list(shard, rt_last, rt_start,
                                 num_entries, marker, entries,
                                 &out_marker, &truncated);
   ...
      if (now >= end) {
      break;
    }
	...
}
```
there may be some object whose index is in [marker, round_start] are not process,
so the next process round we should start from 'marker' not 'round_start',
but it is not easy to get 'marker' in 'RGWObjectExpirer::OEWorker::entry' fuction
and i think it is ok to start from '0' of every process round becasue we already
delete the object whoes index is in [last_run, marker] in last round

### DIFF
--- a/src/rgw/rgw_object_expirer_core.cc
+++ b/src/rgw/rgw_object_expirer_core.cc
@@ -199,8 +199,6 @@ void RGWObjectExpirer::process_single_shard(const string& shard,
 
 void RGWObjectExpirer::inspect_all_shards(const utime_t& last_run, const utime_t& round_start)
 {
-  utime_t shard_marker;
-
   CephContext *cct = store->ctx();
   int num_shards = cct->_conf->rgw_objexp_hints_num_shards;
 
@@ -245,8 +243,6 @@ void *RGWObjectExpirer::OEWorker::entry() {
     ldout(cct, 2) << "object expiration: start" << dendl;
     oe->inspect_all_shards(last_run, start);
     ldout(cct, 2) << "object expiration: stop" << dendl;
-
-    last_run = start;
 
     if (oe->going_down())
       break;


yehudasa:
@rzarzynski can you take a look at this one?

weiqiaomiao:
retest this please

rzarzynski:
@weiqiaomiao, @yehudasa: generally the idea to start each round from 0 looks OK. I see only one situation where it's better to have the current behavior: if a hint shard is polluted with a lot of entries which cannot be decoded (lines [105-109](https://github.com/ceph/ceph/blob/master/src/rgw/rgw_object_expirer_core.cc#L105-L109)). This is rather a corner case.

In theory we could have an additional output parameter to `RGWObjectExpirer::process_single_shard` to let`RGWObjectExpirer::OEWorker::entry` know whether it should do `last_run = start;` (line [249](https://github.com/ceph/ceph/blob/master/src/rgw/rgw_object_expirer_core.cc#L249)) or not. I could take care of that.

@weiqiaomiao: do we have a ticket in the bug tracker for the issue?

weiqiaomiao:
@rzarzynski there is no issue to trace this, do we need one?

rzarzynski:
@weiqiaomiao: yup, having a ticket makes things easier. This is evidently a bug and we would like to backport a fix to Jewel.

BTW: most likely there is another [issue](http://tracker.ceph.com/issues/16684) in the code. I'm working on that. I can also take care of adding the output parameter to `RGWObjectExpirer::process_single_shard`.

weiqiaomiao:
@rzarzynski  OK. I open an issue in bug tracker and the ticket is: ``` http://tracker.ceph.com/issues/16705```

rzarzynski:
@weiqiaomiao: thanks a lot! I created #10330 that is DNM at the moment. Could you please take a look?

weiqiaomiao:
@rzarzynski  It looks good to me.thanks

rzarzynski:
@weiqiaomiao: Is it to OK to close this PR? I just removed the DNM flag from #10330.

weiqiaomiao:
@rzarzynski ok!

PR #8639, opened 2016-04-18
FileStore: make sure _clone happens in the same collection

mslovy:
Prevent clone object between collections

Signed-off-by: Ning Yao <yaoning@unitedstack.com>

### DIFF
--- a/src/os/filestore/FileStore.cc
+++ b/src/os/filestore/FileStore.cc
@@ -2546,10 +2546,13 @@ void FileStore::_do_transaction(
 
     case Transaction::OP_CLONE:
       {
-        coll_t cid = i.get_cid(op->cid);
+        coll_t cid, ncid;
+        cid = ncid = i.get_cid(op->cid);
         ghobject_t oid = i.get_oid(op->oid);
 	_kludge_temp_object_collection(cid, oid);
         ghobject_t noid = i.get_oid(op->dest_oid);
+        _kludge_temp_object_collection(ncid, noid);
+        assert(cid == ncid);
         tracepoint(objectstore, clone_enter, osr_name);
         r = _clone(cid, oid, noid, spos);
         tracepoint(objectstore, clone_exit, r);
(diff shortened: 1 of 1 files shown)


mslovy:
Since during recovery, collection may become a temp collection, which will violate the logic of _clone and _clone_range in FileStore. And two _kludge_temp_object_collection() may give us different result for collection if we use clone or clone_range in recovery process
Therefore, two suggestions for this case:
1. If clone between collection is not allowed, we need to make sure this case will not happen
2. Otherwise, we may alter the _clone (coll_t cid ...),   _clone_range(coll_t cid ...) to 
_clone (coll_t cid, coll_t ncid, ...),  _clone_range(coll_t cid, coll_t ncid ...)

Normally, we don't want object clone between different collections, so which is preferable?

liewegas:
Clone between *ObjectStore* collections is not allowed, but FileStore has the hacky kludge to split the collection into temp and non-temp.  Clone between those is perfectly fine.  Is there anything broken here, or is this just adding a sanity check?

mslovy:
@liewegas , yeah, I agree. But temp and non-temp are still two different coll_t so that we still need to modify the inferface for  _clone and _clone_range to _clone (coll_t cid, coll_t ncid, ...), _clone_range(coll_t cid, coll_t ncid ...) as we need to provide both temp cid and non-temp cid to the function. And at the same time, we need  sanity check to forbidden passing different collection to _clone, is that OK?

athanatos:
I'm not sure what this PR is attempting to do, so I'm going to close it.

clone is never used between a temp object and a non-temp object (currently).  If we did attempt to clone between a temp object and a normal object, I think the current OP_CLONE case does have a bug.  If the goal is to fix that bug, feel free to reopen with a test of that case included (add it to src/test/objectstore/store_test.cc).

clone_range does clone between a temp object and a non-temp object, so those checks are wrong.  The most you could verify there is that ncid and cid are either the same collection OR a temp collection and actual collection for the same PG, which would amount to a check that _kludge_temp_object_collection is working.  I don't think that would be worthwhile.

If I've missed something, or you disagree, feel free to reopen!

mslovy:
@athanatos， but the current logic in clone_range prevent us cloning object between temp and non-temp since the second _kludge_temp_object_collection always update cid to new value. So I want to point that it is confuesed logic here. If it is allowed so we need to alter the current logic, otherwise we need sanity check to avoid this.  Either is reasonable but not this ambiguous one.   Why I want to avoid this here because if this allowed, we need to alter the interface clone_range in order to pass the old cid and new cid.

liewegas:
temp and not-temp are not real collections.. the 'temp collection' concept 
exists *only* as a hack in FileStore itself to avoid more intrusive 
changes.  in the ObjectStore interface, the temp and non-temp objects 
coexist in one collection.

PR #8641, opened 2016-04-18
mds: MDSDaemon::init bug fix

renhwztetecs:
MDSDaemon::init bug fix

### DIFF
--- a/src/mds/MDSDaemon.cc
+++ b/src/mds/MDSDaemon.cc
@@ -595,7 +595,6 @@ int MDSDaemon::init(MDSMap::DaemonState wanted_state)
 
   // schedule tick
   reset_tick();
-  g_conf->add_observer(this);
   mds_lock.Unlock();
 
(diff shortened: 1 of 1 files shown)


src/mds/MDSDaemon.cc
```diff
@@ -545,7 +545,6 @@ int MDSDaemon::init(MDSMap::DaemonState wanted_state)
   // Set up admin socket before taking mds_lock, so that ordering
   // is consistent (later we take mds_lock within asok callbacks)
   set_up_admin_socket();
-  g_conf->add_observer(this);
```
ukernel: we should remove the later one. see commit e430f8e4

renhwztetecs:
done!
@ukernel 
thanks for your review

jcsp:
@ukernel are you happy with the updated code?

jcsp:

PR #8642, opened 2016-04-18
osd/ReplicatedBackend: fix potential memory leak

xiexingguo:
During ```ReplicatedBackend::objects_read_async()```, we may gather multiple
asynchronous read requests in one single run, and may encounter some error on the
intermediate read. We can't just quit as we may leave the callbacks of the
residual read requests hanging, which turns out to cause memory leaks.

This pr solves the above problem by ignoring any reading error if it is
still tolerable, otherwise we simply crash. Therefore in neither way we will
cause memory leaks.

P.S currently we always fill in an extent of asynchronous read with a callback,
see below(```ReplicatedPG::do_osd_ops()```):
```
 ctx->pending_async_reads.push_back(
   make_pair(
   boost::make_tuple(op.extent.offset, op.extent.length, op.flags),
   make_pair(&osd_op.outdata,
   new FillInVerifyExtent(&op.extent.length, &osd_op.rval,
   &osd_op.outdata, maybe_crc, oi.size, osd,
   soid, op.flags))));
```
or
```
  ctx->pending_async_reads.push_back(
    make_pair(
    boost::make_tuple(op.extent.offset, op.extent.length, op.flags),
    make_pair(&osd_op.outdata, new ToSparseReadResult(osd_op.outdata,
    op.extent.length))));
```
That's why we need this change.

### DIFF
--- a/src/osd/ReplicatedBackend.cc
+++ b/src/osd/ReplicatedBackend.cc
@@ -285,7 +285,7 @@ void ReplicatedBackend::objects_read_async(
   for (list<pair<boost::tuple<uint64_t, uint64_t, uint32_t>,
 		 pair<bufferlist*, Context*> > >::const_iterator i =
 	   to_read.begin();
-       i != to_read.end() && r >= 0;
+       i != to_read.end();
        ++i) {
     int _r = store->read(ch, ghobject_t(hoid), i->first.get<0>(),
 			 i->first.get<1>(), *(i->second.first),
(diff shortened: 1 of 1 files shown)


xiexingguo:
@athanatos Mind having a look at this? Thanks:)

xiexingguo:
@athanatos Ping

athanatos:
Hmm, this is actually dead code.  It was written early in the development process for the ECBackend (so that we could test async reads before the EC code was actually written), but it was never meant to be used outside of testing and the tests that used to use that scaffolding now use the actual EC backend.  There is work going on to add actual async reads for replicated backends, so I'm reluctant to change this code right now since it would cause an unnecessary merge conflict.  I'm going to close this PR.

PR #8645, opened 2016-04-18
test: set a default $CEPH_ROOT env variable

tchaikov:
in cf24535, we use $CEPH_ROOT to specify the $top_srcdir to unify
cmake and autotools, but this breaks ceph-qa-suite/tasks/workunit.py,
as it only clones the necessary qa/workunits directory, and does not
pass $CEPH_ROOT to the test scripts. so we need to set a default
$CEPH_ROOT if it is not set.

### DIFF
--- a/qa/workunits/cephtool/test.sh
+++ b/qa/workunits/cephtool/test.sh
@@ -1,6 +1,6 @@
 #!/bin/bash -x
 
-source $CEPH_ROOT/qa/workunits/ceph-helpers.sh
+source $(dirname $0)/../ceph-helpers.sh
 
 set -e
 set -o functrace
--- a/src/test/cephtool-test-mds.sh
+++ b/src/test/cephtool-test-mds.sh
@@ -15,6 +15,8 @@
 # MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 # GNU Library Public License for more details.
 #
+CEPH_ROOT=${CEPH_ROOT:-..}
+
 CEPH_CLI_TEST_DUP_COMMAND=1 \
 MDS=1 MON=1 OSD=3 CEPH_START='mon osd mds' CEPH_PORT=7200 $CEPH_ROOT/src/test/vstart_wrapper.sh \
     $CEPH_ROOT/qa/workunits/cephtool/test.sh \
--- a/src/test/cephtool-test-mon.sh
+++ b/src/test/cephtool-test-mon.sh
@@ -15,6 +15,8 @@
 # MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 # GNU Library Public License for more details.
 #
+CEPH_ROOT=${CEPH_ROOT:-..}
+
 CEPH_CLI_TEST_DUP_COMMAND=1 \
 # uses CEPH_PORT going from 7202 7203 and 7204 because
 # it starts at 7202 and runs 3 mons (see vstart.sh)
--- a/src/test/cephtool-test-osd.sh
+++ b/src/test/cephtool-test-osd.sh
@@ -15,6 +15,8 @@
 # MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 # GNU Library Public License for more details.
 #
+CEPH_ROOT=${CEPH_ROOT:-..}
+
 CEPH_CLI_TEST_DUP_COMMAND=1 \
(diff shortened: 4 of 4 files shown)


qa/workunits/cephtool/test.sh
```diff
@@ -1,6 +1,11 @@
 #!/bin/bash -x
 
-source $CEPH_ROOT/qa/workunits/ceph-helpers.sh
+if [ -z $CEPH_ROOT ]; then
+    source $(dirname $0)/../ceph-helpers.sh
+else
+    # launched by Makefile, so follow its instruction
+    source $CEPH_ROOT/qa/workunits/ceph-helpers.sh
+fi
```
cbodley: I think that `dirname` should work for both cases. I thought @alimaredia had reverted this bit.

tchaikov:
@cbodley i will close this PR once @alimaredia reverts it.

tchaikov:
changelog

- revert cf24535 instead.

liewegas:
lgtm.. pls merge once you're satisfied

alimaredia:
Looks good to me!

tchaikov:
@cbodley and @alimaredia i reverted the change in the qa/workunits, and keep my change in the test/*.

could you help review and merge it? thanks.

wjwithagen:
@tchaikov  @cbodley 

I do not want to rain on the parade, but there are more variables introduced to fix stuff with the Cmake runs, that are not defaulted.
I ran into atleast one of them: the location where the erasure libs are found.
eg. test/mon/mkfs.sh fails due to CEPH_LIB being empty:
test/mon/mkfs.sh:56: mon_run:  ceph-mon --id a --chdir= --mon-osd-full-ratio=.99 --mon-data-avail-crit=1 --erasure-code-dir= --compression-dir= --mon-data=mkfs/a --log-file=mkfs/a/log --mon-cluster-log-file=mkfs/a/log --run-dir=mkfs/a --pid-file=mkfs/a/pidfile --public-addr 127.0.0.1:7110
2016-04-18 23:18:33.461449 804e17000 -1 load dlopen(/libec_jerasure.so): Cannot open "/libec_jerasure.so"

from:
function mon_mkfs() {
    local fsid=$(uuidgen)

    ceph-mon \
        --id $MON_ID \
        --fsid $fsid \
        --erasure-code-dir=$CEPH_LIB \
        --compression-dir=$CEPH_LIB \
        --mkfs \
        --mon-data=$MON_DIR \
        --mon-initial-members=$MON_ID \
        --mon-host=$CEPH_MON \
        "$@"
}


So I guess a lot more needs to be fixed or reverted. 
Suspicious are the variables set in:
 AddCephTest.cmake:
    CEPH_ROOT=${CMAKE_SOURCE_DIR}
    CEPH_BIN=${CMAKE_RUNTIME_OUTPUT_DIRECTORY}
    CEPH_LIB=${CMAKE_LIBRARY_OUTPUT_DIRECTORY}
    CEPH_BUILD_DIR=${CMAKE_BINARY_DIR}
    LD_LIBRARY_PATH=${CMAKE_BINARY_DIR}/lib
 
So all scripts using these need to be able to use a sort of default, If nothing is set.
that is going to require a lot of grepping to find

tchaikov:
that's true. we now define these env variables in Makefile, if one needs to run a test manually, maybe he/she should use ctest (only applies to cmake) or "make check" instead. i found follow command line at the end of test/mon/mkfs.sh,
```sh
# Local Variables:
# compile-command: "cd ../.. ; make TESTS=test/mon/mkfs.sh check"
# End:
```
it is used by emacs, but we can also use it as an example of how to run a single test. did you get a chance to try it out?

wjwithagen:
That once did work, have not tried it recently, Because it takes you thru the whole evaluation of make dependencies. And if one is trying to fix a bug and running the program takes 2 secs, but evaluating the makestuff takes 10 secs, I get annoyed and want to just run the program/script.
I feels like a very convoluted way of doing things.

But if that is the way to go, I can always whipup a my-ceph-test script that does exactly what it requires to set before running. My main reason to call out was because we lost backwards compatibility without any notice at all, and it smacked me in the face.

wjwithagen:
@alimaredia 
I'm sorry If I sounded bitchy, that should not have been the case.
I actually consider parameterising these paths "a good thing".
So fixing it by reverting is not what I was aiming for, more to augment and keep backwards compatible with prevoious behaviour.

PR #8646, opened 2016-04-18
test/rgw: refactor test_multi.py

cbodley:
Moves common code into `rgw_multi.py` to support adding new tests with different multisite/cluster configurations. Also allows running tests that modify the multisite configuration, like `test_multi_period_incremental_sync`, without having side effects on other tests.

### DIFF
--- a/src/mrgw.sh
+++ b/src/mrgw.sh
@@ -11,7 +11,7 @@ port=$2
 
 shift 2
 
-run_root=$script_root/run/$name
+run_root=run/$name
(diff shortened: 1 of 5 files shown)


cbodley:
needs a rebase, but posted so @zaitcev could have a look

oritwas:
@cbodley , what is the status of this pr?

cbodley:
needs a rebase, but because it moves code from test_multi.py into separate files, i'd want to start over instead of trying to rebase. i'll just close this for the time being

PR #8648, opened 2016-04-18
libcephfs: add ceph_ll_mknod wrapper for ll_mknod

ffilz:
Signed-off-by: Frank S. Filz <ffilzlnx@mindspring.com>


liewegas:
:+1:

ffilz:
What qa is needed here? Ganesha is currently the only code that would be a consumer of this new API?

PR #8649, opened 2016-04-19
test: Fix ceph-objectstore-tool test to run manually from src non-cmake

dzafman:
Fix init-ceph to handle it when CEPH_* values set


dzafman:
@alimaredia Next time add "Reviewed-by: " and your name when you merge.

PR #8652, opened 2016-04-19
cephfs: should close file handler before exit in cephfs.cc

JiYou:
cephfs.cc: before exit the program, should close the file handler.

Signed-off-by: You Ji <youji@ebay.com>


gregsfortytwo:
This looks fine to me at a quick glance and I'll make sure it gets tested, but FYI the cephfs tool is getting deprecated.

JiYou:
@gregsfortytwo thanks.

jcsp:
Merged this because as far as I recall our tests don't cover the (soon-to-be-gone) tool anyway

PR #8655, opened 2016-04-19
rgw: forward delete bucket to master before deleting them in a zone

theanalyst:
- `rgw_http_errors.h`- add an entry for -ENOTEMTPY, we currently return a 500 if forward_request returns a 409 as we dont have an entry for it 
- forward delete bucket requests to master before deleting them locally
 
Fixes: http://tracker.ceph.com/issues/15540


theanalyst:
jenkins: test this please

PR #7683, opened 2016-02-17
[DNM] rgw: rework Swift authenication

rzarzynski:
This PR is a very early proposal of Swift's authenication rework. It pulls commits from PR #7296 and #7666 as dependencies. Still in progress.

PR #7326, opened 2016-01-22
rgw: add support for Swift's account ACLs

rzarzynski:
Depends on: #7296.

PR #7296, opened 2016-01-20
rgw: multiple improvements for Swift API's account-related operations

rzarzynski:
This PR provides following improvements:
* Support for full Swift URL schema  - account name is extracted from URL and may be used to load proper tenant identifier. This provides good basics  for i.a. account ACLs and TempURL requests on buckets in the non-empty namespace.
* Split user identity used for authentication/authorization purposes (```req_state::auth_user```) from account owner. Enables account ACL support.
* Swift's reseller admin feature (admin accounts).

The PR bases on #5872.

PR #8657, opened 2016-04-19
rgw: authentication subsystem rework

rzarzynski:
This pull request consists rework of the authentication infrastructure and its related components: HTTP client, Keystone token cache and the revocation mechanism.

Features the rework brings:
* Separation between authentication identities and user's resources a given `RGWOp` operates on (`RGWUserInfo` and friends; account in the Swift's terminology).
* New framework for authentication backends that provides an uniform interface and decouples performing authentication from applying changes to underlying structures (RADOS store, `req_state` and similar).
* Cross-tenants requests over Swift API altogether with the full Swift's URL schema support.
* Account ACLs of Swift API.
* HTTP Referrer-based ACLs of Swift API.
* Cross-tenant ACLs of Swift API for the Keystone authentication backend.
* Admin access (aka *reseller admin* in Swift).
* Increased awareness of Pete's `rgw_keystone_implicit_tenants` across different auth engines.

Enables:
* Account quota of the Swift API.

Derogates following PRs: #7296, #7326, #7683, #8448.

CC: @yehudasa, @mattbenjamin.


rzarzynski:
@mattbenjamin: I've recently pushed a new version with the cross-tenant ACLs handling simplified through the strategy pattern.

rzarzynski:
Rebased due to a merge conflict. Only https://github.com/ceph/ceph/pull/8657/commits/476646e85697ddae995d7af4adff9854d603d4ef has been altered.

src/rgw/rgw_user.cc
```diff
@@ -270,10 +270,10 @@ int rgw_get_user_info_from_index(RGWRados *store, string& key, rgw_bucket& bucke
 int rgw_get_user_info_by_uid(RGWRados *store,
                              const rgw_user& uid,
                              RGWUserInfo& info,
-                             RGWObjVersionTracker *objv_tracker,
-                             real_time *pmtime,
-                             rgw_cache_entry_info *cache_info,
-                             map<string, bufferlist> *pattrs)
+                             RGWObjVersionTracker * const objv_tracker,
```
mattbenjamin: I think that we should gradually move '*' to the type it modifies, as unlike C, that's the style in Stroustrup.  We seem to follow that model for reference but not pointer types here.  I'm not going to put this remark everywhere ;).

src/rgw/rgw_user.cc
```diff
@@ -270,10 +270,10 @@ int rgw_get_user_info_from_index(RGWRados *store, string& key, rgw_bucket& bucke
...
+                             real_time * const pmtime,
```
mattbenjamin: To the point of the change, I think you may be going overboard with the constness.  I don't have a principled objection to it though, so I don't object for now to any of these.  We can always relax it.  Casey noted here, let's not add const to non-pointer/reference arguments.

src/rgw/rgw_common.h
```diff
@@ -300,7 +300,7 @@ class RGWHTTPArgs
   int parse();
   void append(const string& name, const string& val);
   /** Get the value for a specific argument parameter */
-  string& get(const string& name, bool *exists = NULL);
+  const string& get(const string& name, bool *exists = NULL) const;
```
mattbenjamin: actually agree with this one

src/rgw/rgw_common.h
```diff
@@ -309,22 +309,20 @@ class RGWHTTPArgs
...
-    return (iter != sub_resources.end());
+  bool sub_resource_exists(const char *name) const {
+    return (sub_resources.find(name) != std::end(sub_resources));
   }
   map<string, string>& get_params() {
     return val_map;
   }
   map<string, string>& get_sub_resources() { return sub_resources; }
-  unsigned get_num_params() {
+  unsigned get_num_params() const {
```
mattbenjamin: I don't terribly like 'unsigned' as a type in modern code.  Is that really the map's size_type?

src/rgw/rgw_basic_types.h
```diff
@@ -14,6 +14,10 @@ struct rgw_user {
   rgw_user(const std::string& s) {
     from_str(s);
   }
+  rgw_user(const std::string& tenant, const std::string& id)
```
mattbenjamin: ack

src/rgw/rgw_http_client.cc
```diff
@@ -335,6 +335,39 @@ RGWHTTPClient::~RGWHTTPClient()
 }
 
 
+int RGWHTTPHeadersCollector::receive_header(void * const ptr, const size_t len)
```
mattbenjamin: void* const (I guess)

src/rgw/rgw_http_client.cc
```diff
@@ -335,6 +335,39 @@ RGWHTTPClient::~RGWHTTPClient()
...
+{
+  const std::string header_line(static_cast<const char * const>(ptr), len);
```
mattbenjamin: Why is this decl needed?  Should it be inline?

mattbenjamin: But what I do wonder is, could we use a boost::string_ref or similar here?

src/rgw/rgw_user.cc
rzarzynski: > Casey noted here, let's not add const to non-pointer/reference arguments.

I can fully agree with that in the matter of *declarations*. However, when it comes to *definitions* `const` for arguments still could be very useful - readers can be sure (of course, with precision to `const_cast<>` or other nasty things) that they are dealing only with labels, not parts of a context.

src/rgw/rgw_http_client.h
```diff
@@ -102,6 +102,52 @@ class RGWHTTPClient
...
 };
 
+
+class RGWHTTPHeadersCollector : public RGWHTTPClient {
+public:
+  /* Case insensitive comparator for containers carrying HTTP headers. */
+  struct CILess : public std::binary_function<std::string, std::string, bool> {
+    bool operator()(const std::string& lhs,
+                    const std::string& rhs) const {
+      return ::strcasecmp(lhs.c_str(), rhs.c_str()) < 0 ;
```
mattbenjamin: I find calling out to strcasecmp odd here.  Could we use boost::iequals (boost/algorithm/string) here?

src/rgw/rgw_http_client.h
```diff
@@ -102,6 +102,52 @@ class RGWHTTPClient
...
+    }
+  };
+
+  typedef std::string header_name_t;
+  typedef std::string header_value_t;
+  typedef std::set<header_name_t, CILess> header_spec_t;
```
mattbenjamin: Ok, I like the typdefs, but I find header_spec_t to be an obscure name for a std::set of headers.  What about a...header_set (_t?)?

src/rgw/rgw_http_client.h
rzarzynski: This is has been changed in https://github.com/ceph/ceph/pull/8657/commits/45494d40a7b9c648e817b0e00189e07363a3d6c9 to reuse `ltstr_nocase` of `rgw_string.h`. This comparator is quite popular across the code.

src/rgw/rgw_http_client.h
```diff
@@ -102,6 +102,52 @@ class RGWHTTPClient
...
+
+  RGWHTTPHeadersCollector(CephContext * const cct,
+                          const header_spec_t relevant_headers)
+    : RGWHTTPClient(cct),
+      relevant_headers(relevant_headers) {
+  }
+
+  int receive_header(void *ptr, size_t len) override;
```
mattbenjamin: Why is this pointer to void, again?  I guess it's harmless...

src/rgw/rgw_http_client.cc
rzarzynski: I will take deeper look on that. Thanks for pointing!

src/rgw/rgw_user.cc
```diff
@@ -214,8 +214,12 @@ struct user_info_entry {
 
 static RGWChainedCacheImpl<user_info_entry> uinfo_cache;
 
-int rgw_get_user_info_from_index(RGWRados *store, string& key, rgw_bucket& bucket, RGWUserInfo& info,
-                                 RGWObjVersionTracker *objv_tracker, real_time *pmtime)
+int rgw_get_user_info_from_index(RGWRados * const store,
```
mattbenjamin: As above, I think this is mostly a stylistic change.

src/rgw/rgw_http_client.h
```diff
@@ -148,6 +148,83 @@ class RGWHTTPHeadersCollector : public RGWHTTPClient {
 };
 
 
+#if 1
```
mattbenjamin: Ok, what is broken here?  FIXME sounds like something is...not fixed?  Is it fixed later?

src/rgw/rgw_http_client.h
rzarzynski: Those methods are called by callbacks (like `[simple_]receive_http_header`) we passed to libcurl.

src/rgw/rgw_http_client.cc
```diff
@@ -368,6 +368,53 @@ int RGWHTTPHeadersCollector::receive_header(void * const ptr, const size_t len)
   return 0;
 }
 
+int RGWPostHTTPData::send_data(void* ptr, size_t len)
```
mattbenjamin: This is cleaner.  receive_header is grovelly enough to hide, maybe other refactoring could change that.

src/rgw/rgw_keystone.cc
```diff
@@ -16,7 +16,9 @@
 
 #define dout_subsys ceph_subsys_rgw
 
-int rgw_open_cms_envelope(CephContext * const cct, string& src, string& dst)
+int rgw_open_cms_envelope(CephContext * const cct,
```
mattbenjamin: I don't quite see the point of splitting these across 3 lines.

src/rgw/rgw_http_client.h
rzarzynski: I picked up this name to reflect the intention that the type is solely for passing information about in which headers caller is interested. Using `std::set` is only a private implementation detail.

src/rgw/rgw_keystone.h
```diff
@@ -75,13 +75,13 @@ class KeystoneToken {
...
   KeystoneToken() = default;
-  time_t get_expires() { return token.expires; }
-  string get_domain_id() {return project.domain.id;};
-  string get_domain_name()  {return project.domain.name;};
-  string get_project_id() {return project.id;};
-  string get_project_name() {return project.name;};
-  string get_user_id() {return user.id;};
-  string get_user_name() {return user.name;};
+  time_t get_expires() const { return token.expires; }
+  string get_domain_id() const {return project.domain.id;};
```
mattbenjamin: Should these return std::string (copying)?

src/rgw/rgw_keystone.cc
```diff
@@ -149,10 +149,10 @@ KeystoneApiVersion KeystoneService::get_api_version()
   }
 }
 
-bool KeystoneToken::has_role(const string& r)
+bool KeystoneToken::has_role(const string& r) const
```
mattbenjamin: Again, it's not new code.  Naively, it looks like this interface imposes extra (small) string copies?

src/rgw/rgw_rest_s3.cc
```diff
@@ -2993,10 +2993,11 @@ int RGW_Auth_S3_Keystone_ValidateToken::validate_s3token(
 
   /* check if we have a valid role */
   bool found = false;
-  list<string>::iterator iter;
-  for (iter = roles_list.begin(); iter != roles_list.end(); ++iter) {
-    if ((found=response.has_role(*iter))==true)
+  for (const auto role : accepted_roles) {
```
mattbenjamin: I think this should be const auto& role?

rzarzynski: That's right. Will change.

src/rgw/rgw_rest_s3.h
```diff
@@ -357,7 +357,7 @@ class RGW_Auth_S3_Keystone_ValidateToken : public RGWHTTPClient {
   bufferlist rx_buffer;
   bufferlist tx_buffer;
   bufferlist::iterator tx_buffer_it;
-  list<string> roles_list;
+  vector<string> accepted_roles;
```
mattbenjamin: Not for now, but someday, maybe look at sstring for stuff like this...

src/rgw/rgw_swift.cc
```diff
@@ -335,10 +336,9 @@ int RGWSwift::parse_keystone_token_response(const string& token,
   }
 
   bool found = false;
-  list<string>::iterator iter;
-  for (iter = roles_list.begin(); iter != roles_list.end(); ++iter) {
-    const string& role = *iter;
-    if ((found=t.has_role(role))==true)
+  for (const auto role : accepted_roles) {
```
mattbenjamin: const auto&?

src/rgw/rgw_swift.cc
```diff
@@ -335,10 +336,9 @@ int RGWSwift::parse_keystone_token_response(const string& token,
...
+    if (t.has_role(role) == true) {
```
mattbenjamin: is this the same as "if (t.has_role(role))"?

src/rgw/rgw_keystone.cc
```diff
@@ -14,6 +14,11 @@
 #include "common/ceph_crypto_cms.h"
 #include "common/armor.h"
 
+#if 0
```
mattbenjamin: Not new code, but what is broken here?

src/rgw/rgw_keystone.cc
```diff
@@ -293,3 +298,136 @@ void RGWKeystoneTokenCache::invalidate(const string& token_id)
   tokens_lru.erase(e.lru_iter);
   tokens.erase(iter);
 }
+
+
+#if 0
```
mattbenjamin: Um...lot of #if 0--what up with it?

src/rgw/rgw_http_client.h
rzarzynski: I think this was because `#define dout_subsys ceph_subsys_rgw` appeared in the header due to the code movement. It was correct later, in a commit without *ONLY move (...)*.

src/rgw/rgw_keystone.cc
```diff
@@ -13,11 +13,7 @@
 #include "rgw_keystone.h"
 #include "common/ceph_crypto_cms.h"
 #include "common/armor.h"
-
-#if 0
-// FIXME
-#include "rgw_swift.h"
-#endif
+#include "common/Cond.h"
```
mattbenjamin: Note to self or other: let's shift to std forms when we can.

src/rgw/rgw_http_client.cc
```diff
@@ -69,36 +69,49 @@ struct rgw_http_req_data : public RefCountedObject {
 /*
  * the simple set of callbacks will be called on RGWHTTPClient::process()
  */
```
mattbenjamin: "hermitization" means encapsulation?  to do with the matrices?  to do with thierry l'hermite?

src/rgw/rgw_keystone.cc
rzarzynski: Hmm, I think it's because the code movements have been accomplished in separated commits without any further modifications. Those things should be corrected later in commits without the *rgw: ONLY move (...)* prefix.

src/rgw/rgw_http_client.h
```diff
@@ -36,11 +36,50 @@ class RGWHTTPClient
 
 protected:
   CephContext *cct;
-
   param_vec_t headers;
-  int init_request(const char *method, const char *url, rgw_http_req_data *req_data);
-public:
 
+  int init_request(const char *method,
```
mattbenjamin: seems better, yes

src/rgw/rgw_keystone.cc
```diff
@@ -297,8 +297,8 @@ void RGWKeystoneTokenCache::invalidate(const string& token_id)
 
 int RGWKeystoneTokenCache::RevokeThread::check_revoked()
 {
-  string url;
-  string token;
+  std::string url;
```
mattbenjamin: I prefer this as well!

src/rgw/rgw_http_client.h
```diff
@@ -178,36 +178,45 @@ class RGWHTTPHeadersCollector : public RGWHTTPClient {
 };
 
 
-class RGWPostHTTPData : public RGWHTTPClient {
-  bufferlist *bl;
+class RGWHTTPTransceiver : public RGWHTTPHeadersCollector {
```
mattbenjamin: This still seems like a strange class name.

theanalyst: agree on that

src/rgw/rgw_keystone.cc
```diff
@@ -203,6 +203,13 @@ int KeystoneToken::parse(CephContext * const cct,
   return 0;
 }
```
mattbenjamin: Yes, ok.

src/rgw/rgw_keystone.h
```diff
@@ -95,28 +95,33 @@ class KeystoneToken {
...
-      max(_max) {
+      max(cct->_conf->rgw_keystone_token_cache_size) {
   }
+public:
+  RGWKeystoneTokenCache(const RGWKeystoneTokenCache&) = delete;
+  void operator=(const RGWKeystoneTokenCache&) = delete;
+
+  static RGWKeystoneTokenCache& get_instance();
 
   bool find(const string& token_id, KeystoneToken& token);
```
mattbenjamin: not const? ;)

src/rgw/rgw_keystone.cc
```diff
@@ -152,6 +152,88 @@ KeystoneApiVersion KeystoneService::get_api_version()
...
 
+int RGWSwift::get_keystone_url(CephContext * const cct,
+                               std::string& url)
+{
+  // FIXME: it seems we don't need RGWGetRevokedToken here
+  bufferlist bl;
+  RGWGetRevokedTokens req(cct, &bl);
+
+  url = cct->_conf->rgw_keystone_url;
```
mattbenjamin: This use of url seems suspicious--what's going on here?

src/rgw/rgw_keystone.cc
```diff
@@ -152,6 +152,88 @@ KeystoneApiVersion KeystoneService::get_api_version()
...
+
+  if (!cct->_conf->rgw_keystone_admin_token.empty()) {
+    token = cct->_conf->rgw_keystone_admin_token;
+    return 0;
+  }
+
+  KeystoneToken t;
+
+  /* Try cache first. */
+  if (RGWKeystoneTokenCache::get_instance().find_admin(t)) {
```
mattbenjamin: I guess not const.

src/rgw/rgw_admin.cc
```diff
@@ -171,6 +171,7 @@ void _usage()
   cout << "                             of read, write, readwrite, full\n";
```
mattbenjamin: ok

src/rgw/rgw_auth.cc
```diff
@@ -68,3 +72,153 @@ rgw_auth_transform_old_authinfo(req_state * const s)
...
+  int ret = rgw_store_user_info(store, user_info, nullptr, nullptr,
+                                real_time(), true);
+  if (ret < 0) {
+    ldout(cct, 0) << "ERROR: failed to store new user info: user="
+                  << user_info.user_id << " ret=" << ret << dendl;
+    throw ret;
+  }
+}
+
+/* TODO(rzarzynski): we need to handle display_name changes. */
```
mattbenjamin: What remains to be done here?

src/rgw/rgw_keystone.h
rzarzynski: This method does more than just finding a token ;-). It removes a token from the cache if it's expired.

src/rgw/rgw_keystone.cc
rzarzynski: Yup, no const this time :-) as `find_admin` is a wrapper over `find`.

src/rgw/rgw_auth.h
```diff
@@ -51,4 +58,243 @@ inline ostream& operator<<(ostream& out, const RGWIdentityApplier &id) {
...
+  }
+  /* Make the engines non-copyable and non-moveable due to const-correctness
+   * and aggregating applier factories less costly and error-prone. */
+  RGWAuthEngine(const RGWAuthEngine&) = delete;
+  RGWAuthEngine& operator=(const RGWAuthEngine&) = delete;
+
+public:
+  /* Fast, non-throwing method for screening whether a concrete engine may
+   * be interested in handling a specific request. */
+  virtual bool is_applicable() const noexcept = 0;
```
mattbenjamin: Takes no argument?

mattbenjamin: I haven't seen many code which handles exceptions in radosgw to this point.

src/rgw/rgw_auth.h
```diff
@@ -51,4 +58,243 @@ inline ostream& operator<<(ostream& out, const RGWIdentityApplier &id) {
...
+
+/* TODO: introduce extractors for TokenBased. */
+
+/* Keystone. */
+class RGWKeystoneAuthEngine : public RGWTokenBasedAuthEngine {
+protected:
+  const RGWRemoteAuthApplier::Factory * const apl_factory;
+
+  /* Helper methods. */
+  KeystoneToken decode_pki_token(const std::string token) const;
```
mattbenjamin: Can token be passed by reference?

src/rgw/rgw_auth.h
```diff
@@ -51,4 +58,243 @@ inline ostream& operator<<(ostream& out, const RGWIdentityApplier &id) {
...
+  KeystoneToken get_from_keystone(const std::string token) const;
```
mattbenjamin: ditto

src/rgw/rgw_keystone.cc
rzarzynski: This method exports potentially slightly modified (the trailing slash) value of   `cct->_conf->rgw_keystone_url` to a caller. I doubt that `RGWGetRevokedTokens` alters `cct->_conf`.

src/CMakeLists.txt
```diff
@@ -1118,6 +1118,7 @@ if(${WITH_RADOSGW})
     rgw/rgw_acl.cc
     rgw/rgw_acl_s3.cc
     rgw/rgw_acl_swift.cc
+    rgw/rgw_auth.cc
```
mattbenjamin: Not specific to Swift--do more overrides await implementation?

src/rgw/rgw_swift.cc
```diff
@@ -128,98 +128,6 @@ typedef RGWKeystoneHTTPTransceiver RGWValidateKeystoneToken;
 typedef RGWKeystoneHTTPTransceiver RGWGetKeystoneAdminToken;
 typedef RGWKeystoneHTTPTransceiver RGWGetRevokedTokens;
```
mattbenjamin: ok

src/rgw/rgw_keystone.cc
```diff
@@ -384,10 +384,10 @@ int RGWKeystoneTokenCache::RevokeThread::check_revoked()
   bufferlist bl;
   RGWGetRevokedTokens req(cct, &bl);
```
mattbenjamin: ok

src/rgw/rgw_auth.cc
rzarzynski: We need to make some kind of synchronization mechanism to update our local copy of Keystone-dependent pieces of information like the `display_name`. We never had such machinery. As a consequence, we might use an outdated tenant name (name, not UUID) that was stored in `RGWUserInfo` during  RadosGW's user creation.

src/rgw/rgw_acl.cc
```diff
@@ -44,15 +44,13 @@ void RGWAccessControlList::add_grant(ACLGrant *grant)
...
-  }
-  ldout(cct, 5) << "Permissions for user not found" << dendl;
-  return 0;
+int RGWAccessControlList::get_perm(const RGWIdentityApplier& auth_identity,
+                                   const int perm_mask)
+{
+  ldout(cct, 5) << "Searching permissions for identity=" << auth_identity
+                << " mask=" << perm_mask << dendl;
+
+  return perm_mask & auth_identity.get_perms_from_aclspec(acl_user_map);
```
mattbenjamin: correct ordering?

src/rgw/rgw_acl.cc
```diff
@@ -44,15 +44,13 @@ void RGWAccessControlList::add_grant(ACLGrant *grant)
...
-  ldout(cct, 5) << "Searching permissions for uid=" << id << " mask=" << perm_mask << dendl;
-  map<string, int>::iterator iter = acl_user_map.find(id.to_str());
-  if (iter != acl_user_map.end()) {
-    ldout(cct, 5) << "Found permission: " << iter->second << dendl;
-    return iter->second & perm_mask;
-  }
-  ldout(cct, 5) << "Permissions for user not found" << dendl;
-  return 0;
+int RGWAccessControlList::get_perm(const RGWIdentityApplier& auth_identity,
+                                   const int perm_mask)
```
mattbenjamin: Should perm_masks be of explicit width (e.g., uint32_t)?

src/rgw/rgw_acl.cc
```diff
@@ -66,10 +64,12 @@ int RGWAccessControlList::get_group_perm(ACLGroupTypeEnum group, int perm_mask)
   return 0;
 }
 
-int RGWAccessControlPolicy::get_perm(rgw_user& id, int perm_mask) {
-  int perm = acl.get_perm(id, perm_mask);
+int RGWAccessControlPolicy::get_perm(const RGWIdentityApplier& auth_identity,
+                                     const int perm_mask)
```
mattbenjamin: ditto, will stop mentioning this

src/CMakeLists.txt
rzarzynski: I think so. It would be nice to see S3 auth using that as well. :-)

src/rgw/librgw.cc
```diff
@@ -260,6 +260,8 @@ namespace rgw {
     if (ret < 0) {
```
mattbenjamin: ok

src/rgw/rgw_auth.h
```diff
@@ -0,0 +1,51 @@
+// -*- mode:C++; tab-width:8; c-basic-offset:2; indent-tabs-mode:t -*-
+// vim: ts=8 sw=2 smarttab
+
```
mattbenjamin: looks good

src/rgw/librgw.cc
```diff
@@ -227,6 +228,12 @@ namespace rgw {
       goto done;
     }
 
+    /* FIXME: remove this after switching all handlers to the new authentication
+     * infrastructure. */
+    if (nullptr == s->auth_identity) {
```
mattbenjamin: "if (! s->auth->identity)"?  nullptr on the left makes this harder to read

src/rgw/rgw_common.h
```diff
@@ -1251,6 +1253,12 @@ struct req_state {
 
   RGWUserInfo *user;
 
+  /* Object having the knowledge about an authenticated identity and allowing
+   * to apply it during the authorization phase (verify_permission() methods
+   * of a given RGWOp). Thus, it bounds authentication and authorization steps
+   * through a well-defined interface. For more details, see rgw_auth.h. */
+  std::unique_ptr<RGWIdentityApplier> auth_identity;
```
mattbenjamin: unique_ptr?  we allocate one of these...when?

src/rgw/rgw_auth.cc
```diff
@@ -0,0 +1,68 @@
...
+      return id == acct_id;
+    }
+
+    int get_perm_mask() const {
+      return perm_mask;
+    }
+  };
+
+  return std::unique_ptr<RGWIdentityApplier>(
+        new RGWDummyIdentityApplier(s->cct,
```
mattbenjamin: I don't like having to allocate here.

src/rgw/rgw_auth.cc
```diff
@@ -38,17 +38,7 @@ rgw_auth_transform_old_authinfo(req_state * const s)
     }
```
mattbenjamin: ok

src/rgw/rgw_acl_swift.cc
```diff
@@ -136,3 +140,124 @@ void RGWAccessControlPolicy_SWIFT::to_str(string& read, string& write)
   }
 }
 
+void RGWAccessControlPolicy_SWIFTAcct::add_grants(RGWRados * const store,
+                                                  const list<string>& uids,
+                                                  const int perm)
+{
+  for (const auto uid : uids) {
```
mattbenjamin: const auto& ?

src/rgw/rgw_acl_swift.cc
```diff
@@ -136,3 +140,124 @@ void RGWAccessControlPolicy_SWIFT::to_str(string& read, string& write)
...
+    ACLGrant grant;
+    RGWUserInfo grant_user;
+
+    if (uid_is_public(uid)) {
+      grant.set_group(ACL_GROUP_ALL_USERS, perm);
+      acl.add_grant(&grant);
+    } else  {
+      rgw_user user(uid);
```
mattbenjamin: rgw_user feels like it should be optimized more for cases like this, minor, but churns in all paths.

src/rgw/rgw_acl_swift.cc
```diff
@@ -136,3 +140,124 @@ void RGWAccessControlPolicy_SWIFT::to_str(string& read, string& write)
...
+
+    add_grants(store, readonly, SWIFT_PERM_READ);
+  }
+
+  return true;
+}
+
+void RGWAccessControlPolicy_SWIFTAcct::to_str(std::string& acl_str) const
+{
+  list<string> admin;
```
mattbenjamin: ouch!

src/rgw/rgw_op.cc
```diff
@@ -317,6 +340,8 @@ int rgw_build_bucket_policies(RGWRados* store, struct req_state* s)
   if(s->dialect.compare("s3") == 0) {
     s->bucket_acl = new RGWAccessControlPolicy_S3(s->cct);
   } else if(s->dialect.compare("swift")  == 0) {
+    s->user_acl = std::unique_ptr<RGWAccessControlPolicy>(
```
mattbenjamin: I don't like having to allocate here.

src/rgw/rgw_rest_swift.cc
```diff
@@ -110,6 +111,13 @@ static void dump_account_metadata(struct req_state * const s,
 			  iter->second.c_str());
     }
   }
+
```
mattbenjamin: so this happens only on bad request?

src/rgw/rgw_acl.h
```diff
@@ -48,7 +48,7 @@ class ACLPermission
 public:
   ACLPermission() : flags(0) {}
```
mattbenjamin: ok, cool

src/rgw/rgw_auth.h
```diff
@@ -260,19 +260,23 @@ class RGWTokenBasedAuthEngine : public RGWAuthEngine {
   const std::string token;
 
 public:
+  class Extractor {
+  public:
+    virtual ~Extractor() {};
+    virtual std::string get_token() const = 0;
```
mattbenjamin: Is it necessary to copy out a std::string?

src/rgw/rgw_swift_auth.cc
```diff
@@ -10,15 +10,298 @@
...
 using namespace ceph::crypto;
 
-static int build_token(string& swift_user, string& key, uint64_t nonce,
-		       utime_t& expiration, bufferlist& bl)
+
+/* TempURL: applier */
+void RGWTempURLAuthApplier::modify_request_state(req_state * s) const       /* in/out */
+{
+  bool inline_exists = false;
+  string filename = s->info.args.get("filename");
```
mattbenjamin: string&?

src/rgw/rgw_swift_auth.cc
```diff
@@ -10,15 +10,298 @@
...
+  /* TempURL requires that bucket and object names are specified. */
+  if (bucket_name.empty() || s->object.empty()) {
+    throw -EPERM;
+  }
+
+  /* TempURL case is completely different than the Keystone auth - you may
+   * get account name only through extraction from URL. In turn, knowledge
+   * about account is neccessary to obtain its bucket tenant. Without that,
+   * the access would be limited to accounts with empty tenant. */
+  string bucket_tenant;
```
mattbenjamin: somehow...string&

src/rgw/rgw_swift_auth.cc
```diff
@@ -10,15 +10,298 @@
...
+                    << dendl;
+
+  if (rgw_get_user_info_by_uid(store, bucket_info.owner, owner_info) < 0) {
+    throw -EPERM;
+  }
+}
+
+bool RGWTempURLAuthEngine::is_expired(const std::string& expires) const
+{
+  string err;
```
mattbenjamin: rrrrr

src/rgw/rgw_swift_auth.cc
```diff
@@ -10,15 +10,298 @@
...
+
+  if (expiration <= (uint64_t)now.sec()) {
+    dout(5) << "temp url expired: " << expiration << " <= " << now.sec() << dendl;
+    return true;
+  }
+
+  return false;
+}
+
+std::string extract_swift_subuser(const std::string& swift_user_name) {
```
mattbenjamin: ok...

src/rgw/rgw_swift_auth.cc
```diff
@@ -10,15 +10,298 @@
...
+
+  char dest_str[CEPH_CRYPTO_HMACSHA1_DIGESTSIZE * 2 + 1];
+  buf_to_hex((const unsigned char *)dest, sizeof(dest), dest_str);
+
+  return dest_str;
+}
+
+RGWAuthApplier::aplptr_t RGWTempURLAuthEngine::authenticate() const
+{
+  const string temp_url_sig = s->info.args.get("temp_url_sig");
```
mattbenjamin: const string&?

src/rgw/rgw_swift_auth.cc
```diff
@@ -10,15 +10,298 @@
...
+  if (is_expired(temp_url_expires)) {
+    ldout(cct, 5) << "temp url link expired" << dendl;
+    return nullptr;
+  }
+
+  /* We need to verify two paths because of compliance with Swift, Tempest
+   * and old versions of RadosGW. The second item will have the prefix
+   * of Swift API entry point removed. */
+  const size_t pos = g_conf->rgw_swift_url_prefix.find_last_not_of('/') + 1;
+  const std::vector<std::string> allowed_paths = {
```
mattbenjamin: I found a way to do this with boost::string ref (it's in a DNM PR--maybe have a look at it?)

src/rgw/rgw_swift_auth.cc
```diff
@@ -10,15 +10,298 @@
...
+    s->info.request_uri,
+    s->info.request_uri.substr(pos + 1)
+  };
+
+  /* Account owner calculates the signature also against a HTTP method. */
+  std::vector<std::string> allowed_methods;
```
mattbenjamin: boost::string_ref or similar

src/rgw/rgw_swift_auth.cc
```diff
@@ -10,15 +10,298 @@
...
+  if (strcmp("HEAD", s->info.method) == 0) {
+    /* HEAD requests are specially handled. */
+    allowed_methods = { "HEAD", "GET", "PUT" };
+  } else if (strlen(s->info.method) > 0) {
+    allowed_methods = { s->info.method };
+  }
+
+  /* Need to try each combination of keys, allowed path and methods. */
+  for (const auto kv : owner_info.temp_url_keys) {
```
mattbenjamin: yeah, const auto&

I think look at #8722 (mentioned above)

src/rgw/rgw_swift_auth.cc
```diff
@@ -10,15 +10,298 @@
...
+    const int temp_url_key_num = kv.first;
+    const string& temp_url_key = kv.second;
+
+    if (temp_url_key.empty()) {
+      continue;
+    }
+
+    for (const auto path : allowed_paths) {
```
mattbenjamin: const auto&?  etc

src/rgw/rgw_swift_auth.cc
```diff
@@ -10,15 +10,298 @@
...
+  RGWHTTPHeadersCollector validator(cct, { "X-Auth-Groups", "X-Auth-Ttl" });
+
+  ldout(cct, 10) << "rgw_swift_validate_token url=" << url_buf << dendl;
+
+  int ret = validator.process(url_buf);
+  if (ret < 0) {
+    throw ret;
+  }
+
+  std::string swift_user;
```
mattbenjamin: and below...

src/rgw/rgw_swift_auth.cc
```diff
@@ -10,15 +10,298 @@
 #include "auth/Crypto.h"
```
mattbenjamin: conceptually, looks good

src/rgw/rgw_rest_swift.cc
```diff
@@ -1350,38 +1355,131 @@ RGWOp *RGWHandler_REST_Obj_SWIFT::op_options()
...
+      : store(store),
+        acct_override(acct_override) {
+    }
+
+    aplptr_t create_apl_turl(CephContext * const cct,
+                             const RGWUserInfo& user_info) const override {
+      /* TempURL doesn't need any user account override. It's a Swift-specific
+       * mechanism that requires  account name internally, so there is no
+       * business with delegating the responsibility outside. */
+      return aplptr_t(new RGWTempURLAuthApplier(cct, user_info));
```
mattbenjamin: I wish we could avoid allocating here

src/rgw/rgw_rest_swift.cc
```diff
@@ -1350,38 +1355,131 @@ RGWOp *RGWHandler_REST_Obj_SWIFT::op_options()
...
+
+  /* Auth engines. */
+  RGWTempURLAuthEngine tempurl(s, store,                    &aplfact);
+  RGWSignedTokenAuthEngine rgwtk(s->cct, store, token_extr, &aplfact);
+  RGWKeystoneAuthEngine keystone(s->cct,        token_extr, &aplfact);
+  RGWExternalTokenAuthEngine ext(s->cct, store, token_extr, &aplfact);
+  RGWAnonymousAuthEngine anoneng(s->cct,                    &aplfact);
+
+  /* Pipeline. */
+  const std::vector<const RGWAuthEngine *> engines = {
```
mattbenjamin: This I like.

src/rgw/librgw.cc
rzarzynski: I agree. This just landed on the TODO list.

src/rgw/rgw_auth.h
```diff
@@ -238,6 +238,9 @@ class RGWAuthEngine {
   RGWAuthEngine& operator=(const RGWAuthEngine&) = delete;
 
 public:
+  /* Get name of the auth engine. */
+  virtual std::string get_name() const noexcept = 0;
```
mattbenjamin: Would like an idiom that avoids constructing a std::string here.

src/rgw/rgw_common.h
rzarzynski: During the authentication phase (`RGWHandler_REST::authorize`). As a fallback `rgw_auth_transform_old_authinfo` also may allocate that.

src/rgw/rgw_swift_auth.h
```diff
@@ -72,6 +76,10 @@ class RGWSignedTokenAuthEngine : public RGWTokenBasedAuthEngine {
       apl_factory(apl_factory) {
   }
 
+  std::string get_name() const noexcept override {
+    return "RGWSignedTokenAuthEngine";
```
mattbenjamin: Because this could be static.

src/rgw/rgw_acl_swift.cc
rzarzynski: Sure. TODO. :-)

src/rgw/rgw_auth.cc
```diff
@@ -239,3 +240,197 @@ RGWAuthApplier::aplptr_t RGWAnonymousAuthEngine::authenticate() const
...
+class RGWKeystoneHTTPTransceiver : public RGWHTTPTransceiver {
+public:
+  RGWKeystoneHTTPTransceiver(CephContext * const cct,
+                             bufferlist * const token_body_bl)
+    : RGWHTTPTransceiver(cct, token_body_bl,
+                         cct->_conf->rgw_keystone_verify_ssl,
+                         { "X-Subject-Token" }) {
+  }
+
+  std::string get_subject_token() const {
```
mattbenjamin: Likewise here, commented in the abstract interface.

src/rgw/rgw_auth.cc
```diff
@@ -239,3 +240,197 @@ RGWAuthApplier::aplptr_t RGWAnonymousAuthEngine::authenticate() const
...
+};
+
+typedef RGWKeystoneHTTPTransceiver RGWValidateKeystoneToken;
+typedef RGWKeystoneHTTPTransceiver RGWGetKeystoneAdminToken;
+typedef RGWKeystoneHTTPTransceiver RGWGetRevokedTokens;
+
+
+bool RGWKeystoneAuthEngine::is_applicable() const noexcept
+{
+  if (false == RGWTokenBasedAuthEngine::is_applicable()) {
```
mattbenjamin: Is this the same as "if (! RGWTokenBasedAuthEngine::is_applicable())"?  Having "false" on the left-hand side makes the code harder to read.

src/rgw/rgw_auth.cc
```diff
@@ -239,3 +240,197 @@ RGWAuthApplier::aplptr_t RGWAnonymousAuthEngine::authenticate() const
...
+  bufferlist token_body_bl;
+  int ret = rgw_decode_b64_cms(cct, token, token_body_bl);
+  if (ret < 0) {
+    ldout(cct, 20) << "cannot decode pki token" << dendl;
+    throw ret;
+  } else {
+    ldout(cct, 20) << "successfully decoded pki token" << dendl;
+  }
+
+  KeystoneToken token_body;
```
mattbenjamin: Ok.  We should always get RVO.

src/rgw/rgw_auth.cc
```diff
@@ -239,3 +240,197 @@ RGWAuthApplier::aplptr_t RGWAnonymousAuthEngine::authenticate() const
...
+}
+
+RGWRemoteAuthApplier::AuthInfo
+RGWKeystoneAuthEngine::get_creds_info(const KeystoneToken& token,
+                                      const std::vector<std::string>& admin_roles
+                                    ) const noexcept
+{
+  /* Check whether the user has an admin status. */
+  bool is_admin = false;
+  for (const auto admin_role : admin_roles) {
```
mattbenjamin: const auto&?

src/rgw/rgw_auth.cc
```diff
@@ -239,3 +240,197 @@ RGWAuthApplier::aplptr_t RGWAnonymousAuthEngine::authenticate() const
...
+    if (token.has_role(admin_role)) {
+      is_admin = true;
+      break;
+    }
+  }
+
+  return {
```
mattbenjamin: We should make the constructor explicit here.

src/rgw/rgw_swift_auth.cc
rzarzynski: It can recall the PR. It's good. Maybe we could place in on top of this one. What do you think?

src/rgw/rgw_auth.cc
```diff
@@ -239,3 +240,197 @@ RGWAuthApplier::aplptr_t RGWAnonymousAuthEngine::authenticate() const
...
+  };
+}
+
+RGWAuthApplier::aplptr_t RGWKeystoneAuthEngine::authenticate() const
+{
+  KeystoneToken t;
+
+  /* This will be initialized on the first call to this method. In C++11 it's
+   * also thread-safe. */
+  static struct RolesCacher {
```
mattbenjamin: ok

src/rgw/rgw_auth.cc
```diff
@@ -239,3 +240,197 @@ RGWAuthApplier::aplptr_t RGWAnonymousAuthEngine::authenticate() const
...
+      plain.insert(std::end(plain), std::begin(admin), std::end(admin));
+    }
+
+    std::vector<std::string> plain;
+    std::vector<std::string> admin;
+  } roles(cct);
+
+  /* Token ID is a concept that makes dealing with PKI tokens more effective.
+   * Instead of storing several kilobytes, a short hash can be burried. */
+  const auto token_id = rgw_get_token_id(token);
```
mattbenjamin: auto&?

src/rgw/rgw_auth.cc
```diff
@@ -239,3 +240,197 @@ RGWAuthApplier::aplptr_t RGWAnonymousAuthEngine::authenticate() const
...
+  /* Verify expiration. */
+  if (t.expired()) {
+    ldout(cct, 0) << "got expired token: " << t.get_project_name()
+                  << ":" << t.get_user_name()
+                  << " expired: " << t.get_expires() << dendl;
+    return nullptr;
+  }
+
+  /* Check for necessary roles. */
+  for (const auto role : roles.plain) {
```
mattbenjamin: const auto&?

src/rgw/Makefile.am
```diff
@@ -83,7 +83,6 @@ librgw_la_SOURCES = \
 	rgw/rgw_rest_usage.cc \
```
mattbenjamin: CMAKE:  This is good, but the file is still mentioned in CMakeLists.txt

src/rgw/rgw_auth.h
```diff
@@ -5,6 +5,7 @@
 #ifndef CEPH_RGW_AUTH_H
 #define CEPH_RGW_AUTH_H
 
+#include <functional>
```
mattbenjamin: ok

src/rgw/rgw_op.cc
```diff
@@ -124,7 +124,9 @@ static int parse_range(const char *range, off_t& ofs, off_t& end, bool *partial_
   return r;
 }
```
mattbenjamin: I guess this is an improvement.  Again, this isn't something you want to do in headers, right?

src/rgw/rgw_op.cc
```diff
@@ -2875,6 +2875,12 @@ void RGWPutMetadataAccount::execute()
   rgw_get_user_info_by_uid(store, s->user->user_id, orig_uinfo, &acct_op_tracker);
```
mattbenjamin: ok

src/rgw/rgw_acl_swift.h
```diff
@@ -11,17 +11,19 @@
 
 #include "rgw_acl.h"
 
-using namespace std;
```
mattbenjamin: THANK YOU

src/rgw/rgw_acl.cc
```diff
@@ -22,6 +22,9 @@ void RGWAccessControlList::_add_grant(ACLGrant *grant)
   ACLPermission& perm = grant->get_permission();
```
mattbenjamin: looks good

src/rgw/rgw_common.cc
```diff
@@ -858,6 +858,26 @@ string RGWHTTPArgs::sys_get(const string& name, bool * const exists)
   return e ? iter->second : string();
 }
```
mattbenjamin: looks good

src/rgw/rgw_acl_swift.cc
```diff
@@ -22,7 +22,7 @@ using namespace std;
 
 #define SWIFT_GROUP_ALL_USERS ".r:*"
 
-static int parse_list(string& uid_list, list<string>& uids)
+static int parse_list(const string& uid_list, list<string>& uids)
 {
   char *s = strdup(uid_list.c_str());
```
mattbenjamin: Hmm.  I know, not new code.

src/rgw/rgw_acl_swift.cc
```diff
@@ -62,13 +62,13 @@ static bool uid_is_public(const string& uid)
...
 }
 
-void RGWAccessControlPolicy_SWIFT::add_grants(RGWRados *store, list<string>& uids, int perm)
+void RGWAccessControlPolicy_SWIFT::add_grants(RGWRados * const store,
+                                              const list<string>& uids,
+                                              const int perm)
 {
-  list<string>::iterator iter;
-  for (iter = uids.begin(); iter != uids.end(); ++iter ) {
+  for (const auto& uid : uids) {
```
mattbenjamin: yay

src/rgw/rgw_acl_swift.cc
```diff
@@ -194,12 +194,14 @@ void RGWAccessControlPolicy_SWIFT::to_str(string& read, string& write)
       id = SWIFT_GROUP_ALL_USERS;
     }
```
mattbenjamin: ok

src/rgw/rgw_acl.h
```diff
@@ -184,7 +184,7 @@ class ACLGrant
```
mattbenjamin: ok

src/rgw/rgw_user.cc
```diff
@@ -40,11 +40,6 @@ void rgw_get_anon_user(RGWUserInfo& info)
   info.access_keys.clear();
 }
```
mattbenjamin: ok

src/rgw/rgw_auth.cc
rzarzynski: What is your suggestion? I see this function as a factory. IMO there is no business in having anything else than `nullptr` in `req_state::auth_identity` before the auth phase.

src/rgw/rgw_auth.cc
```diff
@@ -53,6 +53,16 @@ rgw_auth_transform_old_authinfo(req_state * const s)
     int get_perm_mask() const {
       return perm_mask;
     }
+
+    std::string to_str() const {
```
mattbenjamin: I think it would be better to avoid stringstream in anything like a common path, at least if we don't rely on our own stream operator overrides.  thoughts?

src/rgw/rgw_auth.cc
```diff
@@ -206,6 +228,19 @@ bool RGWLocalAuthApplier::is_owner_of(const rgw_user& uid) const
   return uid == user_info.user_id;
 }
```
mattbenjamin: ditto

src/rgw/rgw_auth.h
```diff
@@ -49,11 +49,13 @@ class RGWIdentityApplier {
...
     return is_owner_of(rgw_user(RGW_USER_ANON_ID));
   }
+
+  virtual std::string to_str() const = 0;
 };
 
-inline ostream& operator<<(ostream& out, const RGWIdentityApplier &id) {
-//  return out << id.to_str();
-  return out;
+inline std::ostream& operator<<(std::ostream& out,
```
mattbenjamin: and ok, here's a new one--do we really need this idiom?

src/rgw/rgw_auth_decoimpl.h
```diff
@@ -110,6 +119,13 @@ template <typename T>
 const rgw_user RGWThirdPartyAccountAuthApplier<T>::UNKNOWN_ACCT;
 
 template <typename T>
+std::string RGWThirdPartyAccountAuthApplier<T>::to_str() const
+{
```
mattbenjamin: Is this idiom also something we do frequently?

src/rgw/rgw_common.cc
```diff
@@ -183,7 +183,7 @@ void req_info::rebuild_from(req_info& src)
```
mattbenjamin: lgtm

src/rgw/rgw_auth.cc
```diff
@@ -399,6 +399,51 @@ RGWKeystoneAuthEngine::get_creds_info(const KeystoneToken& token,
...
+                                               const std::string id)
+{
+  return tenant + ":" + id;
+}
+
+RGWKeystoneAuthEngine::acl_strategy_t
+RGWKeystoneAuthEngine::get_acl_strategy(const KeystoneToken& token) const
+{
+  /* The primary identity is constructed upon UUIDs. */
+  const auto tenant_uuid = token.get_project_id();
```
mattbenjamin: const auto&?

(and ff.)

src/rgw/rgw_auth.cc
```diff
@@ -399,6 +399,51 @@ RGWKeystoneAuthEngine::get_creds_info(const KeystoneToken& token,
...
+  const auto user_uuid = token.get_user_id();
+
+  /* For Keystone v2 an alias may be also used. */
+  const auto tenant_name = token.get_project_name();
+  const auto user_name = token.get_user_name();
+
+  /* Construct all possible combinations including Swift's wildcards. */
+  const std::vector<std::string> allowed_items = {
```
mattbenjamin: Neat idiom, maybe we could look at small_vector or something here shortly...

src/rgw/rgw_swift.cc
rzarzynski: Sure, TODO.

src/rgw/rgw_swift.cc
rzarzynski: I think so.

src/rgw/rgw_acl.cc
```diff
@@ -66,36 +70,73 @@ int RGWAccessControlList::get_group_perm(ACLGroupTypeEnum group, int perm_mask)
...
+int RGWAccessControlList::get_referer_perm(const std::string http_referer,
+                                           const int perm_mask)
+{
+  ldout(cct, 5) << "Searching permissions for referer=" << http_referer
+                << " mask=" << perm_mask << dendl;
+
+  /* FIXME: C++11 doesn't have std::rbegin nor std::rend. We would like to
+   * switch when C++14 becomes available. */
+  const auto iter = std::find_if(referer_list.crbegin(), referer_list.crend(),
+    [http_referer](const ACLReferer& r) -> bool {
```
cbodley: capture http_referrer by reference here to avoid copying it in

src/rgw/rgw_http_client.h
rzarzynski: Yes, I know this naming is awkward. Unfortunately, I cannot find anything better. :-(

Guys, do have an idea? This class basically sends a request using an HTTP method specified by caller (so, `RGWPostHTTPData` doesn't fit IMO) and collects response.

src/rgw/rgw_auth.h
rzarzynski: All necessary info should be taken and stored by a constructor of a derived class to guarantee that all methods will work on the same context.

src/rgw/rgw_auth.h
rzarzynski: I think so. TODO.

src/rgw/rgw_acl.cc
rzarzynski: Good point. ACLs should use `uint32_t` as `req_state` already does.

src/rgw/rgw_swift_auth.cc
rzarzynski: TODO..

src/rgw/rgw_auth.h
rzarzynski: Hmm, maybe good old `const char *` would work here. `std::string` looks like an overkill.

src/rgw/rgw_rest_swift.cc
rzarzynski: I think `std::vector` might be replaced with `boost::static_vector` or similar. Is it available for us (taking into consideration boost version and platforms we do support)?

src/rgw/rgw_swift_auth.h
rzarzynski: True. TODO.

src/rgw/rgw_auth.cc
rzarzynski: It's the same. TODO.

src/rgw/rgw_auth.h
rzarzynski: It's useful for printing in logs in an uniform manner.

src/rgw/rgw_acl.cc
rzarzynski: That's right, TODO.

src/rgw/rgw_rest_swift.cc
cbodley: `std::array` should work here, right?

rzarzynski: That's right! Thanks for pointing this. TODO.

src/rgw/rgw_auth.cc
rzarzynski: TODO.

src/rgw/rgw_auth.cc
rzarzynski: TODO.

src/rgw/rgw_auth.cc
rzarzynski: TODO.

src/rgw/rgw_swift_auth.cc
rzarzynski: DONE.

src/rgw/rgw_swift_auth.cc
rzarzynski: DONE. I've ported your commit. I hope you don't mind. Authorship and `Signed-off-by` tag have been preserved, of course.

src/rgw/rgw_swift_auth.cc
rzarzynski: DONE.

src/rgw/rgw_swift_auth.cc
rzarzynski: DONE.

src/rgw/rgw_swift_auth.cc
rzarzynski: DONE.

src/rgw/rgw_http_client.cc
rzarzynski: DONE. No allocations now thanks to the `boost::string_ref`.

src/rgw/rgw_keystone.cc
rzarzynski: The line, after applying the `const` modifier, would fairly excess the 80-chars limit.

src/rgw/rgw_rest_s3.cc
rzarzynski: DONE.

src/rgw/rgw_swift.cc
rzarzynski: DONE.

src/rgw/rgw_auth.h
rzarzynski: DONE.

src/rgw/rgw_acl.cc
rzarzynski: DONE.

src/rgw/librgw.cc
rzarzynski: DONE.

src/rgw/rgw_acl_swift.cc
rzarzynski: DONE.

src/rgw/rgw_acl_swift.cc
rzarzynski: DONE.

src/rgw/rgw_auth.h
rzarzynski: DONE.

src/rgw/rgw_acl.cc
rzarzynski: DONE.

src/rgw/rgw_auth.cc
rzarzynski: DONE with `std::array`.

src/rgw/rgw_auth.cc
rzarzynski: DONE.

src/rgw/rgw_auth_decoimpl.h
rzarzynski: Only in printing with high verbosity level.

src/rgw/Makefile.am
rzarzynski: DONE.

src/rgw/rgw_auth.cc
rzarzynski: DONE.

src/rgw/rgw_auth.cc
rzarzynski: DONE.

src/rgw/rgw_auth.cc
rzarzynski: DONE.

src/rgw/rgw_auth.cc
rzarzynski: DONE.

src/rgw/rgw_auth.cc
rzarzynski: DONE.

src/rgw/rgw_swift_auth.h
rzarzynski: DONE.

src/rgw/rgw_auth.h
rzarzynski: DONE.

rzarzynski:
@mattbenjamin, @cbodley: I've pushed the rebased version. GitHub apparently hasn't refreshed the page yet. The latest commit should be `ca88e99618772c512af4e89fd5a3be14885b3cb7`.

The previous version (before the rebase) of this branch [is available](https://github.com/rzarzynski/ceph/commits/wip-rgw-swift-auth-old-20160523) as well. It has all the commits I added to address your comments and currently may be viewed via GitHub. :-)

src/rgw/rgw_auth.h
```diff
@@ -0,0 +1,340 @@
...
+    return is_owner_of(rgw_user(RGW_USER_ANON_ID));
+  }
+
+  virtual std::string to_str() const = 0;
+};
+
+inline std::ostream& operator<<(std::ostream& out,
+                                const RGWIdentityApplier &id) {
+  return out << id.to_str();
+}
```
cbodley: the to_str() interface imposes quite a bit of overhead by requiring the implementations to create a stringstream and convert it to string, just to send it back into an ostream

the MDS provides a good example of polymorphic stream output operators:
```c++
class LogEvent {
...
  virtual void print(std::ostream& out) = 0;
};
inline ostream& operator<<(ostream& out, LogEvent& le) {
  le.print(out);
  return out;
}
```
```c++
class EOpen : public LogEvent {
...
  void print(ostream& out) const {
    out << "EOpen " << metablob << ", " << inos.size() << " open files";
  }
};
```

src/rgw/rgw_auth.h
```diff
@@ -0,0 +1,340 @@
...
+    friend class RGWRemoteAuthApplier;
+  protected:
+    const rgw_user acct_user;
+    const std::string acct_name;
+    const uint32_t perm_mask;
+    const bool is_admin;
+
+  public:
+    AuthInfo(const rgw_user acct_user,
+             const std::string acct_name,
```
cbodley: can pass acct_user and acct_name by const reference

src/rgw/rgw_auth.h
```diff
@@ -0,0 +1,340 @@
...
+  const AuthInfo info;
+
+  virtual void create_account(const rgw_user acct_user,
+                              RGWUserInfo& user_info) const;          /* out */
+
+public:
+  RGWRemoteAuthApplier(CephContext * const cct,
+                       RGWRados * const store,
+                       acl_strategy_t&& extra_acl_strategy,
+                       const AuthInfo info)
```
cbodley: can `create_account()` take `acct_user` by const reference?

can this ctor take `info` by const reference?

src/rgw/rgw_auth.h
```diff
@@ -0,0 +1,340 @@
...
+      extra_acl_strategy(std::move(extra_acl_strategy)),
+      info(info) {
+  }
+
+  virtual uint32_t get_perms_from_aclspec(const aclspec_t& aclspec) const override;
+  virtual bool is_admin_of(const rgw_user& uid) const override;
+  virtual bool is_owner_of(const rgw_user& uid) const override;
+  virtual uint32_t get_perm_mask() const { return info.perm_mask; }
+  virtual std::string to_str() const override;
+  virtual void load_acct_info(RGWUserInfo& user_info) const override; /* out */
```
cbodley: `override` is enough to document that these are virtual functions, so the `virtual` part isn't necessary

src/rgw/rgw_swift_auth.h
```diff
@@ -6,10 +6,123 @@
...
+class RGWXAuthTokenExtractor : public RGWTokenBasedAuthEngine::Extractor {
+protected:
+  const req_state * const s;
+public:
+  RGWXAuthTokenExtractor(const req_state * const s)
+    : s(s) {
+  }
+  std::string get_token() const override {
+    /* Returning a reference here would end in GCC complaining about a reference
+     * to temporary. */
```
cbodley: does it still complain if it returns a const reference?

rzarzynski: Yup, unfortunately.

src/rgw/rgw_auth.h
rzarzynski: That's right. DONE.

src/rgw/rgw_auth.h
rzarzynski: DONE. Switched to the reference type also in the ctor.

src/rgw/rgw_auth.h
rzarzynski: DONE.

src/rgw/rgw_auth.h
rzarzynski: DONE.
Thanks for pointing this out and for snippets! :-)

src/rgw/rgw_keystone.h
rzarzynski: DONE.

src/rgw/rgw_http_client.cc
rzarzynski: DONE.

src/rgw/rgw_auth.cc
rzarzynski: DONE.

src/rgw/rgw_auth.cc
rzarzynski: DONE.

src/rgw/rgw_auth.cc
```diff
@@ -418,7 +420,7 @@ RGWKeystoneAuthEngine::get_acl_strategy(const KeystoneToken& token) const
   const auto& user_name = token.get_user_name();
 
   /* Construct all possible combinations including Swift's wildcards. */
-  const std::vector<std::string> allowed_items = {
+  const std::array<std::string, 6> allowed_items = {
```
mattbenjamin: yee-ha

src/rgw/rgw_rest_swift.cc
```diff
@@ -1482,7 +1483,8 @@ int RGWHandler_REST_SWIFT::authorize()
   RGWAnonymousAuthEngine anoneng(s->cct,                    &aplfact);
 
   /* Pipeline. */
-  const std::vector<const RGWAuthEngine *> engines = {
+  constexpr size_t ENGINES_NUM = 5;
+  const std::array<const RGWAuthEngine *, ENGINES_NUM> engines = {
```
mattbenjamin: yeah, very nice

src/rgw/rgw_swift_auth.cc
```diff
@@ -10,15 +15,330 @@
...
+    for (const auto& path : allowed_paths) {
+      for (const auto& method : allowed_methods) {
+        const char* const local_sig = sig_helper.calc(temp_url_key, method,
+                                                      path, temp_url_expires);
+
+        ldout(s->cct, 20) << "temp url signature [" << temp_url_key_num
+                          << "] (calculated): " << local_sig
+                          << dendl;
+
+        if (sig_helper.is_equal_to(temp_url_sig)) {
```
rzarzynski: FIXED.

mattbenjamin:
ok, if this builds for me, I'm going to set the needs-qa tag

mattbenjamin:
@yuriw  @liewegas ok, we're passing this off for a qa run

rzarzynski:
@mattbenjamin, @cbodley

I've rebased due to a merge conflict we got after merging PR #8186. Only https://github.com/ceph/ceph/pull/8657/commits/2136bbcefe9c541c8e74ad5c7ceb6b880b99a0dc has been altered. Absolutely no other changes.

The previous version of this branch [is available](https://github.com/rzarzynski/ceph/tree/wip-rgw-swift-auth-old-20160525) for comparison.

rzarzynski:
@mattbenjamin, @cbodley: just pushed a patch for using `std::vector` instead of `boost::container::static_vector`. We can revert it when the container will become available on all platforms.

No other changes.

rzarzynski:
@mattbenjamin, @cbodley 

We've got a tiny merge conflict (a new line in https://github.com/ceph/ceph/pull/8657/commits/52b1a57786e7a5cdbc53588ecdf6c20d262a35d7). Rebased. No other changes.

Previous branch if available for comparison: [wip-rgw-swift-auth-old-20160526](https://github.com/rzarzynski/ceph/tree/wip-rgw-swift-auth-old-20160526).

rzarzynski:
@mattbenjamin: I'm rebasing this branch right now.

mattbenjamin:
@rzarzynski  waiting for new push

rzarzynski:
@mattbenjamin: rebased & fixed.

mattbenjamin:
@rzarzynski thanks!

PR #8658, opened 2016-04-19
cmake: Added CMAKE_CEPH_*_VIRTUALENV var to fix build

alimaredia:


alimaredia:
@cbodley if this looks good could you merge it? The cmake build wasn't working

PR #8661, opened 2016-04-20
librbd: put the validation of image snap context earlier

runsisi:
we'd better validate the snap context right after we got the mutable
metadata before we could go any further


dillaman:
lgtm

PR #8663, opened 2016-04-20
BlueStore.cc _do_read fiemap

cooljiansir:
```
    if (op != oend && op->first + op->second.length < offset) {
      ++op;
      continue;
    }
    if (bp != bend && bp->first + bp->second.length <= offset) {
      ++bp;
      continue;
    }
```

### DIFF
--- a/src/os/bluestore/BlueStore.cc
+++ b/src/os/bluestore/BlueStore.cc
@@ -2655,7 +2655,7 @@ int BlueStore::_do_read(
       ++op;
       continue;
     }
-    if (bp != bend && bp->first + bp->second.length <= offset) {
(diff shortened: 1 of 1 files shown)


src/os/bluestore/BlueStore.cc
```diff
@@ -2808,7 +2808,7 @@ int BlueStore::fiemap(
       ++op;
       continue;
     }
-    if (bp != bend && bp->first + bp->second.length <= offset) {
```
yuyuyu101: this should be fine previously?

yuyuyu101:
I dont think original version is wrong, plz correct if I'm stuck into wrong loop

liewegas:
Yeah, original version looks right.  Previous extent might end at our offset.

PR #8666, opened 2016-04-20
rgw: fix handling issues with memory allocation in RGWRados

rzarzynski:


rzarzynski:
@yehudasa: both Tempest and s3-tests haven't found any regression here.

cbodley:
I would prefer not to check for allocation failures manually. Instead, we should aim to be exception-safe and use RAII for cleanup.

In https://github.com/ceph/ceph/commit/685ec652abeb73feb2d27ea3d12ee58fbaef14a0, I took a stab at cleaning up `RGWRados::init_rados()` by using a `std::vector<librados::Rados>` instead of an array of pointers. For exception safety, it initializes the vector on its stack, then swaps it into `RGWRados::rados` on success.

What do you think, @rzarzynski?

cbodley:
@rzarzynski I finally turned that branch into a pr at https://github.com/ceph/ceph/pull/9295, could you take a look?

rzarzynski:
Closing this PR. I think we should go with the #9295.

PR #8670, opened 2016-04-21
rgw: fix memory leak in func 'RGWCreateBucket_ObjStore_S3::get_params'

weiqiaomiao:

### DIFF
--- a/src/rgw/rgw_rest_s3.cc
+++ b/src/rgw/rgw_rest_s3.cc
@@ -930,6 +930,8 @@ int RGWCreateBucket_ObjStore_S3::get_params()
   op_ret = rgw_rest_read_all_input(s, &data, &len, CREATE_BUCKET_MAX_REQ_LEN);
   if ((op_ret < 0) && (op_ret != -ERR_LENGTH_REQUIRED))
     return op_ret;
+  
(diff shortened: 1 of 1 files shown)


oritwas:
How about using std::unique_ptr?

cbodley:
`std::unique_ptr` can be used with pointers that need to be `free()`d, but the syntax is kind of messy:
```c++
  op_ret = rgw_rest_read_all_input(s, &data, &len, CREATE_BUCKET_MAX_REQ_LEN);
  if ((op_ret < 0) && (op_ret != -ERR_LENGTH_REQUIRED))
    return op_ret;

  auto data_deleter = std::unique_ptr<char, decltype(free)*>{data, free};
```
..and then remove any other calls to `free(data)`.

ideally, we would rewrite the c-style interface of `rgw_rest_read_all_input()` to return data as a `std::string`, but that's probably overkill for now

weiqiaomiao:
many thanks for your, I already fixed it ,would you review it again @cbodley

cbodley:
this triggered http://tracker.ceph.com/issues/15595. this commit wasn't actually reverted, but https://github.com/ceph/ceph/pull/8762 was merged as a fix

PR #8672, opened 2016-04-21
rgw: Add access to multi-tenancy test

zaitcev:
This commit adds a code to existing multi-tenancy test, which launches and accesses Gateway, so we can test the tenantized users we created: create some buckets, store some keys/objects.

### DIFF
--- a/src/test/rgw/test_multen.py
+++ b/src/test/rgw/test_multen.py
@@ -3,7 +3,12 @@
 import json
 import sys
 
+from StringIO import StringIO
+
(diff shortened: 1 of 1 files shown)


zaitcev:
retest this please

yehudasa:
@zaitcev remind me, can this one be merged?

zaitcev:
The main problem with the Swift side is not that one final test is missing, but that it imports the swiftclient package. I haven't written a stub yet. This would be a massive new dependency as-is. I'll have to roll this back to S3-only. Unfortunately, they were squashed. Just give me a bit to split and re-test.

PR #8675, opened 2016-04-21
xio: add MNop.h to dist tarball

roidayan:
This file is needed when we want to build a package with xio messenger
but was not added to dist tarball.


maishi:
I git clone the source code,and execute ./genauto.sh && make .then it stop here "CXX      test/encoding/ceph_dencoder-ceph_dencoder.o" .who can help me??

mattbenjamin:
@liewegas for as long as we care about it, can this be incorrect?

PR #8676, opened 2016-04-21
librbd: fix potential double free of SetSnapRequest instance

runsisi:
if image feature `EXCLUSIVE_LOCK` is not enabled we should not try to
initialize the exclusive lock, or we may end with two async Contexts
to finish the same `SetSnapRequest` instance


xiexingguo:
:+1:  Good catch

dillaman:
@runsisi looks good to me ... can I bug you to update the commit message to include:

> Fixes: http://tracker.ceph.com/issues/15571

I opened a new tracker ticket to help with backport (to Jewel) tracking.

runsisi:
hi @dillaman , updated the commit message, thanks for your review :)

PR #8679, opened 2016-04-21
rgw: Have a flavor of bucket deletion to bypass GC.

prallabh:
And to possibly speed up the object deletions by making them async

Fixes: http://tracker.ceph.com/issues/15557

### DIFF
--- a/src/rgw/rgw_admin.cc
+++ b/src/rgw/rgw_admin.cc
@@ -228,6 +228,10 @@ void _usage()
   cout << "   --caps=<caps>             list of caps (e.g., \"usage=read, write; user=read\"\n";
   cout << "   --yes-i-really-mean-it    required for certain operations\n";
   cout << "   --reset-regions           reset regionmap when regionmap update\n";
+  cout << "   --bypass-gc               when specified with bucket deletion, triggers\n";
+  cout << "                             object deletions by not involving GC\n";
+  cout << "   --inconsistent-index      when specified with bucket deletion and bypass-gc set to true,\n";
+  cout << "                             ignores bucket index consistency\n";
   cout << "\n";
   cout << "<date> := \"YYYY-MM-DD[ hh:mm:ss]\"\n";
   cout << "\nQuota options:\n";
@@ -2042,6 +2046,8 @@ int main(int argc, char **argv)
 
   int sync_stats = false;
   int reset_regions = false;
+  int bypass_gc = false;
+  int inconsistent_index = false;
 
   int extra_info = false;
 
@@ -2253,7 +2259,10 @@ int main(int argc, char **argv)
      // do nothing
     } else if (ceph_argparse_binary_flag(args, i, &extra_info, NULL, "--extra-info", (char*)NULL)) {
      // do nothing
-    } else if (ceph_argparse_binary_flag(args, i, &reset_regions, NULL, "--reset-regions", (char*)NULL)) {
+    } else if (ceph_argparse_binary_flag(args, i, &bypass_gc, NULL, "--bypass-gc", (char*)NULL)) {
+     // do nothing
+    } else if (ceph_argparse_binary_flag(args, i, &inconsistent_index, NULL, "--inconsistent-index", (char*)NULL)) {
+     // do nothing
     } else if (ceph_argparse_witharg(args, i, &val, "--caps", (char*)NULL)) {
       caps = val;
     } else if (ceph_argparse_witharg(args, i, &val, "-i", "--infile", (char*)NULL)) {
@@ -3687,6 +3696,7 @@ int main(int argc, char **argv)
   bucket_op.set_check_objects(check_objects);
   bucket_op.set_delete_children(delete_child_objects);
   bucket_op.set_fix_index(fix);
+  bucket_op.set_max_aio(max_concurrent_ios);
 
   // required to gather errors from operations
   std::string err_msg;
@@ -4634,7 +4644,11 @@ next:
   }
 
   if (opt_cmd == OPT_BUCKET_RM) {
-    RGWBucketAdminOp::remove_bucket(store, bucket_op);
+    if (inconsistent_index == false) {
+      RGWBucketAdminOp::remove_bucket(store, bucket_op, bypass_gc, true);
+    } else {
+      RGWBucketAdminOp::remove_bucket(store, bucket_op, bypass_gc, false);
+    }
   }
 
   if (opt_cmd == OPT_GC_LIST) {
--- a/src/rgw/rgw_bucket.cc
+++ b/src/rgw/rgw_bucket.cc
@@ -17,6 +17,7 @@
 #include "rgw_user.h"
 #include "rgw_string.h"
 
+#include "include/rados/librados.hpp"
 // until everything is moved from rgw_common
 #include "rgw_common.h"
 
@@ -459,9 +460,7 @@ int rgw_remove_bucket(RGWRados *store, rgw_bucket& bucket, bool delete_children)
   map<RGWObjCategory, RGWStorageStats> stats;
   std::vector<RGWObjEnt> objs;
   map<string, bool> common_prefixes;
-  rgw_obj obj;
   RGWBucketInfo info;
-  bufferlist bl;
   RGWObjectCtx obj_ctx(store);
 
   string bucket_ver, master_ver;
@@ -469,8 +468,6 @@ int rgw_remove_bucket(RGWRados *store, rgw_bucket& bucket, bool delete_children)
   ret = store->get_bucket_stats(bucket, RGW_NO_SHARD, &bucket_ver, &master_ver, stats, NULL);
   if (ret < 0)
     return ret;
-
-  obj.bucket = bucket;
 
   ret = store->get_bucket_info(obj_ctx, bucket.tenant, bucket.name, info, NULL);
   if (ret < 0)
@@ -490,7 +487,7 @@ int rgw_remove_bucket(RGWRados *store, rgw_bucket& bucket, bool delete_children)
 
     while (!objs.empty()) {
       std::vector<RGWObjEnt>::iterator it = objs.begin();
-      for (it = objs.begin(); it != objs.end(); ++it) {
+      for (; it != objs.end(); ++it) {
         ret = rgw_remove_object(store, info, bucket, (*it).key);
         if (ret < 0)
           return ret;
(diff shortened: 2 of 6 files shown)


robbat2:
@prallabh 
If you fire this no-GC deletion on a large bucket, then kill the RGW instance before it completes, what is the result?

prallabh:
@robbat2 I have tried that on a bucket with ~250K objects. After killing radosgw-admin in the midst of its completion, the bucket is left intact with the remaining objects (possibly some stale as well?). Bucket listing showed the left over ones. Upon restarting the same operation, it completed successfully.

Said that, I would need to verify this to be consistent, as the deletions are async, there might be a chance of not reaping some?

mattbenjamin:
@yehudasa I forgot to bring this up in standup--what are your thoughts on this?

yehudasa:
@mattbenjamin need to review it, but overall it's a useful tool to have.

src/rgw/rgw_bucket.cc
```diff
@@ -524,6 +521,156 @@ int rgw_remove_bucket(RGWRados *store, rgw_bucket& bucket, bool delete_children)
...
+
+  int max = 1000;
+  int max_aio = concurrent_max;
+  ret = list_op.list_objects(max, &objs, &common_prefixes, NULL);
+  if (ret < 0)
+    return ret;
+
+  while (!objs.empty()) {
+    std::vector<RGWObjEnt>::iterator it = objs.begin();
+    for (it = objs.begin(); it != objs.end(); ++it) {
```
yehudasa: @prallabh no need to reassign it

src/rgw/rgw_bucket.cc
```diff
@@ -524,6 +521,156 @@ int rgw_remove_bucket(RGWRados *store, rgw_bucket& bucket, bool delete_children)
...
+      RGWObjState *astate = NULL;
+      rgw_obj obj(bucket, (*it).key.name);
+      ret = store->get_obj_state(&obj_ctx, obj, &astate, NULL);
+      if (ret < 0)
```
yehudasa: `if (ret == -ENOENT) continue; ` probably

yehudasa: @prallabh and also dump a log message

src/rgw/rgw_bucket.cc
prallabh: @yehudasa done

src/rgw/rgw_bucket.cc
prallabh: @yehudasa done

prallabh:
@yehudasa ping

yehudasa:
@prallabh I still need to finish reviewing it. Hopefully can get to it this week.

src/rgw/rgw_admin.cc
```diff
@@ -2229,7 +2230,8 @@ int main(int argc, char **argv)
      // do nothing
     } else if (ceph_argparse_binary_flag(args, i, &extra_info, NULL, "--extra-info", (char*)NULL)) {
      // do nothing
-    } else if (ceph_argparse_binary_flag(args, i, &reset_regions, NULL, "--reset-regions", (char*)NULL)) {
+    } else if (ceph_argparse_binary_flag(args, i, &bypass_gc, NULL, "--bypass-gc", (char*)NULL)) {
```
yehudasa: @prallabh maybe also add this option to the radosgw-admin usage?

prallabh: done

theanalyst:
One of the commits is missing a signed-off-by

prallabh:
@theanalyst yes, the plan is to squash all once the review is done

src/rgw/rgw_admin.cc
```diff
@@ -227,6 +227,8 @@ void _usage()
   cout << "   --caps=<caps>             list of caps (e.g., \"usage=read, write; user=read\"\n";
   cout << "   --yes-i-really-mean-it    required for certain operations\n";
   cout << "   --reset-regions           reset regionmap when regionmap update\n";
+  cout << "   --bypass-gc               when specified with bucket deletion, would trigger\n";
```
theanalyst: This will require a test/cli/ help.t change as well

prallabh: done

src/rgw/rgw_bucket.cc
```diff
@@ -524,6 +521,156 @@ int rgw_remove_bucket(RGWRados *store, rgw_bucket& bucket, bool delete_children)
...
+  int max_aio = concurrent_max;
+  ret = list_op.list_objects(max, &objs, &common_prefixes, NULL);
+  if (ret < 0)
+    return ret;
+
+  while (!objs.empty()) {
+    std::vector<RGWObjEnt>::iterator it = objs.begin();
+    for (it = objs.begin(); it != objs.end(); ++it) {
+      RGWObjState *astate = NULL;
+      rgw_obj obj(bucket, (*it).key.name);
```
yehudasa: @prallabh also need to have `obj.set_instance(key.instance)`, otherwise versioned objects are not going to work

src/rgw/rgw_bucket.cc
```diff
@@ -524,6 +521,156 @@ int rgw_remove_bucket(RGWRados *store, rgw_bucket& bucket, bool delete_children)
...
+        RGWObjManifest& manifest = astate->manifest;
+        RGWObjManifest::obj_iterator miter = manifest.obj_begin();
+
+        if (miter.get_location().ns.empty()) {
+          head_obj = miter.get_location();
+        }
+
+        for (; miter != manifest.obj_end() && max_aio--; ++miter) {
+          if (!max_aio) {
+            ret = drain_handles(handles);
```
yehudasa: @prallabh why drain all the handles here? it's enough to drain the next handle, and increase max_aio by one, so that things are more streamlined

src/rgw/rgw_bucket.cc
```diff
@@ -524,6 +521,156 @@ int rgw_remove_bucket(RGWRados *store, rgw_bucket& bucket, bool delete_children)
...
+            max_aio = concurrent_max;
+          }
+
+          rgw_obj last_obj = miter.get_location();
+          if (last_obj == head_obj) {
+            // have the head obj deleted at the end
+            continue;
+          }
+
+          store->delete_obj_aio(last_obj, bucket, info, astate, handles);
```
yehudasa: not catching error here?

src/rgw/rgw_rados.cc
```diff
@@ -12052,3 +12052,45 @@ librados::Rados* RGWRados::get_rados_handle()
...
+  if (ret < 0) {
+    lderr(cct) << "ERROR: failed to get obj ref with ret=" << ret << dendl;
+    return ret;
+  }
+
+  RGWRados::Bucket bop(this, bucket_info);
+  RGWRados::Bucket::UpdateIndex index_op(&bop, obj, astate);
+
+  ObjectWriteOperation op;
+  ret = index_op.prepare(CLS_RGW_OP_DEL);
```
yehudasa: @prallabh the idea is to avoid accessing the bucket index altogether at this point. If you're removing the entire bucket, there's no reason to update the bucket index for every object. Just remove the objects and delete the bucket index at the end.

src/rgw/rgw_rados.cc
```diff
@@ -12052,3 +12052,45 @@ librados::Rados* RGWRados::get_rados_handle()
...
+  AioCompletion *c = librados::Rados::aio_create_completion(NULL, NULL, NULL);
+  ret = ref.ioctx.aio_operate(ref.oid, c, &op);
+  if (ret < 0) {
+    lderr(cct) << "ERROR: AioOperate failed with ret=" << ret << dendl;
+    return ret;
+  }
+
+  handles.push_back(c);
+
+  ret = delete_obj_index(obj);
```
yehudasa: @prallabh you're only removing the bucket index entry ... not the object itself.

src/rgw/rgw_bucket.cc
prallabh: @yehudasa The idea is to purge all the outstanding aio requests at one go. Please correct me, if we were to increment max_aio by one, wouldn't that result in so many drain calls for each iteration?

src/rgw/rgw_bucket.cc
prallabh: done

src/rgw/rgw_admin.cc
```diff
@@ -2021,6 +2023,7 @@ int main(int argc, char **argv)
   int sync_stats = false;
   int reset_regions = false;
   int bypass_gc = false;
+  int keep_index_consistent = false;
```
yehudasa: @prallabh I'll switch the option to `--inconsistent-index`, so that by default it keeps the index consistent.

src/rgw/rgw_bucket.h
```diff
@@ -280,7 +280,7 @@ class RGWBucket
           map<RGWObjCategory, RGWStorageStats>& calculated_stats,
           std::string *err_msg = NULL);
 
-  int remove(RGWBucketAdminOpState& op_state, bool bypass_gc = false, std::string *err_msg = NULL);
+  int remove(RGWBucketAdminOpState& op_state, bool bypass_gc = false, bool keep_index_consistent = false, std::string *err_msg = NULL);
```
yehudasa: again, by default we should keep index consistent

src/rgw/rgw_admin.cc
prallabh: done

src/rgw/rgw_bucket.h
prallabh: done

src/rgw/rgw_admin.cc
```diff
@@ -229,8 +229,8 @@ void _usage()
   cout << "   --reset-regions           reset regionmap when regionmap update\n";
   cout << "   --bypass-gc               when specified with bucket deletion, would trigger\n";
```
mattbenjamin: this text needs to be moved into the imperative or present tense "when specified with bucket deletion, triggers"--can't be "would"

src/rgw/rgw_admin.cc
```diff
@@ -229,8 +229,8 @@ void _usage()
...
   cout << "                             object deletions by not involving GC\n";
-  cout << "   --keep-index-consistent   set this to keep the bucket index consistent, if the bucket\n";
-  cout << "                             deletion is triggered with bypass-gc set to true\n";
+  cout << "   --inconsistent-index      setting this would not bother to keep the bucket index consistent,\n";
```
mattbenjamin: this text needs to be moved into the imperative or present tense "when specified with bucket deletion, ignore bucket index consistency (since it will be removed)"--can't be "would"

mattbenjamin:
I've verified that 
radosgw-admin -c ${CONF}  bucket rm --bucket=bu_9 --bypass-gc --purge-objects
runs and executes the expected bypass-gc remove-bucket routine
bucket is removed as expected after execution

prallabh:
@mattbenjamin yes, even without this patch, if you want to delete a non empty bucket, you need to provide the purge options.

src/rgw/rgw_admin.cc
prallabh: done

src/rgw/rgw_admin.cc
prallabh: done

prallabh:
@yehudasa @mattbenjamin I have squashed all the commits and rebased for your easy reference, thanks!

yehudasa:
@mattbenjamin merge at will

prallabh:
@mattbenjamin ping

PR #8683, opened 2016-04-21
compressor: fix of error code control in zlib decompress

Ved-vampir:
Little fix for zlib compression plugin,  Z_BUF_ERROR is valid return code for successful decompress process.

### DIFF
--- a/src/compressor/zlib/CompressionZlib.cc
+++ b/src/compressor/zlib/CompressionZlib.cc
@@ -134,7 +134,7 @@ int CompressionZlib::decompress(const bufferlist &in, bufferlist &out)
       strm.avail_out = max_len;
       strm.next_out = c_out;
       ret = inflate(&strm, Z_NO_FLUSH);
-      if (ret != Z_OK && ret != Z_STREAM_END) {
+      if (ret != Z_OK && ret != Z_STREAM_END && ret != Z_BUF_ERROR) {
        dout(1) << "Decompression error: decompress return "
             << ret << dendl;
        inflateEnd(&strm);


Ved-vampir:
All tests are still failing on master, can I do something for make check successful?

src/compressor/zlib/CompressionZlib.cc
```diff
@@ -134,7 +134,7 @@ int CompressionZlib::decompress(const bufferlist &in, bufferlist &out)
       strm.avail_out = max_len;
       strm.next_out = c_out;
       ret = inflate(&strm, Z_NO_FLUSH);
-      if (ret != Z_OK && ret != Z_STREAM_END) {
```
yuyuyu101: this worry me, from http://www.zlib.net/zlib_how.html it looks like we need to exit if meeting Z_BUF_ERROR.

Ved-vampir: @yuyuyu101 In this case strm.avail_out == max_len and we exit from while by the condition, nothing will be added to output, because have == 0 (https://github.com/Ved-vampir/ceph/blob/zlib_fix/src/compressor/zlib/CompressionZlib.cc#L143)
In zlib docs: "But now deflate() **produces no output at all**, and avail_out remains unchanged as CHUNK. That deflate() call wasn't able to do anything, either consume input or produce output, and so it returns Z_BUF_ERROR" (in inflate block link on this description). Also in example they don't check for this return value at all.

yuyuyu101: then we will continue loop?

Ved-vampir: @yuyuyu101 Condition is
while (strm.avail_out == 0)
And it become false, if we get Z_BUF_ERROR (according zlib manual), so we exit from the loop.

yuyuyu101:
LGTM

PR #8684, opened 2016-04-21
librados: copy out data to users' buffer for xio

vuhuong:
rados_aio_read(buf) call objecter->read(). Upon receiving
reply from osd, message->claim_data() is called which
enforces *deep* copy (both bufferlist and data buffers)
because Accelio's registered buffers is NON_SHAREABLE.
Received data is on new bufferlist and buffers but not
in original buffer provided by rados_aio_read() caller.
Another copy to users' buffer is needed

Signed-off-by: Vu Pham <vu@mellanox.com>

### DIFF
--- a/src/librados/IoCtxImpl.cc
+++ b/src/librados/IoCtxImpl.cc
@@ -808,6 +808,7 @@ int librados::IoCtxImpl::aio_read(const object_t oid, AioCompletionImpl *c,
   c->bl.clear();
   c->bl.push_back(buffer::create_static(len, buf));
   c->blp = &c->bl;
+  c->out_buf = buf;
 
   Objecter::Op *o = objecter->prepare_read_op(
     oid, oloc,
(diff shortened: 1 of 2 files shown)


src/librados/IoCtxImpl.cc
```diff
@@ -1630,6 +1631,8 @@ void librados::IoCtxImpl::C_aio_Ack::finish(int r)
   c->cond.Signal();
 
   if (r == 0 && c->blp && c->blp->length() > 0) {
+    if (c->out_buf && c->blp->c_str() != c->out_buf)
```
yuyuyu101: use buffer::is_provided_buffer?

mattbenjamin:
@vuhuong has this passed a rados test suite?

vuhuong:
@mattbenjamin It passed rados_aio_api test suite

liewegas:
Lots of crashes from ceph_test_rados_api_aio in http://pulpito.ceph.com/sage-2016-04-25_22:07:59-rados-wip-sage-testing2---basic-smithi/, i'm guessing it's this.

vuhuong:
It's my mistake when reading feedback from @yuyuyu101 and change from this:

+    if (c->out_buf && c->blp->c_str() != c->out_buf)

to

+    if (!c->blp->is_provided_buffer(c->out_buf))

without re-run test_rados_api_aio again.

It should be

+   if (c->out_buf && !c->blp->is_provided_buffer(c->out_buf))

I'll rework the patch, rerun test_rados_api_aio with simple/async

vuhuong:
This patchset passed rados/basic testsuite with both simple/async messengers

PR #8685, opened 2016-04-22
rgw: check for status >= 0

yehudasa:
Fixes: #15565

check only verified that status was > 0, so we ended not updating mtime for
the target object. This resulted in a continuous sync of objects.


yehudasa:
merged in with #8772

PR #8687, opened 2016-04-22
message: optimazation for message priority strategy

mslovy:
want peering and reply as fast as it can

Signed-off-by: Ning Yao <yaoning@unitedstack.com>

### DIFF
--- a/src/messages/MOSDPGInfo.h
+++ b/src/messages/MOSDPGInfo.h
@@ -31,10 +31,14 @@ public:
   epoch_t get_epoch() { return epoch; }
 
   MOSDPGInfo()
-    : Message(MSG_OSD_PG_INFO, HEAD_VERSION, COMPAT_VERSION) {}
+    : Message(MSG_OSD_PG_INFO, HEAD_VERSION, COMPAT_VERSION) {
+    set_priority(CEPH_MSG_PRIO_HIGH);
+  }
   MOSDPGInfo(version_t mv)
     : Message(MSG_OSD_PG_INFO, HEAD_VERSION, COMPAT_VERSION),
-      epoch(mv) { }
+      epoch(mv) {
+    set_priority(CEPH_MSG_PRIO_HIGH);
+  }
 private:
   ~MOSDPGInfo() {}
 
--- a/src/messages/MOSDPGLog.h
+++ b/src/messages/MOSDPGLog.h
@@ -42,18 +42,24 @@ public:
   spg_t get_pgid() { return spg_t(info.pgid.pgid, to); }
   epoch_t get_query_epoch() { return query_epoch; }
 
-  MOSDPGLog() : Message(MSG_OSD_PG_LOG, HEAD_VERSION, COMPAT_VERSION) { }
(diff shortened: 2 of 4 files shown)


src/messages/MOSDPGPushReply.h
```diff
@@ -31,8 +31,9 @@ class MOSDPGPushReply : public Message {
 
   MOSDPGPushReply() :
     Message(MSG_OSD_PG_PUSH_REPLY, HEAD_VERSION, COMPAT_VERSION),
-    cost(0)
-    {}
+    cost(0) {
+      set_priority(CEPH_MSG_PRIO_HIGH);
```
athanatos: Not this one, ReplicatedBackend bases the priority on the Push priority.

src/messages/MOSDRepOpReply.h
```diff
@@ -102,11 +102,14 @@ class MOSDRepOpReply : public Message {
     result(result_),
     final_decode_needed(false) {
     set_tid(req->get_tid());
+    set_priority(CEPH_MSG_PRIO_HIGH);
```
athanatos: ReplicatedBackend already sets this.  Either remove it from here, or remove it from ReplicatedBackend::sub_op_modify_*

src/messages/MOSDOpReply.h
```diff
@@ -127,6 +127,7 @@ class MOSDOpReply : public Message {
   MOSDOpReply()
     : Message(CEPH_MSG_OSD_OPREPLY, HEAD_VERSION, COMPAT_VERSION) {
     do_redirect = false;
+    set_priority(CEPH_MSG_PRIO_HIGH);
```
athanatos: Same here, ReplicatedBackend already sets this.

athanatos:
@mslovy 

PushReply has it's priority already set in ReplicatedBackend based on the Push priority -- that's better than setting it to HIGH unconditionally.  Remove the PushReply change.

The two OpReply messages already have HIGH set in ReplicatedBackend.  It's confusing to set it in both places.  If you want to switch it to be set in the constructor, please separate that change into its own commit and remove the set_priority calls from ReplicatedBackend.

Setting the peering message priority in the constructor does make some sense, but don't do it in the same commit as the OpReply messages.

mslovy:
@athanatos ,  updated， please review again. I think it is not reasonable to set the priority for MOSDSubOpReply since it is also used by others (like scrubbing). So I revert all modification related to Reply Ops, it would be fine and sometime more instructive to guide us setting different priority in different cases. 
Optimize priority setting for  kick_recovery_op

athanatos:
@mslovy That second commit should be a separate PR.  Currently, there are no tests at all that have different clients at different priorities, I would be surprised if it worked.  That *is* work that needs to be done.  The first step would be to add into the ceph-qa-suite rados/thrash suite a workload with two ceph_test_rados instances using different client priorities.  Once we have that to test against, we can start merging patches to fix that behavior.

mslovy:
@athanatos ,  reset the second commit and I will propose a new PR later as mentioned

athanatos:
Ok, this makes sense, adding the needs-qa label.

PR #8689, opened 2016-04-22
msg/async/EventSocket: fix missing break statement

runsisi:
fix missing `break` statement and conditional compilation for option `--without-eventfd`

### DIFF
--- a/src/test/librbd/test_librbd.cc
+++ b/src/test/librbd/test_librbd.cc
@@ -40,6 +40,7 @@
 #include "test/librados/test.h"
 #include "test/librbd/test_support.h"
 #include "common/errno.h"
+#include "common/event_socket.h"
 #include "include/interval_set.h"
 #include "include/stringify.h"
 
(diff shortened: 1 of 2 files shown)


yuyuyu101:
good catch, it looks like we don't have test to verify the return code. it would be appreciate to add unittest

runsisi:
@yuyuyu101 OK, i will add it :)

runsisi:
hi @yuyuyu101 , sorry for my late response.
unittest added, can u review it for me ? many thx :)

src/test/librbd/test_librbd.cc
```diff
@@ -4047,6 +4048,72 @@ TEST_F(TestLibRBD, TestImageOptionsPP)
...
+  ASSERT_EQ(1, read(pipe_fd[0], buf, 32));
+  ASSERT_EQ('i', buf[0]);
+
+  close(pipe_fd[0]);
+  close(pipe_fd[1]);
+}
+
+TEST_F(TestLibRBD, EventSocketEventfd)
+{
+#ifdef HAVE_EVENTFD
```
xiexingguo: How about move this out of the whole test case? So we don't even try if HAVE_EVENTFD is undefined.

runsisi: i'd prefer ur idea, i just don't want to violate the same style as the old test, i.e., `TEST_F(TestLibRBD, ImagePollIO)` below, though :)

xiexingguo: Sorry, I did not pay attention to that.  Then both are okay with me, suit yourself.

yuyuyu101:
thx!

yuyuyu101:
passed http://pulpito.ceph.com/haomai-2016-05-01_23:40:37-rados-wip-haomai-testing-distro-basic-smithi/

yuyuyu101:
retest this please

liewegas:
lots of errors in http://pulpito.ceph.com/sage-2016-05-06_16:42:17-rados-wip-sage-testing---basic-smithi/, trying again without this

yuyuyu101:
I think these errors are not related to this pr? @liewegas

liewegas:
I will run it through again

yuriw:

PR #8693, opened 2016-04-22
common: show success/fail info in json for 'perf reset' command

cooboos:
When run 'perf reset' command ,there is no return; 
Show the info when run 'perf reset command' in Jason Formater.

Signed-off-by: wangsongbo  wangsongbo@unitedstack.com


yuyuyu101:
I suggest you can add tests in  test/perf_counters.cc. So we can keep format compatible later

cooboos:
@yuyuyu101 , thank you for your suggestion. I will add it later.

PR #8695, opened 2016-04-22
 mds: dispatch assert cleanup

renhwztetecs:
fixed dispatch in the default branch.
    include:
    mdcache->dispatch(m);
    mdcache->migrator->dispatch(m);
    server->dispatch(m);
    balancer->proc_message(m);
    locker->dispatch(m);
    client->handle_request(req);
    server->handle_request(req);

### DIFF
--- a/src/mds/Locker.cc
+++ b/src/mds/Locker.cc
@@ -98,7 +98,8 @@ void Locker::dispatch(Message *m)
     break;
     
   default:
-    assert(0);
+    derr << "locker unknown message " << m->get_type() << dendl;
+    assert(0 == "locker unknown message");
   }
 }
 
--- a/src/mds/MDBalancer.cc
+++ b/src/mds/MDBalancer.cc
@@ -58,10 +58,8 @@ int MDBalancer::proc_message(Message *m)
     break;
 
   default:
-    dout(1) << " balancer unknown message " << m->get_type() << dendl;
-    assert(0);
-    m->put();
-    break;
+    derr << " balancer unknown message " << m->get_type() << dendl;
+    assert(0 == "balancer unknown message");
   }
 
   return 0;
--- a/src/mds/MDCache.cc
+++ b/src/mds/MDCache.cc
@@ -7624,10 +7624,8 @@ void MDCache::dispatch(Message *m)
     break;
     
   default:
-    dout(7) << "cache unknown message " << m->get_type() << dendl;
-    assert(0);
-    m->put();
-    break;
+    derr << "cache unknown message " << m->get_type() << dendl;
+    assert(0 == "cache unknown message");
   }
 }
 
--- a/src/mds/MDSTableClient.cc
+++ b/src/mds/MDSTableClient.cc
@@ -127,7 +127,7 @@ void MDSTableClient::handle_request(class MMDSTableRequest *m)
     break;
 
   default:
-    assert(0);
+    assert(0 == "unrecognized mds_table_client request op");
   }
 
   m->put();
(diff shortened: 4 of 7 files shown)


src/mds/MDSRank.cc
```diff
@@ -576,9 +576,9 @@ bool MDSRank::handle_deferrable_message(Message *m)
       // SERVER
     case CEPH_MSG_CLIENT_SESSION:
     case CEPH_MSG_CLIENT_RECONNECT:
-      ALLOW_MESSAGES_FROM(CEPH_ENTITY_TYPE_CLIENT);
       // fall-thru
     case CEPH_MSG_CLIENT_REQUEST:
+      ALLOW_MESSAGES_FROM(CEPH_ENTITY_TYPE_CLIENT);
```
gregsfortytwo: I'm not certain about this one -- do we never forward client request message types from other daemon types?

renhwztetecs: I understand message types should be derived from CEPH_ENTITY_TYPE_CLIENT rather than the other,
Because the server - > dispatch (m) can only deal with the client type of message
:-)

src/mds/Locker.cc
```diff
@@ -98,7 +98,8 @@ void Locker::dispatch(Message *m)
     break;
     
   default:
-    assert(0);
+    dout(7) << "locker unknown message " << m->get_type() << dendl;
```
jcsp: If we're going to assert out anyway let's make this a derr instead of dout

src/mds/MDBalancer.cc
```diff
@@ -58,10 +58,8 @@ int MDBalancer::proc_message(Message *m)
     break;
 
   default:
-    dout(1) << " balancer unknown message " << m->get_type() << dendl;
-    assert(0);
-    m->put();
-    break;
+    dout(7) << "balancer unknown message " << m->get_type() << dendl;
```
jcsp: as above, s/dout/derr/

src/mds/MDSRank.cc
jcsp: Hmm, in MDSRank::forward_message_mds we convert client-sent MSG_CLIENT_REQUESTs into MClientRequestForwards back to the client, rather than directly sending on to another MDS.

I can't see a path where we actually send an MClientRequest from one mds to another, but it's not the easiest thing to grep for

jcsp:
I'm not sure I see the actual bug fix here, isn't this just tidying up some assertions?

src/mds/MDSRank.cc
gregsfortytwo: I think you're right @jcsp. Running it through the multi-mds test suite should make it pretty obvious if we're wrong, if we want to do a special run.

renhwztetecs: agree, do not modify before not sure

renhwztetecs:
I think set cleanup is more appropriate^-^

src/mds/Locker.cc
renhwztetecs: done

jcsp:
:+1:

jcsp:
I'm happy with this after latest changes, will leave it for Yuri to merge after tests complete.

jcsp:
Ah, I notice that this has had the "mds/MDSRank add msg type judgment" commit dropped.  That's good, because in my test branch it was breaking test_hardlink_reintegration (MDSs *do* send each other client requests, for the "rename" of a stray out of the stray dir)

renhwztetecs:
@jcsp Thank you for confirmation

jcsp:
Passed here: http://pulpito.ceph.com/jspray-2016-06-02_01:58:41-fs-wip-jcsp-testing-20160601---basic-mira/

PR #8698, opened 2016-04-22
rbd: help message distinction between commands and aliases

HeyoQiang:
Fixes BUG #15521:Seeing rollback and revert option for snapshot, which seems to be same and can be confusing for users

Signed-off-by: Yongqiang He <he.yongqiang@h3c.com>

### DIFF
--- a/src/tools/rbd/Shell.cc
+++ b/src/tools/rbd/Shell.cc
@@ -84,6 +84,7 @@ int Shell::execute(const Arguments& cmdline_arguments) {
                                      cmdline_arguments.end());
   std::vector<std::string> command_spec;
   get_command_spec(arguments, &command_spec);
+  bool is_alias = true;
 
   if (command_spec.empty() || command_spec == CommandSpec({"help"})) {
     // list all available actions
@@ -92,12 +93,12 @@ int Shell::execute(const Arguments& cmdline_arguments) {
   } else if (command_spec[0] == HELP_SPEC) {
     // list help for specific action
     command_spec.erase(command_spec.begin());
-    Action *action = find_action(command_spec, NULL);
+    Action *action = find_action(command_spec, NULL, &is_alias);
     if (action == NULL) {
       print_unknown_action(command_spec);
       return EXIT_FAILURE;
     } else {
-      print_action_help(action);
+      print_action_help(action, is_alias);
       return 0;
     }
   } else if (command_spec[0] == BASH_COMPLETION_SPEC) {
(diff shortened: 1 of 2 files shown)


dillaman:
Perhaps as an alternative suggestion, the per-command help page could include the alias (if present):

> $ rbd help snap revert
> usage: rbd snap rollback [--pool <pool>] [--image <image>] [--snap <snap>] 
>                          [--no-progress] 
>                          <snap-spec> 
> alias: rbd snap revert
> 
> Rollback image to snapshot.
> 
...

dillaman:
You most likely will have to update the `src/test/cli/rbd/help.t` test output -- try running the ```test/run-cli-tests``` test

HeyoQiang:
@dillaman I think we can keep it in original format.

src/tools/rbd/Shell.cc
```diff
@@ -302,14 +305,19 @@ void Shell::print_help() {
...
+ }
 
+void Shell::print_action_help(Action *action, bool is_alias) {
   std::stringstream ss;
-  ss << "usage: " << APP_NAME << " "
-     << format_command_spec(action->command_spec);
-  std::cout << ss.str();
+  if (is_alias == false) {
+    ss << "usage: " << APP_NAME << " "
+       << format_command_spec(action->command_spec);
```
dillaman: Probably a good use of the ternary operator to eliminate the dup code: ```format_command_spec(is_alias ? action->alias_command_spec : action->command_spec)```

src/tools/rbd/Shell.h
```diff
@@ -57,12 +57,12 @@ class Shell {
   void get_command_spec(const std::vector<std::string> &arguments,
                         std::vector<std::string> *command_spec);
   Action *find_action(const CommandSpec &command_spec,
-                      CommandSpec **matching_spec);
+                      CommandSpec **matching_spec, bool &is_alias);
```
dillaman: Nit: should be a pointer per the coding style

src/tools/rbd/Shell.cc
HeyoQiang: @dillaman  lgtm.

src/tools/rbd/Shell.h
HeyoQiang: @dillaman I have changed it

dillaman:
@HeyoQiang The latest commit doesn't compile.

dillaman:
@HeyoQiang Last comment: can you change your commit message to include:

> Fixes: https://tracker.ceph.com/issues/15521

instead of:

> Fix BUG #15521 :Seeing rollback and revert option for snapshot, which seems to be same and can be confusing for users

HeyoQiang:
@dillaman I have changed it.

PR #8699, opened 2016-04-22
RFC: change default monitor port

liewegas:
We now have an IANA assigned port that we should use for monitors.

This PR makes (new) clients try the new then old ports when it is not
specified (in mon host, -m option, etc.).

New clusters use the new port.  Old clusters keep their old port.

The big question for me is whether it is okay that a new cluster will start
using the new port immediately, but old (jewel) clients will not try the
new port yet.  That means a new kraken cluster won't get reachable from a
jewel client that specifies only the mon host/ip (without the port).  I don't
think this is very common, and all they need to do to "fix" it is specify the
port in the mon host config option.

### DIFF
--- a/PendingReleaseNotes
+++ b/PendingReleaseNotes
@@ -1,0 +1,16 @@
+v11.0.0
+-------
+
+* The default monitor port has changed from 6789 to 3300, which has
+  been assigned to us by IANA.  Existing monitor will continue to use
+  the ports they were created with, while new clusters will default to the
+  new port.
+
+  Clients will try first the new port and then the old port if it is
+  not explicitly specified, except in one case.  If the monitor
+  addresses are specified in the ``ceph.conf`` file by the ``mon
+  addr`` option, we will default to the new port and will not try the
+  old port as well.  Note that this method specifying monitors is
+  deprecated.  Please use ``mon hosts`` in the ``[global]`` section
+  instead, or modify the options to explicitly specify the old port
+  6789 if needed.
--- a/src/include/msgr.h
+++ b/src/include/msgr.h
@@ -9,14 +9,8 @@
  * Data types for message passing layer used by Ceph.
  */
 
-#define CEPH_MON_PORT    6789  /* default monitor port */
-
-/*
- * client-side processes will try to bind to ports in this
- * range, simply for the benefit of tools like nmap or wireshark
- * that would like to identify the protocol.
- */
-#define CEPH_PORT_FIRST  6789
+#define CEPH_MON_PORT         3300  /* (ce4h) IANA assigned monitor port */
+#define CEPH_MON_PORT_LEGACY  6789  /* old default monitor port */
 
 /*
  * tcp connection banner.  include a protocol version. and adjust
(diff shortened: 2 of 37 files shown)


xiexingguo:
Good news :+1:

src/include/msgr.h
```diff
@@ -9,14 +9,8 @@
...
 
-#define CEPH_MON_PORT    6789  /* default monitor port */
-
-/*
- * client-side processes will try to bind to ports in this
- * range, simply for the benefit of tools like nmap or wireshark
- * that would like to identify the protocol.
- */
-#define CEPH_PORT_FIRST  6789
+#define CEPH_MON_PORT         3300  /* (ce4h) IANA assigned monitor port */
```
runsisi: typo?

tchaikov: what do you mean for typo? "CE4h" stands for "CEPH", imo, it's a very cool port id.

xiexingguo: @runsisi  ce4h means ce4 in Hex format, which is equivalent to 3300 in Dec format.

runsisi: @tchaikov @xiexingguo OK, got it. thank you both :)

PendingReleaseNotes
```diff
@@ -0,0 +1,16 @@
...
+* The default monitor port has changed from 6789 to 3300, which has
+  been assigned to us by IANA.  Existing monitor will continue to use
+  the ports they were created with, while new clusters will default to the
+  new port.
+
+  Clients will try first the new port and then the old port if it is
+  not explicitly specified, except in one case.  If the monitor
+  addresses are specified in the ``ceph.conf`` file by the ``mon
+  addr`` option, we will default to the new port and will not try the
+  old port as well.  Note that this method specifying monitors is
```
jecluis: I think it would be trivial to also support this, but I'm okay with the approach as to ensure the continuing deprecation of 'mon addr'.

jecluis:
LGTM

jecluis:
Just noticed the build has failed. It was not clear whether it was due to these changes, but the units it failed make me wonder. I'm not currently at home to test this and I'll be on vacation next week too, so don't wait on me to merge this once you figure out whether it was the changes that made the tests fail.

PendingReleaseNotes
liewegas: > > @@ -0,0 +1,16 @@
> > +v11.0.0
> > +-------
> > +
> > +* The default monitor port has changed from 6789 to 3300, which has
> > +  been assigned to us by IANA.  Existing monitor will continue to use
> > +  the ports they were created with, while new clusters will default to the
> > +  new port.
> > +
> > +  Clients will try first the new port and then the old port if it is
> > +  not explicitly specified, except in one case.  If the monitor
> > +  addresses are specified in the ``ceph.conf`` file by the ``mon
> > +  addr`` option, we will default to the new port and will not try the
> > +  old port as well.  Note that this method specifying monitors is
> 
> I think it would be trivial to also support this, but I'm okay with the approach as to ensure the continuing deprecation
> of 'mon addr'.

I didn't because the code that parses the config also names the mon's, and 
we need to generate 2 entries to probe both ports.  It'd be a bit kludgey.  
I'd rather take the opportunity to discourage this approach :)

dachary:
@liewegas needs a rebase

liewegas:
this is too stale to be useful.  teh port change will still be tied to msgr2 when we get to it.

PR #8700, opened 2016-04-22
debian: package systemd targets

tchaikov:

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -116,8 +116,7 @@ binary-arch: build install
 	dh_install -a --sourcedir=$(DESTDIR) --list-missing
 	install -d -m0755 debian/ceph-base/etc/logrotate.d
 	install -m0644 debian/ceph.logrotate debian/ceph-base/etc/logrotate.d
-	dh_installinit -p ceph-base --name ceph --no-start
-	dh_installinit -p radosgw --no-start
+
 	# dh_installinit is only set up to handle one upstart script
 	# per package, so do this ourselves
 	install -d -m0755 debian/ceph-base/etc/init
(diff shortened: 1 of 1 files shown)


liewegas:
I don't think invoke-rc.d will do the right thing.  The 'start ceph-all' and 'start ceph-osd-all' etc lines are a bit of a hack so that the *wrapper* meta-service is running after a fresh install, when no daemons actually exist.  That way, when the daemon(s) is/are created, 'stop ceph-osd-all' will stop them.

It's been this way for a long time, and given that we're basic at the end of upstart anyway, I don't think we should change it..

tchaikov:
changelog

- dropped the commit to s/start/invoke-rc.d/

debian/rules
```diff
@@ -155,18 +154,28 @@ binary-arch: build install
...
 
 	install -d -m0755 debian/rbd-mirror/lib/systemd/system
 	install -m0644 systemd/ceph-rbd-mirror@.service debian/rbd-mirror/lib/systemd/system
 	sed -i s./etc/sysconfig/./etc/default/.g debian/rbd-mirror/lib/systemd/system/ceph-rbd-mirror@.service
+	install -m0644 systemd/ceph-rbd-mirror.target debian/rbd-mirror/lib/systemd/system
+
+	dh_systemd_enable
+	dh_installinit -p ceph-base --name ceph --no-start
+	dh_installinit -p radosgw --no-start
+	dh_systemd_start --restart-after-upgrade
```
liewegas: hmm, do we really want --restart-after-upgrade?  traditionally we haven't done that.

liewegas: i think we want --no-restart-on-upgrade

tchaikov: > --restart-after-upgrade
>           Do not stop the unit file until after the package upgrade has been completed.

this reminds me how we do the upgrade

```
   sequential:
   - install.upgrade:
       mon.a:
       mon.b:
   - ceph.restart: [mon.a, mon.b, mon.c, mds.a, osd.0, osd.1, osd.2, osd.3]
```

so i thought we didn't want to stop the service until upgrade is finished. but yeah. i can drop that option.

tchaikov: > i think we want --no-restart-on-upgrade

okay. if we expect our user to start the upgraded service manually.

tchaikov:
changelog

- s/--restart-after-upgrade/--no-restart-on-upgrade/

liewegas:
lgtm!  @javacruft ?

tchaikov:
@javacruft ping?

jmunhoz:
Works on Jessie amd64! btw the commit description uses the old Fixes: # format instead of the new URL format

tchaikov:
changelog

- update the commit message to fix the "Fixed" label

ramsberg:
Hi. Sorry for a lay-man question: When can this fix be expected to be implemented in deb packages for ceph hammer for armhf/jessie (raspberry Pi) ?

tchaikov:
@ramsberge it's a known issue, see http://tracker.ceph.com/issues/11344. but i see there are still some open questions documented in that ticket not yet answered.

ramsberg:
Thanks Kefu. 
It is a little troubling though, that this discussion is over one year old with no directly constructive updates added. 
Is that normal? 
/Olle 
On May 1, 2016 16:27, Kefu Chai <notifications@github.com> wrote:@ramsberge it's a known issue, see http://tracker.ceph.com/issues/11344. but i see there are still some open questions documented in that ticket not yet answered.

liewegas:
The systemd issues are all settled--I closed that bug.

For hammer, we don't have systemd files backported, and don't plan to.  If you want systemd, upgrade to jewel.

Note that we are only doing armhf jessie builds on jewel and later, and only test builds currently--no release ones.

ramsberg:
Ahh! Got it! Will dig into getting jewel running on raspbian/armhf/jessie!
Thanks!

PR #8701, opened 2016-04-22
ceph-osd-prestart.sh: compute correct defaultweight for bluestore

dvanders:
The default crush weight should be the size of the block device, not
the tiny 10M xfs partition.

Backport: jewel


liewegas:
The plan is to fix it with this: https://github.com/ceph/ceph/pull/8357

This is a fine workaround for now, though!

PR #8702, opened 2016-04-22
global-init: fixup chown of the run directory along with log and asok files

kmroz:
As part of deferred setuid/setgid, we also need to chown the run_dir if specified in the config. This ensures that asok files created in a custom run_dir are ultimately owned by the correct user:group.

@liewegas - Please have a look? It adds make your 'bit of a hack' into a bigger one. Further, pid files are not handled either. Adding to a pidfile chown to this code block does not work as the pid files appear to be created later (not sure where). I think a followup PR for this will be needed also.

Hit this while debugging: http://tracker.ceph.com/issues/15553

### DIFF
--- a/src/common/admin_socket.cc
+++ b/src/common/admin_socket.cc
@@ -290,7 +290,7 @@ void* AdminSocket::entry()
 void AdminSocket::chown(uid_t uid, gid_t gid)
 {
   if (m_sock_fd >= 0) {
-    int r = ::fchown(m_sock_fd, uid, gid);
+    int r = ::chown(m_path.c_str(), uid, gid);
     if (r < 0) {
       r = -errno;
       lderr(m_cct) << "AdminSocket: failed to chown socket: "
--- a/src/common/ceph_context.cc
+++ b/src/common/ceph_context.cc
@@ -641,6 +641,11 @@ uint32_t CephContext::get_module_type() const
   return _module_type;
 }
 
+void CephContext::set_init_flags(int flags)
+{
+  _init_flags = flags;
+}
+
 int CephContext::get_init_flags() const
 {
   return _init_flags;
--- a/src/common/ceph_context.h
+++ b/src/common/ceph_context.h
@@ -84,6 +84,7 @@ public:
   /* Get the module type (client, mon, osd, mds, etc.) */
   uint32_t get_module_type() const;
 
+  void set_init_flags(int flags);
   int get_init_flags() const;
 
   /* Get the PerfCountersCollection of this CephContext */
--- a/src/common/common_init.cc
+++ b/src/common/common_init.cc
@@ -12,6 +12,7 @@
  *
  */
 
+#include "common/admin_socket.h"
 #include "common/ceph_argparse.h"
 #include "common/ceph_context.h"
 #include "common/ceph_crypto.h"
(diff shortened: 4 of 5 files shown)


liewegas:
I think my original hack wasn't quite right.. we only need to do the hacky chown if we are deferring the setuid/setgid.  Do you mind fixing that?  Otherwise I'm worried we'll make noise on stderr in cases where we can't chown /var/run but it doesn't need to be chowned.

I'm not sure this addresses the real problem, though. What is creating or chwoning /var/run/ceph in the first place?

smithfarm:
The packaging creates/chowns `/var/run/ceph` but in http://tracker.ceph.com/issues/15553 someone is doing this:

```
[global]
log file = /var/log/ceph/$type.$id/log
pid file = /var/run/ceph/$type.$id/pid
```

The `$type.$id` directories are getting created root.root :-(

kmroz:
Sure... I will wrap the chown calls within a if(flags&CINIT_FLAG_DEFER...)

So we know /var/run/ceph (ie. default run_dir) is created by global_init(). It generally has no issue as the default is re-used by each daemon during startup (worst case hitting EEXIST). It can be a problem, however, in the case when a custom run_dir is specified in the config for each daemon. Then RGW will create it's own (let's say /var/run/ceph/rgw/) with root:root ownership. I suppose this itself doesn't address http://tracker.ceph.com/issues/15553, rather it's just unsightly to have a run dir with root:root ownership, and perhaps some asok file sitting inside that has been chown(ceph:ceph).

As to http://tracker.ceph.com/issues/15553 - I think the problem is that the user specified custom pid_file/log_file variables with directories that were created with root:root ownership. I've no idea where these dirs came from, and as far as I can see, they would not be created by _any_ ceph daemon as no run_dir was specified in the config (thus the default /var/run/ceph would have only been created). Trying to reproduce locally with the users config, I hit directory not found errors when trying to create the pid and logfile. Adding the correct run_dir path fixes the pid problem. However, I don't think there is an equivalent log_dir option. As such, any custom log_file path better exist beforehand.

liewegas:
Hmm, what is creating those subdirectories?  I don't remember doing 
that in the ceph code.

kmroz:
That's just it - looking at the code that generates the run directory... it would only generate such subdirs if the user specified in his config `run_dir = /var/run/ceph/$type.$id/` - which he did not. Not knowing the user's environment, @smithfarm and I did a lot of speculation today. Perhaps some other script being fired that created these on the user's behalf, only with the incorrect permissions.

liewegas:
If someone else is creating the dirs I don't think we have to be 
responsible for their ownership.  Would that turn this into a wontfix, or 
do we still need a chown?

kmroz:
This PR is trying to just address something I uncovered while looking into http://tracker.ceph.com/issues/15553. That is, when a custom run_dir is specified for RGW, because of the possible permissions defer, we will create the run_dir as root:root. It's unsightly, but I suppose not the end of the world as the asok and logfiles would have been created in their respective directories as root:root, and the chown()'d.

liewegas:
Right, sorry--trying to do too many things at once.  This looks right to me.  Thanks!

kmroz:
@liewegas - No worries :D I've updated http://tracker.ceph.com/issues/15553 a bit as well to try to poll for more info regarding some relic scripts that may be firing in the user's environment and generating these subdirs.

kmroz:
Marking this as DNM for now. I'd like to also take care of pid file permissions as part of this PR before it goes to qa, etc.

kmroz:
An update to the original which does a couple extra things:

1. Adds a basic ceph_context:set_init_flags() function
    - Used by global_init() to update init_flags of a daemon in case they've changed (ie. radosgw manually runs global_pre_init(), then adds a CINIT_DEFERRED_FLAG, then global_init()).
2. Guards the chown operations using ceph_context:get_init_flags() and the deferred flag.
3. chown()'s pid files.
4. Moved chown()'ing asok files out of global_init() because _admin_socket->m_sock_fd == 0 at this point :( The chown() needs to happen after ceph_context:start_service_thread() in common_init_finish().
5. **Added a FIXME note to common_init_finish() documenting that, despite successfully invoking AdminSocket::chown(), the ::fchown() operation does not change the owner:group of the socket file. Linux still reports the original owner:group. This does not appear to impact removal of the socket file on daemon shutdown. Requires some more investigation and followup.**

kmroz:
@liewegas - Latest commit changes AdminSocket::chown() method to use chown() instead of fchown(). fchown(), while returning success, does not actually change the asok file ownership. Going with path-based chown() succeeds.

PR #8703, opened 2016-04-22
ceph_test_keyvaluedb: fixup unit test for keyvaluedb

liewegas:
Fixes: http://tracker.ceph.com/issues/15435


liewegas:
https://github.com/ceph/ceph-qa-suite/pull/971

PR #8704, opened 2016-04-22
cmake: removed rdmacm & ibverbs parallel to PR#8637

alimaredia:


mattbenjamin:
+1 verified

PR #8709, opened 2016-04-22
rgw: object delete records timestamp, skip unneeded deletes

yehudasa:
Object delete now records the removed object's timestamp. This is being used by tracking zone to guard the delete (uses the if-not-modified-since logic).
Also, don't need to do the object delete sequence if object does not exist.


cbodley:
closing, these commits should merge with #8772

PR #8710, opened 2016-04-22
Signed-off-by: Tamil Muthamizhan <tmuthami@redhat.com>

tmuthamizhan:
renamed the workunit and test to reflect reality and for better clarity

### DIFF
--- a/qa/workunits/ceph-tests/ceph-admin-commands.sh
+++ b/qa/workunits/ceph-tests/ceph-admin-commands.sh
@@ -1,0 +1,13 @@
+#!/bin/sh -e
+
+#check ceph health
+ceph -s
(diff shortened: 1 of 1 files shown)


yuriw:
@tmuthamizhan you will need to change references in ceph-qa-suite for:
./suites/ceph-ansible/
./suites/ceph-deploy/
alone with this PR, I think, and then cherry-pick on all branches you want to have it changed, unless you want this change to be  from now on (in `kaken`) and keep the old suites unchanged.

tmuthamizhan:
@yuriw , sure there is another PR for ceph-qa-suite - https://github.com/ceph/ceph-qa-suite/pull/972.

PR #8715, opened 2016-04-23
cmake: use LINK_PRIVATE instead of PRIVATE for CMake version == 2.8.11

runsisi:
with version 2.8.11, i.e., the minimum version required by the building
system, we will hit a compile error as follows:
```
  /usr/bin/ld: cannot find -lPRIVATE
  make[2]: *** [lib/librbd.so.1.0.0] Error 1
  make[1]: *** [src/librbd/CMakeFiles/librbd.dir/all] Error 2
```


tchaikov:
lgtm

PR #8717, opened 2016-04-23
osd: make op_is_discardable check optional

ZhenyuLeng:
There is no need to know status of connection after op has been dequeued, this will introduce some lock latency

### DIFF
--- a/src/common/config_opts.h
+++ b/src/common/config_opts.h
@@ -892,6 +892,8 @@ OPTION(osd_bench_large_size_max_throughput, OPT_U64, 100 << 20) // 100 MB/s
 OPTION(osd_bench_max_block_size, OPT_U64, 64 << 20) // cap the block size at 64MB
 OPTION(osd_bench_duration, OPT_U32, 30) // duration of 'osd bench', capped at 30s to avoid triggering timeouts
 
+OPTION(osd_discard_disconnected_ops, OPT_BOOL, true)
+
 OPTION(memstore_device_bytes, OPT_U64, 1024*1024*1024)
 OPTION(memstore_page_set, OPT_BOOL, true)
 OPTION(memstore_page_size, OPT_U64, 64 << 10)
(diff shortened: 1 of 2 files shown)


liewegas:
have you tried to quantify the difference?  the idea is to avoid wasting time on requests when teh client has disconnected.

ZhenyuLeng:
Hi, thank you for reply.
We use Intel p3700 SSD to quantify the difference, the access pattern is random read. We found that it needs about 300us for PGShardWQ to handle a read op, while this function(op_is_discardable) takes nearly 30us to check the state of connection. After dropping the function, we got 9.6% performance improvement. However, if the storage device is disk or the access pattern is write, 30us will be insignificant.
Also, we thought it would be better to treat judging the state of connection as a parameter whether the network is believable or extremely unlikelihood.

liewegas:
The best solution is probably to optimize is_connected() to be fast (using 
e.g. an atomic_t).  But for now it also makes sense to have an OSD option 
whether to discard ops from disconnected clients.  Let's do that...

 osd discard disconnected ops = true

by default, to preserve existing behavior, but you can set it to off for 
nvme.

We'll need to adjust ceph-qa-suite so that some of the rados tests set 
this to get good coverage.

ZhenyuLeng:
Okay, Let's go ahead.

ZhenyuLeng:
How to adjust ceph-qa-suite?

liewegas:
Add a line setting the config option in 1 or 2 of the yaml fragmetns in 
ceph-qa-suite.git/suites/rados/thrash/workload/*.yaml

athanatos:
@ZhenyuLeng Please submit a ceph-qa-suite PR enabling this option for some workload (don't really care which one).  I can add this to wip-sam-testing once I have a ceph-qa-suite branch which actually tests it.

ZhenyuLeng:
@athanatos Okay，i'm working on it.

ZhenyuLeng:
Is the change in ceph-qa-suite not right? Looking for help @liewegas @athanatos

liewegas:
ceph-qa-suite pr merged

ZhenyuLeng:
Should i commit this again for checks?

yuyuyu101:
Hmm, I notice this latency and just for clarify I submitted this PR https://github.com/ceph/ceph/pull/8520 already.

ZhenyuLeng:
Wow, i see it @yuyuyu101  
But if someone are using simple or xio messenger, the problem is still alive.
What do you think?

yuriw:

PR #8722, opened 2016-04-23
[DNM] rgw: reduce string copies in tempURL processing

mattbenjamin:
Introduce a helper class encapsulating the temp URL HMAC computation,
and also save it for later computation.  Use the underlying HMACSHA1
implementation to compute the HMAC incrementally (saves a copy of all
the parameters).

Also avoid some other string copying in the logic prior to taking the
HMAC, in one case by remembering to use auto& in a std::map range-for
loop, and using boost::string_ref to share ReqState::request_uri and
conditionally suffix of same.

### DIFF
--- a/src/rgw/rgw_swift.cc
+++ b/src/rgw/rgw_swift.cc
@@ -5,6 +5,8 @@
 #include <stdlib.h>
 #include <unistd.h>
 
+#include <boost/utility/string_ref.hpp>
+#include "common/ceph_crypto.h"
 #include "common/ceph_json.h"
 #include "rgw_common.h"
 #include "rgw_swift.h"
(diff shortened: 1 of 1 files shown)


mattbenjamin:
@rzarzynski could you look this over?  @cbodley maybe also?

src/rgw/rgw_swift.cc
```diff
@@ -568,24 +570,47 @@ static void temp_url_make_content_disp(req_state * const s)
...
-  calc_hmac_sha1(key.c_str(), key.size(),
-                 str.c_str(), str.size(),
-                 dest);
+  unsigned char dest[CEPH_CRYPTO_HMACSHA1_DIGESTSIZE]; // 20
+  char dest_str[output_size];
 
-  char dest_str[CEPH_CRYPTO_HMACSHA1_DIGESTSIZE * 2 + 1];
-  buf_to_hex((const unsigned char *)dest, sizeof(dest), dest_str);
+public:
+  TempURLSig() {}
```
stiopaa1: why empty constructor?

src/rgw/rgw_swift.cc
```diff
@@ -668,12 +693,17 @@ int authenticate_temp_url(RGWRados * const store, req_state * const s)
...
    * and old versions of RadosGW. The second item will have the prefix
    * of Swift API entry point removed. */
+
+  /* XXX can we search this ONCE? */
   const size_t pos = g_conf->rgw_swift_url_prefix.find_last_not_of('/') + 1;
-  const vector<string> allowed_paths = {
-    s->info.request_uri,
-    s->info.request_uri.substr(pos + 1)
+  boost::string_ref ref_uri = s->info.request_uri;
+  const vector<boost::string_ref> allowed_paths = {
```
stiopaa1: maybe std::array would be better here?

stiopaa1: maybe std::array would be better here?

src/rgw/rgw_swift.cc
```diff
@@ -683,25 +713,25 @@ int authenticate_temp_url(RGWRados * const store, req_state * const s)
...
+      for (const auto& method : allowed_methods) {
+        const char* local_sig = sig_helper.calc(temp_url_key, method, path,
+						temp_url_expires);
 
         ldout(s->cct, 20) << "temp url signature [" << temp_url_key_num
                           << "] (calculated): " << local_sig
                           << dendl;
 
-        if (local_sig != temp_url_sig) {
+        if (!!sig_helper.comp(temp_url_sig)) {
```
rzarzynski: I think that renaming the method to `is_equal_to` or similar would better reflect the intetion. Avoding the implicit to-bool conversion through double negation may be another benefit.

rzarzynski:
@mattbenjamin, @cbodley: I've ported this commit to the PR #8657. There are also further changes to reduce memory allocations (`std::array` and `boost::container::static_vector`, more usage of `boost::string_ref`).

src/rgw/rgw_swift.cc
rzarzynski: In https://github.com/ceph/ceph/pull/8657/commits/fbaf683a24752b61c4596c7305473259ed7d9afa this has been addressed through `boost::container::static_vector` of `boost::string_ref`. Though, I don't expect that switching to `string_ref` for such small strings would bring substantial improvement on those platforms which offer SSO. On the other hand GCC 4.8.4 on Trusty seems to be perfectly unaware about this trick. Anyway, eradicating the `vector` should be profitable.

src/rgw/rgw_swift.cc
mattbenjamin: @stiopaa1 no, not really

PR #8725, opened 2016-04-24
test/encoding/readable.sh: fix make check for autotool

liewegas:
I'm not sure what the cmake plan is here, but we can't
break autotools yet.

### DIFF
--- a/src/test/encoding/readable.sh
+++ b/src/test/encoding/readable.sh
@@ -1,5 +1,7 @@
 #!/bin/sh -e
 
+[ -z "$CEPH_ROOT" ] && CEPH_ROOT=..
+
 dir=$CEPH_ROOT/ceph-object-corpus
 
 set -e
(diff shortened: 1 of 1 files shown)


src/test/encoding/readable.sh
```diff
@@ -8,7 +10,7 @@ failed=0
 numtests=0
 pids=""
 
-myversion=`ceph-dencoder version`
+myversion=`./ceph-dencoder version`
```
tchaikov: we set `PATH=${CMAKE_RUNTIME_OUTPUT_DIRECTORY}:${CMAKE_SOURCE_DIR}/src:$ENV{PATH}` for ceph tests in cmake. so instead of reverting the change in this script, can we have
```sh
if which ceph-dencoder &> /dev/null; then
  CEPH_DENCODER=ceph-dencoder
else
  CEPH_DENCODER=./ceph-dencoder
fi
```

and replace all occurrences of `ceph-dencoder` with "$CEPH_DENCODER"?

liewegas: good idea.  switched it so that it checks the local dir first, though.

tchaikov:
lgtm. lemme test it locally.

tchaikov:
test passed locally. will fix 
```
src/test/cli/crushtool/check-overlapped-rules.t: failed
```
in another PR.

tchaikov:
@xiexingguo posted #8732 to address this failure.

PR #8726, opened 2016-04-24
rgw: fix printing wrong X-Storage-Url in Swift's TempAuth.

rzarzynski:
If the option "rgw_swift_account_in_url" is being set to true,
both user and subuser IDs will be present in the X-Storage-Url
header generated by the implementation of Swift API's TempAuth.
This patch rectifies this behavior and makes that only the user
identifier will be placed there.


rzarzynski:
@yehudasa: I've just added the `Fixes` tag.

yehudasa:
@rzarzynski is there a tracker issue opened for this one? if not, need to create one and add a reference in the commit message.

rzarzynski:
@yehudasa: yes, it is. I created [#15667](http://tracker.ceph.com/issues/15667) and amended the commit message.

oritwas:
@rzarzynski , can you put the tracker url not just the number

rzarzynski:
@oritwas: done. Just updated the commit.

PR #8727, opened 2016-04-24
osd: remove collection attr methods from ObjectStore interface

liewegas:
These have been deprecated since hammer.  We only kept them around so that
pre-hammer FileJournal entries could be replayed.  I think the risk of those
persisting has passed...


athanatos:
Yep.

src/osd/PG.cc
```diff
@@ -2914,35 +2909,6 @@ int PG::peek_map_epoch(ObjectStore *store,
...
-    values.clear();
-    keys.insert(ek);
-    store->omap_get_values(coll_t::meta(), legacy_infos_oid, keys, &values);
-    if (values.size() < 1) {
-      // probably bug 10617; see OSD::load_pgs()
-      return -1;
-    }
-    bufferlist::iterator p = values[ek].begin();
-    ::decode(cur_epoch, p);
   } else {
```
athanatos: Due to 11429, it's still possible (I think) to get orphaned pgs from the past.  I think we just need to return -1 in the else path to preserve the old behavior (OSD::load_pgs skips the pg in that case and emits a warning).

athanatos:
@liewegas Failed 11429.yaml, dropping for now.

src/osd/PG.cc
athanatos: Also, generally, I think skipping the pg in question is probably better than asserting?

liewegas:
Updated to return -1 in that case.

PR #7139, opened 2016-01-07
crush/CrushTester: check for overlapped rules

tchaikov:
list rules if there are more than one of them can be chosen by the CRUSH
for a given ruleset, pool type and a pool size. this is not entirely a
buggy crush map, but it might be worthy to get the attention of the user
that multiple rules overlap with each other, and only the first one will
be used by the CRUSH.

Signed-off-by: Kefu Chai <kchai@redhat.com>

PR #8732, opened 2016-04-25
test/cli: fix crush related ut failure

xiexingguo:
There is a ut failure introduced by https://github.com/ceph/ceph/pull/8502
and can by found at https://jenkins.ceph.com/job/ceph-pull-requests/4668/console,
for example.

### DIFF
--- a/src/test/cli/crushtool/check-overlapped-rules.t
+++ b/src/test/cli/crushtool/check-overlapped-rules.t
@@ -1,6 +1,13 @@
   $ crushtool -c "$TESTDIR/check-overlapped-rules.crushmap.txt" -o "$TESTDIR/check-overlapped-rules.crushmap"
+  ruleset '0' already defined in rule 'rule-r0'
+  
+  ruleset '0' already defined in rule 'rule-r0'
+  
+  ruleset '0' already defined in rule 'rule-r0'
+  
+  ruleset '0' already defined in rule 'rule-r0'
+  
+  ruleset '0' already defined in rule 'rule-r0'
+  
   $ crushtool -i "$TESTDIR/check-overlapped-rules.crushmap" --check
-  overlapped rules in ruleset 0: rule-r0, rule-r1, rule-r2
-  overlapped rules in ruleset 0: rule-r0, rule-r2, rule-r3
-  overlapped rules in ruleset 0: rule-r0, rule-r3
   $ rm -f "$TESTDIR/check-overlapped-rules.crushmap"


xiexingguo:
Passed the cli unit test:
```
make[3]: Leaving directory `/home/jenkins-build/build/workspace/ceph-pull-requests/src/gmock'
u=`ulimit -u` ; \
p=`expr 8 / 2` ; \
n=`expr $p \* 1024` ; \
if true && echo w --jobserver-fds=3,4 -j | grep --quiet -e -j && test $u -lt $n ; then \
	echo "ulimit -u is $u which is lower than $n = $p / 2 * 1024" ; \
	echo "If running make -j$p check you will likely exceed this limit" ; \
	echo "and the tests will fail in mysterious ways." ; \
	echo "Update /etc/security/limits.conf to increase the limit" ; \
	echo "or run make CHECK_ULIMIT=false -j4 check to override this safeguard." ; \
	exit 1 ; \
fi
PASS: src/test/run-cli-tests
```

xiexingguo:
@tchaikov Fixed a ut failure triggered by a recent change and passed, please get this merged. Thanks:)

tchaikov:
we added the check for the overlapped rules merged at #7139, and i don't think we should forbidden such a case as #8502 does.

and #8502 allows the generation of a partial compiled map, with status code of 0 returned by crushtool cli. this is not what we expect.

i'd suggest to revert it. or if we really want to prevent our user from compiling map with overlapped rules, we should have a test verifying that case, instead of covering it in this way.

xiexingguo:
> i'd suggest to revert it. or if we really want to prevent our user from compiling map with overlapped rules, we should have a test verifying that case, instead of covering it in this way.

It's okay with me, suit yourself:)

tchaikov:
@liewegas what do you think?

songbaisen:
@tchaikov  Hi! If the user define the overlapped rule.I will forbid him to generate crushmap.bin form crushmap.txt. I have not seen the generation of a partial compiled crushmap.bin.
I think the two #8502  and #7139 want to do the same thing.But #8502  forbid it earlier than #7139.
If we want to pass the unit test, we have two method as you said,One is revert my code. Two is modify 
the unit test here .I am okay whit all the two method.

tchaikov:
@songbaisen what scares me is the partial crushmap with return code 0. you always can test it using the test case addressed by this PR.

yeah, the intention is the same. but i'd prefer warn our user instead stopping him/her when compiling it.

songbaisen:
@tchaikov  Thank you for your reply.I am also surprise that the unit test can pass here.Because there will not have the check-overlapped-rules.crushmap file generate.But the below command can pass without error.
$ crushtool -i "$TESTDIR/check-overlapped-rules.crushmap" --check

In my opinion if the user defined the overlapped rule in crushmap.txt. It is just a buggy crushmap.txt and we should forbid the user to do that.Just personal advice.

tchaikov:
thanks @xiexingguo, as #8743 which reverts #8502 is merged. i am closing your fix.

xiexingguo:
@tchaikov You are welcome:)

PR #8733, opened 2016-04-25
mon/OSDMonitor: improve reweight_by_utilization() logic

xiexingguo:
By calling reweight_by_utilization() method, we are aiming at an evener result
of utilization among all osds. To achieve this, we shall decrease weights of
osds which are currently overloaded, and try to increase weights of osds which
are currently underloaded when it is possible.
However, we can't do this all at a time in order to avoid a massive pg migrations
between osds. Thus we introduce a max_osds limit to smooth the progress.

The problem here is that we have sorted the utilization of all osds in a descending
manner and we always try to decrease the weights of the most overloaded osds
since they are most likely to encounter a nearfull/full transition soon, but
we won't increase the weights from the most underloaded(least utilized by contrast)
at the same time, which I think is not quite reasonable.

Actually, the best thing would probably be to iterate over teh low and high osds in parallel, and do the ones that are furthest from the average first. 

### DIFF
--- a/src/mon/OSDMonitor.cc
+++ b/src/mon/OSDMonitor.cc
@@ -599,10 +599,13 @@ int OSDMonitor::reweight_by_utilization(int oload,
     util_by_osd.push_back(osd_util);
   }
 
-  // sort and iterate from most to least utilized
-  std::sort(util_by_osd.begin(), util_by_osd.end(), [](std::pair<int, float> l, std::pair<int, float> r) {
-    return l.second > r.second;
-  });
+  // sort by absolute deviation from the mean utilization,
+  // in descending order.
+  std::sort(util_by_osd.begin(), util_by_osd.end(),
+    [average_util](std::pair<int, float> l, std::pair<int, float> r) {
+      return abs(l.second - average_util) > abs(r.second - average_util);
+    }
+  );
 
   OSDMap::Incremental newinc;
 
--- a/src/mon/OSDMonitor.cc
+++ b/src/mon/OSDMonitor.cc
@@ -1709,6 +1709,13 @@ bool OSDMonitor::check_failures(utime_t now)
 
 bool OSDMonitor::check_failure(utime_t now, int target_osd, failure_info_t& fi)
 {
+  // already pending failure?
+  if (pending_inc.new_state.count(target_osd) &&
+      pending_inc.new_state[target_osd] & CEPH_OSD_UP) {
+    dout(10) << " already pending failure" << dendl;
+    return true;
+  }
+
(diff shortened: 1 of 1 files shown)


liewegas:
I wonder if we should flip it around so that we do the low-utilization devices first?

Actually, the *best* thing would probably be to iterate over teh low and high osds in parallel, and do the ones that are furthest from the average first.  That's probably comes down to writing a different sort function that goes by difference from average?

xiexingguo:
Okay, I'll give another try tomorrow:)

xiexingguo:
@liewegas Upgraded. Also write a little test program to verify this change, as below:
```
main()
{
  double average_util = 0;
  std::vector<std::pair<int, float> > util_by_osd;
  int i;

  for (i = 0; i < 10; i++) {
    std::pair<int, float> osd_util;
    osd_util.first = i;
    osd_util.second = (rand() % 100) / 100.0;
    util_by_osd.push_back(osd_util);
    average_util += osd_util.second;
  }

  average_util /= (float)i;

  // sort by absolute deviation from the mean utilization,
  // in descending order.
  std::sort(util_by_osd.begin(), util_by_osd.end(),
    [average_util](std::pair<int, float> l, std::pair<int, float> r) {
      return abs(l.second - average_util) > abs(r.second - average_util);
    }
  );

  cout << "average utilization: " << average_util << endl;
  for (std::vector<std::pair<int, float> >::const_iterator p = util_by_osd.begin();
    p != util_by_osd.end(); ++p)
     cout << "osd." << p->first << " util = " << p->second  << "(deviation = "
          << abs(p->second - average_util) << ")" << endl;
}
```     
                 
**average utilization: 0.637
osd.3 util = 0.15(deviation = 0.487)
osd.9 util = 0.21(deviation = 0.427)
osd.4 util = 0.93(deviation = 0.293)
osd.5 util = 0.35(deviation = 0.287)
osd.7 util = 0.92(deviation = 0.283)
osd.1 util = 0.86(deviation = 0.223)
osd.6 util = 0.86(deviation = 0.223)
osd.0 util = 0.83(deviation = 0.193)
osd.8 util = 0.49(deviation = 0.147)
osd.2 util = 0.77(deviation = 0.133)**

dvanders:
FTR, this was a goof on my part: the python prototype sorted the correct way (like you independently derived) but I forgot about this when we ported it. https://github.com/cernceph/ceph-scripts/blob/master/tools/crush-reweight-by-utilization.py#L89

smithfarm:
@liewegas The hammer backport of e7a3253 needs 3aa0ccef - is it OK to backport that one as well?

PR #8734, opened 2016-04-25
cleanup: use boost::algorithm::starts_with to replace our own implementations

flyd1005:
Don't Reinvent the Wheel.

Signed-off-by: Li Peng <lip@dtdream.com>

### DIFF
--- a/src/osd/OSD.h
+++ b/src/osd/OSD.h
@@ -1060,19 +1060,6 @@ public:
   ~OSDService();
 };
 
-struct C_OSD_SendMessageOnConn: public Context {
-  OSDService *osd;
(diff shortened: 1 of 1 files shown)


tchaikov:
lgtm.

flyd1005:
The build failure is strange. Jenkins can't work now?

tchaikov:
some tests failed. but the build passed.

guess jenkins is using an image with ext4, but we have not replaced that image with xfs since we deprecates ext4 for its lack of support for large xattr. it leads to the test failures. @ErwanAliasr1 have we tested the new image with xfs?

PR #8735, opened 2016-04-25
OSD: set peers need update before reset peers

Aran85:
otherwise the following call of maybe_update_heartbeat_peers() may not update peers immediately...

### DIFF
--- a/src/osd/OSD.cc
+++ b/src/osd/OSD.cc
@@ -6954,6 +6954,8 @@ void OSD::_committed_osd_maps(epoch_t first, epoch_t last, MOSDMap *m)
 
 	hbclient_messenger->mark_down_all();
 
+	heartbeat_set_peers_need_update();
+	last_heartbeat_resample = ceph_clock_now(cct);
 	reset_heartbeat_peers();
       }
     }


src/osd/OSD.cc
```diff
@@ -6954,6 +6954,8 @@ void OSD::_committed_osd_maps(epoch_t first, epoch_t last, MOSDMap *m)
 
 	hbclient_messenger->mark_down_all();
 
+	heartbeat_set_peers_need_update();
+	last_heartbeat_resample = ceph_clock_now(cct);
```
athanatos: start_waiting_for_healthy() above sets last_heartbeat_resample to utime_t().  reset_heartbeat_peers() in that case already does this.

athanatos:
I don't think this is the right fix.  start_waiting_for_healthy sets last_heartbeat_resample to utime_t().  maybe_reset_heartbeat_peers() has special handling for that value which does what you are doing here, and then goes on to actually repopulate the values which you *aren't* doing.  I think this fix will actually leave the map unpopulated until the first timeout making the problem much worse.

I think the right fix would be to call maybe_reset_heartbeat_peers() directly after reset_heartbeat_peers() here?  There may be a reason we can't do that while holding the OSD lock.  @liewegas ?

liewegas:
Yeah, I think a call to maybe_update_heartbeat_peers() will do it.

yuyuyu101:
@athanatos if we update heartbeat peers immediately, and because we will consume osdmap in the following codes. we may get improper hb peers. For example, when the host's cluster network is unavailable, the osds in this host should be marked down. And these osds get notified in _committed_osdmap and update hb peers, because the old osdmap is used to populate hb peers, the same host's osds will uses old osdmap's up osds to indicate we are healthy. So the network hole osds's will quickly pass OSD::_is_healthy and ready to boot. It will cause OSD filpping. So we must consume osdmap at firstly.

yuyuyu101:
Ever more, if we want to get suitable hb peers when ready to be healthy in heartbeat check way, we need to get latest osdmap which indicate enough failure in cluster now. So OSD_TICK period maybe better? Or I think we should pick up smarter hb peer strategy to collect accurate health status from other host even upper crush bucket level's osd.

athanatos:
@yuyuyu101 I'm not sure I'm following.  That call would be in _committed_osd_maps which happens after we've consumed the maps?

yuyuyu101:
I guess you refer to https://github.com/ceph/ceph/pull/8735/files#diff-fa6c2eba8356ae1442d1bf749beacfdfR6972 ?

athanatos:
@yuyuyu101 Ah, apparently I'm pretty confused. Is there an actual bug here?

yuyuyu101:
@athanatos 
1. I think we don't have an actual bug in this pr target to.
2. my previous comment mainly come up with a situation which caused by related codes, when one host is unavailable with cluster network, all osds are marked down. Then the first osd(any osd in the unavailable host)  entering OSD::_committed_osd_maps know itself down, then it consume the osdmap and update hb peers. But the osdmap may still mark other osds up(actually should be down now) in the same host up. Quickly the first osd boot again. So as the time grow, we always see one osd go down/up because of osmdap delay.
3. I think if we force wait "osd_heartbeat_grace" period, then we go to collect hb peers to be ready to boot. Or we could use a smart strategy to construct peer osd set in "OSD::maybe_update_heartbeat_peers", like set a "osd_hb_peer_subtree_level", to force pick up osd from other host and use a strict judgement in OSD::_is_healthy.

athanatos:
@yuyuyu101 
1) This PR should be closed then, right?
2/3) I'm not following.  What are you suggesting we delay for an osd_heartbeat_grace period?

yuyuyu101:
hmm, I mean we delay for a period until we have a stable enough osdmap?

yuyuyu101:
maybe it's not a deterministic condition. so maybe a smart hb peer strategy is better

yuyuyu101:
close this pr now

PR #8736, opened 2016-04-25
CMake: LINK_PRIVATE compatibility 

heroanxiaobo:
Use LINK_PRIVATE keyword instead of PRIVATE in librbd, for backward compatibility with cmake 2.8.11.

Signed-off-by: heroanxiaobo <heroanxiaobo@gmail.com>


tchaikov:
duplicated with #8715. closing

PR #8737, opened 2016-04-25
mds: ceph status outputs mds slow request for better monitoring

david-z:
Currently ceph status or ceph health will output osd slow request summary, but no mds slow request. It is not very friendly and timely for ops engineer to identify slow requests on mds or monitor mds status.

Within this PR, mds slow request summary will be printed out like osd, via ceph status, ceph health and ceph health detail.

<pre>
[ceph@xxx ~]$ ceph -s
    cluster 8dd227cf-5397-4a62-9e03-2accdc976bd7
     health HEALTH_WARN
            mds0: 7 slow requests are blocked > 30 sec
     ...

[ceph@xxx ~]$ ceph health detail
HEALTH_WARN mds0: 7 slow requests are blocked > 30 sec
mds0: 7 slow requests are blocked > 30 sec

[ceph@xxx ~]$ ceph health       
HEALTH_WARN mds0: 11 slow requests are blocked > 30 sec
</pre>

Fixes: #15589


gregsfortytwo:
@ukernel, please review.

@david-z, how far did you go in testing this?

david-z:
@gregsfortytwo 

I think our test is thorough enough. We have integrated this into our monitoring scripts, so far no problem found.

src/common/TrackedOp.cc
```diff
@@ -214,7 +214,11 @@ bool OpTracker::check_ops_in_flight(std::vector<string> &warning_vector)
   //store summary message
   warning_vector.push_back("");
 
-  int slow = 0;     // total slow
+  int _slow = 0;    // total slow
+  if (!slow)
+    slow = &_slow; 
+  else
+    *slow = _slow;  // start from 0 anyway
```
ukernel: I don't see reason why 'slow' needs to be input/output parameter. I think it's better to keep it as output parameter.

The rest change looks good

david-z: @ukernel  yes, you're right, but I chose 'slow' as a input/output parameter to avoid overloading check_ops_in_flight. Now I change it since you ask. Pls review the latest changes. Thanks.

ukernel:
Reviewed-By

gregsfortytwo:
I have the unconfirmed suspicion that this PR is somehow responsible for the failure in http://pulpito.ceph.com/yuriw-2016-05-24_20:35:01-fs-wip-yuri-testing---basic-smithi/212746

>RuntimeError: Unexpected health messages: [u'mds0: Client smithi007:0 failing to advance its oldest client/flush tid', u'mds0: 32628 slow requests are blocked > 30 sec']

in the test_client_oldest_tid test. It's not certain though; we'll see how some testing comes out excluding it and move on from there.

david-z:
@gregsfortytwo 

Yes, looks like this test needs some update. Where is the test case? Pls let me know if you need me to fix this test case.

jcsp:
@david-z the test in question is at https://github.com/ceph/ceph-qa-suite/blob/master/tasks/cephfs/test_client_limits.py#L160

If we're now getting two messages here you'll need to extend CephFSTestCase.wait_for_health to handle that, because it currently only supports waiting for a single health message.

You can try the test on a vstart cluster, see the docstring here: https://github.com/ceph/ceph-qa-suite/blob/master/tasks/cephfs/vstart_runner.py

david-z:
@jcsp Thanks for pointing me the doc and I can run the cephfs tests easily now. I am about to change wait_for_health and do some tests, but it takes quite some time to run one-round cephfs tests. Is there a way to run one specific test case, e.g, run test_client_limits case only?

jcsp:
@david-z yes, you can add the name of the test at the end of the command, with the full python path, like "tasks.cephfs.test_client_limits.TestClientLimits.test_client_oldest_tid".  You can also run all the tests in a module by using a shorter path like "tasks.cephfs.test_client_limits"

david-z:
@jcsp I have updated wait_for_health function in cephfs_test_case.py. Tests have been run on my local env and passed. Pls take a look. Thanks.

src/mds/MDSRank.cc
```diff
@@ -2388,13 +2389,17 @@ void MDSRank::create_logger()
 void MDSRank::check_ops_in_flight()
 {
   vector<string> warnings;
-  if (op_tracker.check_ops_in_flight(warnings)) {
+  int slow;
```
jcsp: This needs to be initialized to zero (error found by valgrind in testing)

src/common/TrackedOp.cc
```diff
@@ -175,6 +175,12 @@ void OpTracker::unregister_inflight_op(TrackedOp *i)
 
 bool OpTracker::check_ops_in_flight(std::vector<string> &warning_vector)
 {
+  int slow;
```
jcsp: Initialize this (unlike the case below this one doesn't matter so much, but still it's cleaner)

src/common/TrackedOp.h
```diff
@@ -121,6 +121,7 @@ class OpTracker {
    * @return True if there are any Ops to warn on, false otherwise.
    */
   bool check_ops_in_flight(std::vector<string> &warning_strings);
+  bool check_ops_in_flight(std::vector<string> &warning_strings, int &slow);
```
jcsp: Maybe you could make the `slow` argument a "int *slow=nullptr" so that you don't need two variants of the function.

jcsp:
@david-z please update to fix the uninitialized variables

src/common/TrackedOp.h
gregsfortytwo: I think the style guide only allows const references anyway; output parameters should all be pointers.

david-z: @jcsp @gregsfortytwo 

The initial commit I made was using pointers as your guys suggested. But Yan also suggested another way. You can see my changes and comments to change this to reference above. So which one do you think is better? Change it back to pointer?

Thanks.

jcsp: @david-z our style guide does ask for pointers for output arguments, so please do it that way (https://github.com/xiexingguo)

david-z:
@jcsp I have reset this PR to my initial one which should be same as we expected. Pls take a look. Thanks.

jcsp:
Looks good, will re-test

PR #8738, opened 2016-04-25
tool/cephfs: make EventOutput::binary() idempotent

xiexingguo:
We may passed a path in which has already been created. So we shall
tolerate the EEXIST error, which shall also make this funciton
idempotent.

Besides, we shall reset return code to errno on mkdir() system call
error for better tracing.


gregsfortytwo:
Looks good to me; @jcsp?

jcsp:
:+1:

yuriw:
tested http://pulpito.ceph.com/yuriw-2016-05-25_20:47:39-fs-wip-yuri-testing---basic-smithi/

PR #8739, opened 2016-04-25
mds,client: sort dentries in hash order 

ukernel:


liewegas:
I like this approach!  I was thinking we'd need to keep a string position in the dir_result_t for every frag and enumerate all frags in parallel.  I think that would also work, but this is much simpler.

gregsfortytwo:
Haven't started reviewing, but this fixes http://tracker.ceph.com/issues/15508, doesn't it? Since it's not mentioned anywhere here.

ukernel:
It's for http://tracker.ceph.com/issues/14271

src/mds/Server.cc
```diff
@@ -3435,12 +3436,21 @@ void Server::handle_client_readdir(MDRequestRef& mdr)
     mdcache->lru.lru_touch(dn);
   }
   
-  __u8 complete = (end && offset_str.empty());  // FIXME: what purpose does this serve
+  bool complete = false;
+  __u16 flags = 0;
+  if (end) {
+    flags = CEPH_READDIR_FRAG_END;
+    complete = offset_str.empty(); // FIXME: what purpose does this serve
```
gregsfortytwo: I realize you didn't introduce this comment, but let's try and figure it out. It looks to me as though this exists from 60ec1cbe0ecc1e57d22c0eccffcb3b30faf7f5b9, and was never referenced in the userspace Client at all. Does the kernel look at it?

gregsfortytwo:
In "client: save readdir result into dir_result_t directly", s/MedaRequest/MetaRequest/ if anything else calls for a respin.

src/client/Client.cc
```diff
@@ -6931,13 +6943,9 @@ int Client::_readdir_get_frag(dir_result_t *dirp)
   req->set_filepath(path); 
   req->set_inode(diri.get());
   req->head.args.readdir.frag = fg;
-  if (dirp->last_name.length()) {
+  if (dirp->last_name.length())
```
gregsfortytwo: I know it's more verbose, but in general we really prefer to use braces even for one-line if functions; it makes updating them much safer and noticing control flows a lot more obvious.

src/mds/MDCache.cc
```diff
@@ -5753,6 +5754,8 @@ void MDCache::opened_undef_inode(CInode *in) {
   dout(10) << "opened_undef_inode " << *in << dendl;
   rejoin_undef_inodes.erase(in);
   if (in->is_dir()) {
+    // FIXME: re-hash dentries if necessary
+    assert(in->inode.dir_layout.dl_dir_hash == g_conf->mds_default_dir_hash);
```
gregsfortytwo: We would need to re-hash in what cases? Why can't we fix that now?

gregsfortytwo: Oh, if it's not the default, got it.

src/mds/Server.cc
ukernel: No, kernel does not look at it

gregsfortytwo: So is this something we should just drop entirely, then? Was it looked at in older kernels, maybe?

ukernel: I don't think old kernel use it

ukernel:

jcsp:
@gregsfortytwo did the most recent commits address your comments?

gregsfortytwo:
Yeah, but I never made it through all the commits. :( It's in my queue...

jcsp:
Tested here (failures accounted for elsewhere) http://pulpito.ceph.com/jspray-2016-05-25_19:00:36-fs-wip-jcsp-testing-20160526---basic-mira

Ready to merge when @gregsfortytwo (who is better placed than me to do the Reviewed-by because of spending time investigating the issue originally) has time to finish review.

jcsp:
@gregsfortytwo ping, could you review please?

src/client/Client.cc
```diff
@@ -6890,26 +6905,42 @@ void Client::fill_dirent(struct dirent *de, const char *name, int type, uint64_t
...
+  }
+
+  // advance
+  fg = fg.next();
+  ldout(cct, 10) << "_readdir_next_frag advance from " << dirp->buffer_frag << " to " << fg << dendl;
+
+  if (dirp->hash_order()) {
+    int64_t new_offset = dir_result_t::make_fpos(fg.value(), 2, true);
+    if (dirp->offset < new_offset) // don't decrease offset
+      dirp->offset = new_offset;
```
gregsfortytwo: Under what scenario could new_offset be less than dirp->offset? I don't really see how that could happen without us also having potentially skipped entries in the next frag. (Mostly, I don't see how that could happen, since we know we're going in hash order.)

src/client/Client.cc
```diff
@@ -1187,6 +1197,26 @@ void Client::insert_readdir_results(MetaRequest *request, MetaSession *session,
...
+	  dirp->start_shared_gen == diri->shared_gen) {
+	if (dirp->cache_index == dir->readdir_cache.size()) {
+	  if (i == 0) {
+	    assert(!dirp->inode->is_complete_and_ordered());
+	    dir->readdir_cache.reserve(dirp->cache_index + numdn);
+	  }
+	  dir->readdir_cache.push_back(dn);
+	} else if (dirp->cache_index < dir->readdir_cache.size()) {
+	  if (dirp->inode->is_complete_and_ordered())
+	    assert(dir->readdir_cache[dirp->cache_index] == dn);
```
gregsfortytwo: Did you mean to leave in this debug code?

src/client/Client.cc
```diff
@@ -2891,6 +2913,9 @@ Dentry* Client::link(Dir *dir, const string& name, Inode *in, Dentry *dn)
     if (in->is_dir() && !in->dn_set.empty()) {
       Dentry *olddn = in->get_first_parent();
       assert(olddn->dir != dir || olddn->name != name);
+      Inode *old_diri = olddn->dir->parent_inode;
+      old_diri->dir_release_count++;
+      clear_dir_complete_and_ordered(old_diri, true);
```
gregsfortytwo: I'm missing why this needs to get added (It doesn't look like it's being moved from elsewhere?). Is it another bug fix, or a result of the change that I'm not understanding?

gregsfortytwo:
Sorry it took so long; this looks pretty good (just a few queries that are probably not issues).

Given the change in listing though, we should definitely run this against our upgrade tests.

src/client/Client.cc
ukernel: To make sure old parent directory's readdir cache is invalid. (I add this code to all places (except snapdir) that Client::unlink is called)

src/client/Client.cc
ukernel: yes

src/client/Client.cc
ukernel: For example:

This can happen after seekdir (forward seek).

src/client/Client.cc
gregsfortytwo: Right, but we needed to do that before as well, didn't we? So why did we add an invalidate step instead of expanding an existing one?

gregsfortytwo:
I built a test branch with this and #9374 and about half the runs are failing on DISCONNECTED inodes not getting cleaned up on shutdown. I'm going to try and investigate some before bed tonight, but if you have ideas @ukernel, let me know. Given that it passed in the last testing round, but it was a while ago, I presume a lifetime has changed somewhere (or the test build was done prior to the last patch being added).
http://pulpito.ceph.com/gregf-2016-06-08_15:07:14-fs-greg-fs-testing-68---basic-mira/

jcsp:
Removing needs-qa until existing failures are understood

ukernel:
On Thu, Jun 9, 2016 at 9:41 AM, Gregory Farnum <notifications@github.com>
wrote:

> I built a test branch with this and #9374
> <https://github.com/ceph/ceph/pull/9374> and about half the runs are
> failing on DISCONNECTED inodes not getting cleaned up on shutdown. I'm
> going to try and investigate some before bed tonight, but if you have ideas
> @ukernel <https://github.com/ukernel>, let me know. Given that it passed
> in the last testing round, but it was a while ago, I presume a lifetime has
> changed somewhere (or the test build was done prior to the last patch being
> added).
>
> http://pulpito.ceph.com/gregf-2016-06-08_15:07:14-fs-greg-fs-testing-68---basic-mira/
>

I think the failures were caused by Fh leaks (Because, in the log,
_release_fh() is not followed by put_inode()). But I can't reproduce this
issue using your branch, I can't find any bug in Fh reference tracking
neither.

gregsfortytwo:
Hahaha, the leak is not because of this branch. Turns out enabling readahead exposed a leak in the readahead code. :D

gregsfortytwo:
Hmm, during my attempt at bisecting the Readahead leak error, one of my runs failed in libcephfs/test.sh's LibCephFS.DirLs test. This ran against master+this PR, but excluding the very last (simultaneous readdir) fix.
http://pulpito.ceph.com/gregf-2016-06-10_01:23:22-fs-greg-fs-testing-69-1---basic-mira/249932/

Obviously that final PR changed the readdir internals quite a lot so if it doesn't turn up again we hopefully don't need to worry about it, but I'm noting it here for a little more investigation or if it sparks any ideas for anybody.

gregsfortytwo:
ahttp://pulpito.ceph.com/gregf-2016-06-10_19:20:53-fs-greg-fs-testing-610---basic-mira/
All failures attributed to the AsyncMessenger compatibility issues.

PR #8742, opened 2016-04-25
os/bluestore: record released extents into bluestore_wal_transaction_t.

majianpeng:


majianpeng:
@liewegas . Could you review those?

liewegas:
Good catch!

PR #8744, opened 2016-04-25
common/mem-leak: Pick address should free ifaddrs when meets error.

JiYou:
When pick up address for ceph-{mon,osd,mds} daemons, when meets error, should release `struct ifaddrs`.

Signed-off-by: You Ji <youji@ebay.com>


liewegas:
Note that these cleanups don't really make a difference--since the process is exiting anyway the only thing we accomplish by freeing memory is making a potential leak checker happy.

JiYou:
ok. Thanks @liewegas

PR #8747, opened 2016-04-25
client: report root's quota in statfs (#15599)

jcsp:
When user is mounted a quota-restricted inode
as the root, report that inode's quota status
as the filesystem statistics in statfs.

This allows us to have a fairly convincing illusion
that someone has a filesystem to themselves, when
they're really mounting a restricted part of
the larger global filesystem.

Fixes: http://tracker.ceph.com/issues/15599

### DIFF
--- a/doc/cephfs/client-auth.rst
+++ b/doc/cephfs/client-auth.rst
@@ -43,6 +43,25 @@ for example, to restrict client ``foo`` to ``mnt/bar`` directory, we will use. :
 
 ./ceph-fuse -n client.foo mnt -r /bar
 
+Free space reporting
+--------------------
+
+By default, when a client is mounting a sub-directory, the used space (``df``)
+will be calculated from the quota on that sub-directory, rather than reporting
+the overall amount of space used on the cluster.
+
+If you would like the client to report the overall usage of the filesystem,
+and not just the quota usage on the sub-directory mounted, then set the
+following config option on the client:
+
+::
+
+    client quota df = false
+
+If quotas are not enabled, or no quota is set on the sub-directory mounted,
+then the overall usage of the filesystem will be reported irrespective of
+the value of this setting.
+
 OSD restriction
 ===============
 
--- a/src/client/Client.cc
+++ b/src/client/Client.cc
@@ -8752,19 +8752,11 @@ int Client::statfs(const char *path, struct statvfs *stbuf)
   tout(cct) << "statfs" << std::endl;
 
   ceph_statfs stats;
-
-  Mutex lock("Client::statfs::lock");
-  Cond cond;
-  bool done;
-  int rval;
-
-  objecter->get_fs_stats(stats, new C_SafeCond(&lock, &cond, &done, &rval));
+  C_SaferCond cond;
+  objecter->get_fs_stats(stats, &cond);
 
   client_lock.Unlock();
-  lock.Lock();
-  while (!done)
-    cond.Wait(lock);
-  lock.Unlock();
+  int rval = cond.wait();
   client_lock.Lock();
 
   memset(stbuf, 0, sizeof(*stbuf));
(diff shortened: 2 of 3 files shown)


jcsp:
Tested by https://github.com/ceph/ceph-qa-suite/pull/974

liewegas:
lgtm

src/common/config_opts.h
```diff
@@ -382,6 +382,7 @@ OPTION(client_notify_timeout, OPT_INT, 10) // in seconds
 OPTION(osd_client_watch_timeout, OPT_INT, 30) // in seconds
 OPTION(client_caps_release_delay, OPT_INT, 5) // in seconds
 OPTION(client_quota, OPT_BOOL, false)
```
xiaoxichen: would we like to make this true by default?

doc/cephfs/client-auth.rst
```diff
@@ -43,6 +43,25 @@ for example, to restrict client ``foo`` to ``mnt/bar`` directory, we will use. :
...
+and not just the quota usage on the sub-directory mounted, then set the
+following config option on the client:
+
+::
+
+    client quota df = false
+
+If quotas are not enabled, or no quota is set on the sub-directory mounted,
+then the overall usage of the filesystem will be reported irrespective of
+the value of this setting.
```
xiaoxichen: could we recursive to parent dir(if MDS cap allow us to ), to see if any quota is setted?
Otherwise it's actually a bug that user can mount a subdir (say /volume/tenantA/FeelFREE) to bypass quota enforced in /volume/tenantA.

liewegas: The client already figures out the ancestor with the quota set, and that's 
what the statfs code is using.  The documentation is just not clear.

src/common/config_opts.h
jcsp: Maybe, but it's a separate thing from the change in this PR

jcsp:
Test run here:
http://pulpito.ceph.com/jspray-2016-04-30_17:56:02-fs-wip-jcsp-testing---basic-mira/

Three valgrind faliures from statfs, but I can't see how I'm doing anything different with the syscall's args, so I'm stumped for the moment (anyone?):
```
<error>
  <unique>0x0</unique>
  <tid>24</tid>
  <kind>SyscallParam</kind>
  <what>Syscall param writev(vector[...]) points to uninitialised byte(s)</what>
  <stack>
    <frame>
      <ip>0xC4A0460</ip>
      <obj>/lib/x86_64-linux-gnu/libc-2.19.so</obj>
      <fn>writev</fn>
      <dir>/build/eglibc-3GlaMS/eglibc-2.19/misc/../sysdeps/unix/sysv/linux</dir>
      <file>writev.c</file>
      <line>54</line>
    </frame>
    <frame>
      <ip>0xA7918E3</ip>
      <obj>/lib/x86_64-linux-gnu/libfuse.so.2.9.2</obj>
    </frame>
    <frame>
      <ip>0xA79366C</ip>
      <obj>/lib/x86_64-linux-gnu/libfuse.so.2.9.2</obj>
    </frame>
    <frame>
      <ip>0xA7936AB</ip>
      <obj>/lib/x86_64-linux-gnu/libfuse.so.2.9.2</obj>
    </frame>
    <frame>
      <ip>0xA795C5B</ip>
      <obj>/lib/x86_64-linux-gnu/libfuse.so.2.9.2</obj>
      <fn>fuse_reply_statfs</fn>
    </frame>
    <frame>
      <ip>0x2A1D5A</ip>
      <obj>/usr/bin/ceph-fuse</obj>
      <fn>fuse_ll_statfs(fuse_req*, unsigned long)</fn>
      <dir>/srv/autobuild-ceph/gitbuilder.git/build/out~/ceph-10.2.0-339-g1eb68e7/src/client</dir>
      <file>fuse_ll.cc</file>
      <line>692</line>
    </frame>
    <frame>
      <ip>0xA795C87</ip>
      <obj>/lib/x86_64-linux-gnu/libfuse.so.2.9.2</obj>
    </frame>
    <frame>
      <ip>0xA79522A</ip>
      <obj>/lib/x86_64-linux-gnu/libfuse.so.2.9.2</obj>
    </frame>
    <frame>
      <ip>0xA791E48</ip>
      <obj>/lib/x86_64-linux-gnu/libfuse.so.2.9.2</obj>
    </frame>
    <frame>
      <ip>0xADDD181</ip>
      <obj>/lib/x86_64-linux-gnu/libpthread-2.19.so</obj>
      <fn>start_thread</fn>
      <dir>/build/eglibc-3GlaMS/eglibc-2.19/nptl</dir>
      <file>pthread_create.c</file>
      <line>312</line>
    </frame>
    <frame>
      <ip>0xC4A947C</ip>
      <obj>/lib/x86_64-linux-gnu/libc-2.19.so</obj>
      <fn>clone</fn>
      <dir>/build/eglibc-3GlaMS/eglibc-2.19/misc/../sysdeps/unix/sysv/linux/x86_64</dir>
      <file>clone.S</file>
      <line>111</line>
    </frame>
  </stack>
  <auxwhat>Address 0x178604a8 is on thread 24's stack</auxwhat>
  <auxwhat>in frame #4, created by fuse_reply_statfs (???)</auxwhat>
</error>

```

gregsfortytwo:
Please check this out @ukernel

ukernel:
I still think we should use get_quota_root() instead of root_ancestor. For example, current code does not work for following case:

ceph-fuse /mnt/ceph
mkdir -p /mnt/ceph/1/2/3
setfattr -n ceph.quota.max_bytes -v 104857600 /mnt/ceph/1/2
umount /mnt/ceph
ceph-fuse -r /1/2/3 /mnt/ceph

jcsp:
@ukernel oh, I see what you mean now.  You're talking about the case where a client is doing a subtree mount, but he isn't limited by "path=" auth caps, so it is possible for us to look up his parents' quotas.

Seems reasonable to support that case, by just calling get_quota_root on root_ancestor -- I've updated the code.

jcsp:
I'm not sure what was confusing valgrind, but this latest version isn't raising the warning any more (locally)

src/client/Client.cc
```diff
@@ -8779,16 +8771,44 @@ int Client::statfs(const char *path, struct statvfs *stbuf)
...
   stbuf->f_favail = -1;
   stbuf->f_fsid = -1;       // ??
   stbuf->f_flag = 0;        // ??
   stbuf->f_namemax = NAME_MAX;
 
+  // Usually quota_root will == root_ancestor, but if the mount root has no
+  // quota but we can see a parent of it that does have a quota, we'll
+  // respect that one instead.
+  assert(root_ancestor != nullptr);
+  Inode *quota_root = get_quota_root(root_ancestor);
```
ukernel: root_ancestor is  real filesystem root (inode number 1). I think this should be get_quota_root(root)

jcsp: Doh, of course.  Fixed.

ukernel:
Reviewed-by

PR #8748, opened 2016-04-26
OSD: misc optimization for advance_pg

mslovy:


src/osd/ReplicatedPG.cc
```diff
@@ -11811,7 +11811,8 @@ void ReplicatedPG::agent_setup()
 void ReplicatedPG::agent_clear()
 {
   agent_stop();
-  agent_state.reset(NULL);
+  if (agent_state)
+    agent_state.reset(NULL);
```
tchaikov: what's the benefit of this change?

src/osd/OSD.cc
```diff
@@ -8847,8 +8847,8 @@ void OSD::process_peering_events(
        ++i) {
     set<boost::intrusive_ptr<PG> > split_pgs;
     PG *pg = *i;
-    pg->lock_suspend_timeout(handle);
     curmap = service.get_osdmap();
+    pg->lock_suspend_timeout(handle);
```
tchaikov: the commit message is not accurate. as `service.get_osdmap()` does not update osdmap. it returns it instead.

src/osd/ReplicatedPG.cc
mslovy: boost scope_ptr always reset the inner by calling a swap (this_type(p).swap(*this);) as well as some sanity checks. Therefore, if it is already null, the clear process is not needed

src/osd/OSD.cc
mslovy: updated

src/osd/OSD.cc
```diff
@@ -1333,6 +1332,7 @@ OSDMapRef OSDService::try_get_map(epoch_t epoch)
   if (epoch > 0) {
     dout(20) << "get_map " << epoch << " - loading and decoding " << map << dendl;
     bufferlist bl;
+    Mutex::Locker l(map_cache_lock);
```
tchaikov: i think the reason why we have a grand `map_cache_lock` is to make sure the consistency of map caches (`map_cache`, `map_bl_cache` and `map_bl_inc_cache`) maintained by OSDServices. as we take them a single cache when trying to   read an osdmap. and as you know, both SharedLRU and SimpleLRU have its own lock.

so we can safely drop this lock if we don't care about performance at all.

but if the performance comes into the picture, we might want to weigh the cost of acquiring the grand lock and the cost of acquiring the locks in caches and possibly cost of unnecessary loading the osdmaps from ObjectStore and decoding them.

so i don't think we can draw a conclusion that this change will be a win in performance just by inspecting the code.

do you have a benchmark to support your change?

src/osd/ReplicatedPG.cc
tchaikov: ic. just not sure if agent_clear() is a hotspot and if it's worth the change.

src/osd/OSD.cc
mslovy: But first of all, map_cache and (map_bl_cache, map_bl_inc_cache), are exactly different cache(may consider cache in different layer), map_cache cache the OSDmap while map_bl_cache(map_bl_inc_cache) cache the encoded OSDMap, based on the logic try_get_map() also tells that something in map_bl_cache may not exist in map_cache. So there is no need to make sure the consistency among map caches (map_cache, map_bl_cache and map_bl_inc_cache), any violated cases here?  I think the reason why we use map_bl_cache here is to prevent a dirty read OSDMap from FileStore as the Update Map transaction write into journal but not apply to the FileStore.
Yeah, I agree that even if we remove the Locker here, the contention still exists and inside SharedLRU. 
I think the issue here is that when several PG running peering process at the same time(this based on the number of thread in osd_tp), advance_map will lead to a heavy map_lock contention. However, once map cached, all other PGs will just a read operation from cache. Also, map advance is consective so that lots of PGs will used a certain range of maps (from oldest_epoch to newest_epoch), so a LRU cache here may not be the best one and we can instead it, like (simple map<epoch, OSDMap> with a RWLock and always trim the OSDMap with oldest epoch?). This would improve the peering process speed and reduce the total Blocked IO time.
And also, when a new OSD start, the best strategy is, waiting all PG advance to the current map epoch and calculate  the past_interval and so on. And then, send out the Boot message (MOSDBoot), otherwise, a Boot will cause osdmap change, and trigger the peering process but still wait for other osd finishing advance pg, which will block client IO for a long time. The best option may use a set<pg_t> wait_for_advance_map and removing the item when each pg advance to the newest OSDMap. A callback is used to check whether wait_for_advance_map is empty and send out the Boot message to Monitor.
so I think I should reset this commit in this PR, and include a more general optimization in another PR @tchaikov , right?
What about the strategy above (1. a simple map_cache not using LRU  2. Boot the OSD when all PG is really ready for), @liewegas ?

athanatos:
It seems very unlikely that either of these changes will make a measurable difference.  They might make sense as part of other changes in this part of the code, but I don't want to merge them on their own.  Feel free to reopen with performance measurements if there is evidence to the contrary.

PR #8752, opened 2016-04-26
vstart_wrapper.sh: allow a lot of the test scripts to run standalone

wjwithagen:
Fixing fallout of a previous Cmake commit that introduced CEPH_.... variables
but failed to keep backwards compatability

### DIFF
--- a/src/test/vstart_wrapper.sh
+++ b/src/test/vstart_wrapper.sh
@@ -16,6 +16,18 @@
 # GNU Library Public License for more details.
 #
 
+if [ -z "$CEPH_ROOT" ] ; then
+    CEPH_ROOT=`readlink -f $(dirname $0)/../..`
+    if [ -d $CEPH_ROOT/build ] ; then
+       # asume we are in Cmake build environment
+        CEPH_BIN=$CEPH_ROOT/build/bin
+        CEPH_LIB=$CEPH_ROOT/build/lib
+    else
+        CEPH_BIN=$CEPH_ROOT/src
+        CEPH_LIB=$CEPH_ROOT/.libs/src
+    fi
+fi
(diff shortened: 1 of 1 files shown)


src/test/vstart_wrapper.sh
```diff
@@ -16,6 +16,7 @@
 # GNU Library Public License for more details.
 #
 
+CEPH_ROOT=${CEPH_ROOT:-..}
```
alimaredia: @wjwithagen wouldn't we want "../.." to be the default relative path instead of ".." in order to reach the qa directory?

wjwithagen: @alimaredia 
I suggested this under the assumption that this is run from the `....../src` directory.
And then the way to `qa/`is just `..`
If you want to run it from the root of the whole ceph-directory, you'd need `../..`
What you would not do is run it whilest standing in src/test.

So in that sense it could do with a lot more code to actually determine where `$CEPHOOT` should be. But it could be done with a range of test to see where `./qa` actually is.

wjwithagen:
retest this

alimaredia:
@tchaikov how does this look to you?

src/test/vstart_wrapper.sh
tchaikov: i'd recommend fall back to `$(readlink -f $(dirname $0)/../..)`, what do you think?

wjwithagen: @tchaikov @alimaredia
I'm going to redo this over the weekend.
Probably a combi of multiple direcoty selection and some readline

wjwithagen:
@tchaikov 
Can you have a look at this?

PR #8756, opened 2016-04-26
rgw: introduce rgw_list_bucket_min_readahead parameter for bucket lis…

Abhishekvrshny:
…ting from rados

Signed-off-by: Abhishek Varshney <abhishek.varshney@flipkart.com>

### DIFF
--- a/src/common/config_opts.h
+++ b/src/common/config_opts.h
@@ -1405,6 +1405,8 @@ OPTION(rgw_period_push_interval_max, OPT_DOUBLE, 30) // maximum interval after e
 
 OPTION(rgw_swift_versioning_enabled, OPT_BOOL, false) // whether swift object versioning feature is enabled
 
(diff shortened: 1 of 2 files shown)


Abhishekvrshny:
Bucket listing turns out to be quite slow when a low value is specified to the URL parameter `max-keys`. For example, the s3a hadoop connector species `max-keys` parameter to be 1 for a certain operation [1]. The expectation of the client here is to get result set with 1 value. Radosgw, in turn, percolates this value down to rados and fetches a key one by one in `RGWRados::Bucket::List::list_objects` function before checking for delimiter etc. This turns out to be highly inefficient and thus the client faces time-outs. 

It would have been better if a config option could be provided to avoid such issues, like, a minimum readahead value for listing objects from rados.

[1] https://github.com/apache/hadoop/blob/trunk/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java#L1012

PR #8759, opened 2016-04-26
osdc/Objecter: fix race condition for sortbitwise flag detection

xiexingguo:
During list_objects/list_nobjects(), we are checking osdmap's
sortbitwise flag setting without holding rwlock in hand, which
is a potential race condition.


xiexingguo:
@liewegas

liewegas:
Good catch!

PR #8763, opened 2016-04-26
os/filestore: Make sure str-pointer is not 0x0 when referenced

wjwithagen:
ceph-osd crashes on reference to where if it is not get_index or lfn_find.
So init with an "empty" indicator "()

### DIFF
--- a/src/os/filestore/FileStore.cc
+++ b/src/os/filestore/FileStore.cc
@@ -4929,7 +4929,7 @@ int FileStore::omap_get_values(const coll_t& _c, const ghobject_t &hoid,
   const coll_t& c = !_need_temp_object_collection(_c, hoid) ? _c : _c.get_temp();
   dout(15) << __func__ << " " << c << "/" << hoid << dendl;
   Index index;
-  const char *where = 0;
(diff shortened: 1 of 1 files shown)


xiexingguo:
Good catch :-)

xiexingguo:
Nit, you may also want to append one more line below https://github.com/ceph/ceph/pull/8763/files#diff-5dd5cdee760fcb36d7bd0ab2144e09b8L4949, like this :
```c++
   r = object_map->get_values(hoid, keys, out);
   if (r < 0 && r != -ENOENT) {
     assert(!m_filestore_fail_eio || r != -EIO);
     where = "(get_values)";
     goto out;
   }
```

wjwithagen:
Thanx..
I'll take you suggestion in an extra commit.

There are more places where 'const char *var = 0` is present.
I've checked several, and they seem harmless for the time being. Until somebody adds a code path that does not set the pointer to a valid string.
So I'm very much inclined to either remove the assignment or replace with 
    const char *var = "";
That is much more defensive than just bluntly initialising them with NULL.

PR #8765, opened 2016-04-26
rgw: extend SIGHUP handler for realm reconfiguration

cbodley:
Adds a signal handler to trigger RGWRealmReloader to shut down and recreate RGWRados to reload the realm configuration and restart multisite sync.

### DIFF
--- a/src/rgw/rgw_main.cc
+++ b/src/rgw/rgw_main.cc
@@ -177,6 +177,19 @@ static RGWRESTMgr *set_logging(RGWRESTMgr *mgr)
 void intrusive_ptr_add_ref(CephContext* cct) { cct->get(); }
 void intrusive_ptr_release(CephContext* cct) { cct->put(); }
 
+RGWRealmReloader *preloader = NULL;
+
+static void reloader_handler(int signum)
+{
+  if (preloader) {
+    bufferlist bl;
+    bufferlist::iterator p = bl.begin();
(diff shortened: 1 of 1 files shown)


yehudasa:
@cbodley I think SIGUSR2 is used internally in valgrind, so maybe we'd want to hook this with SIGHUP instead? tie it into the generic ceph configuration reloader somehow

cbodley:
It looks like we could install a different handler for SIGHUP, which calls both `RGWRealmReloader::handle_notify()` and `g_ceph_context->reopen_logs()`. I'll give that a try.

theanalyst:
is it also possible to hook the `ceph tell` or  `injectargs` into RGW somehow?

yehudasa:
@theanalyst no, you cannot use ceph tell or injectargs with client ceph apps.

cbodley:
@yehudasa updated to piggy-pack realm reconfiguration on SIGHUP. the default `sighup_handler()` only calls `g_ceph_context->reopen_logs()` though.. was the goal to extend this to reload ceph.conf too? i wasn't able to find any examples of other daemons doing this, and i don't think there's a thread-safe way to do it (the config observers look like a step in this direction, but they haven't been fully adopted)

yehudasa:
@cbodley eventually we'd probably want to be able to update all config via sighup, but this is fine for now I think.

PR #8766, opened 2016-04-26
rgw: upgrade default zonegroup and set correct zone[group] id

yehudasa:
Fixes: http://tracker.ceph.com/issues/15597


src/rgw/rgw_rados.cc
```diff
@@ -3361,9 +3361,9 @@ int RGWRados::replace_region_with_zonegroup()
   /* create zonegroups */
   for (iter = regions.begin(); iter != regions.end(); ++iter)
   {
-    /* read region info default has no data */
-    if (*iter != default_zonegroup_name){
+    if (1 /* || *iter != default_zonegroup_name */) {
```
theanalyst: we could drop the comment inside the if

PR #8769, opened 2016-04-26
rgw: deleted zones are removed from period short_zone_ids

cbodley:
Fixes: http://tracker.ceph.com/issues/15618


cbodley:
it turns out that we do want to keep deleted zones in the period, so we can check new zones against them for hash collisions

PR #8770, opened 2016-04-26
cmake: build fixes

alimaredia:
This PR makes all the tests that need environment variables run in `make check` have them set automatically as long as the tests are being executed out of the proper directory.


alimaredia:
@cbodley here's the PR

trociny:
lgtm

src/test/cephtool-test-rados.sh
```diff
@@ -14,6 +14,9 @@
 # MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 # GNU Library Public License for more details.
 #
+source $(dirname $0)/detect-build-env-vars.sh
+CEPH_ROOT=${CEPH_ROOT:-..}
```
cbodley: shouldn't need to set CEPH_ROOT in scripts that run detect-build-env-vars.sh

alimaredia:
@dzafman This is the PR I was talking about in the core meeting

dzafman:
@alimaredia The detect-build-env-vars.sh looks reasonable.  I had a version that used $(pwd) to get an absolute path instead relative paths but I can't say that one is better than the other.

liewegas:
@alimaredia make check test cluster tests are still failing, see jenkins output

alimaredia:
@dachary or anyone else, since this isn't causing Jenkins problems, can this be good to merge?

PR #8771, opened 2016-04-26
pybind: configurable cephfs_vol_client prefix.

xiaoxichen:
Which is useful if we would like to have multiple backends
in manila and potentially set different layout based on prefix.

### DIFF
--- a/src/pybind/ceph_volume_client.py
+++ b/src/pybind/ceph_volume_client.py
@@ -203,17 +203,19 @@ class CephFSVolumeClient(object):
     """
 
     # Where shall we create our volumes?
-    VOLUME_PREFIX = "/volumes"
     POOL_PREFIX = "fsvolume_"
-    POOL_NS_PREFIX = "fsvolumens_"
+    DEFAULT_VOL_PREFIX = "/volumes"
+    DEFAULT_NS_PREFIX = "fsvolumens_"
 
-    def __init__(self, auth_id, conf_path, cluster_name):
+    def __init__(self, auth_id, conf_path, cluster_name, volume_prefix=None, pool_ns_prefix=None):
         self.fs = None
         self.rados = None
         self.connected = False
         self.conf_path = conf_path
         self.cluster_name = cluster_name
         self.auth_id = auth_id
+        self.volume_prefix = volume_prefix if volume_prefix else self.DEFAULT_VOL_PREFIX
+        self.pool_ns_prefix = pool_ns_prefix if pool_ns_prefix else self.DEFAULT_NS_PREFIX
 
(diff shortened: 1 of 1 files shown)


gregsfortytwo:
ping @jcsp, unless @ajarr wants to check it out

src/pybind/ceph_volume_client.py
```diff
@@ -203,17 +203,19 @@ class CephFSVolumeClient(object):
...
 
-    def __init__(self, auth_id, conf_path, cluster_name):
+    def __init__(self, auth_id, conf_path, cluster_name, volume_prefix=None, pool_ns_prefix=None):
         self.fs = None
         self.rados = None
         self.connected = False
         self.conf_path = conf_path
         self.cluster_name = cluster_name
         self.auth_id = auth_id
+        self.volume_prefix = volume_prefix if volume_prefix else DEFAULT_VOL_PREFIX
```
ajarr: s/DEFAULT_VOL_PREFIX/self.DEFAULT_VOL_PREFIX/

src/pybind/ceph_volume_client.py
```diff
@@ -203,17 +203,19 @@ class CephFSVolumeClient(object):
...
+        self.pool_ns_prefix = pool_ns_prefix if pool_ns_prefix else DEFAULT_NS_PREFIX
```
ajarr: s/DEFAULT_NS_PREFIX/self.DEFAULT_NS_PREFIX/

ajarr:
Maybe it would be helpful to refer to tracker issue that this commit "fixes/partially fixes"
http://tracker.ceph.com/issues/15417
I'm not sure about commit tags for patches adding new features.

ajarr:
Tested by https://github.com/ceph/ceph-qa-suite/pull/975

xiaoxichen:
@ajarr ,comments addressed.

ajarr:
@xiaoxichen , the tracker issue number in the commit message is incorrect. And please mention the link there http://tracker.ceph.com/issues/15417 instead of just the number.

xiaoxichen:
@ajarr , updated the number. Actually in practice, we always  only put the number in commit message.

ajarr:
@xiaoxichen , not really.  Checkout, https://github.com/ceph/ceph/commit/65770059cb319da9ffbc2866df1ffa8b35267614

ajarr:
Code looks good to me.

jcsp:
@xiaoxichen merged, thanks!  The number vs. URL thing isn't a big deal, but we are switching to using URLs because the tracker numbers were colliding with github pull request numbers; future patches should use URL.

PR #8772, opened 2016-04-26
rgw: sync fixes

yehudasa:


cbodley:
was able to run through test_multi.py (with mtime comparisons enabled) consistently without reproducing the sync feedback loop :+1:

PR #8773, opened 2016-04-27
Wip add flashcache

houlinfei:


liewegas:
I think we want to make this more generic so that we can layer in any caching option (flashcache, bcache, dm-cache).  Can you send an email to ceph-devel with the proposed ceph-disk interface change so we can get feedback and settle one the approach?  Thanks!

PR #8775, opened 2016-04-27
rbd-nbd: fix rbd-nbd aio callback error handling

cy-lee:
Tracker-Id: #15604

Signed-off-by: Chang-Yi Lee cy.l@inwinstack.com

### DIFF
--- a/src/tools/rbd_nbd/rbd-nbd.cc
+++ b/src/tools/rbd_nbd/rbd-nbd.cc
@@ -186,11 +186,12 @@ private:
 
     if (ret < 0) {
       ctx->reply.error = htonl(-ret);
-    } else if ((ctx->command == NBD_CMD_WRITE || ctx->command == NBD_CMD_READ)
-	       && ret != static_cast<int>(ctx->request.len)) {
-      derr << __func__ << ": " << *ctx << ": unexpected return value: " << ret
-	   << " (" << ctx->request.len << " expected)" << dendl;
-      ctx->reply.error = htonl(EIO);
+    } else if ((ctx->command == NBD_CMD_READ) &&
+                ret < static_cast<int>(ctx->request.len)) {
+      int pad_byte_count = static_cast<int> (ctx->request.len) - ret;
+      ctx->data.append('\x0', pad_byte_count);
+      dout(20) << __func__ << ": " << *ctx << ": Pad byte count: " 
+               << pad_byte_count << dendl;
     } else {
       ctx->reply.error = htonl(0);
     }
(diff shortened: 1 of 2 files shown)


cy-lee:
Here is Ceph-User mailing list discussion:
https://www.mail-archive.com/ceph-users@lists.ceph.com/msg28528.html

src/tools/rbd_nbd/rbd-nbd.cc
```diff
@@ -186,10 +186,9 @@ class NBDServer
 
     if (ret < 0) {
       ctx->reply.error = htonl(-ret);
-    } else if ((ctx->command == NBD_CMD_WRITE || ctx->command == NBD_CMD_READ)
-	       && ret != static_cast<int>(ctx->request.len)) {
+    } else if ((ctx->command == NBD_CMD_WRITE) && ret != 0) {
```
trociny: @cy-lee shouldn't this be something like below instead?
```
else if (ctx->command == NBD_CMD_READ && ret  != ctx->request.len) 
  error "unexpected return value for read"
else if (ctx->command != NBD_CMD_READ && ret  != 0)
  error "unexpected return value"
```
I.e. test that read returns number of bytes requested, any other ops should return 0 on success.

Or may be just check that return value 0 or request.len would be enough?

cy-lee: @trociny I think check return value with 0 might be enough.  
I tried to add read return value check 
``
else if (ctx->command == NBD_CMD_READ && ret  != ctx->request.len)
``
When I mount nbd image  it will show cannot read superblock.

I need others help review it.

dillaman: librbd can clip the output of a read if it's passed the end of the image.  I think we just remove the write return code check since it doesn't provide any value (negative error code already handled).  I would add a line for reads to pad zeroes to the end of returned ```data``` bufferlist if ```ret < ctx->request.len```.

src/tools/rbd_nbd/rbd-nbd.cc
```diff
@@ -186,11 +186,12 @@ class NBDServer
...
       ctx->reply.error = htonl(-ret);
-    } else if ((ctx->command == NBD_CMD_WRITE || ctx->command == NBD_CMD_READ)
-	       && ret != static_cast<int>(ctx->request.len)) {
-      derr << __func__ << ": " << *ctx << ": unexpected return value: " << ret
-	   << " (" << ctx->request.len << " expected)" << dendl;
-      ctx->reply.error = htonl(EIO);
+    } else if ((ctx->command == NBD_CMD_READ) &&
+                ret < static_cast<int>(ctx->request.len)) {
+      int pad_byte_count = static_cast<int> (ctx->request.len) - ret;
+      ctx->data.append('\x0', pad_byte_count);
```
dillaman: Can you swap this for ```ctx->data.append_zero(pad_byte_count)```?

dillaman: ... also, probably good to still explicitly set ```ctx->reply.error = htonl(0)```

cy-lee:
okay, thanks for your help :)

PR #8776, opened 2016-04-27
mon: optimize the monmap build_initial function

songbaisen:
mon: optimize the mommap build_initial function

improve the performance of this method by pulling two loops into one


liewegas:
this happens once on startup, and the loop is short.. not really worth optimizing

songbaisen:
@liewegas  :smile: Yes sir that is right.I just thought the origin code a little long.I close it here.

PR #8777, opened 2016-04-27
mds: fix 'allows_dirfrags' check for fragment_dir command

ukernel:


jcsp:

jcsp:
Skipping backport of this because it's a dev hook

PR #8778, opened 2016-04-27
mds: fix open symlink and truncate bugs

ukernel:


gregsfortytwo:
What testing did you do, @ukernel?

ukernel:
mount /mnt/ceph
cd /mnt/ceph/
touch file
ln -s file syslink
umount /mnt/ceph
mount /mnt/ceph
python -c 'import os; os.open("/mnt/ceph/symlink", os.O_RDWR)' // crash MDS
umount /mnt/ceph
mount /mnt/ceph
python -c 'import os; os.open("/mnt/ceph/symlink", os.O_RDWR|os.O_TRUNC)' // truncate symlink size to 0

src/mds/Server.cc
```diff
@@ -2893,9 +2893,13 @@ void Server::handle_client_open(MDRequestRef& mdr)
...
 
-  // can only open a dir with mode FILE_MODE_PIN, at least for now.
-  if (cur->inode.is_dir())
+  if (!cur->inode.is_file()) {
+    // can only open non-regular inode with mode FILE_MODE_PIN, at least for now.
     cmode = CEPH_FILE_MODE_PIN;
+    // the inode is symlink and client wants to follow it, ignore the O_TRUNC flag.
+    if (cur->inode.is_symlink() && !(flags & O_NOFOLLOW))
+      flags &= ~O_TRUNC;
+  }
```
gregsfortytwo: Is the expected behavior of truncating a symlink documented anywhere? I'm not finding it in initial searches, which makes evaluating this behavior a little hard.
(Although it's definitely better than crashing!)

ukernel: On Fri, May 6, 2016 at 6:43 AM, Gregory Farnum <notifications@github.com>
wrote:

> In src/mds/Server.cc
> <https://github.com/ceph/ceph/pull/8778#discussion_r62267169>:
>
> >      cmode = CEPH_FILE_MODE_PIN;
> > +    // the inode is symlink and client wants to follow it, ignore the O_TRUNC flag.
> > +    if (cur->inode.is_symlink() && !(flags & O_NOFOLLOW))
> > +      flags &= ~O_TRUNC;
> > +  }
>
> Is the expected behavior of truncating a symlink documented anywhere? I'm
> not finding it in initial searches, which makes evaluating this behavior a
> little hard.
> (Although it's definitely better than crashing!)
>
see https://github.com/ceph/ceph-client/blob/testing/fs/open.c#L75

But for this case, client actually wants to follow the symlink, truncate
what the symlink points to

gregsfortytwo: I'm confused; the kernel will return EINVAL on a symlink truncate; this is making us just silently drop it. Where does the client follow the symlink and then truncate the target, and why would we want that behavior to be different from the kernel's?

ukernel: There is a special type of open in kernel, which is called atomic_open. It does lookup and open at the same type. For example. a program calls open("/foo", O_RDWR | O_TRUNC), kernel sends an open request to open path "#1/foo".  If /foo is symlink, the MDS should treat the open request as a lookup.

ukernel:

gregsfortytwo:
Fixes: http://tracker.ceph.com/issues/16013

@ukernel, do you want to stick that in the commit message?

yuriw:
tested http://pulpito.ceph.com/yuriw-2016-05-25_20:47:39-fs-wip-yuri-testing---basic-smithi/

PR #8780, opened 2016-04-27
client: renew caps for read/write if mds session got killed.

ukernel:
When mds session got killed, read/write operation may hang (it also
can return -ESTALE). Client waits for Frw caps, but mds does not know
what caps client wants. To recover this, client sends an open request
to mds. The request will tell mds what caps client wants.


ukernel:

PR #8781, opened 2016-04-27
  mon: optimize the monmap calc_ranks function

songbaisen:
mon: optimize the mommap calc_ranks function


liewegas:
The reason this is 2 loops is because the ranks are assigned in a deterministic fashion, sorting by the entity_addr_t.

songbaisen:
@liewegas Thank you sir!I got it.The paxos need use the mon addr as the rank.

PR #8783, opened 2016-04-27
Event: do not increase EventCenter::file_events so aggresively

runsisi:
resize EventCenter::file_events array with x2 instead of x4


runsisi:
hi @yuyuyu101 , i'm not very sure if increase by x2 is your original intention, can you review this for me? many thx:)

yuyuyu101:
actually it's behavior as expected.

runsisi:
@yuyuyu101 , my fault, but can u explain it a little bit further? thank you very much:)

yuyuyu101:
no much reason, just to reduce extra memory copy when create event

PR #8784, opened 2016-04-27
vstart: replace ifconfig usage with ip addr

ddiss:
Ifconfig is obsolete with Linux 2.0+, and may not be in a user's PATH.
Use ip addr instead.

### DIFF
--- a/src/vstart.sh
+++ b/src/vstart.sh
@@ -387,8 +387,13 @@ if [ -n "$ip" ]; then
     IP="$ip"
 else
     echo hostname $HOSTNAME
+    if [ -x "$(which ip 2>/dev/null)" ]; then
+	IP_CMD="ip addr"
+    else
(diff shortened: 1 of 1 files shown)


theanalyst:
We should probably wrap this around a uname or so to make sure the BSD folks aren't affected (ref: see comments in  https://github.com/ceph/ceph/pull/8152)

ddiss:
Hmm sorry, missed your previous SR. Will resubmit with a uname -o check.
Thanks for the feedback.

ddiss:
Changes since previous version:
- Check for presence of "ip" executable
  + use it if available, otherwise fall back to ifconfig

wjwithagen:
lgtm.
Tested and runs on FreeBSD

theanalyst:
LGTM

ddiss:
Ping, please pull.

PR #8785, opened 2016-04-27
librbd: does not crash if image header is too short

tchaikov:
if something goes wrong with the object, and returns a chunk shorter
than expected, don't panic


jdurgin:
lgtm

PR #8789, opened 2016-04-27
msgr,xio: flexible Messenger::create options

vuhuong:
Widen Messenger::create and XioMessenger constructor to support
per-Messenger instance creation parameters.

This introduces a minimalist Proplist mechanism, based on <map>
and <boost::variant>, to pass in general parameters to
Messenger::create() without changing policy or involving names

Used the new Proplist to apply xio_portal_threads and
xio_max_conns_per_portal to XioMessenger. Tested with
xio_client/xio_server and cluster bringup with and without
Proplist.

We will extend the usage of Proplist to ceph-osd's
"workhorse" and "light" Messenger instances, and other
ceph clients Messenger in next patch

Signed-off-by: Vu Pham <vu@mellanox.com>

### DIFF
--- a/src/msg/Messenger.cc
+++ b/src/msg/Messenger.cc
@@ -21,7 +21,7 @@ Messenger *Messenger::create_client_messenger(CephContext *cct, string lname)
 
 Messenger *Messenger::create(CephContext *cct, const string &type,
 			     entity_name_t name, string lname,
-			     uint64_t nonce, uint64_t features)
+			     uint64_t nonce, uint64_t features, uint64_t cflags)
 {
   int r = -1;
   if (type == "random") {
@@ -36,7 +36,7 @@ Messenger *Messenger::create(CephContext *cct, const string &type,
 #ifdef HAVE_XIO
   else if ((type == "xio") &&
 	   cct->check_experimental_feature_enabled("ms-type-xio"))
-    return new XioMessenger(cct, name, lname, nonce, features);
+    return new XioMessenger(cct, name, lname, nonce, features, cflags);
 #endif
   lderr(cct) << "unrecognized ms_type '" << type << "'" << dendl;
   return nullptr;
--- a/src/msg/Messenger.h
+++ b/src/msg/Messenger.h
@@ -53,6 +53,14 @@ protected:
 
 public:
   /**
+   * Various Messenger conditional config/type flags to allow
+   * different "transport" Messengers to tune themselves
+   */
+  static const int HAS_HEAVY_TRAFFIC    = 0x0001;
+  static const int HAS_MANY_CONNECTIONS = 0x0002;
+  static const int HEARTBEAT            = 0x0004;
+
+  /**
    *  The CephContext this Messenger uses. Many other components initialize themselves
    *  from this value.
    */
@@ -151,13 +159,15 @@ public:
    * @param lname logical name of the messenger in this process (e.g., "client")
    * @param nonce nonce value to uniquely identify this instance on the current host
    * @param features bits for the local connection
+   * @param cflags general set of flags to configure transport resources
    */
   static Messenger *create(CephContext *cct,
                            const string &type,
                            entity_name_t name,
 			   string lname,
                            uint64_t nonce,
-			   uint64_t features = 0);
+			   uint64_t features = 0,
+			   uint64_t cflags = 0);
 
   /**
    * create a new messenger
--- a/src/test/messenger/xio_client.cc
+++ b/src/test/messenger/xio_client.cc
@@ -117,7 +117,7 @@ int main(int argc, const char **argv)
 	messenger = new XioMessenger(g_ceph_context,
 				     entity_name_t::MON(-1),
 				     "xio_client",
-				     0 /* nonce */, XIO_ALL_FEATURES,
+				     0 /* nonce */, XIO_ALL_FEATURES, 0 /* cflags */,
 				     dstrategy);
 
 	// enable timing prints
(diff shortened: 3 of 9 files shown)


vuhuong:
@mattbenjamin @cbodley @yuyuyu101 Could you review and provide feedback on this approach?

yuyuyu101:
Although we will remove heartbeat messenger in the future, this pr is more urgent to make xio work better. The only concern to me is add extra complexity and entry to messenger. 

I think if other guys agrees with this, it's ok to me too.....

src/msg/Messenger.h
```diff
@@ -140,6 +141,14 @@ class Messenger {
   virtual ~Messenger() {}
 
   /**
+   * Proplist notation for passing general parameters to Messenger::create()
+   * to support per-Messenger instance creation parameters
+   */
+  typedef boost::variant< int32_t, std::string > PropElt;
```
cbodley: If we're going to bake this data structure into the interface, it would be nice to support more than int and string. Would `boost::any` be crazy here?

vuhuong:
I think it's a good idea - providing flexibility with a cost of some hundred bytes during setup time.

athanatos:
@cbodley Can you take a look?

src/msg/Messenger.cc
```diff
@@ -36,7 +37,7 @@ Messenger *Messenger::create(CephContext *cct, const string &type,
 #ifdef HAVE_XIO
   else if ((type == "xio") &&
 	   cct->check_experimental_feature_enabled("ms-type-xio"))
-    return new XioMessenger(cct, name, lname, nonce, features);
+    return new XioMessenger(cct, name, lname, nonce, features, opts);
```
cbodley: I think we should go ahead and pass `opts` into the other Messengers' constructors while we're at it

src/test/messenger/xio_client.cc
```diff
@@ -114,11 +114,14 @@ int main(int argc, const char **argv)
 	else
 	  dstrategy = new QueueStrategy(2);
 
+        Messenger::Proplist opts;
+        opts.insert(Messenger::Property("xio_portal_threads", 1));
+        opts.insert(Messenger::Property("xio_max_conns_per_portal", 2));
```
cbodley: I know this is just an example, but I hope that for `ceph_osd.cc` we can avoid using magic numbers. So the actual values should come from ceph.conf; how do you imagine that working?

You mentioned the distinction between 'light' vs. 'workhorse' messengers, so would it be a set of variables like `xio_{light,heavy}_portal_threads` and `xio_{light,heavy}_max_conns_per_portal`?

If this is going to be the strategy, maybe all we really need to pass into `Messenger::create()` is this light/heavy flag. Then we could avoid building up these property lists for each messenger, and the flags would be much more likely to be useful for other messengers.

yuyuyu101:
@cbodley @liewegas if we pick up oob messenger interface task, we may don't need a distinguish between messenger so we don't need specified options?

vuhuong:
@cbodley It's a little bit more than just heavy and light.
1. osd-client, osd-cluster can be considered heavy (multiple portals, hundreds of connections). Mon is also kind of heavy with single portal and a lot of connections.
2. osd-hb-back-server, osd-hb-front-server, osd-hbclient: are medium messengers that need single portal but (conf->osd_heartbeat_min_peers * 4) connections per portal
3. The rest of messengers (osd-objecter, ceph clients mds, rbd, rgw, ceph commands...) are light messengers with single portal and 1-2 connections.
Calling Messenger::create() without "opts" can be mutually treated as light.

Beside the above heavy/light flags may limit the flexibility that "opts" has for passing different properties if any.

I delay to pass opts into the other Messengers' constructors for the subsequent patch after we solidify this one

cbodley:
@vuhuong okay, that makes sense. i still feel that these values need to come from ceph.conf, and that we should plan out which of these messengers lead to which config values

from there, my preference would be to distill these decisions into a generic set of flags that other Messengers have a chance to make use of, rather than xio-only properties. from your outline, this might something look like:
```
MSGR_SHARD_CONNS // unset = single portal thread, set = xio_num_portal_threads
MSGR_MANY_CONNS // unset = xio_few_conns_per_portal, set = xio_many_conns_per_portal
MSGR_HEARTBEAT // unset = xio_few_conns_per_portal, set = osd_heartbeat_min_peers * 4
```
@yuyuyu101 distinguishing heartbeat messengers is part of the picture, but not all of it. we still want to cut down on threads and connections by default, and beef them up for specific messengers. could you see any way that AsyncMessenger could take advantage of these hints?

yuyuyu101:
@cbodley For AsyncMessenger originally facing the same problem as this, later we change to a global(CephContext associated) thread pool instead of per messenger. So currently AsyncMessenger just is a collection of connections, no other associated threads or resources.

vuhuong:
@cbodley your suggestion makes sense. It works for other Messengers and future properties as well (as bit-wise flag). I'll  rework the patch. Thanks!

vuhuong:
@cbodley updated patch with configuration flags as suggested. Please help to review it.

mattbenjamin:
@vuhuong @cbodley  lgtm--see any remaining blockers?

cbodley:
looks great, thanks @vuhuong 

@liewegas @gregsfortytwo any objections?

liewegas:
looks fine to me!

gregsfortytwo:
Sure, looks fine.

yuyuyu101:
so we can go ahead!

yuyuyu101:
retest this please

vuhuong:
I rework the patch to have config flags contained in messenger class with self-documented name and explanations

yuriw:

PR #8793, opened 2016-04-27
radosgw-admin: update usage for zone[group] modify

cbodley:
the help strings indicated that only --master could be changed


theanalyst:
lgtm, needs a backport to jewel probably

cbodley:
added http://tracker.ceph.com/issues/15651 with backport=jewel

theanalyst:
+1

PR #8794, opened 2016-04-28
hammer: [MON] Fixed calculation of %USED.

dzafman:
Fixes backport http://tracker.ceph.com/issues/15635


dachary:
test this please

dzafman:
This is bogus because hammer doesn't have RAW USED %, so not sure where this needs to be backported to.

PR #8797, opened 2016-04-28
debian/control: dh_systemd_start is in the dh-systemd package

liewegas:


tchaikov:
thanks, missed it!

tchaikov:
@liewegas ,could you add "Fixes: http://tracker.ceph.com/issues/15573" to the commit message? so we can backport your fix along with the change to add dh_systemd_start?

liewegas:
https://github.com/ceph/ceph/pull/8812

PR #8798, opened 2016-04-28
common/buffer: clean up alignment checks

majianpeng:


majianpeng:
@tchaikov . please review. Thanks!

src/os/bluestore/KernelDevice.cc
```diff
@@ -375,10 +375,7 @@ int KernelDevice::aio_write(
   assert(off < size);
   assert(off + len <= size);
 
-  if (!bl.is_n_align_sized(block_size) || !bl.is_aligned(block_size)) {
-    dout(20) << __func__ << " rebuilding buffer to be aligned" << dendl;
-    bl.rebuild_aligned(block_size);
-  }
+  bl.rebuild_aligned_size_and_memory(block_size, block_size);
```
liewegas: The only thing we lose here is the debug message saying it had tod o something.  Maybe we can make the bufferlist method return true/false so we know if it had to do a rebuild, and still output the log message?  I'd like to be able to identify cases where a rebuild was needed and ensure the callers can't do a better job aligning their buffers ahead of time.

majianpeng:
update

liewegas:
lgtm!

src/os/bluestore/BlueFS.cc
```diff
@@ -378,7 +378,6 @@ int BlueFS::_write_super()
   ::encode(crc, bl);
   assert(bl.length() <= get_super_length());
   bl.append_zero(get_super_length() - bl.length());
-  bl.rebuild();
```
tchaikov: but `bl.rebuild()` is not identical to `bl.rebuild_aligned()` used by `BlockDevice::aio_write()`. the former create a new memory chunk, and concatenate all memory segments into the newly allocated one, while in the case of `KernelDevice`, the later aligns unaligned buffer segments. and in the case of NVMEDevice, it does not rebuild the buffer segments at all, instead, it splits the buffer to be written into slices of `data_buffer_size` bytes before sending them down to underlying storage.

src/common/buffer.cc
```diff
@@ -1422,6 +1422,18 @@ static simple_spinlock_t buffer_debug_lock = SIMPLE_SPINLOCK_INITIALIZER;
     return true;
   }
 
+  bool buffer::list::is_aligned_size_and_memory(unsigned align_size,
+						  unsigned align_memory) const
+  {
+    for (std::list<ptr>::const_iterator it = _buffers.begin();
```
tchaikov: i'd recommend you to use 

```c++
for (auto &p : _buffers) {
 //...
}
```

src/common/buffer.cc
```diff
@@ -1565,11 +1566,13 @@ static simple_spinlock_t buffer_debug_lock = SIMPLE_SPINLOCK_INITIALIZER;
       _buffers.insert(p, unaligned._buffers.front());
     }
     last_p = begin();
+
+    return  (old_memcopy_count != _memcopy_count);
   }
   
-  void buffer::list::rebuild_page_aligned()
+  bool buffer::list::rebuild_page_aligned()
```
tchaikov: we are changing the public API exposed by librados. maybe it's worth mentioning this in the release notes?

src/os/bluestore/BlueFS.cc
majianpeng: Yes, rebuild is not identical to rebuild_aligned. But in there, for aio_write it don't depend on the input parameter bufferlist whether is contiguous. Other place which call aio_write don't call rebuild.
In fact, rebuild don't consider bufferlist whether already contiguous. Why dose call rebuild? It want data in a contiguous memroy.

yuriw:
@majianpeng please rebase

majianpeng:
@yuriw . update. Thanks!

PR #8801, opened 2016-04-28
jewel: ceph-{mds,mon,osd} packages need scriptlets with systemd code

smithfarm:
http://tracker.ceph.com/issues/15633


smithfarm:
@branto1 Like what you see?

branto1:
LGTM but I just noticed we completely ignored the with_systemd bcond. We should follow-up with the patch that drops it.

smithfarm:
@branto1 Yeah, I realized that and was thinking about it, and then forgot. 

So, like backport all of #8720 to jewel?

smithfarm:
The alternative would be to add `%_with_systemd` conditionals to the jewel backport.

branto1:
Looks like we will be back-porting #8720 to jewel. In the mean-time, I think it should be pretty safe to merge this. We don't use the spec file to do sysvinit builds anyway.

PR #8804, opened 2016-04-28
hammer: Double decreased the count to trim caps which will cause failing to respond to cache pressure

smithfarm:
http://tracker.ceph.com/issues/15512


smithfarm:
@gregsfortytwo This PR passed a cephfs integration run with two failures. Both failures are reproducible in both hammer-backports and hammer itself. One failure is due to http://tracker.ceph.com/issues/11258 ("cephfs-java ftruncate unit test failure") which is fixed in master, but the hammer backport is non-trivial. The second failure is probably an issue with some of the tests. For details, see http://tracker.ceph.com/issues/15895#note-21 

smithfarm:
@jcsp OK to merge?

PR #8808, opened 2016-04-28
osd: fast dispatch heartbeat message

wjin:
As for simple messenger implementation, in a large-scaled ceph cluster, hundreds of threads are used to do heartbeat. Also, the total thread number of one node is approximately linear to qemu VMs (a.k.a librbd client) in extreme case.

Recently, we tested an extreme case: 80 ceph nodes, each has 9 OSDs. 
We ran fio instance in each VM. We observed that when more than 1000 VMs, there were about 30000 thread in our environment.

We observed that  dispatch thread may delay due to system schedule. 
Also, even if heartbeat uses independent messenger, many heartbeat reader threads and the dispatch thread will compete with the disptach queue lock, which might also delay dispatch thread if it fails to get the lock. 

### DIFF
--- a/src/osd/OSD.h
+++ b/src/osd/OSD.h
@@ -1594,6 +1594,20 @@ public:
   struct HeartbeatDispatcher : public Dispatcher {
     OSD *osd;
     explicit HeartbeatDispatcher(OSD *o) : Dispatcher(o->cct), osd(o) {}
+
+    bool ms_can_fast_dispatch_any() const { return true; }
+    bool ms_can_fast_dispatch(Message *m) const {
+      switch (m->get_type()) {
+	case CEPH_MSG_PING:
+	case MSG_OSD_PING:
+          return true;
+	default:
+          return false;
+	}
+    }
+    void ms_fast_dispatch(Message *m) {
+      osd->heartbeat_dispatch(m);
+    }
     bool ms_dispatch(Message *m) {
       return osd->heartbeat_dispatch(m);
     }


wjin:
hi, guys, any risk about fast dispatch heartbeat message? this is just a temp patch, and I fast deal with all three kinds of heartbeat message.

liewegas:
We need to audit the locks taken by the heartbeat dispatch function.  In particular, we can't block on anything that could make us deadlock.

yuyuyu101:
good try. cautious to scheduler mutli large qa tests

wjin:
[reply.txt](https://github.com/ceph/ceph/files/241730/reply.txt)
[send.txt](https://github.com/ceph/ceph/files/241729/send.txt)

Here are two logs under pressure test even though it is not the worst case. 
But we can see that dispatch thread of 'reply side' is blocked for a long time (more than 10 seconds). In my opinion, if it is runnable, it shouldn't wait for schedule for such a long time. It might fail to get the lock and then sleep. Please correct me if I am wrong.

Hi, Sage, any suggestions about 'audit the lock'? I can do more test to figure it out if possible.

yuyuyu101:
audit each heartbeat dispatch logic, to see whether need to acquire messenger/connection's lock

wjin:
It seems that the reader thread does not hold any lock before doing fast dispatch, so it is pretty safe :(

For further analysis, when doing heartbeat logic, it may do share_map_peer or osdmap_subscribe, but these two steps will use cluster messenger instead of heartbeat messenger to send message, so we could skip check this?

Finally, the last suspicious one should be:
      m->get_connection()->send_message(r);

It will use ms_hb_back_server or ms_hb_front_server to send message. It works well if connection is ok. If connection is something wrong, it will drop the message due to server policy.

src/osd/OSD.h
```diff
@@ -1594,6 +1594,21 @@ class OSD : public Dispatcher,
...
     OSD *osd;
     explicit HeartbeatDispatcher(OSD *o) : Dispatcher(o->cct), osd(o) {}
+
+    bool ms_can_fast_dispatch_any() const { return true; }
+    bool ms_can_fast_dispatch(Message *m) const {
+      switch (m->get_type()) {
+	case CEPH_MSG_PING:
+	case MSG_OSD_PING:
+	case CEPH_MSG_OSD_MAP:
+          return true;
```
liewegas: We can fast dispatch pings, but *not* osdmap

liewegas:
Those look okay--the dispatch_op_fast() calls those same methods.

We can't fast dispatch the MOSDMap, though--just the ping messages. With that change I think this is okay!

wjin:
Thanks Sage. I agree that we don't need to fast dispatch MOSDMap as it is not related to heartbeat ping. However, I have a doubt, MOSDMap should always be sent with cluster messenger, when will it be sent with heartbeat messenger? 

Does MOSDMap really exist between ms_hbclient and ms_hb_{back,front}_server?
If it never happens, seems we need to remove this case from OSD::heartbeat_dispatch:
  case CEPH_MSG_OSD_MAP:
    {
      ConnectionRef self = cluster_messenger->get_loopback_connection();
      self->send_message(m);
    }
    break;

liewegas:
Ah, yeah, I think we don't send it anymore.  We could/should just remove 
that case in a separate commit.

yuyuyu101:
👍

zhouyuan:
@liewegas 
Hi Sage, I see there's a [pad ](http://pad.ceph.com/p/messenger)on messenger optimizations and there's a slot on heartbeat. Can you please kindly share some detail thoughts on that?

liewegas:
@zhouyuan how about we discuss it during the CDM call next wednesday?  http://tracker.ceph.com/projects/ceph/wiki/Planning

PR #8810, opened 2016-04-28
osd: fix problematic got_full_map() logic

xiexingguo:
We shall only clear out requested_full_first/last either we got all
we want or we received a map that was newer than requested_full_last.

Also we shall increase requested_full_first properly to reflect exactly
we have got instead of advancing requested_full_first blindly.

NB: if requested_full_first is equal to requested_full_last, we still
have one epoch to request.


tchaikov:
@liewegas does needs-qa implies "lgtm" also?

PR #8812, opened 2016-04-28
debian/control: dh_systemd_start is in the dh-systemd package

liewegas:


ktdreyer:
Just this this myself... :+1:

PR #8813, opened 2016-04-28
test: librbd: aio write/discard completion should always return 0 on success

trociny:


src/test/librbd/test_librbd.cc
```diff
@@ -1278,7 +1278,8 @@ void aio_discard_test_data(librbd::Image& image, off_t off, size_t len, bool *pa
   image.aio_discard(off, len, comp);
   comp->wait_for_complete();
   int r = comp->get_return_value();
-  ASSERT_TRUE(r >= 0);
+  printf("return value is: %d\n", r);
```
dillaman: Nit: this printout is not needed

trociny: @dillaman fixed

PR #8814, opened 2016-04-28
osd reweight-by-*: sort osds by diff from mean

dvanders:
Sort the OSDs by their absolute difference from average_util, instead of
their utilization. Without this, the underloaded OSDs are reweighted last,
which means never unless the admins reweights all OSDs (not recommended).

Backport: jewel


liewegas:
https://github.com/ceph/ceph/pull/8733

PR #8819, opened 2016-04-28
rbd: disk usage now includes all snapshots by default

dillaman:
Added a new "--from-snap" optional to restrict the calculation to
specific subsets of snapshots and the HEAD revision.

### DIFF
--- a/src/tools/rbd/action/DiskUsage.cc
+++ b/src/tools/rbd/action/DiskUsage.cc
@@ -86,8 +86,8 @@ static int compute_image_disk_usage(const std::string& name,
 }
 
 static int do_disk_usage(librbd::RBD &rbd, librados::IoCtx &io_ctx,
-                        const char *imgname, const char *snapname,
-                        Formatter *f) {
+                         const char *imgname, const char *snapname,
+                         const char *from_snapname, Formatter *f) {
   std::vector<std::string> names;
   int r = rbd.list(io_ctx, names);
   if (r == -ENOENT) {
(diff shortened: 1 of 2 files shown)


src/tools/rbd/action/DiskUsage.cc
```diff
@@ -152,6 +153,7 @@ static int do_disk_usage(librbd::RBD &rbd, librados::IoCtx &io_ctx,
       continue;
     }
 
+    bool found_from_snap = (from_snapname == nullptr);
```
jdurgin: !=?

src/tools/rbd/action/DiskUsage.cc
```diff
@@ -167,7 +169,8 @@ static int do_disk_usage(librbd::RBD &rbd, librados::IoCtx &io_ctx,
         goto out;
       }
 
-      if (imgname == NULL || (snapname != NULL && snap->name == snapname)) {
+      if (imgname == nullptr || found_from_snap ||
+         (found_from_snap && snapname != nullptr && snap->name == snapname)) {
         r = compute_image_disk_usage(*name, snap->name, last_snap_name,
```
jdurgin: from_snap instead of last_snap_name? and no need to get last_snap_name anymore?

jdurgin: nvm, making sense now

jdurgin:
lgtm, it'd be good to have a simple test in qa/workunits/copy.sh or similar.

src/tools/rbd/action/DiskUsage.cc
dillaman: Avoiding the need to test "from_snapname == nullptr" two times below by just combing it into the single bool -- if we aren't restricting the starting snapshot, might as well treat it as we've found the starting snapshot.

PR #7715, opened 2016-02-19
[WIP] rgw: add compression to client

Ved-vampir:
Compress and decompress data on client side.

PR #8820, opened 2016-04-28
cmake: fix linking of compressor and plugins

cbodley:
libcommon now includes the compressor objects necessary to load compressor plugins

also removed extra dependencies on the plugins and their unit tests


cbodley:
this fixes the cmake build for #7715

PR #8821, opened 2016-04-28
debian: fix /etc/default/ceph location

aderumier:
commit 7384a14f243519547a61534d22e21e6069aae016 have introduce
the /etc/default/ceph file at a wrong location : /etc/default/ceph/ceph

fix http://tracker.ceph.com/issues/15587


liewegas:
:+1:

FlorentCoppint:
Hi, is it backported to Jewel ? 10.2.1 packaged version has bug.

PR #8822, opened 2016-04-28
cmake: put virtualenv directories under /tmp

cbodley:
this was required for cmake-check to find ceph-disk and ceph-detect-init in the PATH (because that's where automake puts them)


src/ceph-detect-init/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-set(CEPH_DETECT_INIT_VIRTUALENV ${CMAKE_BINARY_DIR}/ceph-detect-init-virtualenv)
+set(CEPH_DETECT_INIT_VIRTUALENV /tmp/ceph-detect-init-virtualenv)
```
tchaikov: can we read "/tmp" from `ENV{TMPDIR}` and default to "/tmp" if it's empty or not set?

cbodley:
@tchaikov thanks, updated to try TMPDIR

adamemerson:
This is a fine patch and I demand^Wrequest that it be merged.

tchaikov:
lgtm also.

sebastian-philipp:
See http://tracker.ceph.com/issues/24710

PR #8823, opened 2016-04-28
etc/default/ceph: set 128MB tcmalloc cache size

aderumier:
Signed-off-by: Alexandre Derumier <aderumier@odiso.com>

commit https://github.com/ceph/ceph/commit/9565a50c58194bf2ffd52f6c3cc797a28047ddbe

have increased tcmalloc cache size to 128M in rpm sysconfig.

We need to do the same for deb /etc/default/ceph

This fix : http://tracker.ceph.com/issues/15588

### DIFF
--- a/etc/default/ceph
+++ b/etc/default/ceph
@@ -3,6 +3,10 @@
 # Environment file for ceph daemon systemd unit files.
 #
 
+# Increase tcmalloc cache size
+TCMALLOC_MAX_TOTAL_THREAD_CACHE_BYTES=128MB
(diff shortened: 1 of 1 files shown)


liewegas:
That other commit hanges it from MB to bytes.. i assume we need to do the same thing here?

aderumier:
yes, you are right, better to have same config. I have pushed a second commit.

ktdreyer:
Why is debian using an entirely different file :(

PR #8824, opened 2016-04-28
DNM: log: use coarse time

liewegas:
```
- coarse time is way faster than gettimeofday.
- log events are ordered, so the exact timestamps mostly only matter when
  comparing events across hosts.  and host clocks drift anyway, so there
  is little value for debugging purposes.
- if you *do* need accurate time for events, modify the code to log it
  explicitly.
```

### DIFF
--- a/src/common/Graylog.cc
+++ b/src/common/Graylog.cc
@@ -84,7 +84,8 @@ void Graylog::log_entry(Entry const * const e)
     m_formatter->dump_string("host", m_hostname);
     m_formatter->dump_string("short_message", s);
     m_formatter->dump_string("_app", "ceph");
-    m_formatter->dump_float("timestamp", e->m_stamp.sec() + (e->m_stamp.usec() / 1000000.0));
+    auto ts = ceph::coarse_real_clock::to_timespec(e->m_stamp);
+    m_formatter->dump_float("timestamp", ts.tv_sec + (ts.tv_nsec / 1000000000.0));
     m_formatter->dump_int("_thread", e->m_thread);
     m_formatter->dump_int("_level", e->m_prio);
     if (m_subs != NULL)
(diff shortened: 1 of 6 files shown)


liewegas:
@adamemerson

adamemerson:
This looks a fine patch to me.

liewegas:
Hmm, this gives us roughly 1ms granularity.  Perhaps the coarse time string conversion should only do 3 digits of precision?

yuyuyu101:
So we may change the log line format? It's ok to me, but I'm afraid that it will break some users log filter like logstash. I think this should be showed in release log...

liewegas:
I'd actually like to update this so that a config option can control the 
timing precision.  Need to futz with it a bit though to dynamically switch 
between coarse and non-coarse time_points...

liewegas:
@adamemerson How would you suggest making coarse vs normal time a config option?  Don't really want to make everything virtual to get around the times not being compatible.

adamemerson:
Probably the most natural thing to do would be to create a log_clock class in one of the Log headers. Rather than having now be static, have now be a member function. It should have a constructor whose parameter tells it whether to use coarse or fine time, and just store CLOCK_REALTIME or CLOCK_REALTIME_COARSE (modulo BSD) in a member variable. Then just have its now function pass that variable as the parameter to clock_gettime. Store the instance in CephContext.

Upside is that it /almost/ satisfies the Clock concept, with the exception being 'now' being a member rather than static function. Also there's no branch, virtual function, or whatever. Downside is it seems a bit messy, but probably the least messy way of doing it.

if that sounds like a decent idea to people I can write up such a thing.

liewegas:
@adamemerson do you mind adopting this PR?  I think it's important to be able to switch between low-res and high-res timestamps via a config option.

adamemerson:
Sure, I can adopt it. I think I have an idea for how. (How DOES one adopt a PR? Can I push things to it and so forth, or should I clone your branch and open a new PR of my own?)

liewegas:
Yeah just push your own branch and open a new pr.  Thanks!

PR #8825, opened 2016-04-28
mds: keep btime

liewegas:
Set it when the inode is created (set it to initial ctime).  Store it.

btime is important for CIFS, and will eventually be exposed via the new statx
syscall on linux.

### DIFF
--- a/src/mds/mdstypes.h
+++ b/src/mds/mdstypes.h
@@ -470,6 +470,7 @@ struct inode_t {
 
   // affected by any inode change...
   utime_t    ctime;   // inode change time
+  utime_t    btime;   // birth time
 
   // perm (namespace permissions)
   uint32_t   mode;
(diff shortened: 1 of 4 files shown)


liewegas:
the client piece should wait until we are ready to spend a feature bit.  branch is here: http://github.com/liewegas/ceph/commits/wip-btime-client

jcsp:
Would be nice to have a test for this.  I guess we could have one that used cache dump to inspect btime, to cover us before we have statx

jcsp:
@liewegas are you planning on creating any tests or do you need someone else to do that?

liewegas:
I don't have time to follow through on this one, sorry!

src/mds/mdstypes.h
```diff
@@ -470,6 +470,7 @@ struct inode_t {
 
   // affected by any inode change...
```
jtlayton: The above comment will be confusing after this, maybe add something like:

// set to ctime when file is created, and only changes afterword with explict setattr

liewegas:
https://github.com/ceph/ceph/pull/9965

PR #8826, opened 2016-04-29
os/bluestore: bitmap-based freelist using merge operator

liewegas:
The xor code here is super inefficient, but I think the infrastructure
and interface changes are appropriate.  Please review, and/or send patches
making the xor stuff fast(er)!


liewegas:
@allensamuels @xiexingguo

src/os/bluestore/BlueStore.cc
```diff
@@ -1750,6 +1752,12 @@ int BlueStore::mkfs()
   {
     dout(20) << __func__ << " initializing freespace" << dendl;
     KeyValueDB::Transaction t = db->get_transaction();
+    {
+      bufferlist bl;
+      bl.append(freelist_type);
+      t->set(PREFIX_SUPER, "freelist_type", bl);
+    }
+    fm->create(t);
```
xiexingguo: Two problems here:
1. it shall be ```fm->create(freelist_type);``` instead of ```fm->create(t);```
2. I think you shall move the initialization of ```fm``` ahead, before we call ``` _open_alloc```, which is intend for space initialization and will access ```fm``` .

xiexingguo: Gotcha your upgrades. So we shall simply drop this.

src/os/bluestore/BlueStore.cc
```diff
@@ -2037,10 +2073,14 @@ int BlueStore::fsck()
...
-  r = _open_alloc();
+  r = _open_fm(false);
   if (r < 0)
     goto out_db;
 
+  r = _open_alloc();
+  if (r < 0)
+    goto out_fm;
+
   r = _open_super_meta();
```
xiexingguo: We shall move ```_open_super_meta()``` ahead here,  just before calling ```_open_fm(false)```(as we did in ```mount()```), because ```_open_fm(false)``` relies on **freelist_type** to be preloaded from db.

src/os/bluestore/BitmapFreelistManager.cc
```diff
@@ -0,0 +1,399 @@
...
+
+  uint64_t first_key = offset & key_mask;
+  uint64_t last_key = (offset + length - 1) & key_mask;
+  dout(20) << __func__ << " first_key " << first_key << " last_key " << last_key
+	   << dendl;
+
+  if (first_key == last_key) {
+    bufferptr p(blocks_per_key >> 3);
+    p.zero();
+    unsigned s = (offset & ~key_mask) / bytes_per_block;
```
xiexingguo: We may want to force block-size to be power of 2 aligned one day, so we can avoid this potential expensive block-index translation. E.g., for `block-size = 4096`, we may introduce a similar member -- "`block_size_shift = 12`"， and perform the above cast more efficiently:
`unsigned s = (offset & ~key_mask) >> block_size_shift`
But I guess the compiler will be smart enough to do the same optimization if block-size is already power of 2 aligned:-)

src/os/bluestore/BitmapFreelistManager.cc
```diff
@@ -0,0 +1,399 @@
...
+      memset(p.c_str(), 0xff, p.length());
+      while (first_key < last_key) {
+	string k;
+	make_offset_key(first_key, &k);
+	bufferlist bl;
+	bl.append(p);
+	dout(30) << __func__ << " " << first_key << ": ";
+	bl.hexdump(*_dout);
+	*_dout << dendl;
+	txn->merge(prefix, k, bl);
```
xiexingguo: I guess that you probably intend to use the global initialized `all_set_bl` here(to replace bl), for the sake of performance:)

src/os/bluestore/BitmapFreelistManager.h
```diff
@@ -0,0 +1,69 @@
...
+  std::mutex lock;
+
+  uint64_t bytes_per_block; ///< bytes per block (bdev_block_size)
+  uint64_t blocks_per_key;  ///< blocks (bits) per key/value pair
+  uint64_t blocks;          ///< size of device (blocks)
+
+  uint64_t block_mask;  ///< mask to convert byte offset to block offset
+  uint64_t key_mask;    ///< mask to convert offset to key offset
+
+  uint64_t free;
```
xiexingguo: Did not see that `free` is used anywhere.

src/os/bluestore/BitmapFreelistManager.cc
```diff
@@ -0,0 +1,399 @@
...
+      }
+    }
+  }
+
+  end = blocks * bytes_per_block;
+  if (enumerate_offset < end) {
+    *length = end - enumerate_offset;
+    dout(10) << __func__ << " " << *offset << "~" << *length << dendl;
+    enumerate_offset = end;
+    return true;
```
xiexingguo: Shall return `false` here, I think.

src/os/bluestore/BitmapFreelistManager.cc
```diff
@@ -0,0 +1,399 @@
...
+	end = get_offset(enumerate_offset, enumerate_bl_pos);
+	dout(30) << __func__ << " found set bit, key " << enumerate_offset
+		 << " bit " << enumerate_bl_pos
+		 << " offset " << end
+		 << dendl;
+	*length = end - *offset;
+	dout(10) << __func__ << " " << *offset << "~" << *length << dendl;
+	return true;
+      }
+      if (enumerate_bl_pos < 0) {
```
xiexingguo: `else` will suffice.

src/os/bluestore/BitmapFreelistManager.cc
```diff
@@ -0,0 +1,399 @@
...
+  while (true) {
+    enumerate_bl_pos = get_next_clear_bit(enumerate_bl, enumerate_bl_pos);
+    if (enumerate_bl_pos >= 0) {
+      dout(30) << __func__ << " found clear bit, key " << enumerate_offset
+	       << " bit " << enumerate_bl_pos
+	       << " offset " << get_offset(enumerate_offset, enumerate_bl_pos)
+	       << dendl;
+      break;
+    }
+    if (enumerate_bl_pos < 0) {
```
xiexingguo: `else` will suffice.

src/os/bluestore/BitmapFreelistManager.cc
```diff
@@ -0,0 +1,399 @@
...
+    }
+    ++start;
+  }
+  return -1; // not found
+}
+
+int get_next_set_bit(bufferlist& bl, int start)
+{
+  const char *p = bl.c_str();
+  int bits = bl.length() << 3;
```
xiexingguo: Shall we add a sanity check like `assert(bits == blocks_per_key)` here?

src/os/bluestore/BitmapFreelistManager.cc
```diff
@@ -0,0 +1,399 @@
...
+{
+  std::lock_guard<std::mutex> l(lock);
+  enumerate_offset = 0;
+  enumerate_bl.clear();
+}
+
+int get_next_clear_bit(bufferlist& bl, int start)
+{
+  const char *p = bl.c_str();
+  int bits = bl.length() << 3;
```
xiexingguo: Shall we add a sanity check like `assert(bits == blocks_per_key)` here?

src/os/bluestore/BitmapFreelistManager.cc
```diff
@@ -0,0 +1,399 @@
...
+	const char *p = k.c_str() + 1;
+	_key_decode_u64(p, &enumerate_offset);
+	enumerate_bl = enumerate_p->value();
+      }
+    }
+  }
+
+  end = blocks * bytes_per_block;
+  if (enumerate_offset < end) {
+    *length = end - enumerate_offset;
```
xiexingguo: Perhaps we shall introduce a member to remember the `size` passed in during initialization.
```c++
int BitmapFreelistManager::create(uint64_t size, KeyValueDB::Transaction txn)
{
  bytes_per_block = g_conf->bdev_block_size;
  blocks = size / bytes_per_block;
  blocks_per_key = g_conf->bluestore_freelist_blocks_per_key;
  if (blocks / blocks_per_key * blocks_per_key != blocks) {
    dout(10) << __func__ << " rounding blocks up from " << blocks << dendl;
    blocks = (blocks / blocks_per_key + 1) * blocks_per_key;
  }
```
As we may round up `blocks` to be `blocks_per_key` aligned as above, which means here `end` could pass over the origin `size` and cause `offset + length` to be greater than origin `size`.

yangdongsheng: What about rounding blocks *down* to blocks_per_key aligned?

src/os/bluestore/BitmapFreelistManager.cc
yangdongsheng: Agreed, compiler would switch it to shift operation I think

yangdongsheng:
Hi @liewegas , IIUC, the principle of bitmap freelist is to store a bitmap to represent each blocks. and '0' means free and '1' means allocated, right? 

In addition, what about the partial-allocated case? If I want to allocate [0, block_size/2]. How can we handle it?

src/os/bluestore/BitmapFreelistManager.cc
```diff
@@ -0,0 +1,399 @@
...
+    dout(20) << __func__ << " " << offset << "~" << length << dendl;
+  }
+}
+
+int BitmapFreelistManager::allocate(
+  uint64_t offset, uint64_t length,
+  KeyValueDB::Transaction txn)
+{
+  dout(10) << __func__ << " " << offset << "~" << length << dendl;
+  return _xor(offset, length, txn);
```
yangdongsheng: Is that necessary to check it before _xor?

src/os/bluestore/BitmapFreelistManager.cc
```diff
@@ -0,0 +1,399 @@
...
+}
+
+int BitmapFreelistManager::release(
+  uint64_t offset, uint64_t length,
+  KeyValueDB::Transaction txn)
+{
+  dout(10) << __func__ << " " << offset << "~" << length << dendl;
+  return _xor(offset, length, txn);
```
yangdongsheng: Ditto

src/os/bluestore/BitmapFreelistManager.cc
yangdongsheng: Yes, so what about just removing this line to fall through to ```dout(10) << "end" ``` and return false.

src/os/bluestore/BitmapFreelistManager.cc
yangdongsheng: Is that necessary to check it before _xor?

src/os/bluestore/BitmapFreelistManager.cc
```diff
@@ -0,0 +1,399 @@
...
+
+bool BitmapFreelistManager::enumerate_next(uint64_t *offset, uint64_t *length)
+{
+  std::lock_guard<std::mutex> l(lock);
+
+  // initial base case is a bit awkward
+  if (enumerate_offset == 0) {
+    dout(10) << __func__ << " start" << dendl;
+    enumerate_p = kvdb->get_iterator(prefix);
+    enumerate_p->lower_bound(string("o"));
```
yangdongsheng: That's could be a stupid question, but could you add a comment here to explain what does "o" mean? thanx :)

yangdongsheng:
@liewegas  about the _xor() I have an idea about it. what about introducing an array as bellow:
```
dictionary[][] = {{10000000B, 11000000B, 11100000B, 11110000B, 11111000B, 11111100B, 11111110B, 11111111B},
                         {00000000B, 01000000B, 01100000B, 01110000B, 01111000B, 01111100B, 01111110B, 01111111B},
                         {00000000B, 00000000B, 00100000B, 00110000B, 00111000B, 00111100B, 00111110B, 00111111B},
                         {00000000B, 00000000B, 00000000B, 00010000B, 00011000B, 00011100B, 00011110B, 00011111B},
                          {00000000B, 00000000B, 00000000B, 00000000B, 00001000B, 00001100B, 00001110B, 00001111B},
                           {00000000B, 00000000B, 00000000B, 00000000B, 00000000B, 00000100B, 00000110B, 00000111B},
                           {00000000B, 00000000B, 00000000B, 00000000B, 00000000B, 00000000B, 00000010B, 00000011B},
                            {00000000B, 00000000B, 00000000B, 00000000B, 00000000B, 00000000B, 00000000B, 00000001B}}
```

Then we can get an xoi factor by dictionary[start][end]. ie, start = 3, end=5. then dictionary[3][5] = 00011100B. 

In this way, I think we can reduce the for loop by 8 times. :)

src/os/bluestore/BitmapFreelistManager.cc
liewegas: We are at the end, but this is the last free extent, and we have to return it to the caller.  The next call will already be at the end with a large enumerate_offset and will return false.

xiexingguo: But we shall have call `enumerate_bl.clear();`(https://github.com/ceph/ceph/pull/8826/files#diff-69a76dbcf69fbc2c8d7361a5996be69cR268) to get here, which means on next entry we shall call get_next_clear_bit first(https://github.com/ceph/ceph/pull/8826/files#diff-69a76dbcf69fbc2c8d7361a5996be69cR254). It won't work then, I guess.

src/os/bluestore/BitmapFreelistManager.cc
liewegas: Not sure what we can check.  We could maybe have a debug option that does a get and verifies bits are not set, but we'd never want to enable that on a running system.

src/os/bluestore/BitmapFreelistManager.cc
liewegas: Oops, cruft.. removing it.

liewegas:
I think first step would be breaking the loop down into word-size chunks.

For the edges, I'm not sure whether constructing a word with teh right bits set or a memory access/array lockup is faster.  I'm guessing the first is?

liewegas:
We will never allocate partial blocks, so a bit per block is okay.

src/os/bluestore/BitmapFreelistManager.cc
xiexingguo: why shall we removing this...  It means we care about the offset key(begins with "o") only... Otherwise we will locate at the meta key which begins with a  "_" here, which is wrong.

liewegas:
* cleaned up bit order (now little endian)
* better debug output
* fixed keys (we have 2 prefixes now)
* fixed enumeration, esp when bdev size does not align with the keys

src/os/bluestore/BitmapFreelistManager.cc
liewegas: we have 2 separate prefixes now!

src/os/bluestore/BitmapFreelistManager.cc
allensamuels: Yes, we should have that as a debug option and yes, it should default to "off".

src/os/bluestore/BitmapFreelistManager.cc
```diff
@@ -0,0 +1,413 @@
...
+}
+
+int BitmapFreelistManager::release(
+  uint64_t offset, uint64_t length,
+  KeyValueDB::Transaction txn)
+{
+  dout(10) << __func__ << " " << offset << "~" << length << dendl;
+  return _xor(offset, length, txn);
+}
+
```
allensamuels: Ditto here, under a debug option we should do a "get" and verify the state of the affected bits.

src/os/bluestore/BitmapFreelistManager.cc
allensamuels: Don't bother. This won't yield any measurable CPU performance savings when you compare a few integer divides to the magnitude of the rest of the code path.

liewegas:
Look ok?

allensamuels:
Yes.


Allen Samuels
Engineering Fellow, Software Architect, Emerging Storage Solutions
[cid:image001.png@01CF4C2D.A7116C90]
951 SanDisk Drive, Milpitas, CA 95035

src/os/bluestore/BitmapFreelistManager.cc
yangdongsheng: my bad. It should be "true" here. We really found a free extend which is the last one. We should return false in next loop.

xiexingguo:
LGTM too

src/os/bluestore/BitmapFreelistManager.cc
```diff
@@ -0,0 +1,533 @@
...
+    *new_value = std::string(rdata, rlen);
+  }
+  virtual void merge(
+    const char *ldata, size_t llen,
+    const char *rdata, size_t rlen,
+    std::string *new_value) {
+    assert(llen == rlen);
+    *new_value = std::string(ldata, llen);
+    for (size_t i = 0; i < rlen; ++i) {
+      (*new_value)[i] ^= rdata[i];
```
yangdongsheng: Hi Sage, sorry for the late. One question here about merge.
(1), alloc the first block ===> txn->merge(bitmap_prefix, k, "10000000");
(2), release the first block ===> txn->merge(bitmap_prefix, k, "00000000");
Then we would merge the them. from what I read from merge() here, it should be:
```"10000000" ^ "00000000" = "10000000"``` 
But,,,,it should "00000000" after a release I think. There must be something wrong in my understanding, please correct me, thanx.

liewegas: The release also sets the bit, so that the xor op will clear it.

PR #8828, opened 2016-04-29
osd/OpRequest: reset connection upon unregister

tchaikov:
this helps to free the resources referenced by the connection, among
other things, in the case of MOSDOp, the OSD::Session and OSDMap. this
helps to free the resource earlier and trim the osdmaps in time.

### DIFF
--- a/src/osd/OpRequest.cc
+++ b/src/osd/OpRequest.cc
@@ -82,6 +82,10 @@ void OpRequest::_unregistered() {
   request->clear_data();
   request->clear_payload();
   request->release_message_throttle();
+  auto conn = request->get_connection();
+  if (conn) {
+    conn->set_priv(nullptr);
+  }
 }
 
 bool OpRequest::check_rmw(int flag) {


xiexingguo:
Bravo. I guess the long hunting is over now?

tchaikov:
@xiexingguo not really, if new op requests keep coming in. the (completed) old ones will be removed from the OpHistory eventually. so this changeset does not resolve the issue, it alleviates it.

xiexingguo:
:100: for the hard working

tchaikov:
note to myself:


should reset the priv in connection instead of the connection in the hammer backport. as `PG::RecoveryState::Stray::react(const MLogRec& logevt)` checks the features bit of the connection of a message for working around an issue

gregsfortytwo:
Did you make sure the history output makes sense once you zap the connection away? We don't want anything to try and dereference it for printing. ;)

tchaikov:
@gregsfortytwo, you are right, the peer address is gone with the zapped connection...

tchaikov:
we could have reset connection in the Message in order to release the reference to `PipeConnection` , but `OpRequest::_dump()` tries to print the `m->get_orig_source()`, which references connection if it is not null. so we just release the Session instead.

tchaikov:
set the performance label, because it helps to free the allocated memory.

tchaikov:
it fails the rbd tests.

PR #8830, opened 2016-04-29
rbd-mirror: support bootstrap canceling 

trociny:


trociny:
@dillaman This is (may be naive) implementation of bootstrap cancelling. When I test it, mirroring image usually becomes in the state that it is impossible to delete local image. Image replay restart fails with "split-brain detected -- skipping image replay"
```
0x7facb000e720 handle_get_remote_tags: decoded remote tag: [mirror_uuid=, predecessor_mirror_uuid=]
 0x7facb000e720 handle_get_remote_tags: decoded remote tag: [mirror_uuid=, predecessor_mirror_uuid=]
 0x7facb000e720 handle_get_remote_tags: local tag data: [mirror_uuid=f668acb0-0ede-4a59-9571-8bdad7416c8b, predecessor_mirror_uuid=]
 0x7facb000e720 handle_get_remote_tags: split-brain detected -- skipping image replay
```

Do you think I should work harder leaving image in more consistent state after canceling? Or is it supposed that `rbd mirror image resync` should trigger all necessary cleanup? Right now it looks like `image resync` is not very useful -- it sets resync flag but I don't see it is used?

dillaman:
Eventually 'rbd mirror image resync' will work -- we'll get it plumbed into the image deletion propagation once that's ready. 

There is most likely a logic error with the current split-brain detection (I kept it simple to pass the simple cases for orderly transition).  When we are still in the ```MIRROR_PEER_STATE_SYNCING``` state, we should be able to re-attach to the peer and re-start the sync from where it left off.

dillaman:
@trociny are you still waiting on additional comments?

trociny:
@dillaman It is outdated (I saw many similar bits came via different commits), I am going to rebase it and look how it runs now. Anyway I think it will become mostly useful when  'rbd mirror image resync' works (which I suppose will be soon).

dillaman:
@trociny Agreed -- definitely will be useful (especially for timely shutdowns of rbd-mirror daemon when in-progress large image syncs)

src/tools/rbd_mirror/ImageReplayer.cc
```diff
@@ -379,6 +379,7 @@ void ImageReplayer<I>::bootstrap() {
 
   {
     Mutex::Locker locker(m_lock);
+    request->get();
```
rjfd: Instead of inheriting from a RefCountedObj, you could use a `std::unique_ptr` to hold the pointer to the BootstrapRequest. The request object would be automatically deleted upon the destruction of the image replayer.

trociny: It will not work for mock (unittest) classes. Also, ImageReplayer has long live, and can be restarted.

Before the change the request objects already had a property that they are automatically destroyed after send() is finished. I wan to keep this property and just need a way to control (delay) this destruction to be sure cancel() is called for yet existing object.

rjfd: Okay, I understand. Just one more question. I saw that in the code before today's commits, you would explicitly delete the request object, instead of relying in RefCountedObj.  Why didn't you kept that approach?

trociny: Because explicitly deleting does not work for unit tests (mock objects). Note, the request objects are always created by `create()` static method, and see how it is overrode in mock, e.g: https://github.com/ceph/ceph/blob/master/src/test/rbd_mirror/test_mock_ImageReplayer.cc#L55

Another solution would be to provide `destroy()` static method and use it instead of explicitly deleting.

rjfd: Thanks for the explanation :-)

src/test/rbd_mirror/image_replayer/test_mock_BootstrapRequest.cc
```diff
@@ -48,7 +48,15 @@ struct ImageSync<librbd::MockImageCtx> {
     s_instance = this;
   }
 
-  MOCK_METHOD0(start, void());
+  ImageSync *put() {
```
dillaman: Should this be a ```void``` return type?

src/tools/rbd_mirror/BaseRequest.h
```diff
@@ -0,0 +1,39 @@
...
+#define CEPH_RBD_MIRROR_BASE_REQUEST_H
+
+#include "common/RefCountedObj.h"
+
+namespace rbd {
+namespace mirror {
+
+class BaseRequest : public RefCountedObject {
+public:
+  BaseRequest(CephContext *cct = nullptr, Context *on_finish = nullptr)
```
dillaman: Minor: can we remove the defaulted parameters?

src/tools/rbd_mirror/BaseRequest.h
```diff
@@ -0,0 +1,39 @@
...
+  virtual void send() = 0;
+  virtual void cancel() {}
+
+protected:
+  CephContext *m_cct;
+  Context *m_on_finish;
+
+  void finish(int r) {
+    if (m_cct) {
+      lsubdout(m_cct, rbd_mirror, 20) << "finish: r=" << r << dendl;
```
dillaman: Minor: include the class name in the output to trace where "finish: r=XYZ" comes from in the log

src/tools/rbd_mirror/image_replayer/BootstrapRequest.cc
```diff
@@ -457,6 +472,13 @@ void BootstrapRequest<I>::handle_update_client(int r) {
     return;
   }
 
+  if (m_canceled) {
+    derr << ": request canceled" << dendl;
```
dillaman: Minor: not sure these should rise to the level of error log messages

src/tools/rbd_mirror/image_replayer/BootstrapRequest.cc
```diff
@@ -530,11 +559,16 @@ void BootstrapRequest<I>::handle_get_remote_tags(int r) {
...
-           m_client_meta->state == librbd::journal::MIRROR_PEER_STATE_REPLAYING))) {
+    if (tag_data.mirror_uuid == librbd::Journal<>::ORPHAN_MIRROR_UUID &&
+	remote_tag_data.mirror_uuid == librbd::Journal<>::ORPHAN_MIRROR_UUID &&
+	remote_tag_data.predecessor_mirror_uuid == m_local_mirror_uuid) {
+      dout(20) << ": local image was demoted" << dendl;
+    } else if (tag_data.mirror_uuid == m_remote_mirror_uuid &&
+	       m_client_meta->state == librbd::journal::MIRROR_PEER_STATE_REPLAYING) {
+      dout(20) << ": local image is in clean replay state" << dendl;
+    } else if (m_client_meta->state == librbd::journal::MIRROR_PEER_STATE_SYNCING) {
+      dout(20) << ": previos sync was canceled" << dendl;
```
dillaman: Spelling

src/tools/rbd_mirror/image_replayer/BootstrapRequest.cc
```diff
@@ -568,18 +602,35 @@ void BootstrapRequest<I>::image_sync() {
...
+    m_image_sync_request->put();
+    m_image_sync_request = nullptr;
+  }
+
   if (r < 0) {
     derr << ": failed to sync remote image: " << cpp_strerror(r) << dendl;
     m_ret_val = r;
   }
 
+  if (m_canceled) {
```
dillaman: Perhaps move this above the preceding ```if``` statement since ```r``` most likely equals ```-ECANCELED``` and we don't need an error message.

dillaman:
The ```ImageCopyRequest``` state machine should check for cancelation in ```ImageCopyRequest::handle_object_copy```

dillaman:
The ```SnapshotCopyRequest``` state machine should check for cancelation during each step (unprotect, remove, create,  protect)

dillaman:
The ```ImageSync``` state machine should keep a ```BaseRequest``` pointer to ```ImageCopyRequest``` and ```SnapshotCopyRequest``` during those states so that they can be canceled.

trociny:
@dillaman 

> The ImageSync state machine should keep a BaseRequest pointer to ImageCopyRequest and SnapshotCopyRequest during those states so that they can be canceled.

I had to provide both  ImageCopyRequest and SnapshotCopyRequest (instead of the shared BaseRequest), because the mock classes are not inherited from BaseRequest.

dillaman:
@trociny Good point ... makes sense

trociny:
@dillaman 

> The ImageCopyRequest state machine should check for cancelation in ImageCopyRequest::handle_object_copy

I do it a little different: in send_next_object_copy m_cancelled is checked and m_ret_val is set. This way if it is cancelled in this point, it will wait for all object_copy requests terminate and will finish in send_flush_sync_point.

trociny:
@dillaman I think I have addressed all your comments. If you have an idea how tests could be improved to test bootstrap canceling, I will be glad implement it.

dillaman:
@trociny Thanks, I'll take a look this afternoon.  For testing, you can use the mock test cases to inject canceling at the various points. Sadly I don't have any test cases build for ```BootstrapRequest```, but for the ones that do have test cases, you can add a new test case, set an expectation for the state before it calls a cancelable state, and instead of directly completing the provided Context, first set the cancel flag and then invoke the context so that you can witness it cancel at the exact step you wish to test.

For example in ```test_mock_ImageCopyRequest.cc```, if the normal sequence is:

```
  InSequence seq;
  expect_get_object_count(mock_remote_image_ctx, 2);
  expect_get_object_count(mock_remote_image_ctx, 2);
  expect_update_client(mock_journaler, 0);
  expect_object_copy_send(mock_object_copy_request);
  expect_update_client(mock_journaler, 0);
```

I could inject a cancel request right before the ```object_copy_send``` state by tweaking the sequence to the following:

```
  InSequence seq;
  expect_get_object_count(mock_remote_image_ctx, 2);
  expect_get_object_count(mock_remote_image_ctx, 2);
  EXPECT_CALL(mock_journaler, update_client(_, _))
    .WillOnce(DoAll(InvokeWithoutArgs([request]() {
                      request->cancel();
                    }),
                    WithArg<1>(CompleteContext(r))));

  request->send();
  ASSERT_EQ(-ECANCELED, ctx.wait());
```

src/tools/rbd_mirror/ImageSync.cc
```diff
@@ -133,23 +144,48 @@ template <typename I>
...
     ImageSync<I>, &ImageSync<I>::handle_copy_snapshots>(this);
-  SnapshotCopyRequest<I> *request = SnapshotCopyRequest<I>::create(
+  m_snapshot_copy_request = SnapshotCopyRequest<I>::create(
     m_local_image_ctx, m_remote_image_ctx, &m_snap_map, m_journaler,
     m_client_meta, m_work_queue, ctx);
-  request->send();
+  m_snapshot_copy_request->get();
+  m_lock.Unlock();
+
+  update_progress("COPY_SNAPSHOTS");
```
dillaman: MInor: duplicate ```update_progress``` invocation in this method

src/tools/rbd_mirror/image_sync/SnapshotCopyRequest.cc
```diff
@@ -167,7 +179,19 @@ void SnapshotCopyRequest<I>::handle_snap_unprotect(int r) {
   CephContext *cct = m_local_image_ctx->cct;
   ldout(cct, 20) << ": r=" << r << dendl;
 
-  if (r < 0) {
+  if (r == 0)
```
dillaman: Minor: this is pretty repetitive for each step in the state machine.  Perhaps miove the logic to a common helper function so you can just run something like ```if (r == 0 && handle_cancellation()) { return; }```

dillaman:
@trociny Just a couple minor comments left plus the unit tests.

trociny:
@dillaman updated

dillaman:
lgtm

PR #8831, opened 2016-04-29
osd: clean heartbeat dispatch message

wjin:
this message should never happen in heartbeat messenger.


liewegas:
:+1:

PR #8832, opened 2016-04-29
common/Graylog.cc: use std move to avoid copy

stiopaa1:

### DIFF
--- a/src/common/Graylog.cc
+++ b/src/common/Graylog.cc
@@ -24,7 +24,7 @@ Graylog::Graylog(const SubsystemMap * const s, std::string logger)
       m_log_dst_valid(false),
       m_hostname(""),
       m_fsid(""),
-      m_logger(logger),
+      m_logger(std::move(logger)),
       m_ostream_compressed(std::stringstream::in |
                            std::stringstream::out |
                            std::stringstream::binary)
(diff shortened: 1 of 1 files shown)


src/common/Graylog.cc
```diff
@@ -24,7 +24,7 @@ Graylog::Graylog(const SubsystemMap * const s, std::string logger)
       m_log_dst_valid(false),
       m_hostname(""),
       m_fsid(""),
-      m_logger(logger),
+      m_logger(std::move(logger)),
```
tchaikov: logger is not a rvalue *reference*. so this does not help with the performance. and neither do i think it's a hotspot here.

cbodley: > logger is not a rvalue _reference_

i'm not sure what you're saying here. `m_logger(logger)` does call the `std::string` copy ctor, and adding `std::move()` causes it to use the move ctor instead

your point remains that this is only called once on initialization, so it won't ever show up in the profiler. but a copy is a copy, and I'd prefer to move when it's appropriate

tchaikov: sorry, i was wrong. thought that the type of `logger` should be `std::string&&`.

> but a copy is a copy, and I'd prefer to move when it's appropriate

makes sense to me.

tchaikov:
agree with @cbodley , lgtm also.

PR #8833, opened 2016-04-29
osd/PG: update info.stats.* mappings on split

liewegas:
These are updated in the init and start_peering_interval paths, but not
on split.

Fixes: http://tracker.ceph.com/issues/15523


src/osd/PG.cc
```diff
@@ -2337,6 +2337,12 @@ void PG::split_into(pg_t child_pgid, PG *child, unsigned split_bits)
   if (get_primary() != child->get_primary())
     child->info.history.same_primary_since = get_osdmap()->get_epoch();
 
+  info.stats.up = up;
```
athanatos: child->info?

PR #8834, opened 2016-04-29
osd: set pg role more carefully

liewegas:


athanatos:
lgtm

src/osd/OSD.cc
```diff
@@ -3480,6 +3479,10 @@ void OSD::handle_pg_peering_evt(
       PG::_create(*rctx.transaction, pgid, pgid.get_split_bits(pp->get_pg_num()));
       PG::_init(*rctx.transaction, pgid, pp);
 
+      int role = osdmap->calc_pg_role(whoami, acting, acting.size());
+      if (!pp->is_replicated() && pgid.shard != whoami)
+	role = -1;
```
xiexingguo: I see you hard coded pg role to be -1, 0, 1 for ec pgs(https://github.com/ceph/ceph/pull/8834/commits/dfce95d124824b69a944983289be5630076d29fa#diff-fa6c2eba8356ae1442d1bf749beacfdfR3228), shall we do this here, too?

liewegas:
repushed.. earlier version didn't even compile

xiexingguo:
looks good now:-)

PR #8836, opened 2016-04-29
mon/OSDMonitor: avoid underflow in reweight-by-utilization if max_change=1

athanatos:
Fixes: http://tracker.ceph.com/issues/15655


athanatos:
Passed tests sjust@teuthology:/a/samuelj-2016-04-28_23:23:57-rados-wip-sam-testing-distro-basic-smithi

PR #8837, opened 2016-04-29
mon/OSDMonitor: max_osds must be > 0 for reweight-by-*

liewegas:

### DIFF
--- a/src/mon/OSDMonitor.cc
+++ b/src/mon/OSDMonitor.cc
@@ -7645,6 +7645,11 @@ done:
     }
     int64_t max_osds = g_conf->mon_reweight_max_osds;
     cmd_getval(g_ceph_context, cmdmap, "max_osds", max_osds);
+    if (max_osds <= 0) {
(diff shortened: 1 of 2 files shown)


src/mon/OSDMonitor.cc
```diff
@@ -7645,6 +7645,11 @@ bool OSDMonitor::prepare_command_impl(MonOpRequestRef op,
     }
     int64_t max_osds = g_conf->mon_reweight_max_osds;
     cmd_getval(g_ceph_context, cmdmap, "max_osds", max_osds);
+    if (max_osds <= 0) {
+      ss << "max_osds " << max_osds << " msut be positive";
```
xiexingguo: typo. msut->must.

xiexingguo:
I have corrected this in https://github.com/ceph/ceph/pull/8758 too, but I don't add these test cases:(

src/mon/OSDMonitor.cc
liewegas: thanks, fixed!

PR #8841, opened 2016-04-30
ceph_time: add operator<< to allow output

yuyuyu101:

### DIFF
--- a/src/common/ceph_time.cc
+++ b/src/common/ceph_time.cc
@@ -103,4 +103,32 @@ namespace ceph {
     m.unsetf(std::ios::right);
     return m;
   }
+
+  std::ostream& operator<<(std::ostream& m, const coarse_mono_time& t) {
+    return m << std::chrono::duration<double>(t.time_since_epoch()).count()
+             << "s";
+  }
+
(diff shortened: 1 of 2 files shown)


tchaikov:
```c++
#include <iostream>
#include <type_traits>

using namespace std;

struct coarse {
  static constexpr const bool is_steady = true;
};

struct mono_coarse {
  static constexpr const bool is_steady = true;
};

struct real {
  static constexpr const bool is_steady = false;
};

template<typename T, typename enable_if<T::is_steady>::type* = nullptr>
std::ostream& 
operator<<(std::ostream& m, const T& t)
{
  return m << "s" << endl;
}

template<typename T, typename enable_if<!T::is_steady>::type* = nullptr>
std::ostream&
operator<<(std::ostream& m, const T& t)
{
  return cout << "tm" << endl;
}

int main()
{
  coarse c;
  cout << c;
  real r;
  cout << r;
  mono_coarse mc;
  cout << mc;
}
```

tchaikov:
@yuyuyu101 could you please take a look at #8850? it tries to follow the idea i proposed here. thanks

tchaikov:
closing as #8850 supersedes this PR.

PR #8844, opened 2016-04-30
common/Graylog.h: remove lexical cast

stiopaa1:
We can use std::to_string instead of boost's lexical_cast, which saves
us from including lexical_cast header.


tchaikov:
lgtm

PR #8845, opened 2016-04-30
hammer: osd: rados cppool omap to ec pool crashes osd

smithfarm:
http://tracker.ceph.com/issues/15647


smithfarm:

athanatos:
lgtm

PR #8848, opened 2016-04-30
[DNM] hammer: kclient: intermittent log warnings client.XXXX isn't responding to mclientcaps(revoke)

smithfarm:
http://tracker.ceph.com/issues/13932

### DIFF
--- a/src/client/Client.cc
+++ b/src/client/Client.cc
@@ -2302,14 +2302,16 @@ void Client::send_reconnect(MetaSession *session)
       bufferlist flockbl;
       _encode_filelocks(in, flockbl);
 
-      in->caps[mds]->seq = 0;  // reset seq.
-      in->caps[mds]->issue_seq = 0;  // reset seq.
-      in->caps[mds]->mseq = 0;  // reset seq.
+      Cap *cap = in->caps[mds];
+      cap->seq = 0;  // reset seq.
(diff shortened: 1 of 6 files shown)


smithfarm:
test this please

smithfarm:
```
mds/Locker.cc: In member function 'void Locker::handle_client_caps(MClientCaps*)':
mds/Locker.cc:2471:2: error: 'session' was not declared in this scope
  session->have_completed_flush(m->get_client_tid())) {
  ^
```

gregsfortytwo:
looks like this would also require 0ca7a49ba8294a9e9406fcda58221fd8c89a7491. That changes SessionMap encoding so you'd need to see if that's a safe change or not. (If the existing hammer already has the encoding version that patch moves from, it should be fine.)

smithfarm:
@gregsfortytwo I tried, but 0ca7a49 is non-trivial to cherry-pick. I rebased the PR, but that's all I'm capable of ATM. What do you suggest we do next?

gregsfortytwo:
I'll try and take a look at some point, but since we have (stable) Jewel out the backport to Hammer isn't as interesting as it used to be for FS bits.

gregsfortytwo:
I don't think I'm ever going to get to this unless we start seeing issues we decide are important, so closing...if somebody else wants to take a look, feel free to reopen!

PR #8854, opened 2016-05-01
PGMonitor: fix %USED field when ceph df

XinzeChi:
We should use the capaticy of the osds belonged to the pool
instead of the total capactity of the cluster.


XinzeChi:
close it first.

PR #8856, opened 2016-05-01
xio: provide keepalive(s) and their ack(s)

avnerbh:
Till now XIO in Ceph could only send Ceph messages, but not notifications like keepalive/ack.  Hence, Ceph with XIO was not robust enough.

From now on, the first byte in the accelio header of any outgoing accelio message will carry the 'tag' of that notification (either data message, keepalive, ack, and so on).
For command notifications (i.e., keepalive/ack) - we'll put all the payload in accelio header since accelio header can easily contain even 200 bytes (which is more than enough for us).

Implementation :
0. add tag to the header of each outgoing accelio message (and switch according to it upon msg receival)
1. splits XioMsg into 2 classes: base XioSend and derived XioMsg.
2. Derive a new class from XioSend - XioCmd for the new notifications (keepalive,ack,...)

### DIFF
--- a/src/msg/xio/XioMsg.cc
+++ b/src/msg/xio/XioMsg.cc
@@ -34,3 +34,12 @@ int XioDispatchHook::release_msgs()
   assert(r);
   return r;
 }
+
+/*static*/ size_t XioMsgHdr::get_max_encoded_length() {
+  ceph_msg_header _ceph_msg_header;
+  ceph_msg_footer _ceph_msg_footer;
+  XioMsgHdr hdr (_ceph_msg_header, _ceph_msg_footer);
+  const std::list<buffer::ptr>& hdr_buffers = hdr.get_bl().buffers();
+  assert(hdr_buffers.size() == 1); /* accelio header is small without scatter gather */
+  return hdr_buffers.begin()->length();
+}
--- a/src/msg/xio/XioMsg.cc
+++ b/src/msg/xio/XioMsg.cc
@@ -43,3 +43,8 @@ int XioDispatchHook::release_msgs()
   assert(hdr_buffers.size() == 1); /* accelio header is small without scatter gather */
   return hdr_buffers.begin()->length();
 }
+
+void XioMsg::print_debug(CephContext *cct, const char *tag) const {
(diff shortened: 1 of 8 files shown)


avnerbh:
Hi Matt,

Can you please review?  I believe all code is okay

TIA,
  Avner

avnerbh:
It was tested using a vstart cluster and using a real cluster. @mattbenjamin is there anything else needed here?

cbodley:
I was under the impression that accelio was handling keepalive on its own, so we didn't need to implement it in Ceph. Does that sound familiar to you @mattbenjamin @vuhuong?

gregsfortytwo:
If Accelio is doing its own keepalives then we need an interface for querying the same sorts of information from the XioMessenger that we use keepalives to provide in the SimpleMessenger. :) We talked about this some at the Ceph.Next offsite but didn't write down as much as I thought we had: http://pad.ceph.com/p/messenger

avnerbh:
Accelio handles keepalive to Accelio layer of peer and it notifies XioMessenger about timeouts. However, **Accelio's keepalives only serve for resource claiming of a timeout'ed connection** 
(I believe you can get similar functionality with TCP sockets using "_setsockopt SO_KEEPALIVE_").  At any case, **XioMessenger doesn't use any API for notifying dispatchers about this timeout.**

**Please note that my pull requests provide 2 functions:**
1. It allows XioMesssenger to send/receive any notification and not only notifications with tag=CEPH_MSGR_TAG_MSG.
2. It allows Ceph to use KEEPALIVE/KEEPALIVE2 with XioMessenger in the same way that Ceph uses keepalive with any other messenger.

src/msg/xio/XioMsg.cc
```diff
@@ -34,3 +34,12 @@ int XioDispatchHook::release_msgs()
   assert(r);
   return r;
 }
+
+/*static*/ size_t XioMsgHdr::get_max_encoded_length() {
```
mattbenjamin: why /* static */ ?

mattbenjamin:
@gregsfortytwo what's here should be unobjectionable, its implementing send_keepalive()

mattbenjamin:
(will merge when github behaves, or by hand if necessary)

PR #8857, opened 2016-05-01
osd: handle boot racing with NOUP set

liewegas:
This is a follow-on to 7139a232d26beef441ffbc13bc087baab3505ea8,
which handled the NOUP set + clear case when the OSD found out
about the flag being cleared.  However, it's possible that the
flag will get cleared but the OSD won't get a map update (because
it hasn't subscribed and is not doing any work).

This means that it is *more* likely than before that we will
restart the boot process even though the OSD did successfully
mark us up.  However, as before, it is unavoidable because there
is no notification of whether our boot request succeeds or not.

And it is still mostly harmless (an extra mark down + up cycle).

### DIFF
--- a/src/osd/OSD.cc
+++ b/src/osd/OSD.cc
@@ -6794,9 +6794,9 @@ void OSD::_committed_osd_maps(epoch_t first, epoch_t last, MOSDMap *m)
       }
     }
 
-    if (osdmap->test_flag(CEPH_OSDMAP_NOUP) &&
-	!newmap->test_flag(CEPH_OSDMAP_NOUP)) {
-      dout(10) << __func__ << " NOUP flag cleared in " << newmap->get_epoch()
+    if (osdmap->test_flag(CEPH_OSDMAP_NOUP) !=
+	newmap->test_flag(CEPH_OSDMAP_NOUP)) {
+      dout(10) << __func__ << " NOUP flag changed in " << newmap->get_epoch()
 	       << dendl;
       if (is_booting()) {
 	// this captures the case where we sent the boot message while


src/osd/OSD.cc
```diff
@@ -6807,6 +6807,12 @@ void OSD::_committed_osd_maps(epoch_t first, epoch_t last, MOSDMap *m)
 	// that is harmless (boot will just take slightly longer).
 	do_restart = true;
       }
+      if (!osdmap->test_flag(CEPH_OSDMAP_NOUP) &&
+	  newmap->test_flag(CEPH_OSDMAP_NOUP)) {
```
xiexingguo: Can we merge these two branches into just one? E.g.,
```c++ 
if (osdmap->test_flag(CEPH_OSDMAP_NOUP) != newmap->test_flag(CEPH_OSDMAP_NOUP)) {
  dout(10) << __func__ << " NOUP flag changed in " << newmap->get_epoch() << dendl;
  do_restart = true;
}

xiexingguo:
Otherwise looks good

yangruifeng:
@liewegas i met the same problem.

is the following changes reasonable?

	diff --git a/src/osd/OSD.cc b/src/osd/OSD.cc
	index de3de5a..8db5e8c 100644
	--- a/src/osd/OSD.cc
	+++ b/src/osd/OSD.cc
	@@ -4135,7 +4135,8 @@ void OSD::heartbeat()
	// hmm.. am i all alone?
	dout(30) << "heartbeat lonely?" << dendl;
	if (heartbeat_peers.empty()) {
	-    if (now - last_mon_heartbeat > cct->_conf->osd_mon_heartbeat_interval && is_active()) {
	+    if (now - last_mon_heartbeat > cct->_conf->osd_mon_heartbeat_interval &&
	+        (is_active() || is_booting()) {
		last_mon_heartbeat = now;
		dout(10) << "i have no heartbeat peers; checking mon for new map" << dendl;
		osdmap_subscribe(osdmap->get_epoch() + 1, false);

liewegas:
I think that would work for this case, but I prefer the more direct approach within the boot and map code.  The dependency on a onetime map subscription is easy to forget about, and who knows what changes the heartbeat code will go through in the future.

PR #8858, opened 2016-05-01
xio: Rename XioRsp to XioCompletion

Adirl:
since this element is not causing real response message on
the wire it was decided to change the name to XioCompletion to
explain that the transaction is completed and the memory
associated to it can be released by accelio.

Signed-off-by: Adir Lev <adirl@mellanox.com>


Adirl:
Tested using vstart cluster. @mattbenjamin is there anything else needed here?

mattbenjamin:
@Adirl I'm fine with the name change; will merge shortly

mattbenjamin:
(merged by hand)

PR #8859, opened 2016-05-01
include/rados: fix scrub-map definition

liewegas:
In commit 40add6a63defd09b47c594171c0f72906d218cc6 we
changed the scrub-map op code, breaking compatibility.


tchaikov:
lgtm. this fix the interoperability with clients before 40add6a.

PR #8861, opened 2016-05-01
fix bug for OP_TRY_RENAME

majianpeng:


majianpeng:
@liewegas . For https://github.com/ceph/ceph/pull/8861/commits/6b3141efcd72705601bbc3959eaaaf10ed9a7414, i'm not sure.  Could you give comments?

athanatos:
@majianpeng I'm confused, what happens if the source object doesn't exist?  It needs to return 0 in that case.

majianpeng:
@athanatos .  If source object doesn't exist. It don't call this code and return -ENOENT.
See the following code:
// object operations
    RWLock::WLocker l(c->lock);
    OnodeRef &o = ovec[op->oid];
    if (!o) {
      // these operations implicity create the object
      bool create = false;
      if (op->op == Transaction::OP_TOUCH ||
          op->op == Transaction::OP_WRITE ||
          op->op == Transaction::OP_ZERO) {
        create = true;
      }    
      ghobject_t oid = i.get_oid(op->oid);
      o = c->get_onode(oid, create);
      if (!create) {
        if (!o || !o->exists) {
          dout(10) << __func__ << " op " << op->op << " got ENOENT on "
                   << oid << dendl;
          r = -ENOENT;
          goto endop;
        }    
      }    
    }    

And _rename don't return -ENOENT.

athanatos:
@majianpeng Ok, for that op, it needs to return 0 if the object doesn't exist.  Can you fix that?

tchaikov:
@majianpeng ping?

liewegas:
@majianpeng ping

dachary:
feel free to re-open once you have time to work on it

tchaikov:
> Ok, for that op, it needs to return 0 if the object doesn't exist. Can you fix that?

`OP_TRY_RENAME` is already fixed in latest master.

but the behavior of `OP_COLL_MOVE_RENAME` in bluestore is different from filestore, where `-ENOENT` is not considered okay. see https://github.com/ceph/ceph/blob/master/src/os/filestore/FileStore.cc#L2816. @liewegas is this expected?

liewegas:
The FileStore semantics are hard to work out because it has to deal with replay; bluestore is a better point of reference.

I confess, though, I don't understand what TRY_RENAME is for or why it exists.  @athanatos ?

athanatos:
The intention is that OP_TRY_RENAME return 0 if the source does not exist.

tchaikov:
closing, as i think we've settled this problem.

PR #8863, opened 2016-05-01
osd/OSD.cc: change shared_ptr to unique_ptr

stiopaa1:
There is no need for osr to be a shared_ptr.


tchaikov:
lgtm

athanatos:
This is done once during mkfs.  I do not think this is worth the code churn.

PR #8864, opened 2016-05-01
qa/workunits/rbd: fix rbd-mirror log file name

trociny:


trociny:
@jdurgin This is a fuxup to 5fe64fa.

jdurgin:
good catch

PR #8866, opened 2016-05-02
jewel: rbd-mirror can crash if start up is interrupted

Abhishekvrshny:
http://tracker.ceph.com/issues/15660


dachary:
test this please

PR #8867, opened 2016-05-02
jewel: Parent image is closed twice if error encountered while opening

Abhishekvrshny:
http://tracker.ceph.com/issues/15661


dachary:
test this please

PR #8873, opened 2016-05-02
xio: use accelio api to get event str

roidayan:
Should not count on current events order in accelio.
This can lead to logging the wrong event string.


mattbenjamin:
This is up to Accelio API, agree

PR #6249, opened 2015-10-13
LTTng-UST tracing should be dynamically enabled

dillaman:

PR #8876, opened 2016-05-02
global: don't link lttng into libglobal

kmroz:
Rely on dynamic initialization instead. Linking lttng in this way had
the unfortunate side effect of causing radosgw to segfault (when
daemonized) during sigterm processing (ie. during lttng_ust_exit()).

This was originally removed in 638738f, but accidentally re-added via
5f61d36.

### DIFF
--- a/src/global/Makefile.am
+++ b/src/global/Makefile.am
@@ -6,9 +6,7 @@ libglobal_la_SOURCES = \
 	common/TrackedOp.cc
 
 libglobal_la_LIBADD = $(LIBCOMMON)
-if WITH_LTTNG
-libglobal_la_LIBADD += -ldl -llttng-ust
-endif
+
 noinst_LTLIBRARIES += libglobal.la
 
 noinst_HEADERS += \


smithfarm:
My only concern is whether `liblttng-ust-fork` is packaged together with `liblttng-ust` in all the supported distros?

smithfarm:
pushed to gitbuilder as "kmroz-wip-fix-rgw-lttng-segfault"

kmroz:
@smithfarm - Thanks Nathan!

kmroz:
@smithfarm - Ceph gitbuilder is happy with the change.

jdurgin:
perhaps we should go with dynamic initialization in rgw to match the libs and osd:

https://github.com/ceph/ceph/pull/6249

ktdreyer:
:+1: to dynamic initialization

kmroz:
@jdurgin / @ktdreyer - Sure thing. I will take a look.

kmroz:
@jdurgin - Looks that https://github.com/ceph/ceph/pull/6249/commits/638738fff1d14f789129011e6c8d6ddb2e29a777 was not properly fully applied to master, leaving behind:

```makefile
if WITH_LTTNG
libglobal_la_LIBADD += -ldl -llttng-ust -llttng-ust-fork
endif
```
from https://github.com/ceph/ceph/commit/811898912f3e43d7312efe5d06c496e1e7e22fa4. Was this on purpose?

jdurgin:
@dillaman do you remember if this keeping the libglobal makefile change was intentional?

dillaman:
@jdurgin Probably missed because at the time because that commit wasn't applicable.

dillaman:
@jdurgin Actually, the log shows that it was applied to master but was reverted by a later merge.

dillaman:
Commit 5f61d36 put it back (by accident it looks like)

jdurgin:
so it seems fine to remove it - let's do that and make sure it passes on the gitbuilders

kmroz:
@jdurgin / @dillaman - PR updated. Would either one of you be kind enough to re-push to gitbuilders, please?

jdurgin:
@kmroz pushed as kmroz-wip-fix-rgw-lttng-segfault

kmroz:
@jdurgin - Great, thanks.

smithfarm:
Guys, I was just looking at 811898912f3e43d7312efe5d06c496e1e7e22fa4 and noticed it says "Backport: hammer" but was never backported. It appears to have been missed, but I guess not backporting it was the right thing to do, after all?

jdurgin:
@smithfarm yes, that shouldn't be backported at this point. I think for hammer we're not compiling with lttng anyway, and at this point it doesn't seem worth trying to backport all the subsequent fixups to make it work there.

kmroz:
Gitbuilders appear satisfied.

smithfarm:
@kmroz Could you open a tracker issue for this, set `Backport: jewel` and `Status: Pending Backport` ? Thanks.

kmroz:
@smithfarm - It was already backported to jewel.

PR #8880, opened 2016-05-02
radosgw-admin: add missing 'zonegroup remove'

cbodley:
'zone delete' is currently the only way to remove a zone from a zonegroup, but it can only be run on that zone's local cluster. 'zonegroup remove' is needed to remove a zone from its zonegroup, a) from another cluster, or b) without deleting the zone itself.

### DIFF
--- a/src/rgw/rgw_admin.cc
+++ b/src/rgw/rgw_admin.cc
@@ -3300,7 +3300,7 @@ int main(int argc, char **argv)
             cerr << "WARNING: failed to initialize zonegroup " << zonegroup_name << std::endl;
             continue;
           }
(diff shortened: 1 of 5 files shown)


yehudasa:
@cbodley lgtm, maybe add a test for it?

cbodley:
@yehudasa would you recommend testing this with test/rgw/test_multi.py? or ceph-qa-suite?

yehudasa:
@cbodley test_mutli.py probably. We'll need to get it integrated with teuthology eventually.

cbodley:
updated with test_zonegroup_remove() in test_multi.py

PR #8883, opened 2016-05-02
librbd: clone fails if config overridden to use image format 1 by default

dillaman:


src/librbd/internal.cc
```diff
@@ -1361,10 +1362,16 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force) {
   {
     CephContext *cct = (CephContext *)p_ioctx.cct();
     ldout(cct, 20) << "clone " << &p_ioctx << " name " << p_name << " snap "
-		   << p_snap_name << "to child " << &c_ioctx << " name "
+		   << p_snap_name << " to child " << &c_ioctx << " name "
 		   << c_name << " opts = " << c_opts << dendl;
 
-    uint64_t format = cct->_conf->rbd_default_format;
+    bool defaut_format_set;
```
jdurgin: defaut?

jdurgin: otherwise lgtm

dillaman:
@jdurgin Thx -- fixed.

PR #8886, opened 2016-05-02
rgw: check for short_zone_id collisions on period update

cbodley:


src/rgw/rgw_rados.cc
```diff
@@ -1614,6 +1614,19 @@ void RGWPeriodMap::decode(bufferlist::iterator& bl) {
...
+static uint32_t gen_short_zone_id(const std::string zone_id)
+{
+  unsigned char md5[CEPH_CRYPTO_MD5_DIGESTSIZE];
+  MD5 hash;
+  hash.Update((const byte *)zone_id.c_str(), zone_id.size());
+  hash.Final(md5);
+
+  uint32_t short_id;
+  memcpy((char *)&short_id, md5, sizeof(short_id));
+  return std::max(short_id, 1u);
```
cbodley: @yehudasa do you remember why this bit was necessary?
```c++
        if (short_id == 0) {
          ++short_id;
        }
```

yehudasa: @cbodley probably because I wanted to be designate 0 as a no-zone-id assigned. E.g., when an object was created on a zone that didn't have a zone id, then we'd get a 0.

cbodley: ok, thanks

PR #8887, opened 2016-05-02
osdc: send error to recovery waiters on shutdown

jcsp:
...instead of sending them '0', which gets things
confused.

Fixes: http://tracker.ceph.com/issues/15689


ukernel:
looks good

gregsfortytwo:
That constituted a Reviewed-by, right @ukernel?

jcsp:
Passing tests here: http://pulpito.ceph.com/jspray-2016-05-12_04:34:36-fs-wip-jcsp-testing-20160511---basic-smithi/

PR #8891, opened 2016-05-03
test/encoding/readable.sh: use bash instead of sh

tchaikov:
aside from the `source` statement, the readable.sh is POSIX shell
script. we can just "." the referenced helper script. but to be
consistent with other encoding/*.sh scripts, let's just change the
shebang line to "#!/bin/bash".

Signed-off-by: Kefu Chai <tchaikov@gmail.com>


badone:
LGTM

PR #8892, opened 2016-05-03
rgw: cleanup: use boost::algorithm::starts_with to replace our own implementations

flyd1005:
Don't reinvent the wheel.

Signed-off-by: Li Peng <lip@dtdream.com>


liewegas:
test this please

mattbenjamin:
+1  I definitely like making more use of boost::algorithm;  I think we have no specific absoluteURI test;  maybe @cbodley can add this to his current testing branch before next run (?)

PR #8902, opened 2016-05-03
mon/OSDMonitor: kill dead code

xiexingguo:
These codes are not applicable any more,
and removing unnecessary code is always the preferred option.

### DIFF
--- a/src/mon/OSDMonitor.cc
+++ b/src/mon/OSDMonitor.cc
@@ -539,14 +539,12 @@ int OSDMonitor::reweight_by_utilization(int oload,
     int num_osd = MIN(1, pgm.osd_stat.size());
     if ((uint64_t)pgm.osd_sum.kb * 1024 / num_osd
 	< g_conf->mon_reweight_min_bytes_per_osd) {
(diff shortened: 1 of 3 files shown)


src/mon/OSDMonitor.cc
```diff
@@ -5897,7 +5897,7 @@ bool OSDMonitor::prepare_command_impl(MonOpRequestRef op,
     }
 
     err = newcrush.adjust_subtree_weightf(g_ceph_context, id, w);
-    if (err < 0)
+    if (err <= 0)
```
liewegas: This might not be right.  If the pending_crush isn't committed yet, even though it is a no-op, we can't reply because it isn't committed yet.

liewegas: Better to wait for a commit cycle in the rare case that it is in flight then to add a bunch of code to tell whether it is not in flight.

liewegas:
The dead code commit looks good.

xiexingguo:
Thanks for the review and dropped than change. Have a nice day:-)

PR #8909, opened 2016-05-03
rgw: remove EEXIST error msg for ZoneCreate

theanalyst:
currently for any admin operations like user create etc. you would
always see:

`RGWZoneParams::create(): error creating default zone params: (17) File Exists`

in stdout as the debug level is set to 0, which doesn't make much sense
for an end user, so increase the debug level when we see an EEXIST, for
other failures we print the error at debug level 0 itself

### DIFF
--- a/src/rgw/rgw_rados.cc
+++ b/src/rgw/rgw_rados.cc
@@ -1496,7 +1496,6 @@ int RGWZoneParams::create(bool exclusive)
 
   r = RGWSystemMetaObj::create(exclusive);
   if (r < 0) {
-    ldout(cct, 0) << "RGWZoneParams::create(): error creating default zone params: " << cpp_strerror(-r) << dendl;
     return r;
   }
 


liewegas:
@yehudasa @cbodley yeah?

cbodley:
I would prefer to just remove this error message.

There are currently two callers of `RGWZoneParams::create()`. One is `radosgw-admin 'zone create'`, which is already printing messages on failure: `cerr << "failed to create zone " << zone_name << ": " << cpp_strerror(-ret) << std::endl;`

The other is `RGWZoneParams::create_default()`, which is the one that's spamming you with EEXIST. `create_default()` doesn't print its own error messages, nor should it.

theanalyst:
@cbodley makes sense, I only kept it around for actual zone create, I'll update this

theanalyst:
@cbodley updated, also created http://tracker.ceph.com/issues/15720 in case we need a backport

cbodley:
@theanalyst sorry, I meant remove the dout entirely, not just for the EEXIST case. do you buy that?

theanalyst:
@cbodley oh I see, and the calls of `zone_params.create_default` does handle non -EEXIST errors as well, so yeah makes sense dropping it, updated it

cbodley:
:+1:

theanalyst:
@cbodley  ping?

cbodley:
merging, but not sure if this qualifies for backport - @yehudasa ?

yehudasa:
@cbodley probably yes for backport, but let's wait with this one to 10.2.2

bobrik:
@yehudasa, doesn't look like it was backported 😕

theanalyst:
@bobrik https://github.com/ceph/ceph/pull/9268

PR #8910, opened 2016-05-03
librbd: unlock image if journal error encountered during lock

dillaman:
Explicitly unlock to prevent a client from accidentally blacklisting
itself when retrying the lock.

Fixes: http://tracker.ceph.com/issues/15709


jdurgin:
maybe we should add a warning and not blacklist ourself in the locking code too

PR #8914, opened 2016-05-03
I have a hatred for warnings

adamemerson:
I wish all the compiler warnings in the repository had just one throat with my fingers clamped around it.

### DIFF
--- a/src/test/librados/c_write_operations.cc
+++ b/src/test/librados/c_write_operations.cc
@@ -211,7 +211,8 @@ TEST(LibRadosCWriteOps, WriteSame) {
   rados_write_op_writesame(op, "four", 4, 4 * 4, 0);
   ASSERT_EQ(0, rados_write_op_operate(op, ioctx, "test", NULL, 0));
   char hi[4 * 4];
-  ASSERT_EQ(sizeof(hi), rados_read(ioctx, "test", hi, sizeof(hi), 0));
+  ASSERT_EQ(sizeof(hi), static_cast<std::size_t>(
+		rados_read(ioctx, "test", hi,sizeof(hi), 0)));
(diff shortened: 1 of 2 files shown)


src/test/librados/c_write_operations.cc
```diff
@@ -211,7 +211,8 @@ TEST(LibRadosCWriteOps, WriteSame) {
   rados_write_op_writesame(op, "four", 4, 4 * 4, 0);
   ASSERT_EQ(0, rados_write_op_operate(op, ioctx, "test", NULL, 0));
   char hi[4 * 4];
-  ASSERT_EQ(sizeof(hi), rados_read(ioctx, "test", hi, sizeof(hi), 0));
+  ASSERT_EQ(sizeof(hi),
+	    (std::size_t) rados_read(ioctx, "test", hi, sizeof(hi), 0));
```
gregsfortytwo: We're trying to switch to proper C++ casts instead of C ones.

static_cast<std::size_t>( /*stuff */ )
etc

gregsfortytwo:
I support your drive to exterminate warnings!

src/test/librados/c_write_operations.cc
adamemerson: Fixed!

wjwithagen:
On 3-5-2016 19:00, Adam C. Emerson wrote:
> I wish all the compiler warnings in the repository had just one throat
> with my fingers clamped around it.
> 

I can second that.
Clang is throwing quite a fit on some parts of the code.

Be it only the mix between class en struct definitions.
Got a stack of them pilled up in my local wip, but do not want to commit
these as I'm still working on them.

--WjW

PR #8915, opened 2016-05-03
client: correct the client_lock order when cleanup

Vicente-Cheng:
As 448bea4, we need to handle a dependency
between client lock and admin socket lock.

Signed-off-by: Vicente Cheng <freeze.bilsted@gmail.com>

### DIFF
--- a/src/client/Client.cc
+++ b/src/client/Client.cc
@@ -490,9 +490,9 @@ int Client::init()
   int r = monclient->init();
   if (r < 0) {
     // need to do cleanup because we're in an intermediate init state
-    objecter->shutdown();
     timer.shutdown();
     client_lock.Unlock();
+    objecter->shutdown();
     objectcacher->stop();
     monclient->shutdown();
     return r;


gregsfortytwo:
How did you discover this? Was there an actual problem? Valgrind warning? Visual inspection?

This is probably okay but I'm a bit leery of it because the timer could conceivably grab the client lock and dispatch new messages into the Objecter.

Vicente-Cheng:
hi @gregsfortytwo ,
sorry for late replay, I think that is not an actual problem.
at first, I just want the consistency about this clean behavior and shutdown (as commit 448bea4)
because I think that behavior should be similar.

then you remind me that would have some problems, so I check the timer shutdown() on other places.
After that, I got confused.
Maybe you are right. I think I still need some time to realize how and when timer shutdown would be fine.

gregsfortytwo:
You know what, I think I was just confused when I made that comment; obviously if we shut down the Timer it won't be re-starting any more messages...

Vicente-Cheng:
hi @gregsfortytwo ,

Ahh, thanks for your explain.
So you mean if we shutdown the timer(), then we could not receive any more messages, right?
(just wanna to more clarify this behavior 

I still would like to check/realize that and thanks for your help!!!

jcsp:
This didn't throw up any issues in jspray-2016-05-25_19:00:36-fs-wip-jcsp-testing-20160526---basic-mira

PR #8917, opened 2016-05-03
civetweb: combine various submodule changes

liewegas:
There were some accidental reverts, fixes, re-reverts,
and follow-on changes that are hard to follow and in the
end did not include what we want.  Merge it all together.


adamemerson:
This is a fine and excellent PR. One of the included commits obviates the needs for my warning fix to Civetweb as well.

PR #8918, opened 2016-05-03
rgw: fix memory leaks

yehudasa:
and other valgrind issues

### DIFF
--- a/src/rgw/rgw_cr_rest.h
+++ b/src/rgw/rgw_cr_rest.h
@@ -25,6 +25,10 @@ public:
       path(_path), params(make_param_list(params)), result(_result)
   {}
 
+  ~RGWReadRESTResourceCR() {
+    request_cleanup();
+  }
+
   int send_request() {
     auto op = boost::intrusive_ptr<RGWRESTReadResource>(
         new RGWRESTReadResource(conn, path, params, NULL, http_manager));
@@ -55,6 +59,7 @@ public:
   void request_cleanup() {
     if (http_op) {
       http_op->put();
+      http_op = NULL;
     }
   }
 };
@@ -78,6 +83,10 @@ public:
       path(_path), params(make_param_list(_params)), result(_result),
       input(_input)
   {}
+
+  ~RGWPostRESTResourceCR() {
+    request_cleanup();
+  }
 
   int send_request() {
     auto op = boost::intrusive_ptr<RGWRESTPostResource>(
@@ -126,6 +135,7 @@ public:
   void request_cleanup() {
     if (http_op) {
       http_op->put();
+      http_op = NULL;
     }
   }
 };
--- a/src/rgw/rgw_cr_rest.h
+++ b/src/rgw/rgw_cr_rest.h
@@ -119,6 +119,7 @@ public:
       op->put();
       return ret;
     }
+    op->put();
     return 0;
   }
 
--- a/src/rgw/rgw_cr_rest.h
+++ b/src/rgw/rgw_cr_rest.h
@@ -39,6 +39,7 @@ public:
     if (ret < 0) {
       log_error() << "failed to send http operation: " << op->to_str()
           << " ret=" << ret << std::endl;
+      op->put();
       return ret;
     }
     std::swap(http_op, op); // store reference in http_op on success
@@ -51,8 +52,10 @@ public:
     if (ret < 0) {
       error_stream << "http operation failed: " << op->to_str()
           << " status=" << op->get_http_status() << std::endl;
+      op->put();
       return ret;
     }
+    op->put();
     return 0;
   }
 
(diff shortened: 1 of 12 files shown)


src/rgw/rgw_cr_rest.h
```diff
@@ -35,6 +39,7 @@ class RGWReadRESTResourceCR : public RGWSimpleCoroutine {
     if (ret < 0) {
       log_error() << "failed to send http operation: " << op->to_str()
           << " ret=" << ret << std::endl;
+      op->put();
```
cbodley: `op` is a `boost::intrusive_ptr`, so its dtor will already `put()` its ref when it goes out of scope here

src/rgw/rgw_cr_rest.h
```diff
@@ -47,14 +52,17 @@ class RGWReadRESTResourceCR : public RGWSimpleCoroutine {
     if (ret < 0) {
       error_stream << "http operation failed: " << op->to_str()
           << " status=" << op->get_http_status() << std::endl;
+      op->put();
       return ret;
     }
+    op->put();
```
cbodley: same as above, `op` will drop its ref when it goes out of scope

src/rgw/rgw_cr_rest.h
yehudasa: @cbodley the object is created with a refcount of 1, assigning it to the intrusive_ptr increases it to 2.

cbodley: oh, gross! `intrusive_ptr()`'s ctor takes a second `bool add_ref = true` argument, let's pass false to keep it at 1 then

liewegas: or, if we always use intrusive_ptr, change the object to init the ref 
count at 0.  (That's what we usually do.)

cbodley:
I'm seeing this segfault in teuthology and in local testing:
```
 1: (ceph::BackTrace::BackTrace(int)+0x2d) [0xb5da69]
 2: /home/cbodley/ceph/build/bin/radosgw() [0xb5d1bb]
 3: (()+0x109f0) [0x7f6efcdc69f0]
 4: (RGWRados::ctx()+0xc) [0x84c9f6]
 5: (RGWSyncErrorLogger::log_error_cr(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, unsigned int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)+0x4a) [0xaa5796]
 6: (RGWDataSyncSingleEntryCR::operate()+0x4bd) [0xae2d91]
 7: (RGWCoroutinesStack::operate(RGWCoroutinesEnv*)+0x15e) [0x8e7296]
 8: (RGWCoroutinesManager::run(std::__cxx11::list<RGWCoroutinesStack*, std::allocator<RGWCoroutinesStack*> >&)+0x25a) [0x8e87d8]
 9: (RGWCoroutinesManager::run(RGWCoroutine*)+0x9b) [0x8e981d]
 10: (RGWRemoteDataLog::run_sync(int, rgw_data_sync_status&)+0x249) [0xacfa8f]
 11: (RGWDataSyncStatusManager::run()+0x33) [0x99da19]
 12: (RGWDataSyncProcessorThread::process()+0x6f) [0x99f547]
 13: (RGWRadosThread::Worker::entry()+0xd8) [0x94eef4]
 14: (Thread::entry_wrapper()+0xc1) [0x7f6f01c6c23b]
 15: (Thread::_entry_func(void*)+0x18) [0x7f6f01c6c170]
```
I added dout()s to the RGWSyncErrorLogger ctor and dtor, and a grep found these:
```
2016-05-04 15:52:42.902111 7f6eda129700 10 rgw meta sync: RGWSyncErrorLogger[0x1b41ae0]
2016-05-04 15:52:44.701446 7f6eda129700 10 rgw meta sync: ~RGWSyncErrorLogger[0x1b41ae0]
2016-05-04 15:53:02.902238 7f6eda129700 10 rgw meta sync: RGWSyncErrorLogger[0x1b419f0]
```
According to gdb, the RGWSyncErrorLogger's `this` pointer was the previously-destructed one, 0x1b41ae0.

cbodley:
re: segfault, `RGWRemoteDataLog::init()` is succeeding in the first call to `RGWDataSyncStatusManager::init()`, so the second call to `RGWRemoteDataLog::init()` is not passing the new error_logger into `sync_env.init()` because it sees `initialized == true`. Looks like `sync_env.init()` should go above the `if (initialized) { return 0; }` block.

yehudasa:
@cbodley, good catch. Yeah, either that, or have un-init RGWDataSyncStatusManager in `finalize()`

cbodley:
> or have un-init RGWDataSyncStatusManager in finalize()

yeah, that way we don't leave RGWRemoteDataLog's HTTPManager thread running

PR #8919, opened 2016-05-03
osd/OSD.h: change some data members to private

stiopaa1:
Move some data members of the OSDService from public to private access
specifier.


athanatos:
@stiopaa1  This is a good thought, but I'd rather you added private: public: where they already are so that these stay grouped near the things they are protecting.  OSDService is really a collection of nested sub-objects, but without actual names for the sub-objects.

stiopaa1:
@athanatos I've changed it to what you proposed

PR #8920, opened 2016-05-04
osd: fix deadlock in OSD::_committed_osd_maps

XinzeChi:
Fixes: #15701

### DIFF
--- a/src/osd/OSD.cc
+++ b/src/osd/OSD.cc
@@ -6993,9 +6993,9 @@ void OSD::_committed_osd_maps(epoch_t first, epoch_t last, MOSDMap *m)
         failure_pending.erase(it++);
       }
     }
-    osd_lock.Unlock();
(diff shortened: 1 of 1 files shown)


XinzeChi:
@liewegas , Do you think it is suitable to fix it?

liewegas:
This looks reasonable to me!  It would be really nice if there were a test for this.. I don't think we have anything that exercises the forced shutdown path.

yuyuyu101:
two possible cases to trigger force shutdown:
1. set osd_max_markdown_count to 1
2. manually set link down, so it can make rebind failed

tchaikov:
better off putting `Fixes: http://tracker.ceph.com/issues/15701` in commit message.

liewegas:
want to run this through one more time since there were some suspicous failures in the last run.

PR #8921, opened 2016-05-04
rbd-mirror: image sync needs to handle snapshot size and protection status

dillaman:


src/tools/rbd_mirror/image_sync/SnapshotCopyRequest.cc
```diff
@@ -79,31 +80,133 @@ void SnapshotCopyRequest<I>::send_snap_remove() {
...
+    librados::snap_t local_snap_id = *snap_id_it;
+    {
+      RWLock::RLocker local_snap_locker(m_local_image_ctx->snap_lock);
+      bool local_unprotected;
+      int r = m_local_image_ctx->is_snap_unprotected(local_snap_id,
+                                                     &local_unprotected);
+      if (r < 0) {
+        lderr(cct) << "failed to retrieve local snap unprotect status: "
+                   << cpp_strerror(r) << dendl;
+        finish(r);
```
trociny: Is it safe to run finish() holding snap_lock?

dillaman: Thx --- I'll fix

jdurgin:
lgtm

src/test/Makefile-client.am
```diff
@@ -317,6 +317,11 @@ noinst_HEADERS += \
 	test/librados_test_stub/TestIoCtxImpl.h
 noinst_LTLIBRARIES += librados_test_stub.la
 
+libjournal_test_mock_la_SOURCES = \
+	test/journal/mock/MockJournaler.cc
+libjournal_test_mock_la_CXXFLAGS = $(UNITTEST_CXXFLAGS)
```
dillaman: missing noinst_HEADERS for MockJournaler.h

jdurgin: thx, not sure why it built on my machine

dillaman: The annoyance of the dist tarball that the github make bit doesn't use

jdurgin: ah right, makes sense

PR #8922, opened 2016-05-04
lockdep: fix assertion expression if ran out of lock ids

runsisi:
we should assert that the `free_ids` is not empty


cbodley:
this is in a block of code that has already verified that free_ids is empty. the assertion is supposed to fail

xiexingguo:
LGTM

tchaikov:
:+1:

PR #8923, opened 2016-05-04
rpm: Fix SELinux relabel on fedora

branto1:
The SELinux userspace utilities stopped providing versions when they
switched to CIL language. We need to use a different technique to
relabel the files.


branto1:
This already incorporates the SELinux changes made in  #8720.

bachradsusi:
ltgm from SELinux point of view.

Note: we (Fedora SELinux maintainers) are working on guidelines how to ship modules in packages, which priority (-X 100) should be used and so on. The priority could change from 100 to something else in future. We'll let you know.

branto1:
This builds fine on gitbuilders, I think we can merge it.

PR #8924, opened 2016-05-04
mon/Monitor: fix memory leak; add leaked key to tracked_conf_key

xiexingguo:
http://tracker.ceph.com/issues/15793

### DIFF
--- a/src/mon/Monitor.cc
+++ b/src/mon/Monitor.cc
@@ -464,6 +464,7 @@ const char** Monitor::get_tracked_conf_keys() const
     "clog_to_graylog",
     "clog_to_graylog_host",
     "clog_to_graylog_port",
+    "host",
     "fsid",
     // periodic health to clog
     "mon_health_to_clog",
--- a/src/mon/Monitor.cc
+++ b/src/mon/Monitor.cc
@@ -3762,14 +3762,14 @@ void Monitor::handle_ping(MonOpRequestRef op)
   MPing *reply = new MPing;
   entity_inst_t inst = m->get_source_inst();
   bufferlist payload;
-  Formatter *f = new JSONFormatter(true);
+  boost::scoped_ptr<Formatter> f(new JSONFormatter(true));
   f->open_object_section("pong");
 
   list<string> health_str;
-  get_health(health_str, NULL, f);
+  get_health(health_str, NULL, f.get());
   {
     stringstream ss;
-    get_mon_status(f, ss);
+    get_mon_status(f.get(), ss);
   }
 
   f->close_section();
--- a/src/mon/Monitor.cc
+++ b/src/mon/Monitor.cc
@@ -86,6 +86,7 @@
 #include "common/config.h"
 #include "common/cmdparse.h"
 #include "include/assert.h"
+#include "include/compat.h"
 
 #define dout_subsys ceph_subsys_mon
 #undef dout_prefix
@@ -5018,7 +5019,7 @@ int Monitor::write_default_keyring(bufferlist& bl)
   err = bl.write_fd(fd);
   if (!err)
     ::fsync(fd);
-  ::close(fd);
+  VOID_TEMP_FAILURE_RETRY(::close(fd));
 
   return err;
 }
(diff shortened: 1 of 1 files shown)


src/mon/Monitor.cc
```diff
@@ -464,6 +464,7 @@ const char** Monitor::get_tracked_conf_keys() const
     "clog_to_graylog",
     "clog_to_graylog_host",
     "clog_to_graylog_port",
+    "host",
```
jecluis: can you please elaborate on why this is needed?

src/mon/Monitor.cc
```diff
@@ -3775,6 +3775,7 @@ void Monitor::handle_ping(MonOpRequestRef op)
   f->close_section();
   stringstream ss;
   f->flush(ss);
+  delete f;
```
tchaikov: i'd suggest use `unique_ptr` here.

src/mon/Monitor.cc
xiexingguo: These tracked keys work as filters, so we can react properly upon "host" configuration changing.
```c++
void Monitor::handle_conf_change(const struct md_config_t *conf,
                                 const std::set<std::string> &changed)
...
 if (changed.count("clog_to_monitors") ||
            changed.count("clog_to_syslog") ||
            changed.count("clog_to_syslog_level") ||
            changed.count("clog_to_syslog_facility") ||
            changed.count("clog_to_graylog") ||
            changed.count("clog_to_graylog_host") ||
            changed.count("clog_to_graylog_port") ||
            changed.count("host") ||
            changed.count("fsid"))
    {
        update_log_clients();
    }
```
E.g., if we change "host" only, the above logic won't catch it because "host" is not tracked.

xiexingguo:
Done, thanks for the review.

src/mon/Monitor.cc
tchaikov: i see now:  graylog prints host in its log entry, so `Monitor::update_log_clients()` also reads from conf.host and pass the `host` read from conf to graylog.

xiexingguo: Sorry, I don't verify this by myself. However I don't see a reason why we shall register all other keys but leaving "host" aside, which is weird:)
Anyway,  would you like me to drop or keep this?

xiexingguo: @tchaikov  Sorry, sir. I misunderstand your wonderful explanation and show no respect for your kind help. I'll fold these into the commit message.

tchaikov: @irq0 what do you think?

xiexingguo: @tchaikov  Thanks for ack and patience :)

jecluis: @tchaikov @xiexingguo I get the point of the commit, was merely looking for an explanation to a change that's not at all obvious in the given patch set.

xiexingguo: @jecluis Thank you for your attention. I also update the commit message correspondingly.

irq0: Looks good to me. Seems like I forgot the "host" key. Sorry and thanks for spotting and fixing.

tchaikov:
lgtm

src/mon/Monitor.cc
```diff
@@ -3762,14 +3762,14 @@ void Monitor::handle_ping(MonOpRequestRef op)
   MPing *reply = new MPing;
   entity_inst_t inst = m->get_source_inst();
   bufferlist payload;
-  Formatter *f = new JSONFormatter(true);
```
tchaikov: @xiexingguo could you file a ticket for this change? so we can backport the fix to jewel?

xiexingguo: @tchaikov Sure. http://tracker.ceph.com/issues/15793
But I doubt I shall update the commit message correspondingly, since yuri is already testing this, right?

src/mon/Monitor.cc
xiexingguo: @irq0 Thanks for your confirmation.

src/mon/Monitor.cc
tchaikov: i think it's okay. as you would change the commit message instead of the code.

xiexingguo: Okay, upgraded.

yuriw:
http://pulpito.ceph.com/yuriw-2016-05-10_10:56:12-rados-wip-yuri-testing-distro-basic-smithi/

PR #8925, opened 2016-05-04
client: fstat should take CEPH_STAT_CAP_INODE_ALL

ukernel:
Fixes: http://tracker.ceph.com/issues/15723
Signed-off-by: Noah Watkins <nwatkins@redhat.com>

### DIFF
--- a/src/client/Client.cc
+++ b/src/client/Client.cc
@@ -8668,16 +8668,16 @@ int Client::_fsync(Fh *f, bool syncdataonly)
   return _fsync(f->inode.get(), syncdataonly);
 }
 
-int Client::fstat(int fd, struct stat *stbuf) 
+int Client::fstat(int fd, struct stat *stbuf, int mask)
 {
   Mutex::Locker lock(client_lock);
-  tout(cct) << "fstat" << std::endl;
+  tout(cct) << "fstat mask " << hex << mask << std::endl;
   tout(cct) << fd << std::endl;
 
   Fh *f = get_filehandle(fd);
   if (!f)
     return -EBADF;
-  int r = _getattr(f->inode, -1);
+  int r = _getattr(f->inode, mask);
   if (r < 0)
     return r;
   fill_stat(f->inode, stbuf, NULL);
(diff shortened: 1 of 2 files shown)


gregsfortytwo:
This shouldn't get merged yet; @noahdesu reported he was getting cap deadlock behavior with it turned on. I doubt this is the actual trigger but we should work it out first.

noahdesu:
Anything I can do to help speed this up by getting logs etc... or is it pretty clear what's going on?

gregsfortytwo:
I haven't looked, have you @ukernel?

noahdesu:
I'm running this again and I'm not able to reproduce it. I think that the prior experiment had an unclean shutdown path in the application and starting it back up caused hang waiting around on timeouts or what not.

ukernel:
I can reproduce it by using following program. 
while this program is running, truncating 'testfile' on other client blocks infinitely. But the behaviour is constant with this patch or without this patch. 

<pre>
#define _FILE_OFFSET_BITS 64
#include <features.h>
#include <sys/types.h>
#include <sys/stat.h>
#include <fcntl.h>
#include <time.h>
#include <stdio.h>
#include <errno.h>
#include <cephfs/libcephfs.h>


int main(int argc, char *argv[]) {
        struct ceph_mount_info *cmount = NULL;

        ceph_create(&cmount, "admin");
        ceph_conf_read_file(cmount, NULL);
        ceph_mount(cmount, NULL);

        int fd = ceph_open(cmount, "testfile", O_RDWR|O_CREAT, 0666);
        int count = 0;
        time_t start = time(NULL);
        for (;;) {
                count++;
                struct stat st;
                int ret = ceph_fstat(cmount, fd, &st);
                if (ret < 0)
                        break;
                if (time(NULL) > start) {
                        printf("%d\n", count);
                        count = 0;
                        start = time(NULL);
                }
        }
        
        ceph_unmount(cmount);
        return 0;
}
</pre>

jcsp:
@ukernel  do you think we should wait for https://github.com/ceph/ceph/pull/8965 before merging this one?

ukernel:
no, I don't think we need to wait #8965

jcsp:

PR #8926, opened 2016-05-04
test/ec: build the libs only when 'make check'

branto1:
Currently, we are always building the erasure code libraries while we
need them only when 'make check' is run. Moving the test libraries to
check_LTLIBRARIES should fix this for us.

We no longer need to remove the libec libs manually, remove the lines
that do that.


smithfarm:
Excellent!

smithfarm:
Making a note to slip this into one of the packaging backport PRs.

PR #8927, opened 2016-05-04
AsyncConnection: continue to read when meeting EINTR

yuyuyu101:


yuyuyu101:
two very rare cases, and the later should only exists in theory. So I guess no need to backport

yuyuyu101:
merge into #8963

PR #8928, opened 2016-05-04
os/bluestore: initial transition to new extent structures

liewegas:


liewegas:
This builds and runs for a little while before falling over.

* we aren't deallocating partial blobs
* the COW logic has several subtle issues.

I don't think it's worth fixing either of these issues until we settle on a better interface and encapsulation for the extent/blob management.

PR #8929, opened 2016-05-04
ceph_volume_client: allow read-only authorization for volumes

ajarr:
Allow clients to be restricted to read-only mount of the volume
by restricting their ceph auth ID's MDS and OSD caps to read-only.

Fixes: http://tracker.ceph.com/issues/15614

Signed-off-by: Ramana Raja <rraja@redhat.com>

### DIFF
--- a/src/pybind/ceph_volume_client.py
+++ b/src/pybind/ceph_volume_client.py
@@ -576,12 +576,13 @@ class CephFSVolumeClient(object):
             else:
                 return self._get_ancestor_xattr(os.path.split(path)[0], attr)
 
-    def authorize(self, volume_path, auth_id):
+    def authorize(self, volume_path, auth_id, readonly=False):
         """
         Get-or-create a Ceph auth identity for `auth_id` and grant them access
         to
         :param volume_path:
         :param auth_id:
+        :param readonly:
         :return:
         """
 
@@ -594,8 +595,11 @@ class CephFSVolumeClient(object):
         # Now construct auth capabilities that give the guest just enough
         # permissions to access the share
         client_entity = "client.{0}".format(auth_id)
-        want_mds_cap = 'allow rw path={0}'.format(path)
-        want_osd_cap = 'allow rw pool={0} namespace={1}'.format(pool_name, namespace)
+        want_access_level = 'r' if readonly else 'rw'
+        want_mds_cap = 'allow {0} path={1}'.format(want_access_level, path)
+        want_osd_cap = 'allow {0} pool={1} namespace={2}'.format(
+            want_access_level, pool_name, namespace)
+
         try:
             existing = self._rados_command(
                 'auth get',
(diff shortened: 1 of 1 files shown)


src/pybind/ceph_volume_client.py
```diff
@@ -582,8 +583,11 @@ def authorize(self, volume_path, auth_id):
         # Now construct auth capabilities that give the guest just enough
         # permissions to access the share
         client_entity = "client.{0}".format(auth_id)
-        want_mds_cap = 'allow rw path={0}'.format(path)
-        want_osd_cap = 'allow rw pool={0} namespace={1}'.format(pool_name, namespace)
+        access_level = ('r' if readonly else 'rw')
+        want_mds_cap = 'allow {0} path={1}'.format(access_level, path)
+        want_osd_cap = 'allow {0} pool={1} namespace={2}'.format(access_level,
+                                                                 pool_name,
+                                                                 namespace)
```
ajarr: I'd need to check if the auth ID already has a conflicting access permission (check for 'ro' access if 'rw' access is requested) for the volume, and if so update the access level of the auth ID or error out based on a 'force' flag,  right?

ajarr:
Tested by https://github.com/ceph/ceph-qa-suite/pull/983

src/pybind/ceph_volume_client.py
xiaoxichen: I would prefer authorize is behave like  plus r or plus rw, means if existing ro and granting rw, it will be rw. existing rw and granting ro still rw. Also similar for deauthorize.

src/pybind/ceph_volume_client.py
```diff
@@ -660,8 +664,14 @@ def deauthorize(self, volume_path, auth_id):
         pool_name = self._get_ancestor_xattr(path, "ceph.dir.layout.pool")
         namespace = self.fs.getxattr(path, "ceph.dir.layout.pool_namespace")
 
-        want_mds_cap = 'allow rw path={0}'.format(path)
-        want_osd_cap = 'allow rw pool={0} namespace={1}'.format(pool_name, namespace)
+        # The auth_id might have read-only or read-write mount access for the
```
ajarr: As @xiaoxichen suggested , I think it'd be better to deauthorize based on access_level.
So the deauthorize function's signature would be,
deauthorize(self, volume_path, auth_id, access_level='rw').
'access_level=rw' to be compatible with the Mitaka version of cephfs_native driver.

src/pybind/ceph_volume_client.py
jcsp: I'm not sure I see the use case for being intelligent about combining ro/rw.  Why not just take whatever the user requested most recently as overwriting what is already present?  Are we expecting that e.g. Manila will benefit?

jcsp: Actually, come to think of it, I think this already behaves the way @xiaoxichen suggests, on the authorize() path.  If someone calls this twice for the same ID, once with readonly=True and once with readonly=False, the result will be two rules.  At the ceph-mds level, these will be interpreted as whichever is broader winning (i.e. if both are present, then it will be RW)

src/pybind/ceph_volume_client.py
jcsp: The ambiguity would be over whether when there is an 'r' rule and a 'rw' rule, should deauthorize(...'rw') remove both rules, or should it leave the 'r' rule in place.  I would suggest the latter, so that this is behaving the same way as authorize().

It would be great to have some tests for these kinds of cases, as a way of demonstrating the behaviour

jcsp:
Sorry for delay, I've added some comments now.   This looks basically fine, we just need clarity on how auth/deauth should handle overlapping r/rw cases -- probably whatever makes Manila implementation easier will be preferable.

ajarr:
Sorry! I managed to over-write the useful comments about authorize(). I think it makes sense to always ensure the access-level requested for an ID is enforced if authorize() returns with no error.

@jcsp , @xiaoxichen , what do you think of this new version?

src/pybind/ceph_volume_client.py
xiaoxichen: Agree on @jcsp . @ajarr , so maybe we would better to review the test case first?

jcsp:
OK, this looks like a clear consistent behaviour (any access updates for a given client ID will override any existing r/rw configuration).  Should be good to merge this once the test PR is updated.

jcsp:
(assuming this is no longer "DNM" from your point of view @ajarr?)

src/pybind/ceph_volume_client.py
```diff
@@ -603,18 +607,30 @@ def authorize(self, volume_path, auth_id):
...
-                    cap_tokens.append(want)
+            # Construct auth caps that if present might conflict with the desired
+            # auth caps.
+            unwanted_access_level = 'r' if want_access_level is 'rw' else 'rw'
+            unwanted_mds_cap = 'allow {0} path={1}'.format(unwanted_access_level, path)
+            unwanted_osd_cap = 'allow {0} pool={1} namespace={2}'.format(
+                unwanted_access_level, pool_name, namespace)
+
+            def cap_update(orig, want, unwanted):
+                # Updates the existing auth caps list such that there is a
```
ajarr: s/list//

ajarr:
@jcsp, sorry for the delay! The test PR has been updated.

PR #8931, opened 2016-05-04
DNM: Repair read

dzafman:


athanatos:
The client side methods can probably be simplified.  I think IoCtx::aio_repair_read can call IoCtx::aio_operate_repair_read which can add the assert and call Objecter::op_submit.  If useful, we can also add an operate_repair_read which wraps aio_operate_repair_read and a repair_read which wraps operate_repair_read.

athanatos:
I think we can simplify the client side further.  Below are the methods I see
in this PR in the format <method> -> <implemented using>

IoCtx::repair_read -> IoCtxImpl::repair_read
IoCtx::aio_repair_read -> IoCtxImpl::aio_repair_read
IoCtx::operate_repair_read -> IoCtxImpl::operate_repair_read

IoCtxImpl::repair_read -> IoCtxImpl::operate_repair_read
IoCtxImpl::aio_repair_read -> IoCtxImpl::aio_operate_repair_read
IoCtxImpl::aio_operate_repair_read -> Objecter::prepare_read_op
IoCtxImpl::operate_repair_read -> IoCtxImpl::operate_repair_read

First, I think we need IoCtx::aio_operate_repair_read.  As written, there is no
way for a librados user to implement an async repair read operation which both
fetches the attrs and a portion of the object data.  It's actually the only
required method: all of the others can be implemented in terms of it.  Second,
it seems like the only IoCtxImpl method we actually need is
IoCtxImpl::aio_operate_repair_read.  I think the right set of methods is
actually:

IoCtx::repair_read -> IoCtx::operate_repair_read
IoCtx::aio_repair_read -> IoCtx::aio_operate_repair_read
IoCtx::operate_repair_read -> IoCtx::aio_operate_repair_read
IoCtx::aio_operate_repair_read -> IoCtxImpl::aio_operate_repair_read

IoCtxImpl::aio_operate_repair_read -> Objecter::prepare_read_op

This way, all IoCtx methods funnel through one IoCtxImpl method.  This also
means that only one method needs to worry about prepare_assert_ops concept.

src/osdc/Objecter.cc
```diff
@@ -2705,6 +2705,9 @@ int Objecter::_calc_target(op_target_t *t, epoch_t *last_force_resend,
 	t->sort_bitwise,
 	sort_bitwise,
 	pg_t(prev_seed, pgid.pool(), pgid.preferred()))) {
+    // If the interval has changed, then don't resend for repair read
```
athanatos: We need to return an error to the callback, right?

jdurgin:
don't forget to add C and python versions - they're much more reusable

src/osd/ReplicatedPG.cc
```diff
@@ -5977,6 +5979,23 @@ int ReplicatedPG::do_osd_ops(OpContext *ctx, vector<OSDOp>& ops)
       }
       break;
 
+    case CEPH_OSD_OP_ASSERT_INTERVAL:
+      ++ctx->num_read;
+      {
+	epoch_t epoch = op.assert_interval.epoch;
+	epoch_t pg_epoch = get_epoch();
```
tchaikov: the osdmap's epoch might be greater than the one of a scrub happened in current interval. i'd suggest use `info.history.same_interval_since` instead.

athanatos: Hmm, he's doing that using pg_has_reset_since below.  Here, he's checking to make sure that the pg has a new enough map to handle this message.  I *think* that this must always pass since the MOSDOp message epoch would have at least as recent an epoch as the scrub epoch, and we would have delayed the message if the pg didn't have a map epoch at least that recent.

dzafman: The assert_interval.epoch is coming from user mode and wouldn't be checked by any other code.  Below I check to make sure it is even a reasonable value and return ERANGE if it is way our of range.

src/osd/ReplicatedPG.cc
```diff
@@ -5977,6 +5979,23 @@ int ReplicatedPG::do_osd_ops(OpContext *ctx, vector<OSDOp>& ops)
...
+	tracepoint(osd, do_osd_op_pre_assert_interval, soid.oid.name.c_str(), soid.snap.val, epoch);
+	dout(5) << "Check interval " << soid << " epoch " << epoch << " pg epoch " << pg_epoch << dendl;
+	if (epoch > pg_epoch) {
```
tchaikov: sorry for the false alarm. this does make sense.

src/librados/IoCtxImpl.cc
```diff
@@ -747,6 +747,65 @@ int librados::IoCtxImpl::operate_read(const object_t& oid,
...
+  o->out_bl.resize(size);
+  o->out_handler.resize(size);
+  o->out_rval.resize(size);
+  o->out_bl[1] = pbl;
+  o->ops[1].op.flags = op_flags;
+  Objecter::Op *objecter_op = objecter->prepare_read_op(oid, oloc,
+	                                      *o, snap_seq, NULL,
+	                                      flags | CEPH_OSD_FLAG_REPAIR_READS | CEPH_OSD_FLAG_IGNORE_OVERLAY,
+	                                      onack, &ver);
+  objecter_op->target.osd = osdid;
```
tchaikov: minor, 
we don't guard the case where osd is not up in `Objecter::op_submit()`. maybe we can check it in `Objecter::_calc_target()`? i have following backtrace if an invalid osdid is passed in:

```
 ceph version Development (no_version)
 1: (ceph::BackTrace::BackTrace(int)+0x2d) [0x4c5e6d]
 2: ./bin/rados() [0x4c5540]
 3: (()+0x10d30) [0x7f0fcae36d30]
 4: (gsignal()+0x38) [0x7f0fbba9b458]
 5: (abort()+0x16a) [0x7f0fbba9c8da]
 6: (ceph::__ceph_assert_fail(char const*, char const*, int, char const*)+0x310) [0x7f0fc1156e8b]
 7: (OSDMap::get_inst(int) const+0x48) [0x7f0fc0a1fd2c]
 8: (Objecter::_get_session(int, Objecter::OSDSession**, ceph::shunique_lock<boost::shared_mutex>&)+0x4fc) [0x7f0fc09f9f1c]
 9: (Objecter::_op_submit(Objecter::Op*, ceph::shunique_lock<boost::shared_mutex>&, unsigned long*)+0x2a0) [0x7f0fc09ff290]
 10: (Objecter::_op_submit_with_budget(Objecter::Op*, ceph::shunique_lock<boost::shared_mutex>&, unsigned long*, int*)+0x266) [0x7f0fc09fe7c4]
 11: (Objecter::op_submit(Objecter::Op*, unsigned long*, int*)+0x82) [0x7f0fc09fe4ec]
 12: (librados::IoCtxImpl::aio_operate_repair_read(object_t const&, ObjectOperation*, librados::AioCompletionImpl*, ceph::buffer::list*, int, int, unsigned int, int)+0x458) [0x7f0f
c09cfddc]
 13: (librados::IoCtxImpl::operate_repair_read(object_t const&, ObjectOperation*, ceph::buffer::list*, int, int, unsigned int, int)+0x6c) [0x7f0fc09cfed8]
 14: (librados::IoCtxImpl::repair_read(object_t const&, ceph::buffer::list&, unsigned long, unsigned long, int, int, unsigned int)+0xf4) [0x7f0fc09d2a44]
 15: (librados::IoCtx::repair_read(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, ceph::buffer::list&, unsigned long, unsigned long, int, i
nt, unsigned int)+0x6e) [0x7f0fc0963738]
 16: ./bin/rados() [0x475100]
 17: ./bin/rados() [0x47f22f]
 18: (main()+0x2294) [0x48c867]
 19: (__libc_start_main()+0xf0) [0x7f0fbba885f0]
 20: (_start()+0x29) [0x4745b9]
```

tchaikov: at https://github.com/ceph/ceph/blob/master/src/osd/OSD.cc#L8651, reply the client with ENXIO. and at that moment the MOSDOp is only partially decoded, so the MOSDOp::ops is still empty. 

if a non-primary osd is specified, the osd will return a MOSDOpReply with empty ops. on the client side, none of the complete call backs will be called at https://github.com/ceph/ceph/blob/master/src/osdc/Objecter.cc#L3270. this makes the client wait in vain.

@dzafman maybe we can support the ec pool by delegating this call to aio_operate_read() ? or just return an error if the client wants to read a shad by specifying a src osd? or we need to give a revamp of current read path of ec backend to support the (osd,shard). but i wonder what the user would want to do with a erasure shard without the erasure decoder? the data shard (m) might still be recognisable, but what if it's a parity shard (k)?

src/osd/ReplicatedPG.cc
athanatos: Makes sense.

src/librados/IoCtxImpl.cc
tchaikov: and might want to report `EINVAL` or `EOPNOTSUPP` if the op is targeting an non-primary OSD, and it's reading from a ec pool?

i have error message like
```
error getting shard from osd 1 of ec_pool/something: (6) No such device or address
```
with log message from osd:
```
2016-06-27 22:32:59.358025 7fc3e9a5a700  7 osd.1 35 we are invalid target
```

dzafman:
The Jenkins build didn't actually run the updated osd-scrub-repair.sh because it has been disabled.

dachary:
jenkins, test this please (osd-scrub-repair.sh is back !)

dachary:
jenkins test this please (unittest_ceph_crypto transient error)

dzafman:
These 3 commits were pull into ceph/ceph#9203

PR #8933, opened 2016-05-04
osd/OSD.h: remove redundant find in map

stiopaa1:
We don't need to use find to get iterator that is then passed to
erase. We can just erase element directly.


liewegas:
pg_epochs is a multiset, and we only want to delete 1 instance of the value, not all of them.

stiopaa1:
You are right, missed that.

PR #8934, opened 2016-05-04
os/bluestore: drop transient extent flags

liewegas:
These were only ever set in memory to control the logic; they
don't belong as part of the (on disk) extent struct.


src/os/bluestore/BlueStore.cc
```diff
@@ -2697,30 +2697,23 @@ int BlueStore::_do_read(
...
+      uint64_t front_extra = x_off % block_size;
+      uint64_t r_off = x_off - front_extra;
+      uint64_t r_len = ROUND_UP_TO(x_len + front_extra, block_size);
+      dout(30) << __func__ << "  reading " << r_off << "~" << r_len << dendl;
+      bufferlist t;
+      r = bdev->read(r_off + bp->second.offset, r_len, &t, &ioc, buffered);
+      if (r < 0) {
+	goto out;
       }
+      r = r_len;
```
xiexingguo: This is not necessary, we'll reset `r = bl.length();` on exit.

xiexingguo:
Looks good except a nit.

PR #8939, opened 2016-05-05
Fix Permission: StupidAllocator.cc

shinobu-x:
It's trivial but...
-rwxrwxr-x. 1 ceph ceph 8.2K Apr 22 18:11 StupidAllocator.cc
-rw-rw-r--. 1 ceph ceph 8.2K Apr 22 18:11 StupidAllocator.cc


liewegas:
you included submodule updates.  make sure you 'git submodule update --init' to sync them up, then make this change.

liewegas:
Please include a Signed-off-by line in your commit message too.  :)  Thanks!

PR #8940, opened 2016-05-05
hammer:ceph-disk: add log level setting 'log-stdout' to command

jordan41177:
http://tracker.ceph.com/issues/15410
with the fix of ceph-deploy: https://github.com/ceph/ceph-deploy/pull/392

This part have fixed in the versions after hammer. But the hammer is
not fixed yet. This may cause the wrong log level showing from using
ceph-disk to create osd.

Signed-off-by: DesmondS <desmond.s@inwinstack.com>

### DIFF
--- a/src/ceph-disk
+++ b/src/ceph-disk
@@ -2751,6 +2751,11 @@ def parse_args():
         help='be more verbose',
         )
     parser.add_argument(
+        '--log-stdout',
+        action='store_true', default=None,
+        help='log to stdout',
+    )
+    parser.add_argument(
         '--prepend-to-path',
         metavar='PATH',
         default='/usr/bin',
@@ -2975,13 +2980,7 @@ def parse_args():
 def main():
     args = parse_args()
 
-    loglevel = logging.WARNING
-    if args.verbose:
-        loglevel = logging.DEBUG
-
-    logging.basicConfig(
-        level=loglevel,
-        )
+    setup_logging(args.verbose, args.log_stdout)
 
     if args.prepend_to_path != '':
         path = os.environ.get('PATH', os.defpath)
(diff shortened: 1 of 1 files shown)


tchaikov:
@jordan41177 can i see your patch as a backport from master?

as it repeats some changes in the master branch, among other things:
* 6cfb4b35ea13f7953d25bf0d2baae75bfe3fb5d0
* 5aeca731aa5bfd0c2071c30e8287f73e34919c21

if that's the case, maybe we can follow the backport process instead?

jordan41177:
@tchaikov ,
This PR is from hammer branch. It is already fixed on master branch(Jewel).
But it seems that 'ceph-disk' are big different from hammer to Jewel.  
I want to fix the wrong log level show in 'ceph-deploy osd create'.
How the backport process work? I am newbie in contributing for Ceph...

Thank you

tchaikov:
@jordan41177 

tl;dr: 

1. mark the tracker issue with "backport: hammer"
1. copy to the issue as a backport issue
1. git-cherry-pick -x <the commit you want to backport>
1. create a PR with the changeset
1. connect the newly created issue with this PR

see http://tracker.ceph.com/projects/ceph-releases/wiki/HOWTO_backport_commits; the idea of this process is to connect of the backported changes with the original changesets, so they can be tracked.

> But it seems that 'ceph-disk' are big different from hammer to Jewel.

agreed. and your changeset is actually involving multiple commits in ceph_disk/main.py targeting master:

- 60c22a9353c0c4245ccda1dbd3416956ac713a3d
- 6cfb4b35ea13f7953d25bf0d2baae75bfe3fb5d0
- 0186062d29c73c2803d21da27457b38490dfd361
- 5aeca731aa5bfd0c2071c30e8287f73e34919c21

so personally, i think we can either 
- do the hard way, say, to backport the involved changes and resolve the conflict manually
- or, explain the rationale why we do not the ordinary backport in the commit message. and manually apply the change to hammer instead using "git cherry-pick" which barely work in this case.

@dachary what do you think?

jordan41177:
@tchaikov, I tried to fix in backport 
Backport issue: http://tracker.ceph.com/issues/15775
New PR : https://github.com/ceph/ceph/pull/8988
I am not sure is there any problem... I am poor in English...

dachary:
Obsoleted by https://github.com/ceph/ceph/pull/8988

PR #8941, opened 2016-05-05
mon/OSDMonitor: more fixes for reweight_by_* command

xiexingguo:


xiexingguo:
@liewegas

src/mon/OSDMonitor.cc
```diff
@@ -638,6 +638,9 @@ int OSDMonitor::reweight_by_utilization(int oload,
       // is a factor to take into account the original weights,
       // to represent e.g. differing storage capacities
       unsigned weight = osdmap.get_weight(p->first);
+      if (weight == 0) {
```
tchaikov: why not move this out of this if block and right before https://github.com/ceph/ceph/pull/8941/commits/3538fff7a58e7668f6876ec398c9ac92edd25b43#diff-0a5db46a44ae9900e226289a810f10e8R634 .

xiexingguo: @tchaikov Upgraded, thank you.

tchaikov:
lgtm

src/mon/OSDMonitor.cc
```diff
@@ -592,6 +592,11 @@ int OSDMonitor::reweight_by_utilization(int oload,
     std::pair<int, float> osd_util;
     osd_util.first = p->first;
     if (by_pg) {
+      if (pgs_by_osd[p->first] == 0) {
```
tchaikov: `pgs_by_osd[p->first]` could read out of the bound if the osd id is greater than the largest osd id serving the specified pools. so maybe we should check in this way:
```c++
if (p->first >= (int)pgs_by_osd.size()) {
```

tchaikov:
withdraw my lgtm, as https://github.com/ceph/ceph/pull/8941#discussion_r62284023 needs to be addressed first.

src/mon/OSDMonitor.cc
xiexingguo: Seems we have set the array size at max_osd at 
https://github.com/ceph/ceph/pull/8941/files#diff-0a5db46a44ae9900e226289a810f10e8R500.

src/mon/OSDMonitor.cc
```diff
@@ -519,6 +519,10 @@ int OSDMonitor::reweight_by_utilization(int oload,
 	if (*q >= (int)pgs_by_osd.size())
 	  pgs_by_osd.resize(*q);
```
tchaikov: @xiexingguo then we can zap this line, or replace it with 
```c++
assert(*q < (int)pgs_by_osd.size());
```

src/mon/OSDMonitor.cc
```diff
@@ -519,6 +519,10 @@ int OSDMonitor::reweight_by_utilization(int oload,
...
 	if (pgs_by_osd[*q] == 0) {
+          if (osdmap.crush->get_item_weightf(*q) <= 0) {
```
tchaikov: is it possible that we will try to increase the weight of a zero-weighted osd?

xiexingguo: @tchaikov  Upgraded https://github.com/ceph/ceph/pull/8941/commits/6effa004dd060f3e1c0a5e057d642a667f057d28 to a safer way.

xiexingguo: And I doubt we shall let this happen, otherwise we will encounter an **division by zero** error at https://github.com/ceph/ceph/pull/8941/files#diff-0a5db46a44ae9900e226289a810f10e8R611

src/mon/OSDMonitor.cc
xiexingguo: oops, wait, I revert this change by accident...

yuriw:
run http://pulpito.ceph.com/yuriw-2016-05-24_13:12:38-rados-wip-yuri-testing2-distro-basic-smithi/
re-run of failed http://pulpito.ceph.com/yuriw-2016-05-24_20:25:47-rados-wip-yuri-testing2-distro-basic-smithi/

PR #8944, opened 2016-05-05
rbd: helpful error message on map failure

vshankar:
Failure to map an rbd image provides pretty much nothing to
debug (or rectify) the nature of failure. This change tries
to improve the same by providing helpful error messages (by
pointing to dmesg, etc..).

Fixes: http://tracker.ceph.com/issues/15721
Signed-off-by: Venky Shankar <vshankar@redhat.com>

### DIFF
--- a/src/tools/rbd/action/Kernel.cc
+++ b/src/tools/rbd/action/Kernel.cc
@@ -153,6 +153,45 @@ static int do_kernel_showmapped(Formatter *f)
   return r;
 }
 
+/*
+ * hint user to check syslog for krbd related messages and provide suggestions
+ * based on errno return by krbd_map(). also note that even if some librbd calls
+ * fail, we atleast dump the "try dmesg..." message to aid debugging.
+ */
+static void print_error_description(const char *poolname, const char *imgname,
+				    const char *snapname, int maperrno)
+{
+  int r;
+  uint8_t oldformat;
+  librados::Rados rados;
+  librados::IoCtx ioctx;
+  librbd::Image image;
+
+  if (maperrno == -ENOENT)
+    goto done;
+
+  r = utils::init_and_open_image(poolname, imgname, snapname,
+				 true, &rados, &ioctx, &image);
+  if (r < 0)
+    goto done;
+
+  r = image.old_format(&oldformat);
+  if (r < 0)
+    goto done;
+
+  /*
+   * kernel returns -ENXIO when mapping a V2 image due to unsupported feature
+   * set - so, hint about that too...
+   */
+  if (!oldformat && (maperrno == -ENXIO)) {
+    std::cout << "RBD image feature set mismatch. You can disable features unsupported by the "
+	      << "kernel with \"rbd feature disable\"." << std::endl;
+  }
+
+ done:
+  std::cout << "In some cases useful info is found in syslog - try \"dmesg | tail\" or so." << std::endl;
+}
+
 static int do_kernel_map(const char *poolname, const char *imgname,
                          const char *snapname)
 {
@@ -181,8 +220,10 @@ static int do_kernel_map(const char *poolname, const char *imgname,
   }
 
   r = krbd_map(krbd, poolname, imgname, snapname, oss.str().c_str(), &devnode);
-  if (r < 0)
+  if (r < 0) {
+    print_error_description(poolname, imgname, snapname, r);
     goto out;
+  }
 
   std::cout << devnode << std::endl;
 


src/civetweb
```diff
@@ -1 +1 @@
-Subproject commit e12494070432b85d5ffce7dece81c0baad457a5f
+Subproject commit 8d271315a541218caada366f84a2690fdbd474a2
```
dillaman: Yank this from your commit

src/tools/rbd/action/Kernel.cc
```diff
@@ -153,6 +153,39 @@ static int do_kernel_showmapped(Formatter *f)
   return r;
 }
 
+static void rbd_check_unsup_features(const char *poolname, const char *imgname,
```
dillaman: Minor: how about something like ```print_error_description``` or similar.

src/tools/rbd/action/Kernel.cc
```diff
@@ -153,6 +153,39 @@ static int do_kernel_showmapped(Formatter *f)
...
+  uint8_t oldformat;
+  librados::Rados rados;
+  librados::IoCtx ioctx;
+  librbd::Image image;
+
+  /*
+   * kernel returns ENXIO when mapping a V2 image due to unsupported feature
+   * set or due to an invalid header for a V1 image format.
+   */
+  if (maperrno != ENXIO)
```
dillaman: It might be nice to ensure that we have helpful coverage for the most common possible failure conditions that would require checking dmesg:

1) CRUSH map incompatibilities (http://docs.ceph.com/docs/master/rados/operations/crush-map/#which-client-versions-support-crush-tunables)
2) Authentication issues
3) RBD feature incompatibilities
4) RBD "fancy" striping (RBD_FEATURE_STRIPING_V2 w/ a stripe unit != 1 and stripe width != object size)

src/civetweb
vshankar: ah! crap.... :(

src/tools/rbd/action/Kernel.cc
vshankar: Hmmm.. In that case we need to handle EINVAL too. It might be a good option to handle all error codes except ENOENT and point the user to dmesg.

src/tools/rbd/action/Kernel.cc
```diff
@@ -188,6 +221,10 @@ static int do_kernel_map(const char *poolname, const char *imgname,
 
   free(devnode);
 out:
+  if (r < 0) {
+    rbd_check_unsup_features(poolname, imgname, snapname, errno);
```
jdurgin: errno could have changed since the write(2) call - just use r instead, since it's already been copied there in safe_write()

src/tools/rbd/action/Kernel.cc
vshankar: Updated patch logs for all cases except ENOENT. Although map request for a  missing image is logged by the krbd, the errno pretty self explanatory and there's no point in trying to open an image which does not exist.

src/tools/rbd/action/Kernel.cc
vshankar: Thanks.

src/tools/rbd/action/Kernel.cc
```diff
@@ -188,6 +221,10 @@ static int do_kernel_map(const char *poolname, const char *imgname,
 
   free(devnode);
 out:
+  if (r < 0) {
+    print_error_description(poolname, imgname, snapname, -r);
```
dillaman: We generally keep all the error codes negative, so just for consistency sake I would recommend keeping all error code comparisons to the negative version.

dillaman:
@idryomov Any comments?

src/tools/rbd/action/Kernel.cc
vshankar: Ah OK. Not a problem. Will resend...

idryomov: Why call it here with another `if (r < 0)` instead of before `goto out;`?

src/tools/rbd/action/Kernel.cc
```diff
@@ -153,6 +153,39 @@ static int do_kernel_showmapped(Formatter *f)
...
+				    const char *snapname, int maperrno)
+{
+  int r;
+  uint8_t oldformat;
+  librados::Rados rados;
+  librados::IoCtx ioctx;
+  librbd::Image image;
+
+  if (maperrno == ENOENT)
+    return;
```
idryomov: I think we need to print the "see dmesg" message on all errors.  On -ENOENT in particular, since the kernel will output something like "pool rbd does not exist" or "image rbd/b does not exist".

src/tools/rbd/action/Kernel.cc
```diff
@@ -153,6 +153,39 @@ static int do_kernel_showmapped(Formatter *f)
...
+  r = utils::init_and_open_image(poolname, imgname, snapname,
+				 true, &rados, &ioctx, &image);
+  if (r < 0)
+    return;
+
+  r = image.old_format(&oldformat);
+  if (r < 0)
+    return;
+
+  std::cout << "check dmesg(1) for rbd related kernel messages.";
```
idryomov: Make it "In some cases useful info is found in syslog - try dmesg | tail or so." and print it last, after all other suggestions, so it looks like:

rbd: sysfs write failed
rbd: map failed: ...
suggestion 1
suggestion 2
there may be something in the dmesg

There may be more than one suggestion we'd like to make in the future, and only a subset of errors is annotated by krbd.

src/tools/rbd/action/Kernel.cc
vshankar: I excluded -ENOENT as its pretty self explanatory. No worries, I'll have that too.

src/tools/rbd/action/Kernel.cc
vshankar: One reason for that is if this routine expands later with possibly other krbd calls, calling `print_error_description()` at a common exit point is neat.

src/tools/rbd/action/Kernel.cc
vshankar: That makes sense. Thanks for reviewing.

src/tools/rbd/action/Kernel.cc
idryomov: It won't (or very unlikely to).  The only call we want to try to interpret errors and/or direct users to syslog for is the map call - everything else is going to be different.

src/tools/rbd/action/Kernel.cc
```diff
@@ -153,6 +153,44 @@ static int do_kernel_showmapped(Formatter *f)
...
+  r = image.old_format(&oldformat);
+  if (r < 0)
+    goto done;
+
+  /*
+   * kernel returns -ENXIO when mapping a V2 image due to unsupported feature
+   * set - so, hint about that too...
+   */
+  if (!oldformat && (maperrno == -ENXIO)) {
+    std::cout << "Possible failure due to mismatching feature set." << std::endl;
```
idryomov: The rectify part is missing.  Maybe something like
`RBD image feature set mismatch?  You can disable features unsupported by`
`the kernel with "rbd feature disable".`

Otherwise looks good.

PR #8945, opened 2016-05-05
radosgw-admin: added missing json output to some commands

cbodley:
The radosgw-admin commands 'zone modify', 'zonegroup modify', and 'zonegroup add' do not print any output on success, so it's hard to tell whether or not they worked.


yehudasa:
@cbodley lgtm. Do we want it in jewel?

cbodley:
added http://tracker.ceph.com/issues/15747 with backport=jewel

theanalyst:
+1, looks harmless enough for jewel :)

PR #8947, opened 2016-05-05
[DNM] rgw: add support for bucket quota of Swift API

rzarzynski:


rzarzynski:
Enhanced version is available in PR #9037.

PR #8949, opened 2016-05-05
tests/rbd: add option to leave behind fsx test data

fullerdj:
Previously, fsx deleted all test data upon successful completion.
Automated tests nuke the cluster anyway and new tests could further
examine the data left behind. Leave test data alone by default and
add an option to fsx to remove it as before.

### DIFF
--- a/src/test/librbd/fsx.cc
+++ b/src/test/librbd/fsx.cc
@@ -140,6 +140,7 @@ unsigned long	debugstart = 0;		/* -D flag */
 int	flush_enabled = 0;		/* -f flag */
 int	holebdy = 1;			/* -h flag */
 bool    journal_replay = false;         /* -j flah */
+int	keep_on_success = 0;		/* -k flag */
 int	do_fsync = 0;			/* -y flag */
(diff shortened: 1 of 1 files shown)


dillaman:
Could be a nice feature to optionally keep the images.  The way I read the code, it only keeps images upon success when specified (👍)  instead of what the commit message states.

fullerdj:
It does default to keep, I just should have renamed the variable when I changed the default behavior. We definitely want the option to keep the images, which was invaluable for debugging krbd issues. I thought it made sense to keep them by default since the typical use case is teuthology, where we nuke on failure anyway. This would allow us to run an object map check at the end of the test as well.

If you want to default to discarding, I'll fix that. If you want to default to keep, I'll change the flag name.

fullerdj:
It does default to keep, I just should have renamed the variable when I changed the default behavior. I thought it made sense since the primary case was teuthology, which nukes the nodes anyway. That way, we can do things like check the object map or do some analysis after the fact. If you'd rather default to discard, I'll change it back. If you want to default to keep, I'll change the flag name so it actually makes sense.

dillaman:
Minor: can you change your commit title?

PR #8951, opened 2016-05-05
rgw: handle stripe transition when flushing final pending_data_bl

yehudasa:
Fixes: http://tracker.ceph.com/issues/15745

When complete_writing_data() is called, if pending_data_bl is not empty
we still need to handle stripe transition correctly. If pending_data_bl
has more data that we can allow in current stripe, move to the next one.

### DIFF
--- a/src/rgw/rgw_rados.cc
+++ b/src/rgw/rgw_rados.cc
@@ -2409,17 +2409,32 @@ int RGWPutObjProcessor_Atomic::complete_writing_data()
     }
     obj_len = (uint64_t)first_chunk.length();
   }
-  if (pending_data_bl.length()) {
+  while (pending_data_bl.length()) {
     void *handle;
-    int r = write_data(pending_data_bl, data_ofs, &handle, false);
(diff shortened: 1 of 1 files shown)


cbodley:
I'm looking into getting the s3readwrite tests to reproduce this, and pushed https://github.com/ceph/ceph-qa-suite/commit/041ca167355ca76c6b16d42e8a2f76ec4cbc536d to raise the file size from 2000 to 8000. Would you expect that to trigger the bug?

I scheduled the wip-rgw-testing-cbodley branch against that with `--filter rgw_readwrite.yaml`: http://pulpito.ceph.com/cbodley-2016-05-09_16:29:18-rgw-wip-rgw-testing-cbodley---basic-smithi/

yehudasa:
@cbodley eventually it should be caught

cfanz:
Hi, I'm still can't find out why pending_data_bl.length() will greater than available capability in current stripe. Function "put_data_and_throttle" guaranteed that when it is returned,  the length of data in pending_data_bl will be less than MIN(max_chunk_size, next_part_ofs - data_ofs), that means complete_writing_data can successfully write all remaining data into current stripe, no need to start a new stripe at all.

PR #8954, opened 2016-05-06
cmake: Remove redundant call to get_processors

badone:
The redundant code causes the following undesirable behaviour.

$ DRY_RUN=echo ./run-cmake-check.sh 2>/dev/null|grep "\-j"
make -j4 -j4


tchaikov:
sorry forgot to add the Reviewed-by label!

PR #8958, opened 2016-05-06
librbd: constrain size of AioWriteEvent journal entries

dillaman:


jdurgin:
lgtm - the fixed size is a bit brittle, but don't see a better way to do it atm, and don't expect it to change too often

dillaman:
@jdurgin Agreed -- very ugly but I made sure the ```asserts``` will fail during unit tests should it change.

PR #8961, opened 2016-05-06
osd: don't get inc osdmap name truncated

xiexingguo:


src/osd/OSD.cc
```diff
@@ -6642,7 +6642,7 @@ void OSD::handle_osd_map(MOSDMap *m)
       OSDMap *o = new OSDMap;
       if (e > 1) {
 	bufferlist obl;
-	get_map_bl(e - 1, obl);
+	assert(get_map_bl(e - 1, obl));
```
liewegas: please assert on a different line.. we don't want any side-effects in the assert itself

src/osd/OSD.h
```diff
@@ -1146,7 +1146,7 @@ class OSD : public Dispatcher,
     return ghobject_t(hobject_t(sobject_t(object_t(foo), 0)));
   }
   static ghobject_t get_inc_osdmap_pobject_name(epoch_t epoch) {
-    char foo[20];
+    char foo[32];
```
tchaikov: @xiexingguo so it should be 22, right?

xiexingguo: @tchaikov You are definitely right. Thanks.

tchaikov:
lgtm

yuriw:

PR #8965, opened 2016-05-06
mds: finish lock waiters in the same order that they were added.

ukernel:
Current code first processes lock waiters who have smaller wait mask.
Lock waiters who have large wait mask can starve if client keeps
sending requests that add waiter with small mask.

### DIFF
--- a/src/mds/mdstypes.h
+++ b/src/mds/mdstypes.h
@@ -1336,6 +1336,7 @@ class MDSCacheObject {
 
 
   // -- wait --
+  const static uint64_t WAIT_ORDERED	 = (1ull<<61);
   const static uint64_t WAIT_SINGLEAUTH  = (1ull<<60);
   const static uint64_t WAIT_UNFREEZE    = (1ull<<59); // pka AUTHPINNABLE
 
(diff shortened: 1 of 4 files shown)


ukernel:
this fixes 'getattr starve setattr' issue described in
https://github.com/ceph/ceph/pull/8925

ukernel:

src/mds/mdstypes.h
```diff
@@ -1545,6 +1545,7 @@ class MDSCacheObject {
   // waiting
  protected:
   compact_multimap<uint64_t, MDSInternalContextBase*>  waiting;
+  static const uint64_t WAIT_LOCK_MASK = ~0xffull;
```
jcsp: This constant needs a comment explaining where it comes from.

Maybe it should also be defined somewhere in the same place as the flags it's detecting to make it clearer?

src/mds/MDSContext.h
```diff
@@ -41,8 +41,20 @@ class MDSContext : public Context
  */
 class MDSInternalContextBase : public MDSContext
 {
+  static uint64_t last_seq;
+  uint64_t seq;
```
jcsp: It feels strange to have this in all MDSInternalContext instances, when only some of them will ever need it.

Given that the sorting happens inside MDSCacheObject, maybe the sequence numbers etc should live there too.  What if MDSCacheObject::waiting stored a std::pair of sequence number and MDSCache object for example, and set the sequence number during add_waiter?

jcsp:
Passing here: https://github.com/ceph/ceph/pull/8965

jcsp:
Created ticket to backport http://tracker.ceph.com/issues/16135

PR #8967, opened 2016-05-06
osdc/Objecter: upper bound watch_check result

liewegas:
This way we always return a safe upper bound on the amount of time
since we did a check.  Among other things, this prevents us from
returning a value of 0, which is confusing.

Fixes: http://tracker.ceph.com/issues/15760


jdurgin:
lgtm

smithfarm:
@liewegas In hammer we have `return age.to_msec();` - so replace with `return 1 + aget.to_msec();` ?

liewegas:
yeah.  thanks!

PR #8972, opened 2016-05-06
jewel: python: clone operation will fail if config overridden with "rbd default format = 1"

Abhishekvrshny:
http://tracker.ceph.com/issues/15697


dachary:
test this please

PR #8973, opened 2016-05-06
librbd: potential concurrent event processing during journal replay

dillaman:


src/librbd/journal/Replay.cc
```diff
@@ -166,7 +166,7 @@ void Replay<I>::shut_down(bool cancel_ops, Context *on_finish) {
 
     // safely commit any remaining AIO modify operations
     if ((m_in_flight_aio_flush + m_in_flight_aio_modify) != 0) {
-      flush_comp = create_aio_flush_completion(nullptr, nullptr);;
+      flush_comp = create_aio_flush_completion(nullptr);;
```
trociny: while here, also remove the extra semicolon?

jdurgin:
lgtm

PR #8976, opened 2016-05-06
[DNM] rbd: Create fewer empty objects

fullerdj:
In many cases, we unconditionally create empty objects (of size 0) when
performing layered writes. While functionally correct, this causes tools
like rbd du to space is allocated (since they conservatively assume all
objects are full). Detect empty writes and bail out early. This is
particularly helpful for operations like flatten, which create all
objects that might exist in an image.

Fixes: http://tracker.ceph.com/issues/15028

### DIFF
--- a/src/librbd/AioObjectRequest.cc
+++ b/src/librbd/AioObjectRequest.cc
@@ -363,6 +363,16 @@ namespace librbd {
       }
       break;
 
+    case LIBRBD_AIO_WRITE_OBJECT_MAP:
+      ldout(m_ictx->cct, 20) << "OBJECT_MAP" << dendl;
+      finished = false;
+      if (r < 0) {
+	m_state = LIBRBD_AIO_WRITE_ERROR;
+	complete(r);
+      }
+      send_write_op();
+      break;
+
     case LIBRBD_AIO_WRITE_FLAT:
       ldout(m_ictx->cct, 20) << "WRITE_FLAT" << dendl;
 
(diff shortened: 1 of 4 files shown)


src/librbd/CopyupRequest.cc
```diff
@@ -180,6 +180,26 @@ class UpdateObjectMap : public C_AsyncObjectThrottle<> {
...
   }
 
+  bool CopyupRequest::is_nop()
+  {
+    if (!m_copyup_data.is_zero()) {
+      return false;
+    }
+
+    for (const AioObjectRequest *req : m_pending_requests) {
+      const AioObjectWrite *wreq = dynamic_cast<const AioObjectWrite *>(req);
```
dillaman: I'd avoid the dynamic cast and add a new ```AioObjectRequest::is_no_op``` virtual method (combines the write check and the ::all_zero check).

src/librbd/AioObjectRequest.cc
```diff
@@ -410,27 +451,10 @@ namespace librbd {
           		       << m_object_off << "~" << m_object_len << dendl;
         m_state = LIBRBD_AIO_WRITE_PRE;
 
-        uint8_t new_state;
```
dillaman: I am most likely missing something, but what is the point of the new state?  This ```::send_pre``` doesn't appear to do any async (librados) action anymore

fullerdj: We need the new state to decide what to do after updating the object map. WRITE_PRE is now unnecessary, though, so I'll remove it.

dillaman: That's what I am still missing I guess -- the PRE state was just around for updating the object map.  The only difference that I see is that instead of calling ```send_write``` it calls ```send_write_op```?

fullerdj: I guess now WRITE_OBJECT_MAP is analogous to WRITE_PRE, though the state diagram isn't exactly the same. We now skip the object map update in the copyup case (because we don't know yet if it's needed) and handle it as part of the copyup operation, if necessary.

fullerdj:
This approach is insufficient because it does not deal with layered writes that may cause object removal.

PR #8980, opened 2016-05-06
osd/OSD.h: remove unneeded line

stiopaa1:
We don't need to create a key with value 0 here.
It is created in statement:
pgid_tracker[pgid]++
if it doesnt exist.

### DIFF
--- a/src/osd/OSD.h
+++ b/src/osd/OSD.h
@@ -1029,7 +1029,6 @@ public:
   void add_pgid(spg_t pgid, PG *pg) {
     Mutex::Locker l(pgid_lock);
     if (!pgid_tracker.count(pgid)) {
-      pgid_tracker[pgid] = 0;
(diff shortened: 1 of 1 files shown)


xiexingguo:
lgtm

tchaikov:
lgtm also

liewegas:
map<spg_t, int> pgid_tracker;

are you sure that the value is reliably initialized to 0, given that int has no ctor?

stiopaa1:
@liewegas I'm pretty sure.

http://stackoverflow.com/questions/4523959/stdmap-default-value-for-build-in-type
http://stackoverflow.com/questions/2667355/mapint-int-default-values

tchaikov:
> If an insertion is performed, the mapped value is value-initialized (default-constructed for class types, zero-initialized otherwise) and a reference to it is returned.

PR #8986, opened 2016-05-08
osd/OSD.h: move shared_ptr instead of copying it

stiopaa1:
By copying map to next_osdmap we are increasing the shared_ptr's
reference count, which is then decreased again when map variable goes
out of scope. This can be avoided by using std move.

### DIFF
--- a/src/osd/OSD.h
+++ b/src/osd/OSD.h
@@ -505,7 +505,7 @@ public:
   Cond pre_publish_cond;
   void pre_publish_map(OSDMapRef map) {
     Mutex::Locker l(pre_publish_lock);
-    next_osdmap = map;
+    next_osdmap = std::move(map);
   }
 
   void activate_map();


adamemerson:
A sensible change.

Linguistic Generalizations Trouble Mortals.

liewegas:
This works because the move assignment operator clobbers map's state such that map's dtor doesn't deref?

stiopaa1:
@liewegas  
When we move from 'map' variable it loses the ownership of the object that it was pointing to and becomes empty, Since it's empty it's destructor won't do anything. So it I understood correctly it's pretty much what you said.

liewegas:
mixed answers from 
http://stackoverflow.com/questions/507971/do-stl-maps-initialize-primitive-types-on-insert

liewegas:
yeah, looks safe.

adamemerson:
Move constructing a shared pointer generally leaves the moved-from pointer nulled (in fact this is exactly what it does in libstdc++ and libc++ and likely everything else but I think the standard gives wiggleroom in the 'valid but unspecified' business)

The advantage of doing so is that, being a move, the ref count doesn't change so one need not invoke any atomic instructions.

PR #8990, opened 2016-05-09
osd: remove all stale osdmaps in handle_osd_map()

tchaikov:
in a large cluster, there are better chances that the OSD fails to trim
the cached osdmap in a timely manner. and sometimes, it is just unable
to keep up with the incoming osdmap if skip_maps, so the osdmap cache
can keep building up to over 250GB in size. in this change

  removed.
  if skip_maps is true. in my test, it happens when the osd only
  receives the osdmap from monitor occasionally because the osd happens
  to be chosen when monitor wants to share a new osdmap with a random
  osd.

### DIFF
--- a/src/osd/OSD.cc
+++ b/src/osd/OSD.cc
@@ -6530,6 +6530,44 @@ void OSD::osdmap_subscribe(version_t epoch, bool force_request)
   }
 }
 
+void OSD::trim_maps(epoch_t oldest, int nreceived, bool skip_maps)
+{
+  epoch_t min = std::min(oldest, service.map_cache.cached_key_lower_bound());
+  if (min <= superblock.oldest_map)
+    return;
+
+  int num = 0;
+  ObjectStore::Transaction t;
+  for (epoch_t e = superblock.oldest_map; e < min; ++e) {
+    dout(20) << " removing old osdmap epoch " << e << dendl;
+    t.remove(coll_t::meta(), get_osdmap_pobject_name(e));
+    t.remove(coll_t::meta(), get_inc_osdmap_pobject_name(e));
+    superblock.oldest_map = e + 1;
+    num++;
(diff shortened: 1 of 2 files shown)


tchaikov:
a pr resembling this one was posted before as #8017.

in this changeset:

- only zap all old maps if we have skipped maps

xiexingguo:
This PR looks gorgeous to me!
By checking our local Ceph cluster, I found that the oldest osdmap ever persisted on disk did not 
keep pace with sb.oldest, as follows:
```
[root@ceph245 ceph]# ceph --admin-daemon /run/ceph/ceph-osd.4.asok status                
{
    "cluster_fsid": "91b7df9f-b868-ea35-5381-67dd25efc345",
    "osd_fsid": "ea1567de-65f9-44b4-a2ed-b95c9a82f403",
    "whoami": 4,
    "state": "active",
    "oldest_map": 2109,
    "newest_map": 7265,
    "num_pgs": 118
}

[root@ceph245 ceph]# ceph osd getmap 7000 -o osdmap                                      
got osdmap epoch 7000
[root@ceph245 ceph]# ceph osd getmap 2109 -o osdmap    
Error ENOENT: there is no map for epoch 2109
[root@ceph245 ceph]# ceph osd getmap 5000 -o osdmap    
Error ENOENT: there is no map for epoch 5000
``` 
Will it also be addressed by this patch?

tchaikov:
@xiexingguo no, this patch tries to *remove* more osdmaps. instead, your osd has missing osdmaps in its local meta collection. maybe you can file a ticket in tracker?

xiexingguo:
Done.
http://tracker.ceph.com/issues/15777

liewegas:
lgtm

liewegas:
http://pulpito.ceph.com/yuriw-2016-05-10_10:56:12-rados-wip-yuri-testing-distro-basic-smithi/

PR #8997, opened 2016-05-09
rgw: "status" is 0 when http status code is between 200 and 299

sunfch:
Signed-off-by: sunspot sunspot0105@gmail.com


sunfch:
the bug has been fixed

PR #8998, opened 2016-05-09
librados: fix wrong raw pointer conversion

dragonylffly:
ObjectOperationImpl has changed, no longer equal to ::ObjectOperation*,
this patch fixes the wrong pointer conversion.


jdurgin:
Good catch! Looks good to me.

PR #8999, opened 2016-05-09
journal: incorrectly computed object offset within set

dillaman:


jdurgin:
lgtm

PR #9000, opened 2016-05-09
Update hardware-recommendations.rst - SSD section

hfunball:
Updated SSD section with more recent information on SSD costs and uses in ceph. Highlights:
- 114: Changed cost estimate from "10x as much per gigabyte" to "from 2x to more than 10x as much per gigabyte"
     - Supported by looking at SSDs and SAS HDD on Amazon as of 5/9/16. Costs ranged from 2x to 10x for 1GB HDD vs SSD drives depending on performance.
- 116/117: added reasons to consider SSDs in vs initial cost.
- 125/126: Included workload example for SSD consideration
- 143: Changed commend as this is more true around sequential writes vs random
- 154: Changed to note vs SSD drive architecture which is a larger consideration than HDD.
- 156: Changed earlier language stating SSDs are not viable for object storage.

### DIFF
--- a/doc/start/hardware-recommendations.rst
+++ b/doc/start/hardware-recommendations.rst
@@ -12,7 +12,6 @@ running Ceph daemons of a specific type on a host configured for that type
 of daemon. We recommend using other hosts for processes that utilize your 
 data cluster (e.g., OpenStack, CloudStack, etc).
 
-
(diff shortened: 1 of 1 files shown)


hfunball:
resubmitting signed

PR #9002, opened 2016-05-09
rgw: fix updating account/container metadata of Swift API.

rzarzynski:
This patch rectifies an issue with handling of user metadata that are
actually stored by an updated resource (account or container).
The expected behaviour is they will be merged with new medadata coming
from an HTTP client.

Backport: Jewel
Fixes: http://tracker.ceph.com/issues/15779


oritwas:
lgtm

oritwas:
lgtm

PR #9007, opened 2016-05-09
jewel: librbd: unlock image if journal error encountered during lock

dillaman:
http://tracker.ceph.com/issues/15712


dachary:
test this please

PR #9009, opened 2016-05-09
jewel: librbd: possible race condition leads to use-after-free

dillaman:
http://tracker.ceph.com/issues/15713


dachary:
test this please

PR #9023, opened 2016-05-10
librbd: journal IO error results in failed assertion in AioCompletion

dillaman:


src/librbd/AioCompletion.cc
```diff
@@ -207,11 +224,12 @@ namespace librbd {
 
   void C_AioRead::finish(int r)
   {
-    ldout(m_cct, 10) << "C_AioRead::finish() " << this << " r = " << r << dendl;
+    CephContext *cct = m_completion->ictx->cct;
```
trociny: Not sure it makes difference here, but taking that in all AioCompletion methods we are always  accessing CephContext pointer only after acquiring the lock, may be it make sens to acquire `m_completion->lock`  (which is done conditionally a little below) before this?

dillaman:
retest this please

smithfarm:
@dillaman Would it make sense to add http://tracker.ceph.com/issues/16077 to the PR description given that the issue is flagged for backport?

dillaman:
@smithfarm Since they were semi-intertwined, I'll just yank the backport tag off of issue 16077 since it will get backported with 15791

src/librbd/Journal.cc
```diff
@@ -1349,6 +1350,10 @@ void Journal<I>::handle_replay_process_safe(ReplayEntry replay_entry, int r) {
   CephContext *cct = m_image_ctx.cct;
 
   m_lock.Lock();
+  assert(m_state == STATE_REPLAYING ||
+         m_state == STATE_FLUSHING_RESTART ||
+         m_state == STATE_FLUSHING_REPLAY);
```
trociny: @dillaman Observing this assert violation in TestJournalReplay.Rename:
http://qa-proxy.ceph.com/teuthology/trociny-2016-06-07_23:26:11-rbd-wip-mgolub-testing---basic-vps/245308/teuthology.log

trociny:
@dillaman There was also a failure in rbd_mirror.sh:
TEST: stop/start/restart mirror via admin socket
After `ceph --admin-daemon cluster1-client.mirror.asok rbd mirror start` it timed out on wait_for_image_replay_started.

http://qa-proxy.ceph.com/teuthology/trociny-2016-06-07_23:26:11-rbd-wip-mgolub-testing---basic-vps/245279/teuthology.log

It may be not related. FYI, my testing branch also includes #9440

src/librbd/Journal.cc
dillaman: Looks like another one of those failures due to a fault injection in the messaging layer.  I don't think it's related to this change (I've been running it in a loop w/ fault injection and it hasn't repeated).

dillaman:
@trociny I can't think of how the rbd-mirror thing is related, so I am hoping it was just a slow vps node.  I've rescheduled that test on mira: http://pulpito.ceph.com/jdillaman-2016-06-08_08:59:22-rbd-wip-mgolub-testing---basic-mira/

trociny:
@dillaman Ah, I recently updated wip-mgolub-testing branch to include only #9440 (just wanted to recheck this first before merging it), so I think  http://pulpito.ceph.com/jdillaman-2016-06-08_08:59:22-rbd-wip-mgolub-testing---basic-mira/ will not test this patch set.

dillaman:
@trociny ACK -- thanks, I'll cancel the job

src/librbd/Journal.cc
dillaman: I looked at the core and it failed due to an existing race condition on the transition between REPLAYING -> READY.  I'll open a tracker ticket.

MoonJustry:
@dillaman Hi, I have a quesion that has this PR fixed the bug #15034? if so, why the rbd_op_threads option still disabled?

dillaman:
@MoonJustry It's not fixed -- there are race conditions that cause various failures. See tracker ticket 17379

PR #9025, opened 2016-05-10
rbd-mirror: avoid potential deadlock

trociny:


dillaman:
@trociny Did you hit a deadlock during testing?  If so, I want to make sure it makes it into the 10.2.1 release.

trociny:
@dillaman, no, I did not hit a deadlock. I just noticed this potential issue when looking at the code recently and decided it will be safer to run update_progress() not holding the lock.

PR #9027, opened 2016-05-10
make-debs.sh: avoid creating recursive conf dir

roidayan:
Before this commit when running make-debs.sh second time ln will create
the conf symlink inside already exisiting conf dir that was created
on the first tune.

### DIFF
--- a/make-debs.sh
+++ b/make-debs.sh
@@ -93,7 +93,9 @@ Suite: stable
 Components: main
 Architectures: i386 amd64 source
 EOF
-ln -s $codename/conf conf
+if [ ! -e conf ]; then
+    ln -s $codename/conf conf
+fi
 reprepro --basedir $(pwd) include $codename WORKDIR/*.changes
 #
 # teuthology needs the version in the version file


make-debs.sh
```diff
@@ -93,7 +93,9 @@ Suite: stable
 Components: main
 Architectures: i386 amd64 source
 EOF
-ln -s $codename/conf conf
+if [ ! -e conf ]; then
+    ln -s $codename/conf conf
```
tchaikov: could also "ln -sf $codename/conf conf"

roidayan: It's not the same. -f is for force. the problem is the target "conf" exists and it's a directory so ln creates a symlink inside conf and then I get conf/conf.

it's more like "ln -sT $codename/conf conf" but then ln will return an error as conf exists and the bash will fail as it runs with -e

tchaikov: thanks @roidayan , got it now!

tchaikov:
lgtm

PR #9035, opened 2016-05-10
cmake: pass CEPH_BUILD_VIRTUALENV to test as an env variable

tchaikov:
Signed-off-by: Kefu Chai <tchaikov@gmail.com>


tchaikov:
@cbodley do you mind taking a look? i have `TMPDIR` set in my env, so cmake build the ceph-disk venv in it, but the ceph-helper.sh still defaults to `/tmp`.

PR #9037, opened 2016-05-10
rgw: add support for account and container quotas of Swift API

rzarzynski:
This pull request brings support for account as well as container quota of Swift API. The first one isn't currently operational because RadosGW doesn't have the *reseller admin* concept yet. It's essential because only administrators should be able to modify account quotas - there is no business in having that kind of quota if we allow account owner to change it. This is stays in contrast to bucket quota that may be altered by anyone with write access to a given bucket.

Derogates: #8947.

Depends on: *reseller admin* feature of #8657.

Includes commits from: #9002.

CC: @yehudasa, @mattbenjamin.

### DIFF
--- a/src/rgw/rgw_op.cc
+++ b/src/rgw/rgw_op.cc
@@ -2797,34 +2797,6 @@ done:
   dispose_processor(processor);
 }
 
-int RGWPutMetadataAccount::handle_temp_url_update(
-  const map<int, string>& temp_url_keys) {
-  RGWUserAdminOpState user_op;
-  user_op.set_user_id(s->user->user_id);
-
-  map<int, string>::const_iterator iter;
-  for (iter = temp_url_keys.begin(); iter != temp_url_keys.end(); ++iter) {
-    user_op.set_temp_url_key(iter->second, iter->first);
-  }
-
-  RGWUser user;
-  op_ret = user.init(store, user_op);
-  if (op_ret < 0) {
-    ldout(store->ctx(), 0) << "ERROR: could not init user ret=" << op_ret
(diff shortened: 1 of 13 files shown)


src/rgw/rgw_op.cc
```diff
@@ -1876,6 +1878,64 @@ static void populate_with_generic_attrs(const req_state * const s,
...
+      return -EINVAL;
+    }
+    add_attrs.erase(iter);
+    extracted = true;
+  }
+
+  /* Put new limit on bucket (container) size. */
+  iter = add_attrs.find(RGW_ATTR_QUOTA_MSIZE);
+  if (iter != add_attrs.end()) {
+    const int64_t max_size =
```
oritwas: why do you need the temp variable?

rzarzynski: Right, it's completely unnecessary. Thanks for pointing that.
I will push the amended version soon.

rzarzynski:
@oritwas: I've pushed a slightly amended version. The temp variable has been removed, debugs improved (a new one with priority `20` in the end of `RGWQuotaHandlerImpl::check_quota` to signalize positive a quota verification) and rounding in `RGWQuotaInfoDefApplier::is_size_exceeded` fixed

rzarzynski:
@oritwas: rebased due to a merge conflict.

oritwas:
@rzarzynski , can you rebase?

rzarzynski:
@oritwas: rebased. The previous version [is available](https://github.com/rzarzynski/ceph/tree/wip-rgw-swift-quota-old-2) as well.

PR #9045, opened 2016-05-10
rpm: Drop SELinux priority setting

branto1:
Older versions of semodule binary that are in Centos/RHEL do not
support priority setting, dropping it.

Fixes: #15822


smithfarm:
lgtm, merging

smithfarm:
@branto1 The old-style "Fixes: #..." is problematic. See 65770059

PR #9051, opened 2016-05-10
Wip 15815 jewel

athanatos:


athanatos:
Checked locally, it seems to work with these backports.  I think it's good to merge.

PR #9056, opened 2016-05-10
osd/OSD.cc: make pistate a local struct

stiopaa1:
It is only used within build_past_intervals_parallel, so it can be
moved inside it.

### DIFF
--- a/src/osd/OSD.cc
+++ b/src/osd/OSD.cc
@@ -3344,16 +3344,15 @@ void OSD::load_pgs()
  * follow the same logic, but do all pgs at the same time so that we
  * can make a single pass across the osdmap history.
  */
-struct pistate {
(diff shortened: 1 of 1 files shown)


tchaikov:
the failure looks like http://tracker.ceph.com/issues/13937

athanatos:
lgtm pending rados run

tchaikov:
tested at 

- http://pulpito.ceph.com/kchai-2016-06-15_04:31:07-rados-wip-kefu-testing---basic-mira/
- http://pulpito.ceph.com/kchai-2016-06-15_10:14:46-rados-wip-kefu-testing---basic-mira/
- http://pulpito.ceph.com/kchai-2016-06-17_02:37:57-rados-wip-kefu-testing---basic-mira/

PR #9064, opened 2016-05-11
os/Objectstore: remove legacy tbl support

majianpeng:


majianpeng:
@liewegas . hope your comments. Thanks!

liewegas:
Note that we can actually remove the legacy tbl support completely now.  Let's do that instead of the tbl fix in the first patch?

majianpeng:
Are you mean no place use tbl? If sure, i can do. In fact, whether or not use_tbl, it use the same parameter: largest_data_off_in_tbl. So i deem add a new patch to remove tbl. Other reconstruct. How about it?

liewegas:
Yeah, I would just drop the tbl fix from this series, and do a separate PR 
that removes the tbl support completely.

Even in hammer it's only ever used during upgrade.  Jewel only needs it in 
case there is an OSD journal that wasn't flushed.

majianpeng:
@liewegas . update. But for decode() part, i'm not sure whether your suggest. Please review. Thanks!

athanatos:
Something seems to be messed up, it claims there are 12 commits. @majianpeng I think a rebase went wrong.

yuyuyu101:
lgtm

src/os/ObjectStore.h
```diff
@@ -1665,123 +1393,73 @@ class ObjectStore {
...
+
+      /* Actual version should be 9, but some version 9
+       * transactions ended up with version 8 */
+      if (struct_v == 8) {
 	bufferlist::iterator bl2 = bl;
 	try {
 	  decode8_5(bl, struct_v);
-	  use_tbl = true;
-	  decoded = true;
+	  assert(0 == "Now not support this format");
```
liewegas: If we don't need v8, then we can drop decode8_5() entirely, right?

liewegas: v8 -> v9 bug was fixed in 2598fc50749f7a1e8450c07b27cec5fa54b3e41f, which precedes jewel, so v8 support is safe to remove entirely.

src/os/ObjectStore.h
```diff
@@ -1665,123 +1393,73 @@ class ObjectStore {
...
-      DECODE_OLDEST(2);
-
-      bool decoded = false;
-      if (struct_v < 8) {
-	decode8_5(bl, struct_v);
-	use_tbl = true;
-	decoded = true;
-      }	else if (struct_v == 8) {
+      DECODE_START(9, bl);
+      DECODE_OLDEST(8);
```
liewegas: DECODE_OLDEST(9);

liewegas:
@majianpeng ping.  would love to get this in!

liewegas:
ping

majianpeng:
@liewegas .Sorry late to response. Update and rebase

liewegas:
Two problems: a crash in decode_attrset:
<pre>
(gdb) bt
#0  0x00007ff4aa7248d9 in raise () from /lib64/libpthread.so.0
#1  0x000055c06509031c in reraise_fatal (signum=6)
    at /home/sage/src/ceph/src/global/signal_handler.cc:72
#2  handle_fatal_signal (signum=6) at /home/sage/src/ceph/src/global/signal_handler.cc:134
#3  <signal handler called>
#4  0x00007ff4a92c2a28 in raise () from /lib64/libc.so.6
#5  0x00007ff4a92c462a in abort () from /lib64/libc.so.6
#6  0x00007ff4a9bf7add in __gnu_cxx::__verbose_terminate_handler() ()
   from /lib64/libstdc++.so.6
#7  0x00007ff4a9bf5936 in ?? () from /lib64/libstdc++.so.6
#8  0x00007ff4a9bf5981 in std::terminate() () from /lib64/libstdc++.so.6
#9  0x00007ff4a9bf5b99 in __cxa_throw () from /lib64/libstdc++.so.6
#10 0x000055c064adfdf7 in ceph::buffer::list::copy_in (crc_reset=<optimized out>, 
    src=<optimized out>, len=<optimized out>, off=<optimized out>, this=<optimized out>)
    at /home/sage/src/ceph/src/common/buffer.cc:1744
#11 0x000055c06539cd61 in ceph::buffer::list::iterator_impl<false>::copy (this=0x7ff49dcccc98, 
    len=<optimized out>, dest=...) at /home/sage/src/ceph/src/common/buffer.cc:1200
#12 0x000055c064eae2c0 in decode (p=..., bp=...)
    at /home/sage/src/ceph/src/include/encoding.h:242
#13 decode<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, ceph::buffer::ptr, std::less<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > (p=..., m=std::map with 1 elements = {...})
    at /home/sage/src/ceph/src/include/encoding.h:736
#14 ObjectStore::Transaction::iterator::decode_attrset (aset=std::map with 1 elements = {...}, 
    this=0x7ff49dcccc80) at /home/sage/src/ceph/src/os/ObjectStore.h:933
#15 FileStore::_do_transaction (this=this@entry=0x55c06f058000, t=..., op_seq=op_seq@entry=31, 
    trans_num=trans_num@entry=0, handle=handle@entry=0x7ff49dccda30)
    at /home/sage/src/ceph/src/os/filestore/FileStore.cc:2622
#16 0x000055c064eb207b in FileStore::_do_transactions (this=this@entry=0x55c06f058000, 
    tls=std::vector of length 1, capacity 1 = {...}, op_seq=31, 
    handle=handle@entry=0x7ff49dccda30)
    at /home/sage/src/ceph/src/os/filestore/FileStore.cc:2227
#17 0x000055c064eb2654 in FileStore::_do_op (this=0x55c06f058000, osr=<optimized out>, 
    handle=...) at /home/sage/src/ceph/src/os/filestore/FileStore.cc:1997
#18 0x000055c0652359cd in ThreadPool::worker (this=0x55c06f058cb0, wt=0x55c06f0fdb90)
    at /home/sage/src/ceph/src/common/WorkQueue.cc:132
#19 0x000055c065236b40 in ThreadPool::WorkThread::entry (this=<optimized out>)
    at /home/sage/src/ceph/src/common/WorkQueue.h:446
#20 0x00007ff4aa71b61a in start_thread () from /lib64/libpthread.so.0
#21 0x00007ff4a93905fd in clone () from /lib64/libc.so.6
(gdb) f 15
#15 FileStore::_do_transaction (this=this@entry=0x55c06f058000, t=..., op_seq=op_seq@entry=31, 
    trans_num=trans_num@entry=0, handle=handle@entry=0x7ff49dccda30)
    at /home/sage/src/ceph/src/os/filestore/FileStore.cc:2622
2622            i.decode_attrset(aset);
</pre>
easily reproducible by running vstart.

also a compile error,
<pre>
diff --git a/src/test/objectstore/test_transaction.cc b/src/test/objectstore/test_transaction.cc
index 31710d2..a452de7 100644
--- a/src/test/objectstore/test_transaction.cc
+++ b/src/test/objectstore/test_transaction.cc
@@ -115,7 +115,6 @@ ObjectStore::Transaction generate_transaction()
 TEST(Transaction, MoveRangesDelSrcObj)
 {
   auto t = ObjectStore::Transaction{};
-  t.set_use_tbl(false);
   t.nop();
 
   coll_t c(spg_t(pg_t(1,2), shard_id_t::NO_SHARD));
</pre>

majianpeng:
@liewegas . I repeat vstart w/ osd objectstore=fielstore and repeat 'reados setxattr'. But no crash happen. What's your commands? Thanks!

liewegas:
replaced by https://github.com/ceph/ceph/pull/9064

PR #9066, opened 2016-05-11
rbd: introduce a simple bench for read

yangdongsheng:
Currently, we have a bench-write for in rbd tool, but we don't have
a simple bench for read.

This patch rename the subcommand of bench-write to bench, at the same time,
introduce a --io-type to the bench command. Then we can do a simple bench
for write or read.

### DIFF
--- a/doc/man/8/rbd.rst
+++ b/doc/man/8/rbd.rst
@@ -369,11 +369,6 @@ Commands
   Release a lock on an image. The lock id and locker are
   as output by lock ls.
 
-:command:`bench-write` [--io-size *size-in-B/K/M/G/T*] [--io-threads *num-ios-in-flight*] [--io-total *total-size-to-write-in-B/K/M/G/T*] [--io-pattern seq | rand] *image-spec*
-  Generate a series of writes to the image and measure the write throughput and
-  latency.  Defaults are: --io-size 4096, --io-threads 16, --io-total 1G,
-  --io-pattern seq.
-
 Image and snap specs
 ====================
 
--- a/doc/man/8/rbd.rst
+++ b/doc/man/8/rbd.rst
@@ -369,6 +369,13 @@ Commands
   Release a lock on an image. The lock id and locker are
   as output by lock ls.
 
+:command:`bench` --io-type <read | write> [--io-size *size-in-B/K/M/G/T*] [--io-threads *num-ios-in-flight*] [--io-total *total-size-for-IO-in-B/K/M/G/T*] [--io-pattern seq | rand] *image-spec*
+  Generate a series of IOs to the image and measure the IO throughput and
+  latency.  Defaults are: --io-size 4096, --io-threads 16, --io-total 1G,
+  --io-pattern seq.
+
+  NOTE: --io-type must be specified.
+
 Image and snap specs
 ====================
 
--- a/src/test/cli/rbd/help.t
+++ b/src/test/cli/rbd/help.t
@@ -5,7 +5,7 @@
   
   Positional arguments:
     <command>
-      bench-write                 Simple write benchmark.
+      bench                       Simple benchmark.
       children                    Display children of snapshot.
       clone                       Clone a snapshot into a COW child image.
       copy (cp)                   Copy src image to dest.
@@ -96,13 +96,13 @@
   
   See 'rbd help <command>' for help on a specific command.
   $ rbd help | grep '^    [a-z]' | sed 's/^    \([a-z -]*[a-z]\).*/\1/g' | while read -r line; do echo rbd help $line ; rbd help $line; done
-  rbd help bench-write
-  usage: rbd bench-write [--pool <pool>] [--image <image>] [--io-size <io-size>] 
-                         [--io-threads <io-threads>] [--io-total <io-total>] 
-                         [--io-pattern <io-pattern>] 
-                         <image-spec> 
+  rbd help bench
+  usage: rbd bench [--pool <pool>] [--image <image>] [--io-size <io-size>] 
+                   [--io-threads <io-threads>] [--io-total <io-total>] 
+                   [--io-pattern <io-pattern>] --io-type <io-type> 
+                   <image-spec> 
   
-  Simple write benchmark.
+  Simple benchmark.
   
   Positional arguments
     <image-spec>         image specification
@@ -111,10 +111,11 @@
   Optional arguments
     -p [ --pool ] arg    pool name
     --image arg          image name
-    --io-size arg        write size (in B/K/M/G/T)
+    --io-size arg        IO size (in B/K/M/G/T)
     --io-threads arg     ios in flight
-    --io-total arg       total size to write (in B/K/M/G/T)
-    --io-pattern arg     write pattern (rand or seq)
+    --io-total arg       total size for IO (in B/K/M/G/T)
+    --io-pattern arg     IO pattern (rand or seq)
+    --io-type arg        IO type (read or write)
   
   rbd help children
   usage: rbd children [--pool <pool>] [--image <image>] [--snap <snap>] 
--- a/src/test/cli/rbd/invalid-snap-usage.t
+++ b/src/test/cli/rbd/invalid-snap-usage.t
@@ -88,7 +88,7 @@
   $ rbd lock remove foo@snap id client.1234
   rbd: snapname specified for a command that doesn't use it
   [22]
-  $ rbd bench-write foo@snap
+  $ rbd bench foo@snap --io-type write
   rbd: snapname specified for a command that doesn't use it
   [22]
 
--- a/src/test/cli/rbd/not-enough-args.t
+++ b/src/test/cli/rbd/not-enough-args.t
@@ -187,7 +187,7 @@
   $ rbd lock remove foo id
   rbd: locker was not specified
   [22]
-  $ rbd bench-write
+  $ rbd bench --io-type write
   rbd: image name was not specified
   [22]
   $ rbd mirror pool enable rbd
--- a/src/tools/rbd/action/Bench.cc
+++ b/src/tools/rbd/action/Bench.cc
@@ -16,13 +16,21 @@
 
 namespace rbd {
 namespace action {
-namespace bench_write {
+namespace bench {
 
 namespace at = argument_types;
 namespace po = boost::program_options;
 
 namespace {
 
+enum io_type_t {
+  IO_TYPE_READ = 0,
+  IO_TYPE_WRITE,
+
+  IO_TYPE_NUM,
+};
+
+struct IOType {};
 struct Size {};
 struct IOPattern {};
 
@@ -50,7 +58,26 @@ void validate(boost::any& v, const std::vector<std::string>& values,
   } else {
     throw po::validation_error(po::validation_error::invalid_option_value);
   }
+}
 
+io_type_t get_io_type(string io_type_string) {
+  if (io_type_string == "read")
+    return IO_TYPE_READ;
+  else if (io_type_string == "write")
+    return IO_TYPE_WRITE;
+  else
+    return IO_TYPE_NUM;
+}
+
+void validate(boost::any& v, const std::vector<std::string>& values,
+              IOType *target_type, int) {
+  po::validators::check_first_occurrence(v);
+  const std::string &s = po::validators::get_single_string(values);
+  io_type_t io_type = get_io_type(s);
+  if (io_type >= IO_TYPE_NUM)
+    throw po::validation_error(po::validation_error::invalid_option_value);
+  else
+    v = boost::any(io_type);
 }
 
 } // anonymous namespace
@@ -58,20 +85,47 @@ void validate(boost::any& v, const std::vector<std::string>& values,
 static void rbd_bencher_completion(void *c, void *pc);
 struct rbd_bencher;
 
+struct bencher_completer {
+  rbd_bencher *bencher;
+  bufferlist *bl;
+
+public:
+  bencher_completer(rbd_bencher *bencher, bufferlist *bl)
+    : bencher(bencher), bl(bl)
+  { }
+
+  ~bencher_completer()
+  {
+    if (bl)
+      delete bl;
+  }
+};
+
 struct rbd_bencher {
   librbd::Image *image;
   Mutex lock;
   Cond cond;
   int in_flight;
+  io_type_t io_type;
+  uint64_t io_size;
+  bufferlist write_bl;
 
-  explicit rbd_bencher(librbd::Image *i)
+  explicit rbd_bencher(librbd::Image *i, io_type_t io_type, uint64_t io_size)
     : image(i),
       lock("rbd_bencher::lock"),
-      in_flight(0)
-  { }
+      in_flight(0),
+      io_type(io_type),
+      io_size(io_size)
+  {
+    if (io_type == IO_TYPE_WRITE) {
+      bufferptr bp(io_size);
+      memset(bp.c_str(), rand() & 0xff, io_size);
+      write_bl.push_back(bp);
+    }
+  }
 
-  bool start_write(int max, uint64_t off, uint64_t len, bufferlist& bl,
-                   int op_flags)
+
+  bool start_io(int max, uint64_t off, uint64_t len, int op_flags)
   {
     {
       Mutex::Locker l(lock);
@@ -79,9 +133,20 @@ struct rbd_bencher {
         return false;
       in_flight++;
     }
-    librbd::RBD::AioCompletion *c =
-      new librbd::RBD::AioCompletion((void *)this, rbd_bencher_completion);
-    image->aio_write2(off, len, bl, c, op_flags);
+
+    librbd::RBD::AioCompletion *c;
+    if (io_type == IO_TYPE_READ) {
+      bufferlist *read_bl = new bufferlist();
+      c = new librbd::RBD::AioCompletion((void *)(new bencher_completer(this, read_bl)),
+					 rbd_bencher_completion);
+      image->aio_read2(off, len, *read_bl, c, op_flags);
+    } else if (io_type == IO_TYPE_WRITE) {
+      c = new librbd::RBD::AioCompletion((void *)(new bencher_completer(this, NULL)),
+					 rbd_bencher_completion);
+      image->aio_write2(off, len, write_bl, c, op_flags);
+    } else {
+      assert(0 == "Invalid io_type");
+    }
     //cout << "start " << c << " at " << off << "~" << len << std::endl;
     return true;
   }
@@ -100,11 +165,15 @@ struct rbd_bencher {
 void rbd_bencher_completion(void *vc, void *pc)
 {
   librbd::RBD::AioCompletion *c = (librbd::RBD::AioCompletion *)vc;
-  rbd_bencher *b = static_cast<rbd_bencher *>(pc);
+  bencher_completer *bc = static_cast<bencher_completer *>(pc);
+  rbd_bencher *b = bc->bencher;
   //cout << "complete " << c << std::endl;
   int ret = c->get_return_value();
-  if (ret != 0) {
+  if (b->io_type == IO_TYPE_WRITE && ret != 0) {
     cout << "write error: " << cpp_strerror(ret) << std::endl;
+    exit(ret < 0 ? -ret : ret);
+  } else if (b->io_type == IO_TYPE_READ && (unsigned int)ret != b->io_size) {
+    cout << "read error: " << cpp_strerror(ret) << std::endl;
     exit(ret < 0 ? -ret : ret);
   }
   b->lock.Lock();
@@ -112,15 +181,17 @@ void rbd_bencher_completion(void *vc, void *pc)
   b->cond.Signal();
   b->lock.Unlock();
   c->release();
+  delete bc;
 }
 
-int do_bench_write(librbd::Image& image, uint64_t io_size,
-                   uint64_t io_threads, uint64_t io_bytes,
-                   bool random)
+int do_bench(librbd::Image& image, io_type_t io_type,
+		   uint64_t io_size, uint64_t io_threads,
+		   uint64_t io_bytes, bool random)
 {
-  rbd_bencher b(&image);
+  rbd_bencher b(&image, io_type, io_size);
 
-  std::cout << "bench-write "
+  std::cout << "bench "
+       << " type " << (io_type == IO_TYPE_READ ? "read" : "write")
        << " io_size " << io_size
        << " io_threads " << io_threads
        << " bytes " << io_bytes
@@ -128,11 +199,6 @@ int do_bench_write(librbd::Image& image, uint64_t io_size,
        << std::endl;
 
   srand(time(NULL) % (unsigned long) -1);
-
-  bufferptr bp(io_size);
-  memset(bp.c_str(), rand() & 0xff, io_size);
-  bufferlist bl;
-  bl.push_back(bp);
 
   utime_t start = ceph_clock_now(NULL);
   utime_t last;
@@ -145,7 +211,7 @@ int do_bench_write(librbd::Image& image, uint64_t io_size,
   uint64_t i;
   uint64_t start_pos;
 
-  // disturb all thread's offset, used by seq write
+  // disturb all thread's offset, used by seq IO
   for (i = 0; i < io_threads; i++) {
     start_pos = (rand() % (size / io_size)) * io_size;
     thread_offset.push_back(start_pos);
@@ -186,7 +252,7 @@ int do_bench_write(librbd::Image& image, uint64_t io_size,
           thread_offset[i] = 0;
       }
 
-      if (!b.start_write(io_threads, thread_offset[i], io_size, bl, op_flags))
+      if (!b.start_io(io_threads, thread_offset[i], io_size, op_flags))
         break;
 
       ++i;
@@ -233,25 +299,42 @@ int do_bench_write(librbd::Image& image, uint64_t io_size,
   return 0;
 }
 
-void get_arguments(po::options_description *positional,
-                   po::options_description *options) {
+void add_bench_common_options(po::options_description *positional,
+			      po::options_description *options) {
   at::add_image_spec_options(positional, options, at::ARGUMENT_MODIFIER_NONE);
-  // TODO
+
   options->add_options()
-    ("io-size", po::value<Size>(), "write size (in B/K/M/G/T)")
+    ("io-size", po::value<Size>(), "IO size (in B/K/M/G/T)")
     ("io-threads", po::value<uint32_t>(), "ios in flight")
-    ("io-total", po::value<Size>(), "total size to write (in B/K/M/G/T)")
-    ("io-pattern", po::value<IOPattern>(), "write pattern (rand or seq)");
+    ("io-total", po::value<Size>(), "total size for IO (in B/K/M/G/T)")
+    ("io-pattern", po::value<IOPattern>(), "IO pattern (rand or seq)");
 }
 
-int execute(const po::variables_map &vm) {
+void get_arguments_for_write(po::options_description *positional,
+                             po::options_description *options) {
+  add_bench_common_options(positional, options);
+}
+
+void get_arguments_for_bench(po::options_description *positional,
+                             po::options_description *options) {
+  add_bench_common_options(positional, options);
+
+  options->add_options()
+    ("io-type", po::value<IOType>()->required(), "IO type (read or write)");
+}
+
+int bench_execute(const po::variables_map &vm, io_type_t bench_io_type) {
   size_t arg_index = 0;
   std::string pool_name;
   std::string image_name;
   std::string snap_name;
+  utils::SnapshotPresence snap_presence = utils::SNAPSHOT_PRESENCE_NONE;
+  if (bench_io_type == IO_TYPE_READ)
+    snap_presence = utils::SNAPSHOT_PRESENCE_PERMITTED;
+
   int r = utils::get_pool_image_snapshot_names(
     vm, at::ARGUMENT_MODIFIER_NONE, &arg_index, &pool_name, &image_name,
-    &snap_name, utils::SNAPSHOT_PRESENCE_NONE, utils::SPEC_VALIDATION_NONE);
+    &snap_name, snap_presence, utils::SPEC_VALIDATION_NONE);
   if (r < 0) {
     return r;
   }
@@ -293,18 +376,39 @@ int execute(const po::variables_map &vm) {
     return r;
   }
 
-  r = do_bench_write(image, bench_io_size, bench_io_threads, bench_bytes,
-                     bench_random);
+  r = do_bench(image, bench_io_type, bench_io_size, bench_io_threads,
+		     bench_bytes, bench_random);
   if (r < 0) {
-    std::cerr << "bench-write failed: " << cpp_strerror(r) << std::endl;
+    std::cerr << "do_bench failed: " << cpp_strerror(r) << std::endl;
     return r;
   }
   return 0;
 }
 
-Shell::Action action(
-  {"bench-write"}, {}, "Simple write benchmark.", "", &get_arguments, &execute, false);
+int execute_for_write(const po::variables_map &vm) {
+  std::cerr << "rbd: bench-write is deprecated, use rbd bench --io-type write ..." << std::endl;
+  return bench_execute(vm, IO_TYPE_WRITE);
+}
 
-} // namespace bench_write
+int execute_for_bench(const po::variables_map &vm) {
+  io_type_t bench_io_type;
+  if (vm.count("io-type")) {
+    bench_io_type = vm["io-type"].as<io_type_t>();
+  } else {
+    std::cerr << "rbd: --io-type must be specified." << std::endl;
+    return -EINVAL;
+  }
+
+  return bench_execute(vm, bench_io_type);
+}
+
+Shell::Action action_write(
+  {"bench-write"}, {}, "Simple write benchmark. (Deprecated, please use `rbd bench --io-type write` instead.)",
+		   "", &get_arguments_for_write, &execute_for_write, false);
+
+Shell::Action action_bench(
+  {"bench"}, {}, "Simple benchmark.", "", &get_arguments_for_bench, &execute_for_bench);
+
+} // namespace bench
 } // namespace action
 } // namespace rbd
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1261,7 +1261,7 @@ if(${WITH_RBD})
     tools/rbd/OptionPrinter.cc
     tools/rbd/Shell.cc
     tools/rbd/Utils.cc
-    tools/rbd/action/BenchWrite.cc
+    tools/rbd/action/Bench.cc
     tools/rbd/action/Children.cc
     tools/rbd/action/Clone.cc
     tools/rbd/action/Copy.cc
--- a/src/test/cli-integration/rbd/formatted-output.t
+++ b/src/test/cli-integration/rbd/formatted-output.t
@@ -33,7 +33,7 @@ clone
   $ rbd clone --image-feature layering,exclusive-lock,object-map,fast-diff bar@snap rbd_other/child
   $ rbd snap create rbd_other/child@snap
   $ rbd flatten rbd_other/child 2> /dev/null
-  $ rbd bench-write rbd_other/child --io-pattern seq --io-total 1B > /dev/null 2>&1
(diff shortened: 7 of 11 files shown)


src/tools/rbd/action/Bench.cc
```diff
@@ -293,18 +332,18 @@ int execute(const po::variables_map &vm) {
...
-    std::cerr << "bench-write failed: " << cpp_strerror(r) << std::endl;
+    std::cerr << "do_bench failed: " << cpp_strerror(r) << std::endl;
     return r;
   }
   return 0;
 }
 
 Shell::Action action(
-  {"bench-write"}, {}, "Simple write benchmark.", "", &get_arguments, &execute);
+  {"bench"}, {}, "Simple benchmark.", "", &get_arguments, &execute);
```
tchaikov: this breaks some tests in 'src/test/cli/rbd/'

see https://jenkins.ceph.com/job/ceph-pull-requests/5474/console .

yangdongsheng: Oh, thanx, I forgot to update the tests.

dillaman: You could leave 'bench-write' and add a 'bench-read' -- each pointing to its own ```execute_bench_write``` or ```execute_bench_read``` method that just invokes ```execute``` with the read or write option.  That way, there are no backwards compatibility issues with users of 'bench-write'

yangdongsheng: thanx @dillaman that's my first version of this feature, but I found there was lots of redundancy. But the compatibility is important for any project. Then I will add a bench-read rather than rename bench-write to bench. And refactor them to reduce the redundancy.

yangdongsheng: Oh, my bad. I found it was my fault to add bench-read. Anyway, will do it as you suggested. thanx @dillaman

trociny:
Sure, compatibility is important, but I like `rbd bench write/read/...` more than `rbd bench-write`, `rbd bench-read`... It is more consistent with our other rbd commands.

May be introduce  `rbd bench write/read` and leave  `rbd bench-write` as an alias marking it deprecated? It looks like it is not hard to do this with the current framework.

doc/man/8/rbd.rst
```diff
@@ -363,6 +363,11 @@ Commands
   latency.  Defaults are: --io-size 4096, --io-threads 16, --io-total 1G,
   --io-pattern seq.
 
+:command:`bench-read` [--io-size *size-in-B/K/M/G/T*] [--io-threads *num-ios-in-flight*] [--io-total *total-size-to-write-in-B/K/M/G/T*] [--io-pattern seq | rand] *image-spec*
```
trociny: s/total-size-to-write/total-size-to-read/ ?

src/tools/rbd/action/Bench.cc
```diff
@@ -16,13 +16,16 @@
 
 namespace rbd {
 namespace action {
-namespace bench_write {
+namespace bench{
```
trociny: white space is missing

src/tools/rbd/action/Bench.cc
```diff
@@ -16,13 +16,16 @@
...
 
 namespace at = argument_types;
 namespace po = boost::program_options;
 
 namespace {
 
+#define BENCH_READ	true
+#define BENCH_WRITE	false
+
```
trociny: Won't we want in future another IO types, e.g read/write mix? May be use enum instead of bool?

doc/man/8/rbd.rst
yangdongsheng: thanx

src/tools/rbd/action/Bench.cc
yangdongsheng: thanx

src/tools/rbd/action/Bench.cc
yangdongsheng: Yes, makes sense to me.

yangdongsheng:
@trociny Yes, making bench-write as an alias and marking it as deprecated sounds good.

yangdongsheng:
@trociny @tchaikov @dillaman Updated, please help to take a look. thanx

src/tools/rbd/action/Bench.cc
```diff
@@ -233,18 +264,30 @@ int do_bench_write(librbd::Image& image, uint64_t io_size,
...
   // TODO
   options->add_options()
-    ("io-size", po::value<Size>(), "write size (in B/K/M/G/T)")
+    ("io-size", po::value<Size>(), "op size (in B/K/M/G/T)")
     ("io-threads", po::value<uint32_t>(), "ios in flight")
-    ("io-total", po::value<Size>(), "total size to write (in B/K/M/G/T)")
-    ("io-pattern", po::value<IOPattern>(), "write pattern (rand or seq)");
+    ("io-total", po::value<Size>(), "total size for IO (in B/K/M/G/T)")
+    ("io-pattern", po::value<IOPattern>(), "io pattern (rand or seq)");
 }
```
trociny: If you use it own get_arguments for bench-write, then I think it is better to leave options descriptions as they were before ('write' instead of 'IO')? Also, I would remove TODO comment in get_arguments_for_write (not sure what it means, but I think we don't want to improve this function in the future).

src/tools/rbd/action/Bench.cc
```diff
@@ -293,18 +336,36 @@ int execute(const po::variables_map &vm) {
...
+  if (vm.count("io-type")) {
+    bench_io_type = vm["io-type"].as<enum io_type>();
+  } else {
+    bench_io_type = bench_read;
+  }
+  return bench_execute(vm, bench_io_type);
+}
+
+Shell::Action action_write(
+  {"bench-write"}, {}, "Simple write benchmark. (Deprecated, please use `rbd bench IMAGE --io-type write` instead.)",
```
trociny: May be just "please use `rbd bench` instead"  to make it shorter? I think this info will be enough for the user to figure out the right command.

src/tools/rbd/action/Bench.cc
yangdongsheng: Oh, true, I should keep the write word here. thanx. About the "TODO", I am not sure, @tchaikov @dillaman , any idea about what does TODO here mean?

src/tools/rbd/action/Bench.cc
yangdongsheng: sounds better. thanx

trociny:
lgtm

doc/man/8/rbd.rst
```diff
@@ -361,6 +361,11 @@ Commands
 :command:`bench-write` [--io-size *size-in-B/K/M/G/T*] [--io-threads *num-ios-in-flight*] [--io-total *total-size-to-write-in-B/K/M/G/T*] [--io-pattern seq | rand] *image-spec*
```
dillaman: Might as well remove this from the man pages at this time.

src/test/cli/rbd/help.t
```diff
@@ -5,7 +5,9 @@
   
   Positional arguments:
     <command>
-      bench-write                 Simple write benchmark.
+      bench                       Simple benchmark.
+      bench-write                 Simple write benchmark. (Deprecated, please use
```
dillaman: Minor: I would just say "deprecated: please use `rbd bench --io-type write` instead"

src/tools/rbd/action/Bench.cc
```diff
@@ -16,13 +16,19 @@
...
-namespace bench_write {
+namespace bench {
 
 namespace at = argument_types;
 namespace po = boost::program_options;
 
 namespace {
 
+enum io_type {
+  bench_read = 0,
```
dillaman: Minor: enums should be named like constants (e.g. IO_TYPE_READ and IO_TYPE_WRITE)

src/tools/rbd/action/Bench.cc
```diff
@@ -63,14 +81,18 @@ struct rbd_bencher {
   Mutex lock;
   Cond cond;
   int in_flight;
+  enum io_type io_type;
+  uint64_t io_size;
 
-  explicit rbd_bencher(librbd::Image *i)
+  explicit rbd_bencher(librbd::Image *i, enum io_type io_type, uint64_t io_size)
```
dillaman: Minor: no need for "enum"

src/tools/rbd/action/Bench.cc
```diff
@@ -81,7 +103,12 @@ struct rbd_bencher {
     }
     librbd::RBD::AioCompletion *c =
       new librbd::RBD::AioCompletion((void *)this, rbd_bencher_completion);
-    image->aio_write2(off, len, bl, c, op_flags);
+    if (io_type == bench_read)
+      image->aio_read2(off, len, bl, c, op_flags);
+    else if (io_type == bench_write)
+      image->aio_write2(off, len, bl, c, op_flags);
+    else
+      cout << "Invalidate io_type: " << io_type << std::endl;
```
dillaman: assert(false) -- this case should never happen

src/tools/rbd/action/Bench.cc
```diff
@@ -293,18 +336,36 @@ int execute(const po::variables_map &vm) {
...
+int execute_for_write(const po::variables_map &vm) {
+  return bench_execute(vm, bench_write);
+}
+
+int execute_for_bench(const po::variables_map &vm) {
+  enum io_type bench_io_type;
+  if (vm.count("io-type")) {
+    bench_io_type = vm["io-type"].as<enum io_type>();
+  } else {
+    bench_io_type = bench_read;
```
dillaman: I'd rather we exit with a failure if the user didn't specify the io-type for the benchmark as opposed to defaulting to read.

yangdongsheng: but other options all have a default value.

src/tools/rbd/action/Bench.cc
yangdongsheng: But:
```
/data/ceph/src/tools/rbd/action/Bench.cc:87:42: error: \u2018io_type\u2019 is not a type
   explicit rbd_bencher(librbd::Image *i, io_type io_type, uint64_t io_size)
```

src/tools/rbd/action/Bench.cc
yangdongsheng: Okey

doc/man/8/rbd.rst
yangdongsheng: Do you mean remove the bench-write related entry from manpage? Why not to keep it with a Deprecated warning until we remove it actually from rbd command?

src/test/cli/rbd/help.t
yangdongsheng: okey

src/tools/rbd/action/Bench.cc
yangdongsheng: What about assert(0 == "Invalidate io_type")?

trociny: Shouldn't be "invalid"? ("invalidate" is a verb)

src/tools/rbd/action/Bench.cc
trociny: It is because type name matches variable name. Change type name to `io_type_t`? Or use CamelCase.

src/tools/rbd/action/Bench.cc
```diff
@@ -63,14 +81,18 @@ struct rbd_bencher {
   Mutex lock;
   Cond cond;
   int in_flight;
+  enum io_type io_type;
```
trociny: Don't forget to remove enum here too (and in all other places if exist).

doc/man/8/rbd.rst
trociny: Agree with Jason -- I would better to remove this entry, so new users even not suspect it exists. Though marking it as deprecated here is fine to me too.

I would even like to see it was removed from mini help too and to output the "deprecated" warning when it is run with this option, but you need to modify Shell::print_help for this.

yangdongsheng: Okey, will try it.

src/tools/rbd/action/Bench.cc
yangdongsheng: That's true. thanx

yangdongsheng:
@trociny @dillaman what about this version? :smile:

trociny:
lgtm, though I am not sure I like how you split the patch. This commit "tools/rbd: make bench-write invisible to user" makes invisible it only in doc, but not in tool. Is it your intention? I would either set invisible flag in the code in this commit, or change the commit message to: "doc: make rbd bench-write invisible to user", or squash it with the main commit.

yangdongsheng:
Oh, my bad, I forgot to git add the change about tool for that commit.

yangdongsheng:
Updated now. :laughing:

src/test/cli/rbd/help.t
```diff
@@ -5,7 +5,7 @@
   
   Positional arguments:
     <command>
-      bench-write                 Simple write benchmark.
```
trociny: This remove (and below) should also be moved to "make bench-write invisible to user" commit.

src/tools/rbd/action/Bench.cc
```diff
@@ -293,18 +336,36 @@ int execute(const po::variables_map &vm) {
...
+int execute_for_write(const po::variables_map &vm) {
+  return bench_execute(vm, BENCH_WRITE);
+}
+
+int execute_for_bench(const po::variables_map &vm) {
+  io_type_t bench_io_type;
+  if (vm.count("io-type")) {
+    bench_io_type = vm["io-type"].as<io_type_t>();
+  } else {
+    bench_io_type = BENCH_READ;
```
dillaman: Getting back to my previous comment: I still don't like the idea of defaulting the IO type.  This is equivalent to requiring "--size" on "rbd create"

trociny:
I am observing a warning when compiling it:
```
tools/rbd/action/Bench.cc:136:46: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
   } else if (b->io_type == BENCH_READ && ret != b->io_size) {
```

Also, have you tried to build it with lttng? Because when I run the binary built with lttng, I have a crash in librbd.cc:Image::aio_read2() -> tracepoint(), when it is accessing bl.c_str().

I have not investigated much what the root cause could be, just noticed that for read operations it is not necessary to prepare bufferlist in the way we prepare it for write, and when I disable this it fixes the crush too.

Could you please look at it?

src/test/cli/rbd/help.t
yangdongsheng: Oh, sorry for my lazy. will correct it.

src/tools/rbd/action/Bench.cc
yangdongsheng: Thanx @dillaman , yes, that's another confusion in my mind. If we want to make the size to be required, what about making it similar with qemu-img? E.g:
```
# qemu-img create test
qemu-img: test: Image creation needs a size parameter
# qemu-img create test 1G
Formatting 'test', fmt=raw size=1073741824 
```

Because ```--xxx``` gives me a feeling that it's optional and programmers have put an default value for my normal use case.

So can we do it in a similar way? such as:
```
# rbd bench read test
and 
# rbd bench write test
```

Both of them, ```rbd create```way and ```qemu-img create``` way, are acceptable to me, but I want to make sure it's a convention in our ceph world, so that if we meet another same situation, we can do it as what ```rbd create``` did. is that correct? @dillaman @trociny

yangdongsheng:
@trociny Okey, will investigate it.

src/tools/rbd/action/Bench.cc
trociny: I like `rbd bench read|write img` a little more than `rbd bench img --io-type read|write`, especially if we want `io-type` option to be non-optional, though I don't have a strong opinion, I would be fine with any way.

dillaman: You can make it a required first positional argument if not specified via "--io-type" (same as rbd <action> --image foo vs rbd action foo)

yangdongsheng:
@dillaman I have made the --io-type working as --image. thanx for your suggestion. 

@trociny I skipped the memset for BENCH_READ, because that's not necessary at all, but I did not reproduce the crash problem you mentioned.

src/tools/rbd/action/Bench.cc
```diff
@@ -26,6 +26,8 @@ namespace {
 enum io_type_t {
   BENCH_READ = 0,
   BENCH_WRITE,
+
+  BENCH_LAST,
```
trociny: I would name it BENCH_UNKNOW and set it the first in the list (= 0).

src/tools/rbd/action/Bench.cc
```diff
@@ -58,17 +60,24 @@ void validate(boost::any& v, const std::vector<std::string>& values,
...
               IOType *target_type, int) {
   po::validators::check_first_occurrence(v);
   const std::string &s = po::validators::get_single_string(values);
-  if (s == "read") {
-    v = boost::any(BENCH_READ);
-  } else if (s == "write") {
-    v = boost::any(BENCH_WRITE);
-  } else {
+  io_type_t io_type = get_io_type(s);
+  if (io_type >= BENCH_LAST)
```
trociny: This could be then 
```
if (io_type == BENCH_UNKNOW)
``

File: src/tools/rbd/action/Bench.cc [hunk 18]
```diff
@@ -352,13 +362,24 @@ int execute_for_write(const po::variables_map &vm) {
...
-    bench_io_type = BENCH_READ;
+    std::string io_type_string;
+    io_type_string = utils::get_positional_argument(vm, arg_index++);
+    if (io_type_string.empty()) {
+      std::cerr << "rbd bench: io-type was not specified." << std::endl;
+      return -EINVAL;
+    }
+    bench_io_type = get_io_type(io_type_string);
+    if (bench_io_type >= BENCH_LAST) {
+      std::cerr << "rbd bench: Invalide bench io-type was specified: " << io_type_string << std::endl;
```
trociny: s/Invalide/Invalid/

trociny:
I think the last two commits can be stashed into one in the final version?

src/tools/rbd/action/Bench.cc
```diff
@@ -16,13 +16,21 @@
...
 namespace po = boost::program_options;
 
 namespace {
 
+enum io_type_t {
+  BENCH_READ = 0,
+  BENCH_WRITE,
+
+  BENCH_LAST,
+};
```
trociny: Also, you could consider using names IO_TYPE_READ, IO_TYPE_WRITE... as it has been suggested by Jason, to follow convention.

src/tools/rbd/action/Bench.cc
dillaman: BENCH_UNKNOWN

yangdongsheng: Thanx @trociny and @dillaman , but from my experience in linux kernel development, put the invalid enum in the end is a good idea.
(1), don't make 0 as UNKNOWN, making 0 as a start point of valid number looks more nature.
(2), BENCH_LAST, can not only tell us the last one, example 2 in this case, is a invalid value, but also it can tell us any one larger than BENCH_LAST is invalid.
(3), BENCH_LAST can also tell us how many io_types we are supporting, because the value of it is the count of the io_type, 2 in my case. Even when we expand our rbd bench for mix mode, then BENCH_LAST would be changed to 3, because it's the last one in our enum definition.

But, It seems that you think that's confusing to you, what about renaming it to IO_TYPE_NUM. and still put it at the last.

Then we can get lots of information from this enum.
(1), if (io_type == IO_TYPE_NUM) -> invalid io_type
(2), if (io_type > IO_TYPE_NUM) -> invalid io_type
(3), the number of io_types we are supporting currently is IO_TYPE_NUM.

src/tools/rbd/action/Bench.cc
yangdongsheng: thanx

src/tools/rbd/action/Bench.cc
yangdongsheng: Okey, good.

yangdongsheng:
@trociny and @dillaman I decided to remove --io-type, but rename rbd bench to "rbd bench read" and "rbd bench write", because there would be a problem in "rbd bench write test --io-type read". So I removed the --io-type.

dillaman:
@yangdongsheng You can use a different "get_arguments" method for each command -- so for "bench write", don't register the "--io-type" option.

yangdongsheng:
@dillaman Actually --io-type did not appear before, that's what I introduced by my patch in last version. But now, I have removed it, then this option should have not been here ever. Then we can do bench like below:
```
$ rbd bench-write [IMAGE]  -> same with what it working.
$ rbd bench write [IMAGE]  -> equal to "rbd bench-write"
$ rbd bench read [IMAGE]  ->  simple bench for read
```
looks good?

dillaman:
I'd prefer not adding "bench read" and "bench write" as separate commands because if in the future we want "readwrite" we need a new command as well.

yangdongsheng:
So you want to add just one command named as bench, and make `read` and `write` as a positional argument for it? Hmmm,,, that's okey to me, but I am not sure what's the problem if we introduce a new command "readwrite" in the future, could you share your opinion more with me, thanx :)

dillaman:
@yangdongsheng Just because a lot of people (myself include) prefer to use the "--<option-name>" instead of the positionals.  The positionals are really awkward to use (especially when you have a lot of possible knobs and levers to turn).

yangdongsheng:
@dillaman Okey, sounds reasonable to me. so, what about this one
```
$ rbd bench-write [IMAGE]  -> same with what it working.
$ rbd bench [IMAGE] --io-type write -> equal to "rbd bench-write"
$ rbd bench [IMAGE] --io-type read ->  simple bench for read
```

And do not support the positional argument version. that means, if you type "rbd bench read IAMGE", you will get an error.

This way, we can expand it for mixed mode such as "rbd bench IMANGE --io-type readwrite".

Hmmm, wait, let me think about why I started to struggle with this problem. Oh, because I don't like to make --io-type to be required. it's a optional argument but it's not optional. So, @dillaman do you still think I should make it required ?

dillaman:
@yangdongsheng No -- that is not what I am saying.  My suggestion is that the command should be registered as "bench" and it should support a first positional of <io-type> along with the optional "--io-type".  If the optional "--io-type" is provided, you shouldn't attempt to read the <io-type> from the positional:

```
if (vm.count("io-type")) {
  io_type = vm["io-type"].as<io_type_t>(); // assumes you have a validator
} else {
  std::string io_type_cmd =utils::get_positional_argument(vm, arg_index);
  io_type = <re-use same validator to convert to enum>(io_type_cmd);
  ++arg_index;
}
```

Also, while I acknowledge that there is great debate about whether or not "--<xyz>" options should be strictly optionals or not (just look at the history of boost::program_options and the fight over allowing options to be flagged as required), I believe in general most people have come down on the side of using them for clarity.  Just look at the OpenStack client CLI -- everything needs to be specified using "optionals" because otherwise things quickly become a giant positional mess.

yangdongsheng:
thanx @dillaman , Okey, I am okey with making --io-type required now. 

but I think we should only provide the --io-type, rather than fall back to check the positional argument. Because there would be a big confusion:
```rbd bench write test --io-type read```
In this case, we will do a read bench, but that looks weird.

So I think let's only support --io-type.
```
rbd bench test --io-type write
```
that's the only correct input. 

Sounds good?

**Comment by dillaman** [comment]
Actually -- in your case above it would attempt to read from the image named "write" (which is what happens with a lot of rbd CLI commands when you mix-n-match the options with the positionals).

**Comment by yangdongsheng** [comment]
@dillaman Oh, yes, it should get image name as "write". But if we have an image named as "write", that command will do another thing different with what we want. I think that's bad. Another example,
```
# ./bin/rbd info test --image yds
rbd image 'yds':
	size 1024 MB in 256 objects
	order 22 (4096 kB objects)
	block_name_prefix: rbd_data.10142ae8944a
	format: 2
	features: layering, exclusive-lock, object-map, fast-diff, deep-flatten
	flags: 
```

Actually, I mean, if we can get an agreement, that "--" options can be set as required, then we have no necessary to introduce positional version of them. that just makes our command more confusing I think.

Maybe just support "--io-type" can make our life more clear and easier.

**Comment by dillaman** [comment]
@yangdongsheng I'm fine with that as well -- we have other required options already.

**Comment by yangdongsheng** [comment]
@dillaman Great. I will update it tomorrow, that's too late now. :crescent_moon:

**Comment by yangdongsheng** [comment]
@dillaman updated

File: doc/man/8/rbd.rst [hunk 20]
```diff
@@ -358,6 +358,13 @@ Commands
   Release a lock on an image. The lock id and locker are
   as output by lock ls.
 
+:command:`bench` [--io-type read | write] [--io-size *size-in-B/K/M/G/T*] [--io-threads *num-ios-in-flight*] [--io-total *total-size-for-IO-in-B/K/M/G/T*] [--io-pattern seq | rand] *image-spec*
```
dillaman: Minor: remove the "[" and "]" around [--io-type read | write] to imply that it is required.  The automatically generated help text already does this.

src/tools/rbd/action/Bench.cc
```diff
@@ -293,18 +348,38 @@ int execute(const po::variables_map &vm) {
...
+    std::cerr << "do_bench failed: " << cpp_strerror(r) << std::endl;
     return r;
   }
   return 0;
 }
 
-Shell::Action action(
-  {"bench-write"}, {}, "Simple write benchmark.", "", &get_arguments, &execute, false);
+int execute_for_write(const po::variables_map &vm) {
+  return bench_execute(vm, IO_TYPE_WRITE);
```
dillaman: Since ```bench-write``` is hidden in the help menu now, it would probably be a good idea to output a warning that this is deprecated and to switch to "bench --io-type write".

dillaman: Something like: `std::cerr << "rbd: bench-write is deprecated, use bench --io-type write" << std::endl;`

dillaman:
You will have to update https://github.com/ceph/ceph/blob/master/src/test/cli-integration/rbd/formatted-output.t when 'bench-write' outputs a deprecated warning.  Probably just easier to switch it to 'rbd bench --io-type write'

dillaman:
lgtm after minor comments

trociny:
I still can trigger crush when running `bench read` for long time on with small io-size:
```
rbd bench --io-type read --io-size 10 --io-total 1024000 test
```

In valgrind it is reported as:
```
==21119== Invalid read of size 8
==21119==    at 0xA70EEC8: ceph::buffer::list::c_str() (buffer.cc:1793)
==21119==    by 0xA5BE017: librbd::Image::aio_read2(unsigned long, unsigned long, ceph::buffer::list&, librbd::RBD::AioCompletion*, int) (librbd.cc:1164)
==21119==    by 0x26A6FA: start_io (Bench.cc:116)
==21119==    by 0x26A6FA: do_bench (Bench.cc:231)
==21119==    by 0x26A6FA: rbd::action::bench::bench_execute(boost::program_options::variables_map const&, rbd::action::bench::(anonymous namespace)::io_type_t) (Bench.cc:352)
==21119==    by 0x26B906: rbd::action::bench::execute_for_bench(boost::program_options::variables_map const&) (Bench.cc:373)
==21119==    by 0x258B81: rbd::Shell::execute(std::vector<char const*, std::allocator<char const*> > const&) (Shell.cc:163)
==21119==    by 0x2243B0: main (rbd.cc:19)
==21119==  Address 0x2e4b3c50 is 0 bytes inside a block of size 32 free'd
==21119==    at 0xA33F360: operator delete(void*) (vg_replace_malloc.c:507)
==21119==    by 0x261CE6: deallocate (new_allocator.h:110)
==21119==    by 0x261CE6: _M_put_node (stl_list.h:347)
==21119==    by 0x261CE6: std::_List_base<ceph::buffer::ptr, std::allocator<ceph::buffer::ptr> >::_M_clear() (list.tcc:79)
==21119==    by 0xA70D6C5: clear (stl_list.h:1304)
==21119==    by 0xA70D6C5: clear (buffer.h:446)
==21119==    by 0xA70D6C5: ceph::buffer::list::claim(ceph::buffer::list&, unsigned int) (buffer.cc:1594)
==21119==    by 0xA5D7FD1: librbd::AioCompletion::finalize(CephContext*, long) (AioCompletion.cc:64)
==21119==    by 0xA5D8961: librbd::AioCompletion::complete_request(CephContext*, long) (AioCompletion.cc:174)
==21119==    by 0xA5D92D4: finish (AioCompletion.h:208)
==21119==    by 0xA5D92D4: librbd::C_AioRead::finish(int) (AioCompletion.cc:226)
==21119==    by 0xA5D1B12: Context::complete(int) (Context.h:64)
==21119==    by 0xA5E56F6: librbd::AioObjectRequest::complete(int) (AioObjectRequest.cc:54)
==21119==    by 0xA5D9113: finish (AioCompletion.cc:242)
==21119==    by 0xA5D9113: complete (Context.h:64)
==21119==    by 0xA5D9113: librbd::C_CacheRead::complete(int) (AioCompletion.cc:237)
==21119==    by 0xA7EE3EE: ThreadPool::worker(ThreadPool::WorkThread*) (WorkQueue.cc:130)
==21119==    by 0xA7EF43F: ThreadPool::WorkThread::entry() (WorkQueue.h:445)
==21119==    by 0x1D68D0A3: start_thread (pthread_create.c:309)
```

I think the problem is that `bufferlist bl`, created in `do_bench`, is shared between all aio_reeds. It was ok for aio_write, because it was used read only there, but for read you need a separate bl for every io.

yangdongsheng:
@dillaman  @trociny okey, I reproduced it and solved it by new bufferlist for each read operation. Please check again.

trociny:
@yangdongsheng see in jenkins output:
```
src/test/cli/rbd/invalid-snap-usage.t: failed 
src/test/cli/rbd/not-enough-args.t: failed
```
Also, I am not very happy that a bufferlist is always passed to start_io(), but is used only for write. I would prefer either:
1) make it symetrical for read/write, i.e. allocate/deallocate bufferlist for every write too (I don't expect it will affect write bench results that it can be noticed), or
2) make write_bl a member of rbd_bencher, so don't pass it to start_io,
3) allocate read_bl  in do_bench(), and pass it to start_io() (I like this less, because it still would look confusing, why NULL is passed to bencher_completer() in write case.

Anyway, it is not mandatory, and I am fine with the current code too.

yangdongsheng:
Hi @trociny sorry for the late, Okey, I have put the write_bl into rbd_bencher. please check again. thanx

src/tools/rbd/action/Bench.cc
```diff
@@ -293,18 +372,39 @@ int execute(const po::variables_map &vm) {
...
+  std::cerr << "rbd: bench-write is deprecated, use bench --io-type write" << std::endl;
+  return bench_execute(vm, IO_TYPE_WRITE);
+}
+
+int execute_for_bench(const po::variables_map &vm) {
+  io_type_t bench_io_type;
+  if (vm.count("io-type")) {
+    bench_io_type = vm["io-type"].as<io_type_t>();
+  } else {
+    std::cerr << "rbd: --io-type must be spedified." << std::endl;
```
trociny: @yangdongsheng s/spedified/specified/

trociny:
otherwise, lgtm

yangdongsheng:
@trociny Good catch, it's updated now.

trociny:
@dillaman Does this latest version look good for your?

run-make-check.sh
```diff
@@ -66,7 +66,7 @@ function run() {
 	$DRY_RUN ./install-deps.sh || return 1
     fi
     $DRY_RUN ./autogen.sh || return 1
-    $DRY_RUN ./configure "$@"  --with-librocksdb-static --disable-static --with-radosgw --with-debug --without-lttng \
+    $DRY_RUN ./configure "$@"  --with-librocksdb-static --disable-static --with-radosgw --with-debug \
```
dillaman: Why is this change included with this PR?

doc/man/8/rbd.rst
```diff
@@ -358,6 +358,13 @@ Commands
   Release a lock on an image. The lock id and locker are
   as output by lock ls.
 
+:command:`bench` --io-type read | write [--io-size *size-in-B/K/M/G/T*] [--io-threads *num-ios-in-flight*] [--io-total *total-size-for-IO-in-B/K/M/G/T*] [--io-pattern seq | rand] *image-spec*
```
dillaman: Minor: "--io-type <read | write>"

run-make-check.sh
yangdongsheng: OMG, it's just a stupid "git commit -a " after a testing with lttng. Really sorry for that, I should check it carefully after pushing to github. thanx @dillaman

doc/man/8/rbd.rst
yangdongsheng: Yes, looks better.

yangdongsheng:
@dillaman updated now. Hoping there is no other foolish problem now.

src/test/cli/rbd/invalid-snap-usage.t
```diff
@@ -89,6 +89,7 @@
   rbd: snapname specified for a command that doesn't use it
   [22]
   $ rbd bench-write foo@snap
+  rbd: bench-write is deprecated, use bench --io-type write
```
trociny: @yangdongsheng sorry, missed this in my previous review. I think it is better to change this to `rbd bench --io-type write ...`, so don't use deprecated command. The same is for not-enough-args.t below.

yangdongsheng:
@trociny Okey, done.

src/test/cli/rbd/invalid-snap-usage.t
```diff
@@ -89,6 +89,7 @@
   rbd: snapname specified for a command that doesn't use it
   [22]
   $ rbd bench-write foo@snap
+  rbd: bench-write is deprecated, use rbd bench --io-type write ...
```
trociny: @yangdongsheng Udate this command too? I think it is better to remove bench-write usage everywhere.

Also, looking at this error, right now `bench --io-type read` returns this too, but I think we can allow bench read for snapshots?

src/test/cli/rbd/not-enough-args.t
```diff
@@ -188,6 +188,7 @@
   rbd: locker was not specified
   [22]
   $ rbd bench-write
+  rbd: bench-write is deprecated, use rbd bench --io-type write ...
```
trociny: The same: change to `rbd bench`?

yangdongsheng: wait, I am sure I have fixed this one, and it passed the run-cli-tests. Let me check it again.

yangdongsheng: Interesting, I have the same Commit ID in local and it contains these fixes, but the remote one in github is different. As they have the same ID, I can't push it now with an error of "Everything is updated.". So I will --ammend and push with --force to force update it.

src/test/cli/rbd/invalid-snap-usage.t
yangdongsheng: Sorry, I am not sure I understand your point. do you mean we will see this warning in read case? but I think this only happen in bench-write.

trociny: No, it is not about "bench-write is deprecated" but about "snapname specified for a command that doesn't use it" error this test actually written for. This error is valid only for write type, but right now it rises for both `--io-type write` and `--io-type read`. For the second one I think we should allow to run bench.

I.e. you should pass `SNAPSHOT_PRESENCE_NONE` flag to `get_pool_image_snapshot_names()` only for write io type.

trociny:
@yangdongsheng ping
Do you agree to fix utils::SNAPSHOT_PRESENCE_NONE issue for read io type I described earlier?

yangdongsheng:
Oh, @trociny Really sorry for the late. I am struggling with some other work these days. But I definitely agree with you about the utils::SNAPSHOT_PRESENCE_NONE. I will update it soon.

trociny:
@yangdongsheng no hurry, just wanted to know your plans about this. Thanks.

yangdongsheng:
@trociny Updated. :)

src/test/cli/rbd/not-enough-args.t
trociny: @yangdongsheng So, you still have these deprecated commands in invalid-snap-usage.t and 
 not-enough-args.t tests?

yangdongsheng:
@trociny Sorry for the late, it's updated in test files. Please help to review again. thanx a lot.

trociny:
lgtm, @dillaman do you have any comments for the latest version?

dillaman:
lgtm

PR #9067, opened 2016-05-11
cls::rbd: mirror_image_status_list returned max 64 items

trociny:
Signed-off-by: Mykola Golub <mgolub@mirantis.com>

PR #9068, opened 2016-05-11
librbd: initial python APIs to support mirroring

trociny:
http://tracker.ceph.com/issues/15656

### DIFF
--- a/src/pybind/rbd/rbd.pyx
+++ b/src/pybind/rbd/rbd.pyx
@@ -18,8 +18,10 @@ from cpython cimport PyObject, ref, exc
 from libc cimport errno
 from libc.stdint cimport *
 from libc.stdlib cimport realloc, free
+from libc.string cimport strdup
 
 from collections import Iterable
+from datetime import datetime
 
 cimport rados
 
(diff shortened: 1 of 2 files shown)


trociny:
Marked as DNM for now as it depends on #9067 to pass tests.

src/pybind/rbd/rbd.pyx
```diff
@@ -566,6 +660,314 @@ class RBD(object):
...
+
+    def __iter__(self):
+        while self.size > 0:
+            for i in range(self.size):
+                yield {
+                    'name'        : decode_cstr(self.images[i].name),
+                    'info'        : {
+                        'global_id' : decode_cstr(self.images[i].info.global_id),
+                        'state'     : self.images[i].info.state,
+                        'primary'   : self.images[i].info.primary,
```
dillaman: Since this is always set to ```false```, perhaps we can use this opportunity to leave it out of the dict?

trociny: I thought it would be good to be consistent with get_image_status, and provide the same set of fields. But thinking more, agree it would be rather confusing for users, who might start rely on it, so I will remove the key.

trociny:
rebased

dillaman:
@trociny the Cython compiles fine on my machine but breaks on the older distros:

http://gitbuilder.sepia.ceph.com/gitbuilder-ceph-deb-trusty-amd64-basic/log.cgi?log=28a6104ff1bc8d69f284c8bc1c51785d78706f71
http://gitbuilder.sepia.ceph.com/gitbuilder-ceph-rpm-centos7-amd64-basic/log.cgi?log=28a6104ff1bc8d69f284c8bc1c51785d78706f71

trociny:
@dillaman I found a machine where it is reproducible and now hope it is fixed. I have also added this to my testing branch to check gitbuiler. centos7 is ok. trusty builder looks like lugging today, still waiting for results.

PR #9070, opened 2016-05-11
doc: words of caution about cache tiering

liewegas:


athanatos:
Seems reasonable to me

bengland2:
typo in words of caution commit: "higher level of disk" => "higher level of risk".  But it is very useful to set people's expectations properly.  

What tools can they use to monitor efficiency of cache (i.e. cache hits/total cache accesses?)  and cache hit rate? .  I had to write something like this for dm-cache, it didn't exist AFAIK.  

Another factor that may not have been covered above is that for reads, a small enough working set of RADOS objects will get cached in RAM, so cache tiering is targeting working sets that are >>  OSD Hosts' Linux buffer cache  but << the entire cold pool, right?  Again, is there any way to estimate the "working set" size?

doc/rados/operations/cache-tiering.rst
```diff
@@ -55,35 +55,75 @@ configure how this migration takes place. There are two main scenarios:
...
+
+Cache tiering will *degrade* performance for most workloads.  Users should use
+extreme caution before using this feature.
+
+* *Workload dependent*: Whether a cache will improve performance is highly
+  dependent on the workload.  Because there is a cost associated with
+  moving objects into or out of the cache, it can only be effective
+  when there is a *large skew* in the access pattern in the data set,
+  such that most of the requests touch a small number of objects.
+
```
fiskn: Probably worth adding something like

1. Your cache tier should be 2x bigger than the size of this skew of hot objects to avoid thrashing the cache
2. If most of the IO is small random reads, don't use cache tiering if you have a large amount of page cache

doc/rados/operations/cache-tiering.rst
```diff
@@ -55,35 +55,75 @@ configure how this migration takes place. There are two main scenarios:
...
+  small writes are not yet supported by the erasure-coded pool, entire
+  (usually 4 MB) objects must be migrated into the cache in order to
+  satisfy a small (often 4 KB) write.  Only a handful of users have
+  successfully deployed this configuration, and it only works for them
+  because their data is extremely cold (backups) and they are not in
+  any way sensitive to performance.
+
+* *RBD with replicated cache and base*: RBD with a replicated base tier does
+  better than when the base is erasure coded, but it is still highly dependent
+  on the amount of skew in the workload, and very difficult to validate.
```
fiskn: I wonder if this would be better under "Unknown Workloads", but along with a warning that the user is expected to understand about their workload and how to tune the cache tier to fit it. I think labelling it as known bad can be discouraging when several people are starting to have good results if their workload is suitable.

It might also be worth amending some of the default settings further down the page. Such as hit_set count and period. They will give very bad performance, as over a 3600 period, lots of stuff will look "hot". Not sure if we want to recommend a higher default recency either?

liewegas:
@athanatos what do you think about the 'unknown' section?

liewegas:
Moved it back to known bad, since Mark's tests aren't encouraging.  We can reconsider this when there is more user success and/or we are able to document what workloads and parameters make it a performance win.

fiskn:
Hi Sage,

I meant for the period to be lower. The problem with larger hit set periods is that even very infrequent accesses mark the object as being hit. So with the old default example of 3600, an access 1 once an hour looks the same as 100x a second.

A setting of something like 60 seconds is probably a good starting place for reasonably active workloads. With the recency of 2 it should mean if an object is hit at least twice in succession within 2 minutes then it is a candidate for promotion.

liewegas:
Whoops!  What settings would you recommend here (for period and count)?

fiskn:
I've been doing all my testing with a count of 10 and a period of 60, I think Mark switched to something similar. I can't say they are the best settings but they are certainly a much better default. 

The period is probably the most crucial setting as it needs to roughly match your top N% of objects, all getting accessed about once in that period.  Too small a period and only very very hot stuff gets promoted. Too long a period and too many objects appear to be hot.

PR #9072, opened 2016-05-11
osd: fix crush location update on startup

liewegas:


tchaikov:
sorry for the latency, looks good.

PR #9074, opened 2016-05-11
cmake: build man pages

tchaikov:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,6 +34,15 @@ if(WITH_CCACHE)
   endif(CCACHE_FOUND)
 endif(WITH_CCACHE)
 
+option(WITH_MANPAGE "Build man pages." ON)
+if(WITH_MANPAGE)
+  find_program(SPHINX_BUILD
+    sphinx-build)
+  if(NOT SPHINX_BUILD)
+    message(FATAL_ERROR "Can't find sphinx-build.")
+  endif(NOT SPHINX_BUILD)
+endif(WITH_MANPAGE)
+
 include_directories(
   ${PROJECT_BINARY_DIR}/src/include
   ${OFED_PREFIX}/include
@@ -363,5 +372,7 @@ add_custom_target(check COMMAND ${CMAKE_CTEST_COMMAND})
 
 add_subdirectory(src)
 
-# man pages must be preprocessed, not supported yet
-#add_subdirectory(man)
+add_subdirectory(doc)
+if(WITH_MANPAGE)
+  add_subdirectory(man)
+endif(WITH_MANPAGE)
--- a/man/CMakeLists.txt
+++ b/man/CMakeLists.txt
@@ -1,43 +1,4 @@
-install(FILES
-	ceph-osd.8
-	ceph-mds.8
-	ceph-mon.8
-	ceph-syn.8
-	crushtool.8
-	osdmaptool.8
-	monmaptool.8
-	ceph-conf.8
-	ceph-run.8
-	ceph.8
-	mount.ceph.8
-	ceph-create-keys.8
-	radosgw.8
-	radosgw-admin.8
-	ceph-authtool.8
-	rados.8
-	librados-config.8
-	ceph-clsinfo.8
-	ceph-debugpack.8
-	cephfs.8
-	ceph-dencoder.8
-	ceph-rest-api.8
-	ceph-post-file.8
-  DESTINATION ${CEPH_MAN_DIR}/man8)
-
-if(HAVE_LIBFUSE)
-  install(FILES
-	ceph-fuse.8
-	rbd-fuse.8
-	DESTINATION ${CEPH_MAN_DIR}/man8)
-endif()
-
-if(WITH_RBD)
-  install(FILES
-	ceph-rbdnamer.8
-	rbd-nbd.8
-	rbd-replay-prep.8
-	rbd-replay.8
-	rbdmap.8
-	rbd.8
-	DESTINATION ${CEPH_MAN_DIR}/man8)
+if(WITH_SELINUX)
+  install(FILES ceph_selinux.8
+    DESTINATION ${CEPH_MAN_DIR}/man8)
 endif()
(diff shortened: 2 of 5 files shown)


tchaikov:
@cbodley would you mind taking a look?

man/CMakeLists.txt
```diff
@@ -1,43 +1,60 @@
...
+
+foreach(manpage ${manpage_srcs})
+  string(REGEX MATCH "^[0-9]/" section ${manpage})
+  get_filename_component(cmd ${manpage} NAME_WE)
+  list(APPEND sphinx_input ${CMAKE_SOURCE_DIR}/doc/man/${manpage})
+  list(APPEND sphinx_output ${CMAKE_BINARY_DIR}/doc/man/${cmd}.${section})
+endforeach()
+
+install(FILES ${sphinx_output} ceph_selinux.8
+  DESTINATION ${CEPH_MAN_DIR}/man8)
```
tchaikov: we could use more sophisticated way to determine the section instead hard-wiring it to man8. but since we only *.8 man pages at this moment, let install all of them into "man8" now.

man/CMakeLists.txt
```diff
@@ -1,43 +1,60 @@
...
-	rados.8
-	librados-config.8
-	ceph-clsinfo.8
-	ceph-debugpack.8
-	cephfs.8
-	ceph-dencoder.8
-	ceph-rest-api.8
-	ceph-post-file.8
-  DESTINATION ${CEPH_MAN_DIR}/man8)
+set(manpages_srcs
```
tchaikov: we list sphinx-build as one of the build dependencies, so i put the generated man pages into the build directory instead of in "/man".

also it might be more natural if we have a CMakeLists.txt in into "/doc/man/8", and collect the man pages there. but i am good either way.

CMakeLists.txt
```diff
@@ -364,4 +373,4 @@ add_custom_target(check COMMAND ${CMAKE_CTEST_COMMAND})
 add_subdirectory(src)
 
 # man pages must be preprocessed, not supported yet
-#add_subdirectory(man)
+add_subdirectory(man)
```
cbodley: make this conditional on `WITH_MANPAGE`

man/CMakeLists.txt
cbodley: > not sure we will have a "make dist" for cmake

that's my understanding as well. i don't think we want to involve cpack, either

> have a CMakeLists.txt in into "/doc/man/8", and collect the man pages there

i would prefer it this way too. for one, it would mean that the entries in `manpages_srcs` would be relative to their CMAKE_CURRENT_SOURCE_DIR, rather than having to add the `${CMAKE_SOURCE_DIR}/doc/man/` prefix. it also solves the install to "man8" piece because each CMakeLists.txt could do its own install(). but i'm fine with putting this off until we need it

cbodley:
Thanks for taking this on @tchaikov, you're my hero! The install() part looks correct to me, but in testing 'make install' I'm getting the strangest errors:
```
-- Installing: /home/cbodley/local/share/man/man8
-- Up-to-date: /home/cbodley/local/share/man/man8
-- Up-to-date: /home/cbodley/local/share/man/man8
-- Up-to-date: /home/cbodley/local/share/man/man8
-- Up-to-date: /home/cbodley/local/share/man/man8
-- Up-to-date: /home/cbodley/local/share/man/man8
-- Up-to-date: /home/cbodley/local/share/man/man8
-- Up-to-date: /home/cbodley/local/share/man/man8
-- Installing: /home/cbodley/local/share/man/man8/ceph_selinux.8
CMake Error at man/cmake_install.cmake:36 (file):
  file INSTALL cannot copy file "/home/cbodley/ceph/man/ceph_selinux.8" to
  "/home/cbodley/local/share/man/man8/ceph_selinux.8".
Call Stack (most recent call first):
  cmake_install.cmake:38 (include)
```
The resulting directory looks like filesystem corruption:
```
$ ll ~/local/share/man/man8
ls: cannot access /home/cbodley/local/share/man/man8/ceph-fuse.8: Permission denied
total 0
?????????? ? ? ? ?            ? ceph-fuse.8
```
..but I was able to reproduce when installing to different devices, and @alimaredia reproduced it on his machine as well.

Digging further, I added some debugging to print the `$sphinx_output` variable and saw this:
`/home/cbodley/ceph/build/doc/man/ceph-fuse.8/;/home/cbodley/ceph/build/doc/man/rbd-fuse.8/;/home/cbodley/ceph/build/doc/man/ceph-rbdnamer.8/;/home/cbodley/ceph/build/doc/man/rbd-nbd.8/;/home/cbodley/ceph/build/doc/man/rbd-replay-prep.8/;/home/cbodley/ceph/build/doc/man/rbd-replay.8/;/home/cbodley/ceph/build/doc/man/rbdmap.8/;/home/cbodley/ceph/build/doc/man/rbd.8/`.

So it looks like we need to trim that trailing / off of the `section` variable that results from `string(REGEX MATCH "^[0-9]/" section ${manpage})`. There's probably a better way to do that, but adding `string(REPLACE "/" "" section ${section})` seemed to work for me and the man pages installed correctly.

man/CMakeLists.txt
cbodley: typo on this line: manpages_srcs -> manpage_srcs

cbodley:
@tchaikov I pushed fixes for the issues I mentioned, feel free to cherry-pick them from https://github.com/cbodley/ceph/commits/wip-cmake-manpages

tchaikov:
@cbodley and @alimaredia thanks for testing this PR!

i merged your changes and restructured the CMakeLists.txt so we don't need to hardcode the "/doc/man" path.

man/CMakeLists.txt
```diff
@@ -1,43 +1,4 @@
...
-if(WITH_RBD)
-  install(FILES
-	ceph-rbdnamer.8
-	rbd-nbd.8
-	rbd-replay-prep.8
-	rbd-replay.8
-	rbdmap.8
-	rbd.8
-	DESTINATION ${CEPH_MAN_DIR}/man8)
+if(WITH_SELINUX)
```
tchaikov: WITH_SELINUX is not supported yet in cmake, will leave it for another PR.

cbodley:
the new stuff in doc/ looks good. retested and everything seemed to work :+1:

man/CMakeLists.txt
tchaikov: added in #9133

PR #9075, opened 2016-05-11
osd/ReplicatedBackend: reuse the already computed cost

runsisi:
no need to compute the cost a second time

### DIFF
--- a/src/messages/MOSDPGPush.h
+++ b/src/messages/MOSDPGPush.h
@@ -28,8 +28,11 @@ public:
   spg_t pgid;
   epoch_t map_epoch;
   vector<PushOp> pushes;
+
+private:
   uint64_t cost;
 
+public:
   void compute_cost(CephContext *cct) {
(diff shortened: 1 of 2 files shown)


xiexingguo:
looks good

src/messages/MOSDPGPush.h
```diff
@@ -43,6 +43,10 @@ class MOSDPGPush : public Message {
     return cost;
   }
 
+  void set_cost(uint64_t c) {
```
tchaikov: cost is a public member variable. if we aways access it via its accessors, maybe we can mark it private?

xiexingguo: @tchaikov  Good idea. Withdraw the needs-qa label for now:-)

runsisi:
@tchaikov @xiexingguo done, thank you both~

yuriw:

PR #9077, opened 2016-05-11
osd: instrument map cache hits and misses

liewegas:
In particular, note misses that are below the map cache lower
bound, and the average distance below that lb.


xiexingguo:
lgtm

yuriw:

PR #9078, opened 2016-05-11
mon, osdmap: optimize check and cleanup of pg_temp for new osdmaps

liewegas:


src/osd/OSDMap.cc
```diff
@@ -1175,81 +1175,80 @@ void OSDMap::dedup(const OSDMap *o, OSDMap *n)
...
+      }
+    }
+    if (num_up == 0) {
+      ldout(cct, 10) << __func__ << "  removing pg_temp " << p->first
+		     << " with all down osds" << p->second << dendl;
+      pending_inc->new_pg_temp[p->first].clear();
+      continue;
+    }
+    // redundant pg_temp?
+    if (pending_inc->new_pg_temp.count(p->first) == 0) {
```
xiexingguo: I think it should be 
```
if (pending_inc->new_pg_temp.count(p->first)) 
```
here.

xiexingguo:
Otherwise looks good.

src/osd/OSDMap.cc
liewegas: Hrm.. actually this an dthe other check are both sort of wrong.  Will revise.

liewegas:
@xiexingguo updated, please look again?  Thanks!

src/osd/OSDMap.cc
```diff
@@ -1235,15 +1235,20 @@ void OSDMap::remove_down_temps(CephContext *cct,
...
 	++num_up;
+	break;
+      }
     }
-    if (num_up == 0)
+    if (num_up == 0) {
+      ldout(cct, 10) << __func__ << "  removing pg_temp " << p->first
+		     << " with all down osds" << p->second << dendl;
       pending_inc->new_pg_temp[p->first].clear();
+      continue;
```
tchaikov: nit, could remove `continue` here.

xiexingguo:
looks good to me now, sir:-)

yuriw:

PR #8963, opened 2016-05-06
msg/async: Separated nonfast message and fast dispatch message thread

yuyuyu101:

PR #9086, opened 2016-05-12
msg/async/AsyncConnection: create writable event for in progress connection

yuyuyu101:
Previously we use a tricky with ceph msgr protocol, if initiator side got
in progress connection state, it will wait until read event. Because if
tcp session built successfully server side will send the banner firstly
and initiator side will get read event. Otherwise, if connection failed,
read event also be called.

But actually man(2)[http://man7.org/linux/man-pages/man2/connect.2.html]
specify if we want to get notification whether connection built, we need
to listen writable event. It means when connection built, send buffer
is ready to be written.

This patch follow the strict nonblock connect process. Not fully sure fix
http://tracker.ceph.com/issues/15849

### DIFF
--- a/src/msg/async/AsyncConnection.cc
+++ b/src/msg/async/AsyncConnection.cc
@@ -1046,9 +1046,12 @@ ssize_t AsyncConnection::_process_connection()
           ldout(async_msgr->cct, 1) << __func__ << " reconnect failed " << dendl;
           goto fail;
         } else if (r > 0) {
+          ldout(async_msgr->cct, 10) << __func__ << " nonblock connect inprogress" << dendl;
+          center->create_file_event(sd, EVENT_WRITABLE, read_handler);
           break;
         }
 
+        center->delete_file_event(sd, EVENT_WRITABLE);
         state = STATE_CONNECTING_WAIT_BANNER;
         break;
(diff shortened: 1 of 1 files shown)


yuyuyu101:
pass http://pulpito.ceph.com/haomai-2016-05-12_02:00:37-rados-wip-15849-distro-basic-smithi http://pulpito.ceph.com/haomai-2016-05-12_04:06:22-rados:thrash-wip-15849-distro-basic-smithi

yuyuyu101:
...ah, hang on ceph_test_msgr

yuyuyu101:
retest this please

liewegas:
this looks good on the cern cluster.

tchaikov:
tested at 

- http://pulpito.ceph.com/kchai-2016-06-15_04:31:07-rados-wip-kefu-testing---basic-mira/
- http://pulpito.ceph.com/kchai-2016-06-15_10:14:46-rados-wip-kefu-testing---basic-mira/
- http://pulpito.ceph.com/kchai-2016-06-17_02:37:57-rados-wip-kefu-testing---basic-mira/

yuyuyu101:
same as #8963

PR #9087, opened 2016-05-12
rados: make all rados op w/ FULL_TRY flag.

majianpeng:
Like rados command 'rm' --force-full.

### DIFF
--- a/src/osd/ReplicatedPG.cc
+++ b/src/osd/ReplicatedPG.cc
@@ -1662,14 +1662,13 @@ void ReplicatedPG::do_op(OpRequestRef& op)
   // op to succeed despite being full).  The except is FULL_FORCE|FULL_TRY ops,
   // which there is no reason to discard because they bypass all full
   // checks anyway.
-  // FIXME: we exclude mds writes for now.
-  if (!(m->get_source().is_mds() || m->has_flag(CEPH_OSD_FLAG_FULL_FORCE|CEPH_OSD_FLAG_FULL_TRY)) &&
+  if (!(m->has_flag(CEPH_OSD_FLAG_FULL_FORCE|CEPH_OSD_FLAG_FULL_TRY)) &&
       info.history.last_epoch_marked_full > m->get_map_epoch()) {
     dout(10) << __func__ << " discarding op sent before full " << m << " "
 	     << *m << dendl;
     return;
   }
-  if (!m->get_source().is_mds() && osd->check_failsafe_full()) {
+  if (osd->check_failsafe_full()) {
     dout(10) << __func__ << " fail-safe full check failed, dropping request"
 	     << dendl;
     return;
@@ -6621,9 +6620,8 @@ int ReplicatedPG::prepare_transaction(OpContext *ctx)
       (pool.info.has_flag(pg_pool_t::FLAG_FULL) ||
        get_osdmap()->test_flag(CEPH_OSDMAP_FULL))) {
     MOSDOp *m = static_cast<MOSDOp*>(ctx->op->get_req());
-    if (ctx->reqid.name.is_mds() ||   // FIXME: ignore MDS for now
-	m->has_flag(CEPH_OSD_FLAG_FULL_FORCE)) {
-      dout(20) << __func__ << " full, but proceeding due to FULL_FORCE or MDS"
+    if (m->has_flag(CEPH_OSD_FLAG_FULL_FORCE)) {
+      dout(20) << __func__ << " full, but proceeding due to FULL_FORCE"
 	       << dendl;
     } else if (m->has_flag(CEPH_OSD_FLAG_FULL_TRY)) {
       // they tried, they failed.
--- a/src/osd/ReplicatedPG.cc
+++ b/src/osd/ReplicatedPG.cc
@@ -1659,11 +1659,11 @@ void ReplicatedPG::do_op(OpRequestRef& op)
   // discard due to cluster full transition?  (we discard any op that
   // originates before the cluster or pool is marked full; the client
   // will resend after the full flag is removed or if they expect the
-  // op to succeed despite being full).  The except is FULL_FORCE ops,
+  // op to succeed despite being full).  The except is FULL_FORCE|FULL_TRY ops,
   // which there is no reason to discard because they bypass all full
   // checks anyway.
   // FIXME: we exclude mds writes for now.
-  if (!(m->get_source().is_mds() || m->has_flag(CEPH_OSD_FLAG_FULL_FORCE)) &&
+  if (!(m->get_source().is_mds() || m->has_flag(CEPH_OSD_FLAG_FULL_FORCE|CEPH_OSD_FLAG_FULL_TRY)) &&
       info.history.last_epoch_marked_full > m->get_map_epoch()) {
     dout(10) << __func__ << " discarding op sent before full " << m << " "
 	     << *m << dendl;
--- a/src/osdc/Objecter.cc
+++ b/src/osdc/Objecter.cc
@@ -2310,22 +2310,32 @@ void Objecter::_op_submit(Op *op, shunique_lock& sul, ceph_tid_t *ptid)
 
   assert(op->target.flags & (CEPH_OSD_FLAG_READ|CEPH_OSD_FLAG_WRITE));
 
+  /*
+   * For FULL_FORCE/FULL_TRY, it can set directly on OP or set by Objcter
+   * like osdmap_full_force & osdmap_full_try
+   */
+  int flags = op->target.flags;
+  if (osdmap_full_force)
+    flags |= CEPH_OSD_FLAG_FULL_FORCE;
+  if (osdmap_full_try)
+    flags |= CEPH_OSD_FLAG_FULL_TRY;
+
   bool need_send = false;
 
-  if ((op->target.flags & CEPH_OSD_FLAG_WRITE) &&
+  if ((flags & CEPH_OSD_FLAG_WRITE) &&
       osdmap->test_flag(CEPH_OSDMAP_PAUSEWR)) {
     ldout(cct, 10) << " paused modify " << op << " tid " << op->tid
 		   << dendl;
     op->target.paused = true;
     _maybe_request_map();
-  } else if ((op->target.flags & CEPH_OSD_FLAG_READ) &&
+  } else if ((flags & CEPH_OSD_FLAG_READ) &&
 	     osdmap->test_flag(CEPH_OSDMAP_PAUSERD)) {
     ldout(cct, 10) << " paused read " << op << " tid " << op->tid
 		   << dendl;
     op->target.paused = true;
     _maybe_request_map();
-  } else if ((op->target.flags & CEPH_OSD_FLAG_WRITE) &&
-	     !(op->target.flags & (CEPH_OSD_FLAG_FULL_TRY |
+  } else if ((flags & CEPH_OSD_FLAG_WRITE) &&
+	     !(flags & (CEPH_OSD_FLAG_FULL_TRY |
 				   CEPH_OSD_FLAG_FULL_FORCE)) &&
 	     (_osdmap_full_flag() ||
 	      _osdmap_pool_full(op->target.base_oloc.pool))) {
--- a/src/osdc/Objecter.cc
+++ b/src/osdc/Objecter.cc
@@ -2599,7 +2599,7 @@ bool Objecter::_osdmap_has_pool_full() const
 
 bool Objecter::_osdmap_pool_full(const pg_pool_t &p) const
 {
-  return p.has_flag(pg_pool_t::FLAG_FULL) && honor_osdmap_full;
+  return p.has_flag(pg_pool_t::FLAG_FULL) && !osdmap_full_force;
 }
 
 /**
@@ -2607,8 +2607,8 @@ bool Objecter::_osdmap_pool_full(const pg_pool_t &p) const
  */
 bool Objecter::_osdmap_full_flag() const
 {
-  // Ignore the FULL flag if the caller has honor_osdmap_full
-  return osdmap->test_flag(CEPH_OSDMAP_FULL) && honor_osdmap_full;
+  // Ignore the FULL flag if the caller set osdmap_full_force
+  return osdmap->test_flag(CEPH_OSDMAP_FULL) && !osdmap_full_force;
 }
 
 void Objecter::update_pool_full_map(map<int64_t, bool>& pool_full_map)
@@ -3024,7 +3024,7 @@ MOSDOp *Objecter::_prepare_osd_op(Op *op)
   if (op->onack)
     flags |= CEPH_OSD_FLAG_ACK;
 
-  if (!honor_osdmap_full)
+  if (osdmap_full_force)
     flags |= CEPH_OSD_FLAG_FULL_FORCE;
 
   op->target.paused = false;
--- a/src/osdc/Objecter.cc
+++ b/src/osdc/Objecter.cc
@@ -3026,6 +3026,8 @@ MOSDOp *Objecter::_prepare_osd_op(Op *op)
 
   if (osdmap_full_force)
     flags |= CEPH_OSD_FLAG_FULL_FORCE;
+  if (osdmap_full_try)
+    flags |= CEPH_OSD_FLAG_FULL_TRY;
 
   op->target.paused = false;
   op->stamp = ceph::mono_clock::now();
--- a/src/tools/rados/rados.cc
+++ b/src/tools/rados/rados.cc
@@ -1734,6 +1734,9 @@ static int rados_tool_common(const std::map < std::string, std::string > &opts,
      goto out;
   }
 
+  // make all osd op of rados cli w/ CEPH_OSD_FLAG_FULL_TRY flag.
+  rados.set_block_on_osdfull(false);
+
   if (create_pool && !pool_name) {
     cerr << "--create-pool requested but pool_name was not specified!" << std::endl;
     usage_exit();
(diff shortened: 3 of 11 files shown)


src/librados/IoCtxImpl.cc
```diff
@@ -1005,6 +1005,29 @@ int librados::IoCtxImpl::aio_remove(const object_t &oid, AioCompletionImpl *c)
   return 0;
 }
 
+int librados::IoCtxImpl::aio_remove(const object_t &oid, AioCompletionImpl *c, int flags)
```
yangdongsheng: Then we can change the original aio_remove to
```
int librados::IoCtxImpl::aio_remove(const object_t &oid, AioCompletionImpl *c)
{
    return aio_remove(oid, c, 0);
}
```

src/tools/rados/rados.cc
```diff
@@ -69,8 +69,8 @@ void usage(ostream& out)
 "   cppool <pool-name> <dest-pool>   copy content of a pool\n"
 "   rmpool <pool-name> [<pool-name> --yes-i-really-really-mean-it]\n"
 "                                    remove pool <pool-name>'\n"
-"   purge <pool-name> --yes-i-really-really-mean-it\n"
-"                                    remove all objects from pool <pool-name> without removing it\n"
+"   purge <pool-name> --yes-i-really-really-mean-it [--force-full]\n"
+"                                    [force no mater full or not]remove all objects from pool <pool-name> without removing it\n"
```
dzafman: s/mater/matter/

src/tools/rados/rados.cc
```diff
@@ -101,8 +101,8 @@ void usage(ostream& out)
 "                                    default is 16 concurrent IOs and 4 MB ops\n"
 "                                    default is to clean up after write benchmark\n"
 "                                    default run-name is 'benchmark_last_metadata'\n"
-"   cleanup [--run-name run_name] [--prefix prefix]\n"
-"                                    clean up a previous benchmark operation\n"
+"   cleanup [--run-name run_name] [--prefix prefix] [--force-full]\n"
+"                                    [force no mater full or not]clean up a previous benchmark operation\n"
```
dzafman: s/mater/matter/

src/tools/rados/rados.cc
```diff
@@ -903,7 +904,10 @@ class RadosBencher : public ObjBencher {
   }
 
   int aio_remove(const std::string& oid, int slot) {
-    return io_ctx.aio_remove(oid, completions[slot]);
+    if (forcefull)
+      return io_ctx.aio_remove(oid, completions[slot], CEPH_OSD_FLAG_FULL_FORCE);
+    else
+      return io_ctx.aio_remove(oid, completions[slot]);
```
dzafman: sync_remove() needs this equivalent change too calling the flags version of IoCtx::remove().  This removes the "benchmark_last_metadata" object.

dzafman: Maybe forceflag state should set an int flags and then just have one call to aio_remove:

<pre>
return io_ctx::aoi_remove(oid, completes[slot], flags);
</pre>

majianpeng:
@dzafman . update.

dzafman:
Either squash the 2 commits, or change the description to reflect the rados command commit.

dzafman:
I tried your branch and it didn't work because of an existing bug in the handling of CEPH_OSD_FLAG_FULL_FORCE flag.

Please cherry-pick the following commit into your branch: c5a4c6c4149e7d4fd38aa2a6f20c591af71cc5c3

majianpeng:
@dzafman,  I changed ceph.conf:  osd failsafe full ratio = .30; mon osd full ratio = .10, The first two patch work. But if mon osd full ratio = osd failsafe full ratio. It can't work. So i add e2414ff.

In MOSDOp *Objecter::_prepare_osd_op(Op *op) 
>> if (!honor_osdmap_full)
 >>   flags |= CEPH_OSD_FLAG_FULL_FORCE;

But only  mds call  unset_honor_osdmap_full(). So for FULL_FORCE, we suppose use know the result of this.It should do whether osd has full flag or osd is failsafe.

majianpeng:
@liewegas . This mainly for how to handle FULL_TRY/FULL_FORCE when osd in full or failsafe statue. Hope your comments. Thanks!

src/librados/IoCtxImpl.h
```diff
@@ -206,6 +206,7 @@ struct librados::IoCtxImpl {
   int aio_writesame(const object_t &oid, AioCompletionImpl *c,
 		    const bufferlist& bl, size_t write_len, uint64_t off);
   int aio_remove(const object_t &oid, AioCompletionImpl *c);
+  int aio_remove(const object_t &oid, AioCompletionImpl *c, int flags);
```
liewegas: @jdurgin adding a new overload breaks the ABI, right?  this should be aio_remove2?

Should we just use aio_operate from the cli instead?

src/osd/ReplicatedPG.cc
```diff
@@ -6631,8 +6630,7 @@ int ReplicatedPG::prepare_transaction(OpContext *ctx)
       (pool.info.has_flag(pg_pool_t::FLAG_FULL) ||
        get_osdmap()->test_flag(CEPH_OSDMAP_FULL))) {
     MOSDOp *m = static_cast<MOSDOp*>(ctx->op->get_req());
-    if (ctx->reqid.name.is_mds() ||   // FIXME: ignore MDS for now
-	m->has_flag(CEPH_OSD_FLAG_FULL_FORCE)) {
+    if (m->has_flag(CEPH_OSD_FLAG_FULL_FORCE)) {
       dout(20) << __func__ << " full, but proceeding due to FULL_FORCE or MDS"
```
liewegas: remove ' or MDS'

liewegas:
The problem here is that the MDS can/will now bypass the failsafe and break an OSD (by overfilling it such that it can't start).  I wonder if the right thing to do is to make the MDS set FULL_TRY instead of FULL_FORCE?

@gregsfortytwo @jcsp thoughts?  It's probably time we had a real plan for ENOSPC here.

gregsfortytwo:
@liewegas unless I'm misremembering, the MDS at this point will only submit delete operations when the osdmap full flag is set; everything else gets an ENOSPC. I guess that still leaves a race between when the OSD actually fills up and when the MDS gets a map saying stuff is full, but that's what the thresholds are supposed to be for. :/

Or else I'm misremembering how much of that is done as I believe @jcsp did most of it.

src/librados/IoCtxImpl.h
jdurgin: yeah, I'd rather use aio_operate() from the cli instead of adding a new method to librados.

liewegas:
@majianpeng So, for the rados CLI issue, I *think* what you actaully want is FULL_TRY.  That will make the OSD prepare the transaction and only allow it if it doesn't increase utilization.. which is enough for the CLI to allow deletes (utilization will clearly go down).

src/tools/rados/rados.cc
```diff
@@ -69,8 +69,8 @@ void usage(ostream& out)
 "   cppool <pool-name> <dest-pool>   copy content of a pool\n"
 "   rmpool <pool-name> [<pool-name> --yes-i-really-really-mean-it]\n"
 "                                    remove pool <pool-name>'\n"
-"   purge <pool-name> --yes-i-really-really-mean-it\n"
-"                                    remove all objects from pool <pool-name> without removing it\n"
+"   purge <pool-name> --yes-i-really-really-mean-it [--force-full]\n"
+"                                    [force no matter full or not]remove all objects from pool <pool-name> without removing it\n"
```
liewegas: please drop the [...] part

src/tools/rados/rados.cc
```diff
@@ -101,8 +101,8 @@ void usage(ostream& out)
 "                                    default is 16 concurrent IOs and 4 MB ops\n"
 "                                    default is to clean up after write benchmark\n"
 "                                    default run-name is 'benchmark_last_metadata'\n"
-"   cleanup [--run-name run_name] [--prefix prefix]\n"
-"                                    clean up a previous benchmark operation\n"
+"   cleanup [--run-name run_name] [--prefix prefix] [--force-full]\n"
+"                                    [force no matter full or not]clean up a previous benchmark operation\n"
```
liewegas: please drop the [..] part

src/tools/rados/rados.cc
```diff
@@ -903,7 +904,8 @@ class RadosBencher : public ObjBencher {
   }
 
   int aio_remove(const std::string& oid, int slot) {
-    return io_ctx.aio_remove(oid, completions[slot]);
+    int flags = forcefull ? CEPH_OSD_FLAG_FULL_FORCE : 0;
```
liewegas: I think this should be FULL_TRY, not FULL_FORCE.

dzafman:
@jcsp @gregsfortytwo @liewegas Let's verify that MDS uses the FULL_TRY flag in all the correct place before turning off the is_mds() checking?

src/osd/ReplicatedPG.cc
```diff
@@ -1673,17 +1673,17 @@ void ReplicatedPG::do_op(OpRequestRef& op)
...
-  // FIXME: we exclude mds writes for now.
-  if (!(m->get_source().is_mds() || m->has_flag(CEPH_OSD_FLAG_FULL_FORCE)) &&
+  if (!m->has_flag(CEPH_OSD_FLAG_FULL_FORCE|CEPH_OSD_FLAG_FULL_TRY) &&
       info.history.last_epoch_marked_full > m->get_map_epoch()) {
     dout(10) << __func__ << " discarding op sent before full " << m << " "
 	     << *m << dendl;
     return;
   }
-  if (!m->get_source().is_mds() && osd->check_failsafe_full()) {
+  if (!m->has_flag(CEPH_OSD_FLAG_FULL_FORCE|CEPH_OSD_FLAG_FULL_FORCE) &&
```
dzafman: This needs to be CEPH_OSD_FLAG_FULL_FORCE|CEPH_OSD_FLAG_FULL_TRY.  I wish this code looked like this:

<pre>
if (!m->has_flag(CEPH_OSD_FLAG_FULL_FORCE|CEPH_OSD_FLAG_FULL_TRY) {
    if (info.history.last_epoch_marked_full > m->get_map_epoch()) {
        dout(...
        return;
    }
    if (osd->check_failsafe_full()) {
        dout(...
        return;
    }
}
</pre>

liewegas:
@majianpeng ok, having thought at this some more, I think:

1. first patch: don't add a new aio_remove op.  Users can use the aio_operate instead and set the flag that way.
2. for the rados cli, I think we can just uncondutionally set FULL_TRY on the rados op.  It just means the op will get an error instead of blocking, which I think is what we always want from the cli.
3. The failsafe check should remove the mds special case--it's dangerous!
4. the last two patches look correct to me as is.

Thank you!

gregsfortytwo:
@dzafman, we don't use FULL_TRY at all in the MDS. I was just making a statement about when it is willing to submit operations to the Objecter. :) Just created a feature ticket, though: http://tracker.ceph.com/issues/15942

dzafman:
@liewegas 

> 2. for the rados cli, I think we can just uncondutionally set FULL_TRY on the rados op. It just means the op will get an error instead of blocking, which I think is what we always want from the cli.

Yes, maybe the cli should set FULL_TRY on all operations.  The only reason not to do that is for scripting that wants to block.  For object removal --force-full sets FULL_FORCE so that space can be freed up.

Proposal:
Default rados cli operations sets FULL_TRY (e.g. "put" and "import" can fail with ENOSPC)
NEW --block-full does not set FULL_TRY so that scripts can hang on out of space
--force-full Object remove operations would pass FULL_FORCE if this option specified

majianpeng:
To make all rados op w/ FULL_TRY, the simplest method is add new api like objecter::set_honor_osdmap_full. I think for all client this simple. So  I add Objecter::set_osdmap_full_op_try(). To be frankly, the name of api is ugly. Hope suggestion.

In fact, at first, i wonder change honor_osdmap_full flag from FULL_FORCE to FULL_TRY. I think FULL_TRY more general and FULL_FORCE maybe set for  individual op. Hope suggestions.

Thanks!

liewegas:
Hmm, yeah.  Let's

1. Do a cleanup patch that renames honor_osdmap_full to osdmap_full_force. This will flip the bool around and change the Objecter interface to set_osdmap_full_force(bool f)
2. Add a the set_osdmap_full_try(bool t).
3. Call the librados API set_block_on_full(bool block) where it defaults to true, and false means we set FULL_TRY.  Ths has to be a Rados method, though, not IoCtx, or else it implies it's only affects that IoCtx.  (And that would require changing all of the IoCtx.cc operate() etc calls to pass the flag explicitly instead of setting it on Objecter.  That might be nice, but it's a lot more work and i don't think we need it. Yet.)

liewegas:
Also, I think this was unclear before, but we should drop 'osd/ReplicatedPG: Handle op which w/ FULL_FORCE if osd is failsafe. …'.  It should no longer be possible to bypass the failsafe check at all.

majianpeng:
@liewegas . update

src/osd/ReplicatedPG.cc
```diff
@@ -1654,17 +1654,18 @@ void ReplicatedPG::do_op(OpRequestRef& op)
...
-  if (!(m->get_source().is_mds() || m->has_flag(CEPH_OSD_FLAG_FULL_FORCE)) &&
+  if (!(m->get_source().is_mds() || m->has_flag(CEPH_OSD_FLAG_FULL_FORCE|CEPH_OSD_FLAG_FULL_TRY)) &&
       info.history.last_epoch_marked_full > m->get_map_epoch()) {
     dout(10) << __func__ << " discarding op sent before full " << m << " "
 	     << *m << dendl;
     return;
   }
-  if (!m->get_source().is_mds() && osd->check_failsafe_full()) {
+  if (!(m->get_source().is_mds() || m->has_flag(CEPH_OSD_FLAG_FULL_TRY)) &&
+      osd->check_failsafe_full()) {
```
liewegas: this line shouldn't change--we shouldn't skip the failsafe check.

src/osd/ReplicatedPG.cc
```diff
@@ -1657,14 +1657,13 @@ void ReplicatedPG::do_op(OpRequestRef& op)
...
-  if (!(m->get_source().is_mds() || m->has_flag(CEPH_OSD_FLAG_FULL_FORCE|CEPH_OSD_FLAG_FULL_TRY)) &&
+  if (!(m->has_flag(CEPH_OSD_FLAG_FULL_FORCE|CEPH_OSD_FLAG_FULL_TRY)) &&
       info.history.last_epoch_marked_full > m->get_map_epoch()) {
     dout(10) << __func__ << " discarding op sent before full " << m << " "
 	     << *m << dendl;
     return;
   }
-  if (!(m->get_source().is_mds() || m->has_flag(CEPH_OSD_FLAG_FULL_TRY)) &&
+  if (!(m->has_flag(CEPH_OSD_FLAG_FULL_TRY)) &&
       osd->check_failsafe_full()) {
```
liewegas: so that here we should end up with just if (osd->check_failsafe_full()) { ... }

src/osdc/Objecter.cc
```diff
@@ -2310,22 +2310,32 @@ void Objecter::_op_submit(Op *op, shunique_lock& sul, ceph_tid_t *ptid)
...
+  /*
+   * For FULL_FORCE/FULL_TRY, it can set directly on OP or set by Objcter
+   * like osdmap_full_force & osdmap_full_try
+   */
+  int flags = op->target.flags;
+  if (osdmap_full_force)
+    flags |= CEPH_OSD_FLAG_FULL_FORCE;
+  if (osdmap_full_try)
+    flags |= CEPH_OSD_FLAG_FULL_TRY;
+
```
liewegas: i think this should just set the bit in op->target.flags, and _prepare_osd_op() can drop the part the sets it when sending.

majianpeng:
@liewegas . Does it need add set_block_on_osdfull for librados c-api?

liewegas:
we probably should!

majianpeng:
@liewegas  add api for librados c-api: rados_set_block_on_osdfull.

src/osd/ReplicatedPG.cc
liewegas: this line still needs to be fixed.  the client should not be able to bypass the failsafe check.

liewegas:
looks good, except for that failsafe check condition (the client should not be able to bypass it with the full_try flag).

jcsp:
This appears to be causing an fs test failure:
http://qa-proxy.ceph.com/teuthology/yuriw-2016-06-07_18:47:13-fs-wip-yuri-testing_JUN_7_2016---basic-smithi/243224/teuthology.log

majianpeng:
@jcsp . I think this because 	a3b4bdd. If osd->check_failsafe_full() return true, for the delete op, it will discard and don't anything. Although the op w/ FULL_FORCE flag.
@liewegas,  i think for FULL_FORCE, it wall bypass check_failsafe_full.

athanatos:
This seemed to cause a crash in sjust@teuthology:/a/yuriw-2016-06-14_09:06:01-rados-wip-yuri-testing_JUN_7_2016-distro-basic-smithi/259449

majianpeng:
For my test, change osd_failsafe_full_ratio = 0.10; mon_osd_full_ratio=0.99; if osd reach the failsafe_full; all ops will discard include rm op. This because sage's suggestion.  At before 2f30a97. mds op can bypass this state.  From the log, i think it don't remove the object and so don't add new object(because osd now in failsafe full state)

athanatos:
@majianpeng I don't understand.  If you need to change the values for the test, do it, and submit a PR for ceph-qa-suite as well.  We can't merge this if it breaks existing tests.

jcsp:
@majianpeng are you still working on this?

dachary:
needs rebase

dachary:
feel free to re-open once you have time to work on it

PR #9088, opened 2016-05-12
qa: dynamic_features.sh: return error only if it failed on alive QEMU

trociny:
Fixes: #15500


dillaman:
lgtm

PR #9090, opened 2016-05-12
hammer: osd: remove all stale osdmaps in handle_osd_map()

tchaikov:
in a large cluster, there are better chances that the OSD fails to trim
the cached osdmap in a timely manner. and sometimes, it is just unable
to keep up with the incoming osdmap if skip_maps, so the osdmap cache
can keep building up to over 250GB in size. in this change

  removed.
  if skip_maps is true. in my test, it happens when the osd only
  receives the osdmap from monitor occasionally because the osd happens
  to be chosen when monitor wants to share a new osdmap with a random
  osd.

(cherry picked from commit 369db9930887d75b498927da9c97733bff4472b6)

Conflicts:
    remove the C++11ism

### DIFF
--- a/src/osd/OSD.cc
+++ b/src/osd/OSD.cc
@@ -5444,6 +5444,7 @@ void OSD::dispatch_session_waiting(Session *session, OSDMapRef osdmap)
   } else {
     register_session_waiting_on_map(session);
   }
+  session->maybe_reset_osdmap();
 }
 
 
@@ -5522,6 +5523,7 @@ void OSD::session_notify_pg_cleared(
   assert(session->session_dispatch_lock.is_locked());
   update_waiting_for_pg(session, osdmap);
   session->waiting_for_pg.erase(pgid);
+  session->maybe_reset_osdmap();
   clear_session_waiting_on_pg(session, pgid);
 }
 
--- a/src/osd/OSD.cc
+++ b/src/osd/OSD.cc
@@ -210,6 +210,7 @@ OSDService::OSDService(OSD *osd) :
   pg_epoch_lock("OSDService::pg_epoch_lock"),
   publish_lock("OSDService::publish_lock"),
   pre_publish_lock("OSDService::pre_publish_lock"),
+  max_oldest_map(0),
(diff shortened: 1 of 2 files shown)


tchaikov:
http://tracker.ceph.com/issues/15193

tchaikov:
should wait until all

* https://github.com/ceph/ceph/pull/8017
* https://github.com/ceph/ceph/pull/8990
* https://github.com/ceph/ceph/pull/9108

are merged.

tchaikov:
retest this please

tchaikov:
the test failed due to env issue
```
FAIL: test/test-ceph-helpers.sh
../test-driver: line 126: test/test-ceph-helpers.sh.trs: No space left on device
../test-driver: line 127: test/test-ceph-helpers.sh.trs: No space left on device
../test-driver: line 128: test/test-ceph-helpers.sh.trs: No space left on device
../test-driver: line 129: test/test-ceph-helpers.sh.trs: No space left on device
make[5]: *** [test/test-ceph-helpers.sh.log] Error 1
make[5]: *** Waiting for unfinished jobs....
```

tchaikov:
retest this please.

vumrao:
test this please.

athanatos:
I think this is causing trouble on hammer (not sure, but this one jumps out) http://tracker.ceph.com/issues/16639

tchaikov:
should remove 5793b13492feecad399451e3a83836722b6e9abc from this changeset.

tchaikov:
but before removing it, i will do a final check.

smithfarm:
@tchaikov Thanks for looking at it. There's at least one other issue in the current hammer-backports branch, which I fixed in https://github.com/ceph/ceph/pull/10227 - once you remove 5793b13 from this PR I can do another round of tests, hopefully with better results.

tchaikov:
changelog

- dropping 5793b13 which leads to http://tracker.ceph.com/issues/16639

smithfarm:

athanatos:
If @tchaikov is ok with it, I'm ok with it.

tchaikov:
yeah, i 👍 for my PR!

PR #9093, opened 2016-05-12
librbd: do not flush rbd cache if journaling is enabled

zhouyuan:
With journaling rbd writes will be persisteted on rbd journal objects.
The journal will be replayed if crash happen. So it's not necessary to
flush rbd_cache in this case. This will improve the flush latency.

This patch adds checking on handling aio_flush: if journaling is
enabled, rbd cache flushing is skipped.
In a system flush(ImageCtx::flush) the cache is flushed even w/ journaling
where we truly do need to flush all IO out to disk.

Fixes: http://tracker.ceph.com/issues/15761

Signed-off-by: Yuan Zhou <yuan.zhou@intel.com>

### DIFF
--- a/src/librbd/AioImageRequest.cc
+++ b/src/librbd/AioImageRequest.cc
@@ -454,23 +454,27 @@ void AioImageFlush::send_request() {
                   !m_image_ctx.journal->is_journal_replaying());
   }
 
-  m_aio_comp->set_request_count(journaling ? 2 : 1);
-
   if (journaling) {
     // in-flight ops are flushed prior to closing the journal
     uint64_t journal_tid = m_image_ctx.journal->append_io_event(
       journal::EventEntry(journal::AioFlushEvent()),
       AioObjectRequests(), 0, 0, false);
 
+    m_aio_comp->set_request_count(2);
+
     C_FlushJournalCommit *ctx = new C_FlushJournalCommit(m_image_ctx,
                                                          m_aio_comp,
                                                          journal_tid);
+    C_AioRequest *req_comp = new C_AioRequest(m_aio_comp);
     m_image_ctx.journal->flush_event(journal_tid, ctx);
     m_aio_comp->associate_journal_event(journal_tid);
+    m_image_ctx.flush_async_operations(req_comp);
+  } else {
+    // flush rbd cache only when journaling is not enabled
+    m_aio_comp->set_request_count(1);
+    C_AioRequest *req_comp = new C_AioRequest(m_aio_comp);
+    m_image_ctx.flush(req_comp);
   }
-
-  C_AioRequest *req_comp = new C_AioRequest(m_aio_comp);
-  m_image_ctx.flush(req_comp);
 
   // track flush op for block writes
   m_aio_comp->start_op(true);


zhouyuan:
@dillaman can you please help to have a look?

dillaman:
Minor: the "Fixes:" keyword should include the full URL to the ticket instead of the number.

zhouyuan:
@dillaman  it's fixed now. Thanks for the review!

zhouyuan:
@ceph-jenkins test this please

src/librbd/ImageCtx.cc
```diff
@@ -849,8 +849,11 @@ struct C_InvalidateCache : public Context {
     // ensure no locks are held when flush is complete
     on_safe = util::create_async_context_callback(*this, on_safe);
 
+    bool journaling = false;
+    journaling = (journal != NULL && !journal->is_journal_replaying());
+
```
trociny: Isn't it better to set this below the assert? I guess journal is safe to access only under the lock?

dillaman: Technically it probably makes sense to move the assertion up -- it will potentially crash either way if called w/o holding the owner lock.

zhouyuan:
@dillaman @trociny updated. Thanks for the look!

zhouyuan:
@ceph-jenkins test this please

src/librbd/ImageCtx.cc
```diff
@@ -850,7 +850,11 @@ struct C_InvalidateCache : public Context {
     on_safe = util::create_async_context_callback(*this, on_safe);
 
     assert(owner_lock.is_locked());
-    if (object_cacher != NULL) {
+
+    bool journaling = false;
+    journaling = (journal != NULL && !journal->is_journal_replaying());
+
+    if (!journaling && object_cacher != NULL) {
```
dillaman: Finally did a non-cursory review.  I think we want this optimization in AioImageFlush (https://github.com/ceph/ceph/blob/master/src/librbd/AioImageRequest.cc#L452).  This is the entry-point for a user-generated flush request.  This flush method is too intertwined right now w/ other expectations that we can't remove it (e.g. if releasing the lock we need to flush out all in-flight IO).

zhouyuan: @dillaman thanks for the review! Do you mean we should move the C_FlushCache stuff to AioImageRequest<flush>::send_request() ? To my understanding this only handles the librbd::aio_flush() API, how about the librbd::flush()?

dillaman: @zhouyuan I just mean move your conditional to ```AioImageFlush``` and have it skip calling ```ImageCtx::flush``` when a user requests a flush op and journaling is enabled.  We need to keep ```ImageCtx::flush``` as is because it handles system flush cases where we truly do need to flush all IO out to disk.

zhouyuan: @dillaman so in a system flush, we do need to flush the cache even if journaling is enabled. is my understanding right here?

dillaman: @zhouyuan Yes -- we only want to optimize away user flushes since those can be safely ignored.

zhouyuan:
@dillaman updated. Thanks for the comments!

src/librbd/AioImageRequest.cc
```diff
@@ -462,8 +462,15 @@ void AioImageFlush::send_request() {
     m_aio_comp->associate_journal_event(journal_tid);
   }
 
-  C_AioRequest *req_comp = new C_AioRequest(cct, m_aio_comp);
-  m_image_ctx.flush(req_comp);
+  if (!journaling && m_image_ctx.object_cacher != NULL)
```
dillaman: I think you'll also need to change line 450: the ```AioCompletion``` doesn't need to wait for the ```ImageCtx::flush``` to complete, the number of pending events should be 1.

zhouyuan: I see, would it be 0 in some case? e.g, `journaling == false && m_image_ctx.object_cacher == NULL`

dillaman: No -- if the cache is disabled, you still might have in-flight writes/discards that you would need to ensure complete safely before finishing the flush.  The ```ImageCtx::flush``` call handles both cases.

zhouyuan: thanks for the education!


Sincerely, Yuan

src/librbd/AioImageRequest.cc
```diff
@@ -462,8 +462,12 @@ void AioImageFlush::send_request() {
     m_aio_comp->associate_journal_event(journal_tid);
   }
 
-  C_AioRequest *req_comp = new C_AioRequest(cct, m_aio_comp);
-  m_image_ctx.flush(req_comp);
+  if (!journaling && m_image_ctx.object_cacher != NULL) {
```
dillaman: I think this just becomes an "} else {" clause to the previous "if (journaling) {" conditional.

dillaman: ... since we still want to flush if the cache is disabled as well.

dillaman:
lgtm

dillaman:
Turns out this is slightly more complicated -- after running this through the test suite it showed lots of seg faults.  I think we want the original code as a starting point, but if journaling is enabled, instead of calling ```ImageCtx::flush```, we would want to call ```ImageCtx::flush_async_operations```.  If journaling is disabled, we would still want to call ```ImageCtx::flush```.  The difference between these two methods is that ```flush_async_operations``` won't flush the cache but will ensure that all preceeding IO is complete before allowing the flush to complete.

zhouyuan:
@dillaman thanks a lot for the check! Let me try to work more on this. Can you please share the logs?

dillaman:
I ran a subset of the rbd test suite (http://pulpito.ceph.com/jdillaman-2016-05-22_12:32:14-rbd-wip-jd-testing---basic-mira/) -- most of the failures were due to environment issues.  The "ceph_test_librbd" failures only occurred when the journaling feature was enabled (RBD_FEATURES=125) and resulted in seg faults.

http://qa-proxy.ceph.com/teuthology/jdillaman-2016-05-22_12:32:14-rbd-wip-jd-testing---basic-mira/207636/teuthology.log

zhouyuan:
@dillaman Hi Jason, I made some changes on the patch, also added some tests for this. Can you please help to check if it looks better?

src/librbd/AioImageRequest.cc
```diff
@@ -447,7 +447,7 @@ void AioImageFlush::send_request() {
                   !m_image_ctx.journal->is_journal_replaying());
   }
 
-  m_aio_comp->set_request_count(cct, journaling ? 2 : 1);
```
dillaman: You will want this original line -- perhaps just move it inside the if and inside the else block so that you can set it to 2 and 1, respectively

src/test/librbd/test_librbd.cc
```diff
@@ -2211,15 +2227,36 @@ TEST_F(TestLibRBD, FlushAio)
...
+    ASSERT_EQ(0, rbd_aio_write(image, offset, TEST_IO_SIZE, test_data, write_comps[i]));
   }
 
   rbd_completion_t flush_comp;
   ASSERT_EQ(0, rbd_aio_create_completion(NULL, NULL, &flush_comp));
   ASSERT_EQ(0, rbd_aio_flush(image, flush_comp));
   ASSERT_EQ(0, rbd_aio_wait_for_complete(flush_comp));
   ASSERT_EQ(1, rbd_aio_is_complete(flush_comp));
+
+  // test with journaling enabled
```
dillaman: You don't need to add subsections within a test for various features.  We run "unittest_librbd" with numerous feature combinations with each rebuild under 'make check':

https://github.com/ceph/ceph/blob/master/src/test/run-rbd-unit-tests.sh

src/test/librbd/test_librbd.cc
```diff
@@ -150,6 +150,22 @@ static int create_image_pp(librbd::RBD &rbd,
   }
 }
 
+static int create_image_full_pp(librbd::RBD &rbd,
```
dillaman: Unneeded since we pick up the features from the ```RBD_FEATURES``` environment variable

src/test/librbd/test_librbd.cc
zhouyuan: @dillaman I see, will remove these tests. Thanks!

dillaman:
@zhouyuan ping

zhouyuan:
@ceph-jenkins test this please

zhouyuan:
Looks like jenkins is still on the old version. local tests passed here.

zhouyuan:
@ceph-jenkins test this please

dillaman:
lgtm

dillaman:
@zhouyuan I apologize -- can you rebase this against master?  It passed a QA test run.

zhouyuan:
@ceph-jenkins test this please

zhouyuan:
@ceph-jenkins test this please

zhouyuan:
@ceph-jenkins test this please

dillaman:
@zhouyuan I feel really badly about this -- can you please rebase for last time (hopefully)?

zhouyuan:
@dillaman thanks! do you happen to have any clues on why unittest_log keeps failing?

dillaman:
@zhouyuan it isn't actually failing -- it's just that the unittest_log has a death test case which results in a false-positive "warning" from Jenkin's attempt to pin-point the real failure.

PR #9094, opened 2016-05-12
osd/PG: exclude removed osd from down set of PriorSet

xiexingguo:
The PriorSet constructor will allow a nonexistent OSD
go into the down set, which is not quite appropriate.

E.g., on receiving a new osdmap, we'll call affected_by_map()
again to check if the given map affects the PriorSet and
will restart Peering process if it is. As affected_by_map()
relies on the down set and the blocked_by set to react
properly, including a perviously nonexistent OSD in the down
set can be rather misleading sometimes, although it may be
rare.

### DIFF
--- a/src/osd/PG.cc
+++ b/src/osd/PG.cc
@@ -7877,19 +7877,20 @@ PG::PriorSet::PriorSet(bool ec_pool,
       int o = interval.acting[i];
       if (o == CRUSH_ITEM_NONE)
 	continue;
-      pg_shard_t so(o, ec_pool ? shard_id_t(i) : shard_id_t::NO_SHARD);
 
-      const osd_info_t *pinfo = 0;
-      if (osdmap.exists(o))
-	pinfo = &osdmap.get_info(o);
+      if (!osdmap.exists(o)) {
+        dout(10) << "build_prior  prior osd." << o
+                 << " no longer exists" << dendl;
+        continue;
+      }
+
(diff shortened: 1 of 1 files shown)


xiexingguo:
retest this please

xiexingguo:
retest this please

xiexingguo:
@liewegas Mind having a look at this? Thanks:-)

liewegas:
Hmm.  I think this might be okay, but I think it is better to treat non-existent as down, because the OSD might get recreated and the osd id re-used.  The only problem with the current code is just that calling it "down" in the pg query results is somewhat confusing, right?

xiexingguo:
In `PG::PriorSet::affected_by_map`:
```c++
  for (set<int>::const_iterator p = down.begin();
       p != down.end();
       ++p) {
    int o = *p;

    if (osdmap->is_up(o)) {
      dout(10) << "affected_by_map osd." << *p << " now up" << dendl;
      return true;
    }

    // did someone in the prior set get lost or destroyed?
    if (!osdmap->exists(o)) {
      dout(10) << "affected_by_map osd." << o << " no longer exists" << dendl;
      return true;
    }
```
I previously thought we might restart the peering here because the non-existent osd still does not exist or it gets replaced by an irrelevant newly created osd and is up in the new osdmap, which is not necessary.

xiexingguo:
retest this please

PR #9095, opened 2016-05-12
osd: make health check more accurate for heartbeat mechanism

xiexingguo:
The original logic will reuse the timestamp which we send a PING to
the specific heartbeat peer to update the last_rx_front[back] field
on receiving the corresponding REPLIES, which later shall be honoured
as the exact timestamps we succeed in getting the corresponding
REPLIES and used to calculate the heartbeat latency and determine
whether the relevant peer is dead.

However this is not accurate enough as there may be a delay between
we receive a PINGREPLY and call `heartbeat_check()`, which is
periodcially scheduled by the `tick()` thread and heartbeat_thread,
to calculate the real heartbeat latency:

1. the `tick()` thread will wake up and call `OSD::heartbeat_check()` at 1s interval,
 but can be blocked and delayed by the big osd_lock significantly.
2. the heartbeat_thread will wake up and call `OSD::heartbeat_check()` at a
 random-choosen interval, which can vary from 0.5s ~6.5S, and 
is blockless, since it requires the heartbeat_lock only.
3. the heartbeat_thread will also be responsible for sending PINGs each time
 it wakes up.

With the prior knowledge above, the following example might be a
little more comprehensible now:
```
1. we send ping at 2016/5/13 14:59:10
2. we get both replies at 2016/5/13 14:59:11 and mark last_rx_* as
   2016/5/13 14:59:10(NB: we always update last_rx_* fields to the
   timestamp we sending pings, instead of receiving replies)
3. heartbeat_thread wakes up at 2016/5/13 14:59:15 and calls
   OSD::heartbeat_check() for the heartbeat latency.
   However, it will figure out that the latency is

     now(2016/5/13 14:59:15) - last_rx_*(2016/5/13 14:59:10) = 5s

   instead of

     2016/5/13 14:59:11 - 2016/5/13 14:59:10 =  1s

   which is obviously wrong.
4. further, heartbeat_thread resends a new ping at 2016/5/13 14:59:15
   meanwhile and the peer is now unable to respond such a call.
5. the tick() thread keeps ticking and the heartbeat latency now may
   appear to be the following series:
     now(2016/5/13 14:59:16) - last_rx_*(2016/5/13 14:59:10) = 6s
     now(2016/5/13 14:59:17) - last_rx_*(2016/5/13 14:59:10) = 7s
     now(2016/5/13 14:59:18) - last_rx_*(2016/5/13 14:59:10) = 8s
     ...
   this is also **not** accurate as we know we send new ping at 2016/5/13 14:59:15
   and shall count down from then.
6. However the last_rx_* fields shall still be consider as the 
   lower bound we can't hear a reply from the front/back side.
```
As both heartbeat interval(osd_heartbeat_interval, default to 6)
and grace(osd_heartbeat_grace, default to 20) are configurable
options, the above case may become fatal if we set a longer
osd_heartbeat_interval or a shorter osd_heartbeat_grace and thus
shall be resolved.
    
Because we keep sending new pings periodically, the old pings
could be deprecated by the newly sent pings, thus we introduces a map
to keep tracking of the ping-history here, which consists of
three elements:
    
    1. "tx_time", worked as map key, indicates the exact timestamp we send
       a ping.
    2. "deadline", indicates we shall receive all replies until then,
       otherwise we consider this peer as "dead".
    3. "rx_refs_remaining", indicates how many replies for the corresponding
       ping are still in-flight. The initial value is 2(the front side and the
       back side).
    
We insert an item into the map above every time we send a ping, and
drop the "rx_refs_remaining" by 1 each time we get a reply for the
tracked ping. If "rx_refs_remaining" drops to 0, we know all the replies
have been successfully collected and we can safely erase the relevant
item from the map as well as older ones if there is any.
    
By comparing the current timestamp with the oldest deadline, we can now
make a much accurate decision about whether the corresponding peer is
healthy or not. And by setting last_rx_* to the timestamp we receiving
the reply, the lower bound that we can't hear a reply from the corresponding
connection is also much clear now.

Other notable changes including:

 **move heartbeat_check() from tick() to tick_without_osd_lock()**--
    The heartbeat_check() logic requires heartbeat_lock only, so it shall
    work without osd_lock in hand. By doing this, we avoid the latency to
    acquire the big osd_lock and detect peer failure much quickly.

 **send heartbeat more randomly**--
    The original scale of randomized interval we sending PINGs is
    too rough, and may raise the ability of PING messages burst for
    large clusters.

### DIFF
--- a/src/osd/OSD.cc
+++ b/src/osd/OSD.cc
@@ -4054,44 +4054,76 @@ void OSD::handle_osd_ping(MOSDPing *m)
     {
       map<int,HeartbeatInfo>::iterator i = heartbeat_peers.find(from);
       if (i != heartbeat_peers.end()) {
-	if (m->get_connection() == i->second.con_back) {
-	  dout(25) << "handle_osd_ping got reply from osd." << from
-		   << " first_tx " << i->second.first_tx
-		   << " last_tx " << i->second.last_tx
-		   << " last_rx_back " << i->second.last_rx_back << " -> " << m->stamp
-		   << " last_rx_front " << i->second.last_rx_front
-		   << dendl;
-	  i->second.last_rx_back = m->stamp;
-	  // if there is no front con, set both stamps.
-	  if (i->second.con_front == NULL)
-	    i->second.last_rx_front = m->stamp;
-	} else if (m->get_connection() == i->second.con_front) {
-	  dout(25) << "handle_osd_ping got reply from osd." << from
-		   << " first_tx " << i->second.first_tx
-		   << " last_tx " << i->second.last_tx
-		   << " last_rx_back " << i->second.last_rx_back
-		   << " last_rx_front " << i->second.last_rx_front << " -> " << m->stamp
-		   << dendl;
-	  i->second.last_rx_front = m->stamp;
-	}
-
-        utime_t cutoff = ceph_clock_now(cct);
-        cutoff -= cct->_conf->osd_heartbeat_grace;
-        if (i->second.is_healthy(cutoff)) {
-          // Cancel false reports
-          if (failure_queue.count(from)) {
-            dout(10) << "handle_osd_ping canceling queued "
-                     << "failure report for osd." << from << dendl;
-            failure_queue.erase(from);
+        auto acked = i->second.pending_pings.find(m->stamp);
+        if (acked != i->second.pending_pings.end()) {
+          utime_t now = ceph_clock_now(cct);
+          int &rx_refs_remaining = acked->second.second;
+          if (m->get_connection() == i->second.con_back) {
+            dout(25) << "handle_osd_ping got reply from osd." << from
+                     << " first_tx " << i->second.first_tx
+                     << " last_tx " << i->second.last_tx
+                     << " last_rx_back " << i->second.last_rx_back << " -> " << now
+                     << " last_rx_front " << i->second.last_rx_front
+                     << dendl;
+            i->second.last_rx_back = now;
+            assert(rx_refs_remaining > 0);
+            --rx_refs_remaining;
+            // if there is no front con, set both stamps.
+            if (i->second.con_front == NULL) {
+              i->second.last_rx_front = now;
+              assert(rx_refs_remaining > 0);
+              --rx_refs_remaining;
+            }
+          } else if (m->get_connection() == i->second.con_front) {
+            dout(25) << "handle_osd_ping got reply from osd." << from
+                     << " first_tx " << i->second.first_tx
+                     << " last_tx " << i->second.last_tx
+                     << " last_rx_back " << i->second.last_rx_back
+                     << " last_rx_front " << i->second.last_rx_front << " -> " << now
+                     << dendl;
+            i->second.last_rx_front = now;
+            assert(rx_refs_remaining > 0);
+            --rx_refs_remaining;
           }
 
-          if (failure_pending.count(from)) {
-            dout(10) << "handle_osd_ping canceling in-flight "
-                     << "failure report for osd." << from << dendl;
-            send_still_alive(curmap->get_epoch(),
-              failure_pending[from].second);
-            failure_pending.erase(from);
+          if (rx_refs_remaining == 0) {
+            // succeeded in getting all replies
+            dout(25) << "handle_osd_ping got all replies from osd." << from
+                     << " , erase pending ping(sent at " << m->stamp << ")"
+                     << " and older pending ping(s)"
+                     << dendl;
+
+            map<utime_t, pair<utime_t, int> > still_pending;
+            ++acked;
+
+            if (acked != i->second.pending_pings.end())
+              still_pending.insert(acked, i->second.pending_pings.end());
+
+            i->second.pending_pings.swap(still_pending);
           }
+
+          if (i->second.is_healthy(now)) {
+            // Cancel false reports
+            if (failure_queue.count(from)) {
+              dout(10) << "handle_osd_ping canceling queued "
+                       << "failure report for osd." << from << dendl;
+              failure_queue.erase(from);
+            }
+
+            if (failure_pending.count(from)) {
+              dout(10) << "handle_osd_ping canceling in-flight "
+                       << "failure report for osd." << from << dendl;
+              send_still_alive(curmap->get_epoch(),
+                failure_pending[from].second);
+              failure_pending.erase(from);
+            }
+          }
+        } else {
+          // old replies, deprecated by newly sent pings.
+          dout(10) << "handle_osd_ping no pending ping(sent at " << m->stamp
+                   << ") is found, treat as covered by newly sent pings "
+                   << "and ignore"
+                   << dendl;
         }
       }
 
@@ -4149,17 +4181,9 @@ void OSD::heartbeat_check()
   }
 
   // check for incoming heartbeats (move me elsewhere?)
-  utime_t cutoff = now;
-  cutoff -= cct->_conf->osd_heartbeat_grace;
   for (map<int,HeartbeatInfo>::iterator p = heartbeat_peers.begin();
        p != heartbeat_peers.end();
        ++p) {
-
-    if (p->second.first_tx == utime_t()) {
-      dout(25) << "heartbeat_check we haven't sent ping to osd." << p->first
-               << "yet, skipping" << dendl;
-      continue;
-    }
 
     dout(25) << "heartbeat_check osd." << p->first
 	     << " first_tx " << p->second.first_tx
@@ -4167,19 +4191,21 @@ void OSD::heartbeat_check()
 	     << " last_rx_back " << p->second.last_rx_back
 	     << " last_rx_front " << p->second.last_rx_front
 	     << dendl;
-    if (p->second.is_unhealthy(cutoff)) {
+    if (p->second.is_unhealthy(now)) {
+      utime_t oldest_deadline = p->second.pending_pings.begin()->second.first;
       if (p->second.last_rx_back == utime_t() ||
 	  p->second.last_rx_front == utime_t()) {
-	derr << "heartbeat_check: no reply from " << p->second.con_front->get_peer_addr().get_sockaddr()
-	     << " osd." << p->first << " ever on either front or back, first ping sent "
-	     << p->second.first_tx << " (cutoff " << cutoff << ")" << dendl;
+        derr << "heartbeat_check: no reply from " << p->second.con_front->get_peer_addr().get_sockaddr()
+             << " osd." << p->first
+             << " ever on either front or back, first ping sent " << p->second.first_tx
+             << " (oldest deadline " << oldest_deadline << ")" << dendl;
 	// fail
 	failure_queue[p->first] = p->second.last_tx;
       } else {
 	derr << "heartbeat_check: no reply from " << p->second.con_front->get_peer_addr().get_sockaddr()
 	     << " osd." << p->first << " since back " << p->second.last_rx_back
 	     << " front " << p->second.last_rx_front
-	     << " (cutoff " << cutoff << ")" << dendl;
+	     << " (oldest deadline " << oldest_deadline << ")" << dendl;
 	// fail
 	failure_queue[p->first] = MIN(p->second.last_rx_back, p->second.last_rx_front);
       }
@@ -4213,6 +4239,8 @@ void OSD::heartbeat()
   dout(5) << "heartbeat: " << service.get_osd_stat() << dendl;
 
   utime_t now = ceph_clock_now(cct);
+  utime_t deadline = now;
+  deadline += cct->_conf->osd_heartbeat_grace;
 
   // send heartbeats
   for (map<int,HeartbeatInfo>::iterator i = heartbeat_peers.begin();
@@ -4222,6 +4250,8 @@ void OSD::heartbeat()
     i->second.last_tx = now;
     if (i->second.first_tx == utime_t())
       i->second.first_tx = now;
+    i->second.pending_pings[now] = make_pair(deadline,
+      HeartbeatInfo::HEARTBEAT_MAX_CONN);
     dout(30) << "heartbeat sending ping to osd." << peer << dendl;
     i->second.con_back->send_message(new MOSDPing(monc->get_fsid(),
 					  service.get_osdmap()->get_epoch(),
@@ -4883,13 +4913,12 @@ bool OSD::_is_healthy()
 
   if (is_waiting_for_healthy()) {
     Mutex::Locker l(heartbeat_lock);
-    utime_t cutoff = ceph_clock_now(cct);
-    cutoff -= cct->_conf->osd_heartbeat_grace;
+    utime_t now = ceph_clock_now(cct);
     int num = 0, up = 0;
     for (map<int,HeartbeatInfo>::iterator p = heartbeat_peers.begin();
 	 p != heartbeat_peers.end();
 	 ++p) {
-      if (p->second.is_healthy(cutoff))
+      if (p->second.is_healthy(now))
 	++up;
       ++num;
     }
--- a/src/osd/OSD.h
+++ b/src/osd/OSD.h
@@ -1609,20 +1609,26 @@ private:
     utime_t last_rx_front;  ///< last time we got a ping reply on the front side
     utime_t last_rx_back;   ///< last time we got a ping reply on the back side
     epoch_t epoch;      ///< most recent epoch we wanted this peer
+    // number of connections we send and receive heartbeat messages
+    const static int HEARTBEAT_MAX_CONN = 2;
+    // pending pings which we haven't collected all replies yet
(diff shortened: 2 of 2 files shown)


yuyuyu101:
osd ping message now is fast dispatch and won't acquire osd_lock

xiexingguo:
Exactly. So we shall do the health-check as soon as we receive replies, instead of delaying this kind of job to the tick() or heartbeat_thread, which are woken up in a timer-driven fashion.

xiexingguo:
retest this please

yuyuyu101:
What's exact difference about check hb status between ASAP and tick. Extra flag always increase code complexity

xiexingguo:
It's more accurate and so we don't wrongly send a failure_report by accident and need to cancel it later.
FWIW, but we still want to do the right thing, do we?

yuyuyu101:
Hmm, I mean if we do fast_dispatch, we don't acquire osd_lock. So we will process ping rely right now. Is it will cause wrong failure?

xiexingguo:
We check peers hb status in `void OSD::heartbeat_check()`, which is scheduled by the tick() thread or heartbeat_thread, which as you can see are woken up in a timer-driven fashion(especially the tick() thread itself sometimes can be blocked by the osd_lock).

During `void OSD::handle_osd_ping(MOSDPing *m)`, we update last_rx_back[front] field to the last rx field(which indicates we succeed in sending a PING) only, and the health-status check is left to tick() and heartbeat_thread themselves.

tchaikov:
assuming

- osd_heartbeat_interval = 6.
- osd_heartbeat_grace = 20.

and:

1. We sended a PING to osd.100 at 2016.5.9 20:20:15
1. We sended a PING to osd.100 at 2016.5.9 20:20:21 <<< should resend the PING at osd_heartbeat_interval
1. We received the corresponding PING-REPLIES from osd.100 at 2016.5.9 20:20:20, and last_rx_* are updated with "2016.5.9 20:20:20"
1. We received the corresponding PING-REPLIES from osd.100 at 2016.5.9 20:20:26, and last_rx_* are updated with "2016.5.9 20:20:26" <<<<<
1. We sended a PING to osd.100 at 2016.5.9 20:20:27 <<< should resend the PING at osd_heartbeat_interval
1. more pings and pongs here. and the last PING_REPLY is received at 49, assuming the round trip takes 5 seconds. and the heartbeat thread is not blocked <<<< 
1. The tick() or heartbeat() thread was blocked(by osd_lock or something alike), and was unable to wake up and perform heartbeat_check() until 2016.5.9 20:20:50
1. During heartbeat_check(), we were **not** going to file a failure_report, as we considered ~~"(2016.5.9 20:20:50 - 2016.5.9 20:20:15) = 35"~~ "(2016.5.9 20:20:50 - 2016.5.9 20:20:49) = 1" ~~already exceeded osd_heartbeat_grace(20 by default) while it actually did not(we got both replies in "2016.5.9 20:20:20 - 2016.5.9 20:20:15 = 5" seconds).~~

tchaikov:
@xiexingguo did you spot the problem while testing or by inspecting the code? if the former, maybe you ran into another issue?

xiexingguo:
@tchaikov Just by inspecting and the example is not real(and is wrong, thanks for reminder):-)




> We received the corresponding PING-REPLIES from osd.100 at 2016.5.9 20:20:20, and last_rx_* are updated with "2016.5.9 20:20:20"
> We received the corresponding PING-REPLIES from osd.100 at 2016.5.9 20:20:26, and last_rx_* are updated with "2016.5.9 20:20:26" <<<<<

It seems we always update last_rx_* to the timestamp we send pings, and later we consider current timestamp(now) - last_rx_*  as the hb latency in `void OSD::heartbeat_check()`, which I think is problematic here ...

xiexingguo:
Check 
```c++
void OSD::handle_osd_ping(MOSDPing *m)
                i->second.last_rx_back = m->stamp;
                // if there is no front con, set both stamps.
                if (i->second.con_front == NULL)
                    i->second.last_rx_front = m->stamp;
```
**m->stamp** is the timestamp we send pings

tchaikov:
yeah, we update last_rx_* with the timestamp of the original PING when receiving its PING_REPLY. but i don't get why it's problematic.

xiexingguo:
Because we don't calculate the latency based on the timestamp receiving the replies.. e.g.:
1. we send ping at 2016/5/13 14:59:10
2. we get both replies at 2016/5/13 14:59:11 and mark last_rx_* as   2016/5/13 14:59:10
3. heartbeat_thread wakes up at  2016/5/13 14:59:15 and calls `void OSD::heartbeat_check()` for the hb latency. However, it will figure out the latency is 
```
    now(2016/5/13 14:59:15) - last_rx_*(2016/5/13 14:59:10) = 5s
```
instead of
```
   2016/5/13 14:59:11 - 2016/5/13 14:59:10 =  1s
```

tchaikov:
the osd tick has its owner timer, and SafeTimer uses a dedicated thread for scheduling the callbacks. i don't think the `tick()` can be delayed in real-world in a significant way leading to wrong decision when checking the healthy status of hb peer.

xiexingguo:
Actually the tick() thread require the osd_lock to be held, which can cause significant latency if pg is stuck or something alike, that's why an extra tick_without_osd_lock() thread is introduced.

xiexingguo:
Wait. So maybe we shall move  `void OSD::heartbeat_check()` from tick() to tick_without_osd_lock() too, since it does not actually require osd_lock...

tchaikov:
> Actually the tick() thread requires the osd_lock to be held, which can cause significant latency if pg is stuck or something alike, that's why an extra tick_without_osd_lock() thread is introduced.

gotcha! but, if it takes too long to acquire the `osd_lock`, i'd consider it's a bug or an environment issue.

xiexingguo:
> gotcha! but, if it takes too long to acquire the osd_lock, i'd consider it's a bug or an environment issue.

Well, dear @tchaikov , I think we can leave this aside as I append a change  https://github.com/ceph/ceph/pull/9095/commits/076cb788348e6674345f262ef6b36ae2a1fe402e which shall let us get rid of the osd_lock problem here.

src/osd/OSD.cc
```diff
@@ -4109,7 +4109,7 @@ void OSD::heartbeat_entry()
   while (!heartbeat_stop) {
     heartbeat();
 
-    double wait = .5 + ((float)(rand() % 10)/10.0) * (float)cct->_conf->osd_heartbeat_interval;
+    double wait = .5 + ((float)(rand() % 10000)/10000.0) * (float)cct->_conf->osd_heartbeat_interval;
```
tchaikov: why not just use RAND_MAX ?

src/osd/OSD.cc
```diff
@@ -4070,6 +4070,10 @@ void OSD::handle_osd_ping(MOSDPing *m)
               failure_pending[from].second);
             failure_pending.erase(from);
           }
+
```
tchaikov: i think 076cb788348e6674345f262ef6b36ae2a1fe402e supersedes this commit. what do you think?

src/osd/OSD.cc
xiexingguo: :-)  I was going to use that. Will revise, thanks.

src/osd/OSD.cc
```diff
@@ -9015,6 +9015,9 @@ const char** OSD::get_tracked_conf_keys() const
     "clog_to_graylog_port",
     "host",
     "fsid",
+#ifdef HAVE_LIBFUSE
+    "osd_objectstore_fuse",
```
tchaikov: it is already. probably you want to guard https://github.com/ceph/ceph/pull/9095/commits/ad3be40a66feb43e1a5295387b8dbcc22e30fdf9#diff-fa6c2eba8356ae1442d1bf749beacfdfR9015 with the `#ifdef` instead?

src/mon/OSDMonitor.cc
```diff
@@ -6593,8 +6593,14 @@ bool OSDMonitor::prepare_command_impl(MonOpRequestRef op,
 	    "--yes-i-really-mean-it if you really do.";
       err = -EPERM;
       goto reply;
-    } else if (!osdmap.exists(id) || !osdmap.is_down(id)) {
-      ss << "osd." << id << " is not down or doesn't exist";
+    } else if (!osdmap.exists(id)) {
+      ss << "osd." << id << " doesn't exist";
```
tchaikov: nit, i'd prefer "does not exist".

xiexingguo:
@tchaikov All fixed, please help to check again, thanks.

xiexingguo:
Consider the discussion above and refactor, label as "feature" since it changes the heartbeat check significantly.

xiexingguo:
Wasn't think big enough, close for now:-(

tchaikov:
> Wasn't thinking big enough, close for now:-(

would you kind enough to elaborate a little bit? thanks.

xiexingguo:
@tchaikov  Sorry for the late response.  In the previous version I don't treat the resending logic carefully, which I get fixed now. Also update the commit message correspondingly for a better understanding.

yuyuyu101:
For the lastest commit, I agree it may be not accurate for interval heartbeat timestamp calculate. But what's the unexpected result will be happen? It seemed fine to own a inaccurate interval?

yuyuyu101:
ok, I'm a little got the idea. You think we should use last PING REPLY timestamp instead of PING sent timestamp since we receive PING REPLY indicate peer is alive.

If so, I don't agree with this change. When we receive PING REPLY it maybe delayed becase of network delay or anything else. And it doesn't indicate an accurate alive timestamp for the peer endpoint.

Because of this design, we already reserver the enough grace interval to allow false negative.

Is my understanding right?

xiexingguo:
> ok, I'm a little got the idea. You think we should use last PING REPLY timestamp instead of PING sent timestamp since we receive PING REPLY indicate peer is alive.
> 
> If so, I don't agree with this change. When we receive PING REPLY it maybe delayed becase of network delay or anything else. And it doesn't indicate an accurate alive timestamp for the peer endpoint.

We use the local time to mark the timestamp we sending pings and receiving pongs, and later to calculate the heartbeat latency, which has nothing to do with the peer's local time, and thus it is OK it there is a clock skew between us and the peer.

Let me make this a little more clear:

_**1**_ On sending a ping to a specific peer, we take down three things:
   
**_1.1_** **tx_time**:  the local time we sending this ping request

**_1.2_** **deadline**: calculated by tx_time + osd_heartbeat_grace, by which time we shall consider peer is dead if we do not get all replies(the back and front side) successfully. **Note that the osd_heartbeat_grace is 20 seconds by default, which is big enough to cover the transmission delay as you mentioned above.**

**_1.3_** **rx_refs_remaining**: indicates how many replies for the corresponding ping are still in-flight. The initial value is 2(as we currently only send pings from the front side and the back side).

We make the above three elements a completed item and insert it into the newly introduced "pending_pings" map, which now looks like:
```
tx_time: 2016/5/13 14:59:10 deadline:2016/5/13 14:59:30 rx_refs_remaining:2
tx_time: 2016/5/13 14:59:12 deadline:2016/5/13 14:59:32 rx_refs_remaining:2
tx_time: 2016/5/13 14:59:17 deadline:2016/5/13 14:59:37 rx_refs_remaining:2
tx_time: 2016/5/13 14:59:18 deadline:2016/5/13 14:59:38 rx_refs_remaining:2
```

 
**_2_** the `tick_without_osd_lock()` thread wakes up at exactly 1s interval, and it will call `heartbeat_check()` to determine if all the relevant peers are still healthy. E.g., for the above
peer, it may act as:

**_2.1_**  ```wakes up at: 2016/5/13 14:59:11; get the oldest deadline:  2016/5/13 14:59:30; compare:
now (2016/5/13 14:59:11) < oldest_deadline(2016/5/13 14:59:30); result: peer is healthy```

**_2.2_**  ```wakes up at: 2016/5/13 14:59:12; get the oldest deadline:  2016/5/13 14:59:30; compare:
now (2016/5/13 14:59:12) < oldest_deadline(2016/5/13 14:59:30); result: peer is healthy```

   ```... (the tick goes on, and the peer is dead)```

**_2.21_**  ```wakes up at: 2016/5/13 14:59:31; get the oldest deadline:  2016/5/13 14:59:30; compare:
now (2016/5/13 14:59:31) > oldest_deadline(2016/5/13 14:59:30); result:```  **peer is unhealthy**


**_3_** On successfully receiving a reply, we simply look its corresponding ping request up from the "pending_pings" map. E.g.:

**_3.1_** we get a reply from the back side at `2016/5/13 14:59:11`, and the related ping is sent at
`2016/5/13 14:59:10`. Then we mark the last_rx_back as  `2016/5/13 14:59:11`, and update the "pending_pings" map, which now looks like:
```
tx_time: 2016/5/13 14:59:10 deadline:2016/5/13 14:59:30 rx_refs_remaining:1
tx_time: 2016/5/13 14:59:12 deadline:2016/5/13 14:59:32 rx_refs_remaining:2
tx_time: 2016/5/13 14:59:17 deadline:2016/5/13 14:59:37 rx_refs_remaining:2
tx_time: 2016/5/13 14:59:18 deadline:2016/5/13 14:59:38 rx_refs_remaining:2
```

**_3.2_** we get a reply from the front side at `2016/5/13 14:59:12`, and the related ping is sent at
`2016/5/13 14:59:10`. Then we mark the last_rx_front as  `2016/5/13 14:59:12`, and update the "pending_pings" map, which now looks like:

~~tx_time: 2016/5/13 14:59:10 deadline:2016/5/13 14:59:30 rx_refs_remaining:0~~
```
tx_time: 2016/5/13 14:59:12 deadline:2016/5/13 14:59:32 rx_refs_remaining:2
tx_time: 2016/5/13 14:59:17 deadline:2016/5/13 14:59:37 rx_refs_remaining:2
tx_time: 2016/5/13 14:59:18 deadline:2016/5/13 14:59:38 rx_refs_remaining:2
```

**_3.3_** Further, we get both replies at `2016/5/13 14:59:19` for the ping sent at `2016/5/13 14:59:18` before getting replies for any earlier sent pings. All the older item in the "pending_pings" will be cleared and the map is now empty:
~~tx_time: 2016/5/13 14:59:12 deadline:2016/5/13 14:59:32 rx_refs_remaining:2~~
~~tx_time: 2016/5/13 14:59:17 deadline:2016/5/13 14:59:37 rx_refs_remaining:2~~
~~tx_time: 2016/5/13 14:59:18 deadline:2016/5/13 14:59:38 rx_refs_remaining:2~~

**_3.4_** The origin last_rx_back[front] is not used to calculate heartbeat latency anymore but record the exact time we hear a reply the corresponding side instead. E.g., if the last_rx_back is `2016/5/13 14:59:38`(and last_rx_front is `2016/5/13 14:59:39`), and later we make a decision the peer is dead, we know that the peer is not reachable from `2016/5/13 14:59:38` at the back side, which is also shall be the lower bound of the peer's real death.

xiexingguo:
@tchaikov @yuyuyu101 Passed our private rados:basic test suit. Does these changes make sense and look good?

yuyuyu101:
hmm, I got your idea. My question is that really make sense to users? User may expect the current impl from the configs. If we use last rx time, it will potentially reduce the heartbeat_grace_timeout interval compare to before. 

For example, heartbeat interval is 1 and grace timeout is 2

we send ping to peer at 0s, and got reply at 0.00001s. Then peer dead at 0.99s, we send ping again at 1s. then we won't get reply. We send ping at 2s. From this pr, we will dead at 2.00001s. So the grace timeout won't be exceed and user may expect we have at least 2s network delay(or anything else), but we actually got 1.0001s.

What I mean is this change may not met user expection.

xiexingguo:
> So the grace timeout won't be exceed and user may expect we have at least 2s network delay(or anything else), but we actually got 1.0001s.

That is not true.
For your example, the new **pending_ping** map may looks like:
```

tx_time: 0 deadline: 2 rx_refs_remaining:2
tx_time: 1 deadline: 3 rx_refs_remaining:2
tx_time: 2 deadline: 4 rx_refs_remaining:2
tx_time: 3 deadline: 5 rx_refs_remaining:2

```
So when we get both replies at 0.0001, we do two things:
- we mark the last_rx_time as 0.0001, which indicates the last time we hear from this peer.
- we erase the ping sent at 0 from the **pending_map**, which now looks like:

```

  tx_time: 1 deadline: 3 rx_refs_remaining:2
  tx_time: 2 deadline: 4 rx_refs_remaining:2
  tx_time: 3 deadline: 5 rx_refs_remaining:2

```

As the tick() thread goes on , it will keep checking the current time against the oldest deadline in the **pending_map** each time it wakes up. Therefore if the specific peer is not reachable thereafter( dies at 0.99 as per your assumption),  the tick() thread will catch that at 4( as **now = 4** > **oldest_deadline = 3**).  Thus we have two conclusions as below:
- the peer is dead now(**=4**) and we shall report to monitor.
- the peer has been dead **at most** for 3.9999(**now = 4** - **last_rx_time=0.0001** ), which is the exact and best estimation we can ever make.

xiexingguo:
@yuyuyu101 @tchaikov Ping.

PR #9096, opened 2016-05-12
mon: optimize the get_rank function

songbaisen:
mon: optimize the get_rank function

### DIFF
--- a/src/mon/MonMap.h
+++ b/src/mon/MonMap.h
@@ -125,15 +125,17 @@ class MonMap {
   }
 
   int get_rank(const string& n) {
-    for (unsigned i=0; i<rank_name.size(); i++)
-      if (rank_name[i] == n)
-	return i;
+    vector<string>::iterator iter = std::find(rank_name.begin(), rank_name.end(), n);
+    if(iter != rank_name.end()) {
+      return std::distance(rank_name.begin(), iter);
+    }
     return -1;
   }
   int get_rank(const entity_addr_t& a) {
-    for (unsigned i=0; i<rank_addr.size(); i++)
-      if (rank_addr[i] == a)
-	return i;
+    vector<entity_addr_t>::iterator iter = std::find(rank_addr.begin(), rank_addr.end(), a);
+    if(iter != rank_addr.end()) {
+      return std::distance(rank_addr.begin(), iter);
+    }
     return -1;
   }
   bool get_addr_name(const entity_addr_t& a, string& name) {


jecluis:
Behavior looks to be functionally the same and I can't think of any thing that would suffer from this.

Code is also simpler and easier on the eyes - plus it's more cpp-ish than before.

I don't think however this can be called an optimization. We're dealing with vectors with a really limited number of elements (we'd expect 1 to a less than a dozen).

The original code has a worst-case complexity of O(n) - we're looking at all the elements after all -, if we end up looking for the last element in the vector. The new code performs no only an std::find(), which has a complexity of O(n), but also an std::distance(), which also has a complexity of O(n). So we're not really optimizing the code.

Given the performance impact on this should be low anyway, regardless of whether we loop once or twice - again, expected to be a really small number of elements -, I'm okay with this just for getting rid of the loop and making things prettier. But the commit message should be adjusted to make it clear this is eye candy rather than an optimization.

jecluis:
Oh, I forgot to mention: some of your lines appear to be over 80 characters. Please make sure to hard wrap them at 80.

src/mon/MonMap.h
```diff
@@ -125,15 +125,17 @@ class MonMap {
   }
 
   int get_rank(const string& n) {
-    for (unsigned i=0; i<rank_name.size(); i++)
-      if (rank_name[i] == n)
-	return i;
+    vector<string>::iterator iter = std::find(rank_name.begin(), rank_name.end(), n);
```
tchaikov: could mark this function `const` also.

tchaikov: i'd suggest use `auto`

src/mon/MonMap.h
```diff
@@ -125,15 +125,17 @@ class MonMap {
...
+    if(iter != rank_name.end()) {
```
tchaikov: add a space after `if`

src/mon/MonMap.h
tchaikov: nit, could use a better name than `iter`, for instance, `name` or `found`.

src/mon/MonMap.h
```diff
@@ -125,15 +125,17 @@ class MonMap {
...
+      return std::distance(rank_name.begin(), iter);
+    }
     return -1;
   }
   int get_rank(const entity_addr_t& a) {
-    for (unsigned i=0; i<rank_addr.size(); i++)
-      if (rank_addr[i] == a)
-	return i;
+    vector<entity_addr_t>::iterator iter = std::find(rank_addr.begin(), rank_addr.end(), a);
+    if(iter != rank_addr.end()) {
```
tchaikov: add a space after `if`.

tchaikov:
agree with @jecluis . and i think the original code is also easy for mortal readers.

songbaisen:
@jecluis :smile: Thank you sir!That is true.And I will close this pr.Before this may i ask you two question.The first is that I seen the mon use the paxos like the fast paxos.But it is diff from that.Each proposal in ceph we need wait for all the quoro agree the proposal.If one quoro  do not reply this in lease time , we will restart the election.But the origin fast paxos do not like this .Only more than half quoro agree the proposal then pass it.And do not need to new election.The second question is that after election the leader and quoro keep headbeat in lease time.If one quoro do not answer in the lease time.The leader will restart the election.It means that the front network for mon must be stable.If one quoro  mon network do not stable such as  constantly  up and down or have a lot of packages loss this will result in the whole mon constantly election.

songbaisen:
@tchaikov  :smile:  Thank you for review！.And  I will close this pr after the consult。

PR #9098, opened 2016-05-12
rgw_rados: remove unused ioctx for domain root in create_bucket

theanalyst:
In `create_bucket`, we open an ioctx for domain root pool and never use
it later, removing this.


theanalyst:
I haven't tracked down where in the code we actually end up creating the system obj of bucket name in domain root pool yet, but definitely not here, guessing it is in one of `rgw_put_system_obj` calls

PR #9102, opened 2016-05-12
rgw: don't unregister request if request is not connected to manager

yehudasa:
That means that request is already complete and done with the manager.

### DIFF
--- a/src/rgw/rgw_http_client.cc
+++ b/src/rgw/rgw_http_client.cc
@@ -41,6 +41,7 @@ struct rgw_http_req_data : public RefCountedObject {
     return ret;
   }
 
+
   void finish(int r) {
     Mutex::Locker l(lock);
     ret = r;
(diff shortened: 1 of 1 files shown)


src/rgw/rgw_http_client.cc
```diff
@@ -790,7 +803,14 @@ void *RGWHTTPManager::reqs_thread_entry()
...
   }
 
+
   RWLock::WLocker rl(reqs_lock);
+  for (auto r : unregistered_reqs) {
+    _finish_request(r, -ECANCELED);
+  }
+
+  unregistered_reqs.clear();
+
```
cbodley: ok, i can see how these could leak since the last call to `manage_pending_requests()`. i was worried that we might still leak them after this point, but after further investigation i'm not seeing how that would happen

cbodley:
looks good to me :+1:

looking more closely at the code during review, i did find that the fds from pipe() were never being closed - see https://github.com/ceph/ceph/pull/9122

PR #9106, opened 2016-05-12
rpm: unconditionally set ceph user's primary group to ceph (SUSE) 

smithfarm:
This commit brings the user/group creation into greater semantic alignment
with the Debian packaging, but only for SUSE.

Fixes: http://tracker.ceph.com/issues/15869

### DIFF
--- a/ceph.spec.in
+++ b/ceph.spec.in
@@ -978,8 +978,13 @@ fi
 if ! getent passwd ceph >/dev/null ; then
     CEPH_USER_ID_OPTION=""
     getent passwd $CEPH_USER_ID >/dev/null || CEPH_USER_ID_OPTION="-u $CEPH_USER_ID"
-    useradd ceph $CEPH_USER_ID_OPTION -r -g ceph -s /sbin/nologin -c "Ceph daemons" -d %{_localstatedir}/lib/ceph 2>/dev/null || :
+    useradd ceph $CEPH_USER_ID_OPTION -r -g ceph -s /sbin/nologin 2>/dev/null || :
 fi
+usermod -c "Ceph storage service" \
+        -d %{_localstatedir}/lib/ceph \   
+        -g ceph \
+        -s /sbin/nologin \
+        ceph
 %endif
 exit 0
 


smithfarm:
Note that in Debian there is an unconditional `usermod` - see https://github.com/ceph/ceph/blob/v10.2.0/debian/ceph-common.postinst#L59-L65

smithfarm:
@ktdreyer @branto1

branto1:
The -o flag comes from ed0cd42d by @dalgaaf. I suppose it is intentional to create the user/group with the default system uid/gid no matter if the user already exists?

smithfarm:
> The -o flag comes from ed0cd42d by @dalgaaf. I suppose it is intentional to create the user/group with the default system uid/gid no matter if the user already exists?

Yes, that was my assumption, too, but I don't think it's good for the ceph packaging to create a second "ceph" users with a differing UID. Do you?

smithfarm:
For now I created http://tracker.ceph.com/issues/15876

branto1:
I'm not sure, here. It depends on how the underlying ceph daemons handle the case of multiple `ceph` users/groups. AFAICR, it was ok (well, even better for scenarios like disk-swapping) for us to use the `-o` option when the user changes were created.

Nevertheless, the underlying daemons should be able to handle this case as we might be on a system that created `ceph` user/group before it was a system user/group and once the system user/group was introduced, the system would have two `ceph` user.

smithfarm:
> Nevertheless, the underlying daemons should be able to handle this case as we might be on a system that created ceph user/group before it was a system user/group and once the system user/group was introduced, the system would have two ceph user.

Yes, but only in RH/CentOS/Fedora. The Debian/Ubuntu and SUSE packaging takes care not to create a second ceph user.

branto1:
> Yes, but only in RH/CentOS/Fedora. The Debian/Ubuntu and SUSE packaging takes care not to create a second ceph user.

I would not be that sure about it. What happens if you have your own ceph user, upgrade the machine and the new system package that creates a ceph user with system (static) uid comes up? I doubt it will modify or delete the existing user.

smithfarm:
@branto1 It does modify the existing user to ensure its primary group is "ceph". Here's what happens in Debian/Ubuntu: https://github.com/ceph/ceph/blob/master/debian/ceph-common.postinst#L39-L65

What happens in SUSE is the subject matter of this PR. . .

smithfarm:
Essentially, the daemons cannot assume that the user "ceph" will be a system user.

branto1:
The daemons don't (afaik, they get the UID from packaging), the users might (especially when hot-swapping disks). I guess that was the reason for forcing non-unique users, there.

Anyway, I originally thought that when you add a new system/static UID to the packaging (package setup in fedora/rhel), it will be automatically created after upgrade. I'm not so sure about it any more, though.

smithfarm:
@ktdreyer Pinging in case you'd like to review. If there are no objections, I will merge.

ktdreyer:
:+1: LGTM

smithfarm:
Note this PR introduced a regression (trailing whitespace) which is fixed by #10707

PR #9108, opened 2016-05-12
osd: reset session->osdmap if waiting_for_pg is empty

tchaikov:
we should release the osdmap reference once we are done with it,
otherwise we might need to wait very long to update that reference with
a newer osdmap ref. this appears to be an OSDMap leak: it is held by an
quiet OSD::Session forever.

Fixes: http://tracker.ceph.com/issues/15879

### DIFF
--- a/src/osd/OpRequest.cc
+++ b/src/osd/OpRequest.cc
@@ -82,6 +82,7 @@ void OpRequest::_unregistered() {
   request->clear_data();
   request->clear_payload();
   request->release_message_throttle();
+  request->set_connection(nullptr);
 }
 
 bool OpRequest::check_rmw(int flag) {
--- a/src/msg/Message.h
+++ b/src/msg/Message.h
@@ -446,10 +446,10 @@ public:
     return get_source_inst();
   }
   entity_name_t get_orig_source() const {
-    return get_orig_source_inst().name;
+    return get_source();
   }
   entity_addr_t get_orig_source_addr() const {
-    return get_orig_source_inst().addr;
+    return get_source_addr();
   }
 
   // virtual bits
--- a/src/osd/OSD.cc
+++ b/src/osd/OSD.cc
@@ -5886,6 +5886,7 @@ void OSD::dispatch_session_waiting(Session *session, OSDMapRef osdmap)
   } else {
     register_session_waiting_on_map(session);
   }
+  session->maybe_reset_osdmap();
 }
 
 
@@ -5964,6 +5965,7 @@ void OSD::session_notify_pg_cleared(
   assert(session->session_dispatch_lock.is_locked());
   update_waiting_for_pg(session, osdmap);
   session->waiting_for_pg.erase(pgid);
+  session->maybe_reset_osdmap();
   clear_session_waiting_on_pg(session, pgid);
 }
 
--- a/src/osd/OSD.h
+++ b/src/osd/OSD.h
@@ -1342,8 +1342,11 @@ public:
       session_dispatch_lock("Session::session_dispatch_lock"), 
       last_sent_epoch(0), received_map_epoch(0)
     {}
-
-
+    void maybe_reset_osdmap() {
+      if (waiting_for_pg.empty()) {
+	osdmap.reset();
+      }
+    }
   };
   void update_waiting_for_pg(Session *session, OSDMapRef osdmap);
   void session_notify_pg_create(Session *session, OSDMapRef osdmap, spg_t pgid);
@@ -1442,6 +1445,7 @@ public:
      */
     session->waiting_on_map.clear();
     session->waiting_for_pg.clear();
+    session->osdmap.reset();
   }
   void register_session_waiting_on_pg(Session *session, spg_t pgid) {
     Mutex::Locker l(session_waiting_lock);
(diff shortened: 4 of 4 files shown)


tchaikov:
changelog

- add a revised version of #8828
- update Message: make get_orig_*() alias of get_*() for better performance and readability. we have `MForward` message, maybe we should make make these two method virtual and override them in `MForward`?

src/osd/OpRequest.cc
```diff
@@ -82,6 +82,7 @@ void OpRequest::_unregistered() {
   request->clear_data();
   request->clear_payload();
   request->release_message_throttle();
+  request->set_connection(nullptr);
```
tchaikov: @gregsfortytwo i re-considered your comment at https://github.com/ceph/ceph/pull/8828#issuecomment-216537466 . and we are safe. see http://tracker.ceph.com/issues/13990#note-105 and https://github.com/ceph/ceph/pull/9108/commits/14207f4390eedb6b59048f07d5467e62fbb6610f .

tchaikov:
changelog

- "Fixes: http://tracker.ceph.com/issues/15879" so we have a 1:1 mapping from PR and its tracker ticket.

liewegas:
test this please

tchaikov:

tchaikov:
retest this please

athanatos:
lgtm, added to wip-sam-testing

athanatos:
-1> 2016-05-14 11:04:55.548056 7f477dc0a700 10 log is not dirty
     0> 2016-05-14 11:04:55.548105 7f47895b7700 -1 osd/OSD.cc: In function 'void OSD::dispatch_session_waiting(OSD::Session*, OSDMapRef)' thread 7f47895b7700 time 2016-05-14 11:04:55.544937
osd/OSD.cc: 5864: FAILED assert(session->osdmap == osdmap)

 ceph version 10.2.0-933-g31704a1 (31704a1009ca563a57c81255541113baeb2d167f)
 1: (ceph::__ceph_assert_fail(char const*, char const*, int, char const*)+0x85) [0x7f479288c435]
 2: (OSD::dispatch_session_waiting(OSD::Session*, std::shared_ptr<OSDMap const>)+0x1d7) [0x7f4792202317]
 3: (OSD::wake_pg_waiters(spg_t)+0x16a) [0x7f479225d62a]
 4: (OSD::handle_pg_peering_evt(spg_t, pg_history_t const&, std::map<unsigned int, pg_interval_t, std::less<unsigned int>, std::allocator<std::pair<unsigned int const, pg_interval_t> > >&, unsigned int, std::shared_ptr<PG::CephPeeringEvt>)+0x15e4) [0x7f4792216794]
 5: (OSD::handle_pg_notify(std::shared_ptr<OpRequest>)+0x5e3) [0x7f47922182b3]
 6: (OSD::dispatch_op(std::shared_ptr<OpRequest>)+0xd0) [0x7f4792219430]
 7: (OSD::_dispatch(Message*)+0x22e) [0x7f479221a02e]
 8: (OSD::ms_dispatch(Message*)+0x21c) [0x7f479221a6ec]
 9: (Messenger::ms_deliver_dispatch(Message*)+0x77) [0x7f47929b3287]
 10: (C_handle_dispatch::do_request(int)+0x11) [0x7f47929b39d1]
 11: (EventCenter::process_events(int)+0x8fa) [0x7f4792952a0a]
 12: (Worker::entry()+0x1f0) [0x7f4792932930]
 13: (()+0x7dc5) [0x7f4790597dc5]
 14: (clone()+0x6d) [0x7f478ec2328d]

tchaikov:
changelog

- do not reset osdmap in OSD::session_notify_pg_create() anymore. see 82b0af7

@athanatos fixed, tested using the failed job at http://pulpito.ceph.com/kchai-2016-05-17_00:23:11-rados:thrash-wip-kefu-testing---basic-smithi/ and repushed.

athanatos:
It actually failed a ton of tests, I just linked one.  I'll rerun it tonight.

xiaoxichen:
@tchaikov , seems we hit this bug,but after upgrade to jewel 10.2.2, the problem is not completed solved. Before upgrading (0.94.6), in a sample OSD we have 327K maps which size are 74GB in total. After upgrading and wait for 2+ days, it still has 100K maps ,around 25GB. The map counts in head/meta doesnt decrease for 1+ days.

tchaikov:
@xiaoxichen, some of the maps won't go away after the upgrade. the fix just stops the osdmap from piling up when new maps are received. and if you are able to reproduce with a new osd node, could you please file a ticket on tracker with more details?

xiaoxichen:
@tchaikov , may i know why ? and how to cleanup the previous leak? I suspect with new code, when everyone re-establish new session, the old map will not get any ref then can be removed?

Hmm, the number of maps is decreased from 300K to 100K but stable at 100K. Cannot tell if "reproduced"

tchaikov:
> may i know why ? 

please read the commit message.

> Cannot tell if "reproduced"

if you install a new osd node with 10.2.2, with the default settings, in a stable environment where no osd is flapping, but the number of osdmap is over 1k, then i'd say it's reproduced.

PR #9111, opened 2016-05-12
test/Throttle.cc: change the order in draining the BackoffThrottle during testing

wjwithagen:
Start + delay could be less than now(), and thus could wait_until be called with the timepoint that is in the past.
It is uncertain what different OSes guarantee in this case. 
On FreeBSD is seems that it leads to no (or very little) possibility to switch thread.
And thus throttle.put() with starve, which would lead to no progress.

### DIFF
--- a/src/test/common/Throttle.cc
+++ b/src/test/common/Throttle.cc
@@ -276,7 +276,8 @@ std::pair<double, std::chrono::duration<double> > test_backoff(
   std::condition_variable c;
   uint64_t total = 0;
   std::list<uint64_t> in_queue;
-  bool stop = false;
+  bool putter_stop = false;
+  bool getter_stop = false;
 
   auto wait_time = std::chrono::duration<double>(0);
   uint64_t waits = 0;
@@ -301,7 +302,7 @@ std::pair<double, std::chrono::duration<double> > test_backoff(
     std::uniform_int_distribution<> dis(0, 10);
 
     std::unique_lock<std::mutex> g(l);
-    while (!stop) {
+    while (!getter_stop) {
       g.unlock();
 
       uint64_t to_get = dis(gen);
@@ -317,25 +318,27 @@ std::pair<double, std::chrono::duration<double> > test_backoff(
   };
 
   auto putter = [&]() {
+    bool empty;
     std::unique_lock<std::mutex> g(l);
-    while (!stop) {
-      while (in_queue.empty())
+    while (!putter_stop) {
+      empty = in_queue.empty();
+      if (in_queue.empty()) {
 	c.wait(g);
+      } else {
+	uint64_t c = in_queue.front();
+	total_observed_total += total;
+	total_observations++;
+	in_queue.pop_front();
+	assert(total <= max);
 
-      uint64_t c = in_queue.front();
-
-      total_observed_total += total;
-      total_observations++;
-      in_queue.pop_front();
-      assert(total <= max);
-
-      g.unlock();
-      std::this_thread::sleep_for(
+	g.unlock();
+	std::this_thread::sleep_for(
 	c * std::chrono::duration<double>(put_delay_per_count*putters));
-      g.lock();
+	g.lock();
 
-      total -= c;
-      throttle.put(c);
+	total -= c;
+	throttle.put(c);
+      }
     }
   };
 
@@ -348,12 +351,24 @@ std::pair<double, std::chrono::duration<double> > test_backoff(
   std::this_thread::sleep_for(std::chrono::duration<double>(5));
   {
     std::unique_lock<std::mutex> g(l);
-    stop = true;
   }
+
+  // Terminate all threads we've forked
+  // First stop the getters filling the queue
+  getter_stop = true;
+  c.notify_all();
   for (auto &&i: gts) i.join();
   gts.clear();
+  cout << __func__<< ":" << __LINE__ << " Switch stop, getters collected." << "\n";
+
+  // Then all putters need to be stopped
+  // Add a bit of waiting to let the threads stabalize
+  putter_stop = true;
+  std::this_thread::sleep_for (std::chrono::milliseconds(500));
+  c.notify_all();
   for (auto &&i: pts) i.join();
   pts.clear();
+  cout << __func__<< ":" << __LINE__ << " Switch stop, putters collected." << "\n";
 
   return make_pair(
     ((double)total_observed_total)/((double)total_observations),


wjwithagen:
@tchaikov @liewegas @athanatos 
Hi, Could any of you label this one for testing.
I sure fixes my problem in unittest_throttle, and it looks like it is also the trickpony I was looking for in the crashes with ceph-osds, that had a faal timeout after 600secs.

wjwithagen:
@tchaikov @liewegas @athanatos

Bummer, ceph-osd still has troubles....
In function 'virtual void SyncEntryTimeout::finish(int)' thread 80599c100 time 2016-05-13
 11:25:36.469757
os/filestore/FileStore.cc: 3670: FAILED assert(0)

So the search continues.

tchaikov:
> On FreeBSD is seems that it leads to no (or very little) possibility to switch thread.

the behaviour of "std::condition_variable::wait_until()" depends on how `libc++` is implemented, (assuming in your installation, libc++ is used). and in libc++, `wait_until()` is implemented using `wait_for()`. see https://github.com/llvm-mirror/libcxx/blob/35ff03b7c2a50d197049edbe223042aec221fec1/include/__mutex_base#L362 . so i don't think your change fixes your issue. but we can take it as a cleanup.

tchaikov:
lgtm.

tchaikov:
and @wjwithagen you might want to update your commit message. as a negative wait time does not block the `wait_until()` call.

wjwithagen:
@tchaikov 
I did look in the llvm code and, yes wait_for is defined as now()+delay.
But the current code is start+delay, and start is never set again to now().
So at some point start+delay < now, and it is not defined (in the "standard") what wait_until does..

Looking at the code, it seems to unlock, wait, lock. And when the time_point is passed it probably does not wait. Using the wait_for actually waits for delay time, giving it a larger window to switch thread.

I'm more than willing to refrase the "try to avoid a negative wait time"' but I have the feeling that this is exactly what the patch does.... How about "enlarge waiting window and try to avoid starvation"

tchaikov:
```c++
#include <iostream>
#include <atomic>
#include <condition_variable>
#include <thread>
#include <chrono>
using namespace std::chrono_literals;
 
std::condition_variable cv;
std::mutex cv_m;
std::atomic<int> i{0};
 
void waits(int idx)
{
    std::unique_lock<std::mutex> lk(cv_m);
    auto now = std::chrono::system_clock::now();
    // if(cv.wait_for(lk, - idx*1000ms, [](){return i == 1;}))

    if(cv.wait_until(lk, now - idx*1000ms, [](){return i == 1;}))
        std::cerr << "Thread " << idx << " finished waiting. i == " << i << '\n';
    else
        std::cerr << "Thread " << idx << " timed out. i == " << i << '\n';
}
 
void signals()
{
    std::this_thread::sleep_for(120ms);
    std::cerr << "Notifying...\n";
    cv.notify_all();
    std::this_thread::sleep_for(100ms);
    i = 1;
    std::cerr << "Notifying again...\n";
    cv.notify_all();
}
 
int main()
{
    std::thread t1(waits, 1), t2(waits, 2), t3(waits, 3), t4(signals);
    t1.join(); 
    t2.join();
    t3.join();
    t4.join();
}
```

@wjwithagen i tested the program above using "clang++ -std=c++14 -O0 -g test.cc -lm -o test". and "test" does finish. does it serve as a supporting material of my theory?

wjwithagen:
@tchaikov 
I think it does... But that is not the point of the fix. It does not deadlock. So the wait_until does not block. The way the code works under FreeBSD starves. The other thread is just not allocated any run-time.
The actual problem is that the code is:
     (*ticket)->wait_until(l, start + delay);
Where start is calculated outside the loop thus making the the wait-time 0, at best. 

I assume your read something like:
==== From "man" page
wait_until causes the current thread to block until the condition variable is notified, a specific time is reached, or a spurious wakeup occurs, optionally looping until some predicate is satisfied. 
Atomically releases lock, blocks the current executing thread, and adds it to the list of threads waiting on *this. The thread will be unblocked when notify_all() or notify_one() is executed, or when the absolute time point timeout_time is reached. It may also be unblocked spuriously. When unblocked, regardless of the reason, lock is reacquired and wait_until exits
====
We enter the wait with l locked, and return from the wait with l locked.
So the wait_until release the lock for an insufficient time for throttle.put() to get in and remove something from the throttle list. (or the scheduling in FreeBSD does not give the other (aka put() thread a chance to run) 

The experience is:
When running with a lot of debugging code in throttle, I observe that the while-loop around this wait_until is running at infinitum, going thru cycles of _get_delay(). But nothing changes and throttle.put() is starved.
Allocating another positive delay-slot does gets the job done.

athanatos:
Oh, this actually isn't correct.  This will cause the waiting thread to potentially wait for much longer than intended.  It seems like we don't yet understand what is happening in the freebsd case.  If the problem is with freebsd scheduling fairness, the answer probably isn't to mess up the waiting logic in this code.

wjwithagen:
@athanatos 
I was afraid that somebody was going to say that. :(

I've run the tests BackoffThrottle tests on both Linux and FreeBSD and the answers are exact equal, and concluded that this code leads to the same.
Note that the "problem" will only occur when throttling with a full queue. Not sure how often that is the case. Probably mostly during rebuilds when everything goes to the max. And that will go in a slower pace. 

I see a few ways forward:
1) I try to get to the bottom of this, with the help of the FreeBSD bunch. That will certainly give some answers, (although one could be: don't do that with timepoints that have passed)
That will take a long time to get to a final result.
(It is something that is on my agenda anyways. But I have to come up with an example that is in C and replicates the issue accurate enough)
2) Differentiate the current code for FreeBSD only to what I have now. And that will allow me to get a bit further with the final objective: Get ceph-osd to run without crashing on that 600sec timeout.

On the other hand, the Linux code is now burning CPU-cycles rather hard, since the current code is actually a busy-waiting loop when start+delay < now. And that I actually see in top, because the CPU-cycles on that process/thread go to the max.

athanatos:
@wjwithagen Can you summarize what the actual problem is?  It's not clear to me from this thread where the issue is.  Is it that with some libc++ implementions, wait_until() doesn't behave properly (I also don't understand what the poor behavior is) with timepoint in the past?

wjwithagen:
On 18-5-2016 19:57, Samuel Just wrote:
> @wjwithagen <https://github.com/wjwithagen> Can you summarize what
> the actual problem is? It's not clear to me from this thread where
> the issue is. Is it that with some libc++ implementions, wait_until()
> doesn't behave properly (I also don't understand what the poor
> behavior is) with timepoint in the past?

The actual problem (from observation of debug output) is, running:

test/Throttle.cc:
auto getter = [&]() {
     std::random_device rd;
     std::mt19937 gen(rd());
     std::uniform_int_distribution<> dis(0, 10);

     std::unique_lock<std::mutex> g(l);
     while (!stop) {
       g.unlock();

       uint64_t to_get = dis(gen);
       auto waited = throttle.get(to_get);

       g.lock();
       wait_time += waited;
       waits += to_get;
       total += to_get;
       in_queue.push_back(to_get);
       c.notify_one();
     }
   };

Where throttle.get() is (I've augmented the code a bit to be able to
print times):
std::chrono::duration<double> BackoffThrottle::get(uint64_t c)
{
   locker l(lock);
   auto delay = _get_delay(c);

   // fast path
   if (delay == std::chrono::duration<double>(0) &&
       waiters.empty() &&
       ((max == 0) || (current == 0) || ((current + c) <= max))) {
     current += c;
     return std::chrono::duration<double>(0);
   }

   auto ticket = _push_waiter();

   while (waiters.begin() != ticket) {
     (*ticket)->wait(l);
   }

   auto start = std::chrono::system_clock::now();
   delay = _get_delay(c);
   auto nextt = start + delay;
   // if (current + c) > max the queue would be full
   // nextt > system_clock::now() means that we have waited long enough
   // Loop until delay has passed and there is room to insert
   while ((nextt > std::chrono::system_clock::now())
	  || !((max == 0) || (current == 0) || ((current + c) <= max))
	) {
     assert(ticket == waiters.begin());
     (*ticket)->wait_for(l, delay);
     delay = _get_delay(c);
     nextt = start + delay;
   }
   waiters.pop_front();
   _kick_waiters();

   current += c;
   return std::chrono::system_clock::now() - start;
}

And we get stuck in the while loop. Which keeps running and running because
nothing is extracted from the full list.

That should be done by:
  auto putter = [&]() {
     std::unique_lock<std::mutex> g(l);
     while (!stop) {
       while (in_queue.empty())
	c.wait(g);

       uint64_t c = in_queue.front();

       total_observed_total += total;
       total_observations++;
       in_queue.pop_front();
       assert(total <= max);

       g.unlock();
       std::this_thread::sleep_for(
	c * std::chrono::duration<double>(put_delay_per_count*putters));
       g.lock();

       total -= c;
       throttle.put(c);
     }
   };

uint64_t BackoffThrottle::put(uint64_t c)
{
   locker l(lock);
   assert(current >= c);
   current -= c;
   _kick_waiters();
   return current;
}

But once we are past the timepoint (start+delay) in the
BackoffThrottle::get() while loop, BackoffThrottle::put() is not running
anymore at all. And that is needed to make any progress.

ATM my conclussion is that the putter-thread does not get allocated any
runtime. I'd need to fiddle around a bit when it is the case accurately
  < 0, ==
0 or a small bit bigger than 0.
Fact is that I get putter runs when I use the wait_for(l, delay) version.

The wait_for code comes from the llvm library, and is typical library
code: If you've written it, it is easy to understand. But for mere
mortals it is a rather big maze.

And like Kefu suggests wait_for(time) == wait_until( now + time ).
But then the code goes of into the dark parts of the library and I'd need
to single step it in gdb to actually see where it is going.

And for the moment I'd rather focus on getting further in the Ceph port.
Although the problem I'm having vaguelly feels like a timeing/threading
issue as well.

--WjW

athanatos:
@wjwithagen Ok, so you are saying that wait_until does not release the lock if the time is in the past, and that it is possible with a very small delay for the difference in time between the loop check and the wait_until call that the wait_until loop spins?

athanatos:
@wjwithagen Ok, I think there are two problems.  First, we need to use wait_for, but adjust delay correctly.  Second, when the throttle reaches the full state, we need to use wait() not wait_for().  Does this branch look like the right direction?

https://github.com/athanatos/ceph/tree/wip-throttle

wjwithagen:
On 18-5-2016 20:45, Samuel Just wrote:
> @wjwithagen <https://github.com/wjwithagen> Ok, so you are saying that
> wait_until does not release the lock if the time is in the past, and
> that it is possible with a very small delay for the difference in time
> between the loop check and the wait_until call that the wait_until loop
> spins?

Yup, that is about it.
I think wait_{for,until} needs a small but positive window for the
scheduler so that at least one quantum is available to give to other
threads. How big that window needs to be remains to determine.

What I've seen in some trials is that the product
  double _expected_throughput * double _max_multiple < 10
And I cannot calculate this back to the delay slots, but it is obvious
from experiments that there is a threshold below which no progress is made.

Now I'm sort of afraid that when I dive into this, it turns into
Schrodingers Cat. Observation kills the experiment, and I end up burning
a lot of time chasing this "feature".
Hence my suggestion to for the time being to conditionalise the code like
#if !defined(__FreeBSD__)
	wait_for(l, delay);
#else
	wait_until(l, start+delay);
#endif
So it keeps the behavior you programmed under Linux.

--WjW

athanatos:
@wjwithagen 

I don't want to add FreeBSD specific code unless I understand the difference in platform and have no other choice.

The second patch in the branch I posted above fixes a bug in the linux version as well (spinning on a small delay value while the throttle is actually full).  I think that bug is what is causing the behavior you are seeing: when the throttle is actually full but the delay value is 0, we keep calling wait_until with a time in the past and thereby never releasing the lock.

Are you sure my branch doesn't fix the problem?

wjwithagen:
@athanatos 
Hi Sam,
I ran the code this morning and it completes all three backoffThrottle tests.
So that would be oke as far as FreeBSD concerns.

Personally I had part of my programming training by E.W. Dijkstra: Goto considered harmfull. 8D
And would thus have written the while(true){break;} loop a bit different.

When the queue is not full we just wait to be signaled, that is fine.
When delay is > 0 we wait delay  to throttle.
But the real question still is, what happens when delay gets negative?
Because as far as I can see can
  `delay = _get_delay(c) - (std::chrono::system_clock::now() - start)`
get negative... That has not changed. 
Stronger even the term `_get_delay(c)` is constant for as long as the content/size of the queue has not changed. Where `(now() - start)` keeps growing with every iteration.
And thus if there is heavy contention on the queue, one of the iterations might turn delay negative
In the current case we break out of the loop, even if the queue is still full: `(current + c) > max)`
And we add to the queue on line 393. Which could result in `current > max`.
I've added an `assert(current <= max);` to validate this, but it has not triggered.

athanatos:
delay going negative is fine (it means that we have already waited longer than the throttle would have delayed us), that's why the loop checks for > 0.

Perhaps time spent waiting on the full throttle shouldn't count towards the delay.  I'm ambivalent on that point.  I'm going to test this version and submit a PR once it passes.

athanatos:
@wjwithagen Also, current *can* exceed max if the throttle is empty and the requested amount is larger than max.  It's an edge case, but it's valid.

wjwithagen:
On 19-5-2016 20:11, Samuel Just wrote:
> @wjwithagen <https://github.com/wjwithagen> Also, current /can/ exceed
> max if the throttle is empty and the requested amount is larger than
> max. It's an edge case, but it's valid.

Oke, lets see what your testrun delivers.
And then get it fixed.

--WjW

athanatos:
https://github.com/ceph/ceph/pull/9235

wjwithagen:
@athanatos @ldachary 
After letting it simmer for a few days, I've come to the conclussion that there might not be something wrong with the library code, but more with the test itself.

First we kill all threads, without giving it much thought, and then wew start collecting them.
Expecting everything to come to a smooth end. But why should it?
There are/could be still things in the queue.

So what would be the way to go:
1) Stop the producers, so the queue doesn't fill up any further.
2) Then collect them to make sure nothing is hanging in there.
3) Stop the consumers, 
4) which might the hard case, since they could be waiting on empty queues.
  ( or even more tricky: there are messages, but not enough to fullfill the request in `get(to_get)`
This could be helped the signal the conditionals now and then.

I've this code below now running for about an hour (3 parr. processes) without any hickup
```
  // Terminate all threads we've forked
  // First stop the getters filling the queue
  getter_stop = true;
  c.notify_all();
  for (auto &&i: gts) i.join();
  gts.clear();
  cout << __func__<< ":" << __LINE__ << " getters collected." << "\n";

  // Then all putters need to be stopped
  putter_stop = true;
  c.notify_all();
  for (auto &&i: pts) i.join();
  pts.clear();
  cout << __func__<< ":" << __LINE__ << " putters collected." << "\n";
```
putter_stop and getter_stop go in while loop of the respective function.

wjwithagen:
@ErwanAliasr1 @athanatos @ldachary 
I required a small delay after the `putter_stop = true` to allow the possible race at the top of the while loop from entering the loop until the `c.wait()` to finish first. And only then signal all thread that are waiting.

This has been running at an overloaded server since yesterday. I think it totals to 10.000 runs

athanatos:
@wjwithagen I'm not convinced the changes to Throttle are necessary or correct.  It seems like the bug is in the test.  Please repush this with only the test changes.  If you think Throttle needs to be changed as well, please explain exactly why.

athanatos:
@wjwithagen If https://github.com/ceph/ceph/pull/9111 is an adequate answer, please close this one.

wjwithagen:
@athanatos 
'mmm Don't quite understand the remark about closing #9111, since that is this pull.
But I'll update it to only contain the fixes to the test

athanatos:
@wjwithagen Sorry, I meant https://github.com/ceph/ceph/pull/9455.

This one and https://github.com/ceph/ceph/pull/9455 both only change the test, right?  Do we need both?  Can I close https://github.com/ceph/ceph/pull/9455?

wjwithagen:
On 17-6-2016 20:58, Samuel Just wrote:
> @wjwithagen <https://github.com/wjwithagen> Sorry, I meant #9455
> <https://github.com/ceph/ceph/pull/9455>.
> 
> This one and #9455 <https://github.com/ceph/ceph/pull/9455> both only
> change the test, right? Do we need both? Can I close #9455
> <https://github.com/ceph/ceph/pull/9455>?

Yup,
The test change that Loic made, is also in the change I made for the
unittest.

athanatos:
@wjwithagen You'll also want to change the PR title.

wjwithagen:
On 18-6-2016 00:34, Samuel Just wrote:
> @wjwithagen <https://github.com/wjwithagen> You'll also want to change
> the PR title.

'mmm
I think the PR is actually by Loic, so he needs to do that.
I can/will change the pull title.

--WjW

wjwithagen:
@athanatos 
Sorry, but I noticed that one diff did not make it into the pull.
And without it, it is still possible to get trapped in the wait.

athanatos:
@wjwithagen https://github.com/ceph/ceph/pull/9111/commits/d075a91bc8c7a8e1a5eedf00c9f3b46fcfa876c1

was the missing one?

wjwithagen:
@athanatos 
Nope.... Git again played some tricks with me.(I'm not at all fluent in Git)

There was a piece of code missing in in putter() that mede sure that it would not loop all by itself on: 
`while (in_queue.empty()) c.wait(g);`

Now how git organises all this, I don't know, but if I look at the resulting file in Git,
it is what I have in my tree.

athanatos:
@wjwithagen I think we can fix this more easily: https://github.com/athanatos/ceph/tree/wip-throttle-test .  Can you test that branch?

wjwithagen:
@athanatos 
Yes I can... But testing has proven a rather tricky thing here. So result might take a few thousand itters.

athanatos:
@wjwithagen That's fine.

wjwithagen:
On 29-6-2016 18:29, Samuel Just wrote:
> @wjwithagen <https://github.com/wjwithagen> That's fine.

I'll let 'm run overnight to see what it brings.
If nothing traps, than I think we're good.

wjwithagen:
@athanatos 
After a good nights of about 7500 runs I see that so now and then we get issues like:
Expected: (results.first) > (35), actual: 29.4161 vs 35
Expected: (results.first) > (40), actual: 39.4381 vs 40
Expected: (results.first) > (40), actual: 35.4503 vs 40
Expected: (results.first) > (85), actual: 78.1828 vs 85

On 7500 tests about 180 fails.
And this will generate a FAIL test.

I have to be careful, but I think the way I rewrote the testcode it did not run into this.

athanatos:
@wjwithagen Oops, apparently the order there was important.  Just repushed a new version for you to test.

wjwithagen:
On 30-6-2016 23:27, Samuel Just wrote:
> @wjwithagen <https://github.com/wjwithagen> Oops, apparently the order
> there was important. Just repushed a new version for you to test.

I'll give it another shot...

wjwithagen:
@athanatos 
I'd say that it works...
No more errors of any sort.
Even the stats are within their limits.

athanatos:
Ok, I'll close this one in favor of that one.

athanatos:
Replaced by https://github.com/ceph/ceph/pull/10094

PR #9112, opened 2016-05-13
mds/mdbalancer: fixup assignment type in the hit_dir

renhwztetecs:
The DecayCounter of val is double type,
so dir_pop&pop_sp&rdp should be double type,
otherwise it will affect the accuracy of calculation.


jcsp:
Test run OK http://pulpito.ceph.com/jspray-2016-05-26_17:59:25-fs-wip-jcsp-testing-20160527---basic-mira/

PR #9113, opened 2016-05-13
Jewel

love-vicky:


tchaikov:
we do the backport by following a process instead of massively merging from master.

PR #9114, opened 2016-05-13
osd/ReplicatedPG: unreserve replicas scrub reservation during on_chan…

xiexingguo:
…ge()

In case there is any potential scrub reservation leak.


xiexingguo:
@tchaikov Not very sure of this, please help to confirm. Thanks.

liewegas:
@athanatos

athanatos:
This isn't correct.  The replicas will call clear_scrub_reserved() on their own since they will see the same map change.  Further, this is after the interval change has logically happened, so the unreserve would be sent in the *next* interval, and so would actually be wrong (the sender isn't even necessarily the primary any more).

PR #9116, opened 2016-05-13
librados: use caller provided snapid for aio_sparse_read()

ddiss:
...instead of the value set via set_snap_read(). This makes
aio_sparse_read() consistent with aio_read(), which also accepts a
snapid parameter.


liewegas:
@jdurgin

jdurgin:
lgtm

PR #9117, opened 2016-05-13
osd: clean pglog code

wjin:


tchaikov:

PR #9118, opened 2016-05-13
rgw: fix realm pull and period pull for apache frontend

oritwas:
add missing content-type and content-length headers

Fixes: http://tracker.ceph.com/issues/15846


cbodley:
:+1:

PR #9123, opened 2016-05-13
qa: rbd: don't modprobe, chown sysfs files or udevadm settle

idryomov:


lande1234:
@idryomov  @dillaman  
HI!
I have a question that why  we drop the sudo. 
when i test the  rbd_simple_big.yaml suite i have meet zhe permission's problem.

the wrong list:
`2016-07-16T15:47:35.430 INFO:tasks.workunit.client.0.plana121.stderr:+ dd if=/dev/zero of=/dev/rbd0 bs=1M count=100000
2
2016-07-16T15:47:35.432 INFO:tasks.workunit.client.0.plana121.stderr:dd: failed to open 鈥�/dev/rbd0鈥�: Permission denied

2016-07-16T15:47:35.432 INFO:tasks.workunit:Stopping ['rbd/simple_big.sh'] on client.0...`

idryomov:
@lande1234 Which lab is this?  Need to make sure teuthology user  is added to the "disk" group.

lande1234:
@idryomov   when I set up the teuthology frame according to zhe teuthology doc ,I didn't notice that how to add zhe user to zhe 'disk' group.  can you explain it or give a link to me?  thanks!

idryomov:
@lande1234 Test boxes are configured with ansible, which will do it for you:

https://github.com/ceph/ceph-cm-ansible/blob/master/roles/testnode/tasks/apt_systems.yml#L70
https://github.com/ceph/ceph-cm-ansible/blob/master/roles/testnode/tasks/yum_systems.yml#L28

PR #8815, opened 2016-04-28
jewel: debian: install systemd target files

ktdreyer:
* enable it using dh_systemd_enable
* start the target using dh_systemd_start
* move the dh_installinit, dh_systemd_enable, dh_systemd_start calls
  down, so they can identify the service files if they care about them.

(cherry picked from commit f950a0a7f5b35e13668379bc69dcb5d98a2dfc1f)

Backport for http://tracker.ceph.com/issues/15646

PR #9127, opened 2016-05-14
debian: install systemd target files

wido:
* enable it using dh_systemd_enable
* start the target using dh_systemd_start
* move the dh_installinit, dh_systemd_enable, dh_systemd_start calls
  down, so they can identify the service files if they care about them.


tchaikov:
@wido we have a PR for this already, see #8815. so i am closing yours.

PR #9130, opened 2016-05-15
journal: live replay might skip entries from previous object set

dillaman:


dillaman:
@jdurgin @trociny I've run this through the "client-killer" stress test a few times and it appears to be stable now.  I've removed the DNM tag from the PR.

vshankar:
@dillaman @jdurgin I tested this out and got a clean journal replay all times..

trociny:
@dillaman I don't like very much we have so much code duplication between rbd_mirror_stress.sh and rbd_mirror.sh. 

May be add `stress` function into `rbd_mirror.sh` instead? It would run setup if needed and then stress tests. It wouldn't executed during normal `rbd_mirror.sh` run, only when it is run as `rbd_mirror.sh stress`. And `rbd_mirror_stress.sh` would just call it.

Another option is to share the functions in `rbd_mirror-helpers.sh` and source it from both scripts.

Still, it is not a stopper, I can do it later if you wish.

dillaman:
@trociny I don't like the duplication either. The problem is that (right now) teuthology workunit tests are expected to be self contained. It would be nice to add a "support_files" tag to the workunit task so that it knows to download shared code dependencies.

qa/workunits/rbd/rbd_mirror_stress.sh
```diff
@@ -0,0 +1,410 @@
...
+
+    rbd --cluster ${cluster} -p ${POOL} create --size ${size} \
+	--image-feature exclusive-lock --image-feature journaling ${image}
+}
+
+write_image()
+{
+    local cluster=$1
+    local image=$2
+    local duration=$(($RANDOM % 35 + 15))
```
trociny: On debian dash is used as /bin/sh, and $RANDOM is not known here.

Change `#!/bin/sh` to `#!/bin/bash`? Or use a more portable way?

https://wiki.ubuntu.com/DashAsBinSh#A.24RANDOM

trociny: An example with awk:
```
duration=$(awk 'BEGIN {srand(); print int(35 * rand()) + 15}')
```

PR #9133, opened 2016-05-15
cmake: add selinux support

tchaikov:


tchaikov:
@cbodley could you take a look?

@branto1 as 'fsadm_manage_pid()' is only found in fedora and rhel/centos, so i think it's okay to do so? @smithfarm what do you think, does SuSE has this interface defined?

CMakeLists.txt
```diff
@@ -344,6 +344,13 @@ include_directories(${Boost_INCLUDE_DIRS})
 
 find_package(Threads REQUIRED)
 
+option(WITH_SELINUX "build SELinux policy" OFF)
```
smithfarm: @tchaikov Does this mean `WITH_SELINUX` is off by default?

smithfarm:
@tchaikov I've never heard of anyone using SELinux with SUSE.

CMakeLists.txt
tchaikov: @smithfarm yes, `option()` is `OFF` by default. and i put `OFF` here to make it explicit.

tchaikov:
thanks for your confirmation, Nathan!

branto1:
@tchaikov I wonder, where have you been hitting the issues with fsadm_manage_pid? The macro is not available with older versions of SELinux policy packeges (like the ones in RHEL 6) but those should be fairly ancient by nowadays standards. Also, I'm not exactly sure whether distro_redhat covers fedora (although I think it does).

btw: Ignoring that line on non-redhat platforms would probably result in denials on that platform.

selinux/ceph.te
```diff
@@ -106,7 +106,9 @@ allow ceph_t self:process setpgid;
 allow ceph_t var_run_t:dir { write create add_name };
 allow ceph_t var_run_t:file { write create open getattr };
 
+ifdef(`distro_redhat', `)
```
branto1: There is a typo, here: the ')' character at the end of the line should not be there.

tchaikov: oh, thanks!

tchaikov:
@branto1 

i was trying to compile my patch on a debian box, hence this change. i had compilation error if `fsadm_manage_pid` is not defined.

> btw: Ignoring that line on non-redhat platforms would probably result in denials on that platform.

i understand the risk. will drop this change anyway. maybe just build failure with '-DWITH_SELINUX=ON' on a not-supported distro would be better than failure at run-time.

tchaikov:
changelog

- dropped the change to ceph.te.

CMakeLists.txt
```diff
@@ -344,6 +344,13 @@ include_directories(${Boost_INCLUDE_DIRS})
...
+if(WITH_SELINUX)
+  find_file(SELINUX_MAKEFILE selinux/devel/include/Makefile
+    PATH /usr/share)
```
cbodley: should it be a fatal error if we don't find this?

tchaikov:
changelog

- bail out with FATAL_ERROR if SELinux's Makefile is not found.

@cbodley thanks. fixed, tested and re-pushed

tchaikov:
@cbodley ping

tchaikov:
i just ported the selinux/Makefile.am to cmake and the ceph.pp file is generated as expected. also they artifacts are installed just fine.

PR #9137, opened 2016-05-16
common/blkdev: accept symbolic link

majianpeng:


majianpeng:
@liewegas . For this function( https://trello.com/c/rX8egn9a/96-bdev-detect-ssd-vs-hdd), i notice Varada Kari don't implement until now. So i implement it.

majianpeng:
@liewegas, ping!

src/common/blkdev.cc
```diff
@@ -68,11 +69,18 @@ int get_block_device_base(const char *dev, char *out, size_t out_len)
   DIR *dir;
   char devname[PATH_MAX], fn[PATH_MAX];
   char *p;
+  char realname[PATH_MAX] = {0};
 
-  if (strncmp(dev, "/dev/", 5) != 0)
-    return -EINVAL;
+  if (strncmp(dev, "/dev/", 5) != 0) {
+    if ((readlink(dev, realname, sizeof(realname)) == -1) || (strncmp(realname, "/dev/", 5) != 0))
+	  return -EINVAL;
```
liewegas: indentation

src/os/bluestore/KernelDevice.cc
```diff
@@ -115,9 +115,12 @@ int KernelDevice::open(string p)
     if (r < 0) {
       goto out_fail;
     }
+
+    rotational = block_device_is_rotational(path.c_str());
     size = s;
   } else {
     size = st.st_size;
+    rotational = true; //file is rotational device
```
liewegas: nit: space between // and comment

src/os/bluestore/KernelDevice.cc
```diff
@@ -142,6 +145,7 @@ int KernelDevice::open(string p)
 	  << " (" << pretty_si_t(size) << "B)"
 	  << " block_size " << block_size
 	  << " (" << pretty_si_t(block_size) << "B)"
+	  << " device is " << (rotational ? "rotational" : "no-rotational")
```
liewegas: s/" device is "/" "/
s/no-rotational/non-rotational/

src/os/bluestore/NVMEDevice.cc
```diff
@@ -834,6 +834,9 @@ int NVMEDevice::open(string p)
     zeros.zero();
   }
 
+  //nvme is no-rotational device.
```
liewegas: // nvme is non-rotational

manali26:
I have few comments : 
1. Can we use ioctl with BLKROTATIONAL parameter to check whether the device is rotational or not. Instead of parsing the file using get_block_device_int_property() ?

2. Also, can we have a enum structure with media_types like {HDD, SSD, NVRAM, OTHER}
    And set it appropriately instead of having a bool which returns the type as rotational or not.
    Is there a use case for this ? If not lets keep it rotational/non rotational.

Thank you
Manali

liewegas:
I think an enum makes sense, although we don't have a use for it yet.  I think the main thing would be to make sure we use an is_rotation() accessor so that if we make that change later it doesn't break things.

We want to symlink handling either way so that we can check other /sys parameters.  No opinion on whether we use the ioctl here or not.

majianpeng:
@liewegas . update.

liewegas:
@manali26 we couldn't tell who you are (name, email).  Can you update your github profile so we can add Reviewed-by: tags for you next time?  thanks!

PR #9138, opened 2016-05-16
osd: replace ceph:atomic_t with std::atomic in osd module.

cxwshawn:
trying to use std more, and atomic_t easier to use and no need
t implement more interfaces, like operator++,++ etc.

### DIFF
--- a/src/osd/OSD.h
+++ b/src/osd/OSD.h
@@ -817,7 +817,7 @@ public:
 
 private:
   /// throttle promotion attempts
-  atomic_t promote_probability_millis; ///< probability thousands. one word.
+  std::atomic_uint promote_probability_millis{1000}; ///< probability thousands. one word.
   PromoteCounter promote_counter;
   utime_t last_recalibrate;
   unsigned long promote_max_objects, promote_max_bytes;
@@ -826,13 +826,13 @@ public:
   bool promote_throttle() {
     // NOTE: lockless!  we rely on the probability being a single word.
     promote_counter.attempt();
-    if ((unsigned)rand() % 1000 > promote_probability_millis.read())
+    if ((unsigned)rand() % 1000 > promote_probability_millis)
       return true;  // yes throttle (no promote)
     if (promote_max_objects &&
-	promote_counter.objects.read() > promote_max_objects)
+	promote_counter.objects > promote_max_objects)
       return true;  // yes throttle
     if (promote_max_bytes &&
-	promote_counter.bytes.read() > promote_max_bytes)
+	promote_counter.bytes > promote_max_bytes)
       return true;  // yes throttle
     return false;   //  no throttle (promote)
   }
@@ -861,9 +861,9 @@ public:
 
   // -- tids --
   // for ops i issue
-  atomic_t last_tid;
+  std::atomic_uint last_tid{0};
   ceph_tid_t get_tid() {
-    return (ceph_tid_t)last_tid.inc();
+    return (ceph_tid_t)last_tid++;
   }
 
   // -- backfill_reservation --
@@ -1124,18 +1124,18 @@ public:
     NOT_STOPPING,
     PREPARING_TO_STOP,
     STOPPING };
-  atomic_t state;
+  std::atomic_int state{NOT_STOPPING};
   int get_state() {
-    return state.read();
+    return state;
   }
   void set_state(int s) {
-    state.set(s);
+    state = s;
   }
   bool is_stopping() {
-    return get_state() == STOPPING;
+    return state == STOPPING;
   }
   bool is_preparing_to_stop() {
-    return get_state() == PREPARING_TO_STOP;
+    return state == PREPARING_TO_STOP;
   }
   bool prepare_to_stop();
   void got_stop_ack();
@@ -1338,32 +1338,32 @@ public:
   }
 
 private:
-  atomic_t state;
+  std::atomic_int state{STATE_INITIALIZING};
 
 public:
   int get_state() {
-    return state.read();
+    return state;
   }
   void set_state(int s) {
-    state.set(s);
+    state = s;
   }
   bool is_initializing() {
-    return get_state() == STATE_INITIALIZING;
+    return state == STATE_INITIALIZING;
   }
   bool is_preboot() {
-    return get_state() == STATE_PREBOOT;
+    return state == STATE_PREBOOT;
   }
   bool is_booting() {
-    return get_state() == STATE_BOOTING;
+    return state == STATE_BOOTING;
   }
   bool is_active() {
-    return get_state() == STATE_ACTIVE;
+    return state == STATE_ACTIVE;
   }
   bool is_stopping() {
-    return get_state() == STATE_STOPPING;
+    return state == STATE_STOPPING;
   }
   bool is_waiting_for_healthy() {
-    return get_state() == STATE_WAITING_FOR_HEALTHY;
+    return state == STATE_WAITING_FOR_HEALTHY;
   }
 
 private:
--- a/src/osd/PG.h
+++ b/src/osd/PG.h
@@ -32,7 +32,6 @@
 #include "include/stringify.h"
 #include "osd_types.h"
 #include "include/xlist.h"
-#include "include/atomic.h"
 #include "SnapMapper.h"
 
 #include "PGLog.h"
@@ -41,6 +40,7 @@
 #include "include/str_list.h"
 #include "PGBackend.h"
 
+#include <atomic>
 #include <list>
 #include <memory>
 #include <string>
@@ -251,7 +251,7 @@ protected:
    * put_unlock() when done with the current pointer (_most common_).
    */  
   mutable Mutex _lock;
-  atomic_t ref;
+  std::atomic_uint ref{0};
 
 #ifdef PG_DEBUG_REFS
   Mutex _ref_id_lock;
--- a/src/osd/osd_types.h
+++ b/src/osd/osd_types.h
@@ -43,6 +43,7 @@
 #include "OpRequest.h"
 #include "include/cmp.h"
 #include "librados/ListObjectImpl.h"
+#include <atomic>
 
 #define CEPH_OSD_ONDISK_MAGIC "ceph osd volume v026"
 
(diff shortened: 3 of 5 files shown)


src/osd/OSD.h
```diff
@@ -1024,18 +1024,18 @@ class OSDService {
     NOT_STOPPING,
     PREPARING_TO_STOP,
     STOPPING };
-  atomic_t state;
+  atomic_uint state;
```
tchaikov: atomic_int

cxwshawn:
@tchaikov  please review again, thanks .

gregsfortytwo:
I think @adamemerson or somebody in his group was working on a more comprehensive set of changes to our atomics. Maybe he has thoughts about how incrementally we want to handle this transition.

adamemerson:
I didn't have a plan to do a big replacement all at once, though I can if people like. It would have a portability advantage since all we'd depend upon is C++ <atomic> support, rather than pulling in libatomic_ops (or falling back on the spinlock.)

It would also open up the opportunity to use looser memory models where appropriate (though I would want to consult with the people more knowledgeable in any given subsystem before relaxing things.)

cxwshawn:
@adamemerson @gregsfortytwo  so what's the meaning ?  is  this PR meaningful ? I am also trying to stdanlize ceph code.

liewegas:
@cxwshawn It means this is a step in the right direction, and nobody was working on it yet.  :)

liewegas:
retest this please

src/osd/OSD.h
```diff
@@ -1241,32 +1241,32 @@ class OSD : public Dispatcher,
   }
 
 private:
-  atomic_t state;
+  atomic_uint state;
```
tchaikov: should use `atomic_int`, as the `state` here is actually .

cxwshawn:
@tchaikov  :+1: ,  thanks for the review, fixed

src/osd/OSD.h
```diff
@@ -796,7 +796,7 @@ class OSDService {
 
 private:
   /// throttle promotion attempts
-  atomic_t promote_probability_millis; ///< probability thousands. one word.
+  atomic_uint promote_probability_millis; ///< probability thousands. one word.
```
cbodley: i'd prefer if we were explicit about the `std::` part, especially in headers - we shouldn't assume that other headers are `using namespace std;`

liewegas: agreed on std::

cxwshawn:
@cbodley thanks for your advice, @liewegas  have fixed, please review.

liewegas:
lgtm

liewegas:
valgrind uninit conditoin errors:
<pre>
  <kind>UninitCondition</kind>
  <what>Conditional jump or move depends on uninitialised value(s)</what>
  <stack>
    <frame>
      <ip>0xD249370</ip>
      <obj>/usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.19</obj>
      <fn>std::ostreambuf_iterator&lt;char, std::char_traits&lt;char&gt; &gt; std::num_put&lt;char, std::ostreambuf_iterator&lt;char, std::char_traits&lt;char&gt; &gt; &gt;::_M_insert_int&lt;unsigned long&gt;(std::ostreambuf_iterator&lt;char, std::char_traits&lt;char&gt; &gt;, std::ios_base&amp;, char, unsigned lon
g) const</fn>
    </frame>
    <frame>
      <ip>0xD24960C</ip>
      <obj>/usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.19</obj>
      <fn>std::num_put&lt;char, std::ostreambuf_iterator&lt;char, std::char_traits&lt;char&gt; &gt; &gt;::do_put(std::ostreambuf_iterator&lt;char, std::char_traits&lt;char&gt; &gt;, std::ios_base&amp;, char, unsigned long) const</fn>
    </frame>
    <frame>
      <ip>0xD2552CD</ip>
      <obj>/usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.19</obj>
      <fn>std::ostream&amp; std::ostream::_M_insert&lt;unsigned long&gt;(unsigned long)</fn>
    </frame>
    <frame>
      <ip>0x465DE2</ip>
      <obj>/usr/bin/ceph-osd</obj>
      <fn>operator&lt;&lt;</fn>
      <dir>/usr/include/c++/4.8</dir>
      <file>ostream</file>
      <line>171</line>
    </frame>
    <frame>
      <ip>0x465DE2</ip>
      <obj>/usr/bin/ceph-osd</obj>
      <fn>OSDService::promote_throttle_recalibrate()</fn>
      <dir>/srv/autobuild-ceph/gitbuilder.git/build/out~/ceph-10.2.0-1558-g22ad94c/src/osd</dir>
      <file>OSD.cc</file>
      <line>629</line>
    </frame>
    <frame>
      <ip>0x4BEE9F</ip>
      <obj>/usr/bin/ceph-osd</obj>
      <fn>OSD::tick()</fn>
      <dir>/srv/autobuild-ceph/gitbuilder.git/build/out~/ceph-10.2.0-1558-g22ad94c/src/osd</dir>
      <file>OSD.cc</file>
      <line>4336</line>
    </frame>
    <frame>
      <ip>0x4C7228</ip>
      <obj>/usr/bin/ceph-osd</obj>
      <fn>Context::complete(int)</fn>
      <dir>/srv/autobuild-ceph/gitbuilder.git/build/out~/ceph-10.2.0-1558-g22ad94c/src/./include</dir>
      <file>Context.h</file>
      <line>64</line>
    </frame>
</pre>
and several others, all in promote_throttle_recalibrate.  Maybe teh atomic isn't getting initialized to 0?  the old atomic_t did that.
/a/yuriw-2016-06-01_08:07:30-rados-wip-yuri-testing-distro-basic-smithi/227666

adamemerson:
Atomic versions of primitive types are not initialized. I'd recommend just sticking a ' = 0' after each in the class declaration.

(Actually this is my preference in most cases. If the class has a whole bunch of stuff that will be initialized to the same value regardless of the constructor called or parameters given, I think it's better style to set its default at declaration. That makes the initializer list in the constructor less noisy.)

cxwshawn:
@liewegas sorry for the latency, maybe this is caused by osd_types.h member variable-not-initialized, have fixed, please try again, @adamemerson yeah, I have changed the uninited variable initiated in class declaration other than constructor initialize list.

src/osd/OSD.h
```diff
@@ -796,7 +796,7 @@ class OSDService {
 
 private:
   /// throttle promotion attempts
-  atomic_t promote_probability_millis; ///< probability thousands. one word.
+  std::atomic_uint promote_probability_millis; ///< probability thousands. one word.
```
athanatos: Should be initialized here.

src/osd/OSD.h
```diff
@@ -840,9 +840,9 @@ class OSDService {
 
   // -- tids --
   // for ops i issue
-  atomic_t last_tid;
+  std::atomic_uint last_tid;
```
athanatos: Should be initialized here.

src/osd/OSD.h
```diff
@@ -1024,18 +1024,18 @@ class OSDService {
     NOT_STOPPING,
     PREPARING_TO_STOP,
     STOPPING };
-  atomic_t state;
+  std::atomic_int state;
```
athanatos: Should be initialized here.

src/osd/OSD.h
```diff
@@ -1241,32 +1241,32 @@ class OSD : public Dispatcher,
   }
 
 private:
-  atomic_t state;
+  std::atomic_int state;
```
athanatos: Should be initialized here.

src/osd/PG.h
```diff
@@ -33,7 +33,7 @@
 #include "osd_types.h"
 #include "include/buffer_fwd.h"
 #include "include/xlist.h"
-#include "include/atomic.h"
+// #include "include/atomic.h"
```
athanatos: Remove this completely.

src/osd/PG.h
```diff
@@ -245,7 +246,7 @@ class PG : DoutPrefixProvider {
    * put_unlock() when done with the current pointer (_most common_).
    */  
   mutable Mutex _lock;
-  atomic_t ref;
+  std::atomic_uint ref;
```
athanatos: Should be initialized here.

src/osd/osd_types.h
```diff
@@ -4280,24 +4281,26 @@ enum scrub_error_type {
 // PromoteCounter
 
 struct PromoteCounter {
-  atomic64_t attempts, objects, bytes;
+  std::atomic_ullong  attempts{0};
```
athanatos: Yep, like this.

src/osd/osd_types.h
```diff
@@ -4280,24 +4281,26 @@ enum scrub_error_type {
...
+  std::atomic_ullong  objects{0};
```
athanatos: Yep, like this.

src/osd/osd_types.h
```diff
@@ -4280,24 +4281,26 @@ enum scrub_error_type {
...
+  std::atomic_ullong  bytes{0};
```
athanatos: Yep, like this.

athanatos:
You seem to have missed several more initializations.  We'll re-add needs-qa once those are fixed.

cxwshawn:
@athanatos  since the missed you said has been initialized by using constructor initialize list, so is there any need to reinitialize using xxx a{0} ?

athanatos:
You are claiming that they are already initialized in the constructor list?  I'd actually prefer you moved the initialization to the declaration like you did with the other variables.  Are those not also initialized in the constructor?

cxwshawn:
@athanatos yes, exactly, the former variables has been initialized in the constructor list, so I think there is no need to reinitialize, but since declaration initialization is preferred, I will change all the atomic variables to this form.

cxwshawn:
@athanatos all modified atomic var has been initialized when declaration, please help merge, thanks!

athanatos:
make check seems to be failing

cxwshawn:
@athanatos it's a little weired, does this failure relate with this change  ?

tchaikov:
@cxwshawn you might want to rebase your changeset against master for a green build.

cxwshawn:
@tchaikov @athanatos rebased, but still failed, is there any hint for this failure ? I cannot see any information related with this change.

tchaikov:

PR #9147, opened 2016-05-16
doc: update required packages list to build doc

flaf:
* The package "zlib1g-dev" is needed and was not mentioned.
* "gcc" and "virtualenv" packages are added explicitly, because
  these packages can be not installed if APT::Install-Recommends
  is set to "false" (personally it's my case in all my Debian
  distributions).
* Debian Jessie and Ubuntu Xenial cases are added.

Signed-off-by: François Lafont <francois.lafont@ac-versailles.fr>

### DIFF
--- a/doc/dev/generatedocs.rst
+++ b/doc/dev/generatedocs.rst
@@ -28,32 +28,22 @@ You should have a full copy of the Ceph repository.
 Install the Required Tools
 --------------------------
 
-To build the Ceph documentation, the following packages are required on
-Ubuntu 14.04:
+To build the Ceph documentation, some dependencies are required.
+To know what packages are needed, you can launch this command::
 
-- ``python-dev``
-- ``python-pip``
-- ``python-virtualenv``
-- ``libxml2-dev``
-- ``libxslt-dev``
-- ``doxygen``
-- ``ditaa``
-- ``graphviz``
-- ``ant``
-- ``cython``
-- ``librbd-dev``
+    cd ceph
+    admin/build-doc
 
-Execute ``apt-get install`` for each dependency that isn't installed
-on your host.::
-
-	sudo apt-get install python-dev python-pip python-virtualenv libxml2-dev libxslt-dev doxygen ditaa graphviz ant cython librbd-dev
-
+If dependencies are missing, the command above will fail
+with a message that suggests you a command to install all
+missing dependencies.
 
 
 Build the Documents
 -------------------
 
-Once you have installed all the dependencies, execute the build::
+Once you have installed all the dependencies, execute the build (the
+same command as above)::
 
 	cd ceph
 	admin/build-doc
--- a/admin/build-doc
+++ b/admin/build-doc
@@ -7,16 +7,31 @@ TOPDIR=`pwd`
 install -d -m0755 build-doc
 
 if command -v dpkg >/dev/null; then
-    for package in python-dev python-pip python-virtualenv doxygen ditaa ant libxml2-dev libxslt1-dev zlib1g-dev cython graphviz; do
-	if [ "$(dpkg --status -- $package|sed -n 's/^Status: //p')" != "install ok installed" ]; then
-            # add a space after old values
-	    missing="${missing:+$missing }$package"
-	fi
+    packages='
+      git
+      gcc
+      python-dev
+      python-pip
+      python-virtualenv
+      libxml2-dev
+      libxslt1-dev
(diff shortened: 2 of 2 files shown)


ktdreyer:
We should probably just recommend `./install-deps.sh` at this point? this list is bound to grow stale over time

ktdreyer:
whoops, this is for the *docs* build. In that case, it's `./admin/build-doc`

flaf:
Sorry, I'm not sure to understand your comment. `./admin/build-doc` is the script to build the doc, ok with that. In my commit, I have just updated the part "what packages are needed **before** to launch `./admin/build-doc`".

So what do you propose exactly?

doc/dev/generatedocs.rst
```diff
@@ -42,12 +44,20 @@ Ubuntu 14.04:
 - ``ant``
 - ``cython``
 - ``librbd-dev``
+- ``zlib1g-dev``
 
 Execute ``apt-get install`` for each dependency that isn't installed
 on your host.::
 
-	sudo apt-get install python-dev python-pip python-virtualenv libxml2-dev libxslt-dev doxygen ditaa graphviz ant cython librbd-dev
+	sudo apt-get install gcc python-dev python-pip python-virtualenv libxml2-dev libxslt-dev doxygen ditaa graphviz ant cython librbd-dev zlib1g-dev
```
tchaikov: @flaf i think what @ktdreyer suggests is to just:

* recommend the document writer or whoever wants to build the doc to use `./admin/build-doc` here. instead of maintaining the dependency list in two places. and this one is easier to grow stale over the time.

and i think we might want to update that script instead if preparing the building environment involves more dependencies.

flaf:
Err... currently `./admin/build-doc` doesn't install any required dependencies, it just build the doc, correct?

So, if I understand well, the suggestion is:

1. Remove this part in the doc http://docs.ceph.com/docs/master/dev/generatedocs/#install-the-required-tools
2. Add some code directly in `./admin/build-doc` to install *automatically* the required packages (in function of the current distribution) before launching the build of the doc.

Do I have well understood now? If yes, in this case, the script must be launched as root, correct? Personally I prefer a script `./admin/build-doc` which makes one thing: just build the doc.

flaf:
@tchaikov, you said: "recommend the document writer or whoever wants to build the doc to use `./admin/build-doc` here". But it's already recommended in the doc => http://docs.ceph.com/docs/master/dev/generatedocs/#build-the-documents

So, I'm sorry but I definitively don't understand.

tchaikov:
> 1. Remove this part in the doc http://docs.ceph.com/docs/master/dev/generatedocs/#install-the-required-tools

probably we can instead reference the output of `./admin/build-doc`, something like "follow the suggested command printed by `./admin/build-doc`.

> you said: "recommend the document writer or whoever wants to build the doc to use `./admin/build-doc` here". But it's already recommended in the doc

right. we suggest to execute it after installing all the dependencies, without noting that  `./admin/build-doc` also prints the command to install the missing packages.

> 2. Add some code directly in `./admin/build-doc` to install *automatically* the required packages (in function of the current distribution) before launching the build of the doc.

not *automatically*. personally, i tent do keep it in its current way: just prints the "apt-get" or "yum" command line. like https://github.com/ceph/ceph/blob/master/admin/build-doc#L18

>  If yes, in this case, the script must be launched as root, correct? 

i agree with you that 

1. we don't want to the script to be launched by root
1. it should only *build* the doc
1. the generatedocs.rst should be update to date.

but it would be easier just to keep the dependency list in a single place. and since user will always run `./admin/build-doc` to build the doc, i guess it would be more user-friendly to show the (missing) packages there, and reference the script in the "Install the Required Tools" section.

flaf:
Ok, understood now. Thx. ;)
I come back asap with new commits.

flaf:
Hi, is it OK for you now (I have made commit + squash)?

I'm sorry, I'm only able to update the "Debian" part (I don't use Redhat at all).

ktdreyer:
Thanks, LGTM (and thanks @tchaikov as always!)

flaf:
Hi,

Maybe I'm wrong but the PR doesn't seem to be merged. Do I have missed something?
Or maybe I'm a little impatient. ;)

tchaikov:
@flaf sorry for the latency! merged!

flaf:
@tchaikov, no problem. ;)
Thx.

PR #9149, opened 2016-05-16
rgw: add_zone only clears master_zone if --master=false

cbodley:
Fixes: http://tracker.ceph.com/issues/15901


yehudasa:
@cbodley lgtm

ktdreyer:
Looks like this needs a rebase @cbodley

cbodley:
rebased, thanks

PR #9151, opened 2016-05-16
rbd: add methods to set and get snapshot limits

fullerdj:
Add a subcommand, rbd snap limit <image> --limit <integer>, to
limit the number of snapshots that may be made of a given image.
Specifying a value less than zero removes the limit. Add an omap
key to the RBD header for implementation. Add object classes to
set and query the limit. Older OSDs will ignore the limit.

Fixes: http://tracker.ceph.com/issues/15706

### DIFF
--- a/src/cls/rbd/cls_rbd.cc
+++ b/src/cls/rbd/cls_rbd.cc
@@ -108,6 +108,8 @@ cls_method_handle_t h_metadata_set;
 cls_method_handle_t h_metadata_remove;
 cls_method_handle_t h_metadata_list;
 cls_method_handle_t h_metadata_get;
+cls_method_handle_t h_snapshot_get_limit;
+cls_method_handle_t h_snapshot_set_limit;
 cls_method_handle_t h_old_snapshots_list;
 cls_method_handle_t h_old_snapshot_add;
 cls_method_handle_t h_old_snapshot_remove;
@@ -1499,6 +1501,7 @@ int snapshot_add(cls_method_context_t hctx, bufferlist *in, bufferlist *out)
 {
   bufferlist snap_namebl, snap_idbl;
   cls_rbd_snap snap_meta;
+  uint64_t snap_limit;
 
   try {
     bufferlist::iterator iter = in->begin();
@@ -1542,7 +1545,16 @@ int snapshot_add(cls_method_context_t hctx, bufferlist *in, bufferlist *out)
     return r;
   }
 
+  r = read_key(hctx, "snap_limit", &snap_limit);
+  if (r == -ENOENT) {
+    snap_limit = UINT64_MAX;
+  } else if (r < 0) {
+    CLS_ERR("Could not read snapshot limit off disk: %s", cpp_strerror(r).c_str());
+    return r;
+  }
+
   int max_read = RBD_MAX_KEYS_READ;
+  uint64_t total_read = 0;
   string last_read = RBD_SNAP_KEY_PREFIX;
   do {
     map<string, bufferlist> vals;
@@ -1550,6 +1562,12 @@ int snapshot_add(cls_method_context_t hctx, bufferlist *in, bufferlist *out)
 			     max_read, &vals);
     if (r < 0)
       return r;
+
+    total_read += vals.size();
+    if (total_read >= snap_limit) {
+      CLS_ERR("Attempt to create snapshot over limit of %lu", snap_limit);
+      return -EDQUOT;
+    }
 
     for (map<string, bufferlist>::iterator it = vals.begin();
 	 it != vals.end(); ++it) {
@@ -2675,6 +2693,50 @@ int metadata_get(cls_method_context_t hctx, bufferlist *in, bufferlist *out)
   return 0;
 }
 
+int snapshot_get_limit(cls_method_context_t hctx, bufferlist *in,
+		       bufferlist *out)
+{
+  int rc;
+  uint64_t snap_limit;
+
+  rc = read_key(hctx, "snap_limit", &snap_limit);
+  if (rc == -ENOENT) {
+    rc = 0;
+    ::encode(UINT64_MAX, *out);
+  } else {
+    ::encode(snap_limit, *out);
+  }
+
+  CLS_LOG(20, "read snapshot limit %lu", snap_limit);
+  return rc;
+}
+
+int snapshot_set_limit(cls_method_context_t hctx, bufferlist *in,
+		       bufferlist *out)
+{
+  int rc;
+  uint64_t new_limit;
+  bufferlist bl;
+
+  try {
+    bufferlist::iterator iter = in->begin();
+    ::decode(new_limit, iter);
+  } catch (const buffer::error &err) {
+    return -EINVAL;
+  }
+
+  if (new_limit == UINT64_MAX) {
+    CLS_LOG(20, "remove snapshot limit\n");
+    rc = cls_cxx_map_remove_key(hctx, "snap_limit");
+  } else {
+    CLS_LOG(20, "set snapshot limit to %lu\n", new_limit);
+    ::encode(new_limit, bl);
+    rc = cls_cxx_map_set_val(hctx, "snap_limit", &bl);
+  }
+
+  return rc;
+}
+
 
 /****************************** Old format *******************************/
 
@@ -2745,6 +2807,17 @@ int old_snapshot_add(cls_method_context_t hctx, bufferlist *in, bufferlist *out)
 
   if (header->snap_seq > snap_id)
     return -ESTALE;
+
+  uint64_t snap_limit;
+  rc = read_key(hctx, "snap_limit", &snap_limit);
+  if (rc == -ENOENT) {
+    snap_limit = UINT64_MAX;
+  } else if (rc < 0) {
+    return rc;
+  }
+
+  if (header->snap_count >= snap_limit)
+    return -EDQUOT;
 
   const char *cur_snap_name;
   for (cur_snap_name = snap_names; cur_snap_name < end; cur_snap_name += strlen(cur_snap_name) + 1) {
@@ -2966,6 +3039,7 @@ int old_snapshot_rename(cls_method_context_t hctx, bufferlist *in, bufferlist *o
     return rc;
   return 0;
 }
+
 
 namespace mirror {
 
@@ -4215,6 +4289,12 @@ void __cls_init()
   cls_register_cxx_method(h_class, "metadata_get",
                           CLS_METHOD_RD,
 			  metadata_get, &h_metadata_get);
+  cls_register_cxx_method(h_class, "snapshot_get_limit",
+			  CLS_METHOD_RD,
+			  snapshot_get_limit, &h_snapshot_get_limit);
+  cls_register_cxx_method(h_class, "snapshot_set_limit",
+			  CLS_METHOD_WR,
+			  snapshot_set_limit, &h_snapshot_set_limit);
 
   /* methods for the rbd_children object */
   cls_register_cxx_method(h_class, "add_child",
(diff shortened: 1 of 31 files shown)


src/cls/rbd/cls_rbd.cc
```diff
@@ -2675,6 +2693,51 @@ int metadata_get(cls_method_context_t hctx, bufferlist *in, bufferlist *out)
...
+		       bufferlist *out)
+{
+  int rc;
+  uint64_t snap_limit;
+
+  rc = read_key(hctx, "snap_limit", &snap_limit);
+  if (rc == -ENOENT) {
+    rc = 0;
+    ::encode(UINT64_MAX, *out);
+  } else {
```
runsisi: maybe we also need to handle `rc < 0 && RC != -ENOENT` ?

src/include/rbd/librbd.hpp
```diff
@@ -247,6 +247,8 @@ class CEPH_RBD_API Image
   int snap_is_protected(const char *snap_name, bool *is_protected);
   int snap_set(const char *snap_name);
   int snap_rename(const char *srcname, const char *dstname);
+  int get_snap_limit(uint64_t *limit);
+  int set_snap_limit(const int limit);
```
runsisi: why we need a `const` here ?

src/cls/rbd/cls_rbd.cc
fullerdj: In that case, we just pass the returned value unmodified the caller. There really isn't much else we can do if read_key fails.

src/librbd/Makefile.am
```diff
@@ -64,6 +64,7 @@ librbd_internal_la_SOURCES = \
 	librbd/operation/SnapshotRenameRequest.cc \
 	librbd/operation/SnapshotRollbackRequest.cc \
 	librbd/operation/SnapshotUnprotectRequest.cc \
+	librbd/operation/SnapshotLimitRequest.cc \
```
runsisi: `src/librbd/CMakeLists.txt` needs updated too.

src/cls/rbd/cls_rbd.cc
```diff
@@ -108,6 +108,8 @@ cls_method_handle_t h_metadata_set;
 cls_method_handle_t h_metadata_remove;
 cls_method_handle_t h_metadata_list;
 cls_method_handle_t h_metadata_get;
+cls_method_handle_t h_get_snapshot_limit;
```
dillaman: Minor: how about snapshot_<verb>_limit?

fullerdj: So you want set_snapshot_limit and snapshot_limit? Since “limit” is a verb, “snapshot_limit” implies “set” to me. Your call, though.

src/cls/rbd/cls_rbd.cc
dillaman: ```snap_limit``` would be uninitialized in that case.  I think it would be good to return the error code if "rc < 0 && rc != -ENOENT" since it is really pointing to a failure in the OSD.

dillaman: Never mind -- I read that wrong.

src/cls/rbd/cls_rbd.cc
dillaman: "limit" in this context is a noun -- "set" and "get" are the verbs

src/cls/rbd/cls_rbd.cc
```diff
@@ -2675,6 +2693,51 @@ int metadata_get(cls_method_context_t hctx, bufferlist *in, bufferlist *out)
...
+
+  CLS_LOG(20, "read snapshot limit %lu", snap_limit);
+  return rc;
+}
+
+int set_snapshot_limit(cls_method_context_t hctx, bufferlist *in,
+		       bufferlist *out)
+{
+  int rc;
+  int new_limit;
```
dillaman: It's encoded as a uint64_t on the getter and an int on the setter.  Why not just treat it as a uint64_t and if it equals UINT64_MAX you can remove it?

fullerdj: Yeah, I don't like that asymmetry either. I wanted the user to be able to specify a negative value on the command line instead of remembering UINT64_MAX and I didn't want the CLI to contain the logic to translate it. I can put it there if that's ok, though.

src/librbd/Operations.cc
```diff
@@ -1024,6 +1025,46 @@ void Operations<I>::execute_snap_unprotect(const char *snap_name,
...
+    execute_snap_limit(limit, &cond_ctx);
+
+    r = cond_ctx.wait();
+  }
+
+  return r;
+}
+
+template <typename I>
+void Operations<I>::execute_snap_limit(const uint64_t limit,
```
dillaman: Probably overly complex -- I don't think there is a need to proxy this request.  The rationale is that the limit won't affect the object map, journaling, etc (things that the exclusive lock is protecting).  I think you can just create a small, synchronous helper method to invoke the "snapshot_set_limit" cls_client helper method.

src/librbd/internal.h
```diff
@@ -128,6 +128,7 @@ namespace librbd {
 	     ProgressContext& prog_ctx);
   int snap_list(ImageCtx *ictx, std::vector<snap_info_t>& snaps);
   int snap_exists(ImageCtx *ictx, const char *snap_name, bool *exists);
+  int get_snap_limit(ImageCtx *ictx, uint64_t *limit);
```
dillaman: Minor: ```snap_get_limit``` to be consistent with the others?

src/librbd/Operations.cc
dillaman: Oh -- I see the rationale now -- for rbd-mirror replication.

src/librbd/Operations.h
```diff
@@ -58,6 +58,9 @@ class Operations {
   int snap_unprotect(const char *snap_name);
   void execute_snap_unprotect(const char *snap_name, Context *on_finish);
 
+  int set_snap_limit(const int limit);
```
dillaman: Minor: ```snap_set_limit``` and ```execute_snap_set_limit```

src/cls/rbd/cls_rbd.cc
dillaman: I'd probably vote for ```rbd snap limit set <image-spec> <limit>``` and ```rbd snap limit clear <image-spec>``` to mirror other commands.

dillaman:
Don't forget to update the C and Python APIs at the same time.

src/librbd/journal/Replay.cc
```diff
@@ -600,6 +604,32 @@ void Replay<I>::handle_event(const journal::DemoteEvent &event,
...
+    return;
+  }
+
+  op_event->on_op_finish_event = new C_RefreshIfRequired<I>(
+    m_image_ctx, new ExecuteOp<I, journal::SnapLimitEvent>(m_image_ctx,
+							   event,
+							   on_op_complete));
+
+  // ignore errors caused due to replay
+  op_event->ignore_error_codes = {-EINVAL};
```
dillaman: Is this necessary? It doesn't look like it should return -EINVAL if you replay the same set limit event.

src/librbd/operation/SnapshotLimitRequest.h
```diff
@@ -0,0 +1,65 @@
...
+   *    v
+   * STATE_LIMIT_SNAPS
+   *    |
+   *    v
+   * <finish>
+   *
+   * @endverbatim
+   *
+   */
+  enum State {
```
dillaman: No technical need for the enum and associated code since it is just a one-and-done op.

src/librbd/Operations.cc
```diff
@@ -1024,6 +1025,46 @@ void Operations<I>::execute_snap_unprotect(const char *snap_name,
...
+  }
+
+  int r = m_image_ctx.state->refresh_if_required();
+  if (r < 0) {
+    return r;
+  }
+
+  {
+    RWLock::RLocker owner_lock(m_image_ctx.owner_lock);
+    C_SaferCond cond_ctx;
```
dillaman: You will have to force-acquire the exclusive lock before executing your async method.  That will ensure that if journaling is enabled, you have the journal opened.

dillaman: ```librbd::ExclusiveLock::request_lock```

src/cls/rbd/cls_rbd.cc
```diff
@@ -1551,6 +1563,12 @@ int snapshot_add(cls_method_context_t hctx, bufferlist *in, bufferlist *out)
     if (r < 0)
       return r;
 
+    total_read += vals.size();
+    if (total_read >= snap_limit) {
+      CLS_ERR("Attempt to create snapshot over limit of %lu", snap_limit);
+      return -ERANGE;
```
dillaman: EDQUOT instead?

src/tools/rbd/action/Info.cc
```diff
@@ -150,6 +154,14 @@ static int do_show_info(const char *imgname, librbd::Image& image,
     }
   }
 
+  if (snap_limit < UINT64_MAX) {
+    if (f) {
+      f->dump_unsigned("snapshot_limit", snap_limit);
+    } else {
+      std::cout << "\tlimited to " << snap_limit << " snapshots" << std::endl;
```
dillaman: "snapshot_limit: xyz" would match the style of the other info fields.

dillaman:
Run "test/run-cli-tests" to see the changes required in "test/cli/rbd/help.t"

fullerdj:
Yeah, I noticed that yesterday. I hadn’t heard of cram before, so I’ve been reading up today.

src/cls/rbd/cls_rbd.cc
fullerdj: Interesting: GitHub munged the part of your comment between the braces when translating to email. Therefore, your comment appeared in the email to be 'Minor: how about snapshot__limit?' I took the double underscore to be a simple typo.

dillaman: I used "<" and ">" originally -- it must have thought I was trying to inject HTML.

fullerdj:
@dillaman Ok, I pushed another revision. Could you take a look when you get a chance? Thanks.

src/librbd/Operations.cc
```diff
@@ -1024,6 +1025,61 @@ void Operations<I>::execute_snap_unprotect(const char *snap_name,
...
+      auto f = [this, limit, &cond_ctx] (int r) {
+	if (r) {
+	  cond_ctx.complete(r);
+	} else {
+	  execute_snap_set_limit(limit, &cond_ctx);
+	}
+      };
+      FunctionContext *fc = new FunctionContext(f);
+
+      m_image_ctx.exclusive_lock->request_lock(fc);
```
dillaman: Minor: can we just use a C_SaferCond here since this is all synchronous.

src/test/cli/rbd/help.t
```diff
@@ -1072,6 +1074,35 @@
...
+  Limit the number of snapshots.
+  
+  Positional arguments
+    <image-spec>         image specification
+                         (example: [<pool-name>/]<image-name>)
+  
+  Optional arguments
+    -p [ --pool ] arg    pool name
+    --image arg          image name
+    --limit arg          snapshot count (negative to remove)
```
dillaman: Minor: leftover?

dillaman:
lgtm

jdurgin:
Need to add the new tracepoints to src/tracing/librbd.tp - apparently the pr bot doesn't have lttng installed so it didn't detect this

fullerdj:
Instead of src/tracing/librbd.h? That’s where I put them.

jdurgin:
src/tracing/*.h are generated from the .tp files

dillaman:
@fullerdj ping

yuriw:
package build failed: http://gitbuilder.sepia.ceph.com/gitbuilder-ceph-rpm-centos7-amd64-notcmalloc/log.cgi?log=082288711322ce052cd97fada22645ab7fad3553

src/librbd/Makefile.am
dillaman: Add 'librbd/operation/SnapshotLimitRequest.h' to ```noinst_HEADERS``` below

PR #9153, opened 2016-05-16
OpenStack setup scripts.

wusui:
A set of scripts used to bring up an OpenStack instance on one site,
connect a ceph cluster, add pools to store OpenStack glance images,
cinder volumes, nova vms, and cinder backups on rbd, and bring up a nova
compute node.


tmuthamizhan:
looks good!

PR #9156, opened 2016-05-16
cmake: Added #defines to config-h.in.cmake

alimaredia:
This PR is a cleanup of pre-processor variables that are defined using CMake. I went through the automake version of acconfig.h (http://paste.fedoraproject.org/367320/63442732) and made sure each defined variable is 1) being still being used in ceph, 2) Set somehow by CMake.

This branch is still a work in progress, I wanted to throw something over the wall and get feedback as I reach completion. This PR may be broken into several different commits over time as well.


CMakeLists.txt
```diff
@@ -321,8 +323,18 @@ if(${HAVE_BABELTRACE})
   set(WITH_BABELTRACE ${BABELTRACE_FOUND})
   set(HAVE_BABELTRACE_BABELTRACE_H ${BABELTRACE_FOUND})
   set(HAVE_BABELTRACE_CTF_EVENTS_H ${BABELTRACE_FOUND})
+  set(HAVE_BABELTRACE_CTF_ITERATOR_H ${BABELTRACE_FOUND})
 endif(${HAVE_BABELTRACE})
 
+option(DEBUG_GATHER "C_Gather debugging is enabled" OFF)
+option(HAVE_KINETIC "Kinetic Client is enables" OFF)
```
alimaredia: @jdurgin is this still necessary? I still see that it's being used in src/kv/KeyValueDB.cc

jdurgin: can get rid of it, along with the kinetic backend code - it's really just of historical interest in it's current form

CMakeLists.txt
```diff
@@ -321,8 +323,18 @@ if(${HAVE_BABELTRACE})
...
+option(HAVE_LIBZFS "LibZFS is enabled" OFF)
```
alimaredia: @liewegas for now I just set this option to OFF to let the user set the variable at the command line. Should I create a Findzfs.cmake for this?

alimaredia:
@dang @cbodley would love to have feedback from both of you

CMakeLists.txt
```diff
@@ -321,8 +323,18 @@ if(${HAVE_BABELTRACE})
...
+option(ENABLE_COVERAGE "Coverage is enabled" OFF)
+option(PG_DEBUG_REFS "PG Ref debugging is enabled" OFF)
+option(HAVE_ARMV8_CRC "Supports ARMv8 CRC instructions" OFF)
```
alimaredia: @dmick I set HAVE_ARMV8_CRC as an option and set it to off just to throw something over the wall. We have builds going on ARMv8 and should something in CMake for platform detection right?

CMakeLists.txt
```diff
@@ -330,8 +337,15 @@ if(${HAVE_BABELTRACE})
   set(WITH_BABELTRACE ${BABELTRACE_FOUND})
   set(HAVE_BABELTRACE_BABELTRACE_H ${BABELTRACE_FOUND})
   set(HAVE_BABELTRACE_CTF_EVENTS_H ${BABELTRACE_FOUND})
+  set(HAVE_BABELTRACE_CTF_ITERATOR_H ${BABELTRACE_FOUND})
 endif(${HAVE_BABELTRACE})
 
+option(DEBUG_GATHER "C_Gather debugging is enabled" OFF)
+option(HAVE_LIBZFS "LibZFS is enabled" OFF)
+option(ENABLE_COVERAGE "Coverage is enabled" OFF)
```
alimaredia: @jdurgin I wasn't aware of what ENABLE_COVERAGE did before I added this option, are we still generating coverage data from the gitbuilder tarballs?

jdurgin: no, the surrounding teuthology infrastructure to analyze it hasn't worked in a long time. Worth keeping the option to enable it if possible though.

CMakeLists.txt
```diff
@@ -330,8 +337,15 @@ if(${HAVE_BABELTRACE})
...
+option(PG_DEBUG_REFS "PG Ref debugging is enabled" OFF)
```
alimaredia: @athanatos is there anything else I need to do with regards to PG_DEBUG_REFS?

src/CMakeLists.txt
```diff
@@ -94,6 +94,17 @@ if(${WITH_PROFILER})
...
+if(${ENABLE_COVERAGE})
+  find_program(HAVE_GCOV gcov)
+  find_program(HAVE_LCOV lcov)
+  if(NOT HAVE_GCOV)
+    message(FATAL_ERROR "Coverage Enabled but gcov Not Found")
+  endif(NOT HAVE_GCOV)
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fprofile-arcs -ftest-coverage -O0")
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fprofile-arcs -ftest-coverage")
+  list(APPEND EXTRALIBS gcov)
+endif(${ENABLE_COVERAGE})
```
alimaredia: @jdurgin is this conditional good enough to ENABLE coverage for the time being?

jdurgin: yeah, looks good. checking for lcov is unnecessary

CMakeLists.txt
athanatos: @alimaredia I think that's it.  It's just an option that can be enabled to make it easier to find pg ref leaks.

src/rgw/rgw_http_client.cc
```diff
@@ -335,7 +335,7 @@ RGWHTTPClient::~RGWHTTPClient()
 }
 
 
-#if HAVE_CURL_MULTI_WAIT
+#ifdef HAVE_CURL_MULTI_WAIT
```
alimaredia: @cbodley This change was made so it would compile

cbodley: this is because you changed:
```
-#cmakedefine HAVE_CURL_MULTI_WAIT 1
+#cmakedefine HAVE_CURL_MULTI_WAIT
```

CMakeLists.txt
dmick: There is a gitbuilder, not certain of its state.  I guess you see that the existing value is set based on compiler test inside AX_ARM_FEATURES()

src/CMakeLists.txt
tchaikov: @alimaredia  agreed with Josh, it's unnecessary to check for lcov.

src/CMakeLists.txt
```diff
@@ -54,7 +54,7 @@ else(no_yasm)
     set(CMAKE_REQUIRED_QUIET ${save_quiet})
     if(not_arch_x32)
       message(STATUS " we are not x32")
-      set(HAVE_GOOD_YASM_ELF64 1)
+      set(HAVE_GOOD_YASM_ELF64 1 PARENT_SCOPE)
```
tchaikov: this breaks the build on amd64 with a capable yasm. `HAVE_GOOD_YASM_ELF64` is not defined in this scope.

src/CMakeLists.txt
```diff
@@ -646,11 +657,11 @@ add_dependencies(ceph-mon erasure_code_plugins)
 install(TARGETS ceph-mon DESTINATION bin)
 
 # OSD
-if(${HAVE_XFS})
+if(${HAVE_LIBXFS})
```
tchaikov: nit, maybe we can be more consistent by just using `if (var)` instead of `if (${var})`. simpler and easier to read. see https://cmake.org/cmake/help/v3.0/command/if.html

CMakeLists.txt
```diff
@@ -367,11 +377,17 @@ endif(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
...
 # enables testing and creates Make check command
 enable_testing()
-set(CMAKE_CTEST_COMMAND ctest -V)
+set(CMAKE_CTEST_COMMAND ctest)
 add_custom_target(check COMMAND ${CMAKE_CTEST_COMMAND})
 
 add_subdirectory(src)
 
+# Now create a useable config.h
+configure_file(
```
tchaikov: as some of the variables (`HAVE_GOOD_YASM_ELF64 ` for instance) used by the this configure_file are defined in `src/CMakeLists.txt`, maybe we can instead moving this `configure_file()` call down into `src/CMakeLists.txt`?

tchaikov: and also move the line of https://github.com/ceph/ceph/pull/9156/files#diff-af3b638bc2a3e6c650974192a53c7291R362 there. so it would be easier for future maintainer to understand our intention.

src/CMakeLists.txt
```diff
@@ -63,7 +63,7 @@ else(no_yasm)
         RESULT_VARIABLE rc
         OUTPUT_QUIET)
       if(NOT rc)
-        set(HAVE_BETTER_YASM_ELF64 1)
+        set(HAVE_BETTER_YASM_ELF64 1 PARENT_SCOPE)
         add_definitions("-DHAVE_BETTER_YASM_ELF64")
```
tchaikov: could remove this line now as it is already defined in the `acconfig.h`.

src/CMakeLists.txt
```diff
@@ -54,7 +54,7 @@ else(no_yasm)
     set(CMAKE_REQUIRED_QUIET ${save_quiet})
     if(not_arch_x32)
       message(STATUS " we are not x32")
-      set(HAVE_GOOD_YASM_ELF64 1)
+      set(HAVE_GOOD_YASM_ELF64 1 PARENT_SCOPE)
       add_definitions("-DHAVE_GOOD_YASM_ELF64")
```
tchaikov: could remove this line now as it is already defined in the `acconfig.h`.

src/CMakeLists.txt
alimaredia: @tchaikov agree! In the "CMake To Do" etherpad, making the syntax consistent is on the list, and will be a top priority when some of the more necessary functionality is working

src/CMakeLists.txt
alimaredia: @tchaikov I'm not sure I completely follow, could you elaborate on this more?

tchaikov: we have following code in `src/CMakeLists.txt`

```cmake
if(HAVE_GOOD_YASM_ELF64)
  set(yasm_srcs
    common/crc32c_intel_fast_asm.S
    common/crc32c_intel_fast_zero_asm.S)
endif(HAVE_GOOD_YASM_ELF64)
```

which checks `HAVE_GOOD_YASM_ELF64`. but if we **only** `set(HAVE_GOOD_YASM_ELF64 1 PARENT_SCOPE)`. the local `HAVE_GOOD_YASM_ELF64` is not changed. so  `yasm_srcs` will be not defined and empty. that's why i'd suggest move the variable of `HAVE_GOOD_YASM_ELF64 ` down to `src/CMakeLists.txt` without worrying about its lexical scope.

PR #9157, opened 2016-05-16
Improvements to ceph-detect-init

onyb:
* More accurate init detection in Ubuntu, RHEL, Fedora, and CentOS
* Init detection in Linux Mint and LMDE
* Added new test cases for the above
* General code refactor
* Fixed Docker integration tests


src/ceph-detect-init/integration/test_main.py
```diff
@@ -54,10 +54,10 @@ def run(os):
 class TestCephDetectInit(testtools.TestCase):
 
     def test_centos_6(self):
-        self.assertEqual('sysvinit', run('centos-6'))
+        self.assertEqual('upstart', run('centos-6'))
```
robbat2: Are you sure that's right?

onyb: Aye. CentOS 6 shipped Upstart, since it was the default init system in RHEL 6 release. I also tested this on a Docker container.
```sh
$ init --version
init (upstart 0.6.5)
```

jdurgin:
looking good to me! Glad to see you updated the docker tests too - I wasn't familar with those before. I don't think it's necessary to stay compatible with python 2.6 at this point, since the oldest platform we're building on is centos 7, but it doesn't do any harm. Just no need to go out of your way to do it.

@dachary perhaps you would like to look at this one too?

dachary:
@onyb ceph-detect-init is meant to match an os version with the init system ceph supports for that version. In other words, when a given os version supports more than one init system, ceph-detect-init will return the init system that has been tested to work. For instance CentOS 7 supports both sysvinit (legacy) and systemd (native) and ceph-detect-init returns systemd because it has the required files for the job.

dachary:
@onyb ping ?

onyb:
@dachary I had a discussion with @jdurgin about this. The results I added for specific platforms are indeed correct (default init) for the respective OS, but haven't checked if Ceph actually supported/used those init systems. Also as correctly pointed out by Josh, many platforms which I've included are too old to be useful, and can be dropped.

dachary:
@onyb that makes sense to me :-)

jdurgin:
@onyb any update on this?

dachary:
@onyb feel free to re-open when you have time to update this pull request

PR #9160, opened 2016-05-17
BitMapAllocator: make allocate() can allocate small blocks

majianpeng:


majianpeng:
@chhabaramesh , @liewegas . This PR mainly make allocate() allow allocate partly blocks rather than only contiguous wanted blocks.

liewegas:
I'm a bit confused here.. why is there a bool partly? We can *always* return fewer blocks (as long as we respect min_alloc_size).  There is never a case where we are require to find large contiguous extents.  (We should just make a best effort attempt to do that.)

liewegas:
@chhabaramesh

majianpeng:
BitMapAllocator::allocate-->BitAllocator::alloc_blocks_res-->BitAllocator::alloc_blocks_int
>> while ((zone = (BitMapZone *) iter.next())) {
 >>   if (!zone_free_to_alloc(zone,
 >>     num_blocks - allocated, wait)) {
  >>    continue;
   >> }

  >>  allocated = zone->alloc_blocks(num_blocks, start_block);

>>    zone->unlock_zone();
>>    if (allocated == num_blocks) {
>>      break;
>>    }

>>    zone->free_blocks(*start_block, allocated);
>>    allocated = 0;
>  }
From that, we can find alloc_blocks_int() only return N contiguous blocks or zero block.

liewegas:
@chhabaramesh can you take a look?  also this needs a rebase?

chhabaramesh:
@liewegas  and @majianpeng  I will review it today. BTW, I am doing the change where we can return fewer blocks and return them in extent. For that we will use api alloc_block_dis that is revised to allocate extents.

chhabaramesh:
since calling alloc_blocks again and again to get desired set of blocks could be costly, the actual change will be to use the alloc_blocks_dis to allocate all required length in multiple extents and in single call to BitAllocator. That will remove the loop from BlueStore.cc and BlueFS.cc allocation logic.

chhabaramesh:
@liewegas  @majianpeng  .Somehow I am not getting notifications from git and I miss where my attention is needed. This PR is > 28 days old and I got notification today :(.

src/os/bluestore/BitMapAllocator.cc
```diff
@@ -64,7 +64,7 @@ int BitMapAllocator::reserve(uint64_t need)
     " total " << m_bit_alloc->size() << dendl;
 
   if (m_bit_alloc->reserve_blocks(nblks)) {
-    return ENOSPC;
+    return -ENOSPC;
```
chhabaramesh: @majianpeng  looks good.

src/os/bluestore/BitMapAllocator.cc
```diff
@@ -107,7 +107,7 @@ int BitMapAllocator::allocate(
 
   count = m_bit_alloc->alloc_blocks_res(nblks, &start_blk);
   if (count == 0) {
-    return ENOSPC;
+    return -ENOSPC;
```
chhabaramesh: @majianpeng  looks good.

PR #9161, opened 2016-05-17
osd: optimize header file dependency to decreace compile time when modify any file

cxwshawn:
1、hmm, everytime modify any file, then compile, then wait&wait&wait, so trying to optimize
    this part, wish to decrease header dependency, try to improve compile speed.
optimization include OSD, PG, PGBackend layer.

2、 I am wandering whether OSD and OSDService can be separated into two single file,
    so other file don't need to ref osd.h, osd.h changed will not affect too much.


liewegas:
needs rebase

PR #9165, opened 2016-05-17
rgw: misc fix for test_multi.py

mikulely:
rgw: misc fix for test_multi.p

remove redundent meta_checkpoint, cause realm.create_user(user) will
call meta_checkpoint by default(via wait_meta = True).

### DIFF
--- a/src/test/rgw/test_multi.py
+++ b/src/test/rgw/test_multi.py
@@ -507,8 +507,6 @@ class RGWMulti:
         user = RGWUser('tester', '"Test User"', gen_access_key(), gen_secret())
         realm.create_user(user)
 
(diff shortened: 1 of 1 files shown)


src/test/rgw/test_multi.py
```diff
@@ -468,6 +468,7 @@ def gen_secret():
 
 def gen_bucket_name():
     global num_buckets
+    global run_prefix
```
cbodley: I don't think this needs to be declared global, since `gen_bucket_name()` isn't modifying `run_prefix`

mikulely: @cbodley  Just want to make reference explicitly. your concerns make sense to me. updated.

PR #9168, opened 2016-05-17
Pipe: take a ref to existing while we are waiting

athanatos:
Otherwise, if it is reaped while we are waiting, it'll be a
use-after-free.

Fixes: http://tracker.ceph.com/issues/15870


athanatos:
Passed tests: sjust@teuthology:/a/samuelj-2016-05-16_20:02:21-rados-wip-sam-testing-distro-basic-smithi

yuyuyu101:
cool!

PR #9173, opened 2016-05-18
rbd-mirror: potential crash during image status update

dillaman:


src/tools/rbd_mirror/ImageReplayer.cc
```diff
@@ -542,33 +542,17 @@ void ImageReplayer<I>::on_start_fail_finish(int r)
...
-      dout(20) << "start interrupted" << dendl;
-      m_state = STATE_STOPPED;
-      m_stop_requested = false;
+    if (r < 0) {
+      derr << "start failed: " << cpp_strerror(r) << dendl;
     } else {
-      assert(m_state == STATE_STARTING);
-      dout(20) << "start failed" << dendl;
-      m_state = (r < 0) ? STATE_UNINITIALIZED : STATE_STOPPED;
+      dout(20) << "start interrupted" << dendl;
```
trociny: Minor: Previous this message was for the case, when stop is called before start is complete (on_start_fail_start(-EINTR)). Now it is for the case when bootstrap returns "remote image is non-primary or local image is primary". It would be nice still to be able to distinguish interrupted start from other errors.

src/tools/rbd_mirror/ImageReplayer.cc
```diff
@@ -1237,10 +1194,36 @@ void ImageReplayer<I>::start_update_status_task()
...
+  m_local_ioctx.close();
+  m_remote_ioctx.close();
+
+  if (on_start != nullptr) {
+    dout(20) << "on start finish complete, r=" << r << dendl;
+    on_start->complete(r);
+  }
+  if (on_stop != nullptr) {
+    dout(20) << "on stop finish complete, r=" << r << dendl;
+    on_stop->complete(0);
```
trociny: on_stop is always completed with r=0. It is ok when handle_stop() is called from on_start_fail_finish() (this is previous and expected behaviour, though in the old code the debug message looked wrong, it should have not printed r).

When handle_stop() is called from on_stop_local_image_close_finish() this looks not correct (we lost error code on failed stop).

src/tools/rbd_mirror/ImageReplayer.cc
```diff
@@ -704,6 +703,7 @@ void ImageReplayer<I>::on_stop_local_image_close_finish(int r)
     std::swap(m_on_stop_finish, on_finish);
   }
 
+  update_mirror_image_status(true, STATE_STOPPED);
```
trociny: At this stage m_state_desc is not cleared (it is cleared in handle_stop), so I observe after stop:

```
zhuzha:/tmp/tmp.rbd_mirror% ceph --admin-daemon /tmp/tmp.rbd_mirror/rbd-mirror.cluster1_daemon.cluster1.asok rbd mirror stop
zhuzha:/tmp/tmp.rbd_mirror% rbd --cluster cluster1 -p mirror mirror pool status --verbose
health: OK
images: 2 total
    2 stopped

test:
  global_id:   8490718b-edf5-4e7a-89d0-1760e7171b6b
  state:       up+stopped
  description: bootstrapping, CLOSE_REMOTE_IMAGE
  last_update: 2016-05-18 14:41:03

test1:
  global_id:   0acf3ba3-7305-4405-9efd-872246344624
  state:       up+stopped
  description: bootstrapping, CLOSE_REMOTE_IMAGE
  last_update: 2016-05-18 14:41:03
```

src/tools/rbd_mirror/ImageReplayer.cc
dillaman: Are you just saying that the -EINTR doesn't clear the "remote image is non-primary..." message?

trociny: I mean, that previously, when I did:
```
start
stop (before start is completed)
```
I got "start interrupted" message in the log.

Now I see "start interrupted" message in the secondary mirror log periodically on every its try to bootstrap (for "remote image is non-primary or local image is primary" case).

I just say that the behavior is changed, personally I liked more as it was before, but I have no strong opinion here.

dillaman: How would I reproduce this?

trociny: 1) run `unittest_rbd_mirror  --gtest_filter=TestImageReplayer.StartInterrupted` with debug log level. Now we have in logs:
```
...::on_start_fail_finish: start failed: (4) Interrupted system call
```
Previously we had:
```
...::on_start_fail_finish: start interrupted
```
2) Run rbd_mirror.sh test. Look at the log for the secondary mirror. You will see
```
...::on_start_fail_finish: start interrupted
```
periodically. Previously we did not have these messages here.

dillaman: ACK -- I thought you were talking about the status description, not the debug log message.

trociny:
@dillaman I am observing status not beeing updated for long time (last_update is not changed). I run
```
RBD_MIRROR_NOCLEANUP=1 RBD_MIRROR_TEMDIR=/tmp/tmp.rbd_mirror ../qa/workunits/rbd/rbd_mirror.sh
```
Then monitor the status in /tmp/tmp.rbd_mirror dir:
```
rbd --cluster cluster1 -p mirror mirror pool status --verbose
```
I will provide more details if you don't see this issue.

dillaman:
On Wed, May 18, 2016 at 8:49 AM, Mykola Golub <notifications@github.com>
wrote:

> I will provide more details if you don't see this issue.
>
>
I was actually looking into that right now -- I noticed that I had a stuck
"entries_behind_master" when double-checking my changes.


-- 
Jason

dillaman:
@trociny Hopefully all better now

src/tools/rbd_mirror/ImageReplayer.cc
```diff
@@ -542,44 +553,34 @@ void ImageReplayer<I>::on_start_fail_finish(int r)
...
-    return false;
+  {
+    Mutex::Locker locker(m_lock);
+    assert(m_state == STATE_STARTING);
+    if (m_on_stop_finish == nullptr) {
+      return false;
+    }
   }
 
+  set_state_description(0, "");
```
trociny: Do you really need this here? `on_start_fail_start(-EINTR)` below will call `set_state_description(-EINTR, "")`. The old code looks like still should work and is simpler due to avoiding the lock dance.

src/tools/rbd_mirror/ImageReplayer.cc
```diff
@@ -1050,197 +1036,265 @@ void ImageReplayer<I>::shut_down_journal_replay(bool cancel_ops)
...
+  m_local_ioctx.close();
+  m_remote_ioctx.close();
+
+  if (on_start != nullptr) {
+    dout(20) << "on start finish complete, r=" << r << dendl;
+    on_start->complete(r);
+  }
+  if (on_stop != nullptr) {
+    dout(20) << "on stop finish complete, r=" << r << dendl;
+    on_stop->complete(on_start == nullptr ? r : 0);
```
trociny: minor: it might still be confusing to see in the debug log message `r`, while complete is actually called with 0.

src/tools/rbd_mirror/ImageReplayer.cc
```diff
@@ -1050,197 +1036,265 @@ void ImageReplayer<I>::shut_down_journal_replay(bool cancel_ops)
...
+
+  // if a deferred update is available, send it -- otherwise reschedule
+  // the timer task
+  if (is_running()) {
+    bool started = false;
+    if (update_status_requested) {
       Mutex::Locker locker(m_lock);
-      std::swap(m_update_status_comp, comp);
+      started = start_mirror_image_status_update(false, true);
     }
```
trociny: I don't like very much we take (and drop) the lock 3 times here (to get update_status_requested, to check is_running, and to start_mirror_image_status_update). Although it looks like race is not possible, it is still alarming and could be avoided (take the lock only once) if use `is_running_()`?

src/tools/rbd_mirror/ImageReplayer.cc
```diff
@@ -1050,197 +1036,265 @@ void ImageReplayer<I>::shut_down_journal_replay(bool cancel_ops)
...
 }
 
 template <typename I>
-void ImageReplayer<I>::reschedule_update_status_task(int new_interval)
-{
-  Mutex::Locker locker(m_threads->timer_lock);
+void ImageReplayer<I>::reschedule_update_status_task(int new_interval) {
+  dout(20) << dendl;
+  m_lock.Lock();
+  m_threads->timer_lock.Lock();
```
trociny: Just wander was some reasons to not use Mutex::Locker here?

src/tools/rbd_mirror/ImageReplayer.cc
dillaman: Good point -- missed that.

src/tools/rbd_mirror/ImageReplayer.cc
dillaman: It was just to avoid indenting the full method within braces -- there are no exit points.

dillaman:
@trociny updated

PR #9174, opened 2016-05-18
OSD: cleanup code for ms_dispatch

majianpeng:


majianpeng:
@tchaikov . Hope your comments.

src/osd/OSD.cc
```diff
@@ -5864,7 +5864,6 @@ bool OSD::ms_dispatch(Message *m)
 
   do_waiters();
   _dispatch(m);
-  do_waiters();
```
liewegas: handle_osd_map drops osd_lock, so it's possible for new waiters to be queued.

liewegas:
I think we still need the first and last patches, since _dispatch drops osd_lock in handle_osd_map (at least)

majianpeng:
@liewegas . I'm not find osd_lock.Unlock in handle_osd_map or the related path. Could you point the place? Thanks very much!.

majianpeng:
@tchaikov . I'm not find osd_lock.Unlock in handle_osd_map or the related path. Could you point the place?  Thanks very much!

tchaikov:
@majianpeng neither can i find the `osd_lock.Unlock()` @liewegas mentioned. we do not drop the osd_lock since 492e6351e5723f7d97defb01d66a3505fd33a82c. maybe i am missing something. but after reviewing the code couple times, no luck.

majianpeng:
@tchaikov . Thanks very much.

src/osd/OSD.cc
```diff
@@ -5856,19 +5847,10 @@ bool OSD::ms_dispatch(Message *m)
...
   }
 
-  while (dispatch_running) {
-    dout(10) << "ms_dispatch waiting for other dispatch thread to complete" << dendl;
-    dispatch_cond.Wait(osd_lock);
-  }
-  dispatch_running = true;
-
-  do_waiters();
+   do_waiters();
```
tchaikov: bad indent.

src/osd/OSD.cc
```diff
@@ -5856,19 +5847,10 @@ bool OSD::ms_dispatch(Message *m)
...
   _dispatch(m);
 
-  dispatch_running = false;
-  dispatch_cond.Signal();
-
-  osd_lock.Unlock();
+   osd_lock.Unlock();
```
tchaikov: wrong indent.

liewegas:
You're right--there is no more unlock.  I was looking at an older checkout.

majianpeng:
@tchaikov  update and add 601f25b. please review. Thanks!

src/osd/OSD.cc
tchaikov: @majianpeng probably we can have an assert here to make sure that the `finished` queue is empty?

src/osd/OSD.cc
```diff
@@ -6070,15 +6070,19 @@ void OSD::do_waiters()
   assert(osd_lock.is_locked());
 
   dout(10) << "do_waiters -- start" << dendl;
-  finished_lock.Lock();
-  while (!finished.empty()) {
-    OpRequestRef next = finished.front();
-    finished.pop_front();
-    finished_lock.Unlock();
-    dispatch_op(next);
+  //take_waiters be called in the status of osd_lock is locked.
```
tchaikov: nit: could add a space after `//`

src/osd/OSD.cc
```diff
@@ -6070,15 +6070,19 @@ void OSD::do_waiters()
...
+  //And !finished.empty() is very rare situation.
+  if (!finished.empty()) {
```
tchaikov: and do early return to reduce the indent level and have less code churn.

majianpeng:
@tchaikov . update

liewegas:
lgtm.  @athanatos do you have any concerns here?  It doesn't seem like we'll want to reintroduce an osd_lock drop.

tchaikov:
lgtm also

src/osd/OSD.cc
```diff
@@ -6089,6 +6071,11 @@ void OSD::do_waiters()
   assert(osd_lock.is_locked());
 
   dout(10) << "do_waiters -- start" << dendl;
+  // take_waiters be called in the status of osd_lock is locked.
+  // And !finished.empty() is very rare situation.
+  if (finished.empty())
```
athanatos: Nothing I have found suggests that std::list::empty() is safe to call with a racing write.  I think this probably invites a rare crash.

src/osd/OSD.cc
```diff
@@ -5842,18 +5833,9 @@ bool OSD::ms_dispatch(Message *m)
...
-  }
-  dispatch_running = true;
-
   do_waiters();
   _dispatch(m);
-  do_waiters();
-
-  dispatch_running = false;
-  dispatch_cond.Signal();
+  assert(finished.empty());
```
athanatos: Same here, I don't have any reason to think that std::list::empty is safe to call with a racing write.

athanatos:
This seems promising to me!

majianpeng:
@athanatos . Like the comments, OSD::take_waites be called when osd_lock is locked. So those the check of finished.empty()  is the safe.

athanatos:
@majianpeng Ah, I had assumed that it was also called from PG threads (there's an unrelated PG::take_waiters).  In that case, can we also kill finished_lock?

majianpeng:
@athanatos  update.

athanatos:
Code looks good, please update your commit messages to have more descriptive first lines.

As an example, the first commit could be:

osd/OSD.cc::ms_dispatch: no need to call do_waiters again after _dispatch

The second could be:

osd/OSD: remove unused take_waiter and take_waiters_front

The third could be

osd/OSD: remove dispatch_running and supporting code

The last can be

osd/OSD: remove now unnecessary finished_lock

When possible, it's good to make sure that the first line is descriptive since that's what we can see at a glance in gitk.  As a rule of thumb, "optimize" is almost never useful as a description.

majianpeng:
@athanatos . update . Thanks your suggestions

tchaikov:
@athanatos ping?

tchaikov:
lgtm.

tchaikov:
tested at 

- http://pulpito.ceph.com/kchai-2016-06-04_19:52:10-rados-wip-kefu-testing---basic-mira/
- http://pulpito.ceph.com/kchai-2016-06-04_09:44:21-rados-wip-kefu-testing---basic-mira/
- http://pulpito.ceph.com/kchai-2016-06-04_05:52:02-rados-wip-kefu-testing---basic-mira/

the failures are caued by environmental problem, test suite issues [1], and teuthology issue[2]

--
[1] https://github.com/ceph/ceph-qa-suite/pull/1031 
[2] http://tracker.ceph.com/issues/16142

PR #9175, opened 2016-05-18
os/bluestore: If using sync_wal_apply, no need start wal_tp thread.

majianpeng:


majianpeng:
@liewegas . hope your comments!

liewegas:
Did you verify the ThreadPool behaves when passed 0 threads?

majianpeng:
@liewegas . For 0 threads, nothing can do. In the end it  killed by suicide timed out.

I try to implement dynamic change wal_apply mode(sync/async). But this will change the order. So i add 	a7a43a5 which don't support change wal_apply mode in running state.

liewegas:
This patch set looks right to me.   Can you rebsae on current master please?

PR #9176, opened 2016-05-18
rbd-mirror: explicitly disable status updates on stopping replay

trociny:
Otherwise, there is still a chance of delayed update being fired by a
working thread after m_local_ioctx is destroyed.

Fixes: http://tracker.ceph.com/issues/15917


trociny:
It is fixed by #9173

PR #9178, opened 2016-05-18
build: freebsd changes, part 2

wjwithagen:
This requires Clang to be at least 3.7
It compiles all ceph code (expect the ones excluded in autogen_freebsd.sh)
It compiles all tests

Tests will run, but not all complete successfully, 
and make check will stall at unittest_chain_xattr.

Also note that there are a lot of warnings (I'm preparing a separate pull for that)
 - struct <-> class
 - unused or uninitialized 
 - some more Clang pickyness

### DIFF
--- a/src/common/Graylog.cc
+++ b/src/common/Graylog.cc
@@ -85,7 +85,7 @@ void Graylog::log_entry(Entry const * const e)
     m_formatter->dump_string("short_message", s);
     m_formatter->dump_string("_app", "ceph");
     m_formatter->dump_float("timestamp", e->m_stamp.sec() + (e->m_stamp.usec() / 1000000.0));
-    m_formatter->dump_int("_thread", e->m_thread);
+    m_formatter->dump_unsigned("_thread", (uint64_t)e->m_thread);
     m_formatter->dump_int("_level", e->m_prio);
     if (m_subs != NULL)
     m_formatter->dump_string("_subsys_name", m_subs->get_name(e->m_subsys));
--- a/src/common/Thread.cc
+++ b/src/common/Thread.cc
@@ -12,6 +12,7 @@
  *
  */
 
+#include "include/compat.h"
 #include "common/Thread.h"
 #include "common/code_environment.h"
 #include "common/debug.h"
@@ -22,6 +23,7 @@
 #include <errno.h>
 #include <iostream>
 #include <pthread.h>
+
 #include <signal.h>
 #include <sstream>
 #include <stdlib.h>
--- a/src/include/compat.h
+++ b/src/include/compat.h
@@ -13,8 +13,56 @@
 #define CEPH_COMPAT_H
 
 #if defined(__FreeBSD__)
+
+/* Make sure that ENODATA is defined in the correct way */
+#ifndef ENODATA
 #define	ENODATA	ENOATTR
+#else
+#if (ENODATA == 9919)
+// #warning ENODATA already defined to be 9919, redefining to fix
+// Silencing this warning because it fires at all files where compat.h
+// is included after boost files.
+//
+// This value stems from the definition in the boost library
+// And when this case occurs it is due to the fact that boost files
+// are included before this file. Redefinition might not help in this
+// case since already parsed code has evaluated to the wrong value.
+// This would warrrant for d definition that would actually be evaluated
+// at the location of usage and report a possible confict.
+// This is left up to a future improvement
+#elif (ENODATA != 87)
+#warning ENODATA already defined to a value different from 87 (ENOATRR), refining to fix
+#endif
+#undef ENODATA
+#define ENODATA ENOATTR
+#endif
+
+#ifndef MSG_MORE
 #define	MSG_MORE 0
+#endif
+
+#ifndef O_DSYNC
+#define O_DSYNC O_SYNC
+#endif
+
+#ifndef HOST_NAME_MAX
+#define HOST_NAME_MAX 64
+#endif
+
+// Fix clock accuracy
+#if !defined(CLOCK_MONOTONIC_COARSE)
+#if defined(CLOCK_MONOTONIC_FAST)
+#define CLOCK_MONOTONIC_COARSE CLOCK_MONOTONIC_FAST
+#else
+#define CLOCK_MONOTONIC_COARSE CLOCK_MONOTONIC
+#endif
+#endif
+
+/* Fix a small name diff */
+#define pthread_setname_np pthread_set_name_np
+/* And include the extra required include file */
+#include <pthread_np.h>
+
 #endif /* !__FreeBSD__ */
 
 #if defined(__APPLE__)
--- a/src/rgw/rgw_file.h
+++ b/src/rgw/rgw_file.h
@@ -802,7 +802,10 @@ namespace rgw {
 			     const uint32_t flags = RGWFileHandle::FLAG_NONE) {
       using std::get;
 
-      LookupFHResult fhr { nullptr, RGWFileHandle::FLAG_NONE };
+      // cast int32_t(RGWFileHandle::FLAG_NONE) due to strictness of Clang 
+      // the cast transfers a lvalue into a rvalue  in the ctor
+      // check the commit message for the full details
+      LookupFHResult fhr { nullptr, uint32_t(RGWFileHandle::FLAG_NONE) };
 
       /* mount is stale? */
       if (state.flags & FLAG_CLOSED)
--- a/src/common/HeartbeatMap.cc
+++ b/src/common/HeartbeatMap.cc
@@ -17,6 +17,8 @@
 #include <sys/stat.h>
 #include <fcntl.h>
 #include <errno.h>
+#include <pthread.h>
+#include <signal.h>
 
 #include "HeartbeatMap.h"
 #include "ceph_context.h"
--- a/src/common/RWLock.h
+++ b/src/common/RWLock.h
@@ -41,6 +41,7 @@ public:
   RWLock(const std::string &n, bool track_lock=true, bool ld=true, bool prioritize_write=false)
     : name(n), id(-1), nrlock(0), nwlock(0), track(track_lock),
       lockdep(ld) {
+#if defined(PTHREAD_RWLOCK_PREFER_WRITER_NONRECURSIVE_NP)
     if (prioritize_write) {
       pthread_rwlockattr_t attr;
       pthread_rwlockattr_init(&attr);
@@ -50,7 +51,10 @@ public:
       pthread_rwlockattr_setkind_np(&attr,
           PTHREAD_RWLOCK_PREFER_WRITER_NONRECURSIVE_NP);
       pthread_rwlock_init(&L, &attr);
-    } else {
+    } else 
+#endif 
+    // Next block is in {} to possibly connect to the above if when code is used.
+    {
       pthread_rwlock_init(&L, NULL);
     }
     ANNOTATE_BENIGN_RACE_SIZED(&id, sizeof(id), "RWLock lockdep id");
--- a/src/common/WorkQueue.cc
+++ b/src/common/WorkQueue.cc
@@ -95,7 +95,9 @@ void ThreadPool::worker(WorkThread *wt)
   
   std::stringstream ss;
   char name[16] = {0};
+#if !defined(__FreeBSD__)
   pthread_getname_np(pthread_self(), name, sizeof(name));
+#endif
   ss << name << " thread " << name;
   heartbeat_handle_d *hb = cct->get_heartbeat_map()->add_worker(ss.str(), pthread_self());
 
@@ -300,7 +302,9 @@ void ShardedThreadPool::shardedthreadpool_worker(uint32_t thread_index)
 
   std::stringstream ss;
   char name[16] = {0};
+#if !defined(__FreeBSD__)
   pthread_getname_np(pthread_self(), name, sizeof(name));
+#endif
   ss << name << " thread " << name;
   heartbeat_handle_d *hb = cct->get_heartbeat_map()->add_worker(ss.str(), pthread_self());
 
--- a/src/common/obj_bencher.cc
+++ b/src/common/obj_bencher.cc
@@ -15,6 +15,7 @@
  * try and bench on a pool you don't have permission to access
  * it will just loop forever.
  */
+#include "include/compat.h"
 #include "common/Cond.h"
 #include "obj_bencher.h"
 
--- a/src/crush/CrushLocation.cc
+++ b/src/crush/CrushLocation.cc
@@ -1,6 +1,7 @@
 // -*- mode:C++; tab-width:8; c-basic-offset:2; indent-tabs-mode:t -*-
 // vim: ts=8 sw=2 smarttab
 
+#include "include/compat.h"
 #include "CrushLocation.h"
 #include "CrushWrapper.h"
 #include "common/config.h"
--- a/src/global/signal_handler.cc
+++ b/src/global/signal_handler.cc
@@ -92,8 +92,10 @@ static void handle_fatal_signal(int signum)
   // presumably dump core-- will handle it.
   char buf[1024];
   char pthread_name[16] = {0}; //limited by 16B include terminating null byte.
+#if !defined(__FreeBSD__)
   int r = pthread_getname_np(pthread_self(), pthread_name, sizeof(pthread_name));
   (void)r;
+#endif
 #if defined(__sun)
   char message[SIG2STR_MAX];
   sig2str(signum,message);
--- a/src/include/cephfs/libcephfs.h
+++ b/src/include/cephfs/libcephfs.h
@@ -15,7 +15,9 @@
 #ifndef CEPH_LIB_H
 #define CEPH_LIB_H
 
+#if defined(__linux__)
 #include <features.h>
+#endif
 #include <utime.h>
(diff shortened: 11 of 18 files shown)


src/common/Graylog.cc
```diff
@@ -85,7 +85,7 @@ void Graylog::log_entry(Entry const * const e)
     m_formatter->dump_string("short_message", s);
     m_formatter->dump_string("_app", "ceph");
     m_formatter->dump_float("timestamp", e->m_stamp.sec() + (e->m_stamp.usec() / 1000000.0));
-    m_formatter->dump_int("_thread", e->m_thread);
+    m_formatter->dump_int("_thread", (int64_t)e->m_thread);
```
liewegas: seems like this should really be dump_unsigned and uint64_t, since it's a pointer

src/include/compat.h
```diff
@@ -13,10 +13,57 @@
 #define CEPH_COMPAT_H
 
 #if defined(__FreeBSD__)
+
+/* Make sure that ENODATA is defined in the correct way */
+#ifndef ENODATA
 #define	ENODATA	ENOATTR
+#else
+#if (ENODATA == 9919)
+// #warning ENODATA already defined to be 9919, redefining to fix
```
liewegas: why is this commented out?

liewegas:
a few comments, but otherwise lgtm!

src/include/compat.h
wjwithagen: On 18-5-2016 22:19, Sage Weil wrote:
> In src/include/compat.h
> <https://github.com/ceph/ceph/pull/9178#discussion_r63774437>:
> 
>>  #define	ENODATA	ENOATTR
>> +#else
>> +#if (ENODATA == 9919)
>> +// #warning ENODATA already defined to be 9919, redefining to fix
> 
> why is this commented out?

I had it active in first, but it blasts so much warnings in a lot of
source files where the boost code is used as well. And it is sort of a
warning overflow (death by warnings :) ). I think it deserves to be
turned on when all references to compat.h are moved to the start of the
code files.

And then it deserves a rewrite to make it a compile check that is done
inline at the location of reference, not generating runtime overhead. I
have to admit that I've never done such a thing.

But it is there to remind the actual origin of the trouble.

--WjW

wjwithagen:
@liewegas fixed and rebased

liewegas:
@tchaikov care to review as well?

tchaikov:
sure! will do.

src/common/Thread.cc
```diff
@@ -22,6 +22,13 @@
 #include <errno.h>
 #include <iostream>
 #include <pthread.h>
+#if defined(__FreeBSD__)
+// pthread non-POSIX stuff is in a separate include file
```
tchaikov: minor, i think this condition macro is self-explained, so no need to have these comments?

src/rgw/rgw_file.h
```diff
@@ -802,7 +802,10 @@ namespace rgw {
 			     const uint32_t flags = RGWFileHandle::FLAG_NONE) {
       using std::get;
 
-      LookupFHResult fhr { nullptr, RGWFileHandle::FLAG_NONE };
+      // This seems like a Clang compiler error.
+      // It result in irgw::RGWFileHandle::FLAG_NONE being undefined during linking.
+      // LookupFHResult fhr { nullptr, RGWFileHandle::FLAG_NONE };
+      LookupFHResult fhr { nullptr, 0 };
```
tchaikov: @wjwithagen i tried with a simple example like
```c++
/// hello.cc
#include <cstdint>

struct Foo
{
  static constexpr uint32_t FOO = 0;
};

struct Bar
{
  int foo(const uint32_t flags = Foo::FOO) {
    return 0;
  }
};
```

```c++
// world.cc
#include "hello.cc"

int main()
{
  Bar b;
  b.foo();
}
```

```makefile
# Makefile
all:
    clang++-3.7 -std=c++11 -O0 -c hello.cc
    clang++-3.7 -std=c++11 -O0 -g world.cc -lm hello.o -o world
```

it compiles fine. could you show me a reproducer?

src/include/compat.h
tchaikov: @wjwithagen i am not following you.. this does not incur runtime overhead. it's compile-time overhead, iiuc.

src/include/compat.h
```diff
@@ -13,10 +13,57 @@
...
+#endif
+
+/* Fix a small name diff */
+#define pthread_setname_np pthread_set_name_np
+/* And include the extra required include file */
+#include <pthread_np.h>
+
 #endif /* !__FreeBSD__ */
 
+// Fix clock accuracy
```
tchaikov: maybe we should remove following `if defined` block? as glibc has `CLOCK_MONOTONIC_COARSE` on Linux?

src/include/compat.h
wjwithagen: @tchaikov 
I think you are refering to the text in the discussion, not the actual commit?
The current commit is all done during initial parsing by the pre-processor, nothing special there.
However: The amount of missery that will occur when ENODATA is actually the wrong value is rather serious. And as such it would IMHO a compile time check when the definition is actually used.
But that does require a more serious macro to be writting, which needs to be expanded upon reference, not upon inclusion.
But that is left for (far) future work, perhaps it is best to ignore that.

src/rgw/rgw_file.h
wjwithagen: I'll have to produce it, which require quite a lot of recompilation. I'll post when I have it.
But it is a rather convoluted error since that same const value is referenced 2 lines earlier, and that does not result in a unknown variable during linking. So it is really fishy.

You are running Clang++-3.7? That is the Linux variant? If so in which distro?

src/include/compat.h
wjwithagen: I can, but I've left it there to please other OSes, like Darwin/osX??

src/rgw/rgw_file.h
tchaikov: i have a debian sid box. and it offers up to clang 3.9 see https://packages.debian.org/search?keywords=clang&searchon=names&suite=unstable&section=all

wjwithagen: @tchaikov interesting I'll see if I can start a debian VM to try that Clang.

wjwithagen: @tchaikov The Clang 3.8 output without the rewite for FLAG_NONE is:
libtool: link: clang++ -Wall -Wtype-limits -Wignored-qualifiers -Winit-self -Wpointer-arith -Werror=format-security -fno-strict-aliasing -fsigned-char -ftemplate-depth-1024 -Wnon-virtual-dtor -Wno-invalid-offsetof -g -DGTEST_USE_OWN_TR1_TUPLE=1 -I/usr/local/include -Wno-unused-function -Wno-unused-local-typedef -Wno-inconsistent-missing-override -Wno-unused-private-field -Wno-varargs -Wno-gnu-designator -Wno-mismatched-tags -std=gnu++11 -g -o .libs/ceph_rgw_multiparser rgw/rgw_multiparser.o -Wl,--export-dynamic  -L/usr/local/lib ./.libs/librgw.so /usr/srcs/Ceph/work/ceph/src/.libs/librados.so ./.libs/librados.so ./.libs/libcls_rgw_client.a ./.libs/libcls_log_client.a ./.libs/libcls_statelog_client.a ./.libs/libcls_timeindex_client.a ./.libs/libcls_user_client.a ./.libs/libcls_replica_log_client.a ./.libs/libcls_lock_client.a ./.libs/libcls_refcount_client.a ./.libs/libcls_version_client.a -lcurl -lexpat -lfcgi /usr/local/lib/libldap.so ./.libs/libglobal.a ./.libs/libcommon.a -lboost_thread -latomic_ops -lboost_random -luuid -lpthread -lcryptopp -lm -lexecinfo -lboost_iostreams -lboost_system -Wl,-rpath -Wl,/usr/local/lib
./.libs/librgw.so: undefined reference to `rgw::RGWFileHandle::FLAG_NONE'
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
Makefile:16057: recipe for target 'ceph_rgw_multiparser' failed

tchaikov: thank you @wjwithagen ! by following the error message, i managed to create a reproducer and have a fix for it.
```c++
#include <cstdint>
#include <tuple>

struct Foo
{
  static constexpr uint32_t NONE = 0;
  typedef std::tuple<uint32_t> Tuple;
  void func() {
    Tuple t{NONE};
  }
};
  
int main()
{
  Foo().func();
}
```

because the constructor of `std::tuple<uint32_t>` always passes its parameters by reference if possible and the values of its parameters are evaluated, `Foo::NONE` is ODR-used[1]. but `Foo::NONE` out side of the class is not found anywhere. this fails the linker. so we can convert this lvalue to rvalue manually:

```diff
--- a/test.cc	2016-05-22 01:24:36.000000000 +0800
+++ b/test.cc	2016-05-22 01:24:43.000000000 +0800
@@ -7,7 +7,7 @@
   static constexpr uint32_t NONE = 0;
   typedef std::tuple<uint32_t> Tuple;
   void func() {
-    Tuple t{NONE};
+    Tuple t{uint32_t(NONE)};
   }
 };
```
so i don't think it is a "Clang compiler error". probably it's GCC who is too permissive.

@wjwithagen does this make sense to you? if so, maybe we can change our code in this way also to fix the issue.

---
[1] http://en.cppreference.com/w/cpp/language/definition

src/include/compat.h
tchaikov: i'd prefer limit this PR only to FreeBSD/clang related fixes.

src/rgw/rgw_file.h
wjwithagen: @tchaikov I really needed to read up on this, and then read it again to get the full grasp.
It really looks like this. 

But what I do not understand is that there are more references to RGWFileHandle::FLAG_NONE ( for instance a few lines earlier ), which I suspect are cancelled out, because of the const character of the variable. Where this line does not get the same treatment, and we end up with an undeclared var...
The actual definition looks like:
```
class class RGWFileHandle {
  static constexpr uint32_t FLAG_NONE =    0x0000;
}
```
And per your link-text, that static declaration would require an explicit definition.
But that sort of surpasses the constexpr declaration.

Interested in what your suggestion for a fix is.

tchaikov: > But what I do not understand is that there are more references to RGWFileHandle::FLAG_NONE ( for instance a few lines earlier ),

yeah because `constexp` *and* passing by rvalue.

> And per your link-text, that static declaration would require an explicit definition.
> But that sort of surpasses the constexpr declaration.

please note the difference of ODR-use and non ODR-use.

> Interested in what your suggestion for a fix is.

```diff
-      LookupFHResult fhr { nullptr, RGWFileHandle::FLAG_NONE };
+      LookupFHResult fhr { nullptr, uint32_t(RGWFileHandle::FLAG_NONE) };
```

wjwithagen: On 23-5-2016 04:18, Kefu Chai wrote:
> In src/rgw/rgw_file.h
> <https://github.com/ceph/ceph/pull/9178#discussion_r64162856>:
> 
>> @@ -802,7 +802,10 @@ namespace rgw {
>>  			     const uint32_t flags = RGWFileHandle::FLAG_NONE) {
>>        using std::get;
>>  
>> -      LookupFHResult fhr { nullptr, RGWFileHandle::FLAG_NONE };
>> +      // This seems like a Clang compiler error.
>> +      // It result in irgw::RGWFileHandle::FLAG_NONE being undefined during linking.
>> +      // LookupFHResult fhr { nullptr, RGWFileHandle::FLAG_NONE };
>> +      LookupFHResult fhr { nullptr, 0 };
> 
>     But what I do not understand is that there are more references to
>     RGWFileHandle::FLAG_NONE ( for instance a few lines earlier ),
> 
> yeah because |constexp| /and/ passing by rvalue.
> 
>     And per your link-text, that static declaration would require an
>     explicit definition.
>     But that sort of surpasses the constexpr declaration.
> 
> please note the difference of ODR-use and non ODR-use.
> 
>     Interested in what your suggestion for a fix is.
> 
> - LookupFHResult fhr { nullptr, RGWFileHandle::FLAG_NONE };
> + LookupFHResult fhr { nullptr, uint32_t(RGWFileHandle::FLAG_NONE) };

Cool, I'll give it a try.

--WjW

wjwithagen: @tchaikov 
Works like a charm. No more complaints in the linker.
I will change and rebase this commit.

src/include/compat.h
tchaikov: > I think you are refering to the text in the discussion, not the actual commit?

yes. thanks for the explanation. now i have a vague understanding now =) what is the "a compile time check" exactly? could you document this as a comment here?

wjwithagen: @tchaikov 
Oke, I'll see if I can write a concise wording of what I mean in the compat.h file.

wjwithagen:
@tchaikov 
I hope I addresses you issues to your satisfaction.
changed compat.h and rgw_file.h
Also added some more pthread includes and excludes in HeartbeatMap.cc WorkQueue.cc

wjwithagen:
@tchaikov 
Wow, I though i sort of understood what was going on, but this even shows I did not full get all of the details. I'll put this in the commit. 

The cynic in me wonders how many of the other developers will find this, once they run into a similar issue. But is a small gem.

tchaikov:
> The cynic in me wonders how many of the other developers will find this, once they run into a similar issue. 

good point. maybe we can quote the error message from clang in the commit message also, so it's more full-text search friendly?

BTW, i forgot to put the footmark in my reply "[1] http://en.cppreference.com/w/cpp/language/definition".

tchaikov:
another reason reason that we need to have a decent commit message is that the cast of `uint32_t(RGWFileHandle::FLAG_NONE)` is unnecessary at the first glance. and people with curiosity might want `git blame` this file and check why it is put this way.

wjwithagen:
@tchaikov 
Valid point, but that would warrant a comment actually in the file at the change that 
 - tells people no to "just" change this.
 - directs people to the commit, if the want to know why.

Oke?

wjwithagen:
@liewegas @tchaikov 
Ping?
Any work to be done before it goes into teuthology?

tchaikov:
@wjwithagen 

could you address the comment in https://github.com/ceph/ceph/pull/9178/commits/81b0dbaabb29f9de502a18898b5781dd39faa785 ?

wjwithagen:
@tchaikov 
I think I addresses that by including the compat.h at the top of the file.
And totally removing the pthread_np.h stuff

So the commit you are refering to, is an old one?

tchaikov:
@wjwithagen oh, i see. you added it in 81b0dba and removed it in 70d3b34. so would you please squash  81b0dba into 70d3b34?

wjwithagen:
@tchaikov 
Yup, I saw that, just rebased

tchaikov:
lgtm

wjwithagen:
@tchaikov 
Are we still waiting for anything to happen before commits?

tchaikov:
@wjwithagen see https://github.com/ceph/ceph/pull/8298#issuecomment-216877063

wjwithagen:
@tchaikov 
Oke, guess I'll have to sit out my impatience. 
How often are things thrown into teuthology?
Or does that not have any forcastable interval?

tchaikov:
@wjwithagen i dunno. i will try to test your changeset along with some of other needs-qa PRs this week-end.

wjwithagen:
@tchaikov 
Oke, thanx, enjoy the weekend then. 8)

tchaikov:
tested at 

- http://pulpito.ceph.com/kchai-2016-06-04_19:52:10-rados-wip-kefu-testing---basic-mira/
- http://pulpito.ceph.com/kchai-2016-06-04_09:44:21-rados-wip-kefu-testing---basic-mira/
- http://pulpito.ceph.com/kchai-2016-06-04_05:52:02-rados-wip-kefu-testing---basic-mira/

the failures are caued by environmental problem, test suite issues [1], and teuthology issue[2]

--
[1] https://github.com/ceph/ceph-qa-suite/pull/1031 
[2] http://tracker.ceph.com/issues/16142

wjwithagen:
On 5-6-2016 08:01, Kefu Chai wrote:
> Merged #9178 <https://github.com/ceph/ceph/pull/9178>.


Yappa, Youze, ...
Thanx Kefu

Off to preparing the next batch of commits

tchaikov:
yw =)

wjwithagen:
On 5-6-2016 14:46, Kefu Chai wrote:
> yw =)

Thanx,

But in the meantime I actually tried building from HEAD and see that
RocksBD is being upgraded to a version that does not really like
FreeBSD. Which I don't get because 4.5.1 is on FreeBSD and that has one
small patch.

I know there has been discussion on this on the developers list.
Ceph seems to be using 4.8.0, and the things that were in 4.5.1 fixing
stuff for some platfoms, that FreeBSD needs: Like PrintStackTrace.
In 4.5.1 it is just and empty wrapper for all OSes that are not
explicitly filled in.

So now I have to figure out why this only bit me now, and not earlier in
my WIP tree. And work out a way to start fixing this.
But since RocksDB is external that will not be easy.

wjwithagen:
@liewegas @tchaikov 
Starting to build part 3: #9504, Getting the Unittests to build

PR #9179, opened 2016-05-18
cls_journal: initialize empty commit position upon client register

vshankar:


vshankar:
@dillaman @rjfd  I've put this up early for early reviews.

Meanwhile, I'll work on writing unit tests and clearing up TODOs.

src/cls/journal/cls_journal.cc
```diff
@@ -540,10 +540,12 @@ int journal_client_register(cls_method_context_t hctx, bufferlist *in,
                             bufferlist *out) {
   std::string id;
   bufferlist data;
+  std::set<cls::journal::Client> clients;
   try {
     bufferlist::iterator iter = in->begin();
     ::decode(id, iter);
     ::decode(data, iter);
+    ::decode(clients, iter);
```
dillaman: No need to pass in the clients (might be stale and it breaks compatibility with older clients).  This class method should just read the current client list keys and perform the logic.

vshankar:
I'll send an updated patch for failures in "test/cls_journal/test_cls_journal.cc"

src/cls/journal/cls_journal.cc
```diff
@@ -567,7 +567,25 @@ int journal_client_register(cls_method_context_t hctx, bufferlist *in,
...
+  cls::journal::ObjectPosition first;
+  cls::journal::ObjectSetPosition object_set_position, least;
+
+  least = cls::journal::ObjectSetPosition();
+  for (std::set<cls::journal::Client>::iterator it = clients.begin();
+	 it != clients.end(); ++it) {
+    object_set_position = (*it).commit_position;
+    if (!object_set_position.object_positions.empty()) {
+      first = object_set_position.object_positions.front();
+      if (!tag_tid || (tag_tid >= first.tag_tid) || (entry_tid >= first.entry_tid)) {
```
dillaman: I believe tag_tid can be zero -- so use an optional or a bool to know if you have a valid entry.

src/cls/journal/cls_journal.cc
vshankar: Calling journal_client_list() was my initial approach, but that meant having similar aggregation logic as in C_ClientList::complete(). Maybe, treat `::decode(clients, iter)` specially in another `try..catch` block and handle failures (falling back to older behaviour)? However, that skips initializing commit positions.

dillaman: It's still potentially stale data -- you can use the current cls_journal::journal_client_list method as a guide for how to do it within the class object (it will be simpler since you don't have to worry about start after/max return).

dillaman: ... well, technically I guess you will have to handle batch reads using an internal "start after"

vshankar: Ah! I see, `::decode(clients, iter)` could still return garbage. I'll make the required changes and resubmit (I had undone my earlier implementation using `journal_client_list`).

src/cls/journal/cls_journal.cc
rjfd: I'm not sure if I'm right about this, but when comparing commit positions there is a precedence of `tag_tid` over `entry_tid`, and therefore you should only compare the `entry_tid` if the `tag_tid`s are equal.
As it is you might register a client with a `tag_tid` that is not the minimum.

vshankar: @rjfd yeh, that logic needs correction.

src/cls/journal/cls_journal.cc
vshankar: @dillaman With the updated patch, I still call `journal_client_list()` from `journal_client_register()` and have helper methods to decode and aggregate clients (with batching).

Also, these helpers reside in cls_journal_client.cc but needs to be called from cls_journal.cc vai externs.

dillaman: I'd rather see both use a helper method to do the decoding in X record batches.  Not that we should have thousands of clients, but it's a good style to use when working in OSD methods to chunk the work.

vshankar: Sure, that way less processing in done on the client side (esp. for `client_list`). What I can do is send my current working patch for review and work on implementing the suggestion above.

src/cls/journal/cls_journal_types.h
```diff
@@ -12,6 +12,10 @@
 #include <set>
 #include <string>
 
+namespace {
+static const uint64_t JOURNAL_MAX_RETURN = 256;
```
dillaman: Minor: Shouldn't have anonymous namespaces in headers. You can move this to the cls::journal namespace.

src/cls/journal/cls_journal.cc
```diff
@@ -224,6 +224,86 @@ int expire_tags(cls_method_context_t hctx, const std::string *skip_client_id) {
...
+              err.what());
+      return -EIO;
+    }
+  }
+
+  return 0;
+}
+
+int find_min_commit_position(cls_method_context_t hctx,
+			     cls::journal::ObjectSetPosition &minset) {
```
dillaman: Minor: use pointers for out values

src/cls/journal/cls_journal.cc
```diff
@@ -224,6 +224,86 @@ int expire_tags(cls_method_context_t hctx, const std::string *skip_client_id) {
   return 0;
 }
 
+int get_client_list_range(cls_method_context_t hctx,
+			  std::set<cls::journal::Client> &clients,
```
dillaman: Minor: use pointers for out values

src/cls/journal/cls_journal.cc
```diff
@@ -224,6 +224,86 @@ int expire_tags(cls_method_context_t hctx, const std::string *skip_client_id) {
...
+  if (r < 0) {
+    return r;
+  }
+
+  bool valid = false;
+  uint64_t tag_tid = 0, entry_tid = 0;
+  cls::journal::ObjectPosition first;
+  cls::journal::ObjectSetPosition object_set_position;
+
+  for (std::set<cls::journal::Client>::iterator it = clients.begin();
```
dillaman: Move this section up into the ```do { } while``` batch loop so that you have a restricted working set.

src/cls/journal/cls_journal.cc
```diff
@@ -224,6 +224,86 @@ int expire_tags(cls_method_context_t hctx, const std::string *skip_client_id) {
...
+       it != clients.end(); ++it) {
+    object_set_position = (*it).commit_position;
+    if (!object_set_position.object_positions.empty()) {
+      first = object_set_position.object_positions.front();
+      if (!valid || (tag_tid > first.tag_tid)
```
dillaman: ```ObjectPosition``` has a ```operator<``` method defined, so actually you could get rid of ```valid```, ```tag_tid```, and ```entry_tid``` and instead say ```if (min_set.empty() || min_set.front() < first)```

src/cls/journal/cls_journal.cc
```diff
@@ -224,6 +224,86 @@ int expire_tags(cls_method_context_t hctx, const std::string *skip_client_id) {
...
+    }
+  } while((r == 0) && !(batch.size() < JOURNAL_MAX_RETURN));
+
+  if (r < 0) {
+    return r;
+  }
+
+  bool valid = false;
+  uint64_t tag_tid = 0, entry_tid = 0;
+  cls::journal::ObjectPosition first;
```
dillaman: Minor: define variables at first use where possible

src/cls/journal/cls_journal.cc
dillaman: Never mind -- it compares ```object_number``` first which isn't wanted in this case

src/cls/journal/cls_journal.cc
```diff
@@ -224,6 +224,86 @@ int expire_tags(cls_method_context_t hctx, const std::string *skip_client_id) {
...
+      return -EIO;
+    }
+  }
+
+  return 0;
+}
+
+int find_min_commit_position(cls_method_context_t hctx,
+                             cls::journal::ObjectSetPosition *minset) {
+  int r;
```
dillaman: Declare all your variables at first use within the tightest possible scope.

src/cls/journal/cls_journal.cc
```diff
@@ -224,6 +224,86 @@ int expire_tags(cls_method_context_t hctx, const std::string *skip_client_id) {
...
+              || ((tag_tid == first.tag_tid) && (entry_tid > first.entry_tid))) {
+            tag_tid = first.tag_tid;
+            entry_tid = first.entry_tid;
+            *minset = cls::journal::ObjectSetPosition(object_set_position);
+            valid = true;
+          }
+        }
+      }
+    }
+  } while((r == 0) && !(batch.size() < max_return));
```
dillaman: ```} while (r == 0 && batch.size() == max_return)```?

src/cls/journal/cls_journal.cc
```diff
@@ -224,6 +224,86 @@ int expire_tags(cls_method_context_t hctx, const std::string *skip_client_id) {
...
+      // update the (minimum) commit position from this batch of clients
+      for(std::set<cls::journal::Client>::iterator it = batch.begin();
+          it != batch.end(); ++it) {
+        object_set_position = (*it).commit_position;
+        if (!object_set_position.object_positions.empty()) {
+          first = object_set_position.object_positions.front();
+
+          // least tag_tid (or least entry_tid for matching tag_tid)
+          if (!valid || (tag_tid > first.tag_tid)
+              || ((tag_tid == first.tag_tid) && (entry_tid > first.entry_tid))) {
```
dillaman: Minor: can you move the logical or to the end of the previous line

src/cls/journal/cls_journal.cc
```diff
@@ -224,6 +224,86 @@ int expire_tags(cls_method_context_t hctx, const std::string *skip_client_id) {
...
+  valid = false;
+  tag_tid = entry_tid = 0;
+
+  start_after = "";
+  max_return = cls::journal::JOURNAL_MAX_RETURN;
+
+  do {
+    batch.clear();
+    r = get_client_list_range(hctx, &batch, start_after, max_return);
+    if ((r == 0) && !batch.empty()) {
```
dillaman: You could just do:

 ```
if (r < 0 || batch.empty()) {
  break;
}
```

It avoids the need to indent the whole section

File: src/cls/journal/cls_journal.cc (hunk 10 above)
Comment (vshankar) [inline src/cls/journal/cls_journal.cc]: This is what I was a bit unsure of, but `C_ClientList::complete()` does something similar (calls `finish(0)`).

File: src/cls/journal/cls_journal.cc (hunk 12 above)
Comment (vshankar) [inline src/cls/journal/cls_journal.cc]: Yeh, but with indent offset of two, that does not make it look worse ;)

So, I didn't care much about that..

File: src/cls/journal/cls_journal.cc (hunk 10 above)
Comment (dillaman) [inline src/cls/journal/cls_journal.cc]: I was just commenting on the negation of a less-than operator -- it looks a bit confusing.  We should know we can break out of the loop if we had an error (won't actually be necessary if we break out above) or if we retrieved less than the requested number of clients.

Comment (vshankar) [inline src/cls/journal/cls_journal.cc]: OK. I was a bit confused about what `cls_cxx_map_get_vals()` returns.

I'll change this one (but leave the other one as it is).

File: src/cls/journal/cls_journal.cc [hunk 13]
```diff
@@ -224,6 +224,87 @@ int expire_tags(cls_method_context_t hctx, const std::string *skip_client_id) {
...
+  int r;
+  bool valid = false;
+  std::string start_after = "";
+  uint64_t tag_tid = 0, entry_tid = 0;
+
+  while (true) {
+    std::set<cls::journal::Client> batch;
+    cls::journal::ObjectPosition first;
+    cls::journal::ObjectSetPosition object_set_position;
+    uint64_t max_return = cls::journal::JOURNAL_MAX_RETURN;
```
vshankar: @dillaman I moved this to a while loop instead so that `batch` can too be declared within the scope (variables which needed default initialization and change state in the loop are kept outside the scope) and hopefully makes it more readable than `do { }while`.

src/cls/journal/cls_journal.cc
```diff
@@ -224,6 +224,87 @@ int expire_tags(cls_method_context_t hctx, const std::string *skip_client_id) {
...
+
+    batch.clear();
```
rjfd: If `batch` is declared within the scope of the while loop, you don't need to clear its contents.

src/cls/journal/cls_journal.cc
```diff
@@ -224,6 +224,87 @@ int expire_tags(cls_method_context_t hctx, const std::string *skip_client_id) {
...
+    }
+
+    start_after = batch.rbegin()->id;
+
+    // update the (minimum) commit position from this batch of clients
+    for(std::set<cls::journal::Client>::iterator it = batch.begin();
+        it != batch.end(); ++it) {
+      object_set_position = (*it).commit_position;
+      if (!object_set_position.object_positions.empty()) {
+        first = object_set_position.object_positions.front();
```
rjfd: I would move the declaration of `first`to here. e.g. `cls::journal::ObjectPosition& first = object_set_position.object_positions.front();`

src/cls/journal/cls_journal.cc
rjfd: Why do you need the `max_return` variable, it will always have the same value. Why not use the constant `cls::journal::JOURNAL_MAX_RETURN` directly?

src/cls/journal/cls_journal.cc
```diff
@@ -224,6 +224,87 @@ int expire_tags(cls_method_context_t hctx, const std::string *skip_client_id) {
...
+    if ((r < 0) || batch.empty()) {
+      break;
+    }
+
+    start_after = batch.rbegin()->id;
+
+    // update the (minimum) commit position from this batch of clients
+    for(std::set<cls::journal::Client>::iterator it = batch.begin();
+        it != batch.end(); ++it) {
+      object_set_position = (*it).commit_position;
```
rjfd: I would move the declaration of `object_set_position` to here. e.g. cls::journal::ObjectSetPosition& object_set_position = (*it).commit_position;`

src/cls/journal/cls_journal.cc
vshankar: OK. Its a matter of preference and possibly carried over by working on past projects (and some of them really had strict rules to be followed). Normally, I tend to assign variables in other namespaces locally to use later.

dillaman: That's a practice left over from very old versions of C where you had to declare everything that would go on the stack at the same time.  That was fixed in C99 since it led to the introduction of way too many bugs (that and not assigning the initial value at declaration).

src/cls/journal/cls_journal.cc
vshankar: Hmm, right. Looks like the the declaring a variable at the appropriate scope is very strictly followed. Other parts of the source does not seem to strictly follow this (yeh, the codebase is too large to keep up with such strictness always).

Thanks! I'll keep this in mind in the future.

src/cls/journal/cls_journal.cc
dillaman: The Ceph project generally follows the Google C++ style guide: https://google.github.io/styleguide/cppguide.html#Local_Variables

vshankar: As an example -- using a value from a long nested namespace with strict limits on the number of chars per line. Now you should be able to relate this with parts of the patch..

dillaman:
lgtm

rjfd:
@dillaman @vshankar Just tested my PR #8473 with the changes of this PR, and all tests from qa/workunits/rbd/rbd_mirror.sh are now passing successfully.
LGTM

PR #9183, opened 2016-05-18
rgw: sync retry

yehudasa:
keep failed data sync buckets in a separate log, go over that log as part of the incremental sync.

### DIFF
--- a/src/rgw/rgw_cr_rados.cc
+++ b/src/rgw/rgw_cr_rados.cc
@@ -289,6 +289,37 @@ int RGWRadosGetOmapKeysCR::send_request() {
   return ioctx.aio_operate(oid, cn->completion(), &op, NULL);
 }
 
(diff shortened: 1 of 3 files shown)


cbodley:
this branch went through [teuthology](http://pulpito.ceph.com/cbodley-2016-05-28_11:42:37-rgw-wip-rgw-testing-cbodley---basic-smithi/) with two valgrind issues:
* `valgrind issue   <kind>Leak_DefinitelyLost</kind> in /var/log/ceph/valgrind/mon.b.log`
* `valgrind issue   <kind>InvalidRead</kind> in /var/log/ceph/valgrind/client.0.log`

the invalid read was in `RGWRados::get_obj_iterate_cb()`, which is unrelated to this pr. i rescheduled that one test and it went through [clean](http://pulpito.ceph.com/cbodley-2016-05-31_13:33:10-rgw-wip-rgw-testing-cbodley---basic-smithi/)

PR #9188, opened 2016-05-18
rgw: improve support for Swift's object versioning.

rzarzynski:
This patch allows RadosGW to pass the RefStack with an accuracy to the [RFC7230 violation issue](https://bugs.launchpad.net/tempest/+bug/1536251)  which is clearly a Tempest bug.

Following areas have been improved:
* reverting previous versions during `DELETE` on an object.
* preserving the `X-Versions-Location` between multiple `PUT` and `POST` requests,
* disabling the versioning through the `X-Remove-Versions-Location` header.

I'm providing below the results of a Tempest run against RefStack's tests. The version of Tempest is `aff9cc072bbbb222b09a3411b203c180b493eae8` (Tue May 10 15:44:57 2016 +0000) with [the fix](https://review.openstack.org/272062) for the RFC7230 issue incorporated in.

CC: @yehudasa.

```
$ ./run_tempest.sh -V "`cat ./refstack.txt`"
...
{2} tempest.api.object_storage.test_object_version.ContainerTest.test_versioned_container [0.918812s] ... ok
{1} tempest.api.object_storage.test_object_services.ObjectTest.test_copy_object_2d_way [0.238136s] ... ok
{3} tempest.api.object_storage.test_object_temp_url.ObjectTempUrlTest.test_get_object_using_temp_url [0.062831s] ... ok
{3} tempest.api.object_storage.test_object_temp_url.ObjectTempUrlTest.test_put_object_using_temp_url [0.145417s] ... ok
{1} tempest.api.object_storage.test_object_services.ObjectTest.test_copy_object_across_containers [0.729780s] ... ok
{1} tempest.api.object_storage.test_object_services.ObjectTest.test_copy_object_in_same_container [0.235504s] ... ok
{1} tempest.api.object_storage.test_object_services.ObjectTest.test_copy_object_to_itself [0.118214s] ... ok
{1} tempest.api.object_storage.test_object_services.ObjectTest.test_create_object [0.103907s] ... ok
{1} tempest.api.object_storage.test_object_services.ObjectTest.test_delete_object [0.065198s] ... ok
{1} tempest.api.object_storage.test_object_services.ObjectTest.test_get_object [0.067509s] ... ok
{1} tempest.api.object_storage.test_object_services.ObjectTest.test_get_object_if_different [0.068058s] ... ok
{1} tempest.api.object_storage.test_object_services.ObjectTest.test_object_upload_in_segments [0.387238s] ... ok
{0} tempest.api.object_storage.test_object_expiry.ObjectExpiryTest.test_get_object_after_expiry_time [12.257810s] ... ok

======
Totals
======
Ran: 13 tests in 19.0000 sec.
 - Passed: 13
 - Skipped: 0
 - Expected Fail: 0
 - Unexpected Success: 0
 - Failed: 0
Sum of execute time for each test: 15.3984 sec.

==============
Worker Balance
==============
 - Worker 0 (1 tests) => 0:00:12.257810
 - Worker 1 (9 tests) => 0:00:02.076278
 - Worker 2 (1 tests) => 0:00:00.918812
 - Worker 3 (2 tests) => 0:00:00.213534
```


rzarzynski:
@yehudasa: full object storage campaign of Tempest as well as s3-tests haven't found any regression here. IMO in the patch is ready for review.

liewegas:
:beers:

src/rgw/rgw_common.h
```diff
@@ -992,11 +992,14 @@ struct RGWBucketInfo
...
   void decode_json(JSONObj *obj);
 
-  bool versioned() { return (flags & BUCKET_VERSIONED) != 0; }
+  bool versioned() const { return (flags & BUCKET_VERSIONED) != 0; }
   int versioning_status() { return flags & (BUCKET_VERSIONED | BUCKET_VERSIONS_SUSPENDED); }
   bool versioning_enabled() { return versioning_status() == BUCKET_VERSIONED; }
 
-  bool has_swift_versioning() { return swift_versioning; }
+  bool has_swift_versioning() const {
+    /* A bucket may be versioned through one mechanism only. */
```
yehudasa: @rzarzynski this makes sense, but maybe we can think about whether there's some restricted way in which both systems can coexist.

src/rgw/rgw_rados.cc
```diff
@@ -5656,30 +5659,83 @@ int RGWRados::BucketShard::init(rgw_bucket& _bucket, rgw_obj& obj)
...
+}
+
+
+int RGWRados::swift_versioning_copy(RGWObjectCtx& obj_ctx,
+                                    const rgw_user& user,
+                                    RGWBucketInfo& bucket_info,
+                                    rgw_obj& obj)
 {
-  if (!bucket_info.has_swift_versioning() || bucket_info.swift_ver_location.empty()) {
+  if (swift_versioning_enabled(bucket_info) == false) {
```
yehudasa: what's wrong with `if (!swift_versioning_enabled())`?

src/rgw/rgw_rados.cc
```diff
@@ -5656,30 +5659,83 @@ int RGWRados::BucketShard::init(rgw_bucket& _bucket, rgw_obj& obj)
...
     return 0;
   }
 
+  RGWObjState * state = nullptr;
+  int r = get_obj_state(&obj_ctx, obj, &state, false);
```
yehudasa: can we avoid this call? it'd be better having this passed from the caller

yehudasa: @rzarzynski specifically we need to think about operation atomicity

src/rgw/rgw_rados.cc
```diff
@@ -5723,12 +5779,107 @@ int RGWRados::swift_versioning_copy(RGWBucketInfo& bucket_info, RGWRados::Object
                NULL, /* void (*progress_cb)(off_t, void *) */
                NULL); /* void *progress_data */
   if (r == -ECANCELED || r == -ENOENT) { /* has already been overwritten, meaning another rgw process already copied it out */
+  ldout(cct, 20) << "++++ ECANC   --- SWIFT VERSION COPY BEGIN ++++ LINE: " << __LINE__ << dendl;
```
yehudasa: @rzarzynski maybe rework this log message

src/rgw/rgw_rados.cc
```diff
@@ -5723,12 +5779,107 @@ int RGWRados::swift_versioning_copy(RGWBucketInfo& bucket_info, RGWRados::Object
...
   return r;
 }
 
+int RGWRados::swift_versioning_restore(RGWObjectCtx& obj_ctx,
+                                       const rgw_user& user,
+                                       RGWBucketInfo& bucket_info,
+                                       rgw_obj& obj,
+                                       bool& restored)             /* out */
+{
+  if (swift_versioning_enabled(bucket_info) == false) {
```
yehudasa: shouldn't this be called after get_bucket_info()

yehudasa: ah, nevermind

src/rgw/rgw_rados.cc
```diff
@@ -5723,12 +5779,107 @@ int RGWRados::swift_versioning_copy(RGWBucketInfo& bucket_info, RGWRados::Object
...
+    return 0;
+  }
+
+  /* Bucket info of the bucket that stores previous versions of our object. */
+  RGWBucketInfo archive_binfo;
+
+  int ret = get_bucket_info(obj_ctx, bucket_info.bucket.tenant,
```
yehudasa: @rzarzynski I wonder if we need to keep also the target bucket id, and do a get_bucket_instance_info() here.. probably not though

src/rgw/rgw_rados.cc
```diff
@@ -5723,12 +5779,107 @@ int RGWRados::swift_versioning_copy(RGWBucketInfo& bucket_info, RGWRados::Object
...
+  if (ret < 0) {
+    return ret;
+  }
+
+  /* Abort the operation if the bucket storing our archive belongs to someone
+   * else. This is a limitation in comparison to Swift as we aren't taking ACLs
+   * into consideration. For we can live with that.
+   *
+   * TODO: delegate this check to un upper layer and compare with ACLs. */
+  if (bucket_info.owner != archive_binfo.owner) {
```
yehudasa: we probably need to document this somewhere

src/rgw/rgw_rados.cc
```diff
@@ -5723,12 +5779,107 @@ int RGWRados::swift_versioning_copy(RGWBucketInfo& bucket_info, RGWRados::Object
...
+      return 0;
+    } else if (ret < 0) {
+      return ret;
+    } else {
+      restored = true;
+    }
+
+    /* Need to remove the archived copy. */
+    ret = delete_obj(obj_ctx, archive_binfo, archive_obj,
+          archive_binfo.versioning_status());
```
yehudasa: @rzarzynski can we allow the archive bucket to have S3 versioning? this can have weird side effects.

yehudasa: @rzarzynski specifically we need to make sure the target object is actually removed, and we don't end up with a delete marker created

yehudasa: for this to happen the archive_obj will need to include a valid object instance

src/rgw/rgw_rados.cc
```diff
@@ -5656,30 +5659,83 @@ int RGWRados::BucketShard::init(rgw_bucket& _bucket, rgw_obj& obj)
 }
 
 
-int RGWRados::swift_versioning_copy(RGWBucketInfo& bucket_info, RGWRados::Object *source, RGWObjState *state,
-                                    rgw_user& user)
+/* Execute @handler on last item in bucket listing for bucket specified
+ * in @bucket_info. @obj_prefix and @obj_delim narrow down the listing
+ * to objects matching these criterias. */
+int RGWRados::on_last_in_bucket(RGWBucketInfo& bucket_info,
```
yehudasa: maybe rename to `on_last_entry_in_listing()`?

src/rgw/rgw_rest_swift.cc
```diff
@@ -503,6 +505,33 @@ static void get_rmattrs_from_headers(const req_state * const s,
...
+static int get_swift_versioning_settings(
+  req_state * const s,
+  boost::optional<std::string>& swift_ver_location)
+{
+  /* Removing the Swift's versions location has lower priority than setting
+   * a new one. That's the reason why we're handling it first. */
+  const std::string vlocdel =
+    s->info.env->get("HTTP_X_REMOVE_VERSIONS_LOCATION", "");
+  if (vlocdel.size()) {
+    swift_ver_location = boost::in_place(std::string());
```
yehudasa: what's wrong with `swift_ver_location.clear()`?

src/rgw/rgw_rest_swift.cc
```diff
@@ -503,6 +505,33 @@ static void get_rmattrs_from_headers(const req_state * const s,
...
+  }
+
+  std::string vloc = s->info.env->get("HTTP_X_VERSIONS_LOCATION", "");
+  if (vloc.size()) {
+    /* If the Swift's versioning is globally disabled but someone wants to
+     * enable it for a given container, new version of Swift will generate
+     * the precondition failed error. */
+    if (! s->cct->_conf->rgw_swift_versioning_enabled) {
+      return -ERR_PRECONDITION_FAILED;
```
yehudasa: maybe -ENOTSUPP instead?

rzarzynski: We need to end with the HTTP's `412 Precondition Failed` error here to be compatible with Swift. It seems that `ENOTSUPP` would be translated into `500 Internal Server Error`.

src/rgw/rgw_rest_swift.cc
rzarzynski: `boost::optional` doesn't have such method. Moreover, at the moment it's empty and we need to construct `std::string` to fill it.

src/rgw/rgw_rados.cc
rzarzynski: I personally think that `== false` is harder to overlook while reading the code. Anyway, I can rework that if you prefer the second option.

src/rgw/rgw_rados.cc
rzarzynski: Oops, accidentally incorporated this line during `git add -p`. I will rework. Thanks for pointing this out.

src/rgw/rgw_rest_swift.cc
yehudasa: this seems like the wrong return code. Would rather return 405 (through ERR_METHOD_NOT_ALLOWED, not ENOTSUPP).

src/rgw/rgw_rados.cc
rzarzynski: This name sounds much better. Thanks!

src/rgw/rgw_rados.cc
rzarzynski: What do you think about putting that as a note in the description of `rgw_swift_versioning_enabled`?

src/rgw/rgw_rest_swift.cc
rzarzynski: In previous version of the Swift's implementation there was no error at all. Take a look [the comment](https://github.com/openstack/swift/blob/master/swift/common/middleware/versioned_writes.py#L454).

src/rgw/rgw_common.h
rzarzynski: I'm not sure. Although `PUT` and `COPY` doesn't seem to be a problem for the coexistence (create an archived copy first, then go to the S3 versioning), `DELETE` might be a headache. The behavior is completely different: restoring a previous version vs putting a deletion marker. This would require some arbitrage, maybe through an API extension allowing clients to pick-up a desired policy.
I'm also afraid we don't have tools for testing this increased complexity. Tempest most likely won't accept such tests. Testing some parts of Swift API in s3-tests doesn't seem to be a good option either.

src/rgw/rgw_rados.cc
rzarzynski: I'm afraid we can't avoid `get_obj_state` here. On the level of `RGWCopyObj` and `RGWPutObj` there is no `RGWObjState` yet.

src/rgw/rgw_rados.cc
rzarzynski: I don't think so. Generally speaking, Swift's idea for restoring is to get listing of a given container and pick-up a best matching object if possible. The list is constructed dynamically. If someone mess with archived copies or even with the whole container, potential items will be simply skipped leading to an empty list in the corner case.

src/rgw/rgw_rados.cc
rzarzynski: DONE.

src/rgw/rgw_rados.cc
rzarzynski: DONE.

src/rgw/rgw_rados.cc
rzarzynski: DONE.

src/rgw/rgw_rados.cc
rzarzynski: DONE. The bucket storing copies cannot be versioned by the S3 mechanism. Otherwise, we'll return `-ERR_PRECONDITION_FAILED`.

rzarzynski:
@yehudasa: rebased and addressed your comments.

src/rgw/rgw_rados.cc
mattbenjamin: I prefer if (!swift_versioning_enabled()) as well

mattbenjamin:
@yehudasa @oritwas are we planning to merge this?

yehudasa:
@mattbenjamin yes. Moreover, we plan to backport it to jewel. @rzarzynski this one needs rebasing.

rzarzynski:
@yehudasa I'm rebasing this right now. Will push updates branch soon.

src/rgw/rgw_rados.cc
rzarzynski: DONE.

rzarzynski:
@oritwas, @yehudasa, @mattbenjamin: rebased.

PR #9191, opened 2016-05-19
rbd-mirror: disable librbd caching for replicated images

dillaman:
Each image has its own cache and each cache uses its own thread. With
a large replicated cluster, this could result in thousands of extra
threads and gigabytes of extra memory.

Fixes: http://tracker.ceph.com/issues/15930


trociny:
lgtm, just a thought though. It will be a little unfortunate that we are running rbd_mirror tests with cache enabled by default, while rbd-mirror will be run with cache disabled.

May be provide and use a way to explicitly tell on image context open that cache should be disabled? Or call shut_down_cache() after open?

dillaman:
@trociny Good point -- can do the same override in the test case

dillaman:
updated

PR #9192, opened 2016-05-19
cmake: use object library instead recompiling popular .cc files

tchaikov:
* common/TextTable.cc
* common/secret.c


cbodley:
works for me :+1: thanks @tchaikov

PR #9194, opened 2016-05-19
Do not link lttng into libglobal

branto1:
Without this patch, radosgw binary is linked against lttng. This causes SELinux and probably even AppArmor denials, we should backport it to jewel.


branto1:
I'll merge this if it builds cleanly in gitbuilders.

kmroz:
@branto1 - Great, thank you.

branto1:
The gitbuilders are quite happy with the change, the jenkins failure has nothing to do with this patch -> merging.

PR #9195, opened 2016-05-19
ceph-disk: parted < 3.2.16 misbehaves with multiple disks

dachary:
When running multiple ceph-disk prepare /dev/sd? /dev/sde at the same
time, partprobe sometime fails with

  ceph-disk: Error: partprobe /dev/vd? failed :
    Error informing the kernel about modifications to partition /dev/vd?1

This misbehavior has been verified with parted 3.1.23. The exact cause
is unclear and the tracker issue should be updated with findings.

Fixes http://tracker.ceph.com/issues/15918

### DIFF
--- a/ceph.spec.in
+++ b/ceph.spec.in
@@ -205,7 +205,7 @@ Requires:      python-setuptools
 Requires:      grep
 Requires:      xfsprogs
 Requires:      logrotate
-Requires:      parted
+Requires:      parted >= 3.2.16
 Requires:      util-linux
 Requires:      hdparm
 Requires:      cryptsetup


smithfarm:
@dachary No deb?

dachary:
@smithfarm I don't think the problem shows on ubuntu/debian. I'm not very confortable enforcing version constraints when I don't really know why they are necessary. I'll dig more :-)

smithfarm:
In that case, `Reviewed-by: Nathan Cutler <ncutler@suse.com>`

ktdreyer:
RHEL 7 includes parted-3.1-23.el7, so this will break RHEL 7 installs

dachary:
@ktdreyer shoot :-) So the only way is to figure out how to workaround the problem with this old (between 2012 and 2014) version of parted ?

ktdreyer:
That sounds right

alfredodeza:
Most of the problems we've seen in ceph-deploy issues regarding ceph-disk calls have been caused by the async nature of the udev rules. 

Have we ensured that this odd behavior in partprobe is not being caused by racing udev rules? What happens when the commands ceph-disk is firing are done in a system that doesn't have Ceph installed (or has it but without any dev rules) ?

dachary:
@alfredodeza ceph-deploy is randomly calling partprobe and creates problems rather than solving any. In any case it's a different problem altogether (see http://tracker.ceph.com/issues/14099 for instance).

PR #9196, opened 2016-05-19
cmake: add cython_rados to vstart target

tchaikov:


tchaikov:
@jcsp could you take a look?

PR #9197, opened 2016-05-19
osd: select more suitable shard as auth shard

XinzeChi:
If there are A,B,C shards for some object. The oi.size == ondisk size
for A and B. The oi.size != ondisk_size for C. The origin algorithm
would select C as auth shard. The PR would select A.

@

### DIFF
--- a/src/osd/PGBackend.cc
+++ b/src/osd/PGBackend.cc
@@ -538,11 +538,6 @@ map<pg_shard_t, ScrubMap *>::const_iterator
       continue;
     }
 
-    // note candidate in case we can't find anything better, because
-    // something is better than nothing.  FIXME.
-    auth = j;
-    *auth_oi = oi;
-
     uint64_t correct_size = be_get_ondisk_size(oi.size);
     if (correct_size != i->second.size) {
       // invalid size, probably corrupt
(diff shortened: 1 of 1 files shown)


liewegas:
Can you decribe in the commit message what the change is and why it is better?

XinzeChi:
@liewegas , Do you think it make sense?

liewegas:
Ah, yeah, that makes sense to me!

@athanatos ?

tchaikov:
> If there are A,B,C shards for some object. The oi.size == ondisk size for A and B. The oi.size != ondisk_size for C. The origin algorithm would select C as auth shard. The PR would select A.

the reason why it selects C is that neither A or B is able to pass all the tests, and C does not have serious problem like read error or OI attr corruption. for example, A's data digest does not match, and B's omap digest fails to match. and we returns with C, simply because "something is better than nothing".

in your example, it selects A instead of C, not because A is better than C, just it has the preference for choosing the first shard (when the first candidate fails the test).

so i don't think this PR is superior than existing algorithm. but i agree, neither of them is perfect.

XinzeChi:
The PR would has the preference for the shard which has no read error, not missing oi attr, data digest match and omap digest match.

tchaikov:
if we has such a candidate, both algorithm return it.

liewegas:
I think @tchaikov  is right--I missed the break statement. We'll stop looking once we find a candidate that passes all the checks.

XinzeChi:
oh, I miss it too. close it.

PR #9198, opened 2016-05-19
os/filestore: add xattr option family to tracked keys

xiexingguo:
Also including some other cleanup changes.


xiexingguo:
@liewegas Minor fixes, simple and easy.

liewegas:
lgtm

PR #9200, opened 2016-05-19
rgw: we should probably return InvalidArgument when XML_parser fails of RGWPutACLs op

weiqiaomiao:
when input xml documention is invalid, bucket put acls fails, but RGW return -EACCESS, which appears as AceessDenied to the client, we should probably return _EINVAL(InvalidArgument) instead


theanalyst:
lgtm

PR #9202, opened 2016-05-19
mon: delete redundant code and add assertions to print

renhwztetecs:
delete the code that will not be executed before assert;
add error exception information in assert.

### DIFF
--- a/src/mon/AuthMonitor.cc
+++ b/src/mon/AuthMonitor.cc
@@ -298,7 +298,7 @@ bool AuthMonitor::preprocess_query(MonOpRequestRef op)
     return false;
 
   default:
-    assert(0);
+    assert(0 == "unknown type!");
     return true;
   }
 }
@@ -315,7 +315,7 @@ bool AuthMonitor::prepare_update(MonOpRequestRef op)
   case CEPH_MSG_AUTH:
     return prep_auth(op, true);
   default:
-    assert(0);
+    assert(0 == "unknown type!");
     return false;
   }
 }
--- a/src/mon/LogMonitor.cc
+++ b/src/mon/LogMonitor.cc
@@ -290,7 +290,7 @@ bool LogMonitor::preprocess_query(MonOpRequestRef op)
     return preprocess_log(op);
 
   default:
-    assert(0);
+    assert(0 == "unknown type!");
     return true;
   }
 }
@@ -306,7 +306,7 @@ bool LogMonitor::prepare_update(MonOpRequestRef op)
   case MSG_LOG:
     return prepare_log(op);
   default:
-    assert(0);
+    assert(0 == "unknown type!");
     return false;
   }
 }
--- a/src/mon/MonmapMonitor.cc
+++ b/src/mon/MonmapMonitor.cc
@@ -136,7 +136,7 @@ bool MonmapMonitor::preprocess_query(MonOpRequestRef op)
   case MSG_MON_JOIN:
     return preprocess_join(op);
   default:
-    assert(0);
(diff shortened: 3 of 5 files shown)


src/mon/PGMonitor.cc
```diff
@@ -2048,7 +2046,7 @@ static void note_stuck_detail(int what,
       whatname = "stale";
       break;
     default:
-      assert(0);
+/bin/bash: indent: command not found
```
xiexingguo: ? typo???

src/mon/OSDMonitor.cc
```diff
@@ -1367,8 +1367,7 @@ bool OSDMonitor::preprocess_query(MonOpRequestRef op)
     return preprocess_remove_snaps(op);
 
   default:
-    assert(0);
-    return true;
```
jecluis: I don't believe removing the return is safe or future proof. Granted, right now ceph is *always* compiled with assertions enabled, but that may not hold true in the future. This is not exactly redundant code - I take it as future proofing assertion handling, as at some point in time we may only compile assertions for debug/dev builds, and we must ensure the appropriate values are still returned.

src/mon/MonmapMonitor.cc
```diff
@@ -136,8 +136,7 @@ bool MonmapMonitor::preprocess_query(MonOpRequestRef op)
   case MSG_MON_JOIN:
     return preprocess_join(op);
   default:
-    assert(0);
-    return true;
```
jecluis: see my other comment @ https://github.com/ceph/ceph/pull/9202/commits/8a0d6f1a47b36cc5d2ff0818576ce687ecb0ab0e#r63887619

src/mon/LogMonitor.cc
```diff
@@ -306,8 +305,7 @@ bool LogMonitor::prepare_update(MonOpRequestRef op)
   case MSG_LOG:
     return prepare_log(op);
   default:
-    assert(0);
-    return false;
```
jecluis: see my other comment @ https://github.com/ceph/ceph/pull/9202/commits/8a0d6f1a47b36cc5d2ff0818576ce687ecb0ab0e#r63887619

src/mon/AuthMonitor.cc
```diff
@@ -315,8 +314,7 @@ bool AuthMonitor::prepare_update(MonOpRequestRef op)
   case CEPH_MSG_AUTH:
     return prep_auth(op, true);
   default:
-    assert(0);
-    return false;
```
jecluis: see my other comment @ https://github.com/ceph/ceph/pull/9202/commits/8a0d6f1a47b36cc5d2ff0818576ce687ecb0ab0e#r63887619

same applies to the chunk above.

renhwztetecs:
@jecluis 
thanks for your review
I agree with you, assert should only open when the debug/dev(whether there is a plan?).
but the current assert place lacks the necessary exceptions to judge and print, so it is not suitable for the application of the assert disable is not appropriate.so there is no modification of the case, I would like to remove the code will not be executed

athanatos:
Put the returns back

renhwztetecs:
@athanatos 
done 
thanks for you attention

renhwztetecs:
ping @athanatos 
:-)

athanatos:
This really doesn't seem worthwhile to me.  @tchaikov @jecluis What do you think?

tchaikov:
@renhwztetecs i am inclined to agree with @athanatos . would be nice if you can accompany this change with more something more significant.

PR #9203, opened 2016-05-19
osd,rados: add repair-copy op for scrub/repair

tchaikov:
this PR is based on #8931 by @dzafman

### DIFF
--- a/doc/rados/troubleshooting/troubleshooting-pg.rst
+++ b/doc/rados/troubleshooting/troubleshooting-pg.rst
@@ -441,8 +441,19 @@ In this case, we can learn from the output:
       ``0xffffffff``, which is calculated from the shard read from OSD.2
     * ``size_mismatch_oi``: the size stored in the object-info is different
       from the one read from OSD.2. The latter is 0.
+  * ``selected_object_info``: the information of the authoritiative copy(s)
+    choosen by OSD. It is the string representation of the object info of this
+    copy. In the example above:
 
-You can repair the inconsistent placement group by executing:: 
+    * ``s 968``: Size is ``968`` bytes
+    * ``dd e978e67f``: Data Digest is ``e978e67f``
+    * ``od ffffffff``: Omap Digest is ``ffffffff``
+
+If ``read_error`` is listed in the ``errors`` attribute of a shard, the
+inconsistency is likely due to disk errors. You might want to check your disk
+used by that OSD.
+
+You could repair the inconsistent placement group by executing::
 
 	ceph pg repair {placement-group-ID}
 
@@ -452,9 +463,28 @@ some predefined criteria. But this does not always work. For example, the stored
 data digest could be missing, and the calculated digest will be ignored when
 choosing the authoritative copies. So, please use the above command with caution.
 
-If ``read_error`` is listed in the ``errors`` attribute of a shard, the
-inconsistency is likely due to disk errors. You might want to check your disk
-used by that OSD.
+If you are able to identify the `good` copy by inspecting it, it is advisable to
+use the ``rados`` command or the RADOS APIs to fix the inconsistencies manually
+instead. In the above example, if the ``selected_object_info`` is hard to decipher
+or does not make sense at all, you will need to download the payload of the shards
+and inspect them manually::
+
+  rados -p rbd --force repair-get foo 0 14 foo.osd-0
+
+This command fetches the replica of ``foo`` to the local file ``foo.osd.0``
+from osd.0 at epoch 14. If you are happy with this replica after looking at
+it, you might want to overwrite bad replicas with good ones::
+
+  rados -p rbd repair-copy foo 2 14 1
+
+This command overwrites the replica on osd.2 with the good ones. Any replica not
+listed in the command line will be considered as good copy, and can be used to
+overwrite the corrupted ones. Please note, this command will refuse to rewrite
+the object if its version is not identical to the specified version, or the PG
+interval is expired. In this case, if ``foo`` object is updated after the scrub,
+and hence has greater version than "1", this command will fail. If the OSDs
+serving the PG holding this object peers after the deep scrub, this command
+will fail also. In that case, you need to re-schedule a scrub to this PG.
 
 If you receive ``active + clean + inconsistent`` states periodically due to 
 clock skew, you may consider configuring your `NTP`_ daemons on your 
--- a/src/include/rados/librados.hpp
+++ b/src/include/rados/librados.hpp
@@ -757,6 +757,7 @@ namespace librados
     int writesame(const std::string& oid, bufferlist& bl,
 		  size_t write_len, uint64_t off);
     int read(const std::string& oid, bufferlist& bl, size_t len, uint64_t off);
+    int repair_read(const std::string& oid, bufferlist& bl, size_t len, uint64_t off, int flags, int32_t osdid, uint32_t epoch);
     int remove(const std::string& oid);
     int remove(const std::string& oid, int flags);
     int trunc(const std::string& oid, uint64_t size);
@@ -967,6 +968,31 @@ namespace librados
 			std::map<uint64_t,uint64_t> *m, bufferlist *data_bl,
 			size_t len, uint64_t off);
     /**
+     * Asynchronously read a replica hosted by the specified osd
+     *
+     * @param oid the name of the object to read from
+     * @param bl where to store the read chunk
+     * @param len the number of bytes to read
+     * @param off the offset to start reading from in the object
+     * @param snapid the snapid of the replica to read
+     * @param flags the flags applied to the underlying read op. Only
+     *        @c CEPH_OSD_OP_FLAG_FAILOK or 0 is allowed. Any flags
+     *        other than @c CEPH_OSD_OP_FLAG_FAILOK will be ignored.
+     *        If @c CEPH_OSD_OP_FLAG_FAILOK is specified, the read payload
+     *        will be returned despite of CRC mismatch or size mismatch.
+     *        Otherwise, @c -EIO will be returned instead.
+     * @param osdid from which osd the replica is read
+     * @param interval the PG interval when the inconsistency was reported
+     * @returns 0 on success, negative error code on failure
+     * @note Please make sure that the @c interval should be the one returned
+     *       by @c get_inconsistent_objects() or @c get_inconsistent_snapsets()
+     *       call. If the PG interval expires when the request is served by OSD,
+     *       @c -ERANGE will be returned.
+     */
+    int aio_repair_read(const std::string& oid, AioCompletion *c,
+                        bufferlist *pbl, size_t len, uint64_t off, uint64_t snapid, int flags,
+                        int32_t osdid, uint32_t interval);
+    /**
      * Asynchronously read existing extents from an object at a
      * particular snapshot
      *
@@ -1058,6 +1084,9 @@ namespace librados
     // compound object operations
     int operate(const std::string& oid, ObjectWriteOperation *op);
     int operate(const std::string& oid, ObjectReadOperation *op, bufferlist *pbl);
+    int aio_operate_repair_read(const std::string& oid, AioCompletion *c,
+                                ObjectReadOperation *op,
+                                uint32_t osdid, int32_t interval);
     int aio_operate(const std::string& oid, AioCompletion *c, ObjectWriteOperation *op);
     int aio_operate(const std::string& oid, AioCompletion *c, ObjectWriteOperation *op, int flags);
     /**
--- a/src/include/rados/librados.hpp
+++ b/src/include/rados/librados.hpp
@@ -758,6 +758,20 @@ namespace librados
 		  size_t write_len, uint64_t off);
     int read(const std::string& oid, bufferlist& bl, size_t len, uint64_t off);
     int repair_read(const std::string& oid, bufferlist& bl, size_t len, uint64_t off, int flags, int32_t osdid, uint32_t epoch);
+    /**
+     * Rewrite the object with the replica hosted by specified osd
+     *
+     * @param bad_shards a list of <osd_id, shard_id> pairs. it is a black list
+     *                   of if the object shards/replicas, we will avoid using
+     *                   shards/replicas from them. if the object specified by @c
+     *                   oid is located in a replicated pool, the shard_id will
+     *                   be ignored.
+     * @param version the version of rewritten object
+     * @param what the flags indicating what we will copy
+     */
+    int repair_copy(const std::string& oid, uint64_t version, uint32_t what,
+                    const std::vector<osd_shard_t>& bad_shards,
+                    uint32_t epoch);
     int remove(const std::string& oid);
     int remove(const std::string& oid, int flags);
     int trunc(const std::string& oid, uint64_t size);
@@ -992,6 +1006,9 @@ namespace librados
     int aio_repair_read(const std::string& oid, AioCompletion *c,
                         bufferlist *pbl, size_t len, uint64_t off, uint64_t snapid, int flags,
                         int32_t osdid, uint32_t interval);
+    int aio_repair_copy(const std::string& oid, AioCompletion *c,
+         uint64_t version, uint32_t what,
+         const std::vector<osd_shard_t>& bad_osd_shards, uint32_t epoch);
     /**
      * Asynchronously read existing extents from an object at a
      * particular snapshot
--- a/src/librados/IoCtxImpl.cc
+++ b/src/librados/IoCtxImpl.cc
@@ -780,6 +780,33 @@ int librados::IoCtxImpl::operate_read(const object_t& oid,
   return r;
 }
 
+int librados::IoCtxImpl::aio_operate_repair_read(const object_t& oid,
+				      ::ObjectOperation *o,
+				      AioCompletionImpl *c,
+				      uint64_t snapid,
+				      int flags, int32_t osdid, epoch_t e)
+{
+  auto onack = new C_aio_Complete(c);
+
+  c->is_read = true;
+  c->io = this;
+
+  // Prepend assert_interval op
+  o->assert_interval(e);
+  for (int i = o->size() - 1; i > 0; i--) {
+    std::swap(o->ops[i - 1], o->ops[i]);
+    std::swap(o->out_bl[i - 1], o->out_bl[i]);
+    std::swap(o->out_rval[i - 1], o->out_rval[i]);
+  }
+  Objecter::Op *objecter_op = objecter->prepare_read_op(oid, oloc,
+	                                      *o, snapid, NULL,
+	                                      flags | CEPH_OSD_FLAG_REPAIR_READS | CEPH_OSD_FLAG_IGNORE_OVERLAY,
+	                                      onack, &c->objver);
+  objecter_op->target.forced_osd = osdid;
+  objecter->op_submit(objecter_op, &c->tid);
+  return 0;
+}
+
 int librados::IoCtxImpl::aio_operate_read(const object_t &oid,
 					  ::ObjectOperation *o,
 					  AioCompletionImpl *c,
--- a/src/librados/IoCtxImpl.cc
+++ b/src/librados/IoCtxImpl.cc
@@ -1416,6 +1416,48 @@ int librados::IoCtxImpl::read(const object_t& oid,
   return bl.length();
 }
 
+int librados::IoCtxImpl::repair_copy(const object_t& oid,
+				     uint64_t ver, uint32_t what,
+				     const std::vector<pg_shard_t>& bad_shards,
+				     epoch_t epoch)
+{
+  ::ObjectOperation op;
+  op.assert_interval(epoch);
+  op.assert_version(ver);
+  op.repair_copy(what, bad_shards);
+  return operate(oid, &op, nullptr, CEPH_OSD_FLAG_REPAIR_WRITES);
+}
+
+int librados::IoCtxImpl::aio_repair_copy(const object_t& oid, AioCompletionImpl *comp,
+					 uint64_t ver, uint32_t what,
+					 const std::vector<pg_shard_t>& bad_shards,
+					 epoch_t epoch)
+{
+  ::ObjectOperation op;
+  op.assert_interval(epoch);
+  op.assert_version(ver);
+  op.repair_copy(what, bad_shards);
+
+  auto onack = new C_aio_Complete(comp);
+
+  comp->io = this;
+  queue_aio_write(comp);
+
+  auto objecter_op = objecter->prepare_mutate_op(oid, oloc, op, snapc,
+						 ceph::real_clock::now(),
+						 CEPH_OSD_FLAG_REPAIR_WRITES,
+						 onack, &comp->objver);
+
+  int src_osd = objecter->pick_random_osd(objecter_op->target, bad_shards);
+  if (src_osd == -1) {
+    onack->complete(-EINVAL);
+    return -EINVAL;
+  }
+  objecter_op->target.forced_osd = src_osd;
+  objecter->op_submit(objecter_op, &comp->tid);
+  return 0;
+}
+
 int librados::IoCtxImpl::mapext(const object_t& oid,
 				uint64_t off, size_t len,
 				std::map<uint64_t,uint64_t>& m)
--- a/src/librados/librados.cc
+++ b/src/librados/librados.cc
@@ -1266,6 +1266,26 @@ int librados::IoCtx::read(const std::string& oid, bufferlist& bl, size_t len, ui
   return io_ctx_impl->read(obj, bl, len, off);
 }
 
+int librados::IoCtx::repair_read(const std::string& oid, bufferlist& bl,
+				 size_t len, uint64_t off, int flags,
+				 int32_t osdid, epoch_t epoch)
+{
+  auto c = librados::Rados::aio_create_completion();
+  int r = aio_repair_read(oid, c, &bl, len, off, CEPH_NOSNAP, flags, osdid, epoch);
+  if (r < 0) {
+    c->release();
+    return r;    
+  }
+  c->wait_for_complete();
+  r = c->get_return_value();
+  if (r >= 0)
+    io_ctx_impl->set_sync_op_version(c->get_version64());
+  c->release();
+  if (r < 0)
+    return r;
+  return bl.length();
+}
+
 int librados::IoCtx::remove(const std::string& oid)
 {
   object_t obj(oid);
@@ -1541,6 +1561,16 @@ int librados::IoCtx::operate(const std::string& oid, librados::ObjectReadOperati
 {
   object_t obj(oid);
   return io_ctx_impl->operate_read(obj, &o->impl->o, pbl);
+}
+
+int librados::IoCtx::aio_operate_repair_read(const std::string& oid,
+					     AioCompletion *c,
+					     librados::ObjectReadOperation *o,
+					     uint32_t osdid, int32_t epoch)
+{
+  return io_ctx_impl->aio_operate_repair_read(oid, &o->impl->o, c->pc,
+					      io_ctx_impl->snap_seq, 0,
+					      osdid, epoch);
 }
 
 int librados::IoCtx::aio_operate(const std::string& oid, AioCompletion *c,
@@ -1928,6 +1958,21 @@ int librados::IoCtx::aio_sparse_read(const std::string& oid, librados::AioComple
 {
   return io_ctx_impl->aio_sparse_read(oid, c->pc,
 				      m, data_bl, len, off, snapid);
+}
+
+int librados::IoCtx::aio_repair_read(const std::string& oid, librados::AioCompletion *c,
+				     bufferlist *pbl, size_t len, uint64_t off,
+				     uint64_t snapid, int flags, int32_t osdid, epoch_t e)
+{
+  if (len > (size_t) INT_MAX)
+    return -EDOM;
+  ::ObjectOperation rd;
+  io_ctx_impl->prepare_assert_ops(&rd);
+  rd.read(off, len, pbl, NULL, NULL);
+  // Only 1 flag can be passed from user
+  rd.set_last_op_flags(flags & CEPH_OSD_OP_FLAG_FAILOK);
+  return io_ctx_impl->aio_operate_repair_read(oid, &rd, c->pc,
+					      snapid, 0, osdid, e);
 }
 
 int librados::IoCtx::aio_write(const std::string& oid, librados::AioCompletion *c,
--- a/src/librados/librados.cc
+++ b/src/librados/librados.cc
@@ -2236,6 +2236,21 @@ void librados::IoCtx::set_assert_version(uint64_t ver)
   io_ctx_impl->set_assert_version(ver);
 }
 
+static int translate_repair_copy(int what)
+{
+  int translated = 0;
+  if (what & librados::repair_copy_t::DATA) {
+    translated |= CEPH_REPAIR_COPY_DATA;
+  }
+  if (what & librados::repair_copy_t::OMAP) {
+    translated |= CEPH_REPAIR_COPY_OMAP;
+  }
+  if (what & librados::repair_copy_t::ATTR) {
+    translated |= CEPH_REPAIR_COPY_ATTR;
+  }
+  return translated;
+}
+
 int librados::IoCtx::repair_copy(const string& oid, uint64_t version, uint32_t what,
 				 const vector<osd_shard_t>& bad_shards,
 				 uint32_t epoch)
@@ -2244,7 +2259,9 @@ int librados::IoCtx::repair_copy(const string& oid, uint64_t version, uint32_t w
   for (const auto& bad_shard : bad_shards) {
     shards.emplace_back(bad_shard.osd, shard_id_t(bad_shard.shard));
   }
-  return io_ctx_impl->repair_copy(object_t(oid), version, what, shards, epoch);
+  return io_ctx_impl->repair_copy(object_t(oid), version,
+				  translate_repair_copy(what),
+				  shards, epoch);
 }
 
 int librados::IoCtx::aio_repair_copy(const std::string& oid, AioCompletion *c,
--- a/src/librados/librados.cc
+++ b/src/librados/librados.cc
@@ -2273,7 +2273,8 @@ int librados::IoCtx::aio_repair_copy(const std::string& oid, AioCompletion *c,
   for (const auto& bad_shard : bad_shards) {
     shards.emplace_back(bad_shard.osd, shard_id_t(bad_shard.shard));
   }
-  return io_ctx_impl->aio_repair_copy(object_t(oid), c->pc, version, what,
+  return io_ctx_impl->aio_repair_copy(object_t(oid), c->pc, version,
+				      translate_repair_copy(what),
 				      shards, epoch);
 }
 
--- a/src/librados/librados.cc
+++ b/src/librados/librados.cc
@@ -2236,6 +2236,30 @@ void librados::IoCtx::set_assert_version(uint64_t ver)
   io_ctx_impl->set_assert_version(ver);
 }
 
+int librados::IoCtx::repair_copy(const string& oid, uint64_t version, uint32_t what,
+				 const vector<osd_shard_t>& bad_shards,
+				 uint32_t epoch)
+{
+  vector<pg_shard_t> shards;
+  for (const auto& bad_shard : bad_shards) {
+    shards.emplace_back(bad_shard.osd, shard_id_t(bad_shard.shard));
+  }
+  return io_ctx_impl->repair_copy(object_t(oid), version, what, shards, epoch);
+}
+
+int librados::IoCtx::aio_repair_copy(const std::string& oid, AioCompletion *c,
+				     uint64_t version, uint32_t what,
+				     const vector<osd_shard_t>& bad_shards,
+				     uint32_t epoch)
+{
+  vector<pg_shard_t> shards;
+  for (const auto& bad_shard : bad_shards) {
+    shards.emplace_back(bad_shard.osd, shard_id_t(bad_shard.shard));
+  }
+  return io_ctx_impl->aio_repair_copy(object_t(oid), c->pc, version, what,
+				      shards, epoch);
+}
+
 void librados::IoCtx::locator_set_key(const string& key)
 {
   io_ctx_impl->oloc.key = key;
--- a/src/osd/PG.cc
+++ b/src/osd/PG.cc
@@ -5271,7 +5271,9 @@ bool PG::can_discard_op(OpRequestRef& op)
     return true;
   }
 
-  if (m->get_map_epoch() < pool.info.last_force_op_resend &&
+  // Don't care about last_force_op_resend for recovery reads
+  if (!(m->get_flags() & CEPH_OSD_FLAG_REPAIR_READS) &&
+      m->get_map_epoch() < pool.info.last_force_op_resend &&
       m->get_connection()->has_feature(CEPH_FEATURE_OSD_POOLRESEND)) {
     dout(7) << __func__ << " sent before last_force_op_resend "
 	    << pool.info.last_force_op_resend << ", dropping" << *m << dendl;
--- a/src/osd/PG.h
+++ b/src/osd/PG.h
@@ -692,7 +692,8 @@ protected:
   eversion_t  oldest_update; // acting: lowest (valid) last_update in active set
   map<pg_shard_t, pg_info_t>    peer_info;   // info from peers (stray or prior)
   set<pg_shard_t> peer_purged; // peers purged
-  map<pg_shard_t, pg_missing_t> peer_missing;
+  map<pg_shard_t, pg_missing_t> peer_missing; // also used by read op issued by
+                                              // repair-copy
   set<pg_shard_t> peer_log_requested;  // logs i've requested (and start stamps)
   set<pg_shard_t> peer_missing_requested;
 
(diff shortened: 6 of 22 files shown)


tchaikov:
tests on their way.

tchaikov:
per sam

> i think it would be simpler to just allow the repair_write call to specify a set of bad shards. For replicated pool, we simply choose one which is not in that set . For EC pools, we use that information to avoid bad shards.

tchaikov:
@dzafman  and @athanatos could you take a look ? next, i will add the test for ec pool `repair-copy` .

src/osd/ReplicatedPG.cc
```diff
@@ -1357,6 +1357,35 @@ int ReplicatedPG::do_scrub_ls(MOSDOp *m, OSDOp *osd_op)
...
 
+// it is pretty much the same as CEPH_OSD_OP_COPY_FROM, but it requires user to
+// specify the src osd, and uses repair_read() to read the object payload, also
+// the src oid is always identical to the dest oid (the one in OpContext)
+int ReplicatedPG::do_repair_copy(OpContext *ctx, const OSDOp& osd_op, bufferlist::iterator& bp)
+{
+  uint32_t what;
+  vector<pg_shard_t> bad_shards;
+  try {
+    ::decode(what, bp);
```
tchaikov: maybe i should wrap it in an arg struct, and add the `encode()/decode()` methods for it.

src/tools/rados/rados.cc
```diff
@@ -1397,7 +1398,10 @@ static void dump_inconsistent(const inconsistent_obj_t& inc,
   f.open_array_section("shards");
   for (auto osd_shard : inc.shards) {
     f.open_object_section("shard");
-    f.dump_int("osd", osd_shard.first);
+    f.dump_int("osd", osd_shard.first.first);
+    auto shard = osd_shard.first.second;
+    if (shard)
```
tchaikov: should be `if (shard == -1)`

src/osd/ReplicatedPG.cc
```diff
@@ -7075,7 +7075,17 @@ int ReplicatedPG::fill_in_copy_get(
       cursor.data_complete = true;
       dout(20) << " got data" << dendl;
     }
-    assert(cursor.data_offset <= oi.size);
```
athanatos: I don't follow how this can happen.  Isn't data_offset always 0 or a result of a previous copy_get?

src/osd/ReplicatedPG.cc
athanatos: Yeah, probably.

src/include/rados/librados.hpp
```diff
@@ -712,6 +712,14 @@ namespace librados
...
     int read(const std::string& oid, bufferlist& bl, size_t len, uint64_t off);
     int repair_read(const std::string& oid, bufferlist& bl, size_t len, uint64_t off, int flags, int32_t osdid, uint32_t e);
+    /**
+     * Rewrite the object with the replica hosted by specified osd
+     *
+     * @param osd from which OSD we will copy the data
+     * @param version the version of rewritten object
+     * @param what the flags indicating what we will copy
+     */
+    int repair_copy(const std::string& oid, uint64_t version, uint32_t what, int32_t osd, uint32_t epoch);
```
athanatos: This bit should probably be squashed into the later commit which changes it to take a set of bad shards, right?

src/include/rados/librados.hpp
```diff
@@ -715,11 +715,17 @@ namespace librados
...
+     *                  of if the object shards/replicas, we will avoid using
+     *                  shards/replicas from them. if the object specified by @c
+     *                  oid is located in a replicated pool, the shard_id will
+     *                  be ignored.
      * @param version the version of rewritten object
      * @param what the flags indicating what we will copy
      */
-    int repair_copy(const std::string& oid, uint64_t version, uint32_t what, int32_t osd, uint32_t epoch);
+    int repair_copy(const std::string& oid, uint64_t version, uint32_t what,
+                    const std::vector<std::pair<int32_t, int8_t>>& bad_osds,
```
athanatos: Do we need to expose the <osd, shard> tuple anywhere else?  Probably worth exposing a librados type to wrap it.

src/osd/ReplicatedPG.h
```diff
@@ -1346,14 +1355,18 @@ class ReplicatedPG : public PG, public PGBackend::Listener {
    *
    * @param cb: The CopyCallback to be activated when the copy is complete
    * @param obc: The ObjectContext we are copying into
-   * @param src: The source object
    * @param oloc: the source object locator
    * @param version: the version of the source object to copy (0 for any)
+   * @param what: the flags composed of CopyOp::COPY_* indicating what to copy
```
tchaikov: revert this change.

src/osd/ReplicatedPG.h
```diff
@@ -1346,14 +1355,18 @@ class ReplicatedPG : public PG, public PGBackend::Listener {
    *
    * @param cb: The CopyCallback to be activated when the copy is complete
    * @param obc: The ObjectContext we are copying into
-   * @param src: The source object
```
tchaikov: revert this change.

src/osd/ReplicatedPG.h
```diff
@@ -1346,14 +1355,18 @@ class ReplicatedPG : public PG, public PGBackend::Listener {
...
    * @param oloc: the source object locator
    * @param version: the version of the source object to copy (0 for any)
+   * @param what: the flags composed of CopyOp::COPY_* indicating what to copy
    */
-  void start_copy(CopyCallback *cb, ObjectContextRef obc, hobject_t src,
-		  object_locator_t oloc, version_t version, unsigned flags,
+  void start_copy(CopyCallback *cb, ObjectContextRef obc,
+		  hobject_t src,
+		  object_locator_t oloc, version_t version,
+		  unsigned flags,
```
tchaikov: and this.

src/osd/ReplicatedPG.cc
tchaikov: see https://github.com/ceph/ceph/pull/9203/files#diff-72747d40a424e7b5404366b557ff12a3L7053 . it could be greater than expected, if we are reading a bad replica which is larger than what object-info claims.

athanatos: I see, gotcha.

tchaikov:
@dzafman i introduced a new librados type `osd_shard_t ` per the suggest of @athanatos, and updated the commits for adding user_version and shard_id. maybe you want to cherry-pick them again?

tchaikov:
i am still testing the test for ec pool.

tchaikov:
@athanatos @dzafman comments addressed. because these two tests are time consuming, i will move them into qa units. so we can exercise them in the integration tests. what do you think?

athanatos:
I would be fine with moving them into qa units.

dachary:
needs rebase

dachary:
setting DNM since the base PR is not merged yet

tchaikov:
@dzafman and @athanatos could you take a look again? once it looks good, i will move replace `faster_objectstore_tool()` with `objectstore_tool()` and move the newly added tests into "qa".

changelog

- @athanatos 's comments at https://github.com/ceph/ceph/pull/8931#issuecomment-219499271 are addressed
- rebased against master
- updated the commits from #8931 accordingly

tchaikov:
```
/home/jenkins-build/build/workspace/ceph-pull-requests/qa/workunits/ceph-helpers.sh:1151: wait_for_clean:  return 1
/home/jenkins-build/build/workspace/ceph-pull-requests/src/test/erasure-code/test-erasure-eio.sh:137: rados_put_get:  return 1
/home/jenkins-build/build/workspace/ceph-pull-requests/src/test/erasure-code/test-erasure-eio.sh:172: rados_get_data_eio:  return 1
/home/jenkins-build/build/workspace/ceph-pull-requests/src/test/erasure-code/test-erasure-eio.sh:313: TEST_rados_get_with_subreadall_eio_shard_0:  return 1
```

tchaikov:
@dzafman i tested the failed test case locally and it passed.

tchaikov:
retest this please.

tchaikov:
rebased against master.

src/osd/PrimaryLogPG.cc
```diff
@@ -1529,6 +1529,46 @@ int PrimaryLogPG::do_scrub_ls(MOSDOp *m, OSDOp *osd_op)
...
+    dout(10) << "corrupted " << __func__ << " arg" << dendl;
+    return -EINVAL;
+  }
+  if (arg.bad_shards.empty() || !arg.what) {
+    return 0;
+  }
+  const auto& oi = ctx->obs->oi;
+  if (!ctx->copy_cb) {
+    // mark the bad shard missing so they are not chosen when repairing
+    if (pool.info.require_rollback()) {
```
athanatos: I think here you really mean pool.info.ec_pool().  Does it actually matter though?  Why not add it to the peer_missing sets unconditionally so that Replicated and EC work uniformly?

src/librados/IoCtxImpl.cc
```diff
@@ -1247,13 +1287,14 @@ int librados::IoCtxImpl::get_inconsistent_objects(const pg_t& pg,
...
-  op.scrub_ls(start_after, max_to_get, objects, interval, nullptr);
+  int rval = 0;
+  op.scrub_ls(start_after, max_to_get, objects, interval, &rval);
   object_locator_t oloc{poolid, pg.ps()};
   Objecter::Op *o = objecter->prepare_pg_read_op(
     oloc.hash, oloc, op, nullptr, CEPH_OSD_FLAG_PGOP, onack,
     nullptr, nullptr);
   objecter->op_submit(o, &c->tid);
-  return 0;
+  return rval;
```
tchaikov: fixed.

src/librados/IoCtxImpl.cc
```diff
@@ -1247,13 +1287,14 @@ int librados::IoCtxImpl::get_inconsistent_objects(const pg_t& pg,
   c->io = this;
 
   ::ObjectOperation op;
-  op.scrub_ls(start_after, max_to_get, objects, interval, nullptr);
+  int rval = 0;
+  op.scrub_ls(start_after, max_to_get, objects, interval, &rval);
```
tchaikov: right.

src/librados/IoCtxImpl.cc
```diff
@@ -780,6 +780,46 @@ int librados::IoCtxImpl::operate_read(const object_t& oid,
...
+  int size = o->ops.size();
+  o->out_bl.resize(size);
+  o->out_handler.resize(size);
+  o->out_rval.resize(size);
+  o->out_bl[1] = pbl;
+  o->ops[1].op.flags = op_flags;
+  Objecter::Op *objecter_op = objecter->prepare_read_op(oid, oloc,
+	                                      *o, snap_seq, NULL,
+	                                      flags | CEPH_OSD_FLAG_REPAIR_READS | CEPH_OSD_FLAG_IGNORE_OVERLAY,
+	                                      onack, &ver);
```
tchaikov: fixed.

tchaikov:
@athanatos all comments addressed and repushed.

src/librados/IoCtxImpl.cc
```diff
@@ -780,6 +780,44 @@ int librados::IoCtxImpl::operate_read(const object_t& oid,
...
+  o->ops[1].op.flags = op_flags;
+  Objecter::Op *objecter_op = objecter->prepare_read_op(oid, oloc,
+	                                      *o, snap_seq, NULL,
+	                                      flags | CEPH_OSD_FLAG_REPAIR_READS | CEPH_OSD_FLAG_IGNORE_OVERLAY,
+	                                      onack, &c->objver);
+  objecter_op->target.osd = osdid;
+  objecter_op->target.use_osd_epoch = true;
+  objecter_op->target.epoch = e;
+  objecter->op_submit(objecter_op, &c->tid);
+  return c->tid;
```
athanatos: I'm fairly sure the expectation is that the return value for these aio_operate* calls should be 0 if successfully initiated.  Getting the tid back I think would be unexpected.

src/osd/PrimaryLogPG.cc
```diff
@@ -1529,6 +1529,44 @@ int PrimaryLogPG::do_scrub_ls(MOSDOp *m, OSDOp *osd_op)
...
+  if (!ctx->copy_cb) {
+    // mark the bad shard missing so they are not chosen when repairing
+    for (const auto& bad_shard : arg.bad_shards) {
+      peer_missing[bad_shard].add(oi.soid, oi.version, eversion_t());
+    }
+    start_repair_copy(ctx, arg.what, arg.bad_shards);
+    return -EINPROGRESS;
+  } else {
+    // finish
+    if (pool.info.require_rollback()) {
```
athanatos: You unconditionally added them to the missing sets, right?  Seems like you need to unconditionally remove them?

tchaikov:
@athanatos all comments addressed and repushed.

src/osdc/Objecter.h
```diff
@@ -2328,6 +2340,13 @@ class Objecter : public md_config_obs_t, public Dispatcher {
     return i;
   }
 
+  /**
+   * get a random osd not listed in blacklist
+   * @note the primary osd will be returned even it is listed in the blacklist
+   *       if the target pool is a erasure pool, the primary osd is returned.
```
dzafman: Confusing because you mention primary osd twice.  Is this sufficient?

the primary osd will be returned even it is listed in the blacklist
if the target pool is a erasure pool

src/osd/PrimaryLogPG.cc
```diff
@@ -5946,6 +5987,24 @@ int PrimaryLogPG::do_osd_ops(OpContext *ctx, vector<OSDOp>& ops)
       }
       break;
 
+    case CEPH_OSD_OP_REPAIR_COPY:
+      ++ctx->num_write;
+      result = do_repair_copy(ctx, osd_op, bp);
```
dzafman: No tracepoint here?

src/tracing/osd.tp
```diff
@@ -790,3 +790,15 @@ TRACEPOINT_EVENT(osd, do_osd_op_post,
         ctf_integer_hex(int, result, result)
     )
 )
+
+TRACEPOINT_EVENT(osd, do_osd_op_pre_assert_interval,
```
dzafman: If you add tracepoint to repair copy don't forget to add here too.

src/tools/rados/rados.cc
```diff
@@ -322,6 +328,94 @@ static int do_get(IoCtx& io_ctx, RadosStriper& striper,
...
+      bad_shards.push_back(bad_shard);
+    } catch (const exception& e) {
+      cerr << "Unable to parse the shards specified: " << strs
+	   << ". " << e.what() << std::endl;
+      return -EINVAL;
+    }
+  }
+  epoch_t epoch = atoi(epoch_str);
+  uint64_t version = atoi(version_str);
+  uint32_t what = (librados::repair_copy_t::DATA |
```
dzafman: Shouldn't we include "omap" in the default or just "data" as the default?  I feel like the command line format is limiting because you can only specify 1 of the elements.

src/librados/librados.cc
```diff
@@ -1930,6 +1965,20 @@ int librados::IoCtx::aio_sparse_read(const std::string& oid, librados::AioComple
 				      m, data_bl, len, off, snapid);
 }
 
+int librados::IoCtx::aio_repair_read(const std::string& oid, librados::AioCompletion *c,
+				     bufferlist *pbl, size_t len, uint64_t off,
+				     uint64_t /*snapid*/, int flags, int32_t osdid, epoch_t e)
```
tchaikov: can we track this in the tracker and support it in another PR?

tchaikov:
@dzafman comments addressed and repushed.

dzafman:
@tchaikov Everything looks good, but the tracepoint change isn't building

tchaikov:
changelog

- squashed related commits,
- rebased against master

tchaikov:
being tested at http://pulpito.ceph.com/kchai-2017-01-30_09:44:58-rados-wip-kefu-testing---basic-mira/

tchaikov:
@athanatos the tests passed, is this change good to merge?

tchaikov:
hey @jdurgin could you please take a look?

jdurgin:
certainly, I started earlier this week and will finish looking through it today

src/librados/librados.cc
```diff
@@ -1266,6 +1266,26 @@ int librados::IoCtx::read(const std::string& oid, bufferlist& bl, size_t len, ui
   return io_ctx_impl->read(obj, bl, len, off);
 }
 
+int librados::IoCtx::repair_read(const std::string& oid, bufferlist& bl,
+				 size_t len, uint64_t off, int flags,
+				 int32_t osdid, epoch_t e)
+{
+  auto c = librados::Rados::aio_create_completion();
```
jdurgin: this could be simplified by calling operate_repair_read()

src/librados/librados.cc
jdurgin: is snap_seq passed in by aio_operate_repair_read() not respected?

src/osd/PG.cc
```diff
@@ -5270,7 +5270,9 @@ bool PG::can_discard_op(OpRequestRef& op)
     return true;
   }
 
-  if (m->get_map_epoch() < pool.info.last_force_op_resend &&
+  // Don't care about last_force_op_resend for recovery reads
```
jdurgin: why not? seems like a bug

src/include/rados/librados.hpp
```diff
@@ -1058,6 +1078,8 @@ namespace librados
     // compound object operations
     int operate(const std::string& oid, ObjectWriteOperation *op);
     int operate(const std::string& oid, ObjectReadOperation *op, bufferlist *pbl);
+    int operate_repair_read(const std::string& oid, librados::ObjectReadOperation *o, bufferlist *pbl,
+		       uint32_t osdid, int32_t epoch, int op_flags);
```
tchaikov: `op_flags` is different from `flags` in this context, `flags` is the bitset of `librados::OPERATION_*`, and it's a part of our public API. while `op_flags` is the internal representation of it, which is part of the rados protocol, and is understood by OSD. we use `translate_flags()` to translate from the former to the later. so i'd prefer to keep `op_flags` , so it's more explicit, and less misleading. 

what do you think?

src/librados/librados.cc
tchaikov:
```diff
@@ -1268,19 +1268,16 @@ int librados::IoCtx::read(const std::string& oid, bufferlist& bl, size_t len, ui

 int librados::IoCtx::repair_read(const std::string& oid, bufferlist& bl,
                                 size_t len, uint64_t off, int flags,
-                                int32_t osdid, epoch_t e)
+                                int32_t osdid, epoch_t epoch)
 {
-  auto c = librados::Rados::aio_create_completion();
-  int r = aio_repair_read(oid, c, &bl, len, off, 0, flags, osdid, e);
-  if (r < 0) {
-    c->release();
-    return r;
-  }
-  c->wait_for_complete();
-  r = c->get_return_value();
-  if (r >= 0)
-    io_ctx_impl->set_sync_op_version(c->get_version64());
-  c->release();
+  if (len > (size_t) INT_MAX)
+    return -EDOM;
+  ObjectReadOperation op;
+  io_ctx_impl->prepare_assert_ops(&op.impl->o);
+  op.read(off, len, &bl, nullptr);
+  // Only 1 flag can be passed from user
+  flags &= CEPH_OSD_OP_FLAG_FAILOK;
+  int r = operate_repair_read(oid, &op, &bl, osdid, epoch, flags);
   if (r < 0)
     return r;
   return bl.length();
```

i'd say it's more of a trade-off. probably, the call path looks more reasonable this way: 

before this change:
- repair_read() => aio_repair_read()
- aio_repair_read() => aio_operate_repair_read()

after this change:
- repair_read() => operate_repair_read() => aio_operate_repair_read(
- aio_repair_read() => aio_operate_repair_read()

src/librados/IoCtxImpl.cc
```diff
@@ -1389,6 +1427,44 @@ int librados::IoCtxImpl::read(const object_t& oid,
...
+  op.assert_version(ver);
+  op.repair_copy(what, bad_shards);
+
+  auto ut = ceph::real_clock::now();
+  auto onack = new C_aio_Complete(comp);
+
+  comp->io = this;
+  queue_aio_write(comp);
+
+  auto objecter_op = objecter->prepare_mutate_op(oid, oloc, op, snapc, ut, 0,
```
tchaikov: fixed.

src/osd/PG.cc
tchaikov: @dzafman neither do i understand this. could you elaborate a little bit?

tchaikov:
apart from https://github.com/ceph/ceph/pull/9203#discussion_r99285234, all other comments addressed.

src/librados/IoCtxImpl.cc
```diff
@@ -799,16 +797,12 @@ int librados::IoCtxImpl::aio_operate_repair_read(const object_t& oid,
...
-  tmp.op.op = CEPH_OSD_OP_ASSERT_INTERVAL;
-  tmp.op.assert_interval.epoch = e;
-  o->ops.insert(o->ops.begin(), tmp);
-  int size = o->ops.size();
-  o->out_bl.resize(size);
-  o->out_handler.resize(size);
-  o->out_rval.resize(size);
-  o->out_bl[1] = pbl;
-  o->ops[1].op.flags = op_flags;
+  o->assert_interval(e);
```
tchaikov: @dzafman i changed the way how you prepend the op, because i think this is less hackery.

src/librados/IoCtxImpl.cc
```diff
@@ -780,6 +780,40 @@ int librados::IoCtxImpl::operate_read(const object_t& oid,
...
+    return 0;
+
+  auto onack = new C_aio_Complete(c);
+
+  c->is_read = true;
+  c->io = this;
+
+  int op = o->ops[0].op.op;
+  ldout(client->cct, 10) << ceph_osd_op_name(op) << " oid=" << oid << " nspace=" << oloc.nspace << dendl;
+  // Prepend assert_interval op
```
jdurgin: instead of modifying the ObjectOperation, you can use the extra_ops arg for the Objecter prepare_* helpers.
adding an Objecter::prepare_repair_read() helper here would be good too

src/osdc/Objecter.cc
```diff
@@ -2736,6 +2774,9 @@ int Objecter::_calc_target(op_target_t *t, bool any_change)
 	t->sort_bitwise,
 	sort_bitwise,
 	pg_t(prev_seed, pgid.pool(), pgid.preferred()))) {
+    // If the interval has changed, then don't resend for repair read
```
jdurgin: won't this make the librados op hang? seems like we should resend, and let the osd give us the error if the interval changed

src/osdc/Objecter.h
```diff
@@ -1206,10 +1216,12 @@ class Objecter : public md_config_obs_t, public Dispatcher {
 
     bool used_replica = false;
     bool paused = false;
+    bool use_osd_epoch = false;
 
     int osd = -1;      ///< the final target osd, or -1
 
     epoch_t last_force_resend = 0;
+    epoch_t epoch; ///< Required epoch
```
jdurgin: how about naming it required_interval (since we care about intervals changing, not a new map that doesn't affect our pg), defaulting to 0, and getting rid of use_osd_epoch?

src/librados/librados.cc
jdurgin: I agree, the first way was duplicating fewer unique things.

In general these librados.cc methods should be doing as little as possible - just the necessities to translate the C++ and C interfaces to a common internal call in IoCtxImpl. You'll notice nothing else in librados.cc calls prepare_assert_ops() for example, all that sort of common code is in IoCtxImpl. This bit can be cleaned up when we add the C interface though, this pr doesn't have to do everything at once.

src/osdc/Objecter.cc
```diff
@@ -2642,6 +2642,44 @@ int64_t Objecter::get_object_pg_hash_position(int64_t pool, const string& key,
...
+    for (auto& shard : blacklist) {
+      auto found = find(acting.begin(), acting.end(), shard.osd);
+      if (found != acting.end()) {
+	acting.erase(found);
+      }
+    }
+    if (acting.empty()) {
+      return -1;
+    }
+    int p = rand() % acting.size();
```
tchaikov: okay. but i think it would fine if we always use the same sequence of the pseudo random number in every Objecter. but i don't feel strong either way. will use `random()` and `srandom()` instead for better randomness.

> POSIX.1-2008 marks rand_r() as obsolete.

quote from rand(3).

src/test/osd/osd-scrub-repair.sh
```diff
@@ -87,6 +92,202 @@ function TEST_corrupt_and_repair_replicated() {
...
+            ;;
+        path)
+            echo $shard
+    esac
+}
+
+#
+# Corrupt one copy of a replicated pool and repair it manually
+#
+# TODO: move this test to ceph-qa-suite
```
jdurgin: we need at least a ceph_test_librados_api* test to get coverage with the various qa-suite options (esp. failure injection for the objecter resending stuff)

ideally there would be RadosModel tests for the new ops as well

src/osdc/Objecter.cc
jdurgin: sounds good. it doesn't matter a ton for this use, but better to be safe for future uses. as long as it's not the ridiculous c++ api http://www.pcg-random.org/posts/cpps-random_device.html

src/include/rados/librados.hpp
```diff
@@ -1058,6 +1078,8 @@ namespace librados
     // compound object operations
     int operate(const std::string& oid, ObjectWriteOperation *op);
     int operate(const std::string& oid, ObjectReadOperation *op, bufferlist *pbl);
+    int operate_repair_read(const std::string& oid, librados::ObjectReadOperation *o,
+		       uint32_t osdid, int32_t epoch);
     int aio_operate(const std::string& oid, AioCompletion *c, ObjectWriteOperation *op);
```
tchaikov: will expose `aio_operate_repair_read()` instead. user is always able to build up his/her own `operate_repair_read()` this method.

src/osd/PrimaryLogPG.cc
```diff
@@ -1529,6 +1529,47 @@ int PrimaryLogPG::do_scrub_ls(MOSDOp *m, OSDOp *osd_op)
...
+	     oi.soid.snap.val,
+	     arg.what);
+
+  if (arg.bad_shards.empty() || !arg.what) {
+    return 0;
+  }
+  if (!ctx->copy_cb) {
+    // mark the bad shard missing so they are not chosen when repairing
+    for (const auto& bad_shard : arg.bad_shards) {
+      peer_missing[bad_shard].add(oi.soid, oi.version, eversion_t());
```
tchaikov: erasure pool select source shards from (acting_set -  peer_missing).

i don't follow, maybe the question should be "what makes it unsafe to use peer_missing in this way?".

src/include/rados/librados.hpp
```diff
@@ -757,6 +757,21 @@ namespace librados
     int writesame(const std::string& oid, bufferlist& bl,
 		  size_t write_len, uint64_t off);
     int read(const std::string& oid, bufferlist& bl, size_t len, uint64_t off);
+    int repair_read(const std::string& oid, bufferlist& bl, size_t len, uint64_t off, int flags, int32_t osdid, uint32_t e);
```
tchaikov: i will add the C and python bindings once the C++ version is stablized.

src/osdc/Objecter.cc
tchaikov: right! and it's simpler.

src/librados/IoCtxImpl.cc
```diff
@@ -780,6 +780,40 @@ int librados::IoCtxImpl::operate_read(const object_t& oid,
...
 }
 
+int librados::IoCtxImpl::aio_operate_repair_read(const object_t& oid,
+				      ::ObjectOperation *o,
+				      AioCompletionImpl *c,
+				      uint64_t snapid,
+				      int flags, int32_t osdid, epoch_t e)
+{
+  if (!o->size())
+    return 0;
```
tchaikov: removed.

src/include/rados/librados.hpp
```diff
@@ -757,6 +757,21 @@ namespace librados
     int writesame(const std::string& oid, bufferlist& bl,
 		  size_t write_len, uint64_t off);
     int read(const std::string& oid, bufferlist& bl, size_t len, uint64_t off);
+    int repair_read(const std::string& oid, bufferlist& bl, size_t len, uint64_t off, int flags, int32_t osdid, uint32_t epoch);
```
tchaikov: not really, the result will be stored to the operation's own rval. so it's still good to use it?

src/osdc/Objecter.h
tchaikov: maybe we can just leave the interval checking to `assert_interval` op, and instead of using `required_interval` or `use_osd_epoch`, we can have a `force_use_osd` to specify the designated osd so object will always use it as long as it's up.

src/osd/PrimaryLogPG.cc
```diff
@@ -1866,7 +1908,8 @@ void PrimaryLogPG::do_op(OpRequestRef& op)
   }
 
   // missing object?
-  if (is_unreadable_object(head)) {
+  if (is_unreadable_object(head) &&
+      !(m->get_flags() & CEPH_OSD_FLAG_REPAIR_READS)) {
```
tchaikov: done.

src/osdc/Objecter.cc
```diff
@@ -2642,6 +2642,44 @@ int64_t Objecter::get_object_pg_hash_position(int64_t pool, const string& key,
   return p->raw_hash_to_pg(p->hash_key(key, ns));
 }
 
+int32_t Objecter::pick_random_osd(const op_target_t& t,
+				  const vector<pg_shard_t>& blacklist) const
+{
+  const pg_pool_t *pi = osdmap->get_pg_pool(t.base_oloc.pool);
```
tchaikov: done.

src/osd/PG.h
```diff
@@ -692,7 +692,8 @@ class PG : protected DoutPrefixProvider {
   eversion_t  oldest_update; // acting: lowest (valid) last_update in active set
   map<pg_shard_t, pg_info_t>    peer_info;   // info from peers (stray or prior)
   set<pg_shard_t> peer_purged; // peers purged
-  map<pg_shard_t, pg_missing_t> peer_missing;
+  map<pg_shard_t, pg_missing_t> peer_missing; // peer_missing[primary] is used
```
tchaikov: fixed.

src/librados/IoCtxImpl.cc
tchaikov: > instead of modifying the ObjectOperation, you can use the extra_ops arg for the Objecter prepare_* helpers.

it's a different use case. `prepare_read_op()` prepend the existing ops in `extra_ops` to the new read op allocated in it. but we want to prepend an `assert_interval` op to the existing read ops in `o`.

> adding an Objecter::prepare_repair_read() helper here would be good too

okay.

src/librados/IoCtxImpl.cc
```diff
@@ -1389,6 +1423,45 @@ int librados::IoCtxImpl::read(const object_t& oid,
...
+  op.assert_interval(epoch);
+  op.assert_version(ver);
+  op.repair_copy(what, bad_shards);
+
+  auto onack = new C_aio_Complete(comp);
+
+  comp->io = this;
+  queue_aio_write(comp);
+
+  auto objecter_op = objecter->prepare_mutate_op(oid, oloc, op, snapc,
```
tchaikov: `Objecter::init_ops()` prepend the ops in given `ObjectOperation` into the new ops.  like
```
[extra,extra,extra,to-be-filled]
```
`prepare_read_op()` fills up the last "to-be-filled slot with a newly created read op.

but in the repair-read case, what we want is
```
[assert-interval,extra,extra,extra]
```
where extra is the read ops passed in by caller. to use `init_ops()` we need to create an `ObjectOperation`, and put assert-interval into it first, and copy the ops in `ObjectOperation` to the ops filled by `init_ops()`. i don't see the advantage of this approach.

src/osd/PrimaryLogPG.cc
```diff
@@ -7110,6 +7185,45 @@ void PrimaryLogPG::start_copy(CopyCallback *cb, ObjectContextRef obc,
...
+  auto found = copy_ops.find(dest);
+  if (found != copy_ops.end()) {
+    cancel_copy(found->second, false);
+  }
+  // we expect the client side to prepend an assert_version() call before
+  // this op, so just pass 0 as the version.
+  auto cb = new CopyFromCallback(ctx);
+  ctx->copy_cb = cb;
+  uint32_t to_copy = 0;
+  if (what & librados::repair_copy_t::DATA)
```
tchaikov: i guess it's fine at this moment: the internal representation can share with the same definitions defined in public interface. and i think it's desirable. if they diverge over the time, we certainly have the luxury to use another set of protocol internally and do the translation in librados. and we are also using other stuff in the `librados` in `PrimaryLogPG.cc` if you search `librados` in that source file.

doc/rados/troubleshooting/troubleshooting-pg.rst
```diff
@@ -452,9 +463,28 @@ some predefined criteria. But this does not always work. For example, the stored
...
+or does not make sense at all, you will need to download the payload of the shards
+and inspect them manually::
+
+  rados -p rbd --force repair-get foo 0 14 foo.osd-0
+
+This command fetches the replica of ``foo`` to the local file ``foo.osd.0``
+from osd.0 at epoch 14. If you are happy with this replica after looking at
+it, you might want to overwrite bad replicas with good ones::
+
+  rados -p rbd repair-copy foo 2 14 1
```
tchaikov: if the interval changes or it is overwritten with new bits, how can we know where the object is not located at, or if its replicas/shards are consistent or not without a scrubbing + get-inconsistent call?

tchaikov:
@jdurgin are you good with the fixes in the `[to-be-squashed]` commits? if yes, i will squash them into the corresponding commits. so i can have less overhead maintaining the commits.

src/osd/PrimaryLogPG.cc
jdurgin: I'd consider those other uses bugs - the fadvise flags in fact have CEPH_OSD_OP_FLAG...* constants defined, and that's important since that header is shared with the kernel.

The other current user is the 2nd librados listing interface, which imo is another misuse of the librados namespace, since it's used in the wire protocol for pgnls. 

We shouldn't be able to break the wire protocol accidentally by changing a librados type.

tchaikov:
i will make sure all comments are addressed after squashing these "to-be-squashed" commits.

src/osdc/Objecter.cc
```diff
@@ -2812,9 +2809,9 @@ int Objecter::_calc_target(op_target_t *t, bool any_change)
     } else {
       int osd;
       bool read = is_read && !is_write;
-      if (t->use_osd_epoch) {
+      if (t->forced_osd != -1) {
 	assert(read);
-	osd = osdmap->is_up(t->osd) ? t->osd : -1;
+	osd = osdmap->is_up(t->forced_osd) ? t->forced_osd : -1;
```
jdurgin: hmm, even if the osd is up, it may no longer hold the pg for the repair read op - in which case it'll drop the op.

I suppose we could check if it's in acting here, and if not cancel the repair read request with ERANGE on the client side... a bit ugly to add that to the client though...

src/osd/PrimaryLogPG.cc
tchaikov: okay. will fix this.

jdurgin:
@tchaikov squashing the existing commits is fine with me

src/osdc/Objecter.cc
tchaikov: thanks and agreed! will fix.

doc/rados/troubleshooting/troubleshooting-pg.rst
jdurgin: consider an admin who detects a bad memory module, replaces it, and then scrubs to find that there are inconsistencies on the formerly bad node.

If there are random writes happening to the inconsistent objects, and no other cluster changes, the admin knows the copies on the formerly bad node still need repairing. Having to rescrub after each write seems pretty inconvenient, and may never allow repair this way if there are more random writes before the admin can use repair-copy with the new scrub output.

This sort of situation makes me wonder if repair-copy should avoid the version assert entirely - what benefit does it provide beyond the interval assert?

src/librados/IoCtxImpl.cc
jdurgin: keeping it as is sounds good, can clean up later if an opportunity appears

src/include/rados/librados.hpp
jdurgin: the overall return value (all that's inspected in repair_copy() or the non-operation versions of repair_read) will be 0 even if the interval assert on the repair_read fails, right?

src/osd/PrimaryLogPG.cc
jdurgin: peer_missing is used in many other places too (e.g. lost_unfound_revert, backfill, recovery) - I'm worried one of those would be affected by adding temporary entries to peer_missing. if the only reason it's used here is to exclude certain shards, how hard is it to pass that information down explicitly?

src/osd/PrimaryLogPG.cc
```diff
@@ -6006,13 +6006,10 @@ int PrimaryLogPG::do_osd_ops(OpContext *ctx, vector<OSDOp>& ops)
 	epoch_t pg_epoch = info.history.same_interval_since;
 	tracepoint(osd, do_osd_op_pre_assert_interval, soid.oid.name.c_str(), soid.snap.val, epoch);
 	dout(5) << "Check interval " << soid << " epoch " << epoch << " pg epoch " << pg_epoch << dendl;
-	if (epoch > pg_epoch) {
+	if (epoch != pg_epoch) {
```
jdurgin: this one is confusing to me - won't the epoch be different if there have been any osdmap changes that don't affect this pg? or is scrub reporting the same_interval_since epoch?

src/osd/PrimaryLogPG.cc
tchaikov: i did investigate the possible ways to pass this down. it's possible, but complicates the code path a lot. marking the corrupted shard missing fits our needs pretty well.

> I'm worried one of those would be affected by adding temporary entries to peer_missing.

i will investigate this. and come back to you.

src/include/rados/librados.hpp
tchaikov: right.

doc/rados/troubleshooting/troubleshooting-pg.rst
tchaikov: the version assert to prevent us from overwriting the changed object with the old copy.

jdurgin: what about having the repair-copy op pull the latest version, like copy-from does?

This assumes the caller (eventually ceph-mgr) is repairing the object's metadata (object info, snapset, clones) before using this call to repair its data/omap/xattrs - and in that case we can trust the local object info's version

tchaikov: @jdurgin i don't think ceph-mgr will be (eventually) smart enough to "repair" the consistencies. instead i think the expected caller of these API is user's own script. it (the script) will be able to inspect the content of object, to check the integrity of it, and to choose the correct one using user's domain knowledge about the content **before** using repair-copy to overwrite the corrupted ones with golden copy.

src/osd/PrimaryLogPG.cc
tchaikov: yes, scrub is reporting the same_interval_since epoch.

src/osdc/Objecter.cc
tchaikov: done.

src/test/osd/osd-scrub-repair.sh
tchaikov: if this test is moved to `qa/suites/rados/basic/tasks`, i think it will have the coverage of the various qa-suite options.

maybe we can postpone this to another PR? i'd really want to have this feature in Luminous if possible.

src/osd/PrimaryLogPG.cc
tchaikov: done.

src/osd/PrimaryLogPG.cc
```diff
@@ -7165,12 +7283,32 @@ void PrimaryLogPG::_copy_some(ObjectContextRef obc, CopyOpRef cop)
...
-				  flags,
-				  gather.new_sub(),
-				  // discover the object version if we don't know it yet
-				  cop->results.user_version ? NULL : &cop->results.user_version);
+  ceph_tid_t tid;
+  if (cop->flags & CEPH_OSD_COPY_FROM_FLAG_REPAIR) {
+    // we are holding the write lock of the object being repaired, so, to read from
+    // it, we need to SKIPRWLOCKS here. and since we are the very one holding the
+    // write lock, it's safe to skip the rwlock.
+    auto objecter_op = osd->objecter->prepare_read_op(cop->src.oid, cop->oloc, op,
```
tchaikov: fixed.

tchaikov:
@jdurgin all comments addressed and repushed.

src/librados/librados.cc
```diff
@@ -2244,7 +2259,9 @@ int librados::IoCtx::repair_copy(const string& oid, uint64_t version, uint32_t w
   for (const auto& bad_shard : bad_shards) {
     shards.emplace_back(bad_shard.osd, shard_id_t(bad_shard.shard));
   }
-  return io_ctx_impl->repair_copy(object_t(oid), version, what, shards, epoch);
+  return io_ctx_impl->repair_copy(object_t(oid), version,
+				  translate_repair_copy(what),
```
jdurgin: looks good, just need to add to the other versions of repair_copy

tchaikov:
see http://pad.ceph.com/p/scrub-repair, closing for now.

PR #9204, opened 2016-05-19
os/filestore: fix omap fsid related fd leak

xiexingguo:


xiexingguo:
@liewegas  Some fd leak fixes. 
Once you are available, thanks:-)

PR #8974, opened 2016-05-06
msg/msg_types: clean up entity_addr_t

liewegas:
- fix up accessors, users
- better operator<<
- remove sockaddr_storage from in-memory struct!!

PR #9205, opened 2016-05-19
Fix compiltaion entity addr t

Adirl:
xio: fix compilation after changes in entity_addr_t

    changes to entity_addr_t caused XIO to break
    see changes in https://github.com/ceph/ceph/pull/8974/files
    this commit is doing the needed changes so that xio can compile

### DIFF
--- a/src/msg/xio/XioMessenger.cc
+++ b/src/msg/xio/XioMessenger.cc
@@ -221,13 +221,13 @@ static string xio_uri_from_entity(const string &type,
   char addr_buf[129];
   string xio_uri;
 
-  switch(addr.addr.ss_family) {
+  switch(addr.get_family()) {
   case AF_INET:
-    host = inet_ntop(AF_INET, &addr.addr4.sin_addr, addr_buf,
+    host = inet_ntop(AF_INET, &addr.u.sin.sin_addr, addr_buf,
 		     INET_ADDRSTRLEN);
     break;
   case AF_INET6:
-    host = inet_ntop(AF_INET6, &addr.addr6.sin6_addr, addr_buf,
+    host = inet_ntop(AF_INET6, &addr.u.sin6.sin6_addr, addr_buf,
 		     INET6_ADDRSTRLEN);
     break;
   default:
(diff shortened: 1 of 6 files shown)


src/msg/xio/XioMessenger.cc
```diff
@@ -221,13 +221,13 @@ static string xio_uri_from_entity(const string &type,
   char addr_buf[129];
   string xio_uri;
 
-  switch(addr.addr.ss_family) {
+  switch(addr.get_family()) {
   case AF_INET:
-    host = inet_ntop(AF_INET, &addr.addr4.sin_addr, addr_buf,
+    host = inet_ntop(AF_INET, &addr.u.sin.sin_addr, addr_buf,
```
cbodley: can you use the helper functions instead of accessing the union directly? `addr.u.sin` -> `addr.in4_addr()`

src/msg/xio/XioMessenger.cc
```diff
@@ -436,7 +436,7 @@ void XioMessenger::learned_addr(const entity_addr_t &peer_addr_for_me)
   if (need_addr) {
     entity_addr_t t = peer_addr_for_me;
     t.set_port(my_inst.addr.get_port());
-    my_inst.addr.addr = t.addr;
+    my_inst.addr.u = t.u;
```
cbodley: `my_inst.addr.set_sockaddr(t.get_sockaddr());` should work here

cbodley:
@Adirl github pulled in some extra commits, can you rebase please?

Adirl:
@cbodley , Helper functions in4_addr() and in6_addr() are not const, therefore i could not use them. i can change them to const if needed. 
?

cbodley:
@Adirl ah, i see. instead of changing the existing ones to const, please add const overloads, i.e.:
```c++
sockaddr_in& in4_addr() { return u.sin; }
const sockaddr_in& in4_addr() const { return u.sin; }

sockaddr_in6& in6_addr() { return u.sin6; }
const sockaddr_in6& in6_addr() const { return u.sin6; }
```

mattbenjamin:
@cbodley I'm uncomfortable blocking this;  I just repeated the work done here, not realizing there was an open PR.  I think using these wrapper methods is maybe just adding verbosity, but, either way, we can always commit that change when it arrives.

PR #9206, opened 2016-05-19
Fix for vstart without -n option

Ved-vampir:
If you try to use vstart.sh without -n but with -r, you'll see, that vstart fails on the step of creating users, because they are already created (if vstart was used with -r option on previous start too). It's confusing behavior, I meet it during some rgw tests. I suppose to fix it.

### DIFF
--- a/src/vstart.sh
+++ b/src/vstart.sh
@@ -750,43 +750,6 @@ do_hitsets $hitset
 
 do_rgw()
 {
-    # Create S3 user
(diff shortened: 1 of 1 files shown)


cbodley:
@Ved-vampir there's an open PR https://github.com/ceph/ceph/pull/8241 that includes this fix and some others; i'd like to merge that once i get a chance to test it

PR #9207, opened 2016-05-19
rbd-mirror: Image resync

rjfd:
Implementation of image resync in the rbd mirror daemon.

Ticket url: http://tracker.ceph.com/issues/15670

### DIFF
--- a/src/journal/Journaler.h
+++ b/src/journal/Journaler.h
@@ -9,6 +9,7 @@
 #include "include/Context.h"
 #include "include/rados/librados.hpp"
 #include "journal/Future.h"
+#include "journal/JournalMetadataListener.h"
 #include "cls/journal/cls_journal_types.h"
 #include <list>
 #include <map>
@@ -70,6 +71,9 @@ public:
 			      int64_t *pool_id, Context *on_finish);
   void get_mutable_metadata(uint64_t *minimum_set, uint64_t *active_set,
 			    RegisteredClients *clients, Context *on_finish);
+
+  void add_listener(JournalMetadataListener *listener);
+  void remove_listener(JournalMetadataListener *listener);
 
   int register_client(const bufferlist &data);
   void register_client(const bufferlist &data, Context *on_finish);
--- a/src/librbd/Journal.cc
+++ b/src/librbd/Journal.cc
@@ -300,7 +300,8 @@ Journal<I>::Journal(I &image_ctx)
     m_lock("Journal<I>::m_lock"), m_state(STATE_UNINITIALIZED),
     m_error_result(0), m_replay_handler(this), m_close_pending(false),
     m_event_lock("Journal<I>::m_event_lock"), m_event_tid(0),
-    m_blocking_writes(false), m_journal_replay(NULL) {
+    m_blocking_writes(false), m_journal_replay(NULL),
+    m_metadata_listener(this) {
 
   CephContext *cct = m_image_ctx.cct;
   ldout(cct, 5) << this << ": ictx=" << &m_image_ctx << dendl;
@@ -1157,6 +1158,8 @@ void Journal<I>::destroy_journaler(int r) {
   delete m_journal_replay;
   m_journal_replay = NULL;
 
+  m_journaler->remove_listener(&m_metadata_listener);
+
   transition_state(STATE_CLOSING, r);
   m_journaler->shut_down(create_async_context_callback(
     m_image_ctx, create_context_callback<
@@ -1174,6 +1177,8 @@ void Journal<I>::recreate_journaler(int r) {
 
   delete m_journal_replay;
   m_journal_replay = NULL;
+
+  m_journaler->remove_listener(&m_metadata_listener);
 
   transition_state(STATE_RESTARTING_REPLAY, r);
   m_journaler->shut_down(create_async_context_callback(
@@ -1276,6 +1281,8 @@ void Journal<I>::handle_initialized(int r) {
       m_image_ctx, create_context_callback<
         Journal<I>, &Journal<I>::handle_get_tags>(this)));
   m_journaler->get_tags(m_tag_class, &tags_ctx->tags, tags_ctx);
+
+  m_journaler->add_listener(&m_metadata_listener);
 }
 
 template <typename I>
@@ -1650,6 +1657,99 @@ void Journal<I>::wait_for_steady_state(Context *on_state) {
   ldout(cct, 20) << this << " " << __func__ << ": on_state=" << on_state
                  << dendl;
   m_wait_for_state_contexts.push_back(on_state);
+}
+
+template <typename I>
+int Journal<I>::check_resync_requested(bool *do_resync) {
+  Mutex::Locker l(m_lock);
+  return check_resync_requested_internal(do_resync);
+}
+
+template <typename I>
+int Journal<I>::check_resync_requested_internal(bool *do_resync) {
+  CephContext *cct = m_image_ctx.cct;
+  ldout(cct, 20) << this << " " << __func__ << dendl;
+
+  assert(m_lock.is_locked());
+  assert(do_resync != nullptr);
+
+  cls::journal::Client client;
+  int r = m_journaler->get_cached_client(IMAGE_CLIENT_ID, &client);
+  if (r < 0) {
+     lderr(cct) << "failed to retrieve client: " << cpp_strerror(r) << dendl;
+     return r;
+  }
+
+  librbd::journal::ClientData client_data;
+  bufferlist::iterator bl_it = client.data.begin();
+  try {
+    ::decode(client_data, bl_it);
+  } catch (const buffer::error &err) {
+    lderr(cct) << "failed to decode client data: " << err << dendl;
+    return -EINVAL;
+  }
+
+  journal::ImageClientMeta *image_client_meta =
+    boost::get<journal::ImageClientMeta>(&client_data.client_meta);
+  if (image_client_meta == nullptr) {
+    lderr(cct) << "failed to access image client meta struct" << dendl;
+    return -EINVAL;
+  }
+
+  *do_resync = image_client_meta->resync_requested;
+
+  return 0;
+}
+
+template <typename I>
+void Journal<I>::handle_metadata_updated() {
+  CephContext *cct = m_image_ctx.cct;
+  ldout(cct, 20) << this << " " << __func__ << dendl;
+
+  std::list<journal::ResyncListener *> resync_private_list;
+
+  {
+    Mutex::Locker l(m_lock);
+
+    if (m_state == STATE_CLOSING || m_state == STATE_CLOSED ||
+        m_state == STATE_UNINITIALIZED || m_state == STATE_STOPPING) {
+      return;
+    }
+
+    bool do_resync = false;
+    int r = check_resync_requested_internal(&do_resync);
+    if (r < 0) {
+      lderr(cct) << "failed to check if a resync was requested" << dendl;
+      return;
+    }
+
+    if (do_resync) {
+      for (const auto& listener :
+                              m_listener_map[journal::ListenerType::RESYNC]) {
+        journal::ResyncListener *rsync_listener =
+                        boost::get<journal::ResyncListener *>(listener);
+        resync_private_list.push_back(rsync_listener);
+      }
+    }
+  }
+
+  for (const auto& listener : resync_private_list) {
+    listener->handle_resync();
+  }
+}
+
+template <typename I>
+void Journal<I>::add_listener(journal::ListenerType type,
+                              journal::JournalListenerPtr listener) {
+  Mutex::Locker l(m_lock);
+  m_listener_map[type].push_back(listener);
+}
+
+template <typename I>
+void Journal<I>::remove_listener(journal::ListenerType type,
+                                 journal::JournalListenerPtr listener) {
+  Mutex::Locker l(m_lock);
+  m_listener_map[type].remove(listener);
 }
 
 } // namespace librbd
--- a/src/tools/rbd_mirror/ImageReplayer.cc
+++ b/src/tools/rbd_mirror/ImageReplayer.cc
@@ -32,6 +32,7 @@
 using std::map;
 using std::string;
 using std::unique_ptr;
+using std::shared_ptr;
 using std::vector;
 
 namespace rbd {
@@ -220,6 +221,19 @@ private:
   Commands commands;
 };
 
+template <typename I>
+struct ResyncListener : public librbd::journal::ResyncListener {
+  ImageReplayer<I> *img_replayer;
+
+  ResyncListener(ImageReplayer<I> *img_replayer)
+    : img_replayer(img_replayer) {
+  }
+
+  virtual void handle_resync() {
+    img_replayer->resync_image();
+  }
+};
+
 } // anonymous namespace
 
 template <typename I>
@@ -234,7 +248,9 @@ void ImageReplayer<I>::BootstrapProgressContext::update_progress(
 }
 
 template <typename I>
-ImageReplayer<I>::ImageReplayer(Threads *threads, RadosRef local, RadosRef remote,
+ImageReplayer<I>::ImageReplayer(Threads *threads,
+                             shared_ptr<ImageDeleter> image_deleter,
+                             RadosRef local, RadosRef remote,
 			     const std::string &local_mirror_uuid,
 			     const std::string &remote_mirror_uuid,
 			     int64_t local_pool_id,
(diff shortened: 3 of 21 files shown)


rjfd:
@dillaman finished the implementation and rebased on top of master.

dillaman:
@rjfd thx -- will take a look today

dillaman:
@rjfd compile failed

src/librbd/Journal.cc
```diff
@@ -23,6 +23,17 @@
 
 namespace librbd {
 
+template<typename T>
```
dillaman: Minor: how about making this a member class and statically allocating it within ```librbd::Journal```?  That avoids the need to expose ```handle_metadata_updated``` as a public method.

src/journal/Journaler.h
```diff
@@ -44,6 +45,7 @@ class Journaler {
   typedef cls::journal::Tag Tag;
   typedef std::list<cls::journal::Tag> Tags;
   typedef std::set<cls::journal::Client> RegisteredClients;
+  typedef JournalMetadata::Listener JournalMetadataListener;
```
dillaman: Minor: maybe move this out to its own file now that it is being exposed outside of libjournal? Then you can include the listener header in ```librbd/Journal.h``` and define the concrete class in the header.

src/tools/rbd_mirror/ImageReplayer.cc
```diff
@@ -411,9 +457,29 @@ void ImageReplayer<I>::handle_bootstrap(int r) {
...
+  if (do_resync) {
+    dout(5) << "image resync was requested" << dendl;
+    resync_image();
+    return;
+  }
+
   {
     Mutex::Locker locker(m_lock);
 
+    m_local_image_ctx->journal->add_listener(
```
dillaman: Should add the listener before checking the resync status just in case there is a race

src/librbd/Journal.cc
```diff
@@ -1582,6 +1600,95 @@ void Journal<I>::wait_for_steady_state(Context *on_state) {
   m_wait_for_state_contexts.push_back(on_state);
 }
 
+template <typename I>
+int Journal<I>::check_resync_requested(bool *do_resync) {
+  CephContext *cct = m_image_ctx.cct;
+  ldout(cct, 20) << this << " " << __func__ << dendl;
+
+  if (!do_resync) {
```
dillaman: ```assert(do_resync != nullptr)```

src/librbd/Journal.cc
```diff
@@ -1582,6 +1600,95 @@ void Journal<I>::wait_for_steady_state(Context *on_state) {
...
+
+  return 0;
+}
+
+template <typename I>
+void Journal<I>::handle_metadata_updated() {
+  CephContext *cct = m_image_ctx.cct;
+  ldout(cct, 20) << this << " " << __func__ << dendl;
+
+  if (m_state == STATE_CLOSING || m_state == STATE_CLOSED ||
```
dillaman: Probably should be holding the lock when checking the state and checking the client for the resync flag

src/tools/rbd_mirror/ImageReplayer.cc
```diff
@@ -1310,6 +1410,64 @@ std::string ImageReplayer<I>::to_string(const State state) {
...
+                                           m_global_image_id);
+
+    if (on_finish) {
+      m_image_deleter->wait_for_scheduled_deletion(m_local_image_name,
+                                                   on_finish, false);
+    }
+    m_lock.Unlock();
+  } else {
+    m_stopping_for_resync = true;
+    if (m_state == STATE_STARTING) {
```
dillaman: Can we just reuse ```stop()``` even when in the "STARTING" state?

src/tools/rbd_mirror/ImageReplayer.cc
```diff
@@ -697,6 +766,9 @@ void ImageReplayer<I>::on_stop_local_image_close_finish(int r)
   m_replay_status_formatter = nullptr;
 
   m_remote_journaler->stop_replay();
+  if (m_stopping_for_resync) {
+    m_remote_journaler->unregister_client();
```
dillaman: Note: under PR #9309 I converted the shut down path to be fully asynchronous to fix an issue.  If we need to unregister the client in the shut down path, we need this to be async as well.  However, I am wondering what would happen if rbd-mirror crashed between unregistering with the remote journal and performing the delete.

src/tools/rbd_mirror/ImageReplayer.cc
```diff
@@ -1271,6 +1344,31 @@ void ImageReplayer<I>::handle_stop(int r, Context *on_start, Context *on_stop) {
...
 
+    if (m_stopping_for_resync) {
+      m_image_deleter->schedule_image_delete(m_local_pool_id,
+                                             m_local_image_id,
+                                             m_local_image_name,
+                                             m_global_image_id);
+      m_stopping_for_resync = false;
+
+      wait_for_image_delete = true;
+      FunctionContext *ctx = new FunctionContext(
```
dillaman: The ```FunctionContext``` can probably just be a re-invoking of ```handle_stop``` similar to the status update finalizer

src/tools/rbd_mirror/ImageReplayer.cc
```diff
@@ -1310,6 +1410,64 @@ std::string ImageReplayer<I>::to_string(const State state) {
 }
 
 template <typename I>
+void ImageReplayer<I>::resync_image(Context *on_finish) {
+  dout(20) << dendl;
+
+  m_lock.Lock();
+  if (m_state == STATE_STOPPED) {
+
+    dout(20) << "replayer is already stopped, skip to delete image step of "
```
dillaman: Can we combine the logic here with the logic used in ```handle_stop``` so that we only have one helper method that unregisters the client, schedules the delete, etc?

src/journal/Journaler.h
rjfd: I agree with your suggestion :-)

src/librbd/Journal.cc
rjfd: Yep, this will become cleaner, `although handle_metadata_updated` is already private, but now I can remove the friend class declaration.

src/tools/rbd_mirror/ImageReplayer.cc
rjfd: Now that you mention that possible race, I cannot check for the resync here, because the journaler will only be initialized later, and even if I add the listener before checking for resync it will not make any difference. I should add the listener as soon as the journaler is initialized, and then check for a resync,

src/tools/rbd_mirror/ImageReplayer.cc
rjfd: Not really, because in the `stop` method, if `m_state == STATE_STARTING` then the execution will not reach the `handle_stop` method.

dillaman: It should interrupt the start by setting ```m_stop_requested``` to true -- Mykola has a pending PR for cancelling the bootstrap / image sync during start up which will integrate into ```stop```

rjfd: hmmm... so after merging Mykola PR, do you think that if I just call `stop`, it will get everything done?

dillaman: Hopefully --- and until it gets merged, right now you'd only have to wait to kick off the resync of someone requested a resync mid-bootstrap (which should be a rare event).  I think executing ```on_start_fail_start``` right now if in the middle of a bootstrap might result in odd behavior / a crash if and when bootstrap completes and executes ```handle_bootstrap```.

src/tools/rbd_mirror/ImageReplayer.cc
rjfd: I that case, when the rbd-mirror recovers it will check in the bootstrap that it has to resync and will finish the work. 
And now that I'm thinking, I forgot to update the `request_resync` flag from the journal metadata, after a successful resync.

dillaman: Wouldn't the ImageDeleter delete the image and thus the associated request_resync flag?

rjfd: duh -- yes, it will remove the local journal as well

rjfd:
@dillaman pushed the new changes

dillaman:
Minor: reduce the commit title length on the journal metadata listener commit

src/librbd/Journal.cc
```diff
@@ -1611,6 +1618,95 @@ void Journal<I>::wait_for_steady_state(Context *on_state) {
...
+  return 0;
+}
+
+template <typename I>
+void Journal<I>::handle_metadata_updated() {
+  CephContext *cct = m_image_ctx.cct;
+  ldout(cct, 20) << this << " " << __func__ << dendl;
+
+  {
+    Mutex::Locker l(m_lock);
```
dillaman: Hold lock while invoking ```check_resync_requested```

src/tools/rbd_mirror/ImageReplayer.cc
```diff
@@ -456,6 +502,28 @@ void ImageReplayer<I>::handle_init_remote_journaler(int r) {
     return;
   }
 
+  {
```
dillaman: Should this be moved to ```handle_bootstrap```? That's the first time that the local image (and associate journal) is available to check.

src/tools/rbd_mirror/ImageReplayer.cc
```diff
@@ -456,6 +502,28 @@ void ImageReplayer<I>::handle_init_remote_journaler(int r) {
...
+    bool do_resync = false;
+    r = m_local_image_ctx->journal->check_resync_requested(&do_resync);
+    if (r < 0) {
+      derr << "failed to check if a resync was requested" << dendl;
+    }
+
+    if (do_resync) {
+      Context *on_finish = m_on_start_finish;
+      FunctionContext *ctx = new FunctionContext([this, on_finish](int r) {
+          resync_image(on_finish);
```
dillaman: If "r < 0", should you skip the resync attempt and just complete the ```on_finish``` context?

src/tools/rbd_mirror/ImageReplayer.cc
```diff
@@ -493,11 +561,13 @@ void ImageReplayer<I>::start_replay() {
 
   {
     Mutex::Locker locker(m_lock);
-    m_replay_handler = new ReplayHandler<I>(this);
-    m_remote_journaler->start_live_replay(m_replay_handler,
-                                          1 /* TODO: configurable */);
+    if (!m_stop_requested) {
```
dillaman: Is it possible to just move ```on_replay_interrupted``` up and return if true?

src/tools/rbd_mirror/ImageReplayer.cc
```diff
@@ -1249,6 +1345,57 @@ std::string ImageReplayer<I>::to_string(const State state) {
...
 template <typename I>
+void ImageReplayer<I>::resync_image(Context *on_finish) {
+  dout(20) << dendl;
+
+  bool is_stopped = false;
+  {
+    Mutex::Locker l(m_lock);
+
+    m_stopping_for_resync = true;
+    if (m_state == STATE_STOPPED) {
```
dillaman: Is this just to service the asok method?  If so, can we just hook the asok method to librbd's request resync method?  If replay has stopped, it will be restarted every 30 seconds so we should detect the request.

src/tools/rbd_mirror/ImageReplayer.cc
rjfd: I can't move it to `handle_bootstrap`, because the JournalMetadata will only start to listen for metadata updates after its initialization in method `init()`. Therefore, in order to not loose any resync request, I can only add the metadata listener after the initialization of the journaler.

src/tools/rbd_mirror/ImageReplayer.cc
rjfd: hmm... I think that a resync should be processed even if the image-replayer start failed. Maybe we are requesting a resync exactly because the image-replayer fails to start...

src/tools/rbd_mirror/ImageReplayer.cc
rjfd: Yes, this is just to service the asok method. To use the librbd's `mirror_image_resync` when the replayer is stopped I still need to open the local image. If you think is still worth to use librbd's request resync method it's also fine by me.

src/tools/rbd_mirror/ImageReplayer.cc
dillaman: But this is the init method for the remote journal, not the local journal.  The local journal is opened when the local image is opened and the exclusive lock is acquired (in ```OpenLocalImageRequest``` invoked from ```BootstrapRequest```).

src/tools/rbd_mirror/ImageReplayer.cc
dillaman: Fair enough

src/tools/rbd_mirror/ImageReplayer.cc
dillaman: I am just concerned about adding more corner cases to ImageReplayer when the user should probably just request the resync via the rbd CLI.

src/tools/rbd_mirror/ImageReplayer.cc
rjfd: @rjfd slaps his face! 
Yes, I can move the code to `handle_bootstrap` .

src/tools/rbd_mirror/ImageReplayer.cc
rjfd: I'm not finding a simple way of hooking the asok method to librbd's request resync. Only through a new method in ImageReplayer as I need to access either to the local image context, or to the local cluster connection and the local image id/name.
What if we completely remove the resync command from the asok?

rjfd:
@dillaman pushed the new changes

src/tools/rbd_mirror/ImageReplayer.cc
```diff
@@ -149,6 +150,25 @@ class FlushCommand : public ImageReplayerAdminSocketCommand {
 };
 
 template <typename I>
+class ResyncCommand : public ImageReplayerAdminSocketCommand {
```
dillaman: I thought you were going to yank the ASOK resync command since you can just use the rbd CLI?

src/tools/rbd_mirror/ImageReplayer.cc
```diff
@@ -907,7 +974,9 @@ void ImageReplayer<I>::handle_process_entry_safe(const ReplayEntry& replay_entry
     return;
   }
 
-  m_remote_journaler->committed(replay_entry);
+  if (m_remote_journaler) {
```
dillaman: Just wondering if you hit a case where the replayer gave you a callback after ```m_remote_journaler``` was destroyed?

src/tools/rbd_mirror/ImageReplayer.cc
rjfd: I just let it because you didn't confirm in the last comment. But I'm all in favor for removing it, and will do it.

src/tools/rbd_mirror/ImageReplayer.cc
rjfd: Yes, it happened. That is why I had to check if it still exists. Do you think it's an erroneous behavior?

rjfd:
@dillaman removed the asok resync command, updated the tests accordingly, and pushed the changes

dillaman:
@rjfd I think you need to update rbd_mirror.sh since a recent merge to master added pool arguments to all commands and it is now looking in pool ```test``` for image ```deleted```:

> + echo 2016-06-03 10:15:32 TEST: simple image resync
> 2016-06-03 10:15:32 TEST: simple image resync
> + request_resync_image cluster1 test
> + local cluster=cluster1
> + local image=test
> + rbd --cluster=cluster1 -p mirror mirror image resync test
> 2016-06-03 10:15:32.570701 7f8e469651c0 -1 WARNING: the following dangerous and experimental features are enabled: *
> 2016-06-03 10:15:32.570973 7f8e469651c0 -1 WARNING: the following dangerous and experimental features are enabled: *
> 2016-06-03 10:15:32.572373 7f8e469651c0 -1 WARNING: the following dangerous and experimental features are enabled: *
> Flagged image for resync from primary
> + wait_for_image_present cluster1 test deleted
> + local cluster=cluster1
> + local pool=test
> + local image=deleted
> + local state=

rjfd:
@dillaman rebased on top of master and updated the rbd_mirror.sh test

dillaman:
lgtm

dillaman:
@rjfd needs rebase

rjfd:
@dillaman done

rjfd:
@dillaman rebased again to resolve conflicts

dillaman:
@rjfd Seeing package build failures:

```
In file included from test/librbd/fsx.cc:53:0:
./journal/Journaler.h:12:45: fatal error: journal/JournalMetadataListener.h: No such file or directory
```

You will need to add ```journal/JournalMetadataListener.h``` to ```journal/Makefile.am```'s noinst_HEADERS list.

rjfd:
@dillaman done -- thanks

PR #9211, opened 2016-05-19
librbd: write-after-write might result in an inconsistent replicated image 

dillaman:
With the librbd cache enabled, there is a possible race condition with write-after-write IO to the same image extent if followed by a client crash.  In such a case, it might result in an inconsistent primary / non-primary image.

### DIFF
--- a/qa/workunits/rbd/rbd_mirror_stress.sh
+++ b/qa/workunits/rbd/rbd_mirror_stress.sh
@@ -325,7 +325,8 @@ write_image()
     local duration=$(($RANDOM % 35 + 15))
 
     timeout ${duration}s rbd --cluster ${cluster} -p ${POOL} bench-write \
-	${image} --io-size 4096 --io-threads 8 --io-total 10G --io-pattern rand || true
+	${image} --io-size 4096 --io-threads 8 --io-total 10G --io-pattern rand \
+        --debug-rbd=20 --debug-journaler=20 2> ${TEMPDIR}/rbd-bench-write.log || true
 }
 
 create_snap()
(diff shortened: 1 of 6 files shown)


qa/workunits/rbd/rbd_mirror_stress.sh
```diff
@@ -325,7 +325,8 @@ write_image()
     local duration=$(($RANDOM % 35 + 15))
 
     timeout ${duration}s rbd --cluster ${cluster} -p ${POOL} bench-write \
-	${image} --io-size 4096 --io-threads 8 --io-total 10G --io-pattern rand || true
+	${image} --io-size 4096 --io-threads 8 --io-total 10G --io-pattern rand \
+        --debug-rbd=20 --debug-journaler=20 2> ${TEMPDIR}/rbd-bench-write.log || true
 }
```
trociny: Is only the latest write usually interesting so overwriting an older log is ok?

I would make it at least `rbd-bench-write.${cluster}.${image}.log` or `rbd-bench-write.${cluster}.${image}.$(date +%s).log` if it is a good idea to keep all logs (or use `2>>`).

trociny:
@dillaman Add `Fixes: http://tracker.ceph.com/issues/15938` to a commit log?

src/librbd/LibrbdWriteback.cc
```diff
@@ -277,8 +298,17 @@ namespace librbd {
 			    len, file_extents);
     for (Extents::iterator it = file_extents.begin();
 	 it != file_extents.end(); ++it) {
-      m_ictx->journal->commit_io_event_extent(original_journal_tid, it->first,
-					      it->second, 0);
+      if (new_journal_tid != 0) {
+        // ensure new journal event is safely committed to dist before
```
trociny: s/to dist/to disk/

trociny:
otherwise looks goot to me

qa/workunits/rbd/rbd_mirror_stress.sh
dillaman: Only the latest one is important since we snapshot and compare the images after each iteration.  I originally had it recording the log with an iteration string, but it didn't provide any value and it just took up space in my /tmp/

dillaman:
@trociny thx -- pushed

PR #9213, opened 2016-05-19
osd/ReplicatedPG.h: remove incorrect use of move

stiopaa1:
The modified functions  take universal references, so they can bind to
either rvalues and lvalues. When they bind to lvalues the use of
std::move is incorrect.
We should use std::forward, so we only move from rvalues.

### DIFF
--- a/src/osd/ReplicatedPG.h
+++ b/src/osd/ReplicatedPG.h
@@ -563,19 +563,19 @@ public:
     list<std::function<void()>> on_success;
     template <typename F>
     void register_on_finish(F &&f) {
-      on_finish.emplace_back(std::move(f));
+      on_finish.emplace_back(std::forward<F>(f));
     }
     template <typename F>
     void register_on_success(F &&f) {
-      on_success.emplace_back(std::move(f));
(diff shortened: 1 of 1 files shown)


liewegas:
@adamemerson can you review?

tchaikov:
i think is more of a cleanup. the only place we call it with a lvalue, is at https://github.com/ceph/ceph/blob/master/src/osd/ReplicatedPG.cc#L8218. in that case, we cast it into lvalue reference, and it's fine, as we reset it with `boost::none` afterwards.

adamemerson:
This is precisely the correct things.

Licentious Goldfish Tempt Minnows

athanatos:
Yeah, that's right.

tchaikov:
tested at http://pulpito.ceph.com/kchai-2016-06-12_09:44:10-rados-wip-kefu-testing---basic-mira/

http://pulpito.ceph.com/kchai-2016-06-12_09:44:10-rados-wip-kefu-testing---basic-mira/254583/ addressed by https://github.com/ceph/ceph-qa-suite/pull/1052

PR #9214, opened 2016-05-19
mds: log on bad seq

bniver:
Cleanup to make the specific case of seqn received greater than seqn expected match the behaviour of the other cases of seqn mis-match.  Not asserting on bad client seqn.

Fixes: http://tracker.ceph.com/issues/9904
Signed-off-by: Brett Niver <bniver@redhat.com>

### DIFF
--- a/src/mds/Server.cc
+++ b/src/mds/Server.cc
@@ -349,10 +349,16 @@ void Server::handle_client_session(MClientSession *m)
 	m->put();
 	return;
       }
+      // We are getting a seq that is higher than expected.
+      // Handle the same as any other seqn error.
+      //
       if (m->get_seq() != session->get_push_seq()) {
-	dout(0) << "old push seq " << m->get_seq() << " != " << session->get_push_seq() 
+	dout(0) << "old push seq " << m->get_seq() << " != " << session->get_push_seq()
 		<< ", BUGGY!" << dendl;
-	assert(0);
+	mds->clog->warn() << "incorrect push seq " << m->get_seq() << " != "
+			  << session->get_push_seq() << ", dropping" << " from client : " << session->get_human_name();
+	m->put();
+	return;
       }
       journal_close_session(session, Session::STATE_CLOSING, NULL);
     }


gregsfortytwo:
@bniver, prepare for your style guide onslaught. :) https://github.com/ceph/ceph/blob/master/SubmittingPatches.rst and https://github.com/ceph/ceph/blob/master/CodingStyle

How about
>Do not assert on bad client seq

instead of

>working 9904

as the title; we like commit headers to stand on their own. You'd then include a line "Fixes: http://tracker.ceph.com/issues/9904" after the longer explanation in the commit body.

Don't forget your Signed-off-by as the last entry in the commit message.

src/mds/Server.cc
```diff
@@ -349,10 +349,16 @@ void Server::handle_client_session(MClientSession *m)
 	m->put();
 	return;
       }
+      //	This looks as if we are getting a seq that is higher than
+      //	expected.  I didn't notice any "wrap" handling, but the seqn is
+      //	64-bits ...
```
gregsfortytwo: We generally don't indent comments, partly because...

src/mds/Server.cc
```diff
@@ -349,10 +349,16 @@ void Server::handle_client_session(MClientSession *m)
...
+      //
       if (m->get_seq() != session->get_push_seq()) {
 	dout(0) << "old push seq " << m->get_seq() << " != " << session->get_push_seq() 
-		<< ", BUGGY!" << dendl;
-	assert(0);
+ 		<< ", BUGGY!" << dendl;
+	mds->clog->warn() << "incorrect push seq " << m->get_seq() << " != " << session->get_push_seq() << ", dropping";
```
gregsfortytwo: The MDS code is old enough that it doesn't conform to this well, but we really prefer 80-character lines in the source when making changes. It's not a totally strict rule but especially when we've got easy breakpoints like the << operator, we try to stick to it.

gregsfortytwo:
Switching from an assert to a clog warning: definitely a good choice. 👍

src/mds/Server.cc
bniver: Sure - I'll have to correct my emacs :-).  Is the formatting, etc the check that is failing?

bniver:
I'm more than happy to learn the hard way since that's how I learn best.
1) looks like sage already changed that for me, thanks!
2) signature, can I add it here, or do I re-submit?
3) wasn't sure about the "assigned" I thought I was assigning a reviewer, not?

I.e what do I need to do next, I've removed the tabs in my comments, I've shortened my lines.  Do I just do a completely new PR?
Thanks,
Brett

liewegas:
Once you update the commits in your local repository (git rebase, git gui, etc.), force push teh same branch and the PR will update (git push remotename -f wip-bniver-9904)

src/mds/Server.cc
```diff
@@ -349,10 +349,17 @@ void Server::handle_client_session(MClientSession *m)
...
+      // expected.  I didn't notice any "wrap" handling, but the seqn is
+      // 64-bits ...
+      //
       if (m->get_seq() != session->get_push_seq()) {
-	dout(0) << "old push seq " << m->get_seq() << " != " << session->get_push_seq() 
+	dout(0) << "old push seq " << m->get_seq() << " != " << session->get_push_seq()
 		<< ", BUGGY!" << dendl;
-	assert(0);
+	mds->clog->warn() << "incorrect push seq " << m->get_seq() << " != "
+			  << session->get_push_seq() << ", dropping0";
```
jcsp: s/dropping0/dropping/

Also, would be nice to include in the log message: session->info.inst.name (the client's unique ID) and session->get_human_name() (a hopefully more recognisable string derived from the metadata), so that the admin has a chance to know which client is buggy.

src/mds/Server.cc
```diff
@@ -349,10 +349,17 @@ void Server::handle_client_session(MClientSession *m)
 	m->put();
 	return;
       }
+      // This looks as if we are getting a seq that is higher than
+      // expected.  I didn't notice any "wrap" handling, but the seqn is
+      // 64-bits ...
```
jcsp: I prefer to avoid uncertain "..." comments like this: this kind of creates more doubt than it removes for the developer reading this years down the line (also the "I" could be anyone when you're reading it later).  I like nice factual statements like "The client message is reporting a higher seq than the server's session state expects, can only happen on bug".

jcsp:
@bniver I see you've updated the PR description with the correct commit message format, you also need to update the commit itself (I usually use "git rebase -i" and then select "r" for the comment when I want to rewrite the message) -- you could squash your two commit into one at the same time.

src/mds/Server.cc
bniver: adding more log info, editing code comments, will update commit and "squash" them if I fiigure out how

src/mds/Server.cc
```diff
@@ -349,10 +349,17 @@ void Server::handle_client_session(MClientSession *m)
...
+      // Handle the same as any other seqn error.
+      //
       if (m->get_seq() != session->get_push_seq()) {
-	dout(0) << "old push seq " << m->get_seq() << " != " << session->get_push_seq() 
+	dout(0) << "old push seq " << m->get_seq() << " != " << session->get_push_seq()
 		<< ", BUGGY!" << dendl;
-	assert(0);
+	mds->clog->warn() << "incorrect push seq " << m->get_seq() << " != "
+			  << session->get_push_seq() << ", dropping0";
+	mds->clog->warn() << "client unique ID : " << session->info.inst.name << " , client name : " << session->get_human_name() ;
```
jcsp: This should all be in one clog->warn(), otherwise some other message could pop up between and make it confusing.  You can probably make it a bit less verbose, like just add "from client " << human name << " (" << inst.name << ")" to the first log line.

Also the "dropping0" typo is still in

jcsp:
First line of commit message should be prefixed "mds: "

gregsfortytwo:
And the commit message lines definitely need to be <80 characters, or git log becomes impossible to parse!

src/mds/Server.cc
bniver: duh, thanks for seeing the typo.  I'll combine the logs.

jcsp:
@bniver so close!  You added the mds: but took away the other bits.

Merge the changes into a single commit, and the commit message should read exactly as follows:
```
mds: log on bad seq

Fixes: http://tracker.ceph.com/issues/9904
Signed-off-by: Brett Niver <bniver@redhat.com>
```

jcsp:
Reviewed-by, will add to next test branch

jcsp:
(this was in http://pulpito.ceph.com/jspray-2016-05-28_13:42:42-fs-wip-jcsp-testing-20160527b---basic-mira/ which had some failures but enough passes to make me comfortable)

PR #9215, opened 2016-05-19
os/bluestore: enable RocksDB row cache

jjhuo:
This PR will enable both RocksDB row cache for bluestore.
Row cache will cache kv pairs after db get(), so next time when db get same kv pair, it won't need to go through sst files to do lookup again.

### DIFF
--- a/src/kv/LevelDBStore.cc
+++ b/src/kv/LevelDBStore.cc
@@ -265,18 +265,18 @@ int LevelDBStore::get(
 
 int LevelDBStore::get(const string &prefix, 
 		  const string &key,
-		  bufferlist *value)
+		  bufferlist *out)
 {
-  assert(value && (value->length() == 0));
+  assert(out && (out->length() == 0));
   utime_t start = ceph_clock_now(g_ceph_context);
   int r = 0;
-  leveldb::ReadOptions options;
-  std::string value;
-  std::string bound = combine_strings(prefix, key);
-  auto status = db->Get(options, bound, &value);
-  if (status.ok()) {
-    value->append(value);
-  } else if (!status.IsNotFound()) {
+  string value, k;
+  leveldb::Status s;
+  k = combine_strings(prefix, key);
+  s = db->Get(leveldb::ReadOptions(), leveldb::Slice(k), &value);
+  if (s.ok()) {
+    out->append(value);
+  } else {
     r = -ENOENT;
   }
   utime_t lat = ceph_clock_now(g_ceph_context) - start;
--- a/src/kv/LevelDBStore.cc
+++ b/src/kv/LevelDBStore.cc
@@ -270,11 +270,13 @@ int LevelDBStore::get(const string &prefix,
   assert(value && (value->length() == 0));
   utime_t start = ceph_clock_now(g_ceph_context);
   int r = 0;
-  KeyValueDB::Iterator it = get_iterator(prefix);
-  it->lower_bound(key);
-  if (it->valid() && it->key() == key) {
-    value->append(it->value_as_ptr());
-  } else {
+  leveldb::ReadOptions options;
+  std::string value;
+  std::string bound = combine_strings(prefix, key);
+  auto status = db->Get(options, bound, &value);
+  if (status.ok()) {
+    value->append(value);
+  } else if (!status.IsNotFound()) {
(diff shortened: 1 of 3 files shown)


jjhuo:
I will do fio_rbd 4KB 100% random read and write benchmarking, to verify if performance of rocksdb block cache is comparable to kernel page cache.

yuyuyu101:
I think this PR only increase rocksdb block cache shard?

jjhuo:
@yuyuyu101 All IO in bluefs are converted to direct IO, then RocksDB won't be able to use kernel page cache. For two read() functions inside NVMEDevice.cc, both of them don't take care of "buffered" flag, hope it's fine.

liewegas:
The big question for me is: if we enable the rocksdb block cache etc and tune it accordingly, can we get rid of the buffered mode in the BlockDevice interface?  This would simplify things greatly.  Note that on the BlueStore side we are doing the cache ourselves and don't need buffered mode there at all.

src/os/bluestore/BlockDevice.h
```diff
@@ -89,7 +89,8 @@ class BlockDevice {
 
   virtual int read(uint64_t off, uint64_t len, bufferlist *pbl,
 	   IOContext *ioc, bool buffered) = 0;
-  virtual int read_buffered(uint64_t off, uint64_t len, char *buf) = 0;
+  virtual int read(uint64_t off, uint64_t len, char *buf,
+           bool buffered) = 0;
```
liewegas: note that in the current implementation these are slightly different... the read() variant assumes reads are block-aligned, whereas read_buffered can do byte-granuarltiy length and offset.  Not sure if this is a problem or not...

liewegas:
This looks good to me!

src/kv/LevelDBStore.cc
```diff
@@ -270,11 +270,13 @@ int LevelDBStore::get(const string &prefix,
   assert(value && (value->length() == 0));
   utime_t start = ceph_clock_now(g_ceph_context);
   int r = 0;
-  KeyValueDB::Iterator it = get_iterator(prefix);
-  it->lower_bound(key);
-  if (it->valid() && it->key() == key) {
-    value->append(it->value_as_ptr());
-  } else {
+  leveldb::ReadOptions options;
+  std::string value;
```
markhpc: This shadows the passed value parameter (fixed in the below RocksDB code).  Fix here:

https://github.com/markhpc/ceph/commit/5bdce852ab81b7910475e97a99fd1698789818af

markhpc: Actually this code was breaking vs the RocksDBStore version.  I rewrote it based on that version and it works fine:

https://github.com/markhpc/ceph/blob/wip-9215/src/kv/LevelDBStore.cc#L273-L281

src/os/bluestore/BlockDevice.h
jjhuo: @liewegas thanks for your review. for read_buffered(), it's only used in BlueFS::_read_random(), which is supposed to do byte-granularity length and offset. So I guess it should be fine. Do I need to rename this new "read" function to be "read_random" to avoid confusion?

src/kv/LevelDBStore.cc
jjhuo: Great. I don't know how to grab this commit? could you please send me a link? thanks!

tchaikov: @jjhuo 

```
git remote add markhpc https://github.com/markhpc/ceph.git
git fetch markhpc
git checkout --track markhpc/wip-9215
```

jjhuo: Thank you, Kefu!

jjhuo:
@tchaikov Thanks for your help! @markhpc I squashed your commits into one, if you don't mind. :-)

markhpc:
Not at all! :)

jjhuo:
@markhpc the crash you saw is caused by unaligned read, Sage was right. RocksDB is not ready to ditch page cache yet, it's still doing unaligned read. @liewegas I removed two commits, this PR only enables row cache now. I will test performance impact of it with bluefs. Thanks!

liewegas:
see https://github.com/ceph/ceph/pull/9533

PR #9216, opened 2016-05-19
test: Disable tests due to recovery race

dzafman:


dzafman:
@liewegas This passed the Jenkins test.  I repushed it with an amended comment only.  Ready to merge.

PR #9225, opened 2016-05-20
tests: rm -fr /tmp/*virtualenv*

dachary:

### DIFF
--- a/run-make-check.sh
+++ b/run-make-check.sh
@@ -76,8 +76,10 @@ function run() {
 function main() {
     if run "$@" ; then
         echo "make check: successful run on $(git rev-parse HEAD)"
+        rm -fr ${CEPH_BUILD_VIRTUALENV:-/tmp}/*virtualenv*
         return 0
     else
+        rm -fr ${CEPH_BUILD_VIRTUALENV:-/tmp}/*virtualenv*
         return 1
     fi
 }


alfredodeza:
we export the temporary location of virtualenvs  (`CEPH_BUILD_VIRTUALENV`) as an environment variable when building Ceph (we don't currently for running make_check), it would be good if we could check for that before removing. This just assumes it will always be `/tmp/`.

Why can't the makefile of the tools involved here remove or ensure this location is clean? For example, in ceph-disk it would be:

https://github.com/ceph/ceph/blob/master/src/ceph-detect-init/Makefile.am#L61

ErwanAliasr1:
+1 for me

tchaikov:
@dachary does this fix the "pkg_resources.DistributionNotFound: ceph-disk==1.0.0" issue?

dachary:
Repushed. The proper solution is to provision a new machine on each run, with a pristine installation of the operating system guaranteed to not have any leftovers from the previous run. Next time around we'll discover that the dependencies installed by run-make-check.sh have to be removed or something similar.

dachary:
@tchaikov I think it can but I don't have access to the jenkins slave so I can't tell for sure.

tchaikov:
@alfredodeza let's try @dachary 's theory first before working on a better "make check". and i am not sure what target we should be using? "make test-clean"?

dachary:
there also is a need for apt-get remove all packages and dependencies so they don't conflict / interfere with another run-make-check that has different requirements.

PR #9228, opened 2016-05-20
os/bluestore: new write path (checksums and compression)

liewegas:
This branch completely replaces the read and write path with a new design.
The disk format has changed, and the data structures are very different.

Compression is only partially supported, but checksums work.

ceph_test_objectstore passes all tests.


src/common/config_opts.h
```diff
@@ -949,7 +949,17 @@ OPTION(bluestore_block_wal_path, OPT_STR, "")
...
+OPTION(bluestore_csum, OPT_BOOL, true)
+OPTION(bluestore_csum_type, OPT_STR, "crc32c")
+OPTION(bluestore_min_csum_block, OPT_U32, 4096)
+OPTION(bluestore_max_csum_block, OPT_U32, 64*1024)
+OPTION(bluestore_min_alloc_size, OPT_U32, 0)
+OPTION(bluestore_min_alloc_size_hdd, OPT_U32, 64*1024)
+OPTION(bluestore_min_alloc_size_ssd, OPT_U32, 4*1024)
+OPTION(bluestore_compression, OPT_STR, "none")  // force|aggressive|passive|none
+OPTION(bluestore_compression_algorithm, OPT_STR, "snappy")
+OPTION(bluestore_compression_min_blob_size, OPT_U32, 256*1024)
```
ifed01: Perhaps we should consider blob size specification in allocation units rather than absolute units. This way we'll always have blobs aligned with alloc units
E.g. 
bluestore_compression_min_blob_size = 8 // i.e = 8 * bluestore_mon_alloc_size = 8 * 64 * 1024 bytes

src/os/bluestore/BlueStore.cc
```diff
@@ -3132,7 +3132,9 @@ int BlueStore::_decompress(bufferlist& source, bufferlist* result)
     derr << __func__ << " can't load decompressor " << chdr.type << dendl;
     r = -EIO;
   } else {
-    r = compressor->decompress(i, *result);
+    bufferlist t;
```
ifed01: Shouldn't we provide decompress with two iterators ( begin,end) instead to avoid data copying?

src/common/config_opts.h
liewegas: We could do that, but we can also enforce a check that errors out if it is 
misconfigured, to.  My only worry is that it is confusing to configure if 
it is in multiples.

Hmm, it probably is better, though, and communicates where the alignment 
requirements/restrictions are...

src/os/bluestore/BlueStore.cc
liewegas: that would be better, yeah!  i was just lazy and didn't want to fuss with 
the Compressor interface

ifed01: Just made a PR to fix that

markhpc:
Am marking this performance as I want to track this both for the compression support performance characteristics and the other potential impacts on the write path.

src/os/bluestore/BlueStore.cc
```diff
@@ -802,6 +801,58 @@ void BlueStore::handle_conf_change(const struct md_config_t *conf,
...
+{
+  comp_min_blob_size = g_conf->bluestore_compression_min_blob_size;
+  comp_max_blob_size = g_conf->bluestore_compression_max_blob_size;
+
+  const char *alg = 0;
+  if (g_conf->bluestore_compression_algorithm == "snappy") {
+    alg = "snappy";
+  } else if (g_conf->bluestore_compression_algorithm == "zlib") {
+    alg = "zlib";
+  } else if (g_conf->bluestore_compression_algorithm.length()) {
```
ifed01: We can burden Compressor with the check for known compression algorithm instead of doing that here. This way adding new algorithm becomes transparent for the BlueStore

src/os/bluestore/BlueStore.cc
```diff
@@ -5599,21 +5599,30 @@ void BlueStore::_do_write_big(
...
-	   << o->onode.extent_map[offset] << dendl;
-  dout(20) << __func__ << "  blob " << *b << dendl;
+  while (length > 0) {
+    int64_t blob;
+    bluestore_blob_t *b = o->onode.add_blob(&blob);
+    wctx->blob_new.push_back(b);
+    b->length = MIN(max_blob_len, length);
+    bufferlist t;
+    blp.copy(b->length, t);
+    wctx->bl_new.push_back(t);
```
ifed01: One can probably do compress here and thus eliminate the additional copy and reduce memory footprint a bit.
Another option is to store <start_iterator, end_iterator> pair instead of  bufferlist copy in bl_new.

src/os/bluestore/BlueStore.cc
```diff
@@ -5363,706 +5225,466 @@ void BlueStore::_pad_zeros_tail(
...
-	} else if (bp->second.is_shared()) {
-	  dout(20) << "  tail shared" << dendl;
-	  length += min_alloc_size;
-	  shared_tail = true;
-	} else {
-	  dout(20) << "  tail 0x" << std::hex << tail << std::dec
-		   << " will presumably WAL" << dendl;
-	}
-      }
+    if (ep->first < ep->second.offset) {
```
ifed01: Probably I'm missing something but thus comparison looks strange - absolute offset vs. relative one.
Shouldn't it be something like
    if ((ep->first % min_alloc_unit) < ep->second.offset)

src/os/bluestore/BlueStore.cc
liewegas: Yeah, I would lean toward a pair/struct and std::move or similar.  I like 
that all of the allocation happens in one place (at least for large 
writes) because we can probably make it a single sequential block (tho we 
don't do that now).

src/os/bluestore/BlueStore.cc
```diff
@@ -5363,706 +5225,466 @@ void BlueStore::_pad_zeros_tail(
...
-	cow_offset += c_len;
-	cow_length -= c_len;
-	x_off += c_len;
-	assert(cow_length == 0);  // fixme ... until we fix this loop
+      if (tail_read) {
+	bufferlist tail_bl;
+	int r = _do_read(c.get(), o, offset + length + tail_pad, tail_read,
+			 tail_bl, 0);
+	assert(r >= 0 && r <= (int)tail_read);
+	b_len += r;
```
ifed01: Shouldn't that be 
b_len += tail_read; 
to have b_len chunk_aligned and synced with padded.length()?

src/os/bluestore/BlueStore.cc
```diff
@@ -5363,706 +5225,466 @@ void BlueStore::_pad_zeros_tail(
...
-  if (length) {
-    dout(20) << "  must alloc 0x" << std::hex << offset << "~0x" << length
-	     << std::dec << dendl;
+  // new blob.
+  int64_t blob;
+  b = o->onode.add_blob(&blob);
+  b->length = min_alloc_size;
+  uint64_t b_off = offset % min_alloc_size;
+  uint64_t b_len = length;
+  _pad_zeros(txc, o, &bl, &b_off, &b_len, block_size);
```
ifed01: _pad_zeros doesn't need txc & o any more

src/os/bluestore/BlueStore.cc
```diff
@@ -5363,706 +5225,466 @@ void BlueStore::_pad_zeros_tail(
...
+  if (csum_type) {
+    // it's little; csum at block granularity.
+    b->init_csum(csum_type, block_size_order, min_alloc_size);
+    checksummer->calculate(b->csum_type, b->get_csum_block_size(),
+			   b_off, b_len, bl, &b->csum_data);
+  }
 
-    // positional hint
-    uint64_t hint = 0;
+  // allocate and write
```
ifed01: IMHO alloc/write code below duplicates one from do_alloc_write - why not reuse it here?

src/os/bluestore/BlueStore.cc
```diff
@@ -5363,706 +5225,466 @@ void BlueStore::_pad_zeros_tail(
...
+      dout(20) << __func__ << " keep blob " << *b << dendl;
+    }
+    if (l.blob >= 0) {
+      txc->write_onode(o);
+    } else {
+      txc->write_bnode(o->bnode);
+    }
+  }
+
+  o->onode.compress_extent_map();
```
ifed01: Suggest to call txc->write_onode(o) if result > 0 here for the sake of code view consistency

PR #9231, opened 2016-05-20
osd: Put background recovery tasks in the main OpWQ

athanatos:


athanatos:
Tested: sjust@teuthology:/a/samuelj-2016-05-19_16:17:25-rados-wip-sam-working-distro-basic-smithi

liewegas:
That osd_recovery_sleep still sucks since it's blocking up the op wq.  Oh well.

liewegas:
LGTM!

PR #9234, opened 2016-05-20
osd/OSDCap: prevent copying vector/strings

stiopaa1:


cbodley:
it'd rather change the ctors to take the arguments by const ref instead of by value.. then we only do a single copy instead of a copy + move

stiopaa1:
@cbodley 
>it'd rather change the ctors to take the arguments by const ref instead of by value.. then we only do a single copy instead of a copy + move

That's true for lvalues, but for rvalues we will have:
one copy (for passing const ref)
move + move (for passing by value)

cbodley:
i'm having trouble understanding enough of the boost::spirit stuff to tell whether it's passing lvalues or rvalues, but fair enough :+1:

yuriw:
tested http://pulpito.ceph.com/yuriw-2016-05-25_20:42:59-rados-wip-yuri-testing-distro-basic-smithi/

PR #9236, opened 2016-05-20
osd/PG: update PGPool to detect map gaps and reset cached_removed_snaps

athanatos:
Fixes: http://tracker.ceph.com/issues/15943

### DIFF
--- a/src/osd/OSD.cc
+++ b/src/osd/OSD.cc
@@ -2986,14 +2986,8 @@ PGPool OSD::_get_pool(int id, OSDMapRef createmap)
     assert(0);
   }
 
-  PGPool p = PGPool(id, createmap->get_pool_name(id),
-		    createmap->get_pg_pool(id)->auid);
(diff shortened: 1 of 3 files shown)


src/osd/OSD.cc
```diff
@@ -3000,14 +3000,9 @@ PGPool OSD::_get_pool(int id, OSDMapRef createmap)
     assert(0);
   }
 
-  PGPool p = PGPool(id, createmap->get_pool_name(id),
+  PGPool p = PGPool(createmap, id, createmap->get_pool_name(id),
```
liewegas: the third argument (name) is redundant now

liewegas: and the 4th

athanatos:
Repushed, DNM pending another test run.

athanatos:
Tested again: sjust@teuthology:/a/samuelj-2016-06-01_11:28:31-rados-wip-sam-testing-distro-basic-smithi/228382/remote

PR #9242, opened 2016-05-20
jewel: rgw: don't unregister request if request is not connected to manager

yehudasa:
http://tracker.ceph.com/issues/15978


dachary:
It passed the rgw suite http://tracker.ceph.com/issues/16344#note-6 (except for one failure which I believe to be unrelated).

PR #9243, opened 2016-05-20
rgw: fix manager selection when APIs customized

robbat2:
When modifying rgw_enable_apis per RGW instance, such as for staticsites, you
can end up with RESTManager instance being null in some cases, which returns a
HTTP 405 MethodNotAllowed to all requests.

Example configuration to trigger the bug:
rgw_enable_apis = s3website

Backport: jewel
X-Note: Patch from Yehuda in private IRC discussion, 2016/05/20.
Fixes: http://tracker.ceph.com/issues/15973
Fixes: http://tracker.ceph.com/issues/15974


robbat2:
Going to regenerate this against master

PR #9248, opened 2016-05-21
mon: expose require_jewel_osds flag to user

xiexingguo:


xiexingguo:
@liewegas Could you help to review this? Thanks.

liewegas:
Hmm, I'm actually not quite sure why this explicit OSDMap flag is there.  @athanatos ?

liewegas:
Hmm, fo rhammer, there was the bitwise sort flag introduced at the same time.  And an individual osd couldn't downgrade.  We didn't have (or quite need) a cluster-wide flag to prevent downgrade.

athanatos:
I think there were features which once used prevent pre-jewel osds from working.  I think users cannot be allowed to cancel that flag.

xiexingguo:
change set:
- add "require_jewel_osds" to white list, so user can access it by the "ceph osd set *" command family(which is previously invisible to user)
- update relevant test cases

@liewegas @athanatos  Please check again, thanks.

xiexingguo:
retest this please

athanatos:
Doesn't it get automatically set once a jewel required feature is used?  I'm not clear on what the use case is.

xiexingguo:
> Doesn't it get automatically set once a jewel required feature is used?

For newly created cluster it will be. 
But it seems we also try to allow user to set this flag manually at https://github.com/ceph/ceph/blob/master/src/mon/OSDMonitor.cc#L6265, perhaps for upgrade from a pre-jewel cluster?  The problem is that we don't add this flag into white list,  so it is actually not invisible to user.

athanatos:
@xiexingguo I get it now.  Yeah, this version makes sense.

PR #9249, opened 2016-05-21
cmake: add ocf support

tchaikov:
a new option is added: WITH_OCF. it is OFF by default.


tchaikov:
@cbodley could you help review this one also? i am preparing a debian/rules which uses cmake. and this is one of the missing pieces to that goal.

tchaikov:
i checked the generated ceph and rbd file, they are the same as the ones generated by autoconf. and all the agent files (ceph, rbd, osd, mds, mon) are installed as expected.

PR #9250, opened 2016-05-21
osdc/Objecter: move C_TwoContexts class to cc file

stiopaa1:


liewegas:
@stiopaa1 there are some items in the Cleanup section of the trello board that might be of interest..

stiopaa1:
@liewegas 
Will have a look, thanks!

PR #9255, opened 2016-05-22
ceph_test_msgr: add verbose info for pending connections

yuyuyu101:


yuyuyu101:
let ticket like (http://tracker.ceph.com/issues/15944) can be dig.

PR #9259, opened 2016-05-23
mon: MonClient may hang on pinging an unresponsive monitor

xiexingguo:
On timedout, WaitUntil() method return a positive ETIMEDOUT error number.
It never returns -ETIMEDOUT.

### DIFF
--- a/src/test/pybind/test_rados.py
+++ b/src/test/pybind/test_rados.py
@@ -143,8 +143,10 @@ class TestRados(object):
         cmd = {'prefix': 'mon dump', 'format':'json'}
         ret, buf, out = self.rados.mon_command(json.dumps(cmd), b'')
         for mon in json.loads(buf.decode('utf8'))['mons']:
-            buf = json.loads(self.rados.ping_monitor(mon['name']))
-            assert buf.get('health')
+            while True:
+                buf = json.loads(self.rados.ping_monitor(mon['name']))
+                if buf.get('health'):
+                    break
 
     def test_create(self):
         self.rados.create_pool('foo')
--- a/src/mon/MonClient.cc
+++ b/src/mon/MonClient.cc
@@ -427,6 +427,7 @@ void MonClient::shutdown()
   monc_lock.Unlock();
 
   if (initialized) {
+    finisher.wait_for_empty();
     finisher.stop();
   }
   monc_lock.Lock();
--- a/src/mon/MonClient.h
+++ b/src/mon/MonClient.h
@@ -72,7 +72,7 @@ struct MonClientPinger : public Dispatcher {
     int ret = 0;
     while (!done) {
       ret = ping_recvd_cond.WaitUntil(lock, until);
-      if (ret == -ETIMEDOUT)
+      if (ret == ETIMEDOUT)
         break;
     }
     return ret;


tchaikov:
lgtm.

`ping_monitor()` is only used by `rados_ping_monitor()` in librados and ceph's `ping` commnad, so it's not necessary to backport this fix.

yuriw:
@xiexingguo we see "TestRados.test_ping_monitor ... ERROR"

in run http://pulpito.ceph.com/yuriw-2016-05-24_20:27:36-rados-wip-yuri-testing-distro-basic-smithi/
Job/log http://qa-proxy.ceph.com/teuthology/yuriw-2016-05-24_20:27:36-rados-wip-yuri-testing-distro-basic-smithi/212674/teuthology.log

Suspecting this PR is at fault per IRC with @athanatos .

xiexingguo:
@yuriw Thanks, yuri. I'll check and verify it locally.

xiexingguo:
@tchaikov I retested this above test case that failed these changes and it passed(http://daisycloud.org:9091/xxg-2016-06-13_15:24:56-rados-wip-xxg-testing---basic-plana/468/).
So the problem I guess is it happened that one of the monitors was unable to up and respond in 300s, which wouldn't be a problem before this change as we would wait forever(so perhaps this test case was going to die under this case?) 

Could you review (and retest, if possible) this pr for me? Thanks!

athanatos:
@xiexingguo It sounds like the failure is non-deterministic.  You need to change the test so that we don't get false failures.

xiexingguo:
Fixed test case and retested here:
http://www.daisycloud.org:9091/teuthology-2016-07-27_10:32:12-rados-wip-xxg-testing---basic-plana/
and passed.

src/test/pybind/test_rados.py
```diff
@@ -143,8 +143,10 @@ def test_ping_monitor(self):
         cmd = {'prefix': 'mon dump', 'format':'json'}
         ret, buf, out = self.rados.mon_command(json.dumps(cmd), b'')
         for mon in json.loads(buf.decode('utf8'))['mons']:
-            buf = json.loads(self.rados.ping_monitor(mon['name']))
-            assert buf.get('health')
+            while True:
```
tchaikov: could you elaborate a little bit on this change?

xiexingguo: See below：
I retested this above test case that failed these changes and it passed(http://daisycloud.org:9091/xxg-2016-06-13_15:24:56-rados-wip-xxg-testing---basic-plana/468/).
So the problem I guess is it happened that one of the monitors was unable to up and respond in 300s, which wouldn't be a problem before this change as we would wait forever(so perhaps this test case was going to die under this case?)

xiexingguo: @tchaikov To be more specific:  
before this change this test case won't be a problem as we'll ping and wait for response forever.
But this change allow monitor to return error on ETIMEDOUT which will fail the test...

yuriw:
re-run
http://pulpito.ceph.com/yuriw-2016-08-03_08:24:51-rados-wip-yuri-testing2_2016_08_01-distro-basic-smithi/

http://pulpito.ceph.com/yuriw-2016-08-01_17:57:24-rados-wip-yuri-testing2_2016_08_01-distro-basic-smithi/

PR #9260, opened 2016-05-23
mds: fix race between StrayManager::{eval_stray,reintegrate_stray}

ukernel:
StrayManager::eval_stray() is called after Server::respond_to_request()
drops locks. So it can race with StrayManager::reintegrate_stray()

Fixes: http://tracker.ceph.com/issues/15920

### DIFF
--- a/src/mds/Server.cc
+++ b/src/mds/Server.cc
@@ -5569,8 +5569,9 @@ void Server::_unlink_local_finish(MDRequestRef& mdr,
   // removing a new dn?
   dn->get_dir()->try_remove_unlinked_dn(dn);
 
-  // clean up?
-  if (straydn) {
+  // clean up ?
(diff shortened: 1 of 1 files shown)


jcsp:
I agree this is correct.  However I think the is_null() check would make more sense up in the Server level, with a comment explaining that we must check in case another request happened between the original request and the rename_finish we're in.(then an assertion in StrayManager).

If we put the check in the StrayManager level it's hard to explain/remember what it's for and why it's important

gregsfortytwo:
@jcsp, latest patch looks good to me? (Just a naive glance though.)

jcsp:
Yep, looks good

jcsp:
http://pulpito.front.sepia.ceph.com/jspray-2016-05-25_19:00:36-fs-wip-jcsp-testing-20160526---basic-mira/

PR #5624, opened 2015-08-20
rgw : setting max number of buckets for users via ceph.conf option

vumrao:
This patch adds a new option "rgw_user_max_buckets" for setting
max number of buckets for users via ceph.conf.

Fixes #12714

Signed-off-by: Vikhyat Umrao <vumrao@redhat.com>

PR #9261, opened 2016-05-23
rgw: Let the default quota settings take effect during user creation

prallabh:
rgw: Let the default quota settings take effect during user creation

Signed-off-by: Pavan Rallabhandi <prallabhandi@walmartlabs.com>

### DIFF
--- a/doc/radosgw/admin.rst
+++ b/doc/radosgw/admin.rst
@@ -400,6 +400,15 @@ To see how much of the quota a user has consumed, execute the following::
 .. note:: You should execute ``radosgw-admin user stats`` with the 
    ``--sync-stats`` option to receive the latest data.
 
+Default Quotas
+--------------
+
+You can set default quotas in the config.  These defaults are used when
+creating a new user and have no effect on existing users. If the
+relevant default quota is set in config, then that quota is set on the
+new user, and that quota is enabled.  See ``rgw bucket default quota max objects``,
+``rgw bucket default quota max size``, ``rgw user default quota max objects``, and
+``rgw user default quota max size`` in `Ceph Object Gateway Config Reference`_
 
 Reading / Writing Global Quotas
 -------------------------------
@@ -468,3 +477,4 @@ ranges for trim operations. ::
 
 .. _radosgw-admin: ../../man/8/radosgw-admin/
 .. _Pool Configuration: ../../rados/configuration/pool-pg-config-ref/
+.. _Ceph Object Gateway Config Reference: ../config-ref/
--- a/doc/radosgw/config-ref.rst
+++ b/doc/radosgw/config-ref.rst
@@ -346,6 +346,40 @@ Ceph configuration file, the default value will be set automatically.
 :Type: Boolean
 :Default: ``false``
 
+
+``rgw bucket default quota max objects``
+
+:Description: Default max number of objects per bucket. Set on new users,
+              if no other quota is specified. Has no effect on existing users.
+:Type: Integer
+:Default: ``-1``
+
+
+``rgw bucket default quota max size``
+
+:Description: Default max capacity per bucket, in bytes. Set on new users,
+              if no other quota is specified. Has no effect on existing users.
+:Type: Integer
+:Default: ``-1``
+
+
+``rgw user default quota max objects``
+
+:Description: Default max number of objects for a user. This includes all
+              objects in all buckets owned by the user. Set on new users,
+              if no other quota is specified. Has no effect on existing users.
+:Type: Integer
+:Default: ``-1``
+
+
+``rgw user default quota max size``
+
+:Description: The value for user max size quota in bytes set on new users,
+              if no other quota is specified.  Has no effect on existing users.
+:Type: Integer
+:Default: ``-1``
+
+
 Regions
 =======
 
--- a/src/common/config_opts.h
+++ b/src/common/config_opts.h
@@ -1398,7 +1398,7 @@ OPTION(rgw_bucket_quota_ttl, OPT_INT, 600) // time for cached bucket stats to be
 OPTION(rgw_bucket_quota_soft_threshold, OPT_DOUBLE, 0.95) // threshold from which we don't rely on cached info for quota decisions
 OPTION(rgw_bucket_quota_cache_size, OPT_INT, 10000) // number of entries in bucket quota cache
 OPTION(rgw_bucket_default_quota_max_objects, OPT_INT, -1) // number of objects allowed
-OPTION(rgw_bucket_default_quota_max_size, OPT_LONGLONG, -1) // Max size of object in kB
+OPTION(rgw_bucket_default_quota_max_size, OPT_LONGLONG, -1) // Max size of object in bytes
 
 OPTION(rgw_expose_bucket, OPT_BOOL, false) // Return the bucket name in the 'Bucket' response header
 
@@ -1409,7 +1409,7 @@ OPTION(rgw_user_quota_sync_interval, OPT_INT, 3600 * 24) // time period for accu
 OPTION(rgw_user_quota_sync_idle_users, OPT_BOOL, false) // whether stats for idle users be fully synced
 OPTION(rgw_user_quota_sync_wait_time, OPT_INT, 3600 * 24) // min time between two full stats sync for non-idle users
 OPTION(rgw_user_default_quota_max_objects, OPT_INT, -1) // number of objects allowed
-OPTION(rgw_user_default_quota_max_size, OPT_LONGLONG, -1) // Max size of object in kB
+OPTION(rgw_user_default_quota_max_size, OPT_LONGLONG, -1) // Max size of object in bytes
 
 OPTION(rgw_multipart_min_part_size, OPT_INT, 5 * 1024 * 1024) // min size for each part (except for last one) in multipart upload
 OPTION(rgw_multipart_part_upload_limit, OPT_INT, 10000) // parts limit in multipart upload
--- a/src/rgw/rgw_quota.cc
+++ b/src/rgw/rgw_quota.cc
@@ -823,8 +823,6 @@ class RGWQuotaHandlerImpl : public RGWQuotaHandler {
   RGWRados *store;
   RGWBucketStatsCache bucket_stats_cache;
   RGWUserStatsCache user_stats_cache;
-  RGWQuotaInfo def_bucket_quota;
-  RGWQuotaInfo def_user_quota;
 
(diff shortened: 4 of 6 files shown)


prallabh:
@cbodley ping

yehudasa:
@prallabh can you describe why this is needed? I'm not 100% sure that it makes sense, but it's a grey area.

prallabh:
@yehudasa This was tracked via http://tracker.ceph.com/issues/12997 and http://tracker.ceph.com/issues/12919, but the fixes made via those trackers didn't seem to address the user/bucket quota settings during user creation. This PR is to fix the same.

cbodley:
@prallabh as I understand this feature, these default quota config variables are only used as a fallback if the given user does not have a quota set. they're only checked during quota enforcement, and aren't ever stored with the users themselves

consider an admin that wants to change these default quota config variables from time to time. if we overwrite the user's quota settings, as your patch does, quota enforcement will always use the values from the user's creation, and won't ever see the updated values from configuration

dang:
@cbodley is correct, the feature was never intended to set quota on anything, only be a last ditch fallback for enforcing quota on something that had no other quota set.   Using them as a set quota would be a different feature.  It may be desired, but it's not a bug as such.

wwang001:
I think the intention on the original commit was a bit lost. I know because I was the one that asked for the feature to have default quotas. As it stands, #6400 has introduced some problems, such as:
- defining a default bucket max obj results in a global value that cannot be overridden at the user level
  - may be true of other vars
- someone debugging quota issues will be confused to see -1 at the user scope, not knowing that there may be a global default set

I would prefer if we took @prallabh's approach. If defaults are defined in ceph.conf, enable user quotas upon user creation, and explicitly inherit the values from ceph.conf. This way we can fall back to the original code that handled quotas on a user basis, essentially reverting #6400. I realize this only affects new users, but that's okay. This is how max_buckets originally was supposed to work, but even that still looks broken. I wanted a var in ceph.conf for max_buckets, so that we can override the hardcoded default of 1000. Instead we have introduced a new variable, rgw_user_max_buckets, which is ignored upon user creation, so we still get the old hardcoded 1000.

jsaintrocc:
Agree with @wwang001 that this approach is more straightforward.  I've been trying to use the current default quota implementation and it doesn't work as expected.  Setting a rgw_bucket_default_quota_max_objects/size will override user/bucket level quotas, if it is set lower than those quotas.  Operationally speaking, it's confusing to have defaults in ceph.conf while quotas are normally set with radosgw-admin, which can't see those defaults.

prallabh:
@cbodley It wasn't clear to me of the original commit intent, but this patch would set user quotas during user creation by reading from the conf variables. As you said, in the current scheme of things, if the conf values were to be changed frequently, the existing users would also get to read the updated values (and irrespective of the existing number of objects/size, the updated values are always honored completely, which am not sure is desired).

If the users use the CLI/API to change the quota per user, then the conf values are anyways overridden and this patch does exactly the same. May be as @dang mentioned, this might be a new requirement, which I thought the mentioned trackers addressed.

dang:
Okay, so I'm trying to understand the issues here.

1. A mysterious 1000 is showing up from somewhere.  I'm not yet sure where.  I agree this needs to be addressed.
2. default bucket quota cannot be overridden on a user level.  This was intentional, bucket quota preempts user quota.  Is there a reason why a user should be able to override any bucket quota?
3. The default quotas do not show up in the admin command outputs.  The proposed solution is to set user quotas to the defaults, but would having the defaults show up in the admin command without the user quotas being set be a solution?

wwang001:
@dang 
1. I tested this using radosgw-admin user creation, since my test env doesn't have Keystone available. Looks like I inherit the hardcoded old max_buckets intead of rgw_user_max_buckets. This is with ceph version 10.2.1 (3a66dd4f30852819c1bdaa8ec23c795d4ad77269). I did validate the rgw_user_max_buckets was properly set before user creation.
2. I think there is some confusion. I believe most operators would expect that user quota takes precedent over global values. For instance, I want to set a lower global number for default rgw user max buckets, default bucket max objs, etc, but I know some tenants/users will need to exceed those values. To satisfy that need, we expect to be able to set the values at a lower scope, such as at the user or user's bucket scope.
3. Not opposed to that idea, as long as it is displayed in a manner than reflects order of operations. However, that being said, I personally think it is cleaner to have a single scope for quotas (user/bucket, and not global). 

Lastly, I saw no docs for the previous commit. There is no way anyone would have found these changes. Happy to help with this, as soon as we know what the behavior is going to be like.

cbodley:
@wwang001 thanks for the feedback. as a followup, what behavior would you expect after running `radosgw-admin quota disable`? should that user be exempt from quota enforcement, or should it fall back to the default quota settings?

wwang001:
@cbodley: I'd expect that user/bucket to be exempt. 

More context. We have a chicken and egg scenario when RGW is configured to use Keystone for auth/user mgmt. There was no way to pre-set a quota, prior to #6400.  Quota management didn't make sense, because the users could be added to Keystone, but until they try to access the RGW, there is no user in the RGW metadata to modify/set quotas. Having defaults allows us to avoid setting quotas on an individual basis for every user.

I would say that the scope on this PR has increased, depending on what we want to do with #6400. Keep globals or collapse back down to user/bucket level.

yehudasa:
@wwang001 does setting quota on the zone (per user/bucket) solve your issues?

yehudasa:
@wwang001 that is, setting user/bucket quota on the zone-group, not the zone

dang:
@wwang001 Just to be clear, the default_max_buckets part is unrelated to default quota, right?  Can you open a separate tracker for that, so we can disambiguate this one?

jsaintrocc:
@dang 

Agree with @wwang001 that, ideally, user quota should take precedent over global quotas.  Otherwise if we set a global quota and we have a single user that needs an exception, we have no way of doing that without increasing the global quota for all users.  Also the current behavior is inconsistent.  If we set the user quota less than the global the user quota applies but if we set the user quota higher than the global quota the global applies.

wwang001:
@dang: That's how this whole thing started. A request from me to make max_buckets configurable, instead of a hard coded value. The scope grew a when I suggested that all quotas should be configurable. I made the (bad) assumption that they would be implemented like max_buckets, where the value is populated into the user's quota upon user creation. I did not realize we were going to create a new global scope, adding all the confusion, and inconsistent behavior. The original ticket got lumped together. I would like to see the fix for all of these be in the same PR, since max_buckets is just another default quota value.

@yehudasa: Sorry, I haven't had a chance to try it, but I don't really see how that will help. It would still be treated like the global defaults.

prallabh:
@dang I would be happy to extend this patch if we agree upon the intended quota behavior.

dang:
@wwang001 The issue is that, code wise, max_buckets is not part of the quota system, so I'd prefer to treat it separately.

@prallabh Additions will be necessary, for sure.  I'll post a description of my understanding in a bit, so that we can all be on the same page.

dang:
Okay, here's what I'm understanding the requested behavior is:

1. If the user quota is set at any point, including creation, it should be used.
2. If the user quota is not set at creation, then the then-current default quota should be set on the user.
3. If the user quota is disabled, the user should not have a quota, regardless of the default quota
4. If the default quota is changed, the new value should only apply to future created users, not to current users using the default quota.

Is this all correct?  Am I missing a behavior?

@wwang001 I agree that default_max_buckets is broken (it seem to have been a simple replace error on the original commit).  I'd like to handle that as a separate issue from this one.

dang:
(Note, there's currently no way to set quota on user creation, but the code should be able to handle that if it's added in the future...)

prallabh:
Have added a commit that essentially doesn't bother to look at config values during quota enforcement. This should at least address the summary points detailed above.

wwang001:
Sorry! I meant to reply earlier. @dang 

1. Correct
2. Correct
3. Correct, however, there should be a way to enable users quotas (user and bucket) by default in ceph.conf. This can be a separate PR/tracker.
4. Correct

dang:
@wwang001 Sorry, for 3 I meant at any point after user creation, if the quota is disabled.

@prallabh Looks good code wise, we just need documentation.  Something like in the branch here:
https://github.com/ceph/ceph/tree/def-quota-doc

Feel free to pull that in if you agree it's sufficient, or add any changes necessary.

prallabh:
@wwang001 For #3, the current PR handles the case of enabling quotas upon user creation.

@dang thanks, will do.

Will update this PR once it's ready for review

wwang001:
Not sure what happened with max_buckets. I can't replicate the issue I had. PEBKAC?

Code logic on the new commit looks good to me. Suggestions on docs:

``rgw bucket default quota max objects``
Default max number of objects per bucket. Set on new users, if no other quota is specified. Has no effect on existing users.

``rgw bucket default quota max size``
Default max capacity per bucket, in KB. Set on new users, if no other quota is specified. Has no effect on existing users.

``rgw user default quota max objects``
Default max number of objects for a user. This includes all objects in all buckets owned by the user. Set on new users, if no other quota is specified. Has no effect on existing users.

``rgw user default quota max size``
Default max capacity in KB for a user. This includes the capacity of all buckets owned by the user. Set on new users, if no other quota is specified. Has no effect on existing users.

vumrao:
> I wanted a var in ceph.conf for max_buckets, so that we can override the hardcoded default of 1000. > Instead we have introduced a new variable, rgw_user_max_buckets, which is ignored upon user creation, so we still get the old hardcoded 1000.

@wwang001 To me looks like some confusion for `max_bucket` : 
- As I had worked for this feature when you had requested this feature : http://tracker.ceph.com/issues/12714 via an internal ticket.
-  PR for this feature :  https://github.com/ceph/ceph/pull/5624
- In my initial testing in our QA also it worked as expected. Once you set `rgw_user_max_buckets` with a value in [global] section of ceph.conf. This option will override the hardcoded default of 1000 for all new user creation.

~~~
in ceph.conf : 
[global]
rgw_user_max_buckets = 5000

[....]

# radosgw-admin user create --uid="test" --display-name="Test User"
{
    "user_id": "test",
    "display_name": "Test User",
    "email": "",
    "suspended": 0,
    "max_buckets": 5000, <----
    "auid": 0,
    
    [....]
~~~

prallabh:
@wwang001 Have corrected the docs.

@vumrao Yes, max_buckets works as expected, thanks!

@dang Have pushed all the commits, looks like the max size is now tracked in bytes, corrected the same in the relevant sections. Please help review/merge the same. I guess the 'pending-discussion' label can be removed now.

PR #9273, opened 2016-05-23
osd/OSD.h: remove closing brace } from comment

stiopaa1:
This is confusing vim when moving to matching brace of OSD class.

### DIFF
--- a/src/osd/OSD.h
+++ b/src/osd/OSD.h
@@ -1582,9 +1582,9 @@ private:
 
 private:
   /**
-   *  @defgroup monc helpers
-   *
-   *  Right now we only have the one
+   * @defgroup monc helpers
+   * @{
+   * Right now we only have the one
    */
 
   /**


src/osd/OSD.h
```diff
@@ -1596,7 +1596,7 @@ class OSD : public Dispatcher,
    * sufficient.
    */
   void osdmap_subscribe(version_t epoch, bool force_request);
-  /** @} monc helpers */
+  /** @ monc helpers */
```
dzafman: This ends a Doxygen grouping.  See defgroup in https://www.stack.nl/~dimitri/doxygen/manual/commands.html   I think  the problem here is a missing **@{**.

stiopaa1:
@dzafman 
Thanks for the link, I have updated the commit. Does this look reasonable now?

dzafman:
@stiopaa1 I'm not sure how, but we should try to build the documentation based on this change.  Maybe execute admin/build-doc????

src/osd/OSD.h
```diff
@@ -1582,9 +1582,9 @@ class OSD : public Dispatcher,
 
 private:
   /**
-   *  @defgroup monc helpers
-   *
-   *  Right now we only have the one
+   * \defgroup monc helpers
```
tchaikov: nit, we use "@defgroup" in general.

tchaikov:
@stiopaa1 @dzafman `admin/build-doc` does not build `OSD.h`. maybe, instead, we can use "doxgen" directly and check the output? please note, the default output format is XML, see Doxyfile. might want to enable HTML output also to read the generated docs.

stiopaa1:
Will have a look at this at the weekend.

src/osd/OSD.h
stiopaa1: @tchaikov 

here:
https://www.stack.nl/~dimitri/doxygen/manual/grouping.html

it says it should be \defgroup?

stiopaa1:
I have run doxygen and it seems to be working:
doxy1

and after clicking on osdmap_subscribe

doxy2

src/osd/OSD.h
tchaikov: it's javadoc style, see https://www.stack.nl/%7Edimitri/doxygen/manual/docblocks.html#docexamples .

stiopaa1: @tchaikov 
I've changed it back to @defgroup

tchaikov:
lgtm

PR #9280, opened 2016-05-24
rbd: check value of config override when setting image-meta

zhuangzeqiang:
Fixes: http://tracker.ceph.com/issues/15522

Signed-off-by: zhuangzeqiang <zhuang.zeqiang@h3c.com>

### DIFF
--- a/src/librbd/internal.cc
+++ b/src/librbd/internal.cc
@@ -2627,6 +2627,15 @@ remove_mirroring_image:
   int metadata_set(ImageCtx *ictx, const string &key, const string &value)
   {
     CephContext *cct = ictx->cct;
+    string start = ictx->METADATA_CONF_PREFIX;
+    size_t conf_prefix_len = start.size();
+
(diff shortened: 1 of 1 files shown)


zhuangzeqiang:
@dillaman thanks for your explanation，is that ok？

dillaman:
lgtm

yuriw:

tchaikov:
@zhuangzeqiang if you were not posting this PR from the future, could you fix your localtime? your commit's timestamp is "Sat Oct 7 16:52:07 2017 +0800". and thanks to the way how github shows the commits you will be always the first one in https://github.com/ceph/ceph/commits/master until we have a newer commit.

zhuangzeqiang:
@tchaikov Very embarrassed for that, and what should i do to fix it?

tchaikov:
@zhuangzeqiang no. we can hardly fix this without rewriting the git history after your commit. but guess it's always good to have a correct time, no? =)

PR #9286, opened 2016-05-24
OSD: op lock completion patch

myoungwon:
op lock completion: minimize pg lock contention
----------

as discussed in last CDM (http://pad.ceph.com/p/osd-locking-optimization),
i introduce op lock completion.

first concept code use original completion codes at initial time.
Then, change op lock completion (for debugging)

a. performance (2 replica)
66K IOPS (original jewel)
76K IOPS (original jewel + op lock completion)

performance will be improved in higher IOPS workload as discussed in CDM.

b. current limitations
1. Replicated backend only
2. only tested by Random write workload. (via KRBD)

c. remain works
1. clean data structure
2. smart completion thread (event driven)
3. EC support
4. ordering completion
5. more test..

Signed-off-by: Myoungwon Oh \<omwmw@sk.com>


tchaikov:
@myoungwon could you remove unrelated commits from your PR, and just keep d05daa6 in it?

myoungwon:
rebase is done. thanks.

src/osd/OSD.cc
```diff
@@ -8725,6 +8751,34 @@ void OSD::enqueue_op(PG *pg, OpRequestRef& op)
...
+    spg_t pgid = pg->get_pgid();
+    const pg_pool_t* pp = osdmap->get_pg_pool(pgid.pool());
+    if (r->get_result() == MSG_OSD_OP_COMP_RESULT && 
+	pp->get_type() == pg_pool_t::TYPE_REPLICATED) {
+      
+
+      PGBackend * pg_backend = pg->get_pgbackend();
+      if (!pg_backend) {
+	dout(0) << __func__ << " PGBackend is null " << dendl;
+	pg->unlock();
```
athanatos: pg isn't locked here.

src/osd/OSD.cc
```diff
@@ -8725,6 +8751,34 @@ void OSD::enqueue_op(PG *pg, OpRequestRef& op)
...
+	assert(0);
+      }
+
+      ReplicatedBackend * rep_backend = dynamic_cast<ReplicatedBackend*>(pg_backend);
+      if (!rep_backend) {
+	dout(0) << __func__ << " ReplicatedBackend is null " << dendl;
+	pg->unlock();
+	assert(0);
+      }
+      rep_backend->sub_op_modify_reply_no_pg_lock(op);
```
athanatos: We can't reply from here.  We are sitting in an ms_fast_dispatch call.  The right thing to do is to change the dequeue path to not lock the pg for this type of message.

src/osd/OSD.h
```diff
@@ -2505,11 +2528,29 @@ class OSD : public Dispatcher,
...
 };
 
 //compatibility of the executable
 extern const CompatSet::Feature ceph_osd_feature_compat[];
 extern const CompatSet::Feature ceph_osd_feature_ro_compat[];
 extern const CompatSet::Feature ceph_osd_feature_incompat[];
 
+// for no pg lock completion
+#define OP_COMP_PRIMARY_OP 1
+#define OP_COMP_REPLICA_OP 2
```
athanatos: Use an enum for this sort of thing.

athanatos:
This is going to take me some time to work through.  I think we can get the same pattern with much less complexity, let me think about this a bit.

src/osd/OSD.cc
myoungwon: Yes, I intend to avoid existing I/O path with PG lock because the role of ack is simply notifying completion of op. also, sending ack is very short path and enqueue to current sharded work queue cause contention. Therefore, I think ack can be issued without PG lock and enqueuing via op lock completion path. (updating PG related data will be processed at lazy_completion_func or comp_thread.)

src/osd/OSD.cc
myoungwon: Right. I thought sending ack in an ms_fast_dispath cause no overhead to messenger code because it can be processed fast. otherwise sending sharded work queue cause consuming additional CPU.

src/osd/OSD.h
myoungwon: OK, i will fix it.

myoungwon:
I look forward to hearing new idea from you.

XinzeChi:
@myoungwon  Thanks for your work. Is there benchmark result for this PR?
filestore_ondisk_finisher_threads and filestore_apply_finisher_threads could help to reduce the pg lock contention.

myoungwon:
@XinzeChi  Right. the number of finisher could help to reduce pg lock contention. However, only increasing thread number cause not only requiring more CPU but also happening same problem in high IOPS workload. Therefore, I think both of the optimizations (increasing thread number and minimize lock contention) are needed to applied for higher performance.
In my presentation in last CDM, page 7 show performance results. parallel completion (this is same method as you mentioned) can help performance around 26K as marked 6. If we apply op lock completion (as marked 5) + parallel completion, more performance (58K) gain is found (please note that this is null test, assuming high IOPS).
Anyway, following is my performance result from default jewel code (10.2.1) via fio.
default jewel
IOPS: 66640 
Latency: 8.6
CPU util: 56.4 %

default jewel + op lock completion
IOPS: 75860
Latency: 7.5
CPU util: 65 %

This is from default version. I think performance will be improved in higher IOPS workload.
(more than 100 KIOPS via performance tuning)

zhouyuan:
@myoungwon Is there any special configuration needed for tests there? I applied your patch on top of Jewel and all pg got stuck when creating

myoungwon:
@zhouyuan no configuration is needed. current code use pg lock completion (original scheme) at initial time, then change op lock completion. (i will make complete code after hearing comment from Samuel) anyway,  this code is tested in the case of MOSDOp (general Read, Write cases). Therefore, please use existing jewel files for jobs such as creating pgs .., then change binary files to patched jewel for testing general Read, Write.
Can you give me logs?

zhouyuan:
@myoungwon I see, thanks for the explanation. I'm testing on a fresh virtual cluster with vstart.sh. Cluster is able to run but the pg creation got stuck. Let me try again with the replacing the binaries on an existing cluster.

src/osd/OSD.cc
athanatos: I mean, you are calling pg->unlock().  I don't think the pg is locked here.

myoungwon: Right. I misunderstood. pg->unlock() is not needed. I will fix it.

src/common/TrackedOp.h
```diff
@@ -162,15 +162,35 @@ class TrackedOp {
...
   atomic<bool> is_tracked; //whether in tracker and out of constructor
+public:
+  Mutex comp_lock;
+  bool can_op_lock_comp;
+  void * rep_gather;
+  uint64_t tid; // for debugging
+  bool all_applied;
+  bool all_committed;
+  bool is_sent_ack;
+  bool completed;
```
athanatos: This is very much the wrong place for this.  The right answer is to leave this information in RepGather and arrange for that state to be separately locked.

src/osd/OSD.cc
```diff
@@ -1455,6 +1459,25 @@ void OSDService::queue_for_peering(PG *pg)
...
+      break;
+    }
+    RWLock::RLocker l(osd->pg_map_lock);
+    ceph::unordered_map<spg_t, PG*>::iterator iter = osd->pg_map.begin();
+    for (; iter != osd->pg_map.end(); ++iter) {
+      PG * pg = iter->second; 
+      if (!pg) assert(0);
+      pg->lazy_completion_func(true);
+    }
+    sleep(1);
```
athanatos: Why the sleep?

src/osd/OSD.cc
```diff
@@ -1455,6 +1459,25 @@ void OSDService::queue_for_peering(PG *pg)
...
+  while (1) {
+    // dout(0) << __func__ << " " << __LINE__ << " enter!! " << dendl;
+    if (!osd) assert(0);
+
+    if (comp_thread_stop) {
+      dout(0) << __func__ << " " << __LINE__ << " out " << dendl;
+      break;
+    }
+    RWLock::RLocker l(osd->pg_map_lock);
+    ceph::unordered_map<spg_t, PG*>::iterator iter = osd->pg_map.begin();
```
athanatos: So we are iterating through all PGs once/sec?  That's kind of odd...

src/osd/OSD.h
```diff
@@ -1175,6 +1175,29 @@ class OSDService {
   }
 #endif
 
+  void check_completion_item();
+#if 0
```
athanatos: Certainly, we don't want to be leaving if'd out code in the code base.

src/osd/OSD.h
```diff
@@ -1175,6 +1175,29 @@ class OSDService {
...
+    CompletionThread(OSD *osd) : osd(osd) {}
+    void *entry() {
+      osd->check_completion_item();
+      return 0;
+    }
+  }comp_thread;
+#endif
+  class CompletionThread : public Thread {
+  public:
+    OSDService * osd_srv;
```
athanatos: Formatting.

src/osd/OSD.h
```diff
@@ -1175,6 +1175,29 @@ class OSDService {
...
+    CompletionThread(OSDService *osd_srv) : osd_srv(osd_srv) {}
+    void *entry() {
+      osd_srv->check_completion_item();
+      return 0;
+    }
+  }comp_thread;
```
athanatos: Formatting.

src/osd/OSD.h
```diff
@@ -1175,6 +1175,29 @@ class OSDService {
...
+
+  bool comp_thread_stop;
```
athanatos: Should be an atomic type, accessed from multiple threads without synchronization.

src/osd/ReplicatedBackend.cc
```diff
@@ -663,6 +771,186 @@ void ReplicatedBackend::op_commit(
...
+      in_progress_ops.erase(op->tid);
+    }
+    in_progress_ops_lock.Unlock();
+    return;
+  }
+  op->op->comp_lock.Unlock();
+  dout(20) << __func__ << " " << __LINE__ << " out tid:" << op->tid << dendl;
+}
+
+void ReplicatedBackend::sub_op_modify_reply_no_pg_lock(OpRequestRef op)
```
athanatos: Do we still need the versions of these that *do* take a pg lock?

myoungwon: @athanatos I think old version codes which based on pg lock mean sub_op_modify_rep(). old version codes can be removed if op lock completion is finished.

src/osd/OSD.h
myoungwon: @athanatos Actually, this is checking whether thread is stop or not. I don't think multiple threads access this value because stop is set rarely. other threads in OSD use same approach (For example, agent thread and heartbeat thread..)

src/osd/OSD.h
myoungwon: @athanatos I will fix it.

src/osd/OSD.cc
myoungwon: @athanatos Need to iteration but not once/sec. this polling function can be changed to event-driven approach. (sleep until happening events). Note that another completion function for processing completion is in execute_ctx. Therefore, most of processing completion will be handled at execute_ctx which already hold pg lock (at issuing context).  This is for preventing unfinished jobs.

src/osd/OSD.cc
myoungwon: @athanatos  Actually, current completion thread use polling for checking completion item.  I think using Cond.wait()  and send event when completion item occur is right answer.

src/common/TrackedOp.h
myoungwon: @athanatos Right. This is wrong place. RepGather is right place. I will fix it.

src/osd/OSD.h
athanatos: There are at least two threads accessing it -- the one stopping and the one being stopped.  It's possible that other things in the OSD are buggy, that doesn't mean we want to introduce new bugs here.

athanatos:
My main problem with this patch series is that it doesn't really attempt to integrate the new changes cleanly into the OSD code.  The basic concept is sound: it makes perfect sense to attempt to perform the op replies without needing the pg lock, and this code is useful as a proof-of-concept. However, this series needs a lot of work to get to a mergeable point.  Readers of this code need to be able to readily identify which pieces of state can be safely accessed without the PG lock, and which ones can't.  The changes also need to include the ECBackend.  I see no reason at all to leave around the old code paths -- they shouldn't be needed once this is complete.

In summary, the basic concept is sound, but the code needs to be radically restructured to make it easy to understand what is going on and to allow us to be confident that the changes are correct.

myoungwon:
@athanatos I carry my intention (showing concept) to you fully. Code will be restructured in order to make mergeable. Do you have any suggestions for this design? if you have, i will apply next code.

athanatos:
As an example, you'll need to be able to cleanly queue things in the OpWQ without requiring the PG lock, so you'll need to rework those interfaces and structures.  I have a branch doing some of that: https://github.com/athanatos/ceph/commits/wip-osd-reply.  I haven't tested it and will definitely be rebasing and changing that branch in the future, but feel free to base your work on it.

Another example is that OpContext users can schedule callbacks using on_commit, on_apply, etc.  Those currently assume that the PG lock will be held when the callbacks are run.  Clearly, that won't be a valid assumption any longer at least for the common op path.  However, some users of the repop system (see the lost-unfound code an example) probably will still need to be able to count on scheduling callbacks with the PG lock, so you'll probably need to add on_commit_repop_lock, on_apply_repop_lock, etc and ensure that you can handle both types of callback.

Another thing you need to handle is that the Replicated and EC schedule "blessed" callbacks to avoid having to explicitly take the PG lock.  You'll have to expand that part of the PGListener interface to allow creating a callback guarded by the Repop lock.

Furthermore, you don't seem to have handled any of the recovery or peering edge cases (or I wasn't able to find it, which amounts to the same thing).  Currently, when a callback is run, it gets the pg lock and is either still valid or is not still valid based on comparing its stored epoch with the PG's last_peering_reset.  That will no longer work, so during peering, you'll have to pass a message (by taking the repop lock and flipping some state) to cause previously scheduled callbacks to fail and to clean up now invalid state.

I *strongly* suggest that before you start writing more code, you take the time to write out your plan for which interfaces you'll need to change so that we can go over it.

myoungwon:
@athanatos Thanks for your comments. I have a additional question. if we don't use a pg lock at processing completion, updating PG metadatas (OMAP) such as lcod (last complete on disk) are delayed via queuing.
I think this delay doesn't cause bad effect. Because there are two reasons. First, if we want to read latest pg related information, just read information after updating delayed information in queue. Second, if system fail occur, PG maintain consistent status because PG metadata is written to disk in order. (note that delayed PG metadata already is written to disk). 
How are you thought about this?

athanatos:
I think it's safe because logically sending the reply to the client was not atomic with updating lcod anyway.

myoungwon:
@athanatos I will start code renewal with applying your comments.

myoungwon:
I am working on restructuring op lock completion patch. Below is interfaces and data structures.
Could you give me comments?

-- Write flow

We make struct CompletionItem in order to handle both type of OP (Primary, Replication)
RepGather and RepModify have CompletionItem.

A. Primary OP (based on struct RepGather)
1. execute_ctx()  >  2. do lazy completion > 3. queue_transaction() > 4. committed(acquire op lock) >  5. send ack(release op lock) > 6. applied(acquire op lock)  >  7. queueing lazy update work > 8. finish (release op lock)
1 ~ 3: with PG Lock

B. Replication OP (based on struct RepModify)
1. do lazy completion >  2. sub_op_modify()  > 3. queue_transaction() > 4. committed(acquire op lock)  >  5. send ack(release op lock)  >  6. applied(acquire op_lock)  >  7. queueing lazy update work >  8. finish (release op lock)
1 ~ 3: with PG Lock

C. Lazy completion worker (based on struct CompletionItem)
1. queueing lazy completion item from committed or applied worker  >  2. wake up and do lazy completion  >  3. sleep  >  go to step 1


-- Data structure.

class CompletionItem {
 // need PG lock
   /* update_peer_last_complete_ondisk from REPOP_REPLY */
   pg_shard_t fromosd;
   eversion_t lcod;
--> saved for update_peer_last_complete_ondisk
   
   /* last update applied */
   eversion_t applied_version;
--> used to check version during lazy completion

   /* completion func which need pg lock, C_UpdateLastRollbackInfoTrimmedToApplied */
   list<Context*> on_applied_with_pg_lock;
--> for calling complete during lazy completion

   /* Replica versions for MSG_OSD_REPOP */
   eversion_t rep_committed_version;
   eversion_t rep_applied_version;
--> for checking version

   version_t epoch;
--> used for making MOSDRepOpReply, send_message_osd_cluster,
  
/* last_update_ondisk, last_complete_ondisk */
   RepGather *repop;   
}


C_OSD_RepopApplied_no_pg_lock class
- handle apply message by calling repop_all_applied_no_pg_lock

C_OSD_RepopCommit_no_pg_lock class
- handle commit message by calling repop_all_committed_no_pg_lock


<in OSDService>
- add following functions

OSDService::check_lazy_completion(float min_comp_interval)
- finish completion jobs by using polling
OSDService::reserve_check_completion(utime_t time, spg_t & pg)
- used to notify completion item is inserted


<in ReplicatedPG class>
- add CompletionItem class to on_applied class
- add completion_ops_lock and map of completion_ops
- add following functions

ReplicatedPG::add_completion_q(CompletionItem *comp_item)
- entry to add item to completion queue

ReplicatedPG::do_lazy_completion(bool need_lock)
- finish completion by calling do_primary_comp and do_rep_comp
- do pg lock according to need_lock flag

ReplicatedPG::do_primary_comp(CompletionItem * comp_item)
- handle last_update_ondisk, last_complete_ondisk, last_update_applied
- send dup commit
- do similar role with eval_repop

ReplicatedPG::do_rep_comp(CompletionItem * comp_item)
- update last complet ondisk
- update on_applied_with_pg_lock

ReplicatedPG::repop_all_committed_no_pg_lock(RepGather *repop)
- call callback function registered in repop->on_commited

ReplicatedPG::repop_all_applied_no_pg_lock(RepGather *repop)
- call callback function registered in repop->on_applied


<in PGBackend class>
Listener::add_completion_q(CompletionItem *)
Listener::op_comp_context(Context *, CompletionItem *)


<in ReplicatedBackend>
- no_pg_lock will be removed if finished
- followings are functions to be changed
ReplicatedBackend::op_applied_no_pg_lock(InProgressOp *op)
ReplicatedBackend::op_commit_no_pg_lock(InProgressOp *op)
ReplicatedBackend::sub_op_modify_reply_no_pg_lock(OpRequestRef op)
ReplicatedBackend::sub_op_modify_applied_no_pg_lock(RepModifyRef rm)
ReplicatedBackend::sub_op_modify_commit_no_pg_lock(RepModifyRef rm)

- functions for calling completion
ReplicatedBackend::check_committed_completion(InProgressOp * op)
ReplicatedBackend::check_applied_completion(InProgressOp * op)
ReplicatedBackend::check_all_completion(InProgressOp * op)
--> called when commit and apply both is finished. add CompletionItem to completion queue

<in OpContext>
list<std::function<void()>> on_applied_op_lock;
list<std::function<void()>> on_committed_op_lock;
list<std::function<void()>> on_finish_op_lock;
list<std::function<void()>> on_success_op_lock;

<in RepGather>
list<std::function<void()>> on_applied_op_lock;
list<std::function<void()>> on_committed_op_lock;
list<std::function<void()>> on_finish_op_lock;
list<std::function<void()>> on_success_op_lock;

athanatos:
On Sun, Jul 10, 2016 at 8:03 PM, myoungwon oh <notifications@github.com>
wrote:

> I am working on restructuring op lock completion patch. Below is
> interfaces and data structures.
> Could you give me comments?
>
> -- Write flow
>
> We make struct CompletionItem in order to handle both type of OP (Primary,
> Replication)
> RepGather and RepModify have CompletionItem.
>
> A. Primary OP (based on struct RepGather)
> 1. execute_ctx() > 2. do lazy completion > 3. queue_transaction() > 4.
> committed(acquire op lock) > 5. send ack(release op lock) > 6.
> applied(acquire op lock) > 7. queueing lazy update work > 8. finish
> (release op lock)
> 1 ~ 3: with PG Lock
>
> B. Replication OP (based on struct RepModify)
> 1. do lazy completion > 2. sub_op_modify() > 3. queue_transaction() > 4.
> committed(acquire op lock) > 5. send ack(release op lock) > 6.
> applied(acquire op_lock) > 7. queueing lazy update work > 8. finish
> (release op lock)
> 1 ~ 3: with PG Lock
>
> C. Lazy completion worker (based on struct CompletionItem)
> 1. queueing lazy completion item from committed or applied worker > 2.
> wake up and do lazy completion > 3. sleep > go to step 1
>

You need to guarantee that the ops (on a particular PG) complete in the
same order as they are started.  That would seem to require that you need
to place the current repop queue state under a second lock and take it
during submission and completion of each op.  I don't want a new worker --
I want you to integrate those items into the OpWQ.  See the branch I
mentioned before for a way to generalize that queue to take a lock other
than the PG lock.


> -- Data structure.
>
> class CompletionItem {
> // need PG lock
> /* update_peer_last_complete_ondisk from REPOP_REPLY */
> pg_shard_t fromosd;
> eversion_t lcod;
> --> saved for update_peer_last_complete_ondisk
>
> /* last update applied */
> eversion_t applied_version;
> --> used to check version during lazy completion
>
> /* completion func which need pg lock,
> C_UpdateLastRollbackInfoTrimmedToApplied
> */ list<Context*> on_applied_with_pg_lock;
> --> for calling complete during lazy completion
>
> /* Replica versions for MSG_OSD_REPOP */
> eversion_t rep_committed_version;
> eversion_t rep_applied_version;
> --> for checking version
>
> version_t epoch;
> --> used for making MOSDRepOpReply, send_message_osd_cluster,
>
> /* last_update_ondisk, last_complete_ondisk */
> RepGather *repop;
>
> }
>
> C_OSD_RepopApplied_no_pg_lock class
>
>    - handle apply message by calling repop_all_applied_no_pg_lock
>
> C_OSD_RepopCommit_no_pg_lock class
>
>    - handle commit message by calling repop_all_committed_no_pg_lock
>
>
>    - add following functions
>
>
Like I said before, I do not want to see new callbacks that end with
_no_pg_lock.  You should be able to replace the existing ones.


>
>
> OSDService::check_lazy_completion(float min_comp_interval)
>
>    - finish completion jobs by using polling
>    OSDService::reserve_check_completion(utime_t time, spg_t & pg)
>    - used to notify completion item is inserted
>
>
>    - add CompletionItem class to on_applied class
>    - add completion_ops_lock and map of completion_ops
>    - add following functions
>
>
No, I want those completions put into the existing OpWQ.

src/osd/ReplicatedPG.cc
```diff
@@ -8311,6 +8359,28 @@ class C_OSD_RepopCommit : public Context {
...
+  ReplicatedPGRef pg;
+  boost::intrusive_ptr<ReplicatedPG::RepGather> repop;
+  ReplicatedPG::RepGather * real_repop;
+public:
+  C_OSD_RepopCommit_no_pg_lock(ReplicatedPG *pg, ReplicatedPG::RepGather *repop)
+    : pg(pg), repop(repop), real_repop(repop) {}
+  void finish(int flag) {
+    //generic_dout(0) << __func__ << " flag is " << flag << dendl;
+    if (flag == OP_COMP_ACK) {
+      pg->repop_all_committed_no_pg_lock(repop.get());
```
asheplyakov: Do we leak repop here? (there's a get(), but no put())

src/osd/ReplicatedPG.cc
```diff
@@ -8311,6 +8359,28 @@ class C_OSD_RepopCommit : public Context {
...
+      return;
+    } else if (flag == OP_COMP_ACK_COMP) {
+      pg->repop_all_committed_no_pg_lock(repop.get());
+    } 
+    CompletionItem * comp_item = new CompletionItem(OP_COMP_PRIMARY_OP);
+    comp_item->rep_gather = real_repop;
```
asheplyakov: On the other hand repop is stored here without incrementing the refcount

src/osd/ReplicatedPG.cc
```diff
@@ -8311,6 +8359,28 @@ class C_OSD_RepopCommit : public Context {
   }
 };
 
+class C_OSD_RepopCommit_no_pg_lock : public Context {
+  ReplicatedPGRef pg;
+  boost::intrusive_ptr<ReplicatedPG::RepGather> repop;
+  ReplicatedPG::RepGather * real_repop;
```
asheplyakov: Managing the same resource both with a raw and a "smart" pointer is very error prone

myoungwon:
I understand. Lazy completion worker can be integrate to the OP_WQ. I think your branch (divide items which need the PG lock via PGOpQueueable) is right way. Can i rework this patch based on your branch?

Anyway, i have a question for completion ordering.
You mentioned completion ordering. I think ordering is guaranteed because completed ops are enqueued to finishers (commit, apply)  in the same order as they are stared. (I do not revise the issue order) 
Could you explain me expected unordered scenario?

src/osd/ReplicatedPG.cc
myoungwon: @asheplyakov  This is the same as existing C_OSD_RepopApplied (which is the PG lock version).
put() will be called in eval_repop()

src/osd/ReplicatedPG.cc
myoungwon: Right. I fixed it already.

src/osd/ReplicatedPG.cc
myoungwon: I removed real_repop in reworking version.

athanatos:
The OpWQ has more than one thread/shard.  If they don't share a lock,
replies from the replicas could be handled out of order.
-Sam

myoungwon:
@athanatos  I created a merge request for discussion (https://github.com/athanatos/ceph/pull/4/commits/840239b5a32a802d3588e1757eb5d62312642aa6) (To do: Peering, EC support, Ordering ..) Completion ordering you mentioned above is protected by sdata_op_ordering_lock.

Could you give me comments?

athanatos:
No, it's not.  We drop that lock before processing the item -- nothing
prevents two threads from grabbing two items in order and then processing
them out of order -- that's why the queue abstraction has a lock concept
and an ordering token.  See my comments on your PR for the direction I'd
like you to go.
-Sam

myoungwon:
Right, the ordering lock is need for completion ordering.

We need to clarify the lock design first before start rest of works.
Below scenarios are my understanding based on your comments.

A. when a write request is issued
 1. PG lock (acquire)  -> 2. Ordering lock (RepopManager::lock, acquire) -> 3. add repop_queue -> 4. Ordering lock (release) -> 5. issue_repop -> 6. PG lock (release)

B. when commit, apply are completed
 1. complete commit or apply. -> 2. Repop(OP) lock (acquire) -> 3. If all commit or apply are finished, send ack to client -> 4. Repop(OP) lock (release) -> 5. delay..  -> 6. queue_for_completion (see D.) 
* I think there is no unordered completion here because current finisher (commit, apply) process completed requests in order (per PG).  Also, sending ack is handled in oder because repop_reply can be processed in order via ordering lock.

C. when osd receives MSD_OSD_REPOPREPLY
 1. Ordering lock (RepopManager::lock, acquire) -> 2. Repop(OP) lock (acquire) -> 3. check completion (applied) -> 4. check completion (committed) -> 5   If all commit or apply are finished, send ack to client -> 6. Repop(OP) lock (release) -> 7. Ordering lock (release) -> 8. delay.. -> 9. queue_for_completion (see D.) 
* repop_reply is handled via ordering lock.

D. queue_for_completion (update PG metadata such as lcod)
 1. PG lock (acquire)  -> 2. Ordering lock (RepopManager::lock, acquire) -> 3. do_lazy_completion (iterate repop_queue for lazy completion and remove repop from repop_queue) -> 4. Ordering lock (release) -> 5. PG lock (release)

athanatos:
On Fri, Jul 15, 2016 at 4:14 AM, myoungwon oh <notifications@github.com>
wrote:

> Right, the ordering lock is need for completion ordering.
>
> We need to clarify the lock design first before start rest of works.
> Below scenarios are my understanding based on your comments.
>
> A. when a write request is issued
>
>    1. PG lock (acquire) -> 2. Ordering lock (RepopManager::lock, acquire)
>    -> 3. add repop_queue -> 4. Ordering lock (release) -> 5. issue_repop -> 6.
>    PG lock (release)
>
>
I'd say that the main piece of design awkwardness is that both the Backend
and ReplicatedPG have state that will need to be managed by the Ordering
lock.  I suggest that you handle that letting ReplicatedPG own the odering
lock (RepopManager::lock), update the PGBackend interface such that
submit_transaction is guarranteed to be called under both the PG::lock and
the RepopManager::lock, and add a variant of the blessed_context concept
which locks with RepopManager::lock rather than PG::lock (you'll probably
need a copy of the last_peering_reset epoch under that lock so that you can
atomically skip obsolete callbacks).


>
>
> B. when commit, apply are completed
>
>    1. complete commit or apply. -> 2. Repop(OP) lock (acquire) -> 3. If
>    all commit or apply are finished, send ack to client -> 4. Repop(OP) lock
>    (release) -> 5. delay.. -> 6. queue_for_completion (see D.)
>       - I think there is no unordered completion here because current
>       finisher (commit, apply) process completed requests in order (per PG).
>       Also, sending ack is handled in oder because repop_reply can be processed
>       in order via ordering lock.
>
>
You should be correct about the ordering here.  If you put the right state
under the lock, much of the time, you may not need to queue_for_completion
at all.  See if you can get the common case to not require the PG lock at
all.


>
>    1.
>
> C. when osd receives MSD_OSD_REPOPREPLY
>
>    1. Ordering lock (RepopManager::lock, acquire) -> 2. Repop(OP) lock
>    (acquire) -> 3. check completion (applied) -> 4. check completion
>    (committed) -> 5 If all commit or apply are finished, send ack to client ->
>    6. Repop(OP) lock (release) -> 7. Ordering lock (release) -> 8. delay.. ->
>    9. queue_for_completion (see D.)
>       - repop_reply is handled via ordering lock.
>
>
As long as you queue in the OSD op_wq using a qitem with the right ordering
token and ordering lock, the ordering here should just work.


>
>    1.
>
> D. queue_for_completion (update PG metadata such as lcod)
>
>    1. PG lock (acquire) -> 2. Ordering lock (RepopManager::lock, acquire)
>    -> 3. do_lazy_completion (iterate repop_queue for lazy completion and
>    remove repop from repop_queue) -> 4. Ordering lock (release) -> 5. PG lock
>    (release)
>
>
It may be the the case that you can avoid taking the PG lock at all in the
common case.  Can lcod be moved under the RepopManager?  You'll have to
learn how it's used and see if the users can tolerate having an async copy.
-Sam

myoungwon:
Below variables and functions are protected by the PG lock when common write is handled. I think these can be updated asynchronously because 1.client does not need them 2. They are already written to disk in order through PG metadata. 3. It does not disappear as long as PG is removed as we was previously discussed. If the most current variables are needed, then we just need to call lazy_completion beforehand. For instance, call lazy_completion before calc_trim_to. (but i need to study corner cases..)

I think the data such as lcod, luod can be handled within RepopManager::lock. However, we need to change lock usage where lcod, luod are used.

A. Primary OP
last_update_ondisk,
last_complete_ondisk,
op_applied(),
update_peer_last_complete_ondisk(),
C_UpdateLastRollbackInfoTrimmedToApplied(),
calc_min_last_complete_ondisk(),
publish_stats_to_osd();

B. Replica OP
last_complete_on_disk(),
op_applied(),
C_UpdateLastRollbackInfoTrimmedToApplied()

myoungwon:
I think below scenarios are more simple ways.
We only need the ordering lock in the case of handling MSG_OSD_REPOPEPLY because the PG lock already protect repop_queue.
lazy completion (Primary, Replica Op) can be handled using repop_queue without additional a queue.

A. when a write request is issued (Primary and Replica Op case)
 1. PG lock (acquire) -> 2. add repop_queue -> 3. issue_repop() (sub_op_modify() in the case of Replica Op)  -> 4. PG lock (release)

B. when commit, apply are completed (Primary and Replica Op case)
 1. complete commit or apply. -> 2. Repop(OP) lock (acquire) -> 3. If all commit or apply are finished, send ack to client -> 5. Repop(OP) lock (release) -> 5. delay.. (6 will happen 1 second later in the batching manner)-> 6. queue_for_completion (see D.)

C. when osd receives MSG_OSD_REPOPREPLY (Primary Op case)
 1. Ordering lock (RepopManager::lock, acquire) -> 2. Repop(OP) lock (acquire) -> 3. check completion (applied) -> 4. check completion (committed) -> 5 If all commit or apply are finished, send ack to client -> 6. Repop(OP) lock (release) -> 7. Ordering lock (release) -> 8. delay.. (9 will happen 1 second later in the batching manner) -> 9. queue_for_completion (see D.)

D. queue_for_completion (update PG metadata such as lcod) (Primary and Replica Op case)
 1. PG lock (acquire) -> 2. do_lazy_completion (iterate repop_queue for lazy completion and remove repop from repop_queue) -> 3. PG lock (release)

athanatos:
This could work, and it has the advantage of allowing the ObjectStore
callbacks to contend only on the same op.  I will not, however, be simple.
Here are the things off the top of my head you'll have to handle (not an
exhaustive list, but make sure you have a strategy for these bits at the
very least before you start):
1) You will need to refactor the PGBackend::submit_transaction to allow the
caller to pass in a per-op locker.  Both the Backend and ReplicatedPG will
have per-op state that needs to be protected, and they need to share a
lock.  This will require care to do cleanly.
2) You will need to refactor the Repop and OpContext callback interfaces
(as well as all users) to distinguish callbacks that require the pg lock
from ones that require the per-op lock.
3) You will need to refactor the PGBackend interface to allow the lazy
completion call in ReplicatedPG to notify the PGBackend implementation that
it can discard state associated with that op (we only have the op lock in
the earlier steps, so we couldn't have cleaned this up).  Alternately, you
could do a much larger refactor to allow the Backend to attach private
state to ReplicatedPG's Repop structure -- I'm not sure how well that will
work the ECBackend, however, make certain you have worked that out first.
4) You will need a clean design for the lazy completion callback.  It
doesn't make sense to put it in the OpWQ, so you'll need something new.
You definitely cannot assume that all of the work can be done by one
thread, so you'll need to map pgs onto N threads.  You can either
generalize the existing Timer class, or write a new one (probably write a
new one).  I suggest you implement it as part of a more general PG tick.  I
think there's a low-cost periodic cache/tiering agent operation that you
could pull in as well.
5) You are going to have to update both ReplicatedBackend and ECBackend to
satisfy these interfaces.  I'm not going to accept a PR that only updates
ReplicatedBackend.  ECBackend would also benefit from the performance
boost, and we won't really know whether your design is suitable for
ECBackend unless you update it as well.

There's another significant challenge with the lazy PG state update.  Most
of the state here is ok to update lazily with one exception (at least, that
I've thought of so far): releasing rw locks and requeuing blocked ops.  You
need to be holding the PG lock to do this correctly (the code currently
makes assumptions about atomicity of lock state w.r.t. the PG lock, and you
*need* to be holding the ordering lock for an op -- the pg lock here -- in
order to queue it at the end of the queue without breaking ordering).  In
the common case for pipelined writes on the same object or writes on
distinct objects, it's not urgent to release the lock state.  It's only
urgent when there is a blocked op.  It's important, therefore, to ensure
that the lazy callback happen ASAP if there is a blocked op.

Here's one possible strategy.  Modify ReplicatedPG to maintain a list on
each RWState of in-progress repops holding locks for that object.  While
obtaining locks prior to submitting an op, if the op needs to block on an
RWState lock, the thread should take the lock of the last in-progress Repop
which will release a lock on that RWState and, depending on whether the op
has completed yet, either flip a bit indicating that the lazy callback
should be scheduled ASAP or drop the op lock, perform the async updates
inline (we are already holding the right locks, I think), and try again to
obtain the RWState lock (it should have been released at this point).
You'll have to find a way to do this that doesn't involve sprinkling this
retry logic everywhere something tries to take an RWState lock.  This may
well not be the best way (I hope not, would rather have something simpler).
-Sam

myoungwon:
Thanks for your comment. I think the things you mentioned should be considered first.

1),3): How is using abstract class? Abstract class allow PGBackend to access RepGather as below.

Class AbstractComp {
 Virtual void lock() = 0;
 Virtual void unlock() = 0;
 Virtual void set_lcod() = 0;
};

Class RepGather: public AbstractComp {
Void lock() {..}
Void unlock() {…}
Void set_lcod(…) {…}
};

Struct InProgressOp {
 AbstractComp * ab_comp;
};

void submit_transaction(........, AbstractComp *);

4): I will investigate your suggestions

5): OK. I will apply op completion to ECBackend. However, I think making common interface should be implemented first.

I understand blocked OP you mentioned.
And, I have a question for the RW lock.
Is the RW locks need to be released or acquired within the PG lock ? (because it is also Mutex in ObjectContext)

athanatos:
On Tue, Jul 19, 2016 at 12:14 AM, myoungwon oh <notifications@github.com>
wrote:

> Thanks for your comment. I think the things you mentioned should be
> considered first.
>
> 1),3): How is using abstract class? Abstract class allow PGBackend to
> access RepGather as below.
>
> Class AbstractComp {
> Virtual void lock() = 0;
> Virtual void unlock() = 0;
> Virtual void set_lcod() = 0;
> };
>
> Class RepGather: public AbstractComp {
> Void lock() {..}
> Void unlock() {…}
> Void set_lcod(…) {…}
> };
>
> Struct InProgressOp {
> AbstractComp * ab_comp;
> };
>
> void submit_transaction(........, AbstractComp *);
>
Well, sure, you'll be creating an interface and passing in something that
implements it.  It makes sense that the implementation will actually be a
Repop (perfectly good place to put your per-op state).  lock(), unlock()
makes sense.  I don't understand set_lcod() here.  This also doesn't
address 3 (I don't think?).  Take a look at ReplicatedBackend, there is an
in_progress_ops map that needs to be cleaned up under the PG lock.


> 4): I will investigate your suggestions
>
> 5): OK. I will apply op completion to ECBackend. However, I think making
> common interface should be implemented first.
>
Oh, sure, you probably want to work out the PGBackend interface changes
first, but think about how they apply the ECBackend as you do it, and the
changes to ECBackend will need to be done as part of the same PR.

myoungwon:
Just ignore set_lcod(). this is my mistake. 
I think an in_progress_ops map can be handled by another lock domain, right? I will investigate an in_progress_ops map with RepopManager::lock.

athanatos:
> Just ignore set_lcod(). this is my mistake. 
> I think an in_progress_ops map can be handled by another lock domain (in_progress_ops_lock in the submitted code) as long as OPs can be processed in order (issue and completion cases), right? For instance,
> 
> a. Issue
> 
> PG lock (acquire) -> 2. issue_repop -> 3. in_progress_ops_lock (acquire) -> 4. insert InProgressOP to an in_progress_ops map -> 5. in_progress_op_lock (release) -> 6. PG lock (release) -> .....

Ok, but that means that the callbacks can contend on that lock.  If we are going to allow that, there are simpler ways to do this.

> b. Completion (MSG_OSD_REPOPREPLY)
> 
> Ordering lock (RepopManager::lock, acquire) -> 2. in_progress_ops_lock (acquire) -> in_progress_ops.find() -> 3. in_progress_ops_lock (release) -> 4. Repop(OP) lock (acquire) -> .......

Ok, like I said, this means that the OpReply handling can contend with op submission.  That would be ok with me, but it would make the entire async callback concept pointless -- I'd insist you instead use the ordering lock in ReplicatedPG, and figure out how to update the state that needs to be updated without a separate PG lock callback.

> We can handle MSG_OSD_REPOPREPLY with no PG lock.
> However, this is not simple... I need to investigate the simple way for InProgressOP lock.
> Do you have any idea?

I think you aren't thinking this out fully.  The whole point of your design is to take only per-op locks in the filestore callbacks.  That means that adding a lock to protect the op_map which needs to be taken in those callbacks should be avoided.  You have the PG lock in the async callback, do it there.

myoungwon:
- in_progress_ops_lock 

However, InProgressOP (ReplicatedBackend), Op (ECBackend) are needed in order to check commit and apply states when OSD receives MSG_OSD_REOPREPLY. We can not avoid this lookup operations. Therefore, I think we should focus on how to minimize this critical section. in_progress_ops_lock, i mentioned above, is only used in the case of insert and find operations. For instance,

a. insert
`in_progress_ops_lock.Lock();`
`InProgressOp &op = in_progress_ops.insert(
      make_pair(
        ......
        orig_op, at_version)
        )
      ).first->second;`
`in_progress_ops_lock.Unlock();`

b. find
`in_progress_ops_lock.Lock();`
 `   if (in_progress_ops.count(rep_tid)) {`
     ` map<ceph_tid_t, InProgressOp>::iterator iter = in_progress_ops.find(rep_tid);`
` in_progress_ops_lock.Unlock();`





- common lock interface which can protect internal ops lists.


There is the operation lists in ReplicatedBackend (in_progress_ops) and ECBackend (writing, tid_to_op_map). We need to protect these structures because the PG lock is remove when op completion is handled. Therefore, I think adding ops_ordering_lock(), ops_ordering_unlock() is need in common PGBackend interface in order to protect internal ops lists. What do you think about these interfaces?

athanatos:
> in_progress_ops_lock
> However, InProgressOP (ReplicatedBackend), Op (ECBackend) are needed in order to check commit and apply states when OSD receives MSG_OSD_REOPREPLY. We can not avoid this lookup operations. Therefore, I think we should focus on how to minimize this critical section. in_progress_ops_lock, i mentioned above, is only used in the case of insert and find operations. For instance,

You need the InProgressOp, *not* the in_progress_ops map.  You could attach a pointer to the InProgressOp to the callback and access it directly without doing a lookup under the Repop lock (which is per-op).

> 
> a. insert
> in_progress_ops_lock.Lock();
> InProgressOp &op = in_progress_ops.insert( make_pair( ...... orig_op, at_version) ) ).first->second;
> in_progress_ops_lock.Unlock();
> 
> b. find
> in_progress_ops_lock.Lock();
> if (in_progress_ops.count(rep_tid)) {
> map<ceph_tid_t, InProgressOp>::iterator iter = in_progress_ops.find(rep_tid);
> in_progress_ops_lock.Unlock();
> 
> common lock interface which can protect internal ops lists.
> There is the operation lists in ReplicatedBackend (in_progress_ops) and ECBackend (writing, tid_to_op_map). We need to protect these structures because the PG lock is remove when op completion is handled. Therefore, I think adding ops_ordering_lock(), ops_ordering_unlock() is need in common PGBackend interface in order to protect internal ops lists. What do you think about these interfaces?

If you choose to go that route, that's fine, but at that point, I claim the asyncronous step can be avoided entirely by carefully placing the state that would be updated under the ops_ordering_lock instead.  Either way, please write something in doc/dev/osd_internals explaining what you plan to do so that we can review the design first.

myoungwon:
How do we know which the InProgressOP is needed when OSD receives MSG_OP_REPOPREPLY without the in_progress_ops map? We should find out the InProgressOP first through rep_tid.

athanatos:
Presumably because you included the right pointer or iterator in the calback -- std::map has stable references.

myoungwon:
@athanatos I updated a commit (https://github.com/athanatos/ceph/pull/4/commits/d144cc047cad1caacf460d643e60946358fff971) as your comments (support ECBackend, replay ordering..). Can you take a look? I want to discuss lock design.

athanatos:
@myoungwon Which of these PRs is the real one?  I don't intend to actually maintain my wip-osd-reply branch... I just pushed it for you to use.  I think you should probably close that PR and update this one with whatever you want to merge.

athanatos:
@myoungwon At first glance, I have two problems:
1) If you are going to take a pg-wide lock at all, I want it to protect the repop queue, and I want you to eliminate the asyncronous completion code entirely.
2) Since 1) needs to protect the repop queue, I don't want *Backend to own the lock, I think ReplicatedPG should own the lock.

myoungwon:
@athanatos I think *Backend lock can be replaced by repop queue lock as below. However i don't understand that i can either take a lock in the completion path, or i can have an async completion. Could you give me a concrete example? (async code is need because it handle jobs which need the PG lock in the batching manner)

*Write Flow*
 A. when a write request is issued (Primary)
 
     1. PG lock (acquire) -> 2. do_lazy_completion(check completed operation) -> 
       3. repop queue lock ->  4. add repop_queue -> 
       -> 5. insert InProgressOP (ReplicateBackend) or Op (ECBackend) -> 
       6. repop queue lock (release) -> 7. queue_transaction() -> 8. PG lock (release)
 
 B. when commit, apply are completed
 
     1. complete commit or apply. -> 2. Repop(OP) lock (acquire) -> 
       3. If all commit or apply are finished, send ack to client -> 
       4. Repop(OP) lock (release) -> 
       5. delay.. (6 will happen a few second later in the batching manner) -> 
       6. queue_for_completion (see D.)
 
 C. when osd receives MSG_OSD_REPOPREPLY or MSG_OSD_EC_WRITE_REPLY
 
     1. Reply Ordering lock (ReplicatedPG::seq_ops_lock, acquire) -> 
       2. repop queue lock  -> 
       3. Get InProgressOp or Op -> 4. repop queue lock  (release) -> 
       5. Repop(OP) lock (acquire) -> 6. check completion (applied) -> 
       7. check completion (committed) -> 
       8. If all commit or apply are finished, send ack to client -> 
       9. Repop(OP) lock (release) -> 10. Reply Ordering lock (release) -> 
       11. delay..(12 will happen a few second later in the batching manner) -> 
       12. queue_for_completion (see D.)
 
 D. queue_for_completion (OSDService::comp_worker, update PG metadatas such as lcod)
 
     1. comp worker wake up 2. PG lock (acquire) -> 2. do_lazy_completion() 
      (iterate repop_queue and repop_sub_queue for lazy completion and remove repop from                    
       repop_queue) -> 3. PG lock (release)


However,asynchronous code is needed because it

athanatos:
I don't see why you need to take a pg-wide lock at all in the local ObjectStore callback path (you do need one to process messages to preserve ordering).  It seems to me that the only thing you really need to do in the local callback path is (possibly) send the response to the client.  That part only requires a per-op lock.  If you are going to do async work later to complete the operation while holding the pg lock, why not cleanup the repop and backend state there?

Conversely, if you *are* going to take a pg-wide lock (a repop lock, for instance), why can't you move the state you need to update down into that lock and do it inline?

Basically, I really don't like the complexity of the async work queue, and I really don't like the complexity (or overhead) of an additional lock.  We obviously need one or the other to avoid the performance problem you have outlined, but I'm not convinced that we have to have both.  You'd have to convince me that there isn't a simpler way.

myoungwon:
I understood.
Main purpose of this patch is to resolve coarse-grained lock contention via fine-grained lock.
Therefore, different lock domains are need when OSDs process operations (when OP is issued, when commit is completed, when apply is completed, when OSD receives reply from replica) which required the PG lock in order to enhance parallelism.
In particular,  per-op lock in local ObjectStore callback path (commit, apply) is important because it can minimize contention from finishers and the PG lock (note that this is for high IOPS).
My plan is that counting commit, apply and sending ack to client are processed within OP lock first because this is critical path for performance and then completion worker call lazy completion for update information which required the PG lock. Therefore, issue and completion can be processed with different the lock and minimum critical section.
However, inevitable ordering (for reply) and finding OP cause additional locks and data structures.

Anyway, if you worried about complexity from my plan, i think current patch can be changed as below with the OP lock scheme. (operations which need the PG lock in completion path can be move into repop_queue lock without async completion as your comment)
Do i understand correctly?

*Write Flow*
 A. when a write request is issued (Primary)
 
     1. PG lock (acquire) ->   
       2. repop queue lock ->  3. add repop_queue -> 
       -> 4. insert InProgressOP (ReplicateBackend) or Op (ECBackend) -> 
       5. repop queue lock (release) -> 6. queue_transaction() -> 7. PG lock (release)
 
 B. when commit, apply are completed
 
     1. complete commit or apply. -> 2. Repop(OP) lock (acquire) -> 
       3. If all commit or apply are finished, send ack to client -> 
       4. Repop(OP) lock (release) -> 
       5. Update PG state with the repop queue lock (occur rarely)
 
 C. when osd receives MSG_OSD_REPOPREPLY or MSG_OSD_EC_WRITE_REPLY 
         I think reply ordering lock can be removed...

     1. Reply Ordering lock (ReplicatedPG::seq_ops_lock, acquire) -> 
       2. repop queue lock  -> 
       3. Repop(OP) lock (acquire) -> 4. check completion (applied or commit) -> 
       5. If all commit or apply are finished, send ack to client -> 
       6. Repop(OP) lock (release) -> 7. Update PG state with the repop queue lock ->
       8. repop queue lock (release) -> 10. Reply Ordering lock (release)

athanatos:
Yeah, you can certainly use the repop queue lock for ordering replies, no problem there.  "Update PG state with the repop queue lock (occur rarely)" is indeed what I was hoping for, but it's going to be tricky.  First: how exactly will you requeue ops blocked on the write you are completing?  See ObjectContext::RWState.  I think that's the single hardest question.

myoungwon:
I think requeue_ops() in release_object_locks() can work because it just call queue_op_front(). The problem is ObcLockManager,get_lock() and ObcLockManager.put_lock() (put_lock() return blocked list) because these function is protected by the PG lock (this patch does not grab the PG lock in completion path). right?
One of simple solution is  ObcLockManager or ObjectContext  has own the lock. I don't know why ObcLockManager is used within the PG lock because ObcLockManager and ObjectContext are for Object not PG.
Is anything I'm missing?

athanatos:
I'm afraid not, take a look at how the dequeueing works.  In order to ensure that ops aren't reordered, you need to be holding the pg lock when you call queue_op_front().

myoungwon:
The PG lock is needed if we need the ordering (and as long as we use the PG lock at issue context). Therefore,
I think requeue_ops with the PG lock is imperative if ObcLockManager has blocked requests. However, we can skip the PG lock if ObcLockManager does not have blocked requests.
How was below idea?

release_object_locks()
ObcLockManager.lock() --> put_locks(get blocked list) --> ObcLockManager.unlock() --> if (blocked list) --> PG lock --> requeue_ops -> PG unlock

athanatos:
Something like that could work, but you need to be holding the PG lock when you actually put the locks from the manager since that needs to be atomic with the requeue.

I'll point out that this problem goes away if you instead do work with the pg lock asynchronously.

myoungwon:
Async work need async completion code. It cause the complexity as you comment. please look at below.

 A. when a write request is issued (Primary)

   1. PG lock (acquire) -> 2. repop queue lock ->  3. add repop_queue -> 4. insert InProgressOP 
(ReplicateBackend) or Op (ECBackend) -> 5. repop queue lock (release) -> 6. queue_transaction() -> 7. PG lock (release)
 
B. when commit, apply are completed

   1. complete commit or apply. -> 2. Repop(OP) lock (acquire) -> 3. If all commit or apply are finished, send ack to client -> 4. Repop(OP) lock (release) -> 5. queue_for_completion
 
C. when osd receives MSG_OSD_REPOPREPLY or MSG_OSD_EC_WRITE_REPLY 
(some lock which protect in_progress_ops is needed at here in order to find out stable InProgressOP.
If we does not use the lock, segfault occur due to wrong reference even though we use unique tid and only need read access. (this is because internal data structure in std:map is changed when OSD inserts many items fast)

   1. repop queue lock -> 2. find out OP -> 3. Repop(OP) lock (acquire) -> 4. check completion (applied or commit) -> 5. If all commit or apply are finished, send ack to client -> 6. Repop(OP) lock (release) -> 7. queue_for_completion -> 8. repop queue lock (release) 

 
D. queue_for_completion (OSDService::comp_worker, update PG metadatas such as lcod)

   1. comp worker wake up 2. PG lock (acquire) -> 2. do_lazy_completion() 
      (iterate repop_queue and repop_sub_queue for lazy completion and remove repop from                    
       repop_queue) -> 3. PG lock (release)

myoungwon:
That could work. We discussed the completion path based on repop_queue lock without the async completion. However, you think that async completion is needed because requeue_ops problem as you mentioned, right?

Async work need async completion code. It cause the complexity as you comment. please look at below and give me comments.

 A. when a write request is issued (Primary)

   1. PG lock (acquire) -> 2. repop queue lock ->  3. add repop_queue -> 4. insert InProgressOP 
(ReplicateBackend) or Op (ECBackend) -> 5. repop queue lock (release) -> 6. queue_transaction() -> 7. PG lock (release)
 
B. when commit, apply are completed

   1. complete commit or apply. -> 2. Repop(OP) lock (acquire) -> 3. If all commit or apply are finished, send ack to client -> 4. Repop(OP) lock (release) -> 5. queue_for_completion
 
C. when osd receives MSG_OSD_REPOPREPLY or MSG_OSD_EC_WRITE_REPLY 
(some lock which protects in_progress_ops is needed at here in order to find out stable InProgressOP.
If we does not use the lock, segfault occur due to wrong reference even though we use unique tid and only need read access. (this is because internal data structure in std:map is changed when OSD inserts many items fast)

   1. repop queue lock -> 2. find out OP -> 3. Repop(OP) lock (acquire) -> 4. check completion (applied or commit) -> 5. If all commit or apply are finished, send ack to client -> 6. Repop(OP) lock (release) -> 7. queue_for_completion -> 8. repop queue lock (release) 

 
D. queue_for_completion (OSDService::comp_worker, update PG metadatas such as lcod)

   1. comp worker wake up 2. PG lock (acquire) -> 2. do_lazy_completion() 
      (iterate repop_queue and repop_sub_queue for lazy completion and remove repop from                    
       repop_queue) -> 3. PG lock (release)

athanatos:
That seems like it could work.  How do you plan on implementing the async part?

myoungwon:
I already made async part in the latest commit which pushed your repository. (https://github.com/athanatos/ceph/pull/4/commits/4eb80626f662bf0d60548b52aee0cb2e9ad227ae) 


Please see OSDService::CompletionWorker.
CompletionWorker has SafeTimers. Context callback is delivered to CompletionWorker in order to check completion.

For instance,
osd->_queue_for_completion(Context*...) ---> CompletionWorker::timer[pg_num%timer_num]->add_event_after --->wake up ---> do_lazy_completion (with the PG lock)

athanatos:
Right, I think you should be able to use the existing queue, I don't want new threads here.

myoungwon:
"new threads" means SafeTimer, right? Actually, CompletionWorker is not thread just data structure for SafeTimer and Timer event will be registered each 1 second per PG (configurable)

"existing queue" means SharededWQ, right? Queueing completon items to SharededWQ can be answer.
However, Delayed batching processing is more helpful if PG does not have blocked requests because we do not need the PG lock at completion path. Therefore, SafeTimer is introduced for batching processing.
How can i enqueue items to SharedeWQ with delay?

athanatos:
I don't think you actually want a delay.
-Sam

myoungwon:
Yes, batch update causes less lock contention and increase ShardedOpWQ thread utilization.
However, if you don't want new threads, using existing queues is OK.

myoungwon:
@athanatos  Right.  Please look at below.  This is about how async completion will be  which is used existing queues (ShardedOpWQ) without delay.

A. when a write request is issued (Primary)

   1. PG lock (acquire) -> 2. repop queue lock -> 3. add repop_queue -> 4. insert InProgressOP (ReplicateBackend) or Op (ECBackend) -> 5. repop queue lock (release) -> 6. queue_transaction() -> 7. PG lock (release)

B. when commit, apply are completed

   1. complete commit or apply. -> 2. Repop(OP) lock (acquire) -> 3. If all commit or apply are finished, send ack to client -> 4. Repop(OP) lock (release) -> 5. queue_for_completion

C. when osd receives MSG_OSD_REPOPREPLY or MSG_OSD_EC_WRITE_REPLY 

   1. repop queue lock -> 2. find out OP -> 3. Repop(OP) lock (acquire) -> 4. check completion (applied or commit) -> 5. If all commit or apply are finished, send ack to client -> 6. Repop(OP) lock (release) -> 7. queue_for_completion -> 8. repop queue lock (release)

D. queue_for_completion (using  ShardedOpWQ threads)

   1. PG lock (acquire) -> 2. do_completion() (iterate repop_queue and repop_sub_queue for completion and remove repop from
repop_queue) -> 3. PG lock (release)

athanatos:
I actually think you're going to want to call do_completion() whenever the pg lock is taken, in addition to when the queued event happens.

myoungwon:
Right. I missed do_completion() in the A case. please look at below (case B is added). 

A. when a write request is issued (Primary)

   1. PG lock (acquire) -> **2. do_completion()** -> 3. repop queue lock -> 4. add repop_queue -> 5. insert InProgressOP (ReplicateBackend) or Op (ECBackend) -> 6. repop queue lock (release) -> 7. queue_transaction() -> 8. PG lock (release)

B. when a write request is issued (Not primary)
 
  1. PG lock (acquire) -> **2. do_completion()** -> 3. add repop_sub_queue -> 4. insert RepModify 
       (ReplicateBackend) or Op (ECBackend) -> 5. queue_transaction() -> 6. PG lock 
       (release)

C. when commit, apply are completed

   1. complete commit or apply. -> 2. Repop(OP) lock (acquire) -> 3. If all commit or apply are finished, send ack to client -> 4. Repop(OP) lock (release) -> 5. queue_for_completion

D. when osd receives MSG_OSD_REPOPREPLY or MSG_OSD_EC_WRITE_REPLY 

   1. repop queue lock -> 2. find out OP -> 3. Repop(OP) lock (acquire) -> 4. check completion (applied or commit) -> 5. If all commit or apply are finished, send ack to client -> 6. Repop(OP) lock (release) -> 7. queue_for_completion -> 8. repop queue lock (release)

**E. queue_for_completion (using  ShardedOpWQ threads)**

   1. PG lock (acquire) -> 2. do_completion() (iterate repop_queue and repop_sub_queue for completion and remove repop from
repop_queue) -> 3. PG lock (release)

athanatos:
This could work, but it's going to be complicated.

myoungwon:
Could you give me the suggestion or comment to solve complicated?

athanatos:
I mean, your approach seems sound, but do what you can to make it clear to future readers how it's supposed to work, and to make it clear to reviewers that you aren't introducing race conditions or ordering bugs.

myoungwon:
It seems that i need to write more detailed documentation  in order to meet your requirement (in doc/dev/osd_internals).
How is updating codes and document based on previous discussion, and then discuss again?

athanatos:
@myoungwon I mean, documentation is always good and I do want a .rst file in doc/dev/osd_internals to explain how it works, but your original series iirc, *duplicated* a whole bunch of logic and didn't make it clear how the new code fit in with the old code.  It also introduced a bunch of subtle ordering bugs.  I want you to really think through all of the implications and choose an implementation strategy that really makes it clear what you are doing and why it works.  I'll be happy to continue to review your design and code as you go though!

myoungwon:
Sorry, I don't understand fully what you say. 

How can I make implementation strategy clear? Do I need to explain the I/O logic (with data structures, interfaces) in the case of  recovercy, srub, basic I/O and so on? Or do I need to re-design completely ?

athanatos:
Disregard my earlier comment.  Go ahead and try to implement the strategy you outlined above, and I'll give you feedback.

myoungwon:
OK. I'll update code and document based on your comments and discussion at here.

myoungwon:
@athanatos I created a new PR (#11188) based on previous discussion. please take a look.

athanatos:
@myoungwon Ok, I'm closing this one then.

PR #9287, opened 2016-05-24
compressor: optimize header file dependency

cxwshawn:
Signed-off-by: Xiaowei Chen chen.xiaowei@h3c.com


src/compressor/Compressor.h
```diff
@@ -16,7 +16,10 @@
 #define CEPH_COMPRESSOR_H
 
 #include "include/int_types.h"
-#include "include/Context.h"
+#include "include/memory.h"
+#include "include/buffer.h"
+#include <string>
+using namespace std;
```
tchaikov: maybe better off specifying the `std` namespace in this header file instead of `using namespace std` here, to avoid the namespace pollution.

src/compressor/Compressor.h
```diff
@@ -16,7 +16,10 @@
 #define CEPH_COMPRESSOR_H
 
 #include "include/int_types.h"
-#include "include/Context.h"
+#include "include/memory.h"
+#include "include/buffer.h"
```
tchaikov: i'd suggest use `#include "include/buffer_fwd.h"` here

src/compressor/Compressor.h
```diff
@@ -16,7 +16,10 @@
 #define CEPH_COMPRESSOR_H
 
 #include "include/int_types.h"
-#include "include/Context.h"
+#include "include/memory.h"
```
tchaikov: should use `<memory>` here instead. it's portable. we added our own `memory.h` back in the pre-C++11 era, when the smart pointers were still in TR1.

src/compressor/Compressor.h
cxwshawn: @tchaikov since u know SnappyCompressor extends Compressor implemented in SnappyCompressor.h header file totally, so I think directly including "include/buffer.h" is ok.

tchaikov: `Compressor.h` is included by other files also. by using `buffer_fwd.h` speed up the compilation, i don't see it has anything to do with `SnappyCompressor.h`. maybe i am missing anything?

cxwshawn: @tchaikov what I mean is if we do so , we should use include/buffer.h other than buffer_fwd.h in SnappyCompressor.h, there will be  both class forward and real buffer definition in SnappyCompressor.h

tchaikov: >  there will be both class forward and real buffer definition in SnappyCompressor.h

i think it is fine.

src/compressor/Compressor.h
```diff
@@ -16,19 +16,21 @@
 #define CEPH_COMPRESSOR_H
 
 #include "include/int_types.h"
-#include "include/Context.h"
+#include <memory>
+#include "include/buffer_fwd.h"
+#include <string>
```
tchaikov: nit, could group the include so it looks like
```c++
#include "include/buffer_fwd.h"
#include "include/int_types.h"

#include <memory>
#include <string>
```

src/compressor/CompressionPlugin.h
```diff
@@ -18,9 +18,9 @@
 #ifndef COMPRESSION_PLUGIN_H
 #define COMPRESSION_PLUGIN_H
 
-#include "common/Mutex.h"
 #include "common/PluginRegistry.h"
 #include "Compressor.h"
+#include <ostream>
```
tchaikov: could use `<iosfwd>` instead

cxwshawn:
@tchaikov  :+1: ,  very thankful for your advice, fixed, :)

cxwshawn:
@tchaikov @liewegas  PING! please help review again  !

badone:
@cxwshawn I'd like to reopen this PR and to help get this in. Alternatively, I would be happy to take over this work as I am doing similar things at the moment? Please let me know.

badone:
Closing. See https://github.com/ceph/ceph/pull/15187

PR #9288, opened 2016-05-24
osd: print log when osd want to kill self

yuyuyu101:
Signed-off-by: Haomai Wang haomai@xsky.com

### DIFF
--- a/src/osd/OSD.cc
+++ b/src/osd/OSD.cc
@@ -6905,12 +6905,12 @@ void OSD::_committed_osd_maps(epoch_t first, epoch_t last, MOSDMap *m)
 	       osd_markdown_log.front() + grace < now)
 	  osd_markdown_log.pop_front();
 	if ((int)osd_markdown_log.size() > g_conf->osd_max_markdown_count) {
-	  dout(10) << __func__ << " marked down "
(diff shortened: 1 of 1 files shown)


yuyuyu101:
we want to know the reason make decision to shutdown, rebind failed or too much failure retries

src/osd/OSD.cc
```diff
@@ -6926,18 +6926,21 @@ void OSD::_committed_osd_maps(epoch_t first, epoch_t last, MOSDMap *m)
 	if (r != 0) {
 	  do_shutdown = true;  // FIXME: do_restart?
           network_error = true;
+          dout(0) << __func__ << " marked down: rebind failed" << dendl;
```
Aran85: Maybe print the class of messager will be better?

yuyuyu101: for qa, I think developers could know the messenger type from job describe. for user, it should definitely be one type defined in ceph.conf

tchaikov: i think it's aways `MOSDMap`?

yuyuyu101: yes, message type should always be MOSDMap

liewegas:
retest this please

PR #9289, opened 2016-05-24
rgw/rados: initialize variable 'rados[]' to zero

Yan-waller:
We should memset `rados[]` to zero before using it, since
it may be a unexpected pointer value while jumping to the `fail:` label
to do the `delete rados[i]` operation.

in addition, we should also release `cr_registry`.

### DIFF
--- a/src/rgw/rgw_rados.cc
+++ b/src/rgw/rgw_rados.cc
@@ -3199,7 +3199,7 @@ void RGWRados::finalize()
 int RGWRados::init_rados()
 {
   int ret = 0;
-
+  uint32_t i;
   num_rados_handles = cct->_conf->rgw_num_rados_handles;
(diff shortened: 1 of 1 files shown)


cbodley:
this reminded me of an old branch that switched `rados[]` to a `std::vector`, which makes the error handling much simpler. i rebased that, added your fix for `cr_registry`, and opened https://github.com/ceph/ceph/pull/9295 - could you take a look?

Yan-waller:
retest this please

Yan-waller:
retest this please

mattbenjamin:
@Yan-waller @cbodley I would also favor moving to std::vector here

mattbenjamin:
close because superceded by 9295

PR #9290, opened 2016-05-24
tools: do not closed stdout ; fix overload of "<" operator

xiexingguo:


xiexingguo:
@liewegas :-)

liewegas:
@xiexingguo did you mean to include the osd patches in this pull request?

liewegas:
the last two patches look good.  note that i don't think teh radosacl code is used by anything?

xiexingguo:
> did you mean to include the osd patches in this pull request?

yep. Would you like me to split them into another PR?

xiexingguo:
> note that i don't think teh radosacl code is used by anything?

Yeah. Then shall we kill the file instead?

liewegas:
please split.  thanks!
>

liewegas:
let's put it on the cleanup list, but wait until we can check with @yehudasa

yuriw:

PR #9291, opened 2016-05-24
rbd-nbd does not properly handle resize notifications

trociny:


trociny:
@dillaman This is still WIP, I just want to know your opinion about the approach in general and API.

Right now it does not work though: I want to be able to call `image.stat()` inside the registered watch handler and this causes deadlock:

```
Thread 7 (Thread 0x7fffe8328700 (LWP 7806)):
#0  pthread_cond_wait@@GLIBC_2.3.2 () at ../nptl/sysdeps/unix/sysv/linux/x86_64/pthread_cond_wait.S:185
#1  0x00005555559b43b5 in Wait (mutex=..., this=0x7fffe8325750) at ./common/Cond.h:56
#2  C_SaferCond::wait (this=this@entry=0x7fffe83256f0) at ./common/Cond.h:202
#3  0x0000555555c89318 in librbd::ImageState<librbd::ImageCtx>::refresh_if_required (this=0x55555f41c740) at librbd/ImageState.cc:232
#4  0x0000555555c9b89f in librbd::info (ictx=ictx@entry=0x55555f419be0, info=..., infosize=infosize@entry=160) at librbd/internal.cc:1572
#5  0x0000555555c4ef80 in rbd_stat (image=0x55555f419be0, info=0x7fffffffde18, infosize=160) at librbd/librbd.cc:1958
#6  0x0000555555b4aef5 in handle_notify (this=<optimized out>) at test/librbd/test_librbd.cc:493
#7  TestLibRBD_StatWatchAndResize_Test::Watcher::cb (cb=<optimized out>, arg=<optimized out>) at test/librbd/test_librbd.cc:487
#8  0x0000555555c8a549 in notify (this=0x55555f41d900) at librbd/ImageState.cc:81
#9  librbd::ImageState<librbd::ImageCtx>::handle_update_notification (this=0x55555f41c740) at librbd/ImageState.cc:181
#10 0x0000555555c7dfa8 in librbd::ImageCtx::notify_update (this=0x55555f419be0, on_finish=on_finish@entry=0x55555f41dd40) at librbd/ImageCtx.cc:1029
#11 0x0000555555cca9b8 in librbd::(anonymous namespace)::C_NotifyUpdate<librbd::ImageCtx>::complete (this=0x55555f41dd40, r=0) at librbd/Operations.cc:67
#12 0x00005555559677e9 in Context::complete (this=0x7fffd00013e0, r=<optimized out>) at ./include/Context.h:64
#13 0x0000555555b07090 in librbd::util::detail::rados_state_callback<librbd::operation::ResizeRequest<librbd::ImageCtx>, &librbd::operation::ResizeRequest<librbd::ImageCtx>::handle_update_header, true> (c=<optimized out>, arg=0x55555f41e410) at ./librbd/Utils.h:39
#14 0x0000555555d75b1e in librados::finish_aio_completion (c=0x7fffd8002690, r=0) at test/librados_test_stub/TestRadosClient.cc:47
#15 0x0000555555a3dfbc in operator() (a0=<optimized out>, this=<optimized out>) at /usr/include/boost/function/function_template.hpp:767
#16 FunctionContext::finish (this=<optimized out>, r=<optimized out>) at ./include/Context.h:460
#17 0x00005555559677e9 in Context::complete (this=0x7fffdc000a70, r=<optimized out>) at ./include/Context.h:64
#18 0x0000555555de459e in Finisher::finisher_thread_entry (this=0x55555f3f1d90) at common/Finisher.cc:68
#19 0x00007fffee41b0a4 in start_thread (arg=0x7fffe8328700) at pthread_create.c:309
#20 0x00007fffec76004d in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:111

Thread 1 (Thread 0x7ffff7fc16c0 (LWP 7793)):
#0  pthread_cond_wait@@GLIBC_2.3.2 () at ../nptl/sysdeps/unix/sysv/linux/x86_64/pthread_cond_wait.S:185
#1  0x00005555559b43b5 in Wait (mutex=..., this=0x7fffffffd820) at ./common/Cond.h:56
#2  C_SaferCond::wait (this=this@entry=0x7fffffffd7c0) at ./common/Cond.h:202
#3  0x0000555555cd081b in librbd::Operations<librbd::ImageCtx>::invoke_async_request(std::string const&, bool, boost::function<void (Context*)> const&, boost::function<void (Context*)> const&) (this=this@entry=0x55555f3f9340, request_type="resize", permit_snapshot=permit_snapshot@entry=false, local_request=..., remote_request=...) at librbd/Operations.cc:1070
#4  0x0000555555cd1d9b in librbd::Operations<librbd::ImageCtx>::resize (this=0x55555f3f9340, size=size@entry=8388608, prog_ctx=...) at librbd/Operations.cc:510
#5  0x0000555555c4ec4f in rbd_resize (image=0x55555f419be0, size=8388608) at librbd/librbd.cc:1937
#6  0x0000555555b862b1 in TestLibRBD_StatWatchAndResize_Test::TestBody (this=<optimized out>) at test/librbd/test_librbd.cc:506
#7  0x0000555555dd8e4c in testing::internal::HandleSehExceptionsInMethodIfSupported<testing::Test, void> (object=0x55555f419730, method=&virtual testing::Test::TestBody(), 
    location=0x555555fb131b "the test body") at ./src/gtest.cc:2078
#8  0x0000555555dd4236 in testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void> (object=0x55555f419730, method=&virtual testing::Test::TestBody(), 
    location=0x555555fb131b "the test body") at ./src/gtest.cc:2114
#9  0x0000555555dbb77f in testing::Test::Run (this=0x55555f419730) at ./src/gtest.cc:2151
#10 0x0000555555dbbf83 in testing::TestInfo::Run (this=0x55555f3d6a10) at ./src/gtest.cc:2326
#11 0x0000555555dbc646 in testing::TestCase::Run (this=0x55555f3d5fa0) at ./src/gtest.cc:2444
#12 0x0000555555dc3171 in testing::internal::UnitTestImpl::RunAllTests (this=0x55555f3c2000) at ./src/gtest.cc:4315
#13 0x0000555555dda24a in testing::internal::HandleSehExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool> (object=0x55555f3c2000, 
    method=(bool (testing::internal::UnitTestImpl::*)(testing::internal::UnitTestImpl * const)) 0x555555dc2ed4 <testing::internal::UnitTestImpl::RunAllTests()>, 
    location=0x555555fb1b70 "auxiliary test code (environments or event listeners)") at ./src/gtest.cc:2078
#14 0x0000555555dd4fee in testing::internal::HandleExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool> (object=0x55555f3c2000, 
    method=(bool (testing::internal::UnitTestImpl::*)(testing::internal::UnitTestImpl * const)) 0x555555dc2ed4 <testing::internal::UnitTestImpl::RunAllTests()>, 
    location=0x555555fb1b70 "auxiliary test code (environments or event listeners)") at ./src/gtest.cc:2114
#15 0x0000555555dc1d41 in testing::UnitTest::Run (this=0x5555563b7a80 <testing::UnitTest::GetInstance()::instance>) at ./src/gtest.cc:3929
#16 0x0000555555959a36 in RUN_ALL_TESTS () at ../src/gmock/gtest/include/gtest/gtest.h:2288
#17 main (argc=1, argv=<optimized out>) at test/librbd/test_main.cc:53
```

I hope this is fixable, but it would be nice to know your opinion before spending time on resolving this.

src/librbd/librbd.cc
```diff
@@ -2753,6 +2779,26 @@ extern "C" int rbd_mirror_image_get_status(rbd_image_t image,
   return 0;
 }
 
+extern "C" int rbd_stat_watch(rbd_image_t image, uint64_t *handle,
+			      rbd_callback_t watch_cb, void *arg)
+{
+  librbd::ImageCtx *ictx = (librbd::ImageCtx *)image;
+  C_StatWatchCB *wctx = new C_StatWatchCB(watch_cb, arg);
```
dillaman: Who deletes this object?

dillaman:
How about running the notification callbacks from a clean thread context by adding the "notify" to ```ImageCtx::op_work_queue```?

src/librbd/librbd.cc
trociny: Ah, it is a missing bit yet. I will fix it. Thanks. Initially I thought the watch context would be deleted automatically on unregister, but later decided that it is up to caller to delete it but forgot to update this part.

trociny:
@dillaman I will try `ImageCtx::op_work_queue`, thanks.

Still, how do you like API?
Do `rbd_stat_watch`, `rbd_stat_unwatch` sound right for you?
May be some useful data could be passed to the callback?
Also, right now I am reusing `rbd_callback_t`, instead of providing its own `rbd_stat_watch_callback_t`. `rbd_callback_t`  has `rbd_completion_t` argument, which is not used by the watcher. And actually I am not sure how `rbd_completion_t` is used in aio interface, I have failed to find examples.

dillaman:
@trociny I might suggest not naming the callback after 'stat' since it will be invoked for things that change outside of stat as well (e.g. snapshot changes, metadata changes).  Maybe just ```rbd_update_watch``` and ```rbd_update_unwatch```?

I would create a new callback type with no expected arguments.  The ```rbd_callback_t``` is associated with a ```rbd_completion_t``` when you create the completion via ```rbd_aio_create_completion```.

trociny:
@dillaman Yet another question. I suppose it will be possible when a new update notification comes while the previous is still in progress (processed by the watcher handler). I have choice here: ether drop this notification for this handler or queue it. What do you thinks is better?

dillaman:
I think you need to queue it up and notify again to avoid a race where a change is missed.

trociny:
@dillaman I tried running the notification callbacks using ImageCtx::op_work_queue (see the updated patch) it still does not work for me, if I want to execute info.stat() in a callback.

As I see, the problem is that info() (in notify callback, executed by the worker) calls refresh_if_required(), which queues the request (to be completed by the same worker) and waits for it to complete, so this gets stuck.

Solutions that I see:
- Forbid (by documenting) execution of Image functions inside update callback (not very convenient for users though, as would need to create/get an additional context to do the real job).
- In ImageUpdateWatchers create/use yet another queue for notification callbacks instead of ImageCtx::op_work_queue.

What do you think?

src/include/rbd/librbd.hpp
```diff
@@ -176,6 +176,17 @@ class CEPH_RBD_API ImageOptions {
...
 
+class CEPH_RBD_API UpdateWatchCtx {
+public:
+  virtual ~UpdateWatchCtx() {}
+  /**
+   * Callback activated when we receive a notify event.
+   */
+  virtual void handle_notify() = 0;
+
+  virtual void handle_unregister() {}
```
dillaman: Instead of using a dummy virtual function to handle the C API deletion problem, could you just have the C API return a pointer to the allocated ```C_UpdateWatchCB``` object via the handle and for unwatch you could cast the handle back to ```C_UpdateWatchCB```, extract the real handle, unwatch, and delete.

src/librbd/ImageCtx.h
```diff
@@ -303,6 +303,8 @@ namespace librbd {
 
     journal::Policy *get_journal_policy() const;
     void set_journal_policy(journal::Policy *policy);
+
+    void notify_watchers();
```
dillaman: Unused?

src/librbd/ImageState.cc
```diff
@@ -23,6 +24,155 @@ using util::create_async_context_callback;
...
+      ldout(cct, 20) << __func__ << ": completing shut down" << dendl;
+      on_shut_down_finish->complete(0);
+    }
+  }
+
+private:
+  I *m_image_ctx;
+  Mutex m_lock;
+  std::map<uint64_t, UpdateWatchCtx*> m_watchers;
+  uint64_t next_handle = 0;
```
dillaman: Minor: ```m_next_handle```

src/librbd/ImageState.cc
```diff
@@ -23,6 +24,155 @@ using util::create_async_context_callback;
...
+    Context *on_shut_down_finish = nullptr;
+
+    {
+      Mutex::Locker locker(m_lock);
+
+      auto in_flight_it = m_in_flight.find(handle);
+      assert(in_flight_it != m_in_flight.end());
+      m_in_flight.erase(in_flight_it);
+
+      if (m_in_flight.find(handle) == m_in_flight.end()) {
```
dillaman: Is it ever possible for this to be false?

src/librbd/ImageState.cc
```diff
@@ -23,6 +24,155 @@ using util::create_async_context_callback;
...
+    int r = 0;
+    {
+      Mutex::Locker locker(m_lock);
+      auto it = m_watchers.find(handle);
+      if (it == m_watchers.end()) {
+	r = -ENOENT;
+      } else {
+	if (m_in_flight.find(handle) != m_in_flight.end()) {
+	  assert(m_pending_unregister.find(handle) == m_pending_unregister.end());
+	  m_pending_unregister[handle] = on_finish;
```
dillaman: Perhaps instead of tracking pending unregisters, just erase the handle from ```m_watchers``` and at ```handle_notify```, if the handle is no longer part of ```m_watchers```, you know you were unregistered?

dillaman: Never mind -- you are using it for the Context

src/librbd/ImageState.cc
```diff
@@ -152,6 +311,45 @@ void ImageState<I>::snap_set(const std::string &snap_name, Context *on_finish) {
 }
 
 template <typename I>
+int ImageState<I>::register_update_watcher(UpdateWatchCtx *watcher,
+					 uint64_t *handle) {
+  CephContext *cct = m_image_ctx->cct;
+  ldout(cct, 20) << __func__ << dendl;
+
+  Mutex::Locker locker(m_lock);
+  if (m_update_watchers == nullptr) {
```
dillaman: Is it possible for the user to have access to the image before it's fully opened?  Perhaps an assert if not.

dillaman:
@trociny I can't think of a good way around the issue.  I think just have ```ImageUpdateWatchers``` start a new thread upon registering the first watcher.  We could modify ```ImageState::handle_update_notification``` to take a boolean to indicate if the update was received from ```ImageWatcher```, which means that we are in a the librados thread.  However, you would hit the same deadlock issue when trying to refresh the image and librados async completions won't be delivered.

src/librbd/ImageState.cc
trociny: Yes. m_in_flight is a mutliset, so it queues notifications with the same handle.
I can use std::map<uint64_t, int> (handle,count) instead if you think it is more readable.

dillaman: Ah -- no worries. Perhaps a short comment would be sufficient

trociny:
@dillaman Updated. May be move ImageUpdateWatchers to a separate file?

dillaman:
@trociny bot failure looks real

trociny:
@dillaman Sorry, missed this. I believe I have fixed this, adding `op_work_queue->drain()` (I thought initially it was not necessary because delete is always after shut down, but forgot that shut down completion is executed by the worker, so a race was still possible).

Still jenkins reported another strange crash:
```
[ RUN      ] TestLibRBD.TestImageOptions
./test/run-rbd-unit-tests.sh: line 9: 16207 Segmentation fault      RBD_FEATURES=$i unittest_librbd
```
https://jenkins.ceph.com/job/ceph-pull-requests/6893/console

I have failed to reproduce this locally. Investigating

dillaman:
@trociny any luck reproducing?

trociny:
@dillaman No luck to reproduce so far. valgrind does not report anything that could look related. Re-reading the patch did not help either. Strangely, it failed in two different places in unittest_librbd, and these were not my watcher tests.

trociny:
retest this please

dillaman:
lgtm -- and now that fsx works, it is finding new (unrelated) bugs

```
ceph_test_librbd_fsx -d -W -R -p 100 -P /home/jdillaman/archive -r 512 -w 512 -t 512 -h 512 -l 250000000 -S 0 -N 10000 -M -U pool_client.0 image_client.0
...
3744 write	0xecc5a00 thru	0xecc92b6	(0x38b7 bytes)
short write: 0x3600 bytes instead of 0x38b7
```

dillaman:
@trociny Looks like there is a race during shut down:

> #0  0x00007fbc2206e272 in notify (this=0x0) at librbd/ImageState.cc:94
> #1  librbd::ImageState<librbd::ImageCtx>::handle_update_notification (this=0x7fbc351f10d0) at librbd/ImageState.cc:276
> #2  0x00007fbc22070a69 in librbd::ImageWatcher::handle_payload (this=this@entry=0x7fbbf40008c0, payload=..., 
>     ack_ctx=ack_ctx@entry=0x7fbbf400b320) at librbd/ImageWatcher.cc:547
> #3  0x00007fbc22078e4c in operator()<librbd::watch_notify::HeaderUpdatePayload> (payload=..., this=<optimized out>)
>     at ./librbd/ImageWatcher.h:217
> #4  internal_visit<librbd::watch_notify::HeaderUpdatePayload const> (operand=..., this=<synthetic pointer>)
>     at /usr/include/boost/variant/variant.hpp:1017

http://teuthology.ovh.sepia.ceph.com/teuthology/jdillaman-2016-06-15_01:52:01-rbd-master---basic-openstack/24158/teuthology.log

dillaman:
http://teuthology.ovh.sepia.ceph.com/teuthology/jdillaman-2016-06-15_01:52:01-rbd-master---basic-openstack/24115/teuthology.log

dillaman:
@trociny needs rebase

trociny:
@dillaman Updated. I changed it to create m_update_watchers on ImageState creation and destroy in ImageState destructor, and m_update_watchers->shut_down() is called in close request to drain pending notifications and unregistrations. New notifications are added only when current state is OPEN, to ensure no new notifications come after shut_down is started.

dillaman:
retest this please

dillaman:
fsx failed against the NBD device complaining of an unexpected size (reported old size, not the new size):

http://teuthology.ovh.sepia.ceph.com/teuthology/jdillaman-2016-06-22_14:28:01-rbd-wip-jd-testing---basic-openstack/27827/teuthology.log

trociny:
@dillaman Thanks, I will try to reproduce. I see in that test they used cache tiering. Do you think it might be related?

dillaman:
@trociny I would hope it wouldn't be related to that

trociny:
@dillaman I easily reproduced this locally (sure it is not cache tiering related). I still failed to make rbd-nbd produce debug logs when started by ceph_test_librbd_fsx (or can't find them), but right now I think it is expected behaviour. The update callback is executed asynchronously, so the race between changing the image size and testing for the new size is expected. When I add retry loop to fsx.cc:check_size(), it always succeeds eventually. (still there are other fsx failures or hangs). BTW, couldn't it be the cause of the initially reported failures?

dillaman:
@trociny Oh -- so it's because of that async callback into the client to notify them of the update? The resize won't complete until it has received ACKs from all active watchers that they have processed the update. Sounds like we need to delay the ack until the client callbacks have completed.

trociny:
> Sounds like we need to delay the ack until the client callbacks have completed.

Will try. Thanks.

trociny:
@dillaman I have added flush method to ImageUpdateWatchers, which is called before sending HeaderUpdate watcher notification ack. It works for me (tested by adding 0.5 sec delay in rbd-nbd callback, which previously triggered fsx check size failure immediately).

If you like this approach I will squash it in the final patchset.

src/librbd/ImageState.cc
```diff
@@ -22,16 +23,231 @@ namespace librbd {
...
+    assert(m_on_shut_down_finish == nullptr);
+
+    destroy_work_queue();
+  }
+
+  void flush(Context *on_finish) {
+    ldout(m_cct, 20) << "ImageUpdateWatchers::" << __func__ << dendl;
+    {
+      Mutex::Locker locker(m_lock);
+      assert(m_on_flush_finish == nullptr);
```
dillaman: Not sure this is a safe assumption -- can get repeated notifications and/or races with multiple updates.

src/librbd/ImageState.cc
```diff
@@ -22,16 +23,231 @@ namespace librbd {
...
+      if (!m_in_flight.empty()) {
+	m_on_flush_finish = on_finish;
```
dillaman: If ```m_in_flight``` isn't empty, can you just append the context to ```m_work_queue```?

trociny: @dillaman Thanks. Updated.

dillaman:
lgtm

dillaman:
@trociny Looks like it can result in a crash if the image header watch is lost: http://teuthology.ovh.sepia.ceph.com/teuthology/jdillaman-2016-07-06_17:56:15-rbd-wip-jd-testing---basic-openstack/33086/teuthology.log

> Program terminated with signal SIGSEGV, Segmentation fault.
> #0  0x00007f181c43f47c in librbd::ImageWatcher::C_ResponseMessage::finish(int) () from /usr/lib/librbd.so.1
> #1  0x00007f181c3a73d5 in Context::complete(int) () from /usr/lib/librbd.so.1
> #2  0x00007f181c42aff1 in librbd::ImageUpdateWatchers::flush(Context*) () from /usr/lib/librbd.so.1
> #3  0x00007f181c42ed17 in librbd::ImageState<librbd::ImageCtx>::flush_update_watchers(Context*) () from /usr/lib/librbd.so.1
> #4  0x00007f181c43b419 in librbd::ImageWatcher::handle_payload(librbd::watch_notify::HeaderUpdatePayload const&, librbd::ImageWatcher::C_NotifyAck*) () from /usr/lib/librbd.so.1
> #5  0x00007f181c43ee61 in librbd::ImageWatcher::reregister_watch() () from /usr/lib/librbd.so.1
>

trociny:
@dillaman Updated -- in ImageWatcher::handle_payload() flush is called only if ack_ctx is not null.

PR #9201, opened 2016-05-19
pybind: misc fixes

marcan:
Three minor fixes discovered while writing AIO support for pybind/rbd (will be a separate PR).

PR #9292, opened 2016-05-24
pybind: AIO bindings for RBD

marcan:
This adds Python bindings for the AIO functions of librbd. The approach is similar to the one used in the rados bindings, but the persistence mechanism for completions is different. Instead of keeping references to completion objects around in the Image object, this just has the Completion keep a reference to itself to prevent Python from freeing it (along with a Cython annotation to make sure it doesn't try to break the reference cycle via garbage collection). This avoids having completions lock on Image-global data structures and IMO makes the code simpler (the GIL implicitly synchronizes completions of course).

There is no `rbd_aio_wait_for_complete_and_cb`, but it's a very useful function to have, so the bindings emulate it using an event object. The intent is for the user to call `Completion.wait_for_complete_and_cb` to "reap" the completion (even if it is known to have completed via some other mechanism), as that will also propagate any asynchronous exceptions raised by the callback.

Note that this is based off of #9201, so only the last commit is to be reviewed here. It's probably independent, but I haven't tested it standalone.

Comes with basic tests, but I'm also testing this daily in a production-ish environment for an RBD backup process with a TB of data or so (and fairly high async parallelism for performance).

cc @jdurgin

If this approach to completion tracking looks good, I'm thinking of changing over the rados bindings to do it the same way.

### DIFF
--- a/src/pybind/rbd/rbd.pyx
+++ b/src/pybind/rbd/rbd.pyx
@@ -14,6 +14,9 @@ method.
 # Copyright 2011 Josh Durgin
 # Copyright 2015 Hector Martin <marcan@marcan.st>
 
+import cython
+import sys
+
 from cpython cimport PyObject, ref, exc
 from libc cimport errno
 from libc.stdint cimport *
@@ -37,8 +40,6 @@ cdef extern from "Python.h":
 
 cdef extern from "time.h":
     ctypedef long int time_t
-
-ctypedef int (*librbd_progress_fn_t)(uint64_t offset, uint64_t total, void* ptr)
 
 cdef extern from "rbd/librbd.h" nogil:
     enum:
@@ -71,6 +72,7 @@ cdef extern from "rbd/librbd.h" nogil:
     ctypedef void* rados_ioctx_t
     ctypedef void* rbd_image_t
     ctypedef void* rbd_image_options_t
+    ctypedef void *rbd_completion_t
 
     ctypedef struct rbd_image_info_t:
         uint64_t size
@@ -122,6 +124,9 @@ cdef extern from "rbd/librbd.h" nogil:
         char *description
         time_t last_update
         bint up
+
+    ctypedef void (*rbd_callback_t)(rbd_completion_t cb, void *arg)
+    ctypedef int (*librbd_progress_fn_t)(uint64_t offset, uint64_t total, void* ptr)
 
     void rbd_version(int *major, int *minor, int *extra)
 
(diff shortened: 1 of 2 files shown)


src/pybind/rbd/rbd.pyx
```diff
@@ -444,6 +465,122 @@ cdef void* realloc_chk(void* ptr, size_t size) except NULL:
...
+    def wait_for_complete(self):
+        """
+        Wait for an asynchronous operation to complete
+
+        This does not imply that the complete callback has finished.
+        """
+        with nogil:
+            rbd_aio_wait_for_complete(self.rbd_comp)
+
+    def wait_for_complete_and_cb(self):
```
dillaman: We should probably fix this in librbd -- pre-Jewel you could always be assured that the callback was fired before ```rbd_aio_is_complete``` would return true.  I don't recall why I moved the assignment of ```AioCompletion::done``` in fde9f785

jdurgin: :+1: not needing it before was why rbd_aio_is_complete_and_cb() didn't get created at the same time as the rados version

src/test/pybind/test_rbd.py
```diff
@@ -3,6 +3,8 @@
 import socket
 import os
 import time
+import threading
```
jdurgin: minor: unused import

jdurgin:
I like this tracking approach better - like you said, it is simpler and avoids extra locking. Doing the same thing in the rados bindings sounds good.

This looks good to me in general. Like @dillaman mentioned, I think we should fix up
rbd_aio_wait_for_complete() in librbd itself, so we can get rid of the threading.Event here.

src/pybind/rbd/rbd.pyx
marcan: I'm not really familiar with the nuances of the async handling in librbd, but if you can't think of a reason for the `done = true` to happen before the callback I can send a PR to move that back and get rid of the event. Let me know if that sounds sensible.

dillaman: Sounds good, thanks

src/test/pybind/test_rbd.py
marcan: Fixed.

marcan:
Updated to remove the Event object. This now depends on (and is rebased on top of) #9326 too. The method is still called `wait_for_complete_and_cb` because I want to emphasize that the callback is waited on and any exceptions are propagated, but now it really just calls `rbd_aio_wait_for_complete` and does the propagation.

dillaman:
@marcan Can you rebase this off the current master now that #9326 has been merged?

marcan:
Done. For some reason I thought I had already done that.

dillaman:
@marcan Did some commits get dropped? I seems to remember some changes with __del__ vs dealloc (or something like that)

marcan:
Those merged in #9201 :)

dillaman:
lgtm -- thx

yuriw:
Per chat with @liewegas this PR fails to build on centos

marcan:
That's weird, I can't think of anything distro-specific in here, unless it's simply about the Cython version. Do you have a build log?

yuriw:
I am not completely sure on details but here is the top error log http://gitbuilder.sepia.ceph.com/gitbuilder-ceph-rpm-centos7-amd64-notcmalloc/log.cgi?log=5202b34e944de997a6edbf8edde81a066a2cbc41

marcan:
It's complaining about an addition in this PR (`Invalid directive: 'no_gc_clear'`), but it's also complaining about a ton of other things that are already there in master. Can you actually build master without this change? It sounds to me like that gitbuilder has an outdated Cython version.

marcan:
FWIW, that directive was added in Cython 0.20 (released 2014-01-18).

dillaman:
@marcan RHEL 7 / CentOS 7 use Cython v0.19 (+ bug fixes)

marcan:
Well, that's unfortunate. I guess I'll have to switch to bare INCREF/DECREF. I'll do that and review the logic tomorrow.

marcan:
Just following up, will get to this on Monday. Busy week.

dillaman:
no worries -- thanks for the update.

marcan:
Updated the branch with a commit to switch from the circular reference to plain INCREF/DECREF. I ran the tests and our production backup code which uses these bindings and it looks like it works fine.

PR #9295, opened 2016-05-24
rgw: use vector for librados handles

cbodley:
using a vector instead of an array of pointers cleans up our initialization/shutdown logic

### DIFF
--- a/src/rgw/rgw_rados.cc
+++ b/src/rgw/rgw_rados.cc
@@ -3199,8 +3199,7 @@ void RGWRados::finalize()
 int RGWRados::init_rados()
 {
   int ret = 0;
-  auto count = cct->_conf->rgw_num_rados_handles;
-  auto handles = std::vector<librados::Rados>{count};
+  auto handles = std::vector<librados::Rados>{cct->_conf->rgw_num_rados_handles};
 
   for (auto& r : handles) {
     ret = r.init_with_context(cct);
@@ -3223,7 +3222,6 @@ int RGWRados::init_rados()
   meta_mgr = new RGWMetadataManager(cct, this);
   data_log = new RGWDataChangesLog(cct, this);
 
-  num_rados_handles = count;
   std::swap(handles, rados);
   return ret;
 }
@@ -12083,7 +12081,7 @@ void RGWStoreManager::close_storage(RGWRados *store)
 
 librados::Rados* RGWRados::get_rados_handle()
 {
-  if (num_rados_handles == 1) {
+  if (rados.size() == 1) {
     return &rados[0];
   } else {
     handle_lock.get_read();
@@ -12097,7 +12095,7 @@ librados::Rados* RGWRados::get_rados_handle()
       handle_lock.put_read();
       handle_lock.get_write();
       uint32_t handle = next_rados_handle.read();
-      if (handle == num_rados_handles) {
+      if (handle == rados.size()) {
         next_rados_handle.set(0);
         handle = 0;
       }
--- a/src/rgw/rgw_rados.cc
+++ b/src/rgw/rgw_rados.cc
@@ -3213,14 +3213,16 @@ int RGWRados::init_rados()
     }
   }
 
-  cr_registry = new RGWCoroutinesManagerRegistry(cct);
-  ret =  cr_registry->hook_to_admin_command("cr dump");
+  auto crs = std::unique_ptr<RGWCoroutinesManagerRegistry>{
+    new RGWCoroutinesManagerRegistry(cct)};
+  ret = crs->hook_to_admin_command("cr dump");
   if (ret < 0) {
     return ret;
   }
 
   meta_mgr = new RGWMetadataManager(cct, this);
   data_log = new RGWDataChangesLog(cct, this);
+  cr_registry = crs.release();
 
   std::swap(handles, rados);
   return ret;
--- a/src/rgw/rgw_rados.cc
+++ b/src/rgw/rgw_rados.cc
@@ -12094,13 +12094,11 @@ librados::Rados* RGWRados::get_rados_handle()
     } else {
       handle_lock.put_read();
       handle_lock.get_write();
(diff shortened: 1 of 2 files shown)


src/rgw/rgw_rados.h
```diff
@@ -1956,17 +1955,7 @@ class RGWRados
...
-      if (rados[i]) {
-        rados[i]->shutdown();
-        delete rados[i];
-      }
-    }
-    if (rados) {
-      delete[] rados;
-    }
-  }
+  virtual ~RGWRados() = default;
```
rzarzynski: We aren't calling `librados::Rados::shutdown()` anymore. Is that OK? Does `librados::Rados::~Rados()` do that for us?

src/rgw/rgw_rados.h
```diff
@@ -1800,9 +1800,8 @@ class RGWRados
 protected:
   CephContext *cct;
 
-  librados::Rados **rados;
-  atomic_t next_rados_handle;
-  uint32_t num_rados_handles;
+  std::vector<librados::Rados> rados;
+  uint32_t next_rados_handle;
```
rzarzynski: This is rather stylistic but, as `next_rados_handle` is used as an index, it would be nice to see `size_t` here.

src/rgw/rgw_rados.cc
```diff
@@ -3199,58 +3199,32 @@ void RGWRados::finalize()
...
+    new RGWCoroutinesManagerRegistry(cct)};
+  ret = crs->hook_to_admin_command("cr dump");
   if (ret < 0) {
-    goto fail;
+    return ret;
   }
 
   meta_mgr = new RGWMetadataManager(cct, this);
   data_log = new RGWDataChangesLog(cct, this);
+  cr_registry = crs.release();
```
rzarzynski: Maybe not now but is there a chance to convert `cr_registry` into a C++11's smart pointer?

src/rgw/rgw_rados.cc
```diff
@@ -12109,28 +12083,26 @@ void RGWStoreManager::close_storage(RGWRados *store)
...
       handle_lock.get_write();
-      uint32_t handle = next_rados_handle.read();
-      if (handle == num_rados_handles) {
-        next_rados_handle.set(0);
-        handle = 0;
-      }
+      const uint32_t handle = next_rados_handle;
       rados_map[id] = handle;
-      next_rados_handle.inc();
+      if (++next_rados_handle == rados.size()) {
```
rzarzynski: Is holding the read-write on `handle_lock` enough to protect `next_rados_handle`? In the previous version it was additionally `atomic_t`. Are we sure that it isn't touched by someone else who doesn't actually take the lock?

src/rgw/rgw_rados.cc
```diff
@@ -3199,58 +3199,32 @@ void RGWRados::finalize()
 int RGWRados::init_rados()
 {
   int ret = 0;
+  auto handles = std::vector<librados::Rados>{cct->_conf->rgw_num_rados_handles};
```
rzarzynski: I like this idea very much!

src/rgw/rgw_rados.h
rzarzynski: It looks we're leaking `cr_registry` but this isn't new (no regression here).

Yan-waller:
lgtm

`Reviewed-by: Yan Jun <yan.jun8@zte.com.cn>`

src/rgw/rgw_rados.cc
cbodley: this function is the only place that `next_rados_handle` is referenced, it doesn't need to be atomic

rzarzynski: Great. Yet another one optimization here. :-)

src/rgw/rgw_rados.h
cbodley: > Does `librados::Rados::~Rados()` do that for us?

```cpp
librados::Rados::~Rados()
{
  shutdown();
}
```

> It looks we're leaking `cr_registry` but this isn't new (no regression here).

correct, `cr_registry` is only freed in `RGWRados::finalize()`. before this patch, it was leaked if the call to `cr_registry->hook_to_admin_command("cr dump");` failed in `RGWRados::init_rados()`. but if `init_rados()` succeeds, it's fair to expect `finalize()` to be called for cleanup

src/rgw/rgw_rados.cc
cbodley: agreed with all members of `RGWRados`, althought it has very specific requirements for the order of construction and destruction. so we wouldn't want to rely on `~RGWRados` to free them automatically

src/rgw/rgw_rados.h
yehudasa: @cbodley this sounds familiar, I might have fixed it somewhere else

src/rgw/rgw_rados.cc
yehudasa: @cbodley can't this function be called concurrently?

cbodley: `next_rados_handle` is only accessed under an exclusive lock on `handle_lock`

cbodley:
this branch went through [teuthology](http://pulpito.ceph.com/cbodley-2016-05-28_11:42:37-rgw-wip-rgw-testing-cbodley---basic-smithi/) with two valgrind issues:
* `valgrind issue <kind>Leak_DefinitelyLost</kind> in /var/log/ceph/valgrind/mon.b.log`
* `valgrind issue <kind>InvalidRead</kind> in /var/log/ceph/valgrind/client.0.log`

the invalid read was in `RGWRados::get_obj_iterate_cb()`, which is unrelated to this pr. i rescheduled that one test and it went through [clean](http://pulpito.ceph.com/cbodley-2016-05-31_13:33:10-rgw-wip-rgw-testing-cbodley---basic-smithi/)

PR #9297, opened 2016-05-24
rgw: support size suffixes for --max-size in radosgw-admin command

vumrao:
rgw: support size suffixes for --max-size in radosgw-admin command

Fixes: http://tracker.ceph.com/issues/16004

### DIFF
--- a/src/rgw/rgw_admin.cc
+++ b/src/rgw/rgw_admin.cc
@@ -233,7 +233,7 @@ void _usage()
   cout << "\nQuota options:\n";
   cout << "   --bucket                  specified bucket for quota command\n";
   cout << "   --max-objects             specify max objects (negative value to disable)\n";
-  cout << "   --max-size                specify max size (in bytes, negative value to disable)\n";
+  cout << "   --max-size                specify max size (in B/K/M/G/T, negative value to disable)\n";
   cout << "   --quota-scope             scope of quota (bucket, user)\n";
   cout << "\nOrphans search options:\n";
   cout << "   --pool                    data pool to scan for leaked rados objects in\n";
@@ -1027,7 +1027,11 @@ void set_quota_info(RGWQuotaInfo& quota, int opt_cmd, int64_t max_size, int64_t 
 
     case OPT_QUOTA_SET:
       if (have_max_objects) {
-        quota.max_objects = max_objects;
+        if (max_objects < 0) {
+          quota.max_objects = -1;
+        } else {
+          quota.max_objects = max_objects;
+        }
       }
       if (have_max_size) {
         if (max_size < 0) {
@@ -2148,7 +2152,7 @@ int main(int argc, char **argv)
         return EINVAL;
       }
     } else if (ceph_argparse_witharg(args, i, &val, "--max-size", (char*)NULL)) {
-      max_size = (int64_t)strict_strtoll(val.c_str(), 10, &err);
+      max_size = strict_si_cast<int64_t>(val.c_str(), &err);
       if (!err.empty()) {
         cerr << "ERROR: failed to parse max size: " << err << std::endl;
         return EINVAL;
--- a/doc/radosgw/admin.rst
+++ b/doc/radosgw/admin.rst
@@ -310,7 +310,7 @@ storage size in megabytes.
   the maximum number of objects. A negative value disables this setting.
   
 - **Maximum Size:** The ``--max-size`` option allows you to specify a quota
-  for the maximum number of bytes. A negative value disables this setting.
+  size in B/K/M/G/T. A negative value disables this setting.
   
 - **Quota Scope:** The ``--quota-scope`` option sets the scope for the quota.
   The options are ``bucket`` and ``user``. Bucket quotas apply to buckets a 
@@ -327,7 +327,7 @@ For example::
 
 For example:: 
 
-	radosgw-admin quota set --quota-scope=user --uid=johndoe --max-objects=1024 --max-size=1024
+	radosgw-admin quota set --quota-scope=user --uid=johndoe --max-objects=1024 --max-size=1024B
 
 
 A negative value for num objects and / or max size means that the
--- a/src/common/strtol.cc
+++ b/src/common/strtol.cc
@@ -187,6 +187,7 @@ T strict_si_cast(const char *str, std::string *err)
 
 template int strict_si_cast<int>(const char *str, std::string *err);
 template long long strict_si_cast<long long>(const char *str, std::string *err);
+template int64_t strict_si_cast<int64_t>(const char *str, std::string *err);
 template uint64_t strict_si_cast<uint64_t>(const char *str, std::string *err);
 template uint32_t strict_si_cast<uint32_t>(const char *str, std::string *err);
 
--- a/src/test/cli/radosgw-admin/help.t
+++ b/src/test/cli/radosgw-admin/help.t
@@ -186,7 +186,7 @@
   Quota options:
      --bucket                  specified bucket for quota command
      --max-objects             specify max objects (negative value to disable)
-     --max-size                specify max size (in bytes, negative value to disable)
+     --max-size                specify max size (in B/K/M/G/T, negative value to disable)
      --quota-scope             scope of quota (bucket, user)
   
   Orphans search options:
--- a/src/test/strtol.cc
+++ b/src/test/strtol.cc
@@ -234,6 +234,21 @@ TEST(StrictSICast, Error) {
     (void)strict_si_cast<int>("1T", &err);
     ASSERT_NE(err, "");
   }
+  {
+    std::string err;
+    (void)strict_si_cast<int64_t>("2E", &err);
+    ASSERT_EQ(err, "");
+  }
+  {
+    std::string err;
+    (void)strict_si_cast<int64_t>("-2E", &err);
+    ASSERT_EQ(err, "");
+  }
+  {
+    std::string err;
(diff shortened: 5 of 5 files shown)


vumrao:
- I have removed `A negative value disables this setting`. As per my understanding we are not using it to disable this setting. 
- As `strict_strtoll()` do not fail with `negative` value and still we are setting `have_max_size = true`.

- Now with `strict_sistrtoll()` we wont accept `negative` value and we wont set `have_max_size = true`.

- And same goes for `have_max_objects`.

vumrao:
- I have tested this change with vstart cluster and it works fine, below are test results : 

~~~
$./radosgw-admin quota set --quota-scope=user --uid=testid --max-size=1G
$./radosgw-admin user info --uid=testid
[...]
"user_quota": {
        "enabled": false,
        "max_size_kb": 1048576,
        "max_objects": -1
    }, 
~~~

- For 2T : 

~~~
$./radosgw-admin quota set --quota-scope=user --uid=testid --max-size=2T
$./radosgw-admin user info --uid=testid
[...]
"user_quota": {
        "enabled": false,
        "max_size_kb": 2147483648,
        "max_objects": -1
    },
~~~

- For `--max-objects` : 

~~~
 $ ./radosgw-admin quota set --quota-scope=user --uid=testid --max-objects=100000
$./radosgw-admin user info --uid=testid
[...]
 "user_quota": {
        "enabled": false,
        "max_size_kb": 2147483648,
        "max_objects": 100000
    },
~~~

cbodley:
hi @vumrao, the `make check` bot doesn't appear to be working, but we have a unit test that runs `radosgw-admin -h` and compares the output against `src/test/cli/radosgw-admin/help.t`. could you please update that file to match your usage changes and make sure the test passes?

you can either run the whole `make check`, or try to run `run-cli-tests` by itself:

if you're building with cmake, from your cmake build directory:
`ctest -V -R run-cli-tests`

if you're building with automake, from ceph/src:
`PATH=$PATH:. test/run-cli-tests` (should work but haven't tested)

src/rgw/rgw_admin.cc
```diff
@@ -2151,14 +2151,14 @@ int main(int argc, char **argv)
...
-      max_size = (int64_t)strict_strtoll(val.c_str(), 10, &err);
+      max_size = (int64_t)strict_sistrtoll(val.c_str(), &err);
       if (!err.empty()) {
         cerr << "ERROR: failed to parse max size: " << err << std::endl;
         return EINVAL;
       }
       have_max_size = true;
     } else if (ceph_argparse_witharg(args, i, &val, "--max-objects", (char*)NULL)) {
-      max_objects = (int64_t)strict_strtoll(val.c_str(), 10, &err);
+      max_objects = (int64_t)strict_sistrtoll(val.c_str(), &err);
```
theanalyst: isn't max_objects supposed to be still a number? (ie. without units?)

theanalyst: nvm, I hadn't read what strict_sistroll was doing, looks good

vumrao:
@cbodley thanks for your inputs. Sure will modify cli test with my changes and test it and then will send the modified patch.

vumrao:
In my local test system all tests have passed.

vumrao:
test this please.

vumrao:
- Not related : 
~~~
make[3]: *** [libec_jerasure.la] Error 1
../libtool: line 1099: .libs/libec_shec_generic.exp: No space left on device
../libtool: line 1099: .libs/libec_jerasure_sse4.exp: No space left on device
~~~~

vumrao:
retest this please.

vumrao:
- Again same :( 
~~~
  CXXLD    ceph_test_cfuse_cache_invalidate
Cannot create temporary file in /tmp/: No space left on device
~~~

cbodley:
@vumrao looking at `RGWQuotaHandlerImpl::check_quota()` in rgw_quota.cc, it does look like negative values for `max_objects` and `max_size_kb` will disable the quota enforcement. unless you've seen different behavior in testing, i think we still need to allow setting them to -1

vumrao:
@cbodley Thanks for your inputs. Below are my findings : 

- As per my understanding `RGWQuotaHandlerImpl::check_quota()` is used for checking  the quota is enabled or not and if enable and it exceeds  then return `ERR_QUOTA_EXCEEDED`. 

- `RGWQuotaHandlerImpl` do have handling of global quota options from command line and here we make use of `max_objects` and `max_size_kb` to check at these options level if we have positive values for these options.  Yes we should accept negative values for these options as if user has enabled the quota and later he want to disable it could be disabled with `radosgw-admin quota --disable` and also setting -1 to `max_objects` and `max_size_kb`.

- Quota values set and enable is performed in function `set_quota_info()` in rgw_admin.cc. 

~~~
void set_quota_info(RGWQuotaInfo& quota, int opt_cmd, int64_t max_size, int64_t max_objects,
                    bool have_max_size, bool have_max_objects)
{
  switch (opt_cmd) {
    case OPT_QUOTA_ENABLE:
      quota.enabled = true;

      // falling through on purpose

    case OPT_QUOTA_SET:
      if (have_max_objects) {
        quota.max_objects = max_objects;
      }
      if (have_max_size) {
        if (max_size < 0) {
          quota.max_size_kb = -1;
        } else {
          quota.max_size_kb = rgw_rounded_kb(max_size);
        }
      }
      break;
    case OPT_QUOTA_DISABLE:
      quota.enabled = false;
      break;
  }
}
~~~

- If we check no where we are making use of either max_size or max_objects to disable the quota means `case OPT_QUOTA_DISABLE` and this case is only used when using `radosgw-admin` with  `quota --disable`.

- If we check above given function set_quota_info() we are just setting `max_size` as -1 if it has any negative value which is correct as default is -1 but it is not same for `max_objects` and I think we should do the same for `max_objects` set always -1 for any negative value. As here `max_size_kb` and `max_objects` as -1 means quota is disabled. 

- As new function `strict_sistrtoll()` do not accept negative values , I will handle negative values with little different approach and will handle setting `max_size` and `max_objects` as -1 at options `--max-size` and `--max-objects` level.

- Will modify and send new patch. 

Please let me know your inputs. Thanks.

vumrao:
- Jenkins slaves still have issues of `No space left on device` : 

~~~
make[3]: *** [libec_jerasure_sse4.la] Error 1
ln: failed to create symbolic link 'libceph_example.la': No space left on device
make[3]: *** [libceph_example.la] Error 1
~~~

src/rgw/rgw_admin.cc
```diff
@@ -1030,7 +1030,11 @@ void set_quota_info(RGWQuotaInfo& quota, int opt_cmd, int64_t max_size, int64_t
 
     case OPT_QUOTA_SET:
       if (have_max_objects) {
-        quota.max_objects = max_objects;
+        if (max_objects < 0) {
+          quota.max_objects = -1;
+        } else {
+          quota.max_objects = max_objects;
+        }
```
cbodley: is this change really needed? the quota checks just use >= 0, so we're not really treating -1 as a special value

src/rgw/rgw_admin.cc
```diff
@@ -2151,10 +2155,14 @@ int main(int argc, char **argv)
...
+      max_size = (int64_t)strict_sistrtoll(val.c_str(), &err);
       if (!err.empty()) {
-        cerr << "ERROR: failed to parse max size: " << err << std::endl;
-        return EINVAL;
+        if (!strcmp(err.c_str(), "strict_sistrtoll: value should not be negative")) {
+          max_size = -1;
+        } else {
+          cerr << "ERROR: failed to parse max size: " << err << std::endl;
+          return EINVAL;
+        }
```
cbodley: since `strict_sistrtoll()` is just a wrapper for `strict_si_cast<uint64_t>()`, i think we can just use `strict_si_cast<int64_t>()` here instead. this gives us the correct type for `max_size`, allows negative values, and doesn't require parsing the error string. can you give that a try?

cbodley: i see that `int64_t` is not one of the template types instantiated for `strict_si_cast()` in src/common/strtol.cc, so we'll want to add that for use here.. which means we'll also need some tests for it in src/test/strtol.cc. would you be willing to look into that, or should i?

src/rgw/rgw_admin.cc
vumrao: Nothing as such. I just thought it would be same for `max_objects` as we are handling for `max_size`.

src/rgw/rgw_admin.cc
vumrao: Ahh got it. Sure, I will give a try and will take your help if I will get any issue.

cbodley: I just pushed a branch to https://github.com/cbodley/ceph/commits/wip-int64-strtol, feel free to cherry-pick that last commit (https://github.com/cbodley/ceph/commit/b708360167d4b05fd547fbce529be6f868cffb4d) into your pr

cbodley: scratch that.. there's already a template for `strict_si_cast<long long>()`, which passes those same unit tests.. just use that instead

vumrao: @cbodley Thanks for your inputs. I have not added all tests as your commit has but if I understood `strtol` test cases correct, I have just replicated the  `(void)strict_si_cast<int>( )` test to `(void)strict_si_cast<int64_t>( )`.

vumrao:
@cbodley thank you.

PR #9300, opened 2016-05-24
cmake: add libcephfs-java support

tchaikov:
@cbodley mind taking a look? thanks.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -118,6 +118,11 @@ if(WITH_OCF)
   add_subdirectory(ocf)
 endif()
 
+option(WITH_CEPHFS_JAVA "build libcephfs Java bindings" OFF)
+if(WITH_CEPHFS_JAVA)
+  add_subdirectory(java)
+endif()
+
 # Python stuff
 find_package(PythonInterp 2 QUIET)
 if(NOT PYTHONINTERP_FOUND)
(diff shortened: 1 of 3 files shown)


src/CMakeLists.txt
```diff
@@ -1332,3 +1337,11 @@ add_custom_target(cephfs_testing DEPENDS
...
     cephfs-data-scan
     cephfs-table-tool)
+
+if (IS_DIRECTORY "${PROJECT_SOURCE_DIR}/.git")
+  add_custom_target(
+    git-update
+    COMMAND git submodule sync
+    COMMAND git submodule update --force --init --recursive
+    WORKING_DIRECTORY "${PROJECT_SOURCE_DIR}")
+endif()
```
cbodley: very cool! does this get run automatically, or do we need something that depends on `git-update`?

cbodley: ah, it's for `make git-update`

tchaikov: i was thinking about 

* using `ExternalProject_Add()`[1] to include the submodules, by overriding its Update/Patch Step so it runs "git submodule update" for a certain submodule when ".git" exists,
* and running "git submodule sync" automatically if ".git" exists

but it is little bit more complicated. if you think it's a viable approach, maybe we can go in this direction.

--
[1] https://cmake.org/cmake/help/v3.0/module/ExternalProject.html

PR #9302, opened 2016-05-24
Improves the README for cmake instructions and slightly improved markdown syntax.

ivancich:


tchaikov:
lgtm, maybe we can rename it to `README.cmake.md`.

alimaredia:
@tchaikov I like that idea, @ivancich could you repush with that change just like you did when you added the signed off by

PR #9304, opened 2016-05-24
Handle EC recovery read errors

dzafman:
I end up with active+degraded pg which doesn't repair if I remove an object on all nodes in a way where it gets recognized as missing.  With this fix the OSD doesn't crash.  Need to figure out how to get recovery to finish.  I assume I can have an error which will have to be discovered later by scrubbing.


tchaikov:
@dzafman is this supposed to be a (yet completed) fix for http://tracker.ceph.com/issues/13937?

dzafman:
@tchaikov I updated the proper commit to show that this pull request does indeed try to address 13937.

dzafman:
@athanatos Any suggestions on how to move forward is appreciated.

dzafman:
@athanatos I updated to make failed_push() common.  I did the following:

created scenario
pg state active+degraded
./rados -p ecpool rm benchmark_data_TrustyTahr_30314_object286 (lost object)
^C
ceph pg #.# scrub (waiting for clean)
DIDN'T START
rados -p ecpool cleanup (hung or was very slow)
^C
ceph pg #.# mark_unfound_lost delete
LOG: pg has 1 objects unfound and apparently lost marking
PG state active+clean+scrubbing

There are 9 removes "waiting for scrub" some removes completed even on the pg going through this
The remove of the lost object itself is "waiting for missing object" but it should have been cleaned up by "mark_unfound_lost delete"

athanatos:
@dzafman I'm not sure what you mean by the last part.  Do you mean that the requests stayed hung?

dzafman:
@athanatos Yes, they are coming out of the dump_ops_in_flight

athanatos:
@dzafman Sounds like a bug in master to me, probably worth tracking down.

src/osd/ECBackend.cc
```diff
@@ -500,6 +500,8 @@ void ECBackend::continue_recovery_op(
 	assert(!op.recovery_progress.first);
 	dout(10) << __func__ << ": canceling recovery op for obj " << op.hoid
 		 << dendl;
+	// XXX: Doing this on the read error handling just loops in recovery
+	// How would this case ever actually happen?
```
athanatos: You can't just use cancel_pull() in the read error handling because you know something the parent doesn't: that some of the sources are faulty.  Here, I think we are failing because a source went down (not in the up or acting set, or it would have triggered peering to restart) and we no longer have enough copies.  I think it's ok to just cancel the recovery op here because the parent would already have updated its missing_loc etc when it got the log update.  Either the object later becomes recoverable again and recovery will retry or not.

src/osd/ReplicatedPG.cc
```diff
@@ -9893,6 +9893,8 @@ void ReplicatedPG::mark_all_unfound_lost(
       [=]() {
 	requeue_ops(waiting_for_all_missing);
 	waiting_for_all_missing.clear();
+	requeue_object_waiters(waiting_for_unreadable_object);
+	// XXX: What about waiting_for_degraded_object, waiting_for_blocked_object?
```
athanatos: I think waiting_for_degraded_object must be empty here.  Same with waiting_for_blocked_object.  You could asserrt to that effect.

Also, I think you should be only requeueing ops for the objects marked lost.

src/test/osd/osd-scrub-repair.sh
```diff
@@ -314,27 +314,35 @@ function TEST_list_missing_erasure_coded() {
...
+    # Stop all osd daemons
+    for id in $(seq 0 2) ; do
+        kill_daemons $dir TERM osd.$id >&2 < /dev/null || return 1
+    done
 
+    id=${osds0[0]}
+    ceph-objectstore-tool --data-path $dir/$id --journal-path $dir/$id/journal \
+        OBJ0 remove || return 1
+    id=${osds0[1]}
+    ceph-objectstore-tool --data-path $dir/$id --journal-path $dir/$id/journal \
```
tchaikov: @dzafman maybe we can switch to https://github.com/ceph/ceph/pull/9203/commits/b6d30ec9e9580bb185b7301bd9bfce1fbd29d6b6#diff-be29cc5661964533b5db541336446042R78 for faster testing?

tchaikov:
reviewed.

src/osd/ECBackend.cc
dzafman: You added this code in d9106ce5e4437ab02279c4d7dfa23ce0d69c59df to fix tracker 8161.  I tried to use an equivalent version in OnRecoveryReadComplete::finish() only to discovery that it didn't work.  I'll have to look into whether I can do something like my new _failed_push() to handle this case too.

athanatos: Yep, that's what I was talking about.  Like I said, here, it's ok because ReplicatedPG/PG already knows that the source died and won't retry until the object is found again.  You need to call something which updates missing_loc (which I think you do earlier in this series with _failed_push).

Is this code causing a bug?

dzafman: I get it now.  I didn't see this code fail in case.   I'll remove the comment.

athanatos: K.  It probably wouldn't hurt to add an explanatory comment if you want to.  I had to spend 20 minutes trying to figure out why it didn't get handled by check_recovery_sources...

src/osd/ReplicatedPG.cc
tchaikov: @dzafman this is not addressed, we can move the missing objects from `waiting_for_blocked_object` into another map and  pass that map to `requeue_object_waiters()` instead.

src/osd/ReplicatedPG.cc
```diff
@@ -9824,12 +9824,12 @@ void ReplicatedPG::recover_got(hobject_t oid, eversion_t v)
...
 }
 
-
-void ReplicatedPG::failed_push(pg_shard_t from, const hobject_t &soid)
+void ReplicatedPG::failed_push(const list<pg_shard_t> &from, const hobject_t &soid)
 {
   assert(recovering.count(soid));
   recovering.erase(soid);
-  missing_loc.remove_location(soid, from);
+  for (list<pg_shard_t>::const_iterator i = from.begin(); i != from.end() ; ++i)
```
tchaikov: could use range-based loop.

src/osd/ECBackend.cc
```diff
@@ -188,6 +188,25 @@ PGBackend::RecoveryHandle *ECBackend::open_recovery_op()
...
+  ECBackend::read_result_t &res = in.second;
+  dout(10) << __func__ << ": Read error " << hoid << " r="
+	   << res.r << " errors=" << res.errors << dendl;
+  dout(10) << __func__ << ": canceling recovery op for obj " << hoid
+	   << dendl;
+  assert(recovery_ops.count(hoid));
+  recovery_ops.erase(hoid);
+
+  list<pg_shard_t> fl;
+  for (map<pg_shard_t, int>::iterator i = res.errors.begin();
```
tchaikov: could use range-based loop.

athanatos:
I think this looks ok to me

dachary:
<pre>
</pre>

tchaikov:
lgtm also.

tchaikov:
@dzafman shall we backport this to jewel?

smithfarm:
@dzafman Ping re: jewel backport. Users are asking about it.

PR #9305, opened 2016-05-25
osd/ReplicatedPG: minor fixes and cleanups

xiexingguo:


xiexingguo:
@liewegas Move the osd relevant changes into this pr. 
Really simple changes, though.

xiexingguo:
@liewegas  And this one:-)

liewegas:
Looks good, except I'm not sure what the motivation for the max_name_len patch is.  I'm not sure it's a good idea not to have a well-defined limit.  Backends may vary between OSD, so i'm not sure when we would want to have behavior be ambiguous...

xiexingguo:
Thanks for the explanation. Dropped that change.

PR #9308, opened 2016-05-25
common: should give a value before use cpp_strerror

songbaisen:
common: should give a value before use cpp_strerror


xiexingguo:
Most of them are not right, as they are not POSIX syscalls and may return a customized error code instead, under which case **errno** may not be correctly set.

songbaisen:
@xiexingguo  Oh learn it. I thought all need set value before.Thank you! :ice_cream:

PR #9310, opened 2016-05-25
rgw/put: fix the 'ret' value of RGWPutObj

Yan-waller:
May be we should set `ret` value to `-ERR_LENGTH_REQUIRED`
if that `ofs != s->content_length` happens.

### DIFF
--- a/src/rgw/rgw_op.cc
+++ b/src/rgw/rgw_op.cc
@@ -2478,7 +2478,7 @@ void RGWPutObj::execute()
   } while (len > 0);
 
   if (!chunked_upload && ofs != s->content_length) {
-    op_ret = -ERR_REQUEST_TIMEOUT;
+    op_ret = -ERR_LENGTH_REQUIRED;
     goto done;
   }
   s->obj_size = ofs;


Yan-waller:
retest this please

oritwas:
I don't think ERR_LENGTH_REQUIRED is the correct error

yehudasa:
@Yan-waller @oritwas Yeah, I don't believe it's ERROR_LENGTH_REQUIRED. We check explicitly that it's not chunked upload, so length will be specified. Note that we reach that point if we got less data than in the content-length header, which means either that the user disconnected (not sure if it'll reach this point if that happens, need to look at the code) -- which means that it doesn't matter what we return, or that when trying to fetch more data we came in short, and since we block on the reads it actually means that we timed out reading.

Yan-waller:
It seems that I made a mistake,  I just misunderstood the errorcode `ERR_REQUEST_TIMEOUT` , sorry for that and many thanks to you. @oritwas @yehudasa

yehudasa:
@Yan-waller thanks, closing

PR #9311, opened 2016-05-25
[RFC] shutdown PGs before osd aborting

yangdongsheng:
When we meet an IO error in objectstore, we will assert(0) to abort our osd. But there is a problem in this case, that there have to be a time window to let heartbeat to detect this osd is down. In this window, the writings from user are blocked. Like below:
```
bench-write  io_size 4096 io_threads 16 bytes 1073741824 pattern sequential
  SEC       OPS   OPS/SEC   BYTES/SEC
    1      2898   2919.88  11959848.38
    2      6144   3082.58  12626249.21
    4      7016   1695.09  6943094.20
    5      7031   1226.59  5024098.20
    6      7046   1158.64  4745802.87
    7      9113   1034.27  4236364.20
    8     11897    958.92  3927740.74
    9     13234   1217.30  4986075.15
   10     14281   1687.51  6912041.28
   11     15326   1688.41  6915719.55
   12     17399   1573.81  6446337.95
   13     18587   1338.00  5480467.11
2016-05-25 03:18:53.431546 7f76907f8700  0 -- 192.168.1.119:0/3735375193 >> 192.168.1.119:6800/17940 pipe(0x7f76800053d0 sd=12 :0 s=1 pgs=0 cs=0 l=1 c=0x7f7680003070).fault
   37     19476    223.10  913799.40
   38     21556    259.58  1063256.69
   39     23537    293.26  1201204.65

```
Okey, that would be reduced by setting the config of osd_heartbeat_interval and osd_heartbeat_grace. But that time window will not disappear. The minimum number is about 5s in my testing.

That would cause the service to be paused, that's not acceptable in some cases. This patch will mark osd to be DOWN and shutdown related PGs before aborting. Then the pause problem is solved.

### DIFF
--- a/src/os/ObjectStore.h
+++ b/src/os/ObjectStore.h
@@ -94,6 +94,7 @@ const int SKIP_MOUNT_OMAP = 1 << 1;
 class ObjectStore {
 protected:
   string path;
+  Context* on_abort = NULL;
 
 public:
   /**
@@ -111,6 +112,10 @@ public:
 			     const string& data,
 			     const string& journal,
 			     osflagbits_t flags = 0);
+
+  void set_on_abort(Context* c) {
+    on_abort = c;
+  }
 
   /**
    * probe a block device to learn the uuid of the owning OSD
--- a/src/os/bluestore/BlueStore.cc
+++ b/src/os/bluestore/BlueStore.cc
@@ -3709,12 +3709,18 @@ void BlueStore::_txc_state_proc(TransContext *txc)
 	if (g_conf->bluestore_sync_submit_transaction) {
 	  _txc_update_fm(txc);
 	  int r = db->submit_transaction(txc->t);
-	  assert(r == 0);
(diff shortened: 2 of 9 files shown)


yangdongsheng:
@liewegas could you take a look, thanx

yangdongsheng:
@liewegas any opinion on this idea？

liewegas:
https://github.com/ceph/ceph/pull/8558 will address this case in a simpler way, I think.

liewegas:
We discussed this idea a few times and eventually decided that something more like #8558 is simpler and less prone to bugs.  We also discussed the idea of having a separate process operate as a watchdog that could send an immediate "osd is dead" if it sees the process go away.  I suspect, though, that just looking for ECONNREFUSED is going to cover all these cases in a simpler way.  Can you think of any cases that it won't cover?

Thanks!

PR #9321, opened 2016-05-25
Add powercrc32

minmfang:
Use power8 crc32 code to accelerate crc32c calculation performance, more information about power8 crc32c can reference the following link:

https://github.com/antonblanchard/crc32-vpmsum


tchaikov:
@minmfang could you squash these two commit into a single one? and put your "Signed-off-by" line at the end of the commit message?

also,  could you prefix the title of your commit message with the subcomponent your are changing ? see https://github.com/ceph/ceph/blob/master/SubmittingPatches.rst#preparing-and-sending-patches

PR #9323, opened 2016-05-25
MDSMonitor.cc: fix mdsmap.<namespace> subscriptions

ukernel:

### DIFF
--- a/src/mon/MDSMonitor.cc
+++ b/src/mon/MDSMonitor.cc
@@ -2411,22 +2411,22 @@ void MDSMonitor::check_subs()
 {
   std::list<std::string> types;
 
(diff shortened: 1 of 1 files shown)


jcsp:
Looks good, what was noticing this bug, #8386?

jcsp:
It's weird that TestMultiFilesystems.test_clients was passing in spite of this bug, worth investigating

ukernel:
I found that if mount non-default namespae, client does not reconnect to mds when mds restarts

jcsp:
http://pulpito.front.sepia.ceph.com/jspray-2016-05-25_19:00:36-fs-wip-jcsp-testing-20160526---basic-mira/

PR #9324, opened 2016-05-25
rgw: properly handle initial slashes in SLO's segment path.

rzarzynski:
Fixes: http://tracker.ceph.com/issues/16015

-- 

CC: @yehudasa.

### DIFF
--- a/src/rgw/rgw_op.cc
+++ b/src/rgw/rgw_op.cc
@@ -1027,13 +1027,26 @@ int RGWGetObj::handle_slo_manifest(bufferlist& bl)
 
   for (const auto& entry : slo_info.entries) {
     const string& path = entry.path;
-    const size_t pos = path.find('/', 1); /* skip first / */
-    if (pos == string::npos) {
+
+    /* If the path starts with slashes, strip them all. */
+    const size_t pos_init = path.find_first_not_of('/');
+    /* According to the documentation of std::string::find following check
+     * is not necessary as we should get the std::string::npos propagation
+     * here. This might be true with the accuracy to implementation's bugs.
+     * See following question on SO:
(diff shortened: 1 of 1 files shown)


src/rgw/rgw_op.cc
```diff
@@ -1027,13 +1027,15 @@ int RGWGetObj::handle_slo_manifest(bufferlist& bl)
 
   for (const auto& entry : slo_info.entries) {
     const string& path = entry.path;
-    const size_t pos = path.find('/', 1); /* skip first / */
-    if (pos == string::npos) {
+    /* If the path starts with slashes, strip them all. */
+    const size_t pos_init = path.find_first_not_of('/');
+    const size_t pos_sep = path.find('/', pos_init);
```
yehudasa: @rzarzynski need to check that pos_init != string::npos here

rzarzynski: @yehudasa: not sure whether I understood you correctly. I think that passing `std::string::npos` to `find` is safe - unlike in case of `substr` when we could get `std::out_of_range` exception. [The documentation]() says about the `pos` argument of `find`:
> If this is greater than the string length, the function never finds matches.

In C++11 the `char` variant also has the no-throw guarantee:
> character (4)	 `size_t find (char c, size_t pos = 0) const noexcept;`

mattbenjamin:
that seems a bit strange to me too, to be honest

src/rgw/rgw_op.cc
rzarzynski: This might be true with the accuracy to implementation's bugs (see [this question on StackOverflow](http://stackoverflow.com/questions/1011790/why-does-stdstring-findtext-stdstringnpos-not-return-npos)).

Pushed a reworked version with the extra check on board.

PR #9326, opened 2016-05-25
librbd: only mark async completion as done after callback

marcan:
This change in behavior was introduced in fde9f78.

cc @dillaman

### DIFF
--- a/src/librbd/AioCompletion.cc
+++ b/src/librbd/AioCompletion.cc
@@ -31,7 +31,7 @@ namespace librbd {
   int AioCompletion::wait_for_complete() {
     tracepoint(librbd, aio_wait_for_complete_enter, this);
     lock.Lock();
-    while (!done)
+    while (state != STATE_COMPLETE)
       cond.Wait(lock);
     lock.Unlock();
     tracepoint(librbd, aio_wait_for_complete_exit, 0);
@@ -95,7 +95,7 @@ namespace librbd {
       ictx->journal->commit_io_event(journal_tid, rval);
     }
 
-    done = true;
+    state = STATE_CALLBACK;
     if (complete_cb) {
       lock.Unlock();
       complete_cb(rbd_comp, complete_arg);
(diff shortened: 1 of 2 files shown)


dillaman:
I remember why I had to move it -- there is a rare possible race condition with completing no-op IO.  How about we switch ```AioCompletion::done``` to be an enum like:

```
enum State {
  STATE_PENDING,
  STATE_CALLBACK,
  STATE_COMPLETE
};
```

The ```AioCompletion::wait_for_complete``` would loop until the state is COMPLETE. Other references to ```done``` would only check for ```state != STATE_PENDING```.  The ```AioCompletion::complete``` would set the state to CALLBACK before invoking the callbacks and would set it to COMPLETE before signaling the condition.

marcan:
Done, PTAL.

src/librbd/AioCompletion.h
```diff
@@ -46,7 +52,7 @@ namespace librbd {
   struct AioCompletion {
     Mutex lock;
     Cond cond;
-    bool done;
+    aio_state_t done;
```
dillaman: Minor: can you rename this to ```state```?

src/librbd/AioCompletion.cc
```diff
@@ -95,12 +95,13 @@ namespace librbd {
...
     }
 
-    done = true;
+    done = STATE_CALLBACK;
     if (complete_cb) {
       lock.Unlock();
       complete_cb(rbd_comp, complete_arg);
       lock.Lock();
     }
+    done = STATE_COMPLETE;
```
dillaman: Minor: I would move this to just above ```cond.Signal()``` below

src/librbd/AioCompletion.h
marcan: Done.

src/librbd/AioCompletion.cc
marcan: Done.

dillaman:
lgtm

yuriw:

PR #9329, opened 2016-05-25
[DNM] rbd: Create fewer empty objects

fullerdj:
In many cases, we unconditionally create empty objects (of size 0) when
performing layered writes. While functionally correct, this causes tools
like rbd du to space is allocated (since they conservatively assume all
objects are full). Detect empty writes and bail out early. This is
particularly helpful for operations like flatten, which create all
objects that might exist in an image.       

Fixes: http://tracker.ceph.com/issues/15028
Signed-off-by: Douglas Fuller dfuller@redhat.com

### DIFF
--- a/src/librbd/AioObjectRequest.h
+++ b/src/librbd/AioObjectRequest.h
@@ -45,6 +45,10 @@ namespace librbd {
       return !m_parent_extents.empty();
     }
 
+    virtual const char *get_op_type() const = 0;
+
+    virtual bool pre_object_map_update(uint8_t *new_state) = 0;
+
   protected:
     bool compute_parent_extents();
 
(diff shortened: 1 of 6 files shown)


src/librbd/AioObjectRequest.h
```diff
@@ -45,6 +45,10 @@ namespace librbd {
       return !m_parent_extents.empty();
     }
 
+    virtual const char *get_op_type() const = 0;
+
+    virtual bool pre_object_map_update(uint8_t *new_state) = 0;
+
```
fullerdj: This is needed for CopyupRequest to handle generic object requests. We could make these private and declare CopyupRequest to be a friend if that is preferable.

liewegas:
@fullerdj ping?

vshankar:
@fullerdj If you're not working on this PR anymore (since you closed it), would you mind if I get this PR started/open again and probably work on this?

fullerdj:
Not at all.

dachary:
@vshankar you need to open your own pull request to continue the work

vshankar:
@dachary my bad, thanks for closing.

PR #9332, opened 2016-05-25
Make Rados ready for PyPI

onyb:
Test package is hosted on http://testpypi.python.org/pypi/rados

**Steps to build from source:**
- First install the development headers and libraries of Python and Rados.

    ```
    $ sudo apt-get install python-dev librados-dev
    ```
- Install Cython

    ```
    $ sudo pip install cython
    ```

- Install Rados bindings

    ```
    $ cd src/pybind/rados
    $ sudo python setup.py install
    ```

- (optional) Install Rados bindings from pre-cythonized source
You don't need to have Cython installed in this case.

    ```
    $ cd src/pybind/rados
    $ sudo python setup.py install --without-cython
    ```

- Building Rados bindings from Ceph top-level

    ```
    $ PYTHON=python3 ./configure
    $ make
    ```

Note that `virtualenv` and `pyvenv` are also supported.

**Steps to run the test package:**
```
$ sudo pip install -i https://testpypi.python.org/pypi rados
```


ktdreyer:
Want to link to http://tracker.ceph.com/issues/5900 ?

jcsp:
Sorry if this has already been discussed, but why is it useful to have our rados modules on pypi?  They are not versioned independently of librados.so, so unless someone happens to have exactly the same versions of both, things are liable to not work.

Given that someone will always need native (librados) packages when using the python modules, what is the advantage of getting the python modules from pypi instead of getting the python modules from the packages that we provide?

onyb:
@jcsp: I am not very sure how the Debian package `python-rados` has been packaged, but I think it doesn't contain the new Cython bindings, and installs only for Python 2.7.

Advantage of using `pip` is that we can ship Cython extensions which will be cythonized and compiled according to the Python version `pip` was launched with. This is also important because bindings compiled on Python 2 won't work on Python 3, and vice-versa. Another reason to go for `pip` is that it allows installing `rados` inside virtual environments.

Maybe @jdurgin will have something to add.

onyb:
@ktdreyer Updated issue [#5900](http://tracker.ceph.com/issues/5900) with a link to this PR.

jdurgin:
@jcsp the version requirement isn't exact, as-is older python bindings will continue to work with new libraries, they'll only fail when newer python bindings are used with older libraries which don't have new functions that the python bindings use.

The drive for this is to play nicer with python packaging for developers, and like @onyb mentioned using different versions of python or the bindings is one aspect. It allows virtualenvs with no system site packages. Compiling just the c generated by cython isn't expensive, so it seems like a good tradeoff to me, rather than trying to bundle librados or compile it in the python package.

jdurgin:
Is it necessary to commit the generated .c file? Could that be created by the usual build process before creating the python package?

src/pybind/rados/setup.py
```diff
@@ -1,51 +1,187 @@
...
-    },
+        ], build_dir=os.environ.get("CYTHON_BUILD_DIR", None)
+    ),
+    classifiers=[
+        'Intended Audience :: Developers',
+        'Intended Audience :: System Administrators',
+        'License :: OSI Approved :: GNU Lesser General Public License v2 or later (LGPLv2+)',
+        'Operating System :: POSIX :: Linux',
+        'Programming Language :: Cython',
+        'Programming Language :: Python :: 2.6',
```
jdurgin: can leave off 2.6 - no distros ceph supports use it, and this setup.py won't work with it (subprocess.check_output() isn't there until 2.7).

Is it necessary to specify all the minor versions rather than just python 2 and python 3?

src/pybind/rados/setup.py
```diff
@@ -1,51 +1,187 @@
...
+        """)
+        fp.write(dummy_prog)
+
+    compiler = new_compiler()
+    customize_compiler(compiler)
+
+    if set(['MAKEFLAGS', 'MFLAGS', 'MAKELEVEL']).issubset(set(os.environ.keys())):
+        # The setup.py has been invoked by a top-level Ceph make.
+        # Set the appropriate CFLAGS and LDFLAGS
+        CEPH_ROOT = subprocess.check_output(['git', 'rev-parse', '--show-toplevel']).strip().decode('utf-8')
```
jdurgin: for out-of-tree builds .libs won't be in the src dir. ceph may also be built from a tarball, so git shouldn't be used to determine the build dir.

The location of this setup.py file should tell us where to find the include dir though.
For the build output dir, I'm not certain of the best way to find it. Perhaps having the makefile preprocess a setup.py.in into setup.py to add the location of .libs.

src/pybind/rados/setup.py
```diff
@@ -1,51 +1,187 @@
...
 
-def get_version():
+# PEP 440 versioning of the Rados package on PyPI
+# Bump this version, after every changeset
+# NOTE: This version is not the same as get_ceph_version()
+
+__version__ = '0.0.1'
+
+
+def get_ceph_version():
```
jdurgin: looks like get_ceph_version() is unused after this patchset

jdurgin:
@alfredodeza perhaps you'd like to take a look at this too

src/pybind/rados/setup.py
onyb: I idea was to be explicit about the versions supported. I kept Python 2.6+ and 3.2+ since it was supported by the bindings, but it is seems logical to drop 2.6, 3.2, and 3.3.

src/pybind/rados/setup.py
onyb: This `get_ceph_version()` didn't return a PEP 440 compliant version, so this needs discussion. Perhaps it can be a number prefixed by Ceph's version, if that makes sense.

src/pybind/rados/setup.py
onyb: @jdurgin I will make the changes. Thanks.

Regarding the build output dir, it can be an environment variable exported just like `CYTHON_BUILD_DIR`.

onyb:
@jdurgin The motivation behind including the cythonized file is to allow users to do a `python setup.py install --without-cython` if they don't want to have Cython as a dependency. Said that, *not* committing the file is an option, but will only work for `pip install`, and not `python setup.py install`.

I can push a commit to exclude `rados.c` from source tree, and only include it in the PyPI source distribution. I will also drop the `--without-cython` option, and instead use the already cythonized .c file only if Cython is not available.

jdurgin:
@onyb that sounds good wrt to rados.c

src/pybind/rados/setup.py
jdurgin: On 05/27/2016 02:50 AM, Anirudha Bose wrote:
> In src/pybind/rados/setup.py
> <https://github.com/ceph/ceph/pull/9332#discussion_r64882338>:
>
>>              )
>> -    ], build_dir=os.environ.get("CYTHON_BUILD_DIR", None)),
>> -    cmdclass={
>> -        "egg_info": EggInfoCommand,
>> -    },
>> +        ], build_dir=os.environ.get("CYTHON_BUILD_DIR", None)
>> +    ),
>> +    classifiers=[
>> +        'Intended Audience :: Developers',
>> +        'Intended Audience :: System Administrators',
>> +        'License :: OSI Approved :: GNU Lesser General Public License v2 or later (LGPLv2+)',
>> +        'Operating System :: POSIX :: Linux',
>> +        'Programming Language :: Cython',
>> +        'Programming Language :: Python :: 2.6',
>
> I idea was to be explicit about the versions supported. I kept Python
> 2.6+ and 3.2+ since it was supported by the bindings, but it is seems
> logical to drop 2.6, 3.2, and 3.3.

Sounds good

src/pybind/rados/setup.py
jdurgin: On 05/27/2016 02:56 AM, Anirudha Bose wrote:
> In src/pybind/rados/setup.py
> <https://github.com/ceph/ceph/pull/9332#discussion_r64882963>:
>
>>
>> -def get_version():
>> +# PEP 440 versioning of the Rados package on PyPI
>> +# Bump this version, after every changeset
>> +# NOTE: This version is not the same as get_ceph_version()
>> +
>> +__version__ = '0.0.1'
>> +
>> +
>> +def get_ceph_version():
>
> This |get_ceph_version()| didn't return a PEP 440 compliant version, so
> this needs discussion. Perhaps it can be a number prefixed by Ceph's
> version, if that makes sense.

I think it would make sense to use a separate version number. librados,
librbd, etc. already have a separate version that they report with
their {rados,rbd}_version calls, and the python bindings may also
change independent of the c libs.

jdurgin: if no one objects to a separate version, let's start it off at 2.0.0 to match the librados so number

jdurgin:
built nicely, and it's qutie easy to reconfigure + rebuild for a new python version. I pushed this as wip-rados-pypi so we can see how it builds on different platforms (http://ceph.com/gitbuilder.cgi)

jdurgin:
We will want to make this work with cmake too (I think that could be a later PR though)
@alimaredia @tchaikov any comments on this?

onyb:
@jdurgin I see builds in Gitbuilder fail with an `error: option --single-version-externally-managed not recognized`. The reason is that the `setup.py` which uses only `distutils` (not `setuptools`), doesn't accept that option. Switching back to using `setuptools` should fix the problem.

jdurgin:
just to clean things up a bit, could you add a subsystem: prefix to your commit message subjects? you can take a look at git log <path> for some examples, I think most would be pybind here.

Also could you rebase to squash the later fixup commits into the original commits that introduced the change, so e.g. rados.c is never in the tree?

jdurgin:
I think the only remaining piece would be to set the rados python module version number. @ktdreyer  are you ok with the python bindings having versions independent from ceph's version, since the usual form doesn't work for python versions?

ktdreyer:
> @ktdreyer are you ok with the python bindings having versions independent from ceph's version, since the usual form doesn't work for python versions?

Sure, I don't mind, but now you've got me curious what doesn't work in particular?

onyb:
@ktdreyer Versions like `10.2.0-2215-gfb2667f` as defined in `CEPH _GIT_NICE_VER` are not valid PEP 440 versions. A version like `10.2.0-r2215` could be normalized to `10.2.0.post2215` (supported), but when there is a commit hash, it becomes impossible to order them reliably. Moreover, this version is not related to Rados pybind in any way.

ktdreyer:
that's fine

tchaikov:
> any comments on this?

@jdurgin will check this changeset later on.

alimaredia:
@tchaikov will you take care of the ramifications to CMake of this PR?

tchaikov:
@alimaredia i will.

onyb:
Please head over to pull request https://github.com/ceph/ceph/pull/9833

PR #9333, opened 2016-05-26
rbd: initial operations for consistency groups

VictorDenisov:
This pull request is specifically for consistency group's operations: create consistency group, remove consistency group, list consistency groups.

Link to the etherpad: http://pad.ceph.com/p/consistency_groups

### DIFF
--- a/.organizationmap
+++ b/.organizationmap
@@ -247,6 +247,7 @@ Mirantis <contact@mirantis.com> Igor Fedotov <ifedotov@mirantis.com>
 Mirantis <contact@mirantis.com> Kiseleva Alyona <akiselyova@mirantis.com>
 Mirantis <contact@mirantis.com> Mykola Golub <mgolub@mirantis.com>
 Mirantis <contact@mirantis.com> Radoslaw Zarzynski <rzarzynski@mirantis.com>
+Mirantis <contact@mirantis.com> Victor Denisov <denisovenator@gmail.com>
 MIT Computer Science and Artificial Intelligence Laboratory <webmaster@csail.mit.edu> Stephen Jahl <stephenjahl@gmail.com>
 MSys Technologies <contact@msystechnologies.com> Rajesh Nambiar <rajesh.n@msystechnologies.com>
 Nebula <info@nebula.com> Anton Aksola <anton.aksola@nebula.fi>
--- a/src/cls/rbd/cls_rbd.cc
+++ b/src/cls/rbd/cls_rbd.cc
@@ -132,12 +132,18 @@ cls_method_handle_t h_mirror_image_status_get;
 cls_method_handle_t h_mirror_image_status_list;
 cls_method_handle_t h_mirror_image_status_get_summary;
 cls_method_handle_t h_mirror_image_status_remove_down;
+cls_method_handle_t h_group_create;
+cls_method_handle_t h_group_dir_list;
+cls_method_handle_t h_group_dir_add;
+cls_method_handle_t h_group_dir_remove;
 
 #define RBD_MAX_KEYS_READ 64
 #define RBD_SNAP_KEY_PREFIX "snapshot_"
 #define RBD_DIR_ID_KEY_PREFIX "id_"
 #define RBD_DIR_NAME_KEY_PREFIX "name_"
 #define RBD_METADATA_KEY_PREFIX "metadata_"
+
+#define GROUP_SNAP_SEQ "snap_seq"
 
 static int snap_read_header(cls_method_context_t hctx, bufferlist& bl)
 {
@@ -4135,6 +4141,212 @@ int mirror_image_status_remove_down(cls_method_context_t hctx, bufferlist *in,
   return 0;
 }
 
+/**
+ * Initialize the header with basic metadata.
+ * Everything is stored as key/value pairs as omaps in the header object.
+ *
+ * Input:
+ * none
+ *
+ * Output:
+ * @return 0 on success, negative error code on failure
+ */
+int group_create(cls_method_context_t hctx, bufferlist *in, bufferlist *out)
+{
+  bufferlist snap_seqbl;
+  uint64_t snap_seq = 0;
+  ::encode(snap_seq, snap_seqbl);
+  int r = cls_cxx_map_set_val(hctx, GROUP_SNAP_SEQ, &snap_seqbl);
+  if (r < 0)
+    return r;
+
+  return 0;
+}
+
+/**
+ * List consistency groups from the directory.
+ *
+ * Input:
+ * @param start_after (std::string)
+ * @param max_return (int64_t)
+ *
+ * Output:
+ * @param map of consistency groups (name, id)
+ * @return 0 on success, negative error code on failure
+ */
+int group_dir_list(cls_method_context_t hctx, bufferlist *in, bufferlist *out)
+{
+  string start_after;
+  uint64_t max_return;
+
+  try {
+    bufferlist::iterator iter = in->begin();
+    ::decode(start_after, iter);
+    ::decode(max_return, iter);
+  } catch (const buffer::error &err) {
+    return -EINVAL;
+  }
+
+  int max_read = RBD_MAX_KEYS_READ;
+  int r = max_read;
+  map<string, string> groups;
+  string last_read = dir_key_for_name(start_after);
+
+  while (r == max_read && groups.size() < max_return) {
+    map<string, bufferlist> vals;
+    CLS_LOG(20, "last_read = '%s'", last_read.c_str());
+    r = cls_cxx_map_get_vals(hctx, last_read, RBD_DIR_NAME_KEY_PREFIX,
+			     max_read, &vals);
+    if (r < 0) {
+      CLS_ERR("error reading directory by name: %s", cpp_strerror(r).c_str());
+      return r;
+    }
+
+    for (pair<string, bufferlist> val: vals) {
+      string id;
+      bufferlist::iterator iter = val.second.begin();
+      try {
+	::decode(id, iter);
+      } catch (const buffer::error &err) {
+	CLS_ERR("could not decode id of consistency group '%s'", val.first.c_str());
+	return -EIO;
+      }
+      CLS_LOG(20, "adding '%s' -> '%s'", dir_name_from_key(val.first).c_str(), id.c_str());
+      groups[dir_name_from_key(val.first)] = id;
+      if (groups.size() >= max_return)
+	break;
+    }
+    if (!vals.empty()) {
+      last_read = dir_key_for_name(groups.rbegin()->first);
+    }
+  }
+
+  ::encode(groups, *out);
+
+  return 0;
+}
+
+/**
+ * Add a consistency group to the directory.
+ *
+ * Input:
+ * @param name (std::string)
+ * @param id (std::string)
+ *
+ * Output:
+ * @return 0 on success, negative error code on failure
+ */
+int group_dir_add(cls_method_context_t hctx, bufferlist *in, bufferlist *out)
+{
+  int r = cls_cxx_create(hctx, false);
+
+  if (r < 0) {
+    CLS_ERR("could not create consistency group directory: %s",
+	    cpp_strerror(r).c_str());
+    return r;
+  }
+
+  string name, id;
+  try {
+    bufferlist::iterator iter = in->begin();
+    ::decode(name, iter);
+    ::decode(id, iter);
+  } catch (const buffer::error &err) {
+    return -EINVAL;
+  }
+
+  if (!name.size() || !is_valid_id(id)) {
+    CLS_ERR("invalid consistency group name '%s' or id '%s'",
+	    name.c_str(), id.c_str());
+    return -EINVAL;
+  }
+
+  CLS_LOG(20, "group_dir_add name=%s id=%s", name.c_str(), id.c_str());
+
+  string tmp;
+  string name_key = dir_key_for_name(name);
+  string id_key = dir_key_for_id(id);
+  r = read_key(hctx, name_key, &tmp);
+  if (r != -ENOENT) {
+    CLS_LOG(10, "name already exists");
+    return -EEXIST;
+  }
+  r = read_key(hctx, id_key, &tmp);
+  if (r != -ENOENT) {
+    CLS_LOG(10, "id already exists");
+    return -EBADF;
+  }
+  bufferlist id_bl, name_bl;
+  ::encode(id, id_bl);
+  ::encode(name, name_bl);
+  map<string, bufferlist> omap_vals;
+  omap_vals[name_key] = id_bl;
+  omap_vals[id_key] = name_bl;
+  return cls_cxx_map_set_vals(hctx, &omap_vals);
+}
+
+/**
+ * Remove a consistency group from the directory.
+ *
+ * Input:
+ * @param name (std::string)
+ * @param id (std::string)
+ *
+ * Output:
+ * @return 0 on success, negative error code on failure
+ */
+int group_dir_remove(cls_method_context_t hctx, bufferlist *in, bufferlist *out)
+{
+  string name, id;
+  try {
+    bufferlist::iterator iter = in->begin();
+    ::decode(name, iter);
+    ::decode(id, iter);
+  } catch (const buffer::error &err) {
+    return -EINVAL;
+  }
+
+  CLS_LOG(20, "group_dir_remove name=%s id=%s", name.c_str(), id.c_str());
+
+  string stored_name, stored_id;
+  string name_key = dir_key_for_name(name);
+  string id_key = dir_key_for_id(id);
+
+  int r = read_key(hctx, name_key, &stored_id);
+  if (r < 0) {
+    if (r != -ENOENT)
+      CLS_ERR("error reading name to id mapping: %s", cpp_strerror(r).c_str());
+    return r;
+  }
+  r = read_key(hctx, id_key, &stored_name);
+  if (r < 0) {
+    if (r != -ENOENT)
+      CLS_ERR("error reading id to name mapping: %s", cpp_strerror(r).c_str());
+    return r;
+  }
+
+  // check if this op raced with a rename
+  if (stored_name != name || stored_id != id) {
+    CLS_ERR("stored name '%s' and id '%s' do not match args '%s' and '%s'",
+	    stored_name.c_str(), stored_id.c_str(), name.c_str(), id.c_str());
+    return -ESTALE;
+  }
+
+  r = cls_cxx_map_remove_key(hctx, name_key);
+  if (r < 0) {
+    CLS_ERR("error removing name: %s", cpp_strerror(r).c_str());
+    return r;
+  }
+
+  r = cls_cxx_map_remove_key(hctx, id_key);
+  if (r < 0) {
+    CLS_ERR("error removing id: %s", cpp_strerror(r).c_str());
+    return r;
+  }
+
+  return 0;
+}
+
 void __cls_init()
 {
   CLS_LOG(20, "Loaded rbd class!");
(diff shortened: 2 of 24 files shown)


dillaman:
Minor: In the future, I would structure similar PRs in themed/grouped commits. For example, I would have one commit that adds new methods to "cls_rbd" and "cls_rbd_client" and associated unit tests (commit message prefix cls_rbd: xyz); another commit to add the C/C++ bindings (commit message prefix librbd: xyz); another commit to add the Python bindings (commit message prefix pybind: xyz); and the last commit to update the org map

.organizationmap
```diff
@@ -247,6 +247,7 @@ Mirantis <contact@mirantis.com> Igor Fedotov <ifedotov@mirantis.com>
 Mirantis <contact@mirantis.com> Kiseleva Alyona <akiselyova@mirantis.com>
 Mirantis <contact@mirantis.com> Mykola Golub <mgolub@mirantis.com>
 Mirantis <contact@mirantis.com> Radoslaw Zarzynski <rzarzynski@mirantis.com>
+Mirantis <contact@mirantis.com> Victor Denisov <denisovenator@gmail.com>
```
dillaman: Minor: update your commit title to "mailmap: Victor Denisov affiliation" (seems to be the style)

src/include/rbd/cg_types.h
```diff
@@ -0,0 +1,25 @@
...
+ * modify it under the terms of the GNU Lesser General Public
+ * License version 2.1, as published by the Free Software
+ * Foundation.  See file COPYING.
+ *
+ */
+
+#ifndef CEPH_CG_TYPES_H
+#define CEPH_CG_TYPES_H
+
+#define CG_SNAP_SEQ "snap_seq"
```
dillaman: Minor: probably just a static const within ```cls_rbd.cc```

src/include/rbd/cg_types.h
```diff
@@ -0,0 +1,25 @@
+// -*- mode:C++; tab-width:8; c-basic-offset:2; indent-tabs-mode:t -*-
```
dillaman: These don't appear to be needed as part of the end-user librbd API.  Therefore, can we move these to "include/rbd_types.h"?  Also, all new header files need to be added to the Makefile*.am file under the noinst_HEADERS list.

src/cls/rbd/cls_rbd.cc
```diff
@@ -132,6 +133,7 @@ cls_method_handle_t h_mirror_image_status_get;
 cls_method_handle_t h_mirror_image_status_list;
 cls_method_handle_t h_mirror_image_status_get_summary;
 cls_method_handle_t h_mirror_image_status_remove_down;
+cls_method_handle_t h_create_cg;
```
dillaman: Since we are going to have a bunch of consistency group-related methods, can we name them like "cg_[optional-subtype}_]{verb}" format?  e.g. cg_create, cg_dir_list, etc

src/cls/rbd/cls_rbd.cc
```diff
@@ -4132,6 +4134,25 @@ int mirror_image_status_remove_down(cls_method_context_t hctx, bufferlist *in,
...
+ *
+ * Output:
+ * @return 0 on success, negative error code on failure
+ */
+int create_cg(cls_method_context_t hctx, bufferlist *in, bufferlist *out)
+{
+  bufferlist snap_seqbl;
+  uint64_t snap_seq = 0;
+  ::encode(snap_seq, snap_seqbl);
+  int r = cls_cxx_map_set_val(hctx, CG_SNAP_SEQ, &snap_seqbl);
```
dillaman: What is the plan for the snap sequence that you are initializing to zero here?  Is this going to be used as an id when creating cg snapshots?

src/cls/rbd/cls_rbd.cc
```diff
@@ -4132,6 +4134,25 @@ int mirror_image_status_remove_down(cls_method_context_t hctx, bufferlist *in,
   return 0;
 }
 
+/**
+ * Initialize the header with basic metadata.
+ * Everything is stored as key/value pairs as omaps in the header object.
+ *
+ * Output:
```
dillaman: Minor: document "Input: none"

src/librbd/internal.cc
```diff
@@ -3513,4 +3513,143 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force,
...
+    uint32_t extra;
+    int remove_r;
+    string id, id_cg, header_oid;
+
+    CephContext *cct = (CephContext *)io_ctx.cct();
+
+    Rados rados(io_ctx);
+    uint64_t bid = rados.get_instance_id();
+
+    id_cg = util::id_cg_name(cg_name);
```
dillaman: I have to be honest, I never saw the purpose of the RBD image v2 id object.  We should add the CG to the directory by providing a name and "unique" id. If the name or id is in-use, the exec call will fail with -EEXISTS/-ESTALE.  Once we have it recorded in the directory, so long as we have "CG name"->"CG id" and "CG id"->"CG name" mappings (see RBD image directory), we don't need the "cg_id.<CG name>" object and on rename we only have to update the directory and not the directory and the id object.

jdurgin: The reason for the separate id object was to make availability of the directory object unnecessary for accessing image. In practice I don't think it's mattered.

src/cls/rbd/cls_rbd.cc
```diff
@@ -4132,6 +4137,210 @@ int mirror_image_status_remove_down(cls_method_context_t hctx, bufferlist *in,
...
+    }
+
+    for (map<string, bufferlist>::iterator it = vals.begin();
+	 it != vals.end(); ++it) {
+      string id;
+      bufferlist::iterator iter = it->second.begin();
+      try {
+	::decode(id, iter);
+      } catch (const buffer::error &err) {
+	CLS_ERR("could not decode id of consistency group '%s'", it->first.c_str());
```
dillaman: Minor: printing the key but log message calls it the consistency group name

src/cls/rbd/cls_rbd.cc
```diff
@@ -4132,6 +4137,210 @@ int mirror_image_status_remove_down(cls_method_context_t hctx, bufferlist *in,
...
+ * Input:
+ * @param name (std::string)
+ * @param id (std::string)
+ *
+ * Output:
+ * @return 0 on success, negative error code on failure
+ */
+int dir_add_cg(cls_method_context_t hctx, bufferlist *in, bufferlist *out)
+{
+  int r = cls_cxx_create(hctx, false);
```
dillaman: Shouldn't have to "create" the object -- within a class method, all modifications will only be applied after the method successfully completes / is committed to disk.

src/include/rbd/librbd.h
```diff
@@ -641,6 +641,11 @@ CEPH_RBD_API int rbd_mirror_image_get_status(rbd_image_t image,
                                              rbd_mirror_image_status_t *mirror_image_status,
                                              size_t status_size);
 
+// RBD consistency groups support functions
+CEPH_RBD_API int rbd_create_cg(rados_ioctx_t p, const char *name);
```
dillaman: Mixed feelings about "cg" naming style in the API since it really isn't self-describing.  Probably can't justify a name as long as "rbd_consistency_group_snap_remove" --- but maybe "rbd_group_XYZ"?

src/librbd/internal.cc
```diff
@@ -3513,4 +3513,143 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force,
       ictx->perfcounter->inc(l_librbd_readahead_bytes, readahead_length);
     }
   }
+
+  // Consistency groups functions
+
+  int create_cg(librados::IoCtx& io_ctx, const char *cg_name)
+  {
+    ostringstream bid_ss;
```
dillaman: Declare local variables with as tight of scope as possible and initialize them with sane values at the point of declaration if needed.

src/librbd/internal.cc
```diff
@@ -3513,4 +3513,143 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force,
...
+    r = cls_client::dir_remove_cg(&io_ctx, CG_DIRECTORY,
+					 cg_name, cg_id);
+    if (r < 0 && r != -ENOENT) {
+      lderr(cct) << "error removing cg from directory" << dendl;
+      return r;
+    }
+
+    string header_oid = util::cg_header_name(cg_id);
+
+    r = io_ctx.remove(header_oid);
```
dillaman: Remove header before removing from the directory so that we don't leave an object behind in case of crash.

src/librbd/internal.cc
```diff
@@ -3513,4 +3513,143 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force,
...
+    do {
+      map<string, string> cgs;
+      r = cls_client::dir_list_cgs(&io_ctx, CG_DIRECTORY,
+			   last_read, max_read, &cgs);
+      if (r < 0) {
+        lderr(cct) << "error listing cg in directory: "
+                   << cpp_strerror(r) << dendl;
+        return r;
+      }
+      for (map<string, string>::const_iterator it = cgs.begin();
```
dillaman: Minor: you can use C++11 foreach loop

src/test/cli/rbd/help.t
```diff
@@ -110,6 +113,40 @@
...
+  Positional arguments
+    <cg-name>            Name of consistency group
+  
+  Optional arguments
+    -p [ --pool ] arg    pool name
+  
+  rbd help cg list
+  usage: rbd cg list [--pool <pool>] [--format <format>] [--pretty-format] 
+  
+  Dump list of consistency groups.
```
dillaman: Minor: "List rbd consistency groups"

src/test/cli/rbd/help.t
```diff
@@ -110,6 +113,40 @@
     --io-total arg       total size to write (in B/K/M/G/T)
     --io-pattern arg     write pattern (rand or seq)
   
+  rbd help cg create
+  usage: rbd cg create [--pool <pool>] 
+                       <cg-name>
```
dillaman: Should really support <cg-spec> where "cg-spec" is "[pool/]cg-name"

src/test/cls_rbd/test_cls_rbd.cc
```diff
@@ -84,10 +85,35 @@ class TestClsRbd : public ::testing::Test {
 
 };
 
+class TestClsRbdCg : public ::testing::Test {
```
dillaman: Can you just re-use the existing test harness for consistency group checks?

src/test/cls_rbd/test_cls_rbd.cc
```diff
@@ -1695,3 +1721,105 @@ TEST_F(TestClsRbd, mirror_image_status) {
...
+
+  ++it;
+  ASSERT_EQ(cg_id2, it->second);
+  ASSERT_EQ(cg_name2, it->first);
+}
+
+void can_add_cg_to_dir(librados::IoCtx ioctx, string cg_id, string cg_name) {
+  ASSERT_EQ(0, dir_add_cg(&ioctx, CG_DIRECTORY, cg_name, cg_id));
+
+  set<string> keys;
```
dillaman: Test dup name / dup id cases

src/test/cls_rbd/test_cls_rbd.cc
```diff
@@ -1695,3 +1721,105 @@ TEST_F(TestClsRbd, mirror_image_status) {
...
+  auto it = cgs.begin();
+  ASSERT_EQ(cg_id1, it->second);
+  ASSERT_EQ(cg_name1, it->first);
+
+  ++it;
+  ASSERT_EQ(cg_id2, it->second);
+  ASSERT_EQ(cg_name2, it->first);
+}
+
+void can_add_cg_to_dir(librados::IoCtx ioctx, string cg_id, string cg_name) {
```
dillaman: Minor: "can_add_cg_to_dir" -> "add_cg_to_dir"?

src/test/librbd/test_ConsistencyGroups.cc
```diff
@@ -0,0 +1,150 @@
...
+#include <vector>
+
+using namespace ceph;
+using namespace boost::assign;
+using namespace librbd::watch_notify;
+
+void register_test_consistency_groups() {
+}
+
+class TestLibCG : public ::testing::Test {
```
dillaman: Reuse existing test harness base class

src/tools/Makefile-client.am
```diff
@@ -33,6 +33,7 @@ rbd_SOURCES = \
 	tools/rbd/action/BenchWrite.cc \
 	tools/rbd/action/Children.cc \
 	tools/rbd/action/Clone.cc \
+	tools/rbd/action/ConsGrp.cc \
```
dillaman: Minor: would like consistent naming scheme.  What about just "group" throughout?

src/cls/rbd/cls_rbd.cc
VictorDenisov: The idea is the same as for images - keep the latest id we assigned to its snapshot. So far this code is only a place holder. Do you want me to remove it until we introduce snapshots?

dillaman: No need -- just wondering

src/include/rbd/librbd.h
VictorDenisov: This was my concern as well. Do you want to change only c bindings or across all the changes?

dillaman: Across all the APIs

src/librbd/internal.cc
VictorDenisov: So, do we want to get rid of cg_id.<name> object or cg_name.<id> object?

dillaman: I'd vote for getting rid of the ```cg_id.<NAME>``` object since we have the same information in the directory.  What's the ```cg_name.``` object?

VictorDenisov: I meant cg_header.98374. This is the object where we keep all the information about the consistency group. Maybe it makes sense to call it cg_header.cgname? Because then we can reach out to a consistency group without getting the directory.

dillaman: No, you would still want the header.  Naming it ```cg_header.<name>``` introduces race conditions if you crash during a rename.  I don't think looking up the id in the directory is a big deal -- you'd have to look it up in the id object anyway.

VictorDenisov: Ok. Then I'll keep cg_header as is.

src/test/cli/rbd/help.t
VictorDenisov: Frankly, I find the current design of cli ambiguous. What is the benefit of being able to specify group name as a positional parameter as well as a named parameter as well as pool using pool/image or using a named parameter? I definitely will implement it the way you want, just wondering.

src/test/cls_rbd/test_cls_rbd.cc
VictorDenisov: Existing test harness doesn't clean up after every test. It's convenient to have a clean pool before the test.

dillaman: Creating pools is very expensive, time-wise, when not using the unit test harness.  I can't think of a reason why you need unique pools.

src/test/cli/rbd/help.t
dillaman: I can't speak for the history, but for now it's based on inertia and staying consistent with the other commands.  If I could start again, I'd vote hyphens between command specs (e.g. cg-create vs cg create) and only named parameters.

src/test/cls_rbd/test_cls_rbd.cc
VictorDenisov: So, if you don't mind I would prefer to keep it.

src/test/librbd/test_ConsistencyGroups.cc
VictorDenisov: do you mean TestFixture class from test/librbd/test_fixture.cc?

dillaman: Yeah -- all tests except for the legacy TestLibRBD use that as a base class.

src/include/rbd_types.h
```diff
@@ -74,6 +74,10 @@
 #define RBD_HEADER_SIGNATURE	"RBD"
 #define RBD_HEADER_VERSION	"001.005"
 
+#define GROUP_HEADER_PREFIX "group_header."
+
+#define GROUP_DIRECTORY "group_directory"
```
dillaman: Minor: not that anyone should be mixing pools, but can we put an "rbd_" prefix on this?

src/include/rbd_types.h
```diff
@@ -74,6 +74,10 @@
 #define RBD_HEADER_SIGNATURE	"RBD"
 #define RBD_HEADER_VERSION	"001.005"
 
+#define GROUP_HEADER_PREFIX "group_header."
```
dillaman: Minor: can you add RBD_ prefixes to your new #defines

src/tools/rbd/ArgumentTypes.h
```diff
@@ -38,17 +38,21 @@ static const std::string DEST_PREFIX("dest-");
...
 static const std::string IMAGE_SPEC("image-spec");
+static const std::string GROUP_SPEC("group-spec");
 static const std::string SNAPSHOT_SPEC("snap-spec");
 static const std::string IMAGE_OR_SNAPSHOT_SPEC("image-or-snap-spec");
 static const std::string JOURNAL_SPEC("journal-spec");
 static const std::string PATH_NAME("path-name");
 
 // optional arguments
 static const std::string CONFIG_PATH("conf");
+static const std::string CG_NAME("cg-name");
```
dillaman: Minor: group name?

VictorDenisov: I think I just forgot to remove it.

dillaman:
lgtm

dillaman:
retest this please

src/tools/rbd/action/Group.cc
```diff
@@ -0,0 +1,148 @@
+// -*- mode:C++; tab-width:8; c-basic-offset:2; indent-tabs-mode:t -*-
+// vim: ts=8 sw=2 smarttab
+
+#include "tools/rbd/ArgumentTypes.h"
+#include "tools/rbd/Shell.h"
+#include "tools/rbd/Utils.h"
+#include "common/errno.h"
+#include "common/Formatter.h"
+
```
dillaman: Minor: need "#include <iostream>" -- build fails for me without it. Not sure how it passed jenkins.

VictorDenisov:
retest this please

VictorDenisov:
I fixed the include, but now build seems to fail at arbitrary points. Is the build stable itself?

src/test/librbd/test_ConsistencyGroups.cc
```diff
@@ -0,0 +1,63 @@
...
+{
+  librados::IoCtx ioctx;
+  ASSERT_EQ(0, _rados.ioctx_create(_pool_name.c_str(), ioctx));
+
+  librbd::RBD rbd;
+  ASSERT_EQ(0, rbd.group_create(ioctx, "mygroup"));
+
+  vector<string> groups;
+  ASSERT_EQ(0, rbd.group_list(ioctx, groups));
+  ASSERT_EQ(1, groups.size());
```
dillaman: Minor: compile warning comparing signed to unsigned.  Should be ```1U```

dillaman:
@VictorDenisov The MDS unittest failure has been an issue for a week or so.  Did you see another failure?

VictorDenisov:
I see that this test fails on Jenkins.
FAIL: test/run-rbd-unit-tests.sh
It works on my machine though. How can I look into the output of that test?
Console log doesn't show details.

dillaman:
Hmm -- that's a new wrinkle. We used to be able to see the output of the test that failed.

PR #9334, opened 2016-05-26
rbd: inherit the parent image features when cloning an image

yangdongsheng:
Fixes: http://tracker.ceph.com/issues/15388


yangdongsheng:
@dillaman

trociny:
@yangdongsheng Recently this fix to librbd::clone has been committed to master:

https://github.com/ceph/ceph/commit/d305eaad8d57a6f64e6d80eebcd0839a65c9a668

Could you please retest that it fixes this issue too? If it does not I believe we should fix this on API levele, not CLI.

trociny:
Actually, looking at it more closely, we still need modification in `rbd`, but I guess we could do it differently: instead of retrieving parent features we just could leave image options unset, relaying on API fuction (clone) to set it properly from parent. Does it make sense?

yangdongsheng:
@trociny yea, good point. we should put this work into API function, otherwise, we have to do the same thing at every caller of this api. thanx, will update it soon.

trociny:
@yangdongsheng Actually API functions should already work. I pointed you to the latest fix for clone. So I expect some work is only required to modify `utils::get_image_options()` to set only those options that are specified by the user.

yangdongsheng:
Hi @trociny sorry for the late. yes, API side is almost ready, but there is something we should do for other options in clone. I have pushed some commits for them. Please help to check again. Thanx

trociny:
I admit I have been only glancing the changes so far, but this additional logic for rbd `utils::get_image_options` looks alarming to me.

Ideally, I would like `utils::get_image_options` only set options specified by the user (so no fallback to rbd_default options from config here). And the logic how to set necessary missing options (either form config option or from parent) are performed internally. This way it will work the same both for users running CLI and API.

@yangdongsheng have you tried this? Do you see any reason why this wouldn't work?

@dillaman what do you think?

yangdongsheng:
@trociny no I did not try it. But that sounds good to me. I will try it if @dillaman has no objection.

dillaman:
I agree with @trociny

xinxinsh:
should process --image-shared flag https://github.com/ceph/ceph/blob/master/src/tools/rbd/Utils.cc#L415

yangdongsheng:
@xinxinsh Yes, good, thanx.

yangdongsheng:
@trociny what about this version?

ErwanAliasr1:
The job was reported failed beause of my work on the CI.
I respawned the job and it's green : https://jenkins.ceph.com/job/ceph-pull-requests/6688/

src/tools/rbd/Utils.cc
```diff
@@ -367,53 +369,27 @@ int get_image_options(const boost::program_options::variables_map &vm,
...
-    if (features_specified && ((features & RBD_FEATURE_STRIPINGV2) != 0)) {
-      std::cerr << "must specify both of stripe-unit and stripe-count when specify striping features" << std::endl;
-      return -EINVAL;
-    }
-    features &= ~RBD_FEATURE_STRIPINGV2;
+    stripe_specified = true;
   }
 
   if (vm.count(at::IMAGE_SHARED) && vm[at::IMAGE_SHARED].as<bool>()) {
     features &= ~RBD_FEATURES_SINGLE_CLIENT;
```
dillaman: The ```features``` haven't potentially been initialized to the default feature set at this time if ```!features_specified```

src/librbd/internal.cc
```diff
@@ -1273,6 +1277,9 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force,
 		   << " stripe_count = " << stripe_count
 		   << dendl;
 
+    if (!order)
```
dillaman: Minor: move this to be an "} else if (order == 0) {" at line 1272

src/librbd/internal.cc
```diff
@@ -2230,29 +2238,33 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force,
...
       opts.set(RBD_IMAGE_OPTION_STRIPE_COUNT, stripe_count);
+    } else {
+      features |= RBD_FEATURE_STRIPINGV2;
     }
     uint64_t order = src->order;
     if (opts.get(RBD_IMAGE_OPTION_ORDER, &order) != 0) {
       opts.set(RBD_IMAGE_OPTION_ORDER, order);
     }
+    if (opts.get(RBD_IMAGE_OPTION_FEATURES, &features) != 0) {
+      opts.set(RBD_IMAGE_OPTION_FEATURES, features);
```
dillaman: Wouldn't you need to unconditionally set the features?  The ```features``` variable has already been initialized from the opts (if provided) or from the parent.  Additionally, the ```features``` have had ```RBD_FEATURE_STRIPINGV2``` applied.

src/librbd/internal.cc
```diff
@@ -2230,29 +2238,33 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force,
...
-    }
     uint64_t format = src->old_format ? 1 : 2;
     if (opts.get(RBD_IMAGE_OPTION_FORMAT, &format) != 0) {
       opts.set(RBD_IMAGE_OPTION_FORMAT, format);
     }
     uint64_t stripe_unit = src->stripe_unit;
     if (opts.get(RBD_IMAGE_OPTION_STRIPE_UNIT, &stripe_unit) != 0) {
       opts.set(RBD_IMAGE_OPTION_STRIPE_UNIT, stripe_unit);
+    } else {
+      features |= RBD_FEATURE_STRIPINGV2;
```
dillaman: Could move this logic to ```create``` so that you can remove the STRIPINGV2 feature is "fancy" striping isn't being used and can auto-apply it if it is being used.  Otherwise, it seems inconsistent behavior depending on whether or not it was specified vs inherited.

src/tools/rbd/action/Clone.cc
```diff
@@ -19,14 +19,6 @@ int do_clone(librbd::RBD &rbd, librados::IoCtx &p_ioctx,
              const char *p_name, const char *p_snapname,
              librados::IoCtx &c_ioctx, const char *c_name,
              librbd::ImageOptions& opts) {
-  uint64_t features;
-  int r = opts.get(RBD_IMAGE_OPTION_FEATURES, &features);
-  assert(r == 0);
-
-  if ((features & RBD_FEATURE_LAYERING) != RBD_FEATURE_LAYERING) {
```
dillaman: This check should be moved into librbd::clone -- right now it only checks the parent features not the new clone's features

yangdongsheng:
@ErwanAliasr1 Great, thanx a lot. :)

src/librbd/internal.cc
yangdongsheng: In this case, features is the features from src before 2261, but if RBD_IMAGE_OPTION_FEATURES is specified in opts, we should use the features from opts. In other words. it's not unconditionally, because we only set the features when RBD_IMAGE_OPTION_FEATURES is not specified.

src/librbd/internal.cc
yangdongsheng: Okey,

src/tools/rbd/action/Clone.cc
yangdongsheng: Agreed,

yangdongsheng:
@dillaman Really sorry for the late reply.  I introduced  two options in image_options for the partial specified features. Please help to review again.

src/include/rbd/librbd.h
```diff
@@ -144,6 +144,8 @@ enum {
   RBD_IMAGE_OPTION_JOURNAL_ORDER = 5,
   RBD_IMAGE_OPTION_JOURNAL_SPLAY_WIDTH = 6,
   RBD_IMAGE_OPTION_JOURNAL_POOL = 7,
+  RBD_IMAGE_OPTION_FEATURES_SET = 8,
```
dillaman: I am not sure I understand the purpose of these new options.  If ```RBD_IMAGE_OPTION_FEATURES``` is set, you should use them -- otherwise use the defaults.

yangdongsheng: Okey, Let me try to explain what I want to do here. Yes, if ```RBD_IMAGE_OPTION_FEATURES``` is set, we shoud use them, otherwise use the defaults. And where are the defaults from, two ways, one from g_conf, the other one from parent. But now, the logic is placed in librbd, rather than rbd tools, then when we want to specify features partially while using default, that's not easy, because we don't know what is the default features at that time.

Example: --image-shared. When we use clone with --image-shared. we want to use the features from parent, at the same time, we want to specify some of the features.

To solve this kind of problem, I introduced two options ```RBD_IMAGE_OPTION_FEATURES_SET``` and ```RBD_IMAGE_OPTION_FEATURES_CLEAR```.  They are working as a additional way for features. Then RBD_IMAGE_OPTION_FEATURES still work as you described, " If RBD_IMAGE_OPTION_FEATURES is set, you should use them -- otherwise use the defaults.". But if we want to do some partial specification, such as clone with --image-shared. we can use these two options. 

I hope I explain myself clearly.

dillaman: In your example above, you could introduce a FEATURES_MASK and use that to mask off the inherited / default features for things like --image-shared.

yangdongsheng: @dillaman  Do you mean pass FEATURES_MASK = RBD_FEATURES_SINGLE_CLIENT. and then we can mask it out in librbd, right?

Yes, the RBD_IMAGE_OPTION_FEATURES_CLEAR is working similarly.
```
+ uint64_t features_clear = 0;
+ uint64_t features_set = 0;
+ opts.get(RBD_IMAGE_OPTION_FEATURES_CLEAR, &features_clear);
+ opts.get(RBD_IMAGE_OPTION_FEATURES_SET, &features_set);
+
+ uint64_t conflict = features_clear & features_set;
+ features_clear &= ~conflict;
+ features_set &= ~conflict;
+
+ features |= features_set;
+ features &= ~features_clear;
```

dillaman: @yangdongsheng Yes -- and then remove RBD_IMAGE_OPTION_FEATURES_SET since RBD_IMAGE_OPTION_FEATURES (or default features) will apply the optional RBD_IMAGE_OPTION_FEATURES_MASK

yangdongsheng: yes MASK can work in my example. but that's not a general solution I think. It can only clear bits in default features but can not set bits in features which is the reason I introduce FEATURE SET

dillaman: Why not? If you optionally specify the mask, librbd should use it on the features (where the features are the defaults if not set or the provided features).

dillaman: ... plus I think having RBD_IMAGE_OPTION_FEATURES and RBD_IMAGE_OPTION_FEATURES_SET is confusing since they essentially do the same thing. What if I specify both? What if I specify RBD_IMAGE_OPTION_FEATURES_SET and MASK? Too many unclear behaviors. Whereas with just the _MASK, it's clear it is always applied last to your feature set before the image is created.

yangdongsheng: Example:
There is an image with no-object-map. But I want to clone a new one with object-map feature enabled. Then I need to not set RBD_IMAGE_OPTION_FEATURES to tell librbd to inherit features from parent and set RBD_IMAGE_OPTION_FEATURES_SET to apply object-map bit in features. 

RBD_IMAGE_OPTION_FEATURES_[SET|CLEAR] are always applied no matter RBD_IMAGE_OPTION_FEATURES specified or not. And I am kicking out the bits which are both specified in SET and CLEAR. 

This is the main logic of it:
https://github.com/ceph/ceph/pull/9334/files#diff-254555dde8dcfb7fb908791ab8214b92R1269

yangdongsheng: Currently, --image-shared is taking the use of RBD_IMAGE_OPTION_FEATURES_CLEAR. Because it want to clear some bits from inherited features.

dillaman: Ah -- I now get what you are trying to accomplish.

src/tools/rbd/Utils.cc
```diff
@@ -529,10 +510,19 @@ int get_image_options(const boost::program_options::variables_map &vm,
...
-  opts->set(RBD_IMAGE_OPTION_STRIPE_UNIT, stripe_unit);
-  opts->set(RBD_IMAGE_OPTION_STRIPE_COUNT, stripe_count);
+  if (order_specified)
+    opts->set(RBD_IMAGE_OPTION_ORDER, order);
+  if (features_specified)
+    opts->set(RBD_IMAGE_OPTION_FEATURES, features);
+  if (features_clear_specified) {
+    opts->set(RBD_IMAGE_OPTION_FEATURES_CLEAR, features_clear);
+  }
+  if (features_set_specified)
```
dillaman: Minor: doesn't appear to be ever set to true

src/include/rbd/librbd.h
yangdongsheng: Ha great sorry for my poor English.

src/tools/rbd/Utils.cc
yangdongsheng: It's always false now because we are not using it so far

dillaman:
Failing some test cases:
http://teuthology.ovh.sepia.ceph.com/teuthology/jdillaman-2016-07-06_17:56:15-rbd-wip-jd-testing---basic-openstack/33066/teuthology.log
http://teuthology.ovh.sepia.ceph.com/teuthology/jdillaman-2016-07-06_17:56:15-rbd-wip-jd-testing---basic-openstack/33068/teuthology.log
http://teuthology.ovh.sepia.ceph.com/teuthology/jdillaman-2016-07-06_17:56:15-rbd-wip-jd-testing---basic-openstack/33075/teuthology.log

yangdongsheng:
@dillaman My bad, I forgot to update ```rbd import``` which is using get_image_options(). It's fixed now. But I can't access sepia for testing. I am not sure it passes integration testing now or not. Could you help to take a look or test again?

yangdongsheng:
I found there is a fail in log about pybind. But I have not addressed it.

dillaman:
@yangdongsheng Thanks for the heads-up -- I'll wait on re-test for now.

yangdongsheng:
@dillaman I addressed the pybind problem in integration testing now. Retest this please. :)

src/librbd/internal.cc
```diff
@@ -1453,6 +1471,7 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force,
     uint64_t order;
     uint64_t size;
     uint64_t p_features;
+    uint64_t stripe_unit, stripe_count;
```
dillaman: Minor: causes compiler warning that these aren't used

src/tools/rbd/Utils.cc
```diff
@@ -422,8 +422,13 @@ int validate_snapshot_name(at::ArgumentModifier mod,
 
 int get_image_options(const boost::program_options::variables_map &vm,
 		      bool get_format, librbd::ImageOptions *opts) {
-  uint64_t order, features, stripe_unit, stripe_count, object_size;
+  uint64_t order, stripe_unit, stripe_count, object_size;
```
dillaman: Minor: order, stripe_unit, and stripe_count cause "may be used uninitialized" compiler warnings

yangdongsheng:
@dillaman Updated. thanx

dillaman:
Hopefully the last of the test failures:

yangdongsheng:
@dillaman Oh, sorry about it. It seems I have to setup a local integration testing environment for a fully testing now. Thanx for your help.

yangdongsheng:
@dillaman I found I can't reproduce the fail by run ```CEPH_CLI_TEST_DUP_COMMAND=1 CEPH_REF=001c65727312bac894ce6f7dc6159c962aaaa2a3 TESTDIR="/tmp/cephtest" CEPH_ARGS="--cluster ceph" CEPH_ID="0" ceph-coverage /tmp/coverage timeout 3h ../qa/workunits/rbd/journal.sh``` in my local ceph environment. Is that possible this failure is not related with this PR?

@yuriw could you help to take this PR into your testing? thanx.

trociny:
@yangdongsheng You don't need to set CEPH_* env variables, just run:
```
../qa/workunits/rbd/journal.sh
```
But make sure your PATH is set correctly and the script runs your binary and lib:
```
which rbd
ldd `which rbd` |grep rbd
```
Also you can try to reproduce manually, basically it failed on this:
```
# create an image without journaling feature  
rbd create --size 256 ${src} 
# copy the image setting journaling features
rbd copy --image-feature exclusive-lock --image-feature journaling \
	--journal-pool rbd \
	--journal-object-size 20M \
	--journal-splay-width 6 \
	${src} ${image}
# check that journaling settings are applied
rbd journal info --image ${image}
```
From teuthology.log it failed reporting that journaling feature was not enabled for the image. Does it work for you?

Another alarming message in teuthology.log: "rbd: image format 1 is deprecated" after `rbd create ...` commands. It is not expected.

yangdongsheng:
```
[root@workstation build]# rbd create --size 256 test
2016-07-22 01:27:13.659840 7fc76127a5c0 -1 WARNING: the following dangerous and experimental features are enabled: *
2016-07-22 01:27:13.660008 7fc76127a5c0 -1 WARNING: the following dangerous and experimental features are enabled: *
2016-07-22 01:27:13.661730 7fc76127a5c0 -1 WARNING: the following dangerous and experimental features are enabled: *
[root@workstation build]# rbd copy --image-feature exclusive-lock --image-feature journaling     --journal-pool rbd     --journal-object-size 20M     --journal-splay-width 6     test copy
2016-07-22 01:27:22.501427 7f713d69b5c0 -1 WARNING: the following dangerous and experimental features are enabled: *
2016-07-22 01:27:22.501591 7f713d69b5c0 -1 WARNING: the following dangerous and experimental features are enabled: *
2016-07-22 01:27:22.503236 7f713d69b5c0 -1 WARNING: the following dangerous and experimental features are enabled: *
Image copy: 100% complete...done.
[root@workstation build]# rbd journal info --image copy
2016-07-22 01:27:28.490416 7f3518cad5c0 -1 WARNING: the following dangerous and experimental features are enabled: *
2016-07-22 01:27:28.490573 7f3518cad5c0 -1 WARNING: the following dangerous and experimental features are enabled: *
2016-07-22 01:27:28.491536 7f3518cad5c0 -1 WARNING: the following dangerous and experimental features are enabled: *
rbd journal '10212ae8944a':
        header_oid: journal.10212ae8944a
        object_oid_prefix: journal_data.0.10212ae8944a.
        order: 25 (32768 kB objects)
        splay_width: 6
        object_pool: rbd
[root@workstation build]# rbd info copy
2016-07-22 01:27:46.543855 7f0b1759d5c0 -1 WARNING: the following dangerous and experimental features are enabled: *
2016-07-22 01:27:46.544664 7f0b1759d5c0 -1 WARNING: the following dangerous and experimental features are enabled: *
2016-07-22 01:27:46.545525 7f0b1759d5c0 -1 WARNING: the following dangerous and experimental features are enabled: *
rbd image 'copy':
        size 256 MB in 64 objects
        order 22 (4096 kB objects)
        block_name_prefix: rbd_data.10212ae8944a
        format: 2
        features: exclusive-lock, journaling
        flags: 
        journal: 10212ae8944a
        mirroring state: disabled
[root@workstation build]# which rbd
/data/ceph/build/bin/rbd
[root@workstation build]# ldd `which rbd` |grep rbd
        librbd.so.1 => /data/ceph/build/lib/librbd.so.1 (0x00007fb94f5c0000)
[root@workstation build]# rbd --version
ceph version v10.2.0-2599-g74a8078 (74a8078a0f72f619642bf6698a8c9f709e940107)
```

Thanx @trociny I manually tested my patch as above. is that same to you?

In addition, I tried the xml format.
```
[root@workstation build]# rbd --format xml journal info --image copy
2016-07-22 01:39:51.267635 7f18f21a55c0 -1 WARNING: the following dangerous and experimental features are enabled: *
2016-07-22 01:39:51.267792 7f18f21a55c0 -1 WARNING: the following dangerous and experimental features are enabled: *
2016-07-22 01:39:51.268863 7f18f21a55c0 -1 WARNING: the following dangerous and experimental features are enabled: *
<journal><journal_id>10212ae8944a</journal_id><header_oid>journal.10212ae8944a</header_oid><object_oid_prefix>journal_data.0.10212ae8944a.</object_oid_prefix><order>25</order><splay_width>6</splay_width><object_pool>rbd</object_pool></journal>[root@workstation build]# 
```

yangdongsheng:
Oh, so we should create test in format 1 what might be the default in this case. I see the problem now.

trociny:
@yangdongsheng Actually, I don't see rbd_default_format is explicitly set in that test, and there many other tests from that teuthology run where this warning was observed. And this is the main question: why it uses old format here as default? Don't you observe this in your local environment (don't you have rbd_default_format = 2 explicitly set in your config)?

I have not built your branch yet, so can't recheck this in my env.

yangdongsheng:
@trociny I tried my local env again. I did not observe the deprecating warning in creating.
```
[root@workstation build]# ldd `which rbd`|grep rbd
        librbd.so.1 => /data/ceph/build/lib/librbd.so.1 (0x00007fcf29338000)
[root@workstation build]# which rbd
/data/ceph/build/bin/rbd
[root@workstation build]# rbd --version
ceph version v10.2.0-2599-g74a8078 (74a8078a0f72f619642bf6698a8c9f709e940107)
[root@workstation build]# rbd create --size 1G test1
2016-07-22 03:17:29.375010 7f660fb9c5c0 -1 WARNING: the following dangerous and experimental features are enabled: *
2016-07-22 03:17:29.375183 7f660fb9c5c0 -1 WARNING: the following dangerous and experimental features are enabled: *
2016-07-22 03:17:29.376296 7f660fb9c5c0 -1 WARNING: the following dangerous and experimental features are enabled: *
[root@workstation build]# rbd info test1
2016-07-22 03:17:33.516528 7ff98ac0f5c0 -1 WARNING: the following dangerous and experimental features are enabled: *
2016-07-22 03:17:33.516821 7ff98ac0f5c0 -1 WARNING: the following dangerous and experimental features are enabled: *
2016-07-22 03:17:33.517604 7ff98ac0f5c0 -1 WARNING: the following dangerous and experimental features are enabled: *
rbd image 'test1':
        size 1024 MB in 256 objects
        order 22 (4096 kB objects)
        block_name_prefix: rbd_data.101f2ae8944a
        format: 2
        features: layering, exclusive-lock, object-map, fast-diff, deep-flatten
        flags: 
[root@workstation build]# ceph daemon mon.a config show|grep rbd_default
    "rbd_default_format": "2",
    "rbd_default_order": "22",
    "rbd_default_stripe_count": "0",
    "rbd_default_stripe_unit": "0",
    "rbd_default_features": "61",
    "rbd_default_map_options": "",
```

yangdongsheng:
Okey, I found the problem now, there is an uninitialized variable of format. when getting format in creating.

yangdongsheng:
@trociny or @dillaman I hope I have fixed the problem above, could you help to queue this PR for integration testing again? thanx a lot.

PR #9335, opened 2016-05-26
packaging: drop python-flask dependency from radosgw

ktdreyer:
The PowerDNS integration code itself is not packaged at this point, so remove the python-flask dependency from the radosgw package.

Maybe the PowerDNS integration bits could live in a separate sub-package, eventually.

Fixes: http://tracker.ceph.com/issues/16032


ktdreyer:
CC'ing @smithfarm here in case he wants to review

smithfarm:

PR #9338, opened 2016-05-26
cephfs: should give a value before use cpp_strerror

songbaisen:
cephfs: should give a value before use cpp_strerror


jcsp:
I've merged this because it was correct and in case anyone wants to backport it for environments where the tool is still in use, but please be aware that this tool is deprecated and will soon be removed (http://tracker.ceph.com/issues/16035)

songbaisen:
@jcsp :ok:  :smile: Thank you sir!

PR #9339, opened 2016-05-26
mon:quickly return on on_active function

songbaisen:
mon:quickly return on on_active function

If we are not quorum must be not leader too.
So here quickly return.


jecluis:
I think you missed the point of that check - it does not mean you're not in quorum, just that you HAD NEVER been in a quorum before. on_active() functions are only called when you are in a quorum.

Returning there makes no significant impact AND prevents the leader from outputting the info on the next check.

songbaisen:
@jecluis  Oh,thank you! I really mistake understand here.BTW would you like help me to review another 
PR as below .Thanks again!
https://github.com/ceph/ceph/pull/8776

PR #9346, opened 2016-05-26
client: enable client_quota by default

jcsp:
It was awkward that our default behaviour
was to let people set the quota vxattrs but
then not enforce them.


liewegas:
:+1:

jcsp:
This is exposing bugs:
http://tracker.ceph.com/issues/16067
http://tracker.ceph.com/issues/16066

Let's fix those before merging this, to keep the nightlies cleaner

liewegas:
\o/

PR #9354, opened 2016-05-26
rgw: Fallback to Host header for bucket name.

robbat2:
RGW should fallback to using the Host header as the bucket name even when it is
NOT a suffix match against the DNS names, or a match against the CNAME rule.

This mirrors AWS S3 behavior for these cases (The AWS S3 does not do any DNS
lookups regardless).

Backport: jewel
Fixes: http://tracker.ceph.com/issues/15975

### DIFF
--- a/src/rgw/rgw_rest.cc
+++ b/src/rgw/rgw_rest.cc
@@ -1604,6 +1604,27 @@ int RGWREST::preprocess(struct req_state *s, RGWClientIO* cio)
   s->info.request_uri_aws4 = s->info.request_uri;
 
   s->cio = cio;
+
+  // We need to know if this RGW instance is running the s3website API with a
+  // higher priority than regular S3 API, or possibly in place of the regular
+  // S3 API.
+  // Map the listing of rgw_enable_apis in REVERSE order, so that items near
+  // the front of the list have a higher number assigned (and -1 for items not in the list).
+  list<string> apis;
+  get_str_list(g_conf->rgw_enable_apis, apis);
+  int api_priority_s3 = -1;
+  int api_priority_s3website = -1;
+  auto api_s3website_priority_rawpos = std::find(apis.begin(), apis.end(), "s3website");
+  auto api_s3_priority_rawpos = std::find(apis.begin(), apis.end(), "s3");
+  if (api_s3_priority_rawpos != apis.end()) {
+    api_priority_s3 = apis.size() - std::distance(apis.begin(), api_s3_priority_rawpos);
+  }
+  if (api_s3website_priority_rawpos != apis.end()) {
+    api_priority_s3website = apis.size() - std::distance(apis.begin(), api_s3website_priority_rawpos);
+  }
+  ldout(s->cct, 10) << "rgw api priority: s3=" << api_priority_s3 << " s3website=" << api_priority_s3website << dendl;
+  bool s3website_enabled = api_priority_s3website >= 0;
+
   if (info.host.size()) {
     ldout(s->cct, 10) << "host=" << info.host << dendl;
     string domain;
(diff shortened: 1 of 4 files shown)


robbat2:
@yehudasa please review.

oritwas:
This breaks simple run (vstart.sh). we get 403 in create bucket and trying to access localhost:8000 in the browser returns invalid bucket name

robbat2:
@oritwas what was the full URL accessed? IIRC vstart runs with `rgw_dns_name=localhost`, so it should have taken the bucket name from the first URI argument. I'll try to put some cycles in today to check (vstart breaks in my environment so I don't use it).

robbat2:
@oritwas:
I've got a new revision up now.
The only change from before is this block:
```
if (subdomain.empty()) {
  subdomain.append(info.host);
  in_hosted_domain = 1;
}
```
To:
```
// If BOTH domain & subdomain variables are empty, then none of the above
// cases matched anything, and we should fall back to using the Host header
// directly as the bucket name.
if (domain.empty() && subdomain.empty()) {
  subdomain.append(info.host);
  in_hosted_domain = 1;
}
```
I realize that I had an equivalent bit in during prior test runs of s3-tests, but lost it in the cleanup prior to posting the PR.

robbat2:
@oritwas hmm, there's a corner case that is broken by changing that if statement again, but does work on AWS S3.

The cases of DNS, for bucket/object level operations:
```
www.example.com. IN CNAME www.example.com.rgw-s3.host.com. # case 1a
www.example.com. IN CNAME altbucketname.rgw-s3.host.com. # case 1b (RGW-specific)
www.example.com. IN CNAME rgw-s3.host.com. # case 2a, where RHS matches rgw dns name
www.example.com. IN CNAME ip-rgw-s3.host.com. # case 2b, where RHS does not match rgw dns name
example.com IN A 1.2.3.4 # case 3, same IP as rgw-s3.host.com, where RHS does not match rgw dns name
```

This really shows that we need to redo the logic here, ceph's magic CNAME case helps complicate matters :-(.

robbat2:
@oritwas 
This new push (commit 5f31b42) again for corners to the corner case; thie one does pass all of s3-tests including s3website pieces (takes ~200 seconds here).

- Case 1a & 1b: should populate both domain & subdomain.Bucket is subdomain.
- Case 2a: subdomain empty, domain NOT empty.
-- If domain == info.host, then it's a service operation (list all buckets) OR a path-style request (depends on the URI path).
-- If domain != info.host, then it's a bucket or object operation. Bucket is info.host.
- Case 2b & case 3 should lead to both subdomain & domain being empty, and the Bucket is info.host.

oritwas:
@robbat2 , this code will not work with IP addresses because rgw_find_host_in_domains doesn't handle IP address correctly. We need to handle IP address as users use it.

robbat2:
@oritwas hmm, we should get some more of this added to s3tests as well, there's certainly no test presently that uses an IP in the Host header (would probably want to explicitly override the Host header vs what we were connecting to, for testing purposes); i did already implemented that for the s3-tests s3website checks.

robbat2:
@oritwas now includes a check for the IP as well as making sure the bucket name is valid before using it (eg if the hostname wasn't a valid bucket name and there was a path-style bucket name next).

src/rgw/rgw_rest.cc
```diff
@@ -1674,6 +1692,32 @@ int RGWREST::preprocess(struct req_state *s, RGWClientIO* cio)
...
+    // cases matched anything, and we should fall back to using the Host header
+    // directly as the bucket name.
+    // As additional checks:
+    // - if the Host header is an IP, we're using path-style access without DNS
+    // - Also check that the Host header is a valid bucket name (per relaxed
+    //   bucket rules) before using it.
+    if (subdomain.empty()
+        && (domain.empty() || domain != info.host)
+        && !looks_like_ip_address(info.host.c_str())
+        && valid_s3_bucket_name(info.host, true)) {
```
oritwas: this is shared code with swift how about using valid_bucket_name instead?

robbat2: @oritwas you mean RGWHandler_REST::validate_bucket_name validate_bucket_name which isn't static; I'll tweak it so it can be used in the static context of RGWREST::preprocess.

robbat2:
@oritwas now pushed with your validate_bucket_name change

robbat2:
@oritwas ping, just wondering if there are any further changes you need before this can be merged (And backported)

oritwas:
@robbat2 , it looks good just waiting for the teuthology run to complete

PR #9355, opened 2016-05-27
tools/rbd: forbid --allow-shrinking flag when image is resized to a l…

xinxinsh:
…arger size

Fixes: #15991

Signed-off-by: xinxin shu <shuxinxin@chinac.com>


yangdongsheng:
Hi @xinxinsh , I am not sure without this change, what kind of problem we will meet?

yangdongsheng:
I mean, expanding the size is allowed by default in resize command. Then what we can do with resize is that ["expanding size", ] by default. But if you specify the --allow-shrinking, the capability of resize command would be ["expanding size", "shrinking size"]. shrinking is allowed now. But why expanding would be disallowed in this case? It's a little confusing to me.

PR #9357, opened 2016-05-27
osd/PGBackend: build_push_op assert() before dereferencing iter

Aran85:
here i meet a segment fault, maybe assert better

Signed-off-by: Zengran Zhang zhangzengran@h3c.com

### DIFF
--- a/src/osd/ReplicatedBackend.cc
+++ b/src/osd/ReplicatedBackend.cc
@@ -1985,6 +1985,7 @@ int ReplicatedBackend::build_push_op(const ObjectRecoveryInfo &recovery_info,
     ObjectMap::ObjectMapIterator iter =
       store->get_omap_iterator(coll,
 			       ghobject_t(recovery_info.soid));
+    assert(iter);
     for (iter->lower_bound(progress.omap_recovered_to);
 	 iter->valid();
 	 iter->next(false)) {


Aran85:
steps to reproduce
./vstart.sh -n
./ceph osd pool set rbd size 2
./rados put testfile /etc/hosts -p rbd
ps: the testfile map to pg0.6 whose upset is [0,1]
rm ./dev/osd0/current/0.6_head/xxxtestfilexxx_0
./ceph osd out osd.1
./ceph osd down osd.1
then osd.0 crash.

bc the fd of testifle is in filestore's fdcache, so it could get the object context of it. so let the osd.0 think testfile is not missing, but when it prepare to push to a new upset osd, it fail to lfn_open it. 

does this situation need to worry about in actual production environment?

yuyuyu101:
I think assert is fine now. Because ceph doesn't have a rule to handle missing file in local filesystem...

yuyuyu101:
And I think this isn't allowed

yuyuyu101:
I don't think this change makes sense, remove this line will result into later segment fault. If you want to have a good error handle. return ENOENT may better.

But since we will hit assert(r == 0) by checking return value. There is no difference in real env. Anyway, return errno give a good sharp to let others handle this case later.

Aran85:
@yuyuyu101 thanks for review. the assert line  added by me,not remove. do you mean we should return errno here?

yuyuyu101:
oh, sorry. my mistake. yes, I think we can return errno.

tchaikov:
- http://pulpito.ceph.com/kchai-2016-11-13_05:09:22-rados-wip-kefu-testing---basic-smithi/
- http://pulpito.ceph.com/kchai-2016-11-11_07:45:56-rados-wip-kefu-testing---basic-smithi/

PR #9360, opened 2016-05-27
rbd-mirror: support multiple replicated pools

dillaman:


src/tools/rbd_mirror/Replayer.cc
```diff
@@ -228,26 +83,22 @@ class MirrorStatusWatchCtx {
 };
 
 Replayer::Replayer(Threads *threads, std::shared_ptr<ImageDeleter> image_deleter,
-                   RadosRef local_cluster, const peer_t &peer,
-                   const std::vector<const char*> &args) :
+                   RadosRef local_cluster, int64_t local_pool_id,
+                   const peer_t &peer, const std::vector<const char*> &args) :
   m_threads(threads),
   m_image_deleter(image_deleter),
   m_lock(stringify("rbd::mirror::Replayer ") + stringify(peer)),
```
trociny: @dillaman I suppose lock name should be updated too?

dillaman: Right now the Replayer is still tied to a single peer -- so that would provide a unique lock name.  Eventually when we support multiple peers, this class might morph into ```PoolReplayer``` which is tied to the local pool with multiple ```PeerReplayers``` associated for each remote peer.

src/tools/rbd_mirror/Replayer.cc
```diff
@@ -228,26 +83,22 @@ class MirrorStatusWatchCtx {
...
   m_peer(peer),
   m_args(args),
   m_local(local_cluster),
   m_remote(new librados::Rados),
-  m_asok_hook(nullptr),
+  m_local_pool_id(local_pool_id),
   m_replayer_thread(this)
 {
-  CephContext *cct = static_cast<CephContext *>(m_local->cct());
-  m_asok_hook = new ReplayerAdminSocketHook(cct, m_peer.cluster_name, this);
```
trociny: It is not that I insist on not removing per replayer asok commands, but anyway, I think it is easy to keep them, just we need to use a different key as a name. I can image situations where running such commands per replayer only could be useful, still sure we can live without it.

dillaman: We would need to come up with a new naming scheme -- the uniqueness is local pool and peer, but we can't have "<pool>/<cluster>" since that conflicts with image naming format.  I am open to suggestions.

dillaman: Maybe just a space between the local pool name and the peer cluster?

rbd mirror <command> <pool> <peer cluster>

dillaman:
@trociny I swapped the last commit for one that keeps the Replayer asok commands but prefixes it with the pool name.

trociny:
@dillaman I like `rbd mirror {command} {pool} {peer cluster}`

trociny:
@dillaman I think in Replayer::print_status() it would be nice to also print "pool" field? Otherwise lgtm.

trociny:
The tests look good.
http://pulpito.ceph.com/trociny-2016-05-28_08:39:09-rbd-wip-mgolub-testing---basic-vps/

dillaman:
@trociny Tweaked the last commit to print out the pool name.

PR #9361, opened 2016-05-27
osd/OSD.h: add override to virtual functions

stiopaa1:

### DIFF
--- a/src/osd/OSD.h
+++ b/src/osd/OSD.h
@@ -1922,7 +1922,7 @@ private:
       : ThreadPool::BatchWorkQueue<PG>(
 	"OSD::PeeringWQ", ti, si, tp), osd(o) {}
 
-    void _dequeue(PG *pg) {
+    void _dequeue(PG *pg) override {
       for (list<PG*>::iterator i = peering_queue.begin();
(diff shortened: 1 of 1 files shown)


src/osd/OSD.h
```diff
@@ -1934,15 +1934,15 @@ class OSD : public Dispatcher,
 	}
       }
     }
-    bool _enqueue(PG *pg) {
+    bool _enqueue(PG *pg) override {
```
tchaikov: can we add `noexcept` for better performance?

stiopaa1: @tchaikov 
I don't have much experience with noexcept, so please correct me if i'm wrong. 
My understanding is that we use noexcept for functions that are guaranteed not to throw exceptions, but here the push_back might throw bad_alloc exception?

tchaikov: > I don't have much experience with noexcept

neither do i =)

i think the actual questions are:

1. we need to mark the overrided method in the parent class with `noexcept`, if we want to have it in the child class.
1. shall we `terminate()` at seeing an exception we do not handle?

tchaikov:
lgtm

PR #9363, opened 2016-05-27
ceph-disk: unmount the osd data when destroy

w11979:
when destroy one OSD by the command of ceph-disk destroy --destroy-by-id ##, the osd dir and dev are not unmount.
This fixes unmount the osd dir when destroy.

Signed-off-by: w11979 <wang.wenfeng@h3c.com>

### DIFF
--- a/src/ceph-disk/ceph_disk/main.py
+++ b/src/ceph-disk/ceph_disk/main.py
@@ -3609,6 +3609,11 @@ def main_destroy_locked(args):
     # Deallocate OSD ID
     _deallocate_osd_id(args.cluster, osd_id)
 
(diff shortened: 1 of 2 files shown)


w11979:
@dachary @liewegas When destroy one OSD by the command of ceph-disk destroy --destroy-by-id ##, the osd dir and dev are not unmount. Could you see and merge it? thank you.

dachary:
prior to destroying the OSD, you must deactivate it. When deactivating an OSD, the matching data partition is unmounted. Is there a scenario we missed where an OSD is deactivated but the path is not umounted ?

PR #9366, opened 2016-05-27
convert parameter type from pointer into reference

snakeAngel2015:
Signed-off-by: snakeAngel2015 li.jieA@h3c.com

### DIFF
--- a/src/os/ObjectStore.cc
+++ b/src/os/ObjectStore.cc
@@ -26,7 +26,7 @@
 #include "kstore/KStore.h"
 
 void decode_str_str_map_to_bl(bufferlist::iterator& p,
-			      bufferlist *out)
+			      bufferlist &out)
 {
   bufferlist::iterator start = p;
   __u32 n;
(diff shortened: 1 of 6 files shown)


snakeAngel2015:
hi Sage,
 this commit can optimize the function's parameters . Could you see and merge it? thank you. @liewegas

snakeAngel2015:
hi Sage,
      this commit can optimize the function's parameters . Could you review and merge it? thank you. @liewegas

dachary:

dachary:
I don't think it will improve performances. If you think otherwise, please provide performance benchmarks showing it does and re-open this pull request.

tchaikov:
also we have an (undocumented) convention to use pointer as the output parameter.

dachary:
@tchaikov you're right ! I did not notice because the comments mentionned "optimize". Sorry about that @snakeAngel2015 this is a fine pull request.

tchaikov:
@dachary actually, i was concurring you. take `decode_str_str_map_to_bl(bufferlist::iterator& p, bufferlist *out)` as an example, the `bl` is actually an output parameter. so using the pointer to pass  bl is desired. maybe we could change the signature of this method (and its cousins) to something like
```
bufferlist decode_str_str_map_to_bl(bufferlist::iterator& p);
```
because it will very likely to be RVO, so no need to pass the output param to eliminate the overhead of default ctor and copy ctor. but that might be another story.

PR #9367, opened 2016-05-27
rgw/common: fix error return value of hex_to_buf.

Yan-waller:
We should return negative value if error happens, otherwise
the caller may run into exceptions. eg, `rgw_swift_verify_signed_token`
will not return as follow:
```c++
  int ret = hex_to_buf(token, p.c_str(), len);
  if (ret < 0)
    return ret;
```


xiexingguo:
looks good

PR #9371, opened 2016-05-27
build ceph-fuse on OSX

ukernel:
To build ceph-fuse

./autogen
./configure --without-libaio --without-libxfs --without-cython CXXFLAGS='-DGTEST_HAS_TR1_TUPLE=0' LDFLAGS=-L/opt/local/lib CPPFLAGS=-I/opt/local/include --without-mon --without-osd --without-mds --without-rbd
make


renhwztetecs:
Is there a requirement for the OSX system version？

src/Makefile-env.am
```diff
@@ -256,8 +256,6 @@ if ENABLE_COVERAGE
 EXTRALIBS += -lgcov
 endif # ENABLE_COVERAGE
 
-LIBCOMMON += -luuid
```
wjwithagen: @ukernel Not sure if Freebsd can link correctly with uuid

src/common/ceph_time.h
```diff
@@ -20,6 +20,19 @@
 
 #include "include/encoding.h"
 
+#if defined(DARWIN)
+#include <sys/_types/_timespec.h>
+#include <mach/mach.h>
+#include <mach/clock.h>
+
+#define CLOCK_REALTIME CALENDAR_CLOCK
```
wjwithagen: @ukernel Perhaps put this in include/compat.h like I do for FreeBSD. That way all clock compatability stuff is "hidden" in there.

src/common/config.cc
```diff
@@ -33,6 +33,7 @@
 #include <sstream>
 #include <stdlib.h>
 #include <string.h>
+#include <libgen.h>
```
wjwithagen: @ukernel Including libgen.h this way will result in complaints on some other platforms. Hence I included it in an FreebsD conditional.

src/common/darwin_errno.cc
```diff
@@ -0,0 +1,224 @@
...
+      case -56:
+        return -EPERM; //TODO EBADRQC
+      case -57:
+        return -EPERM; //TODO EBADSLT
+      case -59:
+        return -EPERM; //TODO EBFONT
+      case -60:
+        return -ENOSTR;
+      case -61:
+        return -ENOATTR;
```
wjwithagen: @ukernel ENOATTR is really a tricky one since there are some loops in the xattr code that depend on ENOATTR == ENODATA. Since that is what Linux expects as error on some of the xattr calls.

src/common/obj_bencher.cc
```diff
@@ -17,6 +17,7 @@
  */
 #include <pthread.h>
 #include "common/Cond.h"
+#include "include/compat.h"
```
wjwithagen: @ukernel I'd suggest putting compat.h as very first include file.
This will prevent other include files from defining things that need to be repaired later.
Like ENODATA

src/crush/CrushLocation.cc
```diff
@@ -6,6 +6,7 @@
 #include "common/config.h"
 #include "include/str_list.h"
 #include "common/debug.h"
+#include "include/compat.h"
```
wjwithagen: @ukernel  Same here

src/rgw/rgw_http_client.cc
```diff
@@ -9,6 +9,7 @@
 #include "rgw_http_client.h"
 #include "rgw_http_errors.h"
 #include "common/RefCountedObj.h"
+#include "include/compat.h"
```
wjwithagen: @ukernel Just put it as first.

src/test/Makefile-client.am
```diff
@@ -200,7 +200,7 @@ bin_DEBUGPROGRAMS += ceph_test_cls_numops
 ceph_test_cls_journal_SOURCES = test/cls_journal/test_cls_journal.cc
 ceph_test_cls_journal_LDADD = \
         libcls_journal_client.la $(LIBRADOS) \
-        $(LIBCOMMON) $(CRYPTO_LIBS) $(UNITTEST_LDADD) $(RADOS_TEST_LDADD) -luuid
+        $(LIBCOMMON) $(CRYPTO_LIBS) $(UNITTEST_LDADD) $(RADOS_TEST_LDADD)
```
wjwithagen: @ukernel Perhaps this would warrant the a UUID_LIB which can be empty on certain platforms.

src/common/Thread.h
```diff
@@ -56,4 +56,6 @@ class Thread {
   int set_affinity(int cpuid);
 };
 
+int ceph_pthread_setname(pthread_t thread, const char *name);
```
wjwithagen: @ukernel I like this change

wjwithagen:
@ukernel Perhaps take the changes for the pthread_get/set_name out into a different pull.
It is sort of complete separate from a lot of the other changes.

src/common/ceph_time.h
ukernel: clock_gettime for OSX is declared here

ukernel:
On Sat, May 28, 2016 at 10:09 AM, huanwen ren <notifications@github.com>
wrote:

> Is there a requirement for the OSX system version？
>
>
I only test it on EI Capitan

src/common/obj_bencher.cc
ukernel: On Sat, May 28, 2016 at 8:28 PM, Willem Jan Withagen <
notifications@github.com> wrote:

> In src/common/obj_bencher.cc
> <https://github.com/ceph/ceph/pull/9371#discussion_r64988367>:
>
> > @@ -17,6 +17,7 @@
> >   */
> >  #include <pthread.h>
> >  #include "common/Cond.h"
> > +#include "include/compat.h"
>
> @ukernel <https://github.com/ukernel> I'd suggest putting compat.h as
> very first include file.
> This will prevent other include files from defining things that need to be
> repaired later.
> Like ENODATA
>

there are some "#ifndef; #define xxx; #endif". I think putting compat.h at
first causes problem

src/Makefile-env.am
ukernel: On Sat, May 28, 2016 at 8:20 PM, Willem Jan Withagen <
notifications@github.com> wrote:

> In src/Makefile-env.am
> <https://github.com/ceph/ceph/pull/9371#discussion_r64988297>:
>
> > @@ -256,8 +256,6 @@ if ENABLE_COVERAGE
> >  EXTRALIBS += -lgcov
> >  endif # ENABLE_COVERAGE
> >
> > -LIBCOMMON += -luuid
>
> @ukernel <https://github.com/ukernel> Not sure if Freebsd can link
> correctly with uuid
>

This commit exists in my old wip-osx-freebsd branch. I did test it.

src/common/darwin_errno.cc
ukernel: On Sat, May 28, 2016 at 8:27 PM, Willem Jan Withagen <
notifications@github.com> wrote:

> In src/common/darwin_errno.cc
> <https://github.com/ceph/ceph/pull/9371#discussion_r64988356>:
>
> > +      case -53:
> > +        return -EPERM; //TODO EBADR
> > +      case -54:
> > +        return -EPERM; //TODO EXFULL
> > +      case -55:
> > +        return -EPERM; //TODO ENOANO
> > +      case -56:
> > +        return -EPERM; //TODO EBADRQC
> > +      case -57:
> > +        return -EPERM; //TODO EBADSLT
> > +      case -59:
> > +        return -EPERM; //TODO EBFONT
> > +      case -60:
> > +        return -ENOSTR;
> > +      case -61:
> > +        return -ENOATTR;
>
> @ukernel <https://github.com/ukernel> ENOATTR is really a tricky one
> since there are some loops in the xattr code that depend on ENOATTR ==
> ENODATA. Since that is what Linux expects as error on some of the xattr
> calls.
>
yes, I think it's better translate -61 to -ENODATA because most ceph code
checks -ENODATA instead of -ENOATTR.

src/common/ceph_time.h
wjwithagen: But it would be nice if they all end up at the same place??

This the discussion with arrisen quite a feew times now.
Most programmes still asume that they are programming against Linux.
And I think it is less error prone if we stick it in compat.h so it is fixed for all architectures/OSes at once.
The advantage being that all this CLOCK* defines are in one place.

src/common/obj_bencher.cc
wjwithagen: I'd be interested to hear of those problems since I'm currently recommending to put it as first include.
Like I actually witnessed: It prevents redefinition of some values.
And the problem with ENODATA <> ENOATRR is real and so easy to fix otherwise.

src/Makefile-env.am
wjwithagen: So what is the source of your uuid routines then??  libc??
I tried that at the beginning, but it bommed out.

I will try that again, in a seperate run.,
Thanx for the pointer.

src/common/darwin_errno.cc
wjwithagen: The first indication that something is amiss  is running the unittest_*attr* stuff.
There is a loop in the attr code that tests on ENODATA but the call returns ENOATTR (on FreeBSD)
And the only exit out of that loop is the get ENODATA, so the program is stuck in this loop.

jcsp:
Although this is aimed at cephfs, lots of it touches core stuff so could probably do with an extra review (@liewegas, @athanatos)

src/test/Makefile-client.am
ukernel: We used to use UUID_LIB in congfiure.ac. But it was removed when moving to boost uuid implementation

wjwithagen: @ukernel 
Oke, I see, Will check if the FreeBSD Boost package has UUID stuff in it, since not all Boost is packaged for FreeBSD.

src/Makefile-env.am
wjwithagen: @ukernel 
Right is seems that I'm able to compile and link without -luuid.
Now running tests, but the first 30 tests are fine

jcsp:
(This has outstanding comments and needs rebase)

src/test/strtol.cc
```diff
@@ -234,21 +234,6 @@ TEST(StrictSICast, Error) {
     (void)strict_si_cast<int>("1T", &err);
     ASSERT_NE(err, "");
   }
-  {
```
tchaikov: i think it's good to have more coverage for `strict_si_cast<int64_t>` even if `int64_t` is identical to `long long int` on some platforms.

src/rgw/rgw_admin.cc
```diff
@@ -2161,7 +2161,7 @@ int main(int argc, char **argv)
         return EINVAL;
       }
     } else if (ceph_argparse_witharg(args, i, &val, "--max-size", (char*)NULL)) {
-      max_size = strict_si_cast<int64_t>(val.c_str(), &err);
+      max_size = strict_si_cast<long long>(val.c_str(), &err);
```
tchaikov: since the type of of `int64_t` is implementation defined, in other words, it's up to `cstdint` to make it an alias of `long int` or `long long`. maybe we can leave it as it is without the assumption? i know, technically `long long` is always 64bit long. but imho, it would be better use the type of the `max_size`. and be standard compliant. (c++ standard requires `long long` to be at least 64bit long)

src/rgw/rgw_http_client.cc
```diff
@@ -2,9 +2,8 @@
 // vim: ts=8 sw=2 smarttab
 
 #include "include/compat.h"
-
 #include <boost/utility/string_ref.hpp>
-
+#include <boost/utility/string_ref.hpp>
```
tchaikov: duplicated include

src/include/compat.h
```diff
@@ -116,25 +116,29 @@
 
 #if defined(HAVE_PTHREAD_SETNAME_NP)
   #if defined(__APPLE__)
-    #define pthread_setname_np(thread, name) ({ \
+    #define ceph_pthread_setname(thread, name) ({ \
       int __result = 0;                         \
       if (thread == pthread_self())             \
-        __result = pthread_setname_np(name)     \
+        __result = pthread_setname_np(name);    \
```
tchaikov: can we just keep this change? and the commit message is stale.

src/common/obj_bencher.cc
```diff
@@ -217,7 +218,8 @@ void *ObjBencher::status_printer(void *_bencher) {
 
 int ObjBencher::aio_bench(
   int operation, int secondsToRun,
-  int concurrentios, size_t op_size, size_t object_size,
+  int concurrentios,
+  uint64_t op_size, uint64_t object_size,
```
tchaikov: just want to note that size_t on arm is 32bit. if people want to reuse the bench after upgrade, he/she need to nuke the bench data.

tchaikov:
@ukernel 

- remove 757e722
- leave pthread_setname_np() unchanged in f0ac83f, and just keep the fixes for osx

other than the above nits, lgtm.

ukernel:
On Wed, Jul 13, 2016 at 7:56 PM, Kefu Chai <notifications@github.com> wrote:

> @ukernel <https://github.com/ukernel>
>
>    - remove 757e722
>    <https://github.com/ceph/ceph/commit/757e722314f968d69d01e00b9dd852183c879a1a>
>    - leave pthread_setname_np() unchanged in f0ac83f
>    <https://github.com/ceph/ceph/commit/f0ac83f21c23ea1b8a726b02ccfe4841ba0e3143>,
>    and just keep the fixes for osx
>
> OSX's pthread_setname_np only accept only one paramter. We can't use macro
to define a two parameters version.  Defining a two parameter static inline
function also does not work.

wjwithagen:
@ukernel @tchaikov 
I would still like to see the Conditional code in `src/common/ceph_time.h` go into compat.h.
And then a next step would be to cleanup the selective code that his in `src/common/ceph_time.c` so that there are no more ifdefs based on `*CLOCK*`

ukernel:
we need to declare clock_gettime() in ceph_time.h. I can't see how to not use ifdefs

src/include/compat.h
tchaikov: @ukernel

> Defining a two parameters version pthread_setname_np() does not work.

@dx9 has addressed this issue, what fails the the osx build is the ";", i think. so no need to define "ceph_pthread_{set,get}name()" right?

tchaikov:
@ukernel i think what @wjwithagen suggests is to

- move the `#if defined(DARWIN)` block into "compat.h" as an inlined function. as we don't have "compat.cc" yet.
- include "compat.h" in "ceph_time.h"

i am good either way.

ukernel:
I don't think include "compat.h" in ceph_time.h is good idea.

wjwithagen:
On 14-7-2016 09:41, Yan, Zheng wrote:
> I don't think include "compat.h" in ceph_time.h is good idea.

As far as I'm concerned every Ceph file should start with compat.h.
And even more so for the clock code, sinec we are trying to centralize
all the differences in one location.

src/include/compat.h
dx9: @ukernel @tchaikov sorry ; is my bad, I couldn't test easily on OSX.

gregsfortytwo:
For the record, I tried this and hit

>configure: error: in '/Users/gfarnum/src/ceph':
>configure: error: CryptoPP build selected but OpenSSL not found
>See 'config.log' for more details

That was with cryptopp and openssl installed via homebrew, and I couldn't make it go with various LDFLAG/CPPFLAG changes either (although I might have just done them wrong). I removed both from my system and just got

>configure: error: no suitable crypto library found

There is a note when you install cryptopp
>Generally there are no consequences of this for you. If you build your
>own software and it requires this formula, you'll need to add to your
>build variables:
>
>    LDFLAGS:  -L/usr/local/opt/openssl/lib
>    CPPFLAGS: -I/usr/local/opt/openssl/include

But when I tried that, configure failed instantly. Any idea what I'm doing differently? (I am running El Capitan 10.11.5)

ukernel:
I think configure failed to find pkgconfig file, try setting PKG_CONFIG_PATH=/usr/local/opt/......

tchaikov:
tested at

- http://pulpito.ceph.com/kchai-2016-07-24_02:17:13-rados-wip-kefu-testing---basic-mira/
- http://pulpito.ceph.com/kchai-2016-07-24_20:41:33-rados-wip-kefu-testing---basic-mira/

tchaikov:
retest this please

tchaikov:
the failure is a known issue.

f1-outsourcing:
I would like to use cephfs on macos, I have been trying this procedure http://docs.ceph.com/docs/mimic/dev/macos/. But run into errors.
Is there an other way to build this? (on 10.12/10.13)

tchaikov:
@f1-outsourcing normally, we don't report bug by commenting in pull requests. could you file a tracker ticket and accompany it with the error messages ?

PR #9372, opened 2016-05-27
jewel: rbd-mirror: Delete local image mirror when remote image mirroring is disabled

dillaman:
http://tracker.ceph.com/issues/15995


dachary:
test this please

PR #9374, opened 2016-05-27
client: use fh min readahead to enable readahead

batrick:
The client was using the configured maximum readahead of 0 which was changed in
95ee699. This prevented readahead from ever
running (without setting a different default from 0).

Resolves #16024.

### DIFF
--- a/src/client/Client.h
+++ b/src/client/Client.h
@@ -710,6 +710,7 @@ private:
     Client *client;
     Fh *f;
     C_Readahead(Client *c, Fh *f);
+    ~C_Readahead();
     void finish(int r);
   };
 
--- a/src/osdc/Striper.cc
+++ b/src/osdc/Striper.cc
@@ -241,10 +241,9 @@ uint64_t Striper::object_truncate_size(CephContext *cct,
 uint64_t Striper::get_num_objects(const file_layout_t& layout,
 				  uint64_t size)
 {
-  __u32 object_size = layout.object_size;
   __u32 stripe_unit = layout.stripe_unit;
   __u32 stripe_count = layout.stripe_count;
-  uint64_t period = (uint64_t)stripe_count * object_size;
+  uint64_t period = layout.get_period();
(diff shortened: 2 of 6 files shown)


gregsfortytwo:
The code looks good to me.
@batrick, please document what testing you've done to verify this.

We also want one of our nightly tests to verify that readahead is actually happening — we didn't notice it was broken previously!

src/osdc/Striper.cc
```diff
@@ -244,7 +244,7 @@ uint64_t Striper::get_num_objects(const file_layout_t& layout,
   __u32 object_size = layout.object_size;
```
gregsfortytwo: Build warning!

> osdc/Striper.cc:244:9: warning: unused variable 'object_size' [-Wunused-variable]

batrick: This is the one I told you about which is fixed in master. Should I rebase?

batrick: Nevermind, fixed.

gregsfortytwo:
retest this please

src/client/Client.h
```diff
@@ -709,7 +709,8 @@ class Client : public Dispatcher, public md_config_obs_t {
   struct C_Readahead : public Context {
     Client *client;
     Fh *f;
-    C_Readahead(Client *c, Fh *f);
+    C_Readahead(Client *c, Fh *f) : client{c}, f{f} {f->get(); f->readahead.inc_pending();};
+    ~C_Readahead() {f->readahead.dec_pending(); client->_put_fh(f);};
```
gregsfortytwo: Drop the spurious semi-colons please, and don't try to squash these into a single line — you can do a single-statement one-line function if you want, but not multiple statements like this.

gregsfortytwo:
http://pulpito.ceph.com/gregf-2016-06-10_19:20:53-fs-greg-fs-testing-610---basic-mira/
All failures attributed to AsyncMessenger incompatibilities.

PR #9375, opened 2016-05-27
rbd-mirror: replicate cloned images

dillaman:


dillaman:
Requires fixed from #9360

trociny:
@dillaman could you please rebase?

Also, add "Fixes: ..." to "rbd-mirror: add support ..." commit log?

qa/workunits/rbd/rbd_mirror.sh
```diff
@@ -416,19 +425,32 @@ test_status_in_pool_dir()
 create_image()
 {
     local cluster=$1
-    local image=$2
+    local pool=$2
+    local image=$3
 
-    rbd --cluster ${cluster} -p ${POOL} create --size 128 \
+    rbd --cluster ${cluster} -p ${pool} create --size 128 \
```
trociny: I think we need to fix all other functions too that use the hardcoded ${POOL}? test_image_replay_state, wait_for_replay_..., test_status_in_pool_dir...

qa/workunits/rbd/rbd_mirror.sh
```diff
@@ -219,94 +219,16 @@ admin_daemon()
     ceph --admin-daemon ${asok_file} $@
 }
 
-status()
```
trociny: Then also need to remove `status` command from the example section in the beginning of the script.

dillaman: I can put it back -- I saw that it wasn't in-use within the script so I yanked it.

trociny: It is useful (at least for me) when testing manually to check the current state, but sure I can live without it (especially if it adds additional work to maintain the script).

dillaman: No worries, I'll put it back and expand it to cover both pools.

dillaman:
@trociny re-added the ```status``` function

trociny:
lgtm
passed local testing
teuthology tests: http://pulpito.ceph.com/trociny-2016-05-31_22:06:08-rbd-wip-mgolub-testing---basic-vps/ (some jobs failed but nothing look related)

PR #9379, opened 2016-05-28
cmake: add ObjectMapIterate.cc to librbd

tchaikov:
it was introduced in a100f02


alimaredia:
@tchaikov can you take a look at the second commit I added to this PR?

PR #9385, opened 2016-05-28
cmake: more fixes

tchaikov:


src/CMakeLists.txt
```diff
@@ -2,6 +2,12 @@ include(GetGitRevisionDescription)
 
 enable_language(C ASM)
 include(GNUInstallDirs)
+set(bindir ${CMAKE_INSTALL_BINDIR})
```
tchaikov: @alimaredia i broke some of configure_file() in 7a602ec. so fix it here.

tchaikov:
yellow at http://gitbuilder.sepia.ceph.com/gitbuilder-ceph-tarball-trusty-amd64-cmake/#origin/wip-cmake-kefu .

i also have a fix to enable the static and shared library building for some of the user facing libraries, but it breaks the linkage of a handful of test executables. didn't get a chance to look into it.

tchaikov:
@cbodley @alimaredia mind taking a look?


todos before "fakeroot debian/rule binary":

- [ ] cython modules installation
- [ ] shared and static libraries building: testing. the change breaks the linkage of lots tests.
- [ ] ceph_test_* files not installed: i doubt that we should install or package at all)
- [ ] usr/lib/ceph/erasure-code/* not installed
- [ ] ceph-disk not installed into usr/sbin
- [x] use cmake in debian/rules

mattbenjamin:
at the moment, on my fedora 23 system, I get:
1. libdir/CMAKE_INSTALL_LIBDIR (src/CMakeLists.txt) is apparently <prefix>/lib
2. at runtime, dlopen of ecplugins is 
2.1 a relative path apparently assumed to be <prefix>
2.2 lib64/* -- which fails to load even when the program is run relative to <prefix>, since the objects were installed in lib
Of course, we don't want a dependency on running from the install prefix--Casey remarked that this works from vstart, but agrees that we didn't want to build that idea into the binary, or at least, that seems questionable;  Didn't the base path for finding ecplugins used to be sourced from ceph.conf (or was that just something cohort added?)

tchaikov:
> 1. libdir/CMAKE_INSTALL_LIBDIR (src/CMakeLists.txt) is apparently /lib

i think it depends on `CMAKE_INSTALL_PREFIX`.

> 2. at runtime, dlopen of ecplugins is 
> 2.1 a relative path apparently assumed to be 
> 2.2 lib64/* -- which fails to load even when the program is run relative to , since the objects were installed in lib

i see. we need to install them into lib64 on redhat and SuSE distros also to follow FHS. will change that also: use ${libdir} instead of "lib".

> Of course, we don't want a dependency on running from the install prefix

yeah, but i think it should at least be consistent about the assumptions of the default paths without the direction of ceph.conf.

tchaikov:
> we need to install them into lib64 on redhat and SuSE distros also to follow FHS. will change that also: use ${libdir} instead of "lib".

probably not now. will continue to use "lib" at this moment. as the deb and rpm packaging scripts are expecting "lib". will update this PR to change "lib64" back to "lib".

maybe we should use "lib64" later on, as redhat and SuSE are respecting FHS. and Debian is using it's own multiarch paths specification. and all these details are dealt by the "GNUInstallDirs.cmake".

src/CMakeLists.txt
```diff
@@ -1009,9 +1030,9 @@ install(TARGETS osdmaptool DESTINATION bin)
 
 set(ceph_psim_srcs
   tools/psim.cc)
-add_executable(ceph-psim ${ceph_psim_srcs})
-target_link_libraries(ceph-psim global)
-install(TARGETS ceph-psim DESTINATION bin)
+add_executable(ceph_psim ${ceph_psim_srcs})
```
alimaredia: @tchaikov (NIT) I was under the impression that executables had '-'s where spaces should go and libraries and other variables should have _'s where spaces should go. Does this make sense for consistency sake and if so can we change the packaging for ceph_psim to ceph-psim?

alimaredia:
@tchaikov I haven't cloned it and played around with it myself, but it looks good from taking a look at all the changes.

Could you explain what "fakeroot debian/rule binary" is and how it furthers CMake?

src/CMakeLists.txt
tchaikov: it does make sense. but i thought we should limit the scope of this PR to cmake only changes. will add a commit to change the packaging and automake also.

tchaikov:
"debian/rule" is the packaging script used to generate the deb packages. it's basically a Makefile. "fakeroot debian/rule binary" is used to generate binary packages[1] without cleaning the source tree. by running it, i can make sure all the artifacts are compiled, installed, and are placed at the expected places at least under debian distros after "make install". but this approach also has its limitation: as @mattbenjamin pointed out, on fedora (and rhel), the file arrangement is somewhat different, so i didn't notice that.

---
[1] https://www.debian.org/doc/manuals/maint-guide/build.en.html#completebuild

src/CMakeLists.txt
tchaikov: @alimaredia i reconsidered this change a little bit: actually there are a bunch of executables named as "ceph_" in ceph-test package, see ceph.spec.in.  probably we can change them all at once, if you think we should go this way. and please note this is a user visible change. i'd like to put them into another PR, so it's more visible to other reviewers. what do you think?

tchaikov:
tested. yellow at http://gitbuilder.sepia.ceph.com/gitbuilder-ceph-tarball-trusty-amd64-cmake/#origin/wip-cmake-kefu.

cbodley:
@tchaikov related to `make install`, i've found that i can no longer use a `CMAKE_INSTALL_PREFIX` under my home directory, because it still tries to put `mount.ceph` under /sbin

mattbenjamin:
kefu--I've verified the most recent python setup and install change, which I need for my workflow;  the the best of my knowledge, you haven't broken anything we depend on :) -- but we were badly broken before, lacking installation of modules built w/cython, so I think this is ready

mattbenjamin:
If you have more things coming incrementally, let's split the new stuff here, and start a new PR?

tchaikov:
@cbodley 

> related to `make install`, i've found that i can no longer use a `CMAKE_INSTALL_PREFIX` under my home directory, because it still tries to put `mount.ceph` under /sbin

i have the same issue. is there any better way we can install `mount.ceph` into /sbin? it was installed into `/usr/sbin`. but the `ceph-fs-common.install` expects it in `/sbin`. i will fix this by revert this change and point it to `/usr/sbin/mount.ceph` instead.

PR #9392, opened 2016-05-30
rgw/rgw_common.cc: modify the end check in RGWHTTPArgs::sys_get

yehudasa:
Fixes:  http://tracker.ceph.com/issues/16072

Signed-off-by: zhao kun <jswps2011@163.com>


yehudasa:
@oritwas I think this can be safely merged

PR #9395, opened 2016-05-30
buffer: handle integer underflow in iterator::copy(large_int, dest)

tchaikov:


src/common/buffer.cc
```diff
@@ -1068,7 +1068,9 @@ static simple_spinlock_t buffer_debug_lock = SIMPLE_SPINLOCK_INITIALIZER;
   template<bool is_const>
   void buffer::list::iterator_impl<is_const>::advance(unsigned o)
   {
-    p_off += o;
+    if (__builtin_add_overflow(p_off, o, &p_off)) {
```
tchaikov: it's but precaution. i am fine if any of you guys wants to drop this change.

tchaikov: may be we can use NDEBUG or something alike to enable debug-only things. like in many other projects. but i checked the definition of `assert()`, seems we don't have such a thing yet.

xiexingguo:
@tchaikov Does not compile :-)

tchaikov:
per @ukernel

> cephfs supports dir fragments. it shouldn't  have so many omap items in single object

and CDir::_omap_fetched() is not able to fetch omap in multiple batches. so we can not lower the max-return in omapgetvals*() calls.

tchaikov:
changelog:

- detect the availability of `__builtin_add_overflow()` before using it.

src/common/buffer.cc
tchaikov: if we throws here, we won't be able to add `noexcept` specifier to these methods. 

quote from Effective Modern C++
> In a noexcept function, optimizers need not keep the runtime stack in an unwindable state if an exception would propagate out of the function, nor must they ensure that objects in a noexcept function are destroyed in the inverse order of construction should an exception leave the function. Functions with “throw()” exception specifications lack such optimization flexibility,

tchaikov:
@athanatos and @liewegas could you take a quick look at this? it addresses an issue @xiaoxichen ran into.

liewegas:
Hmm, can we just make these types size_t/unsigned long?  We will never have a buffer in memory that is larger than the size of a pointer, so they should never overflow.  Or am I missing something?

tchaikov:
@liewegas 
>  can we just make these types size_t/unsigned long? 

we can change in the following way, so when we pass an large enough "unsigned" integer it won't change from a positive number to a negative one.

```diff
-- void advance(int);
++ void advance(ssize_t);
```

but this practically changes the function signature. and the ABI of the function is changed this way. 

> We will never have a buffer in memory that is larger than the size of a pointer, so they should never overflow. 

see http://tracker.ceph.com/issues/16010#note-8, this does happen in real world. @xiaoxichen could you fill it with more details?

xiaoxichen:
What kind of information is helpful now?

xiaoxichen:
But after applying this patch, MDS doesn't complain anymore on " failed to decode message"

tchaikov:
> What kind of information is helpful now?

about why this happens in real world. for example, why directory fragment is disabled by default [1]. and how many files you put in a single directory, and why is it a normal use case.

---
[1] https://github.com/ceph/ceph/blob/master/doc/cephfs/experimental-features.rst#directory-fragmentation

xiaoxichen:
@liewegas , I put 1.6M+(was 4M+ in ticket, but reproceable with 1.6M+) files in one dir, and ls the dir can trigger the bug.  We do have user (CI as a service) that putting ~0.5M small files in one dir and has 100+ this kind of dir , not sure if really have someone want 1.6M.

We do this is just to see if there is any maximum #files per dir or per MDS can hold. It's OK to say cephFS had some headroom that limited #files per dir to 1M or somewhat value. But the problem is:

1.  Once the dir has more files than this, "ls" can create large message and kill the OSD by assert(uint32 overflow). So everytime the user do "ls" it will take down one OSD, which is horrible.

2.  We don't have workable capping / monitoring mechanism to prevent 1) from happening, and the limit is not too large to reach even with this fix(int32 ->uint32). Unless we make it int64.Capping on home directory is not a good solution.

3.  Dir fragment (by bal_frag) doesn't solve the problem because MDS-> Client message will overflow although we avoided the overflow between OSD -> MDS

4.  Does "ls -R" on home dir can trigger the bug even any of the subdir has < 1M files ? 

5.   Does reading a big object (4GB+) via rados API can also trigger the bug?

ukernel:
On Mon, Jun 6, 2016 at 11:18 AM, Xiaoxi Chen <notifications@github.com> wrote:
>
> @liewegas , I put 1.6M+(was 4M+ in ticket, but reproceable with 1.6M+) files in one dir, and ls the dir can trigger the bug. We do have user (CI as a service) that putting ~0.5M small files in one dir and has 100+ this kind of dir , not sure if really have someone want 1.6M.
>
> We do this is just to see if there is any maximum #files per dir or per MDS can hold. It's OK to say cephFS had some headroom that limited #files per dir to 1M or somewhat value. But the problem is:
>
> Once the dir has more files than this, "ls" can create large message and kill the OSD by assert(uint32 overflow). So everytime the user do "ls" it will take down one OSD, which is horrible.
>
> We don't have workable capping / monitoring mechanism to prevent 1) from happening, and the limit is not too large to reach even with this fix(int32 ->uint32). Unless we make it int64.Capping on home directory is not a good solution.
>
> Dir fragment (by bal_frag) doesn't solve the problem because MDS-> Client message will overflow although we avoided the overflow between OSD -> MDS

There are several mechanism (readdir.max_bytes, readdir.max_entries,
dirfrag)  limit size of readdir reply.  I don't think it can overflow
in practice

xiaoxichen:
@ukernel ,what configuration item should I apply for the readdir.maxbyte/max_entries?  What's the behavior if #files > max_entries?

liewegas:
Hmm, it seems like the OSD shouldn't crash when there are lots of omap entries.  We can't currently prevent you from creating them because we don't have a counter.

1. we need to fix the MDS to not let a dirfrag get too big.  This should be pretty straightforward using the fnode_t entry count fields.
2. The buffer code shoudl be fine with size_t as it is talking about data in memory.
3. The calling omap code have a cap on the number of entries that is actually enforced.
4. It seems like the ObjectStore omap_get and omap_get_keys methods shouldn't exist since they have an unbounded result size.  Callers should use the iterator interface and handle the many-key case accordingly.

liewegas:
@ukernel I'm guessing the problem here is that the cdir fetch method tries to load up the entire directory, regardless of what readdir (or cache miss or whatever) triggered it.  It needs to be smarter... not just for this case, but also because the dir could be larger than the cache, and loading big dirs in general is probably a bad idea.  Probably fetching N entries starting from (or around) the given key is what we want.  It might even make sense to have an op like "read up to N keys, including key K".  For small dirs it would be the entire dir, and for big dirs it would be whatever the OSD decides it can do efficiently.

tchaikov:
> The buffer code shoudl be fine with size_t as it is talking about data in memory.

okay, will use `ssize_t` and `size_t` when appropriate in buffer.

liewegas:
I opened http://tracker.ceph.com/issues/16164 for the mds issue.
And http://tracker.ceph.com/issues/16172 for the librados/osd side.

This pr can just clean up the buffer code to use size_t?

xiaoxichen:
Sorry, one question,  when 32-bit OSDs talking with 64-bit OSDs,  will the decoding failed as they have different size_t definition ?

tchaikov:
@xiaoxichen no, the bufferlist encoding won't change in this PR.

tchaikov:
changelog

- remove the assert() detecting integer overflow
- change `advance(unsigned)` and `seek(int)` to `advance(size_t)` and `seek(ssize_t)`.

liewegas:
LGTM.

I hope/pray we don't need to worry about ABI changes with buffer.h.  Or at least agree to ignore them between major updates (i.e., don't backport this to jewel).   @jdurgin

jdurgin:
@liewegas seems we're breaking the ABI in buffer.h pretty regularly across major releases (in the past just by using c++11) and there's been no reported issues (how many c++ users of librados are there?), so this seems ok to me, especially as there's not really a great alternative for this fix

tchaikov:
tested at 

xiaoxichen:
@tchaikov  can this be backport to jewel?

tchaikov:
@xiaoxichen, i am not sure about this. as we are breaking the backward compatibility of the C++ ABI in buffer.h as pointed out by @jdurgin @liewegas . if you really want this to happen, maybe you can bring up some discussion over ceph-devel? it's difficult for me to make this decision on my own.

dillaman:
@tchaikov FYI -- this change breaks ABI compatibility:

> 2016-11-03T10:52:33.094 INFO:tasks.workunit.client.0.vpm047.stderr:ceph_test_librbd_api: symbol lookup error: ceph_test_librbd_api: undefined symbol: _ZN4ceph6buffer4list8iterator7advanceEi
> 2016-11-03T10:52:33.096 INFO:tasks.workunit:Stopping ['rbd/test_librbd_api.sh'] on client.0...

http://qa-proxy.ceph.com/teuthology/trociny-2016-11-03_10:36:34-upgrade-wip-mgolub-testing---basic-vps/513867/teuthology.log

tchaikov:
@dillaman ack, will take a closer look tmr.

tchaikov:
> I hope/pray we don't need to worry about ABI changes with buffer.h. Or at least agree to ignore them between major updates (i.e., don't backport this to jewel). 

this breaks the old librbd client, now it is not able to run with new librbd. and this was my concern at  https://github.com/ceph/ceph/pull/9395#issuecomment-223609290.  so maybe we should instead add a new method? @liewegas @jdurgin ?

trociny:
http://tracker.ceph.com/issues/17809

PR #9398, opened 2016-05-30
Wip multi object write transaction support

dragonylffly:
This patch implements multi-object write transaction. It is originally proposed by Sage at CDS for 
Infernalis (http://tracker.ceph.com/projects/ceph/wiki/Osd_-_Transactions), with more discussion at  CDS for Jewel (http://tracker.ceph.com/projects/ceph/wiki/Rados_-_multi-object_transaction_support). The support for multi-object write transaction at RADOS could potentially greatly simplify the client design, such as CephFS, which can expect many file system calls finish atomically, eliminating the need of journals. Similar idea is presented at  https://www.usenix.org/conference/fast16/technical-sessions/presentation/shin. 

It is theoretically fully on top of the current RADOS mechanism, and in terms of implementation, it keeps the minimal intrusion to the ReplicatedPG code.

src/test/librados/multiobjectwriteoperation.cc gives the usage example, and ceph_test_rados_api_multiobjectwriteoperation runs the test.

### DIFF
--- a/doc/dev/osd_internals/multi_object_write_transaction.rst
+++ b/doc/dev/osd_internals/multi_object_write_transaction.rst
@@ -1,0 +1,113 @@
+==============================
+Multi-Object Write Transaction
+==============================
+
+Multi-object write transaction enables a group of write operations
+on multi-objects complete atomically.
+
+Description of the Process
+--------------------------
+
+The high level process is,
+
+  1. Define the following class,
+
+     class *CEPH_RADOS_API MultiObjectWriteOperation*
+
+     {
+
+       std::map<std::string, ObjectWriteOperation*> slaves;
+
+       ObjectWriteOperation master;
+
+     };
+
+     based on this, client could send a group of ops on multi-objects
+     through *MOSDOp* to the primary osd in charge of the *master* object
+     which is explicitly specified by client, call this osd *MASTER*
+
+  #. *MASTER* recieves the multi-object write operation request, writes a
+     *LOCK* entry into *PGLOG*, creates a meta object, and records the *slaves*
+     information in it;
+
+  #. *MASTER* extracts operations by objects from *slaves*, enclose them in
+     *LOCK* requests, each request includes a group of operations on a single
+     slave object. *MASTER* sends *LOCK* requests to corresponding primary osds
+     individually, call those osds "SLAVE*;
+
+  #. *SLAVE* recieves *LOCK* request, confirms the validity of the operation,
+     then writes a *LOCK* entry into *PGLOG*, creates a meta object, and
+     records the transaction information in it, then sends ack to *MASTER*;
+
+  #. *MASTER* collects response, then does the operations on the *master*
+     object, besides, writes a *COMMIT* entry into *PGLOG*. *MASTER* sends
+     *COMMIT* request to *SLAVE*, then sends ack to client;
+
+  #. *SLAVE* receives *COMMIT*, performs the operations on the slave object,
+     besides, writes a *COMMIT* entry into *PGLOG*. *SLAVE* sends ack to
+     *MASTER*, deletes the meta object and writes a *UNLOCK* entry into
+     *PGLOG*;
+
+  #. *MASTER* collects ack, deletes the meta object and writes a *UNLOCK*
+     entry into *PGLOG*
+
+Fault Tolerance
+---------------
+When restarting, the osd re-creates the transaction in memory according to
+the *PGLOG* entry and meta object
+
+- *MASTER* restart
+
+ - If the transaction is in *LOCK* state
+
+ (1) *MASTER* sends *UNLOCK* to *SLAVE*;
+
+ (2) *SLAVE* receives *UNLOCK*, if the transaction does not exist or in
+ *UNLOCK* state, goto 3; If the transaction is in *LOCK* state, *SLAVE*
+ rolls back;
+
+ (3) *SLAVE* sends ack to *MASTER*;
+
+ (4) *MASTER* collects ack, and rolls back itself.
+
+ - If the transaction is in *COMMIT* state
+
+ (1) *MASTER* sends *COMMIT* to *SLAVE*;
+
+ (2) *SLAVE* receives *COMMIT*, if the transaction is not in *LOCK* state,
+ goto 3; Otherwise, proceed as normal;
+
+ (3) *SLAVE* sends ack to *MASTER*;
+
+
+ (4) *MASTER* collects ack, proceed as normal
+
+- SLAVE restart
+
+ - If the transaction is in *LOCK* state
+
+ *Slave* waits *MASTER* to re/send *LOCK/UNLOCK/COMMIT*
+
+ - If the transaction is in *COMMIT* state
+
+ *SLAVE* does the *UNLOCK* process; if *MASTER* resends *COMMIT*, it directy
+ sends ack to *MASTER*
+
+Dead Lock Avoidance
+-------------------
+
+When a osd receives a new multi-object write operation request, if there is
+pending multi-object write operation operating on the same object,
+return *EDEADLK*, except these two cases:
+
+  1. The osd plays *MASTER* for the new operation;
+
+  2. The osd plays *SLAVE* for the new operation, and the pending operation
+  has entered *COMMIT* step;
+
+In these two cases, the new request could be delayed until the pending one
+finishes
+
+When a new single object write operation request comes, if there is pending
+multi-object write operation operating on the same object, the new request must
+be delayed until the pending one finish
--- a/src/common/config_opts.h
+++ b/src/common/config_opts.h
@@ -814,6 +814,16 @@ OPTION(osd_debug_skip_full_check_in_backfill_reservation, OPT_BOOL, false)
 OPTION(osd_debug_reject_backfill_probability, OPT_DOUBLE, 0)
 OPTION(osd_debug_inject_copyfrom_error, OPT_BOOL, false)  // inject failure during copyfrom completion
 OPTION(osd_debug_randomize_hobject_sort_order, OPT_BOOL, false)
+OPTION(osd_debug_multi_object_write_operation_master_lock_self_crash, OPT_BOOL, false)
+OPTION(osd_debug_multi_object_write_operation_master_lock_slave_crash, OPT_BOOL, false)
+OPTION(osd_debug_multi_object_write_operation_master_commit_self_crash, OPT_BOOL, false)
+OPTION(osd_debug_multi_object_write_operation_master_commit_slave_crash, OPT_BOOL, false)
+OPTION(osd_debug_multi_object_write_operation_master_unlock_self_crash, OPT_BOOL, false)
+OPTION(osd_debug_multi_object_write_operation_master_unlock_slave_crash, OPT_BOOL, false)
+OPTION(osd_debug_multi_object_write_operation_slave_lock_crash, OPT_BOOL, false)
+OPTION(osd_debug_multi_object_write_operation_slave_commit_crash, OPT_BOOL, false)
+OPTION(osd_debug_multi_object_write_operation_slave_unlock_crash, OPT_BOOL, false)
+OPTION(osd_multi_object_operation_namespace, OPT_STR, ".ceph-mooc") // rados namespace for multi operation
 OPTION(osd_enable_op_tracker, OPT_BOOL, true) // enable/disable OSD op tracking
 OPTION(osd_num_op_tracker_shard, OPT_U32, 32) // The number of shards for holding the ops
 OPTION(osd_op_history_size, OPT_U32, 20)    // Max number of completed ops to track
--- a/src/include/rados.h
+++ b/src/include/rados.h
@@ -173,6 +173,7 @@ extern const char *ceph_osd_state_name(int s);
 #define CEPH_OSD_OP_TYPE_EXEC  0x0400
 #define CEPH_OSD_OP_TYPE_PG    0x0500
 #define CEPH_OSD_OP_TYPE_MULTI 0x0600 /* multiobject */
+#define CEPH_OSD_OP_TYPE_MOC   0x0700 /* multi object operation control */
 
 #define __CEPH_OSD_OP1(mode, nr) \
 	(CEPH_OSD_OP_MODE_##mode | (nr))
@@ -260,6 +261,11 @@ extern const char *ceph_osd_state_name(int s);
 	/* ESX/SCSI */							    \
 	f(WRITESAME,	__CEPH_OSD_OP(WR, DATA, 38),	"write-same")	    \
 									    \
+	/* multi object slave op*/				            \
+        f(MOC_LOCK,   __CEPH_OSD_OP(WR, MOC, 1),   "moc-lock")              \
+        f(MOC_COMMIT, __CEPH_OSD_OP(WR, MOC, 2),   "moc-commit")            \
+        f(MOC_UNLOCK, __CEPH_OSD_OP(WR, MOC, 3),   "moc-unlock")            \
+                                                                            \
 	/** multi **/							    \
 	f(CLONERANGE,	__CEPH_OSD_OP(WR, MULTI, 1),	"clonerange")	    \
 	f(ASSERT_SRC_VERSION, __CEPH_OSD_OP(RD, MULTI, 2), "assert-src-version") \
@@ -326,6 +332,10 @@ static inline int ceph_osd_op_type_pg(int op)
 static inline int ceph_osd_op_type_multi(int op)
 {
 	return (op & CEPH_OSD_OP_TYPE) == CEPH_OSD_OP_TYPE_MULTI;
+}
+static inline int ceph_osd_op_type_moc(int op)
+{
+	return (op & CEPH_OSD_OP_TYPE) == CEPH_OSD_OP_TYPE_MOC;
 }
 
 static inline int ceph_osd_op_mode_subop(int op)
--- a/src/include/rados/librados.hpp
+++ b/src/include/rados/librados.hpp
@@ -27,6 +27,7 @@ namespace librados
   class IoCtx;
   struct IoCtxImpl;
   class ObjectOperationImpl;
+  class MultiObjectWriteOperation;
   struct ObjListCtx;
   struct PoolAsyncCompletionImpl;
   class RadosClient;
@@ -361,6 +362,7 @@ namespace librados
     ObjectOperation& operator=(const ObjectOperation& rhs);
     friend class IoCtx;
     friend class Rados;
+    friend class MultiObjectWriteOperation;
   };
 
   /*
@@ -482,6 +484,18 @@ namespace librados
     void cache_unpin();
 
     friend class IoCtx;
+  };
+
+  class CEPH_RADOS_API MultiObjectWriteOperation
+  {
+  protected:
+    std::map<std::string, ObjectWriteOperation*> slaves;
+    void prepare_operate();
+    friend class IoCtx;
+  public:
+    ObjectWriteOperation master;
+    ObjectWriteOperation& slave(const std::string &oid);
+    virtual ~MultiObjectWriteOperation();
   };
 
   /*
@@ -983,8 +997,10 @@ namespace librados
 	         bufferlist& inbl, bufferlist *outbl);
 
     // compound object operations
+    int operate(const std::string& oid, MultiObjectWriteOperation *op);
     int operate(const std::string& oid, ObjectWriteOperation *op);
     int operate(const std::string& oid, ObjectReadOperation *op, bufferlist *pbl);
+    int aio_operate(const std::string& oid, AioCompletion *c, MultiObjectWriteOperation *op);
     int aio_operate(const std::string& oid, AioCompletion *c, ObjectWriteOperation *op);
     int aio_operate(const std::string& oid, AioCompletion *c, ObjectWriteOperation *op, int flags);
     /**
--- a/src/librados/IoCtxImpl.cc
+++ b/src/librados/IoCtxImpl.cc
@@ -766,6 +766,14 @@ int librados::IoCtxImpl::aio_operate(const object_t& oid,
   if (snap_seq != CEPH_NOSNAP)
     return -EROFS;
 
+  if (o->sub_ops.size()) {
+    bool kraken = objecter->with_osdmap([](const OSDMap& osdmap) {
+      return osdmap.test_flag(CEPH_OSDMAP_REQUIRE_KRAKEN);
+    });
+    if (!kraken)
+      return -EOPNOTSUPP;
+  }
+
   Context *onack = c->wants_ack() ? new C_aio_Ack(c) : NULL;
   Context *oncommit = new C_aio_Safe(c);
 
--- a/src/librados/librados.cc
+++ b/src/librados/librados.cc
@@ -584,6 +584,34 @@ void librados::ObjectWriteOperation::cache_unpin()
   o->cache_unpin();
 }
 
+void librados::MultiObjectWriteOperation::prepare_operate()
+{
+  ::ObjectOperation *op = &master.impl->o;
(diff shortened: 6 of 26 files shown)


dragonylffly:
@liewegas , @athanatos , your comments are appreciated

mattbenjamin:
@adamemerson worth a look

liewegas:
@dragonylffly Do you mind adding an rst to doc/dev with a high-level overview of how this works?  What the states are, overall flow of ops, that sort of thing. 

I did a quick skim of the code and it looks pretty reasonable!  I think the key/hard parts are going to be making sure this interacts with peering events and restarts properly to continue or rollback transactions.  It's hard to review that without a high-level understanding of the process.  Thanks!

athanatos:
@dragonylffly I second the request for a doc/dev/osd_internals overview of
1) What problem it is meant to solve
2) A high level description of the protocol used to do it
3) A high level description of the code architecture used to implement it

dragonylffly:
@liewegas @athanatos sure, working on it

athanatos:
There are a bunch of whitespace issues in the branch, fix them.

athanatos:
Also, limit lines to 80 characters.

athanatos:
@dragonylffly Are you sticking the entire encoded multi-object-transaction into an xattr?  That would be a big problem.

athanatos:
I'm going to leave off further review until the formatting issues are cleaned up and there is a commit in doe/dev/osd_internals explaining how the protocol and implementation work (as described above).

dragonylffly:
@liewegas @athanatos updated with the document added, please review

doc/dev/osd_internals/multi_object_write_transaction.rst
```diff
@@ -0,0 +1,117 @@
...
+
+  #. *SLAVE* receives *MOSDOp* from *MASTER*, note each op received by *SLAVE*
+     is a group of write operations on a single slave object. *SLAVE* 
+     performs dead lock check, does the normal *prepare_transaction()* stuff, 
+     if everything is ok, *SLAVE* will lock the object, and delay further 
+     read/write operations on this object until the transaction finishs. 
+     *SLAVE* sends a *QUERY* message to *MASTER*, and returns *MASTER* 
+     *EINPROGRESS*. The *QUERY* message is used for *MASTER* to notify *SLAVE* 
+     when to release transaction context, *MASTER* will not ack *QUERY* until 
+     itself has released transaction context
```
athanatos: At this point, it doesn't actually write anything durably, right?

doc/dev/osd_internals/multi_object_write_transaction.rst
```diff
@@ -0,0 +1,117 @@
...
+
+  #. *MASTER* collects response from *SLAVE*, if all *EINPROGRESS*, then sends
+     *SLAVE* *LOG* message, the transaction enters the *LOG* state
+     
+  #. *SLAVE* receives *LOG*, constructs a write op which persists the 
+     multi-object operation information in the object attribute named 
+     *MOWO_ATTR*, and also synchronizes the op within its replica osds,
```
athanatos: This xattr would possibly contain MBs of write data, right?  That's not going to work.  We'll want to write it out to a temp object, I think.  If we play our cards right, we might be able move the data from the temp object into the target object without needing to do a second write using a bluestore move_range primitive.

doc/dev/osd_internals/multi_object_write_transaction.rst
```diff
@@ -0,0 +1,117 @@
...
+  #. *SLAVE* receives *LOG*, constructs a write op which persists the 
+     multi-object operation information in the object attribute named 
+     *MOWO_ATTR*, and also synchronizes the op within its replica osds, 
+     utilizing the present single object write transaction mechanism, then 
+     returns *MASTER* zero
+     
+  #. *MASTER* collects response from *SLAVE*, if all zero, persists the 
+     multi-object operation information itself, then performs the write op on 
+     the master object, and synchronizes the op within its replica osds, return 
+     client *ACK*
```
athanatos: To be clear, it sends a commit to the client at this point, right?

doc/dev/osd_internals/multi_object_write_transaction.rst
```diff
@@ -0,0 +1,117 @@
...
+     
+  #. *MASTER* sends *SLAVE* *COMMIT & UNLOCK* message, the transaction enters 
+     the *COMMIT* state
+     
+  #. *SLAVE* receives *COMMIT & UNLOCK* message, first processes *COMMIT*, 
+     performs the write op on slave object, and synchronizes the op within its
+     replica osds, returns *MASTER* zero, *SLAVE* processes *UNLOCK*, cleans up
```
athanatos: These two bits can easily be combined, right?

doc/dev/osd_internals/multi_object_write_transaction.rst
```diff
@@ -0,0 +1,117 @@
...
+     
+     
+Fault Tolerance
+---------------
+
+- If *SLAVE* down at any time before persisting transaction to object 
+  attribute, when *MASTER* resends *PREPARE* or (re)sends *LOG*, since *SLAVE* 
+  has lost the in-memory transaction information, it will return *EPERM*. If 
+  *MASTER* receives *EPERM*, it recognizes the situation and redo *PREPARE* for 
+  this *SLAVE* to resend the operation to it to proceed
```
athanatos: The master needs to know to do this.  Is the query message system part of the Objecter where this would happen automatically, or is it a special message sent out-of-band from the Objecter?  I think the best way to handle this would be to structure the PREPARE/LOG portion as a linger op hidden behind the Objecter.  Otherwise, we risk having to reimplement a significant portion of the objecter's interval and map tracking stuff.  Also, hiding it behind the objecter also makes it a lot more trivial to deal with different pools.

doc/dev/osd_internals/multi_object_write_transaction.rst
athanatos: This seems wrong.  master can't actually count on any of this because an interval change would cause the slave primary to forget the lock state, so the responses from the slaves are essentially useless.  Blocking reads and writes doesn't actually become reliable until you've persisted that a transaction is underway in a manner that survives an interval change (like performing a repop).  It seems like this initial prepare needs to be merged into the log step with a success response happening only once the transaction is logged durably.

doc/dev/osd_internals/multi_object_write_transaction.rst
```diff
@@ -0,0 +1,117 @@
...
+  when *MASTER* resends *LOG*/sends *COMMIT*, *SLAVE* will reconstruct the 
+  transaction from the object attribute and proceed. If *SLAVE* down during 
+  *UNLOCK*, when *MASTER* finishes transaction, it will ack *QUERY* to the 
+  *SLAVE* to release the transaction to proceed
+
+- If *MASTER* down at any time after persisting transaction to object 
+  attribute, when client resends request/*SLAVE* resends *QUERY*, *MASTER* will 
+  reconstruct the transaction from the object attribute and send *COMMIT* to 
+  *SLAVE* to proceed
+
```
athanatos: I think there is a case missing here.  If the master undergoes an interval change and the client dies, the slave is left holding the bag on an in-progress transaction which blocks writes and reads on an object.  The slave would need to track whether the master has undergone an interval change and send a query each time -- it can't rely on the client resending the request.

athanatos:
I don't like the way in which the in progress transaction state is being persisted.  I think it is desirable that after an interval change, the primary know about all transactions it is holding state for (all transactions for which the local participating object has entered the LOG/PREPARE state and has reads and writes blocked).  I think the simplest way to do this would be to persist the state for in-progress transactions in a set of omap keys on the PG meta object with the actual transaction data written to temp objects.  State transitions for a transaction should be represented using PG log entries.

When the primary goes active, for each transaction it is master for, it should resend the messages for the current transaction state.  For a slave, I'm not actually sure.  As long as the master persists a MASTER_PREPARE state prior to sending out the initial wave of PREPARE/LOG requests, the slave can actually assume that the master will resend.  Otherwise, the master would only resend in the case that it persisted LOG/PREPARE (in which case, the transaction will roll forward).  This can be worked around by having the slaves send a query if they see from the OSDMap that the master went through an interval change.  The master would then either roll forward or back depending on its local state.

doc/dev/osd_internals/multi_object_write_transaction.rst
dragonylffly: Yep, at PREPARE, it does not persist anything. From the aspect of correctness, PREPARE and LOG could be merged. PREPARE is kept separate for better performance.  As also mentioned by @liewegas , before doing any real transaction stuff, a lightweight in-memory only dead lock and other validity check is desirable, if we do LOG first for all SLAVEs, the roll back overhead will be much higher.

If we agree to keep PREPARE alone, then after PREPARE, any slave must immediately lock the object (in memory) to prevent the object being modified by further write operations, before the transaction finishes. If the SLAVE undergoes an interval change after PREPARE, yes, the new SLAVE will forget the lock and transaction, the MASTER will send LOG/resend PREPARE to the new SLAVE, now, there are two cases,
(1) For the first case, there is no new transaction intervening, then the SLAVE will respond MASTER EPERM to indicate that it does not know this transaction at all, then MASTER will resend PREPARE to this SLAVE, and the SLAVE will redo validity check, lock the object, proceed as before;
(2)	For the second case, there is other transaction intervening, then the SLAVE will find there is other pending transaction on this object, it will fail the dead lock check, and respond MASTER EDEADLK, then MASTER will notify other SLAVEs to roll back.
In either of the cases, it will not cause problem and the semantics of transaction will not be violated. Since the real object modification happens only after MASTER has collected all LOG success message from SLAVES, at that time, all the lock have been persisted.

doc/dev/osd_internals/multi_object_write_transaction.rst
dragonylffly: Exactly

doc/dev/osd_internals/multi_object_write_transaction.rst
dragonylffly: In fact, they are two operations in the same OSDOp message

doc/dev/osd_internals/multi_object_write_transaction.rst
dragonylffly: In fact, this will not be a problem. If both MASTER and client undergo an interval change, the SLAVE will resend QUERY to MASTER, MASTER will respond with EPERM, and then SLAVE will roll back the transaction. Even if SLAVE also undergoes an interval change, when the object is accessed, SLAVE will immediately send QUERY to the MASTER recorded in the persisted transaction, and MASTER will respond with EPERM, and then SLAVE will roll back the transaction

dragonylffly:
In fact, we did not choose to persist the entire transaction at MASTER side because it incurs higher overhead than letting SLAVE to persist their parts separately. If MASTER persists the entire transaction before communicating with SLAVES, it needs two write: one for LOG the transaction which is big operation and the other for committing its part. In our design, MASTER only needs one write: LOG and COMMIT its part done once. Our design, although seems a little bit more tricky, count on the collaboration of MASTER, client and SLAVE, but it makes the minimal revision to the current RADOS mechanism, thus is easier to be implemented correctly. It does it work on the present single object transaction implementation without touching PG LOG and PG metadata data. It could handle all the faulty cases, roll back and forward correctly. The only problem we see is that, as described above, if MASTER, client and SLAVE all undergo an interval change, the persisted transaction may stay at SLAVE for some time until somebody accesses the object, and the object will be locked for a while before SLAVE sending QUERY to MASTER then rolling back the transaction, but that should be ok

doc/dev/osd_internals/multi_object_write_transaction.rst
dragonylffly: Yep, temp object is better option

doc/dev/osd_internals/multi_object_write_transaction.rst
athanatos: Clients don't undergo interval changes, they just die.  The slave can't be continuously sending queries to the master, something would have to trigger it.  It also can't merely do it when it gets an IO on the object -- it would either be redundant or inadequate (what if the master dies after the blocked IO arrives but before it completes the transaction?)  Does the slave monitor osdmap changes to detect an interval change on the primary PG for all operations for which it is the slave?  If so, that seems like it duplicates a great deal of the Objecter.  Does it poll?  If so, at what interval?

doc/dev/osd_internals/multi_object_write_transaction.rst
athanatos: I'm not sure that's a good tradeoff.  Deadlock detection by its nature detects a state which lasts for the duration of a single transaction.  Spending a whole round trip to detect a condition which could easily arise anyway in the mean time doesn't seem worthwhile.  It would make more sense to perform the deadlock detection atomically with the log step.  I think this is particularly important since a user of this mechanism could ensure that there is never a deadlock by construction.

athanatos:
Ok, so the master persists only its own data.  Still, the master needs to be able to immediately identify all in-progress transactions after an interval change to ensure progress and that the slaves will be able to release their locks (either due to success or failure) on time.

My problem with this design is that if the primary, slave, and client all die, the transaction is left in an indeterminate state until an IO happens on the object in question.  That seems bad.  I want it to be the case that as long as both the slave and master pgs are active, the transaction will make progress towards either success or failure without any interaction from clients.  I want this to happen without polling.

I think it suffices for the master pg to have a durable index of in-progress transactions (stored on the pg meta object and mediated by log entries in repops) which it uses on activate to re-establish the on-going transaction state with slaves.  This makes it unnecessary for slaves to query the primary -- when the primary goes active again, it will contact them.  It also means you can structure the primary->slave communications as Objecter linger ops and reuse all of that machinery.

If you want to avoid a write before the master can send the operations to the slaves, then we can complicate the design somewhat by instead adding a similar index on the slave side (updated with the log step repop).  The slaves would monitor the OSDMap for changes for interval changes in the primary and resend the query on such an interval change.

Either way, I really do not want to rely on a client for progress -- as long as both PGs are active, something needs to be ensuring progress.

dragonylffly:
@athanatos @liewegas 
We are implementing the new version according to your comments, before we submit the revised 
code, we wanna make sure we are on the right way, please check the following process,
(1) MASTER recieves the request from client, writes a LOCK entry into PGLOG, creates
a temprorary object, and records the SLAVE information in it;
(2) MASTER sends LOCK requests to SLAVEs, each request includes a group of operations 
on a single slave object;
(3) SLAVE recieves LOCK, does do_osd_ops() to confirm the validity, then writes 
a LOCK entry into PGLOG, creates a temporary object, and records the transaction information in it;
(4) SLAVE sends ack to MASTER;
(5) MASTER collects ack, and does the operations on the master object, besides, writes a COMMIT entry
into PGLOG;
(6) MASTER sends COMMIT to SLAVES;
(7) SLAVE receives COMMIT, performs the operations on the slave object, besides, writes a COMMIT entry into PGLOG;
(8) SLAVE sends ack to MASTER, deletes the temporary object and writes a UNLOCK entry into PGLOG;
(9) MASTER collects ack, deletes the temporary object and writes a UNLOCK entry into PGLOG

athanatos:
@dragonylffly 

Can you elaborate what happens in the failure scenarios I described?

dragonylffly:
MASTER restarts,
reconstructs the transaction in memory according to the PGLOG entry and temporary object,
(1) If the transaction is in LOCK state:

1. MASTER sends UNLOCK to SLAVEs;
2. SLAVE receives UNLOCK, if the transaction does not exist or in UNLOCK state, goto (3); If the transaction is in LOCK state, the SLAVE rolls back by writing a UNLOCK entry into PGLOG, and deletes the temporary object;
3. SLAVE sends ack to MASTER;
4. MASTER collects ack, writes a UNLOCK entry into PGLOG, and deletes the temporary object;

(2) If the transaction is in COMMIT state:

1. MASTER sends COMMIT to SLAVEs;
2. SLAVE receives COMMIT, if the transaction is not in LOCK state, goto (3); Otherwise, does the operations on the slave object, besides, writes a COMMIT entry into PGLOG;
3. SLAVE sends ack to MASTER, deletes the temporary object and writes a UNLOCK entry into PGLOG;
4. MASTER collects ack, deletes the temporary object and writes a UNLOCK entry into PGLOG

SLAVE restart,
constructs the transaction according to the PGLOG entry and temporary object,
(1) If the transaction is in LOCK state:
    it waits MASTER to re/send LOCK/UNLOCK/COMMIT
(2) If the transaction is in COMMIT state:
    it deletes the temporary object and writes a UNLOCK entry to the PGLOG; if MASTER resends COMMIT, it directy sends ack to MASTER;

dragonylffly:
@athanatos @liewegas

athanatos:
@dragonylffly I'll get back to this tomorrow.

athanatos:
@dragonylffly What happens if the slave restarts after the master has sent the message, but before the slave has received it?  What does the primary use the temp object for?

wwformat:
> @dragonylffly What happens if the slave restarts after the master has sent the message, but before the slave has received it? What does the primary use the temp object for?

We use Objecter to send messeage, so if the SLAVE restarts, I think MASTER will resend the LOCK/COMMIT/UNLOCK message in that case.
The temp object in MASTER records all SLAVEs oid information in it, so the MASTER use it to resend COMMIT/UNLOCK message to all SLAVEs when MASTER restarts.

athanatos:
You can record the slave identity (though not the actual transaction) information in the pg log entry, so no need for a temp object there.  I think design is reasonable.

dragonylffly:
@athanatos @liewegas both code and document updated, your comments appreciated

athanatos:
@dragonylffly Can you ping me in 10 days (I'm busy today and tomorrow and on vacation next week)?

dragonylffly:
@athanatos

athanatos:
@dragonylffly I'm sorry, I know I put you off last week, but I won't really have time to look at this for a few weeks.

doc/dev/osd_internals/multi_object_write_transaction.rst
```diff
@@ -0,0 +1,113 @@
...
+     records the transaction information in it, then sends ack to *MASTER*;
+
+  #. *MASTER* collects response, then does the operations on the *master*
+     object, besides, writes a *COMMIT* entry into *PGLOG*. *MASTER* sends
+     *COMMIT* request to *SLAVE*, then sends ack to client;
+
+  #. *SLAVE* receives *COMMIT*, performs the operations on the slave object,
+     besides, writes a *COMMIT* entry into *PGLOG*. *SLAVE* sends ack to
+     *MASTER*, deletes the meta object and writes a *UNLOCK* entry into
+     *PGLOG*;
```
liewegas: I think COMMIT and UNLOCK don't need to be separate events.  A single transaction will atomically write the commit+unlock; then we send the ACK.

If the slave restarts and it's already committed, it does nothing.  If the master didn't get a commit-ack, it can resend commit, and the slave will reply with commit-ack immediately (txn isn't open).  If the master did see the commit-ack, then we are done...

doc/dev/osd_internals/multi_object_write_transaction.rst
```diff
@@ -0,0 +1,113 @@
...
+- SLAVE restart
+
+ - If the transaction is in *LOCK* state
+
+ *Slave* waits *MASTER* to re/send *LOCK/UNLOCK/COMMIT*
+
+ - If the transaction is in *COMMIT* state
+
+ *SLAVE* does the *UNLOCK* process; if *MASTER* resends *COMMIT*, it directy
+ sends ack to *MASTER*
```
liewegas: ...and I think there isn't actually a COMMIT state on the slave.. the COMMIT does the commit and closes out the txn state there.

Am I missing something?

doc/dev/osd_internals/multi_object_write_transaction.rst
```diff
@@ -0,0 +1,113 @@
...
+
+The high level process is,
+
+  1. Define the following class,
+
+     class *CEPH_RADOS_API MultiObjectWriteOperation*
+
+     {
+
+       std::map<std::string, ObjectWriteOperation*> slaves;
```
liewegas: Is the std::string the slave object name?  I think it needs to be more than a std::string in order to accommodate the other info.. object_locator_t + object_t.

liewegas: pair<object_locator_t,object_t> would work?

doc/dev/osd_internals/multi_object_write_transaction.rst
dragonylffly: We could remove the operation of "write unlock entry", then the process is: commit and write commit entry in a transaction; ack sent to master; delete the meta object, so logically remove the UNLOCK state from the state machine. But in the current data structure design, the delete meta object operation could not be merged in the commit step, since one ctx could not touch two objects

doc/dev/osd_internals/multi_object_write_transaction.rst
dragonylffly: As described above

doc/dev/osd_internals/multi_object_write_transaction.rst
dragonylffly: we use string for simplicity, and we could change to the pair

doc/dev/osd_internals/multi_object_write_transaction.rst
```diff
@@ -0,0 +1,113 @@
...
+
+     };
+
+     based on this, client could send a group of ops on multi-objects
+     through *MOSDOp* to the primary osd in charge of the *master* object
+     which is explicitly specified by client, call this osd *MASTER*
+
+  #. *MASTER* recieves the multi-object write operation request, writes a
+     *LOCK* entry into *PGLOG*, creates a meta object, and records the *slaves*
+     information in it;
```
dragonylffly: yes, the list of the object_t slaves

doc/dev/osd_internals/multi_object_write_transaction.rst
liewegas: We can modify as many objects as we want in ObjectStore::Transaction.  The only single-object limit is the pg_log_entry_t, but I think that shouldn't reference the temp meta object, but the master (or slave) object; the meta object will be referenced implicity because of the entry type (COMMIT, etc).

I'll look at the rest of the code...

wwformat: > We can modify as many objects as we want in ObjectStore::Transaction. The only single-object limit is the pg_log_entry_t, but I think that shouldn't reference the temp meta object, but the master (or slave) object; the meta object will be referenced implicity because of the entry type (COMMIT, etc).
> 
> I'll look at the rest of the code...

I think if the primary osd down, the meta object should be peering and recovery to other osd, so I will append pg_log_entry_t::MODIFY to pglog to track meta object, and append pg_log_entry_t::DELETE to pglog to delete meta object.

dragonylffly: we need make peering and recovery know meta object, so that it can be synchronized within the replicas

doc/dev/osd_internals/multi_object_write_transaction.rst
dragonylffly: codes updated with COMMIT and UNLOCK merged, please review the codes

liewegas: @dragonylffly can you update this doc to reflect the updated sequence?

liewegas:
Along with the ceph_test_rados_api_multiobject tests we'll also want to extend RadosModel (test/osd/*) to submit these ops as well for stress testing.

doc/dev/osd_internals/multi_object_write_transaction.rst
```diff
@@ -0,0 +1,113 @@
+==============================
```
amitkumar50: @dragonylffly Please prepend doc before PR title.
`It is required to prefix it with the subsystem or module you are changing`

gregsfortytwo:
Closing this as stale. :)

PR #9404, opened 2016-05-31
hammer: mds: wrongly treat symlink inode as normal file/dir when symlink inode is stale on kcephfs

smithfarm:
http://tracker.ceph.com/issues/16082

### DIFF
--- a/src/mds/Server.cc
+++ b/src/mds/Server.cc
@@ -2747,7 +2747,7 @@ void Server::handle_client_open(MDRequestRef& mdr)
     respond_to_request(mdr, -ENXIO);                 // FIXME what error do we want?
     return;
     }*/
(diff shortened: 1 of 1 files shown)


smithfarm:
@ukernel @gregsfortytwo This PR passed a cephfs integration run with two failures. Both failures are reproducible in both hammer-backports and hammer itself. One failure is due to http://tracker.ceph.com/issues/11258 ("cephfs-java ftruncate unit test failure") which is fixed in master, but the hammer backport is non-trivial. The second failure is probably an issue with some of the tests. For details, see http://tracker.ceph.com/issues/15895#note-21 

gregsfortytwo:
Those test failures don't look like a problem to me.

smithfarm:
@jcsp OK to merge this?

PR #9406, opened 2016-05-31
doc: add breathe_domain_by_extension to conf.py

tchaikov:
so it can render the doxygen generated objects with the proper domain by
checking the extension of rendered source file.

@badone it should work if https://github.com/michaeljones/breathe/pull/263 gets merged.


badone:
LGTM

PR #9408, opened 2016-05-31
rbd: don't allow --allow-shrink when growing image

bniver:
Fixes: http://tracker.ceph.com/issues/15991
Signed-off-by: Brett Niver <bniver@redhat.com>

### DIFF
--- a/src/tools/rbd/action/Resize.cc
+++ b/src/tools/rbd/action/Resize.cc
@@ -75,6 +75,11 @@ int execute(const po::variables_map &vm) {
               << "--allow-shrink flag" << std::endl;
     return -EINVAL;
   }
+  if (info.size < size && vm["allow-shrink"].as<bool>()) {
+    std::cerr << "rbd : growing an image isn't allowed with the "
+	     << "--allow-shrink flag" << std::endl;
+   return -EINVAL;
+  }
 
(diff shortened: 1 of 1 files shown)


idryomov:
IMO this is unnecessary.  This is essentially a force flag and I can't think of any standard command that would error out if force is specified but isn't used.  That goes for our "--yes-i-really[-really]-mean-it" too.

bniver:
@idryomov  - that's fine.   I would assume that if we decide this is extraneous, then the tracker item would be marked "not a bug"?

idryomov:
Redmine's speak for "not a bug" is Rejected.  That's just me though, let's see what @jdurgin and @dillaman think first.

bniver:
Why is it necessary at all?  If new size is smaller, then shrink, right?

dillaman:
@bniver It's required on shrink for the same reason "--yes-i-really-really-mean-it" is there for pool deletions.  If a user accidentally shrinks an image (typo, wrong image, etc), they probably just corrupted the image's file system if they didn't properly shrink everything inside the VM.

jdurgin:
I agree with @idryomov, accepting a force flag like this when it does nothing seems fine, and avoids breaking scripts that would e.g. always pass the option.

bniver:
ok, want me to mark the issue rejected?

jdurgin:
sure, go ahead

PR #9414, opened 2016-05-31
buffer: don't invalidate crc cache of new ptr before zeroing it

branch-predictor:
When doing append_zero, we create new bufferptr and then zero it.
Don't invalidate internal crc cache, because at this point it's
empty anyway.

Signed-off-by: Piotr Dałek <git@predictor.org.pl>


tchaikov:
lgtm.

yuriw:

PR #9415, opened 2016-05-31
packaging: move parted requirement to -osd subpkg

ktdreyer:
Prior to this change, ceph-base required the "parted" package, which meant that any installation of ceph-osd, ceph-mon, or ceph-mds would pull in the parted package.

Move the parted dependency to ceph-osd, since ceph-disk is the only thing that uses parted.

The advantage of this change is that Mon and MDS systems will not need to install the parted package.

Fixes: http://tracker.ceph.com/issues/16095

### DIFF
--- a/ceph.spec.in
+++ b/ceph.spec.in
@@ -205,7 +205,6 @@ Requires:      python-setuptools
 Requires:      grep
 Requires:      xfsprogs
 Requires:      logrotate
-Requires:      parted
 Requires:      util-linux
 Requires:      hdparm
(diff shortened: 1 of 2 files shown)


ktdreyer:
CC @branto1 @smithfarm in case they want to review.

smithfarm:

branto1:
A quick grep on sources reveals that parted is only used in ceph-disk and qa workunit that tests ceph-disk + the gitbuilders look fine -> merging.

PR #9416, opened 2016-06-01
hammer: mon/OSDMonistor: improve reweight_by_utilization() logic

chardan:
http://tracker.ceph.com/issues/15770

Important note: The logic from the lambda function appearing in master was ported into the C++98 friendly function object because hammer does not support C++11.

### DIFF
--- a/src/mon/OSDMonitor.cc
+++ b/src/mon/OSDMonitor.cc
@@ -1494,6 +1494,13 @@ void OSDMonitor::check_failures(utime_t now)
 
 bool OSDMonitor::check_failure(utime_t now, int target_osd, failure_info_t& fi)
 {
+  // already pending failure?
+  if (pending_inc.new_state.count(target_osd) &&
+      pending_inc.new_state[target_osd] & CEPH_OSD_UP) {
+    dout(10) << " already pending failure" << dendl;
+    return true;
+  }
+
   utime_t orig_grace(g_conf->osd_heartbeat_grace, 0);
   utime_t max_failed_since = fi.get_failed_since();
   utime_t failed_for = now - max_failed_since;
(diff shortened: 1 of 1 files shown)


smithfarm:
@chardan Thisooks really good - thanks for helping with this non-trivial backport!

I have just two "nits", both regarding the commit message in a3d5114 

First, please remove the first line so the title is the same as the original commit.

The second "nit" is more of a judgment call - would it be nicer/more useful to expand the conflict resolution sentence to include the "Important note" that you put in the PR description? (Hypothetical future forensic investigations will typically be only looking at the git commits and their descriptions, so all important information should be included.)

smithfarm:
Marking DNM until commit message is fixed and we get a successful make check

smithfarm:
test this please

smithfarm:

smithfarm:
@liewegas @athanatos This PR passed a /200 rados run on Ubuntu. None of the failures were reproducible. For details see http://tracker.ceph.com/issues/15895#note-18

src/mon/OSDMonitor.cc
```diff
@@ -466,7 +466,7 @@ struct Sorter {
 
     double average_util; 
 
-    Sorter(const average_util_)
+    Sorter(const double average_util_)
```
athanatos: This should have been in the previous commit?

athanatos:
There seems to be a problem with the commits.  If you can fix it and verify that the git diff between the two resulting branches is empty, I'm ok with merging it without another round of testing.

chardan:
rebased

smithfarm:
@athanatos We fixed it in `wip-15770-hammer-2` and pushed both branches to ceph/ceph so you can verify:

```
git diff ceph/wip-15770-hammer ceph/wip-15770-hammer-2
```

Shall I merge wip-15770-hammer-2 manually, then?

smithfarm:
@athanatos On second thought, it'll be better to reset this branch to `wip-15770-hammer-2` and force-push - manual merge would mean the PR# doesn't get picked up by the release-notes script.

smithfarm:
@chardan: Please do

```
git checkout wip-15770-hammer
git reset --hard wip-15770-hammer-2
git push -f origin wip-15770-hammer
```

chardan:
Done!

athanatos:
lgtm

PR #9418, opened 2016-06-01
mon/MDSMonitor: fix wrong positive result of jewel flag check

xiexingguo:


xiexingguo:
retest this please

xiexingguo:
@jcsp Some fs-related fixes, please help to review. Thanks.

src/mon/MDSMonitor.cc
```diff
@@ -2385,16 +2390,14 @@ int MDSMonitor::legacy_filesystem_command(
...
       return -EINVAL;
     }
 
     const MDSMap& mdsmap =
       pending_fsmap.filesystems.at(pending_fsmap.legacy_client_fscid)->mds_map;
       
-    if (!mdsmap.allows_multimds() &&
-	maxmds > mdsmap.get_max_mds() &&
-	maxmds > 1) {
+    if (!mdsmap.allows_multimds() && maxmds > 1) {
```
gregsfortytwo: Nope. We need to deal here with upgraded clusters that were set up with multimds prior to the lockout.

src/mon/MDSMonitor.cc
```diff
@@ -2385,16 +2390,14 @@ int MDSMonitor::legacy_filesystem_command(
   if (prefix == "mds set_max_mds") {
     // NOTE: deprecated by "fs set max_mds"
     int64_t maxmds;
-    if (!cmd_getval(g_ceph_context, cmdmap, "maxmds", maxmds) || maxmds < 0) {
+    if (!cmd_getval(g_ceph_context, cmdmap, "maxmds", maxmds) || maxmds <= 0) {
```
gregsfortytwo: I believe you need to be able to set maxmds 0 to do a proper shutdown and cleanup of the FS?

src/mon/MDSMonitor.cc
```diff
@@ -1815,8 +1815,7 @@ class SetHandler : public FileSystemCommandHandler
         return -EINVAL;
       }
 
-      if (!fs->mds_map.allows_multimds() && n > fs->mds_map.get_max_mds() &&
```
jcsp: The problem with this change is in the case where someone has:

1.  setallow_multimds tru
2.  set max_mds to something >1
3.  set allow_multimds false

In this situation, we should allow people to decrease max_mds for their filesystem.

So I think just drop this commit?

jcsp:
This all looks good apart from the "drop weird limit for multimds checking" commit, which I think we should drop.

src/mon/MDSMonitor.cc
jcsp: No, you can just do cluster_down, fail the ranks, and then "fs rm".  I don't think there's any gain to formally deactivating rank 0 if you're about to destroy the FS.

xiexingguo:
@jcsp @gregsfortytwo Thank you both for the quick review. Dropped that change.

jcsp:

jcsp:
Created backport ticket for the bug fix parts of this http://tracker.ceph.com/issues/16136

PR #9421, opened 2016-06-01
mds/cdir: limit max snap id in lookup()

renhwztetecs:
If you give more than Snapid of dentry.last, then you can get dentry
It's not reasonable, I fix it


ukernel:
reviewed-by

jcsp:

PR #9426, opened 2016-06-01
cmake: change libmds back to a static library

mattbenjamin:
Required by ceph-mds.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -974,7 +974,7 @@ if(${WITH_MDS})
     mds/MDLog.cc
     ${CMAKE_SOURCE_DIR}/src/common/TrackedOp.cc
     ${CMAKE_SOURCE_DIR}/src/osdc/Journaler.cc)
-  add_library(mds ${mds_srcs}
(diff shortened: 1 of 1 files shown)


cbodley:
i'm pretty sure we were building libmds as a static library before the change from `ENABLE_SHARED` (where the default was STATIC unless we added `${CEPH_SHARED}` to the `add_library()` call) to `BUILD_SHARED_LIBS` (where the default is SHARED unless we override with STATIC) in 1dbfb264ff1cf32807d5a91e75114fa25d553162. as far as i know, it's only there as an intermediate library that we can link into unit tests and cephfs tools. do you think there are benefits to dynamic linkage here, or should we add STATIC to the add_library(mds)?

cc @tchaikov @alimaredia

mattbenjamin:
I don't care, but this isn't something that should just change accidentally--if that happened, then we're not being very careful

mattbenjamin:
ok,  so now libmds is static once more;  verified;  nothing else checked

mattbenjamin:
(thanks, casey)

PR #9428, opened 2016-06-01
rgw: replace count with find

stiopaa1:

### DIFF
--- a/src/cls/rgw/cls_rgw_types.h
+++ b/src/cls/rgw/cls_rgw_types.h
@@ -733,7 +733,7 @@ struct rgw_usage_log_entry {
   void sum(rgw_usage_data& usage, map<string, bool>& categories) const {
     usage = rgw_usage_data();
     for (map<string, rgw_usage_data>::const_iterator iter = usage_map.begin(); iter != usage_map.end(); ++iter) {
-      if (!categories.size() || categories.count(iter->first)) {
+      if (!categories.size() || categories.find(iter->first) != categories.end()) {
         usage.aggregate(iter->second);
       }
     }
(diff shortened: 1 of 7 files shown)


cbodley:
were any of these using multimap::count()? isn't map::count() equivalent to find() unless you need the iterator?

mattbenjamin:
agree, I don't see the bad count() AND search idiom in any of these;  iiuc, count() is not generally to be avoided, or something

stiopaa1:
@cbodley 
Yes, it is equivalent when it comes to performance for maps, but not multimaps, so I think it is just better to use it everywhere. I also think it reads better than count. 
If none of the above convinces you then I was just doing this task:
https://trello.com/c/YUlqAp4s/105-use-stl-find-instead-of-count
:)

cbodley:
@stiopaa1 cool, i didn't see that on trello. i don't necessarily agree on readability, especially when it adds repetition like `old_info->swift_keys.find(iter->first) != old_info->swift_keys.end()` and leads to extra line breaks

but i'm happy to let @liewegas make the judgement call - should we replace `count()` with `find() != end()` everywhere, or only in cases like the following where it's misused?
```c++
if (foo.count(bar)) {
  cout << foo[bar] << endl;
}
```

liewegas:
Yeah, my concern was in places where it is misused (usually subsequent uses of operator[]).

mattbenjamin:
I think that we should fix the count() + [] (or equivalent) cases, but NOT behave as if count() is inefficient or deprecated in general for map and set.

stiopaa1:
@cbodley @liewegas @mattbenjamin 
Thanks for your input! 
I'm closing this then.

PR #9430, opened 2016-06-01
jewel: backports needed for Manila

jcsp:
I'll hopefully be back in the morning with a link to a test run


jcsp:
Tested here:
http://pulpito.ceph.com/jspray-2016-06-02_15:43:38-fs-wip-manila-backports---basic-mira/

The failures are rgw selinux, osd crashes, boost valgrind false positive, and a PJD failure (which is unknown, but cannot be related to these python changes).

PR #9432, opened 2016-06-01
Clean up hobject_t::max() values in decode, avoid comparing directly to get_max()

athanatos:
Pending tests.  http://tracker.ceph.com/issues/16113


liewegas:
lgtm

athanatos:
Tested sjust@teuthology:/a/samuelj-2016-06-04_19:59:50-rados-wip-16113-distro-basic-smithi

PR #9438, opened 2016-06-02
cmake: rbd_mirror not linking rbd_test_mock

alimaredia:
Added rbd_test_mock lib, also changed name of
journal_test to journal_test_mock to mimic
automake naming.


alimaredia:
@tchaikov @cbodley `make check` won't build without this fix.

PR #9369, opened 2016-05-27
ceph-disk: modify the disk partition sequence of ceph-disk list

houlinfei:
Signed-off-by: houlinfei <hou.linfei@h3c.com>

PR #9442, opened 2016-06-02
ceph-disk: modify the disk partition sequence of ceph-disk list

houlinfei:
before the change, when you execute ceph-disk list, it will show in this order:
    /dev/sda :
     /dev/sda2 ceph journal, for /dev/sda1
     /dev/sda1 ceph data, active, cluster ceph, osd.0, journal /dev/sda2

    after the change, it will show in this order:
    /dev/sda :
     /dev/sda1 ceph data, active, cluster ceph, osd.0, journal /dev/sda2
     /dev/sda2 ceph journal, for /dev/sda1

    Before the change,  the entries are disordered. After the change, the entries will always show in the same order. It makes it look good and easier to visually spot the differences between two runs.
 
Signed-off-by: houlinfei <hou.linfei@h3c.com>

### DIFF
--- a/src/ceph-disk/ceph_disk/main.py
+++ b/src/ceph-disk/ceph_disk/main.py
@@ -3974,7 +3974,7 @@ def list_format_plain(devices):
     for device in devices:
         if device.get('partitions'):
             lines.append('%s :' % device['path'])
-            for p in sorted(device['partitions']):
+            for p in sorted(device['partitions'], key=lambda x: x['path']):
                 lines.append(list_format_dev_plain(dev=p,
                                                    prefix=' '))
         else:


houlinfei:
@liewegas  @dachary @ErwanAliasr1   #9369 reopen here bc some mistake operation.

ErwanAliasr1:
@houlinfei I'm not an expert on that part as @dachary but it would be lovely to explain in your commit message why this change is required.

When some reads a commit it's pretty handy to have an explaination, you explain what you change but not why. What does this commit solves ?

You do explain what you change, how you change but not why (what was your issue before that patch).

dachary:
@houlinfei you could amend the commit by adding something like
<pre>
Sort ceph-disk list output so that entries always show in the same order. It makes it easier to visually spot the differences between two runs.
</pre>

houlinfei:
Hi, i have modified the commit. Thank you for your suggests. @ErwanAliasr1

houlinfei:
Thank you for your idea. I have modified in this way. @dachary

dachary:
@houlinfei thanks that's perfect

PR #9444, opened 2016-06-02
xio: provide dout_prefix for XioConnection

Adirl:
provide dout_prefix for XioConnection with: from>>to, peer-type, and conn/sess
1. this prefix will be added for any log message from XioConnection
   (2. hence, remove info that is now redundeant in few existing log messages)

Signed-off-by: Avner BenHanoch avnerb@mellanox.com

### DIFF
--- a/src/msg/xio/XioConnection.cc
+++ b/src/msg/xio/XioConnection.cc
@@ -77,6 +77,14 @@ void print_ceph_msg(CephContext *cct, const char *tag, Message *m)
   }
 }
 
+#undef dout_prefix
+#define dout_prefix conn_prefix(_dout)
+ostream& XioConnection::conn_prefix(std::ostream *_dout) {
+  return *_dout << "-- " << get_messenger()->get_myinst().addr << " >> " << peer_addr
+                << " peer=" << peer.name.type_str()
+                << " conn=" << conn << " sess=" << session << " ";
+}
+
 XioConnection::XioConnection(XioMessenger *m, XioConnection::type _type,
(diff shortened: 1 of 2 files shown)


src/msg/xio/XioConnection.cc
```diff
@@ -77,6 +77,14 @@ void print_ceph_msg(CephContext *cct, const char *tag, Message *m)
   }
 }
 
+#undef dout_prefix
+#define dout_prefix _conn_prefix(_dout)
+ostream& XioConnection::_conn_prefix(std::ostream *_dout) {
```
mattbenjamin: does this method need to have a leading _ in the name?  I'd like to avoid that, if possible.

avnerbh:
I only copied this format from AsyncMessneger

```
#undef dout_prefix
#define dout_prefix _conn_prefix(_dout)
ostream& AsyncConnection::_conn_prefix(std::ostream *_dout) {
```

i have no problem to remove the leading _ here

Adirl:
@mattbenjamin Anything else needed here? do you us to change leading _ in the name?
@avnerbh 

Thanks

Adirl:
@mattbenjamin @avnerbh 
leading underscore fixed

cbodley:
:+1:

Adirl:
@mattbenjamin  , anything else needed here?

avnerbh:
@mattbenjamin can we progress with this issue?

liewegas:
retest this please

PR #9445, opened 2016-06-02
mds: avoid duplicated call of context; fix potential memory leak

xiexingguo:


xiexingguo:
retest this please

src/mds/MDSContext.h
```diff
@@ -136,10 +138,34 @@ class C_IO_Wrapper : public MDSIOContext
...
+ */
+class MDSIOFinisher: public Context {
+  C_IO_Wrapper *con;
+  Finisher *fin;
+public:
+  MDSIOFinisher(C_IO_Wrapper *c, Finisher *f) : con(c), fin(f) {
+    assert(fin != NULL);
+    assert(con != NULL);
+  }
+  void finish(int r) {
```
jcsp: Wouldn't it be simpler to just change C_IO_Wrapper to have a destructor that deletes the wrapped context?  That way the caller doesn't have to know about the magic ECANCELLED behaviour, he can just delete a context if he never wants to call it back

xiexingguo: @jcsp Done and this idea is really gorgeous. Thanks, John.

src/mds/MDSContext.h
```diff
@@ -135,11 +137,36 @@ class C_IO_Wrapper : public MDSIOContext
   C_IO_Wrapper(MDSRank *mds_, MDSInternalContextBase *wrapped_) : MDSIOContext(mds_), wrapped(wrapped_) {
     assert(wrapped != NULL);
   }
+
+  ~C_IO_Wrapper() {
+    delete wrapped;
+  }
+
   virtual void finish(int r) {
     wrapped->complete(r);
```
jcsp: Since we're now calling delete in the constructor, I think you need to do "wrapped = nullptr;" after calling complete

src/mds/MDSContext.h
```diff
@@ -135,11 +137,36 @@ class C_IO_Wrapper : public MDSIOContext
...
+class MDSIOFinisher: public Context {
+  C_IO_Wrapper *con;
+  Finisher *fin;
+public:
+  MDSIOFinisher(C_IO_Wrapper *c, Finisher *f) : con(c), fin(f) {
+    assert(fin != NULL);
+    assert(con != NULL);
+  }
+  void finish(int r) {
+    if (r == -ECANCELED) {
```
jcsp: Does MDSIOFinisher still need to exist at all?  Couldn't you just do deletion of C_IO_Wrapper now?

xiexingguo: > Does MDSIOFinisher still need to exist at all? Couldn't you just do deletion of C_IO_Wrapper now?

I doubt we can, especially when we simply want to delete the C_IO_Wrapper when we delete the  `C_OnFinisher`. 
( https://github.com/ceph/ceph/pull/9445/files#diff-ea4df8dd62d8f79137676b9f21d58633L1014) .
As you can see, `C_OnFinisher` always passes the context onto the finisher even when it is destroyed, see below:
```
class C_OnFinisher : public Context {
  Context *con;
  Finisher *fin;
public:
  C_OnFinisher(Context *c, Finisher *f) : con(c), fin(f) {
    assert(fin != NULL);
    assert(con != NULL);
  }
  void finish(int r) {
    fin->queue(con, r);
  }
};
```

src/mds/MDSContext.h
xiexingguo: Fixed. Thanks, john.

xiexingguo:
@jcsp  John, I update the code as per your kind reminder, please have a look at this again.

The reason that I introduce a `MDSIOFinisher` is because we want to delete the con simultaneously when we delete the Finisher itself, which the normal C_OnFinisher can't handle properly.

jcsp:
I'm seeing crashes like this:
```
 1: (()+0x4f4852) [0x7f1b8fbe4852]
 2: (()+0x10340) [0x7f1b8f052340]
 3: (C_IO_Wrapper::~C_IO_Wrapper()+0x1e) [0x7f1b8f92dace]
 4: (MDSIOContextBase::complete(int)+0x9a) [0x7f1b8fb488ba]
 5: (Finisher::finisher_thread_entry()+0x206) [0x7f1b8fc11b86]
 6: (()+0x8182) [0x7f1b8f04a182]
 7: (clone()+0x6d) [0x7f1b8d5a147d]
```

src/mds/MDSContext.h
```diff
@@ -135,11 +137,39 @@ class C_IO_Wrapper : public MDSIOContext
...
     assert(wrapped != NULL);
   }
+
+  ~C_IO_Wrapper() {
+    delete wrapped;
+    wrapped = nullptr;
+  }
+
   virtual void finish(int r) {
     wrapped->complete(r);
```
jcsp: You need to set wrapped to null after calling complete().  Contexts either delete themselves in complete(), or we shouldn't be deleting them at all.

src/mds/MDSContext.h
```diff
@@ -135,11 +137,39 @@ class C_IO_Wrapper : public MDSIOContext
...
+    assert(con != NULL);
+  }
+
+  ~MDSIOFinisher() {
+    delete con;
+    con = nullptr;
+  }
+
+  void finish(int r) {
+    fin->queue(con, r);
```
jcsp: Same here, you must set con to null after enqueuing it

src/mds/MDSContext.h
xiexingguo: Right, thanks!

src/mds/MDSContext.h
```diff
@@ -135,11 +137,39 @@ class C_IO_Wrapper : public MDSIOContext
...
   virtual void finish(int r) {
     wrapped->complete(r);
   }
 };
 
+/**
+ * Like C_OnFinisher but cancellable, in which case we do
+ * cleanup manually instead of handling it to the finisher.
+ */
+class MDSIOFinisher: public Context {
```
jcsp: I don't think it's necessary to define a whole separate class: you can just add the deletion behavior to C_OnFinisher

jcsp:
@xiexingguo please see comments inline

xiexingguo:
Thanks, john. I'll fix it right now.

xiexingguo:
@jcsp Upgraded. Please help to confirm again. Thanks!

jcsp:

PR #9446, opened 2016-06-02
cmake changes

tchaikov:


tchaikov:
```
dh_install: Cannot find (any matches for) "usr/lib/librados.la" (tried in "." and "debian/tmp")
dh_install: librados-dev missing files: usr/lib/librados.la
dh_install: Cannot find (any matches for) "usr/lib/libradosstriper.la" (tried in "." and "debian/tmp")
dh_install: libradosstriper-dev missing files: usr/lib/libradosstriper.la
dh_install: Cannot find (any matches for) "usr/lib/librbd.la" (tried in "." and "debian/tmp")
dh_install: librbd-dev missing files: usr/lib/librbd.la
dh_install: Cannot find (any matches for) "usr/lib/libcephfs.la" (tried in "." and "debian/tmp")
dh_install: libcephfs-dev missing files: usr/lib/libcephfs.la
dh_install: Cannot find (any matches for) "usr/lib/librgw.la" (tried in "." and "debian/tmp")
dh_install: librgw-dev missing files: usr/lib/librgw.la
dh_install: Cannot find (any matches for) "usr/bin/ceph_test_*" (tried in "." and "debian/tmp")
dh_install: ceph-test missing files: usr/bin/ceph_test_*
dh_install: Cannot find (any matches for) "usr/bin/librgw_file*" (tried in "." and "debian/tmp")
dh_install: ceph-test missing files: usr/bin/librgw_file*
dh_install: missing files, aborting
```

- [x] debian/rules using cmake. ready, but we need to address following issues first.
- [ ] remove .la files from deb packaging, rpm packaging does not include it as suggested by rpm packaging guide line. as .la files are generated by libtools. shall we ship .pc file? or just leave it to our user to figure out include paths and libraries? 
- [ ] remove some (maybe all) of ceph_test_* executables from packaging. some of them are apparently unit tests. @mattbenjamin do we need to ship them to end user?
- [ ] maybe should remove librgw_file* also. @mattbenjamin and how about these executables?

src/CMakeLists.txt
```diff
@@ -1146,7 +1146,7 @@ endif(BUILD_SHARED_LIBS)
   target_link_libraries(mount.ceph keyutils)
 
   install(TARGETS ceph-syn DESTINATION bin)
-  install(TARGETS mount.ceph DESTINATION /sbin)
+  install(TARGETS mount.ceph DESTINATION ${CMAKE_INSTALL_SBINDIR})
```
tchaikov: @cbodley this addresses the issue you mentioned.

tchaikov:
@alimaredia and @cbodley mind taking a look? we are close now, i think.

mattbenjamin:
@tchaikov I'd like to understand why we need to write a wrapper function to manage libraries when cmake has stable syntax with well understood behavior.  In the last change, a library which only was used statically accidentally turned into a shared object that ceph-mds depended on dynamically;  I think we should avoid making a change that makes library management more magical, and/or more sloppy.

tchaikov:
> I'd like to understand why we need to write a wrapper function to manage libraries when cmake has stable syntax with well understood behavior.

are you talking about https://github.com/ceph/ceph/pull/9446/commits/b7cad2bf82455dc5a6962628eb0ae3bb8b8dee6e#diff-993e65bef9e9430f55305bdf87c0a2edR6? we can repeat the pattern defined by the functions for 4 times, but i agree it is more straightforward. i am okay if we want go this way.

> In the last change, a library which only was used statically accidentally turned into a shared object that ceph-mds depended on dynamically;

i am sorry for that. actually, we relies on the default value of "BUILD_SHARED_LIBS". if user pass `-D BUILD_SHARED_LIBS=TRUE", we are screwed. in my change i made this explicit, and change the intermediate libraries static, but failed to change the one you mentioned. so i think i am heading to the direction: explicit and less magical.

mattbenjamin:
I like your distutils changes a lot, and I see part of the pattern in this, but I would feel most comfortable if we can make use of the standard cmake interfaces to describe library targets, if at all possible.

alimaredia:
@tchaikov I'm with @mattbenjamin on keeping the use of the standard cmake interfaces. Along with what has already been said I feel like the functions add another layer of complexity that would make it a little harder for developers with not as much of a CMake background as us to understand what's going on in those 4 libraries and from editing them.

tchaikov:
okay. i was trying to follow http://pad.ceph.com/p/CMake_To_Do

>     -instead of declarative style, use functions loops and marcos when possible

maybe i misread it..

anyway, will remove `ceph_add_library()`, `ceph_link_libraries()` and `ceph_install_library()`, and unwind them at their caller sites.

tchaikov:
@mattbenjamin and @alimaredia could you take a look again? i believe it's much better now.

cbodley:
@tchaikov can you expand on why we both static and shared versions of these libraries?

tchaikov:
@cbodley because we ship both of them in the *-dev debian packages. if we
decide to not do so and to break the backward compatibility. it would be a
lot easier.

Casey Bodley <notifications@github.com>于2016年6月3日 周五01:31写道：

mattbenjamin:
@cbodley, @tchaikov  it would be great if we could identify .sos we're not really using, and make them vanish :)

tchaikov:
@mattbenjamin 
```
$ find . -name *.so
./lib/libec_jerasure.so
./lib/libcls_kvs.so
./lib/libosd_tp.so
./lib/libos_tp.so
./lib/librbd_tp.so
./lib/librados_tp.so
./lib/libcls_version.so
./lib/libcls_user.so
./lib/libcls_log.so
./lib/libcls_hello.so
./lib/libcls_lock.so
./lib/libcls_timeindex.so
./lib/libcls_statelog.so
./lib/libcls_rbd.so
./lib/libcls_cephfs.so
./lib/libcls_journal.so
./lib/libcls_refcount.so
./lib/libcls_replica_log.so
./lib/libec_jerasure_generic.so
./lib/libec_shec.so
./lib/libgmock.so
./lib/libgtest.so
./lib/libgmock_main.so
./lib/libec_hangs.so
./lib/libec_test_shec_sse3.so
./lib/libec_missing_version.so
./lib/libec_fail_to_initialize.so
./lib/libcls_rgw.so
./lib/libec_fail_to_register.so
./lib/libec_test_jerasure_sse3.so
./lib/libec_test_shec_sse4.so
./lib/libec_test_jerasure_neon.so
./lib/libec_test_shec_generic.so
./lib/libec_test_jerasure_sse4.so
./lib/libec_test_jerasure_generic.so
./lib/libec_test_shec_neon.so
./lib/libec_missing_entry_point.so
./lib/libceph_example.so
./lib/libec_example.so
./lib/libceph_snappy.so
./lib/libceph_zlib.so
./lib/libec_jerasure_sse3.so
./lib/libec_jerasure_sse4.so
./lib/libec_isa.so
./lib/libec_lrc.so
./lib/libec_shec_sse4.so
./lib/libec_shec_generic.so
./lib/libec_shec_sse3.so
./lib/libgtest.so
./lib/libgmock_main.so
./lib/libgmock.so
./lib/librados.so
./lib/libradosstriper.so
./lib/librbd.so
./lib/cython_modules/lib.linux-x86_64-2.7/rados.so
./lib/cython_modules/lib.linux-x86_64-2.7/cephfs.so
./lib/cython_modules/lib.linux-x86_64-2.7/rbd.so
./lib/libcephfs.so
./lib/libcephfs_jni.so
./lib/librgw.so
```

they are 

- tracepoint preload libraries: *_tp
- rados class plugins: libcls_*
- erasure plugins: libec*
- compressor plugins: libceph_* (could use less generic names, though)
- user facing libraries: lib{rbd, rados, radosstriper, cephfs,cephfs_jni,rgw}
- python bindings modules: rados, cephfs, rbd
- gtest libraries: libgtest_main.so, libgmock*

i will change the gtest libraries back into .a, as it actually crashes the unit tests when they exit.

tchaikov:
changelog:

- cmake: generate both static and shared libraries: enable `BUILD_STATIC_LIBS` if `BUILD_SHARED_LIBS` is `OFF`

added following commits
- cmake: compile more radosstriper tests
- cmake: install binaries used by ceph-qa-suite

tchaikov:
@cbodley and @alimaredia i reverted 1dbfb26 and ac47440 in this pull request. is this changeset good to merge now?

PR #9450, opened 2016-06-02
mon: cancel timer events when peon receives paxos update

wonzhq:
In a multi-mon cluster, when a paxos update happens right before the
lease renew, all the events in the mon timer are cancelled, including
the lease renew event, until the paxos update completes. The leader
won't send lease renew to the peons during this period. If for some
reasons a peon is laggy, it is possible that the other peons might
see lease timeout, and initiate a election. From a leader's perspective,
this may not be what it wants. Canceling the lease timeout event on
the peon when it receives the paxos update could solve this issue.

Signed-off-by: Zhiqiang Wang zhiqiang@xsky.com

### DIFF
--- a/src/mon/Paxos.cc
+++ b/src/mon/Paxos.cc
@@ -698,6 +698,7 @@ void Paxos::handle_begin(MonOpRequestRef op)
   // set state.
   state = STATE_UPDATING;
   lease_expire = utime_t();  // cancel lease
+  cancel_events(); // cancel lease timeout event
 
   // yes.
   version_t v = last_committed+1;


tchaikov:
@wonzhq 

i'd take it as a configuration issue, where a proposal fails to be accepted and committed before a peon's lease expires. in other words, `mon_accept_timeout_factor` (2 by default) should be smaller than `mon_lease_renew_interval_factor` (5 by default), am i right?

so maybe you can update `Monitor::sanitize_options()` to enforce this policy instead?

wonzhq:
@tchaikov the default `mon_lease_renew_interval_factor` is 0.6. And I think these 2 factors are independent. The idea here is that the mon leader has canceled the lease renew timer event during the paxos propose. The peon should cancel the lease timeout as well.

tchaikov:
@wonzhq i understand the problem you want to address here. what i don't like is the the window after the peon resets the timer after receiving the begin message, and before it gets the commit message.

what if the leader crashes after it sends out the `OP_BEGIN`, the peon won't bootstrap itself after timeout then with your change.

wonzhq:
@tchaikov yep, we can't simply cancel the timer on the peon. Maybe what we could do is to extend the lease on the peon when it receives the begin message? This means that we treat the begin message as "a sort of lease message".

tchaikov:
> Maybe what we could do is to extend the lease on the peon when it receives the begin message?

we can make a one shot exception in this case: extend the lease for `mon_accept_timeout_factor * g_conf->mon_lease`. does it make sense to you? @jecluis what do you think?

wonzhq:
@jecluis any comments?

dachary:

jecluis:
personally, i think the only way to handle this properly would be to keep updating leases until a new version is available. I don't think canceling the timeouts on begin will work, especially in the case that @tchaikov mentioned - if the leader dies before sending a commit.

Increasing the lease timeout may be a hack we can live with though. At least we keep a timeout around that will keep things sane, even if it takes longer to figure out a mon has died.

liewegas:
I think simply extending the lease period or treating begin as an implicit lease make the most sense...

PR #9452, opened 2016-06-02
osd:preserve allocation hint attribute during recovery

mslovy:
Allocation hint is set into filesystem fsxattr, like XFS.  Or directly assign to onode_t in Blustore and Kstore.  However, it will be lost after recovery or backfill.

I think it does matter for XFS but not BlueStore or KStore since allocation hint can be overwrite in BlueStore or KStore?

### DIFF
--- a/src/osd/osd_types.cc
+++ b/src/osd/osd_types.cc
@@ -4701,7 +4701,7 @@ void object_info_t::encode(bufferlist& bl) const
        ++i) {
     old_watchers.insert(make_pair(i->first.second, i->second));
   }
-  ENCODE_START(15, 8, bl);
+  ENCODE_START(16, 8, bl);
   ::encode(soid, bl);
   ::encode(myoloc, bl);	//Retained for compatibility
   ::encode((__u32)0, bl); // was category, no longer used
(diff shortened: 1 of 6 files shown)


liewegas:
I think it will be cleaner to store this hint information persistently inside object_info_t instead...

mslovy:
Ah, yeah, it seems to be less complicated if we store hint information inside object_info_t. we need one more step to retrieve object_info_t in submit_push_data and reset to the object. I will update a new version later.

mslovy:
@liewegas , update based on the suggestion, please review again

src/osd/osd_types.cc
```diff
@@ -4867,6 +4882,9 @@ ostream& operator<<(ostream& out, const object_info_t& oi)
     out << " dd " << std::hex << oi.data_digest << std::dec;
   if (oi.is_omap_digest())
     out << " od " << std::hex << oi.omap_digest << std::dec;
+  out << " hint [" << oi.expected_object_size
```
liewegas: s/hint /alloc_hint/

liewegas:
@mslovy I have a operator<< formatting nit, but otherwise this looks perfect!

mslovy:
replace hint with alloc_hint as mentioned

liewegas:
@mslovy thanks!

XinzeChi:
@mslovy, if i remember correctly, the object_info_t could store in on chunk in fs xattr (lastest hammer).
If you insert the alloc hint flag in it, could the object_info_t still store in one chunk?

liewegas:
It already spills into the second 254 byte chunk in jewel due to other 
fields.  This won't make a significant impact.  We had to kludge 
local_mtime to fit into one xattr in hammer.

mslovy:
@XinzeChi ,  yeah, I have already checked it before. It won't be a significant impact in Jewel

athanatos:
This passed tests in /a/yuriw-2016-06-22_17:09:03-rados-wip-yuri-testing_2016_6_22-distro-basic-smithi

Are we sure it's harmless to set expected write_size and expected_object_size to 0?

liewegas:
I think we need one more patch to FileStore.cc that makes FileStore::_set_alloc_hint() a no-op if the passed values are 0.  I'm not sure what the ioctl does if passed 0.  Probably nothing, but let's not find out.

athanatos:
Ok, looks like we should change FileStore to make 0 values noops and update the comment on ObjectStore to explain that 0 *must* be a noop for all implementations.

mslovy:
expected_object_size, currently, is not used in Filestore. Not sure it will be used in future. Should we consider also expected_object_size == 0 as noops like the lastest commit?

athanatos:
lgtm, we should retest before merge though

tchaikov:

PR #8528, opened 2016-04-11
rgw: Set Access-Control-Allow-Origin to a Asterisk if allowed in a rule

wido:
http://tracker.ceph.com/issues/15839

Before this patch the RGW would respond with the Origin send by the client in the request
if a wildcard/asterisk was specified as a valid Origin.

This patch makes sure we respond with a header like this:

  Access-Control-Allow-Origin: *

This way a resource can be used on different Origins by the same browser and that browser
will use the content as the asterisk.

We also keep in mind that when Authorization is send by the client different rules apply.
In the case of Authorization we may not respond with an Asterisk, but we do have to
add the Vary header with 'Origin' as a value to let the browser know that for different
Origins it has to perform a new request.

More information: https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS

Fixes: #15348

Signed-off-by: Wido den Hollander <wido@42on.com>

Conflicts:
	src/rgw/rgw_rest.cc

This is the backport fix for #8441

PR #9453, opened 2016-06-02
jewel: rgw: Set Access-Control-Allow-Origin to a Asterisk if allowed in a rule

wido:
http://tracker.ceph.com/issues/16112

Before this patch the RGW would respond with the Origin send by the client in the request
if a wildcard/asterisk was specified as a valid Origin.

This patch makes sure we respond with a header like this:

  Access-Control-Allow-Origin: *

This way a resource can be used on different Origins by the same browser and that browser
will use the content as the asterisk.

We also keep in mind that when Authorization is send by the client different rules apply.
In the case of Authorization we may not respond with an Asterisk, but we do have to
add the Vary header with 'Origin' as a value to let the browser know that for different
Origins it has to perform a new request.

More information: https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS

Fixes: #15348

### DIFF
--- a/src/rgw/rgw_cors.cc
+++ b/src/rgw/rgw_cors.cc
@@ -116,6 +116,13 @@ static bool is_string_in_set(set<string>& s, string h) {
   return false;
 }
 
+bool RGWCORSRule::has_wildcard_origin() {
+  if (allowed_origins.find("*") != allowed_origins.end())
+    return true;
+
+  return false;
+}
+
 bool RGWCORSRule::is_origin_present(const char *o) {
   string origin = o;
   return is_string_in_set(allowed_origins, origin);
--- a/src/rgw/rgw_cors.h
+++ b/src/rgw/rgw_cors.h
@@ -80,6 +80,7 @@ public:
     ::decode(exposable_hdrs, bl);
     DECODE_FINISH(bl);
   }
+  bool has_wildcard_origin();
   bool is_origin_present(const char *o);
   void format_exp_headers(std::string& s);
   void erase_origin_if_present(std::string& origin, bool *rule_empty);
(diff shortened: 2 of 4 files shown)


wido:
This is the same as ##8528, but as a Jewel backport.

Tracker issue: http://tracker.ceph.com/issues/16112

smithfarm:
Changelog:
* added tracker URL to PR description

dachary:
It passed the rgw suite http://tracker.ceph.com/issues/16344#note-6 (except for one failure which I believe to be unrelated).

dachary:
@wido it took a while but the tests finally made it ;-) Would you mind **git cherry-pick -x** instead of manually adding the reference to the cherry-picked commit ? For the sake of consistency: it will help scripts that rely on the specific format that **git cherry-pick -x**  uses. Other than this it's ready to be merged.

dachary:
@wido did you update the commit message ? Or maybe you did and forgot to force-push ?

wido:
@dachary Yes, my bad. I wanted to push asap, but I have a build running locally... Need to wait for that before I can switch branches.

Code is waiting on my desktop, will be there within 3 hours :)

wido:
Done @ldachary !

dachary:
Hey, jenkins, test this please ! (unrelated bluefs mount crash)

wido:
Just as with #8528, is this one ready to merge so it makes it into the next Jewel release?

dachary:
@oritwas do you think it is ready to be merged ? It passed the rgw suite http://tracker.ceph.com/issues/16344#note-6 (except for one failure which I believe to be unrelated).

PR #9454, opened 2016-06-02
os/bluestore: no 'head_read' when check ondisk length.

tanghaodong25:
If we add 'b_off', we should not add 'head_read' again.

### DIFF
--- a/src/os/bluestore/BlueStore.cc
+++ b/src/os/bluestore/BlueStore.cc
@@ -5564,7 +5564,7 @@ void BlueStore::_do_write_small(
     uint64_t tail_read =
       ROUND_UP_TO(b_off + b_len, chunk_size) - (b_off + b_len);
     if ((head_read || tail_read) &&
-	b->get_ondisk_length() >= b_off + b_len + head_read + tail_read) {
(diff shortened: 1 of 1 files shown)


src/os/bluestore/BlueStore.cc
```diff
@@ -5564,7 +5564,8 @@ void BlueStore::_do_write_small(
     uint64_t tail_read =
       ROUND_UP_TO(b_off + b_len, chunk_size) - (b_off + b_len);
     if ((head_read || tail_read) &&
-	b->get_ondisk_length() >= b_off + b_len + head_read + tail_read) {
+	(b->get_ondisk_length() >= b_off + b_len + tail_read) &&
+       b->is_allocated(b_off - head_read, b_len + head_read + tail_read)) {
```
liewegas: I don't think we want the is_allocated() check here unless we have a separate path to fill in zeros.  The read call will give us back zeros if it's not allocated so I don't think it's worth getting tricky here... let's keep it simple and just change the get_ondisk_legnth comparison bug!

tanghaodong25:
@liewegas  Fixed, thanks.

PR #9235, opened 2016-05-20
BackoffThrottle fixes

athanatos:
Also addresses the issue in https://github.com/ceph/ceph/pull/9111

PR #9455, opened 2016-06-02
tests: BackoffThrottle deadlock

dachary:
The stop variable may be changed when wait() exits.  When there are more
getters than putters, one of them may c.wait() and never come back
because all putters have been shut down: add a notify_all() to avoid
that situation.

Fixes http://tracker.ceph.com/issues/16120

### DIFF
--- a/src/test/common/Throttle.cc
+++ b/src/test/common/Throttle.cc
@@ -319,23 +319,24 @@ std::pair<double, std::chrono::duration<double> > test_backoff(
   auto putter = [&]() {
     std::unique_lock<std::mutex> g(l);
     while (!stop) {
-      while (in_queue.empty())
+      if (in_queue.empty()) {
 	c.wait(g);
+      } else {
+	uint64_t c = in_queue.front();
 
-      uint64_t c = in_queue.front();
+	total_observed_total += total;
+	total_observations++;
+	in_queue.pop_front();
+	assert(total <= max);
 
-      total_observed_total += total;
-      total_observations++;
-      in_queue.pop_front();
-      assert(total <= max);
+	g.unlock();
+	std::this_thread::sleep_for(
+	  c * std::chrono::duration<double>(put_delay_per_count*putters));
+	g.lock();
 
-      g.unlock();
-      std::this_thread::sleep_for(
-	c * std::chrono::duration<double>(put_delay_per_count*putters));
-      g.lock();
-
-      total -= c;
-      throttle.put(c);
+	total -= c;
+	throttle.put(c);
+      }
     }
   };
 
@@ -352,6 +353,7 @@ std::pair<double, std::chrono::duration<double> > test_backoff(
   }
   for (auto &&i: gts) i.join();
   gts.clear();
+  c.notify_all();
   for (auto &&i: pts) i.join();
   pts.clear();
 


dachary:
./unittest_throttle it has run in a loop a few hundred times with no hang. Without this patch, it takes about 100 runs to get one hang for sure, as long as the machine has a high load average.

dachary:
it failed with a hang, just now. Either the fix is not really a fix or there is another hang scenario.

dachary:
<pre>
ulimit -c unlimited ; while : ; do echo -n . ; timeout -s SEGV 300 ./unittest_throttle > /dev/null 2>&1 || break ; done
</pre>

wjwithagen:
On 2-6-2016 15:38, Loic Dachary wrote:
> it failed with a hang, just now. Either the fix is not really a fix or
> there is another hang scenario.

'mmmm

Starting to get a deja vue for this.

I had a similar problem, and after a serious discussion we installed a
fix in common/Throttle.cc.  #9235 and #9111
And the hang would be in the while loop in BackoffThrottle::get, not
sure if that would be your case as well.

I asked what would happen if the while-loop now would hit the break, due
to time expiration, and not because of a queue change during the wait.
Jdurgin commented that this would be oke.

Which worked for me on several thousand runs. I think I left it running
overnight, while building/runing/testing other stuff.
But it is a hard nut to crack.

wjwithagen:
On 2-6-2016 15:51, Loic Dachary wrote:
> ulimit -c unlimited ; while : ; do echo -n . ; timeout -s SEGV 300 ./unittest_throttle > /dev/null 2>&1 || break ; done

The throttle tests are hard terminated after 3* 5 sec.
So I guess 60 secs should be more than enough.

dachary:
> Which worked for me on several thousand runs.

Varying the load of the machine will most certainly give you a crash. The window of opportunity is small and if the timing of the test does not vary much the chances of hitting it are very low. Since it happens semi frequently in VMs I guess having a slow machine helps repeat it. Or making it artificially slow.

wjwithagen:
On 2-6-2016 16:12, Loic Dachary wrote:
>     Which worked for me on several thousand runs.
> 
> Varying the load of the machine will most certainly give you a crash.

So timeout actually kicks in, and kills the child process?

The problem is that you would like to catch exactly this process and
attach gdb. And then try to see what is going on. But I haven't really
managed that.
Do you get a usable core when sending a SEGV under Linux?

> The window of opportunity is small and if the timing of the test does
> not vary much the chances of hitting it are very low. Since it happens
> semi frequently in VMs I guess having a slow machine helps repeat it. Or
> making it artificially slow.

I didn't really have a beefy machine to compile and test on.
Now I have a 8-core 4Ghz, 32 Gb ram toy box, that works better.
The test does not do any I/O, and does not really use any memory.
Triggering will require a serious other dummy CPU load.
Maybe cpuburn can help....

I'll also start it in a Centos-7 VM, but those servers are rather
underloaded.

dachary:
> Do you get a usable core when sending a SEGV under Linux?

That's what I'm expected. Why would the core not be useable ?

wjwithagen:
Will cpuburn is good enough to "overload" the system. But now the test gets stuck in:

```
[----------] 3 tests from BackoffThrottle
[ RUN      ] BackoffThrottle.undersaturated
*** Caught signal (Segmentation fault) **
 in thread 803815000 thread_name:
 ceph version Development (no_version)
 1: 0x7c5fe5 <_ZN4ceph9BackTraceC2Ei+0x35> at /usr/srcs/Ceph/work/ceph/src/unittest_throttle
 2: 0x7c4dac <_ZL19handle_fatal_signali+0xcc> at /usr/srcs/Ceph/work/ceph/src/unittest_throttle
 3: 0x801352969 <pthread_sigmask+0x719> at /lib/libthr.so.3
 4: 0x801351d07 <pthread_getspecific+0x1857> at /lib/libthr.so.3
```
Which is wierd in itself, since the the 3 BackoffThrottle tests are "wrapped" in a 5 sec timeout.
```      while (!stop) {
// put in or get from the queu
}

std::this_thread::sleep_for(std::chrono::duration<double>(5));
  {
    std::unique_lock<std::mutex> g(l);
    stop = true;
  }
```
And that is not powerfull enough to terminate. Because either the putter or the getter does not get back into the main loop and exit the 'while(!stop)'.

So my first guess is that since this is with undersaturation, whilest I had the problem with OVERsaturation, that there might be a "similar" issue going on when the queue actually runs emply while the putter is running?

I ended up splashing huge amounts of printf in the common/Throttle.cc code to see where it gets stuck, and then I concluded that 'wait_until()' with a past time was sort of strange. And did not work on FreeBSD. And the code got changed to what it is now.

wjwithagen:
@dachary 
Because the .core contains 8 at least threads in this case, and it is rather tedious to find out what is going on. (Which you can also attribute to my lack of skills. 8) )

dachary:
>  it is rather tedious to find out what is going on. 

I find that http://tracker.ceph.com/issues/16120#note-4 is quite readable actually.

dachary:
> But now the test gets stuck in:

Is this with the current master code ? Or with this pull request ? If you get this crash with this pull request, I'm very interested in the gdb stack traces.

dachary:
<pre>
Thread 5 (Thread 0x7f5715ffb700 (LWP 28199)):
#0  0x00007f57291196d5 in pthread_cond_wait@@GLIBC_2.3.2 () from /lib64/libpthread.so.0
#1  0x00007f57288909ec in std::condition_variable::wait(std::unique_lock<std::mutex>&) () from /lib64/libstdc++.so.6
#2  0x00007f572a04024f in BackoffThrottle::get (this=0x7fff89df19f0, c=c@entry=3) at common/Throttle.cc:381
#3  0x00007f5729fe3e81 in __lambda0::operator() (__closure=0x7f5733796050) at test/common/Throttle.cc:308
#4  0x00007f5728894220 in ?? () from /lib64/libstdc++.so.6
#5  0x00007f5729115dc5 in start_thread () from /lib64/libpthread.so.0
#6  0x00007f5727ffbced in clone () from /lib64/libc.so.6

Thread 4 (Thread 0x7f5725533700 (LWP 28187)):
#0  0x00007f57291196d5 in pthread_cond_wait@@GLIBC_2.3.2 () from /lib64/libpthread.so.0
#1  0x00007f572a03d0f4 in Wait (mutex=..., this=0x7f571c0008c0) at ./common/Cond.h:56
#2  Throttle::_wait (this=this@entry=0x7f5733796020, c=c@entry=7) at common/Throttle.cc:107
#3  0x00007f572a03dd57 in Throttle::get (this=0x7f5733796020, c=7, m=m@entry=0) at common/Throttle.cc:177
#4  0x00007f5729fe96cd in ThrottleTest::Thread_get::entry (this=0x7f5733795db0) at test/common/Throttle.cc:56
#5  0x00007f5729115dc5 in start_thread () from /lib64/libpthread.so.0
#6  0x00007f5727ffbced in clone () from /lib64/libc.so.6

Thread 3 (Thread 0x7f5726535700 (LWP 28166)):
#0  0x00007f5729119a82 in pthread_cond_timedwait@@GLIBC_2.3.2 () from /lib64/libpthread.so.0
#1  0x00007f572a05421a in WaitUntil (when=..., mutex=..., this=0x7f5733794e10) at ./common/Cond.h:72
#2  WaitInterval (interval=..., mutex=..., cct=<optimized out>, this=0x7f5733794e10) at ./common/Cond.h:81
#3  CephContextServiceThread::entry (this=0x7f5733794d90) at common/ceph_context.cc:99
#4  0x00007f5729115dc5 in start_thread () from /lib64/libpthread.so.0
#5  0x00007f5727ffbced in clone () from /lib64/libc.so.6

Thread 2 (Thread 0x7f57274a9700 (LWP 28165)):
#0  0x00007f57291196d5 in pthread_cond_wait@@GLIBC_2.3.2 () from /lib64/libpthread.so.0
#1  0x00007f572a03a2cb in ceph::log::Log::entry (this=0x7f5733745080) at log/Log.cc:459
#2  0x00007f5729115dc5 in start_thread () from /lib64/libpthread.so.0
#3  0x00007f5727ffbced in clone () from /lib64/libc.so.6

Thread 1 (Thread 0x7f5729eb4d00 (LWP 28164)):
#0  0x00007f572911cfcb in raise () from /lib64/libpthread.so.0
#1  0x00007f572a021585 in reraise_fatal (signum=11) at global/signal_handler.cc:69
#2  handle_fatal_signal (signum=11) at global/signal_handler.cc:131
#3  <signal handler called>
#4  0x00007f5729116ef7 in pthread_join () from /lib64/libpthread.so.0
#5  0x00007f5728893fe7 in std::thread::join() () from /lib64/libstdc++.so.6
#6  0x00007f5729fe3518 in test_backoff (low_threshhold=low_threshhold@entry=0.40000000000000002, high_threshhold=high_threshhold@entry=0.59999999999999998, expected_throughput=expected_throughput@entry=1000, high_multiple=high_multiple@entry=2, max_multiple=max_multiple@entry=10, max=max@entry=100, put_delay_per_count=put_delay_per_count@entry=0.001, getters=getters@entry=7, putters=putters@entry=2) at test/common/Throttle.cc:354
#7  0x00007f5729fe4a7d in BackoffThrottle_balanced_Test::TestBody (this=<optimized out>) at test/common/Throttle.cc:394
#8  0x00007f572a012252 in testing::internal::HandleSehExceptionsInMethodIfSupported<testing::Test, void> (object=0x7f5733748a80, method=&virtual testing::Test::TestBody(), location=0x7f572a0fa2bb "the test body") at ./src/gtest.cc:2078
#9  0x00007f572a00d34c in testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void> (object=0x7f5733748a80, method=&virtual testing::Test::TestBody(), location=0x7f572a0fa2bb "the test body") at ./src/gtest.cc:2114
#10 0x00007f5729ff3e3f in testing::Test::Run (this=0x7f5733748a80) at ./src/gtest.cc:2151
#11 0x00007f5729ff4638 in testing::TestInfo::Run (this=0x7f5733740030) at ./src/gtest.cc:2326
#12 0x00007f5729ff4cfc in testing::TestCase::Run (this=0x7f573373fe80) at ./src/gtest.cc:2444
#13 0x00007f5729ffb7c4 in testing::internal::UnitTestImpl::RunAllTests (this=0x7f573373f210) at ./src/gtest.cc:4315
#14 0x00007f572a013638 in testing::internal::HandleSehExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool> (object=0x7f573373f210, method=(bool (testing::internal::UnitTestImpl::*)(testing::internal::UnitTestImpl * const)) 0x7f5729ffb52c <testing::internal::UnitTestImpl::RunAllTests()>, location=0x7f572a0fab10 "auxiliary test code (environments or event listeners)") at ./src/gtest.cc:2078
#15 0x00007f572a00e1ae in testing::internal::HandleExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool> (object=0x7f573373f210, method=(bool (testing::internal::UnitTestImpl::*)(testing::internal::UnitTestImpl * const)) 0x7f5729ffb52c <testing::internal::UnitTestImpl::RunAllTests()>, location=0x7f572a0fab10 "auxiliary test code (environments or event listeners)") at ./src/gtest.cc:2114
#16 0x00007f5729ffa394 in testing::UnitTest::Run (this=0x7f572a3a0da0 <testing::UnitTest::GetInstance()::instance>) at ./src/gtest.cc:3929
#17 0x00007f5729fe111c in RUN_ALL_TESTS () at ../src/gmock/gtest/include/gtest/gtest.h:2288
#18 main (argc=1, argv=0x7fff89df1fb8) at test/common/Throttle.cc:427
</pre>

src/test/common/Throttle.cc
```diff
@@ -319,23 +319,24 @@ std::pair<double, std::chrono::duration<double> > test_backoff(
   auto putter = [&]() {
     std::unique_lock<std::mutex> g(l);
     while (!stop) {
-      while (in_queue.empty())
+      if(in_queue.empty()) {
```
athanatos: space between if and (

wjwithagen:
@dachary 
I did not look at the PR until now.
That is way much better that what I could get out of gdb.

I'm sort of suspicious for this in thread 3 an 4
'''
4: WaitUntil (when=..., mutex=..., this=0x2ab2d2df9060) at ./common/Cond.h:72
'''
but that is because the WaitUntil() has bitten me once before.
But that stack is not really all that deep, so it might be totally oke. 

One thing is sure the 5 secs runtime have expired and all threads are still waiting.
So something should have given a free lock to make progress.

I'm running unittest_throttle without your patch, but just as it is in master.

Got to go and make diner '@home' perhaps more after that.

dachary:
@athanatos space added. I think the [other crash](https://github.com/ceph/ceph/pull/9455#issuecomment-223324010) is unrelated and can be fixed in another pull request. This happens more often and there is value in merging it sooner rather than later because it creates a significant amount of false negative in the make check jenkins run.

wjwithagen:
On 2-6-2016 16:59, Loic Dachary wrote:
>     But now the test gets stuck in:
> 
> Is this with the current master code ? Or with this pull request ? If
> you get this crash with this pull request, I'm very interested in the
> gdb stack traces.

What about reverting the last patch?

Or asserting that the break is never taken without underruning the queue
which is assumed to be oke?

wjwithagen:
On 2-6-2016 16:59, Loic Dachary wrote:
>     But now the test gets stuck in:
> 
> Is this with the current master code ? Or with this pull request ? If
> you get this crash with this pull request, I'm very interested in the
> gdb stack traces.

I took my annotated code and put your patch in.

It actually hangs on the (ticket)->wait(l) in the getter while loop
And to continue it will require lock(l) to be released, and that would
require at least one putter to be active....

But they all ran to completion because stop is set.
You also will not get to the notify_all() because you are still joining
getters.

Note that wait(l) is called when the queue is considered too full to
actually put more in the queue. And we need to wait until somebody
cleared a bit of the queue. Where we check again if there is enough room
in the queue.
(Which makes me wonder is this does not penalize large gets too much,
since smaller ones have a larger chance to actually fit in the cleared
window. )

I first suggested to replace the wait_until(start+delay) with
wait_for(delay), as to prevent a negative wait value.
But Josh was afraid that it would influence throughput in the other
cases where the wait was not required.

So he suggested the current solution, which seemed to work.
But perhaps a timeout on that wait(l), and/or conditionalize it with a
predicate that indicates that the Throttle is being stopped/cleared.
How could BackoffThrottle know that the end is near?

But then the question is what would happen to items that are arriving
when the Throttle is winding down and items go in, but are not taken out.

ErwanAliasr1:
Any updates on that front ?

wjwithagen:
I've been running 3 while's on my box since yesterday evening.
containing: (c-shell)
```
timeout -s SEGV 60 unittest_throttle --debug-throttle=20 --gtest_filter=BackoffThrottle.undersaturated > & /tmp/throttle.under.log
timeout -s SEGV 60 unittest_throttle --debug-throttle=20 --gtest_filter=BackoffThrottle.balanced >& /tmp/throttle.balanced.log
timeout -s SEGV 60 unittest_throttle --debug-throttle=20 --gtest_filter=BackoffThrottle.oversaturated >& /tmp/throttle.over.log
```

I have yet to see one crash or lockup.
The only reason these tests fail is because the run out-of-bounds in the result testing fase:
Like: ```
  ASSERT_LT(results.first, 101);
//  ASSERT_GT(results.first, 85);
//  ASSERT_GT(results.first, 20);
  ASSERT_LT(results.second.count(), 0.002);
  ASSERT_GT(results.second.count(), 0.0005);
```
I've seen reasons in the log file why the bounds might not be appropriate wuith the current version of 'Backthrottle::get'. Major issue I saw is that it is possible to delay for as much as 0.005 in the while loop. And if that is the case than you actually are never going to make:
```
ASSERT_LT(results.second.count(), 0.002);
```
Note that I also run a serious amount of log writting, so that could influence timing as well
Extending the bounds or even removing them has kept thing going without hitch.

**Comment by wjwithagen** [comment]
@ErwanAliasr1 
You should have asked sooner.....
So the lockup as I described above does really happen more frequently

From my annotated log: 
```
    -4> 2016-06-03 16:48:03.028607 803816400 20 throttle::_get_delay:355( 0x7fffffffdf28) :355 Entry c = 8 current = 96
    -3> 2016-06-03 16:48:03.028614 803816400 20 throttle::_get_delay:361( 0x7fffffffdf28) :361 r = 0.96
    -2> 2016-06-03 16:48:03.028620 803816400 20 throttle::_get_delay:375( 0x7fffffffdf28) :375 return r > high_threshhold (queue is full): time = 0.0736 c = 8 duration = 0.0092

    -1> 2016-06-03 16:48:03.028629 803816400 20 throttle::get:413( 0x7fffffffdf28) :413 wait(l)
     0> 2016-06-03 16:48:57.838293 803815000 -1 *** Caught signal (Illegal instruction) ** in thread 803815000 thread_name:
```
Which means that the test got aborted by timeout in throttle::get at the `(*ticket)->wait(l)` at a point where the queue is 96% full. 
So what happens is that one thread needs to write 8, but there is only room for 4, so we wait until something has stirred the queue. But the threads doing that have already stopped.

Reversing putters and getters is not going to work because than it can/will block in the putter-loop on:
```
empty = in_queue.empty();
if (empty) {
		c.wait(g);
```

And then there could be a putter waiting for a getter to actually put something in, before it can continue. This one is relatively easy to "fix", since it is in the part of the unit_testing code, and there it is known that we want to stop AND we know that the throttle is empty. So adding a timeout there would do little harm. It will burn a bit of CPU cycles.But that is the worst it will do.
So I suggest replacing the `wait(g)` with  something like `wait(g, 1ms)` (which is C++14 standard, but you get the idea.)

You can not do the same with the getter, because there the wait is in the library. So you need to push STOP into the library. So 2 ways forward: (or 3 perhaps)
1. Allow get to terminate without getting anything and fix that in the surrounding code by checking on full and than not call get(). Which is like the put() code.
2. implement BackoffThrottle::stop that will signal the wait, and terminate the the `while(true)` in BackoffThrottle
2a. Like in 2. but do not signal the wait, but replace the `wait(l)` with `wait_for(l, 1ms)` 

The reason it cannot be "symmetrical" is due to the fact that throttleing is done by waiting/delaying in the get stage of the throttle

So what I'm wondering: Is this just a (atificial) flaw in the testingcode, where this situation cannot occur in the actual usuage of the BackoffThrottle. Or could this also happen in the code where the Throttle is used?

**Comment by ErwanAliasr1** [comment]
@wjwithagen I don't get why I should have asked sooner. Take your time to find the proper solution. I don't know that part of the code and its usage. 
I'm tracking that issue only because it does lock a build of 20 on the CI ;)
Good luck !

**Comment by wjwithagen** [comment]
@ErwanAliasr1 
It was mere humor. Things we running just fine for a long time, until you asked....
And then I ran into a bug trigger.
So had you asked earlier, it might have also triggered earlier. 
Sort of like a quantum-leap...

**Comment by wjwithagen** [comment]
@dachary 
All I try but this issue keeps getting stuck in the loop
```
while (true) {
    if (!((max == 0) || (current == 0) || (current + c) <= max)) {
      (*ticket)->wait(l);
    }
while (true) {
    if (!((max == 0) || (current == 0) || (current + c) <= max)) {
      dout(20) << ":" << __LINE__ << " wait(l)" << dendl;
      (*ticket)->wait(l);
    } else if (delay > std::chrono::duration<double>(0)) {
      (*ticket)->wait_for(l, delay);
    } else {
      break;
    }
    .....
  }
```
IF there are no more puts, then the condition is not going to change.
Delay will grow negative, and we do not keep stuck here.
TRUE remains TRUE, sad but true.

So the one way to break out of the loop is to change the condition:
` !((max == 0) || (current == 0) || (current + c) <= max))`
Either by setting MAX or CURRENT to 0, but that requires an helper function.

And if a helper function is the way to go, I'd suggest a set_stop() and add a `bool running` to the class. And then still add a positive time-clause to the wait(l),

wjwithagen:
@ldachary 
I looked at my old attempt to fix this in #9111, where I tried fixing the problem that FreeBSD does not like it that the time in wait_until( l, past) is already beyond us.
And that code certainly will not deadlock, because there is a moment in time where (start+delay) < now.

But @athanatos did not like it because it might cause a too long waiting.
And looking at it now, I'm not sure that is correct. Since it wait for maximal delay and then return.
Which is what the original does as well. The delay can be changed if the queue filling changes but then the lock is signaled anyways.
I certainly do not get reports from running unittest_throttle at the tests have execed any boundaries.

So I think if anything  we need to work at setting delay to a correct positive value (negative would stall on FreeBSD). But even that is already adjusted to the current fill degree of the throttle.
Otherwise there perhaps is an easy way to define a minimum time_quantum to wait, based on the parameters passed to the throttle on setup.

wjwithagen:
@athanatos @ldachary @ErwanAliasr1 
I just reopened #9111, an added conditional short waiting when the time would be negative.
This is as close as possible to the original code (before #9235) but still works for FreeBSD.

ErwanAliasr1:
I hope that you'll find a good solution as that still lock a lot of builds.

wjwithagen:
@dachary 
Not sure if the ldachary user notices arive with you as well.
But I think #9111 is actually working. Left it running for the whole day, and no crashes.

dachary:
test this please (making sure it still is passing)

wjwithagen:
@dachary 
Not quite sure what you mean....
This will fail once every few 100 runs, because the throttle is not drained in a correct fashion.
test/common/Throttle.cc needs the 2 fixes that are in:
https://github.com/ceph/ceph/pull/9111/commits/4305248298d68bf7d59370e48860b3df46673dc0
https://github.com/ceph/ceph/pull/9111/commits/08decf5a5e7daccd9e94a74775263a1e2c94726f

That passed in over 10.000 tests on an overloaded server

wjwithagen:
@dachary @athanatos 
I have no capabilities to close pulls other that my own.
And Loic has been silent on the topic for over 15 days, after commenting that his fix still had some problems when testing in large volumes. The problem being that the test could get stuck waiting on a lock

So I worked further on that.
And my working code actually contains the code of this change to test/common/Throttle.cc
And the code that is currently in #9111, to shutdown the throttle in an orderly fashion.

So in my regards the required commits are
#9455 
#9111

athanatos:
Closing in favor of 9111

PR #9456, opened 2016-06-02
msg: don't truncate message sequence to 32-bits

ukernel:
Message sequence is 64-bits number, but the upper 32-bits get lost
when setting/getting message sequence. This causes messenger to
malfunction after receiving 1^32 messages

Fixes: http://tracker.ceph.com/issues/16122

### DIFF
--- a/src/msg/Message.h
+++ b/src/msg/Message.h
@@ -422,8 +422,8 @@ public:
   uint64_t get_tid() const { return header.tid; }
   void set_tid(uint64_t t) { header.tid = t; }
 
-  unsigned get_seq() const { return header.seq; }
-  void set_seq(unsigned s) { header.seq = s; }
+  uint64_t get_seq() const { return header.seq; }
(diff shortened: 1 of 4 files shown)


liewegas:
this looks right to me!

idryomov:
FWIW, I still have this concern: http://www.spinics.net/lists/ceph-devel/msg30599.html.  How is patched messenger supposed to work with the unpatched one?  Did you test it?

idryomov:
@ukernel Nevermind, I think I see what you meant in your reply on ceph-devel.  It's broken as is, so it doesn't matter...

gregsfortytwo:
Looks good to me as well.
Will our existing tests properly cover testing it against an older messenger (just in case) or would we need to write some?

I presume it'll need to get backported at least to Jewel, and that downstreams will want to grab it as well.

lagonix:
Are there by any chance a possibility to dump this up a bit as we are hitting this bug almost every week unless we do a mds failover?
Thank you

PR #9457, opened 2016-06-02
rgw: can set negative max_buckets on RGWUserInfo

yehudasa:
This can be used to disable user's bucket creation

Fixes: http://tracker.ceph.com/issues/14534


ErwanAliasr1:
That's still segfaulting, did you rebased to get @liewegas fix from yesterday ?

yehudasa:
@ErwanAliasr1 I don't think this was rebased, shouldn't matter for testing, as we'd merge it with the latest.

PR #9459, opened 2016-06-02
test/rbd: fsx needs to shut down the journal

dillaman:
Fixes: http://tracker.ceph.com/issues/16123


trociny:
lgtm

PR #9461, opened 2016-06-02
doc: msgr2 protocol spec

liewegas:


doc/dev/msgr2.rst
```diff
@@ -0,0 +1,198 @@
...
+Message exchange
+----------------
+
+Once a session is stablished, we can exchange messages.
+
+* TAG_MSG: a message::
+
+    ceph_msg_header2
+    front
+    middle
```
yuyuyu101: I think we can drop this...why not put middle into front?

liewegas: On Thu, 2 Jun 2016, Haomai Wang wrote:
> 
> In doc/dev/msgr2.rst:
> 
> > +    __le64 flags
> > +
> > +  - Sent by client when it is ready.  Server replies with same.
> > +  - No flags are defined now, but we might use them later.
> > +  - I'm not sure this is necessary.
> > +
> > +Message exchange
> > +----------------
> > +
> > +Once a session is stablished, we can exchange messages.
> > +
> > +* TAG_MSG: a message::
> > +
> > +    ceph_msg_header2
> > +    front
> > +    middle
> 
> I think we can drop this...why not put middle into front?

It's used by CephFS messages. The kernel uses different buffers for these 
so they have separate len fields in the header.  We can of course read 
them into one big bufferptr and split it between two bufferlists on 
the userspace side...

doc/dev/msgr2.rst
```diff
@@ -0,0 +1,199 @@
...
+------
+
+Both the client and server, upon connecting, send a banner::
+
+  "ceph %x %x\n", protocol_features_suppored, protocol_features_required
+
+The protocol features are a new, distinct namespace.  Initially no
+features are defined or required, so this will be "ceph 0 0\n".
+
+If the remote party advertises features we don't support, we can
```
dillaman: s/advertises features/advertises required features/?

doc/dev/msgr2.rst
```diff
@@ -0,0 +1,199 @@
...
+
+Banner
+------
+
+Both the client and server, upon connecting, send a banner::
+
+  "ceph %x %x\n", protocol_features_suppored, protocol_features_required
+
+The protocol features are a new, distinct namespace.  Initially no
+features are defined or required, so this will be "ceph 0 0\n".
```
gregsfortytwo: I'm not sure this will work for us. There's stuff the kernel still doesn't support that might ordered after the new messenger, isn't there?

doc/dev/msgr2.rst
```diff
@@ -0,0 +1,199 @@
...
+
+  - The client first says AUTH_DONE, and the server replies to
+    acknowledge it.
+
+
+Message frame format
+--------------------
+
+Each frame can take one of two forms.  If FLAG_SIGNED or
+FLAG_ENCRYPTED has been specified::
```
dillaman: ... and after the authentication phase

doc/dev/msgr2.rst
```diff
@@ -0,0 +1,199 @@
...
+
+  confounder (block_size bytes of random garbage)
+  __le32 length
+  tag byte
+  payload
+  signature (sig_size bytes)
+  more confounder padding (to pad data from start of __le32 length out to block size)
```
gregsfortytwo: Given that we're ending with confounder padding, perhaps we should just put all the confounder data at the end? Only having to jump once when doing the signing is probably easier, and it means the logic between authenticated and not will have an easier path.

doc/dev/msgr2.rst
```diff
@@ -0,0 +1,199 @@
...
+
+    __le32 len;
+    method specific payload
+
+* TAG_AUTH_DONE::
+    
+    __le64 flags
+      FLAG_ENCRYPTED  1
+      FLAG_SIGNED     2
+    signature
```
gregsfortytwo: Since this is the first thing that gets signed, it needs to include all the necessarily-trusted prior data. Certainly the method, possibly the selection list (although that would make pipelining hard).

doc/dev/msgr2.rst
liewegas: The TAG_IDENT includes the CPEH_FEATURE_* features.  This part is *just* features in the low-level protocol.

doc/dev/msgr2.rst
liewegas: I thought @mdw-at-linuxbox suggested it was important the confounder be at the beginning.  Not sure if that's true though?

gregsfortytwo: It needs to be at the logical beginning for (many kinds of) authorization, but that doesn't mean it needs to be at the beginning of the physical transmission.
I'm not sure how bad having to do jumps back and forth (or do memcpy-in-place) would be for throughput; that would be the reason to keep it physically ordered the same; but I'm also not sure how much benefit there is to random data at the end?

liewegas: We just want to fill out to a block_size boundary so that we can complete the encryption block and push data over the wire.  We can't stick those extra bytes at the beginning easily because length goes there.  Unless we put len before the confounder.

Hmm, I'm not sure how useful the confounder is on *every* if we are encrypted.  Isn't one at the start sufficient?  In that case, we'd have 3 frame formats.. for unprotected, signed, and encrypted.

doc/dev/msgr2.rst
```diff
@@ -0,0 +1,199 @@
...
+
+* TAG_IDENT: identify ourselves::
+
+    entity_addrvec_t addr(s)
+    __u8   my type (CEPH_ENTITY_TYPE_*)
+    __le32 protocol version
+    __le64 features supported (CEPH_FEATURE_* bitmask)
+    __le64 features required (CEPH_FEATURE_* bitmask)
+    __le64 flags (CEPH_MSG_CONNECT_* bitmask)
+    __le64 cookie (a client identifier, assigned by the sender. unique on the sender.)
```
gregsfortytwo: We spent a lot of time talking about this and I'm not sure if we actually came to consensus.

AFAIK, the only reason to have this is to allow the client to quickly identify itself to the server when re-establishing an existing session. So on first connect, it might as well be zeroed out.
But when reconnecting, it needs to be something that is 1) a unique identifier to the server, 2) has been authenticated in both directions.

Which makes me think that:
*) on first connect, the client fills in zero.
*) On any connect, the server replies to the client with both the client-supplied cookie AND a new cookie for the client to use in future reconnects.
(And if the server doesn't fill in a new one, then the client just doesn't get to do a fast reconnect)

Alternatively we could have a different tag with a different struct for reconnects. Or the server could not include the client-supplied cooke in its authenticated reply, but I'm worried (perhaps baselessly) about replay attacks in that case.

doc/dev/msgr2.rst
liewegas: I think the fact that it is signed by the thing that we just negotiated with auth messages is sufficient.  Clearly if we say method A and then use a signature for method B this will fail.

doc/dev/msgr2.rst
```diff
@@ -0,0 +1,199 @@
...
+    that direction should win the race.  The client should wait for that
+    connection to complete.
+
+* TAG_START: client is ready to start the message exchange session::
+
+    __le64 flags
+
+  - Sent by client when it is ready.  Server replies with same.
+  - No flags are defined now, but we might use them later.
+  - I'm not sure this is necessary.
```
gregsfortytwo: I'm not either. What flags might we want to include here that aren't specified elsewhere?
I just want to cut down the round-trip counts and I think the only reason to have explicit tags would be if we need to do one of
*) authenticate data that was previously sent in the clear (and doesn't apply by this point)
*) make a decision that can't be deterministically done on both sides based on established data (which I don't think applies by this point)
Are there cases I'm missing? Does having a TAG_START here enable either of those cases?

liewegas: Yeah I don't think it does.  I'll be happy to remove this!

doc/dev/msgr2.rst
liewegas: If the auth handshake is does properly, we establish a nonce and unique session key that avoids any replay issues in subsequent frames.  So I don't think we have to worry about replay at this stage.

I agree the client value is useless (and may as well be 0) in some case (e.g., client->mds). I think it is needed for osd->osd or mds->mds, though.  I think just leaving/forcing it to 0 unless the flags indicated it is meaningful is sufficient?

I don't think we can change the value on each iteration since we may get a broken tcp connection and reconnect at that layer before successfully negotiating a reconnect.

doc/dev/msgr2.rst
gregsfortytwo: Actually, maybe that's not right. We need to be able to decrypt the length before we read past it, so we probably do need a leading confounder.
(Wow, this is going to *suck* for fast message decode, isn't it.)

liewegas:
Hmm, I think we might be totally overkilling the confounder here.. does it really need to go in *every* frame?  Why not just one at the end of the auth phase, and thereafter we chain together the signatures or encryptino from the previous frames?

doc/dev/msgr2.rst
```diff
@@ -0,0 +1,250 @@
...
+  determines the structure of the payload.
+
+Phases
+------
+
+A connection has two distinct phases:
+
+#. banner
+#. frame exchange for one or more strams
+
```
zhjwpku: :s/strams/streams  ?

PR #9462, opened 2016-06-02
osd/ReplicatedPG: adjust num_pinned in _delete_oid

athanatos:
Fixes: http://tracker.ceph.com/issues/15952


athanatos:
Tested: sjust@teuthology:/a/samuelj-2016-06-01_11:28:31-rados-wip-sam-testing-distro-basic-smithi/228382/remote

PR #9463, opened 2016-06-02
DNM: Wip rgw s3 crypto sse client

aclamk:
This is first version of server side encryption. It is placed here mostly to ask for early review.
Currently 2 modes are supported:
1)  SSE-C, in which client supplies keys for encryption
2) Implicit encryption. When parameter is set like:
rgw crypt default encryption key = 4YSmvJtBv0aZ7geVgAsdpRnLBEwWSWlMIGnRS8a9TSA=
then all objects data will be encrypted with this key.

S3 tests related to encryption are on https://github.com/ceph/s3-tests/pull/114
It is based on very old master, and I am rebaseing it now.


cbodley:
@aclamk should this be closed in favor of https://github.com/ceph/ceph/pull/11049?

aclamk:
@cbodley Yes, this should be closed. Please close it.

PR #9474, opened 2016-06-03
mds: evict MMDSResolve from delayed_resolve map if it is going to be …

xiexingguo:
…re-disptached

So we don't process the same message twice.

### DIFF
--- a/src/mds/MDCache.cc
+++ b/src/mds/MDCache.cc
@@ -3083,6 +3083,7 @@ void MDCache::handle_resolve(MMDSResolve *m)
 
   if (mds->get_state() < MDSMap::STATE_RESOLVE) {
     if (mds->get_want_state() == CEPH_MDS_STATE_RESOLVE) {
+      discard_delayed_resolve(from);
       mds->wait_for_resolve(new C_MDS_RetryMessage(mds, m));
       return;
     }


src/mds/MDCache.cc
```diff
@@ -3083,6 +3083,7 @@ void MDCache::handle_resolve(MMDSResolve *m)
 
   if (mds->get_state() < MDSMap::STATE_RESOLVE) {
     if (mds->get_want_state() == CEPH_MDS_STATE_RESOLVE) {
+      discard_delayed_resolve(from);
```
jcsp: delayed_resolve is only inserted to further down in this function, so I don't see how it's possible for it to already contain entries in the case where "mds->get_state() < MDSMap::STATE_RESOLVE".  Something I'm missing?

xiexingguo: Not very sure about this too, but since the caller can be `process_delayed_resolve `, it can happen theoretically without considering any further:-(

jcsp: Right, but process_delayed_resolve will only call through to handle_resolve if there is something in MDCache::delayed_resolve, and something only gets put in there once we are >=STATE_RESOLVE.  So as long as we are <STATE_RESOLVE, delayed_resolve is guaranteed to be empty

xiexingguo: OK, got it. Close for now.

PR #9477, opened 2016-06-03
cmake: do not create gtest/gmock shared libraries

tchaikov:
otherwise unit tests crash at exit.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -996,7 +996,12 @@ target_link_libraries(crushtool global)
 install(TARGETS crushtool DESTINATION bin)
 
 # Support/Tools
+# the static gtest/gmock libraries load correctly for us
+set(BUILD_SHARED_LIBS_saved ${BUILD_SHARED_LIBS})
+set(BUILD_SHARED_LIBS OFF)
 add_subdirectory(gmock)
+set(BUILD_SHARED_LIBS ${BUILD_SHARED_LIBS_saved})
+
 add_subdirectory(test)
 set(cephfs_srcs cephfs.cc)
 add_executable(cephfstool ${cephfs_srcs})


tchaikov:
@cbodley i am not sure what is the best practice to support shared **and** static library. but though `BUILD_SHARED_LIBS` is a standard way to choose the shared or static library from a cmake user's point of view.

ErwanAliasr1:
Please rebase against master to avoid the bluestore segfaulting.

tchaikov:
@liewegas is this segfault a known issue? it happened in "unittest_bluefs". but i am not able to reproduce locally.

```
[ RUN      ] BlueFS.mkfs
*** Caught signal (Segmentation fault) **
 in thread 2aad4ed43700 thread_name:bstore_aio
 ceph version 10.2.0-1771-gdd6a1a4 (dd6a1a49aae907c41bc04d7aebff02c79de40ec5)
 1: (()+0x18ef82) [0x2aad419d1f82]
 2: (()+0x10340) [0x2aad4b9d5340]
2016-06-03 08:26:12.348574 2aad41b5ba40 -1 did not load config file, using default settings.
2016-06-03 08:26:12.350720 2aad41b5ba40 -1 WARNING: the following dangerous and experimental features are enabled: *
2016-06-03 08:26:12.357733 2aad4ed43700 -1 *** Caught signal (Segmentation fault) **
 in thread 2aad4ed43700 thread_name:bstore_aio
```

ErwanAliasr1:
Complete job trace at https://jenkins.ceph.com/job/ceph-pull-requests/6684/

cbodley:
@tchaikov i'm starting to feel like BUILD_SHARED_LIBS is giving us more trouble than it's worth. we had a lot more control when STATIC was the default and used ${CEPH_SHARED} to override

i'm fine with merging this as a quick fix, but i'd like to see more discussion around this. mainly about the need to ship static libraries in the -dev packages. would you be willing to bring this up on the ceph-devel list to see if we really do need to do that?

tchaikov:
> would you be willing to bring this up on the ceph-devel list to see if we really do need to do that?

good idea, done!

src/CMakeLists.txt
```diff
@@ -996,8 +996,15 @@ target_link_libraries(crushtool global)
...
 # Support/Tools
+# the static gtest/gmock libraries load correctly for us
+set(BUILD_SHARED_LIBS_saved ${BUILD_SHARED_LIBS})
+set(BUILD_SHARED_LIBS OFF)
 add_subdirectory(gmock)
+set(BUILD_SHARED_LIBS ${BUILD_SHARED_LIBS_saved})
+
 add_subdirectory(test)
+
+set(BUILD_SHARED_LIBS )
```
cbodley: why is this last piece needed?

tchaikov:
@cbodley fixed and repushed.

PR #9476, opened 2016-06-03
packaging: do not package unused unittests in deb

tchaikov:

PR #9475, opened 2016-06-03
debian: package compressor plugins and silence dpkg-shlibdeps warnings  

tchaikov:
- package compressor plugins: these plugins are not used yet, but this change tries to sync debian/rules with ceph.spec.in
- silence dpkg-shlibdeps warnings: the plugins do have unresolved symbols. this is expected.

PR #9478, opened 2016-06-03
[RFC] debian: build deb packages using cmake

tchaikov:
just post the changes in hope that it can have some reviews. and so it won't get lost in a hard disk failure. 

this change depends on some other cmake related changes:

- [x] #9446
- [x] #9475 
- [x] #9476 
- [x] to have a "make dist" counterpart for cmake, and update ceph-build to use it accordingly. because gitbuilder building deb packages always do "make dist" and use the created tarball to do the packaging. as maybe cpack can help on this?
- [ ] and more in http://pad.ceph.com/p/CMake_To_Do

@alimaredia @cbodley you might be interested in this change.


tchaikov:
@liewegas just want to make sure that you are aware that there is a wip for the debian/rules for cmake.

liewegas:
oh, excellent.. this is better than my hacky version!

liewegas:
https://github.com/ceph/ceph/pull/9909

PR #9479, opened 2016-06-03
librbd: support setting rbd_default_features config option via string

Ved-vampir:
Proposal to add feature from http://tracker.ceph.com/issues/15593

### DIFF
--- a/src/common/config_opts.h
+++ b/src/common/config_opts.h
@@ -1211,11 +1211,7 @@ OPTION(rbd_default_format, OPT_INT, 2)
 OPTION(rbd_default_order, OPT_INT, 22)
 OPTION(rbd_default_stripe_count, OPT_U64, 0) // changing requires stripingv2 feature
 OPTION(rbd_default_stripe_unit, OPT_U64, 0) // changing to non-object size requires stripingv2 feature
-OPTION(rbd_default_features, OPT_INT, 61)   // only applies to format 2 images
-					    // +1 for layering, +2 for stripingv2,
-					    // +4 for exclusive lock, +8 for object map
-					    // +16 for fast-diff, +32 for deep-flatten,
-					    // +64 for journaling
+OPTION(rbd_default_features, OPT_STR, "layering,exclusive-lock,object-map,fast-diff,deep-flatten")   // only applies to format 2 images
 
 OPTION(rbd_default_map_options, OPT_STR, "") // default rbd map -o / --options
 
--- a/src/include/rbd/features.h
+++ b/src/include/rbd/features.h
@@ -9,6 +9,20 @@
 #define RBD_FEATURE_DEEP_FLATTEN        (1<<5)
 #define RBD_FEATURE_JOURNALING          (1<<6)
 
+#define RBD_FEATURES_DEFAULT             (RBD_FEATURE_LAYERING | \
+                                         RBD_FEATURE_EXCLUSIVE_LOCK | \
+                                         RBD_FEATURE_OBJECT_MAP | \
+                                         RBD_FEATURE_FAST_DIFF | \
+                                         RBD_FEATURE_DEEP_FLATTEN)
+
+#define RBD_FEATURE_NAME_LAYERING        "layering"
+#define RBD_FEATURE_NAME_STRIPINGV2      "striping"
+#define RBD_FEATURE_NAME_EXCLUSIVE_LOCK  "exclusive-lock"
+#define RBD_FEATURE_NAME_OBJECT_MAP      "object-map"
+#define RBD_FEATURE_NAME_FAST_DIFF       "fast-diff"
+#define RBD_FEATURE_NAME_DEEP_FLATTEN    "deep-flatten"
+#define RBD_FEATURE_NAME_JOURNALING      "journaling"
+
 /// features that make an image inaccessible for read or write by
 /// clients that don't understand them
 #define RBD_FEATURES_INCOMPATIBLE 	(RBD_FEATURE_LAYERING |       \
--- a/src/librbd/Utils.cc
+++ b/src/librbd/Utils.cc
@@ -1,9 +1,17 @@
 // -*- mode:C++; tab-width:8; c-basic-offset:2; indent-tabs-mode:t -*-
 // vim: ts=8 sw=2 smarttab
 
+#include <boost/algorithm/string.hpp>
+
 #include "librbd/Utils.h"
 #include "include/rbd_types.h"
 #include "include/stringify.h"
+#include "include/rbd/features.h"
+#include "common/dout.h"
+
+#define dout_subsys ceph_subsys_rbd
+#undef dout_prefix
+#define dout_prefix *_dout << "librbd: "
 
 namespace librbd {
 namespace util {
@@ -31,5 +39,38 @@ librados::AioCompletion *create_rados_ack_callback(Context *on_finish) {
   return create_rados_ack_callback<Context, &Context::complete>(on_finish);
 }
 
+uint64_t parse_rbd_default_features(CephContext* cct) 
+{
+  int ret = 0;
+  uint64_t value = 0;
+  try {
+    value = std::stoi(cct->_conf->rbd_default_features, NULL, 10);
+  } catch (...) {
+    map<std::string, int> conf_vals = {{RBD_FEATURE_NAME_LAYERING, RBD_FEATURE_LAYERING}, 
+                                       {RBD_FEATURE_NAME_STRIPINGV2, RBD_FEATURE_STRIPINGV2},
+                                       {RBD_FEATURE_NAME_EXCLUSIVE_LOCK, RBD_FEATURE_EXCLUSIVE_LOCK},
+                                       {RBD_FEATURE_NAME_OBJECT_MAP, RBD_FEATURE_OBJECT_MAP},
+                                       {RBD_FEATURE_NAME_FAST_DIFF, RBD_FEATURE_FAST_DIFF},
+                                       {RBD_FEATURE_NAME_DEEP_FLATTEN, RBD_FEATURE_DEEP_FLATTEN},
+                                       {RBD_FEATURE_NAME_JOURNALING, RBD_FEATURE_JOURNALING}};
+    std::vector<std::string> strs;
+    boost::split(strs, cct->_conf->rbd_default_features, boost::is_any_of(","));
+    for (std::vector<std::string>::const_iterator i = strs.begin(); i != strs.end(); ++i) {
+    	std::string feature = *i;
+    	boost::trim(feature);
+      if (conf_vals.find(feature) != conf_vals.end()) {
+        value += conf_vals[feature];
+      } else {
+        ret = -EINVAL;
+        ldout(cct, 1) << "Warning: unknown rbd feature " << feature << dendl;
+      }
+    }
+    if (value == 0 && ret == -EINVAL)
+      value = RBD_FEATURES_DEFAULT;
+  }
+  return value;
+}
+
 } // namespace util
+
 } // namespace librbd
--- a/src/librbd/internal.cc
+++ b/src/librbd/internal.cc
@@ -1205,7 +1205,7 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force,
   {
     CephContext *cct = (CephContext *)io_ctx.cct();
     bool old_format = cct->_conf->rbd_default_format == 1;
-    uint64_t features = old_format ? 0 : cct->_conf->rbd_default_features;
+    uint64_t features = old_format ? 0 : librbd::util::parse_rbd_default_features(cct);
     return create(io_ctx, imgname, size, old_format, features, order, 0, 0);
   }
 
@@ -1256,7 +1256,7 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force,
 
     uint64_t features;
     if (opts.get(RBD_IMAGE_OPTION_FEATURES, &features) != 0) {
-      features = old_format ? 0 : cct->_conf->rbd_default_features;
+      features = old_format ? 0 : librbd::util::parse_rbd_default_features(cct);
     }
     uint64_t stripe_unit = 0;
     uint64_t stripe_count = 0;
--- a/src/test/rbd_mirror/test_ImageReplayer.cc
+++ b/src/test/rbd_mirror/test_ImageReplayer.cc
@@ -93,7 +93,7 @@ public:
 					       m_remote_ioctx));
 
     m_image_name = get_temp_image_name();
-    uint64_t features = g_ceph_context->_conf->rbd_default_features;
+    uint64_t features = librbd::util::parse_rbd_default_features(g_ceph_context);
     features |= RBD_FEATURE_EXCLUSIVE_LOCK | RBD_FEATURE_JOURNALING;
     int order = 0;
     EXPECT_EQ(0, librbd::create(m_remote_ioctx, m_image_name.c_str(), 1 << 22,
--- a/src/test/rbd_mirror/test_PoolWatcher.cc
+++ b/src/test/rbd_mirror/test_PoolWatcher.cc
@@ -84,7 +84,7 @@ TestPoolWatcher() : m_lock("TestPoolWatcherLock"),
 
   void create_image(const string &pool_name, bool mirrored=true,
 		    string *image_name=nullptr) {
-    uint64_t features = g_ceph_context->_conf->rbd_default_features;
+    uint64_t features = librbd::util::parse_rbd_default_features(g_ceph_context);
     string name = "image" + stringify(++m_image_number);
     if (mirrored) {
       features |= RBD_FEATURE_EXCLUSIVE_LOCK | RBD_FEATURE_JOURNALING;
@@ -132,7 +132,7 @@ TestPoolWatcher() : m_lock("TestPoolWatcherLock"),
       ictx->state->close();
     }
 
-    uint64_t features = g_ceph_context->_conf->rbd_default_features;
+    uint64_t features = librbd::util::parse_rbd_default_features(g_ceph_context);
     string name = "clone" + stringify(++m_image_number);
     if (mirrored) {
       features |= RBD_FEATURE_EXCLUSIVE_LOCK | RBD_FEATURE_JOURNALING;
--- a/src/tools/rbd/ArgumentTypes.cc
+++ b/src/tools/rbd/ArgumentTypes.cc
@@ -3,6 +3,7 @@
 
 #include "tools/rbd/ArgumentTypes.h"
 #include "tools/rbd/Shell.h"
+#include "tools/rbd/Utils.h"
 #include "include/rbd/features.h"
 #include "common/config.h"
 #include "common/strtol.h"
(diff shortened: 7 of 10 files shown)


src/common/config_opts.h
```diff
@@ -1211,7 +1211,7 @@ OPTION(rbd_default_format, OPT_INT, 2)
 OPTION(rbd_default_order, OPT_INT, 22)
 OPTION(rbd_default_stripe_count, OPT_U64, 0) // changing requires stripingv2 feature
 OPTION(rbd_default_stripe_unit, OPT_U64, 0) // changing to non-object size requires stripingv2 feature
-OPTION(rbd_default_features, OPT_INT, 61)   // only applies to format 2 images
+OPTION(rbd_default_features, OPT_STR, "layering,exclusive lock,object map,fast-diff,deep-flatten")   // only applies to format 2 images
```
dillaman: It would be nice to keep the names consistent between the config and the rbd CLI (i.e. exclusive lock -> exclusive-lock).  I would recommend creating constants in src/include/rbd/features.h for each of the features that both the rbd CLI and librbd can utilize.

src/common/config.cc
```diff
@@ -113,6 +114,27 @@ int ceph_resolve_file_search(const std::string& filename_list,
   return ret;
 }
 
+int parse_rbd_default_features(const std::string& str_value, uint64_t& value)
```
dillaman: This should live in librbd, perhaps in the Utils.h/cc.

src/common/config.cc
```diff
@@ -113,6 +114,27 @@ int ceph_resolve_file_search(const std::string& filename_list,
   return ret;
 }
 
+int parse_rbd_default_features(const std::string& str_value, uint64_t& value) 
+{
+  int ret = 0;
+  value = 0;
+  try {
+    std::stoi(str_value, NULL, 10);
```
dillaman: ```value = std::stoi(...)```

src/common/config.cc
```diff
@@ -113,6 +114,27 @@ int ceph_resolve_file_search(const std::string& filename_list,
...
+  } catch (...) {
+    map<std::string, int> conf_vals = {{"layering", 1}, {"stripingv2", 2}, {"exclusive lock", 4},
+                                       {"object map", 8}, {"fast-diff", 16}, {"deep-flatten", 32}, {"journaling", 64}};
+    std::vector<std::string> strs;
+    boost::split(strs, str_value, boost::is_any_of(","));
+    for (std::vector<std::string>::const_iterator i = strs.begin(); i != strs.end(); ++i)
+      if (conf_vals.find(*i) != conf_vals.end())
+        value += conf_vals[*i];
+      else
+        ret = -1;
```
dillaman: -EINVAL

src/common/config.cc
```diff
@@ -113,6 +114,27 @@ int ceph_resolve_file_search(const std::string& filename_list,
...
 }
 
+int parse_rbd_default_features(const std::string& str_value, uint64_t& value) 
+{
+  int ret = 0;
+  value = 0;
+  try {
+    std::stoi(str_value, NULL, 10);
+  } catch (...) {
+    map<std::string, int> conf_vals = {{"layering", 1}, {"stripingv2", 2}, {"exclusive lock", 4},
```
dillaman: Use the feature constants from src/include/rbd/features.h

src/common/config.cc
```diff
@@ -113,6 +114,27 @@ int ceph_resolve_file_search(const std::string& filename_list,
...
+                                       {"object map", 8}, {"fast-diff", 16}, {"deep-flatten", 32}, {"journaling", 64}};
+    std::vector<std::string> strs;
+    boost::split(strs, str_value, boost::is_any_of(","));
+    for (std::vector<std::string>::const_iterator i = strs.begin(); i != strs.end(); ++i)
+      if (conf_vals.find(*i) != conf_vals.end())
```
dillaman: Might be nice to ```boost::trim``` the tokens in case the features were encoded with whitespace like "layering, object-map"

src/librbd/internal.cc
```diff
@@ -1205,7 +1205,16 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force,
...
     bool old_format = cct->_conf->rbd_default_format == 1;
-    uint64_t features = old_format ? 0 : cct->_conf->rbd_default_features;
+    uint64_t features;
+    if (old_format)
+      features = 0;
+    else {
+      int r = parse_rbd_default_features(cct->_conf->rbd_default_features, features);
+      if (r < 0) {
+        CephContext *cct = (CephContext *)io_ctx.cct();
+        ldout(cct, 1) << "Warning: error in rbd_default_features config value" << dendl;
```
dillaman: Might be nice to move the logging to within the parse method (just pass the CephContext) so that you can log which values caused an error.

src/common/config.cc
dillaman: Can you create a new ```RBD_FEATURES_DEFAULT``` constant in src/include/rbd/features.h and set the value to that constant upon failure.

src/common/config.h
```diff
@@ -266,6 +266,8 @@ typedef enum {
 	OPT_ADDR, OPT_U32, OPT_U64, OPT_UUID
 } opt_type_t;
 
+int parse_rbd_default_features(const std::string& str_value, uint64_t& value);
```
dillaman: Minor: use pointer for out-vals instead of reference

src/test/rbd_mirror/test_ImageReplayer.cc
```diff
@@ -93,7 +93,8 @@ class TestImageReplayer : public ::testing::Test {
 					       m_remote_ioctx));
 
     m_image_name = get_temp_image_name();
-    uint64_t features = g_ceph_context->_conf->rbd_default_features;
+    uint64_t features;
+    parse_rbd_default_features(g_ceph_context->_conf->rbd_default_features, features);
```
dillaman: Probably just set features to the new ```RBD_FEATURES_DEFAULT | RBD_FEATURE_JOURNALING```

src/test/rbd_mirror/test_PoolWatcher.cc
```diff
@@ -84,7 +84,8 @@ TestPoolWatcher() : m_lock("TestPoolWatcherLock"),
 
   void create_image(const string &pool_name, bool mirrored=true,
 		    string *image_name=nullptr) {
-    uint64_t features = g_ceph_context->_conf->rbd_default_features;
+    uint64_t features;
```
dillaman: Just set ```features``` to ```RBD_FEATURES_DEFAULT```

src/test/rbd_mirror/test_PoolWatcher.cc
```diff
@@ -132,7 +133,8 @@ TestPoolWatcher() : m_lock("TestPoolWatcherLock"),
       ictx->state->close();
     }
 
-    uint64_t features = g_ceph_context->_conf->rbd_default_features;
+    uint64_t features;
```
dillaman: Just set ```features``` to ```RBD_FEATURES_DEFAULT```

src/tools/rbd/ArgumentTypes.cc
```diff
@@ -289,7 +289,9 @@ std::string get_short_features_help(bool append_suffix) {
 
     std::string suffix;
     if (append_suffix) {
-      if ((pair.first & g_conf->rbd_default_features) != 0) {
+      uint64_t features;
```
dillaman: Just compare against ```RBD_FEATURES_DEFAULT```

src/test/rbd_mirror/test_ImageReplayer.cc
Ved-vampir: I'm not sure, if it's correct to use default value in tests?

src/tools/rbd/ArgumentTypes.cc
Ved-vampir: And here I'm not sure too, if it's correct: as I understand, this is help message, which show, that feature are turned on now, or I'm not right?

dillaman: You are correct -- it should show the active default features not the hard-coded default features.

src/include/rbd/features.h
```diff
@@ -9,6 +9,16 @@
 #define RBD_FEATURE_DEEP_FLATTEN        (1<<5)
 #define RBD_FEATURE_JOURNALING          (1<<6)
 
+#define RBD_FEATURE_DEFAULT             61
```
dillaman: Use the actual constants -- see examples below

src/include/rbd/features.h
```diff
@@ -9,6 +9,16 @@
...
+
+#define RBD_FEATURE_STR_LAYERING        "layering"
```
dillaman: Minor: substitute "_STR_"-style for "_NAME"  (e.g. RBD_FEATURE_LAYERING_NAME)

src/librbd/Utils.cc
```diff
@@ -32,4 +40,37 @@ librados::AioCompletion *create_rados_ack_callback(Context *on_finish) {
...
+
+int parse_rbd_default_features(CephContext* cct, const std::string& str_value, uint64_t* value) 
+{
+  int ret = 0;
+  *value = 0;
+  try {
+    *value = std::stoi(str_value, NULL, 10);
+  } catch (...) {
+    map<std::string, int> conf_vals = {{RBD_FEATURE_STR_LAYERING, RBD_FEATURE_LAYERING}, 
+    								   {RBD_FEATURE_STR_STRIPINGV2, RBD_FEATURE_STRIPINGV2},
```
dillaman: weird spacing

src/librbd/Utils.cc
```diff
@@ -32,4 +40,37 @@ librados::AioCompletion *create_rados_ack_callback(Context *on_finish) {
...
+    for (std::vector<std::string>::const_iterator i = strs.begin(); i != strs.end(); ++i) {
+    	std::string feature = *i;
+    	boost::trim(feature);
+      if (conf_vals.find(feature) != conf_vals.end())
+        *value += conf_vals[feature];
+      else {
+        ret = -EINVAL;
+        ldout(cct, 1) << "Warning: unkhown rbd feature " << feature << dendl;
+      }
+		}
```
dillaman: weird spacing

src/librbd/Utils.cc
```diff
@@ -32,4 +40,37 @@ librados::AioCompletion *create_rados_ack_callback(Context *on_finish) {
...
+    std::vector<std::string> strs;
+    boost::split(strs, str_value, boost::is_any_of(","));
+    for (std::vector<std::string>::const_iterator i = strs.begin(); i != strs.end(); ++i) {
+    	std::string feature = *i;
+    	boost::trim(feature);
+      if (conf_vals.find(feature) != conf_vals.end())
+        *value += conf_vals[feature];
+      else {
+        ret = -EINVAL;
+        ldout(cct, 1) << "Warning: unkhown rbd feature " << feature << dendl;
```
dillaman: Spelling: "unknown"

src/librbd/internal.cc
```diff
@@ -1205,7 +1205,16 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force,
...
     bool old_format = cct->_conf->rbd_default_format == 1;
-    uint64_t features = old_format ? 0 : cct->_conf->rbd_default_features;
+    uint64_t features;
+    if (old_format)
+      features = 0;
+    else {
+      int r = librbd::parse_rbd_default_features(cct, cct->_conf->rbd_default_features, &features);
+      if (r < 0) {
+        CephContext *cct = (CephContext *)io_ctx.cct();
+        ldout(cct, 1) << "Warning: error in rbd_default_features config value" << dendl;
```
dillaman: These warnings are now probably duplicates of the log message from ```parse_rbd_default_features```

src/librbd/Utils.cc
```diff
@@ -32,4 +40,37 @@ librados::AioCompletion *create_rados_ack_callback(Context *on_finish) {
 }
 
 } // namespace util
+
+int parse_rbd_default_features(CephContext* cct, const std::string& str_value, uint64_t* value)
```
dillaman: Move to within the ```util``` namespace

src/common/config_opts.h
```diff
@@ -1211,7 +1211,7 @@ OPTION(rbd_default_format, OPT_INT, 2)
 OPTION(rbd_default_order, OPT_INT, 22)
 OPTION(rbd_default_stripe_count, OPT_U64, 0) // changing requires stripingv2 feature
 OPTION(rbd_default_stripe_unit, OPT_U64, 0) // changing to non-object size requires stripingv2 feature
-OPTION(rbd_default_features, OPT_INT, 61)   // only applies to format 2 images
+OPTION(rbd_default_features, OPT_STR, "layering,exclusive-lock,object-map,fast-diff,deep-flatten")   // only applies to format 2 images
 					    // +1 for layering, +2 for stripingv2,
```
dillaman: Minor: these comments are no longer required.

src/include/rbd/features.h
```diff
@@ -9,6 +9,20 @@
 #define RBD_FEATURE_DEEP_FLATTEN        (1<<5)
 #define RBD_FEATURE_JOURNALING          (1<<6)
 
+#define RBD_FEATURE_DEFAULT             (RBD_FEATURE_NAME_LAYERING | \
```
dillaman: Minor: RBD_FEATURES_DEFAULT

src/librbd/internal.cc
```diff
@@ -1205,7 +1205,16 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force,
...
     bool old_format = cct->_conf->rbd_default_format == 1;
-    uint64_t features = old_format ? 0 : cct->_conf->rbd_default_features;
+    uint64_t features;
+    if (old_format)
+      features = 0;
+    else {
+      int r = librbd::util::parse_rbd_default_features(cct, cct->_conf->rbd_default_features, &features);
+      if (r < 0) {
+        CephContext *cct = (CephContext *)io_ctx.cct();
+        ldout(cct, 1) << "Warning: error in rbd_default_features config value" << dendl;
```
dillaman: Minor: not needed since the parse function will already output a warning.

src/tools/Makefile-client.am
```diff
@@ -72,6 +72,7 @@ noinst_HEADERS += \
 rbd_LDADD = \
 	libjournal.la libcls_journal_client.la \
 	libcls_rbd_client.la libcls_lock_client.la \
+	librbd_internal.la \
```
dillaman: We probably don't want to pull in librbd.so and librbd_internal.a.  While I don't like the idea of duplicate code, I think you just need to have an rbd CLI version and a librbd version of "parse_rbd_default_features".

src/librbd/Utils.cc
```diff
@@ -31,5 +39,38 @@ librados::AioCompletion *create_rados_ack_callback(Context *on_finish) {
   return create_rados_ack_callback<Context, &Context::complete>(on_finish);
 }
 
+int parse_rbd_default_features(CephContext* cct, const std::string& str_value, uint64_t* value)
```
trociny: As we always ignore return value I think it is better to return features. Also, it seems we always use cct->_conf->rbd_default_features as str_value, so may be do this inside this function, and rename the function just to rbd_default_features()?

It would simplify the code to:
```
uint64_t features = old_format ? 0 : utils::rbd_default_features(cct);
```
instead of
```
uint64_t features;
      if (old_format)
        features = 0;
      else {
        librbd::util::parse_rbd_default_features(cct, cct->_conf->rbd_default_features, &features);
      }
```

src/librbd/Utils.cc
```diff
@@ -31,5 +39,38 @@ librados::AioCompletion *create_rados_ack_callback(Context *on_finish) {
...
+                                       {RBD_FEATURE_NAME_FAST_DIFF, RBD_FEATURE_FAST_DIFF},
+                                       {RBD_FEATURE_NAME_DEEP_FLATTEN, RBD_FEATURE_DEEP_FLATTEN},
+                                       {RBD_FEATURE_NAME_JOURNALING, RBD_FEATURE_JOURNALING}};
+    std::vector<std::string> strs;
+    boost::split(strs, str_value, boost::is_any_of(","));
+    for (std::vector<std::string>::const_iterator i = strs.begin(); i != strs.end(); ++i) {
+    	std::string feature = *i;
+    	boost::trim(feature);
+      if (conf_vals.find(feature) != conf_vals.end())
+        *value += conf_vals[feature];
```
trociny: According to Ceph Coding style we try to follow Google's coding guide: https://google.github.io/styleguide/cppguide.html

And it states: "if one part of an if-else statement uses curly braces, the other part must too"

src/librbd/Utils.cc
Ved-vampir: Good idea, I think, thanks.

dillaman:
lgtm

trociny:
lgtm

dillaman:
retest this please

trociny:
The things complicate a bit :-)
```
RBD_FEATURES=127 LD_LIBRARY_PATH=.libs PYTHONPATH=$(pwd)/build/lib.linux-x86_64-2.7 nosetests -v test/pybind/test_rbd.py -m test_create_defaults              
test_rbd.test_create_defaults ... ERROR

ERROR: test_rbd.test_create_defaults
    self.test(*self.arg)
  File "/SSD/ceph/ceph.upstream/src/test/pybind/test_rbd.py", line 172, in test_create_defaults
    check_default_params(1)
  File "/SSD/ceph/ceph.upstream/src/test/pybind/test_rbd.py", line 168, in check_default_params
    rados.conf_set(k, v)
  File "rados.pyx", line 425, in rados.requires.wrapper.validate_func (/SSD/ceph/ceph.upstream/src/build/rados.c:4235)
  File "rados.pyx", line 740, in rados.Rados.conf_set (/SSD/ceph/ceph.upstream/src/build/rados.c:9285)
Error: error calling conf_set: error code 38
```
Now we can't dynamically change this option (via conf_set()). By default only numeric options are allowed to be changed, and for a string option we need an observer (see common/config.cc:md_config_t::set_val()).

So we would need to protect rbd_default_features() with a lock, use internal variable to keep default values, and add an observer that would update this variable. I guess common/TracepointProvider.h can be used as an example (see e.g. librbd/librbd.cc how it is used).

@dillaman What do you think?

dillaman:
@trociny Hmm -- not sure that will help us in this case since we wouldn't be able to register the listener until the point in time when the user goes to create an image (which would be too late potentially). It seems like this restriction on strings, etc was a short-term workaround to an issue. I think the better solution would be a dedicated, templated ```get_val```-like function that would take a config lock, read the value, and return a copy. We could hide direct access to rbd_default_features to prevent any chance of someone accesssing it the wrong way.

trociny:
@dillaman The templated get_val looks like a good idea, still we need a way to bypass the current md_config_t::set_val restriction for strings. I think we can't just remove it. May be an additional flag to config_option that informs that it is safely to update this option even if it does not have an observer? Or a new type, e.g. OPT_STR_SAFE...

dillaman:
@trociny Yeah, I was thinking either of some change to the current OPTION() macro or a new OPT_ type which would ensure the generated variables wasn't publicly available and permitted to be updated via set_val.

trociny:
@Ved-vampir do you want to work on this?

Ved-vampir:
@trociny @dillaman Yes, I'll return to this, when I'll finish my current compression task, in a ~ week I think

trociny:
@Ved-vampir If you are going to start extending OPTION(), please also look at the recent discussion "prefered behavior for injecting untracked args" on ceph-dev.

dillaman:
@Ved-vampir still interested in working on this?

Ved-vampir:
@dillaman Yes, I'm on vacation now, then I plan to work on it.

dillaman:
@Ved-vampir ping

asheplyakov:
I'll be working on this patch. I've addressed the problem of changing rbd_default_features at the run time and submitted a new PR: https://github.com/ceph/ceph/pull/11175

PR #9484, opened 2016-06-03
rgw: aws4: add STREAMING-AWS4-HMAC-SHA256-PAYLOAD support

jmunhoz:
When authenticating requests using the Authorization header in AWS4, you have
the option of uploading the payload in chunks. You can send data in fixed size
or variable size chunks.

This patch enables streaming mode and signed headers support with chunked
uploads.

Fixes: http://tracker.ceph.com/issues/16146

Signed-off-by: Javier M. Mellid <jmunhoz@igalia.com>

### DIFF
--- a/src/rgw/rgw_op.cc
+++ b/src/rgw/rgw_op.cc
@@ -2519,14 +2519,23 @@ void RGWPutObj::execute()
   }
 
   do {
-    bufferlist data;
-    len = get_data(data);
+    bufferlist data_in;
+    len = get_data(data_in);
     if (len < 0) {
       op_ret = len;
       goto done;
     }
     if (!len)
       break;
+
+    bufferlist data;
+    if (s->aws4_auth_streaming_mode) {
+      /* use unwrapped data */
+      data = s->aws4_auth->bl;
+      len = data.length();
+    } else {
+      data = data_in;
+    }
 
     /* do we need this operation to be synchronous? if we're dealing with an object with immutable
      * head, e.g., multipart object we need to make sure we're the first one writing to this object
@@ -2579,7 +2588,9 @@ void RGWPutObj::execute()
     ofs += len;
   } while (len > 0);
 
-  if (!chunked_upload && ofs != s->content_length) {
+  if (!chunked_upload &&
+      ofs != s->content_length &&
+      !s->aws4_auth_streaming_mode) {
     op_ret = -ERR_REQUEST_TIMEOUT;
     goto done;
   }
@@ -2625,6 +2636,7 @@ void RGWPutObj::execute()
   hash.Final(m);
 
   buf_to_hex(m, CEPH_CRYPTO_MD5_DIGESTSIZE, calc_md5);
+
   etag = calc_md5;
 
   if (supplied_md5_b64 && strcmp(calc_md5, supplied_md5)) {
--- a/src/rgw/rgw_auth_s3.cc
+++ b/src/rgw/rgw_auth_s3.cc
@@ -277,7 +277,11 @@ void rgw_create_s3_v4_canonical_request(struct req_state *s, const string& canon
     if (s->aws4_auth_needs_complete) {
       request_payload_hash = STREAM_IO(s)->grab_aws4_sha256_hash();
     } else {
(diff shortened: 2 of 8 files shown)


mattbenjamin:
Are there test cases available?  e.g., that we could add to s3tests?

jmunhoz:
@mattbenjamin s3tests uses boto under the hood and I think it doesn't support STREAMING-AWS4-HMAC-SHA256-PAYLOAD

I ran a quick search on boto's home here in github and I found the following issue as the only reference:

https://github.com/boto/boto/issues/2921#issuecomment-74827099

As commented in the issue, maybe @tpodowd added support at the end although I didn't find it upstream...

mattbenjamin:
@jmunhoz  cool, I'll check out the test case, thanks very much

src/rgw/rgw_op.cc
```diff
@@ -2519,15 +2519,24 @@ void RGWPutObj::execute()
...
+    bufferlist data_in;
+    len = get_data(data_in);
     if (len < 0) {
       op_ret = len;
       goto done;
     }
     if (!len)
       break;
 
+    bufferlist data;
```
mattbenjamin: This induces buffer sharing.  Would we get the same result if data were a pointer to data_in or s->aws_auth->bl (saving the atomic ops to attach each buffer::ptr)?

jmunhoz: With a reference we should get the expected behaviour without major changes. The next code works:

<pre>
-    bufferlist data;
+    bufferlist &data = data_in;
     if (s->aws4_auth_streaming_mode) {
       /* use unwrapped data */
       data = s->aws4_auth->bl;
       len = data.length();
-    } else {
-      data = data_in;
     }
</pre>

Does it make sense?

mattbenjamin:
@jmunhoz Yes.  I think we should do a change like this as a separate step;  I verified the original change, and merged;  can you send a new PR?

src/rgw/rgw_op.cc
cbodley: move semantics could help here too, assuming you don't reuse aws4_auth->bl:
```diff
-      data = s->aws4_auth->bl;
+      data = std::move(s->aws4_auth->bl);
```

jmunhoz: yes, I think std::move() would work too.

jmunhoz:
@mattbenjamin the new PR is available at https://github.com/ceph/ceph/pull/9689

tpodowd:
Hi guys,

I am just back from vacation. Anything I can help out with? Sounds like it is all resolved though.

Tom.

jmunhoz:
@tpodowd we added the aws4 chunking support and it is working ok although we are running manual validation via aws-java-sdk in this moment. Adding some test case being part of the s3-tests testing suite would be great though.

To be able to add this test case with automatic validation we would need aws4 chunking (STREAMING-AWS4-HMAC-SHA256-PAYLOAD) support in boto2 although I think boto2 is not supporting it upstream now.

I checked you did some work on boto2 related to this support:

https://github.com/tpodowd/boto/commits/s3_sigv4_chunking

In detail this commit looks some support to add those bits in boto2:

https://github.com/tpodowd/boto/commit/0fdcb661b6560cbdad6cae1b0bdd0f0d6059e72e

Are you considering going upstream with this code in boto2?

tpodowd:
Hi @jmunhoz. Yeah. I have 8 outstanding pull requests for the boto project. I guess since boto moved onto the new code base, they have kind of neglected these pull requests. I was trying for a while to get someone to merge them but it has been hard to find the right person. The patches themselves have been used almost every day by us (Cloudian) internally for testing our code base so are pretty stable.

Here is a list of my outstanding pull requests. If you make a little noise on some of the pull requests, maybe folks will merge them into the main code base.

https://github.com/boto/boto/pulls/tpodowd

jmunhoz:
@tpodowd those patchsets are great stuff. I commented on https://github.com/boto/boto/pull/3014#issuecomment-228390782 to know if it would be possible merging it upstream.

Maybe @jamesls could help us to have some of those patches upstream. Having aws chunked support and fixing the double port bug in boto2 would do a big difference.

tpodowd:
Thanks @jmunhoz. @jamesls - I can help if required.

PR #9486, opened 2016-06-03
os/bluestore: add per-store allocation/compression statistics

ifed01:

### DIFF
--- a/src/os/bluestore/BlueStore.cc
+++ b/src/os/bluestore/BlueStore.cc
@@ -33,6 +33,7 @@
 #define dout_subsys ceph_subsys_bluestore
 
 const string PREFIX_SUPER = "S";   // field -> value
+const string PREFIX_STAT = "T";    // field -> value(int64 array)
 const string PREFIX_COLL = "C";    // collection name -> cnode_t
 const string PREFIX_OBJ = "O";     // object name -> onode_t
 const string PREFIX_OVERLAY = "V"; // u64 + offset -> data
@@ -564,6 +565,32 @@ void BlueStore::Cache::_audit_lru(const char *when)
   }
 }
 #endif
+
+struct Int64ArrayMergeOperator : public KeyValueDB::MergeOperator {
+  virtual void merge_nonexistant(
+    const char *rdata, size_t rlen, std::string *new_value) override {
+    *new_value = std::string(rdata, rlen);
+  }
+  virtual void merge(
+    const char *ldata, size_t llen,
+    const char *rdata, size_t rlen,
+    std::string *new_value) {
+    assert(llen == rlen);
+    assert((rlen % 8) == 0);
+    new_value->resize(rlen);
+    const __le64* lv = (const __le64*)ldata;
+    const __le64* rv = (const __le64*)rdata;
+    __le64* nv = &(__le64&)new_value->at(0);
+    for (size_t i = 0; i < rlen >> 3; ++i) {
+      nv[i] = lv[i] + rv[i];
+    }
+  }
+  // We use each operator name and each prefix to construct the
+  // overall RocksDB operator name for consistency check at open time.
+  virtual string name() const {
+    return "int64_array";
+  }
+};
 
 // BufferSpace
 
@@ -1524,6 +1551,7 @@ int BlueStore::_open_db(bool create)
   snprintf(fn, sizeof(fn), "%s/db", path.c_str());
   string options;
   stringstream err;
+  ceph::shared_ptr<Int64ArrayMergeOperator> merge_op(new Int64ArrayMergeOperator);
 
   string kv_backend;
   if (create) {
@@ -1744,8 +1772,9 @@ int BlueStore::_open_db(bool create)
     env = NULL;
     return -EIO;
   }
-  
+
   FreelistManager::setup_merge_operators(db);
+  db->set_merge_operator(PREFIX_STAT, merge_op);
 
   if (kv_backend == "rocksdb")
     options = g_conf->bluestore_rocksdb_options;
@@ -2893,6 +2922,27 @@ int BlueStore::statfs(struct store_statfs_t *buf)
   buf->blocks = bdev->get_size() / block_size;
   buf->bsize = block_size;
   buf->available = (alloc->get_free() - bluefs_len);
+
+  bufferlist bl;
+  int r = db->get(PREFIX_STAT, "bluestore_statfs", &bl);
+  if (r >= 0) {
+       TransContext::volatile_statfs vstatfs;
+     if (size_t(bl.length()) >= sizeof(vstatfs.values)) {
+       auto it = bl.begin();
+       vstatfs.decode(it);
+
+       buf->allocated = vstatfs.allocated();
+       buf->stored = vstatfs.stored();
+       buf->compressed = vstatfs.compressed();
+       buf->compressed_original = vstatfs.compressed_original();
+       buf->compressed_allocated = vstatfs.compressed_allocated();
+     } else {
+       dout(10) << __func__ << " store_statfs is corrupt, using empty" << dendl;
+     }
+  } else {
+    dout(10) << __func__ << " store_statfs missed, using empty" << dendl;
+  }
+
 
   dout(20) << __func__ << *buf << dendl;
   return 0;
@@ -4263,6 +4313,18 @@ BlueStore::TransContext *BlueStore::_txc_create(OpSequencer *osr)
   return txc;
 }
 
+void BlueStore::_txc_update_store_statfs(TransContext *txc)
+{
+  if (txc->statfs_delta.is_empty())
+    return;
+
+  bufferlist bl;
+  txc->statfs_delta.encode(bl);
+
+  txc->t->merge(PREFIX_STAT, "bluestore_statfs", bl);
+  txc->statfs_delta.reset();
+}
+
 void BlueStore::_txc_state_proc(TransContext *txc)
 {
   while (true) {
@@ -4578,6 +4640,7 @@ void BlueStore::_txc_finalize_kv(TransContext *txc, KeyValueDB::Transaction t)
 
   txc->allocated.clear();
   txc->released.clear();
+  _txc_update_store_statfs(txc);
 }
 
 void BlueStore::_kv_sync_thread()
@@ -5655,6 +5718,7 @@ void BlueStore::_do_write_small(
 	bluestore_lextent_t(blob, b_off + head_pad, length, 0);
       b->ref_map.get(lex.offset, lex.length);
       b->mark_used(lex.offset, lex.length);
+      txc->statfs_delta.stored() += lex.length;
       dout(20) << __func__ << "  lex 0x" << std::hex << offset << std::dec
 	       << ": " << lex << dendl;
       dout(20) << __func__ << "  old " << blob << ": " << *b << dendl;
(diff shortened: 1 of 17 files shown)


src/os/bluestore/bluestore_types.h
```diff
@@ -253,6 +253,7 @@ struct bluestore_blob_t {
 
   vector<bluestore_pextent_t> extents; ///< raw data position on device
   uint32_t length;                 ///< logical (decompressed) length
+  uint32_t clength;                ///< compressed length if any
```
liewegas: compressed_length?  No real need to be concise here, I think.

src/os/bluestore/BlueStore.cc
```diff
@@ -3296,9 +3296,9 @@ int BlueStore::_blob2read_to_extents2read(
     uint64_t x_offs = cur->blob_xoffset;
     while (l > 0 && ext_it != ext_end) {
 
-      assert(blob->length >= ext_pos + r_offs);
+      assert(blob->get_payload_length() >= ext_pos + r_offs);
 
-      uint64_t r_len = MIN(blob->length - ext_pos - r_offs, ext_it->length - r_offs);
+      uint64_t r_len = MIN(blob->get_payload_length() - ext_pos - r_offs, ext_it->length - r_offs);
```
liewegas: Am I reading this correctly that these length -> get_payload_length() changes are subtle bug fixes?

src/osd/osd_types.h
```diff
@@ -4301,4 +4301,103 @@ struct PromoteCounter {
...
+  int64_t available;		// Free blocks available
+
+  int64_t allocated;		// Bytes allocated by the store
+  int64_t stored;		// Bytes actually stored by the user
+  int64_t compressed;		// Bytes stored after compression
+  int64_t compressed_allocated;	// Bytes allocated for compressed data
+  int64_t compressed_original;	// Bytes that were successfully compressed
+
+  int64_t files;		// Total file nodes in filesystem, preserved for backward compatibility!
+  int64_t ffree;		// Free file nodes in filesystem, preserved for backward compatibility!
```
liewegas: could put = 0 on all of these...

src/osd/osd_types.h
```diff
@@ -4301,4 +4301,103 @@ struct PromoteCounter {
...
+
+  store_volatile_statfs_t() {
+    reset_volatile();
```
liewegas: and drop this

src/osd/osd_types.h
```diff
@@ -4301,4 +4301,103 @@ struct PromoteCounter {
...
+  void reset_volatile() {
+    available = 0;
+    allocated = 0;
+    stored = 0;
+    compressed = 0;
+    compressed_allocated = 0;
+    compressed_original = 0;
+
+    files = 0;
+    ffree = 0;
```
liewegas: and do *this = store_volitile_stats_t() here.

src/osd/osd_types.cc
```diff
@@ -5578,3 +5578,49 @@ void OSDOp::merge_osd_op_vector_out_data(vector<OSDOp>& ops, bufferlist& out)
...
 }
+
+void store_volatile_statfs_t::dump(Formatter *f) const
+{
+  f->dump_int("available", available);
+  f->dump_int("allocated", allocated);
+  f->dump_int("stored", stored);
+  f->dump_int("compressed", compressed);
+  f->dump_int("compressed_allocated", compressed_allocated);
+  f->dump_int("compressed_orignal", compressed_original);
```
liewegas: s/orignal/original/

src/os/bluestore/BlueStore.cc
```diff
@@ -3189,7 +3192,8 @@ int BlueStore::_read_whole_blob(const bluestore_blob_t* blob, OnodeRef o, buffer
 
   result->clear();
 
-  uint32_t l = blob->get_payload_length();
+
+  uint32_t l = blob->get_aligned_payload_length(block_size);
```
liewegas: can you separate out the compression-related accounting changes into a separate commit, so that the first commit just does the switchover from statfs to store_statfs_t, and the next commit makes bluestore smarter to fill in extra fields?

src/os/bluestore/BlueStore.cc
```diff
@@ -2253,6 +2253,9 @@ int BlueStore::mount()
   _set_csum();
   _set_compression();
 
+  //FIXME:
+  //statfs_ex.set();
+
```
liewegas: drop this?

src/os/bluestore/bluestore_types.h
ifed01: >compressed_length? No real need to be concise here, I think.
We need that for:
- decrement compressed counter on blob release. I.e. to be able to track real compressed size.
- avoid reading the whole blob for small compressed block

liewegas: I just meant s/clength/compressed_length/

src/osd/osd_types.h
```diff
@@ -4387,13 +4379,17 @@ ostream &operator<<(ostream &lhs, const store_volatile_statfs_t &rhs);
 * ObjectStore full statfs information
 */
 struct store_statfs_t : public store_volatile_statfs_t {
```
liewegas: Instead of separating out these types at this level, which is sort of confusing, I would just defined a separate bluestore-specific type that has the int64_t array structure/alignment and copy the ~5 fields explicitly in BlueStore::statfs().

The bluestore type could even be an int64_t array with an enum for indexes or something?  Maybe

src/os/bluestore/BlueStore.cc
ifed01: Yes.

ifed01:
@liewegas - fixed

src/os/bluestore/BlueStore.cc
```diff
@@ -4166,6 +4214,19 @@ void BlueStore::_assign_nid(TransContext *txc, OnodeRef o)
...
 }
 
+void BlueStore::_update_store_statfs(TransContext *txc)
+{
+  if (txc->statfs_delta.is_empty())
+    return;
+
+  bufferlist bl;
+  bufferptr p((char*)(txc->statfs_delta.values), sizeof(txc->statfs_delta.values));
+  bl.append(p);
```
liewegas: This can be simplified to bl.append(ptr, len)

src/os/bluestore/BlueStore.cc
```diff
@@ -4166,6 +4214,19 @@ void BlueStore::_assign_nid(TransContext *txc, OnodeRef o)
   }
 }
 
+void BlueStore::_update_store_statfs(TransContext *txc)
```
liewegas: _txc_update_store_stats ?

src/os/bluestore/BlueStore.h
```diff
@@ -551,6 +551,46 @@ class BlueStore : public ObjectStore,
...
+        STATFS_ALLOCATED = 0,
+        STATFS_STORED,
+        STATFS_COMPRESSED_ORIGINAL,
+        STATFS_COMPRESSED,
+        STATFS_COMPRESSED_ALLOCATED,
+        STATFS_LAST
+      };
+      int64_t values[STATFS_LAST];
+      volatile_statfs() {
+        memset( this, 0, sizeof(volatile_statfs));
```
liewegas: memset(this ...

src/os/bluestore/BlueStore.h
```diff
@@ -551,6 +551,46 @@ class BlueStore : public ObjectStore,
...
+      }
+      int64_t& compressed_allocated() {
+        return values[STATFS_COMPRESSED_ALLOCATED];
+      }
+      bool is_empty() {
+        return allocated() == 0 &&
+          stored() == 0 &&
+          compressed() == 0 &&
+          compressed_original() == 0 &&
+          compressed_allocated();
```
liewegas: compressed_allocated() == 0

This might be simpler as a loop over the int array, tho, to avoid needing an update if this struct changes

src/os/bluestore/BlueStore.cc
```diff
@@ -477,6 +478,32 @@ void BlueStore::BufferCache::trim(uint64_t keep)
...
+    const char *rdata, size_t rlen,
+    std::string *new_value) {
+    assert(llen == rlen);
+    assert((rlen % 8) == 0);
+    new_value->resize(rlen);
+    for (size_t i = 0; i < rlen; i+=8) {
+      int64_t v = *(int64_t*)(ldata + i);
+      v += *(int64_t*)(rdata + i);
+      (int64_t&)(new_value->at(i)) = v;
+    }
```
liewegas: Might be simpler to do
<pre>
 int64_t *lv = (int64_t*)ldata;
 int64_t *vv = (int64_t*)rdata;
</pre>
and avoid the awkward casts inside the loop

src/os/filestore/FileStore.cc
```diff
@@ -701,14 +701,24 @@ void FileStore::collect_metadata(map<string,string> *pm)
...
-  }
+  struct statfs buf;
+  buf0->reset();
+  if (::statfs(basedir.c_str(), &buf) < 0) {
+     int r = -errno;
+     assert(!m_filestore_fail_eio || r != -EIO);
+     assert(r != -ENOENT);
+     return r;
+   }
+  buf0->blocks = buf.f_blocks;
```
liewegas: indentation is weird here?

src/osd/osd_types.h
```diff
@@ -4301,4 +4301,31 @@ struct PromoteCounter {
...
+  uint32_t bsize = 0;                  // Optimal transfer block size
+
+  int64_t allocated = 0;               // Bytes allocated by the store
+  int64_t stored = 0;                  // Bytes actually stored by the user
+  int64_t compressed = 0;              // Bytes stored after compression
+  int64_t compressed_allocated = 0;    // Bytes allocated for compressed data
+  int64_t compressed_original = 0;     // Bytes that were successfully compressed
+
+  int64_t files = 0;                   // Total file nodes in filesystem, preserved for backward compatibility!
+  int64_t ffree = 0;                   // Free file nodes in filesystem, preserved for backward compatibility!
```
liewegas: since we don't actually fill these with correct info, let's drop them.  if a caller needs them we can fabricate a fictitious value there.. but hopefully they don't!

src/rocksdb
```diff
@@ -1 +1 @@
-Subproject commit 5647fa427c4eab87c0caf8bcbb91f5846a4623f9
+Subproject commit fa98456ccbd860120127ac2e91a2088fb222606e
```
liewegas: a rocksdb submodule change snuck in!

liewegas:
This looks great.  Needs a rebase.

A few minor cosmetic issues aside, my only concern here is that this effectively fixes the endianness of the ondisk store so that a disk can't be swapped between a big and little endian box.  To fix that, we should probably make the update function use an encoder method (raw_encode maybe.. we don't want the usual versioning wrappers, just want to use the int64_t encode that flips byte order as needed) and the merge operator to use the __le64 types intead of int64_t so that the arithmetic is done properly.

ifed01:
@liewegas: fixed

src/os/bluestore/BlueStore.cc
```diff
@@ -565,6 +566,31 @@ void BlueStore::Cache::_audit_lru(const char *when)
...
+    const char *ldata, size_t llen,
+    const char *rdata, size_t rlen,
+    std::string *new_value) {
+    assert(llen == rlen);
+    assert((rlen % 8) == 0);
+    new_value->resize(rlen);
+    __le64* lv = (__le64*)ldata;
+    __le64* rv = (__le64*)rdata;
+    for (size_t i = 0; i < rlen >> 3; ++i) {
+      (__le64&)(new_value->at(i << 3)) = lv[i] + rv[i];
```
liewegas: lv[i] = lv[i] + rv[i];
?

liewegas:
rebase please

src/os/bluestore/BlueStore.cc
ifed01: IMHO ldata is supposed to be an input (read only) parameter. And result to be stored to new_value. Am I missing something?

liewegas: oh, you're right.  you can do a __le64* for the new value too, though

ifed01:
@liewegas - done

PR #9487, opened 2016-06-03
Allows developers to turn on diagnostic colors for compiler output when stderr is not a terminal

ivancich:
Useful for those who'd like to do:

    make | less -R

Checks if the C compiler supports the -fdiagnostics-color=always option, and then if cmake will use the value of DIAGNOSTICS_COLOR to set the option. Valid values are "auto" (default), "always" (nice for "make | less -R") and "never".

For example, during cmake invocation one could use:

    cmake -DDIAGNOSTICS_COLOR=always [...]

to make sure the above combination of make and less works.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -90,6 +90,20 @@ else()
   message(FATAL_ERROR "The compiler ${CMAKE_CXX_COMPILER} has no C++11 support.")
 endif()
 
+
+# if compiler allows it and if -DDIAG_COLOR_ALWAYS=yes is provided to
+# cmake, turn on diagnostic colors, even when compiler output does not
+# go to terminal (e.g., when piped to less)
+
+CHECK_C_COMPILER_FLAG("-fdiagnostics-color=always"
+  COMPILER_SUPPORTS_DIAG_COLOR_ALWAYS)
+
+if(COMPILER_SUPPORTS_DIAG_COLOR_ALWAYS AND DIAG_COLOR_ALWAYS)
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fdiagnostics-color=always")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fdiagnostics-color=always")
+endif()
+
+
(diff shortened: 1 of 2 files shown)


alimaredia:
@cbodley @tchaikov this looks good to me. Do either of you see anything that concerns you?

src/CMakeLists.txt
```diff
@@ -90,6 +90,26 @@ else()
   message(FATAL_ERROR "The compiler ${CMAKE_CXX_COMPILER} has no C++11 support.")
 endif()
 
+
+# if compiler allows it and if -DDIAG_COLOR_ALWAYS=yes is provided to
+# cmake, turn on diagnostic colors, even when compiler output does not
+# go to terminal (e.g., when piped to less)
+
+CHECK_C_COMPILER_FLAG("-fdiagnostics-color=always"
+  COMPILER_SUPPORTS_DIAG_COLOR_ALWAYS)
```
cbodley: looks like this line could go inside the `if(DIAG_COLOR_ALWAYS)` block below?

cbodley:
i think it would be nice to initialize `DIAG_COLOR_ALWAYS`, either as an [option()](https://cmake.org/cmake/help/v3.0/command/option.html) (if we only want to support ON/OFF) or with [set()](https://cmake.org/cmake/help/v3.0/command/set.html) as a `CACHE` variable (could support arbitrary strings like OFF/ALWAYS/etc). either approach would let us set a default value and add some descriptive text. speaking of which, is there any reason not to enable it by default?

ivancich:
I think those are good ideas, Casey, and I prefer the set() approach as it would give the developer the most flexibility. I'll go ahead and make that change.

The reason I just didn't turn it on by default was that I didn't want to break anything else. For example, if anything in automated testing grepped through the output looking for strings that might be interrupted with the escape sequences  to colorize the text.

tchaikov:
@ivancich could you remove unrelated commits from your changeset in this PR?

ivancich:
@tchaikov @cbodley @alimaredia Let me know if you think this needs any more work before a merge.

PR #9488, opened 2016-06-03
rados test: Consolidate output of err_t and add test case

dzafman:


dzafman:
@tchaikov I moved this change to its own pull request.

dzafman:
In fact the test case happens to not be impacted by the rados part of the fix.  The test only creates "missing" shards so the output is the same.

tchaikov:
`Reviewed-by: Kefu Chai <kchai@redhat.com>` once your test is green.

dzafman:
Combined this single commit into #9613 in order to also keep improving the test case.

PR #9494, opened 2016-06-04
mds/cdir: fix the type of dentry in the judgment; cleanup undefined functions

renhwztetecs:
1. Perfect is_primary/is_remote/is_null type of dentry in the judgment;
2. first_get() and last_put() are not defined,so cleanup them

### DIFF
--- a/src/mds/CDir.cc
+++ b/src/mds/CDir.cc
@@ -557,10 +557,10 @@ void CDir::link_inode_work( CDentry *dn, CInode *in)
 
 void CDir::unlink_inode(CDentry *dn)
 {
-  if (dn->get_linkage()->is_remote()) {
-    dout(12) << "unlink_inode " << *dn << dendl;
-  } else {
+  if (dn->get_linkage()->is_primary()) {
     dout(12) << "unlink_inode " << *dn << " " << *dn->get_linkage()->get_inode() << dendl;
+  } else {
+    dout(12) << "unlink_inode " << *dn << dendl;
   }
 
   unlink_inode_work(dn);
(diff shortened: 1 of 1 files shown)


src/mds/CDir.cc
```diff
@@ -608,10 +608,8 @@ void CDir::unlink_inode_work( CDentry *dn )
       dn->unlink_remote(dn->get_linkage());
 
     dn->get_linkage()->set_remote(0, 0);
-  } else {
+  } else if (dn->get_linkage()->is_primary()) {
```
jcsp: Do we really ever expect a null dentry here?  If not, then there should be an assert(!dn->get_linkage()->is_null()) after this if block.

src/mds/CDir.cc
```diff
@@ -2187,7 +2185,7 @@ void CDir::_encode_dentry(CDentry *dn, bufferlist& bl,
     bl.append('L');         // remote link
     ::encode(ino, bl);
     ::encode(d_type, bl);
-  } else {
+  } else if (dn->linkage.is_primary()) {
```
jcsp: Same as above: if we really never expect a null dentry here, add an assertion before or after the if block

jcsp:
Is this based on any bug you found, or just from code inspection?

src/mds/CDir.cc
renhwztetecs: null dentry Shouldn't be here，so add assert(!dn->get_linkage()->is_null()) in the else branch

src/mds/CDir.cc
renhwztetecs: Same as above
thank you

renhwztetecs:
just from code inspection

jcsp:
Realised after testing that the "cleanup undefined functions" commit is causing lots of crashes -- they weren't really unused: please remove that commit.

renhwztetecs:
@jcsp 
I change it later

renhwztetecs:
@jcsp 
remove that commit of cleanup undefined functions

jcsp:

PR #9495, opened 2016-06-04
rgw: need to 'open_object_section' before dump stats in 'RGWGetUsage_…

weiqiaomiao:
…ObjStore_S3::send_response()' function

### DIFF
--- a/PendingReleaseNotes
+++ b/PendingReleaseNotes
@@ -17,6 +17,24 @@
   listed in 'osd class default list' requires a capability naming the class
   (e.g. 'allow class foo').
 
+* The 'rgw rest getusage op compat' config option allows you to dump the description of user stats
+  or not in s3 GetUsage API. This config option is of bool type and defaults to false.
+  If the value is true, the reponse data of s3 GetUsage like below:
+
+    "stats": {
+                "TotalBytes": 516,
+                "TotalBytesRounded": 1024,
+                "TotalEntries": 1
+             }
+
+  Or if the value is false, the  reponse of s3 GetUsage like below and as it does before we add this config option:
+
+    {
+         516,
+         1024,
+         1
+    }.
+
 11.0.0
 ------
 
--- a/src/common/config_opts.h
+++ b/src/common/config_opts.h
@@ -1478,6 +1478,8 @@ OPTION(rgw_swift_versioning_enabled, OPT_BOOL, false) // whether swift object ve
 
 OPTION(rgw_list_bucket_min_readahead, OPT_INT, 1000) // minimum number of entries to read from rados for bucket listing
 
+OPTION(rgw_rest_getusage_op_compat, OPT_BOOL, false) // dump description of total stats for s3 GetUsage API
+
 OPTION(mutex_perf_counter, OPT_BOOL, false) // enable/disable mutex perf counter
 OPTION(throttler_perf_counter, OPT_BOOL, true) // enable/disable throttler perf counter
 
--- a/src/rgw/rgw_rest_s3.cc
+++ b/src/rgw/rgw_rest_s3.cc
@@ -430,10 +430,20 @@ void RGWGetUsage_ObjStore_S3::send_response()
        formatter->dump_int("SuccessfulOps", total_usage.successful_ops);
        formatter->close_section(); // total 
        formatter->close_section(); // user
-     } 
+     }
+
+     if (s->cct->_conf->rgw_rest_getusage_op_compat) {
+       formatter->open_object_section("Stats"); 
+     }	
+       
      formatter->dump_int("TotalBytes", header.stats.total_bytes);
      formatter->dump_int("TotalBytesRounded", header.stats.total_bytes_rounded);
      formatter->dump_int("TotalEntries", header.stats.total_entries);
+     
+     if (s->cct->_conf->rgw_rest_getusage_op_compat) { 
+       formatter->close_section(); //Stats
+     }
+
      formatter->close_section(); // summary
    }
    formatter->close_section(); // usage


yehudasa:
@weiqiaomiao this is a public api (albeit rgw extension). Need to bring this change to the mailing list, people might be using it as is now.

yehudasa:
@chenji-kael please take a look at this one

chenji-kael:
@yehudasa @weiqiaomiao 
Thanks for your remind， I think it's just about json decode format issue, and does not much matter, and actually we use this api for production environment ,and change this may cause problems when we update ceph.

weiqiaomiao:
@yehudasa I send a email to the mailing list（the email titile is 'RGW:add description  for GET usage admin api response entity 'stats entry''), and there is no one reponse it. I think  there is no people using this api except chenji. 
@chenji-kael One of our customers planning use this api, and  it is comfused to them what's the figure mean when they decode the api reponse data, add a description is a simple but easy way to help them quickly understand it.How about to fix this when you convenience?

chenji-kael:
@weiqiaomiao 
please see this pr , maybe you need it
https://github.com/ceph/ceph/pull/8043

yehudasa:
@weiqiaomiao @chenji-kael I suggest that we switch to the new api, but make it configurable so that current users can fall back to the old api. What do you think?

weiqiaomiao:
@yehudasa  @chenji-kael  It is a good suggestion to me and  update the commit yet.

src/common/config_opts.h
```diff
@@ -1433,6 +1433,8 @@ OPTION(rgw_swift_versioning_enabled, OPT_BOOL, false) // whether swift object ve
 
 OPTION(rgw_list_bucket_min_readahead, OPT_INT, 1000) // minimum number of entries to read from rados for bucket listing
 
+OPTION(rgw_getusage_dump_description, OPT_BOOL, true) // dump description of total stats for s3 GetUsage API
```
yehudasa: @weiqiaomiao let's rename it to rgw_rest_getusage_op_compat, and default it to false (switch the flag's meaning)

src/rgw/rgw_rest_s3.cc
```diff
@@ -429,10 +429,20 @@ void RGWGetUsage_ObjStore_S3::send_response()
...
        formatter->close_section(); // total 
        formatter->close_section(); // user
-     } 
-     formatter->dump_int("TotalBytes", header.stats.total_bytes);
-     formatter->dump_int("TotalBytesRounded", header.stats.total_bytes_rounded);
-     formatter->dump_int("TotalEntries", header.stats.total_entries);
+     }
+
+     if (s->cct->_conf->rgw_getusage_dump_description) {
+       formatter->open_object_section("Stats");
```
yehudasa: @weiqiaomiao close if here, and reopen it for the close_section(), so that we don't duplicate the stats dump

weiqiaomiao:
@yehudasa  Thanks for your suggestion, and the commit has been updated. Would you take a look at it again?

yehudasa:
@weiqiaomiao this looks ok. Last thing we need is that since this affects compatibility, we need to note it in the relase notes. Please add a description of this change, the new configurable (to override), and the api change to ceph/PendingReleaseNotes. Thanks.

weiqiaomiao:
@yehudasa  I already add this change to ceph/PendingReleaseNote. Would you take a look at it again.Thanks.

PendingReleaseNotes
```diff
@@ -17,6 +17,24 @@
   listed in 'osd class default list' requires a capability naming the class
   (e.g. 'allow class foo').
 
+* The 'rgw_rest_getusage_op_compat' config option allow you to dump the description of user stats
```
theanalyst: config option `allows`

PendingReleaseNotes
```diff
@@ -17,6 +17,24 @@
...
+  or not in s3 GetUsage API.This config is bool type and its default value is false.
+  If the value is true, the reponse data of s3 GetUsage like bellow:
```
theanalyst: `below`

PendingReleaseNotes
```diff
@@ -17,6 +17,24 @@
...
+
+    "stats": {
+                "TotalBytes": 516,
+                "TotalBytesRounded": 1024,
+                "TotalEntries": 1
+            }
+
+  Or if the value is false, the  reponse of s3 GetUsage like bellow and as it do before we add this config:
```
theanalyst: s/bellow/below,  and `as it do` -> `as it does`

theanalyst: also preferably, `before we add this config option`

PendingReleaseNotes
```diff
@@ -17,6 +17,24 @@
   listed in 'osd class default list' requires a capability naming the class
   (e.g. 'allow class foo').
 
+* The 'rgw_rest_getusage_op_compat' config option allow you to dump the description of user stats
+  or not in s3 GetUsage API.This config is bool type and its default value is false.
```
theanalyst: space after period, and `this config is` -> `this config option is of` , also `its default value is false` -> `defaults to false`

weiqiaomiao:
@theanalyst the commit has been updated according to your suggestion. would you take a look at it again. thanks!

PendingReleaseNotes
yehudasa: @weiqiaomiao also, no underscores needed when specifying the config option

yehudasa:
@weiqiaomiao please note the two comments on the last commit

weiqiaomiao:
@yehudasa Thank you for reminding me. The commit has been updated.

yehudasa:
Created and merged another PR due to conflict, see https://github.com/ceph/ceph/pull/11325

yehudasa:
@weiqiaomiao also updated commit message, noted tracker issue

PR #9498, opened 2016-06-04
bluestore: fix is_allocated() method of bluestore_blob_t

xiexingguo:
Seems we forget to increase the iterator properly.

### DIFF
--- a/src/os/bluestore/bluestore_types.cc
+++ b/src/os/bluestore/bluestore_types.cc
@@ -104,9 +104,7 @@ string bluestore_extent_t::get_flags_string(unsigned flags)
 {
   string s;
   if (flags & FLAG_SHARED) {
-    if (s.length())
-      s += '+';
-    s += "shared";
+    s = "shared";
   }
   return s;
 }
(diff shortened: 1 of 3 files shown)


xiexingguo:
@liewegas Ping.

src/os/bluestore/bluestore_types.cc
```diff
@@ -687,7 +687,7 @@ int bluestore_blob_t::verify_csum(uint64_t b_off, const bufferlist& bl) const
     return Checksummer::verify<Checksummer::crc32c>(
       get_csum_block_size(), b_off, bl.length(), bl, csum_data);
   default:
-    return -EOPNOTSUPP;
+    return EOPNOTSUPP;
```
liewegas: Hmm, I think what we actually want then is an output argumetn for offset instead of using the return value.  Otherwise we can't distinguish between a bad csum alg and a csum mismatch.

liewegas:
Looks fine except for the csum error; do you mind updating that one?

liewegas:
I merged the other 3 patches; can you fix the csum one and open a new PR on top of master?  Thanks!

PR #9502, opened 2016-06-05
common/buffer: add noexcept to ensure move ctor is used

tchaikov:
otherwise vector::push_back() will use the copy ctor if it resizes,
to enforce its strong exception guarantee.

### DIFF
--- a/src/common/buffer.cc
+++ b/src/common/buffer.cc
@@ -776,7 +776,7 @@ static simple_spinlock_t buffer_debug_lock = SIMPLE_SPINLOCK_INITIALIZER;
       bdout << "ptr " << this << " get " << _raw << bendl;
     }
   }
-  buffer::ptr::ptr(ptr&& p) : _raw(p._raw), _off(p._off), _len(p._len)
+  buffer::ptr::ptr(ptr&& p) noexcept : _raw(p._raw), _off(p._off), _len(p._len)
   {
     p._raw = nullptr;
     p._off = p._len = 0;
@@ -806,7 +806,7 @@ static simple_spinlock_t buffer_debug_lock = SIMPLE_SPINLOCK_INITIALIZER;
     }
     return *this;
   }
-  buffer::ptr& buffer::ptr::operator= (ptr&& p)
+  buffer::ptr& buffer::ptr::operator= (ptr&& p) noexcept
   {
     release();
     buffer::raw *raw = p._raw;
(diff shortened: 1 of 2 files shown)


src/common/buffer.cc
```diff
@@ -1350,7 +1350,7 @@ static simple_spinlock_t buffer_debug_lock = SIMPLE_SPINLOCK_INITIALIZER;
 
   // -- buffer::list --
 
-  buffer::list::list(list&& other)
+  buffer::list::list(list&& other) noexcept
     : _buffers(std::move(other._buffers)),
```
stiopaa1: @tchaikov 
I think std::list move constructor is not marked noexcept, so I was wondering if it's safe to be marking buffer::list::list(list&& other) noexcept ?

tchaikov: @stiopaa1 yeah, the default `std::allocator`'s ctor is not `noexcept`, and the list's sentry node could be allocated on heap. 

agree with you that it's safe to make sure that every line of the function should be `noexcept` if the function is `noexcept`. but i think the `noexcept` also defines an "exception boundary".

what if we can/do not handle the `bad_alloc` exception? can we add the `noexcept` specifier to the function which throws `bad_alloc` for better performance?

`buffer::list` is both a building block of ceph, it's also a part of librados library. so it is not acceptable for it to make a decision about killing the program on OOM. so as you suggested, i revert this change. but i think we can have it in other part of our project. for example, in #9361.

tchaikov:
tested at 

@cbodley could you take a look?

src/common/buffer.cc
cbodley: > agree with you that it's safe to make sure that every line of the function should be noexcept if the function is noexcept. but i think the noexcept also defines an "exception boundary".
> 
> what if we can/do not handle the bad_alloc exception? can we add the noexcept specifier to the function which throws bad_alloc for better performance?

it had been my understanding that noexcept actually turned off the bookkeeping that supports exception handling and that any exception thrown would lead to std::terminate, but it sounds like your interpretation is closer to the [spec](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2012/n3337.pdf):

> [15.4.9] Whenever an exception is thrown and the search for a handler (15.3) encounters the outermost block of a function with an _exception-specification_ that does not allow the exception, then [...] the function `std::terminate()` is called (15.5.1).

[cppreference](http://en.cppreference.com/w/cpp/language/noexcept_spec) rephrases it:

> If a search for a matching exception handler leaves a function marked `noexcept` or `noexcept(true)`, `std::terminate` is called immediately.

cbodley:
@tchaikov you can add my Reviewed-by. are you saying the failed test is unrelated, or did you want me to look at that too?

tchaikov:
> are you saying the failed test is unrelated

yeah, the failed test is unrelated.

thanks for your review @cbodley!

PR #9504, opened 2016-06-05
Unittesting: freebsd changes, part 3

wjwithagen:
This requires Clang to be at least 3.7
It compiles all ceph code (expect the ones excluded in autogen_freebsd.sh)
It compiles all tests

Tests will run, but not all complete successfully,
  Currently the only one failing is osd-markup.sh

Other tests have been execluded from testing, see README.FreeBSD for that

### DIFF
--- a/qa/workunits/ceph-helpers.sh
+++ b/qa/workunits/ceph-helpers.sh
@@ -117,7 +117,8 @@ function test_setup() {
 function teardown() {
     local dir=$1
     kill_daemons $dir KILL
-    if [ $(stat -f -c '%T' .) == "btrfs" ]; then
+    if [ `uname` != FreeBSD ] \
+        && [ $(stat -f -c '%T' .) == "btrfs" ]; then
         __teardown_btrfs $dir
     fi
     rm -fr $dir
@@ -179,7 +180,7 @@ function kill_daemon() {
             exit_code=0
             break
          fi
-         send_signal=0
+         [ x`uname`x = xFreeBSDx ] || send_signal=0
          sleep $try
     done;
     return $exit_code
@@ -607,7 +608,7 @@ function activate_osd() {
         --mark-init=none \
         $osd_data || return 1
 
-    [ "$id" = "$(cat $osd_data/whoami)" ] || return 1
+    [ "$id" -eq "$(cat $osd_data/whoami)" ] || return 1
 
     ceph osd crush create-or-move "$id" 1 root=default host=localhost
 
@@ -1286,8 +1287,14 @@ function erasure_code_plugin_exists() {
     local plugin=$1
 
     local status
+    local grepstr
+    case `uname` in
+        FreeBSD) grepstr="Cannot open.*$plugin" ;;
+        *) grepstr="$plugin.*No such file" ;;
+    esac
+
     if ceph osd erasure-code-profile set TESTPROFILE plugin=$plugin 2>&1 |
-        grep "$plugin.*No such file" ; then
+        grep "$grepstr" ; then
         status=1
     else
         status=0
@@ -1458,7 +1465,6 @@ function main() {
         code=1
     fi
     teardown $dir || return 1
-    return $code
 }
 
 #######################################################################
--- a/src/ceph-disk/ceph_disk/main.py
+++ b/src/ceph-disk/ceph_disk/main.py
@@ -743,7 +743,11 @@ def is_mounted(dev):
     Check if the given device is mounted.
     """
     dev = os.path.realpath(dev)
-    with file('/proc/mounts', 'rb') as proc_mounts:
+    proc_prefix = ''
+    if platform.system() == "FreeBSD":
+        # Use the Linux compatibility tree
+        proc_prefix = '/compat/linux'
+    with file(proc_prefix + '/proc/mounts', 'rb') as proc_mounts:
         for line in proc_mounts:
             fields = line.split()
             if len(fields) < 3:
--- a/src/ceph-disk/tests/test_main.py
+++ b/src/ceph-disk/tests/test_main.py
@@ -14,6 +14,7 @@
 #
 from mock import patch, DEFAULT
 import os
+import platform
 import io
 import shutil
 import subprocess
@@ -310,6 +311,10 @@ class TestCephDisk(object):
             assert expect == main.list_devices()
 
     def test_list_dmcrypt_data(self):
+        if platform.system() == "FreeBSD":
+            # FreeBSD does not have blkid, and the semantics that go with it
+            print "SKIPPED: no blkid"
+            return
         partition_type2type = {
             main.PTYPE['plain']['osd']['ready']: 'plain',
             main.PTYPE['luks']['osd']['ready']: 'LUKS',
@@ -376,6 +381,10 @@ class TestCephDisk(object):
                 assert expect == main.list_devices()
 
     def test_list_multipath(self):
+        if platform.system() == "FreeBSD":
+            # FreeBSD does not have blkid, and the semantics that go with it
+            print "SKIPPED: no blkid"
+            return
         #
         # multipath data partition
         #
@@ -581,6 +590,10 @@ class TestCephDisk(object):
             assert expect == main.list_devices()
 
     def test_list_other(self):
+        if platform.system() == "FreeBSD":
+            # FreeBSD does not have the same type of devices
+            print "SKIPPED: FreeBSD devices have different properties"
+            return
         #
         # not swap, unknown fs type, not mounted, with uuid
         #
--- a/src/common/config_opts.h
+++ b/src/common/config_opts.h
@@ -1049,18 +1049,35 @@ OPTION(filestore_omap_header_cache_size, OPT_INT, 1024)
 OPTION(filestore_max_inline_xattr_size, OPT_U32, 0)	//Override
 OPTION(filestore_max_inline_xattr_size_xfs, OPT_U32, 65536)
 OPTION(filestore_max_inline_xattr_size_btrfs, OPT_U32, 2048)
+OPTION(filestore_max_inline_xattr_size_zfs, OPT_U32, 2048)
+#if defined(__FreeBSD__)
+// Assuming ZFS without using ZFS-FileStore
+OPTION(filestore_max_inline_xattr_size_other, OPT_U32, 2048)
+#else
 OPTION(filestore_max_inline_xattr_size_other, OPT_U32, 512)
+#endif
 
 // for more than filestore_max_inline_xattrs attrs
 OPTION(filestore_max_inline_xattrs, OPT_U32, 0)	//Override
 OPTION(filestore_max_inline_xattrs_xfs, OPT_U32, 10)
 OPTION(filestore_max_inline_xattrs_btrfs, OPT_U32, 10)
+OPTION(filestore_max_inline_xattrs_zfs, OPT_U32, 10)
+#if defined(__FreeBSD__)
+// Assuming running on ZFS without using ZFS-FileStore
+OPTION(filestore_max_inline_xattrs_other, OPT_U32, 10)
+#else
 OPTION(filestore_max_inline_xattrs_other, OPT_U32, 2)
+#endif
 
 // max xattr value size
 OPTION(filestore_max_xattr_value_size, OPT_U32, 0)	//Override
 OPTION(filestore_max_xattr_value_size_xfs, OPT_U32, 64<<10)
 OPTION(filestore_max_xattr_value_size_btrfs, OPT_U32, 64<<10)
+OPTION(filestore_max_xattr_value_size_zfs, OPT_U32, 64<<10)
+#if defined(__FreeBSD__)
+// Assuming ZFS without using ZFS-FileStore
+OPTION(filestore_max_xattr_value_size_other, OPT_U32, 64<<10)
+#else
 // ext4 allows 4k xattrs total including some smallish extra fields and the
 // keys.  We're allowing 2 512 inline attrs in addition some some filestore
 // replay attrs.  After accounting for those, we still need to fit up to
@@ -1068,6 +1085,7 @@ OPTION(filestore_max_xattr_value_size_btrfs, OPT_U32, 64<<10)
 // to be safe.  This is hacky, but it's not worth complicating the code
 // to work around ext4's total xattr limit.
 OPTION(filestore_max_xattr_value_size_other, OPT_U32, 1<<10)
+#endif
 
 OPTION(filestore_sloppy_crc, OPT_BOOL, false)         // track sloppy crcs
 OPTION(filestore_sloppy_crc_block_size, OPT_INT, 65536)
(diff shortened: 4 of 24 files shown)


src/ceph-disk/tests/ceph-disk.sh
```diff
@@ -54,7 +54,7 @@ CEPH_ARGS+=" --osd-failsafe-full-ratio=.99"
 CEPH_ARGS+=" --mon-host=$MONA"
 CEPH_ARGS+=" --log-file=$DIR/\$name.log"
 CEPH_ARGS+=" --pid-file=$DIR/\$name.pidfile"
-if test -d ../.libs ; then
+if [ -d ../.libs ]; then
```
liewegas: safer to stick with test, i think

wjwithagen:
@tchaikov @liewegas 
I think this is enough for now.
This gets testing to 117 of 126.

There are a few that are easy onliners.
But the scripting in ceph-helpers.sh might be a bit more challenging, although teardown_error is not used yet. It will be in the commits to come.

src/test/cli/ceph-authtool/cap-bin.t
```diff
@@ -2,5 +2,5 @@
   creating kring
 
   $ ceph-authtool --cap osd 'allow rx pool=swimming' kring
-  $ ceph-authtool kring --list|grep -P '^\tcaps '
+  $ ceph-authtool kring --list|grep -E '^\Wcaps '
```
tchaikov: i don't understand why you'd prefer "\W" over "\t"?

wjwithagen: @tchaikov 
-P is perl regex, and that is not in FreeBSD grep.
And `\t`is perl-speak.
So I try to stick to POSIX

Perhaps a bit of a braintwist at my end, but `\W` is abit more liberal in that it catches all types of white space. So it for whatever reason the output changes a bit, this will match that too.

wjwithagen:
@tchaikov 
Anything I can do, other than wait for a qa run?

tchaikov:
@wjwithagen i haven't got a chance to review all your changes. will look at them closer.

wjwithagen:
@tchaikov 
Oke, take your time. Working on my sheet for Ceph day Cern anyways....

wjwithagen:
@tchaikov 
Right, back in town...
So any suggestions are welcome

tchaikov:
@wjwthagen will review your changeset next week.

wjwithagen:
On 17-6-2016 12:41, Kefu Chai wrote:
> @wjwthagen will review your changeset next week.

Oke, no problem. Enjoy the weekend.

--WjW

wjwithagen:
@tchaikov 
Hi Kefu,
I needed to rebase to get conflicts resolved, but now I'm getting the following error, in  tests/ceph-disk.sh.
So I have removed those commit from this pull:

Any suggestions what to fix??
```
tests/ceph-disk.sh:107: command_fixture:  local dir=testdir/test-ceph-disk
tests/ceph-disk.sh:108: command_fixture:  shift
tests/ceph-disk.sh:109: command_fixture:  local command=ceph-conf
tests/ceph-disk.sh:110: command_fixture:  shift
tttests/ceph-disk.sh:111: command_fixture:  which ceph-conf
ttests/ceph-disk.sh:111: command_fixture:  readlink -f /usr/srcs/Ceph/work.wip-wjw-freebsd-tests-2/ceph/src/ceph-conf
tests/ceph-disk.sh:111: command_fixture:  local fpath=/usr/srcs/Ceph/work.wip-wjw-freebsd-tests-2/ceph/src/ceph-conf
ttests/ceph-disk.sh:112: command_fixture:  readlink -f /usr/srcs/Ceph/work.wip-wjw-freebsd-tests-2/ceph/src/ceph-disk/ceph-conf
tests/ceph-disk.sh:112: command_fixture:  '[' x/usr/srcs/Ceph/work.wip-wjw-freebsd-tests-2/ceph/src/ceph-confx = xx ']'
tttests/ceph-disk.sh:112: command_fixture:  pwd
ttests/ceph-disk.sh:112: command_fixture:  readlink -f /usr/srcs/Ceph/work.wip-wjw-freebsd-tests-2/ceph/src/ceph-disk/ceph-conf
tests/ceph-disk.sh:112: command_fixture:  '[' x/usr/srcs/Ceph/work.wip-wjw-freebsd-tests-2/ceph/src/ceph-confx = xx ']'
tests/ceph-disk.sh:112: command_fixture:  return 1

```

qa/workunits/ceph-helpers.sh
```diff
@@ -1452,13 +1472,13 @@ function main() {
...
-        code=0
+        teardown $dir || return 1
+        return 0
     else
         display_logs $dir
-        code=1
+        teardown_error $dir
+        return 1
     fi
-    teardown $dir || return 1
```
tchaikov: @wjwithagen can we leave this change for another changeset? this changes the behaviour of current tests.

qa/workunits/ceph-helpers.sh
```diff
@@ -513,6 +524,9 @@ function destroy_osd() {
     local dir=$1
     local id=$2
 
+    # First mask osd down, otherwise it cannot be deleted later on
+    # is status has not propagated to monitor.
```
tchaikov: could you elaborate this a little bit? why we need to mark this osd down explicitly? and if we really need to do this. maybe you can move it down to line 531. just for consistency.

qa/workunits/ceph-helpers.sh
```diff
@@ -1286,8 +1300,14 @@ function erasure_code_plugin_exists() {
     local plugin=$1
 
     local status
+    local grepstr
+    case x`uname`x in
```
tchaikov: no need to quote `uname` with "x". it is never empty in a sane OS.

qa/workunits/ceph-helpers.sh
```diff
@@ -1286,8 +1300,14 @@ function erasure_code_plugin_exists() {
...
+        xLinuxx) grepstr="$plugin.*No such file" ;;
+        xFreeBSDx) grepstr="Cannot open.*$plugin" ;;
```
tchaikov: maybe we can have a
```
*)
```
to catch all other cases?

src/ceph-disk/tests/test_main.py
```diff
@@ -310,6 +311,10 @@ def get_partition_type(dev):
             assert expect == main.list_devices()
 
     def test_list_dmcrypt_data(self):
+        if platform.system() == "FreeBSD":
```
tchaikov: @dachary does this change look good to you?

src/ceph-disk/tests/test_main.py
```diff
@@ -14,6 +14,7 @@
 #
 from mock import patch, DEFAULT
 import os
+import platform
```
tchaikov: @wjwithagen nit, the commit message could read "FreeBSD does not have blkid"

src/vstart.sh
```diff
@@ -92,7 +92,11 @@ start_rgw=0
 ip=""
 nodaemon=0
 smallmds=0
-short=0
+if [ x`uname`x = xFreeBSDx ] ; then
```
tchaikov: no need to quote `uname` with "x".

src/global/pidfile.cc
```diff
@@ -162,7 +162,12 @@ int pidfh::open(const md_config_t *conf)
   pf_dev = st.st_dev;
   pf_ino = st.st_ino;
 
-  struct flock l = { F_WRLCK, SEEK_SET, 0, 0, 0 };
+  struct flock l;
+  memset(&l, 0, sizeof(l));
+  l.l_type = F_WRLCK;
+  l.l_whence = SEEK_SET;
+  l.l_start = 0;
+  l.l_len = 0;
```
tchaikov: could put

```c++
  struct flock l = {
    .l_type = F_WRLCK,
    .l_whence = SEEK_SET,
    .l_start = 0,
    .l_len = 0
  };
```
instead.

src/vstart.sh
tchaikov: but zfs has longer xattr limits, i assume?

src/ceph-disk/ceph_disk/main.py
```diff
@@ -743,7 +743,11 @@ def is_mounted(dev):
     Check if the given device is mounted.
     """
     dev = os.path.realpath(dev)
-    with file('/proc/mounts', 'rb') as proc_mounts:
+    proc_prefix = ''
+    if platform.system() == "FreeBSD":
+        # Use the Linux compatability tree
```
tchaikov: s/compatability/compatibility/

src/common/config_opts.h
```diff
@@ -1078,7 +1096,11 @@ OPTION(filestore_max_sync_interval, OPT_DOUBLE, 5)    // seconds
 OPTION(filestore_min_sync_interval, OPT_DOUBLE, .01)  // seconds
 OPTION(filestore_btrfs_snap, OPT_BOOL, true)
 OPTION(filestore_btrfs_clone_range, OPT_BOOL, true)
+#if defined(__linux__)
```
tchaikov: maybe we can enable `filestore_zfs_snap` by default? as zfs is included by some major linux distros recently. @ukernel what do you think?

src/common/xattr.h
```diff
@@ -19,12 +19,14 @@
 extern "C" {
 #endif
 
+#if defined(__linux__)
```
tchaikov: i am fine with this change, but it's always good to have `ENOATTR` defined, right? or we need to check if `ENODATA` is defined also here? to avoid cyclic definition? as we have
```
#define ENODATA ENOATTR
```
in `compat.h`.

[OT] if `xattr.c` is the only consumer of `ENOATTR`, maybe we can move this definition into `xattr.c` in another commit.

qa/workunits/ceph-helpers.sh
wjwithagen: I only added the ` [ x`uname`x = xFreeBSDx ] || send_signal=0`.
The part about the OSD maks is not mine. Perhaps a left over from rebasing earlier on.
I will remove that piece of comments.

But this is the wrong place... __teardown_btrfs will not be called on a FreeBSD platform.

qa/workunits/ceph-helpers.sh
wjwithagen: Sane would be relative in this crazy world. 8D
I've seen too many things come up empty... The last one being readlink(1).
But you are right chances are slim.

And I'll put in a wildcard as default catch

qa/workunits/ceph-helpers.sh
wjwithagen: @tchaikov 
It is not my intention to change any of the behaviour of the tests.
Other than:
  - Fix FreeBSD issues
  - Keep data around when something fails, so one can look at the logfiles
So if it does change standard behaviour on Linux, when tests go as planned, that should be a mistake.

But I'm fine with taking this change to a next set of commits for the next pull.
Note that I talked about it with a few people at Ceph day, and it would be nice if tests also start taking care of situations where things actually fail, and people doing testing/debugging/porting would have easy access to more information.

src/ceph-disk/tests/test_main.py
wjwithagen: Will fix

src/ceph-disk/tests/test_main.py
wjwithagen: @tchaikov @dachary 
FreeBSD does not have any of the dm stuff at all. The short cut was to disable here.
As I did at a few other places

src/common/config_opts.h
wjwithagen: @tchaikov @ukernel 
I have not done so, because I would liek to first see that things are working with ZFS as expected.
And then actually start integrating it fully.
Then the options would also become real zfs options instead of:
    `filestore_*_xattr_*_other`
For now I'm just overulling the default on FreeBSD, otherwise there was going to be another lage set of patches for that. But I'll do that in a set of dedicated ZFS patches.

src/common/xattr.h
wjwithagen: The definition was already there, I just wrapped it in ifdef because might case redefinition issues, and compiler warnings... 
But I think ENOATTR is not used in ANY of the code other than compat.h.
And there it is to create compatibility with FreeBSD.
So yes, I think this whole piece could/should go.??

src/global/pidfile.cc
wjwithagen: Oke, that is the nicer style. Are all not mentioned values than set to 0?
Because that is what the memset() does, althoug I suspect that that is not required for the call we're doing

src/vstart.sh
wjwithagen: Fixed the xx's
And yes ZFS has long attributes, but Sage commented during Ceph day that we should use --short for some of the tests. And this is just to make sure that we do that when running FreeBSD, because is is known to sometimes fail the test in ceph-osd. This prevents that

qa/workunits/ceph-helpers.sh
```diff
@@ -17,6 +17,15 @@
 # MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 # GNU Library Public License for more details.
 #
+
+# to be able to run standalone define some VARIABLES
```
tchaikov: this is not related FreeBSD port. could you remove it from this commit?

qa/workunits/ceph-helpers.sh
```diff
@@ -123,6 +132,16 @@ function teardown() {
     rm -fr $dir
 }
 
+function teardown_error() {
```
tchaikov: this function is not called, could you please remove it also?

src/common/xattr.h
tchaikov: i am inclined to keep it as it is, unless there is case that we see redefinition issues or warnings. why we could have redefinition issues because of this `#define`?

i don't think the whole piece should go. see 886f5a91f35d04cd4a721f1a66a2ef4a2c39fd5e

wjwithagen: @tchaikov 
I'll put it back in.

But in that case I would be very mich in favour of reversing the whole deal...
Rip out all references to ENODATA and actually start using ENOATTR every where.
Under Linux we then can declare ENODATA = ENOATTR.
But then again, that is perhaps for another set of commits.

wjwithagen: @tchaikov 
(forget my previous message)

I looked at it again, and it is foolish to put in for anything else than Linux.
If this case would arise under FreeBSD (or other  OS)  I'd rather have it signal an error because at this point we are very likely going to patch up ENOATTR with the wrong value.
So rather than doing that, I is better for the compiler to tell that ENOATTR is not defined. 
The way to fix that is to include include/compat.h.

If you prefer I can add a compiler `#WARNING` to that extend in the `#else` clause.

qa/workunits/ceph-helpers.sh
wjwithagen: @tchaikov 
This is in preparation that we start doing some sensible things when a test actually fails. We then do need the poor remains of the carcass to see what happened. But that can go into another commit/pull as well.

PR #9505, opened 2016-06-06
mon: remove duplicate help information

x11562:
Signed-off-by: Na Xie <xie.na@h3c.com>


xiexingguo:
Actually these two commands are slightly different, and are auth print-(**hyphen**)key and auth print_(**underscore**)key respectively.

That's why jenkins is unhappy with this change.

PR #8890, opened 2016-05-03
client: enable fuse_use_invalidate_cb by default

ukernel:
Fixes: http://tracker.ceph.com/issues/15634
Signed-off-by: Yan, Zheng <zyan@redhat.com>

PR #9509, opened 2016-06-06
client: skip executing async invalidates while umounting

ukernel:
Fixes: http://tracker.ceph.com/issues/16137

### DIFF
--- a/src/client/Client.cc
+++ b/src/client/Client.cc
@@ -3500,31 +3500,29 @@ void Client::wake_inode_waiters(MetaSession *s)
 class C_Client_CacheInvalidate : public Context  {
 private:
   Client *client;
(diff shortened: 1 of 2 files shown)


jcsp:
Reviewed-by, will re-test with https://github.com/ceph/ceph/pull/8890

jcsp:
No client failures in test run http://pulpito.ceph.com/jspray-2016-06-08_02:34:44-fs-wip-jcsp-testing-20160606---basic-mira/

batrick:
Am I correct that this commit just avoids the problem of root becoming invalid? This looks related to #10027. Unfortunately the client logs from the [broken tests](http://tracker.ceph.com/issues/16137) are not available to confirm.

PR #9510, opened 2016-06-06
common: unknown hash type of judgment modification

renhwztetecs:
ceph_str_hash() return -1 if set unknown hash type
but (unsiged)-1 = unsiged 4294967295 
So we can't judge whether the 4294967295 is the normal value of hash, or the return value of the unknown hash type
use assert replace

### DIFF
--- a/src/client/Inode.h
+++ b/src/client/Inode.h
@@ -212,6 +212,7 @@ struct Inode {
     int which = dir_layout.dl_dir_hash;
     if (!which)
       which = CEPH_STR_HASH_LINUX;
+    assert(ceph_str_hash_valid(which));
     return ceph_str_hash(which, dn.data(), dn.length());
   }
 
--- a/src/common/ceph_hash.cc
+++ b/src/common/ceph_hash.cc
@@ -115,3 +115,14 @@ const char *ceph_str_hash_name(int type)
 		return "unknown";
 	}
 }
+
+bool ceph_str_hash_valid(int type)
+{
+        switch (type) {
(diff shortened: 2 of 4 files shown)


liewegas:
I'm worried about cases that have invalid hash function ids. Can you add another function, ceph_str_hash_valid that returns true/false, and fix the callers that are feeding in variables (vs constants) to verify they have a valid function id?  I don't want to introduce a way to crash the system by injecting bad data.

renhwztetecs:
@ liewegas 
thanks for you reviews
I add  ceph_str_hash_valid() to verify hash type, to judge before calling the unknown type hash.
for the type of CEPH_STR_HASH_LINUX/CEPH_STR_HASH_RJENKINS that has been identified is called, it is not to be judged.

but I think is better before commit?  :)

src/client/Inode.h
```diff
@@ -212,6 +212,7 @@ struct Inode {
     int which = dir_layout.dl_dir_hash;
     if (!which)
       which = CEPH_STR_HASH_LINUX;
+    assert(true == ceph_str_hash_valid(which));
```
xiexingguo: `assert(ceph_str_hash_valid(which));` will suffice

renhwztetecs:
@xiexingguo 
done thanks

liewegas:
lgtm, thanks!

yuriw:
http://pulpito.ceph.com/yuriw-2016-07-12_13:07:42-rados-wip-yuri-testing_2016_7_12-distro-basic-smithi

PR #9511, opened 2016-06-06
rgw: update user stat before get it for Get User Info api

weiqiaomiao:
update user stat before  we get it  to retrieve the latest quota stats for Get User Info api


yehudasa:
@weiqiaomiao I'm not sure this is a good idea. Syncing the stats might take a long time, more  than clients expect. Possibly more than the operation timeout expectation.

weiqiaomiao:
@yehudasa  OH~~I did't consider the performance issues before. Thanks for remind me that.

PR #9513, opened 2016-06-06
mon: Forbidden copy and assignment function in monoprequest

songbaisen:
mon: Forbidden copy and assignment function in monoprequest

### DIFF
--- a/src/mon/MonOpRequest.h
+++ b/src/mon/MonOpRequest.h
@@ -74,6 +74,9 @@ struct MonOpRequest : public TrackedOp {
     OP_TYPE_COMMAND,          ///< is a command
   };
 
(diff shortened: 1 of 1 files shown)


liewegas:
The c++11 way to do this is to add '= delete' to the declaration so that you get a compile time (vs link time) error.

songbaisen:
@liewegas  Learn it. Thank you!

dachary:

dachary:
please reword the commit message:
<pre>
Because the class have point as its member,we would better forbid the copy and assign if we do not want this class to do so for safe.
</pre>
is not proper English.  Also make sure you update the commit message and not the github message in the web interface.

songbaisen:
@dachary Thank you for review.Done it.

PR #9514, opened 2016-06-06
client:calculation pg use the correct method 

songbaisen:
client:calculation pg use the correct method 

If we want to know the real up primary pg for the osds .We should use the 

pg_to_raw_up method.If not I do not know why here use the  pg_to_osds method.

Thank you!


jcsp:
I'm not sure this is correct.  When someone uses this function they are probably thinking about where to place e.g. a compute resource to be near the storage.  In that case we should probably tell them where we expect the PG to eventually be, rather than where it currently is (e.g. during backfill)

liewegas:
I think this should actually be pg_to_up_acting_osds()

liewegas:
https://github.com/ceph/ceph/pull/9523

PR #9515, opened 2016-06-06
default to async messenger, wpq

liewegas:


gregsfortytwo:
Hmm, how will this impact our nightlies?

If we want to maintain mixed async/simple clusters, we probably need to swap them. If we don't, I think we might want to change the defaults in our nightlies for a while before changing them in the Ceph source?

ytovy:
hi, Sage, we encoutered a problem a month ago,  haven't resolved yet, we really need your help. 
We build a cluster with 5 hosts, 120 osd, and made a ec pool with k=6, m=1, 2800pgs. ceph version 10.2.0.
We set ms_type = async, and let it go. Making pool is ok, but once we restart the cluster that make some pgs' status stucked in 'peering' or 'remap+peering' and the cluster's health is err. We waited whole day but it couldn't return to health_ok. 
Then we rebuild cluster with 'simple' messenger, it didn't happen, and all other settings are the same with original settings.
So is it bug of async messenger? or any other thing cause it?

liewegas:
The rados suite is covering simplemessenger explicitly.  The other suites will flip to doing just async with this change.

I'm not sure if we want to do ms type = async, random, or to double the test size for them...

PR #9518, opened 2016-06-06
mds: fix MDRequestImpl::print override

cbodley:
Fixes compiler warnings:
```
In file included from /home/cbodley/ceph/src/mds/Locker.cc:22:0:
/home/cbodley/ceph/src/mds/Mutation.h:161:16: warning: ‘virtual void
MutationImpl::print(std::ostream&) const’ was hidden
[-Woverloaded-virtual]
   virtual void print(ostream &out) const {
                ^
/home/cbodley/ceph/src/mds/Mutation.h:340:8: warning:   by ‘void
MDRequestImpl::print(std::ostream&)’ [-Woverloaded-virtual]
   void print(ostream &out);
```

### DIFF
--- a/src/mds/Mutation.cc
+++ b/src/mds/Mutation.cc
@@ -303,7 +303,7 @@ void MDRequestImpl::set_filepath2(const filepath& fp)
   more()->filepath2 = fp;
 }
 
-void MDRequestImpl::print(ostream &out)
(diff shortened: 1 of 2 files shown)


jcsp:
LGTM, merge when jenkins build passes

PR #9523, opened 2016-06-06
osd/OSDMap: pg_to_osds -> pg_to_raw_osds, fix libcephfs caller to get acting osds

liewegas:


gregsfortytwo:
Reviewed-by:

songbaisen:
en, lgtm.

src/client/Client.cc
```diff
@@ -11061,7 +11061,7 @@ int Client::ll_get_stripe_osd(Inode *in, uint64_t blockno,
       pg_t pg = (pg_t)olayout.ol_pgid;
       vector<int> osds;
       int primary;
-      o.pg_to_osds(pg, &osds, &primary);
+      o.pg_to_up_acting_osds(pg, NULL, NULL, &osds, &primary);
```
xiexingguo: Why don't use `pg_to_acting_osds` directly here? Since the up set is irrelevant here.

liewegas: updated, althouh it is just a wrapper around pg_to_up_acting_osds.  also 
fixed to return primary instead of osds[0]

xiexingguo:
lgtm also

PR #9524, opened 2016-06-06
common: add cbits/ctz/clz bit op helpers, replace open-coded calc_bits_for helpers

liewegas:


src/osd/osd_types.cc
```diff
@@ -464,7 +464,7 @@ ostream& operator<<(ostream& out, const spg_t &pg)
 
 pg_t pg_t::get_ancestor(unsigned old_pg_num) const
 {
-  int old_bits = pg_pool_t::calc_bits_of(old_pg_num);
+  int old_bits = cbitsl(old_pg_num);
```
athanatos: I havnen't been following closely, but cbitsl is not cbits.

athanatos: Intentional or harmless typo?

athanatos:
Except for the harmless typo, lgtm!

PR #9531, opened 2016-06-06
rgw: data sync debug logging

yehudasa:
a high level debug logging of the data sync process.


yehudasa:
@cbodley this is a first step towards easier to understand sync logging. Currently only covers data log, but I think it's much better than what we have.

PR #9532, opened 2016-06-07
os/bluestore: make verify_csum catch unsupported csum type error

xiexingguo:
And update related unit-testing cases.

### DIFF
--- a/src/os/bluestore/bluestore_types.cc
+++ b/src/os/bluestore/bluestore_types.cc
@@ -668,23 +668,38 @@ void bluestore_blob_t::calc_csum(uint64_t b_off, const bufferlist& bl)
   }
 }
 
-int bluestore_blob_t::verify_csum(uint64_t b_off, const bufferlist& bl) const
+int bluestore_blob_t::verify_csum(uint64_t b_off, const bufferlist& bl,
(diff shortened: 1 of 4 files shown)


xiexingguo:
@liewegas Upgrade https://github.com/ceph/ceph/pull/9498/commits/ff27e14c160397a1f3bdd06255b3d75ebf10726a here.

src/os/bluestore/bluestore_types.cc
```diff
@@ -668,23 +668,38 @@ void bluestore_blob_t::calc_csum(uint64_t b_off, const bufferlist& bl)
   }
 }
 
-int bluestore_blob_t::verify_csum(uint64_t b_off, const bufferlist& bl) const
+int bluestore_blob_t::verify_csum(uint64_t b_off, const bufferlist& bl,
+  int &b_bad_off) const
```
liewegas: style nit: can you make this int* instead of int& to indicate it is an output argument?

liewegas:
Also, please combine the unittest update in the same commit as the change.

Thanks!

liewegas:
Otherwise it looks good!

xiexingguo:
Updated as you suggested.

PR #9533, opened 2016-06-07
os/bluestore: make rocksdb not to use page cache

jjhuo:
This is second try. Previous PR crashed, because rocksdb does random unaligned read.
I rewrote read_buffered function to read_random, which uses an internal aligned buffer to perform direct read when needed.
Will use a worse case scenario(bluestore_block_buffered_read = false) to test if rocksdb with block cache can perform as with page cache.

### DIFF
--- a/src/common/config_opts.h
+++ b/src/common/config_opts.h
@@ -842,7 +842,8 @@ OPTION(kinetic_use_ssl, OPT_BOOL, false) // whether to secure kinetic traffic wi
 OPTION(rocksdb_separate_wal_dir, OPT_BOOL, false) // use $path.wal for wal
 OPTION(rocksdb_db_paths, OPT_STR, "")   // path,size( path,size)*
 OPTION(rocksdb_log_to_ceph_log, OPT_BOOL, true)  // log to ceph log
-OPTION(rocksdb_cache_size, OPT_INT, 128*1024*1024)  // default leveldb cache size
+OPTION(rocksdb_cache_size, OPT_INT, 128*1024*1024)  // default rocksdb cache size
+OPTION(rocksdb_cache_shard_bits, OPT_INT, 4)  // rocksdb block cache shard bits, 4 bit -> 16 shards
 OPTION(rocksdb_block_size, OPT_INT, 4*1024)  // default rocksdb block size
 // rocksdb options that will be used for omap(if omap_backend is rocksdb)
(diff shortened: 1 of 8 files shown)


jjhuo:
For 5 hours 4K random fio read workload, this new PR gives average 10.4K IOPS, while baseline is 10.3K IOPS. So, this PR has 1% improvement compared to baseline, and IOPS over 5 hours are more stable than baseline.
Both baseline and this PR uses jewel 10.2.1 for this benchmarking, bluestore_block_buffered_read = false, so rocksdb in baseline won't compete for page cache with others, this is kind of best case for baseline I guess; For this PR benchmarking, rocksdb block_cache is set as 512MB, no page cache used.
And this is one OSD on one SSD.
Will do write workload benchmarking as well.

jjhuo:
@liewegas For 5 hours 4K random writes, the numbers are not good, IOPS with rocksdb block cache dropped by half compared to baseline with kernel page cache. After looking into rocksdb code, I found out rocksdb block cache only cache reads not writes. So during this 5 hours random 4K writes, the newly written sst files won't be cached in block cache, then later when compaction kicks in, compaction will have to read sst files from device(as opposed to reading from page cache for baseline case).

jjhuo:
@liewegas Hi Sage, do we need to revert this PR until we have writes working as well? I am evaluating if it's possible to implement an option to have rocksdb block cache to cache writes. Thanks.
And I also will do more benchmarking with bluestore_block_buffered_read = true.

jjhuo:
@liewegas Hi Sage, do we need to revert this PR until we have writes working as well? I am evaluating if it's possible to implement an option in rocksdb to have block cache to cache writes. 
And I also wondering, after bluestore object data uses it's own cache(which you are working on), does rocksdb need to move from page cache to its own block cache anymore? since it won't share kernel page cache with anybody else. Thanks!

markhpc:
Hi jjhuo,  I'm seeing perioids of extensive reads during 4K random writes presumably during compaction while benchmarking recent master.  Do you think this PR may be contributing?

jjhuo:
Hi @markhpc, I think it's not, since you used ramdisk for db+wal, this PR should make no impact on this setup. I agreed with other guys, it looks like compaction stall, in additional to multiple compaction/flusher threads, we should also increase rocksdb write_buffer_size as well, the default 4MB is too small.

PR #9536, opened 2016-06-07
client: fix sanity check of preadv/pwritev

xiexingguo:
The preadv/pwritev call is only supposed to return a postive
number indicating how many bytes we succeed in reading from
or writing into disk, which is obviously not what we want
under the following cases.


jcsp:
Looks good to me.  By the way, can you add the cephfs label when opening these PRs?

xiexingguo:
Sure:-)

jcsp:
http://pulpito.ceph.com/jspray-2016-06-08_02:34:44-fs-wip-jcsp-testing-20160606---basic-mira/

PR #9538, opened 2016-06-07
os/bluestore: Unaligned device length and Correct alloc size fix for BitMap allocator

chhabaramesh:
These two changes are:
1. Take care of situation when total blocks are not aligned to zone boundary.
    a. Pad extra blocks and mark them used.
    b. add unit test cases for the same.
2. Set correct min alloc size as a unit for BitMapAllocator allocation unit.  That is 1M for BlueFS and 4k or 64k for ssd and hdd.

It passes unit + ceph_test_objectstore

### DIFF
--- a/src/os/bluestore/BitAllocator.h
+++ b/src/os/bluestore/BitAllocator.h
@@ -17,6 +17,7 @@
 #include <mutex>
 #include <atomic>
 #include <vector>
+#include "include/intarith.h"
 
 class BitAllocatorStats {
 public:
--- a/src/test/objectstore/BitAllocator_test.cc
+++ b/src/test/objectstore/BitAllocator_test.cc
@@ -408,6 +408,22 @@ TEST(BitAllocator, test_bmap_alloc)
   alloc->free_blocks(0, alloc->size());
   delete alloc;
 
+  // unaligned zones
+  total_blocks = 1024 * 2 + 11;
+  alloc = new BitAllocator(total_blocks, zone_size, CONCURRENT);
+
+  for (int64_t iter = 0; iter < 4; iter++) {
+    for (int64_t i = 0; i < total_blocks; i++) {
+      allocated = alloc->alloc_blocks(1, &start_block);
+      bmap_test_assert(allocated == 1);
+      bmap_test_assert(start_block == i);
+    }
+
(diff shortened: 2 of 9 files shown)


chhabaramesh:

src/os/bluestore/BitAllocator.h
```diff
@@ -11,6 +11,8 @@
 #define debug_assert assert
 #define BITMAP_SPAN_SIZE (1024)
 
+#define NEXT_MULTIPLE(x, m) (!(x) ? 0: (((((x) - 1) / (m)) + 1)  * (m)))
```
liewegas: ROUND_UP_TO is already defined

liewegas:
The ctor argument is really expected_min_alloc_size, right?  This is just improving the situation where you carve up into zones but the zones don't align with the min_alloc_size?

I think in general we want to keep allocation at block_size (bdev's block size, 4k) granularity.  And min_alloc_size is currently an allocate() argument... and could in principle vary per allocation.

src/test/objectstore/BitAllocator_test.cc
```diff
@@ -408,6 +408,22 @@ TEST(BitAllocator, test_bmap_alloc)
   alloc->free_blocks(0, alloc->size());
   delete alloc;
 
+	// unaligned zones	
+	total_blocks = 1024 * 2 + 11;
```
liewegas: bad indentation

src/os/bluestore/BitAllocator.h
chhabaramesh: ok, I will use that.

src/test/objectstore/BitAllocator_test.cc
chhabaramesh: :(. will fix it.

chhabaramesh:
@liewegas  as discussed these are two changes:
1. Handle case when total blocks given to allocator during init are not making complete zones as last few blocks are less than zone.
2. Use bluefs alloc size in the bluefs instance of allocator so that we don't use extra mem by using 4k block size for bluefs instance of allocators.

The changes are good to merge now.

Please do the needful.

chhabaramesh:
@liewegas  please merge this .

PR #9547, opened 2016-06-07
jewel backport: mds: fix mdsmap print_summary with standby replays

jcsp:
http://tracker.ceph.com/issues/15968


dachary:
It passed the fs suite http://tracker.ceph.com/issues/16344#note-15.

PR #9549, opened 2016-06-07
os/bluestore: Fix for blob length on large writes

ifed01:


src/os/bluestore/BlueStore.cc
```diff
@@ -2549,7 +2549,18 @@ int BlueStore::fsck()
 	}
 	// lextents
 	map<int64_t,bluestore_extent_ref_map_t> local_blobs;
+	uint64_t lext_next_offset = 0, lext_prev_offset = 0;
 	for (auto& l : o->onode.extent_map) {
+	  if(l.first < lext_next_offset) {
```
liewegas: s/if(/if (/

PR #9551, opened 2016-06-07
cls_rbd: mirror image status summary should read full directory

dillaman:
Previously only retrieved the status for the first 64 images in
the rbd_mirroring directory.

Fixes: http://tracker.ceph.com/issues/16178


trociny:
lgtm

PR #9563, opened 2016-06-07
kv: explicitly set DISABLE_JEMALLOC for rocksdb in Makefile.am.

markhpc:
Upstream RocksDB changed the way they detect jemalloc which breaks our build:

https://github.com/ceph/rocksdb/commit/0850bc514737a64dc8ca13de8510fcad4756616a

We can explicitly set DISABLE_JEMALLOC depending on whether we are using jemalloc via the existing AM_CONDITIONAL value in configure.ac.  I was able to successfully build master with PR #9466 using tcmalloc or jemalloc on my test cluster after this change.

### DIFF
--- a/src/kv/Makefile.am
+++ b/src/kv/Makefile.am
@@ -25,8 +25,15 @@ else
 NPROC = nproc
 endif
 
+# explicitly disable jemalloc unless we are using it.
+if WITH_JEMALLOC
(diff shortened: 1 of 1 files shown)


markhpc:
I forgot to mention, this change is primarily required to when jemalloc is installed but not used for the build.

PR #9564, opened 2016-06-07
Added -f flag for formatting to rados command line tool

sdieffen:


src/rocksdb
```diff
@@ -1 +1 @@
-Subproject commit fa98456ccbd860120127ac2e91a2088fb222606e
+Subproject commit 5647fa427c4eab87c0caf8bcbb91f5846a4623f9
```
athanatos: Ah, you inadvertently updated the rocksdb sha1.  If you use git gui in amend mode, you can unselect the src/rocksdb entry and recommit.

src/tools/rados/rados.cc
```diff
@@ -3284,12 +3268,40 @@ int main(int argc, const char **argv)
...
+   * since global_init will remove the -f using ceph
+   * argparse functions. */
+  for (j = args.begin(); j != args.end(); ) {
+    if (strcmp(*j, "--") == 0) {
+      break;
+    } else if ((j+1) == args.end()) {
+      // This can't be a formatting call (no format arg)
+      break; 
+    } else if (strcmp(*j, "-f") == 0) {
+      val = *(j+1);
```
athanatos: j+1 might be end(), val here would be constructed from an invalid reference.

src/tools/rados/rados.cc
```diff
@@ -3284,12 +3268,40 @@ int main(int argc, const char **argv)
...
+
+      Formatter *formatter = NULL;
+      formatter = Formatter::create(val.c_str());
```
athanatos: Formatter *formatter = Formatter::create...

No reason to leave it uninitialized.

athanatos: Also, best to use unique_ptr here.

unique_ptr<Formatter> formatter(Formatter::create(val.c_str());

That way, the delete formatter below isn't necessary, and we don't risk future changes here introducing a leak.

PR #9567, opened 2016-06-07
os/bluestore: convert cache to be blob-based (vs onode-based)

liewegas:


liewegas:
@ifed01 can you review?

src/os/bluestore/BlueStore.cc
```diff
@@ -3194,94 +3222,108 @@ int BlueStore::_do_read(
...
-      uint32_t r_off = off - lext->first;
-      l2read = MIN(l, lext->second.length - r_off);
-      if (!ready_intervals_in_cache.contains( off, l2read)) {
-	regions2read_t& regions = blobs2read[bptr];
-	regions.push_back(region_t(off, r_off + lext->second.offset, l2read));
+    unsigned l_off = pos - lp->first;
+    unsigned b_off = l_off + lp->second.offset;
+    unsigned b_len = std::min(std::min(left,
+				       lp->second.length - l_off),
+			      bptr->blob.length - b_off);
```
ifed01: Not sure we need to use bptr->blob.length - b_off as an additional limitation. In a consistent state lp->second.length - l_off <= ptr->blob.length - b_off. Probably appropriate assert is enough.

src/os/bluestore/BlueStore.cc
```diff
@@ -3194,94 +3222,108 @@ int BlueStore::_do_read(
...
+      IOContext ioc(NULL);   // FIXME?
+      bptr->blob.map(
+	0, bptr->blob.get_ondisk_length(),
+	[&](uint64_t offset, uint64_t length) {
+	  bufferlist t;
+	  int r = bdev->read(offset, length, &t, &ioc, false);
+	  assert(r == 0);
+	  compressed_bl.claim_append(t);
+	});
+      if (bptr->blob.csum_type != bluestore_blob_t::CSUM_NONE &&
```
ifed01: we can get rid off the check for CSUM_NONE as blob_t::verify_csum does that by itself

src/os/bluestore/BlueStore.cc
```diff
@@ -5747,25 +5747,24 @@ void BlueStore::_do_write_small(
   }
 
   // new blob.
-  int64_t blob;
-  b = o->onode.add_blob(&blob);
-  b->length = min_alloc_size;
+  b = o->blob_map.new_blob(c->cache);
+  b->blob.length = min_alloc_size;
```
ifed01: we can modify new_blob(..) to assign blob.length by itself

src/os/bluestore/BlueStore.cc
```diff
@@ -6138,8 +6134,6 @@ int BlueStore::_do_zero(TransContext *txc,
   // they may touch.
   o->flush();
 
-  o->bc.discard(offset, length);
```
ifed01: No replacement for that? Since now you are relying on the lextent map to bypass cache usage for holes, right? IMHO that might be dangerous, wouldn't that bring some hidden issue? As for me I'd prefer to have consistent cache content too.

liewegas: Yeah i have a trello card open for this.  We probably just need to make the wctx_finish code that drops lextent refs also invalidate the blob cache buffers (assuming they are clean).

src/os/bluestore/BlueStore.cc
liewegas: Yeah we could.. I'm inclined to keep the BlobMap functions only concerned with the map structure though and not the blob data itself.

src/os/bluestore/BlueStore.cc
```diff
@@ -6505,19 +6497,23 @@ int BlueStore::_clone(TransContext *txc,
 
   if (g_conf->bluestore_clone_cow) {
     if (!oldo->onode.extent_map.empty()) {
-      BnodeRef e = c->get_bnode(newo->oid.hobj.get_hash());
+      if (!oldo->bnode) {
+	oldo->bnode = c->get_bnode(newo->oid.hobj.get_hash());
+      }
+      if (!newo->bnode) {
```
ifed01: Replace with assert( !newo->bnode || newo->bnode == oldo->bnode) ?

ifed01:
Reviewed-by:

PR #9572, opened 2016-06-08
client: fix wrong lock logic of ll_read_block() method

xiexingguo:


jcsp:
This looks correct, but perhaps when cleaning this up you could simplify it further by replacing the C_SafeCond with a C_SaferCond, so that the done/cond vars are no longer needed?  Currently we use a mixture, but when doing cleanup it is nice to replace things with SaferCond where possible

xiexingguo:
Done and label this one as "needs-qa".
Thanks for the review!

jcsp:
Reviewed-by

jcsp:

PR #9573, opened 2016-06-08
cmake: fix rocksdb linkage with jemalloc installed but not enabled

tchaikov:


tchaikov:
@markhpc i amended your fix.

src/kv/Makefile.am
```diff
@@ -26,14 +26,12 @@ NPROC = nproc
...
-DISABLE_JEMALLOC = 0
-else
-DISABLE_JEMALLOC = 1
+if !WITH_JEMALLOC
+export DISABLE_JEMALLOC=1
 endif
 
 rocksdb/librocksdb.a:
-	cd rocksdb && CC="${CC}" CXX="${CXX}" EXTRA_CXXFLAGS="${ROCKSDBCXX_FLAGS}" PORTABLE=1 DISABLE_JEMALLOC="${DISABLE_JEMALLOC}" ${MAKE} -j$(shell ${NPROC}) static_lib
+	cd rocksdb && CC="${CC}" CXX="${CXX}" EXTRA_CXXFLAGS="${ROCKSDBCXX_FLAGS}" PORTABLE=1 ${MAKE} -j$(shell ${NPROC}) static_lib
```
cbodley: why change the strategy for automake to use an environment variable, when the cmake version doesn't?

tchaikov: @cbodley if we put something like
```
a="FOO=bar"
all:
  ${a} make
```
Makefile tries to add the quotes around "FOO=bar". and it will call "FOO=bar" instead, and "make" will be "argv[1]".

but seems cmake does not suffer from this.

cbodley: @tchaikov it wasn't following that `a="FOO=bar"` pattern, though. the variable was `DISABLE_JEMALLOC` and it was being passed as `DISABLE_JEMALLOC="${DISABLE_JEMALLOC}"`. i thought the automake build was working when this merged - was it failing for you?

tchaikov: @cbodley 

automake actually **always** disables JeMalloc even it is enabled using "--with-jemalloc". but we don't pass this in our gitbuilders.

PR #9574, opened 2016-06-08
jewel: Missing export for rados_aio_get_version in src/include/rados/librados.h

Abhishekvrshny:
http://tracker.ceph.com/issues/16153


dachary:
It passed the rados suite http://tracker.ceph.com/issues/16344#note-7 except for one failure which is unrelated).

PR #9578, opened 2016-06-08
jewel: osd:sched_time not actually randomized

Abhishekvrshny:
http://tracker.ceph.com/issues/16080


dachary:
It passed the rados suite http://tracker.ceph.com/issues/16344#note-7 except for one failure which is unrelated).

PR #9579, opened 2016-06-08
jewel: BackoffThrottle spins unnecessarily with very small backoff while the throttle is full

Abhishekvrshny:
http://tracker.ceph.com/issues/16006

### DIFF
--- a/src/common/Throttle.cc
+++ b/src/common/Throttle.cc
@@ -376,11 +376,11 @@ std::chrono::duration<double> BackoffThrottle::get(uint64_t c)
 
   auto start = std::chrono::system_clock::now();
   delay = _get_delay(c);
-  while (((start + delay) > std::chrono::system_clock::now()) ||
+  while ((delay > std::chrono::duration<double>(0)) ||
 	 !((max == 0) || (current == 0) || ((current + c) <= max))) {
     assert(ticket == waiters.begin());
-    (*ticket)->wait_until(l, start + delay);
-    delay = _get_delay(c);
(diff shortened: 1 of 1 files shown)


wjwithagen:
@Abhishekvrshny 
You also might want to look at the discussion in #9111 and #9455.
And it looks like this is a better sollution, but it can still lockup.

But I'm not sure if that is only during unittest_throttle, or that the bug also manifests itself in real life

athanatos:
@liewegas I'd like to sneak this one in for 10.2.2.  I've got a wip-sam-jewel-testing branch building with it, but I think we should just stick it in jewel since we'll be testing over the weekend anyway.

wjwithagen:
@athanatos @liewegas 
You might to augment this with changes to the unittest_throttle code from #9111:
https://github.com/ceph/ceph/pull/9111/commits/4305248298d68bf7d59370e48860b3df46673dc0
https://github.com/ceph/ceph/pull/9111/commits/08decf5a5e7daccd9e94a74775263a1e2c94726f

ErwanAliasr1:
I wonder if that is related but almost all PRs are now failing because of the infinite loop...

wjwithagen:
@ErwanAliasr1 
I do not know about the failing PRs.
But The current discussion for me is about finishing unittest_throttle.
Could be that the same shutdown of the pipe comes into play for other tests, but I would not know

wjwithagen:
@ErwanAliasr1 
Just talked to @liewegas, and he suggested that testing was at a low speed ATM.
But was going to look into it.

PR #8754, opened 2016-04-26
jewel: global-init: fixup chown of the run directory along with log and asok files

kmroz:
http://tracker.ceph.com/issues/15608

PR #9581, opened 2016-06-08
jewel: Ownership of run dir, pid file and asok file not set when deferring permissions drop

Abhishekvrshny:
http://tracker.ceph.com/issues/15970


dachary:
Duplicate of https://github.com/ceph/ceph/pull/8754

PR #9582, opened 2016-06-08
swift /info implementation.

pritha-srivastava:
Implementation for swift /info API.


src/rgw/rgw_rest_swift.cc
```diff
@@ -1325,9 +1325,86 @@ void RGWBulkDelete_ObjStore_SWIFT::send_response()
...
+      s->formatter->open_object_section((it->first).c_str());
+      s->formatter->close_section();
+    }
+    else
+    {
+      it->second(*(s->formatter), *(s->cct->_conf));
+    }
+  }
+
+  s->formatter->close_section();
```
cbodley: the parts of an op that do actual work (including writing to the formatter) usually happen in the `execute()` function. can you move this block, starting with the `open_object_section()`?

src/rgw/rgw_op.h
```diff
@@ -1339,6 +1339,19 @@ class RGWDeleteMultiObj : public RGWOp {
...
+public:
+  RGWInfo() {}
+
+  void execute() {};
+  int verify_permission() { return 0; }
+
+  virtual void send_response() = 0;
+  virtual const string name() { return "get info"; }
+  virtual RGWOpType get_type() { return RGW_OP_GET_INFO; }
+  virtual uint32_t op_mask() { return RGW_OP_TYPE_READ; }
```
cbodley: `send_response()` is already a pure virtual in `RGWOp`, so you shouldn't need it here. for the others, can you use the `override` keyword instead of `virtual`?

src/rgw/rgw_op.h
```diff
@@ -1339,6 +1339,16 @@ class RGWDeleteMultiObj : public RGWOp {
   virtual uint32_t op_mask() { return RGW_OP_TYPE_DELETE; }
 };
 
+class RGWInfo: public RGWOp {
+public:
+  RGWInfo() {}
```
rzarzynski: This constructor is equivalent of the default one. If we really want to be explicit here, I would suggest rather `RGWInfo() = default;`.

src/rgw/rgw_rest.cc
```diff
@@ -1426,7 +1426,7 @@ int RGWHandler_REST::validate_bucket_name(const string& bucket)
 int RGWHandler_REST::validate_object_name(const string& object)
 {
   int len = object.size();
-  if (len > 1024) {
+  if (len > MAX_OBJ_NAME_LEN) {
```
rzarzynski: Removal of magic numbers is a really good thing. :-)

src/rgw/rgw_rest.h
```diff
@@ -350,6 +350,12 @@ class RGWDeleteMultiObj_ObjStore : public RGWDeleteMultiObj {
   virtual int get_params();
 };
 
+class RGWInfo_ObjStore : public RGWInfo {
+public:
+    RGWInfo_ObjStore() {}
+    ~RGWInfo_ObjStore() {}
```
rzarzynski: Default ctor and dtor - see the note on `RGWInfo`.

src/rgw/rgw_rest_swift.h
```diff
@@ -185,6 +185,29 @@ class RGWBulkDelete_ObjStore_SWIFT : public RGWBulkDelete_ObjStore {
...
+public:
+  RGWInfo_ObjStore_SWIFT() {}
+  ~RGWInfo_ObjStore_SWIFT() {}
+
+  void execute() override;
+  void send_response() override;
+  static void list_swift_data(ceph::Formatter& formatter, md_config_t& config, RGWRados& store);
+  static void list_tempurl_data(ceph::Formatter& formatter, md_config_t& config, RGWRados& store);
+  static void list_slo_data(ceph::Formatter& formatter, md_config_t& config, RGWRados& store);
+  static bool is_expired(const std::string& expires);
```
cbodley: these static functions, along with `struct info` and `swift_info`, probably don't need to be members of `RGWInfo_ObjStore_SWIFT`. consider making them static to rgw_rest_swift.cc so you don't need to qualify their definitions with `RGWInfo_ObjStore_SWIFT::`, and you can make changes without having to recompile everything that includes rgw_rest_swift.h

src/rgw/rgw_rest.h
```diff
@@ -385,6 +391,9 @@ class RGWHandler_REST : public RGWHandler {
   static int allocate_formatter(struct req_state *s, int default_formatter,
 				bool configurable);
 public:
+  static const int MAX_BUCKET_NAME_LEN = 255;
+  static const int MAX_OBJ_NAME_LEN = 1024;
```
rzarzynski: `constexpr` for those statics would be even more preferred than `const`. Someone in the future might want to make an arrays basing on them.

src/rgw/rgw_rest_swift.cc
```diff
@@ -1325,9 +1326,143 @@ void RGWBulkDelete_ObjStore_SWIFT::send_response()
...
+{
+  formatter.open_object_section("slo");
+  formatter.dump_int("max_manifest_segments", config.rgw_max_slo_entries);
+  formatter.close_section();
+}
+
+bool RGWInfo_ObjStore_SWIFT::is_expired(const std::string& expires)
+{
+  string err;
+  const utime_t now = ceph_clock_now(g_ceph_context);
```
cbodley: avoid `g_ceph_context` like the plague :) a `CephContext` should be available through req_state via `s->cct`

the `dout()` macro also depends on `g_ceph_context`, so please prefer `ldout(s->cct, 5)`

src/rgw/rgw_rest_swift.h
```diff
@@ -185,6 +185,29 @@ class RGWBulkDelete_ObjStore_SWIFT : public RGWBulkDelete_ObjStore {
...
 };
 
+class RGWInfo_ObjStore_SWIFT : public RGWInfo_ObjStore {
+protected:
+  struct info
+  {
+    bool is_admin_info;
+    void (*list_data)(Formatter& formatter,
+                              md_config_t& config,
+                              RGWRados& store);
```
cbodley: looks like `config` can be passed as a const ref. `store` _should_ be a const ref, but it doesn't look like there's a const version of `RGWRados::get_zonegroup()`

src/rgw/rgw_rest_swift.h
```diff
@@ -185,6 +185,29 @@ class RGWBulkDelete_ObjStore_SWIFT : public RGWBulkDelete_ObjStore {
...
+  };
+
+  static vector<pair<string, struct info>> swift_info;
```
cbodley: this can be `static const`

src/rgw/rgw_rest_swift.cc
```diff
@@ -1325,9 +1326,143 @@ void RGWBulkDelete_ObjStore_SWIFT::send_response()
...
+  const string& swiftinfo_expires = s->info.args.get("swiftinfo_expires");
+
+  if (!swiftinfo_sig.empty() && !swiftinfo_expires.empty() && !is_expired(swiftinfo_expires))
+  {
+    is_admin_info_enabled = true;
+  }
+
+  s->formatter->open_object_section("info");
+
+  for (auto it = swift_info.begin(); it != swift_info.end(); it++)
```
rzarzynski: It would be nice to see the ranged `for` variant here: `for (const auto& pair : swift_info) { ... }`.

src/rgw/rgw_rest_swift.cc
```diff
@@ -1325,9 +1326,143 @@ void RGWBulkDelete_ObjStore_SWIFT::send_response()
   rgw_flush_formatter_and_reset(s, s->formatter);
 }
 
+vector<pair<string, RGWInfo_ObjStore_SWIFT::info>> RGWInfo_ObjStore_SWIFT::swift_info =
+  {
+    {"bulk_delete", {false, NULL}},
```
rzarzynski: I guess that `nullptr` is preferred in a new code.

src/rgw/rgw_rest_swift.cc
```diff
@@ -1325,9 +1326,143 @@ void RGWBulkDelete_ObjStore_SWIFT::send_response()
...
+  formatter.dump_int("max_file_size", config.rgw_max_put_size);
+  formatter.dump_int("container_listing_limit", RGW_LIST_BUCKETS_LIMIT_MAX);
+
+  string ceph_version(CEPH_GIT_NICE_VER);
+  formatter.dump_string("version", ceph_version);
+  formatter.dump_int("max_meta_name_length", 81);
+
+  formatter.open_array_section("policies");
+  RGWZoneGroup& zonegroup = store.get_zonegroup();
+  for (auto it = zonegroup.placement_targets.begin();
```
rzarzynski: Ranged `for` could save us from splitting this across three lines.

src/rgw/rgw_rest_swift.cc
```diff
@@ -1325,9 +1326,143 @@ void RGWBulkDelete_ObjStore_SWIFT::send_response()
...
+  formatter.close_section();
+
+  formatter.dump_int("max_object_name_size", RGWHandler_REST::MAX_OBJ_NAME_LEN);
+  formatter.dump_bool("strict_cors_mode", true);
+  formatter.dump_int("max_container_name_length", RGWHandler_REST::MAX_BUCKET_NAME_LEN);
+  formatter.close_section();
+}
+
+void RGWInfo_ObjStore_SWIFT::list_tempurl_data(Formatter& formatter,
+                                                md_config_t& config,
```
rzarzynski: Spacing.

src/rgw/rgw_rest_swift.cc
```diff
@@ -1325,9 +1326,143 @@ void RGWBulkDelete_ObjStore_SWIFT::send_response()
...
+    return true;
+  }
+
+  return false;
+}
+
 RGWOp *RGWHandler_REST_Service_SWIFT::op_get()
 {
-  return new RGWListBuckets_ObjStore_SWIFT;
+  if(s->decoded_uri.compare("/swift/info") == 0)
```
rzarzynski: We should be aware of the `rgw_swift_url_prefix` configuration option. See how it's being handled [there](https://github.com/ceph/ceph/blob/master/src/rgw/rgw_swift_auth.cc#L514O).

src/rgw/rgw_rest_swift.h
```diff
@@ -185,6 +185,29 @@ class RGWBulkDelete_ObjStore_SWIFT : public RGWBulkDelete_ObjStore {
   void send_response();
 };
 
+class RGWInfo_ObjStore_SWIFT : public RGWInfo_ObjStore {
+protected:
+  struct info
+  {
+    bool is_admin_info;
+    void (*list_data)(Formatter& formatter,
+                              md_config_t& config,
```
rzarzynski: Indentation.

src/rgw/rgw_rest_swift.cc
pritha-srivastava: @rzarzynski : I went thru the code that you have mentioned in your comment above, but I am not sure I understand what we could do with rgw_swift_url_prefix here. Can you please elaborate a little.

cbodley: string comparisons against "/swift/info" should instead compare against "/${rgw_swift_url_prefix}/info"

pritha-srivastava: @cbodley : Thanks got it now.

pritha-srivastava:
@rzarzynski , @cbodley : Review comments incorporated. Please re-check. Thanks.

cbodley:
@pritha-srivastava thanks, looks good to me :+1: 

@rzarzynski do you know of any swift tests we can run to validate this?

rzarzynski:
Yup, there is a dedicated test in Swift `/info`. I will take a care of running that against this PR.

rzarzynski:
@pritha-srivastava, @cbodley: there is an additional problem here. Tempest excepts the `/info` entry point to be located at the root of the hierarchy:
```
http://127.0.0.1:8000/info
```
The same should be (according to [the documentation](http://docs.ceph.com/docs/master/radosgw/config-ref/#swift-settings) and [the comment](https://github.com/ceph/ceph/blob/master/src/rgw/rgw_swift_auth.cc#L514O)) possible in rgw by setting  the `rgw_swift_url_prefix` to `/`. Unfortunately, this doesn't work. After the configuration tuning, I can reach the `/info` using:
```
curl -i -H "X-Auth-Token: $token" "http://127.0.0.1:8000///info"  -X GET
```
It looks we doesn't handle the special case for `rgw_swift_url_prefix` in `RGWHandler_REST_SWIFT::init_from_header` as even to access a plain account I have to issue:
```
curl -i -H "X-Auth-Token: $token" "http://127.0.0.1:8000///v1"  -X GET
```
Of course, no regression here. Anyway, the `RGWHandler_REST_Service_SWIFT::op_get` exhibits a similar issue. I believe the problems are co-existing. We need to fix them all to get the proper behavior.

I'm working on the `RGWHandler_REST_SWIFT::init_from_header` as we'll need to backport the fix to Jewel.

src/rgw/rgw_rest_swift.cc
```diff
@@ -1325,9 +1326,145 @@ void RGWBulkDelete_ObjStore_SWIFT::send_response()
...
+    return true;
+  }
+
+  return false;
+}
+
 RGWOp *RGWHandler_REST_Service_SWIFT::op_get()
 {
-  return new RGWListBuckets_ObjStore_SWIFT;
+  string uri = "/" + s->cct->_conf->rgw_swift_url_prefix + "/info";
```
rzarzynski: We have to support the `/` case of `rgw_swift_url_prefix` here.

cbodley: we should support an empty prefix as well

yehudasa:
@pritha-srivastava @rzarzynski I think the swift at the root is a separate issue that we'll need to fix. For this specific issue we need to find a way to have /info from the root while maintaining a separate swift manager from the swift prefix. Not sure what would be the easiest path. Do users need to be authenticated to use /info? if not then we can leverage the S3 manager (ugly but should work). A better way with much more work would be to create a new api manager for this specific request (should probably just inherit from the swift one) that we'll register on /info.

pritha-srivastava:
@cbodley , @rzarzynski : Please re-review. I have incorporated Yehuda's comment.

src/rgw/rgw_rest_swift.cc
```diff
@@ -1846,3 +1992,11 @@ RGWHandler_REST* RGWRESTMgr_SWIFT::get_handler(struct req_state *s)
 
   return new RGWHandler_REST_Obj_SWIFT;
 }
+
+RGWHandler_REST* RGWRESTMgr_SWIFT_Info::get_handler(struct req_state *s)
+{
+  if (!RGWRESTMgr_SWIFT::get_handler(s))
+    return nullptr;
```
cbodley: this would leak the memory allocated by `RGWRESTMgr_SWIFT::get_handler()`. you probably just want to call `RGWHandler_REST_SWIFT::init_from_header()` here instead

pritha-srivastava:
@cbodley : Done. Do I really need to have RGWRESTMgr_SWIFT_Info inherit from RGWRESTMgr. I don't see any need to do so. Let me know what you think.

src/rgw/rgw_main.cc
```diff
@@ -345,6 +345,8 @@ int main(int argc, const char **argv)
 			   set_logging(new RGWRESTMgr_SWIFT));
   }
 
+  rest.register_resource("info", set_logging(new RGWRESTMgr_SWIFT_Info));
+
```
cbodley: this should be inside the `if (apis_map.count("swift") > 0) {` block above

rzarzynski: @pritha-srivastava: take a look on a [similar code](https://github.com/ceph/ceph/pull/10069/files#diff-703174d51f9d2cd706c9e59b8e7ce8f0R344) in the `crossdomain.xml` implementation.

src/rgw/rgw_rest_swift.cc
```diff
@@ -1325,6 +1326,138 @@ void RGWBulkDelete_ObjStore_SWIFT::send_response()
...
+
+  s->formatter->open_object_section("info");
+
+  for (const auto& pair : swift_info)
+  {
+    if(!is_admin_info_enabled && pair.second.is_admin_info)
+      continue;
+
+    if (!pair.second.list_data)
+    {
```
rzarzynski: I've got an impression that another bracing style is recommended for a new code: 
```
if (true) {
  // NOP
} else {
  // NOP
}
```

src/rgw/rgw_rest_swift.cc
```diff
@@ -1325,6 +1326,138 @@ void RGWBulkDelete_ObjStore_SWIFT::send_response()
...
+      !is_expired(swiftinfo_expires, s->cct))
+  {
+    is_admin_info_enabled = true;
+  }
+
+  s->formatter->open_object_section("info");
+
+  for (const auto& pair : swift_info)
+  {
+    if(!is_admin_info_enabled && pair.second.is_admin_info)
```
rzarzynski: Braces as well?

src/rgw/rgw_rest_swift.cc
```diff
@@ -1735,12 +1873,20 @@ int RGWHandler_REST_SWIFT::init_from_header(struct req_state *s)
...
+  if ((s->decoded_uri[0] != '/' ||
+    s->decoded_uri.compare(0, blen, buf) !=  0) &&
+    s->decoded_uri.compare(uri) != 0) {
+        return -ENOENT;
   }
 
-  int ret = allocate_formatter(s, RGW_FORMAT_PLAIN, true);
+  int ret;
+  //Set the formatter to JSON by default for /info api
+  if (s->decoded_uri.compare(uri) == 0)
```
rzarzynski: Bracing and indentation.

src/rgw/rgw_rest_swift.cc
```diff
@@ -1846,3 +1992,12 @@ RGWHandler_REST* RGWRESTMgr_SWIFT::get_handler(struct req_state *s)
 
   return new RGWHandler_REST_Obj_SWIFT;
 }
+
+RGWHandler_REST* RGWRESTMgr_SWIFT_Info::get_handler(struct req_state *s)
+{
+  int ret = RGWHandler_REST_SWIFT::init_from_header(s);
+  if (ret < 0)
+    return NULL;
```
rzarzynski: IMO it would be nice to see `nullptr` in a new code.

src/rgw/rgw_rest_swift.h
```diff
@@ -264,4 +294,11 @@ class RGWRESTMgr_SWIFT : public RGWRESTMgr {
   virtual RGWHandler_REST *get_handler(struct req_state *s);
 };
 
+class RGWRESTMgr_SWIFT_Info : public RGWRESTMgr_SWIFT {
+public:
+  RGWRESTMgr_SWIFT_Info() = default;
+  virtual ~RGWRESTMgr_SWIFT_Info() = default;
+  virtual RGWHandler_REST *get_handler(struct req_state *s);
```
rzarzynski: Maybe add `override` and skip `virtual`?

cbodley:
@rzarzynski any luck testing this with tempest?

rzarzynski:
@pritha-srivastava, @cbodley: to have meaningful results we would need to test this PR together with #9900 because having `/info` in the root of URL hierarchy is a must-have for Tempest. Similar limitation applies to PRs #10069 and #10213, so maybe creating a bundle from them is an option. What do you think?

pritha-srivastava:
@rzarzynski :  As suggested by Yehuda, by introducing a new RestMgr, now /info has been placed at the root hierarchy. We can use curl -X 'GET' http://localhost:8000/info to get back the results.

rzarzynski:
@pritha-srivastava: #10280 consists a rebased version of this commit. Another patch adapts it to a generic solution that allows to preserve the `info` bucket name for S3 while still running Swift API. Could you please take a look?

pritha-srivastava:
@rzarzynski : I have been taking a look, so far things seem fine. I just want to understand how init_from_headers will work, spending some time understanding that method.

yehudasa:
@pritha-srivastava this one needs to be rebased

rzarzynski:
@yehudasa, @pritha-srivastava: we have an alternative here. We might want to go with #10280 that includes the commit from this PR.

pritha-srivastava:
@yehudasa @rzarzynski : Yes, I agree with Radoslaw, The PR that he has mentioned has all the latest changes.

rzarzynski:
Merged as a part of #10280.

PR #9583, opened 2016-06-08
client: fix potential access violation

xiexingguo:
The _read() call may fail and return a negative result,
and thus when it is passed to "resid", which is of type of
"unsigned", overflow happens due to a forced cast that is
implicitly performed.


jcsp:
Reviewed-by

jcsp:

PR #9584, opened 2016-06-08
rgw: fix Swift API returns double space usage

chiehtu:
Fixes: http://tracker.ceph.com/issues/16188
Signed-off-by: Albert Tu <albert.t@inwinstack.com>

### DIFF
--- a/src/rgw/rgw_rados.cc
+++ b/src/rgw/rgw_rados.cc
@@ -10841,6 +10841,9 @@ int RGWRados::update_containers_stats(map<string, RGWBucketEnt>& m)
   for (iter = m.begin(); iter != m.end(); ++iter) {
     RGWBucketEnt& ent = iter->second;
     rgw_bucket& bucket = ent.bucket;
+    ent.count = 0;
+    ent.size = 0;
+    ent.size_rounded = 0;
 
     map<string, rgw_bucket_dir_header> headers;
     int r = cls_bucket_head(bucket, RGW_NO_SHARD, headers);


src/rgw/rgw_op.cc
```diff
@@ -1491,7 +1491,7 @@ void RGWStatAccount::execute()
     RGWUserBuckets buckets;
 
     op_ret = rgw_read_user_buckets(store, s->user->user_id, buckets, marker,
-				   string(), max_buckets, true, &is_truncated);
```
oritwas: This will means not stat at all , it is not related to the issue

chiehtu:
@oritwas The original code count based on stats of bucket, so that Swift API returns double space usage. In order to prevent this bug, I declared the variables for saving stats of headers and then assign them to bucket at the bottom of `buckets` loop.

src/rgw/rgw_rados.cc
```diff
@@ -10818,8 +10818,14 @@ int RGWRados::omap_del(rgw_obj& obj, const std::string& key)
...
 int RGWRados::update_containers_stats(map<string, RGWBucketEnt>& m)
 {
+  uint64_t total_count;
+  size_t total_size;
+  size_t total_size_rounded;
   map<string, RGWBucketEnt>::iterator iter;
   for (iter = m.begin(); iter != m.end(); ++iter) {
+    total_count = 0;
+    total_size = 0;
+    total_size_rounded = 0;
```
cbodley: could you change these to `ent.total_count = 0;` etc. so we don't need the local variables?

PR #9585, opened 2016-06-08
librbd: Asynchronous (v2) image creation

vshankar:


vshankar:
@dillaman I'm investigating the failure, but the patch should mostly be reviewable.

src/librbd/CreateRequest.cc
```diff
@@ -0,0 +1,633 @@
+// -*- mode:C++; tab-width:8; c-basic-offset:2; indent-tabs-mode:t -*-
```
dillaman: Minor: move to "operation" subdirectory / namespace

src/librbd/CreateRequest.cc
```diff
@@ -0,0 +1,633 @@
...
+  journal_pool(_journal_pool), non_primary_global_image_id(_non_primary_global_image_id),
+  primary_mirror_uuid(_primary_mirror_uuid), on_finish(_on_finish) {
+
+  cct = reinterpret_cast<CephContext *>(ioctx.cct());
+
+  // generate an id for this image
+  uint32_t extra = rand() % 0xFFFFFFFF;
+  ostringstream bid_ss;
+  bid_ss << std::hex << bid << std::hex << extra;
+  id = bid_ss.str();
```
dillaman: I'd like to see the "id" passed in as a param to the state machine

src/librbd/CreateRequest.cc
```diff
@@ -0,0 +1,633 @@
...
+  }
+
+  return true;
+}
+
+} // anonymous namespace
+
+namespace librbd {
+
+class ThreadPoolSingleton : public ThreadPool {
```
dillaman: No need for a dedicated thread if you pass in a ImageCtx instance

src/librbd/CreateRequest.cc
```diff
@@ -0,0 +1,633 @@
...
+#include "librbd/Journal.h"
+#include "librbd/MirroringWatcher.h"
+
+#define dout_subsys ceph_subsys_rbd
+#undef dout_prefix
+#define dout_prefix *_dout << "librbd::CreateRequest: "
+
+namespace {
+
+bool is_feature_set_inter_compatible(CephContext *cct,
```
dillaman: ```validate_features```

src/librbd/CreateRequest.h
```diff
@@ -0,0 +1,140 @@
...
+
+  void send();
+
+private:
+  /**
+   * @verbatim
+   * -- TODO --
+   * @endverbatim
+   */
+  IoCtx ioctx;
```
dillaman: Need ```m_``` prefix for member variables

src/librbd/CreateRequest.h
```diff
@@ -0,0 +1,140 @@
...
+                             non_primary_global_image_id, primary_mirror_uuid,
+                             on_finish);
+  }
+
+  void send();
+
+private:
+  /**
+   * @verbatim
+   * -- TODO --
```
dillaman: Need fancy ASCII art state diagram

src/librbd/CreateRequest.cc
```diff
@@ -0,0 +1,633 @@
...
+  uint32_t extra = rand() % 0xFFFFFFFF;
+  ostringstream bid_ss;
+  bid_ss << std::hex << bid << std::hex << extra;
+  id = bid_ss.str();
+
+  id_obj = util::id_obj_name(imgname);
+  header_obj = util::header_name(id);
+  objmap_name = ObjectMap::object_map_name(id, CEPH_NOSNAP);
+
+  layout = file_layout_t();
```
dillaman: Minor: implicit

src/librbd/CreateRequest.cc
```diff
@@ -0,0 +1,633 @@
...
+  }
+  if (!validate_layout(cct, size, layout)) {
+    complete(-EINVAL);
+    return;
+  }
+
+  do_validate_pool();
+}
+
+void CreateRequest::do_validate_pool() {
```
dillaman: Minor: ```do_validate_pool``` -> ```validate_pool```

src/librbd/CreateRequest.cc
```diff
@@ -0,0 +1,633 @@
...
+  if (*result == 0) {
+    make_id_object();
+    return nullptr;
+  } else if ((*result < 0) && (*result != -ENOENT)) {
+    lderr(cct) << "failed to stat RBD directory: " << cpp_strerror(*result) << dendl;
+    return on_finish;
+  }
+
+  /*
+   * allocate a self-managed snapshot id if this a new pool to force
```
dillaman: Minor: use C++ -style comments "//"

src/librbd/CreateRequest.cc
```diff
@@ -0,0 +1,633 @@
...
+               << ": " << cpp_strerror(r) << dendl;
+    *result = r;
+    return on_finish;
+  }
+
+  make_id_object();
+  return nullptr;
+}
+
+void CreateRequest::make_id_object() {
```
dillaman: Minor: s/make/create/g

src/librbd/CreateRequest.cc
```diff
@@ -0,0 +1,633 @@
...
+  assert(r == 0);
+  comp->release();
+}
+
+Context *CreateRequest::handle_add_image_to_directory(int *result) {
+  ldout(cct, 20) << __func__ << ": r=" << *result << dendl;
+
+  if (*result < 0) {
+    lderr(cct) << "error adding image to directory: " << cpp_strerror(*result) << dendl;
+    save_retval_and_cleanup(*result, &CreateRequest::remove_id_object);
```
dillaman: Seems like overkill to use functors -- just save the retval and invoke ```remove_id_object```

src/librbd/CreateRequest.cc
```diff
@@ -0,0 +1,633 @@
...
+    lderr(cct) << "error writing header: " << cpp_strerror(*result) << dendl;
+    save_retval_and_cleanup(*result, &CreateRequest::remove_from_dir);
+    return nullptr;
+  }
+
+  do_object_map_resize();
+  return nullptr;
+}
+
+void CreateRequest::do_object_map_resize() {
```
dillaman: Minor: s/do_//g

src/librbd/CreateRequest.cc
```diff
@@ -0,0 +1,633 @@
...
+  ldout(cct, 20) << this << " " << __func__ << dendl;
```
dillaman: Skip to next state if object map isn't enabled

src/librbd/CreateRequest.cc
```diff
@@ -0,0 +1,633 @@
...
+}
+
+/*
+ * There doesn't exist native asynchronous call for journal creation.
+ * Executing this synchronously in rados callback context is prone to
+ * deadlocks
+ */
+void CreateRequest::journal_create() {
+  ldout(cct, 20) << this << " " << __func__ << dendl;
+
```
dillaman: Skip if journal feature isn't enabled

src/librbd/CreateRequest.cc
```diff
@@ -0,0 +1,633 @@
...
+  Context *ctx = new FunctionContext([this](int r) {
+      r = Journal<>::create(ioctx, id, journal_order, journal_splay_width,
```
dillaman: As a prior commit, you should add support for async journal creation

src/librbd/CreateRequest.cc
```diff
@@ -0,0 +1,633 @@
...
+void CreateRequest::mirror_image_enable() {
+  ldout(cct, 20) << this << " " << __func__ << dendl;
+
+  mirror_image_internal.state = cls::rbd::MIRROR_IMAGE_STATE_ENABLED;
+  if (non_primary_global_image_id.empty()) {
+    uuid_d uuid_gen;
+    uuid_gen.generate_random();
+    mirror_image_internal.global_image_id = uuid_gen.to_string();
+  } else {
+    mirror_image_internal.global_image_id = non_primary_global_image_id;
```
dillaman: Note: this should be semi-temporary since I think rbd-mirror should do this before it attempts to create the image (w/ state of creating)

src/librbd/CreateRequest.cc
```diff
@@ -0,0 +1,633 @@
...
+  }
+
+  send_watcher_notification();
+  return nullptr;
+}
+
+void CreateRequest::send_watcher_notification() {
+  ldout(cct, 20) << this << " " << __func__ << dendl;
+
+  Context *ctx = new FunctionContext([this](int r) {
```
dillaman: Create async version

src/librbd/CreateRequest.cc
vshankar: That's done in `fetch_mirror_mode()`

src/librbd/CreateRequest.cc
vshankar: Ah nice :)

I had less understanding of possible uses of ImageCtx -- in the sense, use it when there a need to open an image and things of that nature. Looking more into ImageCtx.cc, I understand it can be of used to offload more operations. Thanks! I'll do the required changes.

src/cls/rbd/cls_rbd_client.cc
```diff
@@ -149,6 +149,18 @@ namespace librbd {
                                          parent);
     }
 
+    void create_image(librados::ObjectWriteOperation *op, uint64_t size, uint8_t order,
```
dillaman: I'd leave all the sync versions that are still needed by librbd -- just have the sync version call the async version to setup the object operation.

src/librbd/CMakeLists.txt
```diff
@@ -12,6 +12,7 @@ set(librbd_internal_srcs
   AsyncOperation.cc
   AsyncRequest.cc
   CopyupRequest.cc
+
```
dillaman: Minor: can remove this blank line

src/librbd/internal.cc
```diff
@@ -72,6 +73,19 @@ namespace librbd {
 
 namespace {
 
+struct C_CreateV2Image : public Context {
```
dillaman: Not needed -- pass in the pointer to a stack-allocated ```C_SaferCond``` for the ```Context``` and use the ```wait``` function to retrieve the result.

dillaman: Never mind -- I see that you are using to clean up the temporary ImageCtx. You actually don't need the temporary ImageCtx since you've templated on it -- you can just pass the IoCtx, image id, and image name.

src/librbd/internal.cc
```diff
@@ -1017,8 +1031,8 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force,
   }
 
   int create_v2(IoCtx& io_ctx, const char *imgname, uint64_t bid, uint64_t size,
-		int order, uint64_t features, uint64_t stripe_unit,
-		uint64_t stripe_count, uint8_t journal_order,
+                int order, uint64_t features, uint64_t stripe_unit,
```
dillaman: Minor: assumption is that this will be removed when this PR is fully ready.

src/librbd/image/CreateRequest.cc
```diff
@@ -0,0 +1,678 @@
...
+    m_layout.stripe_count = 1;
+  } else {
+    m_layout.stripe_unit = m_stripe_unit;
+    m_layout.stripe_count = m_stripe_count;
+  }
+
+  m_force_non_primary = !non_primary_global_image_id.empty();
+}
+
+// TODO: dtor()
```
dillaman: Is there something you need in the destructor?

src/librbd/image/CreateRequest.cc
```diff
@@ -0,0 +1,678 @@
...
+    complete(-EINVAL);
+    return;
+  }
+
+  validate_pool();
+}
+
+template<typename I>
+void CreateRequest<I>::validate_pool() {
+  ldout(m_cct, 20) << this << " " << __func__ << dendl;
```
dillaman: This step should be skipped if the rbd_validate_pool config option is false.

src/librbd/image/CreateRequest.cc
```diff
@@ -0,0 +1,678 @@
...
+
+template<typename I>
+Context *CreateRequest<I>::handle_add_image_to_directory(int *result) {
+  ldout(m_cct, 20) << __func__ << ": r=" << *result << dendl;
+
+  if (*result < 0) {
+    lderr(m_cct) << "error adding image to directory: " << cpp_strerror(*result) << dendl;
+
+    m_r_saved = *result;
+    CreateRequest<I>::remove_id_object();
```
dillaman: Minor: no need for ```CreateRequest<I>::``` prefix to call your own member functions.

src/librbd/image/CreateRequest.cc
```diff
@@ -0,0 +1,678 @@
...
+  assert(r == 0);
+  comp->release();
+}
+
+template<typename I>
+Context *CreateRequest<I>::handle_fetch_mirror_image(int *result) {
+  ldout(m_cct, 20) << __func__ << ": r=" << *result << dendl;
+
+  // this should always be the case during create
+  assert(*result == -ENOENT);
```
dillaman: Not a safe assumption -- better to fail gracefully when dealing with out-of-band data inputs.

src/librbd/image/CreateRequest.cc
```diff
@@ -0,0 +1,678 @@
...
+  send_watcher_notification();
+  return nullptr;
+}
+
+template<typename I>
+void CreateRequest<I>::send_watcher_notification() {
+  ldout(m_cct, 20) << this << " " << __func__ << dendl;
+
+  Context *ctx = new FunctionContext([this](int r) {
+      r = MirroringWatcher<>::notify_image_updated(
```
dillaman: Probably a good candidate for a TODO saying make this async.  The current ```image_watcher::Notifier``` code handles async notifications and callbacks.  It should be moved to ```object_watcher::Notifier``` for re-use.

src/librbd/image/CreateRequest.cc
```diff
@@ -0,0 +1,678 @@
...
+template<typename I>
+void CreateRequest<I>::handle_watcher_notify(int r) {
+  ldout(m_cct, 20) << __func__ << ": r=" << r << dendl;
+
+  if (r < 0) {
+    lderr(m_cct) << "failed to send update notification: " << cpp_strerror(r)
+                 << dendl;
+
+    m_r_saved = r;
+    CreateRequest<I>::mirror_image_remove();
```
dillaman: I don't think it's worth failing the create if one-or-more watchers failed to ack it in time.  They will pick up the update when they notice they missed an update (and rescan the full mirroring directory).

src/librbd/image/CreateRequest.h
```diff
@@ -0,0 +1,130 @@
...
+                             stripe_count, journal_order,journal_splay_width, journal_pool,
+                             non_primary_global_image_id, primary_mirror_uuid, on_finish);
+  }
+
+  void send();
+
+private:
+  /**
+   * @verbatim
+   * -- TODO --
```
dillaman: Minor: state diagram

dillaman:
@vshankar looking very good

src/librbd/image/CreateRequest.cc
vshankar: That's a left over from previous patch. I haven't incorporated all earlier comments -- sent this out so as get a successful build run and the last patchset has unittest_librbd failing..

src/librbd/image/CreateRequest.cc
vshankar: My bad.

src/librbd/internal.cc
vshankar: I actually wanted to pack as much as possible into ImageCtx and handover that to CreateRequest. ImageCtx has member variables such as order, features, size, etc. which are filled during Refresh, so, I kept away from filling ImageCtx directly (via additional ctor).

src/cls/rbd/cls_rbd_client.cc
vshankar: The sync versions do use the async helpers which got introduced in the prior commit. This commit changes the sync version to use async helpers.

src/librbd/internal.cc
dillaman: The downside is that you need to handle the deletion of the dummy ImageCtx here and eventually within rbd-mirror (since it will invoke ```CreateRequest::create``` directly).

src/librbd/internal.cc
vshankar: We need to retain the sync version, right? This uses `C_SaferCond` w/ async `create_v2()`.

dillaman: No need to keep the sync version -- the only person going to call it is the rbd CLI (via the API) so that's safe.  It's potentially not safe if an async state machine invokes a blocking sync state machine.

vshankar: Ah yes, create_v2() is internal...

vshankar:
@dillaman 
Reverting commit 6c74381 ("test: fix linkage on rbd/journal tests") caused lots of misbehavior while running unit tests. So, I didn't go that route (and it possibly needs to be fixed as a separate PR). For now, instead of using copy constructor for `IoCtx` (which I had introduced after reverting the above patch), the patch just uses `dup()` in `CreateRequest`.

Also, I'm a bit unsure about the usage of `ContextWQ` in `create()` in `internal.cc`. An explicit `drain()` had to be invoked before the work queue goes out of scope. By the looks of it `PointerWQ` increments/decrements `m_processing` correctly -- so, I'm still seeing why `drain()` is required. I'll see if things can be improved further.

src/cls/rbd/cls_rbd_client.cc
```diff
@@ -862,6 +897,16 @@ namespace librbd {
       return ioctx->exec(oid, "rbd", "dir_remove_image", in, out);
```
dillaman: Reuse ```dir_remove_image``` below

src/cls/rbd/cls_rbd_client.cc
```diff
@@ -1249,11 +1306,7 @@ namespace librbd {
       }
 
       bufferlist::iterator iter = out_bl.begin();
-      r = mirror_image_get_finish(&iter, mirror_image);
```
dillaman: Minor: the rationale why I keep this style is for debugging -- it's super easy to set a debug position on the failure path vs a common path where you hope the variable isn't optimized away for a breakpoint condition.

src/librbd/ImageCtx.h
```diff
@@ -204,6 +217,13 @@ namespace librbd {
       delete this;
     }
 
+    static ImageCtxThreadPoolSingleton *get_thread_pool_instance(CephContext *cct) {
+      ImageCtxThreadPoolSingleton *thread_pool_singleton;
```
dillaman: Minor: move implementation back to C file and avoid #include updates

src/librbd/image/CreateRequest.h
```diff
@@ -0,0 +1,170 @@
+// -*- mode:C++; tab-width:8; c-basic-offset:2; indent-tabs-mode:t -*-
+// vim: ts=8 sw=2 smarttab
+#ifndef CEPH_LIBRBD_CREATEREQUEST_H
```
dillaman: Minor: extra line break spacing between lines 2 + 3 and use "_" between words (CREATE_REQUEST)

src/librbd/internal.cc
```diff
@@ -411,6 +412,14 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force,
 
 } // anonymous namespace
 
+  std::string generate_image_id(uint64_t bid)
```
dillaman: Minor: might be nice to move to Utils since rbd-mirror will need it.

src/test/rbd_mirror/image_replayer/test_mock_CreateImageRequest.cc
```diff
@@ -309,12 +320,10 @@ class TestMockImageReplayerCreateImageRequest : public TestMockFixture {
...
 TEST_F(TestMockImageReplayerCreateImageRequest, Create) {
   librbd::MockTestImageCtx mock_remote_image_ctx(*m_remote_image_ctx);
-
-  CreateCloneImage create_clone_image;
+  MockCreateRequest mock_create_request;
 
   InSequence seq;
-  expect_create_image(create_clone_image, "image name", "global uuid",
-                      "remote uuid", 0);
+  expect_create_image(mock_create_request, 0);
```
dillaman: Minor: you can add a MOCK_METHOD(construct, void(<image name>, <global uuid>, <remote mirror uuid>)) and validate the correct values are passed to the constructor.

src/tools/rbd_mirror/image_replayer/CreateImageRequest.cc
```diff
@@ -59,35 +60,30 @@ template <typename I>
...
+  std::string journal_pool = cct->_conf->rbd_journal_pool;
+
+  // NOTE: bid is 64bit but overflow will result due to
+  // RBD_MAX_BLOCK_NAME_SIZE being too small
+  librados::Rados rados(m_local_io_ctx);
+  uint64_t bid = rados.get_instance_id();
+  std::string id = librbd::generate_image_id(bid);
+
+  librbd::image::CreateRequest<I> *req = librbd::image::CreateRequest<I>::create(
+    m_local_io_ctx, m_local_image_name.c_str(), id, m_remote_image_ctx->size,
```
dillaman: Minor: no need for ```.c_str()```

src/librbd/image/CreateRequest.cc
```diff
@@ -0,0 +1,696 @@
...
+  m_op_work_queue(op_work_queue), m_on_finish(on_finish) {
+
+  m_ioctx.dup(ioctx);
+  m_cct = reinterpret_cast<CephContext *>(m_ioctx.cct());
+
+  m_id_obj = util::id_obj_name(m_imgname);
+  m_header_obj = util::header_name(m_imageid);
+  m_objmap_name = ObjectMap::object_map_name(m_imageid, CEPH_NOSNAP);
+
+  m_layout = file_layout_t();
```
dillaman: Minor: unnecessary

src/librbd/image/CreateRequest.cc
```diff
@@ -0,0 +1,696 @@
...
+    m_layout.stripe_unit = m_stripe_unit;
+    m_layout.stripe_count = m_stripe_count;
+  }
+
+  m_force_non_primary = !non_primary_global_image_id.empty();
+}
+
+template<typename I>
+CreateRequest<I>::~CreateRequest() {
+  m_ioctx.close();
```
dillaman: Minor: unnecessary

src/librbd/image/CreateRequest.cc
```diff
@@ -0,0 +1,696 @@
...
+                 << cpp_strerror(r) << dendl;
+    *result = r;
+    return m_on_finish;
+  }
+
+  r = m_ioctx.selfmanaged_snap_remove(snap_id);
+  if (r < 0) {
+    lderr(m_cct) << "failed to release self-managed snapshot " << snap_id
+                 << ": " << cpp_strerror(r) << dendl;
+    *result = r;
```
dillaman: Minor: ignore the error -- no recovery so no point in failing since we have accomplished the purpose of switching to self-managed snapshot mode on the pool.

src/librbd/image/CreateRequest.cc
```diff
@@ -0,0 +1,696 @@
...
+
+  ostringstream oss;
+  oss << RBD_DATA_PREFIX << m_imageid;
+
+  librados::ObjectWriteOperation op;
+  op.create(true);
+  cls_client::create_image(&op, m_size, m_order, m_features, oss.str());
+  if ((m_stripe_unit || m_stripe_count) &&
+      ((m_stripe_count != 1) || (m_stripe_unit != (1ull << m_order)))) {
+    cls_client::set_stripe_unit_count(&op, m_stripe_unit, m_stripe_count);
```
dillaman: I think this will cause an -ENOENT since by combining the ops, they will be executed in the same transaction (where the image header hasn't been created by the time this method executes and checks that the image exists).

src/tools/rbd_mirror/image_replayer/CreateImageRequest.cc
vshankar: Isn't that needed for `std::string` to `const char *`?

dillaman: Yeah -- but CreateRequest should just take a std::string instead of the const char *

vshankar: Oh OK. that should be fine as create_v2 can directly pass `const char *` pass it to `std::string`.

src/librbd/image/CreateRequest.cc
vshankar: Well, I would have been happy if it returned `-ENOENT` -- what we actually get is `-ENOEXEC` from `require_feature()` in `set_stripe_unit_count()` which happens *after* a `stat()` check on the object. So, the (header) object exist, but reading omap value (`features` key in this case) fails.

vshankar:
@dillaman Thanks for reviewing.
I'm not very sure about how to introduce `construct` in unit test cases. I took a look at other implementations under `src/test` and it seems to me that basic checks for `IoCtx` and `ImageCtx` are performed (not for all parameters in a constructor). I added check for `IoCtx` in the PR --  need to read more about gmock.

Also, thanks for the explanation of `WorkQueue::drain()` -- I was expecting it to just drain pending items in the queue, but it also takes care of finishing execution the current item.

dillaman:
@vshankar For the ```construct``` method, it wasn't so much to test the ```IoCtx``` but rather to test the ids / uuids that were being tested before the change.

vshankar:
@ceph-jenkins retest this please

vshankar:
@ceph-jenkins retest this please

vshankar:
@dillaman all checks have passed -- ready for review.

Thanks.

src/journal/Journaler.cc
```diff
@@ -199,45 +202,46 @@ void Journaler::get_mutable_metadata(uint64_t *minimum_set,
...
+  comp->release();
 }
 
-int Journaler::remove(bool force) {
-  C_SaferCond ctx;
-  m_metadata->shut_down(&ctx);
-  ctx.wait();
+void Journaler::remove(bool force, Context *on_finish) {
+  // chain journal removal (reverse order)
+  librados::AioCompletion *comp = librados::Rados::aio_create_completion(
```
dillaman: Can more into the FunctionContext

src/librbd/CMakeLists.txt
```diff
@@ -30,12 +30,15 @@ set(librbd_internal_srcs
...
   exclusive_lock/StandardPolicy.cc
   image/CloseRequest.cc
+  image/CreateRequest.cc
   image/OpenRequest.cc
   image/RefreshParentRequest.cc
   image/RefreshRequest.cc
   image/SetSnapRequest.cc
   image_watcher/Notifier.cc
   image_watcher/NotifyLockOwner.cc
+  journal/Create.cc
```
dillaman: Minor: suggest rename to ```XyzRequest``` similar to other state machines

src/librbd/ImageCtx.h
```diff
@@ -15,6 +15,7 @@
 #include "common/Readahead.h"
 #include "common/RWLock.h"
 #include "common/snap_types.h"
+#include "common/WorkQueue.h"
```
dillaman: Minor: forward declare ThreadPool so you don't need to include it

src/librbd/Journal.h
```diff
@@ -109,6 +109,9 @@ class Journal {
   static int request_resync(ImageCtxT *image_ctx);
   static int promote(ImageCtxT *image_ctx);
 
+  static void init_safetimer_instance(CephContext *cct, SafeTimer **timer,
```
dillaman: Minor: I think this will be re-usable elsewhere in librbd, so can you move it to ImageCtx and rename it ```get_timer_instance()```?

src/librbd/Journal.cc
```diff
@@ -353,59 +350,28 @@ bool Journal<I>::is_journal_supported(I &image_ctx) {
...
-    lderr(cct) << __func__ << ": "
-               << "failed to create journal: " << cpp_strerror(r) << dendl;
-    return r;
-  }
-
-  cls::journal::Client client;
-  cls::journal::Tag tag;
   journal::TagData tag_data;
+  tag_data.mirror_uuid = (non_primary ? primary_mirror_uuid
+                          : LOCAL_MIRROR_UUID);
```
dillaman: Minor: move colon to end of previous line

dillaman: ... actually, would anyone call this static create method with non_primary and primary_mirror_uuid arguments?  That should only be used in the image::CreateRequest path.

src/librbd/image/CreateRequest.cc
```diff
@@ -0,0 +1,795 @@
...
+template<typename I>
+void CreateRequest<I>::journal_create() {
+  ldout(m_cct, 20) << this << " " << __func__ << dendl;
+
+  Journal<I>::init_safetimer_instance(m_cct, &m_timer, &m_timer_lock);
+
+  // there's no need for calling ->init() as there aren't any journal
+  // objects to remove in case of failure from now on in the state
+  // machine -- header object purging is handled by ->remove()
+  m_journaler = new Journaler(m_op_work_queue, m_timer, m_timer_lock,
```
dillaman: It would probably be safer to have the journal::Create/RemoveRequest state machines instantiate and initialize the journaler object.  For example, if journal::CreateJournal fails before it inits the journal, journal::RemoveJournal will fail since the journal isn't initialized.

src/librbd/image/CreateRequest.h
```diff
@@ -0,0 +1,199 @@
...
+  CreateRequest(IoCtx &ioctx, std::string &imgname, std::string &imageid,
+                uint64_t size, int order, uint64_t features,
+                uint64_t stripe_unit, uint64_t stripe_count, uint8_t journal_order,
+                uint8_t journal_splay_width, const std::string &journal_pool,
+                const std::string &non_primary_global_image_id,
+                const std::string &primary_mirror_uuid,
+                ContextWQ *op_work_queue, Context *on_finish);
+
+  IoCtx m_ioctx;
+  std::string m_imgname;
```
dillaman: Minor: m_image_name / m_image_id

src/librbd/journal/Create.cc
```diff
@@ -0,0 +1,161 @@
...
+  ::encode(m_tag_data, m_bl);
+  m_journaler->allocate_tag(m_tag_class, m_bl, &m_tag, ctx);
+}
+
+template<typename I>
+Context *CreateJournal<I>::handle_journal_tag(int *result) {
+  ldout(m_cct, 20) << __func__ << ": r=" << *result << dendl;
+
+  if (*result < 0) {
+    lderr(m_cct) << "failed to allocate tag: " << cpp_strerror(*result) << dendl;
```
dillaman: Your failure path from this point forward should shut down the journal::Journaler and invoke the journal::RemoveRequest state machine to ensure you've properly cleaned up or the callers of this state machines should expect to potentially remove the journal upon failure.

src/librbd/image/CreateRequest.cc
```diff
@@ -0,0 +1,795 @@
...
+  if (*result == 0) {
+    ldout(m_cct, 20) << "done." << dendl;
+  }
+  return m_on_finish;
+}
+
+// helper
+template<typename I>
+void CreateRequest<I>::shut_down_and_cleanup_journaler(Context *ctx) {
+  ldout(m_cct, 20) << this << " " << __func__ << dendl;
```
dillaman: This can be removed once the lifetime of journal::Journaler is owned by the journal::Create/RemoveRequest state machines.

src/librbd/journal/Remove.cc
```diff
@@ -0,0 +1,101 @@
...
+template<typename I>
+Context *RemoveJournal<I>::handle_remove_journal(int *result) {
+  ldout(m_cct, 20) << __func__ << ": r=" << *result << dendl;
+
+  if (*result < 0) {
+    lderr(m_cct) << "failed to remove journal: " << cpp_strerror(*result) << dendl;
+    return m_on_finish;
+  }
+
+  complete(0);
```
dillaman: No need for the ```complete``` method -- just return ```m_on_finish``` and it will complete and delete the object.

dillaman:
Can squash your last four "yank" commits into a single commit and retitle to something along the lines of "removed unused synchronous API methods"

dillaman:
@vshankar looks very close -- can't wait to see it integrated into rbd-mirror directly.

src/librbd/journal/Remove.cc
vshankar: I kept it for completeness sake -- no worries, will remove `complete()`

src/librbd/image/CreateRequest.cc
vshankar: I'm ok either way. It's done this way as there was a possibility to use `journaler` in multiple place in `CreateRequest`.

src/librbd/Journal.cc
vshankar: I kept that in sync with what done in `Journal::create` -- before the call to `allocate_journaler_tag()`. I see there's just one caller - `update_features()` which passes `non_primary` as `false` (and the other is create_v2() which with this patch will go through `image::CreateRequest`).

I'll use `LOCAL_MIRROR_UUID` directly. Thanks.

src/librbd/journal/Create.cc
vshankar: Hmm.. In that case wouldn't it be beneficial to pass the same `journaler` instance to `journal::RemoveRequest` (which `journal::CreateRequest` already got)?

dillaman: Yes -- it would be nice since you know it's been properly initialized and hasn't been shut-down yet. But for the sake of simplicity, on the error path, an extra round-trip to re-open the journal to remove it probably isn't too bad for the added simplicity of not having to manage the life-cycle / initialization state of journal::Journaler outside of the state machines. Just my 2 cents.

dillaman:
@vshankar Can Journal::ThreadPoolSingleton be yanked now?  I would think everything is async now.

vshankar:
@dillaman That's still needed to instantiate `ContextWQ` as various places (call to `create_v2`, `journal::CreateJournalRequest`, etc..).

src/librbd/journal/CreateJournalRequest.cc
```diff
@@ -0,0 +1,235 @@
...
+Context *CreateJournalRequest<I>::handle_journaler_shutdown(int *result) {
+  ldout(m_cct, 20) << __func__ << ": r=" << *result << dendl;
+
+  if (*result < 0) {
+    lderr(m_cct) << "failed to shut down journaler: " << cpp_strerror(*result) << dendl;
+  }
+
+  delete m_journaler;
+
+  if ((m_r_saved == 0) || (*result < 0)) {
```
vshankar: @dillaman Here, the cleanup cuts short if an error in encountered when shutting down journaler. Would it makes sense to continue deleting journal objects in this case?

dillaman: @vshankar Yeah, I can't think of a reason why not to continue on.

src/librbd/CMakeLists.txt
```diff
@@ -30,12 +30,15 @@ set(librbd_internal_srcs
...
   exclusive_lock/StandardPolicy.cc
   image/CloseRequest.cc
+  image/CreateRequest.cc
   image/OpenRequest.cc
   image/RefreshParentRequest.cc
   image/RefreshRequest.cc
   image/SetSnapRequest.cc
   image_watcher/Notifier.cc
   image_watcher/NotifyLockOwner.cc
+  journal/RemoveJournalRequest.cc
```
dillaman: Minor: since these are in the journal namespace, they could be shortened to just "CreateRequest" and "RemoveRequest" just like all the items in the image, exclusive_lock, etc namespaces.

vshankar:
@ceph-jenkins retest this please

vshankar:
Change in this patchset:
- Renamed `journal::Create/RemoveJournalRequest` to `journal::Create/RemoveReequest`
- Continue removing jounal on shutdown failure

src/librbd/journal/RemoveRequest.cc
```diff
@@ -0,0 +1,155 @@
...
+
+  ImageCtx::get_timer_instance(m_cct, &m_timer, &m_timer_lock);
+  m_journaler = new Journaler(m_op_work_queue, m_timer, m_timer_lock,
+                              m_ioctx, m_image_id, m_image_client_id,
+                              m_cct->_conf->rbd_journal_commit_age);
+
+  using klass = RemoveRequest<I>;
+  Context *ctx = create_context_callback<klass, &klass::handle_init_journaler>(this);
+
+  m_journaler->init(ctx);
```
dillaman: Check for exist before initializing the journal.

src/librbd/journal/RemoveRequest.cc
```diff
@@ -0,0 +1,155 @@
...
+}
+
+template<typename I>
+Context *RemoveRequest<I>::handle_init_journaler(int *result) {
+  ldout(m_cct, 20) << __func__ << ": r=" << *result << dendl;
+
+  if (*result < 0) {
```
dillaman: Can proceed to removal if result == -ENOENT since that might imply the client id isn't registered but the journal still exists.

src/librbd/journal/RemoveRequest.cc
vshankar: I kind of relied on `JournalMetadata::init` which already calls `aio_watch`* on the (header) object which should return `-ENOENT`. Shouldn't that suffice?

[*] https://github.com/ceph/ceph/blob/master/src/journal/JournalMetadata.cc#L449

dillaman: -ENOENT can also be returned if the client isn't registered. If you stat it first and get -ENOENT, you can silently exit the state machine. Otherwise, you can continue to init and remove.

src/librbd/journal/RemoveRequest.cc
vshankar: Hmmm.. I cannot see where does the check for registered clients done (in `init`) -- I might be missing this altogether.

dillaman: https://github.com/ceph/ceph/blob/master/src/journal/JournalMetadata.cc#L769 -- but also notice that the original code ran stat first and then initialized: https://github.com/ceph/ceph/blob/master/src/librbd/Journal.cc#L420

vshankar:
@ceph-jenkins retest this please

dillaman:
@vshankar Passed a QA test suite run -- ready to merge if you tell me you are finished.

vshankar:
@dillaman I resolved a conflict with the master branch due to commit 4de7c8d ("librbd: prevent creation of v2 image ids that are too large"). I moved the required change to commit 1c82132 ("librbd: helper to generate unique image id). Rest should be good.

PR #9586, opened 2016-06-08
client: add check for _lseek() during _write() process

xiexingguo:
The _lseek() call may fail, which can cause uncertain
behaviour if the process goes on instead of returning
immediately.


jcsp:
Reviewed-by

jcsp:

PR #9588, opened 2016-06-08
mds: Ignore create_root_inode's access value

renhwztetecs:
add (void), Ignore create_root_inode's access value

### DIFF
--- a/src/mds/MDSRank.cc
+++ b/src/mds/MDSRank.cc
@@ -954,7 +954,7 @@ void MDSRank::boot_start(BootStep step, int r)
           mdcache->open_root_inode(gather.new_sub());
         } else {
           // replay.  make up fake root inode to start with
-          mdcache->create_root_inode();
+          (void)mdcache->create_root_inode();
         }
         gather.activate();
       }


jcsp:
Can you say exactly how to reproduce the issue this fixes?  I don't quite understand.

renhwztetecs:
@jcsp 
thanks for you reviews :)
I understand the logic of the realization of the purpose is to open a inode root, if this root is not in this MDS, you need to virtual a inode root, that is, to open a new inode root to return to inode root
Before modifying the code, the inode root was created, but not a reference.
I do not know if I explain the，thanks！

jcsp:
I'm not sure what you mean by "but not a reference" -- what kind of reference was missing (and what problem did it cause?)

src/mds/MDSRank.cc
```diff
@@ -954,7 +954,9 @@ void MDSRank::boot_start(BootStep step, int r)
           mdcache->open_root_inode(gather.new_sub());
         } else {
           // replay.  make up fake root inode to start with
-          mdcache->create_root_inode();
```
renhwztetecs: @jcsp 
"but not a reference" mean:
create root inode, but do not use

jcsp:
The created inode added to the inode map and assigned to MDCache::root via create_root_inode->create_system_inode->add_inode

renhwztetecs:
oh,yes
but do not use the following settings, if  do not use the following settings, then why not directly call create_system_inode ()?
I think it is needed to get root inode,not system_inode :)

`i->inode.uid = g_conf->mds_root_ino_uid;
i->inode.gid = g_conf->mds_root_ino_gid;
i->inode.layout = default_file_layout;
i->inode.layout.pool_id = mds->mdsmap->get_first_data_pool();`

PR #9589, opened 2016-06-08
rgw:add a s3 API of make torrent for a object

zhouruisong:
when you put a object, a torrent data of the object will be produced that saved in omap.When you execute the command gettorrent of a object, the torrent data of the object will be return from omap. The torrent data also be removed when the object is removed.
Signed-off-by: zhouruisong <236131368@qq.com>

### DIFF
--- a/src/common/config_opts.h
+++ b/src/common/config_opts.h
@@ -1457,6 +1457,14 @@ OPTION(rgw_list_bucket_min_readahead, OPT_INT, 1000) // minimum number of entrie
 OPTION(mutex_perf_counter, OPT_BOOL, false) // enable/disable mutex perf counter
 OPTION(throttler_perf_counter, OPT_BOOL, true) // enable/disable throttler perf counter
 
+/* The following are tunables for torrent data */
+OPTION(rgw_torrent_tracker, OPT_STR, "")    // torrent field annouce and annouce list
+OPTION(rgw_torrent_createby, OPT_STR, "")    // torrent field created by
+OPTION(rgw_torrent_comment, OPT_STR, "")    // torrent field comment
+OPTION(rgw_torrent_encoding, OPT_STR, "")    // torrent field encoding
+OPTION(rgw_torrent_origin, OPT_STR, "")    // torrent origin
+OPTION(rgw_torrent_sha_unit, OPT_INT, 512*1024)    //torrent field piece length 521K
+
 // This will be set to true when it is safe to start threads.
 // Once it is true, it will never change.
 OPTION(internal_safe_to_start_threads, OPT_BOOL, false)
--- a/src/rgw/rgw_op.cc
+++ b/src/rgw/rgw_op.cc
@@ -1234,6 +1234,7 @@ int RGWGetObj::get_data_cb(bufferlist& bl, off_t bl_ofs, off_t bl_len)
     gc_invalidate_time = start_time;
     gc_invalidate_time += (s->cct->_conf->rgw_gc_obj_min_wait / 2);
   }
+
   return send_response_data(bl, bl_ofs, bl_len);
 }
 
@@ -1332,6 +1333,28 @@ void RGWGetObj::execute()
   op_ret = read_op.prepare(&new_ofs, &new_end);
   if (op_ret < 0)
     goto done_err;
+
+  /* start gettorrent */
+  if (torrent.get_flag())
+  {
+    torrent.init(s, store);
+    torrent.get_torrent_file(op_ret, read_op, total_len, bl, obj);
+    if (op_ret < 0)
+    {
+      ldout(s->cct, 0) << "ERROR: failed to get_torrent_file ret= " << op_ret
+                       << dendl;
+      goto done_err;
+    }
+    op_ret = send_response_data(bl, 0, total_len);
+    if (op_ret < 0)
+    {
+      ldout(s->cct, 0) << "ERROR: failed to send_response_data ret= " << op_ret 
+                       << dendl;
+      goto done_err;
+    }
+    return;
+  }
+  /* end gettorrent */
 
   attr_iter = attrs.find(RGW_ATTR_USER_MANIFEST);
   if (attr_iter != attrs.end() && !skip_manifest) {
@@ -2676,6 +2699,9 @@ void RGWPutObj::execute()
       len = data.length();
     }
 
+    /* save data for producing torrent data */
+    torrent.save_data(data_in);
+
     /* do we need this operation to be synchronous? if we're dealing with an object with immutable
      * head, e.g., multipart object we need to make sure we're the first one writing to this object
      */
@@ -2828,6 +2854,18 @@ void RGWPutObj::execute()
 
   op_ret = processor->complete(etag, &mtime, real_time(), attrs, delete_at,
 			      if_match, if_nomatch);
+  /* produce torrent */
+  if (ofs == torrent.get_data_len())
+  {
+    torrent.init(s, store);
+    torrent.set_create_date(mtime);
+    op_ret =  torrent.handle_data();
+    if (0 != op_ret)
+    {
+      ldout(s->cct, 0) << "ERROR: torrent.handle_data() returned " << op_ret << dendl;
+      goto done;
+    }
+  }
 
 done:
   dispose_processor(processor);
--- a/src/rgw/rgw_op.h
+++ b/src/rgw/rgw_op.h
@@ -34,10 +34,12 @@
 #include "rgw_acl.h"
 #include "rgw_cors.h"
 #include "rgw_quota.h"
+#include "rgw_torrent.h"
 
 #include "include/assert.h"
 
 using namespace std;
+using ceph::crypto::SHA1;
 
 struct req_state;
 class RGWHandler;
@@ -103,6 +105,7 @@ RGWOp() : s(nullptr), dialect_handler(nullptr), store(nullptr),
 
 class RGWGetObj : public RGWOp {
 protected:
+  seed torrent; // get torrent
   const char *range_str;
   const char *if_mod;
   const char *if_unmod;
@@ -642,6 +645,7 @@ class RGWPutObj : public RGWOp {
   friend class RGWPutObjProcessor;
 
 protected:
+  seed torrent;
   off_t ofs;
   const char *supplied_md5_b64;
   const char *supplied_etag;
--- a/src/rgw/rgw_rest.cc
+++ b/src/rgw/rgw_rest.cc
@@ -790,6 +790,19 @@ int RGWGetObj_ObjStore::get_params()
     mod_pg_ver = s->info.env->get_int("HTTP_DEST_PG_VER", 0);
   }
 
+  /* start gettorrent */
+  const char *query_str = s->info.env->get("QUERY_STRING");
+  if (0 == strcmp(query_str, GET_TORRENT))
+  {
+    int ret = 0;
+    ret = torrent.get_params();
+    if (ret < 0)
+    {
+      return ret;
+    }
+  }
+  /* end gettorrent */
+
   return 0;
 }
 
@@ -1000,6 +1013,15 @@ int RGWPutObj_ObjStore::verify_params()
 
 int RGWPutObj_ObjStore::get_params()
 {
+  /* start gettorrent */
+  int ret = 0;
+  ret = torrent.get_params();
+  if (ret < 0)
+  {
+    return ret;
+  }
+  torrent.set_info_name((s->object).name);
+  /* end gettorrent */
   supplied_md5_b64 = s->info.env->get("HTTP_CONTENT_MD5");
 
   return 0;
--- a/src/rgw/rgw_torrent.cc
+++ b/src/rgw/rgw_torrent.cc
@@ -1,0 +1,423 @@
+#include <errno.h>
+#include <stdlib.h>
+
+#include <sstream>
+
+#include "rgw_torrent.h"
+#include "include/rados/librados.hpp"
+
+#define dout_subsys ceph_subsys_rgw
+
+using namespace std;
+using ceph::crypto::MD5;
+using namespace librados;
+using namespace boost;
+using ceph::crypto::SHA1;
+
+seed::seed()
+{
+  seed::info.piece_length = 0;
+  seed::info.len = 0;
+  sha_len = 0;
+  is_torrent = false; 
+}
+
+seed::~seed()
+{
+  seed::info.sha1_bl.clear();
+  bl.clear();
+  torrent_bl.clear();
+  s = NULL;
+  store = NULL;
+}
+
+void seed::init(struct req_state *p_req, RGWRados *p_store)
+{
+  s = p_req;
+  store = p_store;
+}
+
+void seed::get_torrent_file(int &op_ret, RGWRados::Object::Read &read_op, uint64_t &total_len, 
+  bufferlist &bl_data, rgw_obj &obj)
+{
+  string oid, key;
+  rgw_bucket bucket;
+  map<string, bufferlist> m;
+  set<string> obj_key;
+  get_obj_bucket_and_oid_loc(obj, bucket, oid, key);
+  ldout(s->cct, 0) << "NOTICE: head obj oid= " << oid << dendl;
+
+  obj_key.insert(RGW_OBJ_TORRENT);
+  op_ret = read_op.state.io_ctx.omap_get_vals_by_keys(oid, obj_key, &m);
+  if (op_ret < 0)
+  {
+    ldout(s->cct, 0) << "ERROR: failed to omap_get_vals_by_keys op_ret = " << op_ret << dendl;
+    return;
+  }
+
+  map<string, bufferlist>::iterator iter;
+  for (iter = m.begin(); iter != m.end(); ++iter)
+  {
+    bufferlist bl_tmp = iter->second;
+    char *pbuff = bl_tmp.c_str();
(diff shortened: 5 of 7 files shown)


oritwas:
@zhouruisong , can you open a tracker issue for this new feature and it to the commit message
(Fixes: <url to the tracker issue>)
why do you think you need a new pool?

mattbenjamin:
@zhouruisong  I'd like to get feedback from various folks on the different aspects of this, such as the torrent pool concept (e.g., how that might relate to storage classes), and maybe whether this capability should have its own RGWOp.

dang:
So, I'm not liking a separate pool to store torrent files.  How about, instead, the torrent file is generated each time, and the computationally intensive portions of it (such as the SHA) are stored on the primary object as attributes?  That way, when the object is removed it's torrent is removed, and no extra pools or related files are required.

zhouruisong:
@oritwas tracker.ceph.com/issues/16241

zhouruisong:
@oritwas  I need a new torrent pool because I want to save torrent files Individually. it is Acceptable to save pool that has exists. I will modify it in the future.

zhouruisong:
@dang It will cost some time to produce torrent file when you get a object. I test that it cost about two seconds to produce a torrent file. The torrent file's source objcet is about 1.5G. So, I save the torrent file into ceph. if it exists, we wiil get it quickly. I will remove the torrent pool and save torrent file in existed pool. thanks!

zhouruisong:
@dang The  torrent file name and torrent pool name are saved in attributes. torrent file is saved in torrent pool.Do you mean save torrent file in omap?

mattbenjamin:
@zhouruisong I think @dang is suggesting a strategy to materialize checksums/hashes, but not the torrent data;  if someone really wants this kind of space amplification, I don't see how to argue against it for such users, provided there is a durable perf benefit (but that's worth digging more at), but b) I don't think we should force the space amp on everyone

zhouruisong:
@mattbenjamin  The torrent data is only 64k for a object whose capacity is 1.5G. A object whose capacity is 100G only occupy a few megabytes. So I thinks it is will not occupy too mush space for a torrent data of a object. I see your worry. I hope you give me some suggestion  for the solution. thanks！

zhouruisong:
@mattbenjamin  The torrent data is only 64k for a object whose capacity is 1.5G. A object whose capacity is 100G only occupy a few megabytes. So I thinks it is will not occupy too mush space for a torrent data of a object. I see your worry. I hope you give me some suggestion  for the solution. thanks！

src/rgw/rgw_rest.cc
```diff
@@ -779,6 +779,29 @@ void dump_range(struct req_state *s, uint64_t ofs, uint64_t end,
...
+    torrent.set_flag(true);
+    torrent.set_info_piece_length(g_conf->sha_unit);
+    string comment(query_str);
+    torrent.set_comment(g_conf->torrent_comment);
+    torrent.set_create_by(g_conf->torrent_createby);
+    torrent.set_announce(g_conf->torrent_tracker);
+    torrent.set_encoding(g_conf->torrent_encoding);
+    torrent.set_origin(g_conf->torrent_origin);
+    string name = obj.get_orig_obj();
+    torrent.set_info_name(name);
```
cbodley: i think that using `get_orig_obj()` to generate the torrent object names will prevent us from supporting object versioning and object instances. if you overwrite the original object with new data, you'd still find and return the old torrent object for it

i'd lean towards using the omap for this storage so that it's tied to a specific object version/instance, and gets cleaned up automatically when the object is deleted

zhouruisong: thanks, I see what you said and I will modify it

zhouruisong: I need the object name of torrent data beacuse there is a field in torrent file format that is called name。 For example： There is a object named 1.MP4 and when you get its torrent data, a torrent file named 1.MP4.torrent will be produced.But in 1.MP4.torrent, there is a filed name to save the source file. I need get this name when put object and save it into torrent data when producing torrent data. I do not need get the name when get object. If I do not get the object name in geting object, where can I get it? thanks!

zhouruisong:
@dang  I have updated my code according to your suggestion, please help me check thanks!

zhouruisong:
@mattbenjamin  I have updated my code according to your suggestion, please help me check thanks!

zhouruisong:
@cbodley  I have updated my code according to your suggestion, please help me check thanks!

zhouruisong:
@liewegas   I have updated my code according to your suggestion, please help me check thanks!

src/rgw/rgw_rest.cc
cbodley: `get_orig_obj()` is fine for the torrent info name and filename of the .torrent. the problem was only in the object name used to store it in rados

mattbenjamin:
@zhouruisong  lgtm, looking for feedback from @dang

zhouruisong:
@dang I have modified my code according to your suggestion. Now, torrent data has been saved into omap,and it will be removed when the object is removed.

src/rgw/rgw_op.cc
```diff
@@ -1145,9 +1150,500 @@ int RGWGetObj::get_data_cb(bufferlist& bl, off_t bl_ofs, off_t bl_len)
...
+  op_ret = store->omap_set(obj, key, bl);
+  if (op_ret < 0)
+  {
+    INFO "ret = " << op_ret<< dendl;
+    return op_ret;
+  }
+
+  return op_ret;
+}
+
```
cbodley: this seed class belongs in a separate file, like rgw_torrent.cc/h

src/rgw/rgw_op.h
```diff
@@ -36,10 +36,124 @@
 #include "include/assert.h"
 
 using namespace std;
+using ceph::crypto::SHA1;
 
 struct req_state;
 class RGWHandler;
 
+/* get torrent */
+#define INFO     dout(1) <<__FILE__<<":"<<__func__<<":"<<__LINE__<<":"<<
```
cbodley: please take this macro out and use `ldout(s->cct, #)` directly. the file/func/line are unnecessary

src/rgw/rgw_op.cc
```diff
@@ -1145,9 +1150,500 @@ int RGWGetObj::get_data_cb(bufferlist& bl, off_t bl_ofs, off_t bl_len)
...
+    listvalue info_list;
+    info_list.type = BE_LIST;
+    info_list.info =  (*iter);
+    info_list.field = "";
+    list_info.push_back(info_list);
+  }
+  announce_list.clear();
+}
+
+void seed::do_encode()
```
cbodley: i would prefer to see `do_encode()` encode each field directly into the bufferlist, rather than building up a std::list and processing it all at once with `seed_encode()`. there are a lot of allocations associated with std::list that we should be able to avoid here

src/common/config_opts.h
```diff
@@ -166,6 +166,13 @@ OPTION(heartbeat_file, OPT_STR, "")
 OPTION(heartbeat_inject_failure, OPT_INT, 0)    // force an unhealthy heartbeat for N seconds
 OPTION(perf, OPT_BOOL, true)       // enable internal perf counters
 
+OPTION(torrent_tracker, OPT_STR, "")    // torrent_tracker
```
dang: These should probably be prefixed with rgw_ like the other RGW config

src/common/config_opts.h
```diff
@@ -166,6 +166,13 @@ OPTION(heartbeat_file, OPT_STR, "")
...
+OPTION(torrent_createby, OPT_STR, "")    // torrent_createby
+OPTION(torrent_comment, OPT_STR, "")    // torrent_comment
+OPTION(torrent_encoding, OPT_STR, "")    // torrent_encoding
+OPTION(torrent_origin, OPT_STR, "")    // torrent_origin
+OPTION(sha_unit, OPT_INT, 512*1024)    //sha_unit 521K
```
dang: And this one should be rgw_torrent_sha_unit (sha_unit itself it too generic a name)

zhouruisong:
@dang  I have updated my code again according to your advice,please help me check, thanks!

zhouruisong:
@cbodley I have updated my code again according to your advice,please help me check, thanks!

dang:
LGTM

cbodley:
@zhouruisong it's hard to validate the bencoding parts by eye, could you please add some unit tests in src/test/rgw/? just encode each of the basic types, and maybe some combinations of lists/dicts, and compare the resulting bufferlists with their expected output, i.e. `ASSERT_EQ("d3:bar4:spam3:fooi42ee", bl.c_str());`

zhouruisong:
@cbodley ok, I will test it

zhouruisong:
@dang What is the progress of the feature? I have test it and not find any problem

zhouruisong:
@dang What is the progress of the feature? I have test it and not find any problem

dang:
@zhouruisong I'm waiting for unit tests to be added to the PR, as requested by @cbodley

zhouruisong:
@dang  I want to know how to write test code? can you give me a example?

cbodley:
@zhouruisong i see what you mean: all of the encoding logic is tightly coupled to your seed class. moving that stuff into a separate class (or maybe free functions?) would make it easier to write tests against. consider an interface like this, which is similar to Ceph's existing serialization framework:

```c++
// control characters
void bencode_dict(bufferlist& bl) { bl.append('d'); }
void bencode_list(bufferlist& bl) { bl.append('l'); }
void bencode_end(bufferlist& bl) { bl.append('e'); }

// single values
void bencode(int i, bufferlist& bl);
void bencode(const std::string& str, bufferlist& bl);

// dictionary elements
void bencode(const std::string& key, int value, bufferlist& bl);
void bencode(const std::string& key, const std::string& value, bufferlist& bl);
```
Then an example unit test could look like this:
```c++
TEST(Bencode, Dict)
{
  bufferlist bl;

  bencode_dict(bl);
    bencode("foo", 5, bl);
    bencode("bar", "baz", bl);
  bencode_end(bl);

  ASSERT_EQ("d3fooi5e3bar3baze", bl.c_str());
}
```

zhouruisong:
@cbodley Do I run make check to run my unit test? All test will be run, how to make it check my own unit test?

zhouruisong:
@cbodley  I have update my code and add some unit test codes, but when I run make check, I do not konw my unit test have been excuted, please look at my unit test code, I do not know whether the codes in CMakeLists.txt and Makefile-client.am have been executed. Where did I ignore？

src/test/rgw/CMakeLists.txt
```diff
@@ -63,3 +63,29 @@ target_link_libraries(ceph_test_rgw_obj
...
+  global
+  curl
+  uuid
+  expat
+  ${CMAKE_DL_LIBS}
+  ${UNITTEST_LIBS}
+  ${CRYPTO_LIBS}
+  )
+set_target_properties(ceph_test_rgw_bencode PROPERTIES COMPILE_FLAGS
+  ${UNITTEST_CXX_FLAGS})
```
cbodley: this is all you should need here:
```cmake
add_executable(unittest_rgw_bencode test_rgw_bencode.cc)
add_ceph_unittest(unittest_rgw_bencode ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/unittest_rgw_bencode)
target_link_libraries(unittest_rgw_bencode rgw_a)
```
the `add_ceph_unittest()` part is what adds it to make check

src/rgw/rgw_torrent.cc
```diff
@@ -0,0 +1,291 @@
...
+      announce_list.push_back(ann);
+      break;
+    }
+
+    char temp[100] = { 0 };
+    snprintf(temp, location - pstart + 1, "%s", pstart);
+    string anno(temp);
+    announce_list.push_back(anno);
+    pstart = location + 1;
+  }
```
cbodley: please use `get_str_vec()` from "include/str_list.h" here:
```c++
vector<string> announce_list;  // used to get announce list from conf
get_str_vec(announce, ",", announce_list);
```

src/rgw/rgw_torrent.cc
```diff
@@ -0,0 +1,291 @@
...
+
+  list<string>::iterator iter = announce_list.begin();
+  dencode.bencode_key(ANNOUNCE, bl);
+  dencode.bencode_key((*iter), bl);
+  iter++;
```
cbodley: this `iter++` doesn't look right. ANNOUNCE_LIST should contain all of the urls, but you're skipping the first here

src/rgw/rgw_torrent.h
```diff
@@ -0,0 +1,150 @@
...
+    bl.append(info, strlen(info));
+    bencode_end(bl);
+  }
+
+  //single values
+  void bencode(const std::string& str, bufferlist& bl,  bool colon = false) 
+  {
+    int len = str.length();
+    char info[100] = { 0 };
+    if (colon)
```
cbodley: why is the colon optional? this function is otherwise identical to `bencode_key()`

src/test/rgw/CMakeLists.txt
zhouruisong: I have updated my code. but the unittest_rgw_bencode can not generate when I run make check.
I modify add  "#define ENABLE_CLIENT 1" in src/acconfig.h and run make check again. the unittest_rgw_bencode can not generate. I think there are something configrations that I have ignored. I want to know why unittest_rgw_bencode can not be produced? Is it wrong with my code? thanks!

src/rgw/rgw_rest.cc
```diff
@@ -790,6 +790,19 @@ int RGWGetObj_ObjStore::get_params()
     mod_pg_ver = s->info.env->get_int("HTTP_DEST_PG_VER", 0);
   }
 
+  /* start gettorrent */
+  const char *query_str = s->info.env->get("QUERY_STRING");
+  if (0 == strcmp(query_str, GET_TORRENT))
```
cbodley: there may be other params in the query string, right? i think you want to use `s->info.args.exists(GET_TORRENT)` here instead

src/test/Makefile-client.am
```diff
@@ -719,6 +726,15 @@ ceph_test_rgw_obj_LDADD = \
...
 bin_DEBUGPROGRAMS += ceph_test_rgw_obj
 
+ceph_test_rgw_bencode_SOURCES = test/rgw/test_rgw_bencode.cc
+ceph_test_rgw_bencode_LDADD = \ 
+        $(LIBRADOS) $(LIBRGW) $(LIBRGW_DEPS) $(CEPH_GLOBAL) \
+        $(UNITTEST_LDADD) $(CRYPTO_LIBS) \
+        -lcurl -luuid -lexpat
+
+ceph_test_rgw_bencode_CXXFLAGS = $(UNITTEST_CXXFLAGS)
+bin_DEBUGPROGRAMS += ceph_test_rgw_bencode
```
cbodley: note that you've added this block to Makefile-client.am twice

and `bin_DEBUGPROGRAMS` should be `check_TESTPROGRAMS` if you want it to run with make check

the naming convention for these is `unittest_` instead of `ceph_test_`

zhouruisong: @cbodley Is the paramter "#define ENABLE_CLIENT 1" in src/acconfig.h must be added? I think this file acconfig.h is generated when run autogen.sh.

zhouruisong:
@dang  I have finished unit test code.

src/test/rgw/test_rgw_bencode.cc
```diff
@@ -13,7 +13,8 @@ TEST(Bencode, Str)
   decode.bencode("bar", bl);
   decode.bencode("baz", bl);
 
-  ASSERT_EQ("3:foo3:bar3:baz", bl.c_str());
+  const char *p = "3:foo3:bar3:baz";
+  ASSERT_EQ(0, strcmp(p, bl.c_str()));
```
cbodley: oops, ASSERT_STREQ() is what you want for `const char*` comparison (see https://github.com/google/googletest/blob/master/googletest/docs/Primer.md)

zhouruisong: @cbodley I have updated this code, thank you for your advice

src/rgw/rgw_op.cc
```diff
@@ -2581,6 +2604,8 @@ void RGWPutObj::execute()
   map<string, string>::iterator iter;
   bool multipart;
 
+  bufferlist bl_encode;
+
```
cbodley: i don't see this being used anywhere

src/rgw/rgw_torrent.h
```diff
@@ -0,0 +1,139 @@
...
+  void get_torrent_file(int &op_ret, RGWRados::Object::Read &read_op, 
+    uint64_t &total_len, bufferlist &bl_data, rgw_obj &obj);
+  
+  off_t get_data_len();
+  bool get_flag();
+
+  int handle_data();
+  void save_data(bufferlist &bl);
+  void set_create_date(ceph::real_time value);
+  void set_info_name(string value);
```
cbodley: pass the string by `const string&`

zhouruisong:
@cbodley thank you for your advice and I will update my code.

dang:
This is looking good.  @cbodley is out today, but hopefully he can do a final review Monday to make sure all his comments have been addressed.

After that, could you squash all the commits down into a single commit with a good commit message?

dang:
@zhouruisong I had to revert this.  It was causing crashes on normal put.  It appears that this was not the latest version we reviewed.  Did something get messed up in the squash?

PR #9590, opened 2016-06-08
client: fix error process if we fail to choose a target mds

xiexingguo:
The choose_target_mds() always return MDS_RANK_NONE(-1)
if we are currently unable to choose a target mds.
So if it does fail, we'll end up by choosing an invalid
mds, which can't be caught by the current code logic.


jcsp:
Reviewed-by

jcsp:
http://pulpito.ceph.com/jspray-2016-06-14_01:19:46-fs-wip-jcsp-testing-20160610-distro-basic-mira

PR #9592, opened 2016-06-08
librbd: API methods to directly acquire and release the exclusive lock

trociny:


trociny:
@dillaman It is still WIP and serves more as an illustration to the questions I have.

I am not sure how it should coexist with `exclusive-lock` image feature. If `exclusive-lock` is not enabled, then it looks straightforward. But if the feature is enabled and the lock is requested, how safe it is to change the policy, and the back to StandardPolicy, when the lock is released?

E.g. consider the situation: `exclusive-lock` feature is enabled, the image is locked after write is started. Then the user acquires exclusive lock via API, I suppose it should return success. Then the user releases the lock via API, the image is left unlocked while initially it was locked.

Another issue, in theory when the exclusive lock is acquired via API, the policy may differ from Standard. When releasing the lock via API, I suppose the old policy is should be restored?

trociny:
@dillaman So it looks to me it would be (at least) confusing to use exclusive-lock API together with `exclusive-lock` feature enabled. Still `exclusive-lock` feature is needed to enable other useful features.

May be we should add a new feature, something like `manual-exclusive-lock`? It should be used instead (or together with) `exclusive-lock`, and when it is enabled we don't activate/transfer exclusive lock automatically, and any write operation fails with EPERM (EROFS, EINVAL?) until the lock is explicitly acquired by client via API?

dillaman:
@trociny Sorry for the delay getting back to PRs. 

> I am not sure how it should coexist with exclusive-lock image feature. If exclusive-lock is not enabled, then it looks straightforward. But if the feature is enabled and the lock is requested, how safe it is to change the policy, and the back to StandardPolicy, when the lock is released?

With the proper locking of ```owner_lock```, it shouldn't be an issue to swap out the policy.  As an alternative, perhaps we don't swap out StandardPolicy but instead add ```block_release``` and ```unblock_release``` methods to that interface? 

> E.g. consider the situation: exclusive-lock feature is enabled, the image is locked after write is started. Then the user acquires exclusive lock via API, I suppose it should return success. Then the user releases the lock via API, the image is left unlocked while initially it was locked.

Well the good thing is that the IO will force to to automatically re-acquire the lock if needed.  Long term, I'd actually like to see the automatic lock grabbing removed since it was always a workaround for QEMU's lack of locking and the need to support the live-migration case.

> Another issue, in theory when the exclusive lock is acquired via API, the policy may differ from Standard. When releasing the lock via API, I suppose the old policy is should be restored?

For librbd, it should always be StandardPolicy -- but for safety we could add the block/unblock_release methods.

dillaman:
> So it looks to me it would be (at least) confusing to use exclusive-lock API together with exclusive-lock feature enabled. Still exclusive-lock feature is needed to enable other useful features.

Hmm -- I guess I just see this as essentially exclusive-lock w/ the auto-lock acquire disabled. Perhaps I see it in the same vain as our writethrough-until-flush backend tweaks. Before we know that you are properly using the lock API, we treat it as auto-lock.  Maybe even a new config option to disable auto-lock requests?

trociny:
@dillaman  I like your analogy with writethrough-until-flush. I think we don't need (at least for now) new block_release/unblock_release methods then, because we will change the behavior only once, on the first API request, and just changing the Policy should work.

So, the plan could be:

1) Rename Standard policy to AutoPolicy (AutomaticTransferPolicy, AutomaticReleasePolicy?).

2) Provide Standard policy with lock_requested() method being noop (ignore the lock request).

3) Add config option rbd_exclusive_lock_auto_until_requested (default is true).

4) On initialization, if rbd_exclusive_lock_auto_until_requested is false, set policy to Standard. If it is true set to Auto, any update operation will automatically request the lock.

5) When the lock is requested via API and rbd_exclusive_lock_auto_until_requested is true (current policy AutoLock), change the policy to Standard and disable automatic lock request/release on update operations.

What do you think?

trociny:
@dillaman here some working implementation (not sure I found all (and the best) places to hook the checks.

A new configuration option `rbd_exclusive_lock_auto_until_request`  is added.
If the option is true (default), the behavior is as it was previously (auto release) until the exclusive lock is acquired via API. After this it is not automatically released any more, only via API; if an IO operation requires exclusive lock and it is currently released it fails with EINVAL (instead of auto acquiring the lock);  acquiring the lock will block until other client releases the lock.

src/common/config_opts.h
```diff
@@ -1209,6 +1209,7 @@ OPTION(rbd_enable_alloc_hint, OPT_BOOL, true) // when writing a object, it will
 OPTION(rbd_tracing, OPT_BOOL, false) // true if LTTng-UST tracepoints should be enabled
 OPTION(rbd_validate_pool, OPT_BOOL, true) // true if empty pools should be validated for RBD compatibility
 OPTION(rbd_validate_names, OPT_BOOL, true) // true if image specs should be validated
+OPTION(rbd_exclusive_lock_auto_until_request, OPT_BOOL, true) // whether to automatically acquire/release exclusive lock until it is explicitly requested, i.e. before we know the user of librbd is properly using the lock API
```
dillaman: Minor: perhaps ```rbd_auto_exclusive_lock_until_manual_request```?

src/include/rbd/librbd.h
```diff
@@ -299,6 +299,8 @@ CEPH_RBD_API int rbd_set_image_notification(rbd_image_t image, int fd, int type)
 
 /* exclusive lock feature */
 CEPH_RBD_API int rbd_is_exclusive_lock_owner(rbd_image_t image, int *is_owner);
+CEPH_RBD_API int rbd_exclusive_lock_acquire(rbd_image_t image);
```
dillaman: I keep coming back to how we will eventually want to have a unified managed exclusive/shared lock design. How about tweaking these to be something like the following?

```
typedef enum {
  RBD_LOCK_MODE_EXCLUSIVE = 0,
  RBD_LOCK_MODE_SHARED = 1
} rbd_lock_mode_t;

CEPH_RBD_API int rbd_lock_acquire(rbd_image_t image, rbd_lock_mode_t lock_mode);
CEPH_RBD_API int rbd_lock_release(rbd_image_t image);
```

My rationale is that I envision the current ```ExclusiveLock``` and associated namespace being renamed / refactored to support shared locking as well so it would make sense to have a unified API. Internally, you can return -EOPNOTSUPP if the lock mode is shared (for now).

src/librbd/AioImageRequestWQ.cc
```diff
@@ -437,6 +437,17 @@ void AioImageRequestWQ::queue(AioImageRequest<> *req) {
 
   assert(m_image_ctx.owner_lock.is_locked());
   bool write_op = req->is_write_op();
+  bool lock_required = (write_op && is_lock_required()) ||
+    (!write_op && m_require_lock_on_read);
+
+  if (lock_required && m_image_ctx.exclusive_lock_encountered) {
+    lderr(cct) << "op requires exclusive lock" << dendl;
+    req->fail(-EINVAL);
```
dillaman: Minor: -EROFS?

src/librbd/ImageCtx.h
```diff
@@ -70,6 +70,7 @@ namespace librbd {
     // whether the image was opened read-only. cannot be changed after opening
     bool read_only;
     bool flush_encountered;
+    bool exclusive_lock_encountered = false;
```
dillaman: Minor: can we abstract this away into new API method(s) in ```exclusive_lock::Policy``` / ```exclusive_lock::AutomaticReleasePolicy```?

src/librbd/CMakeLists.txt
```diff
@@ -27,6 +27,7 @@ set(librbd_internal_srcs
   Operations.cc
   Utils.cc
   exclusive_lock/AcquireRequest.cc
+  exclusive_lock/AutomaticReleasePolicy.cc
```
dillaman: Minor: perhaps just ```AutomaticPolicy```?

src/librbd/exclusive_lock/StandardPolicy.cc
```diff
@@ -5,15 +5,21 @@
...
 
 void StandardPolicy::lock_requested(bool force) {
   assert(m_image_ctx->owner_lock.is_locked());
   assert(m_image_ctx->exclusive_lock != nullptr);
 
-  // release the lock upon request (ignore forced requests)
-  m_image_ctx->exclusive_lock->release_lock(nullptr);
+  ldout(m_image_ctx->cct, 20) << this << " " << __func__ << ": force=" << force
+			      << dendl;
+
```
dillaman: Might be nice to send a -EROFS error back to the requester of the lock -- otherwise they will keep trying forever and this process will keep ignoring them.

dillaman:
@trociny sorry for the delay --- commented

trociny:
@dillaman updated.

Not sure about a new policy method though, how understandable it would look for other people who read the code. May be it is just due to wrong naming?

dillaman:
@trociny Perhaps "CooporativeReleasePolicy" vs "ManualReleasePolicy"?

trociny:
@dillaman

> Perhaps "CooporativeReleasePolicy" vs "ManualReleasePolicy"?

You are talking about renaming "AutomaticPolicy" to "CooperativeReleasePolicy" and "StandardPolicy" to "ManualReleasePolicy"?

It might be better names. Then may be `rbd_auto_exclusive_lock_until_manual_request` option should be differently named? E.g:
```
 rbd_exclusive_lock_cooperative_release_policy_until_manual_request
```
(thought it is too long...)

But my main concern is not policy names, but a new method `Policy::is_auto()` I have added.

It is used in `librbd::lock_acquire()`:
```
   if (ictx->get_exclusive_lock_policy()->is_auto()) {
     ictx->set_exclusive_lock_policy(
       new exclusive_lock::StandardPolicy(ictx));
   }
```
or, after the policy rename:
```
   if (ictx->get_exclusive_lock_policy()->is_auto()) {
     ictx->set_exclusive_lock_policy(
       new exclusive_lock::ManualReleasePolicy(ictx));
   }
```
and in AioImageRequestWQ::queue():
```
  if (lock_required && !m_image_ctx.get_exclusive_lock_policy()->is_auto()) {
     lderr(cct) << "op requires exclusive lock" << dendl;
     req->fail(-EROFS);
```
I don't very like this. May by, after renaming the config option to `rbd_exclusive_lock_cooperative_release_policy_until_manual_request`, in librbd::lock_acquire() we should rather check that CooperativeReleasePolicy is still used (e.g. by introducing `Policy::type()` method, or using dynamic_cast trick)? So `is_auto()` method will be used only in `AioImageRequestWQ::queue()`, and renamed to something like `may_request_lock()`, `may_auto_request_lock()`, or `may_request_lock_release()`...

dillaman:
@trociny Yeah, I was on the fence about ```is_auto``` as well. I like your suggestion for ```may_request_lock``` or ```may_auto_request_lock```. 

I am fine keeping ```AutomaticPolicy``` now that I know it wasn't your original concern.

trociny:
@dillaman Thanks. So I have just renamed `is_auto()` to `may_auto_request_lock()`.

Still, there is a case where it behaves not as I suppose it is expected (see the comment in TestLibRBD.ExclusiveLock unit test):

When one client opens the image and requests the lock explicitly via API, and then another client opens the image and tries to write, it blocks, though the suggested behavior would be to return -EROFS. It is because `ImageWatcher<I>::handle_request_lock` just reschedules request lock after the error.

Requesting the lock via API (when it is already locked via API by other client) will also block, but this is I think is desirable behavior.

dillaman:
@trociny I think the blocking behavior is fine. Perhaps on the API-side we should have some way to force acquire (i.e. automatically blacklist current lock owner if any) ... probably could wait but would require a new API method when its added.

trociny:
In order not to provide a new API method when adding "force acquire" feature, we could either just add a new mode enum (RBD_LOCK_MODE_EXCLUSIVE_FORCE), or initially use flags instead of mode in our API:
```
typedef enum {
  RBD_LOCK_FLAGS_EXCLUSIVE = 1 << 0,
  RBD_LOCK_FLAGS_SHARED = 1 << 1
} rbd_lock_flags_t;

CEPH_RBD_API int rbd_lock_acquire(rbd_image_t image, rbd_lock_flags_t lock_flags);
CEPH_RBD_API int rbd_lock_release(rbd_image_t image);
```
So RBD_LOCK_FLAGS_FORCE could be added (somewhat similar to libc open(2)). I like the first variant (RBD_LOCK_MODE_EXCLUSIVE_FORCE mode) a little more.

@dillaman What do you think?

dillaman:
@trociny +1 to sticking with the enum. Does this still need the [RFC] prefix?

trociny:
@dillaman I have removed [RFC] prefix.

PR #9593, opened 2016-06-08
rgw/basic_type: use the standard usage of string::npos

Yan-waller:
string::find() returns a value of size_t type, we'd better
use the std::string::npos when it find nothing.

### DIFF
--- a/src/rgw/rgw_basic_types.h
+++ b/src/rgw/rgw_basic_types.h
@@ -52,8 +52,8 @@ struct rgw_user {
   }
 
   void from_str(const std::string& str) {
-    ssize_t pos = str.find('$');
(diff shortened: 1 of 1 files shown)


src/rgw/rgw_basic_types.h
```diff
@@ -52,8 +52,8 @@ struct rgw_user {
   }
 
   void from_str(const std::string& str) {
-    ssize_t pos = str.find('$');
-    if (pos >= 0) {
+    size_t pos = str.find('$');
+    if (pos != std::string::npos) {
```
tchaikov: could also put
```c++
if (pos != str.npos) {
```

mattbenjamin:
this looks fine to me

PR #9595, opened 2016-06-08
librados: get_pool_name() should not return ref of a local variable

tchaikov:

### DIFF
--- a/src/include/rados/librados.hpp
+++ b/src/include/rados/librados.hpp
@@ -1120,7 +1120,7 @@ namespace librados
     int cache_pin(const std::string& o);
     int cache_unpin(const std::string& o);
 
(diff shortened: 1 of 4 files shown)


tchaikov:
test this please.

gregsfortytwo:
Are we allowed to change the signature on this function? It's part of the public interface

jdurgin:
to not break the ABI in jewel (assuming we want to backport this) I think we should add get_pool_name2() instead, and mark the original deprecated

jdurgin:
nevermind, @athanatos points out this was just introduced recently, and works in jewel.

So I'm fine with this fix.

xiexingguo:
@tchaikov Thanks!

PR #9597, opened 2016-06-08
cls_rbd: encode entity_addr_t with features

dillaman:
mirror_image_status_get method call currently fails with EIO


dillaman:
master branch rbd-mirror is currently broken due to the encoding change

trociny:
lgtm

PR #9598, opened 2016-06-08
rbd-mirror: refresh image after creating sync point

dillaman:


dillaman:
@trociny this should be a safe change if you are OK with it -- hoping to get it into v10.2.2

trociny:
lgtm

PR #7374, opened 2016-01-27
WIP:  common: parse according to code_env and alias format for rados-cli

theanalyst:
Not completely sure whether this is the way yet.. here for review

The force option seems to be unused at the moment, removing it, also
this allows us to use the short flag alias `-f` for format, however since `-f` was already used by `foreground` by global_init this could be honored by moving global init after we parse arguments?

_EDIT_ Make description less confusing than it already is

PR #9599, opened 2016-06-08
added -f flag for formatting to rados command line tool

sdieffen:
Added -f flag for formatting to rados command line tool. 

Fixes: http://tracker.ceph.com/issues/15904

Signed-off-by: Steven Dieffenbach <sdieffen@redhat.com>

### DIFF
--- a/src/tools/rados/rados.cc
+++ b/src/tools/rados/rados.cc
@@ -41,6 +41,7 @@ using namespace libradosstriper;
 #include <stdexcept>
 #include <climits>
 #include <locale>
+#include <memory>
 
 #include "cls/lock/cls_lock_client.h"
 #include "include/compat.h"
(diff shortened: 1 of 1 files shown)


athanatos:
The Fixes line seem to be in the wrong place.  You need a blank line between the first line (which becomes the short description in gitk etc) and the rest of the commit message.

src/tools/rados/rados.cc
```diff
@@ -3284,12 +3285,37 @@ int main(int argc, const char **argv)
...
 
-  global_init(NULL, args, CEPH_ENTITY_TYPE_CLIENT, CODE_ENVIRONMENT_UTILITY, 0);
-  common_init_finish(g_ceph_context);
-
   std::map < std::string, std::string > opts;
   std::vector<const char*>::iterator i;
+  std::vector<const char*>::iterator j;
   std::string val;
+
+  /* Necessary to support usage of -f for formatting,
```
tchaikov: could use C++ style comment like 
```
// Necessary to...
```

src/tools/rados/rados.cc
```diff
@@ -3284,12 +3285,37 @@ int main(int argc, const char **argv)
...
+   * since global_init will remove the -f using ceph
+   * argparse procedures. */
+  for (j = args.begin(); j != args.end(); ) {
```
tchaikov: might want to limit `j` lexical scope in the `for` loop, like
```
for (auto j = args.begin(); j != args.end(); ++j)
```

src/tools/rados/rados.cc
```diff
@@ -3284,12 +3285,37 @@ int main(int argc, const char **argv)
...
+      unique_ptr<Formatter> formatter(Formatter::create(val.c_str()));
+      if (formatter) {
+	j = args.erase(j);
+	opts["format"] = val;
+	
+	j = args.erase(j);
+	break;
+      }
+    }
+    ++j;
```
tchaikov: could move this into the `for` statement.

src/tools/rados/rados.cc
```diff
@@ -3284,12 +3285,37 @@ int main(int argc, const char **argv)
   argv_to_vec(argc, argv, args);
   env_to_vec(args);
 
-  global_init(NULL, args, CEPH_ENTITY_TYPE_CLIENT, CODE_ENVIRONMENT_UTILITY, 0);
-  common_init_finish(g_ceph_context);
-
   std::map < std::string, std::string > opts;
   std::vector<const char*>::iterator i;
+  std::vector<const char*>::iterator j;
```
tchaikov: could move the definition of `i` down to where it is used for the first time.

tchaikov:
@sdieffen also you may need to squash these two commit into a single one.

theanalyst:
@tchaikov this was also reported at http://tracker.ceph.com/issues/14501 (pr https://github.com/ceph/ceph/pull/7374/commits) but back then we were discussing not to defer global_init later.. but if we are are doing it, then I agree this patch is way easier than modifying global init and friends

PR #9601, opened 2016-06-09
cmake: add ceph_test_* used by qa units and ceph-qa-suite tasks

tchaikov:


cbodley:
@tchaikov unrelated to this PR, but I'm curious about these `CMAKE_INSTALL_` variables from [GNUInstallDirs](https://cmake.org/cmake/help/v3.0/module/GNUInstallDirs.html). are there important cases where `CMAKE_INSTALL_BINDIR` isn't `bin`? should we prefer the `CMAKE_INSTALL_` form when installing to lib, man, doc, etc? either way, let's add this to the TODO list so that we're consistent about it (a grep for `DESTINATION bin` still shows 40+ matches)

tchaikov:
@cbodley, 

> are there important cases where `CMAKE_INSTALL_BINDIR` isn't `bin`?

to my knowledge, no.

> should we prefer the `CMAKE_INSTALL_` form when installing to lib, man, doc, etc?

in the case of lib, yes. when it comes to multi-arch support, fedora, rhel and SuSE comply to FHS, but  Debian does not. that's why i prefer `$CMAKE_INSTALL_LIBDIR` over the plain `lib`.

in the other case, i tend to be more consistent, so use `CMAKE_INSTALL_*` when possible.

wjwithagen:
On 9-6-2016 16:32, Kefu Chai wrote:
> @cbodley <https://github.com/cbodley>,
> 
>     are there important cases where |CMAKE_INSTALL_BINDIR| isn't |bin|?
> 
> to my knowledge, no.
> 
>     should we prefer the |CMAKE_INSTALL_| form when installing to lib,
>     man, doc, etc?
> 
> in the case of lib, yes. when it comes to multi-arch support, fedora,
> rhel and SuSE comply to FHS, but Debian does not. that's why i prefer
> |$CMAKE_INSTALL_LIBDIR| over the plain |lib|.
> 
> in the other case, i tend to be more consistent, so use
> |CMAKE_INSTALL_*| when possible.

How is this going to work out for FreeBSD where sort of everything is
expected to go into /usr/local/{bin,etc,lib,libexec,....}

If this all could be prefixed in one go, by setting something like
|INSTALL_ROOT| that would be great.

--WjW

cbodley:
> If this all could be prefixed in one go, by setting something like |INSTALL_ROOT| that would be great.

`CMAKE_INSTALL_PREFIX` does this. you can set it on the cmake command line, i.e.
`cmake -DCMAKE_INSTALL_PREFIX=/usr/local`

wjwithagen:
On 9-6-2016 17:11, Casey Bodley wrote:
>     If this all could be prefixed in one go, by setting something like
>     |INSTALL_ROOT| that would be great.
> 
> |CMAKE_INSTALL_PREFIX| does this. you can set it on the cmake command
> line, i.e.
> |cmake -DCMAKE_INSTALL_PREFIX=/usr/local|

Super, that is going to help a lot.

tchaikov:
@alimaredia ping?

PR #9603, opened 2016-06-09
Erasure Coding MOVE API for filestore

manali26:
Signed-off-by: Manali Kulkarni <Manali.Kulkarni@sandisk.com>

### DIFF
--- a/src/ceph.in
+++ b/src/ceph.in
@@ -81,13 +81,13 @@ if MYDIR.endswith('src') and \
 
     python_libpath = os.path.join(MYDIR, 'build', get_pythonlib_dir())
     respawn_in_path(os.path.join(MYDIR, '.libs'), 'pybind', python_libpath)
-    if os.environ.has_key('PATH') and MYDIR not in os.environ['PATH']:
+    if 'PATH' in os.environ and MYDIR not in os.environ['PATH']:
         os.environ['PATH'] += ':' + MYDIR
 
 elif os.path.exists(os.path.join(os.getcwd(), "CMakeCache.txt")) \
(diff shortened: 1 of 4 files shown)


manali26:
MOVE API for Erasure Coding:
Background:
Delayed commit in 2PC will produce multiple modified extends of base object which need to be moved in to appropriate offsets in base object and delete temporary extends. For this reason, "move" api need to be implemented. The callers need to pass this information in below mentioned format.

Move API Logic:
Say, over n intervals, we have n modified extents of the base object. 
Instead of having n temporary objects, this is represented in a single temporary object. This is done using vector of tuple <src offset, dest offset, length> move_info. This move_info has information about which offset in base object needs copy from temp object offset and for what length.
The logic is simple, once move api is called, we iterate through this vector of tuple move_info, and call clone_range() api for each <src offset, dest offset, length> tuple. Once all the ranges are cloned, the temporary object is deleted. And we return.
Move API Implementation:
ECBackend will call the Move API via objectstore transaction. Currently, only filestore implementation is done. Bluestore and other objectstore implementations will be taken up later.
The core function looks like this:
int move_tempobjs(const coll_t& cid, const ghobject_t& tempoid, const ghobject_t& newoid, const vector<boost::tuple<uint64_t, uint64_t, uint64_t> > move_info);
Inputs :
cid : Collection id 
tempoid: The object ID for temporary object, which has the modified contents.
Newoid: new or base object which will take all the modifications from temp object.
Move_info: This is vector of tuple of tuple <src offset, dest offset, length> . This gives the information about what/where needs to be changed in base object.
Output:
Temp object is deleted. Base object [newoid] has all the modifications which were recorded in temp object.

PR #9604, opened 2016-06-09
os/bluestore: add an option to limit max allocation size

ifed01:
…marily for testing. Extends ceph_test_objectstore  with cases for fragmented blob testing.

### DIFF
--- a/src/os/bluestore/BlueStore.cc
+++ b/src/os/bluestore/BlueStore.cc
@@ -1008,6 +1008,7 @@ BlueStore::BlueStore(CephContext *cct, const string& path)
     logger(NULL),
     csum_type(bluestore_blob_t::CSUM_CRC32C),
     min_alloc_size(0),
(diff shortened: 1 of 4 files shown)


src/os/bluestore/BlueStore.cc
```diff
@@ -1317,7 +1320,8 @@ int BlueStore::_open_bdev(bool create)
     ++block_size_order;
   }
 
-  _set_min_alloc();
+  _set_alloc_sizes();
+  max_alloc_size = g_conf->bluestore_max_alloc_size;
```
liewegas: don't need this line i think?

liewegas:
does it need to be a power of 2?  the comment in the header says yes, but the value set in store_test.cc is not

liewegas:
otherwise lgtm!

ifed01:
@liewegas - fixed

PR #9605, opened 2016-06-09
os: simplify store_statfs_t

liewegas:
This PR also fixes ceph_test_objectstore not to muck around with plugins (copying them around! yeesh). 

I'm not sure about the cmake vstart case.  And I'm not sure who you're "supposed" to run vstart and tests with cmake.

Can someone please update README.md with this information?  I'll document the automake way (short-lived though it is).


liewegas:
@ifed01 thoughts?

ifed01:
Perfect!

liewegas:
will rebase after https://github.com/ceph/ceph/pull/9607 goes it

liewegas:
@jcsp @alimaredia do you mind documenting how to run these tests with cmake?  and check my vstart.sh commit?  I don't understand why the plug_dir option was getting set to $CEPH_LIB/compressor when ceph looks in $plugin_dir/compressor already.

PR #9606, opened 2016-06-09
ObjectCacher: fix bh_read_finish offset logic

gregsfortytwo:
If we have an incoming read split across multiple BufferHeads, we want to
line up the BufferHead's bl with the incoming OSDOp's bl at the right offset. We
were erroneously using this nonsense calculation (always equal to zero!) when
a much simpler comparison of the BufferHead's logical object offset to the
incoming OSDOp's logical offset will do the trick nicely.

Fixes: http://tracker.ceph.com/issues/16002

### DIFF
--- a/qa/workunits/osdc/stress_objectcacher.sh
+++ b/qa/workunits/osdc/stress_objectcacher.sh
@@ -14,7 +14,7 @@ do
                     do
                         for MAX_DIRTY in 0 25165824
                         do
-                            ceph_test_objectcacher_stress --ops $OPS --percent-read $READS --delay-ns $DELAY --objects $OBJECTS --max-op-size $OP_SIZE --client-oc-max-dirty $MAX_DIRTY > /dev/null 2>&1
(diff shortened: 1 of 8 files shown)


gregsfortytwo:
This was tested in situ and functions correctly. I'm working on a test for it but the existing ObjectCacher testing framework was insufficient (it doesn't actually store any data so you can't test data-driven failures like this) and doing so ballooned a bit more than I had planned for.

liewegas:
Reviewed-by:

gregsfortytwo:
This now has a test! If you run it without the ObjectCacher patch, it fails; and the test is included as part of the stress_objectcacher.sh script so it will automatically get tested in the RADOS teuthology suite.

jdurgin:
lgtm, if we add more tests like this in the future it'd be nice to use gtest for them, but for just 2 tests this looks good

jcsp:
@gregsfortytwo Did this go into a rados suite run yet?

jdurgin:
Looks like only two tests in the rados suite would use the objectcacher - singleton-nomsgr/all/cache-fs-trunc.yaml and objectstore/objectcacher-stress.yaml.

The rbd and fs suites will get more coverage.

gregsfortytwo:

The rbd suite is still running.

gregsfortytwo:
Jason tells me http://pulpito.ceph.com/gregf-2016-06-27_22:28:07-rbd-greg-fs-testing-627---basic-mira/ is a fine result right now.

PR #9613, opened 2016-06-09
common osd: Improve scrub analysis, list-inconsistent-obj output and osd-scrub-repair test

dzafman:
Improve scrub mechanism to check all shards against object info
Object errors are separate from shard errors
osd: persist the user_version and shard id of scrubbed obj
tools/rados: dump inconsistent obj's version and shard-id
Create list-inconsistent-\* schemas in doc
Add test cases in osd-scrub-repair test

### DIFF
--- a/doc/rados/command/list-inconsistent-obj.json
+++ b/doc/rados/command/list-inconsistent-obj.json
@@ -1,0 +1,185 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "properties": {
+    "epoch": {
+      "description": "Scrub epoch",
+      "type": "integer"
+    },
+    "inconsistents": {
+      "type": "array",
+      "items": {
+        "type": "object",
+        "properties": {
+          "object": {
+            "description": "Identify a Ceph object",
+            "type": "object",
+            "properties": {
+              "name": {
+                "type": "string"
+              },
+              "nspace": {
+                "type": "string"
+              },
+              "locator": {
+                "type": "string"
+              },
+              "version": {
+                "type": "integer",
+                "minimum": 0
+              },
+              "snap": {
+                "oneOf": [
+                  {
+                    "type": "string",
+                    "enum": [ "head", "snapdir" ]
+                  },
+                  {
+                    "type": "integer",
+                    "minimum": 0
+                  }
+                ]
+              }
+            },
+            "required": [
+              "name",
+              "nspace",
+              "locator",
+              "version",
+              "snap"
+            ]
+          },
+          "selected_object_info": {
+              "type": "string"
+          },
+          "union_shard_errors": {
+            "description": "Union of all shard errors",
+            "type": "array",
+            "items": {
+              "enum": [
+                "missing",
+                "stat_error",
+                "read_error",
+                "data_digest_mismatch_oi",
+                "omap_digest_mismatch_oi",
+                "size_mismatch_oi",
+                "ec_hash_error",
+                "ec_size_error",
+                "oi_attr_missing",
+                "oi_attr_corrupted"
+              ]
+            },
+            "minItems": 0,
+            "uniqueItems": true
+          },
+          "errors": {
+            "description": "Errors related to the analysis of this object",
+            "type": "array",
+            "items": {
+              "enum": [
+                "object_info_inconsistency",
+                "data_digest_mismatch",
+                "omap_digest_mismatch",
+                "size_mismatch",
+                "attr_value_mismatch",
+                "attr_name_mismatch"
+              ]
+            },
+            "minItems": 0,
+            "uniqueItems": true
+          },
+          "shards": {
+            "description": "All found or expected shards",
+            "type": "array",
+            "items": {
+              "description": "Information about a particular shard of object",
+              "type": "object",
+              "properties": {
+                "object_info": {
+                  "type": "string"
+                },
+                "shard": {
+                  "type": "integer"
+                },
+                "osd": {
+                  "type": "integer"
+                },
+                "size": {
+                  "type": "integer"
+                },
+                "omap_digest": {
+                  "description": "Hex representation (e.g. 0x1abd1234)",
+                  "type": "string"
+                },
+                "data_digest": {
+                  "description": "Hex representation (e.g. 0x1abd1234)",
+                  "type": "string"
+                },
+                "errors": {
+                  "description": "Errors with this shard",
+                  "type": "array",
+                  "items": {
+                    "enum": [
+                      "missing",
+                      "stat_error",
+                      "read_error",
+                      "data_digest_mismatch_oi",
+                      "omap_digest_mismatch_oi",
+                      "size_mismatch_oi",
+                      "ec_hash_error",
+                      "ec_size_error",
+                      "oi_attr_missing",
+                      "oi_attr_corrupted"
+                    ]
+                  },
+                  "minItems": 0,
+                  "uniqueItems": true
+                },
+                "attrs": {
+                  "description": "If any shard's attr error is set then all attrs are here",
+                  "type": "array",
+                  "items": {
+                    "description": "Information about a particular shard of object",
+                    "type": "object",
+                    "properties": {
+                      "name": {
+                        "type": "string"
+                      },
+                      "value": {
+                        "type": "string"
+                      },
+                      "Base64": {
+                        "type": "boolean"
+                      }
+                    },
+                    "required": [
+                      "name",
+                      "value",
+                      "Base64"
+                    ],
+                    "additionalProperties": false,
+                    "minItems": 1
+                  }
+                }
+              },
+              "required": [
+                "osd",
+                "errors"
+              ]
+            }
+          }
+        },
+        "required": [
+          "object",
+          "union_shard_errors",
+          "errors",
+          "shards"
+        ]
+      }
+    }
+  },
+  "required": [
+    "epoch",
+    "inconsistents"
+  ]
+}
--- a/src/common/scrub_types.cc
+++ b/src/common/scrub_types.cc
@@ -27,6 +27,29 @@ static void encode(const object_id_t& obj, bufferlist& bl)
   reinterpret_cast<const object_id_wrapper&>(obj).encode(bl);
 }
 
+void osd_shard_wrapper::encode(bufferlist& bl) const
+{
+  ENCODE_START(1, 1, bl);
+  ::encode(osd, bl);
+  ::encode(shard, bl);
+  ENCODE_FINISH(bl);
+}
+
+void osd_shard_wrapper::decode(bufferlist::iterator& bp)
+{
+  DECODE_START(1, bp);
+  ::decode(osd, bp);
+  ::decode(shard, bp);
+  DECODE_FINISH(bp);
+}
+
+namespace librados {
+  static void encode(const osd_shard_t& shard, bufferlist& bl)
+  {
+    reinterpret_cast<const osd_shard_wrapper&>(shard).encode(bl);
+  }
+}
+
 void shard_info_wrapper::set_object(const ScrubMap::object& object)
 {
   for (auto attr : object.attrs) {
@@ -43,17 +66,11 @@ void shard_info_wrapper::set_object(const ScrubMap::object& object)
     data_digest_present = true;
     data_digest = object.digest;
   }
-  if (object.read_error) {
-    errors |= SHARD_READ_ERR;
-  }
-  if (object.stat_error) {
-    errors |= SHARD_STAT_ERR;
-  }
 }
 
 void shard_info_wrapper::encode(bufferlist& bl) const
 {
-  ENCODE_START(1, 1, bl);
+  ENCODE_START(2, 1, bl);
   ::encode(errors, bl);
   if (has_shard_missing()) {
     return;
@@ -64,12 +81,13 @@ void shard_info_wrapper::encode(bufferlist& bl) const
   ::encode(omap_digest, bl);
   ::encode(data_digest_present, bl);
   ::encode(data_digest, bl);
+  ::encode(selected_oi, bl);
   ENCODE_FINISH(bl);
 }
 
 void shard_info_wrapper::decode(bufferlist::iterator& bp)
 {
-  DECODE_START(1, bp);
+  DECODE_START(2, bp);
   ::decode(errors, bp);
   if (has_shard_missing()) {
     return;
@@ -80,6 +98,8 @@ void shard_info_wrapper::decode(bufferlist::iterator& bp)
   ::decode(omap_digest, bp);
   ::decode(data_digest_present, bp);
   ::decode(data_digest, bp);
+  if (struct_v > 1)
+    ::decode(selected_oi, bp);
   DECODE_FINISH(bp);
 }
 
@@ -92,28 +112,28 @@ inconsistent_obj_wrapper::inconsistent_obj_wrapper(const hobject_t& hoid)
 void inconsistent_obj_wrapper::add_shard(const pg_shard_t& pgs,
                                          const shard_info_wrapper& shard)
 {
-  errors |= shard.errors;
-  shards[pgs.osd] = shard;
+  union_shards.errors |= shard.errors;
+  shards.emplace(osd_shard_t{pgs.osd, int8_t(pgs.shard)}, shard);
 }
 
 void
 inconsistent_obj_wrapper::set_auth_missing(const hobject_t& hoid,
-                                           const map<pg_shard_t, ScrubMap*>& maps)
+                                           const map<pg_shard_t, ScrubMap*>& maps,
+					   map<pg_shard_t, shard_info_wrapper> &shard_map,
+					   int &shallow_errors, int &deep_errors)
 {
-  errors |= (err_t::SHARD_MISSING |
-             err_t::SHARD_READ_ERR |
-             err_t::OMAP_DIGEST_MISMATCH |
-             err_t::DATA_DIGEST_MISMATCH |
-             err_t::ATTR_MISMATCH);
   for (auto pg_map : maps) {
     auto oid_object = pg_map.second->objects.find(hoid);
-    shard_info_wrapper shard;
-    if (oid_object == pg_map.second->objects.end()) {
-      shard.set_missing();
-    } else {
-      shard.set_object(oid_object->second);
-    }
-    shards[pg_map.first.osd] = shard;
+    if (oid_object == pg_map.second->objects.end())
+      shard_map[pg_map.first].set_missing();
+    else
+      shard_map[pg_map.first].set_object(oid_object->second);
+    if (shard_map[pg_map.first].has_deep_errors())
+      ++deep_errors;
+    else if (shard_map[pg_map.first].has_shallow_errors())
+      ++shallow_errors;
+    union_shards.errors |= shard_map[pg_map.first].errors;
+    shards.emplace(osd_shard_t{pg_map.first.osd, pg_map.first.shard}, shard_map[pg_map.first]);
   }
 }
 
@@ -126,19 +146,24 @@ namespace librados {
 
 void inconsistent_obj_wrapper::encode(bufferlist& bl) const
 {
-  ENCODE_START(1, 1, bl);
+  ENCODE_START(2, 2, bl);
   ::encode(errors, bl);
   ::encode(object, bl);
+  ::encode(version, bl);
   ::encode(shards, bl);
+  ::encode(union_shards.errors, bl);
   ENCODE_FINISH(bl);
 }
 
 void inconsistent_obj_wrapper::decode(bufferlist::iterator& bp)
 {
-  DECODE_START(1, bp);
+  DECODE_START(2, bp);
+  DECODE_OLDEST(2);
   ::decode(errors, bp);
   ::decode(object, bp);
+  ::decode(version, bp);
   ::decode(shards, bp);
+  ::decode(union_shards.errors, bp);
   DECODE_FINISH(bp);
 }
 
--- a/src/common/scrub_types.cc
+++ b/src/common/scrub_types.cc
@@ -113,7 +113,7 @@ inconsistent_obj_wrapper::set_auth_missing(const hobject_t& hoid,
     } else {
       shard.set_object(oid_object->second);
     }
-      shards[pg_map.first.osd] = shard;
+    shards[pg_map.first.osd] = shard;
   }
 }
 
--- a/src/include/rados/rados_types.hpp
+++ b/src/include/rados/rados_types.hpp
@@ -51,24 +51,24 @@ struct object_id_t {
 };
 
 struct err_t {
-  enum {
-    ATTR_UNEXPECTED      = 1 << 0,
+  enum : uint64_t {
     SHARD_MISSING        = 1 << 1,
     SHARD_STAT_ERR       = 1 << 2,
     SHARD_READ_ERR       = 1 << 3,
-    DATA_DIGEST_MISMATCH = 1 << 4,
-    OMAP_DIGEST_MISMATCH = 1 << 5,
-    SIZE_MISMATCH        = 1 << 6,
-    ATTR_MISMATCH        = 1 << 7,
-    ATTR_MISSING         = 1 << 8,
     DATA_DIGEST_MISMATCH_OI = 1 << 9,
     OMAP_DIGEST_MISMATCH_OI = 1 << 10,
     SIZE_MISMATCH_OI        = 1 << 11,
+    SHARD_EC_HASH_MISMATCH  = 1 << 12,
+    SHARD_EC_SIZE_MISMATCH  = 1 << 13,
+    OI_ATTR_MISSING         = 1 << 14,
+    OI_ATTR_CORRUPTED       = 1 << 15,
+    SS_ATTR_MISSING         = 1 << 16,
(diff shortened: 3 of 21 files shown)


athanatos:
@dzafman I thought we weren't going to fill in that part at all unless we were sure?  It seems like we want two concepts:
1) Mismatch between shards/replicas
  a) mismatched attrs
  b) mismatched metadata (which would also imply mismatched attrs)
  c) for ec objects, mismatched stored hashes
  ...
2) Shard does not match known metadata (object_info or ec object stored hashes)
  a) size does not match accepted oi.size
  b) hash (omap or data) does not match object_info hash
  c) hash does not match locally stored hash (ec object)

1. should be communicated by a field on the whole object.  2. should be communicated by a field on the specific shard and probably by a field on the whole object so that the interface consumer knows to check the individual shards.

doc/rados/command/list-inconsistent-obj.json
```diff
@@ -0,0 +1,140 @@
...
+            "description": "Union of all shard errors, including errors in which the faulty shard is NOT identified",
+            "type": "array",
+            "items": {
+              "enum": [
+                "attr_unexpected",
+                "missing",
+                "stat_error",
+                "read_error",
+                "data_digest_mismatch",
+                "omap_digest_mismatch",
```
athanatos: I think here I'd expect that data_digest_mismatch would be present whenever the shards' data digests do not match (ditto for omap).  For an EC pool, this is a bit more complicated because there are two possible mismatches:
1) for shard <x>, shard <x>'s locally stored hash does match the computed hash (not the object_info hash)
2) shard <x>'s stored hash does not match shard <y>'s stored hash (note, each shard stores all of the hashes for all of the shards)
I think 2 should cause data_digest_mismatch and 1 should have another error type (data_digest_mismatch_ec? or a better name...)

doc/rados/command/list-inconsistent-obj.json
```diff
@@ -0,0 +1,140 @@
...
+                "size_mismatch",
+                "attr_mismatch",
+                "attr_missing",
+                "data_digest_mismatch_oi",
+                "omap_digest_mismatch_oi",
```
athanatos: These two would be replicated only, I think, since we can't verify the object_info hash during scrub (right?).  I think they should be present whenever *some* shard's hash does not match the accepted object_info hash.

doc/rados/command/list-inconsistent-obj.json
```diff
@@ -0,0 +1,140 @@
...
+                  "description": "Errors that have been definitively identified as belonging to this shard",
+                  "type": "array",
+                  "items": {
+                    "enum": [
+                      "attr_unexpected",
+                      "missing",
+                      "stat_error",
+                      "read_error",
+                      "data_digest_mismatch",
+                      "omap_digest_mismatch",
```
athanatos: I don't think these would be valid here.  Either it would have to be present on all shards, or we'd be making a guess as to which are correct.  I don't think that makes sense since the above error set already indicates that there is a mismatch and we have the actual hash values for each shard.

doc/rados/command/list-inconsistent-obj.json
```diff
@@ -0,0 +1,140 @@
...
+                      "size_mismatch",
+                      "attr_mismatch",
+                      "attr_missing",
+                      "data_digest_mismatch_oi",
+                      "omap_digest_mismatch_oi",
```
athanatos: I think these *are* valid here since it's not ambiguous which shards do or do not match the accepted object_info hash.

doc/rados/command/list-inconsistent-obj.json
```diff
@@ -0,0 +1,140 @@
...
+          "shards": {
+            "description": "All found or expected shards",
+            "type": "array",
+            "items": {
+              "description": "Information about a particular shard of object",
+              "type": "object",
+              "properties": {
+                "osd": {
+                  "type": "integer"
+                },
```
athanatos: We need the shard id here as well, right?  A single osd could have two copies of the pg at different shard positions.

doc/rados/command/list-inconsistent-obj.json
```diff
@@ -0,0 +1,140 @@
...
+            },
+            "required": [
+              "name",
+              "nspace",
+              "locator",
+              "snap"
+            ]
+          },
+          "errors": {
+            "description": "Union of all shard errors, including errors in which the faulty shard is NOT identified",
```
athanatos: Also, I don't think this should be the union of all shard errors.  We may want different things here from what we have in the shard errors.

athanatos:
I don't see the "guessed" field showing up in the schema?  Does that mean that that field would be represented in another way?

doc/rados/command/list-inconsistent-obj.json
tchaikov: yeah. we do need it. my fault not including it.

src/test/osd/osd-scrub-repair.sh
```diff
@@ -395,8 +437,66 @@ function TEST_corrupt_scrub_replicated() {
...
+    # Check pgid
+    test $(jq -r '.[0]' $dir/json) = $pg || return 1
+
+    rados list-inconsistent-obj $pg > $dir/json || return 1
+    # Get epoch for repair-get requests
+    epoch=$(jq .epoch $dir/json)
+    # Check object count
     test $(jq '.inconsistents | length' $dir/json) = "$total_objs" || return 1
 
+    jq '.inconsistents | sort' > $dir/checkcsjson << EOF
```
tchaikov: the expected output is hard to read, can we pre-fomat it?

doc/rados/command/list-inconsistent-obj.json
tchaikov: see https://github.com/ceph/ceph/pull/9203/commits/fdaaa301a2829624f97cef50fdf7a396c66aa5fe

src/tools/rados/rados.cc
```diff
@@ -1344,7 +1344,7 @@ static void dump_inconsistent(const inconsistent_obj_t& inc,
     auto& osd_shard = shard_info.first;
     f.dump_int("osd", osd_shard.osd);
     auto shard = osd_shard.shard;
-    if (shard >= 0)
+    if (shard != shard_id_t::NO_SHARD)
```
dzafman: @tchaikov I think your commit should use NO_SHARD here.

dzafman:
@tchaikov I added a commit that changes the schema and adds "version" to existing test code based on your changes.

doc/rados/command/list-inconsistent-obj.json
tchaikov: for example, if we have a replica in replica pool, and its data digest does match the calculated digest, all other replicas' `error` should have `data_digest_mismatch`, right?

tchaikov:
> I thought we weren't going to fill in that part at all unless we were sure? It seems like we want two concepts:

yeah, we can differentiate these two concepts by putting the related error into the global error or into the shard's error field. probably, we can do this in another changeset.

doc/rados/command/list-inconsistent-obj.json
dzafman: @tchaikov No, marking all shards with the same error doesn't provide any additional information.

We have the `errors` for the object which should include all errors across all replicas.  A shard is marked with `errors` only if that shard is definitely the one with that error.

doc/rados/command/list-inconsistent-obj.json
dzafman: We might want some additional errors here, or some errors that are never shard specific.  But for all cases where we know there is an inconsistency, but don't know which shard(s) are correct, we mark the object `errors` with the error.

src/osd/PGBackend.cc
```diff
@@ -447,7 +447,7 @@ enum scrub_error_type PGBackend::be_compare_scrub_objects(
 		  << (known ? "known" : "best guess")
 		  << " omap_digest 0x" << auth.omap_digest << std::dec
 		  << " from auth shard " << auth_shard;
-      result.set_omap_digest_mismatch();
+      result.set_omap_digest_mismatch(known);
```
dzafman: @athanatos Here is where we have `bool know` which indicates whether we are guessing that this is the bad shard or not.  If `know` is true then we mark the shard and object with the error flag.  If `know` is false we only mark the object with the error flag by using `guessed`.

doc/rados/command/list-inconsistent-obj.json
tchaikov: oh, i mean, all **other** replicas except the ones whose digest matches with the known hash (i.e. its object_info). and the chance that the corrupt data conflicts with the corrupted object_info is rare. so i think we are safe in this case.

src/osd/PGBackend.cc
athanatos: Ok, so the guessed field on each shard is or'd with the global object errors to build up the global object errors.  I don't think "guessed" is a good name.  We don't know which shard is wrong, but we aren't guessing that there is a mismatch.  Also, is the only difference between the case where we have no oi digest but the shards to not match and the case where we do have a the oi.digest and the shards do not match that in the latter case at least one shard will have have a digest error?

doc/rados/command/list-inconsistent-obj.json
athanatos: If I'm understanding this right, I really think we should be more descriptive in the top level object errors.  I think that the case where the shards do not match each other should have a distinct error from the case where at least one shard does not match the stored digest.  Either, both, or neither error is possible, and it would simplify life for those consuming the interface.

dzafman:
@athanatos @tchaikov Rebased, changed "guessed" to "objonly"

Added a fix for stat_error scrub handling and added read_error/stat_error testing

@tchaikov I want to squash 1eda77bba3fd183febf789d3f144453df082f96c 68cf66b66ce198820ba13ca77aa6c78c1dc9c524 into one of your commits.

tchaikov:
>  I want to squash 1eda77b 68cf66b into one of your commits.

maybe you could move 4b28ba9 (and probably e300832) down after 00daa67, and squash 68cf66b into it?

doc/rados/command/list-inconsistent-obj.json
```diff
@@ -0,0 +1,144 @@
...
+                  "items": {
+                    "enum": [
+                      "missing",
+                      "stat_error",
+                      "read_error",
+                      "data_digest_mismatch",
+                      "omap_digest_mismatch",
+                      "size_mismatch",
+                      "data_digest_mismatch_oi",
+                      "omap_digest_mismatch_oi",
```
athanatos: @dzafman Yeah, that's what I mean.  It seems like either it would be absent on the shard, or it would be data_digest_mismatch_oi.  data_digest_mismatch does not seem to be valid on a particular object (or what would it mean?)

doc/rados/command/list-inconsistent-obj.json
athanatos: @dzafman I don't think this has been addressed.

src/osd/PGBackend.cc
athanatos: @dzafman 
Is the only difference between the case where we have no oi digest but the shards do not match each other and the case where we do have the oi.digest and at least one shard does not match that digest that in the latter case at least one shard will have have a digest error?  I don't think that's good -- I'd rather the top level error list distinguish between those cases explicitly so that the user of this interface knows what to look for.

src/common/scrub_types.cc
```diff
@@ -139,6 +134,8 @@ inconsistent_obj_wrapper::set_auth_missing(const hobject_t& hoid,
     } else {
       shard.set_object(oid_object->second);
     }
+    errors |= shard.errors;
+    errors |= shard.objonly.errors;
```
tchaikov: agreed.

dzafman:
@athanatos @tchaikov I've added _ec_hash_error_ and _ec_size_error_ for those cases and left _read_error_ for EIO from read.   The test/osd/osd-scrub-repair.sh shows the expected output inlined in the additional test case.

src/common/scrub_types.cc
```diff
@@ -133,12 +133,10 @@ inconsistent_obj_wrapper::set_auth_missing(const hobject_t& hoid,
 {
   for (auto pg_map : maps) {
     auto oid_object = pg_map.second->objects.find(hoid);
+    // XXX: Should we just continue if object not found?
```
tchaikov: there is chance that neither does the map have the object. for example, assuming we have a active set of osd.0, osd.1 and osd.2. object `foo` could be missing in osd.1, so, in `PG::build_scrub_map_chunk()`, pg just skips it when preparing the scrub map. but neither osd.0 or osd.2 can provide an authorized copy of `foo`.

so we cannot assert in this case.

and, i think, we should at least inform user that 

1. a file is missing on osd.1. if user knows the active set of that PG interval, we can ignore this.
1. the inconsistencies found on other osds. because we don't update the scrub store in `PGBackend::be_select_auth_object()`. yeah, we are good.

in other words, if user knows its a problem that an object is not found and he/she knows which shard/replica is missing, then we can just continue, otherwise, we cannot continue. we don't provide the active set info in the list-inconsistent-object output yet. but, yes, user can leverage other ceph command for this info.

doc/rados/command/list-inconsistent-obj.json
```diff
@@ -0,0 +1,185 @@
...
+                "corrupt_object_info_attr"
+              ]
+            },
+            "minItems": 0,
+            "uniqueItems": true
+          },
+          "errors": {
+            "description": "Errors related to the analysis of this object",
+            "type": "array",
+            "items": {
```
athanatos: I don't really understand what this part means.

doc/rados/command/list-inconsistent-obj.json
```diff
@@ -0,0 +1,185 @@
...
+              "enum": [
+                "comparing_random_osd",
```
athanatos: I still think this is the wrong way to go.  If I'm understanding this right, we set comparing_random_osd in a case like:

osd 1: hash x
osd 2: hash y
osd 3: hash z

where we don't have an oi hash.  osd 1 and osd 2 would have data_digest_mismatch set and the all_shards union would have data_digest_mismatch set.  I think in that case, we should mark the all_shards one with data_digset_mismatch and simply report the digests for each shard.  I just don't see a point in picking a winner in such a case.  Am I misunderstanding?

athanatos: To expand, if we do have an oi.hash, and it's x, I think we should also set data_digest_mismatch_oi on all_shards and also on 2 and 3 (but not on 1).

dzafman: Yes, to your second comment.  That's how it works now.   In the test code with only 1 replica so osd.0 x oi x osd.1 y oi x.  osd.0 is selected as the authoritative copy with no shard errors.  Here is one the json output that is checked for in osd-scrub-repair.sh.

    {
      "all_shards": [
        "data_digest_mismatch",
        "size_mismatch",
        "data_digest_mismatch_oi",
        "size_mismatch_oi"
      ],
      "errors": [],
      "object": {
        "locator": "",
        "name": "OBJ1",
        "nspace": "",
        "snap": "head",
        "version": 3
      },
      "shards": [
        {
          "data_digest": "0x2ddbf8f5",
          "errors": [],
          "omap_digest": "0xef54f3bc",
          "osd": 0,
          "selected": true,
          "size": 7
        },
        {
          "data_digest": "0x2d4a11c2",
          "errors": [
            "data_digest_mismatch",
            "size_mismatch",
            "data_digest_mismatch_oi",
            "size_mismatch_oi"
          ],
          "omap_digest": "0xef54f3bc",
          "osd": 1,
          "selected": false,
          "size": 9
        }
      ]
    },

dzafman: Actually, I'm about to push the current code which my comment above applies to.

doc/rados/command/list-inconsistent-obj.json
dzafman: @athanatos 

data_digest_mismatch: Shard computed digest doesn't match the **selected** shard
data_digest_mismatch_oi: Shard computed digest doesn't match its own object info digest
ec_hash_error: Shard computed digest doesn't match EC recorded hash

I'm think that the following errors should have a prefix because they are the errors having to do with comparison to selected (authoritative) shard.  Maybe we should add "compare_" or something else?  Also, they are the ones which will be documented as less significant if **comparing_random_osd** is set.

data_digest_mismatch
omap_digest_mismatch
size_mismatch
attr_value_mismatch
attr_key_mismatch

dzafman:
@athanatos @tchaikov  I changed the logging to match up with the json error strings, for example:

2016-08-12 18:25:33.197924 7f22c087b700 10 osd.2 pg_epoch: 71 pg[2.0s0( v 65'7 (0'0,65'7] local-les=71 n=5 ec=18 les/c/f 71/71/0 69/70/18) [2,1,0] r=0 lpr=70 crt=45'4 lcod 45'6 mlcod 0'0 active+clean+scrubbing+deep+inconsistent] be_select_auth_object: rejecting osd 1(1) for obj 2:7580e8de:::OBJ5:head, ec_size_mismatch size_mismatch_oi
2016-08-12 18:25:33.197977 7f22c087b700 10 osd.2 pg_epoch: 71 pg[2.0s0( v 65'7 (0'0,65'7] local-les=71 n=5 ec=18 les/c/f 71/71/0 69/70/18) [2,1,0] r=0 lpr=70 crt=45'4 lcod 45'6 mlcod 0'0 active+clean+scrubbing+deep+inconsistent] be_select_auth_object: rejecting osd 1(1) for obj 2:b56c5d46:::OBJ1:head, read_error size_mismatch_oi

doc/rados/command/list-inconsistent-obj.json
athanatos: I think the osd.1 one should not have data_digest_mismatch set, I think that should only be in the all_shards set.  Also, I don't really like the "selected" field.  If we didn't have an oi digest, would one or the other still end up "selected"?

src/test/osd/osd-scrub-repair.sh
```diff
@@ -394,14 +485,1434 @@ function TEST_corrupt_scrub_replicated() {
...
+            },
+            {
+              "name": "_key3-OBJ8",
+              "value": "val3-OBJ8"
+            }
+          ],
+          "errors": [
+            "attr_value_mismatch",
+            "attr_name_mismatch"
+          ],
```
dzafman: @athanatos This shard out of the 2 didn't have any errors before because it was the selected shard.  Now the selected shard gets marked with any errors that are discovered because we don't really know which is correct.

doc/rados/command/list-inconsistent-obj.json
```diff
@@ -49,19 +49,14 @@
...
             "description": "Union of all shard errors",
             "type": "array",
             "items": {
               "enum": [
                 "missing",
                 "stat_error",
                 "read_error",
-                "data_digest_mismatch",
-                "omap_digest_mismatch",
-                "size_mismatch",
```
athanatos: Hmm, seems like we would want to mark a shard if it doesn't match the oi.size?

athanatos: Nvm, I see the size_mismatch_oi field below now.

dzafman:
Rebased on master after somehow pull request seemed to include extra commits.

dzafman:
jenkins test this please

doc/rados/command/list-inconsistent-obj.json
athanatos: Ok

src/common/scrub_types.cc
```diff
@@ -126,19 +146,24 @@ namespace librados {
 
 void inconsistent_obj_wrapper::encode(bufferlist& bl) const
 {
-  ENCODE_START(1, 1, bl);
+  ENCODE_START(2, 2, bl);
```
athanatos: I assume there's something else that will prevent old clients from seeing this encoding?

athanatos:
LGTM

dzafman:
Passes dzafman-2016-11-02_17:23:59-rados-wip-zafman-testing-distro-basic-smithi

7 errors consist of 3 infrastructure, tracker 2 x 17725, 17726 and 17654.

src/common/scrub_types.cc
dzafman: @athanatos @tchaikov Well, if you upgrade ceph-osd, reboot and scrub and finally use the previous rados command the decode will fail.  We presume that the older version of this feature isn't being used much at this point and the changes I'm making to the results would require rebuilding any other client app.

dzafman:
@tchaikov If you have time please do a final review.

tchaikov:
@dzafman sure, i am on it.

src/include/rados/rados_types.hpp
```diff
@@ -52,23 +52,23 @@ struct object_id_t {
 
 struct err_t {
   enum {
-    ATTR_UNEXPECTED      = 1 << 0,
     SHARD_MISSING        = 1 << 1,
     SHARD_STAT_ERR       = 1 << 2,
     SHARD_READ_ERR       = 1 << 3,
```
dzafman: Yes, we can reuse bits, except for user created client code using librados directly.  We are incompatible but I don't mind since we are making extensive changes.

src/osd/PGBackend.cc
```diff
@@ -513,76 +552,81 @@ map<pg_shard_t, ScrubMap *>::const_iterator
...
+    // XXX: For now we can't pick one shard for repair and another's object info
+    if (i->second.read_error || i->second.ec_hash_mismatch || i->second.ec_size_mismatch)
+      goto out;
+
+    if (auth_version == eversion_t() || oi.version > auth_version ||
+        (oi.version == auth_version && dcount(oi) > dcount(*auth_oi))) {
+      auth = j;
+      *auth_oi = oi;
+      auth_version = oi.version;
+      auth_bl.clear();
```
dzafman: It is hard to remember now, but I think when I did that it did work right because it didn't copy the bufferlist.  The next bl update would change auth_bl too.

src/tools/rados/rados.cc
```diff
@@ -1318,64 +1319,92 @@ static int do_get_inconsistent_pg_cmd(const std::vector<const char*> &nargs,
...
     for (auto kv : shard.attrs) {
       f.open_object_section("attr");
       f.dump_string("name", kv.first);
-      bufferlist b64;
-      kv.second.encode_base64(b64);
-      string v(b64.c_str(), b64.length());
-      f.dump_string("value", v);
+      bool b64;
+      f.dump_string("value", cleanbin(kv.second, b64));
+      f.dump_bool("Base64", b64);
```
dzafman: The string "Base64" is how that encoding is always referred.  So I'm consistent with that instead of with the other key names.

src/test/osd/osd-scrub-repair.sh
```diff
@@ -17,6 +17,18 @@
 source $(dirname $0)/../detect-build-env-vars.sh
 source $CEPH_ROOT/qa/workunits/ceph-helpers.sh
 
+# Set to "yes" in order to ignore diff errors and save results to update test
+getjson="no"
```
dzafman: Only getjson is used for debugging purposes.  The termwidth comes into play if a test fails to see the comparison of the json.  I wouldn't change to diff_opts since there are diffs that don't use -y.

I'll put in additional description for getjson.

src/tools/rados/rados.cc
```diff
@@ -1400,32 +1429,33 @@ static void dump_inconsistent(const inconsistent_obj_t& inc,
...
-  if (inc.has_size_mismatch())
-    f.dump_string("error", "size_mismatch");
-  if (inc.has_attr_mismatch())
-    f.dump_string("error", "attr_mismatch");
-  f.close_section();
-
+  dump_obj_errors(inc, f);
+  dump_errors(inc.union_shards, f, "union_shard_errors");
+  for (const auto& shard_info : inc.shards) {
+    shard_info_t shard = const_cast<shard_info_t&>(shard_info.second);
```
dzafman: I verified that ::decode() needs a non-const.

dzafman:
I pushed code review corrections which will be squashed.  I have rados test running again:

http://pulpito.ceph.com/dzafman-2016-11-07_16:22:05-rados-wip-zafman-testing-distro-basic-smithi

tchaikov:
lgtm. let's merge it once the test passes.

src/osd/PGBackend.cc
```diff
@@ -513,76 +552,81 @@ map<pg_shard_t, ScrubMap *>::const_iterator
...
-    if (i->second.read_error || i->second.stat_error) {
-      // scrub encountered read error or stat_error, probably corrupt
-      dout(10) << __func__ << ": rejecting osd " << j->first
-	       << " for obj " << obj
-	       << "," << (i->second.read_error ? " read_error" : "")
-	       << (i->second.stat_error ? " stat_error" : "")
-	       << dendl;
-      continue;
+    string error_string;
+    shard_map[j->first];
```
tchaikov: @dzafman ping?

dzafman: @tchaikov I was going to leave this alone, but I'm happy to add shard_info.  I'm pretty sure I need the shard_map to include the new entry even if no errors are detected, so I can't just remove the line.

tchaikov:
lgtm.

tchaikov:
@dzafman how's the test going? i think we are close.

dzafman:
Only run-tox-ceph-disk failed

retest this please

PR #9616, opened 2016-06-09
test/osd/osd-markdown: make test more reliable

liewegas:
This can fail if the osd hasn't gotten the map for the
last time it was marked down and come back up.  This
should make this test much more reliable (fewer false
positives).


liewegas:
@dachary mind taking a look?

dachary:
<pre>
</pre>

as soon as make check returns green

liewegas:
thank you!

PR #9617, opened 2016-06-09
test/mon/misc: fix pool set test given new default ec k

liewegas:
k is no longer min_size by default; adjust test accordingly.


liewegas:
@dachary you're awake!  this one too?  make check worked locally for me

dachary:
@liewegas zzzzz :-)

PR #9619, opened 2016-06-10
test: improve rbd-mirror test case coverage

dillaman:


qa/workunits/rbd/rbd_mirror_stress.sh
```diff
@@ -1,412 +1,149 @@
 #!/bin/bash
```
trociny: Minor: it looks like you have removed dependency on bash $RANDOM, so may be change this to /bin/sh?

src/vstart.sh
```diff
@@ -135,6 +136,7 @@ usage=$usage"\t--bluestore use bluestore as the osd objectstore backend\n"
 usage=$usage"\t--memstore use memstore as the osd objectstore backend\n"
 usage=$usage"\t--cache <pool>: enable cache tiering on pool\n"
 usage=$usage"\t--short: short object names only; necessary for ext4 dev\n"
+usage=$usage"\t--no_lockdep disable lockdep\n"
```
trociny: minor: (just ignore if you don't agree) To be consistent with other option names ('--nodaemon)', I would prefer if it were '--nolockdep' instead.

dillaman:
@trociny updated

trociny:
@dillaman I have just merged #9618, could you please rebase?

dillaman:
@trociny rebased

PR #9621, opened 2016-06-10
rbd-mirror: do not propagate deletions when pool unavailable

dillaman:
Fixes: http://tracker.ceph.com/issues/16229

### DIFF
--- a/src/tools/rbd_mirror/PoolWatcher.cc
+++ b/src/tools/rbd_mirror/PoolWatcher.cc
@@ -57,10 +57,12 @@ const PoolWatcher::ImageIds& PoolWatcher::get_images() const
 void PoolWatcher::refresh_images(bool reschedule)
 {
   ImageIds image_ids;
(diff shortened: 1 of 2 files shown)


src/tools/rbd_mirror/PoolWatcher.cc
```diff
@@ -72,7 +74,7 @@ void PoolWatcher::refresh_images(bool reschedule)
   // about new/removed mirrored images
 }
 
-void PoolWatcher::refresh(ImageIds *image_ids) {
+int PoolWatcher::refresh(ImageIds *image_ids) {
```
trociny: @dillaman Just wander, below we are ignoring list_images_v2 failure. May be it is safer just to return error there too?

dillaman: Good catch

dillaman:
@trociny updated

trociny:
lgtm

PR #9623, opened 2016-06-10
rbd-mirror: Throttle in-flight image syncs to only a N concurrent

rjfd:
Fixes: http://tracker.ceph.com/issues/15239

### DIFF
--- a/src/common/config_opts.h
+++ b/src/common/config_opts.h
@@ -1238,6 +1238,7 @@ OPTION(rbd_journal_pool, OPT_STR, "") // pool for journal objects
  * RBD Mirror options
  */
 OPTION(rbd_mirror_sync_point_update_age, OPT_DOUBLE, 30) // number of seconds between each update of the image sync point object number
+OPTION(rbd_mirror_concurrent_image_syncs, OPT_U32, 5) // maximum number of image syncs in parallel
 
 OPTION(nss_db_path, OPT_STR, "") // path to nss db
 
--- a/src/test/rbd_mirror/test_mock_ImageSyncThrottler.cc
+++ b/src/test/rbd_mirror/test_mock_ImageSyncThrottler.cc
@@ -1,0 +1,398 @@
+// -*- mode:C; tab-width:8; c-basic-offset:2; indent-tabs-mode:t -*-
+// vim: ts=8 sw=2 smarttab
+/*
+ * Ceph - scalable distributed file system
+ *
+ * Copyright (C) 2016 SUSE LINUX GmbH
+ *
+ * This is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License version 2.1, as published by the Free Software
+ * Foundation.  See file COPYING.
+ *
+ */
+
+#include "test/rbd_mirror/test_mock_fixture.h"
+#include "librbd/journal/TypeTraits.h"
+#include "test/journal/mock/MockJournaler.h"
+#include "test/librbd/mock/MockImageCtx.h"
+#include "librbd/ImageState.h"
+#include "tools/rbd_mirror/Threads.h"
+#include "tools/rbd_mirror/ImageSync.h"
+
+namespace librbd {
+namespace journal {
+
+template <>
+struct TypeTraits<librbd::MockImageCtx> {
+  typedef ::journal::MockJournaler Journaler;
+};
+
+} // namespace journal
+} // namespace librbd
+
+namespace rbd {
+namespace mirror {
+
+using ::testing::Invoke;
+
+typedef ImageSync<librbd::MockImageCtx> MockImageSync;
+
+template<>
+class ImageSync<librbd::MockImageCtx> {
+public:
+  static std::vector<MockImageSync *> instances;
+
+  Context *on_finish;
+  bool syncing = false;
+
+  static ImageSync* create(librbd::MockImageCtx *local_image_ctx,
+                           librbd::MockImageCtx *remote_image_ctx,
+                           SafeTimer *timer, Mutex *timer_lock,
+                           const std::string &mirror_uuid,
+                           journal::MockJournaler *journaler,
+                           librbd::journal::MirrorPeerClientMeta *client_meta,
+                           ContextWQ *work_queue, Context *on_finish,
+                           ProgressContext *progress_ctx = nullptr) {
+    ImageSync *sync = new ImageSync();
+    sync->on_finish = on_finish;
+
+    EXPECT_CALL(*sync, send())
+      .WillRepeatedly(Invoke([sync]() {
+            sync->syncing = true;
+          }));
+
+    return sync;
+  }
+
+  void finish(int r) {
+    on_finish->complete(r);
+    put();
+  }
+
+  void get() {
+    instances.push_back(this);
+  }
+
+  void put() { delete this; }
+
+  MOCK_METHOD0(cancel, void());
+  MOCK_METHOD0(send, void());
+
+};
+
+
+std::vector<MockImageSync *> MockImageSync::instances;
+
+} // namespace mirror
+} // namespace rbd
+
+
+// template definitions
+#include "tools/rbd_mirror/ImageSyncThrottler.cc"
+template class rbd::mirror::ImageSyncThrottler<librbd::MockImageCtx>;
+
+namespace rbd {
+namespace mirror {
+
+class TestMockImageSyncThrottler : public TestMockFixture {
+public:
+  typedef ImageSyncThrottler<librbd::MockImageCtx> MockImageSyncThrottler;
+
+  virtual void SetUp() {
+    TestMockFixture::SetUp();
+
+    librbd::RBD rbd;
+    ASSERT_EQ(0, create_image(rbd, m_remote_io_ctx, m_image_name, m_image_size));
+    ASSERT_EQ(0, open_image(m_remote_io_ctx, m_image_name, &m_remote_image_ctx));
+
+    ASSERT_EQ(0, create_image(rbd, m_local_io_ctx, m_image_name, m_image_size));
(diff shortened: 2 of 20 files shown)


rjfd:
@dillaman this PR is ready for review

src/tools/rbd_mirror/Mirror.cc
```diff
@@ -103,6 +111,32 @@ class FlushCommand : public MirrorAdminSocketCommand {
   Mirror *mirror;
 };
 
+class SyncSetMaxCommand : public MirrorAdminSocketArgsCommand {
```
dillaman: Unneeded if you are always pulling the max sync from ```_conf``` since there are already asok commands for updating the configuration of a running process.

src/tools/rbd_mirror/ImageSyncThrottler.h
```diff
@@ -0,0 +1,101 @@
...
+ * Copyright (C) 2016 SUSE LINUX GmbH
+ *
+ * This is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License version 2.1, as published by the Free Software
+ * Foundation.  See file COPYING.
+ *
+ */
+
+#ifndef CEPH_RBD_MIRROR_IMAGESYNCTHROTTLER_H
```
dillaman: Minor: underscores between words ```CEPH_RBD_MIRROR_IMAGE_SYNC_THROTTLER_H```

src/tools/rbd_mirror/ImageSyncThrottler.cc
```diff
@@ -0,0 +1,217 @@
...
+             << "/" << m_max_concurrent_syncs << "]" << dendl;
+  }
+
+  if (next_sync_holder) {
+    next_sync_holder->m_sync->send();
+  }
+}
+
+template <typename I, template <typename> class J>
+void ImageSyncThrottler<I, J>::set_max_concurrent_syncs(uint32_t max) {
```
dillaman: Use a ```md_config_obs_t``` to watch for updates

src/tools/rbd_mirror/ImageSyncThrottler.h
```diff
@@ -0,0 +1,101 @@
...
+    Context *m_on_finish;
+
+    C_SyncHolder(ImageSyncThrottler<ImageCtxT> *sync_throttler,
+                 const std::string& local_image_id, Context *on_finish)
+      : m_sync_throttler(sync_throttler), m_local_image_id(local_image_id),
+        m_sync(nullptr), m_on_finish(on_finish) {}
+
+    virtual void finish(int r) {
+      m_on_finish->complete(r);
+      m_sync_throttler->handle_sync_finished(this);
```
dillaman: I would be worried about shut-down ordering -- if we were shutting down rbd-mirror with syncs in-progress, it seems like it would be possible for the ImageReplayers to all shut-down (and ImageSyncThrottler to be destroyed) before this ```handle_sync_finished``` call is executed.  Perhaps swap the order?

src/tools/rbd_mirror/ImageSyncThrottler.cc
```diff
@@ -0,0 +1,217 @@
...
+    }
+  }
+
+  if (sync_holder) {
+    if (running_sync) {
+      dout(10) << "canceled running image sync for local_image_id "
+               << sync_holder->m_local_image_id << dendl;
+      sync_holder->m_sync->cancel();
+    } else {
+      dout(10) << "canceled wating image sync for local_image_id "
```
dillaman: Minor: spelling on "waiting"

src/tools/rbd_mirror/BaseRequest.h
```diff
@@ -32,8 +32,9 @@ class BaseRequest : public RefCountedObject {
 
 private:
   const std::string m_name;
-  CephContext *m_cct;
   Context *m_on_finish;
+protected:
+  CephContext *m_cct;
```
dillaman: Why do you need access to this?

src/test/rbd_mirror/test_mock_ImageSyncThrottler.cc
```diff
@@ -0,0 +1,399 @@
...
+  MockImageSync *sync3 = MockImageSync::instances[2];
+  ASSERT_FALSE(sync3->syncing);
+
+  MockImageSync *sync4 = MockImageSync::instances[3];
+  ASSERT_FALSE(sync4->syncing);
+
+  sync1->finish(0);
+  ASSERT_EQ(0, ctx1.wait());
+
+  ASSERT_TRUE(sync3->syncing);
```
dillaman: Definitely a race condition here unless you swap the order of starting the next sync and invoking the on_finish of the finished sync (from comment below)

src/tools/rbd_mirror/BaseRequest.h
rjfd: I don't :-p it's a leftover of an ancient alternative implementation

src/test/rbd_mirror/test_mock_ImageSyncThrottler.cc
```diff
@@ -0,0 +1,399 @@
...
+  ASSERT_TRUE(sync2->syncing);
+
+  cancel("image_id_2", sync2);
+  ASSERT_EQ(-ECANCELED, ctx2.wait());
+
+  sync1->finish(0);
+  ASSERT_EQ(0, ctx1.wait());
+}
+
+TEST_F(TestMockImageSyncThrottler, Cancel_Wating_Sync) {
```
dillaman: Wating -> Waiting

src/tools/rbd_mirror/ImageSyncThrottler.h
rjfd: Yes, I think you're right, and the easiest solution would be to swap the order.

rjfd:
@dillaman addressed the comments and rebased on top of master

src/common/config_opts.h
```diff
@@ -1235,6 +1235,11 @@ OPTION(rbd_journal_object_flush_bytes, OPT_INT, 0) // maximum number of pending
 OPTION(rbd_journal_object_flush_age, OPT_DOUBLE, 0) // maximum age (in seconds) for pending commits
 OPTION(rbd_journal_pool, OPT_STR, "") // pool for journal objects
 
+/**
+ * RBD Mirror options
+ */
+OPTION(rbd_mirror_concurrent_image_syncs, OPT_U32, 5)
```
dillaman: Minor: can you put a comment at the end of the line explaining it's how many images can be syncing at the same time.

src/tools/rbd_mirror/image_replayer/BootstrapRequest.cc
```diff
@@ -36,22 +37,24 @@ using librbd::util::unique_lock_name;
 
 template <typename I>
 BootstrapRequest<I>::BootstrapRequest(librados::IoCtx &local_io_ctx,
-                                      librados::IoCtx &remote_io_ctx,
```
dillaman: Minor: pull ```local_io_ctx``` down to the next line and align all params w/ 4 spaces

src/tools/rbd_mirror/ImageSyncThrottler.h
```diff
@@ -0,0 +1,104 @@
...
+                          const std::set<std::string> &changed);
+
+  uint32_t m_max_concurrent_syncs;
+  Mutex m_lock;
+  std::list<C_SyncHolder *> m_sync_queue;
+  std::map<std::string, C_SyncHolder *> m_inflight_syncs;
+
+};
+
+}
```
dillaman: Minor: can you put ```// namespace XYZ``` here and below

src/tools/rbd_mirror/ImageSyncThrottler.cc
```diff
@@ -0,0 +1,244 @@
...
+
+template <typename I, template <typename> class J>
+void ImageSyncThrottler<I, J>::handle_sync_finished(C_SyncHolder *sync_holder) {
+  dout(20) << dendl;
+
+  C_SyncHolder *next_sync_holder = nullptr;
+
+  {
+    Mutex::Locker l(m_lock);
+
```
dillaman: I think you need to issue a ```put``` against the ```ImageSync``` since you unconditionally did a ```get``` when it was registered.

rjfd: A call to `put` was already done at this point in the `finish` method of `BaseRequest` class.

rjfd:
@dillaman pushed new changes and rebased on top of master

src/tools/rbd_mirror/ImageSyncThrottler.cc
dillaman: Ah, gotcha.

dillaman:
retest this please

src/tools/rbd_mirror/ImageSyncThrottler.h
```diff
@@ -0,0 +1,104 @@
...
+
+template <typename> class ImageSync;
+
+class ProgressContext;
+
+/**
+ * Manage concurrent image-syncs
+ */
+template <typename ImageCtxT = librbd::ImageCtx,
+          template <typename> class ImageSyncT = ImageSync>
```
dillaman: Why do you need the template param for ImageSync?  You should be able to directly use ImageSync<ImageCtxT> and within your test case you can create a template specialization for ImageSync<MockImageCtx> which no-ops the entire class.

rjfd: You're right, I don't need it. I'll remove that template param.

rjfd:
@dillaman pushed new changes

rjfd:
@dillaman rebased on top of master

dillaman:
@rjfd Rebase please

rjfd:
@dillaman done

PR #9625, opened 2016-06-10
Allow building RGW with LDAP disabled

dang:


dang:
@alimaredia @mattbenjamin ping

mattbenjamin:
@dang per verbal, I think we should just move the parsing of the bindpw into unconditional code, then we don't need #ifdef in librgw.cc and rgw_rest_s3.cc.

PR #9626, opened 2016-06-10
os/bluestore: implement 2Q buffer cache policy

liewegas:


src/common/config_opts.h
```diff
@@ -965,6 +965,7 @@ OPTION(bluestore_compression, OPT_STR, "none")  // force|aggressive|passive|none
 OPTION(bluestore_compression_algorithm, OPT_STR, "snappy")
 OPTION(bluestore_compression_min_blob_size, OPT_U32, 256*1024)
 OPTION(bluestore_compression_max_blob_size, OPT_U32, 4*1024*1024)
+OPTION(bluestore_cache_type, OPT_STR, "lru")
```
ifed01: May be it's worth to enumerate other types as a comment?

src/os/bluestore/BlueStore.cc
```diff
@@ -480,21 +516,13 @@ void BlueStore::Cache::trim(uint64_t onode_max, uint64_t buffer_max)
...
-      } else {
-	break;
-      }
+    assert(b->is_clean());
+    auto p = b->space->buffer_map.find(b->offset);
+    if (i != buffer_lru.begin()) {
+      --i;
     }
+    dout(20) << __func__ << " rm " << *b << dendl;
+    b->space->_rm_buffer(p);
```
ifed01: Looks like one can have a  BufferSpace::_rm_buffer(offset) instead of doing find + _rm_buffer(iterator). There are several similar snippets below

liewegas:
i think there is a bug in the hot list population (it's always size 0)

src/os/bluestore/BlueStore.h
```diff
@@ -490,39 +497,190 @@ class BlueStore : public ObjectStore,
...
+
+    virtual void _add_buffer(Buffer *b, int level, Buffer *near) = 0;
+    /// remove a buffer.  return level to be used by any replacement buffer.
+    virtual int _rm_buffer(Buffer *b) = 0;
+    virtual void _adjust_buffer_size(Buffer *b, int64_t delta) = 0;
+    virtual void _touch_buffer(Buffer *b) = 0;
+
+    virtual void trim(uint64_t onode_max, uint64_t buffer_max) = 0;
+
+#ifdef DEBUG_CACHE
```
ifed01: IMHO it's better to move DEBUG_CACHE definition to Bluestore.cc and fork implementation there only. And let interfaces to be the same for both defined and undefined cases. No much overhead in runtime if any. And the code is more readable and no huge rebuilds if there is a need to change.

src/os/bluestore/BlueStore.h
```diff
@@ -208,28 +212,28 @@ class BlueStore : public ObjectStore,
       assert(writing.empty());
     }
 
-    void _add_buffer(Buffer *b) {
-      cache->_audit_lru("_add_buffer start");
+    void _add_buffer(Buffer *b, int level, Buffer *near) {
```
ifed01: Why not defining such methods really protected? I suppose this is pretty doable.

src/os/bluestore/BlueStore.cc
```diff
@@ -589,6 +591,196 @@ void BlueStore::LRUCache::_audit(const char *when)
...
+    }
+  }
+
+  // onodes
+  int num = onode_lru.size() - onode_max;
+  if (num <= 0)
+    return; // don't even try
+
+  auto p = onode_lru.end();
+  if (num)
```
ifed01: num is always > 0 here

src/os/bluestore/BlueStore.h
```diff
@@ -490,39 +511,190 @@ class BlueStore : public ObjectStore,
...
+	boost::intrusive::list_member_hook<>,
+	&Buffer::lru_item> > buffer_list_t;
 
-    std::mutex lock;                ///< protect lru and other structures
-    buffer_lru_list_t buffer_lru;
-    uint64_t buffer_size = 0;
     onode_lru_list_t onode_lru;
 
-    void _touch_onode(OnodeRef& o);
+    buffer_list_t buffer_hot;      //< "Am" hot buffers
```
ifed01: Probably I'm missing something but it looks like any insert/push to this list happens under some switch that checks for BUFFER_HOT value. But I'm unable to find any code that assigns BUFFER_HOT unless one at Bluestore.cc:647. However it's under BUFFER_HOT check too! Hence I suppose buffer_hot is never used at the moment.

src/os/bluestore/BlueStore.h
```diff
@@ -490,39 +511,190 @@ class BlueStore : public ObjectStore,
...
+      if (!b->is_empty()) {
+	buffer_bytes -= b->length;
+      }
+      switch (b->cache_private) {
+      case BUFFER_WARM_IN:
+	buffer_warm_in.erase(buffer_warm_in.iterator_to(*b));
+	return 1;
+      case BUFFER_WARM_OUT:
+	buffer_warm_out.erase(buffer_warm_out.iterator_to(*b));
+	return 2;
```
ifed01: It's pretty unclear what values 1 & 2 means here. From the comments in the header this should be some level used for replacement. In _add_buffer right? But as far as I can see _add_buffer cares of level values that are greater or less than zero, not 1 or 2. Moreover I can't see any assignment for _rm_buffer's return value hence this stuff looks useless at the moment.

src/os/bluestore/BlueStore.h
liewegas: _add_buffer will insert if cache_private is already set to BUFFER_HOT.  This happens in BufferSpace::write() and did_read(), where cache_private gets the return value from _discard.  _discard returns the highest cache_private value of any buffers we are discarding (replacing).

The key to 2Q is that we get bumped to buffer_hot if we replace a buffer in the WARM_OUT (an empty buffer that we previously trimmed from warm_in) or a previously hot buffer.

src/os/bluestore/BlueStore.cc
liewegas: fixed

liewegas:
fixed the _rm_buffer thing... that was leftover nonsense from when I was doing the original cache_private thing.

src/os/bluestore/BlueStore.h
ifed01: Can you show me the first place where BUFFER_HOT value is assigned. I can see its reuse but not the first assignment. I.e. there is no line x = BUFFER_HOT;

liewegas: <pre>
    case BUFFER_WARM_OUT:
      // move to hot.  fall-thru
    case BUFFER_HOT:
      dout(20) << __func__ << " move to hot " << *b << dendl;
      b->cache_private = BUFFER_HOT;
      buffer_hot.push_front(*b);
      break;
</pre>
in _add_buffer.  the fall-thru is the key.

ifed01: Yeah. Missed that. Thanks!

PR #9635, opened 2016-06-10
make-dist: include src/.git_version and src/ceph_ver.h

liewegas:
This is one of the main (only?) things preventing us from ditching 'make dist'


tchaikov:
the created tarball compiles fine. and `src/ceph_ver.h` is created.

PR #9636, opened 2016-06-10
msg/xio: fix entity_addr_t encoding

mattbenjamin:
this is related to entity_addr_t changes--can someone have a look?


liewegas:
lgtm

PR #9639, opened 2016-06-10
os/bluestore: Add an option to choose bluefs direct or buffered io mode

jjhuo:
For now, set default to buffered io; after rocksdb block cache
has ability to cache writes, we will set default to direct io.

For more info, see
https://github.com/ceph/ceph/pull/9533


jjhuo:
@liewegas Could you please review this? Somnath is benchmarking the master, and it's not easy for him to avoid my previous PR 9533. 
I am requesting rocksdb team to enable block cache to cache writes, after that, we can remove this option or set it to false. Thanks.

liewegas:
As soon as we confirm rocksdb caching is working (I think mainly we need to make it populate the cache during compaction?) I think we should just rip out the buffered option (out of bluefs and the blockdevice).  So this is only useful if it helps for testing in the interim?

jjhuo:
That's exactly right. After we make it populate the cache during compaction, we should remove this option. This is only for testing. Thanks.

liewegas:
retest this please

PR #9640, opened 2016-06-11
librbd: recursive lock possible when disabling journaling

dillaman:
If pool-level mirroring is enabled, deleting the journal
could cause a deadlock attempting to delete remote peer
sync-point snapshots.

Fixes: http://tracker.ceph.com/issues/16235


trociny:
lgtm

PR #9641, opened 2016-06-11
DNM: bluefs: drop lock during log compaction

varadakari:
1. Reorg the compaction process to allow frontend threads to work concurrently with log compaction. Log file is still protected by a global lock.
2. Added utility function to dump the log file contents.
3. Added some test cases to test compaction and read/write files.

Passed make check and unittest_bluefs with new test cases.
Signed-off-by: Varada Kari <varada.kari@sandisk.com>

### DIFF
--- a/src/os/bluestore/BlueFS.cc
+++ b/src/os/bluestore/BlueFS.cc
@@ -867,6 +867,7 @@ uint64_t BlueFS::_estimate_log_size()
   return ROUND_UP_TO(size, super.block_size);
 }
 
+
 void BlueFS::_maybe_compact_log()
 {
   uint64_t current = log_writer->file->fnode.size;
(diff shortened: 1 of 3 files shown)


varadakari:
@xiexingguo @liewegas Could you please review the changes?

src/os/bluestore/BlueFS.cc
```diff
@@ -1315,7 +1561,39 @@ void BlueFS::sync_metadata()
...
+  uint64_t expected = _estimate_log_size();
+  float ratio = (float)current / (float)expected;
+  dout(10) << __func__ << " current " << current
+	   << " expected " << expected
+	   << " ratio " << ratio << dendl;
+
+  if (current < g_conf->bluefs_log_compact_min_size ||
+      ratio < g_conf->bluefs_log_compact_min_ratio) {
+    lock.unlock();
+    return;
```
liewegas: why duplicate maybe_compact_log?  if the locking is different, turn it into a helper like _should_compact_log(() or something

src/os/bluestore/BlueFS.cc
```diff
@@ -929,6 +1164,7 @@ void BlueFS::_compact_log()
   bufferlist bl;
```
liewegas: I think we need to change op_jump_seq to specify a seq *and* a file offset... maybe just rename it jump?  then we can jump to the next 1MB-aligned offset, where the log will continue?

liewegas:
can you reorg this so the first patch adds the dump log function?

src/os/bluestore/BlueFS.cc
```diff
@@ -1217,6 +1217,7 @@ void BlueFS::_pad_bl(bufferlist& bl)
 
 void BlueFS::flush_log()
 {
+  std::lock_guard<std::mutex> l(lock);
```
liewegas: this goes in its own patch

varadakari:
A new PR is raised for the same #10285 , closing this.

PR #9644, opened 2016-06-11
osdc/ObjectCacher: change iterator to const_iterator and add const to member functions

stiopaa1:


src/osdc/ObjectCacher.cc
```diff
@@ -171,7 +171,7 @@ void ObjectCacher::Object::try_merge_bh(BufferHead *bh)
 bool ObjectCacher::Object::is_cached(loff_t cur, loff_t left)
 {
   assert(oc->lock.is_locked());
-  map<loff_t, BufferHead*>::iterator p = data_lower_bound(cur);
+  map<loff_t, BufferHead*>::const_iterator p = data_lower_bound(cur);
```
tchaikov: i'd recommend to use `auto` here.

liewegas:
retest this please

PR #9647, opened 2016-06-12
server: fix potential access violation

xiexingguo:
The mdcache->request_start() method may lose the forward race
against a slave and thus may either drop the last reference of the
passed in message or push the message into the waiting_for_finish
list for a later retry.

In either case we shall stall and exit, otherwise we'll access
violation especifally for the former case(as we drop the last
reference of "req" and it shall be destructed by now).


jcsp:
Reviewed-by

PR #9649, opened 2016-06-12
server: return after respond_to_request() for EROFS

xiexingguo:
Respond_to_request will do cleanup jobs, thus we shall
stall and exit under this case.

### DIFF
--- a/src/mds/Server.cc
+++ b/src/mds/Server.cc
@@ -2879,6 +2879,7 @@ void Server::handle_client_open(MDRequestRef& mdr)
   if ((cmode & CEPH_FILE_MODE_WR) && mdcache->is_readonly()) {
     dout(7) << "read-only FS" << dendl;
     respond_to_request(mdr, -EROFS);
+    return;
   }
   
   set<SimpleLock*> rdlocks, wrlocks, xlocks;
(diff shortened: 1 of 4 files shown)


src/mds/Server.cc
```diff
@@ -3652,8 +3652,8 @@ void Server::handle_client_file_readlock(MDRequestRef& mdr)
     break;
 
   default:
-    dout(10) << "got unknown lock type " << checking_lock.type
-	     << ", dropping request!" << dendl;
+    dout(10) << "got unknown lock type " << checking_lock.type << dendl;
+    respond_to_request(mdr, -EINVAL);
```
jcsp: What is the client behaviour in this case?

jcsp:
The actual EROFS part of this is fine, not so sure about the unknown lock type change until we see how a client would actually handle EINVAL

xiexingguo:
@jcsp Thanks for the careful review, john. Seems it passed the test, but let me know if you want me to drop that change.

jcsp:
On reflection returning EINVAL to clients in the unknown lock case is probably okay, since it indicates bad behavour from the client (and I don't mind upsetting the client with a return code it might not handle gracefully in that case).

jcsp:

gregsfortytwo:
FYI: https://github.com/ceph/ceph/pull/11985 had to revert the Mutation::_dump() patch.

xiexingguo:
> FYI: #11985 had to revert the Mutation::_dump() patch.

Thanks for your reminder, Greg. And sorry for that:-(

PR #9653, opened 2016-06-12
rbd-mirror: FAILED assert(!m_status_watcher)

trociny:
Fixes: http://tracker.ceph.com/issues/16245

### DIFF
--- a/src/tools/rbd_mirror/Replayer.cc
+++ b/src/tools/rbd_mirror/Replayer.cc
@@ -599,7 +599,7 @@ void Replayer::set_sources(const ImageIds &image_ids)
     return;
   }
 
-  if (m_image_replayers.empty()) {
+  if (m_image_replayers.empty() && !existing_image_replayers) {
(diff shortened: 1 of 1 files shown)


trociny:
@dillaman Observed this when running rbd_mirror_stress.sh, on "TEST: create many images".

dillaman:
lgtm

dillaman:
Fails pretty consistently now when running rbd_mirror.sh (although I am not sure why) -- 3 out of 3 tries and passed 2 out of two once reverted:
> ceph --admin-daemon /tmp/tmp.rbd_mirror/rbd-mirror.cluster1_daemon.cluster1.asok rbd mirror flush mirror/test2
> Invalid command:  unused arguments: ['mirror/test2']
> rbd mirror flush :  flush rbd mirror

dillaman:
... looks like a race condition in rbd_mirror.sh although still not sure why it occurs more often with this change.

PR #9657, opened 2016-06-13
librbd: avoid holding owner_lock waiting for exclusive lock

dillaman:
Fixes: http://tracker.ceph.com/issues/16235

### DIFF
--- a/src/librbd/internal.cc
+++ b/src/librbd/internal.cc
@@ -1706,11 +1706,17 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force,
 
       C_SaferCond lock_ctx;
       ictx->exclusive_lock->request_lock(&lock_ctx);
+
+      // don't block holding lock since refresh might be required
+      ictx->owner_lock.put_read();
       r = lock_ctx.wait();
+      ictx->owner_lock.get_read();
+
       if (r < 0) {
         lderr(cct) << "failed to lock image: " << cpp_strerror(r) << dendl;
         return r;
-      } else if (!ictx->exclusive_lock->is_lock_owner()) {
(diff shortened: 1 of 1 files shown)


src/librbd/internal.cc
```diff
@@ -1706,11 +1706,17 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force,
 
       C_SaferCond lock_ctx;
       ictx->exclusive_lock->request_lock(&lock_ctx);
+
+      // don't block holding lock since refresh might be required
+      ictx->owner_lock.put_read();
```
trociny: @dillaman May be it makes sense then to acquire lock before block_writes and block_requests? (not sure how important it is, is there a chance something will unblock while we are not holding the lock?)

Also, don't we need the same when we are releasing the exclusive lock?

Another thing I have noticed, not related to your change, that after acquiring the lock we don't check for acquired_lock value when returning due to a error below. This looks like a potential lock leakage to me?

dillaman: The order for blocking writes/requests was required to prevent an rbd CLI from accepting proxied requests on behalf of the client we might just have taken the lock. The block request call only increments a counter -- the block writes call might result in a refresh if and only if there was active IO (which wouldn't be coming from the rbd CLI). 

The release lock won't trigger a refresh -- that is built into the acquire lock state machine due to a recent bug fix.

The lock will be released whenever the image is closed (or requested) -- that release at the end was just a way to force a specific event order for test cases with regard to the notify update. 

This whole routine needs to be re-written as an async state machine because it's grown to be quite unmanageable with all it needs to do.

trociny:
lgtm

PR #9659, opened 2016-06-13
doc/radosgw/admin: the option '--purge-keys' not purge all keys associated to the uid but only purge the key spedified by '--access-key'

weiqiaomiao:

### DIFF
--- a/doc/radosgw/admin.rst
+++ b/doc/radosgw/admin.rst
@@ -111,12 +111,23 @@ For example::
     "max_buckets": 1000,
     "auid": 0,
     "subusers": [
-          { "id": "johndoe:swift",
-            "permissions": "full-control"}],
+          { 
+            "id": "johndoe:swift",
+            "permissions": "full-control"
+          }
+     ],
     "keys": [
-          { "user": "johndoe",
+          { 
+            "user": "johndoe",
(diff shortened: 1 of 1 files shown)


theanalyst:
@weiqiaomiao could you also add an example for using purge keys to purge a single key alone?

weiqiaomiao:
@theanalyst  ok, I add it

weiqiaomiao:
retest this please

yehudasa:
@weiqiaomiao  @theanalyst I think that the bug here is that purge_keys is not working as expected.

theanalyst:
@yehudasa @weiqiaomiao agreed

weiqiaomiao:
This bug was fixed in  commit [e7b7e1a](https://github.com/ceph/ceph/commit/e7b7e1afc7a81c3f97976f7442fbdc5118b532b5) ，so in jewel(and later) version there is no need to specify '--purge-data' to clean all keys because of the remove subuser opt will purge all associate keys by default, but the bug is still in hammer(and older) version ,  I think this bug have little effect so we need not to backport it to hammer and we can only add the tips in doc . What do you think?

liewegas:
It's probably safe to drop this now? If not please reopen!

PR #9660, opened 2016-06-13
librbd: Check image_watcher before sending update

tdsparrow:
This fixes https://github.com/ceph/go-ceph/issues/20  
  rbd_open didn't initialize image_wathcer which will be used later for notification.
  It will cause go-ceph client crash.

Signed-off-by: Ren Li <sqallowlee@gmail.com>

### DIFF
--- a/src/librbd/ImageCtx.cc
+++ b/src/librbd/ImageCtx.cc
@@ -1020,14 +1020,18 @@ struct C_InvalidateCache : public Context {
   void ImageCtx::notify_update() {
     state->handle_update_notification();
 
-    C_SaferCond ctx;
-    image_watcher->notify_header_update(&ctx);
-    ctx.wait();
+    if (image_watcher != NULL) {
+      C_SaferCond ctx;
+      image_watcher->notify_header_update(&ctx);
+      ctx.wait();
+    }
   }
 
   void ImageCtx::notify_update(Context *on_finish) {
     state->handle_update_notification();
-    image_watcher->notify_header_update(on_finish);
+    if (image_watcher != NULL) {
+      image_watcher->notify_header_update(on_finish);
+    }
   }
 
   exclusive_lock::Policy *ImageCtx::get_exclusive_lock_policy() const {


dillaman:
@tdsparrow what API method are you invoking when this crashes? The ```ImageWatcher``` should be initialized when the image is opened in R/W mode -- and attempting to do any op that changes the image state while in R/O mode should result in a -EROFS error.

david-z:
@dillaman We have also met the same issue recently when we are evaluating jewel or master branch. The steps to reproduce are follows:

1. open image as RO mode.
2. try "rbd_lock_exclusive" on this image, here fails.

The failure reason is as you and @tdsparrow mentioned, that ImageWatcher is initialized at open stage with RW mode, so if open image with RO mode, ImageWatcher will be null. Before jewel, e.g, in hammer, although ImageWatcher is also initialized at open stage with RW mode, calling notify_header_update after exclusive locking won't be via ImageWatcher, so it is good.

I kinda feel current fix is not good enough because notify_header_update to watchers after exclusive locking will be skipped if in RO mode. So the behavior may be different from previous version, e.g, hammer. 

Since accessing notify_header_update needs to be via ImageWatcher in jewel, what I can think of is to fake a ImageWatcher without registering a real one for RO mode and just use it to do notification job. Here is the code diff and tests are passed.

```
diff --git a/src/librbd/image/OpenRequest.cc b/src/librbd/image/OpenRequest.cc
index 7714722..4b43ef0 100644
--- a/src/librbd/image/OpenRequest.cc
+++ b/src/librbd/image/OpenRequest.cc
@@ -13,6 +13,7 @@
 #include "librbd/image/SetSnapRequest.h"
 #include <boost/algorithm/string/predicate.hpp>
 #include "include/assert.h"
+#include "librbd/ImageWatcher.h"
 
 #define dout_subsys ceph_subsys_rbd
 #undef dout_prefix
@@ -339,6 +340,8 @@ void OpenRequest<I>::send_register_watch() {
   m_image_ctx->init();
 
   if (m_image_ctx->read_only) {
+    // for ro mode, fake a image_watcher without registering a real watcher
+    m_image_ctx->image_watcher = new ImageWatcher(*m_image_ctx);
     send_refresh();
     return;
   }
```

What do your guys think? Thanks.

dillaman:
Ok -- so it looks like the advisory locking API methods don't require the image to be in R/W mode.  I think the correct fix is to have the ```ImageCtx::notify_update()``` method invoke a new static ```ImageWatcher::notify_header_update(librados::IoCtx &io_ctx, const std::string &oid)``` method.  Basically -- re-add the static method that existed in Hammer/Infernalis and use it when doing a synchronous header update notification.

dillaman:
Associated tracker ticket: http://tracker.ceph.com/issues/16364

dillaman:
PR #9882 restores the Hammer/Infernalis code used for image notifications in support of advisory locks for R/O images.

PR #9661, opened 2016-06-13
qa/workunits: ensure replay has started before checking position

dillaman:
Fixes: http://tracker.ceph.com/issues/16248


trociny:
lgtm

PR #9662, opened 2016-06-13
test: ceph_test_objectstore: do not copy compressor plugins if not writable

tchaikov:
there is chance that we launch the test with the plugins installed in `/usr`.


tchaikov:
superseded by #9666

PR #9665, opened 2016-06-13
mon/MonClient: fix wrong expiration detection of rotating keys

xiexingguo:
We shall ask for renewing the current rotating key
when the current key is going to expire in 1/4 of its tenure.

E.g., we issue the current key at 2016/6/13 14:30:00,
and its expiration is 2016/6/13 15:30:00.
The tick() thread(which wakes up every 10 seconds) shall detect that
the key is going to expire at 2016/6/13 15:15:10,
by which time we have(cutoff = now + 1/4 ttl):
   **expiration(2016/6/13 15:30:00) < cutoff(2016/6/13 15:30:10)**

Instead the current code logic will detect the expiration of the
current key at 2016/6/13 15:30:40,
by which time we have(cutoff = now - 30):
  **expiration(2016/6/13 15:30:00) < cutoff(2016/6/13 15:30:10)**

### DIFF
--- a/src/mon/MonClient.cc
+++ b/src/mon/MonClient.cc
@@ -827,7 +827,7 @@ int MonClient::_check_auth_rotating()
 
   utime_t now = ceph_clock_now(cct);
   utime_t cutoff = now;
-  cutoff -= MIN(30.0, cct->_conf->auth_service_ticket_ttl / 4.0);
+  cutoff += MAX(30.0, cct->_conf->auth_service_ticket_ttl / 4.0);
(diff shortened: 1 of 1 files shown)


xiexingguo:
@liewegas Mind having a look? Thanks.

liewegas:
I think this is deliberate: 44dc114385b82345008b62afd270c4cf076e252c makes us renew the ticket after it expires, so ensure the mon has a new replacement ticket ready and we don't busy-loop.

xiexingguo:
@liewegas Got it. Thanks!

PR #9666, opened 2016-06-13
test: ceph_test_objectstore: do not override plugin-dir if not necessary

tchaikov:


tchaikov:
tested at http://pulpito.ceph.com/kchai-2016-06-14_03:49:21-rados-wip-ceph_test_objectstore---basic-mira/

tchaikov:
test passed http://pulpito.ceph.com/kchai-2016-06-14_03:49:21-rados-wip-ceph_test_objectstore---basic-mira/

tchaikov:
@ifed01 could you please take a look? thanks.

src/test/objectstore/store_test.cc
```diff
@@ -4485,6 +4485,13 @@ TEST(DummyTest, ValueParameterizedTestsAreNotSupportedOnThisPlatform) {}
 
 #endif
 
+static bool plugin_exists(const string& compressor_dir)
+{
+  string path(compressor_dir);
+  path += "/libceph_snappy.so";
```
ifed01: Not needed IMO

ifed01: Other plugins will depend on snappy presence that's not good IMHO

tchaikov: @ifed01 how about `glob(3)` to see if any plugin exists in this directory?

tchaikov: or you'd suggest we check for the existence of the `compressor_dir` directory instead?

ifed01: What's about check for non-empty compressor_dir directory?

tchaikov:
changelog 
- rebase against master
- do not set "plugin_dir" if `$CEPH_LIB` does not have any plugin in it.

@ifed01 could you take a look again?

ifed01:
lgtm!

PR #9667, opened 2016-06-13
ceph-osd-prestart.sh: drop Upstart-specific code

smithfarm:
http://tracker.ceph.com/issues/15984

### DIFF
--- a/src/ceph-osd-prestart.sh
+++ b/src/ceph-osd-prestart.sh
@@ -23,12 +23,10 @@ journal="$data/journal"
 if [ -L "$journal" -a ! -e "$journal" ]; then
     udevadm settle --timeout=5 || :
     if [ -L "$journal" -a ! -e "$journal" ]; then
-        echo "ceph-osd($UPSTART_INSTANCE): journal not present, not starting yet." 1>&2
-        stop
+        echo "ceph-osd(${cluster:-ceph}-$id): journal not present, not starting yet." 1>&2
         exit 0
     fi
 fi
-
 
 # ensure ownership is correct
 owner=`stat -c %U $data/.`


smithfarm:
@ktdreyer @b-ranto Care to review?

ktdreyer:
LGTM (untested)

b-ranto:
I'm not sure we want to do this. Especially when this is considered in conjunction with http://tracker.ceph.com/issues/15990 where we constantly get the ceph-disk failures because we treat this situation as an error and not just a warning.

I'd suggest to do the same for 15990 and just warn (not error out) in case one of the devices is not up, yet.

src/ceph-osd-prestart.sh
```diff
@@ -21,15 +21,14 @@ data="/var/lib/ceph/osd/${cluster:-ceph}-$id"
...
 
 if [ -L "$journal" -a ! -e "$journal" ]; then
+    echo "ceph-osd(${cluster:-ceph}-$id): journal not present, not starting yet." 1>&2
     udevadm settle --timeout=5 || :
     if [ -L "$journal" -a ! -e "$journal" ]; then
-        echo "ceph-osd($UPSTART_INSTANCE): journal not present, not starting yet." 1>&2
-        stop
-        exit 0
+        echo "ceph-osd(${cluster:-ceph}-$id): journal still not present, aborting." 1>&2
+        exit 1
```
smithfarm: @b-ranto So, change this line to "exit 0" and perhaps change the warning in the previous line to "journal device still not present after 5 seconds; giving up" ?

b-ranto: I think we should be good if we just dropped the 'stop' line. The messages at their current form look sane to me.

smithfarm:
@b-ranto OK, I narrowed the scope of this PR. It no longer attempts to fix http://tracker.ceph.com/issues/15984

smithfarm:
Changelog:
* rebased
* narrowed scope of modifications

smithfarm:
Hey Jenkins, test this please

smithfarm:
@b-ranto Do you think this is OK to merge?

b-ranto:
Yup, seems pretty safe (merged it myself).

PR #9668, opened 2016-06-13
rocksdb: remove rdb source files from dist tarball

tchaikov:
rdb is based on NodeJS, and we don't build rdb to use rocksdb as
a keyvaluestore backend.

Fixes: #13554


tchaikov:
this fixes: http://gitbuilder.sepia.ceph.com/gitbuilder-ceph-rpm-centos7-amd64-basic/log.cgi?log=57ff489dff555ee523f9de758b7420e772bdfad8

PR #9670, opened 2016-06-13
debian: silence couple warnings

tchaikov:


tchaikov:
@dmick here are more `debian/rules` cleanups.

tchaikov:
rebased against master.

tchaikov:
@dmick ping?

debian/control
```diff
@@ -649,7 +648,6 @@ Section: python
 Depends: python-rados (= ${binary:Version}),
          python-rbd (= ${binary:Version}),
          python-cephfs (= ${binary:Version})
-X-Python-Version: >= 2.6
```
dmick: Is this true for versions as old as Ubuntu Trusty?

tchaikov: see http://packages.ubuntu.com/trusty/python

dmick: kefu answers in PM that yes it si

PR #9676, opened 2016-06-13
os/bluestore: Fix enumeration when only single key in BitmapFreelistManager

efirs:
OSD crashes at startup when bluestore_bluefs=false.

enumerate_offset is not enough to detect new enumeration, enumerate_bl_pos needs to be take into account too.


xiexingguo:
looks good

liewegas:
please add your signed-off-by to the git commit

PR #9677, opened 2016-06-13
mds/SessionMap.h: remove duplicated line and add const to member functions

stiopaa1:


gregsfortytwo:
LGTM! It's apparently failing unit tests including some mds thing, but I think it's a bad base and rbd is actually causing the trouble.

PR #9678, opened 2016-06-13
test: Explicitly use tempfile delete= feature instead of os.unlink()

dzafman:
The warning messages are caused by new delete= tempfile feature
trying to unlink a file after test has used os.unlink().


tchaikov:
passed in jenkins' `make check`
```
PASS: test/ceph_objectstore_tool.py
```

tchaikov:
lgtm

PR #9680, opened 2016-06-14
vstart.sh: set plugin-dir for cmake

tchaikov:


tchaikov:
@jcsp mind taking a look?

PR #9681, opened 2016-06-14
server: set terminating_sessions flag correctly

xiexingguo:
So caller won't call terminate_sessions() multiple times.
    

jcsp:
Passed tests http://pulpito.ceph.com/jspray-2016-06-14_13:37:22-fs-wip-jcsp-testing-20160614-distro-basic-mira/

PR #9683, opened 2016-06-14
PIN_FRAGMENTING should be 2 instead of -2

linuxhunter:
PIN_FRAGMENTING in CDentry.h should be 2 instead of -2. I think it a erratum.

### DIFF
--- a/src/mds/CDentry.h
+++ b/src/mds/CDentry.h
@@ -79,7 +79,7 @@ public:
 
   // -- pins --
   static const int PIN_INODEPIN =     1;  // linked inode is pinned
-  static const int PIN_FRAGMENTING = -2;  // containing dir is refragmenting
+  static const int PIN_FRAGMENTING =  2;  // containing dir is refragmenting
   static const int PIN_PURGING =      3;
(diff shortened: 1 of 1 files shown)


jcsp:
This change looks OK

Please update your git commit message to the required style (first line should start "mds: ", and you should add a Signed-of-by line) http://ceph.com/resources/development/ceph-commit-comments/

jcsp:
Passed tests: http://pulpito.ceph.com/jspray-2016-06-14_13:37:22-fs-wip-jcsp-testing-20160614-distro-basic-mira/

Just needs the commit message cleaned up

gregsfortytwo:
I don't think this is correct. Negative values on pins have meaning, although I don't remember what that meaning is offhand (but I think we do comparisons against zero to check if some multi-MDS operations are allowed?). Look at the pin values in CInode for more examples.

jcsp:
@liewegas any chance you can point out where negative values here have an effect?

liewegas:
I'm pretty sure it means you're allowed to have multiple instances of that same pin type.  Although it would only matter if the ref_map debugging is enabled (it still is by default i think?), and I can't find the code that checks/cares.

PR #9684, opened 2016-06-14
os/bluestore: Fix for bit_alloc unit test case stack size and handle device that has less than block size at end.

chhabaramesh:
Hi,
The two changes involves
1. Reduce stack size of the bit_alloc_unit test that was causing segfault in some environments.
2. Handle case when total device size is not multiple of block size used by bluefs (1MB or 64KB). 

Tests:
1. make check
2. cheph_test_objectstore
3. unit test cases for environment that had disk with odd size,


chhabaramesh:

chhabaramesh:

PR #9698, opened 2016-06-14
rados: add option to include clones when doing flush or evict

LiumxNL:
if a cache tier has snaps, or image existing in the pool has self-manage snaps，after io workload, there maybe lots of clone objects. in current logic, when excuting "rados cache-flush", will get an -EBUSY error if clone objects is dirty. when excuting "rados cache-evcit" will get an -EBUSY error if clone exists. it's not very comfortable to flush/evict clones one by one, in addition, for self-manage snapshot, like rbd snapshot, we cannot specify snap in rados command tool. this patch make all these older clones to be flushing/evicting when excuting "rados cache-flush/evict", after clones has flushed/evicted, the op will go through.

Signed-off-by: Mingxin Liu <mingxin@xsky.com>

### DIFF
--- a/src/tools/rados/rados.cc
+++ b/src/tools/rados/rados.cc
@@ -1190,22 +1190,59 @@ static int do_cache_flush_evict_all(IoCtx& io_ctx, bool blocking)
 	io_ctx.locator_set_key(string());
       }
       io_ctx.set_namespace(i->get_nspace());
-      if (blocking)
-	r = do_cache_flush(io_ctx, i->get_oid());
-      else
-	r = do_cache_try_flush(io_ctx, i->get_oid());
+      snap_set_t ls;
+      io_ctx.snap_set_read(LIBRADOS_SNAP_DIR);
+      r = io_ctx.list_snaps(i->get_oid(), &ls);
       if (r < 0) {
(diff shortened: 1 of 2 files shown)


LiumxNL:
@liewegas would you like to take a look? thanks!

athanatos:
There is an operation which exposes the existing snapshots on an object.  I'd prefer to use that along with the existing cache/flush operations to flush the clones.  If you want to make this easier in the rados tool, I think the logic should be implemented entirely in the tool without messing with the actual OSD-side code.

LiumxNL:
@athanatos updated, please review, thanks a lot.

src/tools/rados/rados.cc
```diff
@@ -220,6 +220,8 @@ void usage(ostream& out)
 "   --read-percent                   percent of operations that are read\n"
 "   --target-throughput              target throughput (in bytes)\n"
 "   --run-length                     total time (in seconds)\n"
+"CACHE POOLS OPTIONS:\n"
+"   --with-clones                    include its elder clones to flush/evict\n"
```
liewegas: include clones when doing flush or evict

liewegas:
Aside from the usage string, this looks good!

LiumxNL:
@liewegas  thanks!

tchaikov:

PR #9699, opened 2016-06-14
rbd-mirror:  Update the sync point object number during sync

rjfd:
Fixes: http://tracker.ceph.com/issues/15108

### DIFF
--- a/src/common/config_opts.h
+++ b/src/common/config_opts.h
@@ -1234,6 +1234,11 @@ OPTION(rbd_journal_object_flush_bytes, OPT_INT, 0) // maximum number of pending 
 OPTION(rbd_journal_object_flush_age, OPT_DOUBLE, 0) // maximum age (in seconds) for pending commits
 OPTION(rbd_journal_pool, OPT_STR, "") // pool for journal objects
 
+/**
+ * RBD Mirror options
+ */
+OPTION(rbd_mirror_sync_point_update_age, OPT_DOUBLE, 30) // number of seconds between each update of the image sync point object number
+
 OPTION(nss_db_path, OPT_STR, "") // path to nss db
 
 
--- a/src/test/rbd_mirror/image_sync/test_mock_ImageCopyRequest.cc
+++ b/src/test/rbd_mirror/image_sync/test_mock_ImageCopyRequest.cc
@@ -50,7 +50,7 @@ struct ObjectCopyRequest<librbd::MockImageCtx> {
   Mutex lock;
   Cond cond;
 
-  const ImageCopyRequest<librbd::MockImageCtx>::SnapMap *snap_map;
+  const ImageCopyRequest<librbd::MockImageCtx>::SnapMap *snap_map = nullptr;
   std::map<uint64_t, Context *> object_contexts;
 
   ObjectCopyRequest() : lock("lock") {
@@ -119,7 +119,8 @@ public:
   }
 
   bool complete_object_copy(MockObjectCopyRequest &mock_object_copy_request,
-                            uint64_t object_num, int r) {
+                               uint64_t object_num, int r,
+                               std::function<void()> fn = []() {}) {
     Mutex::Locker locker(mock_object_copy_request.lock);
     while (mock_object_copy_request.object_contexts.count(object_num) == 0) {
       if (mock_object_copy_request.cond.WaitInterval(m_local_image_ctx->cct,
@@ -129,7 +130,12 @@ public:
       }
     }
 
-    m_threads->work_queue->queue(mock_object_copy_request.object_contexts[object_num], r);
+    FunctionContext *wrapper_ctx = new FunctionContext(
+      [&mock_object_copy_request, object_num, fn] (int r) {
+        fn();
+        mock_object_copy_request.object_contexts[object_num]->complete(r);
+      });
+    m_threads->work_queue->queue(wrapper_ctx, r);
     return true;
   }
 
@@ -224,6 +230,20 @@ TEST_F(TestMockImageSyncImageCopyRequest, Throttled) {
   ASSERT_EQ(0, create_snap("snap1"));
   m_client_meta.sync_points = {{"snap1", boost::none}};
 
+  std::string update_sync_age;;
+  ASSERT_EQ(0, _rados->conf_get("rbd_mirror_sync_point_update_age", update_sync_age));
+  ASSERT_EQ(0, _rados->conf_set("rbd_mirror_sync_point_update_age", "1"));
+  BOOST_SCOPE_EXIT( (update_sync_age) ) {
+    ASSERT_EQ(0, _rados->conf_set("rbd_mirror_sync_point_update_age", update_sync_age.c_str()));
+  } BOOST_SCOPE_EXIT_END;
+
+
+  std::string max_ops_str;
+  ASSERT_EQ(0, _rados->conf_get("rbd_concurrent_management_ops", max_ops_str));
+  int max_ops = std::stoi(max_ops_str);
+
+  uint64_t object_count = 55;
+
   librbd::MockImageCtx mock_remote_image_ctx(*m_remote_image_ctx);
   librbd::MockImageCtx mock_local_image_ctx(*m_local_image_ctx);
   journal::MockJournaler mock_journaler;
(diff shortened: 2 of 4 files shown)


src/tools/rbd_mirror/image_sync/ImageCopyRequest.cc
```diff
@@ -198,6 +200,11 @@ void ImageCopyRequest<I>::handle_object_copy(int r) {
       }
     }
 
+    if (!m_updating_sync_point &&
```
dillaman: I'd rather see the SafeTimer be utilized to periodically update the sync point every X seconds (new config option).

src/tools/rbd_mirror/image_sync/ImageCopyRequest.cc
```diff
@@ -210,6 +217,48 @@ void ImageCopyRequest<I>::handle_object_copy(int r) {
...
 
 template <typename I>
+void ImageCopyRequest<I>::send_update_sync_point() {
+  assert(m_lock.is_locked());
+
+  if (m_canceled || m_ret_val < 0) {
+    return;
+  }
+
+  update_progress("UPDATE_SYNC_POINT");
```
dillaman: No need to update the progress since this is an out-of-band state.

src/tools/rbd_mirror/image_sync/ImageCopyRequest.cc
```diff
@@ -210,6 +217,48 @@ void ImageCopyRequest<I>::handle_object_copy(int r) {
...
+  ldout(cct, 20) << ": r=" << r << dendl;
+
+  if (r < 0) {
+    *m_client_meta = m_client_meta_copy;
+    lderr(cct) << ": failed to update client data: " << cpp_strerror(r)
+               << dendl;
+  }
+
+  Mutex::Locker l(m_lock);
+  m_updating_sync_point = false;
```
dillaman: Need to handle the cases where the state machine completes / cancels while there is an in-flight sync point update. In such a case, you cannot complete the state machine until the sync point callback has been invoked / processed.

src/test/rbd_mirror/image_sync/test_mock_ImageCopyRequest.cc
```diff
@@ -396,6 +438,67 @@ TEST_F(TestMockImageSyncImageCopyRequest, Cancel) {
...
+
+  ASSERT_EQ(m_snap_map, wait_for_snap_map(mock_object_copy_request));
+
+  C_SaferCond ctx_0;
+  C_SaferCond ctx_1;
+  C_SaferCond ctx_2;
+
+  ASSERT_TRUE(complete_object_copy_sync(mock_object_copy_request, 0, 0, &ctx_0));
+  ASSERT_TRUE(complete_object_copy_sync(mock_object_copy_request, 1, 0, &ctx_1));
+  ASSERT_TRUE(complete_object_copy_sync(mock_object_copy_request, 2, 0, &ctx_2));
```
dillaman: There is a simpler way to inject the cancel request exactly when object 2 is copied -- modify MockObjectCopyRequest to add a dummy ```construct``` mock method that it's ```create``` method invokes.  Then you can EXPECT_CALL on that ```construct(<object number>)``` and invoke the cancel method directly from the expectation.

rjfd:
@dillaman addressed all your comments and pushed the new changes.

Regarding the unit test problem, I could not do the way you suggested as when the `create` method is called, we are holding the `m_lock` from `ImageCopyRequest` instance, and that prevent us from calling any other method that tries to acquire the same lock again.

dillaman:
In the Invoke, you should be able to enqueue a lambda on the op_work_queue to inject cancel and complete the provide Context.

rjfd:
In the last push, I changed the implementation of the test cases, and I think it's more simple now.

src/tools/rbd_mirror/image_sync/ImageCopyRequest.h
```diff
@@ -100,6 +100,11 @@ class ImageCopyRequest : public BaseRequest {
   uint64_t m_current_ops = 0;
   int m_ret_val = 0;
 
+  bool m_updating_sync_point;
+  Context *m_update_sync_ctx;
+  // TODO: make interval configurable
```
dillaman: Minor: TODO

src/tools/rbd_mirror/image_sync/ImageCopyRequest.cc
```diff
@@ -205,6 +220,88 @@ void ImageCopyRequest<I>::handle_object_copy(int r) {
   update_progress("COPY_OBJECT " + stringify(percent) + "%", false);
 
   if (complete) {
+    bool do_flush = true;
+    {
+      Mutex::Locker l(*m_timer_lock);
```
dillaman: Minor: l -> timer_locker

src/tools/rbd_mirror/image_sync/ImageCopyRequest.cc
```diff
@@ -146,7 +148,20 @@ void ImageCopyRequest<I>::send_object_copies() {
...
+    if (!complete) {
+      m_update_sync_ctx = new FunctionContext([this](int r) {
+          this->send_update_sync_point();
+      });
+    }
+  }
+
+  if (m_update_sync_ctx) {
+    Mutex::Locker l(*m_timer_lock);
+    m_timer->add_event_after(m_update_sync_point_interval,
```
dillaman: You might have to schedule this while holding ```m_timer_lock``` and ```m_lock``` to avoid a potential race if it completes when you drop the lock.

src/tools/rbd_mirror/image_sync/ImageCopyRequest.cc
```diff
@@ -205,6 +220,88 @@ void ImageCopyRequest<I>::handle_object_copy(int r) {
   update_progress("COPY_OBJECT " + stringify(percent) + "%", false);
 
   if (complete) {
+    bool do_flush = true;
+    {
+      Mutex::Locker l(*m_timer_lock);
+      Mutex::Locker locker(m_lock);
+      if (!m_updating_sync_point) {
+        if (m_update_sync_ctx != nullptr) {
+          m_timer->cancel_event(m_update_sync_ctx);
```
dillaman: m_update_sync_ctx = nullptr;

src/tools/rbd_mirror/image_sync/ImageCopyRequest.cc
```diff
@@ -205,6 +220,88 @@ void ImageCopyRequest<I>::handle_object_copy(int r) {
...
+template <typename I>
+void ImageCopyRequest<I>::send_update_sync_point() {
+  Mutex::Locker l(m_lock);
+
+  if (m_canceled || m_ret_val < 0 || m_current_ops == 0) {
+    return;
+  }
+
+  if (m_sync_point->object_number
+      && (m_object_no-1) == m_sync_point->object_number.get()) {
```
dillaman: Minor: move && to previous line

src/tools/rbd_mirror/image_sync/ImageCopyRequest.cc
```diff
@@ -205,6 +220,88 @@ void ImageCopyRequest<I>::handle_object_copy(int r) {
...
+    if (schedule_update) {
+      m_update_sync_ctx = new FunctionContext([this](int r) {
+          this->send_update_sync_point();
+      });
+    }
+  }
+
+  if (schedule_update) {
+    Mutex::Locker l(*m_timer_lock);
+    m_timer->add_event_after(m_update_sync_point_interval, m_update_sync_ctx);
```
dillaman: I think this would be the same race condition if it completed as soon as you dropped the lock.

src/tools/rbd_mirror/image_sync/ImageCopyRequest.h
rjfd: Should I use the `md_config_obs_t` and create a new config option for this?

src/tools/rbd_mirror/image_sync/ImageCopyRequest.cc
rjfd: I don't think I need to hold both locks, although I need to acquire `m_timer_lock` before testing `m_update_sync_ctx != nullptr`.
If the sync completes, and the code starting at line 227 starts to execute,  before line 158, then the code starting in line 227 (which is holding `m_timer_lock`), will assign `m_update_sync_ctx = nullptr`. Then after line 158 we acquire the `m_timer_lock` but we will not call `add_event_after` anymore.

src/tools/rbd_mirror/image_sync/ImageCopyRequest.cc
rjfd: In this case I have to acquire `m_timer_lock` and then test if `m_update_sync_ctx != nullptr`, because it could already have been assign to `nullptr` in line 232 (after pushing the new changes).

src/tools/rbd_mirror/image_sync/ImageCopyRequest.cc
dillaman: Well, if ```complete``` was false on line 150 and when you drop the lock ```handle_object_copy``` invokes and discovers it should complete. At this point, ```m_update_sync_ctx``` is not null, ```handle_object_copy``` cancels the event (which hasn't been scheduled yet), sets it to null (future revision), and continues.  Therefore, you would need to grab the timer lock before testing m_update_sync_ctx for null  on like 159.

rjfd: Yep, that's what I said in my comment above, and already fixed in the implementation, although still didn't pushed the changes

rjfd:
@dillaman done pushing new changes

src/common/config_opts.h
```diff
@@ -1235,6 +1235,11 @@ OPTION(rbd_journal_object_flush_bytes, OPT_INT, 0) // maximum number of pending
 OPTION(rbd_journal_object_flush_age, OPT_DOUBLE, 0) // maximum age (in seconds) for pending commits
 OPTION(rbd_journal_pool, OPT_STR, "") // pool for journal objects
 
+/**
+ * RBD Mirror options
+ */
+OPTION(rbd_mirror_sync_point_update_interval, OPT_DOUBLE, 1) // number of seconds between each update of the image sync point object number
```
dillaman: Minor: can we switch "interval" to "age" and bump the default to something like 30 seconds?

src/tools/rbd_mirror/image_sync/ImageCopyRequest.cc
```diff
@@ -146,7 +149,22 @@ void ImageCopyRequest<I>::send_object_copies() {
...
+    if (!complete) {
+      m_update_sync_ctx = new FunctionContext([this](int r) {
+          this->send_update_sync_point();
+      });
+    }
   }
+
+  {
+    Mutex::Locker timer_locker(*m_timer_lock);
+    if (m_update_sync_ctx) {
```
dillaman: I think this is going to cause helgrind warning since you sometimes hold only ```m_timer_lock``` when touching ```m_update_sync_ctx```, sometimes you hold ```m_lock```, and sometimes you hold both.

rjfd:
@dillaman pushed new changes

dillaman:
@rjfd Did you change anything?

rjfd:
@dillaman yes, I changed the name and default value of config option. I also ran the unit test against helgrind but didn't got any warning.

rjfd:
@dillaman rebased on top of master

PR #9701, opened 2016-06-14
OSD.cc: remove unneeded return

stiopaa1:
There is no need for a return in function returning void.


xiexingguo:
Reviewed-by:

PR #9702, opened 2016-06-14
Python 3 compatibility of Ceph CLI

BlaXpirit:
This replaces tabs with whitespace, fixes syntax and builtins (most controversial change would be `.iteritems()` -> `.items()` but should be fine).

There was also one bug with the validity check that was undetected on Python 2.

Compatibility down to Python 2.6 remains.

### DIFF
--- a/src/ceph.in
+++ b/src/ceph.in
@@ -81,13 +81,13 @@ if MYDIR.endswith('src') and \
 
     python_libpath = os.path.join(MYDIR, 'build', get_pythonlib_dir())
     respawn_in_path(os.path.join(MYDIR, '.libs'), 'pybind', python_libpath)
-    if os.environ.has_key('PATH') and MYDIR not in os.environ['PATH']:
+    if 'PATH' in os.environ and MYDIR not in os.environ['PATH']:
         os.environ['PATH'] += ':' + MYDIR
 
 elif os.path.exists(os.path.join(os.getcwd(), "CMakeCache.txt")) \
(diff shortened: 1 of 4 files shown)


BlaXpirit:
What happened to the check? I don't see what went wrong from the output.

jdurgin:
The check failure is unrelated - it's one that snuck in recently and appears in the bot sometimes but has eluded manual reproduction so far.

jdurgin:
The changes look good, nice and straightforward. Pushed to wip-ceph-argparse-py3 and will run that through the rados suite once it's built.

BlaXpirit:
Don't merge yet, I've discovered other needed changes when porting the `ceph` script. Maybe it should be one bigger pull request?

jdurgin:
alright, sounds like a good idea to make a larger pr since the ceph_argparse tests don't cover as much as the cli tests do

BlaXpirit:
I continued in this pull request. 

This QA suite finished successfully for me, ran like this (from *src*):

```console
$ PATH="$PATH:$(readlink -f .)" PYTHON=python3 ../qa/workunits/cephtool/test.sh --asok-does-not-need-root
```

It doesn't work from just the default environment, because it also needs *rados* and *cephfs* bindings built for Python 3 specifically (I put their *.py/.so* files right inside *pybind* but there is probably a better way).

jdurgin:
lgtm, as discussed on irc we can merge this now since it doesn't break python 2, and fix any remaining python3 issues once we get python3 integration tests working

PR #9705, opened 2016-06-15
librados: restore ability to compile against librados

dillaman:
Partial revert of 62be9268de5e9c9a08bdb977a7dab1ab9c55b2be which
added a new header dependency which was is not part of the librados
API.

### DIFF
--- a/src/include/rados/librados.h
+++ b/src/include/rados/librados.h
@@ -19,7 +19,6 @@
 extern "C" {
 #endif
 
-#include "../compat.h"
 #include <netinet/in.h>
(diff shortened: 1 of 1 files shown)


dillaman:
It's currently impossible to compile against librados due to this missing dependency -- as a result all RBD fio tests are failing in teuthology tests.

jdurgin:
@wjwithagen you'll likely need to add compat.h to just the internal parts of librados (src/librados), or if necessary to compile librados public headers, make a public header for the parts of compat.h that are required in src/include/rados

PR #9707, opened 2016-06-15
DNM: test: Handle object removals in a non-racey way

dzafman:


dzafman:
This commit goes into #9304 because that branch re-enables the osd-scrub-repair.sh test.

PR #9708, opened 2016-06-15
script/make_rocksdb_makefile.sh: do not include rocksdb/tools/rdb/*

tchaikov:
Fixes: http://tracker.ceph.com/issues/13554

### DIFF
--- a/src/script/make_rocksdb_makefile.sh
+++ b/src/script/make_rocksdb_makefile.sh
@@ -4,7 +4,7 @@ cd rocksdb
 git clean -dffx
 cd ..
 echo "EXTRA_DIST += \\" > /tmp/$$
-for f in `find rocksdb -type f | grep -v /.git$ | sort`; do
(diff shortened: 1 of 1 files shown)


tchaikov:
@dmick

dmick:
LGTM; if the test passes I'm good

tchaikov:
but if the test fails, it won't be caused by this change, right?

dmick:
well, this should prevent the test from failing for this reason (if it runs on an RPM system) and have no effect (if it runs on a deb system), so it ought to pass with respect to this bug, right?

dmick:
(and it seems to be running on trusty right at the moment)

tchaikov:
but "make_rocksdb_makefile.sh" is not used by "make check", it is launched manually i think. `git grep make_rocksdb_makefile.sh` does not show anything for me.

dmick:
ugh.  :(

liewegas:
I run this manually when updating the rocksdb submodule commit.  If we don't need those files in teh tarball, though, then this looks good!

dmick:
we not only don't need them, they do damage.

PR #9710, opened 2016-06-15
Server: avoid duplicated call to acquire_locks()

xiexingguo:
For snapped case we may call acquire_locks() twice,
which is not necessary.


ukernel:
The change looks good.  But the commit message is incorrect. The second Locker::acquire_locks does nothing

xiexingguo:
@ukernel  Fixed and thanks!

PR #9711, opened 2016-06-15
mon: monitor time in future, use relative time compute lease_expire 

cxwshawn:
if one  monitor host time is in future[+5s], lease expired immediately, so why could not change
peon monitor's lease_expired to not use the leader's lease_expire, but compute this by
using own local time ? 


liewegas:
Because we don't know how long the MMonLease message transit time is, and we don't want to overestimate the duration of the lease (and think it is valid when it is not).

PR #9713, opened 2016-06-15
server: drop locks and auth pins before waiting for trunc

xiexingguo:

### DIFF
--- a/src/mds/Server.cc
+++ b/src/mds/Server.cc
@@ -3750,6 +3750,8 @@ void Server::handle_client_setattr(MDRequestRef& mdr)
     if (truncating_smaller && pi->is_truncating()) {
       dout(10) << " waiting for pending truncate from " << pi->truncate_from
 	       << " to " << pi->truncate_size << " to complete on " << *cur << dendl;
+      mds->locker->drop_locks(mdr.get());
+      mdr->drop_local_auth_pins();
       cur->add_waiter(CInode::WAIT_TRUNC, new C_MDS_RetryRequest(mdcache, mdr));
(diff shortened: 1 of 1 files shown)


src/mds/Server.cc
```diff
@@ -3750,6 +3750,8 @@ void Server::handle_client_setattr(MDRequestRef& mdr)
     if (truncating_smaller && pi->is_truncating()) {
       dout(10) << " waiting for pending truncate from " << pi->truncate_from
 	       << " to " << pi->truncate_size << " to complete on " << *cur << dendl;
+      mds->locker->drop_locks(mdr.get());
```
jcsp: This doesn't seem unreasonable, but why do it here and not in the other place we use WAIT_TRUNC above?

ukernel:
looks good

xiexingguo:
@ukernel Thanks!

xiexingguo:
@jcsp Thanks for the testing and merging:-)

gregsfortytwo:
I guess I missed the window, but can somebody explain why it's okay to drop locks here? Amongst other things, we've already succeeded in locking and the mdr remembers that, so it won't try to acquire the locks again on the next go-round...

gregsfortytwo:
Oh, I guess we do clear out the done_locking flag. So it's just a question of whether we're okay re-ordering ops. I don't really think we are since we've already started removing the file data, but we haven't changed the truncate sizes yet?

gregsfortytwo:
Sorry for the noise, happy now.
https://github.com/ceph/ceph/pull/9716

PR #9715, opened 2016-06-15
librbd: re-register watch on old format image rename

trociny:
The watching object name is changed when renaming an old format image,
so unregister the watcher before the rename, and register back after,
to avoid "Transport endpoint is not connected" error.

Fixes: http://tracker.ceph.com/issues/16321


trociny:
@dillaman How do you like tit? It the idea is right but the code looks a little ugly, I  can e.g. wrap this into OldFormatRenameRequest class in anonymous name space.

dillaman:
@trociny Looks good to me.  I am actually wondering whether or not we should also disallow renames on v1 images that are in-use.

PR #9717, opened 2016-06-15
mon: do not recalculate 'to_remove' when it's known

songbaisen:
mon: do not recalculate 'to_remove' when it's known


tchaikov:
@songbaisen the change looks good to me.

but could you revise your commit message (and title) to be more specific? imho, "small change in paxos assignment" is not very helpful. something like "do not recalculate 'to_remove' when it's known" would be better.

songbaisen:
@tchaikov  Thank you sir! Done it!

tchaikov:
lgtm.

tchaikov:

PR #9720, opened 2016-06-15
qa/workunits: rados cppool now requires --yes-i-really-really-mean-it

dillaman:
Fixes: http://tracker.ceph.com/issues/16202


trociny:
http://qa-proxy.ceph.com/teuthology/trociny-2016-06-16_02:00:12-rbd-wip-mgolub-testing---basic-mira/263499/teuthology.log

PR #9721, opened 2016-06-15
jewel: qa/workunits/rbd: respect RBD_CREATE_ARGS environment variable

trociny:
http://tracker.ceph.com/issues/16324

### DIFF
--- a/qa/workunits/rbd/import_export.sh
+++ b/qa/workunits/rbd/import_export.sh
@@ -137,7 +137,7 @@ rbd rm sparse
 
 echo "zeros export to sparse file"
 #  Must be tricky to make image "by hand" ; import won't create a zero image
-rbd create sparse --size 4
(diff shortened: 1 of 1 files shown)


dachary:
@trociny we would usually wait until at least one tests passes to confirm the backport is correct and ask for confirmation to the person who merged the original commit in master (that would be @dillaman) in this case.

trociny:
@dachary Sorry, it is not clear from this PR, because I have not left any notes.

I tested this on teuthology, although many days ago. Have you seen any problems after the merge? Actually, this is a specific case, and the fix was needed for hammer branch (already merged many days ago, see the original tracket ticket, it is observed in teuthology upgrade test, when upgrading from hammer to jewel, and teuthology uses hammer branch there). I merged this to jewel just to not diverge.

trociny:
@dachary Sure will follow the procedure more strictly the next time.

PR #9726, opened 2016-06-15
msg: add entity_addrvec_t type

liewegas:
This is a vector of entity_addr_t's that can be swapped in for a legacy entity_addr_t.  The marker that precedes a (newly encoded) entity_addr_t is 1 and the addrvec's is 2, which means we can swap in an addrvec for an addr at any time in the future.


liewegas:
failing make check; i'll reopen once it passes

PR #9730, opened 2016-06-15
rbd-mirror: gracefully handle missing sync point snapshots

dillaman:


trociny:
@dillaman [DNM] prefix may be removed from the title?

qa/workunits/rbd/rbd_mirror_helpers.sh
```diff
@@ -408,14 +408,28 @@ wait_for_replay_complete()
...
+
+        test "${mirror_pos}" = "${last_mirror_pos}" && return 1
+        last_mirror_pos="${mirror_pos}"
+
+        # handle the case where the mirror is ahead of the master
+        master_tag=$(echo "${master_pos}" | grep -Eo "tag_tid=[0-9]*" | cut -d'=' -f 2)
+        mirror_tag=$(echo "${mirror_pos}" | grep -Eo "tag_tid=[0-9]*" | cut -d'=' -f 2)
+        master_entry=$(echo "${master_pos}" | grep -Eo "entry_tid=[0-9]*" | cut -d'=' -f 2)
+        mirror_entry=$(echo "${mirror_pos}" | grep -Eo "entry_tid=[0-9]*" | cut -d'=' -f 2)
+        test "${master_tag}" = "${mirror_tag}" -a ${master_entry} -le ${mirror_entry} && return 0
```
trociny: In future it would be nice to add position compare function.

dillaman:
@trociny If you already tested it, I'm fine with tanking the DNM. I was hoping to figure out how the system could get into such a state (assuming you didn't manually delete the snapshot), but I can create a new ticket/PR if and when that happens.

trociny:
@dillaman testing is still in progress, I will let you know when have results

trociny:
lgtm

PR #9738, opened 2016-06-16
monstoretool: make it more clearly to user why it can not be used

songbaisen:
monstoretool: make it more clearly to user why it can not be used

like below:
    The database have already locked by PID: '420170' task name: '/usr/bin/ceph-mon' please close it first!

### DIFF
--- a/src/tools/ceph_monstore_tool.cc
+++ b/src/tools/ceph_monstore_tool.cc
@@ -709,7 +709,7 @@ void get_name_by_pid(pid_t pid, char *task_name)
 
   if (fd < 0) {
     fd = -errno;
-    std::cerr << "Fail to open '" << proc_pid_path
+    std::cerr << "Failed to open '" << proc_pid_path
               << "' error = " << cpp_strerror(fd)
               << std::endl;
     return;
@@ -718,9 +718,10 @@ void get_name_by_pid(pid_t pid, char *task_name)
   int ret = read(fd, task_name, PATH_MAX-1);
   if (ret < 0) {
     ret = -errno;
-    std::cerr << "Fail to read '" << proc_pid_path
+    std::cerr << "Failed to read '" << proc_pid_path
               << "' error = " << cpp_strerror(ret)
               << std::endl;
+    close(fd);
   }
 
   close(fd);
@@ -747,10 +748,10 @@ int check_lock(const string &path) {
   int fd = open(lock_path.c_str(), O_RDONLY);
   if (fd < 0) {
     fd = -errno;
-    std::cerr << "Fail to open '" << lock_path
+    std::cerr << "Failed to open '" << lock_path
               << "' error = " << cpp_strerror(fd)
               << std::endl;
-    return -1;
+    return -errno;
   }
 
   struct flock f;
@@ -759,8 +760,9 @@ int check_lock(const string &path) {
   f.l_whence = SEEK_SET;
 
   if (-1 == fcntl(fd, F_GETLK, &f)) {
-    std::cerr << " Fail to get the lock " << cpp_strerror(errno) << std::endl;
-    return -1;
+    std::cerr << " Failed to get the lock " << cpp_strerror(errno) << std::endl;
+    close(fd);
+    return -errno;
   }
 
   if (f.l_type != F_UNLCK) {
@@ -771,10 +773,12 @@ int check_lock(const string &path) {
       cout << "The database have already locked by PID: '"
            << f.l_pid << "' task name: '" << task_name
            << "' please close it first!" << std::endl;
-      return -1;
+      close(fd);
+      return -EACCES;
     }
   }
 
+  close(fd);
   return 0;
 }
 
@@ -871,8 +875,9 @@ int main(int argc, char **argv) {
   g_ceph_context->_conf->apply_changes(NULL);
   g_conf = g_ceph_context->_conf;
 
-  if (check_lock(store_path))
-    return 1;
+  int ret = check_lock(store_path);
+  if (ret)
+    return ret;
   // this is where we'll write *whatever*, on a per-command basis.
   // not all commands require some place to write their things.
   MonitorDBStore st(store_path);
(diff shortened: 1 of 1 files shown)


src/tools/ceph_monstore_tool.cc
```diff
@@ -701,6 +701,83 @@ int rebuild_monstore(const char* progname,
...
 
+void get_name_by_pid(pid_t pid, char *task_name)
+{
+  char proc_pid_path[PATH_MAX] = {0};
+  snprintf(proc_pid_path, PATH_MAX, "/proc/%d/cmdline", pid);
+  int fd = open(proc_pid_path, O_RDONLY);
+
+  if (fd < 0) {
+    fd = -errno;
+    std::cerr << "Fail to open '" << proc_pid_path
```
jecluis: s/Fail/Failed/

Same for the other 'Fail' strings.

src/tools/ceph_monstore_tool.cc
```diff
@@ -701,6 +701,83 @@ int rebuild_monstore(const char* progname,
...
+     return 0;
+  }
+
+  int fd = open(lock_path.c_str(), O_RDONLY);
+  if (fd < 0) {
+    fd = -errno;
+    std::cerr << "Fail to open '" << lock_path
+              << "' error = " << cpp_strerror(fd)
+              << std::endl;
+    return -1;
```
jecluis: return `-errno` instead (or, in this case, `fd`, as it is already negative).

src/tools/ceph_monstore_tool.cc
```diff
@@ -701,6 +701,83 @@ int rebuild_monstore(const char* progname,
...
+  }
+
+  struct flock f;
+  memset(&f, 0, sizeof(f));
+  f.l_type = F_WRLCK;
+  f.l_whence = SEEK_SET;
+
+  if (-1 == fcntl(fd, F_GETLK, &f)) {
+    std::cerr << " Fail to get the lock " << cpp_strerror(errno) << std::endl;
+    return -1;
```
jecluis: should return `-errno` instead.

src/tools/ceph_monstore_tool.cc
```diff
@@ -701,6 +701,83 @@ int rebuild_monstore(const char* progname,
...
+
+  if (f.l_type != F_UNLCK) {
+    if (f.l_type == F_WRLCK) {
+      char task_name[PATH_MAX] = "unknown";
+      //get the locked thread name
+      get_name_by_pid(f.l_pid, task_name);
+      cout << "The database have already locked by PID: '"
+           << f.l_pid << "' task name: '" << task_name
+           << "' please close it first!" << std::endl;
+      return -1;
```
jecluis: return `-EBUSY`?

jecluis: or maybe `-EACCES`?

src/tools/ceph_monstore_tool.cc
```diff
@@ -794,6 +871,8 @@ int main(int argc, char **argv) {
   g_ceph_context->_conf->apply_changes(NULL);
   g_conf = g_ceph_context->_conf;
 
+  if (check_lock(store_path))
+    return 1;
```
jecluis: we should return something of value to the user. maybe propagate the return from `check_lock()` as the program's return code?

src/tools/ceph_monstore_tool.cc
```diff
@@ -701,6 +701,83 @@ int rebuild_monstore(const char* progname,
...
+  ostringstream os;
+  os << path.substr(0, path.size() - pos) << "/store.db/LOCK";
+  string lock_path = os.str();
+
+  //The lock file is not existed.
+  if (access(lock_path.c_str(), F_OK)) {
+     return 0;
+  }
+
+  int fd = open(lock_path.c_str(), O_RDONLY);
```
jecluis: this `fd` is never closed in the function. While this is fine if the program exits to user-space, it feels weird and is definitely not ideal if our checks are successful (and thus we continue our work).

songbaisen:
@jecluis  Done it!

src/tools/ceph_monstore_tool.cc
```diff
@@ -701,6 +701,87 @@ int rebuild_monstore(const char* progname,
...
+    std::cerr << " Failed to get the lock " << cpp_strerror(errno) << std::endl;
+    close(fd);
+    return -errno;
+  }
+
+  if (f.l_type != F_UNLCK) {
+    if (f.l_type == F_WRLCK) {
+      char task_name[PATH_MAX] = "unknown";
+      //get the locked thread name
+      get_name_by_pid(f.l_pid, task_name);
```
tchaikov: i don't think the tool is responsible to report who is holding the lock. what it should inform the user is just 

> another process is holding the exclusive lock on a certain file with given path
> (show the specific path here)

and nothing more.

songbaisen: @tchaikov  But I think it is would be better for user who use this tool. when the tool can not use, he can very quickly and clearly know which process handler the lock.and save his time to find the process.

tchaikov: i think this is a gray area, but i am not sure how hard it would be for an administrator to use `fuser(1)` to determine which process is holding the lock. if it's difficult to connect the error message to `fuser`, maybe we should put this into our man page. but i don't think we should embed fuser in whoever want to lock a file. and we are using flock in filestore and bluestore also. shall we add this functionality to them as well? i doubt it.

songbaisen: @tchaikov en . I know a lot of administrator  who use ceph but really do not very family with linux. The reason why I add this commit, it is because I have seen a  lot of user ,who use this tool, but do not know why they can use it through the QQ group of ceph . It have waste they some time to find  out the reason.

tchaikov: okay, then i'd recommend update the man page instead.

songbaisen: @tchaikov  Agree with it too!

PR #9742, opened 2016-06-16
mds/server: fix rare race when waitting for osdmap

xiexingguo:
If wait_for_map() returns true, we have got the requested osdmap
at the specified epoch, so we shall try again to do further
verification whether we have the specific pool or not.
    
The above case can happen because we drop the objecter internal
rwlock during the switch between the objecter->with_osdmap()
and objecter->wait_for_map() methods.
    
### DIFF
--- a/src/mds/Server.cc
+++ b/src/mds/Server.cc
@@ -3038,8 +3038,7 @@ void Server::handle_client_open(MDRequestRef& mdr)
   }
   
   // hit pop
-  if (cmode == CEPH_FILE_MODE_RDWR ||
-      cmode == CEPH_FILE_MODE_WR) 
+  if (cmode & CEPH_FILE_MODE_WR)
     mds->balancer->hit_inode(mdr->get_mds_stamp(), cur, META_POP_IWR);
   else
     mds->balancer->hit_inode(mdr->get_mds_stamp(), cur, META_POP_IRD,
--- a/src/mds/Server.cc
+++ b/src/mds/Server.cc
@@ -4157,6 +4157,76 @@ int Server::parse_quota_vxattr(string name, string value, quota_info_t *quota)
   return 0;
 }
 
+/*
+ * Verify that the file layout attribute carried by client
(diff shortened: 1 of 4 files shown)


src/mds/Server.cc
```diff
@@ -4264,10 +4286,14 @@ void Server::handle_set_vxattr(MDRequestRef& mdr, CInode *cur,
...
-      respond_to_request(mdr, r);
-      return;
+
+      if (r < 0) { // check again, result code could have changed
+        if (r == -ENOENT)
+          r = -EINVAL;
+        respond_to_request(mdr, r);
+        return;
+      }
     }
```
ukernel: lots of duplicated code here, maybe we move them into seperate function

src/mds/Server.cc
```diff
@@ -4501,7 +4501,7 @@ void Server::handle_client_removexattr(MDRequestRef& mdr)
   set<SimpleLock*> rdlocks, wrlocks, xlocks;
   file_layout_t *dir_layout = NULL;
   CInode *cur;
-  if (name == "ceph.dir.layout")
+  if (name.compare(0, 15, "ceph.dir.layout") == 0)
```
ukernel: But inside handle_remove_vxattr(), it still checks name == "ceph.dir.layout"

xiexingguo:
@ukernel Thank you for your careful review.

I have done upgrades based on your gorgeous advice, could you please have a look again?

xiexingguo:
retest this please

jcsp:
@ukernel could you review the updated code please?

xiexingguo:
Tested at:

ukernel:
please remove commit "server: handle attribute name in a more graceful way".  handle_remove_vxattr only support removing whole dir.layout

xiexingguo:
@ukernel Done. Thanks!

xiexingguo:
@ukernel Thanks for the merge, zheng:-)

PR #9743, opened 2016-06-16
jewel : rgw: support size suffixes for --max-size in radosgw-admin command

vumrao:
jewel : rgw: support size suffixes for --max-size in radosgw-admin command
Fixes : http://tracker.ceph.com/issues/16339

### DIFF
--- a/doc/radosgw/admin.rst
+++ b/doc/radosgw/admin.rst
@@ -310,7 +310,7 @@ storage size in megabytes.
   the maximum number of objects. A negative value disables this setting.
   
 - **Maximum Size:** The ``--max-size`` option allows you to specify a quota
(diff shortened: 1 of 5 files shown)


dachary:
@cbodley does this backport look good to merge ? It passed the rgw suite http://tracker.ceph.com/issues/16344#note-12, except for one transient error which seems unrelated to this backport.

cbodley:
@dachary it looks like we'll want to backport the fix from #9934 as well. can @vumrao cherry-pick that into this branch, or should it be tracked as a separate backport?

vumrao:
@cbodley #9934 is getting backported to jewel in backport tracker : http://tracker.ceph.com/issues/16657

PR #9747, opened 2016-06-16
rgw: should clear 'short_zone_ids' when update period

weiqiaomiao:
when delete a zone from zonegroup, update the period and commit the changes,
the zone remove from current periodmap, but the short_zone_ids of zone still in
'short_zone_ids' map of the periodmap, so we should remove it.

### DIFF
--- a/src/rgw/rgw_rados.h
+++ b/src/rgw/rgw_rados.h
@@ -1189,6 +1189,7 @@ struct RGWPeriodMap
   void reset() {
     zonegroups.clear();
     zonegroups_by_api.clear();
+    short_zone_ids.clear();
   }
 
   uint32_t get_zone_short_id(const string& zone_id) const;


theanalyst:
the short zone ids are still used on objects and used as a tie-breaker for sync etc. see http://tracker.ceph.com/issues/15618

mattbenjamin:
@cbodley add a note if this shouldn't be closed :)

cbodley:
@weiqiaomiao i proposed a very similar pr https://github.com/ceph/ceph/pull/8769 that led to some good discussion, and it was decided that we do want to keep the short ids around

because we store the zone's short id with objects it creates, we want to avoid adding a new zone with the same short id. we need to remember old short ids so that we can prevent these conflicts - in the end, https://github.com/ceph/ceph/pull/8886 was merged to accomplish that

PR #9748, opened 2016-06-16
rgw/user: remove needless bufferlist encoding

Yan-waller:
`bl` encoded but never be used.


mattbenjamin:
@Yan-waller thanks for sending these;  for future, a better commit message heading might be "rgw: remove unused buffer::list variable and encoding"

PR #8964, opened 2016-05-06
global: log which process/command sent a signal

songbaisen:
global: log which process/command sent a signal

The below url is the before discuss.
https://github.com/ceph/ceph/pull/6865

Signed-off-by:song baisen <song.baisen@zte.com.cn>

PR #9749, opened 2016-06-16
mon: remove the redundant is_active judge in PaxosService

songbaisen:
mon: remove the redundant is_active judge in PaxosService

In the above line 249 we have judged is_active method.So all the judge below that 
are redundant. 


athanatos:
lgtm pending test

songbaisen:
@tchaikov  @athanatos  @yuriw  Thanks to all! :smile:  Maybe this pr https://github.com/ceph/ceph/pull/8964 also need test.

PR #9752, opened 2016-06-16
jewel: librbd: cancel all tasks should wait until finisher is done

dillaman:
http://tracker.ceph.com/issues/16518


dachary:
It passed the rbd suite http://tracker.ceph.com/issues/16344#note-11 (except for two failures that seem unrelated http://tracker.ceph.com/issues/16922 and http://tracker.ceph.com/issues/16921)

PR #9754, opened 2016-06-16
rgw/rgw_http_client: add compat.h include for TEMP_FAILURE_RETRY

dx9:


wjwithagen:
include compat.h as very first include file.
especially before any boost-libraries, since they tend to define ENODATA on non linux platforms to a value that is not compatible with the xattr-functions. And will cause very unexpected problems.

mattbenjamin:
lgtm (build verified, f23)

PR #9756, opened 2016-06-16
os/bluestore: remove unused bluestore fields

liewegas:


allensamuels:
Reviewed by Allen Samuels <allen.samuels@sandisk.com>

PR #9759, opened 2016-06-16
rbd-mirror: ensure replay status formatter has completed before stopping

dillaman:
Fixes: http://tracker.ceph.com/issues/16352


trociny:
lgtm

PR #9761, opened 2016-06-16
cmake: _FORTIFY_SOURCE built-in definition fix

dx9:


src/CMakeLists.txt
```diff
@@ -25,7 +25,7 @@ include(CheckCCompilerFlag)
 CHECK_C_COMPILER_FLAG("-Wp,-D_FORTIFY_SOURCE=2" HAS_FORTIFY_SOURCE)
```
alimaredia: @dx9 Is there a reason -U_FORTIFY_SOURCE wasn't added here?

dx9:
@alimaredia Nope. Just didn't realize that the feature check failing when warnings exist is the better solution. Closing in favor of #9791. Thanks for looking at it.

PR #9762, opened 2016-06-16
fast containers header file.

allensamuels:


liewegas:
https://github.com/ceph/ceph/pull/11309

PR #9763, opened 2016-06-16
compat: add abstractions for non portable pthread name funcs

dx9:


wjwithagen:
@dx9 
If you are going to take a whack at this in this way, you also need to take care of the fact that in FreeBSD the name is spelled different and compatibility is created thru include/compat.h.
The way you have now organised it, this will not get pthread_setname_np() calls in the FreeBSD version

configure.ac
```diff
@@ -1145,6 +1145,7 @@ LIBS="$PTHREAD_LIBS $LIBS"
 CFLAGS="$PTHREAD_CFLAGS $CFLAGS"
 AC_CHECK_FUNC([pthread_spin_init],
   [AC_DEFINE(HAVE_PTHREAD_SPINLOCK, 1, [Define if you have pthread_spin_init])])
+AC_CHECK_FUNCS([pthread_setname_np pthread_setname_np])
```
yuyuyu101: pthread_setname_np  pthread_getname_np

wjwithagen:
@dx9 @yuyuyu101 
Please note that FreeBSD does not have pthread_getname_np
And that pthread_setname_np is actually called: pthread_set_name_np
Note the extra underscore.

And checking for pthread_setname_np is thus going to fail on FreeBSD,
resulting in the names of the threads not being set.
Unless you can get autoconf to actually set the correct variable when also testing for pthread_set_name_np

wjwithagen:
@dx9 @yuyuyu101
Other alternative would be to include the definition of the HAVE_* in compat.h in the FreeBSD.
But that is sort of weird, since that is hardcore overruling what acconfig.h is trying do do.

dx9:
@wjwithagen @yuyuyu101 Thanks for looking it over. Sneaky underscore! Totally missed that.

Pulled compat.h out of this line because we aren't defining anything that is missing. I don't really know if that is right or not. Just what happened when I looked at it. 

LMK

wjwithagen:
Please do not remove the compat.h file.
It is there to actually to keep portability, but reduce the amount of ifdef's 
So your code is going against the tide, and has grown more ifdef's than was needed.

So if you do that you do not have to have any ifdefs for pthread_setname_np.
Which is the way I ported it to FreeBSD

Now pthread_getname_np is a different cookie, but that I also intended to fix in compat.h
And then remove the ifdef's
Now if you do not like the ifdef's FreeBSD around pthread_getname_np usage than I guess that you could replace the conditional but there is very little need to that. Unless you are aiming to port to another environment .....

src/include/compat.h
```diff
@@ -110,4 +112,18 @@
...
+#elif defined(HAVE_PTHREAD_SET_NAME_NP)
+#define PTHREAD_SETNAME_NP(thread, name) pthread_set_name_np(thread, name)
+#else
+#define PTHREAD_SETNAME_NP(thread, name)
+#endif
+
+#if defined(HAVE_PTHREAD_GETNAME_NP)
+#define PTHREAD_GETNAME_NP(thread, name, len) pthread_getname_np(thread, name, len)
+#else
+#define PTHREAD_GETNAME_NP(thread, name, len)
```
wjwithagen: @dx9 
You might want to have the PTHREAD_GETNAME_NP macro return 0 and have it the first byte of name to '\0'
You can then remove the conditional around the usuage of this macro

ukernel:
Please consider OSX case.  pthread_setname_np() only accepts a single argument. It is only possible to set thread name for the current thread.

CMakeLists.txt
```diff
@@ -368,6 +368,12 @@ include_directories(${Boost_INCLUDE_DIRS})
 include_directories(${PROJECT_BINARY_DIR}/include)
 
 find_package(Threads REQUIRED)
+# Check for non portable pthread extensions
+set(CMAKE_REQUIRED_LIBRARIES ${CMAKE_THREAD_LIBS_INIT})
```
tchaikov: i'd recommend 
```cmake
list(APPEND CMAKE_REQUIRED_LIBRARIES ${CMAKE_THREAD_LIBS_INIT})
list(APPEND CMAKE_REQUIRED_DEFINITIONS -D_GNU_SOURCE)
```
and 
```cmake
list(REMOVE_ITEM CMAKE_REQUIRED_LIBRARIES ${CMAKE_THREAD_LIBS_INIT})
list(REMOVE_ITEM CMAKE_REQUIRED_DEFINITIONS -D_GNU_SOURCE)
```

before and after this test.

src/client/SyntheticClient.cc
```diff
@@ -934,7 +935,9 @@ int SyntheticClient::start_thread()
 
   pthread_create(&thread_id, NULL, synthetic_client_thread_entry, this);
   assert(thread_id);
+#ifdef HAVE_PTHREAD_SETNAME_NP
```
tchaikov: could define an empty `pthread_setname_np` in "compat.h" if `HAVE_PTHREAD_SETNAME_NP` is not `defined` like

```c++
#ifndef HAVE_PTHREAD_SETNAME_NP
#define pthread_setname_np(...) 
#endif
```

src/common/WorkQueue.cc
```diff
@@ -95,7 +97,7 @@ void ThreadPool::worker(WorkThread *wt)
   
   std::stringstream ss;
   char name[16] = {0};
-#if !defined(__FreeBSD__)
+#ifdef HAVE_PTHREAD_GETNAME_NP
```
tchaikov: could squash this commit into 86ac55467fbca1cb8dc6e243ea06e7898b049735

src/global/signal_handler.cc
```diff
@@ -90,7 +92,7 @@ static void handle_fatal_signal(int signum)
   // presumably dump core-- will handle it.
   char buf[1024];
   char pthread_name[16] = {0}; //limited by 16B include terminating null byte.
-#if !defined(__FreeBSD__)
+#ifdef HAVE_PTHREAD_GETNAME_NP
   int r = pthread_getname_np(pthread_self(), pthread_name, sizeof(pthread_name));
```
tchaikov: again, could have an empty definition which evaluates to "0" for this function if `HAVE_PTHREAD_GETNAME_NP` is not defined.

dx9:
@tchaikov Would it be acceptable to squash this down into a single commit with "compat: add abstractions for pthread_setname_np() and pthread_getname_np()"? Then everyone can review again and we can make revisions. Originally kept them separate so the history was there for how we reached a consensus.

tchaikov:
@dx9 sorry, i mistook them as the same one.

src/include/compat.h
```diff
@@ -110,4 +112,25 @@
 #define MSG_DONTWAIT MSG_NONBLOCK
 #endif
 
+#if defined(HAVE_PTHREAD_SETNAME_NP)
+#define PTHREAD_SETNAME_NP(thread, name) pthread_setname_np(thread, name)
```
tchaikov: much better. imo, can we avoid having the all upper case function name, i.e. reuse the name `pthread_setname_np()`, and define it with its equivalent if it's not available? i know, people might want to argue that macro should be named as `FUNCTION()` instead of `function()`.

src/include/compat.h
```diff
@@ -110,4 +112,25 @@
...
+#elif defined(HAVE_PTHREAD_SET_NAME_NP)
+#define PTHREAD_SETNAME_NP(thread, name) pthread_set_name_np(thread, name)
+#else
+/* compiler warning free success noop */
+#define PTHREAD_SETNAME_NP(thread, name) ({ \
+  int __i=0;                                  \
```
tchaikov: add spaces around `=`.

src/include/compat.h
wjwithagen: On 24-6-2016 12:57, Kefu Chai wrote:
> In src/include/compat.h
> <https://github.com/ceph/ceph/pull/9763#discussion_r68383203>:
> 
>> @@ -110,4 +112,25 @@
>>  #define MSG_DONTWAIT MSG_NONBLOCK
>>  #endif
>>  
>> +#if defined(HAVE_PTHREAD_SETNAME_NP)
>> +#define PTHREAD_SETNAME_NP(thread, name) pthread_setname_np(thread, name)
> 
> much better. imo, can we avoid having the all upper case function name,
> i.e. reuse the name |pthread_setname_np()|, and define it with its
> equivalent if it's not available? i know, people might want to argue
> that macro should be named as |FUNCTION()| instead of |function()|.

I'm sort of OLD school here, and I'd prefer MACROs and DEFINES to be
prefably in CAPS. Especially to make a distinction with code wrapped in
classed.

Other solution that was used in the time classes is the prefix the
repair code with ceph_. So there is a signal that something is up.

The advantage of reusing the name that we do not have to educate other
devs to think/look seriously at their code.

I'm all for education. :)

dx9:
Updated. Thanks for the help everyone. Let me know if anything else needs to change.

src/include/compat.h
```diff
@@ -112,4 +112,33 @@
...
 
+#if defined(HAVE_PTHREAD_SETNAME_NP)
+  #if defined(__APPLE__)
+    #define pthread_setname_np(thread, name) ({ \
+      int __result = 0;                         \
+      if (thread == pthread_self())             \
+        __result = pthread_setname_np(name)     \
+      __result; })
+  #else
+    #define pthread_setname_np(thread, name) pthread_setname_np(thread, name)
```
tchaikov: we don't need this `#define` in this case.

tchaikov:
lgtm modulo the nit.

src/include/compat.h
wjwithagen: On 1-7-2016 17:18, Kefu Chai wrote:
> In src/include/compat.h
> <https://github.com/ceph/ceph/pull/9763#discussion_r69315093>:
> 
>> @@ -112,4 +112,33 @@
>>  #define MSG_DONTWAIT MSG_NONBLOCK
>>  #endif
>>  
>> +#if defined(HAVE_PTHREAD_SETNAME_NP)
>> +  #if defined(__APPLE__)
>> +    #define pthread_setname_np(thread, name) ({ \
>> +      int __result = 0;                         \
>> +      if (thread == pthread_self())             \
>> +        __result = pthread_setname_np(name)     \
>> +      __result; })
>> +  #else
>> +    #define pthread_setname_np(thread, name) pthread_setname_np(thread, name)
> 
> we don't need this |#define| in this case.

Not only is it not needed, but I'd expect the preprocessor to start looping.
But I this that the first name needs a _ between set_name.

dx9:
@tchaikov @wjwithagen true. removed cruft. thanks

tchaikov:

dillaman:
Cannot compile from a clean Fedora 22 host on a fresh master branch:

> In file included from /home/jdillaman/ceph/src/common/Thread.h:19:0,
>                  from /home/jdillaman/ceph/src/common/Thread.cc:16:
> /usr/include/pthread.h:448:12: error: expected unqualified-id before ‘{’ token
>  extern int pthread_getname_np (pthread_t __target_thread, char *__buf,
>             ^
> /usr/include/pthread.h:448:12: error: expected ‘)’ before ‘{’ token
> In file included from /home/jdillaman/ceph/src/common/Thread.h:19:0,
>                  from /home/jdillaman/ceph/src/common/Thread.cc:16:
> /usr/include/pthread.h:448:23: error: expected unqualified-id before ‘)’ token
>  extern int pthread_getname_np (pthread_t __target_thread, char *__buf,
>                        ^
> In file included from /home/jdillaman/ceph/src/common/Thread.h:19:0,
>                  from /home/jdillaman/ceph/src/common/Thread.cc:16:
> /usr/include/pthread.h:453:12: error: expected unqualified-id before ‘{’ token
>  extern int pthread_setname_np (pthread_t __target_thread, const char *__name)
>             ^
> /usr/include/pthread.h:453:12: error: expected ‘)’ before ‘{’ token
> /usr/include/pthread.h:453:33: error: expected unqualified-id before ‘)’ token
>  extern int pthread_setname_np (pthread_t __target_thread, const char *__name)
>                                  ^
> src/CMakeFiles/common.dir/build.make:2159: recipe for target 'src/CMakeFiles/common.dir/common/Thread.cc.o' failed
>

dx9:
@dillaman hey man, I can't reproduce this. Ran a clean build on a docker fedora:22 image.

tchaikov:
@dillaman, i ran into the same issue but only to find that i have a leftover `src/acconfig.h`, could you check if you have another copy of `acconfig.h` in your source tree?

dillaman:
@tchaikov Yup -- that was it. All better now.

PR #9764, opened 2016-06-16
mention change of owner for /var/log/ceph

vasukulkarni:
mention change of owner for /var/log/ceph

Signed-off-by: Vasu Kulkarni <vasu@redhat.com>


vasukulkarni:
@athanatos please check

athanatos:
I think this is right.

vasukulkarni:
@athanatos thanks,  The CI failure has nothing to do with this change, I think I can merge this.

PR #6358, opened 2015-10-22
cmake: Use uname instead of arch.

dx9:

PR #9766, opened 2016-06-17
configure.ac: Use uname instead of arch.

dx9:
cmake is using uname -m see https://github.com/ceph/ceph/pull/6358

uname -m returns amd64 on FreeBSD


wjwithagen:
lgtm

PR #6301, opened 2015-10-18
libcephfs: Improve portability by replacing loff_t type usage with off_t

dx9:
?

PR #9767, opened 2016-06-17
safe_io: Improve portability by replacing loff_t type usage with off_t.

dx9:


src/common/safe_io.c
```diff
@@ -119,7 +119,7 @@ ssize_t safe_pwrite(int fd, const void *buf, size_t count, off_t offset)
 }
 
 #ifdef CEPH_HAVE_SPLICE
-ssize_t safe_splice(int fd_in, loff_t *off_in, int fd_out, loff_t *off_out,
+ssize_t safe_splice(int fd_in, off_t *off_in, int fd_out, off_t *off_out,
```
tchaikov: `loff_t` is a linux specific type. and is used by underlying splice(2) system call. yeah, it's a 64bit type. but i don't get what you mean by

> 64-bit behavior on glibc is enforced with __USE_FILE_OFFSET64 compiler error.

dx9: Ya sorry the comment is missing context here. 

off_t == loff_t on glibc when _USE_FILE_OFFSET64 is defined. See https://github.com/ceph/ceph/pull/6301#issuecomment-149608299

Musl libc provides linux splice() [1] with POSIX off_t [2]. 

There is no enforcement of _USE_FILE_OFFSET64 at this level. Maybe we should add a compiler error here similar to https://github.com/dx9/ceph/blob/1ff387047167686331825b592d4d396fc69da4fc/src/include/cephfs/libcephfs.h#L41

We want to make safe_splice() POSIX type compliant and ensure 64-bit usage on glibc.

?

[1] - https://git.musl-libc.org/cgit/musl/tree/include/fcntl.h#n165
[2] - http://pubs.opengroup.org/onlinepubs/009696699/basedefs/sys/types.h.html

dx9:
#10165 may be the better way forward, not sure yet

dx9:
@tchaikov ping

tchaikov:
@dx9 sorry for the latency, makes sense. merging.

dx9:
@tchaikov no worries, thanks.

PR #9769, opened 2016-06-17
mds: fix Session::check_access()

ukernel:
It calls CInode::make_path_string(...) with the second argument is
false. The second argument makes the third argument useless. For
newly created inode, the path string is something like #1xxxxxxxxx.
This can cause the access check to fail.

Fixes: http://tracker.ceph.com/issues/16358

### DIFF
--- a/src/mds/CInode.cc
+++ b/src/mds/CInode.cc
@@ -807,9 +807,14 @@ bool CInode::is_projected_ancestor_of(CInode *other)
   return false;
 }
 
-void CInode::make_path_string(string& s, bool force, CDentry *use_parent) const
+/*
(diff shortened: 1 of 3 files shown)


gregsfortytwo:
This looks good to me (SessionMap is the only one using these extra params), but please add a doc comment to the make_path_string() declaration...its params have already been misused!

jcsp:
Will include on test branch

gregsfortytwo:
http://pulpito.ceph.com/gregf-2016-06-27_22:02:48-fs-greg-fs-testing-627---basic-mira/ looks good, 3 lab infra timeouts

PR #9770, opened 2016-06-17
kv/RocksDB: add perfcounter for submit_transaction_sync operation

tanghaodong25:
We need collect the sum of submit_transaction and submit_transaction_sync respectively.

### DIFF
--- a/src/kv/RocksDBStore.cc
+++ b/src/kv/RocksDBStore.cc
@@ -296,15 +296,16 @@ int RocksDBStore::do_open(ostream &out, bool create_if_missing)
   }
 
   PerfCountersBuilder plb(g_ceph_context, "rocksdb", l_rocksdb_first, l_rocksdb_last);
-  plb.add_u64_counter(l_rocksdb_gets, "rocksdb_get", "Gets");
-  plb.add_u64_counter(l_rocksdb_txns, "rocksdb_transaction", "Transactions");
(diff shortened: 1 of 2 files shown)


src/kv/RocksDBStore.cc
```diff
@@ -297,7 +297,8 @@ int RocksDBStore::do_open(ostream &out, bool create_if_missing)
 
   PerfCountersBuilder plb(g_ceph_context, "rocksdb", l_rocksdb_first, l_rocksdb_last);
   plb.add_u64_counter(l_rocksdb_gets, "rocksdb_get", "Gets");
-  plb.add_u64_counter(l_rocksdb_txns, "rocksdb_transaction", "Transactions");
+  plb.add_u64_counter(l_rocksdb_txns, "rocksdb_submit_transaction", "Submit transactions");
+  plb.add_u64_counter(l_rocksdb_txns_sync, "rocksdb_submit_transaction_sync", "Submit transactions sync");
```
liewegas: while you are here, do you mind removing the "rocksdb_" prefix from all of these?  they are already inside the "rocksdb" perfcounter section so it is redundant.

liewegas:
otherwise looks good to me!

tanghaodong25:
@liewegas  updated. Thanks.

yuriw:
http://pulpito.ceph.com/yuriw-2016-07-12_13:07:42-rados-wip-yuri-testing_2016_7_12-distro-basic-smithi

PR #9772, opened 2016-06-17
msgr: set close on exec flag

tchaikov:
mds execv() when handling the "respawn" command, to avoid fd leakage,
and enormous CLOSE_WAIT connections after respawning, we need to set
FD_CLOEXEC flag for the socket fds.

### DIFF
--- a/src/msg/simple/Accepter.cc
+++ b/src/msg/simple/Accepter.cc
@@ -37,6 +37,18 @@
  * Accepter
  */
 
+static int set_close_on_exec(int fd)
+{
+  int flags = fcntl(fd, F_GETFD, 0);
+  if (flags < 0) {
+    return errno;
+  }
+  if (fcntl(fd, F_SETFD, flags | FD_CLOEXEC)) {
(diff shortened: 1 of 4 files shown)


yuyuyu101:
lgtm

src/msg/simple/Accepter.cc
```diff
@@ -63,6 +63,11 @@ int Accepter::bind(const entity_addr_t &bind_addr, const set<int>& avoid_ports)
     return -errno;
   }
 
+  if (set_close_on_exec(listen_sd)) {
```
tchaikov: please move this change to another commit.

tchaikov: sure.

tchaikov:
@ukernel do you think if mds "respawn" is an unusual case or not? if yes, i will file a tracker ticket and backport this changeset to jewel after it gets merged.

tchaikov:
rebased against master

tchaikov:
tested at 

PR #9773, opened 2016-06-17
mds/CInode: fix potential fin hanging

xiexingguo:
The behaviour of handle_write_error() process is configurable
and therefore is  non-deterministic. Thus we'd better to finish
the context here anyway.


gregsfortytwo:
How about invoking the finisher after we do the handle_write_error() call? That way if it does a respawn or (in the future) something more clever, we won't be trampling on its toes.

xiexingguo:
Done. Thanks, greg.

jcsp:
@gregsfortytwo could you review the updated code?

gregsfortytwo:
Yeah, this looks fine.

xiexingguo:
Tested at:

gregsfortytwo:

PR #9776, opened 2016-06-17
rgw:added zone rename

smanjara:
Added zone rename option to radosgw-admin help


smanjara:
Abandoning the change

PR #9778, opened 2016-06-17
os/bluestore: reduce bluestore blob

ifed01:


src/os/bluestore/BlueStore.cc
```diff
@@ -5588,7 +5587,7 @@ void BlueStore::_do_write_small(
     }
     int64_t blob = ep->second.blob;
     b = c->get_blob(o, ep->second.blob);
-    if (!b->blob.is_mutable()) {
+    if (!b->blob.is_mutable() || b->blob.is_compressed()) {
```
liewegas: this should be redundant.. anything that is compressed will necessarily not have mutable flag set.

liewegas:
great!

src/os/bluestore/BlueStore.cc
ifed01: Just a paranoia check :)

PR #9779, opened 2016-06-17
rgw: added zone rename to radosgw_admin

smanjara:
Added zone rename option to radosgw-admin help


cbodley:
looks good. will merge once the tests finish

cbodley:
unrelated make check failures, we only care about this part:
`PASS: src/test/run-cli-tests` :+1:

PR #9789, opened 2016-06-17
mds: add maximum fragment size constraint

batrick:
This commit adds a new config option

    mds_bal_fragment_size_max = 10000*10

which is an order of magnitude larger than mds_bal_split_size.

This limit prevents a fragment from getting too large which results in large
omap directories.

Right now the limit is enforced only in the RPC paths and in stray directory
entry creation. add_remote_dentry and add_local_dentry do not respect the
limit.

Fixes http://tracker.ceph.com/issues/16164

### DIFF
--- a/src/mds/Server.cc
+++ b/src/mds/Server.cc
@@ -2155,6 +2155,23 @@ bool Server::check_access(MDRequestRef& mdr, CInode *in, unsigned mask)
   return true;
 }
 
+/**
+ * check whether fragment has reached maximum size
+ *
+ */
+bool Server::check_fragment_space(MDRequestRef &mdr, CDir *in)
+{
+  const auto size = in->get_frag_size();
+  if (size >= g_conf->mds_bal_fragment_size_max) {
+    dout(10) << "fragment " << *in << " size exceeds " << g_conf->mds_bal_fragment_size_max << " (ENOSPC)" << dendl;
+    respond_to_request(mdr, -ENOSPC);
+    return false;
+  }
+
+  return true;
+}
+
+
 /** validate_dentry_dir
  *
  * verify that the dir exists and would own the dname.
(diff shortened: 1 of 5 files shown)


batrick:
I'm working on a qa-suite test for this.

src/mds/Server.cc
```diff
@@ -2237,20 +2252,39 @@ CDentry* Server::prepare_null_dentry(MDRequestRef& mdr, CDir *dir, const string&
...
+  } else {
+    assert(straydn->get_projected_linkage()->is_null());
+  }
+
+  // Notify even if a null dentry already existed, because
+  // StrayManager counts the number of stray inodes, not the
+  // number of dentries in the directory.
+  mdcache->notify_stray_created();
+
+  straydn->state_set(CDentry::STATE_STRAY);
```
gregsfortytwo: Please don't open-code the "create" bit of "get_or_create_stray_dentry". It belongs more in MDCache, and if you just put the get_stray_dir(CInode*) and create_stray_dentry(CDir* stray, const string& name) next to each other the diff will be smaller as well. ;)

gregsfortytwo:
Otherwise, looks good at first glance. :)

jcsp:
Is the i16164 test branch still a work in a progress or should I merge it into next test runs?

batrick:
I'm still trying to get a successful teuthology run to see if some test failures I saw in earlier runs are resolved. With that said, I think it's safe to merge into a test branch.

batrick:
There was a [valgrind error](http://qa-proxy.ceph.com/teuthology/pdonnell-2016-06-27_07:08:26-fs-wip-pdonnell-i16164---basic-smithi/280484/remote/smithi019/log/valgrind/client.0.log.gz) in a [run](http://pulpito.ceph.com/pdonnell-2016-06-27_07:08:26-fs-wip-pdonnell-i16164---basic-smithi/280484/) with the current ceph-qa-suite master.

The [error is also present](http://qa-proxy.ceph.com/teuthology/pdonnell-2016-06-27_06:54:23-fs-wip-pdonnell-i16164---basic-smithi/280396/remote/smithi012/log/valgrind/client.0.log.gz) in a [teuthology run](http://pulpito.ceph.com/pdonnell-2016-06-27_06:54:23-fs-wip-pdonnell-i16164---basic-smithi/280396/) with my [branch](https://github.com/ceph/ceph-qa-suite/tree/wip-pdonnell-i16164) containing my test.

That [same error](http://qa-proxy.ceph.com/teuthology/teuthology-2016-06-22_02:10:01-fs-jewel---basic-smithi/270344/remote/smithi002/log/valgrind/client.0.log.gz) is also present in [this jewel nightly](http://pulpito.ceph.com/teuthology-2016-06-22_02:10:01-fs-jewel---basic-smithi/270344/).

So that valgrind error appears unrelated to the branch.

I'm still looking into this test run with a [segmentation fault](http://pulpito.ceph.com/pdonnell-2016-06-27_06:54:23-fs-wip-pdonnell-i16164---basic-smithi/280361/).

PR #9795, opened 2016-06-18
common: Improve linux dcache hash algorithm

cyb70289:
Test code for profiling: [hash-profile2.tar.gz](https://github.com/ceph/ceph/files/325058/hash-profile2.tar.gz)

In ceph_str_hash_linux(), hash value is defined as unsigned **long**,
which is 8 bytes on 64-bit platforms when compiled with gcc. But
the return value is truncated to 4 bytes, and there's no need to
preserve 8 bytes intermediate value in the algorithm. The compiler
doesn't figure out this fact and produces redundant code.

```c
unsigned ceph_str_hash_linux(const char *str, unsigned length)
{
        /* "long" introduces performance penalty, should be deleted. */
        unsigned long hash = 0;

        while (length--) {
                unsigned char c = *str++; 
                hash = (hash + (c << 4) + (c >> 4)) * 11;
        }
        return hash;
}
```

After removing the "long" definition, this route runs much faster.
Following result are tested on x86_64 and aarch64 platforms, built
by gcc 5.3.1 with optimization level -O2. Same output is observed
with gcc 4.8.4 and -O3 optimization.

*ARM Cortex-A57*

String Length | Time w/ long | Time w/o long | Improvement 
------------- | ------------ | ------------- | -----------
 32            | 0.088 us     | 0.067 us      | 24%         
 4096          | 10.26 us     | 8.20 us       | 20%         
 65536         | 164 us       | 131 us        | 20%         
 1048576       | 2624 us      | 2099 us       | 20%         

*Intel i7-4790*

String Length | Time w/ long | Time w/o long | Improvement
--------------- | -------------- | --------------- | -------------
 32            | 0.033 us     | 0.028 us      | 16.3%       
 4096          | 3.87 us      | 3.64 us       | 6.2%        
 65536         | 61.3 us      | 57.8 us       | 5.7%        
 1048576       | 973 us       | 917 us        | 5.8%        

Signed-off-by: Yibo Cai <yibo.cai@linaro.org>

### DIFF
--- a/src/common/ceph_hash.cc
+++ b/src/common/ceph_hash.cc
@@ -82,7 +82,7 @@ unsigned ceph_str_hash_rjenkins(const char *str, unsigned length)
  */
 unsigned ceph_str_hash_linux(const char *str, unsigned length)
 {
-	unsigned long hash = 0;
+	unsigned hash = 0;
 
 	while (length--) {
 		unsigned char c = *str++;


src/common/ceph_hash.cc
```diff
@@ -82,7 +82,7 @@ unsigned ceph_str_hash_rjenkins(const char *str, unsigned length)
  */
 unsigned ceph_str_hash_linux(const char *str, unsigned length)
 {
-	unsigned long hash = 0;
+	unsigned hash = 0;
```
tchaikov: could use `uint_fast32_t`? and change the return type to `uint32_t`?

cyb70289: thanks tchaikov.

i tried "uint_fast32_t", the performance is same as old code, and i found it's defined as "long int" on 64-bit platform(/usr/include/stdint.h). from compiler's point of view, it makes sense as integer with register width in length can be handled efficiently. but it just isn't smart enough to tune all cases.

about "change return type to uint32_t", i think it's necessary as the old code uses "non-type unsigned" extensively which is not clear enough. because there are many codes in this file using "non-type unsigned" and they're related to each other, maybe it's better to push another commit to fix this issue specifically?

tchaikov:
the test code can be found at https://github.com/ceph/ceph/files/325058/hash-profile2.tar.gz 

although the performance change is not that significant in my test bed, it is still worthy to have this change, imo.

```
time-new = 3608 ms, hash = bcfbefaa
time-old = 3677 ms, hash = bcfbefaa
Improves [1.88%] when block_size = 32

time-new = 2199 ms, hash = 59cf0da8
time-old = 2284 ms, hash = 59cf0da8
Improves [3.72%] when block_size = 4096

time-new = 2112 ms, hash = fe750281
time-old = 2172 ms, hash = fe750281
Improves [2.76%] when block_size = 65536

time-new = 3380 ms, hash = 851e19e0
time-old = 3640 ms, hash = 851e19e0
Improves [7.14%] when block_size = 1048576


Intel(R) Xeon(R) CPU E3-1230 V2 @ 3.30GHz
```

tchaikov:
lgtm.

athanatos:
lgtm

yuriw:
test run http://pulpito.ceph.com/yuriw-2016-06-25_17:00:35-rados-wip-yuri-testing2_2016_6_23-distro-basic-smithi/

smithfarm:
@tchaikov @athanatos This is a pretty significant performance boost for ARM. Would it make sense to backport it to jewel?

dachary:
It's really simple, I don't see why not ?

smithfarm:
Opened tracker issue http://tracker.ceph.com/issues/17599

PR #9798, opened 2016-06-18
client: make renew of session cap more fine-grained

xiexingguo:
Sessions may get created or destroyed dynamically.
Therefore we shall renew its cap based on its own ttl
instead of renewing all sessions' cap in a batch,
which can reduce the message traffic and thus is good
for performance.

### DIFF
--- a/src/client/Client.cc
+++ b/src/client/Client.cc
@@ -5794,11 +5794,7 @@ void Client::tick()
   }
 
   if (mdsmap->get_epoch()) {
-    // renew caps?
-    utime_t el = now - last_cap_renew;
-    if (el > mdsmap->get_session_timeout() / 3.0)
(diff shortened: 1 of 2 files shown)


gregsfortytwo:
We also call renew_caps() if we get a CEPH_SESSION_STALE from the MDS. Granted, that shouldn't happen if we're sending off cap messages like we should be, but let's not skip it in case there's some kind of mis-configuration.

Also, I don't think we get any benefit out trying to stagger the messages. We still have to send them out at the same rate (renew interval/3); the existing mechanism promotes batching but works out to be the same over time. Feel free to re-open if I'm missing something.

xiexingguo:
@gregsfortytwo  Thanks for the confirmation. I am OK to drop this change.

PR #9800, opened 2016-06-18
ceph: Fix rados bench seq/rand tests to work when -b op-size != -o ob…

krehm:
…ject-size

rados bench write correctly creates objects whose object_size is a multiple of
op_size, but radios bench seq and rand cannot read them, failing with:
    benchmark_data_alpha1-p200_7856_object0 is not correct!

There are a number of issues concerning this bug, discovered one at a time.
Routine aio_read() in rados.cc is hard-coded to read an object at offset zero,
no matter what offset it is passed.

Routine seq_read_bench() in obj_bencher.cc stops reading after "num_objects"
I/Os, when it should issue "num_objects \* writes_per_object" reads (but see
below), so the test run times are unexpectedly short.

In routine rand_read_bench(), the first "concurrentios" I/Os are actually
sequential reads of the first few objects, not random. Also, the routine
only generates random offsets modulo "num_objects", not
"num_objects \* writes_per_object", so the I/O distribution ignores most of
the higher-numbered objects (but see next).

If a rados bench write test finishes based on time, not number of objects,
then the last object written is usually shorter than the rest. Fixing the
above issues causes the seq and rand tests to issue I/Os beyond the end of
that final shorter object.

The solution is to record the number of writes issued by rados bench write
as an additional parameter in the "bench_metadata" file (num_writes). The
seq test then issues "num_writes" reads and doesn't run off the end of the
last object. The rand test generates a random number modulo "num_writes",
then converts that to an object and offset pair, also avoiding any reads
beyond the end of the last object, and the I/O distribution across the
entire set of objects and across offsets within objects is balanced.

The code in this fix is unencumbered by any other license.
Fixes: #16385
Signed-off-by: krehm

### DIFF
--- a/doc/dev/perf.rst
+++ b/doc/dev/perf.rst
@@ -1,0 +1,39 @@
+Using perf
+==========
+
+Top::
+
+  sudo perf top -p `pidof ceph-osd`
+
+To capture some data with call graphs::
+
+  sudo perf record -p `pidof ceph-osd` -F 99 --call-graph dwarf -- sleep 60
+
+To view by caller (where you can see what each top function calls)::
+
+  sudo perf report --call-graph caller
+
+To view by callee (where you can see who calls each each top function)::  
+
+  sudo perf report --call-graph callee
+
+:note: If the the caller/callee views look the same you may be
+       suffering from a kernel bug; upgrade to 4.8 or later.
+
+Flamegraphs
+-----------
+
+First, get things set up::
+
+  cd ~/src
+  git clone https://github.com/brendangregg/FlameGraph
+
+Run ceph, then record some perf data::
+
+  sudo perf record -p `pidof ceph-osd` -F 99 --call-graph dwarf -- sleep 60
+
+Then generate the flamegraph::
+
+  sudo perf script | ~/src/FlameGraph/stackcollapse-perf.pl > /tmp/folded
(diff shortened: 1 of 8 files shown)


dachary:
this needs rebasing

krehm:
I had no success trying to do a rebase.   What I did on my wip-krehm-16385 branch was:
   git fetch upstream
   git merge upstream/master
   git push origin wip-krehm-16385
to catch up with all the latest commits.  I also corrected a whitespace error.   Then I did:
    git rebase -i 71bf392aaea6f464d01c9b6d4d580c4c3d68517a
I moved my two commits to the end of the list and made the second a fixup:
    pick cf9f5fc ceph: Fix rados bench seq/rand tests to work when -b op-size != -o object-size
    fixup 074ff21 Fix several spacing errors.

What happens is that I start getting merge errors, first about empty cherry-picks, for which I responded with:
    git commit --allow-empty
Then I start getting merge errors on files in which I have no changes:
    src/msg/simple/Pipe.cc
    src/rgw/rgw_file.h
    src/client/Client.h
With multiple merge complaints about this last file.  At that point I gave up, I don't trust that the rebase is working correctly.

Any advice?

Thanks, Kevan

krehm:
I deleted the existing wip-krehm-16385 branch in my fork, then recreated it from master and cherry-picked my mod into it.  I have built ceph successfully with the new branch, and have opened a fresh pull request (#12203) using the new branch.

PR #9801, opened 2016-06-19
msg/msg_types.h: make members private in ...

stiopaa1:
... entity_name_t class.


tchaikov:
@stiopaa1 seems it is not enough to just change the caller of `MDCache::handle_discover()`. there are other places which we access these members, for instance, https://github.com/ceph/ceph/blob/master/src/osd/OpRequest.cc#L147.

PR #9804, opened 2016-06-20
Revert "msg/msg_types.h: make members private in ..."

yuyuyu101:
Reverts ceph/ceph#9801

failed to compiile


tchaikov:
sorry, my bad. didn't notice that we do access the private members in the trace points.

PR #9805, opened 2016-06-20
mds/MDCache: fix handle_dir_update()

renhwztetecs:
put message before return


ukernel:
This is wrong because C_MDS_RetryMessage uses caller's reference

renhwztetecs:
@ukernel 
yes, thanks
i close it

PR #9806, opened 2016-06-20
[DNM]client: fix race for ll_setxattr()

xiexingguo:
After successfully getting the latest osdmap, we shall try again
to check whether the pool does exist or not.

Also we shall do fast abort if the pool existence checking
turns out to be unsuccessful.

### DIFF
--- a/src/client/Client.cc
+++ b/src/client/Client.cc
@@ -10119,7 +10119,15 @@ int Client::ll_setxattr(Inode *in, const char *name, const void *value,
       C_SaferCond ctx;
       objecter->wait_for_latest_osdmap(&ctx);
       ctx.wait();
+
+      // Got the latest osdmap, try again.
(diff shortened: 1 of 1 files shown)


gregsfortytwo:
LGTM

xiexingguo:
http://pulpito.ceph.com/jspray-2016-06-22_02:53:43-fs-wip-jcsp-testing-20160621---basic-mira/270927/
http://pulpito.ceph.com/jspray-2016-06-22_02:53:43-fs-wip-jcsp-testing-20160621---basic-mira/270969/

The above test failures are suspiciously caused by this patch.
Verifying locally.

xiexingguo:
Well, i don't know the exact reason why this change shall fail the above testings. But it is still reproducible in my local test bed.

Anyhow,  close this for now, since this change is not of much importance, as at the end the server shall catch these errors two.

PR #9811, opened 2016-06-20
rgw: fix BUG#16355

linuxhunter:
fix the BUG#16355 with tenant information on rgw.

### DIFF
--- a/src/cls/user/cls_user.cc
+++ b/src/cls/user/cls_user.cc
@@ -54,27 +54,24 @@ static void get_key_by_bucket_name(const string& bucket_tenant, const string& bu
     }
 }
 
-static int get_existing_bucket_entry(cls_method_context_t hctx, const string& bucket_tenant, const string& bucket_name,
+static int get_existing_bucket_entry(cls_method_context_t hctx, const string& bucket_key,
                                      cls_user_bucket_entry& entry)
 {
-  if (bucket_name.empty()) {
+  if (bucket_key.empty()) {
     return -EINVAL;
   }
 
-  string key;
-
-  get_key_by_bucket_name(bucket_tenant, bucket_name, &key);
   bufferlist bl;
-  int rc = cls_cxx_map_get_val(hctx, key, &bl);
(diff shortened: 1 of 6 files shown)


mattbenjamin:
@zaitcev could you scan this?

theanalyst:
I think pin fragmenting should be a separate pr

zaitcev:
I see a couple of issues with the new code.

- It keeps the double-conversion that was mistakenly done in the original code. I think a cleaner approach would be to pass the key to get_existing_bucket_entry() as it's done currently, but rename that argument from bucket_name to bucket_key or such. Then remove the call to get_key_by_bucket_name from the body of get_existing_bucket_entry.

- It adds additional field to serialized structures without any regard for compatibility. I have to wonder if that structure was used by RGW itself. I don't think I could be quite this careless. Actually I left Yehuda's additions intact, but I remember that he was careful with tacking the new fields at the tail and added ifs for structure versions.

linuxhunter:
Hi zaitcev, 
    I had fix the issue 1 by commit 7637d7d2abf6b4edf0fe03f9004bc633e54a0f64, please check it again.

Thanks.

zaitcev:
The double-conversion part now looks good, but what about the incompatible serialization?

linuxhunter:
Hi zaitcev,
    I had try to fix the incompatible serialization on struct cls_user_bucket by commit d45259f, please check it again. 

Thanks.

zaitcev:
My objections were addressed. We still need @yehudasa to take a look, but seems okay. One remaining question - is this the last place that needs fixes? I did not realize the code outside of src/rgw/ would be affected.

Acked-by: Pete Zaitcev <zaitcev@redhat.com>

linuxhunter:
Hi zaitcev, 
    This is the last place that needs to fix for bug#16355.

linuxhunter:
Hi zaitcev,
    Does @yehudasa take a look? What's suggestions?

linuxhunter:
create a new pull request #9938 to replace this which delete the codes unrelated to the bug.

PR #9813, opened 2016-06-20
doc/cephfs: add advice about kernel versions

jcsp:


jcsp:
@ukernel what do you think about this?  Should we just say "4.x", or can we be more specific about which version had all the known fixes at the time Jewel was released?

The idea is to put something here that we can send people a link to when they have kernel version issues.

ukernel:
the advice says 'least 4.x kernel', that should be enough

PR #9816, opened 2016-06-20
librbd: potential race when replaying journal ops

dillaman:
Fixes: http://tracker.ceph.com/issues/16198

### DIFF
--- a/src/librbd/journal/Replay.cc
+++ b/src/librbd/journal/Replay.cc
@@ -147,6 +147,7 @@ Replay<I>::~Replay() {
   assert(m_aio_modify_unsafe_contexts.empty());
   assert(m_aio_modify_safe_contexts.empty());
   assert(m_op_events.empty());
+  assert(m_in_flight_op_events == 0);
 }
 
 template <typename I>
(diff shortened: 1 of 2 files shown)


src/librbd/journal/Replay.cc
```diff
@@ -802,8 +799,15 @@ void Replay<I>::handle_op_complete(uint64_t op_tid, int r) {
...
-    on_flush->complete(0);
+
+  // shut down request might have occurred while lock was
+  // dropped -- handle if pending
+  Mutex::Locker locker(m_lock);
+  assert(m_in_flight_op_events > 0);
+  --m_in_flight_op_events;
+  if (m_flush_ctx != nullptr && m_in_flight_op_events == 0 &&
+      (m_in_flight_aio_flush + m_in_flight_aio_modify) == 0) {
+    m_image_ctx.op_work_queue->queue(m_flush_ctx, 0);
```
trociny: @dillaman m_flush_ctx can be completed from both handle_aio_flush_complete and handle_op_complete. I wonder if a race is possible here and it is completed twice?

dillaman: New IO and op events shouldn't be processed after ```shut_down``` is invoked. With the IO and op counters, it should only be able to be invoked one time -- depending on whether an IO or op finished last.

trociny: Ah, I see, makes sense.

trociny:
lgtm

PR #9817, opened 2016-06-20
rgw_swift: newer versions of boost/utility no longer include in_place

theanalyst:
boost > 1.58 no longer includes in_place in boost/utility, we need to
include in_place_factory explicitly. This causes build failures in
distros that ship with a higher version of boost. Since the only call is for
swift_ver_location, another possibility is to use emplace()
instead (though this requires boost ~ 1.56)

Fixes: http://tracker.ceph.com/issues/16391


rzarzynski:
Thanks for pointing the problem out! I'm putting this PR as a backporting dependency for #9188.

CC: @oritwas.

PR #9820, opened 2016-06-20
configure.ac: fix res_query detection

dx9:


configure.ac
```diff
@@ -354,38 +354,45 @@ if test x"$linux" = x"yes"; then
...
+  LIBS="$2"
+  AS_CASE($1,
+          res_nquery, [resolve_func="res_nquery(0, 0, 0, 0, 0, 0)"],
+          res_query, [resolve_func="res_query(0, 0, 0, 0, 0)"],
+          AC_MSG_FAILURE([unknown resolv func]))
   AC_LINK_IFELSE([AC_LANG_PROGRAM([[
         #include <netinet/in.h>
         #include <resolv.h>
-      ]], [[res_nquery(0, 0, 0, 0, 0, 0);]])],
+      ]], [[$resolve_func;]])],
```
tchaikov: nit, could extract `(0, 0, 0, 0, 0, 0)` out and put it here.

tchaikov: nit, could extract `(0, 0, 0, 0, 0, 0)` out and put it here.

dx9:
@tchaikov Ya much cleaner that way. Also added a comment and fixed the check order so res_nquery always takes precedence over res_query. Thanks

dx9:
@tchaikov oops I broke it... Hang on

dx9:
@tchaikov I think it's good now. LMK

configure.ac
```diff
@@ -354,38 +354,42 @@ if test x"$linux" = x"yes"; then
...
-  AC_MSG_CHECKING([if res_nquery will link (LIBS=$1)])
+  AC_MSG_CHECKING([if $1 will link (LIBS=$3)])
   saved_LIBS="${LIBS}"
-  LIBS="$1"
+  LIBS="$3"
   AC_LINK_IFELSE([AC_LANG_PROGRAM([[
         #include <netinet/in.h>
         #include <resolv.h>
-      ]], [[res_nquery(0, 0, 0, 0, 0, 0);]])],
+      ]], [[$1$2;]])],
```
tchaikov: @dx9 does `"$1"(0, 0, 0, 0, 0, 0)` work here?

dx9: @tchaikov Don't think so? res_query and res_nquery have different number of arguments.

res_nquery(res_state statep, const char *dname, int class, int type, unsigned char *answer, int anslen)
res_query(const char *dname, int class, int type, unsigned char *answer, int anslen)

http://man7.org/linux/man-pages/man3/resolver.3.html

tchaikov: oh, sorry again. bad at counting! =(

tchaikov:
lgtm

tchaikov:
tested at 

PR #9821, opened 2016-06-20
configure.ac: add --with-reentrant-strsignal and --with-thread-safe-r…

dx9:


tchaikov:
nit, the commit message is not accurate:
- `s/--with_thread_safe_res_query/--with-thread-safe-res-query/`
- `s/--with_reentrant_strsignal/--with-reentrant-strsignal/`.

other than this, lgtm.

dx9:
@tchaikov oops, fixed, thanks

PR #9826, opened 2016-06-20
os/bluestore/BlueFS: instrument which file types we write to

liewegas:

### DIFF
--- a/src/os/bluestore/BlueFS.cc
+++ b/src/os/bluestore/BlueFS.cc
@@ -46,18 +46,36 @@ void BlueFS::_init_logger()
 {
   PerfCountersBuilder b(g_ceph_context, "BlueFS",
                         l_bluefs_first, l_bluefs_last);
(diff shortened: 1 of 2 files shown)


src/os/bluestore/BlueFS.cc
```diff
@@ -1399,6 +1426,19 @@ int BlueFS::open_for_write(
...
+    if (logger) {
+      logger->inc(l_bluefs_files_written_wal);
+    }
+  } else if (0 == filename.compare(filename.length() - 4, 4, ".sst")) {
+    (*h)->writer_type = BlueFS::WRITER_SST;
+    if (logger) {
+      logger->inc(l_bluefs_files_written_sst);
+    }
+  }
+
```
varadakari: Better to have a else case to handle any future extensions or unit tests to create any file.

varadakari:
Rest looks good.

src/os/bluestore/BlueFS.cc
liewegas: The default value is 0 (WRITER_UNKNOWN).

PR #9827, opened 2016-06-20
os/bluestore: discard unreferenced buffers as they are overwritten

liewegas:

### DIFF
--- a/src/os/bluestore/BlueStore.cc
+++ b/src/os/bluestore/BlueStore.cc
@@ -813,6 +813,22 @@ bool BlueStore::OnodeSpace::map_any(std::function<bool(OnodeRef)> f)
 }
 
 
+// Blob
+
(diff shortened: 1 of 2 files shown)


src/os/bluestore/BlueStore.cc
```diff
@@ -5692,6 +5713,15 @@ void BlueStore::_wctx_finish(
       if (compressed) {
         txc->statfs_delta.compressed_allocated() -= e.length;
       }
+      if (l.blob < 0) {
```
ifed01: Why doing that inside the loop over released pextents? IMHO this can be done at line 5705 too.

liewegas: oh, right. (i was originally going to use the r list for this.)

ifed01:
LGTM

PR #9830, opened 2016-06-20
cmake: CMake instructions missing -c flag

alimaredia:
The CMake instructions under "Running a test
cluster" were missing the ceph.conf vstart
generates for them.

Also restart and stop with init-ceph instructions
weren't functional.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,11 +147,11 @@ in the section above titled "Building Ceph".
 To run a functional test cluster,
 
 	cd build
-	./path/to/ceph/src/vstart.sh -d -n -x -l
+	../src/vstart.sh -d -n -x -l
 	./bin/ceph -c ceph.conf -s
 
-Note: vstart.sh generates a ceph.conf file in the build directory that should be used
-for subsequent commands.
+Note: vstart.sh generates a ceph.conf file in the build directory that should be 
+used for subsequent commands. 
 
 Almost all of the usual commands are available in the bin/ directory.
 For example,
@@ -161,7 +161,7 @@ For example,
 
 To shut down the test cluster,
 
-	./path/to/ceph/src/stop.sh
+	../src/stop.sh
 
 Running unit tests
 ==================
--- a/README.md
+++ b/README.md
@@ -148,22 +148,20 @@ To run a functional test cluster,
 
 	cd build
 	./path/to/ceph/src/vstart.sh -d -n -x -l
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -148,23 +148,21 @@ To run a functional test cluster,
...
+	./bin/rados -c ceph.conf -p rbd bench 30 write
+	./bin/rbd -c ceph.conf create foo --size 1000
 
 To shut down the test cluster,
 
 	./path/to/ceph/src/stop.sh
 
-To start or stop individual daemons, the sysvinit script should work:
-
-	./bin/init-ceph restart osd.0
```
alimaredia: @liewegas I tried giving this init-ceph a ceph.conf file but the osd still didn't come back up. The command is meant to be run after stop.sh right?

liewegas: usually i do restart, or stop (in leiu of stop.sh)

alimaredia: Neither of those commands worked when I had the everything up. That doesn't sound good to me... @oritwas @tchaikov @jcsp do any of you use init-ceph frequently with CMake + vstart.sh?

tchaikov: i don't use `init-ceph` along with CMake + vstart.sh.

README.md
```diff
@@ -148,23 +148,21 @@ To run a functional test cluster,
 
 	cd build
 	./path/to/ceph/src/vstart.sh -d -n -x -l
-	./bin/ceph -s
+	./bin/ceph -c ceph.conf -s
```
tchaikov: why shall we specify the path to ceph.conf?

README.md
```diff
@@ -148,23 +148,21 @@ To run a functional test cluster,
 
 	cd build
 	./path/to/ceph/src/vstart.sh -d -n -x -l
```
tchaikov: for the sake of consistency, could probably just use `../src/vstart.sh`? as we suggest
```
	cd build
	cmake [options] ..
	make check
```
before

liewegas: +1

README.md
alimaredia: I think it defaults to searching in /etc/ceph for the ceph.conf. When I ran just the ceph command (without specifying the ceph.conf in $PWD) it hung.

liewegas:
-c isn't needed for me.  and let's fix init-ceph.

PR #9831, opened 2016-06-20
cmake: rework NSS and SSL

mattbenjamin:
The cmake build should select NSS or CryptoPP consistently with
current automake logic, as used in our RHEL builds (-DWITH_NSS),
and the SSL linkage must match (for NSS, -lssl3).

With this change, there is no direct linkage with libssl when NSS
is selected (I still see indirect linkage, looking into that).


CMakeLists.txt
```diff
@@ -284,26 +274,45 @@ find_package(libuuid REQUIRED)
...
+  endif(NSS_FOUND)
+else ()
+  find_package(cryptopp)
+  if(CRYPTOPP_FOUND)
+    MESSAGE(STATUS "${CRYPTOPP_LIBRARIES}")
+    set(CRYPTO_LIBS ${CRYPTOPP_LIBRARIES})
+    set(USE_NSS 0)
+    set(USE_CRYPTOPP 1)
+  else()
+    MESSAGE(STATUS "Cryptopp not found using NSS instead")
```
cbodley: this message doesn't appear to be true any more. should the `find_package(cryptopp)` be REQUIRED so we fail in this case?

mattbenjamin: Yes.  This needs to fail if cryptopp is not found, now.

CMakeLists.txt
```diff
@@ -284,26 +274,44 @@ find_package(libuuid REQUIRED)
...
-			set(CRYPTO_LIBS ${NSS_LIBRARIES} ${NSPR_LIBRARIES})
-			#MESSAGE(STATUS "${CRYPTO_LIBS}")
-			#MESSAGE(STATUS "${NSS_INCLUDE_DIR} ${NSPR_INCLUDE_DIR}")
-		endif(NSPR_FOUND)
-	endif(NSS_FOUND)
-endif(CRYPTOPP_FOUND)
+option(WITH_NSS "Use NSS crypto and SSL implementations" OFF)
+if (${WITH_NSS})
+  find_package(NSS REQUIRED)
+  if(NSS_FOUND)
```
tchaikov: if nss is not found, cmake will bail out with the default error message, so no need to check `NSS_FOUND` here?

CMakeLists.txt
```diff
@@ -284,26 +274,44 @@ find_package(libuuid REQUIRED)
...
+    set(USE_NSS 1)
+    set(USE_CRYPTOPP 0)
+    find_package(NSPR)
+    if(NSPR_FOUND)
+      set(CRYPTO_LIBS ${NSS_LIBRARIES} ${NSPR_LIBRARIES})
+      #MESSAGE(STATUS "${CRYPTO_LIBS}")
```
tchaikov: maybe we can remove the commented lines also?

CMakeLists.txt
```diff
@@ -284,26 +274,44 @@ find_package(libuuid REQUIRED)
...
+      #MESSAGE(STATUS "${NSS_INCLUDE_DIR} ${NSPR_INCLUDE_DIR}")
+    endif(NSPR_FOUND)
+  endif(NSS_FOUND)
+else ()
+  find_package(cryptopp REQUIRED)
+  if(CRYPTOPP_FOUND)
+    MESSAGE(STATUS "${CRYPTOPP_LIBRARIES}")
+    set(CRYPTO_LIBS ${CRYPTOPP_LIBRARIES})
+    set(USE_NSS 0)
```
tchaikov: no need to `set(USE_NSS 0)`. it is empty by default. so `if (USE_NSS)` and `if (${USE_NSS})` will be False in this case.

src/CMakeLists.txt
```diff
@@ -1393,13 +1393,13 @@ if(${WITH_RADOSGW})
...
-  target_include_directories(civetweb_common_objs PUBLIC
-	"${OPENSSL_INCLUDE_DIR}")
-  else(HAVE_OPENSSL)
-  endif(HAVE_OPENSSL)
+  if(HAVE_SSL)
+    set_property(TARGET civetweb_common_objs
+      APPEND PROPERTY COMPILE_DEFINITIONS NO_SSL_DL=1)
+    target_include_directories(civetweb_common_objs PUBLIC
+      "${SSL_INCLUDE_DIR}")
+  else(HAVE_SSL)
```
tchaikov: could dispense with this empty "else" block?

CMakeLists.txt
```diff
@@ -284,26 +274,44 @@ find_package(libuuid REQUIRED)
...
-		endif(NSPR_FOUND)
-	endif(NSS_FOUND)
-endif(CRYPTOPP_FOUND)
+option(WITH_NSS "Use NSS crypto and SSL implementations" OFF)
+if (${WITH_NSS})
+  find_package(NSS REQUIRED)
+  if(NSS_FOUND)
+    set(USE_NSS 1)
+    set(USE_CRYPTOPP 0)
+    find_package(NSPR)
```
tchaikov: actually, i think this sure be "REQUIRED".

configure.ac
```diff
@@ -439,6 +439,8 @@ if test "x$have_nss" = "xno" -a "x$with_nss" != "xcheck" -a "x$with_nss" != "xno
     AC_MSG_FAILURE([--with-nss was given, but library was not found])
 fi
 
+AC_MSG_NOTICE([NSS is selected, now we really do something])
```
tchaikov: this message is always printed, do i guess we can remove it?

configure.ac
```diff
@@ -447,6 +449,14 @@ if test "x$have_cryptopp" = "xyes"; then
     #AC_SUBST([CRYPTO_CXXFLAGS], [$CRYPTOPP_CXXFLAGS])
     AM_CXXFLAGS="${AM_CXXFLAGS} ${CRYPTOPP_CXXFLAGS}"
     AC_SUBST([CRYPTO_LIBS], [$CRYPTOPP_LIBS])
+    # OpenSSL -- defines OPENSSL_{INCLUDES,LIBS,LDFLAGS} on success
+    AX_CHECK_OPENSSL([], [found_openssl=1], [found_openssl=0])
```
tchaikov: should be 
```m4
AX_CHECK_OPENSSL([], AC_MSG_FAILURE([CryptoPP build selected but OpenSSL not found])
AC_SUBST([SSL_INCLUDES], [$OPENSSL_INCLUDES])
AC_SUBST([SSL_LIBS], [$OPENSSL_LIBS])
```

tchaikov:
needs rebase.

CMakeLists.txt
mattbenjamin: which should be required?  NSPR?

CMakeLists.txt
mattbenjamin: someone wrote them originally, don't think it's critical

CMakeLists.txt
mattbenjamin: now that it is required, that's true

CMakeLists.txt
mattbenjamin: that's rather fussy

src/CMakeLists.txt
mattbenjamin: that's rather fussy

configure.ac
mattbenjamin: someone wrote it;  if you want to remove random cruft, you could send a patch to do a bunch at once?

configure.ac
mattbenjamin: this is going away in less than 2 weeks...

mattbenjamin:
I'll clean up and rebase tomorrow

liewegas:
Note wip-cmake-debian has 
https://github.com/ceph/ceph/pull/9909/commits/27c861a87f18047757c9962b9af95d24d0d39ed6

Can we cherry-pick this into master, or do you want a fancier version 
that maintains cryptopp?  Don't care either way, as long as it doesn't 
block merging deb.  (The current "check for cryptopp, then check for nss" 
fails on many gitbuilders.)

mattbenjamin:
I'd really rather not dispose of SSL and Crypto choice--I guess if I want that, I need to fix gitbuilders?  In the interest of that, how does that check cryptopp, check nss fail?

liewegas:
The problem is that the gitbuilder detected cryptopp installed but it 
didn't build.  Since right now nobody ever uses it, at a minimum we have 
to make nss the default choice.  The only reason I didn't do that in my 
patch is because it was confusing and I needed it fixed.

configure.ac
mattbenjamin: ok, while Im at it I'll remove all these

mattbenjamin:
ok, makes sense--if you can wait till midday EST tomorrow, I can retest and push it then?

liewegas:
btw just hit this again on my laptop.  libcrypto++ was install and the 
build fails (bad #include somewhere).  worked around it by uninstalling 
libcrypto++

configure.ac
tchaikov: no, it's not cosmetic. the first arg is executed when openssl is found, and the second arg is executed when openssl is not found. in your case, the second arg is `[found_openssl=1]`.

CMakeLists.txt
tchaikov: yes, NSPR. we are including nspr headers if NSS is used.

mattbenjamin:
Ok, still testing this, but, in addition to requested cleanups, NSS is now the default for cmake and automake builds.

mattbenjamin:
Re-pushed--previous set had unintended changes to some generated m4 files.

tchaikov:
@mattbenjamin i am very sorry that i merged unrelated changes into master by mistake, so your changeset conflicts with sage's commit i merged. i have a PR reverting them.

i just tested your changeset, both autotools and cmake builds work great.

lgtm

mattbenjamin:
@tchaikov no problem, would it be easier if I rebased?

PR #9833, opened 2016-06-20
Make Rados ready for PyPI

onyb:
Reopening the former pull request https://github.com/ceph/ceph/pull/9332 again, since it got messed up pretty badly trying to remove a file from the commit history.

This is also same as the `wip-rados-pypi` of `ceph.git`, but with reworded commit messages and `rados.c` nuked out from the history.

### DIFF
--- a/src/pybind/rados/setup.py
+++ b/src/pybind/rados/setup.py
@@ -77,7 +77,7 @@ def check_sanity():
     Test if development headers and library for rados is available by compiling a dummy C program.
     """
 
-    tmp_dir = tempfile.mkdtemp(dir=os.path.dirname(__file__))
+    tmp_dir = tempfile.mkdtemp(dir=os.environ.get('TMPDIR', os.path.dirname(__file__)))
     tmp_file = os.path.join(tmp_dir, 'rados_dummy.c')
 
     with open(tmp_file, 'w') as fp:
--- a/src/pybind/rados/setup.py
+++ b/src/pybind/rados/setup.py
@@ -122,6 +122,22 @@ if (len(sys.argv) >= 2 and
     def cythonize(x, **kwargs):
         return x
 
+if '--without-cython' in sys.argv:
+    if not os.path.isfile('rados.c'):
+        print('Error: Cannot find Cythonized file rados.c', file=sys.stderr)
+        print('Retry without using --without-cython', file=sys.stderr)
+        sys.exit(1)
+
+
+    def cythonize(x, **kwargs):
+        return x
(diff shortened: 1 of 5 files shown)


jdurgin:
this is looking good to me, @tchaikov did you want to look it over too?

src/pybind/rados/setup.py
```diff
@@ -61,6 +69,52 @@ def get_python_flags():
     }
 
 
+def check_sanity():
+    """
+    Test if development headers and library for rados is available by compiling a dummy C program.
+    """
+
+    tmp_dir = tempfile.mkdtemp(dir=os.path.dirname(__file__))
```
tchaikov: nit, would be better if we can respect `$TMPDIR`.

src/pybind/rados/setup.py
```diff
@@ -122,6 +122,22 @@ def check_sanity():
     def cythonize(x, **kwargs):
         return x
 
+if '--without-cython' in sys.argv:
```
tchaikov: under what condition, user would use `setup.py` without cython? and we don't have `rados.c` in our repo yet, right?

src/pybind/rados/setup.py
```diff
@@ -90,12 +88,20 @@ def check_sanity():
     compiler = new_compiler()
     customize_compiler(compiler)
 
+    if set(['MAKEFLAGS', 'MFLAGS', 'MAKELEVEL']).issubset(set(os.environ.keys())):
+        # The setup.py has been invoked by a top-level Ceph make.
+        # Set the appropriate CFLAGS and LDFLAGS
+        CEPH_ROOT = subprocess.check_output(['git', 'rev-parse', '--show-toplevel']).strip().decode('utf-8')
+
+        compiler.add_include_dir(os.path.join(CEPH_ROOT, 'src', 'include'))
```
tchaikov: `CFLAGS` and `LDFLAGS` are passed to `setup.py` explicitly in `Distutils.cmake` and `Makefile.am`, i think they would ready the `compiler` to do its job? iirc, distutils do check the env variables to figure out the cflags and ldflags.

tchaikov: okay, probably `customize_compiler()` is different from unixcompiler...

hansbogert:
Sorry for necromancing an old issue, but I think my question applies here. I am not sure if I misunderstand the intent of all the issues related to pypi, but wasn't the whole idea to have updated versions of 'python-cephlibs' in pypi?

PR #9838, opened 2016-06-21
client: misc fixes

xiexingguo:


gregsfortytwo:
Can you split out futime() into a different PR? And plumb it through libcephfs and anywhere else it should be used? And discuss why we want it? :)

The SyntheticClient stuff looks fine, although it's pretty close to dead code.

src/client/SyntheticClient.cc
```diff
@@ -3288,17 +3288,16 @@ void SyntheticClient::import_find(const char *base, const char *find, bool data)
...
 
-	//client->chmod(f.c_str(), mode & 0777);
-	client->chown(f.c_str(), uid, gid);
+	client->fchown(fd, uid, gid);
 
 	struct utimbuf ut;
 	ut.modtime = mtime;
 	ut.actime = mtime;
-	client->utime(f.c_str(), &ut);
+        client->futime(fd, &ut);
```
xiexingguo: @gregsfortytwo I add the futime() API because I want to use it here, where we already keep the file opened and have fd in hand. I am not sure this is the typical case, though.

gregsfortytwo: Oh, I see. I'm not sure it's worth it just to make the SyntheticClient perform a little better, but I'll take a look.

src/client/Client.cc
```diff
@@ -6818,6 +6818,24 @@ int Client::utime(const char *relpath, struct utimbuf *buf)
   return _setattr(in, &attr, CEPH_SETATTR_MTIME|CEPH_SETATTR_ATIME);
 }
 
+int Client::futime(int fd, struct utimbuf *buf)
+{
+  Mutex::Locker lock(client_lock);
+  tout(cct) << "utime" << std::endl;
```
gregsfortytwo: s/utime/futime/

src/client/Client.cc
```diff
@@ -6818,6 +6818,24 @@ int Client::utime(const char *relpath, struct utimbuf *buf)
...
+  tout(cct) << fd << std::endl;
+  tout(cct) << buf->modtime << std::endl;
+  tout(cct) << buf->actime << std::endl;
+  Fh *f = get_filehandle(fd);
+  if (!f)
+    return -EBADF;
```
gregsfortytwo: How about grabbing the filehandle and printing it out with the fd.

gregsfortytwo:
I guess we can test and merge with that fixup to avoid confusing which function it is. :)

xiexingguo:
Tested at: http://www.daisycloud.org:9091/xxg-2016-06-23_20:47:59-fs-wip-xxg-testing-fs-2016-06-23---basic-plana/

gregsfortytwo:
This missed my recent testing branch, but LGTM

jcsp:
@xiexingguo is the efficiency gain from using futime having a measurable impact for you?  Can you say how you're using SyntheticClient such that this became an issue?

I don't necessarily mind the futime function, but I want to be sure there is a practical need for it before adding more lines to the already huge Client.cc

jcsp:
(and this does pass tests http://pulpito.ceph.com/jspray-2016-06-28_01:48:02-fs-wip-jcsp-testing-20160627-xxg-distro-basic-mira)

xiexingguo:
> @xiexingguo is the efficiency gain from using futime having a measurable impact for you? Can you say how you're using SyntheticClient such that this became an issue?

@jcsp Well, john, I introduce the `futime` API when I happened to inspect code at https://github.com/ceph/ceph/pull/9838/commits/9971e1de77063f371871974786daa13284852f32. So in fact there is no measurement data at all.

And I am OK to drop this change, and it is up to you.

xiexingguo:
@gregsfortytwo @jcsp 
I dropped both the `futime` change and the other one that depends on it.
I guess this one is safe too merge now? Since it passed two round of testing.

xiexingguo:
@jcsp Is this one still relevant?

jcsp:
Remaining bits look fine, merged - thanks

xiexingguo:
@jcsp you are welcome:-)

PR #9840, opened 2016-06-21
os/bluestore: minor improvements for bitmap-freelist-manager

xiexingguo:


src/os/bluestore/BitmapFreelistManager.cc
```diff
@@ -13,7 +13,7 @@
 
 void make_offset_key(uint64_t offset, std::string *key)
 {
-  key->reserve(10);
+  key->reserve(8);
```
liewegas: I'm worried this will require a reallocation since I believe the current std::string will allocate space for a \0 so that c_str() does not require a temporary copy to be (re)allocated.  10 is just enough to be safe.

xiexingguo: @liewegas Dropped this one too.

PR #9783, opened 2016-06-17
msg/async: Support close idle connection feature 

yuyuyu101:
1. use c++ features
2. avoid lock in timer
3. replace some context with inline lambda

impl http://tracker.ceph.com/issues/16366

PR #9841, opened 2016-06-21
msg/async: fix trying to wake up worker event which is already waked up when shuting down msgr

david-z:
Fixes: #16396

This issue was found in jewel when restarting/stopping mds. It took long time for mds to completely stop until mds thread timed out. This is caused by when shutting down async msgr, it will try to dispatch external barrier event to a worker and wake it up, but this worker has been processed clean_handler event before, so its event status is already waked up. Hence this barrier event can't be processed further and msgr thread will be stuck at waiting for barrier count.

In shutting down msgr, we can check if this worker's event status is already waked up. If yes, I think there is no need to barrier this worker.

### DIFF
--- a/src/msg/async/AsyncMessenger.cc
+++ b/src/msg/async/AsyncMessenger.cc
@@ -428,6 +428,12 @@ void WorkerPool::barrier()
   pthread_t cur = pthread_self();
   for (vector<Worker*>::iterator it = workers.begin(); it != workers.end(); ++it) {
     assert(cur != (*it)->center.get_owner());
+    if ((*it)->center.already_wakeup.read()) {
+      ldout(cct, 10) << __func__ << " worker Event=" << &((*it)->center) 
+        << " already_wakeup" << dendl;
+      continue;
+    }
+
     barrier_count.inc();
     (*it)->center.dispatch_event_external(EventCallbackRef(new C_barrier(this)));
   }


david-z:
@yuyuyu101 

When stopping async connection in mark_down, it will dispatch clean_handler event to this worker at last. So the already_wakeup will be set to 1. Then dispatching barrier event won't wake up event again.

yuyuyu101:
Hmm, I review the code it actually may happen what you said. The fact is that "already_wakeup" isn't set to 0 when it already wakeup and execute external event.

So I think the suitable fix is use "external_num_events" to indicate whether we need to wakeup instead of extra flag "already_wakeup"

what do you think?

david-z:
Sorry that I don't fully understand your suggestion. Do you mean to use "external_num_events" in wakeup function to indicate whether we need to wakeup? if yes, we still need to check "already_wakup" in dispatch_event_external function before increasing "external_num_events", right?

Or you just mean discarding "already_wakeup" everywhere?

david-z:
I kinda understand what you were saying. If we just use "external_num_events", is "already_wakeup" no use any more, even if in C_handle_notify callback?

yuyuyu101:
yes, we can remove already_wakeup

On Tue, Jun 21, 2016 at 6:16 PM, Zhi Zhang <notifications@github.com> wrote:

> I kinda understand what you were saying. If we just use
> "external_num_events", is "already_wakeup" no use any more, even if in
> C_handle_notify callback?
>
> —
> Reply to this email directly, view it on GitHub
> <https://github.com/ceph/ceph/pull/9841#issuecomment-227398707>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AA72YZKE4S0zNJ7yQNwQELoOPZnjLyRMks5qN7nwgaJpZM4I6iO3>
> .
>



-- 

Best Regards,

Wheat

david-z:
@yuyuyu101 I just look through the code again. 

1. You are right that "already_wakeup" isn't used for indicting whether we need to wakeup or not. So we don't need to check "already_wakeup" in wakeup function. But I think we could keep "already_wakeup" because it still can be used for C_handle_notify to read pipe asap. It is no harm to keep it.

2. For your suggestion to use "external_num_events" in wakeup function, since time_event will also call wakeup, so I think just checking "external_num_events" is not very reasonable. 
Actually we don't need to check anything (either "already_wakeup" or "external_num_events") in wakeup function. We can set "already_wakeup" to 1 first like now and then write pipe, because once caller wants to call wakeup, that means it needs to do wakeup.

What do you think?

yuyuyu101:
@david-z we will remove wakeup in time_event (https://github.com/ceph/ceph/pull/9783). So I think it's ok ..

yuyuyu101:
@david-z Hmm, I rethink this because we should backport this, but we won't backport #9783 . So I think we should follow your idea firstly. After merging #9783, we can forward to my idea.

david-z:
Thanks Haomai for pointing me #9783. Your idea is right. So before merging #9783, I will change this PR according to our discussion.

david-z:
@yuyuyu101 

I have found out the root cause and will send another PR soon with the fix and also the change we discussed here together. So close this one for now.

PR #9842, opened 2016-06-21
vstart: set osd_check_max_object_name_len_on_startup

jcsp:
For the benefit of those running vstart clusters on local dev machines with a filesystem that trips the object name len check.

### DIFF
--- a/src/vstart.sh
+++ b/src/vstart.sh
@@ -513,6 +513,7 @@ $CMDSDEBUG
 $extra_conf
 [osd]
 $DAEMONOPTS
+        osd_check_max_object_name_len_on_startup = false
         osd data = $CEPH_DEV_DIR/osd\$id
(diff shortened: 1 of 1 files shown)


wjwithagen:
@jcsp 
I my test environment I actually added code to vstart to run it short on certain platforms(FreeBSD), where short would be nice. Perhaps could do that as well of certain disktypes under Linux?

liewegas:
My only concern is that rgw will now fail unpredictably with long object names.  The idea was to make it explicit so that the user was aware or moved off of ext4...

gregsfortytwo:
Well this PR is just for vstart, where all of our new developers (and many long-standing ones) so far gotten stuck on why their ext4-based system won't start up a functioning cluster... ;)

That's obviously not suitable for real systems but I think it's fine here.

liewegas:
Fair enough!

PR #9843, opened 2016-06-21
Wip rgw : bucket create - avoid listing all entries

theanalyst:
Avoid counting the buckets/omap entries during bucket create, update the stats in the bucket header object instead. Now we do have an additional call for omap header update too when we do deletion; on the other hand we avoid getting the full list of keys during bucket create. We can probably improvise here by deferring this check until we actually link the bucket maybe.. (modify one of the user to hold the max_buckets also maybe?)

### DIFF
--- a/src/cls/user/cls_user.cc
+++ b/src/cls/user/cls_user.cc
@@ -100,30 +100,31 @@ static int read_header(cls_method_context_t hctx, cls_user_header *header)
   return 0;
 }
 
-static void add_header_stats(cls_user_stats *stats, cls_user_bucket_entry& entry, bool need_bucket_add=false)
+static void add_header_stats(cls_user_stats *stats, cls_user_bucket_entry& entry, bool sync_bucket_count=false)
 {
   stats->total_entries += entry.count;
   stats->total_bytes += entry.size;
   stats->total_bytes_rounded += entry.size_rounded;
-  if (need_bucket_add)
+  if (sync_bucket_count)
     stats->total_buckets += entry.bucket_count;
 }
 
-static void dec_header_stats(cls_user_stats *stats, cls_user_bucket_entry& entry, bool need_delete=false)
+static void dec_header_stats(cls_user_stats *stats, cls_user_bucket_entry& entry, bool sync_bucket_count=false)
 {
   stats->total_bytes -= entry.size;
   stats->total_bytes_rounded -= entry.size_rounded;
   stats->total_entries -= entry.count;
-  if (need_delete)
+  if (sync_bucket_count)
     stats->total_buckets -= entry.bucket_count;
 }
 
-static void apply_entry_stats(const cls_user_bucket_entry& src_entry, cls_user_bucket_entry *target_entry)
+static void apply_entry_stats(const cls_user_bucket_entry& src_entry, cls_user_bucket_entry *target_entry, bool sync_bucket_count=false)
 {
   target_entry->size = src_entry.size;
   target_entry->size_rounded = src_entry.size_rounded;
   target_entry->count = src_entry.count;
-  target_entry->bucket_count = src_entry.bucket_count;
+  if (sync_bucket_count)
+    target_entry->bucket_count = src_entry.bucket_count;
 }
 
 static int cls_user_set_buckets_info(cls_method_context_t hctx, bufferlist *in, bufferlist *out)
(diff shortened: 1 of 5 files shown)


theanalyst:
Not the final version... just posting here for initial feedback cc @cbodley

cbodley:
I've been reading through more of the quota and user stats code, and see that the gateway is doing some caching and lazy updates of the user's bucket stats. In rgw_quota.cc, the `RGWQuotaHandlerImpl` is checking quotas against a `RGWUserStatsCache` in memory. A `UserSyncThread` is calling `RGWUserStatsCache::sync_all_users()` once in a while, which ends up in `RGWRados::cls_user_sync_bucket_stats()`. This adds up stats from the bucket's headers and passes them to `cls_user_update_buckets()` with add=false to set the stats rather than increment them. And it looks like the `cls_user_bucket_entry::count` represents the number of objects, not buckets.

So we probably want to add a new field for `cls_user_stats::total_buckets` and make sure that it's only incremented when called through `rgw_link_bucket()`. You could add a `bool create` flag in `cls_user_set_buckets_op` to do that.

theanalyst:
@cbodley ah ok makes sense... (I was wondering what count meant.. but yeah this clears it), let's try adding the total buckets in the cls_user_stats. Should we pass down the max-buckets also so that cls-user-set-buckets makes a decision?

cbodley:
> Should we pass down the max-buckets also so that cls-user-set-buckets makes a decision?

We might want to try that part in a different PR.. by the time we get to the `rgw_link_bucket()` in `RGWCreateBucket::execute()`, we've already called `RGWRados::create_bucket()`. So it may take extra work and testing to make sure we can fail gracefully and clean everything up when `rgw_link_bucket()` returns an error

src/cls/user/cls_user.cc
```diff
@@ -262,16 +268,20 @@ static int cls_user_remove_bucket(cls_method_context_t hctx, bufferlist *in, buf
     return ret;
   }
 
-  if (entry.user_stats_sync) {
-    dec_header_stats(&header.stats, entry);
-  }
+
+  /* if (entry.user_stats_sync) { */
```
cbodley: i wasn't able to figure out what this `user_stats_sync` stuff was for; are you sure we can remove this check?

theanalyst: Yeah I'm digging into this right now, initially when I was removing without forcing the sync, the count wasn't decreased have to figure when this is set.

cbodley:
I brought this up with @yehudasa yesterday, and he made a good point about backward compatibility with existing users/buckets.. we'll need a way to detect that bucket_count is missing, and do an actual bucket listing to initialize it. This is tricky, though, because it could race with other operations, and potentially take a long time and lead to laggy osd ops.

theanalyst:
@cbodley hmm you've a point there... Probably can't solve that in an easy way :/

theanalyst:
I guess we could either 1. have another method in `rgw_rados` which only increments the count instead of adding entries to the map 2. maybe do this in the cls itself?  Both of this might not buy us much, but might save a little I guess?

theanalyst:
@cbodley @yehudasa any benefits if we use the above approaches?

yehudasa:
@theanalyst when cls_user accounts the total_buckets it should only add up buckets that hasn't been accounted before. So when adding buckets there we should provide the bucket names (if we don't) and we should add a flag for each bucket entry to specify whether it is accounted in the total_buckets or not.

theanalyst:
@yehudasa won't this approach also have the same problem wrt to the backwards compatibility with the old buckets?
Maybe We could probably get around this when iterating the cls_user_bucket_entry loop and updating as we go probably?

dachary:

theanalyst:
jenkins test this please

theanalyst:
test this please

liewegas:
@theanalyst ping?

theanalyst:
@liewegas I'm still up for fixing this again, kind of fell through the cracks :/
@yehudasa, Let me see if I get this, so we create an flag in bucket entry which indicates whether it is accounted for or not, and in  `cls_user_stats` we have both the buckets count and a set containing bucket names and we use this to check, and probably in cls_user_list_buckets, compare the count and update?

yehudasa:
@theanalyst I'm not 100% sure I remember what I was thinking at the time. Could be something like that:
 - cls_user omap header contains a total_buckets field
 - in order to update the field, rgw sends cls_user an operation with a a list of buckets (bucket instances) it wants to add
 - in the existing data we hold for each bucket in the cls_user omap kv store, we should add a field that specifies whether this bucket was accounted in the header
 - rgw should be able to notice when cls_user hasn't had this new field initialized, and when that happens:
  - rgw lists user's buckets (with pagination)
  - rgw updates cls_user with list of buckets so that it can account these buckets in its internal counting

does that make any sense?

theanalyst:
needs rethinking now

PR #9845, opened 2016-06-21
common: remove basename() dependency

dx9:


wjwithagen:
@dx9 
I think we are here looping in circles.
You code looks suspiciously like code that was once in there.
Then somebody saw fit to replace it with basename()... And I had lots of trouble getting it the way it is now. And now we are going back to handcrafted code?

dx9:
@wjwithagen

:) Ya it's a lot of work keeping things portable. The current code is not portable to musl libc or OSX.

The issue landed here:

https://github.com/ceph/ceph/commit/36af48029378d7e73a57d4ceb1a22ddfd85ef841#diff-1e9db0f73d190a53082545f04ffa1300R279

Here's the thought process behind the change:

We have two choices:

1) Remove the usage - Nixes the portability issue and respects the functional intentions of the current code.

2) Tweak it for POSIX compliance - This would work as well.

In the interest of portability it would be sweet to discourage/question the use of basename(). It's handy, but it's easy to use in the non portable fashion unless you google like crazy. GNU and FreeBSD provide basename(const char*) as default implementations which promote non portable code.

https://www.gnu.org/software/libc/manual/html_node/Finding-Tokens-in-a-String.html
https://www.freebsd.org/cgi/man.cgi?query=basename&sektion=3&manpath=FreeBSD+10.3-RELEASE

Do we really need basename()?

LMK

src/common/config.cc
```diff
@@ -267,8 +263,7 @@ int md_config_t::parse_config_files_impl(const std::list<std::string> &conf_file
      * If cluster name is not set yet, use the prefix of the
      * basename of configuration file as cluster name.
      */
-    const char *fn = c->c_str();
-    std::string name(basename(fn));
+    std::string name = c->substr(c->find_last_of("/") + 1);
```
tchaikov: as `c_str()` returns a "const char*", and per POSIX standard[1], `basename()` expects a "char*", and it might change the string passed in. even the freebsd and glibc kindly offers `basename(const char*)` implementation.

so, yeah, i agree that we do need to change for standard compliance and better portability.

maybe we can have

```c++
auto start = c->rfind('/') + 1;
auto end = c->find(".conf", start);
if (end == name.npos) {
  ...
} else {
  cluster = name.substr(start, end - start);
}
---
[1] http://pubs.opengroup.org/onlinepubs/9699919799/

wjwithagen: On 23-6-2016 08:50, Kefu Chai wrote:
> In src/common/config.cc
> <https://github.com/ceph/ceph/pull/9845#discussion_r68183228>:
> 
>> @@ -267,8 +263,7 @@ int md_config_t::parse_config_files_impl(const std::list<std::string> &conf_file
>>       * If cluster name is not set yet, use the prefix of the
>>       * basename of configuration file as cluster name.
>>       */
>> -    const char *fn = c->c_str();
>> -    std::string name(basename(fn));
>> +    std::string name = c->substr(c->find_last_of("/") + 1);
> 
> as |c_str()| returns a "const char/", and per POSIX standard[1],
> |basename()| expects a "char/", and it might change the string passed
> in. even the freebsd and glibc kindly offers |basename(const char*)|
> implementation.

Yup,
That is where all my hassle started: |const char|.
FreeBSD does grok that, but only if the right library is included.
At the time I looked at several man-pages, but not at the POSIX std.

Note that performance is is not an issue.
So if manual crafted code needs to do it that is fine.

Not sure why it went from manual code to to basename() first place.

--WjW

dx9:
@tchaikov @wjwithagen Updated. Thanks guys!

tchaikov:
lgtm.

wjwithagen:
Oke.

tchaikov:
tested at 

PR #9846, opened 2016-06-21
msg/AsyncMessenger: remove unneeded include

stiopaa1:


yuyuyu101:
actually <map> is useless too. <set> is needed

stiopaa1:
@yuyuyu101 
You are saying that the 
`
#include <list>
`
is needed?

Where?

yuyuyu101:
@stiopaa1 sorry, github hidden my "<" chars. I mean "actually map is useless too. set is needed"

PR #9848, opened 2016-06-21
configure.ac: add _LIBS variables for boost_system and boost_iostreams

dx9:


tchaikov:
@dx9 have you tried to compile ceph on Alpine? as we are moving away from autotools, autotools build will not be maintained in the near future. so i am not sure if the efforts to fix the build on Alpine is worthy.

dx9:
@tchaikov yeppers. Both the autotools and cmake builds work on Alpine with the lastest batch of PRs. I used the autotools build for reference and testing. Also #9853 has a decent pass at packaging based on autotools. I wasn't sure when you guys were gonna cut over. I haven't tried the cmake testing line on Alpine yet.

I'm easy so no worries if you'd just prefer to ditch the configure.ac PRs. I'm fine with moving the packaging over to cmake.

LMK

tchaikov:
we won't drop the configure.ac and Makefile*.am. and iiuc, autotools will not be updated timely as it is after the change. the PR for autotools are still welcome.

tchaikov:
> The checks were breaking downtstream checks for fallocate and others on Alpine.

could you elaborate a little bit how/why it breaks the checks for fallocate?

dx9:
fallocate() check would fail when it should pass. I never figured out why. Worked backwards commenting out blocks until it worked. :) Eventually realized it was the boost_iostreams and boost_system checks. Updated them to look like the other ones that worked. Also had to add them to the linking deps.

tchaikov:
lgtm.

PR #9849, opened 2016-06-21
configure.ac: add fallback to libexecinfo for backtrace(3)

dx9:


tchaikov:
lgtm.

dx9:
@tchaikov I've been testing this more on Alpine and have realized it is not a workable solution. It looks like the better approach is using libunwind.

PR #9850, opened 2016-06-21
librbd: flag image as updated after proxying maintenance op

dillaman:
Fixes: http://tracker.ceph.com/issues/16404


trociny:
lgtm

PR #9851, opened 2016-06-21
rgw: add pg_ver to tombstone_cache

cbodley:
a tombstone cache was added to remember the mtime of deleted objects for
use with the HTTP_IF_MODIFIED_SINCE header, but the comparison was still
failing because of a missing pg_ver. added pg_ver to the tombstone cache
so it can be passed with HTTP_DEST_PG_VER

### DIFF
--- a/src/rgw/rgw_rados.cc
+++ b/src/rgw/rgw_rados.cc
@@ -7757,9 +7757,9 @@ int RGWRados::Object::Delete::delete_obj()
   int64_t poolid = ref.ioctx.get_id();
   if (r >= 0) {
     tombstone_cache_t *obj_tombstone_cache = store->get_tombstone_cache();
-    pair<ceph::real_time, uint32_t> tombstone_entry = make_pair<>(state->mtime, state->zone_short_id);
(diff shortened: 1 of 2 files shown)


theanalyst:
I'll run this by and see if I can see the old behaviour again

cbodley:
@yehudasa confirmed that we don't compare pg_ver unless they're from the same zone:
```c++
  bool operator<(const obj_time_weight& rhs) {
    ...
    if (zone_short_id != rhs.zone_short_id) {
      return (zone_short_id < rhs.zone_short_id);
    }
    return (pg_ver < rhs.pg_ver);
  }
```

yehudasa:
@cbodley perfect

yehudasa:
@theanalyst did you test this?

theanalyst:
Yeah ran this through the same script, no errors so far

mattbenjamin:
The main questions I have are longer-term:
1. what allows us to rely on a memory-only cache?
2. can the cache size limit be overrun?  what happens in that case?

PR #9852, opened 2016-06-21
rgw-ldap: add ldap lib to rgw lib deps based on build config

dx9:
@wjwithagen We've definitely bumped into the same issues. Please take a look.


liewegas:
@wjwithagen ?

mattbenjamin:
looks fine, I was a bit worried about adding this to a static link, but nothing looks amiss in the build objects

dx9:
@mattbenjamin thanks!

PR #9853, opened 2016-06-22
alpine: add alpine linux dev support

dx9:


mattbenjamin:
@fwiw https://www.brianchristner.io/docker-is-moving-to-alpine-linux/

dx9:
https://github.com/alpinelinux/aports/pull/295

dx9:
@liewegas thanks!

PR #9856, opened 2016-06-22
librbd: discard should return error if length too large

hzhang-wx:
rbd discard use 'int' to return the discarded length, but the 'len'
user passed is 'uint64', in some case, the return value will be
truncated and return a negative value which means discard failed.
use ssize_t to avoid.

Signed-off-by: Huan Zhang <zhanghuan@chinac.com>

### DIFF
--- a/src/librbd/librbd.cc
+++ b/src/librbd/librbd.cc
@@ -1160,6 +1160,10 @@ namespace librbd {
   {
     ImageCtx *ictx = (ImageCtx *)ctx;
     tracepoint(librbd, discard_enter, ictx, ictx->name.c_str(), ictx->snap_name.c_str(), ictx->read_only, ofs, len);
+    if (len > std::numeric_limits<int32_t>::max()) {
+        tracepoint(librbd, discard_exit, -EINVAL);
+        return -EINVAL;
+    }
     int r = ictx->aio_work_queue->discard(ofs, len);
     tracepoint(librbd, discard_exit, r);
     return r;


hzhang-wx:
in some case, rbd discard failed when uint64(len) > 'max int size' due to the return value truncated.
use ssize_t instead to fix. please review, thx.

trociny:
@hzhang-wx  Could you please create a tracker ticket and refer it in the commit log?

It would be also nice to clean test/librbd/test_librbd.cc up. For example, in `aio_discard_test_data()` type for `r` var that stores `rbd_aio_get_return_value` return value should be fixed; in printf statements   the cast to int does not look nice. Then may be add a test for this fix, e.g. into TestEmptyDiscard?

dillaman:
The trouble is that the ABI will break if you change ```int``` -> ```ssize_t``` since in a 64bit environment, ```int``` is 4 bytes while ```ssize_t``` is 8.

dillaman:
... or just return -EINVAL if the len is > MAX_INT32 and using the current discard API

trociny:
> or just return -EINVAL if the len is > MAX_INT32 and using the current discard API

may be return 0 or MAX_INT32 instead? So the user knows it succeeded.

hzhang-wx:
yeah, usually, modify the api declaration is not a good way. and in this case,  
if len <= MAX_INT32，change int -> ssize_t， get the correct written len as before.
if len >  MAX_INT32， change int -> ssize_t， get the truncated value as before too.
so it will not affect the running program behavior, right?
the only negative effect is compilation may print a warning or an error depend on compiler setting.
but it's not bad for users, right?

hzhang-wx:
@dillaman @trociny 
for clean test/librbd/test_librbd.cc up: int r = (int64) rbd_aio_get_return_value
many function besides '*discord*'  should be clean, seems draft an another patch is more proper .
how do you think?

dillaman:
@hzhang-wx Actually, you will get a compile error if someone was using a function pointer to the method. Hence why we shouldn't break the API/ABI.

src/librbd/librbd.cc
```diff
@@ -50,6 +50,8 @@
 #undef dout_prefix
 #define dout_prefix *_dout << "librbd: "
 
+#define MAX_INT32 ((uint32_t) -1 >> 1)
```
dillaman: You can just use ```std::numeric_limits<int32_t>::max()```

dillaman:
@hzhang-wx Please squash this down to a single commit or rebase and drop the first two commits. Thanks.

PR #9857, opened 2016-06-22
add socket backlog setting for via ceph.conf

diluga:
RGW : setting socket backlog for via ceph.conf


smithfarm:
@diluga Thanks for your patch. It appears that the second commit fixes a mistake in the first one - if that is the case, the two commits should be squashed into one. You can search on-line for "git interactive rebase" if you don't know how to do this. 

Also, all commits have to be signed - see https://github.com/ceph/ceph/blob/master/SubmittingPatches.rst (tl;dr "git commit -s")

Once you've fixed these nits, you can "git push -f" your branch to update the PR.

PR #9859, opened 2016-06-22
mon: osdmap's epoch should be more than 0

x11562:
Signed-off-by: Na Xie xie.na@h3c.com

### DIFF
--- a/src/mon/OSDMonitor.cc
+++ b/src/mon/OSDMonitor.cc
@@ -3104,11 +3104,9 @@ bool OSDMonitor::preprocess_command(MonOpRequestRef op)
 
     epoch_t epoch = 0;
     int64_t epochnum;
-    cmd_getval(g_ceph_context, cmdmap, "epoch", epochnum, (int64_t)0);
+    cmd_getval(g_ceph_context, cmdmap, "epoch", epochnum, (int64_t)osdmap.get_epoch());
     epoch = epochnum;
-    if (!epoch)
-      epoch = osdmap.get_epoch();
-
+    
     bufferlist osdmap_bl;
     int err = get_version_full(epoch, osdmap_bl);
     if (err == -ENOENT) {


src/mon/OSDMonitor.cc
```diff
@@ -3106,7 +3106,7 @@ bool OSDMonitor::preprocess_command(MonOpRequestRef op)
     int64_t epochnum;
     cmd_getval(g_ceph_context, cmdmap, "epoch", epochnum, (int64_t)0);
     epoch = epochnum;
-    if (!epoch)
+    if (epoch > 0)
```
tchaikov: does this change make any difference? or it is just cosmetic?

x11562:
```
ceph osd getmap 0
```
such as command, get the latest osdmap.
this change tells that 0 is invalid.

tchaikov: this change breaks the behaviour of 
`ceph osd getmap`.

in this case, `epoch` is 0.

x11562: 0 is invalid

tchaikov: the default value of `epoch` is "0" if it's not specified. we use it to tell if user wants to get the latest osdmap.

with your change, if user puts `ceph osd getmap`. he/she will get an error.
```
$ ./bin/ceph osd getmap
Error ENOENT: there is no map for epoch 0
```

tchaikov: further more, you are overwriting the epoch passed by user with the latest one. this does not sound right.

x11562: if epoch is not specified, the latest epoch returned，rather than setted to 0, 
what do you think?

dachary:

tchaikov:
@jecluis agreed. but the change improves the readability, because the last param of `cmd_getval()` is the default value of the val of the command option, and we are using "0" as a flag actually. so i lean towards merging it after running it through rados qa.

jecluis:
@tchaikov fair enough. thumbsup.png

tchaikov:
http://pulpito.ceph.com/kchai-2016-11-25_08:09:33-rados-wip-kefu-testing2---basic-mira/

PR #9860, opened 2016-06-22
client: return errno is err

renhwztetecs:
Fixes: http://tracker.ceph.com/issues/16407


renhwztetecs:
unable to obtain rotating service keys; retrying

PR #9864, opened 2016-06-22
ceph_volume_client: store authentication metadata

ajarr:
ceph_volume_client: store authentication metadata

Store a two-way mapping between auth IDs and volumes.

Enables us to record some metadata on auth ids (which
openstack tenant created it) so that we can avoid exposing
keys to other tenants who try to use the same ceph
auth id.

Enables us to expose the list of which auth ids have access
to a volume, so that Manila's update_access() can be
implemented efficiently.

Fixes: http://tracker.ceph.com/issues/15615
Fixes: http://tracker.ceph.com/issues/15406

### DIFF
--- a/src/pybind/ceph_volume_client.py
+++ b/src/pybind/ceph_volume_client.py
@@ -223,7 +223,6 @@ class CephFSVolumeClient(object):
         self.auth_id = auth_id
         self.volume_prefix = volume_prefix if volume_prefix else self.DEFAULT_VOL_PREFIX
         self.pool_ns_prefix = pool_ns_prefix if pool_ns_prefix else self.DEFAULT_NS_PREFIX
-
         # For flock'ing in cephfs, I want a unique ID to distinguish me
         # from any other manila-share services that are loading this module.
         # We could use pid, but that's unnecessary weak: generate a
@@ -236,10 +235,6 @@ class CephFSVolumeClient(object):
         # TODO: remove .meta files on last rule for an auth ID deletion
         # TODO: implement fsync in bindings so that we don't have to syncfs
         # TODO: version the on-disk structures
-        # TODO: check dirty flag after locking something and call recover()
-        # if we are opening something dirty (racing with another instance
-        # of the driver restarting after failure) -- only required if someone
-        # running multiple manila-share instances with Ceph loaded.
 
     def recover(self):
         # Scan all auth keys to see if they're dirty: if they are, they have
@@ -251,17 +246,28 @@ class CephFSVolumeClient(object):
         # we're updating it with being dirty at the same time.
 
         # First list the auth IDs that have potentially dirty on-disk metadata
-        dir_handle = self.fs.opendir(self.volume_prefix)
+        log.debug("Recovering from partial auth updates (if any)...")
+
+        try:
+            dir_handle = self.fs.opendir(self.volume_prefix)
+        except cephfs.ObjectNotFound:
+            log.debug("Nothing to recover. No auth meta files.")
+            return
+
         d = self.fs.readdir(dir_handle)
         auth_ids = []
-        while d:
-            d = self.fs.readdir(dir_handle)
 
+        if not d:
+            log.debug("Nothing to recover. No auth meta files.")
+
+        while d:
             # Identify auth IDs from auth meta filenames. The auth meta files
             # are named as, "$<auth_id>.meta".
             match = re.search("^\$(.*)\.meta$", d.d_name)
             if match:
                 auth_ids.append(match.group(1))
+
+            d = self.fs.readdir(dir_handle)
 
         self.fs.closedir(dir_handle)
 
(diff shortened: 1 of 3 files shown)


ajarr:
@jcsp , updating work so far.

src/pybind/ceph_volume_client.py
```diff
@@ -848,6 +856,11 @@ def authorize(self, volume_path, auth_id, readonly=False, tenant_id=None):
                 # have mon auth caps that prevent it from accessing those keys
                 # (e.g. limit it to only access keys with a manila.* prefix)
             else:
+                if auth_meta['tenant_id'] != tenant_id.__str__():
```
ajarr: check only if tenant_id is None, i.e., the caller is not multi-tenant aware.

src/pybind/ceph_volume_client.py
```diff
@@ -710,6 +710,23 @@ def _metadata_set(self, path, data):
     def _lock(self, path):
         @contextmanager
         def fn():
+            while(1):
+                fd = self.fs.open(path, os.o_creat, 0755)
+                self.fs.flock(fd, fcntl.lock_ex, self._id)
+
+                # The locked file could have been been deleted e.g, by an
```
ajarr: s/have been been/have been/

src/pybind/ceph_volume_client.py
```diff
@@ -216,6 +232,103 @@ def __init__(self, auth_id, conf_path, cluster_name, volume_prefix=None, pool_ns
...
+            (group_id, volume_id) = volume.split('/')
+            volume_path = VolumePath(group_id, volume_id)
+            access_level = auth_meta['volumes'][volume]['access_level']
+
+            with self._volume_lock(volume_path):
+                vol_meta = self._volume_metadata_get(volume_path)
+
+                if auth_id not in vol_meta['auths']:
+                    del auth_meta['volumes'][volume]
+                    self._auth_metadata_set(auth_id, auth_meta)
```
jcsp: It's not necessary to auth_metadata_set up here, since it's called at the end of this function, right?

src/pybind/ceph_volume_client.py
```diff
@@ -216,6 +232,103 @@ def __init__(self, auth_id, conf_path, cluster_name, volume_prefix=None, pool_ns
...
+
+            (group_id, volume_id) = volume.split('/')
+            volume_path = VolumePath(group_id, volume_id)
+            access_level = auth_meta['volumes'][volume]['access_level']
+
+            with self._volume_lock(volume_path):
+                vol_meta = self._volume_metadata_get(volume_path)
+
+                if auth_id not in vol_meta['auths']:
+                    del auth_meta['volumes'][volume]
```
jcsp: Deleting thing from a container while iterating over it is best avoided.  Build a list of volume IDs to delete, and then delete them in a separate loop.

src/pybind/ceph_volume_client.py
```diff
@@ -216,6 +232,103 @@ def __init__(self, auth_id, conf_path, cluster_name, volume_prefix=None, pool_ns
...
+                    continue
+                if not auth_meta['dirty']:
+                    continue
+                self._recover_auth_meta(auth_id, auth_meta)
+
+    def _recover_auth_meta(auth_id, auth_meta):
+        """
+        Call me after locking the auth meta file.
+        """
+        for volume in auth_meta['volumes']:
```
jcsp: Prefer to use .items() here than repeatedly doing auth_meta['volumes'][volume] within the loop to get the value.

ajarr: Thanks! Agree. Code is hard/ugly to read now.

src/pybind/ceph_volume_client.py
ajarr: Understood.

src/pybind/ceph_volume_client.py
ajarr: Yes. I'll take care of this in the cleanup (in a separate loop) as you'd suggested.

ajarr:
@jcsp, the approaches for recovery, and change in metadata layouts, look ok? checking, as we'd agreed upon a different metadata layout.

src/pybind/ceph_volume_client.py
```diff
@@ -673,13 +1008,76 @@ def cap_update(orig, want, unwanted):
...
+            if auth_meta['dirty']:
+                self._recover_auth_meta(auth_id, auth_meta)
+
+            auth_meta['dirty'] = True
+            auth_meta['volumes'][volume_path_str]['dirty'] = True
+            self._auth_metadata_set(auth_id, auth_meta)
+
+            self._deauthorize_volume(volume_path, auth_id)
+
+            # Filter the volume we're deauthorizing out
```
ajarr: Filter out the volume we're deauthorizing.

src/pybind/ceph_volume_client.py
```diff
@@ -690,12 +1088,11 @@ def deauthorize(self, volume_path, auth_id):
 
         # The auth_id might have read-only or read-write mount access for the
         # volume path.
-        access_levels = ('r', 'rw')
         want_mds_caps = {'allow {0} path={1}'.format(access_level, path)
-                         for access_level in access_levels}
+                         for access_level in ('r', 'rw')}
```
ajarr: these changes are not needed.

src/pybind/ceph_volume_client.py
```diff
@@ -710,6 +710,23 @@ def _metadata_set(self, path, data):
...
+            while(1):
+                fd = self.fs.open(path, os.o_creat, 0755)
+                self.fs.flock(fd, fcntl.lock_ex, self._id)
+
+                # The locked file could have been been deleted e.g, by an
+                # another manila share instance. Perform checks to ensure that
+                # the lock is not applied on an unlinked file.
+                try:
+                    statbuf = self.fs.stat(path)
+                except cephfs.objectnotfound:
```
jcsp: isn't it ObjectNotFound?

src/pybind/ceph_volume_client.py
```diff
@@ -710,6 +710,23 @@ def _metadata_set(self, path, data):
...
+                    self.fs.close(fd)
+                    continue
+
+                fstatbuf = self.fs.fstat(fd)
+                if statbuf.st_ino == fstatbuf.st_ino:
+                    break
+
             fd = self.fs.open(path, os.O_CREAT, 0755)
```
jcsp: You'd want to remove this open/flock call, as it's already been done in the loop above, right?

jcsp:
The format changes look reasonable to me -- we're committing to never having overlapping 'r' and 'rw' rules for one auth ID on one volume, but that is a good thing for simplicity.

src/pybind/ceph_volume_client.py
ajarr: correct.

src/pybind/ceph_volume_client.py
ajarr: correct. missed to remove these lines.

src/pybind/ceph_volume_client.py
ajarr: Done.

src/pybind/ceph_volume_client.py
ajarr: Done.

src/pybind/ceph_volume_client.py
ajarr: Done.

src/pybind/ceph_volume_client.py
ajarr: Done.

src/pybind/ceph_volume_client.py
ajarr: Done.

src/pybind/ceph_volume_client.py
ajarr: Done.

src/pybind/ceph_volume_client.py
ajarr: Done.

src/pybind/ceph_volume_client.py
ajarr: Done.

ajarr:
partially tested by https://github.com/ceph/ceph-qa-suite/pull/1080

src/pybind/ceph_volume_client.py
```diff
@@ -239,6 +249,7 @@ def __init__(self, auth_id, conf_path, cluster_name, volume_prefix=None, pool_ns
         # UUID
         self._id = struct.unpack(">Q", uuid.uuid1().get_bytes()[0:8])[0]
 
+        self.version = _CEPHFSVOLUMECLIENT_VERSION
```
jcsp: We need two versions here, the latest version, and the earliest version we're compatible with.

We should also make this a class attribute, so that callers don't have to have the correct arguments to __init__ before they can query the version.

ajarr: Done.

jcsp:
Reviewed-by me, let's see if we can get http://tracker.ceph.com/issues/16686 resolved today to put this through testing

jcsp:
This is still bumping into pre-existing issue http://tracker.ceph.com/issues/16686 on vstart clusters, but is passing in the lab http://qa-proxy.ceph.com/teuthology/jspray-2016-07-19_05:56:13-fs-wip-jcsp-testing-20160719-distro-basic-mira/322660/teuthology.log

PR #9866, opened 2016-06-22
os/bluestore: replaces interval_set with bitmap for bluestore_blob_t:…

ifed01:
…:unused

### DIFF
--- a/src/os/bluestore/bluestore_types.h
+++ b/src/os/bluestore/bluestore_types.h
@@ -16,6 +16,7 @@
 #define CEPH_OSD_BLUESTORE_BLUESTORE_TYPES_H
 
 #include <ostream>
+#include <bitset>
 #include "include/types.h"
 #include "include/interval_set.h"
 #include "include/utime.h"
@@ -150,9 +151,10 @@ static inline bool operator!=(const bluestore_extent_ref_map_t& l,
 /// blob: a piece of data on disk
 struct bluestore_blob_t {
   enum {
-    FLAG_MUTABLE = 1,     ///< blob can be overwritten or split
-    FLAG_COMPRESSED = 2,  ///< blob is compressed
-    FLAG_CSUM = 4,        ///< blob as checksums
+    FLAG_MUTABLE = 1,         ///< blob can be overwritten or split
+    FLAG_COMPRESSED = 2,      ///< blob is compressed
+    FLAG_CSUM = 4,            ///< blob has checksums
+    FLAG_HAS_UNUSED = 8,      ///< blob has unused map
   };
   static string get_flags_string(unsigned flags);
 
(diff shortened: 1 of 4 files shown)


src/os/bluestore/bluestore_types.h
```diff
@@ -289,25 +298,55 @@ struct bluestore_blob_t {
...
+  bool is_unused(uint64_t offset, uint64_t length, uint64_t chunk_size) const {
+    if (!has_unused()) {
+      return false;
+    }
+    uint64_t start = offset / chunk_size;
+    uint64_t end = ROUND_UP_TO(offset + length, chunk_size) / chunk_size;
+    assert( end <= unused.size());
+    auto i = start;
+    for(; i < end && unused[i]; i++);
+    return i >= end;
```
liewegas: This seems problematic: min_alloc_size / chunk_size could be almost anything, which means we could overflow the 32 bits at any time.  I see two options:

1. we fix the bit vector at 32 bits, and just the blob into 32 chunks unconditionally.  We'll probably be more or less granular than chunk_size, but as long as we underestimate the unused regions we'll be correct.  In practice most cases it will divide evenly and we'll be fine.  If they don't, we'll end up taking the wal path for some of the random writes, but that is not the end of the world.

2. We make the bitmap variably sized, and use the first few bits to encode how big it is.

I like #1.. less complex.  We might end up using 32 bits when 16 is enough.. but on the other hand we could even just say 16 bits is enough.  If the HDD default is 64k min_alloc_size and 4k block size, 16 would be fine for that case.

liewegas:
Can you break this into separate commits?

1. drop *length arg from _pad_zeros (mostly unrelated to the rest of this stuff)
2. get_chunk_size helper and associated cleanup
3. move add_unused into write_item and wctx_finish
4. switch to bitmap representation of unused

src/os/bluestore/bluestore_types.h
```diff
@@ -150,9 +151,10 @@ static inline bool operator!=(const bluestore_extent_ref_map_t& l,
 /// blob: a piece of data on disk
 struct bluestore_blob_t {
   enum {
-    FLAG_MUTABLE = 1,     ///< blob can be overwritten or split
-    FLAG_COMPRESSED = 2,  ///< blob is compressed
-    FLAG_CSUM = 4,        ///< blob as checksums
+    FLAG_MUTABLE = 1,         ///< blob can be overwritten or split
+    FLAG_COMPRESSED = 2,      ///< blob is compressed
+    FLAG_CSUM = 4,            ///< blob has checksums
+    FLAG_UNUSED_PRESENT = 8,  ///< blob h unused map
```
liewegas: FLAG_HAS_UNUSED ?  also typo in comment

src/os/bluestore/bluestore_types.h
```diff
@@ -200,7 +202,7 @@ struct bluestore_blob_t {
   uint8_t csum_chunk_order;        ///< csum block size is 1<<block_order bytes
 
   bluestore_extent_ref_map_t ref_map; ///< references (empty when in onode)
-  interval_set<uint32_t> unused;   ///< portion that has never been written to
+  std::bitset<32> unused;       ///< portion that has never been written to
```
liewegas: let's make that 32 a #define ... and 16?

liewegas: or since we need to know the unsigned equivalent, #define the type (uint16_t) and use sizeof()*8 here.  or whatever works.

ifed01:
@liewegas: updated

liewegas:
https://github.com/ceph/ceph/pull/9962 has whitespace fixup

PR #9868, opened 2016-06-22
include/small_encoding: fix lba encoding with many low zeros

liewegas:


ifed01:
LGTM

PR #9874, opened 2016-06-22
OSDMonitor::prepare_pgtemp: only update up_thru if newer

athanatos:
Fixes: http://tracker.ceph.com/issues/16185


athanatos:
samuelj-2016-06-21_10:29:56-rados-wip-sam-testing-distro-basic-smithi

PR #9875, opened 2016-06-22
OSDMonitor: drop pg temp from sources other than the current primary

athanatos:
Fixes: http://tracker.ceph.com/issues/16127


athanatos:
samuelj-2016-06-21_10:29:56-rados-wip-sam-testing-distro-basic-smithi

PR #9876, opened 2016-06-22
rgw: civetweb uses the same thread timeout as RGWProcess frontends

rzarzynski:


src/rgw/rgw_civetweb_frontend.cc
```diff
@@ -35,13 +35,19 @@ int RGWMongooseFrontend::run() {
   char thread_pool_buf[32];
   snprintf(thread_pool_buf, sizeof(thread_pool_buf), "%d",
 	   (int)g_conf->rgw_thread_pool_size);
+
+  char req_timeout_buf[32];
+  snprintf(req_timeout_buf, sizeof(req_timeout_buf), "%d",
+	   (int)g_conf->rgw_thread_pool_size * 1000);
```
cbodley: should this be `rgw_op_thread_timeout`?

rzarzynski: Yup, it should. I overlooked that while editing. I'm pretty sure you saved me several hours on tomorrow's testing and investigation why the setting doesn't work. Thanks a lot! :-)

rzarzynski:
Reworked and tested.

cbodley:
I don't think the WorkQueue timeouts work the same way that civetweb's do.. civetweb is using `request_timeout_ms` to set a timeout on the socket. But from what I can tell of WorkQueue, it just flags a worker as 'unhealthy' if a request hits this timeout. Nothing in rgw checks whether the workers are healthy.

Also note that this would change the timeout from civetweb's default of 30s to 600s, which doesn't sound right.

rzarzynski:
After reading the documentation for `rgw_op_thread_timeout` I was pretty sure that we are terminating the timed out threads. Now I see that RadoGW only prints a log:
```
2016-06-23 16:20:18.285123 7f806e040700  1 heartbeat_map is_healthy 'tp_rgw_process thread tp_rgw_process' had timed out after 5
```

Do we want to change the behavior for `rgw_op_thread_timeout` or [its documentation](http://docs.ceph.com/docs/master/radosgw/config-ref/#ceph-object-gateway-config-reference)?

cbodley:
I certainly don't think we should be killing threads. We might want to update the documentation to make it clear that the rgw_op_thread stuff only applies to the fcgi frontend.

Are you just trying to make civetweb consistent with RGWProcessFrontend, or is there another issue motivating this pr?

rzarzynski:
Yup, the sole motivation is to unify the behavior between frontends.

What is the reason for `rgw_op_thread_timeout`?

cbodley:
> What is the reason for `rgw_op_thread_timeout`?

Most of ceph's `WorkQueue`s get their timeout values from ceph.conf. I don't it's actually doing anything for rgw, aside from avoiding magic numbers in the code.

cbodley:
@rzarzynski looking at this again, I see that we're using these config variables for the `WorkQueue` in `RGWAsyncRadosProcessor` too, so it's not specific to the fcgi frontend. the documentation is a bit vague on these, but I'm not sure how to improve it. do you think there's anything more to do here?

rzarzynski:
Thanks for the info. I think we need to improve the document for the configurable. I've put this on my TODO list.

PR #9878, opened 2016-06-22
rbd: add explicit shrink check while resizing images

neurodrone:
Currently, if multiple clients issue resizes without coordination it is possible that the image undergoes downsizing due to a race. This might have undesired consequences depending on how strong or weak the client-level coordination works. As Ceph already provides atomic guarantees within its `ResizeRequest` state machine, we leverage that to guard against the unintended shrinking of image by exposing the `rbd_resize2()` API call.

/cc @dillaman 

Signed-off-by: Vaibhav Bhembre vaibhav@digitalocean.com

### DIFF
--- a/src/include/rbd/librbd.h
+++ b/src/include/rbd/librbd.h
@@ -272,6 +272,8 @@ CEPH_RBD_API int rbd_aio_open_read_only(rados_ioctx_t io, const char *name,
 CEPH_RBD_API int rbd_close(rbd_image_t image);
 CEPH_RBD_API int rbd_aio_close(rbd_image_t image, rbd_completion_t c);
 CEPH_RBD_API int rbd_resize(rbd_image_t image, uint64_t size);
+CEPH_RBD_API int rbd_resize2(rbd_image_t image, uint64_t size, bool allow_shrink,
+			     librbd_progress_fn_t cb, void *cbdata);
 CEPH_RBD_API int rbd_resize_with_progress(rbd_image_t image, uint64_t size,
 			     librbd_progress_fn_t cb, void *cbdata);
 CEPH_RBD_API int rbd_stat(rbd_image_t image, rbd_image_info_t *info,
--- a/src/include/rbd/librbd.hpp
+++ b/src/include/rbd/librbd.hpp
@@ -191,6 +191,7 @@ public:
   int aio_close(RBD::AioCompletion *c);
 
   int resize(uint64_t size);
+  int resize2(uint64_t size, bool allow_shrink, ProgressContext& pctx);
   int resize_with_progress(uint64_t size, ProgressContext& pctx);
   int stat(image_info_t &info, size_t infosize);
   int parent_info(std::string *parent_poolname, std::string *parent_name,
--- a/src/librbd/Operations.h
+++ b/src/librbd/Operations.h
@@ -39,8 +39,8 @@ public:
   int rename(const char *dstname);
   void execute_rename(const char *dstname, Context *on_finish);
 
-  int resize(uint64_t size, ProgressContext& prog_ctx);
-  void execute_resize(uint64_t size, ProgressContext &prog_ctx,
+  int resize(uint64_t size, bool allow_shrink, ProgressContext& prog_ctx);
+  void execute_resize(uint64_t size, bool allow_shrink, ProgressContext &prog_ctx,
                       Context *on_finish, uint64_t journal_op_tid);
 
   int snap_create(const char *snap_name);
--- a/src/librbd/WatchNotifyTypes.cc
+++ b/src/librbd/WatchNotifyTypes.cc
@@ -207,17 +207,23 @@ void AsyncCompletePayload::dump(Formatter *f) const {
 }
 
 void ResizePayload::encode(bufferlist &bl) const {
-  ::encode(size, bl);
   AsyncRequestPayloadBase::encode(bl);
+  ::encode(size, bl);
+  ::encode(allow_shrink, bl);
 }
 
 void ResizePayload::decode(__u8 version, bufferlist::iterator &iter) {
-  ::decode(size, iter);
   AsyncRequestPayloadBase::decode(version, iter);
+  ::decode(size, iter);
+
+  if (version >= 4) {
+    ::decode(allow_shrink, iter);
+  }
 }
 
 void ResizePayload::dump(Formatter *f) const {
   f->dump_unsigned("size", size);
+  f->dump_bool("allow_shrink", allow_shrink);
   AsyncRequestPayloadBase::dump(f);
 }
 
@@ -275,7 +281,7 @@ bool NotifyMessage::check_for_refresh() const {
 }
 
 void NotifyMessage::encode(bufferlist& bl) const {
-  ENCODE_START(3, 1, bl);
+  ENCODE_START(4, 1, bl);
   boost::apply_visitor(EncodePayloadVisitor(bl), payload);
   ENCODE_FINISH(bl);
 }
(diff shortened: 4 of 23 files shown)


dillaman:
@neurodrone Drop your first commit from the PR since it is removed in the second commit

src/include/rbd/librbd.h
```diff
@@ -272,9 +272,10 @@ CEPH_RBD_API int rbd_aio_open_read_only(rados_ioctx_t io, const char *name,
 CEPH_RBD_API int rbd_close(rbd_image_t image);
 CEPH_RBD_API int rbd_aio_close(rbd_image_t image, rbd_completion_t c);
 CEPH_RBD_API int rbd_resize(rbd_image_t image, uint64_t size);
+CEPH_RBD_API int rbd_resize2(rbd_image_t image, uint64_t size, bool can_shrink,
```
dillaman: Minor: ```can_shrink``` -> ```allow_shrink```

src/include/rbd/librbd.hpp
```diff
@@ -191,8 +191,8 @@ class CEPH_RBD_API Image
   int aio_close(RBD::AioCompletion *c);
 
   int resize(uint64_t size);
+  int resize2(uint64_t size, bool can_shrink, ProgressContext& pctx);
```
dillaman: Minor: ```can_shrink``` -> ```allow_shrink```

src/librbd/WatchNotifyTypes.h
```diff
@@ -200,11 +200,12 @@ struct ResizePayload : public AsyncRequestPayloadBase {
...
 
-  ResizePayload() : size(0) {}
-  ResizePayload(uint64_t size_, const AsyncRequestId &id)
-    : AsyncRequestPayloadBase(id), size(size_) {}
+  ResizePayload() : size(0), can_shrink(true) {}
+  ResizePayload(uint64_t size_, bool can_shrink_, const AsyncRequestId &id)
+    : AsyncRequestPayloadBase(id), size(size_), can_shrink(can_shrink_) {}
 
   uint64_t size;
+  bool can_shrink;
```
dillaman: Minor: ```can_shrink``` -> ```allow_shrink```

dillaman: You need to modify the ResizePayload ```encode```, ```decode```, and ```dump``` methods to handle the new param. You will need to up the version from 2 -> 3 and only attempt to decode the ```allow_shrink``` param if version >= 3 (see ```AcquiredLockPayload::decode``` for an example)

src/librbd/operation/ResizeRequest.h
```diff
@@ -18,14 +18,15 @@ template <typename ImageCtxT = ImageCtx>
...
+                               uint64_t new_size, bool can_shrink,
+                               ProgressContext &prog_ctx, uint64_t journal_op_tid,
+                               bool disable_journal) {
+    return new ResizeRequest(image_ctx, on_finish, new_size, can_shrink, prog_ctx,
                              journal_op_tid, disable_journal);
   }
 
   ResizeRequest(ImageCtxT &image_ctx, Context *on_finish, uint64_t new_size,
-                ProgressContext &prog_ctx, uint64_t journal_op_tid,
+                bool can_shrink, ProgressContext &prog_ctx, uint64_t journal_op_tid,
```
dillaman: Minor: ```can_shrink``` -> ```allow_shrink```

src/librbd/operation/ResizeRequest.cc
```diff
@@ -220,7 +220,7 @@ Context *ResizeRequest<I>::send_grow_object_map() {
     }
     return this->create_context_finisher();
   } else if (m_new_size < m_original_size) {
-    if (!image_ctx.allow_shrink) {
+    if (!m_can_shrink) {
       ldout(cct, 1) << "cannot shrink since 'allow_shrink' is set to false" << dendl;
```
dillaman: Minor: "shrinking the image is not permitted"

src/librbd/Operations.h
```diff
@@ -39,8 +39,8 @@ class Operations {
   int rename(const char *dstname);
   void execute_rename(const char *dstname, Context *on_finish);
 
-  int resize(uint64_t size, ProgressContext& prog_ctx);
-  void execute_resize(uint64_t size, ProgressContext &prog_ctx,
+  int resize(uint64_t size, bool can_shrink, ProgressContext& prog_ctx);
```
dillaman: Minor: ```can_shrink``` -> ```allow_shrink```

src/librbd/WatchNotifyTypes.h
neurodrone: Updated the inherited methods. Should I bump up the version in this file or some other location?

src/librbd/WatchNotifyTypes.cc
```diff
@@ -208,16 +208,22 @@ void AsyncCompletePayload::dump(Formatter *f) const {
...
 void ResizePayload::encode(bufferlist &bl) const {
   ::encode(size, bl);
+  ::encode(allow_shrink, bl);
   AsyncRequestPayloadBase::encode(bl);
 }
 
 void ResizePayload::decode(__u8 version, bufferlist::iterator &iter) {
   ::decode(size, iter);
+
+  if (version >= 3) {
```
dillaman: Correction -- should be version 4 and update the encoded version here: https://github.com/neurodrone/ceph/blob/950db41a78dac0d07d25436e677d4e5e7b6adae8/src/librbd/WatchNotifyTypes.cc#L284

neurodrone: I see. Good to know! Updated, please take a look.

dillaman:
@neurodrone This requires a rebase against the current state of the master branch

neurodrone:
Sure @dillaman. I am currently AWOL with dismal access, but I am back on Thurs so will do so then.

neurodrone:
@dillaman Rebased against the latest master. LMK what you think.

dillaman:
@neurodrone Unfortunately, it requires an additional rebase

neurodrone:
@dillaman No worries. Updated with the latest.

src/librbd/WatchNotifyTypes.cc
```diff
@@ -208,16 +208,22 @@ void AsyncCompletePayload::dump(Formatter *f) const {
 
 void ResizePayload::encode(bufferlist &bl) const {
   ::encode(size, bl);
+  ::encode(allow_shrink, bl);
   AsyncRequestPayloadBase::encode(bl);
 }
 
 void ResizePayload::decode(__u8 version, bufferlist::iterator &iter) {
   ::decode(size, iter);
+
```
dillaman: Since this is a protocol change, it actually needs to be the last thing decoded. Therefore, move to after ```AsyncRequestPayloadBase::decode``

src/librbd/operation/ResizeRequest.cc
```diff
@@ -212,12 +212,18 @@ Context *ResizeRequest<I>::send_grow_object_map() {
...
   image_ctx.aio_work_queue->unblock_writes();
 
+  CephContext *cct = image_ctx.cct;
   if (m_original_size == m_new_size) {
     if (!m_disable_journal) {
       this->commit_op_event(0);
     }
     return this->create_context_finisher();
   } else if (m_new_size < m_original_size) {
+    if (!m_allow_shrink) {
```
dillaman: This check needs to be performed before the journal event is recorded within the ```send_append_op_event``` method. Once the journal event is recorded, bailing out is considered a split-brain event.

dillaman:
@neurodrone Looking very good -- just a couple of minor issues.

neurodrone:
@dillaman Thanks for your comments and explanation. Updated with the latest.

dillaman:
@neurodrone Thanks -- super minor (and sorry I missed it the first time) but can you also add a test case to test_librbd.cc to exercise a shrink with the allow_shrink set to false and true? In the meantime, I'll run this through the full integration test suite.

neurodrone:
@dillaman Absolutely! I was working on that actually. I will test and send up the addition.

neurodrone:
Added test and rebased against master.

PR #9879, opened 2016-06-22
cmake: fix linking issues

alimaredia:


alimaredia:
@liewegas this fixes any compile issues I ran into.

dmick:
it'd be great if the commit messages were a little more context-free.  for example, what linking problems were caused by the visibility flags, and why is removing them ok, given that we once thought they were necessary?  This is especially critical when there aren't underlying bugs that discuss the issues and tradeoffs

tchaikov:
```
CMakeFiles/ceph_test_rados_striper_api_aio.dir/aio.cc.o: In function `StriperTest_SimpleWrite_Test::TestBody()':
/srv/autobuild-ceph/gitbuilder.git/build/src/test/libradosstriper/aio.cc:53: undefined reference to `rados_striper_aio_write'
```

see http://gitbuilder.sepia.ceph.com/gitbuilder-ceph-tarball-trusty-amd64-cmake/log.cgi?log=48d18030fdbb8243b63378a6782ee8f6ec17fa59 . because 

because libradosstriper does not put the `__attribute__ ((visibility ("default")))` specifier on the exported class/methods. we should not add the visibility to this library.

tchaikov:
#9887 superses this.

PR #9881, opened 2016-06-22
librbd: journal::Replay no longer holds lock while completing callback

dillaman:
Fixes: http://tracker.ceph.com/issues/16433


trociny:
lgtm

PR #9882, opened 2016-06-22
librbd: fix crash while using advisory locks with R/O image

dillaman:
Fixes: http://tracker.ceph.com/issues/16364


trociny:
lgtm

PR #9884, opened 2016-06-23
mds: fix shutting down mds timed-out due to deadlock

david-z:
Fixes: #16396

This PR contains 2 fixes:

* Fix shutting down mds timed-out due to deadlock

The root cause is ms_handle_reset may be called within mds_lock via 'MDSDaemon::handle_signal' -> 'MDSDaemon::suicide' -> 'MDSRankDispatcher::shutdown' -> 'AsyncMessenger::shutdown' -> 'AsyncMessenger::mark_down_all' -> ... -> 'AsyncConnection::stop' -> dispatch 'C_handle_reset' event -> 'Messenger::ms_deliver_handle_reset' in 2 different threads.

* Remove the unnecessary checking to wakup event_wait, discussed with @yuyuyu101 in https://github.com/ceph/ceph/pull/9841

Signed-off-by: Zhi Zhang zhangz.david@outlook.com

### DIFF
--- a/src/mds/MDSRank.cc
+++ b/src/mds/MDSRank.cc
@@ -238,7 +238,11 @@ void MDSRankDispatcher::shutdown()
   progress_thread.shutdown();
 
   // shut down messenger
+  // release mds_lock first because messenger thread might call 
+  // MDSDaemon::ms_handle_reset which will try to hold mds_lock
+  mds_lock.Unlock();
   messenger->shutdown();
+  mds_lock.Lock();
 
   // Workaround unclean shutdown: HeartbeatMap will assert if
   // worker is not removed (as we do in ~MDS), but ~MDS is not
--- a/src/msg/async/Event.cc
+++ b/src/msg/async/Event.cc
@@ -262,15 +262,15 @@ void EventCenter::delete_time_event(uint64_t id)
 
 void EventCenter::wakeup()
 {
-  if (already_wakeup.compare_and_swap(0, 1)) {
     ldout(cct, 1) << __func__ << dendl;
+    already_wakeup.compare_and_swap(0, 1);
+
     char buf[1];
     buf[0] = 'c';
     // wake up "event_wait"
     int n = write(notify_send_fd, buf, 1);
     // FIXME ?
     assert(n == 1);
-  }
 }
 
 int EventCenter::process_time_events()


david-z:
@yuyuyu101 Could you pls take a look and help to review? Thanks.

The issue of restarting mds described in #16396 is due to a deadlock, not I mentioned in https://github.com/ceph/ceph/pull/9841 before.

yuyuyu101:
lgtm

src/mds/MDSDaemon.cc
```diff
@@ -1245,24 +1245,40 @@ bool MDSDaemon::ms_handle_reset(Connection *con)
...
 
-  Mutex::Locker l(mds_lock);
-  if (stopping) {
-    return false;
+  // ms_handle_reset may be called within mds_lock 
+  // via 'MDSDaemon::handle_signal' -> 'MDSDaemon::suicide' 
+  // -> 'MDSRankDispatcher::shutdown' -> 'AsyncMessenger::shutdown' 
+  // -> 'AsyncMessenger::mark_down_all' -> ... -> 'AsyncConnection::stop' 
+  // -> dispatch 'C_handle_reset' event -> 'Messenger::ms_deliver_handle_reset'
+  const bool is_locked = mds_lock.is_locked();
```
jcsp: This should be is_locked_by_me, right?

david-z: @jcsp No, we can't simply use mds_lock.is_locked_by_me() here because one mds thread will hold mds_lock in MDSDaemon::handle_signal(), later it will dispatch an async event 'C_handle_reset' in AsyncConnection::stop() to another msgr thread which will call MDSDaemon::ms_handle_reset().

david-z:
@jcsp I reviewed this part's code logic again and I think it is not always safe to use mds_lock.is_locked here because async msgr might also call MDSDaemon::ms_handle_reset() directly if some faults happen.

Since a 'stopping' flag will be set to true in shutdown within mds_lock and each msgr operation to mds will try to hold mds_lock first and then check this 'stopping' flag, I think we could release mds_lock before shutting down msgr. If there are some other msgr operations to mds happens when we shutdown mds, mds_lock and 'stopping' flag can indicate this situation.

What do you think? Thanks.

jcsp:
Yes, everything that takes mds_lock should check stopping, so this should be safe. Will pull this onto next test run.

jcsp:
(passed http://pulpito.ceph.com/jspray-2016-06-28_01:46:16-fs-wip-jcsp-testing-20160627-distro-basic-mira)

PR #9885, opened 2016-06-23
udev: always populate /dev/disk/by-parttypeuuid

dachary:
ceph-disk activate-all walks /dev/disk/by-parttypeuuid at boot time. It
is not necessary when udev fires ADD event for each partition and
95-ceph-osd.rules gets a chance to activate a ceph disk or journal.

There are various reasons why udev ADD events may not be fired at
boot (for instance Debian Jessi 8.5 never does it and CentOS 7.2 seems
to be racy in that regard when a LVM root is being used).

Populating /dev/disk/by-parttypeuuid fixes ceph-disk activate-all that
would not work without it. And it guarantees disks are activated at boot
time regardless of wether udev fires ADD events at the right time (or at
all).

The new udev file is a partial resurection of the
60-ceph-partuuid-workaround-rules that was removed by
9f77244b8e0782921663e52005b725cca58a8753. It is given a name that
reflects its new purpose.

Fixes http://tracker.ceph.com/issues/16351

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -15,6 +15,7 @@ EXTRA_DIST += \
 	etc/sysconfig/SuSEfirewall2.d/services/ceph-mon \
 	etc/sysconfig/SuSEfirewall2.d/services/ceph-osd-mds \
 	udev/50-rbd.rules \
+	udev/60-ceph-by-parttypeuuid.rules \
 	udev/95-ceph-osd.rules \
 	share/known_hosts_drop.ceph.com \
 	share/id_dsa_drop.ceph.com \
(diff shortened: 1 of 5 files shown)


dachary:
@aderumier does this look sensible to you ? I think it does since that's what you suggested in the first place. Only now I better understand why this is the right solution.

aderumier:
@dachary : I'll test it today or tomorrow. Look fine.

sep76:
@dachary : worked on my debian jessie running   jewel 10.2.2-1~bpo80+1 on backports kernel 4.6.0-0.bpo.1-amd64
all osd's start on reboot now.  thanks :)

aderumier:
Sorry to be late, this is working fine for me too !

PR #9893, opened 2016-06-23
hammer: OSDMonitor: drop pg temps from not the current primary

Vicente-Cheng:
http://tracker.ceph.com/issues/16430

### DIFF
--- a/src/mon/OSDMonitor.cc
+++ b/src/mon/OSDMonitor.cc
@@ -2028,6 +2028,21 @@ bool OSDMonitor::preprocess_pgtemp(MOSDPGTemp *m)
       continue;
     }
 
+    int acting_primary = -1;
+    osdmap.pg_to_up_acting_osds(
+      p->first, NULL, NULL, NULL, &acting_primary);
+    if (acting_primary != from) {
+      /* If the source isn't the primary based on the current osdmap, we know
+       * that the interval changed and that we can discard this message.
+       * Indeed, we must do so to avoid 16127 since we can't otherwise determine
+       * which of two pg temp mappings on the same pg is more recent.
+       */
+      dout(10) << __func__ << " ignore " << p->first << " -> " << p->second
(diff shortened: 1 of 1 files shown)


dachary:
does this compile if you just replace **nullptr** with **NULL** ?

Vicente-Cheng:
@dachary 
hi loicd, 
I use static_cast to simulate nullptr, it seems OK on my machine.

dachary:
make check does not have a problem with it but I don't think it exercises this specific part of the code

Vicente-Cheng:
@dachary 
hi loicd, I will still check other reference about how to handle nullptr.
Could I new a `vector<int>` to null like vector<int>() , and pass it?
please see the latest commit.
how do you think?

smithfarm:
This is the function declaration:

```
void OSDMap::_pg_to_up_acting_osds(const pg_t& pg, vector<int> *up, int *up_primary,
                                   vector<int> *acting, int *acting_primary) const
```

I'm not sure we need to use `new` here -> that creates a structure in memory (right?), which the original code doesn't do. In hammer we could just use "0" or "NULL" in place of `nullptr`.

Vicente-Cheng:
@smithfarm 
hi Nathan,
OK! I following your suggestion and re-push it.

Actually, I still confuse with NULL and nullptr :(
we could just use value(NULL) to replace the pointer(nullptr) ?

thanks!!

smithfarm:
@Vicente-Cheng My understanding is that NULL is used in C, and 0 is used in C++. See https://en.wikipedia.org/wiki/Null_pointer#cite_note-4

smithfarm:
Note that the "null pointer" does not actually point to anything.

Vicente-Cheng:
@smithfarm 
OK! Nathan,
thanks for your information : )

that I remove the DNM tag

smithfarm:

OK to merge?

athanatos:
lgtm

PR #9894, opened 2016-06-23
osd: limit omap data in push op

gaowanlong:
build_push_op assumes 8MB of omap entries is about as much work to read
as 8MB of object data. This is probably false. Add a config
osd_recovery_max_omap_entries_per_chunk with a sane default (128k).

Signed-off-by: Wanlong Gao <wanlong.gao@easystack.cn>

http://tracker.ceph.com/issues/16128

### DIFF
--- a/src/common/config_opts.h
+++ b/src/common/config_opts.h
@@ -752,6 +752,7 @@ OPTION(osd_recovery_delay_start, OPT_FLOAT, 0)
 OPTION(osd_recovery_max_active, OPT_U64, 3)
 OPTION(osd_recovery_max_single_start, OPT_U64, 1)
 OPTION(osd_recovery_max_chunk, OPT_U64, 8<<20)  // max size of push chunk
+OPTION(osd_recovery_max_omap_entries_per_chunk, OPT_U64, 64000) // max number of omap entries per chunk; 0 to disable limit
 OPTION(osd_copyfrom_max_chunk, OPT_U64, 8<<20)   // max size of a COPYFROM chunk
 OPTION(osd_push_per_object_cost, OPT_U64, 1000)  // push cost per object
 OPTION(osd_max_push_cost, OPT_U64, 8<<20)  // max size of push message
--- a/src/osd/ReplicatedBackend.cc
+++ b/src/osd/ReplicatedBackend.cc
@@ -2000,7 +2000,9 @@ int ReplicatedBackend::build_push_op(const ObjectRecoveryInfo &recovery_info,
 	 iter->valid();
 	 iter->next(false)) {
       if (!out_op->omap_entries.empty() &&
-	  available <= (iter->key().size() + iter->value().length()))
+	  ((cct->_conf->osd_recovery_max_omap_entries_per_chunk > 0 &&
+	    out_op->omap_entries.size() >= cct->_conf->osd_recovery_max_omap_entries_per_chunk) ||
+	   available <= iter->key().size() + iter->value().length()))
 	break;
       out_op->omap_entries.insert(make_pair(iter->key(), iter->value()));
 


src/common/config_opts.h
```diff
@@ -751,6 +751,7 @@ OPTION(osd_recovery_delay_start, OPT_FLOAT, 0)
 OPTION(osd_recovery_max_active, OPT_U64, 3)
 OPTION(osd_recovery_max_single_start, OPT_U64, 1)
 OPTION(osd_recovery_max_chunk, OPT_U64, 8<<20)  // max size of push chunk
+OPTION(osd_recovery_max_omap_entries_per_chunk, OPT_U64, 50<<10) // max size of omap entries per chunk
```
liewegas: 50 k seems small to me.. maybe 256 or 512 k?

gaowanlong: @liewegas Yeah, updated to 512K. ;)

athanatos:
512k seems a bit high...  How long does that take to read on a spinning disk?

gaowanlong:
@athanatos @liewegas  Then, what do you think about 256K?

liewegas:
128k or 256k sound ok to me

athanatos:
Let's call it 128k.

gaowanlong:
Good to get a consensus, now updated to 128K.

athanatos:
We also need a ceph-qa-suite branch modifying this config for at least one replicated rados task tests to something like 5 so we can be sure the code will actually be exercised.  Probably suites/rados/thrash/workloads/small_objects.yaml .

athanatos:
https://github.com/athanatos/ceph-qa-suite/tree/wip-osd_recovery_max_omap_entries_per_chunk should do it.

src/osd/ReplicatedBackend.cc
```diff
@@ -1981,20 +1981,25 @@ int ReplicatedBackend::build_push_op(const ObjectRecoveryInfo &recovery_info,
   }
 
   uint64_t available = cct->_conf->osd_recovery_max_chunk;
+  uint64_t omap_available = cct->_conf->osd_recovery_max_omap_entries_per_chunk;
   if (!progress.omap_complete) {
+    uint64_t size;
```
athanatos: Why introduce this uninitialized here?  Move this declaration into the for loop where it is initialized.

athanatos:
Wait, is the intent to limit the *number* of omap entries, or the *total size* of omap entries.  Based on the name of the config, I assumed that you meant to allow up to the smaller of 128000 individual omap entries independent of size *or* osd_recovery_max_chunk total size.  I think that approach makes far more sense.  I'd prefer that you change the config to limit purely the number of entries, we already have a config which limits the total size.

In light of that, does 128000 make sense?  Maybe it should be more like 4000.

athanatos:
@gaowanlong @liewegas

src/osd/ReplicatedBackend.cc
gaowanlong: @athanatos yeah, thank you, will fix in next version.

gaowanlong:
@athanatos @liewegas Now I moved to limit the number of omap entries instead, default to 4K, what do you think? ;)

src/osd/ReplicatedBackend.cc
```diff
@@ -1988,13 +1988,15 @@ int ReplicatedBackend::build_push_op(const ObjectRecoveryInfo &recovery_info,
     for (iter->lower_bound(progress.omap_recovered_to);
 	 iter->valid();
 	 iter->next(false)) {
+      uint64_t size = iter->key().size() + iter->value().length();
```
athanatos: This is an unnecessary refactor, I don't think it should be included in this branch.

athanatos:
@gaowanlong I don't want extra code changes mixed in, you should probably just remove the size change.  Also, have we actually tried to measure how the system behaves with that value?  I'd rather not end up with a totally random config...  Could you set up a test pool, populate it with 1000 objects with 64000 omap entries each, and add an osd with and without this config? 

@gaowanlong Also, you need to arrange for a value of 0 to disable this config.

gaowanlong:
@athanatos Fixed as your comments,  and I'm setting up the test pool and filling omap entries with ceph_omapbench tool.

gaowanlong:
@athanatos Without that value, we can see the push omap entries is 64000 from log:
`send_pushes: sending push PushOp(0:44b8b797:::admin.obj.2:head, version: 12'1, data_included: [], data_size: 0, omap_header_size: 0, omap_entries_size: 64000, attrset_size: 2,`

And with the default value of this config, we can see the push omap entries is 4096 from log:
`send_pushes: sending push PushOp(0:ecaaaeb2:::admin.obj.240:head, version: 29'43, data_included: [], data_size: 0, omap_header_size: 0, omap_entries_size: 4096, attrset_size: 2,`

I think it act as we expected, what do you think?

athanatos:
@gaowanlong  Ok, how long did it take to build the PushOp?  I was hoping you'd get data indicating what a good default would be.  We don't want it to take too long to build the PushOp, but we also don't want to send more messages than we have to.

gaowanlong:
@athanatos My local test result is that, the avg time of no limit (say 64000 entries)  is 1.570614969s and the avg time of limit (say 4K entries) is 0.192918099s.  But to be honest, my 4 osds are on the same hdd,  so that the time is not that convincing.   What's the most acceptable avg time in your heart? I can do more test and find the best default config when I setup more osds on individual hdds.

athanatos:
I'd say around 200-300ms would be reasonable.

gaowanlong:
@athanatos I tested on the 10000rpm SAS, the result is:
`
4096    avg = 15.996865 ms
51200  avg = 125.092096 ms
64000  avg = 235.814020 ms
`
So, the limit of 64000 entries is more reasonable as you said.

athanatos:
Added needs-qa.  Make sure to test against a version of ceph-qa-suite with https://github.com/athanatos/ceph-qa-suite/tree/wip-osd_recovery_max_omap_entries_per_chunk merged in (it's based on an old master, so at least create a new branch based on master and merge it in).

yuriw:
ceph-qa-suite branch `wip-yuri-testing__2016_7_18_qa_suites`

yuriw:

PR #9895, opened 2016-06-23
librbd: ignore snap unprotect -EBUSY errors during journal replay

dillaman:
Fixes: http://tracker.ceph.com/issues/16445

### DIFF
--- a/src/librbd/journal/Replay.cc
+++ b/src/librbd/journal/Replay.cc
@@ -112,24 +112,29 @@ struct C_RefreshIfRequired : public Context {
   C_RefreshIfRequired(I &image_ctx, Context *on_finish)
     : image_ctx(image_ctx), on_finish(on_finish) {
   }
+  virtual ~C_RefreshIfRequired() {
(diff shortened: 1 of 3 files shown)


src/librbd/journal/Replay.cc
```diff
@@ -112,6 +112,15 @@ struct C_RefreshIfRequired : public Context {
...
   }
+  virtual ~C_RefreshIfRequired() {
+    delete on_finish;
+  }
+
+  virtual void complete(int r) override {
+    finish(r);
+    on_finish = nullptr;
+    delete this;
+  }
```
trociny: @dillaman It is just my opinion you can safely ignore. Anyway, I don't like very much when `Context::complete()` is overridden. I consider `Context::complete` method as a part of the interface. And this can easily be avoided here by resetting `on_finish` in `finish()` (and this might look more readable to me).

trociny:
retest this please

trociny:
lgtm

PR #9898, opened 2016-06-23
os/bluestore/BlueFS: do not hold internal lock while waiting for IO

liewegas:
I don't think we should merge this until unittest_bluefs gets some tests that
do parallel threads writing to the same and different files and doing fsync.

### DIFF
--- a/src/os/bluestore/BlueFS.cc
+++ b/src/os/bluestore/BlueFS.cc
@@ -1110,7 +1110,12 @@ int BlueFS::_flush_range(FileWriter *h, uint64_t offset, uint64_t length)
   }
   if (h->file->fnode.size < offset + length) {
     h->file->fnode.size = offset + length;
-    must_dirty = true;
+    if (h->file->fnode.ino != 1) {
+      // we do not need to dirty the log file when the file size
+      // changes because replay is smart enough to discover it on its
+      // own.
+      must_dirty = true;
+    }
   }
   if (must_dirty) {
     h->file->fnode.mtime = ceph_clock_now(NULL);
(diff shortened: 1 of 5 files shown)


src/os/bluestore/BlueFS.cc
```diff
@@ -1239,17 +1270,23 @@ void BlueFS::_fsync(FileWriter *h)
...
   _flush(h, true);
-  _flush_wait(h);
-  if (h->file->dirty) {
-    dout(20) << __func__ << " file metadata is dirty, flushing log on "
-	     << h->file->fnode << dendl;
-    _flush_log();
-    assert(!h->file->dirty);
+  uint64_t old_dirty_seq = h->file->dirty_seq;
+  lock.unlock();
+  wait_for_aio(h);
```
varadakari: There is a problem here, Thread T1 will finish _flush(updates log file) and waits for the IO to complete. Thread T2 can grab the lock, and do _flush(update the log file), releases the lock. T1 will grab and flushes the log file contains T1 contents + T2 contents. If T2 IO fails now, we will have log containing T2 updates.

liewegas: Yep.  Updated this to only allow one concurrent log flush at a time.  It now passes your first test, but I'm unsure if we want to do something more complex here or not...

varadakari: Right now we have the following test cases,
Test case 1:
1. One writer for single(keeps appending to the file).
2. Couple of threads, to generate more files(new).
Both threads keep doing fsync from both. 

Test case 2:
Have multiple threads to generate new files. will keep doing fsync.


Will add the following

Test case 3: 
Have couple of threads to do append only or create new files and have a sync_metadata thread.
This will generate log compaction once the threshould is hit.

If you have something in mind i can add them. This covers most of the uses cases coming to bluefs.

PR #9900, opened 2016-06-23
[DNM, proposal] rgw: fix the handling of rgw_swift_url_prefix.

rzarzynski:
This patch fixes to the support for placing the Swift API in the root of URL hierarchy. Unfortunately, the whole concept exhibits a severe side effect: inability to deploy RadosGW in multi-site configuration.

The sole reason behind this fix is the fact we claimed in documentation that the feature is available.

-- 

This PR is a proposal of quick but very limited fix for the issue we faced in PR #9582. I don't believe it's ready for merge. I sent this patch rather to ignite a discussion.

CC: @cbodley, @pritha-srivastava.

### DIFF
--- a/src/rgw/rgw_main.cc
+++ b/src/rgw/rgw_main.cc
@@ -342,13 +342,32 @@ int main(int argc, const char **argv)
   }
 
   // S3 website mode is a specialization of S3
-  bool s3website_enabled = apis_map.count("s3website") > 0;
-  if (apis_map.count("s3") > 0 || s3website_enabled)
-    rest.register_default_mgr(set_logging(new RGWRESTMgr_S3(s3website_enabled)));
+  const bool s3website_enabled = apis_map.count("s3website") > 0;
+  // Swift API entrypoint could placed in the root instead of S3
+  const bool swift_at_root = g_conf->rgw_swift_url_prefix == "/";
+  if (apis_map.count("s3") > 0 || s3website_enabled) {
+    if (! swift_at_root) {
+      rest.register_default_mgr(set_logging(new RGWRESTMgr_S3(s3website_enabled)));
+    } else {
+      derr << "Cannot have the S3 or S3 Website enabled together with "
+           << "Swift API placed in the root of hierarchy" << dendl;
+      return EINVAL;
+    }
+  }
 
   if (apis_map.count("swift") > 0) {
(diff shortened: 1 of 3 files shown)


src/rgw/rgw_main.cc
```diff
@@ -336,13 +336,26 @@ int main(int argc, const char **argv)
...
+    }
+  }
 
   if (apis_map.count("swift") > 0) {
-    rest.register_resource(g_conf->rgw_swift_url_prefix,
-			   set_logging(new RGWRESTMgr_SWIFT));
+    if (! swift_at_root) {
+      rest.register_resource(g_conf->rgw_swift_url_prefix,
+                             set_logging(new RGWRESTMgr_SWIFT));
+    } else {
```
cbodley: we could error out here if we detect a multisite configuration.. something like this?
```c++
    } else if (store->get_zonegroup()->zones.size() > 1) {
```

rzarzynski: Makes sense.

src/rgw/rgw_main.cc
```diff
@@ -336,13 +336,30 @@ int main(int argc, const char **argv)
...
   if (apis_map.count("swift") > 0) {
-    rest.register_resource(g_conf->rgw_swift_url_prefix,
-			   set_logging(new RGWRESTMgr_SWIFT));
+    if (! swift_at_root) {
+      rest.register_resource(g_conf->rgw_swift_url_prefix,
+                             set_logging(new RGWRESTMgr_SWIFT));
+    } else if (store->get_zonegroup().zones.size() > 1) {
+      derr << "Swift API cannot be placed in the root of hierarchy "
+           << " in a multi-zone deployments" << dendl;
+      return EINVAL;
```
cbodley: I guess I've changed my mind on this part. If we want to support running separate gateways for swift and s3/sync, we shouldn't fail here.. just print a warning that you'll have to run another gateway to handle sync

yehudasa: @rzarzynski maybe it shouldn't be even related to multisite? Just warn if the s3 api is also enabled?

rzarzynski: @yehudasa: I've just pushed a version showing a warning (but not failing) in case of Swift-at-root and multi-site deployment.

rzarzynski:
Bundled in #10280. Closing.

PR #9904, opened 2016-06-23
cmake: add radosgw to 'make vstart' target

cbodley:
Now that all unit tests build by default in cmake (see 824931368fdd99ae20fe30094f8abce4045303f4), developers may want to use 'make vstart' to build the minimal set of targets needed to start a cluster with vstart.sh. This adds the radosgw and radosgw-admin targets to support this workflow for rgw testing.


liewegas:
:+1:

PR #9905, opened 2016-06-23
librbd: mark exclusive lock as released after journal is closed

dillaman:
Fixes: http://tracker.ceph.com/issues/16450


trociny:
lgtm

PR #9906, opened 2016-06-23
msg/AsyncMessenger:delete cpy constr and assign op

stiopaa1:
Delete copy constructor and assignment operator, so the checking can
be moved from linking to compile time - as it was when the above
where only declared, but not defined.


tchaikov:
lgtm.

PR #9914, opened 2016-06-24
[DNM] s/boost::regex/std::regex/

tchaikov:
to use the `std::regex` instead of the one from boost. always good to have less dependency if everything else is the same.


tchaikov:
@mattbenjamin does this change look good to you?

tchaikov:
seems gcc in trusty does not fine with the "std::regex", will test more.

alimaredia:
@tchaikov Matt is out until Tuesday

PR #9915, opened 2016-06-24
rgw: Corrected the logic for container stats calculation.

pritha-srivastava:
The values are being added twice resulting in incorrect stats.
Fixes http://tracker.ceph.com/issues/16188

### DIFF
--- a/src/rgw/rgw_rados.cc
+++ b/src/rgw/rgw_rados.cc
@@ -10853,9 +10853,9 @@ int RGWRados::update_containers_stats(map<string, RGWBucketEnt>& m)
       map<uint8_t, struct rgw_bucket_category_stats>::iterator iter = (hiter->second.stats).find((uint8_t)category);
       if (iter != hiter->second.stats.end()) {
         struct rgw_bucket_category_stats& stats = iter->second;
-        ent.count += stats.num_entries;
-        ent.size += stats.total_size;
-        ent.size_rounded += stats.total_size_rounded;
+        ent.count = stats.num_entries;
+        ent.size = stats.total_size;
+        ent.size_rounded = stats.total_size_rounded;
       }
     }
   }


src/rgw/rgw_rados.cc
```diff
@@ -10853,9 +10853,9 @@ int RGWRados::update_containers_stats(map<string, RGWBucketEnt>& m)
       map<uint8_t, struct rgw_bucket_category_stats>::iterator iter = (hiter->second.stats).find((uint8_t)category);
       if (iter != hiter->second.stats.end()) {
         struct rgw_bucket_category_stats& stats = iter->second;
-        ent.count += stats.num_entries;
-        ent.size += stats.total_size;
-        ent.size_rounded += stats.total_size_rounded;
+        ent.count = stats.num_entries;
+        ent.size = stats.total_size;
+        ent.size_rounded = stats.total_size_rounded;
```
cbodley: i see that this loop is iterating over a `map<string, rgw_bucket_dir_header>`.. and instead of summing up the stats for each, you're only taking stats from the last one. i'm not clear on what these different dir_headers correspond to; can you shed any light on that?

pritha-srivastava:
@cbodley: As far as I can see, the headers contains stats about the bucket, like the total size, total count of objects etc. I am not sure in which case the header for a bucket will have more than 1 header. Based on the assumption that the header returned will be 1 per bucket, I replaced the summation with assignment. I can leave the summation as is, and initialize total size, total count etc to zero for each bucket, just before the beginning of the loop, so that if we ever have more than 1 header, all the values will be taken into account.
Let me know what you think.

pritha-srivastava:
The problem that I am trying to address here is that: the stats are counted twice - once after cls_user_list_buckets is invoked, the bucket stats  gets updated by a BucketsSyncThread thread after some x minutes. After that,  the call to update_container_stats also fetches the same data, and because we have a summation in the loop, the values get counted twice. I think fetching the stats should be done in update_container_stats because 1. I think that is what the code is meant to do 2. Changing here will take care of 'stat' and 'list' methods together. 3. The solution to use 'false' to suppress the call to update_container_stats will not show bucket stats till the sync thread syncs the data.

cbodley:
@pritha-srivastava can you confirm whether there are more than one `rgw_bucket_dir_header` or `rgw_bucket_category_stats` when these double stats are reported?

pritha-srivastava:
@cbodley : I debugged by creating 2 buckets, each containing 2 objects and I saw one header per bucket. Is there any other scenario that you would want me to try out?

cbodley:
it sounds like this only reproduces once the `BucketsSyncThread` calls through `RGWRados::cls_user_sync_bucket_stats()`, which has logic very similar to `update_container_stats()`. maybe trace through that and see if the stats are correct at that point?

pritha-srivastava:
@cbodley: The stats fetched by BucketsSyncThread are correct, but only after the sync happens. And, the sync happens periodically. So before the sync happens, the stats returned by update_container_stats() shows the correct stats. Once the sync happens, the stats gets doubled up in the output of the 'stat' command, and yes so far I have seen only 1 header per bucket. So the correct thing to do would be to reset the total size, count etc to zero, before updating them with the values fetched from the headers.

cbodley:
aren't the `RGWBucketEnt`s that are passed into `update_container_stats()` already zeroed?

pritha-srivastava:
@cbodley : No, they are not zeroed. They are updated as a result of the call cls_user_list_buckets.

cbodley:
@pritha-srivastava thanks for helping me understand the code paths involved. i see that `rgw_read_user_buckets()` is initializing the `RGWBucketEnt` stats with data from `cls_user_list_buckets()`, and then `update_containers_stats()` is adding to those existing stats

i agree that `update_containers_stats()` should zero the count/size/size_rounded fields before looping through the headers

chiehtu:
@cbodley Here you are. #9584

yehudasa:
@cbodley ping?

cbodley:
#9584 was merged for this. closing

PR #9917, opened 2016-06-24
jewel: ceph Resource Agent does not work with systemd

smithfarm:
http://tracker.ceph.com/issues/16461


dachary:

PR #9918, opened 2016-06-24
os/kstore: kill kstore

xiexingguo:
As we have decided to devote on bluestore, it will be a burden
to keep an eye on this deprecated condidate module. And removing
the unnecessary code is always the preferred option.


liewegas:
I think we should keep kstore around as a reference. We may need it to layer on top of new kv devices, or possibly the kinetic API.

PR #9922, opened 2016-06-24
librbd: do not propagate mirror status notification failures

dillaman:
These should be treated as a broadcast since no ACK is required. The
peer will eventually see the notification or its watch will have timed
out and it will re-sync.

Fixes: http://tracker.ceph.com/issues/16470


trociny:
lgtm

PR #9923, opened 2016-06-24
librbd: removal of partially deleted image needs id lookup

dillaman:
Several operations depend on the image id but if the image cannot be
opened to retrieve the id, these cleanup operations cannot be executed.

Fixes: http://tracker.ceph.com/issues/16471

### DIFF
--- a/src/librbd/internal.cc
+++ b/src/librbd/internal.cc
@@ -2127,58 +2127,66 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force,
     }
 
     if (old_format || unknown_format) {
-      ldout(cct, 2) << "removing rbd image from directory..." << dendl;
+      ldout(cct, 2) << "removing rbd image from v1 directory..." << dendl;
       r = tmap_rm(io_ctx, imgname);
       old_format = (r == 0);
       if (r < 0 && !unknown_format) {
         if (r != -ENOENT) {
-	  lderr(cct) << "error removing img from old-style directory: "
-		     << cpp_strerror(-r) << dendl;
+          lderr(cct) << "error removing image from v1 directory: "
+                     << cpp_strerror(-r) << dendl;
         }
 	return r;
       }
     }
     if (!old_format) {
-      r = Journal<>::remove(io_ctx, id);
-      if (r < 0 && r != -ENOENT) {
-        lderr(cct) << "error removing image journal" << dendl;
-        return r;
+      if (id.empty()) {
+        ldout(cct, 5) << "attempting to determine image id" << dendl;
+        r = cls_client::dir_get_id(&io_ctx, RBD_DIRECTORY, imgname, &id);
+        if (r < 0 && r != -ENOENT) {
+          lderr(cct) << "error getting id of image" << dendl;
+          return r;
+        }
       }
+      if (!id.empty()) {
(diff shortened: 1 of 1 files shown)


src/librbd/internal.cc
```diff
@@ -2129,28 +2129,59 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force,
...
+        }
+
+        ldout(cct, 2) << "removing rbd image from v2 directory..." << dendl;
+        r = cls_client::dir_remove_image(&io_ctx, RBD_DIRECTORY, imgname, id);
+        if (r < 0) {
+          if (r != -ENOENT) {
+            lderr(cct) << "error removing img from new-style directory: "
+                       << cpp_strerror(-r) << dendl;
+          }
+          return r;
```
trociny: @dillaman I wonder why we return here even for r == -ENOENT? Don't you expect other removal below still could succeed? In the original version I thought it was because it was the latest op so there was nothing more to delete (well, there was cls_client::mirror_image_remove, but I suppose it was added later).

dillaman: @trociny Probably could be cleaned up since I can't think of it serving any real purpose.

trociny:
@dillaman Unfortunately, now removing non-existent image returns success, which I suppose is not correct, and this python test fails:
```
FAIL: test_rbd.test_remove_dne
    self.test(*self.arg)
  File "/home/mgolub/ceph/ceph.upstream/src/test/pybind/test_rbd.py", line 255, in test_remove_dne
    assert_raises(ImageNotFound, remove_image)
AssertionError: ImageNotFound not raised
```
May be return ENOENT if all operations returned ENOENT?

dillaman:
@trociny I moved that logic to the last operation (remove the rbd_id object) since that is also the first object created during an image creation and the last removed during a failed image creation.

trociny:
retest this please

trociny:
@dillaman Now qa/workunits/rbd/copy.sh fails on this test:
```
        # remove with header and id missing
	rbd create --image-format 2 -s 1 test2
	HEADER=$(rados -p rbd ls | grep '^rbd_header')
	rados -p rbd rm $HEADER
	rados -p rbd rm rbd_id.test2
	rbd rm test2
>>> librbd: removing id object...
>>> rbd: delete error: (2) No such file or directory
	rbd ls | wc -l | grep "^0$"
```

src/librbd/internal.cc
```diff
@@ -2127,59 +2127,68 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force,
...
-	return r;
-      }
-
-      ldout(cct, 2) << "removing image from rbd_mirroring object..." << dendl;
-      r = cls_client::mirror_image_remove(&io_ctx, id);
-      if (r < 0 && r != -ENOENT && r != -EOPNOTSUPP) {
-        lderr(cct) << "failed to remove image from mirroring directory: "
-                   << cpp_strerror(r) << dendl;
         return r;
+        }
```
trociny: @dillaman looks like an extra closing bracket. See the jenkins build failure.

dillaman:
@trociny sorry -- was waiting on a different test to complete before I could compile it.  Passes both copy.sh and the v1 Python tests locally.

trociny:
lgtm

trociny:
retest this please

PR #9925, opened 2016-06-24
mds: Add path filtering for dump cache

fullerdj:
Add a new admin socket command, dump below, to filter
cache dumps by path. An optional integer depth parameter
limits the depth below the specified path.

Fixes: http://tracker.ceph.com/issues/11171

### DIFF
--- a/src/mds/MDCache.cc
+++ b/src/mds/MDCache.cc
@@ -11601,11 +11601,17 @@ void MDCache::dump_cache(Formatter *f)
   dump_cache(NULL, f);
 }
 
+void MDCache::dump_cache(const string& dump_root, int depth, Formatter *f)
+{
+  dump_cache(NULL, f, dump_root, depth);
+}
+
 /**
  * Dump the metadata cache, either to a Formatter, if
  * provided, else to a plain text file.
  */
-void MDCache::dump_cache(const char *fn, Formatter *f)
+void MDCache::dump_cache(const char *fn, Formatter *f,
+			 const string& dump_root, int depth)
 {
   int r = 0;
   int fd = -1;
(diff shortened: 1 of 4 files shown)


Yan-waller:
retest this please

src/mds/MDSRank.cc
```diff
@@ -1758,6 +1758,13 @@ bool MDSRankDispatcher::handle_asok_command(
     } else {
       mdcache->dump_cache(path);
     }
+  } else if (command == "dump below") {
+    string root;
+    int64_t depth;
+    cmd_getval(g_ceph_context, cmdmap, "root", root);
+    if (!cmd_getval(g_ceph_context, cmdmap, "depth", depth))
+      depth = -1;
+    mdcache->dump_cache(root, depth, f);
```
jcsp: Take mds_lock with a Locker around mdcache->dump_cache

src/mds/MDSDaemon.cc
```diff
@@ -339,6 +346,7 @@ void MDSDaemon::clean_up_admin_socket()
   admin_socket->unregister_command("flush_path");
   admin_socket->unregister_command("export dir");
   admin_socket->unregister_command("dump cache");
+  admin_socket->unregister_command("dump below");
```
jcsp: (Subjective) Maybe "dump path" or "dump tree" would be clearer than "dump below"?

src/mds/MDCache.cc
```diff
@@ -11626,11 +11632,28 @@ void MDCache::dump_cache(const char *fn, Formatter *f)
...
        it != inode_map.end();
        ++it) {
     CInode *in = it->second;
+
+    if (!dump_root.empty()) {
+      string ipath;
+      if (in->is_root())
+	ipath = "/";
+      else
+	in->make_path_string(ipath);
```
jcsp: Hmm, this is going to ignore hard links under the requested path filter, but that's probably okay given that we're dumping a flat list of inodes anyway, and not pretending that this is a tree-like output.

jcsp:
No problems here, but let's have a short test in ceph-qa-suite that checks the filtering is applied, and that the depth field is respected.

src/mds/MDSRank.cc
fullerdj: Why, to avoid changes to files under the target path? Just racing here might be an appropriate choice since this is a debug dump.

jcsp: If metadata ops are in flight (e.g. inserting to inode_map) while we are iterating over the list in dump, we'll potentially segfault.

gregsfortytwo:
Hmm, I can't figure out what it is but the make-check bot is failing against us. Probably not our fault here but be careful when merging.

fullerdj:
Updated based on comments. See also https://github.com/ceph/ceph-qa-suite/pull/1066 for tests.

jcsp:
Reviewed-by, will merge with the updated test

gregsfortytwo:
Ping, Is this still waiting on the test update @fullerdj and @jcsp? :)

jcsp:
Indeed it is

PR #9926, opened 2016-06-24
msg/AsyncMessenger: remove unneeded friend decl

stiopaa1:
Removed unneeded friend declaration. Nested class already has access
to all the names(private, protected) to which enclosing class has
access(as of C++11).


yuyuyu101:
rebase

stiopaa1:
@yuyuyu101 
looks like it has been removed by some other commit - can be closed

PR #9933, opened 2016-06-25
kv: In memory keyvalue db implementation

chhabaramesh:
In memory key value db implementation with google btree map.
First version with minimal functional implementation.

### DIFF
--- a/src/common/config_opts.h
+++ b/src/common/config_opts.h
@@ -155,6 +155,7 @@ SUBSYS(bdev, 1, 3)
 SUBSYS(kstore, 1, 5)
 SUBSYS(rocksdb, 4, 5)
 SUBSYS(leveldb, 4, 5)
+SUBSYS(memdb, 4, 5)
 SUBSYS(kinetic, 1, 5)
 SUBSYS(fuse, 1, 5)
 
--- a/src/include/encoding.h
+++ b/src/include/encoding.h
@@ -1018,4 +1018,32 @@ inline void decode(std::deque<T>& ls, bufferlist::iterator& p)
       bl.advance(struct_end - bl.get_off());				\
   }
 
+/*
+ * Encoders/decoders to read from current offset in a file handle and
+ * encode/decode the data according to argument types.
+ */
+inline ssize_t decode_file(int fd, std::string &str)
+{
+  bufferlist bl;
+  __u32 len = 0;
+  bl.read_fd(fd, sizeof(len));
+  decode(len, bl);                                                                                                  
+  bl.read_fd(fd, len);
+  decode(str, bl);                                                                                                  
+  return bl.length();
+}
+
+inline ssize_t decode_file(int fd, bufferptr &bp)
+{
+  bufferlist bl;
+  __u32 len = 0;
+  bl.read_fd(fd, sizeof(len));
+  decode(len, bl);                                                                                                  
+  bl.read_fd(fd, len);
+  bufferlist::iterator bli = bl.begin();
+
+  decode(bp, bli);                                                                                                  
+  return bl.length();
+}
+
 #endif
--- a/src/kv/CMakeLists.txt
+++ b/src/kv/CMakeLists.txt
@@ -1,6 +1,7 @@
 set(kv_srcs
   KeyValueDB.cc
   LevelDBStore.cc
+  MemDB.cc
   RocksDBStore.cc)
 add_library(kv_objs OBJECT ${kv_srcs})
 add_library(kv STATIC $<TARGET_OBJECTS:kv_objs>)
--- a/src/kv/KeyValueDB.cc
+++ b/src/kv/KeyValueDB.cc
@@ -3,6 +3,7 @@
 
 #include "KeyValueDB.h"
 #include "LevelDBStore.h"
+#include "MemDB.h"
 #ifdef HAVE_LIBROCKSDB
 #include "RocksDBStore.h"
 #endif
@@ -29,6 +30,11 @@ KeyValueDB *KeyValueDB::create(CephContext *cct, const string& type,
     return new RocksDBStore(cct, dir, p);
   }
 #endif
+
+  if ((type == "memdb") && 
+    cct->check_experimental_feature_enabled("memdb")) {
+    return new MemDB(cct, dir, p);
+  }
   return NULL;
 }
 
@@ -47,5 +53,9 @@ int KeyValueDB::test_init(const string& type, const string& dir)
     return RocksDBStore::_test_init(dir);
   }
 #endif
+
+  if (type == "memdb") {
+    return MemDB::_test_init(dir);
+  }
   return -EINVAL;
 }
--- a/src/kv/KeyValueDB.h
+++ b/src/kv/KeyValueDB.h
@@ -124,6 +124,7 @@ public:
   virtual int init(string option_str="") = 0;
   virtual int open(std::ostream &out) = 0;
   virtual int create_and_open(std::ostream &out) = 0;
+  virtual void close() { }
 
   virtual Transaction get_transaction() = 0;
   virtual int submit_transaction(Transaction) = 0;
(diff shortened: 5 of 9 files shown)


chhabaramesh:

markhpc:
sweeeeet!  Thanks Ramesh!  Very excited to play with this.

chhabaramesh:
@markhpc you are welcome. But this is just start, might have to tune it for concurrency.

chhabaramesh:
@liewegas  All tests in ceph_test_objectstore and ceph_test_keyvaluedb passes now. Only problem was the libsnappy issue.

markhpc:
@chhabaramesh got a chance to test this today.  A couple of observations:

1) We need to modify the CMakeLists.txt file otherwise the linker get's angry when trying to build this with cmake.   I've done this here:  https://github.com/markhpc/ceph/commit/5318bb6e522ba91a1c3b008db75c92f0cfee5762

2) Pretty quickly I'm hitting segfaults on the OSDs when using this with bluestore.  I've included a snippet from gdb:

[MemDBStore_Segfault.txt](https://github.com/ceph/ceph/files/338362/MemDBStore_Segfault.txt)

chhabaramesh:
@markhpc , Thanks, will look in to it today.

chhabaramesh:
@markhpc , I fixed the make file issue as suggested. 

When do you hit segfault? Can you please tell exact test you are doing?

markhpc:
@chhabaramesh segfault happened about 0.3s after OSD initialization, but may only be on the first couple of OSDs (at least it is in this test).  Looks like the segfaults probably happened during initial default pool creation.

liewegas:
A good test BTW would be to set the mon_keyvaluedb and 
filestore_omap_backend to memdb and see if vstart works as expected.

markhpc:
Did a quick test with mon_keyvaluedb set to memdb which results in:

2016-06-29 11:42:38.494078 7f172b662500 -1 unable to read magic from mon data

No segfaults, mon just doesn't appear to like it.  Even with debug 20 there's not much in the mon log to look at.

chhabaramesh:
@markhpc  @liewegas , So far I did
1. ceph_test_objectstore + ceph_test_keyvaluedb
2. Vstart and then single rbd image and fio on top of it

for me it works ok without segfalut.

src/kv/MemDBStore.h
```diff
@@ -0,0 +1,208 @@
...
+{
+  string out = prefix;
+  out.push_back(KEY_DELIM);
+  out.append(value);
+  return out;
+}
+
+#define CEPH_MS_CONTAINER_NAME "ceph_kv_store"
+
+class MSStore : public KeyValueDB
```
liewegas: Can we call it MemDB{.h,.cc}?

src/kv/MemDBStore.h
```diff
@@ -0,0 +1,208 @@
...
+static inline
+string combine_strings(const string &prefix, const string &value)
+{
+  string out = prefix;
+  out.push_back(KEY_DELIM);
+  out.append(value);
+  return out;
+}
+
+#define CEPH_MS_CONTAINER_NAME "ceph_kv_store"
```
liewegas: unused

src/kv/MemDBStore.h
```diff
@@ -0,0 +1,208 @@
...
+#include <boost/scoped_ptr.hpp>
+#include "include/encoding.h"
+#include "include/cpp-btree/btree.h"
+#include "include/cpp-btree/btree_map.h"
+#include "include/encoding_btree.h"
+#include "KeyValueDB.h"
+#include "osd/osd_types.h"
+
+using std::string;
+#define KEY_DELIM 'X'
```
liewegas: This should be \0

src/kv/MemDBStore.h
```diff
@@ -0,0 +1,208 @@
...
+  CephContext *m_cct;
+  void* m_priv;
+  string m_options;
+  string m_db_path;
+
+  int transaction_rollback(KeyValueDB::Transaction t);
+  int _open(ostream &out);
+  void close();
+  bool _get(const string &prefix, const string &k, bufferlist *out);
+  bool _get_locked(const string &prefix, const string &k, bufferlist *out);
```
liewegas: make this public and make it overload the single-key get() in KeyValueDB

src/kv/MemDBStore.cc
```diff
@@ -0,0 +1,523 @@
...
+void MSStore::_encode(btree::btree_map<string,
+                      bufferptr>:: iterator iter, bufferlist &bl)
+{
+  ::encode(iter->first, bl);
+  ::encode(iter->second, bl);
+}
+
+std::string MSStore::_get_data_fn()
+{
+  string fn = m_db_path+"//"+"MemDB.db";
```
liewegas: single /, spaces between operators and operands

liewegas:
I pushed several patches to https://github.com/liewegas/ceph/commits/wip-mondb.. enough to get the monitor to get most of the way into its startup sequence with 'mon keyvaluedb = memdb'. But it quickly crashes in the iterator code.

liewegas:
Ah, I disabled the snapshot iterator usage in mon and now it works!

src/kv/MemDBStore.cc
```diff
@@ -0,0 +1,523 @@
...
+  }
+
+  m_btree[key] = bufferptr((char *) bl.c_str(), bl.length());
+
+  return 0;
+}
+
+int MSStore::_rmkey(ms_op_t &op)
+{
+  std::lock_guard<std::mutex> l(m_lock);
```
mattbenjamin: could shrink critical section (and elsewhere)?

src/kv/MemDBStore.h
chhabaramesh: Sure, will change it.

src/kv/MemDBStore.h
chhabaramesh: will remove this.

src/kv/MemDBStore.h
chhabaramesh: Will do that.

src/kv/MemDBStore.h
chhabaramesh: will do that.

src/kv/MemDBStore.cc
chhabaramesh: will do that.

src/kv/MemDBStore.cc
chhabaramesh: Yes, I will take utility code out of lock and break code wherever possible. Also, I have in mind overall concurrency improvement for this code using multiple btrees.

src/kv/MemDBStore.h
chhabaramesh: get in KeyValueDB is overloaded and these two functions are just internal helper functions to get value from btree based on lock is needed or not.

src/kv/MemDBStore.cc
chhabaramesh: Only thing I can take out of lock is combine strings. Rest needs to be under the lock. For example getting old value and then erasing it needs to be atomic, otherwise erase might fault.

liewegas:
We need to test and merge https://github.com/ceph/ceph/pull/10102 first, then rebase this on top.

liewegas:
Otherwise this looks good to me!

tchaikov:
#10102 is merged.

chhabaramesh:
@markhpc  @tchaikov  @liewegas  rebased and tested.

markhpc:
Sounds to me like we should merge it!

chhabaramesh:
@markhpc  @tchaikov  any update on this?

mattbenjamin:
> Yes, I will take utility code out of lock and break code wherever possible. Also, I have in mind overall >  concurrency improvement for this code using multiple btrees.

The TreeX class in cohort_lru.h does something along those lines, but is based on the boost::intrusve tree interface.

PR #9934, opened 2016-06-25
common: instantiate strict_si_cast<long> not strict_si_cast<int64_t>

tchaikov:
this fixes the build on armf.

on 32bit platforms, cstdint is very likely to

 typedef long long int int64_t;

this results in compilation error like

 `common/strtol.cc:190:75: error: duplicate explicit instantiation of 'T
 strict_si_cast(const char, std::string) [with T = long long int;
 std::string = std::basic_string]'

 [-fpermissive]
 template int64_t strict_si_cast(const char *str, std::string *err);
 ^`

we can address this by instantiate the primitive type of `long long`
instead of `in64_t`.

Fixes: http://tracker.ceph.com/issues/16398

### DIFF
--- a/src/common/strtol.cc
+++ b/src/common/strtol.cc
@@ -186,8 +186,8 @@ T strict_si_cast(const char *str, std::string *err)
 }
 
 template int strict_si_cast<int>(const char *str, std::string *err);
+template long strict_si_cast<long>(const char *str, std::string *err);
 template long long strict_si_cast<long long>(const char *str, std::string *err);
-template int64_t strict_si_cast<int64_t>(const char *str, std::string *err);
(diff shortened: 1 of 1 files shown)


vumrao:
@tchaikov thanks Kefu. Then I think once this one is merged, I need to backport this to jewel backport : https://github.com/ceph/ceph/pull/9743.

ukernel:
see https://github.com/ceph/ceph/pull/9371/commits/c158e4da8728d1a624ebee07b8b08fcecd9eb788. I think you miss a few places

tchaikov:
@ukernel thanks. replied you in your commit.

ukernel:
got it

tchaikov:
tested at 

tchaikov:

athanatos:
lgtm

PR #9936, opened 2016-06-26
ceph-disk: Compatibility fixes for Python 3

onyb:


src/ceph-disk/ceph_disk/main.py
```diff
@@ -4092,7 +4091,7 @@ def list_devices():
 
     uuid_map = {}
     space_map = {}
-    for base, parts in sorted(six.iteritems(partmap)):
+    for base, parts in sorted(partmap.items()):
```
tchaikov: could you squash this commit into the previous one introducing `six`?

onyb: @tchaikov Done.

src/ceph-disk/ceph_disk/main.py
```diff
@@ -18,6 +18,8 @@
 # GNU Library Public License for more details.
```
jdurgin: minor: the commit message lists some commit subjects that you squashed - you can delete those

jdurgin:
these fixes look good to me - as discussed on irc, I think we should use py34 as the python 3 env, and add python34 to install-deps.sh so 'make check' passes on centos 7 and ubuntu trusty

src/ceph-disk/ceph_disk/main.py
onyb: @jdurgin: Done.

src/ceph-disk/ceph_disk/main.py
```diff
@@ -743,17 +745,17 @@ def is_mounted(dev):
...
-    with file('/proc/mounts', 'rb') as proc_mounts:
+    with open('/proc/mounts', 'rb') as proc_mounts:
         for line in proc_mounts:
             fields = line.split()
             if len(fields) < 3:
                 continue
             mounts_dev = fields[0]
             path = fields[1]
-            if mounts_dev.startswith('/') and os.path.exists(mounts_dev):
+            if mounts_dev.startswith(b'/') and os.path.exists(mounts_dev):
```
tchaikov: could use `os.path.isabs(mounts_dev)` instead of `mounts_dev.startswith(b'/')`

src/ceph-disk/tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = flake8,py27
+envlist = flake8,py27,py35
```
tchaikov: agree with @jdurgin, trusty comes with py34, see http://packages.ubuntu.com/trusty/python3

src/ceph-disk/ceph_disk/main.py
```diff
@@ -323,21 +325,16 @@ def is_systemd():
     """
     Detect whether systemd is running
     """
-    with file('/proc/1/comm', 'rb') as i:
-        for line in i:
-            if 'systemd' in line:
-                return True
-    return False
+    with open('/proc/1/comm', 'rb') as f:
+        return 'systemd' in f.read()
```
BlaXpirit: <code>'systemd' in <b>&lt;bytestring&gt;</b></code> &ndash; please check if this is correct

tchaikov:
needs rebase.

src/ceph-disk/ceph_disk/main.py
```diff
@@ -1987,6 +1987,7 @@ def prepare_file(self):
                     'not the same device as the osd data' %
                     self.name)
         self.space_symlink = space_file
```
jdurgin: it looks like space_symlink should be set to the path, not the file object

jdurgin:
lgtm

src/ceph-disk/ceph_disk/main.py
```diff
@@ -213,7 +212,6 @@ def is_dmcrypt(ptype, name):
     'upstart',
     'sysvinit',
     'systemd',
-    'openrc',
```
tchaikov: @onyb is this change on purpose?

src/ceph-disk/ceph_disk/main.py
```diff
@@ -212,6 +212,7 @@ def is_dmcrypt(ptype, name):
     'upstart',
     'sysvinit',
     'systemd',
+    'openrc',
```
tchaikov: this commit involves at least three changes:

- put the openrc support back into ceph-disk
- run py3.4 instead of py3.5 in run-tox
- address the comment regarding the `os.path.isabs()` versus `startswith('/')`.

could you reset this change and squash the related changes into the corresponding commits?

jdurgin:
tests passed, but this needs a rebase now

onyb:
@jdurgin Rebased and addressed all comments by @tchaikov . Thanks!

tchaikov:
lgtm.

tchaikov:
@jdurgin is this PR still under testing?

PR #9939, opened 2016-06-27
rados/client: fix waiting on the condition variable more efficient.

Yan-waller:
Maybe we can simply use `cond.Wait` if timeout is zero, 
so that it wait and sleep forever until being waked up by
`cond.Signal`. it seems to be more efficient.

### DIFF
--- a/src/librados/RadosClient.cc
+++ b/src/librados/RadosClient.cc
@@ -516,7 +516,7 @@ int librados::RadosClient::wait_for_osdmap()
   bool need_map = false;
   objecter->with_osdmap([&](const OSDMap& o) {
       if (o.get_epoch() == 0) {
-	need_map = true;
+        need_map = true;
       }
     });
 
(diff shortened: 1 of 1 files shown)


xiexingguo:
:+1:

src/librados/RadosClient.cc
```diff
@@ -527,25 +527,21 @@ int librados::RadosClient::wait_for_osdmap()
...
       utime_t start = ceph_clock_now(cct);
       while (objecter->with_osdmap(std::mem_fn(&OSDMap::get_epoch)) == 0) {
-	cond.WaitInterval(cct, lock, timeout);
-	utime_t elapsed = ceph_clock_now(cct) - start;
-	if (!wait_forever && elapsed > timeout) {
-	  lderr(cct) << "timed out waiting for first osdmap from monitors"
-		     << dendl;
-	  return -ETIMEDOUT;
-	}
+        if(timeout.is_zero()) {
```
liewegas: s/if(/if (/ (whitespace)

Yan-waller: @liewegas thanks, I have fixed it.

PR #9940, opened 2016-06-27
common: fix lockdep vs recursive mutexes

aclamk:
The problem is that:
Mutex lock("RGWKeystoneTokenCache", true /\* recursive */);
Mutex::Locker l(lock);
lock.Lock();
fails...
Signed-off-by: Adam Kupczyk akupczyk@mirantis.com

### DIFF
--- a/src/test/common/test_mutex.cc
+++ b/src/test/common/test_mutex.cc
@@ -1,0 +1,66 @@
+// -*- mode:C++; tab-width:8; c-basic-offset:2; indent-tabs-mode:t -*-
+// vim: ts=8 sw=2 &smarttab
+/*
+ * Ceph - scalable distributed file system
+ *
+ */
+
+#include <common/Mutex.h>
+#include "gtest/gtest.h"
+#include "common/ceph_context.h"
+#include "common/config.h"
+
+/*
+ * Override normal ceph assert.
+ * It is needed to prevent hang when we assert() and THEN still wait on lock().
+ */
+namespace ceph
+{
+  void __ceph_assert_fail(const char *assertion, const char *file, int line,
+        const char *func)
+  {
+    throw 0;
+  }
+}
+
+void do_init() {
+  static CephContext* cct = nullptr;
+  if (cct == nullptr) {
+    cct = new CephContext(0);
+    lockdep_register_ceph_context(cct);
+  }
+}
+
+TEST(MutexDebug, NormalAsserts) {
+  Mutex* m = new Mutex("Normal",false);
+  m->Lock();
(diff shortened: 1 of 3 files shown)


liewegas:
Do you mind creating a unit test for this?  We don't actually have any lockdep unit tests, but just a few basic ones would be really nice.  (e.g., demonstrate that a nonrecursive AA or ABBA will assert, and that a recursive AA is okay.  gtest has the assert macros that assert that the test aborts.)

aclamk:
Ok, I will gladly write some unit tests.

liewegas:
LGTM, but needs a rebase now that we're cmake!

tchaikov:
@aclamk ping?

aclamk:
@tchaikov rebased.

src/test/common/test_mutex.cc
```diff
@@ -0,0 +1,66 @@
+// -*- mode:C++; tab-width:8; c-basic-offset:2; indent-tabs-mode:t -*-
+// vim: ts=8 sw=2 &smarttab
+/*
+ * Ceph - scalable distributed file system
+ *
+ */
+
+#include <common/Mutex.h>
+#include "gtest/gtest.h"
```
tchaikov: might want to update the `CMakeLists.txt` to include this test.

tchaikov:
could you prefix the title of your commit message with the subcomponent your are changing ? see https://github.com/ceph/ceph/blob/master/SubmittingPatches.rst#3-describe-your-changes

src/test/common/test_mutex.cc
aclamk: @tchaikov: The unittest_mutex_debug is already a part of src/test/common/CMakeLists.txt, lines 191-194. Can you confirm?

tchaikov: @aclamk it's "test_mutex_debug.cc" not "test_mutex.cc".

tchaikov:
@aclamk f0fe3f6285e4b98374846b8897193e15fe8aead5 is already merged in #12947. could you remove it from this PR?

tchaikov:
retest this please.

liewegas:
can you drop the ceph-disk patch from this PR please?

aclamk:
@liewegas Dropped ceph-disk patch. I have no idea how it even could get there....

tchaikov:
- http://pulpito.ceph.com/kchai-2017-03-20_17:19:01-rados-wip-kefu-testing---basic-smithi/
- http://pulpito.ceph.com/kchai-2017-03-23_06:49:46-rados-wip-kefu-testing---basic-mira/

PR #9943, opened 2016-06-27
ceph-disk: do not activate device that is not ready

b-ranto:
If the journal (or data) device is not ready when we are activating the
data (or journal) device, just print an info message and exit with 0 so
that the ceph-disk systemd service won't fail in this case.

Fixes: http://tracker.ceph.com/issues/15990

### DIFF
--- a/src/ceph-disk/ceph_disk/main.py
+++ b/src/ceph-disk/ceph_disk/main.py
@@ -3334,6 +3334,13 @@ def main_activate(args):
         else:
             raise Error('%s is not a directory or block device' % args.path)
 
+        # exit with 0 if the journal device is not up, yet
+        # journal device will do the activation
+        osd_journal = '{path}/journal'.format(path=osd_data)
+        if os.path.islink(osd_journal) and not os.access(osd_journal, os.F_OK):
+            LOG.info("activate: Journal not present, not starting, yet")
+            return
+
         if (not args.no_start_daemon and args.mark_init == 'none'):
             command_check_call(
                 [
@@ -3341,7 +3348,7 @@ def main_activate(args):
                     '--cluster={cluster}'.format(cluster=cluster),
                     '--id={osd_id}'.format(osd_id=osd_id),
                     '--osd-data={path}'.format(path=osd_data),
-                    '--osd-journal={path}/journal'.format(path=osd_data),
+                    '--osd-journal={journal}'.format(journal=osd_journal),
                 ],
             )
 
@@ -3734,6 +3741,12 @@ def main_activate_space(name, args):
 
         if is_suppressed(path):
             LOG.info('suppressed activate request on %s', path)
+            return
+
+        # warn and exit with 0 if the data device is not up, yet
+        # data device will do the activation
+        if not os.access(path, os.F_OK):
+            LOG.info("activate: OSD device not present, not starting, yet")
             return
 
         (cluster, osd_id) = mount_activate(


b-ranto:
@dachary what do you think? imo, this could work and we would not just ignore the errors. FWIW, we already do the ~same in ceph-osd-prestart.sh.

src/ceph-disk/ceph_disk/main.py
```diff
@@ -3311,6 +3311,12 @@ def main_activate(args):
         else:
             raise Error('%s is not a directory or block device' % args.path)
 
+        # exit with 0 if the journal device is not up, yet 
+        # journal device will do the activation
+        if not os.access('{path}/journal'.format(path=osd_data), os.F_OK):
```
dachary: not all OSD have a journal

dachary:
It looks good, modulo the no journal case :-)

dachary:
And the flake8 / test cases that need to be adjusted accordingly.

b-ranto:
Rebased and hopefully fixed. I inspired with the ceph-osd-prestart.sh and the patch checks whether the journal is a sym-link and only then it tries to follow it.

dachary:
@b-ranto I suspect the test timesout. If you run it locally you'll get more information about why it fails.

b-ranto:
I've tested my ~local mock build and it fixed the original issue (ct#15990) for me. I no longer see any ceph-disk failures when the cluster is being installed.

@dachary: I ran the ./run-make-check.sh manually and I got two failures, no time-out and all three that failed in jenkins passed on my machine. The two failures I got don't seem to be related to this patch:

* unittest_librbd seg-faulted
* failure in  unittest_chain_xattr [1]

[1] test/objectstore/chain_xattr.cc:371: Failure
Value of: get_xattrs(fd).size()
  Actual: 2
Expected: 1UL
Which is: 1

b-ranto:
FYI: After the last rebase, this passed the jenkins build.

b-ranto:
@dachary any objections to merging this?

dachary:
@b-ranto the ceph-disk suite should be run to validate this change. The make check tests are not enough.

src/ceph-disk/ceph_disk/main.py
```diff
@@ -3334,14 +3334,21 @@ def main_activate(args):
         else:
             raise Error('%s is not a directory or block device' % args.path)
 
+        # exit with 0 if the journal device is not up, yet
+        # journal device will do the activation
+        osd_journal = '{path}/journal'.format(path=osd_data)
```
dachary: That won't work for bluestore, because it does not have a journal file.

b-ranto: To be fair, it is not a regression in this respect either -- ceph-disk have passed the --osd-journal argument with this path even before this (old line 3344). If the file does not exist, the next if will simply fail (it is not a link if it does not exist) and that osd_journal path will be sent to ceph-osd --osd-journal flag just like before.

b-ranto:
@dachary are we talking teuthology ceph-disk suite?

dachary:
@b-ranto yes, the  teuthology ceph-disk suite

tchaikov:
ceph-disk test run failed, see https://github.com/ceph/ceph/pull/10135#issuecomment-240927450

tchaikov:
http://pulpito.ceph.com/kchai-2016-08-25_02:27:35-ceph-disk-wip-kefu-testing---basic-vps/ fails,  it has #10824, #10825, #9943 and #10135. @dachary does this mean that either #9943 and/or #10135 is broken?

dachary:
@tchaikov https://github.com/ceph/ceph/pull/10135#issuecomment-242424224 is failing.

tchaikov:
@dachary thanks. rescheduling without #10135.

tchaikov:
passes in http://pulpito.ceph.com/kchai-2016-08-25_17:28:34-ceph-disk-wip-kefu-testing---basic-vps/

PR #9946, opened 2016-06-27
os/bluestore: make trim() of 2Q cache more fine-grained

xiexingguo:
1. count total bytes as per buffer types respectively, 
   so we know the exact size of each kind of buffer.

2. evict by list actual size against expected size,
   which is more accurate.

### DIFF
--- a/src/os/bluestore/BlueStore.cc
+++ b/src/os/bluestore/BlueStore.cc
@@ -654,6 +654,7 @@ void BlueStore::TwoQCache::_add_buffer(Buffer *b, int level, Buffer *near)
   }
   if (!b->is_empty()) {
     buffer_bytes += b->length;
+    buffer_list_bytes[b->cache_private] += b->length;
   }
 }
 
@@ -663,6 +664,8 @@ void BlueStore::TwoQCache::_rm_buffer(Buffer *b)
  if (!b->is_empty()) {
     assert(buffer_bytes >= b->length);
     buffer_bytes -= b->length;
+    assert(buffer_list_bytes[b->cache_private] >= b->length);
+    buffer_list_bytes[b->cache_private] -= b->length;
   }
   switch (b->cache_private) {
   case BUFFER_WARM_IN:
@@ -685,6 +688,8 @@ void BlueStore::TwoQCache::_adjust_buffer_size(Buffer *b, int64_t delta)
   if (!b->is_empty()) {
     assert((int64_t)buffer_bytes + delta >= 0);
     buffer_bytes += delta;
+    assert((int64_t)buffer_list_bytes[b->cache_private] + delta >= 0);
+    buffer_list_bytes[b->cache_private] += delta;
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/os/bluestore/BlueStore.cc
```diff
@@ -698,48 +702,103 @@ void BlueStore::TwoQCache::trim(uint64_t onode_max, uint64_t buffer_max)
...
-      assert(b->is_clean());
-      dout(20) << __func__ << " buffer_hot rm " << *b << dendl;
-      b->space->_rm_buffer(b);
-    }
-    while (buffer_warm_in.size() > kin) {
-      Buffer *b = &*buffer_warm_in.rbegin();
+  // first adjust clean(warm_in and hot) buffer lists
+  if (buffer_bytes > buffer_max) {
+
+    int64_t to_evict_bytes = buffer_bytes - buffer_max;
```
liewegas: It looks like this misses the key part of 2Q that sizes the warm_in vs hot lists.  It looks like this will trim everything from warm_in until it is empty before even touching hot items?

src/os/bluestore/BlueStore.cc
```diff
@@ -698,48 +702,103 @@ void BlueStore::TwoQCache::trim(uint64_t onode_max, uint64_t buffer_max)
...
+               << dendl;
+    } else {
+      dout(20) << __func__ << " reclaimed " << prettybyte_t(reclaimed)
+               << " from hot list, done reclaiming clean buffers"
+               << dendl;
+    }
+  }
+
+  // adjust ghost(warm_out) list too, if necessary
+  uint64_t& ghost_list_bytes = buffer_list_bytes[BUFFER_WARM_OUT];
```
liewegas: also, i'm not sure it makes sense to count warm_out in terms of bytes.  the previous behavior of using average buffer size to count buffers seems like it would be more fair?

liewegas:
The first 2 commits looks good (although it was meant to be 'slack' and not 'back).  I'm not sure that the final patch changing 2q works.. or that it would be an improvement?

xiexingguo:
I understand the 2Q cache shall act like ARC in some way, that the hot_list is always prior to warm_in list and thus shall be evicted first whenever we have to.

Further improvements might include:

- change the way an element jump from the warm_in list to hot_list(frequencybased, may be?) 
 

- apply quotas to each size of the two lists

But I am also ok to drop that patch, as you wish.

liewegas:
2q tries to balance space between hot and warm_in.. 50/50 iirc. see 
section 2 on right side of page 441 of

	http://www.vldb.org/conf/1994/P439.PDF

25% for kin (warm_in), 50% for kout (warm_out), 25% for hot.  Since kout 
are just slot metadata (no bytes), the byte allocation is 50/50 for hot vs 
warm_in.  There's some analysis about how they chose these values at the 
end of the paper but in the end this is what they recommend.

I think the change that does byte counting per list type makes sense so 
that we don't rely on average buffer size for warm_in vs hot.  But I  
would say stick to average buffer size of warm_out?

xiexingguo:
Well, I'll take a careful look at the paper and recommend your suggestion and then try to make some further improvements.

Thanks, sage:-)

liewegas:
will close this for now!

xiexingguo:
retest this please

xiexingguo:
@liewegas See again?

liewegas:
Looks good. will do some testing before merge

xiexingguo:
changeset:

    fix a typo( caculate -> calculate)

PR #9949, opened 2016-06-27
librbd: potential use after free on refresh error

trociny:


trociny:
@dillaman I noticed this reading the code, so did not hit it, and not sure if we need to create a tracker ticket for this.

dillaman:
@trociny I'll just cherry-pick it to Jewel right away so it doesn't get dropped.

PR #9950, opened 2016-06-27
rgw: should delete in_stream_req  if conn->get_obj(...) return not zero value

weiqiaomiao:

### DIFF
--- a/src/rgw/rgw_rest_conn.cc
+++ b/src/rgw/rgw_rest_conn.cc
@@ -163,7 +163,13 @@ int RGWRESTConn::get_obj(const rgw_user& uid, req_info *info /* optional */, rgw
     set_header(mod_pg_ver, extra_headers, "HTTP_DEST_PG_VER");
   }
 
-  return (*req)->get_obj(key, extra_headers, obj);
+  int r = (*req)->get_obj(key, extra_headers, obj);
+  if (r < 0) {
+    delete *req;
+    *req = nullptr;
+  }
+  
+  return r;
 }
 
 int RGWRESTConn::complete_request(RGWRESTStreamReadRequest *req, string& etag, real_time *mtime, map<string, string>& attrs)


weiqiaomiao:
retest this please

src/rgw/rgw_rados.cc
```diff
@@ -6590,7 +6590,7 @@ int RGWRados::fetch_remote_obj(RGWObjectCtx& obj_ctx,
 {
   /* source is in a different zonegroup, copy from there */
 
-  RGWRESTStreamReadRequest *in_stream_req;
+  RGWRESTStreamReadRequest *in_stream_req = NULL;
```
mattbenjamin: should be nullptr these days, but ok :)

src/rgw/rgw_rados.cc
```diff
@@ -6676,6 +6676,9 @@ int RGWRados::fetch_remote_obj(RGWObjectCtx& obj_ctx,
                       dest_mtime_weight.zone_short_id, dest_mtime_weight.pg_ver,
                       true, &cb, &in_stream_req);
   if (ret < 0) {
```
mattbenjamin: does look correct

mattbenjamin:
@cbodley do you have any issue with this?

cbodley:
the patch looks correct, but we might want to attack the issue in `RGWRESTConn::get_obj()` instead. if it returns an error, it should clean up the `RGWRESTStreamReadRequest` itself instead of making the caller do it

cbodley:
i.e. at the bottom of `RGWRESTConn::get_obj()`:
```diff
-  return (*req)->get_obj(key, extra_headers, obj);
+  int r = (*req)->get_obj(key, extra_headers, obj);
+  if (r < 0) {
+    delete *req;
+    *req = nullptr;
+  }
+  return r;
```

mattbenjamin:
@cbodley agree;  if get_obj failed, there was no reason for it to return something for us to delete
@weiqiaomiao can you revise this?

weiqiaomiao:
@mattbenjamin  yes, i revise it.

weiqiaomiao:
retest this please

weiqiaomiao:
retest this please

weiqiaomiao:
@mattbenjamin ping

PR #9954, opened 2016-06-27
mds/mdcache: merge assert in path_traverse()

renhwztetecs:
merge assert and Remove excess judgment

### DIFF
--- a/src/mds/MDCache.cc
+++ b/src/mds/MDCache.cc
@@ -7655,12 +7655,10 @@ int MDCache::path_traverse(MDRequestRef& mdr, Message *req, MDSInternalContextBa
   bool null_okay = (onfail == MDS_TRAVERSE_DISCOVERXLOCK);
   bool forward = (onfail == MDS_TRAVERSE_FORWARD);
 
-  assert(mdr || req || fin);
-  assert(!forward || mdr || req);  // forward requires a request
+  assert(mdr || req || fin || !forward);
 
   snapid_t snapid = CEPH_NOSNAP;
-  if (mdr)
-    mdr->snapid = snapid;
+  mdr->snapid = snapid;
 
   client_t client = (mdr && mdr->reqid.name.is_client()) ? mdr->reqid.name.num() : -1;
 


batrick:
LGTM

src/mds/MDCache.cc
```diff
@@ -7655,12 +7655,10 @@ int MDCache::path_traverse(MDRequestRef& mdr, Message *req, MDSInternalContextBa
   bool null_okay = (onfail == MDS_TRAVERSE_DISCOVERXLOCK);
   bool forward = (onfail == MDS_TRAVERSE_FORWARD);
 
-  assert(mdr || req || fin);
-  assert(!forward || mdr || req);  // forward requires a request
+  assert(mdr || req || fin || !forward);
```
gregsfortytwo: These aren't the same, logically. If we have a fin this passes even if forward is set, which is incorrect. You could just stick the logical conditions together with an AND if you want to reduce the calls to assert(), or maybe

```assert( (!forward && fin) || mdr || req );```

if I'm calculating properly. But that'll require a comment explaining itself.

Comment (xiexingguo) [inline src/mds/MDCache.cc]: @renhwztetecs I am with greg, the merge is somewhat misleading so I'd like it stay this way.

File: src/mds/MDCache.cc [hunk 2]
```diff
@@ -7655,12 +7655,10 @@ int MDCache::path_traverse(MDRequestRef& mdr, Message *req, MDSInternalContextBa
...
 
   snapid_t snapid = CEPH_NOSNAP;
-  if (mdr)
-    mdr->snapid = snapid;
+  mdr->snapid = snapid;
```
xiexingguo: Why do you drop the sanity check here?

I'd suggest to add some sanity check at https://github.com/ceph/ceph/pull/9954/files#diff-2259d0a61fc0c9b2d2be3e40ade57303L7667 and https://github.com/ceph/ceph/pull/9954/files#diff-2259d0a61fc0c9b2d2be3e40ade57303L7667 ... instead...

renhwztetecs:
I misunderstood mdr not be null，in fact mdr&&req can be null
so do not modify, I close it

PR #9957, opened 2016-06-27
os/bluestore: get rid off blob's ref_map for non-shared objects

ifed01:
Ref_map is built on the fly for onodes that are not shared. Hence provides saving for both serialized and in-memory onode representation.
PR is based on PR#9866

### DIFF
--- a/src/os/bluestore/BlueStore.cc
+++ b/src/os/bluestore/BlueStore.cc
@@ -2739,21 +2739,20 @@ int BlueStore::_fsck_verify_blob_map(
 	   << " has no lextent refs" << dendl;
       ++errors;
     }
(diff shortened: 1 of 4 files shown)


liewegas:
I think the approach here makes sense.  Can we

- move the get_ref and put_ref helpers into onode_t, so that we can write unit tests for them in test_bluestore_types.cc?  I think they'll still take the bluestore_blob_t pointer, but it looks like the other stuff in Blob or Onode isn't necessary?
- drop the first commit (looks like rebase noise?)
- separate this into separate patches, like adding the new compressed length field, adding the get/put ref methods with unit tests, then doing the final refmap change.

Thanks!

ifed01:
Replaced with #9988

PR #9958, opened 2016-06-27
rbd-mirror: clean up ImageReplayer before stopping state machine

dillaman:
Fixes: http://tracker.ceph.com/issues/16489


trociny:
lgtm

PR #9959, opened 2016-06-27
os/bluestore: add  some sanity checks

xiexingguo:


xiexingguo:
@liewegas Move the clean ups here.

PR #9963, opened 2016-06-27
os/bluestore: add libzbc library for SMR support in bluestore

shehbazj:
* changed autotools to CMake for libzbc
* added API to call libzbc functions available in zbc_tools.h

Fixes: http://tracker.ceph.com/issues/16174
Signed-off-by: Shehbaz Jaffer <shehbaz.jaffer@mail.utoronto.ca>


liewegas:
It looks like this adds the libzbc files directly into ceph.git.  We should use a git submodule instead.  see 'git submodule add ...' to do that.  I created a fork of libzbc at https://github.com/ceph/libzbc that we can use as the upstream here.  Ping me on irc if you need a hand!

PR #9965, opened 2016-06-27
mds,libcephfs: add ceph_statx support and expose birthtime via it

jtlayton:
The birthtime for an inode is the time when it was first created. It's not updated on any other event, aside from being explicitly set.

This PR builds on the work that Sage started a couple of weeks ago to track the btime in the inode. It also adds a new statx() like interface to libcephfs to allow applications to get at it.

Note that the testcase is not actually tested yet (waiting for the build to finish), but I *think* it'll DTRT.

Also, we may need some cmake changes to ensure that the statx file is installed at install time, and I'm not 100% sure that that is how to want to statx definitions...

### DIFF
--- a/src/client/Client.cc
+++ b/src/client/Client.cc
@@ -835,6 +835,7 @@ Inode * Client::add_update_inode(InodeStat *st, utime_t from,
 
     // immutable bits
     in->ino = st->vino.ino;
+    in->btime = st->btime;
     in->snapid = st->vino.snapid;
     in->mode = st->mode & S_IFMT;
     was_new = true;
@@ -9461,6 +9462,7 @@ Inode *Client::open_snapdir(Inode *diri)
     in->gid = diri->gid;
     in->mtime = diri->mtime;
     in->ctime = diri->ctime;
+    in->btime = diri->btime;
     in->size = diri->size;
 
     in->dirfragtree.clear();
--- a/src/client/Client.cc
+++ b/src/client/Client.cc
@@ -9644,15 +9644,26 @@ Inode *Client::ll_get_inode(vinodeno_t vino)
   return in;
 }
 
-int Client::ll_getattr(Inode *in, struct stat *attr, int uid, int gid)
+int Client::_ll_getattr(Inode *in, int uid, int gid)
 {
-  Mutex::Locker lock(client_lock);
-
   vinodeno_t vino = _get_vino(in);
 
   ldout(cct, 3) << "ll_getattr " << vino << dendl;
   tout(cct) << "ll_getattr" << std::endl;
   tout(cct) << vino.ino.val << std::endl;
+
+  if (vino.snapid < CEPH_NOSNAP)
+    return 0;
+  else
+    return _getattr(in, CEPH_STAT_CAP_INODE_ALL, uid, gid);
+}
+
+int Client::ll_getattr(Inode *in, struct stat *attr, int uid, int gid)
+{
+  Mutex::Locker lock(client_lock);
+
+  int res = _ll_getattr(in, uid, gid);
+  vinodeno_t vino = _get_vino(in);
 
   /* special case for dotdot (..) */
   if (vino.ino.val == CEPH_INO_DOTDOT) {
@@ -9661,11 +9672,6 @@ int Client::ll_getattr(Inode *in, struct stat *attr, int uid, int gid)
     return 0;
   }
 
-  int res;
-  if (vino.snapid < CEPH_NOSNAP)
-    res = 0;
-  else
-    res = _getattr(in, CEPH_STAT_CAP_INODE_ALL, uid, gid);
   if (res == 0)
     fill_stat(in, attr);
   ldout(cct, 3) << "ll_getattr " << vino << " = " << res << dendl;
--- a/src/client/Client.cc
+++ b/src/client/Client.cc
@@ -6620,14 +6620,19 @@ int Client::statx(const char *relpath, unsigned int flags, struct statx *stx,
   tout(cct) << relpath << std::endl;
   filepath path(relpath);
   InodeRef in;
+
   int r = path_walk(path, &in);
   if (r < 0)
     return r;
-  r = _getattr(in, mask);
-  if (r < 0) {
-    ldout(cct, 3) << "statx exit on error!" << dendl;
-    return r;
+
+  if (!(flags & AT_NO_ATTR_SYNC)) {
+    r = _getattr(in, mask);
+    if (r < 0) {
+      ldout(cct, 3) << "statx exit on error!" << dendl;
+      return r;
+    }
   }
+
   fill_statx(in, stx, dirstat);
   ldout(cct, 3) << "statx exit (relpath " << relpath << " mask " << mask << ")" << dendl;
   return r;
@@ -9761,7 +9766,11 @@ int Client::ll_getattrx(Inode *in, unsigned int flags, struct statx *stx, int ui
 {
   Mutex::Locker lock(client_lock);
 
-  int res = _ll_getattr(in, uid, gid);
+  int res = 0;
+
+  if (!(flags & AT_NO_ATTR_SYNC))
+    res = _ll_getattr(in, uid, gid);
+
   vinodeno_t vino = _get_vino(in);
 
   /* special case for dotdot (..) */
--- a/src/client/Client.cc
+++ b/src/client/Client.cc
@@ -94,6 +94,8 @@
 #include "include/assert.h"
 #include "include/stat.h"
 
+#include "include/cephfs/statx.h"
+
 #if HAVE_GETGROUPLIST
 #include <grp.h>
 #include <pwd.h>
@@ -6609,6 +6611,28 @@ int Client::stat(const char *relpath, struct stat *stbuf,
   return r;
 }
 
+int Client::statx(const char *relpath, unsigned int flags, struct statx *stx,
+			  frag_info_t *dirstat, int mask)
+{
+  ldout(cct, 3) << "statx enter (relpath " << relpath << " mask " << mask << ")" << dendl;
+  Mutex::Locker lock(client_lock);
+  tout(cct) << "statx" << std::endl;
+  tout(cct) << relpath << std::endl;
+  filepath path(relpath);
+  InodeRef in;
+  int r = path_walk(path, &in);
+  if (r < 0)
+    return r;
+  r = _getattr(in, mask);
+  if (r < 0) {
+    ldout(cct, 3) << "statx exit on error!" << dendl;
+    return r;
+  }
+  fill_statx(in, stx, dirstat);
+  ldout(cct, 3) << "statx exit (relpath " << relpath << " mask " << mask << ")" << dendl;
+  return r;
+}
+
 int Client::lstat(const char *relpath, struct stat *stbuf,
 			  frag_info_t *dirstat, int mask)
 {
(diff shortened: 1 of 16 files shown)


src/client/Client.cc
```diff
@@ -9669,6 +9746,27 @@ int Client::ll_getattr(Inode *in, struct stat *attr, int uid, int gid)
...
+  Mutex::Locker lock(client_lock);
+
+  int res = _ll_getattr(in, uid, gid);
+  vinodeno_t vino = _get_vino(in);
+
+  /* special case for dotdot (..) */
+  if (vino.ino.val == CEPH_INO_DOTDOT) {
+    stx->stx_mode = S_IFDIR | 0755;
+    stx->stx_nlink = 2;
+    stx->stx_mode = STATX_MODE | STATX_NLINK;
```
jtlayton: Oof, already spotted a bug. That should be the stx_mask...

jtlayton:
Ok, latest PR seems to work properly. Test passes correctly now. The main question at this point is how to best handle the include files so that the interface is usable in programs. Is this the right approach for them?

src/mds/mdstypes.cc
```diff
@@ -285,12 +286,15 @@ void inode_t::encode(bufferlist &bl, uint64_t features) const
   ::encode(last_scrub_version, bl);
   ::encode(last_scrub_stamp, bl);
 
+  if (features & CEPH_FEATURE_FS_BTIME)
+    ::encode(btime, bl);
```
gregsfortytwo: If we're doing conditional encode of the btime (based on features), we should set the encode version conditionally as well.
And actually, since the other end might be say an OSD which doesn't care, we should just encode unconditionally (especially since we're not forcing them to read the btime).

src/include/ceph_features.h
```diff
@@ -84,6 +84,7 @@
 // duplicated since it was introduced at the same time as CEPH_FEATURE_CRUSH_TUNABLES5
 #define CEPH_FEATURE_NEW_OSDOPREPLY_ENCODING (1ULL<<58) /* New, v7 encoding */
 #define CEPH_FEATURE_FS_FILE_LAYOUT_V2       (1ULL<<58) /* file_layout_t */
+#define CEPH_FEATURE_FS_BTIME                (1ULL<<59) /* btime */
```
gregsfortytwo: I know @athanatos has done some work on feature bit reclamation; we should ask him or @liewegas prior to merging something that uses one of our last few.

Actually we might be better off to put btime support in the MDSMap CompatSet stuff, since it should really be a cluster thing. And the admin can decide if they want to let in clients which don't support it or not. Hmm...

src/test/libcephfs/test.cc
```diff
@@ -1374,3 +1374,42 @@ TEST(LibCephFS, OpenNoClose) {
...
+  ASSERT_EQ(0, ceph_conf_parse_env(cmount, NULL));
+  ASSERT_EQ(ceph_mount(cmount, "/"), 0);
+  int64_t old_btime;
+  int32_t old_btime_ns;
+
+
+  char filename[32];
+  sprintf(filename, "/getattrx%x", getpid());
+
+  /* FIXME: check for errors here */
```
gregsfortytwo: Heh, FIXME. ;)

gregsfortytwo:
@jtlayton, what can we do to avoid breaking the ABI?

The shape of this certainly looks good so far.

gregsfortytwo:
Hmm, looks like you're not in the Ceph group yet @jtlayton, is that right? (I can't assign the PR back to you for comments.)

src/mds/mdstypes.cc
jtlayton: Would an OSD ever even request anything here? This is an MDS call so I wouldn't think an OSD would ever see this reply. Also, if we encode unconditionally, won't that throw off the alignment of trailing info in the receiver? How do clients that don't have the btime patches know to skip over the btime in the stream?

src/test/libcephfs/test.cc
jtlayton: Yeah, though TBH we don't really care whether that succeeds or fails. Either it succeeds and the file is removed, or it doesn't and the file isn't. What really matters is whether the O_CREAT open afterward succeeds. We can probably just remove that comment.

jtlayton:
As far as ABI goes, I'm not sure how that works in C++ land...

We are adding new members to class Inode, for instance, so any program compiled against libcephfs.so is going to need to deal with the fact that the alignment of the trailing structure members has changed, isn't it? We could add them to the end, but even that might not be sufficient if the length of the inode matters in any way. Maybe we just need to make sure that we bump the library versions?

src/include/ceph_features.h
jtlayton: I'm fine with either combining this flag with other features added at the same time or indiciating it in some other way. Just point me in the right direction for it...

src/include/cephfs/statx.h
```diff
@@ -0,0 +1,79 @@
+// -*- mode:C++; tab-width:8; c-basic-offset:2; indent-tabs-mode:t -*-
+// vim: ts=8 sw=2 smarttab
+/*
+ * scalable distributed file system
+ *
+ * Copyright (C) 2009-2011 New Dream Network
```
jtlayton: Probably need to fix the copyright statement here. We should probably also add a blurb crediting David Howells for the original statx work.

src/include/ceph_features.h
liewegas: It's just a waiting game.  We'll defer any final merge until there is 
another feature or two to combine with.

src/mds/mdstypes.cc
gregsfortytwo: I'd have to check the call chain to figure out what features we pass down when. Certainly an OSD will never be looking at these structures directly, but we encode them when writing out to omaps for storage on an OSD.

gregsfortytwo:
Yeah, I don't think we should worry about the ABI. librados is in much wider use and I understand it makes breaking changes.

src/include/cephfs/statx.h
```diff
@@ -0,0 +1,79 @@
...
+	uint32_t	stx_nlink;
+	uint32_t	stx_gen;
+	uint32_t	stx_uid;
+	uint32_t	stx_gid;
+	uint16_t	stx_mode;
+	uint16_t	__spare0[1];
+	uint64_t	stx_ino;
+	uint64_t	stx_size;
+	uint64_t	stx_blocks;
+	uint64_t	stx_version;
```
ffilz: Any reason for not using timespec for atime etc?

src/include/cephfs/statx.h
```diff
@@ -0,0 +1,79 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/*
+ * Since the xstat patches are still a work in progress, we borrow its data
+ * structures and #defines to implement ceph_getattrx. Once the xstat stuff
+ * has been merged we should drop this and switch over to using that instead.
+ */
```
ffilz: Oh, sorry, missed that this struct was borrowed from in progress kernel patches...

src/include/cephfs/statx.h
ffilz: Which leads me to ask, is there any way ceph can provide stx_version (I'm assuming that's equivalent to the version field in the kernel inode knfsd uses for the NFS v4 change attribute)?

jtlayton: I think timespec definitions vary quite widely with word size.

I'm happy to look at populating stx_version once the interface is merged.

jtlayton:
Updated PR to address Greg's comments. Encoding of the btime is done unconditionally now, and I cleaned up a few other nits that he noted.

gregsfortytwo:
LGTM. @jtlayton, presumably you've done some testing you can describe?

Oh, assuming that we're all prepared for a very lazy statx. ;) We will want to do it for real once the interface exists!

jtlayton:
I haven't plumbed in "lazy" statx yet, but it wouldn't be hard to do. Just skip the _getattr after doing the lookup. The only problematic part is that the proposed statx syscall requests that behavior via new AT_* flags, and there is some debate about what that should look like in the final draft.

jtlayton:
Latest set does some cleanup of the interface (a'la more consistent variable names and better doc headers). Fix the usage of the "flags" field in statx calls. Add support for "lazy" statx (a'la AT_NO_ATTR_SYNC) and a test for that as well.

liewegas:
FWIW the lazy bit should be trivial, as our internal stat already takes a 
mask.  statx has been a long time coming...

One semi-related gap, BTW, is that we never invalidate the caps when the 
rstats change, and there isn't currently a way to 'force' a stat to get 
fresh rstat values.  Not really sure how this should be handled.

jtlayton:
Fix minor typo in header file comment.

The latest set adds AT_NO_ATTR_SYNC support (aka "lazy" mode), by simply returning whatever is in the inode instead of calling _getattr or _ll_getattr to do it. If you feel that we ought to do something more elaborate there, then please do point it out.

src/client/Client.cc
```diff
@@ -9644,28 +9644,34 @@ Inode *Client::ll_get_inode(vinodeno_t vino)
...
+
+  int res = _ll_getattr(in, uid, gid);
+  vinodeno_t vino = _get_vino(in);
+
   /* special case for dotdot (..) */
   if (vino.ino.val == CEPH_INO_DOTDOT) {
     attr->st_mode = S_IFDIR | 0755;
     attr->st_nlink = 2;
     return 0;
   }
```
gregsfortytwo: Are we okay doing this check after _getattr()? Since it can trigger an MDS request?

gregsfortytwo:
Does statx *still* not provide the interface for specifying which pieces of data to it wants to be up-to-date? if it is available, we can be a lot more interesting than "server or not"; we can compare what it wants with what caps we currently hold to determine if we can satisfy all the "important" pieces out of cache.

src/client/Client.cc
jtlayton: Good point. It would be better to move that afterward. I'll see if I can fix that in a follow-on patch.

jtlayton:
Well, you can set values in the stx_mask before making the call. That tells the call what you're actually interested in. You won't necessarily get all of those back (and you may very well get more back than you requested).

The current interface does not allow you to say "well, I want the very latest size, but the mtime can come out of the cache". If you think that'll be desirable by applications then we probably should hold off on this and rework the interface into something more ceph-specific.

gregsfortytwo:
> Well, you can set values in the stx_mask before making the call. That tells the call what you're actually interested in.

Right, so we should probably look at the stx_mask when deciding whether to make a server request or not, based on whether we have caps for the specific pieces they're requesting?

jtlayton:
We certainly can do that within the confines of the interface. Let me look into it.

liewegas:
Note that _getattr in CLient.cc already makes this judgement, see the mask 
arg here:

	https://github.com/ceph/ceph/blob/master/src/client/Client.cc#L6354

I suspect/hope this is just a matter of translating stx_mask into our mask 
type (see CEPH_SETATTR_* at

	https://github.com/ceph/ceph/blob/master/src/include/ceph_fs.h#L373

).

jcsp:
This branch currently fails to build RPMs (complaining about unpackaged statx.h)

jtlayton:
Yeah, I'm working on a respin of this, with a larger scope. I think I'll go ahead and decline this pull request for now, and open a new one once it's a little more ready.

jtlayton:
Erm, and to be clear...the client just passes the getattr mask parm through to the MDS, AFAICT, which then uses that mask to determine what caps to acquire. It always issues the request to the MDS when you call _getattr.

gregsfortytwo:
@jtlayton, you're missing the part where it says ```return 0;``` if ```in->caps_issued_mask(mask)``` returns true. That's the client deciding if it has sufficient caps to satisfy the request out of cache, or if it has to go read straight from MDS.

PR #9968, opened 2016-06-27
rpm: Do not start targets on update

b-ranto:
We do not want to start the targets on upgrade, this would override the
user configuration and is a bad practice. This commit fixes the
behaviour.

### DIFF
--- a/ceph.spec.in
+++ b/ceph.spec.in
@@ -797,7 +797,7 @@ rm -rf %{buildroot}
 /sbin/ldconfig
 %if 0%{?suse_version}
 %fillup_only
-if [ $1 -ge 1 ] ; then
+if [ $1 -eq 1 ] ; then
   /usr/bin/systemctl preset ceph.target >/dev/null 2>&1 || :
 fi
 %endif
@@ -924,7 +924,7 @@ fi
 
 %post mds
 %if 0%{?suse_version}
-if [ $1 -ge 1 ] ; then
+if [ $1 -eq 1 ] ; then
   /usr/bin/systemctl preset ceph-mds@\*.service ceph-mds.target >/dev/null 2>&1 || :
 fi
 %endif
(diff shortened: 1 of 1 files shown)


b-ranto:
@smithfarm I also noticed that in the suse-only part of the script you use 'if [ $1 -ge q ]'. That would preset the state of the targets also on update. Are you sure you want to do that there?

Also, care to review? ( @ktdreyer: same question :-) )

smithfarm:
LGTM if gitbuilders are OK with it.

@tserong What do you think?

smithfarm:
Oddly, centos7 gitbuilder [1] failed because

```
++ grep '^Version:' ceph.spec
++ awk '{print $2}'
grep: ceph.spec: No such file or directory 
```

[1] http://gitbuilder.sepia.ceph.com/gitbuilder-ceph-rpm-centos7-amd64-basic/log.cgi?log=0efbb5ae8d239af5c7c6e4f9b4374dfb1aa59c51

b-ranto:
@smithfarm the gitbuilder seems to be broken (actually both centos 7 builders looks like they are broken). All new builds seem to fail and yesterday, it took the gitbuilder several (>8) hours to even initiate a build for me.

tserong:
@smithfarm, do those presets need to be there on upgrade to handle the case where upgrading from an earlier version that doesn't have the ceph-{mon,mds,osd}.target files? (I think https://bugzilla.suse.com/show_bug.cgi?id=972369 is the downstream bug in that area)

smithfarm:
@tserong No, I don't think so. The argument passed to the `%post` scriptlet is the number of instances of the package that are installed. Now, let's assume this PR is merged. Now, when e.g. hammer is upgraded to jewel, the argument to the `%post` scriptlet of the "ceph" package is 2 and the preset is not run for that package - but that doesn't matter because `ceph.target` is already enabled. The argument to the other packages' `%post` scriptlets will be 1 (because they are newly installed packages) and the preset fires for those.

tserong:
@smithfarm, ah, gotcha. In that case, LGTM.

smithfarm:
FWIW, here's the bug report explaining why the centos gitbuilders are suddenly not creating `ceph.spec` http://tracker.ceph.com/issues/16501

PR #9969, opened 2016-06-27
include/assert: clean up ceph assertion macros

liewegas:


liewegas:
test this please

src/include/assert.h
```diff
@@ -129,6 +129,13 @@ using namespace ceph;
    ? __CEPH_ASSERT_VOID_CAST (0)					\
    : __ceph_assert_fail (__STRING(expr), __FILE__, __LINE__, __CEPH_ASSERT_FUNCTION))
 
+// this variant will *never* get compiled out to NDEBUG in the future.
+// (ceph_assert currently doesn't either, but in the future it might.)
```
gregsfortytwo: I'm not very familiar with how our current asserts work with debug that lets them be disabled or not. Can we do something to make sure the tooling is set up properly right now, which guarantees ceph_assert_always really does get compiled in?

gregsfortytwo:
Basically looks fine.

src/include/assert.h
liewegas: Our aren't conditional (yet)... It's the lack of an #ifdef NDEBUG where this comment goes that prevents it from compiling out.  And/or the lack of a similar comment says "do not ever compile this out" next to the ceph_assert definition a few lines up.

tchaikov:
tested at 

PR #9970, opened 2016-06-27
rbd-mirror: gracefully handle being blacklisted

dillaman:


trociny:
@dillaman could you please rebase when you have spare time

dillaman:
@trociny rebase pushed

src/tools/rbd_mirror/Replayer.cc
```diff
@@ -311,7 +356,7 @@ int Replayer::init()
     // librados::Rados::conf_parse_argv
     r = cct->_conf->parse_argv(m_args);
```
trociny: @dillaman Need to pass a copy of m_args, because parse_argv removes parsed items from the vector. You can see this running rbd_mirror.sh and observing that log and asok file names are not set properly set for the remote context.

trociny:
@dillaman compilation of test/rbd_mirror/image_replay.cc fails. See jenkins logs.

dillaman:
@trociny updates pushed

trociny:
@dillaman please rebase

dillaman:
@trociny rebased and pushed

trociny:
lgtm

PR #9971, opened 2016-06-27
mds: Kill C_SaferCond in evict_sessions()

fullerdj:
MDSRankDispatcher::evict_sessions waits on a C_SaferCond for
kill_session to complete on each of its victims. Change the
command handling flow to pass command messages all the way down
to MDSRankDispatcher. Extract the MDSDaemon's reply path into a
static function callable from a new context in the MDSRankDispatcher.

See: http://tracker.ceph.com/issues/16288

### DIFF
--- a/src/mds/MDSDaemon.cc
+++ b/src/mds/MDSDaemon.cc
@@ -587,37 +587,12 @@ void MDSDaemon::tick()
   }
 }
 
-/* This function DOES put the passed message before returning*/
-void MDSDaemon::handle_command(MCommand *m)
+void MDSDaemon::send_command_reply(MCommand *m, MDSRank *mds_rank,
+				   int r, bufferlist outbl,
+				   const std::string& outs)
 {
   Session *session = static_cast<Session *>(m->get_connection()->get_priv());
   assert(session != NULL);
-
-  int r = 0;
-  cmdmap_t cmdmap;
-  std::stringstream ss;
-  std::string outs;
-  bufferlist outbl;
-  Context *run_after = NULL;
-
-
-  if (!session->auth_caps.allow_all()) {
-    dout(1) << __func__
-      << ": received command from client without `tell` capability: "
-      << m->get_connection()->peer_addr << dendl;
-
-    ss << "permission denied";
-    r = -EPERM;
-  } else if (m->cmd.empty()) {
-    ss << "no command given";
-    outs = ss.str();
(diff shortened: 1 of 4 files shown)


gregsfortytwo:
NB: make check failure is bluestore unit tests.

src/mds/MDSDaemon.cc
```diff
@@ -587,6 +587,33 @@ void MDSDaemon::tick()
   }
 }
 
+void MDSDaemon::send_command_reply(MCommand *m, MDSRank *mds_rank,
+				   int r, bufferlist outbl,
+				   std::string outs)
```
gregsfortytwo: The string should probably be a const reference?

src/mds/MDSRank.cc
```diff
@@ -2541,10 +2549,11 @@ MDSRankDispatcher::MDSRankDispatcher(
 
 bool MDSRankDispatcher::handle_command(
   const cmdmap_t &cmdmap,
-  bufferlist const &inbl,
+  MCommand *m,
   int *r,
   std::stringstream *ds,
-  std::stringstream *ss)
+  std::stringstream *ss,
+  bool *need_reply)
```
gregsfortytwo: I generally like to set these out params to a known value at function start; otherwise we rely on the caller initializing them correctly. (In which case, we need to document that contract in the header.)

gregsfortytwo:
Looks good to me; just a few nits. Since we test commands in the nightlies this should have plenty of coverage, but you did test it locally, right?

fullerdj:
Comments addressed, should I run this through a suite or something?

gregsfortytwo:
There's probably something available under vstart-runner that tests the commands. If that passes we can mark it as needs-qa and let @jcsp or somebody bundle it up in a testing branch.
(Or you can schedule one if it you like, especially if there's an empty queue!)

jcsp:
I am seeing a hang inside evict sessions when running TestVolumeClient:
http://qa-proxy.ceph.com/teuthology/jspray-2016-06-29_18:28:44-fs-wip-jcsp-testing-20160629-distro-basic-mira/283892/teuthology.log

@fullerdj can you reproduce locally with vstart_runner?

fullerdj:
Yes, I've been working on that.

fullerdj:
This resolves the hang I was seeing, but it looks like you may have encountered a different case. It looks like you are seeing a nearly identical case to http://tracker.ceph.com/issues/16288, but I have never been able to reproduce that one.

PR #9972, opened 2016-06-27
osd: object class loading and execution permissions

noahdesu:
This adds options 'osd class load list' that is a list of classes allowed to be loaded (or * for all), and 'osd class default list' which is a list of classes (or * for all) that may be used without explicit permission.

### DIFF
--- a/src/osd/ClassHandler.cc
+++ b/src/osd/ClassHandler.cc
@@ -125,6 +125,7 @@ ClassHandler::ClassData *ClassHandler::_get_class(const string& cname,
     dout(10) << "_get_class adding new class name " << cname << " " << cls << dendl;
     cls->name = cname;
     cls->handler = this;
+    cls->whitelisted = in_class_list(cname, cct->_conf->osd_class_default_list);
   }
   return cls;
 }
--- a/src/osd/ClassHandler.cc
+++ b/src/osd/ClassHandler.cc
@@ -29,7 +29,9 @@
 int ClassHandler::open_class(const string& cname, ClassData **pcls)
 {
   Mutex::Locker lock(mutex);
-  ClassData *cls = _get_class(cname);
+  ClassData *cls = _get_class(cname, true);
+  if (!cls)
+    return -EPERM;
   if (cls->status != ClassData::CLASS_OPEN) {
     int r = _load_class(cls);
     if (r)
@@ -60,8 +62,9 @@ int ClassHandler::open_all_classes()
       cname[strlen(cname) - (sizeof(CLS_SUFFIX) - 1)] = '\0';
       dout(10) << __func__ << " found " << cname << dendl;
       ClassData *cls;
+      // skip classes that aren't in 'osd class load list'
       r = open_class(cname, &cls);
-      if (r < 0)
+      if (r < 0 && r != -EPERM)
 	goto out;
     }
   }
@@ -80,7 +83,33 @@ void ClassHandler::shutdown()
   classes.clear();
 }
 
-ClassHandler::ClassData *ClassHandler::_get_class(const string& cname)
+/*
+ * Check if @cname is in the whitespace delimited list @list, or the @list
+ * contains the wildcard "*".
+ *
+ * This is expensive but doesn't consume memory for an index, and is performed
+ * only once when a class is loaded.
+ */
+bool ClassHandler::in_class_list(const std::string& cname,
+    const std::string& list)
+{
+  std::istringstream ss(list);
+  std::istream_iterator<std::string> begin{ss};
+  std::istream_iterator<std::string> end{};
+
+  const std::vector<std::string> targets{cname, "*"};
+
+  auto it = std::find_first_of(begin, end,
+      targets.begin(), targets.end());
+
+  if (it == end)
+    return false;
+
+  return true;
+}
+
+ClassHandler::ClassData *ClassHandler::_get_class(const string& cname,
+    bool check_allowed)
 {
   ClassData *cls;
   map<string, ClassData>::iterator iter = classes.find(cname);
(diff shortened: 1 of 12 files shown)


src/test/osd/osdcap.cc
```diff
@@ -270,6 +270,11 @@ TEST(OSDCap, BasicRW) {
   ASSERT_FALSE(cap.is_capable("bar", "", 0, "foo", false, true, false, true));
   ASSERT_FALSE(cap.is_capable("bar", "", 0, "foo", true, false, true, false));
   ASSERT_FALSE(cap.is_capable("bar", "", 0, "foo", true, true, true, true));
+
+  // class whitelist flag works
+  ASSERT_TRUE(cap.is_capable("bar", "", 0, "foo", false, true, false, true, true));
+  ASSERT_TRUE(cap.is_capable("bar", "", 0, "foo", true, false, true, false, true));
+  ASSERT_TRUE(cap.is_capable("bar", "", 0, "foo", true, true, true, true, true));
```
liewegas: This doesn't look right... did you try runing this unit test?

The "allow rw" behavior shouldn't change.  But the "allow x" case should change.. where existing test cases where whitelisting is false should go from true -> false, and new cases should only be true if whitelist is true...

liewegas:
The change looks right except for the unit test.  I think we'll want a whole slew of cases for everywhere either of op_may_class_{read,write} is true with the whitelist false and true cases, where a false changes the is_capable result from true -> false.

src/test/osd/osdcap.cc
noahdesu: My understanding was that if a class was whitelisted, then irregardless of having 'allow x' it should be permitted to call into an object class. So this test is supposed to show that with only 'allow rw' but a whitelisted class (last parameter: true) then may_class_read and may_class_write are OK. Have I got these semantics messed up?

liewegas: Yeah, we want something a bit different:

1. if you have 'allow x' or 'class-read' or 'class-write', then you can 
execute any method on any *whitelisted* class.  We'll whitelist existing 
stuff by, so there is effectively no net change here (no new permissions 
and nothing is lost).

2. if you have 'allow class foo', then you can use class foo regardless of 
whether it is whitelisted or not (provided the osd is loading it). 

@athanatos right?

athanatos: That seems to make sense.

noahdesu: does 'allow *' override the whitelist?

liewegas: yeah, 'allow *' is like root.

noahdesu:
I've pushed up v2 for review. This has 'allow *' allowed to call any object class method, enforces whitelist, and supports the 'allow class foo' format. I updated the tests per your suggestion, and added new tests for the 'allow class foo' cases.

noahdesu:
It just occurred to me that this patch doesn't handle loading class libraries that are not listed in 'osd class load list' unless it is set to `*`. Should we use 'allow class foo' to permit loading of non-default classes?

liewegas:
It seems okay to me to require the admin to either load * or list it 
explicitly.

noahdesu:
Cool. That makes sense.

liewegas:
This looks good to me!  @athanatos want to take a final look?

src/osd/PG.cc
```diff
@@ -1870,7 +1870,9 @@ bool PG::op_has_sufficient_caps(OpRequestRef& op)
 			     op->need_read_cap(),
 			     op->need_write_cap(),
 			     op->need_class_read_cap(),
-			     op->need_class_write_cap());
+			     op->need_class_write_cap(),
+			     op->class_name(),
+			     op->class_whitelisted());
```
athanatos: An op could invoke two different classes, right?

athanatos:
A librados op can have more than one class operation, right?  The OpRequestRef machinery should reflect that.

src/osd/PG.cc
noahdesu: Oh I had only considered the case of classes calling other classes in which case they were transitively permitted. Are you referring to using multiple ObjectOperation::exec?

athanatos: The librados ObjectOperation api lets you perform multiple execs in the same Operation.

noahdesu:
@athanatos ok thanks for catching that problem. i pushed a new version for review that handles multiple execs in the same operation (i didn't squash... seems easier for reviews to not?).

athanatos:
Ok, I think that looks right!

noahdesu:
@tchaikov i was just getting ready to squash the top 3 commits, but it looks like `wip-kefu-testing` might have this branch close to merging? Would it be easier for you to sqash them?

tchaikov:
@noahdesu as long as the squashed code is identical to the one before squashing, they are the same to me. =)

noahdesu:
@tchaikov i pushed up the new version with the top 3 commits squashed. code is identical to before.

tchaikov:
tested at 
- http://pulpito.ceph.com/kchai-2016-07-12_11:14:46-rados-wip-kefu-testing2---basic-mira/
- http://pulpito.ceph.com/kchai-2016-07-12_11:14:46-rados-wip-kefu-testing2---basic-mira/

the failed tests are addressed by #10306.

src/osd/ClassHandler.h
```diff
@@ -101,9 +101,12 @@ class ClassHandler
   Mutex mutex;
   map<string, ClassData> classes;
 
-  ClassData *_get_class(const string& cname);
+  ClassData *_get_class(const string& cname, bool check_allowed);
   int _load_class(ClassData *cls);
 
+  bool in_class_list(const std::string& cname,
```
tchaikov: nit, could be a static member function.

src/osd/ClassHandler.cc
```diff
@@ -80,14 +83,44 @@ void ClassHandler::shutdown()
...
+  std::istringstream ss(list);
+  std::istream_iterator<std::string> begin{ss};
+  std::istream_iterator<std::string> end{};
+
+  const std::vector<std::string> targets{cname, "*"};
+
+  auto it = std::find_first_of(begin, end,
+      targets.begin(), targets.end());
+
+  if (it == end)
```
tchaikov: could just `return it != end`.

tchaikov:
@noahdesu besides some nits, i think it might be worthy to have this features in the PendingReleaseNotes. but i don't want to block you, i am merging it, since sage and sam have given your their blessings.

noahdesu:
@tchaikov i will submit separate PRs for pending release notes and nits. They aren't blockers. Thanks.

PR #9978, opened 2016-06-28
make-dist: create ceph.spec from ceph.ceph.in

tchaikov:
unlike the one created by configure and "make dist", the @VERSION@ in
ceph.spec comes from "git describe" not AC_INIT() or a variable in
configure script. i think this should be fine, as the official release
should always have a most recent tag attached to it, which will be
the same as the one in the configure.ac or the one in set(VERSION "10.2.1")
in CMakeLists.txt. but in the development branch, the VERSION is a
little bit older than the tip of the latest branch, this is okay also.
as we don't release directly from the master branch.

### DIFF
--- a/make-dist
+++ b/make-dist
@@ -11,6 +11,12 @@ outfile="ceph-$version"
 
 echo "version $version"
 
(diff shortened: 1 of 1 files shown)


tchaikov:
@dmick i fix it in a different way.

PR #9981, opened 2016-06-28
cls_rbd: set omap values in batch during image creation

yangdongsheng:


dillaman:
lgtm

PR #9983, opened 2016-06-28
cmake: Fix FCGI include directory

tserong:
config-h.in.cmake and rgw_fcgi.{cc,h} all expect FASTCGI_INCLUDE_DIR to
be set when FastCGI headers are in /usr/include/fastcgi.  Unfortunately,
Findfcgi.cmake sets FCGI_INCLUDE_DIR, i.e. the name doesn't match.  This
breaks the cmake build on SUSE distros (and presumably any other distro
which has the FastCGI headers in /usr/include/fastcgi).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -516,7 +516,7 @@ AS_IF([test "x$with_radosgw" != xno],
               [AC_CHECK_LIB([curl], [curl_easy_init],
                [RADOSGW=1
 	        AC_CHECK_HEADER([fastcgi/fcgiapp.h],
-		 [AC_DEFINE([FASTCGI_INCLUDE_DIR], [1], [FastCGI headers are in /usr/include/fastcgi])])
+		 [RGW_CXXFLAGS="-I/usr/include/fastcgi"])
 	       ],
 	       [if test "x$with_radosgw" != "xcheck"; then
 		    AC_MSG_FAILURE([--with-radosgw was given but libcurl (libcurl-dev on debian) not found])
@@ -529,6 +529,7 @@ AS_IF([test "x$with_radosgw" != xno],
 	     [if test "x$with_radosgw" != "xcheck"; then
 		AC_MSG_FAILURE([--with-radosgw was given but libfcgi (libfcgi-dev on debian) not found])
 	     fi])])
+AC_SUBST(RGW_CXXFLAGS)
 AM_CONDITIONAL(WITH_RADOSGW, test "$RADOSGW" = "1")
 
 AS_IF([test "$RADOSGW" = "1"], [AC_DEFINE([WITH_RADOSGW], [1], [define if radosgw enabled])])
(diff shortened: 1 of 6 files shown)


tchaikov:
quote from `rgw_fcgi.cc`

```c++
#include "acconfig.h"
#ifdef FASTCGI_INCLUDE_DIR
# include "fastcgi/fcgiapp.h"
#else
# include "fcgiapp.h"
#endif
```

and `Findfcgi.cmake` searches for `fcgio.h` in *both* `/usr/include` and `/usr/include/fastcgi`. so 

- if fastcgi header files are installed into `/usr/include/fastcgi`, `FCGI_INCLUDE_DIR` is set, and `#include "fcgiapp.h` is chosen. with this patch, `# include "fastcgi/fcgiapp.h"` is chosen, good.
- if fastcgi header files are installed into `/usr/include`, `FCGI_INCLUDE_DIR` is set, and `#include "fcgiapp.h` is chosen. good. with this patch, `# include "fastcgi/fcgiapp.h"` is chosen, d'oh!

so maybe a better approach is to 

1. always `#include <fcgiapp.h>` in `rgw_fcgi.cc`, please note that we have `find_package(fcgi REQUIRED)` in `CMakeLists.txt`
1. update `src/rgw/CMakeLists.txt`, to add
```
include_directories(${FCGI_INCLUDE_DIR})
```
into it.

tserong:
@tchaikov, ah, thanks for the explanation! I'll try alternate approach... Although, how about this:

````
#ifdef FASTCGI_INCLUDE_DIR
# include "fastcgi/fcgiapp.h"
#else
# include <fcgiapp.h>
#endif
````

That means autotools builds will still pick up the FASTCGI_INCLUDE_DIR branch if necessary, and everything else will use the `<fcgiapp.h>` form.

tchaikov:
so you suggest we keep `FCGI_INCLUDE_DIR` unchanged in the `Findfcgi.cmake`? if that's the case, it works albeit a little bit tricky. personally, i don't really like this way, even it involves less change.

tserong:
Yeah, leave `FCGI_INCLUDE_DIR` unchanged in `Findfcgi.cmake`, add `include_directories(${FCGI_INCLUDE_DIR})` to `src/rgw/CMakeLists.txt` as you suggested earlier, and put `#include <fcgiapp.h>` in the `#else` case in `rgw_fcgi.cc`. That should be the minimal change to fix the cmake build.

If I get rid of the `#ifdef FASTCGI_INCLUDE_DIR` section and always make it `#include <fcgiapp.h>`, that probably breaks the autotools build. But if we don't care about that anymore, then that's actually not a problem :-)

tchaikov:
you are right. the change i suggested breaks autotools build. the autotools part needs more work if we remove `#ifdef FASTCGI_INCLUDE_DIR`.

tchaikov:
lgtm.

tchaikov:
tested at 

PR #9988, opened 2016-06-28
os/bluestore: get rid off blob's ref_map for non-shared objects

ifed01:
Ref_map is built on the fly for onodes that are not shared. Hence provides saving for both serialized and in-memory onode representation.

Signed-off-by: Igor Fedotov ifedotov@mirantis.com

### DIFF
--- a/src/os/bluestore/bluestore_types.h
+++ b/src/os/bluestore/bluestore_types.h
@@ -630,13 +630,13 @@ struct bluestore_onode_t {
 
   /// punch a logical hole.  add lextents to deref to target list.
   void punch_hole(uint64_t offset, uint64_t length,
-		  vector<bluestore_lextent_t> *deref);
+		  vector<std::pair<uint64_t, bluestore_lextent_t> >*deref);
 
   /// put new lextent into lextent_map overwriting existing ones if any and update references accordingly
   void set_lextent(uint64_t offset,
 		   const bluestore_lextent_t& lext,
 		   bluestore_blob_t* b,
-		   vector<bluestore_lextent_t> *deref);
+		   vector<std::pair<uint64_t, bluestore_lextent_t> >*deref);
 
   /// post process removed lextent to take care of blob references
   /// returns true is underlying blob has to be released
--- a/src/os/bluestore/bluestore_types.h
+++ b/src/os/bluestore/bluestore_types.h
@@ -120,6 +120,14 @@ struct bluestore_extent_ref_map_t {
     return ref_map.empty();
   }
 
+  //raw reference insertion that assumes no conflicts/interference with the existing references
+  void fill(uint32_t offset, uint32_t len, int refs = 1) {
+    auto p = ref_map.insert(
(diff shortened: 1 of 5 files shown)


src/os/bluestore/bluestore_types.h
```diff
@@ -196,6 +196,7 @@ struct bluestore_blob_t {
   }
 
   vector<bluestore_pextent_t> extents;///< raw data position on device
+  uint32_t compressed_length_orig = 0;///< original length of compressed blob if any
```
liewegas: We could just call this 'length'...

src/os/bluestore/bluestore_types.h
```diff
@@ -120,6 +120,15 @@ struct bluestore_extent_ref_map_t {
     return ref_map.empty();
   }
 
+  //raw reference insertion that assumes no conflicts/interference with the existing references
+  void fill(uint32_t offset, uint32_t len, int refs = 1) {
+    assert(!intersects(offset,len));
```
liewegas: let's drop this assert (as we don't compile it out)

liewegas:
This looks good to me!  It needs a rebase, though.

liewegas:
I'm not wild about having to calculate a ref map before we can figure out which refs we are discarding, but I don't see a better way.  Perhaps in the future we can keep the (up to date) ref_map in the Blob so that we only generate it on the first put after the Onode is loaded off disk.  (Can do that later though..)

src/os/bluestore/bluestore_types.h
ifed01: I'd prefer to highlight that this member is applied to compressed blobs only. And we ensure its' persistence in this case only. 'length' hides such semantic IMHO.

ifed01:
@liewegas: Fixed and rebased

markhpc:
Hi Igor, does this pass "ceph_test_objectstore  --gtest_filter=*/2"?

ifed01:
@markhpc, just rerun the object store test suite - everything is OK

markhpc:
Adding comment to see if I can get jenkins to rerun

dmick:
Retest this please

markhpc:
Hi Igor, it looks like this is failing one of the unit tests:

The following tests FAILED:
	124 - unittest_bluestore_types (OTHER_FAULT)

Can you please fix and update?

ifed01:
@markhpc - fixed

markhpc:
While testing this I hit a segfault during ceph_test_objectstore, but on subsequent runs I'm not able to reproduce.  This PR may not be responsible for the segfault, but I'm going to try to see if I can track down what's going on before merging.
[9988_segfault.txt](https://github.com/ceph/ceph/files/386733/9988_segfault.txt)

markhpc:
Igor mentioned that he has hit this segfault in the past as well and it seems to be related to running out of memory during the test.  Given that it has been observed before this PR and that I have not been able to reproduce the segfault in question, I'm going to merge this as it is passing all tests now.

PR #9989, opened 2016-06-28
cmake: unittest_rbd_mirror: split this test into two

tchaikov:
this fixes the segfault with cmake build.


dillaman:
@tchaikov I don't want to have to unit tests for a single application.  I haven't been able to recreate the seg fault -- but fixing it is the correct solution instead of working around it.

PR #9990, opened 2016-06-28
cmake: fix unittest_rbd_mirror failures under non-optimized builds

dillaman:


src/test/rbd_mirror/image_sync/test_mock_ImageCopyRequest.cc
```diff
@@ -31,11 +42,11 @@ namespace mirror {
 namespace image_sync {
 
 template <>
-struct ObjectCopyRequest<librbd::MockImageCtx> {
+struct ObjectCopyRequest<librbd::MockTestImageCtx> {
```
tchaikov: oh, i got what you meant. thought we needed to have `ObjectCopyRequest<librbd:: MockImageCtx >`

tchaikov:
looks great to me.

PR #9781, opened 2016-06-17
msg/async: harden error logic handle

yuyuyu101:
This pr don't fix any existing bug. Only fill up the potential error handle leak.

PR #9992, opened 2016-06-29
msg/async: add "queue_reset" argument

yuyuyu101:


yuyuyu101:
fix rebase conflicting introduced(https://github.com/ceph/ceph/pull/9781)

xiexingguo:
How about this(so we don't have the change the default behaviour of  `stop` too much):

```c++
-  void stop(bool queue_reset) {
+  void stop(bool queue_reset = true) {
     lock.Lock();
     bool need_queue_reset = (state != STATE_CLOSED) && queue_reset;

yuyuyu101:
... there is no much difference since it's a internal interface. if possible we should be prefer avoid default argument value

xiexingguo:
Ok, suit yourself.

tchaikov:
lgtm.

PR #9997, opened 2016-06-29
jewel: librados,osd: bad flags can crash the osd

dachary:
http://tracker.ceph.com/issues/16431


dachary:
@athanatos does this backport look good to merge ? It passed the rados suite http://tracker.ceph.com/issues/16344#note-13

athanatos:
lgtm

PR #10001, opened 2016-06-29
jewel: prepare_pgtemp needs to only update up_thru if newer than the existing one

dachary:
http://tracker.ceph.com/issues/16427


dachary:
@athanatos does this backport look good to merge ? It passed the rados suite http://tracker.ceph.com/issues/16344#note-13

athanatos:
lgtm

athanatos:
@dachary btw, for merging backports, /2000 is probably too few tests, I'd like to see at least 300 tests or so run.

PR #10005, opened 2016-06-29
Changes for Inotable repair during Forward scrub 

vishalkanaujia:


src/mds/CInode.cc
```diff
@@ -3837,6 +3839,25 @@ void CInode::validate_disk_state(CInode::validated_data *results,
...
+	      std::stringstream ds;
+	      inode_t& inode = in->inode;
+	      JSONFormatter *f = new JSONFormatter();
+              MDCache *mdcache = in->mdcache;
+
+	      mdcache->mds->inotable->dump(f);
+	      f->flush(ds);
+	      dout(0) << "scrub: inotable: adding ino = " << inode.ino <<dendl;
+	      dout(0) << "scrub: inotable status: before = "<< ds.str() << dendl;
+	      mdcache->mds->inotable->repair_inotable(inode.ino);
```
jcsp: You're returning a bool here but then ignoring it: the following save() should be conditional on the result of this

src/mds/InoTable.cc
```diff
@@ -190,3 +190,22 @@ void InoTable::generate_test_instances(list<InoTable*>& ls)
 {
   ls.push_back(new InoTable());
 }
+
+
+bool InoTable::repair_inotable(inodeno_t id)
+{
+	dout(0) << "repair inotable " << id << dendl;
+	if (free.contains(id)) {
```
jcsp: This check is backwards; repair needs doing if the inode number (in use by the inode we are checking) is marked as free.  We should skip if the inode is marked as *not* free (i.e. correct)

src/mds/InoTable.cc
```diff
@@ -190,3 +190,22 @@ void InoTable::generate_test_instances(list<InoTable*>& ls)
...
+		dout(0) << "repair_inotable: Skipping ino=" << id << "pver =" << projected_version << "ver= " << version << dendl;
+		return false;
+	}
+
+	dout(0) << "repair_inotable: adding ino" << id << "to inotable" << dendl;
+	dout(0) << "repair_inotable: before status. ino=" << id << "pver =" << projected_version << "ver= " << version << dendl;
+	free.insert(id);
```
jcsp: This is backwards, the repair action is to *remove* the inode number from the free sets

vishalkanaujia:
@jcsp Thanks for reviewing the changes. Actually, due to a mistake, I ended up creating pull request. These changes are not at all close to completion. I will update the pull request and request you to review once it is done. But, for sure, I will take care of current review comments.

PR #10008, opened 2016-06-29
jewel: ceph-base requires parted

dachary:
http://tracker.ceph.com/issues/16099


dachary:

PR #10013, opened 2016-06-29
fix install-deps.sh on openSUSE Tumbleweed.

jan--f:
Test for apt-get (install-deps.sh:60) suceeds on Tumbleweed and tries to
install nonexistent equivs package. This patch fixes this by checking
for (apt-get and NOT zypper).
Fixes http://tracker.ceph.com/issues/16522

Signed-off-by: Jan Fajerski <jfajerski@suse.com>

### DIFF
--- a/install-deps.sh
+++ b/install-deps.sh
@@ -60,7 +60,7 @@ if test -f /etc/redhat-release ; then
     $SUDO yum install -y redhat-lsb-core
 fi
 
-if type apt-get > /dev/null 2>&1 ; then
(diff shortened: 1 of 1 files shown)


theanalyst:
so we use `lsb-release` later in the same script for finding the distribution .. maybe we could use the same logic here as well?

jan--f:
The logic in question actually makes sure that lsb-release is installed.

theanalyst:
silly me.. yeah I agree

jan--f:
closed in favour of https://github.com/ceph/ceph/pull/10017

PR #10014, opened 2016-06-29
build: fix wrong indent caused compile warning

gaowanlong:


src/kv/RocksDBStore.cc
```diff
@@ -692,14 +692,14 @@ bool RocksDBStore::RocksDBWholeSpaceIteratorImpl::valid()
...
 {
   if (valid())
-  dbiter->Next();
+    dbiter->Next();
   return dbiter->status().ok() ? 0 : -1;
 }
 int RocksDBStore::RocksDBWholeSpaceIteratorImpl::prev()
 {
   if (valid())
     dbiter->Prev();
```
theanalyst: can you add {} around the if as well.. so that we're explicit

gaowanlong: @theanalyst yeah, updated now.

gaowanlong:
@theanalyst anyone can merge such really simple pr quickly?

gaowanlong:
@liewegas

PR #10019, opened 2016-06-29
librbd: add writesame api

LiumxNL:


src/include/rbd/librbd.h
```diff
@@ -569,6 +569,12 @@ CEPH_RBD_API ssize_t rbd_write(rbd_image_t image, uint64_t ofs, size_t len,
  */
 CEPH_RBD_API ssize_t rbd_write2(rbd_image_t image, uint64_t ofs, size_t len,
                                 const char *buf, int op_flags);
+
+CEPH_RBD_API ssize_t rbd_writesame(rbd_image_t image, uint64_t off, size_t write_len,
+                              const char *buf, size_t data_len);
+CEPH_RBD_API ssize_t rbd_writesame2(rbd_image_t image, uint64_t off, size_t write_len,
```
dillaman: No need to add ```writesame2``` API methods --- just add the ```op_flags``` to the new call.  The xyz2 methods are added to ensure the API doesn't break.

src/librbd/AioImageRequest.h
```diff
@@ -167,6 +169,44 @@ class AioImageWrite : public AbstractAioImageWrite {
...
+    : AbstractAioImageWrite(image_ctx, aio_comp, off, len), m_buf(buf),
+      m_data_len(data_len), m_op_flags(op_flags) {
+  }
+
+protected:
+  virtual aio_type_t get_aio_type() const {
+    return AIO_TYPE_WRITE;
+  }
+  virtual const char *get_request_type() const {
+    return "aio_write";
```
dillaman: Minor: "aio_writesame"

dillaman:
@LiumxNL Might be nice to update ceph_test_librbd_fsx to exercise the new API as well.

src/librbd/AioImageRequest.cc
```diff
@@ -382,6 +390,107 @@ void AioImageWrite::update_stats(size_t length) {
   m_image_ctx.perfcounter->inc(l_librbd_wr_bytes, length);
 }
 
+void AioImageWriteSame::assemble_extent(const ObjectExtent &object_extent,
+                                    bufferlist *bl) {
+  if (object_extent.offset % m_data_len == 0
```
dillaman: Am I reading this incorrectly or is there an assumption that the offset should always be a multiple of the data length?

dillaman:
Any thought about how to support write-same passthrough from the cache? Right now, librbd cache is enabled by default but writesame will essentially be "optimized" away as soon as cache writeback starts.

src/librbd/AioImageRequest.cc
LiumxNL: hmm..i mistake, this should be (object_extent.length % m_data_len == 0), and should make a judgement that striper_units is a multiple of data_len before assemble_extent, otherwise, the current method cannot handle。
should we handle the situation striper_units is not a multiple of data_len, it seems too complex, in block device application scenario, this could never happen？

dillaman: Hard to say since we don't enforce a block size within librbd. Plus, if "fancy" striping is enabled, we can get some odd object extent mappings. I would just update this to support the potential for substrings, odd layouts, etc.

LiumxNL:
@dillaman i am willing to complete these improvements, hope to come back soon.

yuyuyu101:
@LiumxNL

dillaman:
@LiumxNL ping -- are you still interested in working on this?

guihecheng:
@dillaman @LiumxNL We are eager for this command support, so I would like to try my best to finish this feature, is that OK?
Since Liu has already done alot , I shall work based on this pr and keep his author and sign-off.

dillaman:
@guihecheng That is fine w/ me.

LiumxNL:
@dillaman  i fell terrible sorry for laggy so much :( @guihecheng thanks.

guihecheng:
Hi @dillaman @LiumxNL 

I've made a new pr: https://github.com/ceph/ceph/pull/12645.

I changed things as stated in the pr messages, correct me if i am wrong ,Thanks.

dillaman:
Closing to PR #12645

PR #10020, opened 2016-06-29
[DNM] debian reverts

tchaikov:
these commits were merged by mistake.

very sorry for this.


tchaikov:
@mattbenjamin is rebasing his change against master. removing the the revert of  74ec4f9.

tchaikov:
the master before revert is being tested at http://pulpito.ceph.com/kchai-2016-06-29_10:07:50-rados-master---basic-smithi/

tchaikov:
maybe we can keep them, and fix the failures if any. only 1 failure is due to the packaging/cmake in the above rados qa run, and it is fixed by #10025.

liewegas:
I've lost track of the status of this stuff.. not sure if this should merge or not or when?

tchaikov:
i'd recommend not merge. let's fix the failures instead. more fixes in #10059

PR #10021, opened 2016-06-29
rgw: handle Swift auth errors in a way compatible with new Tempests.

rzarzynski:
We have to differentiate the error codes depending on whether user is
anonymous (401 Unauthorized) or he doesn't have necessary permissions
(403 Forbidden). The reason behind that is the change in Tempest. See
commit ID: 6b1cd29b763dbc556137c89c5fed54c624da7f69.

-- 

CC: @yehudasa, @cbodley.


oritwas:
@rzarzynski , can you open a tracker issue and add it to the commit

rzarzynski:
@oritwas: issue created, `Fixes` tag added.

yehudasa:
@rzarzynski lgtm, we need to get it through testing

PR #10022, opened 2016-06-29
../qa/workunits/ceph-helpers.sh: Allow TESTS to run standalone

wjwithagen:
Still falllout of the move to cmake building

### DIFF
--- a/qa/workunits/ceph-helpers.sh
+++ b/qa/workunits/ceph-helpers.sh
@@ -17,6 +17,18 @@
 # MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 # GNU Library Public License for more details.
 #
+if [ -z "$CEPH_ROOT" ] || [ -z "$CEPH_BIN" ] || [ -z "$CEPH_LIB" ]; then
+    CEPH_ROOT=`readlink -f $(dirname $0)/../..`
+    if [ -d $CEPH_ROOT/build ] ; then
+       # asume we are in Cmake build environment
+        CEPH_BIN=$CEPH_ROOT/build/bin
+        CEPH_LIB=$CEPH_ROOT/build/lib
+    else
+        CEPH_BIN=$CEPH_ROOT/src
+        CEPH_LIB=$CEPH_ROOT/.libs/src
+    fi
+fi
+
 TIMEOUT=300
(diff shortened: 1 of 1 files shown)


qa/workunits/ceph-helpers.sh
```diff
@@ -17,6 +17,15 @@
 # MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 # GNU Library Public License for more details.
 #
+if [ x"$SRC_PATH"x = xx ]; then
+  CEPH_DIR=`dirname $0`
+  CEPH_DIR=`(cd $CEPH_DIR/../..; pwd)`
```
tchaikov: could use `readlink` and `dirname`

qa/workunits/ceph-helpers.sh
```diff
@@ -17,6 +17,15 @@
...
+fi
+
+CEPH_SRC=$CEPH_DIR/src
+CEPH_BIN=$CEPH_SRC
```
tchaikov: this is not necessarily true.

wjwithagen: @tchaikov 
You'd like to account for build/bin as well?

wjwithagen: @liewegas @tchaikov 
How desperate are we that the scripts can still be executed standalone?

I can construct `CEPH_{ROOT,BIN,LIB}` relatively easy for standalone, cmake and autobuild,
(updated the commit to match that)
but getting PYTHONPATH requires (in Cmake env):
    `./build/lib/cython_modules/lib.freebsd-11.0-CURRENT-amd64-2.7/`
to find `rados.so` 
Which is sort a far fetched path. 

I see a few solutions, but one relatively nice one would be to Cmake to actually install rados.so into something like `./build/lib` or `./build/lib/pybind`  OR replace `lib` with `bin`

wjwithagen:
retest this please

wjwithagen:
@liewegas @tchaikov 
This one is lingering a long time already.
Any particulars as to why it cannot/should not be merged?

wjwithagen:
@tchaikov @liewegas 
Closing this, since it is too old. And running tests standalone on the commandline is beyond seriously broken at quite a few places.

PR #10023, opened 2016-06-29
DNM: jewel: improve OSD start/in performance

infernix:
This should improve performance when an osd comes up/in and when using do_recovery_sleep.

### DIFF
--- a/src/osd/OSD.cc
+++ b/src/osd/OSD.cc
@@ -8269,11 +8269,13 @@ void OSD::do_recovery(PG *pg, ThreadPool::TPHandle &handle)
 {
   if (g_conf->osd_recovery_sleep > 0) {
     handle.suspend_tp_timeout();
+    pg->unlock();
     utime_t t;
     t.set_from_double(g_conf->osd_recovery_sleep);
     t.sleep();
-    handle.reset_tp_timeout();
     dout(20) << __func__ << " slept for " << t << dendl;
+    pg->lock();
+    handle.reset_tp_timeout();
   }
 
   // see how many we should try to start.  note that this is a bit racy.
--- a/src/osd/PG.cc
+++ b/src/osd/PG.cc
@@ -6738,7 +6738,8 @@ boost::statechart::result PG::RecoveryState::Active::react(const ActMap&)
     pg->queue_snap_trim();
   }
 
-  if (!pg->is_clean() &&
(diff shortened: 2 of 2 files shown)


infernix:
I'm not sure if a backport of  2c50e230c1aa32b14cdc2e91ed70e6e48147343c could also be accepted, or if this can be accepted at all, but I think at least 6a3984c could?

I think it will help with reducing IO pressure when OSDs get started (we are seeing disks maxed out when starting, causing blocked ops to clients).

dachary:
This comes from https://github.com/ceph/ceph/pull/9231 but there does not seem to exist an issue associated with it. @athanatos do you think it is a good idea to backport this ?

src/osd/OSD.cc
```diff
@@ -8269,11 +8269,13 @@ void OSD::do_recovery(PG *pg, ThreadPool::TPHandle &handle)
 {
   if (g_conf->osd_recovery_sleep > 0) {
     handle.suspend_tp_timeout();
+    pg->unlock();
```
athanatos: This is wrong.  In this version of the code, we haven't locked the PG yet here, which is why jewel doesn't already unlock the pg here.  This change was only needed in master due to a change in the way the queuing worked.  Honestly, this should be removed altogether in master and is of questionable value in jewel.

src/osd/PG.cc
```diff
@@ -6738,7 +6738,8 @@ boost::statechart::result PG::RecoveryState::Active::react(const ActMap&)
     pg->queue_snap_trim();
   }
 
-  if (!pg->is_clean() &&
+  if (pg->is_peered() &&
+      !pg->is_clean() &&
```
athanatos: Unless this fixes a specific bug in jewel, there is no reason to backport this.

athanatos:
The one about unlocking the PG in do_recovery is simply not correct in jewel.  I don't think the other one is relevant for jewel either.

infernix:
The idea was really to get https://github.com/ceph/ceph/commit/2c50e230c1aa32b14cdc2e91ed70e6e48147343c backported because at least in hammer right now we have performance issues when OSDs come up/in. I take it that can't be accepted as a backport because only fixes are taken for jewel?

athanatos:
Yeah, we're not going to backport the recovery wq changes to jewel.

PR #10024, opened 2016-06-29
cmake FreeBSD changes

wjwithagen:
- Clang has different warning requirements
 - add /usr/local for include and libs
 - UDEV is not the same on FreeBSD
 - No blkid, keyutils
 - Several libs (dl, uuid, resolv) are in clib, so no need to link explicit
 - uname returns amd64
 - use EventKqueue.* in msg/async

Items excluded because porting is not done.
 - Bluestore, due to incompatible AIO
 - CephFS due to keystore dependancy??
 - RBD excluded because of incomplete porting

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,6 +19,24 @@ endif (POLICY CMP0023)
 
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake/modules/")
 
+# find out which platform we are building on
+if(CMAKE_SYSTEM_NAME MATCHES "Linux")
+  message(STATUS "Building for Linux.")
+  set(LINUX ON)
+  set(UNIX ON)
+  FIND_PACKAGE(Threads)
+elseif(CMAKE_SYSTEM_NAME MATCHES "FreeBSD")
+  message(STATUS "Building for FreeBSD.")
+  set(FREEBSD ON)
+  set(UNIX ON)
+  FIND_PACKAGE(Threads)
+elseif(CMAKE_SYSTEM_NAME MATCHES "Darwin")
+  message(STATUS "Building for Darwin.")
+  set(OperatingSystem "Mac OS X")
+else()
+  message(STATUS "No systemtype selected for building")
+endif(CMAKE_SYSTEM_NAME MATCHES "Linux")
+
 option(WITH_CCACHE "Build with ccache.")
 if(WITH_CCACHE)
   find_program(CCACHE_FOUND ccache)
@@ -54,6 +72,12 @@ link_directories(
   ${OFED_PREFIX}/lib
   ${LEVELDB_PREFIX}/lib
 )
+if(FREEBSD)
+  include_directories(/usr/local/include)
+  link_directories(/usr/local/lib)
+  list(APPEND CMAKE_REQUIRED_INCLUDES /usr/local/include)
+endif(FREEBSD)
+
 
 #put all the libs and binaries in one place
 set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
@@ -103,7 +127,12 @@ CHECK_INCLUDE_FILES("sys/types.h" HAVE_SYS_TYPES_H)
 CHECK_INCLUDE_FILES("sys/vfs.h" HAVE_SYS_VFS_H)
 CHECK_INCLUDE_FILES("sys/prctl.h" HAVE_SYS_PRCTL_H)
 CHECK_INCLUDE_FILES("execinfo.h" HAVE_EXECINFO_H)
+if(LINUX)
 CHECK_INCLUDE_FILES("sched.h" HAVE_SCHED)
+endif(LINUX)
+if(FREEBSD)
+CHECK_INCLUDE_FILES("sched.h" HAVE_SCHED_FREEBSD)
+endif(FREEBSD)
 CHECK_INCLUDE_FILES("valgrind/helgrind.h" HAVE_VALGRIND_HELGRIND_H)
 
 include(CheckTypeSize)
@@ -154,15 +183,24 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ${ENABLE_SHARED})
 
 find_package(execinfo)
 
-find_package(udev REQUIRED)
-set(HAVE_UDEV ${UDEV_FOUND})
+if(LINUX)
+  find_package(udev REQUIRED)
+  set(HAVE_UDEV ${UDEV_FOUND})
 
-option(WITH_AIO "AIO is here ON" ON)
-if(${WITH_AIO})
-find_package(aio REQUIRED)
-set(HAVE_LIBAIO ${AIO_FOUND})
-message(STATUS "${AIO_LIBS}")
-endif(${WITH_AIO})
+  find_package(aio REQUIRED)
+  set(HAVE_LIBAIO ${AIO_FOUND})
+  message(STATUS "${AIO_LIBS}")
+
+  find_package(blkid REQUIRED)
+  set(HAVE_BLKID ${BLKID_FOUND})
+else()
+  set(HAVE_UDEV OFF)
+  message(STATUS "Not using udev")
+  set(HAVE_LIBAIO OFF)
+  message(STATUS "Not using AIO")
+  set(HAVE_BLKID OFF)
+  message(STATUS "Not using BLKID")
+endif(LINUX)
 
 option(WITH_OPENLDAP "OPENLDAP is here" ON)
 if(${WITH_OPENLDAP})
@@ -189,8 +227,6 @@ if(WITH_SPDK)
   find_package(dpdk REQUIRED)
   set(HAVE_SPDK TRUE)
 endif(WITH_SPDK)
-
-find_package(blkid REQUIRED)
 
 # needs mds and? XXX
 option(WITH_LIBCEPHFS "libcephfs client library" ON)
@@ -260,7 +296,10 @@ else(ALLOCATOR)
   endif(Tcmalloc_FOUND)
 endif(ALLOCATOR)
 
-find_package(keyutils REQUIRED)
+if(NOT FREEBSD)
+  # XXX keyutils is available, but not yet recognised
+  find_package(keyutils REQUIRED)
+endif(NOT FREEBSD)
 
 find_package(libuuid REQUIRED)
 
@@ -344,7 +383,9 @@ option(HAVE_LIBZFS "LibZFS is enabled" OFF)
 option(ENABLE_COVERAGE "Coverage is enabled" OFF)
 option(PG_DEBUG_REFS "PG Ref debugging is enabled" OFF)
 
-add_definitions(-D__linux__)
+if(LINUX)
+  add_definitions(-D__linux__)
+endif(LINUX)
 
 if(ENABLE_SHARED)
   set(Boost_USE_STATIC_LIBS   OFF)
@@ -368,18 +409,6 @@ if(WITH_SELINUX)
   endif()
   add_subdirectory(selinux)
 endif(WITH_SELINUX)
-
-# find out which platform we are building on
-if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
-  set(LINUX ON)
-  set(UNIX ON)
-  FIND_PACKAGE(Threads)
-endif(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
-
-# find out which platform we are building on
-if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
-set(OperatingSystem "Mac OS X")
-endif(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
 
 # enables testing and creates Make check command
 add_custom_target(tests
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -298,7 +298,7 @@ elseif(ALLOCATOR STREQUAL "jemalloc")
   set(TCMALLOC_srcs perfglue/disabled_heap_profiler.cc)
 elseif(ALLOCATOR STREQUAL "libc")
   set(TCMALLOC_srcs perfglue/disabled_heap_profiler.cc)
-endif()
+endif(ALLOCATOR STREQUAL "tcmalloc")
 
 # tcmalloc heap profiler
 set(heap_profiler_files ${TCMALLOC_srcs})
@@ -492,7 +492,7 @@ set(libcommon_files
   common/dns_resolve.cc
   common/armor.c
   common/safe_io.c
-  common/addr_parsing.c)
+  common/addr_parsing.c
   ${arch_files}
   ${auth_files}
   $<TARGET_OBJECTS:compressor_objs>
@@ -923,9 +923,14 @@ endif(WITH_SPDK)
 if(NOT ALLOCATOR STREQUAL "jemalloc")
   set(disable_jemalloc "DISABLE_JEMALLOC=1")
 endif()
+set(EXTRA_CXXFLAGS -fPIC -Wno-unused-variable)
+if(FREEBSD)
+  set(EXTRA_CXXFLAGS ${EXTRA_CXXFLAGS} -Irocksdb/include -I/usr/local/include)
+  set(MAKEFLAGS ${MAKEFLAGS})
+endif(FREEBSD)
 add_custom_target(build_rocksdb
     COMMAND
-    PORTABLE=1 ${disable_jemalloc} $(MAKE) static_lib EXTRA_CXXFLAGS='-fPIC -Wno-unused-variable'
+    PORTABLE=1 ${disable_jemalloc} $(MAKE) ${MAKEFLAGS} static_lib EXTRA_CXXFLAGS="${EXTRA_CXXFLAGS}"
     WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/src/rocksdb
     COMMENT "rocksdb building")
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -82,7 +82,7 @@ else(no_yasm)
     COMMAND uname -m
     OUTPUT_VARIABLE arch
     OUTPUT_STRIP_TRAILING_WHITESPACE)
-  if(arch STREQUAL "x86_64")
+  if(arch MATCHES "x86_64|amd64")
     message(STATUS " we are x84_64")
     set(save_quiet ${CMAKE_REQUIRED_QUIET})
     set(CMAKE_REQUIRED_QUIET true)
@@ -110,9 +110,9 @@ else(no_yasm)
     else(not_arch_x32)
       message(STATUS " we are x32; no yasm for you")
     endif(not_arch_x32)
-  else(arch STREQUAL "x86_64")
+  else(arch MATCHES "x86_64|amd64")
     message(STATUS " we are not x86_64 && !x32")
-  endif(arch STREQUAL "x86_64")
+  endif(arch MATCHES "x86_64|amd64")
 endif(no_yasm)
 
 
@@ -169,12 +169,15 @@ try_compile(INTEL_SSE4_1 ${CMAKE_BINARY_DIR} ${sse_srcs}
   COMPILE_DEFINITIONS "-msse4.1")
 try_compile(INTEL_SSE4_2 ${CMAKE_BINARY_DIR} ${sse_srcs}
   COMPILE_DEFINITIONS "-msse4.2")
-try_compile(ARM_NEON ${CMAKE_BINARY_DIR} ${sse_srcs}
-  COMPILE_DEFINITIONS "-mfpu=neon")
-try_compile(ARM_NEON2 ${CMAKE_BINARY_DIR} ${sse_srcs}
-  COMPILE_DEFINITIONS "-march=armv8-a+simd")
-try_compile(ARM_CRC ${CMAKE_BINARY_DIR} ${sse_srcs}
-  COMPILE_DEFINITIONS "-march=armv8-a+crc")
+
+if(NOT arch MATCHES "x84_64|amd64")
+  try_compile(ARM_NEON ${CMAKE_BINARY_DIR} ${sse_srcs}
+    COMPILE_DEFINITIONS "-mfpu=neon")
+  try_compile(ARM_NEON2 ${CMAKE_BINARY_DIR} ${sse_srcs}
+    COMPILE_DEFINITIONS "-march=armv8-a+simd")
+  try_compile(ARM_CRC ${CMAKE_BINARY_DIR} ${sse_srcs}
+    COMPILE_DEFINITIONS "-march=armv8-a+crc")
+if(NOT arch MATCHES "x84_64|amd64")
 
 # clean up tmp file
 file(REMOVE ${sse_srcs})
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -17,17 +17,32 @@ set(prefix ${CMAKE_INSTALL_PREFIX})
 
 add_definitions("-DCEPH_LIBDIR=\"${CMAKE_INSTALL_FULL_LIBDIR}\"")
 add_definitions("-DCEPH_PKGLIBDIR=\"${CMAKE_INSTALL_FULL_PKGLIBDIR}\"")
-add_definitions("-DHAVE_CONFIG_H -D__CEPH__ -D_FILE_OFFSET_BITS=64 -D_REENTRANT -D_THREAD_SAFE -D__STDC_FORMAT_MACROS -D_GNU_SOURCE")
+add_definitions("-DHAVE_CONFIG_H -D__CEPH__ -D_REENTRANT -D_THREAD_SAFE -D__STDC_FORMAT_MACROS")
 
-set(CMAKE_ASM_COMPILER  ${PROJECT_SOURCE_DIR}/src/yasm-wrapper)
-set(CMAKE_ASM_FLAGS "-f elf64")
-set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -rdynamic -Wall -Wtype-limits -Wignored-qualifiers -Winit-self -Wpointer-arith -Werror=format-security -fno-strict-aliasing -fsigned-char")
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wtype-limits -Wignored-qualifiers -Winit-self") 
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wpointer-arith -Werror=format-security -fno-strict-aliasing -fsigned-char")
+
+if(CMAKE_CXX_COMPILER_ID MATCHES "GNU")
+  add_definitions("-D_GNU_SOURCE -D_FILE_OFFSET_BITS=64")
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -rdynamic")
+elseif(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  set(CMAKE_EXE_LINKER_FLAGS "-Wl,-export-dynamic")
+  set(CMAKE_LINKER_FLAGS "${CMAKE_LINKER_FLAGS} -rdynamic -Wl,-export-dynamic -export-dynamic")
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-inconsistent-missing-override")
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-mismatched-tags -Wno-unused-function")
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-unused-local-typedef -Wno-inconsistent-missing-override")
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-unused-private-field -Wno-varargs")
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-gnu-designator -Wno-mismatched-tags")
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-missing-braces -Wno-parentheses -Wno-deprecated-register")
+endif(CMAKE_CXX_COMPILER_ID MATCHES "GNU")
 
 if(NOT CMAKE_BUILD_TYPE)
-  set(CMAKE_BUILD_TYPE "RelWithDebInfo" CACHE STRING "Default BUILD_TYPE is RelWithDebInfo, other options are: Debug, Release, and MinSizeRel." FORCE) 
+  set(CMAKE_BUILD_TYPE "RelWithDebInfo" 
+	CACHE STRING "Default BUILD_TYPE is RelWithDebInfo, other options are: Debug, Release, and MinSizeRel." 
+	FORCE) 
 endif()
 
-if(NOT CMAKE_BUILD_TYPE STREQUAL Debug)
+if(NOT CMAKE_BUILD_TYPE STREQUAL "Debug")
   # we use assert(3) for testing, so scrub the -DNDEBUG defined by default
   string(TOUPPER "${CMAKE_BUILD_TYPE}" build_type_upper)
   foreach(flags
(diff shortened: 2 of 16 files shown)


wjwithagen:
@tchaikov 
Hi Kefu, Would you mind taking a look at this and suggest improvements.
Compilation is getting already pretty far, but comments on how not to do things is very welcom

CMakeLists.txt
```diff
@@ -100,7 +102,12 @@ CHECK_INCLUDE_FILES("sys/types.h" HAVE_SYS_TYPES_H)
 CHECK_INCLUDE_FILES("sys/vfs.h" HAVE_SYS_VFS_H)
 CHECK_INCLUDE_FILES("sys/prctl.h" HAVE_SYS_PRCTL_H)
 CHECK_INCLUDE_FILES("execinfo.h" HAVE_EXECINFO_H)
+if(${LINUX})
```
tchaikov: `if(LINUX)` would suffice.

CMakeLists.txt
```diff
@@ -100,7 +102,12 @@ CHECK_INCLUDE_FILES("sys/types.h" HAVE_SYS_TYPES_H)
...
 CHECK_INCLUDE_FILES("sched.h" HAVE_SCHED)
+endif(${LINUX})
+if(${FREBSD})
+CHECK_INCLUDE_FILES("sched.h" HAVE_SCHED_FREEBSD)
```
tchaikov: do we reference `HAVE_SCHED_FREEBSD` anywhere?

CMakeLists.txt
```diff
@@ -348,7 +363,12 @@ option(HAVE_LIBZFS "LibZFS is enabled" OFF)
 option(ENABLE_COVERAGE "Coverage is enabled" OFF)
 option(PG_DEBUG_REFS "PG Ref debugging is enabled" OFF)
 
+if(${LINUX})
 add_definitions(-D__linux__)
+endif(${LINUX})
+if(${FREEBSD})
```
tchaikov: nit, could use `elseif`.

src/CMakeLists.txt
```diff
@@ -21,6 +21,13 @@ add_definitions("-DHAVE_CONFIG_H -D__CEPH__ -D_FILE_OFFSET_BITS=64 -D_REENTRANT
 set(CMAKE_ASM_COMPILER  ${PROJECT_SOURCE_DIR}/src/yasm-wrapper)
 set(CMAKE_ASM_FLAGS "-f elf64")
 set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -rdynamic -Wall -Wtype-limits -Wignored-qualifiers -Winit-self -Wpointer-arith -Werror=format-security -fno-strict-aliasing -fsigned-char")
+if(__COMPILER_GCC)
```
tchaikov: should use `CMAKE_CXX_COMPILER_ID` instead, see https://cmake.org/cmake/help/v3.0/variable/CMAKE_LANG_COMPILER_ID.html#variable:CMAKE_%3CLANG%3E_COMPILER_ID

src/CMakeLists.txt
```diff
@@ -60,7 +67,7 @@ else(no_yasm)
     COMMAND uname -m
     OUTPUT_VARIABLE arch
     OUTPUT_STRIP_TRAILING_WHITESPACE)
-  if(arch STREQUAL "x86_64")
+  if((arch STREQUAL "x86_64") OR (arch STREQUAL "amd64"))
```
tchaikov: might want to use
```cmake
if(arch MATCHES "amd64|x86_64")
```

src/CMakeLists.txt
```diff
@@ -192,7 +202,13 @@ else(INTEL_SSE4_1)
 endif(INTEL_SSE4_1)
 
 
-set(EXTRALIBS uuid rt dl ${ATOMIC_OPS_LIBRARIES})
+set(EXTRALIBS rt ${ATOMIC_OPS_LIBRARIES})
+if(${LINUX})
+  set(LIB_DL dl)
+  set(LIB_RESOLV resolv)
+  set(LIB_UUID uuid)
+  set(EXTRALIBS "${EXTRALIBS} ${LIB_DL} LIB_UUID")
```
tchaikov: `s/LIB_UUID/${LIB_UUID}/` or just put 
```cmake
list(APPEND EXTRALIBS dl uuid)
```

src/CMakeLists.txt
```diff
@@ -351,6 +367,12 @@ endif(HAVE_GOOD_YASM_ELF64)
 add_library(common_texttable_obj OBJECT
   common/TextTable.cc)
 
+if(${LINUX})
```
tchaikov: i'd recommend use `list(APPEND libcommon_files)` different files for different OS.

src/CMakeLists.txt
```diff
@@ -477,13 +499,15 @@ set(mon_common_files
   auth/cephx/CephxSessionHandler.cc
   erasure-code/ErasureCodePlugin.cc)
 add_library(mon_common_objs OBJECT ${mon_common_files})
+if(${LINUX})
```
tchaikov: not sure about this. these sources should compile on FreeBSD, right?

src/CMakeLists.txt
```diff
@@ -192,7 +202,13 @@ else(INTEL_SSE4_1)
 endif(INTEL_SSE4_1)
 
 
-set(EXTRALIBS uuid rt dl ${ATOMIC_OPS_LIBRARIES})
+set(EXTRALIBS rt ${ATOMIC_OPS_LIBRARIES})
+if(${LINUX})
+  set(LIB_DL dl)
```
tchaikov: we should switch to `CMAKE_DL_LIBS`. see https://cmake.org/cmake/help/v2.8.11/cmake.html#variable:CMAKE_DL_LIBS

src/test/CMakeLists.txt
```diff
@@ -772,7 +772,7 @@ target_link_libraries(unittest_libcephfs_config cephfs)
 # unittest_rbd_replay
 add_executable(unittest_rbd_replay
   test_rbd_replay.cc
-  $<TARGET_OBJECTS:parse_secret_objs>
+#  $<TARGET_OBJECTS:parse_secret_objs>
```
tchaikov: why?

tchaikov:
would be better if you could split 0d2a9de into smaller commits.

CMakeLists.txt
wjwithagen: On 30-6-2016 11:38, Kefu Chai wrote:
> In CMakeLists.txt
> <https://github.com/ceph/ceph/pull/10024#discussion_r69102479>:
> 
>>  CHECK_INCLUDE_FILES("sched.h" HAVE_SCHED)
>> +endif(${LINUX})
>> +if(${FREBSD})
>> +CHECK_INCLUDE_FILES("sched.h" HAVE_SCHED_FREEBSD)
> 
> do we reference |HAVE_SCHED_FREEBSD| anywhere?

I am in one of the CC files, because it tried to used the Linux
scheduler routines. Which are of course not available on FreeBSD.

But FreeBSD does have sched.h, so I need to fix that.
Other solution would be to exclude the HAVE_SCHED_FREEBSD and make the
code deal with the fact that HAVE_SCHED is not defined. And then later
on when the scheduler affinitity code for FreeBSD is done, then fix the
CMakefile

But I also see a typo. FREBSD. that ain't going to work.

CMakeLists.txt
wjwithagen: On 30-6-2016 11:38, Kefu Chai wrote:
> In CMakeLists.txt
> <https://github.com/ceph/ceph/pull/10024#discussion_r69102430>:
> 
>> @@ -100,7 +102,12 @@ CHECK_INCLUDE_FILES("sys/types.h" HAVE_SYS_TYPES_H)
>>  CHECK_INCLUDE_FILES("sys/vfs.h" HAVE_SYS_VFS_H)
>>  CHECK_INCLUDE_FILES("sys/prctl.h" HAVE_SYS_PRCTL_H)
>>  CHECK_INCLUDE_FILES("execinfo.h" HAVE_EXECINFO_H)
>> +if(${LINUX})
> 
> |if(LINUX)| would suffice.

I see both variations (I think), but if this is mode of preference
I'll fix all cases of this.

src/CMakeLists.txt
wjwithagen: On 30-6-2016 11:43, Kefu Chai wrote:
> In src/CMakeLists.txt
> <https://github.com/ceph/ceph/pull/10024#discussion_r69103196>:
> 
>> @@ -21,6 +21,13 @@ add_definitions("-DHAVE_CONFIG_H -D__CEPH__ -D_FILE_OFFSET_BITS=64 -D_REENTRANT
>>  set(CMAKE_ASM_COMPILER  ${PROJECT_SOURCE_DIR}/src/yasm-wrapper)
>>  set(CMAKE_ASM_FLAGS "-f elf64")
>>  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -rdynamic -Wall -Wtype-limits -Wignored-qualifiers -Winit-self -Wpointer-arith -Werror=format-security -fno-strict-aliasing -fsigned-char")
>> +if(__COMPILER_GCC)
> 
> should use |CMAKE_CXX_COMPILER_ID| instead, see
> https://cmake.org/cmake/help/v3.0/variable/CMAKE_LANG_COMPILER_ID.html#variable:CMAKE_%3CLANG%3E_COMPILER_ID

Thanx for the pointer. The manual is rather huge, and getting a grasp on
first reading is rather hard.

src/CMakeLists.txt
wjwithagen: On 30-6-2016 11:50, Kefu Chai wrote:
> In src/CMakeLists.txt
> <https://github.com/ceph/ceph/pull/10024#discussion_r69104015>:
> 
>> @@ -60,7 +67,7 @@ else(no_yasm)
>>      COMMAND uname -m
>>      OUTPUT_VARIABLE arch
>>      OUTPUT_STRIP_TRAILING_WHITESPACE)
>> -  if(arch STREQUAL "x86_64")
>> +  if((arch STREQUAL "x86_64") OR (arch STREQUAL "amd64"))
> 
> might want to use
> 
> if(arch MATCHES "amd64|x86_64")

Looks much nicer...

src/CMakeLists.txt
wjwithagen: On 30-6-2016 11:55, Kefu Chai wrote:
> In src/CMakeLists.txt
> <https://github.com/ceph/ceph/pull/10024#discussion_r69104633>:
> 
>> @@ -351,6 +367,12 @@ endif(HAVE_GOOD_YASM_ELF64)
>>  add_library(common_texttable_obj OBJECT
>>    common/TextTable.cc)
>>  
>> +if(${LINUX})
> 
> i'd recommend use |list(APPEND libcommon_files)| different files for
> different OS.

Found |list(APPEND....)| half way thru.
Once the dust has settled I'll start cleaning this up.
Still running into things that are in/out for FreeBSD on different places.

CMakeLists.txt
tchaikov: yes, please.

src/CMakeLists.txt
wjwithagen: They were giving me a hard time. But now I end up putting them back in.
Just needed to add safe_io during linking.
So I guess this is going to go in the end.

It got excluded because it is for mountcephfs, which will require major plumbing to get working.
And I think the block with common files is needed for other things as well (if only for unittest_safe_io. So I suspect that the files there are not really mountcephfs only??

src/CMakeLists.txt
wjwithagen: On 30-6-2016 12:04, Kefu Chai wrote:
> In src/CMakeLists.txt
> <https://github.com/ceph/ceph/pull/10024#discussion_r69105878>:
> 
>> @@ -192,7 +202,13 @@ else(INTEL_SSE4_1)
>>  endif(INTEL_SSE4_1)
>>  
>>  
>> -set(EXTRALIBS uuid rt dl ${ATOMIC_OPS_LIBRARIES})
>> +set(EXTRALIBS rt ${ATOMIC_OPS_LIBRARIES})
>> +if(${LINUX})
>> +  set(LIB_DL dl)
> 
> we should switch to |CMAKE_DL_LIBS|. see
> https://cmake.org/cmake/help/v2.8.11/cmake.html#variable:CMAKE_DL_LIBS

Will check it out.

src/CMakeLists.txt
tchaikov: they are not, sort of misleading, i agree.

src/test/CMakeLists.txt
wjwithagen: On 30-6-2016 12:05, Kefu Chai wrote:
> In src/test/CMakeLists.txt
> <https://github.com/ceph/ceph/pull/10024#discussion_r69106054>:
> 
>> @@ -772,7 +772,7 @@ target_link_libraries(unittest_libcephfs_config cephfs)
>>  # unittest_rbd_replay
>>  add_executable(unittest_rbd_replay
>>    test_rbd_replay.cc
>> -  $<TARGET_OBJECTS:parse_secret_objs>
>> +#  $<TARGET_OBJECTS:parse_secret_objs>
> 

Sloppy exclude. Sorry.
parse_secrect_objs used KVS and that is not on FreeBSD.
Although it seems like a nice feature

wjwithagen:
On 30-6-2016 12:07, Kefu Chai wrote:
> would be better if you could split 0d2a9de
> <https://github.com/ceph/ceph/commit/0d2a9de0287e83015ea22f7c09ad4fdab7f0a0f9>
> into smaller commits.

Sure, no problem...

Suggestion on how to divide?

CMakeLists.txt
wjwithagen: So but then it could also be |if(FREEBSD)|??

wjwithagen:
@tchaikov 
Clang linker and C-compiler are complaining about:
`[ 58%] Linking CXX executable ../bin/ceph-osd c++: warning: argument unused during compilation: -Wp,-U_FORTIFY_SOURCE,-D_FORTIFY_SOURCE=2`

So this is only valid for C++.
Is there an easy way to make this a C++ compile only option?

src/CMakeLists.txt
Using APPEND adds `;` between the elements?
So that does not always work. 
It didn't work for the C_CFLAGS

wjwithagen:
@tchaikov 
Ping...
If you want even smaller commits, I have to redo them in smaller chuncks in a new WIP

CMakeLists.txt
tchaikov: yeah.

CMakeLists.txt
```diff
@@ -105,9 +127,9 @@ CHECK_INCLUDE_FILES("execinfo.h" HAVE_EXECINFO_H)
 if(${LINUX})
 CHECK_INCLUDE_FILES("sched.h" HAVE_SCHED)
 endif(${LINUX})
-if(${FREBSD})
+if(${FREEBSD})
```
tchaikov: please squash the fix into the commit which introduce `FREBSD`.

src/test/CMakeLists.txt
```diff
@@ -131,6 +131,7 @@ target_link_libraries(ceph_omapbench
   ${CMAKE_DL_LIBS}
   )
 
+if(${WITH_KVS})
```
tchaikov: nit, `if(WITH_KVS)`

CMakeLists.txt
```diff
@@ -19,6 +19,24 @@ endif (POLICY CMP0023)
...
+# find out which platform we are building on
+if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
+  message(STATUS "Building for Linux.")
+  set(LINUX ON)
+  set(UNIX ON)
+  FIND_PACKAGE(Threads)
+elseif(${CMAKE_SYSTEM_NAME} MATCHES "FreeBSD")
+  message(STATUS "Building for FreeBSD.")
+  set(FREEBSD ON)
+  set(UNIX ON)
```
tchaikov: @alimaredia do you know what `UNIX` is used for?

CMakeLists.txt
```diff
@@ -19,6 +19,24 @@ endif (POLICY CMP0023)
 
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake/modules/")
 
+# find out which platform we are building on
+if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
```
tchaikov: could be `if(CMAKE_SYSTEM_NAME MATCHES "Linux")`.

CMakeLists.txt
```diff
@@ -19,6 +19,24 @@ endif (POLICY CMP0023)
...
+  set(UNIX ON)
+  FIND_PACKAGE(Threads)
+elseif(${CMAKE_SYSTEM_NAME} MATCHES "FreeBSD")
+  message(STATUS "Building for FreeBSD.")
+  set(FREEBSD ON)
+  set(UNIX ON)
+  FIND_PACKAGE(Threads)
+elseif(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
+  message(STATUS "Building for Darwin.")
+  set(OperatingSystem "Mac OS X")
```
tchaikov: @alimaredia and `OperatingSystem`?

CMakeLists.txt
```diff
@@ -54,6 +72,12 @@ link_directories(
   ${OFED_PREFIX}/lib
   ${LEVELDB_PREFIX}/lib
 )
+if(${FREEBSD})
```
tchaikov: could be `if(FREEBSD)`

src/CMakeLists.txt
```diff
@@ -732,12 +768,19 @@ if(${WITH_RBD})
   set(DENCODER_EXTRALIBS
       ${DENCODER_EXTRALIBS}
       rbd_replay_types)
+  set(dencoder_srcs
+      ${dencoder_srcs}
+      $<TARGET_OBJECTS:krbd_objs> )
 endif(${WITH_RBD})
+if(${WITH_KVS})
+  set(dencoder_srcs
```
tchaikov: might want to use `list(APPEND` instead.

CMakeLists.txt
```diff
@@ -206,6 +228,7 @@ if(WITH_SPDK)
   set(HAVE_SPDK TRUE)
 endif(WITH_SPDK)
 
+option(WITH_BLKID "Block device library" ON)
```
tchaikov: why would you want to make it an option?

tchaikov:
> Clang linker and C-compiler are complaining about:
> `[ 58%] Linking CXX executable ../bin/ceph-osd c++: warning: argument unused during compilation: -Wp,-U_FORTIFY_SOURCE,-D_FORTIFY_SOURCE=2`
> 
> So this is only valid for C++.
> Is there an easy way to make this a C++ compile only option?

not sure why `CMAKE_C_FLAGS` is used when linking an executable? and i am sure this flag is valid for C sources also. could you double check your settings?

CMakeLists.txt
wjwithagen: On 4-7-2016 06:18, Kefu Chai wrote:
> In CMakeLists.txt
> <https://github.com/ceph/ceph/pull/10024#discussion_r69407120>:
> 
>> @@ -206,6 +228,7 @@ if(WITH_SPDK)
>>    set(HAVE_SPDK TRUE)
>>  endif(WITH_SPDK)
>>  
>> +option(WITH_BLKID "Block device library" ON)
> 
> why would you want to make it an option?

Because it uses blkid, and that is non-existent under FreeBSD.
So for the time being I compile it, and do not run the tests.

wjwithagen:
On 4-7-2016 06:28, Kefu Chai wrote:
>     Clang linker and C-compiler are complaining about:
>     |[ 58%] Linking CXX executable ../bin/ceph-osd c++: warning:
>     argument unused during compilation:
>     -Wp,-U_FORTIFY_SOURCE,-D_FORTIFY_SOURCE=2|
> 
>     So this is only valid for C++.
>     Is there an easy way to make this a C++ compile only option?
> 
> not sure why |CMAKE_C_FLAGS| is used when linking an executable? and i
> am sure this flag is valid for C sources also. could you double check
> your settings?

For flags that only need to be used in linking, I do something like:
  set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS}
-Wl,-export-dynamic")

But the -Wp suggests preprocessor, and the warning complains that it is
used during linking. So there is a valid point in there.
I try to find out what should be used by looking at:
    cmake --system-information
and see what is used where.

Now the challenging selection is for building the libraries, what does
one use where and when.

src/CMakeLists.txt
wjwithagen: On 4-7-2016 06:14, Kefu Chai wrote:
> In src/CMakeLists.txt
> <https://github.com/ceph/ceph/pull/10024#discussion_r69407011>:
> 
>>  endif(${WITH_RBD})
>> +if(${WITH_KVS})
>> +  set(dencoder_srcs
> 
> might want to use |list(APPEND| instead.

That are thing I still have to figure out.
Sometimes |list(APPEND| is usable, but only when cmake uses the list.
Otherwise it is a list wit ';' between the elements, and that does not
work for the other tools.

Same is a bit when to use if(${VAR}) and when if(VAR), but the first
case always seem to work. Which is why I'm using that one.
Got bitten once to many having to rerun
Guess I really do need to read up in that.

tchaikov: cmake expect `CMAKE_*_FLAGS` to be a single string instead of lists. guess they are exceptions.

CMakeLists.txt
wjwithagen: @tchaikov 
This is an extract from `cmake --help-command if`:
```
The if command was written very early in CMake's history, predating
the ``${}`` variable evaluation syntax, and for convenience evaluates
variables named by its arguments as shown in the above signatures.
Note that normal variable evaluation with ``${}`` applies before the if
command even receives the arguments.  Therefore code like::

 set(var1 OFF)
 set(var2 "var1")
 if(${var2})

appears to the if command as::

 if(var1)

and is evaluated according to the ``if(<variable>)`` case documented
above.  The result is ``OFF`` which is false.  However, if we remove the
``${}`` from the example then the command sees::

 if(var2)

which is true because ``var2`` is defined to "var1" which is not a false
constant.
```

And because of that I think I would prefer to keep writting ${var}, just in case there is any medling with constants and variables.

tchaikov: i did read this also, the reason why i'd recommend `if(foo)` is for the sake of simplicity and consistency. they are the equivalent under most circumstances as explained by the document you quote.

wjwithagen: On 4-7-2016 12:32, Kefu Chai wrote:
> In CMakeLists.txt
> <https://github.com/ceph/ceph/pull/10024#discussion_r69438886>:
> 
>> @@ -54,6 +72,12 @@ link_directories(
>>    ${OFED_PREFIX}/lib
>>    ${LEVELDB_PREFIX}/lib
>>  )
>> +if(${FREEBSD})
> 
> i did read this also, the reason why i'd recommend |if(foo)| is for the
> sake of simplicity and consistency. they are the equivalent under most
> circumstances as explained by the document you quote.

Oke, but then "you|me|others"'d better be pretty sure that the reference
is a "constant", because the errors are otherwise hard to track and
understand.

tchaikov: could you be specific what errors you are reference?

wjwithagen: On 4-7-2016 16:20, Kefu Chai wrote:
> In CMakeLists.txt
> <https://github.com/ceph/ceph/pull/10024#discussion_r69464221>:
> 
>> @@ -54,6 +72,12 @@ link_directories(
>>    ${OFED_PREFIX}/lib
>>    ${LEVELDB_PREFIX}/lib
>>  )
>> +if(${FREEBSD})
> 
> could you be specific what errors you are reference?

phew, that is hard.... don't really remember
And I'm not sure that it was with this variable.
But I was trying some conditional code that did not work, until I
wrapped it in |${}|

But I'm on my way to change all the WITH_* I created to |if(WITH_*)|,
and we'll see what comes of it.

wjwithagen:
@tchaikov 
For the time being this is the change set for FreeBSD.
The errors I'm getting were also in the automake tree, so cmake should not have any influence on this

wjwithagen:
@tchaikov 
Ping, possible error will be the already known test problem.

CMakeLists.txt
tchaikov: so this is more about the platform limitation, maybe we can just disable it, like what you already do.

if we follow this logic further, we need to have a knob for every difference on platforms. this does not make sense to me at all.

CMakeLists.txt
tchaikov: okay, if you insist. but better off understanding why before we blindly use `if(${FREEBSD})` all over the places.

CMakeLists.txt
tchaikov: @alimaredia ping?

src/CMakeLists.txt
```diff
@@ -17,14 +17,30 @@ set(prefix ${CMAKE_INSTALL_PREFIX})
...
 
 set(CMAKE_ASM_COMPILER  ${PROJECT_SOURCE_DIR}/src/yasm-wrapper)
 set(CMAKE_ASM_FLAGS "-f elf64")
-set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -rdynamic -Wall -Wtype-limits -Wignored-qualifiers -Winit-self -Wpointer-arith -Werror=format-security -fno-strict-aliasing -fsigned-char")
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wtype-limits -Wignored-qualifiers -Winit-self -Wpointer-arith -Werror=format-security -fno-strict-aliasing -fsigned-char")
+if(CMAKE_CXX_COMPILER_ID MATCHES "GNU")
+  add_definitions("-D_GNU_SOURCE -D_FILE_OFFSET_BITS=64")
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -rdynamic")
+elseif(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  set(CMAKE_CXX_FLAGS_DEBUG "-O0 -g")
```
tchaikov: you can use "-DCMAKE_BUILD_TYPE=Debug" for this.

src/CMakeLists.txt
```diff
@@ -17,14 +17,30 @@ set(prefix ${CMAKE_INSTALL_PREFIX})
...
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-mismatched-tags -Wno-unused-function")
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-unused-local-typedef -Wno-inconsistent-missing-override")
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-unused-private-field -Wno-varargs")
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-gnu-designator -Wno-mismatched-tags")
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-missing-braces -Wno-parentheses -Wno-deprecated-register")
+endif(CMAKE_CXX_COMPILER_ID MATCHES "GNU")
 
 if(NOT CMAKE_BUILD_TYPE)
-  set(CMAKE_BUILD_TYPE "RelWithDebInfo" CACHE STRING "Default BUILD_TYPE is RelWithDebInfo, other options are: Debug, Release, and MinSizeRel." FORCE) 
+  set(CMAKE_BUILD_TYPE "RelWithDebInfo"
```
tchaikov: please do not do reformatting unless you are touch this piece of code.

src/CMakeLists.txt
```diff
@@ -91,11 +109,14 @@ else(no_yasm)
...
     message(STATUS " we are not x86_64 && !x32")
-  endif(arch STREQUAL "x86_64")
+  endif(arch MATCHES "x86_64|amd64")
 endif(no_yasm)
 
 
-set(CMAKE_CXX_FLAGS "${CMAKE_C_FLAGS} -ftemplate-depth-1024 -Wno-invalid-offsetof -Wnon-virtual-dtor -Wno-invalid-offsetof -Wstrict-null-sentinel -Woverloaded-virtual")
+set(CMAKE_CXX_FLAGS "${CMAKE_C_FLAGS} -ftemplate-depth-1024 -Wno-invalid-offsetof -Wnon-virtual-dtor -Wno-invalid-offsetof -Woverloaded-virtual")
+if(CMAKE_CXX_COMPILER_ID MATCHES "GNU")
+set(CMAKE_CXX_FLAGS "${CMAKE_C_FLAGS} -Wstrict-null-sentinel")
```
tchaikov: might want to indent this line.

src/CMakeLists.txt
```diff
@@ -197,7 +222,12 @@ else(INTEL_SSE4_1)
 endif(INTEL_SSE4_1)
 
 
-set(EXTRALIBS uuid rt dl ${ATOMIC_OPS_LIBRARIES})
+set(EXTRALIBS rt ${ATOMIC_OPS_LIBRARIES})
+if(${LINUX})
```
tchaikov: use `if(LINUX)` if you please.

src/CMakeLists.txt
```diff
@@ -341,6 +371,16 @@ endif(HAVE_GOOD_YASM_ELF64)
 add_library(common_texttable_obj OBJECT
   common/TextTable.cc)
 
+if(LINUX)
```
tchaikov: nit, could you move this block down after `set(libcommon_files ..)` and conditional `list(APPEND libcommon_files )` there? so we don't need to introduce `msg_async_event_os_srcs`?

src/CMakeLists.txt
```diff
@@ -489,8 +531,11 @@ if(ENABLE_SHARED)
 endif(ENABLE_SHARED)
 
 add_library(common STATIC ${libcommon_files}
-  $<TARGET_OBJECTS:mon_common_objs>
+  $<TARGET_OBJECTS:mon_common_objs>)
+if(WITH_LIBCEPHFS)
+  list(APPEND common STATIC ${libcommon_files}
```
tchaikov: might need to indent the following line.

src/CMakeLists.txt
```diff
@@ -615,10 +660,9 @@ add_executable(ceph-objectstore-tool
   tools/ceph_objectstore_tool.cc
   tools/RadosDump.cc
   $<TARGET_OBJECTS:common_util_obj>)
-target_link_libraries(ceph-objectstore-tool osd os global ${Boost_PROGRAM_OPTIONS_LIBRARY} fuse dl)
+target_link_libraries(ceph-objectstore-tool osd os global ${Boost_PROGRAM_OPTIONS_LIBRARY} ${CMAKE_DL_LIBS})
```
tchaikov: again, `fuse` is necessary for supporting FuseStore.

src/CMakeLists.txt
tchaikov: @wjwithagen still need to be addressed.

src/CMakeLists.txt
```diff
@@ -865,7 +918,7 @@ if(NOT ALLOCATOR STREQUAL "jemalloc")
 endif()
 add_custom_target(build_rocksdb
     COMMAND
-    PORTABLE=1 ${disable_jemalloc} $(MAKE) static_lib EXTRA_CXXFLAGS='-fPIC -Wno-unused-variable'
+    PORTABLE=1 ${disable_jemalloc} $(MAKE) static_lib V=1 VERBOSE=1 CXX=$(CXX) EXTRA_CXXFLAGS='-Irocksdb/include  -I/usr/local/include -fPIC -Wno-unused-variable'
```
tchaikov: why you'd need `-Irocksdb/include` and `-I/usr/local/include`?

and could you explain why we need `VERBOSE=1` and `V=1`?

src/CMakeLists.txt
```diff
@@ -1161,8 +1216,8 @@ if(WITH_LIBCEPHFS)
   set(mount_ceph_srcs
     mount/mount.ceph.c)
   add_executable(mount.ceph ${mount_ceph_srcs}
-    $<TARGET_OBJECTS:parse_secret_objs>
-    $<TARGET_OBJECTS:common_mountcephfs_objs>)
+    $<TARGET_OBJECTS:common_mountcephfs_objs>
```
tchaikov: please avoid unnecessary reformatting.

src/CMakeLists.txt
```diff
@@ -510,11 +510,15 @@ set(mon_common_files
   erasure-code/ErasureCodePlugin.cc)
 add_library(mon_common_objs OBJECT ${mon_common_files})
 if(${LINUX})
-set(common_mountcephfs_files
-  common/module.c
+  list(APPEND libcommon_files
```
tchaikov: please squash the indent fix into the commit which introduces this `if` block.

tchaikov:
> Suggestion on how to divide?

divide by logic and coherent changes. for example

- conditional set `CMAKE_C_FLAGS` by `CMAKE_CXX_COMPILER_ID`
- do not `try_compile` ARM code on i386 machines
- do not set "resolve" and "uuid" on non Linux platforms (are you sure this applies to OSX and Solaris also)
- include EventKqueue.cc in libcommon on FreeBSD.

src/CMakeLists.txt
```diff
@@ -796,6 +800,10 @@ target_link_libraries(ceph-dencoder
   ${EXTRALIBS}
   ${CMAKE_DL_LIBS}
   )
+if(${WITH_KVS})
```
tchaikov: please squash this change into a dedicated commit which includes the https://github.com/ceph/ceph/pull/10024/commits/60185795b9c13fba8ae7947ba95ce7c18eed79c1#diff-95e351a3805a1dafa85bf20b81d086e6L747 changes. that change only handles the `WITH_KVS` related changes on FreeBSD.

tchaikov:
@wjwithagen could you restructure your commits, some of the changes are interleaved. it would be easier to read them if every commit is self-contained and coherent.

src/CMakeLists.txt
Not quite.... I'd like to also have -O0 during compilation, because that does not remove any inlined code or what so ever. Running things thru the debugger is way much easier if all code is actually visible.

src/CMakeLists.txt
I will rewrite all ${OS_TYPE} references to OS_TYPE.

src/CMakeLists.txt
'mmm I did fix that when you commented at first.
No sure why it is not here.

src/CMakeLists.txt
tchaikov: iirc, `CMAKE_BUILD_TYPE=Debug` implies `-O0` and `-g`.

src/CMakeLists.txt
That might need to go into an if(FREEBSD) box. 
And I'll take care of the V=1 VERBOSE=1 flags in there as well.
I prefer my compilation rather wordy, so I can see what the build system is doing.

src/CMakeLists.txt
Not sure how to divided squashed commits back into loose parts.
For me it is easier to start fresh I think.

src/CMakeLists.txt
tchaikov: `-Irocksdb/include` still looks strange.

src/CMakeLists.txt
tchaikov: @wjwithagen how about "git reset" your change, and "git add -p"?

Going to give it a try. Worst case I have to redo the patching.

wjwithagen:
@tchaikov 
Kefu, I was going to rebase and resubmit.

But todays rebase is so completely different that I'm going to start all over from fresh.
Everything has moved into more subdirectories again,  a good plan as far as I'm concerned.

So I'm going to close this PULL and going to start to submit several pulls to not fall again in the trap of a large pull that is changing too rapid for me and you to keep up.

tchaikov:
@wjwthagen, thanks for your insistence and hard work. could you wait a while after #10351 gets merged? which is a relatively large refactor.

wjwithagen:
first things are in #10352
Not going to add much until this is committed. This is the FreeBSD global stuff and could go in with much distress.

PR #10031, opened 2016-06-29
rgw: finish error_repo cr in stop_spawned_services()

yehudasa:
Fixes: http://tracker.ceph.com/issues/16530

Need to call finish, otherwise drain_all() wouldn't be able to
drain it.


cbodley:
lgtm :+1:

PR #10032, opened 2016-06-30
rgw: add deadlock detection to RGWCoroutinesManager::run()

cbodley:
assert instead of returning if all coroutines are blocked

### DIFF
--- a/src/rgw/rgw_coroutine.cc
+++ b/src/rgw/rgw_coroutine.cc
@@ -570,6 +570,7 @@ int RGWCoroutinesManager::run(list<RGWCoroutinesStack *>& stacks)
   }
 
   lock.get_write();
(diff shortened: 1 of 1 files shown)


yehudasa:
@cbodley did you manage to validate that it actually works?

cbodley:
@yehudasa i did verify that it hit the assert before we fixed the error_repo deadlock. i'll run it through teuthology to make sure we don't hit it accidentally

cbodley:
ran through teuthology without triggering the assertion. only unexpected failure in the results was `kernel: [ INFO: possible recursive locking detected ] ' in syslog`, but not related to this change

http://pulpito.ceph.com/cbodley-2016-08-19_10:23:45-rgw-wip-rgw-cr-deadlock---basic-mira/

PR #10034, opened 2016-06-30
librbd: add consistency groups operations with images

VictorDenisov:


src/cls/rbd/cls_rbd.cc
```diff
@@ -138,9 +140,17 @@ cls_method_handle_t h_group_create;
 cls_method_handle_t h_group_dir_list;
 cls_method_handle_t h_group_dir_add;
 cls_method_handle_t h_group_dir_remove;
+cls_method_handle_t h_group_image_add;
+cls_method_handle_t h_group_image_remove;
+cls_method_handle_t h_group_image_list;
+cls_method_handle_t h_group_image_clean;
+cls_method_handle_t h_group_image_add_uplink;
```
dillaman: I would rename these three functions to ```image_add_group```, ```image_remove_group```, and ```image_get_group``` where the ```image_``` prefix denotes that it is an operation against the image header.

src/cls/rbd/cls_rbd.cc
```diff
@@ -4421,6 +4431,337 @@ int group_dir_remove(cls_method_context_t hctx, bufferlist *in, bufferlist *out)
...
+  int64_t pool_id;
+  try {
+    bufferlist::iterator iter = in->begin();
+    ::decode(image_id, iter);
+    ::decode(pool_id, iter);
+  } catch (const buffer::error &err) {
+    return -EINVAL;
+  }
+
+  string image_key = RBD_IMAGE_KEY_PREFIX + image_id + "_" + stringify(pool_id);
```
dillaman: Minor: I would create a helper function to computing they key and I would store the pool id as zero-padded hex before the image id

src/cls/rbd/cls_rbd.cc
```diff
@@ -4421,6 +4431,337 @@ int group_dir_remove(cls_method_context_t hctx, bufferlist *in, bufferlist *out)
...
+}
+
+/**
+ * List images pertaining to the consistency group.
+ *
+ * Input:
+ * none
+ *
+ * Output:
+ * @param length of the list of images
```
dillaman: Use standard containers instead of a self-rolled solution: std::list<std::tuple<x, y, z>> or std::list<GroupImageSpec>

src/cls/rbd/cls_rbd.cc
```diff
@@ -4421,6 +4431,337 @@ int group_dir_remove(cls_method_context_t hctx, bufferlist *in, bufferlist *out)
...
+ */
+int group_image_add(cls_method_context_t hctx, bufferlist *in, bufferlist *out)
+{
+  CLS_LOG(20, "group_image_add");
+
+  std::string image_id;
+  int64_t pool_id;
+  try {
+    bufferlist::iterator iter = in->begin();
+    ::decode(image_id, iter);
```
dillaman: I would just create a struct that wraps this concept since it is re-used over and over again. For parent images we have a struct that is pool id, image id, and snap id, so this would just be a GroupImageSpec with pool id and image id.

src/include/rbd_types.h
```diff
@@ -106,4 +107,9 @@ struct rbd_obj_header_ondisk {
 	struct rbd_obj_snap_ondisk snaps[0];
 } __attribute__((packed));
 
+enum group_image_link_state {
```
dillaman: Minor: seems like this can be moved to cls_rbd_types.h

dillaman: Minor: GroupImageLinkState

src/include/rbd_types.h
```diff
@@ -106,4 +107,9 @@ struct rbd_obj_header_ondisk {
...
+	LINK_NORMAL,
```
dillaman: Minor: enum values should include enum name: a la GROUP_IMAGE_LINK_STATE_XYZ. "normal" and "dirty" are not too descriptive -- seems like DIRTY implies ATTACHING and NORMAL implies ATTACHED.

src/cls/rbd/cls_rbd.cc
```diff
@@ -4421,6 +4431,337 @@ int group_dir_remove(cls_method_context_t hctx, bufferlist *in, bufferlist *out)
...
+  string last_read = RBD_SNAP_KEY_PREFIX;
+  int r;
+  do {
+    r = cls_cxx_map_get_vals(hctx, RBD_IMAGE_KEY_PREFIX, RBD_IMAGE_KEY_PREFIX,
+			     max_read, &vals);
+    if (r < 0)
+      return r;
+
+    int64_t length = vals.size();
+    ::encode(length, *out);
```
dillaman: In a ```while``` loop that might repeat -- cannot encode the output length here.  Use a standard container and encode everything to the output after exiting the loop.

src/cls/rbd/cls_rbd.cc
```diff
@@ -4421,6 +4431,337 @@ int group_dir_remove(cls_method_context_t hctx, bufferlist *in, bufferlist *out)
...
+  }
+
+  return 0;
+}
+
+/**
+ * List images pertaining to the consistency group.
+ *
+ * Input:
+ * none
```
dillaman: Minor: usually like to see max_return and start_after params.

src/cls/rbd/cls_rbd.cc
```diff
@@ -4421,6 +4431,337 @@ int group_dir_remove(cls_method_context_t hctx, bufferlist *in, bufferlist *out)
...
+	 it != vals.end(); ++it) {
+      bufferlist::iterator iter = it->second.begin();
+      int64_t state;
+      try {
+	::decode(state, iter);
+      } catch (const buffer::error &err) {
+	CLS_ERR("error decoding state for image: %s", it->first.c_str());
+	return -EIO;
+      }
+      typedef boost::tokenizer<boost::char_separator<char>> tokenizer;
```
dillaman: Minor: probably simpler to just do:

```
size_t pos = it->first.find("_");
if (pos == std::string::npos) {
  // ERROR
}
std::string image_id(it->first.substr(0, pos));
std::string pool_id(it->first.substr(pos + 1));
```

src/include/rbd_types.h
```diff
@@ -77,6 +77,7 @@
 #define RBD_GROUP_HEADER_PREFIX "rbd_group_header."
 
 #define RBD_GROUP_DIRECTORY "rbd_group_directory"
+#define RBD_GROUP_UPLINK "rbd_group_uplink"
```
dillaman: Minor: move to cls_rbd.cc (the key name is a hidden implementation detail).

src/cls/rbd/cls_rbd.cc
```diff
@@ -4421,6 +4431,337 @@ int group_dir_remove(cls_method_context_t hctx, bufferlist *in, bufferlist *out)
...
+
+    if ((old_group_id != group_id) || (old_pool_id != pool_id)) {
+      return -EEXIST;
+    } else {
+      return 0; // In this case the values are already correct
+    }
+  }
+
+  bufferlist refbl;
+  ::encode(pool_id, refbl);
```
dillaman: Can reuse that same ```GroupImageSpec``` concept here

src/cls/rbd/cls_rbd.cc
```diff
@@ -4421,6 +4431,337 @@ int group_dir_remove(cls_method_context_t hctx, bufferlist *in, bufferlist *out)
...
+    ::decode(group_id, iter);
+    ::decode(pool_id, iter);
+  } catch (const buffer::error &err) {
+    return -EINVAL;
+  }
+
+  bufferlist existing_refbl;
+
+  int r = read_key(hctx, RBD_GROUP_UPLINK, &existing_refbl);
+  if (r != -ENOENT) { // No entry means this image is not a member of any consistency group. So, we can use it.
```
dillaman:
```
if (r == -ENOENT) {
   // DO SEARCH HERE
} else if (r < 0) {
  // ERROR
}
```

src/include/rbd_types.h
dillaman: ... and do you need a DETACHING state in order to protect against a crash while removing?

src/cls/rbd/cls_rbd.cc
```diff
@@ -4421,6 +4431,337 @@ int group_dir_remove(cls_method_context_t hctx, bufferlist *in, bufferlist *out)
...
+ * Add an image to the consistency group.
+ *
+ * Input:
+ * @param image_id (std::string)
+ * @param pool_id (int64_t)
+ *
+ * Output:
+ * @return 0 on success, negative error code on failure
+ */
+int group_image_add(cls_method_context_t hctx, bufferlist *in, bufferlist *out)
```
dillaman: Minor: could change this to a set state method and remove the need for the extra _clean method.  Also needed if DETACHING state required.

src/cls/rbd/cls_rbd_client.h
```diff
@@ -35,7 +35,8 @@ namespace librbd {
                                     std::map<rados::cls::lock::locker_id_t,
                                              rados::cls::lock::locker_info_t> *lockers,
                                     bool *exclusive_lock, std::string *lock_tag,
-                                    ::SnapContext *snapc, parent_info *parent);
+                                    ::SnapContext *snapc, parent_info *parent,
+				    std::pair<int64_t, std::string> *uplink);
```
dillaman: ```std::pair<int64_t, std::string> *uplink``` -> ```cls::rbd::GroupSpec *group_spec```

src/include/rbd/librbd.h
```diff
@@ -663,6 +663,22 @@ CEPH_RBD_API int rbd_group_create(rados_ioctx_t p, const char *name);
...
+				rados_ioctx_t image_p, const char *image_name);
+CEPH_RBD_API int rbd_group_image_remove(
+				rados_ioctx_t group_p, const char *group_name,
+				rados_ioctx_t image_p, const char *image_name);
+CEPH_RBD_API int rbd_group_image_list(
+				  rados_ioctx_t group_p, const char *group_name,
+				  char *names, size_t *names_size,
+				  int64_t *pools, size_t *pools_size,
+				  int *states, size_t *states_size);
+CEPH_RBD_API int rbd_group_image_get_uplink(
```
dillaman: ```rbd_group_image_get_uplink``` -> ```rbd_get_group```

src/include/rbd/librbd.hpp
```diff
@@ -197,6 +204,7 @@ class CEPH_RBD_API Image
 		      std::string *parent_snapname);
   int old_format(uint8_t *old);
   int size(uint64_t *size);
+  int uplink(std::pair<int64_t, std::string> *uplink);
```
dillaman: ```uplink``` -> ```get_group```

src/librbd/ImageCtx.h
```diff
@@ -119,6 +119,7 @@ namespace librbd {
     std::string id; // only used for new-format images
     parent_info parent_md;
     ImageCtx *parent;
+    std::pair<int64_t, std::string> group_uplink;
```
dillaman: Minor: GroupSpec

src/librbd/internal.cc
```diff
@@ -2107,6 +2107,15 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force,
         return -EBUSY;
       }
 
+      std::map<std::string, bufferlist> out_vals;
+      r = io_ctx.omap_get_vals(header_oid, "", RBD_GROUP_UPLINK, 1, &out_vals);
```
dillaman: Use ```cls_client``` helper method

src/librbd/internal.cc
```diff
@@ -2107,6 +2107,15 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force,
...
+      if (out_vals.size() > 0) {
+	lderr(cct) << "image is in a consistency group - not removing" << dendl;
+	ictx->owner_lock.put_read();
+	ictx->state->close();
+	return -EMLINK;
```
dillaman: Minor: -EBUSY is probably more appropriate

src/librbd/internal.cc
```diff
@@ -3826,4 +3835,212 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force,
...
+                 << cpp_strerror(r)
+        	 << dendl;
+      return r;
+    }
+    string group_header_oid = util::group_header_name(group_id);
+
+    ldout(cct, 20) << "adding image to group name " << group_name
+                   << " group id " << group_header_oid << dendl;
+
+    ImageCtx *imctx = new ImageCtx(image_name, "", "", image_ioctx, true);
```
dillaman: No need to open the image to get its id -- use the dir_get_id cls_client method

src/librbd/internal.cc
```diff
@@ -3826,4 +3835,212 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force,
...
+      lderr(cct) << "error opening image: "
+        	 << cpp_strerror(-r) << dendl;
+      delete imctx;
+      return r;
+    }
+
+    ldout(cct, 20) << "removing image " << image_name
+                   << " image id " << imctx->header_oid << dendl;
+
+    r = cls_client::group_image_add(&group_ioctx, group_header_oid,
```
dillaman: DETACHING state is probably more appropriate

src/include/rbd/librbd.h
```diff
@@ -663,6 +663,22 @@ CEPH_RBD_API int rbd_group_create(rados_ioctx_t p, const char *name);
...
+				rados_ioctx_t group_p, const char *group_name,
+				rados_ioctx_t image_p, const char *image_name);
+CEPH_RBD_API int rbd_group_image_remove(
+				rados_ioctx_t group_p, const char *group_name,
+				rados_ioctx_t image_p, const char *image_name);
+CEPH_RBD_API int rbd_group_image_list(
+				  rados_ioctx_t group_p, const char *group_name,
+				  char *names, size_t *names_size,
+				  int64_t *pools, size_t *pools_size,
+				  int *states, size_t *states_size);
```
dillaman: Minor: use an enum not an int for the state

src/librbd/internal.cc
```diff
@@ -3826,4 +3835,212 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force,
...
+      rados.ioctx_create2(std::get<1>(i), ioctx);
+      std::string image_name;
+      r = cls_client::dir_get_name(&ioctx, RBD_DIRECTORY,
+                                   std::get<0>(i), &image_name);
+      if (r < 0) {
+        return r;
+      }
+      images.push_back(std::make_tuple(image_name,
+                                       std::get<1>(i),
+                                       std::get<2>(i)));
```
dillaman: Safely convert from private API state enum to public state (see how mirror state handles it)

src/test/cli/rbd/help.t
```diff
@@ -21,6 +21,9 @@
       flatten                     Fill clone with parent data (make it
                                   independent).
       group create                Create a consistency group.
+      group image add             Add an image to a consistency group.
+      group image list            Dump list of images in a consistency group.
```
dillaman: Minor: "List images in a consistency group"

src/test/cli/rbd/help.t
```diff
@@ -420,6 +423,55 @@
...
+  
+  Positional arguments
+    <group-spec>         group specification
+                         (example: [<pool-name>/]<group-name>)
+    <image-spec>         image specification
+                         (example: [<pool-name>/]<image-name>)
+  
+  Optional arguments
+    -p [ --pool ] arg    pool name
+    --group arg          group name
```
dillaman: Minor: "--pool" might get confused between group pool and image pool. Also missing "--image" option

src/tools/rbd/action/Info.cc
```diff
@@ -136,7 +148,9 @@ static int do_show_info(const char *imgname, librbd::Image& image,
               << "\tblock_name_prefix: " << prefix
               << std::endl
               << "\tformat: " << (old_format ? "1" : "2")
-              << std::endl;
+	      << std::endl
+	      << "\tconsistency group: " << group_string
```
dillaman: Minor: I'd leave it out of the output if there is no consistency group (like we leave out the journal or mirroring fields).

src/tools/rbd/action/Remove.cc
```diff
@@ -68,6 +68,23 @@ int execute(const po::variables_map &vm) {
...
                 << std::endl;
+    } else if (r == -EMLINK) {
+      librbd::Image image;
+      int image_r = utils::open_image(io_ctx, image_name, true, &image);
+      std::pair<int64_t, std::string> uplink;
+      if (image_r == 0) {
+	image_r = image.uplink(&uplink);
+      }
+      if (image_r == 0)
+	std::cerr << "rbd: error: image pertains to a consistency group "
```
dillaman: Minor: s/pertains/belongs to/g

src/librbd/internal.cc
dillaman: ... never mind, I see that you are using it for nicer error output from the rbd CLI.

src/include/rbd_types.h
VictorDenisov: DETACHING is the same as ATTACHING. It only shows that the link is incomplete. If the link is incomplete you still can repeat addition as well as removal.

dillaman: Yeah, I thought about it again earlier today and agree it should just be attached and "incomplete". I would still like to see the _add and _clean methods just be combined into a _set method where the enum is provided.

src/librbd/internal.cc
VictorDenisov: I need header_oid below for referencing the group. Can I get it without opening the image?

dillaman: There is a helper function in Utils.cc to convert the image id to a header oid.

src/include/rbd_types.h
VictorDenisov: I would like to use this constant in test_cls_rbd.cc. The test won't see it in cls_rbd.cc. Is there any other appropriate header file for this constant?

dillaman: You could put it in cls_rbd_types.h

src/test/cli/rbd/help.t
VictorDenisov: I assume then we need to have something like: --image-pool, and --group-pool options.
Do I understand correctly that this is what you would like to see?

dillaman: The "--pool" is standard across all commands, so I would like to see "--group-pool"

VictorDenisov: A question about another command, but, I guess, relevant.
List images in a consistency groups command only needs group and pool names: would you prefer option 1 or option 2.

Option 1:
group image list
--pool pool name
--group group name

Option 2:
group image list
--group-pool pool name
--group group name

dillaman: I'd hate to be inconsistent -- how would you feel about saying "--pool" implies both image and group pool, or you can override via "--group-pool" if the group is in a different pool?

VictorDenisov: I think it makes sense. The CLI becomes a little bit unstructured, but at least it should be convenient to use. I'll use --group-pool and --image-pool, if any of them is missing or both of them then I'll take their values from --pool, if even that one is missing then I'll use the default pool name.

dillaman:
@VictorDenisov rebase is required

VictorDenisov:
@dillaman rebase is done.

src/cls/rbd/cls_rbd_types.h
```diff
@@ -134,6 +154,69 @@ std::ostream& operator<<(std::ostream& os, const MirrorImageStatusState& state);
...
+    : image_id(image_id), pool_id(pool_id) {}
+
+  static GroupImageSpec empty() {
+    return GroupImageSpec("", -1);
+  }
+
+  static int from_key(const std::string &image_key, GroupImageSpec *spec);
+
+  std::string image_id;
+  int64_t pool_id;
```
dillaman: ... = -1;

src/cls/rbd/cls_rbd_types.h
```diff
@@ -134,6 +154,69 @@ std::ostream& operator<<(std::ostream& os, const MirrorImageStatusState& state);
 
 WRITE_CLASS_ENCODER(MirrorImageStatus);
 
+struct GroupImageSpec {
+  GroupImageSpec() {}
+
+  GroupImageSpec(const std::string &image_id, int64_t pool_id)
+    : image_id(image_id), pool_id(pool_id) {}
+
+  static GroupImageSpec empty() {
```
dillaman: Minor: why not just use the default constructor?

src/cls/rbd/cls_rbd_types.h
```diff
@@ -134,6 +154,69 @@ std::ostream& operator<<(std::ostream& os, const MirrorImageStatusState& state);
...
+		   int64_t pool_id,
+		   GroupImageLinkState state)
+    : spec(image_id, pool_id), state(state) {}
+
+  GroupImageStatus(GroupImageSpec spec,
+		   GroupImageLinkState state)
+    : spec(spec), state(state) {}
+
+  GroupImageSpec spec;
+  GroupImageLinkState state;
```
dillaman: Minor: default initialize to INCOMPLETE

src/cls/rbd/cls_rbd_types.h
```diff
@@ -134,6 +154,69 @@ std::ostream& operator<<(std::ostream& os, const MirrorImageStatusState& state);
...
+
+WRITE_CLASS_ENCODER(GroupImageStatus);
+
+struct GroupSpec {
+  GroupSpec() {}
+  GroupSpec(const std::string &group_id, int64_t pool_id)
+    : group_id(group_id), pool_id(pool_id) {}
+
+  std::string group_id;
+  int64_t pool_id;
```
dillaman: ... = -1;

src/cls/rbd/cls_rbd.cc
```diff
@@ -4412,6 +4418,280 @@ int group_dir_remove(cls_method_context_t hctx, bufferlist *in, bufferlist *out)
...
+      bufferlist::iterator iter = it->second.begin();
+      cls::rbd::GroupImageLinkState state;
+      try {
+	::decode(state, iter);
+      } catch (const buffer::error &err) {
+	CLS_ERR("error decoding state for image: %s", it->first.c_str());
+	return -EIO;
+      }
+      cls::rbd::GroupImageSpec spec;
+      r = cls::rbd::GroupImageSpec::from_key(it->first, &spec);
```
dillaman: This is overwriting the result from ```cls_cxx_map_get_vals``` for the while loop conditional.

src/include/rbd_types.h
```diff
@@ -77,6 +77,7 @@
 #define RBD_GROUP_HEADER_PREFIX "rbd_group_header."
 
 #define RBD_GROUP_DIRECTORY "rbd_group_directory"
+#define RBD_GROUP_REF "rbd_group_ref"
```
dillaman: Minor: keep this implementation detail hidden in cls/rbd. Also, can we rename it to RBD_IMAGE_GROUP_SPEC or something similar to denote that it's tied to the image and it links to a group spec?

src/cls/rbd/cls_rbd.cc
```diff
@@ -4412,6 +4418,280 @@ int group_dir_remove(cls_method_context_t hctx, bufferlist *in, bufferlist *out)
...
+ * @param GroupSpec
+ * @return 0 on success, negative error code on failure
+ */
+int image_get_group(cls_method_context_t hctx,
+		    bufferlist *in, bufferlist *out)
+{
+  CLS_LOG(20, "image_get_group");
+  bufferlist refbl;
+  int r = cls_cxx_map_get_val(hctx, RBD_GROUP_REF, &refbl);
+  if (r < 0 && r != -ENOENT) {
```
dillaman: Why not return -ENOENT if there is no group link?

src/test/cls_rbd/test_cls_rbd.cc
```diff
@@ -953,18 +954,19 @@ TEST_F(TestClsRbd, get_mutable_metadata_features)
   std::string lock_tag;
   ::SnapContext snapc;
   parent_info parent;
+  cls::rbd::GroupSpec uplink;
```
dillaman: Minor: uplink -> group_spec

src/cls/rbd/cls_rbd.cc
dillaman: Never mind -- I see how it's being used.

src/librbd/ImageCtx.cc
```diff
@@ -164,6 +164,7 @@ struct C_InvalidateCache : public Context {
       order(0), size(0), features(0),
       format_string(NULL),
       id(image_id), parent(NULL),
+      group_ref("", -1),
```
dillaman: Minor: allow the default constructor to set it to "invalid"

src/librbd/ImageCtx.h
```diff
@@ -119,6 +119,7 @@ namespace librbd {
     std::string id; // only used for new-format images
     parent_info parent_md;
     ImageCtx *parent;
+    cls::rbd::GroupSpec group_ref;
```
dillaman: Minor: group_spec

src/librbd/internal.cc
```diff
@@ -2217,6 +2218,15 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force,
         return -EBUSY;
       }
 
+      cls::rbd::GroupSpec s;
+      r = cls_client::image_get_group(&io_ctx, header_oid, s);
+      if ((!s.group_id.empty()) && (s.pool_id != -1)) {
```
dillaman: Minor: could add a "is_valid" method to GroupSpec to hide the internal details of a valid vs invalid spec.

src/librbd/internal.cc
```diff
@@ -2217,6 +2218,15 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force,
...
+	lderr(cct) << "image is in a consistency group - not removing" << dendl;
+	ictx->owner_lock.put_read();
+	ictx->state->close();
+	return -EMLINK;
```
dillaman: -EBUSY?

src/librbd/internal.cc
dillaman: Still a valid comment

src/test/librbd/mock/MockImageCtx.h
```diff
@@ -70,6 +69,7 @@ struct MockImageCtx {
       id(image_ctx.id),
       name(image_ctx.name),
       parent_md(image_ctx.parent_md),
+      group_ref("", -1),
```
dillaman: Minor: initialize from value in ```image_ctx```

src/test/cli/rbd/help.t
```diff
@@ -420,6 +423,67 @@
...
+                         (example: [<pool-name>/]<group-name>)
+    <image-spec>         image specification
+                         (example: [<pool-name>/]<image-name>)
+  
+  Optional arguments
+    --group-pool arg     group pool name
+    --group arg          group name
+    --image-pool arg     image pool name
+    --image arg          image name
+    -p [ --pool ] arg    pool name which is used when either image or group or
```
dillaman: "pool name unless overridden"

src/librbd/internal.cc
VictorDenisov: There is a discussion earlier in this pull request. It's used for nicer output. If it's busy then we report that somebody is using it, if it's EMLINK we report that it belongs to a consistency group and specify what consistency group it belongs to.

dillaman:
@VictorDenisov still have that one open comment to avoid opening the image just to retrieve the header oid in ```group_image_remove```.

VictorDenisov:
It looks like I did this for group_image_add, but forgot about group_image_remove. Working on a fix.

VictorDenisov:
@dillaman done

PR #10037, opened 2016-06-30
ReplicatedPG:handled is unnecessary

yonghengdexin735:


xiexingguo:
We shall avoid the side-effect of `assert` in case the high level optimization of compiler translates `assert` to nothing.

PR #10038, opened 2016-06-30
mon: Monitor: validate prefix on handle_command()

tchaikov:
http://tracker.ceph.com/issues/16550


smithfarm:

OK to merge?

athanatos:
lgtm if @tchaikov is ok with the testing.

tchaikov:
lgtm

PR #10039, opened 2016-06-30
osd:OSD.cc:use "bool" instead of "int"

yonghengdexin735:


xiexingguo:
lgtm

athanatos:
This should not be one PR, please split it up.

PR #10049, opened 2016-06-30
librados: improvements async IO in librados and libradosstriper

sponce:
Notably : 
- added IO APis in librados for stat and xattr calls
- added remove/stat async API to striper
- optimized removal of files in striper

### DIFF
--- a/src/include/rados/librados.hpp
+++ b/src/include/rados/librados.hpp
@@ -950,6 +950,7 @@ namespace librados
      * other than SNAP_HEAD
      */
     int aio_remove(const std::string& oid, AioCompletion *c);
+    int aio_remove(const std::string& oid, AioCompletion *c, int flags);
 
     /**
      * Wait for all currently pending aio writes to be safe.
--- a/src/include/rados/librados.hpp
+++ b/src/include/rados/librados.hpp
@@ -983,6 +983,12 @@ namespace librados
     int aio_exec(const std::string& oid, AioCompletion *c, const char *cls, const char *method,
 	         bufferlist& inbl, bufferlist *outbl);
 
+    /*
+     * asynchronous version of unlock
+     */
+    int aio_unlock(const std::string &oid, const std::string &name,
+	           const std::string &cookie, AioCompletion *c);
+
     // compound object operations
     int operate(const std::string& oid, ObjectWriteOperation *op);
     int operate(const std::string& oid, ObjectReadOperation *op, bufferlist *pbl);
--- a/src/include/rados/librados.hpp
+++ b/src/include/rados/librados.hpp
@@ -968,7 +968,10 @@ namespace librados
      * @returns 0 on success, negative error code on failure
      */
     int aio_flush_async(AioCompletion *c);
-
+    int aio_getxattr(const std::string& oid, AioCompletion *c, const char *name, bufferlist& bl);
+    int aio_getxattrs(const std::string& oid, AioCompletion *c, std::map<std::string, bufferlist>& attrset);
+    int aio_setxattr(const std::string& oid, AioCompletion *c, const char *name, bufferlist& bl);
+    int aio_rmxattr(const std::string& oid, AioCompletion *c, const char *name);
     int aio_stat(const std::string& oid, AioCompletion *c, uint64_t *psize, time_t *pmtime);
     int aio_stat2(const std::string& oid, AioCompletion *c, uint64_t *psize, struct timespec *pts);
 
--- a/src/librados/librados.cc
+++ b/src/librados/librados.cc
@@ -1840,6 +1840,11 @@ int librados::IoCtx::aio_remove(const std::string& oid, librados::AioCompletion 
   return io_ctx_impl->aio_remove(oid, c->pc);
 }
 
+int librados::IoCtx::aio_remove(const std::string& oid, librados::AioCompletion *c, int flags)
+{
+  return io_ctx_impl->aio_remove(oid, c->pc, flags);
+}
+
 int librados::IoCtx::aio_flush_async(librados::AioCompletion *c)
 {
   io_ctx_impl->flush_aio_writes_async(c->pc);
--- a/src/librados/librados.cc
+++ b/src/librados/librados.cc
@@ -1646,10 +1646,7 @@ struct AioUnlockCompletion : public librados::ObjectOperationCompletion {
 int librados::IoCtx::aio_unlock(const std::string &oid, const std::string &name,
 			        const std::string &cookie, AioCompletion *c)
 {
-  librados::AioCompletion *completion = librados::Rados::aio_create_completion();
-  int rc = rados::cls::lock::aio_unlock(this, oid, name, cookie, completion);
-  completion->release();
-  return rc;
+  return rados::cls::lock::aio_unlock(this, oid, name, cookie, c);
 }
 
 int librados::IoCtx::break_lock(const std::string &oid, const std::string &name,
--- a/src/librados/librados.cc
+++ b/src/librados/librados.cc
@@ -1882,6 +1882,60 @@ int librados::IoCtx::aio_flush()
   return 0;
 }
 
+struct AioGetxattrDataPP {
+  AioGetxattrDataPP(librados::AioCompletionImpl *c, bufferlist *_bl) :
+    bl(_bl), completion(c) {}
+  bufferlist *bl;
+  struct librados::C_AioCompleteAndSafe completion;
+};
+
+static void rados_aio_getxattr_completepp(rados_completion_t c, void *arg) {
+  AioGetxattrDataPP *cdata = reinterpret_cast<AioGetxattrDataPP*>(arg);
+  int rc = rados_aio_get_return_value(c);
+  if (rc >= 0) {
+    rc = cdata->bl->length();
+  }
+  cdata->completion.finish(rc);
+  delete cdata;
+}
+
+int librados::IoCtx::aio_getxattr(const std::string& oid, librados::AioCompletion *c,
+				  const char *name, bufferlist& bl)
+{
+  // create data object to be passed to async callback
+  AioGetxattrDataPP *cdata = new AioGetxattrDataPP(c->pc, &bl);
+  if (!cdata) {
+    return -ENOMEM;
+  }
+  // create completion callback
+  librados::AioCompletionImpl *comp = new librados::AioCompletionImpl;
+  comp->set_complete_callback(cdata, rados_aio_getxattr_completepp);
+  // call actual getxattr from IoCtxImpl
+  object_t obj(oid);
+  return io_ctx_impl->aio_getxattr(obj, comp, name, bl);
+}
+
+int librados::IoCtx::aio_getxattrs(const std::string& oid, AioCompletion *c,
+				   map<std::string, bufferlist>& attrset)
+{
+  object_t obj(oid);
+  return io_ctx_impl->aio_getxattrs(obj, c->pc, attrset);
+}
+
+int librados::IoCtx::aio_setxattr(const std::string& oid, AioCompletion *c,
+				  const char *name, bufferlist& bl)
+{
+  object_t obj(oid);
+  return io_ctx_impl->aio_setxattr(obj, c->pc, name, bl);
+}
+
+int librados::IoCtx::aio_rmxattr(const std::string& oid, AioCompletion *c,
+				 const char *name)
+{
+  object_t obj(oid);
+  return io_ctx_impl->aio_rmxattr(obj, c->pc, name);
+}
+
 int librados::IoCtx::aio_stat(const std::string& oid, librados::AioCompletion *c,
 			      uint64_t *psize, time_t *pmtime)
 {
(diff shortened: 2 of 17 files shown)


liewegas:
@jdurgin

src/civetweb
```diff
@@ -1 +1 @@
-Subproject commit e12494070432b85d5ffce7dece81c0baad457a5f
+Subproject commit 8d271315a541218caada366f84a2690fdbd474a2
```
ktdreyer: Whoops, looks like this civetweb submodule change is an accident

sponce: Indeed, and somehow rocksdb was also changed. I've now fixed it

src/libradosstriper/RadosStriperImpl.cc
```diff
@@ -1054,21 +1210,24 @@ int libradosstriper::RadosStriperImpl::truncate(const std::string& soid,
       uint64_t new_object_size = Striper::object_truncate_size(cct(), &l, objectno, size);
       int rc;
       if (new_object_size > 0 or 0 == objectno) {
+	// trunc is synchronous as there is no async version
+	// but note that only a single object will be truncated
+	// reducing the overload to a fixed amount
 	rc = m_ioCtx.trunc(getObjectId(soid, objectno), new_object_size);
```
jdurgin: it's not strictly necessary to have aio_* versions of all these operations - they're all available as async via ObjectOperation's methods and aio_operate()

src/libradosstriper/RadosStriperImpl.cc
```diff
@@ -880,33 +880,37 @@ libradosstriper::RadosStriperImpl::internal_aio_write(const std::string& soid,
...
-    librados::AioCompletion *rados_completion =
-      m_radosCluster.aio_create_completion(c, rados_req_write_complete, rados_req_write_safe);
-    r = m_ioCtx.aio_write(p->oid.name, rados_completion, oid_bl, p->length, p->offset);
-    rados_completion->release();
-    if (r < 0) 
-      break;
-  }    
+  // Do not try anything if we are called with empty buffer,
+  // file_to_extents would raise an exception
+  if (len > 0) {
```
jdurgin: if len is 0, how does the caller know it's complete? seems like you'd need to complete the MultiAioCompletionImpl

jdurgin:
Sorry for the delayed review. Mostly looks good to me. 

Could you add a yaml fragment to ceph-qa-suite.git under rados/basic/tasks to run the libradosstriper tests? something like 
```
- install:
- ceph:
- exec:
    client.0:
    - ceph_test_rados_striper_api_io
    - ceph_test_rados_striper_api_aio
    - ceph_test_rados_striper_api_striping
```

That would give us some test coverage of the new librados methods too.

src/libradosstriper/RadosStriperImpl.cc
sponce: The caller knows when the MultiAioCompletionImpl notifies it. In the case where it contains no rados_completion, this is when finish_adding_request is called (just after the end of the if you mention, so executed whatever happens). If you look at the implementation of it, it will call safe and complete in case nothing had to be done.
So I believe this is properly covered

sponce:
I've added the yaml file as you suggested and created a pull request : https://github.com/ceph/ceph-qa-suite/pull/1087
Now I have no knowledge of teuthology and no infrastructure to try running the test, so I have no clue whether my file is correct or not !

src/libradosstriper/RadosStriperImpl.cc
jdurgin: ah yes, that makes sense. lgtm then. Now that the lab's back up I'll schedule a test run with your qa-suite addition (which looks fine to me too)

tchaikov:
i will run your changeset through the tests of ceph/ceph-qa-suite#1087.

tchaikov:
@sponce turns out the tests are not packaged yet.
```
2016-08-02T05:57:56.290 INFO:teuthology.orchestra.run.mira051.stderr:bash: ceph_test_rados_striper_api_io: command not found
```

see http://qa-proxy.ceph.com/teuthology/kchai-2016-08-02_04:49:46-rados-wip-10049---basic-mira/346681/teuthology.log

#10541 is posted to address this.

tchaikov:
```
2016-08-02T11:47:55.052 INFO:teuthology.orchestra.run.mira072.stdout:[----------] 60 tests from SimpleStriping/StriperTestRT
2016-08-02T11:47:55.052 INFO:teuthology.orchestra.run.mira072.stdout:[ RUN      ] SimpleStriping/StriperTestRT.StripedRoundtrip/0
2016-08-02T11:47:55.053 INFO:teuthology.orchestra.run.mira072.stderr:terminate called after throwing an instance of 'ceph::buffer::end_of_buffer'
2016-08-02T11:47:55.053 INFO:teuthology.orchestra.run.mira072.stderr:  what():  buffer::end_of_buffer
2016-08-02T11:47:55.053 ERROR:teuthology.run_tasks:Saw exception from tasks.
  File "/home/teuthworker/src/teuthology_master/teuthology/run_tasks.py", line 66, in run_tasks
    manager = run_one_task(taskname, ctx=ctx, config=config)
  File "/home/teuthworker/src/teuthology_master/teuthology/run_tasks.py", line 45, in run_one_task
    return fn(**kwargs)
  File "/home/teuthworker/src/teuthology_master/teuthology/task/exec.py", line 54, in task
    c],
  File "/home/teuthworker/src/teuthology_master/teuthology/orchestra/remote.py", line 196, in run
    r = self._runner(client=self.ssh, name=self.shortname, **kwargs)
  File "/home/teuthworker/src/teuthology_master/teuthology/orchestra/run.py", line 402, in run
    r.wait()
  File "/home/teuthworker/src/teuthology_master/teuthology/orchestra/run.py", line 166, in wait
    label=self.label)
CommandFailedError: Command failed on mira072 with status 134: 'sudo TESTDIR=/home/ubuntu/cephtest bash -c ceph_test_rados_striper_api_striping'
```
see http://qa-proxy.ceph.com/teuthology/kchai-2016-08-02_11:18:38-rados-wip-10049---basic-mira/346864/teuthology.log

sponce:
Finally looking at it ! Sorry for the delay, the holiday period is always a mess as you know...
So there is indeed a bug in my code that was revealed by that test (good to have tests indeed...) and I've managed to run all tests but this one when checking my code...
fixing it !

sponce:
Should be ok now. Tests are passing and I've also double checked that they are clean with valgrind (after I found some leaks in debugging original problem...)

src/include/rados/librados.hpp
```diff
@@ -393,7 +393,7 @@ namespace librados
     void zero(uint64_t off, uint64_t len);
     void rmxattr(const char *name);
     void setxattr(const char *name, const bufferlist& bl);
-    void setxattr(const char *name, const buffer::list&& bl);
+    void setxattr(const char *name, const ceph::buffer::list&& bl);
```
jdurgin: this was recently fixed in master - probably where the merge conflict is coming from. rebase and you can drop this commit including the accidental submodule update

tchaikov:
@sponce could you rebase against master to resolve the conflict?

sponce:
Done

tchaikov:
```
/home/jenkins-build/build/workspace/ceph-pull-requests/src/test/libradosstriper/rados-striper.sh:86: run:  rados --pool rbd --striper rm toyfile
error removing rbd>toyfile: (16) Device or resource busy
/home/jenkins-build/build/workspace/ceph-pull-requests/src/test/libradosstriper/rados-striper.sh:86: run:  return 1
```

@sponce could you take a look at the test failure reported by jenkins ?

sponce:
This was an interesting one ! it is actually a kind of multi client race condition combined with what I believe is a bug of the rados layer. Let me explain the details : when writing a file with striper, the first thing that happens is to lock the while to avoid concurrent writes leading to non consistent content. Of course, the last thing that will happen is to release the lock, and you guessed it, it did not happen here.

More in details, the lock is actually an external attribute on the first stripe of the file (named <filename>_0...0) and the unlocking consists in removing it. Now this removal involves a complete round trip to the osd from the client and thus can be slightly long. So it is now done asynchronously in the code, and this is the new thing of this commit, we did not have an async unlock before. It means that if you would try in the same piece of code to rm the file immediately after the write, you may get "Device or resource busy" indeed if you are too fast. This is the first issue and I should probably fix it. What do you think ?

Now there was another one, as here, we are issuing 2 commands on the command line so we are not in the same program, and thus you expect that the async callback of the unlock of the first program is completed before it exits. Well it is not the case ! It seems that if you exit too fast from the program, the async unlock is actually never performed ! Is that expected ? I suppose it is still pending in the list of commands to be sent to the osd, and there is no code to make the destruction of this list pend on its emptyness, and the completion of the corresponding actions. Am I correct ?

jdurgin:
Nice analysis. For the 1st issue, I agree it should probably be handled by libradosstriper - perhaps keeping track of in-flight lock status, or by retrying the lock when it gets EBUSY.

You're right about the lack of waiting for the 2nd case with an async op still in flight. This is a common issue when not flushing requests - a simple fix may be to call m_ioCtx.aio_flush() whenever the RadosStriperImpl is deleted - e.g. from rados_striper_destroy(), operator=(), etc.

sponce:
Took me longer than I would have liked, but here are the needed fixes I believe.
Note that I could not run the rados-striper.sh test myself, it seems that the test infrastructure fails on my machine with some permission denied. However other tests are passing and conclusive

jdurgin:
The code looks ok to me, but the test script is failing when I run it from the ```build``` dir via ```ctest -R rados-striper.sh```. The test log in build/Testing/Temporary/LastTest.log shows this error:

```
/home/jdurgin/ceph/src/test/libradosstriper/rados-striper.sh:39: run:  dd if=/dev/urandom of=testdir/rados-striper/toyfile bs=1234 count=1
1+0 records in
1+0 records out
1234 bytes (1.2 kB) copied, 0.000244116 s, 5.1 MB/s
/home/jdurgin/ceph/src/test/libradosstriper/rados-striper.sh:42: run:  rados --pool rbd --striper put toyfile testdir/rados-striper/toyfile
2016-09-08 17:07:13.334381 7fd66c29fa00 -1 libradosstriper: RadosStriperImpl::openStripedObjectForWrite : could not load layout and size for toyfile : rc = -2error putting rbd/toyfile: (2) No such file or directory
```

The same error occurs when run the ```rados put``` command with --striper manually.

sponce:
ok, I now took the time of understanding why I could not run the test, and managed after adding some chown in ceph-helpers.sh.
So things should be fine now, and I also took the opportunity to rebase it on top of master.

jdurgin:
thanks! looks good

sponce:
Will this be part of kraken ? It would be really nice, as I keep rebuilding my own clients for now.

src/test/librados/aio.cc
```diff
@@ -1886,6 +2286,38 @@ TEST(LibRadosAio, MultiWritePP) {
...
+    TestAlarm alarm;
+    ASSERT_EQ(0, rados_aio_wait_for_complete(my_completion));
+  }
+  ASSERT_EQ(0, rados_aio_get_return_value(my_completion));
+  ASSERT_EQ(0, rados_lock_exclusive(test_data.m_ioctx, "foo", "TestLock", "Cookie", "", NULL,  0));
+}
+
+TEST(LibRadosAio, AioUnlockPP) {
+  AioTestDataPP test_data;
+  ASSERT_EQ(0, test_data.m_ioctx.lock_exclusive("foo", "TestLock", "Cookie", "", NULL, 0));
```
tchaikov: you forgot to call `test_data.init()`.

src/librados/librados.cc
```diff
@@ -1646,10 +1646,7 @@ struct AioUnlockCompletion : public librados::ObjectOperationCompletion {
 int librados::IoCtx::aio_unlock(const std::string &oid, const std::string &name,
 			        const std::string &cookie, AioCompletion *c)
 {
-  librados::AioCompletion *completion = librados::Rados::aio_create_completion();
-  int rc = rados::cls::lock::aio_unlock(this, oid, name, cookie, completion);
-  completion->release();
-  return rc;
+  return rados::cls::lock::aio_unlock(this, oid, name, cookie, c);
```
tchaikov: could you prefix the title of your commit message with the subcomponent your are changing ? see https://github.com/ceph/ceph/blob/master/SubmittingPatches.rst#preparing-and-sending-patches, in this case, "librados: " would be good enough.

src/test/libradosstriper/aio.cc
```diff
@@ -22,6 +22,7 @@ class AioTestData
   }
 
   ~AioTestData() {
+    sem_unlink("test_libradosstriper_aio_sem");
```
tchaikov: could you prefix the title of your commit message with the subcomponent your are changing ? see https://github.com/ceph/ceph/blob/master/SubmittingPatches.rst#preparing-and-sending-patches, in this case, "test/librados: " or "ceph_test_rados_api_aio: " would be good enough.

tchaikov:
@sponce 

all tests in http://pulpito.ceph.com/kchai-2016-10-24_15:21:51-rados-wip-kefu-testing---basic-mira/ failed, dropping this PR from the test branch.

sponce:
Just back from vacations today and I see that you had trouble with this merge request.
So I've rebased it on master and fixed the small problem you've noticed (as well as the titles of commits).
I've also redone a full run of the tests and they all run fine (as they already were at the time of my last fixes, 1.5 months ago). So my only guess is that we do not run the tests in the same way.

It's quite true that I do not use all the testing infrastructure. To give details, I do the following : 
  - run ctest -R rados-striper.sh for the ctest part
  - start a test cluster via vstart.sh and run manually ceph_test_rados(_striper)_* for the rest
I've also tried the second set via rados/test.sh as it's what fails for you and I needed to manually set my PATH to build/bin directory, but apart from that, it ran fine both in serial and parallel mode.

Anything else I can do for helping ?
I've put back the needs-qa and reassigned tchaikof for now

tchaikov:
> I've also tried the second set via rados/test.sh

that for fixing and testing. i am pulling this PR into my testing branch again.

src/test/librados/aio.cc
```diff
@@ -669,6 +669,406 @@ TEST(LibRadosAio, RoundTripAppendPP) {
...
+  boost::scoped_ptr<AioCompletion> my_completion
+    (test_data.m_cluster.aio_create_completion
+     ((void*)&test_data, set_completion_completePP, set_completion_safePP));
+  std::map<std::string, bufferlist> attrset;
+  ASSERT_EQ(0, test_data.m_ioctx.aio_getxattrs("foo", my_completion.get(), attrset));
+  {
+    TestAlarm alarm;
+    ASSERT_EQ(0, my_completion->wait_for_complete());
+  }
+  ASSERT_EQ(-ENODATA, my_completion->get_return_value());
```
tchaikov: all failed tests in http://pulpito.ceph.com/kchai-2016-11-05_15:35:54-rados-wip-kefu-testing---basic-mira/ are due to this statement. i think we should `s/-ENODATA/0/` here.

sponce:
@tchaikov : Thanks for finding out where the problem was !
I've now fixed it, and the test should pass properly.
However, I do not know how to run pulpito myself so I rely on you to test all this in that environment.

tchaikov:
@sponce you can run rados/test.sh in your test bed. it should cover ceph_test_rados_aio also. and that's the test failed in my last batch.

sponce:
I've been doing that from the beginning actually. But oddly it doesn't fail every time for me, although I've managed to make it fail but running it in a loop yesterday. Anyway, it should now work

tchaikov:

PR #10057, opened 2016-06-30
rgw: remove some useless codes 

Yan-waller:
We have no need to genarate an `oid_rand` and pass it into the `prepare`
function, which will do nothing with `oid_rand` as following:

```
  virtual int prepare(RGWRados *_store, string *oid_rand) {
    store = _store;
    return 0;
  }
```
On the other hand, the `RGWObjManifest::generator::create_begin` will
genarate a different rand `oid` suffix as well as we need.

### DIFF
--- a/src/rgw/rgw_rados.cc
+++ b/src/rgw/rgw_rados.cc
@@ -7367,9 +7367,7 @@ int RGWRados::Object::complete_atomic_modification()
   store->update_gc_chain(obj, state->manifest, &chain);
 
   string tag = state->obj_tag.to_str();
-  int ret = store->gc->send_chain(chain, tag, false);  // do it async
-
-  return ret;
+  return store->gc->send_chain(chain, tag, false);  // do it async
 }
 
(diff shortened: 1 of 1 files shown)


mattbenjamin:
lgtm, and should save a few cycles.

Yan-waller:
@mattbenjamin , @yehudasa  yeah, the processor prepare api seems to be incomplete, maybe it should handle the input parameter `oid_rand`, store or/and set it as prefix of `manifest` (RGWObjManifest.set_prefix()), so that the `RGWObjManifest::generator::create_begin` later can use this `oid_rand` as what we expected ?

Yan-waller:
@mattbenjamin  I have fixed it, could you help to take a look?

mattbenjamin:
@Yan-waller this looks better, @yehudasa ?

yehudasa:
@Yan-waller @mattbenjamin I think that at the moment this code would have zero effect on anything, as we'll only call prepare() with oid_rand not NULL when dealing with multipart upload, in which case this isn't going to be called. It will just add another path of untested code.

Yan-waller:
thanks all, it's my mistake, dropped it.

Yan-waller:
retest this please

PR #10058, opened 2016-06-30
common: a small modifycation about getting rid of unneccessary contio…

shun-s:
…n judgement in loop while

a small modifycation about getting rid of unneccessary contion judgement in loop while

Signed-off-by: shun-s <song.shun3@zte.com.cn>

### DIFF
--- a/src/common/config.cc
+++ b/src/common/config.cc
@@ -689,7 +689,7 @@ int md_config_t::injectargs(const std::string& s, std::ostream *oss)
   char *p = b;
   while (*p) {
     nargs.push_back(p);
-    while (*p && *p != ' ') p++;
+    while (*p != ' ') p++;
     if (!*p)
       break;
     *p++ = 0;


src/common/config.cc
```diff
@@ -689,7 +689,7 @@ int md_config_t::injectargs(const std::string& s, std::ostream *oss)
   char *p = b;
   while (*p) {
     nargs.push_back(p);
-    while (*p && *p != ' ') p++;
+    while (*p != ' ') p++;
```
tchaikov: `(*p != '\0' && *p != ' ')` and `(*p != ' ')` is not equivalent.

shun-s: yeah, it's not equivalent, but it seems to have nothing to do with that.
As (*p != '\0' && *p != ' ') is always running after while(*p) , so " *p != '\0' " is always true. there is no necessary to judge whether *p != '\0' is true again.

tchaikov: please note `p++` after `while (*p && *p != ' ')`

tchaikov:
i am closing this PR as it does not sound right.

src/common/config.cc
shun-s: yeah, you are right, thank you  ^^

PR #10060, opened 2016-06-30
librbd: discard hangs when 'rbd_skip_partial_discard' is enabled

trociny:
Fixes: http://tracker.ceph.com/issues/16386

### DIFF
--- a/src/test/librbd/journal/test_Entries.cc
+++ b/src/test/librbd/journal/test_Entries.cc
@@ -159,6 +159,9 @@ TEST_F(TestJournalEntries, AioWrite) {
 TEST_F(TestJournalEntries, AioDiscard) {
   REQUIRE_FEATURE(RBD_FEATURE_JOURNALING);
 
+  CephContext* cct = reinterpret_cast<CephContext*>(_rados.cct());
+  REQUIRE(!cct->_conf->rbd_skip_partial_discard);
+
   librbd::ImageCtx *ictx;
   ASSERT_EQ(0, open_image(m_image_name, &ictx));
 
--- a/src/librbd/AioImageRequest.cc
+++ b/src/librbd/AioImageRequest.cc
@@ -264,6 +264,8 @@ void AbstractAioImageWrite::send_request() {
                   !m_image_ctx.journal->is_journal_replaying());
   }
 
+  prune_object_extents(object_extents);
+
   if (!object_extents.empty()) {
     uint64_t journal_tid = 0;
     m_aio_comp->set_request_count(
@@ -392,6 +394,24 @@ uint64_t AioImageDiscard::append_journal_event(
   return tid;
 }
 
+void AioImageDiscard::prune_object_extents(ObjectExtents &object_extents) {
+  CephContext *cct = m_image_ctx.cct;
+  if (!cct->_conf->rbd_skip_partial_discard) {
+    return;
+  }
+
+  for (auto p = object_extents.begin(); p != object_extents.end(); ) {
+    if (p->offset + p->length < m_image_ctx.layout.object_size) {
+      ldout(cct, 20) << " oid " << p->oid << " " << p->offset << "~"
+		     << p->length << " from " << p->buffer_extents
+		     << ": skip partial discard" << dendl;
+      p = object_extents.erase(p);
+    } else {
+      ++p;
+    }
+  }
+}
+
 uint32_t AioImageDiscard::get_cache_request_count(bool journaling) const {
   // extra completion request is required for tracking journal commit
   return (m_image_ctx.object_cacher != nullptr && journaling ? 1 : 0);
@@ -415,8 +435,6 @@ void AioImageDiscard::send_cache_requests(const ObjectExtents &object_extents,
 AioObjectRequest *AioImageDiscard::create_object_request(
     const ObjectExtent &object_extent, const ::SnapContext &snapc,
     Context *on_finish) {
-  CephContext *cct = m_image_ctx.cct;
-
   AioObjectRequest *req;
   if (object_extent.length == m_image_ctx.layout.object_size) {
     req = new AioObjectRemove(&m_image_ctx, object_extent.oid.name,
@@ -427,10 +445,6 @@ AioObjectRequest *AioImageDiscard::create_object_request(
                                 object_extent.objectno, object_extent.offset,
                                 snapc, on_finish);
   } else {
-    if(cct->_conf->rbd_skip_partial_discard) {
-      delete on_finish;
-      return NULL;
-    }
     req = new AioObjectZero(&m_image_ctx, object_extent.oid.name,
                             object_extent.objectno, object_extent.offset,
                             object_extent.length, snapc, on_finish);
--- a/src/librbd/AioImageRequest.h
+++ b/src/librbd/AioImageRequest.h
@@ -113,6 +113,8 @@ protected:
 
   virtual void send_request();
 
+  virtual void prune_object_extents(ObjectExtents &object_extents) {
+  }
   virtual uint32_t get_cache_request_count(bool journaling) const {
     return 0;
   }
@@ -179,6 +181,7 @@ protected:
     return "aio_discard";
   }
 
+  virtual void prune_object_extents(ObjectExtents &object_extents) override;
   virtual uint32_t get_cache_request_count(bool journaling) const override;
   virtual void send_cache_requests(const ObjectExtents &object_extents,
                                    uint64_t journal_tid);
--- a/src/test/librbd/fsx.cc
+++ b/src/test/librbd/fsx.cc
@@ -502,7 +502,7 @@ struct rbd_ctx {
 	int krbd_fd;		/* image /dev/rbd<id> fd */ /* reused for nbd test */
 };
 
-#define RBD_CTX_INIT	(struct rbd_ctx) { NULL, NULL, NULL, -1 }
+#define RBD_CTX_INIT	(struct rbd_ctx) { NULL, NULL, NULL, -1}
 
 struct rbd_operations {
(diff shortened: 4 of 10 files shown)


trociny:
@dillaman It fixes hang but I observe this test failure:
```
[ RUN      ] TestInternal.DiscardCopyup
test/librbd/test_internal.cc:682: Failure
Value of: bl.contents_equal(read_bl)
  Actual: false
Expected: true
[  FAILED  ] TestInternal.DiscardCopyup (24 ms)
```
Am I interpreting the test result correctly, that discard became visible for the snapshot?

dillaman:
@trociny On line 676, I hide the fact that this clone has a parent to ensure it doesn't satisfy the read from the parent. If ```rbd_skip_partial_discard``` is enabled, nothing will be updated if it is a partial object discard, so this test should probably be skipped if it's enabled.

dillaman:
lgtm

dillaman:
@trociny I think we should create a ceph-qa-suite test case where we enable this option and run ceph_test_librbd

dillaman:
@trociny ... also, I wonder if we should record this config setting in the discard journal event. This seems like it is the only config option we have that could result in a mismatched image between primary and non-primary should the settings be different between clusters.

trociny:
@dillaman I was observing this failure:
```
[ RUN      ] TestImageSync.Simple
librbd/AioCompletion.cc: In function 'void librbd::AioCompletion::associate_journal_event(uint64_t)' thread 7fef1d70b700 time 2016-07-01 09:24:34.881967
librbd/AioCompletion.cc: 201: FAILED assert(state == STATE_PENDING)
 ceph version 11.0.0-122-gc8919c2 (c8919c2af59a924cf5202894a10afd0ec7ffe6d6)
 1: (ceph::__ceph_assert_fail(char const*, char const*, int, char const*)+0x82) [0x7fef31fa58f2]
 2: (librbd::AioCompletion::associate_journal_event(unsigned long)+0x51) [0x7fef31d937f1]
 3: (librbd::AioImageDiscard::append_journal_event(std::list<librbd::AioObjectRequest*, std::allocator<librbd::AioObjectRequest*> > const&, bool)+0x89) [0x7fef31e82939]
 4: (librbd::AbstractAioImageWrite::send_request()+0x495) [0x7fef31e81e85]
 5: (librbd::AioImageRequest<librbd::ImageCtx>::send()+0xc4) [0x7fef31e83334]
 6: (librbd::AioImageRequestWQ::process(librbd::AioImageRequest<librbd::ImageCtx>*)+0x80) [0x7fef31d97490]
 7: (ThreadPool::worker(ThreadPool::WorkThread*)+0xb7f) [0x7fef31f9652f]
 8: (ThreadPool::WorkThread::entry()+0x10) [0x7fef31f97580]
 9: (()+0x80a4) [0x7fef28ab40a4]
 10: (clone()+0x6d) [0x7fef26df904d]
```
So I have made associate_journal_event() call in AioImageDiscard::append_journal_event() conditional.

May be a more cleaner solution is to add a new AioObjectRequest class to handle rbd_skip_partial_discard (see [DNM] commit)?

Both ways fixes the assert, but this failure is observed:
```
[ RUN      ] TestMockImageSyncObjectCopyRequest.Trim
seed 2386
 wrote [48644~207480]
unknown file: Failure

Unexpected mock function call - returning directly.
    Function call: aio_update(3, 0, 1, '\x1' (1), @0x7f5791ff87b0 --, 0x7f577c000b90)
Google Mock tried the following 2 expectations, but none matched:

test/rbd_mirror/image_sync/test_mock_ObjectCopyRequest.cc:175: tried expectation #0: EXPECT_CALL(mock_object_map, aio_update(snap_id, 0, 1, state, _, _))...
         Expected: all pre-requisites are satisfied
           Actual: the following immediate pre-requisites are not satisfied:
test/rbd_mirror/image_sync/test_mock_ObjectCopyRequest.cc:153: pre-requisite #0
                   (end of pre-requisites)
         Expected: to be called once
           Actual: never called - unsatisfied and active
test/rbd_mirror/image_sync/test_mock_ObjectCopyRequest.cc:175: tried expectation #1: EXPECT_CALL(mock_object_map, aio_update(snap_id, 0, 1, state, _, _))...
  Expected arg #0: is equal to 4
           Actual: 3
         Expected: to be called once
           Actual: never called - unsatisfied and active
```

trociny:
@dillaman I will create  a ceph-qa-suite test for rbd_skip_partial_discard option.

trociny:
> I wonder if we should record this config setting in the discard journal event. This seems like it is the only config option we have that could result in a mismatched image between primary and non-primary should the settings be different between clusters.

I don't have a strong opinion here. For me it might be not necessary and may be not expected behavior for some users. Also taking into account that we plan to replicate metadata options, if this is enabled/disabled per image, we will replicate this.

I don't have an idea how useful (frequently used) this option in real life to bother with this.

dillaman:
@trociny I opened a new ticket to track it as a feature.  It doesn't matter now, but I can definitely see the potential for the future "rbd mirror image verify" command to flag deltas in the image.

dillaman:
@trociny You could introduce a new virtual method ```prune_object_extents``` that, when skip partial discards is enabled, it will yank those corresponding extents.  You would call that before ```AioCompletion::set_request_count```

trociny:
@dillaman Remade using prune_object_extents.

In TestMockImageSyncObjectCopyRequest.Trim for discard to truncate the test object I force setting rbd_skip_partial_discard to false for remote image. Does it looks ok for you?

trociny:
@dillaman ceph-qa tests with rbd_skip_partial_discard option enabled: https://github.com/ceph/ceph-qa-suite/pull/

dillaman:
lgtm

src/test/librbd/journal/test_Entries.cc
```diff
@@ -172,6 +172,12 @@ TEST_F(TestJournalEntries, AioDiscard) {
   ASSERT_EQ(0, c->wait_for_complete());
   c->put();
 
+  CephContext* cct = reinterpret_cast<CephContext*>(_rados.cct());
+  if (cct->_conf->rbd_skip_partial_discard) {
+    ASSERT_FALSE(wait_for_entries_available(ictx));
```
dillaman: The discard journal entry should still be created -- why was this needed?

trociny: I did not see journal entry being created and I thought it was ok.

I suppose for journal entry to be created, `associate_journal_event()` should be called, which is in `AioImageDiscard::append_journal_event()`, and this only happens when sending object extents?

trociny: Actually, recalling you created a ticket to track rbd_skip_partial_discard in journal entries, I was going to discuss this observation but forgot.

trociny: @dillaman updated (skipping the test if partial discards are enabled).

dillaman:
@trociny looks like the python tests need to be updated:

http://qa-proxy.ceph.com/teuthology/jdillaman-2016-07-22_04:36:50-rbd-wip-jd-testing---basic-vps/328662/teuthology.log

trociny:
@dillaman I have updated the tests (since last your review and testing the changes are only in fsx.cc and test_rbd.py)

I run some teuthology tests and no issues are noticed. Tests you reported failed now pass:
```teuthology-suite --suite rbd --ceph wip-mgolub-testing --owner mgolub --email mgolub@mirantis.com --machine-type vps --filter 'rbd/librbd/{cache/none.yaml cachepool/none.yaml clusters/{fixed-3.yaml openstack.yaml} config/skip-partial-discard.yaml fs/xfs.yaml msgr-failures/few.yaml workloads/python_api_tests_with_journaling.yaml}'
http://pulpito.ceph.com/trociny-2016-07-28_01:57:05-rbd-wip-mgolub-testing---basic-vps/

teuthology-suite --suite rbd --ceph wip-mgolub-testing --owner mgolub --email mgolub@mirantis.com --machine-type vps --filter 'rbd/librbd/{cache/none.yaml cachepool/small.yaml clusters/{fixed-3.yaml openstack.yaml} config/skip-partial-discard.yaml fs/xfs.yaml msgr-failures/few.yaml workloads/fsx.yaml}'
http://pulpito.ceph.com/trociny-2016-07-28_01:57:46-rbd-wip-mgolub-testing---basic-vps/
```
Also, I have been running the following tests:
```
teuthology-suite --suite rbd --ceph wip-mgolub-testing --owner mgolub --email mgolub@mirantis.com --machine-type vps --filter 'fsx'
http://pulpito.ceph.com/trociny-2016-07-28_01:58:45-rbd-wip-mgolub-testing---basic-vps/

teuthology-suite --suite rbd --ceph wip-mgolub-testing --owner mgolub --email mgolub@mirantis.com --machine-type vps --filter 'python_api_tests'
http://pulpito.ceph.com/trociny-2016-07-28_02:00:35-rbd-wip-mgolub-testing---basic-vps/

teuthology-suite --suite rbd --ceph wip-mgolub-testing --owner mgolub --email mgolub@mirantis.com --machine-type vps --filter 'c_api_tests'
http://pulpito.ceph.com/trociny-2016-07-28_02:03:05-rbd-wip-mgolub-testing---basic-vps/
```

Most of the tests have finished, so far the reported failures looks only for valgrind tests due to  "'ceph-fs-common-dbg' was not found".

src/rocksdb
```diff
@@ -1 +1 @@
-Subproject commit 6797e6ffacd90470140f64cea068970bb3a30aaf
+Subproject commit 5647fa427c4eab87c0caf8bcbb91f5846a4623f9
```
tchaikov: this change is introduced by mistake.

dillaman: Thanks for the catch

PR #10061, opened 2016-06-30
mon: small optimize on the Elector handle_propose function

songbaisen:
mon: small optimize on the Elector handle_propose function   

### DIFF
--- a/src/mon/Elector.cc
+++ b/src/mon/Elector.cc
@@ -262,9 +262,8 @@ void Elector::handle_propose(MonOpRequestRef op)
   } else {
     // they would win over me
     if (leader_acked < 0 ||      // haven't acked anyone yet, or
-	leader_acked > from ||   // they would win over who you did ack, or
-	leader_acked == from) {  // this is the guy we're already deferring to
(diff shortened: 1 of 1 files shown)


tchaikov:
i think the original code is clearer and more readable.

xiexingguo:
@songbaisen 
I am with @tchaikov.

scienceluo:
@songbaisen 
I am with @tchaikov too ^_^

songbaisen:
@tchaikov  Thank you sir! The change it is look good to me here.There is not need to judge twice.I want the code here more simple.@scienceluo Do you insomnia last night in there.Remember take back a turkey leg for me.

xiexingguo:
> here is not need to judge twice.

But I think we can afford this, imho,  for a better explanation, which I think can be great help to the newcomers to Ceph.

songbaisen:
@xiexingguo  HE HE。

liewegas:
The compiler should optimize this away.. and even if it doesn't this isn't a hot path.  I think readability wins here.  THanks!

PR #10062, opened 2016-06-30
client/filer: cleanup the redundant judgments of _write&&_fallocate

renhwztetecs:
The running time of redundant judgments influence function,especially _write(),
at the same time,Filer::read/read_trunc/write/write_trunc/truncate/zero type should be void, fix it

### DIFF
--- a/src/client/Client.cc
+++ b/src/client/Client.cc
@@ -8585,13 +8585,10 @@ int Client::_write(Fh *f, int64_t offset, uint64_t size, const char *buf,
     unsafe_sync_write++;
     get_cap_ref(in, CEPH_CAP_FILE_BUFFER);  // released by onsafe callback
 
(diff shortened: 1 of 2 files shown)


gregsfortytwo:
Was there something that led you to make these changes? I mean, they might be fine, but there's nothing inherent about the functions that prevents them from returning error codes, just the current implementation. And I have a hard time believing anything you've removed makes a material difference to throughput.

renhwztetecs:
@gregsfortytwo 
yeah!
For throughput do not necessarily see the effect(), but write() this kind of interface, removing unnecessary judgment is significant

jcsp:
Reviewed-by

Performance aside, I'm in favour of always-succeed functions being void.  For external interfaces I see the value of returning int in case we want to change the implementation, but internally I prefer void.

renhwztetecs:
@jcsp 
:+1:

PR #10063, opened 2016-06-30
osd:ReplicatedPG.cc:add assert for obc

yonghengdexin735:


xiexingguo:
This is not right. I understand the `obc->watchers.empty()` could be **true**?

PR #10065, opened 2016-06-30
rbd-mirror: gracefully fail if object map is unavailable

dillaman:
If the exclusive lock was lost due to a watch failure from an
overloaded cluster, gracefully abort the image sync.

Fixes: http://tracker.ceph.com/issues/16558


trociny:
lgtm

PR #10072, opened 2016-06-30
librbd: failed assertion after shrinking a clone image twice

dillaman:
Fixes: http://tracker.ceph.com/issues/16561


trociny:
lgtm

PR #10075, opened 2016-06-30
rgw: add tenant support to multisite sync

cbodley:
Initial support for syncing buckets/objects created by tenant users. Needs more testing and cleanup.

Fixes: http://tracker.ceph.com/issues/16469

### DIFF
--- a/src/rgw/rgw_bucket.cc
+++ b/src/rgw/rgw_bucket.cc
@@ -305,6 +305,21 @@ void rgw_bucket_instance_key_to_oid(string& key)
   }
 }
 
+// convert bucket instance oids back to the tenant/ format for metadata keys.
+// it's safe to parse 'tenant:' only for oids, because they won't contain the
+// optional :shard at the end
+void rgw_bucket_instance_oid_to_key(string& oid)
+{
+  // find first : (could be tenant:bucket or bucket:instance)
+  auto c = oid.find(':');
+  if (c != string::npos) {
+    // if we find another :, the first one was for tenant
+    if (oid.find(':', c + 1) != string::npos) {
+      oid[c] = '/';
+    }
+  }
+}
+
 int rgw_bucket_parse_bucket_instance(const string& bucket_instance, string *target_bucket_instance, int *shard_id)
 {
   ssize_t pos = bucket_instance.rfind(':');
@@ -2069,9 +2084,11 @@ public:
 
     if (!exists || old_bci.info.bucket.bucket_id != bci.info.bucket.bucket_id) {
       /* a new bucket, we need to select a new bucket placement for it */
+      auto key{entry};
+      rgw_bucket_instance_oid_to_key(key);
       string tenant_name;
       string bucket_name;
-      parse_bucket(entry, tenant_name, bucket_name);
+      parse_bucket(key, tenant_name, bucket_name);
 
       rgw_bucket bucket;
       RGWZonePlacementInfo rule_info;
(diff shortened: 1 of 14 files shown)


src/rgw/rgw_bucket.cc
```diff
@@ -295,29 +295,28 @@ int rgw_bucket_instance_remove_entry(RGWRados *store, string& entry, RGWObjVersi
   return store->meta_mgr->remove_entry(bucket_instance_meta_handler, entry, objv_tracker);
 }
 
+// parse instance in format: [tenant:]instance[:shard_id], returning
```
yehudasa: @cbodley this is ambiguous. can be either tenant:instance, or instance:shard_id. Will only work if we're guaranteed that instance does not parse as integer (and shard_id does). Do we want to tie our hands like that?

src/rgw/rgw_bucket.cc
```diff
@@ -319,6 +321,53 @@ int rgw_bucket_parse_bucket_instance(const string& bucket_instance, string *targ
...
 }
 
+// parse key in format: [tenant/]name:instance[:shard_id]
+int rgw_bucket_parse_bucket_key(CephContext *cct, const string& key,
+                                rgw_bucket *bucket, int *shard_id)
+{
+  boost::string_ref name{key};
+  boost::string_ref instance;
+
+  // split tenant:name
```
yehudasa: comment should be corrected? tenant/name

src/rgw/rgw_bucket.cc
cbodley: is it safe to assume that all bucket instances will have the format "name:zone.instance_id.bucket_id" that results from this line in `RGWRados::create_bucket()`?
```c++
snprintf(buf, sizeof(buf), "%s.%llu.%llu", get_zone_params().get_id().c_str(), (long long)iid, (long long)bid);
```
if so, we can resolve the ambiguity by parsing ':' as shard_id only if it comes after the '.'s

yehudasa: actually it's not even safe to assume that all buckets will even have a bucket instance (e.g., buckets created pre dumpling). I'm not sure whether it's going to always have the same format (could be that the zone was added later). We really shouldn't rely on that string format.

cbodley:
@yehudasa I pushed an updated patch set that changes the bucket instance metadata keys to "tenant/" so we don't have to worry about parsing "tenant:" vs. ":shard" in the datalog

currently passes test_multi.py with all 4 combinations of tenant vs. non-tenant user and rgw_override_bucket_index_max_shards=0 vs. 2, so parsing of bucket instance keys is working in all cases

i'm starting to look into testing the backward compatibility

cbodley:
with the addition of the patch "rgw: convert bucket instance listings back to metadata key format", I was able to get preexisting tenant's buckets to sync after running `radosgw-admin data sync init`

cbodley:
also pushed to https://github.com/ceph/ceph/tree/wip-rgw-tenant-sync so i can schedule a teuthology run tomorrow

cbodley:
http://pulpito.ceph.com/cbodley-2016-07-07_11:21:33-rgw-wip-rgw-tenant-sync---basic-mira/ has 6 failed and 3 dead tests. one was a valgrind issue in the osd, and the rest appear to be networking issues (failure to download packages, timeouts, etc)

cbodley:
added a patch to skip datalog entries for missing buckets, so that old entries missing the 'tenant/' won't stay in the error_repo forever

yehudasa:
@cbodley lgtm, should merge if passing tests (and get it backported to jewel)

PR #10080, opened 2016-07-01
os/bluestore: improve required compression threshold

xiexingguo:
…ssion

Require the net gain of compression at least to be at a specified ratio,
otherwise we don't compress.

Ask for compressing at least 12.5% off, by default.

This is for the sake of performance because if the compression turns out
to be meaningless(saving little space), we can simply shut it down, as we
know the compression/decompression can be rather CPU-consuming.

### DIFF
--- a/src/os/bluestore/BlueStore.cc
+++ b/src/os/bluestore/BlueStore.cc
@@ -5943,8 +5943,12 @@ int BlueStore::_do_alloc_write(
       compressed_bl.claim_append(t);
       uint64_t rawlen = compressed_bl.length();
       uint64_t newlen = ROUND_UP_TO(rawlen, min_alloc_size);
-      if (newlen < final_length) {
-	// pad out to min_alloc_size
+      uint64_t dstlen = final_length *
+        g_conf->bluestore_compression_required_ratio;
+      dstlen = ROUND_UP_TO(dstlen, min_alloc_size);
+      if (newlen <= dstlen && newlen < final_length) {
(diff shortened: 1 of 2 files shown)


src/os/bluestore/BlueStore.cc
```diff
@@ -5943,8 +5943,12 @@ int BlueStore::_do_alloc_write(
       compressed_bl.claim_append(t);
       uint64_t rawlen = compressed_bl.length();
       uint64_t newlen = ROUND_UP_TO(rawlen, min_alloc_size);
-      if (newlen < final_length) {
-	// pad out to min_alloc_size
+      uint64_t dstlen = final_length *
+        g_conf->bluestore_compression_required_ratio;
+      dstlen = ROUND_UP_TO(dstlen, min_alloc_size);
+      if (newlen < dstlen) {
```
liewegas: i would do <=, as .125 is 1/8 and it is likely we'll actually hit that pretty frequently.  maybe also keep newlen < final_length in there too, as the configured min after rounding up might == final_length?

liewegas:
At some point we probably also want a heuristic that makes us stop trying to compress writes to an object if they never compress down..

xiexingguo:
@liewegas Done.



> At some point we probably also want a heuristic that makes us stop trying to compress writes to an object if they never compress down..

Well, I'll try...

liewegas:
we don't have to do it now, but we should think about how it might be 
done.  it'll be tricky!

xiexingguo:
Okay, do you still want this change?

PR #10081, opened 2016-07-01
cmake: remove unnecessary linked libs from libcephfs

tchaikov:
* some of the libs shares the same .cc which has static C++ variables. if
  we link against the different libs sharing the same static C++
  variables, and the dtor of the C++ variables has side-effects, among
  other things, deallocates a memory chunk. then, we are in the trouble of
  double free.
* some of the libs are referenced by the linked lib, so no need to link
  against them again. for example, BLKID_LIBRARIES are linked by
  libcommon, so we can remove it from the linked libs list.


jcsp:
Tested this locally

PR #10084, opened 2016-07-01
 mon: small optimize on the Paxos store_state function

songbaisen:
mon: small optimize on the Paxos store_state function

When we handle the commit message from other paxos mon.If the last_committed is  equal to us,So 

just quick return.Nothing to new store.

### DIFF
--- a/src/mon/Paxos.cc
+++ b/src/mon/Paxos.cc
@@ -359,6 +359,11 @@ bool Paxos::store_state(MMonPaxos *m)
     return false;
   }
 
+  if (last_committed ==  m->last_committed) {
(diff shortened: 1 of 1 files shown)


dachary:

dachary:
please update the commit message (not the github pull request message: it's different). Make it a detailed explanation of why this optimization is a good thing and how it can be measured.

songbaisen:
@dachary Thank you for review.Done it.

dachary:
the logic to figure out if something needs to be stored is different from what you wrote and is not correct. The original logic must be preserved.

PR #10089, opened 2016-07-01
test: fix usage info of omapbench

gaowanlong:
Fix the default value info of "--omaptype" and the wrong indent
of "--name" option.

Signed-off-by: Wanlong Gao <wanlong.gao@easystack.cn>

### DIFF
--- a/src/test/omap_bench.cc
+++ b/src/test/omap_bench.cc
@@ -84,9 +84,9 @@ int OmapBench::setup(int argc, const char** argv) {
       	   << "rand for random sizes between\n"
       	   << "                        0 and max size, uniform for all sizes"
       	   << " to be specified size.\n"
-      	   << "                        (default "<<value_size;
-      cout <<"\n  --name          the rados id to use (default "<<rados_id;
-      cout<<")\n";
+           << "                        (default uniform)\n";
+      cout << "	--name          the rados id to use (default "<< rados_id
+           << ")\n";
       exit(1);
     }
   }


src/test/omap_bench.cc
```diff
@@ -84,9 +84,9 @@ int OmapBench::setup(int argc, const char** argv) {
       	   << "rand for random sizes between\n"
       	   << "                        0 and max size, uniform for all sizes"
       	   << " to be specified size.\n"
-      	   << "                        (default "<<value_size;
-      cout <<"\n  --name          the rados id to use (default "<<rados_id;
-      cout<<")\n";
+           << "                        (default uniform)\n";
+      cout << "	--name          the rados id to use (default "<<rados_id
```
tchaikov: i think the ident of `--name` should be a single space? and might want to add a space before `rados_id`.

gaowanlong: Nope, it's a tab, if you can do "set list" in vim, you can see that other lines are all tabs, that why the "--name" option is not aligned with others. 
While the space before rados_id, anyway, I didn't add the space just want to keep consistent with others above it, like "threads", "objests", etc.

tchaikov: >  I didn't add the space just want to keep consistent with others above it, like "threads", "objests", etc.
please take this chance to add the space since you are change this line anyway.

gaowanlong: > please take this chance to add the space since you are changing this line anyway.

@tchaikov OK, updated now. ;)

tchaikov:
lgtm

PR #10056, opened 2016-06-30
msg/async/AsyncMessenger: make sure all connection cleanup handler finished

yuyuyu101:
Signed-off-by: Haomai Wang <haomai@xsky.com>

PR #10090, opened 2016-07-01
msg/async: remove file event lock

yuyuyu101:
prepare for the next async messenger backend framework

### DIFF
--- a/src/msg/async/AsyncConnection.cc
+++ b/src/msg/async/AsyncConnection.cc
@@ -1331,7 +1331,7 @@ ssize_t AsyncConnection::_process_connection()
         if (net.set_nonblock(sd) < 0)
           goto fail;
 
-        net.set_socket_options(sd);
+        net.set_socket_options(sd, async_msgr->cct->_conf->ms_tcp_nodelay, async_msgr->cct->_conf->ms_tcp_rcvbuf);
         net.set_priority(sd, async_msgr->get_socket_priority());
 
         bl.append(CEPH_BANNER, strlen(CEPH_BANNER));
--- a/src/msg/async/AsyncConnection.cc
+++ b/src/msg/async/AsyncConnection.cc
@@ -362,14 +362,21 @@ ssize_t AsyncConnection::_try_send(bool more)
                              << " remaining bytes " << outcoming_bl.length() << dendl;
 
   if (!open_write && is_queued()) {
-    center->create_file_event(sd, EVENT_WRITABLE, write_handler);
-    open_write = true;
+    if (center->in_thread()) {
+      center->create_file_event(sd, EVENT_WRITABLE, write_handler);
+      open_write = true;
+    } else {
+      center->dispatch_event_external(write_handler);
+    }
   }
 
   if (open_write && !is_queued()) {
-    center->delete_file_event(sd, EVENT_WRITABLE);
-    open_write = false;
-
+    if (center->in_thread()) {
+      center->delete_file_event(sd, EVENT_WRITABLE);
+      open_write = false;
+    } else {
+      center->dispatch_event_external(write_handler);
+    }
     if (state_after_send != STATE_NONE)
       center->dispatch_event_external(read_handler);
   }
@@ -1333,6 +1340,7 @@ ssize_t AsyncConnection::_process_connection()
 
         net.set_socket_options(sd, async_msgr->cct->_conf->ms_tcp_nodelay, async_msgr->cct->_conf->ms_tcp_rcvbuf);
         net.set_priority(sd, async_msgr->get_socket_priority());
+        center->create_file_event(sd, EVENT_READABLE, read_handler);
 
         bl.append(CEPH_BANNER, strlen(CEPH_BANNER));
 
@@ -1796,15 +1804,7 @@ ssize_t AsyncConnection::handle_connect_msg(ceph_msg_connect &connect, bufferlis
       existing->is_reset_from_peer = true;
     }
 
-    // Now existing connection will be alive and the current connection will
-    // exchange socket with existing connection because we want to maintain
-    // original "connection_state"
-    if (existing->sd >= 0)
-      existing->center->delete_file_event(existing->sd, EVENT_READABLE|EVENT_WRITABLE);
     center->delete_file_event(sd, EVENT_READABLE|EVENT_WRITABLE);
-    existing->center->create_file_event(sd, EVENT_READABLE, existing->read_handler);
-
-    reply.global_seq = existing->peer_global_seq;
 
     // Clean up output buffer
     existing->outcoming_bl.clear();
@@ -1815,7 +1815,13 @@ ssize_t AsyncConnection::handle_connect_msg(ceph_msg_connect &connect, bufferlis
     existing->requeue_sent();
     existing->reset_recv_state();
 
-    swap(existing->sd, sd);
+    int new_fd = sd;
+    int pre_exist_fd = existing->sd;
+    std::swap(existing->sd, sd);
+    _stop();
+    // queue a reset on the new connection, which we're dumping for the old
+    dispatch_queue->queue_reset(this);
+    ldout(async_msgr->cct, 1) << __func__ << " stop myself to swap existing" << dendl;
     existing->can_write = WriteStatus::NOWRITE;
     existing->open_write = false;
     existing->replacing = true;
@@ -1827,17 +1833,22 @@ ssize_t AsyncConnection::handle_connect_msg(ceph_msg_connect &connect, bufferlis
     assert(recv_start == recv_end);
 
     existing->write_lock.Unlock();
-
-    ldout(async_msgr->cct, 1) << __func__ << " stop myself to swap existing" << dendl;
-    _stop();
-    // queue a reset on the new connection, which we're dumping for the old
-    dispatch_queue->queue_reset(this);
-    int new_fd = existing->sd;
-    center->submit_to(existing->center->get_id(), [existing, new_fd, connect, reply, authorizer_reply]() mutable {
+    // existing->sd now isn't registering any event while it's new,
+    // previous existing->sd now is closed, no event will notify
+    // existing(EventCenter*) from now.
+    center->submit_to(existing->center->get_id(), [existing, pre_exist_fd, new_fd, connect, reply, authorizer_reply]() mutable {
       Mutex::Locker l(existing->lock);
       if (new_fd != existing->sd)
         return ;
 
+      if (existing->state != STATE_ACCEPTING_WAIT_CONNECT_MSG) {
+        existing->fault();
+        return ;
+      }
+      reply.global_seq = existing->peer_global_seq;
+      if (pre_exist_fd >= 0)
+        existing->center->delete_file_event(pre_exist_fd, EVENT_READABLE|EVENT_WRITABLE);
+      existing->center->create_file_event(new_fd, EVENT_READABLE, existing->read_handler);
       if (existing->_reply_accept(CEPH_MSGR_TAG_RETRY_GLOBAL, connect, reply, authorizer_reply) < 0) {
         // handle error
         existing->fault();
@@ -1961,7 +1972,6 @@ void AsyncConnection::accept(int incoming)
   Mutex::Locker l(lock);
   sd = incoming;
   state = STATE_ACCEPTING;
-  center->create_file_event(sd, EVENT_READABLE, read_handler);
   // rescheduler connection in order to avoid lock dep
   center->dispatch_event_external(read_handler);
 }
@@ -2231,8 +2241,6 @@ void AsyncConnection::_stop()
 
   ldout(async_msgr->cct, 1) << __func__ << dendl;
   Mutex::Locker l(write_lock);
-  if (sd >= 0)
-    center->delete_file_event(sd, EVENT_READABLE|EVENT_WRITABLE);
 
   reset_recv_state();
   dispatch_queue->discard_queue(conn_id);
@@ -2244,14 +2252,8 @@ void AsyncConnection::_stop()
   open_write = false;
   can_write = WriteStatus::CLOSED;
   state_offset = 0;
-  if (sd >= 0) {
-    shutdown_socket();
-    ::close(sd);
-  }
-  sd = -1;
   // Make sure in-queue events will been processed
   center->dispatch_event_external(EventCallbackRef(new C_clean_handler(this)));
-
 }
 
 void AsyncConnection::prepare_send_message(uint64_t features, Message *m, bufferlist &bl)
--- a/src/msg/async/AsyncConnection.cc
+++ b/src/msg/async/AsyncConnection.cc
@@ -950,8 +950,6 @@ void AsyncConnection::process()
 
       case STATE_CLOSED:
         {
-          if (sd >= 0)
-            center->delete_file_event(sd, EVENT_READABLE);
           ldout(async_msgr->cct, 20) << __func__ << " socket closed" << dendl;
           break;
         }
--- a/src/msg/async/AsyncConnection.cc
+++ b/src/msg/async/AsyncConnection.cc
@@ -82,7 +82,7 @@ class C_clean_handler : public EventCallback {
  public:
   explicit C_clean_handler(AsyncConnectionRef c): conn(c) {}
   void do_request(int id) {
-    conn->cleanup_handler();
+    conn->cleanup();
     delete this;
   }
 };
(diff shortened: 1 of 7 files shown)


yuyuyu101:
passed http://pulpito.ceph.com/haomai-2016-07-05_12:29:30-rados-wip-haomai-testing---basic-smithi/

http://pulpito.ceph.com/haomai-2016-07-05_21:09:56-rados-wip-haomai-testing-distro-basic-smithi/

yuyuyu101:
http://pulpito.ceph.com/haomai-2016-07-05_12:29:30-rados-wip-haomai-testing---basic-smithi

src/msg/async/net_handler.h
```diff
@@ -30,6 +30,7 @@ namespace ceph {
     int set_nonblock(int sd);
     void set_close_on_exec(int sd);
     void set_socket_options(int sd);
+    void set_socket_options(int sd, bool nodelay, int size);
```
tchaikov: now we have two `set_socket_options()`?

src/msg/async/AsyncMessenger.cc
```diff
@@ -106,6 +106,7 @@ int Processor::bind(const entity_addr_t &bind_addr, const set<int>& avoid_ports)
   }
   net.set_close_on_exec(listen_sd);
   net.set_socket_options(listen_sd);
+  net.set_socket_options(listen_sd, msgr->cct->_conf->ms_tcp_nodelay, msgr->cct->_conf->ms_tcp_rcvbuf);
```
tchaikov: and call `net.set_socket_options()` twice? might need to squash this commit into the previous one, to make sure every commit is compilable and atomic.

yuyuyu101:
@tchaikov thanks!!! it's a rebase conflicting

src/msg/async/Event.cc
```diff
@@ -126,10 +126,8 @@ EventCenter::~EventCenter()
   }
   assert(time_events.empty());
 
-  if (notify_receive_fd >= 0) {
-    delete_file_event(notify_receive_fd, EVENT_READABLE);
+  if (notify_receive_fd >= 0)
```
tchaikov: could you explain why there is no need to delete `delete_file_event()` in the dtor?

src/msg/async/AsyncConnection.cc
```diff
@@ -362,14 +362,21 @@ ssize_t AsyncConnection::_try_send(bool more)
                              << " remaining bytes " << outcoming_bl.length() << dendl;
 
   if (!open_write && is_queued()) {
-    center->create_file_event(sd, EVENT_WRITABLE, write_handler);
-    open_write = true;
+    if (center->in_thread()) {
```
tchaikov: could you elaborate why we would want to move the `{create,delete}_file_event()` calls into event thread in the commit message?

src/msg/async/Event.cc
yuyuyu101: because delete_file_event is used to delete fd from epoll pool. And we will the whole epoll in the following.

src/msg/async/AsyncConnection.cc
yuyuyu101: sure.

The main idea is we want to remove file_lock in EventCenter, so each event thread which own EventCenter must operate EventCenter in local thread. And it means other threads can't operate others EventCenter and we must move EventCenter operations from other threads into local event thread.

I'm squash commit message into commit at last

wjwithagen: @yuyuyu101 
That will very likely also fix my problem with Kqueue. Where I suspect that the handles are not kept when threads are forked. Hence the requirement to open an new Kqueue-handle.
But I've been side-tracked from this problem by the desire to first go to Cmake.
That is mostly done, so I'll soon return to looking into this

src/msg/async/Event.cc
tchaikov: will what? i guess you mean "nuke" or something? and what does "in the following" stands for? seems this commit does not look right by its own, could you move it to the commit where "we will the nuke whole epoll"?

yuyuyu101: 'delete driver"

yuyuyu101: sorry, delete the epoll

tchaikov:
> so each event thread which own EventCenter must operate EventCenter in local thread

i see you want to have a 1:1 mapping from EventCenter to thread, as you label this PR with performance,  maybe you can put more details on the rationale of why this improves the performance in one of the commits, and accompany it with a micro benchmark on the improvement?

yuyuyu101:
@tchaikov actually the performance isn't the main purpose of this PR. Local eventcenter is required by dpdk. Since this pr is a part of https://github.com/ceph/ceph/pull/9230

yuyuyu101:
the performance is the side effect

tchaikov:
okay, i can hardly tell the actually intention of this move without the context.

yuyuyu101:
@tchaikov sorry, I need to clarify at first.

tchaikov:
sorry, i am not familiar with this area or its background. what i am interested was #10056 which addresses some of qa run failures. let me know if you could document the commit a little bit. or maybe we need some other expert in (async) msgr to review this changeset.

src/msg/async/AsyncConnection.cc
```diff
@@ -1816,46 +1822,85 @@ ssize_t AsyncConnection::handle_connect_msg(ceph_msg_connect &connect, bufferlis
...
+          existing->worker->references--;
+          new_worker->references++;
+          existing->logger = new_worker->get_perf_counter();
+          existing->worker = new_worker;
+          existing->center = new_center;
+          if (existing->delay_state)
+            existing->delay_state->set_center(new_center);
+        } else if (existing->state == STATE_CLOSED) {
+          ::close(new_fd);
+          return ;
```
tchaikov: could remove this line.

yuyuyu101: no, if continue, we will dispatch the later event which we wan't

yuyuyu101:
passed http://pulpito.ceph.com/haomai-2016-07-08_00:06:29-rados-wip-haomai-testing-distro-basic-smithi/

yuyuyu101:
passed http://pulpito.ceph.com/haomai-2016-07-08_09:55:10-rados-wip-haomai-testing-distro-basic-smithi/

yuyuyu101:
.... I would like to merge this firstly since it's still a middle stage from dpdk pull request.

src/msg/async/Event.cc
```diff
@@ -255,6 +255,7 @@ void EventCenter::delete_time_event(uint64_t id)
   }
 
   time_events.erase(it->second);
+  event_map.erase(it);
```
tchaikov: could you make this a dedicated commit or at least note this in the commit message?

yuyuyu101: ok

src/msg/async/AsyncConnection.cc
```diff
@@ -82,7 +82,7 @@ class C_clean_handler : public EventCallback {
  public:
   explicit C_clean_handler(AsyncConnectionRef c): conn(c) {}
   void do_request(int id) {
-    conn->cleanup_handler();
+    conn->cleanup();
```
tchaikov: > It needs all fd used locally which not like kernel socket shared by all cores.

so you mean, we need to pin the fds to its own EventCenter to prepare for the DPDK?  but how is this mapped to DPDK?

yuyuyu101: On Tue, Jul 12, 2016 at 5:37 PM, Kefu Chai <notifications@github.com> wrote:

> In src/msg/async/AsyncConnection.cc
> <https://github.com/ceph/ceph/pull/10090#discussion_r70406327>:
>
> > @@ -82,7 +82,7 @@ class C_clean_handler : public EventCallback {
> >   public:
> >    explicit C_clean_handler(AsyncConnectionRef c): conn(c) {}
> >    void do_request(int id) {
> > -    conn->cleanup_handler();
> > +    conn->cleanup();
>
> It needs all fd used locally which not like kernel socket shared by all
> cores.
>
> so you mean, we need to pin the fds to its own EventCenter to prepare for
> the DPDK? but how is this mapped to DPDK?
>
dpdk stack will be the driver of eventcenter, like EventEpoll now. It will
be EventDPDK. So pin eventcenter means pin dpdk stack instance too.


> —
> Reply to this email directly, view it on GitHub
> <https://github.com/ceph/ceph/pull/10090/files/63d3fbd01f5f15e06477b891c22fd9963af2599b..0ef50041a48f7efaf583ce1bbfb9e56138b3593a#r70406327>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AA72YQgG0YRmmYyFMDjbnnT-QCLBuDhjks5qU2BcgaJpZM4JDDWp>
> .
>



-- 

Best Regards,

Wheat

src/msg/async/AsyncConnection.cc
```diff
@@ -1816,46 +1822,85 @@ ssize_t AsyncConnection::handle_connect_msg(ceph_msg_connect &connect, bufferlis
...
+        }
+      }
 
-      if (existing->state != STATE_ACCEPTING_WAIT_CONNECT_MSG) {
-        existing->fault();
+      // Before changing existing->center, it may already exists some events in existing->center's queue.
+      // Then if we mark down `existing`, it will execute in another thread and clean up connection.
+      // Previous event will result in segment fault
+      auto transfer_existing = [existing, new_fd, connect, reply, authorizer_reply]() mutable {
+      Mutex::Locker l(existing->lock);
```
tchaikov: wrong indent.

src/msg/async/AsyncConnection.cc
```diff
@@ -1816,46 +1822,85 @@ ssize_t AsyncConnection::handle_connect_msg(ceph_msg_connect &connect, bufferlis
...
-        return ;
+    // new sd now isn't registered any event while origin events
+    // have been deleted.
+    // previous existing->sd now is still open, event will continue to
+    // notify previous existing->center from now.
+    // From now, no one will dispatch event to `existing`
+    // Note: we must use async dispatch instead of execute this inline
+    // even existing->center == center. Because we must ensure below
+    // event executed after all pending external events like
+    // "dispatch_state->queue"
```
tchaikov: what "dispatch_state->queue" stands for? should be `dispatch_queue->queue_*` ?

src/test/msgr/test_msgr.cc
```diff
@@ -777,12 +777,24 @@ class SyntheticDispatcher : public Dispatcher {
   }
 
   void ms_handle_fast_connect(Connection *con) {
-    lock.Lock();
+    Mutex::Locker l(lock);
```
tchaikov: so this fixes http://tracker.ceph.com/issues/16554, am i right?

src/test/msgr/test_msgr.cc
```diff
@@ -37,6 +37,14 @@
 
 typedef boost::mt11213b gen_type;
 
+#include "common/dout.h"
```
tchaikov: the change looks fine, but just out of curiosity, why would prefer using logging over cerr in this case?

src/test/msgr/test_msgr.cc
```diff
@@ -278,6 +278,8 @@ TEST_P(MessengerTest, SimpleTest) {
   ASSERT_TRUE(static_cast<Session*>(conn->get_priv())->get_count() == 1);
   client_msgr->shutdown();
   client_msgr->wait();
+  server_msgr->shutdown();
```
tchaikov: might this address http://tracker.ceph.com/issues/16554 also?

src/msg/async/AsyncConnection.cc
```diff
@@ -2083,7 +2083,8 @@ int AsyncConnection::send_message(Message *m)
   } else {
     out_q[m->get_priority()].emplace_back(std::move(bl), m);
     ldout(async_msgr->cct, 15) << __func__ << " inline write is denied, reschedule m=" << m << dendl;
-    center->dispatch_event_external(write_handler);
+    if (can_write != WriteStatus::REPLACING)
+      center->dispatch_event_external(write_handler);
```
tchaikov: okay, iiuc, once we are at `STATE_CONNECTING_READY`, the pending outgoing message will be dispatched.

src/msg/async/AsyncConnection.cc
```diff
@@ -1321,7 +1321,8 @@ ssize_t AsyncConnection::_process_connection()
         async_msgr->ms_deliver_handle_fast_connect(this);
 
         // make sure no pending tick timer
-        center->delete_time_event(last_tick_id);
+        if (last_tick_id)
```
tchaikov: imho, it is not necessary, right? as `last_tick_id` is always greater or equal to 1. if it's 0, `center->delete_time_event()` would be a no-op?

tchaikov: scratch that! as we could be referencing a `center` being changed!

src/msg/async/AsyncConnection.h
```diff
@@ -104,8 +104,16 @@ class AsyncConnection : public Connection {
     return !out_q.empty() || outcoming_bl.length();
   }
   void shutdown_socket() {
-    if (sd >= 0)
+    for (auto &&t : register_time_events)
```
tchaikov: `auto t` would suffice. there is no need to use universal reference here. and the primitive type would be clearer.

src/msg/async/AsyncConnection.h
```diff
@@ -104,8 +104,16 @@ class AsyncConnection : public Connection {
...
+      center->delete_time_event(t);
```
tchaikov: also typo in the title of commit message: s/simply/simplify/

src/msg/async/AsyncConnection.h
```diff
@@ -104,8 +104,16 @@ class AsyncConnection : public Connection {
...
+    register_time_events.clear();
+    center->delete_time_event(last_tick_id);
+    if (sd >= 0) {
+      center->delete_file_event(sd, EVENT_READABLE|EVENT_WRITABLE);
       ::shutdown(sd, SHUT_RDWR);
+      ::close(sd);
```
tchaikov: is sd dupped before? if not, `::close()` would suffice in this case.

src/msg/async/AsyncConnection.h
tchaikov: and strictly speaking this changes the behaviour of `shutdown_socket()`. which has callers besides `AsyncConnection::fault()`. while the commit message implies this change is but a refactor. so this might fix some issue also?

src/msg/async/Event.cc
```diff
@@ -132,7 +128,8 @@ EventCenter::~EventCenter()
     ::close(notify_send_fd);
 
   delete driver;
-  delete notify_handler;
+  if (notify_handler)
```
tchaikov: nit, could just `delete notify_handler`. `delete nullptr` is a no-op in C++.

tchaikov:
reviewed. i don't think this PR is a refactor. it has following non-trivial changes 

- fixes on cleanup
- and shards the fds by the thread handling them
- as a side effect, the lock for inter-thread resource sharing is removed.

yuyuyu101:
@tchaikov thanks!

src/msg/async/AsyncConnection.cc
yuyuyu101: exactly!

src/test/msgr/test_msgr.cc
yuyuyu101: yes, partial fixes

src/test/msgr/test_msgr.cc
yuyuyu101: because I want to get the right sequence which mixes ceph logginer and cerr. Otherwise, cerr output make the logs a little disordered.

src/test/msgr/test_msgr.cc
yuyuyu101: maybe

src/msg/async/AsyncConnection.cc
yuyuyu101: yep!

src/msg/async/AsyncConnection.cc
yuyuyu101: YES! it's surprised that you can get this!!!!

src/msg/async/AsyncConnection.h
yuyuyu101: no, shutdown_socket should do redundant things compared before

src/msg/async/AsyncMessenger.cc
```diff
@@ -419,10 +421,13 @@ void WorkerPool::start()
 {
   if (!started) {
     for (uint64_t i = 0; i < workers.size(); ++i) {
+      pending++;
       workers[i]->create("ms_async_worker");
     }
     started = true;
   }
+  while (pending)
```
yuyuyu101: sure

src/msg/async/Event.cc
yuyuyu101: yes

src/msg/async/AsyncConnection.cc
tchaikov: could you revise the comment accordingly? thanks.

yuyuyu101:
mostly done

PR #10092, opened 2016-07-01
cmake: fixes

dang:
- Add libexecdir to the variables that are replaced in CMake.
- In init-ceph, replace hard coded sbindir with replaced version.


tchaikov:
lgtm

PR #10093, opened 2016-07-01
crush: reset bucket->h.items[i] when removing tree item

tchaikov:
* crush so we don't see the reference after the removing, this keeps
  check_item_loc() happy, and move_bucket() use check_item_loc() to see if
  the removed bucket disappears after the removal.
* test: also add unittest_crush_wrapper::CrushWrapper.insert_item

Fixes: http://tracker.ceph.com/issues/16525

### DIFF
--- a/src/crush/builder.c
+++ b/src/crush/builder.c
@@ -1015,7 +1015,8 @@ int crush_remove_tree_bucket_item(struct crush_bucket_tree *bucket, int item)
 
 		if (bucket->h.items[i] != item)
 			continue;
-		
+
+		bucket->h.items[i] = 0;
 		node = crush_calc_tree_node(i);
 		weight = bucket->node_weights[node];
 		bucket->node_weights[node] = 0;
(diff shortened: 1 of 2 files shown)


tchaikov:
@liewegas mind taking a look?

liewegas:
looks good to me!

tchaikov:
tested at 

xiaoxichen:
@tchaikov , sorry, I tried to reproduce in my env with hammer and jewel but didn't see this bug...but it's really a bug from code...would you mind sharing how this can be trigger?

I did
ceph osd crush add-bucket test root
ceph osd crush add-bucket test1 root
ceph osd crush add-bucket myhost host
ceph osd crush move myhost root=test
ceph osd crush move myhost root=test1

everything looks fine

tchaikov:
@xiaoxichen you could try to reproduce it using the crush map offered by http://tracker.ceph.com/issues/16525.

to trigger this issue, in short

- bucket type should be tree (it's straw2 by default nowadays)
- should be removing a bucket from its parent
- the removed bucket is not the last one

PR #10098, opened 2016-07-01
tools: New "removeall" used to remove head with snapshots

dzafman:
Use --force to allow remove to only remove head object

### DIFF
--- a/cmake/modules/AddCephTest.cmake
+++ b/cmake/modules/AddCephTest.cmake
@@ -2,7 +2,7 @@
 
 #adds makes target/script into a test, test to check target, sets necessary environment variables
 function(add_ceph_test test_name test_path)
-  add_test(NAME ${test_name} COMMAND ${test_path})
+  add_test(NAME ${test_name} COMMAND ${test_path} ${ARGN})
   add_dependencies(tests ${test_name})
   set_property(TEST
     ${test_name}
--- a/src/test/ceph_objectstore_tool.py
+++ b/src/test/ceph_objectstore_tool.py
@@ -171,7 +171,9 @@ def test_failure(cmd, errmsg, tty=False):
         logging.info("Correctly failed with message \"" + matched[0] + "\"")
         return 0
     else:
+        logging.error("Command: " + cmd )
         logging.error("Bad messages to stderr \"" + str(lines) + "\"")
+        logging.error("Expected \"" + errmsg + "\"")
         return 1
 
 
--- a/src/test/ceph_objectstore_tool.py
+++ b/src/test/ceph_objectstore_tool.py
@@ -577,6 +577,75 @@ def test_get_set_inc_osdmap(CFSD_PREFIX, osd_path):
     return errors
 
 
+def test_removeall(CFSD_PREFIX, db, OBJREPPGS, REP_POOL, CEPH_BIN, OSDDIR, REP_NAME, NUM_CLONED_REP_OBJECTS):
+    # Test removeall
+    TMPFILE = r"/tmp/tmp.{pid}".format(pid=os.getpid())
+    nullfd = open(os.devnull, "w")
+    errors=0
+    print "Test removeall"
+    kill_daemons()
+    for nspace in db.keys():
+        for basename in db[nspace].keys():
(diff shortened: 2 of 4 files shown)


dzafman:
Still need to add test case

src/tools/ceph_objectstore_tool.cc
```diff
@@ -1444,20 +1461,52 @@ int do_remove_object(ObjectStore *store, coll_t coll,
...
+    if (!dry_run) {
+      r = remove_object(coll, snapobj, mapper, &_t, &t);
+      if (r < 0)
+        return r;
+    }
   }
 
-  t.remove(coll, ghobj);
+  if (!dry_run)
+    store->apply_transaction(&osr, std::move(t));
```
tchaikov: i think just checking for `dry_run` at this place would be enough. in other words, no need to check for `dry_run` every where.

src/test/ceph_objectstore_tool.py
```diff
@@ -575,6 +575,56 @@ def test_get_set_inc_osdmap(CFSD_PREFIX, osd_path):
...
+
+                    cmd = (CFSD_PREFIX + " '{json}' removeall").format(osd=osd, json=JSON)
+                    logging.debug(cmd)
+                    ret = call(cmd, shell=True, stdout=nullfd, stderr=nullfd)
+                    if ret != 0:
+                        logging.error("removeall failed for {json}".format(json=JSON))
+                        errors += 1
+    vstart(new=False)
+    wait_for_health()
+    cmd = "{path}/rados -p {pool} rmsnap snap1".format(pool=REP_POOL, path=CEPH_BIN)
```
tchaikov: what is this test for?

src/test/ceph_objectstore_tool.py
```diff
@@ -575,6 +575,56 @@ def test_get_set_inc_osdmap(CFSD_PREFIX, osd_path):
...
+                        logging.error("removeall failed for {json}".format(json=JSON))
+                        errors += 1
+
+                    cmd = (CFSD_PREFIX + " '{json}' removeall").format(osd=osd, json=JSON)
+                    logging.debug(cmd)
+                    ret = call(cmd, shell=True, stdout=nullfd, stderr=nullfd)
+                    if ret != 0:
+                        logging.error("removeall failed for {json}".format(json=JSON))
+                        errors += 1
+    vstart(new=False)
```
tchaikov: is there a way so we can check if the object and its snapshots are all removed?

tchaikov:
@dzafman i see you attached "needs-test" label to this PR. so do you plan to add a test in qa suite or want to amend the test in `ceph_objectstore_tool.py` ?

src/test/ceph_objectstore_tool.py
dzafman: This test is to make sure that the object removal correctly updated the SnapMapper so that the snapshot removal doesn't crash the OSD.

src/tools/ceph_objectstore_tool.cc
dzafman: @tchaikov For consistency with the existing code I want to leave the dry_run checking as is for this pull request.  However, I will consider a follow up to simplify every dry_run case throughout the application.  Following your suggestion will produce better error checking.

dzafman:
@tchaikov Check out the additional changes including better testing.  The cmake change isn't actually used.  I used it to add "large" argument for testing in my build tree.

cmake/modules/AddCephTest.cmake
```diff
@@ -2,7 +2,7 @@
 
 #adds makes target/script into a test, test to check target, sets necessary environment variables
 function(add_ceph_test test_name test_path)
-  add_test(NAME ${test_name} COMMAND ${test_path})
+  add_test(NAME ${test_name} COMMAND ${test_path} ${ARGN})
```
tchaikov: yeah, nice to have this feature! could be useful in future.

dzafman:
@tchaikov This is ready to merge once Jenkins test complete successfully.

tchaikov:
lgtm.

PR #10099, opened 2016-07-01
msg/AsynceMessenger: change ptr to reference

stiopaa1:
In Processor class msgr member variable is better suited to be
reference than a pointer - it is never null and we never change it to
point to something else.


yuyuyu101:
I'm not sure it's worth to do this..

yuyuyu101:
closed now, if any reason to reopen, feel free to do

PR #10100, opened 2016-07-01
tools/cephfs: Remove cephfs-data-scan tmap_upgrade

fullerdj:
tmap_upgrade was only for Jewel, so remove it for Kraken.

Fixes: http://tracker.ceph.com/issues/16144


yuyuyu101:
I think we also need to remove https://github.com/ceph/ceph/blob/master/src/test/librados/tmap_migrate.cc#L49

tchaikov:
@yuyuyu101 see #10234.

PR #10101, opened 2016-07-01
os/bluestore: Added libzbc library

shehbazj:
libzbc library provides API that can be used to query SMR drives.

Fixes: http://tracker.ceph.com/issues/16174
Signed-off-by: Shehbaz Jaffer <shehbaz.jaffer@mail.utoronto.ca>


liewegas:
there's a newer branch that does the submodule properly

PR #10102, opened 2016-07-02
mon/MonitorDBStore: do not use snapshot iterator; close on close

liewegas:


liewegas:
@jecluis can you take a look?  My understanding of mon sync is that we always apply the transactions after the sweep across keys, and the transactions are idempotent (puts and deletes), so there is no need to use a snapshotting iterator.  If so, this is the only user in the code base and we can drop it entirely (Yay!).

liewegas:
@chhabaramesh @tchaikov This is blocking the mem kv merge (which doesn't implement snapshot iteration).  @tchaikov do you mind running this through monthrash?  @jecluis your review would be helpful too!  thanks

gregsfortytwo:
I haven't looked at the code, but I think you're right and the mon doesn't need a snapshot iterator.

(Actually, I thought all iterators worked on snapshots at the time this was written?)

chhabaramesh:
@liewegas do you mean that we can remove snapshot iterator code or leave it for compilation and with asserts?

jecluis:
@liewegas missed this; will take a look after dinner.

jecluis:
@liewegas i think this works regardless of having an underlying snapshot or not. I'm quite confident that the way the sync is built - much more so after checking the code - will tolerate not relying on a snapshot iterator.

Reason for this lies in the sync working in one of two ways: either do a full sync, in which case the paxos transactions are always applied in the end; and the !full sync, which solely syncs paxos txs and will apply them as the sync goes.

The snapshot iterators were introduced as part of the first approach to sync, which could potentially end up leaving some garbage in the store if we happened to delete paxos transactions before sync'ing them. I don't believe it happens in the current form of sync because we do sync paxos transactions first.

LGTM

chhabaramesh:
@liewegas I will wait for this PR to merge for my in memory DB PR 
https://github.com/ceph/ceph/pull/9933

src/mon/MonitorDBStore.h
```diff
@@ -662,6 +662,7 @@ class MonitorDBStore
     // there should be no work queued!
     io_work.stop();
     is_open = false;
+    db.reset(NULL);
```
tchaikov: i think we'd better not to `db.reset()`. because, the counter part call of `close()` is `open()`, which just init the db and reset `is_open`.

but the only caller of `MonitorDBStore::close()` i found is ceph_mon.cc and ceph_monstore_tool.cc, neither of them is trying to reuse the db after closing it, though. so it's safe to do so.

tchaikov:
> do you mind running this through monthrash?

i am on it.

tchaikov:
lgtm modulo the nit.

tchaikov:
failures are due to "ceph_test_rados_api_tmap_migrate", which was fixed by #10256.

tchaikov:
another run with the fix: http://pulpito.ceph.com/kchai-2016-07-12_08:41:22-rados:monthrash-wip-kefu-testing---basic-mira/

src/mon/MonitorDBStore.h
liewegas: Good point.  Would love to clean it up, but later.

liewegas:
Thanks!

tchaikov:
all tests passed.

PR #10112, opened 2016-07-02
cmake changes

tchaikov:


src/CMakeLists.txt
```diff
@@ -1091,7 +1091,6 @@ install(PROGRAMS
   ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/ceph-post-file
   ${CMAKE_SOURCE_DIR}/src/ceph-run
   ${CMAKE_SOURCE_DIR}/src/ceph-rest-api
-  ${CMAKE_SOURCE_DIR}/src/vstart.sh
```
alimaredia: @tchaikov I always run vstart out of my build directory and I'm all for removing it, but I have run it from /usr/local/bin. Do you think a subset of ceph developers still runs it that way? @cbodley have anything to weigh in?

tchaikov: it's a developer only script. and only cmake installs it. so no. even if some of us are using it, i'd recommend him/her to run it from the source tree.

alimaredia: Ah I didn't know only cmake installed it. Let's take it out

alimaredia:
This looks good to me. Should I merge, or is there more testing that needs to happen on this branch?

tchaikov:
yes, please. actually they are the result of the tests. and if you get a chance, #10016 is also ready for review. thanks!

PR #10115, opened 2016-07-04
ceph.spec.in: fix rpm package building error

runsisi:
the rpm package building error is as follows:
```
 error: Installed (but unpackaged) file(s) found:
 /usr/lib/udev/rules.d/60-ceph-by-parttypeuuid.rules
```
which was introduced by #9885 


tchaikov:
lgtm

PR #10117, opened 2016-07-04
os/bluestore: fix incorrect pool decoding of bnode

xiexingguo:
"pool" is a pointer, and we shall alter its value instead.


liewegas:
The first patch is right, thanks!  The second one is unnecessary.. the ref_map is used for a single blob, which will never exceed a few megabytes (let alone 4gb).  And the block_mask may be used in the future (and could probably used effectively in several places now).  Can you drop teh last two?  Thanks!

xiexingguo:
Done.

PR #10120, opened 2016-07-04
rgw: remove bucket index objects when deleting the bucket

oritwas:
Fixes: http://tracker.ceph.com/issues/16412


yehudasa:
@oritwas lgtm, needs testing

PR #10123, opened 2016-07-04
librbd: prevent creation of clone from non-primary mirrored image

trociny:
Fixes: http://tracker.ceph.com/issues/16449


dillaman:
lgtm

dillaman:
@trociny The rbd_mirror.sh integration test fails.  I believe it's caused because rbd-mirror daemon cannot create a non-primary clone against the non-primary parent. We need a way to internally disable this check for rbd-mirror's purposes.

trociny:
@dillaman Updated

PR #10125, opened 2016-07-04
msg/AsyncMessenger: remove unused method

stiopaa1:
Change the return type of function send_keepalive to void. It
doesn't make sense to return constant value(zero).


yuyuyu101:
I suggest we remove these lines since it doesn't have user

dachary:

tchaikov:
http://pulpito.ceph.com/kchai-2017-02-18_17:57:32-rados-wip-kefu-testing---basic-mira/

PR #10127, opened 2016-07-05
mon: fix the duplicate osd show on ceph osd df command

songbaisen:
mon: fix the duplicate osd show on ceph osd df command

Fixes: #16510

### DIFF
--- a/src/mon/OSDMonitor.cc
+++ b/src/mon/OSDMonitor.cc
@@ -841,6 +841,7 @@ protected:
 class OSDUtilizationPlainDumper : public OSDUtilizationDumper<TextTable> {
 public:
   typedef OSDUtilizationDumper<TextTable> Parent;
+  set<int> uniq_osd;
 
   OSDUtilizationPlainDumper(const CrushWrapper *crush, const OSDMap *osdmap,
 		     const PGMap *pgm, bool tree) :
@@ -858,6 +859,8 @@ public:
     tbl->define_column("PGS", TextTable::LEFT, TextTable::RIGHT);
     if (tree)
       tbl->define_column("TYPE NAME", TextTable::LEFT, TextTable::LEFT);
+
+    uniq_osd.clear();
 
     Parent::dump(tbl);
 
@@ -889,6 +892,13 @@ protected:
 			 double& var,
 			 const size_t num_pgs,
 			 TextTable *tbl) {
+    //Make sure the osd id is uniq.
+    if (!qi.is_bucket() && uniq_osd.count(qi.id)) {
+      return;
+    }
+    
+    uniq_osd.insert(qi.id);    
+
     *tbl << qi.id
 	 << weightf_t(qi.weight)
 	 << weightf_t(reweight)


songbaisen:
@lande1234 Bingo!

xiaoxichen:
would you mind to share , how did you reproduce the bug?

songbaisen:
@xiaoxichen  Hello.Long time no see. HAHA！ It is very easy to reproduce  this.You can define more crush bucket item include the same osd. Because this command show osd info from crush,but it does not judge   the same osd in diff crush bucket.

xiaoxichen:
how can an osd belongs to more than one bucket?

songbaisen:
@xiaoxichen :rose:  :smile: 

like below:

ID WEIGHT  TYPE NAME             UP/DOWN REWEIGHT PRIMARY-AFFINITY 
-3 0.28000 pod economic                                            
-4 0.28000     row zone0economic                                   
 0 0.28000         osd.0            down  1.00000          1.00000 
-1 0.28000 root default                                            
-2 0.28000     host localhost                                      
 0 0.28000         osd.0            down  1.00000          1.00000

lande1234:
lgtm!

xiaoxichen:
@songbaisen in which use case we would need such osdmap? and how can you generate such map from API?

songbaisen:
@xiaoxichen  :smile: You can learn it from the below doc.

http://docs.ceph.com/docs/master/rados/operations/crush-map/

xiaoxichen:
@songbaisen 
1. pls provide the use case or any advantage that someone need such crushmap( one osd belongs to 2+ bucket)

2. Seems this kind of crushmap cannot generated from "crush add-bucket, crush add, crush move" API , Just can be imported from a manually edited crushmap.

3. if we dont have clear use case why this kind of map is needed, we would better strengthen the check in " set crushmap", to rule out this case.

Make sense?

songbaisen:
@xiaoxichen  Good morning! I think it is the user's right to generate any kind crush map.We should not forbid he or she  behavior to do that.Some osds belong to the diff crush bucket is not very normal.But it may have meaning to the user.I think here  what  we should do  is do not show the repeat osd on ceph osd df command.
@tchaikov  Hi sir! What is you opinion about this issue?

xiaoxichen:
@songbaisen by no means we do some effort to support useless case.

We always hold the constraint that crushmap is a tree-- every item has only one parent. See the logic in get_immediate_parent and get_immediate_parent_id.   Such strange crushmap will break any logic using get_immediate_parent. 

It's more than easy to do trivial fix like this patch, but we have tons of works to do to fix all the bugs around this case....do you still think it worth to ?

songbaisen:
@xiaoxichen  Oh that is true.It is really some  work need to do to allow this.But I am afraid this can strict restriction the user behavior.That only one osd can below to one bucket.I am not very sure we should do this?

xiaoxichen:
@songbaisen ,  so the problem is use case, I think we will only pay this effort if we have very clear use case--- you may contract the bug reporter about why he want to do this, maybe he just dont familiar with the system at all.

songbaisen:
@xiaoxichen  I think his use case is like this:
          He  have one pool use the origin ceph generate crush rule and write some data store to it.But  the osds in his ceph environmental are have diff performance hard disks.Some osds are ssd others are SAS.He want to use all the ssd to run some time sensitive task.But he also do not want to remove these ssd from the origin pool.Because all the ssd have defined as the primary osd in that pool.
So it have the scene one osd below to two bucket.

xiaoxichen:
OK,  the case make sense..would better start a thread in ceph-devel to discuss, as it require lots of change

songbaisen:
@xiaoxichen  Sorry I do not know how to  start a thread in ceph-devel to discuss.Would you help me to do this?

xiaoxichen:
email to ceph-devel mailing list....

songbaisen:
@xiaoxichen  :smile:  :rose:  Ok, I will do that after lunch.

songbaisen:
@xiaoxichen  Hello I have send the email use my company mailbox in 14:00.But until now I have not receive the email..................................

xiaoxichen:
@songbaisen ,pls join the ML first , and pls use plain text

songbaisen:
@xiaoxichen Oh thank you! I have done these.It looks like is my company for safe to do not allow to send email to out. ......................................

songbaisen:
@liewegas  Hello sir! Look at this  problem. here.

liewegas:
It seems like if you're listing things in terms of the tree ('ceph osd df tree') then you'd want to see the OSD in both/all places the osd exists in a bucket.  Even if it was duplicated... what do you think?

songbaisen:
@liewegas  Yes that is true.And according to the @xiaoxichen  opinion we may have a lot of place to change.

PR #10129, opened 2016-07-05
os/filestore/FileJournal.cc:rm split

yonghengdexin735:


yuyuyu101:
it looks useless...

yuyuyu101:
it would be better to fix nit with other useful changes. I suggest freeze these files without outstanding fix.

PR #10130, opened 2016-07-05
src/common/buffer.cc fix judgment for lseek

yonghengdexin735:
lseek()  returns  the resulting offset location as measured in bytes from the beginning of the file.  On error, the value (off_t) -1 is returned and errno is set to indicate the error. And this errno returned ESPIPE.

### DIFF
--- a/src/common/buffer.cc
+++ b/src/common/buffer.cc
@@ -2293,9 +2293,9 @@ int buffer::list::write_fd_zero_copy(int fd) const
    */
   int64_t offset = ::lseek(fd, 0, SEEK_CUR);
   int64_t *off_p = &offset;
-  if (offset < 0 && offset != ESPIPE)
-    return (int) offset;
-  if (offset == ESPIPE)
+  if (offset < 0 && errno != ESPIPE)
+    return -errno;
+  if (errno == ESPIPE)
     off_p = NULL;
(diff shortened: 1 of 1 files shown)


gaowanlong:
better to write the log into commit message, too. otherwise lgtm.

songbaisen:
@gaowanlong  :smile: Hello,Do you remember the  "xiayuhe" stand near by "nanhu" park.

src/common/buffer.cc
```diff
@@ -2293,9 +2293,9 @@ int buffer::list::write_fd_zero_copy(int fd) const
    */
   int64_t offset = ::lseek(fd, 0, SEEK_CUR);
   int64_t *off_p = &offset;
-  if (offset < 0 && offset != ESPIPE)
+  if (offset < 0 && errno != ESPIPE)
     return (int) offset;
```
xiexingguo: How about:
`return -errno`

yonghengdexin735:
@xiexingguo thank you for comment, I fixed it

xiexingguo:
looks good

yuriw:
http://pulpito.ceph.com/yuriw-2016-07-15_16:58:50-rados-wip-yuri-testing2_2016_7_15-distro-basic-smithi/

PR #10132, opened 2016-07-05
ZetaScale integration

efirs:
There is a small HOWTO in the tree: doc/Ceph_Zetascale_HOWTO.txt
It compiles and works in my dev environment, we are doing bigger tests now.


yuyuyu101:
cool, I suggest we split into multi commits? like
1. zetascaleStore
2. kvdb test add
3. integration 1
4. intergration 2
...

efirs:
@yuyuyu101, ok, will do.

markhpc:
excellent!  We should get Ken involved early on to start figuring out how we want to package zetascale as well.

markhpc:
@ktdreyer Looks like zetascale isn't in fedora yet.  Not sure about ubuntu.  Any thoughts on the right way forward?

ktdreyer:
@b-ranto, want to put the zetascale package up for review in Fedora and I'll review it?

b-ranto:
@ktdreyer I can put it on my queue but I'm really more familiar with being a reviewer in fedora than a reviewee. :-)

.gitmodules
```diff
@@ -32,3 +32,7 @@
 [submodule "src/isa-l"]
 	path = src/isa-l
 	url = https://github.com/ceph/isa-l
+[submodule "src/zs"]
+	path = src/zs
+	url = https://github.com/SanDisk-Open-Source/zetascale.git
```
liewegas: we should make the ceph upstream a github.com/ceph clone of this repo so we can have any ceph-specific commits there.  i just created the clone.  we can set up a group to grant commit rights to whoever needs them

autogen.sh
```diff
@@ -28,7 +28,7 @@ else
   exit 1
 fi
 
-if test -d ".git" ; then
+if test -e ".git" ; then
```
liewegas: why?

src/kv/ZetaScaleStore.cc
```diff
@@ -0,0 +1,1030 @@
...
+
+void ZSStore::ZSTransactionImpl::merge(const std::string &prefix,
+				       const std::string &k,
+				       const bufferlist &value)
+{
+  ops.push_back(
+      make_pair(MERGE, make_pair(combine_strings(prefix, k, true), value)));
+}
+
+#define WRITE_BATCH_SIZE 256
```
liewegas: This looks suspicious... it looks like once you get 256 pending k/v updates you do a put.  Is this actually keeping the overall Transaction atomic?

src/kv/ZetaScaleStore.cc
```diff
@@ -0,0 +1,1030 @@
...
+
+    key += '\0';
+    int i = 1;
+    while (i < n_out && values[i].keylen == key.length() + 2 &&
+	   !memcmp(key.c_str(), values[i].key, key.length())) {
+      value_bl.append(bufferptr(values[i].data, values[i].datalen));
+      dtrace << "key=" << decode_key(string(values[i].key, values[i].keylen))
+	     << dendl;
+      free(values[i].data);
+      free(values[i].key);
```
liewegas: There is a buffer constructor that takes over ownership of a malloc'd buffer.. see buffer::claim_malloc().  This would avoid the copy.

somnathr:
@ktdreyer it is added as submodule in the ceph git tree. You should be able to build ZS along with ceph in any OS and then do a 'make install' to install it in proper place.
Since Ceph is the only user as of today (being just open sourced) , we didn't go for packaging it yet.

src/kv/ZetaScaleStore.cc
somnathr: @liewegas we are calling multiput interface of ZS and that is atomic.

efirs: All writes is a part of transaction, so they atomic.

autogen.sh
efirs: In git workspace .git is a file.

src/kv/ZetaScaleStore.cc
efirs: I'll use it, thank you

markhpc:
Retest this please

markhpc:
Note:  The above comment is to trigger jenkins to retest.

dachary:
needs rebase

PR #10133, opened 2016-07-05
test: cephtool/test.sh: fix expect_false() calls

tchaikov:
this fixes expect_false(), we use bash command lines like:

expect_false "ceph osd blacklist ls | grep 192.168.0.1"

to verify that "ceph auth get client.xx | grep caps | grep mon" fails
with non-zero return code. but expect_false() evaluates the command line
using plain "$@", which will send the arguments direct to the shell,
and $0 is "ceph auth get client.xx | grep caps | grep mon", which does
not exist and is not built-in command, so we have failures like:

/home/jenkins-build/build/workspace/ceph-pull-requests/qa/workunits/cephtool/test.sh:
line 32: ceph osd blacklist ls | grep 192.168.0.1: command not found

to support pipe and other stuff in the command line, we need to use
"eval" to evaluate the command line as a whole.

### DIFF
--- a/qa/workunits/cephtool/test.sh
+++ b/qa/workunits/cephtool/test.sh
@@ -486,7 +486,7 @@ function test_auth()
   ceph auth print-key client.xx
   ceph auth print_key client.xx
   ceph auth caps client.xx osd "allow rw"
-  expect_false "ceph auth get client.xx | grep caps | grep mon"
+  expect_false sh <<< "ceph auth get client.xx | grep caps | grep mon"
   ceph auth get client.xx | grep osd | grep "allow rw"
   ceph auth export | grep client.xx
   ceph auth export -o authfile
@@ -1055,14 +1055,14 @@ function test_mon_osd()
   ceph osd dump --format=json-pretty | grep $bl
   ceph osd dump | grep "^blacklist $bl"
   ceph osd blacklist rm $bl
-  expect_false "ceph osd blacklist ls | grep $bl"
+  ceph osd blacklist ls | expect_false grep $bl
 
   bl=192.168.0.1
   # test without nonce, invalid nonce
   ceph osd blacklist add $bl
   ceph osd blacklist ls | grep $bl
   ceph osd blacklist rm $bl
-  expect_false "ceph osd blacklist ls | grep $bl"
+  ceph osd blacklist ls | expect_false grep $expect_false bl
   expect_false "ceph osd blacklist $bl/-1"
   expect_false "ceph osd blacklist $bl/foo"
 
@@ -1073,7 +1073,7 @@ function test_mon_osd()
   ceph osd blacklist add $bl
   ceph osd blacklist ls | grep $bl
   ceph osd blacklist clear
-  expect_false "ceph osd blacklist ls | grep $bl"
+  ceph osd blacklist ls | expect_false grep $bl
 
   #
   # osd crush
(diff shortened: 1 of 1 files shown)


tchaikov:
for an example of the failure, see https://jenkins.ceph.com/job/ceph-pull-requests/8495/consoleFull#12246481177dc3600c-28d8-42e2-9653-291a0c275fdf, 
```
/home/jenkins-build/build/workspace/ceph-pull-requests/qa/workunits/cephtool/test.sh: line 32: ceph osd blacklist ls | grep 192.168.0.1: command not found
```

tchaikov:
@dachary would you please take a look?

dachary:
I think that solution has another problem: it would evaluate expect_false "echo 'foo bar'" as echo foo bar instead of echo 'foo bar'. Given the shell evaluation rules, I don't think there is a reliable and versatile way to provide a script as an argument and that << or <<< should be used instead. For instance:

    expect_false bash <<< "echo foo | grep bar"

Does that make sense ?

tchaikov:
excellent! can we use `expect_false sh << "echo foo | grep bar"` instead? `sh` is likely to be faster than bash, if it points to `dash`.

tchaikov:
and how about using 
```diff
-  expect_false "ceph osd pool get $TEST_POOL_GETSET scrub_min_interval | grep '.'"
+  ceph osd pool get $TEST_POOL_GETSET scrub_min_interval | expect_false grep '.'
```

for simple cases.

tchaikov:
changelog

* use `ceph osd pool get $TEST_POOL_GETSET recovery_priority | expect_false grep '.'` for piped command
* use `expect_false bash <<< "echo foo | grep bar"` for complicated case.

wjwithagen:
@tchaikov 
With the risk of stating the obvious...
But you know that bash allow you to set a variable that determines how pipes return error?
'''
pipefail
                              If  set,  the  return value of a pipeline is the
                              value of the last (rightmost)  command  to  exit
                              with  a non-zero status, or zero if all commands
                              in the pipeline exit successfully.  This  option
                              is disabled by default.
'''

dachary:
:+1:

dachary:
<pre>
</pre>

tchaikov:
@wjwithagen thanks for the quotation, i didn't know this option. but taking it into consideration, how would you suggest we improve this fix? sorry, i don't follow you.

wjwithagen:
@tchaikov It really depends on the type of pipe you have.
And also on how expect_false actually evaluates `$@` internally.

But the option gives you the possibility to actually target the point from where you expect the error.
And it will allow you to detect that 
expect_false ceph osd blacklist ls | grep 192.168.0.1
Fails in either the the expect false of in the grep.

And 
expect_false `ceph osd blacklist ls | grep 192.168.0.1`
would first execute the pipe in a subprocess and return error if any of the commands fail.

Or use the command you suggested:
`ceph osd pool get $TEST_POOL_GETSET recovery_priority | expect_false grep '.'`
And still get an error if your first part errors.

And thus you do not need to wrap the command in "" to send the pipe to expect_false.
And thus prevent the fact that you'd need to unpack a pipe in expect_false, and avoid the "command not found"

qa/workunits/cephtool/test.sh
```diff
@@ -486,7 +486,7 @@ function test_auth()
   ceph auth print-key client.xx
   ceph auth print_key client.xx
   ceph auth caps client.xx osd "allow rw"
-  expect_false "ceph auth get client.xx | grep caps | grep mon"
+  expect_false sh <<< "ceph auth get client.xx | grep caps | grep mon"
```
wjwithagen: @tchaikov 
With pipe fial I thing you could write this as:
`expect_false sh (backtick)ceph auth get client.xx | grep caps | grep mon"(backtick)`

Sorry about the `(backtick)`, but I don't know how to get a ` code format.

tchaikov:
in most case, i think we expect the error in "grep". 

so you are suggesting using the backquotes? i am not sure how we would workout an `expect_false()` for 
```
expect_false `ceph osd blacklist ls | grep 192.168.0.1`
```

wjwithagen:
Eh, right.....
Sort of dumb, now the commandline contains the output of whatever comes out of the pipe.

on the other thing what would be wrong in what you do now at the other places:
`ceph osd blacklist ls | expect_false grep 192.168.0.1`

Main reason why I reacted is that I really don't like the resulting code: `sh <<<` . One really needs to know a lot of bash to fully appriciate the loops you have jumped. And failpipe is a little gem that I am able to fix quite a few of these.

tchaikov:
`<<<` is clearly a bashim. but we have being using this dialect in this script. it's not too hard to guess what it is for.

tchaikov:
the cephtool-test-mds.sh is not found in the jenkins log =(

PR #10139, opened 2016-07-05
cmake: Really add FCGI_INCLUDE_DIR to include_directories for rgw

tserong:
Commit 3cf6c53 was incorrect.  FCGI_INCLUDE_DIR doesn't need to be
set in src/rgw/CMakeLists.txt, but it does need to be set for the
rgw_a target in src/CMakeLists.txt, as well as for the
ceph_test_librgw_file and ceph_test_librgw_file_nfsns targets in
src/test/CMakeLists.txt.  I can only assume that I must not have
done a completely clean rebuild at some point when testing a
reworked version of the earlier commit :-/

This is only a problem for distros that keep the FCGI headers in
/usr/include/fastcgi/ (e.g.: SUSE).

This commit also removes a redundant include of <fcgiapp.h>

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1429,6 +1429,7 @@ if(${WITH_RADOSGW})
   endif(HAVE_SSL)
 
   add_library(rgw_a STATIC ${rgw_a_srcs})
+  target_include_directories(rgw_a PUBLIC ${FCGI_INCLUDE_DIR})
   target_link_libraries(rgw_a librados cls_rgw_client cls_refcount_client
     cls_log_client cls_statelog_client cls_timeindex_client cls_version_client
     cls_replica_log_client cls_user_client curl global expat ${OPENLDAP_LIBS})
--- a/src/test/CMakeLists.txt
+++ b/src/test/CMakeLists.txt
@@ -254,6 +254,7 @@ add_executable(ceph_test_librgw_file
   )
 set_target_properties(ceph_test_librgw_file PROPERTIES COMPILE_FLAGS
   ${UNITTEST_CXX_FLAGS})
+target_include_directories(ceph_test_librgw_file PRIVATE ${FCGI_INCLUDE_DIR})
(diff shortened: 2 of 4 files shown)


tserong:
@theanalyst can you please test to double-check I haven't missed anything?

theanalyst:
@tserong Thanks works now, lgtm

src/test/CMakeLists.txt
```diff
@@ -254,6 +254,7 @@ add_executable(ceph_test_librgw_file
   )
 set_target_properties(ceph_test_librgw_file PROPERTIES COMPILE_FLAGS
   ${UNITTEST_CXX_FLAGS})
+target_include_directories(ceph_test_librgw_file PUBLIC ${FCGI_INCLUDE_DIR})
```
tchaikov: nit, `PRIVATE` would suffice here. as no need to populate ceph_test_librgw_file's INCLUDE_DIRECTORIES to other targets.

tchaikov:
modulo the nit, lgtm.

src/CMakeLists.txt
```diff
@@ -1429,6 +1429,7 @@ if(${WITH_RADOSGW})
   endif(HAVE_SSL)
 
   add_library(rgw_a STATIC ${rgw_a_srcs})
+  target_include_directories(rgw_a PUBLIC ${FCGI_INCLUDE_DIR})
```
tchaikov: @tserong, have you tried to change PUBLIC here to PRIVATE?

tserong: I just did a rebuild with that line changed from PUBLIC to PRIVATE, and it failed, so I think that one needs to remain as-is.

tchaikov: @tserong thanks for testing!

tchaikov:
lgtm.

tchaikov:
@theanalyst very sorry, i forgot to add you as the reviewer. =(

theanalyst:
@tchaikov no problems :)

PR #10141, opened 2016-07-05
msg/AsyncMessenger.cc: remove unneeded cast

stiopaa1:
send_message is a virtual function, so there is no need to be casting
the pointer to AsyncConnection.


tchaikov:
lgtm.

PR #10142, opened 2016-07-05
mds: move Finisher to unlocked shutdown

batrick:
This commit resolves a deadlock reported in i16042 where the thread calling
MDSRankDispatcher::shutdown would hold the mds_lock while asynchronous
callbacks in the Finisher would attempt to lock mds_lock.

For simplicity, I merged the finisher stop with the messenger shutdown as both
need the mds_lock dropped.

Fixes: http://tracker.ceph.com/issues/16042


xiaoxichen:
Manually tested in my env, looks good

PR #10144, opened 2016-07-06
jewel: librbd: fix missing return statement if failed to get mirror image state

dillaman:
http://tracker.ceph.com/issues/16601


dachary:
It passed the rados suite http://tracker.ceph.com/issues/16344#note-18 (except for one known bug that is, I believe, unrelated).

PR #10150, opened 2016-07-06
mon: avoid start election twice when quorum enter

songbaisen:
mon: avoid start election twice when quorum enter

when quorum enter it is no meaning to start election twice

### DIFF
--- a/src/mon/Elector.h
+++ b/src/mon/Elector.h
@@ -425,6 +425,7 @@ class Elector {
    * @post  @p participating is true
    */
   void start_participating();
+
   /**
    * @}
    */
--- a/src/mon/Elector.cc
+++ b/src/mon/Elector.cc
@@ -488,6 +488,5 @@ void Elector::start_participating()
 {
   if (!participating) {
     participating = true;
-    call_election();
   }
 }


liewegas:
It looks like those are the only two callers for start_participating.. why not just change the implementation?

songbaisen:
@ Hello sage! Because when we start election we will check whether we set participating true.So before start election  we must be set the participating true.  And the interface of start_participating can be use for later on.

liewegas:
I don't think we'll have another user, given that the only current user is the admin commands that are there just for debugging purposes anyway.

songbaisen:
@liewegas  Yes sir.Done it.

liewegas:
don't we still need to set teh participating flag = true or false (like your original methods did)?

songbaisen:
@liewegas  Yes sir.Done it.

liewegas:
This looks like the original patch.  start_participating() now has no users.  So we can also remove it.  Or... just change the start_participating to do what your new begin_participating does.  (Which gets us to the same point, except the name is different.)

Either way, let's get rid of the old begin_participating() implementation which is no longer needed?

@jecluis right?

songbaisen:
@liewegas  Oh I got it! Thank you sir!

dachary:

src/mon/Elector.h
```diff
@@ -425,9 +425,6 @@ class Elector {
    * @post  @p participating is true
    */
   void start_participating();
-  /**
```
jecluis: why are you removing the closing tag for this doxygen defgroup? I'm assuming this was a mistake.

songbaisen:
@jecluis Done it!

yuriw:
@liewegas this was tested pls merge

http://pulpito.front.sepia.ceph.com:80/yuriw-2017-02-06_01:16:11-rados-wip-yuri-testing_2017_2_6-distro-basic-smithi/

http://pulpito.ceph.com/yuriw-2017-02-06_15:38:04-rados-wip-yuri-testing_2017_2_6---basic-smithi/

added infra fix  ceph/ceph: Pull Request 13279

http://pulpito.ceph.com/yuriw-2017-02-06_23:30:25-rados-wip-yuri-testing_2017_2_6---basic-smithi/

http://pulpito.front.sepia.ceph.com:80/yuriw-2017-02-07_16:02:15-rados-wip-yuri-testing_2017_2_6-distro-basic-smithi/

PR #10151, opened 2016-07-06
rgw: register the correct handler for cls_user_complete_stats

oritwas:
Fixes: http://tracker.ceph.com/issues/16624


theanalyst:
this might be worth a backport as well?

yehudasa:
@oritwas @theanalyst digging into it a bit, it seems to me that this doesn't have any practical effect on current code, as we never use the method handle (e.g., for unregistering the method). I think that we don't need to backport it.

PR #10154, opened 2016-07-06
cmake: builds with WITH_MDS=OFF

tchaikov:

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -952,6 +952,9 @@ if(${HAVE_LIBFUSE})
 endif(${HAVE_LIBFUSE})
 install(TARGETS ceph-osd DESTINATION bin)
 
+if(WITH_MDS OR WITH_LIBCEPHFS)
+  add_library(mds_flock_obj OBJECT mds/flock.cc)
+endif()
 # MDS
 if(${WITH_MDS})
   set(mds_srcs
@@ -959,7 +962,6 @@ if(${WITH_MDS})
     mds/MDSDaemon.cc
     mds/MDSRank.cc
     mds/Beacon.cc
-    mds/flock.cc
     mds/locks.c
     mds/journal.cc
     mds/Server.cc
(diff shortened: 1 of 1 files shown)


alimaredia:
@tchaikov when I built this it didn't work because the targets in src/test/mds/CMakeLists.txt link lib mds. Also several targets need this library as well, such as cephfs-data-scan, ceph_test_rados_api_watch_notify, cephfs-table-tool, and cephfs-journal-tool. 

Back when I started doing the CMake work at CohortFS I added the option to build without the MDS because we wanted the ability not to build parts of the code base we weren't doing work on for faster builds and because the option was there in automake. Are there other reasons we still have/need this option? It's not used in any #define's in .cc files either. I'm not opposed to keeping it I just wanted to get clarity on why we have this option today.

@wjwithagen you probably want this option because of fuse issues in your FreeBSD port right?

wjwithagen:
@alimaredia 
These are the flags I'm currently using (defined a few myself)
```
        -D CMAKE_BUILD_TYPE=DEBUG \
        -D WITH_UDEV=OFF \
        -D WITH_BLKID=OFF \
        -D WITH_FUSE=OFF \
        -D WITH_AIO=OFF \
        -D WITH_RBD=OFF \
        -D WITH_XFS=OFF \
        -D WITH_KVS=OFF \
        -D WITH_MANPAGE=OFF \
        -D WITH_LIBCEPHFS=OFF \
        -D WITH_CEPHFS=OFF \
        -D WITH_RADOSGW=OFF
```
So there is no WITH_MDS=OFF, but I've taken FUSE out of several other linkage, so that helps there. Where it could go out, I did so conditioned on WITH_FUSE
And I end up with 133 tests (of which 4 fail) and most of the test that are not included have to do with RBD, FUSE or BLKID.

So I think that FreeBSD is going to be oke, if we remove WITH_MDS.

tchaikov:
thanks for testing and explaining why we have this option.

> Are there other reasons we still have/need this option?

no, i disabled it by chance in hope to speed up the compilation. okay, maybe we can drop it then.

alimaredia:
@wjwithagen https://github.com/ceph/ceph/pull/10186 removes the option so that no one tries to build WITH_MDS=OFF and finds it doesn't work

wjwithagen:
On 7-7-2016 20:40, Ali Maredia wrote:
> @wjwithagen <https://github.com/wjwithagen> #10186
> <https://github.com/ceph/ceph/pull/10186> removes the option so that no
> one tries to build WITH_MDS=OFF and finds it doesn't work

sure, no problem here as far as I can tell.

PR #10155, opened 2016-07-06
qa/workunits/rbd: before removing image make sure it is not bootstrapped

trociny:
If an image is being bootstrapped, it implies that the rbd-mirror
daemon currently has the image open. The removal API will prevent the
removal of any image that is opened by another client.

Works-around: http://tracker.ceph.com/issues/16555


dillaman:
lgtm

PR #10157, opened 2016-07-06
rgw: fix error_repo segfault in data sync

cbodley:
RGWDataSyncShardCR will only allocate an error_repo if it's doing
incremental sync, so RGWDataSyncSingleEntryCR needs to guard against a
null error_repo

also, RGWDataSyncShardCR::stop_spawned_services() was dropping the last
reference to the error_repo before calling drain_all(), which meant that
RGWDataSyncSingleEntryCR could still be holding a pointer

Fixes: http://tracker.ceph.com/issues/16603

### DIFF
--- a/src/rgw/rgw_data_sync.cc
+++ b/src/rgw/rgw_data_sync.cc
@@ -865,7 +865,7 @@ class RGWDataSyncSingleEntryCR : public RGWCoroutine {
 
   RGWDataSyncShardMarkerTrack *marker_tracker;
 
-  RGWOmapAppend *error_repo;
+  boost::intrusive_ptr<RGWOmapAppend> error_repo;
   bool remove_from_repo;
 
   set<string> keys;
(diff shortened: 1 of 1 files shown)


yehudasa:
@cbodley need to make sure that we aren't implicitly cross holding a reference (check that error_repo is not holding a reference to the caller via the cr infrastructure), otherwise neither error_repo is going to be freed, or RGWDataSyncSingleEntryCR

cbodley:
@yehudasa i see. probably easier for RGWDataSyncShardCR to hold the error_repo reference until it destructs, i.e.:
```diff
   void stop_spawned_services() {
     lease_cr->go_down();
     if (error_repo) {
       error_repo->finish();
-      error_repo->put();
-      error_repo = NULL;
     }
   }
```

yehudasa:
@cbodley lgtm

PR #10160, opened 2016-07-06
rgw: aws4: add presigned url bugfix in runtime

jmunhoz:
Runtime bugfix to handle presigned urls computed with canonical requests using
the port number once.

Boto2 computes canonical requests using the port number twice although it
should be used once only. This behaviour is a bug supported by AWS S3. Boto2 is
used in RGW S3 as reference implementation.

The client-side tools not supporting this boto2 bug will fail although they
should work too.

In order to support both presigned url implementations this patch adds a config
option to compute a second signature. With this option enabled, the code will
compute two signatures when the first signature is not valid. The aws4 auth
succeed if some of the two signatures is valid.

The config option rgw_s3_auth_aws4_presigned_url_bugfix is disabled by default
so one signature, working with boto2, is computed only.

Fixes: http://tracker.ceph.com/issues/16463

Signed-off-by: Javier M. Mellid jmunhoz@igalia.com

### DIFF
--- a/src/common/config_opts.h
+++ b/src/common/config_opts.h
@@ -1331,6 +1331,7 @@ OPTION(rgw_keystone_verify_ssl, OPT_BOOL, true) // should we try to verify keyst
 OPTION(rgw_keystone_implicit_tenants, OPT_BOOL, false)  // create new users in their own tenants of the same name
 OPTION(rgw_s3_auth_use_rados, OPT_BOOL, true)  // should we try to use the internal credentials for s3?
 OPTION(rgw_s3_auth_use_keystone, OPT_BOOL, false)  // should we try to use keystone for s3?
+OPTION(rgw_s3_auth_aws4_force_boto2_compat, OPT_BOOL, true) // force aws4 auth boto2 compatibility
 
 /* OpenLDAP-style LDAP parameter strings */
 /* rgw_ldap_uri  space-separated list of LDAP servers in URI format */
--- a/src/rgw/rgw_rest_s3.cc
+++ b/src/rgw/rgw_rest_s3.cc
@@ -3307,7 +3307,14 @@ int RGW_Auth_S3::authorize(RGWRados *store, struct req_state *s)
       if (algorithm != "AWS4-HMAC-SHA256") {
         return -EPERM;
       }
-      return authorize_v4(store, s);
+      /* compute first aws4 signature (stick to the boto2 implementation) */
+      int err = authorize_v4(store, s);
+      if ((err==-ERR_SIGNATURE_NO_MATCH) && !store->ctx()->_conf->rgw_s3_auth_aws4_force_boto2_compat) {
+        /* compute second aws4 signature (no bugs supported) */
+        ldout(s->cct, 10) << "computing second aws4 signature..." << dendl;
+        return authorize_v4(store, s, false);
+      }
+      return err;
     }
 
     /* AWS2 */
@@ -3461,7 +3468,7 @@ static std::array<string, 3> aws4_presigned_required_keys = { "Credential", "Sig
 /*
  * handle v4 signatures (rados auth only)
  */
-int RGW_Auth_S3::authorize_v4(RGWRados *store, struct req_state *s)
+int RGW_Auth_S3::authorize_v4(RGWRados *store, struct req_state *s, bool force_boto2_compat /* = true */)
 {
   string::size_type pos;
   bool using_qs;
@@ -3736,7 +3743,7 @@ int RGW_Auth_S3::authorize_v4(RGWRados *store, struct req_state *s)
       }
     }
     string token_value = string(t);
-    if (using_qs && (token == "host")) {
+    if (force_boto2_compat && using_qs && (token == "host")) {
       if (!port.empty() && port != "80" && port != "0") {
         token_value = token_value + ":" + port;
       } else if (!secure_port.empty() && secure_port != "443") {
--- a/src/rgw/rgw_rest_s3.h
+++ b/src/rgw/rgw_rest_s3.h
@@ -449,7 +449,7 @@ private:
   static rgw::LDAPHelper* ldh;
 
   static int authorize_v2(RGWRados *store, struct req_state *s);
-  static int authorize_v4(RGWRados *store, struct req_state *s);
+  static int authorize_v4(RGWRados *store, struct req_state *s, bool force_boto2_compat = true);
   static int authorize_v4_complete(RGWRados *store, struct req_state *s,
 				  const string& request_payload,
 				  bool unsigned_payload);


jmunhoz:
@pritha-srivastava working on my side (test suite passing, boto2 and mc manual tests also running) can you confirm it works on your side? thanks!

src/common/config_opts.h
```diff
@@ -1310,6 +1310,7 @@ OPTION(rgw_keystone_verify_ssl, OPT_BOOL, true) // should we try to verify keyst
 OPTION(rgw_keystone_implicit_tenants, OPT_BOOL, false)  // create new users in their own tenants of the same name
 OPTION(rgw_s3_auth_use_rados, OPT_BOOL, true)  // should we try to use the internal credentials for s3?
 OPTION(rgw_s3_auth_use_keystone, OPT_BOOL, false)  // should we try to use keystone for s3?
+OPTION(rgw_s3_auth_aws4_presigned_url_bugfix, OPT_BOOL, false) // aws4 auth presigned url client-side bug support?
```
pritha-srivastava: Please remove '?' at the end of the statement

src/rgw/rgw_common.h
```diff
@@ -1297,6 +1297,7 @@ struct req_state {
   /* aws4 auth support */
   bool aws4_auth_needs_complete;
   bool aws4_auth_streaming_mode;
+  bool aws4_auth_presigned_url_bugfix;
```
pritha-srivastava: 'rgw_s3_auth_aws4_presigned_url_bugfix' can be used to decide whether to compute the signature once or twice, eliminating the need for this variable - 'aws4_auth_presigned_url_bugfix.

src/rgw/rgw_rest_s3.cc
```diff
@@ -3238,7 +3240,14 @@ int RGW_Auth_S3::authorize(RGWRados *store, struct req_state *s)
       if (algorithm != "AWS4-HMAC-SHA256") {
         return -EPERM;
       }
-      return authorize_v4(store, s);
+      /* compute first aws4 signature (stick to the boto2 implementation) */
+      int err = authorize_v4(store, s);
+      if (err && store->ctx()->_conf->rgw_s3_auth_aws4_presigned_url_bugfix) {
```
pritha-srivastava: How about checking for ERR_SIGNATURE_NO_MATCH error code explicitly here?

src/rgw/rgw_rest_s3.cc
```diff
@@ -3229,6 +3229,8 @@ int RGW_Auth_S3::authorize(RGWRados *store, struct req_state *s)
     return 0;
   }
 
+  s->aws4_auth_presigned_url_bugfix = false;
+
   if (!s->http_auth || !(*s->http_auth)) {
```
pritha-srivastava: The ' else' part also has a call to authorize_v4, I think the same logic of computing the signature twice, needs to be added there also.

src/rgw/rgw_rest_s3.cc
```diff
@@ -3667,7 +3676,7 @@ int RGW_Auth_S3::authorize_v4(RGWRados *store, struct req_state *s)
       }
     }
     string token_value = string(t);
-    if (using_qs && (token == "host")) {
+    if (!s->aws4_auth_presigned_url_bugfix && using_qs && (token == "host")) {
```
pritha-srivastava: Same comment as above - Instead of using s->aws_auth_presigned_url_bugfix, store->ctx()->_conf->rgw_s3_auth_aws4_presigned_url_bugfix can be used here.

src/common/config_opts.h
jmunhoz: Keeping the question mark at the end of the comment looks coherent with the comment style of the previous options (rgw_s3_auth_use_rados and rgw_s3_auth_use_keystone). Maybe it makes sense keeping it. What do you think?

src/rgw/rgw_common.h
jmunhoz: No really, both booleans have a similar name cause they work on the presigned url bug but they need to hold two different boolean states working in tandem. They will only contain the same state/value in the authorize_v4's second run when the admin turned the 'rgw_s3_auth_aws4_presigned_url_bugfix' flag on and the presigned url case happens.

Let me share in detail. The first boolean 'rgw_s3_auth_aws4_presigned_url_bugfix' is the configurable option. It helps to decide if we have to compute the second signature in the presigned url case only.

The second boolean 'aws4_auth_presigned_url_bugfix' is needed to drive the logic in authorize_v4 and, at the same time, keeping the function body with minimal modifications. You will see the only modification with this approach in authorize_v4 is checking this var when the using_qs boolean is 'true' (parameters shipping in the query string as in the case of presigned urls).
 
The key part is the 'aws4_auth_presigned_url_bugfix' needs to be initialized to 'false' and it will have that value but if it is a presigned url case *and* the configurable 'rgw_aws4_auth_presigned_url_bugfix' is also 'true'. In other words, authorize_v4 always uses 'aws4_auth_presigned_url_bugfix' as 'false' in the first run (first signature computation) and then, if it is a presigned url and the admin wants runs the runtime bugfix, it will run a second signature computation with the 'aws4_auth_presigned_url_bugfix' boolean to true.

You can consider 'rgw_s3_auth_aws4_presigned_url_bugfix' as the policy and 'aws4_auth_presigned_url_bugfix' as the mechanism.

src/rgw/rgw_rest_s3.cc
jmunhoz: ERR_SIGNATURE_NO_MATCH is used to return failure when signatures don't match. Bear in mind authorize_v4 calls more functions and more errors can be returned. For example, parsing errors, missed fields, etc. The approach here is checking the signature is ok (err = 0) vs we hit some error (err != 0). Returning the error var, we are propagating the result of the first authorize_v4 call. If err = 0 all things were okay with the first signature and the second signature computation didn't run.

src/rgw/rgw_rest_s3.cc
jmunhoz: Having a look at:

https://github.com/jmunhoz/ceph/blob/8f915e070a5f7847e785f556e89bda09540d5321/src/rgw/rgw_rest_s3.cc#L3679

you'll see it is guarded by 'using_qs'

<pre>
if (!s->aws4_auth_presigned_url_bugfix && using_qs && (token == "host")) {
</pre>

In the case of presigned urls using_qs is true:

https://github.com/jmunhoz/ceph/blob/8f915e070a5f7847e785f556e89bda09540d5321/src/rgw/rgw_rest_s3.cc#L3425

and it is guarded by:

<pre>
if ((!s->http_auth) || !(*s->http_auth)) {
</pre>

This is the same condition guarding the first conditional branch so I think we don't need add the second authorize_v4 call cause it will not be executed due to using_qs.

src/common/config_opts.h
pritha-srivastava: The previous options are questions, hence the question marks. The one related to rgw_s3_auth_aws4_presigned_url_bugfix is not, hence I think the question mark should be removed.

src/rgw/rgw_rest_s3.cc
pritha-srivastava: As far as I understand that computing the signatures twice is needed only when signatures do not match, and that is because the port is appended once or twice. In case of other errors, will computing the signature more than once help?

src/common/config_opts.h
jmunhoz: ok

src/rgw/rgw_rest_s3.cc
jmunhoz: yes, the current patch is a bit conservative. I tested with the new change and it looks working ok. can you confirm? thanks

pritha-srivastava:
@yehudasa : I have run the s3 tests and they are working fine. Can you please review the PR.

jmunhoz:
@mattbenjamin do you have some time to review this PR along this week? It would be great if we get it in to support Minio clients asap. thanks!

yehudasa:
@mattbenjamin ping

src/rgw/rgw_common.h
yehudasa: @jmunhoz I'd comment what bugfix this configurable actually fixes, or maybe rename it to .._boto2_compat or something along these lines

yehudasa:
@pritha-srivastava can you re-review this one following @jmunhoz comments?

src/rgw/rgw_common.h
cbodley: @jmunhoz it looks like this variable doesn't need to be in req_state itself. consider adding a `bool boto2_compat = false` parameter to `RGW_Auth_S3::authorize_v4()`, then have `RGW_Auth_S3::authorize()` pass in `true` the second time around

jmunhoz: @yehudasa @cbodley ok

jmunhoz:
@yehudasa renamed to '_force_boto2_compat'

@cbodley added the suggested  change. it makes the code more simple and compact

@pritha-srivastava the new code is working on my side. feel free to re-review. txs!

pritha-srivastava:
:+1:

jmunhoz:
@yehudasa ping

dachary:

PR #10161, opened 2016-07-06
CMake - stop pip checking for updates

dang:

### DIFF
--- a/src/ceph-detect-init/CMakeLists.txt
+++ b/src/ceph-detect-init/CMakeLists.txt
@@ -3,7 +3,7 @@ set(CEPH_DETECT_INIT_VIRTUALENV ${CEPH_BUILD_VIRTUALENV}/ceph-detect-init-virtua
 add_custom_target(ceph-detect-init ALL
   COMMAND
   ${CMAKE_SOURCE_DIR}/src/tools/setup-virtualenv.sh ${CEPH_DETECT_INIT_VIRTUALENV} &&
-  ${CEPH_DETECT_INIT_VIRTUALENV}/bin/pip install --no-index --use-wheel --find-links=file:${CMAKE_SOURCE_DIR}/src/ceph-detect-init/wheelhouse -e .
+  ${CEPH_DETECT_INIT_VIRTUALENV}/bin/pip install --disable-pip-version-check --no-index --use-wheel --find-links=file:${CMAKE_SOURCE_DIR}/src/ceph-detect-init/wheelhouse -e .
   WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/src/ceph-detect-init
   COMMENT "ceph-detect-init is being created")
 
--- a/src/tools/setup-virtualenv.sh
+++ b/src/tools/setup-virtualenv.sh
@@ -22,12 +22,12 @@ virtualenv --python python2.7 $DIR
 . $DIR/bin/activate
 # older versions of pip will not install wrap_console scripts
 # when using wheel packages
-pip --log $DIR/log.txt install --upgrade 'pip >= 6.1'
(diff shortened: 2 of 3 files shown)


alimaredia:
@tchaikov @dachary any objections?

alimaredia:
@tchaikov @dachary any input?

dachary:
<pre>
</pre>

src/ceph-detect-init/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ set(CEPH_DETECT_INIT_VIRTUALENV ${CEPH_BUILD_VIRTUALENV}/ceph-detect-init-virtua
 add_custom_target(ceph-detect-init ALL
   COMMAND
   ${CMAKE_SOURCE_DIR}/src/tools/setup-virtualenv.sh ${CEPH_DETECT_INIT_VIRTUALENV} &&
-  ${CEPH_DETECT_INIT_VIRTUALENV}/bin/pip install --no-index --use-wheel --find-links=file:${CMAKE_SOURCE_DIR}/src/ceph-detect-init/wheelhouse -e .
+  ${CEPH_DETECT_INIT_VIRTUALENV}/bin/pip install --disable-pip-version-check --no-index --use-wheel --find-links=file:${CMAKE_SOURCE_DIR}/src/ceph-detect-init/wheelhouse -e .
```
tchaikov: it's implied by "--no-index", see https://pip.pypa.io/en/stable/news/ (search for "6.0.3 (2014-12-23)")? so maybe you are facing another case where pip wants to access the network?

alimaredia:
@dang were you able to double check this PR, after I showed you that I still got some warnings yesterday?

dang:
Sorry, forgot.  The problem was tools/setup-virtualenv.sh, which is called by cmake.  I've updated that as well.

alimaredia:
@dang comparing the output from building ceph-detect-init from this PR and master, the only difference is this (http://paste.fedoraproject.org/389111/14680101/):

> You are using pip version 6.0.8, however version 8.1.2 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.

That is what's intended right?

dang:
Correct.

alimaredia:
@tchaikov this looks good to me. Can we merge it?

src/tools/setup-virtualenv.sh
```diff
@@ -22,12 +22,12 @@ virtualenv --python python2.7 $DIR
 . $DIR/bin/activate
 # older versions of pip will not install wrap_console scripts
 # when using wheel packages
-pip --log $DIR/log.txt install --upgrade 'pip >= 6.1'
+pip --disable-pip-version-check --log $DIR/log.txt install --upgrade 'pip >= 6.1'
```
tchaikov: @dang what if pip does not support this option at that moment? see https://github.com/ceph/ceph/pull/10283#issuecomment-232434669

dang: Which pip version doesn't have it?  I didn't realize that there were versions without it, since all versions I have access to, even pretty old ones, have it.

tchaikov: pip 1.5.4. it's shipped by ubuntu trusty. and see https://github.com/pypa/pip/commit/c3df28c7745af243f0c2e6f89af5bf2a93cbca60. the commit introducing this option was added after pip 6.0.

dang: Okay, that's not good.  I guess this should be reverted.

tchaikov: no worries, i have a PR addressing it: https://github.com/ceph/ceph/pull/10283/commits/6f284846457254873fddfb0a4092dc85c3d8e3c9

PR #10162, opened 2016-07-06
rgw: better error message when user has no bucket created yet

gaurav36:
Currently when there is no bucket in the cluster and user execute
 "$radosgw-admin user  stats --uid=testid" command then rgw admin
not throwing meaningful error message.

With this fix it will show proper meaningful error message.

Fixes: http://tracker.ceph.com/issues/16444

Reported-by: Abhishek Lekshmanan <abhishek@suse.com>
Signed-off-by: Gaurav Kumar Garg <garg.gaurav52@gmail.com>

### DIFF
--- a/src/rgw/rgw_admin.cc
+++ b/src/rgw/rgw_admin.cc
@@ -4817,7 +4817,11 @@ next:
     string user_str = user_id.to_str();
     int ret = store->cls_user_get_header(user_str, &header);
     if (ret < 0) {
-      cerr << "ERROR: can't read user header: " << cpp_strerror(-ret) << std::endl;
+      if (ret == -ENOENT) { /* in case of ENOENT */
+        cerr << "User has not been initialized or user does not exist" << std::endl;
+      } else {
+        cerr << "ERROR: can't read user: " << cpp_strerror(ret) << std::endl;
+      }
       return -ret;
     }
 


src/rgw/rgw_admin.cc
```diff
@@ -4817,7 +4817,11 @@ int main(int argc, char **argv)
     string user_str = user_id.to_str();
     int ret = store->cls_user_get_header(user_str, &header);
     if (ret < 0) {
-      cerr << "ERROR: can't read user header: " << cpp_strerror(-ret) << std::endl;
+      if (ret == -2) { /* in case of ENOENT */
```
yehudasa: `if (ret == -ENOENT)`

gaurav36: sure, will do in next patch thanks :)

src/rgw/rgw_admin.cc
```diff
@@ -4817,7 +4817,11 @@ int main(int argc, char **argv)
...
+        cerr << "Error: There is no bucket/object in the cluster. Please create buckets/object first." << std::endl;
```
yehudasa: It's not an error, so no need to use 'Error:'. Also, is this always correct? What if another user have data, aren't we going to get the same error. And there's no need for the second part of the sentence (no need to instruct users what to do).

yehudasa: Sounds to me more like 'User does not exist error, or user has not been initialized'

theanalyst: Yeah that sounds more appropriate

gaurav36: @yehudasa  done in next PR.

src/rgw/rgw_admin.cc
```diff
@@ -4817,7 +4817,11 @@ int main(int argc, char **argv)
     string user_str = user_id.to_str();
     int ret = store->cls_user_get_header(user_str, &header);
     if (ret < 0) {
-      cerr << "ERROR: can't read user header: " << cpp_strerror(-ret) << std::endl;
+      if (ret == -ENOENT) { /* in case of ENOENT */
+        cerr << "User has not been initialized " << std::endl;
```
theanalyst: `User has not been initialized or doesn't exist` would be more appropriate here as the same error is thrown when the user doesn't exist as well

gaurav36: done

src/rgw/rgw_admin.cc
```diff
@@ -4817,7 +4817,11 @@ int main(int argc, char **argv)
     string user_str = user_id.to_str();
     int ret = store->cls_user_get_header(user_str, &header);
     if (ret < 0) {
-      cerr << "ERROR: can't read user header: " << cpp_strerror(-ret) << std::endl;
+      if (ret == -ENOENT) { /* in case of ENOENT */
+        cerr << "User has not been initialized or user does not exist" << std::endl;
+      } else {
+        cerr << "ERROR: can't read user: " << cpp_strerror(-ret) << std::endl;
```
mattbenjamin: for posterity, cpp_strerror coerces whatever code is passed to be positive, so there's no need to make a code negative

gaurav36: @mattbenjamin  done.

yehudasa:
@oritwas ping

PR #10165, opened 2016-07-06
[DNM] switch to _XOPEN_SOURCE=700 vs. GNU_SOURCE

dx9:
Did some quick testing across trusty and alpine and the gcc warnings are more consistent. I think that is pretty cool.

However this maybe a bad idea.

_GNU_SOURCE deps that would need to be handled locally are:
- f_setpipe_sz()
- sched_setaffinity()
- sched_yield()
- pipe2

bzero() usage in erasure code would have to dealt with, POSIX deprecated it and then dropped it.

Putting the feelers out to see if there is any love for this. There would be some more changes required to fix some warnings up.

?


liewegas:
@wjwithagen @tchaikov ?

wjwithagen:
@liewegas 
I moved the `-DGNU_SOURCE` to GCC-compiles, because there are no matches for this define in the FreeBSD parts. I'd expect that setting `_XOPEN_SOURCE=700` will work, since there are some developers in FreeBSD that take standards rather serious. All the parts of strerror_r are there.
One thing i notice, is that we ignore the returned result, but that should not be a problem.

bzero not being there anymore would be a good reason to rewrite and use memset, I think.
It is in FreeBSD, but for compatibility for older programs.

I'll take the patch and run it to see if any errors pop up

wjwithagen:
@dx9 @liewegas 
Right, just putting the XOPEN_SOURCE=700 doesn't work for FreeBSD, because it then starts to formalise on what to include and what not.
Things like usleep start to get hidden.
And I'd need to start adding other flags.

Compiling with strerror_r, and without the XOPEN define works best for FreeBSD

dx9:
@wjwithagen Thanks for trying it! I was confused last night why strerror_r wasn't working consistently with XOPEN_SOURCE. Seems we also need -U_GNU_SOURCE with gcc to get _XOPEN_SOURCE to behave. usleep() got deprecated too so it would need to go as well along with a bunch of other things I've never seen before.

Updated this setup to break the build consistently with similar errors on musl libc, glibc, and BSD libc!! Oh yeah. :grinning: I didn't realize that XOPEN_SOURCE[1] is standardized. Just thought it was a way to instruct glibc to conform better.

I'm curious now if an XSI compliantish build is possible/practical and what it would look like. It could be useful if it blocked some platform issues before they hit the CI level. Saving people the effort of working through things compilers can tell us about.

This definitely spirals more than the original warnings but looks promising. Gonna keep pecking away at it and see where the rabbit hole leads.

[1] http://pubs.opengroup.org/onlinepubs/9699919799/functions/V2_chap02.html

liewegas:
closing this for now; feel free to rebase and reopen!

PR #10166, opened 2016-07-06
test_rados_tool.sh: Make script work under ctest

wjwithagen:


qa/workunits/rados/test_rados_tool.sh
```diff
@@ -52,18 +52,30 @@ run() {
 DNAME="`dirname $0`"
 DNAME="`readlink -f $DNAME`"
 RADOS_TOOL="`readlink -f \"$DNAME/../rados\"`"
-if ! test -f $RADOS_TOOL ; then
-    RADOS_TOOL=$(which rados)
+if [ ! -f "$RADOS_TOOL" ] ; then
```
tchaikov: @wjwithagen maybe we can use `$CEPH_BIN` to tell where rados cli is located if `test -n $CEPH_BIN`? and fallback to `which rados` if `test -n $CEPH_BIN` fails?

wjwithagen: @tchaikov 
We can/could do that.
The readlink stuff makes sense to me, but getting it out will make the code much more legible. But it emphasizes the need to be executed from a parent script setting these vars. I'll try and patch as suggested 

I sort of wonder why we do  `$(which rados)` anyways. it will run rados from an installed ceph installation, and I would not really like to start using that whilest testing.

tchaikov: the reason why we do `$(which rados)` is for the qa run, as the path name ("qa/workunits/rados/test_rados_tool.sh") suggests, `test_rados_tool.sh` is a workunit used by ceph-qa-suite to build the workload for the cluster under testing. in that case, "rados" cli is installed by the package manager (rpm or dpkg), so, `which rados` is a reliable way to locate "rados" in that case.

qa/workunits/rados/test_rados_tool.sh
```diff
@@ -316,9 +316,11 @@ test_xattr() {
     $RADOS_TOOL -p $POOL listxattr $OBJ > $V1
     grep -q foo $V1
     grep -q bar $V1
-    wc -l $V1 | grep -q "^2 "
+    cat $V1
+    [ `cat $V1 | wc -l ` -eq 2 ]
     rm $V1 $V2
     cleanup
+    echo test_xattr done
```
tchaikov: might want to remove this line? maybe no news is good news?

qa/workunits/rados/test_rados_tool.sh
wjwithagen: @tchaikov 
Makes sense. Never too old to learn. Thanx for the info.

qa/workunits/rados/test_rados_tool.sh
wjwithagen: @tchaikov 
During debugging it is always hard to tell when tests have ended...
I would love every routine to have it indicate start and stop, makes tracking way much easier.
But if you insist, I will remove it.

qa/workunits/rados/test_rados_tool.sh
wjwithagen: @tchaikov 
Does this mean that CEPH_TMP, or something like that is also set?
Would be nice to dump all the TEMP stuff in, instead of /tmp.

wjwithagen:
@tchaikov 
This more to your liking?

qa/workunits/rados/test_rados_tool.sh
```diff
@@ -52,18 +52,24 @@ run() {
 DNAME="`dirname $0`"
```
tchaikov: could you remove these crufts? as they are not needed anymore.

qa/workunits/rados/test_rados_tool.sh
```diff
@@ -52,18 +52,24 @@ run() {
...
-fi
-CEPH_TOOL="`readlink -f \"$DNAME/../ceph\"`"
-if ! test -f $CEPH_TOOL ; then
-    CEPH_TOOL=$(which ceph)
+if [ -n "$CEPH_BIN" ] ; then
+   # CMake env
+   RADOS_TOOL="$CEPH_BIN/rados"
+   CEPH_TOOL="$CEPH_BIN/ceph"
+else
+   # executables should be nstalled by the QA env
```
tchaikov: typo.

qa/workunits/rados/test_rados_tool.sh
```diff
@@ -52,18 +52,24 @@ run() {
...
+   # executables should be nstalled by the QA env 
+   RADOS_TOOL=$(which rados)
+   CEPH_TOOL=$(which ceph)
 fi
+
 KEEP_TEMP_FILES=0
 POOL=trs_pool
 POOL_CP_TARGET=trs_pool.2
 
+[ ! -z "$RADOS_TOOL" ] || die "\$RADOS_TOOL was not defined"
```
tchaikov: `-x` would suffce, i guess?

qa/workunits/rados/test_rados_tool.sh
tchaikov: could you please remove it? i'd like to keep this commit as a refactor only change.

qa/workunits/rados/test_rados_tool.sh
wjwithagen: Well the first one -z warns if the logic has not worked.
Especially during porting I've learned to appriciate this kind of defensive programming.
So lets keep these, they will help others in future.

wjwithagen:
@tchaikov ping!

qa/workunits/rados/test_rados_tool.sh
tchaikov: why the logic might not work?

tchaikov:
will revisit this PR once all comments addressed.

qa/workunits/rados/test_rados_tool.sh
tchaikov: we don't use `CEPH_TMP` AFAIK. i am trying to move intermediate stuff into `$TMPDIR` instead.

wjwithagen:
@tchaikov 
The only one outstanding is the -z versus the -x.
-x works but is not as informative.

But hey, you are going to do the commit in the tree. 8D

wjwithagen:
@tchaikov 
Sorry, I got the commit sort of wrong, and nothing of my last changes made it in the pull.
So redoing that.

wjwithagen:
@tchaikov This should do it?
Sorry for the mixup.

qa/workunits/rados/test_rados_tool.sh
wjwithagen: On 7-7-2016 12:29, Kefu Chai wrote:
> In qa/workunits/rados/test_rados_tool.sh
> <https://github.com/ceph/ceph/pull/10166#discussion_r69884463>:
> 
>> @@ -52,18 +52,30 @@ run() {
>>  DNAME="`dirname $0`"
>>  DNAME="`readlink -f $DNAME`"
>>  RADOS_TOOL="`readlink -f \"$DNAME/../rados\"`"
>> -if ! test -f $RADOS_TOOL ; then
>> -    RADOS_TOOL=$(which rados)
>> +if [ ! -f "$RADOS_TOOL" ] ; then
> 
> we don't use |CEPH_TMP| AFAIK. i am trying to move intermediate stuff
> into |$TMPDIR| instead.

But then why don't we/you do that in the |detect-build-env-vars.sh|

And then start to try and propagate it into scripts as we go along.
Because my /tmp looks like a battlezone after running all tests.
Lots of the unittest could be more careful as well.

qa/workunits/rados/test_rados_tool.sh
```diff
@@ -304,7 +305,7 @@ test_xattr() {
     $RADOS_TOOL -p $POOL listxattr $OBJ > $V1
     grep -q foo $V1
     grep -q bar $V1
-    wc -l $V1 | grep -q "^2 "
+    [ `cat $V1 | wc -l ` -eq 2 ]
```
tchaikov: nit, could remove the space before the backtick.

qa/workunits/rados/test_rados_tool.sh
tchaikov: i don't think we need to use `detect-build-env-vars.sh` for propagate `$TMPDIR`, and `detect-build-env-vars.sh` is not used to "propagate" env variables, it is used to *detect* the settings not included in the env vars, which are passed down by shell scripts and cmake.

tchaikov:
lgtm.

wjwithagen:
On 15-7-2016 11:04, Kefu Chai wrote:
> Merged #10166 <https://github.com/ceph/ceph/pull/10166>.

Thanx, --WjW

PR #9489, opened 2016-06-03
ReplicatedPG: call op_applied for submit_log_entries based repops

athanatos:
Signed-off-by: Samuel Just <sjust@redhat.com>

PR #10170, opened 2016-07-07
include write error codes in the pg log

jdurgin:
This depends on https://github.com/ceph/ceph/pull/9489 - if you'd like to see more granular history check out the wip-pg-log-errors-10 branch.

There is at least one issue left to fix - proper cleanup of the repop on shutdown, since the LogUpdateCtx has a ref to it that isn't accounted for yet. Valgrind catches this.


src/osd/PG.cc
```diff
@@ -4547,39 +4547,47 @@ void PG::share_pg_info()
...
-  PGLogEntryHandler rollbacker;
-  pg_log.append_new_log_entries(
-    info.last_backfill,
-    info.last_backfill_bitwise,
-    entries,
-    &rollbacker);
-  rollbacker.apply(this, &t);
-  info.last_update = pg_log.get_head();
+  if (handle_missing) {
+    PGLogEntryHandler rollbacker;
```
athanatos: It seems like it would be easier to just update append_new_log_entries to do the right thing.  Also, append_new_log_entries could indicate whether stats should be invalidated based on the log entries.  That would avoid needing to update the message.

src/osd/PG.cc
```diff
@@ -4547,39 +4547,47 @@ void PG::share_pg_info()
...
 void PG::merge_new_log_entries(
   const list<pg_log_entry_t> &entries,
+  bool handle_missing,
   ObjectStore::Transaction &t)
 {
   dout(10) << __func__ << " " << entries << dendl;
   assert(is_primary());
 
-  append_log_entries_update_missing(entries, t);
+  append_log_entries_update_missing(entries, handle_missing, t);
```
athanatos: It seems like append_log_entries could indicate whether the log requires the missing set to be rebuilt based on the log entries?

src/osd/PGLog.h
```diff
@@ -166,7 +166,8 @@ struct PGLog : DoutPrefixProvider {
     bool get_request(
       const osd_reqid_t &r,
       eversion_t *replay_version,
-      version_t *user_version) const {
+      version_t *user_version,
+      int *return_code) const {
       assert(replay_version);
       assert(user_version);
```
athanatos: assert(return_code); -- someone's going to pass in null at some point.

jdurgin:
this is ready to go now, fixed the mem leak on shutdown and the latest rados run has the same failures as master

PR #10174, opened 2016-07-07
rgw: test_multi.py avoid creating mds

theanalyst:
Avoid creating mds' while bootstrapping new clusters from test_multi.py
as we only require an rgw+ceph cluster for the tests


cbodley:
looks good. i've been using the environment variables to control this, i.e. `MON=1 OSD=1 MDS=0 nosetests ...`

PR #10176, opened 2016-07-07
mon/mds: add err info when load_metadata is abnormal

renhwztetecs:
add a print of the most important reason is that:
when the MDSMonitor::init  call MDSMonitor::load_metadata, no processing error code,
so printing can discover problems


gregsfortytwo:
LGTM

PR #10181, opened 2016-07-07
configure: Add -D_LARGEFILE64_SOURCE to Linux build.

gaurav36:
Fixes: http://tracker.ceph.com/issues/16612

Without -D_LARGEFILE64_SOURCE we can't build against libxfs,
because off64_t must be defined.

(cherry picked from commit 602425a)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -51,6 +51,7 @@ darwin*)
 	;;
 linux*)
 	linux="yes"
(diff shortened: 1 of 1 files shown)


gaurav36:
I think i need to backport it to hammer instead of hammer-bakcport. closing this PR. will send fresh PR.

PR #10182, opened 2016-07-07
hammer: configure: Add -D_LARGEFILE64_SOURCE to Linux build.

gaurav36:
Fixes: http://tracker.ceph.com/issues/16612

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -51,6 +51,7 @@ darwin*)
 	;;
 linux*)
 	linux="yes"
+	CFLAGS="-D_LARGEFILE64_SOURCE ${CFLAGS}"
 	;;
 freebsd*)
 	freebsd="yes"


smithfarm:

That said, I don't know if rados test exercises this fix at all?

smithfarm:
@liewegas @athanatos This PR passed a /200 rados run on Ubuntu. None of the failures were reproducible. For details see http://tracker.ceph.com/issues/15895#note-18

athanatos:
I don't really understand the implications of this.  @gaurav36 Why specifically does this need to be backported to hammer?

gaurav36:
@athanatos  Hammer release is having issue while building because of missing (flag) backport

athanatos:
@gaurav36 Why did it work before?  Is this only very new versions of fedora?

gaurav36:
@athanatos   Its fedora23. Without -D_LARGEFILE64_SOURCE we can't build against libxfs,
because off64_t must be defined. Its a backport patch  http://tracker.ceph.com/issues/16611

athanatos:
@gaurav36 Right, was there something introduced recently that required off64_t, or has this been broken on fedora 23 all along?

cbodley:
@athanatos I think it's been broken all along. I had to work around this issue a couple months ago working on a backport.

athanatos:
@cbodley So it makes sense to backport it?

cbodley:
yes, please

athanatos:
@smithfarm go for it

PR #10184, opened 2016-07-07
doc: rm SysV instructions, add systemd

ktdreyer:

### DIFF
--- a/doc/rados/operations/add-or-rm-osds.rst
+++ b/doc/rados/operations/add-or-rm-osds.rst
@@ -174,13 +174,13 @@ your new OSD before it can begin receiving data. You may use
 ``service ceph`` from your admin host or start the OSD from its host
 machine.
 
-For Debian/Ubuntu use Upstart. ::
+For Ubuntu Trusty use Upstart. ::
 
 	sudo start ceph-osd id={osd-num}
 
-For CentOS/RHEL, use sysvinit. ::
+For all other distros use systemd. ::
 
-	sudo /etc/init.d/ceph start osd.{osd-num}
+	sudo systemctl start ceph-osd@{osd-num}
 
 
 Once you start your OSD, it is ``up`` and ``in``.
@@ -274,7 +274,7 @@ That is, the OSD may be ``up`` and ``out``. You must stop
 your OSD before you remove it from the configuration. :: 
 
 	ssh {osd-host}
-	sudo /etc/init.d/ceph stop osd.{osd-num}
+	sudo systemctl stop ceph-osd@{osd-num}
 
 Once you stop your OSD, it is ``down``. 
 
--- a/doc/rados/operations/operating.rst
+++ b/doc/rados/operations/operating.rst
@@ -2,15 +2,108 @@
  Operating a Cluster
 =====================
 
-.. index:: Upstart; operating a cluster
+.. index:: systemd; operating a cluster
+
+
+Running Ceph with systemd
+==========================
+
+For all distributions that support systemd (CentOS 7, Fedora, Debian Jessie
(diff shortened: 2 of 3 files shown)


doc/rados/operations/add-or-rm-osds.rst
```diff
@@ -174,13 +174,13 @@ your new OSD before it can begin receiving data. You may use
 ``service ceph`` from your admin host or start the OSD from its host
 machine.
 
-For Debian/Ubuntu use Upstart. ::
+For Ubuntu Trusty use Upstart. ::
 
 	sudo start ceph-osd id={osd-num}
 
-For CentOS/RHEL, use sysvinit. ::
```
vasukulkarni: I think we should keep old as well as new unless user is using the writing 'branch' doc or it would be confusing with various versions in use.

doc/rados/operations/operating.rst
```diff
@@ -2,15 +2,108 @@
...
+
 
 Running Ceph with Upstart
 =========================
 
-When deploying Ceph Cuttlefish and beyond with ``ceph-deploy`` on Ubuntu, you 
-may start and stop Ceph daemons on a :term:`Ceph Node` using the event-based 
-`Upstart`_.  Upstart does not require you to define daemon instances in the 
-Ceph configuration file.
+When deploying Ceph with ``ceph-deploy`` on Ubuntu Trusty, you may start and
```
vasukulkarni: Do you think its needed? may be deleted altogether and just use the right way?

vasukulkarni:
:+1:  much needed update for jewel/xenial.

doc/rados/operations/add-or-rm-osds.rst
ktdreyer: If a user needs to refer back to the Hammer docs, it is still available online: http://docs.ceph.com/docs/hammer/rados/operations/operating/

doc/rados/operations/operating.rst
ktdreyer: I'm not sure I understand what you're saying? I think we should keep the Upstart stuff in here for Trusty, until Ceph master stops supporting Trusty

doc/rados/operations/add-or-rm-osds.rst
vasukulkarni: Yeah that is what I was saying they would have to navigate through each branch, but generally the latest docs should handle atleast the last 2 releases? A new user or google redirected search would be misleading if he not aware of the current releases.

doc/rados/operations/operating.rst
vasukulkarni: This isn't the standard way to start/stop in Upstart right? It should be just using the 'service' command? I think here the old docs is talking about the special use cas using ":term"

doc/rados/operations/operating.rst
```diff
@@ -2,15 +2,108 @@
...
+
+Stopping all Daemons by Type
+----------------------------
+
+To stop all daemons of a particular type on a Ceph Node, execute one of the
+following::
+
+        sudo systemctl stop ceph-mon\*.service ceph-mon.target
+        sudo systemctl stop ceph-osd\*.service ceph-osd.target
+        sudo systemctl stop ceph-mds\*.service ceph-mds.target
```
b-ranto: These should work fine even without the \*.service bits in the current master. This is definitely safer, though.

b-ranto:
LGTM.

doc/rados/operations/operating.rst
vasukulkarni: thanks for clarifying, that was just the rst markup which confused me in the view.

PR #10185, opened 2016-07-07
doc: rm reference to Inktank

ktdreyer:
Let's avoid refering to a specific commercial vendor in the upstream documentation.


ktdreyer:
Fixed in https://github.com/ceph/ceph/pull/11182

PR #10186, opened 2016-07-07
cmake: remove WITH_MDS option

alimaredia:

This patch can merge as long as there's agreement that the WITH_MDS option can go away.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -198,9 +198,6 @@ endif(WITH_SPDK)
 
 find_package(blkid REQUIRED)
 
-# probably fuse etc need this -- how to check? XXX
-option(WITH_MDS "MDS is here" ON)
-
 # needs mds and? XXX
 option(WITH_LIBCEPHFS "libcephfs client library" ON)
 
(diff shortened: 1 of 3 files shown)


liewegas:
@ktdreyer ?

ktdreyer:
:+1: for simplicity

liewegas:
:+1: from me too!

ivancich:
Might a better approach had been to use the WITH_CEPHFS to control building and testing of the mds library and the ceph-mds program? With this merge it's now possible to build without the ceph filesystem and the mds will nonetheless be built.

alimaredia:
@liewegas @tchaikov @jcsp @gregsfortytwo  have any thoughts on what @ivancich mentioned?

gregsfortytwo:
I'm not really aware of either of those build options, so nothing detailed.

In general, building the MDS without the filesystem is weird. But you might want to be able to build the clients without building the MDS, if for instance porting/qualifying clients on weird platforms but not going through the effort of the full server. In which case maybe rename WITH_CEPHFS to WITH_CEPHFS and use WITH_CEPHFS for the full stack?
Or not. I don't have strong feelings about it.

PR #10187, opened 2016-07-07
rgw: fix multi-delete query param parsing.

robbat2:
Multi-delete is triggered by a query parameter on POST, but there are
multiple valid ways of representing it, and Ceph should accept ANY way
that has the query parameter set, regardless of what value or absence of
value.

This caused the RubyGem aws-sdk-v1 to break, and has been present since
multi-delete was first added in commit 0a1f4a97da, for the bobtail
release.

Fixes: http://tracker.ceph.com/issues/16618


yehudasa:
@robbat2 lgtm, will need to go through some qa testing though

PR #10192, opened 2016-07-07
mon: use clearer code structure

batrick:
The failed map is not changed in the first if, so the second if should be an
else.


songbaisen:
lgtm!

PR #10193, opened 2016-07-08
rgw: fix issue 16435

yehudasa:
Fix multiple issues related to handling of the old explicit objs manifest.

cc: @robbat2

### DIFF
--- a/src/rgw/rgw_rados.cc
+++ b/src/rgw/rgw_rados.cc
@@ -1911,7 +1911,7 @@ int RGWObjManifest::generator::create_begin(CephContext *cct, RGWObjManifest *_m
 
   bucket = _b;
   manifest->set_tail_bucket(_b);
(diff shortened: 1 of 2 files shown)


oritwas:
lgtm, but needs testing

yehudasa:
@robbat2 @oritwas I tested the specific functionality manually: objects created in dumpling, both regular and multipart, of various sizes (< 1k, 512k, 1M, 4M, 40M). Made sure that after copying this object and removal of source the target is readable and contains correct data. Also made sure that manifest looks correct. I stumbled on a second issue on the hammer branch I was running that copy of the multipart dumpling objects did not succeed. I assume this doesn't happen in @robbat2's environment, so it could be a regression in hammer. The last commit handles that case. Left to do: run this in teuthology, see that it doesn't break current stuff.

PR #10196, opened 2016-07-08
log/LogClient: fill seq & who for syslog and graylog

xiaoxichen:
Previous code fill "seq" and "who" in LogClient::queue(), but
syslog and graylog doesn't go through this path. Which will result
the message without unidentifier.

Fixes http://tracker.ceph.com/issues/16609

Signed-off-by: Xiaoxi Chen xiaoxchen@ebay.com

### DIFF
--- a/src/common/LogClient.cc
+++ b/src/common/LogClient.cc
@@ -238,11 +238,10 @@ void LogChannel::do_log(clog_type prio, const std::string& s)
   int lvl = (prio == CLOG_ERROR ? -1 : 0);
   ldout(cct,lvl) << "log " << prio << " : " << s << dendl;
   LogEntry e;
-  // who will be set when we queue the entry on LogClient
-  //e.who = messenger->get_myinst();
   e.stamp = ceph_clock_now(cct);
-  // seq will be set when we queue the entry on LogClient
-  // e.seq = ++last_log;
+  // seq and who should be set for syslog/graylog/log_to_mon
+  e.who = parent->get_myinst();
+  e.seq = parent->get_next_seq();
   e.prio = prio;
   e.msg = s;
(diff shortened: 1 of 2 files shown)


tchaikov:
lgtm, but we will acquire the lock twice now for each log entry. @jecluis could you please take a look also?

tchaikov:
@jecluis ping?

src/common/LogClient.cc
```diff
@@ -353,6 +350,17 @@ version_t LogClient::queue(LogEntry &entry)
   return entry.seq;
 }
 
+uint64_t LogClient::get_next_seq()
+{
+  Mutex::Locker l(log_lock);
+  return ++last_log;
```
liewegas: We could just turn last_log into std::atomic<uint64_t> instead...

liewegas:
@xiaoxichen ping

xiaoxichen:
@liewegas comment addressed, thx

tchaikov:
lgtm.

src/common/LogClient.cc
```diff
@@ -26,6 +26,7 @@
 #include "mon/MonMap.h"
 
 #include <iostream>
+#include <atomic>
```
tchaikov: could remove this include.

xiaoxichen:
@tchaikov addressed.

liewegas:
lgtm!

tchaikov:
- http://pulpito.ceph.com/kchai-2016-11-13_05:09:22-rados-wip-kefu-testing---basic-smithi/
- http://pulpito.ceph.com/kchai-2016-11-11_07:45:56-rados-wip-kefu-testing---basic-smithi/

PR #10197, opened 2016-07-08
mon: write fsid use the right return value

songbaisen:
mon: write fsid use the right return value


tchaikov:
lgtm.

yuriw:
http://pulpito.ceph.com/yuriw-2016-07-15_16:58:50-rados-wip-yuri-testing2_2016_7_15-distro-basic-smithi/

PR #10198, opened 2016-07-08
hammer: Failing file operations on kernel based cephfs mount point leaves unaccessible file behind on hammer 0.94.7

smithfarm:
http://tracker.ceph.com/issues/16626

### DIFF
--- a/src/mds/Server.cc
+++ b/src/mds/Server.cc
@@ -2732,9 +2732,13 @@ void Server::handle_client_open(MDRequestRef& mdr)
     return;
   }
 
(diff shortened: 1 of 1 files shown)


smithfarm:
@jcsp This PR passed a cephfs integration run with two failures. Both failures are reproducible in both hammer-backports and hammer itself. One failure is due to http://tracker.ceph.com/issues/11258 ("cephfs-java ftruncate unit test failure") which is fixed in master, but the hammer backport is non-trivial. The second failure is probably an issue with some of the tests. For details, see http://tracker.ceph.com/issues/15895#note-21 

PR #10200, opened 2016-07-08
osdc/mds: fixup pos parameter in the journaler

renhwztetecs:
the type of write_pos\read_pos\expire_pos\trimming_pos is uint64_t, 
change the "pos parameter" to uint64_t


jcsp:
Reviewed-by, will include on next test branch

PR #10201, opened 2016-07-08
os/bluestore: misc cleanups

xiexingguo:


src/os/bluestore/BlueStore.cc
```diff
@@ -1465,6 +1465,8 @@ void BlueStore::_set_compression()
     }
   }
   comp_mode = m;
+  if (comp_mode == COMP_NONE)
```
ifed01: Probably it's better to avoid compressor allocation at all if mode is inappropriate. I.e. calculate the mode first and then proceed with allocation

src/os/bluestore/bluestore_types.h
```diff
@@ -283,7 +283,7 @@ struct bluestore_blob_t {
     }
     b_len += b_off;
     while (b_len) {
-      if (!p->is_valid()) {
+      if (p == extents.end() || !p->is_valid()) {
```
ifed01: we should probably remove assert above (line 282) and return false too when iterator reaches the end.

xiexingguo:
@ifed01 Done, please check again!

src/os/bluestore/BlueStore.cc
```diff
@@ -1429,42 +1429,52 @@ void BlueStore::_set_compression()
...
-    derr << __func__ << " unrecognized compression algorithm '"
-	 << g_conf->bluestore_compression_algorithm << "'" << dendl;
-  }
-  if (alg) {
-    compressor = Compressor::create(cct, alg);
-    if (!compressor) {
-      derr << __func__ << " unable to initialize " << alg << " compressor"
-	   << dendl;
-    }
+  CompressionMode m;
```
ifed01: One can omit this and use comp_mode directly...

src/os/bluestore/bluestore_types.h
ifed01: @xiexingguo, I'm sorry for waving back and forth but probably assert is still better here and above. We should always have all pextents present in the list - either valid or not. The lack of specific pextent (or out of range access) should be considered as an unexpected case and hence asserted IMHO. Thus desired fix should be just an assert before p->is_valid call. What do you think?

xiexingguo: @ifed01  This sounds reasonable to me:-)

xiexingguo:
@ifed01  Upgraded again as per your gorgeous ideas. Please check, thanks!

PR #10202, opened 2016-07-08
mds/MDSMap output None instead of 0 when no fs present.

xiaoxichen:
When no fs present, the mdsmap::dump will output metadata_pool: 0,
which is misleading.

Change the initial value of metadata_pool from 0 to -1 , and output None
when seeing -1.

Fixes: http://tracker.ceph.com/issues/16588

### DIFF
--- a/src/mds/MDSMap.h
+++ b/src/mds/MDSMap.h
@@ -225,7 +225,7 @@ public:
       session_autoclose(0),
       max_file_size(0),
       cas_pool(-1),
-      metadata_pool(0),
+      metadata_pool(-1),
       max_mds(0),
       ever_allowed_features(0),
       explicitly_allowed_features(0),


src/mds/MDSMap.cc
```diff
@@ -177,11 +177,20 @@ void MDSMap::dump(Formatter *f) const
...
   }
   f->close_section();
-  f->open_array_section("data_pools");
-  for (set<int64_t>::const_iterator p = data_pools.begin(); p != data_pools.end(); ++p)
-    f->dump_int("pool", *p);
-  f->close_section();
-  f->dump_int("metadata_pool", metadata_pool);
+
+  if (data_pools.empty())
+    f->dump_string("data_pools", "None");
```
jcsp: Let's not output None here: an empty list (the current behaviour) is more appropriate

src/mds/MDSMap.cc
```diff
@@ -177,11 +177,20 @@ void MDSMap::dump(Formatter *f) const
...
+  else {
+    f->open_array_section("data_pools");
+    for (set<int64_t>::const_iterator p = data_pools.begin(); p != data_pools.end(); ++p)
+      f->dump_int("pool", *p);
+    f->close_section();
+  }
+
+  if (metadata_pool == -1)
+    f->dump_string("metadata_pool", "None");
```
jcsp: If we could output null then this would be good, but unfortunately the Formatter class doesn't support that.  Outputting a string is not an improvement either: it makes sense to human eyes but if a piece of code is parsing the output it might be confused.

jcsp:
I like the idea of using None/null here, but the formatter just doesn't support it so I think it doesn't make sense to use a "None" string.  I like initialising it to -1 though.

xiaoxichen:
@jcsp  would this looks better?

  if (metadata_pool == -1) {
    f->open_array_section("metadata_pool");
    f->close_section();
}

jcsp:
@xiaoxichen no, I don't think an empty list is better.  Like I said, None/null would be good, but we just don't have a way to output that at present, so a -1 is probably the most correct thing to output.

xiaoxichen:
@jcsp fixed:)

PR #10204, opened 2016-07-08
cmake: fixes for pypi change

tchaikov:
this fixes the failure like
```
+ echo 'error: Added files:'
error: Added files:
+ cat .git/added-files
src/pybind/rados/rados.egg-info/PKG-INFO
...
```

see http://gitbuilder.sepia.ceph.com/gitbuilder-ceph-tarball-trusty-amd64-cmake/log.cgi?log=07387eb9af16b4916c64259d8aa98267b34b3a6c


jdurgin:
lgtm

PR #10205, opened 2016-07-08
rgw: add missing master_zone when running with old default region config

oritwas:
Fixes: http://tracker.ceph.com/issues/16627

### DIFF
--- a/src/rgw/rgw_rados.cc
+++ b/src/rgw/rgw_rados.cc
@@ -3355,6 +3355,17 @@ int RGWRados::replace_region_with_zonegroup()
       zoneparams.metadata_heap = ".rgw.meta";
       return zoneparams.update();
     }
+    /* update master zone */
+    RGWZoneGroup default_zg(default_zonegroup_name);
+    ret = default_zg.init(cct, this);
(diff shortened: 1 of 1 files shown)


src/rgw/rgw_rados.cc
```diff
@@ -3355,6 +3355,17 @@ int RGWRados::replace_region_with_zonegroup()
...
+    /* update master zone */
+    RGWZoneGroup default_zg(default_zonegroup_name);
+    ret = default_zg.init(cct, this);
+    if (ret < 0 && ret != -ENOENT) {
+      ldout(cct, 0) << __func__ << ": error in initializing default zonegroup: " << cpp_strerror(-ret) << dendl;
+      return ret;
+    }
+    if (ret != -ENOENT && default_zg.master_zone.empty()) {
+      default_zg.master_zone = zoneparams.get_id();
+      return default_zg.update();
```
cbodley: might want to print an error message if this fails? or do we log something higher up the stack?

oritwas: it is logged by the caller

cbodley:
:+1:

PR #10207, opened 2016-07-08
rgw: Fix incorrect content length and range for zero sized objects during range requests

prallabh:
Fixes: http://tracker.ceph.com/issues/16388

### DIFF
--- a/src/rgw/rgw_rest.cc
+++ b/src/rgw/rgw_rest.cc
@@ -655,9 +655,7 @@ void end_header(struct req_state* s, RGWOp* op, const char *content_type,
       s->formatter->close_section();
     }
     s->formatter->output_footer();
-    if (s->obj_size) {
-      dump_content_length(s, s->formatter->get_len());
-    }
+    dump_content_length(s, s->formatter->get_len());
   } else {
     if (proposed_content_length != NO_CONTENT_LENGTH) {
       dump_content_length(s, proposed_content_length);
--- a/src/rgw/rgw_rest.cc
+++ b/src/rgw/rgw_rest.cc
@@ -655,7 +655,9 @@ void end_header(struct req_state* s, RGWOp* op, const char *content_type,
       s->formatter->close_section();
     }
     s->formatter->output_footer();
-    dump_content_length(s, s->formatter->get_len());
+    if (s->obj_size) {
+      dump_content_length(s, s->formatter->get_len());
+    }
   } else {
     if (proposed_content_length != NO_CONTENT_LENGTH) {
       dump_content_length(s, proposed_content_length);
@@ -769,8 +771,13 @@ void dump_range(struct req_state *s, uint64_t ofs, uint64_t end,
 
   /* dumping range into temp buffer first, as libfcgi will fail to digest
    * %lld */
-  snprintf(range_buf, sizeof(range_buf), "%lld-%lld/%lld", (long long)ofs,
-	   (long long)end, (long long)total);
+
+  if (!total) {
+    snprintf(range_buf, sizeof(range_buf), "*/%lld", (long long)total);
+  } else {
+    snprintf(range_buf, sizeof(range_buf), "%lld-%lld/%lld", (long long)ofs,
+		(long long)end, (long long)total);
+  }
   int r = STREAM_IO(s)->print("Content-Range: bytes %s\r\n", range_buf);
   if (r < 0) {
     ldout(s->cct, 0) << "ERROR: s->cio->print() returned err=" << r << dendl;
--- a/src/rgw/rgw_rest_swift.cc
+++ b/src/rgw/rgw_rest_swift.cc
@@ -1206,13 +1206,14 @@ int RGWGetObj_ObjStore_SWIFT::send_response_data(bufferlist& bl,
   set_req_state_err(s, (partial_content && !op_ret) ? STATUS_PARTIAL_CONTENT
 		    : op_ret);
   dump_errno(s);
-  if (s->err.is_err()) {
-    end_header(s, NULL);
-    return 0;
-  }
 
   if (range_str) {
     dump_range(s, ofs, end, s->obj_size);
+  }
+
+  if (s->err.is_err()) {
+    end_header(s, NULL);
+    return 0;
   }
 
   dump_content_length(s, total_len);
--- a/src/rgw/rgw_op.cc
+++ b/src/rgw/rgw_op.cc
@@ -1341,6 +1341,13 @@ void RGWGetObj::execute()
   if (op_ret < 0)
     goto done_err;
 
+  // for range requests with obj size 0
+  if (range_str && !(s->obj_size)) {
+    total_len = 0;
+    op_ret = -ERANGE;
+    goto done_err;
+  }
+
   /* start gettorrent */
   if (torrent.get_flag())
   {


prallabh:
@yehudasa can you please help review

prallabh:
@cbodley, @oritwas can you please help review/assign as appropriate.

oritwas:
This fixes the content-length header but the content-range is still not correct
I get with this fix for an empty object:
Content-Range: bytes 0--1/0

prallabh:
That's true, should the range be 0-0/0 as well?

oritwas:
it should be:
Content-Range: */0

prallabh:
I guess you meant that to be whatever user specifies as the range? Thinking through this, wouldn't any range be irrelevant for zero sized objects?

oritwas:
We need to respond with 416 (Range Not Satisfiable) and 
Content-Range: */0 

see http://httpwg.org/specs/rfc7233.html#status.416

oritwas:
@prallabh , Lets return 416 and remove the content-range header . I  think this will be the simplest

prallabh:
@oritwas Thanks for the link, wasn't aware of it. Realized that we were returning 416 else where as well in the code, was in the process of building the content-range for this case. If we do not bother about the range header, will update the commit with just the content length and 416.

prallabh:
@oritwas I ended up fixing the range header as per the RFC, please take a look at the same.

vsespb:
Here is what Amazon S3 gives (yes, 416):

```
$ curl -v -H "Range: bytes=0-5242880" https://s3-eu-west-1.amazonaws.com/vsetest/s3test
* About to connect() to s3-eu-west-1.amazonaws.com port 443 (#0)
*   Trying 54.231.131.83... connected
* successfully set certificate verify locations:
*   CAfile: none
  CApath: /etc/ssl/certs
* SSLv3, TLS handshake, Client hello (1):
* SSLv3, TLS handshake, Server hello (2):
* SSLv3, TLS handshake, CERT (11):
* SSLv3, TLS handshake, Server key exchange (12):
* SSLv3, TLS handshake, Server finished (14):
* SSLv3, TLS handshake, Client key exchange (16):
* SSLv3, TLS change cipher, Client hello (1):
* SSLv3, TLS handshake, Finished (20):
* SSLv3, TLS change cipher, Client hello (1):
* SSLv3, TLS handshake, Finished (20):
* SSL connection using ECDHE-RSA-AES128-SHA
* Server certificate:
* 	 subject: C=US; ST=Washington; L=Seattle; O=Amazon.com Inc.; CN=*.s3-eu-west-1.amazonaws.com
* 	 start date: 2015-12-08 12:05:08 GMT
* 	 expire date: 2016-09-14 12:00:00 GMT
* 	 subjectAltName: s3-eu-west-1.amazonaws.com matched
* 	 issuer: C=US; O=DigiCert Inc; OU=www.digicert.com; CN=DigiCert Baltimore CA-2 G2
* 	 SSL certificate verify ok.
> GET /vsetest/s3test HTTP/1.1
> User-Agent: curl/7.22.0 (x86_64-pc-linux-gnu) libcurl/7.22.0 OpenSSL/1.0.1 zlib/1.2.3.4 libidn/1.23 librtmp/2.3
> Host: s3-eu-west-1.amazonaws.com
> Accept: */*
> Range: bytes=0-5242880
> 
< HTTP/1.1 416 Requested Range Not Satisfiable
< x-amz-request-id: EBA23471F0510B1A
< x-amz-id-2: RYTrIF6LmawFzMVTsRQd/5dqF3YRJ+8jof6WgXQ5EAr5oIkBorFcdOexrTqVnojpHHl1KMWkvZk=
< Content-Type: application/xml
< Transfer-Encoding: chunked
< Date: Mon, 18 Jul 2016 11:22:15 GMT
< Server: AmazonS3
< 
<?xml version="1.0" encoding="UTF-8"?>
* Connection #0 to host s3-eu-west-1.amazonaws.com left intact
* Closing connection #0
* SSLv3, TLS alert, Client hello (1):
<Error><Code>InvalidRange</Code><Message>The requested range is not satisfiable</Message><RangeRequested>bytes=0-5242880</RangeRequested><ActualObjectSize>0</ActualObjectSize><RequestId>EBA23471F0510B1A</RequestId><HostId>RYTrIF6LmawFzMVTsRQd/5dqF3YRJ+8jof6WgXQ5EAr5oIkBorFcdOexrTqVnojpHHl1KMWkvZk=</HostId></Error>
```

vsespb:
I also wondering why you guys have travis "build", but don't write tests. Different condition-branch for handling zero is something tests designed for.

tchaikov:
@vsespb we don't have travis build yet. why do you say so?

vsespb:
ok, jenkins, not travis.

prallabh:
@oritwas ping

oritwas:
lgtm, but it need more testing.
Can you add a test to s3tests for this?

prallabh:
Thanks, do you mean to add a test under https://github.com/ceph/s3-tests. Have never done that, any pointers/documentation around best practices for that would help.

prallabh:
@oritwas I have verified the fix swift/s3 clients, can you please clarify on the test that you were referring to.

oritwas:
I add a test to s3 tests for an empty object see https://github.com/ceph/s3-tests/pull/119

could you try it an see if it passes?

prallabh:
Thanks for the test, find below the result:

S3TEST_CONF=range.conf ./virtualenv/bin/nosetests s3tests.functional.test_s3:test_ranged_request_empty_object
.
Ran 1 test in 0.568s

OK

prallabh:
@oritwas rebased.

prallabh:
@oritwas ping

prallabh:
@oritwas Realized that having the object size check in the Read::prepare would break while preforming operations like stat/download on a empty object. Moved the checks ahead to RGWGetObj::execute where they should really be IMO. Please take a look, thanks!

oritwas:
@prallabh can you rebase?

prallabh:
@oritwas done

prallabh:
@oritwas please let me know if there is anything pending from my side for merge.

src/rgw/rgw_rest.cc
```diff
@@ -655,7 +655,9 @@ void end_header(struct req_state* s, RGWOp* op, const char *content_type,
       s->formatter->close_section();
     }
     s->formatter->output_footer();
-    dump_content_length(s, s->formatter->get_len());
+    if (s->obj_size) {
```
oritwas: This seems to effect all error response and some require content-length. is it a problem to keep the content-length?

src/rgw/rgw_rest_swift.cc
```diff
@@ -1205,17 +1205,20 @@ int RGWGetObj_ObjStore_SWIFT::send_response_data(bufferlist& bl,
 
   set_req_state_err(s, (partial_content && !op_ret) ? STATUS_PARTIAL_CONTENT
 		    : op_ret);
+
```
oritwas: remove the empty line

src/rgw/rgw_rest_swift.cc
```diff
@@ -1205,17 +1205,20 @@ int RGWGetObj_ObjStore_SWIFT::send_response_data(bufferlist& bl,
...
-    end_header(s, NULL);
-    return 0;
-  }
 
   if (range_str) {
     dump_range(s, ofs, end, s->obj_size);
   }
 
   dump_content_length(s, total_len);
+
```
oritwas: move it before the dump_content_length and than you don't need to remove the content-length in end-header

src/rgw/rgw_rest.cc
prallabh: The content length here is that of the end header's and its not relevant in this context. For range requests on zero sized objects, without suppressing this header, the output looks something like this:

HTTP/1.1 416 Requested Range Not Satisfiable
Content-Range: bytes */0
Content-Length: 0
Accept-Ranges: bytes
X-Trans-Id: tx000000000000000000026-00579f3001-1011-default
Content-Length: 12
Accept-Ranges: bytes
Content-Type: text/plain; charset=utf-8
Date: Mon, 01 Aug 2016 11:18:25 GMT

src/rgw/rgw_rest_swift.cc
prallabh: This won't help as the content length here is that of the object's, which is required.

src/rgw/rgw_rest_swift.cc
prallabh: done

src/rgw/rgw_rest_swift.cc
oritwas: In case of an error the content length of the object doesn't matter, 
it should be the content length of the error message (some error message have body)

src/rgw/rgw_rest.cc
oritwas: you have here two content lengths headers,
it is because you add the content length even in case of an error see my comment on rgw_rest_swift.cc

oritwas:
As bothr swift or AWS are not returning a content-range header in case of such error, we can do the same.
It will simplify the code just to return 416 error if the object is empty

prallabh:
@oritwas Thanks, can you please check now if that is what you are expecting.

src/rgw/rgw_rest_swift.cc
```diff
@@ -1206,16 +1206,18 @@ int RGWGetObj_ObjStore_SWIFT::send_response_data(bufferlist& bl,
...
     dump_range(s, ofs, end, s->obj_size);
   }
 
+  if (s->err.is_err()) {
+    end_header(s, NULL);
+    return 0;
+  }
+
   dump_content_length(s, total_len);
+
```
oritwas: small nit : extra empty line

prallabh:
@oritwas updated and rebased, please help merge.

src/rgw/rgw_rest_swift.cc
prallabh: done, thanks!

oritwas:
looks good

PR #10209, opened 2016-07-08
osd/ReplicatedPG: remove redundant check for balance/localize read

runsisi:
given an op if the PG is a replica and is able to handle it, then the
op must have the BALANCE_READS or LOCALIZE_READS flag(s) set, otherwise
the op was misdirected and had been rejected at the every beginning of
the function.


dachary:
jenkins test this please (jenkins stuck)

dachary:

xiexingguo:
@athanatos  Can you double check this?

athanatos:
lgtm

PR #10213, opened 2016-07-08
rgw: add support for the healthcheck feature of Swift API.

rzarzynski:


rzarzynski:
Bundled in #10280. Closing.

PR #10214, opened 2016-07-08
rgw: drop create_bucket in fwd_request log message

theanalyst:
Since both create and delete bucket use the same function


theanalyst:
jenkins retest this please

PR #10215, opened 2016-07-08
os/bluestore: refactor dirty blob tracking along with some related fixes

ifed01:
This is an intermediate patch before introducing refcounting for Blob instances.

### DIFF
--- a/src/os/bluestore/BlueStore.h
+++ b/src/os/bluestore/BlueStore.h
@@ -820,6 +820,8 @@ public:
 
     set<OnodeRef> onodes;     ///< these onodes need to be updated/written
     set<BnodeRef> bnodes;     ///< these bnodes need to be updated/written
+    set<BlobRef> blobs;       ///< these blobs need to be updated on io completion
+
     KeyValueDB::Transaction t; ///< then we will commit this
     Context *oncommit;         ///< signal on commit
     Context *onreadable;         ///< signal on readable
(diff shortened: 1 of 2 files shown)


markhpc:
retest this please

markhpc:
Note:  The above comment was to get jenkins to retest the build.

markhpc:
Hi Igor, this branch has conflicts.  Can you please fix?  Thanks!

ifed01:
@markhpc - rebased

markhpc:
retest this please

src/os/bluestore/BlueStore.h
```diff
@@ -328,25 +340,25 @@ class BlueStore : public ObjectStore,
       return nullptr;
     }
 
-    Blob *new_blob(Cache *c) {
+    BlobRef new_blob(Cache *c) {
       int64_t id = get_new_id();
       Blob *b = new Blob(id, c);
+      b->get();
```
liewegas: If BlobRef is an intrusive_ptr there shouldn't be any explicit get/put calls anywhere (except the intrustive ptr hooks)...

liewegas:
maybe squash c8a765c  and 6d6d940 since the latter replaces all of the former's code anyway?

liewegas:
otherwise lgtm!

src/os/bluestore/BlueStore.h
ifed01: @liewegas. Additional ref_count increment is needed since Blobs are stored in intrusive::set<Blob> blob_map. Blob has to persist while it's stored in this blob_map even if no corresponding BlobRefs are present.
Other options are either using std::set<BlobRef> or intrusive::set<BlobRef> but IMHO they both have some performance/memory footprint drawbacks. Have I missed something and there is better approach?

liewegas: Oh, I see.  Yep!

ifed01:
@liewegas squashed and rebased.

PR #10218, opened 2016-07-09
crush: fix multi-pick weight anomaly

adamemerson:
List and Straw2 buckets.

(Since Tree is disabled by default, I assume Straw is going to be superseded, and Uniform is uniform. But I'll do Tree and Straw if people think it's important.)

Should close bug 15653

### DIFF
--- a/src/crush/mapper.c
+++ b/src/crush/mapper.c
@@ -52,7 +52,6 @@ int crush_find_rule(const struct crush_map *map, int ruleset, int type, int size
 	return -1;
 }
 
-
 /*
  * bucket choose methods
  *
@@ -70,59 +69,60 @@ int crush_find_rule(const struct crush_map *map, int ruleset, int type, int size
  * Since this is expensive, we optimize for the r=0 case, which
  * captures the vast majority of calls.
  */
-static int bucket_perm_choose(struct crush_bucket *bucket,
+static int bucket_perm_choose(const struct crush_bucket *bucket,
+			      struct crush_work_bucket *work,
 			      int x, int r)
 {
 	unsigned int pr = r % bucket->size;
 	unsigned int i, s;
 
 	/* start a new permutation if @x has changed */
-	if (bucket->perm_x != (__u32)x || bucket->perm_n == 0) {
+	if (work->perm_x != (__u32)x || work->perm_n == 0) {
 		dprintk("bucket %d new x=%d\n", bucket->id, x);
-		bucket->perm_x = x;
+		work->perm_x = x;
 
 		/* optimize common r=0 case */
 		if (pr == 0) {
 			s = crush_hash32_3(bucket->hash, x, bucket->id, 0) %
 				bucket->size;
-			bucket->perm[0] = s;
-			bucket->perm_n = 0xffff;   /* magic value, see below */
+			work->perm[0] = s;
+			work->perm_n = 0xffff;   /* magic value, see below */
 			goto out;
 		}
 
 		for (i = 0; i < bucket->size; i++)
-			bucket->perm[i] = i;
-		bucket->perm_n = 0;
-	} else if (bucket->perm_n == 0xffff) {
+			work->perm[i] = i;
+		work->perm_n = 0;
+	} else if (work->perm_n == 0xffff) {
 		/* clean up after the r=0 case above */
 		for (i = 1; i < bucket->size; i++)
-			bucket->perm[i] = i;
-		bucket->perm[bucket->perm[0]] = 0;
-		bucket->perm_n = 1;
+			work->perm[i] = i;
+		work->perm[work->perm[0]] = 0;
+		work->perm_n = 1;
 	}
 
 	/* calculate permutation up to pr */
-	for (i = 0; i < bucket->perm_n; i++)
+	for (i = 0; i < work->perm_n; i++)
 		dprintk(" perm_choose have %d: %d\n", i, bucket->perm[i]);
-	while (bucket->perm_n <= pr) {
-		unsigned int p = bucket->perm_n;
+	while (work->perm_n <= pr) {
+		unsigned int p = work->perm_n;
 		/* no point in swapping the final entry */
 		if (p < bucket->size - 1) {
 			i = crush_hash32_3(bucket->hash, x, bucket->id, p) %
 				(bucket->size - p);
 			if (i) {
-				unsigned int t = bucket->perm[p + i];
-				bucket->perm[p + i] = bucket->perm[p];
-				bucket->perm[p] = t;
+				unsigned int t = work->perm[p + i];
+				work->perm[p + i] = work->perm[p];
+				work->perm[p] = t;
 			}
 			dprintk(" perm_choose swap %d with %d\n", p, p+i);
 		}
-		bucket->perm_n++;
+		work->perm_n++;
 	}
 	for (i = 0; i < bucket->size; i++)
 		dprintk(" perm_choose  %d: %d\n", i, bucket->perm[i]);
 
-	s = bucket->perm[pr];
+	s = work->perm[pr];
 out:
 	dprintk(" perm_choose %d sz=%d x=%d r=%d (%d) s=%d\n", bucket->id,
 		bucket->size, x, r, pr, s);
@@ -130,14 +130,14 @@ out:
 }
 
 /* uniform */
-static int bucket_uniform_choose(struct crush_bucket_uniform *bucket,
-				 int x, int r)
+static int bucket_uniform_choose(const struct crush_bucket_uniform *bucket,
+				 struct crush_work_bucket *work, int x, int r)
 {
-	return bucket_perm_choose(&bucket->h, x, r);
+	return bucket_perm_choose(&bucket->h, work, x, r);
 }
 
 /* list */
-static int bucket_list_choose(struct crush_bucket_list *bucket,
+static int bucket_list_choose(const struct crush_bucket_list *bucket,
 			      int x, int r)
 {
 	int i;
@@ -153,8 +153,9 @@ static int bucket_list_choose(struct crush_bucket_list *bucket,
 		w *= bucket->sum_weights[i];
 		w = w >> 16;
 		/*dprintk(" scaled %llx\n", w);*/
-		if (w < bucket->item_weights[i])
+		if (w < bucket->item_weights[i]) {
 			return bucket->h.items[i];
+		}
 	}
 
 	dprintk("bad list sums for bucket %d\n", bucket->h.id);
@@ -190,7 +191,7 @@ static int terminal(int x)
 	return x & 1;
 }
 
-static int bucket_tree_choose(struct crush_bucket_tree *bucket,
+static int bucket_tree_choose(const struct crush_bucket_tree *bucket,
 			      int x, int r)
 {
 	int n;
@@ -222,7 +223,7 @@ static int bucket_tree_choose(struct crush_bucket_tree *bucket,
 
 /* straw */
 
-static int bucket_straw_choose(struct crush_bucket_straw *bucket,
+static int bucket_straw_choose(const struct crush_bucket_straw *bucket,
 			       int x, int r)
 {
 	__u32 i;
@@ -255,7 +256,7 @@ static __u64 crush_ln(unsigned int xin)
 	iexpon = 15;
 
 	// figure out number of bits we need to shift and
-	// do it in one step instead of iteratively	
+	// do it in one step instead of iteratively
 	if (!(x & 0x18000)) {
 	  int bits = __builtin_clz(x & 0x1FFFF) - 16;
 	  x <<= bits;
@@ -297,7 +298,7 @@ static __u64 crush_ln(unsigned int xin)
  *
  */
 
-static int bucket_straw2_choose(struct crush_bucket_straw2 *bucket,
+static int bucket_straw2_choose(const struct crush_bucket_straw2 *bucket,
 				int x, int r)
 {
 	unsigned int i, high = 0;
@@ -340,36 +341,41 @@ static int bucket_straw2_choose(struct crush_bucket_straw2 *bucket,
 			high_draw = draw;
 		}
 	}
+
 	return bucket->h.items[high];
 }
 
 
-static int crush_bucket_choose(struct crush_bucket *in, int x, int r)
+static int crush_bucket_choose(const struct crush_bucket *in,
+			       struct crush_work_bucket *work,
+			       int x, int r)
 {
 	dprintk(" crush_bucket_choose %d x=%d r=%d\n", in->id, x, r);
 	BUG_ON(in->size == 0);
 	switch (in->alg) {
 	case CRUSH_BUCKET_UNIFORM:
-		return bucket_uniform_choose((struct crush_bucket_uniform *)in,
-					  x, r);
+		return bucket_uniform_choose(
+			(const struct crush_bucket_uniform *)in,
+			work, x, r);
 	case CRUSH_BUCKET_LIST:
-		return bucket_list_choose((struct crush_bucket_list *)in,
+		return bucket_list_choose((const struct crush_bucket_list *)in,
 					  x, r);
 	case CRUSH_BUCKET_TREE:
-		return bucket_tree_choose((struct crush_bucket_tree *)in,
+		return bucket_tree_choose((const struct crush_bucket_tree *)in,
 					  x, r);
 	case CRUSH_BUCKET_STRAW:
-		return bucket_straw_choose((struct crush_bucket_straw *)in,
-					   x, r);
+		return bucket_straw_choose(
+			(const struct crush_bucket_straw *)in,
+			x, r);
 	case CRUSH_BUCKET_STRAW2:
-		return bucket_straw2_choose((struct crush_bucket_straw2 *)in,
-					    x, r);
+		return bucket_straw2_choose(
+			(const struct crush_bucket_straw2 *)in,
+			x, r);
 	default:
 		dprintk("unknown bucket %d alg %d\n", in->id, in->alg);
 		return in->items[0];
 	}
 }
-
 
 /*
  * true if device is marked "out" (failed, fully offloaded)
@@ -412,7 +418,8 @@ static int is_out(const struct crush_map *map,
  * @parent_r: r value passed from the parent
  */
 static int crush_choose_firstn(const struct crush_map *map,
-			       struct crush_bucket *bucket,
+			       struct crush_work *work,
+			       const struct crush_bucket *bucket,
 			       const __u32 *weight, int weight_max,
 			       int x, int numrep, int type,
 			       int *out, int outpos,
@@ -430,7 +437,7 @@ static int crush_choose_firstn(const struct crush_map *map,
 	int rep;
 	unsigned int ftotal, flocal;
 	int retry_descent, retry_bucket, skip_rep;
-	struct crush_bucket *in = bucket;
+	const struct crush_bucket *in = bucket;
 	int r;
 	int i;
 	int item = 0;
@@ -452,7 +459,7 @@ parent_r %d stable %d\n",
 		skip_rep = 0;
 		do {
 			retry_descent = 0;
-			in = bucket;               /* initial bucket */
+			in = bucket;              /* initial bucket */
 
 			/* choose through intervening buckets */
 			flocal = 0;
@@ -471,9 +478,13 @@ parent_r %d stable %d\n",
 				if (local_fallback_retries > 0 &&
 				    flocal >= (in->size>>1) &&
 				    flocal > local_fallback_retries)
-					item = bucket_perm_choose(in, x, r);
+					item = bucket_perm_choose(
+						in, work->work[-1-in->id],
+						x, r);
 				else
-					item = crush_bucket_choose(in, x, r);
+					item = crush_bucket_choose(
+						in, work->work[-1-in->id],
+						x, r);
 				if (item >= map->max_devices) {
 					dprintk("   bad item %d\n", item);
 					skip_rep = 1;
@@ -516,25 +527,27 @@ parent_r %d stable %d\n",
 							sub_r = r >> (vary_r-1);
 						else
 							sub_r = 0;
-						if (crush_choose_firstn(map,
-							 map->buckets[-1-item],
-							 weight, weight_max,
-							 x, stable ? 1 : outpos+1, 0,
-							 out2, outpos, count,
-							 recurse_tries, 0,
-							 local_retries,
-							 local_fallback_retries,
-							 0,
-							 vary_r,
-							 stable,
-							 NULL,
-							 sub_r) <= outpos)
+						if (crush_choose_firstn(
+							    map,
+							    work,
+							    map->buckets[-1-item],
+							    weight, weight_max,
+							    x, stable ? 1 : outpos+1, 0,
+							    out2, outpos, count,
+							    recurse_tries, 0,
+							    local_retries,
+							    local_fallback_retries,
+							    0,
+							    vary_r,
+							    stable,
+							    NULL,
+							    sub_r) <= outpos)
 							/* didn't get leaf */
 							reject = 1;
 					} else {
 						/* we already have a leaf! */
 						out2[outpos] = item;
-					}
+		}
 				}
 
 				if (!reject) {
@@ -598,7 +611,8 @@ reject:
  *
  */
 static void crush_choose_indep(const struct crush_map *map,
-			       struct crush_bucket *bucket,
+			       struct crush_work *work,
+			       const struct crush_bucket *bucket,
 			       const __u32 *weight, int weight_max,
 			       int x, int left, int numrep, int type,
 			       int *out, int outpos,
@@ -608,7 +622,7 @@ static void crush_choose_indep(const struct crush_map *map,
 			       int *out2,
 			       int parent_r)
 {
-	struct crush_bucket *in = bucket;
+	const struct crush_bucket *in = bucket;
 	int endpos = outpos + left;
 	int rep;
 	unsigned int ftotal;
@@ -676,7 +690,9 @@ static void crush_choose_indep(const struct crush_map *map,
 					break;
 				}
 
-				item = crush_bucket_choose(in, x, r);
+				item = crush_bucket_choose(
+					in, work->work[-1-in->id],
+					x, r);
 				if (item >= map->max_devices) {
 					dprintk("   bad item %d\n", item);
 					out[rep] = CRUSH_ITEM_NONE;
@@ -722,13 +738,15 @@ static void crush_choose_indep(const struct crush_map *map,
 
 				if (recurse_to_leaf) {
 					if (item < 0) {
-						crush_choose_indep(map,
-						   map->buckets[-1-item],
-						   weight, weight_max,
-						   x, 1, numrep, 0,
-						   out2, rep,
-						   recurse_tries, 0,
-						   0, NULL, r);
+						crush_choose_indep(
+							map,
+							work,
+							map->buckets[-1-item],
+							weight, weight_max,
+							x, 1, numrep, 0,
+							out2, rep,
+							recurse_tries, 0,
+							0, NULL, r);
 						if (out2[rep] == CRUSH_ITEM_NONE) {
 							/* placed nothing; no leaf */
 							break;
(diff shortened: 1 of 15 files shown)


src/crush/mapper.c
```diff
@@ -833,6 +928,25 @@ int crush_do_rule(const struct crush_map *map,
...
+		/* This and other reweighting modifications are
+		   externally invisible, so this should be fine. I'm
+		   following the example of the 'perm' array here
+		   which is modified in the course of mapping. */
+		reset_reweight((struct crush_map*)map);
+		/* Instead of modifying the map in-place we could
+		   have builder.c allocate a bookkeeping map, which
+		   is what I would have done if Perm hadn't already
+		   suggested we can keep information like this in
+		   place. */
```
liewegas: FWIW the fact that perm does this means we have to protect crush mappings with a big mutex, which sort of sucks and should really be fixed anyway.  Unfortunately it means we need to build a mapping structure of some sort.  Since crush choices are generally pretty simple (not that many buckets) perhaps a simple flat vector with O(n) lookups is fine?  We're stuck with straight C and can't easily use the kernel rbtree or similar structures since we share the code with userspace.  :/

In any case, though, I think we should do this one the "right" way with a local map so that the crush mapping function can be eventually made const.

src/crush/CrushCompiler.cc
```diff
@@ -203,14 +203,16 @@ int CrushCompiler::decompile(ostream &out)
   if (crush.get_allowed_bucket_algs() != CRUSH_LEGACY_ALLOWED_BUCKET_ALGS)
     out << "tunable allowed_bucket_algs " << crush.get_allowed_bucket_algs()
 	<< "\n";
+  if (crush.get_reweight_items() != 0)
+    out << "tunable reweight_items " << crush.get_reweight_items() << "\n";
```
liewegas: I wonder if we can come up with a more descriptive name.. 'reweight_multi_pick" or something?

liewegas:
Ignoring tree is fine, I think.  Lots of clusters are still using straw but they should just switch to straw2 anyway (it's a close enough mapping that it's safe to do).  Yup!

liewegas:
This looks good, except for the local state thing.  We need to move toward a const mapping anyway so let's do this the right way.  (And if it's not too painful we might fix the perm thing soon as well?)

adamemerson:
I can agree with the local state thing. Is the mapper itself EVER FORBIDDEN TO ALLOCATE MEMORY? Because if I can allocate while running it I can fix this and the perm thing both without changing the interface and we can run without the lock.

liewegas:
Memory allocation is fine.  Try to use kmalloc (it'll get #defined to 
something else in the header if we're not in the kernel). Yay!

liewegas:
ping!

liewegas:
Yay!  I like the single preallocated buffer.  I think we can even make crush_do_rule drop the setup and require callers to do it, too.  I wonder, actaully, if we want to expose explicit functions for initializing the work space so that the user can call it once per map update instead of reinitializing it on every mapping attempt.. otherwise it's O(number of buckets) of additional overhead.

Also, I still think we need a test that specifically addresses the multi-pick reweighting.  I.e., have a bucket with skewed weights, do 10000 3x placements, and assert that the low-weighted object is overfull.. and then repeat with the tunable and assert it is not overfull (for straw2 and list, at least).

adamemerson:
I will happily add a test. Is there a Canonical Way to decide something is overfull?

Running the preparation function only once is troublesome since we still need to set the parent weights and such. From what I can tell there's no /good/ way at present for bucket_choose to know if this is the first time it's being asked to choose from a bucket (since r might have parent_r.) If we could get an easy way to do that, we could just pay to do the setup on the first time the bucket is used. (For straw it's trivial, for list it would just be running up the sum.)

liewegas:
Hrm, yeah.  One option would be to have u64 counter in the front of the 
work structure.  then crush_init_work() would set it to 0, and 
crush_reset_work would increment it.  The bucket could compare it to 
a value in its local workspace and notice when it's changed (implying a 
new mapping call).

Although really, the only difference between mappings should be 'x' and 
'r'.  I dont' think it matters if it's the nested r or the parent r.. 
right?

adamemerson:
Good idea. I had thought of (and discraded) checking the clock for the
same purpose, but a counter should be perfect.

As to r, I just meant that given parent_r, I couldn't test for r == 0,
so the Obvious Lazy Way won't work.

idryomov:
@liewegas, @adamemerson How much work remains here?  If it lands in the next 2-3 weeks, we can look into pulling this into kernel 4.9, along with #10935.

src/crush/mapper.c
```diff
@@ -823,19 +992,46 @@ int crush_do_rule(const struct crush_map *map,
...
+	   potentially useful, but we could get rid of it. */
+
+	if (cwin)
+		cw = cwin;
+	else
+		cw = kmalloc(map->working_size, GFP_KERNEL);
+
+	if (!cw)
+		return 0;
+
```
idryomov: No need for kmalloc or error/crash part - we should always require the caller to provide workspace and just assume it's there.

src/crush/mapper.c
```diff
@@ -776,6 +885,69 @@ static void crush_choose_indep(const struct crush_map *map,
...
+		default:
+			point += sizeof(struct crush_work_bucket);
+			break;
+		}
+		w->work[b]->perm_x = 0;
+		w->work[b]->perm_n = 0;
+		w->work[b]->perm = (__u32*) point;
+		point += m->buckets[b]->size * sizeof(__u32);
+	}
+	assert((char*) point - (char*) w == m->working_size);
```
idryomov: BUG_ON(... != ...);

src/crush/mapper.c
```diff
@@ -992,5 +1190,9 @@ int crush_do_rule(const struct crush_map *map,
 			break;
 		}
 	}
+
+	if (!cwin)
+		kfree(cw);
```
idryomov: kfree can go along with the kmalloc.

src/crush/mapper.c
```diff
@@ -304,6 +314,15 @@ static int bucket_straw2_choose(struct crush_bucket_straw2 *bucket,
 
 	for (i = 0; i < bucket->h.size; i++) {
 		w = bucket->item_weights[i];
+		if (reweight_items && work->picked_sum > 0) {
+			__u64 v = w;
+			v *= (work->weight_sum -
+			      work->picked_sum);
+			v /= work->weight_sum;
```
idryomov: Define
```
#define div64_u64(dividend, divisor) ((dividend) / (divisor))
```
macro in crush_compat.h and use it here (see div64_s64() usage below).

src/crush/mapper.c
```diff
@@ -776,6 +885,69 @@ static void crush_choose_indep(const struct crush_map *map,
...
+	   values and pointers as we go.
+
+	   It's a bit like Forth's use of the 'allot' word since we
+	   set the pointer first and then reserve the space for it to
+	   point to by incrementing the cursor. */
+	char* point = (char*)w;
+	point += sizeof(struct crush_work*);
+	w->work = (struct crush_work_bucket**) point;
+	point += m->max_buckets * sizeof(struct crush_work_bucket*);
+	for (size_t b = 0; b < m->max_buckets; ++b) {
```
idryomov: Why size_t?  max_buckets is __s32.

src/crush/mapper.c
```diff
@@ -776,6 +885,69 @@ static void crush_choose_indep(const struct crush_map *map,
...
+		case CRUSH_BUCKET_STRAW2:
+			;
+			struct crush_work_straw2* s2w
+				= (struct crush_work_straw2 *) point;
+			point += sizeof(struct crush_work_straw2);
+			struct crush_bucket_straw2* s2b
+				= (struct crush_bucket_straw2 *) m->buckets[b];
+			s2w->weight_sum = 0;
+			s2w->picked_sum = 0;
+			for (size_t i = 0; i < s2b->h.size; ++i)
```
idryomov: Why size_t?  If you aren't using a simple int, then use the type of h.size.

src/crush/mapper.c
```diff
@@ -776,6 +885,69 @@ static void crush_choose_indep(const struct crush_map *map,
...
+	w->work = (struct crush_work_bucket**) point;
+	point += m->max_buckets * sizeof(struct crush_work_bucket*);
+	for (size_t b = 0; b < m->max_buckets; ++b) {
+		if (m->buckets[b] == 0)
+			continue;
+
+		w->work[b] = (struct crush_work_bucket*) point;
+		switch (m->buckets[b]->alg) {
+		case CRUSH_BUCKET_STRAW2:
+			;
```
idryomov: Stray semicolon?

src/crush/mapper.c
```diff
@@ -776,6 +885,69 @@ static void crush_choose_indep(const struct crush_map *map,
...
+			struct crush_work_straw2* s2w
+				= (struct crush_work_straw2 *) point;
```
idryomov: Style: '=' should be on the previous line.

src/crush/mapper.c
```diff
@@ -776,6 +885,69 @@ static void crush_choose_indep(const struct crush_map *map,
...
+			struct crush_bucket_straw2* s2b
+				= (struct crush_bucket_straw2 *) m->buckets[b];
+			s2w->weight_sum = 0;
+			s2w->picked_sum = 0;
+			for (size_t i = 0; i < s2b->h.size; ++i)
+				s2w->weight_sum += s2b->item_weights[i];
+
+			break;
+		case CRUSH_BUCKET_LIST:
+			point += sizeof(struct crush_work_list);
```
idryomov: In the straw2 case you assign a straw2 pointer and then advance over the straw2 struct, whereas here you skip the list struct and then assign a list pointer.  I haven't followed the memory layout, so it might be OK, but it stands out.

src/crush/mapper.c
```diff
@@ -776,6 +885,69 @@ static void crush_choose_indep(const struct crush_map *map,
...
+			struct crush_work_list *lw
+				= (struct crush_work_list *) point;
+			struct crush_bucket_list* lb
+				= (struct crush_bucket_list *) m->buckets[b];
```
idryomov: Style nits: '=' should be on the previous line and the placing of asterisks is inconsistent.  I think the preferred style in this file is 'type *var' and '(type *)var' (no space between right paren and var for casts).

liewegas:
@adamemerson I think we still need a test?  I don't see anything that demonstrates the original problem (overfull low-weight items with nrep > 1) and then shows that it goes away with the new tunable.

adamemerson:
Don't know how I saw the name change and missed the testing, but I have some tests now. I have some lumpiness in how things are being picked, with it consistently underdrawing from 0. I think it's just something stupid related to underflow/overflow. Once I have that fixed and pass the conditions I can think of I'll shove again.

liewegas:
Can you elaborate on your first point, that

"We cannot adjust the weights in
subsequent draws: the ratio would approach what we expect as the
adjusted weight goes to infinity. This is terrible."

This just sounds like a special case of the general 'not all weights are possible.'  If it works for larger buckets, then giving up in a conservative way seems sufficient?

The idea of just adjusting the weights down for a given num_rep seems reasonable.  I see two issues:  (1) is that many clusters will have multiple pools, some with num_rep = 3 and some ec pools with k=11 or something else big.  (2) currently all of the crush methods and tools keep the crush bucket weights as a summing tree (bucket weight = sum of child weights), so if we tweak one of them, the next change to the tree will clobber those changes.  We'd need to keep the delta in some other persistent form, or otherwise force the value (or correction factor) to be recalculated when there is a change. :/

In any case, I think we need to discuss this more to see if we can come up with a better solution.  The first patches that get rid of the internal crush state are unrelated to this, though, so let's separate those out and get them tested?  That'll be a nice independent cleanup!

adamemerson:
So!  Regarding "We cannot adjust the weights in subsequent draws: the ratio would approach what we expect as the adjusted weight goes to infinity. This is terrible."

I spent a good bit of time trying to figure out if there was a way to make the first draw using the given weights, then during the second draw adjust the weights to preserve the distribution, do the same on the third draw, and so forth.

What I ended up finding out was that I can't. If I have the weights unmodified during the first draw then, even for possible distributions, adjusting subsequent weights won't converge well on the desired distribution, especially if I have objects of more than two weights.

If you have two weights one could increase the higher weight a whole lot on the second and subsequent draws to try and keep closer to the desired distribution, but it would effectively be not drawing low-weight objects in the second draw. Low-weight objects would have a chance of being the first draw, their chance of appearing at all in the second draw would go down radically as we approached the desired distribution.

I agree about the crush lock removal. I'll split that out and make it its own PR.

Having people use different replication counts on the same CRUSH rules is a big problem with this approach. You could replicate the bucket structure, though that's not attractive. We could give each bucket a set of 'calibrated' replication count weights based on replication count. That idea makes me very uncomfortable, since changing the replication count has not historically resulted in a bunch of data movement, while making changes in replication count de-facto changes in weighting would.

As to 2, yes. Changing weights or adding an OSD would throw off the tweak. I'm not sure how much of a problem this is. Honestly after having looked into this in detail I feel like this kind of adjustment might be more of a special case thing rather than something we would expect people to use in detail. We could, basically, just replace the addition to CRUSH tool with a little tool to help people calculate weights for a desired distribution.

I will make one change to this Matt suggested (clarifying a message) and then push a separate change to crush lock.

dachary:

dachary:
@adamemerson I'd like to help with this. What would you like me to do ?

adamemerson:
Well! The problem right now is that we've sort of hit a UI problem.

- As things are now, it surprises users because they expect the distribution they pick to apply to the sequence of draws taken together when instead if applies to every draw individually.
- There are various drawbacks to making placement depend on the replication count, not the least of which is that it harms CRUSH's monotonicity. Changing the replication count could cause much more data movement after than it did before.

dachary:
@adamemerson thanks for the quick reply ;-) I'll try to describe the problem in my own words, please correct me if I'm mistaken.

When users assigns a weight of 1 to a small device and a weight of 10 to five other devices, they expect the small device to be 10 times less likely to be used. But this is only true for the first device selected in for a three replica pool. When selecting the second device it is possible that the same device is selected again. It is then discarded and another selection is made, until a device that was is not already used is selected. And doing so increases the chances that the small device is selected.

The problem is even easier to understand when there only are three devices in the cluster. When used to support a three replica pool, they will all be equally used, regardless of their weight ;-)

jazeltq:
If we can only choose three from three, i think it is not a surprises that the distrubition is
not accord with the weight....

adamemerson:
You have the gist of it. The one difference is that selecting the device again and discarding it doesn't actually make any difference. consider that you have five OSDs. OSDs a through d have weight ten. OSD e has weight 1.

Our chance of picking OSD e (none of this is normalized) is the sum of the chance of picking it in the first draw (1/41) plus the chance of picking it in the second draw (40/41 * 1/31) which together sum to ~0.05586

Whereas the chance of picking OSD a is (again not normalized) is chance of picking it in the first draw (10/41) plus the chance of picking one of OSDs b through d in the first draw and a in the second (30/41 * 10/31) plus the chance of picking e in the first draw and a in the second draw (1/41 * 10/40) which is ~0.48602

The ratio of the likelihood of drawing a to drawing e is ~8.7. If you set up a map and run crushtool with this kind of configuration and test it, this is roughly the ratio you get. The basic problem is that CRUSH is pretty much correctly executing a draw without replacement from the supplied distribution, but that the result of a correctly executed draw without replacement is surprising and potentially not useful for a sysadmin wanting to manage stuff.

The basic problem is that the ways to fix this work, to some degree, break the monotonicity of RUSH family algorithms. This isn't necessarily a deal breaker, it's just a problem of how to present things to the user in ways that will keep people from shooting themselves or their data in the foot.

dachary:
Thanks for the detailed explanation. My probability class are a distant memory: I only remember getting bad grades ;-)

To rephrase it with my own words, when picking two devices, the chances of getting a particular device during the first draw is the weight of the device divided by the sum of the weight of all devices. In the example you propose where a = 10, b = 10, c = 10, d = 10, e = 1, the chances of getting e is 1/41 and the chances of getting a is 10/41.

Things get more complicated for the second draw because we have to account for a first draw that does not include a given device: it is the sum of the weight of all devices except the one we're interested in, divided by the weight of all devices. So, if we want to know the chances of e showing up in the second draw, the first draw must not include it and this has a 40/41 chance of happening. Also, during the second draw, the chance of getting e is increased because there is one less device to chose from (the one that was picked during the first draw): 1/31 (i.e. 41 - the weight of the device that was chosen). Because the second draw depends on the first draw, the probability must be multiplied: 40/41 * 1/31.

Since the chance of getting the device e in a first draw and getting the device e in a second draw are independent, the chances of getting the device e in both situations is the sum of their probability: 1/41 (first draw) + (40/41 * 1/31) (second draw).

This is a special case because all devices have the same weight except for e. If we are to calculate the probability of a being selected in the second draw, we have to sum the case where e is selected and the case where b, c or d is selected during the first draw, because they do not have the same weight. If e is selected during the first draw, a will be selected during the second draw with a probability of (1/41*10/40). If b, c, or d is selected during the first draw, a will be selected during the second draw with a probability of (30/41*10/31).

The chances of getting the device a in the first draw and the second draw is therefore: 10/41+(30/41*10/31)+(1/41*10/40)

To summarize:
probability of getting e : 1/41*(40/41*1/31) = .05586
probability of getting a : 10/41+(30/41*10/31)+(1/41*10/40) = .48603
it is 8.7 ( 0.48603/0.05586 ) more likely to get e than to get a.

From the point of view of the users, this is counter intuitive because they expect that the weight reflects the probability, which is only true for a single draw.

dachary:
@adamemerson To implement a function to measure the probability that an item is selected in a given bucket for a given set of weights, after N draws. Something like

    item_select(int *weights, int size, int index, int draws) => double

where index is in the range [0, size[. For instance,

    item_select([10, 10, 10],  index=0, draws=1) = 10/30
    item_select([1, 10, 10], index=0, draws=1) = 1/21
    item_select([1, 10, 10], index=0, draws=2) = 1/21 + 20/21 * 1/11
    item_select([1, 10, 10], index=1, draws=2) = 10/21 + 1/11 * 10/20 + 10/11 * 1/10
   etc.

I suspect there exists a formula to do that. I lost myself in https://en.wikipedia.org/wiki/Hypergeometric_distribution :-) Do you happen to know how to do that ?

adamemerson:
@dachary Your understanding in your own words is correct with the caveat that for a given test of n draws, nothing will be picked in more than one draw. (It's a draw without replacement.)

adamemerson:
@dachary There is no **formula** per se. There is an algorithm. If you look in CrushWrapper.cc in this PR, the multidraw function calculates exactly that. There is a good amount of low hanging fruit for optimization which I should do sometime, but it works well enough for a fairly naive numerical way of rebalancing a bucket to have (if possible) an expected distribution.

dachary:
@adamemerson thanks for [multidraw](https://github.com/ceph/ceph/pull/10218/files#diff-c1ea4e8ae4b5520080ef65c1bd8ed666R1858) !

dachary:
> with the caveat that for a given test of n draws, nothing will be picked in more than one draw. (It's a draw without replacement.)

I had that in mind. Since you mention it, it means what I wrote is incorrect. Would you be so kind as to point me where my error is exactly ? I can't see it :-D

adamemerson:
@dachary I misinterpreted

> Since the chance of getting the device e in a first draw and getting the device e in a second draw are 
> independent, the chances of getting the device e in both situations is the sum of their probability: 1/41 
> (first draw) + (40/41 * 1/31) (second draw).

As you saying you could get them in both. Technically getting e in the first draw and getting e in the second draw are not *independent* since getting e in the first draw makes getting it in the second draw impossible. That's just a terminology quibble, though.

Either way. Like I said, the only real blocker for *me* on this bug (i.e. why I stopped working on it) was that I didn't think we hit a clear consensus on the UI for how to deal with it. Based on previous conversations I think I have an idea of what a fix *should* look like if other people agree with it.

dachary:
I'm glad I was not completely mistaken (as you may have noticed, I feel quite insecure when maths are involved ;-). 

Thanks to your patience, I think I now understand the problem.

Your prefered solution is (quoting you from a previous message) "If I had to pick one I think I'd prefer to be able to install several 'rebalanced' versions of buckets in and have a CRUSH rule explicitly use them, since that seems the least likely to surprise people." ?

dachary:
If I write **e in a first draw or getting the device e in a second draw** intsead (replacing the and with or), is it less confusing ?

adamemerson:
So, the two assumptions I'm making are that the times this problem is encountered are fairly rare (you need to have very small buckets. On large buckets the difference between expected and actual distribution is fairly small.) and that we should retain monotonicity, in general. Changing the replication count should NOT induce more data movement than it does now, so I don't want to key off the replication count.

My first idea was just to call a  'rebalancer' on buckets manually and rewrite the CRUSH map (which is what I do in this PR.) People didn't like that idea partially because it made using the same bucket structure for (say) replicated and erasure coded data impossible and because it replaced the original layout so a user couldn't tweak things easily. These are reasonable complaints.

Sage suggested that we try a series of weights from 1 to n, so that the first set is unmodified, the second set of weights is rebalanced so that, given an unmodified previous map it will have the desired distribution and so forth. This is not a *bad* idea, but it has undesirable consequences. The worst is that low weight objects will, when they appear in a placement list, be *much* more likely to be the primary. I strongly suspect that being the primary is more work than being the secondary, so that kind of unevenness would be best avoided. If that actually isn't an issue and we're perfectly fine having some OSDs be the primary in almost all cases where they're used then I don't have too much objection to this idea, so long as it's not the default and instead needs to be turned on.

Otherwise my suggestion would be to, as mentioned before have a sets of shadow weights and some way to slot them in explicitly as part of the CRUSH map, whether as a rule or some property or what I'm not sure.

dachary:
> Sage suggested that we try a series of weights from 1 to n, so that the first set is unmodified, the second set of weights is rebalanced so that, given...

I don't quite understand the idea. Would you have a link to this discussion so that I don't ask you to repeat what has already been said ?

adamemerson:
I don't, it was on the mailing list at some point. Basically have the set of weights on the first draw be unmodified. On the second draw increase the weight of the high-weight items so that when we pick from it the ratio of likelihood is still 10 to 1. And increase the high weight items even more in the third draw and so forth.

The other downside is that, given that our fixed-point calculations have a limited precision and range, as the replication count increases some distributions become impossible since we'd run out of numeric range.

liewegas:
I don't understand why we would have a disproportionate number of PGs with 
primary on the low-weighted device.

Consider a cluster with 5 OSDs, 4 with weight 99 and 1 with weight 4 (for 
a total weight of 400).  2x replication, 2000 PGs, so that there are a 
total of 4000 PG copies/instances.  That means the small OSD should store 
40 PGs (1% * 4000).
  
This is 2x, so half of them should be primary and half secondary.

That would mean for the first choice, we want to choose the small OSD 
for 20 PGs out of the 2000, or 1%... which is what a weight of 4/400 gives 
you in the first pick.

For the second choice (the replica), we'll face either [99 99 99 99] (1% 
of the time) or [99 99 99 4] (99% of the time).  For the second case, 
we'll want to pick the small one 20 out of 1980 times, or 1/99=.010101 (vs 
the 4/301=.013289 we do now).

The total PG count would thus be
  2000 * (.01 + ((99/100) * (1/99))) = .02 * 2000 = 40

Unless the math somehow changes for >2x, I *think* it's just subsequent 
choices that have to have adjusted weights (depending on the previous 
choices).  This makes intutive sense.  Imagine we do 1x and distribute 
them perfectly.  We then go from 1x -> 2x, and "create" a bunch of 
replicas.  That should double the PGs per OSD, but not change any of the 
primaries, or move any previously placed data... just add new replicas.  
Which is a good form of monotonicity....

dachary:
What about adding a way to display the compared probabilities for a given bucketA and a ruleset with two replicas (i.e. what [ratio](https://github.com/adamemerson/ceph/blob/67c714f82be19723452142b5a2b867e1f19b6106/src/crush/CrushWrapper.cc#L1903) would return when comparing them ? 

    $ ceph distribution-matrix bucket A ruleset 2 # 1 replica

    osd.a weight 10
    osd.b weight 10
    osd.c weight 10
    osd.d weight 10
    osd.e weight 1

            osd.a osd.b osd.c osd.d osd.e
    osd.a     1     1     1     1    8.70 
    osd.b     1     1     1     1    8.70 
    osd.c     1     1     1     1    8.70 
    osd.d     1     1     1     1    8.70 
    osd.e    0.11  0.11  0.11  0.11   1

    $ ceph distribution-matrix bucket A ruleset 2 # 1 replica

    osd.a weight 10
    osd.b weight 10
    osd.c weight 10
    osd.d weight 10
    osd.e weight 1

            osd.a osd.b osd.c osd.d osd.e
    osd.a     1     1     1     1    10
    osd.b     1     1     1     1    10
    osd.c     1     1     1     1    10
    osd.d     1     1     1     1    10
    osd.e    0.1   0.1   0.1   0.1    1

And then a command to tweak the weight of a single OSD to match a give probability:

    $ ceph osd set-probability osd.a compared to osd.e to 10

This is very low level... maybe too low level to be useful ?

liewegas:
Yeah, I think any solution has to Just Work.  It's hard enough for us to grok the problem; we can't expect users to do it.

The big question in my mind is: how would/could we adjust the algorithm so that for subsequent picks it would use the correct probability [130.66 130.66 130.66 4] in the case above, so that the small one is 4/396 and the others are 130.66/396.  I can work backwards from the designed 1/99 to get to that point, but it's not obvious how to get there from the other direction.

dachary:
@liewegas would you mind expanding what you mean by "I can work backwards from the designed 1/99" ? Or maybe point me to the URL of a previous discussion explaining it already ?

liewegas:
In the discussion above, I knew I wanted 20 out of 1980 for the second pick, or 1/99.  If the small osd weight is 4, then that's 4/396.  If that is to come from a set of weights, the remaining 392 has to be spread across the other 3 device equally, making their weight 392/3=130.67.  So [130.67 130.67 130.67 4] would work for the second choice.  But I'm not sure how one should arrive at that point without knowing the answer first.

Also, the funny thing about this case is that if the weights were adjusted to [130.67 130.67 130.67 130.67 4] in the second round, they work regardless of whether the first pick is the small one or a big one.  It's a simplistic situation because the big ones mostly just have to be equal to each other.  In the more general case where you have, say, [10 8 6 4 2] the second round of weights (1) are less obvious to calculate and (2) I'm not certain they are independent of what was chosen for the first pick.  It's a bit weird because in the second round *all* the items are always still there... it's just that if we pick one that was already used before we retry, so its weight is effectively 0.  For our [99 99 99 99 4] example that means a single second-round weight vector of [130.67 130.67 130.67 130.67 4] will give us the right result, but is that true in the general?

liewegas:
I think this may not be too bad, actually.  Here's my hypothesis: for round N, we can use an alternate set of weights to get the right distribution for the second choice.  These weights do not depend on what the actual first choice was.  (At least, this will work for a single bucket; I'm not sure about hierarchical descent, but let's ignore that for now.)

We can use conditional probability: P(a | b) = P(a && b) / P(b) (probability of a given b is the probability of a and b / probability of b).  For some osd A,

P(first pick not A) = (total_weight - a_weight) / total_weight
P(pick A | first pick not 4) = P(pick A && first pick not A) / P(first pick not A)

The P(pick A && first pick not A) term is what we *want*.. i.e., the original weight_a/total_weight.  If the item has 1/10th the weight in the bucket, then we want to pick it 1/10th the time for the second replica.  And the P(first pick not A) is easy to calculate.  That gives

= (a_weight / total_weight) / ((total_weight - a_weight) / total_weight)
= a_weight / (total_weight - a_weight)

For the 99 99 99 99 4 example I gave before, this gives us round 2 weights of

99 / (400 - 99) = 99/301 = .3289
4 / (400 - 4) = .010101

...which are the same weights that I derived manually above.

dachary:
For the record @liewegas mentionned trying to implement this approach but it does not seem to work.

liewegas:
I had a stupid error.. this appears to be working reasonably well for the 2nd replica.  There is a bias towards high values (by about 1%) that I can't quite explain, and hopefully we can fix that, but even if we can't it's probably good enough.

Working on the math for subsequent replicas.

liewegas:
see https://github.com/liewegas/ceph/commit/wip-crush-multipick

PR #10220, opened 2016-07-09
[DNM] os/bluestore: fix a Bluestore crash

somnathr:
A bluestore race condition is been fixed by protecting txc data structures
with op sequencer lock.

### DIFF
--- a/src/os/bluestore/BlueStore.cc
+++ b/src/os/bluestore/BlueStore.cc
@@ -4589,11 +4589,19 @@ void BlueStore::_txc_state_proc(TransContext *txc)
       txc->log_state_latency(logger, l_bluestore_state_io_done_lat);
       txc->state = TransContext::STATE_KV_QUEUED;
       // FIXME: use a per-txc dirty blob list?
+      
+      if (txc->first_collection) {
+        (txc->first_collection)->lock.get_read();
+      }
       for (auto& o : txc->onodes) {
-	for (auto& p : o->blob_map.blob_map) {
-	  p.bc.finish_write(txc->seq);
+        for (auto& p : o->blob_map.blob_map) {
+	    p.bc.finish_write(txc->seq);
 	}
       }
+      if (txc->first_collection) {
+        (txc->first_collection)->lock.put_read();
+      }
+      
       if (!g_conf->bluestore_sync_transaction) {
 	if (g_conf->bluestore_sync_submit_transaction) {
 	  _txc_finalize_kv(txc, txc->t);
@@ -5160,7 +5168,6 @@ int BlueStore::queue_transactions(
   }
 
   _txc_write_nodes(txc, txc->t);
-
   // journal wal items
   if (txc->wal_txn) {
     // move releases to after wal


src/os/bluestore/BlueStore.cc
```diff
@@ -5142,6 +5142,7 @@ int BlueStore::queue_transactions(
   txc->onreadable = onreadable;
   txc->onreadable_sync = onreadable_sync;
   txc->oncommit = ondisk;
+  osr->qlock.lock();
```
liewegas: for simplicity i'd just do

 std::lock_guard<std::mutex> l<osd->qlock);

that wya it holds it through the end of hte function (including state_proc)

liewegas:
I get a deadlock on ObjectStore/StoreTest.SmallSequentialUnaligned/2

somnathr:
@liewegas we can't hold the lock for the _txc_state_proc as it will cause deadlock. it can fall through _txc_finish_io and it is trying to hold sequencer lock again..

Also, I was seeing some crash if I include throttle stuff within sequencer lock. Not sure though why ?
Will investigate the hang you are seeing...

Will investigate

somnathr:
@liewegas Yes, I am hitting the hang as well. Sorry, my bad, I forgot to run objectstoretest on this.
Here is the reason it is hanging.
Thread 1 : queue_transaction taking osr->qlock ->_write is calling onode->flush() -> waiting on flush_cond

Thread 2 : Trying to get osr->qlock from _wal_finish() and stuck. Not able to signal flush_cond from _txc_finish

I was not hitting this in my test as I was not doing any test with smaller block_size than 4K and not exercising this WAL path.

I will see how to get around this.

somnathr:
@liewegas I tried to better understood the code in past couple of days. It seems we really don't need to guard the queue_transaction code base with osr->qlock.
Guarding only _wctx_finish() with osr-qlock should be enough to prevent the crash.
No hang in ceph_objectstore test anymore. But, the following 5 tests are failed. This failure is happening in my setup even with master also.

[  FAILED  ] ObjectStore/StoreTest.TooManyBlobsTest/2, where GetParam() = "bluestore" (65415 ms)
[----------] 59 tests from ObjectStore/StoreTest (3086369 ms total)

[==========] 59 tests from 1 test case ran. (3086369 ms total)
[  PASSED  ] 54 tests.
[  FAILED  ] 5 tests, listed below:
[  FAILED  ] ObjectStore/StoreTest.BluestoreStatFSTest/2, where GetParam() = "bluestore"
[  FAILED  ] ObjectStore/StoreTest.BluestoreFragmentedBlobTest/2, where GetParam() = "bluestore"
[  FAILED  ] ObjectStore/StoreTest.Many4KWritesTest/2, where GetParam() = "bluestore"
[  FAILED  ] ObjectStore/StoreTest.Many4KWritesNoCSumTest/2, where GetParam() = "bluestore"
[  FAILED  ] ObjectStore/StoreTest.TooManyBlobsTest/2, where GetParam() = "bluestore"

somnathr:
@liewegas Still crashing after 1 hour (much less frequent but happening). It seems _wctx_fininsh() is not the only place inducing race condition. There are other places, will try to find.
Don't bother reviewing at this point.
@markhpc Don't bother testing this patch yet as you have asked me today.

somnathr:
@liewegas Crash is not reproduced with this. Taking collection lock to protect entire data structure.
@markhpc Please give it a try in your branch.

somnathr:
@markhpc It didn't work for you ?

markhpc:
It did, I merged it via a commit, see: https://github.com/ceph/ceph/commit/bf70bcb6c54e4d6404533bc91781a5ef77d62033

markhpc:
FWIW, locking the entire data strucutre may be introducing a small performance penalty, though there's enough noise that it's tough to tell.  Perhaps at some point we can do better but for now this is enough to avoid segfaults which is the first priority.

somnathr:
@markhpc It may be (for sequential) but shouldn't be a big impact.
 I am locking against collection lock , so, it will only collide with the request on that collection (PG).

PR #10229, opened 2016-07-10
makefile: change librgw_file_* as check_PROGRAMS

badone:
These targets depend on libgmock_main.la which is not built during
a default build.

Fixes: http://tracker.ceph.com/issues/16646


badone:
@tchaikov would you be able to take a look at this Kefu?

tchaikov:
lgtm.

tchaikov:
tested at http://pulpito.ceph.com/kchai-2016-07-12_00:59:27-rados:thrash-wip-kefu-testing---basic-mira/, (actually we are not using autotools anymore in gitbuilders =)

PR #10231, opened 2016-07-11
mon/PaxosService: make the return value type inconsistent

yonghengdexin735:
This is inconsistent problem for the return value type. the type for get_last_committed() and latest_full is the type of version_t, so I fixup the return type of paxos_stash_full_interval for version_t yet.

### DIFF
--- a/src/mon/OSDMonitor.cc
+++ b/src/mon/OSDMonitor.cc
@@ -1833,7 +1833,7 @@ bool OSDMonitor::prepare_failure(MonOpRequestRef op)
   if (m->if_osd_failed()) {
     // add a report
     mon->clog->debug() << m->get_target() << " reported failed by "
-		      << m->get_orig_source_inst() << "\n";
+                       << m->get_orig_source_inst() << "\n";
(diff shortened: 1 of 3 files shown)


src/mon/OSDMonitor.cc
```diff
@@ -1833,7 +1833,7 @@ bool OSDMonitor::prepare_failure(MonOpRequestRef op)
   if (m->if_osd_failed()) {
     // add a report
     mon->clog->debug() << m->get_target() << " reported failed by "
-          	       << m->get_orig_source_inst() << "\n";
+                       << m->get_orig_source_inst() << "\n";
```
tchaikov: in general, we want to avoid changes only involves formatting changes. it renders  the output of "git blame" less useful.

yonghengdexin735:
@yuriw done

yuriw:
http://pulpito.ceph.com/yuriw-2016-07-27_11:33:48-rados-wip-yuri-testing_2016_7_27-distro-basic-smithi/

PR #10233, opened 2016-07-11
rgw: fix wrong  variable definition in cls_version_check func

weiqiaomiao:


cbodley:
retest this please

PR #10210, opened 2016-07-08
osd/ReplicatedPG: fix code indentation

runsisi:
misc: fix minor code indentation issue

Signed-off-by: runsisi <runsisi@zte.com.cn>

PR #10235, opened 2016-07-11
src/librados/IoCtxImpl.cc:Adjust the alignment of function or ldout

yonghengdexin735:
I find src/librados/IoCtxImpl.cc having some place doesn't alignment for function or ldout.
 

yuyuyu101:
same reason for #10210

PR #10236, opened 2016-07-11
os/bluestore: fix potential uninitialized nid of onode

xiexingguo:
The _zero() process may implicitly create a new onode,
thus we shall call _assign_nid() to initialize the nid
properly. And if the onode already has one, _assign_nid()
does nothing.

So it is proper to call _assign_nid() here under any case.

### DIFF
--- a/src/os/bluestore/BitmapFreelistManager.cc
+++ b/src/os/bluestore/BitmapFreelistManager.cc
@@ -56,8 +56,9 @@ BitmapFreelistManager::BitmapFreelistManager(KeyValueDB *db,
 
 int BitmapFreelistManager::create(uint64_t new_size, KeyValueDB::Transaction txn)
 {
-  size = new_size;
   bytes_per_block = g_conf->bdev_block_size;
+  assert(ISP2(bytes_per_block));
+  size = P2ALIGN(new_size, bytes_per_block);
   blocks_per_key = g_conf->bluestore_freelist_blocks_per_key;
 
   _init_misc();
--- a/src/os/bluestore/BlueStore.cc
+++ b/src/os/bluestore/BlueStore.cc
@@ -3656,6 +3656,7 @@ int BlueStore::_do_read(
     } else {
       uint64_t l = length - pos;
       if (pr != pr_end) {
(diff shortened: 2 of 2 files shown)


xiexingguo:
Passed local testing: 
```
rm -rf c store_test_temp_dir && ./ceph_test_objectstore --gtest_filter=*/2
```

```
2016-07-13 16:28:02.521971 7faac0607680  1 bluestore(store_test_temp_dir) fsck finish with 0 errors
[       OK ] ObjectStore/StoreTest.TooManyBlobsTest/2 (23001 ms)
[----------] 59 tests from ObjectStore/StoreTest (2166058 ms total)

[==========] 59 tests from 1 test case ran. (2166058 ms total)
[  PASSED  ] 59 tests.

xiexingguo:
@markhpc Hi, mark. Thanks for the merging. One little suggestion: next time you'd better merge the origin pull request instead of the one pulled into your local branch. Otherwise it will be a pain if next time we want to revert part of the PR which consists of a batch of separated commits.

markhpc:
Hi Xiexingguo,

If at all possible, when you are making a PR could you only include only a single "history worthy" commit so that we can more easily squash simple commits into the history worthy one?  The reason I ask is that I try to avoid adding lots of trivial commits to the history (like slightly changing error messages etc, it makes bisection slow/annoying) while helping keep the history clean by avoiding merges:

https://blog.spreedly.com/2014/06/24/merge-pull-request-considered-harmful/

xiexingguo:
> (like slightly changing error messages etc, it makes bisection slow/annoying

Agreed.  Next time I'll try to get commits well sorted. But I still think we shall merge the origin PR if possible instead of squashing them into a very different one, and I think the well sorted labels(cleanup, bug fix, performance) and comments from the origin PR is good for tracing.

xiexingguo:
> More important (though I'm not sure yours are this way), I want to avoid individual commits that break master, since it makes bisecting super annoying.

Well, I am sure none of them will do. I've been working with Ceph for almost one and half a year and you shall have a little faith in me:-)

PR #10237, opened 2016-07-11
kraken: remove SYSTEMD_RUN from initscript

odivlad:
Closes: http://tracker.ceph.com/issues/16652

`systemd-run` logic in initscripts was introduced because of ticket http://tracker.ceph.com/issues/7627.
If we have systemd-based distro, we should use systemd unit files from systemd directory to start/stop ceph daemons.
Otherwise, `daemon()` from `/etc/init.d/functions` on systemd distro starts service in `system.slice` and everything works well for case, for example, when we use hammer on RH7. With this code it will start daemon with `daemon()` function from `init.d/functions`.
`systemd-run` can not be run on non-systemd distros, so it's not needed on SysV systems.

Backport of: #9871

Signed-off-by: Vladislav Odintsov <odivlad@gmail.com>

### DIFF
--- a/src/init-ceph.in
+++ b/src/init-ceph.in
@@ -17,17 +17,12 @@ if [ -e /lib/lsb/init-functions ]; then
     . /lib/lsb/init-functions
 fi
 
-# detect systemd, also check whether the systemd-run binary exists
-SYSTEMD_RUN=$(which systemd-run 2>/dev/null)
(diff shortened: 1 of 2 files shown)


smithfarm:
I don't think this backport is necessary. The master branch will is still being regularly merged (manually) into kraken.

liewegas:
yep

PR #10238, opened 2016-07-11
hammer: mon: fix memory leak in prepare_beacon

aiicore:
http://tracker.ceph.com/issues/17285

prepare_beacon() case of prepare_update() should put()
message in two more cases, because is the last step
of dispatch()

Signed-off-by: Igor Podoski igor.podoski@ts.fujitsu.com

### DIFF
--- a/src/mon/MDSMonitor.cc
+++ b/src/mon/MDSMonitor.cc
@@ -432,6 +432,7 @@ bool MDSMonitor::prepare_beacon(MMDSBeacon *m)
   // Ignore beacons if filesystem is disabled
   if (!mdsmap.get_enabled()) {
     dout(1) << "warning, MDS " << m->get_orig_source_inst() << " up but filesystem disabled" << dendl;
+    m->put();
     return false;
   }
 
@@ -501,6 +502,7 @@ bool MDSMonitor::prepare_beacon(MMDSBeacon *m)
       dout(0) << "got beacon for MDS in STATE_STOPPING, ignoring requested state change"
 	       << dendl;
       _note_beacon(m);
+      m->put();
       return true;
     }
 


src/mon/MDSMonitor.cc
```diff
@@ -501,6 +502,7 @@ bool MDSMonitor::prepare_beacon(MMDSBeacon *m)
       dout(0) << "got beacon for MDS in STATE_STOPPING, ignoring requested state change"
 	       << dendl;
       _note_beacon(m);
+      m->put();
```
aiicore: Not sure if this put should be here when return 'true', because we can go into propose_pending() from dispatch() in mon/PaxosService.cc. I'm not so familiar with MDS monitor code.

aiicore:
Memory leaks every time when you see in monitor log:
`mon.0@0(leader).mds e1 warning, MDS mds.? XXX.XXX.XXX.XXX:6800/1494583 up but filesystem disabled`
Is not very noticeable, however when cluster is in unstable state, e.g. network cards are dropping random packets, monitors tend to fall into election loop. When two monitors are marking themselves as leaders for a while and then start election from the beginning, MDS messages are forwarded from one to another (to leader). After a while amount of MDS messages is quite big and when they hit monitor which is leader (at that point of time) memory leak is very visible and it leads to OOM kill.

smithfarm:
Adding DNM just to keep this very new PR out of the next hammer integration testing round. Will remove it afterwards.

smithfarm:
@aiicore Is this bug present in infernalis, jewel? Do you know which commit fixed it? When backporting, we prefer to cherry-pick existing fixes.

aiicore:
@smithfarm This code lasts until v10.0.5. There was big MDSMonitor reorganization in 4e9b953, and above code was removed.

smithfarm:
@aiicore OK - is there an existing tracker issue? Or can you make a new one?

aiicore:
@smithfarm I think there is no related tracker issue. I'll make one soon.

aiicore:
@smithfarm I've created bug report http://tracker.ceph.com/issues/17285 and assigned to myself, but I can't edit it anymore (wanted to change it to FIX and clear a typo). Am I missing permissions?

smithfarm:
@aiicore Don't know about your tracker permissions, but I fixed up the issue so it will work with our backporting scripts and saved queries.

aiicore:
@smithfarm Thanks!

smithfarm:
@aiicore Could you please rebase.

smithfarm:
@aiicore Also, please add a `Fixes: http://tracker.ceph.com/issues/17285` line right above `Signed-off-by` thanks.

aiicore:
@smithfarm rebased + additional line in commit

smithfarm:
@jcsp @tchaikov This hammer-only fix passed a rados suite at http://tracker.ceph.com/issues/17151#note-7 with failures that I believe have been addressed (except for http://tracker.ceph.com/issues/15139 which is caused by the build system no longer providing dumpling-era packages).

I have rebuilt the integration branch to include the two fixes and scheduled a new run at http://tracker.ceph.com/issues/17151#note-14

Do you think it's OK to merge provided the second run succeeds?

smithfarm:
Second run succeeded; third run here: http://pulpito.front.sepia.ceph.com/smithfarm-2016-11-21_11:52:41-rados-hammer-backports---basic-smithi/

src/mon/MDSMonitor.cc
tchaikov: `prepare_update()` is responsible to free the message before returns. so this change is correct.

tchaikov:
lgtm.

jcsp:
Looks sensible to me

PR #10241, opened 2016-07-11
rgw: add return value checking to avoid possible subsequent `parse` exception

Yan-waller:
add return value checking to avoid possible subsequent `parse` exception


theanalyst:
the first commit looks good. 

I'm not very keen on the tab vs space change.. while I agree we should fix indentation, this does create problems when doing a git blame or when backporting commits... it is fine doing a ws change when you're kind of modifying the code around, lets see what others think of this?

Yan-waller:
@theanalyst thanks for reminding me of the problems when doing a git blame or when backporting commits, that sounds reasonable. I drop that commit, and fix some necessary indents. Could you help to take a look again?

theanalyst:
lgtm

PR #10242, opened 2016-07-11
bluestore_extent_ref_map_t::put should use vector<bluestore_lextent_t>

sysnote:
bluestore_extent_ref_map_t::put should use vector<bluestore_lextent_t>

Signed-off-by: Wu Dong archer.wudong@gmail.com

### DIFF
--- a/src/os/bluestore/bluestore_types.cc
+++ b/src/os/bluestore/bluestore_types.cc
@@ -210,7 +210,7 @@ void bluestore_extent_ref_map_t::get(uint32_t offset, uint32_t length)
 
 void bluestore_extent_ref_map_t::put(
   uint32_t offset, uint32_t length,
-  vector<bluestore_pextent_t> *release)
+  vector<bluestore_lextent_t> *release)
 {
   map<uint32_t,record_t>::iterator p = ref_map.lower_bound(offset);
   if (p == ref_map.end() || p->first > offset) {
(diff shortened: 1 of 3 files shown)


ifed01:
IMHO this PR is incorrect. bluestore_extent_ref_map_t::put intentionally returns vector of pextents, i.e. PHYSICAL extents that need to be released after specific LOGICAL extent release.
Hence suggest to close

liewegas:
Yep!

sysnote:
@ifed01 in struct bluestore_extent_ref_map_t: map<uint32_t, record_t> ref_map,  the key means offset in the blob, not physical offset, and in function bluestore_extent_ref_map_t::put, it use 'bluestore_pextent_t(p->first, length) to construct pextent', but pextent' offset is physical offset, not ref_map's logical offset.

sysnote:
@ifed01 bluestore_extent_ref_map_t::put is used in bluestore_blot_t::put_ref, this line "ref_map.put(offset, length, &logical)", but here the vector<bluestore_pextent_t> logical just acts as the lextent, so i think this will mislead others.

ifed01:
@sysnote: strictly speaking the container returned from ref_map::put tracks neither logical nor physical adresses (where logical addressing is within object content and physical - for underlying BlueFS). In this case we use 'in-blob' address that we don't have specific structure for. Hence we use vector of bluestore_pextent_t entries as it similar to bluestore_blob_t::put_ref. IMO bluestore_lextent_t is a bit worse for that purpose since it contains additional blob identifier member.

PR #10245, opened 2016-07-11
mds: add health warning for oversized cache

jcsp:
Various issues can lead to an oversized cache, which
is a precursor to an OOM condition: let's give users
a clearer message with some useful counts that might
hint at what is wrong.

Fixes: http://tracker.ceph.com/issues/16570


ukernel:
LGTM

jcsp:
Tested here http://pulpito.ceph.com/jspray-2016-07-11_03:45:31-fs-wip-mds-minor-distro-basic-mira/

PR #10248, opened 2016-07-11
doc: rm "type=rpm-md" from yum repositories

ktdreyer:
As far as I can tell, this is only relevant to SUSE's Zypper, and yum does not need it.


alfredodeza:
this looks OK to me

alfredodeza:
@ktdreyer can you sign your commits please? This will probably need a rebase too

ktdreyer:
Sorry about that. Rebased and re-pushed.

PR #10250, opened 2016-07-11
mds: Return "committing" rather than "committed" member in get_committing

gregsfortytwo:


gregsfortytwo:
Reported at https://github.com/ceph/ceph/commit/9cb5742a61c9d872d6cb7e0962f3623de66d41d0#commitcomment-18109087, checked by inspection.

Luckily this is only used in debug output so far, so I don't think there are going to be any bugs from it.

gregsfortytwo:
I marked this as needs-qa since I haven't even built it. Presumably I got it right and this can just get a pro-forma run through on a testing branch!

PR #10254, opened 2016-07-12
rbd-mirror: include local pool id in resync throttle unique key

dillaman:
Fixes: http://tracker.ceph.com/issues/16536

### DIFF
--- a/src/tools/rbd_mirror/ImageSyncThrottler.cc
+++ b/src/tools/rbd_mirror/ImageSyncThrottler.cc
@@ -49,24 +49,26 @@ ImageSyncThrottler<I>::~ImageSyncThrottler() {
 }
 
 template <typename I>
-void ImageSyncThrottler<I>::start_sync(
-                            I *local_image_ctx, I *remote_image_ctx,
-                            SafeTimer *timer, Mutex *timer_lock,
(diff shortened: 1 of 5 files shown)


rjfd:
lgtm

src/tools/rbd_mirror/ImageSyncThrottler.cc
```diff
@@ -88,33 +90,35 @@ void ImageSyncThrottler<I>::start_sync(
...
 
   if (start) {
-      sync_holder_ctx->m_sync->send();
+    sync_holder_ctx->m_sync->send();
   }
 }
 
 template <typename I>
-void ImageSyncThrottler<I>::cancel_sync(const std::string& local_image_id) {
+void ImageSyncThrottler<I>::cancel_sync(librados::IoCtx &local_io_ctx,
```
rjfd: Why not just pass the uint64_t pool id instead of the IoCtx object?

dillaman: Just to abstract away how the key is formed -- I originally wanted to change the signature to pass in ImageCtxT but that would have involved more work in bootstrap to only cancel if an image sync was scheduled.

trociny:
lgtm

PR #10258, opened 2016-07-12
ceph-fuse: link to libtcmalloc or jemalloc

ukernel:
Fixes: http://tracker.ceph.com/issues/16655

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1163,7 +1163,7 @@ if(WITH_LIBCEPHFS)
       ceph_fuse.cc
       client/fuse_ll.cc)
     add_executable(ceph-fuse ${ceph_fuse_srcs})
-    target_link_libraries(ceph-fuse ${FUSE_LIBRARIES} client global)
+    target_link_libraries(ceph-fuse ${ALLOC_LIBS} ${FUSE_LIBRARIES} client global)
     set_target_properties(ceph-fuse PROPERTIES COMPILE_FLAGS "-I${FUSE_INCLUDE_DIRS}")
     install(TARGETS ceph-fuse DESTINATION bin)
     install(PROGRAMS mount.fuse.ceph DESTINATION ${CMAKE_INSTALL_SBINDIR})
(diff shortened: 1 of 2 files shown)


tchaikov:
shall we backport this fix to jewel? as master is not using autotools anymore.

ukernel:
you mean master is using cmake now? recently I tried using cmake to compile the master several times, most attempts failed.

tchaikov:
both gitbuilders and jenkins "make check" job are using cmake now. if cmake does not work for you, maybe it's the acconfig.h created by `configure`?

ktdreyer:
@tchaikov any objections to merging this?

tchaikov:
@ktdreyer no, not at all.

tchaikov:
but shall we have a more general way to link against the allocator libraries?

wjwithagen:
On 14-7-2016 11:21, Kefu Chai wrote:
> but shall we have a more general way to link against the allocator
> libraries?

My cmake wip (#10024) contains a piece of code where it actually
excludes any allocator libs, since FreeBSD already use jemalloc as
native allocator.

jcsp:
Checked this didn't break anything in http://pulpito.ceph.com/jspray-2016-07-13_08:15:02-fs-wip-jcsp-testing-20160713-distro-basic-mira/

@tchaikov @ktdreyer feel free to merge if this is the needed change for jewel (I held off because the last couple of comments made me uncertain)

tchaikov:
thanks for your input, @wjwithagen.  i am merging this change as it is. let's figure out a better way to address it later on.

tchaikov:
thanks for your input, @wjwithagen.  i am merging this change as it is. let's figure out a better way to address it later on.

wjwithagen:
On 14-7-2016 14:28, Kefu Chai wrote:
> thanks for your input, @wjwithagen <https://github.com/wjwithagen>. i am
> merging this change as it is. let's figure out a better way to address
> it later on.

As long as `${ALLOC_LIBS}` is used it'll keep working as planned.
Because that is the definition I wrapped, and stays empty on FreeBSD.
So as long as this VAR only uses the type of allocator lib, and no more
things will work.

tchaikov:
okay, actually i don't really like how `$ALLOC_LIBS` is initialized, we should set it to variabls like `${Tcmalloc_LIBRARIES}`, but that's another story.

PR #10259, opened 2016-07-12
os: move api

manali26:


src/test/objectstore/store_test.cc
```diff
@@ -2508,6 +2508,69 @@ TEST_P(StoreTest, SimpleCloneTest) {
   }
 }
 
+TEST_P(StoreTest, SimpleMoveTempTest) {
+  ObjectStore::Sequencer osr("test");
+  int r;
+  coll_t cid, cid2;
+ if ((GetParam() == string("filestore")) ) {
```
athanatos: Inconsistent spacing

src/os/kstore/KStore.cc
```diff
@@ -3478,6 +3488,44 @@ int KStore::_clone_range(TransContext *txc,
...
+			 OnodeRef& baseo,
+			 vector<boost::tuple<uint64_t, uint64_t, uint64_t>> move_info)
+{
+  int r = 0;
+  bufferlist bl;
+  baseo->exists = true;
+  _assign_nid(txc, baseo);
+
+//for loop
+ for (unsigned i = 0; i < move_info.size(); ++i) {
```
athanatos: Incorrect spacing

src/os/kstore/KStore.cc
```diff
@@ -2669,6 +2669,16 @@ void KStore::_txc_add_transaction(TransContext *txc, Transaction *t)
       }
       break;
 
+    case Transaction::OP_MERGE_DELETE:
+      {
+	const ghobject_t& boid = i.get_oid(op->dest_oid);
+	OnodeRef bo = c->get_onode(boid, true);
+  vector<boost::tuple<uint64_t, uint64_t, uint64_t>> move_info;
```
athanatos: Incorrect spacing

src/os/filestore/FileStore.cc
```diff
@@ -3736,6 +3753,65 @@ int FileStore::_clone_range(const coll_t& cid, const ghobject_t& oldoid, const g
...
+         goto out3;
+      }
+  }
+
+ dout(10) << "move_tempobjs "  << cid << "/" << oid << " "  <<  " = " << r << dendl;
+
+  out3:
+    lfn_close(b);
+
+  out2:
```
athanatos: Incorrect spacing

src/os/filestore/FileStore.cc
```diff
@@ -3736,6 +3753,65 @@ int FileStore::_clone_range(const coll_t& cid, const ghobject_t& oldoid, const g
...
+    goto out;
+  }
+
+  r = lfn_open(cid, oid, true, &b);
+  if (r < 0) {
+    err = true;
+    goto out2;
+  }
+
+  for (unsigned i = 0; i < move_info.size(); ++i) {
```
athanatos: Incorrect spacing

src/os/filestore/FileStore.cc
```diff
@@ -3736,6 +3753,65 @@ int FileStore::_clone_range(const coll_t& cid, const ghobject_t& oldoid, const g
...
+                              const SequencerPosition& spos)
+{
+//after all processing is done, delete the tempobj.
+  int r = 0;
+  bool err = false;
+  FDRef t, b;
+
+  dout(10) << "move_tempobjs " << temp_cid << "/" << temp_oid << " -> " << cid << "/" << oid << dendl;
+
+	r = lfn_open(temp_cid, temp_oid, false, &t);
```
athanatos: Incorrect spacing.

src/os/filestore/FileStore.cc
```diff
@@ -2688,6 +2689,22 @@ void FileStore::_do_transaction(
       }
       break;
 
+    case Transaction::OP_MERGE_DELETE:
+      {
+        coll_t temp_cid = i.get_cid(op->cid);
```
athanatos: Incorrect spacing

src/os/ObjectStore.h
```diff
@@ -1358,6 +1363,36 @@ class ObjectStore {
...
+
+   /**
+    * Move temporary object to base object.
+    * Data portion is only copied from temp object to base object.
+    * The copy is done according to the move_info vector of tuple, which
+    * has information of src_offset, dest_offset and length.
+    * Finally, the temporary object is deleted.
+    */
+    void move_tempobjs(const coll_t& temp_cid, const ghobject_t& temp_oid, coll_t cid,
+										   ghobject_t oid, const vector<boost::tuple<uint64_t, uint64_t, uint64_t>> move_info) {
```
athanatos: Spacing.

src/os/ObjectStore.h
```diff
@@ -1358,6 +1363,36 @@ class ObjectStore {
...
     }
+
+   /**
+    * Move temporary object to base object.
+    * Data portion is only copied from temp object to base object.
+    * The copy is done according to the move_info vector of tuple, which
+    * has information of src_offset, dest_offset and length.
+    * Finally, the temporary object is deleted.
+    */
+    void move_tempobjs(const coll_t& temp_cid, const ghobject_t& temp_oid, coll_t cid,
```
athanatos: Again, it may not be a temp object -- temp_cid might be the same as cid.  Better to name them source_cid/source_oid.  Also, why is temp_cid passed as a const & while cid is passed by value?  Pass both by const &.

src/os/Transaction.cc
```diff
@@ -232,6 +232,25 @@ void ObjectStore::Transaction::_build_actions_from_tbl()
       }
       break;
 
+    case Transaction::OP_MERGE_DELETE:
+      {
+	coll_t temp_cid;
```
athanatos: source

athanatos: Actually, this shouldn't be encoding the source_cid separately at all.  The source object must be in the same collection (ObjectStore level).  The *filestore* might use the temp version of the collection instead, but that depends on the ghobject_t passed, not on the collection.

src/os/ObjectStore.h
athanatos: Also, we don't need to pass source/temp_cid here at all.  At the ObjectStore level, you can't do a move between collections anyway (the filestore might kludge the source or dest collection to the temp collection if it needs to).

src/os/ObjectStore.h
```diff
@@ -1358,6 +1363,36 @@ class ObjectStore {
       }
       data.ops++;
     }
+
+   /**
+    * Move temporary object to base object.
+    * Data portion is only copied from temp object to base object.
+    * The copy is done according to the move_info vector of tuple, which
+    * has information of src_offset, dest_offset and length.
+    * Finally, the temporary object is deleted.
```
manali26: ok

src/test/objectstore/store_test.cc
manali26: ok

src/os/kstore/KStore.cc
manali26: ok

src/os/filestore/FileStore.cc
manali26: ok

src/os/filestore/FileStore.cc
manali26: ok

src/os/kstore/KStore.cc
manali26: ok

src/os/filestore/FileStore.cc
manali26: ok

src/os/filestore/FileStore.cc
manali26: ok

src/os/ObjectStore.h
manali26: ok

manali26:
I have addressed all comments. I will squash multiple commits into one after the review is done.

src/os/ObjectStore.h
```diff
@@ -1364,34 +1364,33 @@ class ObjectStore {
...
-      data.ops++;
-    }
+  /*
+   * Move source object to base object.
+   * Data portion is only copied from source object to base object.
+   * The copy is done according to the move_info vector of tuple, which
+   * has information of src_offset, dest_offset and length.
+   * Finally, the source object is deleted.
+   */
+  void move_tempobjs(const ghobject_t& src_oid, ghobject_t oid,
```
athanatos: cid generally goes before src_oid and oid.

src/os/filestore/FileStore.cc
```diff
@@ -3805,10 +3805,10 @@ int FileStore::_move_tempobjs(const coll_t& temp_cid, const ghobject_t& temp_oid
     if (err) { return r; }
 
   /* delete the temp obj */
-  lfn_unlink(temp_cid, temp_oid, spos, true);
+  lfn_unlink(src_cid, src_oid, spos, true);
 
   out:
-  dout(10) << "move_tempobjs "  << cid << "/" << oid << " "  <<  " = " << r << dendl;
+    dout(10) << "move_tempobjs "  << cid << "/" << oid << " "  <<  " = " << r << dendl;
```
athanatos: Incorrect whitespace.

src/test/objectstore/store_test.cc
```diff
@@ -2512,7 +2512,7 @@ TEST_P(StoreTest, SimpleMoveTempTest) {
   ObjectStore::Sequencer osr("test");
   int r;
   coll_t cid, cid2;
- if ((GetParam() == string("filestore")) ) {
+  if ((GetParam() == string("filestore")) ) {
```
athanatos: This will need to be implemented for all of the backends.

src/os/filestore/FileStore.cc
```diff
@@ -3780,22 +3780,22 @@ int FileStore::_move_tempobjs(const coll_t& temp_cid, const ghobject_t& temp_oid
   }
```
athanatos: What happens if this operation gets replayed?

src/os/filestore/FileStore.cc
```diff
@@ -3756,7 +3756,7 @@ int FileStore::_clone_range(const coll_t& cid, const ghobject_t& oldoid, const g
 /*
  * Move contents of temp object according to move_info to base object. Once the move_info is traversed completely, delete the temp object.
  */
-int FileStore::_move_tempobjs(const coll_t& temp_cid, const ghobject_t& temp_oid, const coll_t& cid, const ghobject_t& oid,
+int FileStore::_move_tempobjs(const coll_t& src_cid, const ghobject_t& src_oid, const coll_t& cid, const ghobject_t& oid,
```
athanatos: What's the other oid for?

src/os/filestore/FileStore.cc
manali26: ok

src/test/objectstore/store_test.cc
manali26: true. currently only kstore and filestore implementations are present. so i have added that condition. added kstore param here. after bluestore and other implementations, we can remove this condition altogether.

src/os/filestore/FileStore.cc
manali26: added guards to ensure nothing wrong happens

src/os/filestore/FileStore.cc
manali26: one is tempobj [src_oid] and other is base object [oid]. we are moving src_oid --> oid and then deleting the src_oid.

src/os/ObjectStore.h
manali26: ok

src/os/ObjectStore.h
```diff
@@ -1371,8 +1371,7 @@ class ObjectStore {
    * has information of src_offset, dest_offset and length.
    * Finally, the source object is deleted.
    */
-  void move_tempobjs(const ghobject_t& src_oid, ghobject_t oid,
-                     const coll_t& cid,
+  void merge_delete_srcobj(const coll_t& cid, const ghobject_t& src_oid, ghobject_t oid,
```
athanatos: Let's call it move_ranges_destroy_src.  Sorry, I should have been more specific.

src/os/filestore/FileStore.cc
```diff
@@ -3754,18 +3754,22 @@ int FileStore::_clone_range(const coll_t& cid, const ghobject_t& oldoid, const g
...
  */
-int FileStore::_move_tempobjs(const coll_t& src_cid, const ghobject_t& src_oid, const coll_t& cid, const ghobject_t& oid,
+int FileStore::_merge_delete_srcobj(const coll_t& src_cid, const ghobject_t& src_oid, const coll_t& cid, const ghobject_t& oid,
                               const vector<boost::tuple<uint64_t, uint64_t, uint64_t>> move_info,
                               const SequencerPosition& spos)
 {
-//after all processing is done, delete the tempobj.
   int r = 0;
   bool err = false;
+  int dstcmp;
```
athanatos: This isn't ansi C, you don't need to declare variables at the top of the block.  We actively do not want to do this, what if someone tries to use dstcmp before it's initialized?  Please move the declaration to the initialization when it's an option.

src/os/filestore/FileStore.cc
```diff
@@ -3795,20 +3799,24 @@ int FileStore::_move_tempobjs(const coll_t& src_cid, const ghobject_t& src_oid,
...
+    }
     lfn_close(b);
-
   out2:
     lfn_close(t);
     if (err) { return r; }
 
-  /* delete the temp obj */
-  lfn_unlink(src_cid, src_oid, spos, true);
+  if (_check_replay_guard(src_cid, src_oid, spos) > 0) {
```
athanatos: Why is this at a different level of indentation?

src/os/filestore/FileStore.cc
```diff
@@ -3795,20 +3799,24 @@ int FileStore::_move_tempobjs(const coll_t& src_cid, const ghobject_t& src_oid,
       }
   }
 
-  dout(10) << "move_tempobjs "  << cid << "/" << oid << " "  <<  " = " << r << dendl;
+  dout(10) << "merge_delete_srcobj "  << cid << "/" << oid << " "  <<  " = " << r << dendl;
 
   out3:
+    if (dstcmp > 0) {      // if dstcmp == 0 the guard already says "in-progress"
+      _set_replay_guard(**b, spos, &oid);
```
athanatos: You set the replay guard before removing the object here.

src/os/filestore/FileStore.cc
```diff
@@ -3754,18 +3754,22 @@ int FileStore::_clone_range(const coll_t& cid, const ghobject_t& oldoid, const g
...
   int r = 0;
   bool err = false;
+  int dstcmp;
   FDRef t, b;
 
-  dout(10) << "move_tempobjs " << src_cid << "/" << src_oid << " -> " << cid << "/" << oid << dendl;
+  dout(10) << "merge_delete_srcobj " << src_cid << "/" << src_oid << " -> " << cid << "/" << oid << dendl;
+
+  dstcmp = _check_replay_guard(cid, oid, spos);
+  if (dstcmp < 0)
```
athanatos: Is it possible for dstcmp here to be 0?  Can you be sure the source was removed if dstcmp is -1?

src/os/filestore/FileStore.cc
athanatos: Does this usage of _set_replay_guard allow _check_replay_guard to return 0?  Does this design ensure that the source object is actually removed given that you set the replay guard before removing the object?

src/os/filestore/FileStore.cc
athanatos: I'm not convinced that this is correct.  I'd like you to add a comment explaining how you are using the guards and why you believe you have ensured correct replay.  This stuff is really easy to get wrong and leads to really hard to track down data corruption.  Adding an explanation of how replay is supposed to work for this op will help people making changes to this code in the future to understand what is going on.

src/os/filestore/FileStore.cc
```diff
@@ -3754,18 +3754,22 @@ int FileStore::_clone_range(const coll_t& cid, const ghobject_t& oldoid, const g
...
+ * Move contents of src object according to move_info to base object. Once the move_info is traversed completely, delete the src object.
  */
-int FileStore::_move_tempobjs(const coll_t& src_cid, const ghobject_t& src_oid, const coll_t& cid, const ghobject_t& oid,
+int FileStore::_merge_delete_srcobj(const coll_t& src_cid, const ghobject_t& src_oid, const coll_t& cid, const ghobject_t& oid,
                               const vector<boost::tuple<uint64_t, uint64_t, uint64_t>> move_info,
                               const SequencerPosition& spos)
 {
-//after all processing is done, delete the tempobj.
   int r = 0;
   bool err = false;
```
athanatos: Why use a second error value other than r?  This method can only return 0 or -error, right?

src/os/ObjectStore.h
manali26: ok

src/os/filestore/FileStore.cc
manali26: ok

src/os/filestore/FileStore.cc
manali26: the method only returns value of r. 
'err' is used to exit properly. 

there are various possible places when we would like to error out. and sometimes we do not want to error out.
for example:

open(srcoid) failed --> return r
open(srcoid) passed but open(oid) failed --> close(srcoid) and return
open(srcoid) and open(oid) passed,  do_clone_range failed --> close both srcoid, oid, but do not delete srcoid, just return r.
all passed- > close both oids, delete srcoid return r.

src/os/filestore/FileStore.cc
manali26: changed the logic. made it simple and added comments.
check_replay_guard(base object)
   do the cloning
   check replay guard for src obj
         delete the src.
set replay guard (base coll_t)

src/os/filestore/FileStore.cc
```diff
@@ -3731,11 +3750,84 @@ int FileStore::_clone_range(const coll_t& cid, const ghobject_t& oldoid, const g
...
+     srcoff = move_info[i].get<0>();
+     dstoff = move_info[i].get<1>();
+     len = move_info[i].get<2>();
+
+     r = _do_clone_range(**t, **b, srcoff, len, dstoff);
+  }
+
+  dout(10) << __func__  << cid << "/" << oid << " "  <<  " = " << r << dendl;
+
+    lfn_close(b);
```
athanatos: Why is this indented differently?

src/os/filestore/FileStore.cc
```diff
@@ -3731,11 +3750,84 @@ int FileStore::_clone_range(const coll_t& cid, const ghobject_t& oldoid, const g
...
+ * Move contents of src object according to move_info to base object. Once the move_info is traversed completely, delete the src object.
+ */
+int FileStore::_move_ranges_destroy_src(const coll_t& src_cid, const ghobject_t& src_oid, const coll_t& cid, const ghobject_t& oid,
+                              const vector<boost::tuple<uint64_t, uint64_t, uint64_t>> move_info,
+                              const SequencerPosition& spos)
+{
+  int r = 0;
+  bool err = false;
+  FDRef t, b;
+  int dstcmp, srccmp;
```
athanatos: Don't declare these at the top of the block.  Declare them where they are initialized.

src/os/filestore/FileStore.cc
```diff
@@ -3731,11 +3750,84 @@ int FileStore::_clone_range(const coll_t& cid, const ghobject_t& oldoid, const g
...
+    return 0;
+
+  // check the src name too; it might have a newer guard, and we don't
+  // want to clobber it
+  srccmp = _check_replay_guard(src_cid, src_oid, spos);
+  if (srccmp < 0)
+    return 0;
+
+  r = lfn_open(src_cid, src_oid, false, &t);
+  //If we are replaying, it is possible that we do not find src obj as it is deleted before crashing.
```
athanatos: You need to set the replay guard on the destination before returning here!

src/os/filestore/FileStore.cc
```diff
@@ -3731,11 +3750,84 @@ int FileStore::_clone_range(const coll_t& cid, const ghobject_t& oldoid, const g
...
+    r = lfn_unlink(src_cid, src_oid, spos, true);
+    // If crash occurs between unlink and set guard, correct the error.
+    // as during next time, it might not find the already deleted object.
+    if (r < 0 && replaying) {
+      r = 0;
+    }
+  }
+
+  if (r == 0)
+    _set_replay_guard(cid, spos, false); //set replay guard for base obj coll_t.
```
athanatos: You are checking the replay guard on the object, so you need to set the replay guard on the object.

src/os/filestore/FileStore.cc
```diff
@@ -3731,11 +3750,84 @@ int FileStore::_clone_range(const coll_t& cid, const ghobject_t& oldoid, const g
...
+  }
+
+  r = lfn_open(cid, oid, true, &b);
+  if (r < 0) {
+    lfn_close(t);
+    return 0;
+  }
+
+  for (unsigned i = 0; i < move_info.size(); ++i) {
+     uint64_t srcoff;
```
manali26: ok

src/os/filestore/FileStore.cc
manali26: changed the indentation

src/os/filestore/FileStore.cc
manali26: ok

src/os/filestore/FileStore.cc
manali26: ok. corrected.

src/os/filestore/FileStore.cc
manali26: I have added a comment to reflect it. 
scenario can happen like this:
move api starts executing.
         it clones from src obj to dest obj.
         deletes src obj
         and then crashes.
when we restart, we start replaying, and find that src obj is not present [as we have deleted above].
in this case, we just reply back without doing anything. 
I can add an assert saying this is normal behavior if we are replaying else assert.

so i dont think we need to set a guard here.

athanatos: You do, because a subsequent operation might recreate the source object before another crash.

create source
move source destination
delete source
-crash-
move source destination (failed, source does not exist)
recreate source
-crash-
move source destination (suceeded, but causes incorrect contents since the second version of source is different)
remove source (this is actually ok since it's going to be recreated later)
set replay guard

You need to set the replay guard if you are replaying and the source does not exist.

manali26: ok.

src/test/objectstore/DeterministicOpSequence.cc
```diff
@@ -526,6 +561,23 @@ void DeterministicOpSequence::_do_write_and_clone_range(coll_t coll,
...
+                                                        hobject_t& new_obj,
+                                                        vector<boost::tuple<uint64_t, uint64_t, uint64_t>> move_info,
+                                                        bufferlist& bl)
+{
+  ObjectStore::Transaction t;
+
+  note_txn(&t);
+  for (unsigned i = 0; i < move_info.size(); ++i) {
+    uint64_t srcoff = move_info[i].get<0>();
+    t.write(coll, ghobject_t(orig_obj), srcoff, bl.length(), bl);
```
athanatos: Hmm, I'd like to see this happen in another transaction some of the time.

manali26: You mean , write and move api happening separately some times ? 
I have followed how clone and other apis work here.
So, we can probably keep it as is for now.

src/os/kstore/KStore.cc
```diff
@@ -2669,6 +2669,16 @@ void KStore::_txc_add_transaction(TransContext *txc, Transaction *t)
       }
       break;
 
+    case Transaction::OP_MERGE_DELETE:
+      {
+  const ghobject_t& boid = i.get_oid(op->dest_oid);
```
athanatos: Spacing is wrong here.

src/os/memstore/MemStore.cc
```diff
@@ -795,6 +795,12 @@ void MemStore::_do_transaction(Transaction& t)
       }
       break;
 
+    case Transaction::OP_MERGE_DELETE:
+      {
+       assert(0 == "not implemented");
```
athanatos: This does need to be implemented.  We'll want to be able to test ec overwrites in memory...

src/test/objectstore/store_test.cc
```diff
@@ -2508,6 +2508,70 @@ TEST_P(StoreTest, SimpleCloneTest) {
   }
 }
 
+TEST_P(StoreTest, SimpleMoveRangeDelSrcTest) {
+  ObjectStore::Sequencer osr("test");
+  int r;
+  coll_t cid;
+  if ((GetParam() == string("filestore")) || (GetParam() == string("kstore"))) {
```
athanatos: Like I said, we need to implement this for all of the backends.

src/os/bluestore/BlueStore.cc
```diff
@@ -5407,6 +5407,10 @@ void BlueStore::_txc_add_transaction(TransContext *txc, Transaction *t)
       }
       break;
 
+    case Transaction::OP_MERGE_DELETE:
+      assert(0 == "not implemented");
```
athanatos: Nope, we need it implemented.  We *might* make an exception for bluestore only if you talk to the people currently working on it and they don't think now is a good time.

liewegas:
Please wait a few more days for the latest bluestore extent map changes to merge and then rebase this on top.  Most of the bluestore code here will need a rewrite.  Thanks!

manali26:
Hi Sage,

Although it will require a rewrite, can we go along with this for time being ?
And rewrite it when things get changed in bluestore ?
Currently, move api uses clone_range directly.

Thanks!

liewegas:
Oh, sure.

src/os/bluestore/BlueStore.cc
```diff
@@ -5408,7 +5408,16 @@ void BlueStore::_txc_add_transaction(TransContext *txc, Transaction *t)
       break;
 
     case Transaction::OP_MERGE_DELETE:
-      assert(0 == "not implemented");
+      {
+	const ghobject_t& noid = i.get_oid(op->dest_oid);
+	OnodeRef& no = ovec[op->dest_oid];
+	if (!no) {
+	  no = c->get_onode(noid, true);
```
chhabaramesh: is this case when destination object does not exist? is this case possible?

src/os/bluestore/BlueStore.cc
```diff
@@ -5408,7 +5408,16 @@ void BlueStore::_txc_add_transaction(TransContext *txc, Transaction *t)
...
+	}
+  vector<boost::tuple<uint64_t, uint64_t, uint64_t>> move_info;
+  i.decode_move_info(move_info);
```
chhabaramesh: Please fix indentation

src/os/bluestore/BlueStore.cc
athanatos: Yeah

src/os/bluestore/BlueStore.cc
manali26: already fixed in next updated diff.

athanatos:
Clean up the commit history.

src/os/bluestore/BlueStore.cc
manali26: Sam,

If base object is deleted before move transaction by some one, do we continue with move api by creating a new base object or would we like to return without doing anything/or deleting the temp obj?
here, we are creating a new base obj and continuing with move api.

Also, when the move api is called the first time, has the ECbackend code created base object ?
or will move api create it after finding out that its not present?

athanatos: By base object, do you mean the source, or the destination?

src/os/kstore/KStore.cc
athanatos: Spacing is still wrong here.

src/os/kstore/KStore.cc
```diff
@@ -3169,6 +3179,45 @@ int KStore::_clone_range(TransContext *txc,
...
+                           CollectionRef& basec,
+                           OnodeRef& baseo,
+                           vector<boost::tuple<uint64_t, uint64_t, uint64_t>> move_info)
+{
+  int r = 0;
+  bufferlist bl;
+  baseo->exists = true;
+  _assign_nid(txc, baseo);
+
+//for loop
```
athanatos: I'm pretty sure most people will recognize that it's a for loop without the comment.

src/os/bluestore/BlueStore.cc
manali26: destination.

src/os/kstore/KStore.cc
manali26: changed

src/os/kstore/KStore.cc
manali26: added appropriate comments

athanatos:
You shouldn't have any commits fixing things introduced in earlier commits in the same branch.  Please clean that up.

src/os/kstore/KStore.cc
```diff
@@ -2360,6 +2360,16 @@ void KStore::_txc_add_transaction(TransContext *txc, Transaction *t)
...
 
+    case Transaction::OP_MERGE_DELETE:
+      {
+        const ghobject_t& boid = i.get_oid(op->dest_oid);
+        OnodeRef bo = c->get_onode(boid, true);
+        vector<boost::tuple<uint64_t, uint64_t, uint64_t>> move_info;
+        i.decode_move_info(move_info);
+        r = _move_ranges_destroy_src(txc, c, o, cvec[op->dest_cid], bo, move_info);
+      }
+			break;
```
athanatos: bad spacing.

src/os/filestore/FileStore.cc
```diff
@@ -3741,11 +3760,86 @@ int FileStore::_clone_range(const coll_t& cid, const ghobject_t& oldoid, const g
...
+
+  FDRef t;
+  r = lfn_open(src_cid, src_oid, false, &t);
+  //If we are replaying, it is possible that we do not find src obj as it is deleted before crashing.
+  if (r < 0) {
+    lfn_close(b);
+    dout(10) << __func__ << " replaying -->" << replaying  << dendl;
+    if (replaying)
+      _set_replay_guard(**b, spos, &oid);
+    return 0;
```
athanatos: This should return -ENOENT if !replaying

athanatos:
c7a86f519c8ac9f76c4143968a80dbf59478f28c only fixes issues with a previous commit, it should be merged with the commit which introduces the new methods.

src/os/ObjectStore.h
```diff
@@ -1362,6 +1367,34 @@ class ObjectStore {
...
+  void move_ranges_destroy_src(const coll_t& cid, const ghobject_t& src_oid, ghobject_t oid,
+                     const vector<boost::tuple<uint64_t, uint64_t, uint64_t>> move_info) {
+    if (use_tbl) {
+      __u32 op = OP_MERGE_DELETE;
+      ::encode(op, tbl);
+      ::encode(src_oid, tbl);
+      ::encode(cid, tbl);
+      ::encode(oid, tbl);
+      ::encode(move_info, tbl);
+    } else {
```
liewegas: you can drop the use_tbl case here; it's about to be ripped out anyway and will just cause a merge conflict if it's included here.

src/os/bluestore/BlueStore.cc
liewegas: ok

src/os/filestore/FileStore.cc
```diff
@@ -2669,13 +2670,14 @@ void FileStore::_do_transaction(
...
         ghobject_t noid = i.get_oid(op->dest_oid);
-	_kludge_temp_object_collection(cid, noid);
+	_kludge_temp_object_collection(cid, oid);
+	_kludge_temp_object_collection(ncid, noid);
         uint64_t off = op->off;
         uint64_t len = op->len;
         tracepoint(objectstore, clone_range_enter, osr_name, len);
-        r = _clone_range(cid, oid, noid, off, len, off, spos);
+        r = _clone_range(cid, oid, ncid, noid, off, len, off, spos);
         tracepoint(objectstore, clone_range_exit, r);
```
liewegas: no need to change CLONERANGE here.. it's deprecated anyway!

src/os/filestore/FileStore.cc
```diff
@@ -2684,18 +2686,35 @@ void FileStore::_do_transaction(
...
         tracepoint(objectstore, clone_range2_exit, r);
       }
       break;
 
+    case Transaction::OP_MERGE_DELETE:
+      {
+        ghobject_t src_oid = i.get_oid(op->oid);
+        coll_t cid = i.get_cid(op->cid);
+        ghobject_t oid = i.get_oid(op->dest_oid);
+        coll_t src_cid = i.get_cid(op->cid);
```
liewegas: and let's drop teh second coll_t argument here, too, and require that this op be used within the same collection.  just say coll_t src_cid = cid;

src/os/filestore/FileStore.cc
```diff
@@ -2669,13 +2670,14 @@ void FileStore::_do_transaction(
       {
         coll_t cid = i.get_cid(op->cid);
         ghobject_t oid = i.get_oid(op->oid);
-	_kludge_temp_object_collection(cid, oid);
+        coll_t ncid = i.get_cid(op->cid);
```
liewegas: just coll_t ncid = cid;
don't call get_cid() again.

src/os/kstore/KStore.cc
manali26: changed it

src/os/ObjectStore.h
manali26: ok

src/os/filestore/FileStore.cc
manali26: ok

src/os/filestore/FileStore.cc
manali26: This was changed because clone_range() did not allow specifying a source collection different from the destination collection.  Keeping it as it is.
http://tracker.ceph.com/issues/16591

src/os/filestore/FileStore.cc
manali26: This was done to allow specifying src collection different than destination collection only in filestore. Object store level, we have only one collection. At filestore, we call _kludge_temp_object_collection() which can mostly result in different collections for above objects.
 Sam can comment on this more, if required, as why we did this.

src/os/filestore/FileStore.cc
manali26: ok

src/os/filestore/FileStore.cc
liewegas: i see. thanks!

liewegas:
repalced by https://github.com/ceph/ceph/pull/11237 (same code, rebased and cleaned up a bit)

PR #10262, opened 2016-07-12
ceph-disk: timeout ceph-disk to avoid blocking forever

dachary:
When ceph-disk runs from udev or init script, it is in the background
and should it block for any reason, it may keep a lock forever. All
calls to ceph-disk in these context are changed to timeout.

Fixes: http://tracker.ceph.com/issues/16580

### DIFF
--- a/systemd/ceph-disk@.service
+++ b/systemd/ceph-disk@.service
@@ -4,5 +4,5 @@ Description=Ceph disk activation: %f
 [Service]
 Type=oneshot
 KillMode=none
-ExecStart=/bin/sh -c 'flock /var/lock/ceph-disk /usr/sbin/ceph-disk --verbose --log-stdout trigger --sync %f'
+ExecStart=/bin/sh -c 'timeout 120 flock /var/lock/ceph-disk /usr/sbin/ceph-disk --verbose --log-stdout trigger --sync %f'
 TimeoutSec=0


systemd/ceph-disk@.service
```diff
@@ -4,5 +4,5 @@ Description=Ceph disk activation: %f
 [Service]
 Type=oneshot
 KillMode=none
-ExecStart=/bin/sh -c 'flock /var/lock/ceph-disk /usr/sbin/ceph-disk --verbose --log-stdout trigger --sync %f'
+ExecStart=/bin/sh -c 'timeout 120 flock /var/lock/ceph-disk /usr/sbin/ceph-disk --verbose --log-stdout trigger --sync %f'
 TimeoutSec=0
```
ktdreyer: Could you set `TimeoutSec` to something other than `0` here? (or explain in the commit message why not?)

ktdreyer:
LGTM; @smithfarm @b-ranto care to review?

smithfarm:
In SUSE, `timeout` is part of the `coreutils` package which I don't see required in the spec file. It's probably present on all systems, but it wouldn't hurt to add the dependency.

smithfarm:
https://github.com/dachary/ceph/pull/8

b-ranto:
FYI: We could also use the timeout option provided by flock itself. It even seems like a bit safer option to me since it removes one level of misdirection:

flock --timeout 120 /var/lock/ceph-disk ...

Also, you don't need to require coreutils for timeout, it is considered as a part of the base system that you don't need to specify in the spec file (this includes things like bash, sed, coreutils, etc).

Anyway, we use bits from coreutils all the time. After all, rm, mv, .... are all there.

wjwithagen:
On 13-7-2016 11:36, Boris Ranto wrote:
> FYI: We could also use the timeout option provided by flock itself. It
> even seems like a bit safer option to me since it removes one level of
> misdirection:
> 
> flock --timeout 120 /var/lock/ceph-disk ...
> 
> Also, you don't need to require coreutils for timeout, it is considered
> as a part of the base system that you don't need to specify in the spec
> file (this includes things like bash, sed, coreutils, etc).
> 
> Anyway, we use bits from coreutils all the time. After all, rm, mv, ....
> are all there.

What I like about timeout is that it really reports all the possible
errors as return value, making it possible to actually infor the user on
what went wrong. Other than just exiting without any indication.

flock is not basically available in FreeBSD. So there I'd need to search
for a port anyways. And thus I'm not aware of any of the returncode that
flock supplies.

dachary:
@b-ranto 

The flock timeout only applies to locking the file given in argument to flock, it won't timeout the ceph-disk command itself when it blocks forever and that's what we're trying to address.

<pre>
       -w, --wait, --timeout seconds
              Fail if the lock cannot be acquired within seconds.  Decimal fractional values are allowed.  See the
              -E option for the exit code used. The zero number of seconds is interpreted as --nonblock.
</pre>

dachary:
test this please

ktdreyer:
Loic, this does not seem to affect `ceph-disk list`, which was the original command where the reporter found this hanging forever. Can you build this into ceph-disk itself so that `list` does not hang?

b-ranto:
@dachary that is right but timeout won't do a cleanup if it is indeed ceph-disk command that hangs, i.e.:

`* timeout 2 flock sleep.lock sleep 30`
`* ls sleep.lock # the lock file still exists`
`mehmeh`

dachary:
> Can you build this into ceph-disk itself so that list does not hang?

The only reason why ceph-disk list hangs is because there is a ceph-disk activate blocked somewhere. Since that won't happen anymore, ceph-disk list will return eventually, because ceph-disk activate as run by udev will timeout.

dachary:
test this please

dachary:
> that is right but timeout won't do a cleanup if it is indeed ceph-disk command that hangs

It does not matter if the sleep.lock file exists as long as it's not held by a process.

<pre>
loic@fold:~$ timeout 2 flock sleep.lock sleep 30
loic@fold:~$ ls sleep.lock
sleep.lock
loic@fold:~$ time flock sleep.lock sleep 30
real	0m30.007s
user	0m0.000s
sys	0m0.000s
</pre>

tchaikov:
lgtm. and @dachary has addressed all the concerns.

dachary:
Hey jenkins, test this please !

yuriw:
re-run
http://pulpito.ceph.com/yuriw-2016-08-03_08:27:33-rados-wip-yuri-testing_2016_08_02-distro-basic-smithi/

http://pulpito.ceph.com/yuriw-2016-08-02_15:37:57-rados-wip-yuri-testing_2016_08_02-distro-basic-smithi/

PR #10263, opened 2016-07-12
DNM: Parallel Mon client connect

sdieffen:


src/mon/MonClient.cc
```diff
@@ -92,7 +91,7 @@ int MonClient::get_monmap()
   
   _sub_want("monmap", 0, 0);
   if (cur_mon.empty())
-    _reopen_session();
+    _multiple_reopen_session();
```
athanatos: _reopen_session is still pretty descriptive, I think, no need to change.

src/mon/MonClient.cc
```diff
@@ -264,13 +264,41 @@ bool MonClient::ms_dispatch(Message *m)
   }
 
   Mutex::Locker lock(monc_lock);
+  ldout(cct, 10) << "[ms_dispatch] : incoming msg [" << *m << "] from addr ["
```
athanatos: I actually like the [] format, but nothing else uses it.  Most things use << __func__ << ": ... to indicate which method it is.

src/mon/MonClient.cc
```diff
@@ -264,13 +264,41 @@ bool MonClient::ms_dispatch(Message *m)
...
+      ldout(cct, 10) << "[ms_dispatch] : disposing of monmap magic" << dendl;
+      return true;
+    }
+    ldout(cct, 10) << "[ms_dispatch] : accepting monmap magic" << dendl;
+  }
 
-  // ignore any messages outside our current session
-  if (m->get_connection() != cur_con) {
-    ldout(cct, 10) << "discarding stray monitor message " << *m << dendl;
+  if ((m->get_connection() != cur_con) && (m->get_type() != CEPH_MSG_AUTH_REPLY || m->get_type() != CEPH_MSG_MON_MAP)) {
```
athanatos: 80 characters?

src/mon/MonClient.cc
```diff
@@ -593,28 +704,125 @@ string MonClient::_pick_random_mon()
...
+    mon = monmap.get_name(rank);
+  }
+
+  int attempt = monmap.size(); //FIXREF
+
+  //mark down all connections, reset the state_map
+  _mark_down_all();
+
+  // create and fill the connection vector
+  vector<ConnectionRef> con_vec;
```
tchaikov: if the a certain mon is specified, do we still want to initiate multiple connections to all mons?

src/mon/MonClient.cc
```diff
@@ -593,28 +704,125 @@ string MonClient::_pick_random_mon()
...
+  con_vec.push_back(messenger->get_connection(monmap.get_inst(mon)));
+  for (int j = 1; j < attempt; j++) {
+    con_vec.push_back(messenger->get_connection(monmap.get_inst(monmap.get_name(j))));
```
tchaikov: might want to skip the "selected" mon?

src/mon/MonClient.cc
tchaikov: nit, personally, might be better not encode the type of variable into its name, in this case, `conns` would be a better name, imho.

src/mon/MonClient.cc
```diff
@@ -593,28 +704,125 @@ string MonClient::_pick_random_mon()
...
+    _send_mon_message(m, con_vec[j], true); 
+  }
+
+  // for current compatibility issues, save the initial (mon, con) set
+  ldout(cct, 10) << "[_multiple_reopen_session] : cur_mon gets [" << mon << "] and cur_con->addr gets ["
+    << con_vec[0]->get_peer_addr() << "]" << dendl;
+  cur_mon = mon;
+  cur_con = con_vec[0];
+
+  for (map<string,ceph_mon_subscribe_item>::iterator p = sub_sent.begin();
```
tchaikov: nit, could use ranged for loop, like
```c++
for (const auto& sub : sub_sent) {
  if (sub_new.count(sub.first) == 0)
  //...
```

tchaikov: nit, could use ranged for loop, like
```c++
for (const auto& sub : sub_sent) {
  if (sub_new.count(sub.first) == 0)
  //...
```

src/mon/MonClient.cc
```diff
@@ -593,28 +704,125 @@ string MonClient::_pick_random_mon()
...
+    _set_state(con_vec[j]->get_peer_addr(), MC_STATE_NEGOTIATING, true);
+    ldout(cct, 10) << "_set_state on " << con_vec[j]->get_peer_addr()
+	<< " to NEGOTIATING." << dendl;
+  }
+  hunting = true;
+  
+  // send an initial keepalive to ensure our timestamp is valid by the
+  // time we are in an OPENED state (by sequencing this before
+  // authentication).
+  for (int j = 0; j < attempt; j++) {
```
tchaikov: why not just loop in `con_vec`?

tchaikov:
when i was working on this ticket, i thought a cleaner approach might be to extract the session initiation stuff into a separated class like the `MonHunter` in https://github.com/tchaikov/ceph/commit/475a052550eb81cf2557fa488db4a9c7bc2cb87a . what do you think?

src/mon/MonClient.cc
```diff
@@ -74,36 +73,36 @@ MonClient::~MonClient()
 {
   delete auth_supported;
   delete session_established_context;
-  delete auth;
+  auth.reset();
```
athanatos: No need.  unique_ptr will delete its contained pointer in its own destructor (which will be called along with the destructors for all other MonClient members after the body of ~MonClient()).  That's the point of RAII!

src/mon/MonClient.cc
```diff
@@ -74,36 +73,36 @@ MonClient::~MonClient()
...
 {
-  ldout(cct, 10) << "build_initial_monmap" << dendl;
+  ldout(cct, 10) << __func__ << dendl;
   return monmap.build_initial(cct, cerr);
 }
 
 int MonClient::get_monmap()
 {
-  ldout(cct, 10) << "get_monmap" << dendl;
+  ldout(cct, 10) << __func__ << dendl;
```
athanatos: These cleanups should go into their own commit.

src/mon/MonClient.h
```diff
@@ -106,7 +105,8 @@ class MonClient : public Dispatcher {
 public:
   MonMap monmap;
 private:
-  MonClientState state;
+  map<entity_addr_t, MonClientState> state_map;
+  map<entity_addr_t, unique_ptr<AuthClientHandler>> auth_map;
```
athanatos: Hmm, I'm not sure about this.  I think you need to hold references to the pending connections.  I think the best way forward would be to define a:

```
struct MonConnection {
  ConnectionRef conn;
  string name;
  unique_ptr<AuthClientHandler> auth;
  MonClientState state;
};
```

Once you have that, you can replace state_map and auth_map with

```
map<entity_inst_t, unique_ptr<MonConnection> > pending_connections;
```

Next, you can replcae curr_conn, cur_mon, and auth with

```
unique_ptr<MonConnection> active_connection;
```

If I'm understanding this right, you can replace hunting and reopen with:

```
bool hunting() const { return !pending_connections.empty(); }
bool reopen() const { return hunting() || active_connection.get() != nullptr; }
```

and at all times

```
assert(pending_connections.empty() || (active_connection.get() == nullptr);
```

sdieffen:
https://github.com/ceph/ceph/pull/10715

tchaikov:
i am closing this PR as it is superseded by #10715.

PR #10271, opened 2016-07-12
rgw_file: fix rename cases and unify unlink

mattbenjamin:
Rather ordinary rename cases failed in the unlink step.  Fix this,
unifying the unlink path while at it.


mattbenjamin:
depends on 8618

PR #10274, opened 2016-07-13
rgw: collect skips a specific coroutine stack

yehudasa:
Fixes: http://tracker.ceph.com/issues/16665

Instead of drain_all_but() that specifies number of stacks to leave behind,
added drain_all_but_stack() that has a specific stack specified. This is needed
so that we don't call wakeup() through lease_cr->go_down() on a cr stack that
was already collected.

### DIFF
--- a/src/rgw/rgw_coroutine.cc
+++ b/src/rgw/rgw_coroutine.cc
@@ -292,7 +292,7 @@ int RGWCoroutinesStack::unwind(int retcode)
 }
 
 
-bool RGWCoroutinesStack::collect(RGWCoroutine *op, int *ret) /* returns true if needs to be called again */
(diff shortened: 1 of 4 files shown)


cbodley:
looks good. only suggestion would be to make the `skip_stack` argument optional (default to nullpr)

yehudasa:
@cbodley yeah, I wanted to make sure that I visited all call sites when going through the change

cbodley:
I think we should be holding a reference to these `lease_stack`s. Even though we only use it to check for pointer equality in `collect()` and never dereference, it's possible for it to be freed and reallocated by another `spawn()`. This would cause us to skip the wrong stack

PR #10276, opened 2016-07-13
doc: remove i386 from minimal hardware recommendations

tchaikov:
as we don't build or test on i386 machines now. also, we don't build
i386 releases anymore. and regular users are on amd64 or arm.


badone:
lgtm

vumrao:
Thanks Kefu!

PR #10069, opened 2016-06-30
rgw: add support for the /crossdomain.xml resource of Swift API.

rzarzynski:
CC: @yehudasa, @cbodley

PR #10280, opened 2016-07-13
rgw: add suport for Swift-at-root dependent features of Swift API

rzarzynski:
This pull request brings:
* a temporary, provisional fix for the [Swift API-at-root issue](http://tracker.ceph.com/issues/16673) that is small enough to be backported to Jewel.
* implementation of several parts of Swift API that has to be placed in the root of URL hierarchy (`/crossdomain.xml`, `/healthcheck` and `/info` by @pritha-srivastava adapted to this solution). 

Placing Swift-at root-definitely needs an ultimate solution as the temporary, proposed one exhibits severe side effect: inability to deploy RadosGW in multi-site configuration. This is because some S3 operations working in `system_request` mode are necessary for that to operate.

Derogates: #9900, #10069, #10213 and - only if Pritha agrees - #9582.
CC: @pritha-srivastava, @yehudasa, @cbodley.

Tempest and S3 haven't found any regression here. The test for`/info` passes as well though it's [really simple](https://github.com/openstack/tempest/blob/master/tempest/services/object_storage/account_client.py#L148) - check whether response code is `200 OK` and body is a properly formatted JSON.


rzarzynski:
Rebased due to a merge conflict.

cbodley:
looks good to me

src/rgw/rgw_main.cc
```diff
@@ -342,18 +342,51 @@ int main(int argc, const char **argv)
   }
 
   // S3 website mode is a specialization of S3
-  bool s3website_enabled = apis_map.count("s3website") > 0;
-  if (apis_map.count("s3") > 0 || s3website_enabled)
-    rest.register_default_mgr(set_logging(new RGWRESTMgr_S3(s3website_enabled)));
+  const bool s3website_enabled = apis_map.count("s3website") > 0;
+  // Swift API entrypoint could placed in the root instead of S3
+  const bool swift_at_root = g_conf->rgw_swift_url_prefix == "/";
```
pritha-srivastava: A question - do we have to take into account if 'swift' api is enabled, while determining whether swift entry point is at root or not?

pritha-srivastava:
@rzarzynski: Looks good to me too.

src/rgw/rgw_main.cc
rzarzynski: I don't think so. The purpose of the `swift_at_root` is to indicate whether someone _requested_ putting Swift API at root. It could be placed there only if Swift API is enabled and S3 (and S3 website) are turned off.

pritha-srivastava: ok, i asked this question because what if someone changes the default url prefix to '/' and then decides to disable the swift api altogether, then we could allow the s3/ s3 website to be enabled. This may be a case which may be impossible in a real scenario :)

rzarzynski:
Tested as a part of `wip-radek-testing`. Teuthology runs:
 * http://pulpito.ceph.com/rzarzynski-2016-08-22_16:44:18-rgw-wip-radek-testing---basic-mira
 * http://pulpito.ceph.com/rzarzynski-2016-08-29_11:18:44-rgw-wip-radek-testing---basic-mira

The failures seem to be unrelated to the changes.

The Pritha's commit LGTM.

dachary:
@rzarzynski Hi ! Would you be willing to try a backport to jewel ? It looks non trivial :-) If you don't have time for that don't worry about it. For the record the corresponding backport issue is http://tracker.ceph.com/issues/17208

rzarzynski:
@dachary: Hi Loic! I'm putting the backport on my TODO. I think that `/info` would be quite desired in Jewel also because of the [recent change](https://review.openstack.org/#/c/306788/) in Tempest that assumes its availability.

dachary:
@rzarzynski thanks, much appreciated. If it's easier than it looks, just tell me and I'll give it a try :-)

PR #10287, opened 2016-07-13
cmake: add WITH_TESTS option to disable test subdir

mattbenjamin:

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1022,7 +1022,11 @@ install(TARGETS crushtool DESTINATION bin)
 # Support/Tools
 add_subdirectory(googletest/googlemock)
 
-add_subdirectory(test)
+option(WITH_TESTS "Unit tests and 'make check' target enabled" ON)
+if(WITH_TESTS)
+  add_subdirectory(test)
+endif(WITH_TESTS)
+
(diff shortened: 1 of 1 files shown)


src/CMakeLists.txt
```diff
@@ -1022,7 +1022,11 @@ install(TARGETS crushtool DESTINATION bin)
 # Support/Tools
 add_subdirectory(googletest/googlemock)
 
-add_subdirectory(test)
+option(WITH_TESTS "Unit tests and 'make check' target enabled" ON)
```
tchaikov: i am curious what is the use case of `WITH_TEST=OFF`. 

- packaging need it. it's explained in detail later
- "make check" needs it
- maybe it's useful for a developer who don't run "make check" locally, and use vstart.sh for testing?

actually, some qa units also depend on the artifact created in the "test" subdirectory. for example, `ceph_test_cls_rgw_meta` and `ceph_kvstorebench`.

if we really want this option, we might need to update `debian/rules` and `ceph.spec.in` accordingly.

tchaikov:
@mattbenjamin and @cbodley now i understand the use case, when looking at #10872. maybe downstream does not want to do the testing when packaging Ceph. i am closing this one, as #10872 is more complete solution.

PR #10288, opened 2016-07-13
msg/AsyncConnection: move DelayedDelivery class

stiopaa1:

### DIFF
--- a/src/msg/async/AsyncConnection.cc
+++ b/src/msg/async/AsyncConnection.cc
@@ -118,6 +118,61 @@ static void alloc_aligned_buffer(bufferlist& data, unsigned len, unsigned off)
   }
 }
 
+ /**
+ * The DelayedDelivery is for injecting delays into Message delivery off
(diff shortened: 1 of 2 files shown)


src/msg/async/AsyncConnection.cc
```diff
@@ -118,6 +118,61 @@ static void alloc_aligned_buffer(bufferlist& data, unsigned len, unsigned off)
   }
 }
 
+ /**
+ * The DelayedDelivery is for injecting delays into Message delivery off
+ * the socket. It is only enabled if delays are requested, and if they
+ * are then it pulls Messages off the DelayQueue and puts them into the
+ * AsyncMessenger event queue.
+ */
+class AsyncConnection::DelayedDelivery : public EventCallback {
```
tchaikov: could hide it in an anonymous namespace if it's not supposed to be used out side of this compilation unit.

liewegas:
needs rebase

PR #10290, opened 2016-07-14
mds/MDSMap: add 'fsid' set to MDSMap

renhwztetecs:
To track which cluster of MDSMap by the 'fsid';
add display max_mds when run "ceph mds stop"

### DIFF
--- a/src/mds/MDSMap.cc
+++ b/src/mds/MDSMap.cc
@@ -129,6 +129,7 @@ void MDSMap::mds_info_t::generate_test_instances(list<mds_info_t*>& ls)
 void MDSMap::dump(Formatter *f) const
 {
   f->dump_int("epoch", epoch);
+  f->dump_stream("fsid") << get_fsid();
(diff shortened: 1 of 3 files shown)


jcsp:
MDSMaps are implicitly scoped to an FSID: whenever you read one from a mon you know what mon you're talking to, so you know the FSID.  What's the use case for embedding the FSID in the MDSMap?

renhwztetecs:
yeh
but if we use command “ceph mds dump > ***.txt” export to file，unable to identify the mdsmap belonging to the Cluster through the file.  especially in the maintenance of Multi cluster

jcsp:
Sorry, I don't think this makes sense.  We have lots of commands, and most of them don't inline the FSID in the output.  If you have tools that require it, I would suggest running a 'status' before your other command so that the output contains the fsid for you.

renhwztetecs:
thanks for your reviews

PR #9481, opened 2016-06-03
rgw: stop bucket lease only after draining object sync operations

yehudasa:
Fixes: http://tracker.ceph.com/issues/16143

Signed-off-by: Yehuda Sadeh <yehuda@redhat.com>

PR #10293, opened 2016-07-14
jewel: multisite sync races with deletes

theanalyst:
http://tracker.ceph.com/issues/16589


theanalyst:
@cbodley @yehudasa this is a backport for http://tracker.ceph.com/issues/16464 contains #9481 #9851 (which should fix the delete issue) I had to pick #9295 as well for a clean backport, this looks harmless enough  to be backported for jewel we can drop it. What do you think of adding #9295?

cbodley:
#9295 should be harmless to pull in :+1:

yehudasa:
@theanalyst yeah, we can pull it in

theanalyst:
jenkins retest this please

PR #10296, opened 2016-07-14
mon: small change on the HealthMonitor start_epoch  function

songbaisen:
mon: small change on the HealthMonitor start_epoch  function

### DIFF
--- a/src/mon/HealthMonitor.cc
+++ b/src/mon/HealthMonitor.cc
@@ -66,9 +66,10 @@ bool HealthMonitor::service_dispatch(MonOpRequestRef op)
 }
 
 void HealthMonitor::start_epoch() {
+  epoch_t epoch = get_epoch();
   for (map<int,HealthService*>::iterator it = services.begin();
        it != services.end(); ++it) {
-    it->second->start(get_epoch());
+    it->second->start(epoch);
   }
 }
 


src/mon/HealthMonitor.h
```diff
@@ -45,9 +45,10 @@ class HealthMonitor : public QuorumService
   virtual bool service_dispatch(MonOpRequestRef op);
 
   virtual void start_epoch() {
+    epoch_t epoch = get_epoch();
     for (map<int,HealthService*>::iterator it = services.begin();
          it != services.end(); ++it) {
-      it->second->start(get_epoch());
```
tchaikov: i don't think this helps with the performance, as this function will be inlined anyway.

songbaisen: @tchaikov  :smile: Thank you sir!For your review. I am not sure that whether the compiler can inline it.If we do not  statement inline in the constructor function.If that true I will close this pr.

tchaikov: please check the assembly with the default "CMAKE_BUILD_TYPE" if you are not sure.

songbaisen: @tchaikov  Yes sir.That is true.But the modify here is really save time.I write a very simple test like below:

[root@localhost test]# cat test1.cc 
#include <unistd.h>
#include <fcntl.h>
#include <errno.h>
#include <string.h>
#include <stdlib.h>
#include <string>
#include <iostream>
#include <stdio.h>
#include <sys/stat.h>
#include <sys/types.h>
using namespace std;
#define TEST_MAX 10


unsigned long long gettickcount(void) {
  struct timespec ts;
  if (clock_gettime(CLOCK_MONOTONIC, &ts) != 0) {
    return 0;
  }
  return (ts.tv_nsec) + ts.tv_sec * 1000000000;
}

class ctest
{
public:
      void setvalue(int i) {ival = i;};
      int getvalue() {return ival;};
      void test1()
      {
         for(int i = 0; i < TEST_MAX; i++)
         {
             int a =  getvalue();
         }
      };

      void test2()
      {
         int b = getvalue();
         for(int i = 0; i < TEST_MAX; i++)
         {
             int a = b;
         }
      };

private:
     int ival;

};





int main()
{
  ctest  t;
  t.setvalue(5);
  unsigned long long t1, t2;
  t1 = gettickcount();
  t.test1();
  t1 = gettickcount() - t1;
  t2 = gettickcount();
  t.test2();
  t2 = gettickcount() - t2;
  printf("v1: %llu ns \n", t1 );
  printf("v2: %llu ns\n", t2 );
  return 0;
}
[root@localhost test]# g++ -o code test1.cc -O2 --std=c++11 -lrt
[root@localhost test]# ./code 
v1: 313 ns 
v2: 116 ns
[root@localhost test]#

jecluis: I agree with @tchaikov - I seriously doubt this will have any significant impact whatsoever in the monitor's performance. Much more so because this function is not heavily used. But I'm not opposed to merging this if the commit message is something more down the lines of making code readable rather than boasting an optimization.

Besides, unless we expect the state to change (i.e., get_epoch() to return a different value on each iteration, which we don't), it's just good practice to keep the get_epoch() out of the loop.

branch-predictor: On Sun, Jul 17, 2016 at 06:02:17PM -0700, song.baisen wrote:
> >      for (map<int,HealthService*>::iterator it = services.begin();
> >           it != services.end(); ++it) {
> > -      it->second->start(get_epoch());
> 
> @jecluis  Hello.Thank you for review/How about change the tittle to small change  on the HealthMonitor start_epoch function .@branch-predictor  Hi! How do you think about this?

I agree with others, and I also doubt it'll be meaningful change -- unless 
you can prove that this change can save tangible amount of CPU processing 
time over reasonable mon lifetime. Otherwise it's more of code cleanup than
optimization. 

-- 
Piotr Dałek
branch@predictor.org.pl
http://blog.predictor.org.pl

songbaisen: @branch-predictor  Thank you for reply.I change the title.

branch-predictor: On Sun, Jul 17, 2016 at 11:06:57PM -0700, song.baisen wrote:
> >      for (map<int,HealthService*>::iterator it = services.begin();
> >           it != services.end(); ++it) {
> > -      it->second->start(get_epoch());
> 
> @branch-predictor  Thank you for reply.I change the title.

Please change the commit message as well.

songbaisen: @branch-predictor  Already done it.

branch-predictor: On Mon, Jul 18, 2016 at 12:13:18AM -0700, song.baisen wrote:
> >      for (map<int,HealthService*>::iterator it = services.begin();
> >           it != services.end(); ++it) {
> > -      it->second->start(get_epoch());
> 
> @branch-predictor  Already done it.

Thanks!

dachary:

dachary:
this needs rebasing and can be merged after the test pass

songbaisen:
@dachary  @tchaikov  @jecluis  Done it.

PR #10297, opened 2016-07-14
msg/simple: save the errno in case being changed by subsequent codes

Yan-waller:
should save the `errno` which may be changed by `ldout` and/or `<<` operator

Signed-off-by: Yan Jun yan.jun8@zte.com.cn

### DIFF
--- a/src/msg/simple/Pipe.cc
+++ b/src/msg/simple/Pipe.cc
@@ -829,7 +829,8 @@ void Pipe::set_socket_options()
     int r = ::setsockopt(sd, IPPROTO_TCP, TCP_NODELAY, (char*)&flag, sizeof(flag));
     if (r < 0) {
       r = -errno;
-      ldout(msgr->cct,0) << "couldn't set TCP_NODELAY: " << cpp_strerror(r) << dendl;
+      ldout(msgr->cct,0) << "couldn't set TCP_NODELAY: "
+                         << cpp_strerror(r) << dendl;
     }
   }
   if (msgr->cct->_conf->ms_tcp_rcvbuf) {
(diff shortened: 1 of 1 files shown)


src/msg/simple/Pipe.cc
```diff
@@ -828,16 +828,16 @@ void Pipe::set_socket_options()
     int flag = 1;
     int r = ::setsockopt(sd, IPPROTO_TCP, TCP_NODELAY, (char*)&flag, sizeof(flag));
     if (r < 0) {
-      r = -errno;
-      ldout(msgr->cct,0) << "couldn't set TCP_NODELAY: " << cpp_strerror(r) << dendl;
+      ldout(msgr->cct,0) << "couldn't set TCP_NODELAY: "
```
tchaikov: this is intentional, as `ldout` and `<<` operator might change the errno, and they are evaluated before `cpp_strerror(errno)`, so we save the `errno` to `r` first.

Yan-waller: @tchaikov Oh, thanks for your review,  then maybe we should also save the `errno` at  https://github.com/ceph/ceph/pull/10297/files#diff-459b1eb43a0f5b6b6b430e3c67fee6ceL861  and  https://github.com/ceph/ceph/pull/10297/files#diff-459b1eb43a0f5b6b6b430e3c67fee6ceL873  in this function ?

tchaikov: yeah.

Yan-waller: @tchaikov thanks, fixed it as per your suggestion and changed the title.

tchaikov:
http://pulpito.ceph.com/kchai-2016-11-27_12:44:55-rados-wip-kefu-testing2---basic-mira/

PR #10307, opened 2016-07-15
rgw: Rework of s3 LDAP Authentication code.

pritha-srivastava:
The LDAP authentication code has been reworked based
on the new authentication infrastructure.

### DIFF
--- a/src/rgw/rgw_auth.cc
+++ b/src/rgw/rgw_auth.cc
@@ -149,6 +149,11 @@ void RGWRemoteAuthApplier::create_account(const rgw_user& acct_user,
 {
   rgw_user new_acct_user = acct_user;
 
+  if (info.acct_type) {
+    //ldap/keystone for s3 users
+    user_info.type = info.acct_type;
+  }
+
   /* Administrator may enforce creating new accounts within their own tenants.
    * The config parameter name is kept due to legacy. */
   if (new_acct_user.tenant.empty() && g_conf->rgw_keystone_implicit_tenants) {
--- a/src/rgw/rgw_auth.h
+++ b/src/rgw/rgw_auth.h
@@ -117,16 +117,19 @@ public:
     const std::string acct_name;
     const uint32_t perm_mask;
     const bool is_admin;
+    const uint32_t acct_type;
 
   public:
     AuthInfo(const rgw_user& acct_user,
              const std::string& acct_name,
              const uint32_t perm_mask,
-             const bool is_admin)
+             const bool is_admin,
+             const uint32_t acct_type=TYPE_NONE)
     : acct_user(acct_user),
       acct_name(acct_name),
       perm_mask(perm_mask),
-      is_admin(is_admin) {
+      is_admin(is_admin),
+      acct_type(acct_type) {
     }
   };
 
--- a/src/rgw/rgw_common.h
+++ b/src/rgw/rgw_common.h
@@ -527,6 +527,14 @@ void encode_json(const char *name, const RGWUserCaps& val, Formatter *f);
 
 void decode_json_obj(obj_version& v, JSONObj *obj);
 
+enum RGWUserSourceType
+{
+  TYPE_NONE=0,
+  TYPE_RGW=1,
+  TYPE_KEYSTONE=2,
+  TYPE_LDAP=3
+};
+
 struct RGWUserInfo
 {
   uint64_t auid;
@@ -547,6 +555,7 @@ struct RGWUserInfo
   RGWQuotaInfo bucket_quota;
   map<int, string> temp_url_keys;
   RGWQuotaInfo user_quota;
+  uint32_t type;
 
   RGWUserInfo()
     : auid(0),
@@ -554,7 +563,8 @@ struct RGWUserInfo
       max_buckets(RGW_DEFAULT_MAX_BUCKETS),
       op_mask(RGW_OP_TYPE_ALL),
       admin(0),
-      system(0) {
+      system(0),
+      type(TYPE_NONE) {
   }
 
   RGWAccessKey* get_key0() {
@@ -565,7 +575,7 @@ struct RGWUserInfo
   }
 
   void encode(bufferlist& bl) const {
-     ENCODE_START(18, 9, bl);
+     ENCODE_START(19, 9, bl);
      ::encode(auid, bl);
      string access_key;
      string secret_key;
@@ -605,10 +615,11 @@ struct RGWUserInfo
      ::encode(user_quota, bl);
      ::encode(user_id.tenant, bl);
      ::encode(admin, bl);
+     ::encode(type, bl);
      ENCODE_FINISH(bl);
   }
   void decode(bufferlist::iterator& bl) {
-     DECODE_START_LEGACY_COMPAT_LEN_32(18, 9, 9, bl);
+     DECODE_START_LEGACY_COMPAT_LEN_32(19, 9, 9, bl);
      if (struct_v >= 2) ::decode(auid, bl);
      else auid = CEPH_AUTH_UID_DEFAULT;
      string access_key;
@@ -678,6 +689,9 @@ struct RGWUserInfo
     if (struct_v >= 18) {
       ::decode(admin, bl);
     }
+    if (struct_v >= 19) {
+      ::decode(type, bl);
+    }
     DECODE_FINISH(bl);
   }
   void dump(Formatter *f) const;
@@ -727,7 +741,7 @@ struct rgw_bucket {
   }
 
   void encode(bufferlist& bl) const {
-     ENCODE_START(8, 3, bl);
+     ENCODE_START(9, 3, bl);
     ::encode(name, bl);
     ::encode(data_pool, bl);
     ::encode(marker, bl);
@@ -738,7 +752,7 @@ struct rgw_bucket {
     ENCODE_FINISH(bl);
   }
   void decode(bufferlist::iterator& bl) {
-    DECODE_START_LEGACY_COMPAT_LEN(8, 3, 3, bl);
+    DECODE_START_LEGACY_COMPAT_LEN(9, 3, 3, bl);
     ::decode(name, bl);
     ::decode(data_pool, bl);
     if (struct_v >= 2) {
--- a/src/rgw/rgw_json_enc.cc
+++ b/src/rgw/rgw_json_enc.cc
@@ -427,6 +427,26 @@ void RGWUserInfo::dump(Formatter *f) const
   encode_json("bucket_quota", bucket_quota, f);
   encode_json("user_quota", user_quota, f);
   encode_json("temp_url_keys", temp_url_keys, f);
+
+  string user_source_type;
+  switch ((RGWUserSourceType)type) {
+  case TYPE_RGW:
+    user_source_type = "rgw";
+    break;
+  case TYPE_KEYSTONE:
+    user_source_type = "keystone";
+    break;
+  case TYPE_LDAP:
+    user_source_type = "ldap";
+    break;
+  case TYPE_NONE:
+    user_source_type = "none";
+    break;
+  default:
+    user_source_type = "none";
+    break;
+  }
+  encode_json("type", user_source_type, f);
 }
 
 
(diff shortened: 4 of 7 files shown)


pritha-srivastava:
Things pending:
1. Find appropriate files for class definitions.
2. The old code for ldap handling has not been removed.
3. The ACL bit is a little unclear to me.

src/rgw/rgw_rest_s3.cc
```diff
@@ -4342,3 +4345,88 @@ RGWOp* RGWHandler_REST_Service_S3Website::get_obj_op(bool get_data)
...
+      ldh->bind();
+    }
+  }
+}
+
+RGWRemoteAuthApplier::acl_strategy_t RGWLDAPAuthEngine::get_acl_strategy() const
+{
+  //This is based on the assumption that the default acl strategy in
+  // get_perms_from_aclspec, will take care. Extra acl spec is not required.
+  return nullptr;
```
mattbenjamin: I think that -should- be true

src/rgw/rgw_rest_s3.cc
```diff
@@ -4342,3 +4345,88 @@ RGWOp* RGWHandler_REST_Service_S3Website::get_obj_op(bool get_data)
...
+    token = s->info.args.get("AWSAccessKeyId");
+  } else {
+    string auth_str(s->http_auth + 4);
+    int pos = auth_str.rfind(':');
+    if (pos >=0 ) {
+      token = auth_str.substr(0, pos);
+    }
+  }
+  return token;
+}
```
mattbenjamin: github is warning--need newline at end of file

mattbenjamin:
At a high level, this looks very good.  Class definitions can go wherever, we need to support building with and without LDAP support.  The code not removed is the actual auth hooks?  Does this code actually run, then?  I think the current ACL handling is correct, but :)

pritha-srivastava:
@mattbenjamin : This code runs. I replaced the logic in authorize_v2 with something similar to what is there for the swift authorize method and tested it, though I removed the test code before creating the PR.

pritha-srivastava:
@mattbenjamin : The code not removed is the piece which is there in authorize_v2. That bit can be removed when we have all the S3 AuthEngines in place. And the hooks like RGW_Auth_S3::init and init_impl and the respective variables.

pritha-srivastava:
@mattbenjamin : The code to build with and without ldap support is in rgw_ldap.h ( #if defined(HAVE_OPENLDAP)) . Is this what you were looking for or you meant something else?

pritha-srivastava:
Added code for specifying 'type' with the user. Also replaced some parts of the old code with the new AuthEngine and AuthApplier pieces.

src/rgw/rgw_auth.cc
```diff
@@ -149,6 +149,11 @@ void RGWRemoteAuthApplier::create_account(const rgw_user& acct_user,
 {
   rgw_user new_acct_user = acct_user;
 
+  if (!info.acct_type.empty()) {
```
mattbenjamin: cool

src/rgw/rgw_auth.h
```diff
@@ -117,16 +117,19 @@ class RGWRemoteAuthApplier : public RGWAuthApplier {
     const std::string acct_name;
     const uint32_t perm_mask;
     const bool is_admin;
+    const std::string acct_type;
```
mattbenjamin: is std::string the best type for this?  enum class, maybe?

src/rgw/rgw_common.h
```diff
@@ -677,6 +680,7 @@ struct RGWUserInfo
     }
     if (struct_v >= 18) {
       ::decode(admin, bl);
+      ::decode(type, bl);
```
mattbenjamin: for enum class maybe wrap it in a class then use the macros?

src/rgw/rgw_rest_s3.cc
```diff
@@ -3936,52 +3939,48 @@ int RGW_Auth_S3::authorize_v2(RGWRados *store, struct req_state *s)
...
-      << store->ctx()->_conf->rgw_ldap_uri
-      << dendl;
+  public:
+    S3AuthFactory(RGWRados * const store,
+                     const std::string& acct_override)
+      : store(store),
+        acct_override(acct_override) {
+    }
 
-    RGWToken token{from_base64(auth_id)};
```
mattbenjamin: ok

src/rgw/rgw_rest_s3.cc
```diff
@@ -3936,52 +3939,48 @@ int RGW_Auth_S3::authorize_v2(RGWRados *store, struct req_state *s)
...
-	  }
-	}
+  RGWLDAPAuthEngine ldap(s->cct, store, token_extr, &aplfact);
 
-      /* set request perms */
-      s->perm_mask = RGW_PERM_FULL_CONTROL;
-      } /* success */
-    } /* token */
-  } /* ldap */
+  if (!ldap.is_applicable()) {
```
mattbenjamin: this looks great

src/rgw/rgw_rest_s3.cc
```diff
@@ -4342,3 +4341,98 @@ RGWOp* RGWHandler_REST_Service_S3Website::get_obj_op(bool get_data)
...
+  return RGWRemoteAuthApplier::AuthInfo {
+    rgw_user(token.id),
+    token.id,
+    RGW_PERM_FULL_CONTROL,
+    true,
+    "ldap"
+  };
+}
+
+bool RGWLDAPAuthEngine::is_applicable() const noexcept
```
mattbenjamin: so much better than what we had

src/rgw/rgw_rest_s3.cc
```diff
@@ -4342,3 +4341,98 @@ RGWOp* RGWHandler_REST_Service_S3Website::get_obj_op(bool get_data)
...
+
+  if (ldh->auth(base64_token.id, base64_token.key) != 0) {
+    return nullptr;
+  }
+
+  return apl_factory->create_apl_remote(cct, get_acl_strategy(), get_creds_info(base64_token));
+}
+
+std::string RGWLDAPTokenExtractor::get_token() const
+{
```
mattbenjamin: also great

src/rgw/rgw_rest_s3.h
```diff
@@ -627,4 +630,48 @@ static inline int valid_s3_bucket_name(const string& name, bool relaxed=false)
...
+      store(store),
+      apl_factory(apl_factory) {
+      init(cct);
+      base64_token = rgw::from_base64(token);
+  }
+  const char* get_name() const noexcept override {
+    return "RGWLDAPAuthEngine";
+  }
+  bool is_applicable() const noexcept override;
+  RGWAuthApplier::aplptr_t authenticate() const override;
```
mattbenjamin: just pointlessly, "aplptr_t" is not a great type/name--I know it's part of the interface...

src/rgw/rgw_token.h
```diff
@@ -71,7 +71,7 @@ namespace rgw {
 
     virtual uint32_t version() const { return 1; };
 
-    bool valid() {
+    bool valid() const{
```
mattbenjamin: Radek was here? :)

mattbenjamin:
The only significant point from my review is, I think, maybe preferring a non-string (perhaps an enumeration type) for account_type?  Past that, is this ready to test?

pritha-srivastava:
@mattbenjamin : Reworked the 'type' related code based on your comment. Also, replaced other pieces of 'old' ldap code with the new one. And now have completely removed the old ldap code. This is ready to be tested.

src/rgw/rgw_common.h
```diff
@@ -527,6 +527,14 @@ void encode_json(const char *name, const RGWUserCaps& val, Formatter *f);
 
 void decode_json_obj(obj_version& v, JSONObj *obj);
 
+enum RGWUserType
+{
+  TYPE_NONE=0,
+  TYPE_KEYSTONE,
```
yehudasa: @pritha-srivastava I'd put rgw first, then keystone and then ldap. Also, would make the value explicit (even though probably redundant)

src/rgw/rgw_common.h
```diff
@@ -527,6 +527,14 @@ void encode_json(const char *name, const RGWUserCaps& val, Formatter *f);
 
 void decode_json_obj(obj_version& v, JSONObj *obj);
 
+enum RGWUserType
```
yehudasa: a question: is it user type, or user source type?

src/rgw/rgw_common.h
```diff
@@ -605,6 +615,7 @@ struct RGWUserInfo
      ::encode(user_quota, bl);
      ::encode(user_id.tenant, bl);
      ::encode(admin, bl);
+     ::encode(type, bl);
```
yehudasa: need to bump up the encode version in ENCODE_START()

src/rgw/rgw_common.h
```diff
@@ -677,6 +688,7 @@ struct RGWUserInfo
     }
     if (struct_v >= 18) {
       ::decode(admin, bl);
```
yehudasa: `if (struct_v >= 19) {`

src/rgw/rgw_rest_s3.cc
```diff
@@ -1705,11 +1705,13 @@ int RGWPostObj_ObjStore_S3::get_policy()
       err_msg = "Missing signature";
       return -EINVAL;
     }
-
     RGWUserInfo user_info;
 
     op_ret = rgw_get_user_info_by_access_key(store, s3_access_key, user_info);
     if (op_ret < 0) {
+      S3AuthFactory aplfact(store, s->account_name);
+      RGWLDAPTokenExtractor token_extr(s);
```
yehudasa: I don't have the larger context of the code, but should we define an RGWLDAPTokenExtractor object if ldap was not configured to be used? maybe different scoping is possible?

src/rgw/rgw_common.h
```diff
@@ -547,14 +555,16 @@ struct RGWUserInfo
   RGWQuotaInfo bucket_quota;
   map<int, string> temp_url_keys;
   RGWQuotaInfo user_quota;
+  uint32_t type;
```
yehudasa: @pritha-srivastava also need to update the RGWUserInfo json encoding and decoding

src/rgw/rgw_rest_s3.cc
pritha-srivastava: RGWLDAPAuthEngine::applicable has all the logic to determine whether ldap is correctly configured to be used or not including checking the configurable. Moreover the extractor will be common to ldap, keystone and local auth, so may end up eventually getting renamed and one object of the extractor will be defined for all three auth engines.(This depends on the review of local auth rework code.)

pritha-srivastava:
@yehudasa : I have made changes related to the user source type. Can you please re-review.

src/rgw/rgw_common.h
yehudasa: @pritha-srivastava should declare it as RGWUserSourceType?

src/rgw/rgw_json_enc.cc
```diff
@@ -427,6 +427,7 @@ void RGWUserInfo::dump(Formatter *f) const
   encode_json("bucket_quota", bucket_quota, f);
   encode_json("user_quota", user_quota, f);
   encode_json("temp_url_keys", temp_url_keys, f);
+  encode_json("type", type, f);
```
yehudasa: @pritha-srivastava maybe json encode this enum as a string, not as an int (e.g., look at rgw_meta_sync_info)

src/rgw/rgw_common.h
pritha-srivastava: I have left this as uint32_t since:
1. decode doesn't recognize enums
2. I followed the code for rgw_meta_sync_info and there also the type is uint.

mattbenjamin:
This will need a rebase due to refactoring for rebind().  I can do it, if you like?

mattbenjamin:
I've verified the LDAP auth behavior, including rebind.  The one potential problem is related to the new stored user type:  external-auth users already present in a cluster (e.g., RHCS 2.0) will be unable to authenticate.  I worked around this by removing said users.  We need a ruling on whether we need any accommodation (e.g., set user type?)

PR #10309, opened 2016-07-15
mon: OSDMonitor add check only concern our self cluster command

songbaisen:
mon: OSDMonitor add check only concern our self cluster command

we only concern the command send to ourself cluster

### DIFF
--- a/src/mon/OSDMonitor.cc
+++ b/src/mon/OSDMonitor.cc
@@ -7703,6 +7703,14 @@ bool OSDMonitor::preprocess_pool_op(MonOpRequestRef op)
 {
   op->mark_osdmon_event(__func__);
   MPoolOp *m = static_cast<MPoolOp*>(op->get_req());
+  
+  if (m->fsid != mon->monmap->fsid) {
+    dout(0) << __func__ << " drop message on fsid " << m->fsid
+            << " != " << mon->monmap->fsid << " for " << *m << dendl;
+    _pool_op_reply(op, -EINVAL, osdmap.get_epoch());
+    return true;
+  }
+
   if (m->op == POOL_OP_CREATE)
     return preprocess_pool_op_create(op);
 


dachary:
jenkins test this please

src/mon/OSDMonitor.cc
```diff
@@ -7703,6 +7703,13 @@ bool OSDMonitor::preprocess_pool_op(MonOpRequestRef op)
 {
   op->mark_osdmon_event(__func__);
   MPoolOp *m = static_cast<MPoolOp*>(op->get_req());
+  
+  if (m->fsid != mon->monmap->fsid) {
+    dout(0) << "preprocess_pool_op: on fsid " << m->fsid
+            << " != " << mon->monmap->fsid << dendl;
+    return true;
```
jecluis: I'm not entirely sure what the consequences of returning an error here would be, but silently dropping the message doesn't feel right.

We should figure out if there's an appropriate error to be replied to the consumer, and if so, make sure to use it.

src/mon/OSDMonitor.cc
```diff
@@ -7703,6 +7703,13 @@ bool OSDMonitor::preprocess_pool_op(MonOpRequestRef op)
 {
   op->mark_osdmon_event(__func__);
   MPoolOp *m = static_cast<MPoolOp*>(op->get_req());
+  
+  if (m->fsid != mon->monmap->fsid) {
+    dout(0) << "preprocess_pool_op: on fsid " << m->fsid
```
jecluis: also, if we are outputting this error message on level 0, would certainly be nice to know which message is being dropped. I'd recommend fitting the message in here as well.

songbaisen: @jecluis Yes good idea sir.

src/mon/OSDMonitor.cc
songbaisen: @jecluis  Agree with that. I will add it later.

songbaisen:
@jecluis  Done it.

src/mon/OSDMonitor.cc
tchaikov: passing an error string via the `MPoolOpReply ::response_data` is a viable way, but we need to change the interface of librados to return the error string. so returning an error code is fine here in the updated commit.

yuriw:
http://pulpito.front.sepia.ceph.com:80/yuriw-2017-02-06_01:16:11-rados-wip-yuri-testing_2017_2_6-distro-basic-smithi/

http://pulpito.ceph.com/yuriw-2017-02-06_15:38:04-rados-wip-yuri-testing_2017_2_6---basic-smithi/

added infra fix  ceph/ceph: Pull Request 13279

http://pulpito.ceph.com/yuriw-2017-02-06_23:30:25-rados-wip-yuri-testing_2017_2_6---basic-smithi/

http://pulpito.front.sepia.ceph.com:80/yuriw-2017-02-07_16:02:15-rados-wip-yuri-testing_2017_2_6-distro-basic-smithi/

yuriw:
@liewegas this was tested pls merge

PR #10310, opened 2016-07-15
[DNM] hammer: mon/Monitor: clear routed_requests after resend

aiicore:
When monitor which is not the leader, gets message from
e.g. MDS, it routes it to leader, but messages from
routed_requests are not deleted. In hazard cases, such as
bad network connection, when monitors have problem with
forming a quorum, routed_requests are getting quite fat.
Monitor::resend_routed_requests() sends all messages each time,
old and newly added. If this hazard state is long enough,
leader is constantly flooded by those routed messages.

Also entity_name wasn't set when sending routed requests.

Signed-off-by: Igor Podoski <igor.podoski@ts.fujitsu.com>

### DIFF
--- a/src/mon/Monitor.cc
+++ b/src/mon/Monitor.cc
@@ -3203,20 +3203,21 @@ void Monitor::resend_routed_requests()
       dout(10) << " requeue for self tid " << rr->tid << " " << *req << dendl;
       req->set_connection(rr->con);
       retry.push_back(new C_RetryMessage(this, req));
-      delete rr;
     } else {
       dout(10) << " resend to mon." << mon << " tid " << rr->tid << " " << *req << dendl;
       MForward *forward = new MForward(rr->tid, req, rr->con_features,
 				       rr->session->caps);
       forward->client = rr->client_inst;
       forward->set_priority(req->get_priority());
+      forward->entity_name = rr->session->entity_name;
       messenger->send_message(forward, monmap->get_inst(mon));
     }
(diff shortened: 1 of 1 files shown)


aiicore:
Thanks to this routed request flood, discover of memory leak in #10238 was possible. I don't know if not clearing those request are part of something bigger e.g. when the "broken" leader (0) will eventually got down and another monitor will became a leader (2), all historical routed request from routing monitor (1) will be resend to new leader (2). Correct me if I'm wrong.

smithfarm:
Adding DNM just to keep this very new PR out of the next hammer integration testing round. Will remove it afterwards.

smithfarm:
@aiicore Have you checked if this bug is present in jewel or master?

aiicore:
@smithfarm This code lasts until v10.0.5. There was big MDSMonitor reorganization in 4e9b9539f16de15f4c3fdcb7b33625365822e235, and above code was removed.

aiicore:
@smithfarm, looking on the code it could exists in master too. I'll move this PR to the master branch.

smithfarm:
@aiicore Thanks. Once you have verified that the issue is fixed in master, we can look at backporting the fix to the stable releases (jewel and hammer).

aiicore:
@smithfarm Closing this for now on hammer branch. Potentially I would make PR to master since this behavior exist there too, but I need to check one more time if not clearing routed packets is necessary in some case.

PR #10315, opened 2016-07-15
pybind: Rework cephfs/setup.py for PyPI

onyb:
Also use `set_include_dirs` in `distutils.ccompiler` instead of `add_include_dir` for supporting non-standard include directories.

### DIFF
--- a/src/pybind/cephfs/setup.py
+++ b/src/pybind/cephfs/setup.py
@@ -1,56 +1,194 @@
-# Largely taken from
-# https://blog.kevin-brown.com/programming/2014/09/24/combining-autotools-and-setuptools.html
-import os, sys, os.path
+from __future__ import print_function
 
-from setuptools.command.egg_info import egg_info
-from distutils.core import setup
-from distutils.extension import Extension
-from Cython.Build import cythonize
-from Cython.Distutils import build_ext
+import os
+import pkgutil
+import shutil
+import subprocess
+import sys
+import tempfile
+import textwrap
+from distutils.ccompiler import new_compiler
+from distutils.errors import CompileError, LinkError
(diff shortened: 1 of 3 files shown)


src/pybind/cephfs/setup.py
```diff
@@ -1,56 +1,196 @@
...
+except ImportError:
+    print("WARNING: Cython is not installed.")
+
+    if not os.path.isfile('cephfs.c'):
+        print('ERROR: Cannot find Cythonized file cephfs.c', file=sys.stderr)
+        sys.exit(1)
+    else:
+        def cythonize(x, **kwargs):
+            return x
+
```
tchaikov: please remove at least one blank line here.

and could you run this script thru pep8 to kill some warnings.

tchaikov:
lgtm other than a pep8 nit.

onyb:
@tchaikov Thanks. PEP8ified.

jdurgin:
@onyb looks like the build bot failure is real - it fails to find librados.h with a cmake build (happens on my machine too).

If you haven't built with cmake yet, you should start using it - we're just about all switched over. ceph.git's README.md has updated info on using it.

onyb:
@jdurgin Fixed the issue. I've tested this with both CMake and configure.

jcsp:
This didn't cause any issues in this run: http://pulpito.ceph.com/jspray-2016-07-19_05:56:13-fs-wip-jcsp-testing-20160719-distro-basic-mira/

Everyone happy to merge?

jdurgin:
@jcsp I'm happy to merge

dillaman:
Looks like the doc builders are broken -- the include paths from the environment are no longer processed.

PR #10316, opened 2016-07-16
mds: add assertions for standby_daemons invariant

batrick:
These assertions catch state changes of an mds in standby_daemons to a state
other than MDSMap::STATE_STANDBY. Currently this invariant is (sometimes!)
checked in other locations on access of standby_daemons. This commit allows us
catch the violated invariant at the time it occurred.
    
Related to: http://tracker.ceph.com/issues/16592
    

jcsp:
Reviewed-by, will include in next test branch

PR #10319, opened 2016-07-16
src/test/cli/* tests: POSIX Convert grep -P to grep -E

wjwithagen:
- FreeBSD does not have grep -P

### DIFF
--- a/src/test/cli/ceph-authtool/cap-invalid.t
+++ b/src/test/cli/ceph-authtool/cap-invalid.t
@@ -3,10 +3,10 @@
 
 # TODO is this nice?
   $ ceph-authtool --cap osd 'broken' kring
-  $ ceph-authtool kring --list|grep -P '^\tcaps '
+  $ ceph-authtool kring --list|grep -E '^[[:space:]]caps '
   \tcaps osd = "broken" (esc)
 
 # TODO is this nice?
   $ ceph-authtool --cap xyzzy 'broken' kring
-  $ ceph-authtool kring --list|grep -P '^\tcaps '
+  $ ceph-authtool kring --list|grep -E '^[[:space:]]caps '
   \tcaps xyzzy = "broken" (esc)
--- a/src/test/cli/ceph-authtool/cap-overwrite.t
+++ b/src/test/cli/ceph-authtool/cap-overwrite.t
@@ -2,10 +2,10 @@
   creating kring
 
   $ ceph-authtool --cap osd 'allow rx pool=swimming' kring
-  $ ceph-authtool kring --list|grep -P '^\tcaps '
+  $ ceph-authtool kring --list|grep -E '^[[:space:]]caps '
   \tcaps osd = "allow rx pool=swimming" (esc)
 
 # TODO it seems --cap overwrites all previous caps; is this wanted?
   $ ceph-authtool --cap mds 'allow' kring
-  $ ceph-authtool kring --list|grep -P '^\tcaps '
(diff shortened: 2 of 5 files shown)


src/test/cli/ceph-authtool/cap-overwrite.t
```diff
@@ -2,10 +2,10 @@
   creating kring
 
   $ ceph-authtool --cap osd 'allow rx pool=swimming' kring
-  $ ceph-authtool kring --list|grep -P '^\tcaps '
+  $ ceph-authtool kring --list|grep -E '^\Wcaps '
```
tchaikov: i don't think it's necessary to loosen the check for tab.

wjwithagen:
@tchaikov 
This more to your liking?

src/test/cli/ceph-authtool/cap-invalid.t
```diff
@@ -3,10 +3,10 @@
 
 # TODO is this nice?
   $ ceph-authtool --cap osd 'broken' kring
-  $ ceph-authtool kring --list|grep -P '^\tcaps '
+  $ ceph-authtool kring --list|grep -E '^[^I]caps '
```
tchaikov: i never knew that ERE supported the syntax like this! thought it's a character class excluding capital "I"! and seems the spec does not mention this usage, see http://pubs.opengroup.org/onlinepubs/000095399/basedefs/xbd_chap09.html#tag_09_03_05 .

wjwithagen: On 20-7-2016 12:37, Kefu Chai wrote:
> In src/test/cli/ceph-authtool/cap-invalid.t
> <https://github.com/ceph/ceph/pull/10319#discussion_r71502243>:
> 
>> @@ -3,10 +3,10 @@
>>  
>>  # TODO is this nice?
>>    $ ceph-authtool --cap osd 'broken' kring
>> -  $ ceph-authtool kring --list|grep -P '^\tcaps '
>> +  $ ceph-authtool kring --list|grep -E '^[^I]caps '
> 
> i never knew that ERE supported the syntax like this! thought it's a
> character class excluding capital "I"! and seems the spec does not
> mention this usage, see
> http://pubs.opengroup.org/onlinepubs/000095399/basedefs/xbd_chap09.html#tag_09_03_05

It is one of the few versions that I actually got to work with grep -E
Took me a while to get something that works on all ends and pieces.
So I'm very curious what a teuthology runs will bring.

tchaikov: yeah, it works for me, anyway.

tchaikov:
lgtm.

wjwithagen:
On 20-7-2016 13:01, Kefu Chai wrote:
> lgtm.

Looked over it again, and it is definitly not correct. :(
Replace ^I with ^K and it still works.
Back to the drawing board, because ERE does not seem to take the
standard escapes.

So I'm very tempted to use [[:space:]], that would allow either
<space> of <tab>.
So less liberal than \W

wjwithagen:
@tchaikov 
Jenkins likes this too, and that is really according the Posix standard.
Catches blanks, and some other spacing stuff. But not too liberal.

tchaikov:
tested at

- http://pulpito.ceph.com/kchai-2016-07-24_02:17:13-rados-wip-kefu-testing---basic-mira/
- http://pulpito.ceph.com/kchai-2016-07-24_20:41:33-rados-wip-kefu-testing---basic-mira/

wjwithagen:
@tchaikov 
Thanx

PR #10322, opened 2016-07-17
debian: python related changes

tchaikov:
@dmick could you take a look at this PR also? thanks.


debian/rules
```diff
@@ -134,6 +134,14 @@ override_dh_strip:
 override_dh_shlibdeps:
 	dh_shlibdeps -a --exclude=erasure-code --exclude=rados-classes --exclude=compressor
 
+override_dh_python2:
+	for binding in rados cephfs rbd; do                                \
+	  dh_python2 -O--buildsystem=python_distutils -p python-$$binding; \
+	done
+	dh_python2 -p ceph-common
+	dh_python2 -O--buildsystem=python_distutils -p ceph-base
```
dmick: why --buildsystem=python_distutils?  and why not for ceph-common?

dmick:
the build failure is unconnected, most likely, and can be fixed with a rebase I'm assuming

tchaikov:
@dmick ping.

debian/rules
```diff
@@ -134,6 +134,14 @@ override_dh_strip:
 override_dh_shlibdeps:
```
dmick: the commit message confuses me; if the warning as about running dh_python2 against python-rbd, the code below explicitly does that; did you mean to have a different package name in the warning above?...

debian/control
```diff
@@ -53,7 +53,7 @@ Build-Depends: btrfs-tools,
                pkg-config,
                python (>= 2.6.6-3~),
                python-argparse | libpython2.7-stdlib,
-               python-dev,
+               python-all-dev,
```
dmick: Should this be ${python:<something>} anyway?

debian/rules
tchaikov: @dmick sorry for the misleading example, yeah, lemme copy another warning message.

debian/control
tchaikov: no, ${python:`<something>`} subst vars are generated by dh_python*, it by itself is a build dependency listed as "dh-python". so it's like an problem of egg-and-chicken or self-bootstrap. we can hardly let dpkg-buildpackage to figure out what the build-dependency is before the build-dependencies are ready.

tchaikov:
@dmick thanks for reviewing, fixed the commit message, rebased against master, and repushed.

debian/control
dmick: got it.

PR #10326, opened 2016-07-18
ReplicatedPG: EC async read need ondisk read lock.

cxwshawn:
unlike replicated mode, int result = prepare_transaction(ctx); can actually read data and return,
so before prepare_transaction: obc->ondisk_read_lock() and after prepare_transaction: ondisk_read_unlock() is ok, but EC is async read, only lock and unlock prepare_transaction is not enough.


cxwshawn:
@liewegas @athanatos please help review, thanks!

athanatos:
The ondisk.*lock machinery is actually almost completely useless (I have a branch kicking around which removes it entirely).  There is a completely different mechanism for controlling access to objects which doesn't involve blocking a thread (See ObjectContext::RWState).

PR #10328, opened 2016-07-18
Create ceph-mgr

jcsp:
(See ceph-mgr threads on ceph-devel for background, and docs in this PR for the python interface)

This PR includes the core functionality for enabling python modules that provide services (such as the 'rest' module here derived from calamari) and expose CLI hooks.  Python modules can also store persistent pieces of data, which ceph-mgr stashes using the existing "config-key" mon mechanism.

The Mgr class and its encapsulated classes gather the cluster maps, the daemon metadata, and performance counters, and then expose all these to the python modules.

We have a new MgrMonitor class in the mon for tracking the up ceph-mgr daemons and publishing their addresses for consumption by MgrClient instances, such as those embedded in OSDs for sending telemetry to the mgrs.  The perf counters from OSDs are currently sent to mgrs.

There is no auth cap enforcement for access to the mgr yet, that still needs inventing.  The rest module also stores its authentication credentials in config-keys which need some greater restriction on access (either build an alternative store into MgrMonitor, or give the config-key mechanism some way of limiting who can read certain keys).

The ceph-mgr service is built with cmake, but there is no packaging yet.  Add a "MGR=1" environment variable to vstart to have it start an instance for you.  Use "ceph tell mgr enable_auth false" to switch off authentication and point your browser at localhost:8002/api/v2/cluster/osd to see the REST output.


src/pybind/mgr/calamari_rest/management/commands/api_docs.py
```diff
@@ -0,0 +1,406 @@
+#
+# If Django is installed locally, run the following command from the top of
+# the ceph source tree:
+#
+#   PYTHONPATH=src/pybind/mgr \
+#   DJANGO_SETTINGS_MODULE=calamari_rest.settings \
+#   CALAMARI_CONFIG=src/pybind/mgr/calamari.conf \
+#       django-admin api_docs
+#
+# This will create resources.rst (the API docs), and rest.log (which will
```
alfredodeza: This deviates from the standard of using Sphinx to generate documentation from code by introducing custom mechanisms to introspect the code and produce docs per url endpoints. Although fine for a first pass, I don't think it helps in the long term when contributors are faced with a custom system like this.

Sphinx already has introspection and can populate sections in RST files (although it will not create a url-to-api-doc mapping), those mechanisms are well known (well tested!) and documented thoroughly.

Perhaps even more positive than the familiarity a Python developer (or some other contributor) might find with Sphinx, it is the ability to extend and customize using Sphinx (vs. a custom Django command).

src/pybind/mgr/calamari_rest/manager/user_request.py
```diff
@@ -0,0 +1,553 @@
+import json
+import logging
+import uuid
+
+from calamari_rest.types import OsdMap, PgSummary, USER_REQUEST_COMPLETE, USER_REQUEST_SUBMITTED
+from calamari_rest.util import now
+from mgr_module import CommandResult
+
+from rest import logger
+log = logger()
```
alfredodeza: this form of instantiating the logger instance will use the `root` logger. The caveat of doing this is that any log statements in this file will *not* have the ability to:

* show where the log statement is originating from (e.g. `calamari_rest.manager.user_request`)
* be configured differently than the default root logger

src/pybind/mgr/calamari_rest/manager/user_request.py
```diff
@@ -0,0 +1,553 @@
...
+            return "Completed successfully"
+
+    @property
+    def awaiting_versions(self):
+        """
+        Requests indicate that they are waiting for particular sync objects, optionally
+        specifying the particular version they are waiting for (otherwise set version
+        to None).
+
+        :return dict of SyncObject subclass to (version or None)
```
alfredodeza: s/:return/:return:/

src/pybind/mgr/calamari_rest/management/commands/api_docs.py
jcsp: This part is carried forward from the Calamari tree -- revisiting how docs are generated would be welcome if you or anyone else is interested in working on it.

tserong: In my completely biased opinion (what with me being the one who massaged the calamari docs generator into a somewhat functional state for this PR), I'd prefer to leave it as-is now, and fix it later. I am also happy to be the one to attempt to fix it later :-)

src/mgr/PyModules.cc
```diff
@@ -0,0 +1,456 @@
...
+                              std::string> &new_config)
+{
+  Mutex::Locker l(lock);
+
+  dout(4) << "Loaded " << new_config.size() << " config settings" << dendl;
+  config_cache = new_config;
+}
+
+void PyModules::log(const std::string &handle,
+    unsigned level, const std::string &record)
```
tserong: With level being unsigned, I get a build failure:

````
/home/tserong/src/github/SUSE/ceph/src/mgr/PyModules.cc: In member function ‘void PyModules::log(const string&, unsigned int, const string&)’:
/home/tserong/src/github/SUSE/ceph/src/common/dout.h:57:24: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
       char __array[((v >= -1) && (v <= 200)) ? 0 : -1] __attribute__((unused)); \
                         
/home/tserong/src/github/SUSE/ceph/src/common/dout.h:57:24: note: in definition of macro ‘dout_impl’
       char __array[((v >= -1) && (v <= 200)) ? 0 : -1] __attribute__((unused)); \
                        ^~
/home/tserong/src/github/SUSE/ceph/src/common/debug.h:23:17: note: in expansion of macro ‘ldout’
 #define dout(v) ldout((g_ceph_context), v)
                 ^~~~~
/home/tserong/src/github/SUSE/ceph/src/mgr/PyModules.cc:453:3: note: in expansion of macro ‘dout’
   dout(level) << record << dendl;
   ^
/home/tserong/src/github/SUSE/ceph/src/common/dout.h:57:78: error: size of array ‘__array’ is negative
       char __array[((v >= -1) && (v <= 200)) ? 0 : -1] __attribute__((unused)); \
                                                                              ^
/home/tserong/src/github/SUSE/ceph/src/common/dout.h:66:24: note: in expansion of macro ‘dout_impl’
 #define ldout(cct, v)  dout_impl(cct, dout_subsys, v) dout_prefix
                        ^~~~~~~~~
/home/tserong/src/github/SUSE/ceph/src/common/debug.h:23:17: note: in expansion of macro ‘ldout’
 #define dout(v) ldout((g_ceph_context), v)
                 ^~~~~
/home/tserong/src/github/SUSE/ceph/src/mgr/PyModules.cc:453:3: note: in expansion of macro ‘dout’
   dout(level) << record << dendl;
   ^
````

Changing level to an int here (and also in PyModules.h) fixes it.

**Comment by tserong** [comment]
Just an addendum to @jcsp's initial comment - if you want to play with ceph-mgr (say, via vstart), you need django, python-dateutil, gevent and djangorestframework installed. The first three of these are somewhat obvious (you'll see an ImportError in the mgr log for each of those if they're not present), but that doesn't seem to be the case for djangorestframework. You'll just end up with nothing listening on port 8002, and not know why.

TL;DR: do this:

`pip2 install django==1.8 python-dateutil gevent djangorestframework`

(The version restriction on django may or may not be necessary -- I had trouble with 1.9 a while ago, so went with 1.8, and haven't yet re-tested with 1.9)

File: src/mon/OSDMonitor.cc [hunk 5]
```diff
@@ -3321,7 +3321,6 @@ bool OSDMonitor::preprocess_command(MonOpRequestRef op)
             // Drop error, continue to get other daemons' metadata
             dout(4) << "No metadata for osd." << i << dendl;
             r = 0;
-            continue;
```
tchaikov: nice catch!

src/common/common_init.cc
```diff
@@ -68,6 +68,8 @@ CephContext *common_preinit(const CephInitParameters &iparams,
       conf->set_val("keyring", "$mds_data/keyring", false);
     else if (conf->name.is_osd())
       conf->set_val("keyring", "$osd_data/keyring", false);
+    else if (conf->name.is_mgr())
```
tchaikov: can we just hardwire the path of keyring of ceph-mgr to "$mgr_data/keyring" like mon, see https://github.com/ceph/ceph/blob/master/src/mon/Monitor.cc#L709 ? as we don't need to be backward compatible with previous releases here.

src/common/CommandTable.h
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  const std::map<ceph_tid_t, T*> &get_commands() const
+  {
+    return commands;
+  }
+
+  T* get_command(ceph_tid_t tid)
+  {
+    if (commands.count(tid) == 0) {
```
tchaikov: maybe we can avoid the overhead of `count()` and `at()`? and just use `commands.find(tid)` instead?

src/common/CommandTable.h
```diff
@@ -0,0 +1,90 @@
...
+      return nullptr;
+    } else {
+      return commands.at(tid);
+    }
+  }
+
+  void erase(ceph_tid_t tid)
+  {
+    auto ptr = commands.at(tid);
```
tchaikov: ditto.

src/common/CommandTable.h
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  CommandOp(const ceph_tid_t t) : tid(t) {}
+};
+
+template<typename T>
+class CommandTable
+{
+protected:
+  ceph_tid_t last_tid;
```
tchaikov: so we can remove `Client::last_tid`, right?

src/common/cmdparse.cc
```diff
@@ -14,13 +14,37 @@
...
+ * return the prefix "foo baz".
+ */
+std::string cmddesc_get_prefix(const std::string &cmddesc)
+{
+  stringstream ss(cmddesc);
+  std::string word;
+  std::ostringstream result;
+  bool first = true;
+  while (std::getline(ss, word, ' ')) {
+    if (word.find_first_of(",=") == string::npos) {
```
tchaikov: could just break if `word.find_first_of(",=") != word.npos`. or we we can just 
```c++
return cmddesc.substr(0, cmddesc.rfind(' '));
```
assuming there is no space in the parameter list spec.

src/mon/MgrMonitor.cc
```diff
@@ -0,0 +1,241 @@
...
+void MgrMonitor::create_initial()
+{
+}
+
+void MgrMonitor::update_from_paxos(bool *need_bootstrap)
+{
+  version_t version = get_last_committed();
+  if (version == map.epoch) {
+    return;
+  };
```
tchaikov: could remove the `;` here.

src/mon/MgrMonitor.cc
```diff
@@ -0,0 +1,241 @@
...
+ * Handle digest subscriptions separately (outside of check_sub) because
+ * they are going to be periodic rather than version-driven.
+ */
+void MgrMonitor::send_digests()
+{
+  const std::string type = "mgrdigest";
+  if (mon->session_map.subs.count(type) == 0)
+    return;
+  xlist<Subscription*>::iterator p = mon->session_map.subs[type]->begin();
+  while (!p.end()) {
```
tchaikov: nit, could use range based loop also here.

src/mon/MgrMonitor.cc
```diff
@@ -0,0 +1,241 @@
...
+}
+
+/**
+ * Handle digest subscriptions separately (outside of check_sub) because
+ * they are going to be periodic rather than version-driven.
+ */
+void MgrMonitor::send_digests()
+{
+  const std::string type = "mgrdigest";
+  if (mon->session_map.subs.count(type) == 0)
```
tchaikov: might want to avoid `count()` plus `[]`.

src/CMakeLists.txt
```diff
@@ -565,6 +566,24 @@ add_subdirectory(include)
...
+      mgr/PyModules.cc
+      mgr/PyFormatter.cc
+      mgr/PyState.cc
+      mgr/MgrPyModule.cc
+      mgr/Mgr.cc)
+  add_executable(ceph-mgr ${mgr_srcs}
+                 $<TARGET_OBJECTS:heap_profiler_objs>)
+             set_target_properties(ceph-mgr PROPERTIES COMPILE_FLAGS "-I${PYTHON_INCLUDE_DIRS}")
+  target_link_libraries(ceph-mgr mds osdc global
+      -lboost_python ${PYTHON_LIBRARIES} ${BLKID_LIBRARIES} ${CMAKE_DL_LIBS} ${TCMALLOC_LIBS})
```
tchaikov: shall we check for `libboost_python` in `${CMAKE_SOURCE_DIR}/CMakeLists.txt`?

tchaikov:
will continue the review tmr.

src/common/common_init.cc
liewegas: :+1:

jcsp: I can move the setting of "keyring" out of here and into ceph_mgr.cc, but we can't set it directly the way Monitor does, because we're using the MonClient machinery for our keyring rather than constructing one by hand.

src/common/CommandTable.h
jcsp: done

src/common/CommandTable.h
jcsp: In this instance use of ``at`` is intentional: it is required that the tid passed in exists, and if it doesn't they'll get an exception.

src/common/CommandTable.h
jcsp: No, Client::last_tid is still used for normal MDS requests.

Client previously used the same counter for both normal MDS requests and for commands, now there are two counters (which is OK because all the server side logic is separate for commands vs. normal requests).

src/common/cmdparse.cc
jcsp: The parameter list is indeed space separated, so the rfind idea doesn't work.  Changed this to break when we see a ",=" match as suggested.

src/mon/MgrMonitor.cc
jcsp: done

src/mon/MgrMonitor.cc
jcsp: I don't think xlist doesn't supports full range based loops, but I can make this into a proper for loop for sure

tchaikov: @jcsp, see http://en.cppreference.com/w/cpp/language/range-for, and `xlist` does have `begin()` and `end()`, and a proper iterator implementation. see fc9426a5a11e93bb17bbb944b2555fa6ac4183f3. so i guess it supports range based for loop?

jcsp: Ah, you're right.  Updated.

src/CMakeLists.txt
jcsp: Added a `find_package(Boost COMPONENTS python REQUIRED)`

src/rocksdb
```diff
@@ -1 +1 @@
-Subproject commit 6797e6ffacd90470140f64cea068970bb3a30aaf
+Subproject commit 5647fa427c4eab87c0caf8bcbb91f5846a4623f9
```
tchaikov: might want to revert this change.

jcsp: Oops, good catch.

jcsp:
Building on centos gitbuilders but mysteriously failing on ubuntu

http://gitbuilder.sepia.ceph.com/gitbuilder-ceph-deb-trusty-amd64-basic/log.cgi?log=963f2b91b2a6526d6f2ee7f7c6619e2e73bf4797
```
make -f src/CMakeFiles/ceph-mgr.dir/build.make src/CMakeFiles/ceph-mgr.dir/depend
make[4]: Entering directory `/srv/autobuild-ceph/gitbuilder.git/build/out~/ceph-11.0.0-2069-g963f2b9/obj-x86_64-linux-gnu'
cd "/srv/autobuild-ceph/gitbuilder.git/build/out~/ceph-11.0.0-2069-g963f2b9/obj-x86_64-linux-gnu" && /usr/bin/cmake -E cmake_depends "Unix Makefiles" "/srv/autobuild-ceph/gitbuilder.git/build/out~/ceph-11.0.0-2069-g963f2b9" "/srv/autobuild-ceph/gitbuilder.git/build/out~/ceph-11.0.0-2069-g963f2b9/src" "/srv/autobuild-ceph/gitbuilder.git/build/out~/ceph-11.0.0-2069-g963f2b9/obj-x86_64-linux-gnu" "/srv/autobuild-ceph/gitbuilder.git/build/out~/ceph-11.0.0-2069-g963f2b9/obj-x86_64-linux-gnu/src" "/srv/autobuild-ceph/gitbuilder.git/build/out~/ceph-11.0.0-2069-g963f2b9/obj-x86_64-linux-gnu/src/CMakeFiles/ceph-mgr.dir/DependInfo.cmake" --color=
Scanning dependencies of target ceph-mgr
make[4]: Leaving directory `/srv/autobuild-ceph/gitbuilder.git/build/out~/ceph-11.0.0-2069-g963f2b9/obj-x86_64-linux-gnu'
make -f src/CMakeFiles/ceph-mgr.dir/build.make src/CMakeFiles/ceph-mgr.dir/build
make[4]: Entering directory `/srv/autobuild-ceph/gitbuilder.git/build/out~/ceph-11.0.0-2069-g963f2b9/obj-x86_64-linux-gnu'
/usr/bin/cmake -E cmake_progress_report "/srv/autobuild-ceph/gitbuilder.git/build/out~/ceph-11.0.0-2069-g963f2b9/obj-x86_64-linux-gnu/CMakeFiles"
[ 78%] Building CXX object src/CMakeFiles/ceph-mgr.dir/ceph_mgr.cc.o
cd "/srv/autobuild-ceph/gitbuilder.git/build/out~/ceph-11.0.0-2069-g963f2b9/obj-x86_64-linux-gnu/src" && /usr/bin/x86_64-linux-gnu-g++ -DCEPH_LIBDIR=\"/usr/lib\" -DCEPH_PKGLIBDIR=\"/usr/lib/ceph\" -D__linux__ -g -O2 -fstack-protector --param=ssp-buffer-size=4 -Wformat -Werror=format-security -D_FORTIFY_SOURCE=2 -Wall -Wtype-limits -Wignored-qualifiers -Winit-self -Wpointer-arith -Werror=format-security -fno-strict-aliasing -fsigned-char -rdynamic -Wp,-U_FORTIFY_SOURCE,-D_FORTIFY_SOURCE=2 -ftemplate-depth-1024 -Wno-invalid-offsetof -Wnon-virtual-dtor -Wstrict-null-sentinel -Woverloaded-virtual -std=c++11 -I/usr/include/nss -I/usr/include/nspr -fno-builtin-malloc -fno-builtin-calloc -fno-builtin-realloc -fno-builtin-free -fPIE -I"/srv/autobuild-ceph/gitbuilder.git/build/out~/ceph-11.0.0-2069-g963f2b9/obj-x86_64-linux-gnu/src/include" -I"/srv/autobuild-ceph/gitbuilder.git/build/out~/ceph-11.0.0-2069-g963f2b9/src" -I"/srv/autobuild-ceph/gitbuilder.git/build/out~/ceph-11.0.0-2069-g963f2b9/obj-x86_64-linux-gnu/include" -I"/srv/autobuild-ceph/gitbuilder.git/build/out~/ceph-11.0.0-2069-g963f2b9/src/xxHash" -DHAVE_CONFIG_H -D__CEPH__ -D_REENTRANT -D_THREAD_SAFE -D__STDC_FORMAT_MACROS -D_GNU_SOURCE -D_FILE_OFFSET_BITS=64 -I/usr/include/python2.7;/usr/include/x86_64-linux-gnu/python2.7 -o CMakeFiles/ceph-mgr.dir/ceph_mgr.cc.o -c "/srv/autobuild-ceph/gitbuilder.git/build/out~/ceph-11.0.0-2069-g963f2b9/src/ceph_mgr.cc"
x86_64-linux-gnu-g++: fatal error: no input files
compilation terminated.
/bin/sh: 1: /usr/include/x86_64-linux-gnu/python2.7: Permission denied
make[4]: *** [src/CMakeFiles/ceph-mgr.dir/ceph_mgr.cc.o] Error 126
make[4]: Leaving directory `/srv/autobuild-ceph/gitbuilder.git/build/out~/ceph-11.0.0-2069-g963f2b9/obj-x86_64-linux-gnu'
make[3]: *** [src/CMakeFiles/ceph-mgr.dir/all] Error 2
make[3]: *** Waiting for unfinished jobs....
```

liewegas:
This looks great!  Obviously lots of rough edges, but let's get it merged and go from there.  My main concern is around the bulky MPGStats message the OSDs are generating on every report; this might have a noticeable impact on the OSD.  It'll only come up if the there is a ceph-mgr running, though, so I don't think it's a blocker.

Probably the first thing we should address is the build and package mechanics.  That means adding the cmake bits and #ifdefs to deal with python 2 vs 3 so that we can build on all target platforms...

Does anything have to happen before we merge this into the tree?

jcsp:
This will be safe to merge once I've worked out why the .deb gitbuilders are unhappy

jcsp:
Oh, semicolon in the -I argument.

jcsp:
retest this please

src/pybind/mgr/fsstatus/module.py
```diff
@@ -0,0 +1,274 @@
...
+
+            standby_table.add_row([standby['name']])
+
+        output += "\n" + standby_table.get_string() + "\n"
+
+        if len(mds_versions) == 1:
+            output += "MDS version: {0}".format(mds_versions.keys()[0])
+        else:
+            version_table = PrettyTable(["version", "daemons"])
+            for version, daemons in mds_versions.iteritems():
```
tchaikov: shall we collect `version_table.get_string()` also?

tchaikov:
jenkins run failed due to known issue http://tracker.ceph.com/issues/17052.

src/pybind/mgr/fsstatus/module.py
tchaikov: @jcsp ping.

jcsp: good catch, thanks

src/common/CommandTable.h
```diff
@@ -0,0 +1,91 @@
...
+  std::map<ceph_tid_t, T*> commands;
```
athanatos: If someone misuses this interface, it'll just leak the pointers.  That's not great.  I'd like get_commands() and get_command() (since they give up ownership of the pointer) to remove the returned pointers from the map.  I'd further like the destructor for CommandTable to assert that the map is empty.

src/pybind/mgr/rest/app/views/v2.py
```diff
@@ -0,0 +1,770 @@
...
+
+    def list(self, request):
+        servers = self.client.server_list()
+        return Response(self.serializer_class(
+            [DataObject(s) for s in servers],
+            many=True).data)
+
+
+if False:
+    class EventViewSet(DatabaseViewSet, PaginatedMixin):
```
alfredodeza: `DatabaseViewSet` is undefined here because it is commented out on line 8

alfredodeza: Same with `PaginatedMixin`

src/pybind/mgr/rest/app/views/v2.py
```diff
@@ -0,0 +1,770 @@
...
+
+        def list_cluster(self, request):
+            return Response(self._paginate(request, self._filter_by_severity(request, self.queryset.filter_by(fsid=fsid))))
+
+        def list_server(self, request, fqdn):
+            return Response(self._paginate(request, self._filter_by_severity(request, self.queryset.filter_by(fqdn=fqdn))))
+
+
+if False:
+    class LogTailViewSet(RemoteViewSet):
```
alfredodeza: `RemoteViewSet` is also undefined here

src/pybind/mgr/rest/app/views/v2.py
```diff
@@ -0,0 +1,770 @@
...
+        def get_cluster_log(self, request):
+            """
+            Retrieve the cluster log from one of a cluster's mons (expect it to be in /var/log/ceph/ceph.log)
+            """
+
+            # Number of lines to get
+            lines = request.GET.get('lines', 40)
+
+            # Resolve FSID to name
+            name = self.client.get_cluster(fsid)['name']
```
alfredodeza: Not sure where `fsid` is coming from here, it is also undefined

src/pybind/mgr/rest/app/views/v2.py
```diff
@@ -0,0 +1,770 @@
...
+
+    def list(self, request):
+        mons = self._get_mons()
+        return Response(
+            self.serializer_class([DataObject(m) for m in mons],
+                                  many=True).data)
+
+
+if False:
+    class CliViewSet(RemoteViewSet):
```
alfredodeza: this base view set class is also undefined

src/pybind/mgr/rest/app/views/v2.py
```diff
@@ -0,0 +1,770 @@
...
+                raise ParseError("'command' field is required")
+            else:
+                if not (isinstance(command, basestring) or isinstance(command, list)):
+                    raise ParseError("'command' must be a string or list")
+
+            # Parse string commands to list
+            if isinstance(command, basestring):
+                command = shlex.split(command)
+
+            name = self.client.get_cluster(fsid)['name']
```
alfredodeza: and this `fsid` is also undefined here as well

src/mgr/ClusterState.h
```diff
@@ -0,0 +1,107 @@
...
+    monc->with_monmap(cb);
+  }
+
+  template<typename Callback, typename...Args>
+  void with_osdmap(Callback&& cb, Args&&...args)
+  {
+    assert(objecter != nullptr);
+    objecter->with_osdmap(cb);
+  }
+};
```
athanatos: I think what you meant was

```
template<typename... Args>
void with_osdmap(Args &&... args) {
  assert(objecter != nullptr);
  objecter->with_osdmap(std::forward<Args>(args)...);
}
```

src/mgr/ClusterState.cc
```diff
@@ -0,0 +1,337 @@
...
+  pending_inc.version = pg_map.version + 1; // to make apply_incremental happy
+
+  _update_creating_pgs(osd_map, &pending_inc);
+  _register_new_pgs(osd_map, &pending_inc);
+
+  pg_map.apply_incremental(g_ceph_context, pending_inc);
+
+  // TODO: Reinstate check_down_pgs logic?
+}
+
```
athanatos: This makes me extremely nervous.  I really don't want to duplicate this logic, even for a first cut.  I'd rather punt on pg stats altogether for this version and come back to it later.

src/mgr/ClusterState.h
tchaikov: @jcsp ping. the `args` are dropped on the floor.

src/mgr/ClusterState.cc
tchaikov: @athanatos ah, got what you mean!

jcsp: I've pushed an update that eliminates all the copy-paste.  ClusterState still doesn't do all the bits that PGMonitor does, but there isn't the confusing partial-copy now.

src/common/CommandTable.h
jcsp: I've pushed an update that refactors CommandTable

src/upstart/ceph-mgr.conf
```diff
@@ -0,0 +1,26 @@
+description "Ceph MDS"
```
jcsp: Done

src/pybind/mgr/rest/app/views/v2.py
jcsp: I've added a commit to strip out the `if False` blocks and/or replace them with django-rest-framework version checks as appropriate

PR #10330, opened 2016-07-18
rgw: multiple fixes for Swift's object expiration

rzarzynski:


rzarzynski:
Removing the DNM flag after manual testing and successful Tempest run.

PR #10332, opened 2016-07-18
librbd: optimize away unnecessary object map updates 

dillaman:


dillaman:
@neurodrone Hopefully this should resolve the mkfs discard performance issue. It needs to run through the rbd teuthology suite to ensure a regression wasn't introduced.

neurodrone:
That's neat! Thanks for this fix!

dillaman:
@neurodrone Thanks for the traces -- made it really easy to reproduce.

markhpc:
woooot, good job guys!

trociny:
@dillaman 'Fixes: http://tracker.ceph.com/issues/16707' to "rbd-replay: decode and replay discard IO operations" commit log?

src/rbd_replay/rbd-replay-prep.cc
```diff
@@ -427,6 +427,23 @@ class Processor {
...
+      uint64_t length = fields.uint64("len");
+      uint64_t completion = fields.uint64("completion");
+      imagectx_id_t imagectx = fields.uint64("imagectx");
+      require_image(ts, thread, imagectx, name, snap_name, readonly, ios);
+      action_id_t ionum = next_id();
+      IO::ptr io(new DiscardIO(ionum, ts, threadID, m_recent_completions,
+                                imagectx, offset, length));
+      thread->issued_io(io, &m_latest_ios);
+      ios->push_back(io);
+      m_pending_ios[completion] = io;
```
trociny: @dillaman This looks like wrong for not AIO case?

dillaman:
@trociny good catch -- update pushed

trociny:
@dillaman Do I understand the fix correctly, that the source of unnecessary object map updates was in send_pre() where we filtered out only updates that did not change the state, while now we also filter out updates changing state form OBJECT_NONEXISTENT to OBJECT_PENDING and from !OBJECT_PENDING to OBJECT_NONEXISTENT?

Or was send_post() a source of unnecessary updates too?

dillaman:
@trociny There wasn't an issue with the send_post state -- I just consolidated the logic to the same helper method.

The issue was send_pre would only perform an exact check of current state against desired state. When deleting, the desired state is first the _PENDING state.  Therefore, we can avoid updating the object map if our desired state is _PENDING and the current state is _NONEXISTENT.

trociny:
lgtm

PR #10334, opened 2016-07-18
Persist the missing set explicitely

athanatos:
We're going to need this for erasure coding overwrites because the on disk version of an object on a particular shard may not match the most recent log entry if the shard didn't participate.


athanatos:
Tested in samuelj-2016-07-14_14:34:22-rados-wip-ec-pg-log-distro-basic-smithi, failure appear to be unrelated (mostly async messenger failures, see summary.yaml for most of them for the known_bug).

src/osd/osd_types.h
```diff
@@ -2744,80 +2744,360 @@ inline ostream& operator<<(ostream& out, const pg_log_t& log)
...
+  eversion_t have_old(const hobject_t& oid) const override {
+    map<hobject_t, item, hobject_t::ComparatorWithDefault>::const_iterator m =
+      missing.find(oid);
+    if (m == missing.end())
+      return eversion_t();
+    const item &item(m->second);
+    return item.have;
+  }
+
+  void swap(pg_missing_set& o) {
```
jdurgin: I'm confused why this method exists. It looks like the only callers are proc_master_log() and proc_replica_log(), whose callers don't appear to use the modified missing set they passed in. Am I missing something about the code flow?

src/osd/osd_types.h
```diff
@@ -2744,80 +2744,360 @@ inline ostream& operator<<(ostream& out, const pg_log_t& log)
...
+
+template <bool Track>
+class ChangeTracker {
+public:
+  void changed(const hobject_t &obj) {}
+  template <typename F>
+  void get_changed(F &&f) const {}
+  void flush() {}
+  void clean() {
+    return true;
```
jdurgin: bool ret val

src/osd/osd_types.h
athanatos: They are using it to place the incoming missing set into the peer_missing map.  *this is being modified.  Although, it should really be called claim(), and it should be disabled for the variant with the tracker.

src/osd/osd_types.h
athanatos: I...am confused as to how this built.

src/osd/osd_types.h
jdurgin: ok, so it could have just been an assignment operator. and yeah, the tracking didn't make sense to me either

src/osd/osd_types.h
athanatos: Ah, yeah, nothing actually uses clean() for the Track=false instance.  Will fix.

src/osd/PGLog.cc
```diff
@@ -693,3 +685,97 @@ void PGLog::_write_log_and_missing(
...
+      assert(!log_keys_debug->count(i->first));
+      log_keys_debug->insert(i->first);
+    }
+  }
+
+  if (clear_divergent_priors) {
+    //dout(10) << "write_log_and_missing: writing divergent_priors" << dendl;
+    to_remove.insert("divergent_priors");
+  }
+  missing.get_changed(
```
jdurgin: are you thinking there will be more uses for get_changed() later? this implementation of tracking duplicates the objects that are still in the missing set, which could be read from the missing set directly.

athanatos: It should only call the passed function on objects which are different since the last time flush() was called.  It let's us efficiently remove keys for items that are no longer missing or add/replace keys for items that have been added/changed.

src/osd/PGLog.h
```diff
@@ -493,11 +487,11 @@ struct PGLog : DoutPrefixProvider {
   void undirty() {
     dirty_to = eversion_t();
     dirty_from = eversion_t::max();
-    dirty_divergent_priors = false;
     touched_log = true;
     trimmed.clear();
     writeout_from = eversion_t::max();
     check();
+    missing.flush();
```
athanatos: We call flush here, works similarly to the log dirtying code (writeout_from).

src/osd/osd_types.h
athanatos: Well, it invalidates the passed object, so probably not an assignment operator.

jdurgin:
looks good overall, if the missing set is going to be small the duplication is minor

athanatos:
Any delta we build up must (or it's a bug) be submitted to the objectstore before we release the pg lock (as with the rest of the pg metadata).  This whole thing is pretty much to make it so that I don't have to constantly worry about code that messes with the missing set failing to update the persisted on-disk state in some subtle and hard to reproduce way :P.

PR #10335, opened 2016-07-18
rgw: openssl 3

mdw-at-linuxbox:
[ this is an updated copy of PR #8339 ]
Augment configuration so it's easy to support http & https in one civetweb frontend.

Get rid of an incorrect warning when using openssl.

Document the thing so people can use it.

And load shared libraries by soname.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -313,7 +313,7 @@ if(USE_NSS)
 else(USE_NSS)
   #openssl
   find_package(OpenSSL REQUIRED)
-  set(HAVE_OPENSSL ON)
+  set(USE_OPENSSL ON)
   set(SSL_LIBRARIES ${OPENSSL_LIBRARIES})
   message(STATUS "SSL with OpenSSL selected (Libs: ${SSL_LIBRARIES})")
 endif(USE_NSS)
@@ -326,6 +326,47 @@ endif(WITH_XIO)
 
 #option for RGW
 option(WITH_RADOSGW "Rados Gateway is enabled" ON)
+
+if (${WITH_RADOSGW})
+  if (NOT DEFINED OPENSSL_FOUND)
+    message(STATUS "Looking for openssl anyways, because radosgw selected")
+    find_package(OpenSSL)
+  endif()
+  if (OPENSSL_FOUND)
+    execute_process(
+      COMMAND
+	"sh" "-c"
+	"objdump -p ${OPENSSL_SSL_LIBRARY} | sed -n 's/^  SONAME  *//p'"
+      OUTPUT_VARIABLE LIBSSL_SONAME
+      ERROR_VARIABLE OBJDUMP_ERRORS
+      RESULT_VARIABLE OBJDUMP_RESULTS
+      OUTPUT_STRIP_TRAILING_WHITESPACE)
+    if (OBJDUMP_RESULTS)
+      message(FATAL_ERROR "can't run objdump: ${OBJDUMP_RESULTS}")
+    endif()
+    if (NOT OBJDUMP_ERRORS STREQUAL "")
+      message(WARNING "message from objdump: ${OBJDUMP_ERRORS}")
+    endif()
+    execute_process(
+      COMMAND
+	"sh" "-c"
+	"objdump -p ${OPENSSL_CRYPTO_LIBRARY} | sed -n 's/^  SONAME  *//p'"
+      OUTPUT_VARIABLE LIBCRYPTO_SONAME
+      ERROR_VARIABLE OBJDUMP_ERRORS
+      RESULT_VARIABLE OBJDUMP_RESULTS
+      OUTPUT_STRIP_TRAILING_WHITESPACE)
+    if (OBJDUMP_RESULTS)
+      message(FATAL_ERROR "can't run objdump: ${OBJDUMP_RESULTS}")
+    endif()
+    if (NOT OBJDUMP_ERRORS STREQUAL "")
+      message(WARNING "message from objdump: ${OBJDUMP_ERRORS}")
+    endif()
+    message(STATUS "ssl soname: ${LIBSSL_SONAME}")
+    message(STATUS "crypto soname: ${LIBCRYPTO_SONAME}")
+  else()
+    message(WARNING "ssl not found: rgw civetweb may fail to dlopen libssl libcrypto")
+  endif()
+endif (${WITH_RADOSGW})
 
 #option for CephFS
 option(WITH_CEPHFS "CephFS is enabled" ON)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -834,15 +834,23 @@ if(${WITH_RADOSGW})
   add_library(civetweb_common_objs OBJECT ${civetweb_common_files})
   target_include_directories(civetweb_common_objs PUBLIC
 	"${CMAKE_SOURCE_DIR}/src/civetweb/include")
-  if(HAVE_SSL)
+  if(USE_OPENSSL)
     set_property(TARGET civetweb_common_objs
       APPEND PROPERTY COMPILE_DEFINITIONS NO_SSL_DL=1)
     target_include_directories(civetweb_common_objs PUBLIC
       "${SSL_INCLUDE_DIR}")
-  endif(HAVE_SSL)
+  endif(USE_OPENSSL)
   set_property(TARGET civetweb_common_objs
     APPEND PROPERTY COMPILE_DEFINITIONS USE_IPV6=1)
+  if (LIBSSL_SONAME)
+    set_property(TARGET civetweb_common_objs
+      APPEND PROPERTY COMPILE_DEFINITIONS SSL_LIB="${LIBSSL_SONAME}")
+    set_property(TARGET civetweb_common_objs
+      APPEND PROPERTY COMPILE_DEFINITIONS CRYPTO_LIB="${LIBCRYPTO_SONAME}")
+  endif()
+
   add_subdirectory(rgw)
+
 endif(${WITH_RADOSGW})
 
 install(FILES
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -840,6 +840,8 @@ if(${WITH_RADOSGW})
     target_include_directories(civetweb_common_objs PUBLIC
       "${SSL_INCLUDE_DIR}")
   endif(HAVE_SSL)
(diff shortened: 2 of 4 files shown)


djzort:
please pull this asap! :)

mdw-at-linuxbox:
somebody rearranged *all* the cmake stuff.  I've force pushed a fix for that.

alimaredia:
@mdw-at-linuxbox the CMake aspect of this looks fine to me. @tchaikov anything look out of place to you?

src/rgw/rgw_main.cc
```diff
@@ -403,10 +403,7 @@ int main(int argc, const char **argv)
 
       fe = new RGWFCGXFrontend(fcgi_pe, config);
     } else if (framework == "civetweb" || framework == "mongoose") {
-      int port;
-      config->get_val("port", 80, &port);
-
-      RGWProcessEnv env = { store, &rest, olog, port };
+      RGWProcessEnv env = { store, &rest, olog, 0 };
```
yehudasa: @mdw-at-linuxbox you can't remove the port here without addressing its use at RGWMongoose::init_env()

mdw-at-linuxbox: Ok, that's related to : http://tracker.ceph.com/issues/16548 .  I sent mail 20160628 about that where I went through the code that does this.  The basic problem is that the current code path that uses that assumes that it can guess the server port based solely on whether the request came in with ssl or not.  That means not more than 1 ssl and 1 non-ssl port.  Do we really want to enforce that?

yehudasa: @mdw-at-linuxbox we need to have some solution to that, a 0 port number is not a proper one.

src/CMakeLists.txt
```diff
@@ -818,13 +818,21 @@ if(${WITH_RADOSGW})
   add_library(civetweb_common_objs OBJECT ${civetweb_common_files})
   target_include_directories(civetweb_common_objs PUBLIC
 	"${CMAKE_SOURCE_DIR}/src/civetweb/include")
-  if(HAVE_SSL)
+  if(USE_OPENSSL)
```
tchaikov: where is `USE_OPENSSL` initialized?

src/CMakeLists.txt
```diff
@@ -818,13 +818,21 @@ if(${WITH_RADOSGW})
...
     set_property(TARGET civetweb_common_objs
       APPEND PROPERTY COMPILE_DEFINITIONS NO_SSL_DL=1)
     target_include_directories(civetweb_common_objs PUBLIC
       "${SSL_INCLUDE_DIR}")
-  endif(HAVE_SSL)
+  endif(USE_OPENSSL)
+  if (DEFINED LIBSSL_SONAME)
```
tchaikov: could just put `if(LIBSSL_SONAME)`.

src/CMakeLists.txt
```diff
@@ -818,13 +818,21 @@ if(${WITH_RADOSGW})
...
+    set_property(TARGET civetweb_common_objs
+      APPEND PROPERTY COMPILE_DEFINITIONS SSL_LIB="${LIBSSL_SONAME}")
+    set_property(TARGET civetweb_common_objs
+      APPEND PROPERTY COMPILE_DEFINITIONS CRYPTO_LIB="${LIBCRYPTO_SONAME}")
+  endif()
+
   add_subdirectory(rgw)
+
```
tchaikov: could remove this empty line.

smithfarm:
@mdw-at-linuxbox Does this PR fix http://tracker.ceph.com/issues/13718 ? If so, could you add

```
Fixes: http://tracker.ceph.com/issues/13718
```

immediately above the `Signed-off-by:` line of the main/most-relevant commit?

mdw-at-linuxbox:
Yes, this pull request would address #13718 - excepting that the problem Yehuda raised above makes this more complicated and probably a separate patch.

The top level CMakeLists.txt sets "HAVE_OPENSSL".  It should instead set USE_OPENSSL.

The "extra" blank line is because I had stuck a large glop of SSL obscurity in just before, and I didn't think the "add_subdirectory(rgw)" was obviously intentional without the extra blank lines.

Sure "if(LIBSSL_SONAME)" would work too - I take it we prefer this construction?

tchaikov:
> I take it we prefer this construction?

yeah. i think it's more concise.

@mdw-at-linuxbox ping

src/rgw/rgw_main.cc
theanalyst: We could either 
1. parse port=".." the way civetweb does, ie. allow ip:port combo, r, s params that civetweb allows etc.

2.  alternatively, we can set override the env.port after initializing port from civetweb and getting the value of ports civetweb parses using the `mg_get_ports` api ?; ie something like this in rgw_civetweb_frontend:

```c++
(assume ports = new int [num ports] etc)
size_t len = mg_get_ports(ctx, num_ports, ports, ssl_ports)
if (len){
    env.port = ports[0] ;}
``` 
etc 
?

theanalyst:
this should probably address http://tracker.ceph.com/issues/16581 as well

mdw-at-linuxbox:
I've updated the branch wip-rgw-openssl-3.  I've removed "Replace + with , in civetweb port string.".  Whatever we do for that should be a separate pull request.  I've revised the doc accordingly.  I also spelled USE_OPENSSL right and removed DEFINED.  Finally I rebased on the latest master.  Passed my testing.  Hopefully gitbuilder will still like it.

src/rgw/rgw_main.cc
```diff
@@ -403,10 +403,26 @@ int main(int argc, const char **argv)
...
+	return -1;
+      }
+      if (*portend && (*portend != 's' || portend[1])) {
+	derr << "ERROR: trailing stuff on port " << port << dendl;
+	return -1;
+      }
+      if (portno <= 0 || portno > 65535) {
+	derr << "ERROR: port out of range " << port << dendl;
+	return -1;
+      }
```
theanalyst: what happens if we pass an ip:port combo which currently works, albeit setting SERVER_PORT to 0,  here don't we end up not starting rgw at all ?

CMakeLists.txt
```diff
@@ -327,6 +327,47 @@ endif(WITH_XIO)
 #option for RGW
 option(WITH_RADOSGW "Rados Gateway is enabled" ON)
 
+if (${WITH_RADOSGW})
+  if (NOT DEFINED OPENSSL_FOUND)
+    message(STATUS "Looking for openssl anyays, because radosgw selected")
```
theanalyst: nit: typo, anyway

src/rgw/rgw_main.cc
mdw-at-linuxbox: Well, actually, civetweb doesn't start up if you give it a bad port.  It fails in a much more confusing way later on.

However, you are right, it's possible to pass an ip:port in.  Or [ipv6]:port.  And append one of [sr,].  And it's possible to pass in non-zero bits >65535 in portno and have them ignored.  The ssl redir feature isn't useful to us currently, since we don't yet support listening for both ssl|non-ssl.  And we'll never pass in a , because the config file parser eats those elsewhere.  I don't mind breaking rgw for people who set high order bits in portno.

But being able to specify a specific ipaddr is useful, sigh.

mdw-at-linuxbox:
I've force-pushed a new copy of this branch.  Rebased on latest master, there are now 4 commits:
a43ebf3 set -DUSE_IPV6 (NEW)
fe08886 document ssl
4470877 load by soname (FIX TYPO)
a43ebf3 parse and accept portno (UPDATED).

src/rgw/rgw_main.cc
```diff
@@ -187,6 +187,30 @@ static void reloader_handler(int signum)
   sighup_handler(signum);
 }
 
+static int parse_civetweb_portno(const string &port) {
+  char *portend;
+  const char *portstart = port.c_str();
+  const char *cp = strrchr(portstart, ':');
```
yehudasa: @mdw-at-linuxbox could just use port.find_last_of()

src/rgw/rgw_main.cc
```diff
@@ -187,6 +187,30 @@ static void reloader_handler(int signum)
...
+  if (cp)
```
yehudasa: @mdw-at-linuxbox in general we prefer explicit scoping

src/rgw/rgw_main.cc
```diff
@@ -187,6 +187,30 @@ static void reloader_handler(int signum)
...
+    ++cp;
+  else
+    cp = portstart;
+  int portno = strtol(cp, &portend, 10);
```
yehudasa: @mdw-at-linuxbox use strict_strtol

yehudasa: oh, nevermind, I see you need &portend

src/rgw/rgw_main.cc
```diff
@@ -187,6 +187,30 @@ static void reloader_handler(int signum)
...
+  if (cp == portend) {
+    derr << "ERROR: cannot parse port " << port << dendl;
+    return -1;
```
yehudasa: return -EINVAL

yehudasa: also other occurances in this function

src/rgw/rgw_main.cc
```diff
@@ -402,10 +426,15 @@ int main(int argc, const char **argv)
 
       fe = new RGWFCGXFrontend(fcgi_pe, config);
     } else if (framework == "civetweb" || framework == "mongoose") {
-      int port;
-      config->get_val("port", 80, &port);
+      string port;
+      int portno;
+      config->get_val("port", "80", &port);
+      portno = parse_civetweb_portno(port);
+      if (portno == -1) {
```
yehudasa: if (portno < 0)

mdw-at-linuxbox:
port=0s can't work - you already pointed out why.  port=0 would result in a random port assignment.  How is that useful?

yehudasa:
@mdw-at-linuxbox see my comments on the last commit, other than that I think it's ok

djzort:
this seems to have gotten stuck?

mdw-at-linuxbox:
I've created a new PR that has just the soname fix and the documentation fix.  That is now PR # 11571.

liewegas:
need rebase?

mdw-at-linuxbox:
This PR is superseded by https://github.com/ceph/ceph/pull/11571 and https://github.com/ceph/ceph/pull/11776 so I'm closing this one.

PR #10338, opened 2016-07-19
mds/mdsmonitor: kill mds dump

renhwztetecs:
"mds dump" and "fs dump" are repeated, and "mds dump" display incomplete.
see http://tracker.ceph.com/issues/16730

Fixes: 16730

### DIFF
--- a/src/mon/MDSMonitor.cc
+++ b/src/mon/MDSMonitor.cc
@@ -930,53 +930,6 @@ bool MDSMonitor::preprocess_command(MonOpRequestRef op)
       ds << fsmap;
     }
     r = 0;
-  } else if (prefix == "mds dump") {
-    int64_t epocharg;
-    epoch_t epoch;
-
-    FSMap *p = &fsmap;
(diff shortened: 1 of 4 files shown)


xiaoxichen:
"fs dump " is added recently for multi-fs case, legacy client only support "mds dump". Killing the "mds dump" will definitely break hammer + infernalis client compatibility.

gregsfortytwo:
@xiaoxichen, actually, they should keep working since clients get the list of available commands from the monitors.

That said, the witnessed behavior is deliberate and we want to keep both commands, at least for now, as I elaborated in the ticket. Thanks! :)

renhwztetecs:
oh I see，thanks all
by the way,
In theory, multi-fs is completely isolated multiple single-fs，why the multi-fs not stable? 
what we need to do to improve it?

gregsfortytwo:
It's just very fresh code. Also, the security model is almost nonexistent and the UI has a lot of rough edges (eg, the way clients specify which fs to mount by ID instead of name).

PR #10339, opened 2016-07-19
selinux: allow read /proc/<pid>/cmdline

tchaikov:
we read /proc/<pid>/cmdline to figure out who is terminating us.

Fixes: http://tracker.ceph.com/issues/16675

### DIFF
--- a/selinux/ceph.te
+++ b/selinux/ceph.te
@@ -75,6 +75,8 @@ corecmd_exec_shell(ceph_t)
 
 dev_read_urand(ceph_t)
 
+domain_read_all_domains_state(ceph_t)
+
 fs_getattr_all_fs(ceph_t)
 
 auth_use_nsswitch(ceph_t)


b-ranto:
lgtm. I'm just wondering whether it wouldn't be safer to just read where the /proc/<pid>/exe points to instead (i.e. readlink on /proc/<pid>/exe)?

tchaikov:
@b-ranto i am post a PR #10345, which needs to read `/proc/<pid>/cmdline`. if we want the cmdline instead of `argv[0]`, maybe we do need to do so?

tchaikov:
i will hold the PR until someone nah or yah #10345 or timeout =).

b-ranto:
Hmm, now that I think about it -- those bits won't work if the daemons are killed by root? The ceph user can't read `/proc/<pid>/` of a process owned by root.

tchaikov:
in that case, we will fail when trying to open `/proc/<pid>/cmdline`, and "<unknown>" will be printed instead in the log.

b-ranto:
ah, ok, makes sense.

ktdreyer:
@tchaikov when you say "we" read cmdline, does that "we" mean "teuthology's code" or "ceph's code" ?

ktdreyer:
Whoops, I read https://github.com/ceph/ceph/pull/8964 and it's clearer now that it is Ceph itself doing this. Thanks :)

PR #10341, opened 2016-07-19
rbd-mirror: reduce memory footprint during journal replay

dillaman:


src/test/librbd/fsx.cc
```diff
@@ -323,7 +324,7 @@ int register_journal(rados_ioctx_t ioctx, const char *image_name) {
                 return r;
         }
 
-        journal::Journaler journaler(io_ctx, image_id, JOURNAL_CLIENT_ID, 0);
+        journal::Journaler journaler(io_ctx, image_id, JOURNAL_CLIENT_ID, {});
```
rjfd: The default `commit_interval` is 5, but the test was using value 0. If the default value does not matter for the test please ignore this comment!
The same for the remaining similar changes.

dillaman: I ratcheted it down in the cases where the test case truly depended on looking at the commit position.  But in general, this parameter only has any effect after a crash.

src/journal/JournalPlayer.cc
```diff
@@ -492,12 +493,26 @@ void JournalPlayer::prune_tag(uint64_t tag_tid) {
...
+                     << dendl;
+    for (auto &player_pair : m_object_players) {
+      ObjectPlayerPtr object_player(player_pair.second);
+      object_player->set_refetch_state(ObjectPlayer::REFETCH_STATE_IMMEDIATE);
+    }
+  }
+
+  // trim empty player to prefetch the next available object
+  for (auto &player_pair : m_object_players) {
+    ObjectPlayerPtr object_player(player_pair.second);
```
trociny: @dillaman What is the purpose of this line?

dillaman: Apparently nothing -- will clean up

trociny:
lgtm

PR #10345, opened 2016-07-19
global/signal_handler: dump cmdline instead of arg[0]

tchaikov:
sometimes the daemon gets killed by a bash script. and `$0` is always `bash`, which is just less useful.

### DIFF
--- a/src/global/signal_handler.cc
+++ b/src/global/signal_handler.cc
@@ -152,7 +152,7 @@ void install_standard_sighandlers(void)
 #include "common/Thread.h"
 #include <errno.h>
 
-void get_name_by_pid(pid_t pid, char *task_name)
+string get_name_by_pid(pid_t pid)
 {
   char proc_pid_path[PATH_MAX] = {0};
   snprintf(proc_pid_path, PATH_MAX, "/proc/%d/cmdline", pid);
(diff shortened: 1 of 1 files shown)


src/global/signal_handler.cc
```diff
@@ -163,18 +163,21 @@ void get_name_by_pid(pid_t pid, char *task_name)
...
+  char buf[PATH_MAX] = {0};
+  int ret = read(fd, buf, sizeof(buf));
   if (ret < 0) {
     ret = -errno;
-    derr << "Fail to read '" << proc_pid_path 
-         << "' error = " << cpp_strerror(ret) 
+    derr << "Fail to read '" << proc_pid_path
+         << "' error = " << cpp_strerror(ret)
          << dendl;
+    return "<unknown>";
```
xiexingguo: I think you may need an extra
  close(fd);
before return, otherwise there is a leak.

xiexingguo:
Looks good except for the fd leak

src/global/signal_handler.cc
tchaikov: yeah, thanks for pointing this out!

tchaikov:
changelog

- fix the fd leak when `read()` fails
- add a comment, as command line length limit is much larger than max `PATH_MAX`.

yuriw:
http://pulpito.ceph.com/yuriw-2016-07-21_15:02:04-rados-wip-yuri-testing2_2016_7_21-distro-basic-smithi/
http://pulpito.ceph.com/yuriw-2016-07-25_09:25:30-rados-wip-yuri-testing2_2016_7_21-distro-basic-smithi/

PR #10346, opened 2016-07-19
test: ceph-disk.sh do not kill all daemons

tchaikov:
this fixes mysterious "racing" issues when running ceph-helper tests in parallel.


badone:
lgtm

PR #10349, opened 2016-07-19
os/bluestore: add a boundary check of cache read

xiexingguo:
By trapping the "==" case and break, we can finish
the cache read just in time. As a result, we can avoid
two extra operations(adjusting "offset" and "length") and
the final loop check.

Since this is one of key methods of cache, it deserves this
fix.

### DIFF
--- a/src/os/bluestore/BlueStore.cc
+++ b/src/os/bluestore/BlueStore.cc
@@ -1008,6 +1008,8 @@ void BlueStore::BufferSpace::read(
       } else {
 	res[offset].append(b->data);
 	res_intervals.insert(offset, b->length);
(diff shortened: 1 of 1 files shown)


xiexingguo:
2016-07-20 17:58:29.861655 7f3524c71680  1 bluestore(store_test_temp_dir) fsck finish with 0 errors
[       OK ] ObjectStore/StoreTest.TooManyBlobsTest/2 (22225 ms)
[----------] 59 tests from ObjectStore/StoreTest (2145269 ms total)

[==========] 59 tests from 1 test case ran. (2145269 ms total)
[  PASSED  ] 59 tests.

PR #10350, opened 2016-07-19
rgw: fix autotools building for the ASIO frontend.

rzarzynski:
CC: @cbodley, @mattbenjamin.

### DIFF
--- a/src/rgw/Makefile.am
+++ b/src/rgw/Makefile.am
@@ -97,7 +97,7 @@ librgw_la_SOURCES += rgw/rgw_ldap.cc
 endif
 
 librgw_la_CXXFLAGS = -Woverloaded-virtual -fPIC -I$(srcdir)/xxHash \
-	${AM_CXXFLAGS}
+	-I$(srcdir)/Beast/include ${AM_CXXFLAGS}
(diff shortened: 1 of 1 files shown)


src/rgw/Makefile.am
```diff
@@ -160,7 +160,7 @@ radosgw_SOURCES = \
 	rgw/rgw_main.cc
 
 radosgw_CFLAGS = -I$(srcdir)/civetweb/include -fPIC -I$(srcdir)/xxHash
-radosgw_CXXFLAGS = ${RGW_CXXFLAGS} ${AM_CXXFLAGS}
+radosgw_CXXFLAGS = ${RGW_CXXFLAGS} ${AM_CXXFLAGS} -I$(srcdir)/Beast/include
```
cbodley: i think we want this in `librgw_la_CXXFLAGS`. the Beast types could be useful outside of the asio frontend

rzarzynski:
@cbodley: I added the path also to `librgw_la_CXXFLAGS`. However, `radosgw_CXXFLAGS` should preserved as is. Otherwise compilation fails.

cbodley:
thanks!

PR #10352, opened 2016-07-19
Cmake work to get inital FreeBSD stuff

wjwithagen:


CMakeLists.txt
```diff
@@ -154,15 +183,24 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ${ENABLE_SHARED})
...
+  set(HAVE_UDEV ${UDEV_FOUND})
+
+  find_package(aio REQUIRED)
+  set(HAVE_LIBAIO ${AIO_FOUND})
+  message(STATUS "${AIO_LIBS}")
+
+  find_package(blkid REQUIRED)
+  set(HAVE_BLKID ${BLKID_FOUND})
+else()
+  set(HAVE_UDEV OFF)
```
tchaikov: no need to `set(HAVE_UDEV OFF)` explicitly. an empty variable is evaluated to false.

CMakeLists.txt
```diff
@@ -19,6 +19,24 @@ endif (POLICY CMP0023)
 
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake/modules/")
 
+# find out which platform we are building on
+if(CMAKE_SYSTEM_NAME MATCHES "Linux")
+  message(STATUS "Building for Linux.")
+  set(LINUX ON)
+  set(UNIX ON)
```
tchaikov: @alimaredia do we use `UNIX` anywhere? can we just remove this?

CMakeLists.txt
wjwithagen: I know, but I'd rather be explicit about it, and match the counterpart a few lines up.

CMakeLists.txt
```diff
@@ -19,6 +19,24 @@ endif (POLICY CMP0023)
...
+  FIND_PACKAGE(Threads)
+elseif(CMAKE_SYSTEM_NAME MATCHES "FreeBSD")
+  message(STATUS "Building for FreeBSD.")
+  set(FREEBSD ON)
+  set(UNIX ON)
+  FIND_PACKAGE(Threads)
+elseif(CMAKE_SYSTEM_NAME MATCHES "Darwin")
+  message(STATUS "Building for Darwin.")
+  set(OperatingSystem "Mac OS X")
```
tchaikov: no need to set this, cmake have `APPLE`. see https://cmake.org/cmake/help/v3.0/variable/APPLE.html

CMakeLists.txt
```diff
@@ -19,6 +19,24 @@ endif (POLICY CMP0023)
 
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake/modules/")
 
+# find out which platform we are building on
+if(CMAKE_SYSTEM_NAME MATCHES "Linux")
+  message(STATUS "Building for Linux.")
```
tchaikov: could you remove the message? it does not help imho.

CMakeLists.txt
```diff
@@ -103,7 +127,12 @@ CHECK_INCLUDE_FILES("sys/types.h" HAVE_SYS_TYPES_H)
 CHECK_INCLUDE_FILES("sys/vfs.h" HAVE_SYS_VFS_H)
 CHECK_INCLUDE_FILES("sys/prctl.h" HAVE_SYS_PRCTL_H)
 CHECK_INCLUDE_FILES("execinfo.h" HAVE_EXECINFO_H)
-CHECK_INCLUDE_FILES("sched.h" HAVE_SCHED)
+if(LINUX)
+  CHECK_INCLUDE_FILES("sched.h" HAVE_SCHED)
+endif(LINUX)
+if(FREEBSD)
+  CHECK_INCLUDE_FILES("sched.h" HAVE_SCHED_FREEBSD)
```
tchaikov: is this used anywhere?

CMakeLists.txt
tchaikov: please remove this line, see https://cmake.org/cmake/help/v3.0/variable/UNIX.html

CMakeLists.txt
```diff
@@ -19,6 +19,24 @@ endif (POLICY CMP0023)
...
+# find out which platform we are building on
+if(CMAKE_SYSTEM_NAME MATCHES "Linux")
+  message(STATUS "Building for Linux.")
+  set(LINUX ON)
+  set(UNIX ON)
+  FIND_PACKAGE(Threads)
+elseif(CMAKE_SYSTEM_NAME MATCHES "FreeBSD")
+  message(STATUS "Building for FreeBSD.")
+  set(FREEBSD ON)
+  set(UNIX ON)
```
tchaikov: and remove this line.

CMakeLists.txt
wjwithagen: @tchaikov 
"There can only be one?" (free after the Highlander movie.) :)

CMakeLists.txt
wjwithagen: @tchaikov 
Not yet, but sched.h is detected on FreeBSD, but is not compatible.
And that needs work. So either it is just excluded, or we already make a note, and work on the scheduler when the more pressing things are fixed.

CMakeLists.txt
wjwithagen: @tchaikov 
These are all things that I bluntly copied from what was there.
More or less my moto while porting: Leave it alone if you don't know why it is there.
I'll fix both

CMakeLists.txt
tchaikov: okay, can we remove this code block and leave a note here?

CMakeLists.txt
tchaikov: oh, i meant we can remove all the "Building for *" messages.

CMakeLists.txt
tchaikov: > More or less my moto while porting: Leave it alone if you don't know why it is there.

i agree with you. and also did i checked/grep'ed the code couple times. no obvious evidence that it's used. so let's just remove it. thanks.

CMakeLists.txt
wjwithagen: On 20-7-2016 14:25, Kefu Chai wrote:
> In CMakeLists.txt
> <https://github.com/ceph/ceph/pull/10352#discussion_r71514349>:
> 
>> @@ -103,7 +127,12 @@ CHECK_INCLUDE_FILES("sys/types.h" HAVE_SYS_TYPES_H)
>>  CHECK_INCLUDE_FILES("sys/vfs.h" HAVE_SYS_VFS_H)
>>  CHECK_INCLUDE_FILES("sys/prctl.h" HAVE_SYS_PRCTL_H)
>>  CHECK_INCLUDE_FILES("execinfo.h" HAVE_EXECINFO_H)
>> -CHECK_INCLUDE_FILES("sched.h" HAVE_SCHED)
>> +if(LINUX)
>> +  CHECK_INCLUDE_FILES("sched.h" HAVE_SCHED)
>> +endif(LINUX)
>> +if(FREEBSD)
>> +  CHECK_INCLUDE_FILES("sched.h" HAVE_SCHED_FREEBSD)
> 
> okay, can we remove this code block and leave a note here?

It now is for most platforms nothing more that a note.
I'll take the full thing out, and make a note in my todo's that this
should be addressed at one time.

wjwithagen:
@tchaikov 
I think the errors are fallout of other issues in the tree.

wjwithagen:
@tchaikov 
Would it be possible to get this in before the weekend.
Then over the weekend I'll push another small set of changes so we can gradually move on.

tchaikov:
lgtm

PR #10355, opened 2016-07-19
rgw: back off bucket sync on failures, don't store marker

yehudasa:
Fixes: http://tracker.ceph.com/issues/16742

If we fail on any single entry in bucket, skip updating the marker tracker
so that next time we'll go over that entry, and back off. This will trigger
a report to the data sync error repo and eventually a retry on the failing
object.

### DIFF
--- a/src/rgw/rgw_data_sync.cc
+++ b/src/rgw/rgw_data_sync.cc
@@ -2068,6 +2068,7 @@ class RGWBucketSyncSingleEntryCR : public RGWCoroutine {
 
   RGWDataSyncDebugLogger logger;
 
+
 public:
   RGWBucketSyncSingleEntryCR(RGWDataSyncEnv *_sync_env,
                              RGWBucketInfo *_bucket_info,
@@ -2160,9 +2161,9 @@ public:
         yield call(sync_env->error_logger->log_error_cr(sync_env->conn->get_remote_id(), "data", error_ss.str(), retcode, "failed to sync object"));
       }
 done:
-      /* update marker */
-      set_status() << "calling marker_tracker->finish(" << entry_marker << ")";
       if (sync_status == 0) {
+        /* update marker */
+        set_status() << "calling marker_tracker->finish(" << entry_marker << ")";
         yield call(marker_tracker->finish(entry_marker));
         sync_status = retcode;
       }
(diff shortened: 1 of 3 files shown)


src/rgw/rgw_data_sync.cc
```diff
@@ -2286,33 +2288,41 @@ int RGWBucketShardFullSyncCR::operate()
           while (collect(&ret, lease_stack)) {
             if (ret < 0) {
               ldout(sync_env->cct, 0) << "ERROR: a sync operation returned error" << dendl;
+              sync_status = ret;
```
cbodley: `collect()` never returns true, so this block won't execute. removing the `while()` should work. same with `RGWBucketShardIncrementalSyncCR` below

src/rgw/rgw_data_sync.cc
```diff
@@ -2512,33 +2541,55 @@ int RGWBucketShardIncrementalSyncCR::operate()
...
+      while (again) {
+        again = collect(&ret, lease_stack);
+        if (ret < 0) {
+          ldout(sync_env->cct, 0) << "ERROR: a sync operation returned error" << dendl;
+          sync_status = ret;
+          /* we have reported this error */
+        }
+        /* not waiting for child here */
+      }
+    }
```
cbodley: looks correct with the new `collect()`. this logic is almost identical to `RGWCoroutine::drain_children()` though, maybe we could find a way to get the error codes out of that instead?

src/rgw/rgw_sync.h
```diff
@@ -363,13 +355,23 @@ class RGWSyncShardMarkerTrack {
...
-    last_stored_marker = high_marker;
-    return store_marker(high_marker, high_entry.pos, high_entry.timestamp);
+
+    auto last = i;
+    --i;
+    const T& high_marker = i->first;
+    marker_entry& high_entry = i->second;
+    RGWCoroutine *cr = store_marker(high_marker, high_entry.pos, high_entry.timestamp);
+    finish_markers.erase(finish_markers.begin(), last);
+    return cr;
```
cbodley: looks good. it does look like a fix for the meta sync issue in bz 1357641

src/rgw/rgw_data_sync.cc
yehudasa: @cbodley yes, this needs to be cleaned up. But we should probably do it outside the scope of this bug.

PR #10368, opened 2016-07-20
rgw: fix upgrade from old multisite to new multisite configuration

oritwas:
We need to store the updated current period after adding the old converted regions

Fixes: http://tracker.ceph.com/issues/16751

### DIFF
--- a/src/rgw/rgw_rados.cc
+++ b/src/rgw/rgw_rados.cc
@@ -3461,6 +3461,7 @@ int RGWRados::replace_region_with_zonegroup()
          iter ++) {
       RGWZoneParams zoneparams(iter->first, iter->first);
       zoneparams.set_id(iter->first);
+      zoneparams.realm_id = realm.get_id();
       ret = zoneparams.init(cct, this);
       if (ret < 0) {
         ldout(cct, 0) << "failed to init zoneparams  " << iter->first <<  ": " << cpp_strerror(-ret) << dendl;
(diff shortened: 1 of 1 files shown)


src/rgw/rgw_rados.cc
```diff
@@ -3488,16 +3489,37 @@ int RGWRados::replace_region_with_zonegroup()
...
+    }
+  }
 
+  for (auto const& iter : regions) {
+    RGWZoneGroup zonegroup(iter);
+    int ret = zonegroup.init(cct, this, true, true);
+    if (ret < 0) {
+      ldout(cct, 0) << "failed init zonegroup" << iter << ": ret "<< ret << " " << cpp_strerror(-ret) << dendl;
+      return ret;
+    }
```
cbodley: is it possible to avoid reloading the zonegroup here? it doesn't look like `delete_obj()` really depends on this when old_format=true

oritwas: I moved the cleanup to the end so user can keep the old config in case of a failed conversion but it requires to reload the region.
We are deleting the old regions, the flag old_format has to be set otherwise we will delete the new zonegroups

cbodley: i understand the logic, and everything looks correct to me. i was just suggesting an optimization by skipping the step where we reload the zonegroup from rados. but optimizing the upgrade is probably wasted effort. :)

if you're satisfied with testing, i'm happy to merge :+1:

oritwas: please do :)

PR #10369, opened 2016-07-20
msg/async/Event: ensure not refer to member variable which may destroyed

yuyuyu101:
If nonwait is false, another thread is waiting for complete. After calling f()
this thread will notify for finishing and at this moment another thread will
destroy this C_submit_event right now. So we may refer to nonwait when
C_submit_event is disappeared.

### DIFF
--- a/src/msg/async/Event.h
+++ b/src/msg/async/Event.h
@@ -183,8 +183,9 @@ class EventCenter {
       lock.lock();
       cond.notify_all();
       done = true;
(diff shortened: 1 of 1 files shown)


jdurgin:
lgtm, thanks!

jdurgin:
@yuriw let's get this in the next batch of testing since it should help reduce failures a bunch

yuriw:

PR #10372, opened 2016-07-20
rgw multisite: trim data logs as peer zones catch up

cbodley:


src/rgw/rgw_rest_log.cc
```diff
@@ -837,6 +839,89 @@ void RGWOp_DATALog_Delete::execute() {
...
+  }
+  void execute() override;
+  void send_response() override;
+  const string name() override { return "get_metadata_log_status"; }
+};
+
+void RGWOp_MDLog_Status::execute()
+{
+  // construct a temporary status manager to read the sync status
+  RGWMetaSyncStatusManager sync(store, store->get_async_rados());
```
yehudasa: @cbodley shouldn't we have an existing status manager through s->store?

src/rgw/rgw_data_sync.cc
```diff
@@ -1368,6 +1368,62 @@ class RGWDataSyncShardControlCR : public RGWBackoffControlCR {
...
+      yield {
+        rgw_http_param_pair pairs[] = {
+          { "type", "data" },
+          { "status", nullptr },
+          { "source-zone", zone.c_str() },
+          { nullptr, nullptr }
+        };
+        string p = "/admin/log/";
+
+        http_op = new RGWRESTReadResource(conn, p, pairs, NULL, http);
```
yehudasa: @cbodley isn't there a cr that we could call that will wrap that for us?

src/rgw/rgw_rados.cc
```diff
@@ -3771,6 +3804,13 @@ int RGWRados::init_complete()
       thread->start();
       data_sync_processor_threads[iter->first] = thread;
     }
+    sync_log_trimmer = new RGWSyncLogTrimThread(this);
+    ret = sync_log_trimmer->init();
+    if (ret < 0) {
+      ldout(cct, 0) << "ERROR: failed to initialize" << dendl;
```
yehudasa: maybe be more specific about what failed?

yehudasa:
@cbodley overall looks good. Meta should be pretty similar, will need to get bucket index trimming too.

cbodley:
updated to address your comments. the rest api now uses the existing SyncStatusManagers, which required these new commits:
* rgw: expose sync managers through RGWRados
* rgw: change read_sync_status interface
* rgw: use separate cr manager for read_sync_status

also changed so that a `rgw_sync_log_trim_interval` of 0 will disable the trim thread

cbodley:
@yehudasa this was part of the latest wip-cbodley-testing run, but there was a single [s3-tests failure](http://qa-proxy.ceph.com/teuthology/cbodley-2016-09-16_10:00:36-rgw-wip-cbodley-testing---basic-mira/419161/teuthology.log) that hit the assert() that detects deadlocks in `RGWCoroutinesManager::run()`.. at some point during the run, osd.2 started failing every op with -ECANCELED, leading to a cascade of coroutine errors. so I'm not sure whether the cr deadlock was due to this unique failure case, or caused by new coroutines added by this PR. I'll run this branch through more local testing to see if I can learn more.

PR #10373, opened 2016-07-21
Jewel mds snap failover

ukernel:


ukernel:
http://pulpito.ceph.com/zyan-2016-07-19_23:50:41-fs-jewel-mds-snap-failover-testing-basic-smithi

ukernel:
https://github.com/ceph/ceph-qa-suite/pull/1088

PR #10378, opened 2016-07-21
librbd: optionally unregister "laggy" journal clients

trociny:
Fixes: http://tracker.ceph.com/issues/14738

### DIFF
--- a/src/common/config_opts.h
+++ b/src/common/config_opts.h
@@ -1212,6 +1212,7 @@ OPTION(rbd_tracing, OPT_BOOL, false) // true if LTTng-UST tracepoints should be 
 OPTION(rbd_validate_pool, OPT_BOOL, true) // true if empty pools should be validated for RBD compatibility
 OPTION(rbd_validate_names, OPT_BOOL, true) // true if image specs should be validated
 OPTION(rbd_auto_exclusive_lock_until_manual_request, OPT_BOOL, true) // whether to automatically acquire/release exclusive lock until it is explicitly requested, i.e. before we know the user of librbd is properly using the lock API
+OPTION(rbd_mirroring_resync_after_disconnect, OPT_BOOL, false) // automatically start image resync after mirroring is disconnected due to being laggy
 
 /*
  * The following options change the behavior for librbd's image creation methods that
--- a/src/common/config_opts.h
+++ b/src/common/config_opts.h
@@ -1252,6 +1252,7 @@ OPTION(rbd_journal_object_flush_bytes, OPT_INT, 0) // maximum number of pending 
 OPTION(rbd_journal_object_flush_age, OPT_DOUBLE, 0) // maximum age (in seconds) for pending commits
 OPTION(rbd_journal_pool, OPT_STR, "") // pool for journal objects
 OPTION(rbd_journal_max_payload_bytes, OPT_U32, 16384) // maximum journal payload size before splitting
+OPTION(rbd_journal_max_concurrent_object_sets, OPT_INT, 0) // maximum number of object sets a journal client can be behind before it is automatically unregistered
 
 /**
  * RBD Mirror options
--- a/src/journal/JournalMetadata.cc
+++ b/src/journal/JournalMetadata.cc
@@ -749,6 +749,10 @@ void JournalMetadata::handle_refresh_complete(C_Refresh *refresh, int r) {
     Client client(m_client_id, bufferlist());
     RegisteredClients::iterator it = refresh->registered_clients.find(client);
     if (it != refresh->registered_clients.end()) {
+      if (it->state == cls::journal::CLIENT_STATE_DISCONNECTED) {
+	ldout(m_cct, 0) << "client flagged disconnected: " << m_client_id
+			<< dendl;
+      }
       m_minimum_set = MAX(m_minimum_set, refresh->minimum_set);
       m_active_set = MAX(m_active_set, refresh->active_set);
       m_registered_clients = refresh->registered_clients;
@@ -810,8 +814,10 @@ void JournalMetadata::handle_commit_position_task() {
   librados::ObjectWriteOperation op;
   client::client_commit(&op, m_client_id, m_commit_position);
 
-  C_NotifyUpdate *ctx = new C_NotifyUpdate(this, m_commit_position_ctx);
+  Context *ctx = new C_NotifyUpdate(this, m_commit_position_ctx);
   m_commit_position_ctx = NULL;
+
+  ctx = schedule_laggy_clients_disconnect(ctx);
 
   librados::AioCompletion *comp =
     librados::Rados::aio_create_completion(ctx, NULL,
@@ -839,7 +845,7 @@ void JournalMetadata::handle_watch_reset() {
     if (r == -ENOENT) {
       ldout(m_cct, 5) << __func__ << ": journal header not found" << dendl;
     } else {
-      lderr(m_cct) << __func__ << ": failed to watch journal"
+      lderr(m_cct) << __func__ << ": failed to watch journal: "
                    << cpp_strerror(r) << dendl;
     }
     schedule_watch_reset();
@@ -1021,6 +1027,59 @@ void JournalMetadata::async_notify_update() {
 
 void JournalMetadata::handle_notified(int r) {
   ldout(m_cct, 10) << "notified journal header update: r=" << r << dendl;
+}
+
+Context *JournalMetadata::schedule_laggy_clients_disconnect(Context *on_finish) {
+  assert(m_lock.is_locked());
+
+  ldout(m_cct, 20) << __func__ << dendl;
+
+  if (m_settings.max_concurrent_object_sets <= 0) {
+    return on_finish;
+  }
+
+  Context *ctx = on_finish;
+
+  for (auto &c : m_registered_clients) {
+    if (c.state == cls::journal::CLIENT_STATE_DISCONNECTED ||
+	c.id == m_client_id ||
+	m_settings.whitelisted_laggy_clients.count(c.id) > 0) {
+      continue;
+    }
+    const std::string &client_id = c.id;
+    uint64_t object_set = 0;
+    if (!c.commit_position.object_positions.empty()) {
+      auto &position = *(c.commit_position.object_positions.begin());
+      object_set = position.object_number / m_splay_width;
+    }
+
+    if (m_active_set > object_set + m_settings.max_concurrent_object_sets) {
+      ldout(m_cct, 1) << __func__ << ": " << client_id
+		      << ": scheduling disconnect" << dendl;
+
+      ctx = new FunctionContext([this, client_id, ctx](int r1) {
+          ldout(m_cct, 10) << __func__ << ": " << client_id
+                           << ": flagging disconnected" << dendl;
+
+          librados::ObjectWriteOperation op;
+          client::client_update_state(&op, client_id,
+                                      cls::journal::CLIENT_STATE_DISCONNECTED);
+
+          librados::AioCompletion *comp =
+              librados::Rados::aio_create_completion(ctx, nullptr,
+                                                     utils::rados_ctx_callback);
+          int r = m_ioctx.aio_operate(m_oid, comp, &op);
+          assert(r == 0);
+          comp->release();
+	});
+    }
+  }
+
+  if (ctx == on_finish) {
+    ldout(m_cct, 20) << __func__ << ": no laggy clients to disconnect" << dendl;
+  }
+
+  return ctx;
 }
 
 std::ostream &operator<<(std::ostream &os,
--- a/src/librbd/ImageCtx.cc
+++ b/src/librbd/ImageCtx.cc
@@ -945,7 +945,8 @@ struct C_InvalidateCache : public Context {
         "rbd_journal_object_flush_age", false)(
         "rbd_journal_pool", false)(
         "rbd_journal_max_payload_bytes", false)(
-        "rbd_journal_max_concurrent_object_sets", false);
+        "rbd_journal_max_concurrent_object_sets", false)(
+        "rbd_mirroring_resync_after_disconnect", false);
 
     md_config_t local_config_t;
     std::map<std::string, bufferlist> res;
@@ -1002,6 +1003,7 @@ struct C_InvalidateCache : public Context {
     ASSIGN_OPTION(journal_pool);
     ASSIGN_OPTION(journal_max_payload_bytes);
     ASSIGN_OPTION(journal_max_concurrent_object_sets);
+    ASSIGN_OPTION(mirroring_resync_after_disconnect);
   }
 
   ExclusiveLock<ImageCtx> *ImageCtx::create_exclusive_lock() {
--- a/src/librbd/ImageCtx.cc
+++ b/src/librbd/ImageCtx.cc
@@ -944,7 +944,8 @@ struct C_InvalidateCache : public Context {
         "rbd_journal_object_flush_bytes", false)(
         "rbd_journal_object_flush_age", false)(
         "rbd_journal_pool", false)(
-        "rbd_journal_max_payload_bytes", false);
+        "rbd_journal_max_payload_bytes", false)(
+        "rbd_journal_max_concurrent_object_sets", false);
 
     md_config_t local_config_t;
     std::map<std::string, bufferlist> res;
@@ -1000,6 +1001,7 @@ struct C_InvalidateCache : public Context {
     ASSIGN_OPTION(journal_object_flush_age);
     ASSIGN_OPTION(journal_pool);
     ASSIGN_OPTION(journal_max_payload_bytes);
+    ASSIGN_OPTION(journal_max_concurrent_object_sets);
   }
 
   ExclusiveLock<ImageCtx> *ImageCtx::create_exclusive_lock() {
--- a/src/test/journal/RadosTestFixture.cc
+++ b/src/test/journal/RadosTestFixture.cc
@@ -68,10 +68,12 @@ int RadosTestFixture::create(const std::string &oid, uint8_t order,
 
 journal::JournalMetadataPtr RadosTestFixture::create_metadata(
     const std::string &oid, const std::string &client_id,
-    double commit_interval, uint64_t max_fetch_bytes) {
+    double commit_interval, uint64_t max_fetch_bytes,
+    int max_concurrent_object_sets) {
   journal::Settings settings;
   settings.commit_interval = commit_interval;
   settings.max_fetch_bytes = max_fetch_bytes;
+  settings.max_concurrent_object_sets = max_concurrent_object_sets;
 
   journal::JournalMetadataPtr metadata(new journal::JournalMetadata(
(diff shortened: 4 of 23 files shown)


dillaman:
@trociny Instead of unregistering the laggy client (which would work), I put a placeholder ```ClientState``` in the registration record [1]. My original idea was to set this to DISCONNECTED (via a new cls method) and adjust the JournalTrimmer to ignore clients with the state set to DISCONNECTED when determining the minimum set. 

The rbd-mirror daemon can then detect if and when it's client state is set to DISCONNECTED, stop replay (if in-progress), and re-initiate an image sync w/o the need to split-brain.  The sync point record supports a "from snapshot" so it can be modified to only resync from the last known good snapshot (if any). For example, if the system gets disconnected during the initial sync, it can recover w/o starting over from scratch by reconnecting to the journal, creating a second sync point record tied to the first, and once the first sync is complete, it would start with the second (similar to how VM/storage live migration transfers the full base, and then one or more deltas until it can cut over).

The iterative / recoverable image sync wouldn't be part of this effort, but I wanted to lay out my full plans.

[1] https://github.com/ceph/ceph/blob/master/src/cls/journal/cls_journal_types.h#L79

trociny:
@dillaman Got it! Thanks. Will redo.

trociny:
@dillaman How do you like this version?

src/common/config_opts.h
```diff
@@ -1250,6 +1250,7 @@ OPTION(rbd_journal_object_flush_interval, OPT_INT, 0) // maximum number of pendi
 OPTION(rbd_journal_object_flush_bytes, OPT_INT, 0) // maximum number of pending bytes per journal object
 OPTION(rbd_journal_object_flush_age, OPT_DOUBLE, 0) // maximum age (in seconds) for pending commits
 OPTION(rbd_journal_pool, OPT_STR, "") // pool for journal objects
+OPTION(rbd_journal_client_object_sets_behind_max, OPT_INT, 0) // maximum number of object sets a journal client can be behind before it is automatically unregistered
```
dillaman: Minor: perhaps ```rbd_journal_max_concurrent_object_sets```?

src/journal/JournalMetadata.cc
```diff
@@ -748,7 +751,13 @@ void JournalMetadata::handle_refresh_complete(C_Refresh *refresh, int r) {
 
     Client client(m_client_id, bufferlist());
     RegisteredClients::iterator it = refresh->registered_clients.find(client);
-    if (it != refresh->registered_clients.end()) {
+    if (it == refresh->registered_clients.end()) {
+      lderr(m_cct) << "failed to locate client: " << m_client_id << dendl;
+      r = -ENOENT;
+    } else if (it->state == cls::journal::CLIENT_STATE_DISCONNECTED) {
+      lderr(m_cct) << "client flagged disconnected: " << m_client_id << dendl;
+      r = -ENOTCONN;
```
dillaman: Even if the client isn't connected, I think you would still want to initialize the journaler. rbd-mirror, during bootstrap can check its connected and use a helper method to re-connect if disconnected and within ImageReplayer it should have a listener for journal updates to check if it has been disconnected while replaying.

src/journal/JournalMetadata.cc
```diff
@@ -985,6 +991,8 @@ void JournalMetadata::committed(uint64_t commit_tid,
 
     ldout(m_cct, 20) << "updated commit position: " << commit_position << ", "
                      << "on_safe=" << m_commit_position_ctx << dendl;
+
+    disconnect_laggy_clients();
```
dillaman: I would only run this logic as part of the scheduled commit task.  The JournalTrimmer watches for header updates before it proceeds with the trim, so if it's part of the scheduled commit task, it could update the commit position, disconnect the laggy peers, and throw the update event for the Trimmer to detect the newly disconnected client.

src/journal/JournalMetadata.cc
```diff
@@ -1022,6 +1030,79 @@ void JournalMetadata::handle_notified(int r) {
...
+    ctx = new FunctionContext([this, client_id, ctx](int r1) {
+	ldout(m_cct, 1) << __func__ << ": " << client_id
+		        << ": flagging disconnected" << dendl;
+
+	librados::ObjectWriteOperation op;
+	client::client_update_state(&op, client_id,
+				    cls::journal::CLIENT_STATE_DISCONNECTED);
+
+	librados::AioCompletion *comp =
+	librados::Rados::aio_create_completion(ctx, nullptr,
```
dillaman: ```C_NotifyUpdate``` could be re-used multiple times.

src/journal/JournalMetadata.cc
```diff
@@ -1022,6 +1030,79 @@ void JournalMetadata::handle_notified(int r) {
...
+					       utils::rados_ctx_callback);
+	int r = m_ioctx.aio_operate(m_oid, comp, &op);
+	assert(r == 0);
+	comp->release();
+      });
+
+    clients.pop();
+  }
+
+  assert(clients.size() == 1);
```
dillaman: Why the loop and then this extra, duplicate code block?

src/journal/JournalMetadata.cc
```diff
@@ -1022,6 +1030,79 @@ void JournalMetadata::handle_notified(int r) {
...
+
+  ldout(m_cct, 20) << __func__ << dendl;
+
+  if (m_client_object_sets_behind_max <= 0) {
+    return;
+  }
+
+  std::queue<std::string> clients;
+  for (auto &c : m_registered_clients) {
+    if (c.id == m_client_id ||
```
dillaman: I think we need a configurable filter to exclude certain peers from being disconnected.  We would never want to disconnect the ```librbd::Journal::IMAGE_CLIENT_ID```, only other clients.

src/tools/rbd_mirror/image_replayer/BootstrapRequest.cc
```diff
@@ -194,6 +194,10 @@ void BootstrapRequest<I>::handle_get_client(int r) {
     derr << ": failed to retrieve client: " << cpp_strerror(r) << dendl;
     finish(r);
     return;
+  } else if (m_client.state == cls::journal::CLIENT_STATE_DISCONNECTED) {
+    dout(0) << ": client flagged disconnected, re-registering" << dendl;
+    unregister_client();
```
dillaman: So long as this doesn't result in a split-brain, this is probably fine for short term. I think the intermediate term goal would be to set the ```MirrorPeerState``` state back to ```MIRROR_PEER_STATE_SYNCING``` to indicate that a full re-sync is required and re-connect the client. The long term goal would add the smarts to have multiple sync points / incremental syncs upon disconnect.

dillaman:
@trociny Looking good to me.  The only missing piece is the smarts in ```rbd::mirror::ImageReplayer``` to catch a disconnect while we are actively replaying.

src/journal/JournalMetadata.cc
trociny: Because in the loop, I am building the code into the context for all clients but one, while below I execute the code for the last client passing the built context.

Do you suggest just to build the context for all clients, and then run `ctx->complete(0)`?

dillaman: Oh -- I missed the FunctionContext.  Maybe move the bulk librados code section to a helper method so you can re-use it.  You also should be able to just keep adding ```client::client_update_state``` ops to a single ```librados::ObjectWriteOperation``` since each will update different omap keys.

trociny:
@dillaman Updated, still not tested well. Also there are potential issues to discuss:

- A laggy client is disconnected, resync is started, the client is re-registered. While it is resyncing the master commit position grows and the mirror client is unregistered again before resync is complete. Should we add a way the rbd-mirror could tell that the client cannot be disconnected (during  cync/bootstrap)?

- What to do when disconnect is detected when replaying (or just after resync)? Doing resync immediately does not look like a good idea (mirroring is likely overloaded, and resync will only increase the load). Currently it just stops image replay (though without manual flag, so will be restarted on the next pool rescan). May be schedule resync after configurable interval?

trociny:
Added some tests.  Also, experimental `rbd journal disconnect` command is added. It is used in functional tests, not sure though it is useful enough to add it upstream. @dillaman do you think it would be useful  to have such command? It could be also implemented as `rbd mirror image disconnect`.

src/journal/JournalMetadata.cc
```diff
@@ -1023,6 +1031,60 @@ void JournalMetadata::handle_notified(int r) {
...
+  for (auto &c : m_registered_clients) {
+    if (c.state == cls::journal::CLIENT_STATE_DISCONNECTED ||
+	c.id == m_client_id ||
+	m_settings.whitelisted_laggy_clients.count(c.id) > 0) {
+      continue;
+    }
+    const std::string &client_id = c.id;
+    uint64_t object_set = 0;
+    if (!c.commit_position.object_positions.empty()) {
+      auto position = *(c.commit_position.object_positions.begin());
```
dillaman: Minor: ```auto &```

src/test/journal/RadosTestFixture.cc
```diff
@@ -68,10 +68,12 @@ int RadosTestFixture::create(const std::string &oid, uint8_t order,
 
 journal::JournalMetadataPtr RadosTestFixture::create_metadata(
     const std::string &oid, const std::string &client_id,
-    double commit_interval, uint64_t max_fetch_bytes) {
+    double commit_interval, uint64_t max_fetch_bytes,
+    int max_concurrent_object_sets) {
   journal::Settings settings;
   settings.commit_interval = commit_interval;
   settings.max_fetch_bytes = max_fetch_bytes;
+  settings.max_concurrent_object_sets= max_concurrent_object_sets;
```
dillaman: Minor: space before ```=```.

src/tools/rbd_mirror/ImageReplayer.cc
```diff
@@ -500,6 +510,26 @@ void ImageReplayer<I>::handle_init_remote_journaler(int r) {
...
+  if (r < 0) {
+    derr << "error retrieving remote journal client: " << cpp_strerror(r)
+	 << dendl;
+    on_start_fail(r, "error retrieving remote journal client");
+    return;
+  }
+
+  dout(20) << "client=" << client << dendl;
+
+  if (client.state != cls::journal::CLIENT_STATE_CONNECTED) {
```
dillaman: I would put this logic into the bootstrap state machine since it already pulls the remote peer record -- and eventually it will support auto-resyncing.

src/tools/rbd_mirror/ImageReplayer.cc
```diff
@@ -1437,6 +1468,25 @@ void ImageReplayer<I>::handle_shut_down(int r, Context *on_start) {
...
+  if (r < 0) {
+    derr << "failed to retrieve client: " << cpp_strerror(r) << dendl;
+    return;
+  }
+
+  dout(20) << "client=" << client << dendl;
+
+  if (client.state != cls::journal::CLIENT_STATE_CONNECTED) {
+    dout(0) << "client flagged disconnected, stopping image replay" << dendl;
+    stop(); // XXX: Schedule re-sync after configurable interval?
```
dillaman: Should this stop with an error code so that ``` rbd mirror image status``` reports that it has been disconnected?

src/tools/rbd_mirror/image_replayer/BootstrapRequest.cc
dillaman: I think you can just fail with some special result code that ```ImageReplayer``` can use to know that it was disconnected and should stop (with error).

src/tools/rbd/action/Journal.cc
```diff
@@ -973,6 +1062,10 @@ Shell::Action action_reset(
   {"journal", "reset"}, {}, "Reset image journal.", "",
   &get_reset_arguments, &execute_reset);
 
+Shell::Action action_disconnect(
+  {"journal", "disconnect"}, {}, "Flag image journal client disconnected.", "",
```
dillaman: How about "journal client disconnect"? Since I can also see the debugging need for "journal client remove"

src/tools/rbd_mirror/ImageReplayer.cc
trociny: @dillaman  Do you mean "check if it is disconnected" logic? It is here, after add_listener, just to recheck that the client was not disconnected in time window between we checked the state in bootstrap and registered the listener (so we lost the notification). Or do you want the listener is registered in bootstrap?

src/tools/rbd_mirror/image_replayer/BootstrapRequest.cc
trociny: So, do we want that after disconnect the mirror does not reconnect automatically?

Right now, if "disconnected" is detected during replaying, the image repayer stops (without "manual" flag), so it is restarted in max 30 second interval by pool Replayer. And in bootstrap we reregister journal client, so the restart succeeds.

Do you mean the bootstap always returns error here and the "start" is terminated, similarly to "remote image is non-primary or local image is primary" case? So a user intervention is required to manually start image replay. And then how user is supposed to do this? Don't we want an automatic restart, may be after some configurable time interval?

src/tools/rbd_mirror/ImageReplayer.cc
dillaman: Missed that -- I thought that the remote journal was initialized first and then bootstrap was invoked.

src/tools/rbd_mirror/image_replayer/BootstrapRequest.cc
dillaman: Instead of unregister / re-register, why not just update the connected state back to connected (and have the cls method clear the current commit position). How about a new config value for whether or not to automatically resync after a disconnect? Until we get the delta image sync logic, I'd hate to have an overloaded system get behind on replay and then just automatically start a full resync.

src/journal/JournalMetadata.cc
```diff
@@ -986,6 +993,7 @@ void JournalMetadata::committed(uint64_t commit_tid,
 
     ldout(m_cct, 20) << "updated commit position: " << commit_position << ", "
                      << "on_safe=" << m_commit_position_ctx << dendl;
+
```
trociny: TODO: remove empty line

trociny:
@dillaman updated. Now when a journal client is disconnected, rbd-mirror just stops image replayer, until 'mirror image resync' command is issued by user. Also rbd_mirror_resync_after_disconnect configuration option is added to automatically start the resync.

src/tools/rbd/action/Journal.cc
```diff
@@ -985,6 +1084,11 @@ Shell::Action action_import(
   {"journal", "import"}, {}, "Import image journal.", "",
   &get_import_arguments, &execute_import);
 
+Shell::Action action_disconnect(
+  {"journal", "client", "disconnect"}, {},
+  "Flag image journal client disconnected.", "",
```
dillaman: Minor: "... as disconnected"

src/tools/rbd_mirror/ImageReplayer.cc
```diff
@@ -503,6 +513,23 @@ void ImageReplayer<I>::handle_init_remote_journaler(int r) {
...
+  if (r < 0) {
+    derr << "error retrieving remote journal client: " << cpp_strerror(r)
+	 << dendl;
+    on_start_fail(r, "error retrieving remote journal client");
+    return;
+  }
+
+  if (client.state != cls::journal::CLIENT_STATE_CONNECTED) {
+    dout(5) << "client flagged disconnected, stopping image replay" << dendl;
+    on_start_fail(0, "disconnected");
```
dillaman: Should this set an error code so that "mirror image status" shows an error?

src/tools/rbd_mirror/ImageReplayer.cc
```diff
@@ -1442,6 +1473,23 @@ void ImageReplayer<I>::handle_shut_down(int r) {
...
+  cls::journal::Client client;
+  int r = m_remote_journaler->get_cached_client(m_local_mirror_uuid, &client);
+  if (r < 0) {
+    derr << "failed to retrieve client: " << cpp_strerror(r) << dendl;
+    return;
+  }
+
+  if (client.state != cls::journal::CLIENT_STATE_CONNECTED) {
+    dout(0) << "client flagged disconnected, stopping image replay" << dendl;
+    stop(nullptr, false, "disconnected");
```
dillaman: Same comment here: should we set an error code so that an admin can see the problem?

src/librbd/ImageCtx.cc
```diff
@@ -930,7 +930,8 @@ struct C_InvalidateCache : public Context {
         "rbd_journal_object_flush_age", false)(
         "rbd_journal_pool", false)(
         "rbd_journal_max_payload_bytes", false)(
-        "rbd_journal_max_concurrent_object_sets", false);
+        "rbd_journal_max_concurrent_object_sets", false)(
+        "rbd_mirror_resync_after_disconnect", false);
```
dillaman: No need to treat this as a librbd image option -- just use it from rbd-mirror daemon only

dillaman:
@trociny few minor comments, but otherwise lgtm

src/librbd/ImageCtx.cc
trociny: It is not that I insists, though I can imagine a situation when you want the most of the images not to resync after disconnect (in order not to increase the load) while still have a group of "important" images you want to resync ASAP. Setting this option for these images via image-meta makes this possible. Also it simplifies testing -- no need to restart daemon to apply the option change.

@dillaman If you think these arguments are not enough to change your opinion I will remove this.

More logically I think it would be to have something like `rbd mirror image set $image resync_after_disconnect true`, but I guess it is not worth the effort, at least on this stage...

dillaman: Valid point -- it still might be a little confusing to have a "per-image" config option named with the same prefix as "rbd_mirror"-specific config options. Perhaps ```rbd_resync_to_primary_mirror_after_disconnect``` or something similar?

trociny:
@dillaman Updated. I think I addressed all your comments. The only thing is renaming `rbd_mirror_resync_after_disconnect` config option to `rbd_resync_to_primary_mirror_after_disconnect`. I agree that using `rbd_mirror` prefix is unfortunate here, still I am not very happy with `rbd_resync_to_primary_mirror_after_disconnect` name. I changed it to `rbd_mirroring_resync_after_disconnect` instead. What do you think? If you still like `rbd_resync_to_primary_mirror_after_disconnect` more I will change it.

dillaman:
@trociny That's fine with me -- just wanted to ensure it didn't overlap with "rbd_mirror".

dillaman:
@trociny rebase required

trociny:
@dillaman Testing after rebase I noticed the issue with `ImageReplayer<I>::handle_remote_journal_metadata_updated()`: it might call when ImageReplayer was stopping and shutting journal down, and metadata was null so it crashed.

I have updated handle_remote_journal_metadata_updated to check the current ImageReplayer state and return if it is not running.

Now, retesting this locally. I will let you know about test results.

trociny:
@dillaman it passed local tests

PR #10382, opened 2016-07-21
Add ID for Ceph Encrypt

cholcombe973:
I've added an --id field to the ceph-disk prepare function so that a cephx key other than admin can be used when setting up dmcrypt.  If the admin key is not present on the osd cluster ceph-disk prepare fails with a permission denied error.

### DIFF
--- a/src/ceph-disk/ceph_disk/main.py
+++ b/src/ceph-disk/ceph_disk/main.py
@@ -1728,6 +1728,11 @@ class Prepare(object):
             help='unique OSD uuid to assign this disk to',
         )
         parser.add_argument(
+            '--id',
+            metavar='cephx_user',
+            help='Initial cephx user to create lockbox cephx with',
+        )
+        parser.add_argument(
             '--dmcrypt',
             action='store_true', default=None,
             help='encrypt DATA and/or JOURNAL devices with dm-crypt',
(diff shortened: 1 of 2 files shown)


cholcombe973:
Looks like both check fails are coming from BitAllocator.cc from bluestore

tchaikov:
@dachary what do you think?

src/ceph-disk/ceph_disk/main.py
```diff
@@ -2257,26 +2262,34 @@ def create_key(self):
...
-                'ceph',
-                'auth',
-                'get-or-create',
-                'client.osd-lockbox.' + self.args.osd_uuid,
-                'mon',
-                ('allow command "config-key get" with key="dm-crypt/osd/' +
-                 self.args.osd_uuid + '/luks"'),
-            ],
-        )
+        config_key_cmd = [
```
tchaikov: nit, could also have a variable named "ceph_cmd", set it to `['ceph']` `if self.args.id`, and `['ceph', '--id', self.args.id]`, otherwise.
and a variable named `config_key_args` with the args for "config-keys". and then call `command_check_call(ceph_cmd + config_key_args)`.

dachary:
It looks good to me. @cholcombe973 could you also add a minimal test in qa/workunits/ceph-disk/ceph-disk-test.py ?

src/ceph-disk/ceph_disk/main.py
dachary: I agree with @tchaikov that it would make the code a little easier to read

cholcombe973:
Thanks for the quick feedback!  I'll rework the patch a bit and update this PR. :)

cholcombe973:
Diff is updated.  Just waiting on CI to go through

tchaikov:
@cholcombe973 
could you prefix the title of your commit message with the subcomponent your are changing ? see https://github.com/ceph/ceph/blob/master/SubmittingPatches.rst#preparing-and-sending-patches, in this case, you might want to add a prefix of "ceph-disk: ".

tchaikov:
and as @dachary suggested, we could have a minimal test for this feature in `qa/workunits/ceph-disk/ceph-disk-test.py`.

cholcombe973:
@tchaikov yeah no problem.  I'll work on that today

cholcombe973:
@tchaikov I added a tiny test but I'm not entirely sure how to run it.  The cephx key needs to be created by something before we can use it in the test.  Where would that need to happen?  I've not used teuthology before unfortunately.

dachary:
@cholcombe973 you can run it on a virtual machine or bare metal with 3 disks attached, each with at least 20GB, as root and with a Ceph monitor installed. This setup is automatically done by teuthology but if you don't know teuthology, it's probably simpler if you create it manually. Then you can run the test as explained at the beginning of the script.

dachary:
@cholcombe973 ping ?

cholcombe973:
@dachary sorry I haven't had time to circle back and setup the VM to test this.

dachary:
implemented and tested at https://github.com/ceph/ceph/pull/11996

cholcombe973:
@dachary thanks!  Much appreciated

PR #10384, opened 2016-07-21
common/ceph_context.cc: Use CEPH_DEV to reduce logfile noise

wjwithagen:
Logfiles contain a lot of warnings about features set to '*' 
Setting the CEPH_DEV option allows suppression of these warnings.
Can be set in vstart.sh for example.

### DIFF
--- a/src/common/ceph_context.cc
+++ b/src/common/ceph_context.cc
@@ -305,6 +305,17 @@ public:
 	conf->enable_experimental_unrecoverable_data_corrupting_features,
 	cct->_experimental_features);
       ceph_spin_unlock(&cct->_feature_lock);
+      if (getenv("CEPH_DEV") == NULL) {
+        if (!cct->_experimental_features.empty()) {
+          if (cct->_experimental_features.count("*")) {
+            lderr(cct) << "WARNING: all dangerous and experimental features are enabled." << dendl;
+          } else {
+            lderr(cct) << "WARNING: the following dangerous and experimental features are enabled: "
+              << cct->_experimental_features << dendl;
+          }
+        }
+      }
+
     }
     if (changed.count("crush_location")) {
       cct->crush_location.update_from_conf();


src/common/ceph_context.cc
```diff
@@ -254,9 +254,11 @@ class CephContextObs : public md_config_obs_t {
 	conf->enable_experimental_unrecoverable_data_corrupting_features,
 	cct->_experimental_features);
       ceph_spin_unlock(&cct->_feature_lock);
+#if defined(NDEBUG)
```
tchaikov: NDEBUG is defined for non-DEBUG builds, so if we want to respect this macro, better of putting
```c
#ifndef NDEBUG
```

and i believe this warning is intentional, and helpful to our end users. it makes sure they are aware of the risk of using experimental features.

wjwithagen:
@tchaikov 
I was more of the opinion that when NDEBUG is set, it is production code.
And then you'd like to be warned if the flags/options are in a dangerous state.
Where as during a run with debugcode on, you know that there could be things that allow you to shoot your foot.

Reason for this is to reduce the noise in the logfiles in the tests.
Python scripts have a more or less equal annoyance, that I'd like to shut up as well.

wjwithagen:
retest this please

src/common/ceph_context.cc
wjwithagen: @tchaikov 
So we 100% agree. 
- it is for the end users
- it is NonDebug

So if I want to suppress this during developement, because it does not conviene any information to developers other than logfile pollution, it think it needs to be included in a:
```
#ifdef NDEBUG
#endif
```

wjwithagen:
@tchaikov 
Rewrote this to start using CEPH_DEV, which I intend to use at a few more locations to reduce logfile noise.

dachary:
genuine error at https://jenkins.ceph.com/job/ceph-pull-requests/14601/console

wjwithagen:
@dachary 
Yup, :( Always happens when you do thing while you're already with one foot out of the door.
Will fix.

wjwithagen:
@dachary 
I would like to do something similar to build/bin/ceph and the warning
'DEVMODEMSG = *** DEVELOPER MODE: setting PATH, PYTHONPATH and LD_LIBRARY_PATH ***'
Because that is also a rather serious spammer in logfiles, and I think developers will be able to figure this out by them selves??
Something like 'os.environ.has_key(CEPH_DEV)'

Is that a plan?

dachary:
@wjwithagen I'm not sure, it depends on how developers feel about such a change. I don't mind the spam but I have a high tolerance for it. Maybe starting a thread on ceph-devel is the way to go ?

liewegas:
What about instead setting an environment variable (CEPH_DEV=1 or CEPH_DEV_UNSAFE=1)?  That way vstart.sh can just set it and all the annoying warnings will go away, but production deployments (regardless of how the binary was built) will still be noisy.  I'm not sure I like the idea of compiling out the warnings entirely, even for debug builds, as debug builds can also get deployed for various reasons.

wjwithagen:
@liewegas 
That would be possible too, But then I think we need to commit the other PR first.
Once we have that, I'll also take a whack at ceph.py since that nags too about being run as developer, and that seriously clogs up all script testing logs.
Instead of death by chocolate, it is death by logs. ;-)

PR #10386, opened 2016-07-21
Fix attribute handling at lookup time

jtlayton:
This should be a fix for the bug that Frank reported here:

    http://tracker.ceph.com/issues/16668

The root cause of the problem is due to several factors:

1. _lookup doesn't check for the correct caps before filing out the stat structure
2. when an inode has multiple dentries, leases on the other dentries are not revoked after unlinking one
3. the MDS doesn't send any is_target info in the UNLINK reply trace

All of this means that when you remove one link of a hardlinked inode and then do a ll_lookup on the remaining dentry, it will end up filling out the struct stat with stale attributes.

This also has a fix for a minor bug in the handling of the ctime in fill_stat.

Teuthology run is in progress here:

http://pulpito.ceph.com/jlayton-2016-07-21_07:31:09-fs-wip-jlayton-nlink---basic-mira/


jtlayton:
Rebase onto current master.

gregsfortytwo:
Code looks great. But I'm a little confused -- is "parm" an abbreviation I've never seen before or a very common typo for you, @jtlayton?

gregsfortytwo:
Oh, and can you tag the appropriate commit with the Fixes: line for that fracker bug.

jtlayton:
Yeah, "parm" is just an abbreviation of "parameter". I'll make it a bit more clear in the changelogs.

jtlayton:
I've also kicked off a couple of "fs" test suite runs, but a lot of the tests in there are failing, and I don't think it's due to any of the changes in this PR. I'm doing a control run now from the head of the tree sans these patches to see if it fares any better.

jtlayton:
Found a bug in one of the commits, so I'm rerunning the tests now.

jtlayton:
Ran a "fs" test run over the weekend, and all but 9 tests passed. Most of the failures were SSH errors, but there were a few others too:

http://pulpito.ceph.com/jlayton-2016-07-29_18:51:42-fs-wip-jlayton-nlink---basic-mira/

I opened bugs for most of the failures and kicked off a new set of tests for the failing runs. Those all passed, except for one:

http://pulpito.ceph.com/jlayton-2016-08-01_04:07:30-fs-wip-jlayton-nlink---basic-mira/

...so I kicked that one off again after rebaing onto the current master branch, and it still fails with the same error. This is a new forward scrub test however, so I suspect that something is broken in the test itself (or maybe the tighter caps handling in this PR is throwing off the results?).

I'm rerunning the test again now without any of my patches now to see whether it passes.

jtlayton:
The only test that is consistently failing also fails with a recent pull from the master branch. It's an MDS related patch and this is strictly a client-side pull request (now). So, I don't think that's a regression in this set.

gregsfortytwo:
@jtlayton, you mean the "wrongly marked free" inode on scrub, right?

@jcsp ran c24d48071e460702cbb258d6872d51e205728f60 (newer than your wip-jlayton-control's 1194331665ffe878acfd3b24b8ad8a88961d2d9d) and it isn't failing there, although I'm not sure what merged PR might have resolved it. I suggest you wait for @jcsp's next test branch if you're sure this isn't something you did, or else rebase/merge with more current master.

src/client/Client.cc
```diff
@@ -5959,8 +5959,9 @@ int Client::_lookup(Inode *dir, const string& dname, int mask,
       }
       // dir lease?
       if (dir->caps_issued_mask(CEPH_CAP_FILE_SHARED)) {
-	if (dn->cap_shared_gen == dir->shared_gen)
-	  goto hit_dn;
+	if (dn->cap_shared_gen == dir->shared_gen &&
+	    (!dn->inode || dn->inode->caps_issued_mask(mask)))
+	      goto hit_dn;
```
gregsfortytwo: You appear to have some weird tabbing/spacing rules here, btw — not sure why this line is indented 6(?) spaces instead of 2.

doc/cephfs/capabilities.rst
```diff
@@ -0,0 +1,102 @@
...
+        2      4      6       8         20
+
+From the above, we get a number of constants, that are generated by taking
+each bit value and shifting to the correct bit in the word:
+
+::
+        #define CEPH_CAP_AUTH_SHARED  (CEPH_CAP_GSHARED  << CEPH_CAP_SAUTH)
+
+These bits can then be or'ed together to make a bitmask denoting a set of
+capabilities. Some of the logging code will
```
gregsfortytwo: lost sentence fragment?

doc/cephfs/capabilities.rst
```diff
@@ -0,0 +1,102 @@
...
+  check may require getting at ACLs as well, which are stored in xattrs.
+
+* LINK: the link count of the inode
+
+* XATTR: ability to access or manipulate xattrs. Note that since ACLs are
+  stored in xattrs, it's also sometimes necessary to access them when checking
+  permissions.
+
+* FLOCK: file locking capabilities. Covers the ability to access and cache
+  file locks.
```
gregsfortytwo: Uh. We don't actually use the FLOCK cap definitions and I have no idea when they were created. :) Our file locks are all synchronous to the MDS.

> gregf@rex004:~/ceph2 [master]$ git grep CEPH_CAP_SFLOCK
src/include/ceph_fs.h:#define CEPH_CAP_SFLOCK    20
src/include/ceph_fs.h:#define CEPH_CAP_FLOCK_SHARED  (CEPH_CAP_GSHARED   << CEPH_CAP_SFLOCK)
src/include/ceph_fs.h:#define CEPH_CAP_FLOCK_EXCL    (CEPH_CAP_GEXCL     << CEPH_CAP_SFLOCK)
gregf@rex004:~/ceph2 [master]$ git grep CEPH_CAP_FLOCK
src/include/ceph_fs.h:#define CEPH_CAP_FLOCK_SHARED  (CEPH_CAP_GSHARED   << CEPH_CAP_SFLOCK)
src/include/ceph_fs.h:#define CEPH_CAP_FLOCK_EXCL    (CEPH_CAP_GEXCL     << CEPH_CAP_SFLOCK)

doc/cephfs/capabilities.rst
jtlayton: Hah, yeah. I'll fix that up.

src/client/Client.cc
jtlayton: I'm not sure I see the problem there? The goto line is spaced in two spaces inside the conditional.

doc/cephfs/capabilities.rst
jtlayton: Ok, I thought I had seen one place where the FLOCK defs were used but it seems I'm mistaken. I'll remove it from the doc, and maybe we should go ahead and remove the #defines for it as well?

jtlayton:
Thanks @gregsfortytwo!

Ok, fixed up the doc to remove the dangling sentence fragment, and also purged all mention of the FLOCK permissions. I've also added a patch to remove the SFLOCK related constants. It's just confusing having that sort of cruft in there.

On the indentation in _lookup... I dunno, it looks fine to me. Could you send me a patch based upon this branch that shows how you think it ought to be indented?

src/client/Client.cc
gregsfortytwo: When I open this in my emacs and have it indent things it's indenting "goto hit_dn;" 2 spaces past the initial "if" line (it, before the "!dn->inode" bit); it looks like whatever setup you've got is indenting it 2 past the final line. *shrug*

PR #10387, opened 2016-07-21
tracing: link correct version of shared libs

neurodrone:
Presently `dlopen()` fails to find the correct shared library to link with because of a SONAME mismatch. This results in no LTTng-UST trace being generated even when `rbd_tracing` or `rados_tracing` is set to `true` in the `ceph.conf`. This change attempts to fix the above issue without hardcoding the actual versioned names in the code.

The shared object name for both the files are shown as below:

``` bash
$ readelf -d librbd.so.* | grep SONAME
 0x000000000000000e (SONAME)             Library soname: [librbd.so.1]

$ readelf -d librados.so.* | grep SONAME
 0x000000000000000e (SONAME)             Library soname: [librados.so.2]
```

r: @dillaman 

Signed-off-by: Vaibhav Bhembre vaibhav@digitalocean.com

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -329,6 +329,9 @@ if(${WITH_LTTNG})
   find_package(lttng-ust REQUIRED)
 endif(${WITH_LTTNG})
 
+set(LIBRBD_TP_SONAME "librbd_tp.so.1")
+set(LIBRADOS_TP_SONAME "librados_tp.so.2")
+
 #option for Babeltrace
 option(HAVE_BABELTRACE "Babeltrace libraries are enabled" ON)
 if(${HAVE_BABELTRACE})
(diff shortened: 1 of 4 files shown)


dillaman:
@neurodrone I was hoping to re-use the version constants between the source code and the linker settings (https://github.com/ceph/ceph/blob/master/src/tracing/CMakeLists.txt#L48) just to ensure that when one changes, the other automatically changes as well.

neurodrone:
That's a good point. It makes sense to use those rather than have those values in two places. I will look into this.

dillaman:
@neurodrone ping

neurodrone:
Ow, got side-tracked. I will get something in by this week.

tchaikov:
@neurodrone ping?

neurodrone:
I wanted to work on this but don't think I have cycles to spare this week at least. I will try to squeeze some time on this by Friday or on the weekend. Would that be fine?

tchaikov:
@neurodrone thanks!

dachary:
needs rebase

dachary:
feel free to reopen this pull requestion when you have time to work on it

PR #10389, opened 2016-07-21
hammer: Backport pull request #8210 from liewegas/wip-12814

dzafman:
I saw a hammer rados test fail because of tracker 12814


dzafman:
There are other dependencies needed.

PR #10393, opened 2016-07-22
reduce unnecessary mds log flush

ukernel:


src/mds/Server.cc
```diff
@@ -3779,7 +3779,6 @@ void Server::handle_client_setattr(MDRequestRef& mdr)
       mds->locker->drop_locks(mdr.get());
       mdr->drop_local_auth_pins();
       cur->add_waiter(CInode::WAIT_TRUNC, new C_MDS_RetryRequest(mdcache, mdr));
-      mds->mdlog->flush();
```
gregsfortytwo: I think we need to leave this flush in. Adding in an immediate flush once it's logged doesn't matter so much, but in this case isn't the client waiting for the update to be durable before it does anything?

src/mds/Server.cc
```diff
@@ -3839,7 +3839,8 @@ void Server::handle_client_setattr(MDRequestRef& mdr)
 								   truncating_smaller, changed_ranges));
 
   // flush immediately if there are readers/writers waiting
-  if (cur->get_caps_wanted() & (CEPH_CAP_FILE_RD|CEPH_CAP_FILE_WR))
+  if (xlocks.count(&cur->filelock) &&
+      (cur->get_caps_wanted() & (CEPH_CAP_FILE_RD|CEPH_CAP_FILE_WR)))
```
gregsfortytwo: Why do we only want to flush if *we* are xlocking? get_caps_wanted() is telling us that *somebody* wants Frw caps they don't have yet, and we want to service them, right?

src/mds/Locker.cc
```diff
@@ -2197,23 +2191,23 @@ bool Locker::check_inode_max_size(CInode *in, bool force_wrlock,
...
+  bool update_size = new_size > 0;
+  bool update_max = false;
   bool max_increased = false;
 
   if (update_size) {
     new_size = size = MAX(size, new_size);
     new_mtime = MAX(new_mtime, latest->mtime);
     if (latest->size == new_size && latest->mtime == new_mtime)
-      update_size = false;
+      update_size = 0;
```
gregsfortytwo: You set this up as a bool; please don't assign it numbers.

gregsfortytwo:
The rest looks good.

src/mds/Server.cc
ukernel: On Wed, Jul 27, 2016 at 5:35 AM, Gregory Farnum <notifications@github.com>
wrote:

> In src/mds/Server.cc
> <https://github.com/ceph/ceph/pull/10393#discussion_r72341542>:
>
> > @@ -3839,7 +3839,8 @@ void Server::handle_client_setattr(MDRequestRef& mdr)
> >  								   truncating_smaller, changed_ranges));
> >
> >    // flush immediately if there are readers/writers waiting
> > -  if (cur->get_caps_wanted() & (CEPH_CAP_FILE_RD|CEPH_CAP_FILE_WR))
> > +  if (xlocks.count(&cur->filelock) &&
> > +      (cur->get_caps_wanted() & (CEPH_CAP_FILE_RD|CEPH_CAP_FILE_WR)))
>
> Why do we only want to flush if *we* are xlocking? get_caps_wanted() is
> telling us that *somebody* wants Frw caps they don't have yet, and we
> want to service them, right?
>
No, get_caps_wanted()  return what caps clients want. It's irrelevant to
what caps clients have.

src/mds/Server.cc
ukernel: On Tue, Jul 26, 2016 at 1:50 PM, Gregory Farnum <notifications@github.com>
wrote:

> In src/mds/Server.cc
> <https://github.com/ceph/ceph/pull/10393#discussion_r72191154>:
>
> > @@ -3779,7 +3779,6 @@ void Server::handle_client_setattr(MDRequestRef& mdr)
> >        mds->locker->drop_locks(mdr.get());
> >        mdr->drop_local_auth_pins();
> >        cur->add_waiter(CInode::WAIT_TRUNC, new C_MDS_RetryRequest(mdcache, mdr));
> > -      mds->mdlog->flush();
>
> I think we need to leave this flush in. Adding in an immediate flush once
> it's logged doesn't matter so much, but in this case isn't the client
> waiting for the update to be durable before it does anything?
>
Truncating inode has 3 three phrases.  Journal truncate start, truncate
objects and journal truncate finish. Taking inode's xlock trigger log flush
for the first phrase. MDCache::truncate_inode_finish() calls log flush for
the third phrase.

gregsfortytwo: Why does taking the xlock initiate a flush in the first phase? I don't think that's a default behavior?

src/mds/Server.cc
gregsfortytwo: True. But we still want to flush if any other clients are waiting, not just the requesting client (who, after all, can make progress without the change being durably recorded). So we can be more precise than just looking at get_caps_wanted(), but I think checking if we xlocked is too specific?

jcsp:
@ukernel @gregsfortytwo looks like there are oustanding questions here, right?

src/mds/Server.cc
ukernel: Locker::xlock_start() calls nudge_log if it can't get the lock.

src/mds/Server.cc
ukernel: My point is, If setattr does not xlock filelock, It's nothing do with CEPH_CAP_FILE_FOO. why should it flush log for other operation.

src/mds/Server.cc
gregsfortytwo: Okay, I don't remember most of this (even the conversation) but that sounds good.

src/mds/Server.cc
gregsfortytwo: Mmm, true. I guess if we hit this case it would just be band-aiding some other missing wakeup we need to fix.

gregsfortytwo:
Reviewed-by:

jcsp:
Will put in next test branch then merge.

PR #10395, opened 2016-07-22
rgw:add a s3 API of make torrent for a object

zhouruisong:
When you execute the command gettorrent of a object, a torrent file will be produced and returned.
The torrent also will be save into a pool named default.rgw.torrent.
If the torrent of a object exists in default.rgw.torrent, it will be returned.

Signed-off-by: zhouruisong <236131368@qq.com>


zhouruisong:
@dang 
https://github.com/ceph/ceph/pull/9589 
this PR is closed and has some problem. I create a new PR, please merge this PR. thanks!

PR #10177, opened 2016-07-07
Port Python-based tests and remaining Python bindings to Python 3

BlaXpirit:
Compatibility with Python 2.6+ is kept.

The tests I've ran: https://gist.github.com/6ce78ee907b2d13a9fce8de859173dc8
(Probably best to *Download ZIP* and look at the text files locally)

PR #10398, opened 2016-07-22
rgw: make test_multi.py work with python3

theanalyst:


theanalyst:
tested locally, works for me on python2 and 3 right now

src/test/rgw/test_multi.py
```diff
@@ -824,7 +829,7 @@ def init(parse_args):
         with file(path) as f:
```
BlaXpirit: No `file` in Python 3

BlaXpirit:
I actually have a pull request for Python 3 compatibility going. As far as I see, there is no critical difference between them other than one comment I made.

https://github.com/ceph/ceph/pull/10177/files#diff-11

However, my local [testing](https://gist.githubusercontent.com/BlaXpirit/6ce78ee907b2d13a9fce8de859173dc8/raw/2ccdb23959f68f12f0c1024bf428099679efff1d/test_multi.py.3) worked only with `PYTHONHASHSEED=0`, so it would be interesting to confirm that I missed something making the PR or testing went wrong.

theanalyst:
@BlaXpirit yeah didn't see yours when I opened mine, I'll close this one after trying to find what is wrong :)

PR #10402, opened 2016-07-22
qa: Add ceph-ansible installer.

wusui:
Also fix iso rhel handling in openstack script
Updated openstack README.

Fixes: http://tracker.ceph.com/issues/16770
Signed-off-by: Warren Usui wusui@redhat.com

### DIFF
--- a/qa/qa_scripts/openstack/README
+++ b/qa/qa_scripts/openstack/README
@@ -10,9 +10,23 @@ If further development is necessary in the future, these templates should
 probably be removed and direct editing of the OpenStack conf files should
 probably be performed.
 
-These scripts also assume that there is a  rhel iso file named
-rhel-server-7.2-x86_64-boot.iso and a ceph iso file named
-rhceph-1.3.1-rhel-7-x86_64-dvd.iso in the files directory.  These iso files
-can be obtained from the rhel site and are not stored with these scripts.
+These scripts also assume that either there is a rhel iso file named
+rhel-server-7.2-x86_64-boot.iso in the user's home directory, or the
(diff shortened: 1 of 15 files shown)


tmuthamizhan:
in our lab setup, we have stock and non stock images[comes with RH subscription manager settings by default]
it would be nice to check and see if RH subscription manager is enabled on the test setup before enabling subscription manager settings [so as to avoid enabling RH subscription on an already enabled system]. I believe these are independent scripts designed to run on any test setup.

for testing ceph packages downstream, we don't need the following values to be set in "qa/qa_scripts/openstack/ceph_install_w_ansible/execs/ceph_ansible.sh" : ceph_origin=distro and
ceph_stable=true 

please reconsider hardcoding the values for "monitor_interface" and "public_network" as currently its values pertain to magna nodes.

wusui:
I am pushing a new version with tests for making sure the subscription manager is not run if already enabled.  Also, the monitor_interface and public_network values will be made configurable, using their current values as defaults if none are specified.  I will document this.

While retesting these changes, it appears that ceph_origin=distro and ceph_stable=true are both needed to get ceph-ansible to build properly.  The upstream value for ceph_origin breaks the ceph-ansible run almost immediately, and the ceph-stable value, when not set, seems to cause the OSD installations to hang.

I am going to push an new version that should be squished when checked in.

dachary:

liewegas:
Is this safe to merge?

tmuthamizhan:
@liewegas , i think this is safe to merge as it wont interfere with any existing tasks/suites.

PR #10403, opened 2016-07-22
make-dist/ceph.spec.in: Fix rpm build breakage.

ilc:
This allows the make-dist script to create a .spec file that
can actually use the tarball created to build rpms.

Tested on Fedora 24, using mock.


ilc:
Note: This will break the autotools creation of ceph.spec.

PR #10404, opened 2016-07-22
make-dist/ceph.spec.in: Fix srpm build breakage.

ilc:
This allows the make-dist script to create a .spec file that
can actually use the tarball created to build srpms.

Tested on Fedora 24, using mock.

### DIFF
--- a/ceph.spec.in
+++ b/ceph.spec.in
@@ -66,7 +66,7 @@ License:	LGPL-2.1 and CC-BY-SA-1.0 and GPL-2.0 and BSL-1.0 and GPL-2.0-with-auto
 Group:         System/Filesystems
 %endif
 URL:		http://ceph.com/
-Source0:	http://ceph.com/download/%{name}-%{version}.tar.bz2
+Source0:	http://ceph.com/download/@TARBALL_BASENAME@.tar.bz2
 %if 0%{?suse_version}
 %if 0%{?is_opensuse}
 ExclusiveArch:  x86_64 aarch64 ppc64 ppc64le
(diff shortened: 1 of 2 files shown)


ilc:
Note: This will break autotools use of ceph.spec.in.

Given that we want to remove autotools, I didn't address that issue.

tchaikov:
@ilc do you know where we are expected to create an srpm? just want to make sure how this change fit into our work flow. in https://github.com/ceph/ceph-build/blob/master/ceph-setup/build/build, maybe?

ilc:
@liewegas , given that you authored the script I'm fixing. Care to comment on the right thing to do here?

tchaikov:
@ktdreyer and @liewegas ping?

ktdreyer:
Can we fix `make-dist` to do the right thing instead?

liewegas:
@ilc is the problem that the tarbal includes the -nnn-g$sha1 portion, and for rpm we ar eputting that not in the version string but the release string?  Can we just include %release in the url? e.g.,

-Source0:	http://ceph.com/download/%{name}-%{version}.tar.bz2
+Source0:	http://ceph.com/download/%{name}-%{version}-%{release}.tar.bz2

liewegas:
Oh I guess not since we have to munge - to . in the release string.  Yeah, this seems okay to me.

@ktdreyer I think we really do want to stick with creating tarballs with a full version (git describe) string (by default at least).

ktdreyer:
Ok, if we're making the change consciously, then this is ok with me.

tchaikov:
@ilc could you rebase your commit against master to resolve the conflict?

ilc:
I'll fix it.  no problem.  Just need a day. :)

ilc:
@tchaikov Enjoy.
@liewegas Yep, you figured out why I had to make it ugly :(.

PR #10406, opened 2016-07-22
[DNM] Garbage Collection for partially overlapped blobs (with overlap count equal to 3)

Roushan81:
Because of overwrite operation, a new blob is created instead of merging data with the old blob. This leads to overlapping of blobs(example given below).
         ...................
            ......................
                 ......................
Because of overlapping blobs, space is wasted and read-performance can degrade. As part of this code change, we try to garbage collect the overlapping blobs. If the new write request results in overlapping count of more than 3, we extent the write request so that it completely overwrite the blobs adjacent to the new write request.

To keep track of the overlapping count, a new persistent field "blob_depth" is added to lextent. Whenever we add a new lextent, the "blob_depth" is set to the max of all the overlapping lextents + 1 (In fact we need not check all the overlapping lextents, we just need to check only two lextents covering the two ends of the new lextents)  . This will not always give the correct depth. It may over count in certain scenarios but will never undercount. Note: if lextents are splitted at the blobs boundary, it may be possible to correctly keep track of the depth, but that will increase the lextent counts and make "depth" update operation complicated.

### DIFF
--- a/src/os/bluestore/BlueStore.cc
+++ b/src/os/bluestore/BlueStore.cc
@@ -1571,6 +1571,7 @@ void BlueStore::_init_logger()
   b.add_u64(l_bluestore_compressed, "bluestore_compressed", "Sum for stored compressed bytes");
   b.add_u64(l_bluestore_compressed_allocated, "bluestore_compressed_allocated", "Sum for bytes allocated for compressed data");
   b.add_u64(l_bluestore_compressed_original, "bluestore_compressed_original", "Sum for original bytes that were compressed");
+  b.add_u64(l_bluestore_gc_bytes, "garbage_collected_bytes", "Sum for all garbage collected bytes");
   logger = b.create_perf_counters();
   g_ceph_context->get_perfcounters_collection()->add(logger);
 }
@@ -6058,7 +6059,7 @@ void BlueStore::_do_write_small(
       b->dirty_blob().calc_csum(b_off, padded);
       dout(20) << __func__ << "  lexold 0x" << std::hex << offset << std::dec
 	       << ": " << ep->second << dendl;
-      bluestore_lextent_t lex(blob, b_off + head_pad, length);
+      bluestore_lextent_t lex(blob, b_off + head_pad, length, wctx->le_depth);
       o->onode.set_lextent(offset, lex, &b->dirty_blob(), &wctx->lex_old);
       b->dirty_blob().mark_used(lex.offset, lex.length, min_alloc_size);
       txc->statfs_delta.stored() += lex.length;
@@ -6130,7 +6131,7 @@ void BlueStore::_do_write_small(
       dout(20) << __func__ << "  wal write 0x" << std::hex << b_off << "~"
 	       << b_len << std::dec << " of mutable " << blob << ": " << *b
 	       << " at " << op->extents << dendl;
-      bluestore_lextent_t lex(blob, offset - bstart, length);
+      bluestore_lextent_t lex(blob, offset - bstart, length, wctx->le_depth);
       o->onode.set_lextent(offset, lex, &b->dirty_blob(), &wctx->lex_old);
       b->dirty_blob().mark_used(lex.offset, lex.length, min_alloc_size);
       txc->statfs_delta.stored() += lex.length;
@@ -6149,7 +6150,7 @@ void BlueStore::_do_write_small(
   uint64_t b_off = P2PHASE(offset, alloc_len);
   _buffer_cache_write(txc, b, b_off, bl, wctx->buffered ? 0 : Buffer::FLAG_NOCACHE);
   _pad_zeros(&bl, &b_off, block_size);
-  bluestore_lextent_t lex(b->id, P2PHASE(offset, alloc_len), length);
+  bluestore_lextent_t lex(b->id, P2PHASE(offset, alloc_len), length, wctx->le_depth);
   o->onode.set_lextent(offset, lex, &b->dirty_blob(), &wctx->lex_old);
   txc->statfs_delta.stored() += lex.length;
   dout(20) << __func__ << "  lex 0x" << std::hex << offset << std::dec
@@ -6182,7 +6183,7 @@ void BlueStore::_do_write_big(
     blp.copy(l, t);
     _buffer_cache_write(txc, b, 0, t, wctx->buffered ? 0 : Buffer::FLAG_NOCACHE);
     wctx->write(b, l, 0, t, false);
-    bluestore_lextent_t lex(b->id, 0, l);
+    bluestore_lextent_t lex(b->id, 0, l, wctx->le_depth);
     o->onode.set_lextent(offset, lex, &b->dirty_blob(), &wctx->lex_old);
     txc->statfs_delta.stored() += l;
     dout(20) << __func__ << "  lex 0x" << std::hex << offset << std::dec << ": "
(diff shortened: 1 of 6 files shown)


ifed01:
@Roushan81 You may want to perform commit history cleanup via git rebase -i

src/os/bluestore/bluestore_types.cc
```diff
@@ -752,12 +752,14 @@ void bluestore_lextent_t::encode(bufferlist& bl) const
   small_encode_signed_varint(blob, bl);
   small_encode_varint_lowz(offset, bl);
   small_encode_varint_lowz(length, bl);
+  small_encode_varint_lowz(blob_depth, bl);
```
ifed01: no need to use _lowz encoding here as it's intended for values that tend to have low bits set to zero. Suggest to use small_encode_varint or even replace blob_depth type with uint8_t (if level 256 is enough in any case) that requires no small encoding at all.

src/osd/osd_types.h
```diff
@@ -4370,6 +4370,7 @@ struct store_statfs_t
   int64_t compressed = 0;              // Bytes stored after compression
   int64_t compressed_allocated = 0;    // Bytes allocated for compressed data
   int64_t compressed_original = 0;     // Bytes that were successfully compressed
+  int64_t garbage_collected = 0;       // Bytes written to recover space trapped because of overlapped blobs.
```
ifed01: Don't see any use of this field. Missed something?

somnathr:
@ifed01 @markhpc Could you please label it as Bluestore ?

src/os/bluestore/BlueStore.cc
```diff
@@ -6147,6 +6150,142 @@ void BlueStore::_wctx_finish(
...
+  {
+    map<uint64_t,bluestore_lextent_t>::iterator bp =
+	  o->onode.extent_map.begin();
+    while(bp != o->onode.extent_map.end()) {
+        derr << " Extent Start = " << bp->first << " Extent Length = " << bp->second.length
+             << " Depth = " << bp->second.blob_depth << dendl;
+        ++bp;
+    }
+  }
+#endif
```
somnathr: @varadakari Why within #ifdef ? Also why derr ? 
Need to do it with say dout (40)

src/os/bluestore/BlueStore.cc
```diff
@@ -6250,6 +6384,18 @@ int BlueStore::_do_write(
...
+
```
somnathr: Need to do it with dout()

src/os/bluestore/BlueStore.cc
```diff
@@ -6147,6 +6150,131 @@ void BlueStore::_wctx_finish(
...
+  uint64_t *gc_end_offset)
+{
+  uint32_t depth = 0;
+  bool head_overlap = false;
+  bool tail_overlap = false;
+
+  *gc_start_offset = start_offset;
+  *gc_end_offset   = end_offset;
+
+  map<uint64_t,bluestore_lextent_t>::iterator hp =
```
ifed01: please simplify by using bluestore_onode_t::seek_lextent instead

somnathr:
@varadakari @Roushan81 I will go through this more in depth , but, with initial looking it seems this patch will have an impact on write path. I would say please submit a detailed read/write performance report (by creating a cluster) with and without your patch.

src/os/bluestore/BlueStore.cc
```diff
@@ -6147,6 +6150,131 @@ void BlueStore::_wctx_finish(
...
+    if (hp->first + hp->second.length <= start_offset) {
+      ++hp;
+    }
+  }
+  if (hp->first < start_offset && start_offset < hp->first + hp->second.length) {
+    depth = hp->second.blob_depth;
+    head_overlap = true;
+  } 
+
+  map<uint64_t,bluestore_lextent_t>::iterator tp =
```
ifed01: same here - one can use seek_lextent instead

src/os/bluestore/BlueStore.cc
```diff
@@ -6209,38 +6337,33 @@ int BlueStore::_do_write(
...
-    middle_length = length - head_length - tail_length;
-
-    if (head_length) {
-      _do_write_small(txc, c, o, head_offset, head_length, p, &wctx);
+    if (offset > gc_start_offset) {
+      bufferlist head_bl;
+      size_t read_len = offset - gc_start_offset;
+      int r = _do_read(c.get(), o, gc_start_offset, read_len, head_bl, 0);
+      assert(r >= 0 && r <= (int)read_len); 
+      _do_write_data(txc, c, o, gc_start_offset, read_len, head_bl, &wctx);
```
ifed01: By doing separate writes for head, tail and body you considerably reduce the benefit that collector is  expected to bring. E.g. one can have following writes:
1) 0~128K, 128K~128K
2) 64K~128K
3) 64K~64K
Resulting lextents' map will be (IMHO):
0~64, 64~64, 128~128
Not a grate achievement for a costly GC operation.

src/os/bluestore/BlueStore.cc
```diff
@@ -6209,38 +6337,33 @@ int BlueStore::_do_write(
...
 
-    middle_offset = head_offset + head_length;
-    middle_length = length - head_length - tail_length;
-
-    if (head_length) {
-      _do_write_small(txc, c, o, head_offset, head_length, p, &wctx);
+    if (offset > gc_start_offset) {
+      bufferlist head_bl;
+      size_t read_len = offset - gc_start_offset;
+      int r = _do_read(c.get(), o, gc_start_offset, read_len, head_bl, 0);
```
ifed01: Doing sync reads at write path may significantly affect write operation performance. Perhaps better approach would be to write a new blob, log the need for GC and then do GC in background.

src/os/bluestore/BlueStore.cc
```diff
@@ -6147,6 +6150,131 @@ void BlueStore::_wctx_finish(
...
+        if (tp == o->onode.extent_map.end() ||
+            (tp_prev->first + tp_prev->second.length) != tp->first) {
+          tp = tp_prev;
+          break;
+        }
+      }
+      *gc_end_offset = tp->first + tp_prev->second.length;
+    }
+  }
+  if (depth >= 3) {
```
ifed01: This value(3) should definitely be a config setting.

src/osd/osd_types.h
ifed01: Currently we measure amount of data rewritten during GC. Perhaps more useful parameter is space saved by GC procedure, i.e. the difference between previous amount of wasted bytes (allocated - useful) and the resulting one after GC.

src/test/objectstore/store_test.cc
```diff
@@ -903,6 +903,205 @@ TEST_P(StoreTest, CompressionTest) {
   g_ceph_context->_conf->apply_changes(NULL);
 }
 
+TEST_P(StoreTest, garbageCollection) {
```
ifed01: This UT verifies almost nothing - there is neither checks that GC has been executed and produced any results, nor data consistency after its' application. It will successfully pass without any GC functionality as well hence IMO it has  little sense.

src/os/bluestore/BlueStore.cc
Roushan81: Done the changes. (After a series of changes, the final code ended up  being the duplicate of seek_lextent).

src/os/bluestore/BlueStore.cc
Roushan81: Done.

src/os/bluestore/BlueStore.cc
Roushan81: Yes, even my personal preference is to combine them together to reduce the number of smaller/unaligned lextents etc.  But one of Allen's suggestion was to retain/honor native size of current write request and NOT to merge it with the garbage collected data. 

So, I would like to request Allen to comment here please.

src/test/objectstore/store_test.cc
Roushan81: The tricky thing I found was: even reading the data and verifying them also does not say anything about  GC run and its correctness .  
 To do testing:  I used to debug print the "blob_depth" of the lextents before and after doing GC and used to manually verify them.
 
Let me try to use one of your other suggestion regarding adding a new stats counter  ".... space saved by GC procedure" to do the UT.

src/os/bluestore/BlueStore.cc
Roushan81: Agreed that it will degrade the front-end performance. Shall we take that as another enhancement?

src/osd/osd_types.h
Roushan81: I'll try to add that counter.

src/test/objectstore/store_test.cc
ifed01: You may use statistics returned by statfs method to measure amount of wasted data and verify it's decreasing (waste =  statfs.allocated - statfs.stored).

markhpc:
Marking this as DNM for now pending discussion and eventual performance results.

src/osd/osd_types.h
ifed01: Perhaps it's preferable to have such counter(s) implemented via perf counters. statfs structure is rather for generic storage space utilization statistics collection only. IMO we shouldn't overburden it. The counter we are currently discussing is an auxiliary one and hence doesn't deserve to be that common.

src/os/bluestore/BlueStore.cc
ifed01: Maybe have different(configurable) policies: aggressive, passive, etc?

src/os/bluestore/BlueStore.cc
Roushan81: Done.

liewegas:
This approach looks good to me.  I can imagine a few minor refinements, but we can do those later if it seems worth it.  For now,

1. please rebase on current master
2. squash this down, either into one commit, or two (one that switches to the _do_write_data helper, and a second that adds the gc logic).

Thanks!

Roushan81:
Closing this PR, #11232 is addressing the same.

PR #10409, opened 2016-07-23
rgw:fix rgw boot failed after upgrade to master latest version

weiqiaomiao:

### DIFF
--- a/src/rgw/rgw_rados.h
+++ b/src/rgw/rgw_rados.h
@@ -895,11 +895,10 @@ struct RGWZoneParams : RGWSystemMetaObj {
   int fix_pool_names();
   
   void encode(bufferlist& bl) const {
-    ENCODE_START(6, 1, bl);
+    ENCODE_START(7, 1, bl);
     ::encode(domain_root, bl);
     ::encode(control_pool, bl);
     ::encode(gc_pool, bl);
-    ::encode(lc_pool, bl);
     ::encode(log_pool, bl);
     ::encode(intent_log_pool, bl);
     ::encode(usage_log_pool, bl);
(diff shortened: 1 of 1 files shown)


weiqiaomiao:
retest this please

src/rgw/rgw_rados.cc
```diff
@@ -4000,7 +4000,14 @@ int RGWRados::open_gc_pool_ctx()
 
 int RGWRados::open_lc_pool_ctx()
 {
-  const char *lc_pool = get_zone_params().lc_pool.name.c_str();
+  const char *lc_pool;
+
+  if (!get_zone_params().lc_pool.name.empty()){
+    lc_pool = get_zone_params().lc_pool.name.c_str();
+  } else {
+    lc_pool = "default.rgw.lc";
```
cbodley: i'm worried about the "default" part here. in a multisite configuration, we can't have multiple zones using the same "default.rgw.lc" pool. we avoid this by having each zone's pool names start with the zone name instead of "default"

so `lc_pool` should default to `get_zone_params().get_name() + ".rgw.lc"`. and i think we're better off setting this default in `RGWZoneParams::decode()` when we find that `struct_v < 7`

oritwas: this is still problematic, we can have a name collision (after renaming a zone) you will need to use fix_zone_pool_name.
We also need to add the lc pool to new zones in RGWZoneParams::fix_pool_names.

weiqiaomiao: ``lc pool``  had already included in current RGWZoneParams::fix_pool_names

weiqiaomiao:
@cbodley @oritwas  thanks for your review and the commit is updated .

yehudasa:
@oritwas @cbodley looks fine

PR #10410, opened 2016-07-23
os/bluestore: narrow lock scope for cache trim()

xiexingguo:
The cache trim() process can become time consuming
when the cache size is huge. Since it can be operated
safely under its own internal mutex, we shall avoid
to hold irrelevant locks when possible, which is good
for performance.

### DIFF
--- a/src/os/bluestore/BlueStore.cc
+++ b/src/os/bluestore/BlueStore.cc
@@ -3373,14 +3373,21 @@ bool BlueStore::exists(CollectionHandle &c_, const ghobject_t& oid)
   dout(10) << __func__ << " " << c->cid << " " << oid << dendl;
   if (!c->exists)
     return false;
-  RWLock::RLocker l(c->lock);
-  OnodeRef o = c->get_onode(oid, false);
+
+  bool r = true;
+
+  {
+    RWLock::RLocker l(c->lock);
+    OnodeRef o = c->get_onode(oid, false);
+    if (!o || !o->exists)
+      r = false;
+  }
+
   c->cache->trim(
     g_conf->bluestore_onode_cache_size,
     g_conf->bluestore_buffer_cache_size);
-  if (!o || !o->exists)
-    return false;
-  return true;
+
+  return r;
 }
 
 int BlueStore::stat(
@@ -3405,14 +3412,18 @@ int BlueStore::stat(
   if (!c->exists)
     return -ENOENT;
   dout(10) << __func__ << " " << c->get_cid() << " " << oid << dendl;
-  RWLock::RLocker l(c->lock);
-  OnodeRef o = c->get_onode(oid, false);
-  if (!o || !o->exists)
-    return -ENOENT;
-  st->st_size = o->onode.size;
-  st->st_blksize = 4096;
-  st->st_blocks = (st->st_size + st->st_blksize - 1) / st->st_blksize;
-  st->st_nlink = 1;
+
+  {
+    RWLock::RLocker l(c->lock);
+    OnodeRef o = c->get_onode(oid, false);
+    if (!o || !o->exists)
+      return -ENOENT;
+    st->st_size = o->onode.size;
+    st->st_blksize = 4096;
+    st->st_blocks = (st->st_size + st->st_blksize - 1) / st->st_blksize;
+    st->st_nlink = 1;
+  }
+
   c->cache->trim(
     g_conf->bluestore_onode_cache_size,
     g_conf->bluestore_buffer_cache_size);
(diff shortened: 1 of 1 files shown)


xiexingguo:
2016-07-23 12:51:26.319710 7ff4dbc42680  1 bluestore(store_test_temp_dir) fsck finish with 0 errors
[       OK ] ObjectStore/StoreTest.TooManyBlobsTest/2 (22819 ms)
[----------] 59 tests from ObjectStore/StoreTest (2114529 ms total)

[==========] 59 tests from 1 test case ran. (2114531 ms total)
[  PASSED  ] 59 tests.

src/os/bluestore/BlueStore.cc
```diff
@@ -3373,14 +3373,21 @@ bool BlueStore::exists(CollectionHandle &c_, const ghobject_t& oid)
...
+
+  bool r = true;
+
+  {
+    RWLock::RLocker l(c->lock);
+    OnodeRef o = c->get_onode(oid, false);
+    if (!o || !o->exists)
+      r = false;
+  }
+
```
somnathr: We can't do that cache->trim needs to be within collection lock as it is modifying onode structures..

src/os/bluestore/BlueStore.cc
```diff
@@ -3405,14 +3412,18 @@ int BlueStore::stat(
...
+    OnodeRef o = c->get_onode(oid, false);
+    if (!o || !o->exists)
+      return -ENOENT;
+    st->st_size = o->onode.size;
+    st->st_blksize = 4096;
+    st->st_blocks = (st->st_size + st->st_blksize - 1) / st->st_blksize;
+    st->st_nlink = 1;
+  }
+
   c->cache->trim(
```
somnathr: @xiexingguo same as above.

src/os/bluestore/BlueStore.cc
```diff
@@ -3450,22 +3461,23 @@ int BlueStore::read(
...
+    }
 
-  if (offset == length && offset == 0)
-    length = o->onode.size;
+    if (offset == length && offset == 0)
+      length = o->onode.size;
 
-  r = _do_read(c, o, offset, length, bl, op_flags);
+    r = _do_read(c, o, offset, length, bl, op_flags);
+  }
```
somnathr: @xiexingguo same as above.

src/os/bluestore/BlueStore.cc
```diff
@@ -3763,57 +3775,59 @@ int BlueStore::fiemap(
...
-    }
-    if (ep != eend &&
-	ep->first > offset &&
-	ep->first - offset < x_len) {
-      x_len = ep->first - offset;
+      length -= x_len;
     }
-    offset += x_len;
-    length -= x_len;
   }
```
somnathr: @xiexingguo Same

src/os/bluestore/BlueStore.cc
```diff
@@ -739,7 +739,9 @@ void BlueStore::TwoQCache::trim(uint64_t onode_max, uint64_t buffer_max)
       Buffer *b = &*p;
       assert(b->is_clean());
       dout(20) << __func__ << " buffer_warm_in -> out " << *b << dendl;
+      assert(buffer_bytes >= b->length);
```
somnathr: @xiexingguo assert()s are cpu extensive , I don't think we need to do it within loop. Did you face this issue hitting ?
buffer_bytes is already > buffer_cache_size and it is unlikely that individual buffer length we are trimming > buffer_bytes.

src/os/bluestore/BlueStore.cc
```diff
@@ -739,7 +739,9 @@ void BlueStore::TwoQCache::trim(uint64_t onode_max, uint64_t buffer_max)
...
       buffer_bytes -= b->length;
+      assert(buffer_list_bytes[BUFFER_WARM_IN] >= b->length);
       buffer_list_bytes[BUFFER_WARM_IN] -= b->length;
```
somnathr: @xiexingguo please create a new pull request for this as this commit is logically different that the other.
I am not very sure about this is needed. Let's other comment.

src/os/bluestore/BlueStore.cc
xiexingguo: Negative, `cache->trim()` does cache cleanup only and is under its own mutex protection already.



> We can't do that cache->trim needs to be within collection lock as it is modifying onode structures..

Otherwise shouldn't we promote to `WLocker(c->lock)` here?

somnathr: trim() does clear blob_map and onode_map. protecting against cache mutex is not good enough.
All the onode operations should be protected by collection lock. In this case, you are right, it should be promoted to Wlock at the time of clearing the onode structures..
BTW, I don't think taking write lock in the beginning instead of Rlock will be impacting performance much as the requests are been serialized within a PG (collection).

xiexingguo:
retest this please

src/os/bluestore/BlueStore.cc
liewegas: The cache->lock protects the BufferSpace and OnodeSpace structures that get trimmed.

liewegas:
lgtm

somnathr:
@liewegas we saw a crash that Mark reported because the cache->trim() is not happening within c->lock() from _osr_reap_done(). The  pull request #10578 seems to fix that.
onodes are accessed within c->lock() and without cache->lock() (for ex, onode->flush() ) and if from trim() it  decrements reference without c->lock() , it may delete onode object while there still be user.

Am I missing anything ?

liewegas:
trim should operate with only cache->lock, and any place where it affects a structure needs to be protected with cache->lock in the forward path (not the collection lock).

PR #10412, opened 2016-07-23
mon/Monitor.h: add const to member function

stiopaa1:

### DIFF
--- a/src/mon/Monitor.h
+++ b/src/mon/Monitor.h
@@ -557,8 +557,8 @@ private:
 
 public:
   epoch_t get_epoch();
-  int get_leader() { return leader; }
-  const set<int>& get_quorum() { return quorum; }
+  int get_leader() const { return leader; }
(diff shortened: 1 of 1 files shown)


src/mon/Monitor.h
```diff
@@ -586,7 +586,7 @@ class Monitor : public Dispatcher,
 
 public:
   epoch_t get_epoch();
-  int get_leader() { return leader; }
+  int get_leader() const { return leader; }
   const set<int>& get_quorum() { return quorum; }
```
tchaikov: @stiopaa1 could you do this sort of cleanup in a batch. for example, add const in to the should-be-const methods of a class in a commit? to avoid too many merge commit? for example, in this commit we can also mark the `get_quorum()` method as `const`.

tchaikov:
@stiopaa1 ping?

stiopaa1:
@tchaikov 
I've modified the pull request

PR #10415, opened 2016-07-23
ceph-create-keys: fix existing-but-different case

jcsp:
We just have to refrain from calling get-or-create
if the named key already exists, to avoid potentially
having an error when the default creation args don't
match the key as it has already been created, such
as on certain upgrades.

Fixes: http://tracker.ceph.com/issues/16255

### DIFF
--- a/src/ceph-create-keys
+++ b/src/ceph-create-keys
@@ -92,24 +92,41 @@ def get_key(cluster, mon_id):
                 os.fchmod(f.fileno(), 0600)
                 os.fchown(f.fileno(), get_ceph_uid(), get_ceph_gid())
                 LOG.info('Talking to monitor...')
-                returncode = subprocess.call(
-                    args=[
-                        'ceph',
+
+                args_prefix = [
+                        "ceph",
                         '--cluster={cluster}'.format(cluster=cluster),
                         '--name=mon.',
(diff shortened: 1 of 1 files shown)


alfredodeza:
This looks OK to me

src/ceph-create-keys
```diff
@@ -92,24 +92,41 @@ def get_key(cluster, mon_id):
                 os.fchmod(f.fileno(), 0600)
                 os.fchown(f.fileno(), get_ceph_uid(), get_ceph_gid())
                 LOG.info('Talking to monitor...')
-                returncode = subprocess.call(
-                    args=[
-                        'ceph',
+
+                args_prefix = [
+                        "ceph",
```
tchaikov: nit, s/"/'/ for consistency.

src/ceph-create-keys
```diff
@@ -92,24 +92,41 @@ def get_key(cluster, mon_id):
...
                             cluster=cluster,
                             mon_id=mon_id,
                             ),
+                        ]
+
+                # First try getting the key if it already exists, to handle
+                # the case where it exists but doesn't match the caps
+                # we would pass into get-or-create.
+                returncode = subprocess.call(
+                    args=args_prefix + [
```
tchaikov: pep8, add space around `=`.

tchaikov:
lgtm.

PR #10417, opened 2016-07-24
changes for Clang and yasm

wjwithagen:
- Clang does not need all flags and options
- uname -m on FreeBSD returns amd64

PR #10418, opened 2016-07-24
Pass compiler and flags to compile RocksDB into build

wjwithagen:
- make sure the right compiler is used
 - with the correct flags

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -637,9 +637,15 @@ install(TARGETS ceph-mon DESTINATION bin)
 if(NOT ALLOCATOR STREQUAL "jemalloc")
   set(disable_jemalloc "DISABLE_JEMALLOC=1")
 endif()
+
+set(ROCKSDB_EXTRA_CXXFLAG -fPIC -Wno-unused-variable)
+if(NOT CMAKE_BUILD_TYPE STREQUAL Debug)
+  set(ROCKSDB_EXTRA_CXXFLAG ${ROCKSDB_EXTRA_CXXFLAG} -DNDEBUG)
+endif(NOT CMAKE_BUILD_TYPE STREQUAL Debug)
+
 add_custom_target(build_rocksdb
     COMMAND
-    PORTABLE=1 ${disable_jemalloc} $(MAKE) static_lib EXTRA_CXXFLAGS='-fPIC -Wno-unused-variable'
+    PORTABLE=1 ${disable_jemalloc} $(MAKE) static_lib CXX="${CMAKE_CXX_COMPILER}" EXTRA_CXXFLAGS="${ROCKSDB_EXTRA_CXXFLAG}"
     WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/src/rocksdb
     COMMENT "rocksdb building")
 


src/CMakeLists.txt
```diff
@@ -637,9 +637,14 @@ install(TARGETS ceph-mon DESTINATION bin)
 if(NOT ALLOCATOR STREQUAL "jemalloc")
   set(disable_jemalloc "DISABLE_JEMALLOC=1")
 endif()
+set(EXTRA_CXXFLAGS -fPIC -Wno-unused-variable -DNDEBUG)
+if(FREEBSD)
+  set(EXTRA_CXXFLAGS ${EXTRA_CXXFLAGS} -Irocksdb/include -I/usr/local/include)
+  set(MAKEFLAGS ${MAKEFLAGS})
```
tchaikov: what is this for?

src/CMakeLists.txt
```diff
@@ -637,9 +637,14 @@ install(TARGETS ceph-mon DESTINATION bin)
 if(NOT ALLOCATOR STREQUAL "jemalloc")
   set(disable_jemalloc "DISABLE_JEMALLOC=1")
 endif()
+set(EXTRA_CXXFLAGS -fPIC -Wno-unused-variable -DNDEBUG)
+if(FREEBSD)
+  set(EXTRA_CXXFLAGS ${EXTRA_CXXFLAGS} -Irocksdb/include -I/usr/local/include)
```
tchaikov: why rocksdb needs `-Irocksdb/include`? it has 
```
CFLAGS += $(WARNING_FLAGS) -I. -I./include $(PLATFORM_CCFLAGS) $(OPT)
CXXFLAGS += $(WARNING_FLAGS) -I. -I./include $(PLATFORM_CXXFLAGS) $(OPT) -Woverloaded-virtual -Wnon-virtual-dtor -Wno-missing-field-initializers
```
in its `Makefile`. so i assume it's not necessary?

src/CMakeLists.txt
```diff
@@ -637,9 +637,14 @@ install(TARGETS ceph-mon DESTINATION bin)
 if(NOT ALLOCATOR STREQUAL "jemalloc")
   set(disable_jemalloc "DISABLE_JEMALLOC=1")
 endif()
+set(EXTRA_CXXFLAGS -fPIC -Wno-unused-variable -DNDEBUG)
```
tchaikov: could you add a prefix for `EXTRA_CXXFLAGS`, so it is `ROCKSDB_EXTRA_CXXFLAGS`?

so this name has less chance to collide with other (future) names.

src/CMakeLists.txt
It is possible to run the buidl with `make V=1 VERBOSE=1` to get e more verbose output of what is going on.
And the build in RocksDB supports that as well.

src/CMakeLists.txt
I need to run it again without this  to see what actually went wrong, but only when I added that it did build correctly.

src/CMakeLists.txt
tchaikov: i mean, you set `MAKEFLAGS` to `MAKEFLAGS`. this is a no-op.

Right.... Turns out that that whole block is bogus. It stems back from the time that I also had rocksdb installed as package. And then includefiles started to be mixed. And linking was a mess as well.
Looks like things have improved on all ends.
So the only things really required to be set are NDEBUG and CXX.
I'll fix.

src/CMakeLists.txt
tchaikov: why `-NDEBUG`? and could you just set the FLAGS inline without introducing this variable if it is not used anywhere else?

tchaikov: > `rocksdb really likes NDEBUG`

could you be more specific?

src/CMakeLists.txt
```diff
@@ -637,9 +637,23 @@ install(TARGETS ceph-mon DESTINATION bin)
 if(NOT ALLOCATOR STREQUAL "jemalloc")
   set(disable_jemalloc "DISABLE_JEMALLOC=1")
 endif()
+
+set(ROCKSDB_EXTRA_CXXFLAG -fPIC -Wno-unused-variable)
+if(NOT CMAKE_BUILD_TYPE MATCHES "Debug")
```
tchaikov: nit, could be more consistent with existing way to check `CMAKE_BUILD_TYPE` at line:30
```cmake
if(NOT CMAKE_BUILD_TYPE STREQUAL Debug)
```

src/CMakeLists.txt
```diff
@@ -637,9 +637,23 @@ install(TARGETS ceph-mon DESTINATION bin)
...
+  set(ROCKSDB_EXTRA_CXXFLAG ${ROCKSDB_EXTRA_CXXFLAG} -DNDEBUG)
+endif(NOT CMAKE_BUILD_TYPE MATCHES "Debug")
+
+# get number of CPUs to build rockdb with
+if(LINUX)
```
tchaikov: i am not sure about this change, as 

1. we plan to use cmake to build rocksdb once https://github.com/facebook/rocksdb/pull/1135 is merged.
1. rocksdb is not changed very frequently in comparison to other components in ceph.
1. the heuristics of "nprocs/2" might not work well if we are already `make -j $(($(nproc)/2))`

tchaikov: i am not sure about this change, as 

1. we plan to use cmake to build rocksdb once https://github.com/facebook/rocksdb/pull/1135 is merged.
1. rocksdb is not changed very frequently in comparison to other components in ceph.
1. the heuristics of "nprocs/2" might not work well if we are already `make -j $(($(nproc)/2))`

tchaikov:
imo, the changes is not specific to FreeBSD or clang, could you revise your commit message and its title?

src/CMakeLists.txt
"Fun" part is that I've already have line 30 with "matches" as well, in another WIP.
That got deleted when we removed the debug options from the Clang commit.
So I actually copied it from there. :)
But I'll make it consistent.

src/CMakeLists.txt
I looked at Cmake for RocksDB and it is pretty Windows targetted, so some stiff work ahead.
What do you suggest for the NPROC for RocksDB? This is how it is in Automake.

src/CMakeLists.txt
```diff
@@ -637,9 +637,15 @@ install(TARGETS ceph-mon DESTINATION bin)
 if(NOT ALLOCATOR STREQUAL "jemalloc")
   set(disable_jemalloc "DISABLE_JEMALLOC=1")
 endif()
+
+set(ROCKSDB_EXTRA_CXXFLAG -fPIC -Wno-unused-variable)
+if(NOT CMAKE_BUILD_TYPE STREQUAL Debug)
+  set(ROCKSDB_EXTRA_CXXFLAG ${ROCKSDB_EXTRA_CXXFLAG} -DNDEBUG)
+endif(NOT CMAKE_BUILD_TYPE MATCHES "Debug")
```
tchaikov: does not match with `if`.

wjwithagen: On 27-7-2016 10:27, Kefu Chai wrote:
> In src/CMakeLists.txt
> <https://github.com/ceph/ceph/pull/10418#discussion_r72399688>:
> 
>> @@ -637,9 +637,15 @@ install(TARGETS ceph-mon DESTINATION bin)
>>  if(NOT ALLOCATOR STREQUAL "jemalloc")
>>    set(disable_jemalloc "DISABLE_JEMALLOC=1")
>>  endif()
>> +
>> +set(ROCKSDB_EXTRA_CXXFLAG -fPIC -Wno-unused-variable)
>> +if(NOT CMAKE_BUILD_TYPE STREQUAL Debug)
>> +  set(ROCKSDB_EXTRA_CXXFLAG ${ROCKSDB_EXTRA_CXXFLAG} -DNDEBUG)
>> +endif(NOT CMAKE_BUILD_TYPE MATCHES "Debug")
> 
> does not match with |if|.

Aaarrggh... Sorry about that.
I've got this root-canal infection, with splitting headache, for about 2
weeks now. And it is definitly impacting my concentration.

Will fix.

src/CMakeLists.txt
tchaikov: > I looked at Cmake for RocksDB and it is pretty Windows targetted, so some stiff work ahead.

right, the PR is not yet merged.

> What do you suggest for the NPROC for RocksDB? This is how it is in Automake.

i'd suggest not to touch this part. it will be resolved once rocksdb's cmake support is expanded to other platforms.

tchaikov:
lgtm once the jenkins returns.

wjwithagen:
@tchaikov 
Thanx.
Any chance you can have a look at #10417??

src/CMakeLists.txt
tchaikov: no worries. take care.

tchaikov:
@wjwithagen i will check #10417 again once you address all the comments.

wjwithagen:
@tchaikov 
Thanx, GHoing to work on the next now.

PR #10423, opened 2016-07-25
osd: fix the mem leak of RepGather

tchaikov:
ReplicatedPG::new_repop() returns a pointer to RepGather with two refcounts,
one is held by ReplicatedPG::repop_queue, the other is supposed to be
held by the caller of this function. but it's caller
ReplicatedPG::submit_log_entries() assigns it to a
boost::intrusive_ptr<RepGather>() directly, why by default add_ref() in
its constructor. this makes the refcount 3. that's why we have a leak of
RepGather in `ReplicatedPG::new_repop(ObcLockManager&&,
boost::optional<std::function<void ()>>&&)`.

Fixes: http://tracker.ceph.com/issues/16801

### DIFF
--- a/src/osd/ReplicatedPG.cc
+++ b/src/osd/ReplicatedPG.cc
@@ -8619,7 +8619,7 @@ ReplicatedPG::RepGather *ReplicatedPG::new_repop(
   return repop;
 }
 
-ReplicatedPG::RepGather *ReplicatedPG::new_repop(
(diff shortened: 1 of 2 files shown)


tchaikov:
@athanatos could you please take a look?

athanatos:
@tchaikov  Ah, nice catch!  This should be backported to jewel, right?  (lgtm!)

tchaikov:
> This should be backported to jewel, right?

yeah!

PR #10425, opened 2016-07-25
rgw: kill a compile warning for rgw_sync

theanalyst:
killing the compile warning for

```
/ceph/src/rgw/rgw_sync.cc:1462:12:
warning: suggest explicit braces to avoid ambiguous ‘else’ [-Wparentheses]
         if (can_adjust_marker) yield {
            ^
```

### DIFF
--- a/src/rgw/rgw_sync.cc
+++ b/src/rgw/rgw_sync.cc
@@ -1459,18 +1459,18 @@ public:
 
       if (!lost_lock) {
         /* update marker to reflect we're done with full sync */
-        if (can_adjust_marker) yield {
-          sync_marker.state = rgw_meta_sync_marker::IncrementalSync;
-          sync_marker.marker = sync_marker.next_step_marker;
-          sync_marker.next_step_marker.clear();
+        if (can_adjust_marker) {
+	  sync_marker.state = rgw_meta_sync_marker::IncrementalSync;
+	  sync_marker.marker = sync_marker.next_step_marker;
+	  sync_marker.next_step_marker.clear();
+	  ldout(sync_env->cct, 0) << *this << ": saving marker pos=" << sync_marker.marker << dendl;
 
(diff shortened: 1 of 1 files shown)


src/rgw/rgw_sync.cc
```diff
@@ -1459,17 +1459,19 @@ class RGWMetaSyncShardCR : public RGWCoroutine {
...
+            sync_marker.marker = sync_marker.next_step_marker;
+            sync_marker.next_step_marker.clear();
+
+            RGWRados *store = sync_env->store;
+            ldout(sync_env->cct, 0) << *this << ": saving marker pos=" << sync_marker.marker << dendl;
+            using WriteMarkerCR = RGWSimpleRadosWriteCR<rgw_meta_sync_marker>;
+            call(new WriteMarkerCR(sync_env->async_rados, store, pool,
+                                   sync_env->shard_obj_name(shard_id),
+                                   sync_marker));
+          }
```
cbodley: it looks like the local variable `RGWRados *store` is the only reason this yield {} block is needed. if you take that out, we can just do this:
```c++
            yield call(new WriteMarkerCR(sync_env->async_rados, sync_env->store, pool,
                                         sync_env->shard_obj_name(shard_id),
                                         sync_marker));
```

theanalyst: what about the sync_marker clearing, can we do this before the yield call ?

cbodley: yeah, `sync_marker` is a member variable. `yield` is implemented as a switch statement, so you only have to worry about local variable declarations crossing the yield points

theanalyst:
@cbodley updated, and ran through a run of test_multi

cbodley:
@theanalyst thanks! will merge once jenkins approves

PR #10426, opened 2016-07-25
scrub: make requested scrubs jump the queue

gaowanlong:
Make requested scrubs jump the queue by giving them a
higher priority.

Fixes: http://tracker.ceph.com/issues/15789
Signed-off-by: Wanlong Gao wanlong.gao@easystack.cn


gaowanlong:
@athanatos

athanatos:
@gaowanlong They already effectively jump the queue, only scrubs which have gotten to their deadline (or were requested) make it into the queue at all.  Why +1?   Priority is a pretty big range, +1 would likely not matter much.  I don't want to complicate the code base with this change without evidence that it makes a difference to someone's workload.

athanatos:
...I just noticed that I created that bug!  Still, I think the amount that scrub increases the priority by should be configurable (you can just add a config to config_opts.h).

gaowanlong:
@athanatos Yeah, I'll add a config.  Currently, the requested scrubs can jump the sched_scrub_pg queue , but  can not jump the scheduled scrub queue, then a higher priority can make it possible. ;)

gaowanlong:
@athanatos Any more comments ?

src/common/config_opts.h
```diff
@@ -874,6 +874,9 @@ OPTION(osd_snap_trim_priority, OPT_U32, 5)
 OPTION(osd_snap_trim_cost, OPT_U32, 1<<20) // set default cost equal to 1MB io
 
 OPTION(osd_scrub_priority, OPT_U32, 5)
+// set default must scrub priority higher than scrub priority to make the
+// requested scrubs jump the queue of scheduled scrubs
+OPTION(osd_must_scrub_priority, OPT_U32, 6)
```
athanatos: osd_requested_scrub_priority

gaowanlong: @athanatos thank you for your comment, updated to osd_requested_scrub_priority now. ;)

athanatos:
You need to add a test that exercises the new option.

liewegas:
also needs rebase

src/osd/PG.cc
```diff
@@ -2080,6 +2079,7 @@ bool PG::queue_scrub()
     scrubber.must_repair = false;
   }
   requeue_scrub();
+  scrubber.must_scrub = false;
```
dzafman: Why is setting must_scrub moved down?

tchaikov:
@gaowanlong ping?

dachary:
needs a rebase

src/osd/PG.cc
dzafman: Let's not move setting must_scrub to false unless it is needed.

dzafman:
@gaowanlong Still needs rebase.  The config for the request priority should be in common/config_opts.h.  It doesn't need to be a pool parameter.  Recently, someone was complaining about a requested (must) scrub not starting for while.  So we may have seen this issue.

dzafman:
Replaced by ceph/ceph#14488

PR #10311, opened 2016-07-15
cmake: fix the build on AArch64

tone-zhang:
When compile the Ceph, because some compiling options are
missed, the compiler stops at src/erasure-code.
Update the related CMakelists.txt files, and make the
Ceph pass the compiling.

When compile the Ceph UT, because the -msse and -msse2
are hardcoded in the test/CMakeLists.txt, the Ceph UT build
fail in AArm64. Update the source code to fix the issue.

In Ceph erasure UT, because the usful libary is missed, the
link is broken. Update test/erasure-code/CMakeLists.txt to
fix the link error.

Signed-off-by: tone.zhang <tone.zhang@linaro.org>

PR #7515, opened 2016-02-03
Wip wjw freebsd compile

wjwithagen:
Enough changes to get ceph to compile on FreeBSD. 

On FreeBSD the code now halts after testing runc-cli-tests
The next step will be to get the tests running....

PR #10427, opened 2016-07-25
cmake: fix the build on AArch64

tone-zhang:
The PR comes from https://github.com/ceph/ceph/pull/10311, because the origin one is in a mess.

Reconstruct the cmake files, extrace one common module from src/CMakeLists.txt.
And update the related CMakeLists.txt files.

When compile the Ceph UT, because the -msse and -msse2 are hardcoded in 
the test/CMakeLists.txt, the Ceph UT build fail in AArm64. Fix the issue.

Signed-off-by: tone.zhang tone.zhang@linaro.org

### DIFF
--- a/doc/dev/index.rst
+++ b/doc/dev/index.rst
@@ -179,6 +179,20 @@ with::
 
 For more ``vstart.sh`` examples, see :doc:`/dev/quick_guide`.
 
+Backporting
+-----------
+
+All bugfixes should be merged to the ``master`` branch before being backported.
+To flag a bugfix for backporting, make sure it has a `tracker issue`_
+associated with it and set the ``Backport`` field to a comma-separated list of
+previous releases (e.g. "hammer,jewel") that you think need the backport.
+The rest (including the actual backporting) will be taken care of by the
+`Stable Releases and Backports`_ team.
+
+.. _`tracker issue`: http://tracker.ceph.com/
+.. _`Stable Releases and Backports`: http://tracker.ceph.com/projects/ceph-releases/wiki
+
+
 What is merged where and when ?
 ===============================
 
@@ -570,6 +584,7 @@ hardware. Tests designed for this purpose live in the `ceph-qa-suite
 repository`_ and are run via the `teuthology framework`_.
 
 .. _`ceph-qa-suite repository`: https://github.com/ceph/ceph-qa-suite/
+.. _`teuthology repository`: https://github.com/ceph/teuthology
 .. _`teuthology framework`: https://github.com/ceph/teuthology
 
 If you have access to an OpenStack tenant, you are encouraged to run the
@@ -659,11 +674,8 @@ command to complete successfully on x86_64 (other architectures may have
 different constraints). Depending on your hardware, it can take from 20
 minutes to three hours to complete, but it's worth the wait.
 
-When you fix a bug, it's a good idea to add a test. See the `Writing make
-check tests`_ chapter.
-
-Further sections
-----------------
+Future sections
+---------------
 
 * Principles of make check tests
 * Where to find test results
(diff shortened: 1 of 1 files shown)


tchaikov:
@tone-zhang you might be interested in this change. https://github.com/ceph/ceph/pull/10438/commits/7818d91fafd63e2d3b79733e43ebbdf3cf9a8317

it's what we've been talking about.

tone-zhang:
@tchaikov Kefu, yes, it is the update in rocksdb which we need. Thanks.

tchaikov:
this change looks good to me, @alimaredia what do you think?

alimaredia:
Upon first glance this looks good to me. From what I understand one problem is we don't have any gitbuilders for it atm. @dmick @cbodley. Any thoughts?

src/erasure-code/jerasure/CMakeLists.txt
```diff
@@ -70,7 +74,7 @@ if(INTEL_SSE)
   add_dependencies(ec_jerasure_sse3 ${CMAKE_SOURCE_DIR}/src/ceph_ver.h)
   target_link_libraries(ec_jerasure_sse3 crush ${EXTRALIBS})
   set_target_properties(ec_jerasure_sse3 PROPERTIES VERSION 2.0.0 SOVERSION 2
-    COMPILE_FLAGS ${SSE3_FLAGS})
+    COMPILE_FLAGS ${COMPILE_TARGETS_FLAGS})
```
cbodley: this doesn't look right to me. we only want to pass supported sse3 flags for this target, not all supported flags

src/test/CMakeLists.txt
```diff
@@ -356,8 +356,9 @@ target_link_libraries(ceph_multi_stress_watch librados global radostest
 add_executable(ceph_perf_local 
   perf_local.cc
   perf_helper.cc)
-#INTEL_SSE & INTEL_SSE2 flags
-set(PERF_LOCAL_FLAGS "-msse -msse2")
+
+include(CheckCompileTargets)
+set(PERF_LOCAL_FLAGS ${COMPILE_TARGETS_FLAGS})
```
cbodley: just use `${SSE3_FLAGS}` here. they should be empty when unsupported

cbodley:
I like how you moved the detection stuff into a module, but we need to keep the flags separated instead of combining them all into `COMPILE_TARGETS_FLAGS`

src/erasure-code/jerasure/CMakeLists.txt
tone-zhang: @cbodley Casey, thanks, I will roll back the code and use sse3 flags as origin.

src/test/CMakeLists.txt
tone-zhang: @cbodley Only SSE3 and NEON flags will be checked here. The update will be ready soon.

tchaikov:
@tone-zhang #10438 is competing with your PR, and it got merged first. if you'd like to continue working on this PR, we can still have the first commit for refactoring on the compiler detection.

tone-zhang:
@tchaikov Kefu, I will continue on the first. Thanks!

wjwithagen:
@tone-zhang 
Just a personal note.

From my personal experience with Pulls this large is that that is not going to work.
I once had a pull that also had something like 70+ commits, even 30 file pulls did not work: #7515
So I just closed those. 

There is no ill will with the commiters, but the code is mostly changing so rapidly that with 72 files being modified in one pull there is an almost 100% chance that your pull has something to discuss about.
And as a consequence of that the whole of the pull grows old and outdated. And keeping it up to date with HEAD is going to be more an more cumbersome.
@tchaikov is very helpfull in getting me along, but in the end the size of the commit is not going to work. (It only works when you have commit right yourself, and really know what you are doing)

So I personally have moved to smaller commits, with less files changed in one run. Chances of getting such a pull commited are way much bigger. And in the end you will get more commited.
I even make commit of one or 2 lines, 1 2 files. The impact is whay much less. The change is easier to review, and as a consequence is easier approved.

Just my 2 cts.

tone-zhang:
@wjwithagen Willem, it is my fault. I just commit two files I updated, but contain so many non-related commits which are the delta between my working branch and ceph/master. I need more practice to familiar with git commands. Apologize for the mess.

PR #10428, opened 2016-07-25
mon: validate states transmitted in beacons

jcsp:
Since FSMap was added, the state of a daemon can lead
to an entirely invalid map, but we were letting daemons
send any state they wanted.

Especially, we must not allow standby daemons to set
any state other than STANDBY.

Fixes: http://tracker.ceph.com/issues/16592

### DIFF
--- a/src/mon/MDSMonitor.cc
+++ b/src/mon/MDSMonitor.cc
@@ -645,7 +645,23 @@ bool MDSMonitor::prepare_beacon(MonOpRequestRef op)
 			mon->monmap->fsid, m->get_global_id(),
 			m->get_name(), fsmap.get_epoch(), state, seq,
 			CEPH_FEATURES_SUPPORTED_DEFAULT));
+    } else if (info.state == MDSMap::STATE_STANDBY && state != info.state) {
+      // Standby daemons should never modify their own
+      // state.  Reject any attempts to do so.
+      derr << "standby " << gid << " attempted to change state to "
+           << ceph_mds_state_name(state) << ", rejecting" << dendl;
+      return true;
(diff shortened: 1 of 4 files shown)


gregsfortytwo:
Actual build error here:

>/home/jenkins-build/build/workspace/ceph-pull-requests/src/mon/MDSMonitor.cc: In member function ‘bool MDSMonitor::prepare_beacon(MonOpRequestRef)’:
/home/jenkins-build/build/workspace/ceph-pull-requests/src/mon/MDSMonitor.cc:660:16: error: expected primary-expression before ‘<<’ token
                << ceph_mds_state_name(info.state) << " -> "

jcsp:
Oops, updated.

gregsfortytwo:
Reviewed-by:

src/mon/MDSMonitor.cc
```diff
@@ -646,6 +646,23 @@ bool MDSMonitor::prepare_beacon(MonOpRequestRef op)
 			m->get_name(), fsmap.get_epoch(), state, seq,
 			CEPH_FEATURES_SUPPORTED_DEFAULT));
     } else {
+      if (info.state == MDSMap::STATE_STANDBY && state != info.state) {
```
batrick: These `if`s should probably be an `else if` to be consistent I think?

batrick:
`Reviewed-by: Patrick Donnelly <pdonnell@redhat.com>`

src/mon/MDSMonitor.cc
jcsp: Fair point, updated to flatten these conditions into the outer series of `else if`s

PR #10429, opened 2016-07-25
mds: add assertion in handle_slave_rename_prep

jcsp:
This function assumes that the source path
does not point to a null dentry, and that
path_traverse is therefore setting the
passed inode.

Candidate code path for crash
in http://tracker.ceph.com/issues/16807


gregsfortytwo:
LGTM

ukernel:
LGTM

PR #10431, opened 2016-07-25
os/bluestore: proper handling for csum enable/disable settings

ifed01:
There is no need to read 'csum aligned' chunks when reading with csum verification disabled. Hence we can tune read performance.


liewegas:
lgtm

liewegas:
Hmm, I wonder if it makes sense to have a different option for this (verify csum on read)?  That way you can separately disable all the csum checks on read, which might be useful/interesting anyway.

markhpc:
@liewegas Would you want the hypothetical option as a separate PR? (IE are we good to go on this one?)

ifed01:
@liewegas I think that's a nice option with a very little use..

PR #10434, opened 2016-07-25
doc: add client config ref

batrick:
Fixes: http://tracker.ceph.com/issues/16743

### DIFF
--- a/doc/cephfs/client-config-ref.rst
+++ b/doc/cephfs/client-config-ref.rst
@@ -1,0 +1,198 @@
+========================
+ Client Config Reference
+========================
+
+``client acl type``
+
+:Description: Set to ``"posix_acl"`` to enable POSIX ACLs.
+:Type: String
+:Default: ``""`` (N/A)
+
+``client cache mid``
+
+:Description: Set client cache mid-point.
+:Type: Float
+:Default: ``0.75``
+
+``client_cache_size``
+
+:Description: Set client cache object count maximum.
+:Type: Integer
+:Default: ``16384``
+
+``client_caps_release_delay``
+
+:Description: Delay between capability releases.
+:Type: Integer
+:Default: ``5`` (seconds)
+
+``client_debug_force_sync_read``
+
+:Description: Always read synchronously (go to osds).
+:Type: Boolean
+:Default: ``false``
+
+``client_debug_getattr_caps``
+
+:Description: Check if MDS reply contains wanted caps.
+:Type: Boolean
+:Default: ``false``
+
+``client_dirsize_rbytes``
+
+:Description: Use recursive size of directory for directory ``st_size`` stat member.
+:Type: Boolean
+:Default: ``true``
+
+``client_max_inline_size``
+
+:Description: Maximum size of inlined data stored in a file inode (rather than in a separate data object in RADOS). This setting only applies if the ``inline_data`` flag is set on the MDSMap.
+:Type: Integer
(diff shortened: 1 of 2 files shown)


batrick:
There are a few config option descriptions that I'm unsure on (marked by FIXME). Could use some suggestions.

doc/cephfs/client-config-ref.rst
```diff
@@ -0,0 +1,204 @@
...
+
+``client_dirsize_rbytes``
+
+:Description: Use recursive size of directory for directory ``st_size`` stat member.
+:Type: Boolean
+:Default: ``true``
+
+``client_max_inline_size``
+
+:Description: **FIXME**
```
jcsp: It's the largest a file can be and still get inlined (i.e. stored in the inode instead of in a data object).  Only applies if the inline_data flag is set in the mdsmap.

doc/cephfs/client-config-ref.rst
```diff
@@ -0,0 +1,204 @@
...
+
+``client_mountpoint``
+
+:Description: Directory to mount on the CephFS file system.
+:Type: String
+:Default: ``"/"``
+
+``client_notify_timeout``
+
+:Description: **FIXME**
```
jcsp: Not a cephfs option, applies to librados clients using RADOS watch/notify operations

batrick:
Done, thanks @jcsp.

doc/cephfs/client-config-ref.rst
```diff
@@ -0,0 +1,198 @@
...
+
+``client_mountpoint``
+
+:Description: Directory to mount on the CephFS file system.
+:Type: String
+:Default: ``"/"``
+
+``client_oc``
+
+:Description: Enable object caching.
```
gregsfortytwo: I don't think we know if this works, or really expect it to? (Disabling it, that is.) we probably shouldn't document it.

doc/cephfs/client-config-ref.rst
```diff
@@ -0,0 +1,198 @@
...
+
+``client_quota_df``
+
+:Description: Report root directory quota for statfs operation.
+:Type: Boolean
+:Default: ``true``
+
+``client_readahead_max_bytes``
+
+:Description: Maximum bytes of readahead used for future read operations. Overridden by ``client_readahead_max_bytes``.
```
gregsfortytwo: This config option overrides itself? ;)

doc/cephfs/client-config-ref.rst
```diff
@@ -0,0 +1,198 @@
...
+
+``client_tick_interval``
+
+:Description: Interval between capability renewal and other upkeep.
+:Type: Float
+:Default: ``1.0`` (seconds)
+
+``client_trace``
+
+:Description: Trace file path for all file operations.
```
gregsfortytwo: This is for outputting trace data, right? And it can then be replayed in the SyntheticClient, I think. If we're going to document it lets make that a little clearer. (Or just leave it out entirely?)

gregsfortytwo:
Do we really want to point people at the monitor-routed inject-arts instead of the admin socket config set? :/

doc/cephfs/client-config-ref.rst
batrick: I suggest marking the config option as ignored and removing the checks in the code.

batrick:
Thanks for the suggestions Greg. I amended the commit.

PR #10436, opened 2016-07-25
Add documentation about snapshots

gregsfortytwo:
This includes some broader developer data that should give people a base when looking at our snapshotting system. It also adds a warning about multi-FS and snapshot compatibility problems.

### DIFF
--- a/doc/cephfs/experimental-features.rst
+++ b/doc/cephfs/experimental-features.rst
@@ -60,6 +60,13 @@ writing, but there is insufficient testing to provide stability guarantees and
 every expansion of testing has generally revealed new issues. If you do enable
 snapshots and experience failure, manual intervention will be needed.
 
+Snapshots are known not to work properly with multiple filesystems (below) in
+some cases. Specifically, if you share a pool for multiple FSes and delete
+a snapshot in one FS, expect to lose snapshotted file data in any other FS using
+snapshots. See the :doc:`/dev/cephfs-snapshots` page for more information.
+
+Snapshots are known not to work with multi-MDS filesystems.
+
 Snapshotting was blocked off with the "allow_new_snaps" flag prior to Firefly.
 
 Multiple filesystems within a Ceph cluster
@@ -76,6 +83,9 @@ active filesystems in your cluster will require manual intervention and, so far,
 will not have been experienced by anybody else -- knowledgeable help will be
 extremely limited. You also probably do not have the security or isolation
 guarantees you want or think you have upon doing so.
+
+Note that snapshots and multiple filesystems are *not* tested in combination
+and may not work together; see above.
 
 Multiple filesystems were available starting in the Jewel release candidates
 but were protected behind the "enable_multiple" flag before the final release.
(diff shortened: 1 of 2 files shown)


gregsfortytwo:
@ukernel and @jcsp, does this look good?

doc/cephfs/experimental-features.rst
```diff
@@ -60,6 +60,11 @@ writing, but there is insufficient testing to provide stability guarantees and
 every expansion of testing has generally revealed new issues. If you do enable
 snapshots and experience failure, manual intervention will be needed.
 
+Snapshots are known not to work properly with multiple filesystems (below) in
+some cases. Specifically, if you share a pool for multiple FSes and delete
+a snapshot in one FS, expect to lose snapshottedfile data in any other FS using
```
ukernel: snapshottedfile

ukernel: maybe we should add note about snapshot does not work with multiple active MDS

doc/cephfs/experimental-features.rst
```diff
@@ -60,6 +60,13 @@ writing, but there is insufficient testing to provide stability guarantees and
 every expansion of testing has generally revealed new issues. If you do enable
 snapshots and experience failure, manual intervention will be needed.
 
+Snapshots are known not to work properly with multiple filesystems (below) in
+some cases. Specifically, if you share a pool for multiple FSes and delete
```
jcsp: You mean specifically sharing data pools, right?  I think Doug has a patch to forbid that (it's definitely already unsafe).

Then pool sharing will become possible with, http://tracker.ceph.com/issues/15066 so I guess at that point we'll need to put something in to ensure that at most one filesystem using a particular data pool can have snapshots enabled.

doc/cephfs/experimental-features.rst
```diff
@@ -77,5 +84,8 @@ will not have been experienced by anybody else -- knowledgeable help will be
 extremely limited. You also probably do not have the security or isolation
 guarantees you want or think you have upon doing so.
 
+Note that snapshots and multiple filesystems are *not* tested and probably
```
jcsp: Let's not under-sell ourselves, they are tested to an extent.  Maybe we should just fall back to our catch-all language and say they're experimental?

doc/dev/cephfs-snapshots.rst
```diff
@@ -0,0 +1,114 @@
...
+  associate a `SnapContext` with each open file for writing.
+* sr_t: An `sr_t` is the on-disk snapshot metadata. It is part of the containing
+  directory and contains sequence counters, timestamps, the list of associated
+  snapshot IDs, and `past_parents`.
+* snaplink_t: `past_parents` et al are stored on-disk as a `snaplink_t`, holding
+  the inode number and first `snapid` of the inode/snapshot referenced.
+
+Creating a snapshot
+----------
+To make a snapshto on directory "/1/2/3/foo", the client invokes "mkdir" on
```
jcsp: snapshto

doc/cephfs/experimental-features.rst
gregsfortytwo: Ah yeah.

doc/cephfs/experimental-features.rst
gregsfortytwo: I meant tested in combination; will update.

PR #10438, opened 2016-07-26
ARM64 fixes

dmick:
Three things to fix build on arm64.  The changes for jerasure and test should not affect x86, but the rocksdb update certainly might; marking needs-qa because of that.

(We need a CI builder for arm64.  Something happened to the Cavium-supplied gitbuilder and I'm not sure what.)

### DIFF
--- a/src/erasure-code/jerasure/CMakeLists.txt
+++ b/src/erasure-code/jerasure/CMakeLists.txt
@@ -45,6 +45,7 @@ if(ARM_NEON OR ARM_NEON2)
     gf-complete/src/neon/gf_w32_neon.c
     gf-complete/src/neon/gf_w64_neon.c)
   add_library(neon_objs OBJECT ${neon_objs_srcs})
+  set_target_properties(neon_objs PROPERTIES COMPILE_FLAGS ${ARM_NEON_FLAGS})
 
   set(jerasure_neon_srcs
     $<TARGET_OBJECTS:ec_jerasure_objs>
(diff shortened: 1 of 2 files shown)


tchaikov:
@tone-zhang maybe you could try this PR to see if it works for you?

tone-zhang:
@tchaikov No problem, I will double verify the PR in my AArch64 platform.

cbodley:
:+1:

alimaredia:
lgtm. once this is verified someone go ahead and merge it!

dmick:
Reemphasizing this should see a fullish qa run because it's updating rocksdb.  I'm a noob in current process so if I need to do something other than mark needs-qa, let me know before merging.  I could schedule a suite of some sort at least?

dmick:
ah, I also hadn't realized there was a lot of duplicated work in #10427; sorry about that.

tchaikov:
@tone-zhang how is your test going, does this PR work for you?

tone-zhang:
@tchaikov Kefu, sorry for the late, the test bed in our side is very busy. 
I take the update in the PR, and build the Ceph locally, the build has been passed.

tchaikov:
@tone-zhang thanks for testing! merging. as your change actually depends on this one. so we'd better get it merged before yours.

tchaikov:
>  if I need to do something other than mark needs-qa, let me know before merging. I could schedule a suite of some sort at least?

@dmick rocksdb is used by bluestore, and is used as an optional backend of our kv store, which is in turn leveldb by default. we have a test suite named "mon_kv_backend". i will run your branch through it.

tone-zhang:
@tchaikov Kefu, all right. After the merge, I will continue the PR #10427 if there was anything left.

tchaikov:
being tested at http://pulpito.ceph.com/kchai-2016-07-27_00:28:21-rados-wip-arm64---basic-mira/

dmick:
tnx @tchaikov

PR #10440, opened 2016-07-26
msg/async: remove static member variable

tchaikov:
msgr is part of libcommon. and the later is linked into both libcephfs
and librados. and hence into cephfs.so and rados.so. that's why we have
double free.

Fixes: http://tracker.ceph.com/issues/16686

### DIFF
--- a/src/msg/async/AsyncMessenger.cc
+++ b/src/msg/async/AsyncMessenger.cc
@@ -342,8 +342,6 @@ class WorkerPool {
     return coreids[id % coreids.size()];
   }
   void barrier();
-  // uniq name for CephContext to distinguish differnt object
-  static const string name;
 };
 
 void *Worker::entry()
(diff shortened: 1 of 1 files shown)


yuyuyu101:
Reviewed-by:

tchaikov:
@jcsp this change works for me. could you help check if it fixes your issue or not?

ajarr:
@tchaikov , this fix works for me in my local setup. I was able to run the VolumeClient tests using vstart_runner. And thanks for the explanation here, http://permalink.gmane.org/gmane.comp.file-systems.ceph.devel/32519

jcsp:
@tchaikov looks good to me, thanks

jcsp:
(I haven't tried it personally but I see that @ajarr has, which is enough for me)

PR #10448, opened 2016-07-27
rgw: Rework of s3 v2 local authentication code.

pritha-srivastava:
The v2 local auth code has been reworked based on the new new authentication infrastructure.

### DIFF
--- a/src/rgw/rgw_rest_s3.cc
+++ b/src/rgw/rgw_rest_s3.cc
@@ -1714,8 +1714,8 @@ int RGWPostObj_ObjStore_S3::get_policy()
     op_ret = rgw_get_user_info_by_access_key(store, s3_access_key, user_info);
     if (op_ret < 0) {
       S3AuthFactory aplfact(store, s->account_name);
-      RGWGetPolicyLDAPTokenExtractor token_extr(s3_access_key);
-      RGWLDAPAuthEngine ldap(s->cct, store, token_extr, &aplfact);
+      RGWGetPolicyV2Extractor extr(s3_access_key, received_signature_str);
+      RGWLDAPAuthEngine ldap(s->cct, store, extr, &aplfact);
         // try external authenticators
       if (store->ctx()->_conf->rgw_s3_auth_use_keystone &&
 	  store->ctx()->_conf->rgw_keystone_url.empty())
@@ -1754,13 +1754,22 @@ int RGWPostObj_ObjStore_S3::get_policy()
 	  s->perm_mask = RGW_PERM_FULL_CONTROL;
 	}
   } else if (ldap.is_applicable()) {
-    auto applier = ldap.authenticate();
-    if (! applier) {
+    try {
+      auto applier = ldap.authenticate();
+      if (! applier) {
+        return -EACCES;
+      }
+
+      try {
+        applier->load_acct_info(*s->user);
+        s->perm_mask = applier->get_perm_mask();
+        applier->modify_request_state(s);
+        s->auth_identity = std::move(applier);
+      } catch (int err) {
+        return -EACCES;
+      }
+    } catch (int err) {
       return -EACCES;
-    } else {
-      applier->load_acct_info(*s->user);
-      s->perm_mask = applier->get_perm_mask();
-      s->auth_identity = std::move(applier);
     }
       } else {
 	return -EACCES;
@@ -3949,91 +3958,57 @@ int RGW_Auth_S3::authorize_v2(RGWRados *store, struct req_state *s)
   }
 
   S3AuthFactory aplfact(store, s->account_name);
-  RGWLDAPTokenExtractor token_extr(s);
-  RGWLDAPAuthEngine ldap(s->cct, store, token_extr, &aplfact);
+  RGWS3V2Extractor extr(s);
+  RGWLDAPAuthEngine ldap(s->cct, store, extr, &aplfact);
 
   if (ldap.is_applicable()) {
-    auto applier = ldap.authenticate();
-    if (!applier) {
-      external_auth_result = -EACCES;
-    } else {
-      applier->load_acct_info(*s->user);
-      s->perm_mask = applier->get_perm_mask();
-      s->auth_identity = std::move(applier);
-      external_auth_result = 0;
+    try {
+      auto applier = ldap.authenticate();
+      if (! applier) {
+        external_auth_result = -EACCES;
+      } else {
+        try {
+          applier->load_acct_info(*s->user);
+          s->perm_mask = applier->get_perm_mask();
+          applier->modify_request_state(s);
+          s->auth_identity = std::move(applier);
+          external_auth_result = 0;
+        } catch (int err) {
+          ldout(s->cct, 5) << "applier threw err=" << err << dendl;
+          external_auth_result = err;
+        }
+      }
+    } catch (int err) {
+      ldout(s->cct, 5) << "ldap auth engine threw err=" << err << dendl;
+      external_auth_result = err;
     }
   }
 
-  /* keystone failed (or not enabled); check if we want to use rados backend */
-  if (!store->ctx()->_conf->rgw_s3_auth_use_rados
-      && external_auth_result < 0)
-    return external_auth_result;
-
   /* now try rados backend, but only if keystone did not succeed */
   if (external_auth_result < 0) {
-    /* get the user info */
-    if (rgw_get_user_info_by_access_key(store, auth_id, *(s->user)) < 0) {
-      dout(5) << "error reading user info, uid=" << auth_id
-              << " can't authenticate" << dendl;
-      return external_auth_result;
-    }
+      RGWS3V2LocalAuthEngine localauth(s, store, extr, &aplfact);
 
-    /* now verify  signature */
-    string auth_hdr;
-    if (!rgw_create_s3_canonical_header(s->info, &s->header_time, auth_hdr,
-					qsr)) {
-      dout(10) << "failed to create auth header\n" << auth_hdr << dendl;
-      return -EPERM;
-    }
-    dout(10) << "auth_hdr:\n" << auth_hdr << dendl;
-
-    time_t req_sec = s->header_time.sec();
-    if ((req_sec < now - RGW_AUTH_GRACE_MINS * 60 ||
-        req_sec > now + RGW_AUTH_GRACE_MINS * 60) && !qsr) {
-      dout(10) << "req_sec=" << req_sec << " now=" << now
-               << "; now - RGW_AUTH_GRACE_MINS=" << now - RGW_AUTH_GRACE_MINS * 60
-               << "; now + RGW_AUTH_GRACE_MINS=" << now + RGW_AUTH_GRACE_MINS * 60
-               << dendl;
-      dout(0)  << "NOTICE: request time skew too big now=" << utime_t(now, 0)
-               << " req_time=" << s->header_time
-               << dendl;
-      return -ERR_REQUEST_TIME_SKEWED;
-    }
-
-    map<string, RGWAccessKey>::iterator iter =
-      s->user->access_keys.find(auth_id);
-    if (iter == s->user->access_keys.end()) {
-      dout(0) << "ERROR: access key not encoded in user info" << dendl;
-      return -EPERM;
-    }
-    RGWAccessKey& k = iter->second;
-
-    if (!k.subuser.empty()) {
-      map<string, RGWSubUser>::iterator uiter =
-	s->user->subusers.find(k.subuser);
-      if (uiter == s->user->subusers.end()) {
-	dout(0) << "NOTICE: could not find subuser: " << k.subuser << dendl;
-	return -EPERM;
+      if (! localauth.is_applicable()) {
+        return external_auth_result;
       }
-      RGWSubUser& subuser = uiter->second;
-      s->perm_mask = subuser.perm_mask;
-    } else
-      s->perm_mask = RGW_PERM_FULL_CONTROL;
-
-    string digest;
-    int ret = rgw_get_s3_header_digest(auth_hdr, k.key, digest);
-    if (ret < 0) {
-      return -EPERM;
-    }
-
-    dout(15) << "calculated digest=" << digest << dendl;
-    dout(15) << "auth_sign=" << auth_sign << dendl;
-    dout(15) << "compare=" << auth_sign.compare(digest) << dendl;
-
-    if (auth_sign != digest) {
-      return -ERR_SIGNATURE_NO_MATCH;
-    }
-
+      try {
+        auto applier = localauth.authenticate();
+        if (! applier) {
+          return external_auth_result;
+        }
+        try {
+          applier->load_acct_info(*s->user);
+          s->perm_mask = applier->get_perm_mask();
+          applier->modify_request_state(s);
+          s->auth_identity = std::move(applier);
+        } catch (int err) {
+          ldout(s->cct, 5) << "applier threw err=" << err << dendl;
+          return err;
+        }
+      } catch (int err) {
+          ldout(s->cct, 5) << "local auth engine threw err=" << err << dendl;
+          return err;
+      }
     if (s->user->system) {
       s->system_request = true;
       dout(20) << "system request" << dendl;
@@ -4042,7 +4017,7 @@ int RGW_Auth_S3::authorize_v2(RGWRados *store, struct req_state *s)
       RGWUserInfo effective_user;
       if (!effective_uid.empty()) {
         rgw_user euid(effective_uid);
-        ret = rgw_get_user_info_by_uid(store, euid, effective_user);
+        int ret = rgw_get_user_info_by_uid(store, euid, effective_user);
         if (ret < 0) {
           ldout(s->cct, 0) << "User lookup failed!" << dendl;
           return -EACCES;
@@ -4329,9 +4304,9 @@ std::mutex RGWLDAPAuthEngine::mtx;
 
 void RGWLDAPAuthEngine::init(CephContext* const cct)
 {
-  if (!ldh) {
+  if (! ldh) {
     std::lock_guard<std::mutex> lck(mtx);
-    if (!ldh) {
+    if (! ldh) {
       const string& ldap_uri = cct->_conf->rgw_ldap_uri;
       const string& ldap_binddn = cct->_conf->rgw_ldap_binddn;
       const string& ldap_searchdn = cct->_conf->rgw_ldap_searchdn;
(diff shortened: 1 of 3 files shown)


pritha-srivastava:
This PR contains reworked code for s3 v2 local auth. 
1. New classes: RGWS3V2TokenBasedAuthEngine, RGWS3V2TokenExtractor, and RGWS3V2LocalAuthEngine have been introduced.
2. RGWS3V2TokenBasedAuthEngine has the additional functionality of extracting the other tokens required for local auth.
3. The RGWS3V2TokenExtractor class can be re-used for LDAP auth code as well.
4. Both RGWS3V2TokenBasedAuthEngine and RGWS3V2Extractor can be used for V2 Keystone auth as well.
5. Another alternative would be to make RGWS3V2TokenBasedAuthEngine to derive from RGWAuthEngine (and not from RGWTokenBasedAuthEngine) and it takes care of getting all the tokens including auth_id. The auth engine class can then be re-used both for local and keystone auth. The extractor will not be needed in this case. And the ldap code in the other PR can remain unchanged. This approach is simple. Please let me know what you think.

src/rgw/rgw_rest_s3.cc
```diff
@@ -4410,3 +4357,118 @@ RGWOp* RGWHandler_REST_Service_S3Website::get_obj_op(bool get_data)
...
+  }
+  RGWAccessKey& k = iter->second;
+
+  string digest;
+  int ret = rgw_get_s3_header_digest(auth_hdr, k.key, digest);
+  if (ret < 0) {
+    return nullptr;
+  }
+
+  dout(15) << "calculated digest=" << digest << dendl;
```
theanalyst: could we probably replace dout -> ldout if we have s->cct here?

mattbenjamin: harmless for now?

mattbenjamin:
This commit now has conflicts with master, and needs a rebase.  Pritha, can you rebase and re-post?

mattbenjamin:
Also, just quickly:  does this change really only apply to v2?  Will another change be required for v4?

src/rgw/rgw_rest_s3.cc
theanalyst: Yeah harmless, though we do we have req_info, ie. s->cct here :)

pritha-srivastava:
@mattbenjamin : Yes, this change only applies to v2. But I will update this PR with changes for v4 also as you have suggested.

pritha-srivastava:
@mattbenjamin, @rzarzynski  : I need some feedback on the design of classes RGWS3V2TokenBasedAuthEngine and RGWS3V2TokenExtractor. Instead of having an extractor class for just the auth id and the other tokens getting extracted in RGWS3V2TokenBasedAuthEngine, I can just have RGWS3V2TokenBasedAuthEngine derive from RGWAuthEngine and not have any extractor class. The  RGWS3V2TokenBasedAuthEngine class will take care of extracting all auth tokens (like auth id, signature and expires). This class could be re-used for Keystone Auth as well (and maybe for ldap also.)

pritha-srivastava:
@mattbenjamin : In the old code for ldap auth, the code would fallback to local auth incase ldap auth failed. Is that really intended here? Or should we just send an error to the user in case ldap auth fails (and not proceed to ocal auth)?

src/rgw/rgw_rest_s3.h
```diff
@@ -689,7 +689,67 @@ class RGWLDAPTokenExtractor : public RGWTokenBasedAuthEngine::Extractor {
   std::string get_token() const override;
 };
 
-class S3AuthFactory : public RGWRemoteAuthApplier::Factory {
+class RGWS3V2TokenBasedAuthEngine : public RGWTokenBasedAuthEngine {
```
rzarzynski: @pritha-srivastava: I'm not sure whether deriving from `RGWTokenBasedAuthEngine` is suitable way for AWS auth. The class is targeting those authenticators which are token-based. OpenStack Identity API and Kerberos might be seen as typical examples. It's worth to add that under *token* I mean a short-living string that can be used multiple times without alternations. A [comment on StackOverflow](http://stackoverflow.com/a/1592572) might be helpful as well.

AWSv2 isn't token-based in this sense. Instead of providing a constant string, client needs to send `AWSAccessKeyId` and `Signature` according the Amazon's terminology. `Signature` is varying with each request.

The second solution proposed by you seems to be the good one. However, I wouldn't name the class with anything having *Token* inside. Also the extractor concept still could be useful. IIRC the AWS auth stuff could be extracted either from HTTP headers or from query string.

src/rgw/rgw_rest_s3.cc
```diff
@@ -3966,76 +3966,23 @@ int RGW_Auth_S3::authorize_v2(RGWRados *store, struct req_state *s)
...
-    }
-    RGWAccessKey& k = iter->second;
-
-    if (!k.subuser.empty()) {
-      map<string, RGWSubUser>::iterator uiter =
-	s->user->subusers.find(k.subuser);
-      if (uiter == s->user->subusers.end()) {
-	dout(0) << "NOTICE: could not find subuser: " << k.subuser << dendl;
-	return -EPERM;
+    if (! localauth.is_applicable()) {
```
rzarzynski: Please take care of potential exceptions here - both from auth engine and auth applier. You might want to take a look on [the corresponding implementation for the Swift API](https://github.com/ceph/ceph/blob/master/src/rgw/rgw_rest_swift.cc#L1545-L1575).

src/rgw/rgw_rest_s3.cc
```diff
@@ -3966,76 +3966,23 @@ int RGW_Auth_S3::authorize_v2(RGWRados *store, struct req_state *s)
...
+      return external_auth_result;
+    } else {
+      auto applier = localauth.authenticate();
+      if (! applier) {
```
rzarzynski: I really like the style with putting a space after negation operator!

src/rgw/rgw_rest_s3.cc
```diff
@@ -3966,76 +3966,23 @@ int RGW_Auth_S3::authorize_v2(RGWRados *store, struct req_state *s)
...
+        return external_auth_result;
+      } else {
+        applier->load_acct_info(*s->user);
+        s->perm_mask = applier->get_perm_mask();
+        s->auth_identity = std::move(applier);
```
rzarzynski: I guess we want to call `modify_request_state` here even if it actually does nothing.

src/rgw/rgw_rest_s3.cc
```diff
@@ -4420,3 +4367,117 @@ std::string RGWLDAPTokenExtractor::get_token() const
   }
   return token;
 }
+
+bool RGWS3V2TokenBasedAuthEngine::is_applicable() const noexcept
+{
+  if (!RGWTokenBasedAuthEngine::is_applicable()) {
```
rzarzynski: It would be nice the have the space after `!`.

src/rgw/rgw_rest_s3.cc
```diff
@@ -4420,3 +4367,117 @@ std::string RGWLDAPTokenExtractor::get_token() const
...
+    return false;
+  }
+
+  if (!date.empty()) {
+    time_t exp = atoll(date.c_str());
+    if (now >= exp) {
+      return false;
```
rzarzynski: Are you sure that only skipping a given auth engine when `Expires` is in the past  is the intended behavior? I saw that `RGW_Auth_S3::authorize_v2` signalizes `EPERM` in such case.

pritha-srivastava:
@rzarzynski : Thank you very much for your comments.

src/rgw/rgw_rest_s3.cc
```diff
@@ -4393,30 +4385,122 @@ RGWAuthApplier::aplptr_t RGWLDAPAuthEngine::authenticate() const
   RGWUserInfo user_info;
   user_info.user_id = base64_token.id;
   if (rgw_get_user_info_by_uid(store, user_info.user_id, user_info) >= 0) {
-    if (user_info.type != TYPE_LDAP) {
+    /*if (user_info.type != TYPE_LDAP) {
       ldout(cct, 10) << "ERROR: User id of type: " << user_info.type << " is already present" << dendl;
       return nullptr;
-    }
+    }*/
   }
```
pritha-srivastava: I will enable the type checking once we have a plan to migrate existing users to the new format.

src/rgw/rgw_rest_s3.cc
```diff
@@ -4393,30 +4385,122 @@ RGWAuthApplier::aplptr_t RGWLDAPAuthEngine::authenticate() const
...
+      ldout(s->cct, 5) << "error reading user info, uid=" << access_key_id
+              << " can't authenticate" << dendl;
+      throw -EPERM;
+  } /*else {
+    if (s->user->type != TYPE_RGW) {
+      ldout(cct, 10) << "ERROR: User id of type: " << s->user->type << " is present" << dendl;
+      throw -EPERM;
+    }
+  }*/
+
```
pritha-srivastava: Same comment as above, this will be enabled when the users are migrated to the new type.

src/rgw/rgw_rest_s3.cc
```diff
@@ -3949,93 +3976,58 @@ int RGW_Auth_S3::authorize_v2(RGWRados *store, struct req_state *s)
...
+        ldout(s->cct, 5) << "applier threw err=" << err << dendl;
+        return err;
+      }
+    } catch (int err) {
+      ldout(s->cct, 5) << "ldap auth engine threw err=" << err << dendl;
+      return err;
     }
   }
 
-  /* keystone failed (or not enabled); check if we want to use rados backend */
```
pritha-srivastava: Will cascade keystone and ldap auth in the PR for keystone

pritha-srivastava:
Note:
1. I have introduced a RGWS3V2AuthEngine and RGWS3V2Extractor class which are being used by ldap and local authentication now.
2. I can define a standalone Extractor class and use aggregation with AuthEngine classes so that the tokens are not parsed again for every new AuthEngine class that is defined - I will explore this in keystone rework.

pritha-srivastava:
@mattbenjamin @rzarzynski : Please review

src/rgw/rgw_rest_s3.cc
```diff
@@ -1713,8 +1713,26 @@ int RGWPostObj_ObjStore_S3::get_policy()
 
     op_ret = rgw_get_user_info_by_access_key(store, s3_access_key, user_info);
     if (op_ret < 0) {
+      class S3Extractor:public RGWS3V2AuthEngine::Extractor {
+        std::string access_key_id;
+        std::string signature;
+
+        public:
+          S3Extractor(std::string access_key_id, std::string signature)
+          : access_key_id(access_key_id),
```
rzarzynski: I would suggest std::moving here to save additional memory allocation/increasing a reference counter in `std::string`.

src/rgw/rgw_rest_s3.h
```diff
@@ -645,7 +645,48 @@ static inline int valid_s3_bucket_name(const string& name, bool relaxed=false)
...
+  std::string access_key_id;
+  std::string signature;
+  std::string expires;
+  bool qsr;
+
+public:
+  class Extractor {
+  public:
+    virtual ~Extractor() {};
+    virtual void get_tokens(std::string& access_key_id,
```
rzarzynski: Hmm, not sure whether `get_tokens` is a good name for S3.

src/rgw/rgw_rest_s3.cc
```diff
@@ -1713,8 +1713,26 @@ int RGWPostObj_ObjStore_S3::get_policy()
 
     op_ret = rgw_get_user_info_by_access_key(store, s3_access_key, user_info);
     if (op_ret < 0) {
+      class S3Extractor:public RGWS3V2AuthEngine::Extractor {
```
rzarzynski: Is the `S3Extractor` a makeshift solution? At the moment it only stores already extracted data. Is there a plan to make it also responsible for extracting them from `req_state`?

pritha-srivastava: It is kind of a makeshift solution, I will have to rework it to extract the tokens also in this class. I plan to do that later. Right now I want to concentrate on making authorize_v2 perfect, then will come back to this

pritha-srivastava:
@rzarzynski : Reworked.

src/rgw/rgw_rest_s3.cc
```diff
@@ -1754,13 +1754,22 @@ int RGWPostObj_ObjStore_S3::get_policy()
 	  s->perm_mask = RGW_PERM_FULL_CONTROL;
 	}
   } else if (ldap.is_applicable()) {
-    auto applier = ldap.authenticate();
-    if (! applier) {
+    try {
```
rzarzynski: OK.

src/rgw/rgw_rest_s3.cc
```diff
@@ -3949,93 +3958,57 @@ int RGW_Auth_S3::authorize_v2(RGWRados *store, struct req_state *s)
...
-    if (!k.subuser.empty()) {
-      map<string, RGWSubUser>::iterator uiter =
-	s->user->subusers.find(k.subuser);
-      if (uiter == s->user->subusers.end()) {
-	dout(0) << "NOTICE: could not find subuser: " << k.subuser << dendl;
-	return -EPERM;
+      if (! localauth.is_applicable()) {
+        return external_auth_result;
+      }
+      try {
```
rzarzynski: OK.

src/rgw/rgw_rest_s3.h
```diff
@@ -645,7 +645,48 @@ static inline int valid_s3_bucket_name(const string& name, bool relaxed=false)
...
+  const req_state* const s;
+
+public:
+  RGWS3V2Extractor(const req_state * const s)
+    : s(s) {}
+
+  void get_auth_keys(std::string& access_key_id,
+                  std::string& signature,
+                  std::string& expires,
+                  bool& qsr) const;
```
rzarzynski: `override` would be very helpful here. Also indentation of parameters might be improved but this is less than minor.

src/rgw/rgw_rest_s3.h
```diff
@@ -678,42 +719,65 @@ class RGWLDAPAuthEngine: RGWTokenBasedAuthEngine
...
-  RGWGetPolicyLDAPTokenExtractor(std::string access_key_id) {
-    access_key_id = std::move(access_key_id);
+  RGWGetPolicyTokenExtractor(std::string access_key_id, std::string signature) {
+    access_key_id = std::move(access_key_id),
+    signature = std::move(signature);
   }
 
-  std::string get_token() const {
-    return access_key_id;
+  void get_auth_keys(std::string& access_key_id,
```
rzarzynski: As the method implementation doesn't set the `qsr` output parameter a bool member of `RGWS3V2AuthEngine` might be left uninitialized. Is this an intended behavior?

src/rgw/rgw_rest_s3.h
```diff
@@ -678,42 +719,65 @@ class RGWLDAPAuthEngine: RGWTokenBasedAuthEngine
...
+                      std::string& signature,
+                      std::string& expires,
+                      bool& qsr) const {
```
rzarzynski: `override` here.

src/rgw/rgw_rest_s3.cc
```diff
@@ -1714,7 +1714,7 @@ int RGWPostObj_ObjStore_S3::get_policy()
     op_ret = rgw_get_user_info_by_access_key(store, s3_access_key, user_info);
     if (op_ret < 0) {
       S3AuthFactory aplfact(store, s->account_name);
-      RGWGetPolicyLDAPTokenExtractor token_extr(s3_access_key);
+      RGWGetPolicyTokenExtractor token_extr(s3_access_key, received_signature_str);
```
rzarzynski: I still have some doubts regarding the naming. What is *GetPolicyToken*? Does LDAP have the token concept? If not, I would prefer to not use the "token" here.

src/rgw/rgw_rest_s3.cc
```diff
@@ -4331,9 +4304,9 @@ std::mutex RGWLDAPAuthEngine::mtx;
 
 void RGWLDAPAuthEngine::init(CephContext* const cct)
 {
-  if (!ldh) {
+  if (! ldh) {
```
rzarzynski: Nice!

src/rgw/rgw_rest_s3.cc
pritha-srivastava: That name is a mistake - I meant RGWGetPolicyExtractor - which means an extractor for the get_policy method. I will correct that.

src/rgw/rgw_rest_s3.h
pritha-srivastava: Not initializing 'qsr' in this piece of code won't make a difference, or cause any harm. Since the authentication engine (ldap auth) is not using this parameter now.

mattbenjamin:
@pritha-srivastava  I don't think we should just comment out the type checking, since we I believe all agree it should be present (Shilpa raised the issue as a potential bug).

Rather, we should figure out a change following this one which allows dealing with this issue.

pritha-srivastava:
@mattbenjamin : My intention was to keep the type checking commented, till we have the radosgw-admin command in place (or whatever will help in adding the user source type field to the existing users). This is to avoid any confusion if someone may want to use the master branch may see incorrect auth failures. Once the auth rework is done, my plan was to work on the new command and then enable the type checking throughout the code. Let me know what you think.

mattbenjamin:
I think the type checking should be re-enabled.  Yehuda has already agreed that the migration rules can come later, so long as we prioritize them.  If you are not comfortable with that, I'd rather delay merging this change until we have a migration path.

pritha-srivastava:
@mattbenjamin : My only concern was the migration of the existing users. If that can be deffered, then it is fine with me. The code to add the source user type and to check the type is already there. I'll add/ re-enable them and update the PR.

pritha-srivastava:
@mattbenjamin : I have re-enabled the type checking both for ldap and local auth.

mattbenjamin:
@pritha-srivastava I agree it's kind of annoying either way.  I'm honestly responding partly from intuition here.  Changing the behavior between pushes and without discussion felt wrong to me.  I also don't like adding disabled sections terribly much, but then again, I sometimes do it myself.  Here I think we should focus on the real issue that was exposed by this rework, which is the lack of a proper policy governing multiple auth backends.  I share Shilpa's intuition that it's not good to allow dynamic backends like ldap to induce privilege escalation, so I've been trying to force us to address it.

mattbenjamin:
@pritha-srivastava I forgot to add.  Do you have a sense of what strategies we should take for the migration?  I wrote down the ideas I had come up with previously.  Do you have other ideas?  Let's discuss this in standup.

pritha-srivastava:
@mattbenjamin : The note about disabling the user type check was intended to bring about the discussion, but I will bring it up in the stand-up next time. Sorry about that.

Yes, I did think a little about the migrating existing users: We could probably use radosgw-admin modify user command, and give the users an option to add the source type. We could write up a script using this command addressing all users. I am not sure though how many users the end-user will typically have.

mattbenjamin:
@pritha-srivastava I'll merge as soon as the next push re-disabling checks lands.

pritha-srivastava:
@mattbenjamin : Checking teuthology errors, will update the PR once done

pritha-srivastava:
@mattbenjamin : I have updated the PR. But there is still one job in which one of the s3tests timeout randomly because of an IO error. I have run the job twice and each time a new s3 test timed out. I am running it on a different machine now. @cbodley has agreed to check the result and let you know. Please dont merge the PR till then. Thanks!

cbodley:
@pritha-srivastava it hadn't run by the time i left today, but you guys can track progress at http://pulpito.ceph.com/prsrivas-2016-08-11_19:53:53-rgw-wip_s3_v2_local_auth_rework---basic-mira/

PR #10450, opened 2016-07-27
osd: cleanup options and other redundancies

xiexingguo:


xiexingguo:
@athanatos Can you look? Thanks!

src/common/config_opts.h
```diff
@@ -628,10 +628,6 @@ OPTION(osd_pool_default_size, OPT_INT, 3)
 OPTION(osd_pool_default_min_size, OPT_INT, 0)  // 0 means no specific default; ceph will use size-size/2
 OPTION(osd_pool_default_pg_num, OPT_INT, 8) // number of PGs for new pools. Configure in global or mon section of ceph.conf
 OPTION(osd_pool_default_pgp_num, OPT_INT, 8) // number of PGs for placement purposes. Should be equal to pg_num
-OPTION(osd_compression_plugins, OPT_STR,
-       "zlib"
-       " snappy"
-       ) // list of compression plugins
 OPTION(osd_pool_default_erasure_code_profile,
```
athanatos: @Ved-vampir Is this ok to remove?

src/osd/PG.cc
```diff
@@ -4108,7 +4108,7 @@ void PG::chunky_scrub(ThreadPool::TPHandle &handle)
             }
 
             // reset handle once in a while, the search maybe takes long.
-            if (++loop >= g_conf->osd_loop_before_reset_tphandle) {
+            if (++loop >= g_conf->osd_scan_list_ping_tp_interval) {
               handle.reset_tp_timeout();
```
athanatos: osd_scan_list_ping_tp_interval is about IO.  I would actually prefer to have two configs here.  Please remove this commit.

src/osd/OSD.h
```diff
@@ -1857,7 +1857,7 @@ class OSD : public Dispatcher,
     void dequeue_and_get_ops(PG *pg, list<OpRequestRef> *dequeued) {
       ShardData* sdata = NULL;
       assert(pg != NULL);
-      uint32_t shard_index = pg->get_pgid().ps()% shard_list.size();
+      uint32_t shard_index = pg->get_pgid().hash_to_shard(shard_list.size());
       sdata = shard_list[shard_index];
```
athanatos: I don't really like doing cleanups like this when there isn't something substantive changing (anyone working on branches that touch this code would have to deal with a silly merge conflict).  We can leave this until something substrantive changes in these methods, I think.  Please remove this commit.

src/common/config_opts.h
Ved-vampir: @athanatos Now it isn't used, so, yes.

athanatos:
Once you address those comments, ping me and I'll add needs-qa.

xiexingguo:
retest this please

yuriw:
http://pulpito.ceph.com/yuriw-2016-07-29_08:07:15-rados-wip-yuri-testing2_2016_7_29-distro-basic-smithi/

PR #10451, opened 2016-07-27
tools: fix the core dump when get the crushmap do not exist

songbaisen:
tools: fix the core dump when get the crushmap version which it  does not exist


tchaikov:
lgtm.

tchaikov:
could you prefix the title of your commit message with the subcomponent your are changing ? see https://github.com/ceph/ceph/blob/master/SubmittingPatches.rst#preparing-and-sending-patches . in this case, would be better to prefix it with "tools: ".

songbaisen:
@tchaikov  Thank you sir.Done it.

tchaikov:
lgtm.

tchaikov:
tested http://pulpito.ceph.com/kchai-2016-08-18_15:35:36-rados-wip-kefu-testing2---basic-mira/

all failures are known ones.

PR #10452, opened 2016-07-27
test: break out librados-using cephfs test

jcsp:
We are seeing an issue due to the lockdep symbols
in libcephfs and librados clashing, which shows itself
after a fork in the flock tests.  We can avoid this
by splitting the libcephfs tests that require librados
(access.cc) into their own compilation unit so that
the flock tests can run in a libcephfs-only process.

### DIFF
--- a/qa/workunits/libcephfs/test.sh
+++ b/qa/workunits/libcephfs/test.sh
@@ -1,5 +1,6 @@
 #!/bin/sh -e
 
 ceph_test_libcephfs
+ceph_test_libcephfs_access
 
 exit 0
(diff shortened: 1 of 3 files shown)


tchaikov:
@jcsp, fwiw, i will also work on a patch to hide the non-public symbols in librados and libcephfs like what autotools is doing as suggested by @batrick .

tchaikov:
#10472 is the PR.

tchaikov:
@jcsp @badone and me were discussing this issue this morning, probably we can do fork() in a safer way? see http://tracker.ceph.com/issues/16556#note-11.

jcsp:
I'll close this once we've seen a fs suite pass with #10472 (which has now merged)

jcsp:
I'm merging this because without it #9566's tests were failing (presumably due to who-knows-what other fork()-related weirdness that I don't want to investigate right now)

PR #10455, opened 2016-07-27
src/kv/MemDB.cc: the type of the parameter of push_back() does not match the ops's value_type

wjwithagen:
- NULL is a C-type usuage, which is allow by GCC.
   But Clang is more strick in checking:

Error looks like:
src/kv/MemDB.cc:228:7r error: no matching member function
for call to 'push_back'
  ops.push_back(make_pair(DELETE,
  ~~~~^~~~~~~~~
/usr/include/c++/v1/vector:685:36: note: candidate function not viable:
no known conversion from 'pair<[...], pair<[...], nullptr_t>>' to
      'const pair<[...], pair<[...], ceph::buffer::list>>' for 1st argument
    _LIBCPP_INLINE_VISIBILITY void push_back(const_reference __x);
                                   ^

### DIFF
--- a/src/kv/MemDB.cc
+++ b/src/kv/MemDB.cc
@@ -227,7 +227,7 @@ void MemDB::MDBTransactionImpl::rmkey(const string &prefix,
   dtrace << __func__ << " " << prefix << " " << k << dendl;
   ops.push_back(make_pair(DELETE,
                           std::make_pair(std::make_pair(prefix, k),
-                          NULL)));
+                          bufferlist())));
 }
 
 void MemDB::MDBTransactionImpl::rmkeys_by_prefix(const string &prefix)


tchaikov:
lgtm,

@chhabaramesh what do you think?

tchaikov:
@wjwithagen could you please amend your commit message?

wjwithagen:
@tchaikov 
Fixed typos. Or did you have something else that needed amending?

src/kv/MemDB.cc
```diff
@@ -227,7 +227,7 @@ void MemDB::MDBTransactionImpl::rmkey(const string &prefix,
   dtrace << __func__ << " " << prefix << " " << k << dendl;
   ops.push_back(make_pair(DELETE,
                           std::make_pair(std::make_pair(prefix, k),
-                          NULL)));
+                          bufferlist() )));
```
tchaikov: could remove the space after `bufferlist()`

tchaikov:
@wjwithagen i explained the reason why this piece of code failed to compile in the reply to your mail, and put a short version of it in the comment of https://github.com/ceph/ceph/pull/10455#issuecomment-235643268. also the typo in your commit message was not fixed (s/strick/strict/). i'd recommend you to correct the commit message.

src/kv/MemDB.cc
wjwithagen: @tchaikov 
Did that on purpose, to separate the function end from the other )s.
But it does not conform the standard.

wjwithagen:
retest this please

src/kv/MemDB.cc
```diff
@@ -227,7 +227,7 @@ void MemDB::MDBTransactionImpl::rmkey(const string &prefix,
   dtrace << __func__ << " " << prefix << " " << k << dendl;
   ops.push_back(make_pair(DELETE,
                           std::make_pair(std::make_pair(prefix, k),
-                          NULL)));
+                          bufferlist())));
 }
```
chhabaramesh: Looks good to me.

wjwithagen:
@tchaikov 
I think this one is also good to go?

PR #10459, opened 2016-07-27
AddCephTest.cmake: Make PYTHONPATH a bit more flexible

wjwithagen:
- Other OSes will have different diretories where cython results
   will be delivered.
   Path is now made flexibel by using Variables
   A version for FreeBSD is added

### DIFF
--- a/cmake/modules/AddCephTest.cmake
+++ b/cmake/modules/AddCephTest.cmake
@@ -1,5 +1,31 @@
 #AddCephTest is a module for adding tests to the "make check" target which runs CTest
 
+execute_process(
+    COMMAND uname -m
+    OUTPUT_VARIABLE ARCHSTR
+    OUTPUT_STRIP_TRAILING_WHITESPACE)
+
+if(LINUX)
+  set(OSSTR linux)
+elseif(FREEBSD)
+  execute_process(
+    COMMAND uname -r
+    OUTPUT_VARIABLE OSVER
+    OUTPUT_STRIP_TRAILING_WHITESPACE)
+  set(OSSTR freebsd-${OSVER})
+else()
+  message(FATAL_ERROR "unsupported OS, please fix PYTHONPATH construction first.")
+  #
+  # If you get this message for your OS you need to work out where
+  # cython would put its libs.
+  # For Linux is is something like:
+  #    ceph/build/lib/cython_modules/lib.linux-x86_64-2.7
+  # For FreeBSD it is something like:
+  #    ceph/build/lib/cython_modules/lib.freebsd-11.0-CURRENT-amd64-2.7
+endif(LINUX)
+
+set(PYVER "${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR}")
+
 #adds makes target/script into a test, test to check target, sets necessary environment variables
 function(add_ceph_test test_name test_path)
   add_test(NAME ${test_name} COMMAND ${test_path})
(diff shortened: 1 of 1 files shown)


cmake/modules/AddCephTest.cmake
```diff
@@ -1,5 +1,22 @@
 #AddCephTest is a module for adding tests to the "make check" target which runs CTest
 
+execute_process(
+    COMMAND uname -m
+    OUTPUT_VARIABLE ARCHSTR
+    OUTPUT_STRIP_TRAILING_WHITESPACE)
+
+if(LINUX)
+  set(OSSTR linux)
+  set(PYVER 2.7)
```
tchaikov: maybe we can take this chance to set `PYVER` like
```cmake
set(PYVER "${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR}")
```
?

wjwithagen: @tchaikov 
I'm all in favour for as much flexibility as possible.
Especially with with Python going to 3.x, this would prevent more trouble.
I looked at the trace-expand output of Cmake to see if I could find anything you suggest, but did not see any.
But I'll give your suggestion a run.

wjwithagen: @tchaikov 
That does work... 
pushed a new commit

cmake/modules/AddCephTest.cmake
```diff
@@ -1,5 +1,22 @@
...
+
+if(LINUX)
+  set(OSSTR linux)
+elseif(FREEBSD)
+  execute_process(
+    COMMAND uname -r
+    OUTPUT_VARIABLE OSVER
+    OUTPUT_STRIP_TRAILING_WHITESPACE)
+  set(OSSTR freebsd-${OSVER})
+endif(LINUX)
```
tchaikov: might want to fail otherwise? something like `message(FATAL_ERROR, "unsupported OS")`?

tchaikov:
lgtm @alimaredia what do you think?

cmake/modules/AddCephTest.cmake
```diff
@@ -1,5 +1,22 @@
 #AddCephTest is a module for adding tests to the "make check" target which runs CTest
 
+execute_process(
```
tchaikov: if `ARCHSTR` is not used on `LINUX`, could you move this call `execute_process()` into the `elseif(FREEBSD)` block?

cmake/modules/AddCephTest.cmake
wjwithagen: @tchaikov 
Generating a FATAL error would prevent the whole CHECK process from running.
Not even sure if cmake will complete and generate any of the Makefiles.

tchaikov: i think that's exactly the expected behaviour on unsupported platforms. and the the error message will guide the user to the right place (here) to fix it.

if we continue with an empty `$OSSTR`, the error is hidden at this moment, and user will be lost when deciphering the error messages when the tests fail.

wjwithagen: @tchaikov 
Valid point.
I was more thinking about the startup trouble I had whilest starting porting.
But the code here is more than obvious to fix for new platforms.
I'll even throw in some free suggestions.

cmake/modules/AddCephTest.cmake
```diff
@@ -1,5 +1,31 @@
...
+if(LINUX)
+  set(OSSTR linux)
+elseif(FREEBSD)
+  execute_process(
+    COMMAND uname -r
+    OUTPUT_VARIABLE OSVER
+    OUTPUT_STRIP_TRAILING_WHITESPACE)
+  set(OSSTR freebsd-${OSVER})
+else()
+  message(FATAL_ERROR, "unsupported OS, please fix PYTHONPATH construction first.")
```
tchaikov: please remove the comma after `FATAL_ERROR`. my bad.

wjwithagen: @tchaikov 
No prob. I also did not pay attention and just cut-pasted it.

tchaikov:
@wjwithagen  i am closing your PR, as it is superseded by #10459.

wjwithagen:
@tchaikov 
'mmm, I think you ment another PR, since this is #10459

tchaikov:
ahh, sorry should be #10568.

PR #10460, opened 2016-07-28
rgw: add tenant to shard_id in RGWDeleteLC::execute()

weiqiaomiao:
in RGWDeleteLC::execute(), we should add bucket.tenant to shard_id,
otherwise, the lc entry can't delete after this op

Signed-off-by: Wei Qiaomiao wei.qiaomiao@zte.com.cn

### DIFF
--- a/src/rgw/rgw_op.cc
+++ b/src/rgw/rgw_op.cc
@@ -4847,7 +4847,7 @@ void RGWDeleteLC::execute()
             << " returned err=" << op_ret << dendl;
     return;
   }
-  string shard_id = s->bucket.name + ':' +s->bucket.bucket_id;
+  string shard_id = s->bucket.tenant + ':' + s->bucket.name + ':' + s->bucket.bucket_id;
   pair<string, int> entry(shard_id, lc_uninitial);
   string oid; 
   get_lc_oid(s, oid);


src/rgw/rgw_op.cc
```diff
@@ -3914,7 +3914,7 @@ void RGWDeleteLC::execute()
   store->get_bucket_instance_obj(s->bucket, obj);
   store->set_atomic(s->obj_ctx, obj);
   ret = get_system_obj_attrs(store, s, obj, orig_attrs, NULL, &s->bucket_info.objv_tracker);
-  if (op_ret < 0)
```
oritwas: it should be:
op_ret = get_system_obj_attrs
so we can return the error

oritwas: why does the class has field ret, it should use RGWOp::op_ret

weiqiaomiao: I don't know why  but I think we can remove it , and I found the class ``RGWGetLC  `` and `` RGWPutLC``  has the field ret too,   may be we should open other PR to fix them, what would you think?

oritwas: I am ok with fixing it in a new pr

oritwas:
@weiqiaomiao , please rebase

weiqiaomiao:
@oritwas  rebased

src/rgw/rgw_op.cc
```diff
@@ -4621,7 +4621,7 @@ void RGWDeleteLC::execute()
       }
     }
   op_ret = rgw_bucket_set_attrs(store, s->bucket_info, attrs, &s->bucket_info.objv_tracker);
-  string shard_id = s->bucket.name + ':' +s->bucket.bucket_id;
+  string shard_id = s->bucket.tenant + ':' + s->bucket.name + ':' + s->bucket.bucket_id;
```
amitkumar50: @weiqiaomiao Please consider in taking PR title < 50 characters

weiqiaomiao: OK~~

dengxiafubi:
#10460 merged

PR #10461, opened 2016-07-28
src/osd/OSD.cc use "bool" instead of "int"

yonghengdexin735:


yonghengdexin735:
[https://github.com/ceph/ceph/pull/10039](10039)

@athanatos I fix up

athanatos:
This doesn't seem to be fixing a bug.  This is the kind of cleanup which makes sense, but really only if something substantial is changing in the same part of the code in the same PR.

PR #10467, opened 2016-07-28
rgw: workaround for deadlock in early versions of curl_multi_wait()

cethikdata:
radosgw Consumes too much CPU time to synchronize metadata or data between multisite and then stop synchronizing after a while.The reason is that RGWHTTPManager thread doesn't read data from pipe. This makes the pipe full of zeroes and RGWDataSyncProcessorThread suspend on write() call.

### DIFF
--- a/src/rgw/rgw_http_client.cc
+++ b/src/rgw/rgw_http_client.cc
@@ -435,10 +435,10 @@ static int do_curl_wait(CephContext *cct, CURLM *handle, int signal_fd)
     return -EIO;
   }
 
-  if (wait_fd.revents > 0) {
+  if (wait_fd.fd > 0) {
     uint32_t buf;
     ret = read(signal_fd, (void *)&buf, sizeof(buf));
-    if (ret < 0) {
+    if (ret < 0 && ret!=EAGAIN) {
       ret = -errno;
       dout(0) << "ERROR: " << __func__ << "(): read() returned " << ret << dendl;
       return ret;
@@ -789,6 +789,18 @@ int RGWHTTPManager::set_threaded()
     r = -errno;
     ldout(cct, 0) << "ERROR: pipe() returned errno=" << r << dendl;
     return r;
+  }
+  r=fcntl(thread_pipe[0],F_SETFL,O_NONBLOCK);
+  if(r<0){
+  	r=-errno;
+  	ldout(cct, 0) << "ERROR: fcntl() returned errno=" << r << dendl;
+  	return r;
+  }
+  r=fcntl(thread_pipe[1],F_SETFL,O_NONBLOCK);
+  if(r<0){
+  	r=-errno;
+  	ldout(cct, 0) << "ERROR: fcntl() returned errno=" << r << dendl;
+  	return r;
   }
 
   is_threaded = true;


src/rgw/rgw_http_client.cc
```diff
@@ -435,10 +435,10 @@ static int do_curl_wait(CephContext *cct, CURLM *handle, int signal_fd)
     return -EIO;
   }
 
-  if (wait_fd.revents > 0) {
+  if (wait_fd.fd > 0) {
     uint32_t buf;
     ret = read(signal_fd, (void *)&buf, sizeof(buf));
-    if (ret < 0) {
+    if (ret < 0 && ret!=EAGAIN) {
```
yonghengdexin735: if (ret < 0 && errno !=EAGAIN) {

cbodley:
@cethikdata could you please update your commit message to match this format?
```
rgw: <short description of change>

<long description is great as it is>

Fixes: http://tracker.ceph.com/issues/16695

Signed-off-by: Name <email>
```

cbodley:
@cethikdata thanks for looking into this!

tracking down this bug in `curl_multi_wait()` was painful. we were able to backport the relevant fixes to curl 7.29 in RHEL (https://bugzilla.redhat.com/show_bug.cgi?id=1347904), but that might take a while for centos (https://bugs.centos.org/view.php?id=11242). so i'm happy to have a workaround in the meantime, and your approach looks like a good one

src/rgw/rgw_http_client.cc
```diff
@@ -790,6 +790,18 @@ int RGWHTTPManager::set_threaded()
...
+  	r=-errno;
+  	ldout(cct, 0) << "ERROR: fcntl() returned errno=" << r << dendl;
+  	return r;
+  }
+  r=fcntl(thread_pipe[1],F_SETFL,O_NONBLOCK);
+  if(r<0){
+  	r=-errno;
+  	ldout(cct, 0) << "ERROR: fcntl() returned errno=" << r << dendl;
+  	return r;
+  }
```
cbodley: i think that we only need the read side `thread_pipe[0]` to be non-blocking, can you try without the second call to `fcntl()`?

src/rgw/rgw_http_client.cc
```diff
@@ -790,6 +790,18 @@ int RGWHTTPManager::set_threaded()
     ldout(cct, 0) << "ERROR: pipe() returned errno=" << r << dendl;
     return r;
   }
+  r=fcntl(thread_pipe[0],F_SETFL,O_NONBLOCK);
+  if(r<0){
+  	r=-errno;
+  	ldout(cct, 0) << "ERROR: fcntl() returned errno=" << r << dendl;
+  	return r;
+  }
```
cbodley: formatting:
* two spaces for indentation
* spaces between operators and commas:
```c++
  r = fcntl(thread_pipe[0], F_SETFL, O_NONBLOCK);
  if (r < 0) {
    r = -errno;
```

cbodley: on error, need to clean up the fds from `pipe()`:
```c++
    TEMP_FAILURE_RETRY(::close(thread_pipe[0]));
    TEMP_FAILURE_RETRY(::close(thread_pipe[1]));
```

src/rgw/rgw_http_client.cc
cbodley: since we're in non-blocking mode, we could try to read more than 4 bytes as a potential optimization to avoid later wakeups. something like this?
```c++
     std::array<char, 256> buf;
     ret = read(signal_fd, buf.data(), buf.size());
```

cbodley:
@cethikdata ping

cbodley:
@cethikdata ping. if you're busy, i can pick this up

cbodley:
closing in favor of https://github.com/ceph/ceph/pull/10998. thanks for your contribution!

PR #10468, opened 2016-07-28
  cmake: reconstruct the cmake files and extract one common module

tone-zhang:
The PR comes from cmake: fix the build on AArch64 #10427 because #10427 is polluted by myself.
I'm sorry for the inconvenience. And thanks a lot for your patience.

Extract the CheckCompileTargets.cmake module out of
src/CMakeLists.txt. It is a used to generate some the compiler options
according to the CPU type.
Update the CMakeLists.txt files to invoke the common module

Signed-off-by: tone.zhang <tone.zhang@linaro.org>

### DIFF
--- a/cmake/modules/CheckCompileTargets.cmake
+++ b/cmake/modules/CheckCompileTargets.cmake
@@ -1,0 +1,77 @@
+# CMake module to get the compile options to match the target 
+#
+# If ARM NEON instruments set is supported, the following flags are set:
+# ARM_NEON
+# ARM_NEON2
+# If ARM CRC instrument set is supported, the flag HAVE_ARMV8_CRC is set,
+# and the related compile options are set
+# If INTEL SSE instruments set are supported, the related flags and 
+# compile options are set
+
+set(sse_srcs "${CMAKE_BINARY_DIR}/src/test/erasure-code/tmp_sse.c")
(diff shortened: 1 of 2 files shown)


wjwithagen:
@tone-zhang 
LGTM

wjwithagen:
Next step would be to exclude the ARM part if running on amd64, because one way or another ARM_NEON ends up being defined on FreeBSD. Which results in tries to build code that is not compilable.

tone-zhang:
@wjwithagen Thanks for you review.

tone-zhang:
@tchaikov Kefu, the PR #10427 is closed when I forced push the "cherry-pick"ed commit to tone-zhang/tone_cmake1 branch. I need more git practice to avoid such silly error. So could you please move to the new PR and have a review? Thanks!
Sorry for the inconvenience.

tone-zhang:
@dmick @cbodley @alimaredia Hi all, I have changed the PR #10427 to this one. Could you please have a review? Thanks a lot!

tchaikov:
@tone-zhang sorry for the latency, i found that we need to pass `-DINTEL*` to get the SIMD instruction generated for the gf-complete .

#10956 is posted to address this. as it is based on your work, so i added your Signed-off in the commit message.

PR #10469, opened 2016-07-28
src/CMakeLists.txt: supress more warnings

wjwithagen:
- We already have -Wall
   So these options without no- are redundant.
   and overloaded-virtual generates one warning for every line
   when compiling with Clang.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -113,9 +113,9 @@ else(no_yasm)
 endif(no_yasm)
 
 set(CMAKE_CXX_FLAGS "${CMAKE_C_FLAGS} -ftemplate-depth-1024 -Wno-invalid-offsetof")
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wnon-virtual-dtor -Woverloaded-virtual")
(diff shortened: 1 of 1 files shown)


wjwithagen:
please retest this

wjwithagen:
@tchaikov 
This especially pollutes the output of compiling `config_opts.h` with clang.

tchaikov:
@wjwithagen could you pastebin (some) the output? i just want to make sure we are not wallpapering interesting warnings.

wjwithagen:
@tchaikov 
Reran the compile without the flag, and this time no more warnings are reported.
(In the mean time, I upgraded, so it could be the newer version of Clang)

But then the question still is there: Why is this flag added if we are using -Wall

tchaikov:
> But then the question still is there: Why is this flag added if we are using -Wall

`-Woverloaded-virtual` is not enabled by `-Wall`, see
https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html

but clang's `-Wall` does enable this option, see https://github.com/llvm-mirror/clang/blob/master/include/clang/Basic/DiagnosticGroups.td. probably, we can enable this option only for GCC.

wjwithagen:
@tchaikov 
Arrghh, never occurred to me to even compare differences in options.
And naively expected all to mean ALL  
We already have conditioned options for GCC, so best move it there?

PR #10472, opened 2016-07-28
cmake: exclude non-public symbols in shared libraries

tchaikov:

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -518,7 +518,7 @@ endif(HAVE_ARMV8_CRC)
 
 add_library(common_utf8 STATIC common/utf8.c)
 
-target_link_libraries(common json_spirit common_utf8 erasure_code rt uuid resolv ${CRYPTO_LIBS} ${Boost_LIBRARIES} ${BLKID_LIBRARIES} ${Backtrace_LIBRARIES})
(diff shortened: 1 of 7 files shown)


batrick:
This makes a huge reduction in the exported symbols:

````
$ objdump -T librados.so > librados.so.2850b6f
...
$ wc -l lib*so*
   2082 libcephfs.so.2850b6f
  10744 libcephfs.so.2850b6f^
   2009 librados.so.2850b6f
  10401 librados.so.2850b6f^
````

And the final set of global symbols look about right to me. (There could be more pruning done but it's probably not worth it.)

Great job!

dillaman:
@tchaikov Any chance you care to fix librbd while you are at it? It used to have all the private symbols hidden and all the public API methods are already properly tagged. Must have been dropped during the conversion to cmake.

tchaikov:
@dillaman ahh, i completely missed librbd! will do. thanks for the reminder.

tchaikov:
@dillaman fixed in https://github.com/ceph/ceph/pull/10489/commits/f3cba6411a1435b825d002a529add5f9c2f7c2a3

dillaman:
@tchaikov Turns out "ceph_test_librbd" and "ceph_test_rbd_mirror" no longer function after 0edfad0 (from git bisect).  I'll fix in a new PR.

tchaikov:
@dillaman i have the same finding at http://tracker.ceph.com/issues/16882.

PR #10473, opened 2016-07-28
rgw: need to close_section in lc list op

weiqiaomiao:

### DIFF
--- a/src/rgw/rgw_admin.cc
+++ b/src/rgw/rgw_admin.cc
@@ -4744,10 +4744,13 @@ next:
   }
 
   if (opt_cmd == OPT_LC_LIST) {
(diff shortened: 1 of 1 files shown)


weiqiaomiao:
retest this please

src/rgw/rgw_admin.cc
```diff
@@ -4744,10 +4744,13 @@ int main(int argc, char **argv)
   }
 
   if (opt_cmd == OPT_LC_LIST) {
-    formatter->open_array_section("life cycle progress");
+    formatter->open_array_section("lifecycle list");
```
liewegas: FWIW this should also have an underscore or else it'll produce invalid XML

liewegas:
see #12232

PR #10474, opened 2016-07-28
mon:  remove the redudant jugement in LogMonitor tick function

songbaisen:
mon: remove the redudant jugement in LogMonitor tick function


dachary:
@jecluis looks redundant to me :-) Unless is_leader() has a desirable side effect ?

PR #10475, opened 2016-07-28
rgw/rgw_main: fix unnecessary variables defined

yonghengdexin735:

### DIFF
--- a/src/rgw/rgw_main.cc
+++ b/src/rgw/rgw_main.cc
@@ -78,9 +78,8 @@ void signal_shutdown()
     int val = 0;
     int ret = write(signal_fd[0], (char *)&val, sizeof(val));
     if (ret < 0) {
-      int err = -errno;
       derr << "ERROR: " << __func__ << ": write() returned "
-           << cpp_strerror(-err) << dendl;
+           << cpp_strerror(errno) << dendl;
     }
   }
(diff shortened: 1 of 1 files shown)


src/rgw/rgw_main.cc
```diff
@@ -78,9 +78,9 @@ void signal_shutdown()
     int val = 0;
     int ret = write(signal_fd[0], (char *)&val, sizeof(val));
     if (ret < 0) {
-      int err = -errno;
+      ret = -errno;
       derr << "ERROR: " << __func__ << ": write() returned "
-           << cpp_strerror(-err) << dendl;
+           << cpp_strerror(ret) << dendl;
```
theanalyst: actually there is no need ot change the sign of ret here,  cpp_strerror already has an internal check for this purpose, so just call cpp_strerror(errno);

yonghengdexin735:
@theanalyst  I think here defined ret for return and then defined err for 'errno' is unnecessary. 
First, the 'errno'  assign to ret (ret = errno). socond, remove 'int err = -errno' directly, just call cpp_strerror(errno). here I use the latter.

yonghengdexin735:
@theanalyst  done it

theanalyst:
lgtm

PR #10480, opened 2016-07-28
Clean up lifecycle thread

dang:

### DIFF
--- a/src/rgw/rgw_rados.cc
+++ b/src/rgw/rgw_rados.cc
@@ -3176,6 +3176,12 @@ void RGWRados::finalize()
   delete gc;
   gc = NULL;
 
+  if (use_lc_thread) {
+    lc->stop_processor();
+  }
+  delete lc;
+  lc = NULL;
+
   delete obj_expirer;
   obj_expirer = NULL;
 
--- a/src/rgw/rgw_lc.cc
+++ b/src/rgw/rgw_lc.cc
@@ -474,6 +474,7 @@ void RGWLC::start_processor()
 
 void RGWLC::stop_processor()
 {
+  down_flag.set(1);
   if (worker) {
     worker->stop();
     worker->join();
@@ -490,7 +491,7 @@ void RGWLC::LCWorker::stop()
 
 bool RGWLC::going_down()
 {
-  return false;
+  return (down_flag.read() != 0);
 }
 
 bool RGWLC::LCWorker::should_work(utime_t& now)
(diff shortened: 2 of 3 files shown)


yehudasa:
lgtm

cbodley:
still a lot of failures in http://pulpito.ceph.com/cbodley-2016-07-28_15:39:31-rgw-lifecycle-cleanup---basic-mira/. hard to tell which are actually related

testing does show an issue, though - `RGWLC::LCWorker::stop()` just signals the thread, but doesn't actually tell it to shut down. so the logs show this:
```
2016-07-29 10:00:11.931628 7fa040cfa700 -1 received  signal: Terminated from  PID: 1899 task name: -bash  UID: 1000
2016-07-29 10:00:11.931648 7fa040cfa700  1 handle_sigterm
2016-07-29 10:00:11.931670 7fa040cfa700  1 handle_sigterm set alarm for 120
2016-07-29 10:00:11.931684 7fa07079aac0 -1 shutting down
...
2016-07-29 10:00:12.141148 7fa0424fd700  5 life cycle: start
...
2016-07-29 10:00:12.322323 7fa0424fd700  5 life cycle: stop
2016-07-29 10:00:12.322346 7fa0424fd700  5 shedule life cycle next start time: Sat Jul 30 00:00:00 2016
```
looks like this part was unfinished:
```c++
bool RGWLC::going_down()
{
  return false;
}
```

src/rgw/rgw_rados.cc
```diff
@@ -3176,6 +3176,12 @@ void RGWRados::finalize()
   delete gc;
   gc = NULL;
 
+  if (use_lc_thread) {
+    lc->stop_processor();
+  }
```
cbodley: `~RGWLC()` is already calling `stop_processor()`, so we shouldn't need this part

cbodley:
@dang i pushed a fix for this to https://github.com/cbodley/ceph/commits/pr/10480, feel free to cherry-pick https://github.com/cbodley/ceph/commit/08140f6f1b4e4da0e38462decb3aa8b89d836fd5

src/rgw/rgw_rados.cc
dang: This is the same way that GC is, so maybe leave it?

cbodley: k

dang:
@cbodley Better fix, I think, taken from GC.

PR #10481, opened 2016-07-28
librbd: request exclusive lock if current owner cannot execute op

dillaman:
The initial krbd implementation will not support executing maintenance
ops and instead will return -EOPNOTSUPP. In this case, librbd can take
the lock and execute the operation.

Fixes: http://tracker.ceph.com/issues/16171

### DIFF
--- a/src/librbd/Operations.cc
+++ b/src/librbd/Operations.cc
@@ -111,6 +111,7 @@ struct C_InvokeAsyncRequest : public Context {
   boost::function<void(Context*)> remote;
   std::set<int> filter_error_codes;
   Context *on_finish;
+  bool request_lock = false;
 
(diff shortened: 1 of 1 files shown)


dillaman:
@idryomov Here is the patch for testing

idryomov:
@dillaman Saw you update the backport field on the tracker.  I think it would be best to wait until krbd patches are finalized and discuss the older clients in the meantime.

dillaman:
@idryomov This commit actually cannot be applied to Hammer -- it was more about ensuring it isn't just forgotten about once you merge this PR.

idryomov:
@dillaman and I decided that introducing a new feature bit is overkill.

This functionality will be exercised by the krbd suite.

smithfarm:
@dillaman So I can remove the hammer backport issue?

PR #10485, opened 2016-07-28
client: include COMPLETE and ORDERED states in cache dump

gregsfortytwo:


gregsfortytwo:
I noticed this when looking at what information we could get out of cache dump, because somebody reported a problem with file listing, and it's one of the most useful bits of debugging info when looking at those issues. :/

jcsp:
:+1:

jcsp:
Tested here: http://pulpito.ceph.com/jspray-2016-08-02_12:58:37-fs-wip-jcsp-testing-20160802-distro-basic-mira/

PR #10486, opened 2016-07-29
tests: Add test for global static non-POD segfault

badone:
libcommon.a is linked statically to librados and libcephfs. When the global
destructors for these libraries are called malloc code can report "double free
or corrpution" as the same resource may be deleted twice (global static
non-POD variable types). This test attempts to detect this issue.

### DIFF
--- a/src/test/common/CMakeLists.txt
+++ b/src/test/common/CMakeLists.txt
@@ -189,3 +189,12 @@ add_executable(unittest_shunique_lock
   )
 add_ceph_unittest(unittest_shunique_lock ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/unittest_shunique_lock)
 target_link_libraries(unittest_shunique_lock global ${BLKID_LIBRARIES} ${EXTRALIBS})
+
+# unittest_global_doublefree
+if(WITH_CEPHFS)
+  add_executable(unittest_global_doublefree
+    test_global_doublefree.cc
+    )
+  add_ceph_unittest(unittest_global_doublefree ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/unittest_global_doublefree)
+  target_link_libraries(unittest_global_doublefree cephfs librados)
+endif(WITH_CEPHFS)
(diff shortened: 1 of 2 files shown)


tchaikov:
@badone have you tried to reproduce the issues described by http://tracker.ceph.com/issues/16504 and http://tracker.ceph.com/issues/16686 by reverting their fixes?

badone:
This PR is based on 1b1c86f.

$ git log --oneline -2
62e8a97 tests: Add test for global static non-POD segfault
1b1c86f Merge pull request #10355 from yehudasa/wip-16742

If I check out that commit I can indeed reproduce.

Start   4: cephtool-test-mds.sh
*** Error in `/usr/bin/python2': double free or corruption (fasttop): 0x00005612e08c3cd0 ***

I can also reproduce it if I build my test on that commit.

$ bin/unittest_global_doublefree
Segmentation fault (core dumped)

Is that what you mean? 

I'm not sure I ever reproduced http://tracker.ceph.com/issues/16686 since the
steps to reproduce are not particularly clear to me nor is the commit where I
should be able to reproduce. If you can give me some hints I'd be happy to
try?

tchaikov:
lgtm.

src/test/common/CMakeLists.txt
```diff
@@ -189,3 +189,12 @@ add_executable(unittest_shunique_lock
...
 add_ceph_unittest(unittest_shunique_lock ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/unittest_shunique_lock)
 target_link_libraries(unittest_shunique_lock global ${BLKID_LIBRARIES} ${EXTRALIBS})
+
+# unittest_global_doublefree
+if(WITH_CEPHFS)
+  add_executable(unittest_global_doublefree
+    test_global_doublefree.cc
+    )
+  add_ceph_unittest(unittest_global_doublefree ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/unittest_global_doublefree)
+  target_link_libraries(unittest_global_doublefree global cephfs librados)
```
tchaikov: nit, i don't think we need to link against libglobal.

badone: @tchaikov Fixed and tested again

PR #10487, opened 2016-07-29
rbd: introduce v2 format for rbd export/import.

yangdongsheng:
Fixes:  http://tracker.ceph.com/issues/13186


yangdongsheng:
@dillaman 
This PR added this feature http://tracker.ceph.com/issues/13186 , please help to take a look. thanx.

yangdongsheng:
@dillaman I found there is still some bug in this PR, so it's not ready for code-review.

yangdongsheng:
@dillaman it passed my local manual testing now, please help to take a look. thanx

src/tools/rbd/action/MergeDiff.cc
```diff
@@ -196,7 +196,7 @@ static int do_merge_diff(const char *first, const char *second,
     }
   }
 
-  //We just handle the case like 'banner, [ftag], [ttag], stag, [wztag]*,etag',
+  //We just handle the case like 'banner, [ftag], [ttag], [stag], [wztag]*, [etag]',
```
dillaman: I think this was an attempt to say that the "s" and "e" tags are not optional and the ones in the brackets are optional.

src/tools/rbd/Utils.cc
```diff
@@ -762,5 +762,93 @@ std::string timestr(time_t t) {
   return buf;
 }
 
+int export_diff(librbd::Image& image, const char *fromsnapname,
```
dillaman: I'd rather see a new file for the main export/import diff logic instead of putting it into Utils (for lack of a better place).

src/tools/rbd/action/Import.cc
```diff
@@ -188,19 +188,22 @@ static int do_import(librbd::RBD &rbd, librados::IoCtx& io_ctx,
...
-      goto done;
+      goto err;
     }
   }
 
   r = image.close();
 
+err:
+  if (r < 0)
+    rbd.remove(io_ctx, imgname);
```
dillaman: I think the image would still be open in the goto case -- so this remove would fail.

src/tools/rbd/Utils.cc
```diff
@@ -877,6 +871,27 @@ int import_diff(librbd::Image &image, const char *path,
       std::cerr << "rbd: error opening " << path << std::endl;
       return r;
     }
+  }
+
+  r = import_diff_fd(image, fd, no_progress);
+
+  close(fd);
```
dillaman: Don't close stdin (not that it really matters)

src/tools/rbd/action/Export.cc
```diff
@@ -86,12 +86,13 @@ class C_Export : public Context
   SimpleThrottle &m_throttle;
   librbd::Image &m_image;
   bufferlist m_bufferlist;
+  uint64_t f_offset;
```
dillaman: The "m_" prefix is for instance variables -- if you want to distinguish this as the destination offset, name it "m_dest_offset"

src/tools/rbd/action/Export.cc
```diff
@@ -116,34 +117,94 @@ static int do_export(librbd::Image& image, const char *path, bool no_progress)
...
+
+      r = bl.write_fd(fd);
+      if (r < 0) {
+	return r;
+      }
+
+      const char *last_snap = NULL;
+      for (size_t i = 0; i < snaps.size(); ++i) {
+	utils::snap_set(image, snaps[i].name.c_str());
+	r = utils::export_diff_fd(image, last_snap, snaps[i].name.c_str(), false, fd, true);
```
dillaman: I'd rather see ```export_diff_fd``` take a ProgressContext -- in this case, you would pass a derived ProgressContext that subdivides by the number of snapshots.

dillaman:
@yangdongsheng I'd recommend combining Export/ExportDiff and Import/ImportDiff into the same file now that they are going to be interdependent.

dillaman:
@yangdongsheng It might be cleaner to just 100% re-use the existing export/import diff format for the export/import with snaps.  Just create a new "v2" header to distinguish the new behavior that it might contain >1 snapshot segment. Then on import(diff), if it's a v2, it should keep looping when it gets to the end snapshot marker, then the next field should be a from snapshot marker, data, etc, end snapshot, etc.

src/tools/rbd/action/Import.cc
```diff
@@ -132,6 +138,30 @@ static int do_import(librbd::RBD &rbd, librados::IoCtx& io_ctx,
...
+    image_buf[utils::RBD_IMAGE_BANNER.size()] = '\0';
+    if (strcmp(image_buf, utils::RBD_IMAGE_BANNER.c_str())) {
+      // Old format
+      if (from_stdin) {
+	// --with-snaps specified for an old format exported file
+	r = -EINVAL;
+	goto done;
+      }
+      // seek back to the beginning.
+      lseek(fd, 0, SEEK_SET);
```
dillaman: Couldn't you just set "import_snaps" to false here and eliminate the need for new_format?

src/tools/rbd/action/MergeDiff.cc
yangdongsheng: right. :+1:

src/tools/rbd/Utils.cc
yangdongsheng: I need to refer these functions in Export/Import and ExportDiff/ImportDiff. Then I put them in Utils. a new file? Such as src/tools/rbd/Diff.cc?

src/tools/rbd/action/Import.cc
yangdongsheng: will close it first.

src/tools/rbd/Utils.cc
yangdongsheng: okey

src/tools/rbd/action/Export.cc
yangdongsheng: agreed

src/tools/rbd/action/Export.cc
yangdongsheng: There is already a parameter of no_progress in this function and  this progress is subdivides by the number of extent in the diff from src_snap to dest_snap. And when we export-diff, it's more important how many extent different than how many snapshots between the two snapshots.

src/tools/rbd/action/Import.cc
yangdongsheng: Sorry, I am not sure what's your suggestion here, but I can explain the logic here more clear.

(1).import_snaps == true means --with-snaps is specified.
(2).(from_stdin == false )means importing something from file.

Then there are 3 possibility we will into this if statement.
(1). (import_snaps == true) and (from_stdin == true), in this case, when we found it's old format, because it's from stdin, we can't seek back, so return -EINVAL. That means, you have to make sure the you are accepting a new format importing stdin when you specify --with-snaps manually. 

(2). (import_snaps == true) and (from_stdin == false), in this case, when we found it's a old format. we will lseek to the beginning, because it's from a fd. and the all data in file should be imported into image.

(3). (import_snaps == false) and (from_stdin == false), in this case, --with-snaps are not specified, and we are importing data from a file. then we try to check the banner at first, if we found it's old format. lseek to 0 to import it in old manner. 

In this way, we can keep the backforwad compatability for the old format of export/exportdiff

src/tools/rbd/action/Export.cc
dillaman: Understood -- but my rationale is that if I have 10 snapshots on a 100G image, instead of having the progress bar smoothly progress, it will jump as each snapshot is completed. It would be nice to have the granular feedback.

src/tools/rbd/action/Import.cc
```diff
@@ -132,6 +138,30 @@ static int do_import(librbd::RBD &rbd, librados::IoCtx& io_ctx,
 #endif
   }
 
+  if (import_snaps || !from_stdin) {
+    r = safe_read_exact(fd, image_buf, utils::RBD_IMAGE_BANNER.size());
+    if (r < 0)
```
dillaman: Not that anyone should have a very tiny image, but this would return -EDOM if the legacy format and import size < the image banner length.

src/tools/rbd/action/Import.cc
```diff
@@ -132,6 +138,30 @@ static int do_import(librbd::RBD &rbd, librados::IoCtx& io_ctx,
...
+      // Old format
+      if (from_stdin) {
+	// --with-snaps specified for an old format exported file
+	r = -EINVAL;
+	goto done;
+      }
+      // seek back to the beginning.
+      lseek(fd, 0, SEEK_SET);
+    } else {
+      r = safe_read_exact(fd, &size, 8);
```
dillaman: I'd hate to automatically switch to the new format based on the header -- I'd rather see a warning message displayed saying something along the lines of "image appears to contain snapshots, import with --with-snaps".

src/tools/rbd/action/Import.cc
```diff
@@ -132,6 +138,30 @@ static int do_import(librbd::RBD &rbd, librados::IoCtx& io_ctx,
 #endif
   }
 
+  if (import_snaps || !from_stdin) {
+    r = safe_read_exact(fd, image_buf, utils::RBD_IMAGE_BANNER.size());
+    if (r < 0)
+      goto done;
+    image_buf[utils::RBD_IMAGE_BANNER.size()] = '\0';
+    if (strcmp(image_buf, utils::RBD_IMAGE_BANNER.c_str())) {
+      // Old format
```
dillaman: I'd prefer to see an error if "--with-snaps" specified on a legacy image.

src/tools/rbd/action/Import.cc
yangdongsheng: Aha, yes, good catch, actually I noticed and fixed this problem in my local tree yesterday. :) thanx.

src/tools/rbd/action/Import.cc
yangdongsheng: Okey, sound much safer. after some consideration, I agree with you.

src/tools/rbd/action/Import.cc
yangdongsheng: okey, similar with above comment, agree with you. We can make it working in a much simpler way.

yangdongsheng:
Hi @dillaman sorry for the late update. Please help to take a look again. thanx

yangdongsheng:
@dillaman do you have a time to take a look at this PR?

dillaman:
@yangdongsheng Sorry -- the comments were sitting in "pending" state for days since I forgot to publish them.

qa/workunits/rbd/import_export.sh
```diff
@@ -57,12 +57,21 @@ rm ${TMPDIR}/img2 ${TMPDIR}/img3
...
 rm ${TMPDIR}/img ${TMPDIR}/img2 ${TMPDIR}/img3
 
+# try with --with-snaps
+rbd snap create testimg@snap
+rbd export --with-snaps testimg ${TMPDIR}/img
+rbd import --with-snaps $RBD_CREATE_ARGS ${TMPDIR}/img testimg_import
+rbd info testimg_import
+rbd info testimg_import@snap
+rm ${TMPDIR}/img
+rbd snap rm testimg_import@snap
```
yangdongsheng: Okey

src/librbd/ExclusiveLock.cc
```diff
@@ -711,8 +711,11 @@ void ExclusiveLock<I>::handle_shutdown(int r) {
 
   m_image_ctx.aio_work_queue->clear_require_lock_on_read();
   m_image_ctx.aio_work_queue->unblock_writes();
-  m_image_ctx.image_watcher->flush(util::create_context_callback<
-    ExclusiveLock<I>, &ExclusiveLock<I>::complete_shutdown>(this));
+  if (m_image_ctx.image_watcher == nullptr)
```
yangdongsheng: I got a segmentation fault without this change in "rbd export --with-snaps"

src/tools/rbd/action/Export.cc
```diff
@@ -154,6 +213,8 @@ void get_arguments(po::options_description *positional,
   at::add_path_options(positional, options,
                        "export file (or '-' for stdout)");
   at::add_no_progress_option(options);
+  options->add_options()
+    ("with-snaps", po::bool_switch(), "export snapshots at the same time");
```
yangdongsheng: okey

src/tools/rbd/Utils.h
```diff
@@ -8,14 +8,24 @@
...
 #include "tools/rbd/ArgumentTypes.h"
+#include "common/Throttle.h"
 #include <string>
 #include <boost/program_options.hpp>
 
 namespace rbd {
 namespace utils {
 
+static const std::string RBD_IMAGE_BANNER ("rbd image v1\n");
+static const std::string RBD_IMAGE_SNAPS_BANNER ("rbd image snaps v1\n");
```
yangdongsheng: agreed

src/tools/rbd/action/Export.cc
```diff
@@ -116,34 +117,92 @@ static int do_export(librbd::Image& image, const char *path, bool no_progress)
...
     if (r >= 0) {
-      r = ftruncate(fd, info.size);
+      r = ftruncate(fd, file_size);
+      r = lseek(fd, file_size, SEEK_SET);
     }
-    close(fd);
   }
 
-  if (r < 0) {
+  if (with_snaps) {
```
yangdongsheng: sounds good.

src/tools/rbd/action/Import.cc
```diff
@@ -141,75 +171,110 @@ static int do_import(librbd::RBD &rbd, librados::IoCtx& io_ctx,
...
-    if (!from_stdin)
-      pc.update_progress(image_pos, size);
-
-    bufferlist bl(blklen);
-    bl.append(p, blklen);
-    // resize output image by binary expansion as we go for stdin
-    if (from_stdin && (image_pos + (size_t)blklen) > size) {
-      size *= 2;
-      r = image.resize(size);
+    lseek(fd, size, SEEK_CUR);
```
yangdongsheng: Because this will be used when we import a image without --with-snaps. Sometimes, maybe we only want to import a full image even the snaps were exported already.

src/tools/rbd/action/Export.cc
yangdongsheng: But then we can't import a image without snaps in this case.
```
rbd export --with-snaps test test.export
rbd import test.import test.export
```
I mean, sometimes, we can get a exported image with snapshots, but we only want to import a full image from it.

yangdongsheng: So I name it as "with snaps" rather than "snaps mode".

yangdongsheng:
@dillaman When I came back to refine this pr, I found maybe we should introduce a --format rather than --with-snaps. The v2 exporting can export not only content of image, but also snapshots, features, and some other priorities if necessary.  When we are going to import an image with --format v2, we should check the BANNER at first, if fail, exit with an EINVAL. 

Then there would be a big change on this branch, please wait for another review. Thanx.

yangdongsheng:
@dillaman PR updated. Please help to take a look. changelog as below.
(1) squash ExportDiff.cc into Export.cc, same to Import.
(2) instead of --with-snaps, I introduce --import/export-format. The format v2 will export/import snapshots, image_order, image_format, features and so on.
(3) Currently only implement image_order and image_format. will continue for other priorities later.

dillaman:
@yangdongsheng sorry for the delay, I will try to review this tomorrow. In the meantime, can you give your "fixup" commit a better commit title/message (or rebase and place the changes in the originating commit).

src/librbd/ExclusiveLock.cc
```diff
@@ -714,8 +714,11 @@ void ExclusiveLock<I>::handle_shutdown(int r) {
   }
 
   m_image_ctx.aio_work_queue->unblock_writes();
-  m_image_ctx.image_watcher->flush(util::create_context_callback<
-    ExclusiveLock<I>, &ExclusiveLock<I>::complete_shutdown>(this));
+  if (m_image_ctx.image_watcher == nullptr)
```
dillaman: What is this change? If you don't have an image watcher, you shouldn't have an ExclusiveLock either.

yangdongsheng:
@dillaman No problem, I will squash the fixup commit into the previous ones. And figure out why I need the wather related change. Please wait for tomorrow. Thanx a lot.

src/librbd/ExclusiveLock.cc
yangdongsheng: @dillaman I found there are lots of assumption in ExclusiveLock about the image_watcher is not null. seems like the problem is in snap_set(). TBH, I am not clear about the relationship between exclusive_lock and watcher, is that 100% there is at list one watcher if we get an exclusive_lock? Could you help to explain it a bit. thanx a lot.

src/librbd/image/SetSnapRequest.cc
```diff
@@ -38,10 +38,17 @@ SetSnapRequest<I>::~SetSnapRequest() {
 
 template <typename I>
 void SetSnapRequest<I>::send() {
-  if (m_snap_name.empty()) {
-    send_init_exclusive_lock();
+  if (m_image_ctx.read_only) {
```
dillaman: Tracking this fix under PR #11559 so it can be properly backported to jewel

src/tools/rbd/action/Import.cc
```diff
@@ -205,13 +205,11 @@ static int do_import(librbd::RBD &rbd, librados::IoCtx& io_ctx,
...
-      pc.fail();
-    else
-      pc.finish();
-    close(fd);
-  }
+  if (r < 0)
+    pc.fail();
+  else
+    pc.finish();
+  close(fd);
```
dillaman: Need ```if (!from_stdin)``` clause around the ```close``` method

src/tools/rbd/action/Export.cc
```diff
@@ -17,10 +17,6 @@
 
 namespace rbd {
 namespace action {
-namespace export_diff {
```
dillaman: Instead of leaking these methods into the ```action``` namespace, I would combine the ```export_diff``` and ```export_full``` namespaces (and rename methods where needed). This comment applies to the next commit for the ```import_diff``` namespace removal as well.

doc/man/8/rbd.rst
```diff
@@ -236,6 +236,9 @@ Commands
   The --stripe-unit and --stripe-count arguments are optional, but must be
   used together.
 
+  The --import-format accepts '1' or '2' currently. Format 2 allow us to import not only the content
+  of image, but also the snapshots and other priorities, such as image_order, features.
```
dillaman: ```priorities``` -> ```properties```

src/tools/rbd/action/Import.cc
```diff
@@ -21,15 +21,16 @@ namespace rbd {
...
+		   bool no_progress, int format)
 {
   int r;
   struct stat stat_buf;
   utils::ProgressContext pc("Importing image diff", no_progress);
   uint64_t size = 0;
   uint64_t off = 0;
   string from, to;
-  char buf[utils::RBD_DIFF_BANNER.size() + 1];
+  string banner = (format == 1 ? utils::RBD_DIFF_BANNER:utils::RBD_DIFF_BANNER_V2);
```
dillaman: Minor: need space around ternary operator's ":"

src/tools/rbd/Utils.h
```diff
@@ -30,6 +30,8 @@ static const std::string RBD_DIFF_BANNER_V2 ("rbd diff v2\n");
 #define RBD_EXPORT_IMAGE_ORDER		'O'
 #define RBD_EXPORT_IMAGE_FORMAT		'F'
 #define RBD_EXPORT_IMAGE_FEATURES	'T'
+#define RBD_EXPORT_IMAGE_STRIPEUNIT	'U'
```
dillaman: Minor: "_" between words (_STRIPE_UNIT and _STRIPE_COUNT)

doc/man/8/rbd.rst
```diff
@@ -224,10 +224,12 @@ Commands
...
   snapshots, this fails and nothing is deleted.
 
-:command:`export` (*image-spec* | *snap-spec*) [*dest-path*]
+:command:`export` [--export-format *format (1 or 2)*] (*image-spec* | *snap-spec*) [*dest-path*]
   Exports image to dest path (use - for stdout).
+  The --export-format accepts '1' or '2' currently. Format 2 allow us to export not only the content
+  of image, but also the snapshots and other priorities, such as image_order, features.
 
-:command:`import` [--image-format *format-id*] [--object-size *size-in-B/K/M*] [--stripe-unit *size-in-B/K/M* --stripe-count *num*] [--image-feature *feature-name*]... [--image-shared] *src-path* [*image-spec*]
+:command:`import` [--import-format *format (1 or 2)*] [--image-format *format-id*] [--object-size *size-in-B/K/M*] [--stripe-unit *size-in-B/K/M* --stripe-count *num*] [--image-feature *feature-name*]... [--image-shared] *src-path* [*image-spec*]
```
dillaman: Minor: it would be nice to use the same option name for both import/export -- perhaps just "--export-format"

src/tools/rbd/action/Import.cc
```diff
@@ -65,8 +291,15 @@ class C_Import : public Context {
 
 static int do_import(librbd::RBD &rbd, librados::IoCtx& io_ctx,
                      const char *imgname, const char *path,
-		     librbd::ImageOptions& opts, bool no_progress)
+		     librbd::ImageOptions& opts, bool no_progress,
+		     int import_format)
 {
+  // check current supported formats.
+  if (import_format != 1 && import_format != 2) {
```
dillaman: You can use a custom validator for this. See ArgumentTypes.cc ```validate``` for ```ImageFormat```

src/tools/rbd/action/Export.cc
```diff
@@ -116,34 +386,141 @@ static int do_export(librbd::Image& image, const char *path, bool no_progress)
...
+    uint64_t snap_num = snaps.size() + 1;
+    ::encode(snap_num, bl);
+
+    r = bl.write_fd(fd);
+    if (r < 0) {
+      goto out;
+    }
+
+    if (0 == snap_num) {
+      goto out;
```
dillaman: If an image doesn't have any snapshots it won't export?

src/tools/rbd/action/Export.cc
```diff
@@ -116,34 +386,141 @@ static int do_export(librbd::Image& image, const char *path, bool no_progress)
...
 
   utils::ProgressContext pc("Exporting image", no_progress);
-
   SimpleThrottle throttle(max_concurrent_ops, false);
   uint64_t period = image.get_stripe_count() * (1ull << info.order);
-  for (uint64_t offset = 0; offset < info.size; offset += period) {
-    if (throttle.pending_error()) {
-      break;
+
+  if (export_format == 2) {
```
dillaman: Minor: since format 1 and 2 are so different, it would be cleaner to separate the implementation bodies into a v1 and v2 export method

dillaman:
https://github.com/ceph/ceph/blob/master/doc/dev/rbd-diff.rst should be updated to document the new format.

dillaman:
For each of the new tags within the new v2 export format, I would suggest encoding them as follows:

(1 byte): tag
(4 bytes): tag data length
(variable): tag data

That will allow future versions to add new capabilities and allow older rbd CLI tools to import the images (but skip over the tags they don't recognize).

src/tools/rbd/action/Import.cc
yangdongsheng: yes, thanx

src/tools/rbd/action/Import.cc
yangdongsheng: Thanx good suggestion

doc/man/8/rbd.rst
yangdongsheng: thanx

doc/man/8/rbd.rst
yangdongsheng: okey,

yangdongsheng:
Hi @dillaman ,Sorry for the late, just came back from Barcelona. This branch is updated. Please help to take a look. 

Change log:
(1) comments addressed.
(2) rebased against master.
(3) docs added.

Thanx

dillaman:
Thanks -- I'll take a look this week. As an FYI, we are finalizing the Kraken release feature set next week, so this will miss the release and will be scheduled for the Luminous release instead.

yangdongsheng:
That's fine, Luminous sounds okey to me . thanx for your information.

src/tools/rbd/action/Export.cc
```diff
@@ -86,12 +365,159 @@ class C_Export : public Context
...
+  ::encode(tag, bl);
+
+  // write bl to fd.
+  r = bl.write_fd(fd);
+  if (r < 0) {
+    return r;
+  }
+  offset += bl.length();
+
+  lseek(fd, offset, SEEK_SET);
```
trociny: Why is lseek needed? I suppose fd position is set correctly at this point? Also, we might write to stdout...

src/tools/rbd/action/Import.cc
```diff
@@ -63,9 +295,219 @@ class C_Import : public Context {
...
+	r = safe_read_exact(fd, &stripe_count, 8);
+	if (r < 0) {
+	  return r;
+	}
+	if (opts.get(RBD_IMAGE_OPTION_STRIPE_COUNT, &stripe_count) != 0) {
+	  opts.set(RBD_IMAGE_OPTION_STRIPE_COUNT, stripe_count);
+	}
+      } else {
+	std::cerr << "rbd: invalid tag in image properties zone: " << tag << "Skip it." << std::endl;
+	::lseek(fd, length, SEEK_CUR);
```
trociny: Doesn't it fail if it is imported from stdin?

Also, the format specifies le64, but you don't do le64 to native platform conversion (I don't have idea if we have a different platform though).

trociny:
BTW, I suppose we will want export/import for consistency groups when they are functional. It looks like the format is extensible to support this, still it would be nice to keep this in mind.

src/tools/rbd/action/Export.cc
yangdongsheng: Oh, yes, that's a code used in last version. It should be removed now. thanx

src/tools/rbd/action/Import.cc
yangdongsheng: Yes, will check the fd at first. 

About the le and be problem, there is a #define WRITE_INTTYPE_ENCODER(type, etype)  in ceph to declare the encode() and decode for each time of int. So I think just ceph can work well on different platform just with "encode(uint64_t, bufferlist)".

yangdongsheng:
@trociny thanx for your review. I will consider export/import cg later after this merged. thanx

src/tools/rbd/action/Import.cc
```diff
@@ -387,6 +387,9 @@ static int do_import_header(int fd, int import_format, uint64_t &size, librbd::I
 	}
       } else {
 	std::cerr << "rbd: invalid tag in image properties zone: " << tag << "Skip it." << std::endl;
+	if (fd == 0) {
+	  return -EINVAL;
+	}
 	::lseek(fd, length, SEEK_CUR);
```
trociny: @yangdongsheng You can use `read` instead of `lseek` for stdin. Also it is better to use a constant `STDIN_FILENO` instead of 0. And you need to check `lseek` return value (probably falling back to `read` if `errno` is set to `ESPIPE` and failing for any other error).

src/tools/rbd/action/Import.cc
trociny: Yes, but then you need to used decode too, i.e. read to a buffer using bufferlist::read_fd and call decode.

src/tools/rbd/action/Import.cc
yangdongsheng: right thanx

src/tools/rbd/action/Import.cc
yangdongsheng: yes thanx

dillaman:
@yangdongsheng Can you please rebase? I will review this again ASAP if you think it's ready.

yangdongsheng:
Jason, Sorry I am in a business trip now, will update my patches next 
week. thanx

yangdongsheng:
@dillaman rebased. please help to take a look. thanx

src/tools/rbd/action/Export.cc
```diff
@@ -86,12 +365,156 @@ class C_Export : public Context
...
+		        uint64_t period, int max_concurrent_ops, utils::ProgressContext &pc)
+{
+  int r = 0;
+  // header
+  bufferlist bl;
+  bl.append(utils::RBD_IMAGE_BANNER_V2);
+
+  // size in header
+  uint64_t size = info.size;
+  ::encode(size, bl);
```
dillaman: Perhaps this should be omitted and instead encoded in the "snap" section. If I have an image w/ snapshots, the first snapshot could be size X, the second could be size Y, and the HEAD revision could be size Z. When processing the first snapshot, you would need to resize the image down unnecessarily. Therefore, instead, just create a zero-sized image until you get to the first diff section which will tell you the current image size.

src/tools/rbd/action/Export.cc
```diff
@@ -86,12 +365,156 @@ class C_Export : public Context
...
+  uint64_t length;
+  // encode order
+  tag = RBD_EXPORT_IMAGE_ORDER;
+  length = 8;
+  ::encode(tag, bl);
+  ::encode(length, bl);
+  ::encode(uint64_t(info.order), bl);
+
+  // encode format
+  tag = RBD_EXPORT_IMAGE_FORMAT;
```
dillaman: We are trying to eliminate v1 images (already deprecated) -- so I think this is safe to leave out since it will soon provide zero value.

src/tools/rbd/Utils.h
```diff
@@ -16,6 +16,24 @@ namespace utils {
 
 static const std::string RBD_DIFF_BANNER ("rbd diff v1\n");
 
+static const std::string RBD_IMAGE_BANNER_V2 ("rbd image v2\n");
+static const std::string RBD_IMAGE_SNAPS_BANNER_V2 ("rbd image snaps v2\n");
```
dillaman: Nit: instead of "snaps", perhaps just "diffs" since this delineates the start of the diff section (and included the HEAD revision which is technically not a snapshot).

src/tools/rbd/action/Export.cc
```diff
@@ -86,12 +365,156 @@ class C_Export : public Context
...
+  bl.append(utils::RBD_IMAGE_SNAPS_BANNER_V2);
+
+  std::vector<librbd::snap_info_t> snaps;
+  r = image.snap_list(snaps);
+  if (r < 0) {
+    return r;
+  }
+
+  uint64_t diff_num = snaps.size() + 1;
+  ::encode(diff_num, bl);
```
dillaman: I don't think this was documented in the data format document.

doc/dev/rbd-diff.rst
```diff
@@ -67,3 +67,80 @@ End
...
+Every record has a one byte "tag" that identifies the record type,
+followed by length of data, and then some other data.
+
+Metadata records come in the first part of the image.  Order is not
+important, as long as all the metadata records come before the data
+records.
+
+In v2, we have the following metadata in each section:
+(1 Bytes) tag.
+(8 Bytes) length.
```
dillaman: I think it's safe to use a 32bit length -- the only thing that requires 64bit is the write/zero tag, but since the maximum object size is 32M, they technically only require 32bit lengths as well (and should be updated to reflect that).

doc/dev/rbd-export.rst
```diff
@@ -0,0 +1,79 @@
+RBD Export & Import
+===================
+
+This is a file format of an RBD image or snapshot. It's a sparse format
+for the full image. There are three recording sections in the file.
+
+(1) Header.
+(2) Metadatas.
```
yangdongsheng: okey thanx

doc/dev/rbd-diff.rst
yangdongsheng: Yes, but what about keeping 8Bytes here in case of we want to break the limit of 32M someday, and 4Bytes for each tag sounds not a big waste for an output file of a rbd export-diff.

src/tools/rbd/action/Export.cc
yangdongsheng: This is needed for V1 import. When we do a v1 import, there is no diffs, and we need to know the size of the whole image.

doc/dev/rbd-diff.rst
```diff
@@ -67,3 +67,80 @@ End
 ---
 
 - u8: 'e'
+
```
yangdongsheng: I would like to make every tag as simple as possible. Then I think we can introduce a new tags for snap_timestamp and CRC in diffs. This way, we can also keep the backward compatibility, we only need to set the timestamp when we found a tag for it.

src/tools/rbd/action/Export.cc
yangdongsheng: yes, but V1 images are still alive. I am afraid it's bad for users if this change prevents users to export an v1 image.

src/tools/rbd/action/Export.cc
yangdongsheng: thanx done.

src/tools/rbd/Utils.h
yangdongsheng: thanx, agreed.

yangdongsheng:
@dillaman What do you think about this version?

src/tools/rbd/action/Export.cc
dillaman: Why would you do a v1 import on a v2 export? I think that should be disallowed.

src/tools/rbd/action/Export.cc
dillaman: Sorry, I wasn't implying that the users shouldn't be able to export v1 images -- I am saying that we should always import them as v2 images regardless if the original was a v1 image format.

src/tools/rbd/action/Import.cc
```diff
@@ -447,6 +511,8 @@ void get_arguments(po::options_description *positional,
   at::add_image_spec_options(positional, options, at::ARGUMENT_MODIFIER_DEST);
   at::add_create_image_options(options, true);
   at::add_no_progress_option(options);
+  options->add_options()
+    ("import-format", po::value<int>(), "format of the file to be imported");
```
dillaman: For consistency sake, I would still refer to this as the "export-format" since that is the format on-disk that you are attempting to import.

yangdongsheng: Yes, this name was changed in the later commit: 2ad0c6ca2f9e0a2a26a91a922b2ec6d9a0dcc190

src/tools/rbd/action/Export.cc
yangdongsheng: This makes sense to me. thanx

src/tools/rbd/action/Export.cc
yangdongsheng: what about this case:
```
rbd export --export-format 1 test test.export
rbd import --export-format 1 test.export test_import
```

dillaman: Wouldn't that be handled by the ```do_export_v1``` method instead of the ```do_export_v2``` method?

yangdongsheng:
Oh，that's right. thanx

yangdongsheng:
Hi @dillaman is this ready for functional testing?

dillaman:
@yangdongsheng Failure in the "import_export.sh" test case [1] when the default features are overridden / using a data pool.

[1] http://qa-proxy.ceph.com/teuthology/jdillaman-2017-02-16_14:16:19-rbd-wip-jd-testing-distro-basic-smithi/823267/teuthology.log

yangdongsheng:
@dillaman  Yes, the features for data-pool would break this case. Then I changed this unittest as below:
```
--- a/qa/workunits/rbd/import_export.sh
+++ b/qa/workunits/rbd/import_export.sh
@@ -114,12 +114,13 @@ rbd remove testimg
 
 # features
 rbd import --image-feature layering ${TMPDIR}/img testimg
+FEATURES_BEFORE=`rbd info testimg|grep features`
 rbd export --export-format 2 testimg ${TMPDIR}/img_v2
 rbd import --export-format 2 ${TMPDIR}/img_v2 testimg_import
-# only one feature
-rbd info testimg_import|grep "features"|wc -w|grep 2
-# and the feature is layering
-rbd info testimg_import|grep "features"|awk '{print $2}'|grep "layering"
+FEATURES_AFTER=`rbd info testimg_import|grep features`
+if [ "$FEATURES_BEFORE" != "$FEATURES_AFTER" ]; then
+  false
+fi

```

yangdongsheng:
@dillaman Since the timezone problem, can I re-queue the functional testing on your branch wip-jd-testing? Maybe type "retest" on this PR?

tchaikov:
the workunit change breaks the jewel-x-singleton test, see http://pulpito.ceph.com/kchai-2017-03-30_14:32:23-rados-wip-13999-kefu---basic-mira/966061/

```
2017-03-30T14:41:21.411 INFO:tasks.workunit.client.0.mira071.stderr:+ rbd export --export-format 2 testimg /tmp/rbd_import_export_13580/img_v2
2017-03-30T14:41:21.434 INFO:tasks.workunit.client.0.mira071.stderr:rbd: unrecognised option '--export-format'
```

dillaman:
@tchaikov Please open a tracker ticket -- this PR is already merged

tchaikov:
@dillaman sure, will do tomorrow. just wanted to give you guys a quick heads-up in case you run into this.

dillaman:
@tchaikov It's a teuthology issue where the test case specifies that the "jewel" branch of the workunit should be used, but instead it downloads the "head" version (d463f52054ba9a166dfb330c009489e305071dfd) of the workunit.

PR #10488, opened 2016-07-29
rbd-mirror: add additional test scenarios

lande1234:
Signed-off-by: lande1234 <lan.de3@zte.com.cn>


lande1234:
@dillaman    hi
i add this because:
1、we shuld test start ${POOL} ${CLUSTER2} instead of start (the start is just test before)
2、we shuld test restart ${POOL} ${CLUSTER2} after stop the ${POOL} ${CLUSTER2}
3、we shuld check the result after every operation

so the test will be more  effective. thanks!

PR #10492, opened 2016-07-29
os/bluestore: remove cmake warning from extent alloc functions

chhabaramesh:
Removing cmake related warning from alloc extents functions

### DIFF
--- a/src/os/bluestore/BitAllocator.h
+++ b/src/os/bluestore/BitAllocator.h
@@ -194,10 +194,6 @@ public:
   static int get_level(int64_t total_blocks);
   static int64_t get_level_factor(int level);
   virtual bool is_allocated(int64_t start_block, int64_t num_blocks) = 0;
-  virtual bool is_allocated(ExtentList *blocks, int64_t num_blocks, int blk_off) {
-    debug_assert(0);
-    return true;
-  }
   virtual bool is_exhausted() = 0;
   virtual bool child_check_n_lock(BitMapArea *child, int64_t required) {
       debug_assert(0);
@@ -364,6 +360,7 @@ public:
   }
 
   int64_t alloc_blocks(int64_t num_blocks, int64_t *start_block);
+  using BitMapArea::alloc_blocks_dis;
   int64_t alloc_blocks_dis(int64_t num_blocks,
         int64_t blk_off, ExtentList *block_list);  
   void set_blocks_used(int64_t start_block, int64_t num_blocks);
--- a/src/os/bluestore/Allocator.h
+++ b/src/os/bluestore/Allocator.h
@@ -43,9 +43,8 @@ public:
                             uint64_t max_alloc_size, int64_t hint,
                             std::vector<AllocExtent> *extents, int *count) = 0;
 
-  virtual int alloc_extents(uint64_t want_size, uint64_t alloc_unit,
-                            int64_t hint, std::vector<AllocExtent> *extents,
-			    int *count) {
+  int alloc_extents(uint64_t want_size, uint64_t alloc_unit,
+                    int64_t hint, std::vector<AllocExtent> *extents, int *count) {
     return alloc_extents(want_size, alloc_unit, want_size, hint, extents, count);
   }
 


xiexingguo:
LGTM

chhabaramesh:
@xiexingguo  thanks for quick review :)
@ifed01 please merge if looks ok to you.

src/os/bluestore/BitAllocator.h
```diff
@@ -322,6 +322,8 @@ class BitMapZone: public BitMapArea{
   static int64_t total_blocks;
   static void incr_count() { count++;}
   static int64_t get_total_blocks() {return total_blocks;}
+
+  using BitMapArea::is_allocated;
```
ifed01: IMHO one can simply remove 
virtual bool is_allocated(ExtentList *blocks, int64_t num_blocks, int blk_off)
to fix the warning. Looks like it's not used..

src/os/bluestore/BitAllocator.h
```diff
@@ -384,6 +387,8 @@ class BitMapAreaIN: public BitMapArea{
   std::mutex m_blocks_lock;
   BitMapAreaList *m_child_list;
 
+
+  using BitMapArea::is_allocated;
```
ifed01: IMHO not needed if second BitMapArea::is_allocated is removed

src/os/bluestore/BitAllocator.h
```diff
@@ -364,6 +366,7 @@ class BitMapZone: public BitMapArea{
   }
 
   int64_t alloc_blocks(int64_t num_blocks, int64_t *start_block);
+  using BitMapArea::alloc_blocks_dis;
```
ifed01: I have two remarks regarding alloc_blocks_dis method(s):
The first one - I suggest to avoid overload virtuals at all. Hence one can rename alloc_blocks_dis( bool wait...) to something like alloc_blocks_dis_wait..
The second - Probably I missed something but I'm unable to find any use of this bool wait parameter in alloc_blocks[_dis[_int]] methods. I can see they are passed between these methods but don't see real wait var handling. Could you please point out that to me. And if my finding is correct probably one can simplify these methods and get rid of some of them...

chhabaramesh: Wait flag is not used currently as it is for a mode where we want to search for blocks without jumping a zone if another thread is already allocating in that zone.  So there are three modes a. ) where we keep searching and jump zones that are busy, b.) wait for the busy zone on the way to search, but still multiple threads can be active. c.) everything goes serial. I am sure this will be useful when we implement hint or at least we should not remove it till we are more clear on its usage.

src/os/bluestore/BitAllocator.h
chhabaramesh: removed the function

src/os/bluestore/BitMapAllocator.h
```diff
@@ -42,6 +42,7 @@ class BitMapAllocator : public Allocator {
     uint64_t want_size, uint64_t alloc_unit, int64_t hint,
     uint64_t *offset, uint32_t *length);
 
+  using Allocator::alloc_extents;
```
ifed01: The root cause for the warning is double virtual declaration for alloc_extents methods in Allocator.h
In fact there is no much sence in declaring following method virtual
 virtual int alloc_extents(uint64_t want_size, uint64_t alloc_unit,
                            int64_t hint, std::vector<AllocExtent> *extents,
                            int *count) {
    return alloc_extents(want_size, alloc_unit, want_size, hint, extents, count);
  }

To be honest I don't see much sense in having it at all - IMHO it rather complicates Allocator interface,
Anyway if for some reason you prefer to have it than removing virtual declaration ( or renaming the method ) should effectively eliminate the warning.

cbodley:
@chhabaramesh thanks for looking at this. i recently created a tracker issue for it at http://tracker.ceph.com/issues/16766. if this makes those warnings go away, can you please add this to your commit message?
`Fixes: http://tracker.ceph.com/issues/16766`

src/os/bluestore/BitMapAllocator.h
chhabaramesh: yes, it does not need to be virtual. hence removing virtual. Having it makes sense as it eliminates need of one argument.

chhabaramesh:
@cbodley  sure, I will add tracker number in commit message.

chhabaramesh:
@ifed01 please merge this.

src/os/bluestore/BitAllocator.h
ifed01: this 'using' statement isn't needed any more

chhabaramesh:
@ifed01  anything pending on my side to merge this?

ifed01:
@chhabaramesh please remove additional 'using BitMapArea::is_allocated;'

PR #10494, opened 2016-07-29
osd:  add is_split check before _start_split

songbaisen:
osd: add is_split check before _start_split

If the pg is not split,we do not need to start pg split.


dachary:

dachary:

dachary:
jenkins test this please (asok again)

dachary:
please update the commit message explaining why introducing the if is necessary and why it won't work if it is not there

songbaisen:
@dachary  Thank you! Done it.

dachary:
contrary to what the comment suggests, is_split is not just a predicate, it also populates its last argument (children) which is used unconditionally by start_split

PR #10497, opened 2016-07-29
jewel: partprobe intermittent issues during ceph-disk prepare

Abhishekvrshny:
http://tracker.ceph.com/issues/16586


dachary:
It passed the rados suite http://tracker.ceph.com/issues/16344#note-17 (except for one known bug that is, I believe, unrelated).

PR #10505, opened 2016-07-29
src/common/dns_resolve.cc: reorder the includes

wjwithagen:
- dns_resolv.h should only be include after all the std-includes are
   made. Otherwise it will result into undefined structs/variables on
   FreeBSD/Clang

### DIFF
--- a/src/common/dns_resolve.cc
+++ b/src/common/dns_resolve.cc
@@ -11,13 +11,14 @@
  * Foundation.  See file COPYING.
  *
  */
-#include "dns_resolve.h"
 
 #include <sys/types.h>
 #include <netinet/in.h>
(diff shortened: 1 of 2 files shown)


cbodley:
can we try to fix dns_resolve.h instead, so we don't have to worry about other source files that include it? the header itself should include all of its dependencies

wjwithagen:
@cbodley 
That is a more invasive fix, but makes perfect sense.

wjwithagen:
please retest this

xiexingguo:
@wjwithagen 
retest this please

shall work:-)

wjwithagen:
@xiexingguo 
I was wondering.... Thanx.

PR #10511, opened 2016-07-29
src/ceph.in: allow ceph to run from CEPH_ROOT/src in Cmake env

wjwithagen:
- Extent the special case for developers to find the python stuff
  if cwd = CEPH_ROOT/src and everything is build with Cmake.
  So rados.so is somewhere in build/lib/cython_modules

### DIFF
--- a/src/ceph.in
+++ b/src/ceph.in
@@ -84,33 +84,54 @@ if MYDIR.endswith('src') and \
     if 'PATH' in os.environ and MYDIR not in os.environ['PATH']:
         os.environ['PATH'] += ':' + MYDIR
 
-elif os.path.exists(os.path.join(os.getcwd(), "CMakeCache.txt")) \
-     and os.path.exists(os.path.join(os.getcwd(), "bin/init-ceph")):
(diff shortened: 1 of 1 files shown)


wjwithagen:
@jcsp 
Could you take a look at the extra lookup code I added to make ceph work in more cases during testing.

jcsp:
I'm not clear on why this is needed -- what's wrong from just always running from the build directory?

wjwithagen:
@jcsp 
I normally test my programs from src, and if I want to test with ctest I'll go:
```
(cd build; ctest -V -R testprog)
```
Now if you then want to use `ceph` to monitor any of the MON/OSDs it becomes quite a hassle,
where as just typing `ceph .....` makes life a lot easier.
But it is a convenience, and as such only of use to developers.

jcsp:
I feel like I'm missing something, what's the motivation for testing your programs from src instead of operating in your built directory?

Anyway, I don't have particularly strong opinions about this so maybe ask around.

PR #10512, opened 2016-07-30
mds/MDBalancer: cleanup

renhwztetecs:

### DIFF
--- a/src/common/config_opts.h
+++ b/src/common/config_opts.h
@@ -475,7 +475,7 @@ OPTION(mds_log_events_per_segment, OPT_INT, 1024)
 OPTION(mds_log_segment_size, OPT_INT, 0)  // segment size for mds log, default to default file_layout_t
 OPTION(mds_log_max_segments, OPT_U32, 30)
 OPTION(mds_log_max_expiring, OPT_INT, 20)
-OPTION(mds_bal_sample_interval, OPT_FLOAT, 3.0)  // every 5 seconds
+OPTION(mds_bal_sample_interval, OPT_DOUBLE, 3.0)  // every 3 seconds
 OPTION(mds_bal_replicate_threshold, OPT_FLOAT, 8000)
 OPTION(mds_bal_unreplicate_threshold, OPT_FLOAT, 0)
 OPTION(mds_bal_frag, OPT_BOOL, false)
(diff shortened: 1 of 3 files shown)


src/common/config_opts.h
```diff
@@ -475,7 +475,7 @@ OPTION(mds_log_events_per_segment, OPT_INT, 1024)
 OPTION(mds_log_segment_size, OPT_INT, 0)  // segment size for mds log, default to default file_layout_t
 OPTION(mds_log_max_segments, OPT_U32, 30)
 OPTION(mds_log_max_expiring, OPT_INT, 20)
-OPTION(mds_bal_sample_interval, OPT_FLOAT, 3.0)  // every 5 seconds
+OPTION(mds_bal_sample_interval, OPT_DOUBLE, 3.0)  // every 5 seconds
```
xiexingguo: Can you fix the comments also?

renhwztetecs: yes, thanks!

src/mds/MDBalancer.cc
```diff
@@ -162,7 +162,10 @@ mds_load_t MDBalancer::get_load(utime_t now)
   ifstream cpu("/proc/loadavg");
   if (cpu.is_open())
     cpu >> load.cpu_load_avg;
-
+  else
+    derr << "input file '/proc/loadavg' not found" << dendl;
+  cpu.close();
```
jcsp: The close() would go inside the is_open() conditional, calling close on a non-open ifstream is an error.

src/mds/MDBalancer.cc
```diff
@@ -337,6 +340,7 @@ void MDBalancer::queue_split(CDir *dir)
 
 void MDBalancer::queue_merge(CDir *dir)
 {
+  assert(mds->mdsmap->allows_dirfrags());
```
jcsp: This is incorrect, allows_dirfrags controls whether splitting is permitted.  If it is switched off again after some splits have happened, merges are allowed even though allow_dirfrags is false

src/mds/MDBalancer.cc
renhwztetecs: I change it :)
the destructor will release it, do not need to display the call “cpu.close()”

src/mds/MDBalancer.cc
renhwztetecs: this scene I did not take into account,
thanks for you reviews

renhwztetecs:
done :)

PR #10514, opened 2016-07-30
src/vstart.sh: stop script aborting when ceph.cfg is not there

wjwithagen:
When vstart.sh starts with a fresh directory, then there is no ceph.cfg
and the script aborts because it runs with `bash -e`
So better check to see if file is there before trying to delete is.
Otherwise you never get any ceph-{mon,osd,mds} running


batrick:
Thanks!

wjwithagen:
@batrick 
Thanx for swift response.

PR #10516, opened 2016-07-31
src/global/pidfile.cc: Assign elements in structures individually

wjwithagen:
- Assigning structure is better per element.
   The order of elements is not always the same in very OS.
   So "hardcoding" can/will assign to the wrong elements

### DIFF
--- a/src/global/pidfile.cc
+++ b/src/global/pidfile.cc
@@ -162,7 +162,12 @@ int pidfh::open(const md_config_t *conf)
   pf_dev = st.st_dev;
   pf_ino = st.st_ino;
 
-  struct flock l = { F_WRLCK, SEEK_SET, 0, 0, 0 };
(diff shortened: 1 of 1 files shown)


tchaikov:
lgtm

wjwithagen:
@tchaikov 
Can we check this in as well?

tchaikov:
@wjwithagen i'd like to run this change through a rados qa run, but i didn't get a chance to schedule the tests. will do later this week.

wjwithagen:
@tchaikov 
Sure, no problem.
The change is is small and should be oke.
But better safe than sorry

tchaikov:
tested http://pulpito.ceph.com/kchai-2016-08-18_15:35:36-rados-wip-kefu-testing2---basic-mira/

all failures are known ones.

PR #10517, opened 2016-07-31
cmake: FreeBSD specific excludes in CMakeLists.txt files

wjwithagen:
Changes to the CMakeLists.txt files specific for FreeBSD building.
These are the items that can not (yet) be build on FreeBSD
Excluding build target based on
- RBD
- KVS
- BLKID
- LIBFUSE
- AIO
- RADOSGW

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -605,11 +605,20 @@ if(${WITH_RADOSGW})
     cls_rgw_client
   )
 endif(${WITH_RADOSGW})
-if(${WITH_RBD})
+if(WITH_RBD)
   set(DENCODER_EXTRALIBS
       ${DENCODER_EXTRALIBS}
       rbd_replay_types)
-endif(${WITH_RBD})
+  if(LINUX)
+    list(APPEND dencoder_srcs
+      $<TARGET_OBJECTS:krbd_objs>
+      $<TARGET_OBJECTS:parse_secret_objs>)
+    set(DENCODER_EXTRALIBS
+      ${DENCODER_EXTRALIBS}
+      keyutils
+      udev)
+  endif(LINUX)
+endif(WITH_RBD)
 
 add_executable(ceph-dencoder ${dencoder_srcs})
 target_link_libraries(ceph-dencoder
@@ -801,7 +810,7 @@ if(WITH_LIBCEPHFS)
   install(TARGETS ceph-syn DESTINATION bin)
   install(TARGETS mount.ceph DESTINATION ${CMAKE_INSTALL_SBINDIR})
 
-  if(HAVE_LIBFUSE)
+  if(WITH_FUSE)
     set(ceph_fuse_srcs
       ceph_fuse.cc
       client/fuse_ll.cc)
@@ -810,7 +819,7 @@ if(WITH_LIBCEPHFS)
     set_target_properties(ceph-fuse PROPERTIES COMPILE_FLAGS "-I${FUSE_INCLUDE_DIRS}")
     install(TARGETS ceph-fuse DESTINATION bin)
     install(PROGRAMS mount.fuse.ceph DESTINATION ${CMAKE_INSTALL_SBINDIR})
-  endif(HAVE_LIBFUSE)
+  endif(WITH_FUSE)
 endif(WITH_LIBCEPHFS)
 
 add_subdirectory(journal)
@@ -836,7 +845,7 @@ if(WITH_KVS)
   add_subdirectory(key_value_store)
 endif(WITH_KVS)
 
-if(${WITH_RADOSGW})
+if(WITH_RADOSGW)
   set(civetweb_common_files civetweb/src/civetweb.c)
   add_library(civetweb_common_objs OBJECT ${civetweb_common_files})
   target_include_directories(civetweb_common_objs PUBLIC
@@ -848,7 +857,7 @@ if(${WITH_RADOSGW})
       "${SSL_INCLUDE_DIR}")
   endif(HAVE_SSL)
   add_subdirectory(rgw)
-endif(${WITH_RADOSGW})
+endif(WITH_RADOSGW)
 
 install(FILES
   sample.ceph.conf
--- a/src/librbd/CMakeLists.txt
+++ b/src/librbd/CMakeLists.txt
@@ -79,10 +79,15 @@ endif()
 add_library(librbd ${CEPH_SHARED}
   $<TARGET_OBJECTS:osdc_rbd_objs>
   $<TARGET_OBJECTS:common_util_obj>
-  $<TARGET_OBJECTS:krbd_objs>
-  $<TARGET_OBJECTS:parse_secret_objs>  
   ${CMAKE_SOURCE_DIR}/src/common/ContextCompletion.cc
   librbd.cc)
+if(LINUX)
+  list(APPEND librbd 
+    $<TARGET_OBJECTS:parse_secret_objs>  
+    $<TARGET_OBJECTS:krbd_objs>
+    )
+endif(LINUX)
+
 target_link_libraries(librbd LINK_PRIVATE 
   rbd_internal
   rbd_types
--- a/src/os/CMakeLists.txt
+++ b/src/os/CMakeLists.txt
@@ -29,7 +29,6 @@ set(libos_srcs
 
 if(HAVE_LIBAIO)
   list(APPEND libos_srcs
-    bluestore/kv.cc
     bluestore/Allocator.cc
     bluestore/BitmapFreelistManager.cc
     bluestore/BlockDevice.cc
--- a/src/os/CMakeLists.txt
+++ b/src/os/CMakeLists.txt
@@ -21,30 +21,36 @@ set(libos_srcs
   filestore/WBThrottle.cc
   filestore/ZFSFileStoreBackend.cc
   memstore/MemStore.cc
+  kstore/kv.cc
   kstore/KStore.cc
   kstore/kstore_types.cc
-  bluestore/kv.cc
-  bluestore/Allocator.cc
-  bluestore/BitmapFreelistManager.cc
-  bluestore/BlockDevice.cc
-  bluestore/BlueFS.cc
-  bluestore/bluefs_types.cc
-  bluestore/BlueRocksEnv.cc
-  bluestore/BlueStore.cc
-  bluestore/bluestore_types.cc
-  bluestore/ExtentFreelistManager.cc
-  bluestore/FreelistManager.cc
-  bluestore/KernelDevice.cc
-  bluestore/StupidAllocator.cc
-  bluestore/BitMapAllocator.cc
-  bluestore/BitAllocator.cc
   fs/FS.cc
   ${libos_xfs_srcs})
+
+if(HAVE_LIBAIO)
+  list(APPEND libos_srcs
+    bluestore/kv.cc
+    bluestore/Allocator.cc
+    bluestore/BitmapFreelistManager.cc
+    bluestore/BlockDevice.cc
+    bluestore/BlueFS.cc
+    bluestore/bluefs_types.cc
+    bluestore/BlueRocksEnv.cc
+    bluestore/BlueStore.cc
+    bluestore/bluestore_types.cc
+    bluestore/ExtentFreelistManager.cc
+    bluestore/FreelistManager.cc
+    bluestore/KernelDevice.cc
+    bluestore/StupidAllocator.cc
+    bluestore/BitMapAllocator.cc
+    bluestore/BitAllocator.cc
+  )
+endif(HAVE_LIBAIO)
 
 if(WITH_FUSE)
   list(APPEND libos_srcs
     FuseStore.cc)
-endif()
+endif(WITH_FUSE)
 
 if(WITH_SPDK)
   list(APPEND libos_srcs
@@ -55,9 +61,9 @@ add_library(os STATIC ${libos_srcs} $<TARGET_OBJECTS:kv_objs>)
 
 if(HAVE_LIBAIO)
   target_link_libraries(os ${AIO_LIBRARIES})
-endif()
+endif(HAVE_LIBAIO)
 
-if(HAVE_LIBFUSE)
+if(WITH_FUSE)
   target_link_libraries(os ${FUSE_LIBRARIES})
 endif()
 
--- a/src/test/CMakeLists.txt
+++ b/src/test/CMakeLists.txt
@@ -8,7 +8,9 @@ add_subdirectory(cls_hello)
 add_subdirectory(cls_lock)
 add_subdirectory(cls_log)
 add_subdirectory(cls_numops)
-add_subdirectory(cls_rbd)
+if(WITH_RBD)
+  add_subdirectory(cls_rbd)
+endif(WITH_RBD)
 add_subdirectory(cls_refcount)
 add_subdirectory(cls_replica_log)
 add_subdirectory(cls_rgw)
@@ -26,7 +28,9 @@ add_subdirectory(libcephfs)
 add_subdirectory(librados)
 add_subdirectory(librados_test_stub)
 add_subdirectory(libradosstriper)
-add_subdirectory(librbd)
+if(WITH_RBD)
+  add_subdirectory(librbd)
+endif(WITH_RBD)
 add_subdirectory(messenger)
 add_subdirectory(mds)
 add_subdirectory(mon)
@@ -40,7 +44,9 @@ add_subdirectory(pybind)
 if(${WITH_RADOSGW})
   add_subdirectory(rgw)
 endif(${WITH_RADOSGW})
-add_subdirectory(rbd_mirror)
+if(WITH_RBD)
+  add_subdirectory(rbd_mirror)
+endif(WITH_RBD)
 add_subdirectory(system)
 
 # test_timers
@@ -79,14 +85,16 @@ target_link_libraries(test_crypto
 add_executable(test_build_libcommon buildtest_skeleton.cc)
 target_link_libraries(test_build_libcommon common pthread ${CRYPTO_LIBS} ${EXTRALIBS})
 
-if(${WITH_RADOSGW})
-add_executable(test_build_librgw buildtest_skeleton.cc)
-target_link_libraries(test_build_librgw rgw_a global pthread ${CRYPTO_LIBS} ${EXTRALIBS})
-endif(${WITH_RADOSGW})
+if(WITH_RADOSGW)
+  add_executable(test_build_librgw buildtest_skeleton.cc)
+  target_link_libraries(test_build_librgw rgw_a global pthread ${CRYPTO_LIBS} ${EXTRALIBS})
+endif(WITH_RADOSGW)
 
-# From src/test/Makefile-client.am: I dont get this one... testing the osdc build but link in libcephfs?
-add_executable(test_build_libcephfs buildtest_skeleton.cc)
-target_link_libraries(test_build_libcephfs cephfs pthread ${CRYPTO_LIBS} ${EXTRALIBS})
+if(WITH_LIBCEPHFS)
+  # From src/test/Makefile-client.am: I dont get this one... testing the osdc build but link in libcephfs?
+  add_executable(test_build_libcephfs buildtest_skeleton.cc)
+  target_link_libraries(test_build_libcephfs cephfs pthread ${CRYPTO_LIBS} ${EXTRALIBS})
+endif(WITH_LIBCEPHFS)
 
 add_executable(test_build_librados buildtest_skeleton.cc)
 target_link_libraries(test_build_librados librados pthread ${CRYPTO_LIBS} ${EXTRALIBS} osdc osd os common cls_lock_client ${BLKID_LIBRARIES})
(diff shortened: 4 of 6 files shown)


src/CMakeLists.txt
```diff
@@ -678,9 +683,9 @@ add_executable(ceph-osd ${ceph_osd_srcs}
   $<TARGET_OBJECTS:common_util_obj>)
 add_dependencies(ceph-osd erasure_code_plugins)
 target_link_libraries(ceph-osd osd os global ${BLKID_LIBRARIES})
-if(WITH_FUSE)
+if(HAVE_LIBFUSE)
```
tchaikov: i don't think we want this change. one can always disable fuse using `-DWITH_FUSE=OFF`.

src/CMakeLists.txt
```diff
@@ -587,15 +585,22 @@ if(${WITH_RADOSGW})
...
   )
 endif(${WITH_RADOSGW})
-if(${WITH_RBD})
+if(WITH_RBD)
   set(DENCODER_EXTRALIBS
       ${DENCODER_EXTRALIBS}
       rbd_replay_types)
-endif(${WITH_RBD})
+  if(LINUX)
+    list(APPEND dencoder_srcs
```
tchaikov: i understand that the freebsd build does not support krbd, but could you explain this in the commit message also?

src/CMakeLists.txt
```diff
@@ -587,15 +585,22 @@ if(${WITH_RADOSGW})
...
+        $<TARGET_OBJECTS:krbd_objs>)
+  endif(LINUX)
+endif(WITH_RBD)
+if(WITH_KVS)
+  list(APPEND dencoder_srcs
+      $<TARGET_OBJECTS:parse_secret_objs>)
```
tchaikov: could you explain the rationales in the commit message? imo, it's not self-explained.

src/test/librbd/CMakeLists.txt
```diff
@@ -1,3 +1,4 @@
+if(WITH_RBD)
```
tchaikov: you can conditionally 
```cmake
add_subdirectory(librbd)
```
in `src/test/CMakeLists.txt` instead.

tchaikov:
could you elaborate a little bit about "why" in the commit messages?

src/test/pybind/CMakeLists.txt
```diff
@@ -1,2 +1,4 @@
-add_ceph_test(test_ceph_daemon.py ${CMAKE_CURRENT_SOURCE_DIR}/test_ceph_daemon.py)
-add_ceph_test(test_ceph_argparse.py ${CMAKE_CURRENT_SOURCE_DIR}/test_ceph_argparse.py)
+if(NOT FREEBSD)
+  add_ceph_test(test_ceph_daemon.py ${CMAKE_CURRENT_SOURCE_DIR}/test_ceph_daemon.py)
+  add_ceph_test(test_ceph_argparse.py ${CMAKE_CURRENT_SOURCE_DIR}/test_ceph_argparse.py)
```
tchaikov: why this fails under FreeBSD? could you pastebin the output?

src/CMakeLists.txt
If you look at the code in ./CMakeLists.txt the FUSE testing actually purposly sets HAVE_LIBFUSE in the case that FUSE is actually present.
So IMHO it is better to actually use `HAVE_LIBFUSE`, because that guards the fact that we did not set `-DWITH_FUSE=OFF` but still do not have FUSE available.

Note: I agree that having both (and using) `WITH-*` and `HAVE_*` does not help.
Perhaps it should be that the `WITH_*` switches are for commandline-CMake, and should resulting in a matching `HAVE_*`

src/test/librbd/CMakeLists.txt
wjwithagen: @tchaikov 
It was actually there. 
Deleted this commit

src/test/pybind/CMakeLists.txt
wjwithagen: @tchaikov 
Got to dig that out again, but it has probably mostly to do with my inexperience (non existent) to setup nosetests...

wjwithagen: @tchaikov 
After fixing the bang-path in the srcipt to be `/usr/local/bin/nosetests` running it just generates:
```
[~/Ceph/work.patch/ceph] wjw@freetest.digiware.nl> ./src/test/pybind/test_ceph_daemon.py
Unmatched ".
Exit 1
```
And for the time being I'd rather chase after other problems.

src/CMakeLists.txt
tchaikov: `HAVE_LIBFUSE` is for the subvar in `acconfig.h`. we'd better use `WITH_*` for checking the options.

and also, i don't think the purpose of this PR is to consolidate these two variants. also you are not touching this piece of code. so could you please leave it as it is?

OKE

src/test/CMakeLists.txt
```diff
@@ -140,6 +145,10 @@ add_executable(ceph_kvstorebench
   ${kvstorebench_srcs}
   )
 target_link_libraries(ceph_kvstorebench librados global ${BLKID_LIBRARIES} ${CMAKE_DL_LIBS})
+install(TARGETS
```
tchaikov: please don't wrap line like this. if we have multiple targets, we might want to have 1 item for each line.

src/test/CMakeLists.txt
```diff
@@ -495,31 +503,47 @@ add_dependencies(tests
...
   ceph_snappy
   cls_lock
   ceph_test_objectstore
   ceph_erasure_code_non_regression
   ceph_erasure_code
   ceph-disk
-  ceph-detect-init
   cython_modules)
+if(WITH_RBD)
+  list(APPEND tests
```
tchaikov: please don't wrap line. thanks.

wjwithagen:
@tchaikov 
Ping?

src/test/CMakeLists.txt
```diff
@@ -40,7 +38,11 @@ add_subdirectory(pybind)
 if(${WITH_RADOSGW})
   add_subdirectory(rgw)
 endif(${WITH_RADOSGW})
-add_subdirectory(rbd_mirror)
+if(WITH_RBD)
```
tchaikov: maybe we should add the `if` test at where the `add_subdirectory()` was added. please note the order in which the subdirectories were added.

wjwithagen: @tchaikov 
I tried to include all under 1 if test.
And the order within the exclusion is maintained.

wjwithagen:
@tchaikov 
Reorganised the code back to individual excludes in `src/test/CmakeLists.txt` 
Will rebase with current master
But I'm hoping we can get this code in, so most of my Cmake changes for FreeBSD

src/CMakeLists.txt
tchaikov: > - The current KVS depends on keyutils
>    Which is an interesting concept, but still needs to be ported to
>    FreeBSD one way or another.

again keyutils is used by krbd and mount.ceph. we've discussed this before.

CMakeLists.txt
```diff
@@ -325,6 +325,9 @@ endif(WITH_XIO)
 
 #option for RGW
 option(WITH_RADOSGW "Rados Gateway is enabled" ON)
+if(WITH_RADOSGW)
```
tchaikov: could you drop this commit? this message does not help our user, imho.

src/CMakeLists.txt
```diff
@@ -860,6 +860,7 @@ if(WITH_RADOSGW)
     target_include_directories(civetweb_common_objs PUBLIC
       "${SSL_INCLUDE_DIR}")
   endif(HAVE_SSL)
+  message(STATUS "Building directory rgw")
```
tchaikov: and drop this commit, if you please.

CMakeLists.txt
wjwithagen: On 22-8-2016 14:22, Kefu Chai wrote:
> In CMakeLists.txt
> <https://github.com/ceph/ceph/pull/10517#discussion_r75666926>:
> 
>> @@ -325,6 +325,9 @@ endif(WITH_XIO)
>>  
>>  #option for RGW
>>  option(WITH_RADOSGW "Rados Gateway is enabled" ON)
>> +if(WITH_RADOSGW)
> 
> could you drop this commit? this message does not help our user, imho.

I know, this is going to be dropped.
But one way or another the Jenkins stuff refuses to add_directory(rgw).
Which it shows in the Jenkins output, it complains about missing
radosgw-admin.

Trying to figure out why in my CentOS build, but that all of a sudden
REQUIRED python3. So needed to fix that first.
And it is rather busy at the dayjob.

wjwithagen:
@tchaikov 
I don't  quite get this. In my CentOS 7.2 this PR builds without much problems, and includes all of the rgw stuff. 
Building it here in Jenkins on trusty generates a failure, it doesn't even build the required tests and tools

src/CMakeLists.txt
That was in another thread, and I tested it there to find out that KVS depends on secret.c which uses keyutils.h 
```
head -25 ./src/common/secret.c
// -*- mode:C++; tab-width:8; c-basic-offset:2; indent-tabs-mode:t -*-
// vim: ts=8 sw=2 smarttab
/*
 * Ceph - scalable distributed file system
 *
 * Copyright (C) 2011 New Dream Network
 *
 * This is free software; you can redistribute it and/or
 * modify it under the terms of the GNU Lesser General Public
 * License version 2.1, as published by the Free Software
 * Foundation.  See file COPYING.
 *
 */

#include <string.h>
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>
#include <errno.h>
#include <fcntl.h>
#include <keyutils.h>
#include <sys/types.h>
```

tchaikov:
```
/home/jenkins-build/build/workspace/ceph-pull-requests/src/test/run-rbd-unit-tests.sh: line 9: unittest_librbd: command not found
```

might want to guard `run-rbd-unit-tests.sh` with `if(WITH_RBD)` as well.

src/CMakeLists.txt
tchaikov: > I tested it there to find out that KVS depends on secret.c

i cannot find a trace, could you point me to it? thanks.

wjwithagen:
@tchaikov 
I think I found it.
dependencies are not lists(??) and thus list(APPEND tests .....) does not add to the tests-dependencies.
add_dependencies does work in my debug-pull.
Lets see if it works here as well.

wjwithagen:
@tchaikov 
Would this be good enough to pull?

tchaikov:
@wjwithagen could you please address https://github.com/ceph/ceph/pull/10517/commits/c35066eee9c3e6d122b0a77f147fc1529ae8efc1#r75801624 ?

src/CMakeLists.txt
```diff
@@ -687,7 +703,7 @@ add_dependencies(ceph-osd erasure_code_plugins)
 target_link_libraries(ceph-osd osd os global ${BLKID_LIBRARIES})
 if(WITH_FUSE)
   target_link_libraries(ceph-osd ${FUSE_LIBRARIES})
-endif()
+endif(WITH_FUSE)
```
tchaikov: if the `if` is small enough, i am inclined to not add the ending `WITH_FUSE`.

src/CMakeLists.txt
It is like you said: We discussed it before, I looked into it, and like above the include is obvious.
But it could very well be that we discussed this in private Email.

I think I've found ways to actually work around this, but that requires fixing a few source files.
So I'd rather do that after I'm finished with the Cmake stuff.

src/CMakeLists.txt
Will fix

src/CMakeLists.txt
tchaikov: > and like above the include is obvious.

the include means `secret.c` depends on keyutils, but what i don't understand is why KVS depends on `secret.c`.

I think it is like that because that is how it got out of the autobuild - Cmake conversion.
But you are correct that only krbd and mount.ceph reference secret.h.
I even checked calls to any of the 3 functions defined in secret.h.

However secret.c is added to the target: parse_secret_files
Which is used in:
```
./src/tools/rbd_nbd/CMakeLists.txt:2:  $<TARGET_OBJECTS:parse_secret_objs>)
./src/tools/rbd/CMakeLists.txt:41:  $<TARGET_OBJECTS:parse_secret_objs>
./src/rbd_replay/CMakeLists.txt:17:  $<TARGET_OBJECTS:parse_secret_objs>
./src/rbd_replay/CMakeLists.txt:32:    $<TARGET_OBJECTS:parse_secret_objs>
./src/CMakeLists.txt:574:  $<TARGET_OBJECTS:parse_secret_objs>
./src/CMakeLists.txt:757:set(parse_secret_files
./src/CMakeLists.txt:759:add_library(parse_secret_objs OBJECT ${parse_secret_files})
./src/CMakeLists.txt:790:    $<TARGET_OBJECTS:parse_secret_objs>
./src/test/librbd/CMakeLists.txt:55:  $<TARGET_OBJECTS:parse_secret_objs>
./src/test/librbd/CMakeLists.txt:88:  $<TARGET_OBJECTS:parse_secret_objs>
./src/test/librbd/CMakeLists.txt:114:  $<TARGET_OBJECTS:parse_secret_objs>
./src/test/librbd/CMakeLists.txt:131:  $<TARGET_OBJECTS:parse_secret_objs>
./src/test/bench/CMakeLists.txt:25:    $<TARGET_OBJECTS:parse_secret_objs>
./src/test/CMakeLists.txt:779:  $<TARGET_OBJECTS:parse_secret_objs>
./src/test/cls_rbd/CMakeLists.txt:5:  $<TARGET_OBJECTS:parse_secret_objs>
./src/test/rbd_mirror/CMakeLists.txt:81:  $<TARGET_OBJECTS:parse_secret_objs>)
./src/librbd/CMakeLists.txt:78:  $<TARGET_OBJECTS:parse_secret_objs>
```

So your suggestion is that most of those references could/should go?
And just about everything should be able to build without parse_secrets_obj.

But that would not be correct to do in this PR.
But I'll see if there is another way will work in a Linux build.

tchaikov: > So your suggestion is that most of those references could/should go?

@wjwithagen thanks for double checking this! i will give it a try tmr, if it works, probably it worths another PR =)

tchaikov:
#10837 is posted to address the cleanup we discussed yesterday.

src/CMakeLists.txt
```diff
@@ -605,11 +605,20 @@ if(${WITH_RADOSGW})
...
+  if(LINUX)
+    list(APPEND dencoder_srcs
+      $<TARGET_OBJECTS:krbd_objs>
+      $<TARGET_OBJECTS:parse_secret_objs>)
+    set(DENCODER_EXTRALIBS
+      ${DENCODER_EXTRALIBS}
+      keyutils
+      udev)
+  endif(LINUX)
+endif(WITH_RBD)
```
tchaikov: > - The current KVS depends on keyutils
>   Which is an interesting concept, but still needs to be ported to
>   FreeBSD one way or another.

could you remove this from your commit message? as it is not true, see #10837.

src/librbd/CMakeLists.txt
```diff
@@ -77,10 +77,19 @@ endif()
 add_library(librbd ${CEPH_SHARED}
   $<TARGET_OBJECTS:osdc_rbd_objs>
   $<TARGET_OBJECTS:common_util_obj>
-  $<TARGET_OBJECTS:krbd_objs>
-  $<TARGET_OBJECTS:parse_secret_objs>  
   ${CMAKE_SOURCE_DIR}/src/common/ContextCompletion.cc
   librbd.cc)
+if(WITH_KVS)
```
tchaikov: as we discussed, this is not correct. krbd depends on parse_secret_objs.

src/os/CMakeLists.txt
```diff
@@ -21,30 +21,36 @@ set(libos_srcs
...
+    bluestore/BitMapAllocator.cc
+    bluestore/BitAllocator.cc
+  )
+endif(HAVE_LIBAIO)
+
 if(WITH_FUSE)
   list(APPEND libos_srcs
     FuseStore.cc)
-endif()
+endif(WITH_FUSE)
```
tchaikov: could you remove this change for adding the ending `WITH_FUSE`?

src/test/CMakeLists.txt
```diff
@@ -104,8 +112,7 @@ target_link_libraries(ceph_bench_log global pthread rt ${BLKID_LIBRARIES} ${CMAK
 add_executable(ceph_test_mutate
   test_mutate.cc
   )
-target_link_libraries(ceph_test_mutate global librados ${BLKID_LIBRARIES}
-  ${CMAKE_DL_LIBS})
+target_link_libraries(ceph_test_mutate global librados ${BLKID_LIBRARIES} ${CMAKE_DL_LIBS})
```
tchaikov: any reason to change this line?

src/test/CMakeLists.txt
```diff
@@ -131,15 +138,16 @@ target_link_libraries(ceph_omapbench
...
-  ${kvstorebench_srcs}
-  )
-target_link_libraries(ceph_kvstorebench librados global ${BLKID_LIBRARIES} ${CMAKE_DL_LIBS})
+if(WITH_KVS)
+  # ceph_kvstorebench
+  set(kvstorebench_srcs
+    kv_store_bench.cc
+    ${CMAKE_SOURCE_DIR}/src/key_value_store/kv_flat_btree_async.cc
+    )
+  add_executable(ceph_kvstorebench ${kvstorebench_srcs} )
```
tchaikov: could remove the space before `)`.

src/test/pybind/CMakeLists.txt
tchaikov: @wjwithagen i'd recommend leave as it is instead disabling them on FreeBSD if we have not root-caused this issue. imho, a clean "make check" with all failed tests disabled is not what we are after. you can run the test using
```
cd src
PYTHONPATH=pybind /usr/local/bin/nosetests  --stop test/pybind/test_ceph_daemon.py
```

src/test/CMakeLists.txt
wjwithagen: @tchaikov 
Probably a leftover from this PR, and perhaps someboy has wrapped this line.
OR I tried to make to make it consistent with the lines around it.

I do have a some question marks about long line wrapping, but addressing that right now is not really in my focus

src/CMakeLists.txt
Did not delete it, but changed KVS to secret.c

src/librbd/CMakeLists.txt
wjwithagen: @tchaikov 
Made it LINUX dependant, like in src/CMakeLists.txt

src/test/pybind/CMakeLists.txt
wjwithagen: @tchaikov 
Oke, I drop this.

wjwithagen:
@tchaikov 
Jenkins seemed to have an vacation off yesterday.
But is matching all we discussed, I think?

wjwithagen:
@tchaikov 
Rebased and working again.

src/os/CMakeLists.txt
```diff
@@ -21,30 +21,36 @@ set(libos_srcs
   filestore/WBThrottle.cc
   filestore/ZFSFileStoreBackend.cc
   memstore/MemStore.cc
+  kstore/kv.cc
```
tchaikov: @wjwithagen is this change intentional?

src/os/CMakeLists.txt
tchaikov: @wjwithagen ping?

wjwithagen:
@tchaikov 
Not intentional, if anything, it stems from rebasing...
And I missed it during looking at the result.
Things are getting old in my memory, and lots of other merges have happened in the meanwhile.
Will check

src/os/CMakeLists.txt
wjwithagen: @tchaikov @liewegas
Right, this is the deal:
```
> diff -s src/os/bluestore/kv.cc src/os/kstore/kv.cc
Files src/os/bluestore/kv.cc and src/os/kstore/kv.cc are identical
```

So we should ditch one of the 2. And since there is nothing special bluestore with it, I suggest we delete 
`src/os/bluestore/kv.cc`
Unless Sage has a special meaning with kv.cc being copied to bluestore.

wjwithagen:
@tchaikov @liewegas 
I put the answer a bit out of place, but above I explain that both kv.cc's are equal.
So one should go.

wjwithagen:
@tchaikov @liewegas 
Selected kstore/kv.cc to be used (versus bluestore/kv.cc) 
If this is not appropriate, I could use the bluestore one.

wjwithagen:
@tchaikov @liewegas
Could you have a look at the proposed commit. It it is oke, I will rebase it in the other commits.

tchaikov:
i will push this changeset to gb, once the build passes, will merge it.

wjwithagen:
@tchaikov 
Thanx from merging this PR.
There are still some more CMake changes required to complete compiling on FreeBSD.
They are in #10973 
Part is a consequence of the discussion on secret.c.
Part is due to pyhton3 introduction

PR #10519, opened 2016-08-01
jewel: selinux denials in RGW

Abhishekvrshny:
http://tracker.ceph.com/issues/16312


dachary:

PR #10521, opened 2016-08-01
    os/filestore/FileStore.cc: filter out “meta” directory

YankunLi:
load_pg from current directory, but we don't need meta directory

under current

Signed-off-by: YankunLi <lioveni99@gmail.com>


YankunLi:
The check failure is not connected with this patch.

PR #10528, opened 2016-08-01
rgw: fix for object names ending with underscores

theanalyst:
currently if an object name ends with an underscore, we will end up
returning 404, as we are setting the rados namespace for an object to
the object name. So avoid setting a ns when there is no object name left
to set

Fixes: http://tracker.ceph.com/issues/16856

Reported-by: Martin Born <martin.born@udg.de>

### DIFF
--- a/src/rgw/rgw_common.h
+++ b/src/rgw/rgw_common.h
@@ -1584,7 +1584,12 @@ public:
       return;
     }
 
-    init_ns(b, key.name.substr(pos + 1), key.name.substr(1, pos -1));
+    // skip setting ns if object itself ends with an underscore
+    if (pos != (ssize_t)(key.name.size() -1)){
+      init_ns(b, key.name.substr(pos + 1), key.name.substr(1, pos -1));
+    } else {
+      init(b, key.name);
+    }
     set_instance(key.instance);
   }
 
--- a/src/test/rgw/test_rgw_obj.cc
+++ b/src/test/rgw/test_rgw_obj.cc
@@ -143,11 +143,71 @@ void test_obj(const string& name, const string& ns, const string& instance)
   delete formatter;
 }
 
+
+void test_from_index_key(const string& name, const string& ns, const string& instance){
+  rgw_bucket b;
+  init_bucket(&b, "test");
+
+  JSONFormatter *formatter = new JSONFormatter(true);
+
+  formatter->open_object_section("test");
+
+  rgw_obj_key ok1(name,instance);
+
+  rgw_obj o(b, ok1);
+  rgw_obj obj1(o);
+
+  if (!instance.empty()) {
+    obj1.set_instance(instance);
+  }
+  if (!ns.empty()) {
+    obj1.set_ns(ns);
+  }
+
+  check_parsed_correctly(obj1, name, ns, instance);
+  encode_json("obj1", obj1, formatter);
+
+  bufferlist bl;
+  ::encode(obj1, bl);
+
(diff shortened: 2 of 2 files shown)


theanalyst:
Have only done some basic tests so far, where the patch works with object names starting and ending with _ (eg _foo_) not sure where all we use rados namespaces.

theanalyst:
Changelog:
added unit tests

theanalyst:
retest this please

theanalyst:
retest this please

oritwas:
lgtm

src/rgw/rgw_common.h
```diff
@@ -1584,7 +1584,12 @@ class rgw_obj {
       return;
     }
 
-    init_ns(b, key.name.substr(pos + 1), key.name.substr(1, pos -1));
+    // skip setting ns if object itself ends with an underscore
```
yehudasa: @theanalyst it's not clear to me why we should even be here in that case. In the _temp_ example, we should have escaped the first underscore, so it should have turned into __temp_, and we should have returned earlier. Sounds to me that we fail to encode correctly.

theanalyst: Don't we do the _temp_ -> __temp_ only at set_obj.. which is called by this function (ie. init/init_ns)? Or are any other places where we do the underscore check for object names before this point.

theanalyst:
After looking into this a bit further, There is another bug that exists, which this patch doesn't solve yet. Currently if an object name starts with double underscore,  we chop the first underscore and then call `set_obj`, so while doing a get request, before this patch, calling  `_foo_`, however we can still access it by calling a GET `__foo_` ; so then if someone actually uploads an object called `__foo_` the only way it can be accessed is by calling a  GET `__foo_` for instance.

theanalyst:
> After looking into this a bit further, There is another bug that exists, which this patch doesn't solve yet. Currently if an object name starts with double underscore,  we chop the first underscore and then call `set_obj`, so while doing a get request, before this patch, calling  `_foo_`, however we can still access it by calling a GET `__foo_` ; so then if someone actually uploads an object called `__foo_` the only way it can be accessed is by calling a  GET `__foo_` for instance. 

I assume the check for the underscore only exists in order to avoid the blank namespace again,  why can't we just skip over all the initial underscores and do the init; something like 

```c++
    int name_start=0;
    if (key.name[0] != '_') {
      init(b, key.name);
      set_instance(key.instance);
      return;
    } else {
      for (const auto &c:  key.name){
	if (c != '_'){
	  break;
	}
	name_start++;
      }
    }

    ssize_t pos = key.name.find('_', name_start);
```

cc: @yehudasa @oritwas

oritwas:
set_obj append an underscore if the origin obj starts with underscore so there is no bug regarding objects that starts with double underscore (s3tests include a test for that)

yanghonggang:
I try to put/get a object named '_MMM' with my hammer version rgw. The put/get works well.
But the object header name is : 
`
rgwltt-rgwltt.154100.20___MMM   rgwltt-rgwltt.154100.20__MMM
`
Is this the expect result?
I think all normal object's header should be in this style:

${bucket id}_${object name}

src/rgw/rgw_common.h
yanghonggang: @yehudasa 
Why the first underscore should be escaped of object name?

theanalyst:
closing in favor of https://github.com/ceph/ceph/pull/10939

PR #10534, opened 2016-08-02
Wip undo pulls 10312 10514

dzafman:
Pull requests 10312 and 10514 break ceph_objectstore_tool.py test which vstarts with -n and vstarts without -n to bring back cluster from previous state.  These changes break that.


dzafman:
@wjwithagen When I revert #10312 and #10514  the test/ceph_objectstore_tool.py test works again.  Sorry that I haven't figured out what the problem is, but this tests need to start and stop the daemons using the same config and filestore.

dzafman:
The Jenkins test failure is fixed by #10535

wjwithagen:
@dzafman 
If you look at the error message that it generates, the path is actually 115 characters.
So that is 8 (or 9 with \0) more than allowed.
Which I think is due to the rather long Jenkins-prepath.
But it could be fixed by using a less long testdir path.
Now it is:
ceph_objectstore_tool_dir/out/client.admin.25541.asok

perhaps:
objectstore_dir/out/client.admin.25541.asok

Otherwise you'd have to fix other tools and start putting the sockets in /tmp.

wjwithagen:
@dzafman @batrick @yuriw 
 #10514 is a "simple" fix to actually keep going if the config file is not there to be deleted. And was the first obstacle to get ceph_objectstore_tool.py to start running.
#10312 is a more invasive commit, but even there I don't really see anything that extents the path of CEPH_ROOT
So it is a puzzle why we do not get the path to long error right now.

sad thing is also that vstart does not really report error to ceph_objectstore_tool.py when it fails, and then the whole script just runs into scrambles. I Just got DONE, even when it could not delete the config file.

batrick:
Looking into this.

batrick:
#10543 resolves this. Closing.

batrick:
See also:

PR #10539, opened 2016-08-02
mds: add perf counters for MDLog replay and SessionMap

jcsp:


src/mds/MDLog.cc
```diff
@@ -75,6 +75,8 @@ void MDLog::create_logger()
   plb.add_u64(l_mdl_rdpos, "rdpos", "Journaler  read position");
   plb.add_u64(l_mdl_jlat, "jlat", "Journaler flush latency");
 
+  plb.add_u64_counter(l_mdl_replay, "replay", "Events replayed");
```
gregsfortytwo: I'm not sure "replay" is descriptive enough; why not "events_replayed" or "replayed" (past tense) or something?

gregsfortytwo:
👍 otherwise

src/mds/MDLog.cc
jcsp: changed to "replayed"

PR #10540, opened 2016-08-02
mds: fix daemon selection when starting ranks

jcsp:
Previously, if "standby for rank" was set, but
"standby for fscid" wasn't, the daemon wouldn't
be recognised as the right one for the rank,
and we would end up promoting some arbitrary
standby.


jcsp:
(This was only noticeable when enabling multiple MDSs)

gregsfortytwo:
LGTM

jcsp:
Tested here: http://pulpito.ceph.com/jspray-2016-08-02_12:58:37-fs-wip-jcsp-testing-20160802-distro-basic-mira/

PR #10541, opened 2016-08-02
cmake: install ceph_test_rados_striper_api_*

tchaikov:
so we can test them in ceph-qa-suite


jdurgin:
lgtm assuming it builds on gitbuilder

PR #10542, opened 2016-08-02
doc: describe libvirt client logging

ktdreyer:
Describe how to configure logging for RBD libvirt clients.

Thanks to @dillaman on ceph-devel for this information.

### DIFF
--- a/doc/rbd/libvirt.rst
+++ b/doc/rbd/libvirt.rst
@@ -95,6 +95,18 @@ To configure Ceph for use with ``libvirt``, perform the following steps:
    **NOTE:** You can also use `rbd create`_ to create an image, but we
    recommend ensuring that QEMU is working properly.
 
+.. tip:: Optionally, if you wish to enable debug logs and the admin socket for
+   this client, you can add the following section to ``/etc/ceph/ceph.conf``::
+
+	[client.libvirt]
+	log file = /var/log/ceph/qemu-guest-$pid.log
+	admin socket = /var/run/ceph/$cluster-$type.$id.$pid.$cctid.asok
+
+   The ``client.libvirt`` section name should match the cephx user you created
+   above. If SELinux or AppArmor is enabled, note that this could prevent the
+   client process (qemu via libvirt) from writing the logs or admin socket to
+   the destination locations (``/var/log/ceph`` or ``/var/run/ceph``).
+
 
 
 Preparing the VM Manager


Yan-waller:
@ktdreyer Hi, I'm stucking in this problem right now, my client configuration is as below:
```
[client.admin]
debug_ms = 1/20
log file = /var/log/libvirt/qemu/ceph_ms.log
```
But I can't get any logs form the QEMU process  (I can get the logs by using command like "ceph -s" aright).  Is this has something to do with the client name `[client.admin]` or I'm missing something?

ktdreyer:
@Yan-waller it sounds like you'll want to use whatever keyring libvirt is using (eg `[client.libvirt]`)

Yan-waller:
@ktdreyer thanks for your kindly reminding, I have tried this, 
we use the `[client.cinder]` keyring as following:
```
[root@host-192-31-12-131 qemu]# cat /etc/ceph/ceph.client.cinder.keyring 
[client.cinder]
        key = AQAYKY9XVYUeLBAABSEBWIXleloraUz5RRabtA==
```
so I add `[client.cinder]` to the `/etc/ceph/ceph.conf`,  restart service,
but it doesn't work.  is there anything else I missed?
```
[root@host-192-31-12-131 qemu]# cat /etc/ceph/ceph.conf 
[global]
osd_pool_default_size = 3
mon_host = 192.33.12.32,192.33.12.33,192.33.12.31
osd_journal_size = 30720
public_network = 192.33.12.0/16
mon_initial_members = 192.33.12.32
cluster_network = 192.34.12.0/16
fsid = c7e96434-9b71-58c1-5378-0137ce1d2453

[client.cinder]
debug_ms = 1/20
log file = /var/log/libvirt/qemu/ceph_ms.log
```

dillaman:
@ktdreyer Would you mind adding something about the admin socket as well since you're in there:

```admin socket = /var/run/ceph/$cluster-$type.$id.$pid.$cctid.asok```

Also, AppArmor / SElinux might block access to "/var/log/ceph"

**Comment by ktdreyer** [comment]
Thanks @dillaman , amended and re-pushed with your suggestions.

File: doc/rbd/libvirt.rst [hunk 1]
```diff
@@ -95,6 +95,19 @@ To configure Ceph for use with ``libvirt``, perform the following steps:
    **NOTE:** You can also use `rbd create`_ to create an image, but we
    recommend ensuring that QEMU is working properly.
 
+#. **Optional**: Enable libvirt client debugging. If you wish to enable debug
```
dillaman: The "#" is creating a new numeric bullet point for this section. You could also use a directive like ".. tip:: Enable libvirt ..." to create a nicely formatted section.

ktdreyer:
Thanks, amended and repushed with `.. tip::`

PR #10546, opened 2016-08-02
rgw multisite: use a rados lock to coordinate data log trimming

cbodley:
This PR is based on top of https://github.com/ceph/ceph/pull/10372. I'll rebase once that merges.

You can review the delta here: https://github.com/cbodley/ceph/compare/wip-rgw-data-log-trim...cbodley:wip-rgw-log-trim-lease

### DIFF
--- a/src/rgw/rgw_data_sync.cc
+++ b/src/rgw/rgw_data_sync.cc
@@ -2981,20 +2981,32 @@ class LastTimelogTrimCR : public RGWRadosTimelogTrimCR {
   }
 };
 
-} // anonymous namespace
+class DataLogTrimCR : public RGWCoroutine {
+  RGWRados *store;
+  RGWHTTPManager *http;
+  const int num_shards;
+  const utime_t interval; //< polling interval
+  const std::string& zone; //< my zone id
+  std::vector<rgw_data_sync_status> peer_status; //< sync status for each peer
+  std::vector<rgw_data_sync_marker> min_shard_markers; //< min marker per shard
+  std::vector<std::string> last_trim; //< last trimmed marker per shard
+  int ret{0};
 
-RGWDataLogTrimCR::RGWDataLogTrimCR(RGWRados *store, RGWHTTPManager *http,
-                                   int num_shards, utime_t interval)
-  : RGWCoroutine(store->ctx()), store(store), http(http),
-    num_shards(num_shards), interval(interval),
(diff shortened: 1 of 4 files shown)


src/rgw/rgw_data_sync.cc
```diff
@@ -2935,62 +2934,109 @@ class DataLogTrimCR : public RGWCoroutine {
...
+int DataLogTrimPollCR::operate()
+{
+  reenter(this) {
+    for (;;) {
+      set_status("sleeping");
+      wait(interval);
+
+      // prevent others from trimming for our entire wait interval
+      set_status("acquiring trim lock");
+      yield call(new RGWSimpleRadosLockCR(store->get_async_rados(), store,
```
yehudasa: @cbodley, maybe use RGWContinuousLeaseCR instead? and need to shut it down when done

yehudasa:
@cbodley see my comments, also need to rebase this

cbodley:
@yehudasa ok, will rebase - but i'm not seeing any comments?

yehudasa:
@cbodley maybe now?

cbodley:
rebased and renamed to zone_id

src/rgw/rgw_data_sync.cc
yehudasa: @cbodley right.. I see that discussion. I'm not sure I like the idea of just firing a lease without making the effort of releasing it. Is it a problem releasing it at the end here? Or is it just so that we don't trigger the trim code more frequent that originally planned?

cbodley: > Is it a problem releasing it at the end here? Or is it just so that we don't trigger the trim code more frequent that originally planned?

right. if we release the lock then all of the gateways would try to trim, when we only want it to happen once per trim interval

cbodley: @yehudasa do i you still object to this appoach? any ideas for a better solution?

cbodley:
updated comments to clarify use of rados lock:
```
      // request a 'data_trim' lock that covers the entire wait interval to
      // prevent other gateways from attempting to trim for the duration
...
        // if the lock is already held, go back to sleep and try again later
...
      // note that the lock is not released. this is intentional, as it avoids
      // duplicating this work in other gateways
```

cbodley:
jenkins test this please

PR #10547, opened 2016-08-02
test/common/test_util.cc: FreeBSD does not have distro information

wjwithagen:

### DIFF
--- a/src/test/common/test_util.cc
+++ b/src/test/common/test_util.cc
@@ -32,6 +32,7 @@ TEST(util, unit_to_bytesize)
   ASSERT_EQ(65536ll, unit_to_bytesize(" 64K", &cerr));
 }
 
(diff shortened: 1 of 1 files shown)


src/test/common/test_util.cc
```diff
@@ -32,6 +32,7 @@ TEST(util, unit_to_bytesize)
   ASSERT_EQ(65536ll, unit_to_bytesize(" 64K", &cerr));
 }
 
+#if defined(Linux)
```
tchaikov: where is `Linux` defined? maybe we can use `__linux__` instead?

wjwithagen:
@tchaikov 
Sort of a bad min between cc and cmake naming, fixed.

wjwithagen:
@tchaikov 
Could you merge this please

PR #10550, opened 2016-08-03
kv: fix some bugs in memdb

tanghaodong25:
I'm trying to replace rocksdb with memdb to identify the overheads of rocksdb for bluestore, but I hit a few errors.
1. Can't kill ceph-osd process and then restart it.
2. if using memdb for bitmap allocator, it will hit assert fault in 'split_key' function.
3. fix iterator invalidation in memdb.

### DIFF
--- a/src/kv/MemDB.cc
+++ b/src/kv/MemDB.cc
@@ -478,8 +478,8 @@ int MemDB::MDBWholeSpaceIteratorImpl:: prev()
     return -1;
   }
   free_last();
-  m_iter--;
-  if (m_iter != m_btree_p->end()) {
+  if (m_iter != m_btree_p->begin()) {
+    m_iter--;
     fill_current();
     return 0;
   } else {
--- a/src/kv/MemDB.cc
+++ b/src/kv/MemDB.cc
@@ -270,7 +270,7 @@ int MemDB::_setkey(ms_op_t &op)
   }
 
   m_btree[key] = bufferptr((char *) bl.c_str(), bl.length());
-
+  iterator_seq_no++;
   return 0;
 }
 
@@ -284,6 +284,7 @@ int MemDB::_rmkey(ms_op_t &op)
     assert(m_total_bytes >= bl_old.length());
     m_total_bytes -= bl_old.length();
   }
+  iterator_seq_no++;
   /*
    * Erase will call the destructor for bufferptr.
    */
@@ -341,6 +342,7 @@ int MemDB::_merge(ms_op_t &op)
 
   assert((int64_t)m_total_bytes + bytes_adjusted >= 0);
   m_total_bytes += bytes_adjusted;
+  iterator_seq_no++;
   return 0;
 }
 
@@ -403,6 +405,18 @@ bool MemDB::MDBWholeSpaceIteratorImpl::valid()
   return true;
 }
 
+bool MemDB::MDBWholeSpaceIteratorImpl::iterator_validate() {
+  if (this_seq_no != *global_seq_no) {
+    auto key = m_key_value.first;
+    assert(!key.empty());
+    m_iter = m_btree_p->lower_bound(key);
+    if (m_iter == m_btree_p->end()) {
+      return false;
+    }
+  }
+  return true;
+}
+
 void
 MemDB::MDBWholeSpaceIteratorImpl::free_last()
 {
@@ -442,6 +456,10 @@ bufferlist MemDB::MDBWholeSpaceIteratorImpl::value()
 int MemDB::MDBWholeSpaceIteratorImpl::next()
 {
   std::lock_guard<std::mutex> l(*m_btree_lock_p);
+  if (!iterator_validate()) {
+    free_last();
+    return -1;
+  }
   free_last();
   m_iter++;
   if (m_iter != m_btree_p->end()) {
@@ -455,6 +473,10 @@ int MemDB::MDBWholeSpaceIteratorImpl::next()
 int MemDB::MDBWholeSpaceIteratorImpl:: prev()
 {
   std::lock_guard<std::mutex> l(*m_btree_lock_p);
+  if (!iterator_validate()) {
+    free_last();
+    return -1;
+  }
   free_last();
   m_iter--;
   if (m_iter != m_btree_p->end()) {
@@ -477,7 +499,7 @@ int MemDB::MDBWholeSpaceIteratorImpl::seek_to_first(const std::string &k)
   } else {
     m_iter = m_btree_p->lower_bound(k);
   }
-
+  
   if (m_iter == m_btree_p->end()) {
     return -1;
   }
@@ -487,7 +509,6 @@ int MemDB::MDBWholeSpaceIteratorImpl::seek_to_first(const std::string &k)
 int MemDB::MDBWholeSpaceIteratorImpl::seek_to_last(const std::string &k)
 {
   std::lock_guard<std::mutex> l(*m_btree_lock_p);
-
   free_last();
   if (k.empty()) {
     m_iter = m_btree_p->end();
--- a/src/kv/MemDB.cc
+++ b/src/kv/MemDB.cc
@@ -503,6 +503,7 @@ int MemDB::MDBWholeSpaceIteratorImpl::seek_to_first(const std::string &k)
   if (m_iter == m_btree_p->end()) {
     return -1;
   }
+  fill_current();
   return 0;
 }
 
@@ -520,6 +521,7 @@ int MemDB::MDBWholeSpaceIteratorImpl::seek_to_last(const std::string &k)
   if (m_iter == m_btree_p->end()) {
     return -1;
   }
+  fill_current();
   return 0;
 }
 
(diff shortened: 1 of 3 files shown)


src/kv/MemDB.cc
```diff
@@ -425,6 +425,10 @@ bool MemDB::MDBWholeSpaceIteratorImpl::raw_key_is_prefixed(
     const string &prefix)
 {
   string p, k;
+  size_t pos = m_key_value.first.find(KEY_DELIM, 0);
+  if (pos == std::string::npos) {
+    return false;
+  }
```
liewegas: hmm, why is there a key that doesn't have KEY_DELIM?

src/kv/MemDB.cc
```diff
@@ -471,23 +476,25 @@ int MemDB::MDBWholeSpaceIteratorImpl:: prev()
...
   }
+  fill_current();
   return 0;
 }
 
 int MemDB::MDBWholeSpaceIteratorImpl::seek_to_last(const std::string &k)
 {
   std::lock_guard<std::mutex> l(*m_btree_lock_p);
-
+  if (!(*is_valid)) return -1;
```
liewegas: fix whitespace please (statement on next line)

src/kv/MemDB.cc
```diff
@@ -442,6 +445,7 @@ bufferlist MemDB::MDBWholeSpaceIteratorImpl::value()
 int MemDB::MDBWholeSpaceIteratorImpl::next()
 {
   std::lock_guard<std::mutex> l(*m_btree_lock_p);
+  if (!(*is_valid)) return -1;
```
liewegas: here too

src/kv/MemDB.cc
```diff
@@ -455,6 +459,7 @@ int MemDB::MDBWholeSpaceIteratorImpl::next()
 int MemDB::MDBWholeSpaceIteratorImpl:: prev()
 {
   std::lock_guard<std::mutex> l(*m_btree_lock_p);
+  if (!(*is_valid)) return -1;
```
liewegas: and here

src/kv/MemDB.cc
```diff
@@ -125,6 +125,7 @@ void MemDB::_load()
     dout(10) << __func__ << " Key:"<< key << dendl;
     m_btree[key] = datap;
   }
+  m_total_bytes += bytes_done;
```
liewegas: put this in a different commit?

liewegas:
rebase please?

src/kv/MemDB.cc
tanghaodong25: After checking for iterator invalidation, I no longer hit the 'assert_fail'. So I will delete this commit.

tanghaodong25:
@liewegas have updated and rebased.
Thanks.

src/kv/MemDB.cc
```diff
@@ -397,7 +401,7 @@ void MemDB::MDBWholeSpaceIteratorImpl::fill_current()
 
 bool MemDB::MDBWholeSpaceIteratorImpl::valid()
 {
-  if (m_key_value.first.empty()) {
+  if (!(*is_valid) || m_key_value.first.empty()) {
```
chhabaramesh: Rest looks fine, But I am not sure if current iterator becomes invalid then we should return invalid or restart iterator from current position. And if we return invalid, will bluestore take care of this and restart new iterator. I will confirm soon,

chhabaramesh: I did not notice @liewegas  comment . His comment supersedes mine.

chhabaramesh: After keeping track of seqno, I think just restarting the iterator with m_key_value.first  as key should work. m_iter = m_btree_p->lower_bound(k);

tanghaodong25: @chhabaramesh @liewegas Thanks for your comments.

I'm not sure if bluestore will retry to get a valid iterator before, so I just add a bool. Then I will fix it with seqno:)

src/kv/MemDB.h
```diff
@@ -132,12 +133,16 @@ class MemDB : public KeyValueDB
...
       std::mutex *m_btree_lock_p;
+      uint64_t *btree_seq;
+      uint64_t current_seq;
 
   public:
     MDBWholeSpaceIteratorImpl(btree::btree_map<std::string, bufferptr> *btree_p,
-                             std::mutex *btree_lock_p) {
+                             std::mutex *btree_lock_p, uint64_t *btree_seq_) {
         m_btree_p = btree_p;
         m_btree_lock_p = btree_lock_p;
```
chhabaramesh: I think we need to take btree lock here.

src/kv/MemDB.h
```diff
@@ -147,12 +152,13 @@ class MemDB : public KeyValueDB
     int seek_to_first(const std::string &k);
     int seek_to_last(const std::string &k);
 
-    int seek_to_first() { return seek_to_first(NULL); };
-    int seek_to_last() { return seek_to_last(NULL); };
+    int seek_to_first() { return seek_to_first(""); };
```
chhabaramesh: NULL is not good for cpp that is true but can we make empty string more explicit like std::string("NULL") ?

src/kv/MemDB.h
```diff
@@ -132,12 +133,16 @@ class MemDB : public KeyValueDB
       std::pair<string, bufferlist> m_key_value;
       btree::btree_map<std::string, bufferptr> *m_btree_p;
       std::mutex *m_btree_lock_p;
+      uint64_t *btree_seq;
+      uint64_t current_seq;
```
chhabaramesh: Rename it to something : global_seq_no and my_seq_no or this_seq_no. Current seq_node is confusing .

chhabaramesh:
@tanghaodong25  I have some minor comments on latest code. Please fix those and retest.  Overall looks fine to me.

tanghaodong25:
@chhabaramesh  Thanks. :)

src/kv/MemDB.cc
```diff
@@ -472,24 +494,24 @@ int MemDB::MDBWholeSpaceIteratorImpl::seek_to_first(const std::string &k)
 {
   std::lock_guard<std::mutex> l(*m_btree_lock_p);
   free_last();
-  if (k.empty()) {
+  if (k == "NULL") {
```
liewegas: pretty sure k.empty() is the right thing here...

src/kv/MemDB.cc
```diff
@@ -472,24 +494,24 @@ int MemDB::MDBWholeSpaceIteratorImpl::seek_to_first(const std::string &k)
...
   return 0;
 }
 
 int MemDB::MDBWholeSpaceIteratorImpl::seek_to_last(const std::string &k)
 {
   std::lock_guard<std::mutex> l(*m_btree_lock_p);
-
   free_last();
-  if (k.empty()) {
+  if (k == "NULL") {
```
liewegas: and here

src/kv/MemDB.h
```diff
@@ -147,12 +153,13 @@ class MemDB : public KeyValueDB
     int seek_to_first(const std::string &k);
     int seek_to_last(const std::string &k);
 
-    int seek_to_first() { return seek_to_first(NULL); };
-    int seek_to_last() { return seek_to_last(NULL); };
+    int seek_to_first() { return seek_to_first("NULL"); };
+    int seek_to_last() { return seek_to_last("NULL"); };
```
liewegas: seek_to_first(string()); etc

liewegas: actaully, why change this at all?  doesn't seem related to the fix.

src/kv/MemDB.cc
```diff
@@ -478,11 +478,12 @@ int MemDB::MDBWholeSpaceIteratorImpl:: prev()
...
   }
   free_last();
-  m_iter--;
-  if (m_iter != m_btree_p->end()) {
+  if (m_iter != m_btree_p->begin()) {
+    m_iter--;
     fill_current();
     return 0;
   } else {
+    m_iter--;
```
liewegas: if m_iter == m_btree_p->begin() we shouldn't --, right?  just return -1?

tanghaodong25:
@liewegas  updated. Thanks.

src/kv/MemDB.cc
```diff
@@ -477,17 +499,17 @@ int MemDB::MDBWholeSpaceIteratorImpl::seek_to_first(const std::string &k)
   } else {
     m_iter = m_btree_p->lower_bound(k);
   }
-
+  
   if (m_iter == m_btree_p->end()) {
     return -1;
   }
+  fill_current();
```
liewegas: these two changes are fixing an unrelated bug in seek_to_*, right?  Can you put them in a separate commit?

liewegas:
It would be great to have a test for this in test/objectstore/test_kv.cc... :)

chhabaramesh:
@tanghaodong25, please also run ceph_test_objectstore.

tanghaodong25:
@chhabaramesh  @liewegas  All test in "ceph_test_objectstore" and "ceph_test_keyvaluedb" passed. :)

PR #10552, opened 2016-08-03
jewel: expose buffer const_iterator symbols

tchaikov:
Fixes: http://tracker.ceph.com/issues/16507


dachary:
It passed the rados suite http://tracker.ceph.com/issues/16344#note-17 (except for one known bug that is, I believe, unrelated).

PR #10553, opened 2016-08-03
osd:  remove the redundant clear method in consume_map function

songbaisen:
osd:  remove the redundant clear method in consume_map function

In the above loop before this to_remove to clear,we have remove all
the element.So this clear method is redundant.


dachary:
please explain in the commit message where is the first occurrence and why the code in between makes it redundant

songbaisen:
@dachary Done it.

PR #10554, opened 2016-08-03
hammer: radosgw-admin region-map set is not reporting the bucket quota correctly

Vicente-Cheng:
http://tracker.ceph.com/issues/16815


dang:
This is correct, thanks.

PR #10556, opened 2016-08-03
osd: bluestore: Add per-pool compression capability

ifed01:
This PR enables per-pool compression for BlueStore.
Following parameters were added to pool settings:
compress_hint  - determines if pool contains compressible / incompressible data
compress_algorithm - permits to specify different compression algorithm
compress_ratio - specifies maximum compression ratio 

ObjectStore::Transaction::Op was extended to pass these parameters for any write request.
And corresponding handling was added to BlueStore to support per-write compression parameter specification.


src/common/ceph_strings.cc
```diff
@@ -77,9 +77,17 @@ const char *ceph_osd_alloc_hint_flag_name(int f)
 		return "shortlived";
 	case CEPH_OSD_ALLOC_HINT_FLAG_LONGLIVED:
 		return "longlived";
-	case CEPH_OSD_ALLOC_HINT_FLAG_COMPRESSIBLE:
```
liewegas: Hmm, do we really want to remove the alloc_hint flags, or can they coexist?  The nice thing about the existing ones is that they are persistent, so we have the [in]compressible information at any time (not just for a given write).

src/mon/OSDMonitor.cc
```diff
@@ -3906,6 +3924,21 @@ bool OSDMonitor::preprocess_command(MonOpRequestRef op)
...
+            break;
+	  case COMPRESS_RATIO:
+            ss << "compress_ratio: " <<
+	      (unsigned) p->compress_ratio << "\n";
+            break;
+	  case COMPRESS_HINT:
+            ss << "compress_hint: "
+	       << (p->compress_hint ?
+		    ceph_osd_compress_flag_name(p->compress_hint) :
+		    "default")
```
liewegas: s/default/-/?

src/osd/osd_types.h
```diff
@@ -1345,6 +1350,9 @@ struct pg_pool_t {
       stripe_width(0),
       expected_num_objects(0),
       fast_read(false),
+      compress_algorithm(Compressor::COMP_ALG_NONE),
+      compress_ratio(0),
+      compress_hint(0),
```
liewegas: Instead of making these pg_pool_t members, they can go in the key/value area with random option fields.  See pool_opts_t

liewegas:
My original plan here, instead of adding something like write_params_t, was to add an ObjectStore method like collection_set_param(key, value) that would pass in semi-arbitrary options for a collection.  Then the OSD could feed in these per-pool things for each PG collection from the PGPool update method in the OSD whenever the pool is changed, and we wouldn't need to add any new stuff to individual writes.

Adding per-write flags like incompressible is probably useful, but I think we can use the existing flags field for that?

ifed01:
Replaced with PR #11094

PR #10557, opened 2016-08-03
test: More portable use of mmap(MAP_ANON)

wjwithagen:
From the Linux manual:
  MAP_ANONYMOUS
     The  mapping  is  not  backed  by any file; its contents
     are initialized to zero.  The fd and offset arguments are
     ignored; however, some implementations require fd to be -1
     if MAP_ANONYMOUS (or MAP_ANON) is specified, and portable
     applications  should  ensure  this.

FreeBSD is such a case, it wil just return an error.

 - Used in:
     src/test/on_exit.cc
     src/test/system/cross_process_sem.cc

### DIFF
--- a/src/test/on_exit.cc
+++ b/src/test/on_exit.cc
@@ -64,7 +64,7 @@ int main(int argc, char **argv)
 
   // shared mem for exit tests
   shared_val = (int*)mmap(NULL, sizeof(int),
-      PROT_READ|PROT_WRITE, MAP_SHARED|MAP_ANONYMOUS, 0, 0);
+      PROT_READ|PROT_WRITE, MAP_SHARED|MAP_ANONYMOUS, -1, 0);
   assert(shared_val != MAP_FAILED);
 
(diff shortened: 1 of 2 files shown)


tchaikov:
could you prefix the title of your commit message with the subcomponent your are changing ? see https://github.com/ceph/ceph/blob/master/SubmittingPatches.rst#preparing-and-sending-patches

in this case, the prefix could be "test: ".

wjwithagen:
@tchaikov 
Done....
In this case I really did not know what to put there, but it was simpler than what I was thinking off.

tchaikov:
lgtm.

wjwithagen:
@tchaikov 
Could you merge this one as well. It is a "fix" that even the man page suggests.

tchaikov:
merging. since jenkins is happy.

wjwithagen:
@tchaikov 
Thanx, I'll post some more innocent fruits.

PR #10558, opened 2016-08-03
rgw: store oldest mdlog period in rados

cbodley:
`RGWMetadataManager` no longer makes stat calls on logs shards to determine its oldest_log_period. Instead, it records the oldest log's period id and realm_epoch in a rados object. If `RGWMetadataManager::init_oldest_log_period()` finds no such object, it calls `find_oldest_period()` to read period predecessors until it reaches the realm's first period, and writes that to rados.

`RGWMetadataManager::init_oldest_log_period()` is now only called from `RGWRados::init_complete()` if `run_sync_threads` is true, instead of running unconditionally in `RGWMetadataManager::init()`, which allows us to skip building up a period history for radosgw-admin.

Now that `RGWMetadataManager::read_oldest_log_period()` can return ENOENT for an uninitialized mdlog history, `RGWRemoteMetaLog::run_sync()` will retry on ENOENT as well.

Fixes: http://tracker.ceph.com/issues/16894

### DIFF
--- a/src/rgw/rgw_metadata.cc
+++ b/src/rgw/rgw_metadata.cc
@@ -3,6 +3,7 @@
 
 #include <boost/intrusive_ptr.hpp>
 #include "common/ceph_json.h"
+#include "common/errno.h"
 #include "rgw_metadata.h"
 #include "rgw_coroutine.h"
 #include "cls/version/cls_version_types.h"
@@ -345,6 +346,29 @@ public:
 
 static RGWMetadataTopHandler md_top_handler;
 
+
+static const std::string mdlog_history_oid = "meta.history";
+
+struct RGWMetadataLogHistory {
+  epoch_t oldest_realm_epoch;
+  std::string oldest_period_id;
+
+  void encode(bufferlist& bl) const {
+    ENCODE_START(1, 1, bl);
+    ::encode(oldest_realm_epoch, bl);
+    ::encode(oldest_period_id, bl);
+    ENCODE_FINISH(bl);
+  }
+  void decode(bufferlist::iterator& p) {
+    DECODE_START(1, p);
+    ::decode(oldest_realm_epoch, p);
+    ::decode(oldest_period_id, p);
+    DECODE_FINISH(p);
+  }
+};
+WRITE_CLASS_ENCODER(RGWMetadataLogHistory)
+
+
 RGWMetadataManager::RGWMetadataManager(CephContext *_cct, RGWRados *_store)
   : cct(_cct), store(_store)
 {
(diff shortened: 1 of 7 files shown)


cbodley:
Although the existing logic for oldest_log_period was broken, `test_multi_period_incremental_sync()` in test_multi.py was still passing because it detected the wrong realm_epoch and restarted a full sync. I added the log message "epoch=X in sync status comes before remote's oldest mdlog epoch=Y, restarting sync" to `RGWRemoteMetaLog::run_sync()` to make this case more obvious.

I was able to verify that it works correctly on upgrade. After a run of `test_multi_period_incremental_sync()` on the current master branch, the master zone reports the wrong oldest_log_epoch=4 in `RGWOp_MDLog_Info`. After rebuilding the gateway with this branch and restarting, it goes back to reporting the initial oldest_log_epoch=1.

src/rgw/rgw_metadata.cc
```diff
@@ -363,124 +387,153 @@ RGWMetadataManager::~RGWMetadataManager()
...
+  int ret = rgw_get_system_obj(store, ctx, pool, oid, bl, nullptr, nullptr);
+  if (ret < 0) {
+    return ret;
   }
-};
+  try {
+    auto p = bl.begin();
+    state->decode(p);
+  } catch (buffer::error& e) {
+    return -EINVAL;
```
yehudasa: more like -EIO in this case

yehudasa: and add a log message

cbodley:
@yehudasa added log message and changed error code to -EIO

cbodley:
@yehudasa interesting teuthology failure [here](http://pulpito.ceph.com/cbodley-2016-08-22_16:27:39-rgw-wip-cbodley-testing---basic-mira/379664/):
```
Command failed on mira028 with status 22: 'radosgw-admin log show --object mdlog.history'
```
It sounds like this test doesn't want me to put the mdlog bookkeeping object in the same log pool.. what pool would you recommend using for this?

yehudasa:
@cbodley I'm curious why this test hasn't failed before, we keep a lo of other objects in the same pool. Need to look at that test, see why is that.

cbodley:
@yehudasa in tasks/radosgw_admin.py:785, I see it doing a 'radosgw-admin log list' and a 'radosgw-admin log show' on each object listed. The 'log show' is reading object data and decoding it into `rgw_log_entry`. So it's possible that we have objects, but their only data is in the omap?

yehudasa:
@cbodley ah, that would make sense. Yeah, maybe we need to fix that so that log list only returns objects with a certain prefix.

cbodley:
@yehudasa ah, the test is already filtering on some object names:
```python
        if obj[:4] == 'meta' or obj[:4] == 'data' or obj[:18] == 'obj_delete_at_hint':
            continue
```
guess i'll just rename `mdlog.history` to `meta.history`

cbodley:
updated to use `meta.history` as the object name

i also had to change some logic around when to call `init_oldest_log_period()`. there was another teuthology failure with the sync agent, which sets up two zonegroups with a single zone each, and we set `run_sync_thread = false;` when we're the only zone in our group

cbodley:
no teuthology failures other than valgrind: http://pulpito.ceph.com/cbodley-2016-09-21_11:15:35-rgw-wip-cbodley-testing---basic-mira/

cbodley:
@yehudasa should be ready to merge, when you have time for a final review

PR #10559, opened 2016-08-03
osd: drop unused variables/methods

xiexingguo:


xiexingguo:
@athanatos More cleanups, please help to review. Thanks!

src/osd/OSD.cc
```diff
@@ -3572,7 +3567,7 @@ void OSD::handle_pg_peering_evt(
       if (!pp->is_replicated() && role != pgid.shard)
 	role = -1;
 
-      PG *pg = _create_lock_pg(
+        pg = _create_lock_pg(
```
athanatos: Odd spacing.

src/osd/OSD.cc
```diff
@@ -2920,11 +2920,19 @@ void OSD::clear_temp_objects()
     }
     if (!temps.empty()) {
       ObjectStore::Transaction t;
+      unsigned removed = 0;
       for (vector<ghobject_t>::iterator q = temps.begin(); q != temps.end(); ++q) {
 	dout(20) << "  removing " << *p << " object " << *q << dendl;
 	t.remove(*p, *q);
+        if (++removed > 300) {
```
athanatos: Introduce a config here rather than hardcoding 300.  Probably worth updating the other place 300 is hard coded.

src/osd/OSD.cc
```diff
@@ -2772,8 +2772,6 @@ int OSD::shutdown()
   hb_front_server_messenger->shutdown();
   hb_back_server_messenger->shutdown();
 
-  peering_wq.clear();
-
   return r;
```
athanatos: I apparently introduced that in 1493e7dbfb4accd931e27c563526510492a11a8b, but did not mention why.  I think it's because it's possible for other threads to put something into the peering_wq after the threadpool is stopped.  I suspect that this is actually important.

src/osd/OSD.cc
```diff
@@ -6584,12 +6584,9 @@ struct C_OnMapCommit : public Context {
 
 struct C_OnMapApply : public Context {
   OSDService *service;
-  list<OSDMapRef> pinned_maps;
```
athanatos: This is actually really important.  As long as the refs are alive, the OSDMaps will be pinned in the cache and we won't try to read it off of disk.  If you remove this, the maps will not stay in the cache, and we will try to read those OSDMaps before they are actually written resulting in a crash.  You should remove this commit.

xiexingguo:
@athanatos  Fixed as you suggested. Please check again. Thanks!

xiexingguo:
retest this please

yuriw:
http://pulpito.ceph.com/yuriw-2016-08-05_14:30:29-rados-wip-yuri-testing_2016_08_05-distro-basic-smithi/

PR #10560, opened 2016-08-03
[DNM] simple multithread mds

ukernel:
Simple in-memory MDS for verifying lock design. So far only getattr,setattr,mkdir,mknod,symlink, link,unlink,readdir and rename work.

Basically, The mutex in CInode protects inode itself and its primary parent linkage.  The mutex in directory inode also protects inode's CDirs and  CDentries in CDirs.

A strange thing of this lock design is that: updateing an inode needs two locks: inode's mutex and its parent dir inode's mutex. The reason is that inode is stored in dirfrag, updating an inode actually updates dirfrag.


ukernel:
now, the multi-threads MDS can survive fsstress

ukernel:
bring back basic journal code (no trim log code).  we can run some basic performance test on it

ukernel:
image

dachary:

dachary:

src/CMakeLists.txt
```diff
@@ -912,3 +912,38 @@ if (IS_DIRECTORY "${PROJECT_SOURCE_DIR}/.git")
 endif()
 
 add_subdirectory(script)
+
```
amitkumar50: @ukernel It would be good to make Commit title as 
>DNM: simple multithreaded mds

PR #10561, opened 2016-08-03
Wip 16672 jewel

dzafman:


athanatos:
This needs to go into master first, right?  I think we probably do want to leave this in master even if upgrades go through jewel.

dzafman:
@athanatos I went back and forth on whether this should go into master, but it doesn't hurt anything and it would clean up a  filestore corruption.

So I'll put this in master and then change this into cherry-picks.

dzafman:
This branch should not be merged until we merge #10565

dzafman:
@athanatos All test runs accounted for:  http://pulpito.ceph.com/dzafman-2016-08-03_17:22:07-rados-wip-zafman-testing-jewel-distro-basic-smithi/

The cherry-picks were clean no conflicts.

athanatos:
LGTM, not merging because it still says DNM :)

PR #10567, opened 2016-08-04
os/bluestore: add multiple finishers to bluestore

IlsooByun:
- The single finisher of a bluestore can be a bottleneck when using an SSD as a backend device. If too much load is given to the single finisher, client-side IO latency increases. So we add multiple finishers to the bluestore, which shows better performance.
- ‘bluestore_finisher_threads' option is added to be able to configure the number of finisher threads.

Signed-off-by: Ilsoo Byun <ilsoo.byun@sk.com>

### DIFF
--- a/src/os/bluestore/BlueStore.cc
+++ b/src/os/bluestore/BlueStore.cc
@@ -1380,7 +1380,7 @@ BlueStore::BlueStore(CephContext *cct, const string& path)
 	     cct->_conf->bluestore_wal_thread_timeout,
 	     cct->_conf->bluestore_wal_thread_suicide_timeout,
 	     &wal_tp),
-    finisher(cct),
+    m_finisher_num(1),
     kv_sync_thread(this),
     kv_stop(false),
     logger(NULL),
@@ -1390,10 +1390,26 @@ BlueStore::BlueStore(CephContext *cct, const string& path)
   _init_logger();
   g_ceph_context->_conf->add_observer(this);
   set_cache_shards(1);
+
+  if (cct->_conf->bluestore_use_multiple_finishers) {
+    m_finisher_num = cct->_conf->osd_op_num_shards;
+  }
+
+  for (int i = 0; i < m_finisher_num; ++i) {
+    ostringstream oss;
+    oss << "finisher-" << i;
+    Finisher *f = new Finisher(cct, oss.str(), "finisher");
+    finishers.push_back(f);
+  }
 }
 
 BlueStore::~BlueStore()
 {
+  for (auto f : finishers) {
+    delete f;
+    f = NULL;
+  }
+
   g_ceph_context->_conf->remove_observer(this);
   _shutdown_logger();
   assert(!mounted);
@@ -2762,7 +2778,9 @@ int BlueStore::mount()
       goto out_coll;
   }
 
-  finisher.start();
+  for (auto f : finishers) {
+    f->start();
+  }
   wal_tp.start();
   kv_sync_thread.create("bstore_kv_sync");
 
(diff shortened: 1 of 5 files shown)


src/os/bluestore/BlueStore.cc
```diff
@@ -4819,15 +4838,16 @@ void BlueStore::_txc_finish_kv(TransContext *txc)
...
   }
   if (txc->oncommit) {
-    finisher.queue(txc->oncommit);
+    finishers[txc->osr->id % m_finisher_num]->queue(txc->oncommit);
     txc->oncommit = NULL;
   }
   while (!txc->oncommits.empty()) {
-    finisher.queue(txc->oncommits.front());
+    auto f = txc->oncommits.front();
+    finishers[txc->osr->id % m_finisher_num]->queue(f);
```
liewegas: I think the way to do this is not to use random ids, but to instead assign a shard id to the sequencer when it is created.  This can then map to the same mapping scheme that is distributing pg's across teh op_wq shards in the OSD and the collections across cache shards in BlueStore. Probably we should add the shard_hint to the ObjectStore::Sequencer class and initialize it in the OSD/PG code when the sequencer is created.  Elsewhere we use this to do the mapping:

 cid.hash_to_shard(num_shards)

So maybe make the config option a bool, and then use the g_conf->osd_op_num_shards config value that also drives op_wq shard and cache shard counts?

IlsooByun: Thanks. I have modified according to your comment.

liewegas:
Does the new scheme still show an improvement for your workload? Do you have any numbers you can share?

IlsooByun:
Yes, the new scheme has an enhancement of approximately 30%(from 77k iops to 100k iops, 4kb randwrite). But it's not higher than the previous scheme. Both schemes show similar performance. In the previous version, the number of finishers is set to 16, and in the new version, osd_op_num_shards is set to 20, which is the number of finishers.

src/os/bluestore/BlueStore.cc
```diff
@@ -4819,15 +4841,16 @@ void BlueStore::_txc_finish_kv(TransContext *txc)
...
   }
   if (txc->oncommit) {
-    finisher.queue(txc->oncommit);
+    finishers[txc->osr->parent->shard_hint.hash_to_shard(m_finisher_num)]->queue(txc->oncommit);
     txc->oncommit = NULL;
   }
   while (!txc->oncommits.empty()) {
-    finisher.queue(txc->oncommits.front());
+    auto f = txc->oncommits.front();
+    finishers[txc->osr->parent->shard_hint.hash_to_shard(m_finisher_num)]->queue(f);
```
liewegas: unsigned n = txc->osr->parent->shard_hint.hash_to_shard(m_finisher_num); and then use it for all of these?

src/os/bluestore/BlueStore.cc
```diff
@@ -2780,8 +2798,10 @@ int BlueStore::mount()
   _kv_stop();
   wal_wq.drain();
   wal_tp.stop();
-  finisher.wait_for_empty();
-  finisher.stop();
+  for (vector<Finisher*>::iterator it = finishers.begin(); it != finishers.end(); ++it) {
```
liewegas: for (auto a : finishers) {
   a->wait_for_empty();
 ...

liewegas:
Great.  Do you mind squashing these into one commit?  Otherwise, lgtm!

IlsooByun:
No. I don't mind. Thanks!

liewegas:
https://github.com/ceph/ceph/pull/10780 rebased, squash, and config option name changed.

PR #10570, opened 2016-08-04
osd:  small change to should_share_map function

songbaisen:
osd:  small change to should_share_map function

Just return the bool value and do not need to declare local variable.


dachary:
please explain in the commit message why this helps

songbaisen:
@dachary  :smile: Done it.

dachary:
contrary to what you comment suggests, you are not just changing a local variable, you are also changing the function workflow by adding a return statement where there was no return statement before.

PR #10572, opened 2016-08-04
osd: miscellaneous fixes

xiexingguo:


xiexingguo:
@athanatos Again please help to review. Thanks!

src/osd/OSD.h
```diff
@@ -820,7 +820,7 @@ class OSDService {
   std::atomic_uint promote_probability_millis{1000}; ///< probability thousands. one word.
   PromoteCounter promote_counter;
   utime_t last_recalibrate;
-  unsigned long promote_max_objects, promote_max_bytes;
+  std::atomic_ullong promote_max_objects, promote_max_bytes;
```
athanatos: This one you'll have to ask @liewegas about.  IIRC, the claim at the time was that inconsistent access to these variables was ok since we don't actually care as long as they are eventually consistent.  Please remove from this PR for now.

xiexingguo:
@athanatos Dropped that change and thanks for the review:-)

yuriw:

PR #10576, opened 2016-08-04
lockdep: verbose even if no logging is set

wjwithagen:
Set logging level to 1 for start and stop.
So logfiles are not swamped with unneeded lockdep info


wjwithagen:
@dillaman 
Anything I need to do to get this in?

wjwithagen:
@dillaman 
Thanx Jason

PR #10581, opened 2016-08-05
librbd: prevent creation of v2 image ids that are too large

dillaman:
The librbd API is capped at 24 characters for expressing the
object prefix for data blocks (including trailing null byte).

Fixes: http://tracker.ceph.com/issues/16887


trociny:
lgtm

PR #10582, opened 2016-08-05
hammer: osd: Avoid creation of stray TEMP directories in filestore after split

dzafman:
http://tracker.ceph.com/issues/16917

Includes the backport of a simple fix which improves the _destroy_collection() debug output.

### DIFF
--- a/src/osd/OSD.cc
+++ b/src/osd/OSD.cc
@@ -7028,12 +7028,13 @@ void OSD::split_pgs(
     dout(10) << "m_seed " << i->ps() << dendl;
     dout(10) << "split_bits is " << split_bits << dendl;
 
-    parent->split_colls(
+    if (parent->split_colls(
       *i,
       split_bits,
       i->ps(),
       &child->pool.info,
-      rctx->transaction);
+      rctx->transaction))
+        child->set_temp_coll();
     parent->split_into(
       i->pgid,
       child,
--- a/src/osd/PG.h
+++ b/src/osd/PG.h
@@ -1210,7 +1210,8 @@ public:
   virtual void _scrub_clear_state() { }
   virtual void _scrub_finish() { }
   virtual void get_colls(list<coll_t> *out) = 0;
-  virtual void split_colls(
+  virtual void set_temp_coll() { }
+  virtual bool split_colls(
     spg_t child,
     int split_bits,
     int seed,
--- a/src/osd/PGBackend.h
+++ b/src/osd/PGBackend.h
@@ -325,20 +325,21 @@
      if (temp_created)
        out->push_back(temp_coll);
    }
-   void split_colls(
+   bool split_colls(
      spg_t child,
      int split_bits,
      int seed,
      ObjectStore::Transaction *t) {
      coll_t target = coll_t::make_temp_coll(child);
      if (!temp_created)
-       return;
+       return false;
      t->create_collection(target);
      t->split_collection(
        temp_coll,
        split_bits,
        seed,
        target);
+     return true;
    }
 
    virtual void dump_recovery_info(Formatter *f) const = 0;
@@ -352,6 +353,7 @@
     return temp_coll;
    }
    bool have_temp_coll() const { return temp_created; }
+   void set_temp_coll() { temp_created = true; }
 
    // Track contents of temp collection, clear on reset
    void add_temp_obj(const hobject_t &oid) {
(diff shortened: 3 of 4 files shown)


dzafman:
This passes my crude test script that does the following:

Create Hammer cluster with 6 OSds (OSD=6 vstart)

ceph osd out osd.5
ceph osd out osd.4
ceph osd out osd.3

wait for clean cluster

rados -p rbd bench 30 write --no-cleanup

ceph osd in osd.5
ceph osd in osd.4
ceph osd in osd.3

Wait for clean cluster
This should leave some TEMP directories in the filestores on osd.3, osd.4 and osd.5

ceph osd pool set rbd pg_num 16
Wait for this operation to create new PGs
ceph osd pool set rbd pgp_num 16

Stop cluster

With this fix there should NOT be a pg_TEMP directory WITHOUT a corresponding pg_head directory in any of the filestores on osd.3, 4 or 5

dzafman:
Passed rados suite:

http://pulpito.ceph.com/dzafman-2016-08-05_16:01:19-rados-wip-zafman-testing-hammer-distro-basic-smithi

@athanatos You didn't think we should merge this, but I've provided it here in case you want to reconsider.

athanatos:
@dzafman I think we should merge this, but I'd feel happier if we had a test in the rados suite which reproduces one of the crashes resulting from this bug reliably (EEXIST during collection creation would seem like the easiest symptom to reproduce).

athanatos:
@dzafman I also don't think we should hold up 94.8 for this fix, so it's probably worth writing the test first.

smithfarm:
marked DNM pending release of 0.94.8, then I will remove the DNM

smithfarm:
@dzafman Did the test referred to by https://github.com/ceph/ceph/pull/10582#issuecomment-238993698 get written? If so, can you link to it?

smithfarm:
Also, I'm confused about the "This branch is out-of-date with the base branch" - I guess it means the PR needs to be rebased, otherwise I cannot merge it.

smithfarm:
@dzafman @athanatos Although I am uncertain whether the test called for by https://github.com/ceph/ceph/pull/10582#issuecomment-238993698 was actually added to the hammer ceph-qa-suite, for better or worse this backport passed a rados suite at http://tracker.ceph.com/issues/17151#note-7 with failures that I believe have been addressed (except for http://tracker.ceph.com/issues/15139 which is caused by the build system no longer providing dumpling-era packages).

I have rebuilt the integration branch to include the two fixes and scheduled a new run at http://tracker.ceph.com/issues/17151#note-14

Do you think it's OK to merge provided the second run succeeds?

smithfarm:
@dzafman Also, please rebase to get rid of the "This branch is out-of-date" warning.

dzafman:
@smithfarm Rebased.

smithfarm:
Second run succeeded; third run here: http://pulpito.front.sepia.ceph.com/smithfarm-2016-11-21_11:52:41-rados-hammer-backports---basic-smithi/

athanatos:
Did we ever create a test for this @dzafman ?

dzafman:
@athanatos No, I didn't create a test.   In a way this is redundant with a fix in Jewel to handle the stray TEMP directories .  If we were doing a hammer to pre-fix Jewel upgrade, we could verify that this solves the problem from the hammer end.  Is that what you want?

athanatos:
@dzafman So this cannot cause trouble on hammer at all?

dzafman:
@athanatos I removed the commit that improved the _destroy_collection() logging to reduce the chance of errors.  At least we know that it doesn't break hammer based on rados:thrash suite.  I'll talk to Yuri about create a test x-jewel which can go into the hammer suite to test the upgrade.  All existing upgrades terminate at the testing target (jewel-x, hammer-x).

athanatos:
I mean, I don't remember the impact of this bug.  Was it that the condition was harmless on Hammer and only a problem on upgrade?

dzafman:
@athanatos Yes, it was harmless on Hammer and can crash older jewel osds on upgrade.

smithfarm:
@athanatos This passed another rados run, see http://tracker.ceph.com/issues/17151#note-23

athanatos:
@dzafman Ok, is it a good idea to backport this at all?

dzafman:
This isn't a required change, so I'll close.

PR #10584, opened 2016-08-05
mon/PGMonitor: calc the %USED of pool using used/(used+avail)

tchaikov:
we were using
"the raw space used by this pool" / "the raw space of the whole cluster"
as the %USED. but it's wrong:

- there is chance that not all OSDs are assigned to the pool in question
- we can not utilize all the raw space for the pool: there is overhead.
  and the size of available space for a pool is capped by the assigned
  OSD with minimal free space.

so we should use
 USED / (USED + AVAIL)
as the %USED. so once we have no AVAIL space left, %USED will be 100%.

Fixes: http://tracker.ceph.com/issues/16933

### DIFF
--- a/src/mon/PGMonitor.cc
+++ b/src/mon/PGMonitor.cc
@@ -1373,10 +1373,13 @@ void PGMonitor::dump_object_stat_sum(TextTable &tbl, Formatter *f,
     }
   } else {
     tbl << stringify(si_t(sum.num_bytes));
-    int64_t kb_used = SHIFT_ROUND_UP(sum.num_bytes, 10);
     float used = 0.0;
-    if (pg_map.osd_sum.kb > 0)
-      used = (float)kb_used * raw_used_rate * curr_object_copies_rate / pg_map.osd_sum.kb;
+    if (avail) {
+      used = sum.num_bytes * curr_object_copies_rate;
+      used /= used + avail;
+    } else if (sum.num_bytes) {
+      used = 1.0;
+    }
     tbl << percentify(used*100);
     tbl << si_t(avail);
     tbl << sum.num_objects;
--- a/src/mon/PGMonitor.cc
+++ b/src/mon/PGMonitor.cc
@@ -1350,7 +1350,7 @@ inline string percentify(const float& a) {
 //void PGMonitor::dump_object_stat_sum(stringstream& ss, Formatter *f,
 void PGMonitor::dump_object_stat_sum(TextTable &tbl, Formatter *f,
                                      object_stat_sum_t &sum, uint64_t avail,
-                                     float raw_used_rate, bool verbose, const pg_pool_t *pool) const
+                                     float raw_used_rate, bool verbose, const pg_pool_t *pool)
 {
   float curr_object_copies_rate = 0.0;
   if (sum.num_object_copies > 0)
--- a/src/common/TextTable.h
+++ b/src/common/TextTable.h
@@ -49,9 +49,11 @@ private:
   };
 
   std::vector<TextTableColumn> col;	// column definitions
-  std::vector<std::vector<std::string> > row;	// row data array
   unsigned int curcol, currow;		// col, row being inserted into
   unsigned int indent;			// indent width when rendering
+
+protected:
+  std::vector<std::vector<std::string> > row;	// row data array
 
 public:
   TextTable(): curcol(0), currow(0), indent(0) {}
--- a/src/mon/PGMonitor.h
+++ b/src/mon/PGMonitor.h
@@ -147,11 +147,11 @@ private:
 			  int threshold,
 			  vector<string>& args) const;
 
-  void dump_object_stat_sum(TextTable &tbl, Formatter *f,
-			    object_stat_sum_t &sum,
-			    uint64_t avail,
-			    float raw_used_rate,
-			    bool verbose, const pg_pool_t *pool) const;
+  static void dump_object_stat_sum(TextTable &tbl, Formatter *f,
+				   object_stat_sum_t &sum,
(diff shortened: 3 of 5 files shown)


src/mon/PGMonitor.cc
```diff
@@ -1375,8 +1375,11 @@ void PGMonitor::dump_object_stat_sum(TextTable &tbl, Formatter *f,
     tbl << stringify(si_t(sum.num_bytes));
     int64_t kb_used = SHIFT_ROUND_UP(sum.num_bytes, 10);
     float used = 0.0;
-    if (pg_map.osd_sum.kb > 0)
-      used = (float)kb_used * raw_used_rate * curr_object_copies_rate / pg_map.osd_sum.kb;
+    if (avail <= sum.num_bytes / 99.9) {
```
athanatos: I think this can be simplified to avail < 1.  As long as avail is non-zero, all values of sum.num_bytes are safe.

tchaikov: agreed.

tchaikov:
@athanatos fixed and repushed.

src/mon/PGMonitor.cc
```diff
@@ -1375,8 +1375,11 @@ void PGMonitor::dump_object_stat_sum(TextTable &tbl, Formatter *f,
     tbl << stringify(si_t(sum.num_bytes));
     int64_t kb_used = SHIFT_ROUND_UP(sum.num_bytes, 10);
     float used = 0.0;
-    if (pg_map.osd_sum.kb > 0)
-      used = (float)kb_used * raw_used_rate * curr_object_copies_rate / pg_map.osd_sum.kb;
+    if (avail) {
+      used = static_cast<float>(sum.num_bytes) / (sum.num_bytes + avail);
```
athanatos: I think you also need to include the raw_used_rate and curr_object_copies_rate to turn the logical usage into the replicated usage (1GB USED needs to use up 3GB of space with replication = 3).

tchaikov: `avail` is the nominal amount of space available for new objects stored in the pool in bytes, and `sum.num_bytes` is the nominal amount of existing objects stored in the pool in bytes. so we should use 
```c
sum.num_bytes / (sum.num_bytes + avail)
```
right?

tchaikov:
per the discussion with @athanatos 

1. should use: `sum.num_bytes * curr_object_copies_rate / (sum.num_bytes  * curr_object_copies_rate + avail)` to calc "%USE", so we can take the degraded space into account.
1. we need some test to exercise dump_object_stat_sum() (and probably its friends)

src/mon/PGMonitor.cc
dzafman: LGTM :+1:  I played with "ceph df" using vstart and the change works.   

@tchaikov Did you play with rules that limit OSDs for a pool?

tchaikov: @dzafman thanks for testing and the review.

but i think it would be enough to check the %USED with the expected value.

anyway, i tested by

1. creating a vstart cluster with 4 OSDs, and 
1. adding following crush rules to the default crush rule.
1. creating two pools `pool1` and `pool2` and setting crush_ruleset 1 and 2 for them respectively.
1. and setting the `sum.num_object_copies = 100` and set `sum.num_bytes = 1024*1024*1024*100` of pool1 using gdb
```
host host1 {
     id -3
     alg straw
     hash 0
     item osd.0 weight 1.000
     item osd.1 weight 1.000
}

host host2 {
     id -4
     alg straw
     hash 0
     item osd.2 weight 1.000
     item osd.3 weight 1.000
}

rule rep_rule1 {
    ruleset 1
    type replicated
    min_size 1
    max_size 10
    step take host1
    step choose firstn 0 type osd
    step emit
}

rule rep_rule2 {
    ruleset 2
    type replicated
    min_size 1
    max_size 10
    step take host2
    step choose firstn 0 type osd
    step emit
}
```

```
$ ./bin/ceph osd df
ID WEIGHT  REWEIGHT SIZE  USE  AVAIL %USE  VAR  PGS
 2 1.00000  1.00000 1676G 198G 1477G 11.86 1.00  18
 3 1.00000  1.00000 1676G 198G 1477G 11.86 1.00  18
 0 1.00000  1.00000 1676G 198G 1477G 11.86 1.00  22
 1 1.00000  1.00000 1676G 198G 1477G 11.86 1.00  14
 0 1.00000  1.00000 1676G 198G 1477G 11.86 1.00  22
 1 1.00000  1.00000 1676G 198G 1477G 11.86 1.00  14
 2 1.00000  1.00000 1676G 198G 1477G 11.86 1.00  18
 3 1.00000  1.00000 1676G 198G 1477G 11.86 1.00  18
              TOTAL 6706G 795G 5910G 11.86
MIN/MAX VAR: 1.00/1.00  STDDEV: 0

$ ./bin/ceph df
GLOBAL:
    SIZE      AVAIL     RAW USED     %RAW USED
    6706G     5910G         795G         11.86
POOLS:
    NAME      ID     USED        %USED     MAX AVAIL     OBJECTS
    rbd       0            0         0         1970G           0
    pool1     1      102400M      9.22          985G           0
    pool2     2            0         0          985G           0
```

and the `%USED` of pool1 is expected:
```
In [4]: 102400/ (985*1024+102400)
Out[4]: 0.09216589861751152
```

tchaikov:
fixed and repushed

athanatos:
lgtm, adding needs-qa

tchaikov:
hey @yuriw, how's the testing going?

yuriw:
@tchaikov done :)

tchaikov:
@yuriw thanks =)

PR #10587, opened 2016-08-05
logrotate: Run as root/ceph

b-ranto:
Currently, we run the logrotate scripts as ceph/ceph but that way we
cannot rotate the scripts created by qemu (they are root/ceph and 644).
The original su line was introduced in commit 73d7bed9 because logrotate
complained that the directory was writable by a non-root group and it
needed a su line to supress that error. This way, we keep the group
settings and we can access and rotate the qemu logs as well.

### DIFF
--- a/src/logrotate.conf
+++ b/src/logrotate.conf
@@ -8,5 +8,5 @@
     endscript
     missingok
     notifempty
-    su ceph ceph
+    su root ceph
 }


b-ranto:
This is currently being tested, I'll let you know about the results.

b-ranto:
FYI: This seems to fix the issue with the qemu logs and logrotate does not seem to complain about anything.

ktdreyer:
Thanks for the PR!

Want is file as a Redmine ticket so this will be tracked for Jewel?

ktdreyer:
Boris, have you tested this change to ensure there's no regressions with regard to ceph's own logging behavior? Specifically, the unprivileged ceph daemons can write to the log files fine after a rotation, even with this change?

b-ranto:
@ktdreyer Briefly but it worked fine when it was running as root/root before we switched to ceph/ceph, it should work just fine when run as root/ceph. This option does not change the permissions of the files, it just changes the euid/egid of the logrotate daemon when rotating the files. The files will still inherit the old permissions just like they did before.

smithfarm:
@b-ranto :+1: to flag this for jewel backport.

ktdreyer:
@b-ranto would you please backport this to Jewel?

smithfarm:
Give me a minute. I'm setting up the tracker issues now.

b-ranto:
Backport for jewel:

https://github.com/ceph/ceph/pull/11201

PR #10593, opened 2016-08-05
rgw ldap: enforce simple_bind w/LDAPv3

mattbenjamin:
Found by Harald Klein <hklein@redhat.com>.

### DIFF
--- a/src/rgw/rgw_ldap.h
+++ b/src/rgw/rgw_ldap.h
@@ -70,14 +70,22 @@ namespace rgw {
 	(void) init();
 	return bind();
       }
+      return -EINVAL;
     }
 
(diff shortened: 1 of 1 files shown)


src/rgw/rgw_ldap.h
```diff
@@ -70,14 +70,22 @@ namespace rgw {
 	(void) init();
 	return bind();
       }
+      return -1;
```
cbodley: -EINVAL?

mattbenjamin: ok, that's what we use elsewhere

mattbenjamin: updated

PR #10595, opened 2016-08-05
rocksdb: revert the change introduced by dc41731

tchaikov:
dc41731 reverts the submodule reference of rocksdb, and point it to an
older version which fails to support ppc64le and AArch64.


smithfarm:
@tchaikov @dillaman I see jewel is on rocksdb fa98456 which is before 7ca731b so it doesn't support ppc64le and aarch64. . . Is this change backportable, do you think?

tchaikov:
@smithfarm, as a part of #10438, we update the rocksdb submodule and then reverted that change in #10060. in this change, we reverted the revert.

probably we can drop the change of rocksdb in #10060, and note it in "Conflicts", as that patch won't apply without #10438. and also AFAICT, no bug fix relies on the ppc64le and aarch64 support, so we don't need to backport #10438, right?

smithfarm:
@tchaikov Correct, no jewel bugfixes need it. It's a feature request.

michelmno:
@smithfarm , @tchaikov,  I still need the fix of https://github.com/facebook/rocksdb/commit/7ca731b12ce69596d7b9fec4a72d44ca8a3913c3 to be able to build ceph jewel branch for ppc64le.

tchaikov:
@michelmno i understand, but in general, we don't backport features.

smithfarm:
@tchaikov I tend to see https://github.com/facebook/rocksdb/commit/7ca731b12ce69596d7b9fec4a72d44ca8a3913c3 as a build/ops fix. Could it be cherry-picked so we pick up just this one commit?

tchaikov:
@smithfarm yeah. guess we need to have a tracker ticket for this fix then. @michaelndn could you file one for it?

smithfarm:
@michelmno I guess @tchaikov means that no feature like "ppc64(le) support" was contemplated or announced for jewel. The idea here is just to enable the build to complete on this architecture, not to support it.

tchaikov:
@smithfarm @michelmno exactly.

michelmno:
as suggested in https://github.com/ceph/ceph/pull/10595#issuecomment-241388435
I just created new tracker issue http://tracker.ceph.com/issues/17092

smithfarm:
I just opened https://github.com/ceph/rocksdb/pull/12 - once it is merged, we could bump the rocksdb sha1 in the jewel branch.

PR #10597, opened 2016-08-05
rgw: Fix for incorrect access key id in reworked ldap auth code.

pritha-srivastava:
The reworked ldap code does not use the correct access key id in get_policy.

### DIFF
--- a/src/rgw/rgw_rest_s3.cc
+++ b/src/rgw/rgw_rest_s3.cc
@@ -1714,7 +1714,7 @@ int RGWPostObj_ObjStore_S3::get_policy()
     op_ret = rgw_get_user_info_by_access_key(store, s3_access_key, user_info);
     if (op_ret < 0) {
       S3AuthFactory aplfact(store, s->account_name);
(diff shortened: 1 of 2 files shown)


src/rgw/rgw_rest_s3.cc
```diff
@@ -1713,8 +1713,19 @@ int RGWPostObj_ObjStore_S3::get_policy()
 
     op_ret = rgw_get_user_info_by_access_key(store, s3_access_key, user_info);
     if (op_ret < 0) {
+      class LDAPTokenExtractor : public RGWTokenBasedAuthEngine::Extractor {
```
mattbenjamin: is it really ideal to define the class inline like this?

pritha-srivastava: Well, I have defined it inline, because the class will just be used by this method. Like I said I will improve it later.

pritha-srivastava:
@mattbenjamin : Moved the class outside in the header file.

PR #10598, opened 2016-08-05
cmake: correct ceph_test_librbd/ceph_test_rbd_mirror linkage

dillaman:
This now matches the autotools configuration and fixes the failed
assertion during test execution.

Fixes: http://tracker.ceph.com/issues/16882


tchaikov:
tested locally.

PR #10600, opened 2016-08-05
ceph-osd,ceph-mon,ceph-mds: daemonize before creating messenger

vuhuong:
ceph-osd/ceph-mon/ceph-mds daemonize (fork) after creating messengers
Xio messenger will initialize accelio library and register RDMA memory
in the 1st call to XioMessenger constructor. This situation is very
problematic where child process do rdma operations (using rdma resources
created by parent)

http://www.rdmamojo.com/2012/05/24/ibv_fork_init
http://www.spinics.net/lists/linux-rdma/msg03364.html

This patch force to daemonize/fork before creating messenger
(ie. initializing and registering rdma memory before daemonizing
for xio messenger)

Signed-off-by: Avner BenHanoch avnerb@mellanox.com
Signed-off-by: Vu Pham vu@mellanox.com

### DIFF
--- a/src/ceph_mds.cc
+++ b/src/ceph_mds.cc
@@ -135,6 +135,8 @@ int main(int argc, const char **argv)
       << "' is invalid and will be forbidden in a future version.  "
       "MDS names may not start with a numeric digit." << dendl;
   }
+  // daemonize
+  global_init_daemonize(g_ceph_context);
 
   Messenger *msgr = Messenger::create(g_ceph_context, g_conf->ms_type,
 				      entity_name_t::MDS(-1), "mds",
@@ -172,7 +174,6 @@ int main(int argc, const char **argv)
   if (r < 0)
     exit(1);
 
-  global_init_daemonize(g_ceph_context);
   common_init_finish(g_ceph_context);
 
   // get monmap
--- a/src/ceph_mon.cc
+++ b/src/ceph_mon.cc
@@ -660,6 +660,11 @@ int main(int argc, const char **argv)
     }
   }
 
+  if (g_conf->daemonize) {
+    global_init_postfork_finish(g_ceph_context);
+    prefork.daemonize();
+  }
+
   // bind
   int rank = monmap.get_rank(g_conf->name.get_id());
   Messenger *msgr = Messenger::create(g_ceph_context, g_conf->ms_type,
@@ -746,11 +751,6 @@ int main(int argc, const char **argv)
     derr << "compacting monitor store ..." << dendl;
     mon->store->compact();
     derr << "done compacting" << dendl;
-  }
-
-  if (g_conf->daemonize) {
-    global_init_postfork_finish(g_ceph_context);
-    prefork.daemonize();
   }
 
   msgr->start();
--- a/src/ceph_osd.cc
+++ b/src/ceph_osd.cc
@@ -444,6 +444,8 @@ int main(int argc, const char **argv)
 	 << " **          you specify neither or both.                             **"
 	 << TEXT_NORMAL << dendl;
   }
+  // daemonize
+  global_init_daemonize(g_ceph_context);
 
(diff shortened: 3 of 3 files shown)


vuhuong:
I have successfully brought up a cluster with 4 nodes, 8 osds per node,  two monitors and run I/Os (4K-4M blocks size) from several fio instances

tchaikov:
see the discussion at http://www.spinics.net/lists/ceph-devel/msg31823.html

Adirl:
@yuyuyu101 pls see this old PR to address fork issue.
what do you think about this kind of change?

yuyuyu101:
@tchaikov the only reason to bind before fork is checking whether binding is ok and go to fail fast?

yuyuyu101:
@Adirl 
rgw/rgw_main.cc
263:    global_init_daemonize(g_ceph_context);

rgw/rgw_object_expirer.cc
76:    global_init_daemonize(g_ceph_context);

Adirl:
I looked at the pointers you provided @yuyuyu101 
in the previous changes we moved the daemonize to occur before creating the messenger, 
since we are not creating messengers in rgw_main.cc and rgw_object_expirer.cc where do you want to re-place the global_init_daemonize?

@saritz @danielbo

Adirl:
@saritz

saritz:
@yuyuyu101 
Hi, 
I looked at the files you mentioned and I don't find that there is a call to create messenger before calling function: "global_init_daemonize(g_ceph_context);"
Please let us know if you see a need to change the order of the call on these files: 
rgw/rgw_main.cc
rgw/rgw_object_expirer.cc

Or others...

Thank you
@Adirl

yuyuyu101:
oh, looks ok to rgw.

What's your suggestion on this ? @tchaikov

tchaikov:
> the only reason to bind before fork is checking whether binding is ok and go to fail fast?

@yuyuyu101 two reasons, 

1. to let the daemon fail fast
1. print error in the parent process in that case

maybe we can use the `Preforker` to the daemonize? so the child process is able to pass the error code to parent process? see http://www.spinics.net/lists/ceph-devel/msg32212.html

yuyuyu101:
the problem is ibverbs can't support fork. @Adirl  is this right?

Adirl:
ibverbs can support fork only if done before registering any memory
like this 
1. fork
2. ibv_reg_mem()

otherwise child process cannot access the RDMA pinned memory

see more info here http://www.rdmamojo.com/2012/05/24/ibv_fork_init

yuyuyu101:
yes. but we already reg memory before fork... so the option now is unregister memory then register again

Adirl:
yes today we register and then fork, and this is not supported
this patch is suggesting to reverse the order and move the daemonize before creating the messenger.
is that possible?

yuyuyu101:
@saritz I'm not sure your problem is related to this pr....

@tchaikov any comment?

saritz:
@yuyuyu101 @tchaikov
Please ignore the problem. I found out that for some reason the cluster had 3 OSD with id 1,2,3 correspondingly, and there was no OSD 0. 
When configuring the cluster again, all was fine.
I continue to check this patch, and update you
Thank you
@Adirl

yuyuyu101:
call for @liewegas @gregsfortytwo too. this is a barrier to let rdma backend become a easy use option. osd,mon without this pr can't run background now. hope we have a conclusion ASAP

Adirl:
successfully tested on RDMA cluster, now running in background

Adirl:
@yuyuyu101 
ceph_mgr.cc +52
  global_init_daemonize(g_ceph_context);

where can we move it so fork will happen before creating the DaemonServer  and MgrStandby messengers ?

yuyuyu101:
closed now, feel free to open again if any components needs this

PR #10601, opened 2016-08-06
mon/MonClient.h: remove repeated searching of map

stiopaa1:

### DIFF
--- a/src/mon/MonClient.h
+++ b/src/mon/MonClient.h
@@ -210,14 +210,19 @@ private:
   void handle_subscribe_ack(MMonSubscribeAck* m);
 
   bool _sub_want(const string &what, version_t start, unsigned flags) {
-    if ((sub_new.count(what) == 0 &&
-	 sub_sent.count(what) &&
-	 sub_sent[what].start == start &&
-	 sub_sent[what].flags == flags) ||
-	(sub_new.count(what) &&
-	 sub_new[what].start == start &&
-	 sub_new[what].flags == flags))
+    auto sub = sub_new.find(what);
+    if (sub != sub_new.end() &&
+        sub->second.start == start &&
+        sub->second.flags == flags) {
       return false;
+    } else {
+      sub = sub_sent.find(what);
+      if (sub != sub_sent.end() &&
+	  sub->second.start == start &&
+	  sub->second.flags == flags)
+	return false;
+    }
+
     sub_new[what].start = start;
     sub_new[what].flags = flags;
     return true;


src/mon/MonClient.h
```diff
@@ -213,14 +213,19 @@ class MonClient : public Dispatcher {
...
 
   bool _sub_want(string what, version_t start, unsigned flags) {
-    if ((sub_new.count(what) == 0 &&
-	 sub_sent.count(what) &&
-	 sub_sent[what].start == start &&
-	 sub_sent[what].flags == flags) ||
-	(sub_new.count(what) &&
-	 sub_new[what].start == start &&
-	 sub_new[what].flags == flags))
+    auto sub_new_iter = sub_new.find(what);
```
tchaikov: could simply use `sub` as the variable name, as no need to encode the type into the variable name.

src/mon/MonClient.h
```diff
@@ -213,14 +213,19 @@ class MonClient : public Dispatcher {
...
+    if(sub_new_iter != sub_new.end() &&
```
tchaikov: add a space after `if`, please.

src/mon/MonClient.h
```diff
@@ -213,14 +213,19 @@ class MonClient : public Dispatcher {
...
+       sub_new_iter->second.start == start &&
+       sub_new_iter->second.flags == flags) {
       return false;
+    } else {
+      auto sub_sent_iter = sub_sent.find(what);
+      if(sub_sent_iter != sub_sent.end() &&
```
tchaikov: add a space after `if`, please.

src/mon/MonClient.h
stiopaa1: @tchaikov 
What's wrong with adding iter to name of the container? Doesn't it make very explicit what variable is doing?

If I rename sub_new_iter to sub, how should I name sub_sent_iter?

tchaikov: > What's wrong with adding iter to name of the container? Doesn't it make very explicit what variable is doing?

for example, if we want to find an item in a `vector<int>`, one could put
```c++
auto vector_iter = haystack_vector.find(straw_i);
```
or he/she could use
```c++
auto found = v.find(straw);
```

nothing wrong with `sub_new_iter`. but by looking at the (exaggerated) example, probably you can see the difference.

> If I rename sub_new_iter to sub, how should I name sub_sent_iter?

you don't need a new variable name, since it will also be a "sub", so why not just reuse it? like:
```c++
sub = sub_sent.find(what);
```

stiopaa1: with the haystack_vector example I would use 
`auto haystack_iter = haystack_vector.find(straw_i);`
which for me is much more readeable than 
`auto found = v.find(straw);`

haystack_iter tells me straight away that I've got a variable pointing to an element in haystack that is an iterator, so I have to dereference it if I want to access the value.
When I look at a variable named `found` I think about boolean type. 

I have changed it to 'sub' as you proposed earlier, but I think it's now less readable than it was before:) For me at first glance it looks like we are comparing the same thing in both branches of if...

liewegas:
needs rebase

stiopaa1:
@liewegas 
rebased

tchaikov:

failed tests are tracked at

PR #10602, opened 2016-08-06
Change debug levels for certain log messages

smanjara:
Currently while running radosgw-admin command we see these messages thrown at
the endusers since debug level is set to 0:

2016-08-02 14:48:28.672409 7f9e630639c0  0 error in read_id for id  : (2) No
such file or directory
2016-08-02 14:48:28.687251 7f9e630639c0  0 Cannot find zone
id=7a2a89b9-4f5e-4f60-a29e-451ac9acc5a8 (name=us-west), switching to local
zonegroup configuration
2016-08-02 14:48:35.742185 7f9e630639c0  0 error read_lastest_epoch
.rgw.root:periods.a3db7884-2445-4bab-a165-6730f8573b8f:staging.latest_epoch

Raising the log level to 20 since the command succeeds and these are not actual
errors.

Fixes: http://tracker.ceph.com/issues/16935

### DIFF
--- a/src/rgw/rgw_rados.cc
+++ b/src/rgw/rgw_rados.cc
@@ -975,7 +975,7 @@ int RGWPeriod::read_latest_epoch(RGWPeriodLatestEpochInfo& info)
   RGWObjectCtx obj_ctx(store);
   int ret = rgw_get_system_obj(store, obj_ctx, pool, oid, bl, NULL, NULL);
   if (ret < 0) {
-    ldout(cct, 0) << "error read_lastest_epoch " << pool << ":" << oid << dendl;
+    ldout(cct, 1) << "error read_lastest_epoch " << pool << ":" << oid << dendl;
     return ret;
   }
   try {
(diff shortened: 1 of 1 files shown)


wjwithagen:
@smanjara 
When you look at the code, they are actually errors. Even the string says so,
Now if these errors are expected it micght be that you raise the log-level.
But I would not recommend setting them at 20. a simple 1 would maybe be sufficient, since  normally one would run the code without debug set

src/rgw/rgw_rados.cc
```diff
@@ -408,7 +408,7 @@ int RGWSystemMetaObj::init(CephContext *_cct, RGWRados *_store, bool setup_obj,
       r = read_id(name, id);
       if (r < 0) {
         if (r != -ENOENT) {
-          ldout(cct, 0) << "error in read_id for object name: " << name << " : " << cpp_strerror(-r) << dendl;
+          ldout(cct, 20) << "error in read_id for object name: " << name << " : " << cpp_strerror(-r) << dendl;
```
cbodley: the line `if (r != -ENOENT) {` was already added to prevent this from being logged in normal cases, so i think leaving this one at 0 is ok

smanjara: Changed it back.

cbodley:
i tend to agree with @wjwithagen, a debug level of 1 seems appropriate for these messages

smanjara:
Agreed. Makes more sense to change it to debug level 1.

cbodley:
@smanjara thanks, looks good. could you fix up your commit message so the first line has a short description for 'git log'? something like this:
```
rgw: raise log levels for common radosgw-admin errors

Currently while running radosgw-admin command ...
```

smanjara:
@cbodley Done! Thanks.

PR #10603, opened 2016-08-06
[DNM] FreeBSD changes to make run-tox-ceph-disk work

wjwithagen:


src/ceph-disk/tests/ceph-disk.sh
```diff
@@ -381,17 +421,24 @@ function run() {
...
         setup $dir || return 1
         set -x
-        $action $dir || return 1
+        $action $dir 
         set +x
-        teardown $dir || return 1
+	status=$?
+	if [ $status -ne 0 ] ; then
+	    # Keep data for postmortum analysis
+            teardown_no_rm $dir
```
tchaikov: maybe we can have this change for another PR? it don't think this blocks the FreeBSD porting.

src/ceph-disk/run-tox.sh
```diff
@@ -17,6 +17,12 @@
 
 # run from the ceph-disk directory or from its parent
 : ${CEPH_DISK_VIRTUALENV:=/tmp/ceph-disk-virtualenv}
+if [ -z $CEPH_BIN ]; then
```
tchaikov: does this fix the FreeBSD build?

src/ceph-disk/tests/ceph-disk.sh
wjwithagen: @tchaikov 
It is not essential, but rather very convenient during development.
I'll take it so a separate PR

src/ceph-disk/run-tox.sh
wjwithagen: @tchaikov 
It allows it to run, which helps a lot.
But ceph-disk will require quite a lot of low-level device code to actually match what is available as devices in FreeBSD.
So perhaps I should set this to DNM? I just did not want to lose this code, and get some others want to help FreeBSD porting to get going.

src/ceph-disk/tests/ceph-disk.sh
```diff
@@ -150,7 +161,34 @@ function test_path() {
 function test_no_path() {
     local dir=$1
     shift
-    ( export PATH=../ceph-detect-init/virtualenv/bin:virtualenv/bin:$CEPH_BIN:/usr/bin:/bin ; test_activate_dir $dir) || return 1
+    ( export PATH=../ceph-detect-init/virtualenv/bin:virtualenv/bin:$CEPH_BIN:/usr/bin:/bin:/usr/local/bin ; test_activate_dir $dir) || return 1
+}
+
+function run_timeout() {
```
tchaikov: could you move this change into a separated commit?

src/ceph-disk/tests/ceph-disk.sh
```diff
@@ -349,6 +387,8 @@ function test_ceph_osd_mkfs() {
 
 function run() {
     local dir=$1
+    local status
```
tchaikov: is this variable used anywhere?

src/ceph-disk/tests/ceph-disk.sh
```diff
@@ -53,19 +53,30 @@ function setup() {
     touch $dir/ceph.conf # so ceph-disk think ceph is the cluster
 }
 
-function teardown() {
+function teardown_no_rm() {
```
tchaikov: why would you want to extract this function out of `teardown()`?

wjwithagen:
@tchaikov 
Perhaps this needs to go to DNM, since it needs more attention. And the CMkae stuff distracted me full major from working on getting more tests to complete.

Because the way things are now setup, the testdir is delete even if the test does not complete correctly.
So I would like to have 2 cleanups: 
 1 for success that get ride of all files and processes
 2 that keeps the files for post-mortum analysis, like going thru the log files
    or looking at the files in the filestore to see what is really there.

PR #10604, opened 2016-08-07
filestore: fix skip omap clone in _collection_move_rename

XinzeChi:
object_map->clone would be skip because the spos is sync
in _set_replay_guard

Fixes: http://tracker.ceph.com/issues/16948


XinzeChi:
@liewegas , please review.

liewegas:
I think I'm not following the failure case you're thinking about.  Can you explain it in more detail?  This doesn't seem quite right.

Also, I don't think not syncing will help, since another thread can sync the store at any time.

XinzeChi:
make a mistake

PR #10606, opened 2016-08-07
mds: trim null dentries proactively

jcsp:
Instead of leaving null dentries (e.g. left
behind from unlinks) in the cache until they
fall out of the LRU, actively push them
to the bottom of the LRU and then consume
all nulls at the bottom in trim() even if
the cache is not oversized yet.

This fixes the case where standby replay daemons
would otherwise accumulate a cache full of
null dentries resulting from unlinks, and it
makes the behaviour of active daemons more
deterministic.

Fixes: http://tracker.ceph.com/issues/16919

### DIFF
--- a/src/mds/MDCache.cc
+++ b/src/mds/MDCache.cc
@@ -6356,10 +6356,22 @@ bool MDCache::trim(int max, int count)
   bool is_standby_replay = mds->is_standby_replay();
   int unexpirable = 0;
   list<CDentry*> unexpirables;
-  // trim dentries from the LRU
-  while (lru.lru_get_size() + unexpirable > (unsigned)max) {
+
+  // trim dentries from the LRU: only enough to satisfy `max`,
+  // unless we see null dentries at the bottom of the LRU,
+  // in which case trim all those.
+  bool trimming_nulls = true;
+  while (trimming_nulls || lru.lru_get_size() + unexpirable > (unsigned)max) {
     CDentry *dn = static_cast<CDentry*>(lru.lru_expire());
-    if (!dn) break;
+    if (!dn) {
+      break;
+    }
+    if (!dn->get_linkage()->is_null()) {
+      trimming_nulls = false;
+      if (lru.lru_get_size() + unexpirable <= (unsigned)max) {
+        break;
+      }
+    }
     if ((is_standby_replay && dn->get_linkage()->inode &&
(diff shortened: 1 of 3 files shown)


gregsfortytwo:
Okay, so this is really only pushing them to the bottom on replay, right? The commit message concerned me (as obviously we sometimes want to keep them around for fast responses on incomplete dirs).

jcsp:
Tested by https://github.com/ceph/ceph-qa-suite/pull/1111

jcsp:
Right, the extra touch_dentry_bottom calls are only in replay.  There are other places that we do that already, like unlinks and renames, so those will also get those stray dentries thrown away immediately instead of waiting for the cache fill up.  The (important) case of null dentries that cache lookup misses should be unaffected though.

gregsfortytwo:
LGTM

gregsfortytwo:
http://pulpito.ceph.com/gregf-2016-08-25_19:08:49-fs-greg-fs-testing-825---basic-mira/385454/ may be this PR's fault; testing without it.

gregsfortytwo:
http://pulpito.ceph.com/gregf-2016-08-29_04:22:27-fs-greg-fs-testing-828---basic-mira/ has the same branch minus this PR running that test job.

gregsfortytwo:
Yeah, looks like it's busted for some reason. I unzipped the MDS logs and looked a little bit and I'm not quite sure why they're stuck, but it does indeed involve stray inodes and trimming on the export.

You may just have lucked into revealing a bug that was being disguised by timing and isn't any more; not sure.

jcsp:
The offending test (TestStrays.test_migration_on_shutdown) is passing with latest update to this patch:
http://pulpito.ceph.com/jspray-2016-09-07_23:07:40-fs-wip-jcsp-testing-20160907-testing-basic-mira/405046/

I suspect there was an underlying bug here that was triggered by the extra trimming, so I'm going to create a branch that sets cache size to zero to see if it triggers it.

jcsp:
@gregsfortytwo quick re-review?

jcsp:
(the bit that changed was the extra ` if (lru.lru_get_size() + unexpirable <= (unsigned)max) {` check inside the loop.  Previously it was always going through the loop at least once so each tick was trimming one item (which should have still been legal but wasn't the intent of this patch)

gregsfortytwo:
Reviewed-by:

PR #10607, opened 2016-08-07
mds: ensure fragmentation happens promptly

jcsp:
Hit directories during file creation so that we
don't create oversized fragments (previously
would wait until other metadata ops to break them up).

Call MDBalancer::do_fragmenting at the end of
MDSRank dispatch, to avoid waiting for the next
tick to fragment dirs.  Previously, we would end up
processing up to 5 seconds of extra creates before
actually doing the split, leading to oversized fragments.

### DIFF
--- a/src/mds/MDSRank.cc
+++ b/src/mds/MDSRank.cc
@@ -446,6 +446,12 @@ bool MDSRank::_dispatch(Message *m, bool new_msg)
   // finish any triggered contexts
   _advance_queues();
 
+  // In case any dirs hit in this request are now enqueued for
+  // split or merge -- avoid waiting for the next tick() to do so
+  // so that the split happens before any subsequent requests that
+  // might otherwise lead to an oversized fragment.
+  balancer->do_fragmenting();
+
   if (beacon.is_laggy()) {
     // We've gone laggy during dispatch, don't do any
     // more housekeeping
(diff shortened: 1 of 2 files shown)


ukernel:
I think do fragmentation every 5 seconds is good enough. No need to call MDBalancer::do_fragmenting in MDSRank dispatch

jcsp:
I don't think we should wait 5 seconds just because we can: at the point we've decided to split a directory we should go ahead and do it immediately.  It's much easier to test it this way, because the size of dirfrags is strictly limited and we can assert that the limit is not exceeded

Calling it on every dispatch is a bit gratuitous (although it is quite a cheap check), so I should probably change this to do a MDSRank::queue_waiter when we have something waiting to split.

gregsfortytwo:
Well, don't forget that fragmenting a directory requires us to freeze it. I'm not sure we want to enable that on every op since it's not uncommon to see a quick burst of ops on a single inode which we would probably prefer not to interrupt. (The counter-argument is that we're probably better off not freezing *all* dirs on a predictable cycle, but I doubt that's as big a problem. Need data!)

jcsp:
We can have the best of both worlds (not doing it immediately but also being a bit more deterministic) by having a limit on how far fragments are allowed to exceed the split size, so that we don't necessarily split immediately, but we also don't have an unbounded 5-seconds-worth of growth past the limit.

jcsp:
Also worth noting that the current situation is that the user has a random chance of having the split done immediately anyway if they happen to come in right before a tick.  If we really want the behaviour of "wait a bit before splitting" then tick() isn't accomplishing that reliably either.

jcsp:
Closing in favour of https://github.com/ceph/ceph/pull/12022

PR #10609, opened 2016-08-08
[wip] rgw: test_multi use logging module for logs

theanalyst:
This allows for simultaneously logging to stdout and a log file, and
reduce the verbosity of stdout to error messages, also log stderr, which
is only being printed currently (allows printing to both stdout and
logfile if the logging level is set to debug for stdout)

Signed-off-by: Abhishek Lekshmanan abhishek@suse.com

### DIFF
--- a/src/test/rgw/test_multi.py
+++ b/src/test/rgw/test_multi.py
@@ -6,6 +6,8 @@ import string
 import argparse
 import sys
 import time
+import logging
+
 try:
     from itertools import izip_longest as zip_longest
 except ImportError:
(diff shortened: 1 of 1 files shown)


cbodley:
love the idea. there was way too much output from the radosgw-admin commands, and it's great to be able to filter that down. only issue is that we're printing a big WARNING or ERROR for normal output:
```
2016-09-14 16:00:30,687 - ERROR - bootstrapping clusters
2016-09-14 16:00:30,687 - WARNING - running cmd: /home/cbodley/ceph/src/mstart.sh c1 -n --mds_num 0
```

theanalyst:
@cbodley yeah I saw that too, I'll try to move these  messages to INFO or something and set that as the default level

theanalyst:
closed in favor of #14433 which incorporates this

PR #10610, opened 2016-08-08
osd: misc fixes and cleanups

xiexingguo:


xiexingguo:
@athanatos You comments are appreciated.

src/common/WeightedPriorityQueue.h
```diff
@@ -84,6 +84,8 @@ class WeightedPriorityQueue :  public OpQueue <T, K>
       }
       //Get the cost of the next item to dequeue
       unsigned get_cost() const {
+        if (empty())
```
athanatos: I'd rather this were an assert.

src/osd/PG.cc
```diff
@@ -797,11 +798,8 @@ void PG::generate_past_intervals()
     dout(10) << __func__ << " past_intervals " << past_intervals << dendl;
     // Fix it
     info.history.same_interval_since = same_interval_since;
+    dirty_info = true;
   }
-
-  // record our work.
-  dirty_info = true;
-  dirty_big_info = true;
 }
```
athanatos: This commit isn't worth it.  generate_past_intervals() is almost never called -- other PGs provide the history when they provide the info.

src/osd/OSD.cc
```diff
@@ -3153,12 +3153,13 @@ PG *OSD::_create_lock_pg(
   return pg;
 }
```
athanatos: This isn't actually a race.  It's not obvious, but you need to hold the osd_lock (as well as the pg_map_lock) to modify the pg_map.  Holding either the pg_map_lock or the osd_lock is sufficient to ensure that it won't be modified.  The reason for this complexity is that the fast_dispatch path relies on being able to hold the pg_map_lock and not the osd_lock to read (but not modify) the pg_map.  That said, this is a perfectly good patch with the exception that you should return an rvalue PGRef.

src/osd/OSD.cc
```diff
@@ -3153,12 +3153,13 @@ PG *OSD::_create_lock_pg(
...
 
-PG *OSD::get_pg_or_queue_for_pg(const spg_t& pgid, OpRequestRef& op)
+PGRef OSD::get_pg_or_queue_for_pg(const spg_t& pgid, OpRequestRef& op)
 {
+  PGRef out;
```
athanatos: Don't declare this here.

src/osd/OSD.cc
```diff
@@ -3153,12 +3153,13 @@ PG *OSD::_create_lock_pg(
...
   Session *session = static_cast<Session*>(
     op->get_req()->get_connection()->get_priv());
   if (!session)
-    return NULL;
+    return out;
```
athanatos: return PGRef();

src/osd/OSD.cc
```diff
@@ -3171,14 +3172,15 @@ PG *OSD::get_pg_or_queue_for_pg(const spg_t& pgid, OpRequestRef& op)
...
   if (wlistiter == session->waiting_for_pg.end()) {
-    out = i->second;
+    pg = i->second;
   } else {
     wlistiter->second.push_back(op);
     register_session_waiting_on_pg(session, pgid);
   }
   session->put();
+  out.reset(pg);
   return out;
```
athanatos: return PGRef(pg);

xiexingguo:
@athanatos All comments are addressed. Thanks!

src/osd/OSD.cc
```diff
@@ -8556,11 +8556,11 @@ void OSD::handle_op(OpRequestRef& op, OSDMapRef& osdmap)
...
   }
 
-  PG *pg = get_pg_or_queue_for_pg(pgid, op);
-  if (pg) {
+  PGRef pg = get_pg_or_queue_for_pg(pgid, op);
+  if (pg.get()) {
     op->send_map_update = share_map.should_send;
     op->sent_epoch = m->get_map_epoch();
-    enqueue_op(pg, op);
+    enqueue_op(pg.get(), op);
```
athanatos: How annoying would it be to update enqueue_op to take a PGRef?  It's going to build a PGRef anyway, might as well do it here and avoid the .get() nonsense.

src/osd/OSD.cc
```diff
@@ -8644,11 +8644,11 @@ void OSD::handle_replica_op(OpRequestRef& op, OSDMapRef& osdmap)
     peer_session->put();
   }
 
-  PG *pg = get_pg_or_queue_for_pg(m->pgid, op);
-  if (pg) {
+  PGRef pg = get_pg_or_queue_for_pg(m->pgid, op);
+  if (pg.get()) {
```
athanatos: I'm pretty sure you can just do if (pg) here.

src/common/WeightedPriorityQueue.h
```diff
@@ -166,6 +167,8 @@ class WeightedPriorityQueue :  public OpQueue <T, K>
 	ret.first->insert(cost, item, front);
       }
       unsigned get_cost() const {
+        if (empty())
```
athanatos: assert here as well...

src/common/WeightedPriorityQueue.h
```diff
@@ -84,6 +84,7 @@ class WeightedPriorityQueue :  public OpQueue <T, K>
       }
       //Get the cost of the next item to dequeue
       unsigned get_cost() const {
+        assert(!empty());
```
athanatos: spacing is wrong

src/common/WeightedPriorityQueue.h
```diff
@@ -212,7 +215,9 @@ class WeightedPriorityQueue :  public OpQueue <T, K>
       }
       void dump(ceph::Formatter *f) const {
 	f->dump_int("num_keys", next->get_size());
-	f->dump_int("first_item_cost", next->get_cost());
+        if (!empty()) {
```
athanatos: spacing looks wrong

xiexingguo:
@athanatos See again? Thanks!

src/common/WeightedPriorityQueue.h
```diff
@@ -84,7 +84,8 @@ class WeightedPriorityQueue :  public OpQueue <T, K>
       }
       //Get the cost of the next item to dequeue
       unsigned get_cost() const {
-	return lp.begin()->cost;
+        assert(!empty());
```
yuyuyu101: not related to this commit, I'd like to avoid "assert" in fast path in the future changes if not very necessary. Of course, any reasonable assert must be done. To be clarify again, I encourage use assert in necessary path.
@athanatos what do you think?

athanatos: This is the kind of assert which can be compiled out, and it's pretty minimal in terms of cost.  I don't feel any need to avoid this kind of assert.  Are you worried about the overhead?

yuriw:

PR #10615, opened 2016-08-08
rgw multisite: prevent notify and error_repo from duplicating bucket sync

cbodley:
The marker_tracker generally prevents us from starting two sync operations on the same bucket, but there are two cases where we start bucket sync without having a marker:
- in response to async datalog notifications, and
- when retrying buckets in the error_repo

This commit augments RGWDataSyncShardMarkerTrack so that these two cases can still register their sync operation, so that a later attempt knows not to start again. If, after starting, we get a valid marker from the log, we attach it to the sync in progress so that marker is written on completion.

This should help reduce contention on the sync status locks, which will in turn reduce the number of writes we make to the error_repo.

### DIFF
--- a/src/rgw/rgw_data_sync.cc
+++ b/src/rgw/rgw_data_sync.cc
@@ -1116,7 +1116,7 @@ public:
         }
       }
       /* FIXME: what do do in case of error */
-      if (!marker_handle->marker.empty()) {
+      if (sync_status == 0 && !marker_handle->marker.empty()) {
         /* update marker */
         yield call(marker_tracker->finish(marker_handle->marker));
       }
--- a/src/rgw/rgw_data_sync.cc
+++ b/src/rgw/rgw_data_sync.cc
@@ -1158,7 +1158,6 @@ class RGWDataSyncShardCR : public RGWCoroutine {
 
   std::string next_marker;
   list<rgw_data_change_log_entry> log_entries;
-  list<rgw_data_change_log_entry>::iterator log_iter;
   bool truncated;
 
   ceph::mutex inc_lock = ceph::make_mutex("RGWDataSyncShardCR::inc_lock");
@@ -1170,8 +1169,6 @@ class RGWDataSyncShardCR : public RGWCoroutine {
 
   set<string> modified_shards;
   set<string> current_modified;
-
-  set<string>::iterator modified_iter;
 
   int total_entries;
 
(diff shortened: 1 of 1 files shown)


src/rgw/rgw_data_sync.cc
```diff
@@ -782,21 +778,44 @@ class RGWDataSyncShardMarkerTrack : public RGWSyncShardMarkerTrack<string, strin
 				 marker_oid, sync_marker);
   }
 
-  /*
-   * create index from key -> marker, and from marker -> key
-   * this is useful so that we can insure that we only have one
```
jiaopengju: ensure?

src/rgw/rgw_data_sync.cc
```diff
@@ -1227,26 +1257,12 @@ class RGWDataSyncShardCR : public RGWCoroutine {
...
-              if (spawned_keys.find(log_iter->entry.key) == spawned_keys.end()) {
-                spawned_keys.insert(log_iter->entry.key);
-                spawn(new RGWDataSyncSingleEntryCR(sync_env, log_iter->entry.key, log_iter->log_id, marker_tracker, error_repo, false), false);
-                if (retcode < 0) {
-                  stop_spawned_services();
-                  drain_all();
-                  return set_cr_error(retcode);
-                }
-              }
+              spawn(new RGWDataSyncSingleEntryCR(sync_env, log_iter->entry.key, marker_tracker, handle, error_repo, false), false);
```
yehudasa: @cbodley the effect of this I think is that earlier when we squashed entries, because we weren't yielding we didn't need to retry, now need to retry

yehudasa:
@cbodley overall looks much cleaner. How is this being tested?

src/rgw/rgw_data_sync.cc
cbodley: is this about `spawned_keys`? i don't understand that comment about yielding. it looked to me like the call to `marker_tracker->index_key_to_marker()` would have already caught any duplicate keys and flagged them for retry

cbodley:
@yehudasa the only testing it's seen so far is test_multi.py. i'll try adding an option for 'gateways per zone' to see how it behaves with some contention

src/rgw/rgw_data_sync.cc
yehudasa: @cbodley yes, but the difference is that if we squash multiple entries without yielding, for these entries there's no need for retry since we know the spawned entries (that were started earlier) hadn't started yet, so we're not racing. Now that we yield for each spawned entry, we need to set retry for the following entries.

cbodley: okay thanks, that makes sense in theory. but i don't see how `spawned_keys` would ever find a duplicate, because `marker_tracker->index_key_to_marker()` would have already returned false if it found that key in its `key_to_marker` map

i also don't think there's any optimization to be had here, because even if we do set `retry = true`, `RGWDataSyncSingleEntryCR` is resetting it to false before it starts the bucket sync:
```c++
marker_handle->retry = false;
call(new RGWRunBucketSyncCoroutine(sync_env, bs));
```
so we won't actually retry the bucket sync unless it's flagged for retry after the bucket sync has started

yehudasa: @cbodley you're probably right. The `index_key_to_marker` was probably a later addition. That been said, we are going to have lots of duplicates (every bucket change will appear at the data log at least twice),  so if we could squash some entries in a sane fashion we should do it.

cbodley: agreed. i didn't add any yields to this class (afaik), so the marker_tracker should still be squashing any duplicate keys that it reads from the same call to `RGWReadRemoteDataLogShardCR`

but as a future improvement, it would be nice if we could also avoid dropping and reacquiring the lease for retries

yehudasa:
@cbodley one thing that we lose here is that earlier we'd get by with less yields. This could affect cpu utilization.

cbodley:
closing for now

cbodley:
reopened/rebased for more testing. cc @cfanz

cbodley:
this branch is still passing tests as before, but i do see it triggering more of the retries in `RGWDataSyncSingleEntryCR` than before as @yehudasa pointed out in [prior review](https://github.com/ceph/ceph/pull/10615#discussion_r74270399). i think the right solution is to eliminate any yields between the crs spawned for errors/notifications and those spawned for datalog entries. doing this will be easier after the changes in https://github.com/ceph/ceph/pull/23596, so i'll hold off until that's merged

cbodley:
rebased on https://github.com/ceph/ceph/pull/23596 and reorganized to avoid yields

cfanz:
When writing many data into the same bucket in a short time, the timestamp of oldest data not synced reported by `radosgw-admin sync status` will keep the same until this bucket is fully synced. This will make an illusion that the sync is stuck. Maybe using map<time_t, string>  instead of string for Handle::marker can give us a chance to decide whether we should call flush() in the retry loop of RGWDataSyncSingleEntryCR.

src/rgw/rgw_data_sync.cc
```diff
@@ -1129,11 +1148,14 @@ class RGWDataSyncSingleEntryCR : public RGWCoroutine {
       if (sync_env->observer) {
         sync_env->observer->on_bucket_changed(bs.bucket.get_key());
       }
-      /* FIXME: what do do in case of error */
-      if (marker_tracker && !entry_marker.empty()) {
+      if (!marker_handle->marker.empty()) {
```
cfanz: Should we consider only update sync marker when sync_status==0?

cbodley: good question @cfanz, and i think that's essentially what this FIXME comment was asking

it would be technically correct to make this change as you suggest, so that when a bucket sync fails our datalog processing loop in `RGWDataSyncShardCR::incremental_sync()` would keep retrying that entry until it succeeded

however, that would prevent us from making progress on later datalog entries and increase the time it takes for changes in those later buckets to sync

we use sync leases so that multiple gateways can cooperate and share the work of sync, so EBUSY errors from bucket sync are quite common as you add more gateways to the zone. if one EBUSY error blocked all other gateways from advancing in the datalog, that would severely limit how many buckets these gateways could process in parallel

so we added the error_repo as an optimization here. just before this comment, we call `error_repo->append()` to write these failed bucket shards to the error_repo, and `RGWDataSyncShardCR::incremental_sync()` will continue to retry these error_repo entries until they succeed. that allows us to safely update the datalog marker here so we can continue making progress on later buckets in the meantime

does that make sense?

cbodley: @cfanz you're right on both points, thanks!

incremental_sync() will continue to make progress either way. i agree that change would improve the accuracy of 'sync status' - the only other drawback would be around restarts, where we'd have to re-process failed datalog entries - but that happens infrequently

and because of the datalog shard lease, i can't think of any way that other gateways would compete over leases for its bucket shards

would you like open a pr with that change and see how it works?

cbodley: sorry, i see that you already opened https://github.com/ceph/ceph/pull/22866 - and that depends on this pr to work correctly. have you tried testing the two in combination?

cfanz: Not so far. I will try on it.

cbodley:
rebased and passed rgw:multisite suite at http://pulpito.ceph.com/cbodley-2018-10-16_15:19:04-rgw:multisite-wip-rgw-data-sync-retry-distro-basic-smithi/

i folded in the commit from @cfanz following discussion

@yehudasa could you please take another look?

mattbenjamin:
unstale me

cbodley:
@yehudasa i pushed a rebase and will schedule another multisite suite. i'd like to make progress on this, because it could allow us to remove the per-bucket leases and rely only on the datalog shard's lease

yuriw:
[wip-yuri4-testing-2019-03-27-2009-luminous](https://trello.com/c/wBvzoctn/567-wip-yuri4-testing-2019-03-27-2009-luminous)

cbodley:
rebased. after running multisite tests (2 zones, 1 gateway each), i see no lease failures on bucket sync status:
```
~/ceph/build $ grep 'Device or resource busy' run/c1/out/osd.0.log --binary-files=text | grep bucket | wc -l
0
~/ceph/build $ grep 'Device or resource busy' run/c2/out/osd.0.log --binary-files=text | grep bucket | wc -l
0
```
running against master shows several from each zone:
```
~/ceph/build $ grep 'Device or resource busy' run/c1/out/osd.0.log --binary-files=text | grep bucket | wc -l
14
~/ceph/build $ grep 'Device or resource busy' run/c2/out/osd.0.log --binary-files=text | grep bucket | wc -l
15
```

cbodley:
rebased and still passing multisite tests. i still think this is important to avoid unnecessary stalls and retries from lease conflicts

fangyuxiangGL:
@cbodley 

Hi casey, could you share me some docs or links about how to lauch the multisite suites? thanks

cbodley:
hi @fangyuxiangGL, there's a doc in ceph/src/test/rgw/test_multi.md. you can launch the tests from your cmake build directory

fangyuxiangGL:
@cbodley thanks

cbodley:
rebased

src/rgw/rgw_data_sync.cc
```diff
@@ -1097,10 +1116,14 @@ class RGWDataSyncSingleEntryCR : public RGWCoroutine {
         }
       }
       /* FIXME: what do do in case of error */
-      if (marker_tracker && !entry_marker.empty()) {
+      if (sync_status == 0 && !marker_handle->marker.empty()) {
```
fangyuxiangGL: if errors happened，sync_status will < 0，and the entry marker will leave in finish_markers,
which lead the finish_markers hold lots of entries. And also the sync marker can't be updated untill radosgw restarted.

cbodley: good catch, i'll look into that

fangyuxiangGL: @cbodley Hi casey， do you have any fix idea for this issue?

fangyuxiangGL: @cbodley

mattbenjamin:
unstale

fangyuxiangGL:
@cbodley 

Hi, any progress about this issue?

fangyuxiangGL:
@cbodley 

Hi, any progress about this issue?

fangyuxiangGL:
@cbodley 

Hi, any progress about this issue?

cbodley:
closing in favor of https://github.com/ceph/ceph/pull/34094

PR #10616, opened 2016-08-08
rpm: pass --disable-static to configure

smithfarm:
This makes the build consume less resources and we can drop the ancient
"rm -rf" lines from %install.

Fixes: http://tracker.ceph.com/issues/16960

### DIFF
--- a/ceph.spec.in
+++ b/ceph.spec.in
@@ -633,6 +633,7 @@ export RPM_OPT_FLAGS=`echo $RPM_OPT_FLAGS | sed -e 's/i386/i486/'`
                 --libexecdir=%{_libexecdir} \
 		--localstatedir=%{_localstatedir} \
 		--sysconfdir=%{_sysconfdir} \
+                --disable-static \
 %if 0%{?rhel} && ! 0%{?centos}
                 --enable-subman \
 %endif
@@ -684,7 +685,6 @@ make %{?_smp_mflags} check
 %install
 make DESTDIR=%{buildroot} install
 find %{buildroot} -type f -name "*.la" -exec rm -f {} ';'
-find %{buildroot} -type f -name "*.a" -exec rm -f {} ';'
 install -D src/etc-rbdmap %{buildroot}%{_sysconfdir}/ceph/rbdmap
 %if 0%{?fedora} || 0%{?rhel}
 install -m 0644 -D etc/sysconfig/ceph %{buildroot}%{_sysconfdir}/sysconfig/ceph


smithfarm:
pushed `wip-16960` to gitbuilders

smithfarm:
Note this is a jewel-only fix. The master/kraken specfile no longer contains any of this stuff.

smithfarm:
With `--disable-static`, only the following unpackaged files are found:

```
 /usr/lib64/ceph/compressor/libceph_example.la
/usr/lib64/ceph/compressor/libceph_snappy.la
/usr/lib64/ceph/compressor/libceph_zlib.la
/usr/lib64/ceph/erasure-code/libec_isa.la
/usr/lib64/ceph/erasure-code/libec_jerasure.la
/usr/lib64/ceph/erasure-code/libec_jerasure_generic.la
/usr/lib64/ceph/erasure-code/libec_jerasure_sse3.la
/usr/lib64/ceph/erasure-code/libec_jerasure_sse4.la
/usr/lib64/ceph/erasure-code/libec_lrc.la
/usr/lib64/ceph/erasure-code/libec_shec.la
/usr/lib64/ceph/erasure-code/libec_shec_generic.la
/usr/lib64/ceph/erasure-code/libec_shec_sse3.la
/usr/lib64/ceph/erasure-code/libec_shec_sse4.la
/usr/lib64/libcephfs.la
/usr/lib64/libcephfs_jni.la
/usr/lib64/libos_tp.la
/usr/lib64/libosd_tp.la
/usr/lib64/librados.la
/usr/lib64/librados_tp.la
/usr/lib64/libradosstriper.la
/usr/lib64/librbd.la
/usr/lib64/librbd_tp.la
/usr/lib64/librgw.la 
```

smithfarm:
@tchaikov @b-ranto Do you know of a better way to clean up these dangling .la files than `find %{buildroot} -type f -name "*.la" -exec rm -f {} ';'` in %install ?

tchaikov:
probably we can have a post-install hook for the plugins, like

```
install-exec-hook:
	-rm $(DESTDIR)$(pkglibdir)/*.a
	-rm $(DESTDIR)$(pkglibdir)/*.la
```

but i am not quite sure how to deal with the .la files of user facing libraries, as user **might** want link against them using .la files.

smithfarm:
@tchaikov No `.a` files are generated. So we could do:

```
install-exec-hook:
    -rm $(DESTDIR)$(pkglibdir)/ceph/*.la
```

and package these files:

```
/usr/lib64/libcephfs.la
/usr/lib64/libcephfs_jni.la
/usr/lib64/libos_tp.la
/usr/lib64/libosd_tp.la
/usr/lib64/librados.la
/usr/lib64/librados_tp.la
/usr/lib64/libradosstriper.la
/usr/lib64/librbd.la
/usr/lib64/librbd_tp.la
/usr/lib64/librgw.la 
```

Sound right to you?

smithfarm:
Hm, in debian we are packaging the user-facing '.a' and '.la' files (and I guess the `.a` files are only generated when `--disable-static` is **not** passed to configure). So in rpm it makes sense to package these user-facing `.la` files.

smithfarm:
And debian does not package libos or libosd, so these can be dropped as "non-user-facing".

b-ranto:
We should not package the static files unless there is a good reason to package them -- unless there is no other reasonable way to compile a program that needs to be linked with your library and it must be linked statically (we are usually talking about the programs in distribution repositories, here):

https://fedoraproject.org/wiki/Packaging:Guidelines#Packaging_Static_Libraries

In this case, I don't think there is too much point to package these static files. If you really wish to link statically for some reason then you can always use your own copy of ceph to link with. However, generally, static linking is discouraged for obvious reasons (ineffective updates, etc).

smithfarm:
We are packaging `.a` and `.la` files in debian, but not in rpm. Still, I guess we can't just drop them from the debian packaging in jewel.

tchaikov:
i am with @b-ranto.

> Sound right to you?

no, we should not package .la or .a files.

>> as user **might** want link against them using .la files.

i was talking about users which compile the distributed tarball directly instead of using the precompiled packages.

smithfarm:
@tchaikov So you would be in favor of dropping them?

```
smithfarm@wilbur:~/src/ceph/SUSE/ceph/debian> grep -Fr \.a
librbd-dev.install:usr/lib/librbd.a
librgw-dev.install:usr/lib/librgw.a
libcephfs-dev.install:usr/lib/libcephfs.a
libradosstriper-dev.install:usr/lib/libradosstriper.a
librados-dev.install:usr/lib/librados.a
smithfarm@wilbur:~/src/ceph/SUSE/ceph/debian> grep -Fr \.la
librbd-dev.install:usr/lib/librbd.la
librgw-dev.install:usr/lib/librgw.la
libcephfs-dev.install:usr/lib/libcephfs.la
rules:  sed -i "/dependency_libs/ s/'.*'/''/" `find . -name '*.la'`
libradosstriper-dev.install:usr/lib/libradosstriper.la
librados-dev.install:usr/lib/librados.la
```

smithfarm:
@tchaikov We can leave the `rm -rf` in the spec file. Debian, I guess, does not check for unpackaged files.

tchaikov:
> So you would be in favor of dropping them?

@smithfarm i am inclined to keep them. the idea is do not change the content of packaging in minor release changes.

smithfarm:
Cool, I'll close this then.

PR #10620, opened 2016-08-09
subman: use replace instead of format

dachary:
Otherwise all {} are assumed to be substituted. Add a test.

Signed-off-by: Loic Dachary <ldachary@redhat.com>

### DIFF
--- a/src/script/subman
+++ b/src/script/subman
@@ -1,6 +1,7 @@
-#!/usr/bin/env python -B
+#!/usr/bin/env python
 
 import json
+import os
 import re
 import subprocess
 
@@ -13,8 +14,9 @@ for disk in disks:
             df = subprocess.check_output("df --output=used " + partition['path'], shell=True)
             used += int(re.findall('\d+', df)[0])
 
-open("/etc/rhsm/facts/ceph_usage.facts", 'w').write("""
+facts_file = os.environ.get("CEPH_FACTS_FILE", "/etc/rhsm/facts/ceph_usage.facts")
+open(facts_file, 'w').write("""\
 {
 "band.storage.usage": {used}
 }
-""".format(used=used/(1024*1024*1024)))
+""".replace('{used}', str(int(used/(1024*1024*1024)))))
--- a/src/test/test_subman.sh
+++ b/src/test/test_subman.sh
@@ -1,0 +1,28 @@
+#!/bin/bash -e
+
+source $(dirname $0)/detect-build-env-vars.sh
+
+TMP=$(mktemp --tmpdir -d)
+trap "rm -fr $TMP" EXIT
+
+export PATH=$TMP:$PATH
+
+cat > $TMP/ceph-disk <<EOF
+echo '[{"partition":[{"type":"data","path":"/dev/foo/bar"}]}]'
+EOF
+chmod +x $TMP/ceph-disk
+
+cat > $TMP/df <<EOF
+echo Used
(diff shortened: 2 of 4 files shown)


src/script/subman
```diff
@@ -1,6 +1,7 @@
-#!/usr/bin/env python -B
+#!/usr/bin/env python
```
tchaikov: why `-B` is removed?

src/test/test_subman.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash -e
+
+TMP=/tmp/subman$$
```
tchaikov: maybe could use `mktemp --tmpdir -d` ? so we can respect the `$TMPDIR`, which can in turn be overrided by user the make system.

src/test/test_subman.sh
```diff
@@ -0,0 +1,27 @@
...
+cat > $TMP/ceph-disk <<EOF
+echo '[{"partition":[{"type":"data","path":"/dev/foo/bar"}]}]'
+EOF
+chmod +x $TMP/ceph-disk
+
+cat > $TMP/df <<EOF
+echo Used
+echo $((2 * 1024 * 1024 * 1024))
+EOF
+chmod +x $TMP/df
```
tchaikov: why `check_process()` will use the `$TMP/df` instead of the `/bin/df`?

src/test/test_subman.sh
```diff
@@ -0,0 +1,27 @@
...
+
+cat > $TMP/expected <<EOF
+{
+"band.storage.usage": 2
+}
+EOF
+export CEPH_FACTS_FILE=$TMP/facts
+script/subman
```
tchaikov: might want to use `detect-build-env-vars.sh` to initialize the env vars, and use the absolute path instead.

src/script/subman
```diff
@@ -13,8 +14,9 @@ for disk in disks:
...
             used += int(re.findall('\d+', df)[0])
 
-open("/etc/rhsm/facts/ceph_usage.facts", 'w').write("""
+facts_file = os.environ.get("CEPH_FACTS_FILE", "/etc/rhsm/facts/ceph_usage.facts")
+open(facts_file, 'w').write("""\
 {
 "band.storage.usage": {used}
 }
-""".format(used=used/(1024*1024*1024)))
+""".replace('{used}', str(int(used/(1024*1024*1024)))))
```
tchaikov: nit, could also use `{{` and `}}` to escape the substitution.

src/script/subman
dachary: It was meant to ease packaging but turned out to not be used.

src/script/subman
dachary: wow, I completely missed that trick, thanks ! I guess it would be even more confusing in this case but I'll keep it in mind for future use :-)

src/test/test_subman.sh
dachary: of course ! done

src/test/test_subman.sh
dachary: because of the export PATH=$TMP:$PATH above

src/test/test_subman.sh
dachary: done

src/test/test_subman.sh
tchaikov: ahh, right!

tchaikov:
lgtm. as jenkins is also happy.

PR #10627, opened 2016-08-09
rbd-mirror: replaying state should include flush action

dillaman:
Fixes: http://tracker.ceph.com/issues/16970


trociny:
lgtm

PR #10628, opened 2016-08-09
mon: tear down standby replays on MDS rank stop

jcsp:
Fixes: http://tracker.ceph.com/issues/16909


jcsp:
Tested by https://github.com/ceph/ceph-qa-suite/pull/1112

gregsfortytwo:
LGTM

PR #10629, opened 2016-08-09
doc: update rbd glance configuration notes

dillaman:
Starting with the Kilo release of OpenStack, glance drivers are now fully configured within the "[glance_store]" section.

### DIFF
--- a/doc/rbd/rbd-openstack.rst
+++ b/doc/rbd/rbd-openstack.rst
@@ -215,11 +215,22 @@ Edit ``/etc/glance/glance-api.conf`` and add under the ``[glance_store]`` sectio
     rbd_store_ceph_conf = /etc/ceph/ceph.conf
     rbd_store_chunk_size = 8
 
-
-For more information about the configuration options available in Glance please see: http://docs.openstack.org/trunk/config-reference/content/section_glance-api.conf.html.
-
 .. important:: Glance has not completely moved to 'store' yet.
-    So we still need to configure the store in the DEFAULT section.
+    So we still need to configure the store in the DEFAULT section until Kilo.
+
+Kilo
+~~~~
+Edit ``/etc/glance/glance-api.conf`` and add under the ``[glance_store]`` section::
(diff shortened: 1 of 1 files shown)


doc/rbd/rbd-openstack.rst
```diff
@@ -215,11 +215,22 @@ Edit ``/etc/glance/glance-api.conf`` and add under the ``[glance_store]`` sectio
...
-    So we still need to configure the store in the DEFAULT section.
+    [glance_store]
+    stores = rbd
+    default_store = rbd
+    rbd_store_pool = images
+    rbd_store_user = glance
+    rbd_store_ceph_conf = /etc/ceph/ceph.conf
+    rbd_store_chunk_size = 8
+
+For more information about the configuration options available in Glance please see: http://docs.openstack.org/trunk/config-reference/content/section_glance-api.conf.html.
```
trociny: May be update this URL too? It looks like it is outdated (redirecting to http://docs.openstack.org/mitaka/config-reference/index.html).

For Kilo I see documentation:

http://docs.openstack.org/kilo/config-reference/content/ch_configuring-openstack-image-service.html

specifically:

http://docs.openstack.org/kilo/config-reference/content/configuring-image-service-backends.html

The latest (Mitaka):

http://docs.openstack.org/mitaka/config-reference/image-service.html
http://docs.openstack.org/mitaka/config-reference/image-service/backends.html

dillaman: The nice thing about the redirect is that we don't need to keep updating it for each OpenStack release.

trociny: If it had redirected to right link, though...

dillaman: Sadly they took away the ability to link to the most recent docs it appears, so I'd argue removing the link over linking to a specific version that will be out-of-date every 6 months.

trociny:
lgtm

PR #10630, opened 2016-08-09
rgw: Added new counters for monitoring http status

prashant-kr312:
Signed-off-by: Prashant Kumar prashant.kr@flipkart.com

### DIFF
--- a/src/rgw/librgw.cc
+++ b/src/rgw/librgw.cc
@@ -293,6 +293,25 @@ namespace rgw {
     int http_ret = s->err.http_ret;
 
     req->log_format(s, "http status=%d", http_ret);
+    switch (http_ret / 100) {
+      case 1:
+        perfcounter->inc(l_rgw_http_1xx_count);
+        break;
+      case 2:
+        perfcounter->inc(l_rgw_http_2xx_count);
+        break;
+      case 3:
+        perfcounter->inc(l_rgw_http_3xx_count);
+        break;
+      case 4:
+        perfcounter->inc(l_rgw_http_4xx_count);
+        break;
+      case 5:
+        perfcounter->inc(l_rgw_http_5xx_count);
+        break;
+      default:
+        perfcounter->inc(l_rgw_http_NULL_count);
+    }
 
     dout(1) << "====== " << __func__
 	    << " req done req=" << hex << req << dec << " http_status="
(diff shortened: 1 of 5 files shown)


prashant-kr312:
We are capturing req, failed_req, get and put requests from every RGW node. It would be nice to have continuously increasing HTTP counters for HTTP status. This gives more insight into RGW and should help in serviceability enhancement.
Tracker issue: http://tracker.ceph.com/issues/16971

oritwas:
@prashant-kr312 , please rebase

src/rgw/librgw.cc
```diff
@@ -293,6 +293,25 @@ namespace rgw {
     int http_ret = s->err.http_ret;
 
     req->log_format(s, "http status=%d", http_ret);
+    switch (http_ret / 100) {
```
amitkumar50: @prashant-kr312 can't we use a dictionary that maps tag names to a tag handler to replace traditional case statements..

src/rgw/rgw_process.cc
```diff
@@ -199,6 +199,25 @@ int process_request(RGWRados* store, RGWREST* rest, RGWRequest* req,
 
   req->log_format(s, "op status=%d", op_ret);
   req->log_format(s, "http status=%d", http_ret);
+  switch (http_ret / 100) {
```
amitkumar50: @prashant-kr312 same

src/rgw/librgw.cc
amitkumar50: @prashant-kr312 Can you prefix Commit title with "rgw".
>It is required to prefix it with the subsystem or module you are changing.

mattbenjamin:
@cbodley is there a problem with adding more perfcounters?  seems potentially useful

cbodley:
@mattbenjamin nope, this looks useful

src/rgw/rgw_process.cc
```diff
@@ -199,6 +199,25 @@ int process_request(RGWRados* store, RGWREST* rest, RGWRequest* req,
...
+      break;
+    case 4:
+      perfcounter->inc(l_rgw_http_4xx_count);
+      break;
+    case 5:
+      perfcounter->inc(l_rgw_http_5xx_count);
+      break;
+    default:
+      perfcounter->inc(l_rgw_http_NULL_count);
+  }
```
cbodley: this could be a helper function in rgw_rest.cc so it doesn't get duplicated in librgw.cc

PR #10631, opened 2016-08-09
fix civetweb https - autobuild

kmroz:
http://tracker.ceph.com/issues/16957


cbodley:
the autotools build was removed from master in favor of cmake. i'm guessing that we'll need to target fixes like this directly to the jewel branch

kmroz:
@cbodley - Sorry for the late reply. I can certainly open this PR against the jewel branch.

kmroz:
Closing. Re-opened against ceph/jewel: https://github.com/ceph/ceph/pull/11689

PR #10632, opened 2016-08-09
FIX: use ceph::buffer::list instead of buffer::list

ryneli:
use ceph::buffer::list instead of buffer::list so that things outside ceph can use librados.hpp

### DIFF
--- a/src/include/rados/librados.hpp
+++ b/src/include/rados/librados.hpp
@@ -393,7 +393,7 @@ namespace librados
     void zero(uint64_t off, uint64_t len);
     void rmxattr(const char *name);
     void setxattr(const char *name, const bufferlist& bl);
-    void setxattr(const char *name, const buffer::list&& bl);
+    void setxattr(const char *name, const bufferlist&& bl);
     void tmap_update(const bufferlist& cmdbl);
     void tmap_put(const bufferlist& bl);
     void clone_range(uint64_t dst_off,


src/include/rados/librados.hpp
```diff
@@ -393,7 +393,7 @@ namespace librados
     void zero(uint64_t off, uint64_t len);
     void rmxattr(const char *name);
     void setxattr(const char *name, const bufferlist& bl);
-    void setxattr(const char *name, const buffer::list&& bl);
+    void setxattr(const char *name, const ceph::buffer::list&& bl);
```
jdurgin: I was looking quickly this morning and forgot we used the typedef 'bufferlist' throughout - let's use that here too for consistency

jdurgin:
could you prefix the commit message with the subsystem modified (e.g. librados: use bufferlist instead of buffer::list) ? Also add your signed-off-by (git commit -s includes it for you).

ryneli:
Hi @jdurgin, is it ok now?

jdurgin:
Could you add the signed-off-by to the commit message (git commit --amend -s)

jdurgin:
thanks! looks good to me

ryneli:
Hi @jdurgin, could you merge this PR now? After that, I could test my script with latest version of ceph. Thank you so much.

PR #10636, opened 2016-08-09
cephfs: Permit recovering metadata into a new RADOS pool

fullerdj:
Add a procedure that permits reconstructing metadata in a potentially
damaged cephfs metadata pool and writing the results into a
freshly-initialized pool that refers to the same data pool. Add option
flags to override checks that would ordinarily prevent this and add
options to the recovery tools to write output to a separate pool instead of
the one selected for recovery. See docs/cephfs/disaster-recovery.rst for
details.

cf. https://github.com/fullerdj/ceph-qa-suite/tree/wip-djf-15069

Fixes: http://tracker.ceph.com/issues/15068
Fixes: http://tracker.ceph.com/issues/15069
Signed-off-by: Douglas Fuller dfuller@redhat.com

### DIFF
--- a/doc/cephfs/disaster-recovery.rst
+++ b/doc/cephfs/disaster-recovery.rst
@@ -216,3 +216,62 @@ Note that this command acts as a normal CephFS client to find all the
 files in the filesystem and read their layouts, so the MDS must be
 up and running.
 
+Using an alternate metadata pool for recovery
+---------------------------------------------
+
+.. warning::
+
+   There has not been extensive testing of this procedure. It should be
+   undertaken with great care.
+
+If an existing filesystem is damaged and inoperative, it is possible to create
+a fresh metadata pool and attempt to reconstruct the filesystem metadata
+into this new pool, leaving the old metadata in place. This could be used to
+make a safer attempt at recovery since the existing metadata pool would not be
+overwritten.
+
+.. caution::
+
+   During this process, multiple metadata pools will contain data referring to
+   the same data pool. Extreme caution must be exercised to avoid changing the
+   data pool contents while this is the case. Once recovery is complete, the
+   damaged metadata pool should be deleted.
+
+To begin this process, first create the fresh metadata pool and initialize
+it with empty file system data structures:
+
+::
+
+    ceph fs flag set enable_multiple true --yes-i-really-mean-it
+    ceph osd pool create recovery <pg-num> replicated <crush-ruleset-name>
+    ceph fs new recovery-fs recovery <data pool> --allow-dangerous-metadata-overlay
+    cephfs-data-scan init --force-init --filesystem recovery-fs --alternate-pool recovery
+    ceph fs reset recovery-fs --yes-i-realy-mean-it
+    cephfs-table-tool recovery-fs:all reset session
+    cephfs-table-tool recovery-fs:all reset snap
+    cephfs-table-tool recovery-fs:all reset inode
+
+Next, run the recovery toolset using the --alternate-pool argument to output
+results to the alternate pool:
+
+::
+
+    cephfs-data-scan scan_extents --alternate-pool recovery --filesystem <original filesystem name>
+    cephfs-data-scan scan_inodes --alternate-pool recovery --filesystem <original filesystem name> --force-corrupt --force-init <original data pool name>
+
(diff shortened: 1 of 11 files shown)


doc/cephfs/disaster-recovery.rst
```diff
@@ -160,3 +160,63 @@ the range 0-(N_workers - 1), like so:
...
+   data pool contents while this is the case. Once recovery is complete, the
+   damaged metadata pool should be deleted.
+
+To begin this process, first create the fresh metadata pool and initialize
+it with empty file system data structures:
+
+::
+
+    ceph fs flag set enable_multiple true --yes-i-really-mean-it
+    rados mkpool recovery
```
jcsp: Let's not encourage people to use "rados mkpool", "ceph osd pool create" is nicer and forces them to choose sane PG counts

doc/cephfs/disaster-recovery.rst
```diff
@@ -160,3 +160,63 @@ the range 0-(N_workers - 1), like so:
...
+If the damaged filesystem contains dirty journal data, it may be recovered next
+with:
+
+::
+
+    cephfs-journal-tool --rank=<original filesystem name>:0 event recover_dentries list --alternate-pool recovery
+    cephfs-journal-tool --rank recovery-fs:0 journal reset --force
+
+After recovery, some recovered directories will have incorrect link counts. Set
+mds_debug_scatterstat=false to prevent the MDS from checking the link counts,
```
jcsp: mds_debug_scatterstat is false by default, so we should say something like "ensure mds_debug_scatterstat is set to false (the default)" to avoid confusing people who have never heard of the setting

doc/cephfs/disaster-recovery.rst
```diff
@@ -160,3 +160,63 @@ the range 0-(N_workers - 1), like so:
...
+then run a forward scrub to repair them.
+
+::
+
+    ceph-mds -i a --mds_debug_scatterstat=false
```
jcsp: In general MDSs will still be running even if the system is damaged, so let's just note here that we assume that you have a running MDS and should start one if it isn't already running, and use that MDS's ID in the subsequent scrub_path daemon command

src/mon/MDSMonitor.cc
```diff
@@ -1608,6 +1608,19 @@ int MDSMonitor::management_command(
       return -EINVAL;
     }
 
+    for (auto fs : pending_fsmap.get_filesystems()) {
+      const set<int64_t>& data_pools = fs.second->mds_map.get_data_pools();
+      string sure;
+      if ((data_pools.find(data) != data_pools.end()
+	   || fs.second->mds_map.metadata_pool == metadata)
+	  && ((!cmd_getval(g_ceph_context, cmdmap, "allow_overlay", sure)
+	       || sure != "--allow-dangerous-metadata-overlay"))) {
```
jcsp: My original thought was that we would only permit this if one of the filesystems is marked damaged, so that there was no way for the user to have two running MDSs pointing at the same pool.  We should probably still do that, or at least enforce that cluster_down is set for one of the filesystems at all times -- users are careless with the --this-is-so-dangerous-i-cant-even flags.

jcsp:
I'm seeing failures on TestJournalRepair.test_inject_to_empty, and TestDataScan.test_stashed_layout that are probably attributable to this PR.
http://pulpito.ceph.com/jspray-2016-09-05_10:01:52-fs-wip-jcsp-testing-20160902-testing-basic-mira/400996
http://pulpito.ceph.com/jspray-2016-09-05_10:01:52-fs-wip-jcsp-testing-20160902-testing-basic-mira/400938

fullerdj:
Fixed, thanks.

jcsp:
This one's still failing:
http://pulpito.ceph.com/jspray-2016-09-07_23:07:40-fs-wip-jcsp-testing-20160907-testing-basic-mira/405007/

jcsp:
@fullerdj can you look into that test failure?  Is that test passing locally for you?

fullerdj:
Okay, I've updated the ceph-qa-suite PR. It passes for me locally either way, so I'll try to fire off a teuthology run to verify.

jcsp:
@fullerdj could you revive this one please?

fullerdj:
Sure, I’ll have a look. Sorry about that.

PR #10637, opened 2016-08-09
librbd: Zipkin tracing [GSOC 2016]

vears91:
Pass trace information from librbd to librados. This is part of the end-to-end performance visualization Google Summer of Code project, to initialize traces in fio rbd engine and pass them to librados. It is built on the wip-blkin-osdc branch that added Zipkin tracing support to Messenger and Objecter.

Trace information is passed to librbd through new rbd_aio_write_traced and rbd_aio_read_traced functions, that take trace information as one of the parameters. 

Signed-off-by: Victor Araujo <ve.ar91@gmail.com>

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -33,3 +33,6 @@
 [submodule "src/isa-l"]
 	path = src/isa-l
 	url = https://github.com/ceph/isa-l
+[submodule "src/blkin"]
+	path = src/blkin
+	url = https://github.com/linuxbox2/blkin
--- a/src/include/ceph_features.h
+++ b/src/include/ceph_features.h
@@ -76,6 +76,7 @@
 #define CEPH_FEATURE_OSD_HITSET_GMT (1ULL<<54)
 #define CEPH_FEATURE_HAMMER_0_94_4 (1ULL<<55)
 #define CEPH_FEATURE_NEW_OSDOP_ENCODING   (1ULL<<56) /* New, v7 encoding */
+#define CEPH_FEATURE_BLKIN_TRACING (1ULL<<57) // enabled by ifdef, don't overlap
 #define CEPH_FEATURE_MON_STATEFUL_SUB (1ULL<<57) /* stateful mon subscription */
 #define CEPH_FEATURE_MON_ROUTE_OSDMAP (1ULL<<57) /* peon sends osdmaps */
 #define CEPH_FEATURE_OSDSUBOP_NO_SNAPCONTEXT (1ULL<<57) /* overlap, drop unused SnapContext in v12 */
@@ -110,6 +111,13 @@ static inline unsigned long long ceph_sanitize_features(unsigned long long f) {
 		return f;
 	}
 }
+
+// conditionally include blkin in CEPH_FEATURES_ALL/SUPPORTED_DEFAULT
+#ifdef WITH_BLKIN
+#define CEPH_FEATURES_BLKIN CEPH_FEATURE_BLKIN_TRACING
+#else
+#define CEPH_FEATURES_BLKIN 0
+#endif
 
 /*
  * Features supported.  Should be everything above.
@@ -174,6 +182,7 @@ static inline unsigned long long ceph_sanitize_features(unsigned long long f) {
          CEPH_FEATURE_OSD_PROXY_WRITE_FEATURES |         \
 	 CEPH_FEATURE_OSD_HITSET_GMT |			 \
 	 CEPH_FEATURE_HAMMER_0_94_4 |		 \
+	 CEPH_FEATURES_BLKIN | \
 	 CEPH_FEATURE_MON_STATEFUL_SUB |	 \
 	 CEPH_FEATURE_MON_ROUTE_OSDMAP |	 \
 	 CEPH_FEATURE_CRUSH_TUNABLES5 |	    \
--- a/src/include/ceph_features.h
+++ b/src/include/ceph_features.h
@@ -76,7 +76,6 @@
 #define CEPH_FEATURE_OSD_HITSET_GMT (1ULL<<54)
 #define CEPH_FEATURE_HAMMER_0_94_4 (1ULL<<55)
 #define CEPH_FEATURE_NEW_OSDOP_ENCODING   (1ULL<<56) /* New, v7 encoding */
-#define CEPH_FEATURE_BLKIN_TRACING (1ULL<<57) // enabled by ifdef, don't overlap
 #define CEPH_FEATURE_MON_STATEFUL_SUB (1ULL<<57) /* stateful mon subscription */
 #define CEPH_FEATURE_MON_ROUTE_OSDMAP (1ULL<<57) /* peon sends osdmaps */
 #define CEPH_FEATURE_OSDSUBOP_NO_SNAPCONTEXT (1ULL<<57) /* overlap, drop unused SnapContext in v12 */
@@ -85,6 +84,7 @@
 // duplicated since it was introduced at the same time as CEPH_FEATURE_CRUSH_TUNABLES5
 #define CEPH_FEATURE_NEW_OSDOPREPLY_ENCODING (1ULL<<58) /* New, v7 encoding */
 #define CEPH_FEATURE_FS_FILE_LAYOUT_V2       (1ULL<<58) /* file_layout_t */
+#define CEPH_FEATURE_BLKIN_TRACING (1ULL<<59) // don't overlap
 
 #define CEPH_FEATURE_RESERVED2 (1ULL<<61)  /* slow down, we are almost out... */
 #define CEPH_FEATURE_RESERVED  (1ULL<<62)  /* DO NOT USE THIS ... last bit! */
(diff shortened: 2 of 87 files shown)


alimaredia:
@dillaman

alimaredia:
5b5d44368ade3633693016469d2073a224e95b95 might be worth removing. The only part of it that could be useful is the .rst file that gets created, everything else should be removed.

src/include/ceph_features.h
```diff
@@ -76,6 +76,7 @@
 #define CEPH_FEATURE_OSD_HITSET_GMT (1ULL<<54)
 #define CEPH_FEATURE_HAMMER_0_94_4 (1ULL<<55)
 #define CEPH_FEATURE_NEW_OSDOP_ENCODING   (1ULL<<56) /* New, v7 encoding */
+#define CEPH_FEATURE_BLKIN_TRACING (1ULL<<57) // enabled by ifdef, don't overlap
```
cbodley: this was intended to have its own feature bit, rather than share a bit with other flags. the idea was to allow endpoints to interoperate if one was built with blkin support, and the other was not. but i'm not sure that's actually useful - if we're including blkin as a submodule, we might as well enable and package it by default

so i think we should get rid of the conditional `CEPH_FEATURES_BLKIN` stuff below, and just update the value of this bit. i'm guessing that we want to overlap with the KRAKEN bit (can @athanatos confirm?):
```
#define CEPH_FEATURE_BLKIN_TRACING CEPH_FEATURE_SERVER_KRAKEN
```

cbodley:
> 5b5d443 might be worth removing

i'd prefer to keep it, at least in some form, so the original author/sign-offs remain in the git history

.gitmodules
```diff
@@ -33,3 +33,6 @@
 [submodule "src/isa-l"]
 	path = src/isa-l
 	url = https://github.com/ceph/isa-l
+[submodule "src/blkin"]
+	path = src/blkin
+	url = https://github.com/linuxbox2/blkin
```
jdurgin: should we move this under the ceph github org? I can create the repo there

src/include/ceph_features.h
jdurgin: We can include this with the kraken feature bit if blkin will always be enabled. However, if we make blkin optional in the build there are potential issues with upgrades when some endpoints have blkin compiled in and others do not. E.g. a post-kraken osd without blkin and a kraken osd with blkin - one sends no tracing info, the other assumes tracing info is there based on the KRAKEN bit and can't decode the non-blkin message.

Keeping it as a separate feature bit that's only turned on in connections where it's compiled in, like this diff shows, seems to avoid this kind of problem.

src/include/rados/librados.h
```diff
@@ -1902,6 +1902,33 @@ CEPH_RADOS_API int rados_aio_write(rados_ioctx_t io, const char *oid,
...
+ * @param oid name of the object
+ * @param completion what to do when the write is safe and complete
+ * @param buf data to write
+ * @param len length of the data, in bytes
+ * @param off byte offset in the object to begin writing at
+ * @param info blkin trace information
+ * @returns 0 on success, -EROFS if the io context specifies a snap_seq
+ * other than LIBRADOS_SNAP_HEAD
+ */
+ struct blkin_trace_info;
```
jdurgin: does this compile when blkin isn't configured?

src/include/rados/librados.h
```diff
@@ -2007,6 +2034,34 @@ CEPH_RADOS_API int rados_aio_read(rados_ioctx_t io, const char *oid,
...
+ * @param io the context in which to perform the read
+ * @param oid the name of the object to read from
+ * @param completion what to do when the read is complete
+ * @param buf where to store the results
+ * @param len the number of bytes to read
+ * @param off the offset to start reading from in the object
+ * @param info blkin trace information
+ * @returns 0 on success, negative error code on failure
+ */
+CEPH_RADOS_API int rados_aio_read_traced(rados_ioctx_t io,
```
jdurgin: it'd be great to have tests using these new librados and librbd apis in test/librados/io.cc and test/librbd/test_librbd.cc (good place to show example blkin usage too)

src/include/rados/librados.h
vears91: It does, I have tested it without configuring blkin.

src/include/ceph_features.h
cbodley: @jdurgin okay, thanks!

@vears91 could you please update the value of `CEPH_FEATURE_BLKIN_TRACING` here to the first unused bit you can find?

.gitmodules
jdurgin: https://github.com/ceph/blkin and https://github.com/ceph/babeltrace-zipkin are available now

src/include/rados/librados.hpp
```diff
@@ -1003,7 +1003,7 @@ namespace librados
      */
     int aio_operate(const std::string& oid, AioCompletion *c,
 		    ObjectWriteOperation *op, snap_t seq,
-		    std::vector<snap_t>& snaps);
+		    std::vector<snap_t>& snaps, const blkin_trace_info *trace_info = nullptr);
```
dillaman: This breaks the current API/ABI (e.g. an older app linked against this would not have put a null pointer on the stack).

dillaman:
I am having trouble understanding the life cycle of ```blkin_trace_info```.  I am assuming it's the caller's responsibility to allocate one of these objects, but when would it get freed? In the librbd case with the client-side cache enabled, you cannot free it when the read/write's associated ```AioCompletion``` fires since it might still be held in ```ObjectCacher```.  You most likely would need to implement reference counting.

src/include/rados/librados.hpp
vears91: Could it be fixed by adding an aio_operate_traced function in librados, and calling it from AioObjectRequest if there is trace information, or else defaulting to the regular aio_operate?

vears91: Could it be fixed by adding an aio_operate_traced function in librados, and calling it from AioObjectRequest if there is trace information, or else defaulting to the regular aio_operate?

dillaman: Since ```aio_operate``` is already overloaded many times over, you could just add a new overload which librbd would always call.

cbodley:
@dillaman `struct blkin_trace_info` is just 3 integers, and should be stack allocated and copied, rather than allocated on the heap. the same goes for `struct blkin_trace` and `class Trace`

dillaman:
@cbodley (1) right now it is being passed around as a pointer and (2) if it should be passed by value would we ever need to worry about the ```blkin_trace_info``` structure since it will now be part of the librados/librbd API as pass by value?

src/os/filestore/FileJournal.cc
```diff
@@ -1681,6 +1686,12 @@ void FileJournal::submit_entry(uint64_t seq, bufferlist& e, uint32_t orig_len,
     logger->inc(l_os_j_bytes, e.length());
   }
 
+  if (osd_op) {
+    osd_op->mark_event("commit_queued_for_journal_write");
+    osd_op->journal_trace.init("journal", &trace_endpoint, &osd_op->store_trace);
```
cbodley: this `journal_trace.init()` should be conditional on `if (osd_op->store_trace)` instead of just `if (osd_op)`, or we'll create separate traces that aren't attached to a parent

(the call to `osd_op->mark_event()` should remain outside of `if (osd_op->store_trace)`)

src/librbd/AioObjectRequest.h
```diff
@@ -93,6 +96,7 @@ class AioObjectRequest : public AioObjectRequestHandle {
   Context *m_completion;
   Extents m_parent_extents;
   bool m_hide_enoent;
+  const blkin_trace_info *m_trace_info;
```
cbodley: as @dillaman pointed out, it's not safe to hold on to the trace info pointers like this

take a look at how Objecter and Op work for an example; they pass optional traces around by `ZTracer::Trace*`, and store them by value as `ZTracer::Trace`. we should only be using `blkin_trace_info` in our public interface (anything in src/include/rados/), then turning it into a `Trace` object as soon as we can

vears91: I wondered why ZTracer::Trace wasn't being used but I was told to pass the trace info to send_write_op. I will try to change it, can you please explain more about the issue with the trace info pointers? What problems can it cause?

cbodley: looking at your `aio_write_test_data_traced()` test as an example, it's starting the rbd write and waiting for the completion at the same time:
```c++
void aio_write_test_data_traced(...) {
  struct blkin_trace_info trace_info;
  rbd_aio_write_traced(image, off, len, test_data, comp, &trace_info);
  rbd_aio_wait_for_complete(comp);
```
we're passing a pointer to a `blkin_trace_info` that's on the stack here. but it's not a problem, because the stack stays valid until the operation completes - this operation is basically synchronous. but consider an asynchronous case, where we start the operation in one function, and wait for it later in another:
```c++
void start_write(...) {
  struct blkin_trace_info trace_info;
  rbd_aio_write_traced(image, off, len, test_data, comp, &trace_info);
}
void wait_for_write(...) {
  rbd_aio_wait_for_complete(comp);
}
```
again we pass a pointer to `blkin_trace_info` on the stack, but this time it goes out of scope right after it starts the write. so if rbd holds on to this pointer and tries it dereference it later, it won't be pointing to valid memory. that's why we need to store a copy of the trace info instead of just the pointer

PR #10638, opened 2016-08-09
rgw file: remove busy-wait in RGWLibFS::gc()

mattbenjamin:
This is a background thread.  However, CPU is wasted.

Fixes tracker issue #16976

### DIFF
--- a/src/rgw/rgw_file.cc
+++ b/src/rgw/rgw_file.cc
@@ -679,6 +679,7 @@ namespace rgw {
 	  } /* rgw_fh */
 	} /* event::type::READDIR */
       } /* ev */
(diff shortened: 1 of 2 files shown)


src/rgw/rgw_file.cc
```diff
@@ -679,6 +679,7 @@ namespace rgw {
 	  } /* rgw_fh */
 	} /* event::type::READDIR */
       } /* ev */
+      std::this_thread::sleep_for(std::chrono::seconds(120));
```
yehudasa: @mattbenjamin maybe make the 120 configurable, or at least go through #define?

mattbenjamin: @yehudasa ack

PR #10640, opened 2016-08-10
libcephfs: Fix the incorrect integer conversion in libcephfs_jni.cc

vmcloud:
The force conversion from jlong to int will cause the integer overflow, which will make the following logic go wrong. In my local test, this will always lead EFBIG error, while reading or writing a large file.


gregsfortytwo:
@noahdesu, any thoughts?

noahdesu:
lgtm. good catch @vmcloud i'm surprised this wasn't seen before given that it looks like it would trigger on anything bigger than 2gb.

vmcloud:
Yes, this code belongs to libcephfs_jni used by cephfs-java, which is a less popular scenario in ceph. @noahdesu

PR #10642, opened 2016-08-10
rgw/rgw_main.cc: add detailed error message if safe_read_exact() fails

atheism:
Add more details to error message if safe_read_exact() fails.


atheism:
close pull requests for different commits and break them into new separated ones.

PR #10649, opened 2016-08-10
jewel: mkfs.xfs slow performance with discards and object map

dachary:
http://tracker.ceph.com/issues/16867


dachary:
@dillaman it passed a rbd suite run modulo (I think ;-) known false negatives http://tracker.ceph.com/issues/16904#note-5. Good to merge ?

dillaman:
:+1:

PR #10495, opened 2016-07-29
DNM: jewel : 60-ceph-partuuid-workaround-rules still needed by debian jessie (udev 215-17)

Abhishekvrshny:
http://tracker.ceph.com/issues/16577

PR #10654, opened 2016-08-10
jewel: mon: "mon metadata" fails when only one monitor exists

dachary:
http://tracker.ceph.com/issues/16081


dachary:
@jcsp this is the continuation of https://github.com/ceph/ceph/pull/10495 (with @Abhishekvrshny permission ;-). Does it look ok ?

jcsp:
Looks fine to me

PR #10665, opened 2016-08-10
os/bluestore: fix locking for finish_write

liewegas:
In 0633c710584d34261a6164445939d23414bc4799 we adding
locking (which we need), but used first_collection,
which may not be the right collection for the given
onode.  Add an onode->collection pointer so that we can
lock the correct thing.


liewegas:
https://github.com/ceph/ceph/pull/10215 is a better fix

PR #10666, opened 2016-08-10
cmake: install platlib into a subdir of build-base dir

tchaikov:
distutils' install command installs all stuff under $build-base, which
includes the temporary files. and rpm packaging does not like::
  Installed (but unpackaged) file(s)
and to support both py2 and py3, we should encode the python version
into the path, otherwise the built files will be overwritten.


tchaikov:
should wait until gb are happy. i failed to do so last time...

jdurgin:
jenkins failed - perhaps need to adjust AddCephTest again

tchaikov:
@jdurgin thanks. fixed and repushed

jdurgin:
lgtm if the gitbuilders pass

tchaikov:
they are happy.

PR #10667, opened 2016-08-10
rbd-mirror: potential race condition during failure shutdown

dillaman:
Fixes: http://tracker.ceph.com/issues/16980


trociny:
lgtm

PR #10669, opened 2016-08-10
rgw: fix for s3tests failure when ldap auth is not applied

cbodley:
```
FAIL: s3tests.functional.test_s3.test_list_buckets_invalid_auth
  File "/home/cbodley/s3-tests/virtualenv/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/cbodley/s3-tests/s3tests/functional/test_s3.py", line 4611, in test_list_buckets_invalid_auth
    eq(e.error_code, 'InvalidAccessKeyId')
AssertionError: u'AccessDenied' != 'InvalidAccessKeyId'
```

### DIFF
--- a/src/rgw/rgw_rest_s3.cc
+++ b/src/rgw/rgw_rest_s3.cc
@@ -3895,7 +3895,7 @@ int RGW_Auth_S3::authorize_v2(RGWRados *store, struct req_state *s)
   }
 
   /* try keystone auth first */
-  int external_auth_result = -ERR_INVALID_ACCESS_KEY;;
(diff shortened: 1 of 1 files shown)


yehudasa:
@cbodley lgtm, please modify commit message to reference regression. If we expect offending commit to be backported to jewel, also open a tracker issue and reference it too.

cbodley:
@yehudasa ok. do you know if we're planning to backport the s3 auth rework? i would assume not

PR #10673, opened 2016-08-11
common/Formatter: Smarter JSON escape logic

mdw-at-linuxbox:
[This is for radosgw STS.  In theory, escaping /'s should be legal (but only rarely required).  Amazon (for STS) never does this (and doesn't need to), so there are probably clients that don't understand \/.  This change is designed to be something that can be enabled per-use, so won't break the rest of ceph. See github.com/github.com:linuxbox2/linuxbox-ceph wip-rgw-sts-7 src/sts/rgw_rest_sts.cc for intended use. ]

Json doesn't require anything be escaped.  But embedding json in html
requires special treatment for forward slashes.  This change makes doing
that optional, and adds a new "smarter" escape mode that only expands
forward slashes contained inside < > .

### DIFF
--- a/src/common/Formatter.h
+++ b/src/common/Formatter.h
@@ -91,7 +91,12 @@ namespace ceph {
 
   class JSONFormatter : public Formatter {
   public:
-    explicit JSONFormatter(bool p = false);
+    enum Escape {
+      Always = 0,
+      Least = 1,
+      SmartHtml = 2,
+    };
+    explicit JSONFormatter(bool p = false, Escape c = Always);
 
     virtual void set_status(int status, const char* status_name) {};
     virtual void output_header() {};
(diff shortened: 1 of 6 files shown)


mdw-at-linuxbox:
I've pushed a new version of this based on a recent master.  I expect it to fail "make test" above, because that version of master seems to be bad.

dachary:

dachary:

liewegas:
Is this still needed?

oritwas:
@mdw-at-linuxbox , ping

src/common/Formatter.h
```diff
@@ -91,7 +91,12 @@ namespace ceph {
 
   class JSONFormatter : public Formatter {
   public:
-    explicit JSONFormatter(bool p = false);
+    enum Escape {
```
amitkumar50: @mdw-at-linuxbox https://google.github.io/styleguide/cppguide.html#Enumerator_Names

liewegas:
This is a stale pull request and the head branch is in ceph.git.  If this is still a useful change, please reopen a new pull request from your own cloned repo (not the main ceph.git).  You can find the original branch archived in ceph-old.git.  Thanks!

PR #10674, opened 2016-08-11
rgw/rest: don't print empty x-amz-request-id

mdw-at-linuxbox:
[This is required for radosgw STS.  In STS, x-amz-request-id is not used, and not set.  Without this fix, an empty string is emitted, which can't be right.  See github.com/github.com:linuxbox2/linuxbox-ceph wip-rgw-sts-7 for use (no explicit interface/hook.) ]

I don't believe an empty string is ever a valid "x-amz-request-id" record.


oritwas:
lgtm

PR #10676, opened 2016-08-11
rgw/doc: fix indent

Yan-waller:
paragraphs(or new line) with a blank line:

doc

### DIFF
--- a/doc/radosgw/multisite.rst
+++ b/doc/radosgw/multisite.rst
@@ -33,6 +33,7 @@ non-master zones. Following are the basic terminologies that would be used:
   and storage policies. Any configuration change for a non-master zone will
   increment the period's epoch. Changing the master zone to a different zone
   will trigger the following changes:
+
   - A new period is generated with a new period id and epoch of 1
   - Realm's current period is updated to point to the newly generated period id
   - Realm's epoch is incremented


xiexingguo:
:+1:

doc/radosgw/multisite.rst
```diff
@@ -48,7 +51,7 @@ mirroring data changes between the Ceph Object Gateway instances and
 this allows for a much simpler configuration scheme and active-active
 configurations. Please note that metadata operations such as creating
 a user would still need to go through the master zone, however data
-operations such as creation of buckets objects can be handled by any
+operations such as creation of buckets or objects can be handled by any
```
cbodley: this isn't accurate; bucket creation is a metadata operation

theanalyst: hmm actually we need to mention that bucket creation, if done on a non-master zone,  internally forwards the request to the master region (this sort of didn't apply to the user creation as we aren't exactly going via the rest api and thus directly create users on a non-master zone)

doc/radosgw/multisite.rst
```diff
@@ -31,10 +31,13 @@ non-master zones. Following are the basic terminologies that would be used:
   is incremented on every commit operation. Every realm has an associated
   current period, holding the current state of configuration of the zonegroups
   and storage policies. Any configuration change for a non-master zone will
-  increment the period's epoch. Changing the master zone to a different zone
+  increase the period's epoch. Changing the master zone to a different zone
```
theanalyst: we use increment everywhere else, though I'm not really pedantic about that, and okay if this makes it or not :)

Yan-waller: hmm, maybe you are right. but from my point a verb `increase ` sounds reasonable here :)

cbodley: i think i prefer increment, because it's more specific in that it increases by 1

Yan-waller: oh, I got it. @theanalyst @cbodley thanks you two. dropped it

Yan-waller:
I retain the indent modification since that looks more clearer.

PR #10679, opened 2016-08-11
jewel: rbd-nbd does not properly handle resize notifications

dillaman:
http://tracker.ceph.com/issues/16735


dachary:
@dillaman it passed a rbd suite http://tracker.ceph.com/issues/16701#note-5 with two kind of (I think ;-) known errors. If you agree with my assesment, this is ready to merge.

dillaman:
:+1:

PR #10686, opened 2016-08-11
os/bluestore/BlueFS: do not op_file_update deleted files

liewegas:
Make truncate and preallocate no-op if the file has been
deleted.  Otherwise we may get an op_file_update *after*
the op_file_remove in the log, confusing replay.

### DIFF
--- a/src/os/bluestore/BlueFS.cc
+++ b/src/os/bluestore/BlueFS.cc
@@ -1298,7 +1298,10 @@ int BlueFS::_truncate(FileWriter *h, uint64_t offset)
 {
   dout(10) << __func__ << " 0x" << std::hex << offset << std::dec
            << " file " << h->file->fnode << dendl;
-
+  if (h->file->deleted) {
+    dout(10) << __func__ << "  deleted, no-op" << dendl;
+    return 0;
+  }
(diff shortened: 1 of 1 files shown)


markhpc:
Checks are good, but it feels like we might be wallpapering over the issue. :)

liewegas:
It's perfectly legal to operate on an unlinked file in unix/linux, so it's not surprising to me that rocksdb is doing this.

varadakari:
Just a dumb question, If we unlink a file, don't file descriptor should be invalidated? Technically file no longer exist. Should we report the user, that file is already deleted, by returning an error?

tchaikov:
> If we unlink a file, don't file descriptor should be invalidated?

@varadakari no, it is still valid.

> Technically file no longer exist. Should we report the user, that file is already deleted, by returning an error?

no, it's fine. like sage put, it's a common programming practice.

liewegas:
The 'user' is rocksdb.  It is probably doing something like

unlink(name);
delete handle;
	flush(this);

or some other cleanup on file close that happens even when the file is 
unlinked.  It's probably harmless (as it is with a normal posix 
unix-like file system).. I don't think we need to worry about.

PR #10687, opened 2016-08-11
rgw: fix is_admin handling in RGWLDAPAuthEngine and introduce acct_privilege_t

rzarzynski:
@pritha-srivastava, @mattbenjamin: I'm sending the PR early. It still needs some testing.

### DIFF
--- a/src/rgw/rgw_auth.h
+++ b/src/rgw/rgw_auth.h
@@ -120,15 +120,20 @@ public:
     const uint32_t acct_type;
 
   public:
+    enum class acct_privilege_t {
+      IS_ADMIN_ACCT,
+      IS_PLAIN_ACCT
+    };
+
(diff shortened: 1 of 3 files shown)


src/rgw/rgw_auth.h
```diff
@@ -120,15 +120,20 @@ class RGWRemoteAuthApplier : public RGWAuthApplier {
     const uint32_t acct_type;
 
   public:
+    enum class acct_privilage_t {
```
cbodley: privilage -> privilege

pritha-srivastava:
@rzarzynski : Looks fine to me.

src/rgw/rgw_auth.h
rzarzynski: @cbodley: thanks for pointing this out, Casey! Updated.

PR #10688, opened 2016-08-11
Wip 16998

athanatos:


src/osd/PG.cc
```diff
@@ -2546,11 +2546,11 @@ void PG::_update_calc_stats()
 
     // a misplaced object is not stored on the correct OSD
     uint64_t misplaced = 0;
-    unsigned i = 0;
     unsigned in_place = 0;
-    for (vector<int>::iterator p = up.begin(); p != up.end(); ++p, ++i) {
-      pg_shard_t s(*p,
-		   pool.info.ec_pool() ? shard_id_t(i) : shard_id_t::NO_SHARD);
+    for (set<pg_shard_t>::const_iterator p = upset.begin();
```
tchaikov: what is the difference between this approach and the existing way? i think they are the same.

tchaikov: can we skip the shards who are `CRUSH_ITEM_NONE`? and could use `auto` and range based loop.

tchaikov:
nit, might want to connect to http://tracker.ceph.com/issues/16998 using "Fixes: http://tracker.ceph.com/issues/16998"?

src/osd/PG.cc
athanatos: It's really meant to match the actingset and actingbackfill sets.  Mostly, using the up and acting vectors directly has led to a ton of subtle bugs like this one.

athanatos:
Added the Fixes: line (thanks!).

src/osd/PG.cc
tchaikov: also, could use:
```
for (const auto &s : upset) {
```
here, and replace `*p` with `s` at https://github.com/ceph/ceph/pull/10688/files#diff-dfb9ddca0a3ee32b266623e8fa489626R2562 and  https://github.com/ceph/ceph/pull/10688/files#diff-dfb9ddca0a3ee32b266623e8fa489626R2567.

tchaikov:
modulo a nit. lgtm.

tchaikov:
retest this please.

src/osd/PG.cc
athanatos: Eh, I wanted it to be symmetric with the actingset scan above.  Also, we might want to backport these to hammer.

tchaikov:
lgtm.

src/osd/PG.cc
tchaikov: okay.

athanatos:
retest this please.

yuriw:

PR #10691, opened 2016-08-11
Clean up handling of "/.." in ceph client

jtlayton:
This PR is a follow up to a RFC patch posting I made on the ceph-devel list. The libcephfs code has some special handling for ".." at the root that seems to be entirely unnecessary. The typical way this is handled is to make the root its own parent. This patchset makes that change and adds some testcases that verify that it will continue to work properly in the future.

Note that ceph-fuse isn't affected by this, since the Linux VFS layer will detect that there is a mountpoint traversal and handle it appropriately. So, this should only have an effect on applications that are working with libcephfs directly.

The main impetus for this is to simplify the code to prepare for some coming changes (the statx series, in particular).

### DIFF
--- a/src/client/Client.cc
+++ b/src/client/Client.cc
@@ -5910,7 +5910,7 @@ int Client::_lookup(Inode *dir, const string& dname, int mask,
 
   if (dname == "..") {
     if (dir->dn_set.empty())
-      r = -ENOENT;
+      *target = dir;
     else
       *target = dir->get_first_parent()->dir->parent_inode; //dirs can't be hard-linked
     goto done;
@@ -7239,16 +7239,14 @@ int Client::readdir_r_cb(dir_result_t *d, add_dirent_cb_t cb, void *p)
   if (dirp->offset == 1) {
     ldout(cct, 15) << " including .." << dendl;
     uint64_t next_off = 2;
-    if (!diri->dn_set.empty()) {
-      InodeRef& in = diri->get_first_parent()->inode;
-      fill_stat(in, &st);
-      fill_dirent(&de, "..", S_IFDIR, st.st_ino, next_off);
-    } else {
-      /* must be at the root (no parent),
-       * so we add the dotdot with a special inode (3) */
-      fill_dirent(&de, "..", S_IFDIR, CEPH_INO_DOTDOT, next_off);
-    }
+    InodeRef in;
+    if (diri->dn_set.empty())
+      in = diri;
+    else
+      in = diri->get_first_parent()->inode;
 
+    fill_stat(in, &st);
+    fill_dirent(&de, "..", S_IFDIR, st.st_ino, next_off);
 
     client_lock.Unlock();
     int r = cb(p, &de, &st, -1, next_off);
@@ -9683,13 +9681,6 @@ int Client::ll_getattr(Inode *in, struct stat *attr, int uid, int gid)
   ldout(cct, 3) << "ll_getattr " << vino << dendl;
   tout(cct) << "ll_getattr" << std::endl;
   tout(cct) << vino.ino.val << std::endl;
-
-  /* special case for dotdot (..) */
-  if (vino.ino.val == CEPH_INO_DOTDOT) {
-    attr->st_mode = S_IFDIR | 0755;
-    attr->st_nlink = 2;
-    return 0;
-  }
 
   int res;
   if (vino.snapid < CEPH_NOSNAP)
--- a/src/test/libcephfs/test.cc
+++ b/src/test/libcephfs/test.cc
@@ -1408,3 +1408,62 @@ TEST(LibCephFS, Nlink) {
 
   ceph_shutdown(cmount);
 }
+
+TEST(LibCephFS, SlashDotDot) {
+  struct ceph_mount_info *cmount;
+  ASSERT_EQ(ceph_create(&cmount, NULL), 0);
+  ASSERT_EQ(ceph_conf_read_file(cmount, NULL), 0);
+  ASSERT_EQ(0, ceph_conf_parse_env(cmount, NULL));
+  ASSERT_EQ(ceph_mount(cmount, "/"), 0);
+
+  struct stat	st;
+  ASSERT_EQ(ceph_stat(cmount, "/.", &st), 0);
+
+  ino_t ino = st.st_ino;
+  ASSERT_EQ(ceph_stat(cmount, "/..", &st), 0);
+
+  /* At root, "." and ".." should be the same inode */
+  ASSERT_EQ(ino, st.st_ino);
+
+  /* Test accessing the parent of an unlinked directory */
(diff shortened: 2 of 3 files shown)


jtlayton:
I have test run queued up on mira that contains these patches + the whole statx series:

    http://pulpito.ceph.com/jlayton-2016-08-11_17:42:40-fs-wip-jlayton-statx---basic-mira/

src/client/Client.cc
```diff
@@ -7239,17 +7241,16 @@ int Client::readdir_r_cb(dir_result_t *d, add_dirent_cb_t cb, void *p)
   if (dirp->offset == 1) {
     ldout(cct, 15) << " including .." << dendl;
     uint64_t next_off = 2;
-    if (!diri->dn_set.empty()) {
+    if (diri->client->is_root(diri)) {
+      /* must be at the root (no parent), so fill out with same info as "." */
+      fill_stat(diri, &st);
+      fill_dirent(&de, "..", S_IFDIR, st.st_ino, next_off);
+    } else {
       InodeRef& in = diri->get_first_parent()->inode;
```
batrick: I think you're missing a check for an unlinked directory here. (i.e. dn_set.empty() is true)

batrick: Ugh, nevermind. Forgot this wasn't necessary.

src/test/libcephfs/test.cc
```diff
@@ -1408,3 +1408,49 @@ TEST(LibCephFS, Nlink) {
...
+  ASSERT_EQ(ceph_opendir(cmount, "/", &rdir), 0);
+  struct dirent *result = ceph_readdir(cmount, rdir);
+  ASSERT_TRUE(result != NULL);
+  ASSERT_STREQ(result->d_name, ".");
+  ino = result->d_ino;
+  result = ceph_readdir(cmount, rdir);
+  ASSERT_TRUE(result != NULL);
+  ASSERT_STREQ(result->d_name, "..");
+  ASSERT_EQ(ino, result->d_ino);
+
```
batrick: Only thing missing is unlinked directory traversal/parent lookup.

batrick:
Looks good otherwise @jtlayton!

Reviewed-by: Patrick Donnelly <pdonnell@redhat.com>

src/test/libcephfs/test.cc
jtlayton: Ok, added a test for that as well. Look ok? It passes on my own test rig, but the teuthology run I kicked off a little while ago obviously doesn't have it.

src/test/libcephfs/test.cc
```diff
@@ -1425,16 +1425,24 @@ TEST(LibCephFS, SlashDotDot) {
...
   sprintf(dir2, "%s/sub%x", dir1, getpid());
 
   ASSERT_EQ(ceph_mkdir(cmount, dir1, 0755), 0);
   ASSERT_EQ(ceph_mkdir(cmount, dir2, 0755), 0);
+
+  ASSERT_EQ(ceph_chdir(cmount, dir2), 0);
+  ASSERT_EQ(ceph_rmdir(cmount, dir2), 0);
+  ASSERT_EQ(ceph_stat(cmount, dir1, &st), 0);
+  ino = st.st_ino;
+  ASSERT_EQ(ceph_stat(cmount, "..", &st), -ENOENT);
```
batrick: Why should this be -ENOENT?

jtlayton: Because dir2 is now unlinked. We chdir into "dir2", then rmdir "dir2" and then see whether the parent of that dir is now accessible. It's not, so you get -ENOENT back (no parent).

Hmm...come to think of it -- maybe we do need to handle the readdir case here as well? No reason you couldn't issue a readdir from in that now-unlinked directory.

jtlayton:
Just did another push to try to bring this closer into line with Linux VFS behavior. It's still not quite right yet though, as doing a lookup of ".." on an unlinked directory should give you the previous parent.

jtlayton:
Ok, I think this is the best we can do for the time being. The libcephfs behavior on unlinked directories differs from the Linux VFS in two ways:

1) if you're inside a directory that is then unlinked, then getdents on linux will always return -ENOENT. With libcephfs, you'll get a valid entry at position 1 (aka "."), but -ENOENT on position 2 (".."). POSIX says an ENOENT here means that the current position of the directory stream is invalid, so I think this is still OK by the spec.

2) a process chdirs into a directory that is then unlinked. Doing a lookup of ".." at that point will give you the original parent in the Linux VFS, but with libcephfs you get -ENOENT. This is due to the way that the dn_set is handled at unlink time, AFAICT. We could probably fix this by keeping a reference to the original parent inode somehow when the directory is unlinked, but I'm not sure it's worth it.

Neither of these come into play with ceph-fuse either. The Linux VFS behavior supersedes any of these behavioral differences.

jtlayton:
New test run going now:

    http://pulpito.ceph.com/jlayton-2016-08-12_12:10:57-fs-wip-jlayton-dotdot---basic-mira/

jtlayton:
Ok, and that test run failed....

There seem to be some behavioral differences that break find and rm with the earlier set. I've switched it back to the original scheme I had which was to basically treat the root and an unlinked directory the same way. In effect, an unlinked directory becomes a little root of its own. With that, the behavior seems to be as expected. I'll rerun the "fs" suite and see how it does, but I think this is the right way to handle it.

jtlayton:
Latest test run seems to be better:

http://pulpito.ceph.com/jlayton-2016-08-13_01:11:59-fs-wip-jlayton-dotdot---basic-mira/

There are some failures, but they all seem to be known problems that are being worked and are unrelated to this patchset:

known valgrind failures
known problem with the java jni linkage on centos
known ThreesomeInterprocessLocking test failures
kernel lockdep warnings about circular locking, unrelated to ceph -- probably an xfs bug or a false positive

batrick:
Jeff, just to be clear:

For an unlinked directory: returning legitimate "." and ".." in libcephfs readdir (or lookup on "..") will result in errors for rm/find? I think you said earlier that Linux VFS behavior will supersede FUSE for those two cases so changing libcephfs shouldn't even have an effect on userspace right?

ukernel:
http://pulpito.ceph.com/zyan-2016-08-16_08:24:11-fs-wip-jlayton-dotdot-testing-basic-smithi/

jtlayton:
Linux VFS behavior will supersede the behavior when accessing the parent of the root of the filesystem. Once you're inside the filesystem though, then it can have an effect, and I think that mostly comes about when handling readdir.

That said, I wasn't able to reproduce the failure I saw on my own test rig, so I wasn't able to track down the actual failure scenario.

ukernel:
The failure might be caused by old kernel.next time you schedule a fs suite, please use add '-k testing' option. which makes test machine use recent kernel

PR #10694, opened 2016-08-11
WITH_DLIBROCKSDB fixes.

mdw-at-linuxbox:
[ I've been using this with STS for a while now, but there's nothing here specific to STS.  I can avoid building rocksdb (with all the icky in-tree build badness) by saying "-DWITH_ROCKSDB=YES".  I do have to install rocksdb separately first: see http://people.redhat.com/~mwatts/rocksdb/ for rpms. ]

Add new option "WITH_ROCKSDB" to cmake.

And for automake,
When building with WITH_DLIBROCKSDB: reference rocksdb library
as -lrocksdb not rocksdb/librocksdb.a

When linking with WITH_DLIBROCKSDB: add -lz4 -lbz2.

don't try to "ar" -lrocksdb; even if it worked it wouldn't be right.


src/kv/Makefile.am
```diff
@@ -40,7 +40,7 @@ endif
 
 if WITH_DLIBROCKSDB
 libkv_a_SOURCES += kv/RocksDBStore.cc
-libkv_a_LIBADD += -lrocksdb
+# libkv_a_LIBADD += -lrocksdb
```
liewegas: remove this line?

liewegas:
We can probably ignore automake.

For cmake, I don't see any -lrocksdb if we're not building statically... how does it know to link to the so?

tchaikov:
>  how does it know to link to the so?

@liewegas see https://github.com/ceph/ceph/blob/master/src/CMakeLists.txt#L667 and  https://github.com/ceph/ceph/blob/master/src/kv/CMakeLists.txt#L10

without the line of 
```cmake
add_library(rocksdb STATIC IMPORTED)
```
cmake will link against librocksdb in the default library path. that's how we link against "libm", and "libsnappy" and other libraries. ideally, we should have used the absolute paths found by `find_library()`, though.

CMakeLists.txt
```diff
@@ -340,6 +340,7 @@ if(WITH_REENTRANT_STRSIGNAL)
     set(HAVE_REENTRANT_STRSIGNAL 1 CACHE INTERNAL "Reentrant strsignal is supported.")
 endif(WITH_REENTRANT_STRSIGNAL)
 
+option(WITH_ROCKSDB "Use system rocksdb" OFF)
```
tchaikov: might want to `s/WITH_ROCKSDB /WITH_DLIBROCKSDB/` ?

and shall fail the configure phase if librocksdb is not found?

tchaikov:
@mdw-at-linuxbox ping

liewegas:
@mdw-at-linuxbox ping

PR #10695, opened 2016-08-11
mon/MDSMonitor.cc: use std::move on strings

stiopaa1:

### DIFF
--- a/src/mon/MDSMonitor.cc
+++ b/src/mon/MDSMonitor.cc
@@ -3017,7 +3017,7 @@ bool MDSMonitor::try_standby_replay(
 }
 
 MDSMonitor::MDSMonitor(Monitor *mn, Paxos *p, string service_name)
-  : PaxosService(mn, p, service_name)
+  : PaxosService(mn, p, std::move(service_name))
 {
   handlers.push_back(std::make_shared<SetHandler>());
   handlers.push_back(std::make_shared<LegacyHandler<SetHandler> >("mds set"));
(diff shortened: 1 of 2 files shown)


gregsfortytwo:
This looks fine, but should we be pushing it farther up the stack as well? (Or just using const refs? I don't remember how they're used.)

stiopaa1:
@gregsfortytwo 
You are talking about cmd_get val, right? 
Using std::move here avoids one copy, but as we don't need store value of string anywhere changing cmd_get_val to take const string& would be even better option(avoids moving).  I have removed it from this commit and will change it to const string& in a different one.

stiopaa1:
I have also added std::move in PaxosService.h

gregsfortytwo:
Okay, I don't like this new version. Constructors on PaxosService and MDSMonitor are rarely invoked and any confusion about ownership isn't worth it if somebody ends up passing re-used strings or something. I think just using const string& will capture the important bits.

stiopaa1:
@gregsfortytwo 
This is a pretty common way of doing this, which is also used in a different places of the code base, so im not sure why there should be any confusion? could you please elaborate?

gregsfortytwo:
The MDSMonitor constructor is only called once per mon process and the parent PaxosService is called maybe 5 times per process, all at startup. There's very little point in optimizing them like this if there is any chance of that causing bugs in the future. It's not likely to cause trouble, but stealing string values can do so in some cases if they're reused (which we do with other initialization strings from time to time). So let's skip this one.

PR #10696, opened 2016-08-11
Wip buildfix python out of tree

mdw-at-linuxbox:
[ More build fixes from STS - but! changes upstream makes these mostly obsolete.  PR for the sake of making this visible to the people who care, but I expect they will have some other fix they'd rather do instead. ]

Originally, these were 2 things: "out of tree" fixes for autoconf/automake, and python fixes for out of tree and other behaviors.  The automake stuff probably longer matters.

So the python fixes: people have been changing the behavior of src/pybind/*/setup.py to get rid of referencing ceph_ver.h - but the the fixes aren't very systematic and src/pybind/rbd/ still uses this (it's the last such reference).  My fix does include "pep440" changes, which you want if you want to use the ceph version from ceph_ver.h.  Also note rados get_ceph_version() which isn't used (and goes away here.)  It would be good to see a proper systematic fix (probably one that doesn't use ceph_ver.h).

### DIFF
--- a/cmake/modules/Distutils.cmake
+++ b/cmake/modules/Distutils.cmake
@@ -45,6 +45,7 @@ function(distutils_add_cython_module name src)
     LDFLAGS=-L${CMAKE_LIBRARY_OUTPUT_DIRECTORY}
     CYTHON_BUILD_DIR=${CMAKE_CURRENT_BINARY_DIR}
     CEPH_LIBDIR=${CMAKE_LIBRARY_OUTPUT_DIRECTORY}
+    CEPH_BUILD_TOP=${CMAKE_BINARY_DIR}
     CFLAGS=\"-iquote${CMAKE_SOURCE_DIR}/src/include\"
     ${PYTHON${PYTHON_VERSION}_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/setup.py
     build --verbose --build-base ${CYTHON_MODULE_DIR}
@@ -67,6 +68,7 @@ function(distutils_install_cython_module name)
     execute_process(
        COMMAND env
            CYTHON_BUILD_DIR=${CMAKE_CURRENT_BINARY_DIR}
+	   CEPH_BUILD_TOP=${CMAKE_BINARY_DIR}
            CEPH_LIBDIR=${CMAKE_LIBRARY_OUTPUT_DIRECTORY}
            CC=${CMAKE_C_COMPILER}
            CPPFLAGS=\"-iquote${CMAKE_SOURCE_DIR}/src/include\"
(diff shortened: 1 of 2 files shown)


tchaikov:
@mdw-at-linuxbox i will check this PR tmr.

src/pybind/rbd/setup.py
```diff
@@ -20,9 +20,17 @@ def get_version():
     try:
         for line in open(os.path.join(dir, "ceph_ver.h")):
             if "CEPH_GIT_NICE_VER" in line:
-                return line.split()[2][1:-1]
+                l=line.split()[2].strip('"')
+                if len(l) > 1 and l[0] == 'v': l = l[1:]
+                a=l.split("-",2)
+                if len(a) > 2:
+                    return "{}.post{}+{}".format( a[0], a[1], a[2] )
```
tchaikov: i think we can just use git version, i.e. x.y.z for the extension's version. and iiuc, we are not using the post version,

quote from https://www.python.org/dev/peps/pep-0440/#post-releases
> Some projects use post-releases to address minor errors in a final release that do not affect the distributed software (for example, correcting an error in the release notes).

 so X.postY.Z does not apply to our project, right?

cmake/modules/Distutils.cmake
```diff
@@ -45,6 +45,7 @@ function(distutils_add_cython_module name src)
     LDFLAGS=-L${CMAKE_LIBRARY_OUTPUT_DIRECTORY}
     CYTHON_BUILD_DIR=${CMAKE_CURRENT_BINARY_DIR}
     CEPH_LIBDIR=${CMAKE_LIBRARY_OUTPUT_DIRECTORY}
+    CEPH_BUILD_TOP=${CMAKE_BINARY_DIR}
```
tchaikov: nit, could prefix the title of the commit message with `pybind:` instead of `setup.py`.

src/common/Makefile.am
```diff
@@ -80,7 +80,7 @@ libcommon_internal_la_SOURCES = \
 	common/blkdev.cc \
 	common/dns_resolve.cc
 
-common/PluginRegistry.cc: ./ceph_ver.h
+common/PluginRegistry.cc: ceph_ver.h
```
tchaikov: could prefix the title of the commit message with "Makefile: ".

dachary:

liewegas:
what's teh status of this one?

tchaikov:
it's out of my radar, will review it later on.

src/pybind/rados/setup.py
```diff
@@ -26,15 +26,30 @@
 __version__ = '2.0.0'
 
 
-def get_ceph_version():
+def get_version():
```
tchaikov: `get_version()` was not called anywhere. the same applies to `get_ceph_version()`. the version number of the module is hardwired to `__version__` here.

tchaikov:
#17727 supersedes this one

tchaikov:
closing this one, since #17727 has been merged.

PR #10697, opened 2016-08-11
add qa test of envlibrados for rocksdb

ryneli:
Signed-off-by: Ryne Li <lizhenqiangsnake@gmail.com>

### DIFF
--- a/qa/workunits/rados/test_envlibrados_for_rocksdb.sh
+++ b/qa/workunits/rados/test_envlibrados_for_rocksdb.sh
@@ -1,0 +1,95 @@
+#!/bin/bash -ex
+############################################
+#			Helper functions
+############################################
+function install() {
+    for package in "$@" ; do
+        install_one $package
+    done
+    return 0
+}
+
+function install_one() {
+    case $(lsb_release -si) in
+        Ubuntu|Debian|Devuan)
+            sudo apt-get install -y "$@"
+            ;;
+        CentOS|Fedora|RedHatEnterpriseServer)
+            sudo yum install -y "$@"
+            ;;
+        *SUSE*)
+            sudo zypper --non-interactive install "$@"
+            ;;
+        *)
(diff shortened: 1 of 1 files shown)


liewegas:
@jdurgin what do you think?

qa/workunits/rados/test_envlibrados_for_rocksdb.sh
```diff
@@ -0,0 +1,95 @@
+#!/bin/bash
```
jdurgin: looks good, the exit status should indicate if it fails - I like using '#!/bin/bash -ex' for this so we have logging of what precisely failed if something goes wrong

jdurgin:
A couple changes needed: 
1) chmod +x the script 
2) figure out why it's failing 

I pushed it to the wip-rocksdb-librados-test branch and scheduled a run with:

```~/teuthology/virtualenv/bin/teuthology-suite -v -e <my email  addr> -m mira --owner joshd@teuthology -c master -s rados --suite-dir ~/ceph-qa-suite --filter envlibrados --newest 10 --limit 2 workunit_branch.yaml```

where ~/ceph-qa-suite/workunit_branch.yaml contained:
```overrides:
  workunit:
    sha1: 89855ecb68ad4e172978a304b2d8a9566d1c3cf3
```

The results are here:
http://pulpito.ceph.com/joshd-2016-08-22_21:51:53-rados-master---basic-mira/

It failed on ubuntu, I'm guessing from mismatched library/header librados versions, but I'm not sure.
I see some apt-get warnings in the log for the failed run too - these can be fixed by adding ```DEBIAN_FRONTEND=noninteractive``` and ```--force-yes``` to the apt-get commands.

jdurgin:
The failure seems to be a race condition in libnss, or a very subtle bug in common/ceph_crypto.cc. It's not reproducing very reliably anymore, and disappears with timing changes like running manually instead of via scheduled jobs, so it seems safe to merge this. Merged manually to also save the rocksdb test binary for core file analysis.

PR #10698, opened 2016-08-11
bluestore: do not use freelist to track bluefs_extents

liewegas:
We track these explicitly in the bluestore superblock in bluefs_extents--we
dont' need to flip bits.  This speeds up mkfs and balancing significantly.


liewegas:
retest this please

PR #10699, opened 2016-08-11
jewel: doc: fix by-parttypeuuid in ceph-disk(8) nroff

ktdreyer:
Commit 221efb0b893adbfd7a19df171cf967fee87afcc7 altered the rST source for the ceph-disk man page. In Hammer, we also have to modify the nroff sources, because static copies of the generated man pages are stored in Git.

Fixes: http://tracker.ceph.com/issues/15867


tchaikov:
lgtm

dachary:
nice catch

PR #10700, opened 2016-08-11
RFC: buffer: add safe_appender and unsafe_appender

liewegas:
A lot of time in current encoding methods is spent in buffer::append().  This is because it does a lot of bounds checking, and because it has to update both the buffer::list size, buffer::ptr (buffer::list::append_buffer) size.

The idea here is to build a new paradigm around appenders that skip all this housekeeping until the appender is destroyed and it does a flush() to the appropriate list.  That way each individual append is just a memcpy and pointer update.

The safe_appender does bounds checking and flushes as needed.

The idea with unsafe_appender is that when you know an upper bound on how much you will append, you can skip the bounds check on every call.

Initial results:

[ RUN      ] BufferList.appender_bench
appending 1073741824 bytes
buffer::list::append 20.285963
buffer::list encode 19.719120
buffer::list::safe_appender::append 2.588926
buffer::list::safe_appender::append_v 2.837026
buffer::list::safe_appender encode 3.000614
buffer::list::unsafe_appender::append 2.452116
buffer::list::unsafe_appender::append_v 2.553745
buffer::list::unsafe_appender encode 2.200110
[       OK ] BufferList.appender_bench (55637 ms)

The unsafe vs safe results are surprising... unsafe isn't actually much master.  I suspect that this bounds check is actually quite cheap since the CPU is doing branch prediction.  So maybe we don't even need it?

### DIFF
--- a/src/include/buffer.h
+++ b/src/include/buffer.h
@@ -48,6 +48,7 @@
 #include "page.h"
 #include "crc32c.h"
 #include "buffer_fwd.h"
+#include "include/inline_memory.h"
 
 #ifdef __CEPH__
 # include "include/assert.h"
@@ -378,6 +379,110 @@ namespace buffer CEPH_BUFFER_API {
       }
     };
 
+    class appender {
+    protected:
+      list *pls;
+      size_t size;
+      ptr bp;
+      char *pos;
+      union {
+	char *end;
+	appender *parent;
+      };
+
+    public:
+      appender(bufferlist *l, size_t s) : pls(l), size(s) {
+	prepare_buffer();
+      }
+      appender(appender *p)
+	: pls(nullptr),
+	  pos(p->pos),
+	  parent(p) {}
+      appender(appender&& o)
+	: pls(o.pls), size(o.size),
+	  bp(std::move(o.bp)),
+	  pos(o.pos), end(o.end) {
+      }
+      ~appender() {
+	flush();
+      }
+
+      // no copying
+      appender(const appender& other) = delete;
+      appender& operator=(const appender& other) = delete;
+
+    protected:
+      void prepare_buffer(size_t at_least = 0) {
+	size_t l = std::max(size, at_least);
+	bp = buffer::create(l);
+	pos = bp.c_str();
+	end = pos + l;
+      }
+
+      void flush() {
+	if (pls) {
+	  if (bp.length()) {
+	    bp.set_length(pos - bp.c_str());
+	    pls->append(bp);
+	    bp = ptr();
+	  }
+	} else {
+	  parent->pos = pos;
+	}
+      }
+    };
+
+    class unsafe_appender : public appender {
+    public:
+      unsafe_appender(bufferlist *l, size_t s) : appender(l, s) {}
+      unsafe_appender(appender *parent) : appender(parent) {}
+      void append(char *p, size_t l) {
+	maybe_inline_memcpy(pos, p, l, 16);
+	pos += l;
+      }
+      template<typename T>
+      void append_v(T v) {
+	*(T*)pos = v;
+	pos += sizeof(T);
+      }
+    };
+
+    class safe_appender : public appender {
+    public:
+      safe_appender(bufferlist *l, size_t s) : appender(l, s) {}
+      void append(char *p, size_t l) {
+	if (pos + l > end) {
+	  flush();
+	  prepare_buffer(l);
+	}
+	maybe_inline_memcpy(pos, p, l, 16);
+	pos += l;
+      }
+      template<typename T>
+      void append_v(T v) {
+	if (pos + sizeof(T) > end) {
+	  flush();
+	  prepare_buffer();
+	}
+	*(T*)pos = v;
+	pos += sizeof(T);
+      }
+      unsafe_appender reserve(size_t s) {
+	if (pos + s > end) {
+	  flush();
+	  prepare_buffer(s);
+	}
+	return unsafe_appender(this);
+      }
+    };
+
+    unsafe_appender get_unsafe_appender(size_t len) {
+      return unsafe_appender(this, len);
+    }
+    safe_appender get_safe_appender(size_t len=4096) {
+      return safe_appender(this, len);
+    }
+
   private:
     mutable iterator last_p;
     int zero_copy_to_fd(int fd) const;
@@ -601,7 +706,7 @@ namespace buffer CEPH_BUFFER_API {
     int write_fd_zero_copy(int fd) const;
     void prepare_iov(std::vector<iovec> *piov) const;
     uint32_t crc32c(uint32_t crc) const;
-	void invalidate_crc();
+    void invalidate_crc();
   };
 
   /*
--- a/src/test/bufferlist.cc
+++ b/src/test/bufferlist.cc
@@ -2269,6 +2269,322 @@ TEST(BufferList, append_zero) {
   EXPECT_EQ('\0', bl[1]);
 }
 
+TEST(BufferList, appender_bench) {
+  unsigned long long ma = 16 * 16, mb = 65536;
+  cout << "appending " << 16 * (ma * mb) << " uint64_t's" << std::endl;
+  size_t bufmax = 1048576 * 1024;
+  char *buf = new char[bufmax];
+  uint64_t *p;
+  memset(buf, 0, bufmax);
+  {
+    utime_t start = ceph_clock_now(NULL);
+    for (unsigned a = 0; a < ma; ++a) {
+      bufferlist bl;
+      for (unsigned b = 0; b < mb; ++b) {
+	p = (uint64_t*)buf +
+	  (1234567 * a + 45678907 * b) % (bufmax/2/sizeof(uint64_t)) +
+	  123 * a;
+	bl.append((char*)(p++), sizeof(*p));
+	bl.append((char*)(p++), sizeof(*p));
+	bl.append((char*)(p++), sizeof(*p));
+	bl.append((char*)(p++), sizeof(*p));
+	bl.append((char*)(p++), sizeof(*p));
+	bl.append((char*)(p++), sizeof(*p));
+	bl.append((char*)(p++), sizeof(*p));
+	bl.append((char*)(p++), sizeof(*p));
+	bl.append((char*)(p++), sizeof(*p));
+	bl.append((char*)(p++), sizeof(*p));
+	bl.append((char*)(p++), sizeof(*p));
+	bl.append((char*)(p++), sizeof(*p));
+	bl.append((char*)(p++), sizeof(*p));
+	bl.append((char*)(p++), sizeof(*p));
+	bl.append((char*)(p++), sizeof(*p));
+	bl.append((char*)(p++), sizeof(*p));
+      }
+    }
+    utime_t dur = ceph_clock_now(NULL) - start;
+    cout << "buffer::list::append " << dur << std::endl;
+  }
+  {
+    utime_t start = ceph_clock_now(NULL);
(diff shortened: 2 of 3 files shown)


src/include/buffer.h
```diff
@@ -378,6 +378,88 @@ namespace buffer CEPH_BUFFER_API {
...
+      list *pls;
+      size_t size;
+      ptr bp;
+      char *pos, *end;
+
+    public:
+      appender(bufferlist *l, size_t s) : pls(l), size(s) {
+	prepare_buffer();
+      }
+      appender(appender&& o)
```
ifed01: double &?

liewegas: move constructor.  not sure if this is the right way to implement it, but 
we do want to disallow copying and only allow moves so that we have an 
accurate position to flush() when the destructor is called.

src/test/bufferlist.cc
```diff
@@ -2269,6 +2269,120 @@ TEST(BufferList, append_zero) {
...
+    utime_t dur = ceph_clock_now(NULL) - start;
+    cout << "buffer::list::safe_appender encode " << dur << std::endl;
+  }
+
+  {
+    utime_t start = ceph_clock_now(NULL);
+    uint64_t v = 123;
+    for (unsigned a = 0; a < ma; ++a) {
+      bufferlist bl;
+      bufferlist::unsafe_appender ap = bl.get_unsafe_appender(sizeof(v) * mb);
```
ifed01: You pre-allocate 4K for safe allocator and 64K*8 for unsafe one. Hence your comparison for safe vs.  unsafe isn't pretty adequate. Suggest to make similar values and compare. Probably the diff between them wouldn't be that much.

src/include/buffer.h
```diff
@@ -378,6 +378,88 @@ namespace buffer CEPH_BUFFER_API {
...
+      appender& operator=(const appender& other) = delete;
+
+      void prepare_buffer(size_t at_least = 0) {
+	bp = buffer::create(std::max(size, at_least));
+	pos = bp.c_str();
+	end = pos + bp.length();
+      }
+
+      void flush() {
+	if (bp.length()) {
```
ifed01: any sense for this check? I'd say check for pos-bp.c_str() != 0 is more valid

src/include/buffer.h
```diff
@@ -378,6 +378,88 @@ namespace buffer CEPH_BUFFER_API {
...
+      }
+    };
+
+    class safe_appender : public appender {
+    public:
+      safe_appender(bufferlist *l, size_t s) : appender(l, s) {}
+      void append(char *p, size_t l) {
+	if (pos + l > end) {
+	  flush();
+	  prepare_buffer(l);
```
ifed01: Make a single call flush_and_prepare?

liewegas: good catch

liewegas:
I redid this to put 16 calls, which is a more likely pattern in the calling code than a loop.  This shows a huge differnce (2-3x) between safe and unsafe, which makes more sense to me.  ALso fixed the allocation hint so that they are both preallocating the same amount.

[ RUN      ] BufferList.appender_bench
appending 16777216 bytes
buffer::list::append 4.186595
buffer::list encode 4.020799
buffer::list::safe_appender::append 1.868394
buffer::list::safe_appender::append_v 0.526946
buffer::list::safe_appender encode 1.673153
buffer::list::unsafe_appender::append 0.577497
buffer::list::unsafe_appender::append_v 0.574808
buffer::list::unsafe_appender encode 0.572381
[       OK ] BufferList.appender_bench (14001 ms)

markhpc:
ok, cool!  Can't wait to try this for real!

ifed01:
@liewegas: this looks strange:
buffer::list::safe_appender::append 1.868394
buffer::list::safe_appender::append_v 0.526946
> x3 times difference when replacing memcpy with assignment? And this difference doesn't present for unsafe version...
Do you observe the same results when running multiple times?

src/include/buffer.h
somnathr: @liewegas I think you need a move assignment operator too..

liewegas:
It does.  I fixed encode_raw to use append_v and now I get this:

[ RUN      ] BufferList.appender_bench
appending 268435456 uint64_t's
buffer::list::append 4.166772
buffer::list encode 4.217864
buffer::list::safe_appender::append 2.220418
buffer::list::safe_appender::append_v 0.525496
buffer::list::safe_appender encode 0.563121
buffer::list::unsafe_appender::append 0.563079
buffer::list::unsafe_appender::append_v 0.561652
buffer::list::unsafe_appender encode 0.576483
[       OK ] BufferList.appender_bench (13395 ms)

liewegas:
The results are still confusing.  I'm not sure why safe_appender encode got faster but the unsafe_appender encode stayed the same.

ifed01:
Moreover - latest results are for appending 256M 64bits while the previous were for 16M bytes. But numbers are pretty close!

liewegas:
Added the safe_appender::reserve() that returns a temporary unsafe_appender.  With just 16 members gets pretty much all the benefit of the unsafe_appender (which makes sense... doing one bounds check for 16 elements, plus some overhead on construction and flush/assignment).

[ RUN      ] BufferList.appender_bench
appending 268435456 uint64_t's
buffer::list::append 3.973437
buffer::list encode 4.015779
buffer::list::safe_appender::append 3.190428
buffer::list::safe_appender::append (reserve 16) 0.540660
buffer::list::safe_appender::append_v 0.565126
buffer::list::safe_appender::append_v (reserve 16) 0.562342
buffer::list::safe_appender encode 0.816910
buffer::list::unsafe_appender::append 0.533881
buffer::list::unsafe_appender::append_v 0.549342
buffer::list::unsafe_appender encode 0.511045
[       OK ] BufferList.appender_bench (15259 ms)

liewegas:
buffer::list::safe_appender::append_v 0.565126
is sitll the one that doesn't make sense to me.  it's still doing the bounds check.. the only difference between the previous test is that it does *(T*)pos = v; instead of memcpy.  ANd we don't see the same diffrences for the version with no bounds check.  It must be confusing the optimizer or cpu branch prediction or something.  Probably need to look at a disassembly to see wth is going on...

src/include/buffer.h
```diff
@@ -378,6 +378,88 @@ namespace buffer CEPH_BUFFER_API {
...
+	}
+	memcpy(pos, p, l);
```
branch-predictor: I'd use maybe_inline_memcpy for that (https://github.com/ceph/ceph/blob/e9e77c855300cee2072f6980ceb6f0e0cda00bbf/src/include/inline_memory.h#L24)

src/include/buffer.h
```diff
@@ -378,6 +378,88 @@ namespace buffer CEPH_BUFFER_API {
...
+	*(T*)pos = v;
+	pos += sizeof(T);
+      }
+    };
+
+    class unsafe_appender : public appender {
+    public:
+      unsafe_appender(bufferlist *l, size_t s) : appender(l, s) {}
+      void append(char *p, size_t l) {
+	memcpy(pos, p, l);
```
branch-predictor: And that too.

src/include/buffer.h
```diff
@@ -378,6 +378,88 @@ namespace buffer CEPH_BUFFER_API {
...
+      }
+
+      // no copying
+      appender(const appender& other) = delete;
+      appender& operator=(const appender& other) = delete;
+
+      void prepare_buffer(size_t at_least = 0) {
+	bp = buffer::create(std::max(size, at_least));
+	pos = bp.c_str();
+	end = pos + bp.length();
```
yehudasa: @liewegas why call `bp.length()` here if it's `std::max(size, at_least)`? Just reuse the latter (could be that optimizer will be smart enough, but no need to)

markhpc:
I did a quick test with the memcpy changed to maybe_inline_memcpy in both safe_appender and unsafe_appender:

`maybe_inline_memcpy(pos, p, l, 32);`

I compiled and did a quick test on my laptop in both cases and saw some interesting results.  Typically the maybe_inline_memcpy seemed to be about 3-5% faster, except in two cases.  Safe append appeared to be ~36% faster while safe append_v appeared to be ~51% slower.  Not really sure what to make of this.  I think I'll try the same test on another node and see if I see the same results.
[maybe_inline_appender.pdf](https://github.com/ceph/ceph/files/419356/maybe_inline_appender.pdf)

branch-predictor:
On Mon, Aug 15, 2016 at 07:16:47PM -0700, Mark Nelson wrote:
> I did a quick test with the memcpy changed to maybe_inline_memcpy in both safe_appender and unsafe_appender:
> 
> `maybe_inline_memcpy(pos, p, l, 32);`
> 
> I compiled and did a quick test on my laptop in both cases and saw some interesting results.  Typically the maybe_inline_memcpy seemed to be about 3-5% faster, except in two cases.  Safe append appeared to be ~36% faster while safe append_v appeared to be ~51% slower.  Not really sure what to make of this.  I think I'll try the same test on another node and see if I see the same results.
> [maybe_inline_appender.pdf](https://github.com/ceph/ceph/files/419356/maybe_inline_appender.pdf)

Makes perfect sense. In case of append (non-"_v"), replacing memcpy with
maybe_inline_memcpy causes it to go around the algorithm selection logic and
effectively behave as append_v. Append_v, on the other hand, is a template
which uses direct assignment and is translated by compiler to inlined series
of MOVs, just like maybe_inline_memcpy, sans block size selection (the
"switch" statement in maybe_inline_memcpy).

markhpc:
@branch-predictor Why was append_v test faster with the original memcpy code though?

branch-predictor:
On Tue, Aug 16, 2016 at 05:05:26AM -0700, Mark Nelson wrote:
> @branch-predictor Why was append_v test faster with the original memcpy code though?

There is no memcpy call in append_v, only assignment. Compiler does implicit
translation of assignment to a bunch of assembly MOV (memory move)
statements tailored to a particular type in use, making it already fast, and 
using maybe_inline_memcpy only hurts performance in this case.

markhpc:
@branch-predictor I'm still left wondering why switching from memcpy to maybe_inline_memcpy is hurting append_v consistently by so much in those tests...

markhpc:
I got impatient because this is so awesome and decided to throw together a rough first pass at using safe_appender in bluestore small_encode.  I didn't spend much time optimizing it or use it anywhere else (there could be significant other gains!).  I got a ~57% speed up for 4k random writes on incerta using 4 P3700 NVMe cards vs master from about 2 weeks ago.  It looks like we are still CPU bound on incerta so I suspect smartly using the unsafe appender both here and in more places is going to help a lot!

https://github.com/markhpc/ceph/tree/wip-bluestore-append

markhpc:
I spent some time using this over the weekend and discovered a couple of things that I would really like to have:

1) The ability to go back and rewrite a portion of the appended data at some offset.  
2) ability to get the current position (relative to start = 0 is fine).
3*) A fast way to append a buffer::ptr (A simple wrapper around copy_out, and make copy_out faster?).

These two things make DECODE_START and DECODE_FINISH implementable with appenders. This is fairly straightforward for unsafe_appender in it's current state, but safe_appender is trickier as a flush might happen behind your back.  I guess one option might be to track how much data has been appended and either modify the appended data, the bufferlist, or both, depending on the offset and length of the rewrite data.

With those things, it's fairly straightforward to use a single appender to encode an entire Bluestore::BlobMap so long as you are willing to estimate the size in advance.  Even using a single AP in the BlobMap encode loop for the id and bluestore_blob_t gives us another ~10% over the earlier results, though my test code is fairly hacky (though there is also more room for improvement!).

PR #10702, opened 2016-08-12
msg/async/AsyncMessenger: fix osd crash when set ms_async_op_threads=1

sysnote:
set ms_async_op_threads to 1, when osd start, will create several asyncmessengers, which call WorkerPool::get_worker(),
because of the worker load balance, will make a new worker and current_best->create("ms_async_worker"), this is before WorkerPool::start(),
which will fail the assert "  assert(global_centers && !global_centers->centers[idx])", so we need to check whether workerpool started in
WorkerPool::get_worker().

Fixs:http://tracker.ceph.com/issues/17001
Signed-off-by:Dong Wu<archer.wudong@gmail.com>


yuyuyu101:
hmm, it make sense. But we will disable smart thread spawn soon...

sysnote:
Thanks, looking forward to it.

PR #10707, opened 2016-08-12
rpm: Remove trailing whitespace in usermod command (SUSE)

tserong:
Trailing whitespace after the backslash on the -d line of the
usermod command effectively splits it in two, breaking the
rpm %pre script for SUSE builds.

Signed-off-by: Tim Serong tserong@suse.com


tserong:
@smithfarm, mind reviewing? (My test build with this change is yet to complete, but it should be good... give me half an hour or so)

smithfarm:
Good catch, @tserong! Note that this does not affect jewel or hammer.

tserong:
Just to confirm, my test build with this change was OK.

PR #10708, opened 2016-08-12
rbd: bench io-size should not be larger than image size

dillaman:
Fixes: http://tracker.ceph.com/issues/16967


trociny:
lgtm, though the tracker ticket message is not totally correct -- it seg faults only when "--io-size" is greater than image size (and not when equal to).

PR #10709, opened 2016-08-12
doc/install: Use https:// for download.ceph.com

cgwalters:
While the Ceph RPMs etc. are GPG signed, there are various
attacks against signed RPMs without signed/trusted *metadata*
that are elaborated on in https://theupdateframework.github.io/

Since download.ceph.com handles it today, let's make it the default.
(Might want to run this by the server admins though).


liewegas:
@dmick should be good right?

dmick:
Yep, looks fine to me.  Surprised they still say http, honestly.

PR #10711, opened 2016-08-12
librbd: fix possible inconsistent state when disabling mirroring fails

dillaman:


src/librbd/internal.cc
```diff
@@ -910,6 +910,7 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force,
 	if (r < 0) {
 	  lderr(cct) << "error opening image: "
 		     << cpp_strerror(r) << dendl;
+          delete imctx;
```
trociny: @dillaman Also missing close() after "error to flatten image" below?

dillaman:
@trociny good catch -- fixed

trociny:
lgtm

PR #10716, opened 2016-08-12
qa: add rados test script for upgrades

jdurgin:
Disable master tests that are not meant to work on jewel


jdurgin:
Fixes the rados test.sh failures from http://pulpito.ceph.com/teuthology-2016-08-12_04:20:02-upgrade:jewel-x-master-distro-basic-vps/

as seen here:

http://pulpito.ceph.com/joshd-2016-08-12_04:29:31-upgrade:jewel-x-master---basic-vps/

liewegas:
:+1:

PR #10720, opened 2016-08-13
msg/simple/Accepter.cc: replace shutdown() with selfpipe event in poll() (FreeBSD)

wjwithagen:
…socket.

In FreeBSD a shutdown on the socket here can generate:
  accepter.stop shutdown failed:  errno 57 (57) Socket is not connected
But that does not trigger an event with the poll on the socket.

Closing the socket does.

### DIFF
--- a/src/msg/simple/Accepter.cc
+++ b/src/msg/simple/Accepter.cc
@@ -69,7 +69,7 @@ int Accepter::bind(const entity_addr_t &bind_addr, const set<int>& avoid_ports)
 {
   const md_config_t *conf = msgr->cct->_conf;
   // bind to a socket
-  ldout(msgr->cct,10) << "accepter.bind" << dendl;
+  ldout(msgr->cct,10) <<  __func__ << dendl;
   
   int family;
   switch (bind_addr.get_family()) {
@@ -85,16 +85,18 @@ int Accepter::bind(const entity_addr_t &bind_addr, const set<int>& avoid_ports)
 
   /* socket creation */
   listen_sd = ::socket(family, SOCK_STREAM, 0);
+  ldout(msgr->cct,10) <<  __func__ << " socket sd: " << listen_sd << dendl;
   if (listen_sd < 0) {
-    lderr(msgr->cct) << "accepter.bind unable to create socket: "
+    lderr(msgr->cct) << __func__ << " unable to create socket: "
 		     << cpp_strerror(errno) << dendl;
     return -errno;
   }
 
   if (set_close_on_exec(listen_sd)) {
-    lderr(msgr->cct) << "accepter.bind unable to set_close_exec(): "
+    lderr(msgr->cct) << __func__ << " unable to set_close_exec(): "
 		     << cpp_strerror(errno) << dendl;
   }
+  
 
   // use whatever user specified (if anything)
   entity_addr_t listen_addr = bind_addr;
@@ -107,7 +109,8 @@ int Accepter::bind(const entity_addr_t &bind_addr, const set<int>& avoid_ports)
   for (int i = 0; i < conf->ms_bind_retry_count; i++) {
 
     if (i > 0) {
-        lderr(msgr->cct) << "accepter.bind was unable to bind. Trying again in " << conf->ms_bind_retry_delay << " seconds " << dendl;
+        lderr(msgr->cct) << __func__ << " was unable to bind. Trying again in " 
+			 << conf->ms_bind_retry_delay << " seconds " << dendl;
         sleep(conf->ms_bind_retry_delay);
     }
 
@@ -118,7 +121,7 @@ int Accepter::bind(const entity_addr_t &bind_addr, const set<int>& avoid_ports)
         int on = 1;
         rc = ::setsockopt(listen_sd, SOL_SOCKET, SO_REUSEADDR, &on, sizeof(on));
         if (rc < 0) {
-            lderr(msgr->cct) << "accepter.bind unable to setsockopt: "
+            lderr(msgr->cct) << __func__ << " unable to setsockopt: "
                              << cpp_strerror(errno) << dendl;
             r = -errno;
             continue;
@@ -127,14 +130,15 @@ int Accepter::bind(const entity_addr_t &bind_addr, const set<int>& avoid_ports)
         rc = ::bind(listen_sd, listen_addr.get_sockaddr(),
 		    listen_addr.get_sockaddr_len());
         if (rc < 0) {
-            lderr(msgr->cct) << "accepter.bind unable to bind to " << listen_addr
+            lderr(msgr->cct) << __func__ << " unable to bind to " << listen_addr
 			     << ": " << cpp_strerror(errno) << dendl;
             r = -errno;
             continue;
         }
     } else {
         // try a range of ports
-        for (int port = msgr->cct->_conf->ms_bind_port_min; port <= msgr->cct->_conf->ms_bind_port_max; port++) {
+        for (int port = msgr->cct->_conf->ms_bind_port_min; 
+		port <= msgr->cct->_conf->ms_bind_port_max; port++) {
             if (avoid_ports.count(port))
                 continue;
 
@@ -145,16 +149,18 @@ int Accepter::bind(const entity_addr_t &bind_addr, const set<int>& avoid_ports)
                 break;
         }
         if (rc < 0) {
-            lderr(msgr->cct) << "accepter.bind unable to bind to " << listen_addr
+            lderr(msgr->cct) <<  __func__ << "  unable to bind to " << listen_addr
                              << " on any port in range " << msgr->cct->_conf->ms_bind_port_min
                              << "-" << msgr->cct->_conf->ms_bind_port_max
                              << ": " << cpp_strerror(errno)
                              << dendl;
             r = -errno;
-            listen_addr.set_port(0); //Clear port before retry, otherwise we shall fail again.
+            // Clear port before retry, otherwise we shall fail again.
+            listen_addr.set_port(0); 
             continue;
         }
-        ldout(msgr->cct,10) << "accepter.bind bound on random port " << listen_addr << dendl;
+        ldout(msgr->cct,10) << __func__ << " bound on random port " 
+			    << listen_addr << dendl;
     }
 
     if (rc == 0)
@@ -163,8 +169,9 @@ int Accepter::bind(const entity_addr_t &bind_addr, const set<int>& avoid_ports)
 
   // It seems that binding completely failed, return with that exit status
   if (rc < 0) {
-      lderr(msgr->cct) << "accepter.bind was unable to bind after " << conf->ms_bind_retry_count
-                       << " attempts: " << cpp_strerror(errno) << dendl;
+      lderr(msgr->cct) << __func__ << " was unable to bind after " 
+		       << conf->ms_bind_retry_count << " attempts: " 
+		       << cpp_strerror(errno) << dendl;
       ::close(listen_sd);
       listen_sd = -1;
       return r;
@@ -176,7 +183,8 @@ int Accepter::bind(const entity_addr_t &bind_addr, const set<int>& avoid_ports)
   rc = getsockname(listen_sd, (sockaddr*)&ss, &llen);
   if (rc < 0) {
     rc = -errno;
-    lderr(msgr->cct) << "accepter.bind failed getsockname: " << cpp_strerror(rc) << dendl;
+    lderr(msgr->cct) << __func__ << " failed getsockname: " 
+		     << cpp_strerror(rc) << dendl;
     ::close(listen_sd);
     listen_sd = -1;
     return rc;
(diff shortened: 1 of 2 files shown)


wjwithagen:
@liewegas 
Could drop the last 2 commits if you prefer.
But I was there anyways, I had this "cleanup" in my tree anyways.

src/msg/simple/Accepter.cc
```diff
@@ -240,20 +240,20 @@ void *Accepter::entry()
...
 
-  struct pollfd pfd;
-  pfd.fd = listen_sd;
-  pfd.events = POLLIN | POLLERR | POLLNVAL | POLLHUP;
+  struct pollfd pfd[1];
+  pfd[0].fd = listen_sd;
+  pfd[0].events = POLLIN | POLLERR | POLLNVAL | POLLHUP;
   while (!done) {
     ldout(msgr->cct,20) << "accepter calling poll" << dendl;
-    int r = poll(&pfd, 1, -1);
```
yuyuyu101: why not &?

src/msg/simple/Accepter.cc
```diff
@@ -290,26 +290,37 @@ void *Accepter::entry()
...
-  // wait for thread to stop before closing the socket, to avoid
-  // racing against fd re-use.
   if (is_started()) {
+      ldout(msgr->cct,0) << __func__ << " wait for thread to join." << dendl;
     join();
   }
 
   if (listen_sd >= 0) {
-    ::close(listen_sd);
+    if ( ::close(listen_sd) < 0 ) {
```
yuyuyu101: extra space

wjwithagen:
@liewegas @yuyuyu101 
I replaced the shutdown signaling with a selfpipe that is taken along in the FDs that are poll()ed.
Tests on FreeBSD work for the test where it otherwise fails.
Putting it into jenkins to see what happens under Linux.

wjwithagen:
@liewegas 
I noticed that the same idea was also used in ./src/common/admin_socket.cc.
So was there any particular reason not to do the same here?

And there are a few more places where I find shutdown(..., SHUT_..), I guess that those places are also prone to this problem?

wjwithagen:
@liewegas 
This mimics the code already there as close as I can get it.
Can you run this thru QA/Teuthology, please

src/msg/simple/Accepter.cc
```diff
@@ -239,21 +255,46 @@ void *Accepter::entry()
...
   
   int errors = 0;
+  int ch;
 
-  struct pollfd pfd;
-  pfd.fd = listen_sd;
-  pfd.events = POLLIN | POLLERR | POLLNVAL | POLLHUP;
+  struct pollfd pfd[2];
+  memset(pfd, 0, sizeof(pfd));
+  create_selfpipe(&shutdown_rd_fd, &shutdown_wr_fd);
```
liewegas: maybe this selfpipe should be created in the ctor instead of entry() so we avoid an entry() vs stop() race...

src/msg/simple/Accepter.cc
```diff
@@ -280,31 +321,51 @@ void *Accepter::entry()
...
 {
   done = true;
-  ldout(msgr->cct,10) << "stop accepter" << dendl;
-
-  if (listen_sd >= 0) {
-    ::shutdown(listen_sd, SHUT_RDWR);
+  ldout(msgr->cct,10) << __func__ << " accept listening on: " << listen_sd << dendl;
+
+  if (shutdown_wr_fd < 0)
+    return;
```
liewegas: then this goes away

src/msg/simple/Accepter.cc
```diff
@@ -280,31 +321,51 @@ void *Accepter::entry()
...
-  }
-  ldout(msgr->cct,10) << "accepter stopping" << dendl;
+  // if (listen_sd >= 0) {
+  //   ::close(listen_sd);
+  //   listen_sd = -1;
+  // }
+  if (shutdown_rd_fd >= 0) {
+       ::close(shutdown_rd_fd);
+       shutdown_rd_fd = -1;
+   }
```
liewegas: and this close, along with wr_fd, move into stop along with its sibling

liewegas:
minor nits with fd init/close, but otherwise this looks good, thanks!!

src/msg/simple/Accepter.cc
wjwithagen: @liewegas 
Makes sense ... Will give it as shot

wjwithagen: @liewegas 
Not as simple as I though, since it is a scoping problem.
The pipe is now local to entry(), and to use it in the poll() it needs to be known there.
Poll is destructive on the file descriptors struct, so you need to reinit the struct every time over.
And thus it has to become a private member of the class.
Lets see how that works out.

wjwithagen:
@liewegas 
Not sure if the error is due to the changes I made.
Got the feeling that it is not using code I modified

src/msg/simple/Accepter.cc
```diff
@@ -198,7 +214,13 @@ int Accepter::bind(const entity_addr_t &bind_addr, const set<int>& avoid_ports)
 
   msgr->init_local_connection();
 
-  ldout(msgr->cct,1) << "accepter.bind my_inst.addr is " << msgr->get_myaddr()
+  rc = create_selfpipe(&shutdown_rd_fd, &shutdown_wr_fd);
+  if (rc < 0) {
+    lderr(msgr->cct) <<  __func__ << " unable to create signalling pipe " << listen_addr
+		     << ": " << cpp_strerror(rc) << dendl;
```
yuyuyu101: return rc? otherwise, poll will use uninitilazed value

wjwithagen: @yuyuyu101 
Good catch, was in my original code, but forgot to commit it.

wjwithagen:
@liewegas 
Any chance of merging this?

liewegas:
@wjwithagen sorry, lost track of this.  Marked with needs-qa.  Thanks!

wjwithagen:
@liewegas 
No problem. I have trouble keeping up with my own PRs. 
Let alone that everybody is knocking at your door.

wjwithagen:
@liewegas 
Ping?

tchaikov:
@wjwithagen we just need to run your PR through a rados qa test. i will do it this week-end.

wjwithagen:
@tchaikov 
I was under the assumption that that sort went automagically when things were tagged, but it only took some (longer) time...
Sure, thanx....

src/msg/simple/Accepter.cc
yuyuyu101: @wjwithagen this?

src/msg/simple/Accepter.cc
```diff
@@ -263,50 +321,84 @@ void *Accepter::entry()
...
     listen_sd = -1;
   }
+  if (shutdown_rd_fd >= 0) {
+	if (::close(shutdown_rd_fd) < 0) {
+      ldout(msgr->cct,10) << __func__ << "close shutdown_rd_fd failed: "
+	      << " errno " << errno << " " << cpp_strerror(errno) << dendl;
+    }
+	shutdown_rd_fd = -1;
+  }
+  if (shutdown_wr_fd >= 0) {
```
yuyuyu101: shutdown_wr_fd must be -1 now

src/msg/simple/Accepter.cc
```diff
@@ -263,50 +321,84 @@ void *Accepter::entry()
...
+  }  else {
+      ldout(msgr->cct,15) << __func__ << " signaled poll" << dendl;
   }
+  VOID_TEMP_FAILURE_RETRY(close(shutdown_wr_fd));
+  shutdown_wr_fd = -1;
 
   // wait for thread to stop before closing the socket, to avoid
   // racing against fd re-use.
   if (is_started()) {
+      ldout(msgr->cct,0) << __func__ << " wait for thread to join." << dendl;
```
yuyuyu101: suggest level 1

wjwithagen: @yuyuyu101 
Yup, thanx.... Could even be a bit higher, it is really for debugging thread signaling

src/msg/simple/Accepter.cc
wjwithagen: @yuyuyu101 
You mean, it should have been already closed by the other part of the thread?
But it does not hurt to have it here?

It is here to really prevent fd from "escaping" while still open.
Perhaps I should first make this work with an assert, just to be safe and it really doesn't allow file-handle leakage.

wjwithagen: @yuyuyu101 
I see what you mean. `shutdown_wr_fd` is immediately closed right after writing to the pipe. 
So this second attempt can go.
Good catch.

I moved the code actually here from the entry() function on request of Sage.
But did not look up in the same code block.

tchaikov:
not all comments addressed. dropping needs-qa.

wjwithagen:
@tchaikov 
What comment do you think is not addressed?

src/msg/simple/Accepter.cc
```diff
@@ -280,33 +327,67 @@ void *Accepter::entry()
 
   ldout(msgr->cct,20) << "accepter closing" << dendl;
   // don't close socket, in case we start up again?  blech.
-  if (listen_sd >= 0) {
-    ::close(listen_sd);
-    listen_sd = -1;
-  }
-  ldout(msgr->cct,10) << "accepter stopping" << dendl;
+  // if (listen_sd >= 0) {
```
tchaikov: could you remove the commented code?

src/msg/simple/Accepter.cc
```diff
@@ -259,7 +270,7 @@ int Accepter::start()
 
 void *Accepter::entry()
 {
-  ldout(msgr->cct,10) << "accepter starting" << dendl;
+  ldout(msgr->cct,1) << __func__ << " start" << dendl;
```
tchaikov: why are you changing this log level?

src/msg/simple/Accepter.cc
```diff
@@ -310,31 +321,31 @@ void *Accepter::entry()
...
   // don't close socket, in case we start up again?  blech.
   // if (listen_sd >= 0) {
   //   ::close(listen_sd);
   //   listen_sd = -1;
   // }
   if (shutdown_rd_fd >= 0) {
-       ::close(shutdown_rd_fd);
-       shutdown_rd_fd = -1;
-   }
+    ::close(shutdown_rd_fd);
```
tchaikov: could you squash this change into the previous one?

src/msg/simple/Accepter.cc
```diff
@@ -352,41 +363,34 @@ void Accepter::stop()
...
-	if (::close(shutdown_rd_fd) < 0) {
-      ldout(msgr->cct,10) << __func__ << "close shutdown_rd_fd failed: "
-	      << " errno " << errno << " " << cpp_strerror(errno) << dendl;
-    }
-	shutdown_rd_fd = -1;
-  }
-  if (shutdown_wr_fd >= 0) {
-	if (::close(shutdown_wr_fd) < 0) {
-      ldout(msgr->cct,10) << __func__ << "close shutdown_rd_fd failed: "
+    if (::close(shutdown_rd_fd) < 0) {
```
tchaikov: ditto.

tchaikov:
@wjwithagen sorry, your cleanup commit is interleaved with the fixes of another commit. so i missed them.

wjwithagen:
@tchaikov 
Yup, that was me being clumpsy with git rebase....
Let me know if I need to do anything else

src/msg/simple/Accepter.cc
wjwithagen: @tchaikov 
I left it there because it was the original code their, and part of the discussion with Sage on the possibility of having a race with the FD number being closed and opened.
Let me know if it needs to go.

src/msg/simple/Accepter.cc
wjwithagen: @tchaikov 
Changing it to be consistent with other "start" messages.

src/msg/simple/Accepter.cc
tchaikov: i am not sure about this. i think the comment would suffice.

tchaikov:
> Let me know if I need to do anything else

@wjwithagen, https://github.com/ceph/ceph/pull/10720/commits/0eedaa0adc5997faa26ac56aa13b8d2abae50be4#r78334571 would be ideal if you can address this.

and get the confirmation from @yuyuyu101 .

src/msg/simple/Accepter.cc
tchaikov: i am not sure about this. i think the comment would suffice.

wjwithagen:
@tchaikov 
You are referring to actual code changes in the commit about logging ?

Not sure on how to actually move part of a commit into another commit in the same PR.
Short of cleaning things out, and redoing them.
Any suggestions on how to do this?

tchaikov:
> Any suggestions on how to do this?
```
git reset HEAD^
git add -p
```

wjwithagen:
@tchaikov 
Git is not my friend, but after hitting myself with the hamer, we have 2 commits:
 - Code change
 - Log description and level additions and changes

src/msg/simple/Accepter.cc
```diff
@@ -49,11 +51,24 @@ static int set_close_on_exec(int fd)
   return 0;
 }
 
+int Accepter::create_selfpipe(int *pipe_rd, int *pipe_wr) {
```
tchaikov:
```
/home/jenkins-build/build/workspace/ceph-pull-requests/src/msg/simple/Accepter.cc:54:57: error: no 'int Accepter::create_selfpipe(int*, int*)' member function declared in class 'Accepter'
 int Accepter::create_selfpipe(int *pipe_rd, int *pipe_wr) {
                                                         ^
```

@wjwithagen ^

wjwithagen:
@tchaikov 
Forgot to include the .h file in the commit.
Should go better this time.

wjwithagen:
@tchaikov 
That helped.

tchaikov:

failed tests are tracked at

wjwithagen:
@tchaikov @liewegas @yuyuyu101 
Thanx for this merge....

Next big thing is getting the work on EventKqueue in space to be mergeable.

In the mean time I actually reached the point last night that I could run 132 out of 133 tests.
run-tox-ceph-disk is failling for obvious FreeBSD reasons.
But I'm a careful happy camper...
Lets hope this is still the same case after I rebase to get in line with HEAD.

PR #10722, opened 2016-08-14
mon/MonCap.h: add std::move for std::string

stiopaa1:


tchaikov:

failed tests are tracked at

PR #10723, opened 2016-08-14
os/FuseStore: fix several FuseStore issues

liewegas:
This fixes several issues.  Now passes updated scrub_test.py 50x.

https://github.com/ceph/ceph-qa-suite/pull/1118

### DIFF
--- a/qa/workunits/objectstore/test_fuse.sh
+++ b/qa/workunits/objectstore/test_fuse.sh
@@ -81,8 +81,12 @@ test ! -s $MNT/meta/all/#-1:7b3f43c4:::osd_superblock:0#/data
 echo asdfasdfasdf > $MNT/meta/all/#-1:7b3f43c4:::osd_superblock:0#/data
 test -s $MNT/meta/all/#-1:7b3f43c4:::osd_superblock:0#/data
 grep -q asdfasdfasdf $MNT/meta/all/#-1:7b3f43c4:::osd_superblock:0#/data
+truncate --size 4 $MNT/meta/all/#-1:7b3f43c4:::osd_superblock:0#/data
+stat --format=%s $MNT/meta/all/#-1:7b3f43c4:::osd_superblock:0#/data | grep -q ^4$
+expect_false grep -q asdfasdfasdf $MNT/meta/all/#-1:7b3f43c4:::osd_superblock:0#/data
 rm $MNT/meta/all/#-1:7b3f43c4:::osd_superblock:0#/data
 test ! -s $MNT/meta/all/#-1:7b3f43c4:::osd_superblock:0#/data
+
 
 # create pg collection
 mkdir --mode 0003 $MNT/0.0_head
--- a/qa/workunits/objectstore/test_fuse.sh
+++ b/qa/workunits/objectstore/test_fuse.sh
@@ -20,7 +20,7 @@ MNT=store_test_fuse_mnt
 rm -rf $DATA
 mkdir -p $DATA
 
-fusermount -u $MNT || true
+test -d $MNT && fusermount -u $MNT || true
 rmdir $MNT || true
 mkdir $MNT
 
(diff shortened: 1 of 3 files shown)


qa/workunits/objectstore/test_fuse.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh -ex
+#!/bin/bash -ex
```
tchaikov: we removed the last bit of bashim in 15ea140b3bbf6b1bfa9eea19b0f167141fcaa2c9, why would we want to switch to bash again?

qa/workunits/objectstore/test_fuse.sh
```diff
@@ -43,46 +43,46 @@ test -d $MNT/meta/all
 test -d $MNT/meta/by_bitwise_hash
 
 # create object
-mkdir $MNT/meta/all/@-1:7b3f43c4:::osd_superblock:0@
-test -e $MNT/meta/all/@-1:7b3f43c4:::osd_superblock:0@/data
-test -d $MNT/meta/all/@-1:7b3f43c4:::osd_superblock:0@/attr
-test -d $MNT/meta/all/@-1:7b3f43c4:::osd_superblock:0@/omap
-test -e $MNT/meta/all/@-1:7b3f43c4:::osd_superblock:0@/bitwise_hash
-test -e $MNT/meta/all/@-1:7b3f43c4:::osd_superblock:0@/omap_header
+mkdir $MNT/meta/all/#-1:7b3f43c4:::osd_superblock:0#
```
tchaikov: just a note for future readers, we started to use `#` since 4bd31faee0b0de908069287fa14c74dc889b41e5.

src/os/FuseStore.cc
```diff
@@ -772,63 +823,86 @@ static int os_create(const char *path, mode_t mode, struct fuse_file_info *fi)
...
 static int os_release(const char *path, struct fuse_file_info *fi)
 {
-  bufferlist *pbl = reinterpret_cast<bufferlist*>(fi->fh);
-  delete pbl;
+  dout(10) << __func__ << " " << path << dendl;
+  fuse_context *fc = fuse_get_context();
+  FuseStore *fs = static_cast<FuseStore*>(fc->private_data);
+  std::lock_guard<std::mutex> l(fs->lock);
+  FuseStore::OpenFile *o = reinterpret_cast<FuseStore::OpenFile*>(fi->fh);
+  if (--o->ref) {
```
tchaikov: should be `if (--o->ref == 0)`, i think.

src/os/FuseStore.cc
```diff
@@ -594,59 +623,77 @@ static int os_open(const char *path, struct fuse_file_info *fi)
 
   case FN_OBJECT_DATA:
     {
-      pbl = new bufferlist;
-      int r = fs->store->read(cid, oid, 0, 0, *pbl);
+      int r = fs->open_file(
+	path, fi,
+	[&](bufferlist *pbl) {
```
tchaikov: maybe we should pass the local variable `bpl` into `open_file()`, because if we go this way, `pbl` will still be `0` at https://github.com/ceph/ceph/pull/10723/commits/604f6b1f494f23d1964e3951ec9169faed5f1eac#diff-fc25a90da304d483f385da5c44750423R686, and hence the returned `fi` will be empty. so the following `os_read()` will run into a null `fi->fh` .

tchaikov:
@liewegas ping?

liewegas:
Sorry, got busy with the bluestore stuff.  Will get back to this 
this week!

src/os/FuseStore.cc
liewegas: open_file fills in fi->fh.. so that ending block isn't used/needed for the object_data case

liewegas:
Rebased, fixed some header issues (there was some cleanup in master), switched back to sh, and fixed the ref count issue.  Passing the tests.

Will reschedule the teuthology tests again once it builds...

xiexingguo:
retest this please

liewegas:
@tchaikov http://pulpito.ceph.com/sage-scrub-c/

tchaikov:
@liewegas i will review the change again tmr. and will merge it if it's good.

PR #10729, opened 2016-08-15
script: improve ceph-release-notes regex

smithfarm:
Tolerate multiple spaces after ':', remove duplicate "fs", add "log" and
"msgr".


dachary:
I would rather keep the mandatory prefix. It's a little tedious to fix but missing prefixes become less and less common.

smithfarm:
@dachary OK. Please take another look.

PR #10731, opened 2016-08-15
rgw: sync modules, metadata search

yehudasa:
Make data sync more modular, so that we could add sync modules. An example for such sync module is the "log" module that logs every object that needs to be synced. Other examples (yet to be implemented) are meta indexing module, or backup (to external storage) module.


src/rgw/rgw_sync_module.h
```diff
@@ -0,0 +1,141 @@
...
+
+  virtual bool supports_data_export() = 0;
+  virtual int create_instance(CephContext *cct, map<string, string>& config, RGWSyncModuleInstanceRef *instance) = 0;
+};
+
+typedef std::shared_ptr<RGWSyncModule> RGWSyncModuleRef;
+
+
+class RGWSyncModulesManager {
+  Mutex lock;
```
cbodley: unused

yehudasa:
@cbodley addressed you comments, repushed

cbodley:
looks good. is it passing test_multi.py?

yehudasa:
@cbodley yes

PR #10736, opened 2016-08-15
common/ceph_context: 'config diff get' option added

oliveiradan:
This is a feature request from Sebastien (on our ceph-devel list) so we can get the 'diff (current and default) of an config option': # config diff get mon_data_avail_warn 
(1):

```
./ceph --admin-daemon out/osd.1.asok config diff get mon_data_avail_crit 
{
    "diff": {
        "current": {
            "mon_data_avail_crit": "1"
        },
        "defaults": {
            "mon_data_avail_crit": "5"
        }
    },
    "unknown": []
}
```

(2):

```
./ceph --admin-daemon out/osd.1.asok config diff get mon_data_avail_warn
{
    "diff": {
        "current": {
            "mon_data_avail_warn": "10"
        },
        "defaults": {
            "mon_data_avail_warn": "30"
        }
    },
    "unknown": []
}
```

(3):

```
 ./ceph --admin-daemon out/osd.1.asok config diff get num_client
{
    "diff": {
        "current": {
            "num_client": "1"
        },
        "defaults": {
            "num_client": "1"
        }
    },
    "unknown": []
}
```


src/common/ceph_context.cc
```diff
@@ -425,7 +426,45 @@ void CephContext::do_command(std::string command, cmdmap_t& cmdmap,
...
+        f->close_section();   //-- diff
+
+        f->open_array_section("unknown");
+        for (set<string>::iterator p = unknown.begin(); p != unknown.end(); ++p) {
+          f->dump_string("option", *p);
+        } //-- for (set<string>::iterator p = unknown.begin(); p != unknown.end(); ++p)
+        f->close_section();   //-- unknown
+      } //-- else if (!cmd_getval(this, cmdmap, "var", ceph_setting))
+    } //-- else if (command == "config diff get")
+  else if (command == "log flush") {
```
badone: Can we align this better?

src/common/ceph_context.cc
```diff
@@ -425,7 +426,45 @@ void CephContext::do_command(std::string command, cmdmap_t& cmdmap,
...
+        def_conf.set_val("host", _conf->host);
+        def_conf.apply_changes(NULL);
+
+        map<string, pair<string, string> > diff;
+        set<string> unknown;
+        def_conf.diff_setting(_conf, &diff, &unknown, ceph_setting, true);
+        f->open_object_section("diff");
+        f->open_object_section("current");
+
+        for (map<string, pair<string, string> >::iterator p = diff.begin(); p != diff.end(); ++p) {
```
badone: Could we use C++11 style auto and range-based for loops instead of the older style?

src/common/config.cc
```diff
@@ -1227,6 +1227,47 @@ void md_config_t::diff(
...
+            std::map<std::string, std::pair<std::string, std::string>> *diff, 
+            std::set<std::string> *unknown, const std::string& ceph_setting, bool show_unchanged)
+{
+  Mutex::Locker l(lock);
+
+  bool wasItFound = false;
+  char local_buf[4096];
+  char other_buf[4096];
+  for (int i = 0; i < NUM_CONFIG_OPTIONS; i++) {
+    if (wasItFound) break;  //-- If we already found it, we are done.
```
badone: See https://github.com/ceph/ceph/blob/master/CodingStyle#L83

src/common/config.cc
```diff
@@ -1227,6 +1227,47 @@ void md_config_t::diff(
...
+      if (err < 0) {
+        if (err == -ENOENT) {
+          unknown->insert(opt->name);
+        } //-- if (err == -ENOENT)
+        continue;
+      } //-- if (err < 0)
+
+      char *local_val = local_buf;
+      err = _get_val(opt->name, &local_val, sizeof(local_buf));
+      if (err != 0) continue;
```
badone: See https://github.com/ceph/ceph/blob/master/CodingStyle#L83

src/common/config.cc
```diff
@@ -1227,6 +1227,47 @@ void md_config_t::diff(
...
+      if (strcmp(local_val, other_val)) diff->insert(make_pair(opt->name, make_pair(local_val, other_val)));
```
badone: See https://github.com/ceph/ceph/blob/master/CodingStyle#L83

src/common/config.cc
```diff
@@ -1227,6 +1227,47 @@ void md_config_t::diff(
...
+      else {
+        if (show_unchanged) diff->insert(make_pair(opt->name, make_pair(local_val, other_val)));
```
badone: See https://github.com/ceph/ceph/blob/master/CodingStyle#L83

src/common/ceph_context.cc
```diff
@@ -425,7 +426,45 @@ void CephContext::do_command(std::string command, cmdmap_t& cmdmap,
         f->dump_string("option", *p);
       }
       f->close_section(); // unknown
-    } else if (command == "log flush") {
+    }
+    else if (command == "config diff get") {
+      std::string ceph_setting;
+      if (!cmd_getval(this, cmdmap, "var", ceph_setting)) {
+        f->dump_string("error", "syntax error: 'config diff get <var>'");
+      } //-- if (!cmd_getval(this, cmdmap, "var", ceph_setting))
```
tchaikov: nit, could remove the comment, the `if` block is small enough.

src/common/config.h
```diff
@@ -156,8 +156,13 @@ struct md_config_t {
   void show_config(Formatter *f);
 
   /// obtain a diff between our config values and another md_config_t values
-  void diff(const md_config_t *other,
-            map<string,pair<string,string> > *diff, set<string> *unknown);
+  void diff(const md_config_t *other, map<string,pair<string,string> > *diff,
```
tchaikov: could you leave this unchanged if you are not touching it in this commit?

src/common/config.cc
```diff
@@ -1227,6 +1227,47 @@ void md_config_t::diff(
...
+  Mutex::Locker l(lock);
+
+  bool wasItFound = false;
+  char local_buf[4096];
+  char other_buf[4096];
+  for (int i = 0; i < NUM_CONFIG_OPTIONS; i++) {
+    if (wasItFound) break;  //-- If we already found it, we are done.
+    config_option *opt = &config_optionsp[i];
+    std::string ceph_setting_name(opt->name);
+    if (ceph_setting == ceph_setting_name) {
```
tchaikov: nit, could just use `opt->name` for comparison.

src/common/config.cc
```diff
@@ -1227,6 +1227,47 @@ void md_config_t::diff(
   }
 }
 
+void md_config_t::diff_setting(const md_config_t *other,
+            std::map<std::string, std::pair<std::string, std::string>> *diff, 
+            std::set<std::string> *unknown, const std::string& ceph_setting, bool show_unchanged)
+{
+  Mutex::Locker l(lock);
+
+  bool wasItFound = false;
```
tchaikov: our naming convention does not encourage `wasItFound`, might want to use `hello_world` instead.

src/common/config.h
```diff
@@ -156,8 +156,13 @@ struct md_config_t {
   void show_config(Formatter *f);
 
   /// obtain a diff between our config values and another md_config_t values
-  void diff(const md_config_t *other,
-            map<string,pair<string,string> > *diff, set<string> *unknown);
+  void diff(const md_config_t *other, map<string,pair<string,string> > *diff, 
+            set<string> *unknown);
+
+  //-- obtain a diff between our config values and another md_config_t values
```
tchaikov: could you use `///` to start the comment for a function or method. `///` is used by doxygen to start a short description for a function.

src/common/config.cc
```diff
@@ -1227,6 +1227,47 @@ void md_config_t::diff(
...
+      } //-- if (err < 0)
+
+      char *local_val = local_buf;
+      err = _get_val(opt->name, &local_val, sizeof(local_buf));
+      if (err != 0) continue;
+      if (strcmp(local_val, other_val)) diff->insert(make_pair(opt->name, make_pair(local_val, other_val)));
+      else {
+        if (show_unchanged) diff->insert(make_pair(opt->name, make_pair(local_val, other_val)));
+      } //-- if (strcmp(local_val, other_val)) diff->insert(make_pair(opt->name, make_pair(local_val, other_val)))
+      wasItFound = true;
```
tchaikov: why not just `break` here?

src/common/config.cc
tchaikov: please wrap at 80 chars, see https://google.github.io/styleguide/cppguide.html#Line_Length

src/common/config.cc
tchaikov: is it possible that `ceph_setting` can match multiple times?

src/common/config.cc
```diff
@@ -1227,6 +1227,47 @@ void md_config_t::diff(
   }
 }
 
+void md_config_t::diff_setting(const md_config_t *other,
```
tchaikov: maybe be we can consolidate `diff()` and `diff_setting()`?

src/common/config.cc
```diff
@@ -1227,6 +1227,47 @@ void md_config_t::diff(
...
+            std::map<std::string, std::pair<std::string, std::string>> *diff, 
+            std::set<std::string> *unknown, const std::string& ceph_setting, bool show_unchanged)
```
tchaikov: nit, the `std::` is not necessary, as we are `using std::string` and `using std::map` at the top of this source file.

badone:
Could we get a test for this also please @oliveiradan ?

src/common/config.cc
```diff
@@ -1227,6 +1227,47 @@ void md_config_t::diff(
...
+    std::string ceph_setting_name(opt->name);
+    if (ceph_setting == ceph_setting_name) {
+      memset(local_buf, 0, sizeof(local_buf));
+      memset(other_buf, 0, sizeof(other_buf));
+
+      char *other_val = other_buf;
+      int err = other->get_val(opt->name, &other_val, sizeof(other_buf));
+      if (err < 0) {
+        if (err == -ENOENT) {
+          unknown->insert(opt->name);
```
tchaikov: if only one item can be matched, why bother returning `unknown`?

tchaikov:
could you prefix the title of your commit message with the subcomponent your are changing ? see https://github.com/ceph/ceph/blob/master/SubmittingPatches.rst#preparing-and-sending-patches, in this case, the prefix could be "common/ceph_context: ".

and would be ideal if you could put more details about the feature in your commit message?

src/common/ceph_context.cc
oliveiradan: We already had some misaligned block, so I ended up fixing some up the function in order to see this one properly.

src/common/ceph_context.cc
oliveiradan: Done.

src/common/config.h
oliveiradan: Done. Thanks for the heads up.

src/common/config.h
oliveiradan: Done.

src/common/config.cc
oliveiradan: Done, thanks.

src/common/config.cc
oliveiradan: Done, thanks.

src/common/config.cc
oliveiradan: Done, thanks.

src/common/config.cc
oliveiradan: I have not seen any cases where config_optionsp[] would have duplicated values. If so, we would need to think about changing the array in question to a hash (as I suggested during the  work done for  http://tracker.ceph.com/issues/14934).

src/common/ceph_context.cc
oliveiradan: I understand that, but if we have nested blocks, it would look better/more organized. If not *really allowed*, I will take them off.

src/common/config.cc
oliveiradan: Thanks.

src/common/config.cc
oliveiradan: Done, thanks.

src/common/config.cc
oliveiradan: Yep, I agreed. Thanks!

src/common/config.cc
oliveiradan: As md_config_t::diff() does it, we just wanted to keep it the same way/behavior. However it does make sense and I can remove it.

src/common/config.cc
oliveiradan: Humm, we only decided to have a different function **diff_setting()** because we would need to change the **diff()** signature and behavior if a setting was passed or not. But it is a possibility.

src/common/config.cc
oliveiradan: Yes, but probably also not wrong. However, I went ahead and remove it to expedite the process.

src/common/config.cc
oliveiradan: @badone, but 'get a test' do you mean "Add a test for it in the test_config.cc"? 

Thanks.

src/common/ceph_context.cc
tchaikov: our code guidelines do not forbid this explicitly. and i agree this is a gray area, but see https://google.github.io/styleguide/cppguide.html#Implementation_Comments. the scope of the `if` block is obvious, imo.

badone:
@oliveiradan yes, add a test or tests for the new functionality where appropriate.

src/common/ceph_context.cc
```diff
@@ -425,7 +426,48 @@ void CephContext::do_command(std::string command, cmdmap_t& cmdmap,
...
+        def_conf.set_val("host", _conf->host);
+        def_conf.apply_changes(NULL);
+
+        map<string, pair<string, string> > diff;
+        set<string> unknown;
+        def_conf.diff_setting(_conf, &diff, &unknown, ceph_setting, true);
+        f->open_object_section("diff");
+        f->open_object_section("current");
+
+        //for (map<string, pair<string, string> >::iterator p = diff.begin(); p != diff.end(); ++p) {
```
tchaikov: please remove the commented out code.

src/common/config.h
```diff
@@ -156,9 +156,13 @@ struct md_config_t {
   void show_config(Formatter *f);
 
   /// obtain a diff between our config values and another md_config_t values
-  void diff(const md_config_t *other,
+  void diff(const md_config_t *other,
```
tchaikov: please remove the trailing space.

src/common/ceph_context.cc
```diff
@@ -345,10 +346,10 @@ void CephContext::do_command(std::string command, cmdmap_t& cmdmap,
     if (!cmd_getval(this, cmdmap, "var", var)) {
       f->dump_string("error", "syntax error: 'perf reset <var>'");
     } else {
-     if(!_perf_counters_collection->reset(var))
+      if(!_perf_counters_collection->reset(var))
```
tchaikov: could you squash this change into the previous one?

src/common/config.cc
tchaikov: yes, as you are almost duplicating the `diff()` method, so it might be worthy a try.

src/common/ceph_context.cc
```diff
@@ -326,6 +326,7 @@ void CephContext::do_command(std::string command, cmdmap_t& cmdmap,
   }
   lgeneric_dout(this, 1) << "do_command '" << command << "' '"
 			 << ss.str() << dendl;
+
```
tchaikov: again, could you remove the formatting changes if you are not touch them?

tchaikov:
> and would be ideal if you could put more details about the feature in your commit message?

it is not addressed.

src/common/config.h
oliveiradan: Done.

src/common/ceph_context.cc
oliveiradan: Done.

src/common/ceph_context.cc
oliveiradan: Done.

src/common/config.cc
oliveiradan: Well, `md_config_t::diff_setting(const md_config_t *other, map<string, pair<string, string>> *diff, 
const string& ceph_setting, bool show_unchanged)`  and `void md_config_t::diff(const md_config_t *other, map<string,pair<string,string> > *diff, set<string> *unknown)` quite different signatures, so I wonder whether or not the trade offs on consolidating them would make maintenance troublesome.

src/common/ceph_context.cc
```diff
@@ -345,10 +345,10 @@ void CephContext::do_command(std::string command, cmdmap_t& cmdmap,
     if (!cmd_getval(this, cmdmap, "var", var)) {
       f->dump_string("error", "syntax error: 'perf reset <var>'");
     } else {
-     if(!_perf_counters_collection->reset(var))
+      if(!_perf_counters_collection->reset(var))
```
tchaikov: again, could you remove the formatting changes if you are not touch them?

oliveiradan: @tchaikov ,

Pls, based on comment https://github.com/ceph/ceph/pull/10736#discussion_r75415073  you can see my reply to @badone , (We already had some misaligned block, so I ended up fixing some up the function in order to see this one properly). Should we not touch/fix these kind of cases? That was the only reason why it was touched.

tchaikov:
@oliveiradan sorry for the latency, will review your PR next Monday.

src/common/ceph_context.cc
tchaikov: > Should we not touch/fix these kind of cases?

please don't, unless you happen to be on it. otherwise, it would be a pain to `git blame`.

src/common/ceph_context.cc
```diff
@@ -381,17 +381,17 @@ void CephContext::do_command(std::string command, cmdmap_t& cmdmap,
     } else if (command == "config get") {
       std::string var;
       if (!cmd_getval(this, cmdmap, "var", var)) {
-	f->dump_string("error", "syntax error: 'config get <var>'");
+        f->dump_string("error", "syntax error: 'config get <var>'");
```
tchaikov: see https://github.com/ceph/ceph/blob/master/CodingStyle#L77. and again, please don't apply formatting only changes to the code you are not changing otherwise.

src/common/ceph_context.cc
```diff
@@ -425,7 +425,38 @@ void CephContext::do_command(std::string command, cmdmap_t& cmdmap,
...
+        def_conf.name = _conf->name;
+        def_conf.set_val("host", _conf->host);
+        def_conf.apply_changes(NULL);
+
+        map<string, pair<string, string> > diff;
+        def_conf.diff_setting(_conf, &diff, ceph_setting, true);
+        f->open_object_section("diff");
+        f->open_object_section("current");
+
+        for (auto p = begin(diff); p != end(diff); ++p) {
```
tchaikov: why not use range-based for loop?

src/common/ceph_context.cc
```diff
@@ -345,10 +345,10 @@ void CephContext::do_command(std::string command, cmdmap_t& cmdmap,
     if (!cmd_getval(this, cmdmap, "var", var)) {
       f->dump_string("error", "syntax error: 'perf reset <var>'");
     } else {
-     if(!_perf_counters_collection->reset(var))
+      if(!_perf_counters_collection->reset(var))
         f->dump_stream("error") << "Not find: " << var;
-     else
-       f->dump_string("success", command + ' ' + var);
+      else
```
tchaikov: ditto.

src/common/config.cc
tchaikov: they can be very different, if `diff` is not a `map`. and in your case, `diff` could have been a `pair<string, string>`, because the queried setting can only be matched once at most.

src/common/config.cc
```diff
@@ -1227,6 +1227,49 @@ void md_config_t::diff(
...
+      char *local_val = local_buf;
+      err = _get_val(opt->name, &local_val, sizeof(local_buf));
+      if (err != 0) {
+        continue;
+      } 
+      if (strcmp(local_val, other_val)) {
+        diff->insert(make_pair(opt->name, make_pair(local_val, other_val)));
+      }   
+      else {
+        if (show_unchanged) {
```
tchaikov: this parameter is always true.

src/common/config.cc
```diff
@@ -1227,6 +1227,49 @@ void md_config_t::diff(
...
+            const string& ceph_setting, bool show_unchanged)
+{
+  Mutex::Locker l(lock);
+
+  char local_buf[4096];
+  char other_buf[4096];
+  for (int i = 0; i < NUM_CONFIG_OPTIONS; i++) {
+    config_option *opt = &config_optionsp[i];
+    std::string ceph_setting_name(opt->name);
+    if (ceph_setting == ceph_setting_name) {
```
tchaikov: can just `continue` otherwise, to reduce the nested indent level.

src/common/config.cc
```diff
@@ -1227,6 +1227,49 @@ void md_config_t::diff(
...
+      else {
+        if (show_unchanged) {
+          diff->insert(make_pair(opt->name, make_pair(local_val, other_val)));
+        }
+      } 
+      //-- If we already found it, we are done.
+      break;  
+    } //-- if (ceph_setting == ceph_setting_name)
+  } //-- for (int i = 0; i < NUM_CONFIG_OPTIONS; i++)
+} /* void md_config_t::diff_setting(const md_config_t *other,
```
tchaikov: could you remove the comments?

tchaikov:
> and would be ideal if you could put more details about the feature in your commit message?

this comment is not addressed.

src/common/config.cc
oliveiradan: Not really. It can be set as false, so 'unchanged settings will not be shown'.

src/common/config.cc
oliveiradan: Not sure I understand what you mean here.

src/common/config.cc
oliveiradan: Done.

oliveiradan:
@tchaikov , would you mean something more detailed than what we have at the top of the PR https://github.com/ceph/ceph/pull/10736#issue-171281446 ? What am I missing?

src/common/config.cc
tchaikov:
```c++
if (ceph_setting != opt->name)
  continue;
```

src/common/config.cc
tchaikov: the only caller is `def_conf.diff_setting(_conf, &diff, ceph_setting, true);`.

tchaikov:
> What am I missing? 

"what we have at the top of the PR" is in the github comment, not in the commit message.

src/common/config.cc
oliveiradan: Kefu, 

Yes, for now and for that behavior. But what if something wants to call it `def_conf.diff_setting(_conf, &diff, ceph_setting, false);` , which will now show *unchanged setting*.

src/common/ceph_context.cc
oliveiradan: Done.

src/common/config.cc
```diff
@@ -1227,6 +1227,49 @@ void md_config_t::diff(
...
+        continue;
+      } 
+
+      char *local_val = local_buf;
+      err = _get_val(opt->name, &local_val, sizeof(local_buf));
+      if (err != 0) {
+        continue;
+      } 
+      if (strcmp(local_val, other_val)) {
+        diff->insert(make_pair(opt->name, make_pair(local_val, other_val)));
```
oliveiradan: I understand it, but the `map` uses it, not only as a key and it serves as a wrapper also for the `pair`. That's what allows us to use the *iterators*.

src/common/config.cc
oliveiradan: It does make sense. Thanks.

src/common/ceph_context.cc
```diff
@@ -381,17 +381,17 @@ void CephContext::do_command(std::string command, cmdmap_t& cmdmap,
...
-	}
+      	char buf[4096];
+      	memset(buf, 0, sizeof(buf));
+      	char *tmp = buf;
+      	int r = _conf->get_val(var.c_str(), &tmp, sizeof(buf));
+      	if (r < 0) {
+      	  f->dump_stream("error") << "error getting '" << var << "': " << cpp_strerror(r);
+      	} else {
+      	  f->dump_string(var.c_str(), buf);
+      	}
```
tchaikov: could you revert this formatting change?

src/common/config.cc
```diff
@@ -1227,6 +1227,46 @@ void md_config_t::diff(
...
+            map<string, pair<string, string>> *diff, 
+            const string& ceph_setting, bool show_unchanged)
+{
+  Mutex::Locker l(lock);
+
+  char local_buf[4096];
+  char other_buf[4096];
+  for (int i = 0; i < NUM_CONFIG_OPTIONS; i++) {
+    config_option *opt = &config_optionsp[i];
+    std::string ceph_setting_name(opt->name);
```
tchaikov: this variable is never used.

src/common/config.cc
```diff
@@ -1227,6 +1227,46 @@ void md_config_t::diff(
   }
 }
 
+void md_config_t::diff_setting(const md_config_t *other, 
+            map<string, pair<string, string>> *diff, 
+            const string& ceph_setting, bool show_unchanged)
```
tchaikov: we can add `show_unchanged` later once we want to pass `false`.

src/common/config.cc
```diff
@@ -1227,6 +1227,46 @@ void md_config_t::diff(
   }
 }
 
+void md_config_t::diff_setting(const md_config_t *other, 
+            map<string, pair<string, string>> *diff,
```
tchaikov: as we discussed, in your case, at most only one item will be returned. so either we can consolidate these two diff() methods, or we don't need to use a `map` here.

src/vstart.sh
```diff
@@ -431,7 +431,8 @@ test -d gmon && $SUDO rm -rf gmon/*
 
 
 # figure machine's ip
-HOSTNAME=`hostname -s`
+#HOSTNAME=`hostname -s`
+HOSTNAME=`hostname | awk -F "." '{print $1}'`
```
tchaikov: why `hostname -s` does not work for you?

src/common/config.cc
```diff
@@ -1236,21 +1236,45 @@ bool md_config_t::expand_meta(std::string &origval,
...
+
+void md_config_t::diff_helper(
     const md_config_t *other,
     map<string,pair<string,string> > *diff,
-    set<string> *unknown)
+    set<string> *unknown, const string& ceph_setting=string(""))
 {
   Mutex::Locker l(lock);
 
+  bool ceph_param_search = (!ceph_setting.empty()) ? true : false;
```
tchaikov:
```c++
bool ceph_param_search = !ceph_setting.empty();
```
would suffice. and IMHO, no need to introduce `ceph_param_search` just for checking if `ceph_setting` is empty or has the default value.

src/common/config.cc
```diff
@@ -1236,21 +1236,45 @@ bool md_config_t::expand_meta(std::string &origval,
...
+        continue;
+      }
+    }
     memset(local_buf, 0, sizeof(local_buf));
     memset(other_buf, 0, sizeof(other_buf));
 
     char *other_val = other_buf;
     int err = other->get_val(opt.name, &other_val, sizeof(other_buf));
     if (err < 0) {
+      if (ceph_param_search) {
```
tchaikov: why would you want to `continue` here? as `ceph_setting == opt.name` if `ceph_param_search`.

src/common/config.cc
```diff
@@ -1264,6 +1288,12 @@ void md_config_t::diff(
 
     if (strcmp(local_val, other_val))
       diff->insert(make_pair(opt.name, make_pair(local_val, other_val)));
+    else {
```
tchaikov: could just put 
```c++
else if {
  ...
}
```
less indent.

src/common/config.cc
```diff
@@ -1236,21 +1236,45 @@ bool md_config_t::expand_meta(std::string &origval,
 }
 
 void md_config_t::diff(
+  const md_config_t *other,
+  map<string, pair<string, string> > *diff,
+  set<string> *unknown)
```
tchaikov: why do we need this method as we have `diff_helper()` whose last param is optional?

src/common/ceph_context.cc
tchaikov: @oliveiradan ping?

src/common/config.cc
oliveiradan: @tchaikov , we the same function, with 2 different signatures, where once accepts the 'setting' we are searching for. Would you rather have only 1 function (as it was before) and the optional param for the 'setting' search, which would force us to call it passing a param with ""?

src/common/config.cc
oliveiradan: Absolutely correct. Thanks... I think i was spending too much time with Python these days. ;-)

src/common/config.cc
oliveiradan: If we didn't find the param, why spend time going through the rest of the code, as opposed to spend time looking at the next one?

src/vstart.sh
oliveiradan: Sorry, this shouldn't be there... it was from my lab at home. I will take it out.

liewegas:
ping

liewegas:
@oliveiradan ping

badone:
@oliveiradan ping, would you like some assistance here to get this completed?

oliveiradan:
I've talked to @badone and already rebased it. It is in the process of rebuilding it so I can run the tests again tomorrow morning.

src/common/config.cc
tchaikov: if we didn't find the param, maybe we should instead `break`?

src/common/config.cc
tchaikov: > which would force us to call it passing a param with ""

you added an optional param of `ceph_setting`, so i guess we are already ready?

src/common/config.cc
```diff
@@ -1236,21 +1236,45 @@ bool md_config_t::expand_meta(std::string &origval,
...
+  const md_config_t *other,
+  map<string, pair<string, string> > *diff,
+  set<string> *unknown) 
+{
+  diff_helper(other, diff, unknown, "");
+}
+void md_config_t::diff(
+  const md_config_t *other,
+  map<string, pair<string, string> > *diff,
+  set<string> *unknown, const string& ceph_setting = string(""))
```
tchaikov: s/ceph_setting/setting/ as in this context, it's always "ceph" setting.

tchaikov:
@oliveiradan very sorry for the latency. i just lost track of this PR.

src/common/ceph_context.cc
```diff
@@ -496,7 +496,39 @@ void CephContext::do_command(std::string command, cmdmap_t& cmdmap,
         f->dump_string("option", *p);
       }
       f->close_section(); // unknown
-    } else if (command == "log flush") {
+    }
+    else if (command == "config diff get") {
```
tchaikov: please move the `else {` to the previous line, so it looks like
```c
} else if {
...
```

src/common/ceph_context.cc
```diff
@@ -496,7 +496,39 @@ void CephContext::do_command(std::string command, cmdmap_t& cmdmap,
...
+      std::string ceph_setting;
+      if (!cmd_getval(this, cmdmap, "var", ceph_setting)) {
+        f->dump_string("error", "syntax error: 'config diff get <var>'");
+      } 
+      else {
```
tchaikov: ditto

src/common/ceph_context.cc
```diff
@@ -496,7 +496,39 @@ void CephContext::do_command(std::string command, cmdmap_t& cmdmap,
...
+
+        f->open_object_section("defaults");
+        for (const auto& p : diff) {
+          f->dump_string(p.first.c_str(), p.second.first);
+        } 
+        f->close_section();   //-- defaults
+        f->close_section();   //-- diff
+      } //-- else if (!cmd_getval(this, cmdmap, "var", ceph_setting))
+    } //-- else if (command == "config diff get")
+    else if (command == "log flush") {
```
tchaikov: ditto.

src/common/config.cc
```diff
@@ -1342,21 +1342,44 @@ bool md_config_t::expand_meta(std::string &origval,
...
+        continue;
+      }
+    }
     memset(local_buf, 0, sizeof(local_buf));
     memset(other_buf, 0, sizeof(other_buf));
 
     char *other_val = other_buf;
     int err = other->get_val(opt.name, &other_val, sizeof(other_buf));
     if (err < 0) {
+      if (!setting.empty()) {
```
tchaikov: in this case, `setting == opt.name`, and we should not error out with an error indicating that `config_options` has `setting` but `other` does not. `continue` does not help here. it will end up with `ENOENT`.

oliveiradan: Ok, thanks for the heads up. The checking (unneeded) in question was removed.

yuriw:
http://pulpito.ceph.com/yuriw-2017-05-17_00:19:04-rados-wip-yuri-testing2_2017_5_17-distro-basic-smithi/

http://pulpito.ceph.com/yuriw-2017-05-17_15:13:43-rados-wip-yuri-testing2_2017_5_17---basic-smithi/

http://pulpito.ceph.com/yuriw-2017-05-17_17:25:28-rados-wip-yuri-testing2_2017_5_17---basic-smithi/

==============

http://pulpito.ceph.com/yuriw-2017-05-17_17:32:14-rbd-wip-yuri-testing2_2017_5_17-distro-basic-smithi/

http://pulpito.ceph.com/yuriw-2017-05-17_23:59:41-rbd-wip-yuri-testing2_2017_5_17---basic-smithi/

PR #10737, opened 2016-08-16
rgw: fix radosgw daemon core when reopen logs

weiqiaomiao:
the ``rgw_user_st_syn`` thread use the pointer ``meta_mgr`` in ``RGWUserStatsCache::sync_all_users()`` fuction, so we should close the thread before delete the pointer in ``RGWRados::finalize()`` function when close storage before reopen_logs.

Fixes: http://tracker.ceph.com/issues/17036

### DIFF
--- a/src/rgw/rgw_rados.cc
+++ b/src/rgw/rgw_rados.cc
@@ -3164,7 +3164,6 @@ void RGWRados::finalize()
     data_notifier->stop();
     delete data_notifier;
   }
-  delete meta_mgr;
   delete data_log;
   if (async_rados) {
     delete async_rados;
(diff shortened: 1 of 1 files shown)


src/rgw/rgw_rados.cc
```diff
@@ -3201,6 +3200,7 @@ void RGWRados::finalize()
   if (cr_registry) {
     cr_registry->put();
   }
+  delete meta_mgr;
```
tchaikov: does jewel suffer from this issue as well?

weiqiaomiao: although jewel don't enter this process when reopen_logs, but it may hit this problem when the RGWRealmReloader::reload() fuction is called(for example,  change the master zone to a different zone).

PR #10740, opened 2016-08-16
common/WorkQueue: " _lock.Lock()" location tunnel in "ThreadPool::worker"

jimifm:
I think that _lock.Lock() in front of the _void_process_finish will affect the timing of the other threads for _lock in a multithreaded environment , especially when  _void_process_finish is a time-consuming operation. (eg. in FileStore.h OpWQ::_process_finish(OpSequencer *osr) )
Signed-off-by: jimifm 316929299@qq.com


yuyuyu101:
actually workqueue require lock when process_finish in other cases.

PR #10742, opened 2016-08-16
test/rgw: add --gateways-per-zone to test_multi.py

cbodley:
for testing with some contention for the sync locks

### DIFF
--- a/src/test/rgw/test-rgw-common.sh
+++ b/src/test/rgw/test-rgw-common.sh
@@ -85,7 +85,7 @@ function init_first_zone {
   realm=$2
   zg=$3
   zone=$4
-  port=$5
+  endpoints=$5
 
   access_key=$6
   secret=$7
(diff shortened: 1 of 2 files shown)


cbodley:
running this with --gateways-per-zone=2, i'm consistently hitting an assertion in RGWMetaSyncCR. see https://github.com/ceph/ceph/pull/10743

cbodley:
this testing has uncovered another related issue: http://tracker.ceph.com/issues/17047

oritwas:
lgtm

src/test/rgw/test-rgw-common.sh
```diff
@@ -94,13 +94,11 @@ function init_first_zone {
   x $(rgw_admin $id) realm create --rgw-realm=$realm
 
 # create zonegroup, zone
-  x $(rgw_admin $id) zonegroup create --rgw-zonegroup=$zg --endpoints=http://localhost:$port --master --default
```
mikulely: hi~ @cbodley It seems that we no longer set endpoints for zg,but instead just set endpoints on the master zone of zg? If so, maybe we need to update the doc https://github.com/ceph/ceph/blob/master/doc/radosgw/multisite.rst#creating-a-master-zonegroup

PR #10745, opened 2016-08-16
rgw nfs v3 completions

mattbenjamin:


src/include/rados/rgw_file.h
```diff
@@ -220,9 +221,10 @@ int rgw_truncate(struct rgw_fs *rgw_fs,
 */
 #define RGW_OPEN_FLAG_NONE         0x0000
 #define RGW_OPEN_FLAG_CREATE       0x0001
+#define RGW_OPEN_FLAG_V3           0x0002 /* ops have v3 semantics */
```
dang: "V3" is not hugely meaningful to anyone not doing NFS work.  Maybe "STATELESS"?

mattbenjamin: I used that elsewhere, so could do that.

src/rgw/rgw_file.cc
```diff
@@ -844,11 +862,15 @@ namespace rgw {
...
   {
-    lock_guard guard(mtx);
-
+    unique_lock guard{mtx, std::defer_lock};
     int rc = 0;
+
+    if (! (flags & FLAG_LOCKED)) {
+      guard.lock();
+    }
+
```
adamemerson: This is bad style. A unique_lock is the lock and flag all in one, if the function expects to be called with the lock both locked and not locked, it should take a unique_lock by reference.

adamemerson:
Looks pretty good so far, apart from the one locking thing.

src/rgw/rgw_file.cc
mattbenjamin: I don't really want a unique_lock propagating out to rgw_* callers

mattbenjamin:
Removed an empty branch and bogus WS from the main commit.

mattbenjamin:
It appears this doesn't need a rebase, and is isolated.

src/include/rados/rgw_file.h
```diff
@@ -143,8 +143,9 @@ int rgw_statfs(struct rgw_fs *rgw_fs,
 #define RGW_CREATE_FLAG_NONE     0x0000
 
 int rgw_create(struct rgw_fs *rgw_fs, struct rgw_file_handle *parent_fh,
-	      const char *name, struct stat *st, uint32_t mask,
-	      struct rgw_file_handle **fh, uint32_t flags);
+	       const char *name, struct stat *st, uint32_t mask,
+	       struct rgw_file_handle **fh, uint32_t posix_flags,
+	       uint32_t flags);
```
cbodley: do we need some kind of versioning for changes to this public api?

src/common/ceph_timer.h
```diff
@@ -231,6 +235,30 @@ namespace ceph {
...
+      bool adjust_event(uint64_t id, typename TC::duration duration) {
+	return adjust_event(id, TC::now() + duration);
+      }
+
+      // Adjust the timeout of a currently-scheduled event (absolute)
+      bool adjust_event(uint64_t id, typename TC::time_point when) {
+	std::lock_guard<std::mutex> l(lock);
+
+	event key(id);
+	typename event_set_type::iterator it = events.find(key);
```
cbodley: consider `auto it =`, and you wouldn't need the typedefs above

mattbenjamin: it doesn't seem like avoiding type decls in headers is the best use of auto?

src/include/rados/rgw_file.h
mattbenjamin: Do you think a #define would be sufficient?

cbodley: sure

src/common/ceph_timer.h
cbodley: k :+1:

src/rgw/rgw_file.h
```diff
@@ -510,7 +516,12 @@ namespace rgw {
     int readdir(rgw_readdir_cb rcb, void *cb_arg, uint64_t *offset, bool *eof,
 		uint32_t flags);
     int write(uint64_t off, size_t len, size_t *nbytes, void *buffer);
-    int write_finish();
+
+    int commit(uint64_t offset, uint64_t length, uint32_t flags) {
+      return 0;
+    }
```
cbodley: just a stub for later? we might want to return ENOTSUPP for now, instead of returning success while there could still be unstable writes

adamemerson: I wouldn't. If you give NFS3ERR_NOTSUPP in response to COMMIT you'll make the kernel freak out and probably record an item in non-volatile memory to remind itself not to talk to that server again.

mattbenjamin: yeah, atm, I'm just passing this result up through ganesha, which would have the result adam describes;  we -could- return ENOTSUP here if we strongly prefer it, but then the rgw fsal would need to translate it to this

mattbenjamin: ok

PR #10747, opened 2016-08-17
gcc 6.1.1 complains about missing include: <random>. 4.8.3 does not c…

oliveiradan:
…omplain about it

Signed-off-by: Daniel Oliveira <doliveira@suse.com>

### DIFF
--- a/src/test/msgr/test_async_networkstack.cc
+++ b/src/test/msgr/test_async_networkstack.cc
@@ -17,6 +17,7 @@
 #include <algorithm>
 #include <atomic>
 #include <iostream>
+#include <random>
 #include <string>
 #include <set>
 #include <vector>


src/test/msgr/test_async_networkstack.cc
```diff
@@ -20,6 +20,7 @@
 #include <string>
 #include <set>
 #include <vector>
+#include <random>
```
badone: The only nit-pick would be to put the header in alphabetical order, otherwise lgtm

yuyuyu101:
please squash the two commits. lgtm

tchaikov:
@oliveiradan could you squash these two commit into a single one? 


and could you prefix the title of your commit message with the subcomponent your are changing ? see https://github.com/ceph/ceph/blob/master/SubmittingPatches.rst#preparing-and-sending-patches, in this case, it would be "test: ".

oliveiradan:
@tchaikov, @yuyuyu101, @badone, it should be all set now.

badone:
lgtm

tchaikov:
lgtm.

PR #10751, opened 2016-08-17
rgw: do not try to encode or decode time_t and fix compiling warnings

tchaikov:
time_t is `signed int` on linux, which is ambiguous when encode() or decode() on i386-linux.
see also 8f9e7b0.

### DIFF
--- a/src/cls/rgw/cls_rgw.cc
+++ b/src/cls/rgw/cls_rgw.cc
@@ -474,7 +474,7 @@ int rgw_bucket_list(cls_method_context_t hctx, bufferlist *in, bufferlist *out)
       decode_list_index_key(kiter->first, &key, &ver);
 
       start_key = kiter->first;
-      CLS_LOG(20, "start_key=%s len=%lu", start_key.c_str(), start_key.size());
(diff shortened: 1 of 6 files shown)


tchaikov:
pushed to wip-kefu-testing, see http://gitbuilder.sepia.ceph.com/gitbuilder-ceph-deb-trusty-i386-basic/#origin/wip-kefu-testing

src/cls/rgw/cls_rgw.cc
```diff
@@ -474,7 +474,7 @@ int rgw_bucket_list(cls_method_context_t hctx, bufferlist *in, bufferlist *out)
       decode_list_index_key(kiter->first, &key, &ver);
 
       start_key = kiter->first;
-      CLS_LOG(20, "start_key=%s len=%lu", start_key.c_str(), start_key.size());
+      CLS_LOG(20, "start_key=%s len=%zu", start_key.c_str(), start_key.size());
```
dmick: Pleasantly surprised to find this, and to find that it's also standard enough to be part of Solaris.

PR #10752, opened 2016-08-17
osd/OSD: optimize get_nextmap_reserved() and release_map()

yuyuyu101:
For async messenger backend, it won't see any potential lock hold in ms_fast_dispatch. Otherwise will it trigger thundering herd, lots of connections will be affected. It doesn't mean little effect for simple backend. If lock hold happen in async backend, it means lots of pipe threads will race in fast dispatch path too, I think it also hurt for performance hugely.

So I introduce several commits to avoid potential lock in fast path. When osd facing 1w+ messages, it will help for single async worker throughput and latency. Lots of lock wait is avoided.

### DIFF
--- a/src/osd/OSD.h
+++ b/src/osd/OSD.h
@@ -529,54 +529,76 @@ public:
    * down, without worrying about reopening connections from threads
    * working from old maps.
    */
+  using OSDMapRaw = const OSDMap*;
 private:
-  OSDMapRef next_osdmap;
+  // Uses two element sized vector to store two OSDMap references. Since
+  // we must await old osdmap release, so up to 2 element size is enough.
+  std::vector<OSDMapRef> saved_osdmaps;
+  std::atomic<OSDMapRaw> next_osdmap;
   Cond pre_publish_cond;
 
 public:
   void pre_publish_map(OSDMapRef map) {
     Mutex::Locker l(pre_publish_lock);
-    next_osdmap = std::move(map);
+    if (next_osdmap)
+      saved_osdmaps[1] = map;
+    else
+      saved_osdmaps[0] = map;
+    next_osdmap = map.get();
   }
 
   void activate_map();
-  /// map epochs reserved below
-  map<epoch_t, unsigned> map_reservations;
+  // the advanced seq
+  std::atomic<uint64_t> map_reserve_seq;
+  // the completed seq number
+  std::atomic<uint64_t> map_completed_seq;
+  std::atomic<uint64_t> map_reserved_waiters;
+  uint64_t map_wait_version = 0;
+  // wait_version -> pair(seq, count)
+  std::map<uint64_t, std::pair<uint64_t, uint64_t> > map_wait_pending;
 
   /// gets ref to next_osdmap and registers the epoch as reserved
-  OSDMapRef get_nextmap_reserved() {
-    Mutex::Locker l(pre_publish_lock);
-    if (!next_osdmap)
-      return OSDMapRef();
-    epoch_t e = next_osdmap->get_epoch();
-    map<epoch_t, unsigned>::iterator i =
-      map_reservations.insert(make_pair(e, 0)).first;
-    i->second++;
+  OSDMapRaw get_nextmap_reserved(uint64_t &s) {
+    s = ++map_reserve_seq;
     return next_osdmap;
   }
   /// releases reservation on map
-  void release_map(OSDMapRef osdmap) {
-    Mutex::Locker l(pre_publish_lock);
-    map<epoch_t, unsigned>::iterator i =
-      map_reservations.find(osdmap->get_epoch());
-    assert(i != map_reservations.end());
-    assert(i->second > 0);
-    if (--(i->second) == 0) {
-      map_reservations.erase(i);
+  void release_map(uint64_t seq) {
+    map_completed_seq++;
+    if (map_reserved_waiters) {
+      int has_empty = 0;
+      Mutex::Locker l(pre_publish_lock);
+      for (std::map<uint64_t, std::pair<uint64_t, uint64_t> >::iterator it = map_wait_pending.begin();
+           it != map_wait_pending.end(); ++it) {
+        if (seq <= it->second.first)
+          has_empty += (--it->second.second == 0);
+      }
+      if (has_empty)
+        pre_publish_cond.Signal();
     }
-    pre_publish_cond.Signal();
   }
   /// blocks until there are no reserved maps prior to next_osdmap
   void await_reserved_maps() {
     Mutex::Locker l(pre_publish_lock);
     assert(next_osdmap);
-    while (true) {
-      map<epoch_t, unsigned>::const_iterator i = map_reservations.cbegin();
-      if (i == map_reservations.cend() || i->first >= next_osdmap->get_epoch()) {
-	break;
-      } else {
-	pre_publish_cond.Wait(pre_publish_lock);
-      }
+    map_reserved_waiters++;
+    // barrier seq
+    uint64_t cur_seq = map_reserve_seq;
+    uint64_t completed_seq = map_completed_seq;
+    if (cur_seq == completed_seq) {
+      map_reserved_waiters--;
+    } else {
+      assert(cur_seq > completed_seq);
+      uint64_t wait_version = ++map_wait_version;
+      map_wait_pending[wait_version] = make_pair(cur_seq, cur_seq - completed_seq);
+      while (map_wait_pending[wait_version].second)
+        pre_publish_cond.Wait(pre_publish_lock);
+      map_wait_pending.erase(wait_version);
+      map_reserved_waiters--;
+    }
+    if (saved_osdmaps[1]) {
+      std::swap(saved_osdmaps[0], saved_osdmaps[1]);
+      saved_osdmaps[1] = nullptr;
     }
   }
 
--- a/src/osd/OSD.h
+++ b/src/osd/OSD.h
@@ -1421,8 +1421,7 @@ public:
 
     Spinlock sent_epoch_lock;
     epoch_t last_sent_epoch;
-    Spinlock received_map_lock;
-    epoch_t received_map_epoch; // largest epoch seen in MOSDMap from here
+    std::atomic<epoch_t> received_map_epoch; // largest epoch seen in MOSDMap from here
 
     explicit Session(CephContext *cct) :
       RefCountedObject(cct),
--- a/src/osd/OSD.cc
+++ b/src/osd/OSD.cc
@@ -5957,9 +5957,7 @@ void OSD::ms_fast_preprocess(Message *m)
       MOSDMap *mm = static_cast<MOSDMap*>(m);
       Session *s = static_cast<Session*>(m->get_connection()->get_priv());
       if (s) {
-	s->received_map_lock.lock();
 	s->received_map_epoch = mm->get_last();
-	s->received_map_lock.unlock();
 	s->put();
       }
     }
@@ -6177,9 +6175,7 @@ bool OSD::dispatch_op_fast(OpRequestRef& op, OSDMapRef& osdmap)
     Session *s = static_cast<Session*>(op->get_req()->
 				       get_connection()->get_priv());
     if (s) {
-      s->received_map_lock.lock();
       epoch_t received_epoch = s->received_map_epoch;
-      s->received_map_lock.unlock();
       if (received_epoch < msg_epoch) {
 	osdmap_subscribe(msg_epoch, false);
       }
--- a/src/osd/OSD.cc
+++ b/src/osd/OSD.cc
@@ -228,6 +228,9 @@ OSDService::OSDService(OSD *osd) :
   publish_lock("OSDService::publish_lock"),
   pre_publish_lock("OSDService::pre_publish_lock"),
   max_oldest_map(0),
+  saved_osdmaps(2),
+  next_osdmap(nullptr),
+  map_reserve_seq(0), map_completed_seq(0), map_reserved_waiters(0),
   peer_map_epoch_lock("OSDService::peer_map_epoch_lock"),
   sched_scrub_lock("OSDService::sched_scrub_lock"), scrubs_pending(0),
   scrubs_active(0),
@@ -487,7 +490,7 @@ void OSDService::shutdown()
     backfill_request_timer.shutdown();
   }
   osdmap = OSDMapRef();
-  next_osdmap = OSDMapRef();
+  next_osdmap = nullptr;
 }
 
 void OSDService::init()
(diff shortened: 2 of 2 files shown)


gregsfortytwo:
I'm a little confused; have you observed actual performance issues which are reduced by these changes?

I'd be surprised if the recieved_map_epoch lock is really contended more than an atomic is — granted its two ops instead of one — but now that we have scalable atomics we should probably use them.

The second commit though doesn't remove any locks that I'm seeing, despite the commit message?
Didn't look at the third much.

yuyuyu101:
@gregsfortytwo oh, sorry. The previous second commit isn't correct. I move that from previous version...

The first commit doesn't mean we can improve outstanding performance, more than cleaning up codes which doesn't need lock/cond.

The main performance change is the last commit

src/osd/OSD.h
```diff
@@ -1421,8 +1421,7 @@ class OSD : public Dispatcher,
 
     Spinlock sent_epoch_lock;
     epoch_t last_sent_epoch;
-    Spinlock received_map_lock;
-    epoch_t received_map_epoch; // largest epoch seen in MOSDMap from here
+    std::atomic<epoch_t> received_map_epoch; // largest epoch seen in MOSDMap from here
```
gregsfortytwo: Hmm, should we update last_sent_epoch as well?

gregsfortytwo:
I'm not sure how I feel about this logic so I'll leave it to @athanatos.

But I bet he'll complain about assigning to a reference; you should pass a pointer instead. :)

athanatos:
The locking here is subtle and quick to anger.  I will need to think about this tomorrow.

src/osd/OSD.h
```diff
@@ -540,44 +540,54 @@ class OSDService {
...
   /// releases reservation on map
-  void release_map(OSDMapRef osdmap) {
-    Mutex::Locker l(pre_publish_lock);
-    map<epoch_t, unsigned>::iterator i =
-      map_reservations.find(osdmap->get_epoch());
-    assert(i != map_reservations.end());
-    assert(i->second > 0);
-    if (--(i->second) == 0) {
-      map_reservations.erase(i);
+  void release_map(uint64_t seq) {
```
athanatos: I see a lost wake up bug here.

src/osd/OSD.h
```diff
@@ -540,44 +540,54 @@ class OSDService {
...
+    if (map_reserved_waiters) {
```
athanatos: seq 1: Thread A observes map_reserved_waiters is 0

src/osd/OSD.h
```diff
@@ -540,44 +540,54 @@ class OSDService {
...
     assert(next_osdmap);
-    while (true) {
-      map<epoch_t, unsigned>::const_iterator i = map_reservations.cbegin();
-      if (i == map_reservations.cend() || i->first >= next_osdmap->get_epoch()) {
-	break;
-      } else {
-	pre_publish_cond.Wait(pre_publish_lock);
-      }
-    }
+    map_reserved_waiters++;
```
athanatos: seq 2: Thread B increments map_reserved_waiters here

src/osd/OSD.h
```diff
@@ -540,44 +540,54 @@ class OSDService {
...
+    uint64_t completed_seq = map_completed_seq;
+    if (cur_seq == completed_seq) {
+      map_reserved_waiters--;
+      return ;
+    }
+    assert(cur_seq > completed_seq);
+    uint64_t wait_version = ++map_wait_version;
+    map_wait_pending[wait_version] = make_pair(cur_seq, cur_seq - completed_seq);
+    while (map_wait_pending[wait_version].second)
+      pre_publish_cond.Wait(pre_publish_lock);
```
athanatos: seq 3: Thread B goes to sleep on map_reserve_seq != map_completed_seq.

src/osd/OSD.h
```diff
@@ -540,44 +540,54 @@ class OSDService {
...
+      for (std::map<uint64_t, std::pair<uint64_t, uint64_t> >::iterator it = map_wait_pending.begin();
+           it != map_wait_pending.end(); ++it) {
+        if (seq <= it->second.first)
+          has_empty += (--it->second.second == 0);
+      }
+      if (has_empty)
+        pre_publish_cond.Signal();
     }
-    pre_publish_cond.Signal();
+    map_completed_seq++;
```
athanatos: seq 4: Thread A increments map_completed_seq but does not wake up Thread B.

athanatos:
Btw, I agree that this is a good thing to attack!

athanatos:
At the time that I wrote this, I think I assumed the next version would use some kind of userspace rcu.   urcu-qsbr (https://lwn.net/Articles/573424/) actually looks promising.  We'd have to modify all messenger implementations to cooperate, and this reservation machinery would *only* work for fast_dispatch threads, but it would mean no read-side overhead beyond periodically marking each thread as quiescent.

src/osd/OSD.h
yuyuyu101: yep! I move "map_completed_seq" now

src/osd/OSD.h
yuyuyu101: hmm, last_sent_epoch requres a compare and assign op, so I'm not sure we could do it.

gregsfortytwo: There are definitely compare-and-swap atomics in the hardware; I *presume* C++ supports them... If not I will be a sad panda. :(

tchaikov: yeah, `atomic<>` does offer `compare_exchange_strong()`.

src/osd/OSD.h
```diff
@@ -540,44 +540,54 @@ class OSDService {
...
 
   void activate_map();
-  /// map epochs reserved below
-  map<epoch_t, unsigned> map_reservations;
+  // the advanced seq
+  std::atomic_uint map_reserve_seq;
+  // the completed seq number
+  std::atomic_uint map_completed_seq;
+  std::atomic_uint map_reserved_waiters;
+  std::atomic_uint map_wait_version;
```
athanatos: Only accessed under a lock, doesn't need to be atomic.

src/osd/OSD.h
```diff
@@ -540,44 +540,54 @@ class OSDService {
...
   /// gets ref to next_osdmap and registers the epoch as reserved
-  OSDMapRef get_nextmap_reserved() {
-    Mutex::Locker l(pre_publish_lock);
-    if (!next_osdmap)
-      return OSDMapRef();
-    epoch_t e = next_osdmap->get_epoch();
-    map<epoch_t, unsigned>::iterator i =
-      map_reservations.insert(make_pair(e, 0)).first;
-    i->second++;
+  OSDMapRef get_nextmap_reserved(uint64_t &s) {
```
athanatos: I think I see an acquire race here.

athanatos: seq 0: Thread A enters get_nextmap_reserved here.

src/osd/OSD.h
```diff
@@ -540,44 +540,54 @@ class OSDService {
...
-      } else {
-	pre_publish_cond.Wait(pre_publish_lock);
-      }
-    }
+    map_reserved_waiters++;
+    // barrier seq
+    uint64_t cur_seq = map_reserve_seq;
+    uint64_t completed_seq = map_completed_seq;
+    if (cur_seq == completed_seq) {
+      map_reserved_waiters--;
```
athanatos: Seq 0: Thread B gets to here

src/osd/OSD.h
```diff
@@ -540,44 +540,54 @@ class OSDService {
...
-    Mutex::Locker l(pre_publish_lock);
-    if (!next_osdmap)
-      return OSDMapRef();
-    epoch_t e = next_osdmap->get_epoch();
-    map<epoch_t, unsigned>::iterator i =
-      map_reservations.insert(make_pair(e, 0)).first;
-    i->second++;
+  OSDMapRef get_nextmap_reserved(uint64_t &s) {
+    s = ++map_reserve_seq;
     return next_osdmap;
```
athanatos: http://en.cppreference.com/w/cpp/memory/shared_ptr/atomic

I *think* you need to ensure that all accesses to the next_osdmap shared_ptr are done through the atomic accessors.  I don't think the barrier from map_reserve_seq is sufficient.

src/osd/OSD.h
yuyuyu101: I know cas op, what if this case:
      if (session->last_sent_epoch < last_sent_epoch) {
	session->last_sent_epoch = last_sent_epoch;
      }

gregsfortytwo: epoch_t slse = session->last_sent_epoch;
while (slse < last_sent_epoch && !session->last_sent_epoch.compare_exchange_strong(slse, last_sent_epoch)) {
  slse = session->last_sent_epoch;
}

gregsfortytwo: Erm, apparently I've got it wrong how the function call relates to which parameters, but you probably get the idea: grab the current value, CAS with that value, loop if it failed and is still too low.

yuyuyu101:
@athanatos std::atomic_load/std::atomic_store need gcc > 5.0. So we can't use it now....

athanatos:
Well, without a barrier, data races between assigning and dereferencing a shared_ptr can result in segfaults (or even worse, randomly scribbling on someone else's newly malloc'd memory), so this absolutely needs to be fixed one way or another.

athanatos:
At the expense of a pretty obnoxious level of complexity, you could maintain a set of up to two OSDMapRefs: the current one and (possibly) the one we are waiting for threads to finish with.  Probably simplest to make a bounded 2 element queue from a vector.  When you publish a new map, you push_back on the queue.  Then you atomically update a raw OSDMap* pointer to point at the new map.  Then, at the end of await_reserved_maps, you can remove the old list element.  You'll be returning an OSDMap* instead of an OSDMapRef since we are already guaranteeing that a reference will live until at least release() -- this also avoids the overhead of mucking about with the shared_ptr refcounts during the fast path.

athanatos:
I do think it would be worth investigating an rcu solution though.  You'd have to separate the mechanism used in the ms_fast_dispatch path from the one used in OSDService::get_con_osd_cluster since we don't want all of the other threads in the OSD to have to deal with the rcu restrictions.  A quick perusal of the fast_dispatch path shows a few other locks we'll have to be careful about:
- session_dispatch_lock: Not sure about this one
- received_map_lock: Not sure about this one
- MonClient locks in osdmap_subscribe: This one we can bypass by maintaining an atomic greatest_epoch_requested counter (or a vector of same to avoid cache-line bouncing) and doing the subscribe in Tick()
- PG::map_lock: This one I *think* we can bypass by just doing the map check when the op is dequeued.  I don't think ordering actually matters on the enqueue side (it does for enqueue_front though).  It would mean that ops might come out of the queue not in map order.  They'd still be ordered for a particular client.  Not sure if that matters.
- per-shard OpWQ lock: This one would involve updating the implementations to be lock-free.  Probably a good idea.
- pg_map_lock: This one is the important one.  We'd want to replace this whole mechanism on the fast-dispatch side with an rcu publish/subscribe mechanism and structure updates as publishing new maps.

If we do this right, we might be able to avoid any barriers in the common case except the publish barrier for the op queue.

yuyuyu101:
@athanatos hmm, I looked at https://lwn.net/Articles/573424/ again. It looks QSBR (quiescent-state-based RCU) make sense. do we want to make urcu(https://github.com/urcu/userspace-rcu) as submodule? I'm not sure this package whether wildly used in rhel/centos yum repo.

athanatos:
Yeah, I think we can get away with QSBR.  Before we actually start integrating it though, we need to decide what we are going to do with the above locks.  It would probably make sense to go ahead and merge a version of this PR first, and work out how we to do going forward.

yuyuyu101:
retest this please

yuyuyu101:
retest this please

yuyuyu101:
retest this please

yuyuyu101:
@athanatos I follow your suggestion to build a vector to fit two OSDMap, but it doesn't work as expected. Also I don't know why it doesn't work.

The fact is that when consuming OSDMap, some threads may result in segment fault. I also tried to push OSDMap when prepublish_osdmap without pop, but other threads still will stuck into reference to undefined OSDMap instance. 

So I may get your help in this... @tchaikov if free, kefu you man also give a look!

src/osd/OSD.h
```diff
@@ -529,54 +529,76 @@ class OSDService {
...
+  std::vector<OSDMapRef> saved_osdmaps;
+  std::atomic<OSDMapRaw> next_osdmap;
   Cond pre_publish_cond;
 
 public:
   void pre_publish_map(OSDMapRef map) {
     Mutex::Locker l(pre_publish_lock);
-    next_osdmap = std::move(map);
+    if (next_osdmap)
+      saved_osdmaps[1] = map;
```
athanatos: next_osdmap will essentially always be non-null (except just after boot).  I don't really get what's happening here.

tchaikov:
will take a look at this PR tmr.

yuyuyu101:
@tchaikov ping

tchaikov:
@yuyuyu101 i am still looking at this PR.

dachary:

dachary:

rchagam:
@liewegas  rebased with latest master, crashes with rados get operation. raods put operation works fine. 
```
ceph version 11.1.0-6958-g9675e29 (9675e2924e92f355fc9ef06d78a6675f6e75ffab)
 1: (()+0x7fb022) [0x5630788fd022]
 2: (()+0x11390) [0x7f78c11e6390]
 3: (OSD::handle_op(std::shared_ptr<OpRequest>&, std::shared_ptr<OSDMap const>&)+0x2dc) [0x5630783d67dc]
 4: (OSD::dispatch_op_fast(std::shared_ptr<OpRequest>&, std::shared_ptr<OSDMap const>&)+0x126) [0x5630783d7696]
 5: (OSD::dispatch_session_waiting(OSD::Session*, std::shared_ptr<OSDMap const>)+0x6f) [0x5630783d793f]
 6: (OSD::ms_fast_dispatch(Message*)+0x1fa) [0x5630783d7d6a]
 7: (DispatchQueue::fast_dispatch(Message*)+0x7e) [0x7f78c1a3086e]
 8: (AsyncConnection::process()+0x3098) [0x7f78c1ae7ba8]
 9: (EventCenter::process_events(int)+0x9b1) [0x7f78c1af7581]
 10: (()+0x3c3f89) [0x7f78c1afbf89]
 11: (()+0xb8c80) [0x7f78c0affc80]
 12: (()+0x76ba) [0x7f78c11dc6ba]
 13: (clone()+0x6d) [0x7f78c026582d]

```

liewegas:
My guess is that hte problem is related to handle_osd_map doing pre_publish_map() on many maps in a loop, and only doing the await_reserved_maps() in some cases (rarely!).

With the new faster dispatch code, we only need this during fast dispatch for legacy clients.  But we still use this machinery for all messages the OSD sends, so I think it's worth addressing this!

urcu QSBR might be applied here, but the requirement that all threads [using these functions] register and call the quiesce function periodically feels fragile.  My guess is that what you have here (which amounts to an atomic inc+dec per use) should be good enough.

PR #10753, opened 2016-08-17
osd: bail out if transaction size overflows

tchaikov:
with a large MOSDMap message, the transaction size could be greater than
UINT_MAX. so fail early with error messages.

Fixes: http://tracker.ceph.com/issues/16982

### DIFF
--- a/src/osd/OSD.cc
+++ b/src/osd/OSD.cc
@@ -6704,10 +6704,16 @@ void OSD::handle_osd_map(MOSDMap *m)
   }
 
   ObjectStore::Transaction t;
+  uint64_t txn_size = 0;
 
   // store new maps: queue for disk and put in the osdmap cache
   epoch_t start = MAX(superblock.newest_map + 1, first);
   for (epoch_t e = start; e <= last; e++) {
+    if (txn_size >= t.get_num_bytes()) {
+      derr << __func__ << " transaction size overflowed" << dendl;
+      assert(txn_size < t.get_num_bytes());
+    }
+    txn_size = t.get_num_bytes();
     map<epoch_t,bufferlist>::iterator p;
     p = m->maps.find(e);
     if (p != m->maps.end()) {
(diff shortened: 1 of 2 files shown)


athanatos:
Let's talk about this in standup tomorrow.

tchaikov:
will check if we have a guard for file journal because the transaction size could be greater than the size of file journal!

tchaikov:
changelog

- assert() if transaction size is greater than the journal size.

src/osd/OSD.cc
```diff
@@ -6704,10 +6704,16 @@ void OSD::handle_osd_map(MOSDMap *m)
...
 
   ObjectStore::Transaction t;
+  uint64_t txn_size = 0;
 
   // store new maps: queue for disk and put in the osdmap cache
   epoch_t start = MAX(superblock.newest_map + 1, first);
   for (epoch_t e = start; e <= last; e++) {
+    if (txn_size > t.get_num_bytes()) {
+      derr << __func__ << " transaction size overflowed" << dendl;
+      assert(txn_size > t.get_num_bytes());
```
xiexingguo: @tchaikov  I think this assert has no effect(always true)?

tchaikov: right. will fix.

tchaikov:
changelog

- should assert() if `assert(txn_size < t.get_num_bytes())`
- compare with `header.max_size` instead of `max_size`, the latter is the size of device not the size of journal.

athanatos:
lgtm, adding needs-qa

tchaikov:
tested at http://pulpito.ceph.com/kchai-2016-08-20_18:41:17-rados-wip-kefu-testing---basic-mira/

failed ones are tracked by 

* http://tracker.ceph.com/issues/17087
* http://tracker.ceph.com/issues/16157
* http://tracker.ceph.com/issues/16800

PR #10755, opened 2016-08-17
cmake: Add -pie to CMAKE_EXE_LINKER_FLAGS

tserong:
Without this, rpmlint (on openSUSE Tumbleweed) fails with:

ceph-radosgw.x86_64: E: non-position-independent-executable
(Badness: 10000) /usr/bin/radosgw
This executable must be position independent.  Check that it
is built with -fPIE/-fpie in compiler flags and -pie in linker
flags.


tchaikov:
@wjwithagen clang also supports `-pie` right?

PR #10758, opened 2016-08-17
jewel: ceph-disk should timeout when a lock cannot be acquired

dachary:
http://tracker.ceph.com/issues/17007


dachary:
Passed the upgrade suite http://tracker.ceph.com/issues/17487#note-15

PR #10761, opened 2016-08-17
jewel: msgr/async: Messenger thread long time lock hold risk

dachary:
http://tracker.ceph.com/issues/16377


dachary:
@liewegas does this backport look good to merge ? It passed the rados suite http://tracker.ceph.com/issues/17487#note-4 (except for one known bug that is, I believe, unrelated and one environmental failure on an erasure code job).

PR #10762, opened 2016-08-17
librbd: cache was not switching to writeback after first flush

dillaman:
Fixes: http://tracker.ceph.com/issues/16654


trociny:
lgtm

PR #10766, opened 2016-08-17
os/bluestore/BlueFS: disable buffered io

liewegas:


liewegas:
@somnathr

somnathr:
@liewegas Yes, all of my performance run was with bluestore_bluefs_buffered_io = false and this is giving much better and stable performance out.

src/os/bluestore/BlueStore.cc
```diff
@@ -4888,8 +4888,11 @@ void BlueStore::_txc_write_nodes(TransContext *txc, KeyValueDB::Transaction t)
        ++p) {
     bufferlist bl;
     ::encode((*p)->onode, bl);
+    unsigned first_part = bl.length();
     (*p)->blob_map.encode(bl);
-    dout(20) << "  onode " << (*p)->oid << " is " << bl.length() << dendl;
+    dout(20) << "  onode " << (*p)->oid << " is " << bl.length()
+	     << " (" << first_part << " onode + "
+	     << (bl.length() - first_part) << " bnode_map)" << dendl;
```
somnathr: Is it bnode_map or blob_map ?
Bnode is in-memory right ?

PR #10767, opened 2016-08-17
rgw: frontend subsystem rework

rzarzynski:
TODO:
- ~~Document the new interfaces.~~
- ~~Test frontends and RadosGW's APIs. The test matrix will be quite big.~~
- ~~Apply the new filters over `RGWRestfulIO` to all frontends.~~
- ~~Move the AWS Auth v4 signature generator to a dedicated filter.~~ (introduced `AWS_AUTHv4_IO`; the rest of the work will be accomplished in the continuation of the auth rework - I think that the AWSv4 auth engine would be able to decorate `req_state::cio` with a filter for signature calculation).
- ~~Seek and fix API clients which violate the call order of `RGWRestfulIO`. This is not the highest priority as we have `RGWRestfulIOReorderingEngine` now that could be used together with the new ASIO frontend.~~ (applied the reordering filter).


yuyuyu101:
OH, the boost asio......

src/rgw/rgw_rest.cc
```diff
@@ -369,63 +370,88 @@ void dump_errno(struct req_state *s, int http_ret)
...
   } else {
-    r = STREAM_IO(s)->print("ETag: \"%s\"\r\n", etag);
-  }
-  if (r < 0) {
-    ldout(s->cct, 0) << "ERROR: s->cio->print() returned err=" << r << dendl;
+    /* We need two extra bytes for quotes. */
+    char buf[etag.size() + 2 + 1];
+    const auto len = snprintf(buf, sizeof(buf), "\"%s\"", etag.data());
+
+    return dump_header(s, "ETag: \"%s\"\r\n", boost::string_ref(buf, len));
```
cbodley: `return dump_header(s, "ETag", boost::string_ref(buf, len));`

src/rgw/rgw_rest.cc
```diff
@@ -447,21 +473,17 @@ void dump_uri_from_state(struct req_state *s)
...
 	location += s->object.name;
-	STREAM_IO(s)->print("Location: %s\r\n", location.c_str());
+	dump_header(s, "Location", location);
       }
     }
-  }
-  else {
-    STREAM_IO(s)->print("Location: \"%s\"\r\n", s->info.request_uri.c_str());
+  } else {
+    dump_header(s, "Location", "\"" + s->info.request_uri + "\"");
```
cbodley: can we find a way to avoid `operator+()` on strings for stuff like this? maybe add a `dump_quoted_header()` function for cases like this and `lo_etag`?

src/rgw/rgw_client_io.h
```diff
@@ -4,83 +4,236 @@
...
-  virtual int complete_request() = 0; /* XXX signature likely changing */
+  virtual std::size_t complete_header() = 0;
 
-  virtual uint64_t get_bytes_sent() { return 0; }
-  virtual uint64_t get_bytes_received() { return 0; }
-}; /* RGWClient IO */
+  /* Receive body. On success Returns number of bytes sent to the direct
+   * client of RadosGW. On failure throws int containing errno. */
+  virtual std::size_t recv_body(char* buf, std::size_t max) = 0;
+  virtual std::size_t send_body(const char* buf, std::size_t len) = 0;
```
cbodley: can you explain the rationale behind splitting recv/send_body from read/write_data? is it just to support `RGWRestfulIOAccountingEngine`?

src/rgw/rgw_rest.cc
```diff
@@ -369,63 +370,88 @@ void dump_errno(struct req_state *s, int http_ret)
...
-      } else {
-        url_encode(s->bucket_name, b);
-      }
-      STREAM_IO(s)->print("Bucket: %s\r\n", b.c_str());
+  if (g_conf->rgw_expose_bucket && ! s->bucket_name.empty()) {
+    if (! s->bucket_tenant.empty()) {
+      dump_header(s, "Bucket",
+                  url_encode(s->bucket_tenant + "/" + s->bucket_name));
+    } else {
+      dump_header(s, "Bucket", url_encode(s->bucket_name));
```
cbodley: are you sure it's safe to pass a temporary `std::string` into `dump_header()` as a `const string_ref&` argument? it seems like the `string_ref` would end up with a pointer to freed memory

src/rgw/rgw_client_io.h
```diff
@@ -4,83 +4,236 @@
...
+  }
+};
+
+
+/* We're doing this nasty thing only because of extensive usage of templates
+ * to implement the static decorator pattern. C++ templates de facto enforce
+ * mixing interfaces with implementation. Additionally, those classes derive
+ * from RGWRestfulIO defined here. I believe that including in the middle of
+ * file is still better than polluting it directly. */
+#include "rgw_client_io_decoimpl.h"
```
cbodley: this in unfortunate, but i definitely think it's worth the tradeoff. and thanks for the explanation

src/rgw/rgw_client_io.h
```diff
@@ -4,83 +4,236 @@
...
 
-  bool account() { return _account; }
-  void set_account(bool _accnt) {
-    _account = _accnt;
+    int value() {
+      return err;
+    }
+  };
+
+  virtual std::size_t send_status(int status, const char *status_name) = 0;
```
rzarzynski: This is because in C++, unlike in C, [`size_t` lives in the `std` namespace](http://en.cppreference.com/w/cpp/types/size_t). Although `std::size_t` looks really ugly, I would prefer to avoid putting global `using` in headers.

src/rgw/rgw_client_io.h
```diff
@@ -4,83 +4,236 @@
...
+ * (e.g.) REST backends (may be replaced w/dynamic typed req_state). */
+static inline RGWRestfulIO* RESTFUL_IO(struct req_state* s) {
+  return static_cast<RGWRestfulIO*>(s->cio);
+}
 
-  uint64_t get_bytes_sent() { return bytes_sent; }
-  uint64_t get_bytes_received() { return bytes_received; }
-}; /* RGWStreamIO */
+static inline RGWClientIOAccounter* ACCOUNTING_IO(struct req_state* s) {
+  return dynamic_cast<RGWClientIOAccounter*>(s->cio);
```
rzarzynski: TODO.

src/rgw/rgw_client_io.h
```diff
@@ -4,83 +4,236 @@
...
+  virtual uint64_t get_bytes_sent() const = 0;
+  virtual uint64_t get_bytes_received() const = 0;
+};
+
+
+class RGWRestfulIOEngine : public RGWClientIO {
+  template<typename T> friend class RGWDecoratedRestfulIO;
+
+public:
+  class Exception : public std::exception {
```
rzarzynski: TODO.

src/rgw/rgw_client_io_decoimpl.h
```diff
@@ -0,0 +1,442 @@
...
+  size_t sent = 0;
+
+  if (! has_content_length) {
+    sent += RGWDecoratedRestfulIO<T>::send_content_length(data.length());
+    sent += RGWDecoratedRestfulIO<T>::complete_header();
+  }
+
+  if (buffer_data) {
+    if (data.length()) {
+      sent += RGWDecoratedRestfulIO<T>::send_body(data.c_str(),
```
rzarzynski: Definitely TODO.

Thanks a lot for the snippet!

src/rgw/rgw_client_io.h
cbodley: a global `using` isn't needed, though. `::size_t` is already available through system headers, and is widely used by ceph and rgw already

i don't think it's necessary to qualify integer types like `std::uint32_t`, and i don't see what makes `size_t` any different

src/rgw/rgw_client_io.h
rzarzynski: `recv_body` and `send_body` are parts of the public interface for doing **restful** IO while`write_data` and `read_data` live in the stream domain and are actually implementation' details. Although some engines are stream-oriented (both headers and body are handled the same way; FastCGI might be seen as an example here), others aren't. This applies especially to filters (accounting, chunking, buffering, reordering). Having to implement a machine state to differentiate headers and body would unnecessarily increase the complexity. Thus, I believe it's better to leverage the fact that the interface' usage pattern is truly restful.

src/rgw/rgw_rest.cc
rzarzynski: `url_encode(s->bucket_name)` should return a temporary `std::string` living till *the end of the full expression*. I think the same applies to temporary `boost::string_ref` created from it.

I would say the dangling pointer/reference might occur only if an engine buries `boost::string_ref` somewhere inside. That's the reason why `RGWRestfulIOReorderingEngine::send_header()` stores headers as a `std::pair` of `std::string`.

src/rgw/rgw_rest.cc
rzarzynski: I like this idea! TODO.

src/rgw/rgw_client_io.h
rzarzynski: Great. I don't like `std::size_t` too. :-) TODO.

src/rgw/rgw_client_io.h
rzarzynski: DONE.

src/rgw/rgw_client_io_decoimpl.h
rzarzynski: DONE.

src/rgw/rgw_civetweb.cc
```diff
@@ -3,143 +3,106 @@
...
+  if (ret < 0) {
+    throw RGWRestfulIOEngine::Exception(ret);
+  }
+  return ret;
 }
 
-void RGWMongoose::flush()
+void RGWCivetWeb::flush()
 {
+  return;
```
rzarzynski: Removed.

src/rgw/rgw_rest.h
```diff
@@ -529,10 +525,42 @@ extern void dump_start(struct req_state *s);
...
+                        const boost::string_ref& name,
+                        const utime_t& val);
+template <class... Args>
+static inline void dump_header_prefixed(struct req_state* s,
+                                        const boost::string_ref& name_prefix,
+                                        const boost::string_ref& name,
+                                        Args&&... args) {
+  char full_name_buf[name_prefix.size() + name.size() + 1];
+  const auto len = snprintf(full_name_buf, sizeof(full_name_buf), "%s%s",
+                            name_prefix.data(), name.data());
```
rzarzynski: DONE.

src/rgw/rgw_rest.cc
rzarzynski: DONE.

src/rgw/rgw_rest.cc
rzarzynski: DONE.

src/rgw/rgw_client_io_decoimpl.h
```diff
@@ -0,0 +1,442 @@
...
+#include <type_traits>
+#include <string>
+
+#include <boost/optional.hpp>
+
+#include "rgw_common.h"
+#include "rgw_client_io.h"
+
+template <typename T>
+class RGWRestfulIOAccountingEngine : public RGWDecoratedRestfulIO<T>,
```
rzarzynski: At the moment we have `rgw::io::AccountingEngine` (the filter) and `rgw::io::Accounter` (the interface). I'm still looking for better names.

rzarzynski: Hmm, maybe just `rgw::io::AccountingFilter`? What do you think, Casey?

src/rgw/rgw_client_io.h
cbodley: can we just use `size_t` instead of `::size_t` please?

src/rgw/rgw_client_io.h
```diff
@@ -4,83 +4,233 @@
...
+  virtual RGWEnv& get_env() noexcept = 0;
+  virtual int complete_request() = 0;
+}; /* rgw::io::Client */
+
+
+class Accounter {
+public:
+  virtual ~Accounter() {}
+
+  virtual void set_account(bool enabled) = 0;
```
cbodley: this part of the interface seems unnecessary to me.. i only see one caller, at the end of `end_header()` in rgw_rest.cc. can we just have `AccountingEngine::complete_header()` set `enabled = true;` instead?

src/rgw/rgw_client_io.h
rzarzynski: Oops, sorry - I've misunderstood you. I will rework this soon.

src/rgw/rgw_log.cc
```diff
@@ -211,8 +211,12 @@ static void log_usage(struct req_state *s, const string& op_name)
...
   rgw_usage_log_entry entry(u, p, bucket_name);
 
-  uint64_t bytes_sent = s->cio->get_bytes_sent();
-  uint64_t bytes_received = s->cio->get_bytes_received();
+  uint64_t bytes_sent = 0;
+  uint64_t bytes_received = 0;
+  if (ACCOUNTING_IO(s)) {
+    bytes_sent = ACCOUNTING_IO(s)->get_bytes_sent();
+    bytes_received = ACCOUNTING_IO(s)->get_bytes_received();
+  }
```
cbodley: consider storing the result of `ACCOUNTING_IO(s)` in a temporary so we only `dynamic_cast` once

rzarzynski: DONE.

rzarzynski:
@cbodley: to prevent issues with empty-valued HTTP headers (frequently used especially in the Swift API) we might need https://github.com/vinniefalco/Beast/commit/2fc27fb84e3d4e1a893b71a3f1a9d9f9020f56ed for Beast.

cbodley:
> to prevent issues with empty-valued HTTP headers (frequently used especially in the Swift API) we might need vinniefalco/Beast@2fc27fb for Beast.

okay, yeah. we might need to set up a ceph fork for that repo to use for our submodule. feel free to retarget the submodule to whatever commit you need in the meantime

src/rgw/Makefile.am
```diff
@@ -0,0 +1,289 @@
...
+	rgw/rgw_rest_s3website.h \
+        rgw/rgw_torrent.h \
+	civetweb/civetweb.h \
+	civetweb/include/civetweb.h \
+	civetweb/include/civetweb_conf.h \
+	civetweb/src/md5.h
+
+endif # WITH_RADOSGW
+endif # WITH_RADOS
+endif # ENABLE_CLIENT
```
cbodley: autotools is dead :skull: please don't resurrect it!

src/rgw/rgw_client_io_decoimpl.h
```diff
@@ -0,0 +1,438 @@
+// -*- mode:C++; tab-width:8; c-basic-offset:2; indent-tabs-mode:t -*-
+// vim: ts=8 sw=2 smarttab
+
+#ifndef CEPH_RGW_CLIENT_IO_DECOIMPL_H
+#define CEPH_RGW_CLIENT_IO_DECOIMPL_H
```
cbodley: since all of these classes are named Filters, maybe rename this header to `rgw_client_io_filters.h`?

src/rgw/rgw_client_io_decoimpl.h
```diff
@@ -0,0 +1,438 @@
...
+class AccountingFilter : public DecoratedRestfulClient<T>,
+                         public Accounter {
+  bool enabled;
+  uint64_t total_sent;
+  uint64_t total_received;
+
+public:
+  template <typename U>
+  AccountingFilter(U&& decoratee)
+    : DecoratedRestfulClient<T>(std::move(decoratee)),
```
cbodley: `U&& decoratee` here is a universal reference - it will bind to either rvalue- or lvalue refs - so the `std::move()` should be `std::forward()`. same goes with other filters' constructors and their `add_x()` factory functions

src/rgw/rgw_client_io_decoimpl.h
```diff
@@ -0,0 +1,438 @@
...
+  std::string uri_prefix;
+
+  void init_env(CephContext* const cct) override {
+    DecoratedRestfulClient<T>::init_env(cct);
+
+    if (! uri_prefix.empty()) {
+      auto& env = DecoratedRestfulClient<T>::get_env();
+      auto ruri = env.get("REQUEST_URI", cct->_conf->rgw_request_uri.c_str());
+      auto suri = env.get("SCRIPT_URI", cct->_conf->rgw_request_uri.c_str());
+      env.set("REQUEST_URI", uri_prefix + ruri);
```
cbodley: how does this interact with s3 signature verification? wouldn't it cause us to calculate a different signature than the client?

src/rgw/rgw_rest.cc
```diff
@@ -369,63 +370,99 @@ void dump_errno(struct req_state *s, int http_ret)
...
+}
+
+void dump_content_length(struct req_state* const s, const uint64_t len)
+{
+  try {
+    RESTFUL_IO(s)->send_content_length(len);
+  } catch (rgw::io::Exception& e) {
     ldout(s->cct, 0) << "ERROR: s->cio->send_content_length() returned err="
-                     << r << dendl;
+                     << e.code().value() << dendl;
```
cbodley: printing `e.what()` would be more descriptive

src/rgw/rgw_client_io_decoimpl.h
rzarzynski: Thanks for pointing this out, Casey. I'm looking for better way for prefixing than poking with `REQUEST_URI`. Maybe we can accomplish the goal by small modifications to `RGWRESTMgr` interface. I will get back to you soon.

src/rgw/rgw_client_io.h
```diff
@@ -4,83 +4,237 @@
...
-  virtual int send_content_length(uint64_t len) = 0;
+static inline rgw::io::Accounter* ACCOUNTING_IO(struct req_state* s) {
+  return dynamic_cast<rgw::io::Accounter*>(s->cio);
+}
 
-  uint64_t get_bytes_sent() { return bytes_sent; }
-  uint64_t get_bytes_received() { return bytes_received; }
-}; /* RGWStreamIO */
+static inline RGWRestfulIO* AWS_AUTHv4_IO(struct req_state* s) {
+  return static_cast<RGWRestfulIO*>(s->cio);
```
cbodley: i'm curious why the upcast from `rgw::io::BasicClient` to `RGWRestfulIO` is allowed here without `dynamic_cast`

and i suppose the same goes for `RESTFUL_IO` - how do we guarantee that the `BasicClient *cio` is actually a `RestfulClient` and not some other derived class?

src/rgw/rgw_rest.cc
rzarzynski: Done here and in other places as well.

src/rgw/rgw_client_io_decoimpl.h
rzarzynski: DONE.

src/rgw/Makefile.am
rzarzynski: DONE.

src/rgw/rgw_client_io_decoimpl.h
rzarzynski: DONE.

src/rgw/rgw_client_io.h
rzarzynski: The casts are now protected with `assert` according to our discussion.

cbodley:
I pushed this branch to ceph/wip-rgw-frontend-rework, but the centos gitbuilders didn't like it: http://gitbuilder.sepia.ceph.com/gitbuilder-ceph-rpm-centos7-amd64-basic/log.cgi?log=d277a67ec14cd7e33dc451c220f64f48221f30c0

rzarzynski:
@cbodley: fixed and rebased. I hope that after merging the Matt's pull request for in-source Boost we will be able to switch back to `boost::string_ref::to_string` just by reverting https://github.com/ceph/ceph/pull/10767/commits/e9c2b49b82c9eb7965fa3308996f0802c945f9d5.

rzarzynski:
Pushed to `ceph/wip-rgw-frontend-rework` for gitbuilders.

rzarzynski:
Oops, it looks we have an issue with Beast and Boost version on [the CentOS 7 gitbuilder](http://gitbuilder.sepia.ceph.com/gitbuilder-ceph-rpm-centos7-amd64-basic/log.cgi?log=0051815c807a1c666ff5650fea42ea00aa2a784b) :
```
 [100%] Building CXX object src/rgw/CMakeFiles/radosgw.dir/rgw_asio_frontend.cc.o
Linking CXX executable ../../bin/test_cors
Linking CXX executable ../../../bin/unittest_rgw_bencode
Linking CXX executable ../../../bin/unittest_http_manager
[100%] Built target ceph_test_rgw_obj
[100%] Building CXX object src/rgw/CMakeFiles/radosgw.dir/rgw_fcgi_process.cc.o
In file included from /srv/autobuild-ceph/gitbuilder.git/build/rpmbuild/BUILD/ceph-11.0.0/src/rgw/../Beast/include/beast/http/read.hpp:13:0,
from /srv/autobuild-ceph/gitbuilder.git/build/rpmbuild/BUILD/ceph-11.0.0/src/rgw/rgw_asio_frontend.cc:16:
/srv/autobuild-ceph/gitbuilder.git/build/rpmbuild/BUILD/ceph-11.0.0/src/rgw/../Beast/include/beast/core/async_completion.hpp:12:39: fatal error: boost/asio/async_result.hpp: No such file or directory
#include
^
compilation terminated. 
```

rzarzynski:
@cbodley: rabased due to a merge conflict and fixed an issue in the implementation Swift's TempAUTH responsible for many failures in the Teuthology run. No other changes.

Also pushed the branch for gitbuilders.

rzarzynski:
@cbodley: results of the last Teuthology run are [available](http://pulpito.ceph.com/rzarzynski-2016-10-19_17:03:30-rgw-wip-rgw-frontend-rework---basic-mira/). IMO there is no failure that looks like related to this PR.

cbodley:
@rzarzynski could you please update the submodule to point at https://github.com/ceph/Beast?

in the meantime, i'm getting cosbench set up to make sure there aren't any serious performance regressions

rzarzynski:
@cbodley: done - updated https://github.com/ceph/ceph/pull/10767/commits/273e5407a4040308399deaea75d117c495304e81 to accommodate `ceph/Beast`. Absolutely no other changes.

cbodley:
investigating some performance regressions with the civetweb frontend. i ran cosbench on a simple vstart cluster (one mon, one osd, one radosgw). against master, it got ~650 read ops/s and 160 writes/sec. with this pr, it was ~440 reads and 110 writes

src/rgw/rgw_client_io.h
```diff
@@ -239,6 +239,64 @@ class DecoratedRestfulClient : public RestfulClient {
...
+  /* Send exactly @len bytes from the memory location pointed by @buf.
+   * On success returns @len. On failure throws rgw::io::Exception. */
+  virtual size_t write_data(const char *buf, size_t len) = 0;
+};
+
+/* Utility class providing RestfulClient implementations with facilities
+ * for low-level buffering without relying on dynamic memory allocations.
+ * The buffer is carried entirely on stack. This narrows down applicability
+ * to those situations where buffers are relatively small. This perfectly
+ * fits the needs of composing an HTTP heade. Without that a front-end
```
rzarzynski: FIXME: typo.

src/rgw/rgw_client_io.h
```diff
@@ -239,6 +239,64 @@ class DecoratedRestfulClient : public RestfulClient {
...
+  char buffer[BufferSizeV];
+
+public:
+  StaticOutputBufferer(BuffererSink& sink)
+    : sink(sink) {
+    const size_t len = sizeof(buffer) - sizeof(std::streambuf::char_type);
+    std::streambuf::setp(buffer, buffer + len);
+  }
+};
+
```
cbodley: have you looked at using `common/PrebufferedStreambuf.h` for this?

cbodley:
@rzarzynski thanks, after that commit to buffer the headers again, the regression is gone :+1:

src/rgw/rgw_client_io.h
rzarzynski: I'm afraid that `PrebufferedStreambuf` won't help us. IIRC (Adam Kupczyk made an optimization there) it's rather for efficient log handling than buffering network IO. Basically its `overflow` method enlarges the `std::string` instead of flushing the buffered content.

src/rgw/rgw_client_io.h
```diff
@@ -239,6 +239,64 @@ class DecoratedRestfulClient : public RestfulClient {
...
+ * to those situations where buffers are relatively small. This perfectly
+ * fits the needs of composing an HTTP heade. Without that a front-end
+ * might need to issue a lot of small IO operations leading to increased
+ * overhead on syscalls and fragmentation of a message if the Nagle's
+ * algorithm won't be able to form a single TCP segment (usually when
+ * running on extremely fast network interfaces like the loopback). */
+template <size_t BufferSizeV = 4096>
+class StaticOutputBufferer : public std::streambuf {
+  static_assert(BufferSizeV >= sizeof(std::streambuf::char_type),
+                "The buffer must be bigger than one its char_type");
```
rzarzynski: FIXME.

rzarzynski:
@cbodley: thanks a lot for the immediate testing! I will apply similar modifications to other front-ends and also try to avoid the additional `strlen` in CivetWeb on a header's value.

src/rgw/rgw_client_io.h
rzarzynski: DONE.

rzarzynski:
@cbodley: the buffered header composition has been implemented in other front-ends as well. Using `strlen` in frontends' `send_header` methods isn't necessary anymore as the attributes are now sanitized to not include the trailing zero to string's content.

Both s3tests and Tempest haven't found a regression.

src/rgw/rgw_rest.cc
```diff
@@ -382,11 +382,22 @@ void dump_header(struct req_state* const s,
...
 
+static inline boost::string_ref get_sanitized_hdrval(ceph::buffer::list& raw)
+{
+  /* std::string and thus boost::string_ref ARE OBLIGED to carry multiple
+   * 0x00 and count them to the length of a string. We need to take that
+   * into consideration and sanitize the size of a ceph::buffer::list used
+   * to store metadata values (x-amz-meta-*, X-Container-Meta-*, etags).
+   * Otherwise we might send 0x00 to clients. */
+  const size_t len = raw.length();
+  return boost::string_ref(raw.c_str(), len ? len - 1 : 0);
```
cbodley: okay, i see that we're converting strings to bufferlist like this: `bl.append(etag.c_str(), etag.size() + 1);` i'm not sure why we need a null terminator in these bufferlists, because we shouldn't be trying to print them out

can you make this trimming logic conditional, so that we only trim if they're `\0`s? that way we have the option to start removing those trailing nulls

rzarzynski: > i'm not sure why we need a null terminator in these bufferlists, because we shouldn't be trying to print them out

I guess it was a workaround for `ceph::buffer::list::c_str()` that AFAIK doesn't guarantee its caller to obtain a *null-terminated* C-string.

src/rgw/rgw_rados.cc
```diff
@@ -7368,10 +7367,9 @@ int RGWRados::copy_obj(RGWObjectCtx& obj_ctx,
   }
 
   if (petag) {
-    map<string, bufferlist>::iterator iter = attrs.find(RGW_ATTR_ETAG);
-    if (iter != attrs.end()) {
-      bufferlist& etagbl = iter->second;
-      *petag = string(etagbl.c_str(), etagbl.length());
+    const auto iter = attrs.find(RGW_ATTR_ETAG);
+    if (iter != std::end(attrs)) {
+      *petag = iter->second;
```
rzarzynski: OK.

src/rgw/rgw_rest.cc
rzarzynski: DONE.

rzarzynski:
@cbodley, @mattbenjamin: the two last commits have been modified to address Casey's comments. No changes besides that.

Tempest and s3-tests haven't found any regression. Swift's TempAUTH implementation has been tested manually - looks good.

cbodley:
looks good :+1:
`Reviewed-by: Casey Bodley <cbodley@redhat.com>`

liewegas:
Forked to https://github.com/ceph/Beast

mattbenjamin:
@liewegas thanks!

src/rgw/rgw_rest_s3.cc
```diff
@@ -148,12 +148,21 @@ int RGWGetObj_ObjStore_S3::send_response_data(bufferlist& bl, off_t bl_ofs,
   map<string, string>::iterator riter;
   bufferlist metadata_bl;
 
-  if (op_ret)
-    goto done;
-
   if (sent_header)
     goto send_data;
 
+  if (custom_http_ret) {
```
joke-lee: hi @cbodley , when i test radosgw-admin usage , i get object 
```
from boto3.session import Session
import boto3
access_key = "yly"
secret_key = "yly"
url = 'http://127.0.0.1:7480'
session = Session(access_key, secret_key)
s3_client = session.client('s3', endpoint_url=url )

resp = s3_client.get_object(Bucket="test1", Key="512M")
with open('local-backup.bin','wb') as f:
    f.write(resp['Body'].read())
```
before it complete download ,i press ctrl+c to cancal ,and found that the success op =1 and op = 1,  i found the op_ret = -5 , in the current master branch ,it will go send_header and goto send_data,  which will not call set_req_state_err and resorting to 500, and  in the /var/log/ceph/rgw.log file , the message will be GET /bucket/object 200, this is not correct, so should we  remove this patch ?

joke-lee:
@mattbenjamin @cbodley  ping

src/rgw/rgw_rest_s3.cc
joke-lee:
```
2018-08-27 09:50:57.965 7f3efc217700  0 ERROR: flush_read_list(): d->client_cb->handle_data() returned -5
2018-08-27 09:50:57.965 7f3efc217700 20 get_obj_data::cancel_all_io()
2018-08-27 09:50:57.966 7f3efc217700  2 req 1 0.437s s3:get_obj completing
2018-08-27 09:50:57.966 7f3efc217700  2 req 1 0.437s s3:get_obj op status=-5
2018-08-27 09:50:57.966 7f3efc217700  2 req 1 0.437s http status=200
2018-08-27 09:50:57.966 7f3efc217700  1 ====== req done req=0x7f3efc210870 op status=-5 http_status=200 ======
2018-08-27 09:50:57.966 7f3efc217700  1 civetweb: 0x7f3f34f72000: 127.0.0.1 - - [27/Aug/2018:09:50:57 +0800] "GET /ylybucket/1024M HTTP/1.1" 200 321444138 - Boto3/1.8.1 Python/2.7.5 Linux/3.10.0-327.28.3.el7.x86_64 Botocore/1.11.1

```

joke-lee: jewel branch
```
2018-08-27 08:15:22.571433 7faa16f25700  1 ====== starting new request req=0x7faa16f1f710 =====
2018-08-27 08:15:22.890051 7faa16f25700  0 ERROR: flush_read_list(): d->client_c->handle_data() returned -5
2018-08-27 08:15:22.893102 7faa16f25700  0 WARNING: set_req_state_err err_no=5 resorting to 500
2018-08-27 08:15:22.893474 7faa16f25700  0 ERROR: s->cio->send_content_length() returned err=-5
2018-08-27 08:15:22.893482 7faa16f25700  0 ERROR: s->cio->print() returned err=-5
2018-08-27 08:15:22.893485 7faa16f25700  0 ERROR: STREAM_IO(s)->print() returned err=-5
2018-08-27 08:15:22.893502 7faa16f25700  0 ERROR: STREAM_IO(s)->complete_header() returned err=-5
2018-08-27 08:15:22.893670 7faa16f25700  1 ====== req done req=0x7faa16f1f710 op status=-5 http_status=500 ======
2018-08-27 08:15:22.893770 7faa16f25700  1 civetweb: 0x7fae5c000bb0: 127.0.0.1 - - [27/Aug/2018:08:15:22 +0800] "GET /ylybucket/100M HTTP/1.1" 500 0 - Boto3/1.7.6 Python/2.7.5 Linux/3.10.0-327.el7.x86_64 Botocore/1.10.6

```

master branch

```
2018-08-27 08:41:59.843 7f26d34f6700  1 ====== starting new request req=0x7f26d34ef870 =====
2018-08-27 08:41:59.895 7f26d34f6700  0 ERROR: flush_read_list(): d->client_cb->handle_data() returned -5
2018-08-27 08:41:59.895 7f26d34f6700  1 ====== req done req=0x7f26d34ef870 op status=-5 http_status=200 ======
2018-08-27 08:41:59.895 7f26d34f6700  1 civetweb: 0x7f270a75c000: 127.0.0.1 - - [27/Aug/2018:08:41:59 +0800] "GET /ylybucket/100M HTTP/1.1" 200 4866108 - Boto3/1.8.1 Python/2.7.5 Linux/3.10.0-327.28.3.el7.x86_64 Botocore/1.11.1


"user": "yly",
            "categories": [
                {
                    "category": "get_obj",
                    "bytes_sent": 4194304,
                    "bytes_received": 0,
                    "ops": 1,
                    "successful_ops": 1
                }
            ],
```

joke-lee: https://github.com/ceph/ceph/pull/10767/commits/a084b821839d95b1a12b8bd86429001b44255026 ,   if agree with it ,i will pull requests t fix this,

cbodley: > the message will be GET /bucket/object 200, this is not correct

hi @joke-lee - in this case, haven't we already responded to the client with 200 OK?

PR #10768, opened 2016-08-17
os/bluestore: avoid blob_t reencode when unchanged

liewegas:
This improves 4k random io iops by 9% on my nvme vstart cluster.

### DIFF
--- a/src/os/bluestore/BlueStore.cc
+++ b/src/os/bluestore/BlueStore.cc
@@ -1182,7 +1182,7 @@ void BlueStore::BlobMap::encode(bufferlist& bl) const
   ::encode(n, bl);
   for (auto p = blob_map.begin(); n--; ++p) {
     ::encode(p->id, bl);
-    ::encode(p->blob, bl);
+    p->encode(bl);
   }
 }
 
@@ -1195,7 +1195,7 @@ void BlueStore::BlobMap::decode(bufferlist::iterator& p, Cache *c)
     int64_t id;
     ::decode(id, p);
     Blob *b = new Blob(id, c);
-    ::decode(b->blob, p);
+    b->decode(p);
     b->get();
     blob_map.insert(*b);
   }
--- a/src/os/bluestore/BlueStore.cc
+++ b/src/os/bluestore/BlueStore.cc
@@ -2956,21 +2956,22 @@ int BlueStore::_fsck_verify_blob_map(
       ++errors;
       continue;
     }
-    if (b.blob.has_refmap() && pv->second != b.blob.ref_map) {
+    if (b.get_blob().has_refmap() && pv->second != b.get_blob().ref_map) {
       derr << " " << what << " blob " << b.id
-	   << " ref_map " << b.blob.ref_map
+	   << " ref_map " << b.get_blob().ref_map
 	   << " != expected " << pv->second << dendl;
       ++errors;
     }
-    bool compressed = b.blob.is_compressed();
+    bool compressed = b.get_blob().is_compressed();
     if (compressed) {
-      expected_statfs.compressed += b.blob.compressed_length;
-      for (auto& r : b.blob.has_refmap() ? b.blob.ref_map.ref_map : v[b.id].ref_map) {
+      expected_statfs.compressed += b.get_blob().compressed_length;
+      for (auto& r : b.get_blob().has_refmap() ? b.get_blob().ref_map.ref_map :
+	     v[b.id].ref_map) {
 	expected_statfs.compressed_original += r.second.refs * r.second.length;
       }
     }
     v.erase(pv);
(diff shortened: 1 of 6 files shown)


src/os/bluestore/BlueStore.h
```diff
@@ -288,9 +288,14 @@ class BlueStore : public ObjectStore,
   struct Blob : public boost::intrusive::set_base_hook<> {
     std::atomic_int nref;  ///< reference count
     int64_t id = 0;          ///< id
+  private:
     bluestore_blob_t blob;   ///< blob metadata
+  public:
     BufferSpace bc;          ///< buffer cache
 
+    mutable bool dirty = true;
```
ifed01: Can be private.

src/os/bluestore/BlueStore.h
```diff
@@ -324,6 +340,22 @@ class BlueStore : public ObjectStore,
...
+    void encode(bufferlist& bl) const {
+      if (dirty) {
+	::encode(blob, blob_bl);
+	dirty = false;
+      } else {
+	assert(blob_bl.length());
+      }
+      bl.append(blob_bl);
+    }
+    void decode(bufferlist::iterator& p) {
```
ifed01: I'm wondering if we can benefit from a deferred decoding. Just store encoded blob and increment a bufferlist iterator here. And do actual decode when get_blob or dirty_blob are called for the first time. The main question here is if there are decoded but unused blobs in the current implementation.

src/os/bluestore/BlueStore.cc
```diff
@@ -3760,7 +3761,8 @@ int BlueStore::_do_read(
     } else {
       for (auto reg : b2r_it->second) {
 	// determine how much of the blob to read
-	uint64_t chunk_size = bptr->blob.get_chunk_size(csum_type != bluestore_blob_t::CSUM_NONE, block_size);
+	uint64_t chunk_size = bptr->dirty_blob().get_chunk_size(
```
ifed01: can be probably replaced with get_blob if get_chunk_size marked as const

src/os/bluestore/BlueStore.cc
```diff
@@ -5931,7 +5936,10 @@ void BlueStore::_do_write_small(
 	     << " bstart 0x" << std::hex << bstart << std::dec << dendl;
 
     // can we pad our head/tail out with zeros?
-    uint64_t chunk_size = b->blob.get_chunk_size(b->blob.has_csum(), block_size); //blob csum settings to be applied hence ignoring current config settings for csum enable/disable
+    // blob csum settings to be applied hence ignoring current config
+    // settings for csum enable/disable
+    uint64_t chunk_size = b->dirty_blob().get_chunk_size(
```
ifed01: dirst_blob -> get_blob

src/os/bluestore/BlueStore.cc
```diff
@@ -5931,7 +5936,10 @@ void BlueStore::_do_write_small(
...
+      b->get_blob().has_csum(), block_size);
```
ifed01: b->get_blob().has_csum() can be replaced with true here as has_csum is called internally

src/os/bluestore/BlueStore.h
```diff
@@ -288,9 +288,14 @@ class BlueStore : public ObjectStore,
   struct Blob : public boost::intrusive::set_base_hook<> {
     std::atomic_int nref;  ///< reference count
     int64_t id = 0;          ///< id
+  private:
     bluestore_blob_t blob;   ///< blob metadata
+  public:
     BufferSpace bc;          ///< buffer cache
 
+    mutable bool dirty = true;
+    mutable bufferlist blob_bl;    ///< cached encoded blob
```
ifed01: and this one too

ifed01:
lgtm except mentioned minor issues

src/os/bluestore/BlueStore.h
liewegas: Good idea!  I suspect lots of blobs are unused and don't need to be decoded.

ifed01:
@liewegas any performance difference for deferred decode?

liewegas:
I didn't try to measure it.  Care to take a look?

src/os/bluestore/BlueStore.h
```diff
@@ -314,6 +320,36 @@ class BlueStore : public ObjectStore,
       return out << b.id << ":" << b.blob;
```
ifed01: b.get_blob() here?

ifed01:
Just some knitpicking - the last commit contains some fixes unrelated to on-demand blob decoding. Moreover switch to the previous commit 45b757e results in build failures. That's a bit inconvenient if one wants to verify performance gain for the on-demand decode.

PR #10769, opened 2016-08-17
osdc/Filer: move Probe class to .cc file

stiopaa1:


gregsfortytwo:
I think this one is actually clearer with the Probe struct in the header instead of hidden away in the implementation file. It's not quite as simple as our callback Contexts, but it's not so big it can't reside in the header either.

PR #10770, opened 2016-08-17
librbd: interlock image refresh and exclusive lock operations

dillaman:
It was previously possible for an image refresh and lock operation to race and result in sporadic assertion failures.

### DIFF
--- a/src/cls/lock/cls_lock.cc
+++ b/src/cls/lock/cls_lock.cc
@@ -45,6 +45,7 @@ cls_method_handle_t h_break_lock;
 cls_method_handle_t h_get_info;
 cls_method_handle_t h_list_locks;
 cls_method_handle_t h_assert_locked;
+cls_method_handle_t h_set_cookie;
 
 #define LOCK_PREFIX    "lock."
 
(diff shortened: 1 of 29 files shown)


src/cls/lock/cls_lock.cc
```diff
@@ -512,6 +513,94 @@ int assert_locked(cls_method_context_t hctx, bufferlist *in, bufferlist *out)
...
+  if (linfo.lock_type != op.type) {
+    CLS_LOG(20, "lock type mismatch: current=%s, assert=%s",
+            cls_lock_type_str(linfo.lock_type), cls_lock_type_str(op.type));
+    return -EBUSY;
+  }
+
+  if (linfo.tag != op.tag) {
+    CLS_LOG(20, "lock tag mismatch: current=%s, assert=%s", linfo.tag.c_str(),
+            op.tag.c_str());
+    return -EBUSY;
```
trociny: Why EBUSY and not EINVAL here and above?

src/librbd/ExclusiveLock.cc
```diff
@@ -202,13 +202,20 @@ void ExclusiveLock<I>::reacquire_lock(Context *on_reacquired) {
...
+      Action active_action = get_active_action();
+      assert(active_action == ACTION_TRY_LOCK ||
+             active_action == ACTION_REQUEST_LOCK);
+      execute_next_action();
+    } else if (!is_shutdown() &&
+               (m_state == STATE_LOCKED ||
+                m_state == STATE_ACQUIRING ||
+                m_state == STATE_POST_ACQUIRING ||
+                m_state == STATE_WAITING_FOR_PEER)) {
+      // ignore request if shutdown or not in a locked-related state
```
trociny: This comment in this form and in this place looks a little confusing to me. I would expect it applied to statements below, while it is vice versa.

trociny:
@dillaman I am observing this test crash
```
CEPH_LIB=lib RBD_FEATURES=61 ./bin/unittest_librbd --gtest_repeat=1000 --gtest_filter=TestMockExclusiveLock.ReacquireLockError
...
Note: Google Test filter = TestMockExclusiveLock.ReacquireLockError
[==========] Running 1 test from 1 test case.
[----------] 1 test from TestMockExclusiveLock
[ RUN      ] TestMockExclusiveLock.ReacquireLockError
zsh: segmentation fault (core dumped)  CEPH_LIB=lib RBD_FEATURES=61 ./bin/unittest_librbd --gtest_repeat=1000 
```

Also, one time I saw:
```
Note: Google Test filter = TestMockExclusiveLock.ReacquireLockError
[==========] Running 1 test from 1 test case.
[----------] 1 test from TestMockExclusiveLock
[ RUN      ] TestMockExclusiveLock.ReacquireLockError
unknown file: Failure

Unexpected mock function call - returning default value.
    Function call: get_watch_handle()
          Returns: 0
Google Mock tried the following 3 expectations, but none matched:

/home/mgolub/ceph/ceph.upstream/src/test/librbd/test_mock_ExclusiveLock.cc:105: tried expectation #0: EXPECT_CALL(*mock_image_ctx.image_watcher, get_watch_handle())...
         Expected: the expectation is active
           Actual: it is retired
         Expected: to be called any number of times
           Actual: called once - satisfied and retired
/home/mgolub/ceph/ceph.upstream/src/test/librbd/test_mock_ExclusiveLock.cc:105: tried expectation #1: EXPECT_CALL(*mock_image_ctx.image_watcher, get_watch_handle())...
         Expected: the expectation is active
           Actual: it is retired
         Expected: to be called any number of times
           Actual: called once - satisfied and retired
/home/mgolub/ceph/ceph.upstream/src/test/librbd/test_mock_ExclusiveLock.cc:105: tried expectation #2: EXPECT_CALL(*mock_image_ctx.image_watcher, get_watch_handle())...
         Expected: the expectation is active
           Actual: it is retired
         Expected: to be called any number of times
           Actual: never called - satisfied and retired
```

TestMockExclusiveLock.ReacquireLock does not crash for me.

If you fail to reproduce this I can provide more details.

src/cls/lock/cls_lock.cc
dillaman: Just following the existing convention where all mismatches return -EBUSY (e.g. the lock is owned by someone other than the client passed in via the input params).

src/librbd/ExclusiveLock.cc
dillaman: Good catch -- it was a leftover from my initial implementation.

dillaman:
@trociny I fixed the shut down race condition with that test case and re-pushed.

trociny:
lgtm

PR #10771, opened 2016-08-18
os/bluestore: misc fixes

xiexingguo:
We shall round nbytes up to be page-size(4096) aligned instead of
truncating it down, as syncing more is usually okay
while syncing less indicates potential data loss.

### DIFF
--- a/src/os/bluestore/BlueFS.cc
+++ b/src/os/bluestore/BlueFS.cc
@@ -473,7 +473,7 @@ int BlueFS::_replay(bool noop)
   dout(10) << __func__ << " log_fnode " << super.log_fnode << dendl;
 
   FileReader *log_reader = new FileReader(
-    log_file, g_conf->bluefs_alloc_size,
+    log_file, g_conf->bluefs_max_prefetch,
     false,  // !random
     true);  // ignore eof
   while (true) {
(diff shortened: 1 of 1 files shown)


liewegas:
I think this is a poorly named function.. the comment says

   // This asks the OS to initiate flushing the cached data to disk,
   // without waiting for completion.

so RangeStartFlush would have been better.  In any case, it's not a data integrity call, so I think it's okay as it.

What I want to avoid is flushing the same block twice...

xiexingguo:
@liewegas Dropped that change and appended more changes.

src/os/bluestore/BlueFS.cc
```diff
@@ -1379,8 +1379,9 @@ int BlueFS::_flush_range(FileWriter *h, uint64_t offset, uint64_t length)
 		      offset + length - allocated,
 		      &h->file->fnode.extents);
     if (r < 0) {
-      derr << __func__ << " allocated: " << allocated << \
-               " offset: " << offset << " length: " << length << dendl;
+      derr << __func__ << " allocated: 0x" << std::hex << allocated
+           << " offset: 0x" << offset << " length: 0x" << length << std::dec
```
liewegas: please include the std::dec at the end

liewegas:
otherwise lgtm, thanks!

src/os/bluestore/BlueFS.cc
xiexingguo: @liewegas Sir, I did... The `std::dec` lies at the end of this line.

liewegas: Oops, yes, sorry about that!

PR #10772, opened 2016-08-18
cmake: recompile erasure src for different variants

tchaikov:
* instead of reusing the object libraries, we should recompile jerasure
  code for different plugin flavors like neon, sse3, sse4.
* do not version plugin so, as they are not supposed to be used by
  user directly.


cbodley:
:+1: built with `make VERBOSE=1` and verified that jerasure sources were being built with correct cflags for each target

PR #10776, opened 2016-08-18
ceph-objectstore-tool: add a way to split filestore directories offline

jdurgin:
Use the usual split code, and split each dir that meets the
usual split criteria.

### DIFF
--- a/src/os/filestore/HashIndex.cc
+++ b/src/os/filestore/HashIndex.cc
@@ -19,6 +19,7 @@
 
 #include "HashIndex.h"
 
+#include "common/errno.h"
 #include "common/debug.h"
 #define dout_subsys ceph_subsys_filestore
 
@@ -301,6 +302,59 @@ int HashIndex::_split(
     bits,
     match,
     &mkdirred);
+}
+
+int HashIndex::split_dirs(const vector<string> &path) {
+  dout(20) << __func__ << " " << path << dendl;
+  subdir_info_s info;
+  int r = get_info(path, &info);
+  if (r < 0) {
+    dout(10) << "error looking up info for " << path << ": "
+	     << cpp_strerror(r) << dendl;
+    return r;
+  }
+
+  if (must_split(info)) {
+    r = initiate_split(path, info);
+    if (r < 0) {
+      dout(10) << "error initiating split on " << path << ": "
+	       << cpp_strerror(r) << dendl;
+      return r;
+    }
+
+    r = complete_split(path, info);
+    if (r < 0) {
+      dout(10) << "error completing split on " << path << ": "
+	       << cpp_strerror(r) << dendl;
+      return r;
+    }
+  }
+
+  vector<string> subdirs;
+  r = list_subdirs(path, &subdirs);
+  if (r < 0) {
+    dout(10) << "error listing subdirs of " << path << ": "
+	     << cpp_strerror(r) << dendl;
+    return r;
+  }
+  for (vector<string>::const_iterator it = subdirs.begin();
+       it != subdirs.end(); ++it) {
+    vector<string> subdir_path(path);
+    subdir_path.push_back(*it);
+    r = split_dirs(subdir_path);
+    if (r < 0) {
+      return r;
+    }
+  }
+
+  return r;
+}
+
+int HashIndex::apply_layout_settings() {
+  vector<string> path;
+  dout(10) << __func__ << " split multiple = " << split_multiplier
+	   << " merge threshold = " << merge_threshold << dendl;
+  return split_dirs(path);
 }
 
 int HashIndex::_init() {
--- a/src/tools/ceph_objectstore_tool.cc
+++ b/src/tools/ceph_objectstore_tool.cc
@@ -27,6 +27,7 @@
 
 #include "os/ObjectStore.h"
 #include "os/filestore/FileJournal.h"
+#include "os/filestore/FileStore.h"
 #ifdef HAVE_LIBFUSE
 #include "os/FuseStore.h"
 #endif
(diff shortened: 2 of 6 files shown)


jdurgin:
tested by https://github.com/ceph/ceph-qa-suite/pull/1124 - passed here: http://pulpito.ceph.com/joshd-2016-08-18_02:41:43-rados:thrash-wip-offline-split---basic-mira/

src/os/filestore/HashIndex.cc
```diff
@@ -303,6 +304,57 @@ int HashIndex::_split(
...
+      return r;
+    }
+  }
+
+  return r;
+}
+
+int HashIndex::apply_layout_settings() {
+  vector<string> path;
+  return split_dirs(path);
```
tchaikov: do we need to handle the case of `must_merge()` also?

src/tools/ceph_objectstore_tool.cc
```diff
@@ -2259,6 +2260,60 @@ int mydump_journal(Formatter *f, string journalpath, bool m_journal_dio)
...
+  }
+
+  vector<coll_t> collections, in_pool;
+  r = os->list_collections(collections);
+  if (r < 0) {
+    cerr << "Error listing collections: " << cpp_strerror(r) << std::endl;
+    return r;
+  }
+
+  for (vector<coll_t>::const_iterator it = collections.begin();
```
tchaikov: could use `auto` and range based loop.

src/tools/ceph_objectstore_tool.cc
```diff
@@ -2601,6 +2660,16 @@ int main(int argc, char **argv)
     goto out;
   }
 
+  if (object == "apply-layout-settings") {
```
tchaikov: nit, this is a little bit weird, could also be `--op apply-layout-settings <pool>`.

src/os/ObjectStore.h
```diff
@@ -2315,6 +2315,7 @@ class ObjectStore {
   * - num objects - total (including witeouts) object count to measure used space for.
   */
   virtual uint64_t estimate_objects_overhead(uint64_t num_objects) = 0;
+  virtual int apply_layout_settings(coll_t cid) { assert(0); return 0; }
```
tchaikov: IIUC, only FileStore needs to tackle this split-merge problem. can we `dynamic_cast<>` the `ObjectStore` in `ceph_objectstore_tool.cc`, if it is actually a `FileStore`, then we can `apply_layout_settings()` to it.

jdurgin: good idea, we can always make it generic later if other backends have something useful to do with it

src/os/filestore/HashIndex.cc
jdurgin: This command is really intended to just split directories, but I couldn't think of a better name for it. The idea is to be able to split offline, with lower than usual split settings, to avoid needing to split later online, since that is an expensive operation.

Merging is also expensive, but is much less prevalent since clusters are usually adding more objects and expanding rather than removing them, so an offline merge doesn't seem that useful currently.

jdurgin: I'd be happy to have a better name for apply_layout_settings()

src/tools/ceph_objectstore_tool.cc
jdurgin: suppose I might as well, since the backport to hammer needs to change the type anyway.

src/tools/ceph_objectstore_tool.cc
dzafman: Yes, this is not object related so normally I prefer it not to be part of the positional arguments.   I think you can check the 'object' positional argument to hold the pool name.  Though the non-positional usage doesn't have a place to specify the --op needs the pool argument.

ceph-objectstore-tool ..... --op apply-layout-settings <pool>   

Alternatively, you could add an arg --apply-layout-settings which takes the pool.  There the usage is part of the add_options() at the start of main().

ceph-objectstore-tool ..... --apply-layout-settings <pool>

src/tools/ceph_objectstore_tool.cc
```diff
@@ -2259,6 +2260,60 @@ int mydump_journal(Formatter *f, string journalpath, bool m_journal_dio)
...
+  if (r < 0) {
+    cerr << "Error listing collections: " << cpp_strerror(r) << std::endl;
+    return r;
+  }
+
+  for (vector<coll_t>::const_iterator it = collections.begin();
+       it != collections.end(); ++it) {
+    spg_t pgid;
+    if (it->is_pg(&pgid) && pgid.pool() == (uint64_t)poolid) {
+      in_pool.push_back(*it);
```
dzafman: Is there any use to only splitting a particular pgid?  Because if --pgid option is specified instead of pool it could be passed to the code below as coll_t coll(pgid).

jdurgin: yeah, I'll add that. easy to add and useful for testing or more fine-grained usage

src/tools/ceph_objectstore_tool.cc
jdurgin: I originally developed this against hammer which had another command as a positional arg already. Using --op apply-layout-settings sounds better, I'll switch to that. At that point would it make more sense to make the positional object arg optional, and just use --pgid or --pool for this command?

tchaikov:
lgtm. what do you think @dzafman ?

src/tools/ceph_objectstore_tool.cc
```diff
@@ -2228,6 +2229,7 @@ void usage(po::options_description &desc)
     cerr << "ceph-objectstore-tool ... <object> dump" << std::endl;
     cerr << "ceph-objectstore-tool ... <object> set-size" << std::endl;
     cerr << "ceph-objectstore-tool ... <object> remove-clone-metadata <cloneid>" << std::endl;
+    cerr << "ceph-objectstore-tool --op apply-layout-settings (--pool <pool-name>|--pgid <pgid>) " << std::endl;
```
dzafman: @jdurgin A user could infer this usage because apply-layout-settings is listed as an --op in the auto generated usage and you have the message "apply-layout-settings requires either --pool or --pgid."   This line comes out under positional syntax, so I would remove this line.

dzafman:
lgtm, except for 1 nit.

src/tools/ceph_objectstore_tool.cc
jdurgin: good point, it does seem unnecessary now

tchaikov:
@jdurgin better off using the notation of "Fixes: http://tracker.ceph.com/issues/17220" in the commit mesage.

tchaikov:
tested at http://pulpito.ceph.com/kchai-2016-09-10_15:46:37-rados-wip-kefu-testing---basic-mira/408471/

the failed test of http://pulpito.ceph.com/kchai-2016-09-10_15:46:37-rados-wip-kefu-testing---basic-mira/408471/ is irrelevant.

PR #10777, opened 2016-08-18
rgw ldap:  protect rgw::from_base64 from non-base64 input

mattbenjamin:
Also adds unit tests for:
1. empty output from from_base64 (turns out to be harmless)
2. random and specific non-base64 and sort strings

### DIFF
--- a/src/rgw/rgw_rest_s3.h
+++ b/src/rgw/rgw_rest_s3.h
@@ -710,7 +710,12 @@ public:
       store(store),
       apl_factory(apl_factory) {
       init(cct);
-      base64_token = rgw::from_base64(access_key_id);
+      /* boost filters and/or string_ref may throw on invalid input */
+      try {
+	base64_token = rgw::from_base64(access_key_id);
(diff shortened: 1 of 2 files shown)


src/rgw/rgw_rest_s3.h
```diff
@@ -710,7 +710,12 @@ class RGWLDAPAuthEngine: RGWS3V2AuthEngine
       store(store),
       apl_factory(apl_factory) {
       init(cct);
-      base64_token = rgw::from_base64(access_key_id);
+      /* boost filters and/or string_ref may throw on invalid input */
+      try {
+	base64_token = rgw::from_base64(access_key_id);
+      } catch(...) {
+	base64_token = std::string("");
```
cbodley: `base64_token` is already in its default-constructed state, is this necessary?

mattbenjamin: It feels cleaner to re-assert it in an empty state, at least to me.

PR #10778, opened 2016-08-18
rgw multisite: RGWCoroutinesManager::run returns status of last cr

cbodley:
Fixes: http://tracker.ceph.com/issues/17047


oritwas:
lgtm

yehudasa:
@cbodley not sure this is the best way to return status out of RGWCoroutinesManager::run(), works only if we run a single cr. Should probably hold a ref to the cr and take its status afterwards. For this specific issue it'll be ok though.

cbodley:
@yehudasa thanks. agreed that it's not perfect

PR #10779, opened 2016-08-18
rgw multisite: RGWDataSyncCR fails on errors from RGWListBucketIndexesCR

cbodley:
Fixes: http://tracker.ceph.com/issues/17073


oritwas:
lgtm

PR #10780, opened 2016-08-18
os/bluestore: add multiple finishers to bluestore

liewegas:
- The single finisher of a bluestore can be a bottleneck
when using an SSD as a backend device. If too much load
is given to the single finisher, client-side IO latency
increases. So we add multiple finishers to the
bluestore, which shows better performance.

- 'bluestore_shard_finishers' option is added to
be able to configure wheather finsihers is multiple or
not.

- a finisher is selected according to the shard id of a
sequencer.

- the number of finishers is decided by
osd_op_num_shards.

Signed-off-by: Ilsoo Byun <ilsoo.byun@sk.com>


liewegas:
retest this please

PR #10789, opened 2016-08-19
jewel: rbd bench-write: seg fault when --io-size is larger than image size 

dachary:
http://tracker.ceph.com/issues/17059

### DIFF
--- a/doc/man/8/rbd.rst
+++ b/doc/man/8/rbd.rst
@@ -358,6 +358,13 @@ Commands
   Release a lock on an image. The lock id and locker are
   as output by lock ls.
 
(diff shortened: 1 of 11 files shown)


dachary:
@dillaman the https://github.com/ceph/ceph/pull/9066 pull request was backported to avoid conflict (rename BenchWrite.cc into Bench.cc). It introduces a new feature but it looks like it is acceptable for a backport. What do you think ?

dillaman:
@dachary I'd like to avoid backporting new features (outside the feature gaps in rbd-mirror daemon).

dillaman:
@dachary Alternative PR: #10796

dachary:
Thanks for the alternate backport, much appreciated :-)

PR #10791, opened 2016-08-19
jewel: rbd-mirror: FAILED assert(m_state == STATE_STOPPING)

dachary:
http://tracker.ceph.com/issues/17004


dachary:

PR #10792, opened 2016-08-19
jewel: rbd-mirror: FAILED assert(m_on_update_status_finish == nullptr)

dachary:
http://tracker.ceph.com/issues/16978


dachary:

PR #10793, opened 2016-08-19
jewel: rbd-mirror: image deleter should use pool id + global image uuid for key

dachary:
http://tracker.ceph.com/issues/16902


dachary:
Replaced by https://github.com/ceph/ceph/pull/11433

PR #10795, opened 2016-08-19
os/bluestore: include bluefs space in statfs result

liewegas:
We were counting all (free) space allocated to bluefs as used.  Instead,
query bluefs to find out what is used and what is free.

### DIFF
--- a/src/os/bluestore/BlueStore.cc
+++ b/src/os/bluestore/BlueStore.cc
@@ -3388,15 +3388,20 @@ void BlueStore::_sync()
 
 int BlueStore::statfs(struct store_statfs_t *buf)
 {
-  uint64_t bluefs_len = 0;
-  for (interval_set<uint64_t>::iterator p = bluefs_extents.begin();
-      p != bluefs_extents.end(); p++)
-    bluefs_len += p.get_len();
-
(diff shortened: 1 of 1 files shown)


src/os/bluestore/BlueStore.cc
```diff
@@ -3388,15 +3388,18 @@ void BlueStore::_sync()
...
-  for (interval_set<uint64_t>::iterator p = bluefs_extents.begin();
-      p != bluefs_extents.end(); p++)
-    bluefs_len += p.get_len();
-
   buf->reset();
   buf->total = bdev->get_size();
-  assert(alloc->get_free() >= bluefs_len);
-  buf->available = (alloc->get_free() - bluefs_len);
+  buf->available = alloc->get_free();
+
```
somnathr: @liewegas Do we need the above as the below loop is counting usage for all device ? Probably, needed only if bluefs is not enabled ?

somnathr: @liewegas sorry, my bad , it is only collecting usage for all the devices for db/wal. We are good it seems..I will verify..

varadakari:
LGTM

PR #10796, opened 2016-08-19
jewel: rbd: bench io-size should not be larger than image size

dillaman:
http://tracker.ceph.com/issues/17059


dachary:
It passed the rbd suite http://tracker.ceph.com/issues/17487#note-10 (except for two failures that are unrelated http://tracker.ceph.com/issues/17590 and http://tracker.ceph.com/issues/17262)

PR #10798, opened 2016-08-19
DNM: jewel: non-primary image is recording journal events during image sync 

trociny:
http://tracker.ceph.com/issues/16903


dachary:
Replaced by

PR #10800, opened 2016-08-19
ceph-post-file replace DSA with RSA ssh key

djgalloway:
DSA keys are being deprecated: http://www.openssh.com/legacy.html

drop.ceph.com will continue to allow the old DSA key but eventually,
users submitting logs using ceph-post-file will run into issues when
OpenSSH completely drops support for the algorithm.

Fixes: http://tracker.ceph.com/issues/14267


liewegas:
rados suite has a ceph-post-file test

tchaikov:
tested at http://pulpito.ceph.com/kchai-2016-08-20_18:41:17-rados-wip-kefu-testing---basic-mira/

failed ones are tracked by 

* http://tracker.ceph.com/issues/17087
* http://tracker.ceph.com/issues/16157
* http://tracker.ceph.com/issues/16800

@djgalloway could you put 
```
Fixes: http://tracker.ceph.com/issues/14267
```

because `#<number>` could collide with the pull request number. see also https://github.com/ceph/ceph/blob/master/SubmittingPatches.rst#preparing-and-sending-patches

otherwise, lgtm.

djgalloway:
@tchaikov: Fixed commit message

PR #10805, opened 2016-08-22
ceph.spec.in: don't try to package __pycache__

tserong:
When building on openSUSE Tumbleweed, nothing seems to create
the various `__pycache__` directories (so the build fails because
those files don't exist), and in any case they should be
created automatically at runtime, so shouldn't need to be
packaged.

### DIFF
--- a/ceph.spec.in
+++ b/ceph.spec.in
@@ -1433,15 +1433,19 @@ ln -sf %{_libdir}/librbd.so.1 /usr/lib64/qemu/librbd.so.1
 %{python3_sitearch}/cephfs.cpython*.so
 %{python3_sitearch}/cephfs-*.egg-info
 %{python3_sitelib}/ceph_volume_client.py
+%if ! 0%{?suse_version}
 %{python3_sitelib}/__pycache__/ceph_volume_client.cpython*.py*
+%endif
 
 #################################################################################
 %files -n python%{python3_pkgversion}-ceph-argparse
 %defattr(-,root,root,-)
 %{python3_sitelib}/ceph_argparse.py
-%{python3_sitelib}/__pycache__/ceph_argparse.cpython*.py*
 %{python3_sitelib}/ceph_daemon.py
+%if ! 0%{?suse_version}
+%{python3_sitelib}/__pycache__/ceph_argparse.cpython*.py*
 %{python3_sitelib}/__pycache__/ceph_daemon.cpython*.py*
+%endif
 
 #################################################################################
 %files -n ceph-test


tserong:
Probably need to adjust this to use an `%if ! 0%{?suse_version}` guard (see http://tracker.ceph.com/issues/17106)

smithfarm:
@tserong +1

smithfarm:
Also, please add a "Fixes: http://tracker.ceph.com/issues/17106" line before the signature.

tserong:
Done. Running a test build now.

tserong:
...and my test build built fine with this change on openSUSE Tumbleweed.

smithfarm:
Since the changes are in conditional blocks I don't see how they could cause a regression, but pushed `wip-dont-package-__pycache__` out of an abundance of caution.

smithfarm:
This passes gitbuilder. @ktdreyer @b-ranto Care to review?

smithfarm:
**running** http://pulpito.front.sepia.ceph.com:80/smithfarm-2016-08-25_14:02:16-upgrade:jewel-x-wip-dont-package-__pycache__---basic-smithi/

ktdreyer:
It's surprising that you don't have `__pycache__` on Tumbleweed, but LGTM

smithfarm:
Triggered a build in OBS so we can look at the exact error: https://build.opensuse.org/package/show/filesystems:ceph:Unstable/ceph

smithfarm:
pushed `wip-dont-package-__pycache__` to gitbuilder (remind self to delete it afterwards)

dirkmueller:
@toabctl: any opinion about this? It feels like this is papering over some other problem, namely a missing BuildRequires that pulls in the corresponding Packaging magic.. but I don't know enough about that without access to a computer

dirkmueller:
Definitely makes it more consistent with the already existing exception for suse sisters, so lgtm

smithfarm:
full build logs:

* Leap 42.1: https://build.opensuse.org/build/filesystems:ceph:Unstable/openSUSE_42.1/x86_64/ceph/_log
* Tumbleweed: https://build.opensuse.org/build/filesystems:ceph:Unstable/openSUSE_Tumbleweed/x86_64/ceph/_log

(error is the same)

ktdreyer:
Even though the root cause is unknown, `wip-dont-package-__pycache__` builds in Gitbuilder, so let's go ahead with this

smithfarm:
I think we need

```
%py3_compile %{buildroot}%{python3_sitelib}
```

at the end of %install section. Testing this approach now.

smithfarm:
Credits to @dimstar77 for suggesting this!

smithfarm:
See #10894

PR #10808, opened 2016-08-22
osd/ReplicatedPG:wipe off param "op" from generate_subop()

jimifm:
Signed-off-by: jimifm <hong.zhangoole@gmail.com>


tchaikov:
@jimifm please squash these two commits into a single one, also
could you prefix the title of your commit message with the subcomponent your are changing ? see https://github.com/ceph/ceph/blob/master/SubmittingPatches.rst#preparing-and-sending-patches . in this case, it's "osd: " or "osd/ReplicatedPG: ".

PR #10814, opened 2016-08-22
os/bluestore: precondition rocksdb/bluefs during mkfs

liewegas:
Write a bunc hof data through rocksdb so that we get all of the rocksdb log
files preallocated and there isn't a weird performance anamoly once the real
workload starts.

Simplify rocksdb options.

### DIFF
--- a/src/common/config_opts.h
+++ b/src/common/config_opts.h
@@ -960,6 +960,8 @@ OPTION(bluestore_block_wal_path, OPT_STR, "")
 OPTION(bluestore_block_wal_size, OPT_U64, 96 * 1024*1024) // rocksdb wal
 OPTION(bluestore_block_wal_create, OPT_BOOL, false)
 OPTION(bluestore_block_preallocate_file, OPT_BOOL, false) //whether preallocate space if block/db_path/wal_path is file rather that block device.
+OPTION(bluestore_precondition_bluefs, OPT_U64, 128*1024*1024)  // write this much data at mkfs
+OPTION(bluestore_precondition_bluefs_block, OPT_U64, 1048576)
 OPTION(bluestore_csum_type, OPT_STR, "crc32c") // none|xxhash32|xxhash64|crc32c|crc32c_16|crc32c_8
 OPTION(bluestore_min_csum_block, OPT_U32, 4096)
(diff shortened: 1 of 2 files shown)


src/os/bluestore/BlueStore.cc
```diff
@@ -2738,6 +2738,36 @@ int BlueStore::mkfs()
...
+    unsigned n = g_conf->bluestore_precondition_bluefs /
+      g_conf->bluestore_precondition_bluefs_block;
+    bufferlist bl;
+    bufferptr bp(g_conf->bluestore_precondition_bluefs_block);
+    for (unsigned i=0; i < g_conf->bluestore_precondition_bluefs_block; ++i) {
+      bp[i] = rand();
+    }
+    bl.append(bp);
+    string key1("a");
+    string key2("a");
```
ifed01: the same key?

src/common/config_opts.h
```diff
@@ -988,7 +988,7 @@ OPTION(bluestore_freelist_type, OPT_STR, "bitmap") // extent | bitmap
 OPTION(bluestore_freelist_blocks_per_key, OPT_INT, 128)
 OPTION(bluestore_bitmapallocator_blocks_per_zone, OPT_INT, 1024) // must be power of 2 aligned, e.g., 512, 1024, 2048...
 OPTION(bluestore_bitmapallocator_span_size, OPT_INT, 1024) // must be power of 2 aligned, e.g., 512, 1024, 2048...
-OPTION(bluestore_rocksdb_options, OPT_STR, "compression=kNoCompression,max_write_buffer_number=16,min_write_buffer_number_to_merge=3,recycle_log_file_num=16")
+OPTION(bluestore_rocksdb_options, OPT_STR, "compression=kNoCompression,write_buffer_size=16777216,recycle_log_file_num=3")
```
ifed01: In my brief benchmarking (sequential 4k writes to 2000 objects - 8 Gb total) new settings decrease bandwidth from 28Mbs to 8Mbs and increases total duration from 280 s to 1000 s.

liewegas:
dropped the options commit.. we'll need to be more careful there

PR #10815, opened 2016-08-22
Python 3 compatibility for workunits

onyb:


qa/workunits/fs/misc/direct_io.py
```diff
@@ -34,14 +36,15 @@ def main():
...
     os.lseek(fd, 0, os.SEEK_SET)
-    out = os.read(fd, 1);
+    out = os.read(fd, 1)
     if out != '2':
         raise RuntimeError("data were not directly read from object store")
 
-    os.close(fd);
-    print 'ok'
+    os.close(fd)
+    print('ok')
```
tchaikov: do we need 
```py
from __future__ import print_function
```
for this?

onyb: @tchaikov Since the `print` statement is a very simple one, we don't need that line.

tchaikov: oic, `print()` is introduced in py2.6, and we've bumped the supported python version to >= 2.6

tchaikov:
and needs to run through ceph-disk and cephfs qa suites

tchaikov:
ceph-disk test passed at http://pulpito.ceph.com/kchai-2016-09-19_07:05:39-ceph-disk-wip-kefu-testing---basic-vps/

PR #10820, opened 2016-08-23
rgw: fix get object instance  returned NoSuchKey error

yanghonggang:
When accessing the copied destination object, the source object's instance ID 
information is needed, however it's missing now in the destination object’s 
manifest. This will cause an IO error when get the copied object. In order to fix 
this problem, we can record source object's version_id/instance into dest object's 
manifest.

Signed-off-by: Yang Honggang <joseph.yang@xtaotech.com>

### DIFF
--- a/src/rgw/rgw_rados.h
+++ b/src/rgw/rgw_rados.h
@@ -254,6 +254,8 @@ protected:
                              as object might have been copied across buckets */
   map<uint64_t, RGWObjManifestRule> rules;
 
+  string tail_instance; /* tail object's instance */
+
   void convert_to_explicit();
   int append_explicit(RGWObjManifest& m);
   void append_rules(RGWObjManifest& m, map<uint64_t, RGWObjManifestRule>::iterator& iter, string *override_prefix);
@@ -279,6 +281,7 @@ public:
     prefix = rhs.prefix;
     tail_bucket = rhs.tail_bucket;
     rules = rhs.rules;
+    tail_instance = rhs.tail_instance;
 
     begin_iter.set_manifest(this);
     end_iter.set_manifest(this);
@@ -316,7 +319,7 @@ public:
   }
 
   void encode(bufferlist& bl) const {
-    ENCODE_START(4, 3, bl);
+    ENCODE_START(5, 3, bl);
     ::encode(obj_size, bl);
     ::encode(objs, bl);
     ::encode(explicit_objs, bl);
@@ -326,11 +329,12 @@ public:
     ::encode(prefix, bl);
     ::encode(rules, bl);
     ::encode(tail_bucket, bl);
+    ::encode(tail_instance, bl);
     ENCODE_FINISH(bl);
   }
 
   void decode(bufferlist::iterator& bl) {
-    DECODE_START_LEGACY_COMPAT_LEN_32(4, 2, 2, bl);
+    DECODE_START_LEGACY_COMPAT_LEN_32(5, 2, 2, bl);
     ::decode(obj_size, bl);
     ::decode(objs, bl);
     if (struct_v >= 3) {
@@ -352,6 +356,12 @@ public:
 
     if (struct_v >= 4) {
       ::decode(tail_bucket, bl);
+    }
+
+    if (struct_v >= 5) {
+      ::decode(tail_instance, bl);
+    } else { // old object created before 'tail_instance' field added to manifest
+      tail_instance = head_obj.get_instance();
     }
 
     update_iterators();
@@ -409,6 +419,14 @@ public:
 
   const string& get_prefix() {
     return prefix;
+  }
+
+  void set_tail_instance(const string& _ti) {
+    tail_instance = _ti;
+  }
+
+  const string& get_tail_instance() {
+    return tail_instance;
   }
 
   void set_head_size(uint64_t _s) {
--- a/src/rgw/rgw_dencoder.cc
+++ b/src/rgw/rgw_dencoder.cc
@@ -188,6 +188,10 @@ void RGWObjManifest::get_implicit_location(uint64_t cur_part_id, uint64_t cur_st
   }
 
   location->init_ns(*bucket, oid, ns);
+  
+  // Always overwrite instance with tail_instance
+  // to get the right shadow object location
+  location->set_instance(tail_instance);
 }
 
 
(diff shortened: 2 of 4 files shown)


yanghonggang:
@yehudasa   bug report:  http://tracker.ceph.com/issues/17111

src/rgw/rgw_rados.h
```diff
@@ -254,6 +254,14 @@ class RGWObjManifest {
                              as object might have been copied across buckets */
   map<uint64_t, RGWObjManifestRule> rules;
 
+  /* If this object is copied from another versioned object, @src_instance 
+   * is used to store src's version_id. 
+   * 
+   * @copied_obj is used to indicate whether @src_instance is valid.
+   * */
+  string src_instance; 
+  bool copied_obj;
```
yehudasa: @yanghonggang do we really need to keep copied_obj field? Isn't it that we should just check to see that `!src_instance.empty()`?

yanghonggang:
obj a ---copy---->obj b ---copy---> obj c

without copied_obj，we can't decide obj c's src_instance's value. it should be obj b's src_instance (obj a's version_id) or obj b's version_id.

yanghonggang:
@yehudasa

src/rgw/rgw_rados.h
yanghonggang: @yehudasa 
obj a (instance="", src_instance="") ---copy----> obj b (version_id="abc', src_instance="") ---copy---> obj c

without copied_obj，we can't decide obj c's src_instance's value. should it be obj b's src_instance (obj a's version_id) or obj b's version_id? Because empty src_instance is valid which should be copied into dest object's manifest in this senario.

yanghonggang:
@yehudasa

src/rgw/rgw_rados.h
yehudasa: @yanghonggang if src_instance is not empty, use it, otherwise continue as previously. Does that make sense?

yanghonggang: @yehudasa 
`The following example shows that we can not decide dest object's src_instance value 
only by source object's src_instance. All the buckets in this example share the same
data pool.

+------------------+            +--------------------+           +---------------------------+ 
|Bucket A          |            |   Bucket B         |           |   Bucket C                |   
|                  |            |                    |           |                           |  
|versioning CLOSED |            | versioning ENABLED |           | versioning ENABLED        |   
|------------------+            |--------------------+           |---------------------------+      
|Object AA         | - copy1 -> |Object BB           | - copy2 ->|Object CC                  |         
|                  |            |                    |           |                           | 
|instance = ""     |            |instance = "abcd123"|           |instance = "xyz789"        |    
|src_instance = "" |            |src_instance = ""   |           |src_instance = "???abcd123"|   
+------------------+            +--------------------+           +---------------------------+   

copy1:

Because object AA's src_instance is empty, use ""(object AA's instance) as object BB's src_instance.

It's OK.

copy2:

Because object BB's src_instance is empty, use "abcd123"(object BB's instance) as object CC's src_instance.

This will cause an ERROR. Because we should use "" as object CC's src_instance, not "abcd123".

-------------

So with 'coped_obj' we can decide if src object is a object generated by a copy operation. 
If source object is not a copied object, we must copy its 'instance' to dest object's 'src_instance'. 
If source object is a copied object, we can directly copy its 'src_instance' to dest object' 'src_instance'.

In the copy chain, only the first time copy need to copy source object's 'instance' to dest object's 
'src_instance'.

AA --copy*--> BB --copy--> CC --copy--> DD ...     `

yanghonggang: @yehudasa 

The following example shows that **we can not decide dest object's src_instance value 
only by source object's src_instance**.  In the following exmaple, **source object's empty src_instance still need to be copied into dest object's src_instance**. All the buckets in this example share the same
data pool. 
  
I want to copy object AA to generate object BB, and copy object BB to generate object CC, ...

copy chain:
AA --copy1--> BB --copy2--> CC --copy--> DD ...
                  
Bucket A                     |      Bucket B                       |  Bucket C                        
------------------------- |       ---------------------------|   ---------------------------- 
versioning CLOSED  |      versioning ENABLED |    versioning ENABLED              
                                     |                                            |
 Object AA      ->         |      Object BB      ->           |    Object CC 
                                     |                                            |    
instance = ""                |     instance = "abcd123"  |    instance = "xyz789"               
src_instance = ""         |    src_instance = ""          |     src_instance = "???abcd123"            
                 
copy1:
Because object AA's src_instance is empty, use ""(object AA's instance) as object BB's src_instance.

It's OK.

copy2:
Because object BB's src_instance is empty, use "abcd123"(object BB's instance) as object CC's src_instance.

This will cause an **ERROR**. Because we should use "" as object CC's src_instance, not "abcd123".

-------------
So with 'coped_obj' we can decide if src object is a object generated by a copy operation. 
If source object is not a copied object, we must copy its 'instance' to dest object's 'src_instance'. 
If source object is a copied object, we can directly copy its 'src_instance' to dest object' 'src_instance'.

In the copy chain, only the first time copy need to copy source object's 'instance' to dest object's 
'src_instance'.

yanghonggang:
@yehudasa 
This is an optimized version in which I get rid of copied_obj. 
[https://github.com/ceph/ceph/pull/10820/commits/5104f852d255cfc140be33885f696b7f93a6cfad](optimization)

yanghonggang:
@oritwas Your comments are appreciated

src/rgw/rgw_rados.h
yehudasa: @yanghonggang ok, I understand the issue now. Instead of calling the field "tail_instance", it should really be use_instance. By default it should match the instance field, unless it's copied and in that case it should match the copied object instance.

yehudasa:
@yanghonggang looking good. Can you squash all the commits together?

yanghonggang:
@yehudasa All commits are squashed. 
But there are some error:
97% tests passed, 5 tests failed out of 150

Total Test time (real) = 583.88 sec

The following tests FAILED:
	  5 - cephtool-test-mon.sh (Failed)
	  8 - run-rbd-unit-tests.sh (Failed)
	103 - unittest_journal (OTHER_FAULT)
	126 - unittest_bluestore_types (OTHER_FAULT)
	150 - unittest_rbd_mirror (OTHER_FAULT)
Errors while running CTest

yehudasa:
@yanghonggang these test failures are unrelated

PR #10823, opened 2016-08-23
rgw: rgw file fix bug of rgw_lookup can not exact match file name

cmgitdream:
bug reproduce steps:
in nfs-client (connect to nfs-ganesha server):
1. mv file file.rename
2. stat file // get file attrs, not return with -ENOENT

the reason is that:
RGWStatLeafRequest does not exact match the file name,
just take file name as a prefix filter

Signed-off-by: Min Chen <chenmin@xsky.com>

### DIFF
--- a/src/rgw/rgw_file.cc
+++ b/src/rgw/rgw_file.cc
@@ -163,6 +163,15 @@ namespace rgw {
 	if ((rc == 0) &&
 	    (req.get_ret() == 0)) {
 	  if (req.matched) {
+	    // we need rgw object's key name equal to file name, if not return NULL
+	    if ((flags & RGWFileHandle::FLAG_EXACT_MATCH) &&
+		!req.exact_matched) {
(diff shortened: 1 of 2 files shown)


cmgitdream:
@mattbenjamin

mattbenjamin:
@cmgitdream I am unable to verify the bad stat behavior on master without this change, but on the other hand, the change looks plausible and clean, and, I also verified that there is no regression.  So, I'm merging this.  There is a conflict, so I'm pushing a merge commit and will close this PR.

cmgitdream:
@mattbenjamin 
oh, I found this bug on nfs-ganesha v2.2.0 with FSAL_RGW, which is implemented by myself.  So there maybe some different behavors from nfs-ganeshs v2.4-xx with FSAL_RGW upstream.  
The key point of this bug is that  both src file and dest file have the same pattern, eg. src_file = file1, dst_file = file1.rename, and they share the common prefix "file1"

PR #10827, opened 2016-08-23
jewel: librbd: delay acquiring lock if image watch has failed

dillaman:
http://tracker.ceph.com/issues/16950


dachary:

PR #10831, opened 2016-08-23
rgw: delete region map after upgrade to zonegroup map

cbodley:
convert_regionmap() reads the region map and uses it to initialize the zonegroup map.  but it doesn't remove the region_map afterwards, so radosgw (and some radosgw-admin commands) will keep doing this on startup, overwriting any changes made to the period/zonegroup map

Fixes: http://tracker.ceph.com/issues/17051


cbodley:
@robbat2 would you be willing to help validate this fix?

oritwas:
lgtm

robbat2:
Confirmed working, it does correctly delete the region_map object on commit.

PR #10834, opened 2016-08-23
global: fix memory leaks related to global_init_daemonize()

cbodley:
These are two band-aids to fix memory leaks related to `global_init_daemonize()`.

Fixes: http://tracker.ceph.com/issues/17108
Fixes: http://tracker.ceph.com/issues/17109

### DIFF
--- a/src/common/admin_socket.h
+++ b/src/common/admin_socket.h
@@ -77,13 +77,13 @@ public:
 
   bool init(const std::string &path);
 
+  void shutdown();
+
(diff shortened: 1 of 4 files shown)


cbodley:
runs cleanly in local testing, but teuthology still complains about CephContext leaks: http://pulpito.ceph.com/cbodley-2016-08-24_11:02:38-rgw-wip-daemonize-service-thread---basic-mira/

PR #10835, opened 2016-08-24
common: audit OPT_INT/OPT_UINT and handle incorrect signed uint

oliveiradan:
Changes proposed to address issue: BUG #14934 
http://tracker.ceph.com/issues/14934


dachary:
needs rebasing

liewegas:
needs rebase

src/ceph_mon.cc
```diff
@@ -480,7 +480,7 @@ int main(int argc, const char **argv)
            << std::endl;
       exit(-err);
     }
-    if (stats.avail_percent <= g_conf->mon_data_avail_crit) {
+    if (static_cast<uint32_t>(stats.avail_percent) <= g_conf->mon_data_avail_crit) {
```
amitkumar50: @oliveiradan Please keep PR title < 50 characters

src/common/config.cc
```diff
@@ -946,83 +952,177 @@ int md_config_t::set_val_impl(const char *val, const config_option *opt)
 
 int md_config_t::set_val_raw(const char *val, const config_option *opt)
 {
+  const int CEPH_SETTING_SAFE = 1;
```
amitkumar50: @oliveiradan Why local variable is taken as caps?
Conventionally we take macros as this?

oliveiradan: @amitkumar50 , that was when I started working on Ceph (that's an old PR) and I used it as a way to identify a constant. However, as it ended up getting old and a rebase was needed, I ended up not touching it (once it is a pretty big change) 'till we were certain this PR would still be useful or not. Should this PR still be interesting, I will rebase and rework whatever is needed.

Thanks,

liewegas:
This is obsolete with the new options framework.  What we probably want instead is to translate this into min/max values in options.cc

PR #10836, opened 2016-08-24
rgw: fix period update --commit return error

weiqiaomiao:
period commit return error when the current period has a zonegroup which doesn't have a master zone

Fixes: http://tracker.ceph.com/issues/17110


cbodley:
tested and working, thanks!

PR #10837, opened 2016-08-24
cmake: do not link against unused objects or libraries

tchaikov:


wjwithagen:
@tchaikov 
I'll run this thru my FreeBSD config.
But I will first have to fix a conflict between std-includeds and the log system.
I think this due my OS upcrade to the current FreeBSD HEAD :-( 
```
In file included from /usr/include/c++/v1/cmath:301:
/usr/include/c++/v1/math.h:845:37: error: reference to 'log' is ambiguous
log(_A1 __lcpp_x) _NOEXCEPT {return log((double)__lcpp_x);}
                                    ^
/usr/include/c++/v1/math.h:845:1: note: candidate found by name lookup is 'log'
log(_A1 __lcpp_x) _NOEXCEPT {return log((double)__lcpp_x);}
^
/usr/include/c++/v1/math.h:839:46: note: candidate found by name lookup is 'log'
inline _LIBCPP_INLINE_VISIBILITY long double log(long double __lcpp_x) _NOEXCEPT {return logl(__lcpp_x);}
                                             ^
/usr/include/c++/v1/math.h:838:46: note: candidate found by name lookup is 'log'
inline _LIBCPP_INLINE_VISIBILITY float       log(float __lcpp_x) _NOEXCEPT       {return logf(__lcpp_x);}
                                             ^
/usr/include/math.h:247:8: note: candidate found by name lookup is 'log'
double  log(double);
        ^
/home/wjw/ceph/src/common/ceph_context.h:44:13: note: candidate found by name lookup is 'ceph::log'
  namespace log {
            ^

```

wjwithagen:
@tchaikov 
LGTM

Currently I only compile librbd on FreeBSD, but there I already had the stuff you are fixing removed.
So that is a certainly oke on FreeBSD.
With the other fixes there is a big change I can start compiling and test more of the RBD stuff.

PR #10841, opened 2016-08-24
Add udev rule to set readahead on mapped rbd's

fiskn:
Set to 128kb by default, so no change from current behaviour

### DIFF
--- a/udev/50-rbd.rules
+++ b/udev/50-rbd.rules
@@ -1,2 +1,5 @@
 KERNEL=="rbd[0-9]*", ENV{DEVTYPE}=="disk", PROGRAM="/usr/bin/ceph-rbdnamer %k", SYMLINK+="rbd/%c{1}/%c{2}"
 KERNEL=="rbd[0-9]*", ENV{DEVTYPE}=="partition", PROGRAM="/usr/bin/ceph-rbdnamer %k", SYMLINK+="rbd/%c{1}/%c{2}-part%n"
+
+# This is a placeholder, uncomment and edit as necessary
+#KERNEL=="rbd[0-9]*", ENV{DEVTYPE}=="disk", ACTION=="add|change", ATTR{bdi/read_ahead_kb}="128"


wido:
Thanks!

I think it should go in the udev directory right? Afterwards in debian/rules add it to the proper packages. Same goes for the RPM spec file.

idryomov:
Can you add a comment to the rule file, clarifying that this is just a placeholder and doesn't change the kernel default?

fiskn:
As suggest by @wido closing this PR and moving this into the already packaged /usr/lib/udev/50-rbd.rules

fiskn:
@wido Could you just check this new commit before I do another PR

https://github.com/fiskn/ceph/commit/f4febec9788e5b02bf0fc12f0fd2a193d6e8619e

wido:
@fiskn Thanks! However, if you check the current src directory, there is already a file.

Just check src/udev in the source.

You can squash all the commits and force push. No need for a new PR.

fiskn:
@wido, sorry I don't understand, the udev folder was in the root of the repository and I don't see anything in the src folder relating to udev.

Ok, will do that

wido:
Looks good now! Kan you use git rebase to squash all the commits into on? Since you are only changing one file.

Then do a git force push and you should be good.

dillaman:
lgtm

dillaman:
@fiskn Can you run "git commit --amend -s" to add a signed-off-by line to your commit?

idryomov:
I suggested the explanatory comment, but I now think the rule itself needs to be commented out so that we can actually change the kernel default some day and also to save an unnecessary step on map (rbd map CLI waits for udev).
```
# This is a placeholder, uncomment and edit as necessary
#KERNEL== ...
```

dillaman:
@fiskn ping

idryomov:
@fiskn Could you please add a space between # and This and change commit title to "udev: add krbd readahead placeholder" and I'll merge?

PR #10842, opened 2016-08-24
First pile of ceph_statx patches

jtlayton:
This patchset is the first pile of patches to add ceph_statx support to the userland client. The basic idea is to add a new interface patterned after the proposed statx() interface for Linux. The idea there is to extend the traditional stat() interface in several ways:

1) allow access to more inode attribute fields, in particular the birthtime (btime) and a change attribute (a'la NFSv4), in a way that is extensible in the future

2) allow applications to request just specific fields in the ceph_statx struct be filled out. Note that you can (and will) often get more than you request, but you should always get what is requested or an error.

3) allow applications to do a "lazy" statx, where we fill out the attribute fields from whatever we have in cache, without querying the server. This is handy for performance reasons or in cases where we don't really need up-to-date information.

In addition, a ceph_setattrx has also been added, which allows the setting of the btime as well as the other fields that ceph_setattr allows. I also have a companion patchset for samba that converts it to use these new interfaces. That seems to work properly, though I still need to roll a testcase to verify that setting the btime via smb actually works.

Pay close attention to the first patch in the file as well, which I think may help performance in some cases by allowing us to request the caps needed to do an operation during the lookup phase of path-based operations.

I've also started some work to do a patchset for ganesha, but that requires adding a number of new ceph_ll_* calls. I think it's probably best to get this merged first and then do that part afterward.

Testing this has been somewhat problematic. I've been fighting a number of unrelated testcase failures, but the new code _seems_ to do the right thing so far.


src/client/Client.cc
```diff
@@ -9623,6 +9628,7 @@ Inode *Client::open_snapdir(Inode *diri)
     in->ctime = diri->ctime;
     in->btime = diri->btime;
     in->size = diri->size;
+    in->change_attr = diri->change_attr;
```
gregsfortytwo: We shouldn't actually need this on a snapdir; it's just for consistency, right?

src/mds/Locker.cc
```diff
@@ -2991,6 +2992,10 @@ void Locker::_update_cap_fields(CInode *in, int dirty, MClientCaps *m, inode_t *
 	      << " for " << *in << dendl;
       pi->mtime = mtime;
     }
+    if (((dirty & CEPH_CAP_FILE_WR) && change_attr > pi->change_attr) ||
+	((dirty & CEPH_CAP_FILE_EXCL) && change_attr != pi->change_attr)) {
+      pi->change_attr = change_attr;
```
gregsfortytwo: Should we bump change_attr here even if the client value isn't greater? That would let us at least do *something* with older clients.

src/client/Client.cc
```diff
@@ -6524,13 +6524,17 @@ int Client::_do_setattr(Inode *in, struct stat *attr, int mask, int uid, int gid
...
+    in->change_attr++;
     return 0;
+  }
 
 force_request:
   MetaRequest *req = new MetaRequest(CEPH_MDS_OP_SETATTR);
 
   filepath path;
+
+  in->change_attr++;
```
gregsfortytwo: So we're doing the local bump here, but then it gets bu,led again on the remote side, right?

If we want to do it locally on all conditions, should we do it above instead of duplicating between the remote and local-only code paths?

src/mds/mdstypes.h
```diff
@@ -181,18 +182,24 @@ struct frag_info_t : public scatter_info_t {
   }
 
   // *this += cur - acc;
-  void add_delta(const frag_info_t &cur, frag_info_t &acc, bool& touched_mtime) {
+  void add_delta(const frag_info_t &cur, frag_info_t &acc, bool& touched_mtime, bool& touched_chattr) {
```
gregsfortytwo: Since you're touching it, can you switch this to use pointers for the out params? That's what our style guide calls for, although we obviously have a bunch of old "naughty" code like you're seeing here.

src/mds/MDCache.cc
```diff
@@ -11997,6 +12001,8 @@ void MDCache::repair_dirfrag_stats_work(MDRequestRef& mdr)
   if (!good_fragstat) {
     if (pf->fragstat.mtime > frag_info.mtime)
       frag_info.mtime = pf->fragstat.mtime;
+    if (pf->fragstat.change_attr > frag_info.change_attr)
+      frag_info.change_attr = pf->fragstat.change_attr;
```
gregsfortytwo: I'm having trouble reasoning about this right now. When can a client see an individual frag's change_attr? are there any negative consequences if eg it sees the change_attr from a local operation, then a scatter-gather happens and it sees the same directory change_attr despite there being new changes in view from another frag?

src/messages/MClientRequest.h
```diff
@@ -160,7 +162,17 @@ class MClientRequest : public Message {
 
   void decode_payload() {
     bufferlist::iterator p = payload.begin();
-    ::decode(head, p);
+
+    if (header.version >= 4) {
+      struct ceph_mds_request_head_new	new_mds_head;
+
+      ::decode(new_mds_head, p);
+      memcpy(&head, &(new_mds_head.oldest_client_tid), sizeof(head));
```
gregsfortytwo: Eww. We normally do this the other way around: switch to the new structure natively, and instantiate the old one for decoding/encoding if needed. Granted this works but it might not if we change head again. 
(Also we can use the ceph_mds_request_head name for our new structure and just append _old to the classic one that way.)

src/include/ceph_fs.h
```diff
@@ -469,6 +470,77 @@ struct ceph_mds_request_head {
...
+		__le64 start; /* initial location to lock */
+		__le64 length; /* num bytes to lock from start */
+		__u8 wait; /* will caller wait for lock to become available? */
+	} __attribute__ ((packed)) filelock_change;
+} __attribute__ ((packed));
+
+#define CEPH_MDS_REQUEST_HEAD_VERSION	1
+
+struct ceph_mds_request_head_new {
+	__le16 version;
```
gregsfortytwo: We need to note the link between this and the args structure (probably in both places).

src/mds/Locker.cc
jtlayton: Yeah, that may be the safest way to handle it, but I'm not sure this is the right spot to do that. It may also legitimately not have changed here because a client that does support the changeid field had the caps and nothing actually changed.

What we probably want to do is detect when we have an older client and then just preemptively bump the changeid in that situation.  I'll have to think about that piece.

src/client/Client.cc
jtlayton: I'm not sure I understand that bit. Why would it get bumped again on the remote side? My understanding is that when the client returns exclusive caps, the MDS just accepts what is in the cap message from the client as gospel. Is that not the case?

src/mds/mdstypes.h
jtlayton: Sure. Will fix.

src/mds/MDCache.cc
jtlayton: My understanding (quite possibly wrong) is that a client can see the frag attrs whenever it has the shared or exclusive cap for it. To actually _view_ the change_attr though, you would need to do a gather operation and the change_attrs would get synched up at that point.

So I don't think this is problematic, but please do sanity check me here.

src/messages/MClientRequest.h
jtlayton: Understood. I felt it was pretty icky too. I'll see about switching this around handle it that way.

src/include/ceph_fs.h
jtlayton: Ok, will add.

src/mds/Locker.cc
gregsfortytwo: Yeah, that works too. We should just be able to check the connection features; accessible via the message's connection pointer/reference.

src/client/Client.cc
gregsfortytwo: You appear to have removed this, and I didn't mean to say it was *wrong*, I just wanted to double-check. :)
(And actually we do allow snaps to be renamed at present, which could maybe bump the change attr...not sure how that will play out in future or if it's a thing we want to support).

src/client/Client.cc
gregsfortytwo: In this branch we're doing an MDS request to setattr, and handle_client_setattr bumps the change_attr — or I guess we don't include the change_attr on regular requests, just on cap messages, so bumping it on both sides merely keeps them in sync?

But if the MDS rejects the setattr request for any reason, then we've incorrectly bumped it locally. And we fill in our trace from the MDS on a setattr, so we'll update it from the MDS anyway, right? So I think we don't want to bump it ourselves if we're delegating the actual change to the MDS. Does that make sense?

src/mds/MDCache.cc
gregsfortytwo: This sounds right to me as well. Just wasn't sure at midnight last night. ;)

src/client/Client.cc
jtlayton: Ahh yes it does -- thanks. I agree that removing that increment is the right thing to do.

src/mds/Locker.cc
jtlayton: Ok, I'll look at that too. FWIW, what ganesha does currently is fake up the change_attr from the ctime when the underlying storage engine doesn't have a real change attr. So we could follow suit here -- bump the change_attr if the client didn't return a change_attr with the caps, and it looks like the ctime has changed.

gregsfortytwo: 👍

jtlayton: Hmm, actually now that I think about it, I don't think I quite have the right semantics for the change_attr when we have multiple clients. I put the change_attr under CEPH_CAP_FILE_EXCL capability, but I don't think that's quite sufficient.

Suppose we have 2 clients -- one with CEPH_CAP_FILE_EXCL and one with CEPH_CAP_AUTH_EXCL. The AUTH_EXCL client does (e.g.) a mode change, while the other client does a single write. Both bump the change_attr by 1.

Next, the server ends up recalling caps from both clients, and the FILE_EXCL client then checks the change_attr again. It hasn't changed though the mode has...

Actually, now that I think about it, that's probably OK since you're not allowed to infer anything about the value of the change_attr. Applications must treat this as an opaque token that can only be checked for equivalence.

Between the point where the client fetched the mode for its original check and where it does the write, it's ok for the mode to change. All the client knows is that after the write, the change_attr was bumped, and so any cached data and metadata is invalid at that point.

jtlayton: Hmm...but is it a problem on the AUTH_EXCL client? That part makes me a little uneasy since the change_attr is supposed to be under the aegis of CEPH_CAP_SFILE. We can end up bumping it here, even though we are doing a cached changed while only holding the CEP_CAP_AUTH_EXCL.

Still, I don't think it's a huge issue since in order for the client to query for the change_attr, the MDS would need to recall the FILE_EXCL cap. Ugh, these scenarios are hard to work though!

gregsfortytwo: Wait, I've forgotten: is change_attr supposed to increment on inode changes, or on any change to the file (eg, a data overwrite)?

Do we need to bump the change_attr for every client that incremented it, instead of just taking the ceiling? If it's covered by FILE_EXCL then nobody else should be able to read it without getting an up-to-date version from the MDS, but they also shouldn't be able to *set* it except by talking to the MDS. That doesn't seem to be what's happening here with the setattr code.

jtlayton:
I'll close this for now, until the changes are done and have been tested. Thanks for the comments so far!

src/mds/Locker.cc
jtlayton: The rule is basically that the change attr must be different from when you last queried for it iff _something_ changed. That something could be an inode metadata or data change. The application is not allowed to infer anything from the value itself, however. It can only test it for equivalence with an earlier value. So, it doesn't really matter how much the thing changes as long as the above rule is followed.

ffilz: At a quick thought, it seems like it needs to be ceiling+1 if multiple clients bumped it. Because a NFS client of a Ganesha instance that is thusly a ceph client that happened to have the ceiling change attr needs to see it change again to realize the other ceph clients ALSO changed the file...

jtlayton: Actually, this is a little off. We need the mds to take the latest change_attr seen, regardless of what caps are being returned, as long as some of them are dirty (much like we do with the ctime).

You don't really want to bump it if no one else actually changed it. If we were to just bump it like you suggest, then solo NFS clients may end up with spurious cache invalidations, and that's not a good thing either.

While the logic is somewhat complicated, the saving grace here is that the change_attr can only tell you if _something_ has changed since you last checked it. It's semi-opaque in that you can check to see if it's larger than a previous value, but can't infer anything from the _actual_ value. Thus we can allow a single bump of the change_attr to encompass many changes, as long as it shows as changed when a client queries for it.

ffilz: I'm only thinking of doing an extra bump if multiple Ceph clients have indicated a change and bumped it.

I want to make sure the NFS client sees the right thing in the following scenario:

NFS Client -> Ganesha makes change resulting in change bumping from 4 to 5
NFS client does a GETATTR with change 5
On another Ceph client another change is made also makes a change, also bumping change from 4 to 5
Changes are consolidated on Ceph mds

Now if the mds finalizes everything with change 5, and the NFS client does a new GETATTR and sees change 5, it will assume there was no change, thus missing whatever changes occurred on the other Ceph client, whereas if the mds had bumped the change to 6, the NFS client would then know the file has changed beyond what it had already observed.

gregsfortytwo: Yeah, I think this is necessary since clients might be changing things under different caps. One might have Fw while another makes a synchronous change under As; we need to bump the change_attr both for the (mds-mediated) inode metadata change and for the data write, but with each bumping it independently from x a ceiling will just keep x+1 despite two separate (and separately-visible) changes.
Or perhaps I'm misunderstanding and simple data writes don't require a bump to the change attr, just changes to file size or something.

jtlayton: Hmm...what I think I need to do is sit down and write this up in better detail. I think this is OK, but the reason for that is somewhat subtle. Also, given that this PR is closed and that the new one is now open, it's probably better to continue the discussion there.

PR #10843, opened 2016-08-24
messages/MOSDMap: mark as enlighten OSDMap encoder

liewegas:
Back in 70094407c184baaf963577cc0d98edde12e8e9da we require that anyone
encoding an OSDMap pass in a special feature bit indicating that they are
'enlightened' and understand the rules around encoding OSDMaps (basically,
only mons get to do it).  We forgot to update MOSDMap, which may have to
reencode an OSDMap without some features to talk to a really old client.
We haven't noticed until now because the old set of features we had to do
this for are really old.

We are about to introduce new features (for addr2 encoding) that change
that, and this fix is needed to prevent an assert when doing a reencode.

### DIFF
--- a/src/messages/MOSDMap.h
+++ b/src/messages/MOSDMap.h
@@ -109,9 +109,9 @@ public:
 	  OSDMap m;
 	  m.decode(inc.fullmap);
 	  inc.fullmap.clear();
-	  m.encode(inc.fullmap, features);
+	  m.encode(inc.fullmap, features | CEPH_FEATURE_RESERVED);
 	}
-	inc.encode(p->second, features);
+	inc.encode(p->second, features | CEPH_FEATURE_RESERVED);
       }
       for (map<epoch_t,bufferlist>::iterator p = maps.begin();
 	   p != maps.end();
@@ -119,7 +119,7 @@ public:
 	OSDMap m;
 	m.decode(p->second);
 	p->second.clear();
-	m.encode(p->second, features);
+	m.encode(p->second, features | CEPH_FEATURE_RESERVED);
       }
     }
     ::encode(incremental_maps, payload);


yuriw:
@tchaikov waiting for thumb up and then will retest

jdurgin:
lgtm

tchaikov:
quote from the commit message of 70094407c184baaf963577cc0d98edde12e8e9da, 

> Note that we skip this check for legacy encodings, so we don't touch MOSDMap in this patch.

and in the commit message in this PR

> We forgot to update MOSDMap, which may have to reencode an OSDMap without some features to talk to a really old client. We haven't noticed until now because the old set of features we had to do this for are really old.

so we are basically dropping the support of old clients which do not understand these "really old" features bits? because the `feature` passed in `MOSDMap::encode_payload()` is the intersection of the features of both parties of the connection. so if we are sending `OSDMap`s not understood by the client, client will fail to decode the received `MOSDMap`s. and this renders the re-encoding logic meaningless.

liewegas:
We're not changing any behavior for old clients.  The problem is just that the "old clients" and "really old" features are the branch of MOSDMap that does a reencode, and they all used to be old.. but now we've added a new feature bit that requires a reencode and does do the 'enlightened encoder' assert.

yuriw:
FYI
The following tests FAILED:
	  5 - cephtool-test-mon.sh (Failed)
	 14 - run-tox-ceph-disk (Failed)
	 15 - run-tox-ceph-detect-init (Failed)

liewegas:
The same tests passed when run by jenkins.  I suspect an environment issue on your machine?

yuriw:
tests on master passed on smithi049 tho ?!

tchaikov:
lgtm.

yuriw:
http://pulpito.ceph.com/yuriw-2016-08-29_22:13:15-rados-wip-yuri-testing2_2016_08_29-distro-basic-smithi/

PR #10848, opened 2016-08-24
multimds: enforce allow_multimds during conf load

batrick:
Add a configuration option for allow_multimds which allows setting
max_mds. Previously it was possible to set max_mds in ceph.conf which
would circumvent the allow_multimds protection.

Fixes: http://tracker.ceph.com/issues/17105

### DIFF
--- a/doc/cephfs/mds-config-ref.rst
+++ b/doc/cephfs/mds-config-ref.rst
@@ -10,6 +10,12 @@
 :Type: Boolean
 :Default: ``true`` 
 
+``mon allow multimds``
+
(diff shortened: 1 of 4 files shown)


jcsp:
In general I'm not a fan of these (older) config options that are only respected at the point that the system is set up (or in the case when the filesystem is created).  I think I'd prefer to mark the max_mds setting as deprecated (and eventually remove it), rather than adding another config setting.

gregsfortytwo:
Discussed verbally and I think we decided to just ditch setting an MDS count !=1 on fs creation. @batrick is taking that on, right? :)

batrick:
Yup, I'm writing a patch now.

PR #10849, opened 2016-08-25
rgw: remove a redundant judgement when listng objects.

zhangsw:
The judgement is redundant because '!params.delim.empty()' has been judged just before this judgement. So i merge the two judgement into one.


zhangsw:
Could you have a look at this commit ? Thank you very much ~@yehudasa @cbodley

mattbenjamin:
@zhangsw this change looks logically correct, though the eff. gain is pretty small (1 branch in the have-delimiter case).  I assume this is waiting on automated tests.

PR #7741, opened 2016-02-22
Lookup monitors through DNS, feature #14527

rjfd:
If no monitor host info is found in ceph.conf file or from CLI arguments, the construction of monmap fallbacks to query the DNS for SRV records with "ceph-mon" as the default service name.

The service name can be changed with the "mon_dns_srv_name" config option.

Ticket url: http://tracker.ceph.com/issues/14527

PR #10852, opened 2016-08-25
doc: Add docs about looking up Monitors through DNS

wido:
Add some documentation for MONs being looked up using DNS by librados.

As per http://tracker.ceph.com/issues/14527 / #7741

### DIFF
--- a/doc/rados/configuration/mon-lookup-dns.rst
+++ b/doc/rados/configuration/mon-lookup-dns.rst
@@ -1,0 +1,46 @@
+===============================
+Looking op Monitors through DNS
+===============================
+
+Since version 11.0.0 RADOS supports looking up Monitors through DNS.
+
+This way daemons and clients do not require a *mon host* configuration directive in their ceph.conf configuration file.
+
+Using DNS SRV TCP records clients are able to look up the monitors.
+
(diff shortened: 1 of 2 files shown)


doc/rados/configuration/mon-lookup-dns.rst
```diff
@@ -0,0 +1,42 @@
...
+
+With those records now existing we can create the SRV TCP records with the name *ceph-mon* pointing to the three Monitors.
+
+::
+
+    _ceph-mon._tcp.example.com. 60 IN SRV 10 60 6789 mon1.example.com.
+    _ceph-mon._tcp.example.com. 60 IN SRV 10 60 6789 mon2.example.com.
+    _ceph-mon._tcp.example.com. 60 IN SRV 10 60 6789 mon3.example.com.
+
+In this case the Monitors are running on port *6789*.
```
tchaikov: @wido could you please amend this part with something like:

> The priority and weight are not respected yet. So all monitors resolved using DNS are considered equal, and Ceph client will pick a random one from them when connecting to the cluster without taking these settings into consideration.

wido:
@tchaikov Could you look at this one again?

PR #10853, opened 2016-08-25
doc: Add MON docs about pool flags and pool removal config settings

wido:
Documentation on how to configure pools to disallow removal or set flags on new pools

### DIFF
--- a/doc/rados/configuration/mon-config-ref.rst
+++ b/doc/rados/configuration/mon-config-ref.rst
@@ -771,6 +771,43 @@ Client
 :Default: ``100ul << 20``
 
 
+Pool settings
+=============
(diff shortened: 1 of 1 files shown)


wido:
Can we merge this one in?

doc/rados/configuration/mon-config-ref.rst
```diff
@@ -770,6 +770,41 @@ Client
 :Default: ``100ul << 20``
 
 
+Pool settings
+=============
```
tchaikov: @wido some of these settings are also documented at http://docs.ceph.com/docs/master/rados/operations/pools/#set-pool-values, would be ideal if you could add references to them?

wido:
@tchaikov Done! I fixed the default value since it changed with #11665, but also added a reference to the pool values.

PR #10854, opened 2016-08-25
Add ceph_dirfd in libcephfs and the test for it

JevonQ:
Add ceph_dirfd in libcephfs to enable users to get the file descriptor from the named directory stream.

### DIFF
--- a/src/client/Client.cc
+++ b/src/client/Client.cc
@@ -2766,8 +2766,6 @@ void Client::kick_requests_closed(MetaSession *session)
 }
 
 
-
-
 /************
  * leases
  */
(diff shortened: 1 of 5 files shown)


gregsfortytwo:
What's the purpose of this? It looks more like an attack vector (for somebody who manages to steal a dirp out of memory) and less like something a client program (which opened the dirp, after all) should need.

jcsp:
Presumably this is primarily intended to enable people to fstat directories coherently with opendir'ing them (i.e. in case the path changes between a call to ceph_opendir and a call to ceph_stat?).

The other way around to do this would be to add a ceph_fopendir -- afaict that the way it works in normal POSIX file I/O.

@JevonQ can you say more about the intent and purpose of this?

jcsp:
@JevonQ ping?

JevonQ:
@jcsp You're correct, my intention is to allow people to fstat()/fsync() a

JevonQ:
@jcsp You're correct, my intention is to allow people to fstat/fsync a directory which was opened via opendir. For ceph_fopendir, do you mean the one like fdopendir? Just like you've said, people have to call ceph_open to get a fd, then use ceph_fdopendir with the fd to obtain a directory stream. And after that people can do anything. 

However, I think we can not force people to do it in this way if they want to readdir on a directory and then fstat/fsync the directory. I propose to add this interface because I see the interface dirfd exists.

jcsp:
Sorry for the delay.

If possible I would much rather satisfy this use case by mirroring what normal linux IO calls do: in this case, by implementing fdopendir.  Does that sound reasonable?

JevonQ:
@jcsp Sure, I can think of implementing fdopendir.

jcsp:
Closing this for now, please open a new PR if you do the fdopendir change

PR #10856, opened 2016-08-25
os/bluestore: multithreaded '_kv_sync_thread'

tanghaodong25:

### DIFF
--- a/src/os/bluestore/BlueStore.h
+++ b/src/os/bluestore/BlueStore.h
@@ -971,7 +971,7 @@ public:
     std::mutex wal_apply_mutex;
 
     uint64_t last_seq = 0;
-
+    int id =  0;
     OpSequencer()
 	//set the qlock to PTHREAD_MUTEX_RECURSIVE mode
       : parent(NULL) {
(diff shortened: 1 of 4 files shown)


liewegas:
Did you observe a performance improvement with this?  I'm not sure that mulitple kv commit htreads is the way to go... it seems like it won't help too much given the way that rocksdb blocks it's commits (only one in flight to disk at a time, I think).

Also, I think we want to do something a bit different: ensure that each TransContext is submitted in its own thread, and the kv_sync_thread just loops and pushes them to stable storage.

tanghaodong25:
@liewegas  There is performance improvement just when changing rocksdb to memdb. @majianpeng made more effort on optimizing multi-transaction-submission with a new PR #11189 . So I close it.

src/os/bluestore/BlueStore.h
```diff
@@ -1144,11 +1145,16 @@ class BlueStore : public ObjectStore,
   vector<Finisher*> finishers;
 
   KVSyncThread kv_sync_thread;
+  vector<KVSyncThread *> kv_sync_threads;
   std::mutex kv_lock;
   std::condition_variable kv_cond, kv_sync_cond;
+  vector<std::mutex *> kv_locks;
```
yuyuyu101: vector<std::mutex> is ok

src/os/bluestore/BlueStore.h
```diff
@@ -1144,11 +1145,16 @@ class BlueStore : public ObjectStore,
...
+  vector<std::condition_variable *> kv_conds;
```
yuyuyu101: same

PR #10858, opened 2016-08-25
rgw: remove duplicated calls to getattr 

atheism:
We do not need the duplicated calls to getattr. Remove the unnecessary code.

### DIFF
--- a/src/rgw/rgw_file.cc
+++ b/src/rgw/rgw_file.cc
@@ -1284,12 +1284,6 @@ int rgw_getattr(struct rgw_fs *rgw_fs,
   RGWLibFS *fs = static_cast<RGWLibFS*>(rgw_fs->fs_private);
   RGWFileHandle* rgw_fh = get_rgwfh(fh);
 
-  int rc = -(fs->getattr(rgw_fh, st));
-  if (rc != 0) {
-    // do it again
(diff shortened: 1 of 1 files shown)


src/rgw/rgw_file.cc
```diff
@@ -1290,7 +1290,7 @@ int rgw_getattr(struct rgw_fs *rgw_fs,
     rc = -(fs->getattr(rgw_fh, st));
   }
 
-  return -(fs->getattr(rgw_fh, st));
```
mattbenjamin: I can't remember what I was playing with here, but I think we want just this line (return -(expr)), and to zap the two calls that assign rc.  What do you think?

mattbenjamin:
@atheism you're right, this was some kind of experiement, the calls are idempotent, at least in the current code.  I think maybe we should take the more concise form though.

mattbenjamin:
@atheism I'd tweak the commit msg--the reason to fs->getattr() (once :) is that it's the implementation of the API method.

src/rgw/rgw_file.cc
atheism: I think it would be OK. I will change the pull-request.

cbodley:
jenkins, retest this please

PR #10859, opened 2016-08-25
osd/osd_type: check if pool is gone during check_new_interval() 

xiexingguo:


xiexingguo:
@athanatos Can you look? Thanks!

src/osd/osd_types.cc
```diff
@@ -2209,12 +2209,7 @@ bool pg_stat_t::is_acting_osd(int32_t osd, bool primary) const
   if (primary && osd == acting_primary) {
     return true;
   } else if (!primary) {
-    for(vector<int32_t>::const_iterator it = acting.begin();
-        it != acting.end(); ++it)
-    {
-      if (*it == osd)
-        return true;
-    }
+    return std::count(acting.begin(), acting.end(), osd);
```
athanatos: This one isn't worth the churn.

src/osd/osd_types.cc
```diff
@@ -3123,6 +3123,10 @@ bool pg_interval_t::check_new_interval(
       if (*p != CRUSH_ITEM_NONE)
 	++num_acting;
 
+    if (!lastmap->get_pools().count(pgid.pool())) {
+      i.maybe_went_rw = true;
+      return true;
+    }
```
athanatos: Hmm, the fact that the caller is required to pass in old_up (etc) suggests that assert(lastmap->get_pools().count(pgid.pool()) is a precondition of this function.  There appear to be 3 callers.  Two are variants of build_past_intervals and they get their bounding epochs from _calc_past_interval_range.  _calc_past_interval_range gets its lower bound from last_epoch_clean or epoch_created either of which must be an epoch where the pool exists.  The third is in start_peering_interval which can only be called during an interval in which the pg/pool exists.

In summary, I'm pretty sure that's a precondition for the method, make it an assert.

xiexingguo:
@athanatos Seems jenkins won't be happy if we switch to using `assert`.
So would you like me to drop the following test case in **src/test/osd/types.cc** (see below) or revert to original change instead?
```c++
  //
  // pool did not exist in the old osdmap
  //
  {
    ceph::shared_ptr<OSDMap> lastmap(new OSDMap());
    lastmap->set_max_osd(10);
    lastmap->set_state(osd_id, CEPH_OSD_EXISTS);
    lastmap->set_epoch(epoch);

    map<epoch_t, pg_interval_t> past_intervals;

    ASSERT_TRUE(past_intervals.empty());
    ASSERT_TRUE(pg_interval_t::check_new_interval(old_primary,
						  new_primary,
						  old_acting,
						  new_acting,
						  old_up_primary,
						  new_up_primary,
						  old_up,
						  new_up,
						  same_interval_since,
						  last_epoch_clean,
						  osdmap,
						  lastmap,
						  pgid,
                                                  recoverable.get(),
						  &past_intervals));
    ASSERT_EQ((unsigned int)1, past_intervals.size());
    ASSERT_EQ(same_interval_since, past_intervals[same_interval_since].first);
    ASSERT_EQ(osdmap->get_epoch() - 1, past_intervals[same_interval_since].last);
    ASSERT_EQ(osd_id, past_intervals[same_interval_since].acting[0]);
    ASSERT_EQ(osd_id, past_intervals[same_interval_since].up[0]);
  }

athanatos:
"Seems jenkins won't be happy if we switch to using assert." -- I assume you mean that it's causing the above test to fail.  I haven't looked at it yet, but you are asking the wrong question.  What you should be asking is whether:
1) Sam is wrong and it *is* possible for this condition to be false or
2) This test is testing something that violates the preconditions of the function

What is that test testing?

xiexingguo:
> What is that test testing?

As the comment saying:  this test case verifies that "pool did not exist in the old osdmap" but does exist in the current osdmap, and it expects `check_new_interval` to return true for the testing and fires the `assert `in the end.



> 1) Sam is wrong and it is possible for this condition to be false or

The above testing case says **YES** and **it is possible for this condition to be false**. But in reality I am agree with you, so I dropped that test case instead.

athanatos:
The purpose of a test is to test code paths actually needed by the users of the interface.  If no user actually requires this condition to be false, than it makes far more sense for us to narrow the preconditions and update the tests than it does for us to support the case where it is true.

yuriw:
http://pulpito.ceph.com/yuriw-2016-09-12_21:23:04-rados-wip-yuri-testing_2016_09_12-distro-basic-smithi/
rerun of failed
http://pulpito.front.sepia.ceph.com/yuriw-2016-09-13_15:49:46-rados-wip-yuri-testing_2016_09_12-distro-basic-smithi/

PR #10867, opened 2016-08-25
jewel: rgw： period commit return error when the current period has a zonegroup which doesn't have a master zone

dachary:
http://tracker.ceph.com/issues/17118


dachary:

PR #10873, opened 2016-08-25
rgw: Fix Host->bucket fallback logic inversion

robbat2:
The logic (added in 46aae19ee) for falling back to just using the hostname as
the possible bucket name contained an accidental inversion, because
RGWHandler_REST::validate_bucket_name returns success as zero.

Backport: jewel
Fixes: http://tracker.ceph.com/issues/17136
Re-Fixes: http://tracker.ceph.com/issues/15975

### DIFF
--- a/src/rgw/rgw_rest.cc
+++ b/src/rgw/rgw_rest.cc
@@ -1825,7 +1825,7 @@ int RGWREST::preprocess(struct req_state *s, RGWClientIO* cio)
     if (subdomain.empty()
         && (domain.empty() || domain != info.host)
         && !looks_like_ip_address(info.host.c_str())
-        && RGWHandler_REST::validate_bucket_name(info.host)) {
+        && RGWHandler_REST::validate_bucket_name(info.host) == 0) {
       subdomain.append(info.host);
       in_hosted_domain = 1;
     }


cbodley:
:+1:

robbat2:
Still waiting for this to be merged into master.

cbodley:
@robbat2 sorry, i should have run it through teuthology before merging. i had to revert this commit with https://github.com/ceph/ceph/pull/11102, because it causes us to add a `/` to the end of `request_uri`, which breaks s3 signature verification

we'll need to find a way to support this feature without side effects on authentication

robbat2:
@cbodley can you link the teuth output, ideally the s3tests? I'm wondering why it broke in your environment when it passed s3tests here.

cbodley:
@robbat2 an example here: http://qa-proxy.ceph.com/teuthology/cbodley-2016-09-14_14:00:58-rgw-wip-cbodley-testing---basic-mira/415753/teuthology.log

I was able to reproduce locally by changing the `host = localhost` to a fqdn in my s3tests.conf

cbodley:
@robbat2 okay, turns out that these s3test failures were due to a misconfigured `rgw_dns_name`. i'm guessing that yours matches the host you provided in s3tests.conf. vstart.sh was setting `rgw_dns_name = localhost` for me, and teuthology was using the empty default value (the log showed `RGW hostnames: `). in those cases, all requests were treated as virtual hosted buckets

@yehudasa because `rgw_dns_name` defaults to empty, would you be in favor of disabling the virtual hosting feature in that case? (i.e. if `hostnames_set` in rgw_rest.cc is empty, don't allow `in_hosted_domain` to be set)

robbat2:
If `rgw_dns_name` is empty, but the zonegroup has `hostnames`/`hostnames_s3website` set, then virtual hosting is still valid in my opinion.

yehudasa:
@cbodley @robbat2 I'm mostly in favor of documenting this clearly (whatever we choose).

PR #10875, opened 2016-08-25
Revert "osd/ReplicatedPG: for sync-read it don't cacl l_osd_op_r_prep…

athanatos:
…are_lat."

Accessing the OpContext* after execute_ctx is invalid, on error, it
will have been deleted.

This reverts commit 52be772788d9d96accaa7af9eaf9f29a3792df49.

### DIFF
--- a/src/osd/ReplicatedPG.cc
+++ b/src/osd/ReplicatedPG.cc
@@ -2139,13 +2139,12 @@ void ReplicatedPG::do_op(OpRequestRef& op)
   execute_ctx(ctx);
   utime_t prepare_latency = ceph_clock_now(cct);
   prepare_latency -= op->get_dequeued_time();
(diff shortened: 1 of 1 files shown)


gregsfortytwo:
Adding cephfs since this should fix the valgrind test failures we've been running into.

@athanatos, want to reference http://tracker.ceph.com/issues/16908 in the commit message? Need to get a better fix for whatever perfcounter timing thing the original commit was trying to do?

yuriw:
passed in http://pulpito.ceph.com/yuriw-2016-08-26_22:32:43-rados-wip-yuri-testing_2016_08_26-distro-basic-smithi

@athanatos @gregsfortytwo removing wip-yuri-testing and leaving up to you for merge

PR #10876, opened 2016-08-25
cephfs: Support scrub with multiple MDS

fullerdj:
Support scrubbing with more than one MDS. When a subtree boundary
is encountered during scrub, send a new MMDSScrubPath message to the
authoritative MDS and continue without waiting.

Fixes: http://tracker.ceph.com/issues/12274

### DIFF
--- a/src/mds/ScrubStack.cc
+++ b/src/mds/ScrubStack.cc
@@ -20,6 +20,8 @@
 #include "mds/MDCache.h"
 #include "mds/MDSContinuation.h"
 
+#include "messages/MMDSScrubPath.h"
+
 #define dout_subsys ceph_subsys_mds
 #undef dout_prefix
 #define dout_prefix _prefix(_dout, scrubstack->mdcache->mds)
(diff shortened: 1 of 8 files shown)


ukernel:
I take a quick look. This direction seems correct. But there is complete no freezing/freezon check in the scrub code. This will make the code unstable if there are metadata migration. 

First we can add code that check freezing/frozen directories for the directory fragmentation (single MDS). then add code that check metadata migration

src/mds/ScrubStack.cc
```diff
@@ -299,6 +302,24 @@ void ScrubStack::scrub_dirfrag(CDir *dir,
   *is_terminal = false;
   *done = false;
 
+  if (!dir->is_auth() && dir->is_subtree_root()) {
```
jcsp: Because the assert(is_auth) in enqueue_scrub_work went away, couldn't we end up in here with a dir that we are not auth for, but is also not a subtree root?  In that case we would want to send off a MMDSScrubPath as well.

PR #10877, opened 2016-08-25
jewel: client: nlink count is not maintained correctly

dachary:
http://tracker.ceph.com/issues/16946


dachary:
@jcsp does this backport look good to merge ? It passed the fs suite http://tracker.ceph.com/issues/17487#note-13 (except for one known bug that is, I believe, unrelated)

jcsp:
:+1:

PR #10879, opened 2016-08-26
jerasure: include generic objects in neon jerasure lib (like sse3/4)

dmick:


tchaikov:
lgtm.

PR #7338, opened 2016-01-24
Lua object class support

noahdesu:

PR #10887, opened 2016-08-26
fs: Mantle: A Programmable Metadata Load Balancer

michaelsevilla:
- Design/Motivation presented at [August CDM](http://tracker.ceph.com/projects/ceph/wiki/CDM_03-AUG-2016)
- Testing: https://github.com/ceph/ceph-qa-suite/pull/1145
- Relies on the Lua source used in #7338 (that PR is also pending discussion on ceph-devel)

### DIFF
--- a/doc/cephfs/mantle.rst
+++ b/doc/cephfs/mantle.rst
@@ -51,12 +51,12 @@ Mantle with `vstart.sh`
 
 ::
 
-    ./vstart.sh -n -l
+    cd build
+    ../src/vstart.sh -n -l
     for i in a b c; do 
-      ./ceph --admin-daemon out/mds.$i.asok config set debug_ms 0
-      ./ceph --admin-daemon out/mds.$i.asok config set debug_mds 0
-      ./ceph --admin-daemon out/mds.$i.asok config set debug_mds_balancer 2
-      ./ceph --admin-daemon out/mds.$i.asok config set mds_beacon_grace 1500
+      bin/ceph --admin-daemon out/mds.$i.asok config set debug_ms 0
+      bin/ceph --admin-daemon out/mds.$i.asok config set debug_mds 2
+      bin/ceph --admin-daemon out/mds.$i.asok config set mds_beacon_grace 1500
     done
 
 
@@ -64,23 +64,23 @@ Mantle with `vstart.sh`
 
 ::
 
-    ./rados put --pool=cephfs_metadata_a greedyspill.lua mds/balancers/greedyspill.lua
+    bin/rados put --pool=cephfs_metadata_a greedyspill.lua ../src/mds/balancers/greedyspill.lua
 
 
 3. Activate Mantle:
 
 ::
 
-    ./ceph mds set allow_multimds true --yes-i-really-mean-it
-    ./ceph mds set max_mds 5
-    ./ceph mds set balancer greedyspill.lua
+    bin/ceph mds set allow_multimds true --yes-i-really-mean-it
+    bin/ceph mds set max_mds 5
+    bin/ceph fs set cephfs_a balancer greedyspill.lua
 
 
 4. Mount CephFS in another window:
 
 ::
 
-     ./ceph-fuse /cephfs -o allow_other &
+     bin/ceph-fuse /cephfs -o allow_other &
      tail -f out/mds.a.log
 
 
@@ -89,8 +89,7 @@ Mantle with `vstart.sh`
    last MDS tries to check the load of its neighbor, which does not exist.
 
 5. Run a simple benchmark. In our case, we use the Docker mdtest image to
-   create load. Assuming that CephFS is mounted in the first container, we can
-   share the mount and run 3 clients using: 
+   create load:
 
 ::
 
@@ -213,6 +212,27 @@ jargon: a dummy value is pushed onto the stack and the next iterator replaces
 the top of the stack with a (k, v) pair. After reading each value, pop that
 value but keep the key for the next call to `lua_next`. 
 
+Reading from RADOS
+~~~~~~~~~~~~~~~~~~
+
+All MDSs will read balancing code from RADOS when the balancer version changes
+in the MDS Map. The balancer pulls the Lua code from RADOS synchronously. We do
+this with a timeout: if the asynchronous read does not come back within half
+the balancing tick interval the operation is cancelled and a Connection Timeout
+error is returned. By default, the balancing tick interval is 10 seconds, so
+Mantle will use a 5 second second timeout. This design allows Mantle to
+immediately return an error if anything RADOS-related goes wrong.
+
+We use this implementation because we do not want to do a blocking OSD read
+from inside the global MDS lock. Doing so would bring down the MDS cluster if
+any of the OSDs are not responsive -- this is tested in the ceph-qa-suite by
+setting all OSDs to down/out and making sure the MDS cluster stays active.
+
+One approach would be to asynchronously fire the read when handling the MDS Map
+and fill in the Lua code in the background. We cannot do this because the MDS
+does not support daemon-local fallbacks and the balancer assumes that all MDSs
+come to the same decision at the same time (e.g., importers, exporters, etc.).
+
 Debugging
 ~~~~~~~~
 
@@ -227,6 +247,10 @@ the cls logging interface:
 It is implemented by passing a function that wraps the `dout` logging framework
 (`dout_wrapper`) to Lua with the `lua_register()` primitive. The Lua code is
 actually calling the `dout` function in C++.
+
+Warning and Info messages are centralized using the clog/Beacon. Successful
+messages are only sent on version changes by the first MDS to avoid spamming
+the `ceph -w` utility. These messages are used for the integration tests.
 
 Testing
 ~~~~~~
--- a/doc/cephfs/mantle.rst
+++ b/doc/cephfs/mantle.rst
@@ -1,0 +1,235 @@
+Mantle
+======
+
+Multiple, active MDSs can migrate directories to balance metadata load. The
+policies for when, where, and how much to migrate are hard-coded into the
+metadata balancing module. Mantle is a programmable metadata balancer built
+into the MDS. The idea is to protect the mechanisms for balancing load
+(migration, replication, fragmentation) but stub out the balancing policies
+using Lua. Mantle is based on [1] but the current implementation does *NOT*
+have the following features from that paper:
+
+1. Balancing API: in the paper, the user fills in when, where, how much, and
+   load calculation policies; currently, Mantle only requires that Lua policies
+   return a table of target loads (e.g., how much load to send to each MDS)
+2. "How much" hook: in the paper, there was a hook that let the user control
+   the fragment selector policy; currently, Mantle does not have this hook
+3. Instantaneous CPU utilization as a metric
+
+[1] Supercomputing '15 Paper:
+http://sc15.supercomputing.org/schedule/event_detail-evid=pap168.html
+
+Quickstart with vstart
+----------------------
+
+.. warning::
+
+    Developing balancers with vstart is difficult because running all daemons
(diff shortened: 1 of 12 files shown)


src/mds/MDBalancer.cc
```diff
@@ -36,8 +37,9 @@ using std::map;
 using std::vector;
 
 #include "common/config.h"
+#include "common/errno.h"
 
-#define dout_subsys ceph_subsys_mds
+#define dout_subsys ceph_subsys_mds_balancer
```
jcsp: I'm not sure we want this: iirc the --debug-mds flag isn't a prefixish thing, so this change will silence all the balancer and migrator logs for people who currently set --debug-mds

src/mon/MonCommands.h
```diff
@@ -330,7 +330,7 @@ COMMAND("mds set_max_mds " \
 	"set max MDS index", "mds", "rw", "cli,rest")
 COMMAND("mds set " \
 	"name=var,type=CephChoices,strings=max_mds|max_file_size"
-	"|allow_new_snaps|inline_data|allow_multimds|allow_dirfrags " \
+	"|allow_new_snaps|inline_data|allow_multimds|allow_dirfrags|balancer " \
```
jcsp: Please add this to "fs set" instead of "mds set" -- mds set is the legacy mode that operates only on the default filesystem, fs set is the new hotness which lets you work on any filesystem

src/mds/MDBalancer.cc
```diff
@@ -592,6 +604,53 @@ void MDBalancer::prep_rebalance(int beat)
...
+
+    metrics[i].insert(make_pair("auth.meta_load", load.auth.meta_load()));
+    metrics[i].insert(make_pair("all.meta_load", load.all.meta_load()));
+    metrics[i].insert(make_pair("req_rate", load.req_rate));
+    metrics[i].insert(make_pair("queue_len", load.queue_len));
+    metrics[i].insert(make_pair("cpu_load_avg", load.cpu_load_avg));
+  }
+
+  /* execute the balancer */
+  Mantle *mantle = new Mantle();
```
jcsp: Might as well just do a "Mantle mantle" and avoid the pointer/deletion?

src/mds/MDBalancer.cc
```diff
@@ -592,6 +604,53 @@ void MDBalancer::prep_rebalance(int beat)
...
+  /* fill in the metrics for each mds by grabbing load struct */
+  vector < map<string, double> > metrics (cluster_size);
+  for (mds_rank_t i=mds_rank_t(0);
+       i < mds_rank_t(cluster_size);
+       i++) {
+    map<mds_rank_t, mds_load_t>::value_type val(i, mds_load_t(ceph_clock_now(g_ceph_context)));
+    std::pair < map<mds_rank_t, mds_load_t>::iterator, bool > r(mds_load.insert(val));
+    mds_load_t &load(r.first->second);
+
+    metrics[i].insert(make_pair("auth.meta_load", load.auth.meta_load()));
```
jcsp: Might be tidier to use a C++11 map literal here if possible

src/mds/MDBalancer.cc
```diff
@@ -169,6 +171,28 @@ mds_load_t MDBalancer::get_load(utime_t now)
   return load;
 }
 
+int MDBalancer::localize_balancer(string const balancer)
+{
+  int64_t pool_id = mds->mdsmap->get_metadata_pool();
+  string fname = "/tmp/" + balancer;
```
jcsp: Does it have to go into a /tmp/ file?  I would expect these scripts are usually small enough to painlessly hold in memory after we've read it from rados

src/mds/MDBalancer.cc
```diff
@@ -169,6 +171,28 @@ mds_load_t MDBalancer::get_load(utime_t now)
...
+
+  dout(15) << "looking for balancer=" << balancer << " in RADOS pool_id=" << pool_id << dendl;
+  object_t oid = object_t(balancer);
+  object_locator_t oloc(pool_id);
+  bufferlist data;
+  C_SaferCond waiter;
+  mds->objecter->read(oid, oloc, 0, 0, CEPH_NOSNAP, &data, 0, &waiter);
```
jcsp: Ouch, this is a blocking OSD read that's done from inside the mds_lock when this gets called from mantle_prep_rebalance.

I think it would be better to kick off the read when handling the mdsmap (if the balancer has changed), and populate the script in the background when the read completes.  If mantle_prep_rebalance runs before that read has completed it can either no-op or fall back to the native balancer.

src/mds/MDBalancer.cc
```diff
@@ -276,6 +300,17 @@ void MDBalancer::handle_heartbeat(MHeartbeat *m)
     if (mds_load.size() == cluster_size) {
       // let's go!
       //export_empties();  // no!
+
+      int r = mantle_prep_rebalance();
+      if (!r) {
+        mds->clog->info() << "mantle succeeded; "
```
jcsp: Maybe the success message could just go to the MDS log instead of the cluster log?  This will get spammy for anyone who likes to "ceph -w"

src/mds/MDSMap.cc
```diff
@@ -487,6 +487,7 @@ void MDSMap::encode(bufferlist& bl, uint64_t features) const
     ::encode(session_autoclose, bl);
     ::encode(max_file_size, bl);
     ::encode(max_mds, bl);
+    ::encode(balancer, bl);
```
jcsp: This is the legacy encoding for clients too old to handle versioned things, don't add balancer here

src/mds/MDSMap.cc
```diff
@@ -515,6 +516,7 @@ void MDSMap::encode(bufferlist& bl, uint64_t features) const
...
```
jcsp: Same: legacy encoding, leave it out

src/mds/MDSMap.cc
```diff
@@ -551,6 +553,7 @@ void MDSMap::encode(bufferlist& bl, uint64_t features) const
   ::encode(session_autoclose, bl);
   ::encode(max_file_size, bl);
   ::encode(max_mds, bl);
+  ::encode(balancer, bl);
```
jcsp: When you add an item to an encoded representation, you have to add it at the end (i.e. after "damaged" in this case) -- the way older receivers handle newer messages is to ignore trailing bytes

src/mds/MDSMap.cc
```diff
@@ -542,7 +544,7 @@ void MDSMap::encode(bufferlist& bl, uint64_t features) const
     return;
   }
 
-  ENCODE_START(5, 4, bl);
```
jcsp: You need to bump the "ev" below from 10 to 11 instead of modifying this version (the mdsmap encoding is weird!)

src/mds/MDSMap.cc
```diff
@@ -583,7 +586,7 @@ void MDSMap::decode(bufferlist::iterator& p)
   std::map<mds_rank_t,int32_t> inc;  // Legacy field, parse and drop
 
   cached_up_features = 0;
-  DECODE_START_LEGACY_COMPAT_LEN_16(5, 4, 4, p);
+  DECODE_START_LEGACY_COMPAT_LEN_16(6, 4, 4, p);
```
jcsp: Leave this alone, and when you read out 'banalcer' at the very end make it conditional on ev>=11

src/mds/MDSMap.h
```diff
@@ -289,6 +290,9 @@ class MDSMap {
   mds_rank_t get_max_mds() const { return max_mds; }
   void set_max_mds(mds_rank_t m) { max_mds = m; }
 
+  std::string get_balancer() const { return balancer; }
```
jcsp: Could return a const reference here

src/mds/Makefile-client.am
```diff
@@ -1,4 +1,5 @@
 # There are no libmds_types so use the full mds library for dencoder for now
 DENCODER_SOURCES += $(LIBMDS_SOURCES)
+ceph_dencoder_CPPFLAGS = $(LIBLUA_CPPFLAGS)
```
jcsp: autotools is now gone in master so this part can just be removed

src/mds/MDSMap.h
```diff
@@ -193,6 +193,7 @@ class MDSMap {
    */
 
   mds_rank_t max_mds; /* The maximum number of active MDSes. Also, the maximum rank. */
+  string balancer; /* The name and version of the metadata load balancer. */
```
jcsp: I guess it's the name and version if someone named their objects that way, but really this is just a rados object name, right?

jcsp: Please add this field to MDSMap::dump (both plain text and json)

src/mds/Mantle.h
```diff
@@ -0,0 +1,34 @@
+#include <lua.hpp>
+#include <list>
+#include <map>
+using std::list;
```
jcsp: Please avoid `using` directives in headers (we do already have some but I'd rather not add more)

src/mds/Mantle.h
```diff
@@ -0,0 +1,34 @@
...
+using std::map;
+
+#include "include/types.h"
+#include "common/Clock.h"
+#include "CInode.h"
+
+
+
+class MDSRank;
```
jcsp: This block of `class` statements looks to be unneeded

src/mds/Mantle.h
```diff
@@ -0,0 +1,34 @@
+#include <lua.hpp>
```
jcsp: please add #ifdef header guards and the usual modelines+copyright header (copy paste and put your name in).  Same in Mantle.cc

doc/cephfs/mantle.rst
```diff
@@ -0,0 +1,235 @@
+Mantle
```
jcsp: Wow, I wish every PR came with this much documentation!

jcsp:
This looks to be exactly what was discussed at the last CDM, so all of my comments are just detail things.  At a high level: :+1: 

Does #7338 need poking along somehow?

michaelsevilla:
I already poked #7338 and @noahdesu is going to take care of it.

Dumb question: is it better to rebase (keep history clean but lose your comments) or just add more commits?

jcsp:
You can either add more commits (and rebase later before merge) or just do a straight history-squashing rebase.

Usually we leave comments on the "files changed" view (not individual commits) so that they survive a rebase, but I mistakenly put some on commits here -- my fault if they get lost in a rebase.

src/mds/MDBalancer.cc
gregsfortytwo: I don't think we can do daemon-local fallbacks like that. The default balancer assumes everybody else will come to basically the same conclusions as the local instance (about which daemons are paired for export/import, etc).
I haven't looked at the code but I presume we'll need to error out from rebalancing and put big fat warnings in the clog/Beacon if anything goes wrong.

gregsfortytwo:
Luckily Github finally fixed that nonsense, at least as long as you navigate from the PR view into the commit (not sure if you just go straight to a commit). You can tell it's working because all the comments are on the diff, not on a commit. ;)

Personally I prefer commits prefixed with "SQUASH" that are intended to be rebase, as it makes life easier on the reviewer when there are follow-up comments, but that may just be me.

gregsfortytwo:
I see we did get some updated commits at some point that at least invalidated most comments. @jcsp, want to check this again? Be nice to get it in for @batrick's scaling tests.

jcsp:
This is good to go now from my point of view (apologies @michaelsevilla for delay).  Please could you do a quick rebase as something else has touched CMakeLists.txt in the meantime

michaelsevilla:
Done -- @batrick: could you send me info on the scaling tests?

batrick:
See: http://marc.info/?l=ceph-devel&m=147494693010640&w=2

PR #10889, opened 2016-08-26
jewel: rgw: rgw file uses too much CPU in gc/idle thread

dachary:
http://tracker.ceph.com/issues/17143


dachary:

PR #10891, opened 2016-08-26
jewel: rgw: upgrade from old multisite to new multisite fails

dachary:
http://tracker.ceph.com/issues/16793


dachary:

PR #10893, opened 2016-08-26
rados: check for negative return value of rados_create_with_context() as its comment put

yonghengdexin735:
rados.init_with_context returned value is '<=0'

Signed-off-by: zhang.zezhu zhang.zezhu@zte.com.cn


tchaikov:
`librados::Rados::init_with_context()` calls `rados_create_with_context()`, which returns 0 on success.

dachary:
@yonghengdexin735 please explain in the commit message why it is a bug to fail when the return value is a positive integer

PR #10896, opened 2016-08-26
[DNM] rbd-mirror: Safe handling of interrupted image bootstrap

vshankar:
This patchset handles interrupted bootstrap of an image safely by performing a `checkpoint` operation before creating the local image. Checkpointing is done by setting an omap key/value pair  (key formed by using the `global image id` and value being the `image id`) on the `RBD_MIRRORING` object which is removed after the local image has been successfully created. Presence of this omap key/value hints bootstrap of a possible interruption during last run (either due to crash or other factors). Recovery is performed by cleaning up the partial image (identified by the `image id` stored in the omap value) and recreating it freshly (again after performing a checkpoint).

This fix builds over a series of patches that introduce asynchronous image deletion.


vshankar:
Posted this patch for early reviews and doubts in certain parts of implementation. Unit test cases would probably fail -- working on resolving those.

dillaman:
@vshankar don't forget that cloning an image would need to support the same mechanism of providing a suggested image id.

vshankar:
The unittest failure is due to a lockdep assertion. Image removal state machine acquires a read lock  (on `owner_lock`) post image open which is executed in librbd thread context (`tp_librbd`). Later, `TrimRequest::should_complete()` tries to acquire read lock on `owner_lock` resulting in lockdep identifying the recursive lock (from the same thread) and asserting. The synchronous version of `librbd::remove()` acquires `owner_lock` in the caller thread (which executes `rbd` cli command) and therefore doesn't have this issue.

This brings me to back to the some of the implementation questions I had which I kind of delayed in asking, but looks like those questions may converge with what's required to workaround the lockdep assertion.

Currently ImageCtx is initiated within the image remove async state machine. Instead if sync callers can pass in an ImageCtx instance, post open and after acquiring relevant locks, that would be lockdep safe. But again, the caller may be from a different async state machine (such as `BootstrapRequest`) which would result in the same lockdep assertion. My best bet here would be to create context callback via `create_async_context_callback` which invokes callback in a different thread (`ImageCtx::op_work_queue`) and therefore safe for acquiring locks mentioned above. Any other (better) alternatives?

Also, I would like to keep `ImageCtx` instantiation in the async state machine (`RemoveRequest`) as it keeps `BootstrapRequest` simple when cleaning up partial image. Any gotchas in that (?) -- looks good to me if it's cleaned up properly. Or is it a good practice to pass it to the state machine?

@dillaman -- lots of details/questions above, in your spare time.

src/librbd/image/RemoveRequest.cc
```diff
@@ -0,0 +1,905 @@
...
+void RemoveRequest<I>::acquire_exclusive_lock() {
+  ldout(m_cct, 20) << this << " " << __func__ << dendl;
+
+  using klass = RemoveRequest<I>;
+  if (m_force) {
+    Context *ctx =
+      create_context_callback<klass, &klass::handle_exclusive_lock_force>(this);
+    // releasing read lock to avoid a deadlock when upgrading to
+    // write lock in the shut_down process
+    m_image_ctx->owner_lock.put_read();
```
dillaman: Not required since this is an async state machine, not a sync state machine -- don't hold the owner lock when the async state machine isn't running

src/librbd/image/RemoveRequest.cc
```diff
@@ -0,0 +1,905 @@
...
+    m_image_ctx->exclusive_lock->shut_down(ctx);
+  } else {
+    Context *ctx = create_context_callback<klass, &klass::handle_exclusive_lock>(this);
+    m_image_ctx->operations->prepare_image_update(ctx);
```
dillaman: You can just directly request the lock -- no need to use this sync helper function

dillaman:
@vshankar The first problem is how to handle the deletion of an ImageCtx that is owned and managed by RemoveRequest. You cannot delete an ImageCtx while your current calling context is being executed from said ImageCtx's op_work_queue. Perhaps the simple solution to this, since we keep hitting it, is to make the ```op_work_queue``` a singleton like its parent ```ThreadPool```. 

In terms of this particular issue, the ```create_async_context_callback``` helper is used to solve these locking issues throughout the code and could be re-used here. Since this is a maintenance operation, I am not too worried about the extra context switches.

vshankar:
> @vshankar The first problem is how to handle the deletion of an ImageCtx that is owned and managed by RemoveRequest. You cannot delete an ImageCtx while your current calling context is being executed from said ImageCtx's op_work_queue. Perhaps the simple solution to this, since we keep hitting it, is to make the op_work_queue a singleton like its parent ThreadPool.

@dillaman Yeh, I got the part of destroying ImageCtx using it's work queue. If we pass in a separate work queue (if the caller has one already passed to it or instantiate a fresh WQ whose parent is the singleton `ThreadPool`) and use that to destroy `ImageCtx`, that should be safe, correct? The overhead here is maintenance of `WorkQueue` by every caller of the state machine.

> 
> In terms of this particular issue, the create_async_context_callback helper is used to solve these locking issues throughout the code and could be re-used here. Since this is a maintenance operation, I am not too worried about the extra context switches.

dillaman:
@vshankar Yes, passing in an alternate WorkQueue is safe -- just need to switch to that WorkQueue's context before deleting the ImageCtx.

src/tools/rbd_mirror/image_replayer/BootstrapRequest.cc
```diff
@@ -375,29 +376,173 @@ void BootstrapRequest<I>::handle_open_local_image(int r) {
   update_client_image();
 }
 
-template <typename I>
+template<typename I>
 void BootstrapRequest<I>::create_local_image() {
   dout(20) << dendl;
 
-  m_local_image_id = "";
   update_progress("CREATE_LOCAL_IMAGE");
```
jazeltq: Should we put the update_progress in create_image?

jazeltq: Should we put the update_progress in create_image?

vshankar: The global marker set and remove is kind of an internal mechanism to handle interrupted image creation/cloning. Therefore, progress is updated with "CREATE" tag within here.

vshankar:
@ceph-jenkins retest this please

src/cls/rbd/cls_rbd.cc
```diff
@@ -134,6 +134,9 @@ cls_method_handle_t h_mirror_image_status_get;
 cls_method_handle_t h_mirror_image_status_list;
 cls_method_handle_t h_mirror_image_status_get_summary;
 cls_method_handle_t h_mirror_image_status_remove_down;
+cls_method_handle_t h_mirror_image_create_marker_get;
```
dillaman: I don't think new cls methods are necessary -- we just need to adjust the rbd::mirror::image_replayer::BootstrapRequest state machine and add a new state to librbd::journal::MirrorPeerState to indicate the image is being "created".

src/librbd/internal.cc
```diff
@@ -1176,13 +1141,24 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force,
   }
 
   int create(IoCtx& io_ctx, const char *imgname, uint64_t size,
-	     ImageOptions& opts,
+             ImageOptions& opts,
+             const std::string &non_primary_global_image_id,
+             const std::string &primary_mirror_uuid) {
+    std::string image_id = util::generate_image_id(io_ctx);
```
dillaman: Minor: can we just remove this version of ```create``` and have its current caller invoke the new version that takes the generated image id?

src/librbd/image/RemoveRequest.h
```diff
@@ -0,0 +1,161 @@
...
+
+  void remove_client_snaps();
+
+  void remove_snaps();
+  Context *handle_remove_snaps(int *result);
+
+  void unregister_client();
+  Context *handle_unregister_client(int *result);
+
+  void send_close_image(int r, bool destroy_only=false);
```
dillaman: Minor: remove the optional boolean and instead invoke ```switch_thread_context``` directly upon open failure

src/librbd/image/RemoveRequest.cc
```diff
@@ -0,0 +1,910 @@
...
+    fetch_mirror_image();
+  } else {
+    send_close_image(0);
+  }
+
+  return nullptr;
+}
+
+template<typename I>
+void RemoveRequest<I>::fetch_mirror_image() {
```
dillaman: This is a bug with the current remove implementation -- we should attempt to disable mirroring first before we change the image.  I think we should re-use Mykola's work for async feature updates to first disable the journaling feature since that will ensure that mirroring is properly disabled. That way we can avoid the need to implement this mirror disable logic twice.

src/librbd/image/RemoveRequest.cc
```diff
@@ -0,0 +1,910 @@
...
+
+template<typename I>
+void RemoveRequest<I>::remove_v2_image() {
+  ldout(m_cct, 20) << dendl;
+
+  remove_journal();
+}
+
+template<typename I>
+void RemoveRequest<I>::remove_journal() {
```
dillaman: If you disable journaling and object map at the start -- we can also eliminate the duplicate code for cleanup here and just let the update features state machine do it all for us.

src/librbd/image/RemoveRequest.cc
```diff
@@ -0,0 +1,910 @@
...
+  m_image_ctx->state->open(ctx);
+}
+
+template<typename I>
+Context *RemoveRequest<I>::handle_open_image(int *result) {
+  ldout(m_cct, 20) << ": r=" << *result << dendl;
+
+  if (*result < 0) {
+    lderr(m_cct) << "error opening image: " << cpp_strerror(*result) << dendl;
+    send_close_image(*result, true);
```
dillaman: In the case of a partial removal, it's quite possible that the image cannot be opened but the remaining items should be removed. Therefore, only in this special case should the removal proceed with removing the header/id/etc.

src/librbd/internal.cc
vshankar: Sure. Kept it as I thought it might be useful for callers to just call the old version.

src/cls/rbd/cls_rbd.cc
vshankar: Got it. That would mean `image::CreateRequest` need not set mirror image key(s) for non-primary -- all that's done in `BootstrapRequest`

dillaman: Yeah, I think that's correct.  It would still need to configure the journal to be "owned" by the remote peer, though.

dillaman:
@vshankar You can use 'mirror_image_get_image_id' to translate a global image id to a local image id within the RBD_MIRRORING directory.

vshankar:
> @vshankar You can use 'mirror_image_get_image_id' to translate a global image id to a local image id within the RBD_MIRRORING directory.

Yeh, my bad. I had made an incorrect assumption (don't know why, it's crystal clear in `mirror::image_set()`) that both maps are *not* created at the same time (esp. the global -> local mapping). That lead me into thinking about finding out `image id` from a list of `local -> global` mappings. Silly me.

Thanks @dillaman

vshankar:
Changed in this patchset:

* reused MirrorImage for checkpoint/restore
* continue image removal upon ->open() failure
* removed create/clone API that does not accept `imageid` as argument

Some work needs to be done to make use of changed by @trociny, hence referencing this PR to #10980

vshankar:
@ceph-jenkins retest this please

src/tools/rbd_mirror/image_replayer/CreateLocalImageRequest.cc
```diff
@@ -0,0 +1,311 @@
...
+  dout(20) << dendl;
+
+  using klass = CreateLocalImageRequest<I>;
+  Context *ctx = create_context_callback
+    <klass, &klass::handle_remove_local_image_done>(this);
+
+  librbd::image::RemoveRequest<I> *req = librbd::image::RemoveRequest<I>::create(
+    m_local_io_ctx, m_local_image_name, m_curr_image_id, true, m_no_op_prog_ctx,
+    m_op_work_queue, ctx);
+  req->send();
```
yangdongsheng: @vshankar it could be a stupid question, but what's the benifits of an async-ed removing here? I found the most of our operations are being re-implemented as a async-style, I am sure there should be some reasons.  could you explain the advantage of it compared with the blocking way. thanx a lot.

vshankar: @yangdongsheng Calling blocking operations from an async state machine can lead to deadlock. The librbd singleton thread needs to be "free" for the async state machine to make progress. If the singleton thread invokes a blocking operation, which in turn happens to make one or more async calls internally, there is noone to invoke the callback. So, either operations have to be implemented as async state machines or blocking operations should be invoked via a separate `ContextWQ`.

dillaman:
Note: this is dependent upon the async image removal and async clone operation PRs being merged

dillaman:
@vshankar iSCSI is higher priority, but this PR should **finally** be unblocked.

vshankar:
> @vshankar iSCSI is higher priority, but this PR should finally be unblocked.

sure, thanks.

dillaman:
See "rbd-mirror: pre-register image id before creating image" in https://github.com/ceph/ceph/pull/16642

PR #10907, opened 2016-08-29
mon: have mon-specific features & rework internal monmap structures

jecluis:
This pull request is pretty much the result of PR #8126 with a few additions on tests, a somewhat useful admin socket set of commands to change the (in-memory) monmap features on-the-fly, and monmaptool support.

It has been rebased on master, although I suspect it may need yet another rebase before merging. This has been manually tested to make sure features work with pre-kraken monitors and with kraken itself. In the mean time it will go through a rados suite.

Reviews are welcome.

### DIFF
--- a/src/krbd.cc
+++ b/src/krbd.cc
@@ -118,12 +118,14 @@ static int build_map_buf(CephContext *cct, const char *pool, const char *image,
   if (r < 0)
     return r;
 
-  for (map<string, entity_addr_t>::const_iterator it = monmap.mon_addr.begin();
-       it != monmap.mon_addr.end();
-       ++it) {
-    if (it != monmap.mon_addr.begin())
+  list<entity_addr_t> mon_addr;
+  monmap.list_addrs(mon_addr);
+
+  for (const auto &p : mon_addr) {
+    if (oss.tellp() > 0) {
       oss << ",";
-    oss << it->second.get_sockaddr();
+    }
+    oss << p;
   }
 
   oss << " name=" << cct->_conf->name.get_id();
--- a/src/mon/Elector.cc
+++ b/src/mon/Elector.cc
@@ -88,13 +88,16 @@ void Elector::start()
   }
   start_stamp = ceph_clock_now(g_ceph_context);
   electing_me = true;
-  acked_me[mon->rank] = CEPH_FEATURES_ALL;
+  acked_me[mon->rank].cluster_features = CEPH_FEATURES_ALL;
+  acked_me[mon->rank].mon_features = ceph::features::mon::get_supported();
   leader_acked = -1;
 
   // bcast to everyone else
   for (unsigned i=0; i<mon->monmap->size(); ++i) {
     if ((int)i == mon->rank) continue;
-    Message *m = new MMonElection(MMonElection::OP_PROPOSE, epoch, mon->monmap);
+    MMonElection *m =
+      new MMonElection(MMonElection::OP_PROPOSE, epoch, mon->monmap);
+    m->mon_features = ceph::features::mon::get_supported();
     mon->messenger->send_message(m, mon->monmap->get_inst(i));
   }
   
@@ -116,6 +119,7 @@ void Elector::defer(int who)
   leader_acked = who;
   ack_stamp = ceph_clock_now(g_ceph_context);
   MMonElection *m = new MMonElection(MMonElection::OP_ACK, epoch, mon->monmap);
+  m->mon_features = ceph::features::mon::get_supported();
   m->sharing_bl = mon->get_supported_commands_bl();
   mon->messenger->send_message(m, mon->monmap->get_inst(who));
   
@@ -187,12 +191,15 @@ void Elector::victory()
   leader_acked = -1;
   electing_me = false;
 
-  uint64_t features = CEPH_FEATURES_ALL;
+  uint64_t cluster_features = CEPH_FEATURES_ALL;
+  mon_feature_t mon_features = ceph::features::mon::get_supported();
   set<int> quorum;
-  for (map<int, uint64_t>::iterator p = acked_me.begin(); p != acked_me.end();
+  for (map<int, elector_features_t>::iterator p = acked_me.begin();
+       p != acked_me.end();
        ++p) {
     quorum.insert(p->first);
-    features &= p->second;
+    cluster_features &= p->second.cluster_features;
+    mon_features &= p->second.mon_features;
   }
 
   // decide what command set we're supporting
@@ -224,13 +231,16 @@ void Elector::victory()
     if (*p == mon->rank) continue;
     MMonElection *m = new MMonElection(MMonElection::OP_VICTORY, epoch, mon->monmap);
     m->quorum = quorum;
-    m->quorum_features = features;
+    m->quorum_features = cluster_features;
+    m->mon_features = mon_features;
     m->sharing_bl = *cmds_bl;
     mon->messenger->send_message(m, mon->monmap->get_inst(*p));
   }
     
   // tell monitor
-  mon->win_election(epoch, quorum, features, cmds, cmdsize, &copy_classic_mons);
+  mon->win_election(epoch, quorum,
+                    cluster_features, mon_features,
+                    cmds, cmdsize, &copy_classic_mons);
 }
 
 
@@ -315,13 +325,24 @@ void Elector::handle_ack(MonOpRequestRef op)
 	    << " without required features" << dendl;
     return;
   }
-  
+
   if (electing_me) {
     // thanks
-    acked_me[from] = m->get_connection()->get_features();
+    acked_me[from].cluster_features = m->get_connection()->get_features();
+    acked_me[from].mon_features = m->mon_features;
     if (!m->sharing_bl.length())
       classic_mons.insert(from);
-    dout(5) << " so far i have " << acked_me << dendl;
+    dout(5) << " so far i have {";
+    for (map<int, elector_features_t>::const_iterator p = acked_me.begin();
+         p != acked_me.end();
+         ++p) {
+      if (p != acked_me.begin())
+        *_dout << ",";
+      *_dout << " mon." << p->first << ":"
+             << " features " << p->second.cluster_features
+             << " " << p->second.mon_features;
+    }
+    *_dout << " }" << dendl;
     
     // is that _everyone_?
     if (acked_me.size() == mon->monmap->size()) {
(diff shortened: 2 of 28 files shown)


src/mon/Monitor.cc
```diff
@@ -1887,11 +1896,7 @@ void Monitor::win_election(epoch_t epoch, set<int>& active, uint64_t features,
   // when monitors are call elections or participating in a paxos
   // round without agreeing on who the participants are.
   monmon()->election_finished();
-  for (vector<PaxosService*>::iterator p = paxos_service.begin();
-       p != paxos_service.end(); ++p) {
-    if (*p != monmon())
-      (*p)->election_finished();
-  }
+  _finish_svc_election();
```
tchaikov: so `monmon()->election_finished()` would be called twice? it will be fine i guess. as the second call will be basically a no-op.

src/mon/Elector.cc
```diff
@@ -294,13 +304,24 @@ void Elector::handle_ack(MonOpRequestRef op)
...
   if (electing_me) {
     // thanks
-    acked_me[from] = m->get_connection()->get_features();
+    acked_me[from].cluster_features = m->get_connection()->get_features();
+    acked_me[from].mon_features = m->mon_features;
     if (!m->sharing_bl.length())
       classic_mons.insert(from);
-    dout(5) << " so far i have " << acked_me << dendl;
+    dout(5) << " so far i have {";
+    for (map<int, elector_features_t>::const_iterator p = acked_me.begin();
```
tchaikov: could use range-based for loop.

src/test/Makefile-server.am
```diff
@@ -192,6 +192,11 @@ unittest_mon_pgmap_LDADD = $(LIBMON) $(UNITTEST_LDADD) $(CEPH_GLOBAL)
 unittest_mon_pgmap_CXXFLAGS = $(UNITTEST_CXXFLAGS)
 check_TESTPROGRAMS += unittest_mon_pgmap
 
+unittest_mon_montypes_SOURCES = test/mon/test_mon_types.cc
+unittest_mon_montypes_LDADD = $(LIBMON) $(UNITTEST_LDADD) $(CEPH_GLOBAL)
+unittest_mon_montypes_CXXFLAGS = $(UNITTEST_CXXFLAGS)
+check_TESTPROGRAMS += unittest_mon_montypes
+
```
tchaikov: might want to remove this change to resolve the merge conflict.

src/krbd.cc
```diff
@@ -118,12 +118,15 @@ static int build_map_buf(CephContext *cct, const char *pool, const char *image,
   if (r < 0)
     return r;
 
-  for (map<string, entity_addr_t>::const_iterator it = monmap.mon_addr.begin();
-       it != monmap.mon_addr.end();
+  list<entity_addr_t> mon_addr;
+  monmap.list_addrs(mon_addr);
+
+  for (list<entity_addr_t>::const_iterator it = mon_addr.begin();
```
tchaikov: could use range-based for loop.

src/mon/Monitor.cc
```diff
@@ -1973,6 +1975,21 @@ void Monitor::finish_election()
...
 }
 
+void Monitor::_apply_compatset_features(CompatSet &new_features)
+{
+  if (new_features.compare(features) != 0) {
+    CompatSet diff = features.unsupported(new_features);
+    dout(1) << __func__ << " enabling new quorum features: " << diff << dendl;
+    features = new_features;
+
+    MonitorDBStore::TransactionRef t(new MonitorDBStore::Transaction);
```
tchaikov: better off using `std::make_shared<>`.

src/mon/Monitor.h
```diff
@@ -613,7 +613,9 @@ class Monitor : public Dispatcher,
   mon_feature_t get_required_mon_features() const {
     return monmap->get_required_features();
   }
+  void _apply_compatset_features(CompatSet &new_features);
```
tchaikov: this method can be `private` i think.

src/mon/MonMap.cc
```diff
@@ -134,6 +134,30 @@ void MonMap::generate_test_instances(list<MonMap*>& o)
...
+
+    const char *end_p = local_pub_addr_s + strlen(local_pub_addr_s);
+    entity_addr_t local_pub_addr;
+    local_pub_addr.parse(local_pub_addr_s, &end_p);
+
+    m->add("filled_pub_addr", local_pub_addr);
+
+    m->add("empty_addr_zero", entity_addr_t());
+  }
+  o.push_back(m);
```
tchaikov: @jeluis, just want to understand how this MonMap is related to this new feature? or its purpose is just to test monmap with multiple monitors?

src/mon/MonMap.cc
```diff
@@ -40,6 +40,96 @@ void mon_info_t::print(ostream& out) const
...
+      assert(mon_info[p->first].public_addr == p->second);
+    } else {
+      mon_info_t &m = mon_info[p->first];
+      m.name = p->first;
+      m.public_addr = p->second;
+    }
+  }
+}
+
+struct rank_cmp {
```
tchaikov: nit, could move this helper class into an anonymous namespace.

src/mon/MonMap.cc
```diff
@@ -40,6 +40,96 @@ void mon_info_t::print(ostream& out) const
...
+  // 'o' will be our only source of info about the monitors in the
+  // cluster -- and we will use it to populate our 'mon_info' map.
+
+  bool has_mon_info = false;
+  if (mon_info.size() > 0) {
+    assert(o.size() == mon_info.size());
+    has_mon_info = true;
+  }
+
+  for (map<string, entity_addr_t>::const_iterator p = o.begin();
```
tchaikov: could take this change to switch to range-based for loop.

src/mon/MonMap.h
```diff
@@ -105,88 +105,9 @@ class MonMap {
...
-    // map the set to the actual ranks etc
-    unsigned i = 0;
-    for (set<mon_info_t>::iterator p = tmp.begin();
-         p != tmp.end();
-         ++p, ++i) {
-      ranks[i] = p->name;
-    }
-  }
+public:
+  void sanitize_mons(map<string,entity_addr_t>& o);
```
tchaikov: could be private.

src/mon/MonMap.h
```diff
@@ -105,88 +105,9 @@ class MonMap {
...
+  void calc_ranks();
```
tchaikov: ditto.

src/mon/mon_types.h
```diff
@@ -451,6 +483,17 @@ static inline const char *ceph::features::mon::get_feature_name(uint64_t b) {
   return "unknown";
 }
 
+static inline
+mon_feature_t ceph::features::mon::get_feature_by_name(std::string n) {
```
tchaikov: nit, could pass by a const reference.

src/test/mon/test_mon_types.cc
```diff
@@ -112,3 +112,30 @@ TEST(mon_features, set_funcs) {
...
+
+  mon_feature_t FEATURE_A((1ULL << 1));
+  mon_feature_t FEATURE_B((1ULL << 2));
+  mon_feature_t FEATURE_C((1ULL << 3));
+
+  mon_feature_t foo;
+  ASSERT_EQ(foo, ceph::features::mon::FEATURE_NONE);
+
+  foo.set_feature(FEATURE_A);
+  ASSERT_EQ(foo, FEATURE_A);
```
tchaikov: nit, better of putting
```c++
 ASSERT_EQ(FEATURE_A, foo);
```

src/tools/monmaptool.cc
```diff
@@ -24,10 +24,146 @@ using namespace std;
 
 void usage()
 {
-  cout << " usage: [--print] [--create [--clobber][--fsid uuid]] [--generate] [--set-initial-members] [--add name 1.2.3.4:567] [--rm name] <mapfilename>" << std::endl;
+  cout << " usage: [--print] [--create [--clobber][--fsid uuid]]\n"
+       << "        [--generate] [--set-initial-members]\n"
+       << "        [--add name 1.2.3.4:567] [--rm name]\n"
+       << "        [--feature-list [plain|parseable]]\n"
```
tchaikov: why not just support formatter argument? so the output can be parsed by jq or xmlstarlet?

tchaikov:
lgtm. other than some nits.

tchaikov:
@jecluis ping?

liewegas:
@jecluis and rebase?

src/mon/MonMap.cc
jecluis: @tchaikov serves the dual purpose to further test the monmap api (as there were changes in previous patches), as well as further extend the coverage of tests. True purpose: peace of mind.

jecluis:
okay, fixed suggestions, rebased, and will be pushing in a bit.

src/mon/Monitor.cc
jecluis: actually, i'd rather have it only being called once. I'll change this and move the required logic into `_finish_svc_election()`.

jecluis:
okay, looks like the rebase was old enough to cause a conflict... -_-

jecluis:
Okay, I can't for the life of me reproduce the bloody unit test failures. Jenkins keeps on failing on unittest_erasure_code_plugin_jerasure with a segfault, but locally all the tests complete. Even presumed it could need a rebase, but not even that - no matter how close to master I am, it just fails.

Anyone has any ideas of what may be going on?

liewegas:
@jecluis pretty sure it's a problem with jenkins.. it's happening to prs all over, not just you.  you can ignore it for now.

src/krbd.cc
```diff
@@ -118,12 +118,14 @@ static int build_map_buf(CephContext *cct, const char *pool, const char *image,
...
     return r;
 
-  for (map<string, entity_addr_t>::const_iterator it = monmap.mon_addr.begin();
-       it != monmap.mon_addr.end();
-       ++it) {
-    if (it != monmap.mon_addr.begin())
+  list<entity_addr_t> mon_addr;
+  monmap.list_addrs(mon_addr);
+
+  for (auto p : mon_addr) {
```
tchaikov: could be `const auto& p`

tchaikov:
lgtm

voidbag:
@jecluis 
I have problem with doing command 'rbd map foo'.
**"In some cases useful info is found in syslog - try "dmesg | tail" or so.
rbd: map failed: (22) Invalid argument"**

The result of 'dmesg | tail':
**libceph: parse_ips bad ip ~~~**

In the commit(aa5b8e7e4c561de841f3e613984364d7b11e9378) just before this patch,
I can execute the command 'rbd map foo', successfully
After this commit, I cannot map my rbd image.

Could you explain something about this problem?

Thank you

PR #10908, opened 2016-08-29
os/bluestore: Improve 4K overwrite performance by eliminating DB update and doing direct block write when possible

ifed01:
If csum is off and single 4K chunk overwrite takes place one can safely bypass onode metadata update.
This gives ~10-15% performance boost for rand R/W fio benchmarks. (Note: min_alloc_size set to 64K)

Additionally given the assumption that single device block overwrite is atomic one can eliminate WAL  (and corresponding DB update) totally and do the write directly over an old block.

Both fixes gives >100% performance gain for my benchmarks.

The next step using the same concept is to omit new blob creation on 'big' blob overwrite if min_alloc_size == block_size. No need for onode metadata update in this case too.

Signed-off-by: Igor Fedotov ifedotov@mirantis.com

### DIFF
--- a/src/os/bluestore/BlueStore.cc
+++ b/src/os/bluestore/BlueStore.cc
@@ -5278,14 +5278,7 @@ int BlueStore::_wal_apply(TransContext *txc)
     dout(20) << __func__ << " finished sleep" << dendl;
   }
 
-  assert(txc->ioc.pending_aios.empty());
-  vector<OnodeRef>::iterator q = txc->wal_op_onodes.begin();
-  for (list<bluestore_wal_op_t>::iterator p = wt.ops.begin();
-       p != wt.ops.end();
-       ++p, ++q) {
-    int r = _do_wal_op(txc, *p);
-    assert(r == 0);
-  }
+  _do_wal_ops(txc, true);
 
   _txc_state_proc(txc);
   return 0;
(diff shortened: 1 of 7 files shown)


jhyoon01:
Regarding eliminating WAL, is it safe in the point of power failure recovery?

Even if storage device provides atomic block write, without WAL it seems like transactional update of both new data and corresponding KV is not guaranteed. For example, if a power failure happens after KV commit but before data overwrite, committed KV will contain a PGLog entry for the new write but data may still remain as old.

ifed01:
@jhyoon01 tricky thing for this approach is that there is no need for KV update in some circumstances: no csum enabled, no object attribute update, data chunk is completely overwritten, no object size increase etc. Hence I don't eliminate WAL at all but just introduce a 'fast-track' when this is possible, i.e. when there is no need for KV or other data chunk updates.

src/os/bluestore/BlueStore.cc
```diff
@@ -5415,19 +5423,28 @@ int BlueStore::queue_transactions(
...
+  bool queued = _txc_write_nodes(txc, txc->t);
   // journal wal items
   if (txc->wal_txn) {
     // move releases to after wal
     txc->wal_txn->released.swap(txc->released);
     assert(txc->released.empty());
 
     txc->wal_txn->seq = wal_seq.inc();
+
     bufferlist bl;
```
yuyuyu101: redundancy

yuyuyu101:
notes:
1. I think in all normal client op, we always have attribute like "_" updated, I don't know what's the case we can eliminate onode update....
2. one block atomic write is really tricky, there exists several layers between real block device atomic write(or disk durable support) and vfs write. Except explicit atomic write primitive like Fusion IO's driver supports atomic write, NVME atomic write op, even we can't fully trust block size provided by stat. Section atomic write discuss everywhere like postgresql, sqlite..... https://www.reddit.com/r/programming/comments/1eypf/sqlite_only_uses_17_year_old_technology_out_of  https://www.postgresql.org/message-id/4AFD5517.8030207@shopzeus.com
3. avoid wal circumstance is rarely in cluster env, I think it's not worth to maintain this io path since it needs a lots of invader judgements.

PR #10909, opened 2016-08-29
common/secret.c: Do not use keyutils running on FreeBSD

wjwithagen:
And in that case make use of the option already offered on older
kernels:
   " running against older kernel; fall back to secret= in options"

So that'll mean that FreeBSD has to keep the secret in the options.

### DIFF
--- a/src/common/secret.c
+++ b/src/common/secret.c
@@ -18,7 +18,9 @@
 #include <unistd.h>
 #include <errno.h>
 #include <fcntl.h>
+#if defined(__linux__)
 #include <keyutils.h>
+#endif
 #include <sys/types.h>
 
 #include "common/armor.h"
@@ -50,6 +52,7 @@ int read_secret_from_file(const char *filename, char *secret, size_t max_len)
   return 0;
 }
 
+#if defined(__linux__)
 int set_kernel_secret(const char *secret, const char *key_name)
 {
   /* try to submit key to kernel via the keys api */
@@ -85,6 +88,17 @@ int is_kernel_secret(const char *key_name)
   serial = request_key("ceph", key_name, NULL, KEY_SPEC_USER_KEYRING);
   return serial != -1;
 }
+#else
+int set_kernel_secret(const char *secret, const char *key_name)
+{
+  return -ENOSYS;
+}
+
+int is_kernel_secret(const char *key_name)
+{
+  return -ENOSYS;
+}
+#endif
 
 int get_secret_option(const char *secret, const char *key_name,
 		      char *secret_option, size_t max_len)


wjwithagen:
@tchaikov 
First attempt to work the keyutils code out of the FreeBSD port. It will allow for more code and tests to be run.

src/common/secret.c
```diff
@@ -85,6 +88,17 @@ int is_kernel_secret(const char *key_name)
...
 }
+#else
+int set_kernel_secret(const char *secret, const char *key_name)
+{
+  return -EINVAL;
+}
+
+int is_kernel_secret(const char *key_name)
+{
+  return -EINVAL;
```
tchaikov: this is wrong, `is_kernel_secret()` returns 1 on success, 0 otherwise. we do not check the return code to see if it's negative or not. BTW, is mount.ceph and/or krbd compilable on FreeBSD? they are the only two callers of this function.

src/common/secret.c
```diff
@@ -107,7 +121,7 @@ int get_secret_option(const char *secret, const char *key_name,
   if (secret) {
     ret = set_kernel_secret(secret, key_name);
     if (ret < 0) {
-      if (ret == -ENODEV || ret == -ENOSYS) {
+      if (ret == -ENODEV || ret == -ENOSYS || ret == -EINVAL) {
```
tchaikov: `-EINVAL` is a valid return code of `set_kernel_secret()` on GNU/Linux. if the input parameter is invalid, -EINVAL is returned. probably you can return `-ENOSYS` on FreeBSD. it's handled properly already.

wjwithagen: @tchaikov 
I'll return ENOSYS.
And one of the reasons to implement this, is that will allow parts of the FreeBSD variants of mount.ceph and krbd to share more of the common code.
Next to the fact that it now will allow to make less exemptions in the Cmake files. Building a library that includes secret.c.o (for what ever reason) will no longer fail now.

src/common/secret.c
tchaikov: @wjwithagen ping.

wjwithagen:
retest this please

idryomov:
This file is already Linux-specific, so there is no need for ifdefs.  These functions are used by "rbd map" and mount.ceph, both of which require Linux kernel and are of no use on FreeBSD, so you can exclude it from the build entirely.

src/common/secret.c
idryomov: I'd rather we have exemptions in the build system, rather than ifdefs in the code, especially when the entire file is to be excluded.  secret.c is shared between krbd.cc and mount.ceph.c and there is no reason for anything else to include it.

Just like krbd.cc and mount.ceph.c, secret.c shouldn't be built on anything other than Linux.

wjwithagen:
@idryomov 
Real nice that you close the PR without much discussion or anything.
I think it would have relaxed a lot of conditionals in the CMake files.
Guess I'll need to try to get it done there.

idryomov:
Well, we used to have these checks in automake - there is really no reason to build Linux kernel helpers on other OSes.  Why add #if defined(__linux__), return -ENOSYS, etc to the file that doesn't need to be compiled anywhere but Linux?

wjwithagen:
I'm going to try and build the equivalent of krd and mount.ceph on FreeBSD.
There is already example code which does more or less the same for another protocol.
But if needed I'll revice this.
Now I'm trying to get secret.c execluded in CMake without too may if()/endif(), trying to avoid empty expansion of `$<TARGET_OBJECTS:parse_secret_objs>`. That will blow up the remainder of the CMake parsing.

PR #10910, opened 2016-08-29
os/bluestore: fix statfs tests

liewegas:


ifed01:
Reviewed-by: Igor Fedotov <ifedotov@mirantis.com>

src/test/objectstore/store_test.cc
```diff
@@ -1123,6 +1123,7 @@ TEST_P(StoreTest, BluestoreStatFSTest) {
   if(string(GetParam()) != "bluestore")
     return;
   g_conf->set_val("bluestore_compression", "force");
+  g_conf->set_val("bluestore_min_alloc_size", "65536");
```
ifed01: Just realized that this wouldn't work as min_alloc_size is updated from _open_bdev method only.
And I suggest to revert settings to their original values on test case completion.

ifed01: Made PR #10926 to fix that.

PR #10914, opened 2016-08-29
mds: remove max_mds config option

batrick:
It is now required that all changes to max_mds use the run-time `ceph fs set max_mds` command. The rationale for this change is that it is confusing to have a configuration for max_mds which is only observed at file system creation.

Fixes: http://tracker.ceph.com/issues/17105

### DIFF
--- a/src/common/config_opts.h
+++ b/src/common/config_opts.h
@@ -530,7 +530,6 @@ OPTION(mds_inject_traceless_reply_probability, OPT_DOUBLE, 0) /* percentage
 OPTION(mds_wipe_sessions, OPT_BOOL, 0)
 OPTION(mds_wipe_ino_prealloc, OPT_BOOL, 0)
 OPTION(mds_skip_ino, OPT_INT, 0)
-OPTION(max_mds, OPT_INT, 1)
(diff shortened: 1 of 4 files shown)


renhwztetecs:
:+1:

jcsp:
Reviewed-by

src/common/config_opts.h
```diff
@@ -530,7 +530,7 @@ OPTION(mds_inject_traceless_reply_probability, OPT_DOUBLE, 0) /* percentage
 OPTION(mds_wipe_sessions, OPT_BOOL, 0)
 OPTION(mds_wipe_ino_prealloc, OPT_BOOL, 0)
 OPTION(mds_skip_ino, OPT_INT, 0)
-OPTION(max_mds, OPT_INT, 1)
+OPTION(max_mds, OPT_INT, 1) // Obsolete and ignored! Use `ceph fs set max_mds` instead.
```
gregsfortytwo: Why not just kill it entirely? I don't think unknown config options will break anything, and leaving it in is confusing.

batrick:
@gregsfortytwo thanks, I made that change.

PR #10919, opened 2016-08-30
rgw: Replacing '+' with "%20" in canonical uri for s3 v4 auth.

pritha-srivastava:
s3cmd encodes space as "%20" while signature computation and
encodes space as '+' while sending the canonical uri. This
results in a SignatureMismatch Error in rgw, since rgw
computes the signature based on the request received from
the client (s3cmd in this case).

### DIFF
--- a/src/rgw/rgw_rest_s3.cc
+++ b/src/rgw/rgw_rest_s3.cc
@@ -11,6 +11,7 @@
 #include "common/ceph_json.h"
 #include "common/safe_io.h"
 #include <boost/algorithm/string.hpp>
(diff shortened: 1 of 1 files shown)


yehudasa:
@pritha-srivastava lgtm. @oritwas can we test this?

jhaprins:
What happens when we have a literal + sign in the canonical url? So the + sign is not encoded but is meant literal? Wouldn't this then result in a wrong url translation?

pritha-srivastava:
The '+' sign has to be percent encoded in the canonical uri.

PR #10920, opened 2016-08-30
rgw: fix orphan data leak on Bucket deletion

praveengt:
Problem: If we have orphan data in the bucket because of partial Multipart upload. The rados objects corresponding to the Multipart upload gets leaked when we delete the bucket.

Solution: We should return ENOTEMPTY if we find any kind of object and not just objects that are visible to users, i.e. if ent_map.size() is not equal zero, we should return ENOTEMPTY

Tracker ID: http://tracker.ceph.com/issues/17164

### DIFF
--- a/src/rgw/rgw_rados.cc
+++ b/src/rgw/rgw_rados.cc
@@ -7344,15 +7344,9 @@ int RGWRados::delete_bucket(rgw_bucket& bucket, RGWObjVersionTracker& objv_track
     if (r < 0)
       return r;
 
-    string ns;
-    std::map<string, RGWObjEnt>::iterator eiter;
-    rgw_obj_key obj;
-    string instance;
-    for (eiter = ent_map.begin(); eiter != ent_map.end(); ++eiter) {
-      obj = eiter->second.key;
-
-      if (rgw_obj::translate_raw_obj_to_obj_in_ns(obj.name, instance, ns))
-        return -ENOTEMPTY;
+    /* Return ENOTEMPTY if we find any kind of object inside the bucket */
+    if (ent_map.size() != 0) {
+      return -ENOTEMPTY;
     }
   } while (is_truncated);
 


mattbenjamin:
I think this step (ENOTEMPTY returns) is unambiguously correct logically.  @yehudasa, @oritwas do you agree?

yehudasa:
@mattbenjamin yes

oritwas:
@mattbenjamin I agree

mattbenjamin:
I did some manual verification, but also pushed a wip-testing branch.

mattbenjamin:
@praveengt just fyi, I'm seeing at least one s3test failure getting unexpected !empty on deleting a bucket at end of test (thanks for help reading @oritwas );  http://qa-proxy.ceph.com/teuthology/mbenjamin-2016-09-12_16:31:00-rgw-wip-matt-testing-notempty---basic-mira/412201/teuthology.log; will root cause.

praveengt:
Hi Matt,

The fix will prevent leaks. so if you are getting failures while bucket deletion, it means that they are potentially some leak that is about to happen. Did you check whether there are no orphans in the bucket before deleting the bucket

yehudasa:
@mattbenjamin ping

mattbenjamin:
@yehudasa, @praveengt scheduling new teuthology rgw suite today, will update when it completes

dachary:

dachary:

dachary:

mattbenjamin:
With this change,  suite cleanup consistently fails removing a user w/purge-data, will root cause.

2017-02-10T18:44:56.187 INFO:teuthology.orchestra.run.smithi150:Running: 'adjust-ulimits ceph-coverage /home/ubuntu/cephtest/archive/coverage radosgw-admin -n client.0 user rm --uid bar.client.0 --purge-data'

http://pulpito.ceph.com/mbenjamin-2017-02-10_13:13:51-rgw-wip-bucket-del-leak---basic-smithi/

oritwas:
jenkins test this please

Abhishekvrshny:
The RCA for teuthology failure have been updated on [tracker](http://tracker.ceph.com/issues/17164#note-6). Also, a new PR has been raised with the proposed changes here : https://github.com/ceph/ceph/pull/15630

mattbenjamin:
@Abhishekvrshny just waiting for #15630 to retest

Abhishekvrshny:
Closing this, as https://github.com/ceph/ceph/pull/15630 is merged.

PR #10922, opened 2016-08-30
First pile of statx patches

jtlayton:
github wouldn't allow me to reopen my old PR, so opening a new one here. Most of the changes are in response to Greg's review of the PR here:

[https://github.com/ceph/ceph/pull/10842](https://github.com/ceph/ceph/pull/10842)

This patchset is the first pile of patches to add ceph_statx support to the userland client. The basic idea is to add a new interface patterned after the proposed statx() interface for Linux. The idea there is to extend the traditional stat() interface in several ways:

1) allow access to more inode attribute fields, in particular the birthtime (btime) and a change attribute (a'la NFSv4), in a way that is extensible in the future.

2) allow applications to request just specific fields in the ceph_statx struct be filled out. Note that you can (and will) often get more than you request, but you should always get what is requested or an error.

3) allow applications to do a "lazy" statx, where we fill out the attribute fields from whatever we have in cache, without querying the server. This is handy for performance reasons or in cases where we don't really need up-to-date information.

In addition, a ceph_setattrx has also been added, which allows the setting of the btime as well as the other fields that ceph_setattr allows. I also have a companion patchset for samba that converts it to use these new interfaces. That seems to work properly, though I still need to roll a testcase to verify that setting the btime via smb actually works.

Pay close attention to the first patch in the file as well, which I think may help performance in some cases by allowing us to request the caps needed to do an operation during the lookup phase of path-based operations.

I've also started some work to do a patchset for ganesha, but that requires adding a number of new ceph_ll_* calls. I think it's probably best to get this merged first and then do that part afterward.


jtlayton:
Note that while this is largely a feature request, we may want to consider backporting this patch to earlier releases:

     client: pass a mask parameter to path_walk

I think that may be a performance win in many cases.

src/client/Client.cc
```diff
@@ -7571,7 +7576,7 @@ int Client::open(const char *relpath, int flags, mode_t mode, int stripe_unit,
     string dname = dirpath.last_dentry();
     dirpath.pop_dentry();
     InodeRef dir;
-    r = path_walk(dirpath, &dir, true, uid, gid);
+    r = path_walk(dirpath, &dir, true, 0, uid, gid);
```
jtlayton: We probably ought to request CEPH_CAP_AUTH_SHARED here if client_permissions is true. I'll plan to fold that into this patch once others have had a chance to comment.

src/messages/MClientRequest.h
```diff
@@ -160,7 +161,28 @@ class MClientRequest : public Message {
 
   void decode_payload() {
     bufferlist::iterator p = payload.begin();
-    ::decode(head, p);
+
+    if (header.version >= 4) {
+      ::decode(head, p);
+    } else {
+      struct ceph_mds_request_head_legacy *old_mds_head =
+	(struct ceph_mds_request_head_legacy *)&(head.oldest_client_tid);
```
gregsfortytwo: Can't we just create a new (genuine) ceph_mds_request_head_legacy and then invoke a copy function to move it into our real one?
I realize just using the pointer is more efficient but hopefully this is an uncommon case, and doing it with pointer magic makes it really hard to find. Imagine in future when we do an incompat change and totally reorder all the member fields.

gregsfortytwo:
I think I've been over the new bits here; everything else looks good.

(In future, it's a lot easier to do follow-on reviews if you do add-on FIXUP/SQUASH patches that we can look at, then squash down before final merge.)

src/messages/MClientRequest.h
jtlayton: Yeah, we could. I'll hold my nose and do that. ;)

jtlayton:
FWIW, I've done a blog post expliaining why I think that this method of handling the change attribute is OK:

https://jtlayton.wordpress.com/2016/09/01/cephfs-and-the-nfsv4-change-attribute/

ffilz:
Confused by my getting assigned the pr...

gregsfortytwo:
blog comment: https://jtlayton.wordpress.com/2016/09/01/cephfs-and-the-nfsv4-change-attribute/#comment-164

:)

jtlayton:
@ffilz: only assigned to you to make sure you saw the set

@gregsfortytwo: I replied on the blog. There should be no reliance on any particular client or server behavior here. That said, if you can come up with a scenario where this scheme falls down, then I'd be very interested to hear it.

OTOH, if you're talking about the ceph change attr giving guarantees over and above what the NFSv4 spec mandates, then that's another matter entirely.

Are you suggesting we ought to allow clients to inspect the change_attr and somehow determine whether their cached attrs are still valid, even when they don't have the caps that they'd normally need to see them? If so, I'm skeptical but am interested to hear more of what you're thinking.

src/messages/MClientRequest.h
```diff
@@ -165,11 +165,10 @@ class MClientRequest : public Message {
...
       ::decode(head, p);
     } else {
-      struct ceph_mds_request_head_legacy *old_mds_head =
-	(struct ceph_mds_request_head_legacy *)&(head.oldest_client_tid);
-
-      ::decode(*old_mds_head, p);
+      struct ceph_mds_request_head_legacy old_mds_head;
 
+      ::decode(old_mds_head, p);
+      memcpy(&head.oldest_client_tid, &old_mds_head, sizeof(old_mds_head));
```
gregsfortytwo: I think this is getting lost in translation. Enjoy your ability to use C++ and write "invisible" functions; instead of memcpy do something like
```head.copy_from_legacy(old_mds_head);```
or perhaps ```copy_legacy_head(head, old_mds_head)``` since these are declared as shared C structs. That way if we update stuff, it requires changing a single function located next to the ceph_mds_request_head struct instead of searching out everybody who copies them.

Granted, if we only have the one user maybe it's not such a big deal and is fine as it stands. That's just the sort of protective programming we can engage in even though it makes the kernel devs angry for being inefficient. ;)

gregsfortytwo:
Discussed this on irc as well and I think the protocol is all good.

jtlayton:
Now that I thought about it some more, I think both @gregsfortytwo and @ffilz are right, and we need to ensure that any outstanding exclusive caps are revoked when you want to fetch the change attr. The patch that I pushed to the branch late last night should do the right thing there.

src/messages/MClientRequest.h
jtlayton: It does trigger my gag reflex, but ok...

gregsfortytwo:
Updates look good.

Just to be clear, there's nothing in here or the expected wrapper code that will automatically trigger a change_attr lookup unless it's asked for explicitly via statx (or otherwise genuinely needed), right? (I don't see anything, but that would be a really obnoxious performance bug to track down.) I'm thinking specifically of default masks and whatnot.

jtlayton:
Yes, the only to get the change_attr is to specifically request it via CEPH_STATX_VERSION flag, in the "want" mask.

jtlayton:
Note too that I don't think the existing code is quite correct wrt to the ctime. It seems like client1 could cache a mode change (for instance) and if client2 doesn't request As caps, then you might not see the ctime change for that. We may need to consider altering how that works here as well...

gregsfortytwo:
That wouldn't surprise me and might be one of the deliberate cheats Sage snuck in...stuff like times basically forces synchronicity and that isn't much fun. :/
But if there are issues we should open tracker bugs for them and can discuss the cost of the fix, if we want to make it configurable, etc.

jtlayton:
Actually, I think we're ok with the ctime. In order to actually query for the ctime, you have to do a ceph_stat() anyway (or the FUSE equivalent), which means that you're going to ask for all the shared caps anyway.

Where it's not yet correct is in statx_to_mask() function. I'll fix...

PR #10923, opened 2016-08-30
os/bluestore/bluefs: use map to track dirty files

xiexingguo:
Therefore dirty files are arranged by dirty_seq and then
dirty files of same dirty_seq are grouped into list and
as a result we can clean up dirty files much faster during
_flush_and_sync_log().

### DIFF
--- a/src/os/bluestore/BlueFS.cc
+++ b/src/os/bluestore/BlueFS.cc
@@ -156,7 +156,7 @@ int BlueFS::reclaim_blocks(unsigned id, uint64_t want,
 {
   std::unique_lock<std::mutex> l(lock);
   dout(1) << __func__ << " bdev " << id
-          << " want 0x" << std::hex << want << std:: dec << dendl;
+          << " want 0x" << std::hex << want << std::dec << dendl;
   assert(id < alloc.size());
   assert(alloc[id]);
   int r = alloc[id]->reserve(want);
@@ -1770,7 +1770,6 @@ int BlueFS::open_for_write(
     }
     file = new File;
     file->fnode.ino = ++ino_last;
-    file->fnode.mtime = ceph_clock_now(NULL);
     file_map[ino_last] = file;
     dir->file_map[filename] = file;
     ++file->refs;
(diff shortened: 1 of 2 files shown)


src/os/bluestore/BlueFS.cc
```diff
@@ -1401,14 +1415,18 @@ int BlueFS::_flush_range(FileWriter *h, uint64_t offset, uint64_t length)
...
     log_t.op_file_update(h->file->fnode);
     if (h->file->dirty_seq == 0) {
-      dirty_files.push_back(*h->file);
       dout(20) << __func__ << " dirty_seq = " << log_seq + 1
 	       << " (was clean)" << dendl;
     } else {
+      // already in dirty list, erase first
+      assert(dirty_files.count(h->file->dirty_seq));
+      auto it = dirty_files[h->file->dirty_seq].iterator_to(*h->file);
+      dirty_files[h->file->dirty_seq].erase(it);
```
liewegas: this could check for dirty_seq == log_seq+1 so that we don't remove+add in the case that the log_seq is the same.

src/os/bluestore/BlueFS.cc
```diff
@@ -1224,7 +1226,10 @@ void BlueFS::_compact_log_async(std::unique_lock<std::mutex>& l)
 
   // delete the new log, remove from the dirty files list
   _close_writer(new_log_writer);
-  dirty_files.erase(dirty_files.iterator_to(*new_log));
+  if (dirty_files.count(new_log->dirty_seq)) {
```
liewegas: I think this should be if (new_log->dirty_seq).  if that is non-zero, then we know the matching dirty_files list exists for that seq, right?

xiexingguo:
Comments addressed and passed local store_test:
```
2016-08-31 16:09:31.009887 7fc72527f640  1 bluestore(store_test_temp_dir) fsck finish with 0 errors
[       OK ] ObjectStore/StoreTest.TooManyBlobsTest/2 (20885 ms)
[----------] 60 tests from ObjectStore/StoreTest (1842355 ms total)

[==========] 60 tests from 1 test case ran. (1842356 ms total)
[  PASSED  ] 60 tests.

liewegas:
Thanks!

xiexingguo:
my pleasure:-)

PR #10932, opened 2016-08-31
[WIP] rgw: add compression to client, contd.

aclamk:
This is continuation of Alyona's work on compression #7715.

### DIFF
--- a/systemd/ceph-mds@.service
+++ b/systemd/ceph-mds@.service
@@ -15,6 +15,7 @@ PrivateDevices=yes
 ProtectHome=true
 ProtectSystem=full
 PrivateTmp=true
+TasksMax=infinity
 
 [Install]
 WantedBy=ceph-mds.target
--- a/systemd/ceph-mon@.service
+++ b/systemd/ceph-mon@.service
@@ -21,6 +21,7 @@ PrivateDevices=yes
 ProtectHome=true
 ProtectSystem=full
 PrivateTmp=true
+TasksMax=infinity
 
 [Install]
 WantedBy=ceph-mon.target
--- a/systemd/ceph-osd@.service
+++ b/systemd/ceph-osd@.service
@@ -15,6 +15,7 @@ ExecReload=/bin/kill -HUP $MAINPID
 ProtectHome=true
 ProtectSystem=full
 PrivateTmp=true
+TasksMax=infinity
 
 [Install]
 WantedBy=ceph-osd.target
--- a/systemd/ceph-radosgw@.service
+++ b/systemd/ceph-radosgw@.service
@@ -14,6 +14,7 @@ PrivateDevices=yes
 ProtectHome=true
 ProtectSystem=full
 PrivateTmp=true
+TasksMax=infinity
 
 [Install]
 WantedBy=ceph-radosgw.target
--- a/systemd/ceph-rbd-mirror@.service
+++ b/systemd/ceph-rbd-mirror@.service
@@ -17,6 +17,7 @@ PrivateTmp=true
 Restart=on-failure
 StartLimitInterval=30min
 StartLimitBurst=3
+TasksMax=infinity
 
 [Install]
 WantedBy=ceph-rbd-mirror.target


cbodley:
the [teuthology run](http://pulpito.ceph.com/cbodley-2016-09-06_10:26:26-rgw-wip_rgw_compression---basic-mira/) came back with a number of swift and s3test failures. the s3test failures i saw were in s3tests.functional.test_s3.test_region_copy_object - here's an [example](https://gist.github.com/cbodley/7f409438da92e0ba3a9cf309174b5073)

i'm not seeing an obvious pattern to these failures. i'm seeing them happen on both centos and ubuntu machines, and with both the snappy and zlib plugins

cbodley:
@aclamk pushed your latest commits to the gitbuilders, and will schedule another run.

could you rebase at some point please?

aclamk:
Doing it now.

cbodley:
[teuthology results](http://pulpito.ceph.com/cbodley-2016-09-21_09:37:49-rgw-wip_rgw_compression---basic-smithi/) look pretty good, but we're still getting `buffer::end_of_buffer` exceptions in swift tests

src/rgw/rgw_op.cc
```diff
@@ -2855,19 +2936,34 @@ void RGWPutObj::execute()
...
+
+  // add attr to bucket to know original  size of data
+  res = rgw_bucket_compression_info_from_attrset(s->bucket_attrs, bucket_size);
+  if (res < 0)
+    lderr(s->cct) << "ERROR: failed to read decompressed bucket size, cannot update stat" << dendl;
+  op_ret = update_compressed_bucket_size(s->obj_size, bucket_size);
+  if (op_ret < 0)
+    ldout(s->cct, 0) << "NOTICE: put_bucket_info on bucket=" << s->bucket.name
+         << " returned err=" << op_ret << dendl;
+
```
cbodley: @aclamk i'm concerned about this extra request to update the bucket's size for every PUT

i made a quick hack that passes this `s->obj_size` directly through `RGWPutObjProcessor::complete()` -> `RGWRados::Object::Write::write_meta()` -> `RGWRados::Bucket::UpdateIndex::complete()` -> `cls_rgw_bucket_complete_op()`, so that the bucket index tracked this uncompressed size instead of the compressed size calculated by `RGWPutObjProcessor`. this was passing s3tests for me

i brought this up in the standup, and @yehudasa asked that we pass in both the compressed and uncompressed size, so the bucket stats could track the sum for each. i've made the necessary changes in cls/rgw and have been working my way up the chain. i hope to have something ready to show you today or tomorrow

cbodley:
the [latest run](http://pulpito.ceph.com/cbodley-2016-10-03_11:34:04-rgw-wip_rgw_compression---basic-mira/) did have your fixes in it, but i was mistaken about the swift test failures - they aren't the snappy assertions we were seeing before. a lot of the testCopy tests are failing with 'Compression failed with exit code -1' errors when trying to read the result

src/compressor/zlib/ZlibCompressor.cc
```diff
@@ -178,7 +178,7 @@ int ZlibCompressor::decompress(bufferlist::iterator &p, size_t compressed_size,
   strm.avail_in = 0;
   strm.next_in = Z_NULL;
   if (*p == 1)
-    ret = inflateInit2(&strm, -HIST_SIZE);
+    ret = inflateInit2(&strm, -15 /*-HIST_SIZE*/);
```
ifed01: There is a competing PR #11445 to fix the same issue.

src/test/compressor/test_compression_zlib.cc
```diff
@@ -177,6 +177,69 @@ TEST(ZlibCompressor, zlib_isal_compatibility)
   EXPECT_TRUE(exp.contents_equal(after));
 }
 
+TEST(ZlibCompressor, isal_compress_zlib_decompress_random)
```
ifed01: Looks like you're using legacy master fork, compression UTs are a bit different now

cbodley:
@aclamk it should be safe to rebase on current master. feel free to pull in the fix from @ifed01 (temporarily) for teuthology testing

PR #10933, opened 2016-08-31
Wip rebuild monstore

tchaikov:
Fixes: http://tracker.ceph.com/issues/17179

### DIFF
--- a/doc/rados/troubleshooting/troubleshooting-mon.rst
+++ b/doc/rados/troubleshooting/troubleshooting-mon.rst
@@ -383,6 +383,81 @@ example::
 
 	iptables -A INPUT -m multiport -p tcp -s {ip-address}/{netmask} --dports 6789,6800:7300 -j ACCEPT
 
+Monitor Store Failures
+======================
+
+Symptoms of store corruption
+----------------------------
+
+Ceph monitor stores the `cluster map`_ in a key/value store such as LevelDB. If
+a monitor fails due to the key/value store corruption, following error messages
+might be found in the monitor log::
+
+  Corruption: error in middle of record
+
+or::
+
+  Corruption: 1 missing files; e.g.: /var/lib/ceph/mon/mon.0/store.db/1234567.ldb
+
+Recovery using healthy monitor(s)
+---------------------------------
+
+If there is any survivers, we can always `replace`_ the corrupted one with a
+new one. And after booting up, the new joiner will sync up with a healthy
+peer, and once it is fully sync'ed, it will be able to serve the clients.
+
+Recovery using OSDs
+-------------------
+
+But what if all monitors fail at the same time? Since users are encouraged to
+deploy at least three monitors in a Ceph cluster, the chance of simultaneous
+failure is rare. But unplanned power-downs in a data center with improperly
+configured disk/fs settings could fail the underlying filesystem, and hence
+kill all the monitors. In this case, we can recover the monitor store with the
+information stored in OSDs.::
+
+  ms=/tmp/mon-store
+  mkdir $ms
+  # collect the cluster map from OSDs
+  for host in $hosts; do
+    rsync -avz $ms user@host:$ms
+    rm -rf $ms
+    ssh user@host <<EOF
+      for osd in /var/lib/osd/osd-*; do
+        ceph-objectstore-tool --data-path $osd --op update-mon-db --mon-store-path $ms
+      done
+    EOF
+    rsync -avz user@host:$ms $ms
+  done
+  # rebuild the monitor store from the collected map, if the cluster does not
+  # use cephx authentication, there is no need to pass the "--keyring" option.
+  # i.e. use "ceph-monstore-tool /tmp/mon-store rebuild" instead
+  ceph-monstore-tool /tmp/mon-store rebuild -- --keyring /path/to/admin.keyring
+  # backup corrupted store.db just in case
+  mv /var/lib/ceph/mon/mon.0/store.db /var/lib/ceph/mon/mon.0/store.db.corrupted
+  mv /tmp/mon-store/store.db /var/lib/ceph/mon/mon.0/store.db
+
+The steps above
+
+#. collect the map from all OSD hosts,
+#. then rebuild the store,
+#. replace the corrupted store on ``mon.0`` with the recovered copy.
+
+Known limitations
+~~~~~~~~~~~~~~~~~
+
+Following information are not recoverable using the steps above:
+
+- **some added keyrings**: all the OSD keyrings added using ``ceph auth add`` command
+  are recovered from the OSD's copy. And the ``client.admin`` keyring is imported
+  using ``ceph-monstore-tool``. But the MDS keyrings and other keyrings are missing
+  in the recovered monitor store. You might need to re-add them manually.
+
+- **pg settings**: the ``full ratio`` and ``nearfull ratio`` settings configured using
+  ``ceph pg set_full_ratio`` and ``ceph pg set_nearfull_ratio`` will be lost.
+
+- **MDS Maps**: the MDS maps are lost.
+
 
 Everything Failed! Now What?
 =============================
@@ -480,4 +555,6 @@ based on that.
 Finally, you should reach out to us on the mailing lists, on IRC or file
 a new issue on the `tracker`_.
 
+.. _cluster map: ../../architecture#cluster-map
+.. _replace: ../operation/add-or-rm-mons
 .. _tracker: http://tracker.ceph.com/projects/ceph/issues/new
--- a/src/tools/CMakeLists.txt
+++ b/src/tools/CMakeLists.txt
@@ -35,6 +35,7 @@ endif(WITH_TESTS)
 
 add_executable(ceph-objectstore-tool
   ceph_objectstore_tool.cc
+  rebuild_mondb.cc
   RadosDump.cc)
 target_link_libraries(ceph-objectstore-tool osd os global ${Boost_PROGRAM_OPTIONS_LIBRARY} ${CMAKE_DL_LIBS})
 if(WITH_FUSE)
--- a/src/tools/ceph_monstore_tool.cc
+++ b/src/tools/ceph_monstore_tool.cc
@@ -20,8 +20,11 @@
 #include "common/Formatter.h"
 #include "common/errno.h"
 
+#include "auth/KeyRing.h"
+#include "auth/cephx/CephxKeyServer.h"
 #include "global/global_init.h"
 #include "include/stringify.h"
+#include "mon/AuthMonitor.h"
 #include "mon/MonitorDBStore.h"
 #include "mon/Paxos.h"
 #include "mon/MonMap.h"
@@ -217,6 +220,8 @@ void usage(const char *n, po::options_description &d)
   << "                                  (rewrite-crush -- --help for more info)\n"
   << "  inflate-pgmap [-- options]      add given number of pgmaps to store\n"
   << "                                  (inflate-pgmap -- --help for more info)\n"
+  << "  rebuild                         rebuild store\n"
+  << "                                  (rebuild -- --help for more info)\n"
   << std::endl;
   std::cerr << d << std::endl;
   std::cerr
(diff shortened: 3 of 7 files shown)


src/tools/CMakeLists.txt
```diff
@@ -33,6 +33,7 @@ install(PROGRAMS
 
 add_executable(ceph-objectstore-tool
   ceph_objectstore_tool.cc
+  rebuild_mondb.cc
```
athanatos: Probably also needs to go in automake?

athanatos:
@gregsfortytwo Can you take a look?

src/tools/CMakeLists.txt
tchaikov: will do. thought we were ditching off automake =)

doc/rados/troubleshooting/troubleshooting-mon.rst
```diff
@@ -383,6 +383,80 @@ example::
...
+---------------------------------
+
+If there is any survivers, we can always `replace`_ the corrupted one with a
+new one. And after booting up, the new joiner will sync up with a healthy
+peer, and once it is fully sync'ed, it will be able to serve the clients.
+
+Recovery using OSDs
+-------------------
+
+But what if all monitors fails at the same time? Since users are encouraged to
```
tchaikov: s/fails/fail/

src/tools/ceph_monstore_tool.cc
```diff
@@ -509,6 +514,148 @@ int inflate_pgmap(MonitorDBStore& st, unsigned n, bool can_be_trimmed) {
   return 0;
 }
 
+static int update_auth(MonitorDBStore& st, const string& keyring_path)
+{
+  // add a almighty ceph.client.admin.keyring
```
tchaikov: s/a/an/

gregsfortytwo:
Paging @jecluis as well. ;)

doc/rados/troubleshooting/troubleshooting-mon.rst
```diff
@@ -383,6 +383,80 @@ example::
 
 	iptables -A INPUT -m multiport -p tcp -s {ip-address}/{netmask} --dports 6789,6800:7300 -j ACCEPT
 
+Monitor Store Failures
+======================
+
+Symptoms of store corruption
+----------------------------
+
+Ceph monitor stores the `cluster map`_ in a key/value store. If a monitor in a
```
gregsfortytwo: We might want to say something like "a key/value store such as LevelDB".

I realize it's configurable and we might switch to RocksDB, but I think in the medium term anybody looking at this is going to know that LevelDB got messed up and not seeing that this is a repair tool for that case will confuse them. (Avoiding the name is especially weird since we are including LevelDB error messages below.)

doc/rados/troubleshooting/troubleshooting-mon.rst
```diff
@@ -383,6 +383,80 @@ example::
...
+If there is any survivers, we can always `replace`_ the corrupted one with a
+new one. And after booting up, the new joiner will sync up with a healthy
+peer, and once it is fully sync'ed, it will be able to serve the clients.
+
+Recovery using OSDs
+-------------------
+
+But what if all monitors fails at the same time? Since users are encouraged to
+deploy at least three monitors in a Ceph cluster, the chance of simultaneous
+failure is rare. But unplanned power-downs in a data centre could fail the
```
gregsfortytwo: How about "in a data center with improperly configured disk/fs settings"; we don't want to make it sound random.

doc/rados/troubleshooting/troubleshooting-mon.rst
```diff
@@ -383,6 +383,80 @@ example::
...
+underlying filesystem, and hence kill all the monitors. In this case, we can
+recover the monitor store with the information stored in OSDs.::
+
+  ms=/tmp/mon-store
+  mkdir $ms
+  # collect the cluster map from OSDs
+  for host in $hosts; do
+    rsync -avz /tmp/mon-store user@host:$ms
```
gregsfortytwo: Why do we want to send over our existing maps to the remote side first?

Oh, are we sending all ours, then grabbing theirs, then overwriting everything we have locally? Why not just use rsync to merge them?

athanatos: Is that even possible?  Even if it unnecessarily moves stuff around, I think this is a pretty clear explanation of what needs to be where and how the tools are meant to be used, which is what's really important.

athanatos: @tchaikov Oh, what happens if one of the OSDs happens to be really old and there is a hole between its newest map and the oldest map on the others?  I think the correct answer is for the mon to have the newest contiguous chunk of OSDMaps, but I'm not sure if you handled that.

gregsfortytwo:
Still reading code, but I don't see any tests for this. I just *know* it's going to break every time we change storage stuff (like with ceph-mgr or something, for instance) and we aren't going to notice if the nightlies don't tell us.

So a basic scenario where we append a "rebuild" task at the end of one (or several) normal RADOS tests to start with. At least one where we have an OSD we deliberately keep out-of-date like Sam mentioned above. One we'll write when the FS repair stuff is done. Perhaps others but that's all I've got off the top of my head.

doc/rados/troubleshooting/troubleshooting-mon.rst
gregsfortytwo: Oh, this is actually creating a MonDBStore in place; I'd assumed it would be extracting to some intermediate format and then turning them into the mon format at the end. So yes, we definitely need to copy it back and forth to do that.

src/tools/rebuild_mondb.cc
```diff
@@ -0,0 +1,366 @@
...
+
+// rebuild
+//  - auth/${epoch}
+//  - auth/first_committed
+//  - auth/last_committed
+static int update_auth(const string& keyring_path,
+                       const OSDSuperblock& sb,
+                       MonitorDBStore& ms)
+{
+  // stolen from AuthMonitor::prepare_command(), where prefix is "auth add"
```
gregsfortytwo: It sounds like you actually copied the functions? If so we should definitely pull them out into a shared cc file or something, rather than having two copies that can diverge with different behavior or bugs!

src/tools/rebuild_mondb.cc
```diff
@@ -0,0 +1,366 @@
...
+    ntrimmed++;
+  }
+  if (sb.oldest_map > first_committed) {
+    t->put(prefix, first_committed_name, sb.oldest_map);
+  }
+
+  // add inc maps
+  unsigned nadded = 0;
+  const auto last = max(last_committed+1, sb.oldest_map);
+  for (auto e = last; e <= sb.newest_map; e++) {
```
gregsfortytwo: So here, if the OSD has newer maps but an out-of-date newest_map, we're leaving them behind. Can we do a listing of the prefix range and pull them all out instead?

src/tools/rebuild_mondb.cc
```diff
@@ -0,0 +1,366 @@
...
+    int nread = fs.read(coll_t::meta(), oid, 0, 0, bl);
+    if (nread <= 0) {
+      cerr << "missing " << oid << std::endl;
+      return EINVAL;
+    }
+    t->put(prefix, e, bl);
+    nadded++;
+  }
+  // add full maps
+  for (auto e = last; e <= sb.newest_map; e++) {
```
gregsfortytwo: Same here.

src/tools/rebuild_mondb.cc
```diff
@@ -0,0 +1,366 @@
...
+  // trim stale maps
+  unsigned ntrimmed = 0;
+  for (auto e = first_committed; e && e < sb.oldest_map; e++) {
+    t->erase(prefix, e);
+    t->erase(prefix, ms.combine_strings("full", e));
+    ntrimmed++;
+  }
+  if (sb.oldest_map > first_committed) {
+    t->put(prefix, first_committed_name, sb.oldest_map);
+  }
```
gregsfortytwo: Oh wait, if we do have a discontiguous OSD that we read first, and then later we chop off its maps, I think we leak the old ones in our store? Not a huge deal, but we should set up a trim if we run into this case.

gregsfortytwo:
Otherwise looks fine to me (although I only skimmed the pgmap updates).

src/tools/rebuild_mondb.cc
tchaikov: not really, i adapted them.

src/tools/rebuild_mondb.cc
tchaikov: if we have an OSD with osdmap[4..10] first, and

1. case 1, we update our store with osdmap[20..28], then osdmap[4..20) is chopped. no old maps is left in store. and we now have osdmap[20..28]
1. case 2, we update our store with osdmap[6..20], then osdmap[4..6) is chopped. no old maps is left in store. and we now have osdmap[6..20].
1. case 3, we update our store with osdmap[1..3], try to trim osdmap[4..1], so no maps gets trimmed. no maps is added.
1. case 4, we update our store with osdmap[1..6], try to trim osdmap[4..1], so no maps gets trimmed, and no maps is added.

so i think we are good?

src/tools/rebuild_mondb.cc
tchaikov: see https://github.com/ceph/ceph/blob/master/src/osd/OSD.cc#L6824, superblock and new osdmaps are written in the same transaction. so OSD can hardly have "new maps but an out-of-date newest_map". am i right?

src/tools/rebuild_mondb.cc
tchaikov: @gregsfortytwo can we leave the refactor for another PR? as i need to restructure monitor carefully to cater the needs on both sides.

doc/rados/troubleshooting/troubleshooting-mon.rst
tchaikov: @athanatos but I'm not sure if you handled that.

i do handle this. as @gregsfortytwo pointed out. see https://github.com/ceph/ceph/pull/10933#discussion-diff-77090613R217

tchaikov: @gregsfortytwo 
> I'd assumed it would be extracting to some intermediate format and then turning them into the mon format at the end.

i also evaluated this approach, it can be executed in parallel when collecting the OSD side info, hence more efficient. but seems current way is simpler. and this fix is only for disaster recovery which is not supposed to be used very frequently. so i am inclined to use this iterative way.

tchaikov:
all comments addressed except for https://github.com/ceph/ceph/pull/10933#discussion_r77090965 and https://github.com/ceph/ceph/pull/10933#discussion_r77090792.

i am working on a ceph-qa-suite task to 

1. nuke all store.db
1. rebuild one from OSDs
1. restore it to the first mon, mkfs on other mons
1. revive them

tchaikov:
jenkins, retest this please.

tchaikov:
changelog

- rebased against master
- ceph-monstore-tool: instead of getting the client.admin key and add an almighty keyring, import all keyrings from the `--keyring` file, to ready the mon.

tchaikov:
test added at https://github.com/ceph/ceph-qa-suite/pull/1169

src/tools/rebuild_mondb.cc
athanatos: Yeah, if we assume the OSD isn't messed up, than this is true.

src/tools/rebuild_mondb.cc
```diff
@@ -0,0 +1,364 @@
...
+      cerr << "missing " << oid << std::endl;
+      return -EINVAL;
+    }
+    t->put(prefix, ms.combine_strings("full", e), bl);
+  }
+  if (sb.newest_map > last_committed) {
+    t->put(prefix, last_committed_name, sb.newest_map);
+    t->put(prefix, ms.combine_strings("full", "latest"), sb.newest_map);
+  }
+  ms.apply_transaction(t);
```
athanatos: Hmm, can we actually write out all of the OSDMaps in one transaction?

tchaikov: no, will write them in batch.

tchaikov:
retest this please.

tchaikov:
changelog

- write osdmaps in batch of size 30
- use `#pragma once` for guarding multiple inclusion of `rebuild_mondb.h`

tchaikov:
@athanatos test passed at http://pulpito.ceph.com/kchai-2016-09-09_07:30:03-rados:singleton-wip-kefu-testing---basic-mira/

tchaikov:
retest this please.

zphj1987:
can this be test now 
@tchaikov

tchaikov:
@zphj1987 if you'd like to give it a spin, i am happy to learn if it works for you. =)

zphj1987:
@tchaikov 
i will test it in my environment, and tell you the result ，good job!

tchaikov:
changelog

- rebase against master to resolve conflicts.

@athanatos could you take a look again? thanks!

src/tools/rebuild_mondb.cc
```diff
@@ -0,0 +1,392 @@
...
+  const string first_committed_name("first_committed");
+  const string last_committed_name("last_committed");
+  epoch_t first_committed = ms.get(prefix, first_committed_name);
+  epoch_t last_committed = ms.get(prefix, last_committed_name);
+
+  // trim stale maps
+  unsigned ntrimmed = 0;
+  {
+    auto t = make_shared<MonitorDBStore::Transaction>();
+    for (auto e = first_committed; e && e < sb.oldest_map; e++) {
```
athanatos: Why e && e < oldest_map? What does first_committed = 0 mean?

src/tools/rebuild_mondb.cc
```diff
@@ -0,0 +1,392 @@
...
+      t->erase(prefix, e);
+      t->erase(prefix, ms.combine_strings("full", e));
+      ntrimmed++;
+    }
+    if (sb.oldest_map > first_committed) {
+      t->put(prefix, first_committed_name, sb.oldest_map);
+    }
+    ms.apply_transaction(t);
```
athanatos: Don't really need to apply if it empty, right -- move into immediately prior if block?

src/tools/rebuild_mondb.cc
```diff
@@ -0,0 +1,392 @@
...
+  }
+
+  // this number comes from the default value of osd_target_transaction_size,
+  // so we won't OOM or stuff too many maps in a single transaction if OSD is
+  // keeping a large series of osdmap
```
athanatos: If I'm reading this right, we could end up with an incremental map without the corresponding full map, and we can end up with maps outside of [first_committed_name, last_committed_name].  You probably want to make sure that each transaction leaves the store consistent.

src/tools/rebuild_mondb.cc
tchaikov: will do.

src/tools/rebuild_mondb.cc
tchaikov: if there is not such a k/v pair in store.db, `ms.get(prefix, key)` will simply return 0. this creates an illusion for ceph-objectstore-tool that there is always an osdmap#0 in store.db even there is nothing in it. and in real world, the osdmap starts at 1. i will accompany the code with some comments.

src/tools/rebuild_mondb.cc
tchaikov: will do.

tchaikov:
changelog

- add comment explaining why we don't want to trim osdmap#0.
- do not apply transaction if it's empty.
- make sure the store.db is consistent after each transaction adding osdmaps.

@athanatos mind taking a look again? thanks as always.

src/tools/rebuild_mondb.cc
```diff
@@ -0,0 +1,387 @@
...
+
+  // trim stale maps
+  unsigned ntrimmed = 0;
+  // osdmap starts at 1. if we have a "0" first_committed, then there is nothing
+  // to trim. and "1 osdmaps trimmed" in the output message is misleading. so
+  // let's make it an exception.
+  for (auto e = first_committed; first_committed && e < sb.oldest_map; e++) {
+    t->erase(prefix, e);
+    t->erase(prefix, ms.combine_strings("full", e));
+    t->put(prefix, first_committed_name, e + 1);
```
athanatos: Probably do this once at the end?

src/tools/rebuild_mondb.cc
```diff
@@ -0,0 +1,387 @@
...
+  epoch_t first_committed = ms.get(prefix, first_committed_name);
+  epoch_t last_committed = ms.get(prefix, last_committed_name);
+  auto t = make_shared<MonitorDBStore::Transaction>();
+
+  // trim stale maps
+  unsigned ntrimmed = 0;
+  // osdmap starts at 1. if we have a "0" first_committed, then there is nothing
+  // to trim. and "1 osdmaps trimmed" in the output message is misleading. so
+  // let's make it an exception.
+  for (auto e = first_committed; first_committed && e < sb.oldest_map; e++) {
```
athanatos: Seems like you can avoid the first_commited check if you change the loop to go up to MIN(sb.oldest_map, last_committed)?

src/tools/rebuild_mondb.cc
```diff
@@ -0,0 +1,387 @@
...
+      if (nread <= 0) {
+        cerr << "missing " << oid << std::endl;
+        return -EINVAL;
+      }
+      t->put(prefix, ms.combine_strings("full", e), bl);
+    }
+    nadded++;
+
+    // last_committed
+    t->put(prefix, last_committed_name, e);
```
athanatos: Put this into the if block with the apply?

tchaikov: then i have to repeat them in the 2nd if block with the apply.

src/tools/rebuild_mondb.cc
tchaikov: will do.

src/tools/rebuild_mondb.cc
tchaikov: - if `MIN(sb.oldest_map, last_committed)` evaluates to `sb.oldest_map`, we should trim osdmap[first_committed, sb.oldest_map**)**, 
- if `MIN(sb.oldest_map, last_committed)` evaluates to `last_committed`, we should trim osdmap[first_committed, last_committed**]**, 

blindly trim up to sb.oldest_map might be easier to understand.

tchaikov:
changelog

- update the "first_committed" once in the end.

tchaikov:
changelog

- remove trailing spaces.

@athanatos

tchaikov:
being tested at http://pulpito.front.sepia.ceph.com:80/kchai-2016-09-12_18:18:40-rados:singleton-wip-kefu-testing---basic-mira/

athanatos:
LGTM, @gregsfortytwo One last look?

doc/rados/troubleshooting/troubleshooting-mon.rst
jecluis: As mentioned to @tchaikov on IRC, it bothers me that the admin has to copy the osd data from each osd to the node where recovery is happening. This seems slow and unnecessary. We could simply grab the maps from `current/meta/` from all osds.

And, even though this is not something we expect to happen often (ideally, at all), I can imagine how an admin would be closing to tears if he had to copy every. single. osd. to get to the maps during DR. Poor admin. :(

tchaikov:
per the discussion with @jecluis 

* import from OSD's keyring also
* check the crc of osdmaps
* consider rsync'ing and using the meta directory only.

tchaikov:
changelog

- check the crc of osdmaps before putting them into store.db.

tchaikov:
test passed at http://pulpito.ceph.com/kchai-2016-09-15_15:48:11-rados:singleton-wip-kefu-testing---basic-mira/ .

@jecluis @athanatos @gregsfortytwo, thanks for your reviews.

is this changeset good to merge?

athanatos:
It looks good to me, it would probably be good to get a final ack from @gregsfortytwo .

PR #10936, opened 2016-08-31
add ldap auth custom search filter feature patch

haklein:
This patch allows to supply a custom search filter for ldap authentication. Currently rgw ldap auth code allows to limit users based on the search base. This might not be applicable for all environments. E.g. flat structures in smaller environments (ou=People, dc=corp, dc=com). Sometimes the structure is based on geo location and not on organisational units. In many cases one might not want to grant access to all accounts in a unit.

This PR implements a new "rgw search filter" option. These are the possible options:

1) parameter is left empty (default)

rgw LDAP auth will behave as usual, constructing its own search filter based on the "dnattr" option.

2) parameter is set to a "partial" filter like "objectclass=inetOrgperson"

The LDAP auth code takes the user specified partial filter and adds the user id specific filter to it (which is constructed as usual, by using the dn attr and the uid). The resulting filter will be "(&(<custom search filter>)(<dnattr>=<uid>))". So the above example becomes: "(&(objectClass=inetOrgPerson)(uid=hari))" (assuming a username of hari in the authentication request and an "uid" dnattr)

3) parameter is set to a "complete" filter with a special @USERNAME@ placeholder

An example would be "(&(uid=@USERNAME@)(memberOf=cn=ceph-users,ou=groups,dc=mycompany,dc=com))". The LDAP auth code will substitute the "@USERNAME@" placeholder with the uid of the auth request, leading to this search filter: "(&(uid=hari)(memberOf=cn=ceph-users,ou=groups,dc=mycompany,dc=com))"

### DIFF
--- a/src/common/config_opts.h
+++ b/src/common/config_opts.h
@@ -1347,6 +1347,8 @@ OPTION(rgw_ldap_dnattr, OPT_STR, "uid")
 OPTION(rgw_ldap_secret, OPT_STR, "/etc/openldap/secret")
 /* rgw_s3_auth_use_ldap  use LDAP for RGW auth? */
 OPTION(rgw_s3_auth_use_ldap, OPT_BOOL, false)
+/* rgw_ldap_searchfilter  LDAP search filter */
+OPTION(rgw_ldap_searchfilter, OPT_STR, "")
 
 OPTION(rgw_admin_entry, OPT_STR, "admin")  // entry point for which a url is considered an admin request
 OPTION(rgw_enforce_swift_acls, OPT_BOOL, true)
--- a/src/rgw/librgw.cc
+++ b/src/rgw/librgw.cc
@@ -494,12 +494,13 @@ namespace rgw {
     const string& ldap_uri = store->ctx()->_conf->rgw_ldap_uri;
     const string& ldap_binddn = store->ctx()->_conf->rgw_ldap_binddn;
     const string& ldap_searchdn = store->ctx()->_conf->rgw_ldap_searchdn;
+    const string& ldap_searchfilter = store->ctx()->_conf->rgw_ldap_searchfilter;
     const string& ldap_dnattr =
       store->ctx()->_conf->rgw_ldap_dnattr;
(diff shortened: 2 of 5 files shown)


ktdreyer:
Would you please create a Redmine ticket at http://tracker.ceph.com/ , so we make sure it gets backported to the jewel branch?

haklein:
@ktdreyer http://tracker.ceph.com/issues/17185

ktdreyer:
Thanks. Looks like there's two commits now? Mind squashing them into a single one, and adding a reference to that Redmine ticket in the commit log?

haklein:
@ktdreyer sure, all done

mattbenjamin:
I'm unable to compile test_rgw_ldap.cc, because the LDAPHelper ctor needs a search filter.  Is the following minimally  sufficient?  If you can address that and repush, it will be helpful!

[mbenjamin@lemon build]$ git diff
diff --git a/src/test/test_rgw_ldap.cc b/src/test/test_rgw_ldap.cc
index 8a08d8f..7a6c5a8 100644
--- a/src/test/test_rgw_ldap.cc
+++ b/src/test/test_rgw_ldap.cc
@@ -45,10 +45,11 @@ namespace {
   string ldap_binddn = "uid=admin,cn=users,cn=accounts,dc=rgw,dc=com";
   string ldap_bindpw = "supersecret";
   string ldap_searchdn = "cn=users,cn=accounts,dc=rgw,dc=com";
+  string ldap_searchfilter = "";
   string ldap_dnattr = "uid";
 
   rgw::LDAPHelper ldh(ldap_uri, ldap_binddn, ldap_bindpw, ldap_searchdn,
-                     ldap_dnattr);
+                     ldap_searchfilter, ldap_dnattr);
 
 } /* namespace */

mattbenjamin:
I have by-hand verified the following assertions:
1. non-regression w/current scope-search for user account match
2. correct authentication with a full search filter with @USERNAME@ (plus correct failure to authenticate when portions of the search filter do NOT match)

mattbenjamin:
merged from checkout, taking test_rgw_ldap.cc ctor argument fix approved by Harald (IRC)

PR #10937, opened 2016-08-31
Allow compressor build without YASM

dang:
The isa-l compressors need YASM to build.  Conditionalize on that.


cbodley:
i was confused by the existence of a `HAVE_GOOD_YASM_ELF64` variable, but that doesn't actually verify that yasm is present. looks good :+1:

PR #10938, opened 2016-08-31
os/bluestore: fix/improve csum error message

liewegas:


liewegas:
We still do'nt show the object offset, which is sort of lame, but it's nontrivial to add.

ifed01:
retest this please

ifed01:
LGTM

PR #10939, opened 2016-08-31
rgw: fix regression with handling double underscore

oritwas:
Fixes: http://tracker.ceph.com/issues/16856

### DIFF
--- a/src/rgw/rgw_op.cc
+++ b/src/rgw/rgw_op.cc
@@ -308,7 +308,8 @@ static int read_policy(RGWRados *store,
     obj.init_ns(bucket, oid, mp_ns);
     obj.set_in_extra_data(true);
   } else {
-    obj = rgw_obj(bucket, object);
+    obj = rgw_obj(bucket, object.name);
+    obj.set_instance(s->object.instance);
   }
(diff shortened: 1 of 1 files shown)


theanalyst:
tested this on master and the patch seems to work as intended, also solves all the other issues we ran int.  haven't tested on jewel yet.

src/rgw/rgw_op.cc
```diff
@@ -308,7 +308,8 @@ static int read_policy(RGWRados *store,
     obj.init_ns(bucket, oid, mp_ns);
     obj.set_in_extra_data(true);
   } else {
-    obj = rgw_obj(bucket, object);
+    obj = rgw_obj(bucket, object.name);
+    obj.set_instance(s->object.instance);
```
yehudasa: @oritwas while this might fix the issue, I'm having trouble understanding what's going on and why it should or shouldn't work. Sounds like there's a deeper problem that we should be fixing.

yehudasa: @oritwas reading the code further, I now understand the issue. We should get this fix in, and then rework rgw_obj and rgw_obj_key for future versions.

oritwas:
@yehudasa ping

PR #10941, opened 2016-08-31
install-deps: exit non-zero when we cannot match distro 

GregMeno:
exit 1 if we cannot find a match for linux distribution defined in /etc/os_release


tchaikov:
lgtm.

PR #10944, opened 2016-08-31
src/rgw/rgw_coroutine.cc: fix memory leak

bryongloden:
Greetings, on line number `611` of '[/src/rgw/rgw_coroutine.cc](https://github.com/ceph/ceph/blob/master/src/rgw/rgw_coroutine.cc#L611)' there is a memory leak, which is an error. The reason we should delete is that memory is a finite resource within our running programs. Sure in very short running simple programs, failing to delete memory won't have a noticeable effect. However on long running programs, failing to delete memory means we will be consuming a finite resource without replenishing it. Eventually it will run out and our program will abruptly crash. This is why we must free memory.

Signed-off-by: Bryon Gloden, CISSP® <cissp@bryongloden.com>

### DIFF
--- a/src/rgw/rgw_coroutine.cc
+++ b/src/rgw/rgw_coroutine.cc
@@ -608,6 +608,7 @@ int RGWCoroutinesManager::run(RGWCoroutine *op)
   }
   op->put();
 
+  delete stack;
   return r;
 }
 


mattbenjamin:
Something tells me that this change is incorrect.  However, handing it to the folks most knowledgeable about the coroutine lifecycle.

cbodley:
the lifetime of `RGWCoroutinesStack` is controlled by reference counting, so we shouldn't `delete` them manually

if you look at `RGWCoroutinesManager::run(list<RGWCoroutinesStack *>& stacks)`, you'll see that it calls `stack->put()` to drop the reference as each stack finishes. so i'm pretty confident that we're not leaking stacks here (unless you've seen valgrind complain about this)

cbodley:
i see from your commit message that this was caught by static analysis. i'd be open to changes that make this memory management more explicit and easier for tools to validate

for example, `run()` could take a list of `boost::intrusive_ptr<RGWCoroutineStack>` instead

bryongloden:
@cbodley, on the point of making it easier for our tools to validate memory management, suppressions can also be added directly in the code by adding comments that contain special keywords. Consider that the code readability is sacrificed a little after adding such comments. Then the `--inline-suppr` flag can be used by `cppcheck` to suppress the warning. This at least seems to be the approach taken by your colleagues in the [21](https://github.com/ceph/ceph/search?utf8=✓&q=cppcheck-suppress) cases I find.

cbodley:
@bryongloden in general i'd prefer to find ways to improve the code and make it easier to understand (both for tools and other developers) than to add suppressions

would you be willing to try making the reference counting of `RGWCoroutinesManager::run()` explicit with `boost::intrusive_ptr`, to see if cppcheck stops complaining? if that doesn't work, you're welcome to add whatever suppressions are needed here

bryongloden:
Hey there @cbodley! Just so we don't make any mistakes writing the code you have mind, why not just submit a patch to [our ceph fork](https://github.com/bryongloden/ceph). We'll then merge, clone, and rerun our tool.

We've already synced our fork of [ceph](https://github.com/bryongloden/ceph) repository to keep it up-to-date with the [upstream repository](https://github.com/ceph/ceph).

Working this way is easier for us because our computing system is Mac OS X 10.11.6 and we're having some trouble building ceph. We'd like to avoid a case where we implement the code you have in mind, push it upstream, and then discover it doesn't compile.

If you make the code change in our fork (or a branch of the upstream repository) and we rerun our tool on it, we'll both be working (or volunteering) on things we have expertise in. If not, we can continue to plug away on trying to make ceph compile (after all BSD underlies Mac OS X). Then when we're able to compile we could attempt to implement the code you have in mind. Although honestly, it's not immediately clear what the patch code would look like. 👍

cbodley:
@bryongloden i'm closing this pr since there's no actual memory leak, and we're not currently running cppcheck as part of our continuous integration. doing that is certainly open for discussion - if you have suggestions, you're welcome to bring them up on the ceph-devel mailing list. thanks!

PR #10946, opened 2016-09-01
mon/osdmap: load the specified version of the osdmap to the mon

renhwztetecs:
use "ceph-mon -i hostname --mkfs --monmap monmap_table --osdmap osdmap_table" command
to load the specified version of the osdmap to the mon.

The target:
repair the cluster when mon data is all lost,without the need to upgrade osdmap version

Before the reason:
After the original import, can lead to the version number osdmap reset to 1
So that the osd hold high osdmap table version number and mon osdmap version number

Signed-off-by: huanwen ren ren.huanwen@zte.com.cn

### DIFF
--- a/src/mon/OSDMonitor.cc
+++ b/src/mon/OSDMonitor.cc
@@ -80,7 +80,9 @@ OSDMonitor::OSDMonitor(CephContext *cct, Monitor *mn, Paxos *p, const string& se
    full_osd_cache(g_conf->mon_osd_cache_size),
    thrash_map(0), thrash_last_up_osd(-1),
    op_tracker(cct, true, 1)
-{}
+{
+  create_initial_version = 0;
+}
 
 bool OSDMonitor::_have_pending_crush()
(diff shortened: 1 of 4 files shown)


renhwztetecs:
ping @tchaikov @jecluis 
please help me reviews

src/mon/OSDMonitor.cc
```diff
@@ -80,7 +80,9 @@ OSDMonitor::OSDMonitor(CephContext *cct, Monitor *mn, Paxos *p, const string& se
    full_osd_cache(g_conf->mon_osd_cache_size),
    thrash_map(0), thrash_last_up_osd(-1),
    op_tracker(cct, true, 1)
-{}
+{
```
amitkumar50: @renhwztetecs Please keep PR title within 50 characters

liewegas:
I think this is superceded by @tchaikov 's mon recovery tool

PR #10948, opened 2016-09-01
osd: peering optimizations

dragonylffly:
This patch optimizes peering by speculatively prefetching logs in GetInfo state. 
The osd set to contact consists of the current up+acting osds as well as the ones 
from the latest interval. It thus hopefully saves the time to request authoritative log 
in GetLog and request logs in GetMissing.

### DIFF
--- a/src/osd/PG.cc
+++ b/src/osd/PG.cc
@@ -7377,6 +7377,13 @@ PG::RecoveryState::GetLog::GetLog(my_context ctx)
     return;
   }
 
+  if (peer_missing_requested.count(auth_log_shard)) {
+    if (msgs[auth_log_shard]) {
+      msg = msgs[auth_log_shard];
+      post_event(GotLog());
+    }
+    return;
+  }
   // how much log to request?
   eversion_t request_log_from = pg->info.last_update;
   assert(!pg->actingbackfill.empty());
--- a/src/osd/PG.cc
+++ b/src/osd/PG.cc
@@ -7126,6 +7126,8 @@ void PG::RecoveryState::GetInfo::get_infos()
 {
   PG *pg = context< RecoveryMachine >().pg;
   unique_ptr<PriorSet> &prior_set = context< Peering >().prior_set;
+  set<pg_shard_t> &peer_missing_requested =
+    context< Peering >().peer_missing_requested;
 
   pg->blocked_by.clear();
   for (set<pg_shard_t>::const_iterator it = prior_set->probe.begin();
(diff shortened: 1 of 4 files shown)


dragonylffly:
@liewegas @athanatos Your comments appreciated

src/osd/osd_types.h
```diff
@@ -2315,13 +2315,15 @@ struct pg_query_t {
     LOG = 1,
     MISSING = 4,
     FULLLOG = 5,
+    ALL = 6,
```
athanatos: They use different messages, but the MOSDPGLog message returned from FULLLOG request includes the info.  I'd prefer you not change the messages if possible since it would make handling and testing the mixed cluster case much harder.

src/osd/PG.cc
```diff
@@ -7818,14 +7896,19 @@ PG::PriorSet::PriorSet(bool ec_pool,
...
+        shard_id_t(i) : shard_id_t::NO_SHARD));
   }
   // It may be possible to exclude the up nodes, but let's keep them in
   // there for now.
   for (unsigned i = 0; i < up.size(); i++) {
     if (up[i] != CRUSH_ITEM_NONE)
       probe.insert(pg_shard_t(up[i], ec_pool ? shard_id_t(i) : shard_id_t::NO_SHARD));
+      request_log.insert(pg_shard_t(up[i], ec_pool ?
+        shard_id_t(i) : shard_id_t::NO_SHARD));
   }
```
athanatos: This is concerning.  Currently, in the common case where one osd went down, the other osds in the acting set won't need to send any actual log entries.  With this change, every replica from every osd which shares a pg with the down osd will have to send the *entire* log to the primary.  That seems unlikely to be a good tradeoff.  Is there a way to guess at how much log will be required and only request that much?

athanatos:
At a very high level, this seems like it's heading in the right direction.  My main question at this point is: can you characterize how much more data (log entries) will need to be sent with this branch vs current master?

liewegas:
I'm worried that sending logs is expensive, and we'd only want to do this in select cases.  Is it bounding the portion of the log we request?

It seems like the most promising way to optimize peering is to focus on the common cases, e.g., where the primary is the same and the replica set changes (approach: re-use state previous previous interval and skip most of peering entirely) or the new primary is a former replica (approach: only request log entries from our last_update, since we know we aren't divergent).

tchaikov:
@badone

dachary:

PR #10949, opened 2016-09-01
preforker: prevent call to 'write' on an fd that was already closed, …

avnerbh:
…plus minor code fixups

upon ceph_mon parent exiting it calls prefork.exit(err) that was resulting in
writing the exit status to fd[1] whis was previously closed in this process.
for fixing it, I guarded the call to 'signal_exit(r)' with 'if (is_child())'

on the way, I provided 2 minor code fixups in this file

Signed-off-by: Avner BenHanoch <avnerb@mellanox.com>

### DIFF
--- a/src/common/Preforker.h
+++ b/src/common/Preforker.h
@@ -39,7 +39,7 @@ public:
 
   int prefork(std::string &err) {
     assert(!forked);
-    int r = socketpair(AF_UNIX, SOCK_STREAM, 0, fd);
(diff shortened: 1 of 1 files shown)


tchaikov:
tested at http://pulpito.ceph.com/kchai-2016-09-10_15:46:37-rados-wip-kefu-testing---basic-mira/408471/

the failed test of http://pulpito.ceph.com/kchai-2016-09-10_15:46:37-rados-wip-kefu-testing---basic-mira/408471/ is irrelevant.

liewegas:
ok to merge?

PR #10956, opened 2016-09-02
cmake: pass -DINTEL* to gf-complete cflags

tchaikov:
* in addition to the -march=foobar, we also need to pass -DFOOBAR to
  gf-complete sources' cflags, so that the SIMD instructions can be
  generated as expected.
* also extract the compiler support for instruction extensions detections
  into SIMDExt.cmake.
* and only check the compiler support if CMAKE_SYSTEM_PROCESSOR matches
  with the instruction-set/arch to be checked.

Signed-off-by: tone.zhang <tone.zhang@linaro.org>


dachary:
<pre>
</pre>

tchaikov:
@dachary thanks for your review. i just realized that we should `include(SIMDExt)` in `erasure/CMakeLists.txt` instead of in `erasure/jerasure/CMakeLists.txt`, so the list the plugin flavours can be initialized correctly.

PR #10960, opened 2016-09-02
src/CMakeLists.txt: Python 3 is not hard requirement.

wjwithagen:

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -200,8 +200,8 @@ endif()
 # Python stuff
 find_package(PythonInterp 2 REQUIRED)
 find_package(PythonLibs 2 REQUIRED)
-find_package(Python3Interp 3 REQUIRED)
-find_package(Python3Libs 3 REQUIRED)
+find_package(Python3Interp 3 QUIET)
+find_package(Python3Libs 3 QUIET)
 
 if(HAVE_XIO)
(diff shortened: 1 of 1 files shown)


wjwithagen:
@tchaikov 
This PR was explicitly submitted to overcome the problem that you referred to.
I run/ran into Jenkins tests that failed due to missing python3, and I discussed it on the dev-list
and python3 is not yet a hard requirement. So it should not fail if it is missing.

wjwithagen:
@tchaikov 
This is an example Jenkins-run:
https://jenkins.ceph.com/job/ceph-pull-requests/11118/consoleFull#104678606520215f5e-12e1-4837-8c35-e0ce67104d63

jdurgin:
Python 3 isn't a hard requirement, but I think it'll need more conditionals (e.g. a cmake parameter to disable python 3). Otherwise it'll fail later in the make install or cython build stage, or at the very least during the rpm and deb builds (hence http://tracker.ceph.com/issues/17103).

tchaikov:
@wjwithagen i will give it a shot to add an option for generating python2 and python bindings tmr.

wjwithagen:
@jdurgin @tchaikov 
Well it starts with the fact that there are Jenkins instances that do not have python3, and thus the first build already fails... So with that it is never going to get to QA let alone RPM builds.
But then again I have no experience in the path after Jenkins, so I'm happy to see someone else take a shot at it

tchaikov:
#11015 supersedes this change, so i am closing it.

PR #10962, opened 2016-09-02
python-rados: add methods to get/set iterator position.

iain-buclaw-sociomantic:
This change adds python API for the following C methods:
* rados_nobjects_list_get_pg_hash_position
* rados_nobjects_list_seek

Signed-off-by: Iain Buclaw <iain.buclaw@sociomantic.com>

### DIFF
--- a/src/pybind/rados/rados.pyx
+++ b/src/pybind/rados/rados.pyx
@@ -194,6 +194,8 @@ cdef extern from "rados/librados.h" nogil:
 
     int rados_nobjects_list_open(rados_ioctx_t io, rados_list_ctx_t *ctx)
     int rados_nobjects_list_next(rados_list_ctx_t ctx, const char **entry, const char **key, const char **nspace)
+    int rados_nobjects_list_get_pg_hash_position(rados_list_ctx_t ctx)
+    int rados_nobjects_list_seek(rados_list_ctx_t ctx, uint32_t pos)
     void rados_nobjects_list_close(rados_list_ctx_t ctx)
 
     int rados_ioctx_snap_rollback(rados_ioctx_t io, const char * oid, const char * snapname)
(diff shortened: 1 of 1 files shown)


dachary:

dachary:

tchaikov:
@jdurgin i think it's okay to merge this PR, what do you think?

src/pybind/rados/rados.pyx
```diff
@@ -194,6 +194,8 @@ cdef extern from "rados/librados.h" nogil:
 
     int rados_nobjects_list_open(rados_ioctx_t io, rados_list_ctx_t *ctx)
     int rados_nobjects_list_next(rados_list_ctx_t ctx, const char **entry, const char **key, const char **nspace)
+    int rados_nobjects_list_get_pg_hash_position(rados_list_ctx_t ctx)
+    int rados_nobjects_list_seek(rados_list_ctx_t ctx, uint32_t pos)
```
tchaikov: the declarations does not match their counterparts in librados.h.

iain-buclaw-sociomantic: I take it you mean `s/int/uint32_t/`?

tchaikov: @iain-buclaw-sociomantic exactly.

iain-buclaw-sociomantic:
> Please wrap the get_cursor and seek_cursor methods instead; they're better than teh pg hash position ones (which we're likely to deprecate and remove).

OK, what's the difference?  And how would you, say, seek to a random position, or jump to the middle using these?  I suppose that's what `rados_object_list_slice` is for, but that's part of the old object listing.

iain-buclaw-sociomantic:
My last question is still open, though I've not had much use for it now as am using `rados_object_list_slice` to cut my way through a pool nowadays (dropped Python bindings for C++).

iain-buclaw-sociomantic:
@liewegas - I'm still not sure how exposing only get_cursor and seek_cursor alone to the python bindings could be used in any meaningful way.  Not without adding a third method that returns an array of cursor start/end pairs taken from `rados_object_list_slice()`.

As I said, my use-case was being able to jump to a position in the object iterator to parallelize work (pool sizes we deal with have over 650 million objects inside, iterating from 1 .. 650M is no good).

liewegas:
@iain-buclaw-sociomantic Ah, sorry, I misunderstood then.  You're right that you need the slice method too in order to craft cursors for parallelizing work.

PR #10963, opened 2016-09-02
os/bluestore: shard extent map

liewegas:
Rewrote much of the persistence of onode metadata.  The
    highlights:
    
     - extents and blobs stored together (the blob with the
       first referencing extent).
     - extents sharded across multiple k/v keys
     - if a blob if referenced from multiple blobs, it's
       stored in the onode key (called a "spanning blob").
     - when we clone a blob we copy the metadata, but mark
       it shared and put (just) the ref_map on the underlying
       blocks in a shared_blob key.  at this point we also
       assign a globally unique id (sbid = shared blob id)
       so the key has a unique name.
     - we instantiate a SharedBlob in memory regardless of
       whether we need to load the ref_map (which is only
       needed for deallocations!).  the BufferSpace is
       attached to this SharedBlob so we get unified caching
       across clones.


src/test/objectstore/store_test.cc
```diff
@@ -3560,6 +3560,16 @@ class SyntheticWorkloadState {
     return status;
   }
 
+  void fsck() {
+    Mutex::Locker locker(lock);
+    EnterExit ee("fsck");
+    while (in_flight)
+      cond.Wait(lock);
+    store->umount();
+    store->fsck();
```
ifed01: We have fsck_on_(u)mount set to true for this test suite (see main func) hence there is no need to call fsck directly. That's just a waste of time...

liewegas:
This passes tests, except for the bitmap granularity issue in #10999 that is worked around in that PR.  I suggest we merge that one too (with an interim fix) until we do something more clever with min_alloc_size

liewegas:
@chhabaramesh

src/os/bluestore/BlueStore.cc
```diff
@@ -251,34 +252,18 @@ static void get_coll_key_range(const coll_t& cid, int bits,
...
+  _key_encode_u64(sbid, key);
 }
 
-static int get_key_bnode(const string& key, shard_id_t *shard,
-			 int64_t *pool, uint32_t *hash)
+static int get_key_shared_blob(const string& key, uint64_t *sbid)
 {
   const char *p = key.c_str();
-  if (key.length() < 2 + 8 + 4)
+  if (key.length() < 8)
```
allensamuels: 8 --- ugh. Magic number.

src/os/bluestore/BlueStore.cc
```diff
@@ -391,11 +376,35 @@ static int get_key_object(const string& key, ghobject_t *oid)
...
+{
+  key->clear();
+  key->append(onode_key);
+  _key_encode_u32(offset, key);
+  key->append("x");
+}
+
+int get_key_extent_shard(const string& key, string *onode_key, uint32_t *offset)
+{
+  assert(key.size() > 5);
```
allensamuels: more magic numbers. (5)

src/os/bluestore/BlueStore.cc
```diff
@@ -391,11 +376,35 @@ static int get_key_object(const string& key, ghobject_t *oid)
...
+  const char *p = key.data() + key.size() - 5;
+  p = _key_decode_u32(p, offset);
+  return 0;
+}
+
+static bool is_extent_shard_key(const string& key)
+{
+  return key[key.size()-1] == 'x';
```
allensamuels: Let's make a global identifier out of 'x'

src/os/bluestore/BlueStore.cc
```diff
@@ -1177,81 +1246,734 @@ void BlueStore::Blob::discard_unallocated()
...
+  uint64_t length)
+{
+  ref_map.get(offset, length);
+}
+
+bool BlueStore::Blob::put_ref(
+  uint64_t offset,
+  uint64_t length,
+  uint64_t min_release_size,
+  vector<bluestore_pextent_t> *r)
```
allensamuels: If it's not optional, let's pass it by reference.

src/os/bluestore/BlueStore.cc
```diff
@@ -1177,81 +1246,734 @@ void BlueStore::Blob::discard_unallocated()
...
+    while (r_off >= q->length) {
+      vb.add(q->offset, q->length);
+      r_off -= q->length;
+      ++q;
+      assert(q != b.extents.end());
+    }
+    while (r_len > 0) {
+      uint64_t l = MIN(r_len, q->length - r_off);
+      if (q->is_valid()) {
+	r->push_back(bluestore_pextent_t(q->offset + r_off, l));
```
allensamuels: emplace_back.

src/os/bluestore/BlueStore.cc
```diff
@@ -1177,81 +1246,734 @@ void BlueStore::Blob::discard_unallocated()
...
-#define dout_prefix *_dout << "bluestore.blobmap(" << this << ") "
+#define dout_prefix *_dout << "bluestore.extentmap(" << this << ") "
+
+bool BlueStore::ExtentMap::update(Onode *o, KeyValueDB::Transaction t,
+				  bool force)
+{
+  if (o->onode.extent_map_shards.empty()) {
+    if (inline_bl.length() == 0) {
+      unsigned n;
+      if (encode_some(0, 0xffffffff, inline_bl, &n)) {
```
allensamuels: Another odd magic constant.

src/os/bluestore/BlueStore.cc
```diff
@@ -1177,81 +1246,734 @@ void BlueStore::Blob::discard_unallocated()
...
+}
+
+void BlueStore::ExtentMap::reshard(Onode *o)
+{
+  // un-span all blobs
+  auto p = spanning_blob_map.begin();
+  while (p != spanning_blob_map.end()) {
+    auto n = p;
+    ++n;
+    spanning_blob_map.erase(*p);
```
allensamuels: auto n = spanning_blob_map.erase(*p); // eliminates the ugly auto n =p; ++n;

src/os/bluestore/BlueStore.cc
```diff
@@ -1177,81 +1246,734 @@ void BlueStore::Blob::discard_unallocated()
...
+	}
+	sp = next;
+      }
+      dout(20) << __func__ << " old shard end 0x" << std::hex << shard_end
+	       << std::dec << dendl;
+    }
+    if (estimate && estimate + extent_avg > target) {
+      // new shard
+      if (offset == 0) {
+	new_shard_info.push_back(bluestore_onode_t::shard_info());
```
allensamuels: emplace_back might be clearer here as well as a few lines below.

src/os/bluestore/BlueStore.cc
liewegas: We use pointers for output arguments (based on google c++ style)

src/os/bluestore/BlueStore.cc
```diff
@@ -1177,81 +1246,734 @@ void BlueStore::Blob::discard_unallocated()
...
+  uint64_t pos = 0;
+  uint64_t prev_len = 0;
+  for (auto p = start;
+       p != extent_map.end() && p->logical_offset < end;
+       ++p, ++n) {
+    unsigned blobid;
+    bool include_blob = false;
+    if (p->blob->id >= 0) {
+      blobid = p->blob->id << 4;
+      blobid |= 8;
```
allensamuels: More magic numbers and magic shift values :)

liewegas:
Thanks @allensamuels, fixed that.  Also rolled the allocator granularity bug workaround into this PR for now so that the tests pass with all the new fsck's.  It's passing the full test suite for me; I think it's ready to merge.

[==========] 61 tests from 1 test case ran. (2841968 ms total)
[  PASSED  ] 61 tests.

src/os/bluestore/BlueStore.cc
```diff
@@ -1275,51 +1997,77 @@ void BlueStore::Onode::flush()
...
   : store(ns),
-    cache(cs),
-    cid(c),
+    cache(c),
+    cid(cid),
     lock("BlueStore::Collection::lock", true, false),
     exists(true),
-    bnode_set(MAX(16, g_conf->bluestore_onode_cache_size / 128)),
-    onode_map(cs)
+    shared_blob_set(MAX(16, g_conf->bluestore_onode_cache_size / 4)),
```
allensamuels: More magic numbers -- that are different from the old magic numbers :)

src/os/bluestore/BlueStore.cc
```diff
@@ -6987,52 +7841,72 @@ int BlueStore::_clone(TransContext *txc,
...
+	const bluestore_blob_t& blob = e.blob->get_blob();
+	// make sure it is shared
+	if (!blob.is_shared()) {
+	  e.blob->dirty_blob().sbid = _assign_blobid(txc);
+	  c->make_blob_shared(e.blob);
+	  dirtied_oldo = true;  // fixme: overkill
+	} else if (!e.blob->shared_blob->loaded) {
+	  c->load_shared_blob(e.blob->shared_blob);
+	}
+	nb = new Blob;
```
allensamuels: This reads dangerously, because nb is sometimes used for existing blobs and sometimes for a new blob. Perhaps a helper function that creates blobs and returns the reference to the created blob?

src/os/bluestore/BlueStore.h
```diff
@@ -372,192 +463,169 @@ class BlueStore : public ObjectStore,
...
+    friend bool operator==(const Extent &a, const Extent &b) {
+      return a.logical_offset == b.logical_offset;
     }
 
-    void claim(BlobRef b) {
-      assert(b->id == 0);
-      b->id = get_new_id();
-      b->get();
-      blob_map.insert(*b);
+    bool blob_spans(uint32_t o, uint32_t l) {
```
allensamuels: This doesn't read right by itself -- it might be the right logic in the context in which it's used. But it's not a standalone overlap (span) test, if this blob is totally to the "left" of the parameter range, it'll still return true. So either the name is misleading or something....

PR #10965, opened 2016-09-02
common: only call crypto::init once per CephContext

cbodley:
Fixes: http://tracker.ceph.com/issues/17205


cbodley:
@oritwas could you please review?

PR #10966, opened 2016-09-02
os/bluestore/BlueFS: factor unflushed log into runway calculation

liewegas:
Include the size of the in-memory buffer we haven't yet flushed
into the runway calculation when deciding whether to allocate
more log space.

Reported-by: Somnath Roy <somnath.roy@sandisk.com>


liewegas:
@somnathr

varadakari:
LGTM

somnathr:
Looks good..
BTW, with bigger runway I was able to run more than 7 hours and then I got 2 more different assert().
Trying to analyze, will send you a mail soon..

PR #10972, opened 2016-09-03
CmakeLists.txt: use LIB_RESOLV instead of resolv.

wjwithagen:
- FreeBSD does not have a separate libresolv, but keeps it in
   libc

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -458,7 +458,7 @@ endif(ENABLE_SHARED)
 add_library(common STATIC ${libcommon_files}
   $<TARGET_OBJECTS:mon_common_objs>
   $<TARGET_OBJECTS:common_mountcephfs_objs>)
-target_link_libraries(common json_spirit erasure_code rt resolv
+target_link_libraries(common json_spirit erasure_code rt ${LIB_RESOLV}
   ${Boost_LIBRARIES} ${BLKID_LIBRARIES} ${Backtrace_LIBRARIES}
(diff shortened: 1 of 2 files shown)


wjwithagen:
@tchaikov 
Pushed an extra commit to fix the Python problem.
I can drop that once a fix in master is available

tchaikov:
jenkins, retest this please

wjwithagen:
@tchaikov 
```
The following tests FAILED:
	147 - unittest_http_manager (Timeout)
```
And I'm having this also in other jenkins runs that are totally not related.

src/CMakeLists.txt
```diff
@@ -200,8 +200,8 @@ endif()
 # Python stuff
 find_package(PythonInterp 2 REQUIRED)
 find_package(PythonLibs 2 REQUIRED)
-find_package(Python3Interp 3 REQUIRED)
-find_package(Python3Libs 3 REQUIRED)
+find_package(Python3Interp 3 QUIET)
+find_package(Python3Libs 3 QUIET)
```
tchaikov: could you drop this commit? it breaks the build if python3 is missing, see src/pybind/CMakeLists.txt.

I think it is the other way around.
With QUIET the absence of python3 stuff is sillently ignored.

If I test with REQUIRED, things seriously blow up in my face on some Jenkins runs where there is python3 to be installed. And things really stop during CMake processing.
So I added this to actually get some testing. (Now it stops on the last test due to timeout, and unrelated to this commit.)
But perhaps somebody fixed the python stuff in the mean time.
I'll drop this extra commit.

PR #10976, opened 2016-09-04
msg/simple/Pipe.cc: set socket reuse option

wjwithagen:
- This reduces the retries on Pipe connects that I'm seeing in the logs

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -200,8 +200,8 @@ endif()
 # Python stuff
 find_package(PythonInterp 2 REQUIRED)
 find_package(PythonLibs 2 REQUIRED)
-find_package(Python3Interp 3 REQUIRED)
-find_package(Python3Libs 3 REQUIRED)
+find_package(Python3Interp 3 QUIET)
+find_package(Python3Libs 3 QUIET)
 
 if(HAVE_XIO)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -I${XIO_INCLUDE_DIR}")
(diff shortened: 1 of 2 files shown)


gregsfortytwo:
I don't have any experience with this option and not much depth in networking/IP stacks, but I don't see any way this can work based on http://hea-www.harvard.edu/~fine/Tech/addrinuse.html

In particular, "Strategies for Avoidance"->"SO_REUSEADDR" says
>Oddly, using SO_REUSEADDR can actually lead to more difficult "address already in use" errors. SO_REUSADDR permits you to use a port that is stuck in TIME_WAIT, but you still can not use that port to establish a connection to the last place it connected to. What? Suppose I pick local port 1010, and connect to foobar.com port 300, and then close locally, leaving that port in TIME_WAIT. I can reuse local port 1010 right away to connect to anywhere except for foobar.com port 300.

Which sounds sort of like the bug you reported on the mailing list, too.
You might try tracking down the connection retries a little more, depending on what context they're in. We induce a lot of failures deliberately in some tests, or maybe your environment has a restricted set of socket IDs (we crank the available numbers way, way up...).

wjwithagen:
@gregsfortytwo 
The piece you quote does have  a ring to it. Kernels tend to store connections as tuples:
<src:srcport;dst;dstport> 
And what you quote claims that this tuple is going to linger around until TIME_WAIT has passed.

I know that this can lead to DOS attacks where one can run out of table space.
But I would expect modern versions of OSes to work around these problems.

None the less: I'm going to have some more fun digging in the code.

PR #10977, opened 2016-09-05
bug fix for install-deps.sh

jazeltq:
Right now, the install-deps.sh can not run on centos, so fix
it.


liewegas:
Fixed in master

PR #10978, opened 2016-09-05
os/bluestore: Hint based allocation in bitmap Allocator

chhabaramesh:
This pull request has following changes:
1. Hint based allocation in BitMapAllocator.
2. Unit Test cases related to that.
3. Cleanup of wrap argument that is no more in use.

### DIFF
--- a/src/test/objectstore/Allocator_test.cc
+++ b/src/test/objectstore/Allocator_test.cc
@@ -45,7 +45,7 @@ TEST_P(AllocTest, test_alloc_init)
   alloc->shutdown(); 
   blocks = BitMapZone::get_total_blocks() * 2;
   init_alloc(blocks, 1);
-  ASSERT_EQ(alloc->get_free(), 0);
+  ASSERT_EQ(alloc->get_free(), (uint64_t) 0);
 }
 
 TEST_P(AllocTest, test_alloc_min_alloc)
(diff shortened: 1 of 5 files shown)


chhabaramesh:
@liewegas , Please review this.

src/test/objectstore/Allocator_test.cc
```diff
@@ -187,6 +187,49 @@ TEST_P(AllocTest, test_alloc_failure)
   }
 }
 
+TEST_P(AllocTest, test_alloc_hint_bmap)
+{
+  if (GetParam() == std::string("stupid")) {
```
liewegas: != std::string("bitmap") ?

liewegas:
@chhabaramesh I haven't been following bitmap allocator internals.. is there someone else that can review?

chhabaramesh:
@liewegas , I can explain you it to you in short call ? Or Maybe @ifed01 or Manavalan can review?

But this change is quite simple, just makes sure that search of bit start from a given hint.

src/test/objectstore/Allocator_test.cc
chhabaramesh: The stupid allocator was not passing all new hint test cases. So added this to bypass it for stupid allocator.

liewegas:
I see where the hint is used, but I don't see where the search loops around to the beginning if it doesn't find anything > hint (and, ideally, only searches back up to the start point).

chhabaramesh:
@liewegas The iterator BmapEntityListIter takes a wrap argument and that is true only for children of root nodes.

1555 int64_t BitAllocator::alloc_blocks_dis_int(bool wait, int64_t num_blocks,
1556            int64_t hint, int64_t area_blk_off, ExtentList *block_list)
1557 {
1558   return alloc_blocks_dis_int_work(wait, true, num_blocks, hint,
1559                      area_blk_off, block_list);
1560 }

At all other levels it should be false. So it will start at a child at root level and wrap around and end at same child.

I think there is a type bug at:

 959 int64_t BitMapAreaIN::alloc_blocks_dis_int(bool wait, int64_t num_blocks,
 960            int64_t hint, int64_t area_blk_off, ExtentList *block_list)
 961 {
 962   return alloc_blocks_dis_int_work(wait, true, num_blocks, hint,
 963                      area_blk_off, block_list);
 964 }

The wrap ( second ) argument must be false here :(.

chhabaramesh:
@liewegas , Fixed the typo bug mentioned above and updated. Please check again.

PR #8204, opened 2016-03-18
librbd: enable/disable image mirroring automatically for pool mode

rjfd:
Fixes #15143

Ticket url: http://tracker.ceph.com/issues/15143

PR #10980, opened 2016-09-05
rbd-mirror: replicate dynamic feature updates

trociny:
Fixes: http://tracker.ceph.com/issues/16213


src/librbd/operation/EnableFeaturesRequest.cc
```diff
@@ -0,0 +1,444 @@
...
+template <typename I>
+void EnableFeaturesRequest<I>::send_op() {
+  I &image_ctx = this->m_image_ctx;
+  CephContext *cct = image_ctx.cct;
+  assert(image_ctx.owner_lock.is_locked());
+
+  ldout(cct, 20) << this << " " << __func__ << ": features=" << m_features
+		 << dendl;
+
+  if (this->is_canceled()) {
```
dillaman: Minor: no need to check for this -- it's only used for long-running operations (e.g. resize)

src/librbd/operation/DisableFeaturesRequest.h
```diff
@@ -0,0 +1,161 @@
...
+   *
+   * @verbatim
+   *
+   * <start>
+   *    |
+   *    v
+   * STATE_BLOCK_WRITES
+   *    |
+   *    v
+   * STATE_APPEND_OP_EVENT (skip if journaling
```
dillaman: I would move this as far down as possible since any error code will result in a mirroring split-brain (e.g. attempting to disable a feature that wasn't enabled).

src/librbd/operation/DisableFeaturesRequest.cc
```diff
@@ -0,0 +1,655 @@
...
+    lderr(cct) << "failed to lock image: " << cpp_strerror(*result) << dendl;
+    return handle_finish(*result);
+  } else if (m_acquired_lock && (image_ctx.exclusive_lock == nullptr ||
+				 !image_ctx.exclusive_lock->is_lock_owner())) {
+    lderr(cct) << "failed to acquire exclusive lock" << dendl;
+    *result = -EROFS;
+    return handle_finish(*result);
+  }
+
+  if ((m_features & ~image_ctx.features) != 0) {
```
dillaman: This check can only be invoked from the synchronous "Operations::update_features" function since we need to preserve idempotency on image watcher RPC or journal event replay. Same comment for enabling features.

trociny:
@dillaman I addressed your comments and fixed all known issues. I think now it is good enough for review.

I plan to add mock tests for Enable/DisableFeatureRequest operations.

src/librbd/CMakeLists.txt
```diff
@@ -34,10 +34,15 @@ set(librbd_internal_srcs
   exclusive_lock/ReleaseRequest.cc
   exclusive_lock/StandardPolicy.cc
   image/CloseRequest.cc
+  image/CreateObjectMapRequest.cc
```
dillaman: Minor: move to object_map/CreateRequest.cc

src/librbd/CMakeLists.txt
```diff
@@ -34,10 +34,15 @@ set(librbd_internal_srcs
...
   image/CreateRequest.cc
+  image/MirrorDisableRequest.cc
+  image/MirrorEnableRequest.cc
   image/OpenRequest.cc
   image/RefreshParentRequest.cc
   image/RefreshRequest.cc
+  image/RemoveObjectMapRequest.cc
```
dillaman: Minor: move to object_map/RemoveRequest.cc

src/librbd/CMakeLists.txt
```diff
@@ -34,10 +34,15 @@ set(librbd_internal_srcs
   exclusive_lock/ReleaseRequest.cc
   exclusive_lock/StandardPolicy.cc
   image/CloseRequest.cc
+  image/CreateObjectMapRequest.cc
   image/CreateRequest.cc
+  image/MirrorDisableRequest.cc
```
dillaman: Minor: move to mirror/Disable/EnableRequest.cc

src/librbd/ImageWatcher.cc
```diff
@@ -943,6 +955,28 @@ bool ImageWatcher<I>::handle_payload(const RenamePayload& payload,
...
+  RWLock::RLocker owner_locker(m_image_ctx.owner_lock);
+  if (m_image_ctx.exclusive_lock != nullptr) {
+    int r;
+    if (m_image_ctx.exclusive_lock->accept_requests(&r)) {
+      ldout(m_image_ctx.cct, 10) << this << " remote update_features request: "
+                                 << payload.features << " "
+                                 << (payload.enabled ? "enabled" : "disabled")
+                                 << dendl;
+
+      m_image_ctx.operations->execute_update_features(payload.features,
```
dillaman: Minor: move "payload.features" down to next line or align next params

src/librbd/Journal.h
```diff
@@ -105,6 +105,10 @@ class Journal {
   static int get_tag_owner(ImageCtxT *image_ctx, std::string *mirror_uuid);
   static int get_tag_owner(librados::IoCtx& io_ctx, std::string& image_id,
                            std::string *mirror_uuid);
+  static void is_tag_owner(ImageCtxT *image_ctx, ContextWQ *work_queue,
+                           bool *is_tag_owner, Context *on_finish);
+  static void get_tag_owner(ImageCtxT *image_ctx, ContextWQ *work_queue,
```
dillaman: This doesn't seem to be used

src/librbd/Journal.h
```diff
@@ -105,6 +105,10 @@ class Journal {
   static int get_tag_owner(ImageCtxT *image_ctx, std::string *mirror_uuid);
   static int get_tag_owner(librados::IoCtx& io_ctx, std::string& image_id,
                            std::string *mirror_uuid);
+  static void is_tag_owner(ImageCtxT *image_ctx, ContextWQ *work_queue,
```
dillaman: Looks like ```work_queue``` is just extracted from ```image_ctx``` -- do you need to pass it in?

src/librbd/Journal.cc
```diff
@@ -471,6 +624,31 @@ int Journal<I>::get_tag_owner(IoCtx& io_ctx, std::string& image_id,
 }
 
 template <typename I>
+void Journal<I>::is_tag_owner(I *image_ctx, ContextWQ *work_queue,
```
dillaman: Did you still run into an issue reusing journal::OpenRequest/journal::CloseRequest?

src/librbd/Operations.cc
```diff
@@ -1217,6 +1219,79 @@ void Operations<I>::execute_snap_set_limit(const uint64_t limit,
...
+  if ((enabled && (features & RBD_FEATURES_MUTABLE) != features) ||
+      (!enabled && (features & disable_mask) != features)) {
+    lderr(cct) << "cannot update immutable features" << dendl;
+    return -EINVAL;
+  }
+  if (features == 0) {
+    lderr(cct) << "update requires at least one feature" << dendl;
+    return -EINVAL;
+  }
+  if (enabled && (features & m_image_ctx.features) != 0) {
```
dillaman: Minor: hold ImageCtx::snap_lock while testing m_image_ctx.features

src/librbd/image/CreateObjectMapRequest.cc
```diff
@@ -0,0 +1,107 @@
...
+Context *CreateObjectMapRequest<I>::handle_object_map_resize(int *result) {
+  CephContext *cct = m_image_ctx->cct;
+  ldout(cct, 20) << __func__ << ": r=" << *result << dendl;
+
+  {
+    Mutex::Locker locker(m_lock);
+    assert(m_ref_counter > 0);
+    m_ref_counter--;
+
+    if (*result < 0 && *result != -ENOENT) {
```
dillaman: Can -ENOENT be returned? If not, you can use a C_Gather to collect the per-snap resize requests and avoid the need for the lock

src/librbd/operation/EnableFeaturesRequest.cc
```diff
@@ -0,0 +1,435 @@
...
+    send_append_op_event();
+    return;
+  }
+
+  Context *ctx = create_context_callback<
+    EnableFeaturesRequest<I>,
+    &EnableFeaturesRequest<I>::handle_enable_mirror_image>(this);
+
+  image::MirrorEnableRequest<I> *req =
+    image::MirrorEnableRequest<I>::create(&image_ctx.md_ctx, image_ctx.id, ctx);
```
dillaman: Image is already open -- why do you need to re-open the image?

src/librbd/operation/DisableFeaturesRequest.cc
```diff
@@ -0,0 +1,648 @@
...
+    lderr(cct) << "failed to acquire exclusive lock" << dendl;
+    *result = -EROFS;
+    return handle_finish(*result);
+  }
+
+  m_features &= image_ctx.features;
+  m_new_features = image_ctx.features & ~m_features;
+  m_features_mask = m_features;
+
+  if ((m_features & RBD_FEATURE_EXCLUSIVE_LOCK) != 0) {
```
dillaman: Minor: probably good to validate the settings before making any changes.

src/librbd/operation/EnableFeaturesRequest.cc
trociny: Interesting question :-) I do this just because it was in the initial (sync) implementation.

https://github.com/ceph/ceph/pull/8204/commits/ba6c0c9d4cc7675a71741bddf2e8266117b689c8

I thought there were some reasons to do this...

src/librbd/Journal.h
trociny: The async version of `get_tag_owner` is used by async `is_tag_owner` and this later is needed for async mirror enable/disable. Though as `get_tag_owner` is needed only for `is_tag_owner` I think I will hide it.

src/librbd/Journal.h
trociny: I can't extract it inside `Journal<I>`, because in test_mock I->op_work_queue is MockContextWQ and build fails:
```
error: no matching function for call to ‘journal::Journaler::Journaler(librbd::MockContextWQ*&, SafeTimer*&, Mutex*&, librados::IoCtx&, std::string&, const string&, <brace-enclosed initializer list>)’
```

src/librbd/Journal.cc
trociny: @dillaman I can't find journal::OpenRequest/journal::CloseReques in master (only Create/Remove). May be they exist in some development branch? Also, I need to init only Journaler, though it might not make much difference.

Anyway, I think in future it could be optimized not to open/close journaler at all, just to get tags. I suppose a static method could be added to JournalMetadata...

trociny:
@dillaman I have updated it according to your comments. Open issues:

I still reopen image in MirrorEnableRequest, as it was in the original (sync) code. Do you have an idea why it was needed? Might it be due to we were holding snap_lock in the sync code at that time?

Also, I am worried by my reorgaization of DisableFeatures state machine. As you suggested, I made it do all checks it needed before making any changes. In the previous version it disabled mirroring at the beginning (as it is done in our current sync version).

As I understand, in the original version there were at least two reasons why we did disable mirroring at the initial stage:

1) Before acquiring locks, to avoid recursive lock when removing snap shots.
2) Before setting Disable journal policy (otherwise an assertion could trigger when appending remove_snap operation to the journal).

In async version we don't hold constantly locks, so (1) is not needed. As for (2), disabled journal policy was introduced to avoid attempting to open the journal when acquiring the exclusive lock in case the journal is corrupt. So in my new code I set disabled policy before acquiring the lock and return back immediately after acquiring the lock, so `disable mirroring` is run with standard policy.

Not sure this is correct though, and actually I think we don't need to append to journal 'snap remove' operations from 'disable mirroring' in our case?

src/librbd/Journal.cc
dillaman: @trociny Sorry -- yeah, that was in Venky's PR branch. I definitely agree that it would be nicer to just directly extract the values instead of invoking a heavier weight journal open/close. I just have a feeling like the current proposal involves a lot of duplicated logic since it is essentially the first couple states in the "librbd::Journal" state machine. What if you extracted those out to a new "librbd::journal::OpenJournalerRequest" state machine that could be re-used by your logic and "librbd::Journal::create_journaler"?

dillaman:
>
> @dillaman <https://github.com/dillaman> I have updated it according to
> your comments. Open issues:
>
> I still reopen image in MirrorEnableRequest, as it was in the original
> (sync) code. Do you have an idea why it was needed? Might it be due to we
> were holding snap_lock in the sync code at that time?
>
Yes, you are correct. This was a workaround to avoid recursive locking that
most likely shouldn't be needed anymore (if there is a recursive locking
issue, you can just execute the offending state via an async callback to
eliminate the lock ordering issue).

> Also, I am worried by my reorgaization of DisableFeatures state machine.
> As you suggested, I made it do all checks it needed before making any
> changes. In the previous version it disabled mirroring at the beginning (as
> it is done in our current sync version).
>
> As I understand, in the original version there were at least two reasons
> why we did disable mirroring at the initial stage:
>
> 1) Before acquiring locks, to avoid recursive lock when removing snap
> shots.
> 2) Before setting Disable journal policy (otherwise an assertion could
> trigger when appending remove_snap operation to the journal).
>
Yes, the current version has lots of oddities to work around the
synchronous nature of the implementation.

> In async version we don't hold constantly locks, so (1) is not needed. As
> for (2), disabled journal policy was introduced to avoid attempting to open
> the journal when acquiring the exclusive lock in case the journal is
> corrupt. So in my new code I set disabled policy before acquiring the lock
> and return back immediately after acquiring the lock, so disable mirroring
> is run with standard policy.
>
> Not sure this is correct though, and actually I think we don't need to
> append to journal 'snap remove' operations from 'disable mirroring' in our
> case?
>
Definitely don't need to journal "snap remove" events to a journal that is
being removed. I'm not sure that switching back to the standard journal
policy would have any effect, though, unless a refresh was kicked off in
the middle somehow. Therefore, you are probably safer to keep it in the
disabled policy until after you have actually removed the journaling
feature bit.

trociny:
@dillaman Updated. I think it is now in much better shape.

I implemented Journal::is_tag_owner/get_tag_owner methods without opening/closing journaler (they use only journaler get_client and get_tags methods, which actually do not require the journaler init).

Also, I had to change DisabledPolicy::append_disabled method to allow its call -- update_features runs disabling mirroring with this policy, so snap_remove operations are not mirrored.

src/librbd/internal.cc
```diff
@@ -1604,302 +1360,7 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force,
...
-      C_SaferCond lock_ctx;
-      ictx->exclusive_lock->release_lock(&lock_ctx);
-      r = lock_ctx.wait();
-      if (r < 0) {
-        lderr(cct) << "failed to unlock image: " << cpp_strerror(r) << dendl;
-        return r;
-      }
-    }
-    return 0;
+    return ictx->operations->update_features(features, enabled);
```
dillaman: Minor: might as well yank this method and have ```librbd.cc``` just directly invoke the ```Operations``` call

src/librbd/mirror/DisableRequest.cc
```diff
@@ -0,0 +1,424 @@
...
+  if (*result < 0) {
+    lderr(cct) << "failed to remove mirror image: " << cpp_strerror(*result)
+	       << dendl;
+  }
+
+  ldout(cct, 20) << this << " " << __func__
+		 <<  ": removed image state from rbd_mirroring object" << dendl;
+
+  if (m_is_primary) {
+    // TODO: send notification to mirroring object about update
```
dillaman: TODO

src/librbd/mirror/DisableRequest.h
```diff
@@ -0,0 +1,127 @@
...
+   *    |     |                             |             |       *
+   *    |     v                             |             |       *
+   *    |  UNREGISTER_CLIENT ---------------/-------------/ * * * *
+   *    |                                                         *
+   *    | (no more clients                                        *
+   *    |  to unregister)                                         *
+   *    v                                                         *
+   * REMOVE_MIRROR_IMAGE  * * * * * * * * * * * * * * * * * * * * *
+   *    |         (skip if no remove)                             *
+   *    v                                                         *
```
dillaman: NOTIFY_MIRRORING_WATCHER_REMOVED

src/librbd/operation/DisableFeaturesRequest.h
```diff
@@ -0,0 +1,162 @@
...
+   * STATE_REMOVE_OBJECT_MAP (skip if not
+   *    |                     disabling object map)
+   *    v
+   * STATE_SET_FEATURES
+   *    |
+   *    v
+   * STATE_UPDATE_FLAGS
+   *    |
+   *    v
+   * STATE_APPEND_OP_EVENT (skip if journaling
```
dillaman: I think this needs to be before REMOVE_OBJECT_MAP since that would be the first modification state

src/librbd/operation/EnableFeaturesRequest.h
```diff
@@ -0,0 +1,128 @@
...
+   * STATE_SET_FEATURES
+   *    |
+   *    v
+   * STATE_CREATE_OBJECT_MAP (skip if not
+   *    |                     required)
+   *    v
+   * STATE_ENABLE_MIRROR_IMAGE
+   *    |
+   *    V
+   * STATE_APPEND_OP_EVENT (skip if journaling
```
dillaman: Move to before UPDATE_FLAGS state

src/librbd/operation/DisableFeaturesRequest.cc
```diff
@@ -0,0 +1,599 @@
...
+  CephContext *cct = image_ctx.cct;
+
+  ldout(cct, 20) << this << " " << __func__ << dendl;
+
+  Context *ctx = create_context_callback<
+    DisableFeaturesRequest<I>,
+    &DisableFeaturesRequest<I>::handle_disable_mirror_image>(this);
+
+  mirror::DisableRequest<I> *req =
+    mirror::DisableRequest<I>::create(&image_ctx, false, true, ctx);
```
vshankar: I made changes to #10896 to make use `DisableFeaturesRequest` provided by this PR. `rbd-mirror` fails to remove the non-primary image if `force` option isn't provided, which is always passed as `false` here.

Would it make sense to accept "force" as an argument to this state machine and pass it to `mirror::DisableRequest`?

[`mirror_image_disable_internal` passes `force` to `mirror::DisableRequest`, but my changes invokes it via `DisableFeaturesRequest`]

trociny: I am ok with this change. I think you can do it yourself as an additional commit?

@dillaman What do you think?

dillaman: @vshankar I'd say go ahead and change as needed as an additional commit in your PR

vshankar: Sure. Thanks @dillaman @trociny

trociny:
@dillaman I think I have addressed you last comments.

As for "TODO: send notification to mirroring object about update", I fixed this  according to your recommendation adding NOTIFY_MIRRORING_WATCHER_REMOVED state, though I had to add a new cls::rbd::MirrorImageState enum value: MIRROR_IMAGE_STATE_REMOVED, and I guess now it is still not very useful?

Also, I have added mock tests for enable/disable features operations.

src/cls/rbd/cls_rbd_types.h
```diff
@@ -80,7 +80,8 @@ WRITE_CLASS_ENCODER(MirrorPeer);
 
 enum MirrorImageState {
   MIRROR_IMAGE_STATE_DISABLING = 0,
-  MIRROR_IMAGE_STATE_ENABLED   = 1
+  MIRROR_IMAGE_STATE_ENABLED   = 1,
+  MIRROR_IMAGE_STATE_REMOVED   = 2,
```
dillaman: Maybe "DISABLED" to keep it consistent with "DISABLING" state naming?

trociny: updated

trociny:
Updated. 
* New commits added:
 - use counter to track exclusive lock block/unblock requests
 - interlock image state machine and update features operations￼
* TestJournalReplay.UpdateFeatures test is added
* DisableFeatures is changed to handle the compatibility issue fixed in #11155

dillaman:
@trociny Seeing some new test failures:

http://qa-proxy.ceph.com/teuthology/jdillaman-2016-09-23_20:59:27-rbd-wip-jd-testing---basic-smithi/434867/teuthology.log
> 2016-09-24T13:35:45.878 INFO:tasks.qemu.client.0.smithi113.stderr:/srv/autobuild-ceph/gitbuilder.git/build/out~/ceph-11.0.0-2676-g5cdbdc0/src/librbd/AioImageRequestWQ.cc: 243: FAILED assert(m_image_ctx.owner_lock.is_locked())

trociny:
@dillaman I think I have fixed all issues found by testing. Please take a look.

src/librbd/Journal.cc
```diff
@@ -156,6 +156,162 @@ class ThreadPoolSingleton : public ThreadPool {
...
+              cls::journal::Client *client,
+              journal::ImageClientMeta *client_meta,
+              uint64_t *tag_tid, journal::TagData *tag_data,
+              Context *on_finish) {
+  ldout(cct, 20) << __func__ << dendl;
+
+  GetTagsRequest<J> *req =
+    new GetTagsRequest<J>(cct, journaler, client, client_meta, tag_tid,
+                          tag_data, on_finish);
+  req->send();
```
dillaman: Memory leak: http://qa-proxy.ceph.com/teuthology/jdillaman-2016-09-27_15:44:42-rbd-wip-jd-testing-distro-basic-smithi/440609/teuthology.log

Perhaps rename ```GetTagsRequest``` to ```C_GetTagsRequest```, inherit from Context, and invoke "complete(xyz)" where it is currently calling "m_on_finish->complete(xyz)".

trociny: @dillaman Thanks! The update pushed. I decided not to inherit from Context, as it does not look very useful.

PR #10982, opened 2016-09-05
cmake: fix librados compiling error

renzhengeek:
The build error happened with the following message:
    ```
    /var/ceph/ceph/src/librados/librados.cc:47:30: fatal error:
    tracing/librados.h: No such file or directory
     #include "tracing/librados.h"
                                  ^
    compilation terminated.
    ```
Actually, this issue was fixed by commit: 5e03ff098166 (
"cmake: fix the tracing header dependencies"). However,
it comes up again becuase of this commit: 0a717593c0c3 (
"cmake: let librados_api_obj depend on librados-tp"), which
forgets to replace "rados-tp" with "librados-tp". So, complete
this change accordingly.

Signed-off-by: Eric Ren <zren@suse.com>

### DIFF
--- a/src/tracing/CMakeLists.txt
+++ b/src/tracing/CMakeLists.txt
@@ -43,9 +43,9 @@ endfunction()
 
 set(osd_traces oprequest.tp osd.tp pg.tp)
 add_tracing_library(osd_tp "${osd_traces}" 1.0.0)
-add_tracing_library(rados_tp librados.tp 2.0.0)
+add_tracing_library(librados_tp librados.tp 2.0.0)
 add_tracing_library(rbd_tp librbd.tp 1.0.0)
 add_tracing_library(os_tp objectstore.tp 1.0.0)
 
-install(TARGETS rados_tp osd_tp rbd_tp os_tp DESTINATION ${CMAKE_INSTALL_LIBDIR})
+install(TARGETS librados_tp osd_tp rbd_tp os_tp DESTINATION ${CMAKE_INSTALL_LIBDIR})
 


renzhengeek:
Fixes: http://tracker.ceph.com/issues/17209

renzhengeek:
@tchaikov  Hello, could you look at this?

src/tracing/CMakeLists.txt
```diff
@@ -43,9 +43,9 @@ endfunction()
 
 set(osd_traces oprequest.tp osd.tp pg.tp)
 add_tracing_library(osd_tp "${osd_traces}" 1.0.0)
-add_tracing_library(rados_tp librados.tp 2.0.0)
+add_tracing_library(librados_tp librados.tp 2.0.0)
```
tchaikov: i don't understand why it makes different if we use librados_tp as the tracing point library name for librados. and please note, the generated library would be named as "liblibrados_tp.so" if we go this way.

also, "rados-tp" is not "rados_tp". the former is the tracing library, the later is just a target for building the header file for librados's tracing points.

renzhengeek: Yes, I see, thanks! Now, I have no idea how to handle this... how about revoking commit: 0a71759?

tchaikov: @renzhengeek if only we are able to root cause
```
    /var/ceph/ceph/src/librados/librados.cc:47:30: fatal error:
    tracing/librados.h: No such file or directory
     #include "tracing/librados.h"
                                  ^
    compilation terminated.
```
and understand why reverting 0a71759 helps, i am happy to do so.

renzhengeek: @tchaikov Hello, here are some question I'm not sure:
- This fix works because this line? - https://github.com/ceph/ceph/blob/master/src/tracing/CMakeLists.txt#L39:   
```
  set_target_properties(${name} PROPERTIES
  OUTPUT_NAME ${name}
```     
  See [set_target_properties](https://cmake.org/cmake/help/v3.0/command/set_target_properties.html)
- it's weird that I can not find ibrados_tp library (the same for other *_tp libraries). According to   
```
install(TARGETS rados_tp osd_tp rbd_tp os_tp DESTINATION ${CMAKE_INSTALL_LIBDIR})
```
it should be here:
```
ls -1 build/lib/*rados*
build/lib/librados.so
build/lib/librados.so.2
build/lib/librados.so.2.0.0
```

renzhengeek: This building output around librados:   
```
Scanning dependencies of target librados_objs
[ 10%] Building CXX object src/librados/CMakeFiles/librados_objs.dir/IoCtxImpl.cc.o
[ 10%] Building CXX object src/librados/CMakeFiles/librados_objs.dir/RadosXattrIter.cc.o
[ 10%] Building CXX object src/librados/CMakeFiles/librados_objs.dir/RadosClient.cc.o
[ 10%] Built target librados_objs
Scanning dependencies of target librados-tp
[ 10%] generating /home/eric/workspace/ceph/build/include/tracing/librados.h /home/eric/workspace/ceph/build/include/tracing/librados.c
[ 10%] Built target librados-tp
Scanning dependencies of target librados_api_obj
[ 10%] Building CXX object src/librados/CMakeFiles/librados_api_obj.dir/librados.cc.o
[ 10%] Built target librados_api_obj
Scanning dependencies of target librados
[ 10%] Linking CXX shared library ../../lib/librados.so
[ 10%] Built target librados
```

renzhengeek: So at that moment, only target "librados-tp" is built to meet the dependency. But,
unfortunately, there's a another build error following, so "librados_tp" target has not been started to build, right?  
```
Scanning dependencies of target build_rocksdb
[ 17%] rocksdb building
  GEN      util/build_version.cc
  GEN      util/build_version.cc
make[3]: *** No rule to make target 'db/writebuffer.h', needed by 'db/column_family.o'.  Stop.
src/CMakeFiles/build_rocksdb.dir/build.make:57: recipe for target 'src/CMakeFiles/build_rocksdb' failed
make[2]: *** [src/CMakeFiles/build_rocksdb] Error 2
CMakeFiles/Makefile2:424: recipe for target 'src/CMakeFiles/build_rocksdb.dir/all' failed
make[1]: *** [src/CMakeFiles/build_rocksdb.dir/all] Error 2
```

tchaikov: > This fix works because this line?

sorry, i don't follow you. which fix? and why is the output **name** of a library related to the build failure? 

> it's weird that I can not find ibrados_tp library (the same for other *_tp libraries). According to

`make rados-tp`, and try again. or with your change `make librados_tp`.

> there's a another build error following, so "librados_tp" target has not been started to build, right?


with your change, `librados_tp` does not depends on build_rocksdb.

renzhengeek:
@tchaikov It turns out I did some stupid things,  your commit is good, it's my fault. Sorry, not familiar with cmake;-)

tchaikov:
@renzhengeek i am closing this PR. please reopen it if you think otherwise.

renzhengeek:
@tchaikov this error re-occurs when I do it over again:
```
1. rm -rf build
2. ./do_cmake.sh
3. cd build
4. make librados
```
I think this issue is caused by :
```
grep -rn LTTNG
include/acconfig.h:136:/* #undef WITH_LTTNG */
CMakeCache.txt:536:WITH_LTTNG:BOOL=OFF
```
When "WITH_LTTNG" is OFF, "#undef WITH_LTTNG" should be commented out, right?
I have a try with the following change, and the compilation passed:
```
    44 #if WITH_LTTNG == ON^[[>1;4003;0c
    45 #define TRACEPOINT_DEFINE
    46 #define TRACEPOINT_PROBE_DYNAMIC_LINKAGE
    47 #include "tracing/librados.h"
```

renzhengeek:
@tchaikov Do you like the change like this?
```
#if WITH_LTTNG == ON
```
- If so,  many other places should be changed accordingly, right?
```
find -name "*.cc" | xargs grep -rn "WITH_LTTNG"
./src/librados/librados.cc:44:#if WITH_LTTNG == ON
./src/librbd/AioCompletion.cc:18:#ifdef WITH_LTTNG
./src/librbd/librbd.cc:35:#ifdef WITH_LTTNG
./src/os/filestore/FileStore.cc:82:#ifdef WITH_LTTNG
./src/os/filestore/FileStore.cc:2501:#ifdef WITH_LTTNG
./src/osd/OpRequest.cc:16:#ifdef WITH_LTTNG
./src/osd/OpRequest.cc:117:#ifdef WITH_LTTNG
./src/osd/OpRequest.cc:137:#ifdef WITH_LTTNG
./src/osd/OSD.cc:142:#ifdef WITH_LTTNG
./src/osd/OSD.cc:5933:#ifdef WITH_LTTNG
./src/osd/OSD.cc:8731:#ifdef WITH_LTTNG
./src/osd/OSD.cc:8753:#ifdef WITH_LTTNG
./src/osd/PG.cc:55:#ifdef WITH_LTTNG
./src/osd/PG.cc:1919:#ifdef WITH_LTTNG
./src/osd/ReplicatedPG.cc:51:#ifdef WITH_LTTNG
./src/osd/ReplicatedPG.cc:2951:#ifdef WITH_LTTNG
./src/osd/ReplicatedPG.cc:2961:#ifdef WITH_LTTNG
```
- if not, do you have suggestion how to play with this file - src/acconfig.h or build/include/acconfig.h?

tchaikov:
with `WITH_LTTNG=OFF`, neither `#define WITH_LTTNG` nor `#undef WITH_LTTNG` is added to `acconfig.h`.


> Do you like the change like this?

where is `ON` defined? if we go this way?

No, unless i understand why this fixes the issue. if a macro is not defined, its value is 0. so `WITH_LTTNG` is 0, and so does `ON`. so if we don't enable `WITH_LTTNG`, the code block guarded by `#if WITH_LTTNG == ON` is always enabled. 

> if not, do you have suggestion how to play with this file - src/acconfig.h or build/include/acconfig.h?

sorry, no suggestions so far.

BTW, i would try to understand the underneath machinery before taking any wild guess if i wanted to fix the issue.

renzhengeek:
@tchaikov  thanks for keeping keys on this;-)

> with WITH_LTTNG=OFF, neither #define WITH_LTTNG nor #undef WITH_LTTNG is added to acconfig.h. this is expected.
Honestly, i don't understand this because `#define` and `#undef` is contradictory?

> where is ON defined? if we go this way?
`ON` or `OFF` is the value of `WITH_LTTNG`?  according to [CMake Options](https://github.com/ceph/ceph/blob/master/README.md#cmake-options) like:
```
cmake -DWITH_RADOSGW=OFF [path to top level ceph directory]
```
Maybe `WITH_LTTNG` is defined in the similar way?

> if a macro is not defined, its value is 0. so WITH_LTTNG is 0, and so does ON. 
That's the problem.  I didn't define `WITH_LTTNG`, but as the previous comment it has a value `OFF`, which is different with autotool stuff:
```
cd build
grep -rn LTTNG
CMakeCache.txt:536:WITH_LTTNG:BOOL=OFF
```
`#ifdef` doesn't care its value as long as `WITH_LTTNG` is defined, which is different from `#if`, right?  That's why I use `#if` to check its value;-)

> so if we don't enable WITH_LTTNG, the code block guarded by #if WITH_LTTNG == ON is always enabled.
Hmm, if  `WITH_LTTNG` is `0` if not defined as you said,  then `#if 0 == ON` is false, how can the code block be enabled?

Now, I don't even trust myself;-) Could you please check if this issue is valid? if so, please add appropriate labels, looks I don't have the right. Maybe, someone who knows this better could help;-)

renzhengeek:
Oh my god, I mixed cmake stuff with C/C++ syntax. Yes, cmake will translate `ON` to `#define WITH_LTTNG` in C header file while `OFF` doing nothing. Uhm, so this issue is weird;-/

renzhengeek:
Oh my god, I mixed cmake stuff with C/C++ syntax. Yes, cmake will translate `ON` to `#define WITH_LTTNG` in C header file while `OFF` doing nothing. Uhm, so this issue is weird;-/

renzhengeek:
I think I got the root cause this time. It's because I used automake before.
`src/acconfig.h` was left behind, which is included by `common` code. And `librados.cc` contains `#include "common/config.h"`, so `WITH_LTTNG` is introduced this way.  I guess so because when I put  `#include "common/config.h"` after the `WITH_LTTNG` block, it compiles well.

Now, this issue is gone after removing `src/acconfig.h`. Sorry for the noise. Thanks!

PR #10983, opened 2016-09-05
don't start ceph-osd at udev rules

zealoussnow:
When I use cgroup to limit the ceph-osd and ceph-mon resource, I missed a problem. I write some cgroup  rules, and it works fine. But when I reboot the system, I found it doesn't work, because the cgroup will start after udev rules.
Finally, I know when the system reboot, ceph osd daemons will start by udev rules, not the ceph-osd@.serivce. So I want udev rules only mount disk to osd path, and ceph-osd@.service start the osd daemons. It works fine.
Signed-off-by: zealoussnow nguzcf@gmail.com

### DIFF
--- a/src/ceph-disk/ceph_disk/main.py
+++ b/src/ceph-disk/ceph_disk/main.py
@@ -2869,13 +2869,6 @@ def start_daemon(
                     'ceph-osd@{osd_id}'.format(osd_id=osd_id),
                 ],
             )
-            command_check_call(
-                [
-                    'systemctl',
-                    'start',
-                    'ceph-osd@{osd_id}'.format(osd_id=osd_id),
-                ],
-            )
         elif os.path.exists(os.path.join(path, 'openrc')):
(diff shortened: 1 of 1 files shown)


zealoussnow:
Signed-off-by: zealoussnow nguzcf@gmail.com

tchaikov:
@dachary i think this would cause regressions, right?

dachary:
@zealoussnow I'm afraid @tchaikov is right: that patch will break things for people who are not in the same situation as you.

zealoussnow:
I think in the udev rules, the command "ceph-disk activate" only need to mount the device to oad path, the osd service will start at systemd script.
Now I delete the code, I can restart the osd service manually or add restart operation in ceph-deploy.
In most cases, we don't call "ceph-disk" directly.

zealoussnow:
@dachary I thought about it too, it just a suggestion, and I will think a good solution.

PR #10985, opened 2016-09-05
rbd-nbd: fix kernel deadlock during teuthology testing

dillaman:
Fixes: http://tracker.ceph.com/issues/16921

### DIFF
--- a/src/librbd/operation/ResizeRequest.h
+++ b/src/librbd/operation/ResizeRequest.h
@@ -80,10 +80,10 @@ private:
    *    |           STATE_UPDATE_HEADER ----------------------------\
    *    |                                                           |
    *    | (shrink)                                                  |
-   *    |\--------> STATE_TRIM_IMAGE                                |
+   *    |\--------> STATE_INVALIDATE_CACHE                          |
    *    |                 |                                         |
    *    |                 v                                         |
-   *    |           STATE_INVALIDATE_CACHE                          |
+   *    |           STATE_TRIM_IMAGE                                |
    *    |                 |                                         |
    *    |                 v                                         |
    *    |           STATE_POST_BLOCK_WRITES                         |
--- a/src/librbd/operation/ResizeRequest.h
+++ b/src/librbd/operation/ResizeRequest.h
@@ -80,7 +80,10 @@ private:
    *    |           STATE_UPDATE_HEADER ----------------------------\
    *    |                                                           |
    *    | (shrink)                                                  |
-   *    |\--------> STATE_INVALIDATE_CACHE                          |
+   *    |\--------> STATE_FLUSH_CACHE                               |
+   *    |                 |                                         |
+   *    |                 v                                         |
+   *    |           STATE_INVALIDATE_CACHE                          |
    *    |                 |                                         |
    *    |                 v                                         |
    *    |           STATE_TRIM_IMAGE                                |
@@ -120,6 +123,9 @@ private:
 
   Context *send_append_op_event();
   Context *handle_append_op_event(int *result);
+
+  void send_flush_cache();
+  Context *handle_flush_cache(int *result);
 
   void send_invalidate_cache();
   Context *handle_invalidate_cache(int *result);
--- a/src/tools/rbd_nbd/rbd-nbd.cc
+++ b/src/tools/rbd_nbd/rbd-nbd.cc
@@ -236,6 +236,7 @@ private:
       switch (ctx->command)
       {
         case NBD_CMD_DISC:
+          // RBD_DO_IT will return when pipe is closed
 	  dout(0) << "disconnect request received" << dendl;
           return;
         case NBD_CMD_WRITE:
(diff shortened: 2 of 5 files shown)


dillaman:
http://pulpito.ceph.com/jdillaman-2016-09-05_14:09:59-rbd-wip-16921---basic-smithi/

src/tools/rbd_nbd/rbd-nbd.cc
```diff
@@ -614,6 +614,8 @@ static int do_map()
 
       server.start();
       ioctl(nbd, NBD_DO_IT);
+      ioctl(nbd, NBD_CLEAR_QUE);
```
trociny: @dillaman NBD_CLEAR_QUE looks like deprecated and nop.

dillaman: Yeah, I just put these in here to match qemu-nbd's implementation.

dillaman:
@trociny Definitely a different issue:

[ 7241.278129] block nbd0: Other side returned error (22)
[ 7241.283380] blk_update_request: I/O error, dev nbd0, sector 456576

dillaman:
The resize is hanging because rbd-nbd (the exclusive lock owner) isn't responding -- and it isn't responding most likely because its librbd thread is deadlocked.  It looks like you have gdb attached to it so I cannot see what its doing.

trociny:
Sorry, yes I tried attaching with gdb but it hanged and letft the terminal opened. I have kiled my session.

dillaman:
@trociny Ok -- same thing is happening to me. I am thoroughly convinced that the nbd block driver has lots of edge conditions. Looking at the deadlock listed on mira023 (which is running a 4.8-rc5 kernel instead of stock), it is showing a deadlock in the nbd block driver shut down logic and looking at the code it is definitely a problem and I don't immediately see a way around it.

dillaman:
@trociny It appears to be stable now -- I updated the tests to not colocate nbd with the OSDs and I fixed the resize failure.

https://github.com/ceph/ceph-qa-suite/pull/1170

dillaman:
http://pulpito.ceph.com/jdillaman-2016-09-08_20:44:48-rbd-wip-16921---basic-smithi/
http://pulpito.ceph.com/jdillaman-2016-09-08_20:44:36-rbd-wip-16921---basic-vps/
http://pulpito.ceph.com/jdillaman-2016-09-08_15:43:38-rbd-wip-16921---basic-vps/

dillaman:
Kicked off rbd suite run: http://pulpito.ceph.com/jdillaman-2016-09-08_21:48:20-rbd-wip-16921---basic-smithi/

src/librbd/operation/ResizeRequest.h
```diff
@@ -121,6 +121,9 @@ class ResizeRequest : public Request<ImageCtxT> {
   Context *send_append_op_event();
   Context *handle_append_op_event(int *result);
 
+  void send_flush_cache();
+  Context *handle_flush_cache(int *result);
```
trociny: @dillaman Update the diagram?

src/tools/rbd_nbd/rbd-nbd.cc
```diff
@@ -185,13 +185,21 @@ class NBDServer
 
     dout(20) << __func__ << ": " << *ctx << dendl;
 
+    if (ret == -EINVAL) {
+      // if shrinking an image, a pagecache writeback might reference
+      // extents outside of the range of the new image extents
+      dout(20) << __func__ << ": masking IO out-of-bounds error" << dendl;
```
trociny: I think this message could be a lower debug level than 20?

dillaman:
@trociny Pushed updates -- the unrelated blacklisting test failure is fixed by PR #11034 and the dynamic_features failure is covered by open ticket http://tracker.ceph.com/issues/17227

trociny:
lgtm

src/tools/rbd_nbd/rbd-nbd.cc
```diff
@@ -727,6 +729,7 @@ static int rbd_nbd(int argc, const char *argv[])
   env_to_vec(args);
   global_init(NULL, args, CEPH_ENTITY_TYPE_CLIENT, CODE_ENVIRONMENT_DAEMON,
               CINIT_FLAG_UNPRIVILEGED_DAEMON_DEFAULTS);
+  g_ceph_context->_conf->set_val_or_die("pid_file", "");
```
huww98, line 740: Hi Jason (and all reading this), I'm working on starting rbd-nbd with systemd. And it is best to let rbd-nbd write a pid file so that systemd can reliably determine the main process. But this config is disabled here. See [this comment](https://github.com/ceph/ceph/pull/40844#issuecomment-820168209) for more. Do you have any suggestions? thanks.

isodude: Why not go directly and add notify/watchdog support? That would make more sense I think.

PR #10988, opened 2016-09-06
autogen: Fix rocksdb error when make dist

jazeltq:
Rocksdb delete some files, so we should change another way
to get rocksdb's files

Signed-off-by: tianqing <tianqing@unitedstack.com>


src/Makefile-rocksdb.am
```diff
@@ -1,846 +1 @@
...
-  rocksdb/util/transaction_test_util.cc \
-  rocksdb/util/transaction_test_util.h \
-  rocksdb/util/xfunc.cc \
-  rocksdb/util/xfunc.h \
-  rocksdb/util/xxhash.cc \
-  rocksdb/util/xxhash.h \
-  rocksdb/Vagrantfile \
-  rocksdb/WINDOWS_PORT.md \
-  rocksdb/AUTHORS
+EXTRA_DIST += $(find rocksdb/ -type f)
```
tchaikov: probably you can reference the implementation in https://github.com/ceph/ceph/blob/master/src/script/make_rocksdb_makefile.sh ? or revise it so we can leverage it here?

tchaikov:
could you prefix the title of your commit message with the subcomponent your are changing ? see https://github.com/ceph/ceph/blob/master/SubmittingPatches.rst#preparing-and-sending-patches, in this case, it's "Makefile: ".

PR #10991, opened 2016-09-06
rgw: fix failed to create bucket if a non-master zonegroup has a single zone

weiqiaomiao:
If a non-master zonegroup has a single zone, the metadata sync thread not running and
the non-master zonegroup can't sync user from master zonegroup,
so we can't create bucket(or other metadata update) in it
because the authenticated user not found in the zone of non-master zonegroup.


smithfarm:
tracker http://tracker.ceph.com/issues/19756 created to enable jewel backport

PR #10992, opened 2016-09-06
rgw: fix a bug in rgw cache in delete_system_obj and get_system_obj.

zhangsw:
Cache's key is generated by function normal_name(bucket,oid), and these
two parameters are generated by function normalize_bucket_and_obj. But in
delete_system_obj and get_system_obj it doesn't work like that. In other
place ,like put_system_obj_impl, keys are always generated in that way.

### DIFF
--- a/src/rgw/rgw_cache.h
+++ b/src/rgw/rgw_cache.h
@@ -188,9 +188,6 @@ class RGWCache  : public T
   }
 
   void normalize_pool_and_obj(rgw_pool& src_pool, const string& src_obj, rgw_pool& dst_pool, string& dst_obj);
-  string normal_name(rgw_raw_obj& obj) {
-    return normal_name(obj.pool, obj.oid);
-  }
 
   int init_rados() override {
     int ret;
@@ -268,7 +265,7 @@ int RGWCache<T>::delete_system_obj(rgw_raw_obj& obj, RGWObjVersionTracker *objv_
   string oid;
   normalize_pool_and_obj(obj.pool, obj.oid, pool, oid);
 
-  string name = normal_name(obj);
+  string name = normal_name(pool, oid);
   cache.remove(name);
 
   ObjectCacheInfo info;
@@ -290,7 +287,7 @@ int RGWCache<T>::get_system_obj(RGWObjectCtx& obj_ctx, RGWRados::SystemObject::R
   if (ofs != 0)
     return T::get_system_obj(obj_ctx, read_state, objv_tracker, obj, obl, ofs, end, attrs, cache_info);
 
-  string name = normal_name(obj.pool, oid);
+  string name = normal_name(pool, oid);
 
   ObjectCacheInfo info;
 


mattbenjamin:
@zhangsw As I read the code in rgw_cache.h, "normal_name(obj)" is always a call to member function
    string RGWCache<T>::normal_name(rgw_obj& obj)
whose implementation is 
    return normal_name(obj.bucket, obj.get_object());
which is a call to member function string RGWCache<T>::(rgw_bucket& bucket, const std::string& oid)
in the same class (and defined in the class decl).

So it looks like this change just expands the object-arg form of normal_name to the bucket, string- arg form, which the compiler will also do, as member functions in a class decl are always implicitly declared inline.  Am I misreading this?

zhangsw:
@mattbenjamin When we call normal_name(bucket, oid) in this place, the parameters bucket and oid are generated by the function normalize_bucket_and obj(). They may be not  same as the obj.bucket and  obj.get_object(). 
``` 
void RGWCache<T>::normalize_bucket_and_obj(rgw_bucket& src_bucket, const string& src_obj, rgw_bucket& dst_bucket, string& dst_obj)
{
  if (src_obj.size()) {
    dst_bucket = src_bucket;
    dst_obj = src_obj;
  } else {
    dst_bucket = T::get_zone_params().domain_root;
    dst_obj = src_bucket.name;
  }
}```

**Comment by mattbenjamin** [comment]
@zhangsw that's interesting;
@yehudasa what are your thoughts on this?
@zhangsw assuming this is correct, have you identified or hit a scenario where this takes effect?  what is the effect?  and, it seems like it might be possible to write a unit test which checks something like this, would you be up for adding one?

**Comment by yehudasa** [comment]
@mattbenjamin @zhangsw yeah, I see the issue I think. One thing to note is that we only cache system objects, and I think that the whole bucket param there is superfluous. The term bucket is completely inappropriate, we should use a 'pool' string here as we cache rados pool:oid pairs. We use 'bucket' as originally there was a 1:1 mapping of buckets to pools. A better cleanup will clean that up. Anyhow, as a first step we can take this, but please remove the normal_name that takes only obj.

**Comment by mattbenjamin** [comment]
@yehudasa thanks for the detailed analysis;  I'll just push this into teuthology for smoke test

**Comment by zhangsw** [comment]
@mattbenjamin @yehudasa Thanks for your analysis. I've removed the normal_name that takes only obj in the second commit. In most cases, the src_obj's size is bigger than zero. But when we delete user(or in some other situations), the src_obj's size can be zero.

**Comment by zhangsw** [comment]
@mattbenjamin Could this pr be merged now?

**Comment by oritwas** [comment]
@zhangsw please rebase

**Comment by zhangsw** [comment]
@oritwas done

File: src/rgw/rgw_cache.h [hunk 1]
```diff
@@ -268,7 +265,7 @@ int RGWCache<T>::delete_system_obj(rgw_raw_obj& obj, RGWObjVersionTracker *objv_
   string oid;
   normalize_pool_and_obj(obj.pool, obj.oid, pool, oid);
 
-  string name = normal_name(obj);
+  string name = normal_name(pool, oid);
```
amitkumar50: @zhangsw (NIT) Please consider to keep PR title < 50 characters for future PRs.

yuriw:
[wip-yuri4-testing-2017-11-17-1645](https://trello.com/c/covJSIzf/204-wip-yuri4-testing-2017-11-17-1645)

yuriw:
test this please

PR #10993, opened 2016-09-06
crush: make counting of choose_tries consistent

Vicente-Cheng:
When we alloc choose_tries map (for crushtool checking), we need to alloc the number of tries instead of retries.

In previous version, we could get fault when we testing the crush rule which could not done on a small total tries.

### DIFF
--- a/src/crush/CrushWrapper.h
+++ b/src/crush/CrushWrapper.h
@@ -1026,9 +1026,13 @@ public:
 
   void start_choose_profile() {
     free(crush->choose_tries);
-    crush->choose_tries = (__u32 *)malloc(sizeof(*crush->choose_tries) * crush->choose_total_tries);
+    /*
+     * the original choose_total_tries value was off by one (it
+     * counted "retries" and not "tries").  add one to alloc.
+     */
+    crush->choose_tries = (__u32 *)malloc(sizeof(*crush->choose_tries) * (crush->choose_total_tries + 1));
     memset(crush->choose_tries, 0,
-	   sizeof(*crush->choose_tries) * crush->choose_total_tries);
+	   sizeof(*crush->choose_tries) * (crush->choose_total_tries + 1));
   }
   void stop_choose_profile() {
     free(crush->choose_tries);


src/crush/CrushWrapper.h
```diff
@@ -1026,7 +1026,11 @@ class CrushWrapper {
 
   void start_choose_profile() {
     free(crush->choose_tries);
-    crush->choose_tries = (__u32 *)malloc(sizeof(*crush->choose_tries) * crush->choose_total_tries);
+    /*
+     * the original choose_total_tries value was off by one (it
+     * counted "retries" and not "tries").  add one to alloc.
+     */
+    crush->choose_tries = (__u32 *)malloc(sizeof(*crush->choose_tries) * (crush->choose_total_tries + 1));
```
tchaikov: should update `memset()` accordingly. another way to fix this is to change https://github.com/ceph/ceph/commit/5e0722fab5e548564a965b4af25d38f3eb106b49#diff-0df13ad294f6585c322588cfe026d701L446.

tchaikov:
@Vicente-Cheng could you please file a ticket on tracker against this issue? as i think it also impacts hammer and jewel.

src/crush/CrushWrapper.h
Vicente-Cheng: @tchaikov 
yes, I lost it. actually the original fix for me like your suggestion on on 5e0722f.
Maybe change memset() is enough and didn't change anything about mapper record.
I will update it.

Vicente-Cheng:
@tchaikov 
Ok, I will fill a ticket on tracker.

tchaikov:
thanks @Vicente-Cheng , once you file a ticket, it would be ideal if you can connect to it with a "Fixes: " line in your commit message.

liewegas:
@Vicente-Cheng ping

tchaikov:
lgtm.

Vicente-Cheng:
@tchaikov 
hi kefu, sorry.
I already fill a ticket but I don't remind you here.
I see you change the tracker status, I think I just wait for qa right?

tchaikov:
i didn't notice that =)

and yeah! let's wait for a qa run.

tchaikov:

Vicente-Cheng:
@tchaikov 

Hi kefu, I checked the failure log.
I think that did not cause by this commit, right?

and I need to fill the any backport tickets for this pr?

thanks!!!

tchaikov:
> I think that did not cause by this commit, right?

@Vicente-Cheng no, it's not relevant.

> and I need to fill the any backport tickets for this pr?
if you believe that this fix should be backported, you can just update the http://tracker.ceph.com/issues/17229, to add the target branch for the backport to the "Backport" field of that tracker ticket, then mark the status of it to "Pending backport". and the stable release team will help create backport ticket when they are walking through the "Pending backport" ticket, or you can help them by creating backport ticket. see http://tracker.ceph.com/projects/ceph-releases/wiki/HOWTO_backport_commits for more details.

PR #10995, opened 2016-09-06
rgw multisite: feature of bucket sync enable/disable

Aran85:
use bucket info'flag to store bucket sync flag, before adding the data log, we check the bucket flag.

add two rgw-admin command
    radosgw-admin bucket sync enable
    radosgw-admin bucket sync disable

Signed-off-by: Zengran Zhang zhangzengran@h3c.com

### DIFF
--- a/src/rgw/rgw_admin.cc
+++ b/src/rgw/rgw_admin.cc
@@ -324,6 +324,8 @@ enum {
   OPT_BUCKET_SYNC_STATUS,
   OPT_BUCKET_SYNC_INIT,
   OPT_BUCKET_SYNC_RUN,
+  OPT_BUCKET_SYNC_DISABLE,
+  OPT_BUCKET_SYNC_ENABLE,
   OPT_BUCKET_RM,
   OPT_BUCKET_REWRITE,
   OPT_BUCKET_RESHARD,
@@ -555,6 +557,10 @@ static int get_cmd(const char *cmd, const char *prev_cmd, const char *prev_prev_
       return OPT_BUCKET_SYNC_INIT;
     if (strcmp(cmd, "run") == 0)
       return OPT_BUCKET_SYNC_RUN;
+    if (strcmp(cmd, "disable") == 0)
+      return OPT_BUCKET_SYNC_DISABLE;
+    if (strcmp(cmd, "enable") == 0)
+      return OPT_BUCKET_SYNC_ENABLE;
   } else if (strcmp(prev_cmd, "log") == 0) {
     if (strcmp(cmd, "list") == 0)
       return OPT_LOG_LIST;
@@ -1406,6 +1412,58 @@ int do_check_object_locator(const string& tenant_name, const string& bucket_name
   f->close_section();
 
   f->flush(cout);
+
+  return 0;
+}
+
+int set_bucket_sync_enabled(RGWRados *store, int opt_cmd, const string& tenant_name, const string& bucket_name)
+{
+  RGWBucketInfo bucket_info;
+  map<string, bufferlist> attrs;
+  RGWObjectCtx obj_ctx(store);
+
+  int r = store->get_bucket_info(obj_ctx, tenant_name, bucket_name, bucket_info, NULL, &attrs);
+  if (r < 0) {
+    cerr << "could not get bucket info for bucket=" << bucket_name << ": " << cpp_strerror(-r) << std::endl;
+    return -r;
+  }
+
+  if (opt_cmd == OPT_BUCKET_SYNC_ENABLE) {
+    bucket_info.flags &= ~BUCKET_DATASYNC_DISABLED;
+  } else if (opt_cmd == OPT_BUCKET_SYNC_DISABLE) {
+    bucket_info.flags |= BUCKET_DATASYNC_DISABLED;
+  }
+
+  r = store->put_bucket_instance_info(bucket_info, false, real_time(), &attrs);
+  if (r < 0) {
+    cerr << "ERROR: failed writing bucket instance info: " << cpp_strerror(-r) << std::endl;
+    return -r;
+  }
+
+  int shards_num = bucket_info.num_shards? bucket_info.num_shards : 1;
+  int shard_id = bucket_info.num_shards? 0 : -1;
+
+  if (opt_cmd == OPT_BUCKET_SYNC_DISABLE) {
+    r = store->stop_bi_log_entries(bucket_info.bucket, -1);
+    if (r < 0) {
+      lderr(store->ctx()) << "ERROR: failed writing stop bilog" << dendl;
+      return r;
+    }
+  } else {
+    r = store->resync_bi_log_entries(bucket_info.bucket, -1);
+    if (r < 0) {
+      lderr(store->ctx()) << "ERROR: failed writing resync bilog" << dendl;
+      return r;
+    }
+  }
+
+  for (int i = 0; i < shards_num; ++i, ++shard_id) {
+    r = store->data_log->add_entry(bucket_info.bucket, shard_id);
+    if (r < 0) {
+      lderr(store->ctx()) << "ERROR: failed writing data log" << dendl;
+      return r;
+    }
+  }
 
   return 0;
 }
@@ -6002,6 +6060,35 @@ next:
       return -ret;
     }
   }
+
+  if ((opt_cmd == OPT_BUCKET_SYNC_DISABLE) || (opt_cmd == OPT_BUCKET_SYNC_ENABLE)) {
+    if (bucket_name.empty()) {
+      cerr << "ERROR: bucket not specified" << std::endl;
+      return EINVAL;
+    }
+
+    if (ret < 0) {
+      cerr << "could not init realm " << ": " << cpp_strerror(-ret) << std::endl;
+      return ret;
+    }
+    RGWPeriod period;
+    ret = period.init(g_ceph_context, store, realm_id, realm_name, true);
+    if (ret < 0) {
+      cerr << "failed to init period " << ": " << cpp_strerror(-ret) << std::endl;
+      return ret;
+    }
+
+    if (!store->is_meta_master()) {
+      cerr << "failed to update bucket sync: only allowed on meta master zone "  << std::endl;
+      cerr << period.get_master_zone() << " | " << period.get_realm() << std::endl;
+      return EINVAL;
+    }
+
+    rgw_obj obj(bucket, object);
+    ret = set_bucket_sync_enabled(store, opt_cmd, tenant, bucket_name);
+    if (ret < 0)
+      return -ret;
+}
 
   if (opt_cmd == OPT_BUCKET_SYNC_STATUS) {
     if (source_zone.empty()) {
--- a/src/rgw/rgw_data_sync.cc
+++ b/src/rgw/rgw_data_sync.cc
@@ -1710,11 +1710,13 @@ struct bucket_index_marker_info {
   string bucket_ver;
   string master_ver;
   string max_marker;
+  bool syncstopped{false};
 
   void decode_json(JSONObj *obj) {
     JSONDecoder::decode_json("bucket_ver", bucket_ver, obj);
     JSONDecoder::decode_json("master_ver", master_ver, obj);
     JSONDecoder::decode_json("max_marker", max_marker, obj);
+    JSONDecoder::decode_json("syncstopped", syncstopped, obj);
   }
 };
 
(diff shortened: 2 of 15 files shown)


cbodley:
looks like a good start!

i think this is missing some coordination between zones, though - if you disable sync for a bucket in zone A, how does zone B know to disable it there as well?

sync of bucket information (currently only bucket creation/deletion, as far as i know?) is handled by the metadata sync. bucket creation/deletion requests have to go through the metadata master (the master zone of the master zonegroup), and other zones read the master's log to get updates. so these requests to disable/enable data sync should probably go through the metadata master as well, and generate new events in the metadata log

wido:
Looking at the command, does it disable ALL bucket sync? Or can you specify a way to disable the sync for just one bucket.

cbodley:
@wido the command requires a `--bucket` argument to specify the bucket to enable/disable

src/rgw/rgw_admin.cc
```diff
@@ -1338,6 +1344,43 @@ int do_check_object_locator(const string& tenant_name, const string& bucket_name
...
+    op_target.get_bucket_shard(&bs);
+    r = store->data_log->add_entry(bs->bucket, bs->shard_id);
+    if (r < 0) {
+      lderr(store->ctx()) << "ERROR: failed writing data log" << dendl;
+      return r;
+    }
+
+    bucket_info.flags &= ~BUCKET_DATASYNC_DISABLED;
+  }
+  else if (opt_cmd == OPT_BUCKET_SYNC_DISABLE) {
```
yehudasa: move else next to the curly brace

yehudasa:
@Aran85 @cbodley instead we can just disallow running this command not on the master (return error if not on the master)

yehudasa:
@Aran85 I do see a few problems with the current implementation. The bucket index will continue to log entries, even though sync is disabled for that bucket. Another issue is that when flipping the sync from off to on, we need to trigger sync somehow so that other zones will identify the changes that happened while not following. This is a bit contradictory with the first issue I mentioned (for this to happen we need to continue logging). I think that the way to go here would be by re-initializing sync for the bucket and trigger a full sync for it.

Aran85:
@cbodley 
thanks Casey, recently working outside, so sorry for late reply. rgw-admin bucket disable/enable have checked the condition which must be meta master. because store->put_bucket_instance_info() will generate meta log , so the flag in bucket info will sync to slave when been modified. I did some simple tests, but not  sure if there is a more complex error.

Aran85:
enable -> disable : 
stop data-change log, stop bucket index log, delete bucket sync status
disable -> enable :   
restart data-change log,restart  bucket index log, when read data-change log ,trigger full sync automatic.

@yehudasa do it like this?
there is a problem maybe hard to handle.when one bucket sync been disabled, the user wrote different object to different zone's  bucket, how to sync the divergent object?

yehudasa:
@Aran85 when you say stop data-change log, do you mean what you already do here? if so then yes

If the sync would be re-initialized correctly, then the zones will go through a full-sync and then incremental sync starting at the point where it was re-init. I think that for this to happen when disabling we'll need to remove all the sync markers for that bucket on all zones. Objects that diverged will be reconciled by mtime.

Aran85:
@yehudasa yes,i already stop corresponding data-change log when disable bucket sync, but not stop bucket index log yet. now I still hadn't worked out how to remove all slave sync markers in sync status.

yehudasa:
@Aran85 all zones should remove the markers, we should be careful using the 'slave' terminology here as it's an active-active system.

Aran85:
@yehudasa got it,thanks for reminding

Aran85:
@yehudasa @cbodley have a look the impl now plz, i prefer not to stop data-change log now.when backoff  zone receive data-log , if bucket sync disable , uninit the sync status, break sync. Stopping index log not yet implemented

Aran85:
how about move data-change log entry adding to RGWBucketInstanceMetadataHandler::put() method.Thus every zone will add log entry when enable local bucket sync.

Aran85:
@yehudasa @cbodley updated, delete bucket sync status obj in cli sync disable command on meta master zone. delete bucket sync status obj in meta put when sync meta on none-meta master zone.
had considered all source zones'all shards.

cbodley:
I worry about deleting these sync status objects, when there could be a bucket sync (`RGWRunBucketSyncCoroutine`) running that would recreate them as it makes progress.

The sync coroutines use locks on these sync status objects (see `RGWContinuousLeaseCR`) to make sure that two different gateways don't try to sync that shard at the same time. There are two different locks, "sync_lock.incremental" for `RGWBucketShardIncrementalSyncCR` and "sync_lock" for `RGWBucketShardFullSyncCR`.

So it's only safe for radosgw-admin to delete them if it's able to acquire both of those locks first.

After that, we'll need to make sure that `RGWRunBucketSyncCoroutine` and friends do the right thing when they try to acquire a lock and see `-ENOENT` because it was removed.

Aran85:
@cbodley @yehudasa new impl, need review about the sync disable path, whether it is worth continuing?
when user run bucket sync disable on meta master, master modify the bucket metadata, then add a new type datalog(ENTITY_BUCKET_NOSYNC). the metadata will sync to 'slave' zone. the slave zone will put the metadata locally,then add a new type datalog too! the master's new datalog sync to slave, when it arrive the slave ,we change the bucket sync marker's state to a new type (SyncStop), then go to RGWRunBucketSyncCoroutine, when we found the state is SyncStop, we delete the marker object. the datalog from slave to master will be same. 

use add_entry_raw instead of add_entry, because we want the datalog bypass the "data log window".

if you think it is worth continue, todo list:
1.complete bucket sync restart path.
2.stop bucket index log when sync diable
3.stop data sync in RGWRunBucketSyncCoroutine when we found bucket_info's flag is sync disabled.

cbodley:
@Aran85 i like how you've moved the cleanup of sync status objects into the sync process itself - we should be able to use that to make the locking simpler

the only issue i see with your current approach is that you're writing `state=SyncStop` to the status object without holding its lock. if there's a `RGWBucketShardIncrementalSyncCR` or `RGWBucketShardFullSyncCR` running, they would overwrite that state with their own

maybe we could go a step further, and instead of writing a special `NOSYNC` entry to the datalog, we write it to the bucket index log. that way, we'll see it in `RGWBucketShardIncrementalSyncCR` while we already hold the necessary lock, and we can remove the status object for that shard directly without having to transition to a `SyncStop` state. (we'd still have to write the normal `ENTITY_TYPE_BUCKET` entry to the datalog to make sure it kicks off the `RGWRunBucketSyncCoroutine`)

Aran85:
@cbodley approach updated according your suggest!

src/rgw/rgw_data_sync.cc
```diff
@@ -2659,8 +2743,13 @@ int RGWRunBucketSyncCoroutine::operate()
           << " failed, retcode=" << retcode << dendl;
       return set_cr_error(retcode);
     }
+
     yield {
       if ((rgw_bucket_shard_sync_info::SyncState)sync_status.state == rgw_bucket_shard_sync_info::StateFullSync) {
+	if (!bucket_info.datasync_flag_enabled()) { 
+	  ldout(sync_env->cct, 5) << "bucket sync stoped on" << bucket_str{bs.bucket} << dendl;
+	  return set_cr_done();
+	}
```
cbodley: i think we still need to remove the sync status object in this case. when we re-enable sync on this bucket, we'd want to restart full sync from the beginning instead of the current `rgw_bucket_shard_full_sync_marker::position`

you can have `RGWBucketShardFullSyncCR` check this directly after the `set_status("lock acquired");`

Aran85:
@cbodley modified... now the `RGWBucketShardFullSyncCR `  will force meta sync when listing remote bucket shard  failed, its bc remote zone will reject the request when it found it's bucket sync stoped. so sync the meta here first, then check wether bucket sync is stoped will be safe.

Aran85:
@cbodley ping...

Aran85:
@cbodley add a new tag in bilog header, which let `RGWInitBucketShardSyncCR` could handle sync stop early when the shard has not init yet. i put `RGWUnInitBucketShardCR` out full and inc coroutine,and that let 'last' full or inc  coroutine delete the obj. (because i realized the full and inc coroutine may run 
parallelly,and env on two rgw hosts... thanks your explain).
now i break the full sync coroutine when remote disabled sync, not sure it is a good idea.

cbodley:
hey @Aran85! i was on pto last week, sorry for the delay in feedback.

now that https://github.com/ceph/ceph/pull/11598 merged and `RGWRunBucketSyncCR` controls the lease, it should be easier to handle the cleanup on `CLS_RGW_OP_SYNCSTOP`. `RGWBucketShardIncrementalSyncCR` should be able to flag `sync_status.state = rgw_bucket_shard_sync_info::StateStopping;` without actually writing to the sync status object, and `RGWUninitBucketShardSyncStatusCoroutine` could probably just become an alias for `RGWSimpleRadosDeleteCR` (it doesn't need to acquire locks anymore).

for `RGWBucketShardFullSyncCR`, i'd prefer that we don't try to detect when sync is disabled for now - we should just finish the full sync, then incremental sync will see the `CLS_RGW_OP_SYNCSTOP` entry and trigger the cleanup

later on, once we get full and incremental sync running in parallel, and incremental sync sees `CLS_RGW_OP_SYNCSTOP`, `RGWRunBucketSyncCR` can interrupt the full sync if it's running. how does that sound?

Aran85:
@cbodley sorry for  the delay... thanks so much for the works you have done.

i updated the approach, modified it under your suggest...and added a resync log in bilog, i think it will be safer...if we disable/enable too fast, we could squash them.

we have done some tests on previous approach,  the next few days we will run some tests on the new approach.

Aran85:
@cbodley ping...

src/rgw/rgw_data_sync.cc
```diff
@@ -2484,6 +2506,18 @@ int RGWBucketShardIncrementalSyncCR::operate()
...
+
+	yield {
+	  string oid = RGWBucketSyncStatusManager::status_oid(sync_env->source_zone, bs);
+	  RGWRados *store = sync_env->store;
+	  call(new RGWSimpleRadosDeleteCR(sync_env->async_rados, store, store->get_zone_params().log_pool, oid));
+	}
+
+	return set_cr_done();
+      }
+
```
cbodley: i like how you're reading past the `CLS_RGW_OP_SYNCSTOP` in case there's a `RESYNC` entry shortly after. but the definition of 'shortly after' here is kind of arbitrary - they'd have to be in the same batch of entries returned by `RGWListBucketIndexLogCR`. i think it'd be better to run incremental sync all the way to completion before doing this part. if sync is still disabled on this bucket, then we'll probably only see a couple entries after `SYNCSTOP`, if any. but if `RESYNC` comes anywhere later in the log, we really do want to keep reading and syncing entries

dachary:
@Aran85 FYI there is a conflict to resolve

idealguo:
@cbodley, @Aran85 , as the Kraken version will be released soon, can we continue this PR? thanks

src/rgw/rgw_admin.cc
```diff
@@ -1357,6 +1363,58 @@ int do_check_object_locator(const string& tenant_name, const string& bucket_name
...
+      lderr(store->ctx()) << "ERROR: failed writing stop bilog" << dendl;
+      return r;
+    }
+  } else {
+    r = store->resync_bi_log_entries(bucket_info.bucket, -1);
+    if (r < 0) {
+      lderr(store->ctx()) << "ERROR: failed writing resync bilog" << dendl;
+      return r;
+    }
+  }
```
cbodley: oops, i see that `shard_id=-1` causes them to operate on each shard

src/rgw/rgw_admin.cc
```diff
@@ -6003,6 +6061,35 @@ int main(int argc, char **argv)
...
+    if (ret < 0) {
+      cerr << "failed to init period " << ": " << cpp_strerror(-ret) << std::endl;
+      return ret;
+    }
+
+    if (period.get_master_zone() != store->get_zone_params().get_id()) {
+      cerr << "failed to update bucket sync: only allowed on meta master zone "  << std::endl;
+      cerr << period.get_master_zone() << " | " << period.get_realm() << std::endl;
+      return EINVAL;
+    }
```
cbodley: we shouldn't need to load the period here to call `get_master_zone()`. this check can just use `store->is_meta_master()`

Aran85:
@cbodley @idealguo  sorry for delay.. code updated

cbodley:
thanks @Aran85. i think this branch is in good shape, but i still worry about the potential for races - as a next step, can we work on adding tests to `src/test/rgw/test_multi.py`? for example:
* create a bucket and write some objects
* wait for the bucket to sync on all zones
* disable sync on the bucket
* write some more objects
* reenable sync
* verify that all zones end up with a consistent view of the bucket

to run test_multi.py from the ceph/build directory:
`$ nosetests ../src/test/rgw/test_multi.py`

src/rgw/rgw_admin.cc
```diff
@@ -324,6 +324,8 @@ enum {
   OPT_BUCKET_SYNC_STATUS,
```
platoxie: the bucket enable/disable cmd should be added into function "_usage" in rgw_admin.cc？

zhangsw:
@Aran85 Hi~Whether this work is still in process?

cbodley:
@zhangsw this feature is mostly implemented, it just needs some test coverage. we should be able to make progress on that, even if @Aran85 doesn't have the time to do it

cbodley:
i pushed a rebase of this branch to https://github.com/cbodley/ceph/commits/wip-rgw-per-bucket-sync for testing

zhangsw:
@cbodley What's the status of this work now?

cbodley:
@zhangsw i'd love to continue with review and testing. would you be interested in reopening this pr with a rebased version of https://github.com/cbodley/ceph/commits/wip-rgw-per-bucket-sync? i'd be happy to do it otherwise

cbodley:
merged as https://github.com/ceph/ceph/pull/15801 - thanks to @Aran85 and @zhangsw for all the hard work!

PR #10998, opened 2016-09-06
rgw: work around curl_multi_wait bug with non-blocking reads

cbodley:
Fixes: http://tracker.ceph.com/issues/15915
Fixes: http://tracker.ceph.com/issues/16695

### DIFF
--- a/src/rgw/rgw_http_client.cc
+++ b/src/rgw/rgw_http_client.cc
@@ -418,6 +418,16 @@ int RGWHTTPTransceiver::send_data(void* ptr, size_t len)
 }
 
 
+static int clear_signal(int fd)
+{
+  uint32_t buf;
+  int ret = ::read(fd, (void *)&buf, sizeof(buf));
+  if (ret < 0) {
+    return -errno;
+  }
+  return 0;
+}
+
 #if HAVE_CURL_MULTI_WAIT
 
 static int do_curl_wait(CephContext *cct, CURLM *handle, int signal_fd)
@@ -436,10 +446,8 @@ static int do_curl_wait(CephContext *cct, CURLM *handle, int signal_fd)
   }
 
   if (wait_fd.revents > 0) {
-    uint32_t buf;
-    ret = read(signal_fd, (void *)&buf, sizeof(buf));
+    ret = clear_signal(signal_fd);
     if (ret < 0) {
-      ret = -errno;
       ldout(cct, 0) << "ERROR: " << __func__ << "(): read() returned " << ret << dendl;
       return ret;
     }
@@ -491,10 +499,8 @@ static int do_curl_wait(CephContext *cct, CURLM *handle, int signal_fd)
   }
 
   if (signal_fd > 0 && FD_ISSET(signal_fd, &fdread)) {
-    uint32_t buf;
-    ret = read(signal_fd, (void *)&buf, sizeof(buf));
+    ret = clear_signal(signal_fd);
     if (ret < 0) {
-      ret = -errno;
       ldout(cct, 0) << "ERROR: " << __func__ << "(): read() returned " << ret << dendl;
       return ret;
     }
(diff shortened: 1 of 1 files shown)


cbodley:
this made it through teuthology with no new failures: http://pulpito.ceph.com/cbodley-2016-09-02_15:45:38-rgw-wip-16695---basic-mira

oritwas:
this means we do a read every time do_curl_wait is called, correct?
can we detect what curl version is running and only for older versions call read without checking wait_fd.revents?

cbodley:
@oritwas yeah, that was my first inclination. i wasn't going to require this of the original contributor in #10467, but i agree that it's worth doing

cbodley:
tested and verified using unittest_http_manager and an old version of libcurl 7.29 located at /home/cbodley/local-curl/lib/libcurl.so

when run against master, the bug leads to deadlock:
```
$ LD_PRELOAD=/home/cbodley/local-curl/lib/libcurl.so ./unittest_http_manager
[==========] Running 1 test from 1 test case.
[----------] 1 test from HTTPManager
[ RUN      ] HTTPManager.SignalThread
^C
```
when run against this branch, the bug is detected and the workaround avoids deadlock:
```
$ LD_PRELOAD=/home/cbodley/local-curl/lib/libcurl.so ./unittest_http_manager
[==========] Running 1 test from 1 test case.
[----------] 1 test from HTTPManager
[ RUN      ] HTTPManager.SignalThread
2016-09-09 16:55:25.457454 7eff9c96e900  0 WARNING: detected a version of libcurl which contains a bug in curl_multi_wait(). enabling a workaround that may degrade performance slightly. please upgrade to a more recent version of libcurl.
[       OK ] HTTPManager.SignalThread (71 ms)
[----------] 1 test from HTTPManager (71 ms total)

[==========] 1 test from 1 test case ran. (71 ms total)
[  PASSED  ] 1 test.
```
and with my system's libcurl 7.43, the workaround is not applied and the test passes normally:
```
$ ./unittest_http_manager
[==========] Running 1 test from 1 test case.
[----------] 1 test from HTTPManager
[ RUN      ] HTTPManager.SignalThread
[       OK ] HTTPManager.SignalThread (74 ms)
[----------] 1 test from HTTPManager (74 ms total)

[==========] 1 test from 1 test case ran. (74 ms total)
[  PASSED  ] 1 test.
```

src/rgw/rgw_http_client.cc
```diff
@@ -418,6 +418,16 @@ int RGWHTTPTransceiver::send_data(void* ptr, size_t len)
 }
 
 
+static int clear_signal(int fd)
+{
+  uint32_t buf;
+  int ret = ::read(fd, &buf, sizeof(buf));
```
yehudasa: @cbodley need to cast to void * like we had originally? might be that needed for clearing a warning

src/rgw/rgw_http_client.cc
```diff
@@ -420,8 +420,10 @@ int RGWHTTPTransceiver::send_data(void* ptr, size_t len)
 
 static int clear_signal(int fd)
 {
-  uint32_t buf;
-  int ret = ::read(fd, &buf, sizeof(buf));
+  // since we're in non-blocking mode, we can try to read a lot more than
+  // one signal from signal_thread() to avoid later wakeups
+  std::array<char, 256> buf;
+  int ret = ::read(fd, buf.data(), buf.size());
   if (ret < 0) {
```
yehudasa: @cbodley should we handle EAGAIN?

yehudasa: e.g., clear it?

yehudasa:
@cbodley see my comments, otherwise lgtm

cbodley:
@yehudasa thanks for the feedback, updated `clear_signal()` to preserve the (void *) cast and return 0 on EAGAIN

PR #10999, opened 2016-09-06
os/bluestore: use block_size for bitmap granularity

liewegas:


liewegas:
@chhabaramesh

liewegas:
eh, rolled this into #10963 so i can run test.  it's all passing there now.

PR #11004, opened 2016-09-07
autotools: Fix spec file generation

b-ranto:
We recently started using TARBALL_BASENAME in spec file but never
changed the autotools bits to use templating system to define it. This
patch will query git (if available) for the version string and defines
the variable/macro name.


liewegas:
What is still using autotools and can we kill it?

b-ranto:
I use autotools to generate srpm in my scripts but I can update the scripts and we can drop this (make dist-bzip2 does not seem to work any more anyway).

PR #11006, opened 2016-09-07
python-rados: implement new aio_stat.

iain-buclaw-sociomantic:
This change adds python API for the following C methods:
* rados_aio_stat

Signed-off-by: Iain Buclaw <iain.buclaw@sociomantic.com>

### DIFF
--- a/src/test/pybind/test_rados.py
+++ b/src/test/pybind/test_rados.py
@@ -672,6 +672,32 @@ class TestIoctx(object):
         eq(contents, b"bar")
         [i.remove() for i in self.ioctx.list_objects()]
 
+    def test_aio_stat(self):
+        lock = threading.Condition()
+        count = [0]
+        def cb(_, size, mtime):
+            with lock:
+                count[0] += 1
+                lock.notify()
+
+        comp = self.ioctx.aio_stat("foo", cb)
+        comp.wait_for_complete()
+        with lock:
+            while count[0] < 1:
+                lock.wait()
+        eq(comp.get_return_value(), -2)
+
+        self.ioctx.write("foo", b"bar")
+
+        comp = self.ioctx.aio_stat("foo", cb)
+        comp.wait_for_complete()
+        with lock:
+            while count[0] < 2:
+                lock.wait()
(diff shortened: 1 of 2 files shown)


iain-buclaw-sociomantic:
Would this suffer from the same problem as `aio_read`?

jdurgin:
this looks good, could you add a simple test for it to src/test/pybind/test_rados.py?

What problem with aio_read are you encountering?

iain-buclaw-sociomantic:
> What problem with aio_read are you encountering?

I see that with aio_read, it stores the buffer inside the python completion.  Just wondering if the same would be required for aio_stat too.

https://github.com/ceph/ceph/blob/master/src/pybind/rados/rados.pyx#L2125-L2130

iain-buclaw-sociomantic:
Rebased, previous HEAD was https://github.com/ceph/ceph/pull/11006/commits/41e3e0085c37b6acf463b4567ce7d0a485393c59

src/test/pybind/test_rados.py
```diff
@@ -672,6 +672,32 @@ def cb(blah):
...
+                lock.notify()
+
+        comp = self.ioctx.aio_stat("foo", cb)
+        comp.wait_for_complete()
+        with lock:
+            while count[0] < 1:
+                lock.wait()
+        eq(comp.get_return_value(), -2)
+
+        self.ioctx.write("foo", "bar")
```
jdurgin: the data written should be a bytes object for python 3

jdurgin:
I'm not certain whether it's required, but it's more clearly correct to explicitly include the reference to the time and size in the completion object

iain-buclaw-sociomantic:
> I'm not certain whether it's required, but it's more clearly correct to explicitly include the reference to the time and size in the completion object

Well, I've been running a migration script for over a week straight now that uses aio_stat(), and I'm not seeing any leaking of memory from here.

jdurgin:
Good to know. For my own edification I verified that pmtime and psize are saved in a closure for on_complete() by reading the intermediate C code. Thanks!

PR #11010, opened 2016-09-07
os/bluestore/BlueFS: do not start racing async compaction

liewegas:
Compaction is triggred from sync_metadata.  If one compaction is
in progress and another thread also calls sync_metadata, do not
trigger a second async compaction!

### DIFF
--- a/src/os/bluestore/BlueFS.cc
+++ b/src/os/bluestore/BlueFS.cc
@@ -998,10 +998,14 @@ bool BlueFS::_should_compact_log()
   float ratio = (float)current / (float)expected;
   dout(10) << __func__ << " current 0x" << std::hex << current
 	   << " expected " << expected << std::dec
-	   << " ratio " << ratio << dendl;
-  if (current < g_conf->bluefs_log_compact_min_size ||
-      ratio < g_conf->bluefs_log_compact_min_ratio)
+	   << " ratio " << ratio
+	   << (log_flushing ? " (async compaction in progress)" : "")
+	   << dendl;
+  if (log_flushing ||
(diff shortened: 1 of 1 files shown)


liewegas:
@somnathr

somnathr:
Will try..

src/os/bluestore/BlueFS.cc
```diff
@@ -998,10 +998,14 @@ bool BlueFS::_should_compact_log()
   float ratio = (float)current / (float)expected;
   dout(10) << __func__ << " current 0x" << std::hex << current
 	   << " expected " << expected << std::dec
-	   << " ratio " << ratio << dendl;
-  if (current < g_conf->bluefs_log_compact_min_size ||
-      ratio < g_conf->bluefs_log_compact_min_ratio)
+	   << " ratio " << ratio
+	   << (log_flushing ? " (async compaction in progress)" : "")
+	   << dendl;
+  if (log_flushing ||
```
varadakari: Looks good. 
Should we wait till the compaction finishes? Anyways we will not have enough work in compaction if we trigger also. May be we can wait till the old extents are released.

liewegas:
I think we can just skip it in this call path.  We're only doing compaction in sync_metadata() because it's a convenient time to check (and maybe block).  It's probably a bad place to do it, though.. we probably want a separate thread doing the compaction so we don't affect calling threads.  That's a bigger change...

varadakari:
Agreed. we might have to get some throttling in place. We can have some naive algorithm to see how much we have to process or percentage to complete. So that we can make some decision recompact (if needed).

PR #11012, opened 2016-09-08
os/bluestore/BlueFS: For logs of rocksdb & bluefs only use directio.

majianpeng:
Now it use bluefs_buffered_io to control whether use buffer or directio
when write. But in fact for logs of rocksdb & bluefs, whether
bluefs_buffer_io is true or false, the logs only need directio.


markhpc:
This appears to be causing a ~10% performance regression with 4K random writes on NVMe devices.

markhpc:
Upon further testing, it looks like the majority, if not all, of the effect is coming from the second part of the conditional, ie "h->writer_type == WRITER_WAL".  I don't know why this would be the case however.

liewegas:
Huh.  For now, let's just switch the conditional to just ino == 1.

PR #11014, opened 2016-09-08
os/bluestore: make assert conditional with macro for allocator

chhabaramesh:
Change makes the is_allocated related code conditional with a macro.
Rest of code split with different assert so that basic check are still performed.

### DIFF
--- a/src/os/bluestore/BitAllocator.h
+++ b/src/os/bluestore/BitAllocator.h
@@ -8,7 +8,6 @@
 #ifndef  CEPH_OS_BLUESTORE_BITALLOCATOR_H
 #define CEPH_OS_BLUESTORE_BITALLOCATOR_H
 
-#define debug_assert assert
 
 #include <assert.h>
 #include <stdint.h>
@@ -19,6 +18,13 @@
 #include "include/intarith.h"
 #include "os/bluestore/bluestore_types.h"
 
+#define alloc_assert assert
+
+#ifdef BIT_ALLOCATOR_DEBUG
+#define alloc_dbg_assert(x) assert(x)
+#else
+#define alloc_dbg_assert(x) (static_cast<void> (0))
+#endif
 
 class BitAllocatorStats {
 public:
@@ -118,7 +124,7 @@ public:
       return NULL;
     }
 
-    debug_assert(cur_idx < (int64_t)m_list->size());
+    alloc_assert(cur_idx < (int64_t)m_list->size());
     return &(*m_list)[cur_idx];
   }
 
@@ -127,7 +133,7 @@ public:
   }
   void decr_idx() {
     m_cur_idx--;
-    debug_assert(m_cur_idx >= 0);
+    alloc_assert(m_cur_idx >= 0);
   }
 };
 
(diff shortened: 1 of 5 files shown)


src/os/bluestore/BitAllocator.h
```diff
@@ -19,6 +18,18 @@
...
+
+#ifdef BIT_ALLOCATOR_DEBUG
+#define alloc_dbg_assert(x) assert(x)
+#else
+#define alloc_dbg_assert(x) noop(x)
+#endif
+
+static inline void noop(bool var)
+{
+}
```
liewegas: This might not get optimized out.  Better to do a simple

 #define alloc_dbg_assert(x)

chhabaramesh: @liewegas , changed dummy function to what you suggested from assert.h. Please have a look now. I also wanted to define this macro BIT_ALLOCATOR_DEBUG for test case compilation but could not make it work :(.

src/os/bluestore/BlueStore.cc
```diff
@@ -5651,6 +5653,29 @@ ObjectMap::ObjectMapIterator BlueStore::get_omap_iterator(
...
+  assert(new_val > 0);
+  if (min_min_alloc_size > 0 &&
+      new_val >= min_min_alloc_size) {
+    return;
+  }
+
+  if (min_min_alloc_size > 0) {
+    derr << "warning: bluestore_min_alloc_size value decreased from "
+         << min_min_alloc_size << " to " << new_val << "."
+         << " Decreased value could have perofrmanc impact." << dendl;
```
ifed01: typo

chhabaramesh:
@ifed01  @liewegas , fixed the type on message. Please have a look now.

src/os/bluestore/BlueStore.cc
```diff
@@ -5696,6 +5721,21 @@ int BlueStore::_open_super_meta()
     }
   }
 
+  // Min min_alloc_size
+  {
+    bufferlist bl;
+    min_min_alloc_size = 0;
```
liewegas: = min_alloc_size ?  for @somnathr, who will run this on an existing store :)

liewegas: = min_alloc_size

for base case (after mkfs).  or when someone upgrades.

src/os/bluestore/BlueStore.cc
```diff
@@ -5651,6 +5653,29 @@ ObjectMap::ObjectMapIterator BlueStore::get_omap_iterator(
 
 // -----------------
 // write helpers
+void BlueStore::_save_min_min_alloc_size(uint64_t new_val)
+{
+  assert(new_val > 0);
+  if (min_min_alloc_size > 0 &&
+      new_val >= min_min_alloc_size) {
```
liewegas: if new_val > min_min_alloc_size we want to warn with something like

"min_min_alloc_size $minmin < min_alloc_size $min, may impact performance"

chhabaramesh: @liewegas , you meant we should put this message in both cases when
min_min_alloc_size > min_alloc_size and
min_min_alloc_size < min_alloc_size ?

Because once min_alloc_size is increased, we will keep seeing that message every time we boot, is that ok?

liewegas: Yeah, exactly.  Thanks!

src/os/bluestore/BlueStore.cc
chhabaramesh: It is always set after call to _save_min_min_alloc_size. Taking value 0 as argument to this function identifies the condition where we always need to save the min_min_alloc_size. 
The min_alloc_size is set before call to this function. So the cases:

1. min_min_alloc_size : 0, then we need to save current min_alloc_size as min_min_alloc_size. That handles upgrade and mkfs cases where there is no older value.
2. min_min_alloc_size < min_alloc_size, we need to put message and do nothing in this case.
3. min_min_alloc_size > min_alloc_size, we need to put message and save new value.

In all cases at end of this function, min_min_alloc_size == min_alloc_size.

I think it handles all cases ?

chhabaramesh:
@liewegas , a gentle reminder to have a look now and merge if looks ok.

PR #11015, opened 2016-09-08
cmake: make py3 a nice-to-have

tchaikov:
python3 is not a hard requirement to build ceph, so make it optional.
add an option named "WITH_PY3" which accepts ON, OFF, or CHECK.


wjwithagen:
@tchaikov 
Just completed a testrun of head on FreeBSD with only extra my python3 patch.
And that completes all the way thru to testing. So that is good.

I'll give your commit a run for its money this afternoon.

src/CMakeLists.txt
```diff
@@ -200,8 +200,20 @@ endif()
...
+  find_package(Python3Libs 3 QUIET)
+  if(PYTHON3INTERP_FOUND AND PYTHON3LIBS_FOUND)
+    set(WITH_PY3 ON)
+  else()
+    set(WITH_PY3 OFF)
+  endif()
+elseif(WITH_PY3)
+  find_package(Python3Interp 3 REQUIRED)
+  find_package(Python3Libs 3 REQUIRED)
+endif()
```
Why not move all this into `src/pybind/CMakeLists.txt` 
that way everything is close together.
Visually it looks like this will work.

tchaikov: i didn't check if other included `CMakeLists.txt` scripts are using the variables exposed by `find_package(Python3Interp 3)` and `find_package(Python3Libs)` or not. so not sure about this.

wjwithagen:
@tchaikov 
LGTM
My build is pleased and completes oke.

So now I can go back and start figuring out why my OSDs are not noup-ing only to the booting state.

tchaikov:
@wjwithagen thanks for testing! @jdurgin what do you think?

changelog

- rename `WITH_PY3` to `WITH_PYTHON3`, which matches the expectation in http://tracker.ceph.com/issues/17103.

jdurgin:
Looks good for the cmake part of http://tracker.ceph.com/issues/17103. Thanks!

PR #11016, opened 2016-09-08
mds: Set mds_snap_max_uid to 4294967294

wido:
Since kernel version 2.6 the Linux kernel supports 32-bit integers
and thus the limit is no longer 65536.

By setting this to a higher default value we make sure that all users
will be allowed to create snapshots in the future by default.


wido:
@jcsp ping! :)

We might want to move this setting to a 'ceph fs set <fs> snap_min_uid X' / 'ceph fs set <fs> snap_min_max Y' setting as well in the future?

jcsp:
Change looks reasonable, added to list for testing.

I agree that moving this setting into FSMap would make sense, because it's a per-filesystem thing more than a per-daemon thing.

wido:
@jcsp Yes, I think moving it seems logical to me as well. When this config setting was introduced a long time ago this was the only way.

Let's up the max UID limit here and then move it to the FSMap.

PR #11020, opened 2016-09-08
DNM: include/small_encoding: unroll varint encoding loop

liewegas:
This unrolls the loop, and also encodes into a single buffer and does a
single bufferlist append.

Also rewrite the lowz and signed variants in terms of the generic varint to
avoid code duplication.


dachary:

dachary:

PR #11022, opened 2016-09-08
os/bluestore: must write onode when do truncate

yuyuyu101:
If this is a new object and offset is 0, we still need to write onode

Partial revert 420150d
And set exists = true if previous remove this obj

### DIFF
--- a/src/os/bluestore/BlueStore.cc
+++ b/src/os/bluestore/BlueStore.cc
@@ -7540,6 +7540,9 @@ int BlueStore::_do_truncate(
 	   << " 0x" << std::hex << offset << std::dec << dendl;
   _dump_onode(o, 30);
 
+  o->exists = true;
+  txc->write_onode(o);
+
   if (offset == o->onode.size)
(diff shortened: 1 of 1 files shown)


src/os/bluestore/BlueStore.cc
```diff
@@ -7540,6 +7540,9 @@ int BlueStore::_do_truncate(
 	   << " 0x" << std::hex << offset << std::dec << dendl;
   _dump_onode(o, 30);
 
+  o->exists = true;
```
liewegas: touch, write, and zero are the only ops that create an object.

liewegas:
What's triggering this?  I don't think truncate should need to dirty the onode if it doesn't change the size.

src/os/bluestore/BlueStore.cc
yuyuyu101: oh sorry, I made a mistake. truncate doesn't imply creating new object

PR #11024, opened 2016-09-08
osd/OSD.cc: use refrence instead of copying

stiopaa1:

### DIFF
--- a/src/osd/OSD.cc
+++ b/src/osd/OSD.cc
@@ -5180,7 +5180,7 @@ void OSD::handle_pg_stats_ack(MPGStatsAck *ack)
     ++p;
 
     if (ack->pg_stat.count(pg->info.pgid.pgid)) {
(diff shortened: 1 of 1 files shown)


liewegas:
This is only a few bytes to copy; i doubt it'll be slower than a pointer indirection

cbodley:
note that this is an example of the map anti-pattern `if (map.count(key)) { access map[key]; }` which does two searches. can you change that to `auto i = map.find(key); if (i != map.end()) { ...` while you're at it?

stiopaa1:
@liewegas 
You are probably right - haven't noticed that it's so small.

@cbodley 
Yes, this was supposed to be my next commit:)

liewegas:
@stiopaa1 ping

stiopaa1:
@liewegas
 I've closed this.

PR #11025, opened 2016-09-08
systemd: Flush OSD Journal after stop

wido:
After stopping a OSD flush its journal to the data store. This makes sure
the data store holds a full copy of the OSD's data without a need of the journal
device.

When using external journals this makes the OSD disks more hot-pluggable between
different nodes.

On a different node one can simply create a blank journal for the OSD and run the
OSD again.

### DIFF
--- a/systemd/ceph-osd@.service
+++ b/systemd/ceph-osd@.service
@@ -11,6 +11,7 @@ EnvironmentFile=-/etc/sysconfig/ceph
 Environment=CLUSTER=ceph
 ExecStart=/usr/bin/ceph-osd -f --cluster ${CLUSTER} --id %i --setuser ceph --setgroup ceph
 ExecStartPre=/usr/lib/ceph/ceph-osd-prestart.sh --cluster ${CLUSTER} --id %i
+ExecStopPost=/usr/bin/ceph-osd -f --cluster ${CLUSTERD} --id %i --setuser ceph --setgroup ceph --flush-journal
 ExecReload=/bin/kill -HUP $MAINPID
 ProtectHome=true
 ProtectSystem=full


wido:
This is how it looks when starting and stopping a OSD:

```
Sep 09 00:30:52 alpha systemd[1]: Started Ceph object storage daemon.
Sep 09 00:30:52 alpha ceph-osd[3658]: starting osd.0 at :/0 osd_data /var/lib/ceph/osd/ceph-0 /var/lib/ceph/osd/ceph-0/journal
Sep 09 00:30:53 alpha ceph-osd[3658]: 2016-09-09 00:30:53.144664 7f6a7f0808c0 -1 osd.0 81 log_to_monitors {default=true}
Sep 09 00:37:21 alpha systemd[1]: Stopping Ceph object storage daemon...
Sep 09 00:37:21 alpha ceph-osd[3658]: 2016-09-09 00:37:21.529811 7f6a57c4f700 -1 osd.0 86 *** Got signal Terminated ***
Sep 09 00:37:21 alpha ceph-osd[3658]: 2016-09-09 00:37:21.611826 7f6a57c4f700 -1 osd.0 86 shutdown
Sep 09 00:37:21 alpha ceph-osd[3784]: 2016-09-09 00:37:21.834143 7f40039a88c0 -1 flushed journal /var/lib/ceph/osd/ceph-0/journal for object store /var/lib/ceph/osd/ceph-0
Sep 09 00:37:21 alpha systemd[1]: Stopped Ceph object storage daemon.
```

As you can see, it flushed the journal after the OSD has stopped.

gregsfortytwo:
Doesn't this slow down every daemon restart, for instance when upgrading Ceph or restarting to update the OS? It might be a good option but at first glance I don't think we want to do it by default.

wido:
@gregsfortytwo From what I've tested a flush takes 1 to 2 seconds. Most journal data is already flushed and usually the journal doesn't contain a lot of data.

My reasoning behind this is that not everybody is aware of the fact that without flushing you need the journal disk AND data disk to have a complete OSD.

The idea is that this protects users and will have a fully consistent data disk after a clean shutdown of the OSD/system.

So on a ~24 OSD system a restart of all OSDs will take let's say 30 seconds more.

wido:
Thinking about this a bit more: What about a ceph-osd-poststop.sh script?

This can then check for 'osd_flush_journal_after_stop' and flush if that boolean is set to true?

I would vote for setting it to true by default.

gregsfortytwo:
I agree it's convenient for users. I'm just concerned this may be one of the innocuous-seeming changes that blows up in your face when you hit failure conditions like flapping OSDs or something.

@athanatos, thoughts? This is more your area than mine.

athanatos:
I'm extremely uncertain about this.  I'd actually be much more comfortable with the osd flushing the store on it's own during a clean shutdown (are we sure it doesn't already)?

wido:
@athanatos I agree, if the OSD would do this on shutdown that would be even better.

Afaik it doesn't do that currently, so maybe we want to add that?

liewegas:
I agree that the OSD should do this in shutdown() instead of doing it with extra scripts.

wido:
Ok, see #11249

Closing this one

PR #11026, opened 2016-09-08
tools/cephfs: add pg_files command

jcsp:
tools/cephfs: add pg_files command
    
    Sometimes users know that particular data pool PGs
    have been damaged, and they would like to scan
    their files to work out which ones might have
    been affected.
    
    Fixes: http://tracker.ceph.com/issues/17249

### DIFF
--- a/src/tools/cephfs/PgFiles.cc
+++ b/src/tools/cephfs/PgFiles.cc
@@ -1,0 +1,192 @@
+// -*- mode:C++; tab-width:8; c-basic-offset:2; indent-tabs-mode:t -*-
+// vim: ts=8 sw=2 smarttab
+/*
+ * Ceph - scalable distributed file system
+ *
+ * Copyright (C) 2016 Red Hat
+ *
+ * This is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License version 2.1, as published by the Free Software
+ * Foundation.  See file COPYING.
+ *
+ */
+
+#include "common/errno.h"
+#include "osdc/Striper.h"
+
+#include "PgFiles.h"
+
+
+#define dout_subsys ceph_subsys_mds
+#undef dout_prefix
+#define dout_prefix *_dout << "pgeffects." << __func__ << ": "
+
+int PgFiles::init()
+{
+  int r = ceph_create_with_context(&cmount, g_ceph_context);
+  if (r != 0) {
+    return r;
+  }
+
+  return ceph_init(cmount);
+}
+
+PgFiles::PgFiles(Objecter *o, std::set<pg_t> pgs_)
+  : objecter(o), pgs(pgs_)
+{
+  for (const auto &i : pgs) {
+    pools.insert(i.m_pool);
+  }
+}
+
+PgFiles::~PgFiles()
+{
+  ceph_release(cmount);
+}
+
+void PgFiles::hit_dir(std::string const &path)
+{
+  dout(10) << "entering " << path << dendl;
+
+  ceph_dir_result *dr = nullptr;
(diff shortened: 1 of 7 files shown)


jcsp:
Checked this doesn't currently break anything in fs suite but it needs its own test too

src/tools/cephfs/PgFiles.cc
```diff
@@ -0,0 +1,192 @@
...
+  int r = ceph_get_path_layout(cmount, path.c_str(), &l_stripe_unit,
+                               &l_stripe_count, &l_object_size,
+                               &l_pool_id);
+
+  struct file_layout_t layout;
+  layout.stripe_unit = l_stripe_unit;
+  layout.stripe_count = l_stripe_count;
+  layout.object_size = l_object_size;
+  layout.pool_id = l_pool_id;
+  if (r != 0) {
```
gregsfortytwo: This delayed erro check isn't kinds weird?

src/tools/cephfs/PgFiles.cc
```diff
@@ -0,0 +1,192 @@
...
+
+  ceph_dir_result *dr = nullptr;
+  int r = ceph_opendir(cmount, "/", &dr);
+  if (r != 0) {
+    derr << "Failed to open path: " << cpp_strerror(r) << dendl;
+    return;
+  }
+
+  struct dirent de;
+  while((r = ceph_readdir_r(cmount, dr, &de)) != 0) {;
```
gregsfortytwo: Gratuitous semi-colon? Or is there some preventive thing going on?

src/tools/cephfs/PgFiles.cc
```diff
@@ -0,0 +1,192 @@
...
+        pg_num_mask = pool->get_pg_num_mask();
+        pg_num = pool->get_pg_num();
+      }
+    });
+    if (r != 0) {
+      // Can happen if layout pointed to pool not in osdmap, for example
+      continue;
+    }
+
+    target.m_seed = ceph_stable_mod(target.ps(), pg_num, pg_num_mask);
```
gregsfortytwo: Are there no OSDMap or otherwise more "public" functions exposing this?

gregsfortytwo:
My main concern here is that it will get used in place of a proper fsck sequence. It doesn't reconstruct the back traces on head objects, for instance.

Not sure if we should just note that in the docs or if there's a more definite care we can put in place (automaticallly re-creating any lost heads we detect here?)

src/tools/cephfs/PgFiles.cc
jcsp: Indeed it is, fixed.

src/tools/cephfs/PgFiles.cc
jcsp: Fixed

src/tools/cephfs/PgFiles.cc
jcsp: Nope, in existing use in Objecter it's not OSDMap that does this part

jcsp:
I've updated this, and added a 'danger' block to the docs telling people to remove files and replace them rather than overwriting in place.  Ultimately we should also be repairing backtraces from forward scrub probably.

(Tested by https://github.com/ceph/ceph-qa-suite/pull/1216)

jcsp:
New test passing here: http://pulpito.ceph.com/jspray-2016-10-20_14:39:41-fs-wip-jcsp-testing-20161020-distro-basic-mira/487477/

jcsp:
@gregsfortytwo this should be good to go (and https://github.com/ceph/ceph/pull/11578 was inspired by this case too in case that wasn't obvious :-))

PR #11028, opened 2016-09-09
Set ARM_CRC_FLAGS from the CRC test rather than ARM_NEON_FLAGS

dmick:
Fixes: http://tracker.ceph.com/issues/17250


tchaikov:
lgtm.

dmick:
Yay, with this fix make check passes save for ceph-disk

PR #11030, opened 2016-09-09
os/bluestore: fix offset bug in _do_write_small.

amoxic:
The statement "if (ep->logical_offset + b->get_blob().get_ondisk_length() <= offset)" which try to judge if the writing offset is in the current Blob may be wrong.

Signed-off-by: feng yu amoxic@126.com

### DIFF
--- a/src/os/bluestore/BlueStore.cc
+++ b/src/os/bluestore/BlueStore.cc
@@ -6964,7 +6964,7 @@ void BlueStore::_do_write_small(
   if (ep != o->extent_map.extent_map.begin()) {
     --ep;
     b = ep->blob;
-    if (ep->logical_offset + b->get_blob().get_ondisk_length() <= offset) {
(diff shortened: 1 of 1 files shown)


src/os/bluestore/BlueStore.cc
```diff
@@ -6964,7 +6964,7 @@ void BlueStore::_do_write_small(
   if (ep != o->extent_map.extent_map.begin()) {
     --ep;
     b = ep->blob;
-    if (ep->logical_offset + b->get_blob().get_ondisk_length() <= offset) {
+    if ((ep->logical_offset - ep->blob_offset) + b->get_blob().get_ondisk_length() <= offset) {
```
ifed01: No need for parentheses.
Have you run ceph_test_objectstore for your patch?

amoxic: I ran ceph_test_objectstore(only BlueStore), here is the result.

[==========] 61 tests from 1 test case ran. (9295011 ms total)
[  PASSED  ] 61 tests.

PR #11031, opened 2016-09-09
rbd: return error if we specified a wrong image name for rbd du

yangdongsheng:
Fixes: http://tracker.ceph.com/issues/16987

### DIFF
--- a/src/tools/rbd/action/DiskUsage.cc
+++ b/src/tools/rbd/action/DiskUsage.cc
@@ -110,12 +110,14 @@ static int do_disk_usage(librbd::RBD &rbd, librados::IoCtx &io_ctx,
   uint64_t used_size = 0;
   uint64_t total_prov = 0;
   uint64_t total_used = 0;
+  bool found = false;
   std::sort(names.begin(), names.end());
   for (std::vector<string>::const_iterator name = names.begin();
        name != names.end(); ++name) {
     if (imgname != NULL && *name != imgname) {
       continue;
     }
+    found = true;
 
     librbd::Image image;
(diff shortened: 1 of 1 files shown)


src/tools/rbd/action/DiskUsage.cc
```diff
@@ -206,6 +208,10 @@ static int do_disk_usage(librbd::RBD &rbd, librados::IoCtx &io_ctx,
       ++count;
     }
   }
+  if (!found) {
+    std::cerr << "specified image of " << imgname << "is not found." << std::endl;
+    return -EINVAL;
```
vshankar: -ENOENT?

src/tools/rbd/action/DiskUsage.cc
```diff
@@ -206,6 +208,10 @@ static int do_disk_usage(librbd::RBD &rbd, librados::IoCtx &io_ctx,
       ++count;
     }
   }
+  if (!found) {
+    std::cerr << "specified image of " << imgname << "is not found." << std::endl;
```
dillaman: Minor: remove "of" and add extra space before "is not found".

yangdongsheng: Okey

src/tools/rbd/action/DiskUsage.cc
yangdongsheng: Good suggestion

src/tools/rbd/action/DiskUsage.cc
yangdongsheng: @dillaman this is already updated, but I am not sure why it's still expanded here in github.

dillaman:
lgtm

PR #11034, opened 2016-09-09
librbd: potential seg fault when blacklisting an image client

dillaman:
Fixes: http://tracker.ceph.com/issues/17251


vshankar:
LGTM

PR #7649, opened 2016-02-15
librados: implement async watch/unwatch

yuyuyu101:

PR #11035, opened 2016-09-09
librados: Fix deadlock in watch_flush

xiaoxichen:
In previous code, in the watch_flush, it is waiting on conditon
with holding the "lock". The condition will only be signal by
finisher thread, but sadly, in some cases,when finisher queue
is not empty, some context need to take the "lock", thus deadlock.

Fixes:http://tracker.ceph.com/issues/17252

Signed-off-by: Xiaoxi Chen xiaoxchen@ebay.com

### DIFF
--- a/src/librados/RadosClient.cc
+++ b/src/librados/RadosClient.cc
@@ -72,6 +72,7 @@ librados::RadosClient::RadosClient(CephContext *cct_)
     instance_id(0),
     objecter(NULL),
     lock("librados::RadosClient::lock"),
+    shutdown_lock("librados::RadosClient::shutdown_lock"),
     timer(cct, lock),
     refcnt(1),
     log_last_version(0), log_cb(NULL), log_cb_arg(NULL),
@@ -324,6 +325,7 @@ int librados::RadosClient::connect()
 
 void librados::RadosClient::shutdown()
 {
+  Mutex::Locker l(shutdown_lock);
   lock.Lock();
   if (state == DISCONNECTED) {
     lock.Unlock();
@@ -338,7 +340,9 @@ void librados::RadosClient::shutdown()
   if (state == CONNECTED) {
     if (need_objecter) {
       // make sure watch callbacks are flushed
+      lock.Unlock();
       watch_flush();
+      lock.Lock();
     }
     finisher.wait_for_empty();
     finisher.stop();
--- a/src/librados/RadosClient.h
+++ b/src/librados/RadosClient.h
@@ -62,6 +62,7 @@ private:
   Objecter *objecter;
 
   Mutex lock;
+  Mutex shutdown_lock;
   Cond cond;
   SafeTimer timer;
   int refcnt;


yuyuyu101:
we could call this via:

int librados::Rados::watch_flush()
{
  if (!client)
    return -EINVAL;
  return client->watch_flush();
}

so lock.is_locked() may be false

yuyuyu101:
I suggest modify "watch_flush" in RadosClient::shutdown, directly call "objecter->linger_callback_flush(new C_SafeCond(&mylock, &cond, &done));" instead of calling watch_flush

yuyuyu101:
@dillaman what do you think

xiaoxichen:
@yuyuyu101 I guess you may mean objecter->linger_callback_flush(new C_SafeCond(&**lock**, &cond, &done));

src/librados/RadosClient.cc
```diff
@@ -361,15 +361,13 @@ void librados::RadosClient::shutdown()
...
 {
   ldout(cct, 10) << __func__ << " enter" << dendl;
-  Mutex mylock("RadosClient::watch_flush::mylock");
   Cond cond;
   bool done;
-  objecter->linger_callback_flush(new C_SafeCond(&mylock, &cond, &done));
+  objecter->linger_callback_flush(new C_SafeCond(&lock, &cond, &done));
 
-  mylock.Lock();
+  assert(lock.is_locked());
```
dillaman: librados::Rados::watch_flush() invokes this method -- therefore  ```lock``` won't be locked

dillaman:
Do we know when in the cinder code this deadlocked? I can only see this happening if you attempt to shut down the rados connection while you still have an open image.

xiaoxichen:
@dillaman it always hangs in this piece of code, after the remove_snap

with RBDVolumeProxy(self, volume_name) as volume:
            try:
                volume.unprotect_snap(snap_name)
                LOG.debug('******* SNAPSHOT UNPROTECTED *******')
            except self.rbd.ImageBusy:
                raise exception.SnapshotIsBusy(snapshot_name=snap_name)
            volume.remove_snap(snap_name)

        LOG.debug('******* SNAPSHOT DELETED *******')

dillaman:
Why not drop the lock before calling "watch_flush" and picking it back up again and re-checking for CONNECTED before proceeding with the rest of the logic?

yuyuyu101:
this commit move watch_flush outside lock.. I don't know the reason... sorry for the poor commit message

xiaoxichen:
@dillaman seems that will have racing condition
Say 2 concurrent call to shutdown(), the first one dropped the lock and in "watch_flush",the second one take the lock and go just before "watch_flush", after dropping the lock. 

Then, the first one finished the "watch_flush" call,  picking the lock back continue to destroy the finisher.  Now the second one go ahead  in "wathc_flush", insert the Safe_cond to finisher queue --- as the finisher thread is stopped, the second one will be blocked in the cond infinitely.

dillaman:
I should be able to fix the immediate subtle race condition within librbd::ImageWatcher and I'll open a new PR for that shortly. I think this librados issue needs to be fixed just in case someone else hits it, but perhaps we can wait a little while to ensure good runtime and test execution against any changes to librados (especially if we don't know what 7109de7 was designed to fix).

xiaoxichen:
@dillaman 7109de7b80cd50e6ae121eb9e3af033de21896da seems to fix the issue in teuthology test run, see your comment on https://github.com/ceph/ceph/pull/7649

dillaman:
Proposed librbd fix is in PR #11037

liewegas:
@dillaman should this be needs-qa then?

dillaman:
@liewegas librbd already has a workaround for this issue that has been fully backported. It is probably a good idea to fix this in librados as well, but I think this needs additional work since I'd be worried about a potential for the state to change while the lock is not held during the flush.

tchaikov:
http://pulpito.ceph.com/kchai-2017-12-01_08:15:14-rados-wip-kefu-testing-2017-12-01-1256-distro-basic-mira/

gregsfortytwo:
As this is so old and we don't have tests that can detect the issue, I'm closing this PR but leaving the tracker ticket available for future work.

Please feel free to re-open this or update http://tracker.ceph.com/issues/17252 if you're bumping into this issue!

PR #11036, opened 2016-09-09
Added append functionality to rados tool.

tomy-ac:
Signed-off-by: Tomy Cheru <tomy.cheru@sandisk.com>

### DIFF
--- a/src/tools/rados/rados.cc
+++ b/src/tools/rados/rados.cc
@@ -398,12 +398,8 @@ static int do_put(IoCtx& io_ctx, RadosStriper& striper,
 		  bool use_striper)
 {
   string oid(objname);
-  bufferlist indata;
-  bool stdio = false;
-  if (strcmp(infile, "-") == 0)
-    stdio = true;
-
-  int ret;
+  bool stdio = (strcmp(infile, "-") == 0);
+  int ret = 0;
   int fd = STDIN_FILENO;
   if (!stdio)
     fd = open(infile, O_RDONLY);
@@ -411,11 +407,11 @@ static int do_put(IoCtx& io_ctx, RadosStriper& striper,
     cerr << "error reading input file " << infile << ": " << cpp_strerror(errno) << std::endl;
     return 1;
   }
-  char *buf = new char[op_size];
   int count = op_size;
   uint64_t offset = 0;
   while (count != 0) {
-    count = read(fd, buf, op_size);
+    bufferlist indata;
+    count = indata.read_fd(fd, op_size);
     if (count < 0) {
       ret = -errno;
       cerr << "error reading input file " << infile << ": " << cpp_strerror(ret) << std::endl;
@@ -434,7 +430,6 @@ static int do_put(IoCtx& io_ctx, RadosStriper& striper,
       }
       continue;
     }
-    indata.append(buffer::ptr(buffer::create_static(count, buf)));
     if (use_striper) {
       if (offset == 0)
 	ret = striper.write_full(oid, indata);
@@ -446,7 +441,6 @@ static int do_put(IoCtx& io_ctx, RadosStriper& striper,
       else
 	ret = io_ctx.write(oid, indata, count, offset);
     }
-    indata.clear();
 
     if (ret < 0) {
       goto out;
@@ -457,7 +451,6 @@ static int do_put(IoCtx& io_ctx, RadosStriper& striper,
  out:
   if (fd != STDOUT_FILENO)
     VOID_TEMP_FAILURE_RETRY(close(fd));
-  delete[] buf;
   return ret;
 }
 
--- a/src/tools/rados/rados.cc
+++ b/src/tools/rados/rados.cc
@@ -85,6 +85,7 @@ void usage(ostream& out)
 "OBJECT COMMANDS\n"
 "   get <obj-name> [outfile]         fetch object\n"
 "   put <obj-name> [infile]          write object\n"
+"   append <obj-name> [infile]       append object\n"
 "   truncate <obj-name> length       truncate object\n"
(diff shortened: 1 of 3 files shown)


src/tools/rados/rados.cc
```diff
@@ -460,6 +461,53 @@ static int do_put(IoCtx& io_ctx, RadosStriper& striper,
...
 
+static int do_append(IoCtx& io_ctx, RadosStriper& striper,
+                  const char *objname, const char *infile, int op_size,
+                  bool use_striper)
+{
+  string oid(objname);
+  bufferlist indata;
+  bool stdio = false;
+  if (strcmp(infile, "-") == 0)
+    stdio = true;
```
athanatos: bool stdio = (strcmp(infile, "-") == 0);

strcmp acceptable here only because infile from OS.

src/tools/rados/rados.cc
```diff
@@ -460,6 +461,53 @@ static int do_put(IoCtx& io_ctx, RadosStriper& striper,
...
+
+  int ret;
```
athanatos: int ret = 0;

No uninitialized variables without a profoundly good reason.

src/tools/rados/rados.cc
```diff
@@ -460,6 +461,53 @@ static int do_put(IoCtx& io_ctx, RadosStriper& striper,
...
+  int fd = STDIN_FILENO;
+  if (!stdio)
+    fd = open(infile, O_RDONLY);
+  if (fd < 0) {
+    cerr << "error reading input file " << infile << ": " << cpp_strerror(errno) << std::endl;
+    return 1;
+  }
+  char *buf = new char[op_size];
```
athanatos: Use the right unique_ptr variant here.

src/tools/rados/rados.cc
```diff
@@ -460,6 +461,53 @@ static int do_put(IoCtx& io_ctx, RadosStriper& striper,
   return ret;
 }
 
+static int do_append(IoCtx& io_ctx, RadosStriper& striper,
+                  const char *objname, const char *infile, int op_size,
+                  bool use_striper)
+{
+  string oid(objname);
+  bufferlist indata;
```
athanatos: Declare this in the loop, avoids needing to clear it.

src/tools/rados/rados.cc
```diff
@@ -460,6 +461,53 @@ static int do_put(IoCtx& io_ctx, RadosStriper& striper,
...
+  if (fd < 0) {
+    cerr << "error reading input file " << infile << ": " << cpp_strerror(errno) << std::endl;
+    return 1;
+  }
+  char *buf = new char[op_size];
+  int count = op_size;
+  while (count != 0) {
+    count = read(fd, buf, op_size);
+    if (count < 0) {
+      ret = -errno;
```
athanatos: If this is read(2), you need to check for EINTR...

Yes, do_put also appears to have this bug.  I suggest you use safe_read from common/safe_io.h.

athanatos: Add a second commit fixing do_put while you're at it.

src/tools/rados/rados.cc
```diff
@@ -460,6 +461,53 @@ static int do_put(IoCtx& io_ctx, RadosStriper& striper,
...
+    }
+    indata.append(buffer::ptr(buffer::create_static(count, buf)));
+    if (use_striper) {
+      ret = striper.append(oid, indata, count);
+    } else {
+      ret = io_ctx.append(oid, indata, count);
+    }
+    indata.clear();
+
+    if (ret < 0) {
```
tchaikov: i am not sure we need to do the error handling this way. `append()` returns `0` on success, and a negative number on failure. so we can just let `while (count != 0)` take care of the loop condition, and use `break` if ret is not `0`.

src/tools/rados/rados.cc
```diff
@@ -460,6 +461,53 @@ static int do_put(IoCtx& io_ctx, RadosStriper& striper,
...
+  if (!stdio)
+    fd = open(infile, O_RDONLY);
+  if (fd < 0) {
+    cerr << "error reading input file " << infile << ": " << cpp_strerror(errno) << std::endl;
+    return 1;
+  }
+  char *buf = new char[op_size];
+  int count = op_size;
+  while (count != 0) {
+    count = read(fd, buf, op_size);
```
tchaikov: could just use `bufferlist::read_fd()`. it uses `safe_read()`.

athanatos:
@tomy-ac  Please fix do_put as well.

athanatos:
Ok, this branch should have two commits, one for append and one for fixing do_put -- please squash down the review comment commit.  Also, please modify the commit comments to start with 'tools/rados:' to make it easy to determine what the patch is modifying.

athanatos:
@tomy-ac ^

tomy-ac:
@athanatos  squashed commits as requested

athanatos:
@tomy-ac You missed the second part of my comment:

Also, please modify the commit comments to start with 'tools/rados:' to make it easy to determine what the patch is modifying.

For the first patch, 

tools/rados: Added append functionality to rados tool.

tomy-ac:
@athanatos commit comments modified per request

athanatos:
Thanks! Added needs-qa for the next round of integration testing.

yuriw:
http://pulpito.ceph.com/yuriw-2016-10-21_00:30:49-rados-wip-yuri-testing_2016_10_21-distro-basic-smithi/

http://pulpito.front.sepia.ceph.com/yuriw-2016-10-21_15:49:39-rados-wip-yuri-testing_2016_10_21-distro-basic-smithi/

PR #11038, opened 2016-09-09
rpm: move hdparm requirement to ceph-osd package

ktdreyer:
`ceph-objectstore-tool` runs hdparm, but that only happens on OSD nodes.  Move the package requirement to the ceph-osd sub-package.

This reduces dependency bloat for other nodes, since Mon, RGW, and MDS nodes do not need hdparm.

Refs: http://tracker.ceph.com/issues/17177

### DIFF
--- a/ceph.spec.in
+++ b/ceph.spec.in
@@ -227,7 +227,6 @@ Requires:      grep
 Requires:      xfsprogs
 Requires:      logrotate
 Requires:      util-linux
-Requires:      hdparm
 Requires:      cryptsetup
 Requires:      findutils
 Requires:      which
@@ -357,6 +356,7 @@ managers such as Pacemaker.
 Summary:	Ceph Object Storage Daemon
 Group:		System Environment/Base
 Requires:	ceph-base = %{epoch}:%{version}-%{release}
+Requires:	hdparm
 # for sgdisk, used by ceph-disk
 %if 0%{?fedora} || 0%{?rhel}
 Requires:	gdisk


badone:
LGTM

smithfarm:
@ktdreyer Do you plan to add an analogous commit for the deb packaging?

ktdreyer:
The reason I had not bothered with Debian yet is that it's hard to understand why `sdparm` was added (3ad6ccb4a636186695da9391474dc4a06d42f9f6) there years ago, and I wanted to fix that in a separate PR that removes sdparm (reverting that change), and then moves the hdparm dependency to the ceph-osd deb (as you suggest).

That felt like too many changes for something that currently has no test coverage in Teuthology, so I wanted to keep this PR simple to avoid regressions :)

b-ranto:
Looking at the sources, at least one of the tests do seem to use FileJournal (and by that hdparm) as well. We might want to add hdparm to ceph-test sub-package to cover for that.

ktdreyer:
@b-ranto should ceph-test just depend on ceph-osd in that case?

ktdreyer:
@gcsideal can you shed any light on why 3ad6ccb4a636186695da9391474dc4a06d42f9f6 adds sdparm ? Modern versions of hdparm work ok with SCSI devices, so we can drop sdparm now?

b-ranto:
@ktdreyer Good question, I don't think so. AFAIK it links directly to the libs and it does not call the binaries that are provided by the ceph packages. The hdparm is kinda a special case here since it is called directly as a binary from the FileJournal and as such hdparm propagates as a dependency for ceph-test.

I would not be too worried about it, though. This was broken before and there are very likely other missing dependencies for ceph-test package. I suppose this is the reason why hdparm is part of BuildRequires -- it needs to be there if the rpmbuild were to run make check.

badone:
@ktdreyer @b-ranto I think it's worth checking whether we actually need this code any more in master

https://github.com/ceph/ceph/blob/master/src/os/filestore/FileJournal.cc#L171-L239

Since it's designed to warn about an issue with disk write cache on pre 2.6.33 kernels I suspect we can live without it in the current master?

badone:
Further, if we lose this code we can drop the dependency on sdparm/hdparm right?

ktdreyer:
dropping it sounds good to me @badone

badone:
@liewegas would you mind making a call here?

liewegas:
Yeah, let's kill it!

badone:
OK, I'll create a PR to remove this code in the next couple of days, thanks!

ktdreyer:
> OK, I'll create a PR to remove this code in the next couple of days, thanks!

This is now https://github.com/ceph/ceph/pull/11073 - closing this PR in favor of that one. Thanks Brad.

PR #11044, opened 2016-09-11
librbd: ignore partial refresh error when acquiring exclusive lock

dillaman:
Fixes: http://tracker.ceph.com/issues/17227


trociny:
lgtm

PR #11047, opened 2016-09-12
ceph_mon: use readdir() as readdir_r() is deprecated

tchaikov:
see https://lwn.net/Articles/696469/, readdir_r() is deprecated by glibc
since 2.24. so let's use readdir() instead.


jecluis:
lgtm

PR #11052, opened 2016-09-12
OSD: avoid FileStore finisher deadlock in osd_lock when shutdown OSD

yuyuyu101:
When OSD receive signal to shutdown, it will try to call ObjectStore->flush
to make sure all pending IO done. But previous OSDMap committing IO may
complete Context in OSD::_committed_osd_maps which need to acquire osd_lock
firstly. So OSD shutdown thread need to wait for finisher done, and finisher
is acquiring osd_lock. dead lock happended

### DIFF
--- a/src/osd/OSD.cc
+++ b/src/osd/OSD.cc
@@ -6833,11 +6833,11 @@ void OSD::handle_osd_map(MOSDMap *m)
 void OSD::_committed_osd_maps(epoch_t first, epoch_t last, MOSDMap *m)
 {
   dout(10) << __func__ << " " << first << ".." << last << dendl;
-  Mutex::Locker l(osd_lock);
   if (is_stopping()) {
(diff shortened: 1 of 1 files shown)


liewegas:
lgtm

yuriw:
http://pulpito.ceph.com/yuriw-2016-09-19_19:42:27-rados-wip-yuri-testing2_2016_09_19-distro-basic-smithi/

rerun http://pulpito.ceph.com/yuriw-2016-09-20_14:50:21-rados-wip-yuri-testing2_2016_09_19-distro-basic-smithi/

yuriw:
@yuyuyu101 Pls reference a tracker ticket to this PR

yuyuyu101:
@yuriw sorry, I don't have created a ticket before. Do we need one?

PR #11054, opened 2016-09-12
rgw: eradicate dynamic memory allocation in RGWPostObj.

rzarzynski:
s3-tests hasn't found any regression here.

CC: @mattbenjamin, @cbodley.

### DIFF
--- a/src/rgw/rgw_op.cc
+++ b/src/rgw/rgw_op.cc
@@ -2894,22 +2894,6 @@ int RGWPostObj::verify_permission()
   return 0;
 }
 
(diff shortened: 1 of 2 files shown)


mattbenjamin:
@rzarzynski cool;  I think the same can probably be done in the rgw_file write path;  if for some reason you add it, I'll test

rzarzynski:
@mattbenjamin: great! I will work on `RGWWriteRequest` today to remove the dynallocs.

rzarzynski:
@mattbenjamin: the modification for `RGWWriteRequest` has been sent as a separated pull request - #11062.

PR #11058, opened 2016-09-12
[wip] os: improve a little performance by less If judgment

shun-s:
improve a little performance by less If judgment

Signed-off-by: shun-s song.shun3@zte.com.cn

### DIFF
--- a/src/os/filestore/FileJournal.cc
+++ b/src/os/filestore/FileJournal.cc
@@ -888,8 +888,14 @@ int FileJournal::prepare_multi_write(bufferlist& bl, uint64_t& orig_ops, uint64_
   // gather queued writes
   off64_t queue_pos = write_pos;
 
-  int eleft = g_conf->journal_max_write_entries;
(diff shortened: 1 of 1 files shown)


shun-s:
@markhpc 
this pr has been proposed for a long time, could you merge it?
or whether there are some modifications needed?
if that's the case,i'd like to modify, thanks

yuyuyu101:
I don't think this related to performance optimization. It's just moving "if else" to another place.

shun-s:
@yuyuyu101 
Originally,  ' if (bmax) ' and 'if (eleft)' are separatelly to be run many times in the while loop, likely 8 times in my running 4k random write. 
but after modification, only once needs before while loop.
so i think it can have a little benefit to the performance.

markhpc:
I guess the question I have is whether or not this actually shows a measurable difference in any tests?  Ultimately this is Sam's call, but generally we've been rather picky about making sure changes to filestore show clear improvement before merging them.

PR #11059, opened 2016-09-12
bluestore/BlueFS: revert direct IO for WRITER_WAL

markhpc:
When writer_type is WRITER_WAL, forcing direct IO when bluefs_buffered_io = true appears to yield a 6-8% performance degradation on NVMe hardware.  Revert the change for now.  In the future we will need to understand why direct IO in general is causing performance issues on our NVMe test hardware.


markhpc:
retest this please

PR #11061, opened 2016-09-13
os/bluestore: fix shard_info::dump()

xiexingguo:


xiexingguo:
2016-09-13 18:15:03.291922 7f274600b640  1 bluestore(store_test_temp_dir) fsck finish with 0 errors in 0.503416 seconds
[       OK ] ObjectStore/StoreTest.TooManyBlobsTest/2 (23270 ms)
[----------] 61 tests from ObjectStore/StoreTest (2663393 ms total)

[==========] 61 tests from 1 test case ran. (2663393 ms total)
[  PASSED  ] 61 tests.

liewegas:
key == name and key == '' are defined to be the same (although in memory we try to prevent key == name and always have key.empty() in that case).

xiexingguo:
@liewegas 
Thanks for the explanation.

Dropped that change:-)

PR #11062, opened 2016-09-13
rgw: eradicate dynamic memory allocation in RGWWriteRequest.

rzarzynski:
It would be really nice to have Boost 1.56 that brought `boost::optional::emplace()`.

CC: @mattbenjamin.

### DIFF
--- a/src/rgw/rgw_file.cc
+++ b/src/rgw/rgw_file.cc
@@ -945,9 +945,17 @@ namespace rgw {
     /* early quota check skipped--we don't have size yet */
     /* skipping user-supplied etag--we might have one in future, but
      * like data it and other attrs would arrive after open */
-    processor = select_processor(*static_cast<RGWObjectCtx *>(s->obj_ctx),
-				 &multipart);
-    op_ret = processor->prepare(get_store(), NULL);
+    processor.emplace(*static_cast<RGWObjectCtx *>(s->obj_ctx),
+                      s->bucket_info,
+                      s->bucket,
+                      s->object.name,
+                      /* part_size */
+                      s->cct->_conf->rgw_obj_stripe_size,
+                      s->req_id,
+                      s->bucket_info.versioning_enabled());
+    processor->set_olh_epoch(olh_epoch);
+    processor->set_version_id(version_id);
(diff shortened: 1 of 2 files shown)


src/rgw/rgw_file.cc
```diff
@@ -987,9 +995,17 @@ namespace rgw {
       data.swap(orig_data);
 
       /* restart processing with different oid suffix */
-      dispose_processor(processor);
-      processor = select_processor(*static_cast<RGWObjectCtx *>(s->obj_ctx),
-				   &multipart);
+      processor = boost::none;
```
cbodley: is this really needed? does `boost::optional` try to call `RGWPutObjProcessor_Atomic::operator=()` otherwise?

cbodley:
@rzarzynski have you looked at `RGWPutObj::execute()`? would the same strategy work there?

changing it now would cause conflicts with https://github.com/ceph/ceph/pull/10932, but we should get that eventually

rzarzynski:
@cbodley: I'm afraid that for `RGWPutObj::execute()` we would need `static_ptr` to handle the polymorphism. The `RGWPutObj` instantiates `RGWPutObjProcessor_Atomic` or `RGWPutObjProcessor_Multipart` depending on the `multipart` variable.

yehudasa:
@cbodley ping

dachary:

cbodley:
@rzarzynski should be able to use `optional::emplace()` here now?

rzarzynski:
@cbodley: yep, we should be perfectly able to do the switch now. I will send the update soon.

rzarzynski:
@cbodley, @mattbenjamin: fixed the commit and switched to `boost::optional::emplace`.

mattbenjamin:
@rzarzynski I'm fine with using this strategy here, will test

rzarzynski:
@mattbenjamin: thanks! :-)

oritwas:
@rzarzynski please rebase

yehudasa:
@rzarzynski what's the status of this one?

amitkumar50:

cbodley:
see https://github.com/ceph/ceph/pull/24453

PR #11064, opened 2016-09-13
make-srpm.sh: A simple script to make the srpm for ceph.

ilc:
This is a simple script to make the srpm for ceph based
on make-dist.

Yes, it is short, but it will stop many people from
having to solve the same problem.


liewegas:
:+1:

PR #11065, opened 2016-09-13
mon/OSDMonitor: should propose osdmap update when cluster addr changed

LiumxNL:
osd maybe rebind cluster addr but keep public addr unchanged in _committed_osd_maps() method, to be on the safe side, monitor should take cluster addr into account when determining whether should propose osdmap update.

Signed-off-by: Mingxin Liu mingxin@xsky.com

### DIFF
--- a/src/mon/OSDMonitor.cc
+++ b/src/mon/OSDMonitor.cc
@@ -2101,7 +2101,8 @@ bool OSDMonitor::preprocess_boot(MonOpRequestRef op)
 
   // already booted?
   if (osdmap.is_up(from) &&
-      osdmap.get_inst(from) == m->get_orig_source_inst()) {
+      osdmap.get_inst(from) == m->get_orig_source_inst() &&
+      osdmap.get_cluster_addr(from) == m->cluster_addr) {
     // yup.
     dout(7) << "preprocess_boot dup from " << m->get_orig_source_inst()
 	    << " == " << osdmap.get_inst(from) << dendl;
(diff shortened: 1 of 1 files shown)


LiumxNL:
@tchaikov would you mind to take a look? thanks!

src/mon/OSDMonitor.cc
```diff
@@ -2108,7 +2109,9 @@ bool OSDMonitor::prepare_boot(MonOpRequestRef op)
   if (osdmap.is_up(from)) {
     dout(7) << "prepare_boot was up, first marking down " << osdmap.get_inst(from) << dendl;
     // preprocess should have caught these;  if not, assert.
-    assert(osdmap.get_inst(from) != m->get_orig_source_inst());
+    bool addr_changed = (osdmap.get_inst(from) != m->get_orig_source_inst() ||
```
tchaikov: could also put ``osdmap.get_cluster_addr(from) != m->cluster_addr)`` on a new line. like
```c
assert(osdmap.get_inst(from) != m->get_orig_source_inst() ||
       osdmap.get_cluster_addr(from) != m->cluster_addr);
```

 as ideally, `assert()` is a no-op in a NDEBUG build, and compiler or static analyzer would complain at seeing a variable not being used anywhere.

tchaikov:
lgtm modulo the nit.

yuyuyu101:
@LiumxNL ping

LiumxNL:
@tchaikov

PR #11068, opened 2016-09-13
os/bluestore: fix a few leaks

liewegas:


src/os/bluestore/BlueStore.h
```diff
@@ -549,6 +549,12 @@ class BlueStore : public ObjectStore,
     bufferlist inline_bl;    ///< cached encoded map, if unsharded; empty=>dirty
 
     ExtentMap(Onode *o);
+    ~ExtentMap() {
+      auto p = extent_map.begin();
```
ifed01: you may use something like that instead:
extent_map.clear_and_dispose( [](Extent* e) {delete e;});

ifed01:
LGTM

src/os/bluestore/BlueStore.h
liewegas: Oh, cool!  Updated.

PR #11071, opened 2016-09-14
Stop hiding errors from run-tox.sh

dmick:
The grep -v should have been grep, but, why even bother?
Tox doesn't output that much; just dump it, and exit with
tox's error code.

Fixes: http://tracker.ceph.com/issues/17267


tchaikov:
lgtm.

PR #11072, opened 2016-09-14
ceph-disk tests: Let missing python interpreters be non-fatal

dmick:
Let tox ignore missing interpreters, and add py35 and py36 as well.
This lets tests work on Xenial, for example, which only has py35.


dmick:
Any comments, @alfredodeza?

PR #11075, opened 2016-09-14
run-cmake-check.sh: Actually run the tests

badone:
Actually run the tests from run-cmake-check.sh

### DIFF
--- a/run-cmake-check.sh
+++ b/run-cmake-check.sh
@@ -48,10 +48,10 @@ function run() {
 	$DRY_RUN ./install-deps.sh || return 1
     fi
     git submodule update --init --recursive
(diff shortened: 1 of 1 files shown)


liewegas:
rebase?

badone:
@liewegas sorry, fixed.

liewegas:
Oh good, I was going to suggest on the other PR that you just run do_cmake.sh.  FWIW the git submodule line is probably no longer needed now..

badone:
@liewegas thanks, I believe the git submodule line *is* still needed as I fell foul of this while testing (googletest/googlemock non-existent because that submodule not initialised). Thinking about getting rid of the modprobe of rbd.ko in a future commit though...

PR #11078, opened 2016-09-14
mds: remove fail-safe queueing replay request

ukernel:
MDSRank::queue_one_replay() does not create active request
immediately, it just queues corresponding C_MDS_RetryMessage
for execution. So the fail-safe code can queue an extra replay
request. This can cause replay requests be processed out-of-order

Fixes: http://tracker.ceph.com/issues/17271

### DIFF
--- a/src/mds/MDCache.cc
+++ b/src/mds/MDCache.cc
@@ -9073,12 +9073,6 @@ void MDCache::request_cleanup(MDRequestRef& mdr)
   // remove from map
   active_requests.erase(mdr->reqid);
 
-  // fail-safe!
-  if (was_replay && active_requests.empty()) {
-    dout(10) << " fail-safe queueing next replay op" << dendl;
-    mds->queue_one_replay();
-  }
-
   if (mds->logger)
     log_stat();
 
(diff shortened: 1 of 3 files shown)


jcsp:
I think this is breaking clientreplay because this queue_one_replay call was the only thing other than _dispatch that could trigger clientreplay_done.

Seeing test failure (timeout in clientreplay) here:
http://pulpito.ceph.com/jspray-2016-09-21_17:14:24-fs-wip-jcsp-greenish-distro-basic-mira/429390/

src/mds/MDCache.cc
```diff
@@ -9073,12 +9073,6 @@ void MDCache::request_cleanup(MDRequestRef& mdr)
   // remove from map
   active_requests.erase(mdr->reqid);
 
-  // fail-safe!
-  if (was_replay && active_requests.empty()) {
-    dout(10) << " fail-safe queueing next replay op" << dendl;
-    mds->queue_one_replay();
```
jcsp: We need at least to do the same check for whether to call clientreplay_done here that we do inside queue_one_replay.

ukernel: no need to do this if queue_one_replay() is properly called when finishing MDRequest.

ukernel:
updated

jcsp:

PR #11084, opened 2016-09-14
[DNM] Reduce deep scrub impact on normal operations

dzafman:


dzafman:
@athanatos I separated the regular scrub scan and the deep scrub reads.  Since writes are already blocked on the scrub range, I drop the PG lock before doing all the reads on the objects for deep scrub. Also, simplified the code having the primary use the same mechanism as the replicas.  Don't think using PGQueueable buys us anything.

src/osd/PG.cc
```diff
@@ -3465,8 +3465,10 @@ void PG::_request_scrub_map(
   hobject_t start, hobject_t end,
   bool deep, uint32_t seed)
 {
```
athanatos: This I like!  Very slick.

src/osd/PG.cc
```diff
@@ -3725,8 +3727,7 @@ void PG::_scan_snaps(ScrubMap &smap)
 }
 
 /*
- * build a scrub map over a chunk without releasing the lock
- * only used by chunky scrub
+ * build a scrub map over a chunk
  */
```
athanatos: This method iirc is triggered by a OpRequestRef event in the OpWQ -- queued in handle_replica_op<MOSDRepScrub,...> called from OSD::dispatch_op_fast.  I think what you want to do is replace that call to the templated handle_replica_op with a new custom one which queues an event you can requeue until you've finished the scan.

src/osd/PG.cc
```diff
@@ -3752,11 +3753,18 @@ int PG::build_scrub_map_chunk(
...
   }
 
-
-  get_pgbackend()->be_scan_list(map, ls, deep, seed, handle);
+  get_pgbackend()->be_scan_list(map, ls, seed, handle);
   _scan_rollback_obs(rollback_obs, handle);
   _scan_snaps(map);
 
+  if (deep) {
+    // XXX: Drop the PG lock because we are just reading the data and
```
dzafman: @athanatos Doesn't the requeue actually drop the PG lock between each group of N objects?  Even though this code drops the lock after it re-aquires  it, nothing else is done but packaging up and sending the response!

athanatos: Yeah, but it still sits on the thread for the entire duration.  Commonly, the OpWQ only has 1 thread/shard, so even if you drop the lock, no IO can happen.  I really think it would be better to do a bounded amount of work per work queue item.

dzafman: @athanatos Got it.  Didn't know about there being a single work queue thread.

src/osd/PG.cc
```diff
@@ -4080,10 +4078,13 @@ void PG::chunky_scrub(ThreadPool::TPHandle &handle)
...
           while (!boundary_found) {
             vector<hobject_t> objects;
+	    // If we need to exceed osd_scrub_chunk_max by calling this again,
+	    // use osd_scrub_chunk_min as maximum because the boundary checking
+	    // finds the largest possible list of objects on a boundary.
             ret = get_pgbackend()->objects_list_partial(
 	      start,
 	      cct->_conf->osd_scrub_chunk_min,
-	      cct->_conf->osd_scrub_chunk_max,
+	      loop == 0 ? cct->_conf->osd_scrub_chunk_max : cct->_conf->osd_scrub_chunk_min,
```
dzafman: @athanatos  Notice I slightly reduce the number of objects per chunk in the boundary checking code when snapshots cause us to exceed osd_scrub_chunk_max.

src/osd/PG.cc
athanatos: Yep.  More generally, the OpWQ interface essentially assumes that each item only requires a bounded amount of time/resources (for some super fuzzy version of "bounded") characterized by the cost parameter -- that way the prioritization logic has a chance to work.

athanatos: Oh, and you can configure more than 1 thread/shard, but it's still worth ensuring that each item can be processed in a bounded amount of time.

src/msg/Message.cc
```diff
@@ -79,7 +79,7 @@ using namespace std;
 #include "messages/MOSDPGTrim.h"
 #include "messages/MOSDPGMissing.h"
 #include "messages/MOSDScrub.h"
-#include "messages/MOSDRepScrub.h"
+#include "messages/MOSDBuildScrubMap.h"
```
athanatos: I'm not sure we want to change the name (if I suggested it, I am sorry!).  For debugging purposes, it would be nice if two osds in a mixed cluster agree on the name for the same message.

src/osd/PG.cc
```diff
@@ -3858,6 +3860,25 @@ void PG::replica_scrub(
...
+  scrubber.map_state = PG::Scrubber::REG_SCRUBBING;
+  osd->queue_for_build_scrub_map(this, g_conf->osd_scrub_map_start_cost);
+}
+
+void PG::replica_scrub(epoch_t queued, ThreadPool::TPHandle &handle)
+{
+  dout(20) << __func__ << dendl;
+  assert(scrubber.active_rep_scrub);
+  MOSDBuildScrubMap *msg = static_cast<MOSDBuildScrubMap *>(scrubber.active_rep_scrub->get_req());
+  ScrubMap map;
```
athanatos: This'll presumably have to be stashed between queue items?

src/msg/Message.cc
dzafman: @athanatos I just wanted to make the code change clearer.  You didn't suggest it.

src/osd/PG.cc
dzafman: @athanatos Yeah, I'll move it into _scrubber_.

src/common/config_opts.h
```diff
@@ -765,6 +766,7 @@ OPTION(osd_scrub_max_interval, OPT_FLOAT, 7*60*60*24)  // regardless of load
 OPTION(osd_scrub_interval_randomize_ratio, OPT_FLOAT, 0.5) // randomize the scheduled scrub in the span of [min,min*(1+randomize_ratio))
 OPTION(osd_scrub_chunk_min, OPT_INT, 5)
 OPTION(osd_scrub_chunk_max, OPT_INT, 25)
+OPTION(osd_deep_map_chunk_max, OPT_INT, 25)  // Replica deep-scrub map (snapshots can exceed osd_scrub_chunk_max, so divide up deep scrubbing)
```
athanatos: Might be better to just use _chunk_max, not sure.

src/common/config_opts.h
```diff
@@ -878,6 +880,13 @@ OPTION(osd_recovery_priority, OPT_U32, 5)
 // set default cost equal to 20MB io
 OPTION(osd_recovery_cost, OPT_U32, 20<<20)
 
+// XXX: Should the prioirity be based on scrub vs deep-scrub?
```
athanatos: Don't think so, it blocks client requests either way.

src/osd/PG.cc
```diff
@@ -3730,16 +3732,15 @@ void PG::_scan_snaps(ScrubMap &smap)
  */
 int PG::build_scrub_map_chunk(
   ScrubMap &map,
-  hobject_t start, hobject_t end, bool deep, uint32_t seed,
-  ThreadPool::TPHandle &handle)
+  hobject_t start, hobject_t end, uint32_t seed,
+  ThreadPool::TPHandle &handle, vector<hobject_t> &ls)
```
athanatos: Nit: by convention I've been passing handle last.  Also, per the style guide, ls should probably be passed by pointer since it's an out param?

src/osd/PG.cc
```diff
@@ -3847,6 +3850,30 @@ void PG::replica_scrub(
     return;
   }
 
+  scrubber.active_rep_scrub = op;
```
athanatos: I think this is wrong, looks like active_rep_scrub should only be set if we need op_applied or _applied_recovered_object_replica to requeue the op.

src/osd/PG.cc
```diff
@@ -3847,6 +3850,30 @@ void PG::replica_scrub(
...
+           << " objects " << scrubber.ls.size() << dendl;
+  assert(_lock.is_locked());
+  assert(scrubber.active_rep_scrub);
+  assert(scrubber.map_state != PG::Scrubber::NOT_BUILDING);
+  MOSDRepScrub *msg = static_cast<MOSDRepScrub *>(scrubber.active_rep_scrub->get_req());
+
+  if (pg_has_reset_since(queued)) {
+    return;
+  }
+  if (msg->map_epoch < info.history.same_interval_since) {
```
athanatos: Is this possible if !pg_has_reset_since(queued) ?

src/osd/PG.cc
```diff
@@ -3847,6 +3850,30 @@ void PG::replica_scrub(
     return;
   }
 
+  scrubber.active_rep_scrub = op;
+  scrubber.map_state = PG::Scrubber::REG_SCRUBBING;
+  osd->queue_for_build_scrub_map(this, g_conf->osd_scrub_map_start_cost);
+}
+
+void PG::replica_scrub(epoch_t queued, ThreadPool::TPHandle &handle)
```
athanatos: The name of this method should probably be something more like do_scrubmap_work (it runs on the primary as well, right?)

src/osd/PG.cc
```diff
@@ -3847,6 +3850,30 @@ void PG::replica_scrub(
...
+{
+  dout(20) << __func__ << " state " << Scrubber::map_state_string(scrubber.map_state)
+           << " objects " << scrubber.ls.size() << dendl;
+  assert(_lock.is_locked());
+  assert(scrubber.active_rep_scrub);
```
athanatos: As I mentioned earlier, if you want this to be set at this point, you'll probably need to adjust the existing users.

src/osd/PG.cc
```diff
@@ -3855,9 +3882,25 @@ void PG::replica_scrub(
...
-  build_scrub_map_chunk(
-    map, start, end, msg->deep, msg->seed,
-    handle);
+  if (scrubber.map_state == PG::Scrubber::REG_SCRUBBING) {
+    assert(scrubber.scrubmap.objects.empty());
+    assert(scrubber.ls.empty());
+    build_scrub_map_chunk(scrubber.scrubmap, start, end, msg->seed, handle, scrubber.ls);
+  } else {
+    assert(scrubber.map_state == PG::Scrubber::DEEP_SCRUBBING);
+    get_pgbackend()->be_deep_scan_list(scrubber.scrubmap, scrubber.ls, msg->seed, handle);
```
athanatos: It seems to me like you still want to use build_scrub_map_chunk.  Even with a deep scrub, we do want to verify the metadata stuff from build_scrub_map_chunk.

src/osd/PG.cc
```diff
@@ -3855,9 +3882,25 @@ void PG::replica_scrub(
...
+  }
+
+  if (msg->deep && !scrubber.ls.empty()) {
+    scrubber.map_state = PG::Scrubber::DEEP_SCRUBBING;
+    unsigned howmany = scrubber.ls.size() > (unsigned)g_conf->osd_deep_map_chunk_max ? (unsigned)g_conf->osd_deep_map_chunk_max : scrubber.ls.size();
```
athanatos: 80 characters

src/osd/ReplicatedPG.cc
```diff
@@ -9760,9 +9750,14 @@ void ReplicatedPG::_applied_recovered_object(ObjectContextRef obc)
   --active_pushes;
 
   // requeue an active chunky scrub waiting on recovery ops
-  if (!deleting && active_pushes == 0
+  if (!deleting && active_pushes == 0 &&
```
athanatos: Ah, I guess you did update these users.  Still, active_pushes could continue to be 0, right?  If do_scrub_map resets active_rep_scrub again before requeueing the other event, wouldn't this refire?

src/osd/ReplicatedPG.cc
```diff
@@ -12756,11 +12751,11 @@ unsigned ReplicatedPG::process_clones_to(const boost::optional<hobject_t> &head,
  *              [Snapset clones 4]
  * EOL                  obj4 snap 4, (expected)
  */
-void ReplicatedPG::_scrub(
+void ReplicatedPG::scrub_snapshot_metadata(
```
athanatos: Much better name.

src/osd/PG.cc
dzafman: @athanatos It is a little confusing, but a deep scrub goes through the REG_SCRUBBING phase which gets the metadata AND the list of objects in scrubber.ls.  That list is then used after a requeue in DEEP_SCRUBBING phase to call be_deep_scan_list().

src/msg/Message.cc
dzafman: @athanatos I removed the rename commits.

dzafman:
Replaced by #11255

PR #11086, opened 2016-09-14
erasure-code: Runtime detection of SIMD for jerasure and shec

bassamtabbara:
The PR completes the first part of http://pad.ceph.com/p/erasure-coding-runtime-simd-detection.

By switching to a new gf-complete (v3) with SIMD runtime detection, we can now remove all the different flavors of jerasure and shec. This simplifies deployment and configuration of erasure coding, enables hetergenous OSDs, and enables us to take advantage of new performance improvements in jerasure without config/build changes.

This PR removes flavors from cmake, removes ErasureCodePluginSelect___, and fixes unit tests. There is now a single plugin for jerasure and a single plugin for shec.

SIMDExt.cmake was changed so that its a little more generic, and is not polluted with gf-complete specific CFLAG defines.

I also added a small init helper for jerasure that has code that was common between jerasure and shec.


dachary:
there are unrelated changes at src/erasure-code/isa/ErasureCodePluginIsa.h for instance, maybe it just needs a rebase against master ?

dachary:
You also need to consider upgrades. Removing the ability to select the plugin flavor makes this change non backward compatible. It is entirely possible that a sysadmin has a setup where the plugin is set using this feature. There are two ways to deal with this, I think:

* have a new jerasure2 plugin and add a feature bit to make it part of the standard set of plugins and make it the new default plugin
* make this change fully backward compatible by implementing the selection of the plugin flavor (which would always select the same plugin), remove the plugin selection from the documentation and maybe display a deprecation warning when it is used

What do you think ?

bassamtabbara:
@dachary regarding the changes to ErasureCodePluginIsa.h/.cc they are somewhat unrelated. I wanted all EC plugins to be invokable directly for testing and possibly other uses, and made the change to all EC plugins for consistency. I can back this out if you'd like but it seems harmless.

regarding the back compat scenarios, while creating a jerasure2 would be easier I'd like to attempt the bigger cleanup in Kraken. I'm going to need help in making this change. Here what I'm thinking:

* Modify ErasureCodePluginFactory such that a call to factory("jerasure_flavor") are all mapped to just "jerasure" (and the same for shec). This is safe to do for all versions and will not need to check feature bits.
* Find a way to warn users if a flavor is specified in a config file.
* Modify OSDMonitor::normalize_profile such that it generates a warning when a ec flavor is seen in the osdmap, and if the CEPH_OSDMAP_REQUIRE_KRAKEN is set will change the plugin name to just jerasure in the OSDMap.

Does this seem like it would work?

dachary:
> regarding the changes to ErasureCodePluginIsa.h/.cc they are somewhat unrelated.

If you could just separate them into another commit it would be enough.

> Modify ErasureCodePluginFactory such that a call to factory("jerasure_flavor") 

If you mean **ErasureCodePlugin::factory** it may not be the best idea because it is common to all plugins. 

> Modify OSDMonitor::normalize_profile such that it generates a warning when a ec flavor is seen in the osdmap

Yes.

> and if the CEPH_OSDMAP_REQUIRE_KRAKEN is set will change the plugin name to just jerasure in the OSDMap.

I think we should avoid modifying the plugin name. The backward compatibility will stay until the plugin is deprecated. It's a not a significant problem to keep that around. What's annoying is doing it, not so much maintain it, as long as the default for new clusters is to not use that. And you would not need a feature bit such as CEPH_OSDMAP_REQUIRE_KRAKEN.

Please let me know if I misunderstood what you suggest. 

Thanks a lot for this work :-)

bassamtabbara:
Ok will separate the header refactoring into a separate commit.

>If you mean ErasureCodePlugin::factory it may not be the best idea because it is common to all plugins.

Whats a better place for this logic?

> I think we should avoid modifying the plugin name.

I think you're saying that I should just blindly map jerasure_ss4 --> jerasure and warn the user when we see a flavor. No attempt should be done to rewrite the plugin name anywhere. Is this right?

dachary:
> I think you're saying that I should just blindly map jerasure_ss4 --> jerasure and warn the user when we see a flavor. No attempt should be done to rewrite the plugin name anywhere. Is this right?

Yes. It could be as simple as having the legacy names as symlinks so there would be no requirement to change any code, just to display a warning.

bassamtabbara:
Good point about creating symlinks. I'll pursue that.

> just to display a warning.

Whats a good place to show warnings? I assume we want to show them once on startup.

dachary:
The mon logs ?

bassamtabbara:
@dachary take  a look now and let me know if this works. 

I added support for logging warnings when legacy EC plugins are used. Note that I did not go with a symlink approach since it seem quite complicated to get cmake to recognize the link as a target, and also realized that it might break distros. Instead I compile the same code with different names.

src/global/global_init.cc
```diff
@@ -454,3 +456,49 @@ int global_init_shutdown_stderr(CephContext *cct)
...
+	}
+	else if (plugin_name == "shec_generic" ||
+		 plugin_name == "shec_sse3" ||
+		 plugin_name == "shec_sse4" ||
+		 plugin_name == "shec_neon") {
+	  replacement_plugin = "shec";
+	}
+
+	if (replacement_plugin != "") {
+	  derr << "WARNING: osd_erasure_code_plugins contains plugin "
```
dachary: It would probably be better to have dout(0) instead of derr otherwise the user may be confused because it shows as if it was an error but the content of the message suggests it's only a warning.

src/mon/OSDMonitor.cc
```diff
@@ -4404,20 +4404,47 @@ int OSDMonitor::crush_rename_bucket(const string& srcname,
 
   pending_inc.crush.clear();
   newcrush.encode(pending_inc.crush);
-  *ss << "renamed bucket " << srcname << " into " << dstname;
+  *ss << "renamed bucket " << srcname << " into " << dstname;	
   return 0;
 }
 
-int OSDMonitor::normalize_profile(ErasureCodeProfile &profile, ostream *ss)
+void OSDMonitor::check_legacy_ec_plugin(const string& plugin, const string& profile)
```
dachary: const method

src/mon/OSDMonitor.cc
```diff
@@ -4460,7 +4487,7 @@ int OSDMonitor::crush_ruleset_create_erasure(const string &name,
 
 int OSDMonitor::get_erasure_code(const string &erasure_code_profile,
 				 ErasureCodeInterfaceRef *erasure_code,
-				 ostream *ss) const
```
dachary: you can keep the const if the const above is set

dachary:
@bassamtabbara this looks great. I only looked at the last commit 979c5c5, assuming you will split and rework the previous commits for clarity. If they are final I'll review them as well but they also look good overall :-)

src/mon/OSDMonitor.cc
bassamtabbara: i couldn't make it a const method since its calling out and that uses osmap which the compiler was saying needs to be made const.

src/global/global_init.cc
bassamtabbara: ok.

bassamtabbara:
@dachary I think I have all your feedback incorporated. This PR is ready for final review. 

Note also that I did split and rework previous commits for clarity.

src/erasure-code/jerasure/jerasure_init.cc
```diff
@@ -3,7 +3,7 @@
 /*
  * Ceph distributed storage system
  *
- * Copyright (C) 2014 Cloudwatt <libre.licensing@cloudwatt.com>
```
dachary: factoring out jerasure_init.cc and using them for both jerasure / shec should be in another commit. And this they are slightly different, a detailed explanation about why they are factored out to be the same should be provided.

cmake/modules/SIMDExt.cmake
```diff
@@ -1,66 +1,60 @@
 # detect SIMD extentions
 #
```
dachary: the title above should start with erasure-code:

cmake/modules/SIMDExt.cmake
```diff
@@ -1,66 +1,60 @@
...
-# ARM_NEON_FLAGS
-#
 # HAVE_ARMV8_CRC
```
dachary: the renaming of the compilation flags shoudl carefully be done in an independent commit with detailed explanations about the precautions taken to avoid missing a #ifdef HAVE_NEON or #ifdef INTEL_SSE4_1. This is specialy important because it has an impact outside of erasure code.

To be extra conservative and avoid this burden, the new HAVE_* flags could be added instead of replacing the existing flags to reduce the scope of the code change.

dachary:
I made my comments on all but the cmake changes. Before I dive into them, did you verify that rpm/deb packages build as expected ?

src/erasure-code/jerasure/jerasure_init.cc
bassamtabbara: Looking more closely they do seem different. I'll cleanup. thanks.

cmake/modules/SIMDExt.cmake
bassamtabbara: the flags are used primarily in gf-complete and zlib. they were based on #define that originated in gf-complete and are non-standard. I'll attempt to split the commit as you suggest and add some verbiage.

dachary: thanks. The commit does not need to be standalone (i.e work without the other commits) if that makes things easier.

cmake/modules/SIMDExt.cmake
bassamtabbara: done

src/erasure-code/jerasure/jerasure_init.cc
bassamtabbara: I changed jerasure_init to take an array of words to use and its invoked from jerasure and shec.

cmake/modules/SIMDExt.cmake
bassamtabbara: thanks it would be a big change to split this commit and ensure that it builds on each commit. I added verbiage to the commit message around testing. hopefully this satisfies your concern.

bassamtabbara:
Ok ready for review again. All feedback incorporated. Thanks again for keeping me honest.

src/erasure-code/jerasure/jerasure_init.cc
bassamtabbara: One more thought on this issue. If the shec and jerasure plugins are loaded at the same time, don't they share the same galois field arrays? So shec would be enabled for w4 or maybe jerasure would lose w4 depending on the order of loading?

dachary:
nit: could you please prefix the title of all commits with erasure-code: ? except the one which is already prefixed with build/ops:

bassamtabbara:
thanks. updated all commits with "erasure-code" prefix.

src/erasure-code/jerasure/jerasure_init.cc
dachary: I realize just now that the initialization of jerasure has no impact on how the plugin exposes the w parameter. Yesterday when we discussed it I argued against having common code for both jerasure and shec but I was wrong. Sorry for the noise.

to answer your question, the variables in shec and jerasure are different and the initialization done on shec won't be used by jerasure.

dachary:
could you split https://github.com/ceph/ceph/pull/11086/commits/2c9bed482c25e838f2625c0f1a93400d299e96e8 into two commits ? The first one with changes that are outside of the erasure-code directory. The second one with the changes that are in the erasure-code directory. In case a problem is discovered, it will make it easier to analyze.

src/mon/OSDMonitor.cc
```diff
@@ -4407,20 +4407,47 @@ int OSDMonitor::crush_rename_bucket(const string& srcname,
 
   pending_inc.crush.clear();
   newcrush.encode(pending_inc.crush);
-  *ss << "renamed bucket " << srcname << " into " << dstname;
+  *ss << "renamed bucket " << srcname << " into " << dstname;
```
dachary: trailing white space

src/test/erasure-code/test-erasure-code-plugins.sh
```diff
@@ -0,0 +1,109 @@
...
+function TEST_preload_no_warning() {
+    local dir=$1
+
+    for plugin in ${plugins[*]}; do
+        setup $dir || return 1
+        run_mon $dir a --osd_erasure_code_plugins="${plugin}" || return 1 
+        CEPH_ARGS='' ceph --admin-daemon $dir/ceph-mon.a.asok log flush || return 1
+        run_osd $dir 0 --osd_erasure_code_plugins="${plugin}" || return 1 
+        CEPH_ARGS='' ceph --admin-daemon $dir/ceph-osd.0.asok log flush || return 1
+        grep -v "WARNING: osd_erasure_code_plugins contains plugin" $dir/mon.a.log || return 1
```
dachary: grep -v will return true as long as at least one line does not match, you need to use ! grep instead

PendingReleaseNotes
```diff
@@ -51,4 +51,12 @@
   development packages (lib*-dev) any more. As it is unneeded, and per
   https://wiki.debian.org/ReleaseGoals/LAFileRemoval and
   https://www.debian.org/doc/manuals/maint-guide/advanced.en.html,
-  we should remove them.
```
dachary: please make that into a separate commit with doc: prefix

dachary:
contrary to what I thought the documentation (in the doc directory) never mentionned the jerasure/shec variants. It's good because there is no need to update it :-)

cmake/modules/SIMDExt.cmake
```diff
@@ -1,66 +1,60 @@
...
 
 if(CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64|AARCH64")
   CHECK_C_COMPILER_FLAG(-march=armv8-a+crc HAVE_ARMV8_CRC)
   if(HAVE_ARMV8_CRC)
-    set(ARM_CRC_FLAGS "-march=armv8-a+crc -DARCH_AARCH64")
+    set(ARM_CRC_FLAGS "-march=armv8-a+crc")
   endif()
-  CHECK_C_COMPILER_FLAG(-march=armv8-a+simd HAVE_NEON)
-  if(HAVE_NEON)
-    set(ARM_NEON_FLAGS "-march=armv8-a+simd -DARCH_AARCH64 -DARM_NEON")
```
dachary: ARM_NEON_FLAGS is only set for jerasure therefore modifying it has no impact elsewhere :+1:

cmake/modules/SIMDExt.cmake
```diff
@@ -1,66 +1,60 @@
...
-# SSE3_FLAGS
-# SSE4_FLAGS
+# SIMD_COMPILE_FLAGS
 #
 
 if(CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64|AARCH64")
   CHECK_C_COMPILER_FLAG(-march=armv8-a+crc HAVE_ARMV8_CRC)
   if(HAVE_ARMV8_CRC)
-    set(ARM_CRC_FLAGS "-march=armv8-a+crc -DARCH_AARCH64")
+    set(ARM_CRC_FLAGS "-march=armv8-a+crc")
```
dachary: this is unrelated to erasure-code, let's not modify it at all. As far as I can tell it's not used therefore it could be removed. This cleanup should however be in a separate commit.

cmake/modules/SIMDExt.cmake
```diff
@@ -1,66 +1,60 @@
...
-    set(ARM_NEON_FLAGS "-mfpu=neon -DARM_NEON")
+  CHECK_C_COMPILER_FLAG(-mfpu=neon HAVE_ARM_NEON)
+  if(HAVE_ARM_NEON)
+    set(SIMD_COMPILE_FLAGS "${SIMD_COMPILE_FLAGS} -mfpu=neon")
   endif()
 elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "i386|i686|amd64|x86_64|AMD64")
-  set(SSE3_FLAGS)
-  CHECK_C_COMPILER_FLAG(-msse HAVE_SSE)
-  if(HAVE_SSE)
-    set(SSE3_FLAGS "${SSE3_FLAGS} -msse -DINTEL_SSE")
```
dachary: Also only used in jerasure :+1:

cmake/modules/SIMDExt.cmake
```diff
@@ -1,66 +1,60 @@
...
+    set(SIMD_COMPILE_FLAGS "${SIMD_COMPILE_FLAGS} -mssse3")
   endif()
-
-  # pclmul is not enabled in SSE3, otherwise we need another
-  # flavor or an cmake option for enabling it
-
-  set(SSE4_FLAGS ${SSE3_FLAGS})
-  CHECK_C_COMPILER_FLAG(-msse4.1 INTEL_SSE4_1)
-  if(SUPPORTS_SSE41)
-    set(SSE4_FLAGS "${SSE4_FLAGS} -msse41 -DINTEL_SSE4")
```
dachary: Also only used in jerasure :+1:

dachary:
The https://github.com/ceph/ceph-erasure-code-corpus/blob/master/v0.92-988/non-regression.sh will work because you made it backward compatible. But it won't have the desired effect which is to verify that encoding/decoding works exactly the same with or without SIMD. To achieve the same thing the plugin selection must be replaced with something (environment variable ?) instructing jerasure to not use SIMD even if they are available.

If we're no longer doing that, we verify less than what we used to.

src/mon/OSDMonitor.cc
bassamtabbara: odd I didnt make this change. I'll rebase.

src/test/erasure-code/test-erasure-code-plugins.sh
bassamtabbara: done.

PendingReleaseNotes
bassamtabbara: reworded to remove the statement about removing them in future versions.

cmake/modules/SIMDExt.cmake
bassamtabbara: yes ARCH_AARCH64 came from gf-complete and I think it was written this way when the big cmake refactoring was done. Are you saying that I should split just this line into a separate commit? Personally, I find it more readable in one shot.

bassamtabbara:
this latest round of commits factors in most your changes and rebases from master. testing debian packages now and will look into beefing up the regression test

bassamtabbara:
WRT to the non-regression tests I made changes here https://github.com/ceph/ceph-erasure-code-corpus/pull/4. Once those are reviewed and accepted I can update the submodule commit hash in this review.

bassamtabbara:
Ok I picked up the new corpus and fixed regression testing. Also incorporated the new batch of feedback. I was able to build the debian packages (except for signing). Let me know if there is something I missed.

src/erasure-code/jerasure/CMakeLists.txt
```diff
@@ -35,6 +35,13 @@ if(HAVE_INTEL_SSE4_2)
   list(APPEND GF_COMPILE_FLAGS INTEL_SSE4)
 endif()
 
+# set this to TRUE to enabel debugging of SIMD detection
```
dachary: enable

src/test/erasure-code/ceph_erasure_code_non_regression.cc
```diff
@@ -70,8 +68,6 @@ int ErasureCodeNonRegression::setup(int argc, char** argv) {
      "prefix all paths with base")
     ("parameter,P", po::value<vector<string> >(),
      "add a parameter to the erasure code profile")
-    ("path", po::value<string>(), "content path instead of inferring it from parameters")
```
dachary: the   if (vm.count("path")) below and all references to that option should also be removed

src/erasure-code/jerasure/CMakeLists.txt
bassamtabbara: fixed

dachary:
please split https://github.com/ceph/ceph/pull/11086/commits/c10df4d5cd882bcc37dd9c947c6c02519c24823d in two commits (one with erasure-code changes and the other with changes outside of erasure-code)

dachary:
while you're working on the latest (cosmetic) changes, I'll run integration tests to validate it works

bassamtabbara:
Ok split the commit into two parts. All changes are in. Let me know if you see anything with integration testing. Also I would love to be able to run that myself if you can give me access.

dachary:
https://github.com/ceph/ceph/pull/11086/commits/c2096424453413f02f0d8e1eadc7133910299ec4 still has a reference to zlib in the comment, otherwise looks good

dachary:
building in wip-pr-11086

bassamtabbara:
Ok I amended the commit to remove the zlib comment.

dachary:
could you please rebase and repush ? the unrelated make check failures should be fixed now and we can hope for a clean run. thanks !

bassamtabbara:
done. How did the integration testing go? Can you post a link here?

dachary:
@bassamtabbara they did not run yet. They would have failed for the same reason make check failed.

dachary:
building in wip-pr-11086

dachary:
<code>teuthology-suite --verbose --suite-branch master --ceph wip-pr-11086 --subset 2/2000 --suite rados --machine-type smithi --email loic@dachary.org</code>

* **running** http://pulpito.ceph.com/loic-2016-09-26_13:00:32-rados-wip-pr-11086---basic-smithi

dachary:
the tests are making progress today, should complete by friday. So far so good ;-)

bassamtabbara:
@dachary Looks like the integration tested is complete. Not sure I understand why there are 4 failures and 4 dead jobs. They don't seem related to the work here from first look.

dachary:
A breakdown of the failures is as follows:

* *no message*
 * [rados/singleton/{rados.yaml all/dump-stuck.yaml fs/xfs.yaml msgr/random.yaml msgr-failures/many.yaml}](http://pulpito.ceph.com/loic-2016-09-26_13:00:32-rados-wip-pr-11086---basic-smithi/438329)
* *saw valgrind issues*
 * [rados/verify/{rados.yaml 1thrash/none.yaml clusters/{fixed-2.yaml openstack.yaml} fs/btrfs.yaml mon_kv_backend/rocksdb.yaml msgr/simple.yaml msgr-failures/few.yaml tasks/rados_cls_all.yaml validater/valgrind.yaml}](http://pulpito.ceph.com/loic-2016-09-26_13:00:32-rados-wip-pr-11086---basic-smithi/438346)
 * [rados/verify/{rados.yaml 1thrash/none.yaml clusters/{fixed-2.yaml openstack.yaml} fs/btrfs.yaml mon_kv_backend/rocksdb.yaml msgr/async.yaml msgr-failures/few.yaml tasks/mon_recovery.yaml validater/valgrind.yaml}](http://pulpito.ceph.com/loic-2016-09-26_13:00:32-rados-wip-pr-11086---basic-smithi/438286)
* *'mkdir -p -- /home/ubuntu/cephtest/mnt.0/client.0/tmp && cd -- /home/ubuntu/cephtest/mnt.0/client.0/tmp && CEPH_CLI_TEST_DUP_COMMAND=1 CEPH_REF=e23a464a5a6c73c8eb7c16a7b0c33dcedb3d083e TESTDIR="/home/ubuntu/cephtest" CEPH_ARGS="--cluster ceph" CEPH_ID="0" PATH=$PATH:/usr/sbin adjust-ulimits ceph-coverage /home/ubuntu/cephtest/archive/coverage timeout 3h /home/ubuntu/cephtest/workunit.client.0/rados/test_envlibrados_for_rocksdb.sh'*
 * [rados/singleton/{rados.yaml all/test_envlibrados_for_rocksdb.yaml fs/xfs.yaml msgr/async.yaml msgr-failures/many.yaml}](http://pulpito.ceph.com/loic-2016-09-26_13:00:32-rados-wip-pr-11086---basic-smithi/438304)

dachary:
One of the dead jobs is http://pulpito.ceph.com/loic-2016-09-26_13:00:32-rados-wip-pr-11086---basic-smithi/438259/ and the log suggests it was killed before it could complete. The description of the job is
<pre>
  description: rados/upgrade/{rados.yaml hammer-x-singleton/{0-cluster/{openstack.yaml
    start.yaml} 1-hammer-install/hammer.yaml 2-partial-upgrade/firsthalf.yaml 3-thrash/default.yaml
    4-mon/mona.yaml 5-workload/{rbd-cls.yaml rbd-import-export.yaml readwrite.yaml snaps-few-objects.yaml}
    6-next-mon/monb.yaml 7-workload/{radosbench.yaml rbd_api.yaml} 8-next-mon/monc.yaml
    9-workload/{ec-rados-plugin=jerasure-k=3-m=1.yaml rbd-python.yaml rgw-swift.yaml
    snaps-many-objects.yaml test_cache-pool-snaps.yaml}}}
</pre>
and means part if the job is to run rados tests on an erasure coded pool using the [jerasure plugin with k=3,m=1](https://github.com/ceph/ceph-qa-suite/blob/master/suites/rados/upgrade/hammer-x-singleton/9-workload/ec-rados-plugin%3Djerasure-k%3D3-m%3D1.yaml) which is relevant to this pull request. I'll re-run the failed tests while trying to figure out what caused the error to assert if it is transient or not.

dachary:
Re-running the failed/dead tests to figure out what is transient and what is not.
<pre>
run=loic-2016-09-26_13:00:32-rados-wip-pr-11086---basic-smithi
paddles=paddles.front.sepia.ceph.com
eval filter=$(curl --silent http://$paddles/runs/$run/ | jq '.jobs[] | select(.status == "dead" or .status == "fail") | .description' | while read description ; do echo -n $description, ; done | sed -e 's/,$//')
teuthology-suite --verbose --priority=50 --suite-branch master --ceph wip-pr-11086 --filter="$filter" --suite rados --machine-type smithi --email loic@dachary.org
</pre>

* **running** http://pulpito.ceph.com/loic-2016-09-29_07:53:35-rados-wip-pr-11086---basic-smithi/

dachary:
Running upgrade tests involving erasure coded plugins
<pre>
teuthology-suite --verbose --suite-branch master --ceph wip-pr-11086  --suite upgrade/jewel-x/stress-split-erasure-code  --machine-type vps --email loic@dachary.org
</pre>

* **running** http://pulpito.ceph.com/loic-2016-09-29_08:03:24-upgrade:jewel-x:stress-split-erasure-code-wip-pr-11086---basic-vps/

<pre>
teuthology-suite --verbose --suite-branch master --ceph wip-pr-11086  --suite upgrade/jewel-x/stress-split-erasure-code-x86_64  --machine-type vps --email loic@dachary.org 
</pre>

* **running** http://pulpito.ceph.com/loic-2016-09-29_08:02:35-upgrade:jewel-x:stress-split-erasure-code-x86_64-wip-pr-11086---basic-vps/

dachary:
Filed http://tracker.ceph.com/issues/17430 about another dead job http://pulpito.ceph.com/loic-2016-09-26_13:00:32-rados-wip-pr-11086---basic-smithi/ which is unrelated to this pull request as well.

dachary:
The dead job http://pulpito.ceph.com/loic-2016-09-26_13:00:32-rados-wip-pr-11086---basic-smithi/438320/ waits for osd.2 and osd.4 which never come back. The probably died with a core or something but we don't have them: the logs / core are not collected when the job timesout.

dachary:
http://pulpito.ceph.com/loic-2016-09-26_13:00:32-rados-wip-pr-11086---basic-smithi/438338/ log is weird: the last line is only ~30 minutes after it starts and nothing after that. The test infrastructure had a problem of some kind and the job go hanged or the log collection failed. Since the teuthology machine was rebooted two days ago maybe it's related. In any case this is unrelated to this change.

First line:
<pre>
2016-09-28T06:49:31.950 INFO:root:teuthology version: 1.0.0-5146ad1
</pre>
Last line:
<pre>
2016-09-28T07:21:06.638 INFO:teuthology.orchestra.run.smithi073.stderr:nodeep-scrub is unset
</pre>

dachary:
@bassamtabbara could you please confirm that you ran make check on the latest version of this pull request on a cloudlab.us ARMv8 bare metal ?

dachary:
For the record I believe none of the failures / dead jobs of http://pulpito.ceph.com/loic-2016-09-26_13:00:32-rados-wip-pr-11086---basic-smithi are related to this pull request.

dachary:
Running upgrade tests involving erasure coded plugins
<pre>
teuthology-suite --verbose --suite-branch master --ceph wip-pr-11086  --suite upgrade/jewel-x/stress-split-erasure-code  --machine-type vps --email loic@dachary.org
</pre>

* **running** http://pulpito.ceph.com/loic-2016-09-29_08:03:24-upgrade:jewel-x:stress-split-erasure-code-wip-pr-11086---basic-vps/

<pre>
teuthology-suite --verbose --suite-branch master --ceph wip-pr-11086  --suite upgrade/jewel-x/stress-split-erasure-code-x86_64  --machine-type vps --email loic@dachary.org 
</pre>

* **running** http://pulpito.ceph.com/loic-2016-09-29_08:02:35-upgrade:jewel-x:stress-split-erasure-code-x86_64-wip-pr-11086---basic-vps/

and waiting for a confirmation that make check runs on ARMv8 bare metal.

dachary:
@liewegas in my opinion this pull request can be merged (waiting for confirmation on two upgrade tests). Would you mind taking a quick look ?

src/mon/OSDMonitor.cc
```diff
@@ -4418,20 +4418,47 @@ int OSDMonitor::crush_rename_bucket(const string& srcname,
...
+             plugin == "shec_neon") {
+    replacement = "shec";
+  }
+
+  if (replacement != "") {
+    dout(0) << "WARNING: erasure coding profile " << profile << " uses plugin "
+	    << plugin << " that has been deprecated. Please use " 
+	    << replacement << " instead." << dendl;
+  }
+}
```
liewegas: I would change this method to return a bool (or similarly refactor), and add a health check in void OSDMonitor::get_pools_health() so that a HEALTH_WARN comes up if a legacy ec plugin name is in use.  

(A little later we can make the warning only pop up after the upgrade completes, but we haven't allocated a feature bit yet for kraken.)

liewegas: (We can follow up on the warning later, as long as we don't forget.)

liewegas: I take it back.. we should have the mon automatically clean it up after the upgrade is complete.  But we can do that later.  This looks good to go!

bassamtabbara:
Thanks! Rebased off of master. Running make check on bare-metal ARMv8 (via cloud lab) now.

bassamtabbara:
```make check``` completed on bare-metal ARM. one minor issue was found in test-erasure-code-plugins.sh (a new test i added) where it assumed the ISA ec plugin was available on ARM and its not. Fixed the test in commit 2210426 for your review @dachary . I can squash it after you have a look.

dachary:
jenkins test this please

dachary:
2210426 looks good, no need to squash it

dachary:
All upgrade tests above fail with ```ceph-objectstore-tool apply-layout-settings failed with 1```. We should wait until the backport fixing this unrelated problem is merged and try again, some time next week.

dachary:
jenkins test this please

dachary:
Running upgrade tests involving erasure coded plugins
<pre>
teuthology-suite --verbose --suite-branch master --ceph wip-pr-11086  --suite upgrade/jewel-x/stress-split-erasure-code  --machine-type vps --email loic@dachary.org
</pre>

* **failed on wip-pr-11086** http://pulpito.ceph.com/loic-2016-10-03_08:34:37-upgrade:jewel-x:stress-split-erasure-code-wip-pr-11086---basic-vps
* **running on master** http://pulpito.ceph.com/loic-2016-10-03_08:34:26-upgrade:jewel-x:stress-split-erasure-code-master---basic-vps
<pre>
teuthology-suite --verbose --suite-branch master --ceph wip-pr-11086  --suite upgrade/jewel-x/stress-split-erasure-code-x86_64  --machine-type vps --email loic@dachary.org 
</pre>

* **running on wip-pr-11086** http://pulpito.ceph.com/loic-2016-10-03_08:34:55-upgrade:jewel-x:stress-split-erasure-code-x86_64-wip-pr-11086---basic-vps
* **running on master** http://pulpito.ceph.com/loic-2016-10-03_08:34:17-upgrade:jewel-x:stress-split-erasure-code-x86_64-master---basic-vps

and waiting for a confirmation that make check runs on ARMv8 bare metal.

PR #11087, opened 2016-09-14
os/bluestore/BlueFS: use StupidAllocator; fix async compaction bug

liewegas:
StupidAllocator is probably a more appropriate choice for bluefs.
Either way, it needn't be tied to what BlueStore is using.


varadakari:
Looks good to me.

PR #11093, opened 2016-09-15
Small interface cleanups for struct ceph_statx 

jtlayton:
When I originally copied the proposed statx interface for ceph, I think I took it a bit too far and left a lot of the fields as fixed-length when that wasn't really necessary. So, as I mentioned in standup yesterday, this is a set of changes to struct ceph_statx to give the interface a bit more of a userland flavor. The main changes are to keep a dev_t in the structure for the stx_dev and stx_rdev fields, and to convert all of the timestamps to use struct timespec instead of separate fields for the secs and nsecs.

There's also a small bugfix for ll_create, which had a broken caller in the (unused) SyntheticClient code.

### DIFF
--- a/src/client/Client.cc
+++ b/src/client/Client.cc
@@ -6852,6 +6852,7 @@ void Client::fill_statx(Inode *in, unsigned int mask, struct ceph_statx *stx)
   /* These are always considered to be available */
   stx->stx_dev = in->snapid;
   stx->stx_blksize = MAX(in->layout.stripe_unit, 4096);
+  stx->stx_mode = S_IFMT & in->mode;
 
   if (use_faked_inos())
    stx->stx_ino = in->faked_ino;
--- a/src/client/Client.cc
+++ b/src/client/Client.cc
@@ -6517,7 +6517,7 @@ int Client::_do_setattr(Inode *in, struct ceph_statx *stx, int mask, int uid, in
       in->ctime = ceph_clock_now(cct);
       in->cap_dirtier_uid = uid;
       in->cap_dirtier_gid = gid;
-      in->btime = utime_t(stx->stx_btime, stx->stx_btime_ns);
+      in->btime = utime_t(stx->stx_btime);
       mark_caps_dirty(in, CEPH_CAP_AUTH_EXCL);
       mask &= ~CEPH_SETATTR_BTIME;
       ldout(cct,10) << "changing btime to " << in->btime << dendl;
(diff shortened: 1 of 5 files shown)


jtlayton:
Added another patch to remove some currently-unused fields. We can readd those later if we have need for them.

src/test/libcephfs/test.cc
```diff
@@ -1488,26 +1488,26 @@ TEST(LibCephFS, Btime) {
 
   ASSERT_EQ(ceph_fstatx(cmount, fd, &stx, CEPH_STATX_CTIME|CEPH_STATX_BTIME, 0), 0);
   ASSERT_TRUE(stx.stx_mask & (CEPH_STATX_CTIME|CEPH_STATX_BTIME));
-  ASSERT_EQ(stx.stx_btime, stx.stx_ctime);
-  ASSERT_EQ(stx.stx_btime_ns, stx.stx_ctime_ns);
+  ASSERT_TRUE(stx.stx_ctime.tv_sec == stx.stx_btime.tv_sec &&
+	      stx.stx_ctime.tv_nsec == stx.stx_btime.tv_nsec);
   ceph_close(cmount, fd);
```
gregsfortytwo: Do timespecs not have have equality operator you can use instead of calling out the pieces?

src/client/Client.cc
```diff
@@ -11605,6 +11605,9 @@ int Client::ll_create(Inode *parent, const char *name, mode_t mode,
 		      int flags, struct stat *attr, Inode **outp, Fh **fhp,
 		      int uid, int gid)
 {
+  if (fhp)
+    *fhp = NULL;
```
gregsfortytwo: If it's necessary we can do this unconditionally and clean up comments about "may be" NULL?

src/test/libcephfs/test.cc
jtlayton: You would think there would be, but no -- there isn't a standard comparison helper function you can call here. That's a long standing gripe with timespecs, fwiw...

src/client/Client.cc
jtlayton: Yeah, now that you mention it, I don't see any callers that pass a NULL pointer in here. I'll change it to initialize it unconditionally.

jtlayton:
Hopefully this is at least a little neater. Removing the NULL fhp checks also allows us to remove a bit of cruftiness at the bottom of ll_create.

jtlayton:
Also, one more patch here. The proposed upstream interface always sets the file type bits in the stx_mode. Let's do the same, which should allow ganesha (and other callers) to skip requesting CEPH_STATX_MODE (and hence the As cap) if all they want is the file type.

src/client/Client.cc
```diff
@@ -6852,6 +6852,7 @@ void Client::fill_statx(Inode *in, unsigned int mask, struct ceph_statx *stx)
   /* These are always considered to be available */
   stx->stx_dev = in->snapid;
   stx->stx_blksize = MAX(in->layout.stripe_unit, 4096);
+  stx->stx_mode = S_IFMT & in->mode;
```
gregsfortytwo: Is this supposed to be a logical OR? I don't think we actually want to restrict users to only accessing file type data but I'm not very well-versed in these interfaces.

jtlayton: No, we want logical AND. David H.'s original statx set says:

 (2) st_mode.
     The part of this that identifies the file type will always be
     available, irrespective of the setting of STATX_MODE.  The access
     flags and sticky bit are as for class (1).


In our case, the caller needs to check for (or request) CEPH_STATX_MODE if he wants to use the permission bits. If he doesn't have that, then he can still access the type bits, but nothing else in the mode.

In principle, we could just set the full mode unconditionally there and only set CEPH_STATX_MODE flag in stx_mask if we have As caps, but I think it's a little cleaner to avoid setting any of the permission bits if we don't hold As.

gregsfortytwo: Oh, I misunderstood the context here; looks good.

gregsfortytwo:
Reviewed-by:

PR #11094, opened 2016-09-15
os/bluestore: per-pool compression settings

ifed01:
This implements per-pool compression/csum settings to be propagated to object storage(s) as well as corresponding handling at BlueStore.


src/compressor/Compressor.h
```diff
@@ -15,28 +15,59 @@
...
+    COMP_FORCE                  ///< compress always
+  };
+
+  static const char * get_comp_alg_name(int a);
+  static boost::optional<CompressionAlgorithm> get_comp_alg_type(const std::string &s);
+
+  static const char *get_comp_mode_name(int m);
+  static boost::optional<CompressionMode> get_comp_mode_type(const std::string &s);
+
+  Compressor(const std::string& t) : type(t) {
```
liewegas: variant that takes

src/os/bluestore/BlueStore.cc
```diff
@@ -4864,9 +4857,9 @@ int BlueStore::_decompress(bufferlist& source, bufferlist* result)
   bufferlist::iterator i = source.begin();
   bluestore_compression_header_t chdr;
   ::decode(chdr, i);
-  string name = bluestore_blob_t::get_comp_alg_name(chdr.type);
+  const string& name = Compressor::get_comp_alg_name(chdr.type);
```
liewegas: ...otherwise we are going int -> string -> int here?

src/mon/OSDMonitor.cc
```diff
@@ -5328,6 +5339,31 @@ int OSDMonitor::prepare_command_pool_set(map<string,cmd_vartype> &cmdmap,
...
+    else if (var == "compression_required_ratio") {
+      if (interr.length()) {
+        ss << "error parsing integer value '" << val << "': " << interr;
+        return -EINVAL;
+      }
+      if (n < 0 || n>100) {
+        ss << "compression_required_ratio is out of range (0-100): '" << val << "'";
+	return EINVAL;
+      }
+    }
```
liewegas: hmm, if it's 'ratio', i'd go for 0..1 instead of 0..100.

src/os/ObjectStore.h
```diff
@@ -411,6 +411,7 @@ class ObjectStore {
       OP_COLL_HINT = 40, // cid, type, bl
 
       OP_TRY_RENAME = 41,   // oldcid, oldoid, newoid
+      OP_SET_COLLOPTS = 42, // cid, pool_opts_t
```
liewegas: I think we can get away with not making this persistent, actually.. and just make sure the OSD calls it on each PG from load_pgs() and when a pool changes.  That is probably less work for the store implementation

src/os/bluestore/BlueStore.h
```diff
@@ -1450,6 +1450,15 @@ class BlueStore : public ObjectStore,
     b->shared_blob->bc.write(txc->seq, offset, bl, flags);
     txc->shared_blobs_written.insert(b->shared_blob);
   }
+
+  template <typename T, typename F>
+  T select_option(CollectionRef c, const std::string& opt_name, T val1, F f) {
```
liewegas: are the first two arguments needed?

src/os/bluestore/BlueStore.cc
```diff
@@ -7246,8 +7247,25 @@ int BlueStore::_do_alloc_write(
...
+	    if (!cp || cp->get_type_name() != val) {
+	      cp = Compressor::create(cct, val);
+            }
+	    return boost::optional<CompressorRef>(cp);
+	  }
+	  return boost::optional<CompressorRef>();
+	}
+      );
+    }
+    if(	c != nullptr) {
```
liewegas: whitesapce?

src/os/bluestore/BlueStore.cc
```diff
@@ -7246,8 +7247,25 @@ int BlueStore::_do_alloc_write(
...
+	  if (coll->pool_opts.get(pool_opts_t::COMPRESSION_ALGORITHM, &val)) {
+	    CompressorRef cp = compressor;
+	    if (!cp || cp->get_type_name() != val) {
+	      cp = Compressor::create(cct, val);
+            }
+	    return boost::optional<CompressorRef>(cp);
+	  }
+	  return boost::optional<CompressorRef>();
+	}
+      );
```
liewegas: I think an easier helper would be a Collection method that's something like

</pre>
template<T>
T get_option(option, T default_value) {
   ...
}
</pre>
so that all these callers are simply 
<pre>
c = coll->get_option(pool_opts_t::COMPRESSION_ALGORITHM, compressor);
</pre>
or maybe a pool_opts_t method like
<pre>
template<typename T>
T get(int optid, T default) { ... }
</pre>
(ala python's dict get).

src/osd/PG.cc
```diff
@@ -5576,7 +5578,20 @@ void PG::handle_query_state(Formatter *f)
...
+  __u32 crc = bl.crc32c(0);
+  if (crc != last_pool_opts_crc) {
+    ObjectStore::Transaction t;
+    t.set_collection_opts(coll, bl);
+    auto r = osd->store->queue_transaction(osr.get(), std::move(t), nullptr);
+    if (r==0){
+      last_pool_opts_crc = crc;
+    }
+  }
+}
```
liewegas: if we make this ephemeral it'd just be a store->set_collection_opts(coll, opts), no transaction.

liewegas: we can also avoid the bufferlist encode/decode and just pass pool_opts_t by reference...

src/osd/PG.cc
```diff
@@ -5531,6 +5532,7 @@ void PG::handle_advance_map(
   recovery_state.handle_event(evt, rctx);
   if (pool.info.last_change == osdmap_ref->get_epoch())
     on_pool_change();
+  update_store_with_options();
```
liewegas: the only other place we'd need to do this is from load_pgs(), or whatever the loaded() method is that it calls

src/os/bluestore/BlueStore.cc
ifed01: The goal of this method is to provide effective selection between pool and store settings. As per our earlier discussion I'm planning to expand it with an ability to select corresponding setting using some filtering on option names ( e.g. bluestore_disable_per_pool_settings = bluestore_compression_algorithm, csum_type). And I think this logic is rather out of collection/pool_opts_t scope. 
Moreover we can optimize the selection procedure to avoid per-collection option lookup/retrieval if per-store settings prevail. Hence we need to access these settings inside the method (via f arg) not provide their values externally.

src/os/bluestore/BlueStore.h
ifed01: opt_name is needed for the future per-pool vs. per-store setting selection. CollectionRef is redundant.

src/osd/PG.cc
ifed01: I checked if collection options update happens on OSD start and it looks like this takes place without any changes to load_pgs. The changes above seems to be enough. Haven't investigated what triggered this though. Do you still think we need load_pgs modification?

ifed01:
@liewegas updated except a couple of issues (see my comments).

src/osd/PG.cc
liewegas: Ah, it looks like advance_map will always happen (eventually) when the OSD is marked up, but I think it would be better to put this call inside on_pool_change() (so that it only happens when pool options change, not on every epoch change) *and* in the boost::statechart::result PG::RecoveryState::Initial::react(const Load& l) method (which happens explicitly on PG load).

ifed01: @liewegas fixed

src/os/ObjectStore.h
```diff
@@ -1378,26 +1378,27 @@ class ObjectStore {
...
+	::encode(cid, tbl);
+	::encode(type, tbl);
+	::encode(hint, tbl);
+      }
+      else {
+	Op* _op = _get_next_op();
+	_op->op = OP_COLL_HINT;
+	_op->cid = _get_coll_id(cid);
+	_op->hint_type = type;
+	::encode(hint, data_bl);
```
liewegas: this hunk probably doesn't need to be here?  is it just fixing whitespace or something?

src/os/bluestore/BlueStore.cc
```diff
@@ -7288,9 +7288,9 @@ int BlueStore::_do_alloc_write(
 	"compression_required_ratio",
 	g_conf->bluestore_compression_required_ratio,
 	[&]() {
-	  int val;
+	  double val;
 	  if(coll->pool_opts.get(pool_opts_t::COMPRESSION_REQUIRED_RATIO, &val)) {
-	    return boost::optional<double>(double(val) / 100);
+	    return boost::optional<double>(val);
```
liewegas: this hunk goes in previous commit?

liewegas:
Looks good!  There's a DNM patch title in there and some minor cleanup, then we can put it through qa!

ifed01:
@liewegas Fixed.

ifed01:
@liewegas - PR is ready for review.

ifed01:
@liewegas - rebased

ifed01:
@liewegas - rebased

liewegas:
Merged manually in 3b35f2d010561932a477ba7c1a3a87caf47a206e

PR #11095, opened 2016-09-15
os/bluestore/BlueFS: prevent concurrent async compaction

liewegas:
Tried to fix this in dbe23c94c0074358380a40d47a417f7999920696
but got the condition wrong (log_flushing is for
normal fsync).


somnathr:
@liewegas will try this. Shouldn't we also do an assert() if we are trying to access the iterator of an empty vector before the following location ?

bluefs_extent_t& e = log_file->fnode.extents.front();

liewegas:
added!  should be less confusing than a segv :)

somnathr:
Yes..

somnathr:
Do you want to merge it to master ?

liewegas:
yep!

PR #11097, opened 2016-09-15
doc: add "--orphan-stale-secs" to radosgw-admin(8)

ktdreyer:
The radosgw-admin(8) manual page did not include the description of the `--orphan-stale-secs` option of the `orphans find` command. The option sets the number of seconds to wait before declaring an object to be an orphan.

Fixes: http://tracker.ceph.com/issues/17280


ktdreyer:
@cbodley pointed out that we need to fix `--help` as well. That change is https://github.com/ceph/ceph/pull/11098

PR #11098, opened 2016-09-15
radosgw-admin: add "--orphan-stale-secs" to --help

ktdreyer:
The `radosgw-admin --help` did not include the description of the `--orphan-stale-secs` option of the `orphans find` command. The option sets the number of seconds to wait before declaring an object to be an orphan.

Fixes: http://tracker.ceph.com/issues/17280

### DIFF
--- a/src/rgw/rgw_admin.cc
+++ b/src/rgw/rgw_admin.cc
@@ -249,6 +249,7 @@ void _usage()
   cout << "\nOrphans search options:\n";
   cout << "   --pool                    data pool to scan for leaked rados objects in\n";
   cout << "   --num-shards              num of shards to use for keeping the temporary scan info\n";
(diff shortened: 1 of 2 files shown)


src/rgw/rgw_admin.cc
```diff
@@ -249,6 +249,7 @@ void _usage()
   cout << "\nOrphans search options:\n";
   cout << "   --pool                    data pool to scan for leaked rados objects in\n";
   cout << "   --num-shards              num of shards to use for keeping the temporary scan info\n";
+  cout << "   --orphan-stale-secs       num of seconds to wait before declaring an object to be an orphan (default: 86400)"
```
cbodley: missing a `;` at the end

ktdreyer: Thanks! amended and re-pushed.

PR #11104, opened 2016-09-15
rbd-mirror: concurrent access of event might result in heap corruption

dillaman:
Fixes: http://tracker.ceph.com/issues/17283


trociny:
lgtm

PR #11105, opened 2016-09-16
build scripts: Enable dnf for Fedora >= 22

badone:
The use of yum triggers a warning and reference to yum2dnf on Fedora >= 22.
Use either yum or dnf based on the Fedora version.

### DIFF
--- a/install-deps.sh
+++ b/install-deps.sh
@@ -85,11 +85,19 @@ else
 	if [ -n "$backports" ] ; then rm $control; fi
         ;;
     centos|fedora|rhel)
-        echo "Using yum to install dependencies"
-        $SUDO yum install -y redhat-lsb-core
+        yumdnf="yum"
+        builddepcmd="yum-builddep -y"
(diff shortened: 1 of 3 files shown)


install-deps.sh
```diff
@@ -85,11 +85,19 @@ else
 	if [ -n "$backports" ] ; then rm $control; fi
         ;;
     centos|fedora|rhel)
-        echo "Using yum to install dependencies"
-        $SUDO yum install -y redhat-lsb-core
+        yumdnf="yum"
+        builddepcmd="yum-builddep -y"
+        if test "$(echo "$VERSION_ID >= 22" | bc)" -ne 0; then
```
tchaikov: could also put
```sh
if [ $((VERSION_ID >= 22)) -eq 1 ]; then
```

badone: Try it with rhel or CentOS where VERSION_ID may be something like "7.2". Bash is not good at doing float/int comparisons

PR #10943, opened 2016-08-31
tox.ini: remove extraneous coverage --omit option

jdurgin:
This option isn't eliminating anything in the runs I've seen, and it
causes the coverage command (and thus 'make check') to fail spuriously
if a parent directory contains 'test' or 'tox', since it filters out
all of the coverage data.

Signed-off-by: Josh Durgin <jdurgin@redhat.com>

PR #11110, opened 2016-09-17
DNM: tox.ini: remove extraneous coverage --omit option

dmick:
dmick: making a PR just to invoke tests, after rebasing to current master

This option isn't eliminating anything in the runs I've seen, and it
causes the coverage command (and thus 'make check') to fail spuriously
if a parent directory contains 'test' or 'tox', since it filters out
all of the coverage data.


dmick:
Duplicates #10943

PR #11037, opened 2016-09-09
librbd: possible deadlock if cluster connection closed after image

dillaman:
Fixes: http://tracker.ceph.com/issues/17254
Signed-off-by: Jason Dillaman <dillaman@redhat.com>

PR #11112, opened 2016-09-17
librbd: corrected use-after-free in ImageWatcher

dillaman:
Fixes: http://tracker.ceph.com/issues/17289


dillaman:
@trociny fix for use-after-free introduced by #11037

PR #11113, opened 2016-09-17
os/filestore: disable use of splice by default

yuyuyu101:
1. currently splice is actually buggy whatever enabling or not fiemap/seek_data feature
2. like bluestore's clone_range(https://github.com/ceph/ceph/pull/11106/commits/a931dd3e0067de58c83dc41e5adb0578cc5fa2c1) we need to zero target file when source file exist hole, but filestore's zero op maybe heavy which not equal to bluestore's zero. So I'd like pick up full copy strategy instead of fine grain handle. Since clone_range overwrite target object case is rare, I don't think this will be a problem in current clone_range users.

### DIFF
--- a/src/os/filestore/GenericFileStoreBackend.cc
+++ b/src/os/filestore/GenericFileStoreBackend.cc
@@ -58,10 +58,11 @@ GenericFileStoreBackend::GenericFileStoreBackend(FileStore *fs):
   FileStoreBackend(fs),
   ioctl_fiemap(false),
   seek_data_hole(false),
+  use_splice(false),
   m_filestore_fiemap(g_conf->filestore_fiemap),
   m_filestore_seek_data_hole(g_conf->filestore_seek_data_hole),
   m_filestore_fsync_flushes_journal_data(g_conf->filestore_fsync_flushes_journal_data),
-  m_filestore_splice(false) {}
+  m_filestore_splice(g_conf->filestore_splice) {}
 
 int GenericFileStoreBackend::detect_features()
 {
@@ -165,6 +166,9 @@ int GenericFileStoreBackend::detect_features()
   //splice detection
 #ifdef CEPH_HAVE_SPLICE
   if (!m_filestore_splice) {
+    dout(0) << __func__ << ": splice() is disabled via 'filestore splice' config option" << dendl;
+    use_splice = false;
+  } else {
     int pipefd[2];
     loff_t off_in = 0;
     int r;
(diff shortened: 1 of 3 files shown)


src/os/filestore/GenericFileStoreBackend.cc
```diff
@@ -61,7 +61,7 @@ GenericFileStoreBackend::GenericFileStoreBackend(FileStore *fs):
   m_filestore_fiemap(g_conf->filestore_fiemap),
   m_filestore_seek_data_hole(g_conf->filestore_seek_data_hole),
   m_filestore_fsync_flushes_journal_data(g_conf->filestore_fsync_flushes_journal_data),
-  m_filestore_splice(false) {}
+  m_filestore_splice(g_conf->filestore_splice) {}
```
liewegas: does this actually change the behavior?  it's going from false to (tunable) false.

liewegas: I think the detection code in GenericFileStoreBackend.cc shoudl only run if it's true (instead of false) and disable if it's not detected (vs enable if it is).

yuyuyu101: Oh, I found the detection in _detect_feature is wrong.... I update _detect_feature too. Otherwise it will enable splice when m_filestore_splice is false....

src/os/filestore/FileStore.cc
```diff
@@ -3536,17 +3536,30 @@ int FileStore::_do_sparse_copy_range(int from, int to, uint64_t srcoff, uint64_t
...
+  int64_t written = 0;
+  struct stat st;
+  r = ::fstat(to, &st);
+  if (r < 0) {
+    r = -errno;
+    derr << __func__ << ": fstat error at " << to << " " << cpp_strerror(r) << dendl;
+    goto out;
+  }
+
+  if ((uint64_t)st.st_size > dstoff) {
```
asheplyakov: I don't think this condition is correct.

Example: 
- `from` is 8KB long, and first 4KB is a hole
- `to` is 1MB long and has no holes
- srcoff = 0, len = 4096, dstoff = 1MB - 4KB
Such operation should have punched a 4KB hole at 1MB - 4K, however nothing will happen since the (st.size > dstoff) condition does not hold.

The problem is that this method only copies the data, however it should have also punched the same holes as in the source.

yuyuyu101: agreed.....

@liewegas go ahead to complete the sparse copy feature, or totally fall down to full copy? I think both are ok. But I want to get an ack if we prefer to keep these freeze.

liewegas: I suspect in the hole case falling back to a full copy is fine.  That path shouldn't be triggered with real workloads anyway.

yuyuyu101:
@liewegas remove fiemap things, let's disable splice by default firstly.

yuyuyu101:
retest this please

yuyuyu101:
@liewegas ping?

PR #11116, opened 2016-09-17
run-cmake-check.sh: Remove redundant calls

badone:
Calling "git submodule init" is redundant since it is called in do_cmake.sh and
there's no reason to do a modprobe of rbd.ko here either.

### DIFF
--- a/run-cmake-check.sh
+++ b/run-cmake-check.sh
@@ -42,12 +42,10 @@ function run() {
     else
         echo "WARNING: Don't know how to install packages" >&2
     fi
-    sudo /sbin/modprobe rbd
 
     if test -f ./install-deps.sh ; then
 	$DRY_RUN ./install-deps.sh || return 1
     fi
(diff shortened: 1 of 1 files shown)


liewegas:
@idryomov @dachary  do you know why the rbd modprobe was added in the first place?

badone:
Looks like @dachary put it in the original run-make-check.sh (b9ddf97a0c917179e30d4d7bf5fff282c33a60d9) and @dmick removed it from that file with "modprobe won't work on containers; not needed" (bf5ccce9ad36d8dcb4506a2731d1c8e8b322c725). Looks like run-cmake-check.sh was copied from run-make-check.sh somewhere in between. This also raises the question of whether we need both run-make-check.sh and run-cmake-check.sh? I suspect the answer is "no".

idryomov:
FWIW I don't recall any such tests, and we shouldn't be doing a manual modprobe anyway.  LGTM.

PR #11119, opened 2016-09-18
Reduce code of MonClient::ms_dispatch

dongbula:
The first switch in function MonClient::ms_dispatch do not seem necessary, so I modify these code to make it be simpler.

### DIFF
--- a/src/mon/MonClient.cc
+++ b/src/mon/MonClient.cc
@@ -250,19 +250,6 @@ bool MonClient::ms_dispatch(Message *m)
   if (my_addr == entity_addr_t())
     my_addr = messenger->get_myaddr();
 
-  // we only care about these message types
-  switch (m->get_type()) {
-  case CEPH_MSG_MON_MAP:
-  case CEPH_MSG_AUTH_REPLY:
-  case CEPH_MSG_MON_SUBSCRIBE_ACK:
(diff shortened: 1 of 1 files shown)


src/mon/MonClient.cc
```diff
@@ -250,19 +250,6 @@ bool MonClient::ms_dispatch(Message *m)
...
-  case CEPH_MSG_AUTH_REPLY:
-  case CEPH_MSG_MON_SUBSCRIBE_ACK:
-  case CEPH_MSG_MON_GET_VERSION_REPLY:
-  case MSG_MON_COMMAND_ACK:
-  case MSG_LOGACK:
-    break;
-  default:
-    return false;
-  }
-
```
gregsfortytwo: Notice how on line 271 we put() the message and return true, whereas here we are just returning false?

These lines are a necessary filter so that we only bother looking at monitor messages we can handle, and don't steal and throw away messages for other dispatchers in the system.

gregsfortytwo:
As commented, these lines are definitely needed!

src/mon/MonClient.cc
dongbula: Well, it seems like I made a mistake.Thanks for your explaining.

PR #11120, opened 2016-09-18
mds/server: clean up handle_client_open()

renhwztetecs:

### DIFF
--- a/src/mds/Server.cc
+++ b/src/mds/Server.cc
@@ -2889,18 +2889,16 @@ void Server::_lookup_ino_2(MDRequestRef& mdr, int r)
 void Server::handle_client_open(MDRequestRef& mdr)
 {
   MClientRequest *req = mdr->client_request;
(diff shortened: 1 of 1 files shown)


src/mds/Server.cc
```diff
@@ -2892,15 +2892,13 @@ void Server::handle_client_open(MDRequestRef& mdr)
 
   int flags = req->head.args.open.flags;
   int cmode = ceph_flags_to_mode(flags);
-
-  bool need_auth = !file_mode_is_readonly(cmode) || (flags & O_TRUNC);
-
-  dout(7) << "open on " << req->get_filepath() << dendl;
```
gregsfortytwo: We want this output line prior to quitting so that it shows up in debug logs. Please don't move it!

renhwztetecs: ohh, it's done
thanks for your views

jcsp:

PR #11121, opened 2016-09-18
osd/OSD.h: move some members under private

stiopaa1:

### DIFF
--- a/src/osd/OSD.h
+++ b/src/osd/OSD.h
@@ -845,7 +845,6 @@ public:
   Objecter *objecter;
   Finisher objecter_finisher;
 
-
   // -- Watch --
   Mutex watch_lock;
   SafeTimer watch_timer;
@@ -907,6 +906,7 @@ public:
 	  entity_inst_t())));
   }
 
+private:
   // -- pg recovery and associated throttling --
   Mutex recovery_lock;
   list<pair<epoch_t, PGRef> > awaiting_throttle;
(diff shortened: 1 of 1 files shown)


liewegas:
This PR inadvertantly updates the rocksdb submodule

Also, I think these changes are on the threshold of worthwhile.  We clean up the style a bit, but there's no performance impact or readability impact.  I would suggest taking a look at the Cleanup list on the trello board at https://trello.com/b/ugTc2QFH/ceph-backlog# for ideas.  count() + operatorp[] -> find() cleanups, for example, are great, or const cleanups in general.  Another good one is replacing our kludgey xlist/dlist/elist implementation in the MDS with boost::intrusive::list, which is strictly better in all respects (memory utilization being the most important one).

Thanks!

stiopaa1:
@liewegas 
>This PR inadvertantly updates the rocksdb submodule

Ooops, I have corrected this. 

>We clean up the style a bit, but there's no performance impact or readability impact.

Surely making as many members as possible private makes it easier to reason about the code?

>Another good one is replacing our kludgey xlist/dlist/elist implementation in the MDS with boost::intrusive::list, which is strictly better in all respects (memory utilization being the most important one).

I will have a look at this at some point next week, thanks.

liewegas:
Private is good, yes.  I'm just suggesting that some of the other cleanup items on the list are only slightly more involved and would be higher-impact.

Thanks!

PR #11122, opened 2016-09-18
compressor/zlib: switch to raw deflate

branch-predictor:
By using raw deflate instead of default zlib deflate, we're saving 6 bytes on each compression stream and some CPU time by not calculating Adler32 during both compression and decompression (which is waste of time as Bluestore has its own checksumming code that is *also* executed).
CPU time savings are more visible on decompression (depends on compressed stream length; from 5 to 20% cpu time saved), but compression also sees gains (up to 5% in my testing).

Signed-off-by: Piotr Dałek <git@predictor.org.pl>

### DIFF
--- a/src/compressor/zlib/ZlibCompressor.cc
+++ b/src/compressor/zlib/ZlibCompressor.cc
@@ -38,20 +38,30 @@ _prefix(std::ostream* _dout)
 
 #define MAX_LEN (CEPH_PAGE_SIZE)
 
+// default window size for Zlib 1.2.8, negated for raw deflate
+#define ZLIB_DEFAULT_WIN_SIZE -15
+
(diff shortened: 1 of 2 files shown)


liewegas:
sounds great!  hmm, we should figure out if/how users can configure the compression level, though.

branch-predictor:
On Sun, Sep 18, 2016 at 03:28:52PM -0700, Sage Weil wrote:
> sounds great!  hmm, we should figure out if/how users can configure the compression level, though.

How about changing the Compressor instance constructor and passing
compression level through it?

liewegas:
Yeah.. let's follow up with that later though after @ifed01 's pr merged

liewegas:
retest this please

PR #11124, opened 2016-09-19
rgw: remove a redundant judgement in rgw_rados.cc:delete_obj.

zhangsw:
The value of the 'removed' is just the same as 'r>=0'. We can merge the two
judgement and remove the var 'removed'. Because the delete_obj is always called when
we delete obj from ceph, remove the judgement may gain a little improvement.


mattbenjamin:
@zhangsw I guess should just see test suite run

yuriw:
ready for merge http://pulpito.ceph.com/yuriw-2017-07-12_19:38:07-rgw-wip-yuri-testing2_2017_7_12_2-distro-basic-smithi/

PR #11130, opened 2016-09-19
rgw: partial mesh

yehudasa:
Allows configuring partial mesh architecture within a single zone-group. By default it's still a full mesh, that is all zones sync from all other zones. With this change we can specify for each zone whether it's pulling data from all other zones, or which of the zones it'd be pulling the data from.

### DIFF
--- a/src/rgw/rgw_admin.cc
+++ b/src/rgw/rgw_admin.cc
@@ -1959,9 +1959,9 @@ static void sync_status(Formatter *formatter)
     string zone_name;
     string source_str = "source: ";
     string s = source_str + source_id;
-    auto siter = store->zone_name_by_id.find(source_id);
-    if (siter != store->zone_name_by_id.end()) {
-      s += string(" (") + siter->second + ")";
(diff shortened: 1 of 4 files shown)


src/rgw/rgw_admin.cc
```diff
@@ -1993,6 +2006,11 @@ int main(int argc, char **argv)
   std::string zonegroup_name, zonegroup_id, zonegroup_new_name;
   std::string api_name;
   list<string> endpoints;
+  int tmp_int;
+  int sync_from_all_specified = false;
```
cbodley: int -> bool

idealguo:
Hi, this feature is important for multi clusters environment(in same city), How about the progress? Do we have chance to see it in next version?

yehudasa:
@idealguo it's not going to be in kraken, planned for the version that follows it.

yehudasa:
hmm, apparently all the changes in here already found themselves into master as part of the sync modules feature. Closing.

PR #11131, opened 2016-09-19
Rgw lifecycle testing

dang:


src/rgw/rgw_lc.cc
```diff
@@ -73,6 +73,9 @@ void *RGWLC::LCWorker::entry() {
 }
 
 void RGWLC::initialize(CephContext *_cct, RGWRados *_store) {
+#define COOKIE_LEN 16
+  char buf[COOKIE_LEN + 1];
+
```
cbodley: the for loop below has a `char buf[32];` that shadows this one. could you either rename this, or move its declaration down to `gen_rand_alphanumeric()` so their scopes don't overlap?

PR #11133, opened 2016-09-19
fix Unknown lvalue 'TasksMax' in section 'Service'

zealoussnow:
When I run systemctl restart ceph-osd@0, I got some mgs from 'systemctl status ceph-osd@0'
```
Sep 20 01:11:51 ceph1 systemd[1]: [/usr/lib/systemd/system/ceph-osd@.service:17] Unknown lvalue 'TasksMax' in section 'Service'
```

Signed-off-by: LeoZhang <nguzcf@gmail.com>

### DIFF
--- a/systemd/ceph-mon@.service
+++ b/systemd/ceph-mon@.service
@@ -21,7 +21,6 @@ PrivateDevices=yes
 ProtectHome=true
 ProtectSystem=full
 PrivateTmp=true
-TasksMax=infinity
 Restart=on-failure
 StartLimitInterval=30min
 StartLimitBurst=3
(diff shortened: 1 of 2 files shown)


systemd/ceph-mon@.service
```diff
@@ -21,7 +21,6 @@ PrivateDevices=yes
 ProtectHome=true
 ProtectSystem=full
 PrivateTmp=true
-TasksMax=infinity
```
tchaikov: `TasksMax` was introduced by 05cafcf19fb80dd2a76fdbbe5329bae53aa0def4 [0]. seems this setting does not exist in systemd version earlier than 226. and RHEL/Centos7.2 still have v219[1]. so does some relatively old distro releases. debian jessie has systemd v215[2].

---
[0] https://github.com/ceph/ceph/pull/8450#issuecomment-219486514
[1] https://bugzilla.redhat.com/show_bug.cgi?id=1199644
[2] https://packages.debian.org/jessie/systemd

zealoussnow:
@tchaikov OK. I'm so sorry that I only test in CentOS-7.1, and I will remember this issue. Thank you for your advice.

PR #11134, opened 2016-09-19
Add two options to radosgw-admin.rst manpage

tserlin:
Added '--job-id' and '--max-concurrent-ios' to Orphan Search Options


ktdreyer:
Thanks! Here are some things to update:

- Change the commit log message from the general "`Update radosgw-admin.rst`" to "`Add two options to radosgw-admin.rst manpage`" then do a double-line break, and in the body of the commit message `Add '--job-id' and '--max-concurrent-ios' to Orphan Search Options`
-  Add your `Signed-off-by:`
- Add a line that references the Redmine ticket http://tracker.ceph.com/issues/17281 (see "`git log`" for examples of how other people have formatted this)

PR #11135, opened 2016-09-19
osd/PG.h: change map key to string

stiopaa1:
We can't use const char\* for a map key without giving it a comparison
functor. Without it we are comparing pointer addresses not the strings.
For example:

``` c++
std::map<const char*, int> test;
char one[] = "one";
test["one"] = 5;
test[one] = 5;
std::cout << test.size();
```

will return 2.

### DIFF
--- a/src/osd/PG.h
+++ b/src/osd/PG.h
@@ -94,7 +94,7 @@ struct PGRecoveryStats {
     // cppcheck-suppress unreachableCode
     per_state_info() : enter(0), exit(0), events(0) {}
   };
-  map<const char *,per_state_info> info;
+  map<string, per_state_info> info;
   Mutex lock;
 
   PGRecoveryStats() : lock("PGRecoverStats::lock") {}
@@ -105,7 +105,7 @@ struct PGRecoveryStats {
   }
   void dump(ostream& out) {
     Mutex::Locker l(lock);
-    for (map<const char *,per_state_info>::iterator p = info.begin(); p != info.end(); ++p) {
+    for (map<string, per_state_info>::iterator p = info.begin(); p != info.end(); ++p) {
       per_state_info& i = p->second;
       out << i.enter << "\t" << i.exit << "\t"
 	  << i.events << "\t" << i.event_time << "\t"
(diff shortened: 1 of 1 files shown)


badone:
Can you remove the change to subproject commit change to src/rocksdb?

src/osd/PG.h
```diff
@@ -94,7 +94,7 @@ struct PGRecoveryStats {
     // cppcheck-suppress unreachableCode
     per_state_info() : enter(0), exit(0), events(0) {}
   };
-  map<const char *,per_state_info> info;
+  map<string, per_state_info> info;
```
tchaikov: @stiopaa1 even it looks suspicious, it is actually correct, i think. the `const char *` keys are the `NamedState::state_name`, and they live in .rodata section, so

1. they won't go away.
1. and they are unique. in other words, two difference instances of the same subclass of `NamedState` will share the same `state_name`:
```c++
Crashed c1, c2;
assert(c1.state_name == c2.state_name);
```
1. and it's fine to use the addresses of the strings as the keys for info. as we don't care about the order in which the `per_state_info` instances are arranged in info. what we need is actually an unordered_map here.

tchaikov:
```c++
std::map<const char*, int> test;
char one[] = "one";
test["one"] = 5;
test[one] = 5;
std::cout << test.size();
```

well, the `char one[]` is a writable char array located on the stack. while the string literal `"one"` at line of `test["one"] = 5;` lives in the .rodata section. so their addresses are different for sure. and that's not the case here.

stiopaa1:
@tchaikov 
>2. and they are unique. in other words, two difference instances of the same subclass of NamedState will share the same state_name:
Crashed c1, c2;
assert(c1.state_name == c2.state_name);

I don't think this is guaranteed. When you look at the compiler option for gcc: 
-fmerge-constants (https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html)

`Attempt to merge identical constants (string constants and floating-point constants) across compilation units.
This option is the default for optimized compilation if the assembler and linker support it. Use -fno-merge-constants to inhibit this behavior.
Enabled at levels -O, -O2, -O3, -Os. `

An attempt.. so it might happen that the addresses for the same string literal to be different which will  cause creation of multiple entries for the same value.

tchaikov:
makes sense, how about defining static constants in `PG.cc` for these literal strings. and use them instead? because

1. the `state_name`s are actually predefined strings. they are practically the string representation of their `enum`.
1. we can save the cpu cycle doing the lexical comparison.

dachary:

dachary:

liewegas:
closing this for now!

PR #11137, opened 2016-09-19
mds: catch duplicates in DamageTable

jcsp:
There was an implicit assumption in the code that callers
wouldn't hit the notify_*damaged paths twice because they would
have checked is_*_damaged paths first.

However, that's not really true in all cases, e.g. scrub
code isn't required to respect existing damage entries
when trying to load a CDir.

Simply fix this by having the DamageTable notify* functions
check the key they're inserting doesn't already exist.

Fixes: http://tracker.ceph.com/issues/17173

### DIFF
--- a/src/mds/DamageTable.cc
+++ b/src/mds/DamageTable.cc
@@ -46,10 +46,13 @@ bool DamageTable::notify_dentry(
     return true;
   }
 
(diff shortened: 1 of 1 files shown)


jcsp:
@fullerdj quick review please?

fullerdj:
Looks good to me.

PR #11138, opened 2016-09-19
mds: handle blacklisting during journal recovery

jcsp:
EBLACKLISTED was being incorrectly handled as an
indication of metadata damage.

Fixes: http://tracker.ceph.com/issues/17236


jcsp:
@gregsfortytwo could you review please?

jcsp:
Was tested on this branch: http://pulpito.ceph.com/jspray-2016-09-21_18:37:24-fs-wip-jcsp-greenish-distro-basic-smithi/

PR #11140, opened 2016-09-19
hammer: add orphan options to radosgw-admin --help and man page

cbodley:
hammer backport for issues http://tracker.ceph.com/issues/17280 and http://tracker.ceph.com/issues/17281

(required pulling in 6141cfa69764b1861d46ff50880c44cdaab70d65 as well)

### DIFF
--- a/doc/man/8/radosgw-admin.rst
+++ b/doc/man/8/radosgw-admin.rst
@@ -167,6 +167,11 @@ Orphans Search Options
 
 	Number of shards to use for keeping the temporary scan info
 
+.. option:: --orphan-stale-secs
+
+        Number of seconds to wait before declaring an object to be an orphan.
+        Default is 86400 (24 hours).
+
 
 Examples
 ========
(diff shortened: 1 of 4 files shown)


ktdreyer:
Would you please update the nroff code in `man/radosgw-admin.8` as well? Without that change, it will not make it into the hammer man pages that users read with `man radosgw-admin` :cry:

cbodley:
@ktdreyer added a commit for radosgw-admin.8 and tested with `nroff -man`:
```
ORPHANS SEARCH OPTIONS
       --pool Data pool to scan for leaked rados objects

       --num-shards
              Number of shards to use for keeping the temporary scan info

       --orphan-stale-secs
              Number  of  seconds  to wait before declaring an object to be an
              orphan.  Default is 86400 (24 hours).

       --job-id
              Set the job id (for orphans find)

       --max-concurrent-ios
              Maximum concurrent ios for orphans find.  Default is 32.
```

PR #11142, opened 2016-09-20
rbd-mirror: cleanup status global from rbd mirror if image removed

hzhang-wx:
Signed-off-by: Huan Zhang <zhanghuan@chinac.com>


chenliuzhong:
@hzhang-wx  I wonder is there any problem if I don't cleanup status global? Thank you!

PR #11143, opened 2016-09-20
osd: fix rados write op hang

wwformat:
if primary OSD handle a write op and do_osd_ops return errro code,
record_write_error will try to record it into pglog and send log message
to replica.
but if replica OSD crash right now, the primary OSD will call
on_change to free all repop, and do_update_log_missing_reply will not
be called, so the write op will lost and hangs.

this patch fix it.

Signed-off-by: Yunchuan Wen <yunchuan.wen@kylin-cloud.com>


liewegas:
@jdurgin ping

jdurgin:
nice catch! looks good to me

tchaikov:
tested at tested at http://pulpito.ceph.com/kchai-2016-10-24_03:13:40-rados-wip-kefu-testing---basic-mira/

PR #11144, opened 2016-09-20
crush/CrushCompiler: error out as long as parse fails

tchaikov:
do not output compiled crush map if anything goes wrong when parsing
crush map.

Fixes: http://tracker.ceph.com/issues/17306


badone:
lgtm Kefu, and tests OK.

Love your work buddy.

PR #11146, opened 2016-09-20
os/filestore/FileJournal: fail out if FileJournal is not block device…

tchaikov:
… or regular file

otherwise JournalingFileStore will assert when deleting FileJournal
which still has the non block/regular file opened.

Fixes: http://tracker.ceph.com/issues/17307

### DIFF
--- a/src/os/filestore/FileJournal.cc
+++ b/src/os/filestore/FileJournal.cc
@@ -89,13 +89,17 @@ int FileJournal::_open(bool forwrite, bool create)
 
   if (S_ISBLK(st.st_mode)) {
     ret = _open_block_device();
-  } else {
+  } else if (S_ISREG(st.st_mode)) {
     if (aio && !force_aio) {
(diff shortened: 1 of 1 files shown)


dzafman:
:+1:

src/os/filestore/FileJournal.cc
```diff
@@ -134,6 +138,7 @@ int FileJournal::_open(bool forwrite, bool create)
 
  out_fd:
   VOID_TEMP_FAILURE_RETRY(::close(fd));
+  fd = -1;
```
dzafman: @tchaikov This one line is really the fix. but the other changes improve the error output.

tchaikov: thanks for helping me understand this fix.

dzafman: I led you astray.  Actually, the open doesn't fail, so without the S_ISREG check, it would crash later upon read failure.

PR #11148, opened 2016-09-20
librbd: block name prefix might overflow fixed size C-string

dillaman:
The issue which resulted in too large v2 image ids was fixed
under #16887.

Fixes: http://tracker.ceph.com/issues/17310


trociny:
lgtm

PR #11150, opened 2016-09-20
mds/server: merge the snapshot request judgment

renhwztetecs:
merge the snapshot request judgment,
and all return -EROFS


jcsp:
Looks reasonable, and I agree with changing the EPERM to an EROFS to make it consistent with the other case.  Adding to list for testing.

renhwztetecs:
@jcsp 
add to ./src/test/libcephfs/test.cc?
allow_new_snaps is not open, and no snap testunit in test.cc.

PR #11152, opened 2016-09-20
librbd: potential deadlock closing image with in-flight readahead

dillaman:
Fixes: http://tracker.ceph.com/issues/17198


trociny:
lgtm

PR #11156, opened 2016-09-20
compressor/zstd: add zstd compression plugin

liewegas:
- The recursive build of the submodule is not ideal.
- We'll eventually want to drop the submodule.  For now, we need it, though: only xenial has a libzstd-dev, and it is very old.

### DIFF
--- a/src/compressor/zstd/CMakeLists.txt
+++ b/src/compressor/zstd/CMakeLists.txt
@@ -1,0 +1,30 @@
+# zstd
+
+# libzstd - build it statically
+set(ZSTD_EXTRA_CFLAGS -fPIC -Wno-unused-variable -O3)
+
+add_custom_target(build_zstd
+  COMMAND $(MAKE) libzstd CFLAGS="${ZSTD_EXTRA_CFLAGS}"
+  WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/src/zstd/lib
+  COMMENT "libzstd building")
+
+add_library(zstd STATIC IMPORTED)
+set_property(TARGET zstd PROPERTY IMPORTED_LOCATION "${CMAKE_SOURCE_DIR}/src/zstd/lib/libzstd.a")
+add_dependencies(zstd build_zstd)
+set(ZSTD_INCLUDE_DIR ${CMAKE_SOURCE_DIR}/src/zstd/lib)
+
+#
+set(zstd_sources
+  CompressionPluginZstd.cc
+)
+
+add_library(ceph_zstd SHARED ${zstd_sources})
+add_dependencies(ceph_zstd ${CMAKE_SOURCE_DIR}/src/ceph_ver.h)
+target_link_libraries(ceph_zstd zstd)
+set_target_properties(ceph_zstd PROPERTIES VERSION 2.0.0 SOVERSION 2)
+install(TARGETS ceph_zstd DESTINATION ${compressor_plugin_dir})
+
+if(WITH_EMBEDDED)
+  add_library(cephd_compressor_zstd STATIC ${zstd_sources})
+  set_target_properties(cephd_compressor_zstd PROPERTIES COMPILE_DEFINITIONS BUILDING_FOR_EMBEDDED)
+endif()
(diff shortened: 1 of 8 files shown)


alimaredia:
@liewegas @tchaikov @cbodley I found a couple ways to get rid of the repeat builds (which are caused by the libzstd target in the Makefile to have multiple compilations within it). 

Adding zstd/build/cmake to ceph's cmake via add_subdirectory() wasn't an option because both ceph and zstd have an uninstall target and a configure_file problem in zstd's cmake.  

Option #1 (http://paste.fedoraproject.org/438453/51801091/) use `cmake` instead of `make` in add_custom_target. Everything in the previous link is built out of tree!

Option #2 (http://paste.fedoraproject.org/438467/14751809) copy zstd'd cmake into ours. This looks cleaner and give us more control but there's issues with maintaining it and I'm not sure we would be able to compile libzstd.a with the same exact flags as are used when it's build with the submodule's cmake

Thoughts?

liewegas:
Option 1 seems better to me.  It's less maint going forward.  If we find that we have to change zstd build options in some way, we can always switch to option 2.

ktdreyer:
Ubuntu Yakkety ships libzstd v0.8.0, and the Debian packager recently pulled v1.0.0 into debian unstable earlier this month.

What version is new enough?

@b-ranto and I could get this into Fedora and EPEL.

liewegas:
I think 0.8.0 is good enough...

liewegas:
@ktdreyer We can add an option to use the system version like we do with rocksdb and boost.  It's hard to do with upstream builds though because the debian/control build dependencies don't vary per-distro, so  even if we autodetected in cmake it would never get installed in the pbuilder chroot.  I think the best thing to do is include the option and then let the downstream distros explicitly enable the system version as per their preference.

ktdreyer:
I want to keep downstream as close to upstream as possible, because every time we've deviated it's caused problems :(

I've filed a request to update to 0.8.0 in Xenial: https://bugs.launchpad.net/bugs/1629356

Looks like the review request for Fedora is already underway from some other folks. https://bugzilla.redhat.com/1373218

liewegas:
@ktdreyer Until it's in the downstreams I'm not sure which path to take here (autodetect if present or require build flag).  I'm not sure it matters, though, because in both cases a downstream control/spec file change is needed to enable the build dependency.  And I have a general dislike for autodetection because it makes builds nondeterministic...

ktdreyer:
Ok, thanks @liewegas . It sounds like we can revisit this when we drop Trusty support, because we can just depend on the system versions, then.

liewegas:
Put this on hold for now.  95% CPU time is spent initializing the dictionary for each compression call, making this significantly (10x) slower than the alternatives.  Asking upstream if there is a way to address this.  https://github.com/facebook/zstd/issues/408

liewegas:
The startup overhead will not be such an issue for RGW, which is now also using the compression abstraction.  Let's merge this!

src/compressor/zstd/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+# libzstd - build it statically
+set(ZSTD_EXTRA_CFLAGS -fPIC -Wno-unused-variable -O3)
+
+add_custom_target(build_zstd
+  COMMAND $(MAKE) libzstd CFLAGS="${ZSTD_EXTRA_CFLAGS}"
+  WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/src/zstd/lib
+  COMMENT "libzstd building")
+
+add_library(zstd STATIC IMPORTED)
+set_property(TARGET zstd PROPERTY IMPORTED_LOCATION "${CMAKE_SOURCE_DIR}/src/zstd/lib/libzstd.a")
```
tchaikov: this put the built asserts into the source directory. maybe we can put them into the binary directory? see https://github.com/tchaikov/ceph/commit/071562e6e961016bbc9254ef57d1d20e6c2b4f70
the patch follows the example of how we build rocksdb, so we can be more consistent when including external projects.

cbodley: @tchaikov yeah, i agree that we should build outside the source tree

liewegas: @cbodley @alimaredia would someone with more cmake mojo than I have mind take a stab at it?

tchaikov: @liewegas #13075

PR #11159, opened 2016-09-20
rgw_file: fix set_attrs operation

mattbenjamin:
The effective part of this change is to always set a value
for RGW_ATTR_UNIX_KEY1 (because it is expected later).

Secondarily, do not pass the address of the to-set attributes
buffer::list as remove attrs--this is confusing.

### DIFF
--- a/src/rgw/rgw_file.cc
+++ b/src/rgw/rgw_file.cc
@@ -604,6 +604,7 @@ namespace rgw {
     rgw_fh->encode_attrs(ux_key, ux_attrs);
 
     /* save attrs */
+    req.emplace_attr(RGW_ATTR_UNIX_KEY1, std::move(ux_key));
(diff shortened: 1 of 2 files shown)


src/rgw/rgw_op.h
```diff
@@ -1666,6 +1666,11 @@ class RGWSetAttrs : public RGWOp {
   RGWSetAttrs() {}
   virtual ~RGWSetAttrs() {}
 
+  buffer::list* get_attr(const std::string& k) {
+    auto iter = attrs.find(k);
+    return (iter != attrs.end()) ? &(iter->second) : nullptr;
+  }
+
```
cbodley: this function was added but unused - was this intentional?

mattbenjamin: no, I can remove it if actually unused

PR #11160, opened 2016-09-21
librbd: add support for snapshot namespaces

VictorDenisov:


src/cls/rbd/cls_rbd.cc
```diff
@@ -1488,6 +1489,34 @@ int get_snapshot_name(cls_method_context_t hctx, bufferlist *in, bufferlist *out
   return 0;
 }
 
+int get_snapshot_namespace(cls_method_context_t hctx, bufferlist *in, bufferlist *out)
```
dillaman: Minor: add block comment describing expected in/out params

src/cls/rbd/cls_rbd.cc
```diff
@@ -1488,6 +1489,34 @@ int get_snapshot_name(cls_method_context_t hctx, bufferlist *in, bufferlist *out
...
+  bufferlist::iterator iter = in->begin();
+  try {
+    ::decode(snap_id, iter);
+  } catch (const buffer::error &err) {
+    return -EINVAL;
+  }
+
+  CLS_LOG(20, "get_snapshot_namespace snap_id=%llu", (unsigned long long)snap_id);
+
+  if (snap_id == CEPH_NOSNAP)
```
dillaman: Minor:  wrap in braces

src/cls/rbd/cls_rbd.cc
```diff
@@ -1488,6 +1489,34 @@ int get_snapshot_name(cls_method_context_t hctx, bufferlist *in, bufferlist *out
...
+    return -EINVAL;
+
+  cls_rbd_snap snap;
+  string snapshot_key;
+  key_from_snap_id(snap_id, &snapshot_key);
+  int r = read_key(hctx, snapshot_key, &snap);
+  if (r < 0)
```
dillaman: Minor: same here

src/cls/rbd/cls_rbd.cc
```diff
@@ -1488,6 +1489,34 @@ int get_snapshot_name(cls_method_context_t hctx, bufferlist *in, bufferlist *out
...
+  uint64_t snap_id;
+
+  bufferlist::iterator iter = in->begin();
+  try {
+    ::decode(snap_id, iter);
+  } catch (const buffer::error &err) {
+    return -EINVAL;
+  }
+
+  CLS_LOG(20, "get_snapshot_namespace snap_id=%llu", (unsigned long long)snap_id);
```
dillaman: Minor: you can use ```PRIu64``` to directly encode snap_id

src/cls/rbd/cls_rbd_client.cc
```diff
@@ -495,25 +495,29 @@ namespace librbd {
...
            it != ids.end(); ++it) {
         snapid_t snap_id = it->val;
-        bufferlist bl1, bl2, bl3, bl4;
+        bufferlist bl1, bl2, bl3, bl4, bl5;
         ::encode(snap_id, bl1);
         op->exec("rbd", "get_snapshot_name", bl1);
-        ::encode(snap_id, bl2);
-        op->exec("rbd", "get_size", bl2);
+	::encode(snap_id, bl2);
+	op->exec("rbd", "get_snapshot_namespace", bl2);
```
dillaman: Since this cls method won't exist on older OSDs, unfortunately, until we get to the point of only supporting Kraken+, you need to retrieve the snapshot namespaces as part of an individual OSD op in case it errors out with -EOPNOTSUPP (which you could then ignore)

src/cls/rbd/cls_rbd_types.cc
```diff
@@ -316,5 +317,20 @@ bool GroupSpec::is_valid() const {
...
 }
 
+std::ostream& operator<<(std::ostream& os, const UserSnapshotNamespace& ns) {
+  os << "[user]";
+  return os;
+}
+
+std::ostream& operator<<(std::ostream& os, const GroupSnapshotNamespace& ns) {
+  os << "[group]";
+  return os;
```
dillaman: Minor: dump fields?

src/cls/rbd/cls_rbd_types.h
```diff
@@ -216,6 +217,140 @@ struct GroupSpec {
...
+			 const string &_snapshot_id) :group_pool(_group_pool),
+						      group_id(_group_id),
+						      snapshot_id(_snapshot_id) {}
+
+  int64_t group_pool;
+  string group_id;
+  string snapshot_id;
+
+  void encode(bufferlist& bl) const {
+    ::encode(group_pool, bl);
```
dillaman: Minor: move the implementation into the cc file

src/cls/rbd/cls_rbd_types.h
```diff
@@ -216,6 +217,140 @@ struct GroupSpec {
...
+  inline void operator()(T& t) const {
+    t.decode(m_iter);
+  }
+private:
+  bufferlist::iterator &m_iter;
+};
+
+inline void encode(const SnapshotNamespace &c, bufferlist &bl, uint64_t features=0)
+{
+  ENCODE_DUMP_PRE();
```
dillaman: I don't think this is doing what you are expecting -- this for for creating the binary file dump records

src/cls/rbd/cls_rbd_types.h
```diff
@@ -216,6 +217,140 @@ struct GroupSpec {
...
+};
+
+struct UserSnapshotNamespace {
+  static const uint32_t SNAPSHOT_NAMESPACE_TYPE = SNAPSHOT_NAMESPACE_TYPE_USER;
+
+  UserSnapshotNamespace() {}
+
+  void encode(bufferlist& bl) const {}
+  void decode(bufferlist::iterator& it) {}
+
```
dillaman: All the new types should have ```generate_test_instances``` and ```dump``` methods so that they can be registered with the dencoder test case.

dillaman: ... might be either to create a new SnapshotNamespace struct that just inherits from boost::variant.  That would allow you to create the necessary ```generate_test_instances``` static member function for use with dencoder. Then you wouldn't need ```generate_test_instances``` for each individual type.

src/cls/rbd/cls_rbd_types.h
VictorDenisov: I'm not sure I understand this comment. Are you talking about ENCODE_DUMP_PRE(); specifically? I see that WRITE_CLASS_ENCODER is used in order to create encode and decode functions for MirrorImageStatus. WRITE_CLASS_ENCODER adds ENCODE_DUMP_PRE and ENCODE_DUMP post inside encode function it generates.

dillaman: If you look at the code for ENCODE_DUMP_PRE -- most of the time it's a no-op. The only time it actually does something is within the dencoder test where it writes the binary encoding format out to a file. I think you were looking for ENCODE_START/ENCODE_FINISH which add version control.

src/cls/rbd/cls_rbd_client.cc
VictorDenisov: I see. Too bad I didn't think about it.

VictorDenisov:
not ready yet. will update when done with addressing comments

src/cls/rbd/cls_rbd_types.h
VictorDenisov: I tried to implement it using inheritance and couldn't reconcile all types of constructors that are provided by boost::variant. It looks like the solution is to have a wrapper type like NotifyMessage wraps Payload and EventEntry wraps Event type. Those wrappers implement all the required methods.

dillaman:
... all internal plumbing changes, no doc updates required

VictorDenisov:
@dillaman could you please take a look if you like this version at all. In the mean time I'll look into what happened to the tests.

src/cls/rbd/cls_rbd_types.cc
```diff
@@ -316,5 +317,138 @@ bool GroupSpec::is_valid() const {
   return (!group_id.empty()) && (pool_id != -1);
 }
 
+void GroupSnapshotNamespace::encode(bufferlist& bl) const {
+  ENCODE_START(1, 1, bl);
```
dillaman: Minor: probably no need to encode another version here since it's already wrapped in a versioned struct

src/cls/rbd/cls_rbd_types.cc
```diff
@@ -316,5 +317,138 @@ bool GroupSpec::is_valid() const {
...
+  boost::apply_visitor(DecodeSnapshotNamespaceVisitor(p), snapshot_namespace);
+  DECODE_FINISH(p);
+}
+
+void SnapshotNamespaceEntry::dump(Formatter *f) const {
+  boost::apply_visitor(DumpSnapshotNamespaceVisitor(f, "snapshot_namespace_type"), snapshot_namespace);
+}
+
+void SnapshotNamespaceEntry::generate_test_instances(std::list<SnapshotNamespaceEntry *> &o) {
+
```
dillaman: Minor: append a ```UserSnapshotNamespace``` and ```GroupSnapshotNamespace``` example to the list

src/cls/rbd/cls_rbd_types.h
```diff
@@ -216,6 +217,104 @@ struct GroupSpec {
...
+  void dump(Formatter *f) const {}
+
+  inline bool operator==(const UserSnapshotNamespace& usn) const {
+    return true;
+  }
+
+};
+
+WRITE_CLASS_ENCODER(UserSnapshotNamespace);
```
dillaman: Minor: since these will always be encoded via ```SnapshotNamespaceEntry```, no need to create encode/decode free functions for these types.

src/librbd/ImageCtx.h
```diff
@@ -230,6 +231,8 @@ namespace librbd {
     const SnapInfo* get_snap_info(librados::snap_t in_snap_id) const;
     int get_snap_name(librados::snap_t in_snap_id,
 		      std::string *out_snap_name) const;
+    int get_snap_namespace(librados::snap_t in_snap_id,
+			   cls::rbd::SnapshotNamespaceEntry *out_snap_namespace) const;
```
dillaman: Minor: could directly use ```SnapshotNamespace``` variant since ```SnapshotNamespaceEntry``` is only used for encode/decode and requires an indirection.

src/librbd/image/RefreshRequest.h
```diff
@@ -154,6 +155,9 @@ class RefreshRequest {
   void send_v2_get_snapshots();
   Context *handle_v2_get_snapshots(int *result);
 
+  void send_v2_get_snap_namespaces();
```
dillaman: Minor: can you update the state transition diagram in this header

src/librbd/journal/Types.h
```diff
@@ -135,9 +136,10 @@ struct SnapCreateEvent : public SnapEventBase {
 
   SnapCreateEvent() {
   }
-  SnapCreateEvent(uint64_t op_tid, const std::string &snap_name)
-    : SnapEventBase(op_tid, snap_name) {
+  SnapCreateEvent(uint64_t op_tid, const std::string &snap_name, const cls::rbd::SnapshotNamespaceEntry &_snap_namespace)
```
dillaman: Would there be a need to journal group snapshot create events? It seems like the group actions would be journaled via a group journal (eventually). If group snap create events are recorded with the image, they wouldn't be properly associated with a group and you wouldn't be able to remove them (perhaps). Therefore, perhaps this should wait until after group actions can be journaled and replayed.

dillaman:
@VictorDenisov can you add some tests to test_cls_rbd.cc?

src/cls/rbd/cls_rbd_types.cc
VictorDenisov: Should GroupSnapshotNamespace have any meaningful ids in it or any random values are enough?

dillaman: Random but valid is fine -- it's to help ensure that the format isn't unintentionally broken by future changes.

VictorDenisov:
not yet ready. addressing comments is in progress

dillaman:
@VictorDenisov just double checking if this is still not ready

VictorDenisov:
Not ready. I was distracted by something else. Will resume the work on this pull request next week.

dillaman:
@VictorDenisov just a heads-up: we are rapidly approaching the end of the Kraken development cycle.

VictorDenisov:
@dillaman This PR is ready. I'm not sure about the rest of snapshot code. When is the kraken hard code freeze?

dillaman:
@VictorDenisov There is no hard deadline established right now, but I believe once EC overwrite support and some additional BlueStore tweaks are in and stable, that will be the freeze.  I would estimate around the end-of-the-month. If consistency groups aren't completed by then, we will just need to revert the rbd CLI and API changes on the kraken branch.

VictorDenisov:
Ok. I understand.

src/cls/rbd/cls_rbd.cc
```diff
@@ -1512,6 +1553,7 @@ int snapshot_add(cls_method_context_t hctx, bufferlist *in, bufferlist *out)
   try {
     bufferlist::iterator iter = in->begin();
     ::decode(snap_meta.name, iter);
+    ::decode(snap_meta.snapshot_namespace, iter);
```
dillaman: ... also update the block comment above to list the new param

src/librbd/image/RefreshRequest.cc
```diff
@@ -905,7 +953,7 @@ void RefreshRequest<I>::apply() {
         parent = m_snap_parents[i];
       }
 
-      m_image_ctx.add_snap(m_snap_names[i], m_snapc.snaps[i].val,
+      m_image_ctx.add_snap(m_snap_names[i], cls::rbd::UserSnapshotNamespace(), m_snapc.snaps[i].val,
```
dillaman: Minor: should be ```m_snap_namespaces[i]```

src/librbd/librbd.cc
```diff
@@ -1013,8 +1016,18 @@ namespace librbd {
   {
     ImageCtx *ictx = (ImageCtx *)ctx;
     tracepoint(librbd, snap_remove_enter, ictx, ictx->name.c_str(), ictx->snap_name.c_str(), ictx->read_only, snap_name);
+    cls::rbd::SnapshotNamespace snap_namespace;
+    int r = get_snap_namespace(ictx, snap_name, &snap_namespace);
+    if (r < 0) {
+      tracepoint(librbd, snap_remove_exit, r);
+      return r;
+    }
+    if (!(snap_namespace == static_cast<cls::rbd::SnapshotNamespace>(cls::rbd::UserSnapshotNamespace()))) {
```
dillaman: Minor: will this assert if the cast fails?

src/tools/rbd_mirror/image_sync/SnapshotCopyRequest.h
```diff
@@ -104,6 +104,7 @@ class SnapshotCopyRequest : public BaseRequest {
   librados::snap_t m_prev_snap_id = CEPH_NOSNAP;
 
   std::string m_snap_name;
+  cls::rbd::SnapshotNamespace m_snap_namespace;
```
dillaman: Minor: should only handle UserSnapshotNamespace types -- skip GroupSnapshotNamespace

VictorDenisov: Do you mean you want me to declare only UserSnapshotNamespace here? Then when I retrieve a snapshot_namespace from SnapInfo I should check that what arrived is actually UserSnapshotNamespace. How should the code react if the namespace that arrived is not UserSnapshotNamespace? Is an assert(snap_info.snapshot_namespace == UserSnapshotNamespace) enough or you want a more intelligent and verbose handling?

VictorDenisov:
Addressed all except the last two comments.

src/librbd/librbd.cc
dillaman: Sorry, read that backwards. I think it would be cleaner to just use:

```if (boost::get<cls::rbd::UserSnapshotNamespace>(&snap_namespace) == nullptr)```

... also, this should probably be moved into the ```snap_remove``` method so that both the C and C++ API follow the same logic.

File: src/tools/rbd_mirror/image_sync/SnapshotCopyRequest.h (hunk 19 above)
Comment (dillaman) [inline src/tools/rbd_mirror/image_sync/SnapshotCopyRequest.h]: It was more of a comment that the ```SnapshotCopyRequest``` state machine should skip over any snapshots that aren't user snapshots. I think the ```m_remote_snap_ids``` and ```m_local_snap_ids``` collections should be filtered to only include user snapshots.

**Comment by VictorDenisov** [comment]
@dillaman could you please take a look at the current version of my changes. I'm not sure what happens to the tests yet. They seem to fail randomly.

**Comment by dillaman** [comment]
@VictorDenisov Still need to update librbd::watch_notify::SnapCreatePayload to encode/decode/dump the snapshot namespace -- and you need to bump the version to accommodate the change.

File: src/librbd/image/RefreshRequest.cc [hunk 20]
```diff
@@ -140,6 +140,11 @@ Context *RefreshRequest<I>::handle_v1_get_snapshots(int *result) {
     return m_on_finish;
   }
 
+  int n = m_snap_names.size();
```
dillaman: Minor: could just use ```m_snap_namespaces = {m_snap_names.size(), cls::rbd::UserSnapshotNamespace()};```

src/librbd/image/RefreshRequest.cc
```diff
@@ -386,6 +393,52 @@ Context *RefreshRequest<I>::handle_v2_get_snapshots(int *result) {
...
+    bufferlist::iterator it = m_out_bl.begin();
+    *result = cls_client::snap_namespace_list_finish(&it, m_snapc.snaps,
+						     &m_snap_namespaces);
+  }
+  if (*result == -ENOENT) {
+    ldout(cct, 10) << "out-of-sync snapshot state detected" << dendl;
+    send_v2_get_mutable_metadata();
+    return nullptr;
+  } else if (*result == -EOPNOTSUPP) {
+    m_snap_namespaces.resize(m_snapc.snaps.size());
```
dillaman: Minor: default to ```cls::rbd::UserSnapshotNamespace```?

src/tools/rbd_mirror/image_sync/SnapshotCopyRequest.cc
```diff
@@ -354,8 +368,23 @@ void SnapshotCopyRequest<I>::send_snap_protect() {
     librados::snap_t remote_snap_id = *snap_id_it;
 
     m_remote_image_ctx->snap_lock.get_read();
+
```
dillaman: Minor: I don't think you'll need this for the unprotect/protect states since only user snapshots should be protected -- but I think you need to skip over non-user snapshots in the remove state machine for now.

VictorDenisov:
@dillaman done. it looks like those test failures are unrelated. every time I push different tests sigfault.

VictorDenisov:
@dillaman fixed.

dillaman:
@VictorDenisov TestClsRbd failed: http://qa-proxy.ceph.com/teuthology/jdillaman-2016-10-18_21:31:07-rbd-wip-jd-testing---basic-smithi/483908/teuthology.log

VictorDenisov:
ok. looking into it.

src/cls/rbd/cls_rbd.cc
```diff
@@ -1678,7 +1678,8 @@ int snapshot_add(cls_method_context_t hctx, bufferlist *in, bufferlist *out)
 	        (unsigned long long)snap_id.val);
 	return -EIO;
       }
-      if (snap_meta.name == old_meta.name || snap_meta.id == old_meta.id) {
+      if ((snap_meta.name == old_meta.name || snap_meta.id == old_meta.id)
```
dillaman: The snap id should be unique regardless of the namespace -- so the check should be:

```
if (snap_meta.id == old_meta.id ||
    (snap_meta.name == old_meta.name &&
     snap_meta.snapshot_namespace == old_meta.snapshot_namespace)) {
```

VictorDenisov:
done

src/cls/rbd/cls_rbd.cc
```diff
@@ -1678,7 +1678,9 @@ int snapshot_add(cls_method_context_t hctx, bufferlist *in, bufferlist *out)
 	        (unsigned long long)snap_id.val);
 	return -EIO;
       }
-      if (snap_meta.name == old_meta.name || snap_meta.id == old_meta.id) {
+      if ((snap_meta.name == old_meta.name
+	    && snap_meta.snapshot_namespace == old_meta.snapshot_namespace)
```
dillaman: Minor: the conditionals should go at the end of the line and not the start of a new line:

```
if ((snap_meta.name == old_meta.name &&
     snap_meta.snapshot_namespace == old_meta.snapshot_namespace) ||
    snap_meta.id == old_meta.id) {
```

VictorDenisov: oops. sorry. too much haskell before this commit. fixed.

PR #11161, opened 2016-09-21
lttng: remove the source files created by lttng-gen-tp

tchaikov:
they will be created at build time in the build directory.

### DIFF
--- a/src/tracing/librados.c
+++ b/src/tracing/librados.c
@@ -1,6 +1,0 @@
-
-#define TRACEPOINT_CREATE_PROBES
-/*
- * The header containing our TRACEPOINT_EVENTs.
- */
-#include "librados.h"
--- a/src/tracing/librbd.c
+++ b/src/tracing/librbd.c
@@ -1,6 +1,0 @@
-
-#define TRACEPOINT_CREATE_PROBES
-/*
- * The header containing our TRACEPOINT_EVENTs.
- */
-#include "librbd.h"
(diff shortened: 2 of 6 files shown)


tchaikov:
@dillaman could you take a look when you have some time? thanks!

dillaman:
@tchaikov Are the generated C files exactly the same? I can't remember now why the switch to dynamic tp libraries required that I stopped auto-generated the C files -- hopefully it was just an autotools limitation.

tchaikov:
no, they not identical.

the one created by lttng-tp-gen

```c

#define TRACEPOINT_CREATE_PROBES
/*
 * The header containing our TRACEPOINT_EVENTs.
 */
#define TRACEPOINT_DEFINE
#include "tracing/librados.h"
```

the one in repo
```c

#define TRACEPOINT_CREATE_PROBES
/*
 * The header containing our TRACEPOINT_EVENTs.
 */
#include "librados.h"
```

but we are not compiling the ones in repo at least in the cmake build now.

dillaman:
Ah -- so that's what it was.  The ```TRACEPOINT_DEFINE``` macro needed to be in the librados/librbd library, not in the dynamic library.

dillaman:
... I have a feeling that LTTng tracepoints are broken in the master branch now since we really don't have any test cases to make sure they are working.

dillaman:
http://lttng.org/docs/#doc-dynamic-linking -- as per the docs, the ```TRACEPOINT_DEFINE``` macro should only be defined in the application (librbd/librados/etc), not the new shared library.

tchaikov:
@dillaman thanks for your review, will fix it!

PR #11165, opened 2016-09-21
common/throttle: simplify Throttle::_wait()

xiexingguo:
We don't have to check/set the "waited" member when falling into this branch,
which is deemed to be true.


yuriw:
http://pulpito.front.sepia.ceph.com/yuriw-2016-10-20_04:33:22-rados-wip-yuri-testing2_2016_10_19-distro-basic-smithi/
and rerun
http://pulpito.ceph.com/yuriw-2016-10-20_17:45:28-rados-wip-yuri-testing2_2016_10_19-distro-basic-smithi/

PR #11167, opened 2016-09-21
mon: make MDSMonitor tolerant of slow mon elections

jcsp:
Previously MDS daemons would get failed incorrectly
when they appeared to have timed out due to
delays in calling into MDSMonitor that were
actually caused by e.g. slow leveldb writes leading
to slow mon elections.

Fixes: http://tracker.ceph.com/issues/17308

### DIFF
--- a/src/mon/MDSMonitor.cc
+++ b/src/mon/MDSMonitor.cc
@@ -480,26 +480,6 @@ bool MDSMonitor::prepare_update(MonOpRequestRef op)
   return true;
 }
 
-
-namespace {
-class C_Updated : public Context {
-  MDSMonitor *mm;
-  MonOpRequestRef op;
-public:
-  C_Updated(MDSMonitor *a, MonOpRequestRef c) :
-    mm(a), op(c) {}
-  void finish(int r) {
-    if (r >= 0)
-      mm->_updated(op);   // success
-    else if (r == -ECANCELED) {
-      mm->mon->no_reply(op);
-    } else {
-      mm->dispatch(op);        // try again
-    }
-  }
-};
-}
-
 bool MDSMonitor::prepare_beacon(MonOpRequestRef op)
 {
   op->mark_mdsmon_event(__func__);
@@ -699,7 +679,15 @@ bool MDSMonitor::prepare_beacon(MonOpRequestRef op)
   dout(7) << "prepare_beacon pending map now:" << dendl;
   print_map(pending_fsmap);
   
-  wait_for_finished_proposal(op, new C_Updated(this, op));
+  wait_for_finished_proposal(op, new FunctionContext([op, this](int r){
+    if (r >= 0)
+      _updated(op);   // success
+    else if (r == -ECANCELED) {
+      mon->no_reply(op);
+    } else {
+      dispatch(op);        // try again
+    }
+  }));
 
   return true;
 }
--- a/src/mon/MDSMonitor.cc
+++ b/src/mon/MDSMonitor.cc
@@ -2965,8 +2965,25 @@ void MDSMonitor::tick()
     do_propose |= maybe_expand_cluster(i.second);
   }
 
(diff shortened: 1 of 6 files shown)


src/mon/MDSMonitor.cc
```diff
@@ -2952,6 +2952,24 @@ bool MDSMonitor::maybe_promote_standby(std::shared_ptr<Filesystem> fs)
...
+    // for whatever reason) or cluster election slowness (a long gap
+    // between calls while an election happened)
+    dout(4) << __func__ << ": resetting beacon timeouts due to mon delay "
+            "(slow election?) of " << now - last_tick << " seconds" << dendl;
+    for (auto &i : last_beacon) {
+      i.second.stamp = now;
+    }
+  }
+
+  last_tick = now;
```
gregsfortytwo: So this sets last_tick on peon monitors, which means if they get promoted to leader in less time than mds_beacon_interval+mds_beacon_grace they won't go through the beacon reset process...

src/mon/MDSMonitor.cc
```diff
@@ -2968,15 +2986,14 @@ void MDSMonitor::tick()
...
-  utime_t now = ceph_clock_now(g_ceph_context);
   utime_t cutoff = now;
   cutoff -= g_conf->mds_beacon_grace;
 
   // make sure last_beacon is fully populated
   for (const auto &p : pending_fsmap.mds_roles) {
     auto &gid = p.first;
     if (last_beacon.count(gid) == 0) {
-      last_beacon[gid].stamp = ceph_clock_now(g_ceph_context);
+      last_beacon[gid].stamp = now;
```
gregsfortytwo: ...and if we had a (stale) last_beacon map (like from being leader before, because another mon is flapping) then this doesn't get triggered. And down below we will mark the MDS as failed.

src/mon/MDSMonitor.h
```diff
@@ -155,6 +155,11 @@ class MDSMonitor : public PaxosService {
 
   int _check_pool(const int64_t pool_id, std::stringstream *ss) const;
   mds_gid_t gid_from_arg(const std::string& arg, std::ostream& err);
+
+  // When did the mon last call into our tick() method?  Used for detecting
+  // when the mon was not updating us for some period (e.g. during slow
+  // election) to reset last_beacon timeouts
+  utime_t last_tick;
```
gregsfortytwo: So this needs to get reset on monitor elections or something. :/

liewegas:
@jcsp ping

src/mon/MDSMonitor.cc
jcsp: Hmm, usually peons will have an empty last_beacon when they come up, as it's only populated in prepare_beacon and in the leader-specific later part of tick.  That said, there's nothing to reset it when a leader goes back to being a peon, so one *could* end up with stale state there.  Maybe we need to clear out the ephemeral bits like last_beacon explicitly when we go leader->peon.

src/mon/MDSMonitor.h
jcsp: ...yep!  (wrote my earlier reply before reading to the end of your comments)

jcsp:
Updated, also made the last_beacon etc private while we're here.

liewegas:
Seems like we should still move the beacon reset code into a helper and call it explicitly when we win an election, so that the leader always has clean state?  I think the failsafe tick vs last_tick check should stay, though, to handle the loady slow mon case...

jcsp:
@liewegas the intent of the last_beacon.clear() in on_restart is to reset the beacon state on winning an election -- I haven't followed through the paths that actually invoke on_restart(), so is there somewhere else?

liewegas:
aha, yeah, lgtm!

jcsp:
Oops, I wrote Reviewed-by myself :grimacing:

PR #11169, opened 2016-09-21
cmake: remove unittest_memstore_clone from 'make check'

cbodley:
unittest_memstore_clone needs an --osd-data directory to run in, and
(unlike ceph_test_objectstore) it does not create/manage one for you.
removing from 'make check' so the jenkins bot will stop failing


tchaikov:
@cbodley i prepared another fix for this at #11190. do you mind taking a look?

PR #11170, opened 2016-09-21
client: fix segment fault in Client::_invalidate_kernel_dcache().

ukernel:
when umounting, root can be NULL

Fixes: http://tracker.ceph.com/issues/17253

### DIFF
--- a/src/client/Client.cc
+++ b/src/client/Client.cc
@@ -3938,6 +3938,8 @@ public:
 
 void Client::_invalidate_kernel_dcache()
 {
+  if (unmounting)
+    return;
   if (can_invalidate_dentries && dentry_invalidate_cb && root->dir) {
     for (ceph::unordered_map<string, Dentry*>::iterator p = root->dir->dentries.begin();
 	 p != root->dir->dentries.end();


jcsp:
I'm seeing an instance of ceph-fuse failing to terminate when running with this patch:
http://pulpito.ceph.com/jspray-2016-09-21_18:37:24-fs-wip-jcsp-greenish-distro-basic-smithi/429512/

ukernel:
<pre>
2016-09-21 22:20:38.765059 7fe7baf92f00  1 client.4139 dump_inode: DISCONNECTED inode 1000000dd1c #1000000dd1c ref 11000000dd1c.head(faked_ino=0 ref=1 ll_ref=0 cap_refs={1024=0} open={} mode=40755 size=0/0 mtime=2016-09-21 22:00:40.349543 caps=-(0=pAsLsXsFsx) 0x7fe7cb640000)
</pre>

this hang was caused by inode refernece leak, it's unlikely caused by this RP

jcsp:
OK, now I've seen the same failure in a run without this patch: http://pulpito.ceph.com/jspray-2016-09-23_10:14:18-fs-wip-jcsp-greenish-distro-basic-mira/432687/

PR #11175, opened 2016-09-21
rbd: set rbd_default_features config option as a string

asheplyakov:
Now one can use human readable feature names instead of grepping the code for 
the magic numbers

### DIFF
--- a/src/common/config.cc
+++ b/src/common/config.cc
@@ -36,6 +36,10 @@
 #include <sys/stat.h>
 #include <sys/types.h>
 
+#include <utility>
+#include <boost/type_traits.hpp>
+#include <boost/utility/enable_if.hpp>
+
 /* Don't use standard Ceph logging in this file.
  * We can't use logging until it's initialized, and a lot of the necessary
  * initialization happens here.
@@ -60,32 +64,6 @@ const char *CEPH_CONF_FILE_DEFAULT = "$data_dir/config, /etc/ceph/$cluster.conf,
 
 #define _STR(x) #x
 #define STRINGIFY(x) _STR(x)
-
-
-void *config_option::conf_ptr(md_config_t *conf) const
-{
-  void *v = (void*)(((char*)conf) + md_conf_off);
-  return v;
-}
-
-const void *config_option::conf_ptr(const md_config_t *conf) const
-{
-  const void *v = (const void*)(((const char*)conf) + md_conf_off);
-  return v;
-}
-
-struct config_option config_optionsp[] = {
-#define OPTION(name, type, def_val) \
-       { STRINGIFY(name), type, offsetof(struct md_config_t, name) },
-#define SUBSYS(name, log, gather)
-#define DEFAULT_SUBSYS(log, gather)
-#include "common/config_opts.h"
-#undef OPTION
-#undef SUBSYS
-#undef DEFAULT_SUBSYS
-};
-
-const int NUM_CONFIG_OPTIONS = sizeof(config_optionsp) / sizeof(config_option);
 
 int ceph_resolve_file_search(const std::string& filename_list,
 			     std::string& result)
@@ -123,6 +101,7 @@ md_config_t::md_config_t()
 #define OPTION_OPT_U64(name, def_val) name(((uint64_t)1) * def_val),
 #define OPTION_OPT_UUID(name, def_val) name(def_val),
 #define OPTION(name, type, def_val) OPTION_##type(name, def_val)
+#define SAFE_OPTION(name, type, def_val) OPTION(name, type, def_val)
 #define SUBSYS(name, log, gather)
 #define DEFAULT_SUBSYS(log, gather)
 #include "common/config_opts.h"
@@ -137,10 +116,28 @@ md_config_t::md_config_t()
 #undef OPTION_OPT_U64
 #undef OPTION_OPT_UUID
 #undef OPTION
+#undef SAFE_OPTION
 #undef SUBSYS
 #undef DEFAULT_SUBSYS
   lock("md_config_t", true, false)
 {
+  static const std::vector<md_config_t::config_option> s_config_options = {
+#define OPTION4(name, type, def_val, safe) \
+        config_option{ STRINGIFY(name), type, &md_config_t::name, safe },
+#define OPTION(name, type, def_val) OPTION4(name, type, def_val, false)
+#define SAFE_OPTION(name, type, def_val) OPTION4(name, type, def_val, true)
+#define SUBSYS(name, log, gather)
+#define DEFAULT_SUBSYS(log, gather)
+#include "common/config_opts.h"
+#undef OPTION
+#undef SAFE_OPTION
+#undef SUBSYS
+#undef DEFAULT_SUBSYS
+  };
+  static std::shared_ptr<decltype(s_config_options)>
+    s_tbl(new std::vector<md_config_t::config_option>(std::move(s_config_options)));
+  config_options = s_tbl;
+
   init_subsys();
 }
 
@@ -151,8 +148,10 @@ void md_config_t::init_subsys()
 #define DEFAULT_SUBSYS(log, gather) \
   subsys.add(ceph_subsys_, "none", log, gather);
 #define OPTION(a, b, c)
+#define SAFE_OPTION(a, b, c)
 #include "common/config_opts.h"
 #undef OPTION
+#undef SAFE_OPTION
 #undef SUBSYS
 #undef DEFAULT_SUBSYS
 }
@@ -225,7 +224,7 @@ int md_config_t::parse_config_files(const char *conf_files,
     string &s = *p;
     if (s.find("$data_dir") != string::npos) {
       if (data_dir_option.length()) {
-	list<config_option*> stack;
+	list<config_option const *> stack;
 	expand_meta(s, NULL, stack, warnings);
 	p++;
       } else {
@@ -279,12 +278,11 @@ int md_config_t::parse_config_files_impl(const std::list<std::string> &conf_file
 
   std::vector <std::string> my_sections;
   _get_my_sections(my_sections);
-  for (int i = 0; i < NUM_CONFIG_OPTIONS; i++) {
-    config_option *opt = &config_optionsp[i];
+  for (auto& opt: *config_options) {
     std::string val;
-    int ret = _get_val_from_conf_file(my_sections, opt->name, val, false);
+    int ret = _get_val_from_conf_file(my_sections, opt.name, val, false);
     if (ret == 0) {
-      set_val_impl(val.c_str(), opt);
+      set_val_impl(val.c_str(), &opt);
     }
   }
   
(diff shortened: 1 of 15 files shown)


src/common/config.cc
```diff
@@ -790,42 +803,60 @@ int md_config_t::_get_val(const char *key, char **buf, int len) const
     const config_option *opt = &config_optionsp[i];
     if (strcmp(opt->name, k.c_str()))
       continue;
-
-    ostringstream oss;
+ 
     switch (opt->type) {
       case OPT_INT:
-        oss << *(int*)opt->conf_ptr(this);
+        return std::move(config_value_t(*(int*)opt->conf_ptr(this)));
```
dillaman: Minor: return value optimization handles this automatically -- no need for extra "std::move"

dillaman:
@asheplyakov Couple things:
1) Can you separate the config_ops changes and rbd/librbd changes into two separate commits within this PR?
2) It would be nice to define "safe" config variables within a private section so that you are guaranteed that the only way to access them is through ```get_val```.

asheplyakov:
@dillaman 
> 1) Can you separate the config_ops changes and rbd/librbd changes into two separate commits within this PR?

Done

> 2) It would be nice to define "safe" config variables within a private section so that you are guaranteed that the only way to access them is through get_val.

This breaks `offsetof` magic of config_option::conf_ptr. I'm afraid enforcing "access only via get_val" rule requires major refactoring of config related classes (and their users/callers)

asheplyakov:
> >    2) It would be nice to define "safe" config variables within a private section so that you are guaranteed that the only way to access them is through get_val.

> This breaks offsetof magic of config_option::conf_ptr. I'm afraid enforcing "access only via get_val" rule requires major refactoring of config related classes (and their users/callers)

Actually enforcing the access rules is not *that* complicated (although it requires some boring refactoring of config.{cc,h}). The idea is to use pointers to members instead of C-style offsets (and static visitors instead of switch/case). I'll post an updated patch in 3 -- 5 days.

asheplyakov:
@dillaman 
> 2) It would be nice to define "safe" config variables within a private section so that you are guaranteed that the only way to access them is through get_val

Done. That requires a bit of refactoring, though.

src/common/config.h
```diff
@@ -136,6 +212,8 @@ struct md_config_t {
   // No metavariables will be returned (they will have already been expanded)
   int get_val(const char *key, char **buf, int len) const;
   int _get_val(const char *key, char **buf, int len) const;
+  config_value_t get_val(const char *key) const;
```
dillaman: It might be nice to just template this method on the expected value type -- that would eliminate the need for each caller to have to cast the variant and check for failure.

src/common/config.h
```diff
@@ -136,6 +212,8 @@ struct md_config_t {
...
+  config_value_t _get_val(const char *key) const;
```
dillaman: Minor: I assume this can just be moved to the private section

src/common/config.cc
```diff
@@ -61,32 +65,20 @@ const char *CEPH_CONF_FILE_DEFAULT = "$data_dir/config, /etc/ceph/$cluster.conf,
...
-const void *config_option::conf_ptr(const md_config_t *conf) const
-{
-  const void *v = (const void*)(((const char*)conf) + md_conf_off);
-  return v;
-}
-
-struct config_option config_optionsp[] = {
-#define OPTION(name, type, def_val) \
-       { STRINGIFY(name), type, offsetof(struct md_config_t, name) },
+const std::map<std::string, md_config_t::config_option> md_config_t::config_options = {
```
asheplyakov: Done

src/common/config.cc
```diff
@@ -377,8 +373,8 @@ void md_config_t::_show_config(std::ostream *out, Formatter *f)
       f->dump_string(debug_name.c_str(), ss.str());
     }
   }
-  for (int i = 0; i < NUM_CONFIG_OPTIONS; i++) {
-    config_option *opt = config_optionsp + i;
+  for (auto& named_opt: config_options) {
+    config_option const *opt = &(named_opt.second);
```
asheplyakov: Done

src/common/config.cc
```diff
@@ -712,6 +708,25 @@ void md_config_t::set_val_or_die(const char *key, const char *val)
...
+    }
+    template<typename T,
+             typename boost::enable_if_c<!boost::is_integral<T>::value, int>::type = 0>
+    bool operator()(const T md_config_t::* /* member_ptr */) const {
+        return false;
+    }
+};
+
+bool md_config_t::config_option::is_safe() const {
+  return safe ||
```
asheplyakov: I don't quite get what's exactly wrong here. Could you elaborate, please?

src/common/config.h
asheplyakov: OK, I've made _get_val private

src/common/config.h
asheplyakov: I've added a template method:

template<typename T> T get_val(const char *key) const;

However checking for an error is still necessary (say, if there's no such an option, or it has a different type)

src/common/config.cc
dillaman: What I am trying to say is that it makes the code harder to read when you insert a comment line in the middle of a statement:

```
return xyz ||

  // now let's talk about things 
  abc ||

 // some more things to say
 tuv;
```

src/common/config.h
```diff
@@ -259,37 +345,51 @@ struct md_config_t {
...
+struct get_typed_value_visitor : public boost::static_visitor<T> {
+  template<typename U,
+    typename boost::enable_if<boost::is_same<T, U>, int>::type = 0>
+      T operator()(U & val) {
+	return std::move(val);
+      }
+  template<typename U,
+    typename boost::enable_if_c<!boost::is_same<T, U>::value, int>::type = 0>
+      T operator()(U &val) {
+	throw boost::bad_get();
```
dillaman: I would just "assert" here since it would include the full backtrace of the offending ```get_val```

src/common/config.h
dillaman: Since we can't catch such mistakes at compile time, I'd rather assert the failure instead of having sections of code locally re-implement default values should the ```get_val``` fail. If I was attempting to retrieve the "rbd_default_features" and it was misspelled in code or renamed in the config, I'd want my code to fail early.

src/common/config.cc
```diff
@@ -786,45 +811,30 @@ int md_config_t::_get_val(const char *key, char **buf, int len) const
...
+  get_value_generic_visitor gvv(this);
+  return boost::apply_visitor(gvv, opt_it->second.md_member_ptr);
+}
 
+int md_config_t::_get_val(const char *key, char **buf, int len) const
+{
+  assert(lock.is_locked());
+
+  if (!key)
+    return -EINVAL;
```
asheplyakov: returning -EINVAL here is wrong since it looks like as a valid integer valued option

src/common/config.cc
```diff
@@ -786,45 +811,30 @@ int md_config_t::_get_val(const char *key, char **buf, int len) const
   // In key names, leading and trailing whitespace are not significant.
```
asheplyakov: The above -EINVAL is wrong since it looks like a valid integer option from the caller point of the view

src/common/config.h
asheplyakov: Done

src/common/config.cc
asheplyakov: Got it. Does the latest version look OK?

asheplyakov:
@dillaman Could you please point out problems which should be fixed before this patchset can be merged?

src/common/config.cc
```diff
@@ -712,6 +708,24 @@ void md_config_t::set_val_or_die(const char *key, const char *val)
...
+             typename boost::enable_if<boost::is_integral<T>, int>::type = 0>
+    bool operator()(const T md_config_t::* /* member_ptr */) const {
+        return true;
+    }
+    template<typename T,
+             typename boost::enable_if_c<!boost::is_integral<T>::value, int>::type = 0>
+    bool operator()(const T md_config_t::* /* member_ptr */) const {
+        return false;
+    }
+};
```
liewegas: nit: indentation levels should be 2 spaces

src/include/rbd/features.h
```diff
@@ -10,6 +10,20 @@
...
+                                         RBD_FEATURE_DEEP_FLATTEN)
+
+#define RBD_FEATURE_NAME_LAYERING        "layering"
+#define RBD_FEATURE_NAME_STRIPINGV2      "striping"
+#define RBD_FEATURE_NAME_EXCLUSIVE_LOCK  "exclusive-lock"
+#define RBD_FEATURE_NAME_OBJECT_MAP      "object-map"
+#define RBD_FEATURE_NAME_FAST_DIFF       "fast-diff"
+#define RBD_FEATURE_NAME_DEEP_FLATTEN    "deep-flatten"
+#define RBD_FEATURE_NAME_JOURNALING      "journaling"
+
```
dillaman: Can you add support for the new ```RBD_FEATURE_DATA_POOL``` feature bit

asheplyakov: Done

src/common/config.cc
asheplyakov: I've re-indented the code a bit. Does it look OK now?

dillaman:
lgtm -- the ```run-cli-tests``` tests need to be fixed, however

asheplyakov:
> the run-cli-tests tests need to be fixed, however

Apparently the test expects config variables to be printed in a certain order (the same as they declared in config_opts.h). Since users (both human and monitoring tools) might have similar expectations I've adjusted the code accordingly.

dillaman:
retest this please

dillaman:
@asheplyakov I haven't seen the 'cephtool-test-mon.sh' test failure before, so I am not sure if it's related to this PR or not. I can repeat it locally as well with this PR so I suspect some change in this PR is making that test unhappy.

asheplyakov:
@dillaman I've reproduced test_mon_injectargs failure locally and am working on fixing it

asheplyakov:
@dillaman The reason of the failure is floating point (both float and double) options which became "unsafe". I've adjusted config_option::is_safe(), and test_mon_injectargs does not fail any more

dillaman:
retest this please

dillaman:
@asheplyakov unittest_compression is also repeatedly failing -- I haven't seen it fail in other PRs today.

jcsp:
Seems fine to me (removing cephfs label)

asheplyakov:
@dillaman 
> unittest_compression is also repeatedly failing

It has been failing due to static deinitialization order fiasco, that is, some global object (presumable g_ceph_context) destructor has been trying to use the static object md_config_t::config_options after the latter had been destroyed.

Debugging segfaults which happen after main() has returned is always a fun.

asheplyakov:
Random tests still keep failing (this time it's `osd-fast-mark-down.sh`), however I'm unable to reproduce those failures (neither with, nor without this patchset). Any ideas how to find out what could cause [this](https://jenkins.ceph.com/job/ceph-pull-requests/12776/consoleFull#-13568805540526d21-3511-427d-909c-dd086c0d1034)?

dillaman:
retest this please

dillaman:
Associated teuthology runs:
http://pulpito.ceph.com/jdillaman-2016-10-18_21:31:07-rbd-wip-jd-testing---basic-smithi/
http://pulpito.ceph.com/jdillaman-2016-10-19_13:08:35-rgw-wip-jd-testing-distro-basic-smithi/
http://pulpito.ceph.com/jdillaman-2016-10-19_13:09:57-rados-wip-jd-testing-distro-basic-smithi/
http://pulpito.ceph.com/jdillaman-2016-10-19_13:11:45-fs-wip-jd-testing-distro-basic-smithi/

PR #11177, opened 2016-09-21
build/ops: strip virtualenv

dachary:
http://tracker.ceph.com/issues/15260


install-deps.sh
```diff
@@ -167,6 +167,8 @@ function activate_virtualenv() {
     . $env_dir/bin/activate
 }
 
+test "$1" = "skip-virtualenv" && exit 0
```
ktdreyer: Can we make this the default (not an option?)

dachary: If it is made the default all environment that rely on install-deps.sh to run tests will have to be modified. If we don't make it the default all environments that fail to run because install-deps.sh will have to add this option. Unless I'm mistaken there only are two environments that motivate this change : the gitbuilders and the jenkins builders and both can be modified to add this argument. By contrast there is an unknown number of environment (most of which are out of our control) that depend on the current behavior.

src/ceph-detect-init/CMakeLists.txt
```diff
@@ -10,3 +10,10 @@ add_dependencies(tests ceph-detect-init)
 
 include(Distutils)
 distutils_install_module(ceph_detect_init)
+
+add_custom_target(ceph-detect-init-flake8
+  COMMAND
+  flake8 ceph_detect_init tests
+  WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/src/ceph-detect-init
+  COMMENT "syntax check")
+add_dependencies(ceph_detect_init ceph-detect-init-flake8)
```
dachary: @tchaikov I'd like to express that `ceph-detect-init-flake8` should run on make all. Any idea on how this should be done ?

install-deps.sh
ktdreyer: Why do we need a virtualenv to run the tests?

dachary: The tests are run via tox which needs virtualenv. The other reason is to be able to run the tests without network access. a) run install-deps.sh with network access to pull dependencies (packages + python) and do nothing else, b) run make check with no network access

ktdreyer: Tox is nice for virtualenv management, but the tests could run outside of tox during "make check".

Using system packages during "make check" would work without network access in the scenario you describe. It would be simpler because the process would not have to rely on PyPI's availability at all.

dachary: I will remove pip from install-deps.sh and rely on packages instead.

dachary:
@ktdreyer does that look good to you ?

src/ceph-detect-init/CMakeLists.txt
```diff
@@ -1,12 +1,9 @@
-set(CEPH_DETECT_INIT_VIRTUALENV ${CEPH_BUILD_VIRTUALENV}/ceph-detect-init-virtualenv)
+include(Distutils)
+distutils_install_module(ceph_detect_init)
 
-add_custom_target(ceph-detect-init
+add_custom_target(ceph-detect-init-flake8
```
tchaikov: can we add it as a test instead?

src/ceph-detect-init/CMakeLists.txt
```diff
@@ -1,12 +1,9 @@
...
   COMMAND
-  ${CMAKE_SOURCE_DIR}/src/tools/setup-virtualenv.sh ${CEPH_DETECT_INIT_VIRTUALENV} &&
-  ${CEPH_DETECT_INIT_VIRTUALENV}/bin/pip install --no-index --use-wheel --find-links=file:${CMAKE_SOURCE_DIR}/src/ceph-detect-init/wheelhouse -e .
+  flake8 ceph_detect_init tests
```
tchaikov: might want to have
```cmake
find_program(FLAKE8_EXECUTABLE flake8)
```
in `src/CMakeLists.txt`.

src/ceph-detect-init/CMakeLists.txt
```diff
@@ -1,12 +1,9 @@
...
   WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/src/ceph-detect-init
-  COMMENT "ceph-detect-init is being created")
-add_dependencies(tests ceph-detect-init)
-
-include(Distutils)
-distutils_install_module(ceph_detect_init)
+  COMMENT "syntax check")
+add_dependencies(ceph_detect_init ceph-detect-init-flake8)
```
tchaikov: we don't have a target named `ceph_detect_init` yet. `ceph_detect_init` is used to create another target named `ceph_detect_init-clone`  for copying the source files into the build directory.

tchaikov: >  I'd like to express that `ceph-detect-init-flake8` should run on make all. Any idea on how this should be done ?

see https://cmake.org/cmake/help/v3.0/command/add_custom_target.html, we can add `ALL` after the name of the target. i.e.
```cmake
add_custom_target(ceph-detect-init-flake8 ALL
...
)
```

but i'd prefer adding this check as a test.

src/ceph-detect-init/CMakeLists.txt
```diff
@@ -1,12 +1,9 @@
...
+include(Distutils)
+distutils_install_module(ceph_detect_init)
 
-add_custom_target(ceph-detect-init
+add_custom_target(ceph-detect-init-flake8
   COMMAND
-  ${CMAKE_SOURCE_DIR}/src/tools/setup-virtualenv.sh ${CEPH_DETECT_INIT_VIRTUALENV} &&
-  ${CEPH_DETECT_INIT_VIRTUALENV}/bin/pip install --no-index --use-wheel --find-links=file:${CMAKE_SOURCE_DIR}/src/ceph-detect-init/wheelhouse -e .
+  flake8 ceph_detect_init tests
   WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/src/ceph-detect-init
```
tchaikov: nit, could just use `${CMAKE_CURRENT_SOURCE_DIR}` instead.

dachary: thanks !

src/ceph-detect-init/CMakeLists.txt
dachary: ack

src/ceph-detect-init/CMakeLists.txt
dachary: ok

ktdreyer:
If we use `install-deps.sh` to install "make check" dependencies, maybe "python-tox" should move to that shell script, rather than inserting it into the packaging? The package build operation itself doesn't need tox, because it doesn't run "make check" (yet!). Downstream in the distros we do not need to have tox in the buildroots in order to simply build Ceph, if I understand this change correctly. What do you think?

dachary:
@ktdreyer install-deps.sh is not responsible for installing test dependencies. If we are going to change that, it should be done in a separate task and remove all packages that are only required for testing.

ktdreyer:
@dachary in that case, I'm not following why we have to `BuildRequires: python-tox`, in ceph.spec? With the changes in this PR, what particular build step uses tox?

dachary:
@ktdreyer ceph.spec and debian/control build requirements cover the need of both: build and testing. If they are to cover only the need for building, a number of packages can be removed.

ktdreyer:
It is a bummer that we're bloating the package build dependency lists with things that aren't needed to build the packages. That slows down the package build times and complicates life for downstreams.

Thanks for explaining it to me. I think it should be a separate change to fix this (eg split out those deps, or just finally add `make check` during the package build).

alfredodeza:
@dachary can you make a branch in the ceph project with these changes? That way we can check from that branch how/if it builds

alfredodeza:
@dachary this needs a rebase. Mind also creating a branch in the ceph repo so we can try this out with the actual process we use to build/release? Thanks!

dmick:
Did this issue go away for some other reason?  Why is the PR closed and the bug unassigned?

dachary:
@tchaikov I confess I have no clue how to make that work with cmake :-(

tchaikov:
@dachary no worries, let me give it a shot!

dachary:
I'm closing this so that someone knowledgeable can work on it. I don't have the necessary skills.

PR #11180, opened 2016-09-21
messages/MForward: fix encoding features

liewegas:
We were encoding the message with the sending client's
features, which makes no sense: we need to encode with
the recipient's features so that it can decode the
message.

The simplest way to fix this is to rip out the bizarre
msg_bl handling code and simply keep a decoded Message
reference, and encode it when we send.

Fixes: http://tracker.ceph.com/issues/17365

### DIFF
--- a/src/messages/MForward.h
+++ b/src/messages/MForward.h
@@ -22,6 +22,7 @@
 #include "msg/Message.h"
 #include "mon/MonCap.h"
 #include "include/encoding.h"
+#include "include/stringify.h"
 
 struct MForward : public Message {
   uint64_t tid;
@@ -29,9 +30,10 @@ struct MForward : public Message {
   MonCap client_caps;
   uint64_t con_features;
   EntityName entity_name;
-  PaxosServiceMessage *msg;   // incoming message
-  bufferlist msg_bl;          // outgoing message
+  PaxosServiceMessage *msg;   // incoming or outgoing message
 
+  string msg_desc;  // for operator<< only
+  
   static const int HEAD_VERSION = 3;
   static const int COMPAT_VERSION = 1;
 
@@ -44,7 +46,9 @@ struct MForward : public Message {
     client = m->get_source_inst();
     client_caps = m->get_session()->caps;
     con_features = feat;
-    set_message(m, feat);
+    // we may need to reencode for the target mon
+    msg->clear_payload();
+    msg = (PaxosServiceMessage*)m->get();
   }
   MForward(uint64_t t, PaxosServiceMessage *m, uint64_t feat,
            const MonCap& caps) :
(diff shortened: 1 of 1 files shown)


liewegas:
@gregsfortytwo @jecluis Do you mind looking at this?  I'm a little surprised this hasn't bitten us before.  Am I missing something?

gregsfortytwo:
Oh, and indeed that's how the existing implementation seems to work. I guess if both the client and the peon monitor have newer features+code than the leader, we could get in trouble. So we might want to do a logical AND of the client's features and the quorum features for encoding, but that's the only issue I see with current implementation.

liewegas:
The memory management commits seem to be 
2731b19aec3214d22475ab39e193a71d2aa727ca which originally introduced this, 
and f2e240ab8ec2d6a47df1dce9c733c26581221bee, which changed the reference 
counting.  This is where the bug was introduced (encoding with client--not 
target--features).

There was another fix ac10aa5d1e20c230f27ee07d9b7c88c11071f681 which 
addressed the target mon knowing the original client's features.  This is 
what's used to populate the synthetic Connection on the receiving end.  
AFAICS there isn't any dependency on how the payload message is 
encoded/decoded as it passes over the wire between monitors...

I hit this with the upgrade test on wip-addrvec.  The receiving mon 
couldn't decode the payload message.  There is something else going on 
there, but I think it's independent of this bug.

liewegas:
Yeah, the scenario you describe is exactly what's going on (leader mon is older than peon and client).  So the question is, should we AND the features or just use the target?  I suppose there's no harm in ANDing them, and it's possible there's some subtle dependency.  I'll change that.

Unrelated weirdness, though: in this particular case the message came from loopback, and for those virtual Connections we set features to the feature mask passed into the messenger ctor... which is 0 in ceph_mon.cc (and most other places).  The defautl arg is 0, and everyone passes 0, except one caller that passes CEPH_FEATURES_ALL.  And the value is only used by the messenger for this loopback connection.  I'm going to do a separate PR to just rip that all out...

liewegas:
@gregsfortytwo It's doing the & now on the feature set.  And passed the jewel-x test that it was previously failing due to the encoding disparity (http://pulpito.ceph.com/yuriw-2016-09-27_21:00:16-upgrade:jewel-x-wip-mon-forward_2016_09_27-distro-basic-smithi/)

src/messages/MForward.h
```diff
@@ -63,18 +67,14 @@ struct MForward : public Message {
...
-  }
-
 public:
   void encode_payload(uint64_t features) {
     ::encode(tid, payload);
     ::encode(client, payload, features);
     ::encode(client_caps, payload, features);
-    payload.append(msg_bl);
+    // encode client message with intersection of target and source
+    // features.  this probably doesn't matter, but might.
```
gregsfortytwo: This is not a helpful comment. Say when it *does* matter rather than that it probably doesn't.

liewegas:
@gregsfortytwo updated

PR #11181, opened 2016-09-21
rgw_file:  pre-assign times

mattbenjamin:
Set unix timestamps based on RGW values for creation, modification
time for buckets, objects in stat requests.  Regard any saved
value of these in unix attributes as an overlay.

Fixes: http://tracker.ceph.com/issues/17367


ktdreyer:
In the past, we've removed references to downstream bug trackers from upstream commit messages. I don't care strongly about this either way, myself.

mattbenjamin:
I can split it out.

PR #11184, opened 2016-09-21
mon: wait 10m (not 5m) before marking down OSDs out

liewegas:
5m isn't quite enough time for many boxes to reboot.  10m will do
a better job of masking transient crashes without having much
affect on time to repair.


gregsfortytwo:
LGTM

athanatos:
lgtm

dmick:
'effect'.  but yeah.

PR #11187, opened 2016-09-22
lttng: remove the source files created by lttng-gen-tp

tchaikov:
they will be created at build time in the build directory.

### DIFF
--- a/src/tracing/librados.c
+++ b/src/tracing/librados.c
@@ -1,6 +1,0 @@
-
-#define TRACEPOINT_CREATE_PROBES
-/*
- * The header containing our TRACEPOINT_EVENTs.
- */
-#include "librados.h"
--- a/src/tracing/librbd.c
+++ b/src/tracing/librbd.c
@@ -1,6 +1,0 @@
-
-#define TRACEPOINT_CREATE_PROBES
-/*
- * The header containing our TRACEPOINT_EVENTs.
- */
-#include "librbd.h"
--- a/src/tracing/objectstore.c
+++ b/src/tracing/objectstore.c
@@ -1,6 +1,0 @@
-
-#define TRACEPOINT_CREATE_PROBES
-/*
- * The header containing our TRACEPOINT_EVENTs.
- */
-#include "objectstore.h"
--- a/src/tracing/oprequest.c
+++ b/src/tracing/oprequest.c
@@ -1,6 +1,0 @@
-
-#define TRACEPOINT_CREATE_PROBES
-/*
- * The header containing our TRACEPOINT_EVENTs.
- */
-#include "oprequest.h"
--- a/src/tracing/osd.c
+++ b/src/tracing/osd.c
@@ -1,6 +1,0 @@
-
-#define TRACEPOINT_CREATE_PROBES
-/*
- * The header containing our TRACEPOINT_EVENTs.
- */
-#include "osd.h"
--- a/src/tracing/pg.c
+++ b/src/tracing/pg.c
@@ -1,6 +1,0 @@
-
-#define TRACEPOINT_CREATE_PROBES
-/*
- * The header containing our TRACEPOINT_EVENTs.
- */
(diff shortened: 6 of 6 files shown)


tchaikov:
@dillaman i re-read http://lttng.org/docs/#doc-dynamic-linking, and tested lttng tracing locally. seems we misunderstood the doc, so i am reopening #11161 in this PR.

1. build `librados` and `rados-tp` with `-DWITH_LTTNG=ON`
1. start a vstart cluster
1. lttng create my-session
1. lttng enable-event -u -a
1. lttng start
1. LD_PRELOAD="/var/ceph/ceph/build/lib/librados_tp.so"  ./bin/rados put test-obj1 /etc/hosts -p rbd
1. lttng stop
1. lttng view| grep 'librados:rados_' | head
```
[11:07:05.020243985] (+0.018389201) gen8 librados:rados_create_with_context_enter: { cpu_id = 1 }, { cct = 0x5596D12BC430 }
[11:07:05.020359226] (+0.000115241) gen8 librados:rados_create_with_context_exit: { cpu_id = 1 }, { retval = 0, cluster = 0x5596D12CA080 }
[11:07:05.026989314] (+0.006630088) gen8 librados:rados_ioctx_create_enter: { cpu_id = 1 }, { cluster = 0x5596D12CA080, name = "rbd" }
[11:07:05.028028008] (+0.001038694) gen8 librados:rados_ioctx_create_exit: { cpu_id = 1 }, { retval = 0, ioctx = 0x5596D1367BE0 }
[11:10:01.917293405] (+0.011041608) gen8 librados:rados_create_with_context_enter: { cpu_id = 4 }, { cct = 0x556A36F05430 }
[11:10:01.917415324] (+0.000121919) gen8 librados:rados_create_with_context_exit: { cpu_id = 4 }, { retval = 0, cluster = 0x556A36F13080 }
[11:10:01.924791634] (+0.007376310) gen8 librados:rados_ioctx_create_enter: { cpu_id = 4 }, { cluster = 0x556A36F13080, name = "rbd" }
[11:10:01.926059204] (+0.001267570) gen8 librados:rados_ioctx_create_exit: { cpu_id = 4 }, { retval = 0, ioctx = 0x556A36FB0BE0 }
[11:11:04.011121571] (+0.020887079) gen8 librados:rados_create_with_context_enter: { cpu_id = 4 }, { cct = 0x561BB22B3440 }
[11:11:04.011243093] (+0.000121522) gen8 librados:rados_create_with_context_exit: { cpu_id = 4 }, { retval = 0, cluster = 0x561BB22C1090 }
```

so the tracing works, and as per the example in http://lttng.org/docs/#doc-dynamic-linking, the `tp.c` is the same as the one used in "Static linking"
> The tracepoint provider object file used to create the shared library is built like it is using the static library method, only with the `-fpic` option added:
> ```gcc -c -fpic -I. tp.c```


> http://lttng.org/docs/#doc-dynamic-linking -- as per the docs, the ```TRACEPOINT_DEFINE``` macro should only be defined in the application (librbd/librados/etc), not the new shared library.

this is not right, the shared library *can* have `TRACEPOINT_DEFINE` defined. yes, i also tested the build without it defined in the shared library. it works also.

@dillaman

dillaman:
@tchaikov We don't use LD_PRELOAD to load the libraries -- we use the config values "XYZ_tracing" to dlopen the library. If you think about it, ```TRACEPOINT_DEFINE``` tells the header file to define all the necessary functions needed to interact with the probes (if they are available). So by having ```TRACEPOINT_DEFINE``` in both the application (librbd.so) and the dynamically loaded tracing library (librbd_tp.so), you are pulling in the same functions twice.  It may work -- but you shouldn't be doing it.

dillaman:
@tchaikov Note: there is this issue w/ the XYZ_tracing config option: #10387

tchaikov:
@dillaman thanks! fixed at #11196

PR #11189, opened 2016-09-22
[RFC]bluestore: separate kv_sync_thread to two parts.

majianpeng:
To make performance better.

### DIFF
--- a/src/os/bluestore/BlueStore.cc
+++ b/src/os/bluestore/BlueStore.cc
@@ -2501,6 +2501,7 @@ BlueStore::BlueStore(CephContext *cct, const string& path)
 	     &wal_tp),
     m_finisher_num(1),
     kv_sync_thread(this),
+    kv_sync_thread1(this),
     kv_stop(false),
     logger(NULL),
     debug_read_error_lock("BlueStore::debug_read_error_lock"),
@@ -4015,6 +4016,7 @@ int BlueStore::mount()
   }
   wal_tp.start();
   kv_sync_thread.create("bstore_kv_sync");
+  kv_sync_thread1.create("bstore_kv_sync1");
 
   r = _wal_replay();
   if (r < 0)
@@ -6577,6 +6579,48 @@ void BlueStore::_txc_release_alloc(TransContext *txc)
   txc->released.clear();
 }
 
+void BlueStore::_kv_sync_thread1()
+{
+  dout(10) << __func__ << " start" << dendl;
+  std::unique_lock<std::mutex> l(kv_lock1);
+  while (true) {
+    assert(kv_committing2.empty());
+    assert(wal_cleaning2.empty());
+    if (kv_committing1.empty() && wal_cleaning1.empty()) {
+      if (kv_stop)
+	break;
+      dout(20) << __func__ << " sleep" << dendl;
+      kv_sync_cond.notify_all();
+      kv_cond1.wait(l);
+      dout(20) << __func__ << " wake" << dendl;
+    } else {
+      kv_committing2.swap(kv_committing1);
+      wal_cleaning2.swap(wal_cleaning1);
+      kv_cond1.notify_one();
+      utime_t start = ceph_clock_now(NULL);
+      l.unlock();
+
+     while (!kv_committing2.empty()) {
+	TransContext *txc = kv_committing2.front();
+	_txc_state_proc(txc);
+	kv_committing2.pop_front();
+      }
+      while (!wal_cleaning2.empty()) {
+	TransContext *txc = wal_cleaning2.front();
+	_txc_state_proc(txc);
+	wal_cleaning2.pop_front();
+      }
+
+      // this is as good a place as any ...
+      _reap_collections();
+
+      l.lock();
+    }
+  }
+  dout(10) << __func__ << " finish" << dendl;
+}
+
+
 void BlueStore::_kv_sync_thread()
 {
   dout(10) << __func__ << " start" << dendl;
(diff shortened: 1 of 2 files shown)


majianpeng:
For my 4K-RW 300s test
A: if there is compaction of rocksdb in this test; The iops increase from 148650 to 165808(about 11%)
B: If there is no compaction of rocksdb, the IOPS increase from 156441 to 186923(about 19%).
@markhpc . Please review this.

yuyuyu101:
yes, it's a good tradeoff.... this is also my original thought about adding async submit support for rocksdb while we have a lot of time ...

but your iops test here is refer to rbd case? if we can benefit from pipeline design, why we can't benefit much from multi kv threads? I think they're similar

majianpeng:
For multi kv_sync_thread, we implemented and test.But the IOPS don't increase much.  We change the ratio of submit_transaction/submit_transaction_sync. So the performance don't like as expect.

yuyuyu101:
I got your meaning. It looks multi kv thread will decrease ops per transaction, so it will make rocksdb call fsync more frequency.

varadakari:
We are introducing one more thread switch here, instead can we finish what _kv_sync_thread(new one) does in aio_cb thread and defer the rest to _kv_sync_thread1(). That should give us some more cpu.

tanghaodong25:
@varadakari we prefer batch transaction submission sync, cuz it can reduce fsync racing in bluefs. There is something to try if using aio_cb:

1. do submit_transaction and wal cleaning work in multithreaded _aio_thread
2. just do submit_transaction_sync in _kv_sync_thread
3. do other works in _kv_sync_thread1

But I'm not sure multithreaded _aio_thread can bring some benefit.

varadakari:
we have to batch processing for rocksdb to reduce fsyncs, i am not against that. How much of latency the new thread switch and lock are introducing. I get your point for partitioning of processing, i also have similar set of changes. What i mean here, let kv_sync_thread do what kv_sync_thread1 is  doing now and let other work be done on aio_cb itself. In order to do this, we can increase the number of callback threads. Just an idea, have to determine will that increase any performance. I haven't done any perf runs yet!.

markhpc:
very nice!  I'm working on finding a perf regression in the wip-denc branch.  Once I'm done I will try to look this over.  Thanks!

yuyuyu101:
I rethink the whole thing, and suggest we separated the current kv_sync_thread logic into two threads more than this pr. Current kv_sync_thread could only process per TransContext and left submit_transaction_sync into a new submit thread. So we could make less TransContext pending on the queue while thread is submitting. And the submitting thread could ensure all the times is waiting on the wal sync. In the stress benchmark, less TransContext queue latency and less gap between transaction sync

varadakari:
@yuyuyu101 Yeah, that seems a good approach. To wait on wal sync, we need to get some throttling in place for new submitting thread, not to overwhelm the wal syncing. For any reason, wal sync slows down, will have effect on the reads. Need to build that back pressure.

majianpeng:
@yuyuyu101 . Sorry for late response. Are you mean in kv_sync_thread, it only do submit_transaction and in kv_sync_thread1 it do others inlucde submit_transactions_sync & txc_proc_stat.?

liewegas:
I think it's worth rebasing this on top of the current master to see if it is still worthwhile to push the completion work into a separate thread, and/or how this interacts with bluestore_sync_submit_transaction.  What do you think?

src/os/bluestore/BlueStore.cc
```diff
@@ -2195,6 +2195,7 @@ BlueStore::BlueStore(CephContext *cct, const string& path)
 	     &wal_tp),
     m_finisher_num(1),
     kv_sync_thread(this),
+    kv_sync_thread1(this),
```
liewegas: kv_sync_finish_thread

src/os/bluestore/BlueStore.cc
```diff
@@ -3708,6 +3709,7 @@ int BlueStore::mount()
   }
   wal_tp.start();
   kv_sync_thread.create("bstore_kv_sync");
+  kv_sync_thread1.create("bstore_kv_sync1");
```
liewegas: bstore_kv_sync_fin?

src/os/bluestore/BlueStore.cc
```diff
@@ -6265,18 +6309,29 @@ void BlueStore::_kv_sync_thread()
 	_txc_state_proc(txc);
 	wal_cleaning.pop_front();
       }
-
+#endif
```
liewegas: We'd want a config option to control this.  kv_sync_finish_thread (bool)

src/os/bluestore/BlueStore.h
```diff
@@ -1313,11 +1321,14 @@ class BlueStore : public ObjectStore,
...
-  std::mutex kv_lock;
-  std::condition_variable kv_cond, kv_sync_cond;
+  KVSyncThread1 kv_sync_thread1;
+  std::mutex kv_lock, kv_lock1;
+  std::condition_variable kv_cond, kv_sync_cond, kv_cond1;
   bool kv_stop;
   deque<TransContext*> kv_queue, kv_committing;
   deque<TransContext*> wal_cleanup_queue, wal_cleaning;
+  deque<TransContext*> wal_cleaning1, kv_committing1;
+  deque<TransContext*> wal_cleaning2, kv_committing2;
```
liewegas: _finish instead of 1,2 suffix

yuyuyu101:
@majianpeng ping

majianpeng:
@yuyuyu101 . Now i'm doing performance tests with my cluster. Next week I'll firstly do this work.  Sorry  for delay.

liewegas:
see https://github.com/ceph/ceph/pull/14035

PR #11191, opened 2016-09-22
libcephfs: add ceph_fsetattr&&ceph_lchmod&&ceph_lutime

renhwztetecs:
export ceph_fsetattr() && ceph_lchmod() && ceph_lutime()

### DIFF
--- a/src/test/libcephfs/test.cc
+++ b/src/test/libcephfs/test.cc
@@ -1158,6 +1158,7 @@ TEST(LibCephFS, UseUnmounted) {
   EXPECT_EQ(-ENOTCONN, ceph_lremovexattr(cmount, "/path", "name"));
   EXPECT_EQ(-ENOTCONN, ceph_setxattr(cmount, "/path", "name", NULL, 0, 0));
   EXPECT_EQ(-ENOTCONN, ceph_lsetxattr(cmount, "/path", "name", NULL, 0, 0));
+  EXPECT_EQ(-ENOTCONN, ceph_fsetattr(cmount, 0, &st, 0));
   EXPECT_EQ(-ENOTCONN, ceph_chmod(cmount, "/path", 0));
   EXPECT_EQ(-ENOTCONN, ceph_fchmod(cmount, 0, 0));
   EXPECT_EQ(-ENOTCONN, ceph_chown(cmount, "/path", 0, 0));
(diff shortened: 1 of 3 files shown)


jcsp:
Can you comment on why lchmod is needed?  It doesn't seem to part of e.g. Linux's file I/O implementation.

jcsp:
+ Same question about lutime.

renhwztetecs:
@jcsp 
my primitive idea，we can use  lchmod &&lutime of  user mode，not kernel mode.
:-)

jcsp:
Right, but ordinary applications presumably don't use calls like this, since they're not part of the usual Linux stdlib interface.  So why does libcephfs need them?

renhwztetecs:
ignorant and ill-informed, and I get it
thank you

ffilz:
Hmm, would lchmod have allowed setting the mode of a symlink object rather than it's target? nfs-ganesha could actually use such a function, the NFS protocol allows setting attributes of symlink objects and freebsd at least supports lchmod so we could have clients trying it.

renhwztetecs:
@ffilz
thanks for your advices
ping @jcsp

jcsp:
Sorry for the delay.  I'm okay with this now as the fsetattr is at least a normal function on some UNIX variants.

@renhwztetecs please could you add a unit test for this in test/libcephfs/test.cc.

jcsp:
This currently does not compile when merged to master, needs a rebase (change in definition of Client::fsetattr)

renhwztetecs:
@jcsp I will add unit test for fsetattr

renhwztetecs:
@jcsp
I added the unit test for fsetattr, could you take a look?

src/test/libcephfs/test.cc
```diff
@@ -1670,3 +1671,31 @@ TEST(LibCephFS, DirChangeAttr) {
...
+  ceph_unlink(cmount, filename);
+  int fd = ceph_open(cmount, filename, O_RDWR|O_CREAT|O_EXCL, 0666);
+  ASSERT_LT(0, fd);
+  ceph_close(cmount, fd);
+
+  struct ceph_statx stx;
+  uint64_t size = 8388608;
+  stx.stx_size = size;
+
+  ASSERT_EQ(ceph_fsetattr(cmount, fd, &stx, CEPH_SETATTR_SIZE), 0);
```
jcsp: You see to be passing a ceph_statx to a function that expects a `struct stat`?

renhwztetecs: thanks for your reviews
I changed it

renhwztetecs:
@thanks for your reviews
I changed it

renhwztetecs:
@jcsp 
thanks for your reviews
I changed it

src/test/libcephfs/test.cc
```diff
@@ -1158,6 +1158,7 @@ TEST(LibCephFS, UseUnmounted) {
   EXPECT_EQ(-ENOTCONN, ceph_lremovexattr(cmount, "/path", "name"));
   EXPECT_EQ(-ENOTCONN, ceph_setxattr(cmount, "/path", "name", NULL, 0, 0));
   EXPECT_EQ(-ENOTCONN, ceph_lsetxattr(cmount, "/path", "name", NULL, 0, 0));
+  EXPECT_EQ(-ENOTCONN, ceph_fsetattr(cmount, 0, &st, 0);
```
jcsp: Missing closing `)` here.  Please do compile this test and check that it works before updating the PR.

renhwztetecs: I omitted this step, the follow-up I will pay attention to this problem,
Thank you for reminding.

jtlayton:
Ahh, just came into this late. Yeah, I agree that lchmod/lutimes isn't really needed. We have a ceph_setattrx now, which can take an AT_SYMLINK_NOFOLLOW flag. That should allow setting any of those fields on the symlink itself.

renhwztetecs:
@jtlayton thanks
yeah, I have been removed ceph_lchmod&&ceph_lutime

PR #11194, opened 2016-09-22
src/kv/CMakeLists.txt: force rocksdb/include to first include directory

wjwithagen:
- Otherwise on FreeBSD older packages could be used for include
   this then can result to missing declared functions because
   includes and libs are out of sync


wjwithagen:
@tchaikov 
Thanx.

PR #11195, opened 2016-09-22
os/bluestore: a few fixes about  the global csum setting

xiexingguo:


xiexingguo:
Local tested:

2016-09-22 20:58:08.537017 7f4dfee15640  1 bluestore(store_test_temp_dir) fsck finish with 0 errors in 0.521007 seconds
[       OK ] ObjectStore/StoreTest.TooManyBlobsTest/2 (20714 ms)
[----------] 61 tests from ObjectStore/StoreTest (2505935 ms total)

[==========] 61 tests from 1 test case ran. (2505935 ms total)
[  PASSED  ] 61 tests.

src/test/objectstore/store_test.cc
```diff
@@ -4061,7 +4059,6 @@ TEST_P(StoreTest, SyntheticMatrixNoCsum) {
     { "alignment", "512", 0 },
     { "bluestore_min_alloc_size", "65536", "4096", 0 },
     { "bluestore_compression", "force", "none", 0},
-    { "bluestore_csum", "false", 0 },
```
liewegas: need bluestore_csum_type = none here

src/test/objectstore/store_test.cc
```diff
@@ -5106,9 +5101,6 @@ TEST_P(StoreTest, BluestoreOnOffCSumTest) {
   }
   {
     //write with csum disabled followed by read with csum enabled
-    g_conf->set_val("bluestore_csum", "false");
-    g_conf->set_val("bluestore_csum_type", "crc32c");
```
liewegas: need bluestore_csum_type = none here

src/os/bluestore/BlueStore.cc
```diff
@@ -4737,7 +4737,7 @@ int BlueStore::_do_read(
       for (auto reg : b2r_it->second) {
 	// determine how much of the blob to read
 	uint64_t chunk_size = bptr->get_blob().get_chunk_size(
-	  csum_type != bluestore_blob_t::CSUM_NONE, block_size);
+	  csum_type.load() != bluestore_blob_t::CSUM_NONE, block_size);
```
liewegas: I think this doesn't make much sense.  bluestore_csum_type is a kind of csum.. and on read we do'nt get to choose.  I think we need to either just verify csum if it exists or introduce a new option bluestore_csum_verify_on_read that can be disabled if the admin really wants to ignore it.

xiexingguo:
@liewegas Fixed:-)

PR #11198, opened 2016-09-22
[DNM]os/bluestore: using rm_range instead of iterating key ranges

yuyuyu101:
Iterating key ranges introduces ineffective scan which iterate from level1 to levelN. And it won't get benefit from Bloomfilter. In bluestore heavy usage case, iterating is really stupid

### DIFF
--- a/src/kv/KeyValueDB.h
+++ b/src/kv/KeyValueDB.h
@@ -103,6 +103,12 @@ public:
       const std::string &prefix ///< [in] Prefix by which to remove keys
       ) = 0;
 
+    virtual void rm_range_keys(
+      const string &prefix,    ///< [in] Prefix by which to remove keys
+      const string &start,     ///< [in] The start bound of remove keys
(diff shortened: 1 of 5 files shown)


liewegas:
Yes!  I didn't realize rocksdb has a rm_range. 

I suspect the test_kv will crash on leveldb and memdb, though?  Ideally we'd implement a slow iterator-based fallback for those.

yuyuyu101:
yes, need to do.

yuyuyu101:
hmm, it looks rocksdb dev give a joke on the interface.... This interface isn't ready to use, most of implementation is still ongoing.... Of course, this way will happen since myrocksdb also expects this. Refer to https://github.com/facebook/rocksdb/wiki/Delete-A-Range-Of-Keys too.

liewegas:
So sad!  Hopefully this will stabilize sometime soon.

PR #11200, opened 2016-09-22
msg/async: avoid set out of range ms_async_op_threads option

yuyuyu101:

### DIFF
--- a/src/msg/async/Stack.cc
+++ b/src/msg/async/Stack.cc
@@ -69,6 +69,14 @@ NetworkStack::NetworkStack(CephContext *c, const string &t): type(t), started(fa
 {
   const uint64_t InitEventNumber = 5000;
   num_workers = cct->_conf->ms_async_op_threads;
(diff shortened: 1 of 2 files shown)


src/msg/async/Stack.cc
```diff
@@ -69,6 +69,13 @@ NetworkStack::NetworkStack(CephContext *c, const string &t): type(t), started(fa
 {
   const uint64_t InitEventNumber = 5000;
   num_workers = cct->_conf->ms_async_op_threads;
+  if (num_workers >= EventCenter::MAX_EVENTCENTER) {
+    ldout(cct, 0) << __func__ << " max thread limit is " << EventCenter::MAX_EVENTCENTER
+                  << ", change to this now. And no need to spawn thread too much."
```
markhpc: I would change this to read:
", switching to this now.  Higher thread values are unnecessary and currently  unsupported."

Otherwise, looks good!

PR #11201, opened 2016-09-22
jewel: build/ops: backport 'logrotate: Run as root/ceph'

b-ranto:
http://tracker.ceph.com/issues/17382

Currently, we run the logrotate scripts as ceph/ceph but that way we
cannot rotate the scripts created by qemu (they are root/ceph and 644).
The original su line was introduced in commit 73d7bed9 because logrotate
complained that the directory was writable by a non-root group and it
needed a su line to supress that error. This way, we keep the group
settings and we can access and rotate the qemu logs as well.

### DIFF
--- a/src/logrotate.conf
+++ b/src/logrotate.conf
@@ -8,5 +8,5 @@
     endscript
     missingok
     notifempty
-    su ceph ceph
+    su root ceph
 }


smithfarm:
added Backport tracker URL to PR description

dachary:
It passed the upgrade/jewel-x and upgrade/hammer-x (http://tracker.ceph.com/issues/17487#note-22) suites.

PR #11202, opened 2016-09-22
[DNM] bundled Boost submodule

mattbenjamin:
Add Boost submodule at src/boost.
Also add CMake logic to bootstrap src/boost with required components, build it in the cmake step, and use the bundled Boost for build if not overridden.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -38,3 +38,6 @@
 	url = https://github.com/ceph/lua.git
 	branch = lua-5.3-ceph
 	ignore = dirty
+[submodule "src/boost"]
+	path = src/boost
+	url = https://github.com/boostorg/boost.git
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -117,7 +117,6 @@ CHECK_FUNCTION_EXISTS(pthread_setname_np HAVE_PTHREAD_SETNAME_NP)
 CHECK_FUNCTION_EXISTS(pthread_getname_np HAVE_PTHREAD_GETNAME_NP)
 CHECK_FUNCTION_EXISTS(eventfd HAVE_EVENTFD)
 
-CHECK_INCLUDE_FILE_CXX("boost/asio/coroutine.hpp" HAVE_BOOST_ASIO_COROUTINE)
 CHECK_INCLUDE_FILES("inttypes.h" HAVE_INTTYPES_H)
 CHECK_INCLUDE_FILES("linux/types.h" HAVE_LINUX_TYPES_H)
 CHECK_INCLUDE_FILES("linux/version.h" HAVE_LINUX_VERSION_H)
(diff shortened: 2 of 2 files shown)


mattbenjamin:
Took a few tries, but the current changes build to completion for me.  I'm using static linkage to avoid issues w/naming (i.e., the change has a commented line that renames the libs, but then the names of the various libs would differ from the system-boost names, and I didn't adjust for that).

.gitmodules
```diff
@@ -33,3 +33,6 @@
 [submodule "src/isa-l"]
 	path = src/isa-l
 	url = https://github.com/ceph/isa-l
+[submodule "boost"]
+	path = boost
```
tchaikov: better off putting it in `src/boost`

CMakeLists.txt
```diff
@@ -375,17 +374,53 @@ if(LINUX)
...
+option(WITH_SYSTEM_BOOST "require and build with system Boost" OFF)
+
+if (WITH_SYSTEM_BOOST)
+  if(ENABLE_SHARED)
+    set(Boost_USE_STATIC_LIBS OFF)
+  else()
+    set(Boost_USE_STATIC_LIBS ON)
+  endif()
+else()
+  set(BOOST_J 1 CACHE STRING "max jobs for Boost build")
```
tchaikov: @mattbenjamin have you considered using `ExternalProject_Add()`? @cbodley what do you think?

tchaikov: the upside of using `ExternalProject_Add()` is that `cmake` won't take very long to complete, 

- we can only pull the submodules we are interested in using `GIT_REPOSITORY` and `GIT_SUBMODULES`.
- or it can be relatively fast if we just download the dist tarball of boost if we just download it instead of cloning its git repo.
- `cmake` will just do the minimal check and config.

but it might be **very** tedious to add the dependency to boost everywhere we are using it, either header or library.

cbodley: @tchaikov could be worth trying. i found an example here: https://github.com/arnaudgelas/ExternalProject/blob/master/External-Boost.cmake

tchaikov: i was playing with https://paste.fedoraproject.org/433541/43683147/, but `git clone` is not finished yet.

.gitmodules
mattbenjamin: That was my intention, I managed to confuse myself by I think doing it wrong initially and then not cleaning up correctly.

CMakeLists.txt
mattbenjamin: @tchaikov, @cbodley I think Sage made a ruling on the submodule approach, but I don't have a strong preference.  I think this works ok for now (external project)?  Casey mentioned where the objects are built--I think they could be moved into build, if needed.

mattbenjamin: @tchaikov--@cbodley notes that possible advantage of current approach is ability to use findpackage uniformly (note I added the required version number)

tchaikov:
s/path = boost/path = src/boost/

other than this, it looks good to me. @liewegas what do you think

and @alimaredia ?

CMakeLists.txt
liewegas: This hard-coded -j1 worries me...

mattbenjamin: it's not hard coded--it takes on the value of any cmdline provided value of -DBOOST_J ;  I will add a comment on that

liewegas: I mean if you do make -j32, it'll still build boost with -j1...

liewegas: Hmm, so we should make do_cmake.sh set BOOST_J=$(nproc).

mattbenjamin: Yeah, I think it works, I have been verifying that I see the correct number of boost procs when the cmake step runs (and that it is appropriately faster)--are you saying it did not work for you?

liewegas: https://github.com/liewegas/ceph/commit/3a4da1ca6e7a82e0632cc4169666a80d67212a13

liewegas:
detection works now!

but i get build warnings and errors.  i'm on g++ (GCC) 5.3.1 20160406 (Red Hat 5.3.1-6)... mostly this

/home/sage/src/ceph/build/boost/include/boost/type_traits/detail/template_arity_spec.hpp:13:84: note: #pragma message: NOTE: Use of this header (template_arity_spec.hpp) is deprecated

which seems to come from including boost/variant.hpp.  do you see this in your environment too?

mattbenjamin:
I think the current push is ready for re-test.  I fixed several issues that caused occasional failure to install the Boost headers, leading to FindBoost failing.  Also it builds entirely in the cmake build directory, as one would prefer.  I've done multiple builds from clean check-out.

mattbenjamin:
I do see a similar warning;   I don't think variant is going anywhere;  not sure what the actual fix is, but I'll help dig into it

tchaikov:
@mattbenjamin @liewegas this has been fixed upstream (https://github.com/boostorg/iostreams/pull/24) , but 1.61 does not have this fix.

mattbenjamin:
@tchaikov thanks!

mattbenjamin:
I've updated to the Boost commit which takes the warning fix in libs/iostreams.  The result builds, with less warnings.  I notice a few warnings in libs/asio and libs/serialization.  I chased them in gitub, but haven't found closed PRs that address them (there is an open PR from 2015 which maybe addressed the asio warnings, but I think it's safe to say it's stalled out).

liewegas:
https://github.com/ceph/ceph/pull/11265

mattbenjamin:
tx!

PR #11204, opened 2016-09-22
librados: do not request osd ack if no completed completion is set

liewegas:
If an AioCompletion doesn't have a compelte cb set, do not request an ack from the OSD.

This may impact user who use an aio completion and then call wait_for_complete() or poll.  However, there shouldn't be any significant deifference between complete and safe arrival anyway given how the current crop of ObjectStores are implemented.

Notably, this means that all RBD writes get 1 reply instead of 2.  Performanc eshould go up a bit (a few %?).

### DIFF
--- a/src/librados/IoCtxImpl.cc
+++ b/src/librados/IoCtxImpl.cc
@@ -766,7 +766,7 @@ int librados::IoCtxImpl::aio_operate(const object_t& oid,
   if (snap_seq != CEPH_NOSNAP)
     return -EROFS;
 
-  Context *onack = new C_aio_Ack(c);
+  Context *onack = c->wants_ack() ? new C_aio_Ack(c) : NULL;
   Context *oncommit = new C_aio_Safe(c);
 
   c->io = this;
@@ -874,7 +874,7 @@ int librados::IoCtxImpl::aio_write(const object_t &oid, AioCompletionImpl *c,
   if (snap_seq != CEPH_NOSNAP)
     return -EROFS;
 
-  Context *onack = new C_aio_Ack(c);
+  Context *onack = c->wants_ack() ? new C_aio_Ack(c) : NULL;
   Context *onsafe = new C_aio_Safe(c);
 
   c->io = this;
@@ -900,7 +900,7 @@ int librados::IoCtxImpl::aio_append(const object_t &oid, AioCompletionImpl *c,
   if (snap_seq != CEPH_NOSNAP)
     return -EROFS;
 
-  Context *onack = new C_aio_Ack(c);
+  Context *onack = c->wants_ack() ? new C_aio_Ack(c) : NULL;
   Context *onsafe = new C_aio_Safe(c);
 
   c->io = this;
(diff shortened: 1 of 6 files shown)


src/librados/IoCtxImpl.cc
```diff
@@ -982,7 +982,7 @@ int librados::IoCtxImpl::aio_remove(const object_t &oid, AioCompletionImpl *c)
   if (snap_seq != CEPH_NOSNAP)
     return -EROFS;
 
-  Context *onack = new C_aio_Ack(c);
+  Context *onack = c->wants_ack() ? new C_aio_Ack(c) : NULL;
   Context *onsafe = new C_aio_Safe(c);
 
   c->io = this;
```
jdurgin: looking through this file, ```IoCtx::aio_stat() and aio_stat2()``` need to set ```c->is_read``` - the rest of the aio reads already do

liewegas: good catch, added a patch that fixes that

yehudasa:
@liewegas I do see wait_for_complete() in rgw done on AioCompletion with null onack.

liewegas:
Hmm, I wonder if it's a bug.  I'm guessing that aio_wait() is supposed to 
be waiting for a commit, not an ack.

liewegas:
I think the callers should be replaced with wait_for_safe().

liewegas:
@yehudasa added a patch that updates rgw

src/rgw/rgw_rados.cc
```diff
@@ -9201,7 +9201,7 @@ struct get_obj_data : public RefCountedObject {
     librados::AioCompletion *c = iter->second;
     lock.Unlock();
 
-    c->wait_for_complete_and_cb();
+    c->wait_for_safe_and_cb();
```
yehudasa: @liewegas in this case we do have the ack callback set (see at `add_io()`)

liewegas: Is that deliberate?  The ack doesn't mean the write is safe.

yehudasa: probably unintentional

liewegas:
@yehudasa I grepped for all teh aio_create_completion, set_*_callback, *_and_cb, and any other relevant aio operations.  I think I got it all...

liewegas:
should go through rados, rgw, rbd suites

yehudasa:
@liewegas lgtm

liewegas:
good rados run, http://pulpito.ceph.com/sage-2016-10-18_20:36:29-rados-wip-sage-testing2---basic-smithi/

liupan1111:
@liewegas one thing really puzzles me, for filestore, in my opinion:
1) onack: write journal successfully.
2) oncommit(onsafe): write data successfully.

For write, why do we have to register callback_safe? why not just onack? If write journal successfully(use a raw disk as journal disk, and use directIO), I believe the data is safe and writable.

Looking forward your reply!!

liewegas:
On btrfs the fs write and journal write happen in parallel.

liupan1111:
So do you mean the client(rbd) doesn't know the fs in the backend, so you choose "safe" one? For our internal usage, if we choose xfs, how about I just register onack? Maybe big performance improvement, please correct me.

liewegas:
(This discussion is a bit academic now; the ack has been removed from the client and server entirely after this piece merged.)

If you request commit, you would get the fastest reply indicating the write is safe.  If you did ack, you would only know that the write was ordered (and could be read). But in reality that wasn't guaranteed (if the osd restarted the ordering could be lost) and in reality the osd wasn't allowing a read until the previous write was committed anyway.  In short the ack wasn't implemented such that it did any good, and even if it were it would be unreliable and hard to use.  So we removed it.

PR #11206, opened 2016-09-22
osd: Remove extra call to reg_next_scrub() during splits

dzafman:
Add assert() to catch this in the future

Fixes: http://tracker.ceph.com/issues/16474

### DIFF
--- a/src/osd/PG.cc
+++ b/src/osd/PG.cc
@@ -3411,6 +3411,7 @@ void PG::reg_next_scrub()
   double scrub_min_interval = 0, scrub_max_interval = 0;
   pool.info.opts.get(pool_opts_t::SCRUB_MIN_INTERVAL, &scrub_min_interval);
   pool.info.opts.get(pool_opts_t::SCRUB_MAX_INTERVAL, &scrub_max_interval);
+  assert(scrubber.scrub_reg_stamp == utime_t());
   scrubber.scrub_reg_stamp = osd->reg_pg_scrub(info.pgid,
 					       reg_stamp,
 					       scrub_min_interval,
@@ -3420,8 +3421,10 @@ void PG::reg_next_scrub()
 
 void PG::unreg_next_scrub()
 {
-  if (is_primary())
+  if (is_primary()) {
     osd->unreg_pg_scrub(info.pgid, scrubber.scrub_reg_stamp);
+    scrubber.scrub_reg_stamp = utime_t();
+  }
 }
 
 void PG::sub_op_scrub_map(OpRequestRef op)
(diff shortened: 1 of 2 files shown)


athanatos:
lgtm! adding needs-qa

dzafman:
http://pulpito.ceph.com/dzafman-2016-09-22_17:11:48-rados-wip-zafman-testing-distro-basic-smithi/

There is a failure I'm looking at in the run above.  An example is job 432090.  I'm rerunning that job with my wip-zafman-fixes ceph-qa-suite branch.

src/osd/PG.cc
```diff
@@ -3411,6 +3411,7 @@ void PG::reg_next_scrub()
   double scrub_min_interval = 0, scrub_max_interval = 0;
   pool.info.opts.get(pool_opts_t::SCRUB_MIN_INTERVAL, &scrub_min_interval);
   pool.info.opts.get(pool_opts_t::SCRUB_MAX_INTERVAL, &scrub_max_interval);
+  assert(scrubber.scrub_reg_stamp == utime_t());
```
robbat2: Further up, utime_t() was changing to ceph_clock_now(cct). Should this be changed here as well?

src/osd/PG.cc
```diff
@@ -3420,8 +3421,10 @@ void PG::reg_next_scrub()
 
 void PG::unreg_next_scrub()
 {
-  if (is_primary())
+  if (is_primary()) {
     osd->unreg_pg_scrub(info.pgid, scrubber.scrub_reg_stamp);
+    scrubber.scrub_reg_stamp = utime_t();
```
robbat2: Again, should this be ceph_clock_now(cct) instead of utime_t()?

src/osd/PG.cc
dzafman: @rabbat2 No, this was to indicate that no time is set.

src/osd/PG.cc
dzafman: @robbat2 Same here this is supposed to indicate that no time is set.

dzafman: @robbat2 I still see the following at top of master for the utime_t ctor:

utime_t() { tv.tv_sec = 0; tv.tv_nsec = 0; }

robbat2:
We should add in some docs that make it more clear why the unreg scrub uses zero as the value for scrub_reg_stamp. 5e44040e8528bff06cc0a5a3f3293ab146e0e4e1 is what prompted me to ask if ceph_clock_now was more correct.

dzafman:
The _scrub_reg_stamp_  contents is a stashed value when a scrub is registered in reg_next_scrub() that is only used to allow unreg_next_scrub() to find and remove that registration later.  It is not the same as last_scrub_stamp or anything that can be seen outside of the OSD.  I set it to utime_t() to indicate that that scrub isn't registered, so we can assert if you try to remove a scrub twice.  That value is the same as what is initialized by the constructor of PG::Scrubber.

dzafman:
Only one failure:  
79/146 Test  #86: unittest_erasure_code_plugin_jerasure ...***Exception: SegFault  0.13 sec

PR #11207, opened 2016-09-22
rgw: get_zonegroup() uses "default" zonegroup if empty

yehudasa:
Fixes: http://tracker.ceph.com/issues/17372

An empty zonegroup should be replaced with the "default" zonegroup.
This is needed when dealing with zonegroup set in old bucket info,
that predated setting the buckets' region.


oritwas:
@yehudasa, this won't work if you are using a different zonegroup.
How about using the gateway zonegroup?

PR #11209, opened 2016-09-23
mon: clear duplicated logic in MDSMonitor

david-z:
Clear some duplicated logic in MDSMonitor when checking replacement for a failed MDS. It will make this part of logic more clear and readable.

### DIFF
--- a/src/mon/MDSMonitor.cc
+++ b/src/mon/MDSMonitor.cc
@@ -2833,22 +2833,14 @@ void MDSMonitor::maybe_replace_gid(mds_gid_t gid,
     pending_fsmap.promote(sgid, fs, info.rank);
 
     *mds_propose = true;
-  } else if (info.state == MDSMap::STATE_STANDBY_REPLAY) {
-    dout(10) << " failing " << gid << " " << info.addr << " mds." << info.rank << "." << info.inc
-      << " " << ceph_mds_state_name(info.state)
+  } else if (info.state == MDSMap::STATE_STANDBY_REPLAY || 
+             info.state == MDSMap::STATE_STANDBY) {
+    dout(10) << " failing and removing " << gid << " " << info.addr << " mds." << info.rank 
+      << "." << info.inc << " " << ceph_mds_state_name(info.state)
       << dendl;
     fail_mds_gid(gid);
     *mds_propose = true;
-  } else {
-    if (info.state == MDSMap::STATE_STANDBY ||
-        info.state == MDSMap::STATE_STANDBY_REPLAY) {
-      // remove it
-      dout(10) << " removing " << gid << " " << info.addr << " mds." << info.rank << "." << info.inc
-        << " " << ceph_mds_state_name(info.state)
-        << " (laggy)" << dendl;
-      fail_mds_gid(gid);
-      *mds_propose = true;
-    } else if (!info.laggy()) {
+  } else if (!info.laggy()) {
       dout(10) << " marking " << gid << " " << info.addr << " mds." << info.rank << "." << info.inc
         << " " << ceph_mds_state_name(info.state)
         << " laggy" << dendl;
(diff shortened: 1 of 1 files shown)


src/mon/MDSMonitor.cc
```diff
@@ -2833,22 +2833,15 @@ void MDSMonitor::maybe_replace_gid(mds_gid_t gid,
     pending_fsmap.promote(sgid, fs, info.rank);
 
     *mds_propose = true;
-  } else if (info.state == MDSMap::STATE_STANDBY_REPLAY) {
-    dout(10) << " failing " << gid << " " << info.addr << " mds." << info.rank << "." << info.inc
-      << " " << ceph_mds_state_name(info.state)
+  } else if (info.state == MDSMap::STATE_STANDBY_REPLAY || 
+             info.state == MDSMap::STATE_STANDBY) {
+    dout(10) << " failing and removing " << gid << " " << info.addr << " mds." << info.rank 
+      << "." << info.inc << " " << ceph_mds_state_name(info.state)
```
gregsfortytwo: These cases were almost the same, but you've dropped the ```last_beacon.erase(gid);``` line. Did you check git logs to see how we got the duplicated cases?

src/mon/MDSMonitor.cc
```diff
@@ -2833,22 +2833,15 @@ void MDSMonitor::maybe_replace_gid(mds_gid_t gid,
...
-    if (info.state == MDSMap::STATE_STANDBY ||
-        info.state == MDSMap::STATE_STANDBY_REPLAY) {
-      // remove it
-      dout(10) << " removing " << gid << " " << info.addr << " mds." << info.rank << "." << info.inc
-        << " " << ceph_mds_state_name(info.state)
-        << " (laggy)" << dendl;
-      fail_mds_gid(gid);
-      *mds_propose = true;
-    } else if (!info.laggy()) {
+    if (!info.laggy()) {
```
gregsfortytwo: Also this might as well be an "else if" block instead of nested else-if, right?

src/mon/MDSMonitor.cc
david-z: @gregsfortytwo Yes, I dropped last_beacon.erase(gid) because fail_mds_gid(gid) did this work here. 

I checked git log and this duplicated case was introduced back to 2011. At that time this part made sense because the logic was a little different from now.

src/mon/MDSMonitor.cc
david-z: @gregsfortytwo This was an "else if" back to 2011 and then changed to nested one. I think current logic looks good and can be easily understand. If it goes here, it means mds doesn't find a replacement and we can not ignore this mds no matter what current mds state is. We should record laggy_since once and report the warning.   

The whole checking is almost based on mds state. If we pick up a branch to check whether mds is already laggy or not. I don't think the logic is better and more clear than current one. 

What do you think? Thanks.

jcsp:
Tested here: http://pulpito.ceph.com/jspray-2016-09-27_00:45:11-fs-wip-jcsp-testing-20160926-distro-basic-mira

src/mon/MDSMonitor.cc
jcsp: The last_beacon.erase part only made sense in the cases where we were actually removing an MDS.  When we do last_beacon.erase for a GID that is laggy but staying in the map, it just gets replaced at the start of tick() next time (the "make sure last beacon is fully populated") section.

So I think you can delete that last_beacon.erase line, and then collapse the nested if into an `else if` as greg suggests

gregsfortytwo:
Ping @david-z, looks like we're waiting on a few code changes from you. :)

david-z:
@gregsfortytwo sorry for the late reply, I am on vacation in the past week.

david-z:
@gregsfortytwo @jcsp Changes are made as you suggested. Pls help to review. Thanks.

PR #11211, opened 2016-09-23
test: temporarily remove fork()ing flock tests

jcsp:
Usually we fix tests instead of disabling them but in this instance the failure is potentially masking anything else going wrong during the libcephfs tests, and the fix is probably going to be to re-engineer how the tests are written.

http://tracker.ceph.com/issues/16556

### DIFF
--- a/src/test/libcephfs/flock.cc
+++ b/src/test/libcephfs/flock.cc
@@ -70,7 +70,6 @@ static const struct timespec* abstime(struct timespec &ts, long ms) {
 }
 
 /* Basic locking */
-
 TEST(LibCephFS, BasicLocking) {
   struct ceph_mount_info *cmount = NULL;
   STARTUP_CEPH();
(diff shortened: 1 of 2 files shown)


jcsp:
@gregsfortytwo as discussed by email, taking these tests out of circulation.  I've gone with deleting rather than switching off, because it reduces risk of confusion (someone seeing these tests in the tree and assuming they're running).

jcsp:
I noticed that I had also clobbered some non-forking tests, so revised this to selectively DISABLED_ the bad guys.

gregsfortytwo:
I much prefer this version!

gregsfortytwo:
@jcsp, not merging this since I don't have a test run happy — feel free if you already did that. :)

jcsp:
@gregsfortytwo happy to merge this?

jcsp:
(Checked it disables the tests properly here: http://pulpito.ceph.com/jspray-2016-10-14_09:39:15-fs-wip-no-recordlock-test-distro-basic-mira/)

jcsp:
Sorry for noise, I somehow failed to read the last couple of comments

PR #11214, opened 2016-09-23
workunits: build kernel in unique temporary dir

batrick:
This allows multiple instances of the workunit to be run in parallel in
the same directory. I'm using this change for performance testing.


gregsfortytwo:
In teuthology we handle this by creating test directories for each client, above the workunit layer. Can you not do the same?

batrick:
Ya... I'll just do that instead.

PR #11216, opened 2016-09-23
os/bluestore: prevent extent merging across shard boundaries

liewegas:
We cannot have a single extent span a shard boundary.  If we reach
a shard boundary, stop merging extents.

Note that it might be better to move the shard boundary, but it is
awkward to force that to happen at this point.

### DIFF
--- a/src/os/bluestore/BlueStore.cc
+++ b/src/os/bluestore/BlueStore.cc
@@ -1888,6 +1888,20 @@ int BlueStore::ExtentMap::compress_extent_map(uint64_t offset, uint64_t length)
   if (p != extent_map.begin()) {
     --p;  // start to the left of offset
   }
+
+  // identify the *next* shard
+  auto pshard = shards.begin();
+  while (pshard != shards.end() &&
+	 p->logical_offset >= pshard->offset) {
+    ++pshard;
+  }
+  uint64_t shard_end;
+  if (pshard != shards.end()) {
+    shard_end = pshard->offset;
+  } else {
+    shard_end = OBJECT_MAX_SIZE;
+  }
+
   auto n = p;
   for (++n; n != extent_map.end(); p = n++) {
     if (n->logical_offset > offset + length) {
@@ -1896,13 +1910,25 @@ int BlueStore::ExtentMap::compress_extent_map(uint64_t offset, uint64_t length)
     while (n != extent_map.end() &&
 	   p->logical_offset + p->length == n->logical_offset &&
 	   p->blob == n->blob &&
-	   p->blob_offset + p->length == n->blob_offset) {
+	   p->blob_offset + p->length == n->blob_offset &&
+	   n->logical_offset < shard_end) {
+      dout(20) << __func__ << " 0x" << std::hex << offset << "~" << length
+	       << " next shard 0x" << shard_end << std::dec
+	       << " merging " << *p << " and " << *n << dendl;
       p->length += n->length;
       rm(n++);
       ++removed;
     }
     if (n == extent_map.end()) {
       break;
+    }
+    if (n->logical_offset >= shard_end) {
+      ++pshard;
+      if (pshard != shards.end()) {
+	shard_end = pshard->offset;
+      } else {
+	shard_end = OBJECT_MAX_SIZE;
+      }
     }
   }
   return removed;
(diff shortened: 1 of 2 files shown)


liewegas:
retest this please

liewegas:
retest this please

somnathr:
@liewegas I am seeing the following rocksdb tx after 1 hour run of 4K RW.

2016-09-26 14:25:00.624709 7fb8ddbaa700 30 submit_transaction Rocksdb transaction:
Put( Prefix = M key = 0x000000000000083d'.0000000018.00000000000000077820' Value size = 182)
Put( Prefix = M key = 0x000000000000083d'._info' Value size = 855)
Put( Prefix = O key = 0x7f80000000000000016b93fad7217262'd_data.10156b8b4567.000000000001df15!='0xfffffffffffffffeffffffffffffffff001b1000'x' Value size = 583)
Put( Prefix = O key = 0x7f80000000000000016b93fad7217262'd_data.10156b8b4567.000000000001df15!='0xfffffffffffffffeffffffffffffffff Value size = 4689)
Merge( Prefix = b key = 0x00000048de880000 Value size = 16)
Merge( Prefix = b key = 0x00000054a7a00000 Value size = 16)

out of that spanning blob part is 4689. 
If I understood this PR correctly, we don't want an extent span across 2 shards as we don't want to write both the shards in future if IO touches only one (?).
This PR will not be solving the problem of writing the ever growing entire spanning blobs. 
If so, is it possible to write only the portion of the spanning blob that is changed during this IO ?

liewegas:
On Mon, 26 Sep 2016, somnathr wrote:
> @liewegas I am seeing the following rocksdb tx after 1 hour run of 4K RW.
> 
> 2016-09-26 14:25:00.624709 7fb8ddbaa700 30 submit_transaction Rocksdb transaction:
> Put( Prefix = M key = 0x000000000000083d'.0000000018.00000000000000077820' Value size = 182)
> Put( Prefix = M key = 0x000000000000083d'._info' Value size = 855)
> Put( Prefix = O key = 0x7f80000000000000016b93fad7217262'd_data.10156b8b4567.000000000001df15!='0xfffffffffffffffeffffffffffffffff001b1000'x'
> Value size = 583)
> Put( Prefix = O key = 0x7f80000000000000016b93fad7217262'd_data.10156b8b4567.000000000001df15!='0xfffffffffffffffeffffffffffffffff Value size
> = 4689)
> Merge( Prefix = b key = 0x00000048de880000 Value size = 16)
> Merge( Prefix = b key = 0x00000054a7a00000 Value size = 16)
> 
> out of that spanning blob part is 4689.
> If I understood this PR correctly, we don't want an extent span across 2 shards as we don't want to write both the shards in future if IO
> touches only one (?).

Right

> This PR will not be solving the problem of writing the ever growing entire spanning blobs.
> If so, is it possible to write only the portion of the spanning blob that is changed during this IO ?

The spanning blobs should normally be small.  The current exception seems 
to be if you write the whole object sequentialy and get one big blob and 
extent, with 4KB blocks and crc32c, for about 4KB of blob.  I think the 
"fix" is to set a max blob size.. which we don't do currently unless 
compression is enabled.

But it will take a bit of tuning.. we need to make sure the max blob size 
is small enough that the target shard size * (1+target shard size slop) 
will find the big blob's boundaries.  Otherwise, one or more big blobs can 
end up being spanning blobs.  And we don't want *any* of them to be 
spanning blobs if we can help it.  I'm guessing a max blob of 512KB (about 
512 bytes of metadata) would be okay.

*OR*, we could make the reshard smart enough to cleave an uncompressed 
blob in half when it picks a bound.  That is probably better, actually.

src/os/bluestore/BlueStore.cc
```diff
@@ -1888,6 +1888,20 @@ int BlueStore::ExtentMap::compress_extent_map(uint64_t offset, uint64_t length)
   if (p != extent_map.begin()) {
     --p;  // start to the left of offset
   }
+
+  // identify the *next* shard
+  auto pshard = shards.begin();
+  while (pshard != shards.end() &&
```
allensamuels: Really should assert that p != end(). Not sure if seek_lextent really guarantees it. But if it doesn't things will really go south when you do n = p+1 later :(

somnathr: seek_lextent doesn't guarantee that and that's why all the caller of seek_lextent() has a end check except this.
We should add that check or assert here I agree.

PR #11217, opened 2016-09-23
rpm: fix permissions for /etc/ceph/rbdmap

ktdreyer:
Prior to this change, the RPM packaging would install /etc/ceph/rbdmap with exectuable permissions. The execute bit is not necessary and does not match what the Debian packaging does. Remove the execute bit in this case.

Fixes: http://tracker.ceph.com/issues/17395


ktdreyer:
Thanks @mbukatov for reporting this.

smithfarm:
:+1:

badone:
lgtm

liewegas:
Oops, forgot to put Nathan in the merge commit--I'm sorry!

PR #11218, opened 2016-09-24
DNM: pass "UserPerm" struct throughout Client for security checks

gregsfortytwo:
This PR creates a new "UserPerm" struct that can handle a UID and an array of GIDs.
It passes it throughout the request pipeline, which is a significant improvement over our existing uid/gid param pair.

Additionally, it stops the practice of converting "-1" UID/GID params into whatever the Client process is running as. That resulted in problems such as breaking Ganesha's root squash configuration (http://tracker.ceph.com/issues/16367), in addition to just being weird.

We do not yet export UserPerms or similar interfaces to libcephfs users; that is the next step.


gregsfortytwo:
This is DNM as I still need to run tests against the post-master-merge version, but it should be ready for review. Sorry about the messy history, but it wasn't very feasible to squash after a couple rebases. Hopefully it's not too bad!

Note that this needs to run against https://github.com/ceph/ceph-qa-suite/pull/1187 as there are some changes to security options.

jtlayton:
I took a look over the set earlier and had a look at the merge commit and it all looks pretty sane to me. Handing back to @gregsfortytwo for merge when it's ready.

ffilz:
This is looking good, scanned over some of the patches.

gregsfortytwo:
@jtlayton, you'll want to note the freshly-disabled tests. Right now we don't have a way of specifying multiple GIDs in our test suite because of the libcephfs limitations.

We may also decide we need to let our internal Client interface specify multiple GIDs to use on each op, but...I don't think we will?

gregsfortytwo:
@jcsp, did you want to take a look at this before merging?

gregsfortytwo:
(Still DNM as we need a fresh test run, coming shortly, but I believe I've squashed all the bugs that showed up last time so we should be good.)

jcsp:
:+1: I'm happy if you guys are both happy

jtlayton:
Ok, I was seeing some bad frees happening, and I think the UserPerm code still needs a little work. This patch seems to fix it. @gregsfortytwo , want to squash that into an existing commit?

https://github.com/jtlayton/ceph/commit/1672265be7ef35b422a90f64b65ebf724050f025

gregsfortytwo:
Squashed a slightly different commit in that also cleans up the silly practice of passing *this in to deep_copy, since it's not a static function. Gitbuilders are way behind but it passed a local ceph_test_libcephfs (which broke on the last go-round) and assuming next run is happy I'll merge.

PR #11222, opened 2016-09-25
coding style

hhktony:
librbd/internal.cc add space after comma

Signed-off-by: Tony Xu <hhktony@gmail.com>


tchaikov:
@hhktony Tony, i am closing your PR, as it only involves formatting changes. in general we adjust the format when we are on that very area, otherwise it just make the output of "git blame" way harder to read.

PR #11223, opened 2016-09-26
os/bluestore: fix spanning blob leak from ~ExtentMap

somnathr:
If a Blob is spanning is not getting deleted as spanning_blob_map
is holding one ref.Explicitly decreasing Blob ref from ExtentMap
destructor and erasing from spanning_blob_map is fixing the leak.


liewegas:
Good catch!

PR #11225, opened 2016-09-26
client: protect InodeRef with client_lock

ukernel:
Fixes: http://tracker.ceph.com/issues/17392


gregsfortytwo:
Is this everywhere we grab an InodeRef that was unlocked?

ukernel:
yes

jcsp:

PR #11227, opened 2016-09-26
build/ops: Detect crc32 extension support from assembler (AArch64)

agraf:
The used compiler may or may not be recent enough to recognize the
crc32 extended cpu type. However, it does not really have to know about
them either, since all we do is pass inline assembly instructions to
the assembler.

This patch moves the crc cpu extension detection from compiler based
to assembler based, so that we can build optimized code even when the
compiler does not know about the cpu type yet.

Signed-off-by: Alexander Graf <agraf@suse.de>

### DIFF
--- a/cmake/modules/SIMDExt.cmake
+++ b/cmake/modules/SIMDExt.cmake
@@ -15,9 +15,20 @@
 #
 
 if(CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64|AARCH64")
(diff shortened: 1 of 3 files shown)


smithfarm:
@agraf: Needs rebase. While you're doing that, please add the following line to the commit message immediately prior to `Signed-off-by:`

```
Fixes: http://tracker.ceph.com/issues/17516
```

Thanks!

smithfarm:
Closing in favor of #11391

PR #11230, opened 2016-09-26
rgw: bucket resharding

yehudasa:
An offline tool to reshard a bucket's index. Usage:

$ radosgw-admin bucket reshard --bucket=<bucket> --num-shards=<num_shards>

This will create a new linked bucket instance that points to the newly created index objects. The old bucket instance still exists and currently it's up to the user to manually remove the old bucket index objects.


src/rgw/rgw_admin.cc
```diff
@@ -4607,6 +4668,107 @@ int main(int argc, char **argv)
...
+      return -ret;
+    }
+
+    int num_source_shards = (bucket_info.num_shards > 0 ? bucket_info.num_shards : 1);
+
+    RGWBucketInfo new_bucket_info(bucket_info);
+    store->create_bucket_id(&new_bucket_info.bucket.bucket_id);
+    new_bucket_info.bucket.oid.clear();
+
+    new_bucket_info.num_shards = num_shards;
```
oritwas: Don't we need to check num_shards doesn't exceed the max?
also we can optimize by checking if the number of shards is the same.

oritwas:
lgtm

oritwas:
The usage help text is missing.
A short documentation will be nice.

src/cls/rgw/cls_rgw.cc
```diff
@@ -2530,12 +2533,22 @@ static int rgw_bi_list_op(cls_method_context_t hctx, bufferlist *in, bufferlist
     return ret;
   }
 
+  count += ret;
```
oritwas: We can check if we didn't exceed max here and avoid the next listing

oritwas:
@yehudasa , can you rebase?

PR #11232, opened 2016-09-27
os/bluestore: garbage collect partially overlapped blobs

Roushan81:
Because of overwrite operation, a new blob is created instead of merging data with the old blob. This leads to overlapping of blobs(example given below).
...................
......................
......................
Because of overlapping blobs, space is wasted and read-performance can degrade. As part of this code change, we try to garbage collect the overlapping blobs. If the new write request results in overlapping count of more than 3, we extent the write request so that it completely overwrite the blobs adjacent to the new write request.

To keep track of the overlapping count, a new persistent field "blob_depth" is added to lextent. Whenever we add a new lextent, the "blob_depth" is set to the max of all the overlapping lextents + 1 (In fact we need not check all the overlapping lextents, we just need to check only two lextents covering the two ends of the new lextents) . This will not always give the correct depth. It may over count in certain scenarios but will never undercount. Note: if lextents are splitted at the blobs boundary, it may be possible to correctly keep track of the depth, but that will increase the lextent counts and make "depth" update operation complicated.

This pr is based on #10406 , rebased to latest code and fixed the review comments

### DIFF
--- a/src/common/config_opts.h
+++ b/src/common/config_opts.h
@@ -972,6 +972,8 @@ OPTION(bluestore_compression, OPT_STR, "none")  // force|aggressive|passive|none
 OPTION(bluestore_compression_algorithm, OPT_STR, "snappy")
 OPTION(bluestore_compression_min_blob_size, OPT_U32, 256*1024)
 OPTION(bluestore_compression_max_blob_size, OPT_U32, 4*1024*1024)
+OPTION(bluestore_gc_max_blob_depth, OPT_U32, 3)
+OPTION(bluestore_gc_merge_data, OPT_BOOL, true)
 /*
  * Require the net gain of compression at least to be at this ratio,
  * otherwise we don't compress.
--- a/src/os/bluestore/BlueStore.cc
+++ b/src/os/bluestore/BlueStore.cc
@@ -63,7 +63,8 @@ const string PREFIX_SHARED_BLOB = "X"; // u64 offset -> shared_blob_t
 #define BLOBID_FLAG_ZEROOFFSET 0x2  // blob_offset is 0
 #define BLOBID_FLAG_SAMELENGTH 0x4  // length matches previous extent
 #define BLOBID_FLAG_SPANNING   0x8  // has spanning blob id
-#define BLOBID_SHIFT_BITS        4
+#define BLOBID_FLAG_BLOB_DEPTH 0x16 // has blob overlapping count of 1
+#define BLOBID_SHIFT_BITS        5
 
 /*
  * object name key structure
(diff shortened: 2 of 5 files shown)


liewegas:
A few cosmetic nits, but otherwise this looks good to go.  Thanks!

src/os/bluestore/BlueStore.cc
```diff
@@ -2421,6 +2434,7 @@ void BlueStore::_init_logger()
   b.add_u64(l_bluestore_txc, "bluestore_txc", "Transactions committed");
   b.add_u64(l_bluestore_onode_reshard, "bluestore_onode_reshard",
 	    "Onode extent map reshard events");
+  b.add_u64(l_bluestore_gc_bytes, "bluestore_gc_bytes", "Total garbage collection");
```
Roushan81: No. Only the delta is added. Updated the comment.

src/os/bluestore/BlueStore.cc
```diff
@@ -4586,15 +4600,20 @@ struct region_t {
...
-    length(len) {}
+    length(len){}
+    //length(len),
+    //blob_depth(b_depth) {}
   region_t(const region_t& from)
     : logical_offset(from.logical_offset),
     blob_xoffset(from.blob_xoffset),
-    length(from.length) {}
+    length(from.length){}
+    //blob_depth(from.blob_depth) {}
```
Roushan81: Done

src/os/bluestore/BlueStore.cc
```diff
@@ -7436,6 +7456,119 @@ void BlueStore::_wctx_finish(
   }
 }
 
+bool BlueStore::blobs_need_garbage_collection(
```
Roushan81: Done.

src/common/config_opts.h
```diff
@@ -972,6 +972,8 @@ OPTION(bluestore_compression, OPT_STR, "none")  // force|aggressive|passive|none
 OPTION(bluestore_compression_algorithm, OPT_STR, "snappy")
 OPTION(bluestore_compression_min_blob_size, OPT_U32, 256*1024)
 OPTION(bluestore_compression_max_blob_size, OPT_U32, 4*1024*1024)
+OPTION(bluestore_blob_depth_for_gc, OPT_U32, 3)
+OPTION(bluestore_merge_gc_data, OPT_BOOL, true)
```
Roushan81: Done

liewegas:
Sorry, one that thing: can you squash this into a single commit, make sure the commit message is prefixed with "os/bluestore:" (somethign like "os/bluestore: add garbage collection"), and make sure it has a Signed-off-by: line?  Thanks!

PR #11233, opened 2016-09-27
doc: fix start development cluster operation in index.rst

zealoussnow:
Signed-off-by: Leo Zhang <nguzcf@gmail.com>

### DIFF
--- a/doc/dev/index.rst
+++ b/doc/dev/index.rst
@@ -163,21 +163,12 @@ All newcomers are encouraged to read that file carefully.
 Building from source
 --------------------
 
-See instructions at :doc:`/dev/quick_guide` and :doc:`/install/build-ceph`.
+See instructions at :doc:`/install/build-ceph`.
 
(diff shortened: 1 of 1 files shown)


zealoussnow:
@tchaikov is this correct, can you help me review this? Thank you very much.

doc/dev/index.rst
```diff
@@ -171,11 +171,11 @@ Development-mode cluster
 You can start a development-mode Ceph cluster, after compiling the source,
 with::
 
-    cd src
-    install -d -m0755 out dev/osd0
-    ./vstart.sh -n -x -l
+    cd build
```
tchaikov: maybe we can drop this section "Building from source" as a whole, as https://github.com/ceph/ceph/blob/master/doc/dev/quick_guide.rst#running-a-development-deployment also documents this area.

zealoussnow: Now, it looks OK?

PR #11234, opened 2016-09-27
doc/release-notes: v10.2.3 jewel

theanalyst:
Adding release notes for v10.2.3 and to releases timeline


dachary:
The commit messages should be prefixed with the name of the component. It can simply be done by updating the pull request title and the --strict option will verify it is ok. A little tedious but not that difficult and a great help to the reader (IMHO ;-).

theanalyst:
> The commit messages should be prefixed with the name of the component. It can simply be done by updating the pull request title and the --strict option will verify it is ok. A little tedious but not that difficult and a great help to the reader (IMHO ;-).

will update the pr, thanks

PR #11235, opened 2016-09-27
os/bluestore: more cleanups

xiexingguo:


src/os/bluestore/BlueStore.cc
```diff
@@ -7471,28 +7472,31 @@ int BlueStore::_do_write(
...
+        (alloc_hints & CEPH_OSD_ALLOC_HINT_FLAG_RANDOM_WRITE) == 0) {
+      dout(20) << __func__ << " will prefer large blob and csum sizes" << dendl;
+      wctx.comp_blob_size = comp_max_blob_size;
+      wctx.csum_order = min_alloc_size_order;
+    } else {
+      wctx.comp_blob_size = comp_min_blob_size;
+    }
+    dout(20) << __func__ << " prefer csum_order " << wctx.csum_order
+	     << " comp_blob_size 0x" << std::hex << wctx.comp_blob_size
+	     << std::dec << dendl;
```
liewegas: I'm worried about the cost of this lock on every write op.  We are looking at 3 values, comp_mode, and the min and max blob sizes.  They are all a single work.  I don't think it matters if they are updated incoherently (e.g., new max blob  size, old comp_mode).  Is there a specific case you're worried about?  Can we solve it with atomics instead of a lock?

xiexingguo:
Move that one into https://github.com/ceph/ceph/pull/11244 and append more.
Also locally verified:
```
2016-09-28 16:39:53.981001 7fd330a3e640  1 bluestore(bluestore.test_temp_dir) fsck finish with 0 errors in 0.506975 seconds
[       OK ] ObjectStore/StoreTest.TooManyBlobsTest/2 (22015 ms)
[----------] 61 tests from ObjectStore/StoreTest (2555074 ms total)

[==========] 61 tests from 1 test case ran. (2555074 ms total)
[  PASSED  ] 61 tests.

PR #11237, opened 2016-09-27
os: move_ranges_destroy_src

liewegas:


liewegas:
[==========] 248 tests from 1 test case ran. (2739655 ms total)
[  PASSED  ] 248 tests.

manali26:
Sage, I don't see this file src/os/Transaction.cc changes here. Is this desired ?

liewegas:
Yeah, the Transaction.cc change was only needed for the tbl compatibility, 
which we are about to rip out of master. See 
https://github.com/ceph/ceph/pull/9064

dachary:
jenkins test this please

PR #11239, opened 2016-09-27
os/bluestore: add max blob size; fix compressed min blob size logic

liewegas:


liewegas:
@allensamuels

somnathr:
@liewegas If I understood this correctly, it will chunk the blob size in case of bigger block wites. For ex, if I am doing 1M preconditioning , in stead of creating single blob of 1M , it will create 2 X 512K blob.
But, I didn't understand how it is affecting the spanning blob size if I start writing 4K.
For 4K in this case there will be still one blob and similar size extent. 
Are you saying that since the old 1M blob is still there , it is the culprit of spanning across shards ?
Are you seeing any harm if I decrease the max blob size way below if I know my workload will be predominantly small say 4K ?

liewegas:
If the preconditioned big blobs are smaller (say, 512k), then when we 
reshard it will be easier to pick a shard boundary that is close enough to 
where we think it should go.  There is the target blob size and target 
blob size slop (a factor or 20% by default, so a 40% spread).  reshard() 
needs to be able to find a boundary within that range and it'll be easier 
with 512k chunks.  You might also turn up the slop factor (to, say, .3 
or .4).

If we fail to find a boundary, we have no choice but to make the big blob 
a spanning blob...

somnathr:
@liewegas Not helping much till now. After 45 min of 4K RW run after 1M preconditioning of 1TB image , I am getting the following.

2016-09-28 19:32:00.848135 7f3d63dfe700 30 submit_transaction Rocksdb transaction:
Put( Prefix = M key = 0x0000000000000833'.0000000018.00000000000000065470' Value size = 182)
Put( Prefix = M key = 0x0000000000000833'._info' Value size = 855)
Put( Prefix = O key = 0x7f800000000000000149a82a']!rbd_data.10156b8b4567.0000000000000273!='0xfffffffffffffffeffffffffffffffff0039a000'x' Value size = 342)
Put( Prefix = O key = 0x7f800000000000000149a82a']!rbd_data.10156b8b4567.0000000000000273!='0xfffffffffffffffeffffffffffffffff Value size = 4115)
Delete( Prefix = B key = 0x0000006e3648e000)
Put( Prefix = B key = 0x0000006e3648f000 Value size = 8)
Put( Prefix = B key = 0x0000006668a0b000 Value size = 8)

spanning blob is ~4K.

Will see how it behaves after 2-3 hours.

somnathr:
After 2 hour run here is one of the inode size.

2016-09-28 20:48:42.598846 7f3d63dfe700 30 submit_transaction Rocksdb transaction:
Put( Prefix = M key = 0x000000000000088b'.0000000018.00000000000000147583' Value size = 182)
Put( Prefix = M key = 0x000000000000088b'._info' Value size = 855)
Put( Prefix = O key = 0x7f8000000000000001be8d56c7217262'd_data.10156b8b4567.00000000000204bf!='0xfffffffffffffffeffffffffffffffff00222000'x' Value size = 942)
Put( Prefix = O key = 0x7f8000000000000001be8d56c7217262'd_data.10156b8b4567.00000000000204bf!='0xfffffffffffffffeffffffffffffffff002b1000'x' Value size = 740)
Put( Prefix = O key = 0x7f8000000000000001be8d56c7217262'd_data.10156b8b4567.00000000000204bf!='0xfffffffffffffffeffffffffffffffff Value size = 5550)
Delete( Prefix = B key = 0x0000007969bdc000)
Put( Prefix = B key = 0x0000007969bdd000 Value size = 8)
Delete( Prefix = B key = 0x00000055e8b00000)
Put( Prefix = B key = 0x00000055e8aff000 Value size = 8)


Total of 7232 bytes only for onode and out of that spanning blob is of 5.5 K and 2 shards.

Do you want me to try with reduced blob size something like 64K or even lower ?

liewegas:
Try with a larger shard target size slop (.5 maybe)

somnathr:
@liewegas can we merge this as well ?

PR #11246, opened 2016-09-28
os/filestore: Fix erroneous WARNING: max attr too small

badone:
When we read the conf values from the conf file we call handle_conf_change which
calls set_xattr_limits_via_conf but this is before m_fs_type is initialised so
we fall through to the default comaprison with
filestore_max_xattr_value_size_other (1024) and erroneously issue the warning.
This happens at every startup but subsequent calls once the filestore is mounted
proceed as expected. Don't call set_xattr_limits_via_conf unless the filestore
is mounted.

Fixes: http://tracker.ceph.com/issues/17420


yuriw:
@badone was it reviewed ?

badone:
@yuriw not yet, I was hoping setting "needs-review" would help with that.

tchaikov:
lgtm.

tchaikov:

PR #11249, opened 2016-09-28
osd: Flush Journal on shutdown

wido:
This way a data store is consistent and hot-swappable if a OSD had
a clean shutdown.

Make this behavior configurable and enable it by default.

Signed-off-by: Wido den Hollander wido@42on.com

This is a better version of #11025

### DIFF
--- a/src/common/config_opts.h
+++ b/src/common/config_opts.h
@@ -604,6 +604,7 @@ OPTION(osd_uuid, OPT_UUID, uuid_d())
 OPTION(osd_data, OPT_STR, "/var/lib/ceph/osd/$cluster-$id")
 OPTION(osd_journal, OPT_STR, "/var/lib/ceph/osd/$cluster-$id/journal")
 OPTION(osd_journal_size, OPT_INT, 5120)         // in mb
+OPTION(osd_journal_flush_on_shutdown, OPT_BOOL, true) // Flush journal to data store on shutdown
 // flags for specific control purpose during osd mount() process. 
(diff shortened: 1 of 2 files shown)


src/common/config_opts.h
```diff
@@ -604,6 +604,7 @@ OPTION(osd_uuid, OPT_UUID, uuid_d())
 OPTION(osd_data, OPT_STR, "/var/lib/ceph/osd/$cluster-$id")
 OPTION(osd_journal, OPT_STR, "/var/lib/ceph/osd/$cluster-$id/journal")
 OPTION(osd_journal_size, OPT_INT, 5120)         // in mb
+OPTION(osd_journal_flush_on_shutdown, OPT_BOOL, true) // Flush journal to data store on shutdown
```
yuyuyu101: I suggest we set it false by default. Otherwise, I'm not sure a obvious stop behavior change is ok to user. And a long stop will cause lots of ops scripts failed?

liewegas: If we ever want it to be default, then now is the time to change that (kraken or luminous).  I wouldn't expect this would that *that* long, either?

wido:
Small ping on this one @liewegas ? Maybe this can still go into Kraken?

liewegas:
@wido yeah it will, we're a bit backed up on testing and merging is all.  This should go in soon.

tchaikov:

tchaikov:
@liewegas is this good to merge?

PR #11255, opened 2016-09-29
osd: fix scrub boundary to not include a SnapSet

dzafman:


dzafman:
Passed manually testing, and osd-scrub-repair.sh and osd-scrub-snaps.sh.  Running make check now and pushed as wip-zafman-testing2 for gitbuilders to build then I'll run a rados suite.

dzafman:
Rados suite 222 passed and 12 unrelated failures:

teuthology:/a/dzafman-2016-09-30_13:00:52-rados-wip-zafman-testing2-distro-basic-smithi

dzafman:
@athanatos I added this feature to tracker.  Should we add "Feature: http://tracker.ceph.com/issues/17470" to the merge commit or put it on the individual ones?  Or use "Fixes:" ?

src/osd/osd_types.cc
```diff
@@ -5085,7 +5085,7 @@ void ScrubMap::encode(bufferlist& bl) const
 
 void ScrubMap::decode(bufferlist::iterator& bl, int64_t pool)
 {
-  DECODE_START_LEGACY_COMPAT_LEN(3, 2, 2, bl);
+  DECODE_START_LEGACY_COMPAT_LEN(3, 3, 3, bl);
```
athanatos: What's going on here?

dzafman: @athanatos Version 2 is so old, it won't be an issue anymore, so I removed code for version 2 compatibility that would need to be updated with your change.

athanatos: It doesn't seem like it would need to be updated, it's handling the pool encoding change, not related to ordering? (Not that we shouldn't remove it, just not in this patch series hopefully)

src/osd/osd_types.cc
```diff
@@ -5096,20 +5096,6 @@ void ScrubMap::decode(bufferlist::iterator& bl, int64_t pool)
   ::decode(valid_through, bl);
   ::decode(incr_since, bl);
   DECODE_FINISH(bl);
-
-  // handle hobject_t upgrade
-  if (struct_v < 3) {
-    map<hobject_t, object, hobject_t::BitwiseComparator> tmp;
-    tmp.swap(objects);
```
dzafman: @athanatos This wouldn't build because ScubMap objects isn't a simple map with BitwiseCompatator.

athanatos: You're right... why does it seem to build on mine?  Hmm...

athanatos: Oh, it's already fixed in my branch, you should probably reset to mine.

dzafman: @athanatos You changed it to hobject_t::ComparatorWithDefault, so it would build.  If I reset to yours, I'll have to start all the testing over again.

athanatos: You can take a look at the diff between them.  I found another bug with the range_calculation in chunky_scrub other that the one you fixed.  The new patch I added to fix the snap_mapper error necessitates restarting testing anyway -- it's a really really rare race, and we need to run a large set of tests to validate it.

dzafman: @athanatos collection_list() claims to not return the "end" ghobject_t argument.   It is used at the primary and on the replicas from objects_list_partial() and objects_list_range(), respectively.

src/osd/PG.cc
```diff
@@ -4064,54 +4066,54 @@ void PG::chunky_scrub(ThreadPool::TPHandle &handle)
...
+	   * scrubbing that chunk and scrubbing the chunk including head.
+	   * This isn't true for any of the other clones since clones can
+	   * only be created "just to the left of" head.  There is one exception
+	   * to this: promotion of clones which always happens to the left of the
+	   * left-most clone, but promote_object checks the scrubber in that
+	   * case, so it should be ok.  Also, it's ok to "miss" clones at the
+	   * left end of the range if we are a tier because they may legitimately
+	   * not exist (see _scrub).
+	   */
+	  unsigned min = MAX(3, cct->_conf->osd_scrub_chunk_min);
```
dzafman: @athanatos This line is a correction to your branch's version.

dzafman:
1 rados suite had 297 passed tests and 10 accounted for failures.  Another run has been started.

http://pulpito.ceph.com/dzafman-2016-10-05_15:00:09-rados-wip-zafman-testing2-distro-basic-smithi

dzafman:
Second test run had 9 failures all accounted for:

http://pulpito.ceph.com/dzafman-2016-10-06_17:02:30-rados-wip-zafman-testing2-distro-basic-smithi

athanatos:
@liewegas Final review and merge?

PR #11257, opened 2016-09-29
journal: ensure in-flight ops are complete destroying journaler

trociny:

### DIFF
--- a/src/journal/JournalMetadata.cc
+++ b/src/journal/JournalMetadata.cc
@@ -1029,6 +1029,12 @@ void JournalMetadata::async_notify_update(Context *on_safe) {
   comp->release();
 }
 
+void JournalMetadata::wait_for_ops() {
+  C_SaferCond ctx;
+  m_async_op_tracker.wait_for_ops(&ctx);
+  ctx.wait();
+}
+
 void JournalMetadata::handle_notified(int r) {
   ldout(m_cct, 10) << "notified journal header update: r=" << r << dendl;
 }
--- a/src/journal/JournalMetadata.cc
+++ b/src/journal/JournalMetadata.cc
@@ -1014,10 +1014,10 @@ void JournalMetadata::notify_update() {
   m_ioctx.notify2(m_oid, bl, 5000, NULL);
 }
 
-void JournalMetadata::async_notify_update() {
+void JournalMetadata::async_notify_update(Context *on_safe) {
   ldout(m_cct, 10) << "async notifying journal header update" << dendl;
 
-  C_AioNotify *ctx = new C_AioNotify(this);
+  C_AioNotify *ctx = new C_AioNotify(this, on_safe);
   librados::AioCompletion *comp =
     librados::Rados::aio_create_completion(ctx, NULL,
                                            utils::rados_ctx_callback);
--- a/src/journal/JournalMetadata.h
+++ b/src/journal/JournalMetadata.h
@@ -153,6 +153,8 @@ public:
   void notify_update();
   void async_notify_update(Context *on_safe);
 
+  void wait_for_ops();
+
 private:
   typedef std::map<uint64_t, uint64_t> AllocatedEntryTids;
   typedef std::list<JournalMetadataListener*> Listeners;
--- a/src/journal/JournalMetadata.h
+++ b/src/journal/JournalMetadata.h
@@ -151,7 +151,7 @@ public:
   void committed(uint64_t commit_tid, const CreateContext &create_context);
 
   void notify_update();
-  void async_notify_update();
+  void async_notify_update(Context *on_safe);
 
 private:
   typedef std::map<uint64_t, uint64_t> AllocatedEntryTids;
@@ -220,9 +220,10 @@ private:
 
   struct C_AioNotify : public Context {
     JournalMetadata* journal_metadata;
+    Context *on_safe;
 
-    C_AioNotify(JournalMetadata *_journal_metadata)
-      : journal_metadata(_journal_metadata) {
+    C_AioNotify(JournalMetadata *_journal_metadata, Context *_on_safe)
+      : journal_metadata(_journal_metadata), on_safe(_on_safe) {
       journal_metadata->m_async_op_tracker.start_op();
     }
     virtual ~C_AioNotify() {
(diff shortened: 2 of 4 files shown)


trociny:
I had to update the patch to fix yet another race:
```
[ RUN      ] TestImageReplayer.StartInterrupted
/home/jenkins-build/build/workspace/ceph-pull-requests/src/common/AsyncOpTracker.cc: In function 'AsyncOpTracker::~AsyncOpTracker()' thread 7f9c8527b700 time 2016-09-29 08:45:49.670726
/home/jenkins-build/build/workspace/ceph-pull-requests/src/common/AsyncOpTracker.cc: 14: FAILED assert(m_pending_ops == 0)
 ceph version v11.0.0-2801-ge03a70f (e03a70fddad1b4796667ff990158aa6785f638d5)
 1: (ceph::__ceph_assert_fail(char const*, char const*, int, char const*)+0x8b) [0x7f9c9b8fee0b]
 2: (()+0x735332) [0x7f9c9b8dd332]
 3: (journal::JournalMetadata::~JournalMetadata()+0x56) [0x7f9c9b85cdf6]
 4: (journal::JournalMetadata::~JournalMetadata()+0x9) [0x7f9c9b85cfc9]
 5: (RefCountedObject::put()+0xe7) [0x7f9c9b5785f7]
 6: (journal::Journaler::~Journaler()+0x24) [0x7f9c9b8507f4]
 7: (librbd::Journal<librbd::ImageCtx>::get_tag_owner(librados::IoCtx&, std::string&, std::string*)+0x2b6) [0x7f9c9b79f5c6]
 8: (librbd::Journal<librbd::ImageCtx>::is_tag_owner(librados::IoCtx&, std::string&, bool*)+0x25) [0x7f9c9b79f7a5]
 9: (rbd::mirror::image_replayer::BootstrapRequest<librbd::ImageCtx>::handle_open_remote_image(int)+0x5a) [0x7f9c9b60ef4a]
 10: (Context::complete(int)+0x9) [0x7f9c9b576809]
 11: (rbd::mirror::image_replayer::OpenImageRequest<librbd::ImageCtx>::finish(int)+0x66) [0x7f9c9b72e6d6]
 12: (rbd::mirror::image_replayer::OpenImageRequest<librbd::ImageCtx>::handle_open_image(int)+0x65) [0x7f9c9b72eb55]
 13: (Context::complete(int)+0x9) [0x7f9c9b576809]
 14: (librbd::ImageState<librbd::ImageCtx>::complete_action_unlock(librbd::ImageState<librbd::ImageCtx>::State, int)+0x10d) [0x7f9c9b75dc1d]
 15: (librbd::ImageState<librbd::ImageCtx>::handle_open(int)+0xb4) [0x7f9c9b75dfb4]
 16: (Context::complete(int)+0x9) [0x7f9c9b576809]
 17: (ThreadPool::worker(ThreadPool::WorkThread*)+0xb65) [0x7f9c9b905c65]
 18: (ThreadPool::WorkThread::entry()+0x10) [0x7f9c9b906c20]
 19: (()+0x8182) [0x7f9c9ad6d182]
 20: (clone()+0x6d) [0x7f9c908044ad]
```

Althoug not sure about the approach. The actual problem is that m_async_op_tracker.wait_for_ops() is called in shut_down(), but there are methods (get_client, register_client) that do not require init, and they are used without init/shut_down.

May be a better approach is not to call C_NotifyUpdate in such methods if JournalMetadata was not previously initialized?

dillaman:
The reason why the "finish_op" calls are in the destructor was to ensure that their callback was completed before the journal could potentially transition state and destroy itself (had some issues early on). 

It does seem like a good idea to track the in-flight notification and delay whatever action initiated the notification until after the notification is sent. However, I would ensure that a -ETIMEDOUT error on the notify doesn't bubble up and fail the original command.

trociny:
> It does seem like a good idea to track the in-flight notification and delay whatever action initiated the notification until after the notification is sent. However, I would ensure that a -ETIMEDOUT error on the notify doesn't bubble up and fail the original command.

Then may be just ignore the error (see my updated patch)? That would be the same behavior that we had previously. I have split the initial patch into two patches, to separate this part, where there is agreement.

> The reason why the "finish_op" calls are in the destructor was to ensure that their callback was completed before the journal could potentially transition state and destroy itself (had some issues early on).

It looks very useful, but currently it does not work if we don't call shut_down before journal is destroyed (valid for the cases where init is not called). I fixed these failures in the second patch though it does not look like a good solution, as it breaks your rule.

Do you see a better option? May be force users always to call shut_down, regardless if init has been called? Or provide a public method to wait_for_ops, which the user should call?

dillaman:
@trociny I'd be fine w/ doing a sync wait_for_ops if not initialized -- hidden in Journaler.cc or JournalMetadata.cc

trociny:
@dillaman I tried to do a sync wait_for_ops if not initialized, but was not satisfied with the result. Actually it is not clear where to run this. It can't be done by a completion function, as it is run before the op contex is destroyed (so async_op_tracker's finish_op is not called yet). It should be called by the creator of Journaler, before destroying it.

So, in my opinion, the problem is that we use Context for operations, and its complete function calls callback fist and then only destroys the object. So the callback finishes before the operation is tracked in the destructor. I see several ways to resolve this:

1) Override Context's complete function, so callback is called after destruction.
2) Don't inherit from Context and just provide own complete method, that would do things in right order (Actually these ops are rather requests than contexts, so I like this more than 1).
3) Use JournalMethadata's get/put methods to ensure it is destroyed only after all op contexts are destroyed.

I updated the PR to implement (3), although I am fine with any solution. What do you think?

dillaman:
@trociny I'd prefer not to expand the the use of the reference counting since I'd like to eventually remove it. Let me take a deeper look and get back to you.

dillaman:
@trociny Did you hit a crash with one of the "GetXYZ" state machine classes where its "on_finish" callback resulted in the destruction of the Journaler which raced with its "finish_op" call?

dillaman:
@trociny Can you try something like this [1]? I've been running unittest_librbd in a loop for ~15 minutes w/o a crash (it crashed within a minute w/o this fix). I swapped out your second commit and kept your notify callback change.

[1] https://paste.fedoraproject.org/439069/41541147/

trociny:
@dillaman I thouth waiting in destructor would be risky, but apparently it works  -- I was not able to trigger an issue testing locally so far.

I replace my commit with yours in this PR.

dillaman:
retest this please

PR #11260, opened 2016-09-29
librbd: fix rollback if failed to disable mirroring for image

runsisi:
disable mirroring for demoted image stucks in disabling state

### DIFF
--- a/src/librbd/internal.cc
+++ b/src/librbd/internal.cc
@@ -2467,10 +2467,9 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force,
       return r;
     } else {
       bool rollback = false;
-      BOOST_SCOPE_EXIT_ALL(ictx, rollback) {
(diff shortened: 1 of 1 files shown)


src/librbd/internal.cc
```diff
@@ -2551,12 +2550,14 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force,
           }
         }
       }
-    }
 
-    r = mirror_image_disable_internal(ictx, force);
-    if (r < 0) {
-      return r;
+      r = mirror_image_disable_internal(ictx, force);
```
dillaman: I believe this was outside the block due to ```snap_lock``` being held.

runsisi: sorry, did not notice the lock :(
updated and moved the code needs the `snap_lock` into a block.
thanks.

PR #11262, opened 2016-09-29
Revert "osdc: After write try merge bh."

jcsp:
This reverts commit 1a48a8a2b222e41236341cb1241f0885a1b0b9d8.

Fixes: http://tracker.ceph.com/issues/17270

### DIFF
--- a/src/osdc/ObjectCacher.cc
+++ b/src/osdc/ObjectCacher.cc
@@ -1114,7 +1114,6 @@ void ObjectCacher::bh_write_commit(int64_t poolid, sobject_t oid,
       }
     }
 
-    list <BufferHead*> hit;
     // apply to bh's!
     for (map<loff_t, BufferHead*>::const_iterator p = ob->data_lower_bound(start);
 	 p != ob->data.end();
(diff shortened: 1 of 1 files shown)


dillaman:
lgtm

gregsfortytwo:
Can you please explain in the commit and/or ticket why it's bad? At first blush we certainly do want to merge bufferheads whenever possible...

dillaman:
@gregsfortytwo This loop [1] operates on bhs that might be merged together. e.g. if the first and second bhs can be merged, they will -- then the loop will advance to the now merged/destructed second bh and try to merge it.

[1] https://github.com/ceph/ceph/pull/11262/files#diff-9bcd2f7647a2bd574b6ebe6baf8e61b3L1166

gregsfortytwo:
Right, makes sense. We still want to do the merging though so we probably need a better data structure/algorithm instead of just tossing it out.
(I'm not 100% sure we're allowed to have non-merged adjacent buffers...)

gregsfortytwo:
Testing http://pulpito.ceph.com/gregf-2016-10-06_18:07:58-fs-greg-fs-testing-105---basic-mira/ looks pretty good; OSD failure (http://tracker.ceph.com/issues/17553), pjd failure due to missing qa-suite wip-getuid merge, 2xlibcephfs-java failures (http://tracker.ceph.com/issues/17548), 1 dead job.

I'd still prefer we do a better job merging though.

jcsp:
@gregsfortytwo you're right that we should add the behaviour back, but for the moment this needs to merge because the current code crashes.  I've created http://tracker.ceph.com/issues/17589 for the job of adding it back

PR #11263, opened 2016-09-29
os/bluestore: sloppy reshard boundaries to avoid spanning blobs

liewegas:
Make the extent map shard target size sloppy so that we can try to avoid
sharding boundaries that create spanning blobs.

### DIFF
--- a/src/os/bluestore/BlueStore.cc
+++ b/src/os/bluestore/BlueStore.cc
@@ -1488,9 +1488,10 @@ void BlueStore::ExtentMap::reshard(Onode *o)
     }
   }
   unsigned target = g_conf->bluestore_extent_map_shard_target_size;
+  unsigned slop = target * g_conf->bluestore_extent_map_shard_target_size_slop;
   unsigned extent_avg = bytes / extent_map.size();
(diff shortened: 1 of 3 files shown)


src/os/bluestore/BlueStore.cc
```diff
@@ -1521,7 +1523,10 @@ void BlueStore::ExtentMap::reshard(Onode *o)
       dout(20) << __func__ << " old shard end 0x" << std::hex << shard_end
 	       << std::dec << dendl;
     }
-    if (estimate && estimate + extent_avg > target) {
+    // disfavor shard boundaries that span a blob
+    bool would_span = (ep->logical_offset < max_blob_end) || ep->blob_offset;
```
somnathr: @liewegas I don't understand why it would span if ep->blob_offset is non zero ?
Could you also help me understand what is the difference between logical_offset and blob_offset within an extent ?

liewegas: blob_offset means this Extent points to the interior of the blob, not the left-hand side.  That means that if we put a shard boundary starting at ep->logical_offset, the blob would span the boundary.

PR #11264, opened 2016-09-29
os/bluestore: try to split blobs instead of spanning them

liewegas:
If a blob crosses a shard boundary, try to split it into pieces
instead of making it spanning.  This is only possible under
certain conditions (uncompressed, no in-flight writes, no unused
space, shard boundary aligned with chunk boundary).

Generally speaking this will reduce the number of blobs that end
up spanning and reduce the onode key size.

### DIFF
--- a/src/os/bluestore/BlueStore.cc
+++ b/src/os/bluestore/BlueStore.cc
@@ -1294,6 +1294,11 @@ void BlueStore::Blob::discard_unallocated()
       }
       pos += e.length;
     }
+    if (blob.can_prune_tail()) {
+      dirty_blob();
+      blob.prune_tail();
+      dout(20) << __func__ << " pruned tail, now " << blob << dendl;
+    }
   }
 }
 
--- a/src/os/bluestore/BlueStore.cc
+++ b/src/os/bluestore/BlueStore.cc
@@ -1564,7 +1564,7 @@ bool BlueStore::ExtentMap::update(Onode *o, KeyValueDB::Transaction t,
   return false;
 }
 
-void BlueStore::ExtentMap::reshard(Onode *o)
+void BlueStore::ExtentMap::reshard(Onode *o, uint64_t min_alloc_size)
 {
   // un-span all blobs
   auto p = spanning_blob_map.begin();
@@ -1701,7 +1701,8 @@ void BlueStore::ExtentMap::reshard(Onode *o)
 	  for (const auto& sh : shards) {
 	    if (bstart < sh.offset && bend > sh.offset) {
 	      uint32_t blob_offset = sh.offset - bstart;
-	      if (b->get_blob().can_split_at(blob_offset)) {
+	      if (b->get_blob().can_split_at(blob_offset) &&
+		  blob_offset % min_alloc_size == 0) {
 		dout(20) << __func__ << "    splitting blob, bstart 0x"
 			 << std::hex << bstart
 			 << " blob_offset 0x" << blob_offset
@@ -6206,7 +6207,7 @@ void BlueStore::_txc_write_nodes(TransContext *txc, KeyValueDB::Transaction t)
 	t->rmkey(PREFIX_OBJ, s.key);
       }
       o->extent_map.fault_range(db, 0, o->onode.size);
-      o->extent_map.reshard(o.get());
+      o->extent_map.reshard(o.get(), min_alloc_size);
       reshard = o->extent_map.update(o.get(), t, true);
       if (reshard) {
 	dout(20) << __func__ << " warning: still wants reshard, check options?"
--- a/src/os/bluestore/BlueStore.cc
+++ b/src/os/bluestore/BlueStore.cc
@@ -1071,6 +1071,50 @@ void BlueStore::BufferSpace::finish_write(uint64_t seq)
   cache->_audit("finish_write end");
 }
 
+void BlueStore::BufferSpace::split(size_t pos, BlueStore::BufferSpace &r)
+{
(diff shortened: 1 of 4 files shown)


src/os/bluestore/bluestore_types.h
```diff
@@ -579,6 +579,23 @@ struct bluestore_blob_t {
   int verify_csum(uint64_t b_off, const bufferlist& bl, int* b_bad_off,
 		  uint64_t *bad_csum) const;
 
+  bool can_prune_tail() const {
+    return
+      !extents.empty() &&
```
ifed01: can be removed since there is extents.size > 1

src/os/bluestore/BlueStore.cc
```diff
@@ -2026,6 +2158,51 @@ BlueStore::Extent *BlueStore::ExtentMap::set_lextent(
...
+      continue;
+    }
+    if (ep->end() <= pos) {
+      continue;
+    }
+    vector<bluestore_pextent_t> released;
+    if (ep->logical_offset < pos) {
+      // split extent
+      size_t left = pos - ep->logical_offset;
+      Extent *ne = new Extent(pos, 0, ep->length - left, rb);
```
somnathr: @liewegas this is giving compilation error in latest master after I cherry-picked. Extent constructor is taking BlobRef& b now.

liewegas:
rebsaed and fixed!

somnathr:
@liewegas Getting the following onode size (in avg) after almost 7 hours of writing.

Put( Prefix = O key = 0x7f80000000000000019390e2c2217262'd_data.10156b8b4567.00000000000275f9!='0xfffffffffffffffeffffffffffffffff001c1000'x' Value size = 626)
Put( Prefix = O key = 0x7f80000000000000019390e2c2217262'd_data.10156b8b4567.00000000000275f9!='0xfffffffffffffffeffffffffffffffff Value size = 663)

This is encouraging, combining all those recent sharding/blob changes I am getting a total of ~65% performance improvement.

855 bytes of pg_info is significant now , I am hoping for another 35-40% improvement when we cut it to 100 bytes.

src/os/bluestore/BlueStore.cc
```diff
@@ -1583,10 +1683,42 @@ void BlueStore::ExtentMap::reshard(Onode *o)
...
+		dout(20) << __func__ << "    splitting blob, bstart 0x"
+			 << std::hex << bstart
+			 << " blob_offset 0x" << blob_offset
+			 << std::dec << " " << *b << dendl;
+		b = split_blob(b, blob_offset, sh.offset);
+		// switch b to the new right-hand side, in case it
+		// *also* has to get split.
+		bstart += blob_offset;
+	      } else {
+		must_span = true;
```
ifed01: break the loop here?

src/os/bluestore/BlueStore.cc
```diff
@@ -2026,6 +2158,51 @@ BlueStore::Extent *BlueStore::ExtentMap::set_lextent(
...
+  BlobRef rb = onode->c->new_blob();
+  lb->split(blob_offset, rb.get());
+
+  for (auto ep = seek_lextent(pos);
+       ep != extent_map.end() && ep->logical_offset < end_pos;
+       ++ep) {
+    if (ep->blob != lb) {
+      continue;
+    }
+    if (ep->end() <= pos) {
```
ifed01: IMHO that's an overkill since seek_lextent takes care of it.

liewegas:
updated, thanks!

src/os/bluestore/BlueStore.cc
```diff
@@ -1382,6 +1426,62 @@ bool BlueStore::Blob::put_ref(
...
+  unsigned i = 0;
+  size_t left = blob_offset;
+  for (auto p = lb.extents.begin(); p != lb.extents.end(); ++p, ++i) {
+    if (p->length <= left) {
+      left -= p->length;
+      continue;
+    }
+    if (left) {
+      if (p->is_valid()) {
+	rb.extents.emplace_back(bluestore_pextent_t(p->offset + left,
```
ifed01: I'm worrying if we can have pextent with arbitrary (not aligned with min_alloc_size) boundaries. Such pextent might be eventually released and I doubt allocator handles this properly. Or I missed something? Anyway suggest to have corresponding test case in ceph_test_objectstore

liewegas:
good call, added a patch to prevent that.

src/os/bluestore/BlueStore.cc
```diff
@@ -1701,7 +1701,8 @@ void BlueStore::ExtentMap::reshard(Onode *o)
 	  for (const auto& sh : shards) {
 	    if (bstart < sh.offset && bend > sh.offset) {
 	      uint32_t blob_offset = sh.offset - bstart;
-	      if (b->get_blob().can_split_at(blob_offset)) {
+	      if (b->get_blob().can_split_at(blob_offset) &&
+		  blob_offset % min_alloc_size == 0) {
```
ifed01: And another concern - are there any means that ensure blob_offset to be aligned with min_alloc_size often enough. If no and blob_offset is pretty arbitrary then split blob functionality makes no sense...

liewegas:
In practice, with block devices, the extent boundaries will fall on the 
alignment boundaries.  Probably less true with CephFS objects.  We 
probably need to make this capable of splitting extents too.  I think we 
can do that later, though.. the main piece of infrastructure I'm most 
interested here is the split_blob part.  reshard() can be improved as we 
go.

ifed01:
LGTM

PR #11269, opened 2016-09-30
rgw: multipart upload copy

yehudasa:
This builds on top of Javier's branch. Adds the following:
 - read source object in chunks
 - cleanups, parsing changes
 - versioning support

### DIFF
--- a/doc/radosgw/s3.rst
+++ b/doc/radosgw/s3.rst
@@ -70,7 +70,7 @@ The following table describes the support status for current Amazon S3 functiona
 +---------------------------------+-----------------+----------------------------------------+
 | **Copy Object**                 | Supported       |                                        |
 +---------------------------------+-----------------+----------------------------------------+
-| **Multipart Uploads**           | Supported       | (missing Copy Part)                    |
+| **Multipart Uploads**           | Supported       |                                        |
 +---------------------------------+-----------------+----------------------------------------+
 
 Unsupported Header Fields
--- a/src/rgw/rgw_common.h
+++ b/src/rgw/rgw_common.h
@@ -1357,8 +1357,6 @@ struct req_state {
   string req_id;
   string trans_id;
 
-  const char *copy_source;
-
   req_state(CephContext* _cct, RGWEnv* e, RGWUserInfo* u);
   ~req_state();
 };
--- a/src/rgw/rgw_common.h
+++ b/src/rgw/rgw_common.h
@@ -1357,6 +1357,8 @@ struct req_state {
   string req_id;
   string trans_id;
 
+  const char *copy_source;
+
   req_state(CephContext* _cct, RGWEnv* e, RGWUserInfo* u);
   ~req_state();
 };
--- a/src/rgw/rgw_op.cc
+++ b/src/rgw/rgw_op.cc
@@ -2484,6 +2484,27 @@ void RGWDeleteBucket::execute()
 
 int RGWPutObj::verify_permission()
 {
+  if (copy_source) {
+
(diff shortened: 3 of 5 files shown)


jmunhoz:
@yehudasa the new patches look good to me. Bear in mind I am not very familiar with the versioning code under the hood but the code look good on my side. The branch is compiling and it works with my manual tests too. I got some some auth errors (forbidden and sig not match) using the s3-tests but I think they could be related to my testing environment and not the feature.

yehudasa:
@jmunhoz which tests failed exactly, are these aws4?

jmunhoz:
@yehudasa  s3-tests in commit 866411b1acaffc8645bc643470c658f5b34015ac. aws2 and aws4 are failing. As mentioned it is very possible my s3-test environment is broken now after a hard pulling. Are they working on your side?

yehudasa:
@jmunhoz yes, but I tested it against a branch that consisted of some other fixes. Also, there is another broken test that will fail anyway, and some other tests that need special backend config.

jmunhoz:
@yehudasa I set up a new s3-tests environment and ran some tests. As mentioned, it works with manual testing (http://tracker.ceph.com/attachments/download/2426/multipart-uploads-cp-test.py) between allowed buckets, etc. but if I use s3-tests all related test case raise errors on aws4 only. I uploaded a text file with the errors by category. It is available here:

http://tracker.ceph.com/attachments/download/2542/20160930-s3-tests-debug.txt

On my side, and using the s3-tests branch with commit 866411b1acaffc8645bc643470c658f5b34015ac I see errors related to versioning, lifecycle and acl granting.

It is interesting the multipart copy errors show all errors related to auth (404 forbidden) and not existent buckets. I wonder if the versioning and acl support could be failing and some test cases related to multipart could be failing due to not allowed permissions, etc. Does it make sense?

yehudasa:
@jmunhoz there are a few known issues in that branch. Also, you need to run nose with `-a !fails_on_rgw`, so that it filters out other tests that are not supposed to pass. I'm preparing a branch with fixes to some of the known issues.

yehudasa:
@jmunhoz see this branch: https://github.com/yehudasa/ceph/tree/wip-multipart-uploads-cp-2

jmunhoz:
@yehudasa with the new branch the only multipart test case expecting pass and failing is test_multipart_copy_versioned under aws4

yehudasa:
@jmunhoz tested it now, adding your boto aws4 compat branch: bugfix-16463-presigned-url, still see it failing, but not sure it's not a boto bug:
boto created the following canonical header:
`
...
x-amz-copy-source:yehuda-alerinj78zh4llqssrivq5-1/foo
x-amz-copy-source-range:bytes=0-5242879
...
`
whereas rgw created this:
`
x-amz-copy-source:yehuda-alerinj78zh4llqssrivq5-1/foo
x-amz-copy-source-range:bytes=0-5242879
`

It looks like boto is sorting by the whole strings, whereas we do it by the key names. Which is the correct way? In any case, this is not a regression due to the new code.

yehudasa:
@jmunhoz after reading the aws4 spec, and having this test failing also when running it against aws, it really looks like boto bug.

jmunhoz:
@yehudasa I think it could be a boto bug. I think the right case is sorting by the key names. As you mention it is coherent with the aws behaviour. In the minio-go implementation they are sorting by key too https://github.com/minio/minio-go/blob/master/request-signature-v4.go#L126-L134

cbodley:
looks good to me - does it need a run through teuthology?

yehudasa:
@cbodley that can [almost] never hurt

PR #11275, opened 2016-09-30
os/bluestore: drop redundant call of get_blob

xiexingguo:


liewegas:
I made the audit code compile out because the _audit function is called all over the place and won't be inlined, so even if it returns immediately there will be some overhead...

xiexingguo:
Okay, dropped that change.

PR #11277, opened 2016-09-30
cmake: fix compilation with distcc and other compiler wrappers

asheplyakov:
When building with ccache, distcc, and other compiler wrappers (such as STLFilt):

CC='ccache gcc' CXX='ccache g++' cmake /path/to/ceph
make

python modules fail to compile since distutils try to execute the wrapper itself
without specifying the actual compiler. Although cmake has a special magic switch
for compiling with ccache (cmake -DWITH_CCACHE=ON) other tools (distcc) are
not supported, and specifying the compiler as

CC=/whatever/compiler/is

used to work for decades, and it's a good idea to keep it working

### DIFF
--- a/cmake/modules/Distutils.cmake
+++ b/cmake/modules/Distutils.cmake
@@ -32,15 +32,27 @@ endfunction(distutils_install_module)
 function(distutils_add_cython_module name src)
   get_property(compiler_launcher GLOBAL PROPERTY RULE_LAUNCH_COMPILE)
   get_property(link_launcher GLOBAL PROPERTY RULE_LAUNCH_LINK)
-  set(PY_CC \"${compiler_launcher} ${CMAKE_C_COMPILER}\")
-  set(PY_CXX \"${compiler_launcher} ${CMAKE_CXX_COMPILER}\")
-  set(PY_LDSHARED \"${link_launcher} ${CMAKE_C_COMPILER} -shared\")
+  # When using ccache CMAKE_C_COMPILER is ccache executable absolute path
+  # and the actual C compiler is CMAKE_C_COMPILER_ARG1.
+  # However with a naive
+  # set(PY_CC ${compiler_launcher} ${CMAKE_C_COMPILER} ${CMAKE_C_COMPILER_ARG1})
+  # distutils tries to execve something like "/usr/bin/cmake gcc" and fails.
+  # Removing the leading whitespace from CMAKE_C_COMPILER_ARG1 helps to avoid
+  # the failure.
+  string(REGEX REPLACE "^[ ]+(.+)[ ]*$" \\1 _actual_c_compiler "${CMAKE_C_COMPILER_ARG1}")
+  string(REGEX REPLACE "^[ ]+(.+)[ ]*$" \\1 _actual_cxx_compiler "${CMAKE_CXX_COMPILER_ARG1}")
+  # Note: no quotes, otherwise distutils will execute "/usr/bin/ccache gcc"
+  # CMake's implicit conversion between strings and lists is wonderful, isn't it?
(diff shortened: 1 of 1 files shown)


tchaikov:
@asheplyakov have you tried 
```
 cmake -DWITH_CCACHE=ON ..
```
?

asheplyakov:
@tchaikov

First of all there are quite a number of useful tools wrapping the compiler (distcc, STLFilt, etc), and (obviously) cmake don't have magic switches for these.

Secondly (and most importantly) `CC=/whatever/the/compiler/is` used to work for decades, and it's a good idea to keep it working.

liewegas:
@tchaikov ?

cmake/modules/Distutils.cmake
```diff
@@ -32,15 +32,27 @@ endfunction(distutils_install_module)
...
-  set(PY_CXX \"${compiler_launcher} ${CMAKE_CXX_COMPILER}\")
-  set(PY_LDSHARED \"${link_launcher} ${CMAKE_C_COMPILER} -shared\")
+  # When using ccache CMAKE_C_COMPILER is ccache executable absolute path
+  # and the actual C compiler is CMAKE_C_COMPILER_ARG1.
+  # However with a naive
+  # set(PY_CC ${compiler_launcher} ${CMAKE_C_COMPILER} ${CMAKE_C_COMPILER_ARG1})
+  # distutils tries to execve something like "/usr/bin/cmake gcc" and fails.
+  # Removing the leading whitespace from CMAKE_C_COMPILER_ARG1 helps to avoid
+  # the failure.
+  string(REGEX REPLACE "^[ ]+(.+)[ ]*$" \\1 _actual_c_compiler "${CMAKE_C_COMPILER_ARG1}")
```
tchaikov: nit, the scopes of `_actual_c_compiler` and `_actual_cxx_compiler` are limited in the function, so no need to prefix them with `_`. naming the local variables using the lower case chars would suffice. also, i think it might be easier to understand if you could
```cmake
string(STRIP "${CMAKE_C_COMPILER_ARG1}" c_compiler_arg1)
```
and set `PY_CC`  like
```cmake
set(PY_CC ${compiler_launcher} ${CMAKE_C_COMPILER} ${c_compiler_arg1})
```

i can hardly get my head around `${CMAKE_C_COMPILER} ${_actual_c_compiler}` even with your comments. but it would be easier to digest if one just wants to strip off the spaces around the compiler flags before passing it to the compiler.

cmake/modules/Distutils.cmake
```diff
@@ -32,15 +32,27 @@ endfunction(distutils_install_module)
 function(distutils_add_cython_module name src)
   get_property(compiler_launcher GLOBAL PROPERTY RULE_LAUNCH_COMPILE)
   get_property(link_launcher GLOBAL PROPERTY RULE_LAUNCH_LINK)
-  set(PY_CC \"${compiler_launcher} ${CMAKE_C_COMPILER}\")
-  set(PY_CXX \"${compiler_launcher} ${CMAKE_CXX_COMPILER}\")
-  set(PY_LDSHARED \"${link_launcher} ${CMAKE_C_COMPILER} -shared\")
+  # When using ccache CMAKE_C_COMPILER is ccache executable absolute path
```
amitkumar50: @asheplyakov (NIT) Please consider to keep PR title < 50 characters for future PRs.

gregsfortytwo:
Do we still want this PR in?

jenkins test this please

tchaikov:
i am closing this PR due to lack of activity, and it is superseded by #24605 which incorporates the proposed change. and drop the `PY_LDSHARED` change, as we use the cythonized code is C not C++.

PR #11278, opened 2016-09-30
rgw: obsolete 'radosgw-admin period prepare' command

gaurav36:
All that the 'radosgw-admin period prepare' command does is create an
empty period and write it to rados. When we added the staging period
(see 'period get --staging'), uncommitted changes were written to that
instead - so the 'period prepare' command became obsolete. It should be
removed!

Fixes: http://tracker.ceph.com/issues/17387

Reported-by: Casey Bodley <cbodley@redhat.com>
Signed-off-by: Gaurav Kumar Garg <garg.gaurav52@gmail.com>

### DIFF
--- a/src/rgw/rgw_admin.cc
+++ b/src/rgw/rgw_admin.cc
@@ -75,7 +75,6 @@ void _usage()
   cout << "  object rm                  remove object\n";
   cout << "  object unlink              unlink object from bucket index\n";
   cout << "  objects expire             run expired objects cleanup\n";
-  cout << "  period prepare             prepare a new period\n";
   cout << "  period delete              delete a period\n";
   cout << "  period get                 get period info\n";
   cout << "  period get-current         get current period info\n";
@@ -381,7 +380,6 @@ enum {
   OPT_REALM_SET,
   OPT_REALM_DEFAULT,
   OPT_REALM_PULL,
-  OPT_PERIOD_PREPARE,
   OPT_PERIOD_DELETE,
   OPT_PERIOD_GET,
   OPT_PERIOD_GET_CURRENT,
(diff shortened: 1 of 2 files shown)


src/rgw/rgw_admin.cc
```diff
@@ -558,8 +557,10 @@ static int get_cmd(const char *cmd, const char *prev_cmd, const char *prev_prev_
     if (strcmp(cmd, "list") == 0)
       return OPT_BI_LIST;
   } else if (strcmp(prev_cmd, "period") == 0) {
-    if (strcmp(cmd, "prepare") == 0)
-      return OPT_PERIOD_PREPARE;
+    if (strcmp(cmd, "prepare") == 0) {
+      cerr <<"command is obsolete; please check usage and/or man page" <<std::endl;
+      return 0;
+    }
```
cbodley: i don't think it's necessary to keep this message. could you also remove `OPT_PERIOD_PREPARE` from the enum?

gaurav36: @cbodley  Thanks.
I will remove OPT_PERIOD_PREPARE in next patch. 
I think if we remove the message then in next release if user execute period prepare command then he might wonder about period prepare command, means what happened to this command. Should we put any other message or  completely remove this message ("  command is obsolete; please check usage and/or man page")

cbodley: i appreciate you wanting to document the deprecation, but the reason i don't think it's necessary is that we've never documented this command (outside of `radosgw-admin --help`) as part of the [multisite configuration process](http://docs.ceph.com/docs/master/radosgw/multisite/)

gaurav36: @cbodley  thanks, 
right, i will do it in next patch set.

gaurav36:
@cbodley  test case failure for this PR seems to be spurious failure. I saw same test failure on some other PR as well which have nothing to do with code change.

src/rgw/rgw_admin.cc
```diff
@@ -558,8 +556,9 @@ static int get_cmd(const char *cmd, const char *prev_cmd, const char *prev_prev_
     if (strcmp(cmd, "list") == 0)
       return OPT_BI_LIST;
   } else if (strcmp(prev_cmd, "period") == 0) {
-    if (strcmp(cmd, "prepare") == 0)
-      return OPT_PERIOD_PREPARE;
+    if (strcmp(cmd, "prepare") == 0) {
+      return 0;
+    }
```
cbodley: can remove this block entirely now

gaurav36: @cbodley  ahhh, i missed it.

cbodley:
thank you!

PR #11280, opened 2016-09-30
rgw: fixes for virtual hosting of buckets

cbodley:
this brings back the fix from @robbat2 that was reverted in https://github.com/ceph/ceph/pull/10873, with an additional check that disables the virtual hosting feature if no hostnames were configured. this prevents default configurations (such as teuthology) from treating all requests as virtual hosted buckets

also updated the docs - a preview should eventually be available at http://docs.ceph.com/docs/wip-17440/radosgw/s3/commons/

### DIFF
--- a/src/rgw/rgw_rest.cc
+++ b/src/rgw/rgw_rest.cc
@@ -208,8 +208,7 @@ void rgw_rest_init(CephContext *cct, RGWRados *store, RGWZoneGroup& zone_group)
   hostnames_set.insert(cct->_conf->rgw_dns_name);
   hostnames_set.insert(zone_group.hostnames.begin(), zone_group.hostnames.end());
   hostnames_set.erase(""); // filter out empty hostnames
-  string s;
-  ldout(cct, 20) << "RGW hostnames: " << std::accumulate(hostnames_set.begin(), hostnames_set.end(), s) << dendl;
+  ldout(cct, 20) << "RGW hostnames: " << hostnames_set << dendl;
   /* TODO: We should have a sanity check that no hostname matches the end of
    * any other hostname, otherwise we will get ambigious results from
    * rgw_find_host_in_domains.
@@ -223,8 +222,7 @@ void rgw_rest_init(CephContext *cct, RGWRados *store, RGWZoneGroup& zone_group)
   hostnames_s3website_set.insert(cct->_conf->rgw_dns_s3website_name);
   hostnames_s3website_set.insert(zone_group.hostnames_s3website.begin(), zone_group.hostnames_s3website.end());
   hostnames_s3website_set.erase(""); // filter out empty hostnames
-  s.clear();
-  ldout(cct, 20) << "RGW S3website hostnames: " << std::accumulate(hostnames_s3website_set.begin(), hostnames_s3website_set.end(), s) << dendl;
+  ldout(cct, 20) << "RGW S3website hostnames: " << hostnames_s3website_set << dendl;
   /* TODO: we should repeat the hostnames_set sanity check here
    * and ALSO decide about overlap, if any
    */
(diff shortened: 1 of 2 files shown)


src/rgw/rgw_rest.cc
```diff
@@ -1846,10 +1846,12 @@ int RGWREST::preprocess(struct req_state *s, RGWClientIO* cio)
     // As additional checks:
     // - if the Host header is an IP, we're using path-style access without DNS
     // - Also check that the Host header is a valid bucket name before using it.
+    // - Don't enable virtual hosting if no hostnames are configured
     if (subdomain.empty()
         && (domain.empty() || domain != info.host)
         && !looks_like_ip_address(info.host.c_str())
-        && RGWHandler_REST::validate_bucket_name(info.host)) {
+        && RGWHandler_REST::validate_bucket_name(info.host) == 0
+        && !hostnames_set.empty()) {
```
robbat2: This should check hostnames_s3website_set as well. If either of them are non-empty, it should be ok (should be very rare, but want to check anyway).

cbodley: added

robbat2:
I ran into one related impact from this, that should be clarified & documented.
RGW's legacy behavior differs from S3 behavior, and s3-tests was designed around the RGW legacy behavior: using the path calling format, and assuming that the hostname it was connecting to was also configured as the rgw dns name or zonegroup hostname.

Thus, in some cases, we know get a lot more false-positive s3-tests breakage.

```
GET /foobar/ HTTP/1.1
Host: non-existing-ceph-buckettest.example.com
```

AWS-S3 treats this as bucket {{non-existing-ceph-buckettest.example.com}}, and returns:
```
<?xml version="1.0" encoding="UTF-8"?>
<Error>
<Code>NoSuchBucket</Code>
<Message>The specified bucket does not exist</Message>
<BucketName>non-existing-ceph-buckettest.example.com</BucketName>
<RequestId>D1F9F7DCD9D79A11</RequestId>
<HostId>1RGXmTwGt4w2R+BV/xiuJsmJoIwVZsjhwEBH2Qrqh4URbV0zlIvX1+9KczHCCKHnhtV7Mhh7Ps8=</HostId>
</Error>
```

RGW, prior to my patching, treats this as bucket {{foobar}}, and returns:
```
<?xml version="1.0" encoding="UTF-8"?>
<Error>
<Code>NoSuchBucket</Code>
<BucketName>foobar</BucketName>
<RequestId>tx0000000000000000025fb-0057f2e73e-fa74-default</RequestId>
<HostId>fa74-default-default</HostId>
</Error>
```

RGW, with patch, now matches AWS-S3:
```
<?xml version="1.0" encoding="UTF-8"?>
<Error>
<Code>NoSuchBucket</Code>
<BucketName>non-existing-ceph-buckettest.example.com</BucketName>
<RequestId>tx0000000000000000c6085-0057f2e721-10bb00c9-default</RequestId>
<HostId>10bb00c9-default-default</HostId>
</Error>
```

cbodley:
oops, i was cleaning up branches and nuked this one. reopened

robbat2:
@cbodley @yehudasa what happened to having this merged in v10.2.4? I'm writing documentation for StaticSites, and it depends on having this merged.

oritwas:
I scheduled a teuthology run: http://pulpito.ceph.com/owasserm-2016-10-27_16:41:06-rgw-wip-17440---basic-mira/

oritwas:
jenkins test this please

oritwas:
new run (previous runs had enviorment problems): http://pulpito.ceph.com/owasserm-2016-10-31_18:27:38-rgw-wip-orit-17440---basic-mira/

cbodley:
thanks @oritwas !

PR #11283, opened 2016-09-30
mds: include legacy client fsid in FSMap print

jcsp:


gregsfortytwo:
http://pulpito.ceph.com/gregf-2016-10-06_18:07:58-fs-greg-fs-testing-105---basic-mira/ looks pretty good; OSD failure (http://tracker.ceph.com/issues/17553), pjd failure due to missing qa-suite wip-getuid merge, 2xlibcephfs-java failures (http://tracker.ceph.com/issues/17548), 1 dead job.

PR #11284, opened 2016-09-30
mon/OSDMonitor: encode full osdmaps with features all OSDs can understand

liewegas:
Also prompt users to set the require_*_osds flags once tey finish their upgrades.

### DIFF
--- a/src/mon/OSDMonitor.cc
+++ b/src/mon/OSDMonitor.cc
@@ -3075,6 +3075,25 @@ void OSDMonitor::get_health(list<pair<health_status_t,string> >& summary,
       }
     }
 
+    // warn about upgrade flags that can be set but are not.
+    if ((osdmap.get_up_osd_features() & CEPH_FEATURE_SERVER_KRAKEN) &&
+	!osdmap.test_flag(CEPH_OSDMAP_REQUIRE_KRAKEN)) {
+      string msg = "all OSDs are running kraken or later but the"
+	" 'require_kraken_osds' osdmap flag is not set";
+      summary.push_back(make_pair(HEALTH_WARN, msg));
+      if (detail) {
+	detail->push_back(make_pair(HEALTH_WARN, msg));
+      }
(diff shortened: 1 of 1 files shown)


jdurgin:
We should discuss automatically setting the flags - it's silly to add extra steps for users to jump through for every upgrade. Perhaps some heuristic like 'after all up osds are kraken for 24h, require the kraken feature bit'.

liewegas:
Yeah.  It's a bit tricky because we need to avoid situations where all the un-upgraded osds crash or are stopped and only upgraded osds are up.  It could be gated on the cluster being healthy, perhaps... or all in osds being up.

src/mon/OSDMonitor.cc
```diff
@@ -1220,13 +1220,25 @@ void OSDMonitor::encode_pending(MonitorDBStore::TransactionRef t)
     }
   }
 
+  // determine appropriate features
+  uint64_t features = mon->quorum_features;
+  if (!newmap.has_flag(CEPH_ODSMAP_REQUIRE_JEWEL)) {
```
tchaikov: s/ODS/OSD/

tchaikov:
lgtm.

@liewegas we need run this thru at least one upgrade test, and probably you could include https://github.com/tchaikov/ceph/commit/31233d6b2a4666454a24f7dc2ea9cf6eb9969741 in your PR? and try to remove 'failed to encode map' from (at least some of) the upgrade tests. i will take care of this once this change is merged.

tchaikov:
do we want to backport this change to jewel?

liewegas:
Yeah, I think we should backport.

tchaikov:
@liewegas how's test going?

liewegas:
trying to get it through upgrade/jewel-x

liewegas:
http://pulpito.ceph.com/sage-2016-10-15_14:11:58-upgrade:jewel-x-wip-sage-testing---basic-smithi/

PR #11285, opened 2016-09-30
ceph.spec: fix ceph-mgr version requirement

liewegas:


dmick:
Only one that didn't mention epoch.  Looks right to me.  (And epoch is set to 1 in the specfile, in case anyone else besides me didn't realize that.)

ktdreyer:
:+1:

PR #11286, opened 2016-09-30
rgw: don't fail if lost race when setting acls

yehudasa:
Fixes: http://tracker.ceph.com/issues/16930

When concurrently setting acls on object/bucket, we could lose in a race.
Instead of retry, just return success (same effect as if we won and then
other writer overwrote us).


prallabh:
@yehudasa Is this also related to this http://tracker.ceph.com/issues/14208#note-5. I have seen a similar failure during bucket creation on one of our older setups running Jewel 10.2.2, please find the logs that I have attached as part of that comment.

PR #11254, opened 2016-09-29
mds/MDCache: kill a comipler warning

xiexingguo:
/home/jenkins-build/build/workspace/ceph-pull-requests/src/mds/MDCache.cc: In member function ‘void MDCache::request_cleanup(MDRequestRef&)’:
/home/jenkins-build/build/workspace/ceph-pull-requests/src/mds/MDCache.cc:9090:8: warning: unused variable ‘was_replay’ [-Wunused-variable]
   bool was_replay = mdr->client_request && mdr->client_request->is_replay();

Signed-off-by: xie xingguo <xie.xingguo@zte.com.cn>

PR #11287, opened 2016-10-01
mds: fix a compiler warning about unused-variable

runsisi:
#11078 has removed the use of this variable, so remove it should be safe.


runsisi:
has been fixed by #11254 , closing this.

PR #11294, opened 2016-10-03
qa/workunits/rbd: set image-meta on primary image and wait it is replicated

trociny:
After recently added image metadata replication it is not possible any
more to update it on non-primary image.


trociny:
@dillaman

PR #11295, opened 2016-10-03
os/bluestore: make cache settings process-wide

liewegas:
...and a bunch of other stuff, including instrumenting cache onodes, extents, blobs, buffers, bytes.

### DIFF
--- a/src/os/bluestore/BlueStore.h
+++ b/src/os/bluestore/BlueStore.h
@@ -72,8 +72,11 @@ enum {
   l_bluestore_compressed,
   l_bluestore_compressed_allocated,
   l_bluestore_compressed_original,
+  l_bluestore_onodes,
   l_bluestore_onode_hits,
   l_bluestore_onode_misses,
+  l_bluestore_buffers,
+  l_bluestore_buffer_bytes,
   l_bluestore_buffer_hit_bytes,
   l_bluestore_buffer_miss_bytes,
   l_bluestore_write_big,
@@ -714,6 +717,9 @@ public:
 
     virtual void trim(uint64_t onode_max, uint64_t buffer_max) = 0;
 
+    virtual void add_stats(uint64_t *onodes, uint64_t *buffers,
+			   uint64_t *bytes) = 0;
+
 #ifdef DEBUG_CACHE
     virtual void _audit(const char *s) = 0;
 #else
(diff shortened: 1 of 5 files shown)


ifed01:
Wondering if 
shared_blob_set(MAX(16,
 			g_conf->bluestore_onode_cache_size *
 			g_conf->bluestore_shared_blob_hash_table_size_ratio)),
should be fixed as well?

src/os/bluestore/BlueStore.h
```diff
@@ -512,10 +521,24 @@ class BlueStore : public ObjectStore,
     uint8_t  blob_depth;              /// blob overlapping count
     BlobRef blob;                     ///< the blob with our data
 
-    explicit Extent() {}
-    explicit Extent(uint32_t lo) : logical_offset(lo) {}
+    /// ctor for lookup only
+    explicit Extent(uint32_t lo) : logical_offset(lo) { }
+    /// ctor for delayed intitialization (see decode_some())
+    explicit Extent(Cache *cache) {
```
ifed01: IMHO specific blob assignment method that also increments cache counter(s) is more transparent and error prone, e.g.
void Extent::assign_blob(BlobRef b) {
  if(blob)
    blob->shared_blob->bc.cache->rm_extent();
  blob = b;
  if( blob )
    blob->shared_blob->bc.cache->add_extent();
}

src/os/bluestore/BlueStore.h
```diff
@@ -524,22 +524,26 @@ class BlueStore : public ObjectStore,
...
-      cache->add_extent();
+    explicit Extent() {
     }
     /// ctor for general usage
     Extent(uint32_t lo, uint32_t o, uint32_t l, uint8_t bd, BlobRef& b)
-      : logical_offset(lo), blob_offset(o), length(l), blob_depth(bd), blob(b) {
-      if (blob) {
-	blob->shared_blob->bc.cache->add_extent();
-      }
+      : logical_offset(lo), blob_offset(o), length(l), blob_depth(bd) {
```
ifed01: suggest to add ",blob(NULL)" here to avoid potential issues with assign_blob evolution

src/os/bluestore/BlueStore.h
```diff
@@ -524,22 +524,26 @@ class BlueStore : public ObjectStore,
...
+      assign_blob(b);
     }
     ~Extent() {
       if (blob) {
 	blob->shared_blob->bc.cache->rm_extent();
       }
     }
 
+    void assign_blob(const BlobRef& b) {
+      blob = b;
```
ifed01: assert(!blob) here to catch if someday we call assign_blob twice for some reason?

PR #11298, opened 2016-10-03
mgr/MgrClient: fix ms_handle_reset

liewegas:
Return false because we don't handle the reset.


dillaman:
In-progress librbd test suite run is looking more stable with this PR applied -- still seeing the -ENOENT error against objects that should exist.

PR #11299, opened 2016-10-03
mgr: implement con reset handling

liewegas:


jcsp:
Will run the (very small) `mgr` suite on this in the spirit of starting as we mean to go on!

liewegas:
https://www.youtube.com/watch?v=34QXu__ywY8

jcsp:
The lab had a long wait so I tested this locally.

PR #11302, opened 2016-10-04
doc: fix librados example programs

asheplyakov:
basically s/rados_wait_for_(safe|complete)/rados_aio_wait_for_\1/g


jdurgin:
good catch!

PR #11306, opened 2016-10-04
librbd: memory leak in MirroringWatcher::notify_image_updated

trociny:


trociny:
@dillaman

PR #11307, opened 2016-10-04
mgr: squash compiler warnings

jcsp:


wjwithagen:
@jcsp 
LGTM 
(or atleast Clang thinks so)

PR #11308, opened 2016-10-04
mgr/MgrMap: initialize all fields

liewegas:
Fixes: http://tracker.ceph.com/issues/17492


jcsp:
(Tested this locally)

PR #11309, opened 2016-10-04
include/slab_containers: heap-friendly containers

liewegas:


yuyuyu101:
@allensamuels do we really need get_my_actual_allocator? Or do we need reserve if using slab_*? I think reserve is target to original stl structure to reduce extra memory allocation. But if using slab, users actually can specify when construct to specify stacksize/heapsize. whether reserve or not, the allocation number is fixed.

allensamuels:
The reserve functionality of slab_* is purely an optimization. My expectation is that only place where it’s likely to be used is upon deserialization.

Sage has speculated that with the new boost, we could use flat_map and perhaps be even better in many situations.


Allen Samuels
SanDisk |a Western Digital brand

liewegas:
see newer mempool variants

PR #11310, opened 2016-10-04
hammer: osd: adjust scrub boundary to object without SnapSet

dzafman:


dzafman:
The last commit https://github.com/ceph/ceph/pull/11310/commits/ab9a0c42c433936611b6c7b3d06a03658ff32e0b is a cherry-pick from jewel.  It is only needed by Jewel and Hammer.

dzafman:
@athanatos Fix commit added based on first rados suite run.  Repeat of individual job passed with the fix.  New runs started.

src/osd/PG.cc
```diff
@@ -4159,6 +4159,11 @@ void PG::chunky_scrub(ThreadPool::TPHandle &handle)
           return;
         }
 
+        // Account for http://tracker.ceph.com/issues/17491
+        if (!scrubber.primary_scrubmap.objects.empty() &&
+            scrubber.primary_scrubmap.objects.rbegin()->first == scrubber.end)
+          scrubber.primary_scrubmap.objects.erase(scrubber.end);
```
athanatos: Yep!

dzafman:
@athanatos I should in some way indicate that the last commit is a cherry-pick from the _jewel_ branch which is unusual.

athanatos:
@dzafman Probably sufficient to add "From jewel commit: " in front of what -x usually adds.

dzafman:
@liewegas @athanatos I'm satisfied with these 2 final runs validated this even with 1 dead job.

No failures
http://pulpito.ceph.com/dzafman-2016-10-07_13:51:45-rados-wip-zafman-testing-hammer-distro-basic-smithi

One dead job because space ran out with 3 of 6 out OSDs causing backfull_toofull in this run.
http://pulpito.ceph.com/dzafman-2016-10-07_13:43:59-rados-wip-zafman-testing-hammer-distro-basic-smithi

dzafman:
@liewegas Ready for review and merge after master and jewel versions.

athanatos:
@liewegas The workaround in this branch in your opinion should be adequate, right?
@dzafman Have we been running this through the suite?

liewegas:
It's ugly, but I think it's adequate.

dzafman:
@athanatos Tests have been run.  Previous comments show the exact run locations before I removed needs-test.

dzafman:
It has been decided that this change will NOT be merged to Hammer.

PR #11313, opened 2016-10-04
rbd: recognize lock_on_read option

idryomov:
Signed-off-by: Ilya Dryomov <idryomov@gmail.com>

### DIFF
--- a/doc/man/8/rbd.rst
+++ b/doc/man/8/rbd.rst
@@ -474,6 +474,9 @@ Mapping (per block device) options:
 
 * queue_depth=x - queue depth (since 4.2, default is 128 requests).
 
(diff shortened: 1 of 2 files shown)


dillaman:
ltgm

@idryomov do we have a teuthology test that covers the "normal" iSCSI krbd configuration?

idryomov:
No, in the "normal" iSCSI configuration images are mapped through multipath, so nothing iSCSI-specific at the moment.  I was planning to add a simple lock_for_read test case later.

ktdreyer:
@idryomov would you please file a ticket at tracker.ceph.com so that we ensure this gets backported to Jewel?

PR #11316, opened 2016-10-04
mds/SessionMap.cc:put classes in unnamed namespace

stiopaa1:
So they have internal linkage.

### DIFF
--- a/src/mds/SessionMap.cc
+++ b/src/mds/SessionMap.cc
@@ -28,7 +28,7 @@
 #undef dout_prefix
 #define dout_prefix *_dout << "mds." << rank << ".sessionmap "
 
-
+namespace {
 class SessionMapIOContext : public MDSIOContextBase
 {
   protected:
@@ -38,6 +38,7 @@ class SessionMapIOContext : public MDSIOContextBase
     explicit SessionMapIOContext(SessionMap *sessionmap_) : sessionmap(sessionmap_) {
       assert(sessionmap != NULL);
     }
+};
 };
 
 void SessionMap::register_perfcounters()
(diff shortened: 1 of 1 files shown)


gregsfortytwo:
What's the advantage here? Does it reduce the size of binaries, or speed up build times, or something?

stiopaa1:
@gregsfortytwo 
https://gcc.gnu.org/wiki/Visibility

jcsp:
@stiopaa1 that link doesn't mention anything about namespace blocks...?

gregsfortytwo:
http://en.cppreference.com/w/cpp/language/namespace#Unnamed_namespaces points out that they do have internal linkage. External linkage basically implies visibility as I read through that and http://en.cppreference.com/w/cpp/language/storage_duration and this does make some sense.

But I'm not sure how big a deal it actually is, and if we *do* care about leaking symbols (which we might! our builds take *forever*) we probably want to start doing this everywhere.

gregsfortytwo:
http://pulpito.ceph.com/gregf-2016-10-06_18:07:58-fs-greg-fs-testing-105---basic-mira/ looks pretty good; OSD failure (http://tracker.ceph.com/issues/17553), pjd failure due to missing qa-suite wip-getuid merge, 2xlibcephfs-java failures (http://tracker.ceph.com/issues/17548), 1 dead job.

Still have to work out level of care.

jcsp:
OK, so this isn't unreasonable, but I'm not sure I want to handle a separate PR for every file you bump into this.  If this is really helping and making an difference, please do a PR that does this across all the MDS .cc files

PR #11319, opened 2016-10-04
os/bluestore: optimize intrusive sets for size.

markhpc:
This saves us 8 bytes per structure that uses set_base_hook.  Performance may be lower or higher depending on memory locality, the cost of masking operations, and memory saved potentially allowing for more cached onodes.


markhpc:
This may actually provide a slight performance improvement for small random writes over time.  Memory usage appears to decrease by about 3-10% for small random writes depending on the number of cached onodes and how many extents and blobs are present.

PR #11327, opened 2016-10-05
librbd: cls_rbd updates for separate data pool

dillaman:


trociny:
@dillaman Jenkins failure looks related

dillaman:
@trociny should be fixed

PR #11328, opened 2016-10-05
os/bluestore: pack a few more in-memory types

liewegas:


markhpc:
To build this needs the writing_map changed to writing in Blob::can_split (BlueStore.h) and Bufferspace::split (BlueStore.cc).

PR #11331, opened 2016-10-05
mempool: mempool infrastructure, bluestore changes to use it

liewegas:
Signed-off-by: Allen Samuels allen.samuels@sandisk.com

### DIFF
--- a/src/global/mempool.cc
+++ b/src/global/mempool.cc
@@ -91,18 +91,18 @@ void mempool::pool_t::get_stats(
   for (size_t i = 0; i < num_shards; ++i) {
     total->items += shard[i].items;
     total->bytes += shard[i].bytes;
-    if (debug) {
-      std::unique_lock<std::mutex> shard_lock(shard[i].lock);
-      for (const list_member_t *p = shard[i].types.next;
-	   p != &shard[i].types;
-	   p = p->next) {
-	const pool_allocator_base_t *c =
-	  reinterpret_cast<const pool_allocator_base_t *>(p);
-	std::string n = ceph_demangle(c->type_id);
-	stats_t &s = (*by_type)[n];
-	s.bytes = c->items * c->item_size;
-	s.items = c->items;
-      }
+  }
+  if (debug) {
+    std::unique_lock<std::mutex> shard_lock(lock);
+    for (const list_member_t *p = types.next;
+	 p != &types;
+	 p = p->next) {
+      const pool_allocator_base_t *c =
+	reinterpret_cast<const pool_allocator_base_t *>(p);
+      std::string n = ceph_demangle(c->type_id);
+      stats_t &s = (*by_type)[n];
+      s.bytes = c->items * c->item_size;
+      s.items = c->items;
     }
   }
 }
--- a/src/global/mempool.cc
+++ b/src/global/mempool.cc
@@ -72,6 +72,7 @@ size_t mempool::pool_t::allocated_bytes() const
   for (size_t i = 0; i < num_shards; ++i) {
     result += shard[i].bytes;
   }
+  assert(result >= 0);
   return (size_t) result;
 }
 
@@ -81,6 +82,7 @@ size_t mempool::pool_t::allocated_items() const
   for (size_t i = 0; i < num_shards; ++i) {
     result += shard[i].items;
   }
+  assert(result >= 0);
   return (size_t) result;
 }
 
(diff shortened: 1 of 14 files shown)


src/include/mempool.h
```diff
@@ -0,0 +1,521 @@
...
+can't use :: in a variable name.)
+
+In order to use the STL containers, a few additional declarations
+are needed.  For example,
+
+  unittest_1::map<int> myvec;
+
+requires
+
+  MEMPOOL_DEFINE_FACTORY(int, int, unittest_1);
```
allensamuels: I think MEMPOOL_DEFINE_FACTORY doesn't belong here.???

src/include/mempool.h
```diff
@@ -0,0 +1,521 @@
...
+traits, or simply define the allocator for with the cached and
+uncached case and leave one of them unused (but polluting your
+debug dumps).  For example,
+
+  unittest_2::unordered_map<int, std::string> one;       // not cached
+  unittest_2::unordered_map<std::string, int> one;       // cached
+
+needs
+
+ MEMPOOL_DEFINE_UNORDERED_MAP_BASE_FACTORY(unittest_2);  // used by both
```
allensamuels: MEMPOOL_DEFINE_UNORDERED_MAP_BASE_FACTORY. Why not ALWAYS declare one of these in the DEFINE_MEMPOOLS_HELPER macro so that people don't have to worry about it? If it's unused, who cares?

src/include/mempool.h
```diff
@@ -0,0 +1,521 @@
...
+
+You can also interogate a specific pool programatically with
+
+  size_t bytes = unittest_2::allocated_bytes();
+  size_t items = unittest_2::allocated_items();
+
+Note that you cannot easily query per-type, primarily because debug
+mode is optional and you should not rely on that information being
+available.
+
```
allensamuels: Should include a comment about the run-time complexity of these interrogation calls.

liewegas:
I had that before, but it pollutes the dump output with hash table entries 
even when they're not used.

allensamuels:
Why not have the dumper simply suppress type signatures that have no allocations??? That’s probably useful anyways….(Though occasionally misleading). Perhaps a boolean to the commands to force dumping on 0-instance typedefs….


Allen Samuels
SanDisk |a Western Digital brand

src/include/mempool.h
```diff
@@ -0,0 +1,521 @@
...
+// we shard pool stats across many shard_t's to reduce the amount
+// of cacheline ping pong.
+enum { num_shards = 64 };
+
+struct shard_t {
+  std::atomic<size_t> bytes = {0};
+  std::atomic<size_t> items = {0};
+  mutable std::mutex lock;  // only used for containers list
+  list_member_t containers; // protected by lock
```
allensamuels: containers -> container_types ??

xiexingguo:
Does not compile: 
```
CMakeFiles/unittest_mempool.dir/test_mempool.cc.o: In function `pool_allocator':
/home/jenkins-build/build/workspace/ceph-pull-requests/src/include/mempool.h:347: undefined reference to `mempool::pool_allocator<(mempool::pool_index_t)1, std::pair<int, obj> >::base'
make[3]: *** [bin/unittest_mempool] Error 1
make[2]: *** [src/test/CMakeFiles/unittest_mempool.dir/all] Error 2
make[2]: *** Waiting for unfinished jobs....

**Comment by liewegas** [comment]
I pushed another patch.. does that fix it?

**Comment by xiexingguo** [comment]
Negative :-)

**Comment by liewegas** [comment]
What compiler version and/or distro are you using?

**Comment by xiexingguo** [comment]
@liewegas 
It's jenkins itself that is not happy with this change.
Check below:
https://jenkins.ceph.com/job/ceph-pull-requests/12536/console

File: src/global/mempool.cc [hunk 5]
```diff
@@ -0,0 +1,123 @@
...
+    f->close_section();
+  }
+}
+
+void mempool::set_debug_mode(bool d)
+{
+  debug_mode = d;
+  for (size_t i = 0; i < mempool::num_pools; ++i) {
+    if (pools[i]) {
+      pools[i]->debug = d;
```
yuyuyu101: do we need to reset existing debug item?

src/os/bluestore/BlueStore.h
```diff
@@ -1441,6 +1491,44 @@ class BlueStore : public ObjectStore,
...
+      create("bstore_mempool");
+    }
+    void shutdown() {
+      lock.Lock();
+      stop = true;
+      cond.Signal();
+      lock.Unlock();
+      join();
+    }
+  } mempool_thread;
```
yuyuyu101: this trim cache way makes me a little nervous. because if you think get memory stat is a little heavy, we need to handle this in other usages?

athanatos:
Looks reasonable to me (didn't really look at the bluestore stuff).

src/os/bluestore/BlueStore.h
liewegas: I'm not sure I understand.  Which usages?  Right now this is the only thing looking at the mempool usage.  The main thing is that it's expensive enough that I don't want to call trim_cache() after every op any more...

liewegas:
Okay, I've switched back to get_pool() such that we now rely on the array being zeroed on process startup before *any* ctors start to run.  I think this is as safe/robust as it gets.

@allensamuels I take it you don't like the mempool:: prefix?  I did this because I went to create an 'osd' mempool and immediately had problems with the osd namespace colliding with variables named osd (like OSD *osd or int osd).

allensamuels:
Correct, I found the extra prefix ugly. Not a big deal. Personally, I would have just used a prefix like mp_osd for the mempool for the osd rather than nesting the namespaces and uglifiying everthing else.


Allen Samuels
SanDisk |a Western Digital brand

PR #11334, opened 2016-10-05
make-dist: set rpm_release correctly for release builds

dmick:
Similar to autobuild-ceph 17591db, only try to set rpm_release if there
is a release part to the git describe output (i.e. if it's not at a
version tag)


ktdreyer:
:trophy: dmick is this week's bash champion

PR #11336, opened 2016-10-05
os/bluestore: switch spanning_blob_map to std::map

liewegas:
The intrusive map overhead on Blob is expensive when a tiny fraction of
Blobs actually end up in the spanning_blob_map.  Let's pay the extra
allocation overhead (for the map node holding the BlobRef) for those rare
cases.


markhpc:
This is going to be much better I think!  Should let us cache some more onodes at the very least.

PR #11339, opened 2016-10-05
mon/PGMap: PGs can be stuck more than one thing

liewegas:
This was inadvertantly changed waaay back in
9c95bb0454dbad781b6655f4c7116620005101f7.

This fixes the health warnings so that they are
complete and (most importantly) consistent with the
results of 'pg dump_stuck ...'.


tchaikov:
lgtm. actually i spotted this also in last qa run.

tchaikov:
once the dump stuck test passes, it's good to merge!

PR #11341, opened 2016-10-05
make-dist: set rpm_release correctly for release builds (kraken)

dmick:
Similar to autobuild-ceph 17591db, only try to set rpm_release if there
is a release part to the git describe output (i.e. if it's not at a
version tag)

(cherry picked from commit 0ff0c9a8b9fee0f906ce84881cceb163873cb751)


liewegas:
kraken isn't forked yet.. this can jsut go into msater

alfredodeza:
This is already in kraken, I cherry-picked earlier (see 927a3733cc1b9e4afff33eff2c4602f5bafaebdc )

PR #11343, opened 2016-10-05
Update version of civetweb to 1.8

yehudasa:
a simpler, less intusive change than the one in the other two PRs. This one only deals with:
- change the submodule to the version created by Pritha that is based on civetweb 1.8 + rgw related changes broken down to multiple commits
- update rgw code to support the previous change
- add USE_IPV6=1 when compiling civetweb

IMO all other changes if needed can be done on a separate commit.


cbodley:
build failing with `src/rgw/rgw_frontend.h:13:31: fatal error: civetweb/civetweb.h: No such file or directory`

cbodley:
@yehudasa is this ready for review/testing?

yehudasa:
@cbodley this needs testing.. however, there are a couple of functional s3 tests that we should either disable, or make civetweb handle differently; otherwise we're going to fail most of the tests

PR #11345, opened 2016-10-06
messages: fix out of range assertion

jcsp:
clang detects that this is invalid (presumably
using an 8 bit type for the enum)

### DIFF
--- a/src/messages/MMgrReport.h
+++ b/src/messages/MMgrReport.h
@@ -36,7 +36,7 @@ public:
     ::encode(path, bl);
     ::encode(description, bl);
     ::encode(nick, bl);
-    assert(type < 256);
+    static_assert(sizeof(type) == 1, "perfcounter_type_d must be one byte");
     ::encode((uint8_t)type, bl);
     ENCODE_FINISH(bl);
   }
--- a/src/common/perf_counters.h
+++ b/src/common/perf_counters.h
@@ -32,7 +32,7 @@ class CephContext;
 class PerfCountersBuilder;
 class PerfCountersCollectionTest;
 
-enum perfcounter_type_d
+enum perfcounter_type_d : uint8_t
 {
   PERFCOUNTER_NONE = 0,
   PERFCOUNTER_TIME = 0x1,


wjwithagen:
@jcsp 
Can't fool Clang :)

/usr/srcs/Ceph/work/ceph/src/messages/MMgrReport.h:39:17: warning: comparison of constant 255 with expression of type 'const enum perfcounter_type_d' is always true
      [-Wtautological-constant-out-of-range-compare]
    assert(type <=255);
           ~~~~ ^ ~~~

jcsp:
Gaaah!

jcsp:
I don't really want to remove the assert, because there's no guarantee the enums are always going to be char-sized :-/

liewegas:
If that still triggers a clang warning we can just change it to 254, which should be just as effective.

tchaikov:
or we can declare the `perfcounter_type_d` as 

```c++
enum class type : uint8_t {
 // ...
}
```

wjwithagen:
@tchaikov @liewegas @jcsp 
I did not make it a enum class, but did add the `: uint8_t` in the declaration.
Clang is now happy.

And I think it also can get ride of a lot of the casts like:
    `::encode((uint8_t)type, bl)` 

Have not run it thru GCC

wjwithagen:
@tchaikov @liewegas @jcsp
Right, removing the casts actually throws Clang into a selction problem on which of the many en/decode options it can promote to to actually get the work done.
So removing the casts is not an option.

tchaikov:
@wjwithagen yeah, less code churn if unscoped version is used, or instead of keeping the casts, we can add
```c++
template<typename E,
         typename std::enable_if<std::is_enum<E>{}>::type* = nullptr>
void encode(E e, bufferlist& bl) {
  encode(static_cast<typename std::underlying_type<E>::type>(e), bl);
}
```

to delegate the `encode()` calls applying to the enum types to their underlying types.

jcsp:
Updated this to always use uint8_t for the enum, and just assert on the size of the resulting type (the assertion is still strictly always true, so let's see if clang complains about that too!)

src/messages/MMgrReport.h
```diff
@@ -36,7 +36,7 @@ class PerfCounterType
     ::encode(path, bl);
     ::encode(description, bl);
     ::encode(nick, bl);
-    assert(type < 256);
+    assert(sizeof(type) == 1);  // enum type declared as uint8_t
```
tchaikov: nit, could be `static_assert()`.

jcsp: done

tchaikov: could be
```c++
static_assert (sizeof(type) == 1 , "enum type declared as uint8_t");
```

```c++
static_assert ( bool_constexpr )	//	(since C++17)
```

tchaikov:
lgtm.

PR #11347, opened 2016-10-06
[DNM] rgw: Adding code for Role creation, deletion and reading.

pritha-srivastava:
A new class RGWRoles for AWS roles has been added.
Also, code for role create. delete and list for radosgw-admin has been added.

Signed-off-by: Pritha Srivastava prsrivas@redhat.com

### DIFF
--- a/src/rgw/rgw_admin.cc
+++ b/src/rgw/rgw_admin.cc
@@ -162,6 +162,10 @@ void _usage()
   cout << "  role delete                delete a role\n";
   cout << "  role get                   get a role\n";
   cout << "  role modify                modify the assume role policy of an existing role\n";
+  cout << "  role-policy put            add/update permission policy to role\n";
+  cout << "  role-policy list           list policies attached to a role\n";
+  cout << "  role-policy get            get the specified inline policy document embedded with the given role\n";
+  cout << "  role-policy delete         delete policy attached to a role\n";
   cout << "options:\n";
   cout << "   --tenant=<tenant>         tenant name\n";
   cout << "   --uid=<id>                user id\n";
@@ -257,6 +261,8 @@ void _usage()
   cout << "   --role-name               name of the role to create\n";
   cout << "   --path                    path to the role\n";
   cout << "   --assume-role-policy-doc  the trust relationship policy document that grants an entity permission to assume the role\n";
+  cout << "   --policy-name             name of the policy document\n";
+  cout << "   --policy-doc              permission policy document\n";
   cout << "\n";
   cout << "<date> := \"YYYY-MM-DD[ hh:mm:ss]\"\n";
   cout << "\nQuota options:\n";
@@ -412,6 +418,10 @@ enum {
   OPT_ROLE_DELETE,
   OPT_ROLE_GET,
   OPT_ROLE_MODIFY,
+  OPT_ROLE_POLICY_PUT,
+  OPT_ROLE_POLICY_LIST,
+  OPT_ROLE_POLICY_GET,
+  OPT_ROLE_POLICY_DELETE,
 };
 
 static int get_cmd(const char *cmd, const char *prev_cmd, const char *prev_prev_cmd, bool *need_more)
@@ -447,6 +457,7 @@ static int get_cmd(const char *cmd, const char *prev_cmd, const char *prev_prev_
       strcmp(cmd, "regionmap") == 0 ||
       strcmp(cmd, "replicalog") == 0 ||
       strcmp(cmd, "role") == 0 ||
+      strcmp(cmd, "role-policy") == 0 ||
       strcmp(cmd, "subuser") == 0 ||
       strcmp(cmd, "sync") == 0 ||
       strcmp(cmd, "usage") == 0 ||
@@ -788,6 +799,15 @@ static int get_cmd(const char *cmd, const char *prev_cmd, const char *prev_prev_
       return OPT_ROLE_GET;
     if (strcmp(cmd, "modify") == 0)
       return OPT_ROLE_MODIFY;
+  } else if (strcmp(prev_cmd, "role-policy") == 0) {
+    if (strcmp(cmd, "put") == 0)
+      return OPT_ROLE_POLICY_PUT;
+    if (strcmp(cmd, "list") == 0)
+      return OPT_ROLE_POLICY_LIST;
+    if (strcmp(cmd, "get") == 0)
+      return OPT_ROLE_POLICY_GET;
+    if (strcmp(cmd, "delete") == 0)
+      return OPT_ROLE_POLICY_DELETE;
   }
 
   return -EINVAL;
@@ -852,6 +872,24 @@ static void show_user_info(RGWUserInfo& info, Formatter *formatter)
   encode_json("user_info", info, formatter);
   formatter->flush(cout);
   cout << std::endl;
+}
+
+static void show_perm_policy(string perm_policy, Formatter* formatter)
+{
+  formatter->open_object_section("role");
+  formatter->dump_string("Permission policy", perm_policy);
+  formatter->close_section();
+  formatter->flush(cout);
+}
+
+static void show_policy_names(std::vector<string> policy_names, Formatter* formatter)
+{
+  formatter->open_array_section("PolicyNames");
+  for (const auto& it : policy_names) {
+    formatter->dump_string("policyname", it);
+  }
+  formatter->close_section();
+  formatter->flush(cout);
 }
 
 static void show_role_info(RGWRole& role, Formatter* formatter)
@@ -2056,7 +2094,7 @@ int main(int argc, char **argv)
   std::string zone_name, zone_id, zone_new_name;
   std::string zonegroup_name, zonegroup_id, zonegroup_new_name;
   std::string api_name;
-  std::string role_name, path, assume_role_doc;
+  std::string role_name, path, assume_role_doc, policy_name, perm_policy_doc;
   list<string> endpoints;
   int tmp_int;
   int sync_from_all_specified = false;
@@ -2453,6 +2491,10 @@ int main(int argc, char **argv)
       path = val;
     } else if (ceph_argparse_witharg(args, i, &val, "--assume-role-policy-doc", (char*)NULL)) {
       assume_role_doc = val;
+    } else if (ceph_argparse_witharg(args, i, &val, "--policy-name", (char*)NULL)) {
+      policy_name = val;
+    } else if (ceph_argparse_witharg(args, i, &val, "--policy-doc", (char*)NULL)) {
+      perm_policy_doc = val;
     } else if (strncmp(*i, "-", 1) == 0) {
       cerr << "ERROR: invalid flag " << *i << std::endl;
       return EINVAL;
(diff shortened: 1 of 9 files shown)


src/rgw/rgw_role.cc
```diff
@@ -0,0 +1,220 @@
...
+  /* create unique id */
+  uuid_d new_uuid;
+  char uuid_str[37];
+  new_uuid.generate_random();
+  new_uuid.print(uuid_str);
+  id = uuid_str;
+
+  // Creation time
+  creation_date = boost::posix_time::to_iso_extended_string(
+                          boost::posix_time::microsec_clock::universal_time());
```
cbodley: does the standard library support this string format? i'd prefer to use `ceph::real_clock` instead of boost if at all possible

src/rgw/rgw_admin.cc
```diff
@@ -3955,7 +3986,50 @@ int main(int argc, char **argv)
...
+      }
+      RGWRole role(g_ceph_context, store, role_name);
+      ret = role.list();
+      if (ret < 0) {
+        cerr <<"could not list role: " << err_msg << std::endl;
+        return -ret;
+      }
+      show_role_info(role, formatter);
+      return 0;
+    }
```
cbodley: the 'role list' command feels more like a 'role get', doesn't it? will there ever be multiple entries to output here?

src/rgw/rgw_role.cc
```diff
@@ -0,0 +1,220 @@
...
+
+  ret = read_info();
+  if (ret < 0) {
+    ldout(cct, 0) << "ERROR: finding role id in pool: " << pool_name << ": "
+                  << id << ": " << cpp_strerror(-ret) << dendl;
+    return ret;
+  }
+
+  return 0;
+}
```
cbodley: this function is just reading the role into memory, right? can we name it `init()` or `read()` instead?

src/rgw/rgw_role.cc
```diff
@@ -0,0 +1,220 @@
...
+                  ": " << id << dendl;
+    return -EIO;
+  }
+
+  return 0;
+}
+
+int RGWRole::read_name()
+{
+  string pool_name = store->get_zone_params().roles_pool.name;
```
cbodley: shouldn't need to make a copy of the string here. consider changing this to `auto& pool = store->get_zone_params().roles_pool`, so it be used for both the `rgw_get_system_obj()` call, and to access the name as `pool.name`. can do similar in the other functions too

pritha-srivastava:
@cbodley : I didn't add a separate oid for role names because, if a user tries to create a role with a name same as an existing role id, then the method create will return an error saying that the name is in use as an id by another role.
And I think allowing roles with the same name and id is slightly confusing (though with a different prefix). I know the convention has been to separate out ids and names. If we have to strictly follow that then I'll add oid prefixing for the time being till we switch over to rados namespacing. Let me know what you think.

pritha-srivastava:
I gave it some more thought, i guess it will be better to separate out the namespaces for roles, names etc (oid prefixes) for the time being, if going forward we would want to list the role names. For e.g. in case of list roles, we may have to add a separate namespace for path in order to list all roles under the same path.

src/rgw/rgw_role.h
```diff
@@ -0,0 +1,81 @@
...
+    ::decode(name, bl);
+    ::decode(path, bl);
+    ::decode(creation_date, bl);
+    DECODE_FINISH(bl);
+  }
+
+  const string& get_id() { return id; }
+  const string& get_name() { return name; }
+  const string& get_path() { return path; }
+  const string& get_create_date() { return creation_date; }
```
cbodley: these functions can be `const`

src/rgw/rgw_admin.cc
```diff
@@ -158,6 +159,9 @@ void _usage()
   cout << "  orphans find               init and run search for leaked rados objects (use job-id, pool)\n";
   cout << "  orphans finish             clean up search for leaked rados objects\n";
   cout << "  orphans list-jobs          list the current job-ids for orphans search\n";
+  cout << "  role create                create a role\n";
+  cout << "  role delete                delete a role\n";
+  cout << "  role get                   get a role\n";
```
cbodley: can we add a bit more context? something like `create an AWS role for use with STS`? `delete a role` and `get a role` are fine as they are

src/rgw/rgw_admin.cc
```diff
@@ -3955,7 +3983,50 @@ int main(int argc, char **argv)
...
+  case OPT_ROLE_CREATE:
+    {
+      if (role_name.empty()) {
+        cerr << "Empty role name" << std::endl;
+        return -EINVAL;
+      }
+      RGWRole role(g_ceph_context, store, role_name, path);
+      ret = role.create(true);
+      if (ret < 0) {
+        cerr <<"could not create role: " << err_msg << std::endl;
```
cbodley: nit: space after <<

also, where is `err_msg` coming from?

src/rgw/rgw_rados.h
```diff
@@ -937,7 +938,7 @@ struct RGWZoneParams : RGWSystemMetaObj {
   int fix_pool_names();
   
   void encode(bufferlist& bl) const {
-    ENCODE_START(8, 1, bl);
+    ENCODE_START(9, 1, bl);
```
pritha-srivastava: @cbodley : There was a merge conflict here; tier_config was added in master at version number 8, so i have bumped up the version again to 9, to add roles_pool. Is that correct?

cbodley: :+1:

src/rgw/rgw_admin.cc
```diff
@@ -157,6 +158,9 @@ void _usage()
   cout << "  orphans find               init and run search for leaked rados objects (use job-id, pool)\n";
   cout << "  orphans finish             clean up search for leaked rados objects\n";
   cout << "  orphans list-jobs          list the current job-ids for orphans search\n";
+  cout << "  role create                create a AWS role for use with STS \n";
+  cout << "  role delete                delete a role\n";
+  cout << "  role get                   get a role\n";
```
cbodley: nit: extra space after `STS`

also, there's a unit test that compares the output of `radosgw-admin -h` against `test/cli/radosgw-admin/help.t` - could you please update that file to match your changes? jenkins is complaining about it:
```
/home/jenkins-build/build/workspace/ceph-pull-requests/src/test/cli/radosgw-admin/help.t: failed
--- /home/jenkins-build/build/workspace/ceph-pull-requests/src/test/cli/radosgw-admin/help.t
+++ /home/jenkins-build/build/workspace/ceph-pull-requests/src/test/cli/radosgw-admin/help.t.err
@@ -108,6 +108,9 @@
     orphans find               init and run search for leaked rados objects (use job-id, pool)
     orphans finish             clean up search for leaked rados objects
     orphans list-jobs          list the current job-ids for orphans search
+    role create                create a AWS role for use with STS 
+    role delete                delete a role
+    role get                   get a role
   options:
```

src/rgw/rgw_admin.cc
```diff
@@ -2416,6 +2440,10 @@ int main(int argc, char **argv)
       parse_tier_config_param(val, tier_config_add);
     } else if (ceph_argparse_witharg(args, i, &val, "--tier-config-rm", (char*)NULL)) {
       parse_tier_config_param(val, tier_config_rm);
+    } else if (ceph_argparse_witharg(args, i, &val, "--role-name", (char*)NULL)) {
+      role_name = val;
+    } else if (ceph_argparse_witharg(args, i, &val, "--path", (char*)NULL)) {
+      path = val;
```
cbodley: these options need to be documented in the usage also

src/rgw/rgw_admin.cc
```diff
@@ -4017,7 +4045,50 @@ int main(int argc, char **argv)
       cout << std::endl;
     }
     return 0;
-
+  case OPT_ROLE_CREATE:
+    {
+      if (role_name.empty()) {
+        cerr << "Empty role name" << std::endl;
+        return -EINVAL;
+      }
```
cbodley: is `path` allowed to be empty, or should we check that here too?

pritha-srivastava: @cbodley : path can be empty since it is optional. It is set to "/" if the user doesn't set it to anything (according to AWS create role definition.)

cbodley:
@pritha-srivastava looks good! i'm happy to merge once the `[DNM]` tag goes away

pritha-srivastava:
@cbodley : Removed the 'dnm' tag.

src/rgw/rgw_admin.cc
```diff
@@ -4017,7 +4047,50 @@ int main(int argc, char **argv)
...
+      RGWRole role(g_ceph_context, store, role_name);
+      ret = role.delete_obj();
+      if (ret < 0) {
+        cerr << "could not delete role: " << cpp_strerror(-ret) << std::endl;
+        return -ret;
+      }
```
cbodley: could you add a `cout` here so the user knows it succeeded?

src/rgw/rgw_role.cc
```diff
@@ -0,0 +1,245 @@
...
+  if (ret < 0) {
+    ldout(cct, 0) << "ERROR: finding role name in pool: " << pool.name << ": "
+                  << name << ": " << cpp_strerror(-ret) << dendl;
+    return ret;
+  }
+
+  ret = read_info();
+  if (ret < 0) {
+    ldout(cct, 0) << "ERROR: finding role id in pool: " << pool.name << ": "
+                  << id << ": " << cpp_strerror(-ret) << dendl;
```
cbodley: both `read_name()` and `read_info()` are already logging error messages, so i don't think we need to log it again here:
```
$ radosgw-admin role get --role-name=foo
2016-10-18 09:53:41.912480 7f959a04ba80  0 ERROR: failed reading role name from pool: default.rgw.roles: foo: (2) No such file or directory
could not get role: (2) No such file or directory
2016-10-18 09:53:41.912580 7f959a04ba80  0 ERROR: finding role name in pool: default.rgw.roles: foo: (2) No such file or directory
```

pritha-srivastava:
@cbodley: Made the following changes:
1. Code clean up wrt usability.
2. Added error handling for the create when secondary index creation fails.
3. Added code for parsing trust policy document (json) and encoding that as a string for the time being.

src/rgw/rgw_role.cc
```diff
@@ -0,0 +1,275 @@
...
+  this->trust_policy = trust_policy;
+
+  ret = store_info(false);
+  if (ret < 0)
+  {
+    ldout(cct, 0) << "ERROR:  updating role info in pool: " << pool.name << ": "
+                  << id << ": " << cpp_strerror(-ret) << dendl;
+    return ret;
+  }
+}
```
pritha-srivastava: @cbodley: Yeah initially, i thought of breaking the logic as you have suggested, but if we implement the AWS API, then the same code (that will be there in rgw-admin.cc), will be replicated in the class which will have the corresponding method for role modify API. Thats the reason why I put everything under one interface in RGWRole class. Let me know what you think.

cbodley: ok, makes sense

mattbenjamin:
@adamemerson would be good to get your review here

src/rgw/rgw_role.cc
```diff
@@ -82,10 +84,10 @@ int RGWRole::create(bool exclusive)
...
   real_clock::to_timeval(t, tv);
 
-  char buf[30];
-  strftime(buf,30,"%Y-%m-%dT%H:%M:%S", std::gmtime(&tv.tv_sec));
-  sprintf(buf + strlen(buf),".%dZ",(int)tv.tv_usec);
-  creation_date.assign(buf, strlen(buf));
+  std::ostringstream oss;
+  oss << std::put_time(std::gmtime(&tv.tv_sec), "%Y-%m-%dT%H:%M:%S");
+  oss << "." << (tv.tv_usec/1000) << "Z";
+  creation_date = oss.str();
```
pritha-srivastava: @mattbenjamin : Does the above look better as compared to strftime? I came across put_time which is there since C++11.

adamemerson: C++ time formatting is, unfortunately, a bit of a fiasco so I'm not sure there's a really *good* solution. strftime is probably faster since it has less overhead, not needing to call the entire iostream infrastructure into place.

The one thing is, don't use std::gmtime. It's not guaranteed to be thread safe (and strict interpretation of the C standard forbids it to be thread safe, since all calls to gmtime and localtime are required to return the same pointer). Use gmtime_r. It's not C++ but is in the standard C library. It started out as a GNU extension but has been folded into the SIngle Unix Specification so should be available everywhere.

adamemerson:
This seems a fine PR to me, at least once we switch gmtime for gmtime_r.

Labyrinthine Gematria Tallies Modularly.

adamemerson:
This would probably be a pretty good patch to merge even as-is, though I could also put my stuff on top of it, whichever people think is better.

cbodley:
https://github.com/ceph/ceph/pull/12104 is the latest PR with these changes - closing this one

PR #11348, opened 2016-10-06
[DNM]: Wip rgw pool namespace

pritha-srivastava:
Added code for using rados namespace for storing different objects in the same pool.

### DIFF
--- a/src/rgw/rgw_common.h
+++ b/src/rgw/rgw_common.h
@@ -726,6 +726,7 @@ struct rgw_bucket {
   std::string oid; /*
                     * runtime in-memory only info. If not empty, points to the bucket instance object
                     */
+  std::string ns;
 
   rgw_bucket() { }
   // cppcheck-suppress noExplicitConstructor
@@ -754,7 +755,7 @@ struct rgw_bucket {
   }
 
   void encode(bufferlist& bl) const {
-     ENCODE_START(9, 3, bl);
+     ENCODE_START(10, 3, bl);
     ::encode(name, bl);
     ::encode(data_pool, bl);
     ::encode(marker, bl);
@@ -762,10 +763,11 @@ struct rgw_bucket {
     ::encode(index_pool, bl);
     ::encode(data_extra_pool, bl);
     ::encode(tenant, bl);
+    ::encode(ns, bl);
     ENCODE_FINISH(bl);
   }
   void decode(bufferlist::iterator& bl) {
-    DECODE_START_LEGACY_COMPAT_LEN(9, 3, 3, bl);
+    DECODE_START_LEGACY_COMPAT_LEN(10, 3, 3, bl);
     ::decode(name, bl);
     ::decode(data_pool, bl);
     if (struct_v >= 2) {
@@ -791,6 +793,9 @@ struct rgw_bucket {
     if (struct_v >= 8) {
       ::decode(tenant, bl);
     }
+    if (struct_v >= 10) {
+      ::decode(ns, bl);
+    }
     DECODE_FINISH(bl);
   }
 
@@ -803,6 +808,10 @@ struct rgw_bucket {
       return data_pool;
     }
     return data_extra_pool;
+  }
+
+  void set_ns(string& ns) {
+    this->ns = ns;
   }
 
   void dump(Formatter *f) const;
--- a/src/rgw/rgw_json_enc.cc
+++ b/src/rgw/rgw_json_enc.cc
@@ -543,9 +543,24 @@ void RGWQuotaInfo::decode_json(JSONObj *obj)
   JSONDecoder::decode_json("enabled", enabled, obj);
 }
 
+static void parse_pool_name(string& bucket_name, string& name, string& ns)
+{
+  auto pos = bucket_name.find(":");
+  if (pos != string::npos) {
+    name = bucket_name.substr(0, pos);
+    ns = bucket_name.substr(pos + 1);
+  } else {
+    name = std::move(bucket_name);
+  }
+}
+
 void rgw_bucket::dump(Formatter *f) const
 {
-  encode_json("name", name, f);
+  if (! ns.empty()) {
+    encode_json("name", name + ":" + ns, f);
+  } else {
+    encode_json("name", name, f);
+  }
   encode_json("pool", data_pool, f);
   encode_json("data_extra_pool", data_extra_pool, f);
   encode_json("index_pool", index_pool, f);
@@ -555,13 +570,16 @@ void rgw_bucket::dump(Formatter *f) const
 }
 
 void rgw_bucket::decode_json(JSONObj *obj) {
-  JSONDecoder::decode_json("name", name, obj);
+  string bucket_name;
+  JSONDecoder::decode_json("name", bucket_name, obj);
   JSONDecoder::decode_json("pool", data_pool, obj);
   JSONDecoder::decode_json("data_extra_pool", data_extra_pool, obj);
   JSONDecoder::decode_json("index_pool", index_pool, obj);
   JSONDecoder::decode_json("marker", marker, obj);
   JSONDecoder::decode_json("bucket_id", bucket_id, obj);
   JSONDecoder::decode_json("tenant", tenant, obj);
+
+  parse_pool_name(bucket_name, name, ns);
 }
 
 void RGWBucketEntryPoint::dump(Formatter *f) const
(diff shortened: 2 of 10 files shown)


src/rgw/rgw_common.h
```diff
@@ -726,6 +726,7 @@ struct rgw_bucket {
   std::string oid; /*
                     * runtime in-memory only info. If not empty, points to the bucket instance object
                     */
+  std::string name_space;
```
cbodley: can we drop the `_` and use `namespace` to match the librados `set_namespace()` api?

src/rgw/rgw_rados.cc
```diff
@@ -1518,6 +1520,11 @@ int RGWZoneParams::fix_pool_names()
   return 0;
 }
 
+void RGWZoneParams::set_pool_namespaces()
+{
+  roles_pool.name_space = "role";
+}
+
```
cbodley: the namespaces should be parsed from the zone's pool strings, i.e. "pool:namespace". we shouldn't be hard-coding them anywhere

src/rgw/rgw_rados.cc
```diff
@@ -5071,12 +5080,30 @@ int RGWRados::create_pool(rgw_bucket& bucket)
   if (ret < 0)
     return ret;
 
+  if (! bucket.name_space.empty()) {
+    librados::IoCtx io_ctx;
+    ret = rad->ioctx_create(pool.c_str(), io_ctx);
+    if (ret) {
+      return ret;
+    }
+    io_ctx.set_namespace(bucket.name_space);
```
cbodley: the `set_namespace()` call just tells the `librados::IoCtx` object which namespace to use for subsequent rados requests. so this will just initialize an IoCtx and then throw it away

instead, we'll need to go through and find everywhere that we're calling `ioctx_create()` for one of these pools, and add the corresponding calls to `set_namespace()`. there are quite a few, so we should probably refactor those into  a common helper function so we don't miss any. `RGWRados::open_pool_ctx()` looks like a good candidate, if it could take the pool as an `rgw_bucket` instead of a `std::string`

src/rgw/rgw_rados.cc
```diff
@@ -1505,6 +1506,7 @@ int RGWZoneParams::fix_pool_names()
   user_email_pool = fix_zone_pool_name(pool_names, name, ".rgw.users.email", user_email_pool.name);
   user_swift_pool = fix_zone_pool_name(pool_names, name, ".rgw.users.swift", user_swift_pool.name);
   user_uid_pool = fix_zone_pool_name(pool_names, name, ".rgw.users.uid", user_uid_pool.name);
+  roles_pool = fix_zone_pool_name(pool_names, name, ".rgw.authorizationdata", roles_pool.name);
```
cbodley: we'll want to update these defaults (".rgw.users.keys", etc) to use the "pool:namespace" format. i'm not sure which pools we'll want to combine initially, but maybe something like:
```
".rgw.auth:users.keys"
".rgw.auth:users.email"
".rgw.auth:users.swift"
".rgw.auth:users.uid"
".rgw.auth:roles"
```
we'll then need to parse these strings, so that the ".rgw.auth" part ends up in `rgw_bucket::name`, and the part after ':', if any, ends up in `rgw_bucket::namespace`

`fix_zone_pool_name()` will also need a lot more thought:
* it should only be enforcing uniqueness of the pool name part, and ignoring the namespaces
* it can't return a single std::string to be consumed by `rgw_bucket`'s constructor - we'd need to be able to pass in both the pool name and its namespace
* if it 'fixes' a pool name (i.e. `foo.rgw.auth` -> `foo_[random].rgw.auth`) for `user_keys_pool`, it needs to use the same value of `[random]` when it 'fixes' the pool name for other entries

src/rgw/rgw_common.h
pritha-srivastava: since namespace is a C++ keyword, it does not allow me to use that as a variable name. Should I change the api name instead?

src/rgw/rgw_common.h
```diff
@@ -762,6 +763,7 @@ struct rgw_bucket {
     ::encode(index_pool, bl);
     ::encode(data_extra_pool, bl);
     ::encode(tenant, bl);
+    ::encode(name_space, bl);
```
pritha-srivastava: ok, for this one I was curious that the version number was bumped to 9 and there was no additional field in the structure, I dug a little and saw that the version number was bumped to 9 by mistake by one of my auth rework PRs !. So that is why I kept the version at 9 for 'name_space'.

src/rgw/rgw_rados.cc
pritha-srivastava: To avoid parsing the 'poolname:namespace' string, as you have stated in your earlier comment, I added another method to set the namespaces. And anyway we will be hardcoding the names as poolname:namespace, so i hardcoded the namespace here. It is a less intuitive, but simpler. Let me know what you think.

src/rgw/rgw_rados.cc
pritha-srivastava: yes, i missed this one, realized after creating the PR, will add this functionality. Subsequent calls to add objects to in a different namespace in the same pool will result in the 'default' namespace, which we dont want.

src/rgw/rgw_rados.cc
pritha-srivastava: I had a question here - now that we are changing the pool structure to have all the metadata in one pool, will that affect the multisite syncing in any way?

cbodley: for multisite, we put some config objects (realm/zone/period) in the .rgw.root pool, and store sync logs in the log_pool. there are some 'list' commands to list config objects, but that stuff should keep working as long as we're calling `set_namespace()` consistently

src/rgw/rgw_common.h
yehudasa: 'ns'?

src/rgw/rgw_common.h
yehudasa: if 9 was merged upstream then we need to bump this one to 10, even if 9 didn't add any encoding changes

yehudasa:
@pritha-srivastava @cbodley the whole bucke -> pool mapping is in general mess (as @dang pointed out before). I think there's another work that needs to be done in finally separating pools from buckets. I'll take a look at doing it.

yehudasa:
@pritha-srivastava missing updating the json encoding of the namespace (via appending it to the pool name)

src/rgw/rgw_rados.cc
```diff
@@ -4183,7 +4247,7 @@ void RGWRados::build_bucket_index_marker(const string& shard_id_str, const strin
 
 int RGWRados::open_bucket_index_ctx(rgw_bucket& bucket, librados::IoCtx& index_ctx)
 {
-  int r = open_pool_ctx(bucket.index_pool, index_ctx);
+  int r = open_pool_ctx(bucket.index_pool, index_ctx, bucket.ns);
```
pritha-srivastava: @cbodley : Should the namespace be set for index pool also? I don't think we should. Please let me know what is correct.

src/rgw/rgw_json_enc.cc
```diff
@@ -555,13 +559,22 @@ void rgw_bucket::dump(Formatter *f) const
...
   JSONDecoder::decode_json("data_extra_pool", data_extra_pool, obj);
   JSONDecoder::decode_json("index_pool", index_pool, obj);
   JSONDecoder::decode_json("marker", marker, obj);
   JSONDecoder::decode_json("bucket_id", bucket_id, obj);
   JSONDecoder::decode_json("tenant", tenant, obj);
+
+  auto pos = bucket_name.find(":");
+  if (pos != string::npos) {
+    name = std::move(bucket_name.substr(0, pos));
+    ns = std::move(bucket_name.substr(pos + 1));
```
cbodley: the temporary `std::string` returned by `substr()` is already an rvalue, so you don't need to cast to rvalue with `std::move()`

src/rgw/rgw_json_enc.cc
```diff
@@ -555,13 +559,22 @@ void rgw_bucket::dump(Formatter *f) const
...
+  } else {
+    name = std::move(bucket_name);
+  }
```
cbodley: consider making this into a helper function, i.e. `set_pool_ns()`, so we don't duplicate this logic in `decode_json()`

src/rgw/rgw_rados.cc
cbodley: i agree - you can't use `bucket.ns` with `bucket.index_pool` as currently designed. i'm actually worried about using it with `data_pool` and `data_extra_pool` too, now that I'm reading through `RGWRados::set_bucket_location_by_rule()`

those pool names are coming from the zone's `RGWZonePlacementInfo`, which i think would also need to be parsed as `pool:namespace`. then it looks like `rgw_bucket` needs to store a separate namespace for each of those pools, right? hopefully @yehudasa's bucket/pool cleanup can help with this

oritwas:
@pritha-srivastava please rebase

cbodley:
@pritha-srivastava this feature did go in with @yehudasa's cleanup work

PR #11353, opened 2016-10-06
librbd: API changes to support separate data pool

dillaman:


dillaman:
This is rebased on top of PR #11327

trociny:
@dillaman pybind test fails:

```
% RBD_FEATURES=127  PYTHONPATH=../src/test/pybind:../src/pybind:./lib/cython_modules/lib.2 LD_LIBRARY_PATH=./lib nosetests -v test_rbd -m 'test_create_defaults' 
test_rbd.test_create_defaults ... FAIL

FAIL: test_rbd.test_create_defaults
    self.test(*self.arg)
  File "/SSD/ceph/ceph.upstream/build/test_rbd.py", line 194, in test_create_defaults
    check_default_params(2, 20, RBD_FEATURE_STRIPINGV2, 0, 0)
  File "/SSD/ceph/ceph.upstream/build/test_rbd.py", line 150, in check_default_params
    eq(expected_features, image.features())
AssertionError: 2 != 0L

Ran 1 test in 6.695s
```

So it fails when creating an image with 'striping' feature and providing stripe_unit and stripe_count = 0. Previously in such a case it used defaults for these parameters and set the feature bit. Now it creates the image without this feature. Also reproducible with rbd cli:
```
% ./bin/rbd create test_striping --size 1 --rbd_default_features 2 --rbd_default_stripe_count=0 --rbd_default_stripe_unit=0
% ./bin/rbd info test_striping
rbd image 'test_striping':
        size 1024 kB in 1 objects
        order 22 (4096 kB objects)
        block_name_prefix: rbd_data.5e692ae8944a
        format: 2
        features: 
        flags: 
% ./bin/rbd create test_striping1 --size 1 --image-feature striping --stripe-count 0 --stripe-unit 0
% ./bin/rbd info test_striping1
rbd image 'test_striping1':
        size 1024 kB in 1 objects
        order 22 (4096 kB objects)
        block_name_prefix: rbd_data.5e672ae8944a
        format: 2
        features: 
        flags: 
# Now using rbd from master:
% ../../ceph/build/bin/rbd create test_striping2 --size 1 --image-feature striping --stripe-count 0 --stripe-unit 0
% ./bin/rbd info test_striping2
rbd image 'test_striping2':
        size 1024 kB in 1 objects
        order 22 (4096 kB objects)
        block_name_prefix: rbd_data.5e752ae8944a
        format: 2
        features: striping
        flags: 
        stripe unit: 4096 kB
        stripe count: 1
```

trociny:
@dillaman Never mind, I was running a wrong test_rbd.py, I see you have already updated it accordingly.

dillaman:
@trociny Yeah, I expanded on the approach where the STRIPINGV2 feature is auto-applied if you specify stripe options for the DATA_POOL feature. However, I didn't want to introduce unnecessary incompatibility if the data pool was the same as the header's pool, so I also auto-removed the flag if the feature wasn't in use.  Given that, I figured it also probably made sense to remove the STRIPINGV2 flag if fancy striping wasn't actually being used (although I know krbd will just ignore it if it's set w/ non-fancy stripe settings).

PR #11357, opened 2016-10-06
mds: use parse_filesystem in parse_role to handle exceptions and reuse parsing code

batrick:
get_filesystem may fail in parse_role if the fscid does not exist.  This
would result in the program (mon) aborting due to the uncaught
exception.

Fixes: http://tracker.ceph.com/issues/17518

### DIFF
--- a/src/mds/FSMap.cc
+++ b/src/mds/FSMap.cc
@@ -498,7 +498,7 @@ int FSMap::parse_filesystem(
   std::string ns_err;
   fs_cluster_id_t fscid = strict_strtol(ns_str.c_str(), 10, &ns_err);
   if (!ns_err.empty() || filesystems.count(fscid) == 0) {
-    for (auto fs : filesystems) {
+    for (auto &fs : filesystems) {
       if (fs.second->mds_map.fs_name == ns_str) {
         *result = std::const_pointer_cast<const Filesystem>(fs.second);
         return 0;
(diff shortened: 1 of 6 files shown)


jcsp:
iirc it was intentional that this would throw if the filesystem didn't exist (otherwise we open the door to segfaults if the callers don't rigorously check nullness of the return value) -- I'd rather have the caller check for existence (and get an exception if they forget to do it) than have the caller check for null returns (and get a segfault if they forget to do it).  I don't claim all the existing code necessarily sticks to that principle though!

Clearly this is a bit inconsistent wrt the name-based get_filesystem call.  The number-based one is for use internally whereas the name based one is always used for handling user input.  I wonder if the name based one should get renamed to "resolve_filesystem" or something to make it less weird.

batrick:
@jcsp please take another look.

src/mds/FSMap.cc
```diff
@@ -491,21 +491,21 @@ void Filesystem::decode(bufferlist::iterator& p)
 
 int FSMap::parse_filesystem(
       std::string const &ns_str,
-      std::shared_ptr<const Filesystem> *result
+      std::shared_ptr<const Filesystem> &result
```
jcsp: Using pointers for output arguments is a style guide thing (https://github.com/ceph/ceph/blob/master/CodingStyle#L127)

batrick:
@jcsp, thanks for pointing that out. I've removed that commit.

batrick:
retest

batrick:
retest this please

PR #11359, opened 2016-10-06
client: add pid to metadata

batrick:
This allows us to distinguish multiple clients on a machine.

Fixes: http://tracker.ceph.com/issues/17276


src/client/Client.cc
```diff
@@ -1941,6 +1941,12 @@ void Client::populate_metadata()
     ldout(cct, 1) << __func__ << " failed to read hostname (" << cpp_strerror(r) << ")" << dendl;
   }
 
+  {
+    std::stringstream ss;
+    ss << getpid();
+    metadata["pid"] = ss.str();
```
batrick: Thanks for the tip. I made the change.

PR #11360, opened 2016-10-06
auth: tolerate missing mgr keys

liewegas:
During upgrade we won't (yet) have mgr keys.  Tolerate this case
and let the client stumble along without them.

This is imperfect (the auth client sends an additional key request and gets EINVAL; not sure why) but the client behaves fine despite it, and it makes the upgrade tests pass.

### DIFF
--- a/src/auth/cephx/CephxClientHandler.cc
+++ b/src/auth/cephx/CephxClientHandler.cc
@@ -74,7 +74,11 @@ int CephxClientHandler::build_request(bufferlist& bl) const
     return 0;
   }
 
-  if (need) {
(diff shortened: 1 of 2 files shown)


liewegas:
http://tracker.ceph.com/issues/17540

liewegas:
Hmm, similar problem if the mons are jewel and osds are kraken, they request keys and jewel osds return EPERM.  I think we need to avoid requesting mgr keys unless the monitor has the kraken server feature.

liewegas:
replaced by https://github.com/ceph/ceph/pull/11401

PR #11362, opened 2016-10-06
mds: respawn using /proc/self/exe

batrick:
This allows the MDS to respawn using the same executable file even if it
has since been deleted (on Linux). Otherwise, the execv fails because
the readlink returns "/path/to/deleted (deleted)". (There is no path to
the old executable.)

Fixes: http://tracker.ceph.com/issues/17531

### DIFF
--- a/src/mds/MDSDaemon.cc
+++ b/src/mds/MDSDaemon.cc
@@ -1113,12 +1113,12 @@ void MDSDaemon::respawn()
   }
   new_argv[orig_argc] = NULL;
 
-  /* Determine the path to our executable, try to read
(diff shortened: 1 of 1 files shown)


jcsp:
Seems reasonable, I must confess that I don't fully understand how /proc/self/exe is still readable to the process itself when it looks like a dead symlink (I just had to write a program to convince myself that it was so)

batrick:
We already discussed this in standup but for posterity: this is basically Linux magic. The "symlink" aspect of /proc/self/exe is only a user-facing procfs detail. You can basically treat it like a legit hard link to your executable file.

PR #11365, opened 2016-10-07
common: Remove the runtime dependency on lsb_release

badone:
With modern releases we should be able to make do with the call to
os_release_parse only which uses /etc/os-release which should be available on
most (all?) releases we currently support. this then allows us to remove the
runtime dependency which pulls in several other packages and would be nice to
avoid.

Fixes: http://tracker.ceph.com/issues/17425
Signed-off-by: Brad Hubbard bhubbard@redhat.com


alfredodeza:
test this please

badone:
The error seems unrelated. If it fails again will rebase this tomorrow as suggested by @alfredodeza 

Thanks mate :)

liewegas:
:+1:

ktdreyer:
:+1:

yuriw:
http://pulpito.ceph.com/yuriw-2016-10-26_22:43:11-rados-wip-yuri-testing2_2016_10_27-distro-basic-smithi/

PR #11370, opened 2016-10-07
rbd: expose rbd unmap options

idryomov:


ktdreyer:
Would you please file a ticket at tracker.ceph.com to track this issue, so we ensure that it's backported to Jewel eventually?

idryomov:
@ktdreyer I was going to cherry-pick this and #11313 in one go and do a backport project ticket for that.  These are iSCSI-related, and so bound to make it to jewel and then downstream regardless...

ktdreyer:
Perfect, thanks @idryomov

PR #11372, opened 2016-10-08
hammer: doc: add "Upgrading to Hammer" section

tchaikov:
Fixes: http://tracker.ceph.com/issues/17534

### DIFF
--- a/doc/install/upgrading-ceph.rst
+++ b/doc/install/upgrading-ceph.rst
@@ -473,6 +473,20 @@ details. To upgrade from a post-Emperor point release, see the `Firefly release
 notes`_ for details.
 
 
+Upgrading to Hammer
+===================
+
+When upgrading from v0.94.6 (and lower) to Hammer (v0.94.7 and higher), you need
+to upgrade the OSD **before** the monitors.
+
+Due to changes in encoding of osdmap in v0.94.7, upgrading monitors before OSDs
+can lead to serious performance issues on large clusters that contain hundreds of
+OSDs. So, we should upgrade OSDs first to work around this issue. It is fixed
+in Jewel and higher versions of Ceph. So you can follow the general steps when
+upgrading from Hammer to Jewel, or from Jewel to Kraken.
+
+
+
 Ceph Config File Changes
 ------------------------
 


tchaikov:
please note this is not a backport from master. it only applies to hammer as a workaround.

we added 3 more fields to `pg_pool_t` between v0.94.6 and v0.94.7, the last field is `use_gmt_hitset`, and a new feature bit is used to track this change. OSDMap has a field of `map<int64_t,pg_pool_t>`.  the CRC check is added since v0.91, where the OSDMap encoding version was bumped from 8. in this change, we disable CRC check by downgrading the OSDMap encoding version to v7 if the connection.features does not have the GMT feature bit. but this could also remove information that matters that gets added after OSDMap v8. so this change can only stay in hammer.

* hammer (<= v0.94.6) => hammer (with this fix): osdmap messages are sent without CRC enabled if peer does not support GMT feature.
* hammer (<= v0.94.6) => jewel (without #11284): should upgrade to hammer (with this fix) first.
* hammer (with this fix) => jewel (with #11284): monitor will encode full map with `features - CEPH_FEATURE_SERVER_JEWEL`. so we need to make sure that in the next jewel release:
  1. all (sub)fields of OSDMap respect the features parameter when `encode()`
  1. be pessimistic if  `CEPH_FEATURE_SERVER_JEWEL` is missing. say, to encode using the first encoding version when the v0.94.4[1] was released. for example, suppose `foo` and `bar` are added into `pg_pool_t` after v0.94.4 is released, and the encoding version of `pg_pool_t` is then v23. we should stick with v17 if `CEPH_FEATURE_SERVER_JEWEL` is missing when encoding `pg_pool_t`.
* jewel (without #11284) => jewel (with #11284): if new fields are added to osdmap in between these two releases. we should ignore CRC mismatches (feature implemented in #11258).

---
[1] v0.94.4 is the first hammer release allowed to upgrade to jewel. see http://ceph.com/releases/v10-2-0-jewel-released/.  and the encoding version of `pg_pool_t` was v17 then.

tchaikov:
tested at tested at http://pulpito.front.sepia.ceph.com:80/kchai-2016-10-08_08:59:22-upgrade-wip-17386-hammer---basic-mira/ and http://pulpito.ceph.com/kchai-2016-10-08_09:07:08-upgrade-wip-17386-hammer---basic-mira/

dachary:
To clarify that http://tracker.ceph.com/issues/17534 was created and made to be a backport of http://tracker.ceph.com/issues/17386. It's fine that it's a workaround instead of cherry-picks. The Fixes: should be changed to http://tracker.ceph.com/issues/17534 instead.

Does that sound sensible ?

tchaikov:
sounds very sensible!

tchaikov:

athanatos:
I thought we weren't going to use this approach for hammer, and instead modify jewel and newer to not include the new bits in the OSDMap until the user flips the switch on the monitor banning old nodes.

tchaikov:
got it. i will repurpose this PR for updating the document to note down the upgrade process of hammer: osd first then mon.

tchaikov:
@athanatos could you take a look again?

athanatos:
LGTM!

tchaikov:
@dachary could you take a look and merge it if appropriate? thanks

PR #11373, opened 2016-10-08
mds: fix false "failing to respond to cache pressure" warning

ukernel:
the false warning happens in following sequence of events
- MDS has cache pressure, sends recall state messages to clients
- Client does not trim as many caps as MDS expected. So MDS
  does not reset session->recalled_at
- MDS no longer has cache pressure, it stop sending recall state
  messages to clients.
- Client does not release its caps. So session->recalled_at in
  MDS keeps unchanged

### DIFF
--- a/src/mds/Beacon.cc
+++ b/src/mds/Beacon.cc
@@ -385,8 +385,10 @@ void Beacon::notify_health(MDSRank const *mds)
   {
     set<Session*> sessions;
     mds->sessionmap.get_client_session_set(sessions);
+
     utime_t cutoff = ceph_clock_now(g_ceph_context);
     cutoff -= g_conf->mds_recall_state_timeout;
+    utime_t last_recall = mds->mdcache->last_recall_state;
 
     std::list<MDSHealthMetric> late_recall_metrics;
     std::list<MDSHealthMetric> large_completed_requests_metrics;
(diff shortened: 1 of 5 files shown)


src/mds/Beacon.cc
```diff
@@ -396,7 +398,10 @@ void Beacon::notify_health(MDSRank const *mds)
         dout(20) << "Session servicing RECALL " << session->info.inst
           << ": " << session->recalled_at << " " << session->recall_release_count
           << "/" << session->recall_count << dendl;
-        if (session->recalled_at < cutoff) {
+	if (last_recall < cutoff || session->recalled_at < last_recall) {
+	  dout(20) << "  no longer recall" << dendl;
+	  session->clear_recalled_at();
```
jcsp: Hmm, so with the new behaviour we will not give a health warning in the case where one client is bad, but another client is trimming enough cache entries to get the overall MDS total back to within limits.  I think that's okay though, I don't see a better way.

jcsp:
This is causing one of the existing tests to fail: http://pulpito.ceph.com/jspray-2016-10-13_11:02:22-fs-wip-jcsp-testing-20161012-distro-basic-mira/471389

(I haven't investigated whether it's bad behaviour or just a test that needs updating)

ukernel:
updated and tested

gregsfortytwo:
assuming it passes tests.

src/mds/Beacon.cc
gregsfortytwo: Given that one client may just legitimately have that many pinned caps, I think it's okay as long as the MDS has enough space in cache for the system as a whole.

PR #10832, opened 2016-08-23
jewel: rgw - default quota fixes

dang:
This is a backport for http://tracker.ceph.com/issues/16447

PR #11374, opened 2016-10-08
jewel: rgw: fix a bug in quota check. 

zhangsw:
Default bucket quota shouldn't be checked if user's bucket quota is enabled. If checked, and default bucket quota's value is less than user's bucket quota's value, the quota check
may fail, which is not right.

Signed-off-by: Zhang Shaowen <zhangshaowen@cmss.chinamobile.com>


dang:
Default quota has already been fixed on master, and is being backported here:
https://github.com/ceph/ceph/pull/10832

PR #11378, opened 2016-10-08
add timeout to watch/notify

ryneli:
Hi @tchaikov, I have added a timeout parameter to watch-related API. Could you please help me to review my code? Thanks.

### DIFF
--- a/src/include/rados.h
+++ b/src/include/rados.h
@@ -517,6 +517,7 @@ struct ceph_osd_op {
 			__le64 ver;     /* no longer used */
 			__u8 op;	/* CEPH_OSD_WATCH_OP_* */
 			__u32 gen;      /* registration generation */
+			__u32 timeout; /* connection timeout */
 		} __attribute__ ((packed)) watch;
 		struct {
 			__le64 cookie;
--- a/src/include/rados.h
+++ b/src/include/rados.h
@@ -562,6 +562,18 @@ struct ceph_osd_op {
 	__le32 payload_len;
 } __attribute__ ((packed));
 
+/*
+ * Check the compatibility of struct ceph_osd_op
+ *  (2+4+(2*8+8+4)+4) = (sizeof(ceph_osd_op::op) +
+ *                     sizeof(ceph_osd_op::flags) +
+ *                     sizeof(ceph_osd_op::extent) +
+ *                     sizeof(ceph_osd_op::payload_len))
+ */
+#ifdef __cplusplus
+static_assert(sizeof(ceph_osd_op) == (2+4+(2*8+8+4)+4),
+              "sizeof(ceph_osd_op) breaks the compatibility");
+#endif
+
 struct ceph_osd_reply_head {
 	__le32 client_inc;                /* client incarnation */
 	__le32 flags;
--- a/src/include/rados/librados.h
+++ b/src/include/rados/librados.h
@@ -2137,7 +2137,6 @@ typedef void (*rados_watchcb2_t)(void *arg,
  * after 30 seconds. Watches are automatically reestablished when a new
  * connection is made, or a placement group switches OSDs.
  *
- * @note BUG: watch timeout should be configurable
  * @note BUG: librados should provide a way for watchers to notice connection resets
  * @note BUG: the ver parameter does not work, and -ERANGE will never be returned
  *            (See URL tracker.ceph.com/issues/2592)
@@ -2163,12 +2162,11 @@ CEPH_RADOS_API int rados_watch(rados_ioctx_t io, const char *o, uint64_t ver,
  * A watch operation registers the client as being interested in
  * notifications on an object. OSDs keep track of watches on
  * persistent storage, so they are preserved across cluster changes by
- * the normal recovery process. If the client loses its connection to
- * the primary OSD for a watched object, the watch will be removed
- * after 30 seconds. Watches are automatically reestablished when a new
+ * the normal recovery process. If the client loses its connection to the
+ * primary OSD for a watched object, the watch will be removed after
+ * a timeout configured with osd_client_watch_timeout.
+ * Watches are automatically reestablished when a new
  * connection is made, or a placement group switches OSDs.
- *
- * @note BUG: watch timeout should be configurable
  *
  * @param io the pool the object is in
  * @param o the object to watch
@@ -2184,6 +2182,30 @@ CEPH_RADOS_API int rados_watch2(rados_ioctx_t io, const char *o, uint64_t *cooki
 				void *arg);
 
 /**
+ * Register an interest in an object
+ *
+ * A watch operation registers the client as being interested in
+ * notifications on an object. OSDs keep track of watches on
+ * persistent storage, so they are preserved across cluster changes by
+ * the normal recovery process. Watches are automatically reestablished when a new
+ * connection is made, or a placement group switches OSDs.
+ *
+ * @param io the pool the object is in
+ * @param o the object to watch
+ * @param cookie where to store the internal id assigned to this watch
+ * @param watchcb what to do when a notify is received on this object
+ * @param watcherrcb what to do when the watch session encounters an error
+ * @param timeout how many seconds the connection will keep after disconnection
+ * @param arg opaque value to pass to the callback
+ * @returns 0 on success, negative error code on failure
+ */
+CEPH_RADOS_API int rados_watch3(rados_ioctx_t io, const char *o, uint64_t *cookie,
+        rados_watchcb2_t watchcb,
+        rados_watcherrcb_t watcherrcb,
+        uint32_t timeout,
+        void *arg);
+
+/**
  * Asynchronous register an interest in an object
  *
  * A watch operation registers the client as being interested in
@@ -2193,8 +2215,6 @@ CEPH_RADOS_API int rados_watch2(rados_ioctx_t io, const char *o, uint64_t *cooki
  * the primary OSD for a watched object, the watch will be removed
  * after 30 seconds. Watches are automatically reestablished when a new
  * connection is made, or a placement group switches OSDs.
- *
- * @note BUG: watch timeout should be configurable
  *
  * @param io the pool the object is in
  * @param o the object to watch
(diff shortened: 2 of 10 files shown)


src/include/rados.h
```diff
@@ -517,6 +517,7 @@ struct ceph_osd_op {
 			__le64 ver;     /* no longer used */
 			__u8 op;	/* CEPH_OSD_WATCH_OP_* */
 			__u32 gen;      /* registration generation */
+			__u32 timeout; /* connection timeout */
```
tchaikov: i was wondering how we handled the backward/forward compatibility here.  the op struct is wrapped in `MOSDOp::ops::op`. and `ceph_osd_op` is encoded using the raw encoder, i.e. just copy the memory chunk. so the maximum size of the member of union in `ceph_osd_op` is `sizeof(writesame)` = (64 + 64 + 64) bits. in other words, as long as the size of all struct are less or equal to 24 bytes, we are safe.

could you add a 
```c++
static_assert(sizeof(ceph_osd_op) == (2+4+(8+8+8)+4),
              "sizeof(ceph_osd_op) breaks the compatibility");
```
somewhere in this header file in a separated commit? in which,
```
 (2+4+(8+8+8)+4) =    (sizeof(ceph_osd_op::op) +
                                      sizeof(ceph_osd_op::flags) +
                                      sizeof(ceph_osd_op::writesame) +
                                      sizeof(ceph_osd_op::payload_len))
```

@athanatos am i right?

src/osd/ReplicatedPG.cc
```diff
@@ -5470,8 +5470,8 @@ int ReplicatedPG::do_osd_ops(OpContext *ctx, vector<OSDOp>& ops)
 	  result = -ENOENT;
 	  break;
 	}
-        uint64_t cookie = op.watch.cookie;
-        entity_name_t entity = ctx->reqid.name;
+	uint64_t cookie = op.watch.cookie;
+	entity_name_t entity = ctx->reqid.name;
```
tchaikov: better off not `s/<tab>/        /`, if you are not touching this area.

ryneli: The original code seems not indent correctly. So I try to modify the indent to align with the code below.

src/include/rados.h
ryneli: @tchaikov, you are right. There is compile error after adding the static check. Do I need to find another place to pass timeout to OSD?

Following is the compile error.

In file included from /home/ryne/ceph/src/include/ceph_fs.h:16:0,
                 from /home/ryne/ceph/src/include/types.h:32,
                 from /home/ryne/ceph/src/msg/msg_types.h:20,
                 from /home/ryne/ceph/src/common/entity_name.h:24,
                 from /home/ryne/ceph/src/common/config.h:24,
                 from /home/ryne/ceph/src/common/dout.h:20,
                 from /home/ryne/ceph/src/compressor/AsyncCompressor.cc:15:
/home/ryne/ceph/src/include/rados.h:573:1: error: static assertion failed: sizeof(ceph_osd_op) breaks the compatibility
 static_assert(sizeof(ceph_osd_op) == (2+4+(8+8+8)+4),

tchaikov: @ryneli i was wrong the largest struct in the union is
```c++
		struct {
          int64_t offset, length;
          int64_t truncate_size;
          int32_t truncate_seq;
		} __attribute__ ((packed)) extent;
```

so the assert should be
```c++
static_assert(sizeof(ceph_osd_op) == (2+4+(2*8+8+4)+4),
```

ryneli: @tchaikov I have finished these changes. Could you please help to review my code again? Also, this static_assert involves some magic number "2+4+(2*8+8+4)+4". Do I need to add a global constant value to represent this magic number, i.e. "const uint32_t MAX_BYTES_OF_CEPH_OSD_OP = 38;"

src/librados/librados.cc
```diff
@@ -4531,17 +4554,26 @@ extern "C" int rados_watch2(rados_ioctx_t io, const char *o, uint64_t *handle,
...
+}
+
+extern "C" int rados_aio_watch2(rados_ioctx_t io, const char *o,
+                                rados_completion_t completion,
+                                uint64_t *handle,
+                                rados_watchcb2_t watchcb,
+                                rados_watcherrcb_t watcherrcb,
+                                uint32_t timeout, void *arg)
 {
   tracepoint(librados, rados_aio_watch_enter, io, o, completion, handle, watchcb, arg);
```
tchaikov: now that `rados_aio_watch2()` is delegated to do the work. we can probably remove `rados_aio_watch_enter` and `rados_aio_watch_exit` tracing points, and add `rados_aio_watch2_enter` and `rados_aio_watch2_exit` to `rados_aio_watch2()`.

src/include/rados/librados.h
```diff
@@ -2212,6 +2231,34 @@ CEPH_RADOS_API int rados_aio_watch(rados_ioctx_t io, const char *o,
...
+ * after the number of seconds that configured in timeout parameter.
+ * Watches are automatically reestablished when a new
+ * connection is made, or a placement group switches OSDs.
+ *
+ * @param io the pool the object is in
+ * @param o the object to watch
+ * @param completion what to do when operation has been attempted
+ * @param handle where to store the internal id assigned to this watch
+ * @param watchcb what to do when a notify is received on this object
+ * @param watcherrcb what to do when the watch session encounters an error
```
tchaikov: `timeout` is not documented.

src/include/rados.h
```diff
@@ -562,6 +563,18 @@ struct ceph_osd_op {
...
 
+/*
+ * Check the compatibility of struct ceph_osd_op
+ *  (2+4+(2*8+8+4)+4) = (sizeof(ceph_osd_op::op) +
+ *                     sizeof(ceph_osd_op::flags) +
+ *                     sizeof(ceph_osd_op::extent) +
+ *                     sizeof(ceph_osd_op::payload_len))
+ */
+#ifdef __cplusplus
+static_assert(sizeof(ceph_osd_op) == (2+4+(2*8+8+4)+4),
```
tchaikov: > Also, this static_assert involves some magic number "2+4+(2*8+8+4)+4". Do I need to add a global constant value to represent this magic number, i.e. "const uint32_t MAX_BYTES_OF_CEPH_OSD_OP = 38;"

ahh, i just realized that `rados.h` is shared by C and C++ source files! `MAX_BYTES_OF_CEPH_OSD_OP` does not seem right. the size of `ceph_osd_op` **must** be exactly 38 bytes. and we can only add more fields at the end of union structs (`extent` excluded), and the new behavior should be backward compatible if the newly added field(s) reads zero. this static assert is just a better-than-nothing message to posterity.

but i think the comment is good enough to serve this purpose. what do you think?

src/include/rados/librados.h
```diff
@@ -2163,13 +2162,11 @@ CEPH_RADOS_API int rados_watch(rados_ioctx_t io, const char *o, uint64_t ver,
  * A watch operation registers the client as being interested in
  * notifications on an object. OSDs keep track of watches on
  * persistent storage, so they are preserved across cluster changes by
- * the normal recovery process. If the client loses its connection to
- * the primary OSD for a watched object, the watch will be removed
- * after 30 seconds. Watches are automatically reestablished when a new
+ * the normal recovery process. The timeout of the connection between
```
tchaikov: can we keep the original doc, how about amending it like:

> If the client loses its connection to the primary OSD for a watched object, the watch will be removed after a timeout configured with `osd_client_watch_timeout`.

what do you think?

tchaikov:
also, probably you can split the first commit into two separated ones. one is the changes in osd.  the other is for exposing the new feature in librados. and then prefix the title of your commit message withs the subcomponent your are changing ? see https://github.com/ceph/ceph/blob/master/SubmittingPatches.rst#preparing-and-sending-patches

ryneli:
@tchaikov, I have split the commits and add prefix to my commit message. Is there anything else to modify?

src/tracing/librados.tp
```diff
@@ -2327,6 +2327,34 @@ TRACEPOINT_EVENT(librados, rados_watch2_exit,
...
+    TP_ARGS(
+        int, retval,
+        uint64_t, handle),
+    TP_FIELDS(
+        ctf_integer(int, retval, retval)
+        ctf_integer(uint64_t, handle, handle)
+    )
+)
+
 TRACEPOINT_EVENT(librados, rados_aio_watch_enter,
```
ryneli: I have removed both rados_watch2_enter/exit and rados_aio_watch_enter/exit :)

tchaikov:
lgtm. adding needs-qa label.

ryneli:
So could I merge it to the master branch now?

tchaikov:
@ryneli, for most non-trivial changes, we'd like to run them through a qa suite before merging them. and in your case, i'd like to run a rados qa suite. that's what the "needs-qa" label is for.

ryneli:
So could I do something in the QA test or just wait the result?

tchaikov:
i think your change is already covered by existing tests. you can push a branch with your changes to ceph/ceph, and schedule a qa run with enough coverage in sepia after your branch is built. or just wait. normally, it will take 1 week or two.

ryneli:
Since this is not an emergent change. Maybe I can just wait for the coverage test. Thanks for your review and comments, @tchaikov

src/tracing/librados.tp
```diff
@@ -2327,25 +2329,27 @@ TRACEPOINT_EVENT(librados, rados_watch2_exit,
...
 
-TRACEPOINT_EVENT(librados, rados_aio_watch_enter,
+TRACEPOINT_EVENT(librados, rados_aio_watch2_enter,
     TP_ARGS(
         rados_ioctx_t, ioctx,
         const char*, oid,
         rados_completion_t, completion,
         uint64_t*, phandle,
         rados_watchcb2_t, callback,
+        uint32_t timeout,
```
tchaikov: @ryneli you missed a `,`. see http://gitbuilder.sepia.ceph.com/gitbuilder-ceph-deb-trusty-amd64-notcmalloc/log.cgi?log=29c70f92ec89d00399952d13145c9fe55daa7a01.

liewegas:
@ryneli ping

ryneli:
@liewegas I am so sorry for the delay. I didn't receive the message. Just find the message when I check the status of this PR.

ryneli:
@tchaikov I am so sorry for the delay. Due to wrong Github setting, I didn't receive your message. I have corrected the error. Could you please help me check it again?

tchaikov:
@ryneli sure, will do.

tchaikov:
lgtm.

tchaikov:
jenkins, retest this please

src/tracing/librados.tp
```diff
@@ -2327,25 +2329,27 @@ TRACEPOINT_EVENT(librados, rados_watch2_exit,
...
         rados_watchcb2_t, callback,
+        uint32_t, timeout,
         void*, arg),
     TP_FIELDS(
         ctf_integer_hex(rados_ioctx_t, ioctx, ioctx)
         ctf_string(oid, oid)
         ctf_integer_hex(rados_completion_t, completion, completion)
         ctf_integer_hex(uint64_t, phandle, phandle)
         ctf_integer_hex(rados_watchcb2_t, callback, callback)
+        ctf_inteter(uint32_t, timeout, timeout)
```
tchaikov: s/ctf_inteter/ctf_integer/

ryneli: Sorry for the stupid typo. Do you know how to test librados.tp file? It seems that the librados.tp file isn't be used when I compile the source code locally.

tchaikov: @ryneli 
```sh
cmake -DWITH_LTTNG=ON ..
make -C src/tracing
```

src/test/librados/watch_notify.cc
```diff
@@ -770,6 +770,79 @@ TEST_P(LibRadosWatchNotifyPP, WatchNotify2Timeout) {
   comp->release();
 }
 
+TEST_P(LibRadosWatchNotifyPP, WatchNotify3) {
+  notify_oid = "foo";
+  notify_ioctx = &ioctx;
+  notify_cookies.clear();
+  uint32_t timeout = 3; // configured timeout
```
tchaikov: http://pulpito.ceph.com/kchai-2016-11-13_05:09:22-rados-wip-kefu-testing---basic-smithi/ 

quite a few tests with harsh cluster env (high messenger failure rate or thrash interruptions) failed due to -ENOTCONN, i am prolonging the timeout to 12 to see if it helps.

tchaikov: the tests at http://pulpito.ceph.com/kchai-2016-11-13_07:03:13-rados-wip-kefu-testing---basic-smithi/ look good.

@ryneli could you change the timeout to `12` or longer?

ryneli:
@tchaikov I have changed the timeout to 12.

PR #11276, opened 2016-09-30
tools/ceph_monstore_tool: update pgmap_meta also when rebuilding store.db

tchaikov:
we should rebuild pgmap_meta table from the collected osdmaps

Fixes: http://tracker.ceph.com/issues/17400
Signed-off-by: Kefu Chai <kchai@redhat.com>

PR #11381, opened 2016-10-09
tools/ceph_monstore_tool: hint when caps is empty

renhwztetecs:
client.admin lack of caps after rebuild,add hint it.
as follow:
caps: [mds] allow *
caps: [mon] allow *
caps: [osd] allow *

### DIFF
--- a/src/tools/ceph_monstore_tool.cc
+++ b/src/tools/ceph_monstore_tool.cc
@@ -532,6 +532,10 @@ static int update_auth(MonitorDBStore& st, const string& keyring_path)
     KeyServerData::Incremental auth_inc;
     auth_inc.name = k.first;
     auth_inc.auth = k.second;
+    if (auth_inc.auth.caps.empty()) {
+      cerr << "no caps granted to: " << auth_inc.name << std::endl;
+      return -EINVAL;
+    }
     auth_inc.op = KeyServerData::AUTH_INC_ADD;
 
     AuthMonitor::Incremental inc;


src/tools/ceph_monstore_tool.cc
```diff
@@ -532,6 +532,13 @@ static int update_auth(MonitorDBStore& st, const string& keyring_path)
     KeyServerData::Incremental auth_inc;
     auth_inc.name = k.first;
     auth_inc.auth = k.second;
+    if (auth_inc.auth.caps.empty()) {
```
tchaikov: i don't want to fill `auth.cap` with the following caps if it is empty, instead, it would be easier for user to customize the behavior of this tool by pointing it to the right keyring file, in which, keys like
```
[mon.]
    key = AQBZKO5XD71DFhAAKM6WEdA94RgqzHcX5+Eajw==
    caps mon = "allow *"
[client.admin]
    key = AQBZKO5XPZJyFxAAJ0B5//6kYxhmxSw1rcciWw==
    auid = 0
    caps mds = "allow *"
    caps mon = "allow *"
    caps osd = "allow *"
```
are expected.

renhwztetecs: Generally:
generating the ceph.client.admin.keyring  when we deploy ceph, and no caps info in the ceph.client.admin.keyring.
so fill auth.cap with the default caps if it is empty, whether in favor of the use of the tool?

renhwztetecs: the default no [mon.]?

tchaikov: to be specific, i would like ceph-monstore-tool to be a flexible low level tool with minimum assumptions. and the functionality to import keyrings when building the mon-store follows the philosophy.

tchaikov: or we should print out an warning message if the caps is empty?

tchaikov: or we should print out a warning message if the caps is empty?

renhwztetecs: yeah, good idea, and write the doc

renhwztetecs:
I update it, and from https://github.com/ceph/ceph/commit/b4bd4004a836121c11b0bb97d8123df54c271f04

tchaikov:
closing this. as it dups with #11276.

PR #11388, opened 2016-10-10
os/ObjectStore: properly clear object map when replaying OP_REMOVE

ukernel:
To remove an object, filestore needs to unlink corresponding object
file from filesystem and removes corresponding object keys from
DBObjectMap. When replaying OP_REMOVE operation, it's possible the
operation has completed partially, object file has been deleted, but
object keys in DBObjectMap hasn't.

The fix is force clear object keys if object file does not exists

Fixes: http://tracker.ceph.com/issues/17177

### DIFF
--- a/src/os/filestore/FileStore.cc
+++ b/src/os/filestore/FileStore.cc
@@ -472,13 +472,7 @@ int FileStore::lfn_unlink(const coll_t& cid, const ghobject_t& o,
     }
 
     if (!force_clear_omap) {
-      if (hardlink == 0) {
-          if (!m_disable_wbthrottle) {
-	    wbthrottle.clear_object(o); // should be only non-cache ref
-          }
-	  fdcache.clear(o);
-	  return 0;
-      } else if (hardlink == 1) {
+      if (hardlink == 0 || hardlink == 1) {
 	  force_clear_omap = true;
       }
     }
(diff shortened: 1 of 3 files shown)


src/os/filestore/FileStore.cc
```diff
@@ -472,13 +472,7 @@ int FileStore::lfn_unlink(const coll_t& cid, const ghobject_t& o,
...
 
     if (!force_clear_omap) {
-      if (hardlink == 0) {
-          if (!m_disable_wbthrottle) {
-	    wbthrottle.clear_object(o); // should be only non-cache ref
-          }
-	  fdcache.clear(o);
-	  return 0;
-      } else if (hardlink == 1) {
+      if (hardlink == 0 || hardlink == 1) {
```
tchaikov: we always clear omap first, and **then** unlink the object file. so, if the `hardlink` is  1 here, we should have cleared the omap already, am i right?

ukernel:
ping @athanatos  @liewegas @tchaikov

src/os/filestore/FileStore.cc
liewegas: So to clarify the situation is

 - clear omap
 - unlink
 - unlink persists to disk, but omap does not
 - <crash>
 - replay sees hardlink 0 and doesn't clear omap

?

ukernel: yes

liewegas:
lgtm!

athanatos:
lgtm! (Whoa, good catch!)

tchaikov:
lgtm also.

badone:
Thanks @ukernel Hopefully we can get this merged soon.

yuriw:
http://pulpito.ceph.com/yuriw-2016-10-21_00:30:49-rados-wip-yuri-testing_2016_10_21-distro-basic-smithi/

http://pulpito.front.sepia.ceph.com/yuriw-2016-10-21_15:49:39-rados-wip-yuri-testing_2016_10_21-distro-basic-smithi/

PR #11389, opened 2016-10-10
jewel: mon: don't crash on invalid standby_for_fscid

jcsp:
http://tracker.ceph.com/issues/17557


dachary:
tested at http://pulpito.ceph.com/jspray-2016-10-10_15:27:25-fs-wip-jcsp-backports-20161010-distro-basic-mira/

PR #11391, opened 2016-10-10
AArch64: Detect crc32 extension support from assembler

agraf:
The used compiler may or may not be recent enough to recognize the
crc32 extended cpu type. However, it does not really have to know about
them either, since all we do is pass inline assembly instructions to
the assembler.

This patch moves the crc cpu extension detection from compiler based
to assembler based, so that we can build optimized code even when the
compiler does not know about the cpu type yet.

Fixes: http://tracker.ceph.com/issues/17516
Signed-off-by: Alexander Graf <agraf@suse.de>

### DIFF
--- a/cmake/modules/SIMDExt.cmake
+++ b/cmake/modules/SIMDExt.cmake
@@ -16,9 +16,20 @@
 
 if(CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64|AARCH64")
   set(HAVE_ARM 1) 
-  CHECK_C_COMPILER_FLAG(-march=armv8-a+crc HAVE_ARMV8_CRC)
+  set(save_quiet ${CMAKE_REQUIRED_QUIET})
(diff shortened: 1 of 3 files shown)


smithfarm:
@tchaikov Could you include this in your next QA run?

tchaikov:
i am not familiar with the aarch64 assembly, but the check seems sane. pushing to ceph/ceph as "wip-crc-aarch64" to get it built by gitbuilders.

@yghannam does this change looks good to you?

yghannam:
The assembly looks good. It's the same as we're using in the actual CRC32 computation.

Has this been tested on ARMv8 platforms with the CRC32C instructions and on platforms without?

agraf:
@yghannam, the patch doesn't touch the assembly, so I'm not quite sure where your question leads. But yes, it's been tested.

yghannam:
@agraf, right. Sorry I was thinking of something else.

LGTM

PR #11396, opened 2016-10-10
[DNM] rbd-mirror HA: create a generic/re-usable managed lock class

rjfd:
Ticket: http://tracker.ceph.com/issues/17016


rjfd:
Switched this implementation to another PR #11592

PR #11397, opened 2016-10-10
Wip scrub misc

dzafman:
For Kraken from previous work:
    Scrub fixes for semantics of snapshot results.  Add test of snapset corruption.
    Minor cleanup and logging improvements.

### DIFF
--- a/src/osd/ReplicatedBackend.cc
+++ b/src/osd/ReplicatedBackend.cc
@@ -789,7 +789,8 @@ void ReplicatedBackend::be_deep_scrub(
   ScrubMap::object &o,
   ThreadPool::TPHandle &handle)
 {
-  dout(10) << __func__ << " " << poid << " seed " << seed << dendl;
+  dout(10) << __func__ << " " << poid << " seed " 
+	   << std::hex << seed << std::dec << dendl;
   bufferhash h(seed), oh(seed);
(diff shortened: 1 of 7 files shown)


dzafman:
Jenkins only 1 unrelated failure

  79/146 Test  #86: unittest_erasure_code_plugin_jerasure ...***Exception: SegFault  0.22 sec

src/osd/ReplicatedBackend.cc
```diff
@@ -789,7 +789,8 @@ void ReplicatedBackend::be_deep_scrub(
   ScrubMap::object &o,
   ThreadPool::TPHandle &handle)
 {
-  dout(10) << __func__ << " " << poid << " seed " << seed << dendl;
+  dout(10) << __func__ << " " << poid << " seed "
```
tchaikov: no need to restore the iostream state by sending `std::dec` to it. `dout` creates a new `ceph::log::Entry` instance. for every `dout()`. the `ostream` under the hood is the `ceph::log::Entry::m_streambuf` variable, and they are not shared.

dzafman: I'm going to leave it so it is there should anything be added to the end of that message.

dzafman:
jenkins test this please

PR #11401, opened 2016-10-10
auth: tolerate missing MGR keys during upgrade

liewegas:


tchaikov:
this fixes http://tracker.ceph.com/issues/17540

lgtm as long as the upgrade test passes.

PR #11403, opened 2016-10-10
cmake: find gperftools package for tcmalloc_minimal too

bassamtabbara:


CMakeLists.txt
```diff
@@ -257,7 +257,7 @@ find_package(snappy REQUIRED)
 
 #if allocator is set on command line make sure it matches below strings
 if(ALLOCATOR)
-  if(${ALLOCATOR} STREQUAL "tcmalloc")
+  if(${ALLOCATOR} STREQUAL "tcmalloc" OR ${ALLOCATOR} STREQUAL "tcmalloc_minimal")
```
tchaikov: nit, could use 
```cmake
if(${ALLOCATOR} MATCHES "tcmalloc(_minimal)?")
```

bassamtabbara: done.

PR #11404, opened 2016-10-10
cmake: Fix for cross compiling

bassamtabbara:
The check for yasm tool was calling uname -m which will not
work when cross compiling. Use CMAKE_SYSTEM_PROCESSOR instead.


tchaikov:
pushed to wip-kefu-testing for testing

tchaikov:
test passed on amd64 and aarch64.

PR #11405, opened 2016-10-10
cmake: make python package optional

bassamtabbara:
added an option to exclude finding the python package. The default
behavior is to still look for Python2. This is option is helpful
when building in containers or other restricted environments, and
you dont want to pull in mighty python. For example, make ceph-mon
builds without python.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -194,6 +194,8 @@ if(WITH_CEPHFS_JAVA)
 endif()
 
 # Python stuff
+option(WITH_PYTHON "build python bindings" ON)
+if(WITH_PYTHON)
 find_package(PythonInterp 2 REQUIRED)
 find_package(PythonLibs 2 REQUIRED)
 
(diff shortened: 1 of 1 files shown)


src/CMakeLists.txt
```diff
@@ -549,7 +552,10 @@ if(NOT CEPH_BUILD_VIRTUALENV)
   set(CEPH_BUILD_VIRTUALENV /tmp)
 endif()
 
+if(WITH_PYTHON)
 add_subdirectory(pybind)
+endif()
+
```
bassamtabbara: I was hoping not chase down every use of python in the tree, but instead just assume that the user can call a specific target when invoking make that does not require python, for example

```
mkdir build && cd build
cmake .. -DWITH_PYTHON=OFF
make ceph-mon ceph-osd
```

As you point out python is assumed throughout the code base and it would be difficult to chase it all down.

bassamtabbara:
@tchaikov I excluded more subdirs based on your feedback. The goal here is *build* without python available and not necessarily have tests pass etc. without python. The python dependency is quite deep in the ceph tree.

The following works for me:

```
mkdir build && cd build
cmake .. -DWITH_PYTHON=OFF
make -j4
```

bassamtabbara:
@tchaikov any more thoughts on this one?

tchaikov:
> The python dependency is quite deep in the ceph tree.

@bassamtabbara i understand. and your requirement also makes sense. but personally, i am just not very convinced that this is what we want with the `WITH_PYTHON` option. i'd like to wait for more opinions on this change.

bassamtabbara:
@tchaikov I'm happy to try something different or use a different option. What are you concerns with WITH_PYTHON?

bassamtabbara:
closing after discussion with @tchaikov on IRC.

PR #11406, opened 2016-10-10
osd: recover broken maps automatically

jecluis:
We've seen situations in which the osdmaps or incremental maps happen to be completely zeroed out or end up getting corrupt after a series of server failures.

In some of these situations, the osds were unable to boot, dying horrible deaths on init. Recovering from these failures were just a matter of copying the right maps from a healthy osd to the affected osds, but this still had to be a manual task.

This series of patches solves the issue by automating the map recovery, by requesting the broken maps from the monitors and writing them to disk.

Recovery is not enabled by default and depends on an experimental feature ("osd_map_recover_broken") to be enabled, thus delegating to the administrator the responsibility of opting to use the automated recovery. Furthermore, we also add the possibility of proceeding with normal osd init after recovering the maps, although, again, this is not enabled by default.

### DIFF
--- a/src/osd/OSD.h
+++ b/src/osd/OSD.h
@@ -2499,9 +2499,9 @@ public:
   friend class OSDService;
 
   struct corrupt_map : public std::exception {
-    epoch_t epoch;
+    const epoch_t epoch;
     explicit corrupt_map(epoch_t e) : epoch(e) { }
-    const char *what() const throw();
+    const char *what() const override noexcept;
   };
 };
 
--- a/src/osd/OSD.h
+++ b/src/osd/OSD.h
@@ -1016,11 +1016,11 @@ public:
   }
   void pin_map_bl(epoch_t e, bufferlist &bl);
   void _add_map_bl(epoch_t e, bufferlist& bl);
-  bool get_map_bl(epoch_t e, bufferlist& bl) {
+  bool get_map_bl(epoch_t e, bufferlist& bl, bool cache = true) {
     Mutex::Locker l(map_cache_lock);
-    return _get_map_bl(e, bl);
+    return _get_map_bl(e, bl, cache);
   }
-  bool _get_map_bl(epoch_t e, bufferlist& bl);
+  bool _get_map_bl(epoch_t e, bufferlist& bl, bool cache = true);
 
   void add_map_inc_bl(epoch_t e, bufferlist& bl) {
     Mutex::Locker l(map_cache_lock);
@@ -1028,7 +1028,7 @@ public:
   }
   void pin_map_inc_bl(epoch_t e, bufferlist &bl);
   void _add_map_inc_bl(epoch_t e, bufferlist& bl);
-  bool get_inc_map_bl(epoch_t e, bufferlist& bl);
+  bool get_inc_map_bl(epoch_t e, bufferlist& bl, bool cache = true);
 
   void clear_map_bl_cache_pins(epoch_t e);
 
--- a/src/osd/OSD.h
+++ b/src/osd/OSD.h
@@ -2472,6 +2472,8 @@ private:
 
   int init_op_flags(OpRequestRef& op);
 
+  int _recover_broken_maps();
+
 public:
   static int peek_meta(ObjectStore *store, string& magic,
 		       uuid_d& cluster_fsid, uuid_d& osd_fsid, int& whoami);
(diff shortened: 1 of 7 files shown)


jecluis:
weird. jenkins build didn't finish, and yet _it works for me_. I'll try a clean build in the morning and sort it out.

jecluis:
i'm guessing jenkins merges the pull request onto master and then builds. the branch was rebased on top of master and an additional commit, meant to be fixed-up with an earlier commit, was added. Commit left as is for the benefit of the reviewer.

src/osd/OSDMapRecovery.cc
```diff
@@ -0,0 +1,463 @@
...
+
+  for (auto e = first; e <= last; e ++) {
+
+    bool broken = (inc ? _is_broken_inc(e) : _is_broken_map(e));
+
+    if (broken) {
+      dout(0) << __func__ << " " << (inc ? "inc" : "full")
+              << " epoch " << e << " broken" << dendl;
+
+      auto &l = ranges.back();
```
tchaikov: > Calling back on an empty container is undefined.

maybe we can note down the upper bound of missing/corrupted map? and only try to append to it when `ranges` is not empty or the upper bound is not "-1" or some sentry number?

src/osd/OSDMapRecovery.cc
```diff
@@ -0,0 +1,463 @@
...
+ *
+ * 'init' gets basic requirements set up
+ *
+ *  - a dedicated messenger, so we don't mess with the OSDs own client ms
+ *  - a mon client, dully authenticated
+ *
+ */
+int OSDMapRecovery::init()
+{
+  int auth_attempts = 0;
```
jecluis: I could, but then we'd be crossing the initialization with the gotos a few lines down.

dachary:

dachary:
genuine error at https://jenkins.ceph.com/job/ceph-pull-requests/14310/

src/osd/OSD.h
```diff
@@ -2495,6 +2497,12 @@ class OSD : public Dispatcher,
 public:
   OSDService service;
   friend class OSDService;
+
+  struct corrupt_map : public std::exception {
+    const epoch_t epoch;
+    explicit corrupt_map(epoch_t e) : epoch(e) { }
+    const char *what() const override noexcept;
```
tchaikov:
```
/home/jenkins-build/build/workspace/ceph-pull-requests/src/osd/OSD.h:2511:30: error: expected ‘;’ at end of member declaration
     const char *what() const override noexcept;
                              ^
/home/jenkins-build/build/workspace/ceph-pull-requests/src/osd/OSD.h:2511:39: error: expected unqualified-id before ‘noexcept’
     const char *what() const override noexcept;
                                       ^
/home/jenkins-build/build/workspace/ceph-pull-requests/src/osd/OSD.h:2511:17: error: looser throw specifier for ‘virtual const char* OSD::corrupt_map::what() const’
     const char *what() const override noexcept;
                 ^
In file included from /usr/include/c++/4.8/new:40:0,
                 from /usr/include/c++/4.8/ext/new_allocator.h:33,
                 from /usr/include/x86_64-linux-gnu/c++/4.8/bits/c++allocator.h:33,
                 from /usr/include/c++/4.8/bits/allocator.h:46,
                 from /usr/include/c++/4.8/vector:61,
                 from /home/jenkins-build/build/workspace/ceph-pull-requests/src/common/config.h:19,
                 from /home/jenkins-build/build/workspace/ceph-pull-requests/src/objclass/class_api.cc:4:
/usr/include/c++/4.8/exception:68:25: error:   overriding ‘virtual const char* std::exception::what() const noexcept (true)’
     virtual const char* what() const _GLIBCXX_USE_NOEXCEPT;
```

should be
```c++
const char *what() const noexcept override;
```

liewegas:
This doesn't feel like a good candidate for merging upstream.  Maybe it's just because I haven't seen this particular failure condition.. is it that common?  Why were the maps corrupted in the first place?

Especially with bluestore, we have crcs on the map objects, so we'll notice if they are garbled on disk.  That doesn't fix the problem (the osd would need to re-request it from the mon in that case), but that seems like a much more limited workaround/fix ("if i get eio reading a map, re-fetch it from the mon") than all of this infrastructure.

What do you think?

jdurgin:
Seems to be a large amount of complexity for a rare situation. Bluestore makes this much better as Sage mentioned, so closing.

PR #11409, opened 2016-10-11
jewel: librados memory leaks from ceph::crypto (WITH_NSS)

dachary:
http://tracker.ceph.com/issues/17341

### DIFF
--- a/src/common/ceph_context.cc
+++ b/src/common/ceph_context.cc
@@ -583,8 +583,10 @@ void CephContext::put() {
 
 void CephContext::init_crypto()
 {
-  ceph::crypto::init(this);
-  _crypto_inited = true;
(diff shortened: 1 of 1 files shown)


dachary:
@oritwas does this backport look good to merge ? It passed the rados (http://tracker.ceph.com/issues/17487#note-12), upgrade/jewel-x and upgrade/hammer-x (http://tracker.ceph.com/issues/17487#note-15) suites (except for two bugs that are, I believe, unrelated)

oritwas:
lgtm

liupan1111:
@dachary, this is a good fix. But have one question: is it better to set this init_crypto as private function, and be called inside the CephContext construction func?

PR #11411, opened 2016-10-11
jewel: tests: scsi_debug fails /dev/disk/by-partuuid

dachary:
http://tracker.ceph.com/issues/17245


dachary:
Passed the ceph-disk suite http://tracker.ceph.com/issues/17487#note-23

PR #11417, opened 2016-10-11
jewel: ceph-create-keys: sometimes blocks forever if mds allow is set

dachary:
http://tracker.ceph.com/issues/17347


dachary:
@liewegas does this backport look good to merge ? It passed the rados (http://tracker.ceph.com/issues/17487#note-12), upgrade/jewel-x and upgrade/hammer-x (http://tracker.ceph.com/issues/17487#note-15) suites (except for two bugs that are, I believe, unrelated)

PR #11421, opened 2016-10-11
Small pile of random cephfs fixes and cleanup

jtlayton:
This is a small patch pile of random bugfixes and cleanups. Most of the bugs fixed are either in the new statx handling code, though there are a couple of fixes to bugs introduced during the UserPerm conversion as well.

There is also a fix for what appears to be an inode refcount leak in the ceph_ll_link code, that I found by inspection. Please take a hard look at that one and make sure I'm not missing anythingt here.

Finally, this also converts most of the internal libcephfs callers to use the statx-based APIs. For now, I've left the java and python APIs the same, but converted them to use statx APIs internally. Eventually we may want to convert them to something more statx-like. That means fixing up their callers though and I'm a little unclear on how best to audit all of them.

### DIFF
--- a/src/client/Client.cc
+++ b/src/client/Client.cc
@@ -6842,12 +6842,10 @@ void Client::fill_statx(Inode *in, unsigned int mask, struct ceph_statx *stx)
   /* These are always considered to be available */
   stx->stx_dev = in->snapid;
   stx->stx_blksize = MAX(in->layout.stripe_unit, 4096);
-  stx->stx_mode = S_IFMT & in->mode;
 
-  if (use_faked_inos())
-   stx->stx_ino = in->faked_ino;
-  else
-    stx->stx_ino = in->ino;
+  /* Type bits are always set, even when CEPH_STATX_MODE is not */
+  stx->stx_mode = S_IFMT & in->mode;
+  stx->stx_ino = use_faked_inos() ? in->faked_ino : (ino_t)in->ino;
   stx->stx_rdev = in->rdev;
   stx->stx_mask |= (CEPH_STATX_INO|CEPH_STATX_RDEV);
 
--- a/src/client/Client.cc
+++ b/src/client/Client.cc
@@ -6658,7 +6658,7 @@ int Client::setattrx(const char *relpath, struct ceph_statx *stx, int mask,
 
   filepath path(relpath);
   InodeRef in;
-  int r = path_walk(path, &in, perms, flags & AT_SYMLINK_NOFOLLOW);
+  int r = path_walk(path, &in, perms, !(flags & AT_SYMLINK_NOFOLLOW));
   if (r < 0)
     return r;
   return _setattrx(in, stx, mask, perms);
@@ -6738,7 +6738,7 @@ int Client::statx(const char *relpath, struct ceph_statx *stx,
 
   unsigned mask = statx_to_mask(flags, want);
 
-  int r = path_walk(path, &in, perms, flags & AT_SYMLINK_NOFOLLOW, mask);
+  int r = path_walk(path, &in, perms, !(flags & AT_SYMLINK_NOFOLLOW), mask);
   if (r < 0)
     return r;
 
--- a/src/client/Client.cc
+++ b/src/client/Client.cc
@@ -6047,8 +6047,7 @@ int Client::get_or_create(Inode *dir, const char* name,
 }
 
 int Client::path_walk(const filepath& origpath, InodeRef *end,
-		      const UserPerm& perms, bool followsym,
-		      int mask, int uid, int gid)
+		      const UserPerm& perms, bool followsym, int mask)
 {
   filepath path = origpath;
   InodeRef cur;
--- a/src/client/Client.cc
+++ b/src/client/Client.cc
@@ -9905,7 +9905,7 @@ Inode *Client::ll_get_inode(vinodeno_t vino)
   return in;
 }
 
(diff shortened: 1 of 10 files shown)


src/client/Client.cc
```diff
@@ -8675,10 +8675,6 @@ int Client::_write(Fh *f, int64_t offset, uint64_t size, const char *buf,
   // use/adjust fd pos?
   if (offset < 0) {
     lock_fh_pos(f);
-    /*
-     * FIXME: this is racy in that we may block _after_ this point waiting for caps, and size may
-     * change out from under us.
-     */
```
jtlayton: Probably could flesh out the justification for removing this a bit. Basically in order to do the _lseek to SEEK_END, you need to fetch the file size. So _lseek does a _getattr with CEPH_STAT_CAP_SIZE (aka CEPH_CAP_FILE_SHARED). At that point the size is effectively frozen until the client returns that cap back to the MDS, but that can't happen until the client_lock mutex is released.

src/client/Client.cc
```diff
@@ -11441,7 +11441,6 @@ int Client::ll_link(Inode *in, Inode *newparent, const char *newname,
   if (r == 0) {
     assert(target);
     fill_stat(target, attr);
-    _ll_get(target.get());
```
gregsfortytwo: Can you pull out this patch for later discussion? You're probably right we have a problem here with non-fuse callers, but this fix doesn't look right either and isn't really in theme with the others.

src/client/Client.cc
```diff
@@ -8115,18 +8115,14 @@ loff_t Client::lseek(int fd, loff_t offset, int whence, const UserPerm& perms)
...
 
-loff_t Client::_lseek(Fh *f, loff_t offset, int whence, const UserPerm& perms)
+loff_t Client::_lseek(Fh *f, loff_t offset, int whence)
 {
   Inode *in = f->inode.get();
   int r;
 
-  if (!may_open(f->inode.get(), f->flags, perms)) {
-    return -EPERM;
-  }
```
gregsfortytwo: Do we have any idea yet *why* this was causing breakage?

Maybe the cmount->default_perms in libcephfs weren't matching up with what was being used elsewhere?

src/client/Client.cc
jtlayton: My understanding (and if I'm wrong, please let me know) is that at the end of the _getattr(), we'll hold the shared caps that were requested (Fs in this case). Since we won't release the mutex again before we issue the write, we'll still hold Fs when we go to do the write.

Is that wrong?

src/client/Client.cc
jtlayton: Sure, will do. It's pretty orthogonal to the rest of the set.

src/client/Client.cc
jtlayton: No. I'm not sure how to tell from the teuthology logs -- could be anything really. This did deal in java after all, and I'm not 100% sure how the jre handles process cred changes. Might two threads have hit a name collision and be running different tests vs. the same files?

Also, the EPERM here clobbers the return from may_open, which might have told us a bit more about the failure. If you're curious we could change that to log a warning when that occurs without failing the operation, and see if we can catch it in the act?

src/client/Client.cc
```diff
@@ -6656,7 +6656,7 @@ int Client::setattrx(const char *relpath, struct ceph_statx *stx, int mask,
 
   filepath path(relpath);
   InodeRef in;
-  int r = path_walk(path, &in, perms, flags & AT_SYMLINK_NOFOLLOW);
+  int r = path_walk(path, &in, perms, !(flags & AT_SYMLINK_NOFOLLOW));
```
gregsfortytwo: Oh right, this is new code and it was added there.

src/client/Client.cc
gregsfortytwo: We may have high-debug client logs? But probably logging the warning there would be easier.

src/client/Client.cc
jtlayton: Well...we do fetch the size there rather early. What we probably could do is fetch extra caps if we know we'll need to do an O_APPEND write, and then delay setting up the offset until just before then, so that we don't lose them when we go to issue it.

That's a bigger patch though. I'll just drop this patch for now and just leave the comment there, but I'll also open a tracker bug to see if we can improve the O_APPEND semantics with a scheme like that.

src/client/Client.cc
jtlayton: Maybe. OTOH...do we really want to incur that permissions check? Those are not exactly cheap, and I don't see myself spending a lot of time tracking down something that's not necessarily a bug anyway.

jtlayton:
Rebased onto master, and stripped out the patch to remove the FIXME comment and the Inode reference leak in the link codepath.

jtlayton:
One more commit to remove a redundant caps check.

gregsfortytwo:
Well, I still think that lseek thing is an information leak, but I guess we can fix it if/when everybody else does, so all LGTM.

@jtlayton, I assume you've tested this and it won't blow up when John runs it through an integration suite? ;)

jtlayton:
Yeah, here's the latest run. Most of the failures are known stuff (ThreesomeInterProcessLocking test in particular), and the others look like problems getting enough mira nodes.

http://pulpito.ceph.com/jlayton-2016-10-14_01:19:59-fs-wip-jlayton-cephfs---basic-mira/

PR #11422, opened 2016-10-11
os/bluestore: fix race condtion during blob spliting

xiexingguo:


src/os/bluestore/BlueStore.cc
```diff
@@ -1665,31 +1665,35 @@ void BlueStore::ExtentMap::reshard(Onode *o, uint64_t min_alloc_size)
 	// shard boundary.
 	bool must_span = false;
 	BlobRef b = ep->blob;
-	if (b->can_split()) {
```
liewegas: I think a simpler fix would be to take the cache lock *just* for the can_split() writing.empty() check.  We have the collection lock so no new writes will race with us.. only IO completion.  So if writing is empty now it's safe the drop the lock and assume it will still be empty below when we do the split.

liewegas:
The other patches look good.  Thanks!

xiexingguo:
@liewegas Done.

PR #11423, opened 2016-10-11
[DNM] rbd-mirror HA: add new lock released/acquired and heartbeat messages

trociny:
Fixes: http://tracker.ceph.com/issues/17018
Signed-off-by: Mykola Golub mgolub@mirantis.com

### DIFF
--- a/qa/workunits/rbd/rbd_mirror_ha.sh
+++ b/qa/workunits/rbd/rbd_mirror_ha.sh
@@ -1,0 +1,135 @@
+#!/bin/sh
+#
+# rbd_mirror_ha.sh - test rbd-mirror daemons in HA mode
+#
+
+. $(dirname $0)/rbd_mirror_helpers.sh
+
+wait_for_leader()
+{
+    local s instance
+
+    for s in 1 2 4 4 4 4 4 4; do
+	sleep $s
+	for instance in `seq 0 9`; do
+	    admin_daemon "${CLUSTER1}:${instance}" \
+			 rbd mirror status ${POOL} ${CLUSTER2} |
+		grep '"leader": true' || continue
+	    LEADER=${instance}
+	    return 0
+	done
+    done
+
+    return 1
+}
+
+test_replay()
+{
+    local image
+
+    for image; do
+	wait_for_image_replay_started ${CLUSTER1}:${LEADER} ${POOL} ${image}
+	write_image ${CLUSTER2} ${POOL} ${image} 100
+	wait_for_replay_complete ${CLUSTER1}:${LEADER} ${CLUSTER2} ${POOL} \
+				 ${image}
+	wait_for_status_in_pool_dir ${CLUSTER1} ${POOL} ${image} 'up+replaying' \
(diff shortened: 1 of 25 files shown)


dachary:
needs rebase :-)

trociny:
@dillaman I have updated it to use librbd::ManagedLock. So, it basically works now (passes rbd_mirror_ha.sh workunit tests), still not worth a thorough review I think until I rebase it on your #12364 (I expect many changes).

I have a question though. A missing bit is blacklisting a dead leader. For this I would need to know its address. E.g. we could extend ManagedLock interface to return the current owner address, so I could use it (or may be blacklisting should be a part of ManagedLock functionality?) or the address could be sent in the heartbeat messages.  What do you suggest?

dillaman:
@trociny I would suggest adding a "blacklist_lock_owner()" (or similar) method to ```ManagedLock``` and moving the blacklisting logic from ```managed_lock::AcquireRequest``` to its own ```managed_lock::BlacklistRequest```.

trociny:
@dillaman
Working on improving the case when there is no heartbeat from the leader. The plan is to:
1) blacklist the previous lock owner;
2) try acquire the lock.

I use try_acquire_lock instead of acquire_lock, as it may block because the lock has already been acquired by another non-leader. Is it what you supposed?

Earlier we discussed to extend ManagedLock interface to provide "blacklist_lock_owner()"  that could be used for blacklisting. But now I am not sure how to use it. The lock might already be reacquired by another rbd-mirror daemon (so the owner is changed), while I need to blacklist the original owner (or skip blacklisting in this case). I think I still need to cache the address of the previous leader, so I could blacklist it either via rados interface or ManagedLock (not sure how much it could be useful in this case to extend it for this functionality). 

Also, considering a situation, when currently the lock is hold by a non-responding (no heatbeat) leader. Is it enough just to blacklist it for acquire_lock to succeed? I suppose it will succeed only after the watcher timeout? I suppose it could be improved by sending break_lock but this looks racy (the lock might be already acquired by another alive daemon).

dillaman:
@trociny I was thinking that you should be able to use ```acquire_lock``` in the derived class, but since right now there really is no difference in the base class between ```try_acquire``` and ```acquire```, it shouldn't matter (or that should be fixed in ```ManagedLock``` so that it only invokes the derived class's virtual callback function when in LOCK mode and fails fast when in TRY_LOCK mode).

You would want to break the lock when you decide that the current lock owner isn't responding. I should have used that term instead of blacklist since blacklisting is just one step in breaking the lock (i.e. I was attempting to say move the current break lock logic out into its own state machine class).

I agree with your conclusion that w/o being able to cache the last known lock owner, attempting to break the current lock owner would be racy. That would imply that you would need to cache the current lock owner after a failed lock attempt. If you notice that the lock owner changes, you would reset your "dead owner" timer. That implies that the break lock state machine needs to support an (optional) lock owner filter where the state machine can be restricted to breaking the lock of only a specific client (or failing if a new client owns the lock).

PR #11426, opened 2016-10-11
rgw: loses realm/period/zonegroup/zone data: period overwritten if somewhere in the cluster is still running Hammer

oritwas:
Fix corruption of configuration when running an old radosgw-admin (hammer or older) on a new radosgw (jewel or newer)

Fixes: http://tracker.ceph.com/issues/17371


cbodley:
ceph.com is down so I can't read the ticket, but this looks correct. older radosgw-admin will recreate region objects, but new radosgw won't re-convert them if it sees this new 'converted' object :+1:

dachary:
jenkins test this please

PR #11428, opened 2016-10-11
src/CMakeLists.txt: Moved some more modules to generic build.

wjwithagen:
FreeBSD needs these modules also when building.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -49,6 +49,9 @@ if(NOT CMAKE_BUILD_TYPE STREQUAL Debug)
       CMAKE_C_FLAGS_${build_type_upper})
     string(REGEX REPLACE "(^| )[/-]D *NDEBUG($| )" " " "${flags}" "${${flags}}")
   endforeach()
+else()
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${CMAKE_CXX_FLAGS_DEBUG} -O0 -g" )
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${CMAKE_C_FLAGS_DEBUG} -O0 -g" )
 endif()
 
 include(CheckCCompilerFlag)
@@ -431,6 +434,9 @@ set(libcommon_files
   common/Graylog.cc
   common/fs_types.cc
   common/dns_resolve.cc
+  common/armor.c
+  common/safe_io.c
+  common/addr_parsing.c
   ${arch_files}
   ${auth_files}
   $<TARGET_OBJECTS:compressor_objs>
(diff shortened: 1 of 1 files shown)


src/CMakeLists.txt
```diff
@@ -49,6 +49,9 @@ if(NOT CMAKE_BUILD_TYPE STREQUAL Debug)
       CMAKE_C_FLAGS_${build_type_upper})
     string(REGEX REPLACE "(^| )[/-]D *NDEBUG($| )" " " "${flags}" "${${flags}}")
   endforeach()
+else()
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${CMAKE_CXX_FLAGS_DEBUG} -O0 -g" )
```
tchaikov: user can always customize `CMAKE_CXX_FLAGS` and `CMAKE_C_FLAGS`  explicitly.  so again, i'd suggest remove this change.

src/CMakeLists.txt
```diff
@@ -470,8 +478,13 @@ if(ENABLE_SHARED)
 endif(ENABLE_SHARED)
 
 add_library(common STATIC ${libcommon_files}
-  $<TARGET_OBJECTS:mon_common_objs>
-  $<TARGET_OBJECTS:common_mountcephfs_objs>)
+  $<TARGET_OBJECTS:mon_common_objs>)
+
+if(LINUX)
+  add_library(common STATIC ${libcommon_files}
```
tchaikov: why libcommon needs double doze of `module.c`?

src/CMakeLists.txt
```diff
@@ -431,6 +434,9 @@ set(libcommon_files
   common/Graylog.cc
   common/fs_types.cc
   common/dns_resolve.cc
+  common/armor.c
+  common/safe_io.c
+  common/addr_parsing.c
```
tchaikov: and mount.ceph needs `addr_parsing.c` at least, i believe?

src/CMakeLists.txt
If you look at ./do_freebsd.sh you see that I'm trying to get  those options set, but they just don't make it into the CMake. So that is why I keep trying. :)
If you have any suggestion why the commandline options do not get thru I'll fix that, and we can really remove them.

wjwithagen:
@tchaikov 
I'm moving the 3 modules up to the generic libcommon_files because they were in a `if(LINUX)`-clause, and they are generally required. Not just only on LINUX. I guess that that was an oversight from the last set of CMake changed we did for FreeBSD.

wjwithagen:
@tchaikov 
'mmm, I think I got fooled by a rebase that I merged incorrectly. Rerunning the build with the original src/CMakeLists.txt from HEAD works just fine.
other than the CXX_DEBUG stuff, but perhaps I should give them a different name, because they are also set by CMake itself.

So I'm closing this pull

src/CMakeLists.txt
tchaikov: @wjwithagen i will take a look at this.

PR #11429, opened 2016-10-11
src/common/ceph_context.h: Help Clang decide for the correct log() reference

wjwithagen:
Clang is confused about selecting the correct log() function.
Including math.h helps for the time being.
(until this is fixed in Clang, or it turns out the be very tricky stuff in
the include order)

When compiling under Clang 3.8.0 it reports:
In file included from /usr/srcs/Ceph/work/ceph/src/common/AsyncOpTracker.cc:6:
In file included from /usr/srcs/Ceph/work/ceph/src/include/Context.h:19:
In file included from /usr/srcs/Ceph/work/ceph/src/common/dout.h:20:
In file included from /usr/srcs/Ceph/work/ceph/src/common/config.h:24:
In file included from /usr/srcs/Ceph/work/ceph/src/common/entity_name.h:22:
In file included from /usr/srcs/Ceph/work/ceph/src/include/encoding.h:304:
In file included from /usr/srcs/Ceph/work/ceph/src/include/unordered_map.h:6:
In file included from /usr/include/c++/v1/unordered_map:369:
In file included from /usr/include/c++/v1/__hash_table:19:
In file included from /usr/include/c++/v1/cmath:301:
/usr/include/c++/v1/math.h:845:37: error: reference to 'log' is ambiguous
log(_A1 __lcpp_x) _NOEXCEPT {return log((double)__lcpp_x);}
                                    ^
/usr/include/c++/v1/math.h:845:1: note: candidate found by name lookup is 'log'
log(_A1 __lcpp_x) _NOEXCEPT {return log((double)__lcpp_x);}
^
/usr/include/c++/v1/math.h:839:46: note: candidate found by name lookup is 'log'
inline _LIBCPP_INLINE_VISIBILITY long double log(long double __lcpp_x) _NOEXCEPT {return logl(__lcpp_x);}
                                             ^
/usr/include/c++/v1/math.h:838:46: note: candidate found by name lookup is 'log'
inline _LIBCPP_INLINE_VISIBILITY float       log(float __lcpp_x) _NOEXCEPT       {return logf(__lcpp_x);}
                                             ^
/usr/include/math.h:247:8: note: candidate found by name lookup is 'log'
double  log(double);
        ^
/usr/srcs/Ceph/work/ceph/src/common/ceph_context.h:49:13: note: candidate found by name lookup is 'ceph::log'
  namespace log {
            ^
1 error generated.

### DIFF
--- a/src/common/ceph_context.h
+++ b/src/common/ceph_context.h
@@ -15,6 +15,11 @@
 #ifndef CEPH_CEPHCONTEXT_H
 #define CEPH_CEPHCONTEXT_H
 
+#if defined(__FreeBSD__)
+// Help Clang decide for the correct reference for log() 
+#include <math.h>
+#endif
+
 #include <iosfwd>
 #include <stdint.h>
 #include <string>


wjwithagen:
@tchaikov 
I'd rather not make a commit like this, but at the moment I see no other alternative.
Still discussing the issue on the FreeBSD-toolchain list, which ended in the suggestion to reorder the includes. But I have not found any incarnation that works.
So for the time being this "hack"

Unless you have other suggestions?

wjwithagen:
@liewegas @tchaikov 
Perhaps this skipped your attention, but next to the denc.h bool fixes, this is one of the last Mohicans to allow people to start compiling Ceph on FreeBSD.
I'm finding a few more FreeBSD hickups like duplicate getopt stuff, but that needs to be fixed by the user, and I'll put tests for that in do_freebsd.sh

cbodley:
i haven't been following this closely, but is it possibly the result of a header doing `using namespace std;`?

wjwithagen:
@cbodley 
Interesting suggestion.
I'll dump the preprocessor output, and scan the file.

wjwithagen:
@cbodley 
If I do not include the 
```
#if defined(__FreeBSD__)
// Help Clang decide for the correct reference for log() 
#include <math.h>
#endif
```
This it the first reference with 'using' and '::log'
```
# 305 "/usr/include/c++/v1/cmath" 3


namespace std {inline namespace __1 {
.....
using ::log;
using ::logf;
using ::log10;
using ::log10f;
.....
}
```
But I also find:
'''
# 62 "/usr/srcs/Ceph/work/ceph/src/include/types.h"
using namespace std;
# 1 "/usr/srcs/Ceph/work/ceph/src/include/object.h" 1
# 19 "/usr/srcs/Ceph/work/ceph/src/include/object.h"
# 1 "/usr/include/c++/v1/stdio.h" 1 3
# 20 "/usr/srcs/Ceph/work/ceph/src/include/object.h" 2
using namespace std;
'''

cbodley:
how about `using namespace ceph`?

wjwithagen:
@cbodley 
In this exact phrasing
'''
# 79 "/usr/srcs/Ceph/work/ceph/src/include/assert.h"
}
using namespace ceph;
'''
'''
# 21 "/usr/srcs/Ceph/work/ceph/src/common/Mutex.h" 2
using namespace ceph;
'''
'''
# 28 "/usr/srcs/Ceph/work/ceph/src/include/encoding.h" 2
using namespace ceph;
'''

wjwithagen:
@cbodley 

I deleted the empty lines for brevity.
But as far as I can tell tell these are on the global level. There are no extra indents so I would expect them not to be wrapped in another class. (if that was what you were wondering about)

tchaikov:
i am not able to reproduce this issue. and i found http://stackoverflow.com/questions/3871474/how-to-resolve-a-name-collision-between-a-c-namespace-and-a-global-function, which is quite the same as this problem. 

i was trying to remove the `using namespace ceph` and `using namespace std` from popular header files in hope to avoid the ambiguity. and the effort turns into a tedious work changing hundreds of source files. i will continue the adventure later on, on a boring weekend probably.

wjwithagen:
@tchaikov
Not even with Clang? 
I'd attribute it to the pedantic nature of Clang, and accept the rather hackish fix I suggested.
And once you feel like that boring weekend is up, put some music on and go for it.
OTOH changing hundreds of files is going to pollute the repo quite a bit.

tchaikov:
@wjwithagen no, not even with clang. i'd rather understanding the root cause better before merging your change set.

wjwithagen:
@tchaikov 
Sure, I understand that.
Let me know if I can help. Next 2 days I'm out of the office.

I'll think about making a patchset to be applied before running ceph:master thru FreeBSD jenkins.
That way I can always be creative, and that matches sort of the way the ports/packages do it.

wjwithagen:
@tchaikov 
From the link above it seems that this is something that happens more often.
Starting to prefix just about everything is no fun.
And what if we were to call the namespace `logging`.
From what I see there are not that many references to `log::` or `namespace log`

tchaikov:
@wjwithagen love your idea! maybe we can give it a try and see if people like it or not!

wjwithagen:
@tchaikov 
It seems to compile oke, and changing was about 15 minutes. 8=)
Now the catch could be that rocksdb does use log as well. Got to wait until the whole build is finished.

wjwithagen:
@tchaikov 
I've got a PR that seems to work in Clang for me: #11650.
Lets see what jenkins here thinks of it.

cbodley:
i don't see how clang would consider `ceph::log` to be a candidate, unless there was a `using namespace ceph` somewhere before that

wjwithagen:
@cbodley 
Like in encoding.g or assert.h??
They add the ceph namespace to the global namespace.
But I guess that is what @tchaikov is refering to, removing that will require a lot of fixes to references that would otherwise be available at the top namespace. 
Hence my suggestion to avoid the conflict by changing the namespace name.
Which is a "relativey" easy fix in #11650

PR #11430, opened 2016-10-11
FreeBSD/EventKqueue.{h,cc} Added code to restore events on (thread)fork

wjwithagen:
According the FreeBSD man page of kqueue(), the kq-descriptors become invalid
upon fork. It looks like the same happens when a kq-handle is created and then
a thread is created.

So we keep a list of assigned events with each kq-handle and when the handle
has beccome invalid, we recreated the kq-handle and the events that go with it.

### DIFF
--- a/src/msg/async/EventKqueue.cc
+++ b/src/msg/async/EventKqueue.cc
@@ -26,7 +26,6 @@
 
 int KqueueDriver::test_kqfd() {
   struct kevent ke[1];
-  // EV_SET(&ke[0], fd, EVFILT_READ, EV_ADD, 0, 0, NULL);
   if (kevent(kqfd, ke, 0, NULL, 0, NULL) == -1) {
     ldout(cct,0) << __func__ << " invalid kqfd = " << kqfd 
                  << cpp_strerror(errno) << dendl;
@@ -106,7 +105,8 @@ int KqueueDriver::add_event(int fd, int cur_mask, int add_mask)
   int oldkqfd = kqfd;
 
   if (!pthread_equal(mythread, pthread_self())) {
-    ldout(cct,10) << __func__ << " We changed thread from " << mythread << " to " << pthread_self() << dendl;
+    ldout(cct,10) << __func__ << " We changed thread from " << mythread 
+		  << " to " << pthread_self() << dendl;
     mythread = pthread_self();
     kqfd = -1;
   } else if ((kqfd != -1) && (test_kqfd() < 0)) {
@@ -130,7 +130,7 @@ int KqueueDriver::add_event(int fd, int cur_mask, int add_mask)
     }
   }
 
-  ldout(cct,10) << __func__ << " add event kqfd = " << kqfd << " fd = " << fd 
+  ldout(cct,20) << __func__ << " add event kqfd = " << kqfd << " fd = " << fd 
 	<< " cur_mask = " << cur_mask << " add_mask = " << add_mask 
 	<< dendl;
 
@@ -157,12 +157,13 @@ int KqueueDriver::del_event(int fd, int cur_mask, int del_mask)
   int num = 0;
   int mask = cur_mask & del_mask;
 
-  ldout(cct,10) << __func__ << " delete event kqfd = " << kqfd 
+  ldout(cct,20) << __func__ << " delete event kqfd = " << kqfd 
 	<< " fd = " << fd << " cur_mask = " << cur_mask 
 	<< " del_mask = " << del_mask << dendl;
 
   if (!pthread_equal(mythread, pthread_self())) {
-    ldout(cct,10) << __func__ << " We changed thread from " << mythread << " to " << pthread_self() << dendl;
+    ldout(cct,10) << __func__ << " We changed thread from " << mythread 
+		  << " to " << pthread_self() << dendl;
     mythread = pthread_self();
   } else if ((kqfd != -1) && (test_kqfd() < 0)) {
     ldout(cct,10) << __func__ << " Recreating old kqfd."  << dendl;
@@ -213,7 +214,8 @@ int KqueueDriver::event_wait(vector<FiredFileEvent> &fired_events, struct timeva
 
   ldout(cct,10) << __func__ << " kqfd = " << kqfd << dendl;
   if (!pthread_equal(mythread, pthread_self())) {
-    ldout(cct,10) << __func__ << " We changed thread from " << mythread << " to " << pthread_self() << dendl;
+    ldout(cct,10) << __func__ << " We changed thread from " << mythread 
+		  << " to " << pthread_self() << dendl;
     mythread = pthread_self();
     kqfd = -1;
   } else if ((kqfd != -1) && (test_kqfd() < 0)) {
@@ -238,19 +240,19 @@ int KqueueDriver::event_wait(vector<FiredFileEvent> &fired_events, struct timeva
   if (tvp != NULL) {
       timeout.tv_sec = tvp->tv_sec;
       timeout.tv_nsec = tvp->tv_usec * 1000;
-      ldout(cct,10) << __func__ << " "
+      ldout(cct,20) << __func__ << " "
 		<< timeout.tv_sec << " sec "
 		<< timeout.tv_nsec << " nsec"
 		<< dendl;
       retval = kevent(kqfd, NULL, 0, res_events, size, &timeout);
   } else {
-      ldout(cct,10) << __func__ << " event_wait: "
+      ldout(cct,20) << __func__ << " event_wait: "
 		<< " NULL"
 		<< dendl;
       retval = kevent(kqfd, NULL, 0, res_events, size, NULL);
   }
 
-  ldout(cct,10) << __func__ << " kevent retval: "
+  ldout(cct,25) << __func__ << " kevent retval: "
 		<< retval
 		<< dendl;
   if (retval < 0) {
@@ -270,7 +272,7 @@ int KqueueDriver::event_wait(vector<FiredFileEvent> &fired_events, struct timeva
     for (j = 0; j < numevents; j++) {
       int mask = 0;
       struct kevent *e = res_events + j;
-      ldout(cct,10) << __func__ << " j: " << j << dendl;
+      ldout(cct,20) << __func__ << " j: " << j << dendl;
 
       if (e->filter == EVFILT_READ) mask |= EVENT_READABLE;
       if (e->filter == EVFILT_WRITE) mask |= EVENT_WRITABLE;
--- a/src/msg/async/EventKqueue.cc
+++ b/src/msg/async/EventKqueue.cc
@@ -26,7 +26,7 @@
 
 int KqueueDriver::test_kqfd() {
   struct kevent ke[1];
-  if (kevent(kqfd, ke, 0, NULL, 0, NULL) == -1) {
+  if (kevent(kqfd, ke, 0, NULL, 0, KEVENT_NOWAIT) == -1) {
     ldout(cct,0) << __func__ << " invalid kqfd = " << kqfd 
                  << cpp_strerror(errno) << dendl;
     return -errno;
@@ -37,21 +37,20 @@ int KqueueDriver::test_kqfd() {
 int KqueueDriver::restore_events() {
   struct kevent ke[2];
   int i;
-  int num = 0;
 
   ldout(cct,10) << __func__ << " on kqfd = " << kqfd << dendl;
   for(i=0;i<size;i++) {
+    int num = 0;
     if (sav_events[i].mask == 0 )
       continue;
     ldout(cct,10) << __func__ << " restore kqfd = " << kqfd 
-	<< " fd = " << i << " mask " << sav_events[i].mask
-	<< dendl;
+                  << " fd = " << i << " mask " << sav_events[i].mask << dendl;
     if (sav_events[i].mask & EVENT_READABLE)
       EV_SET(&ke[num++], i, EVFILT_READ, EV_ADD, 0, 0, NULL);
     if (sav_events[i].mask & EVENT_WRITABLE)
       EV_SET(&ke[num++], i, EVFILT_WRITE, EV_ADD, 0, 0, NULL);
     if (num) {
-      if (kevent(kqfd, ke, num, NULL, 0, NULL) == -1) {
+      if (kevent(kqfd, ke, num, NULL, 0, KEVENT_NOWAIT) == -1) {
         ldout(cct,0) << __func__ << " unable to add event: "
                      << cpp_strerror(errno) << dendl;
         return -errno;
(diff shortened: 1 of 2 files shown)


wjwithagen:
@yuyuyu101 
I've taken that rather rigorous approach to redo the events on a kqfd after the thread has forked.
But it seems to work, since all tests that are available run to completion. (This is for the first time.)

So would you care to review the code, and comment....
More that happy to fix any suggestions.

liewegas:
@yuyuyu101

wjwithagen:
@liewegas 
I pinged @yuyuyu101 already to see what he thinks of the code.
Probably is not going to win a first prize in a beauty contest, but it does get the thing done.

src/msg/async/EventKqueue.h
```diff
@@ -25,18 +25,33 @@
 
 class KqueueDriver : public EventDriver {
   int kqfd;
-  struct kevent *events;
+  pthread_t mythread;
+  struct kevent *res_events;
   CephContext *cct;
   int size;
 
+ private:
```
yuyuyu101: nit: don't need to private here

src/msg/async/EventKqueue.cc
```diff
@@ -62,17 +146,43 @@ int KqueueDriver::add_event(int fd, int cur_mask, int add_mask)
...
+      lderr(cct) << __func__ << " unable to do kqueue: "
+                             << cpp_strerror(errno) << dendl;
+      return -errno;
+    }
+    if (restore_events()< 0) {
+      lderr(cct) << __func__ << " unable restore all events "
+                             << cpp_strerror(errno) << dendl;
+      return -errno;
+    }
+  }
```
yuyuyu101: I think we can move these checks to a function to make it shared

src/msg/async/EventKqueue.cc
```diff
@@ -99,22 +211,66 @@ int KqueueDriver::event_wait(vector<FiredFileEvent> &fired_events, struct timeva
   int retval, numevents = 0;
   struct timespec timeout;
 
+  ldout(cct,10) << __func__ << " kqfd = " << kqfd << dendl;
+  if (!pthread_equal(mythread, pthread_self())) {
+    ldout(cct,10) << __func__ << " We changed thread from " << mythread << " to " << pthread_self() << dendl;
+    mythread = pthread_self();
+    kqfd = -1;
+  } else if ((kqfd != -1) && (test_kqfd() < 0)) {
```
yuyuyu101: if thread id isn't changed, it may destroy kqfd?

src/msg/async/EventKqueue.h
```diff
@@ -25,18 +25,33 @@
...
+  // Keep what we set on the kqfd
+  struct SaveEvent{
+    int fd;
+    int mask;
+  };
```
yuyuyu101: I think we can move struct kevent to this structure.

src/msg/async/EventKqueue.cc
```diff
@@ -22,34 +22,118 @@
...
+      return -errno;
+    }
+    if (restore_events()< 0) {
+      lderr(cct) << __func__ << " unable restore all events "
+                             << cpp_strerror(errno) << dendl;
+      return -errno;
+    }
+  }
+
+  ldout(cct,10) << __func__ << " add event kqfd = " << kqfd << " fd = " << fd
```
yuyuyu101: 20 should be better. 10 is too verbose

src/msg/async/EventKqueue.cc
```diff
@@ -62,17 +146,43 @@ int KqueueDriver::add_event(int fd, int cur_mask, int add_mask)
...
+int KqueueDriver::del_event(int fd, int cur_mask, int del_mask)
 {
-  ldout(cct, 20) << __func__ << " del event fd=" << fd << " cur mask=" << cur_mask
-                 << " delmask=" << delmask << dendl;
   struct kevent ke[2];
   int num = 0;
-  int mask = cur_mask & delmask;
+  int mask = cur_mask & del_mask;
+
+  ldout(cct,10) << __func__ << " delete event kqfd = " << kqfd
```
yuyuyu101: 20

src/msg/async/EventKqueue.cc
```diff
@@ -99,22 +211,66 @@ int KqueueDriver::event_wait(vector<FiredFileEvent> &fired_events, struct timeva
   int retval, numevents = 0;
   struct timespec timeout;
 
+  ldout(cct,10) << __func__ << " kqfd = " << kqfd << dendl;
```
yuyuyu101: I hope we don't print this or improve this to 30

src/msg/async/EventKqueue.cc
```diff
@@ -99,22 +211,66 @@ int KqueueDriver::event_wait(vector<FiredFileEvent> &fired_events, struct timeva
...
   } else {
-      retval = kevent(kqfd, NULL, 0, events, size, NULL);
+      ldout(cct,10) << __func__ << " event_wait: "
+		<< " NULL"
+		<< dendl;
+      retval = kevent(kqfd, NULL, 0, res_events, size, NULL);
   }
 
-  if (retval > 0) {
+  ldout(cct,10) << __func__ << " kevent retval: "
```
yuyuyu101: 20 or 30

src/msg/async/EventKqueue.cc
```diff
@@ -99,22 +211,66 @@ int KqueueDriver::event_wait(vector<FiredFileEvent> &fired_events, struct timeva
...
+  } else {
     int j;
 
     numevents = retval;
     fired_events.resize(numevents);
     for (j = 0; j < numevents; j++) {
       int mask = 0;
-      struct kevent *e = events + j;
+      struct kevent *e = res_events + j;
+      ldout(cct,10) << __func__ << " j: " << j << dendl;
```
yuyuyu101: hope we don't need this

wjwithagen:
@yuyuyu101 
Glad you think it is not a total mess.... I will work on the changes.

And yes, I would love to do this at a higher level, but on the other hand it would perhaps obfuscate the code at that level. This is obviously a Kqueue issue, and as such should be hidden in the kqueue-code.
Perhaps for me the most important thing: I'm not 100% confident that I understand all details of the thread creation in the stack module.
Perhaps we should later on, when the dust has settled, and we are quite sure that this is a workable solution that has passed QA.

wjwithagen:
@yuyuyu101 
If you want to go the handle_fork way, you will have to give me some pointers/example code of what your idea is. And I'll start thinking about it, and we can see if it delivers cleaner code.

yuyuyu101:
@wjwithagen hmm, I think we can go through this way at first. Let's see it will work firstly. Then if available, I could just refactor this workable codes and make it in higher level. So let's make it first!

Thanks!!

src/msg/async/EventKqueue.h
wjwithagen: @yuyuyu101 
That would actually make it hard to use `res_events[]` the structure that `kevent()` can return the triggerd events in. So we'd again need to `alloc(nevents-size)` an array for that, or do that on the stack. Blowing the stack out of proportion on every call to `event_wait()`

src/msg/async/EventKqueue.cc
wjwithagen: @yuyuyu101 
Will fix all verbose levels, since they are too low.

src/msg/async/EventKqueue.cc
wjwithagen: @yuyuyu101 
I have seen cases where it triggered an "invalid" kqfd. But perhaps that was during testing and the code still had bugs. Lets keep this code in for now. And in due time if we full understand what is going on, perhaps change it into an 'assert()' if this actually should not happen. But now I'm not 100% certain.

wjwithagen:
@yuyuyu101 
Great, thanx for the positive feedback.

src/msg/async/EventKqueue.cc
wjwithagen: @yuyuyu101 
Or even just return a negative result. The layer above is not equipped to handle kqueue-event faults, and assert in most cases anyways.

wjwithagen:
@yuyuyu101 
I think I addressed most/all of your comments?
If you are oke with this, I'll rebase it again into 2 commits.

src/msg/async/EventKqueue.cc
```diff
@@ -99,22 +205,48 @@ int KqueueDriver::event_wait(vector<FiredFileEvent> &fired_events, struct timeva
...
+  } else {
     int j;
 
     numevents = retval;
     fired_events.resize(numevents);
     for (j = 0; j < numevents; j++) {
       int mask = 0;
-      struct kevent *e = events + j;
+      struct kevent *e = res_events + j;
+      ldout(cct,20) << __func__ << " j: " << j << dendl;
```
yuyuyu101: I dont know why need to print j.... since we already know the number of events returned

src/msg/async/EventKqueue.cc
yuyuyu101: this?

src/msg/async/EventKqueue.cc
wjwithagen: @yuyuyu101 
This is really debug stuff to make sure all events are processed.
So yes it can go.

wjwithagen:
@yuyuyu101 
Ping...

src/msg/async/EventKqueue.cc
```diff
@@ -22,34 +22,129 @@
 #undef dout_prefix
 #define dout_prefix *_dout << "KqueueDriver."
 
+#define KEVENT_NOWAIT 0
```
yuyuyu101: unused?

src/msg/async/EventKqueue.cc
```diff
@@ -22,34 +22,129 @@
...
+    if (sav_events[i].mask & EVENT_READABLE)
+      EV_SET(&ke[num++], i, EVFILT_READ, EV_ADD, 0, 0, NULL);
+    if (sav_events[i].mask & EVENT_WRITABLE)
+      EV_SET(&ke[num++], i, EVFILT_WRITE, EV_ADD, 0, 0, NULL);
+    if (num) {
+      if (kevent(kqfd, ke, num, NULL, 0, NULL) == -1) {
+        ldout(cct,0) << __func__ << " unable to add event: "
+                     << cpp_strerror(errno) << dendl;
+        return -errno;
+      }
```
yuyuyu101: I think you need to reset num = 0 here? otherwise, num will continue to grow.....

src/msg/async/EventKqueue.cc
yuyuyu101: ping

src/msg/async/EventKqueue.h
```diff
@@ -25,18 +25,33 @@
...
+  struct SaveEvent *sav_events;
+  int sav_index;
+  int restore_events();
+  int test_kqfd();
+  int test_thread_change(const char* funcname);
+
  public:
-  explicit KqueueDriver(CephContext *c): kqfd(-1), events(NULL), cct(c), size(0) {}
+  explicit KqueueDriver(CephContext *c): kqfd(-1), res_events(NULL), cct(c), size(0) {}
```
yuyuyu101: need to init sav_events = 0

src/msg/async/EventKqueue.h
```diff
@@ -25,18 +25,33 @@
...
   CephContext *cct;
   int size;
 
+  // Keep what we set on the kqfd
+  struct SaveEvent{
+    int fd;
+    int mask;
+  };
+  struct SaveEvent *sav_events;
+  int sav_index;
```
yuyuyu101: unused

src/msg/async/EventKqueue.h
```diff
@@ -25,18 +25,33 @@
...
+  struct kevent *res_events;
   CephContext *cct;
   int size;
 
+  // Keep what we set on the kqfd
+  struct SaveEvent{
+    int fd;
+    int mask;
+  };
+  struct SaveEvent *sav_events;
```
yuyuyu101: because you save all ongoing kqueue events, I think you need to resize in KqueueDriver::resize_events in case of fd is exceed the init number

src/msg/async/EventKqueue.cc
wjwithagen: @yuyuyu101 
Leftover from previous code. Usually I replace hard values with descriptive constants.
I'll see if it still fits the purpose.

src/msg/async/EventKqueue.h
wjwithagen: Right the idea was to keep save usage here.
Best to name it sav_max and keep the size of sav_events for resizing.

src/msg/async/EventKqueue.h
wjwithagen: @yuyuyu101 
Right, I was hoping not to run into that :), but it does require action.
Would be a lot when 1 OSD has like 5000 fd's open.

wjwithagen:
@yuyuyu101 
Good review, I'm currently also busy with other things, so it might take more than a few hours to rework this.

wjwithagen:
@yuyuyu101 
I think I got all the issues you touched and did a bit of cleaning up.
Especially catching the `num=0` was a nice one, could be the reason that I sometimes saw restore_event errors, and reset the FD. Probably events were reinserted that were already there.
So can you do another review?

src/msg/async/EventKqueue.cc
```diff
@@ -22,75 +22,193 @@
...
+    sav_events = (struct SaveEvent*)realloc( sav_events, 
+										sizeof(struct SaveEvent)*newsize);
+    if (!sav_events) {
+      lderr(cct) << __func__ << " unable to realloc memory: "
+                             << cpp_strerror(errno) << dendl;
+      return -ENOMEM;
+    }
+	memset(&sav_events[size], 0, sizeof(struct SaveEvent)*(newsize-sav_max));
+	sav_max = newsize;
+  }
```
yuyuyu101: looks format is totally unaligned

src/msg/async/EventKqueue.cc
yuyuyu101: @wjwithagen this line... we don't want to print level 10 log when entering event_wait each time

yuyuyu101:
others look better to me

wjwithagen:
@yuyuyu101 
You are right. Looks like vi did not pick up on tabbing.
Fixed that.

yuyuyu101:
Because the qa won't compile out Kqueue driver, I don't think we need qa process. But before merging, I want to make sure @wjwithagen have tested the fresh commits. And got ack from you....

wjwithagen:
@yuyuyu101 
Before I left yesterday, i fired up a testrun with this code in my work tree:
```
99% tests passed, 1 tests failed out of 129
Total Test time (real) = 3531.31 sec
The following tests FAILED:
         12 - run-tox-ceph-disk (Failed)
```

And the failure on the last on is expected, because I have not done any work on it.
So I think it is good to go.

wjwithagen:
@yuyuyu101 
Rebased to compress the commits into 3 pieces.

wjwithagen:
@yuyuyu101 
thank ....

PR #11431, opened 2016-10-11
jewel: build/ops: include more files in "make dist" tarball

ktdreyer:
http://tracker.ceph.com/issues/17560

Include all the documentation, qa, and CMakeLists.txt files in the `make dist` tarball.

Kraken and newer releases will already do this. This change is restricted to releases that still use autotools (jewel in this case).

The purpose of this change is to make it easier to apply downstream patches on an unpacked upstream tarball, because often those patches are cherry-picks of commits that touch files under `doc/`, `qa/`, or `CMakeLists.txt`.


dachary:
I'll create the matching Backport issue in the tracker as soon as it comes back online.

smithfarm:
No objections.

dachary:
It passed the upgrade/jewel-x and upgrade/hammer-x (http://tracker.ceph.com/issues/17487#note-22) suites.

PR #11432, opened 2016-10-11
doc: Update install-ceph-gateway.rst

hansbogert:
Make the s3test.py script PIP-8 compliant, and more importantly it no longer fails because of the wrong indentation on the line of the for-loop

### DIFF
--- a/doc/install/install-ceph-gateway.rst
+++ b/doc/install/install-ceph-gateway.rst
@@ -510,24 +510,24 @@ Execute the following steps:
 
 #. Add the following contents to the file::
 
-    import boto
     import boto.s3.connection
 
     access_key = 'I0PJDPCIYZ665MW88W9R'
     secret_key = 'dxaXZ8U90SXydYzyS5ivamEP20hkLSUViiaR+ZDA'
     conn = boto.connect_s3(
-            aws_access_key_id = access_key,
-            aws_secret_access_key = secret_key,
-            host = '{hostname}', port = {port},
-            is_secure=False, calling_format = boto.s3.connection.OrdinaryCallingFormat(),
-            )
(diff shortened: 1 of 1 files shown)


doc/install/install-ceph-gateway.rst
```diff
@@ -510,24 +510,24 @@ Execute the following steps:
 
 #. Add the following contents to the file::
 
-    import boto
```
hansbogert: I think it's superfluous. I checked it btw, and it works. I used pycharm which says it's an unused import.

tchaikov: @hansbogert. okay, i trust you. but how come python knows what `boto` is without this statement. weird...

hansbogert: Couldn't really stand it as well so I looked it up. `connect_s3` is defined in the `__init__.py` of top level of the boto package, which gets read even when you import a subpackage like `boto.s3.connection`

hansbogert: Btw, do **I** need to sign when I commit something like this? I am not familiar with the linux kernel's process of signage, which you guys seem to be using.

hansbogert:
There is one error with the automated tests. Unlikely this is of my doing, but should I rebase?

tchaikov:
@hansbogert the last thing,  could you prefix the title of your commit message with the subcomponent your are changing ? see https://github.com/ceph/ceph/blob/master/SubmittingPatches.rst#preparing-and-sending-patches, in your case, it would be "doc: ".

hansbogert:
Well, this trivial patch served as good pilot for contributing to Ceph. Hope all is well now.

PR #11436, opened 2016-10-12
kv/MemDB: making memdb code adapt to generic maps

chhabaramesh:
Changing code to use generic map for MemDB implementation.

### DIFF
--- a/src/kv/MemDB.cc
+++ b/src/kv/MemDB.cc
@@ -50,8 +50,7 @@ static string make_key(const string &prefix, const string &value)
   return out;
 }
 
-void MemDB::_encode(btree::btree_map<string,
-                      bufferptr>:: iterator iter, bufferlist &bl)
+void MemDB::_encode(mdb_iter_t iter, bufferlist &bl)
 {
   ::encode(iter->first, bl);
   ::encode(iter->second, bl);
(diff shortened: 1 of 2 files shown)


chhabaramesh:
@liewegas , please review and merge this, code changes are to make types generic to try different maps

yuyuyu101:
why change to std::map, performance reason?

chhabaramesh:
@yuyuyu101 , yes, seems btree map is not scaling much and is compute intensive.  Std::map also does not help much with just this change. But change is more of making it use any map by single line change.

liewegas:
This also lets us avoid the annoying iterator invalidation logic.  Can you add a flag for that too?

chhabaramesh:
@liewegas , will do it today, apologies for the delay in responding.

chhabaramesh:
@liewegas , Iterartor invalidation works this way now:

. If seq does not change, it is valid
. if using btree and seq no has changed, restart iter.
. If not using btree and seq no has changed, but current key still exists, iter is valid.
. If not using btree and seq not has changed and current key does not exists, restart the iter.

I am still keeping seqno logic for all maps since it should be faster then checking "if key exists" every time.

liewegas:
Looks good, thanks!

PR #11438, opened 2016-10-12
rbd-nbd: support disabling auto-exclusive lock transition logic

trociny:
Fixes: http://tracker.ceph.com/issues/17488

### DIFF
--- a/src/tools/rbd_nbd/rbd-nbd.cc
+++ b/src/tools/rbd_nbd/rbd-nbd.cc
@@ -64,6 +64,7 @@ static void usage()
             << "  --device <device path>                    Specify nbd device path\n"
             << "  --read-only                               Map readonly\n"
             << "  --nbds_max <limit>                        Override for module param\n"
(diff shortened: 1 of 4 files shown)


src/tools/rbd_nbd/rbd-nbd.cc
```diff
@@ -64,13 +64,15 @@ static void usage()
             << "  --device <device path>                    Specify nbd device path\n"
             << "  --read-only                               Map readonly\n"
             << "  --nbds_max <limit>                        Override for module param\n"
+            << "  --single-client                           Forbid other clients write\n"
```
dillaman: Minor: how about just "--exclusive" or "--exclusive-client"? Also, can you add the argument to the "rbd nbd map" CLI as well?

dillaman:
... also need to update the rbd CLI help.t cram test case

trociny:
@dillaman Thanks, updated.

dillaman:
lgtm

PR #11443, opened 2016-10-12
do_freebsd.sh: Set options for debug building.

wjwithagen:
And that requires that the options need to be passed on completely
to the cmake program as well. Thus adding """'s around the option
argument with do_cmake.sh as well

### DIFF
--- a/do_cmake.sh
+++ b/do_cmake.sh
@@ -6,7 +6,7 @@ if test -e build; then
 fi
 mkdir build
 cd build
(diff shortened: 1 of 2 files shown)


wjwithagen:
@tchaikov 
Could you have whack at this one? 
It fixes my challenge to get my options for debugging into Cmake.

wjwithagen:
@tchaikov 
Thanx...

PR #11446, opened 2016-10-12
tools/cephfs: add scan_links command which fixes linkages errors

ukernel:

### DIFF
--- a/src/tools/cephfs/DataScan.cc
+++ b/src/tools/cephfs/DataScan.cc
@@ -35,6 +35,7 @@ void DataScan::usage()
     << "  cephfs-data-scan scan_extents [--force-pool] <data pool name>\n"
     << "  cephfs-data-scan scan_inodes [--force-pool] [--force-corrupt] <data pool name>\n"
     << "  cephfs-data-scan pg_files <path> <pg id> [<pg id>...]\n"
+    << "  cephfs-data-scan scan_links\n"
     << "\n"
     << "    --force-corrupt: overrite apparently corrupt structures\n"
     << "    --force-init: write root inodes even if they exist\n"
@@ -233,8 +234,8 @@ int DataScan::main(const std::vector<const char*> &args)
   }
 
   // Initialize data_io for those commands that need it
-  if (command == "scan_inodes"
-     || command == "scan_extents") {
+  if (command == "scan_inodes" ||
+      command == "scan_extents") {
     if (data_pool_name.empty()) {
       std::cerr << "Data pool not specified" << std::endl;
       usage();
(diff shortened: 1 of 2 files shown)


src/tools/cephfs/DataScan.cc
```diff
@@ -821,6 +824,223 @@ bool DataScan::valid_ino(inodeno_t ino) const
...
+      if (r == -EINVAL) {
+	dout(10) << "Not a dirfrag: '" << oid << "'" << dendl;
+	continue;
+      } else {
+	// parse_oid can only do 0 or -EINVAL
+	assert(r == 0);
+      }
+
+      if (!valid_ino(dir_ino)) {
+	dout(10) << "Not a difrag (invalid ino): '" << oid << "'" << dendl;
```
gregsfortytwo: s/difrag/dirfrag

ukernel:
Two passes design is for reduce memory usage. Scan metadata pool is fast in most case.

This tool is for repair linkage errors, such as incorrect inode link count, duplicated primary dentries.  The online scrub can't fix this type of error easily.

ukernel:
I don't know what 'scan a single directory and compare against the stray dir' can do?  find bad remote link ? force stray reintegration? both can be done easily by scrub

gregsfortytwo:
You were suggesting this on the mailing list as a solution for reintegrating strays into their new locations, and I'm not sure this is a very efficient way of doing that.

The other cases aren't very likely but are definitely worth being able to repair. Please write some tests demonstrating this works.

gregsfortytwo:
Also I'm really not thrilled about the in-memory map of all inodes; that's going to out some serious limits on scalability. But it will do for getting on with (although we should also add a fracker ticket about it).

ukernel:
On Wed, Oct 19, 2016 at 1:05 PM, Gregory Farnum <notifications@github.com>
wrote:

> You were suggesting this on the mailing list as a solution for
> reintegrating strays into their new locations, and I'm not sure this is a
> very efficient way of doing that.
>

No, I didn't suggest the user to use this tool for reintegrating strays.
(The tool does not have this function either). The user's cephfs was
corrupted (there are duplicated primary dentries). I write this tool for
repairing the corruption

gregsfortytwo:
@ukernel, I will merge this when you add tests. I don't see any. :)

ukernel:
https://github.com/ceph/ceph-qa-suite/pull/1219

jcsp:
Test failure: http://qa-proxy.ceph.com/teuthology/jspray-2016-10-24_17:11:20-fs-wip-jcsp-testing-20161024-distro-basic-mira/493088/teuthology.log

ukernel:
Test case fixed

PR #11448, opened 2016-10-12
Rewrite mount.fuse.ceph (to python) and move ceph-fuse options to fs_mntops

ledgr:
Rewrote mount.fuse.ceph to move ceph-fuse options to `fs_mntops`, where it should be.
Bash version with options in `fs_spec` is counterintuitive and cause issues in some
situations (systemd, for example)

### DIFF
--- a/doc/cephfs/fstab.rst
+++ b/doc/cephfs/fstab.rst
@@ -29,15 +29,14 @@ FUSE
 To mount Ceph FS in your file systems table as a filesystem in user space, add the
 following to ``/etc/fstab``::
 
-	#DEVICE                                  PATH         TYPE      OPTIONS
-	id={user-ID}[,conf={path/to/conf.conf}] /mount/path  fuse.ceph defaults,_netdev 0 0
+       #DEVICE PATH       TYPE      OPTIONS
+       none    /mnt/ceph  fuse.ceph ceph.id={user-ID}[,ceph.conf={path/to/conf.conf}],_netdev,defaults  0 0
 
 For example::
 
-	id=admin  /mnt/ceph  fuse.ceph defaults 0 0 
-	id=myuser,conf=/etc/ceph/cluster.conf  /mnt/ceph2  fuse.ceph defaults 0 0 
+       none    /mnt/ceph  fuse.ceph ceph.id=myuser,_netdev,defaults  0 0
+       none    /mnt/ceph  fuse.ceph ceph.id=myuser,ceph.conf=/etc/ceph/foo.conf,_netdev,defaults  0 0
 
-The ``DEVICE`` field is a comma-delimited list of options to pass to the command line.
 Ensure you use the ID (e.g., ``admin``, not ``client.admin``). You can pass any valid 
 ``ceph-fuse`` option to the command line this way.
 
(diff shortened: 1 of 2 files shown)


jcsp:
I don't think any of our automated testing covers this.  How have you tested it?

ledgr:
We have used this script for half a year for our project. Then switched to kernel client.
It works with standard fuse and systemd .mount and .automount units. We didn't do any sophisticated tests.

ledgr:
Systemd `unit` file used:

```
[Unit]
Description=Mount /home as ceph-fuse client
DefaultDependencies=yes
After=remote-fs-pre.target network.target network-online.target

[Install]
WantedBy=multi-user.target

[Mount]
What=none
Type=fuse.ceph
Options=ceph.name=client.cephfuse-client.domain.tld,nonempty,_netdev
Where=/home
```

_netdev is necessary: https://www.freedesktop.org/software/systemd/man/systemd.mount.html#Automatic%20Dependencies

>  In some cases this is not sufficient (for example network block device based mounts, such as iSCSI), in which case _netdev may be added to the mount option string of the unit, which forces systemd to consider the mount unit a network mount.

Systemd `automount` unit used:
```
[Unit]
Description=/home ceph-fuse Automount Point
DefaultDependencies=yes
After=remote-fs-pre.target network.target network-online.target

[Automount]
Where=/home
```

jcsp:
OK, I think I'm on board with this, but it seems like it wouldn't be hard to make it backward compatible for existing fstabs (i.e. detect if the first column is not none, and parse out the old-style args in this case).

An update to doc/cephfs/fstab.rst is also needed.

jcsp:
@ledgr could you add the backward compatibility change and docs update so that we can merge it?  I appreciate this works already in your environment but if we merged it as-is it would break some people's systems and confuse anyone reading the docs.

jcsp:
(note to self, don't merge until we branch Kraken)

jcsp:
@ledgr are you still interested in coming back to this?

ledgr:
@jspc, yes, I will come back to it later this week.

jspc:
@ledgr - I think you want @jcsp 😁

ledgr:
Haha, yes, a typo 😁

ledgr:
@jcsp

ledgr:
Updated for backward compatibility with existing fstabs.
Updated docs.

Tested python versions:
```
Python 2.6.6
Python 2.7.9
Python 3.4.5
```

Tested mount versions:
```
mount from util-linux 2.25.2 (libmount 2.25.0: selinux, assert, debug)
mount from util-linux 2.20.1 (with libblkid and selinux support)
mount from util-linux 2.28.2 (libmount 2.28.2: selinux, btrfs, force-mountinfo, assert, debug)
mount from util-linux 2.23.2 (libmount 2.23.0: selinux, debug, assert)
```
CentOS 7.2 had problems with mount: https://bugzilla.redhat.com/show_bug.cgi?id=1248003
Which is fixed by updating `libmount to 2.23.2-33.el7`

@jcsp 
> I don't think any of our automated testing covers this.

I didn't write any tests for this, nor did I test it against multiple `ceph-fuse` versions

jcsp:
:+1: thank you @ledgr

PR #11449, opened 2016-10-12
test/osd-scrub-repair.sh: Use test case specific object names to help…

dzafman:


dzafman:
@tchaikov This is changing a unit test which has been deactivated.  I have other branches I want to merge on top of this, so I'm putting it in first.  Jenkins not able to run, but it wouldn't be running this test anyway.

PR #11450, opened 2016-10-12
mds/DamageTable: move classes to .cc file

stiopaa1:


gregsfortytwo:
Sure, looks good. I see this also adds the anonymous namespacing.

stiopaa1:
@gregsfortytwo 
Even if we don't want don't want to modify existing code to introduce anonymous namespaces I guess there is no harm adding them if I'm changing to code in this area anyway?
Some time ago with very similar commit I was even asked by @tchaikov to add anonymous namespace.

PR #11451, opened 2016-10-12
os/bluestore: fix writes that span existing shard boundaries

liewegas:
For now I take the simple strategy and force a reshard if we happen to write across a boundary.  I susepct the optimal solution is a bit more nuanced.  For example, big writes could be split, but small writes probably shouldn't be.  Or, better yet, this sort of write could just move the shard boundary over, or otherwise trigger a partial reshard.  I'm happy leaving that for future work, though!

### DIFF
--- a/src/os/bluestore/BlueStore.cc
+++ b/src/os/bluestore/BlueStore.cc
@@ -1569,9 +1569,6 @@ void BlueStore::ExtentMap::reshard(Onode *o, uint64_t min_alloc_size)
 	   << " target " << target << " slop " << slop << dendl;
 
   // reshard
-  auto sp = o->onode.extent_map_shards.begin();
-  auto esp = o->onode.extent_map_shards.end();
-  unsigned shard_end = 0;
   unsigned estimate = 0;
   unsigned offset = 0;
   vector<bluestore_onode_t::shard_info> new_shard_info;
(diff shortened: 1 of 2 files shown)


src/os/bluestore/BlueStore.cc
```diff
@@ -1675,6 +1676,7 @@ void BlueStore::ExtentMap::reshard(Onode *o, uint64_t min_alloc_size)
       }
     }
   }
+  needs_reshard = false;
```
ifed01: Put that at function entrance to cover returns from the mid of the func?

src/os/bluestore/BlueStore.h
```diff
@@ -627,6 +629,21 @@ class BlueStore : public ObjectStore,
       return -1; // not found
     }
 
+    /// check if a range spans a shard
+    bool spans_shard(uint32_t offset, uint32_t length) {
+      if (shards.empty()) {
+	return false;
+      }
+      size_t s = seek_shard(offset);
```
ifed01: seek_shard returns int

ifed01: handling for not found shard? Or assert(s>=0) if this is unexpected.

src/os/bluestore/BlueStore.cc
```diff
@@ -2088,6 +2090,9 @@ BlueStore::Extent *BlueStore::ExtentMap::set_lextent(
   b->ref_map.get(offset, length);
   Extent *le = new Extent(logical_offset, offset, length, blob_depth, b);
   extent_map.insert(*le);
+  if (spans_shard(offset, length)) {
```
ifed01: replace with if( !needs_reshard && spans_shard....)?

ifed01:
LGTM except mentioned nits

PR #11452, opened 2016-10-12
os/bluestore: fix onode vs extent key suffix

liewegas:
I set the extent key suffix as 'x', thinking that was
not a valid hex character and would let us quickly
identify extent keys (vs onode keys, which end with an
encoded u64). But that doesn't work: the keys encode
integer values in raw form--not in hex
(pretty_binary_string just prints it out that way in
the debug log).

Fix by appending 'o' to the onode keys, so that we will
always have a trailing 'o' or 'x' and can use the last
char reliably to determine the type.


xiexingguo:
lgtm

PR #11453, opened 2016-10-13
Don't loop forever when reading data from 0 sized semgment.

mdw-at-linuxbox:

This version of the patch: no assert, handles DLO with only empty segments.

### DIFF
--- a/src/rgw/rgw_op.cc
+++ b/src/rgw/rgw_op.cc
@@ -715,6 +715,10 @@ int RGWGetObj::read_user_manifest_part(rgw_bucket& bucket, RGWObjEnt& ent, RGWAc
     return -EPERM;
   }
 
+  if (ent.size == 0) {
+    return 0;
+  }
+
   perfcounter->inc(l_rgw_get_b, cur_end - cur_ofs);
   while (cur_ofs <= cur_end) {
     bufferlist bl;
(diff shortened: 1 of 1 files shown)


src/rgw/rgw_op.cc
```diff
@@ -715,6 +715,14 @@ int RGWGetObj::read_user_manifest_part(rgw_bucket& bucket, RGWObjEnt& ent, RGWAc
     return -EPERM;
   }
 
+  if (ent.size == 0) {
+    /* This is a 0 sized segment, which can be easily created in a DLO.
+     * Return now; we're done.
+     * (BZ #1380196)
```
ktdreyer: Can we please refer to a tracker.ceph.com ticket here, instead of Red Hat's Bugzilla?

src/rgw/rgw_op.cc
```diff
@@ -872,6 +886,15 @@ int RGWGetObj::handle_user_manifest(const char *prefix)
   if (r < 0)
     return r;
 
+  if (!total_len) {
+    /* There were no non-zero segments in the DLO.
+     * So we still need to return status.
+     * (BZ #1380196)
```
ktdreyer: Can we please refer to a tracker.ceph.com ticket here, instead of Red Hat's Bugzilla?

ktdreyer:
Is there a tracker.ceph.com ticket for this?

I want to be sure this is fixed in master and jewel, too.

src/rgw/rgw_op.cc
```diff
@@ -715,6 +715,14 @@ int RGWGetObj::read_user_manifest_part(rgw_bucket& bucket, RGWObjEnt& ent, RGWAc
     return -EPERM;
   }
 
+  if (ent.size == 0) {
```
yehudasa: "wיןבי can be easily created in DLO" is completely uninteresting as a comment (and can be confusing)

yehudasa:
@mdw-at-linuxbox see comments, otherwise lgtm

ktdreyer:
Thanks Marcus for filing http://tracker.ceph.com/issues/17572 for this PR.

mdw-at-linuxbox:
I found very few comments pointing back to tracker (and none to BZ) so I just took that out.  I try to put comments in to clarify confusing points, but I also admire terseness.

The last push I did was purely cosmetic: should not affect any testing results.

ktdreyer:
Closing in favor of https://github.com/ceph/ceph/pull/11628

PR #11455, opened 2016-10-13
mds/Mutation.h: simplify constructors

stiopaa1:

### DIFF
--- a/src/mds/Mutation.h
+++ b/src/mds/Mutation.h
@@ -38,8 +38,8 @@ class MMDSSlaveRequest;
 
 struct MutationImpl {
   metareqid_t reqid;
-  __u32 attempt;      // which attempt for this request
-  LogSegment *ls;  // the log segment i'm committing to
+  __u32 attempt = 0;      // which attempt for this request
+  LogSegment *ls = nullptr;  // the log segment i'm committing to
 
 private:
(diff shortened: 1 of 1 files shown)


gregsfortytwo:
Hmm. I guess I know you can do this in some cases and I'm guessing recent C++ revisions have added more. But I don't know the rules; is there a benefit beyond Don't Repeat Yourself? Is there a cost?

stiopaa1:
@gregsfortytwo 
Advantages would be as you mentioned no code duplication and then in case you ever wanted to add additional constructors in the future, there is no way to leave those member uninitialised even if you forget to initialise them.
There is no additional cost: (http://en.cppreference.com/w/cpp/language/initializer_list)
>If a non-static data member has an default member initializer and also appears in a member initializer list, then member initializer list is executed and the default member initializer is ignored:
struct S {
    int n = 42;   // default member initializer
    S() : n(7) {} // will set n to 7, not 42
};
(since C++11)

The only surprising effect it has in C++11 is that it makes a class a not an aggregate. More information about it here:
http://stackoverflow.com/questions/27118535/c11-aggregate-initialization-for-classes-with-non-static-member-initializers

gregsfortytwo:
LGTM

PR #11456, opened 2016-10-13
mon: send updated monmap to its subscribers

tchaikov:
prior to this change, we send monmap when serving the subscription
requests, but the updates are not sent to the subscribers anymore.
so we need to send latest monmap to subscribers and update the
subscription status accordingly when the monmap is updated.

http://tracker.ceph.com/issues/17558

### DIFF
--- a/src/mon/Monitor.cc
+++ b/src/mon/Monitor.cc
@@ -4332,7 +4332,7 @@ void Monitor::handle_subscribe(MonOpRequestRef op)
 	pgmon()->check_sub(s->sub_map["osd_pg_creates"]);
       }
     } else if (p->first == "monmap") {
-      check_sub(s->sub_map["monmap"]);
+      monmon()->check_sub(s->sub_map[p->first]);
     } else if (logmon()->sub_name_to_id(p->first) >= 0) {
       logmon()->check_sub(s->sub_map[p->first]);
     } else if (p->first == "mgrmap" || p->first == "mgrdigest") {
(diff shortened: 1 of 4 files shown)


tchaikov:
@jecluis do you mind taking a look?

tchaikov:
@gregsfortytwo could you take a look also?

jecluis:
LGTM. Must admit I'm a bit baffled this is a thing we were not doing.

gregsfortytwo:
This looks good, but since we apparently weren't doing this and think we should have, this needs a test before merging please.

tchaikov:
thanks for the reviews! i am preparing a test in ceph-qa-suite.

tchaikov:
@gregsfortytwo test posted at https://github.com/ceph/ceph-qa-suite/pull/1217

a test passed at http://pulpito.ceph.com/kchai-2016-10-20_03:33:37-rados-wip-kefu-testing---basic-mira/

could you take a look again?

jazeltq:
If the osd lose mon connection, it will hunt mon using old monmap. Why it won't connect for ever?  Can u explain more about this? thanks a lot.

tchaikov:
> Why it won't connect for ever?

i don't really understand what you mean.

jazeltq:
Oh, i view a little from the http://tracker.ceph.com/issues/17558, and
In number 4, it said that osd.0 will not be able to connect cluster any more.
I do not know why osd.0 can not connect cluster any more, so i ask for that ....


reproduce steps.

1. create a cluster with 3 mon, 1 osd
2. create another mon manually, wait until the quorum is formed. update the ceph.conf also, otherwise the "ceph" cli won't be able to get updated monmap
3. "ceph mon remove a" 
4. osd.0 won't be able to connect to cluster anymore.

if the updated monmap is sent to osd.0, osd.0 will be able to switch to the new monitor, even it is not in its initial monmap.

tchaikov:
> if the updated monmap is sent to osd.0, osd.0 will be able to switch to the new monitor, even it is not in its initial monmap.

but without this fix, the updated monmap won't be sent to osd.0. am i missing anything?

jazeltq:
I see. Thanks a lot....

PR #11468, opened 2016-10-13
os/bluestore: traverse buffer_map in reverse order when spliting BufferSpace

xiexingguo:


xiexingguo:
Locally tested:
```
2016-10-13 18:38:29.206106 7fe98a623640  1 bluestore(bluestore.test_temp_dir) fsck finish with 0 errors in 5.328099 seconds
[       OK ] ObjectStore/StoreTest.TooManyBlobsTest/2 (37320 ms)
[----------] 63 tests from ObjectStore/StoreTest (9766848 ms total)

[==========] 63 tests from 1 test case ran. (9766848 ms total)
[  PASSED  ] 63 tests.

**Comment by liewegas** [comment]
Why is going in reverse more efficient?

File: src/os/bluestore/BlueStore.cc [hunk 1]
```diff
@@ -1042,13 +1042,14 @@ void BlueStore::BufferSpace::split(size_t pos, BlueStore::BufferSpace &r)
 {
   std::lock_guard<std::recursive_mutex> lk(cache->lock);
   assert(r.cache == cache);
-  auto p = buffer_map.begin();
-  while (p != buffer_map.end() &&
-	 p->second->end() <= pos) {
-    dout(30) << __func__ << " skip " << *p->second << dendl;
-    ++p;
-  }
```
xiexingguo: @liewegas 
So we can avoid the skip process above, especially if we have to skip all.

liewegas:
Ah, of course

PR #11469, opened 2016-10-13
jewel: rgw:user email can modify to empty when it has values

dachary:
http://tracker.ceph.com/issues/17538


dachary:
@cbodley does this backport look good to merge ? It passed the rgw suite http://tracker.ceph.com/issues/17487#note-18

PR #11479, opened 2016-10-13
mon/pgmap: DIRTY display "15E" when run command "ceph df detail"

renhwztetecs:
num_objects_dirty may be a negative number,when executing clearing DIRTY flag.
if num_objects_dirty is negative number,and changing the type of num_objects_dirty
from int64_t to int64_t in the dump_object_stat_sum()

add the judgment to ensure num_objects_dirty not less than zero

Fixes:http://tracker.ceph.com/issues/17565

Signed-off-by: huanwen ren ren.huanwen@zte.com.cn


tchaikov:
@renhwztetecs i don't understand why `num_objects_dirty` can be a negative number. seems you are covering a bug here with this patch, IMO.

tchaikov:
@renhwztetecs ping?

renhwztetecs:
sorry，I am late. 
I have not found the reason, here is only just a patch to prevent overflow
thanks

tchaikov:
instead of  covering a bug, we need to fix it. so i am closing the PR.

PR #11481, opened 2016-10-13
cmake: check WITH_RADOSGW for fcgi and expat dependencies

ddiss:
The fcgi and expat libraries are only used by rgw, so only make them
hard requirements if WITH_RADOSGW is set.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -240,13 +240,6 @@ find_package(leveldb REQUIRED)
 find_file(HAVE_LEVELDB_FILTER_POLICY leveldb/filter_policy.h PATHS ${LEVELDB_INCLUDE_DIR})
 endif(${WITH_LEVELDB})
 
-option(WITH_EXPAT "EXPAT is here" ON)
-if(${WITH_EXPAT})
-find_package(EXPAT REQUIRED)
-endif(${WITH_EXPAT})
-
-find_package(fcgi REQUIRED)
-
 find_package(atomic_ops REQUIRED)
 message(STATUS "${ATOMIC_OPS_LIBRARIES}")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -325,6 +318,10 @@ endif(WITH_XIO)
 
 #option for RGW
 option(WITH_RADOSGW "Rados Gateway is enabled" ON)
+if(${WITH_RADOSGW})
```
tchaikov: could you put `if(WITH_RADOSGW)` instead, just to be more consistent with other places where we check options. (yes, i know we use `if(${WITH_RADOSGW})` somewhere else. but we can take every chance to fix it)

CMakeLists.txt
```diff
@@ -325,6 +318,10 @@ endif(WITH_XIO)
...
+find_package(EXPAT REQUIRED)
```
tchaikov: indentation with two spaces.

CMakeLists.txt
```diff
@@ -325,6 +318,10 @@ endif(WITH_XIO)
...
+find_package(fcgi REQUIRED)
```
tchaikov: ditto.

ddiss:
Thanks for the feedback. Will fix and resubmit.

ddiss:
Changes since last version:
- s/${WITH_RADOSGW}/WITH_RADOSGW/
- indent nested find_package() calls

tchaikov:
lgtm if gitbuilders are happy.

ddiss:
The gitbuilder failure appears to be unrelated to the change here.

tchaikov:
pushed to ceph/ceph:wip-11481

PR #11482, opened 2016-10-13
os/bluestore: GC infra refactor, more UTs and GC range calculation fixes

ifed01:

### DIFF
--- a/src/os/bluestore/BlueStore.h
+++ b/src/os/bluestore/BlueStore.h
@@ -636,7 +636,7 @@ public:
       }
       int s = seek_shard(offset);
       assert(s >= 0);
-      if (s == shards.size() - 1) {
+      if (s == (int)shards.size() - 1) {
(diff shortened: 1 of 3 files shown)


src/os/bluestore/BlueStore.h
```diff
@@ -735,6 +735,18 @@ class BlueStore : public ObjectStore,
   };
   typedef boost::intrusive_ptr<Onode> OnodeRef;
 
+  //Made a specific class to be able to verify GC logics in Unit Tests
+  struct GarbageCollector {
```
liewegas: I think it's simpler to just make this a method of ExtentMap, which already includes several unit tests in test_bluestore_types.cc

ifed01: Yeah, that's better...

ifed01:
@liewegas - fixed

PR #11483, opened 2016-10-13
os/bluestore: a few cleanups

liewegas:


xiexingguo:
I think you shall withdraw the changes related to StupidAllocator in https://github.com/ceph/ceph/pull/11483/commits/e91b16c8dede2fc9527a91c4306e939399d97e7a.

Seems the original design is that a pre-initialized(and allocated) `extents `array(**for efficiency**) is allowed to passed in and used to initialize a `ExtentList `later.   So we can not `assert(extents->empty());`.

See https://github.com/ceph/ceph/blob/master/src/os/bluestore/BlueFS.cc#L1720 
and https://github.com/ceph/ceph/blob/master/src/test/objectstore/Allocator_test.cc#L70
and https://github.com/ceph/ceph/blob/master/src/os/bluestore/BlueStore.cc#L7540
and so on...

Or introduce more changes to fix above all..

liewegas:
yeah, i'll drop that one for now.  i don't really like the AllocExtent stuff but it's not pressing.

Eventually I think we want to add a bdev field to bluestore_pextent_t, just like the bluefs_extent_t,s o that we can support multiple devices.  A bunch of other stuff needs to be adjusted to handle multiple bdev lba spaces, though (like ref_map_t and a bunch of stuff in fsck()).

xiexingguo:
Reviewed-by:

PR #11485, opened 2016-10-13
rgw structures rework

yehudasa:
Create new types:
- rgw_pool (to represent pools instead of rgw_bucket)
- rgw_raw_obj (to represent the raw rados object instead of rgw_obj)
- rgw_obj_index_key

 Don't store placement info in rgw_bucket (if possible, still keeping explicit placement for backward compatibility when dealing with older data). head object will always reside in the bucket's default placement target pool, (from the zone config), tail depends on what's in the manifest.

Rework rgw_obj; rename some of the fields, don't refer to oid as object anymore and don't auto generate the oid. Don't encode the oid, but the name, ns, and instance. Split rgw_obj_key and rgw_obj_index_key, and rgw_obj handles both as needed.

Pools can now be defined with a namespace (<pool>[:ns]), and consolidate a few of the default pools.

### DIFF
--- a/src/rgw/rgw_common.h
+++ b/src/rgw/rgw_common.h
@@ -912,7 +912,7 @@ struct rgw_bucket {
   std::string name;
   std::string marker;
   std::string bucket_id;
-  rgw_data_placement_target placement;
+  rgw_data_placement_target explicit_placement;
 
   std::string oid; /*
                     * runtime in-memory only info. If not empty, points to the bucket instance object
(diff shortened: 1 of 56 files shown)


cbodley:
@yehudasa did you forget to commit `test_rgw_common.cc`? getting this error from cmake:
```
CMake Error at src/test/rgw/CMakeLists.txt:25 (add_library):
  Cannot find source file:

    test_rgw_common.cc
```

yehudasa:
@cbodley fixed now

src/rgw/rgw_rados.h
```diff
@@ -2810,18 +3052,21 @@ class RGWRados
...
+  int delete_obj(RGWObjectCtx& obj_ctx,
+                         const RGWBucketInfo& bucket_owner,
                          const rgw_obj& src_obj,
                          int versioning_status,
                          uint16_t bilog_flags = 0,
                          const ceph::real_time& expiration_time = ceph::real_time());
 
+  /** Delete a raw object.*/
+  int delete_raw_obj(const rgw_raw_obj& obj);
+
```
yehudasa: trying to separate the notion of a raw obj vs a system obj. Maybe this can be reworked later to avoid duplicate code though.

src/rgw/rgw_common.h
```diff
@@ -788,29 +974,37 @@ struct rgw_bucket {
...
+      }
+      if (struct_v >= 7) {
+        ::decode(explicit_placement.data_extra_pool.name, bl);
+      }
     }
     if (struct_v >= 8) {
       ::decode(tenant, bl);
     }
+    if (struct_v >= 10) {
+      bool decode_explicit = !explicit_placement.data_pool.empty();
```
yehudasa: no need to initialize at all, bad copy paste

yehudasa:
@cbodley PR that adds ragweed support here: https://github.com/ceph/ceph/pull/13644

yehudasa:
Upgrade test is green here:
http://pulpito.ceph.com/yehudasa-2017-02-24_15:31:11-rgw:upgrade-wip-bucket-cleanup-3---basic-smithi/

smithfarm:
@yehudasa I understand you'd like to backport a9ec5e8ce184e19c009863db4d3519f9d8af91bd from this PR, but to which stable release(s)?

src/rgw/rgw_rados.cc
```diff
@@ -1956,13 +2039,12 @@ void RGWObjManifest::obj_iterator::operator++()
...
-int RGWObjManifest::generator::create_begin(CephContext *cct, RGWObjManifest *_m, rgw_bucket& _b, rgw_obj& _h)
+int RGWObjManifest::generator::create_begin(CephContext *cct, RGWObjManifest *_m, const string& placement_rule, rgw_bucket& _b, rgw_obj& _obj)
 {
   manifest = _m;
 
-  bucket = _b;
-  manifest->set_tail_bucket(_b);
-  manifest->set_head(_h, 0);
+  manifest->set_tail_placement(placement_rule, _b);
+  manifest->set_head(placement_rule, _obj, 0);
```
fangyuxiangGL: so the tail stripes and head object will still stored as a same placement_rule?

fangyuxiangGL:
@yehudasa 

"head object will always reside in the bucket's default placement target pool, (from the zone config), tail depends on what's in the manifest."

head object and tail stripes still have same rados store way as your code.

PR #11486, opened 2016-10-14
client: get gid from MonClient

batrick:
This is consistent with other parts of the code that set whoami.


gregsfortytwo:
How'd you spot this?

batrick:
Just looking around the gid code while working on #11487.

gregsfortytwo:
LGTM to, but these kinds of changes can do nasty things so let's test carefully.

PR #11487, opened 2016-10-14
cephfs: add gid to asok status

batrick:
This is useful e.g. for building a database with a unique ID for each
entity.


gregsfortytwo:
LGTM, assuming it builds.

PR #11489, opened 2016-10-14
os/bluestore: parallel txn submission changes

majianpeng:


src/os/bluestore/BlueStore.cc
```diff
@@ -6466,11 +6464,27 @@ void BlueStore::_kv_sync_thread()
...
 	// kv metadata updates
 	_txc_finalize_kv(*it, t);
+	_txc_update_store_statfs(*it, &vstatfs);
 	// cleanup the wal
 	string key;
 	get_wal_key(wt.seq, &key);
 	t->rm_single_key(PREFIX_WAL, key);
       }
+      {
+	if (vstatfs.is_empty()) {
```
ifed01: !is_empty ???

src/os/bluestore/BlueStore.cc
```diff
@@ -6409,33 +6409,42 @@ void BlueStore::_kv_sync_thread()
...
+	}
+	if (high_blobid + g_conf->bluestore_blobid_prealloc/2 > blobid_max) {
+	  blobid_max = high_blobid + g_conf->bluestore_blobid_prealloc;
+	  bufferlist bl;
+	  ::encode(blobid_max, bl);
+	  t->set(PREFIX_SUPER, "blobid_max", bl);
+	  dout(10) << __func__ << " blobid_max now " << blobid_max << dendl;
+	}
+      }
+
```
liewegas: This might appear better, but it doesn't actually work, because the transactions that increased the max are submitting to rocksdb in parallel, and rocksdb is free to persist them without our txn here that increases max.  That means we might commit something that exceeds the committed max.  In order to make || submission work I think the parallel thread needs to block if it exceeds max, or somehow make it fall back to submitting it via the kv_sync_thread.

liewegas:
I take it you're testing this.. how does the performance compare when you do parallel submission?

src/os/bluestore/BlueStore.cc
```diff
@@ -6045,10 +6045,42 @@ void BlueStore::_txc_state_proc(TransContext *txc)
       _txc_finalize_kv(txc, txc->t);
       if (!sync_transaction) {
 	if (sync_submit_transaction) {
+	  std::lock_guard<std::mutex> l(id_lock);
```
majianpeng: @liewegas . I review the code and found how to update nid_max/blobid_max for parallel submit-txn
a)before update oid/blobkid of txn(which larger than *_max) it must firstly update *_max
b)don't occur this problem: t1->max_nid > t2->max_nid, But the submit order is t1, t2.
I firstly use the following code:
__assign_nid {
if (nid_last > nid_max && (parallel_submit_txn)
   wait(kv_sync_thread update *_max).
}
The other is this commit. I don't know which is better.  Hope your comments! Thanks!

liewegas: I think a wait would work, but I suspect it'll perform better if we just fall back to committing this particular txn in kv_sync_thread in this case.  It'll happen to any other concurrent txn that needs a nid or blobid too.  But all of them will commit *with* the max update.  If we do the wait like you suggest, then they have to wait for the max increase to commit, and *then* they will commit with the batch after that.

src/os/bluestore/BlueStore.cc
```diff
@@ -6046,14 +6046,13 @@ void BlueStore::_txc_state_proc(TransContext *txc)
         sb->bc.finish_write(txc->seq);
       }
       txc->shared_blobs_written.clear();
+      _txc_finalize_kv(txc, txc->t);
```
liewegas: This change breaks if you are using ExtentFreelistManager.

We should add an additional virtual method to FreelistManager.h called something bool supports_parallel_transactions() that is true for the bitmap version and false for the extent version.  Then we drop this change, and add a new patch that prevents you from using sync_transaction or sync_submit_transaction entirely.  (We could try to make sync_submit_transaction work but I don't think there's any reason to bother.)

src/os/bluestore/BlueStore.cc
```diff
@@ -6000,21 +6000,17 @@ BlueStore::TransContext *BlueStore::_txc_create(OpSequencer *osr)
...
+  vstatfs->allocated() += txc->statfs_delta.allocated();
+  vstatfs->stored() += txc->statfs_delta.stored();
+  vstatfs->compressed() += txc->statfs_delta.compressed();
+  vstatfs->compressed_allocated() += txc->statfs_delta.compressed_allocated();
+  vstatfs->compressed_original() += txc->statfs_delta.compressed_original();
 
-  bufferlist bl;
-  txc->statfs_delta.encode(bl);
-
-  txc->t->merge(PREFIX_STAT, "bluestore_statfs", bl);
```
liewegas: I don't think this change is safe, either, because rocksdb could choose to persist one of these earlier transactions (with an actual change) before persisting the final sync transaction (with the batched stats changes), and then teh system crashes, and the stats are left out of sync.

liewegas:
Hrm, I think there are different issues with each of these patches, but they all stem from having a global flag indicating which path we take.  Please take a look at https://github.com/ceph/ceph/pull/11537.  I think that if we do this first, then we can make a per-txc decision to take the non-sync path if, for example, our nid or blobid exceeds the committed max.  That way we get the more-parallel path most of the time, but we can take the marginally slower path if we have to.  This is waaay less fragile and complicated (I think) than blocking for a nid/blobid max update.  What do you think?

src/os/bluestore/BlueStore.cc
majianpeng: @liewegas .  I see https://trello.com/c/YJZymq8O/8-bluestore-fix-and-test-inline-parallel-txn-submission also related parallel submit for Allocator. But i don't know why for extentfreelist dose not support paralle? Because the global lock or other reasons?  Thanks!

liewegas:
I've updated #11537.  Does that look reasonable?

PR #11490, opened 2016-10-14
os/bluestore: use iterator for erase() method directly

xiexingguo:


ifed01:
Looks like there is duplicate write_onode call for _do_truncate and _do_clone_range too.

ifed01:
I'm not completely sure if this duplication's worth fixing but I'd prefer to have ALL write_onode calls at the same level for the sake of uniformity, i.e. either at entry points (_zero, _truncate, _clone, etc) or at some underlying ones (_do_zero, _do_truncate, _do_clone_range, etc). Just IMHO though...

xiexingguo:
@ifed01 I am with you.

Dropped that change.

liewegas:
@ifed01 @xiaoxichen agreed on being consistent.  Probably _do_*....  in a handful of cases that will include a duplicate insert, but it's probably safer?  No strong preference though.

PR #11494, opened 2016-10-14
[RGW] Wip rgw compression

aclamk:
Continuation of #10932.
I closed old one by mistake.

### DIFF
--- a/src/rgw/rgw_rados.cc
+++ b/src/rgw/rgw_rados.cc
@@ -2197,11 +2197,12 @@ void RGWObjVersionTracker::generate_new_write_ver(CephContext *cct)
   append_rand_alpha(cct, write_version.tag, write_version.tag, TAG_LEN);
 }
 
-int RGWPutObjProcessor::complete(string& etag, real_time *mtime, real_time set_mtime,
+int RGWPutObjProcessor::complete(size_t accounted_size, const string& etag,
+                                 real_time *mtime, real_time set_mtime,
                                  map<string, bufferlist>& attrs, real_time delete_at,
-                                 const char *if_match, const char * if_nomatch)
+                                 const char *if_match, const char *if_nomatch)
 {
-  int r = do_complete(etag, mtime, set_mtime, attrs, delete_at, if_match, if_nomatch);
+  int r = do_complete(accounted_size, etag, mtime, set_mtime, attrs, delete_at, if_match, if_nomatch);
   if (r < 0)
     return r;
 
@@ -2511,8 +2512,10 @@ int RGWPutObjProcessor_Atomic::complete_writing_data()
   return 0;
 }
 
-int RGWPutObjProcessor_Atomic::do_complete(string& etag, real_time *mtime, real_time set_mtime,
-                                           map<string, bufferlist>& attrs, real_time delete_at,
+int RGWPutObjProcessor_Atomic::do_complete(size_t accounted_size, const string& etag,
+                                           real_time *mtime, real_time set_mtime,
+                                           map<string, bufferlist>& attrs,
+                                           real_time delete_at,
                                            const char *if_match,
                                            const char *if_nomatch) {
   int r = complete_writing_data();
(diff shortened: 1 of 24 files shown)


cbodley:
unlucky timing on the rebase - gitbuilders, which use WITH_LTTNG=ON, hit some unrelated build failures. i had to rebase the ceph/wip_rgw_compression branch on top of 9983cd36fa15dbd2ff8bc9d5dd1a21f058f79141 to fix

cbodley:
the teuthology queue is very sluggish lately, but most of the results are in from the [latest run](http://pulpito.ceph.com/cbodley-2016-10-14_16:33:02-rgw-wip_rgw_compression---basic-mira/). so far all of the failures are unrelated known issues: valgrind, sync agent (HTTPConnectionPool timeouts and 'object not deleted from destination zone'), and a single s3tests failure caused by http://tracker.ceph.com/issues/17465

cbodley:
there's an interesting new merge conflict from https://github.com/ceph/ceph/pull/11567, which fixes a bug with zero-length segments in `RGWGetObj::read_user_manifest_part()`. this pr replaces that loop with a call to `RGWRados::Object::Read::iterate()`, so we'll need to make sure that function doesn't have to same issue with empty segments. i'll work with @mdw-at-linuxbox to get that tested

cbodley:
@yehudasa could you please do a final review? only two remaining blockers: my XXX comment in `RGWRados::copy_obj_data()`, and the merge conflict with the zero-length segment fix from @mdw-at-linuxbox

src/rgw/rgw_rados.cc
```diff
@@ -7572,7 +7605,8 @@ int RGWRados::copy_obj_data(RGWObjectCtx& obj_ctx,
     }
   }
 
-  ret = processor.complete(etag, mtime, set_mtime, attrs, delete_at);
+  // XXX: need to copy over compression attr and its orig_size here?
+  ret = processor.complete(ofs, etag, mtime, set_mtime, attrs, delete_at);
```
aclamk: It is OK.

Reasoning:
Here we copy object, including attrs. Copying process is compression agnostic. However, all attrs are preserved, I assume including RGW_ATTR_COMPRESSION. Basing on that object can be read.

cbodley: (discussed briefly during standup)
i agree that the compression attribute is preserved. i just worry about `ofs` as the first argument to `complete()` - my changes involved passing the uncompressed size here, so the bucket could track both compressed and uncompressed sizes

cbodley:
@aclamk with help from @mdw-at-linuxbox, I was able to run his test successfully on a rebased version of this branch. this means that, when rebasing, you can resolve the conflicts in `RGWGetObj::read_user_manifest_part()` normally. the conflict in `cls_rgw.cc` is also trivial

i pushed my rebased branch to https://github.com/cbodley/ceph/commits/wip-rgw-compression for comparison if that helps

src/rgw/rgw_rados.cc
cbodley: i added some code to decode the uncompressed size from the compression attribute here, and included the commit https://github.com/cbodley/ceph/commit/8e0d9a5aa9f73ad91d1aa29f361672ce95231e7f in https://github.com/cbodley/ceph/commits/wip-rgw-compression - feel free to cherry-pick that (it's passing s3 and swift tests)

cbodley:
would love to see this make the kraken freeze

@aclamk could you please do the rebase and pull in my fix for `copy_obj_data()` at https://github.com/cbodley/ceph/commit/8e0d9a5aa9f73ad91d1aa29f361672ce95231e7f?

@yehudasa could you please review my commits related to tracking `accounted_size` in the bucket stats? specifically, that I'm not missing anything w.r.t. backward compatibility?

cbodley:
thanks @aclamk, looks good. i opened https://github.com/aclamk/ceph/pull/2 against this branch to add some documentation

cbodley:
thanks to @aclamk and @Ved-vampir for their great work on this!

liewegas:
Yay!

PR #11499, opened 2016-10-14
cmake: find GIT_VER variables if there is no .git dir

alimaredia:
find CEPH_GIT_VER & CEPH_GIT_NICE_VER if there is no
ceph/.git dir by parsing ceph/src/.git_version.

scenario is for building from source tarballs generated
by make-dist, since make-dist generates .git_version
by running make_version, and then packs .git_version
into the tarball.


src/CMakeLists.txt
```diff
@@ -182,6 +182,13 @@ option(ENABLE_GIT_VERSION "build Ceph with git version string" ON)
 if(${ENABLE_GIT_VERSION})
   get_git_head_revision(GIT_REFSPEC CEPH_GIT_VER)
   git_describe(CEPH_GIT_NICE_VER --always)
+  #if building from a source tarball via make-dist
+  if(${CEPH_GIT_VER} STREQUAL "GITDIR-NOTFOUND")
+    message(STATUS "Ceph/.git directory not found, parsing ${CMAKE_CURRENT_SOURCE_DIR}/.git_version for CEPH_GIT_VER and CEPH_GIT_NICE_VER")
+    file(STRINGS ${CMAKE_CURRENT_SOURCE_DIR}/.git_version CEPH_GIT_SHA_AND_TAG)
```
alimaredia: @alfredodeza if file() can't find src/.git_version then it errors when running the `cmake` command and won't build. Is that fine or should I have a further conditional for if it can't find that file?

alfredodeza: that should be enough I think

cbodley: agreed - the user can set `ENABLE_GIT_VERSION=OFF` if all else fails

alfredodeza:
Installing from: https://4.chacra.ceph.com/r/ceph/wip-kraken-no-git-dir/c5c6b59f296f0ace3a6cfd939027118e82ab50a4/ubuntu/xenial/flavors/default/

Gives the correct result:

    $ ceph --version
    ceph version v11.0.1-1-gc5c6b59 (c5c6b59f296f0ace3a6cfd939027118e82ab50a4)

PR #11500, opened 2016-10-14
CMakeLists.txt: enable LTTNG by default

liewegas:
This makes dev builds more closely match package builds.


tchaikov:
lgtm.

PR #11501, opened 2016-10-14
DNM: test: Remove check for "load: jerasure.*lrc"

dzafman:
Fix for broken test-erasure-code.sh and test-erasure-eio.sh


dachary:
I'm curious about why this is necessary ?

dzafman:
@dachary This commit is the only one removing the logging which includes "load: " that was merged Aug 17 d0cd88b3ee805854c6c0c5272f14793e95211b43   It doesn't look like the erasure code plug-in would be loaded by async msg code but what do I know.

dzafman:
Added another minor fix for a problem I saw on my build machine due to low space condition.

dachary:
The message is logged by ErasureCodePluginRegistry::load
<pre>
  (*plugin)->library = library;

  *ss << __func__ << ": " << plugin_name << " ";
</pre>

dachary:
loading the plugin happens at a very early stage in the mon bootstrap sequence. But there can be a race where
<pre>
CEPH_ARGS='' ceph --admin-daemon $dir/ceph-mon.a.asok log flush || return 1
</pre>
succeeds before the erasure code plugins are initialized. The log flush command is registered in the CephContext constructor and only later is global_init_preload_erasure_code called.

I think a safe way to avoid this rare case is to run ceph -s after the log flush. Does that make sense ?

dzafman:
@ldachary The test-erasure-*.sh tests aren't even running in master.  I know I saw them failing somewhere.

dachary:
They are not running ? Do you mean they have been removed from make check at some point ?

dzafman:
@ldachary Oops, I disabled them in master a while ago, but had them enabled in a new branch.   So I still want a variation of this fix.

dzafman:
@ldachary I tried a "ceph -s" that failed.  Then added a 30 second sleep it still fails the grep!  It isn't a rare race for in my build tree.

dzafman:
@ldachary Using gdb I not only never hit the load() function but the address for it isn't even available for the class " ErasureCodePluginRegistry"  Something must be statically linked or something else by default in my build tree.  I used vstart.sh to first config up and 1 MON 6 OSD set-up.  Then I kill and use gdb to check the ceph-mon start-up.

`
(gdb) info breakpoints
Num     Type           Disp Enb Address    What
1       breakpoint     keep y   <PENDING>  ErasureCodePluginRegistry::load
`

dzafman:
@ldachary Just to verify I put the greps back and Jenkins fails for both test the same way on the mon log.  Now I'm going to push it with my fix again.  I can't get a sleep or ceph -s to make it work.  See my gdb comment above.

```
/home/jenkins-build/build/workspace/ceph-pull-requests/src/test/erasure-code/test-erasure-code.sh:35: run:  grep 'load: jerasure.*lrc' testdir/test-erasure-code/mon.a.log
/home/jenkins-build/build/workspace/ceph-pull-requests/src/test/erasure-code/test-erasure-code.sh:35: run:  return 1
```

dachary:
I see they have been disabled in 5bc55338f5a1645bc651811fae2f89ad855ff86e. What I don't quite understand is the lack of message because [global_init_preload_erasure_code](https://github.com/ceph/ceph/blob/master/src/ceph_mon.cc#L493) calls [ErasureCodePluginRegistry::preload](https://github.com/ceph/ceph/blob/master/src/global/global_init.cc#L494) which loops over the [osd_erasure_code_plugins](https://github.com/ceph/ceph/blob/master/src/common/config_opts.h#L644) which has at least **jerasure** and **lrc** and will return on error and abort the mon/osd unless it [adds the expected message](https://github.com/ceph/ceph/blob/master/src/erasure-code/ErasureCodePlugin.cc#L179) to be [printed at verbose level 10](https://github.com/ceph/ceph/blob/master/src/global/global_init.cc#L501).

Late september the [global_init_preload_erasure_code was moved](https://github.com/ceph/ceph/commit/e7e0b1bc1e159e35b151eeb9d5cc64bc57cd1b76) created to factorize ceph_mon.cc and ceph_osd.cc and the **dout(10)** was not modified. However they have
<pre>
#define dout_subsys ceph_subsys_mon
</pre>
and
<pre>
#define dout_subsys ceph_subsys_osd
</pre>
which commands a different level of debug than what is in global_init.cc
<pre>
#define dout_subsys ceph_subsys_
</pre>

I think that raising the proper debug level will fix the problem. @bassamtabbara would have noticed it while working on the patch if the tests had not been disabled.

dzafman:
This change has been applied to ceph/ceph#9304.  I did change the dout(10) to dout(0) for the plugin load message.

PR #11505, opened 2016-10-14
rgw: clean up RGWShardedOmapCRManager on early return

cbodley:
ShardedOmapCRManager is spawning coroutines that sleep. if we don't
finish them before trying to drain_all(), they'll lead to deadlock

Fixes: http://tracker.ceph.com/issues/17571

### DIFF
--- a/src/rgw/rgw_sync.cc
+++ b/src/rgw/rgw_sync.cc
@@ -814,6 +814,7 @@ public:
       }
       if (get_ret_status() < 0) {
         ldout(cct, 0) << "ERROR: failed to fetch metadata sections" << dendl;
+        yield entries_index->finish();
         yield lease_cr->go_down();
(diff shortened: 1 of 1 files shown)


yehudasa:
@cbodley lgtm, do we need to run specific tests on this one?

cbodley:
@yehudasa this one would be hard to reproduce in the absence of other bugs

i would (someday) love to hook up unit tests to each of these coroutines so we can get coverage of the error paths, but it'll take some work to be able to pass in mocks for each of the child crs they spawn. would you like me to try FetchAllMetaCR as a start?

yehudasa:
@cbodley sounds good

cbodley:
unit tests added as separate pr - they succeed when this fix is applied, and deadlock without

PR #11506, opened 2016-10-14
rgw: clear data_sync_cr if RGWDataSyncControlCR fails

cbodley:
async notifications will still try to call wakeup() on RGWDataSyncControlCR if it fails, leading to segfault

Fixes: http://tracker.ceph.com/issues/17569


yehudasa:
@cbodley lgtm

PR #11510, opened 2016-10-15
librbd: batch object map updates during trim

vshankar:
Shrinking an image can result in huge number of ObjectMap
updates -- two for each object being removed. This results
in lots of requests to be send to OSDs especially when
trimming a gigantus image. This situation can be optimized
by sending batch ObjectMap updates, significantly cutting
down the OSD traffic.

Fixes: http://tracker.ceph.com/issues/17356
Signed-off-by: Venky Shankar vshankar@redhat.com

### DIFF
--- a/src/librbd/AioObjectRequest.h
+++ b/src/librbd/AioObjectRequest.h
@@ -330,11 +330,14 @@ private:
 
 class AioObjectTrim : public AbstractAioObjectWrite {
 public:
+  // we'd need to only conditionally specify if a post object map
+  // update is needed. pre update is decided as usual (by checking
+  // the state of the object in the map).
   AioObjectTrim(ImageCtx *ictx, const std::string &oid, uint64_t object_no,
-                const ::SnapContext &snapc, Context *completion)
+                const ::SnapContext &snapc, Context *completion,
+                bool post_object_map_update)
     : AbstractAioObjectWrite(ictx, oid, object_no, 0, 0, snapc, completion,
-                             true) {
-  }
+                             true), m_post_object_map_update(post_object_map_update) { }
 
 protected:
   virtual void add_write_ops(librados::ObjectWriteOperation *wr) {
@@ -350,8 +353,11 @@ protected:
   }
 
   virtual bool post_object_map_update() {
-    return true;
+    return m_post_object_map_update;
   }
+
+private:
+  bool m_post_object_map_update;
 };
 
 class AioObjectTruncate : public AbstractAioObjectWrite {
--- a/src/librbd/operation/TrimRequest.cc
+++ b/src/librbd/operation/TrimRequest.cc
@@ -46,7 +46,7 @@ public:
     ldout(image_ctx.cct, 10) << "removing (with copyup) " << oid << dendl;
 
     AioObjectRequest<> *req = new AioObjectTrim(&image_ctx, oid, m_object_no,
-                                                m_snapc, this);
+                                                m_snapc, this, false);
     req->send();
     return 0;
   }
@@ -131,8 +131,18 @@ bool TrimRequest<I>::should_complete(int r)
 
   RWLock::RLocker owner_lock(image_ctx.owner_lock);
   switch (m_state) {
+  case STATE_PRE_COPYUP:
+    ldout(cct, 5) << " PRE_COPYUP" << dendl;
+    send_copyup_objects();
+    break;
+
   case STATE_COPYUP_OBJECTS:
     ldout(cct, 5) << " COPYUP_OBJECTS" << dendl;
+    send_post_copyup();
+    break;
+
+  case STATE_POST_COPYUP:
+    ldout(cct, 5) << " POST_COPYUP" << dendl;
     send_pre_remove();
     break;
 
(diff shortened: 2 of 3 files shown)


src/librbd/operation/TrimRequest.cc
```diff
@@ -170,11 +180,11 @@ bool TrimRequest<I>::should_complete(int r)
 
 template <typename I>
 void TrimRequest<I>::send() {
-  send_copyup_objects();
+  send_update_objectmap();
 }
 
 template <typename I>
-void TrimRequest<I>::send_copyup_objects() {
+void TrimRequest<I>::send_update_objectmap() {
```
dillaman: Minor: this state doesn't seem to do anything besides pick which method to invoke next -- why not just combine it w/ ```send_pre_copyup``` and skip to ```send_pre_remove``` if copyup isn't needed.

src/librbd/operation/TrimRequest.cc
```diff
@@ -245,6 +263,46 @@ void TrimRequest<I>::send_remove_objects() {
...
+      ldout(image_ctx.cct, 5) << this << " send_pre_copyup: "
+                              << " copyup_start=" << m_copyup_start
+                              << " copyup_end=" << m_copyup_end << dendl;
+      m_state = STATE_PRE_COPYUP;
+
+      assert(image_ctx.exclusive_lock->is_lock_owner());
+
+      Context *ctx = this->create_callback_context();
+      RWLock::WLocker object_map_locker(image_ctx.object_map_lock);
+      if (!image_ctx.object_map->aio_update(m_copyup_start, m_copyup_end,
```
dillaman: I don't think this is correct. The purpose of the copy-up state is to ensure that if you have snapshots *and* if there are one-or-more blocks that haven't been copied up to the child image, we need to ensure that these snapshots have the backing parent image data just in case the child image is flattened.

This path would only be executed if you are shrinking an image and will not be executed upon an image removal (since you cannot have snapshots when you remove the image). Since we need to read each affected block from the parent image for copyup to know if the parent image has backing data, I am not sure we can optimize this path to a batch object map update.

src/librbd/AioObjectRequest.h
```diff
@@ -348,10 +348,6 @@ class AioObjectTrim : public AbstractAioObjectWrite {
   virtual void pre_object_map_update(uint8_t *new_state) {
     *new_state = OBJECT_PENDING;
   }
-
-  virtual bool post_object_map_update() {
```
dillaman: I didn't realize that the trim state machine already performed a batch update -- I think this removal is correct and is what was (incorrectly) causing per-object object map updates.

vshankar: no worries...

dillaman:
@vshankar It appears that all the fsx tests have failed due to IO mismatches:
http://pulpito.ceph.com/jdillaman-2016-10-24_09:21:03-rbd-wip-jd-testing-distro-basic-smithi/

src/librbd/AioObjectRequest.cc
```diff
@@ -623,6 +624,13 @@ void AioObjectTruncate::send_write() {
   }
 }
 
+void AioObjectTrim::pre_object_map_update(uint8_t *new_state) {
+  *new_state = OBJECT_PENDING;
+
+  // object map state (before pre) decides if a post updation is needed
+  need_post = m_ictx->object_map->update_required(m_object_no, *new_state);
```
dillaman: Why is this needed? Shouldn't https://github.com/vshankar/ceph/blob/master/src/librbd/AioObjectRequest.cc#L485 cover the cause when the PRE didn't update the object map (since https://github.com/vshankar/ceph/blob/master/src/librbd/ObjectMap.cc#L98 would essentially no-op the update if the object is already non-existent), thus the POST would be no-op as well (since it's still flagged as non-existent).

vshankar: Before `send_pre()` is called (for each object), the object state for the range `[start, end)` has already been moved to `OBJECT_PENDING`. Therefore, pre operation for every object being trimmed becomes a no-op. But in `send_post()`, the object map state for every object is still in `OBJECT_PENDING` state as bulk object map update of `[start, end)` is done *after* objects in this range are trimmed (in `TrimRequest::send_post_copyup()`). This post object map update is done for each object being trimmed.

This is the reason to introduce `need_post` in `AioObjectTrim` -- post update is done only if the object state was not already in `OBJECT_PENDING` state.

dillaman: It should only be set to PENDING if the current state is EXISTS/EXISTS_CLEAN

vshankar: I think that's kind of implicit. `->update_required()` in `send_pre()` would become a no-op for other cases. `AioObjectTrim::pre_object_map_update()` does something similar and sets `need_post` accordingly. Actually, I had initially thought to introduce `AioObejctTrim::send_post` that would set `need_post` thereby avoiding redundant call to `->update_required` and avoiding calling `pre_object_map_update()` call inside `object_map_lock`.

vshankar:
@dillaman - I'm unable to get a hold of your concerns in this PR. So, let me briefly explain the reason for the changes where the concerns are:

`AioObjectTrim` needs to handle cases where `pre/post` object map updates are already done by the caller (`TrimRequest::send_{pre, post}_copyup()`) and the case where `pre/post` operations needs to be done by AioObjectTrim itself (call from `TrimRequest::send_clean_boundary()`). For the latter, we need `pre/post` to go through for `AioObjectTrim` (as object state would need to be moved from `EXIST` to `PENDING` in pre and from `PENDING` to `NONEXISTENT` in post). So, I still think the changes are required. Plus, `fsx` runs to completion successfully with these changes.

Let me know if this is completely off...

dillaman:
@vshankar I think I'd rather see ```TrimRequest::<anon>::C_CopyupObject``` either pass a "disable object map updates" flag to ```AioObjectTrim``` or create a new class that explicitly disables object map updates since it was already handled in batch. Right now, I think the current implementation isn't very clear what it's attempting to handle. The ```AioObjectTrim``` was added for one special case instead of adding lots of conditions to the original ```AioObjectRemove```. (not to mention, your ```need_post``` is named as if it is a local variable)

vshankar:
@dillaman - Right, So its more from an implementation point. I'll do the required changes.

Thanks for reviewing.

vshankar:
Teuthology jobs which failed earlier (fsx runs) have now passed: http://pulpito.ceph.com/vshankar-2016-11-03_13:12:19-rbd-wip-17356---basic-mira/

dillaman:
lgtm

PR #11511, opened 2016-10-16
mds/MDSMap.h: add const to member functions

stiopaa1:

### DIFF
--- a/src/mds/MDSMap.h
+++ b/src/mds/MDSMap.h
@@ -233,13 +233,13 @@ public:
       cached_up_features(0)
   { }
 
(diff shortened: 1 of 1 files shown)


src/mds/MDSMap.h
```diff
@@ -610,9 +612,9 @@ class MDSMap {
     }
   }
 
-  int get_inc_gid(mds_gid_t gid) {
+  int get_inc_gid(mds_gid_t gid) const {
     if (mds_info.count(gid))
```
tchaikov: could you take this chance to fix the anti-pattern of `count + []`?

stiopaa1: @tchaikov 
done

PR #11512, opened 2016-10-16
FreeBSD: update require packages

wjwithagen:
Leftover from the Cmake move,
and ceph-mgr requires boost-python

### DIFF
--- a/install-deps.sh
+++ b/install-deps.sh
@@ -23,9 +23,10 @@ if [ x`uname`x = xFreeBSDx ]; then
     $SUDO pkg install -yq \
         devel/git \
         devel/gmake \
-        devel/automake \
+        devel/cmake \
         devel/yasm \
         devel/boost-all \
+        devel/boost-python-libs \
         devel/valgrind \
         devel/pkgconf \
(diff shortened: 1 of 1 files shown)


tchaikov:
is `boost-python-libs` part of `boost-all`?

wjwithagen:
@tchaikov 
I guess not, since I had installed boost-all, and got complaints that is was missing when MGR got introduced.
And it seems packages things the same, boost-libs does not contain the python stuff.
```
% pkg info | grep boost
boost-all-1.55.0               The "meta-port" for boost libraries
boost-docs-1.55.0              Documentation for libraries from boost.org
boost-jam-1.55.0               Build tool from the boost.org
boost-libs-1.55.0_12           Free portable C++ libraries (without Boost.Python)
boost-python-libs-1.55.0       Framework for interfacing Python and C++
```
And on a fresh system boost-all installs:
```
New packages to be INSTALLED:
        boost-all: 1.55.0
        boost-libs: 1.55.0_13
        boost-jam: 1.55.0
        boost-docs: 1.55.0
```

PR #11517, opened 2016-10-17
Fix #17562 (backtrace check fails when scrubbing directory created by fsstress)

ukernel:


src/mds/inode_backtrace.cc
```diff
@@ -145,18 +145,7 @@ int inode_backtrace_t::compare(const inode_backtrace_t& other,
...
-          *divergent = true;
-        return -1;
-      } else if (ancestors[i-1].version > other.ancestors[i-1].version) {
-        if (comparator < 0)
-          *divergent = true;
-        return 1;
-      } else {
-        assert(ancestors[i-1].version == other.ancestors[i-1].version);
-        return 0;
-      }
```
gregsfortytwo: Removing this block is incorrect. You're right that this *should* be unreachable code if everything in the system is behaving, but we are trying to detect corruptions and bugs here. We need to detect it as divergent if we somehow have a newer ancestor version on the side with the older parent.

src/mds/CInode.cc
```diff
@@ -3840,15 +3840,11 @@ void CInode::validate_disk_state(CInode::validated_data *results,
...
+	// we're divergent, or on-disk version is newer
+	results->backtrace.error_str << "On-disk backtrace is divergent or newer";
       } else {
-        if (divergent || memory_newer <= 0) {
-          // we're divergent, or don't have a newer version to write
-          results->backtrace.error_str <<
-              "On-disk backtrace is divergent or newer";
-	  goto next;
-        }
+        results->backtrace.passed = true;
```
gregsfortytwo: This part looks like it's probably correct to me.

src/mds/inode_backtrace.cc
ukernel: Nothing do with corruption.

The code first set 'comparator' according comparison between ancestors[i].version and other.ancestors[i].version. Then it increases variable 'i' and execute above code block. It's impossible that the '*divergent = true' line get reached.

gregsfortytwo:
Are these checks backwards, then? :/

>

ukernel:
Yes, they are backwards. that's why the check can not be true.

src/mds/inode_backtrace.cc
gregsfortytwo: No, I just checked. We set comparator >0 if we are newer based on the immediate parent. That means if the other guy at any point has a newer ancestor than we do, something has gone terribly wrong and we set the divergent flag. We don't update the comparator on each loop though and this is not dead code!

src/mds/inode_backtrace.cc
```diff
@@ -140,7 +141,8 @@ int inode_backtrace_t::compare(const inode_backtrace_t& other,
        */
       break;
     }
-    if (ancestors[i].dirino != other.ancestors[i].dirino) {
+    if (ancestors[i].dirino != other.ancestors[i].dirino ||
+        ancestors[i].dname != other.ancestors[i].dname) {
```
gregsfortytwo: Okay, but do we really care about that when this is in use? We already update backtrackes on a local rename and if we don't have the object we're reconstructing the fragments anyway. What do we get by not considering them equivalent when the directory inodes still match, besides potentially extra IO?

ukernel:
On Wed, Oct 19, 2016 at 1:13 PM, Gregory Farnum <notifications@github.com>
wrote:

> *@gregsfortytwo* commented on this pull request.
> ------------------------------
>
> In src/mds/inode_backtrace.cc <https://github.com/ceph/ceph/pull/11517>:
>
> > @@ -145,18 +145,7 @@ int inode_backtrace_t::compare(const inode_backtrace_t& other,
>      if (ancestors[i].dirino != other.ancestors[i].dirino ||
>          ancestors[i].dname != other.ancestors[i].dname) {
>        *equivalent = false;
> -      if (ancestors[i-1].version < other.ancestors[i-1].version) {
> -        if (comparator > 0)
> -          *divergent = true;
> -        return -1;
> -      } else if (ancestors[i-1].version > other.ancestors[i-1].version) {
> -        if (comparator < 0)
> -          *divergent = true;
> -        return 1;
> -      } else {
> -        assert(ancestors[i-1].version == other.ancestors[i-1].version);
> -        return 0;
> -      }
>
> No, I just checked. We set comparator >0 if we are newer based on the
> immediate parent. That means if the other guy at any point has a newer
> ancestor than we do, something has gone terribly wrong and we set the
> divergent flag. We don't update the comparator on each loop though and this
> is not dead code!
>

which code does the job "if the other guy at any point has a newer ancestor
than we do" ? Above code use 'i-1' as index, which is immediate parent.

jcsp:
(This is passing tests at http://pulpito.ceph.com/jspray-2016-10-17_21:55:28-fs-wip-jcsp-testing-20161017-distro-basic-mira , http://pulpito.ceph.com/jspray-2016-10-18_15:21:25-fs-wip-jcsp-testing-20161017---basic-mira , but I'm leaving it open because of ongoing discussion)

gregsfortytwo:
@jcsp said he would look at this too.

src/mds/CInode.cc
gregsfortytwo: Well, we can use the equivalent part to do other stuff like electing to force out a backtrace if the on-disk one is stale, even if they're still consistent.

src/mds/inode_backtrace.cc
gregsfortytwo: Sounds good to me then.

src/mds/inode_backtrace.cc
gregsfortytwo: ...okay, I'm just totally blind. Wasn't seeing the comparator adjustments farther down for some reason.

gregsfortytwo:
I think I'm good with this now. Thanks for double-checking @jcsp. Won't merge because of your outstanding questions.

jcsp:
OK, I now think that setting divergent=true (instead of equivalent=false) was correct to begin with in the case of ancestors[0] not matching.  I think we could still see some false positives here if someone scrubs unflushed metadata, but that's kind of a separate issue.

PR #11519, opened 2016-10-17
jewel: RGW loses realm/period/zonegroup/zone data: period overwritten if somewhere in the cluster is still running Hammer

dachary:
http://tracker.ceph.com/issues/17576


dachary:
@cbodley does this backport look good to merge ? It passed the rgw suite http://tracker.ceph.com/issues/17487#note-18

PR #11525, opened 2016-10-18
tools/ceph-monstore-update-crush.sh: FreeBSD getopt is not compatible…

wjwithagen:
…, use the one from packages


wjwithagen:
@tchaikov 
I have installed a jenkins builder, and from that a few things follow that are required to getting the environment more compatible. Hence all the small commit of tools, bits and pieces.

PR #11526, opened 2016-10-18
client: get caller's uid/gid on every libcephfs operation

ukernel:
Fixes: http://tracker.ceph.com/issues/17591

### DIFF
--- a/src/client/UserPerm.h
+++ b/src/client/UserPerm.h
@@ -65,8 +65,8 @@ public:
     return *this;
   }
 
-  uid_t uid() const { return m_uid; }
-  gid_t gid() const { return m_gid; }
+  uid_t uid() const { return m_uid != (uid_t)-1 ? m_uid : ::geteuid(); }
+  gid_t gid() const { return m_gid != (gid_t)-1 ? m_gid : ::getegid(); }
   bool gid_in_groups(gid_t gid) const {
     if (gid == m_gid) return true;
     for (int i = 0; i < gid_count; ++i) {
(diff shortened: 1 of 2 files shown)


jcsp:
Looks sane to me, @gregsfortytwo please could you review

gregsfortytwo:
I was really trying to move away from this model, but I guess until we have @jtlayton's new libcephfs and bindings done we need our tests to pass. :/ Did we run this through the samba test suite?

src/client/UserPerm.h
```diff
@@ -65,8 +65,8 @@ struct UserPerm
     return *this;
   }
 
-  uid_t uid() const { return m_uid; }
-  gid_t gid() const { return m_gid; }
+  uid_t uid() const { return m_uid != (uid_t)-1 ? m_uid : ::geteuid(); }
+  gid_t gid() const { return m_gid != (gid_t)-1 ? m_gid : ::getegid(); }
```
jtlayton: For the record, we should really be using the fsuid and fsgid here (at least on Linux) and not the euid/egid. In most cases, they are identical but there are cases where they can be different. Unfortunately, Linux has no getfsuid() syscall -- you have to issue setfsuid(-1) which will always fail, but will return the existing fsuid.

What we should probably do is roll our own getfsuid() routine that does that on Linux and just calls out to geteuid() on other platforms. That said, we should probably do that in the context of a different PR.

jtlayton: Hmm...maybe I'm wrong there actually. The setfsuid manpage says "don't use it", so maybe geteuid is fine. I think you can disregard my comment...

ukernel:
http://pulpito.ceph.com/zyan-2016-10-20_07:49:59-samba-wip-zyan-testing---basic-smithi/

PR #11527, opened 2016-10-18
test/encoding/readable.sh: fix shell script warning

wjwithagen:
shell complains about:
`continue: only meaningful in a`for', `while', or`until' loop`

### DIFF
--- a/src/test/encoding/readable.sh
+++ b/src/test/encoding/readable.sh
@@ -69,7 +69,7 @@ test_object() {
       if [ -n "$incompat" ]; then
         if [ -z "$incompat_paths" ]; then
           echo "skipping incompat $type version $arversion, changed at $incompat < code $myversion"
-          continue
+	  return
         else
           # If we are ignoring not whole type, but objects that are in $incompat_path,
           # we don't skip here, just give info.


liewegas:
In that case the loop needs to be restructured?

wjwithagen:
@liewegas 
I have the feeling that it is just a "left-over" from cut&paste.
But I'll try and see if I can figure it out.

wjwithagen:
@liewegas 
It is weird indeed. Perhaps this would explain possible errors on my jenkins VM as well.
My dev box seems to work without much trouble.

wjwithagen:
@liewegas 
Nope, code does need to be restructured, since there is no outside loop to reference , and this test really does want to skip. At least that is what the comment and code is hinting. 
Will make a new patch

src/test/encoding/readable.sh
```diff
@@ -69,7 +69,6 @@ test_object() {
       if [ -n "$incompat" ]; then
         if [ -z "$incompat_paths" ]; then
           echo "skipping incompat $type version $arversion, changed at $incompat < code $myversion"
-          continue
```
tchaikov: should be `return`, i guess.

wjwithagen:
@liewegas @tchaikov 
Yup, I replaced the continue with return before I went to my meeting and that helps.
No more complaints about language, and my FreeBSD tests do not abort.
I'll commit a fix.

tchaikov:
lgtm.

PR #11528, opened 2016-10-18
vstart: decrease pool size if <3 OSDs

jcsp:
Makes it easy to do OSD=1 for e.g. doing cephfs
stuff where we don't care much about the OSDs
behaviour.


theanalyst:
useful to have in case of other components also which doesn't depend on OSD behaviour

PR #11530, opened 2016-10-18
common/config_opts.h: Optimized RocksDB WAL settings.

markhpc:
Increasing the total amount of WAL space improves performance of 4K randwrites on NVMe configurations.  Spreading the logs over many files does not particularly make sense.  In our test setup:

```
buffer count     buffer size    aggregate    4K randwrite (MB/s)
16               4m             64m          105.73
4                16m            64m          134.906
2                32m            64m          196.793
16               16m            256m         227.959
4                64m            256m         221.089
2                128m           256m         234.898
4                128m           512m         256.793
4                256m           1024m        265.873
4                512m           2048m        262.495
4                1024m          4096m        270.052
```

Performance increases are notable up to 4 256MB logs.  Small gains may be made by increasing the aggregate log size considerably beyond that, however the gains appear to be minimal.

Signed-off-by: Mark Nelson mnelson@redhat.com

### DIFF
--- a/src/common/config_opts.h
+++ b/src/common/config_opts.h
@@ -1002,7 +1002,7 @@ OPTION(bluestore_freelist_type, OPT_STR, "bitmap") // extent | bitmap
 OPTION(bluestore_freelist_blocks_per_key, OPT_INT, 128)
 OPTION(bluestore_bitmapallocator_blocks_per_zone, OPT_INT, 1024) // must be power of 2 aligned, e.g., 512, 1024, 2048...
 OPTION(bluestore_bitmapallocator_span_size, OPT_INT, 1024) // must be power of 2 aligned, e.g., 512, 1024, 2048...
-OPTION(bluestore_rocksdb_options, OPT_STR, "compression=kNoCompression,max_write_buffer_number=16,min_write_buffer_number_to_merge=3,recycle_log_file_num=16")
+OPTION(bluestore_rocksdb_options, OPT_STR, "compression=kNoCompression,max_write_buffer_number=4,min_write_buffer_number_to_merge=1,recycle_log_file_num=4,write_buffer_size=268435456")
 OPTION(bluestore_fsck_on_mount, OPT_BOOL, false)
 OPTION(bluestore_fsck_on_umount, OPT_BOOL, false)
 OPTION(bluestore_fsck_on_mkfs, OPT_BOOL, true)


yuyuyu101:
...I think it may make sense for nvme ssd. But for hdd, I think 256MB is too huge. Like hdd, it may totally make full use of hdd bandwidth.

markhpc:
Hi Haomai, for HDD you may be right, though there does seem to be some correlation with larger WAL size and higher throughput.  I suspect it's a minor increase due to slightly better write coalescing.  Since the min_alloc size is higher by default with HDD, there is much less metadata per IO to write so I suspect the majority of the log write throughput will not be from metadata.

HDD Only:

`log count    log size    aggregate    4K randwrite (MB/s)
16           4m          64m          6.2266
4            16m         64m          7.04785
2            32m         64m          7.29004
16           16m         256m         6.75098
4            64m         256m         7.78516
2            128m        256m         9.24414`

HDD With NVMe WAL/DB:

`log count    log size    aggregate    4K randwrite (MB/s)
16           4m          64m          8.2739
4            16m         64m          8.03418
2            32m         64m          9.1875
16           16m         256m         9.24414
4            64m         256m         8.62305
2            128m        256m         8.91113`

yuyuyu101:
I guess 256mb buffer size may cause io shaking hugely when flushing write buffer?

liewegas:
@markhpc can you clarify what "buffer count" is?  Is that max_write_buffer_number  or min_write_buffer_number_to_merge?

liewegas:
@yuyuyu101 The write buffer is flushed out each time there is a synchronous commit, so it'll never be a huge 256MB write to the write buffer.  But... when it gets compacted for the level0 SST, it'll turn into something between 256MB and 0MB, depending on how much of the data is obsolete (e.g., temp wal keys or dup onode updates).  I agree that 256MB is way too much, even for an upper bound.

@markhpc I don't understand what buffer count is or why you're varying it.  The results don't make sense to me.. if it's the max buffers it should have minimal effect as long as it's not tiny, and if its the number_to_merge it should be roughly identical to the same buffer size * that value, but it's not (and I don't see a value of 1).  Which makes me think it's the max, in which case the results are just noisy?

markhpc:
sage: that's max_write_buffer_number.  The only results with min_write_buffer_number_to_merge are the ones with max_write_buffer_number = 16 (ie the current default).

Haomai:  what we've been seeing is sort of the opposite.  It appears that during level 0 compaction you can run out of WAL buffer space and writes stall.  It may be that if we tune the compactions to happen frequently we could mitigate some of that.  The fear with larger WAL sizes I think is that there are more entries to sort through, but so far we do not appear to be at sizes where this is a problem.

liewegas:
Ah ok.  So I can see that for small buffer sizes then max_buffers might matter.. but once it's largish (64mb?) it seems crazy that it would take more than 1/4 second on nvme to write it out (which is what you'd need to sustain ~200 MB/sec.

markhpc:
sage: for berevity I included shorter labels, but here's the spreadsheet with the real labels:


[rocksdb_wal_tuning.xlsx](https://github.com/ceph/ceph/files/536465/rocksdb_wal_tuning.xlsx)

markhpc:
sage: I don't think the WAL flushing speed is probably what matters here.  I think we are avoiding stalls during level0 compaction with big buffers.

liewegas:
What I'm curious about is what happens on NVMe with smallish buffer sizes (say, 32M) but a huge max (let's say 128 just to take any throttling off the table).  How does the 4k rw throughput vary then when it's *just* the buffer size and subsequent l0 sst size that is being varied?

somnathr:
@markhpc do you have any data min_write_buffer_number_to_merge 1 vs 2 ? The updates will never be merged if we go with 1. 
Hope, you ran the tests for at least 60 min or so as with small run compaction will not start and for very small run all will be served from memory. Bigger buffer size will help you on that.
we need to see how the following compaction settings affects performance as well specially with min_alloc_size=4k

compaction_threads
flusher_threads
max_background_compactions
max_background_flushes
max_bytes_for_level_base
level0_file_num_compaction_trigger


Regarding recycle log number, during fixing the bug on that path in rocksdb , I saw lot of files are reused over time and not going for reuse will only increase extra create/delete. Again, small run will not trigger this reuse and we will not be knowing the impact. What is the difference you are seeing with 4 vs 16 ?

markhpc:
somnath: one point sage made is that for RGW workloads, having 4 large SST files in level 0 that need to be compacted might be worse than 16 small SST files.  I'm concerned though that we might start throttling if we write too many files into level 0 (ie say we have max_write_buffer_number set to 16+ with 32MB buffers).  I'm going to try to test out some of this today.

somnathr:
You can control level0 compaction with the following

level0_file_num_compaction_trigger

But, generally I saw the earliest you start compaction is better because that will smooth the performance..

somnathr:
Also, the following control is there for stall/slowing down writes..
level0_slowdown_writes_trigger
level0_stop_writes_trigger

I didn't find this much helpful though if compaction is *not* starting early..

markhpc:
Today I reran some tests with max_write_buffer_number = [4, 32] and write_buffer_size = [256MB, 32MB] respectively (write_buffer_number_to_merge was kept at 1 in each case).  In these tests 8 32GB partitions were prefilled and then overwritten with 4k random writes (until all volumes were completely overwritten once).  I threw together a parser that extracts the compaction statistics from the RocksDB compaction finish event in osd log files and dumps them into a tab delimited output along with some general statistics across multiple input files.

ie:

`ceph_rocksdb_log_parser.py -l1 16k_32_32/osd.0.log 4k_32_32/osd.0.log 16k_4_256/osd.0.log 4k_4_256/osd.0.log`

A rough first pass at the tool can be found in the cbt tools directory here:

https://github.com/ceph/cbt/blob/master/tools/ceph_rocksdb_log_parser.py

And a spreadsheet made with the output is attached.

The tests with larger buffers with lower max_write_buffer_number were significantly faster, likely due to the low write_buffer_number_to_merge ratio when max_write_buffer_number is set high.  In these cases, significantly more data appears to be written to level 1 and more total time is spent in compaction despite having smaller compaction windows.  I don't really know why this is the case yet however.
[wal_compaction_stats.xlsx](https://github.com/ceph/ceph/files/537947/wal_compaction_stats.xlsx)

markhpc:
In the logs, we are regularly seeing write stalls with the associated message in the logs:

"rocksdb: [default] Stalling writes because we have 20 level-0 files rate 2097152"

This indicates that compaction is not keeping up with writes.  There are three primary recommendations from the rocksdb folks for dealing with this:

> Increase max_background_compactions to have more compaction threads.
> Increase write_buffer_size to have large memtable, to reduce write amplification.
> Increase min_write_buffer_number_to_merge.

https://github.com/facebook/rocksdb/wiki/Write-Stalls

Increasing the max_background_compactions will help with l1->l2 compactions, but not level0->level1:

> Compactions at levels 0 and 1 are tricky. Files at level 0 usually span the entire key space. When compacting L0->L1 (from level 0 to level 1), compaction includes all files from level 1. With all files from L1 getting compacted with L0, compaction L1->L2 cannot proceed; it must wait for the L0->L1 compaction to finish. If L0->L1 compaction is slow, it will be the only compaction running in the system most of the time, since other compactions must wait for it to finish.
> 
> L0->L1 compaction is also single-threaded. It is hard to achieve good throughput with single-threaded compaction. To see if this is causing issues, check disk utilization. If disk is not fully utilized, there might be an issue with compaction configuration. We usually reommend making L0->L1 as fast as possible by making the size of level 0 similar to size of level 1.

https://github.com/facebook/rocksdb/wiki/RocksDB-Tuning-Guide

So ultimately it appears to me that short of changing the amount of data and/or we we inject data into rocksdb, primarily the options are to make sure that L0 and L1 are the same size and increase the buffer size to reduce write amplification.  Next up will be to test small new rados writes with different buffer sizes to see how bad large buffers are in this case.

markhpc:
I ran additional tests looking at compaction stats and client performance for rados 4k object creation/writes.  The overall trends remained roughly the same.  Performance increased with larger buffers.  These tests ran for a shorter period of time as the NVMe drives quickly filled to capacity and rewrites were not performed.

Compaction statistics were gathered for all of the tests.  A sample of them in comparison with the RBD tests is attached.  Primarily, it appears in both the RBD and in the rados bench case, write amplification of metadata in rocksdb is the primary variable affecting performance in governed almost entirely by how much data can be written to the WAL before a flush (which in turn is governed by several of the individual settings being tested).
[wal_compaction_stats_rbd_rados.xlsx](https://github.com/ceph/ceph/files/550713/wal_compaction_stats_rbd_rados.xlsx)

markhpc:
I updated the spreadsheet to now include 4k rados object creation/write tests, 256B rados OMAP write tests, filestore tests, and also tested using universal compaction.  Filestore performance appears to have improved in RBD since jewel (still need to verify this), likely due to the pg info fastinfo work.  We still beat it with large WAL buffer space though.  Bluestore performance is exceptionally faster, especially with large buffers, when writing many small objects or OMAP entries.
[20161025_wal_compaction_stats_rbd_rados_omap.xlsx](https://github.com/ceph/ceph/files/558600/20161025_wal_compaction_stats_rbd_rados_omap.xlsx)

PR #11532, opened 2016-10-18
pybind: avoid "exception 'int' object is not iterable"

javeme:
Run a command: ceph tell osd.0 pg 1.0 list_missing
If it's interrupted by CTRL+C, it would output a strange error message:
"exception 'int' object is not iterable".

The reason is the caller expects run_in_thread() to return a tuple, but
it returns an int. this patch we fix the bug, then it would just output
a message like: "Error EINTR: Interrupted!"
```
Detailed error message:
  File "./bin/ceph", line 949, in <module>
    retval = main()
  File "./bin/ceph", line 882, in main
    sigdict, inbuf, verbose)
  File "./bin/ceph", line 480, in new_style_command
    inbuf=inbuf)
  File "ceph/src/pybind/ceph_argparse.py", line 1334, in
json_command
    raise RuntimeError('"{0}": exception {1}'.format(argdict, e))
RuntimeError: "{'prefix': u'pg', u'cmd': u'list_missing', u'pgid':
u'1.0'}": exception "['{"prefix": "pg", "cmd": "list_missing", "pgid":
"1.0"}']": exception 'int' object is not iterable
```
Signed-off-by: Javeme <javaloveme@gmail.com>

### DIFF
--- a/src/pybind/ceph_argparse.py
+++ b/src/pybind/ceph_argparse.py
@@ -1191,7 +1191,7 @@ def run_in_thread(target, *args, **kwargs):
         interrupt = True
 
     if interrupt:
-        t.retval = -errno.EINTR
+        t.retval = -errno.EINTR, None, 'Interrupted!'
     if t.exception:
         raise t.exception
     return t.retval


javeme:
The failure of this test case "run-rbd-unit-tests" may be caused by [issue #17566](http://tracker.ceph.com/issues/17566)

tchaikov:
you commit title reads:

```
fix bug: exception 'int' object is not iterable
```

could you prefix it with the subcomponent your are changing ? see https://github.com/ceph/ceph/blob/master/SubmittingPatches.rst#preparing-and-sending-patches, in this case, it could be "pybind: " 

and it would be ideal if you can be more concise when describing the change, i'd recommend something like:
```
pybind: return a tuple in run_in_thread() if interrupted

<describe the change in more detail>
```

javeme:
@tchaikov Thank you! I have updated the commit message.

tchaikov:
lgtm.

tchaikov:
tested at http://pulpito.ceph.com/kchai-2016-10-24_03:13:40-rados-wip-kefu-testing---basic-mira/

PR #11535, opened 2016-10-18
[DNM] rgw: continuation of the auth rework

rzarzynski:


dachary:

src/rgw/rgw_keystone.cc
```diff
@@ -140,68 +142,82 @@ bool rgw_decode_pki_token(CephContext * const cct,
...
-  url = cct->_conf->rgw_keystone_url;
-  if (url.empty()) {
-    ldout(cct, 0) << "ERROR: keystone url is not configured" << dendl;
-    return -EINVAL;
-  }
+  static std::string url = g_ceph_context->_conf->rgw_keystone_url;
 
-  if (url[url.size() - 1] != '/') {
-    url.append("/");
+  if (url.empty() || boost::algorithm::ends_with(url, "/")) {
```
mattbenjamin: *ah*

src/rgw/rgw_auth.h
```diff
@@ -348,4 +350,164 @@ class RGWAnonymousAuthEngine : public RGWTokenBasedAuthEngine {
...
+ *
+ * The authentication process consists two steps:
+ *  - Engine::authenticate() supposed to be called before *initiating*
+ *    any modifications to RADOS store that are related to an operation
+ *    a client wants to perform (RGWOp::execute).
+ *  - Completer::complete() supposed to be called, if completer has been
+ *    returned, after the authenticate() step but before *committing*
+ *    those modifications or sending a response (RGWOp::complete).
+ *
+ * The lifetime of an engine outperforms both Applier and Completer. It's
```
rzarzynski: Definitely. Thanks!

src/rgw/rgw_auth.cc
```diff
@@ -508,3 +508,44 @@ RGWAuthApplier::aplptr_t RGWKeystoneAuthEngine::authenticate() const
...
 }
+
+rgw::auth::Engine::result_t
+rgw::auth::Strategy::authenticate(const req_state* const s) const
+{
+  for (const stack_item_t& kv : auth_stack) {
+    const rgw::auth::Engine& engine = kv.first;
+    const auto& policy = kv.second;
+
+    auto&& res = engine.authenticate(s);
```
rzarzynski: I'm trying to figure this out and it looks that making confusion is the sole outcome from requesting the R-value reference here. :-( Undoubtedly needs to be reworked. Thanks for pointing this out.

rzarzynski:
@mattbenjamin: I would prefer to avoid linking those cache *instances* due to potential collisions. Anyway, making some common infrastructure looks like an idea worth research.

src/rgw/rgw_auth.cc
cbodley: i think `auto&&` makes perfect sense here. because we're returning this result by value, the caller will already require a move construction to store the result. changing this to `auto res =` would incur an extra call to the move constructor

p.s. the `return std::move(res);` below should be equivalent to `return res;` in either case

rzarzynski: > changing this to `auto res =` would incur an extra call to the move constructor

IIRC the standard explicitly allows (but not enforces) compilers to apply NVRO in such situation. So, depending on compiler, we can obtain the benefit without introducing the visual clutter of `auto&&`.

rzarzynski:
Merged in #12893.

PR #11537, opened 2016-10-18
os/bluestore: refactor bluestore_sync_submit_transaction

liewegas:


liewegas:
unit test failure is fixed in latest master

src/os/bluestore/BlueStore.cc
```diff
@@ -6164,10 +6168,17 @@ void BlueStore::_txc_state_proc(TransContext *txc)
...
+	  dout(20) << __func__
+		   << " last_{nid,blobid} exceeds max, submit via kv thread"
+		   << dendl;
+	} else {
+	  _txc_finalize_kv(txc, txc->t);
+	  txc->kv_submitted = true;
+	  int r = db->submit_transaction(txc->t);
+	  assert(r == 0);
+	}
       }
```
majianpeng: @liewegas .  Consider this:
txc1/txc2 last_nid both are larger nid_max.
  thread1                           thread2
 update nid_max
                                        submit_transaction
 submit_transaction but failed.
When replay check, txc2->last_nid > nid_max.
I think we in kv_sync_thread, we must make sure the new nid_max must update and then update nid_max.

src/os/bluestore/FreelistManager.h
```diff
@@ -38,6 +38,10 @@ class FreelistManager {
   virtual void release(
     uint64_t offset, uint64_t length,
     KeyValueDB::Transaction txn) = 0;
+
+  virtual bool supports_parallel_transactions() {
+    return false;
+  }
```
majianpeng: @liewegas . I don't know for extentfreelistmanager  why don't support parallel transaction? Hope your explain. Thanks!

liewegas:
Oh sorry, forgot to answer this.

The extent manager stores the freelist as offset=length k/v pairs. When 
you take something off the freelist, it usually means deleing an old pair 
and reinserting a new (smaller) one.  This works as long as these updates 
are ordered.  But if parallel racing threads and doing these updates, the 
updates won't make any sense.  E.g.,

start with 0=100
txn1
	rm 0=100
	set 1=99

txn2
	rm 1-99
	set 2=88

If they commit in order all is well.  If the order is reversed the 
freelist is corrupted.

This was one of the motivations for the bitmap implementation, which uses 
a commutative merge operator (xor).

majianpeng:
@liewegas . Thanks your detail explain. Even if it don't use parallel submit transaction, we also don't make tx1 must firstly submit to kv_queue.Is it so?

varadakari:
LGTM

varadakari:
@majianpeng after aio _txc_finish_io() will make sure we are submitting the transaction in sequence.

src/os/bluestore/BlueStore.cc
```diff
@@ -6165,21 +6169,40 @@ void BlueStore::_txc_state_proc(TransContext *txc)
...
-	int r = db->submit_transaction_sync(txc->t);
-	assert(r == 0);
       }
       {
 	std::lock_guard<std::mutex> l(kv_lock);
 	kv_queue.push_back(txc);
 	kv_cond.notify_one();
+	if (!txc->kv_submitted) {
+	  kv_queue_unsubmitted.push_back(txc);
+	  ++txc->osr->kv_committing_serially;
```
somnathr: @liewegas Sorry missed the check , gave my comment above.

src/os/bluestore/BlueStore.cc
```diff
@@ -6165,21 +6169,40 @@ void BlueStore::_txc_state_proc(TransContext *txc)
...
-      if (!g_conf->bluestore_sync_transaction) {
-	if (g_conf->bluestore_sync_submit_transaction) {
+      if (g_conf->bluestore_sync_submit_transaction &&
+	  fm->supports_parallel_transactions()) {
+	if (txc->last_nid >= nid_max ||
+	    txc->last_blobid >= blobid_max) {
+	  dout(20) << __func__
+		   << " last_{nid,blobid} exceeds max, submit via kv thread"
+		   << dendl;
+	} else if (txc->osr->kv_committing_serially) {
```
somnathr: @liewegas I think this may hurt the parallelism, may be we should guard against a number rather than even 1 ?
I will run through my test and see how frequent we are hitting this.

liewegas:
We do, because in the non-sync mode, the _txc_finalize_kv is called from 
kv_sync_thread in kv_committing order, and they are submitted to rocksdb 
in that same order.

liewegas:
we should only fall into this case for low values of _prealloc_* tunables.  
I set them low in store_test.cc to exercise this path but in the real 
world they will be quite bit (1024 or something).

PR #11538, opened 2016-10-18
os/bluestore: No garbage collection for uncompressed blobs

Roushan81:
This is a bug fix.  As part of the change, we skip garbage collection of uncompressed blobs. Also, if overlapping blob is compressed, then blob_depth for newly created blob is set to minimum(1).


liewegas:
https://github.com/ceph/ceph/pull/11539 (rebased)

PR #11542, opened 2016-10-19
systemd: add ceph-fuse service file

batrick:
Signed-off-by: Patrick Donnelly pdonnell@redhat.com

### DIFF
--- a/ceph.spec.in
+++ b/ceph.spec.in
@@ -1156,6 +1156,8 @@ fi
 %{_bindir}/ceph-fuse
 %{_mandir}/man8/ceph-fuse.8*
 %{_sbindir}/mount.fuse.ceph
+%{_unitdir}/ceph-fuse@.service
+%{_unitdir}/ceph-fuse.target
 
 %files -n rbd-fuse
 %defattr(-,root,root,-)
--- a/debian/rules
+++ b/debian/rules
@@ -96,6 +96,11 @@ override_dh_installinit:
 	sed -i s./etc/sysconfig/./etc/default/.g debian/ceph-mds/lib/systemd/system/ceph-mds@.service
 	install -m0644 systemd/ceph-mds.target debian/ceph-mds/lib/systemd/system
 
+	install -d -m0755 debian/ceph-fuse/lib/systemd/system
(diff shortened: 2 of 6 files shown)


jcsp:
Hmm.  What's the use case for this vs. having something in fstab?

batrick:
I'm not really sure if fstab is a better way to do this. My
understanding is that mount units are considered the way forward and
they do provide more configuration options.

We might want to change this `ceph-fuse@.service` unit to a `.mount`
unit but the naming requirements are unfortunately strict (the mount
unit name **must** match the mount point name... but we don't know
where users will want to mount ceph-fuse).

batrick:
I had to make small modification to the unit as the namespace modification directives like "PrivateDevices" cause the mounted file system to be unavailable to the rest of the system. That's fixed now and I've tested the unit file with my Linode tests without any issues.

jcsp:
So I was googling about this topic and saw:

> Mount units may either be configured via unit files, or via /etc/fstab (see fstab(5) for details). Mounts listed in /etc/fstab will be converted into native units dynamically at boot and when the configuration of the system manager is reloaded. **In general, configuring mount points through /etc/fstab is the preferred approach**. See systemd-fstab-generator(8) for details about the conversion.
(https://www.freedesktop.org/software/systemd/man/systemd.mount.html)

However I can see the utility of having the ceph-fuse service file as it lets us provide the "network-online.target local-fs.target time-sync.target" requirements to avoid users having weird issues when they forget "netdev" etc.

I guess I have no problem merging this, can you add something to the docs in this PR to explain how to use it (could just add to fstab.rst)?  I'm guessing one is supposed to create a systemd unit with the mount point in the name?

jcsp:
@batrick this is good from my POV once it has docs

batrick:
Sorry this fell off my radar. I'll finish this up soon.

batrick:
@jcsp, I've added the docs and did some testing to make sure it works properly via ceph-ansible. I think this is ready to go.

jcsp:
(I'm waiting until Kraken branches to merge this to master)

PR #11543, opened 2016-10-19
mds: use unique_ptr to simplify resource mgmt

batrick:
[FYI, this was originally part of a large branch I've been sitting on for a few months. Unfortunately that branch grew too large making it difficult to go through compile errors. So, I'm breaking it up over time into smaller PRs.]

### DIFF
--- a/src/mds/CDir.cc
+++ b/src/mds/CDir.cc
@@ -183,7 +183,6 @@ CDir::CDir(CInode *in, frag_t fg, MDCache *mdcache, bool auth) :
   first(2),
   dirty_rstat_inodes(member_offset(CInode, dirty_rstat_item)),
   projected_version(0),  item_dirty(this), item_new(this),
-  scrub_infop(NULL),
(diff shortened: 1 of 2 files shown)


src/mds/CDir.cc
```diff
@@ -183,7 +183,6 @@ CDir::CDir(CInode *in, frag_t fg, MDCache *mdcache, bool auth) :
   first(2),
   dirty_rstat_inodes(member_offset(CInode, dirty_rstat_item)),
   projected_version(0),  item_dirty(this), item_new(this),
-  scrub_infop(NULL),
   num_head_items(0), num_head_null(0),
   num_snap_items(0), num_snap_null(0),
   num_dirty(0), committing_version(0), committed_version(0),
```
renhwztetecs: bloom(null) is redundant

jcsp:
:+1:

batrick:
@jcsp, I've made a few corrections due to build errors. I think this is ready for testing.

PR #11544, opened 2016-10-19
librbd: group snapshots

VictorDenisov:


VictorDenisov:
@dillaman This PR lives on top of yet unmerged snapshot namespaces PR. It doesn't have tests yet and snapshot remove operation. Could you please take a look and say if you like it in principle. Specifically this part: https://github.com/ceph/ceph/pull/11544/files#diff-d2dae46d02584254961fdcc4318289caR411
Once snapshot namespaces PR is merged I'll rebase this PR on top of master.

VictorDenisov:
Now the PR has create, remove, list operations. It's been rebased on top of master. It still lacks tests. I'm working on the tests.

src/cls/rbd/cls_rbd.cc
```diff
@@ -4797,6 +4803,246 @@ int image_get_group(cls_method_context_t hctx,
...
+/**
+ * Save initial snapshot record.
+ *
+ * Input:
+ * @param GroupSnapshot
+ *
+ * Output:
+ * @return 0 on success, negative error code on failure
+ */
+int group_snap_save(cls_method_context_t hctx,
```
dillaman: Minor: probably better named as ```group_snap_add```

src/cls/rbd/cls_rbd.cc
```diff
@@ -4797,6 +4803,246 @@ int image_get_group(cls_method_context_t hctx,
...
+  CLS_LOG(20, "group_snap_update");
+  cls::rbd::GroupSnapshot gs;
+  try {
+    bufferlist::iterator iter = in->begin();
+    ::decode(gs, iter);
+  } catch (const buffer::error &err) {
+    return -EINVAL;
+  }
+
+  std::string key = gs.snap_key();
```
dillaman: Probably shouldn't allow all fields to be updated (e.g. the uuid shouldn't be updatable).

src/cls/rbd/cls_rbd.cc
```diff
@@ -4797,6 +4803,246 @@ int image_get_group(cls_method_context_t hctx,
...
+  int r = cls_cxx_map_set_val(hctx, key, &obl);
+
+  return r;
+}
+
+/**
+ * Retrieve snapshot record.
+ *
+ * Input:
+ * @param name Snapshot name
```
dillaman: Probably should be the uuid since you should be able to rename group snaps.

src/cls/rbd/cls_rbd.cc
```diff
@@ -4797,6 +4803,246 @@ int image_get_group(cls_method_context_t hctx,
...
+  std::string name;
+  try {
+    bufferlist::iterator iter = in->begin();
+    ::decode(name, iter);
+  } catch (const buffer::error &err) {
+    return -EINVAL;
+  }
+
+  cls::rbd::GroupSnapshot gs;
+
```
dillaman: TODO?

src/cls/rbd/cls_rbd.cc
```diff
@@ -4797,6 +4803,246 @@ int image_get_group(cls_method_context_t hctx,
...
+  return 0;
+}
+
+/**
+ * Remove snapshot record.
+ *
+ * Input:
+ * @param id Snapshot id
```
dillaman: Use the uuid

src/cls/rbd/cls_rbd.cc
```diff
@@ -4797,6 +4803,246 @@ int image_get_group(cls_method_context_t hctx,
...
+
+  int max_read = RBD_MAX_KEYS_READ;
+  std::map<string, bufferlist> vals;
+  string last_read = start_after.snap_key();
+  std::vector<cls::rbd::GroupSnapshot> res;
+  int keys_read;
+
+  do {
+    keys_read = cls_cxx_map_get_vals(hctx, last_read,
+				     RBD_SNAP_KEY_PREFIX, // TODO  replace with prefix from start after
```
dillaman: TODO

src/cls/rbd/cls_rbd.cc
```diff
@@ -4797,6 +4803,246 @@ int image_get_group(cls_method_context_t hctx,
...
+				     max_read, &vals);
```
dillaman: Minor: max_read should take into account the number of keys already in ```res``` and ```max_return```

src/cls/rbd/cls_rbd.cc
```diff
@@ -4797,6 +4803,246 @@ int image_get_group(cls_method_context_t hctx,
...
+    ::decode(start_after, iter);
+    ::decode(max_return, iter);
+  } catch (const buffer::error &err) {
+    return -EINVAL;
+  }
+
+  int max_read = RBD_MAX_KEYS_READ;
+  std::map<string, bufferlist> vals;
+  string last_read = start_after.snap_key();
+  std::vector<cls::rbd::GroupSnapshot> res;
```
dillaman: Minor: ```res``` -> ```group_snapshots```

src/cls/rbd/cls_rbd_client.cc
```diff
@@ -1770,5 +1770,96 @@ namespace librbd {
...
+      bufferlist in, out;
+      ::encode(snap_id, in);
+
+      return ioctx->exec(oid, "rbd", "group_snap_remove", in, out);
+    }
+
+    int group_snap_list(librados::IoCtx *ioctx, const std::string &oid,
+			const cls::rbd::GroupSnapshot &start,
+			uint64_t max_return,
+			std::vector<cls::rbd::GroupSnapshot>& snapshots) {
```
dillaman: Minor: out params should be pointers not references

src/cls/rbd/cls_rbd_types.h
```diff
@@ -310,6 +310,55 @@ struct SnapshotNamespaceOnDisk {
...
 WRITE_CLASS_ENCODER(SnapshotNamespaceOnDisk);
 
+enum GroupSnapshotState {
+  GROUP_SNAPSHOT_STATE_PENDING = 0,
+  GROUP_SNAPSHOT_STATE_COMPLETE = 1,
+};
+
+inline void encode(const GroupSnapshotState &state, bufferlist& bl,
+		   uint64_t features=0)
+{
```
dillaman: Minor: you can move this to the previous line

src/cls/rbd/cls_rbd_types.h
```diff
@@ -310,6 +310,55 @@ struct SnapshotNamespaceOnDisk {
...
+WRITE_CLASS_ENCODER(ImageSnapshotRef);
+
+struct GroupSnapshot {
+  static const uint64_t invalid_id = -1;
+  uint64_t id = invalid_id;
+  std::string uuid;
+  std::string name;
+  GroupSnapshotState state;
+
+  vector<ImageSnapshotRef> snaps;
```
dillaman: At first glance, I am not sure we need two-way linkage between the group snapshot and the associate image snapshot. The image's snapshot already has the link back to the group snapshot and the group has links to all of the associated images. Therefore, I can iterate through all the images in the group to get the associated snapshots. I think we only need this if there is some high performance need to bypass the extra lookup.

src/cls/rbd/cls_rbd_types.h
```diff
@@ -310,6 +310,55 @@ struct SnapshotNamespaceOnDisk {
...
+  void decode(bufferlist::iterator& it);
+
+  void dump(Formatter *f) const;
+};
+WRITE_CLASS_ENCODER(ImageSnapshotRef);
+
+struct GroupSnapshot {
+  static const uint64_t invalid_id = -1;
+  uint64_t id = invalid_id;
+  std::string uuid;
```
dillaman: Minor: would be good to keep the naming consistent. ```GroupSnapshotNamespace``` refers to ```group_id``` which should be this ```uuid```, so I'd recommend renaming to ```id``` (since the sequence id isn't needed)

src/cls/rbd/cls_rbd.cc
```diff
@@ -4797,6 +4803,246 @@ int image_get_group(cls_method_context_t hctx,
...
+{
+  CLS_LOG(20, "group_snap_save");
+  cls::rbd::GroupSnapshot gs;
+  try {
+    bufferlist::iterator iter = in->begin();
+    ::decode(gs, iter);
+  } catch (const buffer::error &err) {
+    return -EINVAL;
+  }
+
```
dillaman: Need some basic validation of the provided group snapshot record. UUID empty, name empty, duplicate name, etc. Same comment applies to the update method.

src/include/rbd/librbd.h
```diff
@@ -159,6 +159,11 @@ typedef struct {
 } rbd_group_spec_t;
 
 typedef enum {
+  GROUP_SNAP_STATE_PENDING,
```
dillaman: Don't forget to add the C and Python API methods to match the new C++ API methods

src/include/rbd/librbd.hpp
```diff
@@ -173,6 +180,13 @@ class CEPH_RBD_API RBD
   int group_image_list(IoCtx& io_ctx, const char *group_name,
 		       std::vector<group_image_status_t>& images);
 
+  int group_snap_create(IoCtx& io_ctx, const char *group_name,
+			const char *snap_name);
+  int group_snap_remove(IoCtx& io_ctx, const char *group_name,
+			const char *snap_name);
+  int group_snap_list(IoCtx& group_ioctx, const char *group_name,
```
dillaman: Minor: should have support for rename as well.

src/librbd/Group.cc
```diff
@@ -379,4 +382,405 @@ int image_get_group(ImageCtx *ictx, group_spec_t *group_spec)
 
   return 0;
 }
+
+int group_snap_name_check_duplicate(librados::IoCtx& group_ioctx,
```
dillaman: Should be enforced in cls_rbd not here

src/librbd/Group.cc
```diff
@@ -379,4 +382,405 @@ int image_get_group(ImageCtx *ictx, group_spec_t *group_spec)
...
+  std::vector<group_image_status_t> images;
+  r = librbd::group_image_list(group_ioctx, group_name, images);
+  if (r < 0) {
+    return r;
+  }
+  int n = images.size();
+  std::vector<librbd::IoCtx*> io_ctxs;
+  std::vector<librbd::ImageCtx*> ictxs;
+  std::vector<C_SaferCond*> on_finishes;
+  for (auto i: images) {
```
dillaman: Minor: ```i``` -> ```image```

src/librbd/Group.cc
```diff
@@ -379,4 +382,405 @@ int image_get_group(ImageCtx *ictx, group_spec_t *group_spec)
...
+    librbd::IoCtx* image_io_ctx = new librbd::IoCtx;
```
dillaman: Minor: include ```()``` at end here and below for C_SaferCond

src/librbd/Group.cc
```diff
@@ -379,4 +382,405 @@ int image_get_group(ImageCtx *ictx, group_spec_t *group_spec)
...
+
+    r = rados.ioctx_create2(i.pool, *image_io_ctx);
+    if (r < 0) {
+      ldout(cct, 1) << "Failed to create io context for image" << dendl;
+    }
+
+    librbd::ImageCtx* image_ctx = new ImageCtx(i.name.c_str(), "", "",
```
dillaman: "snap_name" param should be nullptr -- not ""

src/librbd/Group.cc
```diff
@@ -379,4 +382,405 @@ int image_get_group(ImageCtx *ictx, group_spec_t *group_spec)
...
+      delete ictxs[i];
+      ictxs[i] = nullptr;
+      ret_code = r;
+    }
+  }
+  if (ret_code != 0) {
+    goto cleanup;
+  }
+  for (auto i: ictxs) {
+    i->exclusive_lock->block_requests(-EBUSY);
```
dillaman: Need to hold ```owner_lock``` when touching ```ImageCtx::exclusive_lock``` -- and should verify that it isn't a nullptr before dereferencing. Same comment applies below.

src/librbd/Group.cc
```diff
@@ -379,4 +382,405 @@ int image_get_group(ImageCtx *ictx, group_spec_t *group_spec)
...
+    if (r < 0) {
+      delete ictxs[i];
+      ictxs[i] = nullptr;
+      ret_code = r;
+    }
+  }
+  if (ret_code != 0) {
+    goto cleanup;
+  }
+  for (auto i: ictxs) {
```
dillaman: Minor: ```i``` -> ```image_ctx``` or ```ictx``` (use variable names w/ meaning). Comment applies throughout.

src/librbd/Group.cc
```diff
@@ -379,4 +382,405 @@ int image_get_group(ImageCtx *ictx, group_spec_t *group_spec)
...
+    return r;
+  }
+  string group_header_oid = util::group_header_name(group_id);
+
+  std::vector<group_image_status_t> images;
+  r = librbd::group_image_list(group_ioctx, group_name, images);
+  if (r < 0) {
+    return r;
+  }
+  int n = images.size();
```
dillaman: Minor: ```n``` -> ```image_count``` or something similar

src/librbd/Group.cc
```diff
@@ -379,4 +382,405 @@ int image_get_group(ImageCtx *ictx, group_spec_t *group_spec)
...
+  }
+
+  r = cls_client::group_snap_next_seq(&group_ioctx, group_header_oid, &snap_seq);
+  if (r < 0) {
+    ret_code = r;
+    goto cleanup;
+  }
+
+  gs.id = snap_seq;
+  gs.uuid = "";
```
dillaman: Generate a UUID

src/librbd/Group.cc
```diff
@@ -379,4 +382,405 @@ int image_get_group(ImageCtx *ictx, group_spec_t *group_spec)
...
+
+  r = cls_client::group_snap_save(&group_ioctx, group_header_oid, gs);
+  if (r < 0) {
+    ret_code = r;
+    goto cleanup;
+  }
+
+  image_snaps = vector<cls::rbd::ImageSnapshotRef>(n, cls::rbd::ImageSnapshotRef());
+
+  ind_snap_name = calc_ind_image_snap_name(group_ioctx.get_id(),
```
dillaman: Can just re-use the group snapshot uuid -- since it's guaranteed to be unique within the group namespace

src/librbd/Group.cc
```diff
@@ -379,4 +382,405 @@ int image_get_group(ImageCtx *ictx, group_spec_t *group_spec)
...
+  for (int i = 0; i < n; ++i) {
+    ImageCtx *ictx = ictxs[i];
+    cls::rbd::SnapshotNamespace ne =
+			 cls::rbd::GroupSnapshotNamespace(group_ioctx.get_id(),
+							  group_id,
+							  snap_seq);
+
+    C_SaferCond* on_finish = new C_SaferCond;
+
+    ictx->operations->snap_create(ind_snap_name.c_str(), ne, on_finish);
```
dillaman: Just though about this: the ```add_snap``` method in cls_rbd should validate that the image is a member of the associated group before permitting the creation of a ```GroupSnapshotNamespace```

src/librbd/Group.cc
```diff
@@ -379,4 +382,405 @@ int image_get_group(ImageCtx *ictx, group_spec_t *group_spec)
...
+  for (int i = 0; i < n; ++i) {
+    r = on_finishes[i]->wait();
+    delete on_finishes[i];
+    if (r < 0) {
+      ret_code = r;
+    } else {
+      ImageCtx *ictx = ictxs[i];
+      ldout(cct, 1) << "Get snap id with name " << ind_snap_name << dendl;
+      ictx->snap_lock.get_read();
+      snap_t snap_id = ictx->get_snap_id(ind_snap_name);
```
dillaman: If this double-lookup is required, I'd rather have the SnapshotCreateRequest state machine take an optional snap_id pointer to save the assigned snap id into before completing to avoid this lookup.

src/librbd/Group.cc
```diff
@@ -379,4 +382,405 @@ int image_get_group(ImageCtx *ictx, group_spec_t *group_spec)
...
+cleanup:
+  for (int i = 0; i < n; ++i) {
+    if (ictxs[i] != nullptr) {
+      ictxs[i]->state->close();
+    }
+  }
+  return ret_code;
+}
+
+int group_snap_list_cls(librados::IoCtx& group_ioctx, const char *group_name,
```
dillaman: C++ supports overloading -- no need for a unique name

src/librbd/Group.cc
```diff
@@ -379,4 +382,405 @@ int image_get_group(ImageCtx *ictx, group_spec_t *group_spec)
...
+				 group_name, &group_id);
+  if (r < 0) {
+    lderr(cct) << "error reading consistency group id object: "
+	       << cpp_strerror(r)
+	       << dendl;
+    return r;
+  }
+  string group_header_oid = util::group_header_name(group_id);
+
+  r = group_snap_list_cls(group_ioctx, group_name, snaps);
```
dillaman: Already looked up the id and now ```group_snap_list_cls``` is going to look it up again

src/librbd/Group.cc
```diff
@@ -379,4 +382,405 @@ int image_get_group(ImageCtx *ictx, group_spec_t *group_spec)
...
+  if (r < 0) {
+    return r;
+  }
+  cls::rbd::GroupSnapshot *gs = nullptr;
+  int n = snaps.size();
```
dillaman: Just need a ```for (auto &gs : groups_names)``` loop and break out when you find a match

src/librbd/Group.cc
```diff
@@ -379,4 +382,405 @@ int image_get_group(ImageCtx *ictx, group_spec_t *group_spec)
...
+  ret_code = 0;
+  for (int i = 0; i < n; ++i) {
+    r = on_finishes[i]->wait();
+    delete on_finishes[i];
+    if (r < 0) {
+      delete ictxs[i];
+      ictxs[i] = nullptr;
+      ret_code = r;
+    }
+    if (ictxs[i]->id != gs->snaps[i].image_id) {
```
dillaman: Can't rename image ids

src/librbd/Group.cc
```diff
@@ -379,4 +382,405 @@ int image_get_group(ImageCtx *ictx, group_spec_t *group_spec)
...
+
+  ldout(cct, 20) << "Opened participating images. Deleting snapshots themselves." << dendl;
+
+  image_snap_name = calc_ind_image_snap_name(group_ioctx.get_id(),
+					     group_id,
+					     gs->id);
+  for (int i = 0; i < n; ++i) {
+    ImageCtx *ictx = ictxs[i];
+    ldout(cct, 20) << "Removing individual snapshot with name: " << image_snap_name << dendl;
+    int r = ictx->state->refresh_if_required();
```
dillaman: Just opened the image -- no need to refresh (plus it will be handled automatically within ```Operations``` if required)

dillaman:
@VictorDenisov Need to update the CLI test to handle the new help text and need unit and workunit tests to exercise the new cls/librbd APIs/rbd CLI.

VictorDenisov:
work in progress. I'll update when it's done.

src/cls/rbd/cls_rbd_types.h
VictorDenisov: little clarification request: does it mean want to have something like this:
struct GroupSnapshot {
   std::string id;

dillaman: Yes, please.

src/librbd/Group.cc
dillaman: You should be opening the image via its id and not its name -- no need to lookup the image name.

VictorDenisov: Hm. I thought I couldn't do it with id only. I reread the doc and now I can see it's possible. Thanks.

src/librbd/Group.cc
VictorDenisov: Sorry, not sure I completely understand what you mean. Is this what you mean:
```
cls::rbd::GroupSnapshot *gs = nullptr;
int n = snaps.size();
for (auto &snap : snaps) {
  if (snap.name == string(snap_name)) {
    gs = &snap;
    break;
  }
}
```

dillaman: Yes

src/librbd/Group.cc
VictorDenisov: Can it be done is cls_rbd? When we are handling snapshot_add we have access to the image's object. If we want to check whether the image belongs to the group then we need to go to potentially another object on another osd. Is it possible from cls_rbd layer? My understanding is that this only can be done from librbd layer.

VictorDenisov:
@dillaman do I understand correctly that python bindings are in src/pybind/rbd/rbd.pyx?

dillaman:
@VictorDenisov Yes, that's correct

dachary:

dachary:
<pre>
Error compiling Cython file:
------------------------------------------------------------
...
        char *description
        time_t last_update
        bint up

    ctypedef struct rbd_group_image_spec_t:
        char *name;
                 ^
------------------------------------------------------------

rbd.pyx:134:18: Syntax error in C variable declaration
Compiling rbd.pyx because it changed.
Cythonizing rbd.pyx
  File "/home/jenkins-build/build/workspace/ceph-pull-requests/src/pybind/rbd/setup.py", line 188, in <module>
    os.path.join(os.path.dirname(__file__), "..", "rados")
  File "/usr/lib/python2.7/dist-packages/Cython/Build/Dependencies.py", line 798, in cythonize
    cythonize_one(*args[1:])
  File "/usr/lib/python2.7/dist-packages/Cython/Build/Dependencies.py", line 915, in cythonize_one
    raise CompileError(None, pyx_file)
Cython.Compiler.Errors.CompileError: rbd.pyx
</pre>

VictorDenisov:
@dillaman I have a question about python bindings. They convert ENOENT and EEXIST error codes into ImageExists and ImageNotFound exceptions. I'm thinking about renaming them to ObjectExists and ObjectNotFound, since the same errors can occur with groups, but this change will break backward compatibility. What is your opinion on it?

VictorDenisov:
@dillaman So far I decided to rename the Exceptions to ObjectNotFound. If you don't like it I'll rename back. This PR is ready for review. This PR doesn't have snapshot clone operations for creating consistency groups from it and renaming groups and snapshots. I would prefer to have the in a separate pull request unless you insist.

dillaman:
@VictorDenisov I'd prefer not to break backwards compatibility by changing the class names. You could always just directly throw your own exception in those two cases in the group_create/remote functions

VictorDenisov:
Ok. I'll change it.

VictorDenisov:
Done.

VictorDenisov:
@dillaman I presume you are busy with Kraken. I just wanted to make sure that you noticed the update.

dillaman:
@VictorDenisov Thanks -- once Kraken is released, I plan to get back to the Luminous features.

VictorDenisov:
I'm reworking this code. I'll update when it's done.

dillaman:
retest this please

dillaman:
@VictorDenisov is this ready for review?

VictorDenisov:
@dillaman Yes. It's ready.

dillaman:
@VictorDenisov I seemed to have missed this before, but please add all new types to ```test/encoding/types.h``` to ensure they are properly registered.

src/cls/rbd/cls_rbd_types.h
```diff
@@ -248,13 +253,13 @@ struct GroupSnapshotNamespace {
...
-							group_id(_group_id),
-							snapshot_id(_snapshot_id) {}
+			 const string &_snapshot_id) :group_pool(_group_pool),
+						      group_id(_group_id),
+						      snapshot_id(_snapshot_id) {}
 
   int64_t group_pool;
   string group_id;
-  snapid_t snapshot_id;
+  string snapshot_id;
```
dillaman: Nit: I think we should either drop the "group_" prefix from all the variables (since it's implied w/ the class name) or be consistent and add it here. I am leaning towards the first choice since it's less verbose.

src/include/rbd/librbd.h
```diff
@@ -146,7 +146,7 @@ typedef enum {
 
 typedef struct {
   char *name;
-  int64_t pool;
+  char *pool;
```
dillaman: Why the change to a C-string? I think it's safe to let the caller decode the pool name -- plus, the pool id is actually more helpful if trying to create IoCtx's because it's safer from races.

src/include/rbd/librbd.h
```diff
@@ -156,10 +156,20 @@ typedef struct {
...
```
dillaman: Same comment here

src/include/rbd/librbd.h
```diff
@@ -758,6 +768,15 @@ CEPH_RBD_API void rbd_group_image_status_cleanup(
...
 					      rbd_group_image_status_t *images,
 					      size_t len);
+CEPH_RBD_API int rbd_group_snap_create(rados_ioctx_t group_p, const char *group_name,
+					const char *snap_name);
+CEPH_RBD_API int rbd_group_snap_remove(rados_ioctx_t group_p, const char *group_name,
+					const char *snap_name);
+CEPH_RBD_API int rbd_group_snap_list(rados_ioctx_t group_p, const char *group_name,
+				     rbd_group_snap_spec_t *snaps, size_t *snaps_size);
+CEPH_RBD_API void rbd_group_snap_spec_cleanup(rbd_group_snap_spec_t *group_snap_spec);
+CEPH_RBD_API void rbd_group_snap_spec_list_cleanup(rbd_group_snap_spec_t *snaps,
```
dillaman: Nit: rename to ```rbd_group_snap_list_cleanup```

src/include/rbd/librbd.h
```diff
@@ -758,6 +768,15 @@ CEPH_RBD_API void rbd_group_image_status_cleanup(
 CEPH_RBD_API void rbd_group_image_status_list_cleanup(
 					      rbd_group_image_status_t *images,
 					      size_t len);
+CEPH_RBD_API int rbd_group_snap_create(rados_ioctx_t group_p, const char *group_name,
+					const char *snap_name);
+CEPH_RBD_API int rbd_group_snap_remove(rados_ioctx_t group_p, const char *group_name,
+					const char *snap_name);
+CEPH_RBD_API int rbd_group_snap_list(rados_ioctx_t group_p, const char *group_name,
+				     rbd_group_snap_spec_t *snaps, size_t *snaps_size);
+CEPH_RBD_API void rbd_group_snap_spec_cleanup(rbd_group_snap_spec_t *group_snap_spec);
```
dillaman: What would this be used for?

src/include/rbd/librbd.hpp
```diff
@@ -15,14 +15,18 @@
 #ifndef __LIBRBD_HPP
 #define __LIBRBD_HPP
 
+#include <boost/variant.hpp>
 #include <stdbool.h>
 #include <string>
 #include <list>
 #include <map>
 #include <vector>
+#include "common/Formatter.h"
```
dillaman: This is not part of the public API -- neither are the other two new includes, so you cannot add them.

src/include/rbd/librbd.hpp
```diff
@@ -35,9 +39,51 @@ namespace librbd {
...
+    }
+  } group_snap_namespace_t;
+
+  typedef boost::variant<user_snap_namespace_t, group_snap_namespace_t, unknown_snap_namespace_t> snap_namespace_t;
+
+  typedef struct {
     uint64_t id;
     uint64_t size;
     std::string name;
+    snap_namespace_t snap_namespace;
```
dillaman: Cannot change the existing API -- this would break ABI and crash older clients using the newer library. You will just need to add a new helper method to retrieve the namespace of individual snapshots by id.

src/include/rbd/librbd.hpp
```diff
@@ -35,9 +39,51 @@ namespace librbd {
...
+    }
+
+    std::string to_string() const {
+      return "group snapshot member - " + group_pool +
+				   "/" + group_name +
+				   "@" + snap_name;
+    }
+  } group_snap_namespace_t;
+
+  typedef boost::variant<user_snap_namespace_t, group_snap_namespace_t, unknown_snap_namespace_t> snap_namespace_t;
```
dillaman: Cannot use a variant since future additions would break the ABI

src/include/rbd/librbd.hpp
dillaman: Since you will have to support the C API as well, and you might as well keep the interfaces semi-similar, perhaps add a new snapshot namespace enum, a new ```snap_get_namespace_type``` call that will retrieve the namespace enum, and a new ```snap_get_group_spec``` (or similar) method to retrieve the group information should the namespace type by a group.

src/include/rbd/librbd.hpp
```diff
@@ -75,15 +121,22 @@ namespace librbd {
 
   typedef struct {
     std::string name;
-    int64_t pool;
+    std::string pool;
```
dillaman: Same comment about keeping the pool id

src/include/rbd/librbd.hpp
```diff
@@ -75,15 +121,22 @@ namespace librbd {
...
     group_image_state_t state;
   } group_image_status_t;
 
   typedef struct {
     std::string name;
-    int64_t pool;
+    std::string pool;
```
dillaman: Same comment about keeping the pool id

src/librbd/ImageCtx.h
```diff
@@ -68,6 +68,7 @@ namespace librbd {
                                         // a format librados can understand
     std::map<librados::snap_t, SnapInfo> snap_info;
     std::map<std::string, librados::snap_t> snap_ids;
+    std::map<cls::rbd::SnapshotNamespace, librados::snap_t> snap_namespace_ids;
```
dillaman: Please merge ```snap_ids ``` w/ ```snap_namespace_ids```. Use a std::pair of the namespace + the snapshot name for the unique key.

src/tools/rbd/ArgumentTypes.h
```diff
@@ -45,6 +45,7 @@ static const std::string JOURNAL_SPEC("journal-spec");
 static const std::string PATH_NAME("path-name");
 
 // optional arguments
+static const std::string ALL_NAME("all");
```
dillaman: Nit: just declare/define this new argument within Snap.cc since that's the only place it's currently used. The fields are are re-used between other actions.

dillaman:
@VictorDenisov Can you create a simple script in "qa/workunits/rbd" that tests all the new group commands available from the rbd CLI (i.e. create/delete groups, attach/detach images, create/remove snapshots, etc)?

VictorDenisov:
Sure. I'll add the test.

VictorDenisov:
@dillaman What kind of types should be added to test/encoding/types.h . I looked at types from cls_rbd_types.h. It looks like all enums are not added. Also struct MirrorImageStatus is not added. Is there something special about MirrorImageStatus or was it just overlooked?

dillaman:
@VictorDenisov Basically, any non-POD type w/ an encode/decode should be added to that test. In the case of MirrorImageStatus, it was overlooked.

src/cls/rbd/cls_rbd_types.h
VictorDenisov: After trying to rename it I realized that it looks confusing from user's perspective.
If I have code in librbd/internal.cc like:
```
rados.ioctx_create2(snap_namespace.group_pool, group_ioctx);
```
or 
```
string group_header_oid = util::group_header_name(snap_namespace.group_id);
```
When it only has id instead of group_id then the impression is that it's a namespace id rather then a group id that this namespace references. In this regard I would prefer to add group to snapshot_id.

src/cls/rbd/cls_rbd_types.h
```diff
@@ -310,6 +329,52 @@ struct SnapshotNamespaceOnDisk {
...
+  void dump(Formatter *f) const;
+};
+WRITE_CLASS_ENCODER(ImageSnapshotRef);
+
+struct GroupSnapshot {
+  std::string id;
+  std::string name;
+  GroupSnapshotState state;
+
+  vector<ImageSnapshotRef> snaps;
```
VictorDenisov: Images can be removed and added to a group. Snapshots should stay intact. This list is primarily for keeping the list of images. Snapshot id in this reference comes in handy when we need to list image snapshots in this group snapshot.

src/include/rbd/librbd.hpp
VictorDenisov: I presume it's also applicable to group_snap_namespace_t in this file, isn't it?

src/cls/rbd/cls_rbd_types.h
dillaman: How does OpenStack behave? It seems odd to me to remove an image from a group w/o also deleting the group's associated snapshots within the image. Since we only permit group snapshots to be manipulated via the group, that seems like an oddity (i.e. the image is not part of a group anymore, but you cannot delete the image w/o deleting the linked group snapshots -- which would affect all images still within the group).

src/include/rbd/librbd.hpp
dillaman: Yup -- but I skipped over it since I knew you would have to refactor it to eliminate the variant.

src/cls/rbd/cls_rbd_types.h
VictorDenisov: According to my conversation with Xing, images can be removed and added. Here is the link to my email thread with him: http://lists.openstack.org/pipermail/openstack-dev/2016-November/106958.html

VictorDenisov:
Work in progress. I'll update when it's done.

src/include/rbd/librbd.h
VictorDenisov: You mean it shouldn't be exposed as a public API? This method is used by rbd_group_snap_spec_list_cleanup. Python bindings only use rbd_group_snap_spec_list_cleanup, I just thought I would add this method for consistency. If we can cleanup a list of snapshot specs, we should be able to cleanup an individual snapshot spec. Do you still want me to remove this method?

dillaman: Correct --- I don't see a use-case for cleaning up just one structure vs freeing up the entire list you just retrieved via the _list method.

VictorDenisov: Fair enough. I'll inline it in rbd_group_snap_spec_list_cleanup.

src/librbd/ImageCtx.h
VictorDenisov: I tried merging snap_ids and snap_namespace_ids when I just started working on this and it required so much refactoring of the existing code that I decided to do it in a separate PR if at all. Here the issues I've encountered. Probably, you'll be able to repel my fears. If I use pair <SnapshotNamespace, String> as the key for each entry then I can't search by name only? It means that everywhere where we want to determine the id of a snapshot by its name we need to specify its namespace as well. It means changing even command line tools in order to allow the user to specify what namespace he wants to look in. Which seems to me like a too big change.

If we use reversed pair <string, SnapshotNamespace> then we can look for snapshots using only their names, but then we can't look for snapshots using only their Namespace. But this is why we want this index in the first place. Taking all of this into account I decided to just have a separate index for namespaces. What is your thought on this?

dillaman: I am probably missing something, but you would need to do this regardless since you can have duplicate names between namespaces. There really would never be a valid case where you need to search by name only since you would need the namespace disambiguation. For the legacy snapshots, you would just default construct an image namespace and the snapshot name. For group snapshots, you would require the full group namespace to find the correct snapshot.

I would actually recommend breaking this piece of refactoring into its own commit within the PR since it is an atomic unit of work that needs to be done and would facilitate the review.

VictorDenisov: Should we allow users to rename snapshots from GroupSnapshotNamespace?

VictorDenisov: I'll do this refactoring in a separate branch and will show you. If you like it then I'll rebase this PR on top of the refactoring branch.

VictorDenisov: @dillaman I created another PR #12970 for this refactoring. I would like to move through that PR gradually. Could you please take a look at the current version. It has only changes to ImageCtx. Please take a look and tell me if this is what you want to do to ImageCtx in principle. Thanks.

VictorDenisov: The PR is very short yet. It shouldn't take long to review.

VictorDenisov:
These changes are rebased onto master, but they still require some comments to be addressed and some commits to be squashed. I'll update when it's ready.

VictorDenisov:
@dillaman How should those qa/workunits tests be invoked?

dillaman:
@VictorDenisov You can add new workunit test cases in this directory following the example of existing tasks [1]

[1] https://github.com/ceph/ceph/tree/master/qa/suites/rbd/singleton/all

VictorDenisov:
@dillaman  could you please take a look at the change. I addressed the comments. Ceph-test-helpers keeps failing, but it's not clear to me how it's related to my changes.

VictorDenisov:
rebase on the latest master fixed the issue.

src/librbd/api/Group.cc
```diff
@@ -26,20 +30,280 @@ using librados::Rados;
...
+
+  std::string image_snap_name;
+
+  cls::rbd::SnapshotNamespace ne;
+
+  ldout(cct, 20) << "Removing snapshots" << dendl;
+  int snap_count = group_snap.snaps.size();
+
+  for (int i = 0; i < snap_count; ++i) {
+    librbd::IoCtx* image_io_ctx = new librbd::IoCtx;
```
dillaman: Unnecessary to allocate on the heap -- it internally reference counts (plus ```ImageCtx``` just creates its own immediately)

src/librbd/api/Group.cc
```diff
@@ -26,20 +30,280 @@ using librados::Rados;
...
+  std::stringstream ind_snap_name_stream;
+  ind_snap_name_stream << std::setw(16) << std::setfill('0') << std::hex <<
+			  pool_id <<
+			  "_" << group_id <<
+			  "_" << snap_id;
+  return ind_snap_name_stream.str();
+}
+
+static int group_image_list(librados::IoCtx& group_ioctx, const char *group_name,
+    std::vector<cls::rbd::GroupImageStatus> *image_ids)
```
dillaman: Nit: indentation

src/tools/rbd/action/Snap.cc
```diff
@@ -29,13 +43,23 @@ int do_list_snaps(librbd::Image& image, Formatter *f)
   if (r < 0)
     return r;
 
+  if (!all_snaps) {
+    snaps.erase(remove_if(snaps.begin(),
+			  snaps.end(),
+			  boost::bind(is_not_user_snap_namespace, &image, _1)),
+		snaps.end());
```
dillaman: Nit: indentation

src/tools/rbd/action/Snap.cc
```diff
@@ -48,17 +72,40 @@ int do_list_snaps(librbd::Image& image, Formatter *f)
...
     if (f) {
       f->open_object_section("snapshot");
       f->dump_unsigned("id", s->id);
       f->dump_string("name", s->name);
       f->dump_unsigned("size", s->size);
       f->dump_string("timestamp", tt_str);
+      if (all_snaps) {
+	f->open_object_section("namespace");
+	if (get_group_res == 0) {
+	  f->dump_int("pool", group_snap.group_pool);
```
dillaman: Nit: translate pool id to pool name

src/tools/rbd/action/Snap.cc
```diff
@@ -48,17 +72,40 @@ int do_list_snaps(librbd::Image& image, Formatter *f)
...
+	f->close_section();
+      }
       f->close_section();
     } else {
-      t << s->id << s->name << stringify(prettybyte_t(s->size)) << tt_str
-        << TextTable::endrow;
+      std::string namespace_string;
+      if (all_snaps && (get_group_res == 0)) {
+	ostringstream oss;
+	oss << "group snapshot member - " << group_snap.group_pool << "/" <<
```
dillaman: Nit: translate pool id to pool name

dillaman: Nit: can probably shorten "group snapshot member - " to just "group snapshot"

src/librbd/internal.h
```diff
@@ -195,6 +195,39 @@ namespace librbd {
   int metadata_list(ImageCtx *ictx, const string &last, uint64_t max, map<string, bufferlist> *pairs);
   int metadata_get(ImageCtx *ictx, const std::string &key, std::string *value);
 
+  int mirror_mode_get(IoCtx& io_ctx, rbd_mirror_mode_t *mirror_mode);
```
dillaman: Rebase issue here -- pulled in lots of deleted lines

src/librbd/librbd.cc
```diff
@@ -3796,3 +3853,92 @@ extern "C" void rbd_group_image_status_list_cleanup(
...
+    *snaps_size = cpp_snaps.size();
+    tracepoint(librbd, group_snap_list_exit, -ERANGE);
+    return -ERANGE;
+  }
+
+  for (size_t i = 0; i < cpp_snaps.size(); ++i) {
+    group_snap_spec_cpp_to_c(cpp_snaps[i], &snaps[i]);
+  }
+
+  r = cpp_snaps.size();
```
dillaman: Nit: return zero upon success and always update ```snap_size``` w/ the correct count

src/cls/rbd/cls_rbd_types.h
```diff
@@ -328,6 +336,68 @@ struct SnapshotNamespaceOnDisk {
 };
 WRITE_CLASS_ENCODER(SnapshotNamespaceOnDisk);
 
+SnapshotNamespaceType get_namespace_type(const SnapshotNamespace& snapshot_namespace);
+
+enum GroupSnapshotState {
+  GROUP_SNAPSHOT_STATE_PENDING = 0,
```
dillaman: Nit: can we rename this to ```GROUP_SNAPSHOT_STATE_INCOMPLETE```? If there is ever an issue, a PENDING group snapshot will never migrate to COMPLETE.

src/pybind/rbd/rbd.pyx
```diff
@@ -2553,3 +2841,98 @@ cdef class SnapIterator(object):
         if self.snaps:
             rbd_snap_list_end(self.snaps)
             free(self.snaps)
+
+cdef class ImageIterator(object):
```
dillaman: Nit: rename to ```GroupImageIterator```

qa/workunits/rbd/rbd_consistency_groups.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/sh -ex
```
dillaman: Nit: rename this to "rbd_groups.sh"

qa/suites/rbd/cli/workloads/rbd_cli_consistency_groups.yaml
```diff
@@ -0,0 +1,5 @@
+tasks:
```
dillaman: Nit: rename this to "rbd_cli_groups.yaml"

qa/workunits/rbd/rbd_consistency_groups_helpers.sh
```diff
@@ -0,0 +1,136 @@
+#!/bin/sh
```
dillaman: Nit: combine this w/ rbd_groups.sh -- a separate file isn't much use when you don't have multiple tests referencing common functions

src/librbd/internal.cc
```diff
@@ -2056,6 +2056,73 @@ int validate_pool(IoCtx &io_ctx, CephContext *cct) {
     }
   }
 
+  int snap_get_namespace_type(ImageCtx *ictx,
+			      uint64_t snap_id,
+			      snap_namespace_type_t *namespace_type) {
+    const SnapInfo *snap_info;
+    {
+      RWLock::RLocker l(ictx->snap_lock);
+      snap_info = ictx->get_snap_info(snap_id);
```
dillaman: Check for nullptr -- cannot reference ```snap_info``` outside lock scope

src/librbd/internal.cc
```diff
@@ -2056,6 +2056,73 @@ int validate_pool(IoCtx &io_ctx, CephContext *cct) {
...
+					   snap_namespace.group_snapshot_id,
+					   &group_snapshot);
+
+      group_snap.group_pool = group_ioctx.get_id();
+      group_snap.group_name = group_name;
+      group_snap.group_snap_name = group_snapshot.name;
+    }
+  };
+
+  int snap_get_group(ImageCtx *ictx,
```
dillaman: Nit: I would move this to api/Group

src/librbd/internal.cc
```diff
@@ -2056,6 +2056,73 @@ int validate_pool(IoCtx &io_ctx, CephContext *cct) {
...
+		     uint64_t snap_id,
+		     group_snap_t *group_snap) {
+    const SnapInfo *snap_info;
+    {
+      RWLock::RLocker l(ictx->snap_lock);
+      snap_info = ictx->get_snap_info(snap_id);
```
dillaman: Same comment here

src/librbd/internal.cc
```diff
@@ -2056,6 +2056,73 @@ int validate_pool(IoCtx &io_ctx, CephContext *cct) {
...
+    }
+    *namespace_type = static_cast<snap_namespace_type_t>(get_namespace_type(snap_info->snap_namespace));
+    return 0;
+  }
+
+  class GetGroupVisitor : public boost::static_visitor<void> {
```
dillaman: Make this a "int" visitor so that you can return your result code

src/librbd/internal.cc
```diff
@@ -2056,6 +2056,73 @@ int validate_pool(IoCtx &io_ctx, CephContext *cct) {
...
+  public:
+    librados::IoCtx *image_ioctx;
+    group_snap_t group_snap;
+    bool found = false;
```
dillaman: No need for these two variables once you can directly return your result code

src/librbd/internal.cc
```diff
@@ -2056,6 +2056,73 @@ int validate_pool(IoCtx &io_ctx, CephContext *cct) {
...
+    template <typename T>
+    inline void operator()(const T&) const {
+      // ignore other than GroupSnapshotNamespace types.
+    }
+
+    inline void operator()(const cls::rbd::GroupSnapshotNamespace& snap_namespace) {
+      found = true;
+      librados::Rados rados(*image_ioctx);
+      IoCtx group_ioctx;
+      rados.ioctx_create2(snap_namespace.group_pool, group_ioctx);
```
dillaman: Check for success

src/librbd/internal.cc
```diff
@@ -2056,6 +2056,73 @@ int validate_pool(IoCtx &io_ctx, CephContext *cct) {
...
+
+      cls::rbd::GroupSnapshot group_snapshot;
+
+      std::string group_name;
+      r = cls_client::dir_get_name(&group_ioctx, RBD_GROUP_DIRECTORY,
+				   snap_namespace.group_id, &group_name);
```
dillaman: Return result upon failure

src/librbd/internal.cc
```diff
@@ -2056,6 +2056,73 @@ int validate_pool(IoCtx &io_ctx, CephContext *cct) {
...
 
+      string group_header_oid = util::group_header_name(snap_namespace.group_id);
+
+      r = cls_client::group_snap_get_by_id(&group_ioctx,
+					   group_header_oid,
+					   snap_namespace.group_snapshot_id,
+					   &group_snapshot);
+
```
dillaman: Return result upon failure

src/librbd/internal.cc
```diff
@@ -2056,6 +2056,73 @@ int validate_pool(IoCtx &io_ctx, CephContext *cct) {
...
+      snap_info = ictx->get_snap_info(snap_id);
+    }
+    *namespace_type = static_cast<snap_namespace_type_t>(get_namespace_type(snap_info->snap_namespace));
+    return 0;
+  }
+
+  class GetGroupVisitor : public boost::static_visitor<void> {
+  public:
+    librados::IoCtx *image_ioctx;
+    group_snap_t group_snap;
```
dillaman: Just pass the pointer so you can directly update

src/librbd/internal.cc
dillaman: Nit: move to api/Group within an anonymous namespace

src/librbd/api/Group.cc
```diff
@@ -318,80 +628,273 @@ int Group<I>::image_list(librados::IoCtx& group_ioctx,
...
+	       << dendl;
     return r;
+  }
+  string group_header_oid = util::group_header_name(group_id);
 
-  if (-1 != ictx->group_spec.pool_id) {
-    librados::Rados rados(ictx->md_ctx);
-    IoCtx ioctx;
-    rados.ioctx_create2(ictx->group_spec.pool_id, ioctx);
+  r = group_snap_list(group_ioctx, group_name, &snaps);
```
dillaman: Good use for a ```group_snap_get_by_name``` function. That would allow the removal of cls_rbd::check_duplicate_snap_name because it can just use the a helper "get_by_name" function

src/librbd/api/Group.cc
```diff
@@ -26,20 +30,280 @@ using librados::Rados;
 namespace librbd {
 namespace api {
 
-// Consistency groups functions
+// Auxiliary functions
 
-template <typename I>
-int Group<I>::create(librados::IoCtx& io_ctx, const char *group_name)
+static string generate_uuid(librados::IoCtx& io_ctx)
```
dillaman: Put all static functions in an anonymous namespace near the top of the file

dillaman:
@VictorDenisov looking close -- can you remove any newly added "consistency" verbiage references since we are otherwise just referring to this functionality as groups.

VictorDenisov:
Yep. Will do. Most likely will be ready after the weekend.

src/librbd/api/Group.cc
VictorDenisov: I presume static keyword should be removed.

VictorDenisov:
I was too optimistic. I'll update when I'm done.

src/librbd/api/Group.cc
dillaman: Yes, the ```static``` keyword is not required when placed in an anonymous namespace

src/tools/rbd/action/Snap.cc
VictorDenisov: This indentation is meant to show that snaps.end() is erase's argument as well as remove_id. Do you want me to join it all in one line and wrap based on line length?

src/librbd/librbd.cc
VictorDenisov: I can assign this to snaps_size, but it looks like all list functions return list size in case of success, not zero.

src/tools/rbd/action/Snap.cc
dillaman: OK -- disregard then.

src/librbd/librbd.cc
dillaman: The C API is half-and-half -- I'd prefer to return zero upon success and avoid signed vs unsigned int issues

VictorDenisov:
@dillaman ready for another review.

dillaman:
@VictorDenisov Looks like lots of related test failures:
http://pulpito.ceph.com/jdillaman-2017-05-15_11:02:46-rbd-wip-jd-testing-distro-basic-smithi/

dillaman:
@VictorDenisov ... one more thing: it looks like your recent commits are missing the "librbd" prefix. In fact, they can probably just be squashed down to the commit for the comment they are addressing.

src/tools/rbd/action/Snap.cc
```diff
@@ -220,6 +281,13 @@ int execute_list(const po::variables_map &vm) {
   if (r < 0) {
     return r;
   }
+  bool all_snaps = vm[ALL_NAME].as<bool>();
+  r = utils::get_pool_image_snapshot_names(
```
dillaman: Looks like this is a rebase error

VictorDenisov:
@dillaman How can I trigger this test run(http://pulpito.ceph.com/jdillaman-2017-05-15_11:02:46-rbd-wip-jd-testing-distro-basic-smithi/) again and see the results?

dillaman:
@VictorDenisov The majority of the tests you can run locally if you just start a cluster via "vstart" and manually run the workunit test.

VictorDenisov:
@dillaman I fixed python tests, but I don't understand why rbd_groups.sh fails. It works for me locally. Is it possible that you have some other changes in your wip branch that interfere with my code?

dillaman:
@VictorDenisov It might have been a rebase error that you have since corrected. Let me attempt to run it locally.

dillaman:
@VictorDenisov The rbd_group test fails for me locally if I don't override the ```rbd default features``` and also if I use the data pool feature.

VictorDenisov:
I see. I'll look into it. Thanks,

VictorDenisov:
It looks like the problem is that I open images using their ids instead of names. Opening images by ids has a completely different workflow. I'm still looking into it, just wanted to share my findings.

VictorDenisov:
@dillaman I fixed the problem with datapool. It should be good for another run of teuthology.

dillaman:
@VictorDenisov ack -- thanks

Songweibin:
@dillaman  @VictorDenisov  Hi, is this PR will merge into the version of Luminous? Thanks.

dillaman:
@Songweibin I think mimic is the most likely release

VictorDenisov:
@dillaman I rebased the changes. Let me know if anything else is required to merge this PR.

dillaman:
@VictorDenisov thanks -- I'll give it a quick once-over tomorrow and re-run it through testing.

Songweibin:
Hi, @VictorDenisov 
I tested this in my local and got an rbd error as follow:
```
root@ceph:/ceph-dev/build# ./bin/rbd snap list i1 -a
SNAPID NAME                                            SIZE TIMESTAMP                NAMESPACE                  
     4 0000000000000003_103974b0dc51_103f3d1b58ba 102400 kB Tue Nov 28 18:25:02 2017 group snapshot - rbd/g1@s1 
root@ceph:/ceph-dev/build# ./bin/rbd du i1          
2017-11-28 18:37:35.588 7f9e977fe700 -1 librbd::image::OpenRequest: failed to set image snapshot: (2) No such file or directory
rbd: error opening snapshot i1@0000000000000003_103974b0dc51_103f3d1b58ba: (2) No such file or directory
NAME PROVISIONED USED
```

src/tools/rbd/action/Group.cc
```diff
@@ -287,20 +288,19 @@ int execute_list_images(const po::variables_map &vm) {
...
       state_string = "incomplete";
     }
     if (f) {
       f->dump_string("image name", image_name);
-      f->dump_int("pool id", pool_id);
+      f->dump_int("pool", image.pool);
       f->dump_int("state", state);
     } else
-      std::cout << pool_id << "." << image_name << " " << state_string << std::endl;
+      std::cout << image.pool << "/" << image_name << " " << state_string << std::endl;
```
Songweibin: @VictorDenisov  Why not present `pool_name/image_name` which i think is more readable than `pool_id/image_name`, thanks.

dillaman:
@VictorDenisov Similar issue: I run "rbd group rm <name>" on a group w/ snapshots and I see "librbd::Operations: No such snapshot found." errors (and the snapshots are not properly deleted from the image).

dillaman:
Unrelated to these changes, but ```execute_add``` and ```execute_remove_image``` in ```src/tools/rbd/action/Group.cc``` both leak memory due to the second use of ```util::init``` when the second call should be ```util::init_io_ctx``` instead.

src/cls/rbd/cls_rbd_types.h
```diff
@@ -328,6 +336,67 @@ struct SnapshotNamespaceOnDisk {
...
+  void dump(Formatter *f) const;
+
+  static void generate_test_instances(std::list<ImageSnapshotSpec *> &o);
+};
+WRITE_CLASS_ENCODER(ImageSnapshotSpec);
+
+struct GroupSnapshot {
+  std::string id;
+  std::string name;
+  GroupSnapshotState state;
```
dillaman: Nit: default initialize to a valid value

src/tools/rbd/action/Group.cc
```diff
@@ -396,6 +588,15 @@ Shell::Action action_remove_image(
 Shell::Action action_list_images(
   {"group", "image", "list"}, {}, "List images in a consistency group.",
   "", &get_list_images_arguments, &execute_list_images);
+Shell::Action action_group_snap_create(
+  {"group", "snap", "create"}, {}, "Make a snapshot of a group.",
+  "", &get_group_snap_create_arguments, &execute_group_snap_create);
+Shell::Action action_group_snap_remove(
+  {"group", "snap", "remove"}, {}, "Remove a snapshot from a group.",
```
dillaman: Nit: add "rm" alias for "remove" for CLI consistency

src/cls/rbd/cls_rbd.h
```diff
@@ -164,4 +164,15 @@ struct cls_rbd_snap {
 };
 WRITE_CLASS_ENCODER(cls_rbd_snap)
 
+namespace group {
```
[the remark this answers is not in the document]
VictorDenisov: @dillaman I need this function to be available in unit tests: ceph/src/test/cls_rbd/test_cls_rbd.cc:2308:21. Since the test was not about building snap_key I didn't want to duplicate this code. What is your opinion?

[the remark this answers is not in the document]
dillaman: OK -- leave it then.

src/include/rbd/librbd.hpp
VictorDenisov: Rename will be addressed in a separate PR.

VictorDenisov: https://github.com/ceph/ceph/pull/12431

src/include/rbd/librbd.h
```diff
@@ -894,6 +910,18 @@ CEPH_RBD_API void rbd_group_image_status_cleanup(
...
 					      size_t len);
+CEPH_RBD_API int rbd_group_snap_create(rados_ioctx_t group_p, const char *group_name,
+					const char *snap_name);
+CEPH_RBD_API int rbd_group_snap_remove(rados_ioctx_t group_p, const char *group_name,
+					const char *snap_name);
+CEPH_RBD_API int rbd_group_snap_list(rados_ioctx_t group_p, const char *group_name,
+				     rbd_group_snap_spec_t *snaps, size_t *snaps_size);
+CEPH_RBD_API void rbd_group_snap_list_cleanup(rbd_group_snap_spec_t *snaps,
+						   size_t len);
+
```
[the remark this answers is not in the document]
VictorDenisov: Can I add these features as part of this PR: https://github.com/ceph/ceph/pull/12637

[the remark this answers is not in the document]
dillaman: I'd prefer it to be part of this one to (1) avoid getting lost and (2) since this one added the functions.

src/include/rbd/librbd.hpp
dillaman: Yup -- that was a very old comment.

liewegas:
@dillaman what's the status of this one?

VictorDenisov:
I'm addressing the comments and fixing one bug.

dillaman:
Manually resolved remaining issues and merged to master branch

PR #11545, opened 2016-10-19
osdc: After write try merge bh.

majianpeng:
Fixes:http://tracker.ceph.com/issues/17270

### DIFF
--- a/src/osdc/ObjectCacher.cc
+++ b/src/osdc/ObjectCacher.cc
@@ -1114,6 +1114,7 @@ void ObjectCacher::bh_write_commit(int64_t poolid, sobject_t oid,
       }
     }
 
+    vector<pair<loff_t, BufferHead*>> hit;
     // apply to bh's!
     for (map<loff_t, BufferHead*>::const_iterator p = ob->data_lower_bound(start);
 	 p != ob->data.end();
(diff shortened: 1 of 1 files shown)


src/osdc/ObjectCacher.cc
```diff
@@ -182,6 +182,7 @@ void ObjectCacher::Object::merge_left(BufferHead *left, BufferHead *right)
 
   // hose right
   delete right;
+  right = nullptr; //see bh_write_commit
```
liewegas: Isn't this a local variable?

src/osdc/ObjectCacher.cc
```diff
@@ -1156,6 +1159,13 @@ void ObjectCacher::bh_write_commit(int64_t poolid, sobject_t oid,
 		       << dendl;
       }
     }
+
+    while (!hit.empty()) {
+      BufferHead *bh = hit.front();
+      hit.pop_front();
+      if (bh)
```
liewegas: bh should always be non-null?

majianpeng: #11262 . In hit, the first may merged w/ the second. So i need know which be merged(delete). Yes as you said, set the local var to nullptr after delete don't indicate the related var in hit. But change the merge_left right to BufferHead **right. I think it's not a goog method.

majianpeng:
@liewegas . I think this version can work.

liewegas:
Can you remove the Fixes line from the commit?  It make me think this fixes a bug but it doesn't--it just optimizes in a non-buggy way :)  Thanks!

jcsp:
Tested here: http://pulpito.ceph.com/jspray-2016-11-10_21:38:43-fs-wip-jcsp-testing-20161108-distro-basic-mira/

PR #11553, opened 2016-10-19
rgw multisite: fix the increamtal bucket sync init

Aran85:
in the `RGWBucketShardFullSyncCR::operate`, inc_marker will assigned with remote bilog's max_marker.
but the sync_status's inc_marker cant be assigned.so the next step inc sync will always sync
from null log,which means at beginning log.

### DIFF
--- a/src/rgw/rgw_data_sync.cc
+++ b/src/rgw/rgw_data_sync.cc
@@ -1747,13 +1747,14 @@ class RGWInitBucketShardSyncStatusCoroutine : public RGWCoroutine {
 
   string lock_name;
   string cookie;
-  rgw_bucket_shard_sync_info status;
+  rgw_bucket_shard_sync_info& status;
 
   bucket_index_marker_info info;
 public:
   RGWInitBucketShardSyncStatusCoroutine(RGWDataSyncEnv *_sync_env,
-                                        const rgw_bucket_shard& bs)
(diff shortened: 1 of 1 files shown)


cbodley:
thanks, this looks correct. i'll run some tests to validate

cbodley:
@Aran85 I created a ticket for this bug at http://tracker.ceph.com/issues/17624 so we can make sure this gets backported to jewel. could you please add `Fixes: http://tracker.ceph.com/issues/17624` to your commit message?

Aran85:
@cbodley done

cbodley:
tested successfully using two vstart clusters with one zone each. i started a gateway on the first zone and created a bucket with ~20 objects. i then started the second zone's gateway, and watched it go through full sync. without this fix, the first request for incremental sync uses an empty marker. with the fix, it uses a valid marker :+1: 

that said, this points out another issue with `RGWRunBucketSyncCoroutine` - it shouldn't be keeping a cached copy of the sync status outside of the lock. rather than each of the `InitBucketShardSyncStatus`, `BucketShardFullSync`, and `BucketShardIncrementalSync` coroutines taking and releasing that lock, `RunBucketSync` should acquire it at the beginning and hold it over all of the other calls

`RunBucketSync` also doesn't appear to be handling the parallel full and incremental sync correctly, so we'll want to get that working for https://github.com/ceph/ceph/pull/10995. i'll work on a design that covers both of these issues, and share with the ceph-devel list

PR #11555, opened 2016-10-19
rgw: 0lenfix1

mdw-at-linuxbox:
The first commit is to fix http://tracker.ceph.com/issues/17572.  It also turns out that somebody broke byte ranges on SLO/DLO, so this patch removes the code that broke that.

The 2nd commit redoes "end" counts to get rid of a bunch of +1/-1 stuff, hopefully to make bugs like the first commit less likely.  It also fixes the range parse logic to be a bit more strict (ie, less likely to return wrong results for bad ranges.)

I found the swift test logic in teuthology and ran it directly on my test setup, seems to work.

### DIFF
--- a/src/rgw/rgw_op.cc
+++ b/src/rgw/rgw_op.cc
@@ -825,6 +825,10 @@ int RGWGetObj::read_user_manifest_part(rgw_bucket& bucket,
     return -EPERM;
   }
 
+  if (ent.size == 0) {
+    return 0;
+  }
+
   perfcounter->inc(l_rgw_get_b, cur_end - cur_ofs);
   while (cur_ofs <= cur_end) {
     bufferlist bl;
@@ -834,6 +838,12 @@ int RGWGetObj::read_user_manifest_part(rgw_bucket& bucket,
 
     off_t len = bl.length();
     cur_ofs += len;
+    if (!len) {
+        ldout(s->cct, 0) << "ERROR: read 0 bytes; ofs=" << cur_ofs
+	    << " end=" << cur_end << " from obj=" << ent.key.name
+	    << "[" << ent.key.instance << "]" << dendl;
+        return -EIO;
+    }
     op_ret = 0; /* XXX redundant? */
     perfcounter->tinc(l_rgw_get_lat,
                       (ceph_clock_now(s->cct) - start_time));
@@ -1103,6 +1113,11 @@ int RGWGetObj::handle_user_manifest(const char *prefix)
     return r;
   }
 
+  if (!total_len) {
+    bufferlist bl;
+    send_response_data(bl, 0, 0);
+  }
+
   return 0;
 }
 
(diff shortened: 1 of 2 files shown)


src/rgw/rgw_op.cc
```diff
@@ -110,23 +112,47 @@ static int parse_range(const char *range, off_t& ofs, off_t& end, bool *partial_
...
 
   *partial_content = true;
+  end = -2;
 
+  if (pos == 0 && s.length() == 1)
+    goto done;
   ofs_str = s.substr(0, pos);
   end_str = s.substr(pos + 1);
+  for (il = ofs_str.length(); il > 0; --il)
+    if (!isspace(ofs_str[il-1])) break;
```
oritwas: You can use find_last_not_of

src/rgw/rgw_op.cc
```diff
@@ -110,23 +112,47 @@ static int parse_range(const char *range, off_t& ofs, off_t& end, bool *partial_
...
+  if (il < ofs_str.length())
+    ofs_str.erase(il);
+  for (il = end_str.length(); il > 0; --il)
+    if (!isspace(end_str[il-1])) break;
```
oritwas: see previous comment

src/rgw/rgw_op.cc
```diff
@@ -110,23 +112,47 @@ static int parse_range(const char *range, off_t& ofs, off_t& end, bool *partial_
     goto done;
 
   *partial_content = true;
+  end = -2;
```
oritwas: why? doesn't it overrides the end calculations that is used for the out of range error

ktdreyer:
The ticket that corresponds to this fix is http://tracker.ceph.com/issues/17572

src/rgw/rgw_op.cc
mdw-at-linuxbox: this is for a range that's "x-" -- so end has to be returned as -1 to say "to the end of the file".  the later end tests already know "-1" is an indefinite "end of the file" value.

src/rgw/rgw_op.cc
mdw-at-linuxbox: the code above this already uses isspace().  Also strtoll internally uses isspace().  And with "find_last_not_of" there's an icky problem with deciding exactly which "whitespace" characters one wants to count and I don't think we want that to be different from what strtoll & the isspace code above does.

src/rgw/rgw_op.cc
oritwas: please add a comment.

src/rgw/rgw_op.cc
```diff
@@ -1347,13 +1388,6 @@ void RGWGetObj::execute()
   if (op_ret < 0)
     goto done_err;
 
-  // for range requests with obj size 0
-  if (range_str && !(s->obj_size)) {
```
oritwas: don't we still need this check?

ktdreyer:
The first commit in this PR was already merged to master in https://github.com/ceph/ceph/pull/11567 , so I think this needs to be edited and rebased?

dachary:
needs rebase :-)

oritwas:
@mdw-at-linuxbox please rebase

liewegas:
@mdw-at-linuxbox is this still needed?

liewegas:
This is a stale pull request and the head branch is in ceph.git.  If this is still a useful change, please reopen a new pull request from your own cloned repo (not the main ceph.git).  You can find the original branch archived in ceph-old.git.  Thanks!

PR #11556, opened 2016-10-19
Don't loop forever when reading data from 0 sized segment.

mdw-at-linuxbox:


ktdreyer:
This is to fix http://tracker.ceph.com/issues/17572

ktdreyer:
Closing in favor of https://github.com/ceph/ceph/pull/11626

PR #11557, opened 2016-10-19
os/bluestore: kill overlay related options

xiexingguo:
These two options are not used anymore.


xiexingguo:
@ifed01 Thanks, Igor:-)

PR #11561, opened 2016-10-19
DNM: os/bluestore: improve move_ranges implementation

liewegas:


src/os/bluestore/BlueStore.cc
```diff
@@ -8625,30 +8625,81 @@ int BlueStore::_move_ranges_destroy_src(TransContext *txc,
...
+	skip_front = srcoff - e.logical_offset;
+      } else {
+	skip_front = 0;
+      }
+      if (e.logical_offset + e.length > srcoff + length) {
+	skip_back = e.logical_offset + e.length - (srcoff + length);
+      } else {
+	skip_back = 0;
+      }
+      Extent *ne = new Extent(
```
ifed01: Let's consider following disposition: 
e(10~90) -> blob(0~0x1000)
mv (10~90 - > 0x2000)
it results in
skip_front = 0, skip_back = 0
ne.logical_offset = 10 + 0 + 0x2000 - 10 = 0x2000
ne.blob_offset = 10 + 0 = 10
ne.length = 90 - 0 - 0 = 90
i.e.
new extent, located at alloc unit(AU) boundary(4K) has non zero blob offset. This violates our invariant that blobs are AU aligned and IMHO might have some negative impact on subsequent operations. Makes sense?

src/os/bluestore/BlueStore.cc
```diff
@@ -8625,30 +8625,81 @@ int BlueStore::_move_ranges_destroy_src(TransContext *txc,
...
+	e.logical_offset + skip_front + dstoff - srcoff,
+	e.blob_offset + skip_front,
+	e.length - skip_front - skip_back, e.blob_depth, ep->blob);
```
ifed01: I doubt blob_depth is inherited from the old object - that's rather a matter of disposition in the new object - hence needs recalculation

liewegas:
Perhaps we should restrict this operation to move ranges at identical 
offsets.  That simplifies the API somewhat and avoids the alignment 
issues...

liewegas:
@athanatos does fixing srcoff=dstoff seem ok to you?

athanatos:
@liewegas Hmm, ideally not.  It would mean relying on the rollforward object being sparse.  I supposed that could be ok.

liewegas:
https://github.com/ceph/ceph/pull/11595 .. i'll rebase onto that

liewegas:
Rebased and retested; ready for review

src/test/objectstore/store_test.cc
```diff
@@ -2823,7 +2823,16 @@ TEST_P(StoreTest, SimpleMoveRangeDelSrcTest) {
     cerr << "move temp object" << std::endl;
     r = apply_transaction(store, &osr, std::move(t));
     ASSERT_EQ(r, 0);
-
+  }
+  if (true) {
+    r = store->umount();
+    ASSERT_EQ(0, r);
+    r = store->fsck();
```
ifed01: umount/mount perform fsck with the current settings. hence we have triple fsck here

src/os/bluestore/BlueStore.cc
```diff
@@ -8731,28 +8731,79 @@ int BlueStore::_move_ranges_destroy_src(
...
+      } else {
+	skip_front = 0;
+      }
+      if (e.logical_offset + e.length > offset + length) {
+	skip_back = e.logical_offset + e.length - (offset + length);
+      } else {
+	skip_back = 0;
+      }
+      Extent *ne = new Extent(
+	e.logical_offset + skip_front + offset - offset,
```
ifed01: remove +offset-offset

src/os/bluestore/BlueStore.cc
```diff
@@ -8731,28 +8731,79 @@ int BlueStore::_move_ranges_destroy_src(
...
+	e.blob_offset + skip_front,
+	e.length - skip_front - skip_back, e.blob_depth, ep->blob);
```
ifed01: blob_depth can't be inherited from an old object

liewegas:
Gonna do something different.

PR #11564, opened 2016-10-19
doc: ref cephfs docs from os-recommendations

jcsp:
...instead of having potentially conflicting kernel
version information.

### DIFF
--- a/doc/start/os-recommendations.rst
+++ b/doc/start/os-recommendations.rst
@@ -13,7 +13,7 @@ Linux Kernel
 
 - **Ceph Kernel Client**
 
-  We currently recommend:
+  For RBD, we currently recommend:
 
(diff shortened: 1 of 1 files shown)


dachary:

jcsp:
@dillaman I haven't actually changed the RBD-related versions here, but could you confirm they're still correct?

idryomov:
@jcsp I pushed https://github.com/idryomov/ceph/commits/wip-doc-kernel-vers.  Do you want to pull it into your branch or should I do a follow-up?

jcsp:
@idryomov I see that you've already done the merge to get my commit on top of yours, so probably simplest to just close this PR and you open one with both changes in it

idryomov:
Opened #13369.

PR #11566, opened 2016-10-19
rgw: fix put_acls for objects starting and ending with underscore

oritwas:
Fixes: http://tracker.ceph.com/issues/17625


oritwas:
@yehudasa , add a test in https://github.com/ceph/s3-tests/pull/131

PR #11567, opened 2016-10-19
Don't loop forever when reading data from 0 sized segment.

mdw-at-linuxbox:

Also, don't object to byte ranges on DL/SL objects.

### DIFF
--- a/src/rgw/rgw_op.cc
+++ b/src/rgw/rgw_op.cc
@@ -825,6 +825,10 @@ int RGWGetObj::read_user_manifest_part(rgw_bucket& bucket,
     return -EPERM;
   }
 
+  if (ent.size == 0) {
+    return 0;
+  }
+
   perfcounter->inc(l_rgw_get_b, cur_end - cur_ofs);
   while (cur_ofs <= cur_end) {
     bufferlist bl;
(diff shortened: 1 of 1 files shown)


src/rgw/rgw_op.cc
```diff
@@ -1347,13 +1362,6 @@ void RGWGetObj::execute()
   if (op_ret < 0)
     goto done_err;
 
-  // for range requests with obj size 0
```
yehudasa: @mdw-at-linuxbox was this block in the original patch?

mdw-at-linuxbox:
Yehuda asked me to remove the fix for byte ranges on SL/DL objects
from this patch.  This concerns code that was apparently added to fix the problem identified in http://tracker.ceph.com/issues/16388 .  Without that fix this code will not pass my tests.  However, the patch should be very nearly identical to what I had for jewel.

Anyways, I've done a "force push", so should be all set there.

liewegas:
Please use the usual convention for naming pull requests (subsystem: short description) as this feeds directly into the script that generates release notes and it's tedious to clean up at release time.  (We should also use the same convention for commits themselves, too, so it's clear e.g., that this commit relates to rgw)

tchaikov:
@liewegas fwiw, i just included your comment in #11570. so we can reference it in future.

PR #11571, opened 2016-10-20
Wip rgw openssl 4

mdw-at-linuxbox:
This is a simplified update of PR # 10335.  Just 2 commits,

840c12cc95 load libssl.so and libcrypto.so by name.

722cc904d document that radosgw now supports SSL.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -3,7 +3,7 @@
 	url = https://github.com/ceph/ceph-object-corpus.git
 [submodule "src/civetweb"]
 	path = src/civetweb
-	url = https://github.com/ceph/civetweb
+	url = https://github.com/mdw-at-linuxbox/civetweb
 [submodule "src/erasure-code/jerasure/jerasure"]
 	path = src/erasure-code/jerasure/jerasure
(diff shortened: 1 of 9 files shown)


mdw-at-linuxbox:
With this patch, this is "expected" and normal behavior:
error parsing int: 443s: The option value '443s' seems to be invalid

mdw-at-linuxbox:
I've pushed a new version of this.  This version requires a patch be made to civetweb, to add "mg_get_local_addr".  I've made a separate pr for that, https://github.com/ceph/civetweb/pull/14 .  This version can listen on multiple ports, some of which can have ssl enabled.  It should also work with s3 v4 and swift preauth urls.

mdw-at-linuxbox:
Yehuda asked for a version of this updated to use civetweb 1.8.  I've made a separate pair of PRs for that, https://github.com/ceph/ceph/pull/11776 https://github.com/ceph/civetweb/pull/15

mdw-at-linuxbox:
Since civetweb-1.8 has been committed, I'm going to close this one in favor of #11776.

PR #11572, opened 2016-10-20
fix:rgw crash when client post object with null conditions

diluga:
for: http://tracker.ceph.com/issues/17635

Signed-off-by: Feng Guo <diluga@gmail.com>

### DIFF
--- a/src/common/config.cc
+++ b/src/common/config.cc
@@ -36,6 +36,10 @@
 #include <sys/stat.h>
 #include <sys/types.h>
 
+#include <utility>
+#include <boost/type_traits.hpp>
+#include <boost/utility/enable_if.hpp>
+
 /* Don't use standard Ceph logging in this file.
  * We can't use logging until it's initialized, and a lot of the necessary
  * initialization happens here.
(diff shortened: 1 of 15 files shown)


src/rgw/rgw_policy_s3.cc
```diff
@@ -276,6 +276,10 @@ int RGWPolicy::from_json(bufferlist& bl, string& err_msg)
       int i;
       for (i = 0; !citer.end() && i < 3; ++citer, ++i) {
 	JSONObj *o = *citer;
+        string condition_data = o->get_data();
```
tchaikov: better off using a const reference for holding the returned value of `o->get_data()`. and reference it at line https://github.com/ceph/ceph/pull/11572/commits/e887f62ff8f9602d1ee785f5e256079296554112#diff-d71c299d7f07bbeccbb5a0d3aca46659R283

tchaikov:
could you prefix the title of your commit message with the subcomponent your are changing ? see https://github.com/ceph/ceph/blob/master/SubmittingPatches.rst#preparing-and-sending-patches, in this case, it could be "rgw: ".

tchaikov:
and could you tag your fix using the notation of "Fixes: <URL of the tracker ticket>" in your commit message, see https://github.com/ceph/ceph/blob/master/SubmittingPatches.rst#preparing-and-sending-patches ?

PR #11552, opened 2016-10-19
include/object: pass "snapid_t&" to bound_encode()

tchaikov:
otherwise the STL container's encoders will segfault when trying to
figure out the element size: they use "denc(*(A*)nullptr, elem_size)"
to do the calculation.

Signed-off-by: Kefu Chai <kchai@redhat.com>

PR #11577, opened 2016-10-20
denc: don't pass null instances into encoder fns

jcsp:
In cases like inodeno_t, the object ref passed in
is dereferenced and passing in a null reference
causes a segfault.

Clearly the constructor overhead is best avoided,
but until someone finds a way to do this better, let's
not crash!

This was causing the MDS to crash on startup after
the new encoding code was merged.

Fixes: http://tracker.ceph.com/issues/17636


tchaikov:
i fixed a similar issue (#11552) yesterday!

PR #11578, opened 2016-10-20
mds: repair backtraces during scrub

jcsp:
Fixes: http://tracker.ceph.com/issues/17639


jcsp:
Tested by https://github.com/ceph/ceph-qa-suite/pull/1218

PR #11579, opened 2016-10-20
build/ops: run ceph-osd-prestart.sh with bash

smithfarm:
Newer Debian and Ubuntu systems have /bin/sh linked to "dash" and scripts
containing bashisms (like this one) will fail.

Fixes: http://tracker.ceph.com/issues/17637

### DIFF
--- a/src/ceph-osd-prestart.sh
+++ b/src/ceph-osd-prestart.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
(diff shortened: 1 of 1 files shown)


src/ceph-osd-prestart.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
tchaikov: can we instead remove the bashism in this script? as `ceph-osd` does not depend on "bash".

PR #11580, opened 2016-10-20
rgw: dump remaining coroutines when cr deadlock is detected

cbodley:
when we detect a coroutine deadlock, dump remaining coroutines to lderr before asserting. this should help us figure out what's causing the deadlocks in teuthology runs. example output:
```
2016-10-20 12:17:28.236470 7f52959dea80 -1 run(): ERROR: deadlock detected, dumping remaining coroutines:
[
    {
        "stack": "0x55f9f6bdc3d0",
        "run_count": 0,
        "ops": [
            {
                "type": "17RGWFetchAllMetaCR",
                "spawned": [
                    "0x55f9f6bde3c0"
                ],
                "status": {
                    "status": "acquiring lock",
                    "timestamp": "2016-10-20 16:17:28.235549Z"
                }
            }
        ]
    },
    {
        "stack": "0x55f9f6bde3c0",
        "run_count": 0,
        "ops": [
            {
                "type": "11MockLeaseCR"
            }
        ]
    }
]
```

### DIFF
--- a/src/rgw/rgw_coroutine.cc
+++ b/src/rgw/rgw_coroutine.cc
@@ -588,7 +588,19 @@ int RGWCoroutinesManager::run(list<RGWCoroutinesStack *>& stacks)
   }
 
   lock.get_write();
-  assert(context_stacks.empty() || going_down.read()); // assert on deadlock
+  if (!context_stacks.empty() && !going_down.read()) {
+    JSONFormatter formatter(true);
+    formatter.open_array_section("context_stacks");
+    for (auto& s : context_stacks) {
+      ::encode_json("entry", *s, &formatter);
+    }
+    formatter.close_section();
+    lderr(cct) << __func__ << "(): ERROR: deadlock detected, dumping remaining coroutines:\n";
+    formatter.flush(*_dout);
+    *_dout << dendl;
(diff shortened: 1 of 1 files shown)


src/rgw/rgw_coroutine.cc
```diff
@@ -588,7 +588,19 @@ int RGWCoroutinesManager::run(list<RGWCoroutinesStack *>& stacks)
...
-  assert(context_stacks.empty() || going_down.read()); // assert on deadlock
+  if (!context_stacks.empty() && !going_down.read()) {
+    JSONFormatter formatter(true);
+    formatter.open_array_section("context_stacks");
+    for (auto& s : context_stacks) {
+      ::encode_json("entry", *s, &formatter);
+    }
+    formatter.close_section();
+    lderr(cct) << __func__ << "(): ERROR: deadlock detected, dumping remaining coroutines:\n";
+    formatter.flush(*_dout);
```
yehudasa: @cbodley shouldn't we use _derr?

cbodley: this is referring to the `std::ostream* _dout` variable from the `dout_impl` macro in common/dout.h, which the other `dout`/`derr` macros are using

PR #11581, opened 2016-10-20
rgw: handle empty POST condition

yehudasa:
Fixes: http://tracker.ceph.com/issues/17635

Before accessing json entity, need to check that iterator is valid.


cbodley:
looks good. just waiting on the s3test for it?

cbodley:
i ran this against the test in https://github.com/ceph/s3-tests/pull/132, and the gateway didn't crash :+1:  but the test failed with `AssertionError: 403 != 400`. is this pr missing the `-EINVAL` change you mentioned?

yehudasa:
@cbodley ah, right... failed to push the change. Pushed now.

PR #11582, opened 2016-10-20
core: set dumpable flag after setuid

batrick:
When ceph-* drops drops privileges via setuid, core dumps are no longer
generated because its DUMPABLE flag is cleared. We have to manually
turn that back on.

From prctl(2):

      Normally, this flag is set to 1.  However, it is reset to the current value contained in the file /proc/sys/fs/suid_dumpable (which by default has the value 0), in the fol‐
      lowing circumstances:

      *  The process's effective user or group ID is changed.

      *  The process's filesystem user or group ID is changed (see credentials(7)).

      *  The process executes (execve(2)) a set-user-ID or set-group-ID program, or a program that has capabilities (see capabilities(7)).

Fixes: http://tracker.ceph.com/issues/17650

### DIFF
--- a/src/global/global_init.cc
+++ b/src/global/global_init.cc
@@ -31,6 +31,10 @@
 #include <grp.h>
 #include <errno.h>
 
+#ifdef HAVE_SYS_PRCTL_H
+#include <sys/prctl.h>
+#endif
+
 #define dout_subsys ceph_subsys_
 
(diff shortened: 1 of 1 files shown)


badone:
lgtm

badone:
test this please

badone:
Failure is known issue, http://tracker.ceph.com/issues/17561

PR #11584, opened 2016-10-20
cmake: default lttng to off on opensuse systems 

jan--f:
Package is not available by default. Fixes
http://tracker.ceph.com/issues/17648


cbodley:
i really don't think this kind of thing belongs in the build system. we shouldn't try to detect the perfect defaults for everyone - that's why they're options in the first place. if your build environment doesn't have lttng, just run cmake with `-DWITH_LTTNG=OFF`

liewegas:
Yeah, I think this selection of options belongs in ceph.spec.in along iwth all of the other distro-specific magic.  The key reason to default this on in cmake is because most distros do enable it and it's helpful for developers to notice build failures early.

PR #11585, opened 2016-10-20
test/rgw: unit tests for RGWFetchAllMetaCR

cbodley:
this is an experiment to build up unit tests for rgw sync coroutines. RGWFetchAllMetaCR is a convenient place to start, because it doesn't have a deep call stack - this means the factory interface is small

note that the tests for `ReadSectionsEIO` and `ReadKeysENOENT` will deadlock without https://github.com/ceph/ceph/pull/11505

### DIFF
--- a/src/rgw/rgw_cr_rados.h
+++ b/src/rgw/rgw_cr_rados.h
@@ -1007,7 +1007,19 @@ public:
   }
 };
 
-class RGWContinuousLeaseCR : public RGWCoroutine {
+// interface for unit testing
+class RGWBaseContinuousLeaseCR : public RGWCoroutine {
+ public:
+  RGWBaseContinuousLeaseCR(CephContext *cct) : RGWCoroutine(cct) {}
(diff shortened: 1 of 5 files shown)


cbodley:
this gets pretty good coverage for `RGWFetchAllMetaCR`. i worry that the approach won't scale well though, because we have to inject factories for all of the child crs it can possibly spawn - for something like `RGWMetaSyncCR`, this is going to get messy fast

yehudasa:
@cbodley just a thought, but maybe instead of overriding all the different crs, we can hook into RGWCoroutinesManager::run(), or create a version of it where it only really executes RGWFetchAllMetaCR; figuring the type by the typeid().name(), and then for each child cr it has a callback that modifies what it needs to accordingly. This could scale better (if feasible).

cbodley:
pulled the fix into a separate pr: https://github.com/ceph/ceph/pull/11586

dachary:
needs rebase :-)

cbodley:
closing for now

PR #11586, opened 2016-10-20
rgw: fix break inside of yield in RGWFetchAllMetaCR

cbodley:
the yield macro is implemented with for/switch, so the breaks in
RGWFetchAllMetaCR weren't being applied to the for loop as expected -
so any of these breaks send RGWFetchAllMetaCR into an infinite loop

changed the breaks into returns, so the next call to operate() will
continue with the next iteration of the loop

Fixes: http://tracker.ceph.com/issues/17655

(exposed by unit tests in https://github.com/ceph/ceph/pull/11585)

### DIFF
--- a/src/rgw/rgw_sync.cc
+++ b/src/rgw/rgw_sync.cc
@@ -837,27 +837,25 @@ public:
         }
         iter = result.begin();
         for (; iter != result.end(); ++iter) {
-          RGWRados *store;
-          int ret;
-          yield {
-            if (!lease_cr->is_locked()) {
-              lost_lock = true;
-              break;
-            }
-	    ldout(cct, 20) << "list metadata: section=" << *sections_iter << " key=" << *iter << dendl;
-	    string s = *sections_iter + ":" + *iter;
-            int shard_id;
-            store = sync_env->store;
-            ret = store->meta_mgr->get_log_shard_id(*sections_iter, *iter, &shard_id);
-            if (ret < 0) {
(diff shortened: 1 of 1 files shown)


dachary:

dachary:
jenkins test this please (jenkins failure)

dachary:

src/rgw/rgw_sync.cc
```diff
@@ -840,7 +840,7 @@ class RGWFetchAllMetaCR : public RGWCoroutine {
           yield {
             if (!lease_cr->is_locked()) {
               lost_lock = true;
-              break;
+              return 0;
```
cbodley: will try replacing the yield {} block with a single `yield;` statement - then the `break`s will apply to the for loop as intended

cbodley:
@yehudasa removed the `yield {}` block and added a single `yield;` statement once per iteration of the loop (i put the yield before the local variables so it wouldn't complain about jumping across their declarations)

i rebased the unit tests from https://github.com/ceph/ceph/issues/11585 for testing - without this pr, they hit an infinite loop - with this pr, they pass

yehudasa:
@cbodley should also test the multisite suite

cbodley:
passes test_multi.py (requires fixes from https://github.com/ceph/ceph/pull/12229 and https://github.com/ceph/ceph/pull/12105)

yehudasa:
@cbodley need to get this one through teuthology too

cbodley:
@yehudasa success http://pulpito.ceph.com/cbodley-2017-01-17_15:16:18-rgw-wip-cbodley-testing---basic-mira

PR #11588, opened 2016-10-21
osd: fix duplicated id of incompat feature "fastinfo"

xiexingguo:


liewegas:
good catch, thanks!

PR #11589, opened 2016-10-21
kstore: added DB sharding for rocksdb to store data.

Ted-Chang:
kstore: added interfaces in KeyValueDB for DB sharding and implemented them in rocksdb with rocksdb column family.


yuyuyu101:
I think this is the last way to resolve rocksdb performance, so this pr may need to wait a lot

Ted-Chang:
I did a simple test for this implementation and for 4K small IO rados bench performance increase by about 20 percent in my test environment with 4 rocksdb shards. This is a benign feature if you didn't turn it on. It only take effect if you set kstore_num_shards great than 1. Do we have a check-in plan or timeline for this patch? Thanks a lot.

liewegas:
This seems like a last resort.  Do you understand why this is faster?  Lock contention in rocksdb or something?

It looks like this is only sharing the object data?

I do'nt think we want to carry this weirdness in teh generic KeyValueDB interface until we *really* have to...

PR #11592, opened 2016-10-21
rbd-mirror HA: create a generic/re-usable managed lock class

rjfd:
Fixes: http://tracker.ceph.com/issues/17016

Signed-off-by: Ricardo Dias rdias@suse.com


dillaman:
@rjfd I spent some time thinking today and I don't think we want the generic ```LockWatcher``` -- the rbd-mirror lock messages are going to be different (more of announcements instead of requests) and they won't require the ```ClientId``` kludge like the ```ImageWatcher``` notification types (since librbd couldn't control whether or not each image would be using a unique librados instance).

src/librbd/Lock.cc
```diff
@@ -0,0 +1,620 @@
...
+      execute_action(ACTION_TRY_LOCK, on_tried_lock);
+      return;
+    }
+  }
+
+  on_tried_lock->complete(r);
+}
+
+template <typename L>
+void Lock<L>::request_lock(Context *on_locked) {
```
dillaman: I think this can be moved back to ```ExclusiveLock``` since ```rbd-mirror``` won't require it.

src/librbd/Lock.cc
```diff
@@ -0,0 +1,620 @@
...
+  on_released->complete(r);
+}
+
+template <typename L>
+void Lock<L>::reacquire_lock(Context *on_reacquired) {
+  {
+    Mutex::Locker locker(m_lock);
+
+    if (!is_shutdown_locked() &&
+               (m_state == STATE_LOCKED ||
```
dillaman: Minor: spacing looks odd in the review tool -- not sure if this is indented past ```!is_shutdown_locked()```, but if it is, it should be moved back

src/librbd/Lock.cc
```diff
@@ -0,0 +1,620 @@
...
+  }
+
+  {
+    m_lock.Lock();
+    assert(m_state == STATE_ACQUIRING);
+
+    Action action = get_active_action();
+    assert(action == ACTION_TRY_LOCK || action == ACTION_REQUEST_LOCK);
+    if (action == ACTION_REQUEST_LOCK && r < 0 && r != -EBLACKLISTED &&
+        r != -EPERM) {
```
dillaman: I think the ```-EPERM``` was an error that would only be returned by the journal policy, so that check should be moved up to ```ExclusiveLock```

src/librbd/Lock.cc
```diff
@@ -0,0 +1,620 @@
...
+
+  // expect to be destroyed after firing callback
+  for (auto ctx : action_contexts.second) {
+    ctx->complete(r);
+  }
+}
+
+} // namespace librbd
+
+#include "librbd/managed_lock/LockWatcher.h"
```
dillaman: Minor: wrong spot for #includes

src/librbd/Lock.h
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+private:
+
+  /**
+   * @verbatim
+   *
+   *       <start>
+   *          |
+   *          |
```
dillaman: Minor: missing WAITING_FOR_REGISTER state

src/librbd/exclusive_lock/AcquireRequest.h
```diff
@@ -14,15 +14,21 @@ class Context;
 
 namespace librbd {
 
+namespace managed_lock{
+class LockWatcher;
+}
+
 template <typename> class Journal;
+template <typename> class Lock;
+typedef Lock<librbd::managed_lock::LockWatcher> LockT;
```
dillaman: Minor: move this into the class instead of polluting the librbd namespace. Of course, might not be needed anymore, either.

src/librbd/exclusive_lock/ReleaseRequest.h
```diff
@@ -11,17 +11,22 @@ class Context;
...
 namespace librbd {
 
-class ImageCtx;
+namespace managed_lock{
+class LockWatcher;
+}
+
+struct ImageCtx;
+template <typename> class Lock;
+typedef Lock<librbd::managed_lock::LockWatcher> LockT;
```
dillaman: Minor: same comment here

src/librbd/internal.cc
```diff
@@ -1528,7 +1528,8 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force,
         ictx->state->close();
         return r;
       }
-      if (watchers.size() > 1) {
+      if ((ictx->exclusive_lock != nullptr && watchers.size() > 2) ||
```
dillaman: Definitely don't want to have two watchers per image per librbd client

src/librbd/Lock.cc
```diff
@@ -0,0 +1,620 @@
...
+template <typename L>
+Lock<L>::Lock(librados::IoCtx &ioctx, const string& oid,
+           Policy *policy)
+  : m_ioctx(ioctx), m_cct(reinterpret_cast<CephContext *>(ioctx.cct())),
+    m_oid(oid),
+    m_policy(policy),
+    m_lock(util::unique_lock_name("librbd::Lock<L>::m_lock", this)),
+    m_state(STATE_UNLOCKED) {
+
+  if (m_policy == nullptr) {
```
dillaman: ... and coming back to this now that I've seen how it's used, I think the policy can be scrapped at this level and be moved to back to ```ImageWatcher``` / ```ExclusiveLock```-only.

src/librbd/managed_lock/AcquireRequest.cc
```diff
@@ -0,0 +1,332 @@
...
+  m_on_finish->complete(m_error_result);
+  delete this;
+}
+
+} // namespace managed_lock
+} // namespace librbd
+
+#include "librbd/managed_lock/LockWatcher.h"
+template class librbd::managed_lock::AcquireRequest<
+                                            librbd::managed_lock::LockWatcher>;
```
dillaman: Minor: I would just recommend a single indent when you run into overflow cases instead of as far indented as possible. General comment that applies everywhere.

src/librbd/managed_lock/NotifyLockOwner.cc
```diff
@@ -0,0 +1,105 @@
+// -*- mode:C++; tab-width:8; c-basic-offset:2; indent-tabs-mode:t -*-
```
dillaman: Minor: I think this is a good generic helper for ```ObjectWatcher```, but not necessarily tied into the lock

src/librbd/managed_lock/Policy.h
```diff
@@ -0,0 +1,20 @@
+// -*- mode:C++; tab-width:8; c-basic-offset:2; indent-tabs-mode:t -*-
```
dillaman: I think the policy is tied to ```ExclusiveLock``` / ```ImageWatcher```

src/librbd/managed_lock/LockWatcher.h
```diff
@@ -0,0 +1,85 @@
+// -*- mode:C++; tab-width:8; c-basic-offset:2; indent-tabs-mode:t -*-
+// vim: ts=8 sw=2 smarttab
+
+#ifndef CEPH_LIBRBD_LOCK_WATCHER_H
```
dillaman: Minor: here and throughout the ```managed_lock``` namespace -- change "LOCK" to "MANAGED_LOCK"

src/librbd/watcher/RewatchRequest.h
```diff
@@ -0,0 +1,75 @@
+// -*- mode:C++; tab-width:8; c-basic-offset:2; indent-tabs-mode:t -*-
```
dillaman: Move to ```object_watcher``` namespace

rjfd:
@dillaman during the implementation of the 'Lock' class I also thought that the 'LockWatcher' class would not be necessary and instead use the 'ImageWatcher' class as the template parameter of the 'Lock' class. This way we keep only one watcher per image. What do you think?

Regarding the 'Watcher' class, the idea was always to improve the 'ObjectWatcher' class, either by substituting it, or merging it with the 'Watcher' class. I'll do this in future commits by updating the code that uses the 'ObjectWatcher' class. I also plan to make the 'ImageWatcher' class to extend the new 'ObjectWatcher' class implementation.

src/librbd/Lock.cc
```diff
@@ -0,0 +1,620 @@
...
+struct C_SendRequest : public Context {
+  R* request;
+  explicit C_SendRequest(R* request) : request(request) {
+  }
+  virtual void finish(int r) override {
+    request->send();
+  }
+};
+
+class ThreadPoolSingleton : public ThreadPool {
```
rjfd: The reason of why I used a new thread pool and consequently a new 'ContextWQ' was because the 'op_work_queue' from the 'ImageCTx' class is being mocked in the unit tests and I didn't want to make the 'Lock' class to depend on 'ImageCtx' as a template parameter of 'Lock'.

dillaman:
@rjfd I think the lock support should be separate from the RPC support since they might not necessarily be tied together. The ```ManagedLock``` class should only handle acquiring/releasing the lock (with all the extras for breaking the lock if the current owner is dead).

src/librbd/Lock.cc
dillaman: Do any of the test cases really care / test how Contexts get queued into the ContextWQ? If not, you can use the ```get_image_ctx``` helper from Utils.h to get the non-mocked version.

rjfd:
@dillaman Understood -- to make it clear, the ManagedLock should not have any dependency with any Watcher, right? It will just provide some callbacks at important points.

dillaman:
@rjfd In general, yes. I think the extent of the relationship between ```ManagedLock``` and ```ObjectWatcher``` should be about handling the watch handle. For other things (like if I lost the lock due a lost watch and I couldn't reacquire), either callbacks or (if using inheritance) invoke some virtual method to kick the derived class to do what it needs to do.

src/librbd/ExclusiveLock.h
```diff
@@ -94,13 +95,8 @@ class ExclusiveLock {
     STATE_LOCKED,
     STATE_INITIALIZING,
     STATE_ACQUIRING,
-    STATE_POST_ACQUIRING,
```
rjfd: I'm still not convinced that the _PRE/_POST states are required. Regarding the release of the exclusive-lock, in the current implementation, the managed-lock is only released after the object_map and journal are closed.
Regarding the acquire of the exclusive-lock, the managed-lock is acquired before the object-map and journal are opened. The only problem I found was with the RefreshRequest that would see that the managed-lock was already acquired (although the exclusive-lock was still in m_state == STATE_ACQUIRING) , and would try to open the object-map and journal, which conflicted with the open of object-map and journal in the AcquireRequest class.

src/librbd/ExclusiveLock.cc
```diff
@@ -206,19 +184,9 @@ void ExclusiveLock<I>::reacquire_lock(Context *on_reacquired) {
     Mutex::Locker locker(m_lock);
     assert(m_image_ctx.owner_lock.is_locked());
 
-    if (m_state == STATE_WAITING_FOR_REGISTER) {
```
rjfd: Doh -- forget this, since LockWather will disappear and ManagedLock will have very few interactions with a Watcher object. Will fix this at the ExclusiveLock class

rjfd:
@dillaman finished the changes to address your comments and the last discussions. I think the PR is ready for a second review iteration

rjfd:
@dillaman Now that I'm thinking, maybe it's worthwhile to split this PR into two: 1) the generalization of object watcher and several Watcher classes refactoring; 2) generic managed lock class implementation

dillaman:
@rjfd sounds like a good plan

dachary:
needs rebase :-)

rjfd:
@dillaman rebased the PR on top of yesterdays commits. I'm just missing a final commit with the implementation of a `request_shared_lock` in the generic lock implementation, but I think everything else is ready for review.

src/librbd/ExclusiveLock.cc
```diff
@@ -528,44 +470,6 @@ void ExclusiveLock<I>::handle_reacquire_lock(int r) {
...
-            });
-          execute_action(ACTION_REQUEST_LOCK, ctx);
-        }
-      }
-    }
-  } else {
-    m_cookie = m_new_cookie;
-  }
 
   complete_active_action(STATE_LOCKED, 0);
```
dillaman: If the re-acquire fails, you'd still need to execute the release / reacquire steps above

src/librbd/ManagedLock.cc
```diff
@@ -0,0 +1,501 @@
...
+    } else {
+      lderr(m_cct) << this << " " << __func__ << ": "
+                 << "failed to update lock cookie: " << cpp_strerror(r)
+                 << dendl;
+    }
+
+    if (!is_shutdown_locked()) {
+      // queue a release and re-acquire of the lock since cookie cannot
+      // be updated on older OSDs
+      execute_action(ACTION_RELEASE_LOCK, nullptr);
```
dillaman: This needs to happen at the upper layer (ExclusiveLock) since the release step involves shutting down the journaler / object map. If you used ManagedLock as a parent class of ExclusiveLock, you could have the necessary hooks via virtual methods to start the release/reacquire process

src/librbd/ManagedLock.h
```diff
@@ -0,0 +1,184 @@
...
+  void shut_down(Context *on_shutdown);
+
+  void acquire_lock(Context *on_acquired);
+  void release_lock(Context *on_released);
+
+  void reacquire_lock(Context *on_reacquired = nullptr);
+
+  void assert_locked(librados::ObjectWriteOperation *op, ClsLockType type);
+
+  inline librados::IoCtx& io_ctx() const {
```
dillaman: I'd vote to yank these accessors and instead directly pass the necessary values to the state machine instances -- which can still be templated on an (unused) ```ImageCtx``` instead of ```ManagedLock<DerivedLock>```

src/librbd/ManagedLock.cc
```diff
@@ -0,0 +1,501 @@
...
+    ldout(m_cct, 5) << "unable to acquire exclusive lock" << dendl;
+  } else if (r < 0) {
+    lderr(m_cct) << "failed to acquire exclusive lock:" << cpp_strerror(r)
+               << dendl;
+  } else {
+    ldout(m_cct, 5) << "successfully acquired exclusive lock" << dendl;
+  }
+
+  State next_state = (r < 0 ? STATE_UNLOCKED : STATE_LOCKED);
+  if (r == -EAGAIN) {
```
dillaman: I would pass this error down to ```ExclusiveLock``` since it has the logic to (potentially) request the lock from a peer if it's in-use by an alive peer. This state machine should be interlocked w/ ```ExclusiveLock``` anyway so ```ExclusiveLock::handle_acquire_lock``` shouldn't need to check whether or not the lock was acquired since the error code will be the status.

src/librbd/managed_lock/ReacquireRequest.h
```diff
@@ -1,32 +1,32 @@
 // -*- mode:C++; tab-width:8; c-basic-offset:2; indent-tabs-mode:t -*-
 // vim: ts=8 sw=2 smarttab
 
-#ifndef CEPH_LIBRBD_EXCLUSIVE_LOCK_REACQUIRE_REQUEST_H
-#define CEPH_LIBRBD_EXCLUSIVE_LOCK_REACQUIRE_REQUEST_H
+#ifndef CEPH_LIBRBD_LOCK_REACQUIRE_REQUEST_H
```
dillaman: Minor: ```MANAGED_LOCK```

src/librbd/managed_lock/ReleaseRequest.h
```diff
@@ -0,0 +1,61 @@
+// -*- mode:C++; tab-width:8; c-basic-offset:2; indent-tabs-mode:t -*-
+// vim: ts=8 sw=2 smarttab
+
+#ifndef CEPH_LIBRBD_LOCK_RELEASE_REQUEST_H
```
dillaman: Minor: ```MANAGED_LOCK```

dachary:
jenkins test this please (eio)

dachary:

src/librbd/ManagedLock.h
```diff
@@ -0,0 +1,184 @@
...
+#include "common/Mutex.h"
+#include <list>
+#include <string>
+#include <utility>
+
+class ContextWQ;
+
+namespace librbd {
+
+template <typename Watcher>
```
rjfd: I tried to change the template to ImageCtx, and use the Watcher base class but it required so many changes to the unit tests that I don't think it's worth it. The main problem is that by using the Watcher base class without template it, mocking the watcher in very hard, as it required mocking using inheritance.

rjfd: I think I'm not following what you said. The `librbd::Watcher` class is not templated and therefore I cannot create a specialization in the unit tests.
Are you suggesting to also template the `Watcher` class on `ImageCtx` to easily create mocks of the Watcher class?

dillaman: Sorry, didn't realize it wasn't templated. Looking at this again, the only reason you appear to pass the ```Watcher``` is for the assert that needs to be removed and to get the watch handle on acquire/reacquire.  You already need to do some refactoring to support reacquire, so if you go w/ inheritance to ```ManagedLock```, you could add a pure virtual method to retrieve the watch handle (via ```ExclusiveLock```) -- or ```ExclusiveLock``` could call a set function to provide the handle prior to acquire / reacquire.

rjfd:
@dillaman rebased on top of todays commits. I think it's ready for review. Thanks

src/librbd/ManagedLock.h
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  ManagedLock(librados::IoCtx& ioctx, ContextWQ *work_queue,
+              const std::string& oid, Watcher *watcher);
+  virtual ~ManagedLock();
+
+  bool is_lock_owner() const;
+
+  void shut_down(Context *on_shutdown);
+  void acquire_lock(Context *on_acquired, bool try_lock = false);
```
dillaman: Minor: really just my personal preference, but I really dislike defaulted variables. I think it would be clearer to have a ```try_acquire_lock``` method and a ```acquire_lock``` method since you need to differentiate the states.

src/librbd/ManagedLock.h
```diff
@@ -0,0 +1,181 @@
...
+    STATE_LOCKED,
+    STATE_ACQUIRING,
+    STATE_WAITING_FOR_REGISTER,
+    STATE_REACQUIRING,
+    STATE_RELEASING,
+    STATE_PRE_SHUTTING_DOWN,
+    STATE_SHUTTING_DOWN,
+    STATE_SHUTDOWN,
+
+    // ExclusiveLock Specific
```
dillaman: I think all of these states could be made generic enough so that you don't need to call out that these are exclusive lock states oddly embedded in the base class. Derived classes can provided a ```initialization_handler``` virtual method, ```post_acquiring``` and ```pre_releasing```, and rename waiting for peer to ```waiting_for_lock``` (i.e. rbd-mirror might want to "wait for the lock" by starting a timer to re-attempt the acquire).

src/librbd/ManagedLock.cc
```diff
@@ -0,0 +1,588 @@
...
+  }
+}
+
+template <typename I>
+void ManagedLock<I>::shutdown_handler(int r, Context *on_finish) {
+  on_finish->complete(r);
+}
+
+template <typename I>
+void  ManagedLock<I>::pre_acquire_lock_handler(Context *on_finish) {
```
dillaman: Nit: extra space between ```void``` and class name here and below

src/librbd/ManagedLock.cc
```diff
@@ -0,0 +1,588 @@
...
+  Mutex::Locker locker(m_lock);
+  post_acquire_lock_handler(r, new FunctionContext([this, next_state](int r) {
+    assert(m_lock.is_locked());
+
+    if (r < 0 && next_state == STATE_LOCKED) {
+      // release_lock without calling pre and post callbacks
+      using managed_lock::ReleaseRequest;
+      ReleaseRequest<I>* req = ReleaseRequest<I>::create(m_ioctx, m_watcher,
+          m_work_queue, m_oid, m_cookie,
+          new FunctionContext([this, r](int ret) {
```
dillaman: These lambdas are getting pretty large and complex -- I think, in general, it would read cleaner to just create helper methods that ```create_context_callback``` can just wrap.

src/librbd/ManagedLock.cc
dillaman: Has this been addressed somewhere?

src/librbd/ManagedLock.cc
```diff
@@ -0,0 +1,588 @@
...
+
+  Mutex::Locker locker(m_lock);
+  std::string cookie = m_cookie;
+
+  pre_release_lock_handler(true, new FunctionContext([this, cookie](int r) {
+    ldout(m_cct, 10) << ": r=" << r << dendl;
+    using managed_lock::ReleaseRequest;
+    ReleaseRequest<I>* req = ReleaseRequest<I>::create(m_ioctx, m_watcher,
+        m_work_queue, m_oid, cookie,
+        new FunctionContext([this](int r) {
```
dillaman: Same comment here -- these embedded lambdas are not as clear to read.

src/librbd/ManagedLock.cc
rjfd: In the new implementation, the acquire/release virtual handlers will be called upon the processing of `ACTION_TRY_LOCK`, `ACTION_ACQUIRE_LOCK`, and `ACTION_RELEASE_LOCK`, and therefore the reacquire logic can be stored in the `ManagedLock` class

rjfd:
@dillaman embedded lambdas in `ExclusiveLock` class are also pretty complex and big, I'm going to follow your suggestion and also turned them into helper methods

src/librbd/ManagedLock.h
```diff
@@ -0,0 +1,190 @@
...
+
+  bool is_lock_owner() const;
+
+  void shut_down(Context *on_shutdown);
+  void acquire_lock(Context *on_acquired);
+  void try_acquire_lock(Context *on_acquired);
+  void release_lock(Context *on_released);
+  void reacquire_lock(Context *on_reacquired = nullptr);
+
+  virtual void shutdown_handler(int r, Context *on_finish);
```
dillaman: Minor: these functions should move to the protected section

src/librbd/ManagedLock.h
```diff
@@ -0,0 +1,190 @@
...
+    return is_shutdown_locked();
+  }
+
+  bool is_locked_state() const {
+    return m_state == STATE_LOCKED;
+  }
+
+  static bool decode_lock_cookie(const std::string &tag, uint64_t *handle);
+
+protected:
```
dillaman: Minor: most of the declarations below should move to a private section

src/test/librbd/mock/MockWatcher.h
```diff
@@ -0,0 +1,33 @@
...
+#define CEPH_TEST_LIBRBD_MOCK_WATCHER_H
+
+#include "librbd/Watcher.h"
+#include "gmock/gmock.h"
+
+class Context;
+
+namespace librbd {
+
+struct MockWatcher : public Watcher {
```
dillaman: I'd prefer to keep the mocks as a clean, independent implementation (e.g. it doesn't half bootstrap a base class and override virtual methods). If you look at "journal/TypeTraits.h", you can see an example of how to derive the correct ```Watcher``` class based upon the type of ```ImageCtxT``` template parameter. You could add a ```librbd::watcher::Traits``` class in watcher/Types.h to accomplish the same thing.

src/librbd/ManagedLock.h
```diff
@@ -0,0 +1,190 @@
...
+public:
+  static const std::string WATCHER_LOCK_TAG;
+
+  static ManagedLock *create(librados::IoCtx& ioctx, ContextWQ *work_queue,
+                             const std::string& oid, Watcher *watcher) {
+    return new ManagedLock(ioctx, work_queue, oid, watcher);
+  }
+
+  ManagedLock(librados::IoCtx& ioctx, ContextWQ *work_queue,
+              const std::string& oid, Watcher *watcher);
```
dillaman: See comment below re: use of templated type traits struct to derive the correct "Watcher" class (```Watcher``` for API code, ```MockWatcher``` for test code).

Example: https://github.com/ceph/ceph/blob/master/src/librbd/Journal.h#L180

src/librbd/ManagedLock.cc
```diff
@@ -0,0 +1,661 @@
...
+    m_oid(oid),
+    m_watcher(watcher),
+    m_lock(util::unique_lock_name("librbd::ManagedLock<I>::m_lock", this)),
+    m_state(STATE_UNLOCKED) {
+}
+
+template <typename I>
+ManagedLock<I>::~ManagedLock() {
+  Mutex::Locker locker(m_lock);
+  assert(m_state == STATE_SHUTDOWN || m_state == STATE_UNLOCKED);
```
dillaman: Minor: also UNINITIALIZED?

src/librbd/ManagedLock.cc
```diff
@@ -0,0 +1,661 @@
...
+
+    if (m_state == STATE_WAITING_FOR_REGISTER) {
+      // restart the acquire lock process now that watch is valid
+      ldout(m_cct, 10) << ": " << "woke up waiting acquire" << dendl;
+      Action active_action = get_active_action();
+      assert(active_action == ACTION_TRY_LOCK ||
+             active_action == ACTION_ACQUIRE_LOCK);
+      execute_next_action();
+    } else if (!is_shutdown_locked() &&
+        (m_state == STATE_LOCKED || m_state == STATE_ACQUIRING)) {
```
dillaman: I think you are missing some states here: STATE_POST_ACQUIRING, STATE_WAITING_FOR_LOCK

src/librbd/ManagedLock.cc
```diff
@@ -0,0 +1,661 @@
...
+
+template <typename I>
+void  ManagedLock<I>::post_release_lock_handler(bool shutting_down, int r,
+                                                Context *on_finish) {
+  on_finish->complete(r);
+}
+
+template <typename I>
+void ManagedLock<I>::assert_locked(librados::ObjectWriteOperation *op,
+                          ClsLockType type) {
```
dillaman: Minor: spacing issue?

src/librbd/ManagedLock.cc
```diff
@@ -0,0 +1,661 @@
...
+template <typename I>
+bool ManagedLock<I>::is_transition_state() const {
+  switch (m_state) {
+  case STATE_ACQUIRING:
+  case STATE_WAITING_FOR_REGISTER:
+  case STATE_REACQUIRING:
+  case STATE_RELEASING:
+  case STATE_PRE_SHUTTING_DOWN:
+  case STATE_SHUTTING_DOWN:
+
```
dillaman: Minor: why the added line breaks here and below?

src/librbd/ManagedLock.cc
```diff
@@ -0,0 +1,661 @@
...
+  assert(m_lock.is_locked());
+  if (m_state == STATE_LOCKED) {
+    complete_active_action(STATE_LOCKED, 0);
+    return;
+  }
+
+  ldout(m_cct, 10) << dendl;
+  m_state = STATE_ACQUIRING;
+
+  if (!m_watcher->is_registered()) {
```
dillaman: Minor: potential race here between ```is_registered``` call and ```get_watch_handle``` call. The original logic retrieved the watch handle and if it was zero it entered the WAITING_FOR_REGISTER state.

src/librbd/ManagedLock.cc
```diff
@@ -0,0 +1,661 @@
...
+  uint64_t watch_handle = m_watcher->get_watch_handle();
+  m_cookie = ManagedLock<I>::encode_lock_cookie(watch_handle);
+
+  pre_acquire_lock_handler(util::create_context_callback<
+      ManagedLock<I>, &ManagedLock<I>::handle_pre_acquire_lock>(this));
+}
+
+template <typename I>
+void ManagedLock<I>::handle_pre_acquire_lock(int r) {
+  ldout(m_cct, 10) << ": r=" << r << dendl;
```
dillaman: Minor: if "r < 0", perhaps you should just invoke ```handle_acquire_lock``` so that it can clean up.

src/librbd/ManagedLock.cc
dillaman: Ah -- yes, thanks.

src/librbd/ManagedLock.cc
```diff
@@ -0,0 +1,661 @@
...
+  m_state = STATE_RELEASING;
+
+  pre_release_lock_handler(false, util::create_context_callback<
+      ManagedLock<I>, &ManagedLock<I>::handle_pre_release_lock>(this));
+}
+
+template <typename I>
+void ManagedLock<I>::handle_pre_release_lock(int r) {
+  ldout(m_cct, 10) << ": r=" << r << dendl;
+
```
dillaman: Minor: if "r < 0", perhaps you should just invoke ```handle_release_lock``` so that it can clean up.

src/librbd/ManagedLock.cc
```diff
@@ -0,0 +1,661 @@
...
+  post_release_lock_handler(false, r, util::create_context_callback<
+      ManagedLock<I>, &ManagedLock<I>::handle_post_release_lock>(this));
+}
+
+template <typename I>
+void ManagedLock<I>::handle_post_release_lock(int r) {
+  ldout(m_cct, 10) << ": r=" << r << dendl;
+
+  assert(m_lock.is_locked());
+  complete_active_action(r < 0 ? STATE_LOCKED : STATE_UNLOCKED, r);
```
dillaman: If the failure occurred at the derived class handler -- the lock might have been released but you revert to the LOCKED state.

src/test/librbd/mock/MockWatcher.h
rjfd: Thanks! Until now I didn't know why the journal TypeTraits was needed. Now I understand :-)

rjfd:
@dillaman pushed new changes.

src/librbd/ManagedLock.cc
dillaman: Was this change not needed for some reason?

src/librbd/ManagedLock.cc
```diff
@@ -0,0 +1,672 @@
...
+
+  using managed_lock::ReleaseRequest;
+  ReleaseRequest<I>* req = ReleaseRequest<I>::create(m_ioctx, m_watcher,
+      m_work_queue, m_oid, m_cookie,
+      new FunctionContext([this, r](int ret) {
+        Mutex::Locker locker(m_lock);
+        if (ret >= 0) {
+          complete_active_action(STATE_UNLOCKED, r);
+        } else {
+          ldout(m_cct, 10) << ": failed to revert to unlocked state:"
```
dillaman: What are the valid failure cases here? If we get ENOENT / EBLACKLISTED, I think it's safe to assume we are unlocked (probably could mask those errors within the ReleaseRequest state machine to simplify things). Otherwise, I think we should still treat an error as we are unlocked -- especially since the derived class has already thrown an error so it isn't set up for a locked state.

src/librbd/ManagedLock.cc
```diff
@@ -0,0 +1,672 @@
...
+}
+
+template <typename I>
+void ManagedLock<I>::shutdown_handler(int r, Context *on_finish) {
+  on_finish->complete(r);
+}
+
+template <typename I>
+void ManagedLock<I>::pre_acquire_lock_handler(Context *on_finish) {
+  m_work_queue->queue(on_finish);
```
dillaman: I would feel better if it were possible to invoke all the ```xyz_handler``` methods without holding ```m_lock```. It would make for a consistent calling convention that the handlers are always invoked asynchronously and can invoke the provided ```on_finish``` w/o having to worry about lock ordering / recursive locking.

src/librbd/ManagedLock.cc
dillaman: .... looks like it is hardcoded to return 0, so I'd assert if r != 0 and remove the failure path.

src/librbd/ExclusiveLock.cc
```diff
@@ -2,126 +2,87 @@
...
-    m_state(STATE_UNINITIALIZED), m_watch_handle(0) {
+  : ML<I>(image_ctx.md_ctx, image_ctx.op_work_queue, image_ctx.header_oid,
+          image_ctx.image_watcher),
+    m_image_ctx(image_ctx), m_pre_post_callback(nullptr),
+    m_shutting_down(false)  {
+  ML<I>::m_state = ML<I>::STATE_UNINITIALIZED;
 }
 
 template <typename I>
 ExclusiveLock<I>::~ExclusiveLock() {
```
dillaman: Minor: is this destructor needed anymore?

src/librbd/ExclusiveLock.cc
```diff
@@ -132,610 +93,242 @@ void ExclusiveLock<I>::init(uint64_t features, Context *on_init) {
...
-    execute_action(ACTION_SHUT_DOWN, on_shut_down);
-  }
+  ML<I>::shut_down(on_shut_down);
 
   // if stalled in request state machine -- abort
   handle_peer_notification();
 }
 
 template <typename I>
 void ExclusiveLock<I>::try_lock(Context *on_tried_lock) {
```
dillaman: Minor: it would be great, while you are in here, to just remove the ```try_lock``` method and '''request_lock``` method and modify librbd to use ```try_acquire_lock``` and ```acquire_lock``` for simplicity / clarity.

src/librbd/ExclusiveLock.h
```diff
@@ -40,15 +30,12 @@ class ExclusiveLock {
   void try_lock(Context *on_tried_lock);
   void request_lock(Context *on_locked);
   void release_lock(Context *on_released);
```
dillaman: Minor: no need for these pass-through methods -- just let the ```ManagedLock``` version be exposed

src/librbd/ManagedLock.cc
rjfd: I forgot to address this, thanks

rjfd:
@dillaman addressed your comments and pushed the changes.

PR #11596, opened 2016-10-21
mon/OSDMonitor: encode OSDMap::Incremental with same features as OSDMap

liewegas:
The Incremental encode stashes encode_features, which is
what we use later to reencode the updated OSDMap.  Use
the same features so that the encoding will match!

### DIFF
--- a/src/mon/OSDMonitor.cc
+++ b/src/mon/OSDMonitor.cc
@@ -1220,6 +1220,9 @@ void OSDMonitor::encode_pending(MonitorDBStore::TransactionRef t)
     }
   }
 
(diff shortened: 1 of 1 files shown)


liewegas:
http://pulpito.ceph.com/sage-2016-10-24_00:49:43-upgrade:jewel-x-wip-sage-testing2---basic-smithi/

tchaikov:
adding this PR to http://tracker.ceph.com/issues/17386, as it should be backported to jewel as well, i believe.

tchaikov:
lgtm.

PR #11600, opened 2016-10-21
ceph-disk: missing return when a file is used for journal

jayaajay:
http://tracker.ceph.com/issues/17662

ceps-disk is missing a return statement when preparing the journal space when it is regular file.

### DIFF
--- a/src/ceph-disk/ceph_disk/main.py
+++ b/src/ceph-disk/ceph_disk/main.py
@@ -1927,6 +1927,7 @@ class PrepareSpace(object):
                 raise Error('%s is not a block device' % name.capitalize,
                             getattr(args, name))
             self.type = self.FILE
+            return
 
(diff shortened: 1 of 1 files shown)


dachary:
good catch !

dachary:
I'll add a test for that so it stays fixed.

dachary:
@jayaajay I amended your commit to add tests and fix another issue, could you please review it ? https://github.com/ceph/ceph/pull/11619

jayaajay:
Done!

On Oct 21, 2016, at 2:06 PM, Casey Bodley <notifications@github.com<mailto:notifications@github.com>> wrote:


thank you! could you please add Fixes: http://tracker.ceph.com/issues/17662 to your commit message? we also need a Signed-off-by: line (you can git commit --amend -s)

PR #11610, opened 2016-10-22
messages/MForward: reencode forwarded message if target has differing features

liewegas:
This ensures we reencode the payload with the
appropriate set of features if the client, us, or the
target do not have identical features.  Otherwise we
may forward an encoding with more features than the
target can handle.

### DIFF
--- a/src/messages/MForward.h
+++ b/src/messages/MForward.h
@@ -77,6 +77,9 @@ public:
     // message are changed when reencoding with more features than the
     // client had originally.  That should never happen, but we may as
     // well be defensive here.
+    if (con_features != features) {
+      msg->clear_payload();
+    }
     encode_message(msg, features & con_features, payload);
     ::encode(con_features, payload);
     ::encode(entity_name, payload);
--- a/src/msg/Message.cc
+++ b/src/msg/Message.cc
@@ -190,7 +190,12 @@ void Message::encode(uint64_t features, int crcflags)
 {
   // encode and copy out of *m
   if (empty_payload()) {
+    assert(middle.length() == 0);
     encode_payload(features);
+
+    if (byte_throttler) {
+      byte_throttler->take(payload.length() + middle.length());
+    }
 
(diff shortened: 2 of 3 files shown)


liewegas:
http://pulpito.ceph.com/sage-2016-10-24_00:49:43-upgrade:jewel-x-wip-sage-testing2---basic-smithi/

tchaikov:
adding this to http://tracker.ceph.com/issues/17386 also, it should be backported to jewel along with #11596.

src/messages/MForward.h
```diff
@@ -77,6 +77,10 @@ struct MForward : public Message {
     // message are changed when reencoding with more features than the
     // client had originally.  That should never happen, but we may as
     // well be defensive here.
+    if (CEPH_FEATURES_SUPPORTED_DEFAULT != features ||
```
tchaikov: assuming we have three parties: osd.0 <=> mon.1 <=> mon.0. in which, osd.0 is connected to mon.1, which is a peon, and mon.0 is the leader of the quorum. 

1. osd.0 sends a request to mon.1, and mon.1 **forwards** it to mon.0

> the `features` here is `mon[0].features & mon[1].features`. and the `msg->payload` is encoded using `con_features` (i.e. `osd[0].features & mon[1].features`).
> in this patch we re-encode the request before forwarding it to mon.0. i think should always re-encode the wrapped request as long as `con_features != features`, regardless if `mon[0].features` is equal to `mon[1].feature` or not.

2. mon.0 sends the reply of this message to mon.1, and mon.1 **routes** it to osd.0

> ahh, we always `clear_payload()`, okay then.

tchaikov: in short, we don't need to check "CEPH_FEATURES_SUPPORTED_DEFAULT != features"

tchaikov:
modulo the nit where i'd recommend remove "CEPH_FEATURES_SUPPORTED_DEFAULT != features", lgtm.

PR #11612, opened 2016-10-23
hammer: ObjectCacher doesn't correctly handle read replies on split BufferHeads

smithfarm:
http://tracker.ceph.com/issues/16546

NOTE: the original PR #9606 has 6 commits, and this (the hammer backport) has only 5. The reason is that the last commit in #9606 ( 2ee3d02 ) is cmake-specific and therefore not needed in hammer.

### DIFF
--- a/src/test/osdc/object_cacher_stress.cc
+++ b/src/test/osdc/object_cacher_stress.cc
@@ -21,6 +21,7 @@
 #include "osdc/ObjectCacher.h"
 
 #include "FakeWriteback.h"
+#include "MemWriteback.h"
 
 // XXX: Only tests default namespace
 struct op_data {
@@ -170,6 +171,24 @@ int stress_test(uint64_t num_ops, uint64_t num_objs,
   return EXIT_SUCCESS;
 }
 
+int correctness_test(uint64_t delay_ns)
+{
+  Mutex lock("object_cacher_stress::object_cacher");
+  MemWriteback writeback(g_ceph_context, &lock, delay_ns);
+
+  ObjectCacher obc(g_ceph_context, "test", writeback, lock, NULL, NULL,
+		   g_conf->client_oc_size,
+		   g_conf->client_oc_max_objects,
+		   g_conf->client_oc_max_dirty,
+		   g_conf->client_oc_target_dirty,
+		   g_conf->client_oc_max_dirty_age,
+		   true);
+  obc.start();
+
+  std::cout << "Testing ObjectCacher correctness" << std::endl;
+  return 0;
+}
+
 int main(int argc, const char **argv)
 {
   std::vector<const char*> args;
@@ -184,6 +203,8 @@ int main(int argc, const char **argv)
   long long num_objs = 10;
   float percent_reads = 0.90;
   int seed = time(0) % 100000;
+  bool stress = false;
+  bool correctness = false;
   std::ostringstream err;
   std::vector<const char*>::iterator i;
   for (i = args.begin(); i != args.end();) {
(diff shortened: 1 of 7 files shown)


src/test/osdc/object_cacher_stress.cc
```diff
@@ -173,20 +173,178 @@ int stress_test(uint64_t num_ops, uint64_t num_objs,
...
+
+  // write some 1-valued bits at 256-KB intervals for checking consistency
+  std::cerr << "Writing some 0xff values" << std::endl;
+  ceph::buffer::ptr ones(1<<16);
+  memset(ones.c_str(), 0xff, ones.length());
+  ceph::bufferlist ones_bl;
+  ones_bl.append(ones);
+  for (int i = 1<<18; i < 1<<22; i+=1<<18) {
+    ObjectCacher::OSDWrite *wr = obc.prepare_write(snapc, ones_bl,
+						   ceph::real_time::min(), 0,
```
asheplyakov: $ git grep -e 'real_time' |wc -l
0

src/test/osdc/object_cacher_stress.cc
```diff
@@ -173,20 +173,178 @@ int stress_test(uint64_t num_ops, uint64_t num_objs,
...
+  }
+  bufferlist ones_maybe;
+  ones_maybe.substr_of(readbl, (1<<18)+(1<<16), ones_bl.length());
+  assert(0 == memcmp(ones_maybe.c_str(), ones_bl.c_str(), ones_bl.length()));
+
+  std::cout << "validated that data is 0xff where it should be" << std::endl;
+  
+  lock.Lock();
+  C_SaferCond flushcond2;
+  done = obc.flush_all(&flushcond2);
```
asheplyakov: ObjectCacher::flush_all() is not in hammer yet

src/test/osdc/object_cacher_stress.cc
```diff
@@ -173,20 +173,178 @@ int stress_test(uint64_t num_ops, uint64_t num_objs,
...
+  SnapContext snapc;
+  ceph_tid_t journal_tid = 0;
+  std::string oid("correctness_test_obj");
+  ObjectCacher::ObjectSet object_set(NULL, 0, 0);
+  ceph::bufferlist zeroes_bl;
+  zeroes_bl.append_zero(1<<20);
+
+  // set up a 4MB all-zero object
+  std::cerr << "writing 4x1MB object" << std::endl;
+  std::map<int, C_SaferCond> create_finishers;
```
asheplyakov: This code is invalid in C++ 98: Context can't be copied (it can't be moved so the above works with C++11)

src/test/osdc/object_cacher_stress.cc
```diff
@@ -173,20 +173,178 @@ int stress_test(uint64_t num_ops, uint64_t num_objs,
...
+  for (int i = 0; i < 4; ++i) {
+    ObjectCacher::OSDWrite *wr = obc.prepare_write(snapc, zeroes_bl,
+						   ceph::real_time::min(), 0,
```
asheplyakov: utime_t()?

src/test/osdc/object_cacher_stress.cc
```diff
@@ -173,20 +173,178 @@ int stress_test(uint64_t num_ops, uint64_t num_objs,
...
+						   ++journal_tid);
```
asheplyakov: $ git grep -e 'journal_tid' | wc -l
0

asheplyakov:
This patchset fails to compile, just cherry-picking patches is not enough. I propose this PR https://github.com/ceph/ceph/pull/11618 instead

smithfarm:
@asheplyakov Thanks! Closing this in favor of #11618

PR #11616, opened 2016-10-24
osd: Include OSD ID in metadata data

wido:
This way the output returned can be easily JSON parsed and includes
the OSD's ID in the JSON.


wido:
Never mind, the ID already seems included. Can't see why or where I missed this.

PR #11621, opened 2016-10-24
Implement slab containers and enhance mempools.

allensamuels2:
signed-off-by: Allen Samuels allen.samuels@sandisk.com

### DIFF
--- a/src/include/slab_containers.h
+++ b/src/include/slab_containers.h
@@ -19,43 +19,60 @@
 
 namespace mempool {
 
-//
-// The slab_xxxx containers are made from standard STL containers with a custom allocator.
-//
-// When you declare a slab container you provide 1 or 2 additional integer template parameters that
-// modify the memory allocation pattern. The point is to amortize the memory allocations for nodes
-// within the container so that the memory allocation time and space overheads are reduced.
-//
-//  slab_map     <key,value,stackSize,heapSize = stackSize,compare = less<key>>
-//  slab_multimap<key,value,stackSize,heapSize = stackSize,compare = less<key>>
-//  slab_set     <value,    stackSize,heapSize = stackSize,compare = less<key>>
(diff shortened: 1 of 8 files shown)


dachary:
needs rebase :-)

liewegas:
@rzarzynski care to resurrect this?

rzarzynski:
@liewegas: sure!

allensamuels2:
If you have any questions, I’m happy to answer them to the best of my memory….

src/include/slab_containers.h
```diff
@@ -0,0 +1,728 @@
...
+  };
+  //
+  // Each slot has a pointer to it's containing slab PLUS either one Object OR if it's free exactly one pointer 
+  // in a per-slab freelist.
+  // Since this is raw memory, we don't want to declare something of type "T" to avoid
+  // accidental constructor/destructor calls.
+  //
+  struct slot_t {
+    slab_t *slab; // Pointer to my slab, NULL for slots on the stack (Technically, this could be an index but I'm lazy :))
+    enum { OBJECT_IN_POINTERS = (sizeof(T) + sizeof(void *) - 1) / sizeof(void *) };
```
rzarzynski: Maybe `std::aligned_storage` instead.

allensamuels2:
If the underlying object has an alignment restriction greater than a pointer, this code will fail miserably. I don’t know if an assert to that effect can be done or not (maybe something like: assert(alignof(T) <= sizeof(void *));    ????

The usage of the OBJECT_IN_POINTERS is done because there is an implicit union of the template type <T> with a pointer (when the slot is not in use, a pointer to the free list is stored). The ENUM’s computation guarantees that if sizeof(T) is less than sizeof(void *) you are still reserving sufficient space. So you could just make this explicit with a union -> that would probably make the code a bit clearer.

liewegas:
see @rzarzynski 's newer implementations of this

PR #11623, opened 2016-10-24
librbd: ensure consistency groups will gracefully fail on older OSDs

dillaman:
Attempting to retrieve the group spec will fail on older OSDs, so it
must be executed as an individual step in the refresh state machine.
Also fixed code style issues for out parameters.

Signed-off-by: Jason Dillaman dillaman@redhat.com


dillaman:
retest this please

trociny:
lgtm

PR #11624, opened 2016-10-24
Wip rgwfile pybind

yuyuyu101:


yuyuyu101:
@mattbenjamin would you like to give a look about the python api design?

mattbenjamin:
@yuyuyu101 absolutely, thanks for doing this

mattbenjamin:
@yuyuyu101 I'm not an expert on Python bindings, but this looks clean.   I'll ask Dan on my team (who is a Python expert) to look over as well.

@dang :)

src/pybind/rgw/rgw_file.pyx
```diff
@@ -0,0 +1,663 @@
...
+                rgw_umount(self.fs, 0);
+            self.state = "shutdown"
+
+    def __enter__(self):
+        self.mount()
+        return self
+
+    def __exit__(self, type_, value, traceback):
+        self.shutdown()
+        return False
```
tchaikov: nit, it returns `None` by default.

src/pybind/rgw/setup.py
```diff
@@ -0,0 +1,201 @@
...
+    except LinkError:
+        print('\nLink Error: RGW library not found', file=sys.stderr)
+        return False
+    else:
+        return True
+    finally:
+        shutil.rmtree(tmp_dir)
+
+
+if 'BUILD_DOC' in os.environ.keys():
```
tchaikov: @yuyuyu101 could you update `doc/api/index.rst` to include the rgw pybindings API references? you can follow the example of `doc/rbd/librbdpy.rst` by adding a `doc/radosgw/api.rst`, which lists some important classes. and update `admin/build-doc` accordingly. if you have no intention to update the doc, i can help on it after your change gets merged.

yuyuyu101:
@tchaikov sure, let's do doc things after merging in case of we need to modify api

@dang do you have time recently to review this? I want to make this merge at K release

src/pybind/rgw/setup.py
```diff
@@ -0,0 +1,201 @@
...
+
+        compiler.set_library_dirs([os.environ.get('CEPH_LIBDIR')])
+
+    try:
+        compiler.define_macro('_FILE_OFFSET_BITS', '64')
+
+        link_objects = compiler.compile(
+            sources=[tmp_file],
+            output_dir=tmp_dir,
+            extra_preargs=['-iquote{path}'.format(path=os.path.join(CEPH_SRC_DIR, 'include'))]
```
dang: RBD does this by usind compiler_set_include_dirs() above by the library_dirs.  This seems like a cleaner (and more maintainable) solution, and my test build worked.  Could we switch to that?

dmick:
This breaks the build on Trusty, perhaps the older version of cython (trusty: 0.20.1 vs xenial: 0.23.4).  Don't know where CentOS stands.

Error compiling Cython file:
------------------------------------------------------------
...
                'f_bfree': statbuf.f_bfree,
                'f_bavail': statbuf.f_bavail,
                'f_files': statbuf.f_files,
                'f_ffree': statbuf.f_ffree,
                'f_favail': statbuf.f_favail,
                'f_fsid': statbuf.f_fsid,
                                ^
------------------------------------------------------------

rgw_file.pyx:408:33: Cannot convert 'uint64_t [2]' to Python object

voidbag:
I have the same problem

My cython version is also 0.20.1 on Ubuntu

PR #11627, opened 2016-10-24
jewel: rgw: work around curl_multi_wait bug with non-blocking reads

cbodley:
pulls in some earlier fixes to resolve merge conflicts

Backport tracker: http://tracker.ceph.com/issues/17343
http://tracker.ceph.com/issues/17846

PR #11630, opened 2016-10-24
rgw: remove suggestion to upgrade libcurl

cbodley:
@ktdreyer pointed out on the ceph-users list that it's a bad idea to ask users to maintain their own packages for curl


cbodley:
@oritwas could you please review? it's blocking the backport of #11627

PR #11632, opened 2016-10-24
rbd: fix parsing of group and image specific pools

VictorDenisov:
Signed-off-by: Victor Denisov <denisovenator@gmail.com>


VictorDenisov:
@dillaman this is a quick fix for already merged group image add command.

PR #11641, opened 2016-10-25
librbd: assert failure when using data pool

vshankar:
This fixes a silly assert that's hit during image creation
(cli/api) when the data pool specified is same as the pool
specified by -p/--pool option (or the default).

Signed-off-by: Venky Shankar <vshankar@redhat.com>


dillaman:
@vshankar Actually, I'd rather have the constructor set the data pool id to -1 and clear the data pool feature bit if the pools are the same.

PR #11647, opened 2016-10-25
libcephfs client API overhaul and update

jtlayton:
This is the first pile of patches for a proposed overhaul of libcephfs client API.  The brunt of the changes are there to support @gregsfortytwo 's UserPerm work. Applications can now set up a UserPerm object and pass that in to different calls.

Additionally, this plumbs in the use of ceph_statx in more calls, in particular, the readdirplus code.

All of these API changes add up to backward incompatible changes to the libcephfs ABI as well, so this also bumps the SOVERSION on the library from 1 to 2 (per @liewegas's suggestion on the mailing list), and changes the packaging accordingly.

I have nfs-ganesha changes that will allow it build against this. I also have some older samba patches, but those may need some work after the latest changes.

Note that this is just the first pile. I think we also will want to make some other changes to better support ganesha here (in particular, allowing it to pass in a userperm on read, write and fsync), but this is a good spot to stop and take a pause.


jtlayton:
Test run mostly passed. A couple of jobs ended up dead for reasons that aren't clear to me, but I suspect that those are unrelated to the changes in this PR:

    http://pulpito.ceph.com/jlayton-2016-10-25_18:12:41-fs-wip-jlayton-cephfs---basic-mira/

jtlayton:
Note that there are companion PRs for teuthology and ceph-qa-suite that should be merged at the same time as this one:

https://github.com/ceph/ceph-qa-suite/pull/1223
https://github.com/ceph/teuthology/pull/972

PR #11648, opened 2016-10-25
rgw: fix for bucket delete racing with mdlog sync

cbodley:
proposed fix for bucket deletion on a remote racing with mdlog sync

Fixes: http://tracker.ceph.com/issues/17698

### DIFF
--- a/src/rgw/rgw_op.cc
+++ b/src/rgw/rgw_op.cc
@@ -2481,6 +2481,14 @@ void RGWDeleteBucket::execute()
   }
 
   op_ret = store->delete_bucket(s->bucket, ot);
+
+  if (op_ret == -ECANCELED) {
+    // lost a race, either with mdlog sync or another delete bucket operation.
+    // in either case, we've already called rgw_unlink_bucket()
+    op_ret = 0;
+    return;
+  }
+
   if (op_ret == 0) {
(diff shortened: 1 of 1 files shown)


src/rgw/rgw_op.cc
```diff
@@ -2481,6 +2481,12 @@ void RGWDeleteBucket::execute()
   }
 
   op_ret = store->delete_bucket(s->bucket, ot);
+
+  if (op_ret == -ECANCELED && !store->is_meta_master()) {
```
yehudasa: @cbodley can't this happen with multiple deletes going to the same bucket on the master zone also?

cbodley: probably, yeah. should we map `ECANCELED` to `ENOENT` in that case, or return success there as well?

theanalyst: @yehudasa @cbodley was able to reproduce the multiple deletes going to master zone and master raising a -125; but no debug logs; so yes we should handle this  case too

cbodley:
updated to handle races on the metadata master as well. also added a return, because we've already called `rgw_unlink_bucket()` in either case

theanalyst:
the latest change worked for both the cases, ie. races in master zone & slave zone for bucket removal

PR #11650, opened 2016-10-25
Log: Replace namespace log with logging

wjwithagen:
- This to avoid a naming conflict that Clang finds with
   std::log from cmath

### DIFF
--- a/src/common/Graylog.cc
+++ b/src/common/Graylog.cc
@@ -15,7 +15,7 @@
 #include "include/uuid.h"
 
 namespace ceph {
-namespace log {
+namespace logging {
 
 Graylog::Graylog(const SubsystemMap * const s, std::string logger)
     : m_subs(s),
@@ -167,5 +167,5 @@ void Graylog::log_log_entry(LogEntry const * const e)
   }
 }
 
-} // ceph::log::
+} // ceph::logging::
 } // ceph::
--- a/src/common/Graylog.h
+++ b/src/common/Graylog.h
@@ -21,7 +21,7 @@ namespace ceph {
 
 class Formatter;
 
-namespace log {
+namespace logging {
 
 struct Entry;
 class SubsystemMap;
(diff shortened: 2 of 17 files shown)


src/common/Graylog.cc
```diff
@@ -15,7 +15,7 @@
 #include "include/uuid.h"
 
 namespace ceph {
-namespace log {
+namespace loging {
```
tchaikov: s/loging/logging/

wjwithagen:
@tchaikov 
Redid the patch since I used loging instead of logging. But it was easy since I just fixed it in the patchfile and reapplied the patch.
Lets see if Jenkins likes it.
I'm off for the evening, so I'll see the result when I return.

wjwithagen:
@tchaikov 
I guess that this is going to take a full QA run before it can be merged? And thus will take some time.?
Do you know if any developer info/doc needs to be changed if this PR would be merged?

tchaikov:
this change looks good to me.

@liewegas is it good to merge if it builds on gitbuilder and the qa run passes?

tchaikov:
@wjwithagen i don't think there is any dev doc is referencing the changed namespace "log". maybe what we need is just a bless from another developer.

wjwithagen:
@liewegas @cbodley  
Clang trips over the use of namespace log. Most likely due to the loading of the ceph namespace t global level. @tchaikov has looked into removing this, and rewriting references as needed. Which turns out a lot of trivial changes in many files. 
So the alternative is to avoid this naming conflict and rename the namespace.
Which turns out to be a rather limited change. 
Would you care to take a look at this and bless it?

liewegas:
Yep, lgtm :)

wjwithagen:
@liewegas 
Thanx, and waiting for QA to do it thing.

tchaikov:

wjwithagen:
@tchaikov 
Thanx for helping.

PR #11651, opened 2016-10-25
mds/MDSRank.h: make destructor protected

stiopaa1:
MDSRank class is inherited from, but doesn't have virtual destructor.
Making destructor protected prevents possibility of deleting derived
classes using base class pointer.


gregsfortytwo:
Please test this

gregsfortytwo:
(LGTM)

PR #11655, opened 2016-10-26
mon/MonClient: fix race condition when shuting down the mon client

runsisi:
we have to shutdown the hunting timer and reset `cur_con` at the same place, or the hunting timer may set `cur_con` before it get shutdown, which results segfault as follows:

```
  #0  0x00007fb09ffca989 in raise () from /lib64/libc.so.6
  #1  0x00007fb09ffcc098 in abort () from /lib64/libc.so.6
  #2  0x00007fb08ea52677 in ceph::__ceph_assert_fail(char const*, char const*, int, char const*) () from /lib64/librados.so.2
  #3  0x00007fb08e93144c in ceph::log::SubsystemMap::should_gather(unsigned int, int) [clone .part.120] () from /lib64/librados.so.2
  #4  0x00007fb08e97eb15 in RefCountedObject::put() () from /lib64/librados.so.2
  #5  0x00007fb08eae3f9e in MonClient::~MonClient() () from /lib64/librados.so.2
  #6  0x00007fb08e97c2d5 in librados::RadosClient::~RadosClient() () from /lib64/librados.so.2
  #7  0x00007fb08e97c319 in librados::RadosClient::~RadosClient() () from /lib64/librados.so.2
  #8  0x00007fb08e94684a in rados_shutdown () from /lib64/librados.so.2
  #9  0x00007fb098074210 in __pyx_pw_5rados_5Rados_7shutdown () from /usr/lib64/python2.7/site-packages/rados.so
  ...
```

### DIFF
--- a/src/mon/MonClient.cc
+++ b/src/mon/MonClient.cc
@@ -418,6 +418,8 @@ void MonClient::shutdown()
     waiting_for_session.pop_front();
   }
 
+  timer.shutdown();
+
   if (cur_con)
     cur_con->mark_down();
   cur_con.reset(NULL);
@@ -429,10 +431,6 @@ void MonClient::shutdown()
     finisher.wait_for_empty();
     finisher.stop();
   }
-  monc_lock.Lock();
-  timer.shutdown();
-
-  monc_lock.Unlock();
 }
 
 int MonClient::authenticate(double timeout)


tchaikov:
@runsisi how do you connect the segfault to the destruction of `Connection` (`cur_con`)? is the logging subsystem referencing something already got destructed?

runsisi:
hi @tchaikov 
1. `RadosClient::RadosClient` gets a ref to `RadosClient::cct`
2. `rados_shutdown`
3. `RadosClient::~RadosClient` releases the only ref and delete `RadosClient::cct`
4. `RadosClient::~RadosClient` deconstructs `RadosClient::monclient`
5. `MonClient::~MonClient` deconstructs `ConnectionRef cur_con`
6. `Connection::put`, i.e., `RefCountedObject::put` uses the already deleted `RefCountedObject::cct` to do logging

BTW, this segfault can be easily reproduced by setting:
```
client_mount_timeout = 1.0
mon_client_hunt_interval = 0.5
mon_client_hunt_interval_backoff = 1.0
```
and with unreachable mon hosts

tchaikov:
> uses the already deleted `RefCountedObject::cct` to do logging

the `cct` held by `Connection` comes from the `messenger.cct`. where the messenger is either async msgr or simple msgr used by the mon client. and the `cct` held by msgr is in turn comes from `RadosClient` in your case. `RadosClient` holds a strong reference of `cct`. but `~RadosClient()` relinquish the reference to its `cct` *before* other member variables holding it are destructed, for instance,  `monclient`, `mgrclient`. that's why `monclient` is not happy when it tries to reference the `cct` member variable.

this only happens if the "debug refs" is enabled. a better approach might be using the deleter to postpone the deletion of cct, see https://github.com/ceph/ceph/pull/11659. could you give it a try?

runsisi:
yeah, your solution is more elegant :), e.g., if the `MonClient::finisher` is not an anonymous instance, then, as for my solution, `Finisher::~Finisher` should be a big problem.

but i do not understand one of your statements:

> this only happens if the "debug refs" is enabled

i think if we are referencing the deleted `cct`, the segfault should always be triggerred ?

i will close this PR until you have answered my question above :) thx.

runsisi:
wait, i think we should not close this PR, otherwise, if this race condition happens, we will be referencing an already deleted `RadosClient::messenger`.
@tchaikov do you think so?

tchaikov:
@runsisi could you elaborate a little bit on how RadosClient::messenger would be referenced?

runsisi:
@tchaikov never mind, i did not read the code carefully.

but do you think that we may leave a dangling `con` there and wait the msgr to recycle is elegant enough ? and i am still wondering why did you state `this only happens if the "debug refs" is enabled`?

tchaikov:
> but do you think that we may leave a dangling `con` there and wait the msgr to recycle is elegant enough ?

i don't see there is any problem this way as long as we hold a strong reference of cct until the last minute of RadosClient.

> and i am still wondering why did you state `this only happens if the "debug refs" is enabled`?

sorry, i was wrong, cct is always dereferenced no matter what the debug level of `refs` is.

runsisi:
@tchaikov thank you, i am closing this now.

PR #11659, opened 2016-10-26
librados: postpone cct deletion

tchaikov:
when `debug refs` is enabled, RefCountedObject::put() prints log using
cct, but RadosClient relinquishes the reference to cct too early, so its
when its `RefCountedObject` member variables print log when they are
destructed, the cct is freed. so we need postpone destructure of it
after all other RefCountedObject member variables are destroyed.


liewegas:
lgtm

tchaikov:
changelog

- fix syntax error in commit message.

tchaikov:
changelog

- update commit message: the crash always happens, no matter what the debug level of "refs" is.

tchaikov:

PR #11661, opened 2016-10-26
cmake: fix git version string, cleanup

liewegas:
The version string grew a 'v' starting with the cmake transition.  Remove it, and drop the unused ceph_ver.h in the tarball.


liewegas:
gitbuilder is happy at least

liewegas:
@alimaredia ?

make-dist
```diff
@@ -32,8 +32,9 @@ bin/git-archive-all.sh --prefix ceph-$version/ \
 		       $outfile.tar
 
 # populate files with version strings
-echo "including src/.git_version, src/ceph_ver.h, ceph.spec"
-src/make_version -g src/.git_version -c src/ceph_ver.h
+echo "including src/.git_version, ceph.spec"
+
+git rev-parse HEAD 2> /dev/null; git describe 2> /dev/null > src/.git_version
```
dang: This is dropping the 'git rev-parse HEAD'.  The old script was capturing STDOUT, so it got both lines, but this one is only redirecting the 'git describe'

liewegas:
D'oh, @dang fixed

PR #11665, opened 2016-10-27
mon: Do not allow pools to be deleted by default

wido:
This prevents accidental pool removals on cluster which may contain
very valuable data.

Users can still revert this setting and allow a pool to be removed,
but that would be a deliberate action of the admin.

Signed-off-by: Wido den Hollander wido@42on.com

### DIFF
--- a/src/common/config_opts.h
+++ b/src/common/config_opts.h
@@ -270,7 +270,7 @@ OPTION(mon_pg_check_down_all_threshold, OPT_FLOAT, .5) // threshold of down osds
 OPTION(mon_cache_target_full_warn_ratio, OPT_FLOAT, .66) // position between pool cache_target_full and max where we start warning
 OPTION(mon_osd_full_ratio, OPT_FLOAT, .95) // what % full makes an OSD "full"
 OPTION(mon_osd_nearfull_ratio, OPT_FLOAT, .85) // what % full makes an OSD near full
-OPTION(mon_allow_pool_delete, OPT_BOOL, true) // allow pool deletion
+OPTION(mon_allow_pool_delete, OPT_BOOL, false) // allow pool deletion
 OPTION(mon_globalid_prealloc, OPT_U32, 10000)   // how many globalids to prealloc
 OPTION(mon_osd_report_timeout, OPT_INT, 900)    // grace period before declaring unresponsive OSDs dead
 OPTION(mon_force_standby_active, OPT_BOOL, true) // should mons force standby-replay mds to be active


jcsp:
I was thinking of this setting as a way for professional deployers or tools to set up a cluster, and then flip this switch to protect it from the less-expert administrators.  So switching it on would be a conscious handover action.

If it was on by default, then there would need to be various updates to tests and documentation, and I'd be concerned that new people would look at the number of steps to delete a pool and think "This Ceph thing is really awkward to use" (even though we're being awkward on purpose).

I'm sure there will be varying opinions here!

wido:
@jcsp Yes, I know. So that's why I went ahead and opened the PR.

My opinion is that since Ceph cares so much about data consistency it should also make it very hard to remove data from the system. Imho it's to easy now to remove a 2PB pool should one have it.

We have the options there, but I never run into a system in the wild where these are used.

I wouldn't say only less-expert admins, think about this pseudo code:

<pre>for pool in [pool1, pool2, pool3]:
  rados_pool_remove(pool)</pre>

Some dev makes a mistake and removes all pools from the system.

I know this goes both ways and the PR is just flipping a true to false which might have big implications.

Would love to know what other think!

liewegas:
I'm less worried about the test side of things.. for teuthology at least we can set the default in the teuthology ceph.conf.

Given the recent unhappy camper on ceph-users I'm leaning toward changing this!

wido:
Good to hear @liewegas ! Yes, the user on ceph-users yesterday inspired me to open this PR, but it has been in my mind for a longer period.

At customers I always set this setting to false and set the nodelete flag on all pools.

My point is just protecting users against themselves. Loosing data in any way is just bad, very bad.

With ZFS for example you can still recover a pool you just threw away, but with Ceph there is no such thing. Gone is gone.

src/common/config_opts.h
```diff
@@ -270,7 +270,7 @@ OPTION(mon_pg_check_down_all_threshold, OPT_FLOAT, .5) // threshold of down osds
 OPTION(mon_cache_target_full_warn_ratio, OPT_FLOAT, .66) // position between pool cache_target_full and max where we start warning
 OPTION(mon_osd_full_ratio, OPT_FLOAT, .95) // what % full makes an OSD "full"
 OPTION(mon_osd_nearfull_ratio, OPT_FLOAT, .85) // what % full makes an OSD near full
-OPTION(mon_allow_pool_delete, OPT_BOOL, true) // allow pool deletion
+OPTION(mon_allow_pool_delete, OPT_BOOL, false) // allow pool deletion
```
tchaikov: @wido we might want to update https://github.com/ceph/teuthology/blob/master/teuthology/ceph.conf.template before merging this change. because quite a few of our qa suites are deleting pools without setting `mon_allow_pool_delete`.

tchaikov: dropping from my testing branch.

tchaikov: @liewegas what do you think?

liewegas:
updated ceph.conf

liewegas:
I wonder if we should consider a more robust safeguard, too.  Like, don't actually delete pools immediately, but rename them, and purge them only after N days.

wido:
@liewegas We had that discussion on the ML before the feature and the pool flags were introduced and came up with this solution. Some sort of grace period would be indeed be welcome I think.

A pool is removed, gets a special flag which makes it impossible to do I/O on and after 72 hours it is deleted by the MONs.

However, you should still be able to drop a pool right now if you want to due to maybe space or PG constraints.

tchaikov:

tchaikov:

PR #11671, opened 2016-10-27
rgw_file:  apply missed base64 try-catch

mattbenjamin:
Fixes a case missed in 0a4c91ec.  Affects only librgw/rgw_file
(i.e., RGW-NFS).

Fixes: http://tracker.ceph.com/issues/17663

Signed-off-by: Matt Benjamin mbenjamin@redhat.com


yehudasa:
@mattbenjamin lgtm, needs testing?

PR #11677, opened 2016-10-28
rocksdb: do not build with --march=native

tchaikov:
we can not assume that building host has the same arch (or the same cpu
capability) with the target machine.


bassamtabbara:
LGTM

badone:
lgtm

liewegas:
How should we fix this upstream in rocksdb so that we don't have to carry this patch?  Add a cmake option?

tchaikov:
change posted to upstream https://github.com/facebook/rocksdb/pull/1429

tchaikov:
@liewegas patch submitted to upstream. if we really want to have some optimization on some arch, for example i686. we should be using a toolchain cmake file to customize the CCFLAGS, etc. and use it instead.

tchaikov:
being tested at http://pulpito.front.sepia.ceph.com:80/kchai-2016-10-28_12:33:09-rados-wip-rocksdb-no-march-native---basic-mira/

tchaikov:
i am merging this change in hope to bring the master branch back to normal before the test passes.

PR #11678, opened 2016-10-28
librbd: silence the unused variable warning

tchaikov:


src/test/librbd/image/test_mock_RefreshRequest.cc
```diff
@@ -190,8 +190,9 @@ class TestMockImageRefreshRequest : public TestMockFixture {
   }
 
   void expect_snap_namespace_list(MockRefreshImageCtx &mock_image_ctx, int r) {
-    auto &expect = EXPECT_CALL(get_mock_io_ctx(mock_image_ctx.md_ctx),
-                               exec(mock_image_ctx.header_oid, _, StrEq("rbd"), StrEq("get_snapshot_namespace"), _, _, _));
+    EXPECT_CALL(get_mock_io_ctx(mock_image_ctx.md_ctx),
```
dillaman: Should actually be:

```
    if (r < 0) {
      expect.WillOnce(Return(r));
    } else {
      expect.WillOnce(DoDefault());
    }
```

dillaman:
It would be better to have the correct implementation -- otherwise ```r``` is unused

tchaikov:
@dillaman updated as suggested

PR #11680, opened 2016-10-28
test/ceph_test_msgr: do not use Message::middle for holding transient…

tchaikov:
… data

Message::middle is used for holding encoded data, so we we can not stuff
it with payload and leave the "payload" field empty. this change
refactors the ceph_test_msgr by introducing a Payload class which
encodes all test data in it.

Fixes: http://tracker.ceph.com/issues/17728


yuyuyu101:
really cool, lgtm if already pass test

PR #11682, opened 2016-10-28
DNM: RFC: EC Overwrites

athanatos:
It's passing enough tests now to be worth looking at.  For the up-to-the-minute version, see https://github.com/athanatos/ceph/tree/wip-ec-cache.

There are a few DNM commits in there to make reading the logs less annoying, I'll remove them when I update this published branch.

You'll probably want to start with the updated docs to get an overview.  Then, the series has two pieces:
1) refactor everything to use the new and improved PGTransaction
2) use the new structure to make overwrites work


athanatos:
I closed this one because github and jenkins didn't seem to be able to see more recent commits

athanatos:
Replaced by https://github.com/ceph/ceph/pull/11701

PR #11683, opened 2016-10-28
src/vstart.sh: Only execute btrfs if it is available

wjwithagen:

### DIFF
--- a/src/vstart.sh
+++ b/src/vstart.sh
@@ -661,7 +661,9 @@ if [ "$start_osd" -eq 1 ]; then
 EOF
 
 	    rm -rf $CEPH_DEV_DIR/osd$osd || true
-	    for f in $CEPH_DEV_DIR/osd$osd/*; do btrfs sub delete $f &> /dev/null || true; done
+            if command -v btrfs > /dev/null; then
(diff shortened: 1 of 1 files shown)


src/vstart.sh
```diff
@@ -661,7 +661,9 @@ if [ "$start_osd" -eq 1 ]; then
 EOF
 
 	    rm -rf $CEPH_DEV_DIR/osd$osd || true
-	    for f in $CEPH_DEV_DIR/osd$osd/*; do btrfs sub delete $f &> /dev/null || true; done
+            if [ -x /usr/bin/btrfs ]; then
```
tchaikov: `btrfs` is considered as a system administration command, and usually lives in `/usr`. probably you can use `command -v` instead to avoid this fuss? i.e.
```sh
if command -v btrfs > /dev/null; then
  ...
fi
```

wjwithagen: @tchaikov 
Never too old to learn. 8=) checking out `command`

wjwithagen: @tchaikov 
Ping.

wjwithagen:
@tchaikov 
FreeBSD seems happy...

tchaikov:
lgtm

PR #11684, opened 2016-10-29
common/admin_socket: add config for admin socket permission bits

runsisi:
this is to fix https://github.com/influxdata/telegraf/issues/1657, so other tools, such as `telegraf`, running as non-root user can gather info from the ceph daemons.

### DIFF
--- a/src/common/admin_socket.cc
+++ b/src/common/admin_socket.cc
@@ -299,6 +299,18 @@ void AdminSocket::chown(uid_t uid, gid_t gid)
   }
 }
 
+void AdminSocket::chmod(mode_t mode)
+{
+  if (m_sock_fd >= 0) {
+    int r = ::chmod(m_path.c_str(), mode);
+    if (r < 0) {
+      r = -errno;
+      lderr(m_cct) << "AdminSocket: failed to chmod socket: "
+                   << cpp_strerror(r) << dendl;
+    }
+  }
+}
+
 bool AdminSocket::do_accept()
 {
   struct sockaddr_un address;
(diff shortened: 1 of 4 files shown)


runsisi:
hi @liewegas  , can you review this for me ?
thank you :)

overskylab:
any update on this?

src/common/common_init.cc
```diff
@@ -132,4 +133,23 @@ void common_init_finish(CephContext *cct)
...
+
+  md_config_t *conf = cct->_conf;
+
+  if (!conf->admin_socket.empty() && !conf->admin_socket_mode.empty()) {
+    int ret = 0;
+    std::string err;
+
+    ret = strict_strtol(conf->admin_socket_mode.c_str(), 8, &err);
+    if (err.empty()) {
+      if (!(ret & (~07777))) {
```
tchaikov: could use `ALLPERMS` or better off using `ACCESSPERMS`. for example, what is the use case of applying `setuid`?

src/common/admin_socket.cc
```diff
@@ -299,6 +299,18 @@ void AdminSocket::chown(uid_t uid, gid_t gid)
   }
 }
 
+void AdminSocket::chmod(mode_t mode)
+{
+  if (m_sock_fd >= 0) {
+    int r = ::chmod(m_path.c_str(), mode);
```
liewegas: since we have the fd, then let's fchmod instead

runsisi: If `fildes` refers to a socket, the behavior of fchmod() is unspecified.[1]

[1] https://linux.die.net/man/3/fchmod

runsisi:
@tchaikov @liewegas updated, thanks!

PR #11686, opened 2016-10-29
ceph-rest-api: understand the new style entity_addr_t representation

tchaikov:
the new entity_addr_t representation was introduced by 3561eb3, we
should update ceph_rest_api.py accordingly.

Fixes: http://tracker.ceph.com/issues/17742
Signed-off-by: Kefu Chai kchai@redhat.com


tchaikov:
test passed at http://pulpito.ceph.com/kchai-2016-10-30_09:52:32-rados-wip-ceph-rest-api-new-addr---basic-mira/

PR #11687, opened 2016-10-29
[DNM] rgw_file: prevent unlink of non-empty directories

mattbenjamin:
Now that directory name objects can have attributes, we need to
avoid deleting them if they have children.

Note that this guard has the same atomicity guarantee as other
cross-object namespace operations, so races with other endpoints
to add or delete children are tolerated.

Signed-off-by: Matt Benjamin mbenjamin@redhat.com

### DIFF
--- a/src/rgw/rgw_file.cc
+++ b/src/rgw/rgw_file.cc
@@ -270,9 +270,19 @@ namespace rgw {
 	rgw_fh->mtx.lock(); /* LOCKED */
       }
 
+      /* to preserve attributes, avoid deleting the directory vnode
+       * if children exist */
+      if (rgw_fh->has_children(RGWFileHandle::HC_FLAG_SKIP_BUCKETS)) {
+	rgw_fh->mtx.unlock();
+	unref(rgw_fh);
+	return -ENOTEMPTY;
+      }
+
       std::string oname = rgw_fh->relative_object_name();
-      if (rgw_fh->is_dir())
+      if (rgw_fh->is_dir()) {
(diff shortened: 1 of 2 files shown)


mattbenjamin:
@yuyuyu101 I have to actually test it...

src/rgw/rgw_file.cc
```diff
@@ -724,6 +734,41 @@ namespace rgw {
...
+      rc = rgwlib.get_fe()->execute_req(&req);
+    } else {
+      RGWReaddirRequest req(fs->get_context(), fs->get_user(),
+			    const_cast<RGWFileHandle*>(this), has_children_cb,
+			    &cb_result, &offset);
+      req.set_default_max(1);
+      rc = rgwlib.get_fe()->execute_req(&req);
+    }
+
+    return rc;
```
cmgitdream: should return cb_result
for get the first dentry

src/rgw/rgw_file.cc
```diff
@@ -724,6 +734,41 @@ namespace rgw {
...
+    if (! is_dir())
+      return false;
+
+    int rc = 0;
+    bool cb_result = false;
+    uint64_t offset = 0;
+
+    if (is_root() &&
+	(flags & HC_FLAG_SKIP_BUCKETS)) {
+      return false;
```
cmgitdream: why not check the first level dir(bucket)?

oritwas:
@cmgitdream please rebase

PR #11695, opened 2016-10-31
os/bluestore: update cache logger after 'trim_cache' operation

tanghaodong25:


liewegas:
Can you move the _update_cache_logger call into trim_cache() instead of duplicating it at every caller?

tanghaodong25:
Updated. Thanks.

PR #11696, opened 2016-10-31
kv/rocksdb: enable rocksdb write path breakdown

tanghaodong25:
bluestore has unstable write performance partly because of
rocksdb internal overhead. So we add more breakdown in
rocksdb writh path.


liewegas:
lgtm, but needs rebase!

tanghaodong25:
Rebased.

PR #11702, opened 2016-10-31
qa/workunits/cephtool/test.sh: FreeBSD has no distro.

wjwithagen:

### DIFF
--- a/qa/workunits/cephtool/test.sh
+++ b/qa/workunits/cephtool/test.sh
@@ -1137,8 +1137,10 @@ function test_mon_osd()
   expect_false ceph osd find xyz
   expect_false ceph osd find 0.1
   ceph --format plain osd find 1 # falls back to json-pretty
-  ceph osd metadata 1 | grep 'distro'
-  ceph --format plain osd metadata 1 | grep 'distro' # falls back to json-pretty
+  if [ `uname` == Linux ]; then
+    ceph osd metadata 1 | grep 'distro'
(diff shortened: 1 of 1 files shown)


qa/workunits/cephtool/test.sh
```diff
@@ -1137,8 +1137,10 @@ function test_mon_osd()
   expect_false ceph osd find xyz
   expect_false ceph osd find 0.1
   ceph --format plain osd find 1 # falls back to json-pretty
-  ceph osd metadata 1 | grep 'distro'
-  ceph --format plain osd metadata 1 | grep 'distro' # falls back to json-pretty
+  if [ `uname` != FreeBSD ]; then
```
tchaikov: the distro info collected by `ceph osd metadata` comes from the output `lsb_release(1)`, which in turn is a Linux specific command. so, maybe we should put something like:
```sh
if [ `uname` == Linux ]; then
 ...
fi
```

wjwithagen: @tchaikov 
That is the alternative, but it excludes any other platform that is going to try to be Linux compatible and does have lsb_release.
I've also considered to "tweak" the lsb-data gathering, but then this is only informational and could set people on the wrong foot, thinking they can do more processing with it Linux-style.
And then the mis-information only propagates further.
So please confirm you want the `Linux`-test

wjwithagen:
@tchaikov 
Ping?
FreeBSD => Linux

PR #11706, opened 2016-10-31
msg/async/AsyncMessenger.cc: use explicit reference from my_inst.

wjwithagen:
Very likely Clang is being picky here, but if we do not reference the
msgr->my_inst ifor addr, then the nonce is incorrectly updated and
communication starts to fail.

And cephtool-test-rados.sh starts to fail.

### DIFF
--- a/src/msg/async/AsyncMessenger.cc
+++ b/src/msg/async/AsyncMessenger.cc
@@ -167,7 +167,7 @@ int Processor::rebind(const set<int>& avoid_ports)
   ldout(msgr->cct, 10) << __func__ << " new nonce " << nonce << " and inst " << msgr->my_inst << dendl;
 
   ldout(msgr->cct, 10) << __func__ << " will try " << addr << " and avoid ports " << new_avoid << dendl;
-  return bind(addr, new_avoid);
+  return bind(msgr->my_inst.addr, new_avoid);
 }
 
 void Processor::start()


src/msg/async/AsyncMessenger.cc
```diff
@@ -167,7 +167,7 @@ int Processor::rebind(const set<int>& avoid_ports)
   ldout(msgr->cct, 10) << __func__ << " new nonce " << nonce << " and inst " << msgr->my_inst << dendl;
 
   ldout(msgr->cct, 10) << __func__ << " will try " << addr << " and avoid ports " << new_avoid << dendl;
-  return bind(addr, new_avoid);
+  return bind(msgr->my_inst.addr, new_avoid);
```
yuyuyu101: why clang failed to refer to addr? what's the error message?

wjwithagen: @yuyuyu101 
There is no error message during compilation.
The code just doesn't work. 
And that shows during reconnecting where connections refused because the nonce does not match.

wjwithagen: @yuyuyu101 
I think it is a compiler bug, that does not correctly uses the new value of addr, but an older value (exactly 100000 less_.
When I use `msgr->my_inst.addr` as value for bind, the problem goes away.
And both values are equivalent from user perspective.
So I see no problem in changing the reference, and it helps me in avoiding thru hoops in getting the compiler fixed. (I've reported the problem, but it is very hard to reporduce in an "simple" example.

wjwithagen:
retest this please

src/msg/async/AsyncMessenger.cc
tchaikov: @wjwithagen i discussed this with @yuyuyu101 over IRC just now. it's not a bug of clang. we should pass the `addr` with the new nonce, and the current code fails to do so. i am posting a new PR replacing yours so it can be reviewed by @yuyuyu101 sooner. also it's actually a bug in ceph, i am filing a ticket on tracker at http://tracker.ceph.com/issues/17807

tchaikov:
superseded by #11804

src/msg/async/AsyncMessenger.cc
wjwithagen: @tchaikov  @yuyuyu101 
Oke, good. I'm testing that now.
I need a godd IRC client that does not disconnect every 5 minutes, AND where I can kill all un/subscribe messages. Otherwise I just lose track of the discussions.
Challenge is that my terminals run W10.

PR #11710, opened 2016-10-31
Fix long stalls when calling ceph_fsync()

jtlayton:
This PR should fix [tracker bug 17563](http://tracker.ceph.com/issues/17563)

The userland client fsync codepath is quite slow in some (rather common) workloads. I have a libcephfs test program that creates a file, writes 16k to it, calls ceph_fsync on it and then ceph_close. This is done in a loop 256 times.

Without this patchset when I run this program vs a vstart cluster under "time":
```
real    21m19.619s
user    0m0.431s
sys     0m0.364s
```

With this patchset:
```
real    0m11.491s
user    0m0.247s
sys     0m0.127s
```

Roughly 100x speedup. I first noticed this when testing with ganesha, backed by ceph, so this should help its workload as well.

### DIFF
--- a/src/client/Client.cc
+++ b/src/client/Client.cc
@@ -3417,7 +3417,7 @@ void Client::check_caps(Inode *in, unsigned flags)
       if (in->cap_snaps.size())
 	flush_snaps(in, true);
       if (in->flushing_caps)
-	flush_caps(in, session);
+	flush_caps(in, session, flags & CHECK_CAPS_SYNCHRONOUS);
     }
 
     int flushing;
@@ -3429,8 +3429,8 @@ void Client::check_caps(Inode *in, unsigned flags)
       flush_tid = 0;
     }
 
-    send_cap(in, session, cap, false, cap_used, wanted, retain, flushing,
-	     flush_tid);
+    send_cap(in, session, cap, flags & CHECK_CAPS_SYNCHRONOUS, cap_used, wanted,
+	     retain, flushing, flush_tid);
   }
 }
 
@@ -4133,7 +4133,7 @@ void Client::flush_caps()
   }
 }
 
-void Client::flush_caps(Inode *in, MetaSession *session)
+void Client::flush_caps(Inode *in, MetaSession *session, bool sync)
 {
   ldout(cct, 10) << "flush_caps " << in << " mds." << session->mds_num << dendl;
   Cap *cap = in->auth_cap;
@@ -4142,7 +4142,14 @@ void Client::flush_caps(Inode *in, MetaSession *session)
   for (map<ceph_tid_t,int>::iterator p = in->flushing_cap_tids.begin();
        p != in->flushing_cap_tids.end();
        ++p) {
-    send_cap(in, session, cap, false, (get_caps_used(in) | in->caps_dirty()),
+    bool req_sync = false;
+
+    /* If this is a synchronous request, then flush the journal on last one */
+    if (sync && (p->first == in->flushing_cap_tids.rbegin()->first))
+      req_sync = true;
+
+    send_cap(in, session, cap, req_sync,
+	     (get_caps_used(in) | in->caps_dirty()),
 	     in->caps_wanted(), (cap->issued | cap->implemented),
 	     p->second, p->first);
   }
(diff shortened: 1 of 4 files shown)


jtlayton:
Mostly passing fs sepia run here:

http://pulpito.ceph.com/jlayton-2016-10-31_20:35:33-fs-wip-jlayton-fsync---basic-mira/

jtlayton:
Note too that if this protocol change looks ok, then we'll almost certainly want to propagate this change to the kernel client as well.

jtlayton:
One thing that may make sense is to turn this bool into an (advisory) flags field or an enum that describes how the flush should occur. We could keep one bit for a "sync" flag, and one for "more data coming".

The latter could hint to the MDS that it should delay a log flush until it receives one from a client with that flag cleared. That could help the mds batch up updates from a single client more effectively.

src/messages/MClientCaps.h
```diff
@@ -117,7 +120,8 @@ class MClientCaps : public Message {
       time_warp_seq(0),
       osd_epoch_barrier(0),
       oldest_flush_tid(0),
-      caller_uid(0), caller_gid(0) {
+      caller_uid(0), caller_gid(0),
+      sync(false) {
```
liewegas: Need to set HEAD_VERSION = 10 above, and make the decoding of the new field conditional on that

src/messages/MClientCaps.h
```diff
@@ -263,6 +269,7 @@ class MClientCaps : public Message {
     if (header.version >= 9) {
       ::decode(btime, p);
       ::decode(change_attr, p);
+      ::decode(sync, p);
```
liewegas: if (header.version >= 10)

src/messages/MClientCaps.h
jtlayton: Jewel didn't have btime or change_attr. Those fields would both be new as of kraken.

So to be clear, that means that we do need to worry about backward compatability vs. development builds since jewel? If not, then we can still call it struct version 9 and it should still work. If we care about sane behavior vs. earlier development builds though, then we probably will need to do as you suggest.

gregsfortytwo:
I'd like to go through this as well, self-assigning for that.

jtlayton:
Ok, I think my latest PR should address all of the comments. @liewegas care to re-review?

src/mds/Locker.cc
```diff
@@ -2706,7 +2708,7 @@ void Locker::handle_client_caps(MClientCaps *m)
     }
 
     // filter wanted based on what we could ever give out (given auth/replica status)
-    bool need_flush = false;
+    bool need_flush = m->flags & CLIENT_CAPS_SYNC;
```
jcsp: To make it clearer that we're always respecting this flag, maybe the need_flush assignments below should be changed into `need_flush |=`?

jtlayton:
@jcsp : I think the SQUASH patch I just added should address your concern. Untested as of yet but it should do the right thing. I'll spin up a test run with it once the build is done.

src/client/Client.cc
```diff
@@ -4105,22 +4118,30 @@ void Client::flush_caps()
...
   p = cap_list.begin();
   while (!p.end()) {
+    unsigned flags = CHECK_CAPS_NODELAY;
     Inode *in = *p;
+
     ++p;
-    check_caps(in, true);
+    if (p.end())
+      flags |= CHECK_CAPS_SYNCHRONOUS;
+    check_caps(in, flags);
```
gregsfortytwo: It looks like we're here forcing the caps immediately to disk on any invocation of flush_caps(). Can we rename the function to indicate that behavior?
Since we obviously call the per-file/per-session flush_caps() in lots of places where we don't want to force it instantly to disk.

jtlayton: To be clear, this is not the per-file/per-session flush_caps, -- this is the flush_caps function with no argument that flushes all caps back to the mds for unmount or syncfs. Can we rename this one to distinguish it from the other? Absolutely. I'm build-testing a patch for that now.

jcsp:
This (or at least a sufficiently recent version for me) passed tests here: http://pulpito.ceph.com/jspray-2016-11-10_21:38:43-fs-wip-jcsp-testing-20161108-distro-basic-mira/

@jtlayton I'm happy for this to merge as soon as it's squashed

jtlayton:
Squashed. @jcsp you may merge when ready.

PR #11711, opened 2016-10-31
rgw: set a default bucket quota of 100000

mattbenjamin:
The new default value is chosen is a known-safe maximum value
for bucket entries in one bucket index shard.


mattbenjamin:
@yehudasa is this what we mean by default quota value near expected max?

ktdreyer:
Matt, would you please file a Redmine ticket to track this issue, so we ensure that it's backported to Jewel?

ktdreyer:
Should a docs change accompany this fix?

ktdreyer:
Matt, I'm told that this change is going to be abandoned now?

PR #11715, opened 2016-11-01
Fix racey test by setting noout flag (tracker 17757)

dzafman:
Want to see what Jenkins does with this showing ceph pg dump pgs in the log.


dzafman:
jenkins test this please

dzafman:
Passed Jenkins once, updated sign-off so we'll see if it passes again.

dachary:
@dzafman do you want me to run into a loop to verify it always works ?

dzafman:
@dachary It seemed to reproduce every time on Jenkins and never on my build machine.  So I don't think it is necessary, but go ahead if you prefer.

PR #11717, opened 2016-11-01
os/bluestore: Record correctly wal_bytes for TransContext.

majianpeng:


src/os/bluestore/BlueStore.cc
```diff
@@ -6782,6 +6782,7 @@ int BlueStore::queue_transactions(
     _txc_add_transaction(txc, &(*p));
   }
 
+  txc->wal_bytes = txc->bytes;
```
liewegas: I think we don't want this one?  wal_bytes should probably only be the bytes that we write as wal records...

PR #11719, opened 2016-11-01
rgw: Anonymous users shouldn't be able to access requester pays buckets.

zhangsw:
Anonymous users can't access requester pays buckets in S3.
We can find that in http://docs.aws.amazon.com/AmazonS3/latest/dev/RequesterPaysBuckets.html.

Fixes: http://tracker.ceph.com/issues/17175

Signed-off-by: Zhang Shaowen <zhangshaowen@cmss.chinamobile.com>


idealguo:
@yehudasa @cbodley , please help to review, thanks

mattbenjamin:
@zhangsw this seems necessarily correct

PR #11720, opened 2016-11-01
os/bluestore: increase gc depth only if overlap truely happens

xiexingguo:

### DIFF
--- a/src/os/bluestore/BlueStore.cc
+++ b/src/os/bluestore/BlueStore.cc
@@ -2258,7 +2258,8 @@ bool BlueStore::ExtentMap::do_write_check_depth(
 
   bool do_collect = true;
   if (depth < g_conf->bluestore_gc_max_blob_depth) {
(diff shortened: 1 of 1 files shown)


src/os/bluestore/BlueStore.cc
```diff
@@ -2258,7 +2258,8 @@ bool BlueStore::ExtentMap::do_write_check_depth(
 
   bool do_collect = true;
   if (depth < g_conf->bluestore_gc_max_blob_depth) {
-    *blob_depth = 1 + depth;
+    if (head_overlap || tail_overlap)
```
ifed01: no much sense in this patch - when head_overlap ==false && tail_overlap==false depth is equal to 0 and we just performed a no-op by reassigning blob_depth to 1

xiexingguo: Okay, thanks for the explanation, Igor:-)

PR #11721, opened 2016-11-01
tests: TestErasureCodePluginJerasure must stop the log thread

dachary:
When a log entry is created in dout.h, it holds a pointer to the

   static size_t _log_exp_length

variable. When OnExitManager::~OnExitManager from Log.cc is called
during global destruction, it may happen after the static variables
referenced by some entries have been deallocated. The flush() function
will try to access these deallocated variables via the hint_size()
method and core dump.

To prevent this race, the Log::stop() method must be called explicitly
to ensure no entries are left when the global destructors are called.

Fixes: http://tracker.ceph.com/issues/17561

Signed-off-by: Loic Dachary <ldachary@redhat.com>

### DIFF
--- a/src/test/erasure-code/CMakeLists.txt
+++ b/src/test/erasure-code/CMakeLists.txt
@@ -81,7 +81,7 @@ target_link_libraries(unittest_erasure_code
 add_executable(unittest_erasure_code_plugin_jerasure
   TestErasureCodePluginJerasure.cc
   )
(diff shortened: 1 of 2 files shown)


tchaikov:
lgtm. but i'd recommend to keep http://tracker.ceph.com/issues/17561 open until we have a complete solution to this problem.

dachary:
@tchaikov agreed

PR #11725, opened 2016-11-01
os/bluestore: fix extent shard span check

liewegas:


xiexingguo:
@liewegas 
Seems these patches will cause `ceph_test_objectstore` noises.

Check attachment for detail.

[assert.txt](https://github.com/ceph/ceph/files/565969/assert.txt)

ifed01:
LGTM.

liewegas:
@xiexingguo that assert looks unrelated

PR #11728, opened 2016-11-02
qa: remove EnumerateObjects from librados upgrade tests

jdurgin:
These rely on new rados functionality not present in jewel


jdurgin:
tested as part of preparing the hammer-jewel-x suite

PR #11729, opened 2016-11-02
kv: Rocksdb statistics

majianpeng:
print rocksdb statistics

### DIFF
--- a/src/kv/RocksDBStore.cc
+++ b/src/kv/RocksDBStore.cc
@@ -242,6 +242,9 @@ int RocksDBStore::do_open(ostream &out, bool create_if_missing)
       return -EINVAL;
     }
   }
+
+  opt.statistics = rocksdb::CreateDBStatistics();
+
   opt.create_if_missing = create_if_missing;
   if (g_conf->rocksdb_separate_wal_dir) {
     opt.wal_dir = path + ".wal";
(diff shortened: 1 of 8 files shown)


src/kv/RocksDBStore.cc
```diff
@@ -242,6 +242,9 @@ int RocksDBStore::do_open(ostream &out, bool create_if_missing)
       return -EINVAL;
     }
   }
+
+  opt.statistics = rocksdb::CreateDBStatistics();
```
ifed01: Suggest to make statistics collection an optional feature,enabled by a config parameter. RocksDB manual has a mention that it might impact a performance a bit

src/kv/RocksDBStore.cc
```diff
@@ -366,6 +366,7 @@ void RocksDBStore::get_statistics(Formatter *f) {
   bool status = db->GetProperty("rocksdb.stats", &stats);
   if (status) {
     f->open_object_section("bluestore rocksdb statistics");
+    f->write_raw_data("\n We usually observe an overhead of 5%-10% for this command. Be carefully before using\n\n");
```
ifed01: I'm afraid that's not enough - performance impact starts when you create statistics object and assign it to DB instance in do_open method. Hence performance is affected permanently, not after command execution.

liewegas:
We just merged another PR to master with a rocksdb_perf config option.  Can you use that so that this is only conditionally enabled?

varadakari:
This PR is addressed with #12287. Closing this.

PR #11731, opened 2016-11-02
FileStore: Only verify split when it has been really done and done correctly

dragonylffly:


tchaikov:
lgtm.

tchaikov:
this option is not enabled by default, and is not enabled by any unit tests or qa suites.

PR #11733, opened 2016-11-02
common: release g_ceph_context before returns

tchaikov:
http://tracker.ceph.com/issues/17762

### DIFF
--- a/src/global/global_init.cc
+++ b/src/global/global_init.cc
@@ -130,11 +130,12 @@ void global_pre_init(std::vector < const char * > *alt_def_args,
   g_conf->complain_about_parse_errors(g_ceph_context);
 }
 
-void global_init(std::vector < const char * > *alt_def_args,
-		 std::vector < const char* >& args,
-		 uint32_t module_type, code_environment_t code_env,
-		 int flags,
-		 const char *data_dir_option, bool run_pre_init)
+boost::intrusive_ptr<CephContext>
+global_init(std::vector < const char * > *alt_def_args,
+	    std::vector < const char* >& args,
+	    uint32_t module_type, code_environment_t code_env,
+	    int flags,
+	    const char *data_dir_option, bool run_pre_init)
 {
   // Ensure we're not calling the global init functions multiple times.
   static bool first_run = true;
@@ -325,6 +326,18 @@ void global_init(std::vector < const char * > *alt_def_args,
     output_ceph_version();
 
   g_ceph_context->crush_location.init_on_startup();
+
+  return boost::intrusive_ptr<CephContext>{g_ceph_context, false};
+}
+
+void intrusive_ptr_add_ref(CephContext* cct)
+{
+  cct->get();
+}
+
+void intrusive_ptr_release(CephContext* cct)
+{
+  cct->put();
 }
 
 void global_print_banner(void)
--- a/src/test/erasure-code/TestErasureCodePluginJerasure.cc
+++ b/src/test/erasure-code/TestErasureCodePluginJerasure.cc
@@ -17,13 +17,12 @@
 
 #include <errno.h>
 #include <stdlib.h>
-#include "global/global_init.h"
 #include "erasure-code/ErasureCodePlugin.h"
-#include "common/ceph_argparse.h"
 #include "log/Log.h"
 #include "global/global_context.h"
 #include "common/config.h"
 #include "gtest/gtest.h"
+#include "test/unit.h"
 
 TEST(ErasureCodePlugin, factory)
 {
@@ -59,24 +58,6 @@ TEST(ErasureCodePlugin, factory)
                                   &erasure_code, &cerr));
     EXPECT_TRUE(erasure_code.get());
   }
-}
-
-int main(int argc, char **argv)
-{
-  vector<const char*> args;
-  argv_to_vec(argc, (const char **)argv, args);
-
-  global_init(NULL, args, CEPH_ENTITY_TYPE_CLIENT, CODE_ENVIRONMENT_UTILITY, 0);
-  common_init_finish(g_ceph_context);
-
-  const char* env = getenv("CEPH_LIB");
-  string directory(env ? env : ".libs");
-  g_conf->set_val("erasure_code_dir", directory, false, false);
-
-  ::testing::InitGoogleTest(&argc, argv);
-  int status = RUN_ALL_TESTS();
-  g_ceph_context->_log->stop();
-  return status;
 }
 
 /*
--- a/src/test/unit.h
+++ b/src/test/unit.h
@@ -18,7 +18,9 @@
 #include "include/types.h" // FIXME: ordering shouldn't be important, but right 
                            // now, this include has to come before the others.
 
+#include "common/ceph_argparse.h"
 #include "common/code_environment.h"
+#include "common/config.h"
 #include "global/global_context.h"
 #include "global/global_init.h"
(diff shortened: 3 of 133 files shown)


badone:
LGTM

dachary:
testing manually with the reproducer to verify it fixes the issue

tchaikov:
- http://pulpito.ceph.com/kchai-2016-11-04_10:40:58-rados-wip-kefu-testing---basic-mira/
- http://pulpito.ceph.com/kchai-2016-11-05_15:35:54-rados-wip-kefu-testing---basic-mira/

failed tests are caused by #10049

tchaikov:
`Log::entry()` is flushing when the program terminates, that's why the producer crashes.

liewegas:
This looks right to me...

tchaikov:
@liewegas i am changing this PR so the `global_init()` will return a `unique_ptr` which calls `g_ceph_context->put()` in its dtor. and also fixes some other `g_ceph_context` leakage.

src/global/global_init.cc
```diff
@@ -325,6 +326,10 @@ void global_init(std::vector < const char * > *alt_def_args,
     output_ceph_version();
 
   g_ceph_context->crush_location.init_on_startup();
+
+  using Deleter = std::function<void(CephContext*)>;
+  return std::unique_ptr<CephContext, Deleter>{g_ceph_context,
+                                               [](CephContext *p) {p->put();}};
```
cbodley: since `CephContext` is a ref-counted type, i think `boost::intrusive_ptr` is a more natural fit than `std::unique_ptr` - see src/rgw/rgw_main.cc for an example

dachary:

dachary:

dachary:
<pre>
  9%] Performing configure step for 'rocksdb_ext'
CMake Error: The source directory "/home/jenkins-build/build/workspace/ceph-pull-requests/src/rocksdb" does not appear to contain CMakeLists.txt.
Specify --help for usage, or press the help button on the CMake GUI.
make[3]: *** [src/rocksdb_ext-prefix/src/rocksdb_ext-stamp/rocksdb_ext-configure] Error 1
make[2]: *** [src/CMakeFiles/rocksdb_ext.dir/all] Error 2
make[2]: *** Waiting for unfinished jobs....
[  9%] Building CXX object src/kv/CMakeFiles/kv_objs.dir/MemDB.cc.o
[  9%] Building CXX object src/kv/CMakeFiles/kv_objs.dir/RocksDBStore.cc.o
/home/jenkins-build/build/workspace/ceph-pull-requests/src/kv/RocksDBStore.cc:13:24: fatal error: rocksdb/db.h: No such file or directory
</pre>

tchaikov:
retest this please.

src/global/global_init.cc
tchaikov: @cbodley comment addressed and repushed.

tchaikov:
@athanatos mind taking a look?

athanatos:
I like it!

liewegas:
LGTM!  much better

dachary:
@tchaikov excellent :-) In the commit message for f4c4f42 s/quites/quits/ + s/and ./and /. If at all possible it would be good to split  f4c4f42 into several so that it's more readable (the unit.h change for instance).

tchaikov:
@dachary will do.

tchaikov:
changelog

- just fixed the typos in commit message of f4c4f42

tchaikov:
retest this please

tchaikov:
./bin/unittest_erasure_code_plugin_jerasure' segfault. rerunning.

src/test/erasure-code/TestErasureCodePluginJerasure.cc
```diff
@@ -61,24 +60,6 @@ TEST(ErasureCodePlugin, factory)
   }
 }
 
-int main(int argc, char **argv)
```
dachary: this is different from test/unit.h and needs to be adapted

tchaikov:
changelog

- do not take a refcount when constructing the `intrusive_ptr<>`.

@liewegas , sorry! we need to reschedule a rados qa test.

src/test/erasure-code/TestErasureCodePluginJerasure.cc
dachary: my bad, it was an autotools leftover

tchaikov:
crushtool times out.

src/tools/crushtool.cc
```diff
@@ -274,6 +274,9 @@ int main(int argc, const char **argv)
   auto cct = global_init(NULL, env_args, CEPH_ENTITY_TYPE_CLIENT,
 			 CODE_ENVIRONMENT_UTILITY,
 			 CINIT_FLAG_NO_DEFAULT_CONFIG_FILE);
+  // crushtool times out occasionally when quits. so do not
+  // release the g_ceph_context.
+  cct->get();
   common_init_finish(g_ceph_context);
```
dachary: It would be nice to comment on the root cause of this timeout. A leak there is harmless but knowing the root cause would be good to check if that's because of a problem that may be shared by other parts of the code, maybe with side effects that are more subtle and more difficult to diagnose afterwards. Am I making sense ?

tchaikov: @dachary i tried to reproduce this problem in a heavy loaded machine, the crushtool times out twice out of 200 runs. it could take more than 6 seconds or 16 seconds to complete the check. and i try to catch this by wrap the crushtool with a python script which launches gdb to print out the backtrace of all threads if it times out but the gdb reports that the debugged process is a zombie and cannot be attached.

in short, 1) the root cause is still unknown. 2) the leak is not a regression. i just make it more obvious.

dachary: Ok

tchaikov:
changelog

- `unit.h`: pass 0 instead of `CINIT_FLAG_NO_DEFAULT_CONFIG_FILE`. as we are also using the `ceph.conf` to customize the behavior of unit tests.

src/test/unit.h
```diff
@@ -32,10 +34,19 @@
  * initialization for you.
  */
 int main(int argc, char **argv) {
-  std::vector<const char*> args;
-  global_init(NULL, args, CEPH_ENTITY_TYPE_CLIENT, CODE_ENVIRONMENT_UTILITY,
-	      CINIT_FLAG_NO_DEFAULT_CONFIG_FILE);
```
dachary: CINIT_FLAG_NO_DEFAULT_CONFIG_FILE may be here so that tests are not influenced by an existing ceph installation on the developer machine (i.e. no attempt to read /etc/ceph/ceph.conf even if it's available). This is just a thought: I did not actually verify that's going to happen but ... that's what I thought.

tchaikov:
tested at 

- http://pulpito.ceph.com/kchai-2016-11-24_15:35:58-rados-master---basic-mira/ without the "unit.h: pass 0 instead of CINIT_FLAG_NO_DEFAULT_CONFIG_FILE" change
- http://pulpito.ceph.com/kchai-2016-11-24_15:35:58-rados-master---basic-mira/ with the "unit.h: pass 0 instead of CINIT_FLAG_NO_DEFAULT_CONFIG_FILE" change

src/test/unit.h
dachary: right

dillaman:
FYI -- this change caused rbd-nbd to immediately seg fault. I am fixing under http://tracker.ceph.com/issues/18070

PR #10515, opened 2016-07-31
mon/Monitor.h: add override to virtual functions

stiopaa1:
Signed-off-by: Michal Jarzabek <stiopa@gmail.com>

PR #11734, opened 2016-11-02
jewel: monitor should send monmap updates when the monmap is updated

Abhishekvrshny:
http://tracker.ceph.com/issues/17679

### DIFF
--- a/src/mon/Monitor.cc
+++ b/src/mon/Monitor.cc
@@ -4292,7 +4292,7 @@ void Monitor::handle_subscribe(MonOpRequestRef op)
 	pgmon()->check_sub(s->sub_map["osd_pg_creates"]);
       }
     } else if (p->first == "monmap") {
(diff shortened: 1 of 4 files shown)


Abhishekvrshny:
This backport does not apply cleanly and the build fails with error 
```
mon/Monitor.cc:4384:48: error: no ‘bool Monitor::ms_handle_refused(Connection*)’ member function declared in class ‘Monitor’
 bool Monitor::ms_handle_refused(Connection *con)
```
It looks like that it is dependent on commits which are not in jewel : https://github.com/ceph/ceph/commit/d58d7d3dbd21a8dca0a19964f51cb9bf78814a75 and https://github.com/ceph/ceph/pull/10515/commits/3f20c28f645b3d2df12af97bf08b1eb8d72f08f8

tchaikov:
superseded by #11743. might want to resolve the conflict in another way.

PR #11741, opened 2016-11-02
rbd-nbd: disallow mapping images >2TB in size

trociny:
Fixes: http://tracker.ceph.com/issues/17219

### DIFF
--- a/qa/workunits/rbd/rbd-nbd.sh
+++ b/qa/workunits/rbd/rbd-nbd.sh
@@ -4,6 +4,7 @@
 
 POOL=rbd
 IMAGE=testrbdnbd$$
+TOO_LARGE_IMAGE=${IMAGE}_large
 SUDO=sudo
 SIZE=64
 DATA=
@@ -16,6 +17,7 @@ setup()
     DATA=${TEMPDIR}/data
     dd if=/dev/urandom of=${DATA} bs=1M count=${SIZE}
     rbd --dest-pool ${POOL} --no-progress import ${DATA} ${IMAGE}
+    rbd -p ${POOL} create ${TOO_LARGE_IMAGE} --size 3T
 
     if [ `id -u` = 0 ]
     then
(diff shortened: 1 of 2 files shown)


qa/workunits/rbd/rbd-nbd.sh
```diff
@@ -4,6 +4,7 @@
 
 POOL=rbd
 IMAGE=testrbdnbd$$
+TOOLARGE_IMAGE=${IMAGE}_large
```
dillaman: Minor: suggest adding an underscore between "TOO" and "LARGE" to make it clearer

trociny:
@dillaman Thanks, updated.

src/tools/rbd_nbd/rbd-nbd.cc
```diff
@@ -594,6 +594,14 @@ static int do_map()
   }
 
   size = info.size;
+
+  if (size > (1UL << 32) * 512) {
+    r = -EINVAL;
```
vshankar: minor: would E2BIG be a more appropriate errcode?

trociny: Not sure. E2BIG is usually used when user input parameters are too big. This looks a little different. Still I don't have a strong opinion. @dillaman what do you think?

dillaman: I don't think ```E2BIG``` is a correct error code for the reason Mykola stated. You might be able to get away with ```EFBIG``` for "file too large"

trociny:
EFBIG looks much better! updated

edisno:
I am having problems doing discard/trim on newly created rbd volumes.  The krbd client in linux 4.8.0 does not support certain default rbd flags, and rbd-fuse seems to not support trim/discard.  rbd-nbd does, but it is limited to 2TB and my volumes are much larger.  Any idea if rbd-nbd might support >2TB (properly) soon?  I read somewhere that this problem might only affect 32-bit systems, so would I risk data corruption by commenting out the check on my (64-bit) system?

trociny:
@edisno See http://tracker.ceph.com/issues/18335
So we have plans to allow >2TB for not affected kernels. I am going to work on this ticket soon.

PR #11743, opened 2016-11-02
jewel: mon: send updated monmap to its subscribers

tchaikov:
http://tracker.ceph.com/issues/17679


Abhishekvrshny:
Thanks @tchaikov :)

dachary:
It passed the rados (http://tracker.ceph.com/issues/17851#note-4) suite. It also passed the upgrade/jewel-x and upgrade/hammer-x (http://tracker.ceph.com/issues/17851#note-7) suites, except for one job which should be fixed by https://github.com/ceph/ceph-qa-suite/pull/1256 and does not seem related to this pull request. Note that it won't be included in 10.2.4, reason why it targets jewel-next.

PR #11744, opened 2016-11-02
rbd-mirror: snap protect of non-layered image results in split-brain

trociny:
Fixes: http://tracker.ceph.com/issues/16962


dillaman:
lgtm

PR #11747, opened 2016-11-02
mon: MonmapMonitor: return success when monitor will be removed

jecluis:
We were cascading without changing the return value to success. This was incorrect, obviously.


tchaikov:
lgtm.

PR #11748, opened 2016-11-02
os/bluestore: fix _split_collections race with osr_reap

liewegas:
The SharedBlobSet may not yet be empty at split time because previous
transactions may not have been reaped, leaving some Blobs still alive
even after the cache refs are cleared.  Drop them explicitly here (they
will go away shortly).

### DIFF
--- a/src/os/bluestore/BlueStore.cc
+++ b/src/os/bluestore/BlueStore.cc
@@ -1143,6 +1143,12 @@ void BlueStore::OnodeSpace::clear()
   onode_map.clear();
 }
 
+bool BlueStore::OnodeSpace::empty()
+{
+  std::lock_guard<std::recursive_mutex> l(cache->lock);
+  return onode_map.empty();
+}
+
 void BlueStore::OnodeSpace::rename(OnodeRef& oldo,
 				     const ghobject_t& old_oid,
 				     const ghobject_t& new_oid,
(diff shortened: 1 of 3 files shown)


liewegas:
@ifed01 mind taking a look?

src/os/bluestore/BlueStore.cc
```diff
@@ -8930,6 +8930,12 @@ int BlueStore::_split_collection(TransContext *txc,
   c->onode_map.clear();
   d->onode_map.clear();
 
+  // We have an awkward race here: previous pipelinex transactions may
+  // still reference blobs and their shared_blobs.  They will be flushed
+  // shortly by _osr_reap_done, but it's awkward to block for that (and
+  // a waste of time).  Instead, explicitly remove them from the shared blob
+  // map.
+  c->shared_blob_set.violently_clear();
   assert(c->shared_blob_set.empty());
```
ifed01: no need for this assert any more.

src/os/bluestore/BlueStore.cc
```diff
@@ -8930,6 +8930,12 @@ int BlueStore::_split_collection(TransContext *txc,
...
   assert(d->shared_blob_set.empty());
```
ifed01: Not sure if this makes much sense but is it possible that destination collection struggles (or will struggle one day) with the same issue? Does it worth to clear its shared_blob_set the same way? Just in case...

liewegas:
The dest collection is always brand new and empty, so if anything we 
would want to assert that's the case.  Weird that we're clearing it 
currently.. I'll clean it up tho

PR #11750, opened 2016-11-02
[DNM] Make PG::queue_op lockless for common case

dlambrig:
PG::queue_op won't queue a request if the epoch is newer
than the one in the request. This patch changes that check
to be done in a lockless manner.

This is a  WIP patch, uploaded so other's could take a peek. 
more testing / analysis needed.

(1) obtain osdmap using an RCU managed pointer
(2) manage the osdmap's epoch using an atomic counter
(3) maintain request ordering for the session

Signed-off-by: Dan Lambright <dlambrig@gmail.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -182,9 +182,14 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ${ENABLE_SHARED})
 
 find_package(Backtrace)
 
+find_package(urcu REQUIRED)
+
 if(LINUX)
   find_package(udev REQUIRED)
   set(HAVE_UDEV ${UDEV_FOUND})
+
+  find_package(urcu REQUIRED)
+  set(HAVE_URCU ${URCU_FOUND})
 
   find_package(aio REQUIRED)
   set(HAVE_LIBAIO ${AIO_FOUND})
@@ -192,6 +197,8 @@ if(LINUX)
   find_package(blkid REQUIRED)
   set(HAVE_BLKID ${BLKID_FOUND})
 else()
+  set(HAVE_URCU OFF)
+  message(STATUS "Not using urcu")
   set(HAVE_UDEV OFF)
   message(STATUS "Not using udev")
   set(HAVE_LIBAIO OFF)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -220,6 +220,8 @@ if(HAVE_XIO)
   list(APPEND EXTRALIBS ${XIO_LIBRARY} pthread rt)
 endif(HAVE_XIO)
 
+set(RCU_LIBS ${GPERFTOOLS_TCMALLOC_LIBRARY})
+
 # sort out which allocator to use
 if(ALLOCATOR STREQUAL "tcmalloc")
   set(ALLOC_LIBS ${GPERFTOOLS_TCMALLOC_LIBRARY})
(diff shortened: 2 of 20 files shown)


dachary:

dachary:

CMakeLists.txt
```diff
@@ -182,16 +182,23 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ${ENABLE_SHARED})
 
 find_package(Backtrace)
 
+find_package(urcu REQUIRED)
```
mattbenjamin: @dlambrig looks like we have an unconditional find_package(urcu), and again in if(LINUX);  only the LINUX block sets HAVE_URCU;  is the LINUX check needed now?

CMakeLists.txt
```diff
@@ -182,16 +182,23 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ${ENABLE_SHARED})
...
+  find_package(urcu REQUIRED)
+  set(HAVE_URCU ${URCU_FOUND})
+
   find_package(aio REQUIRED)
   set(HAVE_LIBAIO ${AIO_FOUND})
 
   find_package(blkid REQUIRED)
   set(HAVE_BLKID ${BLKID_FOUND})
 else()
+  set(HAVE_URCU OFF)
```
mattbenjamin: @dlambrig and can the !HAVE_URCU case happen now?

src/CMakeLists.txt
```diff
@@ -220,6 +220,8 @@ if(HAVE_XIO)
   list(APPEND EXTRALIBS ${XIO_LIBRARY} pthread rt)
 endif(HAVE_XIO)
 
+set(RCU_LIBS ${GPERFTOOLS_TCMALLOC_LIBRARY})
```
mattbenjamin: @dlambrig looks odd--and different from form used in the target_link_libraries(...) locations

src/common/ceph_context.h
```diff
@@ -146,6 +146,8 @@ class CephContext {
    */
   CryptoHandler *get_crypto_handler(int type);
 
+  pthread_key_t registered;
```
mattbenjamin: @dlambrig could use a more specific var name?  (nit)

src/mon/OSDMonitor.cc
```diff
@@ -138,12 +138,12 @@ void OSDMonitor::create_initial()
 void OSDMonitor::update_from_paxos(bool *need_bootstrap)
 {
   version_t version = get_last_committed();
-  if (version == osdmap.epoch)
+  if (version == osdmap.get_epoch())
```
mattbenjamin: @dlambrig change to use epoch accessor is just stylistic? (not objecting)

src/msg/async/AsyncConnection.cc
```diff
@@ -327,6 +327,8 @@ void AsyncConnection::process()
   bool already_dispatch_writer = false;
   std::lock_guard<std::mutex> l(lock);
   last_active = ceph::coarse_mono_clock::now();
+
+  rcu_quiescent_state();
```
mattbenjamin: maybe add a comment above the rcu_quiescent_state() call noting its function?

src/msg/async/Event.cc
```diff
@@ -331,11 +332,18 @@ int EventCenter::process_events(int timeout_microseconds)
 
   ldout(cct, 10) << __func__ << " wait second " << tv.tv_sec << " usec " << tv.tv_usec << dendl;
   vector<FiredFileEvent> fired_events;
+
+  rcu_thread_offline();
```
mattbenjamin: similarly here;  I understand this logic pretty well from NEBR (an epoch-based reclaimation variant), but find the offline() online() notation surprising and might not have been 100% sure of the behavior if I were just encountering it

src/msg/simple/Pipe.cc
```diff
@@ -1518,17 +1519,22 @@ void Pipe::stop_and_wait()
 void Pipe::reader()
 {
   pipe_lock.Lock();
-
+
```
mattbenjamin: ws?

src/osd/OSDMap.cc
```diff
@@ -874,7 +874,7 @@ void OSDMap::Incremental::generate_test_instances(list<Incremental*>& o)
 
 void OSDMap::set_epoch(epoch_t e)
 {
-  epoch = e;
```
mattbenjamin: @dlambrig epoch is an RCU var?

dlambrig:
Have moved this PR to branch wip-lockless-queue-op.

PR #11751, opened 2016-11-02
rgw: extract host name from host:port string

yehudasa:
Fixes: http://tracker.ceph.com/issues/17788


mattbenjamin:
@yehudasa this seems sane, I think I should verify.

PR #11701, opened 2016-10-31
osd: EC Overwrites

athanatos:
It's passing enough tests now to be worth looking at. For the up-to-the-minute version, see https://github.com/athanatos/ceph/tree/wip-ec-cache.

You'll probably want to start with the updated docs to get an overview. Then, the series has two pieces:
1) refactor everything to use the new and improved PGTransaction
2) use the new structure to make overwrites work

PR #11754, opened 2016-11-03
osd/ReplicatedPG: do_update_log_missing: take the pg lock in the callback

athanatos:
Fixes: http://tracker.ceph.com/issues/17789

### DIFF
--- a/src/osd/ReplicatedPG.cc
+++ b/src/osd/ReplicatedPG.cc
@@ -9894,13 +9894,13 @@ void ReplicatedPG::do_update_log_missing(OpRequestRef &op)
   assert(m->get_type() == MSG_OSD_PG_UPDATE_LOG_MISSING);
   ObjectStore::Transaction t;
   append_log_entries_update_missing(m->entries, t);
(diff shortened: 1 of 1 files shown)


athanatos:
This is also included in the ec overwrites branch.

https://github.com/ceph/ceph/pull/11701

src/osd/ReplicatedPG.cc
```diff
@@ -9894,13 +9894,13 @@ void ReplicatedPG::do_update_log_missing(OpRequestRef &op)
...
   append_log_entries_update_missing(m->entries, t);
-  // TODO FIX
 
   Context *complete = new FunctionContext(
-      [=](int) {
-	MOSDPGUpdateLogMissing *msg =
-	  static_cast<MOSDPGUpdateLogMissing*>(
-	    op->get_req());
+    [=](int) {
+      MOSDPGUpdateLogMissing *msg = static_cast<MOSDPGUpdateLogMissing*>(
```
tchaikov: nit, could just use local variable `m` here.

PR #11763, opened 2016-11-03
osd: speedup the performance of multi-replication flow

liupan1111:
For multi-replication flow, it is enough to send message to primary only once, and we do this work in sub_op_modify_commit. Don't need to do it again in sub_op_modify_applied.


liupan1111:
@liewegas , please help take a look. Thanks.

src/osd/ReplicatedBackend.cc
```diff
@@ -1259,8 +1265,13 @@ void ReplicatedBackend::sub_op_modify_applied(RepModifyRef rm)
...
     ack->set_priority(CEPH_MSG_PRIO_HIGH); // this better match commit priority!
-    get_parent()->send_message_osd_cluster(
-      rm->ackerosd, ack, get_osdmap()->get_epoch());
+    if (!isBlueStore) {
+      get_parent()->send_message_osd_cluster(
+        rm->ackerosd, ack, get_osdmap()->get_epoch());
+    }
+    else {
+      ack->put();
+    }
```
liewegas: We shouldn't need a conditional on the lower-level backend here.  Why is bluestore different than filestore?  Is it because the bluestore ack callback happens before the commit callback?  In that case, the client should not request an ack at all (which we just changed with librados and librbd, btw).  Can you share a bit more about the motivation?

liupan1111:
Hi Sage, sorry for reply late.

I delete the ack message between replications, because bluetore ack callback happens before the commit callback. Bluestore ack and commit callback are queued one by one in function "BlueStore::_txc_finish_kv". Therefore, just only one callback is needed, we can delete one of them. For bluestore, we can delete the ack callback betweent replications, and the ack between client and osd can be deleted too.

liupan1111:
@liewegas

liewegas:
I think a simpler fix is to just switch the callback order in bluestore to 
do the onack callback after the oncommit one...

liupan1111:
Right, much simpler... Thanks! I will close this pull request and create a new one with your suggestion.

PR #11764, opened 2016-11-03
cmake: Support for embedding Ceph Daemons

bassamtabbara:
This PR adds support for embedding Ceph Daemons. Currently only MON and OSD are supported. See http://pad.ceph.com/p/embedded-ceph for more context.

Note that WITH_EMBEDDED is set to on by default to ensure that the bits don't go stale. I plan on adding unit tests in a separate PR for this work.

To ease the review, I recommend you look at each commit in order.


src/compressor/zlib/CompressionPluginZlib.cc
```diff
@@ -14,43 +14,11 @@
...
-      isal = false;
-    }
-    if (compressor == 0 || has_isal != isal) {
-      compressor = CompressorRef(new ZlibCompressor(isal));
-      has_isal = isal;
-    }
-    *cs = compressor;
-    return 0;
-  }
-};
```
liewegas: Is it really necessary to move these into the headers?

bassamtabbara: an alternative would be to split __ceph_plugin_init from CompressionPluginZlib.cc and into a separate file (plugin.cc?). I can then include CompressionPluginZlib.cc in libcephd. Let me know if you'd like me to do that.

tchaikov:
will do.

src/osd/ClassHandler.cc
```diff
@@ -75,9 +83,11 @@ int ClassHandler::open_all_classes()
 
 void ClassHandler::shutdown()
 {
-  for (auto& cls : classes) {
-    if (cls.second.handle) {
-      dlclose(cls.second.handle);
+  if (!disable_dlclose) {
```
tchaikov: do we really need to check `disable_dlclose`? because `_load_class()` is never called if the cls are embedded in libcephd, `cls.second.handle` is not set. so this piece of code just works in both cases (embedded or not), am i right?

src/libcephd/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+  os
+  osd
+  osdc
+  rocksdb)
+
+if(HAVE_ARMV8_CRC)
+  list(APPEND merge_libs common_crc_aarch64)
+endif(HAVE_ARMV8_CRC)
+
+merge_static_libraries(cephd ${merge_libs})
```
tchaikov: i am wondering what APIs are supposed to be exposed via libcephd? the ones start with "cephd_"?

src/cls/CMakeLists.txt
```diff
@@ -1,152 +1,220 @@
...
+
+set(cls_lua_client_srcs
+    lua/cls_lua_client.cc)
+add_library(cls_lua_client STATIC ${cls_lua_client_srcs})
+
+list(APPEND cls_embedded_srcs ${cls_lua_srcs} ${cls_lua_client_srcs})
+
+if(WITH_EMBEDDED)
+  include(MergeStaticLibraries)
+  list(REMOVE_DUPLICATES cls_embedded_srcs)
```
alimaredia: Is this REMOVE_DUPLICATES a backup in case something goes wrong? I'm not seeing why this is necessary.

src/cls/CMakeLists.txt
```diff
@@ -1,152 +1,220 @@
...
+  add_library(cephd_cls_base STATIC ${cls_embedded_srcs})
+  # while not necessary this seems to bring in the lua's include directories
+  # so that cls_lua srcs build correctly
+  target_link_libraries(cephd_cls_base liblua)
+  set_target_properties(cephd_cls_base PROPERTIES COMPILE_DEFINITIONS BUILDING_FOR_EMBEDDED)
```
alimaredia: @tchaikov should the `BUILDING_FOR_EMBEDDED` here have quotation marks or a "-D" in front of it?

cmake/modules/MergeStaticLibraries.cmake
```diff
@@ -0,0 +1,87 @@
...
+    endif(LINUX)
+
+    message("-- MergeStaticLibraries: ${target}: merged ${libs}")
+
+    # we want to set the target_link_libraries correctly for the new merged
+    # static library. First we get the list of link libraries for each
+    # of the libs we are merging
+    set(link_libs)
+    foreach(lib ${libs})
+    if(TARGET ${lib})
```
bassamtabbara: thats right they should all be cmake targets.

src/CMakeLists.txt
```diff
@@ -151,6 +151,9 @@ if(LINUX)
   list(APPEND EXTRALIBS ${LIB_RESOLV})
 endif(LINUX)
 
+# Embedded Library
+option(WITH_EMBEDDED "build the embedded ceph daemon library" ON)
```
bassamtabbara: the argument for making it the default is that the code will not rot. I'm going to be adding unit tests to ensure the functionality does not break as we make changes in the code base. If we turn it off it will never get tested and will eventually rot. the increased build time is minimal from what I can tell.

alimaredia: I don't think we should, @liewegas have an opinion?

src/include/cephd/libcephd.h
```diff
@@ -0,0 +1,48 @@
...
+#endif
+
+#define LIBCEPHD_VER_MAJOR 0
+#define LIBCEPHD_VER_MINOR 1
+#define LIBCEPHD_VER_PATCH 0
+
+#define LIBCEPHFD_VERSION(maj, min, extra) ((maj << 16) + (min << 8) + extra)
+#define LIBCEPHFD_VERSION_CODE LIBCEPHD_VERSION(LIBCEPHD_VER_MAJOR, LIBCEPHD_VER_MINOR, LIBCEPHD_VER_PATCH)
+
+#if __GNUC__ >= 4
```
bassamtabbara: sure. also a cut and paste from other headers in the code base like librados.h

src/libcephd/CMakeLists.txt
bassamtabbara: yes. the API is a start right now. more will be added.

src/libcephd/libcephd.cc
```diff
@@ -0,0 +1,234 @@
...
+  if (pminor)
+    *pminor = (n >= 2) ? minor : 0;
+  if (ppatch)
+    *ppatch = (n >= 3) ? patch : 0;
+  return v;
+}
+
+extern "C" int cephd_generate_fsid(char *buf, size_t len)
+{
+    if (len < 37) {
```
bassamtabbara: ok.

src/libcephd/libcephd.cc
```diff
@@ -0,0 +1,234 @@
...
+    fsid.generate_random();
+    fsid.print(buf);
+
+    return 0;
+}
+
+extern "C" int cephd_generate_secret_key(char *buf, size_t len)
+{
+    CephInitParameters iparams(CEPH_ENTITY_TYPE_MON);
+    CephContext *cct = common_preinit(iparams, CODE_ENVIRONMENT_LIBRARY, 0);
```
bassamtabbara: I'll fix.

src/libcephd/libcephd.cc
```diff
@@ -0,0 +1,234 @@
...
+    cct->_conf->apply_changes(NULL);
+    cct->init_crypto();
+
+    CryptoKey key;
+    key.create(cct, CEPH_CRYPTO_AES);
+
+    string keyString;
```
bassamtabbara: ok.

liewegas:
It's just the plugins that will build twice, right?

As long as there arne't new dependencies and the build time doesn't 
explode, it should build by default or it will rot.

src/include/config-h.in.cmake
```diff
@@ -123,9 +123,22 @@
 /* Accelio conditional compilation */
 #cmakedefine HAVE_XIO
 
+
 /* AsyncMessenger RDMA conditional compilation */
 #cmakedefine HAVE_RDMA
 
+/* define if embedded enabled */
+#cmakedefine WITH_EMBEDDED
```
bassamtabbara: The two flags have different semantics. WITH_EMBEDDED says we are building the embedded library. BUILDING_FOR_EMBEDDED means that the source code (usually a plugin) is currently being compiled for the use of the embedded library, and it should not compile any entry points etc. 

I do think the a single run of make is the best approach here. other wise we would need to run a special build just for embedded. given the small amount of recompile involved I dont believe this is needed. MySql for comparison follows a similar path, compiling twice only the code needed to generate libmysqld (they're cmake based as well).

src/osd/ClassHandler.cc
bassamtabbara: the destructor calls ClassHandler::shutdown and that calls dlclose.

src/include/cephd/libcephd.h
```diff
@@ -0,0 +1,48 @@
+#ifndef CEPH_LIBCEPHD_H
+#define CEPH_LIBCEPHD_H
```
bassamtabbara: librados, librbd and libcephfs all use old style header guards. I'm tempted to leave this as is for consistency. I like the pragma once as well :-) Let me know if you feel strongly about this one.

src/include/cephd/libcephd.h
bassamtabbara: similar case here with librados, librbd and libcephfs . they all have the `__GNUC__ >=4` check. I'm tempted to leave this the same also for consistency with the other libraries.

src/libcephd/libcephd.cc
```diff
@@ -148,6 +169,57 @@ void cephd_preload_embedded_plugins()
   }
 }
 
+void cephd_preload_rados_classes(OSD *osd)
+{
+  // intialize RADOS classes
+  {
```
bassamtabbara: the parentheses are a scope for the Mutex class

src/libcephd/libcephd.cc
```diff
@@ -0,0 +1,234 @@
...
+    }
+    strcpy(buf, keyString.c_str());
+    return keyString.length();
+}
+
+// load the embedded plugins. This is safe to call multiple
+// times in the same process
+void cephd_preload_embedded_plugins()
+{
+  int r;
```
bassamtabbara: fixed

bassamtabbara:
@liewegas yes only a few plugin source files are rebuilt for embedded. the build time increase is negligible.

@tchaikov @alimaredia I just pushed a set of rebased changes that capture your feedback. please let me know if there is something else.

adamemerson:
Rather than having a single, static library for everything, it seems like it would be of more general use to have a separate libceph-osd and libceph-mon that could be dynamically loaded as plugins.

bassamtabbara:
@adamemerson one of the key design points is that you can statically compile the daemons. You can certainly build the dynamic plugins ontop of this work, so its not an either or.

For example, you can have libcephd-mon dynamic plugin that statically compiled in libcephd and never called the OSD code. The linker will remove unused code.

CMakeLists.txt
```diff
@@ -240,6 +240,9 @@ option(WITH_KVS "Key value store is here" ON)
 # remote block storage
 option(WITH_RBD "Remote block storage is here" ON)
 
+# embedded ceph daemon library
+option(WITH_EMBEDDED "build the embedded ceph daemon library" ON)
```
liewegas: Perhaps a comment here and/or in libcephd.cc would be a good place to clarify the authors' intent of the source code license with respect to linking (IANAL but I am led to understand that authors' understanding and intent is important).  Something like:

Ceph is mostly LGPL (see COPYING), which means that static linking brings with it significant restrictions.  While it is permissible to dynamically link LGPL code with proprietary software (and redistribute the combined work), statically linking LGPL code means that the resulting derivative work must also be LGPL, or the proprietary code be provided in object form so that it can be relinked.  See http://www.gnu.org/licenses/gpl-faq.html#LGPLStaticVsDynamic.

bassamtabbara: thanks. I'll add a comment.

bassamtabbara: @liewegas added a comment in 72d4c46

src/cls/CMakeLists.txt
tchaikov: @alimaredia no, i don't think so.

src/include/cephd/libcephd.h
tchaikov: i don't feel strongly about this, but every time we introduce new code into this project. we have a chance to improve it in one way or another. by "improve", we could add more features using the old programming practice, or introduce the improvement with something (not really) new, take C++11 for example, if we really want to be consistent with the existing code, we would be forced to use the C++98 subset. the reason i recommend to use something different in the new code is, taking "#ifdef" as an example, that we could 1) lower the risk of changing the code in a future overhaul to replace the "#ifdef" with "#pragma" 2) the "git blame" would be easier to read after the overhaul.

not to mention that it's not encouraged to change the code only for the cosmetic reasons, IIUC.

bassamtabbara: @tchaikov thanks for the feedback. I'll go ahead and make the changes you suggested, and favor improvement over consistency.

bassamtabbara:
jenkins retest please

bassamtabbara:
@liewegas @tchaikov anything else left of this one?

liewegas:
Hmm, jenkins didn't like it when i did an integration build:
<pre>
-- Installing: /build/ceph-11.0.2-1550-gb88a9f4/debian/tmp/usr/lib/libcephd.a
CMake Error at src/libcephd/cmake_install.cmake:48 (FILE):
  file INSTALL cannot find
  "/build/ceph-11.0.2-1550-gb88a9f4/src/libcephd/include/rados/libcephd.h".
Call Stack (most recent call first):
  src/cmake_install.cmake:343 (INCLUDE)
  cmake_install.cmake:37 (INCLUDE)
</pre>

bassamtabbara:
@liewegas can you point me at the whole log?

liewegas:
https://jenkins.ceph.com/job/ceph-dev-build/ARCH=x86_64,AVAILABLE_ARCH=x86_64,AVAILABLE_DIST=trusty,DIST=trusty,MACHINE_SIZE=huge/2469/consoleFull#-14067203567641bc0b-bacc-4481-8290-0cf28df0ce68

bassamtabbara:
@liewegas I see the issue. Not sure why this has not come up before --maybe I've not done a `make install`. Fix on the way.

liewegas:
@bassamtabbara thanks!

bassamtabbara:
@liewegas fix pushed. waiting for jenkins to validate.

dachary:

bassamtabbara:
@liewegas this is good to go now

src/ceph_mon.cc
```diff
@@ -156,7 +156,7 @@ int check_mon_data_empty()
   return code;
 }
 
-void usage()
+void print_mon_usage()
```
tchaikov: can we just make this function static? or hide it in an anonymous namespace? the same applies to changes to other `usage()` functions.

src/libcephd/libcephd.cc
tchaikov: sorry, i was wrong, should call `cct->put()`  instead.

src/osd/ClassHandler.cc
tchaikov: i am not following you. as i think you are not addressing my comment. also, could you elaborate a little bit? whose destructor?

src/libcephd/libcephd.cc
```diff
@@ -0,0 +1,234 @@
...
+    plugin = new CompressionPluginZlib(g_ceph_context);
+    r = reg->add("compressor", "zlib", plugin);
+    if (r == -EEXIST) {
+      delete plugin;
+    }
+    assert(r == 0);
+  }
+}
+
+void cephd_preload_rados_classes(OSD *osd)
```
tchaikov: > its expected by ClassHandler. I believe we are calling a few methods that assert that the mutex is locked.

that's my question. why we would need it? just because the new code asserts it's locked?

CMakeLists.txt
```diff
@@ -234,6 +234,13 @@ option(WITH_KVS "Key value store is here" ON)
 # remote block storage
 option(WITH_RBD "Remote block storage is here" ON)
 
+# embedded ceph daemon static library
+# NOTE: Ceph is mostly LGPL (see COPYING), which means that
+# static linking brings with it restrictions. Please be sure
+# to look at the LGPL license carefully before linking this library to
+# your code. See http://www.gnu.org/licenses/gpl-faq.html#LGPLStaticVsDynamic.
+option(WITH_EMBEDDED "build the embedded ceph daemon library" ON)
```
tchaikov: @liewegas what do you think?

liewegas:
I think we have to turn it on by default or else it will immediately 
break.  And if I heard correctly this doesn't have a significant impact on 
build time?

src/ceph_mon.cc
bassamtabbara: good idea. will try that now.

src/libcephd/libcephd.cc
bassamtabbara: ahh ok. will fix now.

src/libcephd/libcephd.cc
bassamtabbara: @tchaikov `cephd_preload_rados_classes` calls `add_embedded_class` which does not require the mutex as you noted. However, the next call to each of the CLS implementation like `hello_cls_init` in turn call `register_class` which asserts that the mutex is locked. The whole ClassHandler class assumes that there is a mutex locking it. Do you see a problem with using the mutex inside cephd_preload_rados_classes?

bassamtabbara:
@tchaikov I factored your recent changes (removed dont_dlclose and made the usage functions static). @liewegas I'm testing on Centos 7 now and will get back to you soon.

cmake/modules/MergeStaticLibraries.cmake
```diff
@@ -0,0 +1,85 @@
...
+    endif(LINUX)
+
+    message("-- MergeStaticLibraries: ${target}: merged ${libs}")
+
+    # we want to set the target_link_libraries correctly for the new merged
+    # static library. First we get the list of link libraries for each
+    # of the libs we are merging
+    set(link_libs)
+    foreach(lib ${libs})
+      get_property(trans TARGET ${lib} PROPERTY LINK_LIBRARIES)
```
bassamtabbara: @tchaikov note that this was changed from `INTERFACE_LINK_LIBRARIES` to `LINK_LIBRARIES` so that it would work with cmake 2.8.11 on Centos7. This was the issue @liewegas saw.

bassamtabbara:
@liewegas @tchaikov Centos issue resolved. it was an incompatibility with cmake 2.8.11. I believe I factored the recent round of feedback and comments. please let me know if I missed anything.

dachary:
jenkins test this please (eio)

bassamtabbara:
@liewegas @tchaikov ping. I have some more changes that build on this commit that I'd like to propose once this one is merged.

liewegas:
Requires(rpmlib): rpmlib(FileDigests) <= 4.6.0-1 rpmlib(PayloadFilesHavePrefix) <= 4.0-1 rpmlib(CompressedFileNames) <= 3.0.4-1
Checking for unpackaged file(s): /usr/lib/rpm/check-files /home/jenkins-build/build/workspace/ceph-dev-build/ARCH/x86_64/AVAILABLE_ARCH/x86_64/AVAILABLE_DIST/centos7/DIST/centos7/MACHINE_SIZE/huge/release/11.0.2-1893-g6fb1d26/rpm/el7/BUILDROOT/ceph-11.0.2-1893.g6fb1d26.el7.x86_64
error: Installed (but unpackaged) file(s) found:
   /usr/include/cephd/libcephd.h
   /usr/lib64/libcephd.a


RPM build errors:
    Installed (but unpackaged) file(s) found:
   /usr/include/cephd/libcephd.h
   /usr/lib64/libcephd.a
+ rm -fr /tmp/install-deps.3625
Build step 'Execute shell' marked build as failure

https://jenkins.ceph.com/job/ceph-dev-build/ARCH=x86_64,AVAILABLE_ARCH=x86_64,AVAILABLE_DIST=centos7,DIST=centos7,MACHINE_SIZE=huge/2837//consoleFull

bassamtabbara:
@liewegas I removed libcephd.a and libcephd.h from the cmake install phase. As a result they will not be part of the distro packages (rpm and deb).  I've honestly not thought through the distribution of these yet, and can add that in the future. For now, libcephd will continue to build as part of `make` and hopefully will not rot.

Sorry for the back and forth -- I wish I knew how to run these "dev-builds" on jenkins myself to reduce the iterations.

liewegas:
QA run looks ok.  Kefu, is this ok to merge?

src/osd/ClassHandler.h
```diff
@@ -110,10 +109,13 @@ class ClassHandler
       const std::string& list);
 
 public:
+  Mutex mutex;
```
tchaikov: can we lock the mutex in `add_embedded_class()`, so we don't need to expose it as a public member variable?

CMakeLists.txt
tchaikov: @bassamtabbara 

if the sole purpose of enabling this option by default is to avoid the code from rotting, maybe we can at least disable it in `debian/rules` and `ceph.spec.in`.

src/osd/ClassHandler.h
bassamtabbara: @tchaikov I don't think that can be done easily. The issue is that call to `cls_init` methods immediately after `add_embedded_class` call `register_class` and that method expects the mutex to be already locked.

tchaikov: oic.

CMakeLists.txt
tchaikov: @bassamtabbara ping.

bassamtabbara: @tchaikov libcehpd no longer is installed and is not part of any deb or rpm. We're on the same page. See my last comment below.

bassamtabbara:
@tchaikov I've set `WITH_EMBEDDED=OFF` in `ceph.spec.in` and `debian/rules` in rebased `c6067f4`

liewegas:
good idea!

PR #11765, opened 2016-11-03
mon: if crushtool config is empty use internal crush test

bassamtabbara:
currently we expect the tool ```crushtool``` to be installed and available for crush map testing. This is not always the case. The change here will use the internal crush tester if the ```crushtool```
config is empty.


liewegas:
lgtm.  Note that you lose the timeout safeguard and crash-tolerance in this mode, though...

tchaikov:
http://pulpito.ceph.com/kchai-2016-11-16_11:34:58-rados-wip-kefu-testing---basic-mira/

PR #11766, opened 2016-11-03
librbd: snap_get_limit compatibility check

trociny:
This is not critical and it may fail e.g. with ENOTSUP when running
an upgraded client on an older cluster.

### DIFF
--- a/src/librbd/internal.cc
+++ b/src/librbd/internal.cc
@@ -1774,8 +1774,13 @@ int mirror_image_disable_internal(ImageCtx *ictx, bool force,
 
   int snap_get_limit(ImageCtx *ictx, uint64_t *limit)
   {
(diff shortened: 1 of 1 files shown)


trociny:
@dillaman May be also don't fail when `image.get_group` returns an error? It looks like it does not return ENOTSUP in upgrade:client tests, still the failure does not look critical to abort `rbd info`.

dillaman:
@trociny Looking at the implementation of "get_group", it should just return an empty group spec w/o returning an error if unsupported (since the group information is fetched during the RefreshRequest state machine and cached).

Perhaps we should handle the EOPNOTSUPP case within internal.cc and set the max to UINT64_MAX if you get this error back from the OSD.

trociny:
@dillaman Thanks, updated

dillaman:
lgtm

PR #11772, opened 2016-11-03
osd,mon: require sortbitwise flag to upgrade beyond jewel

liewegas:


liewegas:
This will need to go through jewel-x

liewegas:
and hammer-jewel-x, when it is ready

PR #11773, opened 2016-11-03
mds/MDSRank.cc: remove unneeded searching of map

stiopaa1:


liewegas:
I think the original is fine.  If anything, we'd want to change to a find() and make the function call conditional, but what's there is fine.

PR #11777, opened 2016-11-04
rgw: fix Content-Type for S3 user create

lnsyyj:
Bugfix,
the Content-Type is not returned when the s3 API user create.
Causing the client to not be able to parse the data from Content-Type.

http://tracker.ceph.com/issues/17818

Signed-off-by: YuJiang lnsyyj@hotmail.com


liewegas:
Please rebase on current master, and include a Signed-off-by line in the git commit message

lnsyyj:
@liewegas I'm sorry, I changed a new one.

https://github.com/ceph/ceph/pull/11828

PR #11778, opened 2016-11-04
rbd: specify the snap_name during image open if needed

yangdongsheng:

### DIFF
--- a/qa/workunits/rbd/copy.sh
+++ b/qa/workunits/rbd/copy.sh
@@ -74,7 +74,7 @@ test_others() {
 
     # remove snapshots
     rbd snap rm --snap=snap1 testimg1
-    rbd info --snap=snap1 testimg1 2>&1 | grep 'error setting snapshot context: (2) No such file or directory'
+    rbd info --snap=snap1 testimg1 2>&1 | grep 'error opening testimg1@snap1: (2) No such file or directory'
 
     remove_images
     rm -f $TMP_FILES
--- a/src/tools/rbd/Utils.cc
+++ b/src/tools/rbd/Utils.cc
@@ -764,17 +764,18 @@ int init_io_ctx(librados::Rados &rados, const std::string &pool_name,
 }
 
 int open_image(librados::IoCtx &io_ctx, const std::string &image_name,
-               bool read_only, librbd::Image *image) {
+               const std::string &snap_name, bool read_only,
+	       librbd::Image *image) {
   int r;
   librbd::RBD rbd;
   if (read_only) {
-    r = rbd.open_read_only(io_ctx, *image, image_name.c_str(), NULL);
+    r = rbd.open_read_only(io_ctx, *image, image_name.c_str(), snap_name.c_str());
   } else {
-    r = rbd.open(io_ctx, *image, image_name.c_str());
+    r = rbd.open(io_ctx, *image, image_name.c_str(), snap_name.c_str());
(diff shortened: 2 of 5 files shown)


dillaman:
lgtm

dillaman:
http://qa-proxy.ceph.com/teuthology/jdillaman-2016-11-07_00:13:39-rbd-wip-jd-testing-distro-basic-smithi/528491/teuthology.log

With this change, running "rbd info image@<missing snapshot name>" no longer returns a nice error message about a missing snapshot. Instead, it states that the image is missing. This results in a test failure.

yangdongsheng:
@dillaman Okey, the error message of ```rbd info --snap snap image1``` changed from ```error setting snap ``` to ```error opening image```.

There is an idea to pass a callback to open() to get the state of opening state machine, then give user a better message. But that means we have to make OpenRequest and rbd tool to share a enum of opening states. I am afraid this is not a good thing. And I don't think that's worthy for a cleanup. 

Or, just change the unittest for the new manner?

Or, close this patch, just keep what it was? I think this option is good enough now. it works well currently.

What's your suggestion?

trociny:
May be just change the open error message
```
std::cerr << "rbd: error opening image " << image_name << ...
```
to something like
```
std::cerr << "rbd: error opening " << image_name << (snap_name.empty() ? "" : "@" + snap_name) << ... 
```
?

yangdongsheng:
Yes, this was an option to me, but I am not sure  is this unittest 
introduced to ensure some backward compatibility.
If application is relaying on this kind of output of rbd command, we 
will break it.

trociny:
I don't think this check was a backward compatibility test (so just update the test accordingly). I think the main point is to have a useful and not confusing message.

And sure let's wait for Jason opinion.

yangdongsheng:
Yes, Mykola, agree with you. ;-)

dillaman:
@yangdongsheng I'm fine w/ the change so long as the error message isn't confusing. I also think there was an librbd-internal error message being printed to the console, so that message would need to get downgraded from a ```lderr``` to an ```ldout```

yangdongsheng:
@dillaman error message updated. But about the lderr downgrading, do you mean this one? https://github.com/ceph/ceph/blob/master/src/librbd/image/OpenRequest.cc#L471

If yes, I think this error is printed in log and there is exactly an error happend in setting snap, so I think we can keep it as lderr.

although there is already a similar error printed in console, this kind of error message in ceph.log can help us to find more details about addressing the problem. and lderr can make sure this message will not be missed by user. That means, I think the error message printed in console should not affect the messages in ceph.log. @dillaman @trociny What's you opinion?

dillaman:
@yangdongsheng Yes, that is probably the one. The issue is that std::cout/cerr and lderr will get intermingled on the console and it's a terrible user experience for issues that are very likely to occur (like fat fingering an image or snapshot name).

yangdongsheng:
@dillaman Okey, sounds a general problem about std::cerr in rbd tool and lderr in librbd. But if we change the lderr in librbd to ldout, this would affect other api for librbd, such as python-binding. can we prevent lderr in console by some means?

dillaman:
@yangdongsheng ... or we just take the approach that this cleanup is causing more work than is worthwhile

yangdongsheng:
@dillaman  Haha, yes, it's just a cleanup ;-)

PR #11780, opened 2016-11-04
os/bluestore: a few more cleanups

xiexingguo:


src/os/bluestore/BlueStore.cc
```diff
@@ -6693,8 +6693,8 @@ void BlueStore::_kv_sync_thread()
       KeyValueDB::Transaction synct = db->get_transaction();
 
       // increase {nid,blobid}_max?  note that this covers both the
-      // case wehre we are approaching the max and the case we passed
-      // it.  in either case, we increase the max in the earlier txn
+      // case we are approaching the max and the case we passed it.
```
ifed01: should be: *where* we are approaching...?

xiexingguo: @ifed01 Thanks, Igor.

PR #11783, opened 2016-11-04
mds: enable rmxattr on pool_namespace attrs 

jcsp:


gregsfortytwo:

Looks like routing it through handle_set_vxattr takes care of users trying to change the layout on already-written files, so this should be okay.

PR #11796, opened 2016-11-05
rgw: add bucket size limit check to radosgw-admin

mattbenjamin:
http://tracker.ceph.com/issues/17925

The change adds a new list of all buckets * all users, with
fields for bucket name, tenant name, current num_objects,
current num_shards, current objects per shard, and the
corresponding fill_status--the latter consisting of 'OK',
'WARN <n>%', or 'OVER <n>%.'

The warning check is relative to two new tunables.  The threshold
max objects per shard is set as rgw_bucket_safe_max_objects_per_shard,
which defaults to 100K.  The value rgw_bucket_warning_threshold is
a percent of the current safe max at which to warn (defaults to
10).

### DIFF
--- a/src/common/config_opts.h
+++ b/src/common/config_opts.h
@@ -1651,6 +1651,10 @@ OPTION(rgw_realm_reconfigure_delay, OPT_DOUBLE, 2) // seconds to wait before rel
 OPTION(rgw_period_push_interval, OPT_DOUBLE, 2) // seconds to wait before retrying "period push"
 OPTION(rgw_period_push_interval_max, OPT_DOUBLE, 30) // maximum interval after exponential backoff
 
+OPTION(rgw_safe_max_objects_per_shard, OPT_INT, 100*1024) // safe max loading
+OPTION(rgw_shard_warning_threshold, OPT_DOUBLE, 90) // pct of safe max
+						    // at which to warn
+
 OPTION(rgw_swift_versioning_enabled, OPT_BOOL, false) // whether swift object versioning feature is enabled
 
 OPTION(mgr_module_path, OPT_STR, CEPH_PKGLIBDIR "/mgr") // where to load python modules from
--- a/src/rgw/rgw_admin.cc
+++ b/src/rgw/rgw_admin.cc
@@ -1,3 +1,4 @@
+// -*- mode:C++; tab-width:8; c-basic-offset:2; indent-tabs-mode:t -*-
 // vim: ts=8 sw=2 smarttab
 
 #include <errno.h>
@@ -72,6 +73,7 @@ void _usage()
   cout << "  key create                 create access key\n";
   cout << "  key rm                     remove access key\n";
   cout << "  bucket list                list buckets\n";
+  cout << "  bucket limit check         show bucket sharding stats\n";
   cout << "  bucket link                link bucket to specified user\n";
   cout << "  bucket unlink              unlink bucket from specified user\n";
   cout << "  bucket stats               returns bucket statistics\n";
@@ -285,6 +287,9 @@ void _usage()
   cout << "   --categories=<list>       comma separated list of categories, used in usage show\n";
   cout << "   --caps=<caps>             list of caps (e.g., \"usage=read, write; user=read\")\n";
   cout << "   --yes-i-really-mean-it    required for certain operations\n";
+  cout << "   --warnings-only           when specified with bucket limit check, list\n";
+  cout << "                             only buckets nearing or over the current max\n";
+  cout << "                             objects per shard value\n";
   cout << "   --bypass-gc               when specified with bucket deletion, triggers\n";
   cout << "                             object deletions by not involving GC\n";
   cout << "   --inconsistent-index      when specified with bucket deletion and bypass-gc set to true,\n";
@@ -338,6 +343,7 @@ enum {
   OPT_KEY_CREATE,
   OPT_KEY_RM,
   OPT_BUCKETS_LIST,
+  OPT_BUCKETS_LIMIT_CHECK,
   OPT_BUCKET_LINK,
   OPT_BUCKET_UNLINK,
   OPT_BUCKET_STATS,
@@ -555,6 +561,10 @@ static int get_cmd(const char *cmd, const char *prev_cmd, const char *prev_prev_
   } else if (strcmp(prev_cmd, "buckets") == 0) {
     if (strcmp(cmd, "list") == 0)
       return OPT_BUCKETS_LIST;
+    if (strcmp(cmd, "limit") == 0) {
+      *need_more = true;
+      return 0;
+    }
   } else if (strcmp(prev_cmd, "bucket") == 0) {
     if (strcmp(cmd, "list") == 0)
       return OPT_BUCKETS_LIST;
@@ -576,14 +586,18 @@ static int get_cmd(const char *cmd, const char *prev_cmd, const char *prev_prev_
       *need_more = true;
       return 0;
     }
-  } else if ((prev_prev_cmd && strcmp(prev_prev_cmd, "bucket") == 0) &&
-	     (strcmp(prev_cmd, "sync") == 0)) {
-    if (strcmp(cmd, "status") == 0)
-      return OPT_BUCKET_SYNC_STATUS;
-    if (strcmp(cmd, "init") == 0)
-      return OPT_BUCKET_SYNC_INIT;
-    if (strcmp(cmd, "run") == 0)
-      return OPT_BUCKET_SYNC_RUN;
+  } else if (prev_prev_cmd && strcmp(prev_prev_cmd, "bucket") == 0) {
+    if (strcmp(prev_cmd, "sync") == 0) {
+      if (strcmp(cmd, "status") == 0)
+	return OPT_BUCKET_SYNC_STATUS;
+      if (strcmp(cmd, "init") == 0)
+	return OPT_BUCKET_SYNC_INIT;
+      if (strcmp(cmd, "run") == 0)
+	return OPT_BUCKET_SYNC_RUN;
+    } else if ((strcmp(prev_cmd, "limit") == 0) &&
+	       (strcmp(cmd, "check") == 0)) {
+      return OPT_BUCKETS_LIMIT_CHECK;
+    }
   } else if (strcmp(prev_cmd, "log") == 0) {
     if (strcmp(cmd, "list") == 0)
       return OPT_LOG_LIST;
@@ -2437,6 +2451,7 @@ int main(int argc, const char **argv)
 
   int sync_stats = false;
   int bypass_gc = false;
+  int warnings_only = false;
   int inconsistent_index = false;
 
   int verbose = false;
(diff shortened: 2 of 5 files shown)


mattbenjamin:
@yehudasa could you give this a look over?  tx

src/rgw/rgw_admin.cc
```diff
@@ -4234,6 +4237,42 @@ int main(int argc, char **argv)
     }
   }
 
+  if (opt_cmd == OPT_BUCKETS_LIMIT_CHECK) {
+
+	  void *handle;
+      metadata_key = "user";
```
yehudasa: indentation?

src/rgw/rgw_bucket.cc
```diff
@@ -1411,6 +1413,116 @@ static int bucket_stats(RGWRados *store, const std::string& tenant_name, std::st
...
+  const size_t max_entries =
+    store->ctx()->_conf->rgw_list_buckets_max_chunk;
+
+  const size_t safe_max_objects_per_shard =
+    store->ctx()->_conf->rgw_safe_max_objects_per_shard;
+
+  uint16_t shard_warn_pct =
+    store->ctx()->_conf->rgw_shard_warning_threshold;
+  if (shard_warn_pct > 100)
+    shard_warn_pct = 10;
```
yehudasa: is it 10 on purpose here? If > 100 then reset to 100 I think

src/rgw/rgw_bucket.cc
```diff
@@ -1411,6 +1413,116 @@ static int bucket_stats(RGWRados *store, const std::string& tenant_name, std::st
...
+	uint64_t num_objects = 0;
+
+	/* need stats for num_entries */
+	string bucket_ver, master_ver;
+	std::map<RGWObjCategory, RGWStorageStats> stats;
+	ret = store->get_bucket_stats(bucket, RGW_NO_SHARD, &bucket_ver,
+				      &master_ver, stats, nullptr);
+	if (! ret) {
+	  for (const auto& s : stats) {
+	    num_objects = s.second.num_objects;
```
yehudasa: `num_objects += s.second.num_objects;`

src/rgw/rgw_bucket.cc
```diff
@@ -1411,6 +1413,116 @@ static int bucket_stats(RGWRados *store, const std::string& tenant_name, std::st
...
+	  if (objs_per_shard > safe_max_objects_per_shard) {
+	    ss << boost::format("OVER %4f%%") %
+	      (safe_max_objects_per_shard/objs_per_shard * 100);
+	  } else {
+	    double safe_fill_pct =
+	      100 - (objs_per_shard / safe_max_objects_per_shard * 100);
+	    if (safe_fill_pct <= shard_warn_pct) {
+	      ss << boost::format("WARN %4f%%") % safe_fill_pct;
+	    } else {
+	      ss << "OK";
```
yehudasa: @mattbenjamin maybe we need to have two modes of operations? one that just dumps everything (like here), another one that just dumps warnings and errors?

src/rgw/rgw_admin.cc
```diff
@@ -4234,6 +4237,42 @@ int main(int argc, char **argv)
...
+		  return -ret;
+	  }
+
+	  bool truncated;
+
+	  /* list users in groups of max-keys, then perform user-bucket
+	   * limit-check on each group */
+	  do {
+		  std::list<std::string> user_ids;
+		  ret = store->meta_mgr->list_keys_next(handle, max, user_ids,
```
yehudasa: @mattbenjamin maybe if uid is passed in the command line, just show stats of a specific user?

src/rgw/rgw_bucket.cc
mattbenjamin: ok

src/rgw/rgw_bucket.cc
mattbenjamin: ack

src/rgw/rgw_bucket.cc
```diff
@@ -1411,6 +1413,116 @@ static int bucket_stats(RGWRados *store, const std::string& tenant_name, std::st
...
+    formatter->open_object_section("user");
+    formatter->dump_string("user_id", user_id);
+    bool done;
+    formatter->open_array_section("buckets");
+    do {
+      RGWUserBuckets buckets;
+      string marker;
+      bool is_truncated;
+
+      ret = rgw_read_user_buckets(store, user_id, buckets,
```
mattbenjamin: ok

src/rgw/rgw_bucket.cc
mattbenjamin: ok, that's what I had originally, then changed it...

src/rgw/rgw_admin.cc
mattbenjamin: ack

ktdreyer:
If we plan to backport this to Jewel, we'll need a ticket at http://tracker.ceph.com/

mattbenjamin:
@ktdreyer Ok;  @yehudasa is this version ready for merge, removing DNM?

src/common/config_opts.h
```diff
@@ -1518,6 +1518,10 @@ OPTION(rgw_realm_reconfigure_delay, OPT_DOUBLE, 2) // seconds to wait before rel
 OPTION(rgw_period_push_interval, OPT_DOUBLE, 2) // seconds to wait before retrying "period push"
 OPTION(rgw_period_push_interval_max, OPT_DOUBLE, 30) // maximum interval after exponential backoff
 
+OPTION(rgw_safe_max_objects_per_shard, OPT_INT, 100*1024) // safe max loading
+OPTION(rgw_shard_warning_threshold, OPT_DOUBLE, 10) // pct of safe max
```
yehudasa: @mattbenjamin let's change this configurable to be the inverse of this one (100-x)

yehudasa:
@mattbenjamin did you push a newer version? I still see my comments unaddressed

ktdreyer:
It would be great to have the ticket URL in the commit log

src/common/config_opts.h
mattbenjamin: ack, otw

mattbenjamin:
@ktdreyer ack

mattbenjamin:
@yehudasa updated

mattbenjamin:
@yehudasa I broke a submodule commit (ec corpus), will repush...again

mattbenjamin:
@yehudasa updated radosgw-admin help.t
The Jenkins build now verifies the radosgw-admin usage output which it had stopped on previously. The current make check failures are unrelated to this change.

@yehudasa this does remind me that I still haven't updated doc/radosgw/admin.rst. I think it makes sense to do so, but it looks like we independently need writeup on bucket (re-)sharding.

dachary:

src/rgw/rgw_bucket.cc
yehudasa: @mattbenjamin should probably set it to 90

yehudasa:
@mattbenjamin sorry for failing to send these comments earlier, new github review workflow was a bit confusing

mattbenjamin:
@yehudasa pushed with requested changes;  I haven't run w/tenants yet, fyi.

yehudasa:
jenkins test this please

oritwas:
@mattbenjamin please rebase

mattbenjamin:
@orit done

src/rgw/rgw_admin.cc
```diff
@@ -4811,6 +4828,50 @@ int main(int argc, const char **argv)
     }
   }
 
+  if (opt_cmd == OPT_BUCKETS_LIMIT_CHECK) {
+    void *handle;
+    std::list<std::string> user_ids;
+    metadata_key = "user";
+    int max = 1000;
+    int ret = store->meta_mgr->list_keys_init(metadata_key, &handle);
```
cbodley: this call isn't needed in the `if (! user_id.empty()) {` case (and will leak memory because `list_keys_complete()` is only called in the `} else {` block)

src/rgw/rgw_admin.cc
```diff
@@ -4811,6 +4828,50 @@ int main(int argc, const char **argv)
...
+    } else {
+      /* list users in groups of max-keys, then perform user-bucket
+       * limit-check on each group */
+      do {
+	ret = store->meta_mgr->list_keys_next(handle, max, user_ids,
+					      &truncated);
+	if (ret < 0 && ret != -ENOENT) {
+	  cerr << "ERROR: buckets limit check lists_keys_next(): "
+	       << cpp_strerror(-ret) << std::endl;
+	  return -ret;
```
cbodley: an early return here will leak memory allocated by `list_keys_init()`. `break;` should work instead

src/rgw/rgw_admin.cc
mattbenjamin: ack

src/rgw/rgw_admin.cc
mattbenjamin: ack

ktdreyer:
Looks like a real test failure in `src/test/cli/radosgw-admin/help.t` ?

src/rgw/rgw_bucket.cc
```diff
@@ -1401,6 +1403,123 @@ static int bucket_stats(RGWRados *store, const std::string& tenant_name, std::st
...
+	/* need info for num_shards */
+	RGWBucketInfo info;
+	RGWObjectCtx obj_ctx(store);
+
+	ret = store->get_bucket_info(obj_ctx, bucket.tenant, bucket.name,
+				     info, nullptr);
+	if (! ret) {
+	  if (info.num_shards > 0)
+	    num_shards = info.num_shards;
+	}
```
cbodley: why ignore errors from `get_bucket_info()` and `get_bucket_stats()` below?

i see that a `continue;` here would skip the `marker = bucket.name;` statement at the end of this for loop. but if you were to move that assignment above this call to `get_bucket_info()`, it would be safe to change these to:
```c++
if (ret < 0) {
  continue;
}
```

src/test/cli/radosgw-admin/help.t
```diff
@@ -231,6 +232,9 @@
      --categories=<list>       comma separated list of categories, used in usage show
      --caps=<caps>             list of caps (e.g., "usage=read, write; user=read")
      --yes-i-really-mean-it    required for certain operations
+     --warnings-only           when specified with bucket limit check, list
+                               only buckets nearing or over the current max
+			       objects per shard value
```
cbodley: indentation here

src/rgw/rgw_bucket.cc
```diff
@@ -1401,6 +1403,123 @@ static int bucket_stats(RGWRados *store, const std::string& tenant_name, std::st
...
+      ret = rgw_read_user_buckets(store, user_id, buckets,
+				  marker, string(), max_entries, false,
+				  &is_truncated);
+      if (ret < 0)
+        return ret;
+
+      map<string, RGWBucketEnt>& m_buckets = buckets.get_buckets();
+
+      for (const auto& iter : m_buckets) {
+	auto& bucket = const_cast<rgw_bucket&>(iter.second.bucket);
```
mattbenjamin: I guess, but I think this is overreaching

src/rgw/rgw_bucket.cc
mattbenjamin: this comment may be important, but I do think it would be more helpful to more clearly state the problem;  I don't need help picking between return continue, or other operators

mattbenjamin:
@cbodley updated again, left a marker assignment at bol

